PR #30, opened 2015-01-15
avoid unnecessary long startup time

fraenki:
re-introduce the short autoboot_delay, avoiding a unnecessary long startup time


fichtner:
This might sound stupid, but I have to ask. ;) Are you ok with licensing this under the BSD 2-Clause license?

A similar patch is needed in the installer and maybe even the tools. I can add them as well so this is not just a "if you click save settings in the GUI it gets better" but an across-the-board change for all images and installations. @jschellevis are you ok with this?

jschellevis:
Lets go ahead with this.

fraenki:
@fichtner Yes, it's OK to use this under the BSD 2-Clause license.

fichtner:
Thanks, I'll pull this in as soon as 15.1.2 is out. :)

fichtner:
Applied via cherry-pick, thank you. :)

PR #135, opened 2015-04-12
adding 1999 BSD license for Whistle Communications

dotike:

I passed this missing licence file while splunking the codebase, appears it got dropped in FreeBSD at some point, (and then the code apparently got dropped).

Their licence is just a somewhat modified BSD style 2-clause.


fichtner:
Applied via git-cherry-pick(8). Thanks. :)

PR #136, opened 2015-04-12
Scrub upstream

dotike:
This pull request is non trivial, and requires a full review of the diff.

If you want to take some parts and leave others, just note what you want and reject it- I'll happily pick through the bits for you.
(Surprisingly, the tedium of this commit is a happy respite compared to the EN .pot file stuff I've been doing in here :)

Best,
.ike


fichtner:
Most of the renaming/pruning that hasn't been done yet is a slippery slope for various reasons:

(1) The bulk load of leftover functions functions are declared in the php-pfSense module which is in ports and can go out of sync with the core causing severe operational mishaps as it does low level interface handling. Those functions need to be rewritten in PHP for portability.

(2) Renaming files still causes hiccups and spurious warnings in the web GUI due to the fact of php-cgi processes not refreshing the include locations.

I'll see what I can pull after a line by line check and apply a partial version of your commit.

fichtner:
.ike, I took what I could without knowingly/potentially breaking the code taking out the file move and and most of the pfsense_*() renaming. Thanks for the patch. More of this is always welcome. We're getting there eventually. :)

    $ git grep -i pfsense | wc -l
         191

PR #149, opened 2015-04-23
Locale.en.canonical catch upstream

dotike:
Still playing catch-up to changes, not quite up to head yet.
This PR is for files files that have been removed and are not longer necessary to translate.

More coming, but I thought it best to get this PR up sooner than later so any early adopter translators don't waste time on deleted lines.


fichtner:
Looking good, thanks! :)

PR #171, opened 2015-05-11
Locale.en.canonical xcatch upstream

dotike:
Long story short: I completely obliterated the English .pot with this automated version, for one good reason:  this paves the way to better automating translation changes in the codebase.

Small tear in my eye for the sweat lost not knowing about xgettext(1), but this provides a perfect template for translators.

Franco pointed out xgettext to me, here,
https://github.com/opnsense/core/commit/87a1e6157a286958fead33e57e84b7ad171ebda4

With that, had I known this tool existed, (and boy did I search for it), I would have saved myself a *lot* of time.

--
Note for Franco: your xargs version is certainly cleaner than my for loop, (and perhaps more robust).  Yet, 2 flag issues are worth being raised now:

  -j "join messages with existing file"
I believe this is what we want long term, to ensure translation messages don't pile up, and that they retain multiple comment lines for source references, etc...  Tell me if I'm wrong, I'm not 100% on how it's behaviour works.

-L PHP , this certainly reduces the C warning/errors, but, not all the files to be translated will be PHP- for example, console based interface setup, etc...
Regardless, I think this issue can be resolved in the future by refining the find/xgettext calls according to the situation- and perhaps just the type of file.


fichtner:
Isaac, thanks for looking at this. :) I'm merging this in, but will play with -j and -L to see if the output can be improved. Some things are weird like different endings for PHP files, shebang usage in rc scripts, etc.

dotike:
Cool- I'm screwing around a bit with your makefile line right now too-

Thanks so much for bringing this to the table- now that it's here, it's quite obviously the way to go :)

dotike:
p.s. - now that a nice one-liner blows away my prior month of toil, I'm delighted to feel a bit like a jackass, as this will be so much easier for future maintinence...  When the payoff is high, I'll happily learn and move on :)

fichtner:
Always excited to learn something new. :)
This is probably needed for -j: https://github.com/opnsense/core/commit/77c4d929373252b6b49f79bed5e9a529ca0f7437

PR #172, opened 2015-05-11
Locale.en.canonical xcatch upstream

dotike:
Adding a few lines to ensure the .pot file isn't so confusing/daunting.

Ostensibly, language translators will start by precisely copying this file- so a bit of OPNsense is good in the header IMHO.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -56,7 +56,7 @@ translate:
 	@: > src/share/locale/en/LC_MESSAGES/OPNsense.pot
 	find src | xargs xgettext -j -L PHP --from-code=UTF-8 -F \
             --copyright-holder="OPNsense project." \
-            --package-name="OPNsense" \
+            --package-name="OPNsense `git describe --abbrev=0`" \
             --msgid-bugs-address="https://forum.opnsense.org/" \
 	    -o src/share/locale/en/LC_MESSAGES/OPNsense.pot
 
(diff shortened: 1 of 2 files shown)


fichtner:
I found a bit of nastiness using your C-format pot file diffed against the PHP-format file generated by the Makefile: https://github.com/opnsense/core/commit/321d458ea07d86417ca0dbc2176ca231ee4b089d

Gonna clean those up before merging this :)

dotike:
Yikes!  Franco: Just nuke the .pot file and generate one using 'make translate'.

Start it from scratch now while there's a clean chance at it!

fichtner:
Hmm, I don't like using git versioning info from there, git might not be installed.
The file is already in version control once checked in, if we use -F the diff is easily readable/mergabe && when we wrap up a package the file will be generated as *.mo anyway. We don't even have to ship the po/pot files in the package.

dotike:
Understood- I'll pull it out now.

I guess the fact that xgettext seems to get the date in there is more than sufficient for translators watching to see how the file tracks the source.

dotike:
Closing- too much for too little.

fichtner:
ACK, will regenerate once more after the current fixing session, pushing -L PHP back (seems to remove some C-format duplication too). I'll let you know. :)

dotike:
10-4 I'll leave the ball in your court from here, and go bootstrap JP...

/salute

PR #173, opened 2015-05-11
Master

dotike:


fichtner:
Isaac, do you want to reroll using the master? I think now it's stable.

dotike:
Early (perhaps premature) submission for Japanese.

Don't feel rushed to pull this PR in, but the header and first few lines are most important, they show how simple it is for translators to get going...

dotike:
ha- we press the button same time, yes- I will now-

PR #174, opened 2015-05-11
jp .pot file kickoff

dotike:
This PR may be a bit early, but it may show other languages how easy it is to get going!

Japanese translation is now officially going, with some way to go...


fichtner:
I'll take it, thanks!!

dotike:
ありがとぢます！

fichtner:
Renamed this to .po ;)

PR #175, opened 2015-05-11
This may need to become "ja_JP.UTF-8" down the road

dotike:


fichtner:
No moar renames for a while now :D

dotike:
whew. :)

Issue #55, opened 2015-02-22
RRD graphs are obnoxiously large


phillijw:
I find it difficult to read with the extra large size of these graphs. The pixelated image actually makes it harder to look at even though it's bigger and easier to see. I would suggest making these a bit smaller.

StopTrain:
I'm in total agreement with this.

cbasolutions:
Pull request submitted.

fichtner:
Merged, thanks! :)

PR #233, opened 2015-07-03
Update authgui.inc

cbasolutions:
Added autocapitalize="off" autocorrect="off" to the input for the logon form usernamefld. When using from an iPhone/iPad the first character of the username is capitalized automatically.

PR #234, opened 2015-07-03
RRD graphs are obnoxiously large #55

cbasolutions:
Removed width=100% and set inline style="max-width:100%; height:auto"


fichtner:
Looks good. Will merge following licensing feedback on #233.

Issue #211, opened 2015-06-09
widgets: system log widget


fichtner:

SashLi:
Hey I created a small widget

![screen_dashboard_system log](https://cloud.githubusercontent.com/assets/10072887/8618104/2d212794-270b-11e5-9730-ac2cbdcaa1f7.png)

Maybe you can take a look at it.

fichtner:
Hi Sascha,

that looks great already. :) A few minor things:

* What we need is a 2-Clause BSD copyright header with your copyright (if you agree to the licensing) 
* A pass for at least PSR 1 coding style: http://www.php-fig.org/psr/psr-1/ (We also have PSR 2, but we can certainly mop that up later.)
* The RSS widget has a hide/show feature for settings. That would be the proper style for the log lines setting and the save button.

Feel free to create a pull request. I will merge this into master with license agreement and PSR 1 fixup.

Thank you,
Franco

fichtner:
Oh, one more thing: Header("Location: /"); should be redirecting to /index.php instead, we saw redirect problems from time to time with the dashboard not reachable (fringe case, but still).

fabianfrz:
Can you also use gettext to make your widget translatable?
See this link for PHP reference: https://secure.php.net/manual/de/function.gettext.php

SashLi:
Could we do a small review. If everything is fine I will create a pull request. Thanks ;-)

fichtner:
Looks good now, thanks! :) Will test/merge/fixup (if needed) with PR.

PR #258, opened 2015-07-15
Small widget that shows the system log #211

SashLi:
A small widget that shows the system log


fichtner:
Merged via git-cherry-pick. Will also be in 15.7.3 (most likely shipping on Friday). Thank you :)

Issue #259, opened 2015-07-15
widgets: submit button text should be translatable


fichtner:

fichtner:
@fabianfrz fixed, thanks a lot! :)

PR #263, opened 2015-07-19
work on #259

fabianfrz:
did some work on issue #259

### DIFF
--- a/src/www/widgets/widgets/carp_status.widget.php
+++ b/src/www/widgets/widgets/carp_status.widget.php
@@ -74,14 +74,14 @@ if ($carp_enabled == false) {
 }
 if ($ipaddress) {
 ?> &nbsp;
-        <?=htmlspecialchars($status);?> &nbsp;
+        <?=htmlspecialchars(gettext($status));?> &nbsp;
         <?=htmlspecialchars($ipaddress);
 }?>
 </td></tr><?php
     }
 } else {
 ?>
-		<tr><td class="listr">No CARP Interfaces Defined. Click <a href="carp_status.php">here</a> to configure CARP.</td></tr>
+  <tr><td class="listr"><?= gettext('No CARP Interfaces Defined. Click <a href="carp_status.php">here</a> to configure CARP.') ?></td></tr>
 <?php
 } ?>
 </table>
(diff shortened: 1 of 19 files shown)


src/www/widgets/widgets/carp_status.widget.php
```diff
@@ -74,14 +74,14 @@
 }
 if ($ipaddress) {
 ?> &nbsp;
-        <?=htmlspecialchars($status);?> &nbsp;
+        <?=htmlspecialchars(gettext($status));?> &nbsp;
```
fichtner: gettext() requires static strings

fabianfrz: there are three values possoble if you look at the file MASTER, BACKUP, INIT and DISABLED - I will solve that with a switch-case

fichtner: It's not worth fixing for now. Typical chicken-and-egg problems. Until we have a good model that wraps all of this we are stuck here (strings vs. integers for state). :)

fabianfrz:
fixed it

fichtner:
No, this makes it worse, 3 lines now for the same string.. if anyone touches a single string the whole thing breaks.

(a) no translation for that particular bit, or
(b) rework the status completely to return integers and only convert to strings on print (fully translated that is).

In order to move forward (a) is probably a the sensible option.

fabianfrz:
it seems that the code of the widget was wrong the function returning the string translated them and the widget used only hard coded strings so a translation file will break this widget - so we found a bug

fabianfrz:
is there still something wrong?

fichtner:
Looks good now, but will commit tomorrow after a final look. :)

fichtner:
squashed all commits into https://github.com/opnsense/core/commit/caea4e1b10e99c5e6115c66c2690801155a8d50e

please also note https://github.com/opnsense/core/commit/9f5be0799f03af09b8f63ad3b61f2cab9ae41fae -- we don't want html in gettext strings. lots of code does it but we'll have to get rid of it eventually. :)

fabianfrz:
I've seen this in the translation files very often so I thought it is common to do this - will do it this way in the future

can we close the related issue?

Issue #269, opened 2015-07-22
GUI: Firewall Logs - wrong icon for "pass"


fraenki:
I've noticed that under _Status -> System logs -> Firewall_ the icon for "blocked" and "passed" traffic is the same:

![pass_not_block](https://cloud.githubusercontent.com/assets/909706/8821740/45a6572c-3060-11e5-8acc-11149ab970ae.png)

I would expect that blocked traffic is indicated by a red "x" and passed traffic by a green tick/check.

fabianfrz:
should be possible - I also dislike the '/'

fabianfrz:
created a pull request - please check it

fabianfrz:
this is a duplicate of #206

fichtner:
Thanks, closing now.

PR #270, opened 2015-07-22
work on #269

fabianfrz:
contains fix for #269 and some language issues in the file

NOTE: Was not able to test it because I get this when I put this into my VM with the stable release:
Fatal error: Cannot redeclare get_port_with_service() (previously declared in /usr/local/www/diag_logs_filter.php:380) in /usr/local/etc/inc/filter_log.inc on line 278


AdSchellevis:
Thanks Fabian, I will pull it in and check the redeclare issue on my box. I'm doing some major restructures so I probably caused this error. sorry about that.

PR #276, opened 2015-07-24
fixed issue also in dynamic view and corrected comment

fabianfrz:
I found the GUI Error also in the dynamic view and fixed it


fichtner:
Thanks :)

fabianfrz:
Sorry this was too late - just found it after the upgrading to 15.7.4 but there is still a problem:
the headings are wrong (and make problems with the translation (e.g. german)):
+ Act -> Action (Reagieren -> Aktion)
+ Time -> Timestamp (Zeit -> Zeitstempel) or Date and Time (Datum und Uhrzeit) because the field contains information like this: Jul 24 17:32:06
+ IF -> Interface (Falls -> Interface) even a collision with another word in english
+ Proto -> Protocol - I don't think the 3 letters are too much text

fichtner:
There's always (more) work. 15.7.5 is scheduled for Monday or Tuesday. Just PR as usual. I'll try to bring everything over to stable/15.7.

Issue #251, opened 2015-07-08
diagnostics/proxy log: needs a face lift


fichtner:
reverse doesn't work, no zebra style

fabianfrz:
created a pull request that converts the pre to a table and added the zebra style but it is still not reverse sortable

fichtner:
Done in https://github.com/opnsense/core/commit/0528a90aaf9132e0e56797f98c794222d3c753de

PR #277, opened 2015-07-25
proxylog pre to table

fabianfrz:
some work on #251

contains no fix for reverse


fichtner:
I'm declining this, even though it's good work. Let me explain: it introduces a third layout type for the log viewer pages and will create more work on said legacy pages that could be rather quickly ported to MVC. There is a lot of duplication going on and that is work that has very little "bang for bucks" in the long run.

I'll gladly take the pre-zapping and reverse log via config option though.

Issue #57, opened 2015-02-22
Aiccu Support


sjorge:
It would be nice to bundle aiccu to also over ipv6 tunnels that use AYIYA instead of heartbeat.

michbu:
+1

fichtner:
Will add it as an optional port, but somebody should help with progressing the GUI integration (if needed).

fichtner:
I've pushed the client to the package mirror (as soon as 15.1.10 is released today). To install:

    # pkg install sixxs-aiccu

The client itself is from 2007, don't expect any wonders. I've also disabled GNUTLS support due to heavy GNU dependencies and I don't want another SSL library in the packages. It should be rewritten to use OpenSSL/LibreSSL by default (if there is enough demand for this I'll do it). Closing this for now. Cheers! :)

mibuthu:
Thanks for inclusion, it is working fine.

The most important missing part is that the interface isn't available in the GUI, so no firewall rules can be set for it. But I found some good hints in [this thread](https://forum.pfsense.org/index.php?topic=62136.0) in the pfsense forum.

I want to do the full GUI integration. I have already tested a bit, it should relatively easy.

But before I goto the details I have some open questions I would like to discuss:
* The interface which will be added is called "tun0" independent of the "ipv6_interface" option value in the config file. There is also a hint in the config file that on BSD the ipv6_interface should be set to gifX for proto-41 tunnels or tunX for AYIYA tunnels. In the link above there is already given a fix for the GUI to include tun interfaces in the web interface. This is working fine but it is also proposed to rename the interface to a more unique name like "aiccu". I would prefer this also, aiccu should respects the given interface name in the config file and I then would use "aiccu" as the name. Would this be a problem? If yes, can the interface name "tun0" can cause problems with other interfaces that are named tunX or is it o.k. to add all tun interfaces to the web UI?
* I would add an additional tab in the interfaces assign screen called "Aiccu". In all other tabs there is an overview of all available interfaces. For aiccu it is only possible to set 1 interface, so the question is if it make sense to have an overview page. I can only allow to create 1 single interface. The alternative would be to remove the overview page at all and only show the options with an enable/disable checkbox. What do you prefer?
* Do you have any other suggestions which I should respect?

Of course it would be great if you can do the SSL parts described above.

fichtner:
@mibuthu thanks for the confirmation and your help :)

[...]

PR #284, opened 2015-07-28
added tun0 interface in interface assign (required for sixxs-aiccu)

mibuthu:
This is a temporary solution to allow using sixxs-aiccu service without
manual code change until the aiccu service is correctly included in the
web interface (issue #57).

### DIFF
--- a/src/www/interfaces_assign.php
+++ b/src/www/interfaces_assign.php
@@ -167,6 +167,19 @@ if (isset($config['ppps']['ppp'])) {
 	}
 }
 
+/* add tun0 interface (required for sixxs-aiccu) */
+//   This is a temporary solution to allow using sixxs-aiccu without manual code change
+//   until the aiccu service is correctly included in the web interface.
+//   'mac' is used to add the description in the available network ports list
+//   (to avoid additional temporary code in the interface_assign_description function).
+$tunfound = "";
+exec("/sbin/ifconfig | /usr/bin/grep -c '^tun0'", $tunfound);
+if (intval($tunfound[0]) > 0) {
+	$portlist['tun0'] = array();
+	$portlist['tun0']['if'] = 'tun0';
+	$portlist['tun0']['mac'] = 'sixxs-aiccu';
+}
+
 $ovpn_descrs = array();
 if (isset($config['openvpn'])) {
 	if (isset($config['openvpn']['openvpn-server'])) {


fichtner:
Sorry I haven't had the time to look at this yet. We're rolling out 15.7.6 today, let's talk about merging this into the development branch for 15.7.7.

Is there a reason why you only use "tun0" and where does "tun0" come from? I can see clashes of tun0 when other things are configured, e.g. OpenVPN.

mibuthu:
Thanks for the info.

In my tests the name of the interface seems to be hard-coded in sixxs-aiccu. Even if a tun0 interface is already available the programs tries to use this interface name and tries overwrites the existing one. So probably this behaviour must be changed in the sixxs-aiccu program.
At the moment I don't think that this is a big issue in opnsense because "normal" tunnel interfaces like OpenVPN are called "ovpn*" at the moment. But if this shall be changed, like you have written it in #57 this would result in clashes.
Because of this I have decided to only check if a tun0 interface is available and if yes add it to the interface assign options. If another tun0 interface would be available which wasn't created by sixxs-aiccu the only issue would be the wrong description which is o.k. for this temporary solution in my opinion. It should be added as an interface assign option anyway.

The above descriptions are the result of my own tests. But probably I have missed something, so please correct me if my conclusions are wrong.

fichtner:
Alright, thanks. :) Works for me, but will not be committed to the stable branch. The package you'll have to use is "opnsense-devel" in this case (pkg install -y opnsense-devel)

fichtner:
WTF. Meh, das war doch der falsche Branch. -.-

fichtner:
Fixed, sorry.

8191:
> In my tests the name of the interface seems to be hard-coded in sixxs-aiccu. Even if a tun0 interface is already available the programs tries to use this interface name and tries overwrites the existing one. So probably this behaviour must be changed in the sixxs-aiccu program.

Actually, aiccu only uses `tun` interfaces for AYIYA tunnels. For static proto-41 tunnels, `gif` interfaces are used.

Also there is the aiccu configuration option `ipv6_interface` which allows defining the interface to be used (e.g. tun9 or gif3), so it's actually not hard-coded `tun0`.

```cfg
# Interface names to use (default: aiccu)
# ipv6_interface is the name of the interface that will be used as a tunnel interface.
# On *BSD the ipv6_interface should be set to gifX (eg gif0) for proto-41 tunnels
# or tunX (eg tun0) for AYIYA tunnels.
ipv6_interface gif0
```

That said, this patch only works for AYIYA tunnels and also only if the `ipv6_interface` is not set.

Since I'm working on a aiccu plugin right now, I'm planning to fix that issue there.

PR #285, opened 2015-07-28
Translation - German 45%

fabianfrz:
sorry, poedit rewrote some lines so the first commit has that many changes.

@fichtner can you ask in the forum for volunteers for german translation.


fichtner:
Fabian, I had some translations in the pipe. Can you rebase your changes? If possible, could you also split the PR so that the "Files changed" section has diff output so we can add comments/discuss?

fabianfrz:
The diff works if you click on a commit but the overall diff does not work - maybe it is too large. I will look if I can split this PR into two

fichtner:
Migrated to #290

PR #288, opened 2015-07-29
minor LDAP fixes

chrisch1974:
usermanager: disable username and descr field for ldap users
usermanager: hide password field for ldap users
import ldap user: fixed a bug on Active Directory where the name attribute isn't handled correctly
import ldap user: sort entries by name
import ldap user: add scrollbars on IE, Firefox & Opera
auth servers: move bind credentials up on the screen


fichtner:
@chrisch1974 great stuff, thank you!!

PR #293, opened 2015-07-31
missing prototype.js in protochart directory

tomcheung789:
fix blank chart on the diag_logs_filter_summary page.

### DIFF
--- a/src/www/protochart/prototype.js
+++ b/src/www/protochart/prototype.js
@@ -1,0 +1,4221 @@
+/*  Prototype JavaScript framework, version 1.6.0.2
+ *  (c) 2005-2008 Sam Stephenson
+ *
+ *  Prototype is freely distributable under the terms of an MIT-style license.
+ *  For details, see the Prototype web site: http://www.prototypejs.org/
+ *
+ *--------------------------------------------------------------------------*/
+
+var Prototype = {
+  Version: '1.6.0.2',
+
+  Browser: {
+    IE:     !!(window.attachEvent && !window.opera),
+    Opera:  !!window.opera,
+    WebKit: navigator.userAgent.indexOf('AppleWebKit/') > -1,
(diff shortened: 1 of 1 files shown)


fabianfrz:
Are you sure that Prototype does not conflict with jQuery?

fichtner:
It's a sensible fix in general, but @fabianfrz is right: we don't want prototype back.

AdSchellevis:
I will remove the include later today.. probably we forgot to remove this one.

fichtner:
And how do we adapt the missing chart?

AdSchellevis:
I thought to replace that with http://d3js.org/ which was an earlier idea of @fabianfrz for some other insights and if possible I would like to standardize on one graphics library (d3js is very powerful).

But replacing the current code costs me a bit more time than I figured, what if?
- we pull this in for now to make the current graphs work
- add an issue assigned to me, to replace the graph library for d3js
- kill prototype when the other issue is solved

fichtner:
ok, if there is an issue in the bug tracker to remember this I'm in :)

AdSchellevis:
Ok, I'm pulling this in then

AdSchellevis:
For future reference, the issue which should remove this again:
https://github.com/opnsense/core/issues/294

fabianfrz:
@AdSchellevis no the page I created did exactly the same but it used a sample code from d3 and I am not sure about the license if we can use it so it is still not used (also used a barchart instead)

AdSchellevis:
@fabianfrz the license of d3js shouldn't be an issue it's a quite similar license as prototype.js. But the implementation side we have to do some work on I guess. Eventually we could also use d3 for dashboard widgets, like http://bl.ocks.org/tomerd/1499279 for example.

I just haven't had the time to dive into it yet.

fabianfrz:
yes but the samples are gists which don't contain license information.

PR #295, opened 2015-07-31
fix menu collapses when select DHCP Server interfaces tab 

tomcheung789:
Services -> DHCP Server


AdSchellevis:
@tomcheung789 Thank you for fixing!

PR #348, opened 2015-08-26
fixes and closes #334

### LINKED ISSUE #334: Display of Routing Table under diagnostics does not show column headers
Display of Routing Table under diagnostics does not show column headers, so it is not clear what the values in the ip4 and ip6 columns are.

fabianfrz:
the problem was that the thead was missing so the headers have not been printed.


fichtner:
Merged manually, thanks :)

PR #358, opened 2015-08-28
Adjust gettext message in password page and fr_FR.po update

gitdevmod:


fichtner:
Maybe we can make the message even clearer by saying something like "Change your password", removing the need to do any kind of username display?

gitdevmod:
I though this page was also for all users,  "Change your password" is better

fichtner:
I removed the column completely, also merged your two newest translation patches. Some additions failed, you need to rebase your branch before you continue (the po and pot file changed).

PR #364, opened 2015-09-01
work on routing table page

fabianfrz:


AdSchellevis:
Hi Fabian,

Thanks again for your support!

I'm working on a full example on how to use the framework, detach services using configd and how to wrap things into plugins.
The sample code is already available (https://github.com/opnsense/plugins/tree/master/devel/helloworld), the wiki page will probably be available tomorrow.
Unfortunately it took a bit long to write things out, so busy with a lot of cleanups and bugfixes. I'm sorry for that. 

Maybe we should look where this component should land eventually,  adding a module only to display the routing table is probably a bit of an overkill ;)
What if we re-factor ShowRoutingTables to Routing and name the action a bit more logical (something like list)? so we leave space for the rest of the routing features in this module.

Next thing to think about is the naming for the actual template (maybe diagnostics) and name the action in the index template something accordingly. 

Cheers,

Ad

AdSchellevis:
The first version of the wiki documentation has arrived: 
https://wiki.opnsense.org/index.php/Develop:Creating_the_hello_world_module

fabianfrz:
Hi @AdSchellevis ,
you can change the code as you like ;)

fabianfrz:
changed files

fichtner:
Your copyrights are missing :)

fabianfrz:
should be done now

fichtner:
Since you've rewritten it, what do you think of https://github.com/opnsense/core/issues/366 ?

fabianfrz:
I also use the output of netstat but if it does not work we should also patch netstat or create a php native extension.

fichtner:
Both options were done by pfSense, both of them are impossible to maintain long-term as we want to run on standard FreeBSD and PHP some day, sorry. :)

I rewrote some function that had the same defect, there must be other tools that do this more gracefully.

Last way out would be a new netstat clone in ports that could do this, but it's also duplicating lots of code.

fabianfrz:
Maybe the author of netstat can add a parameter suppressing this behaviour.

fichtner:
They did, widening the character count from 5 to 7 using -W ;)

https://www.freebsd.org/cgi/man.cgi?query=netstat&sektion=1

Maybe the problem is we use awkward interface renames in the first place? Lots of decisions to make, lots of breakage along the way if we don't respect FreeBSD limitations.

fichtner:
Long story short: no sure what to do without going backwards.

fichtner:
Please note we're not going to solve the netstat issue at all. Too much work for too little benefit.

fichtner:
I'm ok with merging this especially for translation fixes... But as a follow up, there should be a configd backend call that does the actual netstat :)

@AdSchellevis what do you think?

AdSchellevis:
@fichtner please keep this here for now, this definitely needs configd calls before it may enter the code base and the class structure needs some additional thinking to0 (in relation to routing).

As soon as I can find some time, I will pull it in and change things at once.

gonzopancho:
https://redmine.pfsense.org/issues/5484 has all the detail you need to see that the "FreeBSD limitations" you deplore have been lifted.

PR #383, opened 2015-09-14
added user agent to crash reporter

### LINKED ISSUE #382: Add browser version and os to crash report
Would be very useful to have details on the browser and os version when submitting a crash report.
Certainly when an issue is related to css of javascript thing may differ from browser to browser.

fabianfrz:
this should close #382


fichtner:
Thanks! :)

PR #408, opened 2015-09-27
print voucher series

fabianfrz:


fabianfrz:
contains lodash

fabianfrz:
closed it because wrote wrong words

PR #489, opened 2015-11-30
fixes #485: make NAT type configurable for IPsec

### LINKED ISSUE #485: IPsec: make NAT type configurable
Currently the NAT type for an IPsec Phase 2 entry is [chosen automatically](https://github.com/opnsense/core/blob/master/src/etc/inc/filter.inc#L1769-L1773). In some situations this may not be desired.

I think we need a GUI option to choose between _NAT_ and _BINAT_.

fraenki:
This PR adds an option to manually configure the NAT type for IPsec. It's fallback is to use the "auto" mode (to match the previous behaviour), so it shouldn't break anything for existing configurations.

This resolves issue #485.

### DIFF
--- a/src/www/vpn_ipsec_phase2.php
+++ b/src/www/vpn_ipsec_phase2.php
@@ -82,12 +82,21 @@ function pconfig_to_idinfo($prefix, $pconfig)
     $type = $pconfig[$prefix."id_type"];
     $address = isset($pconfig[$prefix."id_address"]) ? $pconfig[$prefix."id_address"] : null;
     $netbits = isset($pconfig[$prefix."id_netbits"]) ? $pconfig[$prefix."id_netbits"] : null;
+    $nattype = $pconfig[$prefix."id_nattype"];
 
     switch ($type) {
         case "address":
-            return array('type' => $type, 'address' => $address);
+            if (!empty($nattype)) {
+                return array('type' => $type, 'address' => $address, 'nattype' => $nattype);
+            } else {
+                return array('type' => $type, 'address' => $address);
+            }
         case "network":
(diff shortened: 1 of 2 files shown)


src/www/vpn_ipsec_phase2.php
```diff
@@ -272,12 +286,19 @@ function getIndexByUniqueId($uniqid)
...
                         $input_errors[] = gettext("A valid NAT local network IP address must be specified.");
-                    } elseif (is_ipaddrv4($pconfig['natlocalid_address']) && ($pconfig['mode'] != "tunnel"))
-                    $input_errors[] = gettext("A valid NAT local network IPv4 address must be specified or you need to change Mode to IPv6");
-                    elseif (is_ipaddrv6($pconfig['natlocalid_address']) && ($pconfig['mode'] != "tunnel6"))
-                    $input_errors[] = gettext("A valid NAT local network IPv6 address must be specified or you need to change Mode to IPv4");
+                    } elseif (is_ipaddrv4($pconfig['natlocalid_address']) && ($pconfig['mode'] != "tunnel")) {
+                        $input_errors[] = gettext("A valid NAT local network IPv4 address must be specified or you need to change Mode to IPv6");
+                    } elseif (is_ipaddrv6($pconfig['natlocalid_address']) && ($pconfig['mode'] != "tunnel6")) {
+                        $input_errors[] = gettext("A valid NAT local network IPv6 address must be specified or you need to change Mode to IPv4");
+                    }
```
fraenki: This is just a style fix and optional.

fichtner: melikes style fixing :)

fraenki:
I've updated my commit to fix a text bug.

PR #496, opened 2015-12-04
firewall: adopt formatting of RFC 1918 and IANA rules

8191:
Adopt style of RFC 1918 rule and reserved IANA rule to match user created
rules in style.

That tiny style inconsistency makes me crazy since using OPNsense the first time... :wink:


AdSchellevis:
Time to pull it in then :) Thanks for your help!

fichtner:
@8191 thanks :) please make sure to set your name and email for git on the box so they end up in the commits

Issue #495, opened 2015-12-04
Inconsistency in firewall rules: RFC 1918 and IANA reserved rules are being logged without "log" icon


8191:
The firewall rules for blocking RFC 1918 addresses (on WAN ports) and the one for blocking IANA reserved addresses are shown without the "log" icon (:information_source:), but the generated pf rules add `log` to the rules.

fichtner:
Interesting, these are settings globally set for both types on System: Settings: Logs/Reports (menu since 15.7.21, logging for both types is on by default)

This is quite funny, because the edit takes you to the interface configuration, but the log settings are elsewhere.

8191:
Maybe this usability issue could be solved by making the icons "block" (:x:) and "log" (:information_source:) clickable and toggle between "block"/"block (disabled)" and "log"/"log (disabled)", respectively, instead of linking to the System: Settings with the pencil icon?

8191:
I've created PR #497 to fix the initial reported issue (missing :information_source: icon).

fichtner:
It's tricky because logging is a global setting, blocking a per-interface thing. Don't want to mess with this before 16.1 is out. The interface page design needs a fundamental rework.

This PR already made it 100% more usable. Close this and pick other low-hanging fruit for the time being? :)

8191:
AH, no I got your point. Yes, that's really tricky.
Maybe showing two pencil icons, one with an log icon and one with the drop icon?
Just collecting ideas here... ;)

fichtner:
I'd probably rather root for a flat solution... move the global log setting to a per-interface setting so the edit button is the right button already?

fichtner:
@8191 I think closing this would be best until we get around to rewrite the logging in MVC?

8191:
Agree.

PR #498, opened 2015-12-04
firewall: show "log" icon for RFC 1918 and IANA rules

8191:
Show a "log" icon if the RFC 1918 rule or the reserved IANA rules
are configured to be logged.

Fixes parts of #495


fichtner:
Merged this manually, thanks! Commit author/email all good now.

Issue #487, opened 2015-11-30
firewall/logs: normal view "Act" column needs a brush-up


fichtner:
The X-is clickable, but counter-intuitive.

fichtner:
@8191 what do you think, close this then? :)

fichtner:
@8191 ping :)

8191:
I'm not totally convinced with the current solution (neither the "normal" nor the "dynamic" view), since it's not completely intuitive. But I cannot imaging a better solution neither... :confused:

fichtner:
True. Ok, let's leave this open.

fichtner:
Was discussed internally, need to close without further sense of direction in favour of a new ticket, as soon as a direction is proposed.

PR #499, opened 2015-12-04
firewall: raise usability for normal log view

8191:
Change cursor when hoovering over symbol within "Act" column of
normal firewall log to indicate that icon may be clicked.

Addresses #487


fichtner:
Funny, the dynamic view already worked this way. Nice.

(merged manually)

PR #503, opened 2015-12-06
firewall: show implicit IPv6 block rule if IPv6 disabled

8191:
If IPv6 is not enabled in global networking settings, show the implicitly
generated IPv6 block rule in floating firewall rules.

See also [this forum post](https://forum.opnsense.org/index.php?topic=1777.0).


fichtner:
Very cool, thanks :)

PR #506, opened 2015-12-07
firewall: show warning banner if firewall disabled

8191:
If firewall is globally disabled, show a warning banner in all relevant
firewall configuration and diagnostic screens (filter and NAT).

### DIFF
--- a/src/www/diag_logs_filter.php
+++ b/src/www/diag_logs_filter.php
@@ -606,6 +606,10 @@ include("head.inc");
 		<div class="container-fluid">
 			<div class="row">
 
+                                <?php if (isset($config['system']['disablefilter'])): ?>
+                                <?php print_warning_box(gettext("The firewall has globally been disabled. Configured rules are currently not enforced."));?>
+                                <?php endif; ?>
+
 				<?php if (isset($input_errors) && count($input_errors) > 0) print_input_errors($input_errors); ?>
 
 			    <section class="col-xs-12">
--- a/src/www/diag_logs_filter.php
+++ b/src/www/diag_logs_filter.php
@@ -606,10 +606,7 @@ include("head.inc");
 		<div class="container-fluid">
 			<div class="row">
 
-                                <?php if (isset($config['system']['disablefilter'])): ?>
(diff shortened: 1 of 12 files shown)


fichtner:
How about removing the firewall section from the menu? Someone disabling the firewall is intentional so it doesn't hurt to scale down complexity. It would also give us more opportunity to clean up the menu in terms of which pages are really firewall pages and which pages are not... There are some candidates e.g. Virtual IP, Traffic Shaper (different subsystem system) and some network diagnostics pages.

8191:
@fichtner Removing the complete menu if disabled also sounds like a good idea. Nevertheless, I think then the setting to enable/disable the firewall needs to be more prominent, so that one can easily check why the menu is not there. If someone is not familiar with OPNsense and he is looking for firewall rules and cannot find it, it's not directly clear that the feature has been disabled. Maybe the firewall menu could be moved to the bottom and being renamed from __Firewall__ to _Firewall (disabled)_?

What about aliases? Are they used outside of the firewall section?

8191:
What about if someone disables a service (e.g. unbound). Is it intended to show a warning banner like this PR? Or do you plan having a centralized page for enabling/disabling features/services?

fichtner:
We can merge this conceptually, but we need to minimise the impact for each page, e.g. no duplication of strings or print_warning_box (print_info_box). The easiest way would be a src/www/firewall.inc which can handle this message and is included on each page where the code now resides. That way we also don't pollute guiconfig.inc any further.

The warnings are good and should probably be there for administrative pages (not diagnostics, logs, status). It should probably point to the respective settings page though...

Unbound/Dnsmasq and other services are different, because these services are triggered from their respective menu entry. We talk about global settings that affect the behaviour of other pages.

Another idea would be to move system Firewall/NAT settings to firewall settings, making it clearer that the firewall can be disabled in the firewall section where it belongs. When the firewall is disabled, and all pages except settings are missing it's 100% clear where to enable it... that's the theory, but then again one could argue that the firewall section should be completely gone.

8191:
@fichtner Worked in your comments. :wink:

fichtner:
Thanks, I'll finish up in the master branch. :)

PR #509, opened 2015-12-07
firewall: extend pfInfo by active rules

8191:
Extends pfInfo by showing active rules incl. match statistics.

Especially for troubleshooting the generated rules by OPNsense are quite useful.


AdSchellevis:
Nice catch! thanks!

PR #510, opened 2015-12-07
tiny bugfix: unbound: fix JS to enable/disable interface selectors

8191:
If unbound is enabled/disabled globally, the interface selectors should be
enabled/disabled as well.


AdSchellevis:
Thanks!

PR #513, opened 2015-12-08
unbound: fix starting of unbound via Status: Services

8191:
unbound could not be started via Status: Services.


fichtner:
Nice catch, been there since we started. :)

Issue #354, opened 2015-08-27
DNS Forwarder


fatchoy:
Hi, it's not an issue but in the services.inc dnsmasq is identified as "DNS forwarder" and Unbound is named "Unbound DNS Forwarder" but in the services list, they are respectively "DNS Forwarder" and "DNS Resolver", I propose to rename the Unbound service to "Unbound DNS Resolver"

fichtner:
On a related note: I like neither the forwarder nor resolver thing to be honest. There's a DNS server that happens to run dnsmasq, there's another one using unbound. The GUIs are different when they shouldn't. In the long run, there will only be one server. It's been hell to have to support both and the associated confusion.

I would rather use something like "Dnsmasq DNS Server" and "Unbound DNS Server"?

fatchoy:
Agreed, my note was just to point out the difference between the naming in the services submenu and the status services items.

fichtner:
Fair enough, and thanks for that. I always try to go a little further so I thought it was good to write this down in a related ticket and leave it here for discussion. :)

The naming fix we can probably get into 15.7.12.

8191:
Which services/functions can unbound not provide at the moment, which dnsmasq can? Or: what keeps you from dropping dnsmasq right now? 

btw: DNS = Domain Name _System_ 😉

fichtner:
Hehe, you got me. :)

Dnsmasq works more reliable. Unbound was at one point imported from packages and you already found an issue with it. The integration of unbound (because of having been a package once) is a bit cleaner though. I don't know what the endgame was for importing unbound, now there's two and nothing happens anymore...

fichtner:
(I realise that doesn't answer your question. I don't know a better one.)

8191:
So the plan is dropping dnsmasq or unbound? Or dropping both for the sake of a new, clean application for dnsmasq/unbound/bind/etc...  :grinning:

fatchoy:
Having the option to run a resolver such as OPNsense is great in those time of censorship. Blockage of website in europe are done with DNS poisoning at the ISP level. 
Having your own resolver make you independent of your ISP decision over the DNS service.

8191:
I think dnsmasq also is capable of resolving, not only forwarding.

fatchoy:
Unfortunately not, dnsmasq rely on external DNS.

fichtner:
overcome by events :)

PR #515, opened 2015-12-08
service status: added references to service config pages

8191:
- Added links to main pages, log pages, and status pages of services shown in the System: Diagnosics: Services page.

- Changed incorrect name of unbound service from "Unbound DNS Forwarder" to "Unbound DNS Resolver"

### DIFF
--- a/src/www/shortcuts.inc
+++ b/src/www/shortcuts.inc
@@ -143,18 +143,28 @@ $shortcuts['dhcp6']['log'] = "diag_logs_dhcp.php";
 $shortcuts['dhcp6']['status'] = "status_dhcpv6_leases.php";
 
 $shortcuts['ipsec'] = array();
+$shortcuts['ipsec']['main'] = "vpn_ipsec.php";
+$shortcuts['ipsec']['log'] = "diag_logs_ipsec.php";
+$shortcuts['ipsec']['status'] = "diag_ipsec.php";
 $shortcuts['ipsec']['service'] = "ipsec";
 
 $shortcuts['openvpn'] = array();
+$shortcuts['openvpn']['log'] = "diag_logs_openvpn.php";
+$shortcuts['openvpn']['status'] = "status_openvpn.php";
 $shortcuts['openvpn']['service'] = "openvpn";
 
 $shortcuts['forwarder'] = array();
+$shortcuts['forwarder']['main'] = 'services_dnsmasq.php';
 $shortcuts['forwarder']['service'] = 'dnsmasq';
 
 $shortcuts['resolver'] = array();
+$shortcuts['resolver']['main'] = "services_unbound.php";
 $shortcuts['resolver']['service'] = 'unbound';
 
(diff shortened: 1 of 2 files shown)


8191:
I've changed a `gettext` obtained string, shall I adopt the language file, or will this be done before release of a new version?

fichtner:
I'm working to get rid of shortcuts.inc. It's trying to do too many things at once and is inconsistent while doing so (some pages navigate to a log page, but the log page is for two facilities so it routes back the wrong way...) and the service start/stop widget ties into this as well.

For now, all items except 'service' should be zapped pending more menu consolidation. It shouldn't be necessary.

For gettext(): if it's necessary it's just changed, the full regeneration is only executed every few weeks to keep the lines changed churn minimal, see this workflow:

https://github.com/opnsense/core/commit/f2e052754f4539d7b066e73ac5a4e13ab84cb153
https://github.com/opnsense/core/commit/08749b414f0c2c5a67737383296abfe503d7dc8f
https://github.com/opnsense/core/commit/1d744002922155c841ab043650be4525cd5f206d

Translation progress from Pootle is mainly merged before an update:

https://github.com/opnsense/core/commit/42c658c73f8c92fa573188a7078ff4666da1267d

Must-see ticket for the DNS resolver/forwarder controversy: https://github.com/opnsense/core/issues/354

8191:
Agreed, the shortcuts.inc is something awful. Nevertheless currently it's a bit awkward finding all the different sections for each service. What about adding these icons until a better way of navigation comes up?

fichtner:
I've kept logs and service entries for now, but killed status and main. If they are not correctly grouped in the menu they should be reordered to be correctly grouped there (Menu.xml).

src/www/shortcuts.inc
```diff
@@ -190,3 +200,11 @@ $shortcuts['squid'] = array();
...
 $shortcuts['squid']['log'] = "diag_logs_proxy.php";
 $shortcuts['squid']['service'] = "squid";
+
+$shortcuts['sshd'] = array();
+$shortcuts['sshd']['main'] = 'system_advanced_admin.php';
+$shortcuts['sshd']['service'] = 'sshd';
+
+$shortcuts['radvd'] = array();
+$shortcuts['radvd']['main'] = 'services_dhcpv6.php';
+$shortcuts['radvd']['service'] = 'radvd';
```
fichtner: are those shortcuts even used? see $shortcut_section assigment on individual pages

fichtner:
Ok, here's what we can do: keep/add log entries where the log entries for shortcuts that are not in the same submenu (e.g. we don't need a squid log shortcut). Fix all currently available shortcuts to appear on pages that are unambiguous and not a log page (e.g. like upnp on the settings and status page)

fichtner:
Thanks, pushed the rename part manually, no more shortcut links though. See https://github.com/opnsense/core/commit/832b801195165b7f2cd31f0186f382edc631a1f4

Issue #511, opened 2015-12-07
Feature Request: Additional DNS Options


majorsl:
I'm interested in having OPNsense take over my DNS duties as it seems to do everything I need for my relatively simple LAN.

I need one host to have an MX record, but I can't seem to find an option to add it. Maybe a simple checkbox "[x] Add an MX Record for this host" could be done at some point?

Alternately, I love *LOVE* how you devs did the "Additional BOOTP/DHCP Options" section on the current release (yay, I can finally have option 252) - something like this for advanced DNS would be pretty sweet.

fichtner:
Are you talking forwarder or resolver?

Forwarder (dnsmasq) has advanced options for the command line, see --mx-host in http://www.thekelleys.org.uk/dnsmasq/docs/dnsmasq-man.html

Resolver (unbound) has advanced options for the config file, see MX in https://forum.teksyndicate.com/t/need-some-help-with-unbound-dns/68740 et al.

We didn't change DHCP options there... Not sure what you mean. :)

majorsl:
Oh, okay, so in Resolver I could just directly add:

local data: "example.com. IN MX 10 mail.example.com."

Sorry, I'm accustomed to hand editing BIND. It appears the TTL is globally set in unbound, yes?

Oh, I was complimenting you all on how well the Additional DHCP options are organized. I was looking for something similar to that for DNS which is why I overlooked the raw text entry box, but that works.

fichtner:
Yes, it's *cache-min-ttl*. No worries. :)

Maybe something could be added to the host overrides. It's interesting that it has never been there in the first place.

8191:
Does not completely fit in here, but what about moving the (very?) uncommonly used feature "Listen Port" from the GUI to the advanced options? I cannot imagine any use-case, where someone would like to configure unbound to a different port than 53...

fichtner:
Agreed, PR welcome. :)

8191:
Regarding the host override: Currently there is a _unique key_ on the _host_ value of each domain override. So it is not possible to add e.g. `host1.domain1` and `host1.domain2`, because both host override entries use the name host portion of the host override.

If we are now thinking of extending the host override feature by oder RRs than A and AAAA, we should maybe think about this restriction: does it make sense?

I personally cannot see a reason why I should not be able to override the same host portion in two different domains with two different IPs.

[...]

PR #519, opened 2015-12-10
unbound: introducing overwriting of MX records

8191:
First PR to #511:

- Split unbound host and domain overwrites from _Services: DNS Resolver: General_ to _Services: DNS Resolver: Overwrites_
- Introducing resource record types
- Allow MX records to be overwritten (in addition to already existing A and AAAA)


src/www/services_unbound_overrides.php
```diff
@@ -0,0 +1,267 @@
+<?php
+
+/*
+	Copyright (C) 2014-2015 Deciso B.V.
```
fichtner: This is definitely wrong. Which file did you copy? And where is *your* copyright? :)

8191: @fichtner I've copied `services_unbound.php`. I guess the pfSense copyright should also go along with that file, since parts of the content were moved from an existing file...

fichtner: Yup, Warren Baker needs to be added.

You can push more commits to the branch to update the pull request.

I will need to get home before I have time to test.

src/etc/inc/unbound.inc
```diff
@@ -518,13 +519,25 @@ function unbound_add_host_entries() {
...
             if (!$added_item[$current_host]) {
-                $host_entries .= "local-data-ptr: \"{$host['ip']} {$host['host']}{$host['domain']}\"\n";
-                if (is_ipaddrv6($host['ip']))
-                    $host_entries .= "local-data: \"{$host['host']}{$host['domain']} IN AAAA {$host['ip']}\"\n";
-                else
-                    $host_entries .= "local-data: \"{$host['host']}{$host['domain']} IN A {$host['ip']}\"\n";
-                if (!empty($host['descr']) && isset($config['unbound']['txtsupport']))
+                /* Backwards compatibility for records created before introducing RR types. */
+                if (!isset($host['rr'])) {
+                    $host['rr'] = (is_ipaddrv6($host['ip'])) ? 'AAAA' : 'A';
```
fichtner: Here you're using "AAAA", but later you're using only "A" for both. If we do resource records, we ought to split A and AAAA.

src/www/services_unbound_host_edit.php
```diff
@@ -58,9 +59,17 @@ function hosts_sort() {
...
+	if (!isset($a_hosts[$id]['rr'])) {
+		$a_hosts[$id]['rr'] = 'A';
+	}
+
 	$pconfig['host'] = $a_hosts[$id]['host'];
 	$pconfig['domain'] = $a_hosts[$id]['domain'];
+	$pconfig['rr'] = $a_hosts[$id]['rr'];
 	$pconfig['ip'] = $a_hosts[$id]['ip'];
+	$pconfig['mxprio'] = $a_hosts[$id]['mxprio'];
+	$pconfig['mx'] = $a_hosts[$id]['mx'];
```
fichtner: The model is a bit too fragile for my taste: MX, like A or AAAA is a property of the domain itself and those entries can coexist, but here it is flat. What if we need multiple records for the same host? Is it wise to handle them "flat"?

fichtner:
Hi Manuel,

Except for the two question the code looks good to me. I am only concerned with backwards compatibility and consistency in the modelling, even though it's just a temporary (0.5 to 1 year by my weak estimate) solution until we can move this to the MVC framework. Sooner or later more people will ask for different records, multiple records of the same type, etc.


Cheers,
Franco

src/www/services_unbound_host_edit.php
8191: @fichtner I'm also not happy with that restriction, but did not change it intentionally yet, since it would change the current behavior of the host overrides. Please see my last comment on #511 regarding that.

src/etc/inc/unbound.inc
8191: The idea was, that the user does not need to care about the difference between `A` and `AAAA` records, since without this PR the user does not need to care about it neither. So the differentiation between A and AAAA is done programmatically and is shown as "A/AAAA".
I have no problem in separating A and AAAA RRs...

8191:
Hi Franco,

please see my comments above.

Regarding modelling (in)consistency; where exactly are your concerns?
Not directly related, but how do you plan migrating an existing "static PHP" service to MVC? Do you plan to migrate the XML configuration to a new structure or reuse the old one?

Regar

8191:
@fichtner I've removed the restriction that a single host cannot be overwritten with two values, since I cannot see any technical justification for that limitation. Also sometimes it might be even required to do so, e.g. DNS round robin.

fichtner:
@8191 cool! I'll try to get this in today :)

8191:
@fichtner Great... :)
What about A/AAAA differentiation? Shall we split it and let the user choose between A and AAAA? Currently OPNsense determines the type by the specified IP address.

fichtner:
Now that multiple host entries are allowed it doesn't matter. What we should do is get this in now and remodel as soon as we start the MVC code for DNS. It make sense to move the model closer to the specific domain model, so A/AAAA split we will do then.

fichtner:
In it went, thank you very much! :)

Issue #431, opened 2015-10-13
fix remaining submit inputs vs. gettext issues


fichtner:
# git grep "<input.*type=[\"']submit[\"'].*gettext" | wc -l
         151

fichtner:
overcome by @AdSchellevis ! thank you :)

PR #547, opened 2015-12-21
fix translation issue; closes #546

### LINKED ISSUE #546: Firewall: NAT: Outbound German Form Bug
Hello,

when you want to activate as example Manual Outbound NAT rule generation
(AON - Advanced Outbound NAT) its not working when the language is set to German (or others?).
I think this is because the formdata includes in the save field "Speichern". Maybe it just apply when save value is "Save"?

Post:
Formdata German (nothing happens)
save:Speichern
mode:advanced

Formdata English (working)
save:Save
mode:advanced

fabianfrz:


fichtner:
The correct fix is explained in #431 and how over hundred of code lines are still wrong...

PR #569, opened 2015-12-27
fix CARP support for OpenVPN client

fraenki:
This fixes the crash report I've just submitted. The function `get_carp_interface_status()` cannot be found otherwise...

### DIFF
--- a/src/etc/inc/openvpn.inc
+++ b/src/etc/inc/openvpn.inc
@@ -29,6 +29,8 @@
 	POSSIBILITY OF SUCH DAMAGE.
 */
 
+require_once('pfsense-utils.inc');
+
 global $openvpn_verbosity_level;
 $openvpn_verbosity_level = array(
 	0 => gettext('0 (none)'),


AdSchellevis:
@fraenki we rather not stack the includes in the old code, to keep the decencies more clear. do you perhaps  know what the original caller was for this issue? (if you do, can you add the include there?)

fraenki:
@AdSchellevis The problematic code is this: https://github.com/opnsense/core/blob/master/src/etc/inc/openvpn.inc#L770

AdSchellevis:
just for reference, this was what the original code dependencies looked like before we restructured all includes:
vpn_openvpn_server php
(dashed lines are duplicates)

now the same file (vpn_openvpn_server.php) looks like this (output of scripts/development/crawl_legacy_deps.py):

vpn_openvpn_server php

AdSchellevis:
@fraenki we need to focus on the usage of openvpn_restart then, let me find the missing part here...

AdSchellevis:
@fraenki this https://github.com/opnsense/core/commit/2458389394a7db637a2ce18c7d2f890f42b2056f should do the trick (openvpn_restart references openvpn_resync).

PR #570, opened 2015-12-27
CARP virtual IP: make sure to save a value for advskew

fraenki:
Currently if we add a CARP virtual IP and leave "advskew" at the default value of "0", then this value will not be saved to the config.xml. This is problematic, because the CARP configuration will be synced with the *same* value of "0" to the secondary node, instead of automatically adding +100 to it.

This leads to some serious issues, because the secondary node will easily become the master. And besides that, the master node will not become the CARP MASTER after a reboot (for example).

I'm not sure if there is a better place to fix this. Maybe something else should be changed to allow saving of the value "0".


AdSchellevis:
@fraenki my mistake, I used an empty() here, but "0" is also treated as empty by php. It's probably better to fix the check in this case. I will push a patch to solve the check for this form.

PR #574, opened 2015-12-27
bring back gateway descriptions

fraenki:
Currently the descriptions for gateways are not visible in `System -> Gateways -> All`, but can still be seen when editing a gateway. This patch brings back the gateway descriptions.

I think they got lost during [a code cleanup](https://github.com/opnsense/core/commit/a453828f3abd249892cf980eed1f62456973e71f#diff-a544724430f51cdfcf966a33f5b8549f).


AdSchellevis:
Thanks for spotting this one.

PR #575, opened 2015-12-27
bring back loadbalancer descriptions

fraenki:
Currently loadbalancer descriptions are configured with white (#FFFFFF) font. :-) This patch removes the `<font>` tags and relies on CSS to provide the required font color.


fichtner:
Nice catch. They were white ever since we forked, I don't even want to know why. :)

PR #578, opened 2015-12-28
fix label for option Source/Invert

fraenki:
In `Firewall->Rules->Edit` the option "Source / Invert" is erroneously labeled "Destination / Invert". This patch corrects the label.


AdSchellevis:
Thanks for pushing this one.

PR #587, opened 2015-12-31
change dialog style

fabianfrz:


fichtner:
looks good to me, but @AdSchellevis should decide here

AdSchellevis:
@fabianfrz @fichtner I think I have some more committed the last days, but your right, it should have danger style in stead of info. I will try to pay closer attention on those and pull this one in.

Issue #576, opened 2015-12-27
Style issues and missing translation


DokuKaefer:
system_advanced_network.php
Line 139
	"class="text-warning" is missing at Note

system_authservers.php
Line 285
	- missing translation
Line 483
	- nothing for "showhelp"-function

System -> Access Control -> Servers -> click on "Add Servers"
	- the Button for "Add Server" is still available
	- the Button for "full help" is missing
	- the table on the right side jumps to left when opening a help text at:
		- Hostname or IP address
		- Authentication containers

![without_help_text](https://cloud.githubusercontent.com/assets/11537653/12012488/62b28c6e-acf7-11e5-81e9-04fa601c9d25.png)

![show_help_text](https://cloud.githubusercontent.com/assets/11537653/12012489/6f9b5b36-acf7-11e5-8a2a-2951b9741670.png)

system_advanced_admin.php
Line 402
	"class="text-warning" is missing at Note

diag_logs_settings.php
Line 402 and 462
	"class="text-warning" is missing at Note

AdSchellevis:
@DokuKaefer can you recheck the current master branch? I walked through the first few issues and they seemed to be solved already..

AdSchellevis:
Looks fixed, if anything is missing feel free to open a new issue for that part.

PR #593, opened 2016-01-01
fix 576

fabianfrz:
@DokuKaefer please check

### DIFF
--- a/src/www/diag_logs_settings.php
+++ b/src/www/diag_logs_settings.php
@@ -399,7 +399,7 @@ $(document).ready(function() {
                       <div class="hidden" for="help_for_logfilesize">
                         <?=gettext("Logs are held in constant-size circular log files. This field controls how large each log file is, and thus how many entries may exist inside the log. By default this is approximately 500KB per log file, and there are nearly 20 such log files.") ?>
                         <br /><br />
-                        <strong><span class="text-danger"><?=gettext("NOTE:");?></span></strong>
+                        <strong><span class="text-warning"><?=gettext("NOTE:");?></span></strong>
                         <?=gettext("Log sizes are changed the next time a log file is cleared or deleted. To immediately increase the size of the log files, you must first save the options to set the size, then clear all logs using the \"Reset Log Files\" option farther down this page. "); ?>
                         <?=gettext("Be aware that increasing this value increases every log file size, so disk usage will increase significantly."); ?>
                         <?=gettext("Disk space currently used by log files: ") ?><?= exec("/usr/bin/du -sh /var/log | /usr/bin/awk '{print $1;}'"); ?>.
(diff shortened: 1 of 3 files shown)


src/www/diag_logs_settings.php
```diff
@@ -399,7 +399,8 @@ function check_everything() {
                       <div class="hidden" for="help_for_logfilesize">
                         <?=gettext("Logs are held in constant-size circular log files. This field controls how large each log file is, and thus how many entries may exist inside the log. By default this is approximately 500KB per log file, and there are nearly 20 such log files.") ?>
                         <br /><br />
-                        <?=gettext("NOTE: Log sizes are changed the next time a log file is cleared or deleted. To immediately increase the size of the log files, you must first save the options to set the size, then clear all logs using the \"Reset Log Files\" option farther down this page. "); ?>
+                        <strong><span class="text-danger"><?=gettext("NOTE:");?></span></strong>
```
fichtner: I neiher recommend the uppercase, nor the "danger/strong" combo for the word. If it's a note it's enough as it was (plus lower-case, if more severe than a note (warning or danger, the text "Warning:" or "Danger:" can be used.... but not a danger colour with a note...

src/www/system_authservers.php
```diff
@@ -465,7 +465,8 @@ function select_clicked() {
                   <td>
                     <input name="ldap_host" type="text" id="ldap_host" size="20" value="<?=$pconfig['ldap_host'];?>"/>
                     <div class="hidden" for="help_for_ldap_host">
-                      <?= gettext("NOTE: When using SSL, this hostname MUST match the Common Name (CN) of the LDAP server's SSL Certificate."); ?>
+                      <span class="text-warning"><strong><?=gettext("Note:");?>&nbsp;</strong></span>
```
fichtner: Same here, is it a note or a warning?

src/www/diag_logs_settings.php
fabianfrz: This is what ticket #576 suggests. @DokuKaefer should answer this question.

DokuKaefer: I found it under System - Networking
Several "Notes" are orange (see screenshot)
But the first "Note" at "Allow IPv6" is black.
Wouldn't it be better if all notes would be in orange?

networking_settings_system

fichtner: I'd like no more colours, no more uppercase for those leading words. It leads to exaggerated exposure of minor notes (unless they are not minor, but then they should neither be notes nor explanation text above is conceptually correct hiding important details in further notes). I'm also against splitting up the "Note:" and the rest of the text, creating multiple gettext calls, it loses context for translation.

fabianfrz: @fichtner @DokuKaefer close this pull request?

fichtner:
Sorry I spent too many hours at the doctor's the past few days... @fabianfrz I'd like a fix to go in, but the patches should go into simplifying and unifying the code that we have. :)

PR #614, opened 2016-01-07
user manager: fixed button titles

8191:


fichtner:
Looking good, thank you.

Issue #606, opened 2016-01-06
UX: change icon "add new based on this one"


8191:
In the Firewall: Rules screen the button "add new rule based on this one" uses a :heavy_plus_sign: icon. In my opinion that's not really straightforward, since I'd expect something like "add new rule below/above this one" when seeing the :heavy_plus_sign: icon.

What about changing the icon to something more intuitive like a copy or clone icon (e.g. [something like this](http://fortawesome.github.io/Font-Awesome/icon/clone/))?

Similar functions use the same icon, e.g.: copy IPsec P1/P2 rules, copy gateways, etc...

AdSchellevis:
Hi Manuel,

Those are probably quite some icons to replace, but it's a good idea to choose a different icon for the copy feature to make it a bit more intuitive.

Do you have time to do this replacement?

Regards,

Ad

8191:
Hi Ad, 
Yes, I can create a PR for that. Just wanted to clarify if it's also in your intention changing the icons. 

Manuel

fichtner:
Colours may be counter-intuitive and create problems in other themes. It would be better to use a copy icon, namely fa-copy. :)

8191:
@fichtner I'm afraid there is [no such icon](http://fortawesome.github.io/Font-Awesome/icon/copy/) named `fa-copy` in Font Awesome... But there is [`fa-files-o`](http://fortawesome.github.io/Font-Awesome/icon/files-o/).

fichtner:
Whoops, should have been fa-clone :)

8191:
Is it intended to migrate the glyphicons to font awesome, or can they co-exist at one page?

fichtner:
font awesome took over the menu (except for the firewall icon, the font awesome one looks blunt), but glyphicons are widely used for action buttons. they must coexist anyway :)

fichtner:
@8191 Will you close? Other smaller things we can merge without a specific PR now since you've established a great baseline now.

8191:
Yes, the initial request/report was merged.

PR #616, opened 2016-01-07
Reworked action buttons for list operations

8191:
Reworked action buttons of several lists:
* IPsec phase 1 and phase 2 entries
* Firewall rules
* NAT rules (all 4 pages)
* Firewall schedules
* Virtual IPs
* Gateways
* Gateway groups
* Routes

Changes applied (to each page):
* Changed plus icon of "add new item based on this one" to a clone icon; changed text to "clone item" (#606)
* Applied consistent button order of move, remove, add, clone
* Changed tooltips "click to toggle enable/disable rule" to "disable rule" and "enable rule" corresponding to the current state, respectively
* Added missing tooltips for buttons

Additionally the title of the manual NAT rules table was changed from "Mappings" to "Manual rules", since the automatically generated rules are called "Automatic rules" and also the NAT mode only uses the word "rules" and never "mapping".

### DIFF
--- a/src/www/firewall_nat.php
+++ b/src/www/firewall_nat.php
@@ -303,7 +303,7 @@ $( document ).ready(function() {
 <?php                 endif; ?>
                       </td>
                       <td>
-                        <a href="#" class="act_toggle" id="toggle_<?=$nnats;?>" data-toggle="tooltip" data-placement="left" title="<?=gettext("click to toggle enabled/disabled status");?>">
+                        <a href="#" class="act_toggle" id="toggle_<?=$nnats;?>" data-toggle="tooltip" data-placement="left" title="<?=(!isset($natent['disabled'])) ? gettext("disable rule") : gettext("enable rule");?>">
 <?php                     if (!empty($natent['associated-rule-id'])): ?>
 <?php                     if(isset($natent['disabled'])):?>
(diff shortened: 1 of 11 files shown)


fichtner:
@AdSchellevis will you take this?

AdSchellevis:
@fichtner no problem, I will check and pull in.

AdSchellevis:
@8191 nice work! this really improves consistency.

fichtner:
I agree. Since we're here, the data placement for the tooltips (left) is a awkward with (almost always) buttons further left. Can we change all of those to top (default)?

8191:
@fichtner Yes, that would be a good idea, since sometimes the tooltip covers buttons left of the currently highlighted button.

Also one more thing: currently both, the X and the "trash bin" icon are used for deleting records (mostly the X, but at least for "delete API key" in the user manager it's the trash icon). I personally find the trash bin icon more readable for "delete" than the X. What's your opinion on that?

fichtner:
trash bin

AdSchellevis:
+1

8191:
OK, so I'll also unify the delete actions to trash bins.

Can I still modify my already merged branch `8191:change_copyicons` or do I need to create a new one?

fichtner:
The PR was merged, best rebase the branch, then add new commits and a new PR :)

Issue #624, opened 2016-01-08
Bug: Registered DHCP leases in DNS have wrong domain name


8191:
The feature "register DHCP leases in DNS" registers the DHCP leases with wrong domain name. The domain name of the OPNsense system is used instead of the configured domain name of the DHCP service.

AdSchellevis:
@8191 I don't think we can fix this without glueing too many components together, with our latest changes we collect and monitor all the leases, but I'm not sure how we should determine the dhcp pool they belong too easily. Any suggestions?

AdSchellevis:
timeout

PR #625, opened 2016-01-08
unbound: Fix "register DHCP leases in DNS" config option

8191:
The option to register DHCP leases within unbound DNS was configurable via
DNS Resolver GUI, but the setting was ignored. DHCP leases were always applied.

There is still a bug in the "register DHCP leases in DNS" feature which might be related to this, but I'm not yet sure if it's a bug in the dhcpd or in unbound config generators: #624


fichtner:
Indeed, great catch! :)

PR #626, opened 2016-01-09
Enhance firewall rule display

8191:
Fixes #493, show tooltips with alias and schedule description within firewall rules and various NAT pages.

PR #628, opened 2016-01-10
Unify action buttons of lists

8191:
* Changed "X" icon for delete action to "trash bin" for all delete actions, which modify the config (e.g. remove firewall rule). This means, e.g. removing a DHCP lease or a state from the pf state table still has an "X" and not a "trash bin" icon. #616

* Moved position of most tooltips from left (most) to top. Not moved were tooltips of horizontal button rows (e.g. user manager: add/remove groups) and the services table. #616 

_Unfortunately this PR depends on #626._

### DIFF
--- a/src/etc/inc/util.inc
+++ b/src/etc/inc/util.inc
@@ -1004,6 +1004,28 @@ function alias_expand($name) {
 	}
 }
 
+/* get description of alias */
+function get_alias_description($name)
+{
+    global $config;
+
+    if (is_alias($name))
+    {
+        if (isset($config['aliases']['alias']))
+        {
+            foreach ($config['aliases']['alias'] as $alias)
+            {
+                if ($alias['name'] == $name)
+                {
+                    return (isset($alias['descr'])) ? $alias['descr'] : "";
+                }
+            }
+
+            return null;
+        }
+    }
+}
+
 function subnet_size($subnet) {
 	if (is_subnetv4($subnet)) {
 		list ($ip, $bits) = explode("/", $subnet);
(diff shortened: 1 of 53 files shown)


fichtner:
please don't base PRs on other PRs, it creates really messy histories. I will cherry-pick the commits instead here for clarity

AdSchellevis:
oops, I placed some comments in the other commit, but to be sure. can we move get_alias_description to legacy_bindings.inc and skip the is_alias there?

fichtner:
The stacked PRs do not even play nice with github. this one should go in first, the others need to be rebased afterwards.

fichtner:
Merged manually. Please rebase your other PR :)

8191:
@AdSchellevis I'll move `get_alias_description` to legacy_bindings.inc, but what do you mean by _skip is_alias_?

AdSchellevis:
@8191 thanks, the is_alias checks the global $aliastable for key existence, which in return is populated by the same data that the get_alias_description is traversing. 
If performance is a bottleneck (which I don't expect) it might be a good idea to combine actions somehow (build a class which keeps it's own reference in stead of trusting application globals).
So my strategy would be to first try the code without caching, then if needed solve those.
For profiling, I use webgrind from time to time which can easily be installed on OPNsense.
Eventually I would like to kill all globals (except $config, which is too tangled into the codebase).

8191:
@AdSchellevis Thanks for explanation, I just found your post in the other commit comment. ;)
I've removed the `is_alias` call and moved the function to legacy_bindings.inc.

Issue #106, opened 2015-03-25
NAT: automatic vs. manual mappings


fichtner:
"Well, either Automatic doesn't show the manual mapping a at all, but perhaps "remembers" them when coming from Manual, or do allow the manual mapping's to be used before the automatically generated ones. (With an option to enable/disable them)"

via https://forum.opnsense.org/index.php?topic=229.msg931#msg931

8191:
@fichtner Was this closed intentionally? Currently I still can see the manual rules when using automatic outbound NAT.

fichtner:
No activity for too long so I closed it, but if you take this I can reopen?

8191:
Sure, I'll take it.

fichtner:
thank you :)

PR #629, opened 2016-01-10
Enhance outbound NAT

8191:
Hide _Manual rules_ of outbound NAT if modes "automatic" or "disabled" are chosen, which do not obey manual rules. #106

### DIFF
--- a/src/etc/inc/util.inc
+++ b/src/etc/inc/util.inc
@@ -1004,6 +1004,28 @@ function alias_expand($name) {
 	}
 }
 
+/* get description of alias */
+function get_alias_description($name)
+{
+    global $config;
+
+    if (is_alias($name))
+    {
+        if (isset($config['aliases']['alias']))
+        {
+            foreach ($config['aliases']['alias'] as $alias)
+            {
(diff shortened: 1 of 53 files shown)


src/etc/inc/util.inc
```diff
@@ -1004,6 +1004,28 @@ function alias_expand($name) {
 	}
 }
 
+/* get description of alias */
+function get_alias_description($name)
```
AdSchellevis: can we move this function into legacy_bindings.inc? it helps to identify the additions to the legacy code better.

src/etc/inc/util.inc
```diff
@@ -1004,6 +1004,28 @@ function alias_expand($name) {
...
+{
+    global $config;
+
+    if (is_alias($name))
+    {
```
AdSchellevis: it should be save to ditch is_alias, the block below does approx. the same thing :)

fichtner:
cherry-picked the remaining chunk, also changed the "or" according to the comment earlier. Thanks a bunch! :)

PR #630, opened 2016-01-10
changed file type

### LINKED ISSUE #623: Created API key cannot be downloaded using Google Chrome
At least Google Chrome blocks the ini file which gets downloaded when creating an API key for a user.

I don't know if there's actually a use-case for the API keys right know, but as soon as there is one we should address this one, I'd say, since the ini file contains the secret, which is not shown within the GUI.

fabianfrz:
@8191 should test this before it should be merged.
If it works this closes #623


8191:
@fabianfrz Works for me. Thanks!

PR #632, opened 2016-01-10
Outbound NAT: enhanced description of NAT modes

8191:
Removed overlong information box of outbound NAT rules and replaced it with short descriptions within the configuration options. Also small HTML enhancement to easier select the options by clicking the text instead of clicking directly the radio button.


fichtner:
Cherry-picked, many thanks! :)

PR #660, opened 2016-01-13
Menu: fixed menu for IPsec Mobile Clients after apply

8191:
VPN: IPsec: Mobile Clients looses menu hierarchy after clicking "Apply"


fichtner:
Nice catch, thanks!

Issue #653, opened 2016-01-12
Bug: OpenVPN Client Export only exports singing CA of server certificate without its chain


8191:
The OpenVPN client export (VPN: OpenVPN: Client Export) only exports the direct CA singing the server certificate without the CA's chain. The client then fails to verify the exported CA, if it was not self-signed (i.e. it's an intermediate CA).

fichtner:
@8191 I'm leaving this open for a week to collect more reports before closing. Worst case, we're reverting for 16.1, but I don't think we need to do that. :)

fichtner:
All good, many thanks. :)

PR #662, opened 2016-01-13
OpenVPN Client Export: add CA chain to CA certificates

8191:
Addresses #653

I think this PR requires some more testing, since the modified function is used for several different client packages to be exported. I only tested some of them, maybe someone can help with testing the Windows Installers?

Generally, the [OpenVPN man page](https://openvpn.net/index.php/open-source/documentation/manuals/65-openvpn-20x-manpage.html) states:

> `--ca file`
> `Certificate authority (CA) file in .pem format, also referred to as the root certificate. This file can have multiple certificates in .pem format, concatenated together.`

So _theoretically_ there should not be a problem with any implementation...

### DIFF
--- a/src/etc/inc/openvpn-client-export.inc
+++ b/src/etc/inc/openvpn-client-export.inc
@@ -95,7 +95,7 @@ function openvpn_client_export_validate_config($srvid, $usrid, $crtid) {
 	{
 		$input_errors[] = gettext("Could not locate server certificate.");
 	} else {
-		$server_ca = isset($server_cert['caref']) ? lookup_ca($server_cert['caref']) : null;
+		$server_ca = isset($server_cert['caref']) ? str_replace("\n\n", "\n", str_replace("\r", "", ca_chain($server_cert))) : null;
 		if (!$server_ca) {
 			$input_errors[] = gettext("Could not locate the CA reference for the server certificate.");
 		}
@@ -333,7 +333,7 @@ function openvpn_client_export_config($srvid, $usrid, $crtid, $useaddr, $verifys
 			file_put_contents("{$tempdir}/{$prefix}.ovpn", $conf);
 
 			$cafile = "{$tempdir}/{$cafile}";
-			file_put_contents("{$cafile}", base64_decode($server_ca['crt']));
+			file_put_contents("{$cafile}", $server_ca);
 			if ($settings['tls']) {
 				$tlsfile = "{$tempdir}/{$prefix}-tls.key";
 				file_put_contents($tlsfile, base64_decode($settings['tls']));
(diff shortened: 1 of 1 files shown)


src/etc/inc/openvpn-client-export.inc
```diff
@@ -95,7 +95,7 @@ function openvpn_client_export_validate_config($srvid, $usrid, $crtid) {
 	{
 		$input_errors[] = gettext("Could not locate server certificate.");
 	} else {
-		$server_ca = isset($server_cert['caref']) ? lookup_ca($server_cert['caref']) : null;
+		$server_ca = isset($server_cert['caref']) ? str_replace("\n\n", "\n", str_replace("\r", "", ca_chain($server_cert))) : null;
```
fichtner: This suggest ca_chain() does a have a weird output format oO

8191: @fichtner Sorry? Didn't get that... :S

fichtner: I mean I'd rather have ca_chain() return sane output (the transformation makes sense, that's how the output should look) or give it a switch for sane output?

8191: The `ca_chain()` function adds an empty line after each cert. I've adopted the double-`str_replace()` from the [CaptivePortal](https://github.com/opnsense/core/blob/b83fed8cfcc655f07db931e3961e6a92b85318cc/src/opnsense/scripts/OPNsense/CaptivePortal/generate_certs.php#L59). Didn't want to change the function, since it's used in the web gui config, too.

fichtner: Ok, I'll mop it up later. :)

8191: I've just tested the certificate file of lighttp without the line break and it also works. I'll add the `ca_chain` sanitation to my PR...

8191: oh... you were faster. ;)

fichtner:
FWIW, patch looks very good, yet can't test this right now. I'll let this sit here for a day and talk to a few people about it.

src/etc/inc/openvpn-client-export.inc
fichtner: it's yours if you want

fichtner:
Tested by: @stoptrain

8191:
Maybe let's keep that PR unmerged until #664 was resolved?

fichtner:
I think #664 can wait. I'll try to get two more testers to be sure and it'll stay on the devel branch for 15.7.25.

fichtner:
Yay for progress, thanks Manuel. I'll keep this off 15.7.25 to make sure we get some testing before we take this into 16.1.

PR #663, opened 2016-01-13
CA Generation: fix internal intermediate CA creation

8191:
Found a bug while working on something else: Internal intermediate CAs were not correctly referred back to the signing CA.


AdSchellevis:
Nice catch

Issue #620, opened 2016-01-08
Feature: Firewall: also show floating rules on interface tabs


8191:
Actually I thought that was already introduced some time ago, but never actually checked it. So it turns out, that currently this feature is already "implemented" as a bug in some cases (i.e. only one assigned interface).
But anyway, I think it would be handy to merge interface-only and floating rules at the interface tabs, since one might overlook a floating rule when creating new rules or investigating existing rules. The rules of cause should be tagged somehow as floating.

oparoz:
I agree, maybe in a separate block, like in the outbound nat rule panel?

AdSchellevis:
timeout

PR #666, opened 2016-01-13
Firewall: show floating rules on all affected interfaces

8191:
This needs some discussion. Basically its a fix for #620, but since no one responded on the feature request, maybe it's not even appreciated.

Changes:
  * Show floating rules on each interface tab it was configured to
  * Show "floating rule" icon in front of floating rules when an interface tab
    is selected
  * Indicate non-quick floating rules on interface tab (since all other rules
    are quick, quick rules are not indicated)

My thoughts about that: hiding floating rules on the interface tabs is error potential, since they are still applied but not show.
Also we are getting very many different icons on the rules page. Maybe we should switch to some other indication like background color or having a separate table for floating rules?!

### DIFF
--- a/src/www/firewall_rules.php
+++ b/src/www/firewall_rules.php
@@ -373,7 +373,7 @@ $( document ).ready(function() {
 <?php
                 $interface_has_rules = false;
                 foreach ($a_filter as $i => $filterent):
-                if ( (isset($filterent['interface']) && $filterent['interface'] == $selected_if) ||
+                if ( (isset($filterent['interface']) && in_array($selected_if, explode(',', $filterent['interface']))) ||
                      (isset($filterent['floating']) && $selected_if == "FloatingRules" )):
                   $interface_has_rules = true;
                   // select icon
@@ -422,6 +422,9 @@ $( document ).ready(function() {
                   <tr ondblclick="document.location='firewall_rules_edit.php?id=<?=$i;?>';">
                     <td>
                       <input type="checkbox" name="rule[]" value="<?=$i;?>"  />
+<?php                 if ($selected_if != "FloatingRules" && isset($filterent['floating']) && $filterent['floating'] === 'yes'): ?>
+                      <i class="fa fa-random" data-toggle="tooltip" title="<?= gettext('floating rule') ?>"></i>
+<?php                 endif; ?>
                     </td>
                     <td>
                       <a href="#" class="act_toggle" id="toggle_<?=$i;?>" data-toggle="tooltip" title="<?=(empty($filterent['disabled'])) ? gettext("disable rule") : gettext("enable rule");?>"><span class="glyphicon <?=$iconfn;?>"></span></a>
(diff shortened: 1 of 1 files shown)


fichtner:
Sorry, I simply forgot. This can be in 16.1 pending some review. Good night for now :)

8191:
> Sorry, I simply forgot

Should not have been an accusation... ;) I was not sure if this feature is intended to exist, since you've mentioned that it is a bug that floating rules with only one assigned interface are shown in the corresponding interface section.

src/www/firewall_rules.php
```diff
@@ -373,7 +373,7 @@
 <?php
                 $interface_has_rules = false;
                 foreach ($a_filter as $i => $filterent):
-                if ( (isset($filterent['interface']) && $filterent['interface'] == $selected_if) ||
+                if ( (isset($filterent['interface']) && in_array($selected_if, explode(',', $filterent['interface']))) ||
```
fichtner: If no interfaces are set for a floating rule, all interface match, that currently doesn't happen

src/www/firewall_rules.php
```diff
@@ -422,6 +422,9 @@
                   <tr ondblclick="document.location='firewall_rules_edit.php?id=<?=$i;?>';">
                     <td>
                       <input type="checkbox" name="rule[]" value="<?=$i;?>"  />
+<?php                 if ($selected_if != "FloatingRules" && isset($filterent['floating']) && $filterent['floating'] === 'yes'): ?>
+                      <i class="fa fa-random" data-toggle="tooltip" title="<?= gettext('floating rule') ?>"></i>
+<?php                 endif; ?>
```
fichtner: We should make a second line for the icon key table, split the display in half and add a floating icon on/off line like all the other icons do.

fichtner:
I am really worried about reordering. The idea was that each tab makes for a block of rules that are evaluated in order (floating rules last).

If we have floating rules in the interface section one can rearrange floating rules on top of interface rules.

My gut feeling here is to either leave it as is and fix the leakage of floating rules into interface tabs (short term, maybe for 16.1) and later remove the need for floating/interface distinction altogether (mid-term).

AdSchellevis:
@fichtner I also think we should keep the behavior as it is and fix the leakage. Ordering of rules belonging to different interfaces will always lead to strange results.

8191:
Maybe we can have a optical separation between the interface and floating rules? And then prevent the moving of rules between the sections.

fichtner:
All of this without adding code would be fantastic, but I think it doesn't work without adding code that we hope to get rid of again eventually, which then maybe never happens.

16.1 is almost out, there's a chance we can move this way:

* Remove the quick-only restriction from interface rules.
* Merge all the interface tabs into one screen (*)
* Let the user do whatever he wants (*)

Or this: let it be the advanced edit mode which needs to be enabled in the settings (*)

All in all, we really ought to fix the leakage first, whatever we do next. :)

fichtner:
Also, we need to come up with a definition of a floating rule: what is a "floating rule" actually ;)

fraenki:
> Also, we need to come up with a definition of a floating rule: what is a "floating rule" actually

Hm? A floating rule may apply to *multiple* interfaces, whereas all other rules are specific to only *one* interface. Or isn't it that simple?

fichtner:
So why can we add floating rules that apply to one interface at all? The point is, it's a completely arbitrary definition with a big hole. The former distinction of not being "quick" by default somehow helped the definition of "floating" as they would match last if any, but that's pretty arbitrary as well. The kicker is that the order for rules is already wrong. If one creates a floating rule for WAN and then creates a new rule for WAN the floating rule will be on top. Really, I'd just like to get rid of the notion of "floating", unify the rules screen and simply keep the interface tabs as a filter device. Wouldn't break anything underneath and then one is free to use the rules in the most flexible way without invisibly shooting oneself in the proverbial foot.

fraenki:
> unify the rules screen and simply keep the interface tabs as a filter device

:+1: Makes sense to me. The key to not get lost in all the rules (on one screen) is a really good filter. With a subpar filter this could easily become really annoying. Maybe move this to 16.7 to have some time for testing on some devel releases.

fichtner:
floating rules are a continuous annoyance in the code, this can't be merged as is since there are other fixes in the code now due to other side (design, really) effects.

PR #696, opened 2016-01-22
use netaddr to lookup mac address in list_arp.py and list_ndp.py

gitdevmod:
net/py-netaddr package required

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -97,6 +97,7 @@ CORE_DEPENDS?=		apinger \
 			py27-requests \
 			py27-sqlite3 \
 			py27-ujson \
+			py27-netaddr \
 			python27 \
 			radvd \
 			rate \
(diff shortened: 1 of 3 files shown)


fichtner:
@gitdevmod Thanks, I like this solution. Will you add the dependency to the main Makefile for completeness? It's as simple as that :)

Also, do we really need to import all (*) from netaddr?

gitdevmod:
I do not know about the import all need, I just followed the documention.

AdSchellevis:
@fichtner @gitdevmod before we merge this one, I think we should check where / how this library gets it's data. I kind of missed the piece that the netmap organisation list wasn't installed, so I used the one available on my box. I'm not against using this lib, (without the import *) as long as it doesn't require an internet connection to fetch the manufacturer every time.

fichtner:
I checked with @gitdevmod yesterday, the python module installs flat files (package is 1 mb packed, 5 mb unpacked). no online querying, I like this solution better than handling nmap or merging its file into core.git

AdSchellevis:
@fichtner Ok, I couldn't find it in the original source, but if it's in I like this one also. If you need help fixing the import, just let me know.

fichtner:
Looks good, small issue with an exception when mac is unknown, but will follow in the next commit.

thanks @gitdevmod :)

PR #703, opened 2016-01-25
translation fixes

fabianfrz:


src/www/services_dhcp.php
```diff
@@ -890,7 +890,7 @@ function show_netboot_config() {
 										<td width="22%" valign="top" class="vncell"><?=gettext("Failover peer IP:");?></td>
 										<td width="78%" class="vtable">
 											<input name="failover_peerip" type="text" class="form-control host" id="failover_peerip" size="20" value="<?=htmlspecialchars($pconfig['failover_peerip']);?>" /><br />
-											<?=gettext("Leave blank to disable.  Enter the interface IP address of the other machine.  Machines must be using CARP. Interface's advskew determines whether the DHCPd process is Primary or Secondary. Ensure one machine's advskew<20 (and the other is >20).");?>
+											<?=htmlspecialchars(gettext("Leave blank to disable.  Enter the interface IP address of the other machine.  Machines must be using CARP. Interface's advskew determines whether the DHCPd process is Primary or Secondary. Ensure one machine's advskew<20 (and the other is >20)."));?>
 										</td>
```
AdSchellevis: Maybe we should look at the use of "htmlspecialchars(gettext(" in a bit wider context for 16.7, almost all occurrences of gettext should be escaped because we don't know the end result.... maybe there's an easy way to wrap this for all frontend outputs.

fabianfrz: So I will remove this function call now

fichtner: what's worse is that htmlspecialchars() is not enough, I've added html_safe() to deal with encoding of single quotes which can break JavaScript code, see French translation and https://github.com/opnsense/core/commit/1e5fb3e35a77f0e0de6cadf45ef90afdb6aad95d

fichtner:
Merged, thanks @fabianfrz

PR #717, opened 2016-01-28
"device polling" changes also require a reboot

oparoz:
The changes won't happen until the interface is re-configured, just like for the other settings.


AdSchellevis:
Thanks for this addition!

PR #729, opened 2016-01-31
Firewall log not dispalaying intentionally logged IGMP packets

dotike:
Intentional IGMP log messages dropped because the packets did not have source and destination set.


fichtner:
Looks good, thanks @dotike  :)

Issue #707, opened 2016-01-25
Broken hardware crypto info on dashboard


oparoz:
I get something like this:

`Hardware crypto	<AES-CBC`

oparoz:
Also, that row doesn't always appear on the dashboard.

gonzopancho:
I've created a pull request https://github.com/opnsense/core/pull/732 that should close this issue.  (We fixed it months ago in pfSense.)

I predict that @fichtner won't take it, and someone else will clone the work.  You'll still get a fix that way.

oparoz:
>The code in question was removed after a bit of tinkering with the direction to take.

Care to briefly elaborate? I think it's quite useful to have some info on the dashboard about the hardware and available features. Maybe in a separate widget?

fichtner:
The output this code parsed is enclosed in dmesg (visible via crash reporter) or available in the system logs. The real setting under System: Settings: Misc: Cryptographic Hardware Acceleration is authoritative and works fine.

PR #732, opened 2016-02-01
Make the display of hardware crypto information more reliable for AESNI

gonzopancho:
From pfSense https://github.com/pfsense/pfsense/commit/a2e6d7f82f4811c743542837bf56c4336a6711eb / issue #4911 

Should close opnsense issue #707


gonzopancho:
In case there is any question, you can take this code under a BSD 2-Clause license.

gonzopancho:
That's another way to fix it, I suppose.

fichtner:
The code in question was removed after a bit of tinkering with the direction to take. Thank you for your contribution.

PR #781, opened 2016-02-18
fix filter reload with invalid relayd config, fixes #780

### LINKED ISSUE #780: relayd: applying configuration changes disrupts firewall traffic
When applying Loadbalancer settings, most network traffic stops working and seems to be filtered.

It can easily be fixed by manually reloading the filter rules (Firewall->Diagnostics->Filter Reload), but there seems to be something wrong around the reload of the Loadbalancer.

I've seen this already on 15.7.x and can confirm that this is still an issue on 16.1.3.

fraenki:
fix for #780

### DIFF
--- a/src/www/load_balancer_monitor.php
+++ b/src/www/load_balancer_monitor.php
@@ -44,8 +44,8 @@ if ($_POST) {
 
 	if ($_POST['apply']) {
 		$retval = 0;
-		$retval |= filter_configure();
 		$retval |= relayd_configure();
+		$retval |= filter_configure();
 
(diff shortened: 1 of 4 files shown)


fichtner:
@fraenki nice, thanks. when it's running the reload is ok too with this?

fraenki:
@fichtner Yes, I've created and delete some _Pools_ and _Virtual Servers_ and experienced no more network disruption. The configuration changes were applied as expected.

fichtner:
Thanks for this then, great stuff. :)

PR #782, opened 2016-02-18
fix relayd startup/restart (ref #780)

fraenki:
@fichtner In contrast to my original report in #780, a manual start/restart of the relayd service will disrupt network traffic too. Maybe this is a very specific edge case in my OPNsense setup/configuration, but adding a filter_configure() fixes this too.


fichtner:
LGTM, thanks again :)

PR #786, opened 2016-02-20
translation

fabianfrz:


src/www/services_dhcp.php
```diff
@@ -990,17 +990,15 @@ function removeRow() {
...
-                          <?=gettext("and the UEFI 32bit filename  ");?>
-                          <input name="filename32" type="text" id="filename32" value="<?=$pconfig['filename32'];?>" /><br />
-                          <?=gettext("and the UEFI 64bit filename  ");?>
-                          <input name="filename64" type="text" id="filename64" value="<?=$pconfig['filename64'];?>" /><br />
+                          <?=sprintf(gettext("Enter the IP of the next-server %s and the default bios filename %s and the UEFI 32bit filename %s and the UEFI 64bit filename %s"),
+                                    '<input name="nextserver" type="text" id="nextserver" value="' . $pconfig['nextserver'] . '" /><br />',
+                                    '<input name="filename" type="text" id="filename" value="' . $pconfig['filename'] . '" /><br />',
+                                    '<input name="filename32" type="text" id="filename32" value="' . $pconfig['filename32'] . '" /><br />',
+                                    '<input name="filename64" type="text" id="filename64" value="' . $pconfig['filename64'] . '" /><br />'
+                                    ) ?>
```
AdSchellevis: @fabianfrz I would rather leave the input tags out of the parsed string data, but that's probably just a matter of style preference. 
Maybe @fichtner disagrees with me here, so I will leave this pull request for him :)
As always, thanks for you help!

fabianfrz: @AdSchellevis it is required here because the input tags are between the parts of the sentence.

fichtner:
Did not merge this, needed to break up the individual inputs instead. Thanks for noticing, @fabianfrz  :)

PR #789, opened 2016-02-23
Add NetMaskRequired option into NetworkField and NetworkValidator

evbevz:
If I need to ensure that user enters the ip with netmask into network field, I can use `<NetMaskRequired>Y</NetMaskRequired>` in my model.


AdSchellevis:
Thank you for the addition.
I will pull this one in and pretty it up a bit, can you please make sure that your code complies to PSR-2 the next time?

AdSchellevis:
@evbevz  I changed it a bit to comply with the code style and only except "Y"/"y" like the other types.
Changes are in https://github.com/opnsense/core/commit/7c4bc839679fb860e4ee22ba027d1f50be05ffaf

PR #792, opened 2016-02-24
Added masquerade logon controller

zvs44:
Added a means to masquerade (dynamic logon) as a user to the Captive Portal as discussed on the forum: https://forum.opnsense.org/index.php?topic=2028.msg6402

### DIFF
--- a/src/opnsense/mvc/app/controllers/OPNsense/CaptivePortal/Api/MasqueradeController.php
+++ b/src/opnsense/mvc/app/controllers/OPNsense/CaptivePortal/Api/MasqueradeController.php
@@ -1,0 +1,174 @@
+<?php
+/**
+ *    Copyright (C) 2015 Deciso B.V.
+ *
+ *    All rights reserved.
+ *
+ *    Redistribution and use in source and binary forms, with or without
+ *    modification, are permitted provided that the following conditions are met:
+ *
+ *    1. Redistributions of source code must retain the above copyright notice,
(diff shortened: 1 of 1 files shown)


AdSchellevis:
@zvs44 I looked briefly at your code, I like the idea of authenticating an arbitrary client ip using the API, but rather not see a copy of the default access controller.
Is the authentication feature the only thing your looking for here? (maybe I'm missing something) If that is the case we could easily extend the SessionController with a connectAction (the disconnect is already there).

src/opnsense/mvc/app/controllers/OPNsense/CaptivePortal/Api/MasqueradeController.php
```diff
@@ -0,0 +1,174 @@
...
+        // get username, IP and auth server from post
+        $userName = $this->request->getPost("user", "striptags", null);
+        $clientIp = $this->request->getPost("ip", "striptags", null);
+        $authServer = $this->request->getPost("server", "striptags", null);
+        $timeout = $this->request->getPost("timeout", "striptags", null);
+        
+        // If an explicit client IP wasn't provided
+        if(!$clientIp)
+        {
+            $clientIp = $this->getClientIp();
```
fabianfrz: If you want to allow a client then this line is wrong - you should return an error instead.

AdSchellevis:
time-out, no feedback

PR #794, opened 2016-02-25
Unbound dhcpd

AdSchellevis:
Refactor Unbound + dhcp leases, removes the /etc/hosts parsing and add a watcher daemon. 
Shareable components have been placed in src/site-python/, which may be used by the other scripts as well (todo).

### DIFF
--- a/src/etc/inc/system.inc
+++ b/src/etc/inc/system.inc
@@ -458,8 +458,7 @@ function system_dhcpleases_configure()
     global $config, $g;
 
     /* Start the monitoring process for dynamic dhcpclients. */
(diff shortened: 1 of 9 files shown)


fichtner:
Leases, dynamic and static, land in their respective files now. The work looks good, however, on the current master I seem to be unable to resolve from my internal clients with unbound:

````
This webpage is not available

DNS_PROBE_FINISHED_BAD_CONFIG
````

Related note: I noticed that static dhcp for dnsmasq doesn't seem to work.

fichtner:
# sudo killall -HUP mDNSResponder

Fixed this according to @AdSchellevis , thanks for the work!!

PR #796, opened 2016-02-26
Icap support

fabianfrz:
i think it works now but it needs some additional testing


AdSchellevis:
@fabianfrz I will pull this in now and test / optimize later today.

AdSchellevis:
@fabian I tested it on my end it all seems to be working fine, only issue I had was with c-icap, the ports version doesn't seem to be working properly for testing).
Thanks for your contribution!

fabianfrz:
@AdSchellevis thanks for testing.
We will see what people will think about this feature.

Have a nice day,

Fabian

PR #816, opened 2016-03-13
Fix 792

fabianfrz:
Fixed the issues with pull request #792.

@zvs44 I think your copyright should be added with the full name but because I have only your GitHub username I cannot add your real name. Just let us know how you would like to be attributed.

This closes #792


AdSchellevis:
@fabianfrz I'm still waiting for a response from @zvs44 on the original pull request, from an architecture point of view it seems more logical to add the login feature to the already existing SessionController. (the login function itself is only a couple of lines of code and the  SessionController already has functions in the same area).

fabianfrz:
@AdSchellevis please check again

AdSchellevis:
@fabianfrz I totally missed the file name here, sorry about that. I will review this code later, but would rather see feedback first before we start fixing things. A general connect option is quite easy to add with only a few lines of code, I'm just not convinced all of the original connect option applies to the use-case (first thing that comes to mind are the client restrictions).

Let's discuss this on irc later this week.

AdSchellevis:
time-out, no feedback

PR #844, opened 2016-03-21
Added ability to change the hostname and admin email in Proxy

thelinuxfr:
Added ability to change the hostname and admin email in Proxy

_(Small addition for me to test if things are well done?)_

### DIFF
--- a/src/opnsense/mvc/app/models/OPNsense/Proxy/Proxy.xml
+++ b/src/opnsense/mvc/app/models/OPNsense/Proxy/Proxy.xml
@@ -69,6 +69,14 @@
                 <default>0</default>
                 <Required>N</Required>
             </suppressVersion>
+            <VisibleEmail type="TextField">
+                <default>localhost</default>
+                <Required>N</Required>
+            </VisibleEmail>
+            <VisibleHostname type="TextField">
+                <default>admin@localhost</default>
+                <Required>N</Required>
+            </VisibleHostname>
             <cache>
                 <local>
                     <enabled type="BooleanField">
--- a/src/opnsense/mvc/app/models/OPNsense/Proxy/Proxy.xml
+++ b/src/opnsense/mvc/app/models/OPNsense/Proxy/Proxy.xml
@@ -71,14 +71,14 @@
             </suppressVersion>
             <VisibleEmail type="EmailField">
                 <default>admin@localhost</default>
-                <ValidationMessage>Please enter a valid email address</ValidationMessage>
                 <Required>N</Required>
+                <ValidationMessage>Please enter a valid email address.</ValidationMessage>
             </VisibleEmail>
             <VisibleHostname type="TextField">
                 <default>localhost</default>
                 <Required>N</Required>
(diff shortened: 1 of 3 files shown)


fichtner:
Yay :)

Since none of those are required, each template line needs "if helpers.exists ..." wrapping. They should not land in the template when empty (and produce a warning/error).

src/opnsense/mvc/app/models/OPNsense/Proxy/Proxy.xml
```diff
@@ -69,6 +69,14 @@
                 <default>0</default>
                 <Required>N</Required>
             </suppressVersion>
+            <VisibleEmail type="TextField">
```
AdSchellevis: You probably best change this to EmailField type

src/opnsense/mvc/app/models/OPNsense/Proxy/Proxy.xml
```diff
@@ -69,6 +69,14 @@
...
+                <default>localhost</default>
```
AdSchellevis: I guess hostname and email settings are just the other way around for the defaults :)

src/opnsense/mvc/app/models/OPNsense/Proxy/Proxy.xml
```diff
@@ -69,6 +69,14 @@
...
+                <Required>N</Required>
+            </VisibleEmail>
+            <VisibleHostname type="TextField">
+                <default>admin@localhost</default>
+                <Required>N</Required>
```
AdSchellevis: Maybe add a regexp match here to match only valid hostnames (we don't have a hostname type yet)

AdSchellevis:
Nice! I added some comments to your code, mainly for data validation reasons.

src/opnsense/mvc/app/models/OPNsense/Proxy/Proxy.xml
```diff
@@ -69,6 +69,17 @@
...
             </suppressVersion>
+            <VisibleEmail type="EmailField">
+                <default>admin@localhost</default>
+                <Required>N</Required>
+                <ValidationMessage>Please enter a valid email address.</ValidationMessage>
+            </VisibleEmail>
+            <VisibleHostname type="TextField">
+                <default>localhost</default>
+                <Required>N</Required>
+                <mask>/^([0-9a-zA-Z.,_\-]){0,1024}$/u</mask>
```
fabianfrz: the dot needs a backslash

fichtner: Some more... IPv6 needs ":" and the escape for "-" is superfluous.

thelinuxfr: @fichtner What do you mean for "-" ?

fichtner: I think that "-" only needs to be written as "\-" when it would imply a sequence, but you don't want that ("[0-9]" vs. "[0\-9]" or rather "[09-]"). It's just cosmetic. :)

Issue #848, opened 2016-03-22
Bug: Pulldown menus - Last entry not shown completely


b2000699:
OPNsense 16.1.7-amd64 & Firefox 45.0.1:

When pulling down pull-down-menus with a lot of entries, such as /Firewall/Rules/Edit rule/Protocol, the last entry at the bottom of the pull-down-menu is only shown half, i.e. it is cut off horizontally approximately at it's middle line.
Another example would be "Destination" in the same menu.

Cheers

thelinuxfr:
Same issue on OPNsense 16.7.a_648-0c6e70a3c [16.7.a_456-01b168c51]

![issue-848](https://cloud.githubusercontent.com/assets/43656/13967940/0ce02802-f07a-11e5-8bbe-2236c14b9571.png)

fichtner:
The current fix forces reversed artefacts such as the following:

<img width="121" alt="screen shot 2016-03-25 at 5 35 49 pm" src="https://cloud.githubusercontent.com/assets/1915288/14048730/52d61aaa-f2b0-11e5-8e42-467413f704df.png">

I've said this before on the PR: having the last item "cut off" looks a bit weird but is in some UIs used to indicate that further entries are available. I am unsure whether it should be fixed or kept as is.

fichtner:
some attempts have been made, but this won't fix easily. for better or for worse assume this working as intended.

PR #852, opened 2016-03-23
Proposal fix issue #848

thelinuxfr:
Proposal fix issue #848

### DIFF
--- a/src/opnsense/www/themes/opnsense/assets/stylesheets/bootstrap/_dropdowns.scss
+++ b/src/opnsense/www/themes/opnsense/assets/stylesheets/bootstrap/_dropdowns.scss
@@ -34,7 +34,7 @@
   display: none; // none by default, but block on "open" of the menu
   float: left;
   min-width: 160px;
-  padding: 5px 0;
+  padding: 5px 0 15px 0;
   margin: 2px 0 0; // override default ul
   list-style: none;
   font-size: $font-size-base;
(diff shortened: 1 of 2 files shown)


AdSchellevis:
@thelinuxfr the idea is good, but it should be fixed in the sass file (main.scss) with an override and then be compiled into the css file again.
This css stuff is a bit magic from time to time, the idea of the bootstrap components is that you use the standards and extend them where needed, but I can't guarantee that none of the files are changed at the moment. (but I would try to minimize changes to standards where possible)
Need help with creating the override?

thelinuxfr:
@AdSchellevis I modified the scss, I will help in the following

AdSchellevis:
@thelinuxfr I checked the issue again with your fix, your fix functions perfectly, but should only apply to this type of dropdown box.
What about his fix https://github.com/opnsense/core/commit/b32d93da219b025387ca005b02fcb1e5c23514fb ?

fichtner:
FWIW, I've seen such intentional behaviour on iOS to indicate that there is more data in the box as opposed to rendering it "correctly" where it's not clear that there is more entries after the last one. I do not think this has to be a bug.

AdSchellevis:
I don't have a strong opinion on this one, both work for me, I'm happy to revert the change if  the slightly cut-of version seems more obvious for people.

thelinuxfr:
I test tomorrow, but your solution seems correct , I would close my PR if it's ok

thelinuxfr:
@AdSchellevis Fix your suits me, it is true that this introduces another problem when there are only two elements ... I closed my PR

AdSchellevis:
@thelinuxfr the fix might not make it into the release because of other issues, see https://github.com/opnsense/core/issues/848

PR #861, opened 2016-03-26
(legacy) description and consistency cleanup in firewall_virtual_ip_edit.php

azdps:


AdSchellevis:
@azdps Thanks!

azdps:
No problem. Also noticed in the drop down box for Edit Virtual IP that CARP is all in lower case. I figured CARP should be capitalized since it's an acronym. Was't sure if case sensitive was a factor so I left it alone because I didn't want to inadvertently introduce a bug.

AdSchellevis:
The gettext() string may change to upper case, no problem, as long as the actual field value doesn't change. Your absolutely right, it looks better when its in upper case.

src/www/firewall_virtual_ip_edit.php
```diff
@@ -476,7 +476,7 @@ function find_last_used_vhid() {
               <span class="text-danger">
                 <strong><?=gettext("Note:");?><br /></strong>
               </span>
-              <?=gettext("Proxy ARP and Other type Virtual IPs cannot be bound to by anything running on the firewall, such as IPsec, OpenVPN, etc.  Use a CARP or IP Alias type address for these cases.");?>
+              <?=gettext("Proxy ARP and other type Virtual IPs cannot be bound to by anything running on the firewall, such as IPsec, OpenVPN, etc.  Use a CARP or IP Alias type address for these cases.");?>
               <br /><br />
               <?=gettext("For more information on CARP and the above values, visit the OpenBSD ");?><a href='http://www.openbsd.org/faq/pf/carp.html'> <?=gettext("CARP FAQ"); ?></a>.
```
fabianfrz: This line needs a changes too.

azdps:
fabianfrz are you referring to "Note:"? If this the case "Note:" is consistent across most of the pages that I can see. It also seems to be appropriate for it's use.

fichtner:
I think he means the embedding of the html tags using %s so that the gettext string is one single translation line: 'Texttext text: %sCARP FAQ%s.'

fichtner:
Blazingly fast... https://github.com/opnsense/core/pull/862

PR #862, opened 2016-03-26
translation

fabianfrz:


src/www/firewall_virtual_ip_edit.php
```diff
@@ -476,9 +476,9 @@ function find_last_used_vhid() {
               <span class="text-danger">
                 <strong><?=gettext("Note:");?><br /></strong>
               </span>
-              <?=gettext("Proxy ARP and other type Virtual IPs cannot be bound to by anything running on the firewall, such as IPsec, OpenVPN, etc.  Use a CARP or IP Alias type address for these cases.");?>
+              <?=gettext("Proxy ARP and other type Virtual IPs cannot be bound to by anything running on the firewall, such as IPsec, OpenVPN, etc. Use a CARP or IP Alias type address for these cases.");?>
```
fichtner: In the future, please don't invalidate translations by removing spurious spaces. It takes more time to get the translation back than it's worth having an invisible space removed. :)

fichtner:
Merged, thanks!

PR #866, opened 2016-03-29
add new form input type "header"

fraenki:
If I'm not mistaken there is currently no (easy) way to add a proper header for forms. This makes long forms rather confusing:

forms_info

This patch adds the new input type "header" to allow headers in forms:

```
    <field>
        <label>Tuning Options</label>
        <type>header</type>
    </field>
```

forms_header

(I'd be more happy to have a margin of just 10px instead of 15px, but I don't wanted to add a new css class just for this.)

### DIFF
--- a/src/opnsense/mvc/app/views/layout_partials/form_input_tr.volt
+++ b/src/opnsense/mvc/app/views/layout_partials/form_input_tr.volt
@@ -46,9 +46,13 @@ allownew    :   allow new items (for list) if applicable
             {% if help|default(false) %}
                 <a id="help_for_{{ id }}" href="#" class="showhelp"><i class="fa fa-info-circle"></i></a>
             {% elseif help|default(false) == false %}
-                <i class="fa fa-info-circle text-muted"></i>
+                <i class="fa fa-info-circle text-muted{% if type == "header" %} hidden{% endif %}"></i>
             {% endif %}
-            <b>{{label}}</b>
+            {% if type == "header" %}
+                <h4 class="modal-title __mb __mt">{{label}}</h4>
+            {% else %}
+                <b>{{label}}</b>
+            {% endif %}
         </div>
     </td>
     <td >


AdSchellevis:
@fraenki you can use tabs to split content, I'm not fully sure about the headers in this case. If you don't use the tabs and want to divide the content within the same screen, a separate partial which renders the same structure in to different blocks might be better. (better spacing on the page)

fraenki:
@AdSchellevis Thanks for your response. :) I've decided to add this new type, because it's also the font size (and possibly font style) that matters in this case. To me this seemed like an easy way to achieve the goal without major drawbacks. Am I missing something?

> you can use tabs to split content

I'm already using tabs for general application settings. But this form is used to create/edit new frontends for a Load Balancer – there can be many frontends. I don't see how I could use tabs in this case...

> If you don't use the tabs and want to divide the content within the same screen, a separate partial which renders the same structure in to different blocks might be better.

Is there an example in the code somewhere that demonstrates this?
Still, this sounds like a lot work to simply make the form look more pretty. :(

AdSchellevis:
@fraenki I just don't like the broken striping very much, the idea is good. Do you mind if I move your idea into "base_form" so we can fix the tables a bit better?

fichtner:
I kind of agree with Ad here, two things are important:

1. *every* section needs a caption/description, not just the added ones
2. the sections are the stacked tabs, but...

As far as 2. goes I always felt that the stacked tabs aren't very intuitive and they could very well be rendered together on one page, which is what Frank is looking for. Maybe we could use a form render switch for this in the xml?

fichtner:
Note that "tabs" and "stacked tabs" are different things, the proxy pages make use of stacked tabs...

AdSchellevis:
@fichtner I guess the additional type from @fraenki should be fine here (only fix the presentation), the tabs can be made more intuitive at a later point in time.

fraenki:
@AdSchellevis No objections from my side :)

AdSchellevis:
@fraenki is this one https://github.com/opnsense/core/commit/95522e0935b3d805fd2f3b741437da458de6c21f good enough for you? (drop in replacement for this pull)

fraenki:
Looks good. Thanks, @AdSchellevis!

AdSchellevis:
@fraenki your welcome!

PR #869, opened 2016-03-30
support multiple values in ModelRelationField, fixes #868

fraenki:
proposal for #868

Issue #868, opened 2016-03-30
[framework] support n-m relation in ModelRelationField


fraenki:
Currently it is limited to 1-n relation, but I'd require n-m relation. :)
As discussed with @AdSchellevis on IRC...

AdSchellevis:
details https://github.com/opnsense/core/pull/869#issuecomment-203821570

fraenki:
@AdSchellevis I was wrong in https://github.com/opnsense/core/pull/869#issuecomment-203821570. My test failed because I forgot to add the tag `Multiple>Y</Multiple>`. D'oh! I did a new test and now it's working as expected.

AdSchellevis:
@fraenki thanks for testing again! this indeed explains all :)

PR #876, opened 2016-04-04
support multiple values in CertificateField

fraenki:
Basically the same as #869, but this time for CertificateField. This functionality is required for SNI-capable services/plugins and thus support multiple certs at once.


AdSchellevis:
@fraenki just to be sure, doesn't this have the same issue as https://github.com/opnsense/core/issues/868 ? (accepts multiple inputs when only one may be accepted)

fraenki:
@AdSchellevis You're right, I've fixed this with a second commit.

AdSchellevis:
@fraenki Thank you!

PR #878, opened 2016-04-06
Simple names to complete drop-down

dotike:
tiny translation addition


fichtner:
Thanks, I've added these two to Pootle as well.

PR #887, opened 2016-04-11
enable jinja2 loopcontrols extension

fraenki:
Please :) This enables the break/continue loop controls in jinja. I'll need it for a plugin...


AdSchellevis:
that shouldn't be a problem :)

PR #894, opened 2016-04-15
translation

fabianfrz:
was not able to test this (I do not have a ppp interface) -> requires additional check


fichtner:
Merged, thanks!

PS: you can always use "make lint" to run simple syntax checks on the repo before creating a PR.

fabianfrz:
I did a syntax check but it is also important to me to check if everything is still correctly displayed.

fichtner:
Looked good from here. :)

PR #1044, opened 2016-06-27
fix for the boot issue in openvpn_get_engines(); closes #1038

### LINKED ISSUE #1038: Issue on Boot
It looks like an openssl command fails on boot:

Error while Booting

running:
OPNsense 16.7.b_394-amd64
FreeBSD 10.2-RELEASE-p18
LibreSSL 2.3.5

fabianfrz:


src/etc/inc/openvpn.inc
```diff
@@ -193,6 +193,13 @@ function openvpn_get_digestlist()
...
 {
     $openssl_engines = array('none' => 'No Hardware Crypto Acceleration');
+    exec('/usr/local/bin/openssl version', $openssl_version);
+    $is_libressl = explode(' ', $openssl_version[0])[0] == 'LibreSSL';
+    if ($is_libressl) {
+        /* LibreSSL does not have the engines command */
+        return $openssl_engines;
+    }
+    
     exec('/usr/local/bin/openssl engine -t -c', $openssl_engine_output);
```
fichtner: Thanks for pinning this down :)

Instead of checking the version, stderr should be muted instead. Although I'm unsure if openvpn_get_engines() is even needed for bootup?!

fichtner:
Thanks, I'll cherry-pick this tomorrow. The command is used to verify the engine before config write. It seems ok as it can depend on the kernel module being loaded and may prevent OpenVPN startup.

fichtner:
Merged manually, thanks!

PR #1058, opened 2016-07-10
translation

fabianfrz:


src/www/interfaces_gre_edit.php
```diff
@@ -183,12 +183,12 @@
                     <td>
                       <input name="remote-addr" type="text" value="<?=$pconfig['remote-addr'];?>" />
                       <div class="hidden" for="help_for_remote-addr">
-                        <?=gettext("Peer address where encapsulated GRE packets will be sent ");?>
+                        <?=gettext("Peer address where encapsulated GRE packets will be sent");?>
```
fichtner: needs "."

src/etc/inc/upgrade_config.inc
```diff
@@ -737,7 +737,7 @@ function upgrade_040_to_041() {
 		$config['sysctl']['item'][15]['value'] =   "default";
 
 		$config['sysctl']['item'][16]['tunable'] = "net.inet.tcp.inflight.enable";
-		$config['sysctl']['item'][16]['descr'] =    gettext("The system will attempt to calculate the bandwidth delay product for each connection and limit the amount of data queued to the network to just the amount required to maintain optimum throughput. ");
+		$config['sysctl']['item'][16]['descr'] =    gettext("The system will attempt to calculate the bandwidth delay product for each connection and limit the amount of data queued to the network to just the amount required to maintain optimum throughput.");
```
fichtner: I don't like any of these in this file: very very old compatibility layer with m0n0wall creating unnecessary work for translators. If anything, we should strip all gettext() here.

fichtner:
Merged, thanks!

PR #1060, opened 2016-07-12
some fixes for vpn_openvpn_export.php

fabianfrz:


src/www/vpn_openvpn_export.php
```diff
@@ -490,9 +490,9 @@
                         <?=gettext("Use a password to protect the pkcs12 file contents or key in Viscosity bundle.");?>
                       </div>
                       <div id="usepass_opts" style="display:none">
-                        <label ><?=gettext("Password :");?></label>
+                        <label for="pass"><?=gettext("Password:");?></label>
                         <input name="pass" id="pass" class="form-control" type="password" value="" />
-                        <label ><?=gettext("Confirm :");?></label>
+                        <label for="conf"><?=gettext("Confirm:");?></label>
```
fichtner: I was going to suggest moving the colon outside of the issue. "Password" is readily translated, but "Confirm" is tricky. We should opt for "Confirmation".

src/www/vpn_openvpn_export.php
```diff
@@ -686,7 +686,7 @@
                     <div class="hidden" for="help_for_clientpkg">
                       <br/>
                       <?= gettext("NOTES:") ?> <br/>
-                      <?= gettext("The &quot;XP&quot; Windows installers work on Windows XP and later versions. The &quot;win6&quot; Windows installers include a new tap-windows6 driver that works only on Windows Vista and later.") ?><br/>
+                      <?= gettext('The "XP" Windows installers work on Windows XP and later versions. The "win6" Windows installers include a new tap-windows6 driver that works only on Windows Vista and later.') ?><br/>
```
fichtner: nice catch, html escaping isn't favourable for translation strings :)

src/www/vpn_openvpn_export.php
fabianfrz: or "Confirm Password:"

fichtner:
Merged, thanks!

PR #1065, opened 2016-07-21
Fix undefined calculate_ipv6_delegation_length()

devste:
Error message "Fatal error: Call to undefined function calculate_ipv6_delegation_length() in /usr/local/etc/inc/services.inc on line 1060"
in
Services -> DHCPv6 -> Advertisements
Services -> DHCPv6 -> Relay

### DIFF
--- a/src/etc/inc/services.inc
+++ b/src/etc/inc/services.inc
@@ -32,6 +32,7 @@
 require_once('dyndns.class');
 require_once('unbound.inc');
 require_once('miniupnpd.inc');
+require_once('pfsense-utils.inc');
 
 function generate_ipv6_from_mac($mac)
 {


fichtner:
Hi Stephen,

Thanks for your fix, but master does not have "pfsense-utils.inc" anymore. I'm guessing this is about the removal in the 16.7-RC series or at least opnsense-devel usage? Can you post your exact version info from the dashboard to confirm?

We also try to register dependencies from the top to avoid frequent dependency loops. Although require_once is built to avoid this, the low level dependencies make refactoring harder than it should be as the include fines end up being "one big single include file".

If you can confirm the version we'll know what to do. :)


Cheers,
Franco

fichtner:
Backporting https://github.com/opnsense/core/commit/21f651c6747eb7773982c46eb8046da266c2b8e9 caused this, sorry. Thanks for the analysis!

devste:
Hi Franco, thanks for fixing it. I did forget to mention that I noticed this after the upgrade from 16.1.18 to 16.1.19.

fichtner:
Just FYI: We're pushing 16.1.20 today instead of Monday to fix this for all. Thanks again!

PR #1066, opened 2016-07-21
Language correction: more generic expression

devste:
Probably home users would call their device "router", while professional users with the stronger systems would call their device "firewall". I opted for a more neutral expression here, "device" sounds too technical in my opinion.


fichtner:
Works for me, thanks!

PR #1080, opened 2016-07-28
Fix IPv6 Outbound NAT

djGrrr:
Rules using "Interface Address" as the NAT target IP did not select the
IPv6 address when generating the rule


djGrrr:
This patch brings in full IPv6 outbound NAT support, so that it works the same as IPv4 does

AdSchellevis:
@djGrrr looks good, thank you!

PR #1086, opened 2016-07-30
Put custom scrub rules first

djGrrr:
It appears that the custom normalization rules do not function unless "disable interface scrub" is enabled.

This is simply because the "interface scrub" rules appear before the custom rules, and since they match all traffic on each interface, the custom scrub rules are never reached.

This patch puts the custom rules first.

### DIFF
--- a/src/etc/inc/filter.inc
+++ b/src/etc/inc/filter.inc
@@ -570,26 +570,7 @@ function filter_generate_scrubing(&$FilterIflist)
 
     $scrubrules = '';
 
-    /* scrub per interface options */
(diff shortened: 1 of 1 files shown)


fichtner:
This is a good idea, thanks :) I'm leaving this for @AdSchellevis to pick up.

fichtner:
PS: Can you rebase your branch cleanly against master?

fichtner:
PPS: You can rebase the branch and force push, it will "fix" the pull request :)

AdSchellevis:
@djGrrr thanks, it seems that I misinterpreted the evaluation order.

Issue #1095, opened 2016-08-01
[framework] CertificateField parameters ignored when using multiple fields


fraenki:
I've tried to use multiple `CertificateField` input fields with different `Type` parameter:

```
            <sslcertificatetest type="CertificateField">
                <Required>N</Required>
                <Type>cert</Type>
                <ValidationMessage>Please select a valid certificate from the list</ValidationMessage>
            </sslcertificatetest>
            <sslcertificate type="CertificateField">
                <Required>N</Required>
                <Type>ca</Type>
                <ValidationMessage>Please select a valid CA certificate from the list</ValidationMessage>
            </sslcertificate>
```

Expected result:
- input field `sslcertificatetest` should contain a list of _all_ certificates
- input field `sslcertificate` should contain a list of all *CA* certificates

Actual result:
- both input fields contain a list of all certificates

Note that when using _only_ the `ca` input field and no `cert` input field anywhere it will contain the expected list of CA certificates (see the Proxy module/plugin for an example).

fraenki:
Tested, works as expected. Thanks, @AdSchellevis! :+1:

fichtner:
I'm queueing this up for 16.7.2 along with the HAProxy 1.4 bump. Cheers!

PR #1096, opened 2016-08-01
allow to use CRLs with CertificateField

fraenki:
Currently the `CertificateField` is limited to types `ca` and `cert`. This patch allows to use type `crl` too.
Use case: advanced certificate validation in HAProxy plugin.

Is related to #1095 and not useful (for my use case) until the referenced bug is fixed.


AdSchellevis:
@fraenki I'll pull this one in first and will assign your issue to myself.

PR #1119, opened 2016-08-05
Incorrect units capitalization in proxy form

pv2b:
k = kilo (1000 or 1024)
B = byte
b = bit
K = nothing

So kilobyte is kB (was written as Kb in the form)
And kilobit per second is kbps not Kbps.


fichtner:
@pv2b looks good. Can you, however, rebase and force push your branch to get rid of the "merge pull request" commit?

Please also note that whenever strings are touched, it degrades the translations since they can't be resolved anymore (there is no fuzzy matching). So such changes are welcome as long as they are small (like this one, so no worries at all).

pv2b:
Done!

fichtner:
Merged, thank you! :)

Issue #1123, opened 2016-08-07
Certain common API method implementations ought to be in base class?


pv2b:
Hi!

I'm seeing a lot of boiler plate when implementing API controllers, such as a lot of logic in the getAction() and setAction() methods. This may be abstractable into a common base class or perhaps as traits.

Here's some example of boiler plate code that has to be repeated a lot (from the https://docs.opnsense.org/development/examples/helloworld.html example):

```
public function getAction()
{
    // define list of configurable settings
    $result = array();
    if ($this->request->isGet()) {
        $mdlHelloWorld = new HelloWorld();
        $result['helloworld'] = $mdlHelloWorld->getNodes();
    }
    return $result;
}
public function setAction()
{
    $result = array("result"=>"failed");
    if ($this->request->isPost()) {
        // load model and update with provided data
        $mdlHelloWorld = new HelloWorld();
        $mdlHelloWorld->setNodes($this->request->getPost("helloworld"));

        // perform validation
        $valMsgs = $mdlHelloWorld->performValidation();
        foreach ($valMsgs as $field => $msg) {
            if (!array_key_exists("validations", $result)) {
                $result["validations"] = array();
            }
            $result["validations"]["general.".$msg->getField()] = $msg->getMessage();
        }

        // serialize model to config and save
        if ($valMsgs->count() == 0) {
            $mdlHelloWorld->serializeToConfig();
            Config::getInstance()->save();
            $result["result"] = "saved";
        }
    }
    return $result;
}
```

Things that have to be done to make this general:
1. The name "helloworld" that's used in getAction() to define where the result goes.
2. Code to instantiate a new model action.
3. Potential hooks to allow the programmer to easilly influence the behavior of the method.

I'm going to have a bash at this and see what I come up with.

PR #1124, opened 2016-08-07
Extract ApiModelControllerBase and ApiMutableModelControllerBase classes

pv2b:
As per issue https://github.com/opnsense/core/issues/1123 I hereby submit a proposed refactoring.

In order to support the very common case of a controller that that implements getters and setters on a model, I've extracted out two classes, ApiModelControllerBase that can be inherited by read-only API controllers, and ApiMutableModelControllerBase that can be inherited by read-write API controllers.

I have also refactored the IDS and traffic shaper using these two classes. The set code path probably needs review, especially the validation may be a little bit iffy. I also think there's some room for further simplification, specifically in whether OPNsense\Base\ApiModelControllerBase\getModelNodes is correct in the general case. That, or if there needs to be an extention to the model to mark in the metadata if a field should be exposed in the API or not.

I have done basic validation of the get methods, ensuring that I get the same result before and after refactoring, but I have not done testing of the set methods or validation.

### DIFF
--- a/src/opnsense/mvc/app/controllers/OPNsense/Base/ApiModelControllerBase.php
+++ b/src/opnsense/mvc/app/controllers/OPNsense/Base/ApiModelControllerBase.php
@@ -1,0 +1,61 @@
+<?php
+/**
+ *    Copyright (C) 2016 IT-assistans Sverige AB
+ *
+ *    All rights reserved.
+ *
+ *    Redistribution and use in source and binary forms, with or without
+ *    modification, are permitted provided that the following conditions are met:
+ *
+ *    1. Redistributions of source code must retain the above copyright notice,
+ *       this list of conditions and the following disclaimer.
+ *
+ *    2. Redistributions in binary form must reproduce the above copyright
+ *       notice, this list of conditions and the following disclaimer in the
+ *       documentation and/or other materials provided with the distribution.
+ *
+ *    THIS SOFTWARE IS PROVIDED ``AS IS'' AND ANY EXPRESS OR IMPLIED WARRANTIES,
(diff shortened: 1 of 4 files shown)


pv2b:
getModelName() may also be unneccesary if that can be pulled out of the model metadata.

AdSchellevis:
@pv2b I also think its better to move getModelName to the model metadata. The idea to add simple model classes is good, I'll cherry pick your changes in step by step and refactor where needed.

AdSchellevis:
@pv2b just did some refactoring on your code in https://github.com/opnsense/core/commit/7a67a2c03d07559c929d817aea28b7abd1edf8d0 

Let me briefly explain the things I've changed:

- move the internal model name (internalModelName) to a property, we may even remove that tag eventually or put a default in there... its only used to set/get the start of the in/out structure from the api point of view
- define class name and namespace for your model in $this->internalModelClass
- move getModel() to the model base, the default behaviour is mostly the same, you can extend if needed.
- validate properties on initialize(), your abstract method forces validation which I couldn't do for the properties, so hence the property validation
- getModelNodes() probably better to use the model already available on $this item
- and some typo's, nothing special

I will refactor the proxy code in the next commit.

pv2b:
Okay, I like this improved version, but as we discussed on IRC probably it's better to make internalModelClass and internalModelName into statics. I wanted to make them static methods in my version, but in PHP you can't make abstract static functions for some reason. protected static member should not be a problem though.

I think it would be better to put a default in for internalModelName instead, or perhaps just automatically grab it from the class name somehow... or it might as I said be better if that were actually defined as part of the model itself.

pv2b:
Closing this PR to work on #1134 now.

PR #1125, opened 2016-08-08
Filter URL Only at HTTPS Filter

fabiomello:
Some users just want to block https urls by transparent https proxy


AdSchellevis:
@fabiomello thanks for the contribution, it looks good, I've pulled it in and will do some additional testing now

PR #1126, opened 2016-08-08
ApiMutableTableModelControllerBase

pv2b:
As per discussions with @AdSchellevis here's my first stab at an ApiMutableTableModelControllerBase class. The use case for this abstract base class is to have a base class that can be derived from to act as an API controller for a gridview.

I've based this code on the pipe methods in OPNsense\TrafficShaper\SettingsController. A [diff](https://github.com/opnsense/core/files/407603/diff.txt) between these methods and my methods has been attached. This might be useful for understanding what I've done.

A few questions remain before this can be merged:

- What about validation? It seems that the code I looked at does not do validation. I'm not sure if that's correct in the general case.
- I'm not sure if I've implemented addItemAction() correctly.
- In addItemAction(), the traffic shaper code sets an origin node in the model. I'm not sure if this is something that needs to happen in the general case. And if not, do we need to add a hook for it?
- In searchItemsAction(), the fetchBindRequest method is passed a list of column headers. Who should supply this list? Can this be gleaned from metadata rather than hardcoded? And what about the "number" parameter?

### DIFF
--- a/src/opnsense/mvc/app/controllers/OPNsense/Base/ApiMutableModelControllerBase.php
+++ b/src/opnsense/mvc/app/controllers/OPNsense/Base/ApiMutableModelControllerBase.php
@@ -123,7 +123,7 @@ abstract class ApiMutableModelControllerBase extends ApiControllerBase
         foreach ($valMsgs as $field => $msg) {
             // replace absolute path to attribute for relative one at uuid.
             if ($node != null) {
-                $fieldnm = str_replace($node->__reference, static:$internalModelName, $msg->getField());
+                $fieldnm = str_replace($node->__reference, static::$internalModelName, $msg->getField());
                 $result["validations"][$fieldnm] = $msg->getMessage();
             } else {
                 $result["validations"][$msg->getField()] = $msg->getMessage();
--- a/src/opnsense/mvc/app/controllers/OPNsense/Base/ApiMutableModelControllerBase.php
+++ b/src/opnsense/mvc/app/controllers/OPNsense/Base/ApiMutableModelControllerBase.php
@@ -108,6 +108,38 @@ abstract class ApiMutableModelControllerBase extends ApiControllerBase
     }
 
     /**
+     * validate and save model after update or insertion.
+     * Use the reference node and tag to rename validation output for a specific node to a new offset, which makes
+     * it easier to reference specific uuids without having to use them in the frontend descriptions.
+     * @param $node reference node, to use as relative offset
+     * @return array result / validation output
+     */
+    protected function save($node = null)
+    {
+        $mdl = $this->getModel();
+        $result = array("result"=>"failed","validations" => array());
+        // perform validation
+        $valMsgs = $mdl->performValidation();
+        foreach ($valMsgs as $field => $msg) {
+            // replace absolute path to attribute for relative one at uuid.
+            if ($node != null) {
+                $fieldnm = str_replace($node->__reference, static:$internalModelName, $msg->getField());
+                $result["validations"][$fieldnm] = $msg->getMessage();
+            } else {
+                $result["validations"][$msg->getField()] = $msg->getMessage();
+            }
+        }
+        // serialize model to config and save when there are no validation errors
+        if (count($result['validations']) == 0) {
+            // save config if validated correctly
+            $mdl->serializeToConfig();
+            Config::getInstance()->save();
+            $result = array("result" => "saved");
+        }
+        return $result;
+    }
+
+    /**
      * update model settings
      * @return array status / validation errors
      */
@@ -118,22 +150,7 @@ abstract class ApiMutableModelControllerBase extends ApiControllerBase
             // load model and update with provided data
             $mdl = $this->getModel();
             $mdl->setNodes($this->request->getPost(static::$internalModelName));
-
-            // perform validation
-            $valMsgs = $mdl->performValidation();
-            foreach ($valMsgs as $field => $msg) {
-                if (!array_key_exists("validations", $result)) {
-                    $result["validations"] = array();
-                }
-                $result["validations"][static::$internalModelName.".".$msg->getField()] = $msg->getMessage();
-            }
-
-            // serialize model to config and save
-            if ($valMsgs->count() == 0) {
-                $mdl->serializeToConfig();
-                Config::getInstance()->save();
-                $result["result"] = "saved";
-            }
+            $result = $this->save();
         }
         return $result;
     }
--- a/src/opnsense/mvc/app/controllers/OPNsense/Base/ApiMutableModelControllerBase.php
+++ b/src/opnsense/mvc/app/controllers/OPNsense/Base/ApiMutableModelControllerBase.php
@@ -111,12 +111,13 @@ abstract class ApiMutableModelControllerBase extends ApiControllerBase
      * validate and save model after update or insertion.
      * Use the reference node and tag to rename validation output for a specific node to a new offset, which makes
      * it easier to reference specific uuids without having to use them in the frontend descriptions.
+     * @param $mdl The model to save
      * @param $node reference node, to use as relative offset
+     * @param $actionHook A closure containing an action to do after validation but before saving. Must return null if successful, or an error message if not.
      * @return array result / validation output
      */
-    protected function save($node = null)
+    protected function save($mdl, $node = null, $actionHook = null)
     {
-        $mdl = $this->getModel();
         $result = array("result"=>"failed","validations" => array());
         // perform validation
         $valMsgs = $mdl->performValidation();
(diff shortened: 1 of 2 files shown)


src/opnsense/mvc/app/controllers/OPNsense/Base/ApiMutableTableModelControllerBase.php
```diff
@@ -0,0 +1,183 @@
...
+{
+    // FIXME What about validation?
+
+    static protected $modelPathPrefix = '';
+    private function getNodes() {
+        $ref = static::$modelPathPrefix
+             + static::$internalModelName;
+        return $this->getModel()->getNodeByReference($ref);
+    }
+    private function getNodeByUUID($uuid) {
```
AdSchellevis: forgot something here, getNodes can return null if the reference wasn't found. 
So we probably should do something like this:
```
$nodes = $this->getNodes();
return !empty($nodes) ? $nodes->$uuid : null;
```

src/opnsense/mvc/app/controllers/OPNsense/Base/ApiMutableTableModelControllerBase.php
```diff
@@ -0,0 +1,183 @@
...
+    public function addItemAction()
+    {
+        $result = array("result"=>"failed");
+        if ($this->request->isPost() && $this->request->hasPost(static::$internalModelName))) {
+            $mdl = $this->getModel();
+            // FIXME Is this correct?
+            $node = $this->getNodes()->add();
+            $node->setNodes($this->request->getPost(static::$internalModelName)));
+            // FIXME What do we do with this? Is this traffic-shaper specific? Do we need a hook here?
+            $node->origin = "TrafficShaper"; // set origin to this component.
```
AdSchellevis: origin is a field not mapped in the shaper's edit form, you can delete this.

src/opnsense/mvc/app/controllers/OPNsense/Base/ApiMutableTableModelControllerBase.php
```diff
@@ -0,0 +1,183 @@
...
+
+/**
+ * Class ApiMutableTableModelControllerBase, inherit this class to implement
+ * an API that exposes a model for tables.
+ * @package OPNsense\Base
+ */
+abstract class ApiMutableTableModelControllerBase extends ApiControllerBase
+{
+    // FIXME What about validation?
+
```
AdSchellevis: not sure which validation your referring to here, the model contains field validations.

src/opnsense/mvc/app/controllers/OPNsense/Base/ApiMutableTableModelControllerBase.php
```diff
@@ -0,0 +1,183 @@
...
+     * add new item and set with attributes from post
+     * @return array
+     */
+    public function addItemAction()
+    {
+        $result = array("result"=>"failed");
+        if ($this->request->isPost() && $this->request->hasPost(static::$internalModelName))) {
+            $mdl = $this->getModel();
+            // FIXME Is this correct?
+            $node = $this->getNodes()->add();
```
AdSchellevis: it looks good, need to test. only upper case Add()

src/opnsense/mvc/app/controllers/OPNsense/Base/ApiMutableTableModelControllerBase.php
```diff
@@ -0,0 +1,183 @@
...
+            $node->setNodes($this->request->getPost(static::$internalModelName)));
+            // FIXME What do we do with this? Is this traffic-shaper specific? Do we need a hook here?
```
AdSchellevis: remove, same as above

src/opnsense/mvc/app/controllers/OPNsense/Base/ApiMutableTableModelControllerBase.php
```diff
@@ -0,0 +1,183 @@
...
+     * @return array
+     */
+    public function searchItemsAction()
+    {
+        $this->sessionClose();
+        $mdl = $this->getModel();
+        $grid = new UIModelGrid($this->getNodes());
+        return $grid->fetchBindRequest(
+            $this->request,
+            // FIXME Where do we get this list from? Is this something to be supplied by class implementor?
```
AdSchellevis: yes, we should wrap these. these are the fields displayed in the grid. in case we want to template the grid a bit more, we might add an extra definition (xml?) which describes its content.

src/opnsense/mvc/app/controllers/OPNsense/Base/ApiMutableTableModelControllerBase.php
pv2b: Analogous to the validation on setAction that happens here: https://github.com/opnsense/core/blob/master/src/opnsense/mvc/app/controllers/OPNsense/Base/ApiMutableModelControllerBase.php#L122

Come to think of it, I wonder if setAction and setItemAction might be refactorable into using a common method that does most of the work.

src/opnsense/mvc/app/controllers/OPNsense/Base/ApiMutableTableModelControllerBase.php
pv2b: Fixed :-)

src/opnsense/mvc/app/controllers/OPNsense/Base/ApiMutableTableModelControllerBase.php
pv2b: I think it's a mistake to tunnel vision on the UIModelGrid use case here, remember, the API may also be used for automation. Is it not better to return everything and then decide in the HTML what to actually expose? Or perhaps the list of requested columns should be coming from the javascript as a paramter rather than hardcoded here.

src/opnsense/mvc/app/controllers/OPNsense/Base/ApiMutableTableModelControllerBase.php
pv2b: removed

src/opnsense/mvc/app/controllers/OPNsense/Base/ApiMutableTableModelControllerBase.php
pv2b: removed

src/opnsense/mvc/app/controllers/OPNsense/Base/ApiMutableTableModelControllerBase.php
pv2b: Why uppercase? Isn't this the function we're trying to call? https://github.com/opnsense/core/blob/master/src/opnsense/mvc/app/models/OPNsense/Base/FieldTypes/ArrayField.php#L76

pv2b:
I figured out why I felt there was validation missing. I forgot to pull in a general analog to the save() method as found on https://github.com/opnsense/core/blob/master/src/opnsense/mvc/app/controllers/OPNsense/TrafficShaper/Api/SettingsController.php#L51

Going to fix now

src/opnsense/mvc/app/controllers/OPNsense/Base/ApiMutableModelControllerBase.php
```diff
@@ -108,6 +108,38 @@ protected function getModel()
...
+        $result = array("result"=>"failed","validations" => array());
+        // perform validation
+        $valMsgs = $mdl->performValidation();
+        foreach ($valMsgs as $field => $msg) {
+            // replace absolute path to attribute for relative one at uuid.
+            if ($node != null) {
+                $fieldnm = str_replace($node->__reference, static:$internalModelName, $msg->getField());
+                $result["validations"][$fieldnm] = $msg->getMessage();
+            } else {
+                $result["validations"][$msg->getField()] = $msg->getMessage();
```
pv2b: This is actually a slight difference compared to the previous code. The previous code would prepend static::$internalModelName here. I'm not sure if this matters, but either way it seemed inconsistent. What do here?

AdSchellevis: as long as you send a $node, static:$internalModelName is prepended. looking at the shaper code, the != null could probably go away because the save isn't used to save anything else.

pv2b:
I made a major commit on here, which adds functionality to "hook" the calls.

The reason for these hooks are in here is that you need to be able to take an action sometimes in addition to updating the model. Just using standard method overrides will not work, because the changes have to happen *after* the model is validated but *before* the model is serialized to the configuration.

src/opnsense/mvc/app/controllers/OPNsense/Base/ApiMutableModelControllerBase.php
```diff
@@ -118,22 +166,7 @@ public function setAction()
...
-                $result["validations"][static::$internalModelName.".".$msg->getField()] = $msg->getMessage();
-            }
-
-            // serialize model to config and save
-            if ($valMsgs->count() == 0) {
-                $mdl->serializeToConfig();
-                Config::getInstance()->save();
-                $result["result"] = "saved";
-            }
+            $result = $this->save($mdl, null, $this->setActionHook);
```
AdSchellevis: @pv2b phalcon won't like this passing, "Notice: Access to undefined property setActionHook". We probably better split the save into validate > save if you want to hook functions.

AdSchellevis:
@pv2b I've annotated some of my changes to ApiMutableModelControllerBase in https://github.com/opnsense/core/commit/618036bc55b1251095e8b520f27201ce38c9a37b I left the other controller unchanged.

PR #1135, opened 2016-08-13
(get_interface_list) Properly handle an interface not appearing in dmesg

pv2b:
get_interface_list() tries to extract interface information by grepping through dmesg. This fails if there are interfaces present that for some reason do not appear in dmesg. The code attempts to handle it, but only handles the case where the first line returned from the grep is unparsable, it does not handle the case where no lines are returned at all. This leaks an undesired PHP info or warning out on stdout if not checked.

This may be the case for example, if a user has manually cloned VLAN interfaces on the FreeBSD end. (I'm sure other things break at least, but at least we don't have to leak this particular warning any more.)

### DIFF
--- a/src/etc/inc/util.inc
+++ b/src/etc/inc/util.inc
@@ -948,12 +948,13 @@ function get_interface_list($only_active = false)
         }
 
         $dmesg_arr = array();
+        $toput['dmesg'] = null;
         exec("/sbin/dmesg |grep $ifname", $dmesg_arr);
-        preg_match_all("/<(.*?)>/i", $dmesg_arr[0], $dmesg);
-        if (isset($dmesg[1][0])) {
-            $toput['dmesg'] = $dmesg[1][0];
-        } else {
-            $toput['dmesg'] = null;
+        if (count($dmesg_arr) > 0) {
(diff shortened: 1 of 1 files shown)


fichtner:
grep(1) can do '-c' for us ;)

How useful is the dmesg inspect, especially in an inline setting like this? can we ditch it, or make it optional?

pv2b:
I'm not sure how grep -c would be useful, you'd just end up shelling out twice instead of once, and introducing a race condition in case the dmesg output changes between calls.

Anyway, because this is legacy code that's due to be ripped out, I didn't really consider the raison d'être of the code, I just saw some errors fall out of it, so I decided to fix the corner case rather than re-architect it.

The only place this field seems to be used (after a quick code search) appears to be here:
https://github.com/opnsense/core/blob/17f656da4132834afea9a9fe378506b152b197de/src/etc/inc/config.console.inc#L89

This looks like it's the console code for interface assignments, specifically the part that gives you the make and model of the network card.

I'll leave it up to you if you want to nuke it, but I think it can stay in for now :-)

fichtner:
Ok, -c won't help as the output needs to be inspected after, you're right. :) thanks, I'll make it optional for that one instance. :)

Issue #1136, opened 2016-08-13
PHP errors in API calls end up breaking the output


pv2b:
Hi.

If you have a PHP error when you call an API call, it goes straight out and intermingles with the JSON output. This ends up breaking JSON parsing and is generally annoying, especially when it comes out *before* the JSON output.

fichtner:
Sounds reasonable. display_errors should be turned off for the API, as long as the messages still make it to the log file (either the one we have or a different one, whatever makes more sense?).

AdSchellevis:
it should throw the errors (most of the time they point to actual issues), report back in json and log the traceback. 
https://github.com/opnsense/core/commit/9ed6406bdc8786e7d280a101ac652dc4b76c3b77 should fix this to the proper behaviour.

PR #1140, opened 2016-08-14
Improve error messages for API calls

pv2b:
Related to https://github.com/opnsense/core/issues/1136

Example before: `{"errorMessage":"Undefined offset: 0 [8]"}`
Example after: `{"errorMessage":"Error at \/usr\/local\/etc\/inc\/util.inc:952 - Undefined offset: 0 (errno=8)"}`

### DIFF
--- a/src/opnsense/mvc/app/controllers/OPNsense/Base/ApiControllerBase.php
+++ b/src/opnsense/mvc/app/controllers/OPNsense/Base/ApiControllerBase.php
@@ -59,13 +59,29 @@ class ApiControllerBase extends ControllerRoot
 
     /**
      * Raise errors, warnings, notices, etc.
-     * @param $errno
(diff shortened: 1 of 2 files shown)


pv2b:
Not sure where that escaping of backslashes is from or if it's neccessary?

pv2b:
Turns out json_escape escapes backslashes by default because in HTML </ may not appear in a <script> tag. We're not emitting the exception as part of HTML so we can safely disable this escaping in this case.

With this code in place, an exception will look like this:

{"errorMessage":"Error at /usr/local/etc/inc/util.inc:952 - Undefined offset: 0 (errno=8)"}

And that makes me happy :-)

AdSchellevis:
@pv2b looks good, thanks for the addition.

PR #1155, opened 2016-08-25
changes in services_unbound_advanced.php

fabianfrz:
* translation fixes (for example submit button)
* create some options in a function
* remove some not operators
* make short echo tags look more like a template

### DIFF
--- a/src/www/services_unbound_advanced.php
+++ b/src/www/services_unbound_advanced.php
@@ -106,7 +106,7 @@ include_once("head.inc");
         <?php if (isset($input_errors) && count($input_errors) > 0) print_input_errors($input_errors); ?>
         <?php if (isset($savemsg)) print_info_box($savemsg); ?>
         <?php if (is_subsystem_dirty('unbound')): ?><br/>
-        <?php print_info_box_apply(gettext("The configuration of the DNS Resolver, has been changed") . ".<br />" . gettext("You must apply the changes in order for them to take effect."));?><br />
+        <?php print_info_box_apply(gettext("The configuration of the DNS Resolver, has been changed.") . "<br />" . gettext("You must apply the changes in order for them to take effect."));?><br />
         <?php endif; ?>
         <section class="col-xs-12">
           <div class="tab-content content-box col-xs-12">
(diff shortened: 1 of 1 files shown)


src/www/services_unbound_advanced.php
```diff
@@ -133,64 +133,64 @@
                       <tr>
                         <td><a id="help_for_hideversion" href="#" class="showhelp"><i class="fa fa-info-circle"></i></a> <?=gettext("Hide Version");?></td>
                         <td>
-                          <input name="hideversion" type="checkbox" id="hideversion" value="yes" <?=!empty($pconfig['hideversion']) ? "checked=\"checked\"" : "";?> />
```
AdSchellevis: @fabianfrz I don't mind switching the order, but don't see the big gain (issues slip in very easily). For the future, can you keep the order as it is if the result is good? Thanks in advance.

src/www/services_unbound_advanced.php
```diff
@@ -338,21 +338,21 @@
...
-                            <?=gettext("Select the log verbosity.");?>
+                            <?= gettext("Select the log verbosity.") ?>
                           </div>
                         </td>
                       </tr>
                       <tr>
                         <td></td>
                         <td>
-                          <input type="submit" name="Save" class="btn btn-primary" id="save" value="Save" />
+                          <button type="submit" name="Save" class="btn btn-primary" id="save" value="Save"><?= gettext("Save") ?></button>
```
AdSchellevis: @fabianfrz shouldn't this be a normal input/submit? the submit is part of the <form/> tag.

fabianfrz: Hi @AdSchellevis, the button allows to use a text between the tags so I can use the translation, with an input type=submit this does not work (I would have to write that into the value).

AdSchellevis: @fabianfrz I understand, only thing I'm not sure about if this is compatible with all browsers, have you tested that?

fabianfrz: Works with Firefox and Vivaldi without issues

AdSchellevis: ok, I will test Safari on my end.

PR #1160, opened 2016-08-29
Camellia can only be used with IKEv2

fabianfrz:
forgot this in the commit before


fichtner:
can we please solve this in a flexible way, there's plenty of room in $p1_ealgos to register flags...

AdSchellevis:
@fichtner @fabianfrz I'll pull this one in and adjust $p1_ealgos afterwards

PR #1179, opened 2016-09-08
Update main.xml

Strykar:
Add CSV capability information to 'SSL no bump sites' form tool tip


AdSchellevis:
I don't know if this is the place to add this, all tokenize inputs support the same in OPNsense. Maybe we better put this in the docs (for example here https://docs.opnsense.org/manual/gui.html)

Strykar:
I saw the same tool tip in other places and assumed it was okay. Perhaps it is better in the docs, not my call.

fichtner:
This can't really hurt, although I don't suggest updating the help entry of every single CSV-capable field as this would remove a lot of good translations for little benefit.

Slightly modified version committed as https://github.com/opnsense/core/commit/e6f5a191901063e64c6a22fd3a45528127808b65

fichtner:
And thanks @Strykar :)

PR #1186, opened 2016-09-15
LDAP fix: handle legal boolean return result from searchUsers()

evbevz:
In case of serachUsers() returns false, this value handles wrong and Tester tool always returns successful login if in LDAP server settings does not specified "Bind credentials".


AdSchellevis:
@evbevz thank you!

fichtner:
Very nice catch, thank you!

PR #1222, opened 2016-10-17
include a custom configuration file to allow custom suricata rules

fabianfrz:


AdSchellevis:
@fabianfrz I'm not sure if this solves anything, you can already put files in the rule director which will be loaded automatically.

fabianfrz:
@AdSchellevis in this case it was my mistake and I will close it.

Kind regards

Fabian

PR #1263, opened 2016-11-16
Update unbound.inc

Tikimotel:
192.254 is a legal IPv4 range. (used by shmee150.com for instance whoops)
169.254 is the correct private IPv4 address space.
Moved the order of the private address so they go from small to big.


fichtner:
Agreed, nice catch! :)

PR #1265, opened 2016-11-19
fix bug in firewall_rules_edit.php reported by tillsense in the forum…

fabianfrz:
… at https://forum.opnsense.org/index.php?topic=3951.0;topicseen


AdSchellevis:
@fabianfrz thank you

PR #1278, opened 2016-11-26
add validation message

fabianfrz:
just add a message if the validation fails to the commit of @AdSchellevis


AdSchellevis:
@fabianfrz thank you!

PR #1282, opened 2016-11-28
Improved password hashes.

pioorg:
Replaced crypt($password, '$6$') with password_hash and password_verify in order to produce salted hashes from passwords.

crypt($password, '$6$') doesn't create any randomness in password hash, because the second parameter should contain a random salt (and optionally a number of rounds). With only '$6$' neither is provided. Switching from SHA512 to bcrypt seems to be justified be the speed of possible password cracking.

### DIFF
--- a/src/etc/inc/auth.inc
+++ b/src/etc/inc/auth.inc
@@ -488,7 +488,7 @@ function local_user_del($user)
 
 function local_user_set_password(&$user, $password)
 {
-    $user['password'] = crypt($password, '$6$');
+    $user['password'] = generate_password_hash($password, 10);
 
     // Converts ascii to unicode.
     $astr = (string) $password;
--- a/src/www/system_usermanager_passwordmg.php
+++ b/src/www/system_usermanager_passwordmg.php
@@ -40,7 +40,7 @@ if (isset($_POST['save'])) {
     do_input_validation($_POST, $reqdfields, $reqdfieldsn, $input_errors);
 
     if ($_POST['passwordfld1'] != $_POST['passwordfld2'] ||
-        $config['system']['user'][$userindex[$username]]['password'] != crypt($_POST['passwordfld0'], '$6$')) {
+        !password_verify($_POST['passwordfld0'], $config['system']['user'][$userindex[$username]]['password'])) {
         $input_errors[] = gettext("The passwords do not match.");
     }
 
@@ -58,7 +58,7 @@ if (isset($_POST['save'])) {
 
     if (count($input_errors) == 0) {
         // all values are okay --> saving changes
-        $config['system']['user'][$userindex[$username]]['password'] = crypt($_POST['passwordfld1'], '$6$');
+        $config['system']['user'][$userindex[$username]]['password'] = generate_password_hash($_POST['passwordfld1'], 10);
         local_user_set($config['system']['user'][$userindex[$username]]);
 
         write_config();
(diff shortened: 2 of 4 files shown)


fichtner:
As long as this is backwards compatible I'm fine with it. I'll do a test round soon.

Also, PHP docs suggest cost value of 12?

src/www/system_usermanager_passwordmg.php
```diff
@@ -40,7 +40,7 @@
     do_input_validation($_POST, $reqdfields, $reqdfieldsn, $input_errors);
 
     if ($_POST['passwordfld1'] != $_POST['passwordfld2'] ||
-        $config['system']['user'][$userindex[$username]]['password'] != crypt($_POST['passwordfld0'], '$6$')) {
+        !password_verify($_POST['passwordfld0'], $config['system']['user'][$userindex[$username]]['password'])) {
```
fichtner: Does password_verify() accept $2b$?

src/etc/inc/auth.inc
```diff
@@ -488,7 +488,7 @@ function local_user_del($user)
 
 function local_user_set_password(&$user, $password)
 {
-    $user['password'] = crypt($password, '$6$');
+    $user['password'] = generate_password_hash($password, 10);
```
fichtner: I'd favour not hardcoding this in the function prototype, especially as a static value.

src/www/system_usermanager_passwordmg.php
```diff
@@ -58,7 +58,7 @@
 
     if (count($input_errors) == 0) {
         // all values are okay --> saving changes
-        $config['system']['user'][$userindex[$username]]['password'] = crypt($_POST['passwordfld1'], '$6$');
+        $config['system']['user'][$userindex[$username]]['password'] = generate_password_hash($_POST['passwordfld1'], 10);
```
fichtner: Same here...

src/www/system_usermanager_passwordmg.php
pioorg: Yes, it does.

src/etc/inc/auth.inc
pioorg: I've tested this with my APU1D, and if I had chosen 12, then password verification would take about half a second. Might be too much for a default for everyone.

fichtner: Fair point on 10 vs. 12, ok. What is 10 in numbers on the same box to compare? :)

The prototype for password_hash() is pretty good, generate_password_hash() is too oriented towards the newly implemented default. In the future, there may be more calls from plugins, we can't expect to fix all of them because we e.g. want to bump the default from 10 to something else in the future. We may also have to dash out MD5 for non-local credentials, ideally routing all of this through the same function while offering sane defaults underneath is desirable.

pioorg: The time is 2^cost_factor for bcrypt, so 10 is about 4 times faster than 12. And according to my research bcrypt with 10 is about 6 times longer than current impl with crypt/SHA512 (at my test box). (Although the cost factor has 'second importance' here, this pull request is about salting it the first place.)

Regarding the defaults, there might be none which can satisfy everyone, because the hashing speed varies between environments.

As above, the idea is not to bump the default in the sources, but to make it adjustable depending on the device it's running, so the admin can focus on _time_, not cost _factor_.
The intent of this push is more of a bugfix, not a feature. The feature should come in the future. ;-)

fabianfrz: wrote a little benchmark tool to test it on my laptop (Linux on an i7-2620M) using the bcrypt gem:

100 times, cost=10, delta t=6.92085047, mean=0.0692085047
100 times, cost=12, delta t=27.533339248, mean=0.27533339248

fichtner:
Thanks, I cherry-picked your commit and wrapped the implementation into the existing functions.

PR #1285, opened 2016-11-29
Fix bug in some languages with quotes

MyTheValentinus:
Hello,
I fixed bugs with quote in my language (French)

Thanks


fichtner:
Hi Valentin,

I think the fix is simply html_safe() around the gettext() calls in JavaScript (like the other line you changed. Would you mind changing it? Otherwise it breaks on double quote now.


Thanks,
Franco

fichtner:
Thanks, I merged this and added more cleanups :)

MyTheValentinus:
Hello, thanks. I did not have the time to change html_safe.
Have nice day ;)

PR #1288, opened 2016-12-02
Update system_advanced_misc.php

Tikimotel:
This adds the "normal" power state to the powerd - system_advanced_misc options.
If the powerd utillity can not determine the system power options (ac or battery) it will use the "normal" setting rendering the "ac" and "battery" settings useless for these systems.

_powerd -v_ (from the command-line, on my intel Avoton system)
**Quote**
powerd: unable to determine AC line status <=== Important for "-n" option of powerd !!!
load  10%, current freq 1200 MHz (13), wanted freq 1200 MHz
load   6%, current freq 1200 MHz (13), wanted freq 1200 MHz
load   8%, current freq 1200 MHz (13), wanted freq 1200 MHz
load  10%, current freq 1200 MHz (13), wanted freq 1200 MHz
load   6%, current freq 1200 MHz (13), wanted freq 1200 MHz

_powerd manpage:_
**Quote**
     -n   mode    Selects the mode to use normally when the AC line state is
       unknown.

### DIFF
--- a/src/www/system_advanced_misc.php
+++ b/src/www/system_advanced_misc.php
@@ -82,16 +82,20 @@ if ($_SERVER['REQUEST_METHOD'] === 'GET') {
     /* if the old use_mfs_tmpvar is found, set these flags, too */
     $pconfig['use_mfs_var'] = isset($config['system']['use_mfs_tmpvar']) || isset($config['system']['use_mfs_var']);
     $pconfig['use_mfs_tmp'] = isset($config['system']['use_mfs_tmpvar']) || isset($config['system']['use_mfs_tmp']);
-    $pconfig['powerd_ac_mode'] = "hadp";
     $pconfig['rrdbackup'] = !empty($config['system']['rrdbackup']) ? $config['system']['rrdbackup'] : null;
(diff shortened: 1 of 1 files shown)


fichtner:
Whoops, nice catch. This seems to have gone missing in action even before 15.1 came out:

https://github.com/opnsense/core/commit/b899e523

src/www/system_advanced_misc.php
```diff
@@ -397,6 +402,28 @@ function thermal_modules()
                 </td>
               </tr>
               <tr>
+                <td><i class="fa fa-info-circle text-muted"></i>  <?=gettext('On Battery Normal Mode') ?></td>
```
Tikimotel: It's my first time editing php...
I thought the comment should be added to add some explanation for the normal option.

fichtner: If you want I can mop it up. Just wanted to leave you with the choice :)

Tikimotel: `<td><a id="help_for_powerd_normal_mode" href="#" class="showhelp"><i class="fa fa-info-circle text-circle"></i></a>  <?=gettext('On Normal Power Mode'); ?></td>`

would this work?

Tikimotel:
Edited to add the "help" text.

fichtner:
Looks good, thank you very much! 👍

Issue #1316, opened 2016-12-24
[17.1b] Destination port range fields are disabled


deviantintegral:
When adding a new firewall rule, the port range fields are disabled. I'm seeing this both in Firefox 52 and Chrome stable. AFAICT, the only rules you can add are those that don't have any port restrictions.

![rules _ firewall _ opnsense localdomain - google chrome 2016-12-24 10 25 20](https://cloud.githubusercontent.com/assets/255023/21467631/a64d1a4a-c9c3-11e6-9bc4-06c606c53631.png)

AdSchellevis:
@deviantintegral are you sure you selected tcp and/or udp as protocol?

deviantintegral:
Thanks, that's totally it.

I wonder if there's a better way we can call this out in the UI. A quick solution would be to add a note about the dependency in the help text, but it'd be even better to show it directly when no protocol is selected. I've seen other systems group protocol with the port range, but it feels like there's other issues that'd cause.

AdSchellevis:
Adding a reference to tcp/udp in the help text would probably be better in this case.

AdSchellevis:
closed with https://github.com/opnsense/core/pull/1317

PR #1317, opened 2016-12-25
Note that src and dst ports only apply to TCP and UDP

deviantintegral:
Followup from #1316. Completely untested!

### DIFF
--- a/src/www/firewall_nat_edit.php
+++ b/src/www/firewall_nat_edit.php
@@ -685,7 +685,7 @@ $( document ).ready(function() {
                       </tbody>
                     </table>
                     <div class="hidden" for="help_for_srcport">
-                      <?=gettext("Specify the source port or port range for this rule"); ?>.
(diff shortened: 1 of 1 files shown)


AdSchellevis:
@deviantintegral looks good, thanks! I'll leave this one for @fichtner to pull in, because he might want to freeze the translation for our 17.1 release.

fichtner:
Thanks for this! I doubt it is the last change we're going to do as the help text is not always read and intuitive design should be further sought out. But I don't have a vision as to how we could handle this other than what is already in there. :)

wrobelda:
Note that in 22.7 this change is only seen in `firewall_nat_edit.php`, but not `firewall_rules_edit.php` nor `firewall_rules_edit.php`: https://github.com/opnsense/core/search?q=Specify+the+port+or+port+range+for+the+destination+of+the+packet+for+this+mapping.

PR #1318, opened 2016-12-29
Optimized dhcp_clean_leases()

sxnxl:
Current implementation of dhcp_clean_leases() takes too much time to remove static MACs from dhcpd leases database, which can sometimes exceed the max_execution_time, especially when leases database is huge or too many static MACs are configured.
This implementation uses regular expressions to speed up the cleanup process.

### DIFF
--- a/src/www/services_dhcp.php
+++ b/src/www/services_dhcp.php
@@ -47,45 +47,25 @@ function dhcp_clean_leases()
         return;
     }
 
-    /* Build list of static MACs */
-    $staticmacs = array();
+    /* Build lease patterns for static MACs */
+    $lease_patterns = array();
     foreach (legacy_config_get_interfaces(array("virtual" => false)) as $ifname => $ifarr) {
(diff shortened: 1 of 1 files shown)


AdSchellevis:
@sxnxl I just tested this on my end and it works like a charm! thanks!

fichtner:
Please keep in mind that file_get_contents() can hit the PHP memory limit on reads. This still needs to be optimised. But still very good work, thank you. :)

AdSchellevis:
just for reference, I fixed the status page before in https://github.com/opnsense/core/commit/493a3d18f890f3fa66fbc12ab9177aa993579056 we should probably wrap  https://github.com/opnsense/core/blob/ce94277186f685241e72e60f0fc64c4eb1bafa49/src/www/status_dhcp_leases.php#L252-L281 into another function sometime.... 

I'm not sure about the performance implications of switching to that option, although I expect it to be faster then it originally was (and consuming way less memory).

PR #1323, opened 2017-01-03
Add file_notice() to MVC

kekek2:


AdSchellevis:
@kekek2 sorry, legacy code may not be included in our mvc base to avoid pollution... I'm not against adding a new noticing system, but gluing the old one in (which has multiple issues) is against our architecture principles.

evbevz:
@AdSchellevis, maybe move notification subsystem to MVC in form of plugin? I mean notifications in SMTP, Growl, Web-gui.

PR #1325, opened 2017-01-05
Fix setting locale

kekek2:
Setting locale by code 
`$ret = new ViewTranslator(array(
            'directory' => '/usr/local/share/locale',
            'defaultDomain' => 'OPNsense',
            'locale' => $lang_encoding,
        ));`
don't work correctly. After this $ret->_locale = false


fichtner:
Interesting. Does this reliably fix the "search bar" translation?

kekek2:
Without this patch calling `gettext("some text")` randomly return "???????"

fichtner:
Ok, nice catch and thanks! :)

PR #1327, opened 2017-01-06
Squid plugins

evbevz:
This way allows to write plugins, which adds configs to /usr/local/etc/squid/pre-auth/ and /usr/local/etc/squid/post-auth/
No templates intrusions to original squid.conf needed.
The auth plugins can be done by same way.
Developer must pay attention to removing generated plugin config file from /../etc/squid/.. folder on uninstall procedure and  accurately regenerate plugin config on enabling/disabling plugin feature. 

I think that
{% include "OPNsense/Proxy/squid.user.pre_auth.conf" ignore missing with context %}
and
{% include "OPNsense/Proxy/squid.user.post_auth.conf" ignore missing with context %}
is not convinient way for squid plugins.


AdSchellevis:
Hi Evgeny,

Your right, the include * is definitely better here. We should however also cleanup the old method when adding this one, 17.1 is probably a good timing for that.

I'm keeping this open, until I can change the old part and check our existing plugins for usage.

Regards,

Ad

fichtner:
note that for consistency the old and new positions for pre/post differ, I don't know which ones are more suitable/correct

and if we can, please let us avoid dummy files (include context missing maybe works too)

also note that this could be prone to racing with multiple files. is there really a need for multiple squid plugins (that can cooperate nicely?)

AdSchellevis:
@fichtner good point, the old order seems to be the correct one:

1. local configured auth helpers
2. pre scripts
3. internal acl's
4. post scripts

evbevz:
Franco, squid directive 'include' wants to find any file at specified path. Dummy file exists for case, when no squid plugins installed. If directory is empty, squid start fails with error.
This include method can be used for more than one plugin. Plugin configs can be ordered by numeric prefixes (for ex., 100-ipcheck.conf, 200-super-acl.conf).

PR #1332, opened 2017-01-11
Non-transparent SSL Bumping

mmorev:
In some circumstances transparent proxying is not applicable, but SSL traffic filtering is still needed. This small patch allows use SSL bumping in non-transparent mode through standard 3128 port.


AdSchellevis:
@mmorev isn't the outcome of this the same as sslbumpport to port 3128?

mmorev:
@AdSchellevis sslbumpport is used within intercept-mode https listener on loopback interface. User cannot connect to this listener directly if not redirected by firewall. Moreover, not all web browsers support different proxy servers for HTTP and HTTPS sites (like Firefox does), so using port 3128 for both http and ssl is the best way of handling user traffic.

AdSchellevis:
@mmorev your problem is with the listener, I see. I'm not sure what the downside is with enabling this by default, if there is non, I'm fine with it enabling sslbump on the default listener as well. 
It looks like the http_port tags are the same as the ones in the sslbump_httpsconfig macro, can you change them to use the macro and keep/change the help text, I don't think it's a good idea to remove the forward reference which most people use.

fichtner:
but please don't remove the redirect rule and the associated feature. it looks like changing a feature into something else, not an addition.

mmorev:
@AdSchellevis okay, I just did'nt pay attention to that macro, sorry. Will review this evening. My php isn't so good as my english :).
@fichtner of cause I will not remove anything. These are different features, which, I think, should not depend each other.

mmorev:
@AdSchellevis btw a part of code that I have modified should use this macro to create IPv4 and IPv6 listeners by design, isn't it? Currently it creates IPv4 listener only.

AdSchellevis:
@mmorev it should indeed create both listeners, but if I'm reading it correctly, it doesn't appear to be doing so with our standard rules at the moment. So sticking to what's already there is fine, adding IPv6 (for both) is also good.

mmorev:
@AdSchellevis i have made some changes, please, review

src/opnsense/mvc/app/controllers/OPNsense/Proxy/forms/main.xml
```diff
@@ -223,23 +223,24 @@
...
-                  Enable sslbump mode,
-                  which makes the proxy act as a man in the middle between the internet and your clients.<br/>
+                  Enable SSL inspection mode,
+                  which allows to log HTTPS connections information, such as requested URL and/or
+                  make the proxy act as a man in the middle between the internet and your clients.<br/>
                   Be aware of the security implications before enabling this option.
-                  <br/><br/>
-                  Transparent HTTP proxy needs to be enabled and you need nat rules to reflect your traffic
-                  for this feature to work.<br/>
-                  <a href="/firewall_nat_edit.php?template=transparent_proxy&https=1"> Add a new firewall rule </a>
```
AdSchellevis: can we leave some comment in there about adding firewall rules? Most people use this in transparant mode.

src/opnsense/service/templates/OPNsense/Proxy/squid.conf
```diff
@@ -3,38 +3,44 @@
...
-  {% endif %}
+{# wrap listener configuration for reuse #}
+{% macro listener_config(network, port='3129', tags='', protocol='') -%}
+{%     if protocol == 'ssl' %}
+{%         set listener_type = 'https_port' %}
+{%     else %}
+{%         set listener_type = 'http_port' %}
+{%     endif %}
+{%     if helpers.exists('OPNsense.proxy.forward.sslbump') and OPNsense.proxy.forward.sslbump == '1' %}
+{%         set sslparams = 'ssl-bump cert=/var/squid/ssl/ca.pem dynamic_cert_mem_cache_size=10MB generate-host-certificates=on' %}
```
AdSchellevis: you may need to set sslparams to an empty string it not in sslbump mode, I haven't test this, but I'm not sure either it will like an uninitialised string.

src/opnsense/mvc/app/controllers/OPNsense/Proxy/forms/main.xml
mmorev: This link was duplicated from proxy.forward.transparentMode field comment, where it's really necessary. Looks like someone considered these two options dependent, but they aren't.

src/opnsense/service/templates/OPNsense/Proxy/squid.conf
mmorev: Fixed that. Actually, template engine treats undefined variable as an empty string when printed (http://jinja.pocoo.org/docs/dev/templates/#variables), so it wasn't necessary

src/opnsense/mvc/app/controllers/OPNsense/Proxy/forms/main.xml
AdSchellevis: You're right, sorry about the noise here.

fichtner: so what happens to our transparent_proxy https rule template? it's now unused.

mmorev: @fichtner look above in file, it's used on line 221

mmorev:
Thank you! Hope I'll find some time to make more improvements )

src/opnsense/mvc/app/controllers/OPNsense/Proxy/forms/main.xml
fichtner: nope, the other one uses "template=transparent_proxy", and this one used "template=transparent_proxy&https=1", the former would redirect port 80, the other redirects 443.

mmorev: @fichtner Oh, yes, now I see. I think, we can merge these templates, cause no reason to use transparent mode for either http or https without each other. Could you point me where they are processed?

mmorev: @fichtner besides, both type transparent listeners are created in squid.conf when transparent mode is enabled, and it was so before

fichtner: @mmorev sure: https://github.com/opnsense/core/blob/master/src/www/firewall_nat_edit.php#L90-L119

mmorev: @fichtner I see it is nontrivial to create such link for adding two rules simultaneously. I'll think if we can create rules another way.

fichtner: @mmorev we have pluggable rules now, they could be used instead. though I liked the simplicity of the approach: rule for http transparent use, another one for https transparent use as it gives the flexibility to do the one or the other or both. plus the user could change the rule, set it for a specific range of sources or destinations. If we use a checkbox from the proxy, this flexibility will go away.

AdSchellevis: @fichtner @mmorev should we add the link for now? I also forgot about the template action here....

fichtner: I don't think there is a way around adding it back until we have a better solution

AdSchellevis: I agree, I will add it back.

fichtner:
when "Enable Transparent HTTP proxy" is on and "Enable SSL inspection" is off squid does not start anymore:

    squid: Bungled /usr/local/etc/squid/squid.conf line 9: https_port 127.0.0.1:3129 intercept

mmorev:
@fichtner See it, [here](https://github.com/mmorev/opnsense-core/commit/956ea297f8e10e147b408a10abd448682f5075cf) is a fix. Tested in all combinations of these checkboxes.

fichtner:
@mmorev many thanks, tested and committed 👍

mmorev:
@fichtner Glad to help) I very like your work and product)

fichtner:
thanks and likewise we love contributions. please do keep them coming! :)

PR #1334, opened 2017-01-13
Translate wizards.

kekek2:


fichtner:
Mid-term, this wizard implementation will be removed. Let me check how many strings are pushed into the translations, if it's not too many we can add this until we have a better replacement.

fichtner:
Merged, thanks!

Issue #918, opened 2016-04-29
replace CSRF protection


fichtner:
MVC is ok, but /usr/local/www pages need a replacement.

PR #1339, opened 2017-01-20
CSRF replacement for https://github.com/opnsense/core/issues/918

AdSchellevis:
@fichtner I think this should be it, can you review and test? 
Thanks in advance.


fichtner:
Thank you for doing the work! Check worked in all tests I could think of. Manually added to master to test normal workflow in the next days.

Great work and way less code :)

PR #1341, opened 2017-01-22
fix for https://forum.opnsense.org/index.php?topic=4309.0;topicseen

fabianfrz:
the person who reported it, confirmed that this patch works


fichtner:
Merged, thanks!

PR #1342, opened 2017-01-22
Part 2 of 2 - EAP-RADIUS ipsec/strongswan support

GurliGebis:
This pull request is split across the tools and core repositories, this is the core part.

1. commit improves readability of the generated strongswan.conf by fixing indentions, and ipsec.conf by adding some missing spaces.
2. commit implements the config file generation parts that is missing.
3. commit implements the ui part, adding two fields for entering the RADIUS server and secret.

### DIFF
--- a/src/etc/inc/ipsec.inc
+++ b/src/etc/inc/ipsec.inc
@@ -626,7 +626,7 @@ EOD;
         if (isset($a_client['enable'])) {
             $strongswan .= "\t\tattr {\n";
             if ($a_client['pool_address'] && $a_client['pool_netbits']) {
-                $strongswan .= "\t\tsubnet = {$a_client['pool_address']}/{$a_client['pool_netbits']}\n";
+                $strongswan .= "\t\t\tsubnet = {$a_client['pool_address']}/{$a_client['pool_netbits']}\n";
             }
             $cfgservers = array();
             foreach (array('dns_server1', 'dns_server2', 'dns_server3', 'dns_server4') as $dns_server) {
@@ -635,7 +635,7 @@ EOD;
                 }
             }
             if (!empty($cfgservers)) {
-                $strongswan .= "\t\tdns = " . implode(",", $cfgservers) . "\n";
+                $strongswan .= "\t\t\tdns = " . implode(",", $cfgservers) . "\n";
             }
             unset($cfgservers);
             $cfgservers = array();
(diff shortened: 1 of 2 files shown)


fichtner:
I merged the tools part, but still waiting for a final merge in a 17.1.x and feedback from Ad on this. :)

GurliGebis:
Good :) It has been years since I have touched PHP, so it might need a few changes here and there.

I have attached the <ipsec> part of my config here, for people to test: https://forum.opnsense.org/index.php?topic=4323.0

AdSchellevis:
@GurliGebis I will handle this one next week.
I have one small question, does this patch also add support for accounting, or is there more work needed for that? Our pluggable auth framework provides support for radius authentication, but doesn't provide accounting.

GurliGebis:
@AdSchellevis Great :)
All this patch does is make sure the right thing is added to the strongswan.conf and ipsec.conf for EAP-RADIUS.

The only parts added to ipsec.conf is the plugins/eap-radius part, with the "server" and "secret".
From what I know, that should make EAP-RADIUS use it for both auth and acc, but I haven't tested, since I have never had the need for accounting.

As for the RADIUS part in the pluggable auth framework, I haven't touched that.
I think that later on, it might be an idea to replace the two text boxes in my patch with a dropdown, where you can select multiple RADIUS servers defined in the auth framework.
But my plan with this is to get this up and running first, and then replace it later with the dropdown selecting one or more radius servers from the auth framework.

Does it make sense to go that route, or should I rework this to use the auth framework for selecting RADIUS servers?

GurliGebis:
Just checked the strongswan documentation - my patch does not enable accounting.
https://wiki.strongswan.org/projects/strongswan/wiki/EAPRadius

I think it would be better to save that for when it is changed to use the auth framework, since it would be possible to define it on the server there, what should be used and what shouldn't.

GurliGebis:
Okay, I know how this should be done.
Please don't merge this pull request until I give an okay (in case the idea I have does not work).

I will get back with a new pull request :)

GurliGebis:
Just got confirmation back from strongswan people - it is not possible to define different servers for different connections.

https://wiki.strongswan.org/issues/2229

So I plan on doing it the other way around - so to use this, you have to add one or more RADIUS servers to the auth framework.
Then, when you enable EAP-RADIUS for a connection, it takes all the RADIUS servers that is defined in there.

Does that sound like a usefull idea?

GurliGebis:
I will be recreating my repo, since the patches should not have been commited to the master branch (and it cannot be rolled back now that I has been pushed to github).

I will change my implementation like mentioned above, and create a new pull request from the new repository

GurliGebis:
Done, I have reimplemented it using the servers in the auth framework.

Please see the new pull request here: https://github.com/opnsense/core/pull/1345

I'm closing this one, since this version should not be pulled.

PR #1345, opened 2017-01-24
Part 2 of 2 - EAP-RADIUS ipsec/strongswan support, 2. try

GurliGebis:
This pull request is split across the tools and core repositories, this is the core part.
It has been changed, so it uses the radius servers defined in the auth framework.

1: commit improves readability of the generated strongswan.conf by fixing indentions, and ipsec.conf by adding some missing spaces.
2: commit implements the config file generation parts that is missing.
3: commit implements the ui part, adding some helper text, and hiding fields not to be used for EAP-RADIUS.

### DIFF
--- a/src/etc/inc/ipsec.inc
+++ b/src/etc/inc/ipsec.inc
@@ -626,7 +626,7 @@ EOD;
         if (isset($a_client['enable'])) {
             $strongswan .= "\t\tattr {\n";
             if ($a_client['pool_address'] && $a_client['pool_netbits']) {
-                $strongswan .= "\t\tsubnet = {$a_client['pool_address']}/{$a_client['pool_netbits']}\n";
+                $strongswan .= "\t\t\tsubnet = {$a_client['pool_address']}/{$a_client['pool_netbits']}\n";
             }
             $cfgservers = array();
             foreach (array('dns_server1', 'dns_server2', 'dns_server3', 'dns_server4') as $dns_server) {
@@ -635,7 +635,7 @@ EOD;
                 }
             }
             if (!empty($cfgservers)) {
-                $strongswan .= "\t\tdns = " . implode(",", $cfgservers) . "\n";
+                $strongswan .= "\t\t\tdns = " . implode(",", $cfgservers) . "\n";
             }
             unset($cfgservers);
             $cfgservers = array();
@@ -646,7 +646,7 @@ EOD;
                 $cfgservers[] = $a_client['wins_server2'];
             }
             if (!empty($cfgservers)) {
-                $strongswan .= "\t\tnbns = " . implode(",", $cfgservers) . "\n";
+                $strongswan .= "\t\t\tnbns = " . implode(",", $cfgservers) . "\n";
             }
             unset($cfgservers);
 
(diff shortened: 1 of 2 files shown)


fichtner:
happy to see this went it, thanks!

AdSchellevis:
@GurliGebis I pulled your commit, changed some things and added into core, my changes can be found in this commit https://github.com/opnsense/core/commit/78e265796f7f7d9c07884bbe72623a9dcad03011 

I have no radius server available at the moment to test, but the the output should be the same as it was with your code.
The radius server(s) to use can be selected now from within the gui (the xauth should eventually do the same, but switch to pam first), I rather don't automatically use all available servers.

Can you test and confirm if all is ok now?

GurliGebis:
I will give it a try later.

How do we handle if multiple connections has been created that use EAP-RADIUS, but with different servers defined?
strongswan does not support defining different servers for different connections (which is why I went with using all the defined radius servers)

GurliGebis:
Also, why has the line clearing $pconfig['caref'] when radius is used been removed?
Sending the CA when using EAP-RADIUS prints errors to the strongswan log, which does not happen when it is disabled (which is why I disabled it).

AdSchellevis:
Thanks. We need to check the config it creates now when using different ipsec tunnels, I'm not entirely sure it's generating the right config. I though originally the mobile servers should be limited to 1, although given your response I'm doubting what it does now.
Another option could be to drop the auth servers from here, and move them to the mobile tab (which is probably the better place anyway).
 
The caref input should already be disabled in the jquery part, if it's not reset, we should fix the general behaviour. Although I think we're already good here.

GurliGebis:
Moving it to the mobile tab might be a better solution indeed.

So it would be like if you configure them at the Mobile config, it forces the created phase1 to use EAP-RADIUS?
Then we need to find some way to prevent other phase1's from using EAP-RADIUS, since it does not make sense for them to be able to select it there - is there a way to do that?

GurliGebis:
It works file - I'll delete my branch, since there is no reason to keep it around.

There is one thing we need to have a look at.

When EAP-RADIUS has just been configured, and the config has been save and strongswan has been reloaded - it is not possible to connect until the strongswan processes has been stopped and started completly.

If you try connecting before doing so, this is what gets printed to the log:

`Jan 31 14:34:23 gateway1 charon: 02[NET] sending packet: from 172.16.15.220[4500] to 172.16.15.128[4500] (1856 bytes)
Jan 31 14:34:23 gateway1 charon: 15[NET] received packet: from 172.16.15.128[4500] to 172.16.15.220[4500] (96 bytes)
Jan 31 14:34:23 gateway1 charon: 15[ENC] parsed IKE_AUTH request 2 [ EAP/RES/ID ]
Jan 31 14:34:23 gateway1 charon: 15[IKE] received EAP identity 'HOME\TestUser'
Jan 31 14:34:23 gateway1 charon: 15[IKE] loading EAP_RADIUS method failed
Jan 31 14:34:23 gateway1 charon: 15[ENC] generating IKE_AUTH response 2 [ EAP/FAIL ]
Jan 31 14:34:23 gateway1 charon: 15[NET] sending packet: from 172.16.15.220[4500] to 172.16.15.128[4500] (80 bytes)`

It is not a bug in OPNsense, but it would be nice if we somehow could detect that there has been changes to the EAP-RADIUS settings, and the offer to restart the service along with regenerating the config.

AdSchellevis:
@GurliGebis I just checked, you should not be able to create more then one mobile phase1 tunnel, only if you feed the url manually (which is not really intended use, things will break...)

The restart thing is quite difficult to fix, without adding a lot of unwanted complexity, maybe you can add an issue for that... maybe for another time.
I will refactor the server selection, when that is done, can you do a final test?

AdSchellevis:
@GurliGebis can you test https://github.com/opnsense/core/commit/f12e4e4b2f19803a443766131f4350b3eafc12e7 ? This should be it, disabling Xauth when radius is used, eventually we should switch to pam for other authentication options anyway, maybe dropping the auth setting in the mobile page.

L1ghtn1ng:
> It is not a bug in OPNsense, but it would be nice if we somehow could detect that there has been changes to the EAP-RADIUS settings, and the offer to restart the service along with regenerating the config

Could we not have a button that did this maybe with help text on what it does?

GurliGebis:
@AdSchellevis it works fine - the result is that the xauth-generics section is gone from the strongswan.conf - so it seems to be working :)

@L1ghtn1ng that might work - I have no idea on how to create a button that triggers a restart of the service though.

Issue #1348, opened 2017-01-25
Why does auth_get_authserver replicate the connection string configuration?


EugenMayer:
Eventhough the Auth-Backend Implementations do their own thing in parsing the config and creating the configuration strings, auth_get_authserver replicates that, introducing possible erros and confusion, while later, passed to the OPNsense\Auth\AuthenticationFactory anyway in authenticate_user .. beside that little stunt here https://github.com/opnsense/core/blob/master/src/etc/inc/auth.inc#L754 which should go into LDAP https://github.com/opnsense/core/blob/master/src/opnsense/mvc/app/library/OPNsense/Auth/LDAP.php as a precondition into the constructor.

Nevertheless, the point is here, $authCFG is not even used, its created and passed here https://github.com/opnsense/core/blob/master/src/etc/inc/authgui.inc#L207 and then gets thrown away since the factory ignore is, out of good reasons since it implements config parsing itself while https://github.com/opnsense/core/blob/master/src/etc/inc/auth.inc#L701 is a plain duplication which is error prone

AdSchellevis:
timeout due to inactivity

Issue #1346, opened 2017-01-24
ldap starttls support


EugenMayer:
Seems like pfsense does already have this, but this is hugely important to be supported.

Especially considered that openldap deprecated ldaps since 2.0 should be reason enough to start working on this http://www.openldap.org/faq/data/cache/605.html

fitch pointed me on https://github.com/opnsense/core/blob/master/src/opnsense/mvc/app/library/OPNsense/Auth/LDAP.php#L353 so thats the starting point.
I guess we want http://php.net/manual/de/function.ldap-start-tls.php there, including an option the the form of the server configuration ( or rather say an option in the select ). Port should stay 389 as usual for startTLS.

In the end we need to run https://github.com/opnsense/core/blob/master/src/opnsense/mvc/app/library/OPNsense/Auth/LDAP.php#L251 those 3 things after the ldap_connect

    if($this->useStartTLS) {
      ldap_set_option($con, LDAP_OPT_PROTOCOL_VERSION, 3);
      ldap_set_option($con, LDAP_OPT_REFERRALS, 0);
      $encrypted = (ldap_start_tls($this->ldapHandle));
    }

And of course we need to patch https://github.com/opnsense/core/blob/master/src/opnsense/mvc/app/library/OPNsense/Auth/LDAP.php#L193 to include the property beside introducing those in the fields https://github.com/opnsense/core/blob/master/src/opnsense/mvc/app/library/OPNsense/Auth/LDAP.php#L97

For me at least thats more or less the show-stopper for opnsense and that would be horrible ;/

EugenMayer:
I could provide a PR for the connection part, but i am not able to find out how to introduce the option / were the FAPI is.

Looks like the factory is generating the options: https://github.com/opnsense/core/blob/master/src/opnsense/mvc/app/library/OPNsense/Auth/AuthenticationFactory.php#L147 .. but ldap does not implement getConfigurationOptions

Cannot really find the configuration tempaltes in the views either https://github.com/opnsense/core/tree/master/src/opnsense/mvc/app/views/OPNsense

EugenMayer:
That one gets in our way https://github.com/opnsense/core/issues/1348 .. but not really since it gets ignore anyway in the end, never used

fichtner:
This commit should include all relevant files: https://github.com/opnsense/core/commit/0fb3fd78cbfe059001016a30b14fbe481f169282

[...]

PR #1350, opened 2017-01-25
Implement startTLS for ldap #1346

EugenMayer:
ignore the implementation in #1348 since it is a non-used replication

### DIFF
--- a/src/opnsense/mvc/app/library/OPNsense/Auth/LDAP.php
+++ b/src/opnsense/mvc/app/library/OPNsense/Auth/LDAP.php
@@ -97,6 +97,10 @@ class LDAP implements IAuthConnector
     private $userDNmap = array();
 
     /**
+     * @var bool if true, startTLS will be initialized
+     */
+    private $useStartTLS = false;
+    /**
      * close ldap handle if open
      */
     private function closeLDAPHandle()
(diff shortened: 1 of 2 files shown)


EugenMayer:
This is untested, since i have yet not clue how to drive a test setup. I rather expect it to work, but not sure entirely because of #1348

EugenMayer:
Just tested on my FreeBSD11, opnsense 17.1 server with `opsense-patch b9e4d0eb11811a2619126b33c8a39aa299002b7f` and it worked nicely.

The only thing i now need to re-verify is, that encryption was actually really used during the process, since i did not block 336 plaintext yet. Anybody has a good idea on how tot do that?

fichtner:
We'll keep this in -devel for a bit, asking for broader testing. So far: thanks very much, this is great!

fichtner:
whoops, that wasn't the master branch :D

EugenMayer:
well we need for sure someone who can ensure startTLS is actually used, so either packet sniffing or a startTLS only server - just so we dont fool anybody here :)

EugenMayer:
As far as i can see, the container select does not work in the form when startTLS is activated

fichtner:
step by step, the picker needs the settings passed, see https://github.com/opnsense/core/blob/master/src/www/system_usermanager_settings_ldapacpicker.php

EugenMayer:
this is seriously the 3rd dublication of the authentication code? Considering what we are talking about, that is .. bad ;/

fichtner:
in other news I just got phpstorm up and running. great stuff, thanks for the recommendation!

PR #1354, opened 2017-01-26
fix spelling issue in firmware message

fabianfrz:
just because this should not be in the final release ;)


fichtner:
Nice catch! Still up for review, which seems to have worked. Many thanks.

PR #1355, opened 2017-01-26
UIModelGrid: use empty string instead of "##Unlinked"

fraenki:
Hmm, does it serve any purpose to set `##Unlinked` instead of just using an empty string?
It looks ugly if empty fields get filled with `##Unlinked` :)

unlinked


AdSchellevis:
@fraenki you're probably right, if the field is supposed to be linked to something it would be better to tell something about the status, but we can't really make the proper destinction here.

PR #1382, opened 2017-02-07
DHCP leases: do not link to WOL page if WOL plugin is missing

fraenki:
When clicking on the MAC address in DHCP Leases page, you'll get a "404 - Not Found". This is caused by a missing WOL plugin.

My proposal is to check if the WOL plugin is installed. If not, do not link to the WOL page. To make this check possible I've added a new function `is_plugin_installed()`. It may be used for similar glue in the future. Still, there may be better ways to solve this. :)

Reported on IRC by stormy1.

### DIFF
--- a/src/etc/inc/services.inc
+++ b/src/etc/inc/services.inc
@@ -2198,3 +2198,46 @@ function service_control_restart($name, $extras)
 
     return sprintf(gettext("%s has been restarted."), htmlspecialchars($name));
 }
+
+function is_plugin_installed($name)
+{
+    // NOTE: Based on infoAction() from Core/Api/FirmwareController.php
+
+    $backend = new \OPNsense\Core\Backend();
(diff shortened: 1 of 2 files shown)


fichtner:
I think for a short-term fix it's better to probe for the plugin file /usr/local/etc/inc/plugins.inc.d/wol.inc

fichtner:
Or actually, probe for the file that is linked to... :)

fraenki:
I think it wouldn't be ideal to check for the presence of a file, because this might break very easily. (I think of it this way: The WOL plugin is rather small and that's why a good candidate to migrate it to the MVC framework rather sooner than later. This would break such a check immediately.)

fabianfrz:
@fraenki In this case it would also break the link as long as they are not generated by the mvc routing like it is done in Ruby on Rails (you get a constant like wol_path, which is used in link tags)

PR #1388, opened 2017-02-08
DHCP leases: do not link to WOL page if WOL plugin is missing (2nd attempt)

fraenki:
a follow-up to #1382

### DIFF
--- a/src/www/status_dhcp_leases.php
+++ b/src/www/status_dhcp_leases.php
@@ -76,6 +76,8 @@ function remove_duplicate($array, $field)
 
 $interfaces = legacy_config_get_interfaces(array('virtual' => false));
 $leasesfile = services_dhcpd_leasesfile();
+/* Check if WOL plugin is installed */
+$plugin_installed = count(plugins_scan('wol'));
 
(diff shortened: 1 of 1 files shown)


fichtner:
this looks better. sorry for nitpicking, but:

mark the plugin_enabled with a /* XXX refactor plugin detection */

and:

value = boot ? true : false;

equals

value = bool;

fichtner:
I've added a commit, you can now check for

value = count(plugins_scan('wol'));

fraenki:
True ;) Will push a fix that uses the enhanced `plugins_scan()` function.

fraenki:
Done. Please review :)

src/www/status_dhcp_leases.php
```diff
@@ -417,9 +419,13 @@ function remove_duplicate($array, $field)
                   <td><?=$data['int'];?></td>
                   <td><?=$data['ip'];?></td>
                   <td>
+                      <? if ((int)$plugin_installed > 0) { ?>
```
fichtner: $plugin_installed is enough

fraenki:
And now? :)

fichtner:
Merged, thanks! :)

PR #1392, opened 2017-02-09
Add priority settings to advanced firewall rule options

djGrrr:
This will add 2 additional advanced options for firewall rules. One to set the priority, and one to match based on the priority. This should be useful for Google Fiber users to unlock full upload speed.

### DIFF
--- a/src/etc/inc/filter.inc
+++ b/src/etc/inc/filter.inc
@@ -2231,7 +2231,7 @@ function filter_generate_user_rule(&$FilterIflist, $rule)
     $int = "";
     $aline = array();
     // initialize array with empty tags
-    foreach (array('schedlabel','divert','icmp-type','icmp6-type','tag','tagged','route','os','reply','prot','log', 'priority', 'priority-match') as $tag) {
+    foreach (array('schedlabel','divert','icmp-type','icmp6-type','tag','tagged','route','os','reply','prot','log', 'set-prio', 'set-prio-alt', 'prio') as $tag) {
         $aline[$tag] = "";
     }
 
@@ -2389,13 +2389,19 @@ function filter_generate_user_rule(&$FilterIflist, $rule)
         $aline['icmp6-type'] = "icmp6-type {$rule['icmptype']} ";
     }
 
-    if (!empty($rule['priority'])) {
-    	// need to use a hack for priority values since rule apparantly cannot store "0"
-		$aline['priority'] = ' set prio '.($rule['priority'] - 1).' ';
+    if (isset($rule['set-prio']) && $rule['set-prio'] !== '') {
+    	if (isset($rule['set-prio-alt']) && $rule['set-prio-alt'] !== '') {
+    		$prio = '('.$rule['set-prio'].','.$rule['set-prio-alt'].')';
+		}
+		else {
+			$prio = $rule['set-prio'];
+		}
+
+		$aline['set-prio'] = ' set prio '.$prio.' ';
 	}
 
-	if (!empty($rule['priority-match'])) {
-		$aline['priority-match'] = ' prio '.($rule['priority-match'] - 1).' ';
+	if (isset($rule['prio']) && $rule['prio'] !== '') {
+		$aline['prio'] = ' prio '.$rule['prio'].' ';
 	}
 
     if (!empty($rule['tag'])) {
@@ -2549,7 +2555,7 @@ function filter_generate_user_rule(&$FilterIflist, $rule)
     $line .= $aline['type'] . $aline['direction'] . $aline['log'] . $aline['quick'] . $aline['interface'] .
       $aline['reply'] . $aline['route'] . $aline['ipprotocol'] . $aline['prot'] . $aline['src'] . $aline['os'] . $aline['dst'] .
       $aline['divert'] . $aline['icmp-type'] . $aline['icmp6-type'] . $aline['tag'] . $aline['tagged'] .
-      $aline['priority'] . $aline['priority-match'] . $aline['allowopts'] . $aline['flags'] . $aline['schedlabel'];
+      $aline['set-prio'] . $aline['prio'] . $aline['allowopts'] . $aline['flags'] . $aline['schedlabel'];
 
     unset($aline);
 
--- a/src/etc/inc/filter.inc
+++ b/src/etc/inc/filter.inc
@@ -2231,7 +2231,7 @@ function filter_generate_user_rule(&$FilterIflist, $rule)
     $int = "";
     $aline = array();
     // initialize array with empty tags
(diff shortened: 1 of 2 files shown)


src/etc/inc/filter.inc
```diff
@@ -2388,6 +2388,16 @@ function filter_generate_user_rule(&$FilterIflist, $rule)
     if (isset($rule['protocol']) && $rule['protocol'] == "icmp" && $rule['icmptype'] && $rule['ipprotocol'] == "inet6") {
         $aline['icmp6-type'] = "icmp6-type {$rule['icmptype']} ";
     }
+
+    if (!empty($rule['priority'])) {
+    	// need to use a hack for priority values since rule apparantly cannot store "0"
+		$aline['priority'] = ' set prio '.($rule['priority'] - 1).' ';
```
fichtner: also, pf.conf says you can set "set prio (2, 5)", we should account for this right away, also in the GUI, so we need a third value "set-prio-ack"

src/www/firewall_rules_edit.php
```diff
@@ -349,13 +349,20 @@ function is_posnumericint($arg) {
         $input_errors[] = gettext("If you specify TCP flags that should be set you should specify out of which flags as well.");
 
 
+    // must use value + 1 in comparison due to not being able to store "0" in rule config
+    if (!empty($pconfig['priority']) && ($pconfig['priority'] < 1 || $pconfig['priority'] > 8))
+        $input_errors[] = gettext('Priority must be between 0 and 7.');
```
fichtner: curly brackets, always

src/www/firewall_rules_edit.php
```diff
@@ -349,13 +349,20 @@ function is_posnumericint($arg) {
...
+
+    if (!empty($pconfig['priority-match']) && ($pconfig['priority-match'] < 1 || $pconfig['priority-match'] > 8))
+        $input_errors[] = gettext('Priority must be between 0 and 7.');
```
fichtner: curly too

src/www/firewall_rules_edit.php
fichtner: From the error texts, make clear which value is wrong. This isn't unlikely to hit since the value is provided via select box, but still...

src/www/firewall_rules_edit.php
```diff
@@ -1202,6 +1209,50 @@ function is_posnumericint($arg) {
                         </div>
                       </td>
                   </tr>
+<?
+$priorities = [
```
fichtner: we use array(), not []

If you create this array, map the values as well:

'' => gettext("None"),
'1' => gettext("Background"),
etc.

src/www/firewall_rules_edit.php
```diff
@@ -1202,6 +1209,50 @@ function is_posnumericint($arg) {
...
+                        <select name="priority-match">
+                          <option value="0"></option>
+<?  foreach ($priorities as $prio => $priority) {
+    $prio_value = $prio + 1;
+?>
+                          <option value="<?=$prio_value;?>"<?=(!empty($pconfig['priority-match']) && $pconfig['priority-match'] == $prio_value ? ' selected="selected"' : '');?>><?=$prio.' - '.htmlspecialchars($priority);?></option>
+<? } ?>
+                        </select>
+                        <div class="hidden" for="help_for_priority_match">
+                          <?= sprintf(gettext("match on the priority (Class of Service) of packets ."),'<b>','</b>') ?>
```
fichtner: You use sprintf but it has no %s. Also upper case for "Match".

src/www/firewall_rules_edit.php
```diff
@@ -1202,6 +1209,50 @@ function is_posnumericint($arg) {
...
+                        <select name="priority">
+                            <option value="0"></option>
+<?  foreach ($priorities as $prio => $priority) {
+    $prio_value = $prio + 1;
+?>
+                            <option value="<?=$prio_value;?>"<?=(!empty($pconfig['priority']) && $pconfig['priority'] == $prio_value ? ' selected="selected"' : '');?>><?=$prio.' - '.htmlspecialchars($priority);?></option>
+<? } ?>
+                        </select>
+                        <div class="hidden" for="help_for_priority">
+                            <?= sprintf(gettext("Set the priority (Class of Service) of packets matching this rule."),'<b>','</b>') ?>
```
fichtner: Here too: spurious sprintf

src/etc/inc/filter.inc
AdSchellevis: @fichtner compare with "" of 0 is a valid option

fichtner: won't this will also enable value 0 all the time?

AdSchellevis: I'm not sure what you mean, $rule['priority'] == "" won't store empty values, to be on the safe site for parameter initialisation, you could add isset($rule['priority'])) && $rule['priority'] == ""

Or do you mean something else?

fichtner: Not sure. It's early in the morning. :)

We need these values: "", 0, 1, 2, 3, 4, 5, 6, 7

AdSchellevis: true, it's early :) 

I expect that "" should not lead to a 'set prio()', in that case you should be fine.

fichtner: Ah, should probably use strong type compare here!

"" !== 0

AdSchellevis: yes, sorry, like you said.... early :)

djGrrr:
After testing the functionality in pf itself, i realized that "set prio" is not functioning, or at the very least, it is not setting the vlan pcp value, which makes it kind of useless.

After realizing this, I did some searching and realized that the vlanpcp setting in ifconfig was also brought into FreeBSD 11, and it does work to set a static outgoing pcp for the vlan, which I have created a (much smaller) patch for in #1394. This new patch has been tested by a user on google fiber, and it does work as expected, unlocking the full upload speed when the VLAN Priority is set to 3.

Based on this, I am wondering if we should just close this pull request, and perhaps re-visit this once the bugs are worked out of the pf set prio rules.

djGrrr:
I have pushed fixes for everything mentioned (i think), figured out how to use 0 directly in the rules, and also added support for dual priorities in set prio.

Keep in mind that as far as i can tell, set prio does not work (a FreeBSD issue), I have no way to to test the match priority, but the rules are generated correctly.

djGrrr:
After testing this on stock FreeBSD 11, I can definitely confirm that set prio does not work, even a rule as simple as:
pass out quick on vtnet0.6 set prio 2

Then using:
tcpdump -eni vtnet0 'vlan'
The pcp value (show as simply p) stays 0 on every vlan packet, yet when changing the vlanpcp setting with ifconfig, it is immediately seen in the dump.

fichtner:
merged, thanks! some more changes follow in a bit

PR #1394, opened 2017-02-10
add VLAN Priority (PCP) setting to VLAN config

djGrrr:
This will allow setting a static outbound VLAN PCP (priority code point) in the VLAN settings.

This allows Google Fiber users to obtain full upload speeds.

### DIFF
--- a/src/etc/inc/interfaces.inc
+++ b/src/etc/inc/interfaces.inc
@@ -213,7 +213,7 @@ function interface_vlan_configure(&$vlan)
         mwexecf('/usr/sbin/ngctl name %s: %s', array($tmpvlanif, $vlanif));
     }
 
-    legacy_vlan_tag($vlanif, $if, $vlan['tag']);
+    legacy_vlan_tag($vlanif, $if, $vlan['tag'], $vlan['pcp']);
 
     interfaces_bring_up($vlanif);
     interfaces_bring_up($if);
--- a/src/etc/inc/interfaces.inc
+++ b/src/etc/inc/interfaces.inc
@@ -213,7 +213,8 @@ function interface_vlan_configure(&$vlan)
         mwexecf('/usr/sbin/ngctl name %s: %s', array($tmpvlanif, $vlanif));
     }
 
-    legacy_vlan_tag($vlanif, $if, $vlan['tag'], $vlan['pcp']);
+    $pcp = isset($vlan['pcp']) ? $vlan['pcp'] : 0;
+    legacy_vlan_tag($vlanif, $if, $vlan['tag'], $pcp);
 
     interfaces_bring_up($vlanif);
     interfaces_bring_up($if);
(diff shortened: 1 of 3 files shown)


djGrrr:
I noticed that you don't like using [] for arrays so i fixed that

src/etc/inc/interfaces.lib.inc
```diff
@@ -150,9 +150,10 @@ function legacy_bridge_member($ifs, $member)
     }
 }
 
-function legacy_vlan_tag($ifs, $member, $tag)
+function legacy_vlan_tag($ifs, $member, $tag, $pcp)
```
fichtner: You need a default here to avoid errors with old configs

src/etc/inc/interfaces.inc
```diff
@@ -213,7 +213,7 @@ function interface_vlan_configure(&$vlan)
         mwexecf('/usr/sbin/ngctl name %s: %s', array($tmpvlanif, $vlanif));
     }
 
-    legacy_vlan_tag($vlanif, $if, $vlan['tag']);
+    legacy_vlan_tag($vlanif, $if, $vlan['tag'], $vlan['pcp']);
```
fichtner: Make sure this doesn't cause a warning for old configs where the value is not set

src/etc/inc/interfaces.lib.inc
```diff
@@ -150,9 +150,10 @@ function legacy_bridge_member($ifs, $member)
...
 {
-    $cmd = '/sbin/ifconfig ' . escapeshellarg($ifs) . ' vlandev ' . escapeshellarg($member) . ' vlan ' . escapeshellarg($tag);
+	$pcp = (int)$pcp;
```
fichtner: Zap this

src/www/interfaces_vlan_edit.php
```diff
@@ -200,6 +206,31 @@
                       </div>
                     </td>
                   </tr>
+<?
+$priorities = array(
+    1   => gettext('1 - Background'),
+    0   => gettext('0 - Best Effort (default)'),
```
fichtner: Default value first

djGrrr: Actually, it doesn't make sense to put the default value first in this case, since they are in order of priority, 1 - Backgroud is lower priority than 0 - Best Effort

fichtner: I agree but the browser does not care and if the user isn't explicitly selecting the priority he sets a lower priority. think about users who edit their old vlan id and hit save..

djGrrr: Except that is not the case, 0 is the default, it will always be selected when no existing value is set (NULL == 0). It's already 0 in ifconfig as well.

djGrrr:
I've pushed fixes for the other things you mentioned

src/www/interfaces_vlan_edit.php
fichtner: Good point, ok :)

fichtner:
merged, thanks!

PR #1395, opened 2017-02-12
Fix ipv4 subnet matching

djGrrr:
This will fix the issue with tun interfaces (ie openvpn) having their subnet detected incorrectly. It also brings the code more in line with the IPv6 code.


fichtner:
Nice catch, thanks!

Issue #1353, opened 2017-01-26
Add Google Domains to the Dynamic DNS client list


oneoffdallas:
The addition of Google Domains to the Dynamic DNS client list would be great. It appears to follow a similar format to other Dynamic DNS clients.

fichtner:
available in 17.1.3

PR #1411, opened 2017-02-18
Dynamic DNS: Support Google Domains

alasley:
Addresses issue https://github.com/opnsense/core/issues/1353

### DIFF
--- a/src/etc/inc/dyndns.class
+++ b/src/etc/inc/dyndns.class
@@ -248,7 +248,7 @@
 				case 'eurodns':
 				case 'gratisdns':
 				case 'ovh-dynhost':
-                                case 'googledomains':
+				case 'googledomains':
 				case 'citynetwork':
 				case 'duckdns':
 					$this->_update();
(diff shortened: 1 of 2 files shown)


src/etc/inc/dyndns.class
```diff
@@ -702,6 +703,19 @@
...
 					curl_setopt($ch, CURLOPT_URL, $server . '?domains=' . str_replace('.duckdns.org', '', $this->_dnsHost) . '&token=' . $this->_dnsUser);
 					break;
+				case 'googledomains':
+					if ($this->_dnsVerboseLog)
+						log_error("Google Domains: ({$this->_dnsHost}): DNS update() starting.");
+					$server = "https://domains.google.com/nic/update";
+					$post_data['hostname'] = $this->_dnsHost;
+					$post_data['myip'] = $this->_dnsIP;
+					$post_data['offline'] = 'no';
+					curl_setopt($ch, CURLOPT_SSL_VERIFYPEER, FALSE);
```
fabianfrz: Is that really required as it is bad for security reasons?

src/etc/inc/dyndns.class
```diff
@@ -702,6 +703,19 @@
...
+					curl_setopt($ch, CURLOPT_USERPWD, $this->_dnsUser.':'.$this->_dnsPass);
+					curl_setopt($ch, CURLOPT_POSTFIELDS, $post_data);
+					curl_setopt($ch, CURLOPT_URL, $server);
+					curl_setopt($ch, CURLOPT_IPRESOLVE, CURL_IPRESOLVE_V4);
```
fabianfrz: This may make it unable to update IPv6 addresses

src/etc/inc/dyndns.class
alasley: After removing it and testing, I now get this error:
`opnsense: /services_dyndns_edit.php: Curl error occurred: SSL certificate problem: unable to get local issuer certificate`

Google recently began operating its own CA, which it now uses for domains.google.com. The root certificate does not appear to be in `/usr/local/share/certs/ca-root-nss.crt`, although I believe the root cert is cross-signed and I'm not 100% sure that is where curl is looking for it.

On the other hand, `CURLOPT_SSL_VERIFYPEER` is set to false in every other case except the custom one, where the user has an option in the GUI to set it to true. So this may be part of a larger issue in the codebase.

It looks like we may need to explicitly set `CURLOPT_CAINFO` to tell curl where to find certificates, if there is not a default that points to something useful.

src/etc/inc/dyndns.class
alasley: That is correct, although there should be no impact because (at least, last I checked) Google Domains does not support dynamic DNS for IPv6 addresses anyway. A helpful error message might be useful though.

I included that line because I ran into a bug similar to this:
https://redmine.pfsense.org/issues/3858

I think there needs to be a more general solution which correctly chooses to resolve IPv4 or IPv6 based on what we are trying to update.

src/etc/inc/dyndns.class
fabianfrz: If I look at the certificate, Google uses an own CA as an intermediate CA which is signed by GeoTrust Global CA.

spectacle db3271

Maybe curl does not check for the next certificate in the chain if it is missing. In this case it would be better, if the CA is added to the global store or pinned for this curl request.

alasley: After further research and testing, I found that the issue is limited to installations using LibreSSL, and it looks like it has already been fixed for the next release (https://github.com/libressl-portable/portable/issues/80).

In the meantime, `CURLOPT_SSL_VERIFYPEER` can be set to true and it will work correctly for OpenSSL installations.

fabianfrz: This is probably the best option for now - with an entry in the release notes that it will work only on OpenSSL until the LibreSSL patch is released due to a bug in LibreSSL.
With `CURLOPT_SSL_VERIFYPEER` set to false it is insecure and may be attacked by a man in the middle attack and therefore the token could be stolen. In my opinion insecure code cannot be merged.

@fichtner  What do you think?

fichtner: Not want to ship a known broken feature, even if it's "just broken for LibreSSL".

On a very related note, almost all, if not all of the DynDNS services set CURLOPT_SSL_VERIFYPEER to false.

fabianfrz: @fichtner so wait for the LibreSSL patch?

fichtner:
@alasley unfortunately, I have to ask where this code is from? did you write it?

src/etc/inc/dyndns.class
fichtner: while we wait, why not fix all the other insecure spots? ;)

but seriously, in general, this can go on master (-devel) no problem. for the release we should be more careful.

alasley:
@fichtner The code is extremely similar to the implementation in the pfSense version of `dyndns.class`, which was added at some point after the fork. Given the similarity of the two files it would not be possible to implement the `googledomains` case very much differently.

Nonetheless, I apologize if I misrepresented anything.

fichtner:
@alasley fair enough, thank you :)

fichtner:
Merged, thanks!

Issue #1400, opened 2017-02-15
[BUG] DNS Resolver registers ipv6 link-local addresses


phpb-com:
When DNS Resolver is configured to register static mapping, it places firewalls link local addresses in the record. That should not happen and breaks resolution to the proper address.

phpb-com:
Forgot to mention that I am on the latest version, 17.1.1

phpb-com:
Also, it seems that localhost (::1,127.0.0.1) addresses are also returned by unbound when "Register DHCP static mappings in the DNS Resolver" flag is checked (not expected), with WAN address missing (as expected).

fichtner:
For reference, this is the code, by no means conditional on these:

https://github.com/opnsense/core/blob/master/src/etc/inc/plugins.inc.d/unbound.inc#L538

I know dnsmasq/dns forwarder does not generate all of these, but I feel impartial to changing something here as long as we don't break "good" and "expected" behaviour.

What do you think?

phpb-com:
The problem with the current approach is that when the client relies on firewall for DNA resolution and tries to resolve firewall name for whatever purpose (ping,etc) it results in wrong address used. I have multiple local interfaces (non-WAN) and client in one subnet gets IPv6 link-local address from another. In this case it cannot contact firewall. Localhost results in even worse behavior and that is not expected from resolver. I think we should filter all of link local (fe80::) and local hosts from the dns entries.

phpb-com:
And by localhost I mean the firewall name resolving to 127.0.0.1 or ::1

phpb-com:
If it is important to have lladdr in there, maybe remove it if more than one LAN interface is defined or have an option to disable it?

phpb-com:
Digging through the internet, found an interesting newsletter about DNS and IPv6 http://www.kame.net/newsletter/20000216/ . Perhaps one more approach could be to remove link-local addresses from the DNS as soon as the interface(s) has globally routable ipv6 address. In my case, I would prefer to keep them away from DNS, that includes any reference to localhost as well, unless it points to localhost.<your domain> or localhost.

phpb-com:
Fix submitted and merged, thanks.

PR #1416, opened 2017-02-20
Add option to do not register ipv6 link-local addresses in the DNS Resolver

phpb-com:
This should provide a good compromise to solve #1400


fichtner:
This looks good, but we should flip the config value to retain the default behaviour? Only the config underneath, the dialog should stay the same. If you're ok with this I'll do it after merge.

phpb-com:
Thanks, yes please go ahead and fix it after merge.

fichtner:
Alright... Merged, thank you :)

PR #1418, opened 2017-02-21
Add modulate state as an option for state tracking

phpb-com:
Seems like most of the work for that was in the code already, only GUI option was missing. This commit adds the option in the GUI with appropriate checks, that are identical to synproxy state.

### DIFF
--- a/src/www/firewall_rules_edit.php
+++ b/src/www/firewall_rules_edit.php
@@ -247,7 +247,7 @@ if ($_SERVER['REQUEST_METHOD'] === 'GET') {
     if ($pconfig['protocol'] == "icmp" && !empty($pconfig['icmptype']) && $pconfig['ipprotocol'] == "inet46") {
         $input_errors[] =  gettext("You can not assign a ICMP type to a rule that applies to IPv4 and IPv6");
     }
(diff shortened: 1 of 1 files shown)


fichtner:
I'm leaving this for @AdSchellevis -- it looks ok, but the new rules have a "modulate" state and sometimes use it. I don't know if "modulate" or "modulate state" is the expected tag here...

phpb-com:
modulate state is the correct tag that is expected, already tested the change and done pfctl -f /tmp/rules.debug , loads without any issues.

fichtner:
I'm not questioning that. :) I merely found other parts that were untouched and look suspiciously out of sync:

https://github.com/opnsense/core/blob/master/src/etc/inc/filter.lib.inc#L195
https://github.com/opnsense/core/blob/master/src/opnsense/mvc/app/library/OPNsense/Firewall/FilterRule.php#L211-L222

AdSchellevis:
@phpb-com thanks for your contribution, when we use the new code base for generating rules this should still work because our default parsing ignores the additional "state" https://github.com/opnsense/core/blob/master/src/opnsense/mvc/app/library/OPNsense/Firewall/FilterRule.php#L221

PR #1422, opened 2017-02-22
enable tls certificate checks for some services

fabianfrz:
* Google Domains
* DuckDNS

they should work as they have a valid certificate (browser does trust them as well as curl URL works from my computer without `-k`)

### DIFF
--- a/src/etc/inc/dyndns.class
+++ b/src/etc/inc/dyndns.class
@@ -514,7 +514,7 @@
 					if ($this->_dnsVerboseLog)
 						log_error("HE.net Tunnelbroker: DNS update() starting.");
 					$server = "https://ipv4.tunnelbroker.net/ipv4_end.php?";
-					curl_setopt($ch, CURLOPT_SSL_VERIFYPEER, FALSE);
+					curl_setopt($ch, CURLOPT_SSL_VERIFYPEER, TRUE);
 					curl_setopt($ch, CURLOPT_USERPWD, $this->_dnsUser . ':' . $this->_dnsPass);
 					curl_setopt($ch, CURLOPT_URL, $server . 'tid=' . $this->_dnsHost);
 					break;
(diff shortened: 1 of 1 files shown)


fabianfrz:
Waiting for the LibreSSL patch to merge

fichtner:
Thanks. I forgot to ask which fix, which version of LibreSSL works/will work?

fabianfrz:
The links was: https://github.com/libressl-portable/portable/issues/80

The next version on the repository is 2.5, so this should be the minimum requirement.


Tested it now on my OPNsense device - with `LibreSSL 2.4.5` on `17.1` with the behaviour described in the other pull request. DuckDNS does work (valid certificate)

fabianfrz:
* DHS (https://members.dhs.org) has an expired certificate - so it will fail anyway - keep verify on FALSE
* no-ip (https://dynupdate.no-ip.com/): https://dynupdate.no-ip.com/ - Firefox rejects the connection because of an insecure cipher suite - curl on Linux accepts it - need more testing!
* easyDNS: working on Firefox and curl on Linux, but not on LibreSSL - enable the check.
* hn: VERIFY_PEER is disabled but it is HTTP - WHY???
* opendns: valid -> enable check

fabianfrz:
@fichtner looks like there is a developer package available: https://www.freshports.org/security/libressl-devel/

Can You make a build of libressl 2.5.1 and curl to test if it fixes the issue?

fabianfrz:
* GratisDNS works -> enable TLS verify
* OVH works -> enable TLS verify
* citynetwork works -> enable
* HE.net Tunnelbroker works -> enable

fichtner:
@fabianfrz Thanks! Looks like a lot changed in the last years. I don't mind pulling this into devel. LibreSSL 2.5 will be shipped with OpenBSD 6.1, I don't know when that is so we will simply wait with the final merge into 17.1.x, ok? Unless you have more to clean up :)

fabianfrz:
Hi @fichtner , I do not use them but I think it would be bad to keep this stuff in an insecure way inside.
We can add more but this can be its own pull request as well.

fichtner:
okey, merged. thank you!

fabianfrz:
@fichtner thanks, lets see if it really breaks something - I don't think so - only LibreSSL users which is because of the bug.

fabianfrz:
Let's add a link into the forum as well.

fichtner:
@fabianfrz make sure your patch you advertise on the forum applies to 17.1.2, I don't think it does because google domains is missing

PR #1427, opened 2017-02-25
Further clean-up of generated firewall rules.

phpb-com:
Since pf already provides syntax to refer to interface IPs and directly attached networks, it is redundant to have this in the rule-set. I have not observed any place where those generated tables are used, but maybe missing something. It also does not work for group interfaces.


phpb-com:
This PR removes generation of redundant network aliases for the firewall, and changes how route-to rules for the firewall are generated, making them a little more robust and cleaner. Please consider for the merge, thank you.

AdSchellevis:
@phpb-com thanks! looks good to me.

PR #1428, opened 2017-02-25
Add ruleset-optimization option in firewall settings. 

phpb-com:
This setting will help with de-duplication of rules and reordering for better performance. From man pages:
<pre>
     set ruleset-optimization
           none      Disable the ruleset optimizer.
           basic     Enable basic ruleset optimization.  This is the default
                     behaviour.  Basic ruleset optimization does four things
                     to improve the performance of ruleset evaluations:

                     1.   remove duplicate rules
                     2.   remove rules that are a subset of another rule
                     3.   combine multiple rules into a table when
                          advantageous
                     4.   re-order the rules to improve evaluation performance

           profile   Uses the currently loaded ruleset as a feedback profile
                     to tailor the ordering of quick rules to actual network
                     traffic.

</pre>

### DIFF
--- a/src/etc/inc/filter.inc
+++ b/src/etc/inc/filter.inc
@@ -445,6 +445,12 @@ function filter_configure_sync($verbose = false)
         $limitrules .= "set limit table-entries {$config['system']['maximumtableentries']}\n";
     }
 
+    if (!empty($config['system']['rulesetoptimization'])) {
+        $limitrules .= "set ruleset-optimization {$config['system']['rulesetoptimization']}\n";
+    } else {
+        $limitrules .= "set ruleset-optimization basic\n";
+    }
+
     if ($config['system']['optimization'] <> "") {
         $limitrules .= "set optimization {$config['system']['optimization']}\n";
         if ($config['system']['optimization'] == "conservative") {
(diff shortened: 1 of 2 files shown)


phpb-com:
I'll submit a patch to add GUI option for that today, consider this PR as WIP until then.

phpb-com:
Alright, this one commit should be all that is needed to add ruleset optimization. Since basic is the default, it is the one that is selected and added without user interaction. Please consider this PR for the merge.

src/etc/inc/filter.inc
```diff
@@ -446,6 +446,12 @@ function filter_configure_sync($verbose = false)
         $limitrules .= "set limit table-entries {$config['system']['maximumtableentries']}\n";
     }
 
+    if ($config['system']['rulesoptimization'] <> "") {
```
AdSchellevis: @phpb-com better to check for empty($config['system']['rulesoptimization']) here, rulesoptimization is optional available and may not be initialised.

phpb-com: Thanks, let me rework the diff and submit a new one. I think that I also better rename it to rulesetoptimization for consistency.

AdSchellevis: @phpb-com sounds good, thanks

phpb-com:
Ok, I've updated the diff and changed how the check for config presence is done, as well as the name of the configuration. Thanks for the comments @AdSchellevis

AdSchellevis:
@phpb-com thanks for fixing!

PR #1430, opened 2017-02-26
Fix a203c69 to prevent matching interfaces with 'ip'

phpb-com:
Fix for overaggressive regex in a203c69 . If you create interface with "ip" in the end (group interface "testip") it will wrongly match and cause breakage of rules. This should narrow-down the match and fix the issue.

### DIFF
--- a/src/etc/inc/filter.inc
+++ b/src/etc/inc/filter.inc
@@ -2048,7 +2048,7 @@ function filter_generate_address(&$FilterIflist, &$rule, $target = 'source', $is
         $matches = "";
         if ($network_name == '(self)') {
             $src = "(self)";
-        } elseif (preg_match("/(.*)ip$/", $rule[$target]['network'], $matches)) {
+        } elseif (preg_match("/^(wan|lan|opt[0-9]+)ip$/", $rule[$target]['network'], $matches)) {
             $src = "({$FilterIflist["{$matches[1]}"]['if']})";
         } else {
             $src = "({$FilterIflist[$network_name]['if']}:network)";


phpb-com:
Sorry about the first PR that had wrong hash of the commit in the branch name and commit message, this should be the correct one.

src/etc/inc/filter.inc
```diff
@@ -2048,7 +2048,7 @@ function filter_generate_address(&$FilterIflist, &$rule, $target = 'source', $is
         $matches = "";
         if ($network_name == '(self)') {
             $src = "(self)";
-        } elseif (preg_match("/(.*)ip$/", $rule[$target]['network'], $matches)) {
+        } elseif (preg_match("/(^[wan|lan|opt]+[0-9]*)ip$/", $rule[$target]['network'], $matches)) {
```
fichtner: First one must be round parenthesis, no plus at the end. Lan or wan don't have numbers so shuffle number match into first as suffix to opt with a plus instead of asterisk.

phpb-com:
The first match is used to find interface in config, so I made that all names of different interfaces (e.g., wan, wanip, lan, lanip, opt1, opt1ip) match. I am not sure what you suggest in your comment, could you please give an example? Thanks.

src/etc/inc/filter.inc
fabianfrz: Let's write what @fichtner describes:
```ruby
/(^(wan|lan|opt[0-9]+))ip$/
```

phpb-com: @fabianfrz thanks, that makes sense when you put it this way :) I'll fix the commit soon.

fichtner:
Almost, the outer parenthesis should be removed :)

phpb-com:
Thanks @fichtner and @fabianfrz for helping with regex.

fichtner:
Merged, thank you :)

PR #1438, opened 2017-02-28
Add ability to add custom settings to squid.conf via GUI.

phpb-com:
This will allow addition of custom setting to squid.conf. One use-case is addition of custom refresh_pattern directives to improve caching.


AdSchellevis:
@phpb-com we have a template hook to add custom settings in squid, plaintext input can't be validated, that's why we don't wan't to expose these in the frontend.

phpb-com:
@AdSchellevis I see, sorry that I did not realize this was the case. I will retract this PR then. Thanks.

AdSchellevis:
@phpb-com no problem at all

PR #1444, opened 2017-03-04
translation fix and fix copyright year

fabianfrz:


src/opnsense/mvc/app/views/layouts/default.volt
```diff
@@ -250,7 +250,7 @@
         <!-- page footer -->
     <footer class="page-foot col-sm-push-3 col-lg-push-2">
       <div class="container-fluid">
-        <a target="_blank" href="https://opnsense.org/" class="redlnk">OPNsense</a> (c) 2014-2016 <a href="https://www.deciso.com" class="tblnk">Deciso B.V.</a>
+        <a target="_blank" href="https://opnsense.org/" class="redlnk">OPNsense</a> (c) 2014-2017 <a href="https://www.deciso.com" class="tblnk">Deciso B.V.</a>
```
fichtner: not for now, but we need to softcode this like in the static pages

AdSchellevis: when soft-coding these properties, they probably have to be pushed in somewhere around here (or using setVars in the same file). https://github.com/opnsense/core/blob/17.1.2/src/opnsense/mvc/app/controllers/OPNsense/Base/ControllerBase.php#L219-L222

AdSchellevis:
@fabianfrz looks good, thanks!

src/opnsense/mvc/app/views/layouts/default.volt
fichtner: I will add a json file with the current remaining $g globals and make it a @sample file so it can be overwritten by users

AdSchellevis: @fichtner sure, if you need help anywhere, just let me know.

PR #1453, opened 2017-03-08
make csrf error page translateable, fix HTML errors in csrf.inc

fabianfrz:
* the script tag cannot be between head and body -> move into head
* the input was invalid (looks like a wrongly copied escape sequence
* add gettext


fabianfrz:
Is there any problem because this is still open?

fichtner:
no assignee, sorry for the delay

AdSchellevis:
@fabianfrz sorry for the late response, looks good. Thanks for committing.

Issue #1456, opened 2017-03-09
RFC2136 service always updates IPv4 and IPv6


wevcode:
In OPNsense 17.1.2 you can only update both address types (IPv4 and IPv6) if you want to use RFC2136. Sometimes you do not want to update both address types or you want to split it on two domains. pfSense has integrated this feature with pfsense/pfsense#1578 so it is currently not in OPNsense.

I will try to port it to OPNsense as it is not really complicated.

PR #1457, opened 2017-03-09
RFC2136 RecordType Option

wevcode:
As explained in #1456 this PR makes it possible to control whether A, AAAA or both record types are changed when updating using RFC2136.

### Tasks

- [x] Modify edit page
- [x] Modify service function
- [x] Modify table overview

### Open Questions

_pfSense do not changed their table overview template, so you cannot see which record types are updated. Should we add this in OPNsense as it might be useful because you do not have to open the edit page for each update task?_ (see discussion)

### DIFF
--- a/src/etc/inc/services.inc
+++ b/src/etc/inc/services.inc
@@ -1748,7 +1748,7 @@ EOD;
                 $need_update = false;
 
                 /* Update IPv4 if we have it. */
-                if (is_ipaddrv4($wanip)) {
+                if (is_ipaddrv4($wanip) && $dnsupdate['recordtype'] != "AAAA") {
                     if (($wanip != $cachedipv4) || (($currentTime - $cacheTimev4) > $maxCacheAgeSecs) || $forced) {
                         $upinst .= "update delete {$dnsupdate['host']}. A\n";
                         $upinst .= "update add {$dnsupdate['host']}. {$dnsupdate['ttl']} A {$wanip}\n";
@@ -1764,7 +1764,7 @@ EOD;
                 }
 
                 /* Update IPv6 if we have it. */
-                if (is_ipaddrv6($wanipv6)) {
+                if (is_ipaddrv6($wanipv6) && $dnsupdate['recordtype'] != "A") {
                     if (($wanipv6 != $cachedipv6) || (($currentTime - $cacheTimev6) > $maxCacheAgeSecs) || $forced) {
                         $upinst .= "update delete {$dnsupdate['host']}. AAAA\n";
                         $upinst .= "update add {$dnsupdate['host']}. {$dnsupdate['ttl']} AAAA {$wanipv6}\n";
--- a/src/etc/inc/services.inc
+++ b/src/etc/inc/services.inc
@@ -1748,7 +1748,7 @@ EOD;
                 $need_update = false;
 
                 /* Update IPv4 if we have it. */
-                if (is_ipaddrv4($wanip) && $dnsupdate['recordtype'] != "AAAA") {
+                if (is_ipaddrv4($wanip) && (empty($dnsupdate['recordtype']) || $dnsupdate['recordtype'] == 'A')) {
                     if (($wanip != $cachedipv4) || (($currentTime - $cacheTimev4) > $maxCacheAgeSecs) || $forced) {
                         $upinst .= "update delete {$dnsupdate['host']}. A\n";
                         $upinst .= "update add {$dnsupdate['host']}. {$dnsupdate['ttl']} A {$wanip}\n";
@@ -1764,7 +1764,7 @@ EOD;
                 }
 
                 /* Update IPv6 if we have it. */
-                if (is_ipaddrv6($wanipv6) && $dnsupdate['recordtype'] != "A") {
+                if (is_ipaddrv6($wanipv6) && (empty($dnsupdate['recordtype']) || $dnsupdate['recordtype'] == 'AAAA')) {
                     if (($wanipv6 != $cachedipv6) || (($currentTime - $cacheTimev6) > $maxCacheAgeSecs) || $forced) {
                         $upinst .= "update delete {$dnsupdate['host']}. AAAA\n";
                         $upinst .= "update add {$dnsupdate['host']}. {$dnsupdate['ttl']} AAAA {$wanipv6}\n";
--- a/src/www/services_rfc2136.php
+++ b/src/www/services_rfc2136.php
@@ -140,7 +140,7 @@ $main_buttons = array(
                       <td>
 <?php
                         $filename = "/conf/dyndns_{$rfc2136['interface']}_rfc2136_" . escapeshellarg($rfc2136['host']) . "_{$rfc2136['server']}.cache";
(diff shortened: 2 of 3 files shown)


wevcode:
I will try to change the overview of all RFC2136 update tasks!

src/www/services_rfc2136_edit.php
```diff
@@ -58,6 +58,9 @@
     $pconfig['usetcp'] = isset($a_rfc2136[$id]['usetcp']);
     $pconfig['usepublicip'] = isset($a_rfc2136[$id]['usepublicip']);
 
+    $pconfig['recordtype'] = $a_rfc2136[$id]['recordtype'];
+    if (!$pconfig['recordtype']) $pconfig['recordtype'] = "both";
```
fichtner: For compatibility the record type should remain empty in the case of "both".

wevcode:
One should keep it as simple as possible. In order to follow this idea I have added an extra check so only ip address types which should be updated are displayed in the overview table.

src/www/services_rfc2136_edit.php
```diff
@@ -222,6 +226,14 @@
                     </td>
                   </tr>
                   <tr>
+                    <td valign="top" class="vncellreq"><?=gettext("Record Type");?> </td>
+                    <td class="vtable">
+                      <input name="recordtype" type="radio" value="A" <?php if ($pconfig['recordtype'] == "A") echo "checked=\"checked\""; ?> /> <?=gettext("A (IPv4)");?> &nbsp;
+                      <input name="recordtype" type="radio" value="AAAA" <?php if ($pconfig['recordtype'] == "AAAA") echo "checked=\"checked\""; ?> /> <?=gettext("AAAA (IPv6)");?> &nbsp;
+                      <input name="recordtype" type="radio" value="both" <?php if ($pconfig['recordtype'] == "both") echo "checked=\"checked\""; ?> /> <?=gettext("Both");?>
```
fichtner: value="" and empty($pconfig['recordtype']) is better, also the default selection should be the first in the list.

src/etc/inc/services.inc
```diff
@@ -1748,7 +1748,7 @@ EOD;
                 $need_update = false;
 
                 /* Update IPv4 if we have it. */
-                if (is_ipaddrv4($wanip)) {
+                if (is_ipaddrv4($wanip) && $dnsupdate['recordtype'] != "AAAA") {
```
fichtner: logic works, but may not be resilient against broadening recordtype to other types, maybe:

is_ipaddrv4($wanip) && (empty($dnsupdate['recordtype']) || $dnsupdate['recordtype'] == 'A')

fichtner:
And BTW: thanks for working on this 👍

src/www/services_rfc2136_edit.php
wevcode: Sorry, if I did not get the point but if there is no recordtype set, you are currently updating A and AAAA records which is reflected by "both".

src/www/services_rfc2136_edit.php
wevcode: I will remove the "values", for the second change please read my comment above.

src/etc/inc/services.inc
wevcode: Well I see yout point. Do you think, we should rename "both" to "ALL" which represents all normal DNS records as A, AAAA (and maybe AAAAAAAA in some years)?

Any other record type has to be set by an extra configuration option as MX, right?

src/www/services_rfc2136_edit.php
fichtner: yes. the config semantics are changed here so this imposes a silent migration only when the entry is edited. we cannot guarantee this won't be problematic in the future so we need to keep the empty value as default. :)

wevcode:
OpenSource: Everyone can use it, but you should also try to give something back! :+1:

src/etc/inc/services.inc
fichtner: That's the idea, exactly. And indeed "All" sounds best for future use.

src/www/services_rfc2136_edit.php
wevcode: Ok, I will change it! :)

fichtner: Cool. If you need anything or have questions I'm happy to help. :)

wevcode: @fichtner I want to test my changes but there is a problem: opnsense/tools#46

fichtner: From a working OPNsense VM you can install the repository and upgrade from there:

    # opnsense-code -a wevcode opnsense-core
    # cd /usr/opnsense-core
    # git checkout rfc2136-recordtype-option
    # opnsense-update -t opnsense-devel
    # make upgrade

wevcode: @fichtner I cannot find this in the tools repository as a documentation, please add this tip for other contributors.

fichtner: Most of it is in the handbook: https://docs.opnsense.org/development/workflow.html

But it needs some updates on the refined tools :)

wevcode:
@fichtner Updated branch; I will test it now. :checkered_flag:

src/www/services_rfc2136_edit.php
```diff
@@ -58,6 +58,8 @@
     $pconfig['usetcp'] = isset($a_rfc2136[$id]['usetcp']);
     $pconfig['usepublicip'] = isset($a_rfc2136[$id]['usepublicip']);
 
+    $pconfig['recordtype'] = $a_rfc2136[$id]['recordtype'];
```
fichtner: This needs something like:

$pconfig['server'] = isset($id) &&!empty($a_rfc2136[$id]['server']) ? $a_rfc2136[$id]['server'] : null;

src/www/services_rfc2136.php
```diff
@@ -140,7 +140,7 @@
                       <td>
 <?php
                         $filename = "/conf/dyndns_{$rfc2136['interface']}_rfc2136_" . escapeshellarg($rfc2136['host']) . "_{$rfc2136['server']}.cache";
-                        if (file_exists($filename) && !empty($rfc2136['enable'])) {
+                        if (file_exists($filename) && !empty($rfc2136['enable']) && $rfc2136['recordtype'] != "AAAA") {
```
fichtner: same logic as in services.inc should be used here. We cannot expect it to be set, it would probably produce a warning now.

wevcode: Missed that, sorry.

src/www/services_rfc2136_edit.php
wevcode: You are right, I will change it!

wevcode:
@fichtner Should be fixed! :)

wevcode:
@fichtner Also added a help text to undestand the new 'all' option.

wevcode:
***Ready for review.***

fichtner:
Merged, thanks!

fichtner:
Two small follow-ups committed. I think this will land in 17.1.3 next week. Thanks again.

Issue #1431, opened 2017-02-26
Feature: hardware off-loading control per interface


phpb-com:
It would be good to have per interface control of the hardware off-loading. Since most of the people will likely run IPS on wan, it is not necessary to turn-off off-loading on all of the interfaces. I would like to have it still "on" on the local internal interfaces. What do you think?

phpb-com:
I think that adding "Network Interfaces Hardware Control" section in Interface > [Interface] will be the most intuitive, while retaining global settings for off-loading in the same place. If there are no objections I would like to take a shot at adding that.

AdSchellevis:
I'm not entirely sure it's a good idea to change this to a per interface option, because we've seen more issues with offloading in the past (strange link issues etc.).
Do you have a specific use-case where offloading really helps? 
It might be an option to support an override on a per interface basis, as long as we have defaults for all like it is now (I really want to keep our defaults the way they are now).

phpb-com:
That is my idea to have global off-loading control as is and only add per interface override. I do also think that keeping global default as is, is important.

PR #1463, opened 2017-03-10
Implement hardware offloading override

phpb-com:
Implement hardware offloading override for selected interfaces. Allow to have global flag to turn-off all overrides. Also clean-up how hardware offloading is set/unset, since flags are idempotent.  #1431

### DIFF
--- a/src/etc/inc/interfaces.lib.inc
+++ b/src/etc/inc/interfaces.lib.inc
@@ -297,38 +297,36 @@ function legacy_interface_details($intf)
 function configure_interface_hardware($ifs)
 {
     global $config;
-    $intf_details = legacy_interface_details($ifs);
     /* skip vlans for checksumming  */
-    if (!stristr($ifs, "_vlan") && is_array($intf_details)) {
-        // get current settings
-        $csum_set = in_array('rxcsum', $intf_details['options']) || in_array('txcsum', $intf_details['options']);
-        $csumv6_set = in_array('rxcsum6', $intf_details['options']) || in_array('txcsum6', $intf_details['options']);
-        $tso_set = in_array('tso4', $intf_details['options']) || in_array('tso6', $intf_details['options']);
-        $lro_set = in_array('lro', $intf_details['options']);
+    if (!stristr($ifs, "_vlan")) {
+
+        //Get current override settings
+        if(isset($config['system']['offloadoverride'])) {
+            $overrideifs = isset($config['system']['overridemembers']) ? explode(' ', $config['system']['overridemembers']) : array();
+        } else {
+            $overrideifs = array();
+        }
+
+	$override = in_array($ifs, $overrideifs);
 
         // hardware checksum offloading offloading
-        if (isset($config['system']['disablechecksumoffloading']) && $csum_set) {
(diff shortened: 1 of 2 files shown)


src/etc/inc/interfaces.lib.inc
```diff
@@ -297,38 +297,36 @@ function legacy_interface_details($intf)
...
-        if (isset($config['system']['disablechecksumoffloading']) && $csumv6_set) {
-            legacy_interface_flags($ifs, '-txcsum6 -rxcsum6', false);
-        } elseif (!isset($config['system']['disablechecksumoffloading']) && !$csumv6_set) {
-            legacy_interface_flags($ifs, 'txcsum6 rxcsum6', false);
+
+	$override = in_array($ifs, $overrideifs);
+
+        // hardware checksum offloading offloading
+        if (isset($config['system']['disablechecksumoffloading']) && !$override) {
+            legacy_interface_flags($ifs, '-txcsum -rxcsum -txcsum6 -rxcsum6', false);
```
fichtner: we should merge this before we merge the feature

src/etc/inc/interfaces.lib.inc
```diff
@@ -297,38 +297,36 @@ function legacy_interface_details($intf)
...
+            legacy_interface_flags($ifs, 'txcsum rxcsum txcsum6 rxcsum6', false);
         }
 
         // TCP segmentation offloading
-        if (isset($config['system']['disablesegmentationoffloading']) && $tso_set) {
-            legacy_interface_flags($ifs, '-tso', false);
-        } elseif (!isset($config['system']['disablesegmentationoffloading']) && !$tso_set) {
-            legacy_interface_flags($ifs, 'tso', false);
+        if (isset($config['system']['disablesegmentationoffloading']) && !$override) {
+            legacy_interface_flags($ifs, '-tso -tso4 -tso6', false);
```
fichtner: also merge this earlier here (if it makes sense, but tso may include tso4 and tso6?)

src/etc/inc/interfaces.lib.inc
```diff
@@ -297,38 +297,36 @@ function legacy_interface_details($intf)
...
+            $overrideifs = isset($config['system']['overridemembers']) ? explode(' ', $config['system']['overridemembers']) : array();
+        } else {
+            $overrideifs = array();
         }
-        if (isset($config['system']['disablechecksumoffloading']) && $csumv6_set) {
-            legacy_interface_flags($ifs, '-txcsum6 -rxcsum6', false);
-        } elseif (!isset($config['system']['disablechecksumoffloading']) && !$csumv6_set) {
-            legacy_interface_flags($ifs, 'txcsum6 rxcsum6', false);
+
+	$override = in_array($ifs, $overrideifs);
```
fichtner: I think this merges logic for checkums and tso and lro, we shouldn't want that if we do it right

phpb-com:
@fichtner thanks for review. I agree on all counts, will first retract the interface.lib.inc changes and submit as a separate PR. Then will change the override itself to per option for each interface. Thanks.

phpb-com:
I will close this PR for now, since there is a need to fix the way flags are assigned to the interface, and the GUI in front of it.

Issue #1462, opened 2017-03-10
Feature: Traffic Graph IP table improvements


jeffg2k:
The IP table that lists the current traffic is too fleeting with IPs popping in and out, trying to spot who is consuming what is difficult.

Instead of dropping off the IPs as they flow in and out of traffic, let them persist for the duration of the overall traffic graph (showing maybe the top xx), dynamically updating the traffic for each.  You could add small graphs to each row for the individual usage (like the size of CPU usage in the dashboard).  It would also be great to have a column that displays the hostname.  Would make a great dashboard widget too.

This would be so much more effective when trying to look at the traffic graph.  It's great to see the overall traffic, but usually I'm looking at it to see who is consuming what in real time.

fichtner:
This was merged recently, thanks again!

Issue #1449, opened 2017-03-06
[BUG] DNS Resolver seems to be not started after boot


fabianfrz:
After booting the DNS-Resolver seems to be not starting on `17.1.2`
It runs after
* Starting it via the services page
* Saving the configuration

fabianfrz:
I checked the log, which says the following:
```
Mar  9 14:27:24 opn unbound: [24343:0] error: cannot chdir to directory:  (No such file or directory)
```

So it may be a race condition in the boot process.

fabianfrz:
Now after booting, this one shows up in the log:
```
opnsense: /usr/local/etc/rc.bootup: The command 'chroot -u unbound -g unbound / '/usr/local/sbin/unbound-anchor' -a '/var/unbound/root.key'' returned exit code '1', the output was ''
```

fichtner:
Can you give more context than the error itself? There is some kind of ordering issue, but I have no idea where and cannot reproduce.

fabianfrz:
It happens when the device is booting. The service can be started later without issues. I think this is because the service is started before the configuration is generated

fichtner:
Yes, but with some more log messages we can find the call spot. ;)

fabianfrz:
sure:
```
Mar 11 17:27:42 	opnsense: /index.php: Successful login for user 'root' from: 10.230.231.9
Mar 11 17:26:46 	kernel:
Mar 11 17:26:46 	kernel:
Mar 11 17:26:45 	kernel:
Mar 11 17:26:33 	squid[36296]: Squid Parent: (squid-1) process 36475 started
Mar 11 17:26:33 	squid[36296]: Squid Parent: will start 1 kids
Mar 11 17:26:23 	squid: No valid signing SSL certificate configured for HTTPS_port 127.0.0.1:3129
Mar 11 17:26:13 	ftp-proxy[23677]: listening on 127.0.0.1 port 8021
Mar 11 17:26:08 	configd.py: generate template container OPNsense/Syslog
Mar 11 17:26:08 	configd.py: generate template container OPNsense/Sample/sub2
Mar 11 17:26:08 	configd.py: generate template container OPNsense/Sample/sub1
Mar 11 17:26:07 	configd.py: generate template container OPNsense/Sample
Mar 11 17:26:06 	configd.py: generate template container OPNsense/Quagga
Mar 11 17:26:03 	configd.py: generate template container OPNsense/Proxy
[...]

Issue #1465, opened 2017-03-10
DHCP static mapping not listing IP


jeffg2k:
I find the viewing of static mapped DHCP addresses (IP reservations) confusing.

Feedback - Suggestions
1) On the Server page it lists the "DHCP Static Mappings for this interface."  However, it doesn't list the IP addresses assigned to those reservations.  I haven't turned on Static ARP, so maybe it would display it there, but it just seems weird that the IP reservation doesn't show the IP reserved.  As a interface placement, I also wonder at the point of having it at the bottom of the server tab.  To me it might make more sense to just have it all on the Leases page - add a LAN / WAN tab there if needed.

2) On the Leases page we have the same issue as above where the static entries don't contain their static reservations.  The entries also do not have edit or delete buttons, where the Server page does.  In addition, it doesn't include the lease period.  As far as I know, an IP reservation (static mapping) is still a lease - it's just the system will not give that IP out to anyone else.  And it does show this information, above in the table.  It has two entries in the table, a static one and an active one.  I find this confusing.  Just have it once, note it as reserved along with the IP assigned and the lease period, along with the static description.

    - The static hostname lists what the admin defines and the active hostname lists what the computer reports.  If you give the option to edit the hostname, then it should probably reflect that, unless you want to give an option to add an alias that doesn't effect the name reported.  Sometimes I'll add a prefix to the hostname so I can identify who it belongs to, where I use the description to define where it is located.

djGrrr:
Have you actually configured IP addresses on these leases? IP addresses show up in both places for me, and there are no duplicate leases for an IP with static reservation, only the static ones get listed. Can you please include some screenshots so we can see what you are seeing?

What version of OPNsense are you running?

jeffg2k:
I'll have to wait until Monday to provide a screenshot, but no, I have not configured IP addresses on the leases.  If I recall, it states that if you specify an ip, the ip must be outside the DHCP scope, which seemed odd.  If you leave it blank, it will use an IP from the DHCP pool, which is usually the one currently assigned to it.  I want to reserve its current IP, which is in the pool.  I'm running the latest version of OPNsense, 17.1.

djGrrr:
The IP is not static unless you put it in the reservation, it just allows you to do dynamic addresses with "allow only known clients" enabled.

I would suggest making the dynamic range a subset of the entire range, then assigning static ips in the rest of the range.

jeffg2k:
[...]

Issue #1013, opened 2016-06-14
The script in core/src/opnsense/www/js/d3.min.js has stopped working (Firefox 47.0)


Swoy:
Leaving a tab with Opnsense on dashboard with the following widgets:

System information
Gateways
Traffic Graphs
Service Status
Interfacelist
Thermal Sensors
Interface Statistics
System Log
Firewall Logs

causes script hang on d3.min.js after a random amount of time. If the user allows the script to continue, the memory is leaking up to several GB for that tab.

Please let me know what additional information you require to locate the issue.

AdSchellevis:
timeout

PR #1472, opened 2017-03-14
Traffic Graph improvements

jeffg2k:
This update includes many of the items discussed in feature request https://github.com/opnsense/core/issues/1462.  

This allows the hosts entries to persist for the duration of the overall traffic graph (showing the top 20), dynamically updating the traffic for each. It includes a small micrograph for each item displaying "bw in" and "bw out".  It also includes new columns for Total In, Total Out, as well as new sorting for Total In, Total Out, Bw In Avg, Bw Out Avg.  The Avg sorting helps reduce fluctuation in the table.

This pull request is based on code from the last commit of Mar 13, 2017. https://github.com/opnsense/core/issues/1449

### DIFF
--- a/src/www/status_graph.php
+++ b/src/www/status_graph.php
@@ -74,7 +74,7 @@ if ($_SERVER['REQUEST_METHOD'] === 'GET') {
             $intsubnet = gen_subnet(find_interface_ip($real_interface), $netmask) . "/$netmask";
             $cmd_args = $pconfig['filter'] == "local" ? " -c " . $intsubnet . " " : " -lc 0.0.0.0/0 ";
             $cmd_args .= $pconfig['sort'] == "out" ? " -T " : " -R ";
-            $cmd_action = "/usr/local/bin/rate -i {$real_interface} -nlq 1 -Aba 20 {$cmd_args} | tr \"|\" \" \" | awk '{ printf \"%s:%s:%s:%s:%s\\n\", $1,  $2,  $4,  $6,  $8 }'";
+            $cmd_action = "/usr/local/bin/rate -v -i {$real_interface} -nlq 1 -Aba 20 {$cmd_args} | tr \"|\" \" \" | awk '{ printf \"%s:%s:%s:%s:%s\\n\", $1,  $2,  $4,  $6,  $8 }'";
             exec($cmd_action, $listedIPs);
             for ($idx = 2 ; $idx < count($listedIPs) ; ++$idx) {
                 $fields = explode(':', $listedIPs[$idx]);
@@ -104,9 +104,64 @@ include("head.inc");
 ?>
 <body>
 <?php include("fbegin.inc"); ?>
-
+<style>
+path {
+    stroke: steelblue; 
+    stroke-width: 1;
+    fill: none;
+}
+.minigraph {
+	border-style: solid;
+	border-color: lightgray;
+	border-width: 0px 1px 1px 1px;
+	padding: 0px;
+	margin: 0px;
+	margin-bottom: -5px;
+}
+</style>
 <script type="text/javascript">
-  $( document ).ready(function() {
+    var graphtable = {};
+    function formatSizeUnits(bytes){
+        if      (bytes>=1000000000) {bytes=(bytes/1000000000).toFixed(2)+'G';}
+        else if (bytes>=1000000)    {bytes=(bytes/1000000).toFixed(2)+'M';}
+        else if (bytes>=1000)       {bytes=(bytes/1000).toFixed(2)+'k';}
+        else if (bytes>1)           {bytes=bytes.toFixed(2) +'b';}
+        else if (bytes==1)          {bytes=bytes.toFixed(2) +'b';}
+        else                        {bytes='0.00b';}
+        return bytes;
+    }
+    function containsHost(obj, list) {
+        var i;
+        for (i = 0; i < list.length; i++) {
+            if (list[i].host === obj.host) {
+                return true;
+            }
+        }
+        return false;
+    }
+    function sortProperties(obj, item)
+	{
+	  // convert object into array
+		var sortable=[];
+		for(var key in obj)
+			if(obj.hasOwnProperty(key))
+				sortable.push([key, obj[key]][item]); // each item is an array in format [key, value]
+
+		// sort items by value
+		sortable.sort(function(a, b)
+		{
+		  return a[1]-b[1]; // compare numbers
+		});
+		return sortable; // array in format [ [ key1, val1 ], [ key2, val2 ], ... ]
+	}
+    // define dimensions of graph
+	var m = [3, 3, 1, 3]; // margins
+	var w = 150 - m[1] - m[3]; // width
+	var h = 30 - m[0] - m[2]; // height
+	var datasize = 45;
+	var maxvalue = 0;
+
+  	$( document ).ready(function() {
       function update_bandwidth_stats() {
         $.ajax("status_graph.php", {'type': 'get', 'cache': false, 'dataType': 'json', 'data': {'act': 'traffic'}}).done(function(data){
             traffic_widget_update($("[data-plugin=traffic]")[0], data);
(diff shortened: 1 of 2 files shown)


jeffg2k:
snapshot

jeffg2k:
One thing that I haven't done yet with this is better Hostname Display for LAN addresses by checking against the DHCP lease files.  I'm looking at the status_dhcp_leases.php, but not sure if there is an easier way to get the data.

fichtner:
Hi Jeff,

Thanks, this is looking good!

I will assign it to @AdSchellevis who originally converted this page for review/merge. In the meantime, you may consider adding your copyright to src/www/status_graph.php

Ad: please use "squash and merge" for merging if you can.


Cheers,
Franco

jeffg2k:
I'm testing performance on the page and the Traffic Widget has a memory leak.  Doesn't appear to be javascript, maybe detached DOM tree?  Watching it now, my page consumption is at 3.3Gig and growing.  If I remove the Widget and just display the IP table of the top 30, it only consumes around 250mb.

fichtner:
Same as this maybe? https://github.com/opnsense/core/issues/1013

jeffg2k:
That seems likely to me.  I fixed it with the latest commit.

jeffg2k:
There is also an exception I see periodically in the Traffic Widget.  It's not causing much issue as it just continues on, but I thought I would mention it.

Uncaught TypeError: Cannot read property 'x' of null
    at tr (d3.min.js:2)
    at Ze (d3.min.js:2)
    at ar (d3.min.js:2)
    at t (d3.min.js:4)
    at Object.ao.geom.voronoi (d3.min.js:4)
    at i (nv.d3.min.js:8)

That error is described in this StackOverflow post, where an answer was that the nvd3.js was out of sync with d3.js.  http://stackoverflow.com/questions/17624111/nvd3-line-chart-uncaught-typeerror-cannot-read-property-x-of-undefined

jeffg2k:
Upon further testing - the memory leak remains.  :(

jeffg2k:
I'm thinking the leak in the Traffic Widget may be in nv.d3.  It seems to happen in the "// load data" phase.

fichtner:
I think this will corroborates your findings:

https://www.reddit.com/r/PFSENSE/comments/5v8vjm/pfsense_233_release_now_available/de0y2ig/?context=3&st=j0bz9aor&sh=06d26d46

jeffg2k:
Yeah, looking at this issue for it @ nvd3 https://github.com/novus/nvd3/issues/1956  However, I don't see that their fix has fixed it.  I tried PFSense's modified nvd3 and the latest nvd3 1.8.5.  Here is an example of the problem which I think recreates what we're seeing in OPNsense. http://jsfiddle.net/gzzdmz7j/

jeffg2k:
Created a new issue for it: https://github.com/novus/nvd3/issues/1992

src/www/status_graph.php
```diff
@@ -48,6 +50,24 @@
     }
 }
 
+//Create array of hostnames from DHCP
+foreach ($interfaces as $ifname => $ifarr) {
+	if (isset($config['dhcpd'][$ifname]['staticmap'])) {
```
AdSchellevis: Hi Jeff, this could be simplified by adding an additional loop and reuse the the logic underneath, like:
```

foreach (array('dhcpd', 'dhcpdv6') as $dhcp) {
    foreach($config[$dhcp][$ifname]['staticmap'] as $entry) {
        ...
    }
}
```

src/www/status_graph.php
```diff
@@ -48,6 +50,24 @@
...
+		foreach($config['dhcpd'][$ifname]['staticmap'] as $entry) {
+			if (htmlentities($entry['hostname']) !== "") {
```
AdSchellevis: probably better to check !empty($entry['hostname']) and leave the entities in the line below.

jeffg2k:
Updated code as suggested - we still need something similar for the active leases.
I was hoping something like 'activemap' might exist, similar to 'staticmap' in the config. But that term didn't work.

Update on the Memory leak - I found that I only get the leak on Chrome while running the LastPass extension.  My troubleshooting indicates it's not anything directly in their code, but more with just loading the content script and perhaps not releasing elements of the dom tree.  So I'm not sure if this is a nv.d3 issue, LP issue, or Chrome issue.  Some magic combination.

AdSchellevis:
Hi Jeff, the config indeed doesn't know anything about the service status.
For dhcp the status is kept in two files, /var/dhcpd/var/db/dhcpd.leases and /var/db/dhcpd6.leases, but to prevent adding a lot of additional glue in the ui to parse those, we better add a configd system call to request the known leases using a single call.
configd is our new system to control services and request status information about them, which all new code should use (it's a requirement to eventually be able to stop using root for the UI)

If you like, I can add a service section for dhcp and add some simple calls for you to use, which would then work something like this:

`$leases = json_decode(configd_run("dhcpd show leases"), true);
`

And then delivers a lists of all known leases.

Maybe with some additional parameters to filter the results if needed (like only active).

Best regards,

Ad

jeffg2k:
That sounds great Ad.  Please do.
If you're reworking how DHCP queried and displayed, please take a look at this request as well https://github.com/opnsense/core/issues/1465  It was one area in the UI that I found a bit confusing.  If they're both pulling from the same config, it will be easier to display a description and hostname change of active leases if defined and details for static leases.

AdSchellevis:
@jeffg2k I finally found some time to add the configd command to list the leases (https://github.com/opnsense/core/commit/07f5033bff78c54874e92724c7dbf9e54f2fada7)

```
// list active leases
$leases = json_decode(configd_run("dhcpd list leases"), true);

// list all leases
$leases = json_decode(configd_run("dhcpd list leases"), true);
```

To test from the command line:

`configctl dhcpd list leases 1
`

Best regards,

Ad

jeffg2k:
@AdSchellevis  Excellent - I'll try to copy those files to my test instance and give it a run but feel free to update it if you have a chance.  I've been a bit busy and haven't had time lately.  

The examples given for the active leases and all leases are identical.  Should there be a number 0 or 1 in there?

AdSchellevis:
@jeffg2k yes, my mistake, 0/1 to switch active/all leases. I've changed the example as well.

sjorge:
@AdSchellevis was looking into this code for another issue stumbled in here. This is my IPv6 leases file.

```
# The format of this file is documented in the dhcpd.leases(5) manual page.
# This lease file was written by isc-dhcp-4.3.5

# authoring-byte-order entry is generated, DO NOT DELETE
authoring-byte-order little-endian;

server-duid "\000\001\000\001 \227J\332\000\"\006\377\020\001";

ia-na "\000\000\000\000\000\001\000\001\035-cU\240\231\233\017\002\013" {
  cltt 5 2017/05/05 20:44:47;
  iaaddr 2001:470:7ee7:10::241 {
    binding state active;
    preferred-life 2250;
    max-life 3600;
    ends 5 2017/05/05 21:44:47;
  }
}

ia-na "\001\000\000\000\000\001\000\001\035-cU\240\231\233\017\002\013" {
  cltt 5 2017/05/05 14:55:11;
  iaaddr 2001:470:7ee7:20::250 {
    binding state active;
    preferred-life 377962;
    max-life 604740;
    ends 5 2017/05/12 14:54:11;
  }
}
```

AdSchellevis:
@sjorge thanks!

fichtner:
@jeffg2k @AdSchellevis @sjorge  stable/17.7 is now branched so this won't hit 17.7 right now... should we merge this for later inclusion? I like this work, not sure if done

sjorge:
This can always go in 18.1 right? Or a later 17.7.x once it works well?

fichtner:
yes

rate utility to gather statistics does not support ipv6

sjorge:
I believe part of the lease lookup also needs also needed  get support added, @AdSchellevis was worked on that?

But if it is IPv4 only with no IPv6 support in sight I see no problems with getting this in 17.7.

AdSchellevis:
I'm currently not working on this topic, as far as I know there's also still a memory leak issue which prevents it from being stable enough for production.

fichtner:
the memory leak is in the library, it has been reported multiple times and can be reproduces with leaving the dashboard + traffic graph widget open ;)

AdSchellevis:
yes, I know,  in the library. Sorry if it looks like I was suggesting something else here.

fichtner:
merged, thanks!

fichtner:
FWIW: colors are gone, traffic widget doesn't update. Will take a closer look...

AdSchellevis:
@fichtner are you sure we should have pulled this in?

fichtner:
I'll keep this on -devel, work on it...

knxuf:
can you enable the settings from the traffic graph dashboard applet to hide interfaces

PR #1498, opened 2017-03-22
harden ciphers and tls versions

fabianfrz:
Note: SSL cannot be disabled because it does not exist (at least not in LibreSSL)

Important Note: This directive must be updated when upgrading to Squid 4.

### DIFF
--- a/src/opnsense/service/templates/OPNsense/Proxy/squid.conf
+++ b/src/opnsense/service/templates/OPNsense/Proxy/squid.conf
@@ -51,6 +51,9 @@
 sslcrtd_program /usr/local/libexec/squid/ssl_crtd -s /var/squid/ssl_crtd -M {{ OPNsense.proxy.forward.ssl_crtd_storage_max_size|default('4') }}MB
 sslcrtd_children {{ OPNsense.proxy.forward.sslcrtd_children|default('5') }}
 
+sslproxy_cipher HIGH:MEDIUM:!RC4:!aNULL:!eNULL:!LOW:!3DES:!MD5:!EXP:!PSK:!SRP:!DSS
+sslproxy_options NO_TLSv1
+
 # setup ssl bump acl's
 acl bump_step1 at_step SslBump1
 acl bump_step2 at_step SslBump2


fabianfrz:
@fichtner can you check this one?

fichtner:
Ad is in tomorrow, I'd prefer his judgement

AdSchellevis:
@fabianfrz it looks good to me, if people complain we might consider adding an option "allow insecure" or something like it, but let's wait for feedback first. When we're switching to squid 4, probably more directives will change, so that shouldn't be an issue either. 

I'm pulling this one in, thanks again!

fichtner:
I'll skip this for 17.1.4, I don't want it in the images just yet for compatibility reasons.

fabianfrz:
@fichtner that is a really BAD idea. This allows squid to use weak ciphers like RC4 so if interception is enables the client has a secure connection to squid but squid can have in insecure connection to the destination host.

AdSchellevis:
@fichtner @fabianfrz when we're delivering new images it's better to keep the current known working situation.

fichtner:
Fair enough, but then I'm still not considering this without the ability to turn it off for compatibility reasons. The risk of breaking existing setups is moderately high with this and we're in a minor iteration with no way to turn it back off.

fichtner:
@fabianfrz if we are worried about e.g. RC4, shouldn't we put a smaller impact one on stable/17.1 for 17.1.4?

fabianfrz:
@fichtner may be an idea to block only 3DES and RC4 on 17.1.4. It would not have the risk to impact so many sites. Browsers, for example Firefox, block RC4 anyway so it should not hurt. Most servers use AES.

fichtner:
@fabianfrz what should I add in terms of lines?

fabianfrz:
you could omit the second line which disables TLS 1.0
and you can omit ":!LOW" and ":!PSK:!SRP:!DSS" if you like.

fabianfrz:
however the next update should contain the full list of disabled ciphers.

fabianfrz:
You can check it here: https://badssl.com/

fichtner:
@fabianfrz @AdSchellevis like this? https://github.com/opnsense/core/commit/d4ac31717a

fabianfrz:
should work. If not openssl has a constant ALL, which needs to be added

fichtner:
worked on openssl and libressl from here, thanks!

Issue #1491, opened 2017-03-20
dnsmasq: deprecate default in favour of a full resolver dnssec replacement


fichtner:
* ~~Includes additions to fully unwind the DNS subsystem with the plugin framework~~
* ~~Unbound default config as full resolver~~
* ~~Unhook Dnsmasq from resolver logs page~~
* ~~Add log file for dnsmasq and/or make it pluggable for syslog plugin foo~~
* ~~Unbound menu "Unbound DNS"~~
* Still need to wire DNS remote log output https://github.com/opnsense/core/blob/master/src/etc/inc/system.inc#L838
* Slow FS may prevent Unbound start or crash? #1449 #1515

PR #1504, opened 2017-03-25
move from plugins: unbound debugging

fabianfrz:


fichtner:
thanks, let me know when you want this pulled in. there is alignment work for #1491 that will likely shift / enhance these files as discussed.

fabianfrz:
@fichtner this should be it for now, have fun.

fichtner:
merged, thanks!

AdSchellevis:
@fabianfrz @fichtner are we sure we want ruby as a core dependency? Personally I would rather stick to python or php, given the fact that both are used for all the other components.

fichtner:
my task https://github.com/opnsense/core/issues/1505 as discussed with Fabian

AdSchellevis:
Thanks! If you need help, just let me know.

fabianfrz:
@AdSchellevis Franco wanted that to be moved to core from here: https://github.com/opnsense/plugins/pull/101

AdSchellevis:
@fabianfrz I just missed the comment, sorry about that, I agree it's a good idea to move it to core.

fabianfrz:
@AdSchellevis no problem. Nobody will see it anyway as there is currently no GUI for it ;)
It is made to make it possible to generate some diagnostics pages for unbound.

AdSchellevis:
@fabianfrz nice!

PR #1524, opened 2017-04-06
Fixed bug which caused the some cache files not being created

theq86:
Hello, since opnSense is a fork of pfSense I found a solution for a problem I had with opnSense in the pfSense bugtracker. See: https://redmine.pfsense.org/issues/3669

They state, that there are 2 bugs which cause the cache files /var/db/wan_cacheip and /var/db/wan_cacheipv6 not being created.

One of the bugs is already fixed in pfSense as well as opnSense.
The other bug concerning IPv6 is fixed by my change and working for me.

For me, the bug caused OpenVPN restart every 15 minutes, when WAN has IPv6 enabled. Now it works.

However, I do not know the whole opnSense and can not say if this fix would have other side-effects.

### DIFF
--- a/src/etc/rc.newwanipv6
+++ b/src/etc/rc.newwanipv6
@@ -98,7 +98,7 @@ if (is_ipaddrv6($curwanipv6)) {
 
 log_error("rc.newwanipv6: on (IP address: {$curwanipv6}) (interface: {$interface}) (real interface: {$interface_real}).");
 
-$oldipv6 = "";
+$oldipv6 = "::";
 if (file_exists("/var/db/{$interface}_cacheipv6")) {
     $oldipv6 = file_get_contents("/var/db/{$interface}_cacheipv6");
 }


fichtner:
Hi @theq86,

Thanks! There is indeed something wrong here, but it looks like we should get rid of the cacheipv6 file trickery and align the code a bit more. IPv4 doesn't use this workaround so IPv6 should be the same.

I'll see what I can do and report back. :)


Cheers,
Franco

fichtner:
I've added https://github.com/opnsense/core/commit/f729d881c749db5f146f3dad9e331a3bd97fe0ab but won't be able to test this before tomorrow.

theq86:
Thanks. I will apply the changes live on my opnSense and test as well.

fichtner:
There are some devel changes, I don't think it's fully compatible with 17.1.4 at this point.  Be careful 😊

theq86:
You're right. I reverted to my temporary fix. It works and it will do until your changes find the way into the release.

Woi:
@fichtner Will this fix make it to the 17.1.5 release?

fichtner:
opnsense-devel on 17.1.5 will have it. If someone could test it would speed up inclusion. When 17.1.5 is out I can also provide a clean patch for this change only if something else on -devel doesn't work, which could be, but normally it's fine too. 😊

Woi:
If I run `pkg install -y opnsense-devel` as described in https://docs.opnsense.org/contribute.html#testing do I end up with an pre-release of 17.1.5 or an development version of 17.7?

fichtner:
# opnsense-update -t opnsense-devel

It'll install the latest 17.7 prerelease GUI package, kernel and base are the same as the 17.1 release. -devel is fixed build, it currently does not include these changes, but will once 17.1.5 is out.

With git you can get the latest -devel changes afterwards anyway:

# opnsense-code core
# cd /usr/core
# make upgrade

Woi:
I'll try to test the fix, but I don't expect to find time for this during the next two weeks.

fichtner:
The backported patch for 17.1.5:

# opnsense-patch f31964a

(run once to apply, run again to remove)

theq86:
Since this pull request will never make it as it is we should reject and close it?

fabianfrz:
@theq86 a pull request cannot be converted to a ticket.

theq86:
Okay, then I close it because the issue presented in this pull request is going to be resolved by other commits.

Woi:
I'll give feedback https://github.com/opnsense/core/commit/f31964a

fichtner:
@theq86 your patch will go into 17.1.7 via https://github.com/opnsense/core/commit/805cac40 -- thanks!

we're reworking and unifying the IP renewal for IPv4 and IPv6 completely for 17.7, so this buys some time. The code hasn't been dusted off in a very long time. :)

PR #1533, opened 2017-04-10
Setting locale in ApiControllerBase

kekek2:


src/opnsense/mvc/app/controllers/OPNsense/Base/ControllerBase.php
```diff
@@ -44,32 +44,15 @@ class ControllerBase extends ControllerRoot
...
-
-        $lang_encoding = $lang . '.UTF-8';
-        $textdomain = 'OPNsense';
-
         $ret = new ViewTranslator(array(
             'directory' => '/usr/local/share/locale',
-            'defaultDomain' => $textdomain,
-            'locale' => $lang_encoding,
+            'defaultDomain' => 'OPNsense',
+            'locale' => "en_US.UTF-8",
```
AdSchellevis: I don't think it's a good idea to feed in a static locale here, according to the docs (https://docs.phalconphp.com/en/latest/api/Phalcon_Translate_Adapter_Gettext.html) Gettext() expects a proper one here.

src/opnsense/mvc/app/controllers/OPNsense/Base/ApiControllerBase.php
```diff
@@ -91,6 +91,7 @@ public function initialize()
     {
         // disable view processing
         set_error_handler(array($this, 'APIErrorHandler'));
+        self::setLocale();
```
AdSchellevis: Eventually it might be better to move both setLocales() to ControllerRoot, for now this is good enough.

src/opnsense/mvc/app/controllers/OPNsense/Base/ControllerRoot.php
```diff
@@ -48,6 +49,31 @@ protected function sessionClose()
     }
 
     /**
+     * Set locale for gettext
+     */
+    public static function setLocale()
```
AdSchellevis: if we skip the static here, we could save our locale in the controller object and add a getLocale() to request our setting back (which can then be used in getTranslator() for GetText() ).

src/opnsense/mvc/app/controllers/OPNsense/Base/ControllerBase.php
kekek2: https://github.com/phalcon/cphalcon/blob/master/phalcon/translate/adapter/gettext.zep#L198
don't setting locale under OPNSense, because setlocale() return FALSE. We can pass any value of 'locale' to ViewTranslator(). The effective setting 'locale' is below.

src/opnsense/mvc/app/controllers/OPNsense/Base/ControllerRoot.php
kekek2: I call \OPNsense\Base\ControllerRoot::setLocale() from /usr/local/opnsense/scripts/*

kekek2:
I remove setting static locale.

AdSchellevis:
@kekek2 thanks!

PR #1535, opened 2017-04-10
Fix autocomplete, autocapitalize and autocorrect in Captive Portal login form

johannrichard:
New iOS versions do autocapitalization, autocrrection and/or autocompletion in Web Forms. All three can be annoying when entering always changing usernames with sometimes lower-case first characters, as in the Captive Portal Vouchers' usernames. 

This should fix the behaviour and make the portal more graceful towards iOS devices.

Source for the attributes and their values: https://developer.apple.com/library/content/documentation/AppleApplications/Reference/SafariHTMLRef/Articles/Attributes.html

### DIFF
--- a/src/opnsense/scripts/OPNsense/CaptivePortal/htdocs_default/index.html
+++ b/src/opnsense/scripts/OPNsense/CaptivePortal/htdocs_default/index.html
@@ -173,7 +173,7 @@
               <form class="form-signin">
                   <h2 class="form-signin-heading">Please sign in</h2>
                   <label for="inputUsername" class="sr-only">Username</label>
-                  <input type="text" id="inputUsername" class="form-control" placeholder="Username" required autofocus>
+                  <input type="text" id="inputUsername" class="form-control" placeholder="Username" required autofocus autocomplete="none" autocapitalize="none" autocorrect="off">
                   <label for="inputPassword" class="sr-only">Password</label>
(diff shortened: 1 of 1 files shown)


src/opnsense/scripts/OPNsense/CaptivePortal/htdocs_default/index.html
```diff
@@ -173,7 +173,7 @@
               <form class="form-signin">
                   <h2 class="form-signin-heading">Please sign in</h2>
                   <label for="inputUsername" class="sr-only">Username</label>
-                  <input type="text" id="inputUsername" class="form-control" placeholder="Username" required autofocus>
+                  <input type="text" id="inputUsername" class="form-control" placeholder="Username" required autofocus autocomplete="none" autocapitalize="none" autocorrect="off">
```
johannrichard: `autocapitalize`
Specifies the auto-capitalization behavior of text elements. Possible values are:

`none`
Do not capitalize any text automatically.
Available in iOS 5.0 and later.

`autocorrect`
Specifies a form field allows automatic correction. If `off`, deactivates automatic correction for a form or input element; `on`, activates automatic correction for the element.

Support Level
Apple extension.

`autocomplete`
Specifies a form field accepts AutoFill information. If `off`, deactivates automatic completion for an input element; `on`, activates automatic completion for the element.

Related Tags
`input`

AdSchellevis:
@johannrichard looks good, thanks!

PR #1545, opened 2017-04-16
Changed paths to change Referrer

mimugmail:
Seems this was copied from pfSense


fichtner:
It's simple: we moved it. It's difficult to embed paths of the menu into the strings, as we can never guarantee consistency unless we do an "automatic" path conversion. Maybe we could do that, but for now let's merge this, thanks! :)

PR #1548, opened 2017-04-18
Translation

fabianfrz:
@fichtner There are even more translation issues to fix.

The idea with the partials is from @DokuKaefer, I have added even more places.


fichtner:
Merged, thanks!

PR #1551, opened 2017-04-20
Force restart radvd on TZ change

evbevz:
If radvd not restarted after TZ was changed, it continues write system log timestamps in old timezone.
services_radvd_configure() only forces reread config if radvd is running.

### DIFF
--- a/src/www/system_general.php
+++ b/src/www/system_general.php
@@ -133,6 +133,9 @@ if ($_SERVER['REQUEST_METHOD'] === 'GET') {
     }
 
     if (count($input_errors) == 0) {
+
+        $timezone_changed = $config['system']['timezone'] != $pconfig['timezone'];
+
         $config['system']['hostname'] = $pconfig['hostname'];
         $config['system']['domain'] = $pconfig['domain'];
         $config['system']['timezone'] = $pconfig['timezone'];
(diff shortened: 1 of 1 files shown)


src/www/system_general.php
```diff
@@ -229,6 +232,13 @@
         system_timezone_configure();
         filter_configure();
 
+        if ($timezone_changed) {
+            // force real radvd restart
+            $radvd = find_service_by_name('radvd');
+            killbypid($radvd['pidfile'], 'TERM', true);
+            services_radvd_configure();
```
fichtner: Maybe we should either fix the reload behaviour of radvd here or pass an argument to do it. HUP is a nice feature, but it never really works in edge cases?

evbevz: In this case HUP doesn't work.

fichtner: Yes, I'll adapt radvd to fully restart to simplify this code bit.

But keep in mind that time zone changes probably affect a lot more services, some can't even cope with time going backwards so the safest bet is to reboot.

fichtner:
@evbevz merged this manually, thanks! can you double-check for me?

fichtner:
@evbevz I reverted your change, because now that radvd always properly restarts the dhcp restart in that page takes care of it automatically. I'll credit the change to you in the notes anyway.

evbevz:
Are you sure? dhcp restarts radvd before call system_timezone_configure().

fichtner:
Oh, nice catch! system_timezone_configure() should be the first call after write_config() then

PR #1559, opened 2017-04-25
Removed unnecessary semicolon (Issue 1558 proposal) 

theq86:
This semicolon is not needed but seems also not to break the config.


fichtner:
Merged, thanks. :)

PR #1561, opened 2017-04-27
Fix Voucher server for squid authenticate

kekek2:
Voucher server creates user names with capital letters. Squid converts users names to lower case. because of this authenticate fails. Then squid does not have enough rights to write to voucher data base.

### DIFF
--- a/src/etc/rc.recover
+++ b/src/etc/rc.recover
@@ -62,7 +62,6 @@ audit:*:77:
 www:*:80:
 _ypldap:*:160:
 hast:*:845:
-voucher:*:1003:squid
 nogroup:*:65533:
 nobody:*:65534:
 
--- a/src/etc/rc.recover
+++ b/src/etc/rc.recover
@@ -62,6 +62,7 @@ audit:*:77:
 www:*:80:
 _ypldap:*:160:
 hast:*:845:
+voucher:*:1003:squid
 nogroup:*:65533:
 nobody:*:65534:
 
--- a/src/opnsense/mvc/app/library/OPNsense/Auth/Voucher.php
+++ b/src/opnsense/mvc/app/library/OPNsense/Auth/Voucher.php
@@ -91,14 +91,21 @@ class Voucher extends Base implements IAuthConnector
      */
     private function openDatabase()
     {
-        $db_path = '/conf/vouchers_' . $this->refid . '.db';
+        $db_dir = "/conf/db";
+        $db_path = $db_dir . '/vouchers_' . $this->refid . '.db';
+        if (!is_dir($db_dir))
+        {
+            mkdir($db_dir);
+            chgrp($db_dir, "squid");
+            chmod($db_dir, 0770);
+        }
         $this->dbHandle = new \SQLite3($db_path);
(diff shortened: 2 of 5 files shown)


src/opnsense/service/templates/OPNsense/Proxy/squid.user.local_auth.conf
```diff
@@ -1,5 +1,20 @@
+{% set voucher_method = [] %}
```
AdSchellevis: it's probably better to move this a new file using  https://github.com/opnsense/core/blob/stable/17.1/src/opnsense/service/templates/OPNsense/Proxy/squid.conf#L278 , something like  /usr/local/etc/squid/post-auth/local_voucher.conf

To make it more explicit where the feature is for, if I'm seeing things correctly it only tries to set "auth_param basic casesensitive on" when a voucher provider is used.

kekek2:
Fixed comments

src/opnsense/mvc/app/library/OPNsense/Auth/Voucher.php
```diff
@@ -134,6 +136,7 @@ private function userNameExists($username)
 
     private function setStartTime($username, $starttime)
     {
+        $this->dbHandle->exec("PRAGMA journal_mode=off");
```
AdSchellevis: why? disabling transactions might corrupt the database file.....

src/etc/rc.recover
```diff
@@ -62,6 +62,7 @@ audit:*:77:
 www:*:80:
 _ypldap:*:160:
 hast:*:845:
+voucher:*:1003:squid
```
AdSchellevis: @fichtner can you take a look at this? I'm not sure this is the proper way to add a group, next question is, what happens on update...

src/opnsense/mvc/app/library/OPNsense/Auth/Voucher.php
kekek2: When writing to the database, SQLIte tries to create temporary file for logging in /conf/ He does not have to do this right. Or you need to give write permission on the directory /conf (which is not very good). Or deny to create a temporary file logging. You can still create the directory /conf/db is that SQLite will have rights to create temporary files.

AdSchellevis: I see, this needs more thinking. we might need to move the file away into it's own directory, but that has some migration challenges.

src/etc/rc.recover
fichtner: this is the freebsd template, we should not clobber it

proxy has some legacy workaround already here:

https://github.com/opnsense/core/blob/master/src/etc/inc/system.inc#L1143-L1145

But it looks like an authentication framework pre-command so not the best place in loading kernel modules

kekek2:
How to do the migration I don't know yet

AdSchellevis:
Probably better to use pam_auth (https://linux.die.net/man/8/pam_auth) and create pam templates for the service.  Will also need some work in core (https://github.com/opnsense/core/blob/master/src/sbin/opnsense-auth), but will likely be more portable.

PR #1567, opened 2017-04-28
add lodash.js

fabianfrz:
required for building templates running client side and messing with objects and arrays

### DIFF
--- a/src/opnsense/www/js/lodash.js
+++ b/src/opnsense/www/js/lodash.js
@@ -1,0 +1,17084 @@
+/**
+ * @license
+ * Lodash <https://lodash.com/>
+ * Copyright JS Foundation and other contributors <https://js.foundation/>
+ * Released under MIT license <https://lodash.com/license>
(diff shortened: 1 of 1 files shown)


fabianfrz:
@fichtner If you are going to do a release this week, don't forget about this as it will break the quagga plugin if missing.

fichtner:
just master or both? hoping for a ok from @AdSchellevis (seems no need to add this to the default template)

fichtner:
https://github.com/opnsense/plugins/commit/e273ba6f491dddc30b59c166e558f8cc08fcab41 makes more sense to me ;)

fabianfrz:
will do it too - hoped to include it in core so it would be shared between plugins and could be used for refactoring of pages. However the impact is minimal as the paths are the only thing that needs to be adjusted in such a case.

AdSchellevis:
@fichtner we could always add it to core later if if we're going to use it anywhere else, although having a directory for the plugin itself is probably the best for now.

PR #1604, opened 2017-05-03
Update phpDynDNS.inc

evbevz:
An API exception occured
Error at /usr/local/etc/inc/plugins.inc.d/dyndns/phpDynDNS.inc:92 - Methods with the same name as their class will not be constructors in a future version of PHP; updatedns has a deprecated constructor (errno=8192)


fichtner:
looks good, thank you very much :)

PR #1607, opened 2017-05-03
allow json output of the firewall rule

fabianfrz:
this was just too easy...

If this is included add &format=json to the URL and you will get the rule as json.


fabianfrz:
pleas do a squashed commit - this does not make sense to be 2 commits.

fabianfrz:
@fichtner do you still need something here?

fichtner:
@AdSchellevis and me discussed this a a few times. We agree that starting from the bottom with the rule generation is the better approach, currently targeting NAT rule creation, working our way up to a full model, which can be used by plugins, before we start to go to the top and start replacing the GUI / introducing the API.

PR #1617, opened 2017-05-05
Allow CNAME overrides for unbound

### LINKED ISSUE #1614: [17.1.6] Overrides for unbound can't be CNAME
Only A, AAAA and MX are options.
It would be nice to have at least CNAME also.  Maybe SRV and TXT too.

Usage example for CNAME.
host-a.vlan10.example.org A 172.16.10.123
host-a.vlan20.example.org A 172.16.20.123
host-a.example.org CNAME host-a.vlan10.example.org

sjorge:
We allow for CNAME to be used in the host override for unbound.

This closes #1614.

### DIFF
--- a/src/etc/inc/plugins.inc.d/unbound.inc
+++ b/src/etc/inc/plugins.inc.d/unbound.inc
@@ -599,6 +599,9 @@ function unbound_add_host_entries()
                 case 'MX':
                     $unbound_entries .= "local-data: \"{$host['host']}{$host['domain']} IN MX {$host['mxprio']} {$host['mx']}\"\n";
                     break;
+                case 'CNAME':
+                    $unbound_entries .= "local-data: \"{$host['host']}{$host['domain']} IN {$host['rr']} {$host['cname']}\"\n";
+                    break;
             }
 
             if (!empty($host['descr']) && isset($config['unbound']['txtsupport'])) {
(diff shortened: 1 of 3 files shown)


fichtner:
looks good, thanks!

sjorge:
Hey @fichtner although the CNAME are properly returned (yay) some revolvers seem to have issues with them.

This is because unbound does not include the recursive A and/or AAAA records :(
There is an excellent thread here http://unbound.net/pipermail/unbound-users/2009-March/000507.html

Looks like it will probably never work with unbound without having an authoritative server behind it.

sjorge:
Example difference between dnsmasq with --cname=test.example.org,test.vlan10.example.org vs unbound with CNAME.

**dnsmasq**
```
$ dig +short test.example.org
test.vlan10.example.org.
127.0.0.2
```

**unbound***
```
dig +short test.example.org
test.vlan10.example.org.
```

sjorge:
@fichtner bump

fichtner:
yeah, agreed, I'll back it out.

sjorge:
Yeah I switched back to dnsmasq for now, nothing wrong with it but I was initially realling impressed with unbound too.

3rd3:
@fichtner You wrote "CNAME does not fully work in unbound at this point". Might this be fixed/implemented in Unbound since?

This seems to be useful for e.g. safe search/content enforcement. Currently there is only a general switch/checkbox for that, but adding custom CNAME host overrides might allow for more granular control (in my case my client wants SafeSearch on Google and Bing but it has been found too restrictive on YouTube).

PR #1627, opened 2017-05-08
Add plugins_reconfigure to squid

kekek2:
If we reconfigure squid, we may need to reconfigure all plugins. Or reconfiguration of one plugin may require reconfiguration of other plugin.

### DIFF
--- a/src/etc/inc/plugins.inc
+++ b/src/etc/inc/plugins.inc
@@ -241,3 +241,25 @@ function plugins_xmlrpc_sync()
     }
     return $sync_settings;
 }
+
+function plugins_squid($hook)
+{
+    foreach (plugins_scan() as $name => $path) {
+        require_once $path;
+        $func = sprintf('%s_squid', $name);
+        if (function_exists($func)) {
+            $workers = $func();
+            foreach ($workers as $when => $worker) {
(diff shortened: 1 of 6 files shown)


fichtner:
I'm pretty sure you can use plugins_configure() for this use case already, adding your own reconfigure points like "vendorproxyplugin" in your plugins, then just calling that. We just need to see how much core patching you need to call this cleanly without adding seemingly unused hooks that could break during development.

AdSchellevis:
You probably better add your own custom frontend template for your system and add the calls for your own system in there if you need to refresh components for your own system. 
(override the menu item, hook-in your own proxy component and keep that up to date)

Eventually we might look into an option to extend configd calls with plugins (add dependencies for example), but that's not very high on my list.

kekek2:
Move squid plugins from services() to plugins_squid()

fichtner:
ok, this can be even simpler: plugin_configure('webproxy'), no need for plugin_squid()

kekek2:
Done

fichtner:
Merging this as preparation for https://github.com/opnsense/plugins/pull/266 , but we need to tweak change a few things.

fichtner:
Thanks! :)

PR #1628, opened 2017-05-08
Fix bootup script

kekek2:
If bootup-script contains "-" in name, then in logs:
Oct 12 13:37:45	kernel: <118>eval: -ng_opnsense_bootup_run: not found
Oct 12 13:37:32	kernel: <118>eval: -freshclam_opnsense_bootup_run: not found

### DIFF
--- a/src/etc/rc.freebsd
+++ b/src/etc/rc.freebsd
@@ -21,7 +21,7 @@ ls /usr/local/etc/rc.d/* /etc/rc.d/* | while read rc_filename; do
 	# start/stop service
 	if [ "$is_enabled" == "YES" ]; then
 		if [ "$1" == "start" ]; then
-			pre_run_var="`basename $rc_filename`_opnsense_bootup_run"
+			pre_run_var="`basename $rc_filename`_opnsense_bootup_run | /usr/bin/sed 's/-/_/g'"
 			eval "pre_run=\$"$pre_run_var""
 			if [ ! -z "$pre_run" ]; then
 				eval "$pre_run"


fichtner:
Hmm, we should really be using $name, not $rc_filename

AdSchellevis:
have you tested this? on my end this appears to result in something like
`pre_run=$squid_opnsense_bootup_run | /usr/bin/sed 's/-/_/g'`

which then renders as non existing.... (so squid setup isn't started anymore at my end)

AdSchellevis:
@fichtner I think it should be https://github.com/opnsense/core/commit/f8547f27cd9adcb32bf809bbff4a02014f90f3f5 I don't think we can use $name here

fichtner:
@AdSchellevis dashes are not supported in shell vars, so this thing breaks. why not use name? it's the same thing sanitised for variable expansion, see e.g. /etc/rc.d/ftp-proxy, which has "ftpproxy".

AdSchellevis:
@fichtner I think you have to refactor the complete script to use $name here, it's not in it's scope https://github.com/opnsense/core/blob/master/src/etc/rc.freebsd#L7-L9

fichtner:
can do

AdSchellevis:
ok, thanks

fichtner:
@AdSchellevis @kekek2 please double-check https://github.com/opnsense/core/commit/e438bfc9022ff671b4077afde93458ba71f63188

AdSchellevis:
@fichtner checked, looks good. thanks!

PR #1633, opened 2017-05-10
Translate squids UIBootgrid

kekek2:


src/opnsense/mvc/app/views/OPNsense/Proxy/index.volt
```diff
@@ -52,7 +52,17 @@ POSSIBILITY OF SUCH DAMAGE.
                     'set':'/api/proxy/settings/setRemoteBlacklist/',
                     'add':'/api/proxy/settings/addRemoteBlacklist/',
                     'del':'/api/proxy/settings/delRemoteBlacklist/',
-                    'toggle':'/api/proxy/settings/toggleRemoteBlacklist/'
+                    'toggle':'/api/proxy/settings/toggleRemoteBlacklist/',
+                    'options': {
+                        'labels': {
+                            'all': "{{ lang._('All') }}",
+                            'infos': "{{ lang._('Showing {') }}" + "{ctx.start}" + "{{ lang._('} to {') }}" + "{ctx.end}" + "{{ lang._('} of {') }}" + "{ctx.total}" + "{{ lang._('} entries') }}",
```
fabianfrz: This line is not good - it breaks a sentence into fragments.

kekek2:
Fixed

fichtner:
no, you need to use formatting: https://docs.phalconphp.com/en/3.0.0/reference/volt.html#filters

something like {{ lang._('Please setup a voucher server first (%sgoto auth servers%s)') | format('<a href="/system_authservers.php">','</a>') }}

kekek2:
Done

fabianfrz:
@fichtner because this is very generic, do you think it is possible to patch bootgrid after loading the script so it applies to all tables?

kekek2:
_lang('') work only in .volt files. In included .js it not work

fabianfrz:
I know - I am thinking about the base layout file here: https://github.com/opnsense/core/blob/master/src/opnsense/mvc/app/views/layouts/default.volt

fabianfrz:
If you look at it, you just have to overwrite a property of the library after it is loaded:
https://github.com/opnsense/core/blob/master/src/opnsense/www/js/jquery.bootgrid.js#L1294-L1300

fabianfrz:
@fichtner I made a replacement for this which works globally.

PR #1635, opened 2017-05-11
Generate ssl options for IPv6 web configurator

ghostwheel42:
In lighttpd < 1.4.46 ssl options like ssl.use-sslv2 and
ssl.cipher-list have to be included twice.


fichtner:
Merged, thanks! Closing the other one without merge as we copy over the master change in the different file anyway.

fichtner:
@ghostwheel42 @sjorge hey guys, I'm working on strict IP binding for lighttpd and noticed this workaround is still there, but we've gone from 1.4.45 to 1.4.48... is this still an issue or can we remove this?

fichtner:
@ghostwheel42 thanks, that's what I did subsequently since we went from 1.4.45 directly to 1.4.48. If you could check out opnsense-devel / master to double check that would be great. It works fine on my end.

ghostwheel42:
@fichtner Sorry for the late answer. Yes, the new configuration works for me.

fichtner:
@ghostwheel42 thank you 👍

PR #1636, opened 2017-05-11
Toggle nat rule and associated rule if it exists

kekek2:


AdSchellevis:
I don't think it's a good idea to add more copies of similar functionality, I think it's better to use a similar construction as https://github.com/opnsense/core/blob/622e8dba35e51c002e6f9f1a5dac0bd404cf90b8/src/www/firewall_nat_edit.php#L261-L267 and cleanup the current delete_id too.

kekek2:
I removed toggle_id(). delete_id() called twice in firewall_nat.php, it is better to leave.

src/www/firewall_nat_edit.php
```diff
@@ -256,16 +255,33 @@
...
+                    if (isset($item['associated-rule-id']) && $item['associated-rule-id'] == $a_nat[$id]['associated-rule-id']) {
+                        unset($config['filter']['rule'][$key]);
+                        break;
+                    }
                 }
+                mark_subsystem_dirty('filter');
+            }
+
+            // toggle nat rule and associated rule if it exists
+            if (isset($a_nat[$id]["disabled"]) != $natent['disabled']) {
```
AdSchellevis: not sure this is a good idea, it could lead to side affects. When a user disables an associated filter rule, it will automatically re-enable on save here. For an explicit toggle it might work, but I'm really in doubt if we should do this (people clone and disable associated rules). @fichtner what do you think?

fichtner:
I don't like the associated rules. They are convenient, but do they pose a real issue if not disabled? They should not even be explicit, plugins code should inject them or people should use manual pass rules. Ideally I'd want them out of the config.xml, not sidestep them 😔

Let me get back to this tomorrow.

AdSchellevis:
timeout

PR #1647, opened 2017-05-18
routing tab in menu

fabianfrz:
Move Routing stuff in its own subtab

### DIFF
--- a/src/opnsense/mvc/app/models/OPNsense/Base/Menu/Menu.xml
+++ b/src/opnsense/mvc/app/models/OPNsense/Base/Menu/Menu.xml
@@ -41,24 +41,6 @@
             <Settings order="100" url="/system_hasync.php"/>
             <Status order="200" url="/status_habackup.php"/>
         </HighAvailability>
-        <Routes order="40" cssClass="fa fa-map-signs fa-fw">
-            <All order="10" url="/system_routes.php">
-                <Edit url="/system_routes_edit.php*" visibility="hidden"/>
-            </All>
-            <Log order="100" VisibleName="Log File" url="/diag_logs_routing.php"/>
-        </Routes>
-        <Gateways order="30" cssClass="fa fa-location-arrow fa-fw">
-            <All url="/system_gateways.php">
-                <All url="/system_gateways.php*" visibility="hidden"/>
-                <Edit url="/system_gateways_edit.php*" visibility="hidden"/>
-            </All>
-            <AllStatus VisibleName="Status" url="/status_gateways.php"/>
-            <Group url="/system_gateway_groups.php">
-                <Edit url="/system_gateway_groups_edit.php*" visibility="hidden"/>
-            </Group>
-            <GroupStatus VisibleName="Group Status" url="/status_gateway_groups.php"/>
-            <Log VisibleName="Log File" url="/diag_logs_gateways.php"/>
-        </Gateways>
         <Settings order="20" cssClass="fa fa-cogs fa-fw">
             <General order="100" url="/system_general.php">
                 <workAround url="/system_general.php*"/>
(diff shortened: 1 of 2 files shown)


fabianfrz:
spectacle l12877
@fichtner @AdSchellevis hope it looks good this way

fabianfrz:
@fichtner Please wait with the merge until https://github.com/opnsense/plugins/pull/163 is ready.

fichtner:
This will take some coordination and tweaking anyway. It's a long overdue change, but it's difficult to make this look perfectly clear, e.g. when Quagga isn't installed it would say "Routing: Static Routes" and it sounds off.

fabianfrz:
@fichtner quagga is the "Dynamic Routing" entry while "Static Routes" has been moved from "System"

fabianfrz:
I have also moved Gateways to this top level menu so all relevant layer 3 things are together.

fichtner:
I can see that, but it doesn't look "enterprisey" ;)

fabianfrz:
@fichtner Improvements?

fichtner:
Possibly, but I need to crunch a few old docs for that

AdSchellevis:
do you know you can move items in your plugins by deleting the original one?

fabianfrz:
@AdSchellevis no, I don't. Is there some documentation about that?

AdSchellevis:
@fabianfrz I don't have documentation, but an example is easy to provide, the following xml hides the documentation, forum and support items:
```

<menu>
    <Help>
        <Documentation  Visibility="delete"/>
        <Forum  Visibility="delete"/>
        <Support Visibility="delete"/>
    </Help>
</menu>
```

fabianfrz:
thanks @AdSchellevis should we close this and I restructure the menu only if the plugin is present?

AdSchellevis:
@fabianfrz I think it's better to do so, the current location is fine when quagga isn't used in my opinion.

fichtner:
Hey, I want to do more research ;)

mimugmail:
@fichtner Here some example views from "enterprisey" vendors.

Sophos:
image


Cisco ASA:
image

SecurePoint (doesn't offer dynamic routing):
image

fichtner:
@AdSchellevis does your opinion still stand? if yes we'll address it in FRR. In general I don't mind either way, though. It could help clarity and future growth.

mimugmail:
I could provide another screenshot of Sonicwall if needed :)

AdSchellevis:
@fichtner yes, although I don't have a very strong meaning about the subject, when FRR is used it seems to make more sense.

fichtner:
It could complicate documentation a bit. Routes and Gateways don't perfectly fit System or Interfaces at the moment, both of which are very "rich" menus.

Another approach would be to use "Networks" and "Devices" instead of "Interfaces" and shuffle assigned interfaces into Networks along with routes/gateways and keep all non-network device configuration in Devices. But that seems way more confusing than breaking out routing.

AdSchellevis:
Network and devices could indeed be better, but we shouldn't do that pre 18.7 in my opinion.

fichtner:
do you mean skip for 18.7 or wait for 18.7 to ship this?

AdSchellevis:
wait for, too big a change for a minor release

fichtner:
Here's a *test* for a separate Routing entry... 
screen shot 2018-04-02 at 8 08 48 pm

Put it on a branch for review (patching is futile, I changed menu stuff in master) https://github.com/opnsense/core/tree/routing_menu_maybe

fichtner:
Just seeing it, routing -> routes looks silly, but the gateway stuff ruins every attempt at consistency

mimugmail:
Why dont call it Static in menu above?

fabianfrz:
@fichtner just remove the intermediate heading routes. It is not needed.

mimugmail:
FYI

https://blog.webernetz.net/basic-mp-bgp-lab-cisco-router-palo-alto-fortinet/

Some views from Palo Alto how they are doing things

fichtner:
Continued in ticket...

muchahagrande:
IMHO, from a functional point of view, having the Routes and Gateway menues mixed with all other system general menues is not very well structured.
The action of routing is a very specific task and in is better placed in its own menue group.
The same is for gateway stuff. It is also a routing task.
We are used to find these functions under System menu beause they were included there in pfSense from the beggining, but for new users it would be no matter nor confusing in finding these functions under a Routing main menu item.

Chers

fichtner:
@mimugmail didn’t want the routing menu last we talked about it. It’s rather exhausting to bring everyone on the same side on a seemingly trivial matter (it’s not unfortunately)  so I personally do not have an strong opinion anymore.

muchahagrande:
It's ok, but I didn't want to lose the opportunity of giving my opinion.
I opened another issue asking this change, #2282, and it was closed because the subject was being discussed here, but I never read this until today. A little late :-)

Thank you for your reply

PR #1667, opened 2017-06-01
Add ability to have IPv6-only OpenVPN vpns

wget:
As discussed with Franco Fichtner on IRC, I wanted to have an IPv6 only VPN but the check in the WebUI was a little bit to restrictive as the network tunnel IPv4 seems to be mandatory.

I think this is a simple fix. Therefore, I decided to bring this PR on the table. This has not been tested (I don't have a CI for OPNsense - and I don't like to push untested code - sorry), but I assume (read hope) this will be enough to have this feature :)

### DIFF
--- a/src/www/vpn_openvpn_server.php
+++ b/src/www/vpn_openvpn_server.php
@@ -298,8 +298,10 @@ if ($_SERVER['REQUEST_METHOD'] === 'GET') {
         }
 
         if ($pconfig['dev_mode'] != "tap") {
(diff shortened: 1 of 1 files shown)


wget:
Ok. Testing in the PHP files made my VPN crash. Gonna investigate this on my side :) Before repushing a fix (and rebasing of course).

wget:
Not possible for now [because of upstream](https://community.openvpn.net/openvpn/ticket/208).

PR #1670, opened 2017-06-06
add local-zone typetransparent to resolving ptr zone

ra1nb0w:
fix ptr resolving. without this patch, unbound doesn't resolv ptr hostname with private ips. Probably there exists a better way.

### DIFF
--- a/src/etc/inc/plugins.inc.d/unbound.inc
+++ b/src/etc/inc/plugins.inc.d/unbound.inc
@@ -508,6 +508,9 @@ function unbound_add_domain_overrides($pvt = false)
         if ($pvt == true) {
             $domain_entries .= "private-domain: \"$domain\"\n";
             $domain_entries .= "domain-insecure: \"$domain\"\n";
+            if ((substr($domain, -14) == ".in-addr.arpa.") || (substr($domain, -13) == ".in-addr.arpa")) {
+                $domain_entries .= "local-zone: \"$domain\" typetransparent\n";
+            }
         } else {
             $domain_entries .= "stub-zone:\n";
             $domain_entries .= "\tname: \"$domain\"\n";


src/etc/inc/plugins.inc.d/unbound.inc
```diff
@@ -508,6 +508,9 @@ function unbound_add_domain_overrides($pvt = false)
         if ($pvt == true) {
             $domain_entries .= "private-domain: \"$domain\"\n";
             $domain_entries .= "domain-insecure: \"$domain\"\n";
+            if ((substr($domain, -14) == ".in-addr.arpa.") || (substr($domain, -13) == ".in-addr.arpa")) {
```
fichtner: looks good in general, but we should make this a regex with preg_match, maybe '/\.in-addr\.arpa\.?$/' works here?

ra1nb0w:
```
if (preg_match('/\.in-addr\.arpa\.?$/', $domain)) {
```
works flawless. thanks

fichtner: cool :) as a small follow up, maybe we should only match non-empty left side with .+  ?

ra1nb0w: nice idea.
```
if (preg_match('/.+\.in-addr\.arpa\.?$/', $domain)) {
```

ra1nb0w: do you prefer that I rebuild the patch?

fichtner: can do too, I just thought I'd stay clear before we both do the same work. Just let me know.

ra1nb0w: Ok. I do.

ra1nb0w: unfortunately I already deleted the fork so I create a new pull request. sorry.

Issue #1326, opened 2017-01-05
 [plugins] inject new NAT rules


fraenki:
Similar to #993 it should be possible (for plugins) to inject NAT rules as required.

fichtner:
Doesn't this cover all yet? If not, I know @fbrendel and @fabianfrz are waiting for this already.

Don't think this will be before 17.1.x though.

AdSchellevis:
it's on my list..... all firewall logic should follow the same rules, but given the impact of changing the standard parsing, we (@fichtner and I) decided to hold development off for some time.
Using anchors you should be able to inject rules already (like some services do).

fraenki:
>  Using anchors you should be able to inject rules already (like some services do).

I've seen the examples to inject regular FW rules. Am I missing something... does this work for NAT rules too?

AdSchellevis:
if I'm not mistaken both relayd and ftpproxy use an anchor to forward traffic, but a registration option will be added later....

fbrendel:
The ftpproxy plugin only adds the anchor and the ftp-proxy(8) daemon injects the rules.
And I believe the same does relayd.

Furthermore the plugin depends on regular FW rules to allow access to the ftp-proxy daemon.
This rules could be made automatically by the plugin. But this requires cluster aware services.

AdSchellevis:
@fichtner I'm refactoring the code behind "enablenatreflectionhelper", to me this looks like a lot of magic which could be simplified a lot.

Do you know if there's a good reason why we shouldn't replace the two lines below

https://github.com/opnsense/core/blob/ad1744ce240fc721a473c6f434269b55c6fee135/src/etc/inc/filter.inc#L847-L848

with something which looks more like:

`nat on [emX] proto tcp from [emX]:network to [redirect ip] port [redirect port] -> [emX] port 1024:65535
`

Where emX is the interface configured to forward the session. At my end this seems todo exactly the same, but maybe I'm missing something.

fichtner:
@AdSchellevis Judging by the commit history this was introduced by Scott in 2005 and kept that way. The [emX]:network network trick is what seems to make your suggestion work in one line. I'm all for it. We still have RC1 and RC2 to gather feedback so please go ahead. :)

AdSchellevis:
ok, will do :)

fichtner:
Im closing this since substantial work was done and we should discuss individual things in dedicated tickets.

fichtner:
[...]

PR #1679, opened 2017-06-12
firewall, replace static logic with plugin code

AdSchellevis:
There's still more work to do here (for example https://github.com/opnsense/core/issues/1326), but this batch cleans up most of the filter logic underneath. 
It would probably be best to cherry-pick the changes into master, the commits are quite logically ordered.


fichtner:
@AdSchellevis one more, I now get:

    06-12-17 18:44:18 [ There were error(s) loading the rules: /tmp/rules.debug:123: keep state on block rules doesnt make sense - The line in question reads [123]: block in  log quick on enc0 inet from {192.168.77.30} to {any} keep state label USER_RULE ]

Rule in question is a user rule on IPsec tab

fichtner:
reject has the same issue

AdSchellevis:
@fichtner will check right away

AdSchellevis:
fixed https://github.com/opnsense/core/tree/88c5005d0cf3f567a842e136de051a995f50a428

fichtner:
works, thanks!

Issue #1034, opened 2016-06-21
Feature Request - Option to resolve IPs to hostnames via DNS in Netflow/Insight


JulianSiS:
As the title says... it would be nice to have the option to resolve IPs to hostnames via DNS for the Netflow/Insight reporting.
See https://forum.opnsense.org/index.php?topic=3223.0

juj:
This would be an amazing feature, I was also looking on how to do this.

AdSchellevis:
![image](https://user-images.githubusercontent.com/9783985/32572411-be5f0cee-c4ca-11e7-8794-f8f2c7e2c244.png)

![image](https://user-images.githubusercontent.com/9783985/32572400-b2322b4a-c4ca-11e7-8df2-a32810a83da2.png)

It needs https://github.com/opnsense/core/blob/master/src/opnsense/mvc/app/controllers/OPNsense/Diagnostics/Api/DnsController.php and https://github.com/opnsense/core/commit/97be1979a7e3b9950518987cf89d303b081b6dd5

When reverse lookup is selected, the charts expands to full width, domain names tend to be longer then ip addresses which limits the visible addresses to only a few due to how nvd3 handles the legends.

fichtner:
All done, thanks!

PR #1685, opened 2017-06-16
Resolve IPs in Insight

HFsi:
Hi! Is there any way to do a reverse DNS in the info showed by insight?
It would be very usefull to see that info instead of IPs, or having to export and resolve by other means.
If not, could be an enhancement request?
Thanks


AdSchellevis:
@HFsi can you please rebase your changes on top of the current master branch? this is quite a strange pull...

fichtner:
it's not a PR, there  are no changes. We already have a ticket here: https://github.com/opnsense/core/issues/1034

PR #1690, opened 2017-06-23
Fix DynDNS update for IPv6

Alex9779:
I just stared using OPNsense with a downloaded version which was 17.1.4 where a custom v6 DynDNS update worked.
Then I did the upgrade to latest stable 17.1.8 and it just gave me the IPv4 address.
Seems the problem was just a little typo when refactoring...
This fixes it...


fichtner:
Looks good, thank you. Will get this into 17.1.9.

Issue #1692, opened 2017-06-25
[Bug] [Patch] fixup_default_gateway() vs apinger restart


speed47:
I currently have 3 configured gateways, with the 3rd one set as "default" for OPNsense itself.

I noticed that the "default" setting seemed to be ignored, because always the first gateway (in the PHP array order) was set as default on the system (as seen by `netstat -nr`).

After some digging I found out why: when applying a new gateway configuration, the "default" gateway is properly set, but right after, in the code path, `fixup_default_gateway()` is called, and decides to override the default gateway, because the "status" element of the array of each and every gateway is `none`. This is most probably because apinger has been restarted at the same time, and needs some seconds to decide if each gateway is really up or down.

In this case, the first configured gateway is chosen as the default gateway by  `fixup_default_gateway()` (even if its status is also reported as `none`).

A proposition of PR is coming up in some minutes, referencing this issue.

speed47:
Actually I wrote a patch for my system which is under 17.1.8, but I see that on master, fixup_default_gateway() logic has been entirely rewritten, and this race condition might no longer exist as a result. I'll try to cherry-pick minimum changes from master to confirm if it fixes the problem or not.

fichtner:
Yeah, that's correct. You can try "opnsense-update -t opnsense-devel", I think the latest changes are there.

fichtner:
Moving back to stable by "opnsense-update -t opnsense" ;)

speed47:
Hmm, opnsense-devel seems to be dated May 30, but the needed patches are just a bit more recent than that!

fichtner:
If your patch is small and makes sense maybe we should add it to stable/17.1 instead. The next opnsense-devel will have the rework?

speed47:
After reading the code carefully and adding some debug, it seems to me that the logic of `fixup_default_gateway` in stable is actually flawed and it's not just about a race condition with apinger restart.

As I understand it, the code will always elect the first configured gateway as the proper one to set - as long as it's not marked as "down" - regardless of the fact that it's selected as default or not, and regardless of the fact that the actual default gateway is up or not.

This is because in the foreach loop inspecting all gateways, it'll always exit after the first iteration if the first configured gateway is not the default, because of a `break` in the code: (comments are mine):

```
foreach ($gateways_arr as $gwname => $gwsttng) {
    // looping thru all defined gateways
    if ($gwsttng['ipprotocol'] == $ipprotocol) {
[...]

PR #1694, opened 2017-06-26
fix logic error in the behaviour of fixup_default_gateway()

speed47:
Fixes issue #1692


fichtner:
@AdSchellevis should go into stable/17.1? the code was reworked on master, but needs an extra round on -devel to materialise, so this PR is a candidate for 17.1.9 only.

AdSchellevis:
@fichtner sure, approximately the same change is also in https://github.com/opnsense/core/commit/67c9c5d62bd2311245702560d14140865b6dc317 on master.

PR #1724, opened 2017-07-18
Update base_form.volt

evbevz:
If style of header is not set in form xml, then PHP Notice appeared in log:
PHP Notice: Undefined index: style in ......


AdSchellevis:
@evbevz thanks!

PR #1730, opened 2017-07-19
Host-Uniq field in PPPoE connection

polxmod:
Add a custom field in the "advanced configuration" tab in pppoe connection page.
The custom field "host-uniq" is needed in order to connect with some ISP (for example Vodafone fiber in Italy).

Thread on forum: https://forum.opnsense.org/index.php?topic=5307.0

host-uniq1

### DIFF
--- a/src/etc/inc/interfaces.inc
+++ b/src/etc/inc/interfaces.inc
@@ -1582,6 +1582,8 @@ EOD;
             $provider = isset($ppp['provider']) ? $ppp['provider'] : "";
             $mpdconf_arr[] = "set pppoe service \"{$provider}\"";
             $mpdconf_arr[] = "set pppoe iface {$port}";
+            if (!empty($ppp['hostuniq'])) {
+              $mpdconf_arr[] = "set host-uniq {$ppp["hostuniq"]}";
         } elseif ($ppp['type'] == "pptp" || $ppp['type'] == "l2tp") {
             $mpdconf_arr[] = "set {$ppp['type']} self {$localips[$pid]}";
             $mpdconf_arr[] = "set {$ppp['type']} peer {$gateways[$pid]}";
--- a/src/www/interfaces_ppps_edit.php
+++ b/src/www/interfaces_ppps_edit.php
@@ -49,7 +49,7 @@ if ($_SERVER['REQUEST_METHOD'] === 'GET') {
     }
     // plain 1-on-1 copy
     $copy_fields = array('ptpid', 'type', 'username', 'idletimeout', 'uptime', 'descr', 'simpin', 'pin-wait',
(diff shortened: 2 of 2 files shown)


fichtner:
Hi Paolo,

Thanks for this. I have one question: are you sure this only applies to PPPoE? That's what the config file code assumes at the moment.


Cheers,
Franco

polxmod:
Hi Franco,
I don't know if you can use the custom tag for other type of connection, I can't find documentation regarding the usage of this tag outside a pppoe connection.

I have noticed that this field also appears in PPP, PPTP, L2TP connection. Maybe I should hide it if it's not pppoe.

Paolo

fichtner:
ok let's instead keep it for all, if empty it does not cause issues anyway :) just need to move the config write block a tiny bit. I'll take care of it.

fichtner:
alright...upon further research, it looks like a pppoe-only thing

fichtner:
I had to change a few things, I hope you don't mind. interfaces.php also needed the GUI change.

Can you test with:

    # opnsense-revert opnsense
    # opnsense-patch 7681a29


Cheers.
Franco

polxmod:
I've tested the connection and it works well. thanks !!

fichtner:
Super, let's get this into 17.7 right away then. 😊

polxmod:
Sorry I haven't fully tested the patch correctly. It needs more tweaks, I'll try better tomorrow evening and I'll post results.

polxmod:
Both. 
Line 1586 in interfaces.inc  should be:
$mpdconf_arr[] = "set service \"{$ppp['hostuniq']}|\"";
because is not "set host-uniq STRING" but "set service "string|"" (with the "" and the | inside it)

and it's assumed that is the binary version of you host-uniq

In this way I can get IP from ISP, but I can't navigate pages (I think is a problem with my net conf).
Sorry for the typo in the conf.

fichtner:
Ah, ok, try this then:

    # opnsense-revert opnsense
    # opnsense-patch 7681a29 ac40e52

polxmod:
Ok, I tried now and after reading the documentation of ng_pppoe it works (now I'm connected with it).
From the docs I read that the host-uniq tag is placed in "set pppoe service" attribute.
So: **set pppoe service "0xHOSTUNIQ|SERVICENAME"**
e.g. my conf for host-uniq is: **set pppoe service "0x577635332630543030313631393531|"**
The "0x" before and the "|"(pipe) after are are absolutely necessary.

I don't know if you prefer using only the "Service Name" field on the webGUI or you prefer make a code that merges the two values in a single string. 

Thank you 👍 I think all the people in Italy that wanted to get rid of the VodafoneStation router will be pleased.

PR #1740, opened 2017-07-27
Fix anchor name quoting

evbevz:
To have possibility in plugins to set anchors:
anchor myanchor quick
or
anchor "myanchor/nested" quick

Quoting of anchor name is auto-fixed if slash exists in anchor name.

### DIFF
--- a/src/opnsense/mvc/app/library/OPNsense/Firewall/Plugin.php
+++ b/src/opnsense/mvc/app/library/OPNsense/Firewall/Plugin.php
@@ -173,8 +173,12 @@ class Plugin
         foreach (explode(',', $types) as $type) {
             foreach ($this->anchors as $anchorKey => $anchor) {
                 if (strpos($anchorKey, "{$type}.{$placement}") === 0) {
+                    $fixed_anchor = "";
+                    foreach (explode(' ', $anchor) as $token) {
+                        $fixed_anchor .= strpos($token, '/') === FALSE ? "$token " : sprintf("\"%s\" ", str_replace('"', '', $token));
+                    }
                     $result .= $type == "fw" ? "" : "{$type}-";
-                    $result .= "anchor \"{$anchor}\"\n";
+                    $result .= "anchor {$fixed_anchor}\n";
                 }
             }
         }


fichtner:
something looks off here even in the code that we have... why does it matter how this is quoted? what do we do here that is not desired?

fichtner:
or is this about the attribute "quick" being mishandled?

evbevz:
man pf.conf:
Optionally, anchor	rules can specify packet filtering parameters using
     the same syntax as	filter rules.  When parameters are used, the anchor
     rule is only evaluated for	matching packets.  This	allows conditional
     evaluation	of anchors, like:

	   block on $ext_if all
	   anchor spam proto tcp from any to any port smtp

So, there are many different additional parameters can be with anchor.
Simplest solution is to make quoting in right way.

fichtner:
but the quoting is done to prevent this, a safety measure as you will, because we only ever intended to pass the name. if we want to do it right, we can't simply pass the string as is... what happens on "\n" ? are we going to inject other rules after a newline?

evbevz:
I think that plugins subsystem is developers area. It is not place, where strings goes to configuration from web-interface. If developer wants to break system, he can do it.
I thought that there is synthax check here for anchor name quoting.

fichtner:
The question is simple... how much flexibility do you need? We don't have to change the world in one day. If you only use quick here, we should add support for quick.

We don't want to restrict -- we want to provide safe constraints to develop within. For everthing else there is pf.conf. Because if not, these cases will bite us back during support hours. The developer who "broke" it will be gone, but someone else will have to find out what went wrong and users won't know how to debug. And if we support verbatim strings who does the validation of the rule syntax in them? Can we be sure the rules can't be manipulated by the user in a bad way through a config field on the GUI that allows adding unintended side effects to this?

evbevz:
Yes, I need only quick parameter.

fichtner:
does this work for you? f4acddb

evbevz:
Franco, it's good. Merge ancor_quick branch and delete my pull-request.

fichtner:
@evbevz thanks, done :)

PR #1744, opened 2017-07-27
Change Mirror URL

computeralex92:
To separate Mirror-Traffic and Website Traffic, a change of the URL is needed.
The website / mirror site behind both URLs is the same.


fichtner:
Merged, thanks!

PR #1749, opened 2017-07-30
MVC: rewrite static routes

fabianfrz:
ToDo:

- [x] add hooks to model or controller to add / remove / update static route
- [x] add UUID in migration
- [x] make gateway a select box
- [x] design stuff (make it look like the other pages)
- [x] remove old pages, run "make plist-fix" to regenerate file index

this is backward compatible to the old data - only adding an UUID is required.


src/opnsense/mvc/app/models/OPNsense/Routes/Route/Migrations/M0_0_1.php
```diff
@@ -0,0 +1,10 @@
+<?php
+namespace OPNsense\Routes\Route\Migrations;
+use OPNsense\Base\BaseModelMigration;
+class M0_0_1 extends BaseModelMigration
+{
+    public function run($model)
+    {
+       $model.save();
```
AdSchellevis: this looks a bit off, If I'm not mistaken it should add uuid without an empty stub as well (as long as the version number is higher)

fabianfrz: @AdSchellevis This is what I am thinking about how that is done. I made the model in a way it will not change any of the values so nothing breaks in mean time.

fichtner: I like this approach, but unsure whether this works as-is or requires further tweaks in the base model?

fabianfrz: @fichtner I am also not sure. Do you have an idea how to populate the gateway list. It comes from a php function in `gwlb` which returns an array of associative arrays

AdSchellevis: @fabianfrz I don't think we can easily pull out the gateway list the way it is now, we did quite some cleanups some time ago, but there still is way too much logic in there. Ideally you should be able to read the current gateways either from the config. 

Another option could be to serialise (parts of) the legacy data using a configd call which hooks back into the model (so we can cleanly do it now and use legacy parts in the meantime),  we only might need to look at the caching mechanism in there.

fabianfrz: @AdSchellevis I used a configd call to solve the issue.

fabianfrz:
This is how it's currently looking:
spectacle kn4043

fabianfrz:
@mimugmail This is for the static routes ;)

src/opnsense/mvc/app/models/OPNsense/Routes/Route/Migrations/M0_0_1.php
AdSchellevis: @fabianfrz ok, sounds good, I will try to read the code tomorrow to see if there's work to be done in the current configd model class.

mimugmail:
@fabianfrz I'm watching it for a couple of days :) Is it  already useable? Then I could throw up some test machines.

fabianfrz:
@mimugmail the routes are stored correctly in the XML but they are not applied to the system.

src/opnsense/mvc/app/controllers/OPNsense/Routes/Api/RoutesController.php
```diff
@@ -0,0 +1,169 @@
...
+                if ($node != null) {
+                    $this->backend_execute_route('delete', $node);
+                    $node->setNodes($this->request->getPost("route"));
+                    $validations = $mdlRoute->validate($node->__reference, "route");
+                    if (count($validations)) {
+                        $result['validations'] = $validations;
+                    } else {
+                        // serialize model to config and save
+                        $mdlRoute->serializeToConfig();
+                        $this->backend_execute_route('add', $node);
```
fabianfrz: @AdSchellevis I know that this is not clean yet as it currently ignores the `disabled` flag.

src/opnsense/mvc/app/models/OPNsense/Routes/Route.xml
```diff
@@ -0,0 +1,24 @@
...
+    <items>
+        <route type="ArrayField">
+          <network type="NetworkField">
+            <NetMaskRequired>Y</NetMaskRequired>
+            <Required>Y</Required>
+          </network>
+          <gateway type="JsonKeyValueStoreField">
+            <Required>Y</Required>
+            <ConfigdPopulateAct>interface gateways list</ConfigdPopulateAct>
+            <SourceFile>/tmp/gateway_list.json</SourceFile>
```
AdSchellevis: @fabianfrz you probably want to set ConfigdPopulateTTL to a low value here, to avoid caching, see https://github.com/opnsense/core/blob/master/src/opnsense/mvc/app/models/OPNsense/Base/FieldTypes/JsonKeyValueStoreField.php#L166-L171

src/opnsense/mvc/app/controllers/OPNsense/Routes/Api/RoutesController.php
AdSchellevis: @fabianfrz It might be better to write to a config file and apply on reload, keeps components more separated and easier to test. Eventually the bootup process could use the same approach then. 

something like, this in a config file:

```
[routeXXX]
version=4
net=192.168.0.0/24
gateway=10.0.0.1
```

And a simple script comparing the actual routes against the one in the list performing appropriate action (add/delete).

We might need to change some more nuts and bolts to integrate it with the rest (mainly system_host_route() ), but it would help detaching the routing subsystem in the process.

fabianfrz: @AdSchellevis This is highly dangerous as this may kill routes learned by DHCP and Quagga. The current code applies them during update as a configd call (without an apply step).

src/opnsense/mvc/app/models/OPNsense/Routes/Route.xml
fabianfrz: Probably a good idea

fabianfrz: done

src/opnsense/mvc/app/controllers/OPNsense/Routes/Api/RoutesController.php
AdSchellevis: @fabianfrz If we can't distinguish our routes with the ones created by dhcpd and/or quagga your definitely right on that point. I'm not sure what the best option is here, but apply immediate could lead to unexpected results (loss of access to the gui for example).

I need to think some more about this, ideally it would save a route configuration which we can load, but that means we need to track our changes somehow (without adding too much logic).

I don't mind if we first work on a version that just works using the current approach and refactor some in devel later on.

fabianfrz:
@AdSchellevis I reused the old help strings as is if possible so the GUI should look like the old one ;)

The only thing in question is the migration. What should I do here?

src/opnsense/mvc/app/models/OPNsense/Core/ACL/ACL.xml
```diff
@@ -820,15 +820,10 @@
     <page-system-staticroutes>
         <name>System: Static Routes</name>
         <patterns>
-            <pattern>system_routes.php*</pattern>
+            <pattern>ui/routes/</pattern>
```
fichtner: can this be ui/routes/* for future compat?

fichtner: ps: firmware uses "ui/core/firmware/*", I'm all for simplicity, but what is better now?

fichtner:
I've added another TODO ;)

fabianfrz:
@fichtner I had to rebase the branch because of conflicts but your changes should exist now.

fabianfrz:
@AdSchellevis ready for review

AdSchellevis:
@fabianfrz thanks! Can we keep the two legacy files in core for now? It's a bit easier to test on my end.

fabianfrz:
@AdSchellevis @fichtner asked me to delete them.

AdSchellevis:
@fabianfrz @fichtner was quite enthusiastic :) let's keep it like it is then.

fabianfrz:
note that you may have to run `/usr/local/opnsense/mvc/script/run_migrations.php` as the model needs to be migrated.

AdSchellevis:
I know :)

fichtner:
package upgrade or make upgrade takes care of migration ;)

fabianfrz:
@fichtner please don't edit the routes pages. This breaks this PR.

fichtner:
Please don't be like this. Nobody should complain about merge issues. I see them from time to time but it's a waste to go and vent about them, because they are part of the working process.

Neither of the last two edits have been added to nefariously tamper with this PR. They have been added to fix real world issues, namely PHP 7.1 support and a GitHub license file. There haven't and won't be any scheduled changes on the old routing pages, which won't ever guarantee we won't have to touch the code (again) before this is merged.

In retrospect my request to have the old pages deleted was suboptimal, because the merge time frame is not as fast as I expected. I'm sorry.

AdSchellevis:
@fichtner do you mind if I merge this in and keep the old versions in place as well? I'm having some doubts about applying the routes on save, when it's in master it's easier to test and propose changes.

fichtner:
Ok, sure

AdSchellevis:
@fabianfrz at my end `configctl interface gateways list` only listed null routes, I've realigned the requires to match the ones currently used in  https://github.com/opnsense/core/commit/a013aec6f5ac0b023c7006b2c3a2472788e115c3

AdSchellevis:
@fabianfrz @fichtner I will open an issue for the new static routes code, easier to discuss options. https://github.com/opnsense/core/issues/1774

Issue #1745, opened 2017-07-27
[regression] [17.7] disabling the 'quick' keyword in floating rules does not work anymore


fraenki:
After upgrading from 17.1.11 to 17.7 I've noticed that all firewall rules, where a Gateway is defined, no longer work as expected. The Gateway directive seems to be ignored. I'm still seing the `route-to` directive in rules.debug...

It should be noted that the Gateway in this case is connected through OpenVPN.

fraenki:
Hm, interesting, the _real_ issue is something different. It seems that the order in which rules are processed has changed. Now some of my (old) floating rules match first, so that my gateway rules never match.

Is this something intentional?

fichtner:
It depends on what the actual issue is. Floating rules should always match first, so that sounds reasonable.

One thing we've changed is not redirecting outgoing traffic from an interface itself through the respective gateway anymore, unless a specific gateway is set in the floating rules. This was done to allow better multi-wan capabilities for squid, which always seemed to have suffered from this limitation.

fraenki:
In this example I was using the gateway option in firewall rules to make sure certain outgoing traffic passes a OpenVPN tunnel (gateway) instead of my default WAN gateway.

bq. One thing we've changed is not redirecting outgoing traffic from an interface itself through the respective gateway anymore, unless a specific gateway is set in the floating rules.

Hm, I'm not sure if I fully understand this change. Does this mean that the gateway setting in _normal_ firewall rules was is less functional now? Do I _need_ to use floating rules for this kind of stuff from now on?

Can you point me to the issue/commit that introduced this change? Maybe it'll help me to better understand the underlying issue.

For me this turned out to be a breaking change. I'm not entirely sure this side-effect was intended. Maybe add a clear note to the 17.7 release notes or errata regarding this change and it's side-effects?

AdSchellevis:
@fraenki I think you're looking for this one https://github.com/opnsense/core/issues/1681, if you send me both versions of rules.debug (17.1.11 and 17.7) maybe I an easily check if this is related.

fraenki:
Got it! Floating rules have the keyword `quick` added in 17.7, which was missing in previous releases. For example:

```
# floating rule on 17.1.x
pass  on {  vtnet0  vtnet1  vtnet2  } inet proto tcp  from $local_networks to any port $http_https flags S/SA keep state  label "USER_RULE: allow HTTP/HTTPS to any"

# floating rule on 17.7
[...]

PR #1751, opened 2017-07-30
filter: fix floating rules default for quick parameter

fraenki:
Fix for #1745. Turns out this was a simple typo, but an important one, because it made it impossible to have floating rules without the `quick` keyword.


fichtner:
Merged, thanks!

PR #1752, opened 2017-07-31
Optimize Netflow

kekek2:
High CPU load during reading /var/log/flowd.log. Instead of sharing using the file, I did the communication through the socket between the flowd and flowd_aggregate.py


AdSchellevis:
I'm not sure this is a good idea, the current solution also caches data in the flowd files in case the daemon can't process it fast enough. Parsing the file logically shouldn't be the largest load in the process..

AdSchellevis:
we'er not going to include this, for reasons mentioned in my earlier response

PR #1753, opened 2017-07-31
Restore setLocale from Phalcon

kekek2:
After [this comit](https://github.com/opnsense/tools/commit/e9882554428519b38fd95ddb92604ee63737cd32) setting locale from Phalcon works fine.


fichtner:
Merged, thanks!

PR #1760, opened 2017-08-03
Fix setting locale

kekek2:


fichtner:
Sometimes chars come back as "?", right?

kekek2:
Yes

fichtner:
splendid, thanks!

PR #1762, opened 2017-08-03
proposal: new dialog to request confirmation

fraenki:
I needed a generic dialog (for plugins) to request a users confirmation for various actions. Currently there's only `stdDialogRemoveItem()` available, which really only suits the specific situation where one wants to remove an item, because the title reads "Remove":

dialog_remove

The new dialog is more generic and the title only says "Confirmation Required", that makes it very reusable:

dialog_confirmation

Additionally, I am suggesting to replace `stdDialogRemoveItem()` with `stdDialogConfirmation()` everywhere. The old dialog is a little redundant in itself, the header says "Remove" and the text usually repeats this somehow. If you agree I'd do the work and submit an additional PRs for core and plugins.

### DIFF
--- a/src/opnsense/mvc/app/views/OPNsense/CaptivePortal/clients.volt
+++ b/src/opnsense/mvc/app/views/OPNsense/CaptivePortal/clients.volt
@@ -95,7 +95,7 @@ POSSIBILITY OF SUCH DAMAGE.
                     grid_clients.on("loaded.rs.jquery.bootgrid", function(){
                         grid_clients.find(".command-disconnect").on("click", function(e) {
                             var sessionId=$(this).data("row-id");
-                            stdDialogRemoveItem('{{ lang._('Disconnect selected client?') }}',function() {
+                            stdDialogConfirmation('Remove','{{ lang._('Disconnect selected client?') }}',function() {
                                 ajaxCall(url="/api/captiveportal/session/disconnect/" + zoneid + '/',
                                         sendData={'sessionId': sessionId}, callback=function(data,status){
                                             // reload grid after delete
--- a/src/opnsense/mvc/app/views/OPNsense/CaptivePortal/clients.volt
+++ b/src/opnsense/mvc/app/views/OPNsense/CaptivePortal/clients.volt
@@ -95,7 +95,7 @@ POSSIBILITY OF SUCH DAMAGE.
                     grid_clients.on("loaded.rs.jquery.bootgrid", function(){
                         grid_clients.find(".command-disconnect").on("click", function(e) {
                             var sessionId=$(this).data("row-id");
-                            stdDialogConfirmation('Remove','{{ lang._('Disconnect selected client?') }}',function() {
+                            stdDialogConfirmation('Remove','{{ lang._('Disconnect selected client?') }}','OK','Cancel',function() {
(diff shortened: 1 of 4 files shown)


fichtner:
we should change these to WARNING, DANGER makes not a lot of sense for a confirmation even in the remove case

the current dialog in the js file cannot be translated, we need to pass all labels including title....which would also help with not duplicating this function.

fraenki:
@fichtner Thanks, I've committed the following changes:

* changes DANGER to WARNING
* made the title customizable
* replaced stdDialogRemoveItem() with stdDialogConfirmation()
* updated all occurences of stdDialogRemoveItem()

I'd do the same for plugins (acme) if you agree to this change.

fichtner:
as an intermediate step could we maybe emulate stdDialogRemoveItem() using stdDialogConfirmation() ? I don't feel comfortable with axing the function right away.

fichtner:
PS: note that since 17.7 is out and stdDialogConfirmation() not in it, what people like to do is not update the core, but install newer plugins right away so the javascript will fail if we don't put in safeguards :/

fraenki:
> as an intermediate step could we maybe emulate stdDialogRemoveItem() using stdDialogConfirmation()?

Do you mean something like this?

```
function stdDialogRemoveItem(message, callback) {
  stdDialogConfirmation('Remove', message, 'Yes', 'Cancel', callback)
}
```

fichtner:
Yes, exactly 😊

AdSchellevis:
Maybe it's better to wrap stdDialogConfirmation into a dialog for the type it's reporting, like:
```

stdDialogConfirmation(title, message, accept, decline, level, callback) {}

stdDialogWarning(title, message, accept, decline, callback) {
   stdDialogConfirmation(title, message, accept, decline, BootstrapDialog.TYPE_WARNING, callback);
}
```

What do you think?

fichtner:
that's fine too, but I wouldn't like adding more stdDialogXXX variants. ideally there should just be stdDialogConfirmation() that is used everywhere in the way that it needs to be used.

fraenki:
@fichtner Added the emulation for stdDialogRemoveItem() with https://github.com/opnsense/core/pull/1762/commits/607d955ed03584765cbb5fca57ea3ac2a847acc7.

fichtner:
I'll take the src/opnsense/www/js/opnsense_ui.js change only for now. Can you make sure that you have your own stub (different name) in the plugins for now? I don't think we should switch to depending on this right away, broken javascript because of missing functions is really hard to support.

fraenki:
@fichtner Yeah, I've got a copy of the function in https://github.com/opnsense/plugins/commit/9bc93a30203c76ba40c198e8509f5bca052f0076 already.

fichtner:
but it's the same name, no? not sure what happens if you have two...

fichtner:
I added your proposal, but changed it a bit: we now have stdDialogConfirm() and stdDialogInform() to avoid clashing definitions. Translations for the captive portal, bootgrid annotations that it currently doesn't translate.

fichtner:
This wasn't linked https://github.com/opnsense/core/commit/fb6125ccb2a73d5f3d8ecbd2ff316f2cfd81d532

PR #1763, opened 2017-08-05
use strings which I guessed from dhcp6c source and copied from RFC 24…

### LINKED ISSUE #1178: Translation issues in interfaces.php
This range is unclear (configuration directives instead of normal language) - it should be changed to be understandable by users:
https://github.com/opnsense/core/blob/master/src/www/interfaces.php#L2564-L2630

This should be kept together in the gettext call:
https://github.com/opnsense/core/blob/master/src/www/interfaces.php#L2683

fabianfrz:
…71; closes #1178


fichtner:
Very cool, thank you! :)

PR #1769, opened 2017-08-09
Fix DHCP Leases PHP Illegal Offset Warnings

wevcode:
I had some issues with "illegal offset" warnings so I was not able to see dhcp lease times.
This fixed it for me and it seems that these isset-checks are missing.

### DIFF
--- a/src/www/status_dhcp_leases.php
+++ b/src/www/status_dhcp_leases.php
@@ -48,15 +48,10 @@ function adjust_gmt($dt)
     }
 
     foreach ($dhcpd as $dhcpditem) {
-        $dhcpleaseinlocaltime = $dhcpditem['dhcpleaseinlocaltime'];
-        if ($dhcpleaseinlocaltime == "yes") {
-            break;
+        if (isset($dhcpditem['dhcpleaseinlocaltime']) && $dhcpleaseinlocaltime == "yes") {
+            $ts = strtotime($dt . " GMT");
+            return strftime("%Y/%m/%d %I:%M:%S%p", $ts);
         }
-    }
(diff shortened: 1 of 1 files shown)


src/www/status_dhcp_leases.php
```diff
@@ -48,9 +48,11 @@ function adjust_gmt($dt)
     }
 
     foreach ($dhcpd as $dhcpditem) {
-        $dhcpleaseinlocaltime = $dhcpditem['dhcpleaseinlocaltime'];
-        if ($dhcpleaseinlocaltime == "yes") {
-            break;
+        if (isset($dhcpditem['dhcpleaseinlocaltime'])) {
```
AdSchellevis: probably better to replace with isset($dhcpditem['dhcpleaseinlocaltime']) && $dhcpditem['dhcpleaseinlocaltime'] == "yes"  and ditch $dhcpleaseinlocaltime

wevcode: Do you not need it after `foreach`?

fichtner: you can inline the trailing code for $dhcpleaseinlocaltime:

if (isset($dhcpditem['dhcpleaseinlocaltime']) && $dhcpditem['dhcpleaseinlocaltime'] == "yes") {
         $ts = strtotime($dt . " GMT");
         return strftime("%Y/%m/%d %I:%M:%S%p", $ts);
}

wevcode:
@AdSchellevis @fichtner Changed.

AdSchellevis:
@wevcode thanks!

PR #1791, opened 2017-08-27
patch translations into bootgrid at runtime, based on the work of kekek2

fabianfrz:
patch it into the library in the template so it will be applied to all templates.

Strings used from the original pull request.

closes #1633


AdSchellevis:
@fabianfrz thanks, looks good

PR #1809, opened 2017-09-08
Icap whitelist

kekek2:
Exclude domains from antivirus check.

### DIFF
--- a/src/opnsense/mvc/app/controllers/OPNsense/Proxy/forms/main.xml
+++ b/src/opnsense/mvc/app/controllers/OPNsense/Proxy/forms/main.xml
@@ -537,6 +537,24 @@
                 <allownew>true</allownew>
                 <advanced>true</advanced>
             </field>
+            <field>
+                <id>proxy.forward.icap.whiteList</id>
+                <label>Whitelist</label>
+                <type>select_multiple</type>
+                <style>tokenize</style>
+                <help><![CDATA[Whitelist destination domains.<br/>
+        You may use a regular expression, use a comma or press Enter for new item.<br/>
+        <div class="alert alert-info">
+            <b>Examples:</b><br/>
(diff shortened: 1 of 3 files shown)


src/opnsense/mvc/app/controllers/OPNsense/Proxy/forms/main.xml
```diff
@@ -537,6 +537,24 @@
                 <allownew>true</allownew>
                 <advanced>true</advanced>
             </field>
+            <field>
+                <id>proxy.forward.icap.whiteList</id>
+                <label>Whitelist</label>
```
fabianfrz: I would rename that as "Whitelist" can be misleading. Basically ICAP is some kind of an remote procedure call protocol and not a ACL enforcement tool (even if it's most often used for that). I would prefer something like "Exclusion List" which fits the use case better (traffic will not be sent to the ICAP server).

mimugmail:
I applied the patches and did some testing, works good 👍

fichtner:
This may be relevant as well while we're at it:

https://forum.opnsense.org/index.php?topic=5866.msg24763#msg24763

fabianfrz:
@fichtner I took the link from the config for the docs:
https://github.com/opnsense/plugins/blob/master/www/c-icap/src/opnsense/service/templates/OPNsense/CICAP/virus_scan.conf#L4

fabianfrz:
ups - my mistake, we should change the default settings in squid

fabianfrz:
see #1822 - please note that this will be only used for new installations or installations, where the config of the proxy was never saved.

fichtner:
@fabianfrz that's alright, we worry about the future, not the past 👍

fichtner:
also added to 17.7.4, thanks!

PR #1816, opened 2017-09-10
jinja: Include any file - otherwise jinja cannot include files outside the template dir

fabianfrz:


AdSchellevis:
@fabianfrz why would we want this? I don't think it's a good idea to include arbitrary files outside the template scope per se.

fabianfrz:
I need to generate a key for Tor which needs to be included in the template.

fichtner:
this will be all kinds of ugly, which encoding is there, race conditions in writing / template generation, etc. If you have a secret it must be stored in the config.xml for writing...

AdSchellevis:
If it's part of the configuration, you probably best store it in the config and read it from there.
A bit like how the keys in Tinc work.
1) generate key command
https://github.com/opnsense/plugins/blob/master/security/tinc/src/opnsense/scripts/OPNsense/Tinc/generate_keypair.py
2) link in configd
https://github.com/opnsense/plugins/blob/master/security/tinc/src/opnsense/service/conf/actions.d/actions_tinc.conf#L1
3) use on save.
https://github.com/opnsense/plugins/blob/master/security/tinc/src/opnsense/mvc/app/controllers/OPNsense/Tinc/Api/SettingsController.php#L79

fabianfrz:
@AdSchellevis this worked. https://github.com/opnsense/plugins/pull/258/commits/3124ac81a7a0467b18514460145ab7b8a96a25e7

AdSchellevis:
@fabianfrz nice!

Issue #1821, opened 2017-09-13
Multi WAN testbeds


mimugmail:
I did some testing with Multi WAN in combination with (hopefully not too soon) deprecated "Enable gateway switching". My intention is to test if services enabled at WAN side are reachable (WebGUI and just ping wan address)

Scenario: LAN, WAN1 (Static IP, default route), WAN2 (dhcp, receives default gw, but not set to default)

Test1:
No gateways in FW rules.
No gateway groups
GW switching active

Result1:
Failover works
Both WANs reachable

---

Test2:
No gateways in FW rules.
No gateway groups
GW switching *NOT* active

Result2:
Failover doesn't work
Both WANs not reachable

Sidenote2: 
Enabling gw switching afterwards doesn't fix this situation without reboot. Reenabling WAN makes default gw available again but in UI it's marked as pening. Only restart of apinger fix this (just cosmetic, perhaps some syscall missing here)

---

Test3:
No gateways in FW rules.
GWGROUP, WAN1 Tier1, WAN2 Tier2
GW switching active

Result3:
Failover works
Both WANs reachable

Sidenode3: This time reenabling WAN1 didn't end with apinger status pending

---

Test4:
FW Rule LAN to ANY set to GWGROUP.
GWGROUP, WAN1 Tier1, WAN2 Tier2
GW switching active

Result4:
Same as Result3

---

Test5:
FW Rule LAN to ANY set to GWGROUP.
GWGROUP, WAN1 Tier1, WAN2 Tier2
GW switching *NOT* active

Result5:
Both WANs not reachable
Failover for LAN rule works

I'm not sure if this results helps in any way, but I hope it let's gateway switching live a bit longer, since as a MSP we have the requirement to reach our managed firewalls via both/all WANs. I think this also relates to services like OpenVPN to make them redundantly available.

Perhaps other people also want to contribute their experiences.

L1ghtn1ng:
As gateway switching uses apinger to do a ping check I can right a python script to do a ping check to replace apinger if anyone is interested in that and gets agreed to?

I would need someone to wire it in to opnsense though as I do not know PHP.


________________________________

mimugmail:
For the archive: 
WAN static, set to default, everything works but gateway status "Pending"
Switched to WAN DHCP as default

Result: default gateway route lost, not reproduceable when status for both OK. Will wait until this happens again. Restarting apinger doesn't help. /usr/local/etc/rc.reload_interfaces will fix it.
Will also try reload routing next time

[...]

PR #1829, opened 2017-09-15
make configctl call "newip" public to cron

mimugmail:
For Multi WAN with DHCP it might be good to have periodic renewal of the interface to fix error with stall routes ( documented in #1821 ).

It could also help in other situations e.g. buggy dhcp servers in the wan world to renew the address in given time ranges.

### DIFF
--- a/src/opnsense/service/conf/actions.d/actions_interface.conf
+++ b/src/opnsense/service/conf/actions.d/actions_interface.conf
@@ -16,6 +16,7 @@ command:/usr/local/etc/rc.newwanip
 parameters:%s
 type:script
 message:rc.newwanip starting %s
+description:Periodic interface renewal
 
 [newipv6]
 command:/usr/local/etc/rc.newwanipv6


fichtner:
I thought we were talking about a workaround... ;)

maybe we don't need this here, you could always copy the file command to have it retained locally. this can also be solved by ifconfig up/down or maybe even probing the dhcp pid with an appropriate signal?

mimugmail:
You mean creating a file e.g.
/usr/local/opnsense/service/conf/actions.d/actions_dhcpupdate.conf

Add the action with descripton and set in cron?
So this file wont be overwritten?

fichtner:
yes, exactly

mimugmail:
Ok, I'll close this one, but I'd not be sad if this would also go into master some time 👍

fichtner:
we will find the root cause and fix that so this isn't needed anymore, I promise.

mimugmail:
I'm testing right now the openvpn multi remote patch and now I'm in this unknown state. Motivated to troubleshoot or are you in weekend yet? :)

PR #1837, opened 2017-09-19
Captive portal: add expiry time

Stephanowicz:
adding expiry time so that a voucher can only be used til that time. Default is no expiry.

### DIFF
--- a/src/opnsense/mvc/app/library/OPNsense/Auth/Voucher.php
+++ b/src/opnsense/mvc/app/library/OPNsense/Auth/Voucher.php
@@ -242,7 +242,7 @@ class Voucher extends Base implements IAuthConnector
                     $generatedPasswordHash = crypt($generatedPassword, '$6$');
                     $stmt = $this->dbHandle->prepare('
                                 insert into vouchers(username, password, vouchergroup, validity, expirytime, starttime)
-                                values (:username, :password, :vouchergroup, :validity, :expirytime,:starttime)
+                                values (:username, :password, :vouchergroup, :validity, :expirytime, :starttime)
                                 ');
                     $stmt->bindParam(':username', $generatedUsername);
                     $stmt->bindParam(':password', $generatedPasswordHash);
--- a/src/opnsense/mvc/app/library/OPNsense/Auth/Voucher.php
+++ b/src/opnsense/mvc/app/library/OPNsense/Auth/Voucher.php
@@ -105,6 +105,8 @@ class Voucher extends Base implements IAuthConnector
                     , vouchergroup  varchar2  -- group of vouchers
                     , validity      integer   -- voucher credits
                     , starttime     integer   -- voucher start at
+                    , expirytime    integer   -- voucher valid until - '0' = disabled
+			   
                     , vouchertype   varchar2  -- (not implemented) voucher type
                     , primary key (username)
                 );
@@ -177,9 +179,10 @@ class Voucher extends Base implements IAuthConnector
      * @param int $count number of vouchers to generate
      * @param int $validity time (in seconds)
      * @param int $starttime valid from
+     * @param int $expirytime valid until ('0' means no expiry time)
      * @return array list of generated vouchers
      */
-    public function generateVouchers($vouchergroup, $count, $validity, $starttime = null)
+    public function generateVouchers($vouchergroup, $count, $validity, $expirytime, $starttime = null)
     {
         $response = array();
         if ($this->dbHandle != null) {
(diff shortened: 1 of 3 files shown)


src/opnsense/mvc/app/library/OPNsense/Auth/Voucher.php
```diff
@@ -237,20 +241,22 @@ public function generateVouchers($vouchergroup, $count, $validity, $starttime =
                     // save user, hash password first
                     $generatedPasswordHash = crypt($generatedPassword, '$6$');
                     $stmt = $this->dbHandle->prepare('
-                                insert into vouchers(username, password, vouchergroup, validity, starttime)
-                                values (:username, :password, :vouchergroup, :validity, :starttime)
+                                insert into vouchers(username, password, vouchergroup, validity, expirytime, starttime)
+                                values (:username, :password, :vouchergroup, :validity, :expirytime,:starttime)
```
fabianfrz: add a space after the comma

Stephanowicz: hi - just a numb question: how to find Squash and merge? :D 
this is my first one - so I dunno...

fabianfrz: This can be done on GitHub using the button for that. The alternative is doing it on the shell using an interactive rebase.

fichtner: we can do squash with final merge, but it would be better to add all commits to the same PR, you can just update your git PR branch and GitHub will update the PR

Stephanowicz: ok, seems like I can't do that without writepermissions...
probably it's not what I wanted... the aim is to have all 3 pullrequests in one

fabianfrz: @Stephanowicz I would recommend you to merge all your branches together. When @fichtner will merge it squashed, all merge commits are gone anyway.
If you want to improve your git skills, you can try `git rebase -i master` on that branch.

fichtner: no worries, take your time to sort this out. it should be possible when all commits are on the same branch and pushed

AdSchellevis:
@Stephanowicz I don't think we should add this, it will break existing installations (missing field) and you can very likely achieve the exact same behaviour by calculating the validity (instead of storing another) and setting the start time to now

Stephanowicz:
...had already feared something like that :D 

Well, the idea behind is a little bit different:
You have a general expiry time - after this the voucher gets invalid, if it has been used or not.

But, whatever - if we will use captive portal for our youth center, we will need this and I can implement this - just thought, that it would be a desireable feature - as I have seen on other portals, too...^^

Cheers,

Stephan

AdSchellevis:
@Stephanowicz I see, well, we can implement an absolute limit, but we have to take care of migration for existing installations. 

Can we agree that you first polish your version, then I can look into the migration part of the issue before we will add it to the core package.

Stephanowicz:
Ok, just thought, that a workaround would be to make different groups that You can delete after a day or so, for example. But this requires that someone has an eye on this, which is mostly not the case :D 

Migrating means, to import existing data into the new database structure... hmm, dunno if I can help here. 

What do You think needs to be polished? Any help appreciated! ^^

Cheers,

Stephan

fichtner:
@Stephanowicz I think Ad means merging everything into one PR to see the whole picture :)

Stephanowicz:
Yes - this was my first question, how to do that :D 
Still not sure, how to manage this!
Can You guide me? - btw I have no git shell...

fichtner:
if you grant me write access to your fork I can fix up the branch

fichtner:
that should be all from patch-2, patch-3 and patch-4 ?

Stephanowicz:
Yep! Thank You!

fichtner:
ok, github editor probably creates new branches for your changes. make sure you stay on patch-1 branch when editing, then this PR is automatically updated.

AdSchellevis:
@Stephanowicz finally I found some time to look into your feature and did some cleanups / added backwards compatibility for existing installations (https://github.com/opnsense/core/commit/83bde8ef1bfe9701e90eb42cf7c61f6f39fd9947)

Can you do some additional testing and confirm this is how it's supposed to be?

Stephanowicz:
Hi, that's great - thank You!

Well, @moment I don't know when I'll find time for testing, probably next WE.

Cheers, Stephan

AdSchellevis:
You're welcome!

No rush, just let us know your findings, so @fichtner can pull it in stable when ready.

PR #1851, opened 2017-09-28
Don't used OPNsense.proxy.forward.icap.UsernameHeader

kekek2:


src/opnsense/service/templates/OPNsense/Proxy/squid.conf
```diff
@@ -246,6 +246,9 @@ icap_client_username_encode on
 {%   else %}
 icap_client_username_encode off
 {%   endif %}
+{%   if helpers.exists('OPNsense.proxy.forward.icap.UsernameHeader') %}
```
fabianfrz: the empty check is missing: `and ... != ''`

mimugmail:
I applied the patch with #295 and it's working as expected :)

fichtner:
Merged, thanks!

PR #1870, opened 2017-10-09
Include an "info" action variant for bootgrid.

dharrigan:
-=david=-


fichtner:
Ideally, we want to find a way to register buttons automatically instead of defining new groups for different kinds of clusters. But for now this works well enough. Thanks!

PR #1873, opened 2017-10-09
Support non-whitespace field separators for URLtable lists

shonjir:
Adds support for non-whitespace field separators to URLtable IP lists.

Some IP lists use alternate characters as field separators - commas, semicolons, vertical bars.  For example, feeds from http://osint.bambenekconsulting.com/feeds/ use commas to separate list values.

### DIFF
--- a/src/etc/inc/filter.inc
+++ b/src/etc/inc/filter.inc
@@ -175,7 +175,7 @@ function process_alias_urltable($name, $alias_type, $url, $freq, $forceupdate = 
                     continue;
                 }
                 // cleanse line item
-                $line = preg_split('/\s+/', $line)[0];
+                $line = preg_split('/[\s,;|]+/', $line)[0];
                 if ($alias_type == "urltable_ports") {
                     // todo: add proper validation for ports here
                     fwrite($output_file_handle, "{$line}\n");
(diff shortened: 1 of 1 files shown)


src/etc/inc/filter.inc
```diff
@@ -175,7 +175,7 @@ function process_alias_urltable($name, $alias_type, $url, $freq, $forceupdate =
                     continue;
                 }
                 // cleanse line item
-                $line = preg_split('/\s+/', $line)[0];
+                $line = preg_split('/[\s,;|]+/', $line)[0];
```
fichtner: wouldn't "#" make sense as well here?

src/etc/inc/filter.inc
```diff
@@ -175,7 +175,7 @@ function process_alias_urltable($name, $alias_type, $url, $freq, $forceupdate =
...
                 if ($alias_type == "urltable_ports") {
                     // todo: add proper validation for ports here
```
fichtner: Err.. I know this is out of scope, but this shouldn't be very hard to check if port is valid :)

src/etc/inc/filter.inc
shonjir: It would.

fichtner: would you update or shall I post-merge?

fichtner:
thanks a lot for this! :)

PR #1881, opened 2017-10-18
Firewall plugin: register table

evbevz:
It is needed to allow register tables in plugins.


AdSchellevis:
@evbevz nice addition, thanks.

PR #1882, opened 2017-10-18
Captive portal plugins access

evbevz:
Plugins ability to register access paths to custom controllers in captive portal config.
For example, I made captive portal plugin for sms-authentication, and clients must have access to my sms-authentication controller.

### DIFF
--- a/src/opnsense/mvc/app/models/OPNsense/CaptivePortal/CaptivePortal.xml
+++ b/src/opnsense/mvc/app/models/OPNsense/CaptivePortal/CaptivePortal.xml
@@ -117,5 +117,12 @@
                 </content>
             </template>
         </templates>
+        <PluginsAccess>
+            <Access type="ArrayField">
+                <Route type="TextField">
+                    <Required>Y</Required>
+                </Route>
+            </Access>
+        </PluginsAccess>
     </items>
 </model>
--- a/src/opnsense/service/templates/OPNsense/Captiveportal/lighttpd-zone.conf
+++ b/src/opnsense/service/templates/OPNsense/Captiveportal/lighttpd-zone.conf
@@ -104,6 +104,11 @@ proxy.server = ( "/api/captiveportal/access/" => (
 		( "host" => "127.0.0.1",
 		  "port" => 8999 )
 	)
+{% if helpers.exists('OPNsense.captiveportal.PluginsAccess') %}
+{% 	for access in helpers.toList('OPNsense.captiveportal.PluginsAccess.Access') %}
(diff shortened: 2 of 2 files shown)


AdSchellevis:
@evbevz  I'm not sure we should add this, there's no use case for it in the standard product, maybe it's better to look for a template include construction for custom settings so we don't have to extend the standard model (more similar like the proxy/squid approach).

evbevz:
For custom functionality CP template must have access to custom controller. Access to controllers is guarded by authorisation system. I can't find another way to get access to custom controller.

AdSchellevis:
@evbevz for new overwrite options, see https://github.com/opnsense/core/issues/1934. With this you can create and maintain your own templates for the installed ones.

pzwahlen:
@evbevz This has been rejected and closed a long time ago, however I need to solve the exact same problem and implement some sort of SMS-based captive portal. I posted in the forum a long time ago without any reply from devs (https://forum.opnsense.org/index.php?topic=9749.msg44477#msg44477).

Would you be willing to share some code samples that you used at the time or point me at a way to solve the problem using "overwrites" as mentioned by @AdSchellevis ?

Patrick

evbevz:
Hi, @pzwahlen. My solution has:
- access controller with sms-auth logic
- settings controller to store sms-gate settings
- service controller to get some stats and current state
- sms-athenticator implements IAuthConnector
- proxy hook
- captive portal config hook to get unauthorised access to access controller (it can be made by override, abouth whish @AdSchellevis says)

If you don't need proxy logs, some parts can be skipped.

pzwahlen:
Hi,

I'm looking for a solution in a private environment (large house with 2 families, kids en friends...). The idea is to avoid sharing main wifi passwords and provide a somewhat bandwidth-limited guest access. 

So far, the built-tin template without any authentication ("click to connect") does the job. I just wanted to improve on it, without having to use vouchers or a user DB.

I don't intend to intercept the trafic, so proxy integration won't be used.

My main concern is that the solution needs to survive upgrades, or it will be hard to maintain over time.

Would you be willing to share or can we discuss pricing for a private setup ?

Anyway, thanks a ton for having taken the time to respond.

pzwahlen:
@evbevz Would you be willing to share your Captive Portal solution ?

evbevz:
@pzwahlen unfortunately I can’t share the sources anymore, as the work was done in the interests of another customer who didn’t consent to share it. sorry.

pzwahlen:
@evbevz Thanks, this is perfectly clear and fine.

PR #1884, opened 2017-10-20
Translate Proxies tabs.

kekek2:


src/opnsense/mvc/app/controllers/OPNsense/Base/ControllerBase.php
```diff
@@ -53,7 +53,11 @@ private function parseFormNode($xmlNode)
                     }
                     $tab = array();
                     $tab[] = $node->attributes()->id;
-                    $tab[] = $node->attributes()->description;
+                    if (function_exists("gettext")) {
```
fichtner: I don't think we need the function_exists() dance. There is one other spot here that had this originally, but it should be removed from there as well. Ok?

fichtner:
merged, thank you!

PR #1889, opened 2017-10-23
 improve sorting of DHCP leases

ljm42:
* replaces strcmp with strnatcasecmp, which is much better at sorting ip addresses
* sorts the list by IP address by default
* does a secondary sort by IP address, which gives better results when sorting by status or lease type
* fixes sorting on the "status" column (the underlying data structure is keyed on "online" instead of "status")
* removes the option to sort by interface, since that doesn't work (the data doesn't exist at the time the array is sorted)

### DIFF
--- a/src/www/status_dhcp_leases.php
+++ b/src/www/status_dhcp_leases.php
@@ -33,11 +33,6 @@ require_once("config.inc");
 require_once("services.inc");
 require_once("interfaces.inc");
 
-function leasecmp($a, $b)
-{
-    return strcmp($a[$_GET['order']], $b[$_GET['order']]);
-}
-
 function adjust_gmt($dt)
 {
     global $config;
@@ -242,9 +237,11 @@ if ($_SERVER['REQUEST_METHOD'] === 'GET') {
         }
     }
 
-    if ($_GET['order']) {
-        usort($leases, "leasecmp");
+    $order = ( $_GET['order'] ) ? $_GET['order'] : 'ip';
+    usort($leases, function ($a, $b) use ($order) {
+        return strnatcasecmp($a[$order].$a[ip],$b[$order].$b[ip]);
     }
+  );
 } elseif ($_SERVER['REQUEST_METHOD'] === 'POST') {
     if (!empty($_POST['deleteip']) && is_ipaddr($_POST['deleteip'])) {
         // delete dhcp lease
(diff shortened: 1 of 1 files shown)


src/www/status_dhcp_leases.php
```diff
@@ -242,9 +237,13 @@ function remove_duplicate($array, $field)
         }
     }
 
-    if ($_GET['order']) {
-        usort($leases, "leasecmp");
-    }
+    $order = ( $_GET['order'] ) ? $_GET['order'] : 'ip';
+    usort($leases, 
+        function ($a, $b) use ($order) {
+            return strnatcasecmp($a[$order].$a[ip], $b[$order].$b[ip]);
```
AdSchellevis: @ljm42 you seem to miss some quotes here ip -> 'ip', I'm not 100% sure we should concat the ip like this either as it might mangle the requested sort order due to the variable length of the fields. a solution could be to use sprintf() to fix the size of the requested column and concat the address (in the same command) then.

ljm42: Oh right, I'll add those quotes.

The concatenation only affects the comparison when the first term is the same, for instance comparing "static192.168.10.87" to "static192.168.10.101". If the first term is different, the concatenation doesn't matter since the comparison will already have been determined.

AdSchellevis: @ljm42 as long as the field length is the same for the first field, the sorting is fine, when it differs it may lead to unexpected results.

ljm42: It doesn't matter though. If you are comparing "aardvark192.168.10.87" to "zebra192.168.10.101", the one that starts with "aardvark" is always going to sort before the one that starts with "zebra", whether or not an IP address is appended. Changing the second option to "zebra<space><space><space>192.168.10.101" will not change the outcome of the comparison.

The only time the IP address has any effect on the sort value is when the first terms are exactly the same. And then it makes a huge impact on the usability of the page, since you have a nicely sorted list instead of a random one.

(hmm.... I suppose if you were comparing "aardvark192.168.10.87" to "aardvarktwo192.168.10.101", then the IP address would have an effect. But this is highly unlikely given the dataset we're working with, and the unexpected result from that one comparison is far less confusing than the completely unsorted listed we have today)

I encourage you to give this change a try, it is a night and day difference in terms of finding a specific IP on the page.

fichtner: the traditional way, since you already have a custom compare function, is to compare individual sections one after the other and return their compare value if not zero otherwise keep comparing :)

ljm42: Thanks. How's it look now?

fichtner: looks good, thanks!

AdSchellevis:
@ljm42 looks good, thanks! I will pull this one in.

PR #1949, opened 2017-11-22
Default form: ability to hide header under 'advanced' tag

evbevz:
For the case where I want to make the settings block "advanced" with the block header.


AdSchellevis:
looks ok, thanks

Issue #1964, opened 2017-11-28
XSS in custom (root) certificate


binaryfigments:
I do not think the impact will be big, it only works with crafted certificates logged in as root.

It works on:
OPNsense 17.1.1-amd64
FreeBSD 11.0-RELEASE-p7
OpenSSL 1.0.2k 26 Jan 2017

And a just complete upgraded firewall installation.

URLs:
https://firewall/system_camanager.php
https://firewall/system_certmanager.php

While adding a custom certificate authority, you can trigger an XSS in the admin panel.

First, I went to Authorities and added the CA. No problems here.
![opnsense0](https://user-images.githubusercontent.com/2205121/33309907-1c5e4698-d420-11e7-86e1-b9569bc2f659.png)

Then I click further to certificates. (for adding a certificate to that fake root.
Here is the first pop-up.

![opnsense1](https://user-images.githubusercontent.com/2205121/33309720-5e43ff18-d41f-11e7-87ef-1fe732b8c0e9.png)

Clicking these away this is the result:

![opnsense2](https://user-images.githubusercontent.com/2205121/33309962-52866610-d420-11e7-819b-048675fea01c.png)

Adding a certificate over there:
![opnsense3](https://user-images.githubusercontent.com/2205121/33310356-93f84e78-d421-11e7-9616-26f64b07da2b.png)

Here are the CA and certificate files I used.
[fakeca.zip](https://github.com/opnsense/core/files/1509040/fakeca.zip)

AdSchellevis:
@binaryfigments thanks for reporting, we will fix this before the next release.

AdSchellevis:
this https://github.com/opnsense/core/commit/1219f2f1db708f089cb5f4d41a70b0717750b11e should do the trick. can you test on your end to be sure?
`
opnsense-patch  1219f2f
`

binaryfigments:
Sure! I'll give it a try.

fabianfrz:
@AdSchellevis maybe there should be a separate channel for those bugs like an email address with a public GPG key to encrypt the mails to make sure there will not be any exploits in the wild using already fixed bugs before the next release. For example this (stored XSS) can be used for privilege escalation.

CC @fichtner

fabianfrz:
@AdSchellevis not completely fixed,

found also an XSS issue in certificate details view:
[...]

PR #1969, opened 2017-11-30
fix: system_certmanager.php xss

fabianfrz:
With this patch, the correct issuer and subject will be shown to the user:

screenshot_20171130_164702

Follow up for #1964


AdSchellevis:
@fabianfrz better output indeed, thanks

Issue #1915, opened 2017-11-07
Feature request: make the "full help" setting "sticky"


ljm42:
As a new user, I find that I am clicking the "full help" button in the upper right corner on nearly every page load.

Can we make that a "sticky" setting? So click "full help" once to have the help text load by default on every page until you click the "full help" button again to turn it off?

This would be an enormous benefit to new users. Experienced users probably never turn help on, but if they do it would only be a single additional click to turn it back off.

ljm42:
Yes, I think a cookie would be fine. The setting doesn't need to be stored in your user profile, if you move between browsers you can deal with turning help on as needed.

fabianfrz:
@AdSchellevis You can use the local storage like for the scroll lock.

AdSchellevis:
@fabianfrz I know, I added that one too ;) Time is the only thing keeping this from happening at the moment.  I can assign myself, eventually this will be there.... The question has been there more often and it's quite practical to have.

AdSchellevis:
In case anyone likes to test, the command below installs the feature (on a recent version of OPNsense)

`opnsense-patch 7e3906a09`

ljm42:
Thanks for this @AdSchellevis it works great! This should be extremely helpful.

fichtner:
Nice, thank you!

fichtner:
@AdSchellevis can we do the same for advanced?

AdSchellevis:
sure, will fix in a minute

AdSchellevis:
@fichtner https://github.com/opnsense/core/commit/2826d6206e0e5bf9d4ebecc323426c2ce80b10aa

fichtner:
perfect, thank you <3

fichtner:
I have to reopen this briefly for a fix (I'll do it after investigation):

https://forum.opnsense.org/index.php?topic=6444.0

Sometimes the stores are deactivated and should therefore not stop all the pages from working. Maybe we can add an if check or something....

fabianfrz:
This is quite strange but the setting `dom.storage.enabled` to false, will cause that `window.localStorage`  will be undefined. A possible fix would be disabling the feature be wrapping it into an if statement looking like this:
```js
if ('localStorage' in window) {
//code
}
```

fichtner:
@fabianfrz if you want to propose a fix please do, that looks promising. Till can help text over at the forum.

fabianfrz:
@fichtner will create a PR which should fix this.

[...]

PR #1972, opened 2017-12-02
fix the localStorage issue from #1915

fabianfrz:
Disables some features to keep the rest working if the browser has some features disabled.

### DIFF
--- a/src/opnsense/www/js/opnsense.js
+++ b/src/opnsense/www/js/opnsense.js
@@ -36,9 +36,24 @@
   * @param value encoded text
   * @return string decoded text
   */
- function htmlDecode(value) {
-     return $("<textarea/>").html(value).text();
- }
+function htmlDecode(value) {
+    return $("<textarea/>").html(value).text();
+}
+ 
+
+/**
+ * Check if sessionStorage is available
+ * 
+ * @return boolean if sessionStorage is available.
+ */
+function checkSessionStorageExists() {
+  if ('sessionStorage' in window) {
+    if (window.sessionStorage != null) {
+      return true;
+    }
+  }
+  return false;
+}
 
  /**
  *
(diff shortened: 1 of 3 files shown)


fichtner:
shouldn't we simply test for defined sessionStorage in place? I don't think the function helper gives us a lot of flexibility

fichtner:
There is also more here... 

```
$ git grep sessionStorage
src/opnsense/www/js/opnsense.js:        sessionStorage.setItem('scrollpos', current_location()+"|"+$(window).scrollTop());
src/opnsense/www/js/opnsense.js:        var scrollpos = sessionStorage.getItem('scrollpos');
src/opnsense/www/js/opnsense_ui.js:            sessionStorage.setItem('all_help_preset', 1);
src/opnsense/www/js/opnsense_ui.js:            sessionStorage.setItem('all_help_preset', 0);
src/opnsense/www/js/opnsense_ui.js:    if (sessionStorage.getItem('all_help_preset') == 1) {
src/opnsense/www/js/opnsense_ui.js:    if (sessionStorage.getItem('show_advanced_preset') == 1) {
src/opnsense/www/js/opnsense_ui.js:            sessionStorage.setItem('show_advanced_preset', 1);
src/opnsense/www/js/opnsense_ui.js:            sessionStorage.setItem('show_advanced_preset', 0);
src/www/head.inc:          sessionStorage.setItem('all_help_preset', 1);
src/www/head.inc:          sessionStorage.setItem('all_help_preset', 0);
src/www/head.inc:      if (sessionStorage.getItem('all_help_preset') == 1) {
```

fabianfrz:
@fichtner it is defined but it is `null`.

src/opnsense/www/js/opnsense.js
```diff
@@ -36,9 +36,24 @@
   * @param value encoded text
   * @return string decoded text
   */
- function htmlDecode(value) {
-     return $("<textarea/>").html(value).text();
- }
+function htmlDecode(value) {
+    return $("<textarea/>").html(value).text();
+}
```
fabianfrz: @fichtner this had an one space indent

fichtner:
I'm really against function to check by calling from multiple files, a portable null check is better as this will definitely break during weird caching or if we make a mistake later during refactor.

fabianfrz:
@fichtner now without the function

AdSchellevis:
@fabianfrz looks good to me too, thanks for fixing!

PR #1998, opened 2017-12-17
Squid Proxy „Before Auth“ is missing in GUI

opnsenseuser:
Sorry for my Bad english. I Even do not know if this is the right Place for my concern.

With the current Version of opnsense its not possible to get Windows Updates working if Proxy transparent Mode is selected. You get an error Message(0x801901f7)

To get this 100% working the only way is to modify the „before auth“ in squid. But this field is not available in opnsense. Is there a change to get this Problem fixed?

Read this:
https://wiki.squid-cache.org/SquidFaq/WindowsUpdate

This is the solution for pfsense. And this works without any problems in pfsense!. I have tested it for one year of using pfsense.
PFsense  -> Menu Package -> Proxy Server -> General Settings -> Show Advanced Options -> "Before Auth": but the code in this field:

code:

acl DiscoverSNIHost at_step SslBump1
acl NoSSLIntercept ssl::server_name_regex microsoft.com                     
acl NoSSLIntercept ssl::server_name_regex .microsoft.com                   
acl NoSSLIntercept ssl::server_name_regex windowsupdate.com
acl NoSSLIntercept ssl::server_name_regex .windowsupdate.com
acl NoSSLIntercept ssl::server_name_regex update.microsoft.com.akadns.net

ssl_bump splice NoSSLIntercept
ssl_bump peek DiscoverSNIHost
ssl_bump bump all

acl BrokenButTrustedServers dstdomain download.microsoft.com
acl BrokenButTrustedServers dstdomain update.microsoft.com
acl BrokenButTrustedServers dstdomain update.microsoft.com.akadns.net
acl BrokenButTrustedServers dstdomain update.microsoft.com.nsatc.net
acl DomainMismatch ssl_error SQUID_X509_V_ERR_DOMAIN_MISMATCH
sslproxy_cert_error allow BrokenButTrustedServers DomainMismatch
sslproxy_cert_error deny all


fabianfrz:
@opnsenseuser Something is probably wrong with your git branch as you are having more than 250 commits on this branch which leads to a not reviewable diff.  **This patch would also allow MITM attacks and therefore should be avoided** (for example it would allow a connection to my server because it has a valid certificate and the only issue is that the domain would not match).

Issue #2004, opened 2017-12-19
Show ! on traffic shaper rules table on invert selection


tkald:
For now rules with invert source / dest selected are indistinguishable in traffic shaper rules table.
Please add exclamation mark infront of source / dest row if either invert rule is selected (as implemented with firewall rules).

fabianfrz:
I made a patch in #2006. You may want to try it.

AdSchellevis:
@fabianfrz sure, thanks again.

PR #2006, opened 2017-12-19
traffic shaper: add a ! for negated fields

fabianfrz:
probably not the cleanest solution but it works. -> ticket #2004 

bildschirmfoto am 2017-12-19 um 19 06 58


AdSchellevis:
@fabianfrz not bad at all, thanks!

fabianfrz:
@AdSchellevis not sure if I can get the json out of it.

AdSchellevis:
@fabianfrz I see what you mean, let me try something

AdSchellevis:
@fabianfrz here you are https://github.com/opnsense/core/commit/78686814a472df4a9cce69ee15b7f31d9cf1ffa2 , you can easily ditch the json conversion because the object is send to the conversion function.

fabianfrz:
@AdSchellevis this will be probably a requirement for the firewall sections too when the migration will start.

AdSchellevis:
@fabianfrz I'm not sure bootgrid will be good enough to restructure the firewall section, but we'll see

PR #2017, opened 2017-12-24
IPv6 Static over parent v4 PPPoE link

marjohn56:
Certain ISP's allow static v6 address assignments even though dhcp6 works too. The only issue with using statics is that the setup for apinger gateway attempts to use the real interface, i.e. igb0, this does not work as the traffic is going via the pppoe link.

This patch addresses that and sets the gateway to correctly use the pppoe link.

This is an optional setting and an additional option has been added to the interfaces as a checkbox to enable/disable this option.

### DIFF
--- a/src/www/interfaces.php
+++ b/src/www/interfaces.php
@@ -365,6 +365,7 @@ if ($_SERVER['REQUEST_METHOD'] === 'GET') {
     $pconfig['dhcp6usev4iface'] = isset($a_interfaces[$if]['dhcp6usev4iface']);
     $pconfig['adv_dhcp6_debug'] = isset($a_interfaces[$if]['adv_dhcp6_debug']);
     $pconfig['track6-prefix-id--hex'] = sprintf("%x", empty($pconfig['track6-prefix-id']) ? 0 :$pconfig['track6-prefix-id']);
+    $pconfig['ipv6usev4iface'] = isset($a_interfaces[$if]['ipv6usev4iface']);
 
     // ipv4 type (from ipaddr)
     if (is_ipaddrv4($pconfig['ipaddr'])) {
@@ -1053,6 +1054,9 @@ if ($_SERVER['REQUEST_METHOD'] === 'GET') {
                     $new_config['subnetv6'] = $pconfig['subnetv6'];
                     if ($pconfig['gatewayv6'] != "none") {
                         $new_config['gatewayv6'] = $pconfig['gatewayv6'];
+                    }
+					if (!empty($pconfig['ipv6usev4iface'])) {
+                        $new_config['ipv6usev4iface'] = true;
                     }
                     break;
                 case "slaac":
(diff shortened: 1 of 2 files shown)


src/www/interfaces.php
```diff
@@ -1053,6 +1054,9 @@ function get_wireless_channel_info($interface) {
                     $new_config['subnetv6'] = $pconfig['subnetv6'];
                     if ($pconfig['gatewayv6'] != "none") {
                         $new_config['gatewayv6'] = $pconfig['gatewayv6'];
+                    }
+					if (!empty($pconfig['ipv6usev4iface'])) {
```
fichtner: looks like tabs here, we try use 4 spaces as per PSR2

marjohn56: Noted.  I'll attempt to find the correct setting in my editor!

AdSchellevis:
@marjohn56 @fichtner wouldn't it be better to move the current dhcp6usev4iface setting (and change its help text)? It kind of suggests doing the same and there are only a few cases where $realv6iface is set to true.

All in interface_bring_down()

https://github.com/opnsense/core/blob/17.7.10/src/etc/inc/interfaces.inc#L1131
https://github.com/opnsense/core/blob/17.7.10/src/etc/inc/interfaces.inc#L1140
https://github.com/opnsense/core/blob/17.7.10/src/etc/inc/interfaces.inc#L1150

marjohn56:
As a new opnsense user whatever you think is best. But as the pppoe link only appears when static IPv6 is selected and use v4 link only appears when dhcp is selected.

I'm not confused, but then I know what it's doing. :)

AdSchellevis:
@marjohn56 I understand, no problem, our aim is to simplify the codebase, hence we try to look a step further every time. Let me propose a fix for you to test.

marjohn56:
What is not wanted is to break a working system. It's an oddity where a user can use static addresses but is using PPPoE to sign in. In our case, IPv6 works perfectly well when set to static, the only thing that does not work is the gateway pinger, it gets set to the physical address where it needs to be routed over pppe0.

As I said, it's an oddity which probably only comes to light when you have an unusual scenario of the type we can use with our ISP.

AdSchellevis:
I understand, but the other setting likely was there for a similar reason.... Just trying to prevent adding more and more features that do (almost) the same.

AdSchellevis:
@marjohn56 this https://github.com/opnsense/core/commit/f722081b8779ac8d183f5b39efd8dd7c80b22ea0 should do the trick, we still have to do some work in the interface area to ditch unneeded code, but that's for 2018

marjohn56:
Cheers, I'll check it out shortly, Christmas day and all that. :)

AdSchellevis:
nu rush, Christmas indeed :)

marjohn56:
OK, that appears to work. Threw me a couple of times as I had to disable and re-enable the wan and re-save the gateway info, but it's all up and running now.

Nice one... now onto the next problem. :)

fichtner:
@AdSchellevis @marjohn56 looks good too and thanks for testing. closing this one. :)

kwillers:
Tested here and working well. Nice job

PR #2018, opened 2017-12-25
Add PAC support to proxy

fabianfrz:
Add PAC file support to the UI to allow Proxy autoconfiguration of some clients that support that standard


src/opnsense/mvc/app/controllers/OPNsense/Proxy/forms/dialogEditBlacklist.xml
```diff
@@ -32,7 +32,6 @@
     <field>
         <id>blacklist.filter</id>
         <label>categories (if available)</label>
-        <type>text</type>
```
fichtner: maybe we should break this out and merge it immediately?

fabianfrz: It is not important at the moment - it is overwritten in the next line but it should be removed in the future as it might cause problems in the future (if for some reasons the other element is used).

fabianfrz:
@fichtner the only thing still missing is a template reload button on all three tabs.

src/opnsense/mvc/app/views/OPNsense/Proxy/index.volt
```diff
@@ -466,7 +466,6 @@
                             <td></td>
                             <td>
                                 <button data-action="add" type="button" class="btn btn-xs btn-default"><span class="fa fa-plus"></span></button>
-                                <button data-action="deleteSelected" type="button" class="btn btn-xs btn-default"><span class="fa fa-trash-o"></span></button>
```
fabianfrz: @fichtner this button did nothing so I removed it

fabianfrz:
@fichtner the model needs some improvements regarding to validation but the rest seems good now.

src/opnsense/mvc/app/views/OPNsense/Proxy/index.volt
```diff
@@ -230,7 +447,7 @@
             <tr>
                 <td colspan="2">
                     <div id="remoteACLchangeMessage" class="alert alert-info" style="display: none" role="alert">
-                        {{ lang._('Note: after changing categories, please remember to download the ACL again to apply your new settings') }}
+                        {{ lang._('Note: After changing categories, please remember to download the ACL again to apply your new settings') }}
```
fichtner: Better to ditch „Note: „ prefix if this is changed.

src/opnsense/mvc/app/views/OPNsense/Proxy/index.volt
fichtner: Ok if @AdSchellevis doesn’t want to keep it

src/opnsense/service/templates/OPNsense/Proxy/+TARGETS
```diff
@@ -8,3 +8,4 @@ post-auth.conf:/usr/local/etc/squid/post-auth/dummy.conf
 pre-auth.conf:/usr/local/etc/squid/pre-auth/dummy.conf
 rc.conf.d:/etc/rc.conf.d/squid/squid
 squid.conf:/usr/local/etc/squid/squid.conf
+wpad.dat:/usr/local/www/wpad.dat
```
fichtner: It’s a bit strange to drop files into www but I know what you are after

fabianfrz: Do you have a better idea? The URL does not really match the pattern of standard OPNsense rules. Otherwise we would have to heavily rewrite routes in lighttpd.

src/opnsense/mvc/app/views/OPNsense/Proxy/index.volt
AdSchellevis: @fabianfrz @fichtner it can go, I guess someone copied it in there a long time ago....

src/opnsense/service/templates/OPNsense/Proxy/+TARGETS
fichtner: No, I’m only raising awareness. In any case this will not merge before stable/18.1 is branched so we have time to think about this. For now this is good enough for -devel

fabianfrz: the only thing I could think of is adding a custom rewrite to a controller without authentication.

fabianfrz:
@fichtner it's time for a test run ;)

src/opnsense/mvc/app/views/OPNsense/Proxy/index.volt
fabianfrz: @AdSchellevis It did probably come from IDS because it is the only view with a list containing checkable checkboxes.

fabianfrz:
@fichtner the generated JS is valid and it looks good. It may be published in `devel`.

fichtner:
Not yet, or else this lands in 18.1-RC1...

fabianfrz:
fixed conflict

fabianfrz:
@fichtner I found a small GUI issue - maybe you have an idea: Some fields will be hidden when the modal shows up: https://github.com/opnsense/core/pull/2018/files#diff-23d36dcc2f04f176686a881d17c288bdR190

so the event will be emitted here for example where the data should already be mapped: https://github.com/opnsense/core/blob/master/src/opnsense/www/js/opnsense_bootgrid_plugin.js#L217
However it seems $.val() may return a wrong value as the form may show an incorrect dependent field (Some kind of a race condition?)

fabianfrz:
@fichtner the issue was fixed a week ago

fabianfrz:
ping

Additional PRs:
* #2097 
* #2088

fichtner:
Let me finish 18.1.6 for new images, then we pull this in. Is everything ready from your end?

fichtner:
(Your PR summary says otherwise)

fabianfrz:
@fichtner It was for a long time as I can remember. This pull request only adds a template to Squid to allow to create a PAC file as OPNSENSE-INTERFACE-HOST/wpad.dat which can be used to automatically configure the proxy in Firefox, Windows, Atom and so on. The Other Pull requests are more simple - they only add the WPAD DNS-Records and DHCP Options. I will give this branch a try again.

fabianfrz:
found a bug but the rest looks good.

fabianfrz:
fixed conflict of "+Targets" via GitHub online editor.

fichtner:
@fabianfrz will you rebase / fix conflict? thanks

fabianfrz:
I hope the github merge did what it was intended to do.

src/opnsense/mvc/app/controllers/OPNsense/Proxy/forms/dialogEditBlacklist.xml
fichtner: merge glitch

fichtner:
Merged, thanks!

PR #2019, opened 2017-12-26
Full IDN support from SmartSoft

kekek2:


AdSchellevis:
@kekek2 this seems to interfere with the current changes we have in stock for the new alias processing, we definitely don't want to insert the host aliases back as they where.

PR #2021, opened 2017-12-26
Fixed DUID

marjohn56:
Allow the user to fix the dhcp6 duid.

The existing DUID may be recalled and stored in the config file or a new DUID may be created.

On the call to function interface_dhcp6_configure the existince of a config variable 'system''global-v6duid' is checked, if it exists then the duid file is written before dhcp6c starts.

The loss of DUID on reboot if using RAM disk or the loss when upgrading or rebuilding is therefore mitigated.

There is one issue I cannot resolve, it's probably simple but I cannot see what I am missing.

The call $input_errors[] = gettext("A valid DUID must be specified"); at line 85 in system_advanced_network.php does not work. This should flag an error if the user inputs an invalid DUID, but it does nothing.

Much of this is taken from the darkside, but as I wrote it in the first place I have no qualms!

### DIFF
--- a/src/etc/inc/interfaces.inc
+++ b/src/etc/inc/interfaces.inc
@@ -3111,6 +3111,13 @@ function interface_dhcpv6_configure($interface = 'wan', $wancfg)
     $wanif = get_real_interface($interface, "inet6");
     $dhcp6cconf = "";
     $dhcp6cconf .= "interface {$wanif} {\n";
+    
+    if (!empty($config['system']['global-v6duid'])) {
+        // Write the DUID file
+        if(!write_dhcp6_duid($config['system']['global-v6duid'])) {
+            log_error(gettext("Failed to write user DUID file!"));
+		}
+    }
 
     /* for SLAAC interfaces we do fire off a dhcp6 client for just our name servers */
     if ($wancfg['ipaddrv6'] == "slaac") {
--- a/src/etc/inc/util.inc
+++ b/src/etc/inc/util.inc
@@ -1370,3 +1370,92 @@ function is_install_media()
 
     return true;
 }
+/* format a string to look (more) like the expected DUID format:
+ * 1) Replace any "-" with ":"
+ * 2) If the user inputs 14 components, then add the expected "0e:00:" to the front.
+ *    This is convenience, because the actual DUID (which is reported in logs) is the last 14 components.
+ * 3) If any components are input with just a single char (hex digit hopefully), put a "0" in front.
+ *
+ * The final result should be closer to:
+ *
+ * "0e:00:00:01:00:01:nn:nn:nn:nn:nn:nn:nn:nn:nn:nn"
+ *
+ * This function does not validate the input. is_duid() will do validation.
+*/
+function format_duid($dhcp6duid) {
+    $values = explode(":", strtolower(str_replace("-", ":", $dhcp6duid)));
+    if (count($values) == 14) {
+        array_unshift($values, "0e", "00");
+    }
+
+    array_walk($values, function(&$value) {
+        $value = str_pad($value, 2, '0', STR_PAD_LEFT);
+    });
+
+    return implode(":", $values);
+}
+
+/* returns true if $dhcp6duid is a valid duid entry */
+function is_duid($dhcp6duid) {
+    $values = explode(":", $dhcp6duid);
+    if (count($values) != 16 || strlen($dhcp6duid) != 47) {
+        return false;
+    }
+    for ($i = 0; $i < 16; $i++) {
+        if (ctype_xdigit($values[$i]) == false)
+            return false;
+        if (hexdec($values[$i]) < 0 || hexdec($values[$i]) > 255)
+            return false;
+    }
+    return true;
+}
+
+/* Write the DHCP6 DUID file */
+function write_dhcp6_duid($duidstring) {
+    // Create the hex array from the dhcp6duid config entry and write to file
+
+    if(!is_duid($duidstring)) {
+        log_error(gettext("Error: attempting to write DUID file - Invalid DUID detected"));
+        return false;
+    }
+    $temp = str_replace(":","",$duidstring);
+    $duid_binstring = pack("H*",$temp);
+    if ($fd = fopen("/var/db//dhcp6c_duid", "wb")) {
+        fwrite($fd, $duid_binstring);
+        fclose($fd);
+        return true;
+    }
+    log_error(gettext("Error: attempting to write DUID file - File write error"));
+    return false;
+}
+
+/* returns duid string from 'vardb_path']}/dhcp6c_duid' */
+function get_duid_from_file() {
+
+    $duid_ASCII = "";
+    $count = 0;
+
+    if (file_exists("/var/db//dhcp6c_duid") &&
+        ($fd = fopen("/var/db/dhcp6c_duid", "r"))) {
+        if(filesize("/var/db//dhcp6c_duid")==16) {
+            $buffer = fread($fd,16);
+            while($count < 16) {
+            $duid_ASCII .= bin2hex($buffer[$count]);
+                $count++;
+                if($count < 16) {
+                    $duid_ASCII .= ":";
+                }
+            }
+        }
+        fclose($fd);
+    }
+    else
+    {
+        log_error(gettext("Error: Could not open /var/db/dhcp6c_duid"));
+    }
+	//if no file or error with read then the string returns blanked DUID string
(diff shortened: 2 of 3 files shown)


src/www/system_advanced_network.php
```diff
@@ -170,7 +180,35 @@
...
+                <td width="22%"><strong><?= gettext('DHCP6c DUID') ?></strong></td>
+                <td width="78%" align="right">
+                  <small><?=gettext("full help"); ?> </small>
+                  <i class="fa fa-toggle-off text-danger"  style="cursor: pointer;" id="show_all_help_page" type="button"></i>
+                </td>
+              </tr>
+              <tr>
+                  <td><a id="help_for_fixed_DUID" href="#" class="showhelp"><i class="fa fa-info-circle"></i></a> <?=gettext("Fixed DUID"); ?></td>
+                  <td>
+                    <input name="global-v6duid" type="text" id="global-v6duid" value="<?=htmlspecialchars($pconfig['global-v6duid']);?>" />
```
fabianfrz: html_safe wraps htmlspecialchars

src/www/system_advanced_network.php
```diff
@@ -170,7 +180,35 @@
...
+                      <?= gettext('This field can be used to enter a fixed DUID for use by dhcp6c. The DUID ' .
+                          'may change when the system is rebooted if using RAM disks, it may also change ' .
+                          'when a firmware update takes place. Entering the DUID here will store it in ' .
+                          'in the configuration file.') ?><br />
+<?php
+            
+                      $duid = get_duid_from_file();
+                      if (!empty($duid)):
+?>
+                      <a onclick="document.getElementById('global-v6duid').value='<?= html_safe($duid) ?>';" href="#"><?=gettext("Insert the existing DUID here"); ?></a><br />
```
fabianfrz: may be shorter using jQuery: 
```js
$('#global-v6duid').val('<?= html_safe($duid) ?>');
```

src/etc/inc/util.inc
```diff
@@ -1370,3 +1370,92 @@ function is_install_media()
...
+function write_dhcp6_duid($duidstring) {
+    // Create the hex array from the dhcp6duid config entry and write to file
+
+    if(!is_duid($duidstring)) {
+        log_error(gettext("Error: attempting to write DUID file - Invalid DUID detected"));
+        return false;
+    }
+    $temp = str_replace(":","",$duidstring);
+    $duid_binstring = pack("H*",$temp);
+    if ($fd = fopen("/var/db//dhcp6c_duid", "wb")) {
```
fabianfrz: "//"?

fichtner: It was mentioned in the forum that this is bad for /var MFS. The directory is kinda silly, it would be better to have a /var/db/dhcp6c/duid structure so we could add /var/db/dhcp6c to the known MSF dirs, we do the same to make the package database persistent.

src/www/system_advanced_network.php
fichtner: All pconfig values are escaped by default

src/etc/inc/interfaces.inc
```diff
@@ -3111,6 +3111,13 @@ function interface_dhcpv6_configure($interface = 'wan', $wancfg)
     $wanif = get_real_interface($interface, "inet6");
     $dhcp6cconf = "";
     $dhcp6cconf .= "interface {$wanif} {\n";
+    
+    if (!empty($config['system']['global-v6duid'])) {
```
fichtner: Not sure about “global-“ suffix for a global value. How about ipv6duid ?

src/etc/inc/interfaces.inc
```diff
@@ -3111,6 +3111,13 @@ function interface_dhcpv6_configure($interface = 'wan', $wancfg)
...
+        // Write the DUID file
+        if(!write_dhcp6_duid($config['system']['global-v6duid'])) {
+            log_error(gettext("Failed to write user DUID file!"));
```
fichtner: No gettext() for log_error() it makes debugging hard in translations.

src/etc/inc/interfaces.inc
```diff
@@ -3111,6 +3111,13 @@ function interface_dhcpv6_configure($interface = 'wan', $wancfg)
     $wanif = get_real_interface($interface, "inet6");
     $dhcp6cconf = "";
     $dhcp6cconf .= "interface {$wanif} {\n";
+    
+    if (!empty($config['system']['global-v6duid'])) {
+        // Write the DUID file
+        if(!write_dhcp6_duid($config['system']['global-v6duid'])) {
```
fichtner: Code style: “if (“, beware of tabs as well in this area.

src/etc/inc/util.inc
```diff
@@ -1370,3 +1370,92 @@ function is_install_media()
...
+ *    This is convenience, because the actual DUID (which is reported in logs) is the last 14 components.
+ * 3) If any components are input with just a single char (hex digit hopefully), put a "0" in front.
+ *
+ * The final result should be closer to:
+ *
+ * "0e:00:00:01:00:01:nn:nn:nn:nn:nn:nn:nn:nn:nn:nn"
+ *
+ * This function does not validate the input. is_duid() will do validation.
+*/
+function format_duid($dhcp6duid) {
```
fichtner: Can this functions be private?

src/etc/inc/util.inc
```diff
@@ -1370,3 +1370,92 @@ function is_install_media()
...
+
+    array_walk($values, function(&$value) {
+        $value = str_pad($value, 2, '0', STR_PAD_LEFT);
+    });
+
+    return implode(":", $values);
+}
+
+/* returns true if $dhcp6duid is a valid duid entry */
+function is_duid($dhcp6duid) {
```
fichtner: Private function or push to gui if only for validation

src/etc/inc/util.inc
```diff
@@ -1370,3 +1370,92 @@ function is_install_media()
...
+            return false;
+        if (hexdec($values[$i]) < 0 || hexdec($values[$i]) > 255)
+            return false;
+    }
+    return true;
+}
+
+/* Write the DHCP6 DUID file */
+function write_dhcp6_duid($duidstring) {
+    // Create the hex array from the dhcp6duid config entry and write to file
```
fichtner: Should be in interfaces.inc as dhcp6c belongs there, like “interfaces_dhcp6_duid_write” or better yet inline into dhcp6c config code

marjohn56: Whatever you think is best.

src/www/system_advanced_network.php
marjohn56: I borrowed that bit from the existing spoofmac code.

src/etc/inc/interfaces.inc
marjohn56: I guess because when it was discusses elsewhere there was a lot of discussion of to where to put the DUID setting, some were unaware that you can only run one instance of dhcp6c and therefore the DUID is in fact global. It is whatever you wish to call it. :)

fichtner:
If you want I can propose a rework on top of your changes. I feel we should add and use duid getter and setter for configd. It’s a bit of a sticky situation here to get all moving parts act according to new and even future use.

I’ll propose a config mal after skimming similar code and will ask for your ok. Ok? :)

marjohn56:
Franco I will learn from you, whatever you see fit to do then do it. 

I will then look at your code and learn from it. :)

fichtner:
@marjohn56 

there is no right or wrong really, just discussion for clarity and future stability

$input_errors didn't work because the page had no validation being checkboxes/preselect only.

I'll propose a commit shortly for you to ACK or NAK :)

fichtner:
@marjohn56 how do you feel about this one? https://github.com/opnsense/core/commit/945bc19e7

The impact has been minimised by only touching interfaces.inc and the network settings, usual label simplification, two small issues in the page (the missing input validation for your setting and also the faulty use of $_POST) and the addition of removing /var/db/dhcp6c_duid file when the DUID is cleared on save.

One thing that I'd like to do is store the DUID in the /conf directory during reboot to prevent the /var MFS splatter that you mentioned. That should make a good solid feature. :)

fichtner:
Ok, here we go... 533ccbde0c867 and https://github.com/opnsense/bsdinstaller/commit/e6441137cb0a

kwillers:
Thanks Franco 
MarJohn is busy with Christmas family stuff today so ask me to let you know it may be tomorrow before he can get to this

fichtner:
No rush guys, enjoy the holidays. :)

PR #2024, opened 2017-12-26
dhcp6c No Release Option

marjohn56:
dhcp6c sends a release signal on exit by default. This can cause the loss of the allocated prefix or address and a new one on the next connection. This is the case with many ISPs especially Sky UK.

This option allows the user to disable the release send.

I added the  -n option to the FreeBSD dhcp6c quite a while back.

### DIFF
--- a/src/etc/inc/interfaces.inc
+++ b/src/etc/inc/interfaces.inc
@@ -3195,10 +3195,13 @@ function interface_dhcpv6_configure($interface = 'wan', $wancfg)
     unset($dhcp6cscript);
     chmod("/var/etc/dhcp6c_{$interface}_script.sh", 0755);
 
+    $noreleaseoption = isset($wancfg['dhcp6norelease']) ? "-n" : "";
+    
     $dhcp6ccommand = exec_safe(
-        "/usr/local/sbin/dhcp6c %s -c %s -p %s %s",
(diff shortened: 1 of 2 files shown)


src/etc/inc/interfaces.inc
```diff
@@ -3195,10 +3195,13 @@ function interface_dhcpv6_configure($interface = 'wan', $wancfg)
...
     chmod("/var/etc/dhcp6c_{$interface}_script.sh", 0755);
 
+    $noreleaseoption = isset($wancfg['dhcp6norelease']) ? "-n" : "";
+    
     $dhcp6ccommand = exec_safe(
-        "/usr/local/sbin/dhcp6c %s -c %s -p %s %s",
+        "/usr/local/sbin/dhcp6c %s %s -c %s -p %s %s",
         array(
             empty($wancfg['adv_dhcp6_debug']) ? '-d' : '-D',
+            "{$noreleaseoption},
```
fabianfrz: something went wrong here...

marjohn56:
What the f***...

Hold on, looks like copy did not work, my appols!

marjohn56:
Ok, that was stupid.... fixed.

fichtner:
@marjohn56 can you please reuse your PR branches? GitHub supports this, we don’t mind squash and (force) push :)

PR #2025, opened 2017-12-26
dhcp6c No Release Option

marjohn56:
dhcp6c sends a release signal on exit by default. This can cause the loss of the allocated prefix or address and a new one on the next connection. This is especially the case with many ISPs.

This option allows the user to disable the release send.

I added the  -n option to the FreeBSD dhcp6c quite a while back.

Sorry for the previous PR, copy paste error!


fichtner:
@marjohn56 thanks! Small tweaks via https://github.com/opnsense/core/commit/6c13cbf4dd

PR #2029, opened 2017-12-28
DHCP6c Uses REASONS to call newipv6

marjohn56:
Use the REASONS given by dhcp6c when it calls its script. This then only calls newipv6 when 'REQUEST' is the reason.

RENEW, REBIND or INFO do not have changes to the leases, therefore no call to update is needed and this prevents reloading which can affect VPNs etc.

### DIFF
--- a/src/etc/inc/interfaces.inc
+++ b/src/etc/inc/interfaces.inc
@@ -3195,8 +3195,38 @@ function interface_dhcpv6_configure($interface = 'wan', $wancfg)
     }
     unset($dhcp6cconf);
 
+    /* Use the REASONS given by dhcp6c when it calls its script. This then only calls newipv6 when 'REQUEST' is the reason. */
+    /* RENEW, REBIND or INFO do not have changes to the leases, therefore no call to update is needed and this prevents */
+    /* reloading which can affect VPNs etc */
+
     $dhcp6cscript = "#!/bin/sh\n";
+    $dhcp6cscript .= "case \$REASON in\n";
+    $dhcp6cscript .= "REQUEST)\n";
     $dhcp6cscript .= "/usr/local/opnsense/service/configd_ctl.py interface newipv6 {$wanif}\n";
+    if (empty($wancfg['adv_dhcp6_debug') {
+        $dhcp6cscript .= "/usr/bin/logger -t dhcp6c \"dhcp6c REQUEST on {$wanif} - running newipv6\"\n";
+    }
+    $dhcp6cscript .= ";;\n";
+    $dhcp6cscript .= "REBIND)\n";
+    if (empty($wancfg['adv_dhcp6_debug') {
+        $dhcp6cscript .= "/usr/bin/logger -t dhcp6c \"dhcp6c rebind on {$wanif}\"\n";
+    }
+    $dhcp6cscript .= ";;\n";
+    if (isset($wancfg['dhcp6norelease'])) {
+        $dhcp6cscript .= "EXIT)\n";
+    } else {
+        $dhcp6cscript .= "RELEASE)\n";
+    }
+    if (empty($wancfg['adv_dhcp6_debug') {
+        $dhcp6cscript .= "/usr/bin/logger -t dhcp6c \"dhcp6c EXIT or RELEASE on {$wanif} running newipv6\"\n";
+    }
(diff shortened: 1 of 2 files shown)


marjohn56:
Forgot to add that the REASONS EXIT and RELEASE also call newipv6 as the lease will have been removed from the interface.

fichtner:
@marjohn56 please update your initial github branch on changes, it's easier to track changes that way instead of shifting PRs :)

src/etc/inc/interfaces.inc
```diff
@@ -3195,8 +3195,39 @@ function interface_dhcpv6_configure($interface = 'wan', $wancfg)
...
+    $dhcp6cscript .= ";;\n";
+    if (isset($wancfg['dhcp6norelease'])) {
+        $dhcp6cscript .= "EXIT)\n";
+    } else {
+        $dhcp6cscript .= "RELEASE)\n";
+    }
+    if ($debugOption == '-D') {
+        $dhcp6cscript .= "/usr/bin/logger -t dhcp6c \"dhcp6c EXIT or RELEASE on {$wanif} running newipv6\"\n";
+    }
+    $dhcp6cscript .= "/usr/local/sbin/fcgicli -f /etc/rc.newwanipv6 -d \"interface={$wanif}&dmnames=\${dmnames}&dmips=\${dmips}\"\n";
```
fichtner: not sure this works, should probably be like above using configd_ctl.py

src/etc/inc/interfaces.inc
```diff
@@ -3195,8 +3195,39 @@ function interface_dhcpv6_configure($interface = 'wan', $wancfg)
...
+    /* Use the REASONS given by dhcp6c when it calls its script. This then only calls newipv6 when 'REQUEST' is the reason. */
+    /* RENEW, REBIND or INFO do not have changes to the leases, therefore no call to update is needed and this prevents */
+    /* reloading which can affect VPNs etc */
+    $debugOption = isset($wancfg['adv_dhcp6_debug']) ? "-D" : "-d";
+    $noreleaseOption = isset($wancfg['dhcp6norelease']) ? "-n" : "";
     $dhcp6cscript = "#!/bin/sh\n";
+    $dhcp6cscript .= "case \$REASON in\n";
+    $dhcp6cscript .= "REQUEST)\n";
     $dhcp6cscript .= "/usr/local/opnsense/service/configd_ctl.py interface newipv6 {$wanif}\n";
+    if ($debugOption == '-D') {
```
fichtner: you can probably leave out the '-X' assignments and use isset() exclusively, folding this into if ($debugOption) etc.

src/etc/inc/interfaces.inc
```diff
@@ -3195,8 +3195,39 @@ function interface_dhcpv6_configure($interface = 'wan', $wancfg)
...
+        $dhcp6cscript .= "/usr/bin/logger -t dhcp6c \"dhcp6c REQUEST on {$wanif} - running newipv6\"\n";
+    }
+    $dhcp6cscript .= ";;\n";
+    $dhcp6cscript .= "REBIND)\n";
+    if ($debugOption == '-D') {
+        $dhcp6cscript .= "/usr/bin/logger -t dhcp6c \"dhcp6c rebind on {$wanif}\"\n";
+    }
+    $dhcp6cscript .= ";;\n";
+    if (isset($wancfg['dhcp6norelease'])) {
```
fichtner: $noreleaseOption is unused because here it's correctly tested against already :)

marjohn56:
Hold on... I'm having issues with my Github Desktop  (Windows) . It's driving me nuts today, that's why so many cancelled PR's, For some unknown reason it keeps either pulling a commit I've already reversed or using an old file...

There are days when life is just tedious..

src/etc/inc/interfaces.inc
marjohn56: It should not, no... I'l post a a change in a mo, that's after I've thrown this PC out of the window and started again... sigh.

src/etc/inc/interfaces.inc
marjohn56: OK, strange this all got through p****** when I write it there. :)

marjohn56:
@fichtner  Really sorry,  my GitHub Desktop has now gone completely bananas, now getting fatal error, I'll need to delete the branch and start again. I may just do a full clone and start from there....

fichtner:
Alright, no worries.

PR #2030, opened 2017-12-28
dhcp6c - Use Reasons

marjohn56:
Use the REASONS given by dhcp6c when it calls its script. This then only calls newipv6 when 'REQUEST' is the reason.

RENEW, REBIND or INFO do not have changes to the leases, therefore no call to update is needed and this prevents reloading which can affect VPNs, Unbound etc.

On RELEASE or EXIT, once again newipv6 is called as the lease addresses should have been removed by dhcp6c.

### DIFF
--- a/src/etc/inc/interfaces.inc
+++ b/src/etc/inc/interfaces.inc
@@ -3196,7 +3196,33 @@ function interface_dhcpv6_configure($interface = 'wan', $wancfg)
     unset($dhcp6cconf);
 
     $dhcp6cscript = "#!/bin/sh\n";
+    $dhcp6cscript .= "case \$REASON in\n";
+    $dhcp6cscript .= "REQUEST)\n";
     $dhcp6cscript .= "/usr/local/opnsense/service/configd_ctl.py interface newipv6 {$wanif}\n";
+    if (isset($wancfg['adv_dhcp6_debug'])) {
+        $dhcp6cscript .= "/usr/bin/logger -t dhcp6c \"dhcp6c REQUEST on {$wanif} - running newipv6\"\n";
+    }
+    $dhcp6cscript .= ";;\n";
+    $dhcp6cscript .= "REBIND)\n";
+    if (!empty($wancfg['adv_dhcp6_debug'])) {
+        $dhcp6cscript .= "/usr/bin/logger -t dhcp6c \"dhcp6c rebind on {$wanif}\"\n";
+    }
+    $dhcp6cscript .= ";;\n";
+    if (!empty($wancfg['dhcp6norelease'])) {
+        $dhcp6cscript .= "EXIT)\n";
+    } else {
+        $dhcp6cscript .= "RELEASE)\n";
+    }
+    if (!empty($wancfg['adv_dhcp6_debug'])) {
+        $dhcp6cscript .= "/usr/bin/logger -t dhcp6c \"dhcp6c EXIT or RELEASE on {$wanif} running newipv6\"\n";
+    }
+    $dhcp6cscript .= "/usr/local/opnsense/service/configd_ctl.py interface newipv6 {$wanif}\n";
+    $dhcp6cscript .= ";;\n";
+    $dhcp6cscript .= "RENEW|INFO)\n";
+    if (!empty($wancfg['adv_dhcp6_debug'])) {
+        $dhcp6cscript .= "/usr/bin/logger -t dhcp6c \"dhcp6c renew, no change - bypassing update on {$wanif}\"\n";
+    }
+    $dhcp6cscript .= "esac\n";
 
     if (!@file_put_contents("/var/etc/dhcp6c_{$interface}_script.sh", $dhcp6cscript)) {
         printf("Error: cannot open dhcp6c_{$interface}_script.sh in interface_dhcpv6_configure() for writing.\n");


marjohn56:
New Hard Drive,  let's see if it's more reliable now!

src/etc/inc/interfaces.inc
```diff
@@ -3196,7 +3196,35 @@ function interface_dhcpv6_configure($interface = 'wan', $wancfg)
     unset($dhcp6cconf);
 
     $dhcp6cscript = "#!/bin/sh\n";
+    $dhcp6cscript .= "dmips=\${new_domain_name_servers}\n";
```
fichtner: unused

src/etc/inc/interfaces.inc
```diff
@@ -3196,7 +3196,35 @@ function interface_dhcpv6_configure($interface = 'wan', $wancfg)
...
+    $dhcp6cscript .= "dmnames=\${new_domain_name}\n";
```
fichtner: unused :)

marjohn56: Yes, sorry I was playing around with those, forgot they were still there...

src/etc/inc/interfaces.inc
```diff
@@ -3196,7 +3196,35 @@ function interface_dhcpv6_configure($interface = 'wan', $wancfg)
...
+    if (isset($wancfg['adv_dhcp6_debug'])) {
+        $dhcp6cscript .= "/usr/bin/logger -t dhcp6c \"dhcp6c REQUEST on {$wanif} - running newipv6\"\n";
+    }
+    $dhcp6cscript .= ";;\n";
+    $dhcp6cscript .= "REBIND)\n";
+    if (!empty($wancfg['adv_dhcp6_debug'])) {
+        $dhcp6cscript .= "/usr/bin/logger -t dhcp6c \"dhcp6c rebind on {$wanif}\"\n";
+    }
+    $dhcp6cscript .= ";;\n";
+    if (!empty($wancfg['dhcp6norelease'])) {
```
fichtner: what's the logic here? it flips EXIT to RELEASE but wouldn't it make more sense to omit the block when dhcp6norelease is set?

src/etc/inc/interfaces.inc
marjohn56: I will attempt to remove them and issue a new commit, this is where it all went pear shaped last time.... here goes.

src/etc/inc/interfaces.inc
marjohn56: No, even if dhcp6c exits with the no release option, it still exits and removes the leases, but the reason is different. Makes sense to catch it either way. You could just have both in there, but I was trying to reduce code when I wrote the original for p*****e a while back.

fichtner: Ah, so in a nutshell we want to do updates under more restrictive circumstances?

marjohn56:
Ok, that worked, should have removed the offending articles.   :)

src/etc/inc/interfaces.inc
marjohn56: Yes, it also creates a framework for what may happen in future.

I am still working my way through this area of code, it was ripe for issues with dhcp6c, under p*****e and the pitfalls are here too, well they would be, :) 

I fairly easily managed to get two instances of dhpc6c running, I and my testers spent many happy hours/days resolving the issues with dhcp6c, however I assume that this section is up for a complete re-write? 

If you want me to, I can easily recreate the same things that were done on p******, I have no qualms about using code I originally wrote.

marjohn56: Ah, I misunderstood your query about exit or release, hmm, I'm trying to remember why that was done, because  its easier to use an or like in the case of info RENEW|INFO. Looking at it now I can see no reason at all to do that, either will result in the same action..

I think it was just down to testing, I have just looked at some old interfaces.inc files and there were just more debug messages while we ( my testers ) were working out what would happen with various responses. 

I'll clean that bit up too,

fichtner: Alright, thanks for explaining. :) In that case "\$REASON" could be used to condense the debug messages. I'll take a closer look next week.

marjohn56: Yes. :) 

I thought of just doing a single if(dbeg-blah) log reason is {$REASON}
but there was a reason I didn't, now I cannot remember what it was.

Having issues trying to get this dhcp6 VLAN prio to work, but that's another topic.

fichtner: hehe, if it comes back to memory just let me know

I've added a side note to that topic, not sure if useful https://forum.opnsense.org/index.php?topic=6714.0

fichtner:
Thanks, as usual, a bit of cleanup follows.

fichtner:
https://github.com/opnsense/core/commit/b5f0902407c52ecead0c8a8d613e7a350b6e665c

I have a small question regarding RELEASE and EXIT flipping: shouldn't we always adhere to RELEASE and only add EXIT to the mix if we have norelease unset?

fichtner:
@marjohn56 after testing this at the office for the better part of a week it looks good. let me know if anything is not as you expect.

marjohn56:
:) I rolled out the changes to one of my testers on his 17,7.11 system, he was delighted so I guess it's all good. Regarding the EXIT,  no keep it. There's a reason for that that will come to light shortly, but basically we can use a usr1 signal I implemented in dhcp6c to overide the unset no-release option and force dhcp6c to exit with no- release. Why you may ask.... because we can. :)

PR #2033, opened 2017-12-30
DHCPv6 Vlan priority

marjohn56:
Certain ISP's, Orange France for example use a VLAN, it is a requirement that the dhcpv6 packets are sent with a different priority, in Orange Frances case it is 6.

This PR allows the user to select whether or not to enable the setting of the VLAN priority for dhcpv6 packets and the value to use.

The is then used in the filter.lib.inc to add the extra set-prio value when the fillter is created.

This PR requires the the sysctl net.link.vlan.mtag_pcp be set to 1.

### DIFF
--- a/src/etc/inc/filter.lib.inc
+++ b/src/etc/inc/filter.lib.inc
@@ -289,11 +289,20 @@ function filter_core_rules_system($fw, $defaults)
                     'interface' => $intf, 'label' =>'allow dhcpv6 client in ' . $intfinfo['descr']),
               $defaults['pass']
             );
-            $fw->registerFilterRule(1,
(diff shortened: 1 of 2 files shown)


fichtner:
Maybe it's better here to have a "do not modify" default setting (empty) and otherwise set 0-7 :)

marjohn56:
Indeed... Now why didn't I think of that. :)

fichtner:
@marjohn56 thanks, I've pulled this in and will rework as suggested

fichtner:
https://github.com/opnsense/core/commit/ba725881f895db5908dee0ac36c07748229af917
https://github.com/opnsense/core/commit/617dcaee88c11eedcd91e9f783048257bf2db566

Now if we have the same for dhclient on IPv4 we are almost done :)

PR #2039, opened 2018-01-01
Eliminate 300 ms Tap Delay

NOYB:
Implement touch-action: manipulation CSS property to eliminate 300 ms tap delay.

Does not prevent zoom.  Zoom still available by pinch.  Only prevents double tap zoom in order to eliminate the 300 ms tap delay.

What Exactly Is..... The 300ms Click Delay
https://www.telerik.com/blogs/what-exactly-is.....-the-300ms-click-delay

Mozilla developers page.  Updated Oct 25, 2017.
https://developer.mozilla.org/en-US/docs/Web/CSS/touch-action

W3C Recommendation - The touch-action CSS property
http://www.w3.org/TR/pointerevents/#the-touch-action-css-property

### DIFF
--- a/src/opnsense/www/themes/opnsense/assets/stylesheets/main.scss
+++ b/src/opnsense/www/themes/opnsense/assets/stylesheets/main.scss
@@ -121,6 +121,7 @@ html, body{
 }
 
 body{
+  touch-action: manipulation;
   min-width: 320px;
 }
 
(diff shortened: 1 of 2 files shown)


fichtner:
this requires update of the main.scss file as well, as main.css is generated from there and this change could end up being overwritten

fichtner:
@NOYB oh, please ping when pushing, did not get a github notice for the update.

I'll pull this in now, thank you!

PR #2040, opened 2018-01-02
Allow input fields to be rendered as readonly

dharrigan:
Hi,

Allow input fields to be rendered as readonly. This helps when displaying data in a form (to adhere to the form layout of the other input elements), but clearly mark it as readonly so modifications are not allowed.

-=david=-

### DIFF
--- a/src/opnsense/mvc/app/views/layout_partials/form_input_tr.volt
+++ b/src/opnsense/mvc/app/views/layout_partials/form_input_tr.volt
@@ -38,6 +38,7 @@ style       :   css class to add
 maxheight   :   maximum height in rows if applicable
 width       :   width in pixels if applicable
 allownew    :   allow new items (for list) if applicable
+readonly    :   if true, then textareas (textboxes) will be readonly
 
 #}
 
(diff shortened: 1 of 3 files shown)


AdSchellevis:
Hi David, 

Wouldn't the same also count for the other input types? Adding this to the templates is no problem, but maybe we should extend it to the input tags as well....

Best regards,

Ad

dharrigan:
Hi,

Absolutely, but for now, I only (selfish reasons :-)) have need for it on the textarea input element. I would be happy to add it to others in another PR in a little while.

-=david=-

AdSchellevis:
yes please :)

dharrigan:
Could you merge in this one first please, if acceptable? I sort of need it for my scp-backup plugin (to render the public key on the page, but as readonly :-))

src/opnsense/mvc/app/views/layout_partials/form_input_tr.volt
```diff
@@ -65,7 +66,7 @@ allownew    :   allow new items (for list) if applicable
         {% elseif type == "password" %}
             <input type="password" class="form-control" size="{{size|default("50")}}" id="{{ id }}" >
         {% elseif type == "textbox" %}
-            <textarea rows="{{height|default("5")}}" id="{{ id }}"></textarea>
+            <textarea rows="{{height|default("5")}}" id="{{ id }}" {{ readonly ? 'readonly' : '' }}></textarea>
```
fichtner: readonly="readonly"

is XHTML compatible

src/opnsense/mvc/app/views/layout_partials/form_input_tr.volt
```diff
@@ -38,6 +38,7 @@ style       :   css class to add
 maxheight   :   maximum height in rows if applicable
 width       :   width in pixels if applicable
 allownew    :   allow new items (for list) if applicable
+readonly    :   if true, then textareas (textboxes) will be readonly
```
fichtner: better to omit "(textboxes)" if we want to fix all

fichtner: PS: read-only

fichtner: (in the description)

Issue #47, opened 2015-02-19
IPv6/IPv6 Scripts


mannkind:
Is IPv6 currently working on OPNSense? I have my WAN set w/DHCP6, but I don't see it running.

When trying to start it manually, I found */var/etc/dhcp6c_wan.conf* references */var/etc/dhcp6c_wan_script.sh* which tries to start */usr/local/etc/rc.newwanipv6* via *fcgicli*

jschellevis:
issue is related to the fact that fcgicli is still trying to use php-fpm and fpm is not running... WIP

jschellevis:
some changes have been made, please test again with latest git or 15.1.6 when released.

mannkind:
I've updated to 15.1.6.1 –

After running */var/etc/dhcp6c_wan_script.sh* (which just runs */usr/local/etc/rc.newwanipv6 vmx1*) manually, I don't get and IPv6 address and all I can see in the logs is as follows…

* opnsense: /usr/local/etc/rc.newwanipv6: rc.newwanipv6: Informational is starting vmx1.
* opnsense: /usr/local/etc/rc.newwanipv6: rc.newwanipv6: Failed to update WAN[wan] IPv6, restarting…


Let me know if there's anything else I can provide/configure/etc.

pdf:
@mannkind `/var/etc/dhcp6c_wan_script.sh` doesn't actually start `dhcp6c`, it's _called_ by it.  The `dhcp6c` client appears to be started by `/var/etc/rtsold_${interface}_script.sh`, which is in turn called by `rtsold`.

With that said, @jschellevis I came here to report that `rtsold` is extremely flakey on boot - it almost never initializes the `dhcp6c` client.  I've found that by changing rtsold from one-shot to daemon mode in `interfaces.inc` this is very much more reliable.

jschellevis:
@mannkind @pdf thanks for your feedback. @pdf do you think that running rtsold in deamon mode will solve the entire issue or is more investigation needed?

pdf:
@jschellevis I can't be certain what @mannkind's issue is, however dhcp6c was certainly not being started reliably on my system so it seems likely to be related.  I'm not using DHCPv6 to obtain an address in my setup though, just a PD.

[...]

PR #2044, opened 2018-01-02
dhcp6 Client Adanced Options - Enter Prefix Interface Name

marjohn56:
Currently, when using Advanced for dhcp6 client, the config file has the real WAN name as the interface on which the prefix will be applied, this needs to be a LAN ( real ) interface. Simply enter the interface name.

No error checking is carried out as this is already in an advanced selection.

### DIFF
--- a/src/etc/inc/interfaces.inc
+++ b/src/etc/inc/interfaces.inc
@@ -3355,7 +3355,7 @@ function DHCP6_Config_File_Advanced($interface, $wancfg, $wanif)
 
         if (is_numeric($wancfg['adv_dhcp6_prefix_interface_statement_sla_id'])) {
             $id_assoc_statement_prefix .= "\n\tprefix-interface";
-            $id_assoc_statement_prefix .= " {$wanif}";
+            $id_assoc_statement_prefix .= " {$wancfg['adv_dhcp6_prefix_selected_interface']}";
             $id_assoc_statement_prefix .= " {\n";
             $id_assoc_statement_prefix .= "\t\tsla-id {$wancfg['adv_dhcp6_prefix_interface_statement_sla_id']};\n";
             if (($wancfg['adv_dhcp6_prefix_interface_statement_sla_len'] >= 0) &&
--- a/src/etc/inc/interfaces.inc
+++ b/src/etc/inc/interfaces.inc
@@ -3132,44 +3132,49 @@ function interface_dhcpv6_configure($interface = 'wan', $wancfg)
         $dhcp6cconf .= "};\n";
     } else {
         /* skip address request if this is set */
-        if (!isset($wancfg['dhcp6prefixonly'])) {
-            $dhcp6cconf .= "  send ia-na 0;   # request stateful address\n";
-        }
-        if (is_numeric($wancfg['dhcp6-ia-pd-len'])) {
-            $dhcp6cconf .= "  send ia-pd 0;  # request prefix delegation\n";
+        $iflist = link_interface_to_track6($interface);
+        $count = 0;
+        foreach ($iflist as $friendly => $ifcfg) {
+            if (!isset($wancfg['dhcp6prefixonly'])) {
+              $dhcp6cconf .= "  send ia-na {$count};   # request stateful address\n";
+            }
+            if (is_numeric($wancfg['dhcp6-ia-pd-len'])) {
+              $dhcp6cconf .= "  send ia-pd {$count};  # request prefix delegation\n";
+            }
+         ++$count;
         }
 
         $dhcp6cconf .= "\trequest domain-name-servers;\n";
         $dhcp6cconf .= "\trequest domain-name;\n";
         $dhcp6cconf .= "\tscript \"/var/etc/dhcp6c_{$interface}_script.sh\"; # we'd like some nameservers please\n";
-
         $dhcp6cconf .= "};\n";
 
-        if (!isset($wancfg['dhcp6prefixonly'])) {
-            $dhcp6cconf .= "id-assoc na 0 { };\n";
-        }
-
-        if (is_numeric($wancfg['dhcp6-ia-pd-len'])) {
-            /* Setup the prefix delegation */
-            $dhcp6cconf .= "id-assoc pd 0 {\n";
-            $preflen = 64 - $wancfg['dhcp6-ia-pd-len'];
-            if (isset($wancfg['dhcp6-ia-pd-send-hint'])) {
-                $dhcp6cconf .= "  prefix ::/{$preflen} infinity;\n";
+        $count = 0;
+        foreach ($iflist as $friendly => $ifcfg) {
+            if (!isset($wancfg['dhcp6prefixonly'])) {
+                $dhcp6cconf .= "id-assoc na {$count} { };\n";
             }
-            $iflist = link_interface_to_track6($interface);
-            foreach ($iflist as $friendly => $ifcfg) {
-                if (is_numeric($ifcfg['track6-prefix-id'])) {
-                    $realif = get_real_interface($friendly);
-                    $dhcp6cconf .= "  prefix-interface {$realif} {\n";
-                    $dhcp6cconf .= "    sla-id {$ifcfg['track6-prefix-id']};\n";
-                    $dhcp6cconf .= "    sla-len {$wancfg['dhcp6-ia-pd-len']};\n";
-                    $dhcp6cconf .= "  };\n";
+            if (is_numeric($wancfg['dhcp6-ia-pd-len'])) {
+                /* Setup the prefix delegation */
+                $dhcp6cconf .= "id-assoc pd {$count} {\n";
+                $preflen = 64 - $wancfg['dhcp6-ia-pd-len'];
+                if (isset($wancfg['dhcp6-ia-pd-send-hint'])) {
+                    $dhcp6cconf .= "  prefix ::/{$preflen} infinity;\n";
                 }
             }
-            unset($preflen, $iflist, $ifcfg);
+            if (is_numeric($ifcfg['track6-prefix-id'])) {
+                $realif = get_real_interface($friendly);
+                $dhcp6cconf .= "  prefix-interface {$realif} {\n";
+                $dhcp6cconf .= "    sla-id 0;\n";
+                $dhcp6cconf .= "    sla-len {$wancfg['dhcp6-ia-pd-len']};\n";
+                $dhcp6cconf .= "  };\n";
+            }
             $dhcp6cconf .= "};\n";
+            ++$count;
         }
+        unset($preflen, $iflist, $ifcfg);
     }
+
 
     // DHCP6 Config File Advanced
     if ($wancfg['adv_dhcp6_config_advanced']) {
@@ -3354,8 +3359,9 @@ function DHCP6_Config_File_Advanced($interface, $wancfg, $wanif)
         }
 
         if (is_numeric($wancfg['adv_dhcp6_prefix_interface_statement_sla_id'])) {
+            $ad_if = get_real_interface($wancfg['adv_dhcp6_prefix_selected_interface']);
             $id_assoc_statement_prefix .= "\n\tprefix-interface";
-            $id_assoc_statement_prefix .= " {$wancfg['adv_dhcp6_prefix_selected_interface']}";
+            $id_assoc_statement_prefix .= " {$ad_if}";
             $id_assoc_statement_prefix .= " {\n";
             $id_assoc_statement_prefix .= "\t\tsla-id {$wancfg['adv_dhcp6_prefix_interface_statement_sla_id']};\n";
             if (($wancfg['adv_dhcp6_prefix_interface_statement_sla_len'] >= 0) &&
--- a/src/www/interfaces.php
+++ b/src/www/interfaces.php
@@ -2625,8 +2625,25 @@ include("head.inc");
                         <tr class="dhcpv6_advanced">
                           <td><i class="fa fa-info-circle text-muted"></i> <?=gettext("Prefix Interface");?></td>
                           <td>
-                            <i><?=gettext("Prefix Interface name ( real \"igb1\" - \"em1\" etc )"); ?></i>
-                            <input name="adv_dhcp6_prefix_selected_interface" type="text" id="adv_dhcp6_prefix_selected_interface" value="<?=$pconfig['adv_dhcp6_prefix_selected_interface'];?>" />
+                            <select name='adv_dhcp6_prefix_selected_interface' class='selectpicker' data-style='btn-default' >
+<?php
+                            foreach (get_configured_interface_with_descr(false, true) as $iface => $ifacename):
+                              switch($config['interfaces'][$iface]['ipaddrv6']) {
+                                case "track6":
+                                    break;
+                                default:
+                                    continue 2;
+                              }?>
+                                <option value="<?=$iface;?>" <?=$iface == $pconfig['adv_dhcp6_prefix_selected_interface'] ? " selected=\"selected\"" : "";?>>
+                                    <?= htmlspecialchars($ifacename);?>
+                                </option>
+<?php
+                            endforeach;?>
+                            </select>
+                          </td>
+                        <tr class="dhcpv6_advanced">
+                          <td><i class="fa fa-info-circle text-muted"></i> <?=gettext("Prefix Interface");?></td>
+                          <td>
                             <i><?=gettext("Site-Level Aggregation Identifier"); ?></i>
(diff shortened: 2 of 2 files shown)


fichtner:
Can't we determine this programmatically? Preferably in automatic fashion, otherwise as a "checkbox" type. Maybe it's liked to useipv4interface anyway?

marjohn56:
It's not the WAN interface that the prefix delegation gets assigned to, it's a LAN interface. You can have more than one LAN interface, dhcp6c, when given the right info can support multiple WANs, thus you could have WAN(1) delegating a prefix to LAN_A and WAN(2) delegating a prefix to LAN_B.

This is purely prefix delegation. What you could do is a list box similar the interface assignment, but how can you determine if the interface is a LAN... hmm, I suppose if the interface is set to tracking, then we could create a list from that and the user could select?

fichtner:
@marjohn56 yes, your tracking interface collection suggestion sounds way better

marjohn56:
Try that, see if it's better. :)

fichtner:
It's becoming clearer now, but two points to consider:

* We know which interface is tracking, why not auto-select it or calculate from the backend? Worst case we need a checkbox to enable this, best case no user-interaction at all.
* Multiple interfaces can track a single WAN connection. I believe with the prefix ID this works pretty  well so we want to make that work too.

marjohn56:
Hmm....

So, we look at the LAN(s) interface and see which WAN it/they are tracking, and do it that way?

fichtner:
I see, the tricky part is we need to reload a WAN config if the trackers change then :)

marjohn56:
Maybe not so tricky. We already have this in the Interfaces_dhcpv6_configure(WAN)..

 if ($wancfg['adv_dhcp6_config_advanced']) {
        $dhcp6cconf = DHCP6_Config_File_Advanced($interface, $wancfg, $wanif);
    }

We would need either to find the correct interfaces in the LAN configuration function and call it from there, and I've never looked at that part of the code, or we write a new function in that section of code to update the conf file. 

It will require a restart of dhcp6c either way,

marjohn56:
@fichtner 

I've had some time today to look closer at this. I think with the advanced setting it will need to be limited to one WAN interface. If you try and take it beyond that you are effectively creating a GUI to replace the config file override. 

For example, in the Send Options you must specify the wan interface, the real interface not the friendly name so to a degree it's already beyond the click and select type of configuration. If you want to have multiple WAN interfaces then we would not only have to list them for selection, but we would then need to pull back the user entered options to work out the ia-na and ia-pd values and create the id_assoc na and pd's accordingly, In doing so as I said we are just creating a GUI for the config override.

I think at this point it's enough just to have the select LAN interface option that we had with this PR and use that. Can I have your thoughts on this Franco?

What I have done today is to modify the basic dhcpv6 handling. I have changed the function in interface_dhcpv6_configure where the dhcp_WAN.conf is created, scanning the LAN interfaces to see if they are tracking the given wan interface and adding them to the conf file. I have also implemented an increment counter so that both the ia and pd values are incremented for each prefix, this means that each LAN interface now gets a separate prefix delegation. Without this the prefix is the same on both LAN interfaces and is effectively useless as you cannot route to two destinations with the same prefix. 

I will post these changes as an addition to this PR tomorrow when I have squashed them.

fichtner:
This is still odd. Can't we simply determine the track6 interfaces belonging to WAN during config write time and write one block for each of them?

fichtner:
@marjohn56 it might help to understand this by posting a dhcp6c config file skeleton of what you have in mind (WAN plus a LAN and DMZ tracking it)?

marjohn56:
Actually, a light just came on in the brain. If you look at the last change I posted a minute ago that's exactly what I have done in the basic settings. For some reason whilst I was looking at the advanced I failed to notice that the first set of changes in the basic settings get pulled into the advanced settings.. that makes life MUCH simpler than I thought.

Here's a working dhcp6c_wan_conf generated with two LANs and one WAN,

interface igb0 {
  send ia-na 0;   # request stateful address
  send ia-pd 0;  # request prefix delegation
  send ia-na 1;   # request stateful address
  send ia-pd 1;  # request prefix delegation
	request domain-name-servers;
	request domain-name;
	script "/var/etc/dhcp6c_wan_script.sh"; # we'd like some nameservers please
};
id-assoc na 0 { };
id-assoc pd 0 {
  prefix-interface igb1 {
    sla-id 0;
    sla-len 0;
  };
};
id-assoc na 1 { };
id-assoc pd 1 {
  prefix-interface igb2 {
    sla-id 0;
    sla-len 0;
  };
};

marjohn56:
You can cherry pick the bit in interface_dhcpv6_configure that does that creates the conf file, works nicely. Of course  the fix for the routing still needs sorting or you have to manually create a gateway and route!

fichtner:
Hehe, so we don't need any of the advanced bits?

marjohn56:
You still need some of them, other options need to be sent,  RAW for example, but we can take the top half of the routines in the basic config and create a sub that will then be used in the basic and the advanced. Add the other options in the advanced and then call a second function that uses the bottom half of those new routines...

Job done?

fichtner:
@marjohn56 

good morning! :)

After a fresh start this seems to be  a weird change in general. I had to modify and fix the config write as it did not do quite what you wanted.

Now looking at what it does...  it is splitting one prefix request in two, parting with the usual way of doing things within the scope of this project. and there is no way backwards for the users who liked this approach: get one prefix, assign different prefix ID /64 subnets to the tracking interfaces.

The only use case where this is strictly necessary is when the prefix is 64 and we actually need two prefixes for the tracking interfaces, because two /64 won't fit into one. If that's the case we need to make this an optional doing split requests, but fixed to /64 to be on the safe side....

The config itself is generated correctly from here without your patch with regard to interface usage, although maybe there is a pppoe oddity at play. This is from my test setup:

```
interface em1 {
  send ia-pd 0;  # request prefix delegation
	request domain-name-servers;
	request domain-name;
	script "/var/etc/dhcp6c_wan_script.sh"; # we'd like some nameservers please
};
id-assoc pd 0 {
  prefix-interface run0_wlan1 {
    sla-id 0;
    sla-len 4;
  };
  prefix-interface em2 {
    sla-id 2;
    sla-len 4;
  };
};
```

marjohn56:
Morning.. :)

In the current code you cannot do two tracking interfaces using the basic settings, what you get is a slaac generated address on the one interface with no prefix assigned. Strange that yours does not do as it should, I have a /48 from my ISP to my primary router. Here's my ifconfig output from the second test router running the patch I sent.


igb0: flags=8843<UP,BROADCAST,RUNNING,SIMPLEX,MULTICAST> metric 0 mtu 1500
        options=4400b8<VLAN_MTU,VLAN_HWTAGGING,JUMBO_MTU,VLAN_HWCSUM,VLAN_HWTSO,TXCSUM_IPV6>
        ether 00:0d:b9:44:72:70
        hwaddr 00:0d:b9:44:72:70
        inet6 fe80::20d:b9ff:fe44:7270%igb0 prefixlen 64 scopeid 0x1
        inet6 2a02:8010:6228:0:20d:b9ff:fe44:7270 prefixlen 64 autoconf
        inet6 2a02:8010:6228:0:eeee:ba91:1110:aa8 prefixlen 128
        inet6 2a02:8010:6228:0:eeee:cef5:309f:df75 prefixlen 128
        inet 192.168.1.15 netmask 0xffffff00 broadcast 192.168.1.255
        media: Ethernet autoselect (1000baseT <full-duplex>)
        status: active
igb1: flags=8843<UP,BROADCAST,RUNNING,SIMPLEX,MULTICAST> metric 0 mtu 1500
        options=4400b8<VLAN_MTU,VLAN_HWTAGGING,JUMBO_MTU,VLAN_HWCSUM,VLAN_HWTSO,TXCSUM_IPV6>
        ether 00:0d:b9:44:72:71
        hwaddr 00:0d:b9:44:72:71
        inet 192.168.3.254 netmask 0xffffff00 broadcast 192.168.3.255
        inet6 2a02:8010:6228:ff:20d:b9ff:fe44:7271 prefixlen 64
        inet6 fe80::1:1%igb1 prefixlen 64 scopeid 0x2
        nd6 options=21<PERFORMNUD,AUTO_LINKLOCAL>
        media: Ethernet autoselect (1000baseT <full-duplex>)
        status: active
igb2: flags=8843<UP,BROADCAST,RUNNING,SIMPLEX,MULTICAST> metric 0 mtu 1500
        options=4400b8<VLAN_MTU,VLAN_HWTAGGING,JUMBO_MTU,VLAN_HWCSUM,VLAN_HWTSO,TXCSUM_IPV6>
        ether 00:0d:b9:44:72:72
        hwaddr 00:0d:b9:44:72:72
        inet 192.168.4.254 netmask 0xffffff00 broadcast 192.168.4.255
        inet6 2a02:8010:6228:fe:20d:b9ff:fe44:7272 prefixlen 64
        inet6 fe80::1:1%igb2 prefixlen 64 scopeid 0x3
        nd6 options=21<PERFORMNUD,AUTO_LINKLOCAL>
        media: Ethernet autoselect
        status: no carrier

As you can see, I have two separate /64 prefixes. or am I missing something?

fichtner:
/64 are always separate, sure. Well, ok, looks like this does the trick then, both seem to be in your /48. I hope this is not provider-specific trickery.

The modified patch is here... https://github.com/opnsense/core/commit/d2ba69f5dae817b1b24ef29cd9ce1591da92bbe7

marjohn56:
My problem is I have one tester who is on the same ISP as me, gets a /48 so I know his will work as mine does. I have another who's on a /56 but has kids so has to chose the right moment to be able to test and another who won't make the move across until the stuff we have already done is released ( useless :-/ ).

My other issue is that I am seeing multiple dhcp6c clients if I do a reload all services... not good.

fichtner:
Comparing your config and mine again... I haven't tested this yet, but hopping over to the office within the hour... that being said... 

Could it simply be that "prefix-interface" cannot occur twice per id-assoc pd statement?

Do you affirm that without this change the second interface does not receive a delegated prefix?

marjohn56:
I'll test your bits now and see what happens.... give me 5

marjohn56:
I've used your conf, but changed the interfaces... I get diddly squat,. Here's the output from dhcp6c. Upper 48 bits removed to protect the innocent.

8:38:02 OPNsense dhcp6c[77769]: reset a timer on igb0, state=REQUEST, timeo=6, retrans=29739
Jan  9 08:38:02 OPNsense dhcp6c[77769]: receive reply from fe80::277:70ff:fecd:1e00%igb0 on igb0
Jan  9 08:38:02 OPNsense dhcp6c[77769]: get DHCP option IA_PD, len 41
Jan  9 08:38:02 OPNsense dhcp6c[77769]:   IA_PD: ID=0, T1=0, T2=0
Jan  9 08:38:02 OPNsense dhcp6c[77769]: get DHCP option IA_PD prefix, len 25
Jan  9 08:38:02 OPNsense dhcp6c[77769]:   IA_PD prefix: xx:ff::/64 pltime=4500 vltime=128784594377760
Jan  9 08:38:02 OPNsense dhcp6c[77769]: get DHCP option client ID, len 14
Jan  9 08:38:02 OPNsense dhcp6c[77769]:   DUID: 00:01:00:01:21:e5:4e:c8:00:0d:b9:44:72:70
Jan  9 08:38:02 OPNsense dhcp6c[77769]: get DHCP option server ID, len 14
Jan  9 08:38:02 OPNsense dhcp6c[77769]:   DUID: 00:01:00:01:21:e2:d5:62:02:77:70:cd:1e:00
Jan  9 08:38:02 OPNsense dhcp6c[77769]: get DHCP option DNS, len 16
Jan  9 08:38:02 OPNsense dhcp6c[77769]: dhcp6c Received REQUEST
Jan  9 08:38:02 OPNsense dhcp6c[77769]: nameserver[0] xx:0:20d:b9ff:fe44:7271
Jan  9 08:38:02 OPNsense dhcp6c[77769]: make an IA: PD-0
Jan  9 08:38:02 OPNsense dhcp6c[77769]: create a prefix xx:ff::/64 pltime=4500, vltime=6841882909728
Jan  9 08:38:02 OPNsense dhcp6c[77769]: invalid prefix length 64 + 4 + 64
Jan  9 08:38:02 OPNsense dhcp6c[77769]: invalid prefix length 64 + 4 + 64
Jan  9 08:38:02 OPNsense dhcp6c[77769]: T1(2250) and/or T2(3600) is locally determined
Jan  9 08:38:02 OPNsense dhcp6c[77769]: executes /var/etc/dhcp6c_wan_script.sh
Jan  9 08:38:02 OPNsense dhcp6c: dhcp6c REQUEST on igb0
Jan  9 08:38:02 OPNsense dhcp6c: dhcp6c REQUEST on igb0 - running newipv6
Jan  9 08:38:06 OPNsense dhcp6c[77769]: script "/var/etc/dhcp6c_wan_script.sh" terminated
Jan  9 08:38:06 OPNsense dhcp6c[77769]: removing an event on igb0, state=REQUEST
Jan  9 08:38:06 OPNsense dhcp6c[77769]: removing server (ID: 00:01:00:01:21:e2:d5:62:02:77:70:cd:1e:00)
Jan  9 08:38:06 OPNsense dhcp6c[77769]: got an expected reply, sleeping.

fichtner:
What's your dhcp6c wan conf?

marjohn56:
interface igb0 {
  send ia-na 0; # request stateful address for lan
  send ia-pd 0; # request prefix delegation for lan
  send ia-na 1; # request stateful address for opt1
  send ia-pd 1; # request prefix delegation for opt1
  request domain-name-servers;
  request domain-name;
  script "/var/etc/dhcp6c_wan_script.sh"; # we'd like some nameservers please
};
id-assoc na 0 { };
id-assoc pd 0 { # more prefix delegation for lan
  prefix-interface igb1 {
    sla-id 0;
    sla-len 0;
  };
};
id-assoc na 1 { };
id-assoc pd 1 { # more prefix delegation for opt1
  prefix-interface igb2 {
    sla-id 0;
    sla-len 0;
  };
};

fichtner:
It looks ok oO

marjohn56:
When you get a moment, can you do a reload all services and see if you get multiple dhcp6c clients?

fichtner:
When using option 11 from the console there are *sometimes* multiple dhcp6c clients. killall doesn't work, but issuing SIGINT to the stray process seems to convince it to exit.

When I track two interfaces via the old master state, I get a prefix for both as expected after reboot... but it is the same prefix although it should be within a /63 where there are two. After reload, however, there is only one prefix on the first LAN.

When I use the split approach, I only get a prefix on the second LAN, the first is never working. :/

marjohn56:
OK, well it's a work in progress, I have now split out the dhcp6c_wan.conf from the rest of that code. I need to add some stuff to take into account different setups. I will also look at the dhcp6c issue. One question there. Do we really need RTSOLD to be running all the time, would a one shot work as well in opnsense?

marjohn56:
So do you only have single /64 at  wan interface or is it a /63?

Look at dhcp6c log and see what it is showing.

fichtner:
Ancient reason for not running one-shot mode in rtsold: https://github.com/opnsense/core/issues/47

I'm trying to fix the dhcp6c startup race, it seems it takes longer than a second to stop, it's not so useful then :D

fichtner:
@marjohn56 try https://github.com/opnsense/core/commit/e6dc3265

fichtner:
Now it works on master (no actual changes to dhcp6c config), /63 prefix, one for ID 0 and one for ID 1 gives me two separate /64 on LAN 1 and 2.

fichtner:
Funky, the prefix delegation is correct but radvd is giving out LAN2 IPs for LAN1, which bounces off the antispoof, no IPv6 connectivity for clients at all... wtf

marjohn56:
Don't waste your time trying to fix dhcp6c dupes. I have a solid system that I put in place across the road. I already have an opnsense interfaces.inc with all the changes in it. I was just waiting for the dhcp6c multiple clients issue to raise its head. I'll try and get it over to Github tonight, I am onsite with a client at present working on their issues.

marjohn56:
@fichtner 


Try this [.](https://github.com/marjohn56/core/commit/8c75f704247e4439a3e2446ac89cfb0a5360110b) I've broken the basic conf creation into a separate function to tidy things up. It will create the sort of dhcp6c_*.conf that we are used to, but will auto add any tracking interfaces. The same thing now happens with the advanced, that also automatically adds tracking interfaces. Try it and let me know. If we want to assign a different prefix to each LAN interface, then we know how to do that and it's an option we can add to interfaces.php.

The other reason I've broken it away is to simplify the interface_dhcpv6_configure(), what's left will get some attention next once your happy that this works.

marjohn56:
closed in favour of PR #2072

PR #2046, opened 2018-01-03
HTML Compliance - Element <button> Descendant of <a> Element

NOYB:
Error: The element button must not appear as a descendant of the a element.

Specification:
http://w3c.github.io/html/textlevel-semantics.html#the-a-element
"Allowed ARIA role attribute values: link (default - do not set), button, checkbox, radio, switch, tab or treeitem"
"The <a> element may be wrapped around entire paragraphs, lists, tables, and so forth, even entire sections, so long as there is no interactive content within (e.g., buttons or other links)."


fichtner:
Thank you!

Issue #1899, opened 2017-10-27
Legacy 2 MVC


kekek2:
What are the plans in the translation of legacy code to MVC ? Do you need help ?

AdSchellevis:
We are still working in some core areas (mainly firewall), but we can't manage too many changes at once. 
Help is certainly welcome, preferable for features which aren't too big and can easily be migrated, such as diagnostics pages (dns lookup, ping, ...).

Every pull request costs us time to manage and test, which is fine, but we need to spend the limited time we have wisely. 
https://github.com/opnsense/core/issues/1774 for example also needs work, we did pull in the code, but can't spend enough time at the moment to finish it.

Thanks for your support!

mimugmail:
@kekek2 how about certificate Code .. would close #1905

AdSchellevis:
@mimugmail @kekek2  I would suggest starting with smaller things, the trust/certificate section is quite a blob of work, which might result in quite some work for the core team to review as well (which we might not have available at the moment).

kekek2:
The trust/certificates are almost ready, tested. Will soon make pull-request.

PR #2053, opened 2018-01-04
Trust MVC from Smart-Soft

kekek2:


src/opnsense/mvc/app/controllers/OPNsense/Trust/Api/TrustBase.php
```diff
@@ -0,0 +1,108 @@
...
+        if (isset($post["dn_commonname"]) && preg_match("/[\!\@\#\$\%\^\(\)\~\?\>\<\&\/\\\,\"\']/", $post["dn_commonname"])) {
+            $result["validations"]["{$method}.dn_commonname"] = gettext("The field 'Distinguished name Common Name' contains invalid characters.");
+        }
+        if (isset($post["keylen"]) && !in_array($post["keylen"], array_keys($this->keylens))) {
+            $result["validations"]["{$method}.keylen"] = gettext("Please select a valid Key Length.");
+        }
+        if (isset($post["digest_alg"]) && !in_array($post["digest_alg"], array_keys($this->digest_algs))) {
+            $result["validations"]["{$method}.digest_alg"] = gettext("Please select a valid Digest Algorithm.");
+        }
+        return $result;
```
fabianfrz: you should override performValidatoion like in #2018.

src/opnsense/mvc/app/controllers/OPNsense/Trust/forms/externalCert.xml
```diff
@@ -0,0 +1,67 @@
+<form>
+    <field>
+        <id>External.descr</id>
+        <label>Descriptive name</label>
```
fabianfrz: N

src/opnsense/mvc/app/controllers/OPNsense/Trust/forms/externalCert.xml
```diff
@@ -0,0 +1,67 @@
...
+        <type>text</type>
+    </field>
+    <field>
+        <type>header</type>
+        <label>External Signing Request</label>
+    </field>
+    <field>
+        <id>External.keylen</id>
+        <label>Key length (bits)</label>
```
fabianfrz: Length

src/opnsense/mvc/app/controllers/OPNsense/Trust/forms/externalCert.xml
```diff
@@ -0,0 +1,67 @@
...
+        <type>dropdown</type>
+    </field>
+    <field>
+        <id>External.digest_alg</id>
+        <label>Digest Algorithm</label>
+        <type>dropdown</type>
+        <help>NOTE: It is recommended to use an algorithm stronger than SHA1 when possible.</help>
```
fabianfrz: no NOTE:

src/opnsense/mvc/app/controllers/OPNsense/Trust/forms/intermediateCA.xml
```diff
@@ -0,0 +1,71 @@
...
+    <field>
+        <id>Intermediate.keylen</id>
+        <label>Key length (bits)</label>
+        <type>dropdown</type>
+    </field>
+    <field>
+        <id>Intermediate.digest_alg</id>
+        <label>Digest Algorithm</label>
+        <type>dropdown</type>
+        <help>NOTE: It is recommended to use an algorithm stronger than SHA1 when possible.</help>
```
fabianfrz: no NOTE:

src/opnsense/mvc/app/models/OPNsense/Trust/Trust.xml
```diff
@@ -0,0 +1,81 @@
...
+                <serial type="IntegerField">
+                </serial>
+            </ca>
+        </cas>
+        <certs>
+            <cert type="ArrayField">
+                <refid type="TextField">
+                </refid>
+                <cauuid type="TextField">
+                </cauuid>
```
fabianfrz: this should probably be a `ModelRelationField`

src/opnsense/mvc/app/models/OPNsense/Trust/Trust.xml
```diff
@@ -0,0 +1,81 @@
...
+                <csr type="TextField">
+                </csr>
+            </cert>
+        </certs>
+        <crls>
+            <crl type="ArrayField">
+                <refid type="TextField">
+                </refid>
+                <cauuid type="TextField">
+                </cauuid>
```
fabianfrz: Maybe a `ModelRelationField`

src/opnsense/mvc/app/models/OPNsense/Trust/Trust.xml
```diff
@@ -0,0 +1,81 @@
...
+                </descr>
+                <method type="OptionField">
+                    <OptionValues>
+                        <existing>existing</existing>
+                        <internal>internal</internal>
+                    </OptionValues>
+                </method>
+                <serial type="IntegerField">
+                </serial>
+                <lifetime type="IntegerField">
```
fabianfrz: minimum is probably 0 so please try: `<MinimumValue>0</MinimumValue>`

src/opnsense/mvc/app/views/OPNsense/Trust/authorities.volt
```diff
@@ -0,0 +1,132 @@
...
+                                "<button type=\"button\" class=\"btn btn-xs btn-default command-exp\" data-type=\"key\" data-row-id=\"" + row.uuid + "\" title=\"{{ lang._('export CA private key') }}\" alt=\"{{ lang._('export CA private key') }}\"><span class=\"glyphicon glyphicon-download\"></span></button>" +
+                                "<button type=\"button\" class=\"btn btn-xs btn-default command-delete\" data-row-id=\"" + row.uuid + "\" title=\"{{ lang._('delete ca') }}\" alt=\"{{ lang._('delete ca') }}\"><span class=\"fa fa-trash-o\"></span></button>";
+                        },
+                        "Distinguished": function (column, row) {
+                            return row[column.id] +
+                            "<table width=\"100%\" style=\"font-size: smaller\"> \
+                                <tr> \
+                                <td>&nbsp;</td> \
+                            <td width=\"20%\">{{ lang._('Valid From') }}:</td> \
+                            <td width=\"70%\">" + row["startdate"] + "</td> \
```
fabianfrz: it should not contain width - this should be a CSS property .
The ":" should be in the translation call.

src/opnsense/mvc/app/views/OPNsense/Trust/authorities.volt
```diff
@@ -0,0 +1,132 @@
...
+                            </tr> \
+                            <tr> \
+                            <td>&nbsp;</td> \
+                            <td>{{ lang._('Valid Until') }}:</td> \
+                            <td>" + row["enddate"] + "</td> \
```
fabianfrz: same here

fabianfrz: to be secure - please escape

src/opnsense/mvc/app/views/OPNsense/Trust/authorities.volt
fabianfrz: please escape

src/opnsense/mvc/app/views/OPNsense/Trust/certificates.volt
```diff
@@ -0,0 +1,178 @@
...
+                        "Name": function (column, row) {
+                            return "<span class=\"glyphicon glyphicon-certificate __iconspacer\"></span>" + row[column.id] + "<br>" + row["Purpose"];
+                        },
+                        "Distinguished": function (column, row) {
+                            return row[column.id] +
+                                "<table width=\"100%\" style=\"font-size: smaller\"> \
+                                    <tr> \
+                                    <td>&nbsp;</td> \
+                                <td width=\"20%\">{{ lang._('Valid From') }}:</td> \
+                            <td width=\"70%\">" + row["startdate"] + "</td> \
```
fabianfrz: same here

src/opnsense/mvc/app/views/OPNsense/Trust/certificates.volt
```diff
@@ -0,0 +1,178 @@
...
+                            </tr> \
+                            <tr> \
+                            <td>&nbsp;</td> \
+                            <td>{{ lang._('Valid Until') }}:</td> \
+                            <td>" + row["enddate"] + "</td> \
```
fabianfrz: and here

src/opnsense/mvc/app/views/OPNsense/Trust/certificates.volt
```diff
@@ -0,0 +1,178 @@
...
+            }).end();
+            // info item
+            grid.find(".command-certinfo").on("click", function (e) {
+                ajaxGet(url = "/api/trust/certificates/info/" + $(this).data("row-id"),
+                    sendData = {}, callback = function (data, status) {
+                        if (status == 'success') {
+                            BootstrapDialog.show({
+                                title: "{{ lang._('Certificate') }}",
+                                type: BootstrapDialog.TYPE_INFO,
+                                message: "<pre>" + data.message + "</pre>",
```
fabianfrz: escape

src/opnsense/mvc/app/views/OPNsense/Trust/revocation.volt
```diff
@@ -0,0 +1,123 @@
...
+                                ret += "<button type=\"button\" class=\"btn btn-xs btn-default command-edit\" data-row-id=\"" + row.uuid + "\" title=\"{{ lang._('Edit CRL') }}\" alt=\"{{ lang._('Edit CRL') }}\"><span class=\"fa fa-pencil\"></span></button> ";
+                            }
+                            return ret +
+                                "<button type=\"button\" class=\"btn btn-xs btn-default command-delete\" data-row-id=\"" + row.uuid + "\" title=\"{{ lang._('Delete CRL') }}\" alt=\"{{ lang._('Delete CRL') }}\"><span class=\"fa fa-trash-o\"></span></button>";
+                        },
+                        "Name": function (column, row) {
+                            return "<span class=\"glyphicon glyphicon-certificate __iconspacer\"></span>" + row[column.id] + "<br>" + row["Purpose"];
+                        },
+                        "Distinguished": function (column, row) {
+                            return row[column.id] + "<br>{{ lang._('Valid From') }}: " + row["startdate"] + "<br>{{ lang._('Valid Until') }}: " + row["enddate"];
```
fabianfrz: escape

src/opnsense/mvc/app/views/OPNsense/Trust/revocation.volt
```diff
@@ -0,0 +1,123 @@
...
+                            if (row.InternalBool) {
+                                ret += "<button type=\"button\" class=\"btn btn-xs btn-default command-edit-internal\" data-row-id=\"" + row.uuid + "\" title=\"{{ lang._('Edit CRL') }}\" alt=\"{{ lang._('Edit CRL') }}\"><span class=\"fa fa-pencil\"></span></button> ";
+                            } else {
+                                ret += "<button type=\"button\" class=\"btn btn-xs btn-default command-edit\" data-row-id=\"" + row.uuid + "\" title=\"{{ lang._('Edit CRL') }}\" alt=\"{{ lang._('Edit CRL') }}\"><span class=\"fa fa-pencil\"></span></button> ";
+                            }
+                            return ret +
+                                "<button type=\"button\" class=\"btn btn-xs btn-default command-delete\" data-row-id=\"" + row.uuid + "\" title=\"{{ lang._('Delete CRL') }}\" alt=\"{{ lang._('Delete CRL') }}\"><span class=\"fa fa-trash-o\"></span></button>";
+                        },
+                        "Name": function (column, row) {
+                            return "<span class=\"glyphicon glyphicon-certificate __iconspacer\"></span>" + row[column.id] + "<br>" + row["Purpose"];
```
fabianfrz: escape

src/www/vpn_openvpn_client.php
```diff
@@ -882,7 +887,7 @@ function addRow() {
 <?php
               if (!isset($config['cert']) || count($config['cert']) == 0) :?>
                 <b><?=gettext("No Certificates defined.");?></b> <br /><?=gettext("Create one under");?>
-                <a href="system_certmanager.php"><?=gettext("System: Certificates");?></a> <?=gettext("if one is required for this connection.");?>
+                <a href="/ui/trust/certificates/"><?=gettext("System: Certificates");?></a> <?=gettext("if one is required for this connection.");?>
```
fabianfrz: Isn't that "System:Trust"? Needs probably more work to do because the string needs to be a template.

src/www/vpn_openvpn_server.php
```diff
@@ -806,25 +810,26 @@
...
+                              <?=htmlspecialchars($ca->descr->__toString());?>
                             </option>
 <?php
                           endforeach; ?>
                           </select>
 <?php
                         else :?>
                           <b><?=gettext("No Certificate Authorities defined.");?></b>
-                          <br /><?=gettext("Create one under")?> <a href="system_camanager.php"> <?=gettext("System: Certificates");?></a>.
+                          <br /><?=gettext("Create one under")?> <a href="/ui/trust/authorities/"> <?=gettext("System: Certificates");?></a>.
```
fabianfrz: System: Trust

src/www/vpn_openvpn_server.php
```diff
@@ -861,40 +864,41 @@
...
+                            <?=htmlspecialchars($cert->descr->__toString() . $caname . $inuse . $revoked);?>
                           </option>
 <?php
                         endforeach; ?>
                         </select>
 <?php
                       else :?>
                           <b><?=gettext("No Certificates defined.");?></b>
-                          <br /><?=gettext("Create one under");?> <a href="system_certmanager.php"><?=gettext("System: Certificates");?></a>.
+                          <br /><?=gettext("Create one under");?> <a href="/ui/trust/certificates/"><?=gettext("System: Certificates");?></a>.
```
fabianfrz: System: Trust

src/www/wizard.php
```diff
@@ -594,14 +595,14 @@ function showchange() {
...
-				foreach($config['ca'] as $ca) {
-					if (!empty($field['internal']) && empty($ca['prv'])) {
+				foreach($mdlTrust->cas->ca->getChildren() as $uuid => $ca) {
+					if (!empty($field['internal']) && empty($ca->prv->__toString())) {
 						continue;
 					}
-					$name = htmlspecialchars($ca['descr']);
+					$name = htmlspecialchars($ca->descr->__toString());
 					$SELECTED = "";
 					if ($value == $name) $SELECTED = " selected=\"selected\"";
```
fabianfrz: selected is probably bad because uppercase is usually understand as a constant

src/opnsense/mvc/app/controllers/OPNsense/Trust/Api/TrustBase.php
kekek2: Usually the input form matches a data model. So it is convenient to validate the data entered at the model level. In the case of a Trust/Certificates, we introduce some data, and on the basis of their calculated values to be written to the model. Therefore, you need to validate the entered data, because the data for the model can simply not now contain.

evbevz:
Fabian, how do you see the model? How can it be improved?

src/opnsense/mvc/app/models/OPNsense/Trust/Trust.xml
kekek2: In my opinion, even better to use CertificateField. I tried to use CertificateField, but I'm having problems with the migration. The certificates are not migrated. Does not pass the validator "InclusionIn". It checks the list of available CA, but they have just been added and the validator does not know them. In all other cases CertificateField works great. ModelRelationField also uses the validator "InclusionIn", most likely, will have the same problems with migration.

I will think about how you can use CertificateField and use it with migration.

fabianfrz: @fichtner @AdSchellevis looks like a problem with the framework

fabianfrz:
@evbevz do as many validations in the XML or in Field types - New Field Types can be created so it is possibe to add a PEM field which includes the certificate or Base64 if it is encoded. If you look at Trust.xml it does not really contain validation. But something like that exists: https://github.com/opnsense/core/pull/2053/files#diff-29bb443b0cb80ae7ef701e4f4625dcb8R76

src/opnsense/mvc/app/models/OPNsense/Trust/Trust.xml
kekek2: The field "refid" is necessary only for migration. Then it is not being used.

AdSchellevis:
@kekek2 The chances that this PR will make it into core won't be very large, please coordinate in the future about planned work and the effort you require from our team for bringing features is. 

We really try to review a lot of code that is send in, we try to be reasonable with extensions that are manageable, but Smart-Soft can't expect our team to fund the amount of work which is needed for these kind of changes. 

Hence my earlier note in https://github.com/opnsense/core/issues/1899 about suggestions for work items that might be worth our time to review. 

The quality and stability of the core components i very important for us, more issues cost more maintenance which we try to prevent. We have walked a long way and really need to keep the code quality to our standards.

cc @evbevz

kekek2:
@fabianfrz, the model does not contain "keylens" and "digest_algs". These data are used only to generate the CA or certificate. After that they are stored in encrypted and signed form in the certificate. The open form model to store them is not necessary. So for them there is no separate field. And therefore cannot be validated by means of model.

kekek2:
Plan for a phased transfer Trust/Certificates to MVC.
1. Create classes to work with CA and certificates:
`class Cert {
    public function getCrt();
    public function setCrt($crt);
    public function getPrv();
    ....
}
class CA {
    ....
}
`
2. Doing these classes worked with the current config Trust/Certificates.
3. One by one we change in different modules for direct calls to $config["cert"] the challenges of classes Cert/CA/CRL.
4. Do the migration and transition to a new model of Trust/Certificates.
5. One by one changing the page to work with certificates on the Controller/View

AdSchellevis:
@kekek2 You probably better pick something smaller to convert to the mvc framework if you want to help out, for example some of the diagnostic pages are still old styled php.  
Huge projects like this needs quite some investment to coordinate, test and provide feedback.

PR #2061, opened 2018-01-06
HTML Compliance - Attribute "for" on Element <div>

NOYB:
Error: Attribute "for" not allowed on element "div" at this point.
Attributes for element "div": http://www.w3.org/html/wg/drafts/html/master/single-page.html#the-div-element

HTML attribute reference: https://developer.mozilla.org/en-US/docs/Web/HTML/Attributes
The "Output" element: https://developer.mozilla.org/en-US/docs/Web/HTML/Element/output

Attributes for element "output": http://www.w3.org/html/wg/drafts/html/master/single-page.html#the-output-element


fichtner:
Thank you for this! :)

PR #2064, opened 2018-01-07
mv PR from plugins: add known hosts support to OPNsense for SSH hardening

fabianfrz:
This should be extended to do SSH Key management like in Trust and provide settings to plugins. This is for the file `known_hosts` which can be used for public key pinning.

Changes done while moving:
* Correct (C) year -> we have 2018 now.
* Remove second (C) block in volt template

Original PR:
opnsense/plugins#472

### DIFF
--- a/src/etc/inc/auth.inc
+++ b/src/etc/inc/auth.inc
@@ -414,6 +414,18 @@ function local_user_set(&$user, $force_password = false)
 {
     global $config;
 
+    /* load SSH Key DB*/
+    $available_ssh_keys_raw = config_read_array('OPNsense', 'ssh-keys', 'key_pair');
+    if (!is_array($available_ssh_keys_raw)) {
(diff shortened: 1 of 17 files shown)


fabianfrz:
@fichtner I added the functionality to store SSH keys and validate them (at least by the outer format)

fabianfrz:
@fichtner I have added user manager support as well. The old data are not able to be migrated because it had absolutely NO validation. The last step for that is adding it to the place where authorized_keys are written. The big question is if I should add a separate ArrayField to the Model or make Private Key in key_pair optional?

fabianfrz:
User Settings are now looking like this:
screenshot_20180113_102425

fichtner:
I don't think this will cut it, if anything we should migrate ssh away from static pages. SCP backup will be solved in a minimal fashion.

PR #2072, opened 2018-01-10
dhcp6c conf file handling

marjohn56:
Basic dhcp6c conf creation is now separate function. Advanced conf creation now automatically adds any tracked interfaces.

The reasoning behind breaking out the basic conf creation is that with a ( to do ) modification to dhcp6c, a sigup should be sufficient to handle any config change. It also reduces the overall complexity of the interface_dhcp6c_configure() function.


marjohn56:
@fichtner 

I have an update to this, it now also handles a change in the LAN interface(s) such that a change from to or from a tracking interface will re-write the config and trigger a dhcp6c SIGHUP. Shall I just issue a new PR?

marjohn56:
Closed in favour of a more complete option #2090

PR #2073, opened 2018-01-10
Fix multiple dhcp6c issues

marjohn56:
OK, this is the big one. multiple dhcp6c clients were a nightmare on pf, and they are occurring on opn.

The changes in this PR are distilled from the changes I made to pf with some minor adjustments I never bothered to do in pf as it worked, no more stray clients, but there was always room for improvement!

The changes are numerous, wit extra scripts created depending on mode and new functions that handle dhcp6c in a similar way to that of dhclient.

I have switched RTSOLD to one shot, I know of the reason for leaving it running continuously but in many weeks and months of testing on pf with these changes it was never found to be an issue.

Try it and test it thoroughly.

### DIFF
--- a/src/etc/inc/interfaces.inc
+++ b/src/etc/inc/interfaces.inc
@@ -1019,6 +1019,8 @@ function interfaces_configure($verbose = false)
 {
     global $config;
 
+    // Delete this file - Can get left behind if hard shutdown.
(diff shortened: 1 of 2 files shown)


marjohn56:
@fichtner 

Sending you the modified dhcp6c client by email. It now re-reads the config on SIGHUP.

PR #2077, opened 2018-01-10
Fix IPv6 PD auto routing

marjohn56:
Add the missing exec() calls.

### DIFF
--- a/src/sbin/prefixes.php
+++ b/src/sbin/prefixes.php
@@ -60,7 +60,8 @@ foreach ($duid_arr as $entry) {
 // echo "add routes\n";
 if (count($routes) > 0) {
     foreach ($routes as $address => $prefix) {
-        echo "/sbin/route change -inet6 {$prefix} {$address}\n";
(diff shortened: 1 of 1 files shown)


src/sbin/prefixes.php
```diff
@@ -60,7 +60,8 @@
 // echo "add routes\n";
 if (count($routes) > 0) {
     foreach ($routes as $address => $prefix) {
-        echo "/sbin/route change -inet6 {$prefix} {$address}\n";
```
fichtner: hmmm, I think this is it, "route change" was a custom patch a long time ago in a project far far away

fichtner:
Can you try c62f01f993b2ab on master instead? We should still fix the command pipe in abit, this is fishy... but first things first. Very nice catch!

marjohn56:
Appears to work. :)

Applied the patch, manually deleted the route on the primary, reset the secondary WAN and the lease is back...

fichtner:
Thank you, I'll weave in your original changes with the next commit. Very very good find. I owe you a pint. 👍

PR #2080, opened 2018-01-11
HTML Compliance - Lobby: License

NOYB:
Error: Stray end tag span.

Error: No p element in scope but a p end tag seen.
The ordered list makes paragraph tag unnecessary.


fichtner:
Looks good, thanks!

PR #2082, opened 2018-01-11
HTML Compliance - Lobby: Dashboard Widgets

NOYB:
Error: Attribute alt not allowed on element span at this point.

Error: Table columns in range 2…4 established by element td have no cells beginning in them.

Error: Self-closing syntax (/>) used on a non-void HTML element. Ignoring the slash and treating as a start tag.
Error: Row 1 of a row group established by a tbody element has no cells beginning on it.
Warning: A table row was 0 columns wide, which is less than the column count established by the first row (5).

Error: Duplicate ID submita.
Error: Bad value 100% for attribute width on element img: Expected a digit but saw % instead.
Error: Bad value 100% for attribute height on element img: Expected a digit but saw % instead.
Warning: The border attribute is obsolete. Consider specifying img { border: 0; } in CSS instead.

Avoid potential future Error: Duplicate ID submit.
Error: Attribute autocomplete is only allowed when the input type is color, date, datetime-local, email, hidden, month, number, password, range, search, tel, text, time, url, or week.

### DIFF
--- a/src/www/widgets/widgets/picture.widget.php
+++ b/src/www/widgets/widgets/picture.widget.php
@@ -76,7 +76,7 @@ if ($_POST) {
       </tr>
       <tr>
         <td>
-          <input id="submita" name="submita" type="submit" class="btn btn-primary formbtn" value="<?= gettext('Upload') ?>" />
+          <input id="submit_pictfile_widget" name="submit_pictfile_widget" type="submit" class="btn btn-primary formbtn" value="<?= gettext('Upload') ?>" />
         </td>
       </tr>
     </table>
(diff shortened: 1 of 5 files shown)


src/www/widgets/widgets/picture.widget.php
```diff
@@ -76,7 +76,7 @@
       </tr>
       <tr>
         <td>
-          <input id="submita" name="submita" type="submit" class="btn btn-primary formbtn" value="<?= gettext('Upload') ?>" />
+          <input id="submitb" name="submitb" type="submit" class="btn btn-primary formbtn" value="<?= gettext('Upload') ?>" />
```
fichtner: can you please make these a bit more specific to their use case, e.g. "submit_widget"

src/www/widgets/widgets/system_log.widget.php
```diff
@@ -61,7 +61,7 @@
           </select>
         </td>
         <td>
-          <input id="submit" name="submit" type="submit" class="btn btn-primary formbtn" value="<?= gettext('Save') ?>" autocomplete="off">
+          <input id="submite" name="submite" type="submit" class="btn btn-primary formbtn" value="<?= gettext('Save') ?>">
```
fichtner: same here

NOYB:
Updated submit id and name.

fichtner:
splendid, merged. thanks!

Issue #2076, opened 2018-01-10
openvpn: Respect the radius Framed-IP-Address attribute for client ip assignment


EugenMayer:
Yet, one can choose Radius as an authentication for openvpn already, but assigning an IP is yet not possible, since the Framed-IP-Address attribute is yet not respected.

https://wiki.freeradius.org/guide/Ippool-and-radius-clients should be a reference 

![image](https://user-images.githubusercontent.com/136934/34780756-3f8d058e-f624-11e7-81f9-e3fc346dcc83.png)

As far as i understand, one would do ccd for openvpn - at least that is how i was doing that in the past with custom bash foo when booking in a new client. Not sure there is a better way not involving creating files on the FS?

Problably something where we could neat-pick? https://roessner-network-solutions.com/openvpn-radius-mysqlldap-howto/

Offering funding of this one.

mimugmail:
Just to add it here: https://forum.opnsense.org/index.php?topic=5883.0
Would love this too since this would allow "real" user management via API :)

EugenMayer:
We had a chat with @fichtner about how to implement this, i try to rephrase / not down what was "decided" and hope i nail it - if not, punch me.


1. we create a ccd.php in https://github.com/opnsense/core/tree/master/src/etc/inc/plugins.inc.d/openvpn which handle the actuall resolving of ccd meta data. 
So in here, we will ask, who ever wants to provide, in this case its the freeradius server, what the meta-data for the client are. We will handle over the commono name from the auth, which already has been happened ( and succeded ).  After fetching the clients information from <radius> we will write them with https://github.com/opnsense/core/blob/master/src/etc/inc/plugins.inc.d/openvpn.inc#L992 as a ccd entry and return. If not client infos could be fetched, we should *DELETE* the ccd file in the openvpn server ccd folder for this folder

2. we call the ccd.php script from https://github.com/opnsense/core/blob/master/src/etc/inc/plugins.inc.d/openvpn/ovpn_auth_verify#L13 so right after the auth has been verified, passing along all the client details like common name. this will be called always right now, we do not have a "disable ccd" in the openvpn configuration. 

(If we introduce such an option, we would need to pass it on here https://github.com/opnsense/core/blob/master/src/etc/inc/plugins.inc.d/openvpn.inc#L684 )

EugenMayer:
I started with the implementation. I added a checkbox to the openvpn server config "user dynamic ccd lookup" so one can enable and disable this for a openvpn server

[...]

PR #2084, opened 2018-01-12
Implement a flag for openvpn servers to opt-in for dynamic CCD

EugenMayer:
Basically complete for now. Concept is here https://github.com/opnsense/core/issues/2076

This PR just handles the actual flagging of openvpn server to allow dynamic CCD in general - be it by auth or by CURD. Which strategy to take should not be not part of this ( would be needed for both )

### DIFF
--- a/src/www/vpn_openvpn_server.php
+++ b/src/www/vpn_openvpn_server.php
@@ -1529,7 +1529,7 @@ endif; ?>
                           <input name="dynamic-ccd-lookup" type="checkbox" value="1" <?=!empty($pconfig['dynamic-ccd-lookup']) ? 'checked="checked"' : "" ;?> />
                           <output class="hidden" for="help_for_dynamic-ccd-lookup">
                               <span>
-                                <?=gettext("When checked a dynamic lookup for CCD values like ip/netmask will be done with the auth backend like radius"); ?><br />
+                                <?=gettext("When checked a dynamic lookup for Client-Specific-Overrides or so called Client-Config-Dir values like IP/Netmask will be done with the auth backends like RADIUS"); ?><br />
                               </span>
                           </output>
                       </td>


src/www/vpn_openvpn_server.php
```diff
@@ -1523,6 +1523,17 @@
                     <tr>
                       <td colspan="2"><strong><?=gettext("Advanced configuration"); ?></strong></td>
                     </tr>
+                    <tr class="table table-striped opnsense_standard_table_form">
+                      <td style="width:22%"><a id="help_for_dynamic-ccd-lookup" href="#" class="showhelp"><i class="fa fa-info-circle"></i></a> <?=gettext("Use dynamic CCD Lookup"); ?></td>
+                      <td>
+                          <input name="dynamic-ccd-lookup" type="checkbox" value="1" <?=!empty($pconfig['dynamic-ccd-lookup']) ? "checked=\"checked\"" : "" ;?> />
```
fabianfrz: `'checked="checked"'` is better readable

EugenMayer: You mean, i should not copy & paste from above and improve all the ~ 15 cases where this looks like that? :)

I fixed that, but i am serious, i am just copying code and i am not really looking to improve a file which is a template, controller and inline-JS defintion is "stylistics". Its like it is, would be my way to deal with it :)

src/www/vpn_openvpn_server.php
```diff
@@ -1523,6 +1523,17 @@
                     <tr>
                       <td colspan="2"><strong><?=gettext("Advanced configuration"); ?></strong></td>
                     </tr>
+                    <tr class="table table-striped opnsense_standard_table_form">
+                      <td style="width:22%"><a id="help_for_dynamic-ccd-lookup" href="#" class="showhelp"><i class="fa fa-info-circle"></i></a> <?=gettext("Use dynamic CCD Lookup"); ?></td>
+                      <td>
+                          <input name="dynamic-ccd-lookup" type="checkbox" value="1" <?=!empty($pconfig['dynamic-ccd-lookup']) ? 'checked="checked"' : "" ;?> />
+                          <output class="hidden" for="help_for_dynamic-ccd-lookup">
+                              <span>
+                                <?=gettext("When checked a dynamic lookup for CCD values like ip/netmask will be done with the auth backend like radius"); ?><br />
```
fabianfrz: Use the not abbreviated version of CCD in the help text if possible and upcase the following words:
* IP
* RADIUS

EugenMayer: sure, done

EugenMayer:
even though i closed https://github.com/opnsense/plugins/pull/485 i would keep this one open since this will make sense for this particular task anyway. Of course, if not desired, close this one too

csszep:
Is there any plan to finish this feature?

This is the last pfsense feature i'm using, and blocking the migration :)

Thx
csszep

fichtner:
master was reworked to support this another way

PR #2085, opened 2018-01-13
HTML Compliance - MVC Form Input - Attribute "for" on Element <tr,div,a,small,span>

NOYB:
/usr/core/src/mvc/app/views/layout_partials/form_input_tr.volt
Error: Attribute "for' not allowed on element <tr> at this point.
Error: Attribute "for" not allowed on element <div> at this point.
Error: Attribute "for" not allowed on element <a> at this point.
Error: Attribute "for" not allowed on element <small> at this point.
Error: Attribute "for" not allowed on element <span> at this point.

Error: Duplicate ID "clear-options".

### DIFF
--- a/src/opnsense/mvc/app/views/layout_partials/form_input_tr.volt
+++ b/src/opnsense/mvc/app/views/layout_partials/form_input_tr.volt
@@ -42,16 +42,16 @@ readonly    :   if true, input fields will be readonly
 
 #}
 
-<tr for="{{ id }}" {% if advanced|default(false)=='true' %} data-advanced="true"{% endif %}>
+<tr id="row_{{ id }}" {% if advanced|default(false)=='true' %} data-advanced="true"{% endif %}>
     <td >
-        <output class="control-label" for="{{ id }}">
+        <div class="control-label" id="control_label_{{ id }}">
             {% if help|default(false) %}
                 <a id="help_for_{{ id }}" href="#" class="showhelp"><i class="fa fa-info-circle"></i></a>
             {% elseif help|default(false) == false %}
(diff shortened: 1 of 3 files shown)


AdSchellevis:
@NOYB I'm not sure about the padding, but functionally the patch is fine. The old non compliant setting used a padding-top of 0px, output has a 7px top padding by default it seems, which seem to make it all a bit spacey. @fichtner what do you think? 

I'll pull the changes in, we can decide what todo with the padding later.

NOYB:
Took a look at the top spacing in services captive portal administration edit zone and not sure which I like better.  Without the extra top space it kind of seem to run together.  So guess maybe I give the edge to the extra top space.  But really no big deal either way.

If removing the extra top space is desired here are a couple possibilities.

1)
output[for*="help_for"] {
 padding-top: 0px; }

2)
Replace &lt;output&gt; element and "for" attribute with &lt;div&gt; element and "data-for" attribute.
Requires changing the "for" attribute selector to "data-for" in initFormHelpUI() (opensense_ui.js).

Thanks for the merge.

NOYB:
Kind of like this to remove the top space and yet keep it from running together by including some bottom space to create some separation from the next option.
output[for*="help_for"] {
	padding-top: 0px;
	padding-bottom: 7px; }

NOYB:
Like this even better.  This is my favorite.
output[for*="help_for"] {
 padding-top: 7px;
 padding-bottom: 7px; }

AdSchellevis:
@NOYB I just tried some different options too and agree with the last one, both 0.4em seems to align nicely.  I'll push a commit for this.

PR #2088, opened 2018-01-13
add WPAD to dhcp server

fabianfrz:
It deploys the default hostname:

```
nmap --script broadcast-dhcp-discover -e vboxnet1

Starting Nmap 7.60 ( https://nmap.org ) at 2018-01-13 21:53 CET
Pre-scan script results:
| broadcast-dhcp-discover: 
|   Response 1 of 1: 
|     IP Offered: x.x.x.x
|     DHCP Message Type: DHCPOFFER
|     Server Identifier: x.x.x.x
|     IP Address Lease Time: 5m00s
|     WPAD: https://wpad.sense:443/wpad.dat
|     Subnet Mask: 255.255.255.0
|     Router: x.x.x.x
|     Domain Name Server: x.x.x.x
|_    Domain Name: sense
WARNING: No targets were specified, so 0 hosts scanned.
Nmap done: 0 IP addresses (0 hosts up) scanned in 1.32 secon
```

### DIFF
--- a/src/etc/inc/services.inc
+++ b/src/etc/inc/services.inc
@@ -432,6 +432,7 @@ option domain-name "{$config['system']['domain']}";
 option ldap-server code 95 = text;
 option domain-search-list code 119 = text;
 option arch code 93 = unsigned integer 16; # RFC4578
+option pac-webui code 252 = text;
 {$custoptions}
 default-lease-time 7200;
 max-lease-time 86400;
@@ -786,6 +787,19 @@ EOD;
         // tftp-server-name
         if (!empty($dhcpifconf['tftp'])) {
             $dhcpdconf .= "  option tftp-server-name \"{$dhcpifconf['tftp']}\";\n";
+        }
+
+
+
+        // add pac url if it applies
+        if (isset($dhcpifconf['wpad']) && !empty($config['system']['hostname']) && !empty($config['system']['domain'])) {
+            $protocol = !empty($config['system']['webgui']['protocol']) ? $config['system']['webgui']['protocol'] : 'https';
+            // take hostname from system settings - it can be used to be resolved to anything based on client IP
+            $host = implode('.', array($config['system']['hostname'], $config['system']['domain']));
+            $port = !empty($config['system']['webgui']['port']) ? $config['system']['webgui']['port'] : 443;
+            $webui_url = "$protocol://$host:$port/wpad.dat";
+
+            $dhcpdconf .= "  option pac-webui \"$webui_url\";\n";
         }
 
         // Handle option, number rowhelper values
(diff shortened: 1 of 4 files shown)


src/www/services_dhcp.php
```diff
@@ -962,6 +963,13 @@ function removeRow() {
                         </div>
                       </td>
                     </tr>
+                    <tr>
+                      <td><i class="fa fa-info-circle text-muted"></i> <?=gettext("WPAD");?> </td>
+                      <td>
+                        <input name="wpad" id="wpad" type="checkbox" value="yes" <?=!empty($pconfig['wpad']) ? "checked=\"checked\"" : ""; ?> />
+                        <strong><?= sprintf(gettext("Enable Web Proxy Auto Discovery on the  %s interface"),!empty($config['interfaces'][$if]['descr']) ? htmlspecialchars($config['interfaces'][$if]['descr']) : strtoupper($if));?></strong>
```
fichtner: "Enable Web Proxy Auto Discovery on this interface", the rest is too confusing (and there is a double space in front of %s anyway)

src/etc/inc/services.inc
```diff
@@ -788,6 +789,19 @@ EOD;
             $dhcpdconf .= "  option tftp-server-name \"{$dhcpifconf['tftp']}\";\n";
         }
 
+
+
+        // add pac url if it applies
+        if (isset($dhcpifconf['wpad']) && !empty($config['system']['hostname']) && !empty($config['system']['domain'])) {
```
fichtner: Can't we set host names / domains per interface?

src/etc/inc/services.inc
```diff
@@ -788,6 +789,19 @@ EOD;
...
+            $protocol = !empty($config['system']['webgui']['protocol']) ? $config['system']['webgui']['protocol'] : 'https';
+            // take hostname from system settings - it can be used to be resolved to anything based on client IP
+            $host = implode('.', array('wpad', $config['system']['domain']));
```
fichtner: hostname is checked in first if, but not added here?

src/etc/inc/services.inc
fabianfrz: theoretically yes but is there a reason to? It must point to an IP of the Firewall and the system hostname makes that always true. On that interface the wpad.dat file must be served.

fichtner: true

src/etc/inc/services.inc
fabianfrz: Hosts try the hostname `wpad` in the domain. which is created here. This one needs DNS support.

fichtner: but why the sanity check of !empty($config['system']['hostname']) ? it should never be empty and if it is WPAD not working is the least of our issues

fabianfrz: hostname is used for the CNAME in Unbound:
https://github.com/opnsense/core/pull/2097/files#diff-a89985242e1eea6a91d3e103e3353d5cR586

src/etc/inc/services.inc
```diff
@@ -788,6 +789,19 @@ EOD;
...
+            $port = !empty($config['system']['webgui']['port']) ? $config['system']['webgui']['port'] : 443;
```
fichtner: 443 is only true for https :)

fabianfrz: fixed

fichtner:
Merged, thanks!

PR #2090, opened 2018-01-14
dhcp6c handling, multiple LAN tracking, advanced dhcp6 fix.

marjohn56:
This commit corrects an issue when using advanced dhcp6c settings in that it now correctly configures the LAN interface(s) names. It also allows multiple LAN interfaces  to be used.

A complete break out of the configuration creation and script creation has been done, this is because to correctly allow for a change to a LAN interfaces that may affect tracking, the script creation MUST take place in the interface_configure routine  as a change to LAN tracking must be caught and the configs adjusted before calls to interfaces_dhcp6_configure.

New routines have been added that launch or reset dhcp6c plus kill and find & reset. These are all there to try and mitigate the chances of multiple dhcp6c clients. The kill_dhcp6c_client_process takes advantage of feature added to dhcp6c last year. Using the USR1 signal will force dhcp6c to exit cleanly but with no release. This is important, as a lot of the issues with dhcp6c are caused by timing. On a normal TERM  kill, dhcp6c will try to release any assigned values by signalling the server and waiting for a response, this takes time. Thus if we are taking down the interface or making changes we want an immediate exit and as a new REQUEST will be sent we are not interested in releasing, thus dhcp6c exits quickly.

RTSOLD has been changed to a one shot. I am aware of a previous issue but the clean-up here around dhcp6c should mean that there is no requirement for it to be resident.


marjohn56:
Blimey that was quick!

src/etc/inc/interfaces.inc
```diff
@@ -3195,88 +3309,202 @@ function interface_dhcpv6_configure($interface = 'wan', $wancfg)
...
+                if (isset($wancfg['dhcp6incslaid'])) {
+                    $sla_id = $count;
+                } else {
+                    $sla_id = 0;
+                }
+                if (is_numeric($ifcfg['track6-prefix-id'])) {
+                    $realif = get_real_interface($friendly);
+                    $dhcp6c_conf .= "  prefix-interface {$realif} {\n";
+                    $dhcp6c_conf .= "    sla-len {$wancfg['dhcp6-ia-pd-len']};\n";
+                    $dhcp6c_conf .= "    sla-id {$sla_id};\n";
```
fichtner: is there any reason why we omit track6-prefix-id usage here now? that's a detail I don't want to change without further tweaking, e.g. allowing the value to be empty in the GUI so this can be auto-assigned?

fichtner:
Thanks for the cleanups, I'll merge this early next week.

src/etc/inc/interfaces.inc
marjohn56: No, that's a mistake on my part, it should follow the config. Sorry, Are you OK to correct that?

fichtner: I will cherry-pick and modify a tiny bit then, no need to change this PR. Ok? :)

marjohn56: Your're comment elsewhere about dhcp6c being slow to exit reminded me of the reason for the  addition of the USR1 sig,

marjohn56: That's cool Franco. :)

fichtner: 👍

marjohn56: I wanted to use a php call to the run_dhcp6client_process() from within the rtsold script, but although I could pass the params, I had to include util.inc and even then it would not correctly get all the variables. It would have been nice though to have only one place where dhcp6c can be launched.

marjohn56:
fixed the bug you saw, also fixed another where ia-na request was being sent even though only an ia-pd was selected.

marjohn56:
fixed a path issue in rtsold script creation

src/etc/inc/interfaces.inc
```diff
@@ -3196,87 +3310,197 @@ function interface_dhcpv6_configure($interface = 'wan', $wancfg)
...
-        "/usr/local/sbin/dhcp6c %s -c %s -p %s %s",
-        array(
-            '-' . (empty($wancfg['adv_dhcp6_debug']) ? 'd' : 'D' ) . (!empty($wancfg['dhcp6norelease']) ? 'n' : ''),
-            "/var/etc/dhcp6c_{$interface}.conf",
-            "/var/run/dhcp6c_{$wanif}.pid",
-            "{$wanif}"
-        )
-    );
+    /* non ipoe Process */
+    if (!isset($wancfg['dhcp6sendsolicit'])) {
```
fichtner: problem with this flip/flop logic is that rtsold is the only source of a gateway and if we don't even run rtsold we never get it... that's ok if there is never a router advertisement, but if there is one after SOLICIT we end up missing it.

src/etc/inc/interfaces.inc
```diff
@@ -1019,6 +1019,8 @@ function interfaces_configure($verbose = false)
 {
     global $config;
 
+    // Delete this file - Can get left behind if hard shutdown.
+    @unlink("/tmp/dhcp6c_{$realif}_lock");
```
fichtner: can't find any reference to this in the diff?

src/etc/inc/interfaces.inc
marjohn56: RTSOLD is always run. In one mode it runs and it launches dhpc6c, in the other mode dhcp6c runs and it launches rtsold.

Try it. :)

marjohn56:
Removed unused lock file deletion reference.

fichtner:
@marjohn56 can we split out the smallest required atomic part and reiterate until we have all we need? I see some code that seems to be borrowed elsewhere and in part seems to revert the code back to where it was in terms of code style... both of which make me doubt we are going forward with the whole of this patch :/

marjohn56:
It needs it all as a whole to work. OK, let's break this down into parts.

There are new functions that handle the starting, killing and process finding of dhcp6c. Those a pretty straight forward, the reason for them is to help mitigate multiple dhcp6c clients. Using them takes advantage of the extra bits that exist in dhcp6c that opnsense is not currently using,  See line 2524, it explains it. The Key is the use of SIGUSR1. There is also a function reset_dhclient_process(), This uses the additional function I have added to dhcp6c to re-read the conf on sigup, thus if the process is running we do not kill it, we just send a sigup and let it carry on, no chance of a new dhcp6c process.

The next major breakout is the function **interface_dhcpv6_configure**, all I have done there is break out all of the configuration functions from that call and there is a very good reason for this. Whenever LAN config changes, i.e. tracking, then that change MUST be reflected in an update to dhcp6c_*.conf, but we do not always want to call a interface_dhcpv6_configure, especially not at that point, the call to re-write the config is in interface_configure. I have made a comment at around line 2747 that explains it. 

Now, we also need to take into account that there may be an advanced dhcp6c needed, this is also handled in the function configure_dhcp6c_conf()  which calls the config routines needed, whether they are  basic or advanced_config functions, it also writes the RTSOLD script as needed.

Now, I know you have this thing about rtsold needing to be running constantly, and I have removed the -1 parameter to leave it resident, but the issue there is that if there is an RS before dhcp6c solicit and know for a fact that Sky UK will then ignore the dhcp6c request, hence scripts either have dhcp6c launching RTSOLD or RTSOLD launching DHCP6C.

I am happy if you wish to break this all down into separate commits, but they do form a whole process.

Where do you see code regression?

marjohn56:
Slept on it...

OK.. First thing we need to do before any of this commit is to get the dhcp6c changes integrated and sort out dhclient vlan priority. Once those are done I will re-work  this commit so it's done in separate stages. The dhcp6c changes I have already committed, dhclient appears to use raw sockets hence it;s refusal to play nicely.

marjohn56:
Ignore this PR, breaking down into smaller chunks and will then delete

fichtner:
@marjohn56 Thanks, much appreciated. We still have the dhclient patch on this branch.... https://github.com/opnsense/src/commits/dhclient_77

I'm not opposed to this, but it needs testing and maybe fixing as I've tried to break it down and simplify as much as possible.

marjohn56:
Not sure what you mean with 'I'm not opposed to this, but it needs testing and maybe fixing'. Do you mean dhclient?

fichtner:
yes, patching dhclient to enable vlan because set-prio doesn't seem to work. it's not a straightforward way, but we can always reiterate and find a better solution later when we have a working base implementation. :)

marjohn56:
OK, I'll take a gander...

fichtner:
Also continued in #2258

Issue #2092, opened 2018-01-16
Widget controls not working correctly with touch screen input


NOYB:
Chrome:
The title link and buttons work with touch input.  Though vertical position offset a little bit.  Have to touch slightly below them.
Drag and drop sorting doesn't work with touch input.
None of them work at all with active stylus input.

IE:
None of them work at all with touch.
Only drag and drop sorting works with stylus input.

Adding delay: 50, and distance: 2 options to the jquery sortable function gets the link and buttons to work with both touch and active stylus input.
Kind of a cursor grabbing action.

Adding "touch-action: none" css to the sort handle header element gets the drag and drop sorting working with touch input.

I'll make a commit for these changes which fixes this for IE.  But Chrome is still not working.

A bit surprised that if this is not working in Chrome it hasn't been reported.  So maybe it's just my Chrome environment that is not working properly.  Can someone confirm that the widget title link, settings, minimize, and remove buttons all work as well as drag and drop sorting works with touch screen input?

NOYB:
Added a modification that works with both Android and Desktop Chrome.
https://github.com/opnsense/core/pull/2137

So nice to have touch access to the widget settings.

fichtner:
Submitted and subsequently released into 18.1.x

PR #2093, opened 2018-01-16
Widgets Controls on Touch Screen

NOYB:
Touch events not working with the widget controls (link, buttons, drag/drop sort).
Add distance, delay and touch-action css for the widget controls to work with touch screen input.

Widget controls not working correctly with touch screen input
https://github.com/opnsense/core/issues/2092

### DIFF
--- a/src/www/index.php
+++ b/src/www/index.php
@@ -276,6 +276,8 @@ include("fbegin.inc");?>
       // sortable widgets
       $(".dashboard_grid_column").sortable({
         handle: '.content-box-head',
+        delay: 50,
+        distance: 2,
         group: 'dashboard_grid_column',
         itemSelector: '.widgetdiv',
         containerSelector: '.dashboard_grid_column',
@@ -369,7 +371,7 @@ include("fbegin.inc");?>
           }?>
           <section class="widgetdiv" data-sortkey="<?=$widgetItem['sortKey'] ?>" id="<?=$widgetItem['name'];?>"  style="display:<?=$divdisplay;?>;">
             <div class="content-box">
-              <header class="content-box-head container-fluid">
+              <header class="content-box-head container-fluid" style="touch-action: none">
                 <ul class="list-inline __nomb">
                   <li><h3>
 <?php


fichtner:
Thanks. Does not seem to do much on iOS with Safari, but there is definitely room for improvement as the controls are behaving weirdly indeed.

NOYB:
A brief internet search turns up lots of people having the same problem with the jquery sortable in Chrome on touch screens.  I trust this fix will not be delayed due to those other browsers not working.  It has been an ongoing issue with them for a long time with no solution in sight.  IE should not be penalized for their sake.  Requesting that this commit be included in the 18.1 release.  Thanks.

fichtner:
If I had that feedback 2 days ago, yes. Now 18.1 is all built and pending release. Sorry. I’ll see to inclusion in 18.1.1 if this works on mobile Chrome, Firefox and Edge.

NOYB:
This does not break Chrome on desktop or Android.  But there is no fix for Chrome.  Search shows Chrome not working with jquery sortable for a long time and no solution in sight.

So please do not delay and withhold this from others for Chromes sake.

Rant:
Chrome needs to stop going down their own non industry specification proprietary path like they did with using the viewport width=device-width (or less) for an unintended purpose to disable double tap zoom.  Now the viewport width=device-width cannot be used for it's intended purpose without it also disabling double tap zoom.  Which is undesirable.  Double tap zoom is a useful and convenient capability even when the viewport width is set to device width or less.

1) Quick and convenient zoom in on something.  Just because viewport width is device width doesn't mean zooming is not useful and desirable.  And double tap is more convenient than pinch.

2) Quick and convenient return to scale factor from a zoom-in.

3) I'm sure there are more.

Was tinkering with the viewport width last night.  Setting it to 1 grater than the device-width to prevent Chrome from disabling double tap zoom.  So convenient to have the double tap zoom/return capability.

fichtner:
Note my initial response was on behalf of Safari, but then what you are seeing on Chrome seems to indicate it's a Webkit issue. Which is ok, but it's also a lot of users combined.

I haven't had the time to verify the former state. From initial tests it seemed to do nothing (mea culpa for testing on the wrong mobile platforms). Still, if I can't verify immediately or at least double-check that the non-working behaviour is the same without the patch, there is no apparent reason to merge into a release something that could be good or bad or both at the same time.

It's good that you are convinced of your work and maybe I'm not doing it the justice it deserves in this particular case, but want to feel the same motivation about what is being proposed so be able to pull it into a release. And besides, 18.1.1 is not so far out. :)

NOYB:
Well it only does two things.

1) Disables touch action on the widget headers.  Which should already be the case for Chrome via the viewport width setting.  Don't know what iOS and Safari are doing.  Probably nothing.  So until they implement the touch-action css property it probably shouldn't have any affect.  This is a good read on iOS 8 and the tap delay.  Gets into iOS 8 solution about half way down.  https://developer.telerik.com/featured/300-ms-click-delay-ios-8/

2) Setting the JQuery sortable delay and distance options.

But yeah I understand and appreciate your position wanting to know how high the bridge is be for leaping.

Issue #2023, opened 2017-12-26
WPAD support


fabianfrz:
The PAC related stuff is in #2018 and this file can be deployed automatically using DHCP and DNS which might be both controlled by OPNsense.

* In case of DHCP we need option 252 to be a URL to the file (with #2018 this is /wpad.dat)
* and in case of DNS we need to set the address of wpad.example.com to the IP of OPNsense from which the file wpad.dat (which is the reason for the name) can be downloaded via Port 80. If the web configuration runs on that port this works too - else it will probably work via the HAProxy Plugin forwarding it to the right port.

For WPAD,
* DHCP should get an option (checkbox) to add the option automatically and
* Unbound should get a Host Override via Checkbox or something like that.

fabianfrz:
DHCP is in #2088

fabianfrz:
And DNS in #2097

PR #2097, opened 2018-01-16
Unbound: add WPAD

fabianfrz:
part of #2023


src/www/services_unbound.php
```diff
@@ -267,6 +270,16 @@
                           </output>
                         </td>
                       </tr>
+                      <tr>
+                        <td><a id="help_for_enable_wpad" href="#" class="showhelp"><i class="fa fa-info-circle"></i></a> <?=gettext("WPAD Records");?></td>
+                        <td>
+                          <input name="enable_wpad" type="checkbox" value="yes" <?=!empty($pconfig['enable_wpad']) ? "checked=\"checked\"" : "";?> />
+                          <output class="hidden" for="help_for_enable_wpad">
```
fichtner: we use a different style for help text now

fichtner:
General question: is dnsmasq capable of this? or is this an unbound-only feature?

fabianfrz:
I have not tried dnsmasq but as far as I know it can fake the A/AAAA record by adding the IP / Host to the hosts file.

fichtner:
Merged, thanks!

PR #2098, opened 2018-01-17
Interfaces Widget Name & Title Link

NOYB:
Widget names are derived from the widget file base name, which in turn is used to obtain the widget title link.
This restores the original interfaces widget file base name, which fixes the broken title link.


fichtner:
Looks good, thank you!

fichtner:
Okay, not so nice, interface_list is used in the default config and we need to realign everything accordingly...

fichtner:
This should do it 3bf091ed513 :)

PR #2099, opened 2018-01-17
Firewall Logs Widget Tweaks

NOYB:
Widget purpose is to provide an overview, i.e. abbreviated information, of the firewall log.  Details are available in the firewall log view.
  Source port is not all that interesting.  Consumes considerable space for not much value in this view.
  Remove seconds from time field.  Verbosity is available in the firewall log view.

Apply "table-condensed" class for better space utilization.

Settings:
  Provide some separation between check boxes and make their labels active.  Better accessibility and touch screen friendliness.


AdSchellevis:
@NOYB looks good, thanks.

PR #2107, opened 2018-01-18
Firewall Logs Widget - Add Update Interval Option

NOYB:
Add update interval configuration option.

### DIFF
--- a/src/www/widgets/widgets/log.widget.php
+++ b/src/www/widgets/widgets/log.widget.php
@@ -36,6 +36,7 @@ require_once("interfaces.inc");
 
 if (is_numeric($_POST['filterlogentries'])) {
     $config['widgets']['filterlogentries'] = $_POST['filterlogentries'];
+    $config['widgets']['filterlogentriesupdateinterval'] = $_POST['filterlogentriesupdateinterval'];
 
     $acts = array();
     if ($_POST['actpass']) {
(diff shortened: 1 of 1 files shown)


src/www/widgets/widgets/log.widget.php
```diff
@@ -179,6 +181,22 @@ function fetch_log(){
...
+            <td>
+              <?= gettext('Update interval (seconds):') ?>
+            </td>
+          </tr>
+          <tr>
+            <td>
+              <select name="filterlogentriesupdateinterval" class="formfld unknown" id="filterlogentriesupdateinterval">
+<?php
+              for ($i = 1; $i <= 60; $i++):?>
+                <option value="<?=$i?>" <?= $updateinterval == $i ? "selected=\"selected\"" : ""?>><?=$i;?></option>
```
fabianfrz: maybe display with a unit (note: untested):
```php
<?=  sprintf($i == 1 ? gettext('%s Second') : gettext('%s Seconds'), $i) ?>

NOYB: Doable I'm sure.  Though think it adds clutter.  The units being noted in the label and then just the number alone in the selector is cleaner.

In context, along with the label units note, it should be pretty clear to target audience that it is seconds without suffixing it.

Thought about not including 1 anyway.  That's pretty intense updating.  But figured somebody, somewhere, sometime, for some obscure reason will want it.  So who am I to withhold it from them.

Though it may be a good idea to add a test when setting the time out loop to prevent anything less than 1 second from being scheduled.  Just in case of an issue such as failing to read the element value for some reason.

fabianfrz: What do you think about using input type number?
You could set `min` and `max` as a property and it would be more mobile friendly if @fichtner is fine with it.

https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/number

NOYB: Had considered using input rather than selector but wanted to be consistent with the other inputs.  In particular the number of entries selector.  I'm okay either way, but whichever way is done I think those two should use the same method.

Maybe we get the update interval option in, then change the method on those two it that is what's desired.

AdSchellevis:
@NOYB thanks!

PR #2112, opened 2018-01-19
pppoe Virtual IP fix #2

### LINKED ISSUE #2: should not ping the internet by default
this revolves around the initial apinger setup. via @timelessp (Twitter)

marjohn56:
PPPoE VIP fix

First version broke IPv6, this corrects that and leaves things intact unless the VIP is v4 AND is pppoe.

With pppoe we do not know the gateway in advance, so when ifconfig'ing, we determine if we have a pppoe connection on the interface and get the gateway from the gateway file.

This is for those users who want to have those IP addresses on the interface. Maybe and option in the GUI?

### DIFF
--- a/src/etc/inc/interfaces.inc
+++ b/src/etc/inc/interfaces.inc
@@ -1823,6 +1823,7 @@ function interface_ipalias_configure(&$vip)
         return;
     }
 
+    $wancfg = $config['interfaces'][$vip['interface']];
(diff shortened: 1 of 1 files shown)


marjohn56:
Forget this, it;s working perfectly well. If someone wants the real ip's on the WAN or PPPoE interface it can be resurrected.

githubatf2f10:
no worries,.....called TekSavvy and they found it out this subnet was assigned to another person as well, which caused packet being missed.. they gave me a new subnet and with only port-forward, no 1:1 nat, things work well now. thanks

PR #2115, opened 2018-01-20
HTML Compliance - Attribute "content" value on Element <meta>

NOYB:
Error: A meta element with an http-equiv attribute whose value is X-UA-Compatible must have a content attribute with the value IE=edge.

The "Chrome=1" value was used to spawn Chrome Frame in IE 6,7,8,9.  However, "Google Chrome Frame was discontinued in January 2014", and think all IE<11 is beyond EOL.
So is this even needed/useful?

If needed/useful it would probably be better to support deployment by HTTP headers method instead.

Google Chrome Frame
https://en.wikipedia.org/wiki/Google_Chrome_Frame

### DIFF
--- a/src/etc/inc/authgui.inc
+++ b/src/etc/inc/authgui.inc
@@ -356,7 +356,7 @@ function display_error_form($http_code, $desc)
   <head>
 
     <meta charset="UTF-8" />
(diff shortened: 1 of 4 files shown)


fichtner:
Looks good, thanks!

Issue #2103, opened 2018-01-17
Minor: do not open openvpn client in same window


EugenMayer:
When clicking on any of those clients, it loads in the same window. Should be _blank rather

![image](https://user-images.githubusercontent.com/136934/35049702-31edbe78-fba1-11e7-8560-ee237de1f301.png)

fabianfrz:
Created a fix for your issue.

fichtner:
Fixed in 18.1.1. Thanks!

PR #2116, opened 2018-01-20
fix: add js code to open links in a new tab

fabianfrz:
fix for #2103


fichtner:
with all due respect using javascript here is overkill ;)

fabianfrz:
@fichtner May be better now.

fichtner:
merged, thank you!

PR #2124, opened 2018-01-23
Fix CSS paths rev #2

marjohn56:
change css paths in default.volt and head.inc to use theme paths.

Copy *.css files from opensense/www/css to opnsense/www/themes/opnsense/build/csss EXCEPT the *font.css


fichtner:
So I agree with this, but we need to make sure we put a symbolic link to the old location in order to not break any other pages that would expect those files where they are. Let me take care of this post-merge this afternoon.

fichtner:
I revise that last comment: break existing themes that users may have created and that are missing those files... a symbolic link will not entirely fix it... mhhh...

marjohn56:
I did do a global search for all references and I think I have got them all. 

That was also the reason for leaving the existing css files where they are too. :)

AdSchellevis:
I agree we should move all css into the theme directory, but we should add some glue here to use the default when there's none specified to avoid breakage. I'll let this here for a moment,  as soon I'll have time to fix the "use default" I will pull this in.

marjohn56:
It will take you far less time to do it than me and it will be correct first time. :)

AdSchellevis:
@marjohn56 ok, no problem, I'll keep this PR open until it's fixed.

marjohn56:
The only other part of the themes that needs correction, but will need a full re-write as it's old style is the *gateways*.php files. Currently they are using fixed colours and I have no idea where to start on that.

AdSchellevis:
@marjohn56 this https://github.com/opnsense/core/commit/2461ca2a8504671d76946f37f7ab68d8964423a8 should provide you with the possibility to override the css files in your theme. @fichtner this won't solve the duplication of fonts, because their compiled into the theme css. Maybe we better leave those as they are.

PR #2131, opened 2018-01-26
Optimize exec('/usr/local/bin/openssl version')

kekek2:


fichtner:
Hi @kekek2,

I've seen this before in one of your PRs and wasn't completely convinced. The problem I have is that we rely on php71-openssl to tell us the truth during compile time... not runtime... but we really need to know the state of the binary to know which crypto is installed. Or is that not problem from your experience?


Cheers,
Franco

fichtner:
Pondering further... I'd like to see this in action on opnsense-devel so I'm merging... thanks!

PR #2133, opened 2018-01-26
Double Tap Zoom w/o 300 ms Delay

NOYB:
Eliminate 300 ms tap delay only on user action elements, classes, and ID's.
This permits double tab zoom to remain enabled elsewhere.

Prevent Chrome from disabling double tap zoom based on viewport width <= device-width.

This is something I've been tinkering with on Android Chrome.  Being able to double tap zoom or return to scale 1 from a pinch zoom is very convenient.

Don't know what your interest in this would be, but it will be in my repertoire for sure.


src/etc/inc/authgui.inc
```diff
@@ -432,6 +439,13 @@ function display_login_form($Login_Error)
     <meta name="copyright" content="" />
     <meta name="viewport" content="width=device-width, initial-scale=1, minimum-scale=1" />
 
+    <script>
+      // Prevent Chrome from disabling double tap zoom based on viewport width <= device-width.
+      var device_width = screen.width + 1;
+      var viewport = document.querySelector("meta[name=viewport]");
+      viewport.setAttribute('content', 'width='+ device_width +', initial-scale=1, minimum-scale=1');
+    </script>
+
```
fabianfrz: It would be better to extract some parts into a partial.

src/www/head.inc
```diff
@@ -42,6 +42,13 @@ $pagetitle .= html_safe(sprintf(' | %s.%s', $config['system']['hostname'], $conf
     <meta name="copyright" content="" />
     <meta name="viewport" content="width=device-width, initial-scale=1, minimum-scale=1" />
 
+    <script>
+      // Prevent Chrome from disabling double tap zoom based on viewport width <= device-width.
+      var device_width = screen.width + 1;
+      var viewport = document.querySelector("meta[name=viewport]");
+      viewport.setAttribute('content', 'width='+ device_width +', initial-scale=1, minimum-scale=1');
+    </script>
+
```
fabianfrz: use the partial form authgui.inc when created

NOYB: Can you point me at an example I can follow?
Thanks

NOYB:
Changed touch-action: none; to touch-action: manipulation;
touch-action: none; prevented scrolling in nav menu.

src/www/head.inc
fabianfrz: What I am thinking about is having a function to generate the head stuff that is identical at every place. In that case you just need to have your change at that place and a second time for MVC.

probably a function like

```php
<?php
function create_head_base_content($themename) {
?>
    <meta charset="UTF-8" />
    <meta http-equiv="X-UA-Compatible" content="IE=edge">

    <meta name="robots" content="noindex, nofollow, noodp, noydir" />
    <meta name="keywords" content="" />
    <meta name="description" content="" />
    <meta name="copyright" content="" />
    <meta name="viewport" content="width=device-width, initial-scale=1, minimum-scale=1" />

    <script>
      // Prevent Chrome from disabling double tap zoom based on viewport width <= device-width.
      var device_width = screen.width + 1;
      var viewport = document.querySelector("meta[name=viewport]");
      viewport.setAttribute('content', 'width='+ device_width +', initial-scale=1, minimum-scale=1');
    </script>

    <link href="/ui/themes/<?= $themename ?>/build/css/main.css" rel="stylesheet">
    <link href="/ui/themes/<?= $themename ?>/build/images/favicon.png" rel="shortcut icon">

    <script src="/ui/js/jquery-3.2.1.min.js"></script>
    <script src="/ui/js/jquery-migrate-3.0.1.min.js"></script>

<?php
}
?>
```

So the other pages don't need to include have the same headers - you can simply do something like this:
```php
<?php create_head_base_content($themename); ?>
```

fabianfrz: It is just for not repeating any stuff on any page.  The block does not include the title tag or something page specific. Maybe even more JS libraries can be moved inside.

NOYB:
Where do you think the function should reside?  index.php? guiconfig.inc?  Other?  Kind of reluctant to corrupt guiconfig.inc with html.

Also if this is to be done I think it should be a separate effort not dependent on this.  Seems like it is something that would have value in its own right.  If that is something Franco would like.

fabianfrz:
@fichtner probably knows best. At least it should be as [DRY](https://en.wikipedia.org/wiki/Don%27t_repeat_yourself) compliant as possible in my opinion.

src/www/head.inc
NOYB: Have this working on the legacy side.  Don't know how to deal with the PHP function in MVC side.  Where to put the PHP and how to properly include/call it from in the .volt files.  Need some guidance.

On the legacy side I have the PHP function in authgui.inc

fabianfrz: MVC supports partials: by default but as far as I know there is no need for that because there is only a single layout.

Here is an example for how to use a partial:
https://github.com/opnsense/core/blob/master/src/opnsense/mvc/app/views/layout_partials/base_form.volt#L104

NOYB: There is the MVC default and captive portal has it's own head as well.  Maybe captive portal needs to be reworked to use the default.

NOYB:
Changed to using document.documentElement.clientWidth instead of screen.width.
Better for landscape mode where devices typical have some of the screen width allocated to soft buttons.

NOYB:
Updated to adjust for orientation change.

Fabian's suggestion to create/use head "partial":
https://github.com/opnsense/core/pull/2145
If accepted, will make use of it for the viewport width adjustment script.

PR #2134, opened 2018-01-26
Translate plugins descriptions.

kekek2:


fabianfrz:
API calls should never be translated because this may break other application that expect the responses in english.

fichtner:
I spend a bit of time to weigh this and I agree with @fabianfrz as the translation effort here is too fragile.

We could restructure pkg-descr to provide multiple languages, but even that effort is going to be an uphill battle.

I'd rather have it we could leave this in English. We will also start adding release notes to pkg-descr so it will be even harder to maintain.

Closing this without merge but open for suggestions.

PR #2137, opened 2018-01-27
Widgets Controls on Touch Screen - Update (Chrome)

NOYB:
This works with both Android and Desktop Chrome (v63) as well as IE 11.
exception: there is apparently an incompatibility between Chrome and JQuery sortable.  So drag/drop sort doesn't work in Chrome.

This is a modification of the previous:
https://github.com/opnsense/core/commit/3f7ee41b5307204fdc4819d09a09aef9e254dbc5#gitext://gotocommit/3f7ee41b5307204fdc4819d09a09aef9e254dbc5

Touch events not working with the widget controls (link, buttons, drag/drop sort).
Add distance, delay and touch-action css for the widget controls to work with touch screen input.

Widget controls not working correctly with touch screen input
https://github.com/opnsense/core/issues/2092

### DIFF
--- a/src/opnsense/www/themes/opnsense/assets/stylesheets/main.scss
+++ b/src/opnsense/www/themes/opnsense/assets/stylesheets/main.scss
@@ -125,6 +125,9 @@ body{
   min-width: 320px;
 }
 
+.widget_sort_handle {
+  touch-action: none; }
+
 .page-head{
   background: map-get($colors, darkgrey);
   top: 0;
--- a/src/opnsense/www/themes/opnsense/build/css/main.css
+++ b/src/opnsense/www/themes/opnsense/build/css/main.css
@@ -5478,6 +5478,9 @@ body {
   touch-action: manipulation;
   min-width: 320px; }
 
+.widget_sort_handle {
+  touch-action: none; }
+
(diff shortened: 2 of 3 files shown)


fichtner:
Thanks, I'll give this a try in a sec! :)

fichtner:
Mobile drag + drop, widget controls works fine. Small issue though: changing the tab in the IPsec widget only partially works:
* ok in macOS Firefox, Safari
* iOS Safari works after drag/drop reposition
* macOS Chrome never works

NOYB:
IPsec widget doesn't like the change from document ready to window load.  The tabs change as expected with document ready.  (IE11, Android and Desktop Chrome 63)

fichtner:
@NOYB ah thanks! @AdSchellevis to the rescue, pretty please :)

AdSchellevis:
@fichtner @NOYB I'm not sure why the window.load doesn't bind the click event (likely another handler detaches it), but the simplest fix is to move back to the document.ready and make sure the event only binds once. The dashboard isn't ideal, but I guess this workaround is solid enough. 

I will commit a change so it will be functional again, @NOYB thanks for testing.

PR #2139, opened 2018-01-27
use label instead of font color

fabianfrz:
do some code optimisation and use label instead of font color. It looks like this now:
screenshot_20180127_125712

### DIFF
--- a/src/www/widgets/widgets/gateways.widget.php
+++ b/src/www/widgets/widgets/gateways.widget.php
@@ -34,45 +34,41 @@
       data.map(function(gateway) {
           var tr_content = [];
           var tr_id = "gateways_widget_gw_" + gateway['name'];
+          var status_color = '';
           if (tbody.find("#"+tr_id).length == 0) {
               // add new gateway
               tr_content.push('<tr id="'+tr_id+'">');
               tr_content.push('<td><small><strong>'+gateway['name']+'</strong><br/>'+gateway['address']+'</small></td>');
               tr_content.push('<td>'+gateway['delay']+'</td>');
               tr_content.push('<td>'+gateway['loss']+'</td>');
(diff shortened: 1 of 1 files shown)


marjohn56:
Nice one... I just mentioned that to Ad, literally seconds before your message came in!  👍

src/www/widgets/widgets/gateways.widget.php
```diff
@@ -34,45 +34,41 @@ function gateways_widget_update(sender, data)
...
             case 'none':
-              status_color = 'green';
-              break;
-            default:
-              status_color = '';
+              status_color = 'success';
               break;
           }
-          $("#"+tr_id+" > td:eq(3) > font").removeAttr('color');
+          console.log($("#"+tr_id+" > td:eq(3) > span"))
```
fichtner: zap

src/www/widgets/widgets/gateways.widget.php
```diff
@@ -34,45 +34,41 @@ function gateways_widget_update(sender, data)
...
+          $("#"+tr_id+" > td:eq(3) > span").removeClass("label-danger label-warning label-success label");
+          console.log(status_color);
```
fichtner: zap! :)

marjohn56:
It might be even better to specify the gateway label colours in the _variables.scss file.

fichtner:
I think this is ok, they should be adjusted to warning/success/danger. Some other widgets made a similar transition.

fabianfrz:
@fichtner did your requested changes - thanks for finding it - I used it for debugging because firefox did not respect my breakpoints for some reasons.

fichtner:
Sure, that's what review are for. Thanks, merged.

PR #2140, opened 2018-01-27
fix: themes in System: Gateways: Status page

fabianfrz:
ref: opnsense/plugins#508

don't use hardcoded colors in status_gateway.php

requested by @marjohn56

### DIFF
--- a/src/www/status_gateways.php
+++ b/src/www/status_gateways.php
@@ -104,7 +104,7 @@ include("head.inc");
                         $gateway_label_class = "success";  // green
                     } else {
                         $online = gettext("Pending");
-                        $gateway_label_class = "info";  // blue (no other color available in bootstrap 3 - its ok but would prefer gray)
+                        $gateway_label_class = "default";  // gray
                     }
 ?>
                       <div class="label label-<?= $gateway_label_class ?>">
--- a/src/www/status_gateways.php
+++ b/src/www/status_gateways.php
@@ -85,29 +85,29 @@ include("head.inc");
                         $status = $gateways_status[$gname];
                         if (stristr($status['status'], "force_down")) {
                             $online = gettext("Offline (forced)");
-                            $bgcolor = "#F08080";  // lightcoral
+                            $gateway_label_class = "danger";  // red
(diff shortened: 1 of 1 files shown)


src/www/status_gateways.php
```diff
@@ -85,29 +85,29 @@
...
+                            $gateway_label_class = "success";  // green
                         }
                     } else if (isset($gateway['monitor_disable'])) {
                         $online = gettext("Online");
-                        $bgcolor = "#90EE90";  // lightgreen
+                        $gateway_label_class = "success";  // green
                     } else {
                         $online = gettext("Pending");
-                        $bgcolor = "#D3D3D3";  // lightgray
+                        $gateway_label_class = "info";  // blue (no other color available in bootstrap 3 - its ok but would prefer gray)
```
fichtner: Default is not grey? https://getbootstrap.com/docs/3.3/components/#labels

fabianfrz: yes that worked - thanks

marjohn56:
That certainly works for online and offline, biut I am unsure as to the effect on the rebellion theme as it's grey anyway...

fichtner:
@marjohn56 this is where you want to redefine default, success, danger, etc to fit the theme perfectly and all the pages behave accordingly :)

src/www/status_gateways.php
fichtner: can we drop all the // colour annotations? they are not necessarily true if modified in the theme

marjohn56:
I thought that.. I thought Green for Danger, Red for OK and blue for default. :)

fichtner:
@marjohn56 haha, your theme, your rules :)

marjohn56:
I'm going to tone down the green a little, the others should not be too far out, although info looks like it could do with a little reduction too.

$brand-success:         #9BD275 !default;
$brand-info:               #B0CDDB !default;
$brand-warning:        #f0ad4e !default;
$brand-danger:          #F05050 !default;

fichtner:
Merged, thanks. I'll remove the superfluous comments inline.

PR #2141, opened 2018-01-27
Widget Sort Handle Tweaks

NOYB:
Delay and distance no longer needed now that the handle does not overlap the controls (title link and buttons).

Be consistent with class naming convention (s/_/-/).


fichtner:
Merged, thank you!

PR #2149, opened 2018-01-30
IPSEC Widget Tabs - Tweak

NOYB:
This is more appropriate, as the tabs are more akin to a label than an output.
Also better for touch screens because some browsers select the tab label text (highlight for copying) when output element tapped.  Not a functionality issue, but looks bad and initially confusing (why is that text selected?)

Using data attributes
https://developer.mozilla.org/en-US/docs/Learn/HTML/Howto/Use_data_attributes

HTML5 is designed with extensibility in mind for data that should be associated with a particular element but need not have any defined meaning. data-* attributes allow us to store extra information on standard, semantic HTML elements without other hacks such as non-standard attributes, extra properties on DOM, or Node.setUserData().


fichtner:
Looks good, thank you.

PR #2152, opened 2018-01-30
Firewall Logs Widget - time min width

NOYB:
Minimum time field width so date (month day, e.g. Jan 24) doesn't wrap and consume additional line unnecessarily.  Particularly in mobile portrait view.

### DIFF
--- a/src/opnsense/www/themes/opnsense/assets/stylesheets/main.scss
+++ b/src/opnsense/www/themes/opnsense/assets/stylesheets/main.scss
@@ -647,3 +647,8 @@ select {
   background-repeat: no-repeat;
   background-position: right;
   background-image: url(/ui/themes/opnsense/build/images/caret.png) !important;}
+
+#grid-log th[data-column-id="__timestamp__"], 
+#filter-log-entries th[data-column-id="__timestamp__"] {
+	min-width: 3.5em;
+}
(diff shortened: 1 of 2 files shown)


src/www/widgets/widgets/log.widget.php
```diff
@@ -230,7 +230,7 @@ function fetch_log(){
   <thead>
     <tr>
       <th data-column-id="action" data-type="icon" class="text-center"><?=gettext("Act");?></th>
-      <th data-column-id="__timestamp__" data-type="time"><?=gettext("Time");?></th>
+      <th data-column-id="__timestamp__" data-type="time" style="min-width: 3.5em"><?=gettext("Time");?></th>
```
fichtner: How do you feel about "white-space: nowrap;" instead?

fichtner: oh, there is also text-nowrap class

fichtner: sorry, silly me, this would then go to the output lines, not the header

NOYB:
Change to use the table header column id attribute value as selector for css.

fichtner:
@NOYB why do you prefer a min width over preventing text wrap? honest question, I'm not attached to a particular solution

NOYB:
The content of the field is short month, day, hours and minutes. e.g. Jan 24 22:33
Preventing text wrap means it on goes on one line.  It's too wide for portrait mobile view.
Min width keeps the month and day on single line and allows time to wrap to second line.
Without min width it wraps to 3 lines instead of only 2.

There are other ways, such as replacing the regular space between month and day with a non breaking space.  But altering the content as means of format control seem like more of a hack.

fichtner:
Alright, thanks for explaining. Merged. :)

PR #2156, opened 2018-02-02
Touch Friendlyize - Interface Overview Interface Selection

NOYB:
The chevron down icon is problematic to touch select here.  Extend selection area to entire vertical and horizontal row space.


fichtner:
Makes sense, thanks! :)

PR #2157, opened 2018-02-02
HTML Compliance - Interfaces Overview

NOYB:
Create opportunistic word breaks between raw and unit-ized (KB/MB/GB) In/Out Packets for better portrait mobile view line wrap formating.

These are not necessarily ISP DNS servers.

Error: script element between head and body.


fichtner:
Merged, thanks!

PR #2159, opened 2018-02-02
add CRUD action helpers to ApiMutableModelControllerBase

fabianfrz:
some DRY for Plugin Controllers this makes CRUD actions one liners (tested with tor acl).


AdSchellevis:
@fabianfrz Thanks, we have been working on some shared code for this a long time ago, but I think back then we tried to incorporate too much functionality at once. We should add some phpdoc lines to keep the styling in order, but I can handle that.

PR #2167, opened 2018-02-04
Interfaces Overview - Relinquish DHCP Lease

NOYB:
Send gratuitous DHCP release message to server.

Dependency:
To fully relinquish the DHCP lease, the "RELEASE" reason is added to /usr/local/sbin/dhclient-script.ext.
s/EXPIRE|FAIL)/EXPIRE|FAIL|RELEASE)/

Without the 'RELEASE' reason, the DHCP release message will still be sent to the server, however the local dhclient leases will not reflect the relinquishment.  Therefore, DHCP renewal will do a DHCP request for the previous address rather than doing a DHCP discovery and being more likely to be assigned a different address.


NOYB:
Added tooltip.

NOYB:
Update: Break post item naming correction and styling change into separate commits for better clarity of the relinquish DHCP lease feature commit.

NOYB:
Update: Include all the interface types in the CSS layout and post item naming changes.

NOYB:
Update: Quote the input values for "if" and "ipv".

PR #2169, opened 2018-02-05
Services - Unbound DNS - General - System Domain Local Zone Type Option

NOYB:
Expose system domain unbound local-zone type to be configurable.

### DIFF
--- a/src/etc/inc/plugins.inc.d/unbound.inc
+++ b/src/etc/inc/plugins.inc.d/unbound.inc
@@ -516,7 +516,14 @@ function unbound_add_host_entries()
 {
     global $config;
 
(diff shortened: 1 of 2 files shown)


fichtner:
We should retain compatibility and use "transparent" as the empty default. I'll do this post-merge. Thanks!

fichtner:
Please check 0bd1d466f064a1d7ba85b91831d0ae9e91e957e5

NOYB:
It should already be using "transparent" as the default if the setting doesn't match a valid type.  That should include empty.  (unbound.inc)

fichtner:
It does, but it previously forced "transparent" into the config.xml causing both "" and "transparent" to mean the same thing.

PR #2175, opened 2018-02-07
Help For - Data For Attribute

NOYB:
Help For - Data For Attribute

Use original element (&lt;div&gt;/&lt;small&gt;) with "data-for" attribute.

The old jquery "for" attribute selectors can be removed once conversion to "data-for" attribute is complete (including plugins).

The old jquery selectors to remove once conversion is complete.
  opnsense_ui.js & head.inc
  $("*[for='" + $(this).attr('id') + "']")....
  $('[for*="help_for"]')....

If you'd like to go this route (with or without the css padding) let me know and I'll make a plugins commit for the same.  If you're wanting to do something else that's fine too.  Just close this PR.  I probably should have gone the data-for attribute route to being with rather than changing to the output element.  A good little learning exercise.


fichtner:
Merged, thanks!

PR #2180, opened 2018-02-07
inline nmea checksum calculator; use bootstrap style for button

fabianfrz:
This was probably not planned that bootstrap breaks the line before and after the input field. Also the button was missing bootstrap style.

screenshot_20180207_190737


fichtner:
Looks like we could get get rid of the leading "$" and trailing "*checksum" as it says so in the label?

fabianfrz:
@fichtner the "checksum" string will be replaced by the checksum after the button is pushed so it cannot be removed.

fabianfrz:
now looks like this:
screenshot_20180208_173140

and the advanced button has been formatted too:
screenshot_20180208_173222

fichtner:
Thanks, I'll take a look on the offset of text and button in the morning.

PR #2181, opened 2018-02-07
changing reference for fonts/

teliz:
when using another theme the font remains the same as the @ font-face is with fixed path.
:)


AdSchellevis:
The font path is compiled using https://sass-lang.com/ your theme should contain the correct path, change https://github.com/opnsense/core/blob/master/src/opnsense/www/themes/opnsense/assets/stylesheets/main.scss and recompile for your theme and you should be good.

PR #2184, opened 2018-02-09
Buttons for services control.

kekek2:


src/opnsense/www/js/opnsense_ui.js
```diff
@@ -153,6 +153,43 @@ function updateServiceStatusUI(status) {
     $('#service_status_container').html(status_html);
 }
 
+function updateServiceControlUI(ServiceName)
```
fabianfrz: variables should not start with a uppercase character -> would indicate a class name

src/opnsense/www/js/opnsense_ui.js
```diff
@@ -153,6 +153,43 @@ function updateServiceStatusUI(status) {
...
+                            $("#startService").removeClass("btn-danger").removeClass("btn-success");
+                            if (data['status'] == "running") {
+                                $("#startService").addClass("btn-success");
+                            } else if (data['status'] == "stopped") {
+                                $("#startService").addClass("btn-danger");
+                            }
+                        });
+                    });
+                });
+            })((commands[i]));
```
fabianfrz: too many brackets

src/opnsense/www/js/opnsense_ui.js
```diff
@@ -153,6 +153,43 @@ function updateServiceStatusUI(status) {
...
+        }
+
+        var buttons = '<span id="startService" class="glyphicon glyphicon-play btn ' + status_html + '"></span>';
+        buttons += '<span id="restartService" class="glyphicon glyphicon-refresh btn"></span>';
+        buttons += '<span id="stopService" class="glyphicon glyphicon-stop btn"></span>';
+        $('#service_status_container').html(buttons);
+
+        var commands = ["start", "restart", "stop"];
+        for (var i = 0; i < commands.length; i++) {
+            (function (command) {
```
fabianfrz: is that function really needed?

kekek2: Variable Commands[i] used in call-back function. The value "i" will always be 3.

fabianfrz: This does not work?
```js
commands.forEach(function(command) {
// do something
});
```

AdSchellevis:
Wrapping service control in updateServiceControlUI(), although I’m not sure it’s a good idea to automatically depend on the processing modal dialog. Could be optional or the modal could have an other name

fichtner:
I'm merging this, but it needs an extensive rework for a consistent look and feel with the legacy pages...

fichtner:
Thanks!

PR #2200, opened 2018-02-14
RADVD Problem on PPPeO link without public IPv6 address

marjohn56:
Forum message
https://forum.opnsense.org/index.php?topic=7304.0

No one else has complained of this which is odd, @fichtner  you have pppoe with a link local I  thought.

### DIFF
--- a/src/etc/inc/services.inc
+++ b/src/etc/inc/services.inc
@@ -174,8 +174,10 @@ function services_radvd_configure($blacklist = array())
 
         $stanzas = array();
         $ifcfgsnv6 = get_interface_subnetv6($dhcpv6if);
-        $subnetv6 = gen_subnetv6($ifcfgipv6, $ifcfgsnv6);
-        $stanzas[] = "{$subnetv6}/{$ifcfgsnv6}";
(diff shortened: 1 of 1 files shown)


src/etc/inc/services.inc
```diff
@@ -174,8 +174,10 @@ function services_radvd_configure($blacklist = array())
 
         $stanzas = array();
         $ifcfgsnv6 = get_interface_subnetv6($dhcpv6if);
-        $subnetv6 = gen_subnetv6($ifcfgipv6, $ifcfgsnv6);
-        $stanzas[] = "{$subnetv6}/{$ifcfgsnv6}";
+        if ($ifcfgsnv6) {
```
fichtner: can we make this !empty() ... should be the same?

fichtner:
I don't have a pppoe but this looks reasonable.

marjohn56:
Whatever way floats your boat... :)

src/etc/inc/services.inc
marjohn56: Whatever way floats your boat... :)

marjohn56:
Done

PR #2203, opened 2018-02-16
Fix Activity.

kekek2:
System/Activity don't show processes,

### DIFF
--- a/src/opnsense/scripts/systemhealth/activity.py
+++ b/src/opnsense/scripts/systemhealth/activity.py
@@ -44,7 +44,7 @@ if __name__ == '__main__':
         output_stream.seek(0)
         rownum = 0
         for line in output_stream.read().strip().split('\n'):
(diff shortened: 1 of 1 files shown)


fichtner:
@kekek2 can you explain "don't show processes" ?

kekek2:
screenshot

AdSchellevis:
@kekek2 which version is this? On my end it looks normal.

kekek2:
17.7.12

AdSchellevis:
That does the following display on a terminal at your end?

`top -aHSn 99999 `

kekek2:
top

AdSchellevis:
@kekek2 your using zfs?

kekek2:
ZFS in the testing state.

AdSchellevis:
that's where the extra line came from, we should probably search for the header rule differently to make sure it always works.

PR #2204, opened 2018-02-16
Add support for more DNS Filter

unicorn667:
Hi,

i don´t want to use opendns as dns filter. 
It´s possible to interegrate something else that is free like https://github.com/pi-hole/pi-hole 

Best regards


AdSchellevis:
@unicorn667 please don't use PR's to create feature requests.

Issue #2206, opened 2018-02-17
Can't switch WAN from PPPoE to DHCP directly


coaxial:
Versions: 

```
OPNsense 18.1.2_2-amd64
FreeBSD 11.1-RELEASE-p6
OpenSSL 1.0.2n 7 Dec 2017
```

If I set my WAN interface to PPPoE, save it, and then want to change it to DHCP, it fails with the cryptic error message `You have to reassign the interface to be able to configure as dhcp.`

If instead of DHCP, I set it to static with any valid IP, save, and then switch to DHCP, no problem.

I believe this is a bug because the static intermediary step shouldn't be necessary and isn't documented. The error message when switching from PPPoE to DHCP is not very clear either and doesn't help understand how to solve the issue.

PR #2207, opened 2018-02-17
GUI Issue when changing WAN PPPoE to DHCP

marjohn56:
Issue #2206 Can't switch WAN from PPPoE to DHCP directly

When changing from PPPoE to DHCP on the WAN interface, re-assigning the interface will still not allow you to select dhcp, static is OK.

This simply forces it to DHCP and allows the user to continue.


AdSchellevis:
We better drop the validation, the end result looks the same when changing to dhcp on assignment.

PR #2209, opened 2018-02-18
Accessibility and Touch Friendlyize - Checkboxes / Radio Buttons

NOYB:
Add label element to checkboxes and radio buttons to make them touch friendly and improve accessibility i.e. screen readers, dexterity, etc.

Many of the checkboxes are absent a label.  No big deal for visual usage.  But not sure how well that works out for screen readers.  Also is less than idle accessibility for the dexterity impaired.  For these checkboxes the label element has been added but the  label content needs to be added to them.  I realize it may be a design decision to not put a label on some of them when the left column pretty much explains it.  But again, not sure how well that works for screen readers.  From what I've read it seems to be best practice to use a label on input elements for best accessibility experience.

Also as best practice according to Mozilla, set the for attribute on label nested elements.  "...some assistive technologies do not understand implicit relationships between labels and widgets."  https://developer.mozilla.org/en-US/docs/Learn/HTML/Forms/How_to_structure_an_HTML_form

Added an mvc form element to accommodate labeling form elements.  Just include it in the form xml.  e.g. <input_label>Enable this feature</input_element>

Adding label content to these could be a good way for people to easily make contributions.


src/www/services_dhcpv6.php
```diff
@@ -430,8 +430,10 @@ function show_netboot_config() {
                     <tr>
                       <td><i class="fa fa-info-circle text-muted"></i>  <?=gettext("Enable");?></td>
                       <td>
-                        <input name="enable" type="checkbox" value="yes" <?php if ($pconfig['enable']) echo "checked=\"checked\""; ?> />
-                        <strong><?= sprintf(gettext("Enable DHCPv6 server on " . "%s " ."interface"),!empty($config['interfaces'][$if]['descr']) ? htmlspecialchars($config['interfaces'][$if]['descr']) : strtoupper($if));?></strong>
+                        <label for="enable">
+                          <input name="enable" id="enable" type="checkbox" value="yes" <?php if ($pconfig['enable']) echo "checked=\"checked\""; ?> />
+                          <strong><?= sprintf(gettext("Enable DHCPv6 server on " . "%s " ."interface"),!empty($config['interfaces'][$if]['descr']) ? htmlspecialchars($config['interfaces'][$if]['descr']) : strtoupper($if));?></strong>
```
fabianfrz: please fix the string

NOYB: If you wish to change the string you can create a commit.  Merge conflicts will then be resolved in this commit.

fabianfrz: it is just about a removal of this useless string concatenation - the gettext string does not need that.

NOYB: Submit a commit to change it.  That is in the master.  Not something introduced here.

PR #2212, opened 2018-02-18
Harden GUI by adding a CSP, XSS protection and referrer hiding

fabianfrz:
this should stay on master for longer time since it may have side effects like blocking resources.

### DIFF
--- a/src/opnsense/mvc/app/controllers/OPNsense/Base/ControllerBase.php
+++ b/src/opnsense/mvc/app/controllers/OPNsense/Base/ControllerBase.php
@@ -204,5 +204,10 @@ class ControllerBase extends ControllerRoot
 
         // append ACL object to view
         $this->view->acl = new \OPNsense\Core\ACL();
+        $this->response->setHeader('Content-Security-Policy', "default-src 'self'; script-src 'self' 'unsafe-inline' 'unsafe-eval'; style-src 'self' 'unsafe-inline' 'unsafe-eval';");
+        $this->response->setHeader('X-Frame-Options', "SAMEORIGIN");
+        $this->response->setHeader('X-Content-Type-Options', "nosniff");
+        $this->response->setHeader('X-XSS-Protection', "1; mode=block");
+        $this->response->setHeader('Referrer-Policy', "same-origin");
     }
 }
--- a/src/opnsense/mvc/app/controllers/OPNsense/Base/IndexController.php
+++ b/src/opnsense/mvc/app/controllers/OPNsense/Base/IndexController.php
@@ -64,16 +64,4 @@ class IndexController extends ControllerBase
         }
         return false;
     }
-
-    /**
-     * before routing event
-     * @param Dispatcher $dispatcher
-     * @return null|bool
-     */
-    public function beforeExecuteRoute($dispatcher)
-    {
-        parent::beforeExecuteRoute($dispatcher);
-        // by default, don't allow our content to be rendered in an [i]frame from another location.
-        $this->response->setHeader('X-Frame-Options', 'SAMEORIGIN');
-    }
 }
--- a/src/www/head.inc
+++ b/src/www/head.inc
@@ -28,12 +28,6 @@ $themename = html_safe(get_current_theme());
 $pagetitle = gentitle(array_reverse($menuBreadcrumbs), ' | ');
 $pagetitle .= html_safe(sprintf(' | %s.%s', $config['system']['hostname'], $config['system']['domain']));
 
-// hardening
-header("Content-Security-Policy: default-src 'self'; script-src 'self' 'unsafe-inline' 'unsafe-eval'; style-src 'self' 'unsafe-inline' 'unsafe-eval';");
-header('X-Frame-Options: SAMEORIGIN');
-header('X-Content-Type-Options: nosniff');
-header('X-XSS-Protection: 1; mode=block');
-header('Referrer-Policy: same-origin');
 
 ?><!doctype html>
 <!--[if IE 8 ]><html lang="<?=system_get_language_code();?>" class="ie ie8 lte9 lte8 no-js"><![endif]-->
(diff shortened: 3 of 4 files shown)


src/opnsense/mvc/app/controllers/OPNsense/Base/ControllerBase.php
```diff
@@ -204,5 +204,10 @@ public function beforeExecuteRoute($dispatcher)
 
         // append ACL object to view
         $this->view->acl = new \OPNsense\Core\ACL();
+        $this->response->setHeader('Content-Security-Policy', "default-src 'self'; script-src 'self' 'unsafe-inline' 'unsafe-eval'; style-src 'self' 'unsafe-inline' 'unsafe-eval';");
```
fabianfrz: will only allow resources directly from the firewall and additional embedded in the HTML, style attributes and JavaScript eval.

src/opnsense/mvc/app/controllers/OPNsense/Base/ControllerBase.php
```diff
@@ -204,5 +204,10 @@ public function beforeExecuteRoute($dispatcher)
...
+        $this->response->setHeader('X-Frame-Options', "SAMEORIGIN");
```
fabianfrz: Allow framing only in the context of the firewall itself (no framing by another website) - avoid clickjacking

src/opnsense/mvc/app/controllers/OPNsense/Base/ControllerBase.php
```diff
@@ -204,5 +204,10 @@ public function beforeExecuteRoute($dispatcher)
...
+        $this->response->setHeader('X-Content-Type-Options', "nosniff");
+        $this->response->setHeader('X-XSS-Protection', "1; mode=block");
```
fabianfrz: block entire page if the browser detects XSS

src/opnsense/mvc/app/controllers/OPNsense/Base/ControllerBase.php
```diff
@@ -204,5 +204,10 @@ public function beforeExecuteRoute($dispatcher)
...
+        $this->response->setHeader('Referrer-Policy', "same-origin");
```
fabianfrz: do not show the referrer to other pages if the user clicks a link in the GUI.

L1ghtn1ng:
You can test the csp using report-uri.com as that will let you know of any issues
________________________________

fabianfrz:
looks like it works on static pages as well.

src/opnsense/mvc/app/controllers/OPNsense/Base/ControllerBase.php
AdSchellevis: @fabianfrz this one is already in there https://github.com/opnsense/core/blob/18.1.2/src/opnsense/mvc/app/controllers/OPNsense/Base/IndexController.php#L77 , we could move it to Base and make sure the classes deriving from index move to base. What do you think?

src/www/head.inc
```diff
@@ -28,6 +28,13 @@ $themename = html_safe(get_current_theme());
 $pagetitle = gentitle(array_reverse($menuBreadcrumbs), ' | ');
 $pagetitle .= html_safe(sprintf(' | %s.%s', $config['system']['hostname'], $config['system']['domain']));
 
+// hardening
+header("Content-Security-Policy: default-src 'self'; script-src 'self' 'unsafe-inline' 'unsafe-eval'; style-src 'self' 'unsafe-inline' 'unsafe-eval';");
+header('X-Frame-Options: SAMEORIGIN');
+header('X-Content-Type-Options: nosniff');
+header('X-XSS-Protection: 1; mode=block');
+header('Referrer-Policy: same-origin');
+
```
AdSchellevis: probably better to move the headers here https://github.com/opnsense/core/blob/18.1.2/src/www/guiconfig.inc#L38 as the csrf checks are also there.

src/opnsense/mvc/app/controllers/OPNsense/Base/ControllerBase.php
fabianfrz: So it should be removed from the index controller since it inherits controllerbase and security should be enforced on all controllers - even those which are not IndexControllers.

src/opnsense/mvc/app/controllers/OPNsense/Base/IndexController.php
```diff
@@ -64,16 +64,4 @@ public function handleErrorAction($message = null, $sender = null)
...
     }
-
-    /**
-     * before routing event
-     * @param Dispatcher $dispatcher
-     * @return null|bool
-     */
-    public function beforeExecuteRoute($dispatcher)
-    {
-        parent::beforeExecuteRoute($dispatcher);
```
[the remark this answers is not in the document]
AdSchellevis: @fichtner should be safe to delete, falls back to ControllerBase->beforeExecuteRoute()

fabianfrz:
ping

AdSchellevis:
pong, sorry a bit busy. will pull it in now.

AdSchellevis:
@fabianfrz I'm going to drop the "default-src 'self';" in the security policy for the mvc code, this prevents plugins from using anything like open streetmap, which causes issues on some of our plugins.  Not sure if this is the only issue with the more tightened security yet though. We might look into something at a later point to add trusted domains in the controller area, although I'm not sure what the best option is there yet.

fabianfrz:
Maybe inject or replace the values with properties like the values for the mutable controller. So secure defaults are kept.

BTW: Which plugin is using OSM? It should only require image-src to add the domain.
Openlayers or leaflet should be placed in the plugin.

AdSchellevis:
some custom work. adding the desired domains would indeed be better, but this needs a bit of work for a clear hook.

fabianfrz:
I dont think so. Adding a property $csp-image-source to the controller base should work and the defaults are the values of my pull request. Any controller that inherits the base controller, can override the default just by adding it in the class. In that case it would be controller wide which should not be a huge issue because most controllers do have 1-3 actions

AdSchellevis:
@fabianfrz this https://github.com/opnsense/core/commit/a8f54d2e7771dc8556544c961535ccc1e76974dc should do the trick

fabianfrz:
Exactly the implementation I had in mind

PR #2222, opened 2018-02-23
Enable charon.install_routes from GUI

namezero111111:
When routing the default route over IPSec, installing 0.0.0.0/1 and 128.0.0.0/1 routes effectively overrides the default gateway, making it impossible for the firewall itself to use the route, because this hack (https://doc.pfsense.org/index.php/Why_can%27t_I_query_SNMP,_use_syslog,_NTP,_or_other_services_initiated_by_the_firewall_itself_over_IPsec_VPN) no longer works.

This is alleviated by adding the option to not automatically install routes via the charon.install_routes directive (https://wiki.strongswan.org/projects/strongswan/wiki/Strongswanconf).

The rationale behind providing this via GUI is that "Passthrough Networks" also is provided for that exact reason; providing the new switch rounds off this experience.

### DIFF
--- a/src/etc/inc/plugins.inc.d/ipsec.inc
+++ b/src/etc/inc/plugins.inc.d/ipsec.inc
@@ -880,6 +880,9 @@ function ipsec_configure_do($verbose = false, $interface = '')
 
         $cnf_add_to_charon_section = "";
         $cnf_add_to_charon_section .= $aggressive_psk ? "\ti_dont_care_about_security_and_use_aggressive_mode_psk=yes\n":"";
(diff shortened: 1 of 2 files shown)


AdSchellevis:
@namezero111111 looks good, practical to have. I spotted a small indentation issue, but we'll fix that after merge. Thanks!

namezero111111:
Thanks for the quick merge; sorry about the indentation; I was copy&pasting with Notepad++ :}

AdSchellevis:
no problem, was easy to fix

PR #2232, opened 2018-03-03
Check IP Services - Plugin Support

NOYB:
Update get_dyndns_ip function to support adding other services and disabling the factory default service (Check IP Services - Plugin).
  Factory default check IP service as array
  Support verify peer
  Support authentication

### DIFF
--- a/src/etc/inc/services.inc
+++ b/src/etc/inc/services.inc
@@ -1570,6 +1570,21 @@ function services_dhcrelay6_configure($verbose = false)
     }
 }
 
+/* Factory default check IP service. */
+function factory_default_checkipservice()
+{
+    return array(
+        "enable" => true,
+        "name" => 'Default',
+        "url" => 'http://checkip.dyndns.org',
+#       "url" => 'http://ip.3322.net/',    /* Chinese alternative */
+#       "username" => '',
+#       "password" => '',
+#       "verifysslpeer" => true,
+        "descr" => 'Factory Default Check IP Service'
+    );
+}
+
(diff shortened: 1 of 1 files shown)


src/etc/inc/services.inc
```diff
@@ -1570,6 +1570,21 @@ function services_dhcrelay6_configure($verbose = false)
     }
 }
 
+/* Factory default check IP service. */
+function factory_default_checkipservice()
```
fichtner: currently missing in the other PR, creating a fatal error. We should avoid calling this twice. The MVC model gives nice migration for such cases, maybe we should create a model and use it from the static pages? I can help with that.

NOYB: I'm not following what you mean that it is missing in the other PR.  The other PR includes services.inc and calls the function to get the factory default check ip service.
Been running this for quite some time and not throwing any error.

NOYB: Ah, think I understand what you mean.  Think you are getting error because this PR has not been merged yet and therefor the function does not exist.

NOYB: Not sure how you want to proceed here.  As constructed the factory_default_checkipservice() function is obtained in services_checkip.php by require_once inclusion of services.inc.
So once this commit is merged it will work.  You mentioned using an MVC model.  I'm not sure how that works for this or how to create that.  The function needs to be available to both the services_checkip.php and the services.inc: get_dyndns_ip() function.

fichtner:
Merged, thanks!

PR #2233, opened 2018-03-03
HTML Compliance - Reporting

NOYB:
System Health
  Error: Element style not allowed as child of element section in this context.
  Error: Duplicate ID loading.
  Error: Row 1 of a row group established by a thead element has no cells beginning on it.
  Warning:  The type attribute for the style element is not needed and should be omitted.

Insight
  Error: Element style not allowed as child of element section in this context.
  Warning: A table row was 6 columns wide and exceeded the column count established by the first row (5).
  Warning: A table row was 4 columns wide, which is less than the column count established by the first row (6).
  Warning:  The type attribute for the style element is not needed and should be omitted.

Traffic Graph
  Error: Element style not allowed as child of element div in this context.
  Error: Table column 3 established by element td has no cells beginning in it.
  Error: Table column 5 established by element td has no cells beginning in it.

### DIFF
--- a/src/opnsense/mvc/app/views/OPNsense/Diagnostics/networkinsight.volt
+++ b/src/opnsense/mvc/app/views/OPNsense/Diagnostics/networkinsight.volt
@@ -26,12 +26,18 @@ POSSIBILITY OF SUCH DAMAGE.
 
 #}
 
-<style type="text/css">
-.panel-heading-sm{
-    height: 28px;
-    padding: 4px 10px;
-}
-</style>
+<script>
+    var sheet = document.createElement('style');
+
+    sheet.innerHTML = '\
+.panel-heading-sm {\
+    height: 28px;\
+    padding: 4px 10px;\
+}\
+';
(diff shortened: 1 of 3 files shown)


AdSchellevis:
@NOYB what are you using for validation? HTML Validator doesn't come up with the same issues on my end (at least for the insight page) and I'm not sure it's a good idea to replace the style for a script replacement here, so I would like to try the same on my end and think this over a bit before pulling in.

NOYB:
I'm using https://validator.w3.org/#validate_by_input

AdSchellevis:
got it, my plugin was mixing stuff up apparently.

AdSchellevis:
on my end it looks like omitting  `type="text/css"` in the style element is enough to loose the warning, I guess we could preserve the style tag and fix the other things like you suggested.

NOYB:
Think w3 validator will still complain about the style element as child of section.  But I'll take what I can get.

PR #2234, opened 2018-03-03
dhcp6c & rtsold conf  & script creation breakout

marjohn56:
This commit is a basic breakout of the conf and script file creation only. It does not attempt to fix the underlying issues with duplicate dhcp6c clients, existing rtsold and dhcp6c launch and handling remains unchanged. 

It does fix an issue with advanced dhcp6c config and it also allows for changes in LAN config that may affect dhcp6c config.

This is the first part of a replacement for #2090

### DIFF
--- a/src/etc/inc/interfaces.inc
+++ b/src/etc/inc/interfaces.inc
@@ -2637,34 +2637,54 @@ function interface_configure($interface = 'wan', $reloadall = false, $linkupeven
 	/* with dhcp6.            							     */
 	/* We call the function configure_dhcp6c_conf() when     */
 	/* either its a WAN dhcpv6 or a LAN track6 setting.      */
+	
+	/* We need to know the WAN interface that's using dhcp6c */
+	/* In this itaration, the first one, we only support     */
+	/* one for now.                                          */
+    
+    $iflist = get_configured_interface_list();
+	foreach ($iflist as $if => $ifname) {
+        if (!empty($config['interfaces'][$if]['ipaddrv6']) && $config['interfaces'][$if]['ipaddrv6'] == 'dhcp6') {
+            $dhcp6if = $ifname;
+            $dhcp6cfg = $config['interfaces'][$dhcp6if];
+            break;
+        } 
+    }
 
     if (isset($wancfg['ipaddrv6'])) {
         switch ($wancfg['ipaddrv6']) {
             case 'slaac':
             case 'dhcp6':
-				configure_dhcp6c_conf('wan', $config['interfaces']['wan']);
+                configure_dhcp6c_conf($dhcp6if, $dhcp6cfg);
                 interface_dhcpv6_configure($interface, $wancfg);
                 break;
             case '6rd':
+                configure_dhcp6c_conf($dhcp6if, $dhcp6cfg);
                 interface_6rd_configure($interface, $wancfg);
                 break;
             case '6to4':
+                configure_dhcp6c_conf($dhcp6if, $dhcp6cfg);
                 interface_6to4_configure($interface, $wancfg);
                 break;
             case 'track6':
-				configure_dhcp6c_conf('wan', $config['interfaces']['wan']);
+				configure_dhcp6c_conf($dhcp6if, $dhcp6cfg);
                 interface_track6_configure($interface, $wancfg, $reloadall || $linkupevent);
                 break;
             default:
                 /* XXX: Kludge for now related to #3280 */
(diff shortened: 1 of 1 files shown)


fichtner:
Thanks for the breakdown! Modified the patch a little for clarity to start with: https://github.com/opnsense/core/commit/37f7e72799f

One comment inline follows.

src/etc/inc/interfaces.inc
```diff
@@ -2628,10 +2628,21 @@ function interface_configure($interface = 'wan', $reloadall = false, $linkupeven
...
+	/* Initialy this will only support one WAN interface     */
+	/* with dhcp6.            							     */
+	/* We call the function configure_dhcp6c_conf() when     */
+	/* either its a WAN dhcpv6 or a LAN track6 setting.      */
+
     if (isset($wancfg['ipaddrv6'])) {
         switch ($wancfg['ipaddrv6']) {
             case 'slaac':
             case 'dhcp6':
+				configure_dhcp6c_conf('wan', $config['interfaces']['wan']);
```
fichtner: Shouldn't this be $interface and $wancfg ?

src/etc/inc/interfaces.inc
```diff
@@ -2641,6 +2652,7 @@ function interface_configure($interface = 'wan', $reloadall = false, $linkupeven
                 interface_6to4_configure($interface, $wancfg);
                 break;
             case 'track6':
+				configure_dhcp6c_conf('wan', $config['interfaces']['wan']);
```
fichtner: And here we need to look up the track interface instead of guessing WAN, right?

src/etc/inc/interfaces.inc
marjohn56: If you use $interface then I would  think it would try to set up dhcp6c conf using the LAN interface settings, as interface_configure is also called when the LAN interface is being configured, hence my forcing it to use WAN... or am I reading this wrong, It works as expected and re-writes the config on a change to LAN or WAN.

fichtner: Don‘t know yet. Statically assuming WAN is likely wrong. First review sweep for just asking questions. But this patch is manageable now, thanks again. We will figure this out together. 😊

src/etc/inc/interfaces.inc
marjohn56: I guess we could, but at this point I am very wary of a user trying to use two WAN dhcp6 interfaces, that opens up a whole new can of worms I am yet to look at. My reading of dhcp6c says it's possible, and one of the reasons for this approach. I just have gone there .... yet. :)

fichtner: the problem is: people may add a second wan (e.g. opt1) and delete wan... so we come up empty here

marjohn56:
OK, so I can change that to pull the first interface that uses dhcp6, would that do the trick?

src/etc/inc/interfaces.inc
```diff
@@ -3341,26 +3379,32 @@ function DHCP6_Config_File_Advanced($interface, $wancfg, $wanif)
             }
             $id_assoc_statement_prefix .= ";";
         }
-
-        if (is_numeric($wancfg['adv_dhcp6_prefix_interface_statement_sla_id'])) {
```
fichtner: I believe this block is for the dhcp6 interface itself. if you want to iterate for the track6 interfaces you have to do it in the same way as in basic as an additional block.

fichtner: (please no work here, I am modifying the diff on the other branch so all thoughts are helpful)

src/etc/inc/interfaces.inc
fichtner: should be the same interface here, that's where it was split from :)

src/etc/inc/interfaces.inc
fichtner: here we'll have to rewrite the interface that is being tracked

marjohn56:
I'll hold off until you tell me.

src/etc/inc/interfaces.inc
marjohn56: Not only that, I have also realised that if its changed away from track-interface, then the config needs to reflect that. I am currently working on that, but I'll hold until you have finished whatever else you are doing.

marjohn56:
OK, have a look at 0639d0b, I think it does what we need. I have not squashed the commits but I have edited in your baby steps where they fell within my original commit.

fichtner:
Continued in #2258

PR #2235, opened 2018-03-03
Check IP Services - Plugin

NOYB:
Configurable check IP services.
Used by Dynamic DNS services to check IP addresses.
Configure as many check IP services as desired.  The first (highest in the list) that is enabled will be used as the default.
Include verify peer option.
A future possibility could be to add an option to the Dynamic DNS entires to use a specific check IP service.
Allows additional/private check IP services to be added and set as the default.

### DIFF
--- a/src/www/services_checkip.php
+++ b/src/www/services_checkip.php
@@ -1,0 +1,192 @@
+<?php
+
+/*
+    Copyright (C) 2014-2016 Deciso B.V.
+    Copyright (C) 2008 Ermal Luçi
+    All rights reserved.
+
+    Redistribution and use in source and binary forms, with or without
+    modification, are permitted provided that the following conditions are met:
+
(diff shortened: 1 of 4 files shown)


src/www/services_checkip.php
```diff
@@ -0,0 +1,192 @@
+<?php
+
+/*
```
fichtner: No Copyrights for you?

NOYB: Only thing I'm interested in is retaining my right to use personally or in other projects.
As far as anyone else using it.  Far as I'm concerned my portions of the code is public domain.
If necessary for that I guess you could add a NOYB at Github copyright.

fichtner:
Ok, fine, since this is BSD license your right to use is unrestricted so no need to do anything.

fichtner:
Merged, thanks!

PR #2251, opened 2018-03-09
Add Nextcloud backup

fabianfrz:
note: this is a work in progress until the "WIP:" prefix is removed.


AdSchellevis:
Hi Fabian,

Maybe we can make the backup options pluggable and move the google drive into the same as well, some of the patterns can easily be reused.

Ideally the flow should be more or less the following:

- gather files to upload (at the moment we only pack config.xml, but we might want to extend that at some point)
- encrypt content (if selected)
- upload using selected method, which I should put in src/opnsense/scripts/remote_backup.php at the moment, to automatically fit the cron schedule.

If we move the backup options into classes, we could also provide some simple interfaces (a bit like the authentication framework), like:

- listParameters() {returns required parameters for this service, which we can also use in the ui}
- test(in:config parameters) {perform test, either directly or using configd if root access is needed to test the service}
- backup() {perform actual backup}

What do you think? 

Best,

Ad

fabianfrz:
@AdSchellevis I like the idea here is some sample to discuss:

```php
<?php
$uploaders = array();
class BackupUploader {
    public abstract function performUpload($file);
    public abstract function getRequiredFields();
    protected function encryptAndCompressFile($file);
}

class SCPUploader extends BackupUploader {
// code
}
$uploaders[] = 'SCPUploader';

class NextcloudUploader  extends BackupUploader  {
// code
}
$uploaders[] = 'NextcloudUploader';

class Backup {
  public function initialize($uploader)
  private function collectFiles();
  public function uploadFiles();
}
```

mimugmail:
Really nice addition :)

AdSchellevis:
@fabianfrz I'm a bit late to the party, but in concept I agree, we should only split them up into different files so we can auto probe the options using a factory class on top. If you like, I can prepare the base classes for you to work with.

fabianfrz:
@AdSchellevis The next weeks I will probably not be able to handle this in the next weeks.

AdSchellevis:
@fabianfrz no rush, I can add some classes in the time between so you can pick up where you left as soon as you have time to spare.

fabianfrz:
@AdSchellevis no problem - when the API exists, adjusting this PR should be simple.

AdSchellevis:
@fabianfrz the basics are in, google drive backup is using our new framework now.

fabianfrz:
@AdSchellevis @fichtner  will create a new PR since this one is obsolete now.

fabianfrz:
rewritten in PR #2289

PR #2258, opened 2018-03-12
Interfaces: dhcp6-creation-breakout-step2

marjohn56:
Additions to Franco's changes. Moved script creation to interface_dhcpv6_prepare().

Removed call to interface_dhcpv6_prepare() from interface_dhcpv6_configure(), the conf and script creation needs to be called before we get there. Checkout L2644 onwards.

### DIFF
--- a/src/etc/inc/interfaces.inc
+++ b/src/etc/inc/interfaces.inc
@@ -2627,31 +2627,63 @@ function interface_configure($interface = 'wan', $reloadall = false, $linkupeven
             break;
     }
 
+    /* configure all ipv6 dhcp conf & scripts at this point  */
+    /* We need to take into account that a LAN change        */
+    /* may also affect it. It will get called everytime      */
+    /* interface is reconfigured for track or WAN dhcpv6,    */
+	/* but we have to watch all interfaces not just WAN.     */
+	/* Initialy this will only support one WAN interface     */
+	/* with dhcp6.            							     */
+	/* We call the function configure_dhcp6c_conf() when     */
+	/* either its a WAN dhcpv6 or a LAN track6 setting.      */
+	
+	/* We need to know the WAN interface that's using dhcp6c */
+	/* In this itaration, the first one, we only support     */
+	/* one for now.                                          */
+    
+    $iflist = get_configured_interface_list();
+	foreach ($iflist as $if => $ifname) {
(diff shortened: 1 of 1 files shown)


fichtner:
looking good... almost there :)

* the part about the "deconfigure" should be done in interfaces.php, we cannot catch all corner cases in the backend
* your first commit seemed better with only affecting dhcpv6 and track6, what made you change it? in dhcpv6 case we need to configure the same interface, in the track6 case we look up the tracked interface.

marjohn56:
We need to be aware that if the user changes away from track6 on the LAN to ANY other setting then the dhcp6c*.conf is not changed unless we call a dhcp6 config change, I found that bit out whilst testing yesterday. The next stage of this is to use the modified dhcp6c which can use sigup to re-read the config.

fichtner:
yes, the "deconfigure" (it's a prepare() on the tracked if setting to be deleted) should be executed in interfaces.php, we can't always rewrite configs especially on unrelated actions later.

marjohn56:
But unless you re-write the config and re-trigger dhcp6c, it will continue to put the tracked interface IP's on the LAN until you restart the WAN, at least that's what happened when I was testing.

marjohn56:
@fichtner  - Your thoughts on my last comment?

I have some hours free this week ( at the moment ) so I can move on with this once this bit is sorted.

fichtner:
I want to pull in the other refactor right away, but discuss the other thing further... it seems the "track6" modelling seems a bit wrong as it cannot be cleaned up properly. Ideally the tracking should be set would be from the parent interface, which makes the config generation and cleanup easier... but with that being said we shouldn't worry too much about configuring after a cleanup and rather focus on the true use cases for setting track6 and reconfiguring it. all other code is bound to severe bitrot and ends up hogging cycles for unknown benefits.

marjohn56:
So for now we can ignore the possibility of a problem if the conf file is not re-written during a LAN config change, and just issue a warning that WAN should be re-saved and applied?

fichtner:
Yes, one step at a time. I'm committing a modified commit that tries to retain what we have now.. there was a small mistake in it for rc.newwanip

marjohn56:
OK, next step?

fichtner:
See 2f12b3d, we detach interface_dhcpv6_prepare() but need to prefix both calls of interface_dhcpv6_configure() now to retain the same behaviour.

With this base change it becomes easier to get the hardest step done to reload the respective tracking configuration with it.

PR #2284, opened 2018-03-22
Inplement_Dpinger

marjohn56:
I am sure there is a lot of tidying up to do on this, and a lot of testing - however it is working on my 18.7 VM which is a good start!

### DIFF
--- a/src/etc/inc/gwlb.inc
+++ b/src/etc/inc/gwlb.inc
@@ -29,121 +29,33 @@
     POSSIBILITY OF SUCH DAMAGE.
 */
 
-function return_apinger_defaults()
-{
-    return array(
-      "latencylow" => "200",
-      "latencyhigh" => "500",
-      "losslow" => "10",
-      "losshigh" => "20",
-      "interval" => "1",
-      "down" => "10",
-      "avg_delay_samples" => "10",
-      "avg_loss_samples" => "50",
-      "avg_loss_delay_samples" => "20"
-    );
+function return_dpinger_defaults() {
+	return array(
+		"latencylow" => "200",
+		"latencyhigh" => "500",
+		"losslow" => "10",
+		"losshigh" => "20",
+		"interval" => "500",
+		"loss_interval" => "2000",
+		"time_period" => "60000",
+		"alert_interval" => "1000",
+		"data_payload" => "0");
+      
 }
 
-function setup_gateways_monitor($verbose = false)
-{
+function setup_gateways_monitor($verbose = false) {
     global $config;
 
     if ($verbose) {
         echo 'Setting up gateway monitors...';
         flush();
     }
-
-    killbypid('/var/run/apinger.pid', 'TERM', true);
-    @unlink('/var/run/apinger.status');
-
+    stop_dpinger();
     @mkdir('/var/db/rrd', 0775);
     @chown('/var/db/rrd', 'nobody');
 
     $gateways_arr_all = return_gateways_array(true);
 
-    $apinger_default = return_apinger_defaults();
-    $apingerconfig = <<<EOD
-
-# OPNsense apinger configuration file. Automatically Generated!
-
-## User and group the pinger should run as
-user "root"
-group "wheel"
-
-## Mailer to use (default: "/usr/lib/sendmail -t")
-#mailer "/var/qmail/bin/qmail-inject"
-
-## Location of the pid-file (default: "/var/run/apinger.pid")
-pid_file "/var/run/apinger.pid"
-
-## Format of timestamp (%s macro) (default: "%b %d %H:%M:%S")
-#timestamp_format "%Y%m%d%H%M%S"
-
-status {
-  ## File where the status information should be written to
-  file "/var/run/apinger.status"
-  ## Interval between file updates
-  ## when 0 or not set, file is written only when SIGUSR1 is received
-  interval 5s
-}
-
-########################################
-# RRDTool status gathering configuration
-# Interval between RRD updates
-rrd interval 60s;
-
-## These parameters can be overridden in a specific alarm configuration
-alarm default {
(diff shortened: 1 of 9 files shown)


fichtner:
We can take this on a branch, it should no go directly to master. Whether or not the service is called apinger or dpinger, keeping function names cleans up the PR, and I'm almost entirely sure we should make this optional / switchable in the beginning.

marjohn56:
Well it cannot be made a plugin as there are script changes. Making it optional to use Apinger or Dpinger seems a lot of work to me.

marjohn56:
So what branch do you want to put this on?

marjohn56:
I'll have a go at making this an option to use Apinger or Dpinger...

fichtner:
@marjohn56 I will help, but I have a book review to finish this weekend...

marjohn56:
Going to look at bit deeper at this. Although it all works, when the WAN goes down, sometimes and sort of one in six cases, the WAN v6 instance of dpinger fails to restart. Now, I am also playing with Monit, that's monitoring the instances nicely. My thoughts are that it may be better to fire of dpinger instances as separate scripts or something like that so Monit or other can restart it required.

marjohn56:
OK, so now I have it working as a user option setting, but now I need some help with the system_gateways_edit.php. When we select advanced we need to switch between the settings for apinger and the settings for dpinger. Although some of the params are the same, most are not. So my thoughts are that when advanced is selected, currently it displays the apinger defaullts, we need that switchable to display the dpinger defaults if dpinger is selected and write the values as dpinger variables for that gateway. This is where I am stuck though, when we get into the html bits I go cross-eyed!

fichtner:
Is it a global setting?

marjohn56:
Is what a global setting, use dpinger? Yes.

fichtner:
if you have the diff against the edit page I can add the switch if you want.

marjohn56:
OK, I'll do that. :)

But first we need decide how we save those advanced settings, currently, each gateway in the xml is like this:
 <gateways>
    <gateway_item>
      <interface>wan</interface>
      <gateway>dynamic</gateway>
      <name>WAN_PPPOE</name>
      <weight>1</weight>
      <ipprotocol>inet</ipprotocol>
      <interval/>
      <descr>Interface WAN_PPPOE Gateway</descr>
      <avg_delay_samples/>
      <avg_loss_samples/>
      <avg_loss_delay_samples/>
      <monitor>8.8.8.8</monitor>
      <defaultgw>1</defaultgw>
    </gateway_item>

I propose for dpinger we use something like:

 <gateways>
    <gateway_item>
      <interface>wan</interface>
      <gateway>dynamic</gateway>
      <name>WAN_PPPOE</name>
      <weight>1</weight>
      <ipprotocol>inet</ipprotocol>
      <interval/>
      <descr>Interface WAN_PPPOE Gateway</descr>
      <dp_avg_delay_samples/>
      <dp_avg_loss_samples/>
      <dp_avg_loss_delay_samples/>
      <monitor>8.8.8.8</monitor>
      <defaultgw>1</defaultgw>
    </gateway_item>

to differentiate, obviously, apinger and dpinger don't use all the same variables, in apinger some are in seconds whereas dpinger it's milliseconds etc.

fichtner:
This came in mangled by github, but I see your proposal during edit... I think the names should be the same but increase depth by &lt;options&gt;&lt;/options&gt; block perhaps, but it also means we can't easily reuse/migrate them in either case. If we skip the extra resolution for now for compat would that be bad for the time being?

marjohn56:
I think we’ll need to increase the options, as for example Apinger you set an alarm if number of failed pings, with dpinger it says percentage of failed pings.

 

Here’s what dpinger says.

 

options:

    -f run in foreground

    -R rewind output file between reports

    -S log warnings via syslog

    -P priority scheduling for receive thread (requires root)

    -B bind (source) address

    -s time interval between echo requests (default 500ms)

    -l time interval before packets are treated as lost (default 4x send interval)

    -t time period over which results are averaged (default 60s)

    -r time interval between reports (default 1s)

    -d data length (default 0)

    -o output file for reports (default stdout)

    -A time interval between alerts (default 1s)

    -D time threshold for latency alarm (default none)

    -L percent threshold for loss alarm (default none)

    -C optional command to be invoked via system() for alerts

    -i identifier text to include in output

    -u unix socket name for polling

    -p process id file name

 

  notes:

    IP addresses can be in either IPv4 or IPv6 format

 

    time values can be expressed with a suffix of 'm' (milliseconds) or 's' (seconds)

    if no suffix is specified, milliseconds is the default

 

    the output format is "latency_avg latency_stddev loss_pct"

    latency values are output in microseconds

    loss percentage is reported in whole numbers of 0-100

    resolution of loss calculation is: 100 * send_interval / (time_period - loss_interval)

 

    the alert_cmd is invoked as "alert_cmd dest_addr alarm_flag latency_avg loss_avg"

    alarm_flag is set to 1 if either latency or loss is in alarm state

    alarm_flag will return to 0 when both have have cleared alarm state

marjohn56:
What we can do is leave the 'extra bits' in dpinger at the default settings. With the second/millisecond thing that's easy enough to resolve, just multiply the value by 1000 if using dpinger. I'll try and get something though this week. Rather busy week ahead with real work but I'll see what I can get done.

fichtner:
Sounds like a plan. To ease your testing for yourself and others in the future, dpinger is now a dependency in the devel version.

marjohn56:
Cool. What I am doing is 'translating' from one set of defaults to the other in system_gateways_edit.php, Seems the only logical way of doing it and keeping just the one set of values in the config if set.

marjohn56:
It's even easier than I thought. Reading a bit more says that you can use Seconds instead of mS just by specifying in the command line. I'll try and get something new on this this afternoon/evening. I have worked out how to hide the non-relevant bits in the web part of the php.

marjohn56:
@fichtner - it is going to be better to write the options separately for dpinger. That way you can switch between apinger and dpinger without affecting any custom options.

marjohn56:
Closed in favour of #2326

PR #2293, opened 2018-03-25
Fix help text in Traffic Shaper masks

mimugmail:
Make it clearer, reference:
https://www.hyperois.com/members/knowledgebase.php?action=displayarticle&id=1


src/opnsense/mvc/app/controllers/OPNsense/TrafficShaper/forms/dialogPipe.xml
```diff
@@ -28,8 +28,9 @@
         <label>mask</label>
         <type>dropdown</type>
         <help><![CDATA[Dynamic queue creation by source or destination address.
-        <li> choose destination here to share the total bandwidth of this pipe among all connected clients. </li>
-        <li> choose source to provide all connected clients up to a maximum of this pipe configured bandwidth  </li>
+        <li> Choose destination to give every IP in destination field of rules the specified bandwith. A pipe with 1Mbit e.g. would let 3 clients lend 1Mbit each so 3Mbit max. Normally this is used for download pipes. </li>
```
fichtner: Can we lose the " " padding before and after the string?

AdSchellevis:
@mimugmail thanks, this should help

AdSchellevis:
@mimugmail just read over our text again, our description is a bit confusing still, how about https://github.com/opnsense/core/commit/6f23e9cc1ff51694c431ba15f45dcc61f357924d ?

AdSchellevis:
@mimugmail @jschellevis with all the discussions going on, I thought to read some of the documentation too.  We probably should also add "buckets" as an option, now our shaper is limited to 256 flows per pipe or queue, set using "net.inet.ip.dummynet.hash_size" in our ipfw startup, which is a fine default, but likely to small for some installations/rule combinations.

I'll add a new issue for this one and add the needed sections.

PR #2295, opened 2018-03-26
updated link for anti lockout

mimugmail:
Seems this option was in _admin and moved to _firewall. Now it links to the correct page for editing anti lockout rule


AdSchellevis:
nice catch!

fichtner:
looks good, thanks!

PR #2297, opened 2018-03-26
Add delay option to pipe in Traffic Shaper

mimugmail:
(only supported in pipe's)


AdSchellevis:
@mimugmail looks good, thanks!

PR #2301, opened 2018-03-27
Fix incorrect display

marjohn56:
Delegated Prefixes display was incorrect.


fichtner:
Looks good, merged... thanks!

Issue #2292, opened 2018-03-25
Larger CPU Graph


bigops:
Is it possible to have a larger CPU Graph in the dashboard?  The current graph is very compressed and it is difficult to get an accurate status of the CPU load.  This could even by a separate widget which can be added in addition to the small one in the dashboard.  The attached image is from a test for CPU utilization when downloading at gigabit speeds and it is very difficult to ascertain how much CPU is being used.  

![image](https://user-images.githubusercontent.com/26052053/37871623-dfae7256-2fb7-11e8-80b7-d1b409f4602a.png)

marjohn56:
How large would you like it?

![image](https://user-images.githubusercontent.com/17962918/38080685-0e0e916e-333a-11e8-8cd1-2edf576db033.png)

marjohn56:
Merged

fichtner:
Very cool, thanks!

bigops:
Thanks

PR #2307, opened 2018-03-29
Standalone cpu load widget

marjohn56:
Someone wanted a larger CPU widget... actually so do I :)

Ref Issue #2292


marjohn56:
Somethings amiss with my repo... time for a cleanup

PR #2309, opened 2018-03-29
Large CPU widget

marjohn56:


marjohn56:
Large CPU Widget ref #2292

fichtner:
Merged, thanks. Need another file and a fixup for the plist, follows in a second. :)

marjohn56:
What's the plist for and where and what needs doing to it... and why is Donald Trump president of the USA.... so many questions, so little time..

fichtner:
plist is for QA purpose, it lists all files that the repo installs:

    # make plist

or

    # make plist-fix

marjohn56:
Ah, understood, Should I have adjusted that or is that down to your good self?

fichtner:
If new files are added or deleted it's ok to push it with the PR, "make upgrade" installs the current git state in the system and does all the sanity checks so it tells you if e.g. plist fails

fichtner:
(of course, that only works in a live system)

marjohn56:
I see it...

PR #2312, opened 2018-03-30
Fix gateway creation for GRE/GIF tunnels

DasTestament:
I've noticed that there is no GWv4 created for GIF tunnel with configured IPv4 address, but obviously useless GWv6 is successfully created. Current base code does not have any logic for GWv4 creation at all which made me very devastatingly upset so even sun stopped rising every morning upon my house.

This fix allows gateway creation accordingly to the currently configured address space on given iface.

I hope this piece of nano fix will do good.

### DIFF
--- a/src/etc/inc/gwlb.inc
+++ b/src/etc/inc/gwlb.inc
@@ -498,6 +498,7 @@ function return_gateways_array($disabled = false, $localhost = false, $inactive 
                 $ctype = $ifcfg['ipaddr'];
                 break;
             default:
+                $tunnelif = substr($ifcfg['if'], 0, 3);
                 if (substr($ifcfg['if'], 0, 4) ==  "ovpn") {
                     // if current iface is an ovpn server endpoint then skip it
                     if (substr($ifcfg['if'], 4, 1) == 's') {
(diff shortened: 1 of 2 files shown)


fichtner:
Just a remark: The v6 Gateway for gif(4) is not useless, it is for Hurricane Electric Tunnel setups :) https://docs.opnsense.org/manual/how-tos/ipv6_tunnelbroker.html

FWIW, I think this is a fair addition. Handing over to @AdSchellevis for merge.

DasTestament:
You misunderstood me. I meant that if **GIF/GRE tunnel local address** has IPv4 address then GWv6 is totally useless, not **GIF/GRE remote address** =)

fichtner:
You are right, the current implementation is too simplistic / pragmatic in this regard. 😊

AdSchellevis:
sorry, but this looks overly complicated to me.. how about https://github.com/opnsense/core/commit/8cf31215f55d14c47d5a44f836ef0a1a2ea9333c ?

fichtner:
looks good at first glance , but need to redo tunnel setup to be 100% sure. :)

DasTestament:
Well, if 8cf3121 implements the very same logic as in this PR then it's ok by me :)

fichtner:
@DasTestament please test to make sure we're not missing anything! thanks 👍

DasTestament:
Sure thing. Will do it 2morrow.

DasTestament:
Ran few tests and it seems proposed code works as expected ... but, I'm a bit worried that **legacy_interfaces_details()** depends strictly on output order of **ifconfig** ... unlike my code. In theory, proposed code will fail on output order of **ifconfig** change.

AdSchellevis:
**ifconfig** contains what's actually configured, which is generally better/safer to use.

DasTestament:
okay

fichtner:
@DasTestament this will be in 18.1.7 :)

DasTestament:
Ok, waiting for .7 release =)

PR #2313, opened 2018-03-31
add support for wildcard entries in unbound

EugenMayer:
Using * as a hostname, one now create wildcard domains in unbound.
```
redirect
                 The query is answered from the local data for the zone  name.
                 There  may  be  no  local  data  beneath the zone name.  This
                 answers queries for the zone, and all subdomains of the  zone
                 with the local data for the zone.  It can be used to redirect
                 a domain to return a different  address  record  to  the  end
                 user,    with   local-zone:   "example.com."   redirect   and
                 local-data: "example.com. A 127.0.0.1" queries for  www.exam-
                 ple.com and www.foo.example.com are redirected, so that users
                 with web browsers  cannot  access  sites  with  suffix  exam-
                 ple.com.
```

### DIFF
--- a/src/etc/inc/plugins.inc.d/unbound.inc
+++ b/src/etc/inc/plugins.inc.d/unbound.inc
@@ -590,8 +590,15 @@ function unbound_add_host_entries()
             switch ($host['rr']) {
                 case 'A':
                 case 'AAAA':
-                    $unbound_entries .= "local-data-ptr: \"{$host['ip']} {$host['host']}{$host['domain']}\"\n";
-                    $unbound_entries .= "local-data: \"{$host['host']}{$host['domain']} IN {$host['rr']} {$host['ip']}\"\n";
+                    // handle wildcard entries which have * as a hostname. Since we added a . above, we match on *.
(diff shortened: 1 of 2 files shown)


AdSchellevis:
@EugenMayer practical addition, can you extend the help text to reflect the new feature as well https://github.com/opnsense/core/blob/18.1.5/src/www/services_unbound_host_edit.php#L183 ?


Thanks in advance

EugenMayer:
That makes a lot of sense, surely. On it

EugenMayer:
@AdSchellevis does this make sense to you this way?

AdSchellevis:
Yes, thanks!

EugenMayer:
Thank you for merging! Also tested with  https://github.com/EugenMayer/opnsense-unbound-plugin 
  and https://github.com/EugenMayer/opnsense-cli using REST api, worked without any issues - so its working on the REST side too

EugenMayer:
if anybody is interested in pre-testing, that would be the patch line

`opnsense-patch c4c28ff6b1b07bf2949a0073778fbfd99a917f81`

PR #2315, opened 2018-03-31
Fix hostname sanitizing to allow asterisk for wildcards

EugenMayer:
Follow up for https://github.com/opnsense/core/pull/2313 , fixing form validation not allowing an asterisk ( i was testing with REST before, sorry )


AdSchellevis:
things happen :)

AdSchellevis:
@fichtner thanks for catching that one, I totally overlooked it.

EugenMayer:
good catch, that's what happens if you just exit using the editor on github.com. sorry

PR #2326, opened 2018-04-03
Dpinger   with selectable use of apinger or dpinger

marjohn56:
This version of Dpinger allows the user to select whether to use dpinger or apinger, the setting is in System->General->Settings.

 I have tried wherever possible to keep functions seperate, thus retaining as much of the existing flow as possible. Alarms checked and working.  Some  advanced settings are now hidden when dpinger is selected.

@Franco can you look at deleting the RRD files whenever the system flips from apinger to dpinger or vice versa, I searched but could not find that.

### DIFF
--- a/src/etc/inc/gwlb.inc
+++ b/src/etc/inc/gwlb.inc
@@ -1,6 +1,7 @@
 <?php
 
 /*
+    Copyright (C) 2018 Martin Wasley <martin@team-rebellion.net>
     Copyright (C) 2016-2018 Franco Fichtner <franco@opnsense.org>
     Copyright (C) 2008 Bill Marquette <bill.marquette@gmail.com>
     Copyright (C) 2008 Seth Mos <seth.mos@dds.nl>
@@ -52,13 +53,19 @@ function setup_gateways_monitor($verbose = false)
         echo 'Setting up gateway monitors...';
         flush();
     }
-
+    stop_dpinger();
     killbypid('/var/run/apinger.pid', 'TERM', true);
     @unlink('/var/run/apinger.status');
 
     @mkdir('/var/db/rrd', 0775);
     @chown('/var/db/rrd', 'nobody');
 
+    if (isset($config['system']['prefer_dpinger'])) {        
+        setup_dpinger_gateways_monitor($verbose);
+        return;
+    }
+    
+    $monitor_ips = array();
     $gateways_arr_all = return_gateways_array(true);
 
     $apinger_default = return_apinger_defaults();
@@ -394,33 +401,71 @@ function return_gateways_status($byname = false)
 
     $status = array();
 
-    if (isvalidpid('/var/run/apinger.pid')) {
-        /* always get the latest status from apinger */
-        killbypid('/var/run/apinger.pid', 'USR1');
+    if (!isset($config['system']['prefer_dpinger'])) {
+    
+        if (isvalidpid('/var/run/apinger.pid')) {
+            /* always get the latest status from apinger */
+            killbypid('/var/run/apinger.pid', 'USR1');
 
-        /* we may read the wrong file here as it's async: */
-        if (file_exists('/var/run/apinger.status')) {
-            $apingerstatus = file('/var/run/apinger.status');
+            /* we may read the wrong file here as it's async: */
+            if (file_exists('/var/run/apinger.status')) {
+                $apingerstatus = file('/var/run/apinger.status');
 
-            foreach ($apingerstatus as $line) {
-                $info = explode('|', $line);
-                if ($byname == false) {
-                    $target = $info[0];
-                } else {
-                    $target = $info[2];
+                foreach ($apingerstatus as $line) {
+                    $info = explode('|', $line);
+                    if ($byname == false) {
+                        $target = $info[0];
+                    } else {
+                        $target = $info[2];
+                    }
+
+                    $status[$target] = array();
+                    $status[$target]['monitorip'] = $info[0];
+                    $status[$target]['srcip'] = $info[1];
+                    $status[$target]['name'] = $info[2];
+                    $status[$target]['delay'] = sprintf('%0.1f ms', empty($info[6]) ? 0.0 : round($info[6], 1));
+                    $status[$target]['stddev'] = 'n/a';
+                    $status[$target]['loss'] = sprintf('%0.1f %%', empty($info[7]) ? 0.0 : round($info[7], 1));
+                    $status[$target]['status'] = trim($info[8]);
                 }
-
-                $status[$target] = array();
-                $status[$target]['monitorip'] = $info[0];
-                $status[$target]['srcip'] = $info[1];
-                $status[$target]['name'] = $info[2];
-                $status[$target]['delay'] = sprintf('%0.1f ms', empty($info[6]) ? 0.0 : round($info[6], 1));
-                $status[$target]['loss'] = sprintf('%0.1f %%', empty($info[7]) ? 0.0 : round($info[7], 1));
-                $status[$target]['status'] = trim($info[8]);
             }
         }
-    }
+    } else {
 
+        $dpinger_gws = running_dpinger_processes();
+        $status = array();
+
+        $gateways_arr = return_gateways_array();
+        foreach ($dpinger_gws as $gwname => $gwdata) {
+            // If action is disabled for this gateway, then we want a detailed status.
+            // That reports "highdelay" or "highloss" rather than just "down".
+            // Because reporting the gateway down would be misleading (gateway action is disabled)
+            $detailed = $gateways_arr[$gwname]['action_disable'];
+            $dpinger_status = get_dpinger_status($gwname, $detailed);
+            if ($dpinger_status === false) {
+                continue;
+            }
+
+                                                   
+                                                
+            if ($byname == false) {
+                $target = $dpinger_status['targetip'];
+            } else {
+                $target = $gwname;
+            }
+
+            $status[$target] = array();
+            $status[$target]['monitorip'] = $dpinger_status['targetip'];
+            $status[$target]['srcip'] = $dpinger_status['srcip'];
+            $status[$target]['name'] = $gwname;
+            $status[$target]['delay'] = empty($dpinger_status['latency_avg']) ? "0ms" : $dpinger_status['latency_avg'] . "ms";
+            $status[$target]['stddev'] = empty($dpinger_status['latency_stddev']) ? "0ms" : $dpinger_status['latency_stddev'] . "ms";
+            $status[$target]['loss'] = empty($dpinger_status['loss']) ? "0.0%" : round($dpinger_status['loss'], 1) . "%";
+            $status[$target]['status'] = $dpinger_status['status'];
+        
+        }
+
+    }
     /* tack on any gateways that have monitoring disabled
      * or are down, which could cause gateway groups to fail */
     $gateways_arr = return_gateways_array();
@@ -450,6 +495,9 @@ function return_gateways_status($byname = false)
             $target = $gwitem['name'];
             $status[$target]['name'] = $gwitem['name'];
             $status[$target]['delay'] = '0.0 ms';
+            if (isset($config['system']['prefer_dpinger'])) {
+                $status[$target]['stddev'] = '0.0 ms';
+            }
             $status[$target]['loss'] = '100.0 %';
             $status[$target]['status'] = 'down';
         } else {
@@ -457,6 +505,9 @@ function return_gateways_status($byname = false)
             $status[$target]['srcip'] = $srcip;
             $status[$target]['name'] = $gwitem['name'];
             $status[$target]['delay'] = '0.0 ms';
+            if (isset($config['system']['prefer_dpinger'])) {
+                $status[$target]['stddev'] = '0.0 ms';
+            }
             $status[$target]['loss'] = '0.0 %';
             $status[$target]['status'] = 'none';
         }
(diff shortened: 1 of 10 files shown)


marjohn56:
Update 1 - Added rc.gateway-alarm script. We can use it without, but we lose logging info.
Update 2 - Added fix for monit, it does no like ':' in the pid filenames, so str_replace to '-',

src/etc/inc/plugins.inc.d/core.inc
```diff
@@ -82,16 +82,48 @@ function core_services()
...
-    if (is_apinger_enabled()) {
-        $pconfig = array();
-        $pconfig['name'] = "apinger";
-        $pconfig['description'] = gettext("Gateway Monitoring Daemon");
-        $pconfig['php']['restart'] = array('setup_gateways_monitor');
-        $pconfig['php']['start'] = array('setup_gateways_monitor');
-        $pconfig['pidfile'] = '/var/run/apinger.pid';
-        $services[] = $pconfig;
-    }
+    if (!isset($config['system']['prefer_dpinger'])) {
```
fichtner: if you push this to is_apinger_enabled and is_dpinger_enabled this doesn't have to be evaluated here and dpinger plugs in seamlessly here

src/etc/inc/plugins.inc.d/core.inc
```diff
@@ -82,16 +82,48 @@ function core_services()
...
+
+        foreach($gateways_arr_all as $name => $gateway) {
+            if (isset($gateway['monitor_disable'])) {
+                continue;
+            }
+
+            $pattern = "/^dpinger_({$name})~([^~]+)~([^~]+)\.pid$/";
+            foreach ($pidfiles as $pidfile) {
+                if (preg_match($pattern,basename($pidfile), $matches)) {
+                    $dpinger_pid = $pidfile;
```
fichtner: wouldn't it be enough to construct the pid file here instead of glob and find ?

this loop must also take into account monitoring_disabled flag and skip that particular dpinger process because it doesn't exist

src/etc/inc/plugins.inc.d/core.inc
```diff
@@ -82,16 +82,48 @@ function core_services()
...
+                }
+            }
+            $pconfig = array();
+            $pconfig['name'] = "dpinger {$gateway['name']}";
+            $pconfig['description'] = gettext("Gateway Monitoring Daemon");
```
fichtner: We need to embed the gateway name here, e.g. sprintf(gettext('Gateway Monitor %s'), $the_gateway_name)

src/etc/inc/plugins.inc.d/core.inc
```diff
@@ -82,16 +82,48 @@ function core_services()
...
+            }
+
+            $pattern = "/^dpinger_({$name})~([^~]+)~([^~]+)\.pid$/";
+            foreach ($pidfiles as $pidfile) {
+                if (preg_match($pattern,basename($pidfile), $matches)) {
+                    $dpinger_pid = $pidfile;
+                }
+            }
+            $pconfig = array();
+            $pconfig['name'] = "dpinger {$gateway['name']}";
```
fichtner: and just 'dpinger' here

src/etc/inc/plugins.inc.d/core.inc
marjohn56: just belts and braces... but I'll remove it if you think it's unnecessary.

src/www/system_gateways_edit.php
```diff
@@ -764,6 +767,7 @@ function calculated_change() {
                     </div>
                   </td>
                 </tr>
+<?php           if (!isset($config['system']['prefer_dpinger'])):?>
```
fichtner: this looks very nice, thanks ❤️

src/www/system_general.php
```diff
@@ -46,6 +46,7 @@
     $pconfig['hostname'] = $config['system']['hostname'];
     $pconfig['language'] = $config['system']['language'];
     $pconfig['prefer_ipv4'] = isset($config['system']['prefer_ipv4']);
+    $pconfig['prefer_dpinger'] = isset($config['system']['prefer_dpinger']);
```
fichtner: note to self: should be in misc, along with the default gateway switching option which is currently under firewall

(it's not a basic setting)

src/etc/rc.gateway_alarm
```diff
@@ -0,0 +1,21 @@
+#!/bin/sh
+#
```
fichtner: this file is better off with a copyright block and your copyright line

src/etc/inc/services.inc
```diff
@@ -1897,3 +1896,26 @@ function service_control_restart($name, $extras)
...
+function is_dpinger_enabled()
+{
+    global $config;
+
+    $gwcount=0;
+
+    if (isset($config['gateways']['gateway_item'])) {
+        foreach($config['gateways']['gateway_item'] as $gwkey => $gateway) {
+            if (!isset($gateway['monitor_disable']) || $gateway['monitor_disable'] == "0") {
+                if (isset($config['system']['prefer_dpinger'])) {
```
fichtner: ah, already here... nice, but you can move this to the top :)

src/www/widgets/widgets/gateways.widget.php
```diff
@@ -78,6 +80,7 @@ function gateways_widget_update(sender, data)
         <tr>
             <th><?=gettext('Name')?></th>
             <th><?=gettext('RTT')?></th>
+            <th><?=gettext('Dev')?></th>
```
fichtner: Let's use STDEV consistently... https://www.abbreviations.com/abbreviation/Standard+Deviation

src/www/widgets/api/plugins/gateway.inc
```diff
@@ -49,6 +50,9 @@ function gateway_api()
                 case "delay":
                     $gatewayItem['status_translated'] = gettext("Latency");
                     break;
+                case "stdev":
+                    $gatewayItem['status_translated'] = gettext("StDev");
```
fichtner: STDEV

src/etc/inc/plugins.inc.d/core.inc
marjohn56: The pid filename is created in gwlib and the name is horribly complex. At first I could see no reason for the name being as complex as it is... but after thinking about it, I can see the reason behind it,.

src/www/system_gateways.php
```diff
@@ -297,6 +297,7 @@ function delete_gateway_item($id, $a_gateways)
                     <th class="hidden-xs hidden-sm hidden-md"><?=gettext("Gateway"); ?></th>
                     <th class="hidden-xs hidden-sm hidden-md"><?=gettext("Monitor IP"); ?></th>
                     <th class="text-nowrap hidden-xs"><?=gettext("RTT"); ?></th>
+                    <th class="text-nowrap hidden-xs"><?=gettext("RTTd"); ?></th>
```
fichtner: STDEV

src/etc/inc/gwlb.inc
```diff
@@ -52,13 +52,20 @@ function setup_gateways_monitor($verbose = false)
         echo 'Setting up gateway monitors...';
```
fichtner: this file should hold your copyright, too

src/etc/inc/plugins.inc.d/core.inc
marjohn56: No, there are multiple dpinger instances.

src/etc/rc.gateway_alarm
marjohn56: It's a rip of the pf* file, I cannot really change it to copyright opnsense... :)

src/etc/inc/plugins.inc.d/core.inc
fichtner: which is? :)

src/www/widgets/api/plugins/gateway.inc
marjohn56: Actually it should be RTTd, - That changed in V2 but fell through my review.

src/www/widgets/widgets/gateways.widget.php
marjohn56: Still think it should be RTTd :)

src/www/widgets/api/plugins/gateway.inc
fichtner: Fine, it should simply be the same everywhere

src/etc/inc/plugins.inc.d/core.inc
marjohn56: You may have multiple targets on the same interface.

src/etc/rc.gateway_alarm
fichtner: it's a problem, because if they don't assert a license this is not open source licensed code tainting their product

then again, this isn't even called anywhere?

src/etc/inc/plugins.inc.d/core.inc
fichtner: but we only have one interface to monitor?

src/etc/inc/plugins.inc.d/core.inc
marjohn56: Currently looks like this:

image

src/etc/inc/interfaces.inc
```diff
@@ -4769,3 +4769,20 @@ function get_carp_interface_status($carpinterface)
     }
     return;
 }
+function interface_wait_tentative($interface, $timeout = 10) {
```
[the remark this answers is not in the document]
marjohn56: OK, I'll test it without and see if there is any difference, Isn't it better to have something rather than nothing though?

src/etc/inc/gwlb.inc
```diff
@@ -1034,3 +1084,411 @@ function validate_address_family($ipaddr, $gwname)
...
+	/* Do not try to bind IPv6 where interface is in tentative state */
+	if (is_ipaddrv6($gateway['gwifip'])) {
+		$err = interface_wait_tentative(get_real_interface(
+		    $gateway['interface']));
+		if ($err == false) {
+			log_error(gettext("Timeout waiting for IPv6 address in tentative state.  dpinger will not run."));
+			return (false);
+		}
+	}
+	return mwexec("/usr/bin/dpinger {$params} {$gateway['monitor']} >/dev/null");
```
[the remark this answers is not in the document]
marjohn56: Okey dokey.

src/etc/rc.gateway_alarm
marjohn56: line 1421 in gwlib.

src/etc/inc/plugins.inc.d/core.inc
marjohn56: I suppose because the target is the variable?  best ask Denny on that one. :)

src/etc/rc.gateway_alarm
fichtner: ok, here's a clean version based on our code. you can use it as is: /usr/local/etc/rc.monitor

fichtner: (https://github.com/opnsense/core/commit/6fb26f989)

src/etc/inc/plugins.inc.d/core.inc
marjohn56: I've moved the catches over to the services...

src/etc/rc.gateway_alarm
marjohn56: I could, but I would lose the alarm logging info wouldn't I?

fichtner: refactored what we have in apinger to prepare for your required changes so you plug the logging into it

marjohn56:
OK.. I think I have it all in place now. Confused on the gateway-alarm bit, are we calling gateway alarm or rc.monitor... or do you want me to take rc.monitor and use that as a template for rc.gateway-alarm?

fichtner:
rc.monitor is a common file for both apinger and dpinger... it mainly reloads the proper services in the right order (there was a fix here recently so we should keep that code in one place for maintenance and clarity)... and so if dpinger is used, which sets a second argument the log line can be added in full

src/etc/inc/gwlb.inc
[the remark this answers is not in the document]
fichtner: this :)

src/etc/inc/interfaces.inc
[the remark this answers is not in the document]
fichtner: it just waits here trying to avoid the error. I want to see the error first... there should be a better way to handle this

marjohn56:
OK..  try that.. I have not touched rc.monitor at this point. Needs the catch for the logging etc.

I have added a little bit in system_general in an attempt to delete the rrd files when the user changes between apinger <> dpinger.

marjohn56:
f*&&^&*!!!!!  GIT it's done me again!!

I'll have to rebuild the PR from my test VM

fabianfrz:
`git reset --hard head^` may do it (removes the last commit)

marjohn56:
Would be nice to think so, but sadly not. It only ever happens with Windows GIT and shell... says it all really.

Fortunately my VM has all the files on there as I copy and test them before a PR, I also copied over the files after modifying them this morning so they are still there. I am making some changes @fichtner wanted to the placement of the config entry and then I'll  push it again..

marjohn56:
OK... back to normal. ;)

Modified rc.monitor a little
Added mwexec call to rm *quality.rrd files on change
moved the config to system->advanced-misc

Think that's it... until you find something else. :)

src/www/system_advanced_misc.php
```diff
@@ -221,8 +228,8 @@ function thermal_modules()
     }
 ?>
       <section class="col-xs-12">
-        <form method="post" name="iform" id="iform">
```
fichtner: looks like this clashes with a newer commit on master, reverting the style update

marjohn56:
Try that.

marjohn56:
@fichtner 

There is a problem if you change the name of the service to just dpinger in core,inc like so:

$pconfig = array();
            $pconfig['name'] = "dpinger";
            $pconfig['description'] = sprintf(gettext('Gateway Monitor %s'), $gateway['name']);
            $pconfig['php']['restart'] = array('setup_gateways_monitor');
            $pconfig['php']['start'] = array('setup_gateways_monitor');
            $pconfig['pidfile'] = $dpinger_pid;
            $services[] = $pconfig;    
            unset($dpinger_pid);

What happens  then is that in the services widget, you have two services called dpinger, and you cannot stop or restart them properly. Afraid it is going to have to stay the way it is, unless you can think of another way,

fichtner:
Solved this for OpenVPN a good while back. I'll take care of it in a manual merge. It needs an additional ID parameter... or the gateway name or so. Easy. :)

marjohn56:
Okay Obi-Wan,  I'll leave you to play with that one. :)

marjohn56:
When you hit start on either one it restarts both anyway, that can be lived with stopping it's not quite so good,

src/etc/rc.monitor
```diff
@@ -37,3 +37,13 @@ fi
...
     "dyndns reload ${GATEWAY}" \
     "rfc2136 reload ${GATEWAY}"
+
+ALARM_ADDR="${2}"
+ALARM_FLAG="${3}"
+ALARM_RTT="${4}"
+ALARM_RTTSD="${5}"
+ALARM_LOSS="${6}"
+
+if [ -z "$#" -eq 6 ]; then
```
fichtner: -z only takes one argument and "sh -n src/etc/rc.monitor" should return a syntax error... best to compare a string here right away: [ "${#}" = "6" ]

src/etc/rc.monitor
```diff
@@ -37,3 +37,13 @@ fi
...
+    echo ">>> Gateway alarm: ${GW} (Addr:${ALARM_ADDR} Alarm:${ALARM_FLAG} RTT:${ALARM_RTT}ms STDEV:${ALARM_RTTSD}ms Loss:${ALARM_LOSS}%)" | /usr/bin/logger -p daemon.info -i -t rc.monitor
```
fichtner: It's ${GATEWAY} in this script

src/etc/rc.monitor
marjohn56: How about if [ "$#" -eq 6 ]

fichtner: works too, practically the same as the string match minus the certainty of how "$#" is converted into a number... :)

marjohn56:
So what I have done will work? I find bourne shell a bit strange,

fichtner:
Yep, will merge this till Friday. Currently without proper Internet and therefore test opportunity -.-

marjohn56:
OK, well I won't be around after tomorrow night for a couple of weeks+ so good luck. :)

fichtner:
alright... enjoy the time off... :)

I'll push out the dhcp6c switch and the associated configuration code rework in 18.1.7. It's ready for release now that 18.1.6 is behind us.


Cheers,
Franco

src/www/widgets/api/plugins/gateway.inc
fabianfrz: Or "Standard Derivation"

marjohn56: Not Derivation, that's from derived.

Deviation :)

fabianfrz: oops

marjohn56:
RRD Fix

fichtner:
Closing as discussed, mostly manually merged so several conflicts in here.

PR #1892, opened 2017-10-25
Add support UTF-8 domain names in black, white and exclude lists.

kekek2:

PR #2327, opened 2018-04-04
Fix errors in IDN.

kekek2:


src/opnsense/mvc/app/controllers/OPNsense/Proxy/Api/SettingsController.php
```diff
@@ -287,8 +288,12 @@ public function setAction()
      */
     public static function encode($domains)
     {
+        if (($domains = $domains) == "") {
+            return "";
+        }
         $result = array();
         foreach (explode(",", $domains) as $domain) {
+            $result[] = ($domain[0] == "." ? "." : "") . idn_to_ascii($domain);
```
AdSchellevis: this looks odd, why are you adding the same output twice?

kekek2:
idn_to_ascii(".samdomain.com") return "samdomain.com". It loses the lead point. To do this, I restore the leading point.

AdSchellevis:
I meant the next line... 

`if ($domain != "") { ..same.. }`

kekek2:
I'm not checking for an empty string. I'm checking to make sure the first character is a dot.

src/opnsense/mvc/app/controllers/OPNsense/Proxy/Api/SettingsController.php
```diff
@@ -287,8 +288,12 @@ public function setAction()
...
             if ($domain != "") {
```
AdSchellevis: ^^^^

kekek2:
I was wrong.

AdSchellevis:
Looking at this again, I kind of doubt this is the best solution for the issue. Data gets converted on multiple occasions now, and if I understand it correctly, we only needed idna names in the configuration file.

I would suggest that we revert the previous commit https://github.com/opnsense/core/pull/1892 and add output filters to Jinja2 to be able to only translate when needed.

Which would result in a template change, a bit like :

`
acl whiteList url_regex {{element|encode_idna|replace(".","\.")}}
` 

And a very small addition to configd, being https://github.com/opnsense/core/commit/cf3ebcae01d2b5a3bc7ea4ca4c9fd55e763981ff

fichtner:
Would this mean we can remove the PHP module for it?

AdSchellevis:
maybe, but there is another pull request which might need it too, although I'm not sure we should pull that one in.

AdSchellevis:
@fichtner we can remove it, the other pull request is not going to be in OPNsense.

PR #2329, opened 2018-04-04
Update classification.config

mimugmail:
We need an update to classification config. It's semi deprecated but VERY useful regarding UX.
You can list categories by dropdown menu, and via serch field also search by application. 

The description text is debateable, it open for suggestions @AdSchellevis @fichtner


AdSchellevis:
are you sure we need this? the ones in the user interface come from the rules itself if I remember correctly.

AdSchellevis:
my mistake, let's pull this in then, if someone has a better description we can always update.

PR #2333, opened 2018-04-06
add PfctlTable wrapper class for dynamic table updates

fabianfrz:
* should allow to allow or block connections dynamically based on a table


AdSchellevis:
I don't think we should do this, rather control the tables using configd to avoid root dependency here. Best would be to also refactor https://opnsense.localdomain/diag_tables.php so our UI and API use the same logic.

fabianfrz:
Ok

PR #2337, opened 2018-04-09
Speedup reloading interfaces.

kekek2:
If WAN-interface has no connection then restart of all services is very slow. The IP address is received by DHCP for a long time and the DNS server is restarted for a long time. Moreover, the DNS server is rebooted three times: once per initialization of each interface and once again on the general reboot of interfaces. I removed the unnecessary reset of the DNS servers on each interface. Restart all services accelerated in half.

### DIFF
--- a/src/etc/inc/interfaces.inc
+++ b/src/etc/inc/interfaces.inc
@@ -955,7 +955,7 @@ function interfaces_configure($verbose = false)
     interfaces_group_setup();
 
     if ($reload == INTERFACE_RELOAD_ALL) {
-        system_routing_configure('', $verbose);
+        system_routing_configure($verbose);
         ipsec_configure_do($verbose);
         plugins_configure('dns', $verbose);
         services_dhcpd_configure('all', array(), $verbose);
--- a/src/etc/inc/interfaces.inc
+++ b/src/etc/inc/interfaces.inc
@@ -29,6 +29,10 @@
  * POSSIBILITY OF SUCH DAMAGE.
  */
 
+const INTERFACE_BOOTING = 1;
+const INTERFACE_RELOAD_SINGLE = 2;
+const INTERFACE_RELOAD_ALL = 3;
+
 require_once("interfaces.lib.inc");
 require_once("gwlb.inc");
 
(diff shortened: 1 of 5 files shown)


src/etc/inc/interfaces.inc
```diff
@@ -950,8 +954,8 @@ function interfaces_configure($verbose = false)
 
     interfaces_group_setup();
 
-    if ($reload) {
-        system_routing_configure($verbose);
+    if ($reload == INTERFACE_RELOAD_ALL) {
+        system_routing_configure('', $verbose);
```
fichtner: this has merge conflicts

src/etc/inc/interfaces.inc
```diff
@@ -905,7 +909,7 @@ function interfaces_configure($verbose = false)
     $track6_list = array();
 
     /* This is needed to speedup interfaces on bootup. */
-    $reload = !file_exists('/var/run/booting');
+    $reload = file_exists('/var/run/booting') ? INTERFACE_BOOTING : INTERFACE_RELOAD_ALL;
```
fichtner: is that really necessary?

src/etc/rc.initial.setlanip
```diff
@@ -558,7 +558,7 @@ echo "done.\n";
 system_hosts_generate(true);
 system_resolvconf_generate(true);
 interface_bring_down($interface);
-interface_configure($interface, true, false, true);
+interface_configure($interface, INTERFACE_RELOAD_SINGLE, false, true);
```
fichtner: isn't that determined by setting $interface in the first place?

src/etc/inc/interfaces.inc
kekek2: https://github.com/kekek2/core/blob/c4b02532caf071a05bff96312041511906837920/src/etc/inc/interfaces.inc#L957

In this line uses variable $reload.

fichtner:
After reviewing this and internal code the conclusion is that reconfiguring should be thorough, not fast. I've adjusted a few things on master, but this here won't go in.

PR #2341, opened 2018-04-10
Fixed creation of GIF tunnel with an outer IPv6 remote address (remote-addr)

noctarius:
When setting up the remote address to be an IPv6 address on an interface with a legal IPv6 address, the tunnel needs to be created using the inet6 parameter before the tunnel keyword, otherwise the command silently fails (return value ignored?).

### DIFF
--- a/src/etc/inc/interfaces.inc
+++ b/src/etc/inc/interfaces.inc
@@ -849,7 +849,11 @@ function interface_gif_configure(&$gif, $gifkey = "")
     }
 
     /* Do not change the order here for more see gif(4) NOTES section. */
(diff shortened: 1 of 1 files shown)


noctarius:
Wow... that was quick!

Glad to help and get my issues fixed :)

fichtner:
Sure, small fix quick merge :)

I did a follow up as GRE had the same issue, touched your code again to use the more preferable mwexecf() for automatic escaping of variables. Would be nice if you could test this for GIF. Looked good from here for both types:

https://github.com/opnsense/core/commit/743eaed9b

PR #2351, opened 2018-04-11
Add Destination IP to Alert Search

jeffg2k:
This adds Destination IP as a search filter for the IDS Alerts.  It also adds a tooltip to the trash can.


AdSchellevis:
@jeffg2k looks good, thanks!

Issue #2349, opened 2018-04-11
IDS/IPS - Remote Syslog Contents


jeffg2k:
Would be nice to have an option in the Remote Syslog Contents for the IDS/IPS sending to Syslog if possible.

fichtner:
Yupp, we've recently enabled syslog by default in IDS... it's been a longer journey on this front due to the focus on eve logging. I'll add this as  separate option... although sending all logs should already work on 18.1.4 and up if enabled in IDS.

jeffg2k:
It would also be cool if we could get the IDS Alerts in a Dashboard Widget.

tuwid:
@jeffg2k I was testing it manually and its possible to send the contents of /var/log/suricata/eve.log via syslog but this would require manually changing the config file here /etc/syslog.conf which suricata can potentially override.

PR #2353, opened 2018-04-12
Intrusion Detection Widget

jeffg2k:
Adds a Intrusion Detection Widget pulling from /var/log/suricata.log as mentioned in #2349 

I just copied the system_log widget and renamed / redirected things.  Worked in the Dashboard without issues that I could see.

### DIFF
--- a/src/www/widgets/widgets/ids_log.widget.php
+++ b/src/www/widgets/widgets/ids_log.widget.php
@@ -1,0 +1,82 @@
+<?php
+
+/*
+    Copyright (C) 2015 S. Linke <dev@devsash.de>
+    All rights reserved.
+
+    Redistribution and use in source and binary forms, with or without
+    modification, are permitted provided that the following conditions are met:
+
+    1. Redistributions of source code must retain the above copyright notice,
+       this list of conditions and the following disclaimer.
+
+    2. Redistributions in binary form must reproduce the above copyright
+       notice, this list of conditions and the following disclaimer in the
+       documentation and/or other materials provided with the distribution.
+
+    THIS SOFTWARE IS PROVIDED ``AS IS'' AND ANY EXPRESS OR IMPLIED WARRANTIES,
+    INCLUDING, BUT NOT LIMITED TO, THE IMPLIED WARRANTIES OF MERCHANTABILITY
+    AND FITNESS FOR A PARTICULAR PURPOSE ARE DISCLAIMED. IN NO EVENT SHALL THE
+    AUTHOR BE LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY,
+    OR CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF
+    SUBSTITUTE GOODS OR SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS
+    INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN
+    CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE)
+    ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
+    POSSIBILITY OF SUCH DAMAGE.
+*/
+
+require_once("guiconfig.inc");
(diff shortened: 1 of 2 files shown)


src/www/widgets/widgets/ids_log.widget.php
```diff
@@ -0,0 +1,82 @@
...
+    ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
+    POSSIBILITY OF SUCH DAMAGE.
+*/
+
+require_once("guiconfig.inc");
+
+$ids_logfile = '/var/log/suricata.log';
+
+if (!$config['widgets']['idslogfiltercount']){
+  $syslogEntriesToFetch = isset($config['syslog']['nentries']) ? $config['syslog']['nentries'] : 20;
```
fabianfrz: Is that really what you want?
syslog != suricata

src/www/widgets/widgets/ids_log.widget.php
```diff
@@ -0,0 +1,82 @@
...
+
+<div id="ids_log-settings" class="widgetconfigdiv" style="display:none;">
+  <form action="/widgets/widgets/ids_log.widget.php" method="post" name="iform">
+    <table class="table table-striped">
+      <tr>
+        <td><?=gettext("Number of Log lines to display");?>:</td>
+        <td>
+          <select name="logfiltercount" id="logfiltercount">
+            <?php for ($i = 1; $i <= 50; $i++) {?>
+            <option value="<?= html_safe($i) ?>" <?php if ($syslogEntriesToFetch == $i) { echo "selected=\"selected\"";}?>><?= html_safe($i) ?></option>
```
fabianfrz: $i is always html_safe - but it can stay as it is.

src/www/widgets/widgets/ids_log.widget.php
```diff
@@ -0,0 +1,82 @@
...
+?>
+
+<div id="ids_log-settings" class="widgetconfigdiv" style="display:none;">
+  <form action="/widgets/widgets/ids_log.widget.php" method="post" name="iform">
+    <table class="table table-striped">
+      <tr>
+        <td><?=gettext("Number of Log lines to display");?>:</td>
+        <td>
+          <select name="logfiltercount" id="logfiltercount">
+            <?php for ($i = 1; $i <= 50; $i++) {?>
```
fabianfrz: (optional) if you want to use the template form - use a : instead of the opening curly bracket here

src/www/widgets/widgets/ids_log.widget.php
```diff
@@ -0,0 +1,82 @@
...
+            <option value="<?= html_safe($i) ?>" <?php if ($syslogEntriesToFetch == $i) { echo "selected=\"selected\"";}?>><?= html_safe($i) ?></option>
+            <?php } ?>
```
fabianfrz: (optional) if you want to use the template form - use endif instead of the closing curly bracket here

src/www/widgets/widgets/ids_log.widget.php
jeffg2k: probably not, but didn't know enough about it.  This entire file is a copy of the system log widget.  Fix as needed :)

fabianfrz: the $config is the root element of the config.xml. 'syslog' is the syslog tag and 'nentries' is a tag inside that which contains the data (here a number). I hope that explains everything.

src/www/widgets/widgets/ids_log.widget.php
```diff
@@ -0,0 +1,82 @@
...
+          <input id="submit_ids_log_widget" name="submit_ids_log_widget" type="submit" class="btn btn-primary formbtn" value="<?= gettext('Save') ?>">
+        </td>
+      </tr>
+    </table>
+  </form>
+</div>
+
+<div id="ids_log-widgets" class="content-box" style="overflow:scroll;">
+  <table class="table table-striped" style="cellspacing:0; cellpadding:0">
+    <?php dump_clog($ids_logfile, $syslogEntriesToFetch); ?>
```
AdSchellevis: Eventually we should feed these through configd, but for now this is ok. Currently you need to be root to be able to read most of the log files in "/var/log", which is kind of a long term goal to stop the web ui from running as root user.

AdSchellevis:
@jeffg2k if you plan to work on it again, feel free to reopen.

PR #2371, opened 2018-04-23
Fixed RA setup of tracked interfaces which should be handled by the s…

noctarius:
…econd foreach loop, not the first one

### DIFF
--- a/src/etc/inc/services.inc
+++ b/src/etc/inc/services.inc
@@ -115,7 +115,7 @@ function services_radvd_configure($blacklist = array())
     foreach ($config['dhcpdv6'] as $dhcpv6if => $dhcpv6ifconf) {
         if (isset($config['interfaces'][$dhcpv6if]['track6-interface'])) {
             continue;
-        } else if (!isset($config['interfaces'][$dhcpv6if]['enable'])) {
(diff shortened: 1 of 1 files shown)


src/etc/inc/services.inc
```diff
@@ -113,7 +113,9 @@ function services_radvd_configure($blacklist = array())
 
     /* handle manually configured DHCP6 server settings first */
     foreach ($config['dhcpdv6'] as $dhcpv6if => $dhcpv6ifconf) {
-        if (!isset($config['interfaces'][$dhcpv6if]['enable'])) {
+        if (isset($config['interfaces'][$dhcpv6if]['track6-interface'])) {
+            continue;
+        } else if (!isset($config['interfaces'][$dhcpv6if]['enable'])) {
```
fichtner: elseif as per PSR style :)

overall this looks like a leftover of manual/static configuration, but it should go in as there is no way to clean this up from the GUI at the moment.

noctarius:
Whatever PSR style is, but it was probably just like common habit.

fichtner:
Boring details: https://www.php-fig.org/psr/psr-2/

Anyway, merged, thanks!

PR #2378, opened 2018-04-27
Update helptext in dialogRule.xml

mimugmail:
We really should define some standards here. In plugins labels start always (or 90%) with upper case and helptext also starts upper case and ends with a dot (if not an IP address or something like that)

### DIFF
--- a/src/opnsense/mvc/app/controllers/OPNsense/TrafficShaper/forms/dialogRule.xml
+++ b/src/opnsense/mvc/app/controllers/OPNsense/TrafficShaper/forms/dialogRule.xml
@@ -38,7 +38,7 @@
         <id>rule.src_port</id>
         <label>src-port</label>
         <type>text</type>
-        <help>source port number or well known name (imap,imaps,http,https,...), for ranges use a dash</help>
+        <help>source port number or well known name (imap, imaps, http, https, ...), for ranges use a dash</help>
     </field>
     <field>
         <id>rule.destination</id>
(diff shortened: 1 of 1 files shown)


fichtner:
Sure here we go:

* first letter uppercase
* all other letters preferably lowercase unless some kind of name or acronym
* the dot only if multiple sentences are used (so at least two dots in total)

Sounds good?

mimugmail:
Does this break translation?
Why a dot only when there is a second sentence .. where does this logic come from?

Perhaps a new issue for this?

fichtner:
yes, it will flush the translation.

bullet points are usually not sentences unless you want to write more elaborate bullet points. I guess it doesn't really matter here if we use a trailing dot

yes please

fichtner:
No, if you want to make consistent progress for the price of invalidating translations your PR should:
* Uppercase the first letter of the line touched
* add a space after comma for readablity

src/opnsense/mvc/app/controllers/OPNsense/TrafficShaper/forms/dialogRule.xml
```diff
@@ -56,7 +56,7 @@
         <id>rule.dst_port</id>
         <label>dst-port</label>
         <type>text</type>
-        <help>destination port number or well known name (imap,imaps,http,https,...), for ranges use a dash</help>
+        <help>destination port number or well known name (imap, imaps, http, https, ...), for ranges use a dash</help>
```
fabianfrz: Möglicherweise ist es auf deinem Computer anders, aber AltGr + `.` ergibt `…` bei mir (drei Punkte als ein Zeichen)

fichtner: Unicode vermeiden :)

fichtner:
@mimugmail If we bounce the translation the first letter in uppercase is really not a thing "tbd"

The question about tbd is for form labels, which asks if every first letter of every word is uppercase or not :)

PR #2387, opened 2018-05-04
Fix IPsec RADIUS secret

rcano-miambiente:
Add double quotes around the RADIUS secret to prevent some corner cases where the secret was not fully read.


AdSchellevis:
@rcano-miambiente nice catch, thanks!

PR #2399, opened 2018-05-08
dpinger STDEV GUI Additions

marjohn56:
Hide STDEV from GUI when Dpinger not enabled

### DIFF
--- a/src/www/system_gateways_edit.php
+++ b/src/www/system_gateways_edit.php
@@ -37,9 +37,9 @@ foreach ($a_gateways as $gw) {
     $a_gateways_arr[] = $gw;
 }
 $a_gateways = $a_gateways_arr;
-if (!isset($config['system']['prefer_dpinger'])) {
-    $apinger_default = return_apinger_defaults();
-} else {
+$apinger_default = return_apinger_defaults();
+
+if (isset($config['system']['prefer_dpinger'])) {
     $apinger_default = return_dpinger_defaults();
 }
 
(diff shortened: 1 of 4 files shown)


src/www/system_gateways_edit.php
```diff
@@ -37,9 +37,9 @@
     $a_gateways_arr[] = $gw;
```
fichtner: you can drop this whole file from the PR

src/www/widgets/widgets/gateways.widget.php
```diff
@@ -78,6 +102,11 @@ function gateways_widget_update(sender, data)
         <tr>
             <th><?=gettext('Name')?></th>
             <th><?=gettext('RTT')?></th>
+<?php
+            if (isset($config['system']['prefer_dpinger'])) :?>                         
+            <th><?=gettext('STDEV')?></th>
```
fichtner: Should it be STDEV or RTTd / RTTD consistently for the visible GUI labels ?

marjohn56: Well my original.. original... was RTTd, then someone said it should be STDEV... so which would you like, Personally I prefer RTTd

fichtner: Word! Let's make it RTTd then.

src/www/system_gateways_edit.php
marjohn56: What ,,, so it becomes

$a_gateways_arr = array();
$a_gateways = return_gateways_array(true, false, true);
if (!isset($config['system']['prefer_dpinger'])) {
...

fichtner: it's ok as is now: return_apinger_defaults is a cheap call and is fixed up when prefer_dpinger is set below

the diff later also rewrites part of unrelated fixes on master

src/www/widgets/widgets/gateways.widget.php
marjohn56: So are you fixing this PR or do you want me to make any changes?

Little confused by the comment above

"it's ok as is now: return_apinger_defaults is a cheap call and is fixed up when prefer_dpinger is set below
the diff later also rewrites part of unrelated fixes on master"

Better I leave you to finish it if it's affecting other files.

fichtner:
@marjohn56 remove all modifications from src/www/system_gateways_edit.php from this PR -- the state in this PR reverts master state in this file and it already works with prefer_dpinger :)

marjohn56:
Slight screw up with the merge for some reason. Reverted it - there should now only be 3 files in the commit 6c61f85.

If you can tell me how to get rid of those merge commits I will.

fichtner:
No need, squashed for maximum effect. 👍

PR #2404, opened 2018-05-15
Dpinger - gateway.inc - missing 'd'

marjohn56:
Missing 'd' in identifiier, does it make a difference?


fichtner:
Sorry, no convincing argument so far, let's just delete this...

PR #2417, opened 2018-05-21
feature: Top level menu item link pivots + security improvements

MaxOrelus:
This feature adds the ability to have parent level menu items that are clickable and that can pivot to an internal link as well as a external link.

I made some improvements on the security stand point for external links due to the fact we are using `target="_blank"`. You can read more about the potential security risks here:

* [Target="_blank" - the most underestimated vulnerability ever](https://www.jitbit.com/alexblog/256-targetblank---the-most-underestimated-vulnerability-ever/)

I've added `rel="noopener noreferrer"` to legacy code as well as additional feature code.

### DIFF
--- a/src/opnsense/mvc/app/views/layout_partials/base_menu_system.volt
+++ b/src/opnsense/mvc/app/views/layout_partials/base_menu_system.volt
@@ -1,53 +1,74 @@
 <aside id="navigation" class="page-side col-xs-12 col-sm-3 col-lg-2 hidden-xs">
     <div class="row">
         <nav class="page-side-nav">
-            <div id="mainmenu" class="panel" style="border:0px" >
+            <div id="mainmenu" class="panel" style="border:0px">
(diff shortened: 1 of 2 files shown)


fabianfrz:
I do already catch the referrer issue here: https://github.com/opnsense/core/blob/master/src/opnsense/mvc/app/controllers/OPNsense/Base/ControllerBase.php#L236

Maybe the `noopener` is enough.

fabianfrz:
BTW: The OpenVPN page has some external links to the client download pages.

MaxOrelus:
FF does not support `noopener`, so adding `noreferrer` as well takes care of it all. Doing a quick `ack` against the repo there's a couple places with external links, but for the scope of this feature request, I made sure to touch both new/legacy.

AdSchellevis:
@MaxOrelus Iooks good, I'll going to pull it in first and do some additional testing then.

PR #2420, opened 2018-05-24
Enhancement: URLs load correctly regardless of trailing slash

MaxOrelus:
In the current OPNsense, URLs that don't end in an extension (`php`) have to end with a trailing slash. If you navigate to the same URL without the trailing slash, the page is not found and you get a "Page not found" page. I went ahead and fixed the routing so that URLs are now dynamic in the aspect of whether you go to `/ui/plugin-page` or `/ui/plugin-page/` the content for that page loads.

The functionality is standard Phalcon stuff, and you can read more about the change I made at the following location:

link: [Dealing with extra/trailing slashes](https://olddocs.phalconphp.com/en/3.0.0/reference/routing.html#dealing-with-extra-trailing-slashes)

I also went ahead and fixed the one menu item that was using a trailing `/` to not use it.

### DIFF
--- a/src/opnsense/mvc/app/library/OPNsense/Core/Routing.php
+++ b/src/opnsense/mvc/app/library/OPNsense/Core/Routing.php
@@ -128,16 +128,16 @@ class Routing
         // where module is mapped to the corresponding namespace
         foreach ($registered_modules as $module_name => $module_configs) {
             $namespace = array_shift($module_configs)['namespace'];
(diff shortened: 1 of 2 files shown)


AdSchellevis:
Hi Max,

Looks good to me, we do need to keep this on devel for some time to make sure there are no side affects, but I'm pulling it right away.

Thanks,

Best regards,

Ad

fichtner:
I like!

PR #2429, opened 2018-05-29
ZFS support in rc and opnsense-importer (from Smart-Soft)

alexpro:
Implementation of ZFS support in Opensense
It is assumed that Opensense root is mounted on a ZFS pool with a random name in
the “root” volume (i.e. "<some_name>"/root). Later any other ZFS filesystems can be mounted in the usual way.

It is assumed that ZFS root does not have fstab entry, same as in FreeBSD system.

Scanning all pools present in the running system is performed in the importer. If ZFS pool
uses the corresponding partition (i.e. - /dev/ada0s1a or /dev/ada0p3 for example) and contains “root” volume – this volume is mounted and then the importer works as usual. Pool is exported before exiting.

### DIFF
--- a/src/etc/rc
+++ b/src/etc/rc
@@ -47,6 +47,7 @@ mount -fr / 2> /dev/null
 
 GROWFS_MARKER=/.probe.for.growfs.nano
 ROOT_IS_UFS=
+ZFS_DEV=
 
 while read FS_PART FS_MNT FS_TYPE FS_MORE; do
 	# only tune our own file systems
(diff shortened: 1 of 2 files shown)


evbevz:
We have ZFS support not so long. About two months in production.

alexpro:
Only last five commits are discarded by my request, all commits from yesterday. I works with latest revisions of rc and opnsense-importer files.

fichtner:
It's all fine and pretty sure it works, but listen, we can't forcefully load the ZFS module on every box. The former code tried to avoid this and it looks like you just trade one implementation for the other which is why I said no.

evbevz:
@fichtner ,  what about bsdinstaller with zfs support?

fichtner:
https://github.com/opnsense/bsdinstaller/issues/11

evbevz:
We can make pull-request to bsdinstaller, if you are interested.

fichtner:
Sure, although I have to say aligning with FreeBSD is a main design goal. This is important for opnsense-bootstrap capabilities.

PR #2438, opened 2018-06-01
fix error string in openvpn wizard

fabianfrz:


fichtner:
Isn't this already on master?? https://github.com/opnsense/core/commit/0b6a41dc

fabianfrz:
then close it

PR #2447, opened 2018-06-06
Fix rules toggle.

kekek2:
If you switch the state of more than one hundred rules in IDS, they are not always switched. In parallel, run multiple AJAX requests simultaneously writing changes to the config.xml. Because of this, some changes may be lost. I made AJAX requests to be sent sequentially.

### DIFF
--- a/src/opnsense/mvc/app/views/OPNsense/IDS/index.volt
+++ b/src/opnsense/mvc/app/views/OPNsense/IDS/index.volt
@@ -212,7 +212,7 @@ POSSIBILITY OF SUCH DAMAGE.
                     const localKeyset = keyset;
                     promise = promise.then(function () {
                         return ajaxCall(url + localKeyset.join(',') + '/' + url_suffix, sendData = {}, null);
-                    })
+                    });
                 }
 
                 // refresh when all toggles are executed
--- a/src/opnsense/mvc/app/views/OPNsense/IDS/index.volt
+++ b/src/opnsense/mvc/app/views/OPNsense/IDS/index.volt
@@ -196,25 +196,23 @@ POSSIBILITY OF SUCH DAMAGE.
                 }
 
                 var keyset = [];
-                $.each(rows, function(key,uuid){
+                $.each(rows, function (key, uuid) {
                     keyset.push(uuid);
-                    if ( combine == undefined || keyset.length > combine) {
-                        (function (localKeyset) {
-                            promise = promise.then(function () {
-                                return ajaxCall(url + localKeyset.join(',') + '/' + url_suffix, sendData = {}, null);
-                            })
-                        }) (keyset);
+                    if (combine == undefined || keyset.length > combine) {
+                        const localKeyset = keyset;
+                        promise = promise.then(function () {
+                            return ajaxCall(url + localKeyset.join(',') + '/' + url_suffix, sendData = {}, null);
+                        });
                         keyset = [];
                     }
                 });
 
                 // flush remaining items
                 if (keyset.length > 0) {
-                    (function (localKeyset) {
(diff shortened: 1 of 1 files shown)


src/opnsense/mvc/app/views/OPNsense/IDS/index.volt
```diff
@@ -199,19 +199,30 @@ POSSIBILITY OF SUCH DAMAGE.
                 $.each(rows, function(key,uuid){
                     keyset.push(uuid);
                     if ( combine == undefined || keyset.length > combine) {
-                        deferreds.push(ajaxCall(url + keyset.join(',') +'/'+url_suffix, sendData={},null));
+                        (function (localKeyset) {
+                            promise = promise.then(function () {
+                                return ajaxCall(url + localKeyset.join(',') + '/' + url_suffix, sendData = {}, null);
+                            })
+                        }) (keyset);
```
fabianfrz: this function is probably not needed

src/opnsense/mvc/app/views/OPNsense/IDS/index.volt
```diff
@@ -199,19 +199,30 @@ POSSIBILITY OF SUCH DAMAGE.
...
                 });
 
                 // flush remaining items
                 if (keyset.length > 0) {
-                    deferreds.push(ajaxCall(url + keyset.join(',') +'/'+url_suffix, sendData={},null));
+                    (function (localKeyset) {
+                        promise = promise.then(function () {
+                            return ajaxCall(url + localKeyset.join(',') + '/' + url_suffix, sendData = {}, null);
+                        })
+                    })(keyset);
```
fabianfrz: same here

src/opnsense/mvc/app/views/OPNsense/IDS/index.volt
kekek2: First I did without closure. All calls AjaxCall was with the last value of the keyset. I had to add this feature.

fabianfrz: I am talking about the outer function, not the promise

kekek2:
var keyset = [];
                $.each(rows, function (key, uuid) {
                    keyset.push(uuid);
                    if (combine == undefined || keyset.length > combine) {
                        promise = promise.then(function () {
                            return ajaxCall(url + keyset.join(',') + '/' + url_suffix, sendData = {}, null);
                        })
                        keyset = [];
                    }
                });

                if (keyset.length > 0) {
                    promise = promise.then(function () {
                        return ajaxCall(url + keyset.join(',') + '/' + url_suffix, sendData = {}, null);
                    })
                }
`
These screenshots I made with this code. All requests are the same.

kekek2:
Screenshots with my code

fabianfrz:
looks like you are loosing the reference - should be fixable by `let` and local variables.

src/opnsense/mvc/app/views/OPNsense/IDS/index.volt
```diff
@@ -196,25 +196,23 @@ POSSIBILITY OF SUCH DAMAGE.
...
                 if (keyset.length > 0) {
-                    (function (localKeyset) {
-                        promise = promise.then(function () {
-                            return ajaxCall(url + localKeyset.join(',') + '/' + url_suffix, sendData = {}, null);
-                        })
-                    })(keyset);
+                    const localKeyset = keyset;
+                    promise = promise.then(function () {
+                        return ajaxCall(url + localKeyset.join(',') + '/' + url_suffix, sendData = {}, null);
+                    })
```
fabianfrz: add a semicolon

PR #2448, opened 2018-06-06
Remove rules with deinstall plugin.

kekek2:


AdSchellevis:
https://github.com/opnsense/core/commit/ac35e91deef0773b737f3763aab32e618677d591 cleans up uninstalled rules after update, leaving the configuration in place.

PR #2460, opened 2018-06-13
Track6_Allow_DHCPD6

marjohn56:
This is the initial commit. There is much more work to do, dhcpd6.conf will need to be re-writen on a prefix change is the major omission at this point, but this commit will allow the user to manually set dhcpd6 range and RADVD options. To enable, go to the System->Settings->General Page and enable Manual IPv6 DHCPD and RA,

### DIFF
--- a/src/etc/inc/services.inc
+++ b/src/etc/inc/services.inc
@@ -113,7 +113,7 @@ function services_radvd_configure($blacklist = array())
 
     /* handle manually configured DHCP6 server settings first */
     foreach ($config['dhcpdv6'] as $dhcpv6if => $dhcpv6ifconf) {
-        if (isset($config['interfaces'][$dhcpv6if]['track6-interface'])) {
+        if (isset($config['interfaces'][$dhcpv6if]['track6-interface']) && !isset($config['interfaces'][$dhcpv6if]['dhcpd6track6allowoverride'])) {
             continue;
         } elseif (!isset($config['interfaces'][$dhcpv6if]['enable'])) {
             continue;
@@ -134,7 +134,7 @@ function services_radvd_configure($blacklist = array())
         }
 
         $ifcfgipv6 = get_interface_ipv6($dhcpv6if);
-        if (!is_ipaddrv6($ifcfgipv6)) {
+        if (!is_ipaddrv6($ifcfgipv6) && !isset($config['interfaces'][$dhcpv6if]['dhcpd6track6allowoverride'])) {
             continue;
         }
 
@@ -1087,16 +1087,23 @@ function services_dhcpdv6_configure($blacklist = array(), $verbose = false)
             $ifcfgipv6 = Net_IPv6::getNetmask($ifcfgipv6, 64);
             $trackifname = $config['interfaces'][$ifname]['track6-interface'];
             $trackcfg = $config['interfaces'][$trackifname];
+            $ifcfgipv6arr = array();
             $pdlen = calculate_ipv6_delegation_length($trackifname);
             $ifcfgipv6arr =explode(":", $ifcfgipv6);
             $dhcpdv6cfg[$ifname] = array();
             $dhcpdv6cfg[$ifname]['enable'] = true;
             /* range */
-            $ifcfgipv6arr[7] = "1000";
-            $dhcpdv6cfg[$ifname]['range'] = array();
-            $dhcpdv6cfg[$ifname]['range']['from'] = Net_IPv6::compress(implode(":", $ifcfgipv6arr));
-            $ifcfgipv6arr[7] = "2000";
-            $dhcpdv6cfg[$ifname]['range']['to'] = Net_IPv6::compress(implode(":", $ifcfgipv6arr));
+            if(!isset($config['interfaces'][$ifname]['dhcpd6track6allowoverride']) || (isset($config['interfaces'][$ifname]['dhcpd6track6allowoverride']) && !isset($config['dhcpdv6'][$ifname]['enable']))) {
+                $ifcfgipv6arr[7] = "1000";
+                $dhcpdv6cfg[$ifname]['range'] = array();
+                $dhcpdv6cfg[$ifname]['range']['from'] = Net_IPv6::compress(implode(":", $ifcfgipv6arr));
+                $ifcfgipv6arr[7] = "2000";
+                $dhcpdv6cfg[$ifname]['range']['to'] = Net_IPv6::compress(implode(":", $ifcfgipv6arr));
+            } else {
+               // Get config entry and marry it to the live prefix          
+               $dhcpdv6cfg[$ifname]['range']['to'] = make_ipv6_64_address($ifcfgipv6,$config['dhcpdv6'][$ifname]['range']['to']);
+               $dhcpdv6cfg[$ifname]['range']['from'] = make_ipv6_64_address($ifcfgipv6,$config['dhcpdv6'][$ifname]['range']['from']);
+            }
             /* prefix length > 0? We can add dhcp6 prefix delegation server */
             if ($pdlen > 2) {
                 $pdlenmax = $pdlen;
@@ -1198,7 +1205,7 @@ EOD;
         if (is_ipaddrv6($ifcfgipv6)) {
             $dhcpdv6conf .= "\nsubnet6 {$subnetv6}/{$ifcfgsnv6}";
         } elseif (!empty($dhcpv6ifconf['range']['from'])) {
-            $subnet6 = gen_subnetv6($dhcpv6ifconf['range']['from'], "64");
+            $subnet6 = gen_subnetv6($dhcpv6ifconf['range']['from'], "64", !isset($dhcpdv6cfg[$ifname]['dhcpd6track6allowoverride']));
             $dhcpdv6conf .= "\nsubnet6 {$subnet6}/64";
         }
         $dhcpdv6conf .= " {\n";
--- a/src/etc/inc/util.inc
+++ b/src/etc/inc/util.inc
@@ -150,9 +150,9 @@ function gen_subnet($ipaddr, $bits)
 }
 
 /* return the subnet address given a host address and a subnet bit count */
-function gen_subnetv6($ipaddr, $bits)
+function gen_subnetv6($ipaddr, $bits, $validate = true)
 {
-    if (!is_ipaddrv6($ipaddr) || !is_numeric($bits)) {
+    if ($validate && (!is_ipaddrv6($ipaddr) || !is_numeric($bits))) {
         return "";
     }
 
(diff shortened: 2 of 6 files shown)


fichtner:
download

Nice, looks simple enough, but review will have to wait a couple of days.

marjohn56:
There's no rush on this, need to get this one right. pf had it but it used to fail when prefix changed, that I want to make sure does not happen here. We'll need a catch somewhere that tells us the prefix has changed and then a call to re-write the dhcpd6.conf. 

1. How do we signal the clients that the prefix has changed?
2. We will need to find a way to update the statically assigned addresses
3. We will need to update Unbound to reflect the changes also.

marjohn56:
After looking at this, I am inclined to disable the subnet prefix ability when in this mode. The issue being that if I request a /56 prefix, I have no way of saying where in the /56 subnet range my LAN interface is going to get an address, and as I cannot have the ranges crossing it makes it possible that a clash can happen.

Thoughts..

marjohn56:
Well it does work, the PD delegation appears to work fine.

fichtner:
Do we need to restart from WAN on IPv6 change? Something similar to this? https://github.com/opnsense/core/commit/3b7c0bca06c776c0d2acdfe78826d27237b62032#diff-1332c372788c9e1a8c6c9bae9ebb55a5R3818

I recently removed a side effect here, maybe it made it worse... https://github.com/opnsense/core/commit/eb1e3966e5ecb9c7f7245c61e7155fee892648a5#diff-1332c372788c9e1a8c6c9bae9ebb55a5L2595

src/www/system_general.php
```diff
@@ -424,6 +426,20 @@
                 </div>
               </td>
             </tr>
+            <tr>
+              <td><a id="help_for_dhcpd6_opt" href="#" class="showhelp"><i class="fa fa-info-circle"></i></a> <?=gettext("Manual IPv6 DHCPD and RA"); ?></td>
+              <td>
+                <input name="dhcpd6track6allowoverride" type="checkbox" value="yes" <?= $pconfig['dhcpd6track6allowoverride'] ? 'checked="checked"' : '' ?>/>
```
fichtner: Shouldn't be global. Either we set this in the interface, or we don't set anything and figure it out from the dhcpv6 server side

src/www/system_general.php
```diff
@@ -307,7 +309,7 @@
             <tr>
               <td><a id="help_for_theme" href="#" class="showhelp"><i class="fa fa-info-circle"></i></a> <?=gettext("Theme"); ?></td>
               <td>
-                <select name="theme" class="selectpicker">
+                <select name="theme" class="selectpicker" data-size="10" data-width="auto">
```
fichtner: merge issue

src/etc/inc/util.inc
```diff
@@ -152,8 +152,10 @@ function gen_subnet($ipaddr, $bits)
 /* return the subnet address given a host address and a subnet bit count */
 function gen_subnetv6($ipaddr, $bits)
 {
-    if (!is_ipaddrv6($ipaddr) || !is_numeric($bits)) {
+    if(!isset($config['system']['dhcpd6track6allowoverride'])) {
```
fichtner: if we need to push such high level override logic into these innocent functions the whole approach is flawed ;)

src/www/system_general.php
marjohn56: We could add it to the interface itself I suppose.

src/www/system_general.php
marjohn56: Need to do a pull... my bad, sorry!

src/etc/inc/util.inc
marjohn56: Well if we do it by interface then we'll still have to do this.

fichtner: makes zero sense to me, the caller of gen_subnetv6() should do what is necessary. I'm guessing that's in the radvd or dhcp6 code?

marjohn56: The key there is the line at 155, if we don't override it there and let the function continue then we return nothing, and we need a range for dhcp6.conf

fichtner: sure, but that's dhcp's problem, not gen_subnetv6(). this code here merely introduces a multitude of regressions into unrelated code if dhcpd6track6allowoverride were set, like OpenVPN...

marjohn56: But it's called by the function that generates dhcpd6.conf, that needs to be updated on a prefix change.

Line 1249 onward in services.inc...

`        if (is_ipaddrv6($ifcfgipv6)) {
            $dhcpdv6conf .= "\nsubnet6 {$subnetv6}/{$ifcfgsnv6}";
        } elseif (!empty($dhcpv6ifconf['range']['from'])) {
            $subnet6 = gen_subnetv6($dhcpv6ifconf['range']['from'], "64");
            $dhcpdv6conf .= "\nsubnet6 {$subnet6}/64";
        }
        $dhcpdv6conf .= " {\n";`

marjohn56: I guess I could create a new function specifically for that, then it does not affect anything else.

fichtner: yes and no, we need to address this specifically in "Line 1249 onward in services.inc..."

marjohn56: We could use the same lines 161 to 163 in from util.inc - the Net_IPV6:: calls and place them directly into services.inc..

fichtner: sooo you want to bypass the sanity check to generate a subnet anyway? why not

```
function gen_subnetv6($ipaddr, $bits, $validate = true)
```

marjohn56: There is no sanity on my desk.. :) Sorry I got distracted setting up my new new Qotom.

OK, so we call the gen_subnetv6() which by default is validate true, but we use the status of the validate flag instead of the config->dhcpd6track6allowoverride. Is that your thought?

fichtner: yep, sounds like a step forward. as far as semantics go "dhcpd6track6allowoverride" doesn't really tell us why we override it so we also need a comment there why we disable validation in this particular case, but first things first...

marjohn56:
OK... let's see where this takes us. :)

src/www/system_general.php
```diff
@@ -137,7 +137,7 @@
         }
 
         $config['system']['dnsallowoverride'] = !empty($pconfig['dnsallowoverride']);
-
+
```
fichtner: small merge hiccup still, shouldn't touch this file :)

src/etc/inc/services.inc
```diff
@@ -1087,16 +1087,67 @@ function services_dhcpdv6_configure($blacklist = array(), $verbose = false)
...
             $ifcfgipv6arr =explode(":", $ifcfgipv6);
             $dhcpdv6cfg[$ifname] = array();
             $dhcpdv6cfg[$ifname]['enable'] = true;
             /* range */
-            $ifcfgipv6arr[7] = "1000";
-            $dhcpdv6cfg[$ifname]['range'] = array();
-            $dhcpdv6cfg[$ifname]['range']['from'] = Net_IPv6::compress(implode(":", $ifcfgipv6arr));
-            $ifcfgipv6arr[7] = "2000";
-            $dhcpdv6cfg[$ifname]['range']['to'] = Net_IPv6::compress(implode(":", $ifcfgipv6arr));
+            if(!isset($config['interfaces'][$ifname]['dhcpd6track6allowoverride']) || (isset($config['interfaces'][$ifname]['dhcpd6track6allowoverride']) && !isset($config['dhcpdv6'][$ifname]['enable']))) {
```
fichtner: this whole block could be clearer, let me try to merge tomorrow. nice work so far!

src/www/interfaces.php
```diff
@@ -598,7 +599,7 @@ function get_wireless_channel_info($interface) {
         if (isset($config['dhcpd']) && isset($config['dhcpd'][$if]['enable']) && (! preg_match("/^staticv4/", $pconfig['type']))) {
             $input_errors[] = gettext("The DHCP Server is active on this interface and it can be used only with a static IP configuration. Please disable the DHCP Server service on this interface first, then change the interface configuration.");
         }
-        if (isset($config['dhcpdv6']) && isset($config['dhcpdv6'][$if]['enable']) && (! preg_match("/^staticv6/", $pconfig['type6']))) {
+        if (isset($config['dhcpdv6']) && isset($config['dhcpdv6'][$if]['enable']) && (! preg_match("/^staticv6/", $pconfig['type6'])) && !isset($config['system']['dhcpd6track6allowoverride'])) {
```
fichtner: Note to self: this validation is very strange. we should opt for a clean service change, not block a reconfigure. Also, this might have been fixed via a0bb26b529b

marjohn56: What's even stranger is this one has sneaked through with $config['system'], so much for a global file search!

src/etc/inc/services.inc
marjohn56: I'm sure there's a better way, but I can't see it at present.

marjohn56:
Heavy mods to services_dhcpv6.php to attempt to make it foolproof. No override of any sanity checks.

marjohn56:
More cleanup... much tidier

fichtner:
I like it ❤️ 

Merge coming up soon...

src/etc/inc/services.inc
```diff
@@ -113,7 +113,7 @@ function services_radvd_configure($blacklist = array())
 
     /* handle manually configured DHCP6 server settings first */
     foreach ($config['dhcpdv6'] as $dhcpv6if => $dhcpv6ifconf) {
-        if (isset($config['interfaces'][$dhcpv6if]['track6-interface'])) {
+        if (isset($config['interfaces'][$dhcpv6if]['track6-interface']) && isset($config['interfaces'][$dhcpv6if]['dhcpd6track6allowoverride'])) {
```
fichtner: @marjohn56 trying to make sense of this. can you walk me through it please?

src/etc/inc/services.inc
```diff
@@ -134,7 +134,7 @@ function services_radvd_configure($blacklist = array())
         }
 
         $ifcfgipv6 = get_interface_ipv6($dhcpv6if);
-        if (!is_ipaddrv6($ifcfgipv6)) {
+        if (!is_ipaddrv6($ifcfgipv6) && !isset($config['interfaces'][$dhcpv6if]['dhcpd6track6allowoverride'])) {
```
fichtner: especially with this in mind...

src/etc/inc/services.inc
fichtner: to be more precise, shouldn't it be !isset() ?

src/etc/inc/services.inc
marjohn56: Out earning pennies at moment
I'll have a look when I get home.

Issue #2459, opened 2018-06-12
Certificates generated using deprecated extensions


hyperlumic:
Any certificate generated in the certificate management interface is generated with a Netscape Cert Type extension indicating the purpose of the certificate. As of OpenSSL v1.1, the nsCertType extension in X.509 certificates are no longer supported. This extension is old and has been deprecated for a long time. OpenVPN has also deprecated the use of this extension quite some time ago, and is soon to remove the ns-cert-type option from OpenVPN entirely.

The correct method is to set the X.509v3 Key Usage and X.509v3 Extended Key Usage fields correctly.

Accordingly, the web UI for OPNsense should stop looking for the nsCertType extension to determine/display if the certificate is a client or server certificate, and follow the KU/EKU extensions.

```
X509v3 extensions:
            X509v3 Basic Constraints: 
                CA:FALSE
            Netscape Cert Type: 
                SSL Server
            Netscape Comment: 
                OpenSSL Generated Server Certificate
            X509v3 Subject Key Identifier: 
                AB:77:8C:16:BA:39:51:49:14:79:F0:33:C5:A6:C6:62:4F:31:22:8D
            X509v3 Authority Key Identifier: 
                keyid:FA:D8:C3:33:05:F7:23:F7:F6:B4:19:E1:F4:0F:27:3B:66:BF:A1:FC
                DirName:/CN=Test CA/C=US
                serial:5B:1E:C6:B5:0A:5C:54:10
            X509v3 Extended Key Usage: 
                TLS Web Server Authentication, 1.3.6.1.5.5.8.2.2
            X509v3 Key Usage: 
                Digital Signature, Key Encipherment
```

AdSchellevis:
question, are the cert's we create now missing  KU/EKU extensions or incomplete? Or is this  issue covering more then UI checks and OpenVPN settings? 

If  the certs created from OPNsense have been good for quite some time, we could probably easily switch, if not, switching will impact users and have to be planned carefully.

hyperlumic:
I should have clarified that!

The certs created now DO include the correct KU/EKU extensions, yes (displayed in the provided output), so that's not a problem.

AdSchellevis:
no problem, thanks for the clarification.

whislock:
It appears that the OpenVPN client configuration required for this is already in place, though with a conflicting comment.

```
src/www/vpn_openvpn_export.php:

// Prevent MITM attacks by verifying the server certificate.
[...]

PR #2463, opened 2018-06-14
Confirm KU/EKU for server certs

whislock:
Addresses https://github.com/opnsense/core/issues/2459.

Server certificates should have KU set to include "Digital Signature" and "Key Encipherment," and EKU set to "TLS Web Server Authentication."

OpenVPN requires both KU and EKU set correctly to successfully validate if client is using remote-cert-tls server configuration parameter.

### DIFF
--- a/src/etc/inc/certs.inc
+++ b/src/etc/inc/certs.inc
@@ -417,6 +417,7 @@ function cert_get_purpose($str_crt, $decode = true)
     if (
         isset($crt_details['extensions']['extendedKeyUsage'])
         && strstr($crt_details['extensions']['extendedKeyUsage'], "TLS Web Server Authentication") !== false
+        && isset($crt_details['extensions']['keyUsage'])
         && strpos($crt_details['extensions']['keyUsage'], "Digital Signature") !== false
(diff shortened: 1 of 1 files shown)


AdSchellevis:
@whislock how did you test your change? 

`Parse error: syntax error, unexpected 'else' (T_ELSE), expecting end of file in /usr/local/etc/inc/certs.inc on line 425
`

whislock:
Wrote a stub PHP script that would call the function against an array of certificates with varying combinations of KU/EKU settings, and verified that each certificate was properly classified. I didn’t have a full development environment set up, so it’s possible that I flubbed copying the modified code back to GitHub. 😣 I won’t do that again.

fabianfrz:
@AdSchellevis at least for that errors a CI check would catch them

fichtner:
# make lint

Problem solved. ;)

fabianfrz:
@fichtner sure but a CI can add an (X) to the PR so it will not be merged because it is already visible that syntax errors exist.

fichtner:
Sorry, that's no excuse for pushing untested changes, where "untested" means the code that is to be included in OPNsense actually works, not if the idea tested elsewhere worked.

PR #2465, opened 2018-06-15
Further KU/EKU fixes.

whislock:
Related to #2459. 

The matching KUs for EKU "TLS Web Server Authentication" are Digital Signature AND (Key Encipherment OR Key Agreement). Added for compatibility with externally-generated certificates.

Removed the nsComment extension from all cert profiles. It's deprecated like nsCertType is, though nsCertType was retained in the server profile for backwards compatibility with ancient OpenVPN clients using the ns-cert-type configuration parameter.

Tested by porting modified code to a running OPNsense firewall and generating a bunch of certificates, all of which function properly.

### DIFF
--- a/src/etc/inc/certs.inc
+++ b/src/etc/inc/certs.inc
@@ -418,8 +418,8 @@ function cert_get_purpose($str_crt, $decode = true)
         strstr($crt_details['extensions']['extendedKeyUsage'], "TLS Web Server Authentication") !== false &&
         isset($crt_details['extensions']['keyUsage']) &&
         strpos($crt_details['extensions']['keyUsage'], "Digital Signature") !== false &&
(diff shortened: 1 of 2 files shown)


whislock:
I had originally intended to remove the nsCertType extension in server cert generation, but I decided that leaving it there for now didn't cause any problems. It's worth noting that OpenSSL 1.1 and later do away with nsCertType and nsComment entirely (they show up as unknown extensions), so it may in fact be worth removing completely. Thoughts?

fichtner:
Can you split this into two separate requests? opnsense.cnf should be modified in one go later on.

You can also read about style at https://www.php-fig.org/psr/psr-2/

whislock:
Closing to split.

whislock:
Rolled back the opnsense.cnf change.

fichtner:
Merged, thanks!

PR #2467, opened 2018-06-16
Make optional cert fields optional.

### LINKED ISSUE #2317: Certificate generation requires entries for optional fields
Attempting to generate a certificate/CA/CSR via the Trust application within the web GUI without providing content for all fields returns errors such as:

```

The field Distinguished name State or Province is required.
The field Distinguished name City is required.
The field Distinguished name Email Address is required.
```

This is not consistent with `RFC 5280, Internet X.509 Public Key Infrastructure Certificate and Certificate Revocation List (CRL) Profile`, namely section 4.1.2.6 that defines how the subject of a certificate (CA or leaf) must be specified.

Specifically, all fields presented in the `Distinguished name` group in the web interface are, technically, optional, though it's hardly recommended to leave them all blank. The simplest solution would be to make the `Common Name` field mandatory, and all others optional, with a warning that some commercial CAs will not accept CSRs without a `Country Code` element present in the `Subject` field of the CSR.

whislock:
Fixes #2317. Makes cert fields other than common name and country code optional.

Tested by generating internal CAs, intermediate CAs, client certs, server certs. Omitted fields are not included as part of the distinguishedName.

### DIFF
--- a/src/etc/ssl/opnsense.cnf
+++ b/src/etc/ssl/opnsense.cnf
@@ -127,17 +127,17 @@ string_mask = utf8only
 
 [ req_distinguished_name ]
 countryName			= Country Name (2 letter code)
-countryName_default		= AU
+#countryName_default		= AU
 countryName_min			= 2
 countryName_max			= 2
 
 stateOrProvinceName		= State or Province Name (full name)
-stateOrProvinceName_default	= Some-State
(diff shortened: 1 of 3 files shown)


readonly24:
Hi, I just have installed OPNsense 18.1.6-amd64 and I get this error, too. When will this fix be in a public release?

PR #2471, opened 2018-06-21
interfaces: check existence of `status`

t123yh:
I was trying to use `get_interfaces_info()` function in one of my plugin apis. I got an error `Undefined index` here because some `intf_detail` didn't have the `status` field. So let's use `isset` to check it first to avoid this error.


fichtner:
Merged, thanks!

Issue #2427, opened 2018-05-27
Show transport encryption fingerprints


ypid:
When using Qubes OS, or just being someone who takes security seriously, you never want to be required to have to trust the network. One uses encryption to ensure this. To be able to detect MITM, you need to verify authenticity of X.509 certificates and SSH host keys. OPNsense could help make this more easy for admins. As I setup a test system now, was able to never need to trust the network during setup. This is what I did:

1. Install using direct console access.
2. Configure network and login on console.
3. Manually print the X.509 SHA256 fingerprint for verification when first connecting to the web interface via HTTPS.
3.1. `vi /conf/config.xml`
3.2. Search `<crt>` and yank the body of the tag.
3.3. Open a new file in ~~vim~~ vi and save the body there, example `/tmp/cert.b64`.
3.4. base64 decode the certificate: `b64decode -r /tmp/cert.b64 > /tmp/cert.pem`.
3.5. Use OpenSSL to show the fingerprint: `openssl x509 -in /tmp/cert.pem -noout -fingerprint -sha256`
4. Connect to the web interface and verify the fingerprint.
5. Initial setup, new hostname. New internal CA and server certificate.
5.1. In `/system_certmanager.php` the "show certificate info" is useful to show the certificate before activating it. The issue is that it only shows the `-text` but at that point you can download the `.pem` and do the `openssl x509 -in ${cert_pem} -noout -fingerprint -sha256` locally.
5.2. Activate the new certificate with correct DNS name and verify the new fingerprint.
6. Activate SSH via web interface, setup SSH public key.
6.1. Login to console directly again to show the SSH host key fingerprint with: `ssh-keygen -l -f /conf/sshd/ssh_host_ecdsa_key.pub`
6.2. Compare on initial ssh connection attempt.

Concluding from that, the following improvements could be made to make this easier:

- [ ] Introduce a new option in `src/sbin/opnsense-shell` to show all fingerprints of X.509 certificates and SSH host keys (3 algos by default as of 18.1.6). This would ideally be implemented as a PHP script so that the `config_read_array` function can be used. PR: ~~https://github.com/opnsense/core/pull/2476~~, https://github.com/opnsense/core/pull/2481
- [ ] Include output of `openssl x509 -in -noout -fingerprint -sha256` in `src/www/system_certmanager.php`. PR: https://github.com/opnsense/core/pull/2475

I could implement and test this if it sounds good to you?

[...]

PR #2475, opened 2018-06-22
Add `openssl x509 -fingerprint -sha256` to "Show certificate info"

ypid:
Nice thing is that `openssl x509` respects the order of supplied arguments. It was chosen to print the one line of fingerprint above of the X509 text dump. `-sha256` is needed because OpenSSL 1.0.2k-freebsd 26 Jan 2017 seems to default to SHA1 currently.

Example of the first line:

```
SHA256 Fingerprint=F0:E6:EB:31:E8:87:AF:52:16:4E:84:05:3B:6C:03:2C:C1:DF:5A:E7:36:F4:32:44:3B:B5:57:63:97:45:C3:77
```

URL path of the GUI page involved: /system_certmanager.php

### DIFF
--- a/src/www/system_certmanager.php
+++ b/src/www/system_certmanager.php
@@ -209,7 +209,7 @@ if ($_SERVER['REQUEST_METHOD'] === 'GET') {
     } elseif ($act == "info") {
       if (isset($id)) {
           // use openssl to dump cert in readable format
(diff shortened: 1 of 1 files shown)


fichtner:
Merged, thanks!

ypid:
You are welcome, thanks!

PR #2476, opened 2018-06-22
Add "Show host fingerprints" shell option: SSH host keys, X.509 cert

ypid:
Example output:

```
Enter an option: 14


Fingerprints of this host follow. Please compare them when connecting to this host to prevent MITM attacks.

Fingerprints of SSH host keys:
256 SHA256:fcMIAgT/vZR/TWP0j8AFROTNnudkU1tP9sRhbsIa8vM root@OPNsense.localdomain (ECDSA)
256 SHA256:lDenOc5wy2WU0e6sSz2hR9nEFnMqx5c3u1F/pHxgJlY root@OPNsense.localdomain (ED25519)
2048 SHA256:dsw9srlQHL0hPJlEdR9rL769N30BTZgXG9gXbdZGOkU root@OPNsense.localdomain (RSA)

Fingerprints of HTTPS X.509 certificate:
SHA256 Fingerprint=F0:E6:EB:31:E8:87:AF:52:16:4E:84:05:3B:6C:03:2C:C1:DF:5A:E7:36:F4:32:44:3B:B5:57:63:97:45:C3:77
SHA1 Fingerprint=2A:DE:8C:09:90:9E:76:6B:01:4D:05:7B:81:22:7E:B6:B6:0F:1A:86
```

Previously (#2427) I suggested to extract the X.509 certificate from the xml config but the difficult part for me who is not so familiar with the implementation of OPNsense is to find the certificate which is actually used by the local web server. I found that `/var/etc/cert.pem` is used in the configuration of the local web server and assume that this is the easier way to implement this in the expectation that the file name does not change without being also changed in this script and that the file exists. If it does not exist, OpenSSL would complain with a useful error message.

Tested on: OPNsense 18.1.10-amd64 (and forward ported it to master)
Alternative implementation to: https://github.com/opnsense/core/issues/2427

### DIFF
--- a/src/opnsense/scripts/shell/fingerprints.php
+++ b/src/opnsense/scripts/shell/fingerprints.php
@@ -1,0 +1,41 @@
+#!/usr/local/bin/php
+<?php
+
+/*
+ * Copyright (C) 2018 Robin Schneider <ypid@riseup.net>
+ * All rights reserved.
+ *
+ * Redistribution and use in source and binary forms, with or without
+ * modification, are permitted provided that the following conditions are met:
+ *
+ * 1. Redistributions of source code must retain the above copyright notice,
+ *    this list of conditions and the following disclaimer.
+ *
+ * 2. Redistributions in binary form must reproduce the above copyright
+ *    notice, this list of conditions and the following disclaimer in the
+ *    documentation and/or other materials provided with the distribution.
+ *
+ * THIS SOFTWARE IS PROVIDED ``AS IS'' AND ANY EXPRESS OR IMPLIED WARRANTIES,
+ * INCLUDING, BUT NOT LIMITED TO, THE IMPLIED WARRANTIES OF MERCHANTABILITY
+ * AND FITNESS FOR A PARTICULAR PURPOSE ARE DISCLAIMED. IN NO EVENT SHALL THE
+ * AUTHOR BE LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY,
+ * OR CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF
+ * SUBSTITUTE GOODS OR SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS
+ * INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN
+ * CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE)
+ * ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
+ * POSSIBILITY OF SUCH DAMAGE.
+ */
+
+echo "\nFingerprints of this host follow. Please compare them when connecting to this host to prevent MITM attacks.\n";
+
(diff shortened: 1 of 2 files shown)


fabianfrz:
@fichtner I think this fits better into the ssh banner since it would be shown before the user is logging into the system. That way, anyone could connect via SSH to see the certificates without login.

fichtner:
Yes, I concur. Good idea. The output should be short and precise, ideally one line per fingerprint and no additional blank lines.

We could use a separate script, but I’m unsure about the option 14) and the reformatting of the whole block. Not a lot of people will use this even though they should and all the available scripts in there now have large impact on reconfiguration of the system, but this does not.

ypid:
> Not a lot of people will use this even though they should

I like the idea to make the fingerprints easily visable also for that reason you mentioned. Maybe people will think about it "Why are those fingerprints shown there? They must be there for some reason …" :)

> I think this fits better into the ssh banner since it would be shown before the user is logging into the system. That way, anyone could connect via SSH to see the certificates without login.

With ssh banner I would associate this:

```
# cat /etc/motd
----------------------------------------------
|      Hello, this is OPNsense 18.1          |         @@@@@@@@@@@@@@@
|                                            |        @@@@         @@@@
| Website:      https://opnsense.org/        |         @@@\\\   ///@@@
| Handbook:     https://docs.opnsense.org/   |       ))))))))   ((((((((
| Forums:       https://forum.opnsense.org/  |         @@@///   \\\@@@
| Lists:        https://lists.opnsense.org/  |        @@@@         @@@@
| Code:         https://github.com/opnsense  |         @@@@@@@@@@@@@@@
----------------------------------------------
```

The issue here is that this is only shown after the login. When "/system_advanced_admin.php" -> "Password protect the console menu" is enabled in OPNsense, that is not shown before login.

The only thing that is shown is the summary of interfaces. The fingerprints could be cut down to 4 lines (3 ssh, 1 X.509) and shown after the interfaces. I implemented this in https://github.com/opnsense/core/pull/2481 which is alternative to this PR.

ypid:
> As an SSH banner, I understand something like the one is running on my server. In case of OPNsense it may include the fingerprints.

The ssh `Banner` option is not used by OPNsense but it could be. `/etc/motd` and `/usr/local/etc/rc.initial` are shown post ssh authentication. In either case, only the X.509 certificate fingerprint makes sense because at that point, we are already post ssh host key checking so SSH can handle host key updates itself.

fichtner:
SSH fingerprints are still useful in those cases even post login to ensure the fingerprints are correct after they changed (e.g. factory reset). It’s probably also time to fix the banner post login, but never found a good way to show a banner without showing it twice after console login but maybe it doesn’t matter that much anyway.

fichtner:
PS: messing with motd is possible, but I wouldn’t make it a priority as it does not block progress in any way.

fabianfrz:
@fichtner SSH can disable the motd to prevent showing it twice.

fabianfrz:
@ypid there is a difference. You have not authenticated at this step and you may detect a proxy because it does not rewrite the banner.

fichtner:
Hmm, I am unsure. The MOTD is not for administration purposes... we can't always show the banner because some accounts won't use it / have no use for it so this would be primarily a console thing for root anyway or when those SSH accounts log in as root either directly, or su or sudo (console menu).

fichtner:
@ypid shall we close this then?

ypid:
Works for me. Thanks for the discussion and merging the alternative PR!

PR #2477, opened 2018-06-23
Unset rcvar first to ensure any previous assignment of rcvar is unset…

ndejong:
… since it is being used in a global context not locally scoped within the rc_enabled() function

### DIFF
--- a/src/etc/rc.freebsd
+++ b/src/etc/rc.freebsd
@@ -41,6 +41,7 @@ rc_enabled()
 {
 	rc_filename=${1}
 	name=${2}
+	unset rcvar
 
 	# check if service has a name
 	if [ -z "${name}" ]; then


fichtner:
It looks good, but I am interested in which bug you have seen in relation to this?

ndejong:
Thanks - the issue turned up for me because I was debugging a malformed(?) rc.d script that has no `rcvar` line entry and uses a non-standard `rcvar` name by not having an `_enabled` suffix:-
https://github.com/digitalocean/freebsd-fix

Either way, the Digital Ocean FreeBSD "fix" needs to be re-crafted to suit the OPNsense startup tooling anyway.

src/etc/rc.freebsd
```diff
@@ -41,6 +41,7 @@ rc_enabled()
 {
 	rc_filename=${1}
 	name=${2}
+	unset rcvar
```
fichtner: understood, small request though: can we use this idiom instead?

rcvar=

ndejong: yes, that looks reasonable - I'm not going to be able to (easily) test it anymore though...

fichtner: I‘ll use your commit but edit as discussed. Will do testing, although mostly for regressions which are not expected either way :)

PR #2480, opened 2018-06-24
do not start lighttpd when nginx is installed

fabianfrz:
the nginx plugin takes over and handles those requests so the port can be reused for other things like a reverse proxy or serving local files.

### DIFF
--- a/src/etc/inc/plugins.inc.d/webgui.inc
+++ b/src/etc/inc/plugins.inc.d/webgui.inc
@@ -143,9 +143,6 @@ function webgui_configure_do($verbose = false, $interface = '')
         return;
     }
 
-    /* only stop the frontend when the generation was successful */
-    killbypid('/var/run/lighty-webConfigurator.pid', 'TERM', true);
-
     /* flush Phalcon volt templates */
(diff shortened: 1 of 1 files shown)


AdSchellevis:
@fabianfrz I think we should make this more generic, someone else might prefer apache which in this case requires our core package to be changed again for just disabling it. Cluttering dependencies is usually not a good idea.

AdSchellevis:
If I'm not mistaken, you should be able to add a similar construction like this https://github.com/opnsense/core/blob/18.1.9/src/etc/inc/plugins.inc.d/webgui.inc#L31-L38 for Nginx, which only leaves the minor problem of disabling the core webserver.

fabianfrz:
@AdSchellevis the problem is that nginx is started via RC and because lighttpd is started early, I would have to kill the process already running which means I would have to manipulate the RC file. I also do use the files generated here (certificate, dh parameters etc. so the user does not see any difference).

AdSchellevis:
I understand, but still don't think we should clutter these things together, so maybe we need to split some functionality first to offer a swap-in replacement, adding all kinds of hooks specific for software not installed is something we should try to avoid.

At the moment I don't have a lot of time available, so I can't help out with a better solution now, maybe after 18.7 comes out.

PR #2483, opened 2018-06-26
Change <select> class to "selectpicker".

kekek2:


AdSchellevis:
It's probably better to split the pull request, if there's javascript code in the page, there are scenario's where you can't just replace the type.

kekek2:
Ok.

PR #2490, opened 2018-06-27
monit: add dashboard widget

fbrendel:


fichtner:
Merged as discussed, thanks!

fichtner:
FYI, it works but keeps polling indefinitely without delay for me, monit isn't running and it shows zero results.

PR #2493, opened 2018-06-28
monit: fix widget issues

fbrendel:
* poll time is in seconds of course :innocent:
* polling was running twice
  possibly due to the `<script>` tag location within `<body>`
  once while parsing the `<body>` and once when DOM was ready


fichtner:
> poll time is in seconds of course

:D

fichtner:
Merged, thanks!

PR #2500, opened 2018-07-01
Typo in interfaces.php

marjohn56:
I think it should be dhcpdv6 not dhcpv4, :)


fichtner:
I edited a bit too much here, thanks. Though we should keep it in line with the menu entry, which is "DHCPv6".

marjohn56:
Done

fichtner:
Merged, thanks!

PR #2501, opened 2018-07-02
DHCPv6_Help

marjohn56:
Add help comment to DHCP6 for use with tracking interface.


fichtner:
Merging slightly modified version, thanks!

PR #2504, opened 2018-07-02
Delete Tooltip from Large CPU Widget

marjohn56:
Delete Tooltip from large CPU Widget


fichtner:
added manually in 4b89ead11, thanks!

PR #2513, opened 2018-07-04
Change <select> class to "selectpicker".

kekek2:


AdSchellevis:
@kekek2  please test these kind of changes before committing, the first file I'm trying to validate has style issues 

image
vs the original
image

Which isn't an improvement, eventually changed it to (https://github.com/opnsense/core/commit/8857213367d650fff8f5569ab514aaa59ef3d88b)
image

Which obviously cost more time then just do a search replace on the `<select>` tag

Next one on the list wash the phase1 page, which has items that are nicely hidden using javascript, which needs the proper changes to prevent the following from happening:

image

Please feel free to open another pull request if you're going to test the pages locally first, I'm going to close this one for now since it has too many issues.

PR #2519, opened 2018-07-06
Individual templates on deny ACL.

kekek2:


fabianfrz:
I like the idea to be more verbose about the reason why a page has been blocked but they may look more OPNsense branded than the squid standard templates.

AdSchellevis:
also better to use keep it flexible, either by allowing others to override (like https://github.com/opnsense/core/blob/master/src/opnsense/service/templates/OPNsense/Auth/motd.in) or add functionality for it using the gui.

AdSchellevis:
timeout

PR #2525, opened 2018-07-08
Use openssh_enabled() in src/opnsense/scripts/shell/banner.php

ypid:
This is only an improvement and unification of `src/opnsense/scripts/shell/banner.php`.

Using `openssh_enabled()` both times in this file is preferred over one time using `isset($config['system']['ssh']['enabled'])` and the other time using `openssh_enabled()`.

Updates: 00f9b21
Updates: 81e50ab
Updates: https://github.com/opnsense/core/pull/2481

### DIFF
--- a/src/opnsense/scripts/shell/banner.php
+++ b/src/opnsense/scripts/shell/banner.php
@@ -126,7 +126,7 @@ foreach ($iflist as $ifname => $friendly) {
 
 echo PHP_EOL;
 
-if (isset($config['system']['ssh']['enabled']) || $config['system']['webgui']['protocol'] == 'https') {
+if (openssh_enabled() || $config['system']['webgui']['protocol'] == 'https') {
     echo PHP_EOL;
 }
 


fichtner:
Nice catch, cherry-picked manually as 3f86721ce3.

I thought about moving both fingerprint bodies to their respective plugins.inc.d files alongside the rest of their code base?

ypid:
Thanks!

About `plugins.inc.d`: I am too unfamiliar with the structure but I would say that because I used `passthru`, the code is not really portable right now. Also this code is only used in one place right now.

fichtner:
Think of it like a class for each service... it's always cleaner to bring the private parts together. We could scrape the output and pass it on, maybe even in a more well-formatted way. But it's just an idea for a rainy day. :)

fichtner:
PS: Using file names only known to the service is something that is definitely "private" in nature.

ypid:
Thanks for the details!

PR #2530, opened 2018-07-09
Add a special case for /31 tunnel networks to the OpenVPN config generator (fixes #2529).

### LINKED ISSUE #2529: OpenVPN config generator is handling /31 tunnel networks incorrectly
As [RFC3021](https://tools.ietf.org/html/rfc3021) says, "In a point-to-point link with a 31-bit subnet mask, the two addresses above MUST be interpreted as host addresses." 
Right now, if you use a /31 network for the tunnel subnet, it ignores this special case, and treats the first address as a network address, thus running over the intended range, and producing non-RFC compliant and counterintuitive results.

Since OpenVPN interfaces are inherently point to point, and it makes perfect sense to use point to point network for them rather than waste /30's on them, this is quite annoying.

Reproducing steps:

1. Create an OpenVPN client interface
2. Enter a /31 prefix in the "Tunnel network" option, e.g. 10.0.0.4/31
3. Expected ifconfig statement as per RFC3021 and the "first address is the server address" logic is "ifconfig 10.0.0.5 10.0.0.4".
4. Look into the /var/etc/openvpn/client$num.conf file: the actual result is "ifconfig 10.0.0.6 10.0.0.5"

A pull request will follow shortly.

dmbaturin:
A fix for #2529 

I believe this should do the trick, though I have to admit I couldn't test it because I haven't figured out the procedure for testing it on a live system yet (editing the file in /usr/local/www and restarting services seems to have no effect) and I don't have a FreeBSD machine handy. If someone points me to a correct procedure for testing changes on a live system, I'll be grateful (if there's no such thing I can make a build setup of course).

Since I haven't contributed to OPNSense before, if I broke any process or coding conventions, just let me know and I'll fix it.

### DIFF
--- a/src/www/vpn_openvpn_export.php
+++ b/src/www/vpn_openvpn_export.php
@@ -701,11 +701,18 @@ function openvpn_client_export_sharedkey_config($srvid, $useaddr, $proxy, $zipco
         $conf .= openvpn_gen_routes($settings['local_networkv6'], 'ipv6');
     }
     if (!empty($settings['tunnel_network'])) {
-        list($ip, $mask) = explode('/', $settings['tunnel_network']);
-        $mask = gen_subnet_mask($mask);
+        list($ip, $prefix_length) = explode('/', $settings['tunnel_network']);
+        $mask = gen_subnet_mask($prefix_length);
         $baselong = ip2long32($ip) & ip2long($mask);
-        $ip1 = long2ip32($baselong + 1);
-        $ip2 = long2ip32($baselong + 2);
+
+        if ($prefix_length == "31") {
+            // As per RFC3021, in a /31 network, first address is the first host address
+            $ip1 = long2ip32($baselong);
+            $ip2 = long2ip32($baselong + 1);
+        } else {
+            $ip1 = long2ip32($baselong + 1);
+            $ip2 = long2ip32($baselong + 2);
+        }
         $conf .= "ifconfig $ip2 $ip1\n";
     }
     $conf .= "keepalive 10 60\n";


AdSchellevis:
@dmbaturin looks good, thanks!

fichtner:
@dmbaturin thank you ❤️ 

@AdSchellevis stable/18.1 or stable/18.7 ?

AdSchellevis:
@fichtner stable/18.1 should be good, it doesn't interfere with other setups.

dmbaturin:
@AdSchellevis, @fichtner  Could you point me to the procedure for live testing changes? I've found it quite puzzling that editing files in /usr/local/www didn't have any effect for me?

Also, since this is a behaviour change, I think it should be prominently mentioned in release notes. I've run into this bug when setting up a site to site connection to a VyOS router, but OPNSense to OPNsense connections with /31 tunnel networks probably have worked, until someone tried to make two tunnels using consecutive /31's.

fichtner:
ah, right, this is the export page. great, will do.

@dmbaturin you can apply your patch to a known install using:

    # opnsense-patch df338ae

Make sure you don't have "os-cache" installed, the plugin holds on to old code until you restart the web gui completely.

dmbaturin:
Interesting. Even after running opnsense-patch and restarting services, it still seems to have no effect. I wonder if I'm doing something wrong.

AdSchellevis:
@dmbaturin you know you changed the client export? this file has no relation with the service itself.

dmbaturin:
@AdSchellevis Hhm, wait. You mean this is not the place where openvpn config is generated from the client settings in the GUI? Please point me to the right place then, if I've taken it up, I guess I should complete it.

fichtner:
I was under the impression this was part of a field test prior to PR creation :D

see https://github.com/opnsense/core/blob/master/src/etc/inc/plugins.inc.d/openvpn.inc

dmbaturin:
@fichtner No, it's just my lack of familiarity with OPNSense source code. ;)

Any reason why the common functionality is not factored out of the plugin and client config export?

A more practical question, right now openvpn_get_interface_ip() takes the bit mask. Should I refactor it to take the prefix length string instead?

dmbaturin:
...or rather compare the bit mask integer with hex representation of /31?

fichtner:
> Any reason why the common functionality is not factored out of the plugin and client config export?

I'm not sure I understand this question. The plugin code should be all the functionality of a subsystem, but the client export code came from a package originally so the two do not align 100%. FWIW, most the code for operating an OpenVPN locally is in that plugin file with a few extra scripts next to it for this purpose or that.

> A more practical question, right now openvpn_get_interface_ip() takes the bit mask. Should I refactor it to take the prefix length string instead?

It's up to you. :)

dmbaturin:
Ok, I suppose I'll compare the bitmask, since it's the least intrusive thing to do.

I suppose I should make it a separate pull request since this one is already merged.

PR #2534, opened 2018-07-10
Add a special case for /31 networks to openvpn_get_interface_ip() (another instance of #2529)

dmbaturin:
Finally, I've fixed it in the place where I originally intended to fix it (#2529).

This time I've verified that it works, thanks to people on the IRC who explained me how to restart the web GUI properly.


fichtner:
Merged, thanks!

PR #2545, opened 2018-07-14
Nextcloud password: allow to set a separeate pw and allow no encryption

fabianfrz:
* allow to use no password (no encryption)
* allow to use a different pw

### DIFF
--- a/src/opnsense/mvc/app/library/OPNsense/Backup/Nextcloud.php
+++ b/src/opnsense/mvc/app/library/OPNsense/Backup/Nextcloud.php
@@ -75,6 +75,13 @@ class Nextcloud extends Base implements IBackupProvider
                 "value" => null
             ),
             array(
+                "name" => "password_encryption",
+                "type" => "password",
+                "label" => gettext("Encryption Password (Optional)"),
+                "help" => gettext("A password to encrypt your configuration"),
+                "value" => null
+            ),
+            array(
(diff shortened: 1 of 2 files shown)


src/opnsense/mvc/app/library/OPNsense/Backup/Nextcloud.php
```diff
@@ -128,13 +135,16 @@ function backup()
             $username = (string)$nextcloud->user;
             $password = (string)$nextcloud->password;
             $backupdir = (string)$nextcloud->backupdir;
+            $crypto_password = (string)$nextcloud->password_encryption;
             $hostname = $config->system->hostname . '.' .$config->system->domain;
-            $configname = 'config-' . $hostname . '-' .  date("Y-m-d_h:m:s") . '.xml';
+            $configname = 'config-' . $hostname . '-' .  date("Y-m-d_H:i:s") . '.xml';
```
fichtner: Or have we not fixed the time everywhere yet?

fabianfrz: my branch was probably too old - however git should not show it if it is equal on both branches...

fichtner: hm, no github is bugging out here in this PR making it useless for opnsense-patch

PR #2555, opened 2018-07-18
Opnsense_theme_patch

marjohn56:
Patch Opnsense theme for Sidebar

### DIFF
--- a/src/opnsense/www/themes/opnsense/assets/stylesheets/bootstrap/_responsive-utilities.scss
+++ b/src/opnsense/www/themes/opnsense/assets/stylesheets/bootstrap/_responsive-utilities.scss
@@ -127,6 +127,12 @@
   @include responsive-invisibility('.hidden-sm');
 }
 
+@media (max-width: 768px), (max-height: 669px) {
+  .toggle-sidebar {
+    display: none !important;
+  }
+}
+
 @media (min-width: $screen-md-min) and (max-width: $screen-md-max) {
   @include responsive-invisibility('.hidden-md');
 }
--- a/src/opnsense/www/themes/opnsense/assets/stylesheets/main.scss
+++ b/src/opnsense/www/themes/opnsense/assets/stylesheets/main.scss
@@ -4,6 +4,7 @@ $colors: (
   lightergrey: #FBFBFB,
   darkgrey: #3C3C3B,
   bordergrey: #E5E5E5,
+  black: #000000,
 );
 
 $zindex: (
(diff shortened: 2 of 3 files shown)


AdSchellevis:
thanks for the quick service Martin, I'm not sure I'm able to test and review this week, but will do so as fast as I can.

AdSchellevis:
Hi Martin,

I'll pull this one in, although there are still some issues with the new menu system, I expect the other theme's to suffer from similar things.

On my dev machine I seem to have two issues, a vertical line when there's a side menu collapsed and I can't seem to navigate outside the visible area (the menu doesn't scroll when collapsed).
image

So we shouldn't pull this in production yet, but the progress looks promising. 

Best regards,

Ad

marjohn56:
Indeed... I can see it on rebellion, but it's very dark so I hadn't noticed it... let me check Cicada and let René know.

AdSchellevis:
oops, while clicking some further, I noticed another glitch in the normal menu. 
image

The colouring changed and something is off with the current selected item.

marjohn56:
OK, well the color is sorted, now need to find the vertical bar issue...

opnsenseuser:
I know that the field is outside the viewboard, because no slide function is activated. I did not intentionally do this because it looks very strange. My approach to a solution would be to automatically collapse the collapsed menu into the visible area should the collapsed menu be outside the viewport. This should be possible with javascript. I have a link where someone has a similar solution. I would have imagined something like that. then you would not need a scrollbar. and it would also be compatible across browsers. here is the link: http://blog.moagrius.com/javascript/jquery-test-if-element-is-in-viewport-visible-on-screen/

AdSchellevis:
@opnsenseuser Maybe it's better to open an issue to keep track of some ideas to fix this, at the moment we're quite busy with our release.

marjohn56:
It appears to be OK in cicada and tukan, so I've missed something when I  updated rebellion and opnsense.

opnsenseuser:
@AdSchellevis  i will try to find a solution!
@marjohn56 no this can´t be. the main problem is in every template.

opnsenseuser:
@AdSchellevis and @marjohn56
I'm talking about this problem -> navigate outside the visible area (the menu doesn't scroll when collapsed).
This is the main problem. All other things work without any problem

@marjohn56 I think you missed something!

marjohn56:
Here's a snapshot from the current Tukan version and the 18.7.r152

image

opnsenseuser:
Are you using chrome? Did you clean the Browser Cache then close the Browser and try it again.

Regards,
Rene

Samstag, 28. Juli 2018, 21:43 +0200 von notifications@github.com  <notifications@github.com>:

marjohn56:
Yes I am using Chrome, no I did not clear the cache, I  switched to Tukan them from Opnsense. However I have now cleared the cache and reloaded.. I still have no vertical line  <https://github.com/notifications/beacon/ARIXpjrYDv8k-g-6Aid-sfHfmPNy27w2ks5uLMABgaJpZM4VULDy.gif>

opnsenseuser:
So they Must have changed something. 

Regards,
Rene

Samstag, 28. Juli 2018, 22:00 +0200 von notifications  <notifications@github.com>:

opnsenseuser:
What Vertical line do you mean?

Regards,
Rene

Samstag, 28. Juli 2018, 22:02 +0200 von rene@team-rebellion.net  <rene@team-rebellion.net>:

marjohn56:
<https://github.com/notifications/beacon/ARIXpoGwrli-74zUyaArcdGY85NpyFUNks5uLMOngaJpZM4VULDy.gif> The vertical line that Ad pointed out that started this thread. It’s not there in Tukan or Cicada, but is there in Rebellion and Opnsense.

PR #2558, opened 2018-07-18
After install, give basic tip that GUI IP can be set in console

stilez:
If the user's existing subnet isn't 192.168.1.1/xx, then telling the user to use this IP will fail. So tell the user how to set a different IP after reboot, rather than just telling them to use an IP that won't work for them.

Kept short this time - 1 line only :)

Also edited wording re certificate, to be clearer about this as well, by starting with what they will actually see (*"Your browser may..."* rather than *"You might need to..."*). Added just 1 line there as well.  

The wording *"and cannot be validated by an external root authority"* is so people don't worry about "WHY ISN'T IT TRUSTED, IS IT INSECURE!" - makes clear there isn't **external** validation. Also better understanding is good. Possible reword: "This is normal, as the initial certificate will be self-signed, and not signed by an external root authority."

Hopefully not overdoing it. (Life should be easy for newcomers, and install is where newcomers first see if a thing works for them!) Perhaps they can be squashed down to save text space.  If so please do.

### DIFF
--- a/src/sbin/opnsense-installer
+++ b/src/sbin/opnsense-installer
@@ -83,11 +83,13 @@ echo
 echo "The installation finished successfully."
 echo
 echo "After reboot, open a web browser and navigate to"
-echo "https://192.168.1.1 (or the LAN IP address)."
+echo "https://192.168.1.1 (or the LAN IP address). The console"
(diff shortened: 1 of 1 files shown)


fichtner:
Merged, thanks!

PR #2560, opened 2018-07-18
Plugins restrictions don't work for FTP.

kekek2:


AdSchellevis:
Incomplete fix, if we do want  to add functional constraints to ftp, we should look at the other acl parts as well (squid.acl.conf)

PR #2572, opened 2018-07-24
Plugin ProxyUserACL version 2

kekek2:
https://github.com/opnsense/plugins/pull/754


AdSchellevis:
@kekek2 how is this supposed to work for people using sslbump or any of the other functions you've removed here?

kekek2:
Now all this can be done in the updated user ACL plugin. The white list for SSL Bump is set here: https://github.com/opnsense/core/pull/2572/commits/485b2bd50b8a91c31d6d8ace2fbde19d3dc384c6#diff-014704489997936870d40579abee8cc0R229 and https://github.com/opnsense/plugins/pull/754/commits/a66359f45f192b4833bd86cb069aed0017745b67#diff-131b3ee7ed0a7faf5083ee2bc3663d8d. You can not throw the old code for the ACL, it will not interfere.

Issue #2573, opened 2018-07-25
IPv6 DUID-EN and others


marjohn56:
Reading a reddit conversation it appears IPv6 DUID-EN is needed. I'll take a look and get this done. I also want to add a function to quickly clear the existing DUID and force a new one.

cminnoy:
Hi, for my internet provided I also need to provide a DUID to get DHCPV6 to work with a fixed IP.
This used to work in a previous version of opnsense, but since a recent upgrade doesn't anymore.

What's strange is that the DUID accepted by opnsense always needs to have two more bytes (0e:00) before the rest of the identifier. The internet provider doesn't accept the first two bytes.
Always wondered why the difference?

That said, it used to work in a previous release, but in OPNsense 18.1.13_1-amd64 doesn't work anymore for me.

marjohn56:
Nothing was changed in 18.13 in the duid area. Set the dhcp6c debug flag and then look at the dhcp6c log entries. Post them here and I'll try and see what's wrong. The first two bytes are for use by dhcp6c itself only.

fichtner:
This has been merged into master.

PR #2574, opened 2018-07-25
Force DUID renew

marjohn56:
Clears existing DUID on save. No duid file forces dhcp6c to regenerate a new DUID.

This is in part a preparation for Issue #2573

### DIFF
--- a/src/www/interfaces.php
+++ b/src/www/interfaces.php
@@ -1124,6 +1124,13 @@ if ($_SERVER['REQUEST_METHOD'] === 'GET') {
                     $new_config['adv_dhcp6_config_advanced'] = $pconfig['adv_dhcp6_config_advanced'];
                     $new_config['adv_dhcp6_config_file_override'] = $pconfig['adv_dhcp6_config_file_override'];
                     $new_config['adv_dhcp6_config_file_override_path'] = $pconfig['adv_dhcp6_config_file_override_path'];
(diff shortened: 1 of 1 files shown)


fichtner:
looks more like a side-effect than a feature. we need to find a better way do to this :(

marjohn56:
Side-effect of what? There are several reports of users being unable to get a dhcp v6 address from their ISP, it seems that one of the reasons is the duid needs to be cleared and reset. Whichever way you do it, you need to delete the existing DUID file and configs if you want dhcp6c to generate a new one. You can of course edit the fixed DUID in Interface->Advanced, this is just  quicker and easier. There is no reason not to have more than one method of cracking an egg, and I'll still need this when I add the extra DUID types.

fichtner:
This is not a programatic solution. I would expect the checkbox to always purge the DUID at some point (e.g. on reboot). Like it is it will not be more practical than editing the DUID in the interface settings page. Worse still, the DUID is backed up and restored by the backup framework upon reboot so this needs further tinkering.

marjohn56:
It purges it on apply. I was under the impression that the backup file in conf, as well as the duid file itself and the config entry was all that was needed to clear the duid , is there somewhere else too?

Sitting outside a bar in Leeds so answering on my mobile.

marjohn56:
Closed in favour of #2602

PR #2580, opened 2018-07-28
Cause data through saveFormToEndpoint callback_ok to be passed through

ndejong:
When calling saveFormToEndpoint I'd like to be able to reference data in the front end that was returned through the ajax call - as in the example below

```javascript
$("#someAction").click(function(){
    saveFormToEndpoint(url="/api/plugin/feature",formid='frm_Form',callback_ok=function(data){
        $("#someElement").html(data['message']);
    });
});
```

The PR allows this data to be passed through as required.

I'd highlight however that I do not know the codebase well enough to know if this has other side effects or could be leveraged in ways to cause unwanted behaviors - while it seems probably okay I'd ask project maintainer(s) to provide guidance if required.

### DIFF
--- a/src/opnsense/www/js/opnsense_ui.js
+++ b/src/opnsense/www/js/opnsense_ui.js
@@ -90,7 +90,7 @@ function saveFormToEndpoint(url,formid,callback_ok, disable_dialog) {
                 }
             } else if ( callback_ok != undefined ) {
                 // execute callback function
-                callback_ok();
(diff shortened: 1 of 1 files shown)


AdSchellevis:
Hi @ndejong, 

This looks safe and indeed practical, I'll pull it in and check some of our use-cases.

Thanks!

Best regards,

Ad

PR #2586, opened 2018-07-30
radvd & dhcpd6 configure fix

marjohn56:
Safety net to configure dhcpd6 and radvd6

### DIFF
--- a/src/etc/inc/interfaces.inc
+++ b/src/etc/inc/interfaces.inc
@@ -2590,6 +2590,8 @@ function interface_track6_configure($interface = 'lan', $lancfg, $linkupevent = 
                  */
                 $parentrealif = get_real_interface($lancfg['track6-interface']);
                 killbypid("/var/run/dhcp6c_{$parentrealif}.pid", 'HUP');
+            }  else {
+                services_dhcpd_configure('inet6', array(),true);
             }
             break;
     }


src/etc/inc/interfaces.inc
```diff
@@ -2590,6 +2590,8 @@ function interface_track6_configure($interface = 'lan', $lancfg, $linkupevent =
                  */
                 $parentrealif = get_real_interface($lancfg['track6-interface']);
                 killbypid("/var/run/dhcp6c_{$parentrealif}.pid", 'HUP');
+            }  else {
+                services_dhcpd6_configure_only(array(),true);
```
fichtner: why not services_dhcpd_configure('inet6') ?

marjohn56: Because I did not want to affect anything else by calling all the other stuff in that call, just the two radvd and dhcpd6

fichtner: inet(4) is skipped, but you need the chroot fixup if dhcp wasn't started already

fichtner:
under which conditions is this reproducible? I need to inspect the call stack

src/etc/inc/interfaces.inc
marjohn56: OK, if you think it better... you da boss.

marjohn56:
My test system for this one is very basic. Clean install, upgrade and switch to dev.

OPNsense 18.7.r_127-amd64
FreeBSD 11.1-RELEASE-p11
OpenSSL 1.0.2o 27 Mar 2018

Slaved off my live unit.

dhcp6c settings are:

directly send solicit
debug is on
prevent release is on

apart from adding dns servers in general it's default.

fichtner:
services are not up after bootup ? radvd and dhcp6c or dhcpd?

marjohn56:
Oh, and it shows when the system boots. dhcpd6 service is showing not running, radvs shows running but the config is wrong,

fichtner:
thanks, how is dhcpd6 involved.... are you also tracking LAN from WAN on a slave unit?

marjohn56:
dhcpd6 server shows stopped, and will not start, config has no addresses. Not even sure it should be trying to start unless the override is selected, that may be another issue.

My live unit is all statics, the slave LAN is tracking the WAN.

fichtner:
Relevant topic: https://www.reddit.com/r/OPNsenseFirewall/comments/945z2j/ipv6_problems_after_migration_from_pf/

Number 3) is interesting, a /60 PD is actually sent as /62 ?

fichtner:
Slightly modified the version, no functional changes.

fichtner:
And: thanks! :)

marjohn56:
Issue 3  on reddit is sorted. See PR's also posted patch ID on reddit,

fichtner:
great, thank you ❤️

PR #2592, opened 2018-07-31
IPv6_DUID

marjohn56:
Part one of new feature. This is the GUI behind it. All that's left to do is generate the other types of DUID within opnsense, but pushing this out first for some feedback. It does work for the Auto setting.


marjohn56:
Closed in favour of completed PR #2602

PR #2602, opened 2018-08-04
Multiple Type DHCPv6 DUID support and generation.

marjohn56:
As it says, it's also possible to just use the  dhcp6c restart button just to re-fire that on its owm.

### DIFF
--- a/src/etc/inc/interfaces.inc
+++ b/src/etc/inc/interfaces.inc
@@ -2858,13 +2858,7 @@ function interface_dhcpv6_configure($interface = 'wan', $wancfg)
 
     /* write DUID if override was set */
     if (!empty($config['system']['ipv6duid'])) {
-        $temp = str_replace(':', '', $config['system']['ipv6duid']);
-        $duid_binstring = pack('H*', $temp);
-        $fd = fopen('/var/db/dhcp6c_duid', 'wb');
-        if ($fd) {
-            fwrite($fd, $duid_binstring);
-            fclose($fd);
-        }
+         write_ipv6_duid($config['system']['ipv6duid']);
     }
 
     $wanif = get_real_interface($interface, "inet6");
--- a/src/etc/inc/util.inc
+++ b/src/etc/inc/util.inc
@@ -1378,3 +1378,22 @@ function is_install_media()
 
     return true;
 }
+
+function write_ipv6_duid($duid_to_write)
+{
+    $values = explode(":", $duid_to_write);
+    $num_values = count($values);
+    $num_values = sprintf("%02x",$num_values);
+    array_unshift($values, $num_values, '00');
+    $duid_to_write = implode(':', $values);
+    log_error("write: {$duid_to_write}");
+    $temp = str_replace(':', '', $duid_to_write);
+    
+    $duid_binstring = pack('H*', $temp);
+  
+    $fd = fopen('/var/db/dhcp6c_duid', 'wb');
+    if ($fd) {
+        fwrite($fd, $duid_binstring);
+        fclose($fd);
+    }
+}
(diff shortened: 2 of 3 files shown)


marjohn56:
Just noticed there's a stray '$wanif = get_real_interface('wan', "inet6");' which needs removing. It's a leftover from a direction I was taking to get the MAC address, changed my mind and went a different route.

marjohn56:
Because the length of the DUIDs can be up to 128 chars, the manual entry "input" needs to be changed to a textarea, I can get that working, what I am having issues with is updating the textarea on the 'onclick' event in help field.

fichtner:
Works from here if I manually change the &lt;input/> to &lt;textarea>&lt;/textarea>. Maybe you want to push the "faulty" change for closer inspection.

marjohn56:
OK, gimme two.

marjohn56:
Line 466.. This is obviously wrong as it does not work. I've tried various ways of doing it but no joy. Do I need to create a function?

fichtner:
don't delete the attribute id= and it will work

marjohn56:
example por favour.. :)

fichtner:
huh :D like this...

    <textarea name="ipv6duid" id="ipv6duid" rows="2" ><?=$pconfig['ipv6duid'];?></textarea>

marjohn56:
Nay :) You're not reading what I have written. The textarea at 464 is fine, it's the onclick event at line 466, used to dynamically update the textarea with the read_duid() function. This will not work for example, although I read various web postings saying it should or should not.

`<a onclick="$('#ipv6duid').val('<?= read_duid() ?>');" href="#"><?=gettext("Insert the existing DUID here"); ?></a>`

marjohn56:
Still cannot get it to work...

fichtner:
```
--- /usr/local/www/system_advanced_network.php.orig	2018-08-05 14:10:20.911241000 +0200
+++ /usr/local/www/system_advanced_network.php	2018-08-05 14:11:32.016932000 +0200
@@ -461,9 +461,9 @@
               <tr>
                 <td><a id="help_for_persistent_duid" href="#" class="showhelp"><i class="fa fa-info-circle"></i></a> <?=gettext("Stored DHCP6 Unique Identifier"); ?></td>
                 <td>
-                  <textarea name="ipv6duid" rows="2" ><?=$pconfig['ipv6duid'];?></textarea>            
+                  <textarea name="ipv6duid" id="ipv6duid" rows="2" ><?=$pconfig['ipv6duid'];?></textarea>            
                   <div class="hidden" data-for="help_for_persistent_duid">
-                  <a onclick="$(pconfig['ipv6duid']('<?= read_duid() ?>');" href="#"><?=gettext("Insert the existing DUID here"); ?></a>
+                  <a onclick="$('#ipv6duid').val('<?= html_safe(read_duid()) ?>');" href="#"><?=gettext("Insert the existing DUID here"); ?></a>
                     <?= gettext('<br>This field can be used to enter an explicit DUID for use by IPv6 DHCP clients.<br>
                         The correct format for each DUID Type is as follows, all entries to be in hex format "xx" seperated by a colon.<br>
                         LLT: 4 bytes "00:01:00:01" followed by 4 bytes Unix time e.g. "00:01:02:03", followed by six bytes of the mac address.<br>
```

marjohn56:
And it doesn't work for me. :)

marjohn56:
Yes I was.. fixed... Ta muchly.

marjohn56:
bump.

fichtner:
@marjohn56 I spent 3 hours breaking this down due to workflow and style concerns. Ended up rewriting the proposed integration in what is about half of what was originally added in code. It's not perfect yet, but it's a step forward. It does take a lot longer to merge if hours have to be spent reviewing / reworking so I wanted to stress that point.

PR #2603, opened 2018-08-05
Tracking dhcpv6 dhcpd6 overide PD fix - 18.7 Stable patch only!!!

marjohn56:
Missing the overide section for PD!


marjohn56:
I've issued this PR on the stable branch so users can just patch it. I'll issue a PR for master in a few moments.

fichtner:
patching works regardless of branches, especially because both branches carry the same code

marjohn56:
When I created the patch using master there were other changes that appeared in the diff that did not apply to 18.7 stable, I was taking no chances. :)

PR #2604, opened 2018-08-05
Fix PD when using DHCPD6 override on tracked interface

marjohn56:
Missing PD override section

### DIFF
--- a/src/etc/inc/services.inc
+++ b/src/etc/inc/services.inc
@@ -1117,23 +1117,30 @@ function services_dhcpdv6_configure($blacklist = array(), $verbose = false)
                $dhcpdv6cfg[$ifname]['range']['from'] = make_ipv6_64_address($ifcfgipv6, $config['dhcpdv6'][$ifname]['range']['from']);
             }
             /* prefix length > 0? We can add dhcp6 prefix delegation server */
-            if ($pdlen > 2) {
-                $pdlenmax = $pdlen;
-                $pdlenhalf = $pdlenmax -1;
-                $pdlenmin = (64 - ceil($pdlenhalf / 4));
-                $dhcpdv6cfg[$ifname]['prefixrange'] = array();
-                $dhcpdv6cfg[$ifname]['prefixrange']['prefixlength'] = $pdlenmin;
+            if (!isset($config['interfaces'][$ifname]['dhcpd6track6allowoverride']) || !isset($config['dhcpdv6'][$ifname]['enable'])) {
+                if ($pdlen > 2) {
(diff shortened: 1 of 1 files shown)


fichtner:
Merged in a slightly modified way that merges the existing if-else with the newly added. Please double check  3a26aff :)

Thanks!

marjohn56:
Looks kosher...

marjohn56:
Going to have to change this a bit. It's forcing a prefix rather than just the range. We cannot force the prefix otherwise it defeats the whole object of making it dynamically aware of a PD change. This was a problem with pfSense, don't know if the fixed it on their end. I'll try and get it done for 18.1, certainly done for 18.2. In the meantime it works with 1ccafed. Note also posted on reddit thread.

fichtner:
You mean 1ccafed as well as what went in as d8c7a9b  or does the latter have a defect?

marjohn56:
No, the fix you did based on 1c*. It will hard code the prefix given by the user into dhcpd6.conf. we need the user just to enter the prefix range and length, i.e. ::20 > ::30 /60 and make up the rest for the dynamic info, same as with the /64 ranges.
null

marjohn56:
You are easily confuses. :) d8c7* will ALSO work in the meantime. De-confused?

fichtner:
Yes, kthxbai :)

marjohn56:
Hmm I looked that up... That's not very nice...

> kthxbai means “you're an asshole”. While the term literally parses out to Ok, thank you, goodbye, whoever is saying it to you does not mean any of that. It's a more firm way of saying “you're an asshole” than just a simple k, whose meaning is often ambiguous and subject to interpretation.

fichtner:
Not sure which dictionary you use.... I merely meant: Generally used [...] for hasty exit

https://www.urbandictionary.com/define.php?term=kthxbai

Sorry for the misunderstanding.

marjohn56:
:)
[https://www.google.co.uk/search?q=kthxbai&rlz=1C1CHBF_enGB763GB763&oq=kthxbai&aqs=chrome..69i57.552j0j8&sourceid=chrome&ie=UTF-8](url)

marjohn56:
Slight issue I have come across here. I can see no way at present of finding the prefix length that has been allocated to the LAN, i.e /48 /56 /60 so on. Do you know of a way? 
If not then I will need to modify dhcp6c to put that into an envar that gets dumped when the REASON is pushed out to the script.
It's needed as without it there is no way of knowing that the user has entered a valid prefix delegation size.
Otherwise we just have to leave it to the user to enter the data correctly with no checks.

PR #2605, opened 2018-08-05
blank problem advanced button

opnsenseuser:
since i used opnsense i noticed that the advanced slider button has an ugly blank on hover and focus. this problem occurs in every theme. has to do with a wrong placed blank.  Its also visible in the original theme of opnsense. I've fixed this with a standard html syntax. its visible on every page which uses an advanced button. like monitor, proxy and so on...see the screenshot

grafik

grafik

grafik


fichtner:
@opnsenseuser does 7a5f1fef6ecead work for you in the same way?

Issue #2583, opened 2018-07-30
UI - Firewall/Rules representation/rendering 


commonism:
I'd love to improve the representation of the rules in OPNSense to make a pfSense migration possible.

This is some screenshots of the rules view on a 2560x display using full/half/even less of the screen:

![opn_rules_2551](https://user-images.githubusercontent.com/164513/43380852-e680bdd0-93d2-11e8-9fd1-e2ecb5932de1.png)
![opn_rules_1914](https://user-images.githubusercontent.com/164513/43380838-df3f7f5c-93d2-11e8-83ed-76a96ddd7102.png)
At 1400x it starts to stack the icons:
![opn_rules_1400](https://user-images.githubusercontent.com/164513/43380845-e3335192-93d2-11e8-9531-41fcbafb2280.png)

To improve the usability of this very important view, there are things I'd like to change:
  * if the category changes, the category of a rule is shown above the rule
  * category filter box has the categories *sorted* by name and only the categories of the current interface
  * description is a row below the rule
  * remove Gateway column
  * remove Schedule column
  * add options column for things like interface(s), direction, Gateway, Schedule
  * rules with things not shown have an "advanced icon"
  * all clickable things for a rule are in the same spot
  * disabled rules font is grey/italic for the whole row

Similar to:
![example](https://user-images.githubusercontent.com/164513/43382777-2e6f18e8-93d9-11e8-8814-e404e1a1c11a.png)

If we could have the category to be coloured (attribute on the XML Element is an backwards compatible change) - even better.

As you may guess I really miss separators and as you propose to use category instead, lets make it a working replacement.

In case you have a large test config using aliases, floating rules, multiple interfaces, categories and all the other nice things, I could render it as an example. Small configs do not work, as there is very little gain cleaning up the rendering for them.

AdSchellevis:
@commonism there's more work needed in the firewall area to eventually drop the legacy code and move to new api enabled mvc code.  I the meantime it's no problem to improve the visibility as long as the changes are manageable (small steps, easy to review). I'm not sure about the proposed layout, it should eventually fit with the rest of the system in look in feel for as far as possible.

fabianfrz:
@AdSchellevis it would be also possible to provide an API via the legacy file which would make the new components working. When the backend is ready, only the API endpoints need to be moved.

AdSchellevis:
[...]

PR #2611, opened 2018-08-07
Firewall Rules - visibility adjustments

commonism:
partially addressing #2583

### DIFF
--- a/src/www/firewall_rules.php
+++ b/src/www/firewall_rules.php
@@ -480,6 +480,7 @@ $( document ).ready(function() {
                 $interface_has_rules = false;
 				$category='';
 				$interface='';
+				$offset = 0;
                 foreach ($a_filter as $i => $filterent):
                 if (
                     (!isset($filterent['floating']) && $selected_if == $filterent['interface']) ||
@@ -490,8 +491,10 @@ $( document ).ready(function() {
                 ):
 				  if( $selected_if != $interface )
                   {
+					$offset = $i;
                     $interface = $selected_if;
                   }
+                  $idx = $i-$offset;
 
 
                   if ($filterent['category'] !== $category) {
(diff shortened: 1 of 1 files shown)


src/www/firewall_rules.php
```diff
@@ -367,8 +376,8 @@
                       <th class="hidden-xs hidden-sm"><?=gettext("Port");?></th>
                       <th class="hidden-xs hidden-sm"><?=gettext("Destination");?></th>
                       <th class="hidden-xs hidden-sm"><?=gettext("Port");?></th>
-                      <th class="hidden-xs hidden-sm"><?=gettext("Gateway");?></th>
-                      <th class="hidden-xs hidden-sm"><?=gettext("Schedule");?></th>
+                      <th class="hidden"><?=gettext("Gateway");?></th>
```
AdSchellevis: I don't think we should hide gateway (and schedule) columns without them being visible on the overview page completely, policy based routing mistakes tend to cause quite some issues in peoples setups.

src/www/firewall_rules.php
```diff
@@ -477,6 +489,28 @@
                         $selected_if == 'FloatingRules'
                      )
                 ):
+				  if( $selected_if != $interface )
+                  {
+					$offset = $i;
+                    $interface = $selected_if;
+                  }
+                  $idx = $i-$offset;
```
AdSchellevis: I'm not sure what this is supposed to be doing, if we only want to add a number to the rule, this feels a bit too much.

src/www/firewall_rules.php
```diff
@@ -477,6 +489,28 @@
...
+                  if ($filterent['category'] !== $category) {
+                    $category = $filterent['category']
+?>
+
+                  <tr class="category text-uppercase text-danger text-center font-weight-bold" data-category="<?=!empty($filterent['category']) ? $filterent['category'] : "";?>">
+                    <td colspan=2></td>
+                    <td colspan=3><?=$category?></td>
+                    <td></td>
+                    <td></td>
+                  </tr>
```
AdSchellevis: Optically this doesn't look really nice on my end, I've attached a screenshot of what it looks like now
image

src/www/firewall_rules.php
commonism: Just trying to serve the 80% here.
A single column wich can be used to indicate 'special' configuration options like Gateway, Schedule, statetype, .... maybe?

src/www/firewall_rules.php
commonism: Numbering rules per interface helps addressing the *same* rule in a discussion.
I'm not aware of an easier way to get the number.
The rules data structure is not per interface, therefore I choose to store the interfaces offset.

src/www/firewall_rules.php
commonism: I think rendering issue is caused by an empty category for rule 3?

fichtner:
this needs to be redone due to conflicts

PR #2619, opened 2018-08-09
Fix disabled Reject action icon display code

framer99:
I posted about this on the forum at opnsense.org and ended up finding and fixing the problem.
[https://forum.opnsense.org/index.php?topic=9373.0](https://forum.opnsense.org/index.php?topic=9373.0)

This is tested to work on 18.7.


fichtner:
Merged, thanks!

PR #2620, opened 2018-08-10
Auto Prefix Delegation tracking.

marjohn56:
This adds the ability of prefix delegation to follow the change of the tracked interface.

Also added some changes to services_shcpv6.php in terms of help info and to limit the number of chars a user can enter when entering a value for PD ranges when tracking overide is in place. These changes reqiure a modified dhcp6c client that now gives out the prefix range and mask as an env var. This is picked up in interfaces.inc  and echoed to a file in /tmp.

Added correction for override ignoring additional options apart from range, such as dns servers etc as PR #2633


marjohn56:
Closed in favour of #2702

PR #2621, opened 2018-08-10
Update php.ini include paths

machadovilaca:


fichtner:
Committed slightly modified, we don't want /usr/local/share to override our own files 6655ee9

Thanks!

PR #2627, opened 2018-08-11
Swap stub-zone for forward-zone

johnkeates:
stub-zone doesn't work for non-authorative upstream DNS servers, but forward-zone does.

### DIFF
--- a/src/etc/inc/plugins.inc.d/unbound.inc
+++ b/src/etc/inc/plugins.inc.d/unbound.inc
@@ -486,12 +486,12 @@ function unbound_add_domain_overrides($pvt = false)
                 $domain_entries .= "local-zone: \"$domain\" typetransparent\n";
             }
         } else {
-            $domain_entries .= "stub-zone:\n";
+            $domain_entries .= "forward-zone:\n";
             $domain_entries .= "\tname: \"$domain\"\n";
             foreach ($ips as $ip) {
-                $domain_entries .= "\tstub-addr: $ip\n";
+                $domain_entries .= "\tforward-addr: $ip\n";
             }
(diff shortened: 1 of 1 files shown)


src/etc/inc/plugins.inc.d/unbound.inc
```diff
@@ -486,12 +486,12 @@ function unbound_add_domain_overrides($pvt = false)
...
         } else {
-            $domain_entries .= "stub-zone:\n";
+            $domain_entries .= "forward-zone:\n";
             $domain_entries .= "\tname: \"$domain\"\n";
             foreach ($ips as $ip) {
-                $domain_entries .= "\tstub-addr: $ip\n";
+                $domain_entries .= "\tforward-addr: $ip\n";
             }
-            $domain_entries .= "\tstub-prime: no\n";
+            //$domain_entries .= "\tstub-prime: no\n";
```
fichtner: it's better to remove stale code. we always have git history :)

fichtner:
Merged, thanks!

src/etc/inc/plugins.inc.d/unbound.inc
johnkeates: That is very true.

johnkeates:
This one: https://forum.opnsense.org/index.php?topic=9279.0

fichtner:
There are plans to move to MVC, unbound and dnsmasq are good targets. We could do configd service and template regeneration without going MVC as an interim step where service rework is slightly easier than the template rework.

OTOH, it’s easy to maintain as is and works reliably. Changing it will cause a regression or two for sure and both are to be considered critical infrastructure.

I can help guide / review this work if you want to submit it.

Is that what you were asking? 😊

PR #2630, opened 2018-08-12
fix table headers: tr elements cannot have a colspan, th elements can

fabianfrz:
fix for an UI bug found while developing the nginx 1.1 plugin. Due to the really long headings, they would break in the first column because the colspan is set incorrectly.


fichtner:
oO nice catch!

PR #2633, opened 2018-08-13
DNS servers set in DHCPDv6 config not being written to dhcpd6.conf

marjohn56:
Confirmed forum user report that dns entries made in dhcpd6 setting were not being written to dhcpd6.conf.

### DIFF
--- a/src/etc/inc/services.inc
+++ b/src/etc/inc/services.inc
@@ -1091,6 +1091,15 @@ function services_dhcpdv6_configure($blacklist = array(), $verbose = false)
         if (isset($blacklist[$ifname])) {
             continue;
         }
+        if(!empty($config['dhcpdv6'][$ifname]['dnsserver'][0])){
+            // user specified DNS servers
+            $dns_arrv6 = array(); //reset
+            foreach($config['dhcpdv6'][$ifname]['dnsserver'] as $dnsserver) {
+                if (is_ipaddrv6($dnsserver)) {
+                    $dns_arrv6[] = $dnsserver;
+                }
+            }            
+        }
         if (!empty($config['interfaces'][$ifname]['track6-interface'])) {
             $realif = get_real_interface($ifname, "inet6");
             $ifcfgipv6 = get_interface_ipv6($ifname);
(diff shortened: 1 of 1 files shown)


fichtner:
to me it seems dns server overrides work -- what am I missing?

the other change I'm half understanding although the code is pretty hard to read in general. in the old auto mode we clear all these things but with the manual override we want to keep all the settings and rewrite the range? that seems to make the prefixrange switcheroo completely unnecessary? :)

fichtner:
Ah I get the DNS issue now. I think sidestepping the config clear should fix DNS for the manual override case. For auto it should be cleared as it was. Does that make sense? :)

marjohn56:
Yes..:) I'm off to bed, got an early start in the morning.

marjohn56:
User reports issue has returned in 18.7.2

marjohn56:
They didn't work. my patch fixed it, however the re-work of that patch ignored the fact that the user entered DNS servers never get entered in my patch it was this:

if(!empty($config['dhcpdv6'][$ifname]['dnsserver'][0])){
            // user specified DNS servers
            $dns_arrv6 = array(); //reset
            foreach($config['dhcpdv6'][$ifname]['dnsserver'] as $dnsserver) {
                if (is_ipaddrv6($dnsserver)) {
                    $dns_arrv6[] = $dnsserver;
                }
            }            
        }

fichtner:
This patch chunk has a reverse effect on the code as it clears the system servers and makes them unusable. ;(

Let's go back one step.

Servers are handed here:

https://github.com/opnsense/core/blob/master/src/etc/inc/services.inc#L1237

Which didn't work because of: 

https://github.com/opnsense/core/pull/2633/files#diff-23f22aca2e953811c28d5b034d367737L1104

It cleared the dhcpv6 configured servers. But you fixed that. So that's all we have to do.

I challenge you to try to emit system-entered dns servers with your original patch in place. it will not work anymore.

Everything is fine here and you did good work.

Can we go please back to being productive?

fichtner:
Or maybe it's just duplicated code. I could be wrong. The code base is hard to review and coerce to do the right thing all the time. Sorry.

marjohn56:
I'll look at it again. The user reports not working, I cannot quickly test it as my test unit relies on #2702. That's not working now either so I need to fix that again before I can test this again.

fichtner:
Ok, different proposal: I'll add a patch for this for 18.7.3 and test it again to be sure we fix this. Ok?

There is a lot of changes going on in master which may break larger commits. I'm just trying to clean up the code base and try to understand what's going on before adding new things.

marjohn56:
I've got #2702 working again, but there are changes around that area I have not yet got my head around. What I do know is that if I go with the default master code in my test unit then my PD override settings do not work. Remember when I said to you in London the issue with knowing the real length of the prefix delegation, that's why I modified dhcp6c, looking again this is not required as the PD length has to be what was requested in the WAN config. I'll re-work this now, will be much simpler.

fichtner:
Ok, thanks, highly appreciated. You mean calculate_ipv6_delegation_length() will have your answer?

I've been looking at this and the function's return value can be misleading so it was changed to return "-1" if no delegation can be used, that's why there are some changes in the code to cope with this.

PR #2635, opened 2018-08-14
don't wrap first column (date/time) of log viewer

ghostwheel42:
the date/time column of the log viewer wraps when viewing logs with
very long lines (for example when viewing the acme client log)
this adds white-space: nowrap to the first column


fichtner:
it would be better to directly inject the style into the required column. this PR is missing the SCSS anyway.

fichtner:
Not opposed to this but I'd like to point out that this will further diminish the readability on mobile devices in portrait mode.

ghostwheel42:
sorry, I didn't see the scss part.
styling the first column this way uses fewer bytes than injecting the style into each first cell of each row - but it's gzipped anyway, so I think it wouldn't matter.
on my tests firefox android ignores the nobr in portrait mode.

fichtner:
except this will put the burden on all other theme maintainers to carry the fix too.

ghostwheel42:
@fichtner you are right - doing it in the first cell is more manageable

fichtner:
okay, minimal width mode disables text-nowrap so we're good here :)

PR #2636, opened 2018-08-14
Fix for advanced dhcp6 config

marjohn56:
Section of code I've removed had been removed previously as during recent PR's the ability to iterate through the interfaces was added. This 'solo' bit of code appears to have made its way back in, don't know where and don't know when as the song goes.


fichtner:
The block is skipped if "Prefix Interface Site-Level Aggregation Identifier" is set to "" instead of "0" contrary to what is stated in https://docs.opnsense.org/manual/how-tos/orange_fr_fttp.html -- effectively it tries to use the prefix twice for WAN and LAN.

marjohn56:
Removing the block stops it either way, you cannot have the wan there.

marjohn56:
That block was there from the original ORIGINAL code, way back, before we added the ability for it to iterate through the interfaces.

fichtner:
it obsoletes adv_dhcp6_prefix_interface_statement_sla_id option from the GUI, so either this PR is incomplete or not relevant because a change in the documentation does what you want. Either way, the documentation needs to be updated...

marjohn56:
Yes, you are correct. Give me 15 minutes to re-work this. 

Strange I tested it and the dhcp6c_*.config looked fine and worked.

marjohn56:
Try that now...

fichtner:
I don't understand. adv_dhcp6_prefix_interface_statement_sla_id is not embedded in the config. adv_dhcp6_prefix_interface_statement_sla_id is for the WAN. If you leave it out, it won't generate that block you get rid of by force, leaving the GUI to keep asking for adv_dhcp6_prefix_interface_statement_sla_id even though it's never written.

marjohn56:
Oh !**$$, what's up with my brain this morning.... hold on.

marjohn56:
OK... I think I see the error of my ways...

I think 'adv_dhcp6_prefix_interface_statement_sla_id' is is obsolete. The prefix ID statement on the tracking interface sets it, I cannot see why we need it at all. Can you think of a scenario where it would be needed?

fichtner:
It is used slurp a separate prefix for the WAN as well... at least that is what the original code seemed to intend. Happy to remove it, unless it seems marginally useful in which case we'd need to update a bit on online help / direct users not to input an ID here if they don't know what they are doing.

marjohn56:
What... a prefix delegation on the WAN? When would you ever use that?

fichtner:
Unsure and somewhat indifferent to the idea. :)

marjohn56:
Let's bin it then.. If someone comes back and asks for it we'll cross that bridge then. The thing is, with that bit of code in you get this in dhcp6c.conf and dhcp6c reports an error.

`nterface igb1_vlan832 {
  send ia-pd 0;
  send raw-option 6 00:0b:00:11:00:17:00:18;
  send raw-option 15 00:2b:46:53:56:44:53:4c:5f:6c:69:76:65:62:6f:78:2e:49:6e:74:65:72:6e:65:74:2e:73:6f:66:74:61:74:68:6f:6d:65:2e:6c:69:76:65:62:6f:78:33;
  send raw-option 16 00:00:04:0e:00:05:73:61:67:65:6d;
  send raw-option 11 00:00:00:00:00:00:00:00:00:00:00:66:74:69:2f:7a:37:78:76:77:68:77;
  script "/var/etc/dhcp6c_wan_script.sh";
};
id-assoc pd 0 {
  prefix-interface igb1_vlan832 {
    sla-id 0;
    sla-len 8;
  };
  prefix-interface igb0 {
    sla-id 0;
    sla-len 8;
  };
};
nterface igb1_vlan832 {
  send ia-pd 0;
  send raw-option 6 00:0b:00:11:00:17:00:18;
  send raw-option 15 00:2b:46:53:56:44:53:4c:5f:6c:69:76:65:62:6f:78:2e:49:6e:74:65:72:6e:65:74:2e:73:6f:66:74:61:74:68:6f:6d:65:2e:6c:69:76:65:62:6f:78:33;
  send raw-option 16 00:00:04:0e:00:05:73:61:67:65:6d;
  send raw-option 11 00:00:00:00:00:00:00:00:00:00:00:66:74:69:2f:7a:37:78:76:77:68:77;
  script "/var/etc/dhcp6c_wan_script.sh";
};
id-assoc pd 0 {
  prefix-interface igb1_vlan832 {
    sla-id 0;
    sla-len 8;
  };
  prefix-interface igb0 {
    sla-id 0;
    sla-len 8;
  };
};
`

marjohn56:
So after messing around with it it's pointless. The sla-length is used though. So 'Prefix Interface Site-Level Aggregation Identifier' it needs removing from interfaces.php too.

fichtner:
ok, agreed. let's ditch it then 👍

marjohn56:
OK, I've done a doc update, so what do you want to do, just use the doc update for now until I do multi-wan dhcp6c or remove it now?

PR #2644, opened 2018-08-17
Added probe for ZFS pool on partition in opnsense-importer (from Smart-Soft)

alexpro:
Added probe for ZFS pool on partition.

I would like to suggest adding the ability to import a configuration from a pool with an arbitrary name.
How it works:
User selects the device, and probe_for_part() determines partition. If probe_zfs_pool() finds ZFS pool on that partition, configuration importing is performed from ZFS pool.

1. import all possible pools without mounting any filesystems
2. enumerate pools and search for the one that uses selected partition
3. export all imported pools except root mounted pool, if exists

Now user must specify exactly "zroot" for import configuration from ZFS, i.e., ZFS supported only on "zroot/ROOT/default"

### DIFF
--- a/src/sbin/opnsense-importer
+++ b/src/sbin/opnsense-importer
@@ -123,16 +123,19 @@ zfs_load()
 probe_for_part()
 {
 	local DEV=${1}
+	local zcnt=
 
+	zcnt=$(echo ${POOLS} | grep -wc ${DEV})
 	if [ -e "/dev/${DEV}s1a" ]; then
 		# MBR UFS
 		export PART="/dev/${DEV}s1a"
-		probe_zfs_pool
 		return 0
 	elif [ -e "/dev/${DEV}p3" ]; then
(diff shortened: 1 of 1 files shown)


alexpro:
@fichtner 
Yes, I know the way. And proposed function 'probe_zfs_pool()' does it.
I use 'zpool get -H cachefile' command, it lists all imported pools, one per line.

fichtner:
@alexpro good :) it should list all pools here so the users knows what options there are to import from https://github.com/opnsense/core/blob/master/src/sbin/opnsense-importer#L181

alexpro:
I reworked function for ZFS pools probing. Now all available ZFS pools added to the tail of DEVS variable. Any ZFS pool now can be used same as any other device, from command line argument or can be specified manually when prompted.

fichtner:
Merged, thanks!

PR #2648, opened 2018-08-18
WIP: Add ignore_hotplug_events interface option

nallar:
Adds an option to ignore hotplug events for an interface. Currently we also always ignore hotplug events for interfaces with both static v4 and v6 addresses set, this option does not change that behaviour.

OpenWRT has a similar setting called "force link" which ignores hotplug events as well as doing some other things.

iOpenWRT settings screenshot showing force link

Currently adding it under the block bogons option.

image

### DIFF
--- a/src/etc/rc.linkup
+++ b/src/etc/rc.linkup
@@ -40,6 +40,12 @@ function handle_argument_group($iface, $argument2)
     global $config;
 
     if (isset($config['interfaces'][$iface])) {
+        if (isset($config['interfaces'][$iface]['ignore_hotplug_events'])) {
+            $friendly = convert_friendly_interface_to_friendly_descr($iface);
+            log_error("Hotplug event {$argument2} detected for {$friendly}({$iface}) but ignoring as ignore_hotplug_events is set");
+            return;
+        }
+
         // set defaults
         $ipaddr = '0.0.0.0';
         $ip6addr = '::';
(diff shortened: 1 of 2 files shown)


nallar:
Alternate approach here: e040d0044d853cd763886

src/etc/rc.linkup
```diff
@@ -40,6 +40,12 @@ function handle_argument_group($iface, $argument2)
     global $config;
 
     if (isset($config['interfaces'][$iface])) {
+        if (isset($config['interfaces'][$iface]['ignore_hotplug_events'])) {
```
AdSchellevis: better use` !empty()`

src/www/interfaces.php
```diff
@@ -359,6 +359,7 @@ function get_wireless_channel_info($interface) {
     $pconfig['lock'] = isset($a_interfaces[$if]['lock']);
     $pconfig['blockpriv'] = isset($a_interfaces[$if]['blockpriv']);
     $pconfig['blockbogons'] = isset($a_interfaces[$if]['blockbogons']);
+    $pconfig['ignore_hotplug_events'] = isset($a_interfaces[$if]['ignore_hotplug_events']);
```
AdSchellevis: same here

nallar: The lines around this for the existing checkbox options use isset - happy to change but does that mean the existing code is wrong?

AdSchellevis: @nallar I see, you're right, it writes a new array structure instead of using the existing. Better to keep it aligned with the rest, although I would replace the first occurrence in rc.linkup, when set with 0 we should also consider the item false.

nallar: Ok, will do.

nallar:
Been running without trouble after working around opnsense/plugins/issues/800 by disabling dynamic DNS, so I'm closing this merge request as I don't need this change.

PR #2649, opened 2018-08-19
Add ability for user to hide interfaces in widgets

marjohn56:
Request for users to be able to hide selected interfaces from Interfaces & Intefaces Statistics widdgets. Added new flag to config Interface - 'hide'


marjohn56:
In the Interfaces widget that is the case, in the statistics widget.
In the statistics widget that's using a bit of java in the <script>, and I have not got my head around those java bits... yet. Any clues?

fichtner:
Focus on the easier widget first. We can figure out / replace the JS one later.

marjohn56:
The first one is only a change to the widget, and that works fine.

fichtner:
But it needs its own widget setting. System log widget has an example.

marjohn56:
OK, so do not use the 'Hide in widgets' in the interfaces.php?

fichtner:
Yes, because it’s not an operational setting.

marjohn56:
OK, first widget done, allows the user to enter the interface name() to hide in the same way as services.

marjohn56:
OK, all working, changes to widgets only.

marjohn56:
Bump ..

@fichtner - Don't forget these, modified as you requested. :)

src/www/widgets/widgets/gateways.widget.php
```diff
@@ -25,71 +25,108 @@
...
+          <td>
+            <input id="submitd" name="submitd" type="submit" class="btn btn-primary" value="<?=gettext("Save");?>" />
+          </td>
+        </tr>
+      </tbody>
+    </table>
+  </form>
+</div>
+
+<div id="gateways-widgets" class="content-box";">
```
fichtner: there's an error here, but let me merge manually in a bit...

fichtner:
2/3 widgets changed as the code needed a larger dust off in general.. one larger bug in the new layout, but you'll like it when it's ready hopefully tomorrow :)

AdSchellevis:
@fichtner I'll put it on my list :)

marjohn56:
I tried select picker for about 40 minutes, gave up and went for the simple method.

fichtner:
The browser standard works ok which is what I've put in, but the selectpicker is certainly an improvement over it. ;)

fichtner:
BTW, thanks for the submissions!

fabianfrz:
@marjohn56 The reason why [lodash](https://lodash.com/docs/4.17.10) is one of my preferred libraries is that it has all functions you need for working with objects and collections. In your case it it probably filter in combination with includes. This is a backbone model but it is more or less the same:
https://github.com/opnsense/plugins/blob/master/www/nginx/src/opnsense/www/js/nginx/src/models/LogLinesCollection.js#L14 (please be careful because this is ES6, so some features may not work without transpiling)

marjohn56:
Very nice, but I know not what you want me to do with it. 🤔

marjohn56:
@fichtner - It works, but to be honest I prefer the "type in" method. At least I could quickly see what I was hiding. With the current method if you have a lot of interfaces it's a pain to just hide one.

fichtner:
Let's invert the selection then... it's just a line of code. The UX problem is still there either way, now if you have a long list and only need one interface listed that's the same amount of clicking. It would maybe be more useful if we had an invert selection checkbox?

marjohn56:
Ooh too many choices... :)


Sent from my Samsung Galaxy smartphone.
null

fichtner:
I'll take that as a "definitely maybe". :)

marjohn56:
or even "maybe definitely"...

Shall we have a vote? 🤪

fichtner:
This as an Oasis reference. I'll add the checkbox tomorrow and set the default to invert.

marjohn56:
Perfect... :)

Issue #2650, opened 2018-08-19
Feature request:  DHCP Dynamic DNS Key Algorithm


itn3rd77:
Hi,

after struggeling with DHCP Dynamic DNS updates I found out that OPNsense is only supporting the key algorithm ```hmac-md5```. Therfore I would like to request adding a drop-down box allowing to select the key algorithm  ```hmac-md5``` or ```hmac-sha512```.

Thanks for the greate work!

Regards
Ingo

itn3rd77:
Closed. Merged with #2651

PR #2651, opened 2018-08-20
Add DHCP Dynamic DNS key algorithm drop-down list

itn3rd77:
I tried my luck and opened my first pull request. Please let me know if anything further has do be done. Here are some open points I couldn't figure out:

- Do I need to add translations for the description of the drop-down list?
- Do I need to handle dhcpv6 as well?

This pull request tries to solve my feature request #2650.

Regards
Ingo

### DIFF
--- a/src/etc/inc/services.inc
+++ b/src/etc/inc/services.inc
@@ -950,7 +950,7 @@ EOD;
                 $newzone['dns-servers'] = array($dhcpifconf['ddnsdomainprimary']);
                 $newzone['ddnsdomainkeyname'] = $dhcpifconf['ddnsdomainkeyname'];
                 $newzone['ddnsdomainkey'] = $dhcpifconf['ddnsdomainkey'];
-                $newzone['ddnsdomainalgorithm'] = $dhcpifconf['ddnsdomainalgorithm'];
+                $newzone['ddnsdomainalgorithm'] = !empty($dhcpifconf['ddnsdomainalgorithm']) ? $dhcpifconf['ddnsdomainalgorithm'] : "hmac-md5";
                 $ddns_zones[] = $newzone;
             }
         }
--- a/src/etc/inc/services.inc
+++ b/src/etc/inc/services.inc
@@ -950,6 +950,7 @@ EOD;
                 $newzone['dns-servers'] = array($dhcpifconf['ddnsdomainprimary']);
                 $newzone['ddnsdomainkeyname'] = $dhcpifconf['ddnsdomainkeyname'];
                 $newzone['ddnsdomainkey'] = $dhcpifconf['ddnsdomainkey'];
+                $newzone['ddnsdomainalgorithm'] = $dhcpifconf['ddnsdomainalgorithm'];
                 $ddns_zones[] = $newzone;
             }
         }
@@ -1012,7 +1013,7 @@ function services_dhcpd_zones($ddns_zones)
                     if (!empty($zone['ddnsdomainkeyname']) && !empty($zone['ddnsdomainkey'])) {
                         if (!in_array($zone['ddnsdomainkeyname'], $added_keys)) {
                             $dhcpdconf .= "key {$zone['ddnsdomainkeyname']} {\n";
-                            $dhcpdconf .= "  algorithm hmac-md5;\n";
+                            $dhcpdconf .= "  algorithm {$zone['ddnsdomainalgorithm']};\n";
                             $dhcpdconf .= "  secret {$zone['ddnsdomainkey']};\n";
                             $dhcpdconf .= "}\n";
                             $added_keys[] = $zone['ddnsdomainkeyname'];
--- a/src/www/services_dhcp.php
+++ b/src/www/services_dhcp.php
@@ -894,10 +894,6 @@ include("head.inc");
                             if ($pconfig['ddnsdomainalgorithm'] == $algorithm) {
                               $selected = "selected=\"selected\"";
                             }
-                          } else {
-                            if ($algorithm == "hmac-md5") {
-                              $selected = "selected=\"selected\"";
-                            }
                           }?>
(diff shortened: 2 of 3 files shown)


src/etc/inc/services.inc
```diff
@@ -950,6 +950,7 @@ EOD;
                 $newzone['dns-servers'] = array($dhcpifconf['ddnsdomainprimary']);
                 $newzone['ddnsdomainkeyname'] = $dhcpifconf['ddnsdomainkeyname'];
                 $newzone['ddnsdomainkey'] = $dhcpifconf['ddnsdomainkey'];
+                $newzone['ddnsdomainalgorithm'] = $dhcpifconf['ddnsdomainalgorithm'];
```
AdSchellevis: @itn3rd77 you need to set the default, like:

`$newzone['ddnsdomainalgorithm'] = !empty($dhcpifconf['ddnsdomainalgorithm']) ? $dhcpifconf['ddnsdomainalgorithm'] : "hmac-md5";
`

Otherwise you will break running configs.

AdSchellevis:
Hi Ingo,

At a first glance it looks good, minus some minor remarks. 
I'm just not 100% sure about the dhcpd edit for single hosts, did you test the outcome when configuring different values for the static dhcp items as well?

IPv6 seems to have the same construction, are you planning to add this option there as well?

Best regards,

Ad

src/www/services_dhcp_edit.php
```diff
@@ -493,6 +493,24 @@ function show_tftp_config() {
...
+<?php
+                        foreach (array("hmac-md5", "hmac-sha512") as $algorithm) :
+                          $selected = "";
+                          if (! empty($pconfig['ddnsdomainalgorithm'])) {
+                            if ($pconfig['ddnsdomainalgorithm'] == $algorithm) {
+                              $selected = "selected=\"selected\"";
+                            }
+                          } else {
+                            if ($algorithm == "hmac-md5") {
+                              $selected = "selected=\"selected\"";
```
fichtner: First algo in the list is selected by default by the browser which makes this superfluous.

src/www/services_dhcp.php
```diff
@@ -885,6 +885,24 @@ function removeRow() {
...
+                        foreach (array("hmac-md5", "hmac-sha512") as $algorithm) :
+                          $selected = "";
+                          if (! empty($pconfig['ddnsdomainalgorithm'])) {
+                            if ($pconfig['ddnsdomainalgorithm'] == $algorithm) {
+                              $selected = "selected=\"selected\"";
+                            }
+                          } else {
+                            if ($algorithm == "hmac-md5") {
+                              $selected = "selected=\"selected\"";
+                            }
```
fichtner: Same here

src/etc/inc/services.inc
```diff
@@ -1012,7 +1013,7 @@ function services_dhcpd_zones($ddns_zones)
                     if (!empty($zone['ddnsdomainkeyname']) && !empty($zone['ddnsdomainkey'])) {
                         if (!in_array($zone['ddnsdomainkeyname'], $added_keys)) {
                             $dhcpdconf .= "key {$zone['ddnsdomainkeyname']} {\n";
-                            $dhcpdconf .= "  algorithm hmac-md5;\n";
+                            $dhcpdconf .= "  algorithm {$zone['ddnsdomainalgorithm']};\n";
```
fichtner: Could it be there is another stray hmac-md5 in this file that needs replacement?

src/etc/inc/services.inc
itn3rd77: fixed. thanks for catching.

src/etc/inc/services.inc
itn3rd77: mmmh didn't find any more occurrences of ```hmac-md5```. have you any concrete line for a blind man?

fichtner: Can’t check without a laptop. It just seems odd there are two select boxes added but only one authorative line changed. I think Ad mentioned the same thing. Maybe it’s ok as is and one of the selects doesn’t do anything?

itn3rd77:
@AdSchellevis 

> I'm just not 100% sure about the dhcpd edit for single hosts, did you test the outcome when configuring different values for the static dhcp items as well?

To be honest I have not tested dhcpd edit for single hosts. After taking a short test it seems the fields like ```ddnsdomainkeyname``` or ```ddnsdomainkey``` are not taken into account when edit a single dhcp host.

This is what I found in ```etc/inc/services.inc```

```
...
                if (isset($sm['ddnsupdate'])) {
                    if (($sm['ddnsdomain'] <> "") && ($sm['ddnsdomain'] != $dhcpifconf['ddnsdomain'])) {
                        $smdnscfg .= "    ddns-domainname \"{$sm['ddnsdomain']}\";\n";
                    }
                    $smdnscfg .= "    ddns-update-style interim;\n";
                }
...
```

and this is what is written to ```/var/dhcpd/etc/dhcpd.conf``` when adding a DHCP Static host ...

```
host s_lan_0 {
  hardware ethernet 00:11:32:54:56:33;
  fixed-address 192.168.100.199;
  option host-name "test1";
}
```

In my humble opinion most fields in ```services_dhcp_edit.php``` are superfluous and not taken into account. Is this right?

AdSchellevis:
@itn3rd77 I've cleaned up a lot of these fields earlier, but it's easy to overlook. In case these fields seem to be unused, we should cleanup these area's. If you leave the `services_dhcp_edit.php` as it was, I'll merge you're patch and drop the unused properties.

itn3rd77:
@AdSchellevis I wasn't able to revert the commits regarding ``services_dhcp_edit.php``, so I changed the content back as it was. Hope that is ok.

AdSchellevis:
@itn3rd77 the effect is the same :) thanks

PR #2653, opened 2018-08-20
fix adding interfaces LAN bug of German interface

tk-wfischer:
The XML variable name ($newifname) should always be "lan" (written in lowercase letters). This commit removes the gettext translation function for the variable name. With the former behaviour, removing the LAN interface and trying to add new (optional) interfaces was broken when a user had switched the interface language to German (the variable name was translated to "LAN" in capital letters, which caused any new interface to overwrite this "LAN" entry).


AdSchellevis:
@tk-wfischer thanks!

PR #2656, opened 2018-08-20
Translate console menu.

kekek2:


kekek2:
https://github.com/opnsense/lang/pull/41

fabianfrz:
@fichtner didn't we agree not to translate things on the console because of encoding issues?

fichtner:
With the gettext in shell scripts figured out and FreeBSD finally making use of proper UTF-8 that may be feasible, although I doubt this will be added everywhere. A Brazilian user asked for installer language support but that also relies on serial console behaving normally so I’m not overly enthusiastic.

fabianfrz:
@fichtner usually the serial connection with Linux machines works flawlessly because they usually use UTF-8 in the terminals.

nhirokinet:
Even if UTF-8 is treated CJK texts are unreadable in OPNsense VGA, or many other terminal places, because of lack of fonts. Even in 2019 this is not uncommon in Japan.
Adding font packs for all available languages is one option, but I strongly vote for keeping terminal English.

There are two same files. Upper one is ssh to OPNsense in WSL mintty and below one is the OPNsense in VMware ESXi VGA.
default

kekek2:
screenshot
Need to use vt-driver.

nhirokinet:
In the instance it was enabled (as default), still "これは日本語で書かれたメッセージです。" is shown like above.

nhirokinet:
This is because of font data, and probably hw.vga.textmode setting.

I read https://www.next-hop.net/~hiraga/FreeBSD/japanese-vt.shtml (Japanese page, sorry, but please see only code block) and tried

```
hw.vga.textmode = 0
```
to /boot/loader.conf, reboot, and

```
fetch http://www.wheel.gr.jp/~dai/fonts/unifont-8.0.01.fnt
sudo vidcontrol -f unifont-8.0.01.fnt
```

Then Japanese texts are displayed in GNU Unifont.

default

Currently all of supported languages are LTR, so installing GNU Unifont seems to do a job for internationalization of console CUI. However, GNU Unifont is licensed under GPLv2, so you may want to find some alternative.

http://unifoundry.com/unifont/index.html

fichtner:
As much as this may be convenient, we can't do this and should keep it to english on non-GUI elements. It does not affect most of the system and NLS port options are turned off from even before we forked.

Issue #2664, opened 2018-08-22
Accessing the UUID of a config item in templates


ndejong:
I started a discussion in the forum [with the same subject](https://forum.opnsense.org/index.php?topic=9489.0) but I think I need to move this one here to discuss the code in more detail.

My goal is to implement a [template helper](https://github.com/opnsense/core/blob/master/src/opnsense/service/modules/addons/template_helpers.py) function called `getUUIDs()` to facilitate being able to write the following template style:-
```
{% for tunnel_uuid in helpers.getUUIDs('VerbNetworks.autossh.tunnels.tunnel') %}
{% set tunnel = helpers.getUUID(tunnel_uuid) %}
Host {{ tunnel_uuid }}
  User {{ tunnel.user }}
  HostName {{ tunnel.host }}
{% endfor %}
```
Following the code, I observe that the special `__uuid_tags__` attribute that is populated via [this code ](https://github.com/opnsense/core/blob/master/src/opnsense/service/modules/config.py#L76)
```
self.__uuid_tags[item.attrib['uuid']] = item.tag
```
... and this code provides data in the helper that looks like this
```
  "__uuid_tags__": {
    "8a777695-656c-497e-ba31-3e9301edb15f": "alert",
    "4a0bab44-cd2d-485a-8d07-75967c8e521f": "key",
    "c467f1c4-375b-4be1-987a-e52a07dd6de1": "tunnel",
...
```
The "thing" about this is that the tag-name value (ie, "alert", "key", "tunnel") only provide the last part of the item tag-path which means it is impossible to select all UUIDs that belong to a given full-tag without the risk of collecting tags from different plugins that happen to have the same name.

Is there actually any straight-forward way to resolve this?  I imagine that the flow on effects of expanding the tag value to include the full tag-path would create a whole set of new issues elsewhere.
```
  "__uuid_tags__": {
    "8a777695-656c-497e-ba31-3e9301edb15f": "OPNsense.monit.general.alert",
    "4a0bab44-cd2d-485a-8d07-75967c8e521f": "VerbNetworks.autossh.keys.key",
    "c467f1c4-375b-4be1-987a-e52a07dd6de1": "VerbNetworks.autossh.tunnels.tunnel",
...
```

@AdSchellevis - I'm figuring this question is for you.

AdSchellevis:
@ndejong I think we should just add the attributes to the items, so you could point to something like `{{tunnel.__uuid__}}` I'll take a look at this when I can.

[...]

PR #2665, opened 2018-08-22
Implements accessing the UUIDs of a config items by tag reference

ndejong:
Raised as issue https://github.com/opnsense/core/issues/2664
Discussed in [forum here](https://forum.opnsense.org/index.php?topic=9489.0)


ndejong:
Even tried creating a new branch to prevent those previous commits getting tangled up with this PR here - oh, git...

src/opnsense/service/modules/addons/template_helpers.py
```diff
@@ -111,6 +111,18 @@ def getUUID(self, uuid):
...
+        """ retrieve a list of uuids from the item list returned by tag
+        :param tag: tag in dot notation (section.item)
+        :return: []
+        """
+        uuids = []
+        for item in self.toList(tag):
+            for uuid in self._template_in_data['__uuid__']:
+                if self._template_in_data['__uuid__'][uuid] == item:
+                    uuids.append(uuid)
+        return uuids
```
fabianfrz: should be done for a single element

ndejong:
This updated diff makes it more explicit about there the `uuid` string is coming from - also adds a `break` once the `uuid` is found since we don't need to keep searching and can skip to the next

AdSchellevis:
@ndejong if we add the attributes to the object, would that be sufficient for you? I'm not sure we should add an extractor for all uuids if most use-cases need the uuid of the item traversing.

ndejong:
@AdSchellevis  - I'm probably happy with any solution that is clear and easy to use.

If step back for a moment and consider why I wrote an extractor - As a relative newbie in the OPNsense codebase I felt like I wanted a solution that made the templates easier to write and understand.  To get there I first dug through other templates and then through code to discover the data-structures inside the template helpers at which point it did not occur to me that I could simply use the item object itself as the comparator as @fabianfrz pointed out in the Nginx [work around](https://github.com/opnsense/plugins/blob/94a55d5e61668ed13cc4cc245911294d9dadcfc2/www/nginx/src/opnsense/service/templates/OPNsense/Nginx/http.conf#L138-L144) - my extractor uses the same approach.

The idea of injecting a special `__uuid__` attribute into each item certainly works, perhaps the comment I'd make is that, again as a newbie it took me until I discovered where the template helper was and then finding a way to observe the data-structures inside that the `__uuid_tags__` and `__uuid__` elements became apparent - there was all a bit magic  :)  and so I wonder if the per item `__uuid__` attribute will be too much magic and/or get confused with the existing `__uuid__` element that feels like it could have been called `__uuids__` (with the "s")

I'm perhaps biased here, but I feel like the explicit function that does what it says `getUUIDs()` makes it easier for the next plugin developer that comes stumbling along.

Totally your call.

fabianfrz:
@AdSchellevis the uuid as a property would be exactly what I need. be careful because you may overwrite data. In PHP i get it as [(at) attibutes]["uuid"]. The advantage is, that an @ cannot be a valid XML tag.

AdSchellevis:
@fabianfrz I know, As soon as I have a proper fix, I'll post it in the issue https://github.com/opnsense/core/issues/2664 and close this one when done.

AdSchellevis:
@ndejong @fabianfrz this https://github.com/opnsense/core/commit/3add6c7cc9adea860061a8849286047cb388e777 should do the trick. 

To use a uuid, just use something similar to:

```
  {{item['@uuid']}}
```

ndejong:
Most excellent indeed

ndejong:
@AdSchellevis yes please do

PR #2669, opened 2018-08-24
dhcpd: treat EFI BC the same as EFI x86-64

andi-makandra:
Several vendors use EFI BC instead of the EFI x86-64 value for the arch, some thinkpads for example. We noticed this while we tried to use PXE on all our machines and that some EFI systems with 64-bit didn't work and some did. We then found the old commit that change the if clause from 00:07 to 00:09 and tested around. Thus we found out that some machines use 7 and some 9, so IMHO it's correct to redirect both cases to the 64bit syslinux filename. This commit doesn't break old configs, just adds the additional option.

Would be nice to see that in the next update so we can start using PXE :) Thanks


AdSchellevis:
@andi-makandra thanks, looks good to me, we'll add it to the next release

andi-makandra:
thanks for the quick merge :)

PR #2674, opened 2018-08-27
Copy pfsync0 MTU from real interface

Kali-:
in some situation, like OpenStack, mtu is less then default (1500) and
it cause replication issue
we have noticed on a low used firewall with 2/3k State table size more
then 100k State table size on the backup server

copy mtu from real interface fix this issue

### DIFF
--- a/src/etc/inc/interfaces.inc
+++ b/src/etc/inc/interfaces.inc
@@ -1525,6 +1525,8 @@ function interfaces_carp_setup($verbose = false)
 {
     global $config;
 
+    $intf_stats = legacy_interfaces_details();
+
     if ($verbose) {
         echo 'Configuring CARP settings...';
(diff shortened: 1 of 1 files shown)


fichtner:
Looks reasonable. I would split off the mtu statement as a second call so the carp init doesn’t fail if the mtu is wrong or empty.

fichtner:
If this is ok for @adschellevis I’ll merge next week with a few cleanups in this area it of this PR scope.

AdSchellevis:
@fichtner it looks good to me, maybe default to 1500 (or not setting the mtu) when the parent mtu can't be found for some odd reason.

fichtner:
Good idea, will do 😊

fichtner:
Merged, thanks!

PR #2679, opened 2018-08-29
Update system_hasync.php

### LINKED ISSUE #2487: "Disable preempt" in High Availability settings requires reboot to take effect without a hint about that
Since the setting only changes a sysctl I dont think a reboot would actually be required but even if it is better to reboot the machine there should be hint in the description about it.

Minor issue but it cost me some time yesterday, using 18.1.10

mimugmail:
Closes #2487


fichtner:
Merged, thanks!

Issue #2675, opened 2018-08-28
dhcpd6 not starting


marjohn56:
This is on the dev path, I've not checked production as I think we would have heard by now if the problem existed on that path.

I switched back from using dhcpd6 override to auto to check out that everything still works OK, and found that dhcpd6 is not starting. Tracked this down to newwanipv6 now has a check in it for $manual_dhcpv6. If this is false then services_dhcpd_configure('inet6') does not get called, result = no dhcpdv6. What was the intended purpose of this, as whether it is auto or manual dhcpd_configure('inet6') still needs calling.... or am  I missing something?

Commit ID: dcc4ec0 is the beast.

al-ec:
Experienced the same. This actually shuts down my DNS connectivity (I guess most my DNS traffic is ipv6?) and it makes me unable to reach any update repos for OPNsense/FBSD.

marjohn56:
Until @fichtner rejoins us just replace that check and block with this.

`foreach (link_interface_to_track6($interface, true) as $lanif => $lancfg) {
    services_dhcpd_configure('inet6');
}
`

fichtner:
What setting in dhcpv6 /ra mode do you use? I’m entirely sure what dhcpv6 does in tracking, but Martin‘s patch is probably what is needed.

It looks like an interface_bring_down() temporarily disables dhcpv6 with that „blacklist” but it doesn’t come up afterwards again.

I removed a side effect in 18.7 that would unconditionally reload dhcpv6 so that’s probably still what we’re chasing. At least it gives us the opportunity to reload it correctly now.

marjohn56:
Just had another look, we're not passing any interface to the services_dhcpd_configure() function so I cannot at present ( I've only been awake for a short while! ) see the reason for the 'for' loop.

fichtner:
The for loop was for figuring out the amount of manual dhcpv6 usage. If it was above zero we would start dhcpv6 once. 😉

But then we simply have to say: if (count(link_interface...)) {

fichtner:
Blacklist skips down interfaces which ends up stopping dhcpv6 because it has no more interfaces

marjohn56:
So really we need to check if the interface is up and act accordingly. So perhaps a check in the services_dhcpd_configure function itself is the way forward. Not sure, not wide awake. :)

fichtner:
[...]

PR #2681, opened 2018-08-30
fix for dhcpd6 not being configured.

marjohn56:
dhcpd6 needs configuring whether it's manual override or not. This patch checks to see if track6 is set and uses that as the conditional.


fichtner:
My review was stuck in "pending" mode. I've added a simpler fix as mentioned in #2675 via dc02745. Thanks!

PR #2687, opened 2018-09-04
core: fix configd asList helper

fabianfrz:
fixes the issue that if the element does not exist, I will get an array with a single element which is null element, which again breaks for loops

### DIFF
--- a/src/opnsense/service/modules/addons/template_helpers.py
+++ b/src/opnsense/service/modules/addons/template_helpers.py
@@ -78,6 +78,8 @@ class Helpers(object):
         :return: []
         """
         result = self.getNodeByTag(tag)
+        if result is None:
+            return []
         if type(result) != list:
             # wrap result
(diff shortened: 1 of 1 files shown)


fabianfrz:
@fichtner this one is important to merge because @mimugmail found a bug in the nginx plugin caused by this misbehaviour. This bug may break other config templates too (maybe not found yet)

AdSchellevis:
@fabianfrz thanks!

mimugmail:
Does this touch ModelRelationFields? I have some reports that Wireguard Peers doesnt work after 18.7.2 but Wireguard itself stays same version. No idea where to look else

fabianfrz:
no, this should only affect `ArrayField` where this helper is used.
You will now get an empty array instead of one containing an empty value usually known as `null`. With the change, you cannot run into a NPE so easily because the for loops work as expected.

Issue #1003, opened 2016-06-09
Various style errors on the iPhone 6s PLUS


DokuKaefer:
![firmware](https://cloud.githubusercontent.com/assets/11537653/15922410/cd817170-2e27-11e6-9fc9-aa05167a278e.png)

![usermanager](https://cloud.githubusercontent.com/assets/11537653/15922411/cd842c6c-2e27-11e6-912b-caacbb0dcef8.png)

![voucher](https://cloud.githubusercontent.com/assets/11537653/15922412/cd89cb72-2e27-11e6-95b8-bc06b6888fa7.png)

fichtner:
Hi Andreas,

Screenshot Nr.1: agreed, saw that with the dashboard too. Not very good and kind of addressed now. If there is too little space the elements "squash" together. There is nothing to be done other than removing this or that element for smaller screens in the long run. Lots of work, needs lots of help frankly...

Screenshot Nr.2: this can always happen, the ultimate price of a small screen.

Screenshot Nr.3: Agreed. Need to port MVC pages from tables to divs in order to make those dialogs better. That's for 17.1.

There is a patch, which I'm sure add will disagree with on a few levels, but for the sake of getting a discussion going I've added it to the -devel version.


Cheers,
Franco

fichtner:
Yes, so from what I can see bootstrap elements get width="100%", which does not properly inherit in tables from col-X-Y values. We mostly don't set them, but even when set it does not properly work.

fichtner:
I've reverted my table-layout style changes, this needs @AdSchellevis to save the day. ;)

fichtner:
Aware of this, but takes too much effort to fix table layouts instead of migrating all to div layout.

fichtner:
Hooray, this was very likely fixed by @NOYB via https://github.com/opnsense/core/commit/76aab5cb52a

PR #2696, opened 2018-09-07
Wrap the IP address

marjohn56:
On lobby with more than two columns, in the interface_list.widget the IPv6 address can be longer than the box. This wraps it.


fichtner:
The width specification is another fix. We need to add both, probably to all width: columns

marjohn56:
I  tried just using the width, but it's not reliable, especially if you go over three columns.

fichtner:
Because there's not enough space to begin with... https://github.com/opnsense/core/issues/1003 and other reports have indicated that text breaks out if not limited by width in some form or another. Adding text-wrap on top doesn't help?

marjohn56:
I'm a bit baffled now, I had this working, wrapping the IPv6 address nicely, now I've broken it. Not even my PR will bring it back....
duh... got it. This sorts out this widget, the others are not quite as friendly.

fichtner:
Ok, thanks, I know what do to.

Issue #2110, opened 2018-01-18
Feature request: use of suricata 'ignoring traffic' features


petererdos:
This would be the starting point:
https://suricata.readthedocs.io/en/latest/performance/ignoring-traffic.html

It would be very nice if these could be implemented in the GUI, at least capture filters, as they look like (fairly) easy to implement (hopefully they are) :)

The reason for this is that sometimes connections and services brake when using IDS + IPS. There are a few hosts which can be "whitelisted", eg. Windows Update servers (which many times fail without any apparent reasons while nothing is logged as blocked - hunting down the rule(s) causing this behavior is difficult).

It would be awesome (well, maybe useless, if not possible, in the case of WU) to use wildcard dns records somehow, as capture filters look like only support IPs and cidr net addresses. Maybe this can be worked around with aliases (?). If not, at least a wsus server (internal) could be whitelisted and clients could get the updates from there (this would be another workaround).

Many thanks!

AdSchellevis:
Pass rules should be quite simple to integrate into the "user rules" tab which is already there, although this traffic will still flow trough suricata. 

BPF filters would fully skip suricata, but I haven't tested it with netmap (IPS mode) yet, although it seems to be supported. 
The disadvantage is that according to the examples we need to specify the filter statically at startup, maybe there is a possibility somewhere to feed in a list of some sort (or a pipe we could feed dynamically), but I haven't looked into that.

There are quite some use cases to skip traffic, like skipping high traffic hosts like netflix, it would be good to investigate the options we have before adding knobs for a specific host.

fichtner:
Netmap is before BPF AFAIK so that wouldn't work.

AdSchellevis:
that was I was I thought too, but https://github.com/OISF/suricata/blob/master-4.0.x/src/runmode-netmap.c#L113 suggests these filters also apply on netmap. not sure what the gain is then.

mimugmail:
I'd take this the next weeks.
Proposal: make User rules a subtab and add subtab whitelist. Also space for future use with a new tab supporting user defined rules.

Suggestions?

AdSchellevis:
@mimugmail I'm not fully convinced we should use the bpf filters at the moment, the pass operation is simple, although wouldn't have the full gain of skipping traffic completely (easy to integrate, only add an option type). bpf filters need a full restart of suricata and might add more complexity then we would like to have. What did you want to build for this item? I think it's good to first discuss options together with measurements if needed before adding complexity here.

[...]

PR #2697, opened 2018-09-07
IDS, add whitelists

mimugmail:
As discussed in #2110 with @AdSchellevis here are whitelists.

I didn't realized it's that simple. Two new fields for source/destination. When empty it's any, so existing rules won't be changed. In dropdown there's now a "Pass" to add a whitelist.

Since we have in suricata.yaml:

```
action-order:
  - pass
  - drop
  - reject
  - alert
```

.. we don't run into problems when the SID is higher than the match. 

Here's my test rule:

`pass ip   10.10.10.1  any ->  192.168.55.5  any (msg:"test pass";  sid:4294967294; rev:1;)
`

If I have a FP on let's say some SIP attacks from a real phone, normally you'd just disable the rule to get it running. But it would also allow this kind of attack. Now you can just create a whitelist from your phone to your PBX and that's it.

So, what do you think?

### DIFF
--- a/src/opnsense/mvc/app/controllers/OPNsense/IDS/forms/dialogUserDefined.xml
+++ b/src/opnsense/mvc/app/controllers/OPNsense/IDS/forms/dialogUserDefined.xml
@@ -6,6 +6,18 @@
         <help>Enable this fingerprint rule.</help>
     </field>
     <field>
+        <id>rule.source</id>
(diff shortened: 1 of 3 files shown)


AdSchellevis:
@mimugmail easy indeed :) that's what I meant last week. I'll pull it in and give it another test round

PR #2700, opened 2018-09-08
ipsec: move styles to theme

### LINKED ISSUE #2695: IPSec widget not following css.
User points out that in the IPSec widget there are hard codec colours, in the <script> section.

fabianfrz:
move all styles to sass file and recompile the CSS

closes #2695


fabianfrz:
@AdSchellevis our them developers usually fix those issues in approximately a week to a maximum of two releases. For that time the patch can be hold back on master until the themes are ready but that is up to you.

AdSchellevis:
@fabianfrz there are way more theme's around then the ones publicly available, hence the major release switch. Other parties should have the opportunity to change their products in time.

fabianfrz:
Then I would suggest keeping it on master until the next major release, the public templates can update (it would not have any impact because the selectors / properties) would not change anything because the element style is higher weighted then the class selector.

marjohn56:
I've made no changes to Rebellion, but I now see this. On 18.7.2:
image

On 19.1.a_187 With @fabianfrz patch:

image

Don't think  I need to do anything.

src/opnsense/www/themes/opnsense/assets/stylesheets/main.scss
```diff
@@ -884,3 +884,20 @@ div[data-for*="help_for"] {
 .dashboard_grid_column table {
   table-layout: fixed;
 }
+
+$ipsec_bright_color: #EEEEEE;
+$ipsec_dark_color: #777777;
+
+#ipsec {
+  #ipsec-mobile, #ipsec-tunnel, #ipsec-Overview {
```
fichtner: I'd prefer this to be lowercase all the way

fabianfrz: me too but the ids aren't mine ;)

fichtner: but you touch the same lines so check mate! :P

marjohn56:
I'm just heading out of the door so manyana. :)

fichtner:
unfortunately this needs to be redone due to conflicts

fabianfrz:
@fichtner do I have to do that or the theme team?

fichtner:
good question. I'm not even sure what @AdSchellevis prefers as a solution here. the firewall live log has embedded classes now that can be safely overwritten...

AdSchellevis:
we probably better opt for a similar solution as the firewall live log, embed css, let the theme overwrite if needed. 

@fabianfrz do you have time to create a pull request for that?

fabianfrz:
@AdSchellevis done - see the new PR

PR #2702, opened 2018-09-08
DHCPD6 Auto Prefix change handling

marjohn56:
This Commit replaces #2620 which has so many conflicts now that it needed re-working. This commit still requires the updated [https://github.com/opnsense/dhcp6c/pull/4](url) - as it uses the new envvar created by the client to feedback the PD length to the GUI.

This commit is somewhat reduced in complexity as much of what was in the original commit has been adopted in one form or another.


marjohn56:
@fichtner - If this is going to be ignored then I'll not bother. I have modded this twice now along with #2633 and the bug is back. I've raised the issue. I'll leave it to you to fix.

marjohn56:
Closed.. re-working

PR #2705, opened 2018-09-08
hover for sidebar

opnsenseuser:
sorry it took so long, but it was quite a job for me!

I will customize the tukan, cicada and opnsense themes too and also upload. 
There are some program lines in the main.css of any themes that are no longer needed.
There are also some important fixes in tukan and cicada and also opnsense (font-size,.....)

That's why it's very important to me. 

Thank you very much Franco.

best regards rené

### DIFF
--- a/src/opnsense/www/js/opnsense_theme.js
+++ b/src/opnsense/www/js/opnsense_theme.js
@@ -31,6 +31,12 @@
 $(document).ready(function () {
     // traverse loaded css files
     var toggle_sidebar_loaded = false;
+	var nomouse = "mouseenter mouseleave";
+	var layer1_a = $('#navigation > div > nav > #mainmenu > div > a');
+	var layer1_div = $('#navigation > div > nav > #mainmenu > div > div');
+	var layer2_a = $('#navigation > div > nav > #mainmenu > div > div > a');
+	var layer2_div = $('#navigation > div > nav > #mainmenu > div > div > div');
+	
     $.each(document.styleSheets, function(sheetIndex, sheet) {
         if (sheet.href != undefined && sheet.href.match(/main\.css(\?v=\w+$)?/gm)) {
           $.each(sheet.cssRules || sheet.rules, function(ruleIndex, rule) {
@@ -40,7 +46,23 @@ $(document).ready(function () {
           });
         }
     });
-
+	/* disable mouseevents on toggle and resize */
+	function mouse_events_off() {
+		layer1_a.off(nomouse);
+		layer2_a.off(nomouse);
+		layer1_div.off(nomouse);
+		layer2_div.off(nomouse);
+								}
+	/* trigger mouseevents on startup */				
+	function trigger_sidebar() {
+		layer1_a.first().trigger("mouseenter");
+		layer1_a.first().trigger("mouseleave");
+		layer1_div.removeClass("in");
+								}	
+	/* transition duration - time */
+	function transition_duration(time) {
+		$.fn.collapse.Constructor.TRANSITION_DURATION = time;
+									}								
     function opnsense_sidebar_toggle(store) {
         $("#navigation").toggleClass("col-sidebar-left");
         $("main").toggleClass("col-sm-9 col-sm-push-3 col-lg-10 col-lg-push-2 col-lg-12");
@@ -48,8 +70,10 @@ $(document).ready(function () {
         if ($("#navigation").hasClass("col-sidebar-left")) {
             $(".brand-logo").css("display", "none");
             $(".brand-icon").css("display", "inline-block");
+			trigger_sidebar();
             if (store && window.sessionStorage) {
                 sessionStorage.setItem('toggle_sidebar_preset', 1);
+				transition_duration(0);
             }
         } else {
             $(".brand-icon").css("display", "none");
@@ -57,6 +81,8 @@ $(document).ready(function () {
             $("#navigation.page-side.col-xs-12.col-sm-3.col-lg-2.hidden-xs").css("width", "");
             if (store && window.sessionStorage) {
                 sessionStorage.setItem('toggle_sidebar_preset', 0);
+				mouse_events_off();
+				transition_duration(350);
             }
         }
     }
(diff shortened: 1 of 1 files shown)


src/opnsense/www/js/opnsense_theme.js
```diff
@@ -31,6 +31,12 @@
 $(document).ready(function () {
     // traverse loaded css files
     var toggle_sidebar_loaded = false;
+	var nomouse = "mouseenter mouseleave";
+	var layer1_a = $('#navigation > div > nav > #mainmenu > div > a');
+	var layer1_div = $('#navigation > div > nav > #mainmenu > div > div');
+	var layer2_a = $('#navigation > div > nav > #mainmenu > div > div > a');
```
fabianfrz: IDs are unique - the front part before can be omitted

src/opnsense/www/js/opnsense_theme.js
```diff
@@ -40,23 +46,43 @@ $(document).ready(function () {
...
         $("#navigation").toggleClass("col-sidebar-left");
         $("main").toggleClass("col-sm-9 col-sm-push-3 col-lg-10 col-lg-push-2 col-lg-12");
         $(".toggle-sidebar > i").toggleClass("fa-chevron-right fa-chevron-left");
         if ($("#navigation").hasClass("col-sidebar-left")) {
             $(".brand-logo").css("display", "none");
             $(".brand-icon").css("display", "inline-block");
+			trigger_sidebar();
             if (store && window.sessionStorage) {
                 sessionStorage.setItem('toggle_sidebar_preset', 1);
+				transition_duration(0);
```
fabianfrz: tabs / spaces

src/opnsense/www/js/opnsense_theme.js
```diff
@@ -40,23 +46,43 @@ $(document).ready(function () {
...
             }
         } else {
             $(".brand-icon").css("display", "none");
             $(".brand-logo").css("display", "inline-block");
             $("#navigation.page-side.col-xs-12.col-sm-3.col-lg-2.hidden-xs").css("width", "");
             if (store && window.sessionStorage) {
                 sessionStorage.setItem('toggle_sidebar_preset', 0);
+				mouse_events_off();
+				transition_duration(350);
```
fabianfrz: here too

src/opnsense/www/js/opnsense_theme.js
```diff
@@ -65,35 +91,64 @@ $(document).ready(function () {
         /* navigation toggle */
         $(".toggle-sidebar").click(function () {
             opnsense_sidebar_toggle(true);
-            $(this).blur();
+            $(this).blur();
```
fabianfrz: no whitespace after the semicolon

src/opnsense/www/js/opnsense_theme.js
```diff
@@ -65,35 +91,64 @@ $(document).ready(function () {
         /* navigation toggle */
         $(".toggle-sidebar").click(function () {
             opnsense_sidebar_toggle(true);
-            $(this).blur();
+            $(this).blur();			
         });
-
-        /* sidebar mouseevents */
-        $("#navigation").hover(function () {
+		/* sidebar mouseenter */
```
fabianfrz: again a tab issue

src/opnsense/www/js/opnsense_theme.js
```diff
@@ -65,35 +91,64 @@ $(document).ready(function () {
...
+																}	
+										});
+		/* sidebar mouseleave */								
+		$("#mainmenu").mouseleave(function () {
+			if ($("#navigation").hasClass("col-sidebar-left")) {
+				layer1_a.attr("aria-expanded","false");
+				layer1_a.next("div").removeClass("in");
+															}
+												});
+		/* on resize - toggle sidebar / main navigation */
```
fabianfrz: f*** this is messed up…

src/opnsense/www/js/opnsense_theme.js
```diff
@@ -65,35 +91,64 @@ $(document).ready(function () {
...
+					mouseenter: function() { 	$("#navigation.col-sidebar-left").css("width", "380px");
+												$(this).trigger("click");},
+					mouseleave: function() { 	$("#navigation.col-sidebar-left").css("width", "70px");},
+							});
+				
+				layer2_div.on({
+					mouseenter: function() { 	$("#navigation.col-sidebar-left").css("width", "380px");
+												$(this).trigger("click");},
+					mouseleave: function() { 	$("#navigation.col-sidebar-left").css("width", "70px");},
+							});
```
fabianfrz: Two issues:
* https://en.wikipedia.org/wiki/Don%27t_repeat_yourself
* this code layout is awful

opnsenseuser:
I am sorry that the source code is so bad. 
I really do not understand why that is and how I can avoid this in the future. 
what would happen if we take this 1:1 to the opnsense source with these errors?

can someone explain to me how I can determine this when editing in notepad ++?

fabianfrz:
I am an Arch Linux user - Notepad++ is Windows only. Somewhere in the editor you should have the following settings:
Tabs: set it to replace it by 4 Spaces
Encoding: UTF-8 (should be correctly detected)
Line Endings: Unix (LF, \n)

Kate as an example:
screenshot_20180908_232126

The issue can be avoided if you would set the default of your editor correctly.
The tab (\t) is a special control character which is replaced by the editor by an indent but our files use 4 spaces. If as you can see, this is machine dependent - so the position of the brackets is completely misplaced on another one or, as you can see, in the diff of GitHub.

What are the consequences?
* the code becomes unmaintainable (bad readable, in case of python also error prone)
* it makes the review hard
* may also make it harder to debug - for example to place breakpoints

opnsenseuser:
ok, thank you for explaining that to me. can you tell me which encoding should I set?

grafik

fabianfrz:
you are already using the correct one

opnsenseuser:
UTF-8 without BOM should be right?
So, the only problem are the tabs? I´m i right ?

fabianfrz:
yes

opnsenseuser:
I should get that right next time! 

And thanks again for the explanation. that is a really great help for me !!

opnsenseuser:
I would like to upload it again to see if it is at least a little better. I must study!! Thank you Fabian

PR #2708, opened 2018-09-09
Sidebar - hover modification

opnsenseuser:
Sidebar - hover modification

### DIFF
--- a/src/opnsense/www/js/opnsense_theme.js
+++ b/src/opnsense/www/js/opnsense_theme.js
@@ -30,8 +30,14 @@
 
 $(document).ready(function () {
     // traverse loaded css files
-    var toggle_sidebar_loaded = false;
-    $.each(document.styleSheets, function(sheetIndex, sheet) {
+	var toggle_sidebar_loaded = false;
+	var nomouse = "mouseenter mouseleave";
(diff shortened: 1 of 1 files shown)


opnsenseuser:
@fabianfrz -> is it better now or even worse?

src/opnsense/www/js/opnsense_theme.js
```diff
@@ -40,70 +46,119 @@ $(document).ready(function () {
...
+		$(".toggle-sidebar > i").toggleClass("fa-chevron-right fa-chevron-left");
+		if ($("#navigation").hasClass("col-sidebar-left")) {
+			$(".brand-logo").css("display", "none");
+			$(".brand-icon").css("display", "inline-block");
+			trigger_sidebar();
+			if (store && window.sessionStorage) {
+				sessionStorage.setItem('toggle_sidebar_preset', 1);
+				transition_duration(0);
+				}
+			} else {
```
fabianfrz: To answer your question: Only from the indent in this diff view - to which if does it belong and to which dies it really belong?

opnsenseuser:
@fabianfrz -> i will re-upload the opnsense_theme.js and fix also the brackets problem. now it should be fine!

PR #2714, opened 2018-09-11
dhcpd6 PD Auto tracking additons

marjohn56:
Does not use dhcp6c mods. uses existing functions to determine the PD length available.


fichtner:
merged, thanks!

PR #2716, opened 2018-09-12
Make IPSec Phase1 defaults more secure

mimugmail:
3DES shouldn't be used anymore, same to SHA1. Also, DH, the higher the number doesn't mean it's more secure. DH22-24 should be avoided in a new draft of RFC:

https://tools.ietf.org/html/draft-ietf-ipsecme-rfc4307bis-18#section-2.4

The new values are marked as MUST or SHOULD+ values from the RFC.


fichtner:
Merged, thanks!

PR #2721, opened 2018-09-15
add OPNsense UI-Bootgrid timestamp formatter

fabianfrz:
needed for the nginx plugin and it is so generic that I can go to core

### DIFF
--- a/src/opnsense/www/js/opnsense_bootgrid_plugin.js
+++ b/src/opnsense/www/js/opnsense_bootgrid_plugin.js
@@ -1,5 +1,6 @@
 /**
  *    Copyright (C) 2015 Deciso B.V.
+ *    Copyright (C) 2018 Fabian Franz
  *
  *    All rights reserved.
  *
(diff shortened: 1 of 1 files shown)


AdSchellevis:
Maybe, but if we do, we need to consider the other types already in the codebase as well, this isn’t used anywhere else yet, others may be duplicate already. Besides that I’m unsure if additions merge at the moment or overwrite the presets in this file (Which would lead to unpredictable results),

There’s quite some room for improvement in the current wrapper, but if we change things we should try to make things more consistent for all components using this.

fabianfrz:
if it is set in another file, this function here is lost because the object will be replaced.

fichtner:
@fabianfrz what's the status here?

AdSchellevis:
@fichtner keep it open for now, it should be moved to https://github.com/opnsense/jquery-bootgrid

fabianfrz:
@fichtner A copy of it has been added to opnsense version of jquery-bootgrid. Maybe it fits better there. Awaiting a final decision if we patch it into our calling function or if we maintain our own fork of bootgrid.

PR #2730, opened 2018-09-18
monit: change mail format field for alerts to textbox

fbrendel:
Change the Mail Format field for Alerts to textbox to configure it multi-line.
The problem came up with opnsense/plugins#820


fbrendel:
yup :grinning:

fichtner:
Merged, thanks!

PR #2735, opened 2018-09-19
Add Active Schedule icon to firewall_rules display

framer99:
This makes reviewing a bunch of Firewall rules with various schedules much nicer.

active_sched_icon


framer99:
oops, didn't realize I had 3 commits in this PR, closing and will resubmit

fichtner:
@framer99 we can squash them via github so it's not no-no anymore :)

I like this idea, but would maybe reuse the same schedule icon and show it muted if it is currently not active maybe?

framer99:
OK, thanks, I was trying to figure out how to revert and squash on my merge/update to resubmit... I'm not too skilled with git or github. Now i have a mess with a revert on top of the 3 commits above :)

Looking at my firewall rule page now, I agree it would may be better to have it always present but different color or muted.

Are you suggesting to keep the existing calendar icon and just making it muted/non-muted? Or to make the "new" clock icon always present and muted/non muted?

I can see combining them but the tool tip would need to contain combined Edit/active or Edit/nonActive text.

If we make the clock icon always present and muted/non muted, then we probably should make it function the same way on the Schedules page... and perhaps other places schedules appear that I'm not familiar with.

framer99:
OK, let me just close this PR and come back with something new later after i figure out my mess :)
Probably not today, maybe this weekend.

fichtner:
no rush, thanks for your work! :)

Issue #2726, opened 2018-09-16
Wierd IPv6 issue when using PPPoE


marjohn56:
We noticed that if you save and apply the WAN interface, don't take it down - just save and apply then when the interface comes back up there is no IPv6 address on the PPPoE interface. I think this is because the PPPoE interface is not ready for it. Sort of proved this by putting a sleep(5) ( maybe a bit long ) in the case 'ppp': call where it's setup for V4.

Interfaces.inc Lines 211 on.

> switch ($wancfg['ipaddr']) {
        case 'dhcp':
            interface_dhcp_configure($interface);
            break;
        case 'pppoe':
        case 'l2tp':
        case 'pptp':
        case 'ppp':
            interface_ppps_configure($interface);
            sleep(5);
            break;
        default:
            /* XXX: Kludge for now related to #3280 */
            if (!in_array($tunnelif, array("gif", "gre", "ovp"))) {
                if (is_ipaddrv4($wancfg['ipaddr']) && $wancfg['subnet'] <> "") {
                    legacy_interface_setaddress($realif, "{$wancfg['ipaddr']}/{$wancfg['subnet']}");
                }
            }
            break;
    }

Very likely not the way to fix this, but I wonder if this is causing some of the odd v6 behaviour sometimes seen on the forum.

marjohn56:
Have a little free time to take a look at this. It appears it only happens if you have a static v6 with PPPoE on v4. I think if you are using dhcp6c the natural sequence of things allows that to work, but with static v6 it appears it's trying to add the v6 address before the interface is ready. I know it fixes it using a generic sleep(5) before applying the static, I'll reduce that until it no longer works then increase it and also add a catch so that it only sleeps if dhcp4 is pppoe and v6 is static.

marjohn56:
Use IPv4 connectivity is checked, as it's using the PPPoE link.

fichtner:
@marjohn56 noted via https://github.com/opnsense/core/pull/2736

> That bits OK, but the dev version has something amiss in the firewall, it was not allowing any of my rules for my servers, everything was hitting the default deny.

Did you solve this or potentially still happening? There are a few minor firewalling changes, but nothing drastic that would force this.

marjohn56:
No, I had to switch back to production due to complete lack of email coming through, SWMBO was muttering. I'll try and have another look when I can.

fichtner:
Alias issue solved in -devel

marjohn56:
[...]

PR #2736, opened 2018-09-19
Fix for static IPv6 when using PPPoE IPv4 on WAN

marjohn56:
When using a static WAN IPv6 address and using PPPoE for v4,  if you save and apply the interface - this is when the interface is already up, the IPv6 address does not get applied, an error from ifconfig is generated. It appears as though the interface is not ready to accept the v6 address. This fix adds a slight delay in calling ifconfig if both the IPv6 address is static and the v4 is pppoe.

This is a fix for issue 2726 [https://github.com/opnsense/core/issues/2726](url)

### DIFF
--- a/src/etc/rc.newwanip
+++ b/src/etc/rc.newwanip
@@ -129,6 +129,13 @@ if (isset($config['interfaces'][$interface]['ipaddrv6'])) {
                 interface_dhcpv6_configure($interface, $config['interfaces'][$interface]);
             }
             break;
+       default:
+                if (!in_array($tunnelif, array("gif", "gre", "ovp"))) {
+                    if (is_ipaddrv6($config['interfaces'][$interface]['ipaddrv6']) && $config['interfaces'][$interface]['subnetv6'] <> "") {
+                        mwexec("/sbin/ifconfig " . escapeshellarg($interface_real) . " inet6 {$config['interfaces'][$interface]['ipaddrv6']} prefixlen " . escapeshellarg($config['interfaces'][$interface]['subnetv6']));
+                    }
+                }
+            break;
     }
 }
 


fichtner:
For static IPv6 on top of a PPPoE this may be more suited for rc.newwanip, see my question on the Use IPv4 toggle state?

marjohn56:
Use IPv4 connectivity is checked, as it's using the PPPoE link.

fichtner:
Ok, that means the spot you are looking for is here, no sleep required:

https://github.com/opnsense/core/blob/master/src/etc/rc.newwanip#L118


Cheers,
Franco

marjohn56:
Interesting, as it only happens if you save and and apply when the interface is already up, if you take disable the interface - save & apply then enable - save & apply it's OK.

marjohn56:
We are talking static IPv6, not SLAAC or dhcp6c, even though it's using the v4 connectivity,

fichtner:
the switch just needs:

https://github.com/opnsense/core/blob/master/src/etc/inc/interfaces.inc#L2443-L2450

marjohn56:
OK.. done that, I get warnings that the route cannot be applied, but it all seems to work fine.

fichtner:
looks good, ready for pull? :)

marjohn56:
Want kev to check his too. We may be the only two who use this strange setting. :) 

I've emailed him.

fichtner:
ok, let me clean up related fluff in interfaces.inc ... that shouldn't get in the way of the merge now

marjohn56:
He'll test it and post here, but it could be tomorrow.

fichtner:
ok, no problem. just let me know :)

nivek1612:
tested works like a charm :-)

fichtner:
Thanks, merged! :)

fichtner:
@marjohn56 would you mind testing this on top of master @ 9d37f in the next few days... I made a reusable function. works for my static v4/v6, but I don't have the PPPoE / Use v4 part.

marjohn56:
My live system is now development... ooerr!

Works fine. 
Reboot OK
Save and apply OK.

fichtner:
brilliant, many thanks ❤️

marjohn56:
That bits OK, but the dev version has something amiss in the firewall, it was not allowing any of my rules for my servers, everything was hitting the default deny.

marjohn56:
@fichtner - Finally had chance to test again regarding my last comment. Dev version definately NOT working. Something is very broken in the Aliases. Firstly, the existing aliases are not imported into the GUI. Re-creating the alias entries creates some real strangeness in the config.xml and still they do not work. Going to email the xml files to you.

fichtner:
thanks, looking at it now. we have the new aliases planned for 18.7.4 which is basically ready so worst case we have to pull 18.7.4 last minute. keep you posted.

marjohn56:
I will switch my test unit to release to check whatever is needed, won't have another chance to mess with the live unit but I can edit the aliases into the test unit config.xml

PR #2737, opened 2018-09-19
TrafficShaper: API enable Status and fix command

mimugmail:
Legacy code has ipfw pipe show, but it should be ipfw sched show.
While here ... what have you api enabled today? :)

### DIFF
--- a/src/opnsense/mvc/app/controllers/OPNsense/TrafficShaper/Api/ServiceController.php
+++ b/src/opnsense/mvc/app/controllers/OPNsense/TrafficShaper/Api/ServiceController.php
@@ -1,7 +1,6 @@
 <?php
 /**
  *    Copyright (C) 2015-2016 Deciso B.V.
- *    Copyright (C) 2018 Michael Muenz <m.muenz@gmail.com>
  *
  *    All rights reserved.
  *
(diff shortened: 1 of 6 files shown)


AdSchellevis:
@mimugmail I kind of ignored the legacy page since I think we should parse the data when we API enable id, let me think about this. It's not wrong, but it doesn't real bring much to another client trying to gather stats..... other tools can't really do anything with the data gathered, since they will always need to parse it themselves.

mimugmail:
@AdSchellevis for me at the beginning it was also an option to just change the command in the old static diag.php page. My thought was you want to get rid of all static php pages and when Smart-Soft pulled a PR for Trust MVC you said it's better to start with smaller things like Diagnostic stuff, so I tried to move this to the new code base ...

AdSchellevis:
@mimugmail I understand, still think that it's indeed better to translate smaller (stand alone) diagnostic pages first like I mentioned back then. Part of the translation should hoever be parsed data that could be useful to other subscribers as well, otherwise it doesn't really make sense to migrate the old pages. 

My concern is that eventually we end up with the legacy diagnostics pages we have now into basically the same but having a different url, which looks like window dressing. Fixing this properly takes time, but I'm convinced we shouldn't settle for less. 

For example take a look at the arp diagnostics (/ui/diagnostics/interface/arp).

If you would be willing to invest more time in this, I would be happy to help you extend this into something more meaningful, but a quick fix to the legacy code is also acceptable.

Just let me know what you prefer.

mimugmail:
I'm sure you have the broader overview to this project, so I can accept. Closing this and open a new one later today for the legacy page

PR #2739, opened 2018-09-20
Translate systemhealth.

kekek2:


kekek2:
https://github.com/opnsense/lang/pull/42

src/etc/inc/rrd.inc
```diff
@@ -51,6 +51,17 @@ function rrd_configure($verbose = false)
     $rrddbpath = "/var/db/rrd/";
     $rrdgraphpath = "/usr/local/www/rrd";
 
+    gettext("system");
```
fichtner: I don't think this is future-proof. It should somehow be tied to the controller file as an array of translations, e.g.

```
[...]
traffic => gettext('Traffic'),
vpnusers => gettext('VPN Users'),
[...]
```

and let the controller fill the label name through the API for the volt page if it's found in the translation array

PR #2746, opened 2018-09-22
Update diag_limiter_info.php

mimugmail:
Changed ipfw pipe show to ipfw sched show, output is the same but with the correct scheduler.

### DIFF
--- a/src/www/diag_limiter_info.php
+++ b/src/www/diag_limiter_info.php
@@ -30,7 +30,7 @@
 require_once("guiconfig.inc");
 
 if (isset($_POST['getactivity'])) {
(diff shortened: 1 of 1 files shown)


AdSchellevis:
@mimugmail thanks!

fichtner:
also pushed to stable, thanks ❤️

mimugmail:
IMHO in medium-term this or something similar should replace the static page .. it's true that api doesn't make sense for just printing a command output, but you could easily add actions to just print drop counters and let montioring systems like zabbix alert you, when they receive thresholds over X via an api call ...

Issue #2701, opened 2018-09-08
OpenVPN: Client Specific Overrides problem in 18.7.2 


fprina:
After the 18.7.2 upgrade seems OpenVPN do not create anymore the CSO files.

Also the RADISU Framed-Route and ramed-IP-Address attributes are ignored

AdSchellevis:
upgrade from where?

AdSchellevis:
what does the openvpn log report?

fprina:
This is an authentiation session
`
Sep  8 12:45:28 nme-fw1-prod openvpn[24686]: MULTI: multi_create_instance called
Sep  8 12:45:28 nme-fw1-prod openvpn[24686]: 1xx.xx.xx.xx:55114 Re-using SSL/TLS context
Sep  8 12:45:28 nme-fw1-prod openvpn[24686]: 1xx.xx.xx.xx:55114 LZO compression initializing
Sep  8 12:45:28 nme-fw1-prod openvpn[24686]: 1xx.xx.xx.xx:55114 Control Channel MTU parms [ L:1622 D:1140 EF:110 EB:0 ET:0 EL:3 ]
Sep  8 12:45:28 nme-fw1-prod openvpn[24686]: 1xx.xx.xx.xx:55114 Data Channel MTU parms [ L:1622 D:1450 EF:122 EB:406 ET:0 EL:3 ]
Sep  8 12:45:28 nme-fw1-prod openvpn[24686]: 1xx.xx.xx.xx:55114 Local Options String (VER=V4): 'V4,dev-type tun,link-mtu 1602,tun-mtu 1500,proto UDPv4,comp-lzo,keydir 0,cipher AES-128-CBC,auth SHA512,keysize 128,tls-auth,key-method 2,tls-server'
Sep  8 12:45:28 nme-fw1-prod openvpn[24686]: 1xx.xx.xx.xx:55114 Expected Remote Options String (VER=V4): 'V4,dev-type tun,link-mtu 1602,tun-mtu 1500,proto UDPv4,comp-lzo,keydir 1,cipher AES-128-CBC,auth SHA512,keysize 128,tls-auth,key-method 2,tls-client'
[...]

PR #2748, opened 2018-09-22
openvpn, option to match CSO against common_name or login

fprina:
Openvpn option to select with between CN and username must be check as CSO


fprina:
Related to issue #2701

AdSchellevis:
@fprina thanks, I will pull it in later with some minor adjustments

PR #2753, opened 2018-09-24
Fix translate.

kekek2:


src/etc/inc/config.inc
```diff
@@ -272,7 +272,7 @@ function make_config_revision_entry($desc = '')
     }
 
     if (empty($desc)) {
-        $desc = sprintf('%s made changes', $_SERVER['SCRIPT_NAME']);
+        $desc = sprintf(gettext('%s made changes'), $_SERVER['SCRIPT_NAME']);
```
fichtner: This was intentional, please don't change it.

src/www/firewall_rules_edit.php
```diff
@@ -756,7 +756,7 @@ function is_posnumericint($arg) {
                     <td>
 <?php
                     if (!empty($pconfig['floating'])): ?>
-                      <select name="interface[]" title="Select interfaces..." multiple="multiple" class="selectpicker" data-live-search="true" data-size="5" tabindex="2" <?=!empty($pconfig['associated-rule-id']) ? "disabled" : "";?>>
+                      <select name="interface[]" title="<?=gettext("Select interfaces...") ?>" multiple="multiple" class="selectpicker" data-live-search="true" data-size="5" tabindex="2" <?=!empty($pconfig['associated-rule-id']) ? "disabled" : "";?>>
```
fichtner: requires html_safe() around gettext()

AdSchellevis:
timeout, no feedback.

PR #2769, opened 2018-09-28
make Schedule icon reflect current schedule state

framer99:
2018-09-28 13_06_55-lan _ rules _ firewall
On firewall_rules.php, there is no indication whether a schedule-based
rule is active. This change to the schedule icon applies the same styles
that are applied to a disabled/enabled Pass rule icon (text-muted and
text-success).

The break added to the foreach loop is needed to retain reference to the
attached schedule for the filter_get_time_based_rule_status() call

### DIFF
--- a/src/www/firewall_rules.php
+++ b/src/www/firewall_rules.php
@@ -652,6 +652,7 @@ $( document ).ready(function() {
                                 if ($schedule['name'] == $filterent['sched'])
                                 {
                                     $schedule_descr = (isset($schedule['descr'])) ? $schedule['descr'] : "";
+                                    break;
                                 }
                             }
                         }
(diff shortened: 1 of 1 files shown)


AdSchellevis:
@framer99 I think it's a nice addition, but can you extend the footer as well to reflect the new/changed links? 
image

fichtner:
Yup, I agree. Thanks for the idea and contribution! :)

framer99:
I didn't noticed that there was a Schedule icon in your image you posted. I was focused on that green triangle so I jut added the schedule icon to the top two rows of the footer.

As I was taking a new screenshot I noticed the existing Schedule further down so I made another change for that.

I think you probably wanted the second change only. 

Screenshot:
2018-09-29 10_45_29-window

AdSchellevis:
@framer99 yes, let's take the icon on the bottom only, it clarifies the new feature enough. I'll cherry-pick your changes and close the pull, thanks!

AdSchellevis:
https://github.com/opnsense/core/commit/480fb97b8a33ed84701c1a8b138d55a3871a8792 https://github.com/opnsense/core/commit/c1c5ef115e29ff8fae5011bac83e07dd9880d2b8

PR #2781, opened 2018-10-02
Add RFC3118 Option

marjohn56:
This allows us to call functions defined in a new file. rfc3118_lib.inc which will generate the correct credentials where they need to be hashed or not. In the function in this PR, the function returns all of the request and send options required for  Orange France and fills in the client fields autonatically.

Also in the case of Orange France an extra entry box appears to allow the entry of the LiveBox_ID.

The PR is a work in progress, but all that is needed to nake it work for Orange France is the hash routine.

It;'s also possible to use the functions create the hex strings needed for other ISPs and auto fill the options as needed.

### DIFF
--- a/src/etc/inc/rfc3118_lib.inc
+++ b/src/etc/inc/rfc3118_lib.inc
@@ -1,0 +1,63 @@
+<?php
+
+/*
+ *    Copyright (C) 2016 Deciso B.V.
+ *    All rights reserved.
+ *
+ *    Redistribution and use in source and binary forms, with or without
+ *    modification, are permitted provided that the following conditions are met:
+ *
+ *    1. Redistributions of source code must retain the above copyright notice,
+ *       this list of conditions and the following disclaimer.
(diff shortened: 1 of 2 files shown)


nivek1612:
Initial testing this looks good just needs a few minor tweaks. We now need to work on building the option-90 (dhcp) and raw-option-11 (dhcp6c) authentication field

fichtner:
Thanks! I have merged it a little different moving the function from its file to interfaces.php.

I think this is a great way to deal with the UI, but we may have to change the approach underneath to generate the options from the configuration: think about further changes and a firmware update-- we want to be able to deploy the changes directly after reboot which doesn't work yet because the user would have to go back to the page and hit "save". Also need to be careful when writing the config file / merging advanced DHCP options set in the respective field. Maybe both could work in tandem eventually?

I'm not around for the next few days. Talk to you next week!


Cheers,
Franco

PR #2785, opened 2018-10-03
opnsense theme chrome browser outline fix

opnsenseuser:
opnsense theme chrome browser outline fix
@fichtner


fichtner:
Merged, thanks!

PR #2788, opened 2018-10-03
core: update config for squid version 4

fabianfrz:
Don't merge as long as we are using v3


fabianfrz:
@fichtner don't forget about this.

fichtner:
Sure, blocked until 19.1.1 where we can safely use both squid packages so devel can move to version 4.

fabianfrz:
@fichtner ok, I thought this is for 19.1.0, then it's ok.

fichtner:
merged, thanks!

PR #2789, opened 2018-10-03
Correct missing '"'' at the end of the userclass field

marjohn56:
Correct missing '"' and copy paste error with a "text" field.


fichtner:
Merged, thanks!

PR #2790, opened 2018-10-04
Add FTI/ to password - tidy end of useridhex

marjohn56:
Add the missing FTI/ to the start of the $username


fichtner:
Merged, thanks!

PR #2802, opened 2018-10-09
Use single guillemets for previous/next page

MichaelDeciso:
Currently, pagination uses double guillemets for first/last page and lower than/greater than signs for previous next (like so: `« < 1 2 3 > »`). This looks ugly. Change this to use single guillemets, for visual consistency (like so: `« ‹ 1 2 3 › »`).

### DIFF
--- a/src/opnsense/www/js/jquery.bootgrid.js
+++ b/src/opnsense/www/js/jquery.bootgrid.js
@@ -469,7 +469,7 @@
 
                 renderPaginationItem.call(this, pagination, "first", "&laquo;", "first")
                     ._bgEnableAria(current > 1);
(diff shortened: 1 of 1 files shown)


AdSchellevis:
@MichaelDeciso we've forked bootgrid to https://github.com/opnsense/jquery-bootgrid, I agree this looks a bit nicer, but ideally we should patch our bootgrid fork as well. 

Even better would be if we could build it from there and copy the results, but for some reason I could get the javascript build functional ;(  (so with the one patch we have I decided to at least make sure the sources match)

MichaelDeciso:
I have created https://github.com/opnsense/jquery-bootgrid/pull/1 to properly modify the non-compiled version as well.

PR #2807, opened 2018-10-11
dhcp: add OMAPI configuration support

yuri-moens:
This adds support for configuring the OMAPI service in the dhcpd4 configuration, both with and without key.


fichtner:
Merged manually due to conflicts, see 6c91d7f87a1c7f

Thanks! ❤️

PR #2810, opened 2018-10-13
border between <a>´s are missing

opnsenseuser:
I noticed that the separation of the individual <a> links is missing. Now it is up to you to say if we should install it again or not. if you do not want it please just delete this pr. have added two screenshots as a comparison. Thank you, best regards, René

new would be:
grafik

old was:
grafik

### DIFF
--- a/src/opnsense/www/themes/opnsense/assets/stylesheets/main.scss
+++ b/src/opnsense/www/themes/opnsense/assets/stylesheets/main.scss
@@ -386,7 +386,7 @@ main.page-content.col-lg-12 {
         padding-left: 0px;
         padding-right: 0px;
         padding-top: 12px;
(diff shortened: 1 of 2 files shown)


opnsenseuser:
@fichtner if it does not suit you, just delete the pr. Thank you

fichtner:
Good idea, thanks, merged!

opnsenseuser:
@fichtner 
@AdSchellevis sorry, lightergrey doesnt work

grafik

opnsenseuser:
sorry, posted my comment in the wrong closed pr.:-(

opnsenseuser:
@fichtner i just read your commit -> The normal theme menu does not use separators so the sidebar
shouldn't use them as well. Noticed this morning with a fresh
pot of coffee.

ok!

PR #2813, opened 2018-10-14
Restore area - line break is missing! - fix

opnsenseuser:
Restore area - line break is missing! - fix
i changed it using <p> </p>

see my screenshots for comparison

old:
grafik

new:
grafik

### DIFF
--- a/src/www/diag_backup.php
+++ b/src/www/diag_backup.php
@@ -384,14 +384,14 @@ $( document ).ready(function() {
                 <tr>
                   <td>
                     <?=gettext("Restore area:"); ?>
-                    <select name="restorearea" id="restorearea" class="selectpicker">
+                    <p><select name="restorearea" id="restorearea" class="selectpicker">
                       <option value=""><?=gettext("ALL");?></option>
 <?php
                     foreach($areas as $area => $areaname):?>
                       <option value="<?=$area;?>"><?=$areaname;?></option>
 <?php
                     endforeach;?>
-                    </select><br/>
+                    </select></p>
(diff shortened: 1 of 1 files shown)


opnsenseuser:
@fichtner 
by the way....why does the "search" button appear in german even though i have set the language in english? regards rené

opnsenseuser:
@fabianfrz fixed it with an div element!

fabianfrz:
@opnsenseuser a br tag is only a line break which is always the line height. There are lot's of this br tags left but It is always better to separate them and style it with CSS when needed so you can override it in your themes.

opnsenseuser:
I noticed that all "search" buttons are in German!

fabianfrz:
@opnsenseuser the bootgrid or the the old style pages?

opnsenseuser:
i think these are the old style pages! or am i wrong?
grafik

fabianfrz:
you are right, but it is built like the new ones in this specific case. This input element is rendered by your browser and cannot be modified. You can switch it's language by switching the browser language.

Those input fields cannot be styled and a common workaround is hiding it and then forward the events from a custom element to it.

opnsenseuser:
I've just found out that it works with javascript, but whether it's worth the effort, someone else has to decide

opnsenseuser:
We would have to find an alternative! But that certainly needs a lot of effort.

fabianfrz:
Not really - it can be done via JS in a global file.

opnsenseuser:
@fichtner 
@AdSchellevis sorry, but lightergrey doesn´t work

see->
grafik

Issue #2811, opened 2018-10-13
Monit: support commandline parameters in Custom Services


poggenpower:
Hi,
found that monit supports command line parameters at least since the latest release.
But the UI verifies the if the PATH  is a file only. 
Would be great if command line parameters are supported here. 
In the munit config path  +  parameters need to be given as one string in quotes. 
Tried this out by manually editing the monit.config and it is working find from a monit prospective. 
Here an example:
`check program raid-md127 with path "/sbin/mdadm --misc --test /dev/md127"
   if status != 0 then alert`

If you touch the code there you may also want to allow to give the timeout parameter. 

Thanks for your great work
   Thomas

fbrendel:
Hi Thomas,

the program path is quoted now and the timeout parameter can be set as an advanced option.

Cheers
Frank

poggenpower:
Hi Frank,

thanks for the fast response. And fix. 
Just for verification, as it is not mentioned in you comment. You allow parameter in addition to the quotes? And the verification check if the first part is an existing file?

Thanks again
   Thomas

fbrendel:
> Just for verification, as it is not mentioned in you comment. You allow parameter in addition to the quotes? And the verification check if the first part is an existing file?
> 
The additional parameters have already worked.
The quotes are necessary due to the timeout parameter.
And the verification only checks if the first part is a path regardless if the file exists or not.

Hope that helps :wink:

poggenpower:
Had no luck creating a custom service (without timeout) until I have added the quotes in the config manually. Before monit was complaining about the config file. 
However if it is fixed now I am happy. WHich release will contain this fixes?

fichtner:
18.7.6 or later. Needs review to see how error-prone it is against injection.

PR #2815, opened 2018-10-15
monit: add quotes and timeout to custom program path

fbrendel:
Add quotation marks to custom program path and the timeout parameter.
See #2811 and [Monit Service Checks](https://mmonit.com/monit/documentation/monit.html#Program)


fichtner:
Discussed with @fbrendel -- merged, thanks!

Issue #2114, opened 2018-01-20
PC Engines APU* LED support


ktk:
Unless I missed something it looks like OPNsense does not include default support for onboard LEDs on [PC Engines APU*](http://pcengines.ch/apu2.htm) boards.

It looks like there is a kernel module available for them called [apuled](https://bugs.freebsd.org/bugzilla/show_bug.cgi?id=189772). According to the FreeBSD issue it supports APU1, APU2 and APU3.

Would be great if this could be shipped/included in the default configuration for serial builds so one can enable it. On ALIX boards/pfsense a popular package was [BlinkLED](https://doc.pfsense.org/index.php/BlinkLED_Package), which would probably easy to re-implement once I have the LEDs available.

Thanks for the awesome work!

fichtner:
Hi @ktk,

The led driver nodes should be there for your APU already...

    # ls /dev/led/led*

And it is easy to operate via shell script:

https://www.freebsd.org/cgi/man.cgi?query=led&sektion=4


Cheers,
Franco

ktk:
@fichtner oh great didn't realize that thanks! Will update the posts in the forum accordingly. Also I will try to see if I can port BlinkLED! Will report here as well for reference.

fichtner:
Best way of integrating would be to rewrite the management parts in the MVC framework. That would automatically provide an API for future use... :)

https://docs.opnsense.org/development/examples/helloworld.html

ktk:
@fichtner tnx will check that out!

ktk:
@fichtner just checked on 17.7, I only see:

```
ls /dev/led/*
/dev/led/igb0   /dev/led/igb1   /dev/led/igb2
```

fichtner:
I don't think the drivers disappeared, but seeing igb* maybe the new APU leds are not found... I have to check on Monday on an old APU1 to see if they are available there, ok?

ktk:
@fichtner sure thanks!

fichtner:
please ping me when I'm slacking off :)

fabianfrz:
Also having an APU1 board so I can check for /dev/led as well

fabianfrz:
there is no led on APU1d4:

```
root@opn:~ # ls /dev/*
/dev/acpi       /dev/ctty       /dev/da0s1      /dev/geom.ctl   /dev/mem        /dev/random     /dev/ttyu0.init /dev/ufssuspend /dev/ugen6.2
[...]

PR #2819, opened 2018-10-17
Add runlevel 'boot' for syshook (from Smart-Soft)

alexpro:
To implement as described in opnsense/core#2114
This runlevel for running tasks before filesystems are mounted and checked.

### DIFF
--- a/src/etc/rc
+++ b/src/etc/rc
@@ -42,6 +42,8 @@ export HOME PATH
 
 echo "Mounting filesystems..."
 
+/usr/local/etc/rc.syshook boot
+
 # tunefs may refuse otherwise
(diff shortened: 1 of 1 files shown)


fichtner:
Anything earlier than "early" is more harmful than helpful even though your motivations for the LED operation are clear.

evbevz:
@fichtner you want to move apuled syshook to 'early'?

alexpro:
Moved to 'early' runlevel.
I suggested running daemon before any actions performed by rc in order to startup indication begins immediately after rc runs. In some cases (such as emergency shutdown) - there will be no indication of system startup until all filesystems are checked.

PR #2823, opened 2018-10-18
Added CaptivePortal API connect action

zvs44:
Added a connect action to the CaptivePortal API as discussed here https://forum.opnsense.org/index.php?topic=2028.msg6402


AdSchellevis:
@zvs44 ok, thanks!

PR #2829, opened 2018-10-23
monit: change status page look-and-feel

fbrendel:
The status page uses css classes for text colors now.
Hence it's customizable via themes.
opnsense/plugins#928


fichtner:
Merged, thanks!

PR #2830, opened 2018-10-23
Move selected ... to end

tbandixen:
What I did:

- Some upper and lower case corrections on the tooltips.
- Rearranged the buttons order in "Firewall: Settings: Normalization"
- Correction for the "move to end" function for "Firewall: Rules: xxx"


fichtner:
Thanks for this, merged! ❤️

PR #2845, opened 2018-10-26
www/nginx: adjust some core scripts to make it compatible with exteral JS

fabianfrz:
This PR provides the possibility to store some json data in hidden fields so data is exchanged without messing a lot with the core API.


src/opnsense/www/js/opnsense.js
```diff
@@ -135,26 +140,27 @@ function setFormData(parent,data) {
                         targetNode.empty(); // flush
                         $.each(node[keypart],function(indxItem, keyItem){
                             if (keyItem["selected"] != "0") {
-                                targetNode.append("<option value='"+indxItem+"' selected>" + keyItem["value"] + " </option>");
+                                targetNode.append(`<option value="${indxItem}" selected>${keyItem["value"]}</option>`);
                             } else {
-                                targetNode.append("<option value='"+indxItem+"'>" + keyItem["value"] + " </option>");
+                                targetNode.append(`<option value="${indxItem}">${keyItem["value"]}</option>`);
```
fabianfrz: @fichtner this is not supported by IE11 but by Edge, Chome and Firefox. should I undo this two lines?

fabianfrz: this is standard but it seems like IE does not get modern features anymore. If you look at this page, you see that IE 11 does not get modern JS features anymore:
https://kangax.github.io/compat-table/es6/

fabianfrz: done btw.

src/opnsense/www/js/opnsense.js
```diff
@@ -142,19 +147,20 @@ function setFormData(parent,data) {
                         });
                     } else if (targetNode.prop("type") == "checkbox") {
                         // checkbox type
-                        if (node[keypart] != 0) {
-                            targetNode.prop("checked",true);
-                        } else {
-                            targetNode.prop("checked",false);
-                        }
+                        targetNode.prop("checked", node[keypart] != 0);
```
AdSchellevis: nice catch

src/opnsense/www/js/opnsense.js
```diff
@@ -142,19 +147,20 @@ function setFormData(parent,data) {
...
                     } else if (targetNode.is("span")) {
                         if (node[keypart] != null) {
                             targetNode.text("");
                             targetNode.append(htmlDecode(node[keypart]));
                         }
                     } else {
                         // regular input type
-                        targetNode.val(htmlDecode(node[keypart]));
+                        if (targetNode.hasClass('json-data')) {
```
AdSchellevis: better to move up `else if (targetNode.hasClass('json-data'))`

src/opnsense/www/js/opnsense.js
```diff
@@ -142,19 +147,20 @@ function setFormData(parent,data) {
...
+                            // if the input field is JSON data, serialize the data into the field
+                            targetNode.val(JSON.stringify(node[keypart]));
```
AdSchellevis: better to use a data tag for this, should avoid json serialisation:

`targetNode.data('json', node[keypart]);
`

src/opnsense/www/js/opnsense.js
```diff
@@ -94,7 +94,12 @@ function getFormData(parent) {
                     }
                 } else {
                     // regular input type
-                    node[keypart] = sourceNode.val();
+                    if (sourceNode.hasClass("json-data")) {
+                        // deserialize the field content - used for JS maintained fields
+                        node[keypart] = JSON.parse(sourceNode.val());
```
AdSchellevis: better to use a data tag for this, should avoid json serialisation:

`node[keypart] = sourceNode.data('json');`

(We're doing something similar for another project)

src/opnsense/www/js/opnsense.js
```diff
@@ -94,7 +94,12 @@ function getFormData(parent) {
                     }
                 } else {
                     // regular input type
-                    node[keypart] = sourceNode.val();
+                    if (sourceNode.hasClass("json-data")) {
```
AdSchellevis: better to move up, `else if (sourceNode.hasClass("json-data"))`

src/opnsense/www/js/opnsense.js
fabianfrz: data does not use a data attribute but a jQuery cached data stucture but it is better to handle anyway. The only difference is that I used 'data' as a name, since it is just a JavaScript structure and has theoretically nothing to do with JSON.

src/opnsense/mvc/app/models/OPNsense/Base/Constraints/SetIfConstraint.php
```diff
@@ -0,0 +1,75 @@
...
+        if ($node_class == "OPNsense\Base\FieldTypes\BooleanField") {
+            return empty((string)$node);
+        } elseif (empty((string)$node) || (string)$node == "0") {
+            return true;
+        }
+        return false;
+    }
+    /**
+     * Executes validation, expects a list of fields in "addFields" which to check for content.
+     * Fields are concerned empty if boolean false or containing an empty string
```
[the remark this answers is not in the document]
fabianfrz: @AdSchellevis I've copied some code as a template and changed the worker method. Now the copied empty check is also gone so there is only the check method left.

fabianfrz:
@AdSchellevis all your change requested should be included.

AdSchellevis:
@fabianfrz looks good, thanks!

PR #2859, opened 2018-10-28
dhcpv6 static leases with tracked interface: apply prefix to fixed IPv6 address

PiQuer:
When operating a dhcpv6 server on an interface with IPv6 configuration mode "Track Interface" with "Allow manual adjustment of DHCPv6 and Router Advertisements", static leases with fixed IP don't work as expected. One is required to enter a full IPv6 address, the static lease will break whenever the dynamic prefix changes. The same problem is already solved for the configuration field "Range": when using a dynamic prefix it is possible to enter only the suffix part, the configuration file is automatically generated with the current prefix.
This patch uses the same technique to generate a fixed IPv6 address for static leases. The help message advises to use only the suffix when a dynamic prefix is used, and in the generated configuration file the prefix will be added.

### DIFF
--- a/src/etc/inc/services.inc
+++ b/src/etc/inc/services.inc
@@ -1386,6 +1386,10 @@ host s_{$dhcpv6if}_{$i} {
 
 EOD;
                 if (!empty($sm['ipaddrv6'])) {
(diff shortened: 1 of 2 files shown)


fichtner:
@PiQuer thanks, this looks good! I'll merge after testing at the office. :)

PR #2863, opened 2018-11-01
certmanager: allow wildcard DNS subjectAltName

clandmeter:


clandmeter:
Ah somebody had a similar idea :).

fichtner:
@clandmeter your submission looks good too, yet the other one is already queued up for release in 18.7.7. Thanks anyway! :)

clandmeter:
@fichtner Ive seen this popup another time in the code (maybe more). So it could be a cleaner way to add an option to `is_hostname ` to allow wildcard hostnames and reuse it. Anyways im glad its resolved now.

AdSchellevis:
@clandmeter @fichtner I didn't add it to `is_hostname()` since I couldn't find other practical use for it. With a wildcard it technically isn't really a hostname anymore, which kind of makes it a different type of validation in my opinion which tends to get rather application specific.

PR #2865, opened 2018-11-01
sidebar - submenu may be cropped due to small height (fix)

opnsenseuser:
@fichtner ready for testing (Submenus now open upwards if the space down is not sufficient.)
and by the way....
i need a pro (like you or fabian) or someone else who can optimize the jquery code because it´s memory usage.the calculations of the current postition with offset and scrollTop brings jquery to the performance limit. and unfortunately you can not calculate that, unfortunately.

Perhaps @fabianfrz can help?

i did my best and I like to learn!

### DIFF
--- a/src/opnsense/www/js/opnsense_theme.js
+++ b/src/opnsense/www/js/opnsense_theme.js
@@ -26,19 +26,62 @@
  * ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
  * POSSIBILITY OF SUCH DAMAGE.
  */
-
-/* hook-in additional theme functionality */
+ 
+// hook-in additional theme functionality
 
 $(document).ready(function () {
-    // traverse loaded css files
-    var toggle_sidebar_loaded = false;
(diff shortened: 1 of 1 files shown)


opnsenseuser:
@fichtner I can remove the blanks in the evening, if you still give me the time. I am unfortunately working now and can only continue working in the evening and remove the blanks. 

best regards rené

fichtner:
Merged via  622ef1d, cleaned up whitespaces and comment style differences for a cleaner diff.

First things first: this solution is very practical, many thanks for working on it! More on this in the parent topic in a bit...

Small note on comments: /* */ to me are usually persistent code comments and documentation and // are temporary comments and disabled code lines that are being worked on. I've left all previous styles in tact, but the file uses a mixed style so I'd like to mention it.

Getting the cleanest possible diff helps others to review, understand and review the changes more effectively. :)


Thanks,
Franco

opnsenseuser:
@fichtner Naturally . I'm sorry, it actually wanted to make it better. but now understand it and will do it the future as you say. thanks, best regards, rene

Issue #2766, opened 2018-09-28
Lobby - Dashboard - Gateways / Gateway Status Unknown


TheNec:
Hello,

since I updated to 18.7.3 while I didn't change anything else the dashboard shows an unknown gateway status. The problem persists even after the update to 18.7.4:

Name WAN_PPPOE
          ~
RTT ~
RTTd ~
Loss ~
Status [Unknown]

Under System - Gateways - Single everything seems to be fine:

Name WAN_PPPOE (default)
Interface WAN
Gateway [External Gateway IP Address]
Monitor IP [Monitor IP]
RTT 17.9ms
RTTd 0.3 ms
Loss 0.0%
Status [Online]
Description Interface WAN_PPPOE Gateway

So perhaps this is just a small read/write error and not really a problem for me. Or was something changed with gateways and version >= 18.7.3 that you need to adjust any settings?

The behavior is language and apinger/dpinger independent while I am using the LibreSSL flavour.

If you need any futher information please let me know.

Best regards from Germany :)

fichtner:
You're using PPPoE on WAN, but monitoring IPv4 or IPv6? Or is it for both?

There was a gateway widget update in 18.7.3 so that's probably the issue, not the data underneath indeed.

Greetings from Leipzig :)

TheNec:
Hallo Franco :)

I am using PPPoE IPv4 only but with vlan7 (Deutsches Telekom / Interface: pppoe0 (re2_vlan7) ) while using the usually online redirector.heise.de [193.99.144.80] as an alternative monitor IP xD (btw. Spelling Error/Schreibfehler: Monitor IP Help/Hilfe nitcht -> nicht). If I delete the IP and save it it shows status [offline] in gateway settings because it is not pingable but still [unknown] in dashboard widget. Same after adding the IP again: [online] in settings, [unknown] in dashboard.

As I understood correctly the widget update (should) only allow(s) (you) to hide and unhide single/multiple gateways from the widget but there is only one gateway I have and that is shown. Hiding and unhiding doesn't change anything.

Perhaps the probem is the name of the gateway: [WAN_PPPOE (default)] under system settings but [WAN_PPPOE] without (default) in the widget or something like that.

Schöne Grüße aus Karlsruhe und schönes Wochenende ;)

fichtner:
@TheNec hi, can you try this patch?

    # opnsense-patch f394fed

TheNec:
Hi Franco,

[...]

PR #2868, opened 2018-11-02
Lobby - Dashboard - Gateways / Gateway Status Unknown  #2766

NOYB:
Gateways Widget Status Unknown Fix

### DIFF
--- a/src/www/widgets/widgets/gateways.widget.php
+++ b/src/www/widgets/widgets/gateways.widget.php
@@ -58,9 +58,10 @@ $gateways = return_gateways_array();
 <script>
   function gateways_widget_update(sender, data)
   {
+      var tbody = sender.find('tbody');
       data.map(function(gateway) {
           var tr_id = "gateways_widget_gw_" + gateway['name'];
-          if (find("#"+tr_id).length != 0) {
+          if (tbody.find("#"+tr_id).length != 0) {
               $("#"+tr_id+" > td:eq(0)").html('<small><strong>'+gateway['name']+'</strong><br/>'+gateway['address']+'</small>');
               $("#"+tr_id+" > td:eq(1)").html(gateway['delay']);
 <?php if (isset($config['system']['prefer_dpinger'])): ?>


src/www/widgets/widgets/gateways.widget.php
```diff
@@ -58,9 +58,10 @@
 <script>
   function gateways_widget_update(sender, data)
   {
+      var tbody = sender.find('tbody');
       data.map(function(gateway) {
           var tr_id = "gateways_widget_gw_" + gateway['name'];
-          if (find("#"+tr_id).length != 0) {
+          if (tbody.find("#"+tr_id).length != 0) {
```
fabianfrz: for me, this looks like a workaround for a bug (id is not unique), which would be a better idea to fix.

fichtner: maybe a browser-specific fix. we don't have tbody in the original file. maybe find without an object misses the point of searching for a html element

fabianfrz: @fichtner the tbody may be automatically generated by the browser (I have seen that multiple times in the developer tools). An id must be unique in a document by the HTML specification.

fichtner: yep, and this can fail when you use child selectors. IDs shouldn't be duplicated here, it just misses finding any ID so status stays at "~". sender.find("#"+tr_id) should work too?!

NOYB:
Don't consider correct syntax as being a browser specific fix.
  For Chrome there may as well not be a test because it will always run the code block even when it shouldn't.

  if (find("#"+tr_id).length != 0) { code block }
    find("#"+tr_id).length    returns undefined
    tr id exists:             the code block runs
    tr id NOT exists:         the code block runs

  For Edge the code block may as well be null because it will never run.

  With the correct syntax both browsers work correctly as expected.
  With the incorrect syntax each browser does it's own thing.

  Only reason the status happens to be populated with Chrome is because the failure mode is to execute the code block anyway.  Thus it is broken for both Chrome and Edge.  Failure mode is just different.  Correct syntax will fix for both Chrome and Edge.

NOYB:
Even sender.find("#"+tr_id) is overkill.  if ("#"+tr_id) should work just as well.  All it needs is to know whether or not the id exists.
BTW.  Same code (with tbody) is in interface_list widget.

PR #2872, opened 2018-11-03
sidebar - fix for some problems with sticky sub-submenu

opnsenseuser:
@fichtner - sidebar - fix for some problems with sticky sub-submenu


fichtner:
#2873

PR #2873, opened 2018-11-03
Sidebar   fix for some problems with sticky sub submenu

fabianfrz:
cleaned up #2872, but I am unable to merge it back / create a pull request to rene´s repository.

@opnsenseuser please check if the memory issue still applies. I've run auto format on the code, so it is cleaner and I removed some redundant code.

### DIFF
--- a/src/opnsense/www/js/opnsense_theme.js
+++ b/src/opnsense/www/js/opnsense_theme.js
@@ -30,166 +30,177 @@
 /* hook-in additional theme functionality */
 
 $(document).ready(function () {
-    // traverse loaded css files
-    var toggle_sidebar_loaded = false;
-    var winheight = $(window).height();
-    var nomouse = "mouseenter mouseleave";
-    var layer1_a = $('#mainmenu > div > a');
-    var layer1_div = $('#mainmenu > div > div');
-    var layer2_a = $('#mainmenu > div > div > a');
-    var layer2_div = $('#mainmenu > div > div > div');
-    var events = {
-        mouseenter: function() {
-            $("#navigation.col-sidebar-left").css("width", "415px");
-            if ($(this).next("div").hasClass("in")) {
-                /* no action needed */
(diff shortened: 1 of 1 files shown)


opnsenseuser:
@fabianfrz 
@fichtner 
unfortunately no real difference. I leave it as I wrote it. Then I understand it easier and can understand it a bit easier ;-)
I hope you are not angry with me. :-(
In any case very interesting how to write jquery optimized.
But thanks Fabian for your support. I suspect that it depends heavily on which side I am in opnsense. on the dashboard I have 6 columns active, this is also the traffic widget, etc. I also try to provoke the navigation to the utmost during the test and drive with the mouse wildly and very fast over all subemnüs up and down. I suspect there are several things that come together. performance number of widgets, etc.
Anyway ... I would like to thank you again for your support and wish you a nice weekend. best regards rené

opnsenseuser:
@fabianfrz 
@fichtner 

I have now tested further and found the following:
the reason for the load is not on the sidebar but on the real-time data of the individual widgets such as cpu, traffic, etc. all that gives real-time data affects the sidebar but also the normal navigation quite negative. Of course I never noticed 6 columns full of widgets before. but all users who use only weak cpu's will probably never be able to benefit from the full scope of opnsense without sacrificing significant performance.

In my opinion, would have to partially optimize the reporting, the traffic, cpu widget to get here even with low cpu performance appealing results.

reagrds rené

fichtner:
merged via 42beaef3aa464c8, thanks!

opnsenseuser:
Thx Franco for mergin
Would be also great if you could merge the tukan and cicada mods. They are also important for the sidebar mod
https://github.com/opnsense/plugins/pull/953

Thx, regards rene

Regards,
Rene

Sonntag, 4. November 2018, 18:13 +0100 von notifications  <notifications@github.com>:

PR #2877, opened 2018-11-04
sidebar - last entry in a submenu does not close the superordinate su…

opnsenseuser:
…bmenu (fix)

@fichtner sidebar - last entry in a submenu does not close the superordinate submenu (fix)

now its done !!!!

### DIFF
--- a/src/opnsense/www/js/opnsense_theme.js
+++ b/src/opnsense/www/js/opnsense_theme.js
@@ -46,7 +46,7 @@ $(document).ready(function () {
             var that = $(this);
             if (that.next("div").hasClass("in")) {
                 /* no action needed */
-            } else if (that.next().is("a")) {
+            } else if ((that.next().is("a")) || (that.is(".list-group-item:last-child"))) {
                 activeremove(this);
             } else {
                 var divtop = that.offset().top - $window.scrollTop();


mimugmail:
I installed latest master and tested with and without this commit. The services submenu is now cropped at the top .. see screenshot

image

opnsenseuser:
There is no way to Solveig this .there are to many Plugins in the Service. See my last post

opnsenseuser:
The routing entry in the Main Menu is away and so the Space to the top is smaller. 
Solution:
https://github.com/opnsense/core/issues/2878

opnsenseuser:
@fichtner this fix is for the problem with the last entry in a submenu. its important and has to be merged to get rid of this technical problem. 

But the problem of the cropped menu on top can´t be done better. when using the "routing" tab in the main menu it works. without routing tab the list of plugins is every time to long. 

@mimugmail screenshot of your last post with routing tab.
grafik


see the difference. solution https://github.com/opnsense/core/issues/2878

opnsenseuser:
@mimugmail 
without the "routing" tab you get up to 19 plugins without cropping.
with "routing" tab you get up to 23 plugins without cropping. 
-> you know what i mean?

mimugmail:
The previous screenshot was from a different test machine. I'd guess only 2% of all users will have the routing menu.

fichtner:
looks like we should make the routing menu the standard.... ;)

fichtner:
merged, thanks!

PR #2884, opened 2018-11-06
sidebar - change ".list-group-items:last-child" to "a:last-child"

opnsenseuser:
@fichtner small correction changing ".list-group-items:last-child" to "a:last-child"


fichtner:
merged, thanks!

PR #2887, opened 2018-11-08
Fix checking empty string.

kekek2:
If string is "0,1,2,3" result is "01,2,3"
empty($row[$fieldname]) return true if $row[$fieldname] is "0"


AdSchellevis:
@kekek2 thanks

PR #2902, opened 2018-11-12
Fix use of undefined constant where string was intended

MichaelDeciso:
function_exists() takes a string, but an undefined constant was passed. If a constant is undefined, PHP will convert it to a string, but this is very dirty behaviour.


fichtner:
nice catch, thanks!

PR #2903, opened 2018-11-12
Fix old-style constructor in PEAR.inc

MichaelDeciso:
Since PHP5, __construct() is preferred. As of PHP7, the old PHP4 style constructor is deprecated.

There are also a dozen occurrences in contrib, but I assume those are supposed to be fixed upstream.


fichtner:
Merged, thanks!

PR #2904, opened 2018-11-12
Clean up array usage in interfaces.php

MichaelDeciso:


src/www/interfaces.php
```diff
@@ -285,13 +285,11 @@ function get_wireless_modes($interface) {
...
 
-        $interface_channels = "";
+        $interface_channels = [];
         exec("$chan_list | $stack_list | sort -u | $format_list 2>&1", $interface_channels);
-        $interface_channel_count = count($interface_channels);
 
-        $c = 0;
-        while ($c < $interface_channel_count) {
-            $channel_line = explode(",", $interface_channels["$c"]);
+        for ($c = 0; $c < count($interface_channels); $c++) {
```
fichtner: Better make this something like foreach($interface_channels as $c => $interface_channel) and get rid of more fluff

MichaelDeciso: Ok, I have pushed a commit to do that.

fichtner:
Merged, thanks!

Issue #2892, opened 2018-11-10
Is Monit vulnerable to MITM?


NOYB:
Using SMTP server, either over SSL/TLS on port 465, or STARTTLS on port 587, messages go through even though there is no trusted CA for the mail server certificate.

This seems to indicate that the certificate/peer is not being validated.  Making it vulnerable to MITM.  Is this the case?  Or am I missing something?

Monit Secure Connection is enabled.  Packet capture bares this out.

fbrendel:
Yes, the SSL certificate verification is disabled by default. See Monit [SSL Options](https://mmonit.com/monit/documentation/monit.html#SSL-OPTIONS).
I'll add these options to the General Settings.

@fichtner should we enable the verification by default?

NOYB:
My philosophy would be for default secure so that it is required to make a conscious choice to be less secure.
Maybe a verify peer checkbox along side the secure connection enable checkbox.

Thanks

P.S.  Will Monit be using OpenSSL/LibreSSL or Certify?  Think OpenSSL/LibreSSL is more flexible for self signed certs frequently used with private mail servers.

fichtner:
Agree with default to on as it's easily turned off after update if required. We'll add this to the release notes. Monit follows ports crypto so that's OpenSSL or LibreSSL indeed.

fichtner:
thanks guys!

PR #2906, opened 2018-11-12
monit: add SSL options to mail server connection

fbrendel:
This PR adds two SSL options to the alert mail server connection.
One to select the SSL version to be used and the other to verify the server certificate.
See #2892

### DIFF
--- a/src/opnsense/mvc/app/models/OPNsense/Monit/Monit.xml
+++ b/src/opnsense/mvc/app/models/OPNsense/Monit/Monit.xml
@@ -1,6 +1,6 @@
 <model>
    <mount>//OPNsense/monit</mount>
-   <version>1.0.4</version>
+   <version>1.0.5</version>
    <description>Monit settings</description>
    <items>
       <general>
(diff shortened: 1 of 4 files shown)


NOYB:
A cursory check has the patch working here, and answered my question; uses OpenSSL/LibreSSL
Thanks

src/opnsense/mvc/app/models/OPNsense/Monit/Monit.xml
```diff
@@ -47,6 +47,23 @@
             <default>0</default>
             <Required>Y</Required>
          </ssl>
+         <sslversion type="OptionField">
+            <default>auto</default>
+            <Required>Y</Required>
+            <OptionValues>
+               <auto>AUTO</auto>
+               <sslv2>SSLV2</sslv2>
```
fichtner: can remove these, SSL2 and SSL3 are dead and hardwired to off in LibreSSL/OpenSSL

src/opnsense/mvc/app/models/OPNsense/Monit/Monit.xml
```diff
@@ -47,6 +47,23 @@
...
+               <sslv3>SSLV3</sslv3>
+               <tlsv1>TLSV1</tlsv1>
+               <tlsv11>TLSV11</tlsv11>
+               <tlsv12>TLSV12</tlsv12>
+               <tlsv13>TLSV13</tlsv13>
+            </OptionValues>
+         </sslversion>
+         <sslverify type="BooleanField">
+            <Required>N</Required>
```
fichtner: since it's easy to turn off after update it should be Required Y and default 1 for proper migration and security standard

PR #2907, opened 2018-11-12
De-PHP4 notices.smtp.inc

MichaelDeciso:
This class is older than the project, and had hardly anything done to it since.

This PR fixes the PHP4 warts, formatting some other odd stuff this class pulled. Commits are atomic, so reviewing should be easy if you check the diff for each commits, rather than the whole diff at once.


MichaelDeciso:
Since SMTP notices are going to go away anyway, I guess it's better to close this?

PR #2926, opened 2018-11-17
[WIP]: add password field to certmanager

mimugmail:
I thought this might be the way to go but I get some errors when creating a certificate:

```

openssl library returns: error:0E06D06C:configuration file routines:NCONF_get_string:no value
openssl library returns: error:0E06D06C:configuration file routines:NCONF_get_string:no value
openssl library returns: error:0E06D06C:configuration file routines:NCONF_get_string:no value
openssl library returns: error:0E06D06C:configuration file routines:NCONF_get_string:no value
```

I'm quite sure it's badly escaped or probably it's not the way to go, just an idea.

### DIFF
--- a/src/www/system_certmanager.php
+++ b/src/www/system_certmanager.php
@@ -192,7 +192,6 @@ if ($_SERVER['REQUEST_METHOD'] === 'GET') {
             if (!empty($pconfig['password'])) {
                 $res_pw = $pconfig['password'];
             }
-
             openssl_pkcs12_export($res_crt, $exp_data, $res_key, $res_pw, $args);
             $exp_size = strlen($exp_data);
             restore_error_handler();
@@ -420,6 +419,8 @@ if ($_SERVER['REQUEST_METHOD'] === 'GET') {
                 }
 
                 $cert['descr'] = $pconfig['descr'];
+
+                $cert['password'] = $pconfig['password'];
 
                 $old_err_level = error_reporting(0); /* otherwise openssl_ functions throw warings directly to a page screwing menu tab */
 
--- a/src/www/system_certmanager.php
+++ b/src/www/system_certmanager.php
@@ -127,6 +127,7 @@ if ($_SERVER['REQUEST_METHOD'] === 'GET') {
         $pconfig['cert_type'] = "usr_cert";
         $pconfig['cert'] = null;
         $pconfig['key'] = null;
+        $pconfig['password'] = null;
         $pconfig['dn_country'] = null;
         $pconfig['dn_state'] = null;
         $pconfig['dn_city'] = null;
@@ -187,8 +188,12 @@ if ($_SERVER['REQUEST_METHOD'] === 'GET') {
             $exp_data = "";
             $res_crt = openssl_x509_read(base64_decode($a_cert[$id]['crt']));
             $res_key = openssl_pkey_get_private(array(0 => base64_decode($a_cert[$id]['prv']) , 1 => ""));
+            $res_pw = null;
+            if (!empty($pconfig['password'])) {
+                $res_pw = $pconfig['password'];
+            }
 
-            openssl_pkcs12_export($res_crt, $exp_data, $res_key, null, $args);
+            openssl_pkcs12_export($res_crt, $exp_data, $res_key, $res_pw, $args);
             $exp_size = strlen($exp_data);
             restore_error_handler();
 
@@ -443,6 +448,7 @@ if ($_SERVER['REQUEST_METHOD'] === 'GET') {
                         $pconfig['caref'],
                         $pconfig['keylen'],
                         $pconfig['lifetime'],
+                        $pconfig['password'],
                         $dn,
                         $pconfig['digest_alg'],
                         $pconfig['cert_type']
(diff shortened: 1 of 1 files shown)


fichtner:
is this for exporting ?

mimugmail:
Yes, the PHP function defaults to null, thought this might work. My first journey to legacy code :)

fichtner:
the biggest issue here is the p12 export is a small button and then you'd have to type the password *twice* in order to reliably export a protected file so that requires a modal and further tweaking for the page POST behaviour...

fichtner:
You want to store the export password in the config? Per certificate or globally? It would be better to have a password prompt at export time. That’s why you want the input twice to verify the password is correct. See config export with password...

fabianfrz:
@fichtner or use a JS PRNG to generate a password in the web interface and print it in a modal afterwards - should work too.

mimugmail:
@fichtner at first I need a little success to motivate myself working with legacy stuff. :)
I'll look at the config export part, thx.

mimugmail:
Ok, now you have a Password field when creating a new certificate. It's stored in config.xml and when exporting p12 it'll take the one from config and set it correctly :)

src/www/system_certmanager.php
```diff
@@ -421,6 +420,8 @@ function () {
 
                 $cert['descr'] = $pconfig['descr'];
 
+                $cert['password'] = $pconfig['password'];
+
```
fabianfrz: you can also use an @ to suppress the warning if possible.

fichtner:
Okay, I like the idea to add a password input on the certificate details page, but storing it in the config is probably counter-productive in the long run. Why not remove the p12 export button from the list and add a small extra field for password input and download?

fabianfrz:
@fichtner I would use the bootstrap dialog for that. Just for linking the button to another JS function.

fichtner:
@fabianfrz yes, what I said here https://github.com/opnsense/core/pull/2926#issuecomment-439620550 but it's likely tricky for @mimugmail to pull this off just yet

fabianfrz:
@mimugmail can you try this for the dialog solution?

```diff
diff --git a/src/www/system_certmanager.php b/src/www/system_certmanager.php
index 43c06e1e2..50847680e 100644
--- a/src/www/system_certmanager.php
+++ b/src/www/system_certmanager.php
@@ -559,6 +559,39 @@ if (empty($act)) {
       });
     });
 
+    $('.p12btn').on('click', function(event) {
+        event.preventDefault();
+        var link = $(this).attr('href');
+        BootstrapDialog.show({
+            type:BootstrapDialog.TYPE_INFO,
+            title: "<?= gettext("Certificates");?>",
+            message: "<?=sprintf(gettext('If you want to protect the file with a password, please enter it here (leave empty for no password): %s'), '<input type=\\"text\\" name=\\"p12pw\\" id=\\"p12pw\\" />')?>",
+            buttons: [
+                {
+                    label: "<?=gettext("Close");?>",
+                    action: function(dialogRef) {
+                        dialogRef.close();
+                    }
+                }, {
+                    label: "<?=gettext("Download");?>",
+                    action: function(dialogRef) {
+                        dialogRef.close();
+                        $.post(link, "password=" + encodeURI($('#p12pw').val()) ,function (result)
+                        {
+                            var tmp=document.createElement('a');
+                            tmp.href=window.URL.createObjectURL(new Blob([result]));
+                            tmp.download="certificate.p12";
+                            tmp.style.display = 'none';
+                            document.body.append(tmp);
+                            tmp.click();
+                            tmp.remove();
+                        });
+                    }
+                }
+            ]
+        });
+    });
+
     $(".act_info").click(function(event){
         event.preventDefault();
         var id = $(this).data('id');
@@ -1305,7 +1338,7 @@ $( document ).ready(function() {
                     <i class="fa fa-download fa-fw"></i>
                   </a>
 
-                  <a href="system_certmanager.php?act=p12&amp;id=<?=$i;?>" class="btn btn-default btn-xs" data-toggle="tooltip" title="<?=gettext("export ca+user cert+user key in .p12 format");?>">
+                  <a href="system_certmanager.php?act=p12&amp;id=<?=$i;?>" class="btn btn-default btn-xs p12btn" data-toggle="tooltip" title="<?=gettext("export ca+user cert+user key in .p12 format");?>">
                       <i class="fa fa-download fa-fw"></i>
                   </a>
 <?php
```

fabianfrz:
@mimugmail FYI: this does not include the backend code.

mimugmail:
@fichtner tbh the only reason for static password is that this is the only way how I'd solve this with my little knowledge :) I'm using here a dedicated OPNsense instance just for PKI stuff and p12 password export is one of the things where I have to move it to a linux machine, extract and make a p12 again :) This could also just be treated as a PoC code to convince you this is a good feature to add. 

@fabianfrz seems to work, but I have no idea how to go on with this. For me it's totally fine when we close this one and go on with your approach :)

mimugmail:
Actually I have many open issues in plugins and no idea of Bootstrap code. If you think it's only a couple of minutes for you and you can afford them, I'd be more than happy to close my PR :)

AdSchellevis:
@mimugmail just leave it open, I'll look at the bootstrap code for you/

fabianfrz:
@AdSchellevis All the frontend code, that is required is already in my diff (just needs to be applied). It is only the backend code for handling the post request missing.

AdSchellevis:
@fabianfrz I'll glue it all together :)

AdSchellevis:
@mimugmail @fabianfrz here you go https://github.com/opnsense/core/commit/8a4c4614420f332222d6fbf006c17bd6a35f7765 can you test and confirm?

fabianfrz:
works

mimugmail:
yep, really nice :) Thanks! Also thanks to @fabianfrz for taking the time!

Issue #2848, opened 2018-10-26
[UI] bootstrap-select upgrade to v1.13.3


reza-ebrahimi:
The last upgrade for bootstrap-select is for v1.9.3, the current stable version is v1.13.3 that some critical bugs has been resolved.

https://github.com/snapappointments/bootstrap-select/releases/tag/v1.13.3

AdSchellevis:
This will likely have to wait for a major release, the style sheet changes don't seem to be compatible with the ones in our theme's now.

reza-ebrahimi:
And when we can have a major release?

AdSchellevis:
The next one is scheduled for Januari 2019, we'll try to add this.

AdSchellevis:
https://github.com/opnsense/core/pull/2939

PR #2929, opened 2018-11-18
jQuery 3 migration - some progress

fabianfrz:
use the jQuery 3 syntax (for many cases - this is not complete).
* In most cases the function name must be moved into a trigger, on or off call.
* Libraries are not (yet) compatible (for example tokenize, bootstrap select)
* Plugins need migration as well

### DIFF
--- a/src/opnsense/mvc/app/views/layouts/default.volt
+++ b/src/opnsense/mvc/app/views/layouts/default.volt
@@ -87,7 +87,7 @@
                     }
                 });
                 // hide submenu items
-                $('#mainmenu .list-group-item').click(function(){
+                $('#mainmenu .list-group-item').on('click', function(){
                     if($(this).attr('href').substring(0,1) == '#') {
                         $('#mainmenu .list-group-item').each(function(){
                             if ($(this).attr('aria-expanded') == 'true'  && $(this).data('parent') != '#mainmenu') {
@@ -151,11 +151,11 @@
                 });
 
                 // change search input size on focus() to fit results
-                $("#menu_search_box").focus(function(){
+                $("#menu_search_box").on('focus', function(){
                     $("#menu_search_box").css('width', '450px');
                     $("#menu_messages").hide();
                 });
-                $("#menu_search_box").focusout(function(){
+                $("#menu_search_box").on('focusout', function(){
                     $("#menu_search_box").css('width', '250px');
                     $("#menu_messages").show();
                 });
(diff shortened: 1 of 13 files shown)


AdSchellevis:
@fabianfrz I can't find a reference of `click()` (https://api.jquery.com/click/) being deprecated, same seems to be the case for `focus()` and `unbind()`, can you explain these changes?

To be honest, I personally like the shortcuts better, but maybe there's some design document I missed somewhere, we have been running JQuery 3 + JQuery migrate for quite some time now, and I haven't seen any deprecation messages there either.

fabianfrz:
@AdSchellevis the warnings are only shown if you don't use the .min.js variant. See:
https://github.com/jquery/jquery-migrate#development-vs-production-versions

And here is a reference:
https://github.com/jquery/jquery-migrate/blob/bf48bce49625e0c558c2f07dd514f2dfa5665334/warnings.md#jqmigrate-jqueryfnclick-event-shorthand-is-deprecated

AdSchellevis:
@fabianfrz thanks for explaining, I just find it odd that there seems to be no reference of these deprecations in the jQuery api docs.

We should might consider ditching jquery-migrate as well at some point, as it seems to be of minimal use at the moment.

fabianfrz:
@AdSchellevis we cannot. Some libraries won't work with it anymore.
See this screenshot for examples:

grafik

AdSchellevis:
@fabianfrz are you sure? the click() events seem to be working fine without the jquery-migrate import on my end.

fabianfrz:
@AdSchellevis yes but due to the deprecation I want to get rid of it - it is currently defined in the jQuery source: https://github.com/jquery/jquery/blob/master/src/event/alias.js#L10

There must be a reason why this message is printed and the warning is thrown.

AdSchellevis:
@fabianfrz to be honest, most libs just seem to be using these binds (including the latest bootstrap-select, which was requested here https://github.com/opnsense/core/issues/2848) and the docs from jQuery aren't in sync with their desired behaviour. I kind of expect it's unlikely this will go away soon.

Maybe it's better to wait until there is a version out there that doesn't support it and see other libraries pick this up as well. I don't want to be in the situation that we need to upstream these kind of patches to the rest of the world.

Your intentions are good, don't get me wrong, but I'm unsure we should try to fix this...

fabianfrz:
@AdSchellevis I've looked into it - the short syntax should have been deprecated with 3.3 but it looks like it has ben forgotten. I have created a ticket for clarification in the jQuery project - you can see the issue above this comment.

AdSchellevis:
@fabianfrz thanks, let's wait for the jQuery people then.

L1ghtn1ng:
@fabianfrz Just an FYI this branch has a conflict with src/opnsense/www/js/opnsense_theme.js

AdSchellevis:
3.4.1 had some shorthands deprecated (https://github.com/opnsense/core/commit/ac970161058b1d7944284434bce909c0bb48fbdb), we better fix this when needed, a lot of code depends on the ones we use now (including some of our dependancies)

PR #2930, opened 2018-11-18
system: expose defaults for security-related sysctls

fraenki:
It took me some time to figure out that the default value for these sysctls is `0` on OPNsense. So I thought it would be a good idea to expose these defaults in the GUI and make it more obvious.


fichtner:
Merged, thanks!

PR #2931, opened 2018-11-19
Check IP Services

NOYB:
Allow other, public and private, check IP services to be used.

Features:
  * Interface specific configurable check IP services.  The service with matching interface and IP version, or the one that is enabled as the default, will be used by Dynamic DNS and RFC2136 services to check IP addresses.
  * Verify SSL peer option
  * Service backend test facility

services.inc changes overview:
  * The get_dyndns_ip function updated to support using check IP services based on service name or interface and IP version, and a default check IP service other than factory default.
  * Factory default check IP service as array from MVC Check IP model Factory_Default_Service class.
  * Facilitate verify SSL peer option
  * Facilitate authentication
  * Facilitate service specific regular expression address capture

### DIFF
--- a/src/etc/inc/services.inc
+++ b/src/etc/inc/services.inc
@@ -49,6 +49,10 @@ require_once('plugins.inc.d/openssh.inc');
 require_once('plugins.inc.d/openvpn.inc');
 require_once('plugins.inc.d/unbound.inc');
 
+//use \OPNsense\DynDNS\Service;
+//use \OPNsense\DynDNS\Factory_Default_Service;
+use \OPNsense\DynDNS\CheckIP;
+
 function generate_ipv6_from_mac($mac)
 {
     $elements = explode(":", $mac);
@@ -1705,7 +1709,7 @@ function services_dhcrelay6_configure($verbose = false)
     }
 }
 
-function get_dyndns_ip($int, $ipver = 4)
+function get_dyndns_ip($int, $ipver = 4, $service = null)
 {
     global $config;
 
@@ -1729,17 +1733,54 @@ function get_dyndns_ip($int, $ipver = 4)
     }
 
     if ($ipver != 6 && is_private_ip($ip_address)) {
-        /* Chinese alternative is http://ip.3322.net/ */
-        $hosttocheck = 'http://checkip.dyndns.org';
+
+        // Get the check IP services.
+//        $mdlCheckIPService = new Service();
+        $mdlCheckIPService = new CheckIP();
+        $a_checkipservices = $mdlCheckIPService->service->getNodes();
+
+        // Append the factory default check IP service to the list.
+//        $mdlCheckIPFDS = new Factory_Default_Service();
+        $mdlCheckIPFDS = new CheckIP();
+        $a_checkipservices['FDS'] = $mdlCheckIPFDS->factory_default_service->getNodes();
+
+        // Is the factory default service disabled?
+        if ($mdlCheckIPService->disable_factory_default_service->__toString() == "1") {
+            $a_checkipservices['FDS']['default'] = 0;
+        }
+
+        /* Use the service specified by name. /*
+        /* Else use the first service found with a matching interface and IP version assignment. /*
+        /* Else use the enabled default service. */
+        foreach ($a_checkipservices as $i => $checkipservice) {
+            if ($checkipservice['name'] == $service || 
+               (!$service && 
+                $checkipservice['interfaces'][$int]['selected'] && 
+                $checkipservice['ipv']['ipv'.$ipver]['selected'])) {
+                    $found = true;
+            }
+            if ($found || $checkipservice['default']) {
+                $hosttocheck = $checkipservice['url'];
+                $regex = $checkipservice['regex'];
+                $username = $checkipservice['username'];
+                $password = $checkipservice['password'];
+                $verifysslpeer = $checkipservice['verifysslpeer'] ? true : false;
+                if ($found) { break; }
+            }
+        }
+
         $ip_ch = curl_init($hosttocheck);
         curl_setopt($ip_ch, CURLOPT_RETURNTRANSFER, 1);
+        curl_setopt($ip_ch, CURLOPT_SSL_VERIFYPEER, $verifysslpeer);
         curl_setopt($ip_ch, CURLOPT_INTERFACE, $ip_address);
         curl_setopt($ip_ch, CURLOPT_CONNECTTIMEOUT, 5);
         curl_setopt($ip_ch, CURLOPT_TIMEOUT, 30);
         curl_setopt($ip_ch, CURLOPT_IPRESOLVE, CURL_IPRESOLVE_V4);
+        curl_setopt($ip_ch, CURLOPT_HTTPAUTH, CURLAUTH_ANY);
+        curl_setopt($ip_ch, CURLOPT_USERPWD, "{$username}:{$password}");
         $ip_result = curl_exec($ip_ch);
         if ($ip_result !== false) {
-            preg_match('=<body>Current IP Address: (.*)</body>=siU', $ip_result, $matches);
+            preg_match('=' . $regex . '=siU', $ip_result, $matches);
             $ip_address = trim($matches[1]);
         } else {
             log_error('Aborted IPv4 detection: ' . curl_error($ip_ch));
--- a/src/opnsense/mvc/app/controllers/OPNsense/DynDNS/Api/CheckipserviceController.php
+++ b/src/opnsense/mvc/app/controllers/OPNsense/DynDNS/Api/CheckipserviceController.php
@@ -1,0 +1,65 @@
+<?php
+/**
+ *    Copyright (C) 2015 Deciso B.V.
+ *
+ *    All rights reserved.
+ *
+ *    Redistribution and use in source and binary forms, with or without
+ *    modification, are permitted provided that the following conditions are met:
+ *
+ *    1. Redistributions of source code must retain the above copyright notice,
+ *       this list of conditions and the following disclaimer.
+ *
+ *    2. Redistributions in binary form must reproduce the above copyright
+ *       notice, this list of conditions and the following disclaimer in the
+ *       documentation and/or other materials provided with the distribution.
+ *
+ *    THIS SOFTWARE IS PROVIDED ``AS IS'' AND ANY EXPRESS OR IMPLIED WARRANTIES,
+ *    INCLUDING, BUT NOT LIMITED TO, THE IMPLIED WARRANTIES OF MERCHANTABILITY
+ *    AND FITNESS FOR A PARTICULAR PURPOSE ARE DISCLAIMED. IN NO EVENT SHALL THE
+ *    AUTHOR BE LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY,
+ *    OR CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF
+ *    SUBSTITUTE GOODS OR SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS
+ *    INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN
+ *    CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE)
+ *    ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
+ *    POSSIBILITY OF SUCH DAMAGE.
+ *
+ */
+//namespace OPNsense\CheckIP\Api;
+namespace OPNsense\DynDNS\Api;
+
+use \OPNsense\Base\ApiControllerBase;
+use \OPNsense\Core\Backend;
+
+/**
+ * Class ServiceController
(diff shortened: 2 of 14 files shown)


AdSchellevis:
Can you move this over to `OPNsense\DynDNS\CheckIP`? This service is specifically for dyndns if I understand correctly, since there are changes needed in services.inc I understand that a plugin is not an option in this case, otherwise it would have been better to start with a plugin first.

I'll try to read the rest of the code at a later moment, this was just a small thing I noticed directly.

src/opnsense/mvc/app/controllers/OPNsense/CheckIP/Api/SettingsController.php
```diff
@@ -0,0 +1,277 @@
...
+    {
+        $mdlCheckIPFDS = new Factory_Default_Service();
+
+        $fds = $mdlCheckIPFDS->factory_default_service->getNodes();
+        $fds['default'] = ($mdlCheckIPService->disable_factory_default_service->__toString() == "1") ? "0" : "1";
+        $fds['uuid'] = 'FDS';
+
+        $grid['rows'][] = $fds;
+        $grid['rowCount'] = $grid['rowCount'] + 1;
+        $grid['total'] = $grid['total'] + 1;
```
fabianfrz: please note that this looks like the response expected by bootgrid.
This contains some data to query.

src/opnsense/mvc/app/controllers/OPNsense/CheckIP/Api/TestController.php
```diff
@@ -0,0 +1,99 @@
...
+            $nodes = $this->includeFDS($mdlCheckIPService, $nodes);
+
+            // Populate the service options.
+            if (is_array($nodes)) {
+                foreach ($nodes as $nodeUuid => $node) {
+                    $servicename = $node['name'];
+                    $selected = ($node['default'] == "1") ? 1 : 0;
+
+                    $suffix1 = ($nodeUuid == "FDS") ? " (FDS)" : "";
+                    $suffix2 = ($node['default'] == "1") ? " (default)" : "";
```
fabianfrz: probably used in UI -> translate

src/opnsense/mvc/app/views/OPNsense/CheckIP/index.volt
```diff
@@ -0,0 +1,209 @@
...
+
+  // Determine if at least one service is enabled as the default.  Warn if not.
+  function is_service_default() {
+    ajaxCall(url="/api/checkip/settings/is_default/service/", sendData={}, callback=function(data,status) {
+      if (status == 'success' && data['result']) {
+         $("#responseMsg").css({'background-color': '', 'color': ''});
+         $("#responseMsg").addClass('hidden');
+         $("#responseMsg").html('');
+      } else {
+         $("#responseMsg").html('<b>WARNING: </b>There is no default check IP service enabled!');
```
fabianfrz: translation missing

src/opnsense/mvc/app/views/OPNsense/CheckIP/index.volt
```diff
@@ -0,0 +1,209 @@
...
+  function dom_update() {
+    // suppress FDS selection checkbox and command buttons, and keep unselected
+    $('table#grid-services tbody tr[data-row-id="FDS"]').each(function() {
+      $(this).find('td.select-cell input.select-box').addClass('hidden');
+      $(this).find('td button').filter('.command-edit,.command-copy,.command-delete').addClass('hidden');
+/**/
+      $(this).find('td.select-cell input.select-box').prop('checked', false);
+      $(this).attr('aria-selected', false);
+      $(this).removeClass('active');
+/**/
```
fabianfrz: this empty comments are probably useless

src/opnsense/mvc/app/views/OPNsense/CheckIP/index.volt
```diff
@@ -0,0 +1,209 @@
...
+    });
+
+    // suppress fa-times icon for verify ssl peer disabled
+    $('table#grid-services tbody tr td:nth-child(5) span').removeClass('fa-times');
```
fabianfrz: this selector is probably not very stable (nth-child) because the content may change and there will be some strange behaviour then.

src/opnsense/mvc/app/views/OPNsense/CheckIP/index.volt
```diff
@@ -0,0 +1,209 @@
...
+
+    // hyper-link each service URL
+    $('table#grid-services tbody tr td:nth-child(4)').each(function() {
```
fabianfrz: same here

src/opnsense/mvc/app/views/OPNsense/CheckIP/index.volt
```diff
@@ -0,0 +1,209 @@
...
+      $(this).html('<a target="Check_IP" rel="noopener noreferrer" href="'+$(this).text()+'">'+$(this).text()+'</a>');
```
fabianfrz: this looks quite strange to have href and visible text the same

src/opnsense/mvc/app/views/OPNsense/CheckIP/index.volt
```diff
@@ -0,0 +1,209 @@
...
+    mapDataToFormUI(data_get_map).done(function(data){
+      // place actions to run after load, for example update form styles.
+        formatTokenizersUI();
+        $('.selectpicker').selectpicker('refresh');
+        // request service status on load and update status box
+        updateServiceControlUI('checkip');
+    });
+  });
+
+  $('#btn_test').unbind('click').click(function(){
```
fabianfrz: this function is deprecated (v3.0), use off instead: https://api.jquery.com/unbind/

src/opnsense/mvc/app/views/OPNsense/CheckIP/index.volt
```diff
@@ -0,0 +1,209 @@
...
+  observer.observe(targetNode, config);
+
+  $('div#services').removeClass('hidden');
+
+
+  /**
+   * service test
+   */
+
+  $('[data-toggle][href=#test]').click(function(){
```
fabianfrz: there is currently some inconsistency if the short syntax is deprecated. There are two docs telling something different. See  #2929 for details

src/opnsense/scripts/OPNsense/CheckIP/service_test.php
```diff
@@ -0,0 +1,43 @@
...
+
+  $service = (isset($argv[1])) ? trim($argv[1], " \n")    : '';
+$interface = (isset($argv[2])) ? trim($argv[2], " \n")    : 'wan';
+      $ipv = (isset($argv[3])) ? trim($argv[3], "ipv \n") : 4;
+
+echo      "Service: " . ($service ? $service : 'selected by interface and IP version') . 
+      ", Interface: " . $interface . 
+     ", IP Version: " . $ipv . 
+     ", IP Address: " . get_dyndns_ip($interface, $ipv, $service);
+?>
```
fabianfrz: when returning structured data, we usually use JSON.

NOYB:
@AdSchellevis
The services.inc get_dyndns_ip() function, i.e. check IP, is used by Dynamic DNS and also RFC2136 entries that have the "use public IP" option enabled.  Both the Dynamic DNS and RFC2136 services are plug-ins.  Not clear on what you mean by move over to OPNsense\DynDNS\CheckIP.

src/opnsense/mvc/app/views/OPNsense/CheckIP/index.volt
NOYB: Looks are not everything.  ;)
Provides browsing link to service for view response for making address capture regex and for troubleshooting and whatnot.

src/opnsense/scripts/OPNsense/CheckIP/service_test.php
NOYB: Changed to JSON

AdSchellevis:
@NOYB I mean rename namespaces to DynDNS, like:

**Controllers:**
```
src/opnsense/mvc/app/controllers/OPNsense/CheckIP/Api/ServiceController.php  -->  src/opnsense/mvc/app/controllers/OPNsense/DynDNS/Api/CheckIPServiceController.php
src/opnsense/mvc/app/controllers/OPNsense/CheckIP/Api/SettingsController.php -->  src/opnsense/mvc/app/controllers/OPNsense/DynDNS/Api/CheckIPSettingsController.php
src/opnsense/mvc/app/controllers/OPNsense/CheckIP/Api/TestController.php -->  src/opnsense/mvc/app/controllers/OPNsense/DynDNS/Api/CheckIPTestController.php 
src/opnsense/mvc/app/controllers/OPNsense/CheckIP/IndexController.php --> src/opnsense/mvc/app/controllers/OPNsense/DynDNS/CheckIPController.php
src/opnsense/mvc/app/controllers/OPNsense/CheckIP/forms/service.xml --> src/opnsense/mvc/app/controllers/OPNsense/DynDNS/forms/checkip_service.xml
src/opnsense/mvc/app/controllers/OPNsense/CheckIP/forms/test.xml -->  src/opnsense/mvc/app/controllers/OPNsense/DynDNS/forms/checkip_test.xml
```

**Use a single model:**
```
[
src/opnsense/mvc/app/models/OPNsense/CheckIP/Factory_Default_Service.php
src/opnsense/mvc/app/models/OPNsense/CheckIP/Factory_Default_Service.xml
src/opnsense/mvc/app/models/OPNsense/CheckIP/Test.php
src/opnsense/mvc/app/models/OPNsense/CheckIP/Test.xml
src/opnsense/mvc/app/models/OPNsense/CheckIP/Service.php
] --> combine into
src/opnsense/mvc/app/models/OPNsense/DynDNS/CheckIP.php|xml
```

ACL's and Menu are generic for DynDNS in that case (we could move other items in there later):
```
src/opnsense/mvc/app/models/OPNsense/CheckIP/ACL/ACL.xml --> src/opnsense/mvc/app/models/OPNsense/DynDNS/ACL/ACL.xml
src/opnsense/mvc/app/models/OPNsense/CheckIP/Menu/Menu.xml --> src/opnsense/mvc/app/models/OPNsense/DynDNS/Menu/Menu.xml
src/opnsense/mvc/app/views/OPNsense/CheckIP/index.volt --> src/opnsense/mvc/app/views/OPNsense/DynDNS/checkip.volt
```

**Scripts the same, configd actions into dynamicdns, since dyndns is already taken (remember to name the calls logical, so you could  execute `dynamicdns checkip test`),** 
```
src/opnsense/scripts/OPNsense/CheckIP/service_test.php  --> src/opnsense/scripts/OPNsense/DynDNS/checkip_test.php
src/opnsense/service/conf/actions.d/actions_checkip.conf  -->  src/opnsense/service/conf/actions.d/actions_dynamicdns.conf
```

I haven't had time to look at the rest of the code yet, but it's probably easier to restructure this first and do another review then.

Eventually it might be better to move all dynamic dns stuff into a single plugin and think about how this can be hooked as well, but it doesn't hurt to put in on a logical spot first.

NOYB:
Using the single model how can I restrict a new object to just the service nodes?

So something like this to only work with the <service> nodes.
`$mdlCheckIPService = new CheckIP();`

CheckIP.xml:
```
<items>
<FDS>...</FDS>
<dFDS>...</dFDS>
<service>...</service>
<test>...</test>
</items>
```
Think what I'd like is a Service class in CheckIP.php.  To be able to do like this.
$mdlCheckIPService = new Service();

Trying to avoid having the non services static nodes such as FDS, dFDS and test from being stored in the config by this.
```
$mdlCheckIPService->serializeToConfig();
Config::getInstance()->save();
```

This is why I initially used separate models for them.

Thanks

AdSchellevis:
I'm not sure what you mean, the model isn't a service. If you only change part of a model, the rest will always be left unaltered. 

You can use sections in your model to distinct functionality, which all belong to the same component, for example like https://github.com/opnsense/core/blob/master/src/opnsense/mvc/app/models/OPNsense/IDS/IDS.xml or use different models (with their own mountpoints) for different functional blocks, like Michael does in Freeradius https://github.com/opnsense/plugins/tree/4368fa910ada7f32dd4b9ac2f6494ccdaf53a06e/net/freeradius/src/opnsense/mvc/app/models/OPNsense/Freeradius

Seeing the scope of the plugin, I would probably go for one model, which can be used a bit like this then:
```
$mdlCheckIP = new CheckIP();
$mdlCheckIP->service->my_data ...;
```

NOYB:
Misspoke about "disable_factory_default_service" in the previous comment.  It is a flag that is handled in the code.  Not in the model.  It is correctly stored in the config.  Updated previous comment to reflect this.

AdSchellevis:
Sorry, we don't have a lot of time available to review this, ideally all the dyndns code would be in a plugin (and using a new and more cleaned-up codebase), which is maintained outside of core. 

We have to be quite strict in which changes we accept in our core repository, since they might involve more than the target audience of the feature at hand. 

Looking briefly at the current changes, at a first glance they seem to be somewhat over engineered.
To avoid disappointments, I thought I rather mention this now, since there is a realistic change that we can't add this to core...

Small and manageable changes to both legacy plugins might be more realistic at the moment, if that's a possibility. We might consider to move `get_dyndns_ip` to both legacy plugins, since it doesn't seem to be used elsewhere.

NOYB:
For the case that both Dynamic DNS and RFC2136 plugins are installed, what would be the preferred means of avoiding get_dyndns_ip() redeclaration?
Stick it in an include file and require/include once?  If so preferred location?
An if !function_exists('get_dyndns_ip') block?
Other?

PR #2933, opened 2018-11-19
Add stats page for Unbound

MichaelDeciso:


src/opnsense/mvc/app/models/OPNsense/Unbound/Menu/Menu.xml
```diff
@@ -11,6 +11,7 @@
                 <All url="/services_unbound_acls.php*" visibility="hidden"/>
             </ACL>
             <LogFile VisibleName="Log File" order="50" url="/diag_logs_resolver.php"/>
```
fichtner: Log File should be last

src/opnsense/mvc/app/views/OPNsense/Unbound/stats.volt
```diff
@@ -0,0 +1,165 @@
...
+POSSIBILITY OF SUCH DAMAGE.
+
+#}
+<script>
+    'use strict';
+
+    const descriptionMapThread = {
+        'recursion': {
+            'time': {
+                'avg': 'Recursion time (average)',
```
fichtner: all of the values are missing translations like this:

```suggestion
                'avg': ''{{ lang._('Recursion time (average)') }}',
```

src/opnsense/mvc/app/views/OPNsense/Unbound/stats.volt
```diff
@@ -0,0 +1,165 @@
...
+                'median': 'Recursion time (median)'
+            }
+        },
+        'tcpusage': 'TCP usage',
+        'num': {
+            'queries_ip_ratelimited': 'Num IP ratelimited queries',
```
fichtner: "Num " prefix can be dropped from these two

src/opnsense/mvc/app/views/OPNsense/Unbound/stats.volt
```diff
@@ -0,0 +1,165 @@
...
+                    });
+                }
+            }
+        );
+    }
+
+    $(document).ready(function() {
+
+
+        $('#refresh').on('click', function() {
```
fichtner: any reason not to do an auto update ? it would increase UX, we don't have a static page to deal with :)

MichaelDeciso: I took inspiration from another stats page which didn't auto-update. That's the only reason, really, I think auto-updating is no problem.

How often do you think it should auto-update? (As in: once per x seconds/minutes?)

fichtner: somewhere between 10-30 seconds, probably on the lower half. whatever is good to read the values and understand them before they get flushed (e.g. copy values from the page before wiped again).

the firewall live log has a checkbox to stop auto-update... maybe that would satisfy all needs and the timeout can be as low as 10 seconds

MichaelDeciso: That sounded like a good plan so I implemented it. ^^

fichtner:
merged, thanks!

soder10:
Just a question: will it include extended stats / can it have the ability to trigger the extended stat collection? Services\Unbound GUI does not allow toggle on/off extended stats, so I had to add it as custom option. But I am not sure this stats query can handle these extra entries, or only the standard stats.

PR #2936, opened 2018-11-19
Fix Unbound stats autorefresh not working

MichaelDeciso:
Noticed this when checking the stats clearing problem. It would only update once.


fichtner:
Thanks 👍

PR #2938, opened 2018-11-19
sidebar - performance & optimisation

opnsenseuser:
@fichtner -> sidebar - performance & optimisation -> i hope this works -> notepad++ totally mess. my fault!


opnsenseuser:
@fichtner Do you think you will manage to get the opnsense-theme.js into the dev today or tomorrow? Because this update also contains a really important fix that fixes the calculation for "scroll position"!

best regards, rene

fichtner:
merged manually via 04df3de895a

PR #2943, opened 2018-11-20
Fix HTML errors

MichaelDeciso:
This fixes some typos, deprecated elements and the usage of the `alt` attribute on `<div>` elements (which is not allowed and does not work either).

### DIFF
--- a/src/etc/inc/interfaces.inc
+++ b/src/etc/inc/interfaces.inc
@@ -4357,7 +4357,7 @@ function get_interfaces_info()
         if ($bridge) {
             $bridge_text = `/sbin/ifconfig {$bridge}`;
             if (stristr($bridge_text, "blocking") <> false) {
-                $ifinfo['bridge'] = "<b><font color='red'>" . gettext("blocking") . "</font></b> - " . gettext("check for ethernet loops");
+                $ifinfo['bridge'] = "<b><span style='color: red'>" . gettext("blocking") . "</span></b> - " . gettext("check for ethernet loops");
                 $ifinfo['bridgeint'] = $bridge;
             } elseif (stristr($bridge_text, "learning") <> false) {
                 $ifinfo['bridge'] = gettext("learning");
--- a/src/etc/inc/interfaces.inc
+++ b/src/etc/inc/interfaces.inc
@@ -4357,7 +4357,7 @@ function get_interfaces_info()
         if ($bridge) {
             $bridge_text = `/sbin/ifconfig {$bridge}`;
             if (stristr($bridge_text, "blocking") <> false) {
-                $ifinfo['bridge'] = "<b><span style='color: red'>" . gettext("blocking") . "</span></b> - " . gettext("check for ethernet loops");
+                $ifinfo['bridge'] = "<b><span class='text-danger'>" . gettext("blocking") . "</span></b> - " . gettext("check for ethernet loops");
                 $ifinfo['bridgeint'] = $bridge;
             } elseif (stristr($bridge_text, "learning") <> false) {
                 $ifinfo['bridge'] = gettext("learning");
--- a/src/www/diag_ipsec_spd.php
+++ b/src/www/diag_ipsec_spd.php
@@ -73,9 +73,9 @@ legacy_html_escape_form_data($spd);
                 <tr>
                   <td colspan="2"></td>
                   <td colspan="3">
-                    <span class="fa fa-arrow-right fa-fw" aria-hidden="true" alt="in"></span>
+                    <span class="fa fa-arrow-right fa-fw" aria-hidden="true"></span>
                     <?= gettext("incoming (as seen by firewall)"); ?> <br/>
-                    <span class="fa fa-arrow-left fa-fw" aria-hidden="true" alt="out"></span>
+                    <span class="fa fa-arrow-left fa-fw" aria-hidden="true"></span>
                     <?= gettext("outgoing (as seen by firewall)"); ?>
                   </td>
                 </tr>
--- a/src/www/diag_limiter_info.php
+++ b/src/www/diag_limiter_info.php
@@ -68,7 +68,7 @@ include("head.inc");
       });
   }
   function activitycallback(transport) {
-    jQuery('#limiteractivitydiv').html('<font face="Courier" size="2"><pre style="text-align:left;">' + transport.responseText  + '<\/pre><\/font>');
+    jQuery('#limiteractivitydiv').html('<pre class="text-left">' + transport.responseText  + '</pre>');
     setTimeout('getlimiteractivity()', 2000);
   }
   setTimeout('getlimiteractivity()', 1000);
--- a/src/www/services_dhcp.php
+++ b/src/www/services_dhcp.php
@@ -1024,7 +1024,7 @@ include("head.inc");
                             foreach($numberoptions as $item):?>
                               <tr>
                                 <td>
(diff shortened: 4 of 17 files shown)


src/www/diag_ipsec_spd.php
```diff
@@ -73,9 +73,9 @@
                 <tr>
                   <td colspan="2"></td>
                   <td colspan="3">
-                    <span class="fa fa-arrow-right fa-fw" aria-hidden="true" alt="in"></span>
+                    <span class="fa fa-arrow-right fa-fw" aria-hidden="true"></span>
                     <?= gettext("incoming (as seen by firewall)"); ?> <br/>
-                    <span class="fa fa-arrow-left fa-fw" aria-hidden="true" alt="out"></span>
+                    <span class="fa fa-arrow-left fa-fw" aria-hidden="true"></span>
```
fabianfrz: Probably title was meant -> creates a tooltip

src/www/diag_limiter_info.php
```diff
@@ -68,7 +68,7 @@ function getlimiteractivity() {
       });
   }
   function activitycallback(transport) {
-    jQuery('#limiteractivitydiv').html('<font face="Courier" size="2"><pre style="text-align:left;">' + transport.responseText  + '<\/pre><\/font>');
+    jQuery('#limiteractivitydiv').html('<pre style="text-align:left;">' + transport.responseText  + '<\/pre>');
```
fabianfrz: The backslash is probably not needed and the pre would make more sense when it is not generated in js if possible

src/www/services_dhcp.php
```diff
@@ -1071,7 +1071,7 @@ function removeRow() {
                             <tfoot>
                               <tr>
                                 <td colspan="4">
-                                  <div id="addNew" style="cursor:pointer;" class="btn btn-default btn-xs" alt="add"><i class="fa fa-plus fa-fw"></i></div>
+                                  <div id="addNew" style="cursor:pointer;" class="btn btn-default btn-xs"><i class="fa fa-plus fa-fw"></i></div>
```
fabianfrz: Very likely title was meant - also gettext is missing

src/www/wizard.php
```diff
@@ -885,7 +885,7 @@ function showchange() {
 				echo $field['typehint'];
 			}
 			if($field['warning'] <> "") {
-				echo "<br /><b><font color=\"red\">" . $field['warning'] . "</font></b>";
+				echo "<br /><b><span style='color: red;'>" . $field['warning'] . "</span></b>";
```
fabianfrz: Use a suitable bootstrap class instead

src/www/diag_ipsec_spd.php
MichaelDeciso: I know `title` creates a tooltip, but I didn't think one was intended - it's all in lower case.
Of course, I could add one now - but what shall I take as text? And shall I add gettext as well?

src/www/diag_limiter_info.php
MichaelDeciso: I think you're right.

src/www/wizard.php
MichaelDeciso: The most suitable would be `text-warning`, though that would be orange, not red. Is that OK?

fichtner: for this case text-alert would be more appropriate despite the "warning" designation

MichaelDeciso: text-alert is not defined in any CSS file. Did you mean "text-info"?

fichtner: sorry, I meant text-danger

MichaelDeciso: Hm, would that really be more appropriate? I thought this field showed warnings that weren't that urgent? To me, it seems that classifying them as 'danger' would be a bit overkill.

fichtner: ok, let's back up a little :) how often is 'warning' used in the wizard xml files?

MichaelDeciso: I cannot find any occurrence in `wizard/system.xml` and `wizard/openvpn.xml`.

If you really think text-danger is more appropriate, I'll just go with that. :)

fichtner: if nothing uses it the point is moot. better remove the "functionality" ?

MichaelDeciso: You know the system better than I do. Are the two XML files I mentioned the only places this "warning" could come from?

I'll leave it to you to decide whether to remove it. :)

fichtner: yeah, only those two are left with the wizard system deprecated and in need of a better replacement eventually. please remove then, thanks!

src/etc/inc/interfaces.inc
```diff
@@ -4357,7 +4357,7 @@ function get_interfaces_info()
         if ($bridge) {
             $bridge_text = `/sbin/ifconfig {$bridge}`;
             if (stristr($bridge_text, "blocking") <> false) {
-                $ifinfo['bridge'] = "<b><font color='red'>" . gettext("blocking") . "</font></b> - " . gettext("check for ethernet loops");
+                $ifinfo['bridge'] = "<b><span style='color: red'>" . gettext("blocking") . "</span></b> - " . gettext("check for ethernet loops");
```
fichtner: how about this one in text-danger then? :D

MichaelDeciso: As you wish ;)

src/www/widgets/widgets/carp_status.widget.php
```diff
@@ -54,13 +54,13 @@
 <?php
       if (get_single_sysctl('net.inet.carp.allow') <= 0 ) {
           $status = gettext("DISABLED");
-          echo "<span class=\"fa fa-remove fa-fw text-danger\" title=\"$status\" alt=\"$status\" ></span>";
+          echo "<span class=\"fa fa-remove fa-fw text-danger\" title=\"$status\" ></span>";
```
fichtner: does this still display tooltips?

src/www/vpn_openvpn_server.php
```diff
@@ -1077,7 +1077,7 @@
                     </tr>
                     <tr>
                       <td style="width:22%" id="ipv4_tunnel_network"><a id="help_for_tunnel_network" href="#" class="showhelp"><i class="fa fa-info-circle"></i></a> <?=gettext("IPv4 Tunnel Network"); ?></td>
-                      <td witdh="78%">
+                      <td width="78%">
```
fichtner: fixed the remaining ones via https://github.com/opnsense/core/commit/1bfc8ef5df

src/www/widgets/widgets/carp_status.widget.php
MichaelDeciso: `title` is the attribute that displays tooltips. `alt` is only used when an img(!) cannot load (except in IE, which interprets the `alt` as `title` if no `title` is specified).

src/www/vpn_openvpn_server.php
fichtner: maybe this and the other one should be changed to style= as well?

$ git grep td.witdh
src/www/vpn_openvpn_server.php:                      <td witdh="78%">
$ git grep td.width
src/www/interfaces.php:                          <td width="22%"><i class="fa fa-info-circle text-muted"></i> <?=gettext('Service Provider') ?></td>

MichaelDeciso: Done.

fichtner:
merged, many thanks!

PR #2945, opened 2018-11-20
 Fix implicit Javascript global variable assignments 

MichaelDeciso:


fabianfrz:
Some could be `const` instead of `let`

MichaelDeciso:
You're right. I changed the ones that could be `const` to `const`.

MichaelDeciso:
This should be good to go now - addressed Fabian's point.

PR #2951, opened 2018-11-21
Fix some HTML errors

MichaelDeciso:
This fixes a duplicated closing tag and two instances of incorrect CSS usage to what was intended.


src/www/firewall_scrub_edit.php
```diff
@@ -422,7 +422,7 @@ function formNetworks()
                   <tr>
                     <td><a id="help_for_srcport" href="#" class="showhelp"><i class="fa fa-info-circle"></i></a> <?=gettext("Source port"); ?></td>
                     <td>
-                      <div style="display:hidden;">
+                      <div style="visibility:hidden;">
```
fichtner: better use display: none or class="hidden" which is more common in our code

MichaelDeciso: Fixed.

fichtner:
Merged, thanks!

PR #2952, opened 2018-11-21
PHP: Fix lots of refs to undefined vars

MichaelDeciso:


src/etc/inc/certs.inc
```diff
@@ -330,7 +330,7 @@ function cert_get_subject($str_crt, $decode = true)
         }
     }
 
-    return $subject;
+    return $subject ?? '';
```
fichtner: is that a clever "if not defined syntax"? I would prefer a $subject = ''; definition at the top

src/etc/inc/certs.inc
```diff
@@ -378,7 +378,7 @@ function cert_get_issuer($str_crt, $decode = true)
         }
     }
 
-    return $issuer;
+    return $issuer ?? '';
```
fichtner: same, unless the caller can cope with null on the other end

src/etc/inc/plugins.inc.d/dnsmasq.inc
```diff
@@ -169,6 +169,7 @@ function dnsmasq_configure_do($verbose = false)
         }
     }
 
+    $dns_rebind = '';
```
fichtner: why not use $args and move the if below to the top?

src/etc/inc/plugins.inc.d/unbound.inc
```diff
@@ -252,6 +252,7 @@ EOF;
     }
 
     // Set up forwarding if it configured
+    $forward_conf = '';
```
fichtner: also safe string interpolation later. the real WTF is that it's embedded and written even if empty, which forces an unwanted newline. but why not... but delete else { $forward_conf = ""; } ;)

src/etc/inc/services.inc
```diff
@@ -1746,7 +1746,7 @@ function get_dyndns_ip($int, $ipver = 4)
             $ip_address = '';
         }
         curl_close($ip_ch);
-    } elseif (ipver == 6 && is_linklocal($ip_address)) {
+    } elseif ($ipver == 6 && is_linklocal($ip_address)) {
```
fichtner: nice catch

src/www/diag_confbak.php
```diff
@@ -226,7 +226,7 @@
                   </td>
                   <td>
                     <?= gettext('Be aware of how much space is consumed by backups before adjusting this value.'); ?>
-                    <?php if (count($confvers)) {
+                    <?php if (isset($confvers) && count($confvers) > 0) {
```
fichtner: been over this recently and it's set in POST and else branch without an early way to skip so it looks like a false positive. Ideally I'd want this to break if it fails after a rework

src/www/wizard.php
```diff
@@ -172,6 +172,7 @@ function find_ip_interface($ip, $bits = null) {
 function update_config_field($field, $updatetext, $unset, $arraynum, $field_type) {
 	global $config;
 	$field_split = explode("->",$field);
+    $field_conv = '';
```
fichtner: indent in this file is probably still whacky tabs?

src/etc/inc/certs.inc
fabianfrz: https://secure.php.net/manual/en/language.operators.comparison.php#language.operators.comparison.coalesce

src/opnsense/mvc/app/library/OPNsense/Auth/AuthenticationFactory.php
```diff
@@ -126,7 +126,7 @@ public function get($authserver)
                 $localUserMap = $this->fetchUserDNs();
             }
 
-            if ($authObject != null) {
+            if (isset($authObject) && $authObject != null) {
```
fabianfrz: isset also checks for != null

src/www/wizard.php
MichaelDeciso: Yes :'(
What should I do with it? I can open a PR to reformat this file.

fichtner: sure, can do this after merging this one. wizard.php, system.xml and openvpn.xml (all wizard files) still use tabs. thanks!

PR #2954, opened 2018-11-22
Change tabs to spaces

MichaelDeciso:
As requested by @fichtner .

### DIFF
--- a/src/etc/inc/legacy_bindings.inc
+++ b/src/etc/inc/legacy_bindings.inc
@@ -34,8 +34,8 @@ require_once('script/load_phalcon.php');
  */
 function configd_run($cmd, $detach = false)
 {
-	$backend = new OPNsense\Core\Backend();
-	return $backend->configdRun("{$cmd}", $detach);
+    $backend = new OPNsense\Core\Backend();
+    return $backend->configdRun("{$cmd}", $detach);
 }
 
 /*************************************************************************************************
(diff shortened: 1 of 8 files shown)


src/etc/inc/gwlb.inc
```diff
@@ -33,15 +33,15 @@
 function return_dpinger_defaults()
```
fichtner: keep these... only leading tabs please (^\t)

src/etc/inc/system.inc
```diff
@@ -721,16 +721,16 @@ function system_syslogd_start($verbose = false, $restart = false)
     if (!isset($syslogcfg['disablelocallogging'])) {
```
fichtner: keep these, too

src/www/diag_backup.php
```diff
@@ -70,7 +70,7 @@ function restore_config_section($section_name, $new_contents)
 }
```
fichtner: keep :)

src/etc/inc/gwlb.inc
MichaelDeciso: That violates PSR-2, which says spaces should be used throughout. (And even most tab-intended code bases use spaces for aligning comments.)

MichaelDeciso: Ok, I will revert them for now.

I really hope the PHP files will be PSR-2 compliant sooner than later, though. I like the code base to be clean and consistent.

fichtner:
Thanks!

PR #2955, opened 2018-11-22
Fix many PHPdoc declarations and some function signatures

MichaelDeciso:


fichtner:
looks, good so will merge. one word of caution: there are subtle side-effects in this. best to batch PHPdocs fixes but make each signature change a commit to be able to revert it easily for users when necessary.

fichtner:
thanks!

PR #2957, opened 2018-11-22
fix advanced field bug in dialogs

fabianfrz:
copied that line from the base form - it makes advanced headers working in dialogs (base_form already has it).


AdSchellevis:
@fabianfrz thanks!

fabianfrz:
@AdSchellevis no problem - I've found it while working on the nginx plugin  because the headers kept stale.

PR #2966, opened 2018-11-23
Replace uncommon <> notation with !=

MichaelDeciso:
`!=` is way more familiar to most people than `<>`. This makes the code base more consistent and slightly easier to understand.

### DIFF
--- a/src/etc/inc/system.inc
+++ b/src/etc/inc/system.inc
@@ -190,7 +190,7 @@ function system_resolvconf_generate($verbose = false)
         $dnsgw = "dns{$dnscounter}gw";
         if (isset($syscfg[$dnsgw])) {
             $gwname = $syscfg[$dnsgw];
-            if (($gwname != "") && ($gwname != "none")) {
+            if (($gwname != '') && ($gwname != 'none')) {
                 $gatewayip = lookup_gateway_ip_by_name($gwname);
                 if (is_ipaddrv4($gatewayip)) {
                     /* dns server array starts at 0 */
--- a/src/etc/inc/system.inc
+++ b/src/etc/inc/system.inc
@@ -190,7 +190,7 @@ function system_resolvconf_generate($verbose = false)
         $dnsgw = "dns{$dnscounter}gw";
         if (isset($syscfg[$dnsgw])) {
             $gwname = $syscfg[$dnsgw];
-            if (($gwname <> "") && ($gwname <> "none")) {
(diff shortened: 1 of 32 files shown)


src/etc/inc/system.inc
```diff
@@ -190,7 +190,7 @@ function system_resolvconf_generate($verbose = false)
         $dnsgw = "dns{$dnscounter}gw";
         if (isset($syscfg[$dnsgw])) {
             $gwname = $syscfg[$dnsgw];
-            if (($gwname <> "") && ($gwname <> "none")) {
+            if (($gwname != "") && ($gwname != "none")) {
```
fichtner: taking this as an example, all lines touched should be further tweaked if possible. double quoted strings (") are slightly slower than single quoted ones ('), in all trivial cases a replacement is recommended (trivial = where no string interpolation or escaping like \n takes place)

MichaelDeciso: I don't really like touching too many unrelated things within a commit. Especially since the code base is full of these. Also, checking 100 separate places for single/double quotes is not really nice to do.

I'd rather tackle strings in another PR - I'd also be able to address more of them.

fichtner: I don't see a real argument against doing it only for the lines already touched here. At some point I'll have to veto cleanups because the code base is being treated with automated cleanups lacking functional testing. It's far better to clean up code while working on a particular subsystem to ensure integrity. As a hybrid approach, cleanup PRs could still be small and address multiple things in a self-contained way at the same time.

As a general cautionary tale, all well-intentioned clean ups can end in bugs eventually. We just don't know when and how they'll surface.

MichaelDeciso: I have done it, but not because I agree with you. This has cost more time than necessary, while only addressing a handful of trivial strings. It's especially jarring in wizard.php, where a find-and-replace throughout would have resulted in a cleaner file, while taking less time.

MichaelDeciso:
Since the commits are not really related, I strongly recommend that this gets merged using 'rebase and merge', rather than squashing it.

fichtner:
Thanks. I'm pushing some more cleanups now in the same files. I just want to make sure that automated search and replace is a nice tool, but is in no way a substitute for going the extra mile of manual review because of "cost more time than necessary". We are talking short term cost and missed opportunities to get to know the code vs. long term readability and knowing the code.

PR #2973, opened 2018-11-26
proxy: move bump_step1 down in squid.conf

mimugmail:
If sslurlonly is 0 `ssl_bump peek bump_step1 all` is listed twice.
Might be cosmetic only since it's working though ...


AdSchellevis:
@mimugmail nice catch, thanks!

PR #2974, opened 2018-11-26
resize function adapts to the sidebar height

opnsenseuser:
since the navigation bar can change in height at any time due to the installation of plugins, and since we may want to redistribute the plugins, I have decided to dynamically adapt the resize function for navigation. now it does not matter how many menu items are in the sidebar. the resize function of sidebar navigation is now dynamic.


opnsenseuser:
@fichtner the toggle button would now have to adapt dynamically too.. but this works only by js. we would have to remove the part from the .css file. what do you mean?

fichtner:
Nice and clean PR, thank you!

PR #2983, opened 2018-11-29
Fix undefined variable

MichaelDeciso:
Doing the quotation marks as well where we're at it, which hopefully pleases @fichtner ;)


fichtner:
\o/ thanks!

PR #2984, opened 2018-11-29
Replace some double quoted strings with single quotes ones

MichaelDeciso:
I have attempted to make sure that if I touched a line, all occurrences on that line were fixed. Hopefully, I didn't miss any.


fichtner:
Merged, thanks!

PR #2985, opened 2018-11-29
Clean up incorrect/unused variable assignments

MichaelDeciso:
This fixes several problems regarding variable assignments or initialisation. Most of these are variables that are declared but never used or declared but immediately overwritten. It also contains cases of an assignment operator instead of a compound string concatenate operator, and some cases of (nonworking) "C-style" variable declaration.

### DIFF
--- a/src/etc/inc/filter.inc
+++ b/src/etc/inc/filter.inc
@@ -323,8 +323,8 @@ function filter_configure_sync($verbose = false, $flush_states = false)
     }
 
     $natrules = "\n# NAT Redirects\n";
-    $natrules = "no nat proto carp all\n";
-    $natrules = "no rdr proto carp all\n";
+    $natrules .= "no nat proto carp all\n";
+    $natrules .= "no rdr proto carp all\n";
     $natrules .= $fw->outputNatRules();
 
     if ($verbose) {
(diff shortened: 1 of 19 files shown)


src/etc/inc/filter.inc
```diff
@@ -323,8 +323,8 @@ function filter_configure_sync($verbose = false, $flush_states = false)
     }
 
     $natrules = "\n# NAT Redirects\n";
-    $natrules = "no nat proto carp all\n";
-    $natrules = "no rdr proto carp all\n";
+    $natrules .= "no nat proto carp all\n";
```
fichtner: oO

MichaelDeciso: ... yeah, that was my reaction as well.

fichtner:
merged. thanks! I'll add the filter thingy directly to 18.7.9

MichaelDeciso:
Great!
Not sure if it's necessary, but the error was introduced in https://github.com/opnsense/core/commit/34e7a98df314c7e90f7e6fa87a48b613b5064c7c

fichtner:
Yeah, I remember it vividly, but didn't do great review apparently.

MichaelDeciso:
Oh, don't be too hard on yourself - it's very hard to spot.

PR #2997, opened 2018-12-03
Remove nonexistant CSS properties

MichaelDeciso:
The table attributes 'cellspacing' and 'cellpadding' were moved inside the style attribute via a script. However, they are not valid CSS properties, so browsers should (and do) ignore them.
This commit removes them. The 'table' class, set on most tables, should take care of proper formatting anyway.


fichtner:
Merged, thanks!

PR #2999, opened 2018-12-03
Fix idiosyncratic way of returning

MichaelDeciso:


fichtner:
ah, ok, horrible by design, I see. It would be nice to get rid of this return behaviour at some point (function &)

MichaelDeciso:
A test that Ad did showed that doing `return false` from this function _does_ work, even with error_reporting set to `E_ALL`.

Furthermore, the PHP manual has this to say:
```
    Note: Unlike parameter passing, here you have to use & in both places - to indicate that you want to return by reference, not a copy, and to indicate that reference binding, rather than usual assignment, should be done for $myValue. 
```
I checked all occurrences of this function, but none of them adheres to this. This means that they all create a regular variable, rather than a reference.

In addition, the stated reason it returns a reference in the first place is because the return value is sometimes passed into a function that takes an argument by reference. But that is not needed - you can create a regular variable and pass it into a function that takes a variable by reference without any issue:
```
function test(&$var)
{
    $var++;
}
$var = 2;
test($var);
var_dump($var);
```
Output:
```
int(3)
```

AdSchellevis:
No it doesn't...
```
PHP Notice:  Only variable references should be returned by reference in ...test.php on line 28
Notice: Only variable references should be returned by reference in ...test.php on line 28
```

http://php.net/manual/en/language.references.return.php

MichaelDeciso:
Which bit are you reacting to? The statement about your test? I didn't see any error output on your screen, so I must have misread it.

The rest still stands, though, especially the point about how this function is called in our code.

PR #3001, opened 2018-12-03
monit: Support declaring dependencies

awerner:
This PR allows setting dependencies in the UI. This allows advanced check behaviour, e.g. VPN <- hosts behind VPN <- service behind VPN on host.

At the moment, the selection of the service as its own dependency is possible, monit shows a dependency loop error in that case with a helpful message. I'm unsure how to restrict the available services to exclude itself from the dependency list.


fichtner:
@fbrendel thanks for review. version bump sounds good. Ideall both should hit 19.1, though I yield to @AdSchellevis on this topic

AdSchellevis:
@awerner thanks, I clarified the help message a bit, just let me know if it should be different.

PR #3008, opened 2018-12-04
Fix some JS errors

MichaelDeciso:
This most importantly involves duplicate case statement labels, declaration errors and empty if/catch blocks.

### DIFF
--- a/src/opnsense/www/js/opnsense.js
+++ b/src/opnsense/www/js/opnsense.js
@@ -200,17 +200,17 @@ function clearFormValidation(parent) {
  */
 function ajaxCall(url, sendData, callback) {
     return $.ajax({
-        type: "POST",
+        type: 'POST',
         url: url,
-        dataType:"json",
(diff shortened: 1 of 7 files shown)


src/opnsense/www/js/opnsense.js
```diff
@@ -205,12 +205,12 @@ function ajaxCall(url, sendData, callback) {
...
         contentType: "application/json",
         complete: function(data, status) {
-            if ( callback == null ) {
-                null;
-            } else if ( "responseJSON" in data ) {
-                callback(data['responseJSON'],status);
-            } else {
-                callback(data,status);
+            if ( callback != null ) {
+                if ( "responseJSON" in data ) {
```
fabianfrz: if you are already here, you could use single quotes here

MichaelDeciso: Done!

fichtner:
Merged, thanks!

PR #3009, opened 2018-12-04
Fix sorting of SIZE and RES columns in system activity

MichaelDeciso:
This required teaching jQuery-bootgrid about how to decode the data type.

Previously, it would treat the values as a string, sorting 1G before 10K, and sorting 100M before 11M. Both issues are now fixed.

Alternatively, we could add these data types and converters to our fork of jquery-bootgrid, if we expect to need them in other places.


AdSchellevis:
@MichaelDeciso looks good, thanks

PR #3014, opened 2018-12-05
Clean up ipsec_convert_to_modp()

MichaelDeciso:


src/etc/inc/plugins.inc.d/ipsec.inc
```diff
@@ -101,6 +101,26 @@ $p2_protos = array(
     'ah' => 'AH'
 );
 
+const DH_GROUP_TO_MODP = [
```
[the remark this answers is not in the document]
MichaelDeciso: Done!

src/etc/inc/plugins.inc.d/ipsec.inc
```diff
@@ -604,67 +604,33 @@ function ipsec_find_id(&$ph1ent, $side = 'local')
 }
 
 /* include all configuration functions */
-function ipsec_convert_to_modp($index)
+function ipsec_convert_to_modp(int $index): string
```
fichtner: does it make sense to force the int here? we likely get a string anyway as it comes from the config and it could produce errors?

MichaelDeciso: Since we don't declare strict mode, it will coerce the string to an int. `'30'` will be coerced to `30` with no error. The exception to this would be an empty string.

Considering the not-very-precise code in the rest of the file, it might be better to drop the type declarations. I did a test, and doing `array_key_exists('2', $map);` works as expected.

MichaelDeciso: Changed it.

fichtner:
Merged, thanks! ❤️

PR #3018, opened 2018-12-06
Fix #2960: Limit line length in log widget

### LINKED ISSUE #2960: system log widget - lines limiting?
in my opinion, the lines should be limited, because the user can anyway come at any time with a click in the detail view of the system logs. And so on the dashboard would also allow a much clearer presentation. and it would not turn out to be an endless line view as I have stated in my screenshot.

grafik

MichaelDeciso:
For consistency, I added the line char limit parameter to both `dump_clog` and `dump_log`.

### DIFF
--- a/src/www/diag_logs_common.inc
+++ b/src/www/diag_logs_common.inc
@@ -63,7 +63,16 @@ function print_dump($logarr)
     }
 }
 
-function dump_clog($logfile, $tail, $grepfor = '')
+function limit_log_line_length(string $line, int $lineCharLimit): string
+{
+    if (strlen($line) <= $lineCharLimit) {
+        return $line;
+    } else {
+        return substr($line, 0, $lineCharLimit - 1) . '…';
+    }
+}
+
+function dump_clog($logfile, $tail, $grepfor = '', int $lineCharLimit = 0)
 {
     global $config;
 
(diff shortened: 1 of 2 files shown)


opnsenseuser:
@MichaelDeciso 👍 

https://github.com/opnsense/core/issues/2960

MichaelDeciso:
@opnsenseuser Have you been able to test it? Does it work for you?

opnsenseuser:
@MichaelDeciso  will Test this in a few hours and will give you Feedback on this.

opnsenseuser:
@MichaelDeciso i can´t see any real difference . i just replaced the two files with your new ones. i also deleted the browser cache. but i can´t see any difference.

this is a short screenshot
widget_log_before

opnsenseuser:
@MichaelDeciso i think it would be better to Limit the line breaks and not the sum of the characters. But thats only my opinion. Regards rene

opnsenseuser:
@MichaelDeciso I have still come up with an idea. 
limit the lines and at the end of a sign with which you can open the entire entry (only if the user wants to see the entire text).

opnsenseuser:
@MichaelDeciso yes, in a few Hours!

opnsenseuser:
@MichaelDeciso i tested again and cleared the browser cache too. but this time the system log is completely empty! 

grafik

opnsenseuser:
@MichaelDeciso you did it! 👍 

grafik

PR #3027, opened 2018-12-08
Updated help text for verbosity level

Northguy:
Added help text for verbosity from https://nlnetlabs.nl/documentation/unbound/unbound.conf/

### DIFF
--- a/src/www/services_unbound_advanced.php
+++ b/src/www/services_unbound_advanced.php
@@ -345,7 +345,7 @@ include_once("head.inc");
                           endfor;?>
                           </select>
                           <div class="hidden" data-for="help_for_log_verbosity">
-                            <?= gettext("Select the log verbosity.") ?>
+                            <?= gettext("Select the log verbosity. Level 0 means no verbosity, only errors. Level 1 gives operational information. Level 2 gives detailed operational information. Level 3 gives query level information, output per query. Level 4 gives algorithm level information. Level 5 logs client identification for cache misses. Default is level 1. ") ?>
                           </div>
                         </td>
                       </tr>


src/www/services_unbound_advanced.php
```diff
@@ -345,7 +345,7 @@
                           endfor;?>
                           </select>
                           <div class="hidden" data-for="help_for_log_verbosity">
-                            <?= gettext("Select the log verbosity.") ?>
+                            <?= gettext("Select the log verbosity. Level 0 means no verbosity, only errors. Level 1 gives operational information. Level 2 gives detailed operational information. Level 3 gives query level information, output per query. Level 4 gives algorithm level information. Level 5 logs client identification for cache misses. Default is level 1. ") ?>
```
fichtner: default 1 seems to clash with this GUI setting once saved even if unintentional. first one is 0  so that's what it'll set?

fichtner: nevermind, checked code and it sets 1 for the GUI page and the backend config :)

fichtner:
Merged. Thanks, Patrick!

src/www/services_unbound_advanced.php
Northguy: Not sure what you mean, Franco, Do we have a different default level in verbosity in the GUI? If so, I need to adapt default to 0. These levels (and default) I obtained from Unbound.conf documentation. I've reset my lvl to 2, so cannot find default setting from gui anymore :-)

fichtner: this https://github.com/opnsense/core/blob/master/src/www/services_unbound_advanced.php#L59

the loop sets "0" first in the dropdown so default GUI selection would differ but it's all working as expected

PR #3029, opened 2018-12-09
Fix to add the duid length to th duid file

marjohn56:
When using different DUID types the lenght varies. Although not seen in the GUI, you must add the length as the first byte to the DUID file that dhcp6c reads, the second byte is null, then the duid itself.

### DIFF
--- a/src/etc/inc/interfaces.inc
+++ b/src/etc/inc/interfaces.inc
@@ -2858,13 +2858,19 @@ function interface_dhcpv6_configure($interface = 'wan', $wancfg)
 
     /* write DUID if override was set */
     if (!empty($config['system']['ipv6duid'])) {
-        $temp = str_replace(':', '', $config['system']['ipv6duid']);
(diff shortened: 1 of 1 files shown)


fichtner:
If we want to really fix it, we need to fix read and write, see 5568b35

marjohn56:
Ah, I assumed only the write bit had been broken, not the read bit as well.

fichtner:
No worries. I just felt there was a bit missing. I went to the dhcp6c code to figure out the format and saw that we earlier always prefixed 00:14: so it auto-fixed itself on write but that changed with the addition of the multiple length types. Then spent some hours with the PHP pack/unpack magic and verify the DUID file with a hex dump.

Nevertheless, if you can spare a moment please double check.

marjohn56:
Will do manyana…

fichtner:
¡Sí, claro, hasta mañana!

PR #2452, opened 2018-06-08
Fix hint for proxy.forward.acl.unrestricted.

kekek2:

PR #3030, opened 2018-12-09
Updated the help section for "Forwarded_for"

Northguy:
Updated the help section for "Forwarded_for", based on original Squid documentation at http://www.squid-cache.org/Doc/config/forwarded_for/

Other option could be to use lesser verbosity and use the squid documentation hyperlink in the help text, but this might make the help section sensitive for URL changes outside the OPNsense project itself.


Northguy:
Had the same thought regarding inclusion of hyperlinks in the help texts. I also don't like creating an additional layer of information on the OPNsense website, so let's abandon that path. Saw the markup in other 
help texts, so thought it was regular practice. Will cleanup when I have time available. Thanks for your comments.

fichtner:
It's true for the web proxy, it's full of markup, but it's also one of the earliest MVC components :)

For a tiny bit more info on this situation see https://github.com/opnsense/lang/issues/1

src/opnsense/mvc/app/controllers/OPNsense/Proxy/forms/main.xml
```diff
@@ -11,7 +11,8 @@
                 <id>proxy.general.icpPort</id>
                 <label>ICP port</label>
                 <type>text</type>
-                <help>The port number where Squid sends and receives ICP queries to and from neighbor caches. Leave blank to disable (default). The standard UDP port for ICP is 3130.</help>
+                <help>The port number where Squid sends and receives ICP queries to and from neighbor caches. 
+                    Leave blank to disable (default). The standard UDP port for ICP is 3130.</help>
```
fichtner: Merge with line above. XML will add lots of spaces here...

Northguy: Was not ready yet with editing. I am not really experienced with Git. When editing a file in online mode which is already part of a pull request, will automatically trigger a new pull request?

fichtner: No rush, if you need more time or help just let me know. Online edit on this branch should create new commits for this PR.

src/opnsense/mvc/app/controllers/OPNsense/Proxy/forms/main.xml
```diff
@@ -239,10 +229,9 @@
                 <help><![CDATA[
                   Enable SSL inspection mode,
                   which allows to log HTTPS connections information, such as requested URL and/or
-                  make the proxy act as a man in the middle between the internet and your clients.<br/>
+                  make the proxy act as a man in the middle between the internet and your clients.
                   Be aware of the security implications before enabling this option.
-                  <br/><br/>
-                  If you plan to use transparent HTTPS mode, you need nat rules to reflect your traffic.<br/>
+                  If you plan to use transparent HTTPS mode, you need nat rules to reflect your traffic.
                   <a href="/firewall_nat_edit.php?template=transparent_proxy&https=1">Add a new firewall rule </a>
```
Northguy: How about this hypertext? This has a function now (linking to the template for automatic adding of Port forward. This will loose it's functionality when rigidly applying the development direction for XML text strings?

Northguy: How about this hypertext? This has a function now (linking to the template for automatic adding of Port forward. This will loose it's functionality when rigidly applying the development direction for XML text strings?

fichtner: Need this and the other one as it offers a special NAT rule template on page open to avoid configuration errors. That's why we should have no markup, but require it anyway. In older components this would have been easier to keep out as we could control what goes into help texts and what is embedded as link via something such as sprintf(gettext('foo%s'), 'bar')

Northguy: OK, guess I am done. Left these specific URLs in place, but removed the wikipedia URL at the SNI section as this is non-functional.

Northguy:
Hi @fichtner , I guess I am finished with removing markup in the whole XML. Just wondering about the CDATA part. Should that stay? Comments in  #2452 suggest that CDATA is not necessary when HTML tags are not present in the text string?

fichtner:
Thanks! Yes they can be removed. Pseudo markup such as TIP: or NOTE: should be stripped as well. Either they are relevant in which case the sentences remain or they are irrelevant so all the text can be removed. How does that sound?

Northguy:
Sounds good. Will have a look at CDATA and Pseudo markup and notify once finished.

Regarding embedded HTML links in the help text, isn't it a better idea to remove them from the <help> tag? 

I see multiple options:
1. Leave them the way they are
2. You could put them in the <hint> tag, making them always visible, instead of dependent on the help toggle. More logical anyhow IMHO. Hint text will possibly exported to Translation strings, so is a shift of the 'issue'?
3. You could restructure the PHP page by creating a new text line under the help text and hint text, defined by a separate XML tag i.e. <Pagereference> This will make the <help> and <hint> text independent from the hyperlinks and might increase maintainability (although impacting overall design of the GUI pages).  This would also make it possible to either show it permanently or toggle it together with the help text.

Any thoughts? 

1) and 2) are easy. Need some guidance on PHP markup if you decide for approach 3)

fichtner:
3 is preferable, but requires to redesign MVC in this regard, not just that page (there is no PHP page for this, it is entirely generated from XML and Volt templates). The modelling of the new feature isn't so great, it's a "not really help text but may be useful for quick setup" type of thing.

NAT rules don't have firewall plugin hooks yet, they are one possible option 4 to sidestep the need for these helper links so we can make checkboxes for automatic rules. So going from 1. to 4. is the best approach without creating spurious work in the meantime.

Northguy:
I'll work on finishing 1. and update once done. Implementing 4 is out of my league, sorry...

fichtner:
Merged, thanks!

PR #3039, opened 2018-12-12
Fix #2990: Browser crashes when there are more than 10000 addresses in a pfTable alias

### LINKED ISSUE #2990: GEO_IP doesn´t work - crashes!
First of all, I've chosen alias as all countries that I need and then create the rule. but if I want to look at the pftable, the browser crashes and no content is displayed.

geo_ip_alias

geo_ip_crashes_browser

MichaelDeciso:


src/opnsense/mvc/app/controllers/OPNsense/Firewall/Api/AliasUtilController.php
```diff
@@ -132,13 +165,16 @@ public function flushAction($alias)
     /**
      * delete item from alias table
      * @param string $alias name
+     * @param $address
      * @return array status
+     * @throws \Exception
      */
-    public function deleteAction($alias)
+    public function deleteAction($alias, $address)
```
AdSchellevis: Just to be sure, did you test this with different types of items (ip[4|6] address, network)? 
This could cause issues with escaping, which is the reason why the address was initially send with the post itself (+ easy to extent the call to support multiple addresses at ones) 

We probably better keep this call like it was, people might already be using this endpoint in other scripts/applications too.

src/opnsense/mvc/app/controllers/OPNsense/Firewall/Api/AliasUtilController.php
```diff
@@ -90,14 +90,47 @@ public function aliasesAction()
...
     public function listAction($alias)
     {
         $this->sessionClose();
+
+        $itemsPerPage = intval($this->request->getPost('rowCount', 'int', 9999));
+        $currentPage = intval($this->request->getPost('current', 'int', 1));
+        $offset = ($currentPage - 1) * $itemsPerPage;
+
         $backend = new Backend();
         $entries = json_decode($backend->configdpRun("filter list table", array($alias, "json")));
```
AdSchellevis: Good to check with a large list, I kind of expect the configd call with > 1 Million results will still work like a charm, but if it doesn't we might need another approach here.  Did you test with a very large alias? and what was the reaction time in the browser for the api call.

src/opnsense/mvc/app/controllers/OPNsense/Firewall/Api/AliasUtilController.php
MichaelDeciso: I ran into the escaping issue, so I encoded the address with `urlencode`. I have tested it with regular IPv4 addresses and ranges like `192.168.1/24`. I haven't tested it with IPv6, though, but since `urlencode` handles `:`, it should not cause any surprises.

I would have preferred not touching this, but the `del` action on Bootgrid sends this value as a GET parameter, with an empty POST body. I have tried working around it, but it seems that I would have to bypass the `del` action entirely. I actually tried doing so, but I ran into a race issue with that.

The ValueA framework has a nice way of adding custom commands, but that does not appear to be present here. If it were to be backported, it would provide an easy way out.

What would be the best course of action, you think?

src/opnsense/mvc/app/controllers/OPNsense/Firewall/Api/AliasUtilController.php
MichaelDeciso: I tried it with the same alias as René, i.e. one with all Asian and African countries. That amounts to 61461 entries. The reaction time is indistinguishable from that of the bogons alias.

src/opnsense/mvc/app/controllers/OPNsense/Firewall/Api/AliasUtilController.php
AdSchellevis: I see, hooking custom functions should be rather simple, if I'm not mistaken this  does something similar:

https://github.com/opnsense/core/blob/e0692a85b32239b5d6edbab742b7ed073034d7b6/src/opnsense/mvc/app/views/OPNsense/IDS/index.volt#L247-L249

Better to revert to the original call then.

opnsenseuser:
see also -> https://github.com/opnsense/core/issues/2990

src/opnsense/mvc/app/controllers/OPNsense/Firewall/Api/AliasUtilController.php
MichaelDeciso: It doesn't look like that function does what I need - it only adds a custom button (which I had no trouble with), but it still uses the default edit action (it has a `command-edit` class, which means that Bootgrid automatically adds the default edit action to it). My problem was that I need to define a custom action, because the default delete action does not work with this call.

MichaelDeciso: I found a solution.

PR #3054, opened 2018-12-18
Change foreach loops that don't use values to use array_keys()

MichaelDeciso:
As requested by @fichtner .


fichtner:
merged, thanks!

PR #3072, opened 2018-12-26
dhcpd6 leases not always correctly displayed

marjohn56:
Taken Ad's version as it's much cleaner and works on both my live system with multiple leases and my test system that has only one lease.

### DIFF
--- a/src/www/status_dhcpv6_leases.php
+++ b/src/www/status_dhcpv6_leases.php
@@ -97,15 +97,24 @@ $interfaces = legacy_config_get_interfaces(array('virtual' => false));
 $leasesfile = services_dhcpdv6_leasesfile();
 
 if ($_SERVER['REQUEST_METHOD'] === 'GET') {
-    $awk = "/usr/bin/awk";
+    $leases_content = array();
 
-    /* this pattern sticks comments into a single array item */
-    $cleanpattern = "'{ gsub(\"^#.*\", \"\");} { gsub(\"^server-duid.*\", \"\");} { gsub(\";$\", \"\"); print;}'";
-    /* We then split the leases file by } */
-    $splitpattern = "'BEGIN { RS=\"}\";} {for (i=1; i<=NF; i++) printf \"%s \", \$i; printf \"}\\n\";}'";
-
-    /* stuff the leases file in a proper format into a array by line */
-    exec("/bin/cat {$leasesfile} | {$awk} {$cleanpattern} | {$awk} {$splitpattern} | /usr/bin/grep '^ia-.. '", $leases_content);
+    $fin = @fopen($leasesfile, "r");
+    $section = array();
+    if ($fin) {
+        while (($line = fgets($fin, 4096)) !== false) {
+            if (strpos($line, "ia-") !== false) {
+                $section = array();
+            } elseif (strpos($line, "}") === 0 && count($section) > 0) {
+                $output = implode($section, "");
+                $leases_content[] = $output;
+            }
+            $section[] = rtrim($line, "\n;");
+        }
+        fclose($fin);
+    }
+    
+        
     $leases_count = count($leases_content);
     exec("/usr/sbin/ndp -an", $rawdata);
     $ndpdata = array();


src/www/status_dhcpv6_leases.php
```diff
@@ -97,15 +97,49 @@ function parse_duid($duid_string) {
...
-    /* this pattern sticks comments into a single array item */
-    $cleanpattern = "'{ gsub(\"^#.*\", \"\");} { gsub(\"^server-duid.*\", \"\");} { gsub(\";$\", \"\"); print;}'";
-    /* We then split the leases file by } */
-    $splitpattern = "'BEGIN { RS=\"}\";} {for (i=1; i<=NF; i++) printf \"%s \", \$i; printf \"}\\n\";}'";
-
-    /* stuff the leases file in a proper format into a array by line */
-    exec("/bin/cat {$leasesfile} | {$awk} {$cleanpattern} | {$awk} {$splitpattern} | /usr/bin/grep '^ia-.. '", $leases_content);
+    $leases_content = array();
+
+    $t_dhcpd6_leases = file_get_contents($leasesfile);
```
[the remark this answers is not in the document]
fabianfrz: @AdSchellevis @marjohn56  It does not hurt early - there is a user who is reading a 40MiB log file which means that you have to have this in memory first. I had to do this change as well because this is almost the limit configured in PHP.

You can look at opnsense/plugins#1075 to see how I implemented this some days ago.

src/www/status_dhcpv6_leases.php
```diff
@@ -97,15 +97,49 @@ function parse_duid($duid_string) {
...
+    while(1) {
+        $t_start_pointer = strpos($t_dhcpd6_leases,"ia-pd",$tpos);
+        if($t_start_pointer == false) {
+            break;
+        }
+        $t_end_pointer = strpos($t_dhcpd6_leases,"}",$t_start_pointer);
+        $t_end_pointer = strpos($t_dhcpd6_leases,"}",$t_end_pointer+1);
+        $t_entry = substr($t_dhcpd6_leases,$t_start_pointer, ($t_end_pointer+1) - $t_start_pointer);
+        $t_entry = str_replace(";"," ", $t_entry);
+        $t_entry = str_replace("\n"," ", $t_entry);
```
fabianfrz: I am not sure but you probably want to use this function:

https://secure.php.net/manual/de/function.rtrim.php

src/www/status_dhcpv6_leases.php
```diff
@@ -97,15 +97,49 @@ function parse_duid($duid_string) {
...
+        // find end of lease entry, twice as all entries contain double brackets
+        $t_end_pointer = strpos($t_dhcpd6_leases,"}",$t_start_pointer);
+        $t_end_pointer = strpos($t_dhcpd6_leases,"}",$t_end_pointer+1);
+        // we have an entry. Now extract it.    
+        $t_entry = substr($t_dhcpd6_leases,$t_start_pointer, ($t_end_pointer+1) - $t_start_pointer);
+        // strip CR and semis
+        $t_entry = str_replace(";"," ", $t_entry);
+        $t_entry = str_replace("\n"," ", $t_entry);
+        // Put entry in array
+        array_push($leases_content,$t_entry);
```
fabianfrz:
```suggestion
        $leases_content[] = $t_entry;
```

src/www/status_dhcpv6_leases.php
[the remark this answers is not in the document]
marjohn56: I started out looking at using fread, but a single lease entry is across multiple lines, that means that an may cross the buffer boundary. Guess I could do it by copying the partial entry to a variable first, then reading more into the buffer etc etc, or I could just use read fgets and assemble thereafter.

src/www/status_dhcpv6_leases.php
marjohn56: If I use the fgets method yes.

src/www/status_dhcpv6_leases.php
fabianfrz: it is shorter and more commonly used.

AdSchellevis:
@marjohn56 this should basically do the trick for both types, but to be honest, more of the existing code may have some cleanup too... 

```
$leases_content = array();

$fin = @fopen($leasesfile, "r");
$section = array();
if ($fin) {
    while (($line = fgets($fin, 4096)) !== false) {
        if (strpos($line, "ia-") !== false) {
            $section = array();
        } elseif (strpos($line, "}") === 0 && count($section) > 0) {
            $output = implode($section, "");
            $leases_content[] = $output;
        }
        $section[] = rtrim($line, "\n;");
    }
    fclose($fin);
}

```

marjohn56:
And that lovely simple piece of code works nicely!
I've obviously over indulged in the eggnog over Christmas.

AdSchellevis:
@marjohn56 thanks!

PR #3073, opened 2018-12-26
DUID-LL Generator fix

marjohn56:
Timestamp is not used for DUID-LL , only the mac address. There is some confusion in the RFC3315 doc. Section 9.4 states that DUID-LL should be the link layer address in canonical form, as described in RFC2464. That doc shows the link-layer part of the DUID would be  xx:xx:xx:ff:fe:xx:xx:xx, however, looking at 'in use' examples of DUID-LL the additional ff:fe centre pair are not used, so I've not added them for Opnsense.

### DIFF
--- a/src/www/system_advanced_network.php
+++ b/src/www/system_advanced_network.php
@@ -64,12 +64,8 @@ function generate_new_duid($duid_type)
             }
             $new_duid = $new_duid.':'.$timestamp.':'.$mac;
             break;
-        case '2': //LL
-            $ts = time() - 946684800;
-            $hts = dechex($ts);
-            $timestamp = sprintf("%s",$hts);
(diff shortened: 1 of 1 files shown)


Franck78:
```
       case '2': //LL - NO TIMESTAMP: Just 03:00:01:00: + Link layer address in canonical form, so says RFC.
            $mac = get_mac_address(getenv('REMOTE_ADDR'));
            $type = "\x00\x03\x00\x01";
```
Just 03:00:01:00: 
or
Just 00:03:00:01 ???

marjohn56:
Typo in my comment.. sorry, I'll correct that, It is the latter.

fichtner:
Nice catch, thanks!

PR #3074, opened 2018-12-27
Stable/18.7> dhcpdv4: adding opt67 (bootfilename) and staticmap netboot

terminar:
This is my first attempt on contributing to OPNsense and maybe not the correct path but anyway...

I switched from an old (manually configured) router to OPNsense which works great for most of the situations i have to solve. Two problems remained.

1. DHCP option 67 can't be set for a sub pool
----------------------------------------------------
This problem occurs with VoIP networks and different phone devices which should be autoprovisioned via option 66 (tftp server) and option 67 (bootfile name). Phone vendors "misuse" these two options.
Also, often the options (configuration path in 67) are completely different and can't be set network-wide. Using an asterisk softpbx like "Gemeinschaft" or "Gemeinschaft" based softpbx systems are an example for this situation.
There are several devices which are found via vendor MAC-Prefix (like snom with 00:04:13) and then auto-provisioned/redirected to 66> http://ip.of.pbx with 67> gemeinschaft/prov/snom/settings.php={mac}.

The first commit solves this problem, adding Option67 to the TFTP Setting.

2. Netboot settings can't be set for a static configured host
------------------------------------------------------------------
It seems to be impossible to maintain a static list of configured host with different netboot setting per-host. The second commit solves this problem, adding the "netboot" block to the staticmap.
While i was adding it, i found old settings "filename" and "rootpath" in the services.inc - it seems that it was possible in the past to configure netboot in the staticmap? Maybe, this is just re-adding the options.

-----
One new problem will maybe still remain: this MR is for stable/18.7 because i am using it.
I have no clue if MR's are also accepted to stable branches and what i have to do for the master branch (or how i can test it).

### DIFF
--- a/src/etc/inc/services.inc
+++ b/src/etc/inc/services.inc
@@ -934,12 +934,27 @@ EOD;
                     $dhhostname = str_replace(".", "_", $dhhostname);
                     $dhcpdconf .= "  option host-name \"{$dhhostname}\";\n";
                 }
-                if (!empty($sm['filename'])) {
-                    $dhcpdconf .= "  filename \"{$sm['filename']}\";\n";
-                }
 
-                if (!empty($sm['rootpath'])) {
-                    $dhcpdconf .= "  option root-path \"{$sm['rootpath']}\";\n";
+                if (isset($sm['netboot'])) {
+                    if (!empty($sm['nextserver'])) {
+                        $dhcpdconf .= "  next-server {$sm['nextserver']};\n";
+                    }
+                    if (!empty($sm['filename']) && !empty($sm['filename32']) && !empty($sm['filename64'])) {
+                        $dhcpdconf .= "  if option arch = 00:06 {\n";
+                        $dhcpdconf .= "    filename \"{$sm['filename32']}\";\n";
+                        $dhcpdconf .= "  } else if option arch = 00:07 {\n";
+                        $dhcpdconf .= "    filename \"{$sm['filename64']}\";\n";
+                        $dhcpdconf .= "  } else if option arch = 00:09 {\n";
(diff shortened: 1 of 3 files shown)


fichtner:
Hi @terminar,

Thanks for this! Looks reasonable on first glance but please give me a few days for review and double-checking. :)


Cheers,
Franco

fichtner:
PS: we do not merge into stable/18.7. It should be in master, but I can do this manually at no trouble.

fichtner:
Yes, normally you switch to development from there. You will get latest master branch aligned with when the 18.7.x releases are done. If you need newer changes there's opnsense-patch to cherry-pick commits from GitHub or you use:

    # opnsense-code core
    # cd /usr/core
    # make upgrade

To get an up-to-date git with the master branch... :)

terminar:
@fichtner Any update on this pull request? I just want to know if I should try to create a new PR to master and close this one.

fichtner:
@terminar I was actually asking for you to kindly rebase the patch on the development code, but let me merge this manually after review this time then...

fichtner:
I have merged 8e361f3 as a modified version of your bootfile-name patch.

We will need to redo the UEFI files changes for the static mappings, because the implementation did not take into account this existing code:

https://github.com/opnsense/core/blob/master/src/www/services_dhcp_edit.php#L360-L381

This latches on to the netboot setting of the dhcp server so changing it in the proposed way will change the behaviour and I'm not sure this was intended / anticipated. We have to assume some people are using this as is.

Please test 8e361f3 on top of 19.1.2:

    # opnsense-patch 8e361f3

For the next PR please use the master branch and only post a single change per PR, it's easier to work on it and a lot quicker to be merged.


Cheers,
Franco

PR #3077, opened 2018-12-29
Add missing entries to Dpinger advanced section

marjohn56:
It's only taken 8 months to do this.. impressive. :)

### DIFF
--- a/src/etc/inc/gwlb.inc
+++ b/src/etc/inc/gwlb.inc
@@ -33,15 +33,15 @@
 function return_dpinger_defaults()
 {
     return array(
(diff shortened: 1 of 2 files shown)


marjohn56:
Typo correction

fichtner:
Applied modified version. Thanks! data_payload is still unused. Do we need it?

marjohn56:
Could not work out what that's supposed to be used for... I guess not until we do.

fichtner:
Let‘s remove it then, -d 0 option seems to be the default anyway.

PR #3080, opened 2018-12-30
handle <if>|<ip> interface notation

devopsoffice-ivan:
This patch was necessary to initialize GRE tunnel with IP alias VIP.

### DIFF
--- a/src/etc/inc/interfaces.inc
+++ b/src/etc/inc/interfaces.inc
@@ -3996,6 +3996,13 @@ function get_interface_ip($interface = 'wan')
         return get_configured_carp_interface_list($interface);
     }
 
+    // handle <if>|<ip> notation. it's used for ip alias vip ips,
+    // e.g. gre parent interface with ip alias
+    $pipe_pos = strpos($interface, '|');
+    if ($pipe_pos !== FALSE) {
+	return substr($interface, $pipe_pos + 1);
+    }
+
     return null;
 }
 


AdSchellevis:
We probably should hack this into `get_interface_ip()`, but find the real culprit here. Could you describe the issue in more detail? Maybe we can figure out the correct patch here.

devopsoffice-ivan:
I wanted to setup a GRE over IPsec tunnel, so first I configured IPsec with IP from a /31 to each side.
Then I wanted to configure GRE but I couldn't select IPsec interface so I configured IP alias for interface which was used for IPsec tunnel, IP address for IP alias is the same as IPsec local net.
I could select this IP alias in GRE configuration page from Parent interface selectbox.

I could save this configuration but I realized that tunnel net is not configured for GRE (only inet addresses) so it wasn't usable.
After some testing I found root cause and created this patch.
I didn't find better solution yet.

Working GRE interface:
gre3: flags=a051<UP,POINTOPOINT,RUNNING,LINK1,MULTICAST> metric 0 mtu 1476
        options=80000<LINKSTATE>
        tunnel inet 172.20.0.15 --> 172.20.0.14
        inet 172.20.0.17 --> 172.20.0.16  netmask 0xfffffffe 
        inet6 fe80::ae1f:6bff:fe76:32ee%gre3 prefixlen 64 scopeid 0x10 
        nd6 options=21<PERFORMNUD,AUTO_LINKLOCAL>
        groups: gre 

Not working GRE interface:
gre3: flags=a051<UP,POINTOPOINT,RUNNING,LINK1,MULTICAST> metric 0 mtu 1476
        options=80000<LINKSTATE>
        inet 172.20.0.17 --> 172.20.0.16  netmask 0xfffffffe 
        inet6 fe80::ae1f:6bff:fe76:32ee%gre3 prefixlen 64 scopeid 0x10 
        nd6 options=21<PERFORMNUD,AUTO_LINKLOCAL>
        groups: gre

devopsoffice-ivan:
I looked at related code and my plan to refactoring code is the following:
- refactor get_configured_ip_aliases_list() following the same logic used for get_configured_carp_interface_list(). it will generate unique alias id as key for returned hash. modified get_configured_ip_aliases_list() will be used in get_interface_ip().
- modify web interface related php files to use generated alias id. currently each php files generate ip alias id on it's own but with same logic. (<if>|<ip>)

What's your opinion?

AdSchellevis:
at a first glance it sounds a bit too complicated for the issue at hand, maybe we could pinpoint the issue better, let me see.

devopsoffice-ivan:
OK. Let me know your suggestions and I will implement required modification happily.

devopsoffice-ivan:
As I see affected files would be in core:
src/etc/inc/util.inc
src/www/services_ntpd.php
src/www/vpn_openvpn_server.php
src/www/vpn_openvpn_server.php
src/www/interfaces_gif_edit.php
src/www/interfaces_gre_edit.php
src/www/vpn_ipsec_phase1.php
src/www/vpn_openvpn_client.php
src/www/vpn_ipsec.php

A possible simper way to handle ip alias in get_interface_ip() is a new function which validates data and returns IP.

Another way to handle this issue without touching get_interface_ip() is to rethink GRE configuration page because GRE doesn't depend on interface it use IP only for configuration. A list of local IP addresses is enough instead of interface list. (I don't know it's easier or not because I don't know related code.)

AdSchellevis:
ok, how about https://github.com/opnsense/core/commit/8c0683540d592c873dd4e46efd296ce43d2cd7a0 ?

Only impacts gre interfaces, there are better options, but in most cases there's some migration involved.

fichtner:
I think the original issue is:
 
https://github.com/opnsense/core/blob/8c0683540d592c873dd4e46efd296ce43d2cd7a0/src/www/vpn_openvpn_server.php#L758

And CARP is similar:

https://github.com/opnsense/core/blob/8c0683540d592c873dd4e46efd296ce43d2cd7a0/src/www/vpn_openvpn_server.php#L754

devopsoffice-ivan:
as i see interface|address format originates from web interface related php files (under www folder), 
these all use the return value of get_configured_ip_aliases_list() and generated this form of identifier on it's own.

i checked get_configured_carp_interface_list() which can generate a a tiny identifier using configd result (vhid key), but ipaliases has no such numeric identifier to generate such tiny id.
I think to handle the root cause first ip aliases would need a similar id than carp addresses.

    [3] => Array
        (
            [type] => single
            [subnet_bits] => 24
            [mode] => carp
            [interface] => lan
            [descr] => MGMT VIP
            [subnet] => 192.168.x.x
            [vhid] => 5
            [advskew] => 0
            [advbase] => 1
            [password] => xzy
        )

    [4] => Array
        (
            [type] => single
            [subnet_bits] => 24
            [mode] => ipalias
            [interface] => lan
            [descr] => Other MGMT IP
            [subnet] => 10.x.x.x
        )

devopsoffice-ivan:
Maybe we may say that interface|ip notation is a valid identifier (yes it can be misleading, but it fits for ipv4 and ipv6).

wan, lan, optX is handled by find_interface_ip in get_interface_ip()
optX_vipX is handled by get_configured_carp_interface_list()
interface|ip is handled by a modified get_configured_ip_aliases_list()

AdSchellevis:
In the openvpn server conf, it's saved as two different entities (interface, ipaddr):

https://github.com/opnsense/core/blob/8c0683540d592c873dd4e46efd296ce43d2cd7a0/src/www/vpn_openvpn_server.php#L85-L87

There's only one tiny block of code that doesn't do anything when providing an address:

https://github.com/opnsense/core/blob/8c0683540d592c873dd4e46efd296ce43d2cd7a0/src/etc/inc/interfaces.inc#L696-L698

Although I expect this is quite unnecessary here (should also be easy to test).

If we would go for the same approach as the openvpn implementation, we could consider saving both values (`if` + `if_addr`) in stead of my quick fix of flushing the address to the `if` field

AdSchellevis:
if you can validate that https://github.com/opnsense/core/commit/8c0683540d592c873dd4e46efd296ce43d2cd7a0 does the trick, we could probably pull it in for the next version. after 19.1 we could investigate a better solution, which probably involves adding id's to the virtual ip's.

AdSchellevis:
Happy New Year to you too :)

devopsoffice-ivan:
I checked 8c06835 and it seems to be good. Tunnel was recreated and other side responds to ping.
Many thanks again!

fichtner:
Looks like this was resolved....

PR #3088, opened 2019-01-02
mail/rspamd: add a field dependent range validator

fabianfrz:
required to check when there is a max limit of a value, that is additionally limited by another field. For example 

0 < a < b < c < 100

see opnsense/plugins#1114 for more details


AdSchellevis:
Hi @fabianfrz, looks like a nice addition to have, if you  have a bit of time available, could you also add a unit test for this constraint type ? (https://github.com/opnsense/core/tree/master/src/opnsense/mvc/tests)

We don't have everything covered at the moment, but I would like to keep up with new additions. 
If you need help, just let me know, it shouldn't be very difficult to add some relevant test cases for this type.

Thanks,

Ad

AdSchellevis:
@fabianfrz perfectly fine, it differs a bit from the other tests, this is more isolated, which has a slight disadvantage in the amount of plumbing needed, but has less risk of accidental interference by other tests. Thanks!

https://github.com/opnsense/core/blob/23431cb7ea94f36595c38f3ad4aeffc57a392f8d/src/opnsense/mvc/tests/app/models/OPNsense/Base/BaseModel/TestModel.xml#L65-L92 

https://github.com/opnsense/core/blob/23431cb7ea94f36595c38f3ad4aeffc57a392f8d/src/opnsense/mvc/tests/app/models/OPNsense/Base/BaseModelTest.php#L313-L353

PR #3091, opened 2019-01-03
 Also sync Virtual IPs with a VHID group to backup host, closes #3034

### LINKED ISSUE #3034: xmlrpc-sync, alias+vhid
Synchronise VIP's using vhid's, partially discussed here https://github.com/opnsense/core/issues/2795

MichaelDeciso:


MichaelDeciso:
I ran the following tests (after the third commit):
```
Create:
- CARP: works
- VIP without VHID: works
- VIP with VHID: works:

Modify:
- CARP: works
- VIP without VHID: works
- VIP with VHID: works

Delete:
- VIP with VHID: works
```

MichaelDeciso:
@AdSchellevis I ran the test that you requested (making changes to IPalias+VHID) and it indeed fails to properly sync (it creates duplicates, as you expected).

A solution could be to remove all pre-existing CARPs and IPalias+VHID, then create new ones based on the XMLRPC data. But I'm not sure if that is an acceptable solution, performance-wise.

PR #3093, opened 2019-01-03
Clarify incorrect gateway IP error

MichaelDeciso:
I ran into this error while attempting to add a VIP. The old message implies that the error is in the regular IP address field, rather than in the gateway IP address field.

### DIFF
--- a/src/www/firewall_virtual_ip_edit.php
+++ b/src/www/firewall_virtual_ip_edit.php
@@ -110,7 +110,7 @@ if ($_SERVER['REQUEST_METHOD'] === 'GET') {
             }
         }
         if (!empty($pconfig['gateway']) && !is_ipaddr($pconfig['gateway'])) {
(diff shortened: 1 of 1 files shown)


src/www/firewall_virtual_ip_edit.php
```diff
@@ -110,7 +110,7 @@ function find_last_used_vhid() {
             }
         }
         if (!empty($pconfig['gateway']) && !is_ipaddr($pconfig['gateway'])) {
-            $input_errors[] = gettext("A valid IP address must be specified.");
+            $input_errors[] = gettext("A valid Gateway IP address must be specified.");
```
fichtner:
```suggestion
            $input_errors[] = gettext("A valid gateway IP address must be specified.");
```

fichtner:
Thanks! ❤️

PR #3094, opened 2019-01-03
Cleanup/misc

MichaelDeciso:
Redundant closing tags are forbidden by PSR-2, as they can cause subtle bugs.


fichtner:
Merged, thanks!

PR #3104, opened 2019-01-08
Update status_dhcpv6_leases.php

marjohn56:
Fix to correct UTC and local time correctly. Invert the local/gmt selection and subtract the offset rather than add corrects it.

V4  is also incorrect, not looked at that one.

### DIFF
--- a/src/www/status_dhcpv6_leases.php
+++ b/src/www/status_dhcpv6_leases.php
@@ -50,11 +50,11 @@ function adjust_gmt($dt)
 
     $timezone = $config['system']['timezone'];
     $ts = strtotime($dt . " GMT");
-    if ($dhcpv6leaseinlocaltime == "yes") {
+    if ($dhcpv6leaseinlocaltime != "yes") {
         $this_tz = new DateTimeZone($timezone);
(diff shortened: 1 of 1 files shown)


fichtner:
Merged, thanks a lot for this!

src/www/status_dhcpv6_leases.php
```diff
@@ -50,11 +50,11 @@ function adjust_gmt($dt)
 
     $timezone = $config['system']['timezone'];
     $ts = strtotime($dt . " GMT");
-    if ($dhcpv6leaseinlocaltime == "yes") {
+    if ($dhcpv6leaseinlocaltime != "yes") {
```
fichtner: I don't know, this looked more correct before: assume time is GMT and if not add the offset (it can be negative) unless we assume the time is never GMT from dhcp in the first place, but then the help text is wrong...

marjohn56: leasetime in dhcp will always be local time surely?

fichtner: dunno. some part of the code is definitely still lying to us oO

PR #3107, opened 2019-01-08
Add support IDN in aliases.

kekek2:


kekek2:
I tested it.

AdSchellevis:
ok, thanks

PR #3109, opened 2019-01-08
Fix usage of include_once in load_phalcon.php

MichaelDeciso:
include_once returns TRUE if the file has already been loaded. This is not what you want if you're setting a variable from it.
http://php.net/manual/en/function.include-once.php

This fixes notices when including certain old scripts from modern controllers, as well as a notice when running run_migrations.php.


AdSchellevis:
it's only included once, so `include_once` and `include` act equally in this case

PR #3110, opened 2019-01-08
Port DHCPv4 leases page to MVC; fix #3095

### LINKED ISSUE #3095: DHCP Leases Columns Sortable?
On the DHCP leases page, the column headings are clickable (cursor does not change but can still click on them.)  The page refreshes like it tries to sort the columns but nothing changes.  Not sure if the columns are suppose to be sort-able or not.

MichaelDeciso:


AdSchellevis:
doesn't add enough value at the moment in the state it's in. let's skip for now.

luismompohanden:
Would be really helpful aka valuable. We have to maintain hundreds of IP reservations manually.

soif:
It would be really valuable to get an DHCP API, to list the leases, as well as to add/delete static leases...

Would'nt you reconsider reviewing/merging this PR? 🙏 

Or please advice if we should first improve something?

Thank you

PR #3117, opened 2019-01-09
Generate correct config for "Hybrid-RSA + XAuth" IPSec

max-weller:
This PR allows to use the Hybrid-RSA + XAuth mode of IKEv1. Without these changes, strongswan fails with following error message:

    looking for HybridInitRSA peer configs matching x.x.x.x...x.x.x.x[x.x.x.x]
    found 1 matching config, but none allows HybridInitRSA authentication using Main Mode

-----

First commit fixes a logical problem in the config form for IPSec Phase 1.

Second commit fixes a problem in the generated `ipsec.conf`: To use Hybrid-RSA with XAuth the authentication must be configured exactly as leftauth=pubkey, rightauth=xauth.
No rightauth2/leftauth2 may be set.
This can be seen here in the source code:
https://github.com/strongswan/strongswan/blob/be1c7e38157b336655894d99012d25f1b3b75429/src/libcharon/sa/ikev1/phase1.c#L506

### DIFF
--- a/src/www/vpn_ipsec_phase1.php
+++ b/src/www/vpn_ipsec_phase1.php
@@ -201,6 +201,9 @@ if ($_SERVER['REQUEST_METHOD'] === 'GET') {
             $reqdfieldsn = array(gettext("Pre-Shared Key"));
             break;
         case "hybrid_rsa_server":
+            $reqdfields = explode(" ", "certref");
+            $reqdfieldsn = array(gettext("Certificate"));
+            break;
         case "xauth_rsa_server":
         case "rsasig":
             $reqdfields = explode(" ", "caref certref");
@@ -491,6 +494,9 @@ include("head.inc");
                     }
                     break;
                 case 'hybrid_rsa_server':
+                    $(".auth_eap_tls").show();
+                    $(".auth_eap_tls :input").prop( "disabled", false );
+                    break;
                 case 'xauth_rsa_server':
                 case 'rsasig':
                 case 'rsa_eap-mschapv2':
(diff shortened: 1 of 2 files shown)


src/www/vpn_ipsec_phase1.php
```diff
@@ -491,6 +494,9 @@ function ipsec_ikeid_next() {
                     }
                     break;
                 case 'hybrid_rsa_server':
+                    $(".auth_eap_tls").show();
```
MichaelDeciso:
```suggestion
                    $('.auth_eap_tls').show();
```

src/www/vpn_ipsec_phase1.php
```diff
@@ -491,6 +494,9 @@ function ipsec_ikeid_next() {
...
+                    $(".auth_eap_tls :input").prop( "disabled", false );
```
MichaelDeciso:
```suggestion
                    $('.auth_eap_tls :input').prop('disabled', false);
```

src/www/vpn_ipsec_phase1.php
max-weller: Hi @MichaelDeciso 
 I just copied from the code above, which also uses double quotes everywhere, to make it more consistent. Why should the quotes be changed?
Thanks

MichaelDeciso: It's what we do for all new code. The old code mostly uses double quotes, you're correct about that.

max-weller: okay

AdSchellevis:
@max-weller the ipsec.conf seems to be faulty indeed, thanks for the fix!

PR #3118, opened 2019-01-11
Port license page to MVC

MichaelDeciso:


src/opnsense/mvc/app/models/OPNsense/Base/Menu/Menu.xml
```diff
@@ -5,7 +5,6 @@
             <Root url="/" visibility="hidden"/>
             <RootArgs url="/?*" visibility="hidden"/>
         </Dashboard>
-        <License order="1" url="/license.php" cssClass="fa fa-balance-scale fa-fw"/>
```
fichtner: I'd leave this here modified for the new URL unless we want to create a separate ACL page too.. the general reasoning is "should it deinstal / install as a plugin?" if yes, all separate files makes sense, otherwise it can probably stay. But I'm not opposed to the removal in general.

src/opnsense/mvc/app/views/OPNsense/Core/license.volt
```diff
@@ -0,0 +1,21 @@
+<section class="col-xs-11">
+    <p><a href="https://opnsense.org/" target="_blank">OPNsense&reg;</a> is Copyright &copy; {{ product_copyright_years }} {{ product_copyright_owner }}<br>All rights reserved.</p>
```
fichtner: opnsense.org and "OPNsense" should be available as product_... view variables as well.

MichaelDeciso: I have changed it. After discussing it with Jos, I have also taken precautions to ensure that if someone changes the product name, it will still show our copyright correctly.

Makefile
```diff
@@ -71,6 +71,7 @@ CORE_MAINTAINER?=	project@opnsense.org
 CORE_ORIGIN?=		opnsense/${CORE_NAME}
 CORE_PACKAGESITE?=	https://pkg.opnsense.org
 CORE_PRODUCT?=		OPNsense
+CORE_PRODUCT_LEGAL?=OPNsense®
```
fichtner: what have I done ;( shuffling UTF-8 through make and json is not so great

src/opnsense/mvc/app/views/OPNsense/Core/license.volt
fichtner: makes sense, thanks 👍

src/opnsense/mvc/app/models/OPNsense/Base/Menu/Menu.xml
MichaelDeciso: I have put the License under 'Core'. For some reason, the general ACL file is already in that location, while the general Menu file is under 'Base'.

Since the Base/Menu/Menu.xml file only appears to list legacy pages, I think putting it into its own Menu.xml is the best way to go.

fichtner: why not move base menu file to core?

Makefile
MichaelDeciso: It seems to handle it correctly, though.

src/opnsense/mvc/app/models/OPNsense/Base/Menu/Menu.xml
MichaelDeciso: It seemed too important for me to touch, but if you'd like me to, I'll do it.

fichtner: yes please, "Base" prefix predates some later works that have a more aligned structure like ACL.

MichaelDeciso: Done.

Makefile
fichtner: If people go through the trouble of rebranding they can zap the (r) in the volt or you make it conditional as well. This is too little benefit to have it included and opnsense-version doesn't know about it.

MichaelDeciso: I discussed this approach with Jos and he preferred it this way.

I think we can use UTF-8 in Makefiles anno 2018. As I mentioned, I did test it and it works correctly.

fichtner: ok fine, I disagree with having this piggybacked through a build system

MichaelDeciso: It's no worse than having our website URL piggybacked through it, IMO. As for zapping the ® manually: I find that very inconsistent with your initial request of using template variables. Either we should use template variables throughout or require manual editing, not a mix.

I will change the approach, but I heavily disagree: I find your stance _very_ inconsistent.

fichtner: You don't have to change it for me. Ad can merge it when he is back. I just don't see a point in passing both "OPNsense" and "OPNsense(r)" for a single use in a file that has no functional use. Moreover, the user neither sees nor cares about the change so why do it in the first place? If you feel like I'm being inconsistent it is because you're trying to argue if it's already inconsistent it's only consistent to keep building on the inconsistencies. :)

MichaelDeciso: No, I felt you were being inconsistent because we already have a lot of stuff that isn't _strictly_ make-related (such as URLs) in our makefile. To me, adding something to that should be fine.

I changed it because I asked Jos again and he said that, since you felt strongly about it and you handle most of the build stuff, it would probably be better to follow your instructions. So I did.

There are a lot of unreviewed, unmerged PRs already -- I don't really want to stall this one just because I disagree with you over what, in the end, isn't that major. I'm taking your approach, and I will complain no further about it.

fichtner: Look, take a bit of time to look at the history of the code you're changing if in doubt. The previous location for the product data was worse. And I'd replace this current one in a heartbeat if there was a better idea. So all I am asking is not starting to clutter stuff that is not great in an of itself. I'm fully aware...

Thing is I would have not started reviewing if that change was in the original PR or said no directly. The purpose of the PR is not to sneak in more later because that derails the review process unless it's agreed that such a change is necessary. The other change was great and it was correct to take that risk, but one good and one bad change still stalls the PR.

fichtner:
Merged, thanks!

Makefile
fichtner: You can thumbs down this all you like, because my point still stands. :)

MichaelDeciso: You requested using more template variables, I discussed what approach I should take with Jos and then applied it. That is _not_ sneaking stuff. You can accuse me of plenty of things, but not of "sneaking stuff it into PRs", and I don't appreciate this one bit.

fichtner: I'm sorry for phrasing it badly. In the scope of this PR, previously undiscussed changes were added that inadvertently altered the review state in my opinion. I can only judge this PR. I'm not judging you.

MichaelDeciso: Ok, thanks!

Issue #3063, opened 2018-12-21
Monit: adding ping check causes syntax failure


Taomyn:
opnSense v18.7.9-amd64
Monit v5.25.2

When I enable "ping" for one of my service settings I receive the error:

`monit: /usr/local/etc/monitrc:17: syntax error 'failed'`

When I compare the monitrc file the difference is:

```
check system bart.domain.com
   if failed ping then alert 
   if memory usage is greater than 75% then alert 
   if cpu usage is greater than 75% then alert 
   if loadavg (1min) is greater than 8 then alert 
   if loadavg (5min) is greater than 6 then alert 
   if loadavg (15min) is greater than 4 then alert 
```

and

```
check system bart.domain.com
   if memory usage is greater than 75% then alert 
   if cpu usage is greater than 75% then alert 
   if loadavg (1min) is greater than 8 then alert 
   if loadavg (5min) is greater than 6 then alert 
   if loadavg (15min) is greater than 4 then alert 
```
Line 17 points at "if failed ping then alert"

fichtner:
done via #3119

Issue #2918, opened 2018-11-15
Monit: service tests and relation to service type


AdSchellevis:
The monit service tests configuration contains different test options, but can be selected on every service now, in reality this would lead to configuration errors.

For example, when adding a file service, it's possible to select a ping test now.

I would like to suggest to add a service type field to the "Service Tests Settings" section, which can then be used as validation or selection filter in the service definition.

AdSchellevis:
@fbrendel what do you think?

fbrendel:
Basically sounds good.
However, the challenge is to find out which test is valid for which service.
E.g. existence tests are valid for files, directories, processes etc.
So we need a mapping between service types and test types (Fault tolerance, existence, resource...).
Much work and we have to watch for new test types on every Monit release.

On the other side Monit has already syntax checks and knows better what works and what not.
Yes, usability is important but I'd rather rely on Monit.

Don't get me wrong. It's a really cool idea.
But is it worth the effort?

If so, I'd implement it only in the UI.
Instead of the Condition textfield I'd add a Test Type dropdown list and
guide the user through all the possible options.

AdSchellevis:
The valid combinations should be in the monit docs, my problem is that the current solution isn't very user friendly, with error messages like : 
![image](https://user-images.githubusercontent.com/9783985/48637622-19583980-e983-11e8-8f42-182a6efac998.png)

Which should be rather easy to prevent using proper validations in my humble opinion.

The original plan for adding Monit to core instead of a plugin was to remove the legacy notifications, which is a good plan given the Monit feature set, but it's a bit tough to use at the moment.
If we're actively going to steer people in using this, I kind of expect some support challenges, @fichtner what do you think?

fichtner:
with the marginal notification support still left any replacement is better than what we have now. I'm happy to work on this, but maybe 19.1 is not the greatest time frame for removing the old ones... maybe just adding a first viable replacement as an optional method and keep improving from there?

architecturally I don't think any way other than too complicated to get started is worth going for.  small steps.. .we can still make it more complicated later?

(please don't expect a technical answer until I know the system well enough)

AdSchellevis:
@fichtner I guess dropping the old could still be viable for 19.1, I really want to prevent any patches like https://github.com/opnsense/core/pull/2907 being needed for stuff that should go away. I expect we can add some simple tests, disabled by default like `RootFS` thing now. 

I'll work on the needed tests, this validation issue can wait until after 19.1 in my opinion.

[...]

PR #3119, opened 2019-01-11
monit: add Service/Test type relation; improve Service Tests UI

fbrendel:
This solves #2918 and should prevent issues like #3063 :smirk: 

I couldn't implement the connect checks. And for all other special
cases I've added a custom condition with the old freestyle behaviour.

For backward compatibility the migration process adds the 'type=custom' property to all tests.


AdSchellevis:
@fbrendel we've merged it partially, I had quite some working aligning the code to our standards (see also https://github.com/opnsense/core/commit/df8d12c8a0e8abb4d3d180f56cc6129b237cb505).

The javascript code is left out of now, it's quite hard to read and seems to be over complicated. I like the idea of offering the configuration options to the user, but can't find the time to refactor that as well. 

It's better now than it was, a very good addition would be to create migration code to fix the known test types after migration (setting everything to Custom isn't very useful). If you have some time to fix that, just let us know.

PR #3130, opened 2019-01-14
Add option to create combined client and server certificate, closes #2800

### LINKED ISSUE #2800: Feature Request: Create Certificates with Both Server & Client Extended Key Usage
When creating a certificate in Trust section of OPNsense, you have to select either Server, Client (or CA) as certificate extended key usage (Type row).

Now I am confronted with a software problem where I need both usage types within the same certificate.
See here: [Problem with solution](https://community.hortonworks.com/content/supportkb/151790/errorjavasecuritycertcertificateexception-javaxnet.html)

Unfortunately, it is not possible to generate such a certificate with OPNsense, which means that I have to generate the certificate outside of OPNsense.

It would be nice if it is possible to select multiple values in the Type of the certificate dropdown so that it is possible to generate certificates with server and client extended key usage for cluster servers and nodes that authenticate and serve with the same certificate.

Thanks!

MichaelDeciso:
Confirmed to work by the OP of #2800.

### DIFF
--- a/src/etc/ssl/opnsense.cnf
+++ b/src/etc/ssl/opnsense.cnf
@@ -373,7 +373,6 @@ keyUsage = digitalSignature, keyEncipherment
 [ combined_server_client ]
 
 basicConstraints=CA:FALSE
-nsCertType			   = server, client
 nsComment			   = "OPNsense Generated Combined Client/Server Certificate"
 subjectKeyIdentifier   = hash
 authorityKeyIdentifier = keyid,issuer:always
--- a/src/etc/ssl/opnsense.cnf
+++ b/src/etc/ssl/opnsense.cnf
@@ -369,3 +369,16 @@ extendedKeyUsage=serverAuth,1.3.6.1.5.5.8.2.2
 keyUsage = digitalSignature, keyEncipherment
 
 ###OPNsense:subjectAltName###
+
+[ combined_server_client ]
+
+basicConstraints=CA:FALSE
+nsCertType			   = server, client
+nsComment			   = "OPNsense Generated Combined Client/Server Certificate"
+subjectKeyIdentifier   = hash
+authorityKeyIdentifier = keyid,issuer:always
+extendedKeyUsage       = clientAuth,serverAuth,1.3.6.1.5.5.8.2.2
+# This is typical in keyUsage for a client certificate.
+keyUsage               = nonRepudiation, digitalSignature, keyEncipherment
+
+###OPNsense:subjectAltName###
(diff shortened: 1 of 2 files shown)


src/www/system_certmanager.php
```diff
@@ -864,6 +864,7 @@ function internalca_change() {
                     <select name="cert_type">
                         <option value="usr_cert" <?=$pconfig['cert_type'] == 'usr_cert' ? 'selected="selected"' : '';?>> <?=gettext("Client Certificate");?> </option>
                         <option value="server_cert" <?=$pconfig['cert_type'] == 'server_cert' ? 'selected="selected"' : '';?>> <?=gettext("Server Certificate");?> </option>
+                        <option value="combined_server_client" <?=$pconfig['cert_type'] == 'combined_server_client' ? 'selected="selected"' : '';?>> <?=gettext("Combined Client/Server Certificate");?> </option>
```
fichtner:
```suggestion
                        <option value="combined_server_client" <?=$pconfig['cert_type'] == 'combined_server_client' ? 'selected="selected"' : '';?>> <?=gettext("Client/Server Certificate");?> </option>
```

MichaelDeciso: Are you sure this is clear enough to our average user?

src/etc/ssl/opnsense.cnf
```diff
@@ -369,3 +369,16 @@ extendedKeyUsage=serverAuth,1.3.6.1.5.5.8.2.2
 keyUsage = digitalSignature, keyEncipherment
 
 ###OPNsense:subjectAltName###
+
+[ combined_server_client ]
+
+basicConstraints=CA:FALSE
+nsCertType			   = server, client
```
fichtner: there was chatter a while back that says we don't need "nsCertType" and "nsComment" anymore.

FWIW, server, client seems wrong while the user said order matters but since it's most likely unused it might be a good start to omit these and go from there?

src/www/system_certmanager.php
fichtner: Looking at the drop down choices all at once, yes. "Combined" is a kind of doubling.

src/etc/ssl/opnsense.cnf
MichaelDeciso: I have looked on the internet and `nsCertType` indeed appears to be unused nowadays. Most likely, `extendedKeyUsage` now takes care of it.

Not sure if dropping `nsComment` is wise - I haven't found an alternative for it, and I still think it could be useful to have a comment.

src/www/system_certmanager.php
fichtner: the shorter the text the clearer the meaning. factor in translators who will "interpret" the meaning of the string. just want to point out best practice so deviating from it for a valid reason is a conscious decision and not just a gut feeling. your call, I don't mind either way.

MichaelDeciso: Ah, I see. Well, I think the original text is also clearer to translators (especially if they are not fully aware of the context). If you are okay with keeping the original text, then I'd like to keep it.

fichtner: Okay, sure

MichaelDeciso:
With the phrasing of the option discussed and `nsCertType` removed, is there anything else I need to do?

fichtner:
Nope, will merged. Thanks!

PR #3134, opened 2019-01-15
Create MVC log viewer infrastructure

MichaelDeciso:
This should allow for the elimination of almost 30 old-style pages. For now, I have only ported syslog, configd.log, lighttpd.log and ppps.log.

### DIFF
--- a/src/opnsense/scripts/diagnostics/deletelog.php
+++ b/src/opnsense/scripts/diagnostics/deletelog.php
@@ -1,0 +1,49 @@
+#!/usr/local/bin/php
+<?php
+/**
+ *    Copyright (C) 2019 Deciso B.V.
+ *
+ *    All rights reserved.
+ *
+ *    Redistribution and use in source and binary forms, with or without
+ *    modification, are permitted provided that the following conditions are met:
+ *
+ *    1. Redistributions of source code must retain the above copyright notice,
+ *       this list of conditions and the following disclaimer.
+ *
+ *    2. Redistributions in binary form must reproduce the above copyright
+ *       notice, this list of conditions and the following disclaimer in the
+ *       documentation and/or other materials provided with the distribution.
+ *
+ *    THIS SOFTWARE IS PROVIDED ``AS IS'' AND ANY EXPRESS OR IMPLIED WARRANTIES,
+ *    INCLUDING, BUT NOT LIMITED TO, THE IMPLIED WARRANTIES OF MERCHANTABILITY
+ *    AND FITNESS FOR A PARTICULAR PURPOSE ARE DISCLAIMED. IN NO EVENT SHALL THE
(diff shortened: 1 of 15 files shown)


src/opnsense/scripts/diagnostics/deletelog.php
```diff
@@ -0,0 +1,21 @@
+#!/usr/local/bin/php
+<?php
```
fichtner: license missing

src/opnsense/scripts/diagnostics/viewlog.php
```diff
@@ -0,0 +1,74 @@
+#!/usr/local/bin/php
+<?php
```
fichtner: license missing

AdSchellevis:
not going to add this now, doesn't really bring something new to the table.

src/opnsense/scripts/diagnostics/viewlog.php
```diff
@@ -0,0 +1,102 @@
...
+echo json_encode($result) . PHP_EOL;
+exit(0);
+
+function doFileChecks($logfile)
+{
+    if (is_dir($logfile)) {
+        echo json_encode(['status' => sprintf(gettext('File %s is a directory.'), $logfile)]) . PHP_EOL;
+        exit(1);
+    }
+    if (!file_exists($logfile)) {
```
fabianfrz: is_readable

src/opnsense/scripts/diagnostics/viewlog.php
```diff
@@ -0,0 +1,102 @@
...
+
+if ($type == 'clog') {
+    exec("/usr/local/sbin/clog " . escapeshellarg($logfile) . " {$grepline}| grep -v \"CLOG\" | grep -v \"\033\" | /usr/bin/tail {$tailParam} -n " . escapeshellarg($numlines), $logarr);
+} else {
+    exec("cat " . escapeshellarg($logfile) . " {$grepline} | /usr/bin/tail {$tailParam} -n " . escapeshellarg($numlines), $logarr);
+}
+
+$result = checkResult($logarr, $logfile);
+
+echo json_encode($result) . PHP_EOL;
```
fabianfrz: the eol is not needed, isn't it?

fabianfrz:
@AdSchellevis it brings at least privilege separation. Otherwise, @MichaelDeciso may take a look at mine under https://github.com/opnsense/plugins/blob/master/www/nginx/src/opnsense/www/js/nginx/src/logviewer.js, which is a more powerful log viewer on client side.

AdSchellevis:
@fabianfrz we're not going to do this now, it could be simplified (probably a lot), but costs us all valuable time at the moment. We could probably handle this all with only a couple of controllers, reducing our footprint a lot.

PR #3138, opened 2019-01-19
security/usermapping: add groups to local auth, allow custom overrides of before exec route

fabianfrz:
required for my new plugin, which is a WIP


AdSchellevis:
Since authentication should eventually be able to only travel through pam or its helper program (in order to cut privileges on the web interface) and we don't have an active use-case for this feature, we should probably not do this now. 

In preparation of further auth detachment we added https://github.com/opnsense/core/commit/875a1e2d4a90aa83ec7d216520a084344853ec28 last year, but OpenVPN still depends on direct config access for authentication (https://github.com/opnsense/core/issues/3266).

Issue #3126, opened 2019-01-13
Dashboard: Monit Widget 6 Columns pagination view problem


opnsenseuser:
![6columns_monit_widget](https://user-images.githubusercontent.com/34602360/51083209-6142d500-1716-11e9-9de0-1d37dfee7dc7.jpg)

fixed should look like this:
![fixed](https://user-images.githubusercontent.com/34602360/51083214-6d2e9700-1716-11e9-88ba-44d158c4c059.jpg)

fabianfrz:
this is a dangerous css option but probably does what you expect (white-space: no-wrap):
https://www.w3schools.com/cssref/playit.asp?filename=playcss_white-space&preval=nowrap

opnsenseuser:
@fabianfrz maybe it would be even better to take away the pagination and replace it with a scrollbar, if the list is longer than .... whatever!!

fbrendel:
Yes, a scrollbar would be better than the pagination.
But that requires a fixed widget height. How high should it be?
@AdSchellevis can we set it within the css of the theme?

AdSchellevis:
@fbrendel it would probably be best to start with a class added in the widget itself, so the theme can overwrite it if needed. I know it's suboptimal, but custom theme's might render odd defaults otherwise.

MichaelDeciso:
It looks like this has been fixed (at least on my dashboard) by #3147 and this can be closed.

fichtner:
Indeed, thanks for the heads-up.

PR #3147, opened 2019-01-23
monit: widget improvements

fbrendel:
This PR solves #3126.

Changes:
- fix issue with deleting services via GUI
- replace the pagination with a scrollbar
- make the number of rows to be shown configurable
- add a search filter

Unfortunately I have two small problems that I cannot solve.
The number of rows that are effectively shown is based on a rough estimation of the rendered row height. That can lead to different table heights on different themes.

And there is no translation for the word 'Rows'.

### DIFF
--- a/src/opnsense/mvc/app/views/OPNsense/Monit/index.volt
+++ b/src/opnsense/mvc/app/views/OPNsense/Monit/index.volt
@@ -211,7 +211,7 @@
          'get':'/api/monit/settings/getService/',
          'set':'/api/monit/settings/setService/',
          'add':'/api/monit/settings/addService/',
-         'del':'/api/monit/settings/del/service/',
+         'del':'/api/monit/settings/delService/',
          'toggle':'/api/monit/settings/toggleService/'
       });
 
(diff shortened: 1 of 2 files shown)


src/opnsense/mvc/app/views/OPNsense/Monit/index.volt
```diff
@@ -211,7 +211,7 @@
          'get':'/api/monit/settings/getService/',
          'set':'/api/monit/settings/setService/',
          'add':'/api/monit/settings/addService/',
-         'del':'/api/monit/settings/del/service/',
+         'del':'/api/monit/settings/delService/',
```
AdSchellevis: @fbrendel nice catch, thanks!

AdSchellevis:
@fbrendel if this version is better than what's in master now, I'm considering to pull it in now and see how we can manage the issues later. At the moment I don't have enough time available to review/test in detail. What do you think?

fbrendel:
I think it's much better than the pagination and it's worth the vague computation of the row height.
That's really not that bad. I just wanted to mention it.
The changes are tested with Chrome 71, Firefox 60 and Internet Explorer 11.
So I'd merge it. :smiley:

AdSchellevis:
@fbrendel thanks!

PR #3150, opened 2019-01-24
monit: bug fix

fbrendel:
Issues fixed:
- wrong test type validation when saving a service
- unable to delete services via GUI (already fixed in #3147 but not merged yet)
- view/hide fields in modals broken 

The field id's for the General tab are prefixed with monit (rows have row_monit).
In the other tabs they aren't.
Is that intended?

### DIFF
--- a/src/opnsense/mvc/app/models/OPNsense/Monit/Monit.php
+++ b/src/opnsense/mvc/app/models/OPNsense/Monit/Monit.php
@@ -42,18 +42,18 @@ class Monit extends BaseModel
      *
      */
     private $testSyntax = [
-        'process'    => ['Existence', 'Process Resource', 'Process Disk I/O',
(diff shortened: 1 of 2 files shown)


AdSchellevis:
@fbrendel thanks, the prefix is different indeed since the general setter uses the full modal path prefixed with https://github.com/opnsense/core/blob/6d11ffc2ab374b7f0db19cf6f1bf8aba495ddd8f/src/opnsense/mvc/app/controllers/OPNsense/Monit/Api/SettingsController.php#L43

Items setters use a relative path, defined by https://github.com/opnsense/core/blob/6d11ffc2ab374b7f0db19cf6f1bf8aba495ddd8f/src/opnsense/mvc/app/controllers/OPNsense/Monit/Api/SettingsController.php#L65

Your fixes look good, maybe we can simplify things a bit at some point in time, but since we likely wish to add the condition builder at some point in time we improve things later if needed.

Thanks!

PR #3154, opened 2019-01-25
monit: remove Test Type from GUI until validations are complete

fbrendel:


fichtner:
ah I see. since it does come back at some point, wouldn't it be better to wrap it in an XML comment block?

fichtner:
merged, thanks!

PR #3155, opened 2019-01-25
monit: add validation for test type

fbrendel:
This PR adds test type validation.
It simply checks if a service has a link to a test and refuses to change the type of the test if so.
That means if you want to change the type of a test you have either to remove all links from services or delete and recreate it which does the same in turn.
If I'm not mistaken deleting an alias which is in use by a rule has a similar behavior.

The validation is not performed on full model validations only if the field has changed.
The test type - service type validation is made by the service validation.

I've put it in an extra model function isTestServiceRelated to enable/disable the Test Type dropdown in the GUI later based on the result of this function.
(Don't know how to grab the uuid yet :smiley: )


AdSchellevis:
@fbrendel I finally found a bit of time to check your pull request, did some small cleanups in https://github.com/opnsense/core/commit/5bde17012df138d36a211433db7704749e0b391e, looks good further. Thanks!

Issue #3173, opened 2019-02-02
19.1 upgrade sets WAN MTU to 576 bytes


mimugmail:
I followed the reports about strange behavior with some guys after updating. When you check forums or Reddit it seems the WAN MTU was set to 576 bytes and reconfigure fixes it. I cant reproduce, perhaps it's only related to Realtek as most users reporting had them. 

I also had 2 long Teamviewer sessions with one of them and I'm quite sure it's not a local configuration issue.

Any ideas?

immanuelfodor:
Setting it to 1500 via web gui on the wan interface (previously that box was empty = default) solved all my issues after 18.07->19.1 upgrade. I have Intel NICs though.

mimugmail:
Ah, you are the reddit guy :) Thanks for reporting. I myself am not affected, I'll update some more boxes ...

immanuelfodor:
Yeah, I looked for the issue right away to subscribe if I can help with something to the team if there are any questions.

mimugmail:
https://redmine.pfsense.org/issues/8507

immanuelfodor:
Removed the custom 1500 MTU setting from the GUI (empty box), applied it and all my issues came back, slow page load speeds, inaccessible websites, etc. I didn't even had to reboot, the _evil_ 576 appeared instantly.

```bash
ifconfig # of WAN interface
vtnet1: flags=8843<UP,BROADCAST,RUNNING,SIMPLEX,MULTICAST> metric 0 mtu 576
        options=c00b8<VLAN_MTU,VLAN_HWTAGGING,JUMBO_MTU,VLAN_HWCSUM,VLAN_HWTSO,LINKSTATE>
        ether xxxxxxxx
        hwaddr xxxxxxxx
        inet6 xxxxxxxx%vtnet1 prefixlen 64 scopeid 0x2
        inet xxxxxxxxx netmask 0xfffffc00 broadcast 255.255.255.255
        nd6 options=21<PERFORMNUD,AUTO_LINKLOCAL>
        media: Ethernet 10Gbase-T <full-duplex>
        status: active

netstat -4rnW
Routing tables

Internet:
Destination        Gateway            Flags       Use    Mtu      Netif Expire
default            xxxxxxxxx     UGS         181    576     vtnet1
127.0.0.1          link#4             UH        45217  16384        lo0
xxxxxxxx    link#2             U           690    576     vtnet1
xxxxxxxx      link#2             UHS         285  16384        lo0
[...]

PR #3178, opened 2019-02-03
DHCP Overide MTU Option for #3173

marjohn56:
It seems DHCP in 11.2 is honouring the ISPs MTU if it is sent. It also seems there are some ISPs who send a stupid value. This addition allows the user to ignore the ISP supplied MTU . A checkbox has been added to the dhcp settings in WAN, this then adds a 'supersede interface-mtu 0' to the config creation. It can also be done by adding the said string in the option modifiers.

Use it or abuse it..

### DIFF
--- a/src/etc/inc/interfaces.inc
+++ b/src/etc/inc/interfaces.inc
@@ -3237,7 +3237,9 @@ alias {
 
 EOD;
     }
-
+    if (isset($wancfg['dhcpoveridemtu'])) {
+        $dhclientconf .= "supersede interface-mtu 0;\n";      
+    }
     // DHCP Config File Override
     if (!empty($wancfg['adv_dhcp_config_file_override'])) {
         $dhclientfile = $wancfg['adv_dhcp_config_file_override_path'];
--- a/src/www/interfaces.php
+++ b/src/www/interfaces.php
@@ -436,6 +436,7 @@ if ($_SERVER['REQUEST_METHOD'] === 'GET') {
     $pconfig['lock'] = isset($a_interfaces[$if]['lock']);
     $pconfig['blockpriv'] = isset($a_interfaces[$if]['blockpriv']);
     $pconfig['blockbogons'] = isset($a_interfaces[$if]['blockbogons']);
+    $pconfig['dhcpoveridemtu'] = isset($a_interfaces[$if]['dhcpoveridemtu']);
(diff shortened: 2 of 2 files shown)


fichtner:
I think we should flip-flop this and make it opt-out to retain compatibility with the way that dhclient handles this on 11.2? I'm unsure the behaviour will change because it seems to indicate this is an "intended fix" for some thing that didn't really need fixing so badly. ;(

fichtner:
cc @AdSchellevis

marjohn56:
OK, So you want me to make it inverse, tick the box to opt out?

fichtner:
Just discuss for now. What my motivation for the inversion would be is to make upgrades from 18.7 work as they would be intended?

marjohn56:
Ah.. well I've reversed it now. :)

I think probably reversed is best. It appears that the change to dhcp has broken more than its fixed.

AdSchellevis:
@fichtner I tend to agree, although it's a bit odd this. https://github.com/freebsd/freebsd/commit/4ece63474c38029e5fd9512af1e8b1db35fef3b5 seems to  disable https://github.com/freebsd/freebsd/commit/413d7799506ad5f6ed856a7e68590cb726f150a9, since some providers send "random" mtu values... it feels odd, but I think we should keep the default disabled as you suggested.

fichtner:
Default disabled as in do not use the ISP MTU by default?

AdSchellevis:
yes, which was the behaviour before 19.1 (reversed logic like you suggested)

fichtner:
Neat, thanks guys. I'll merge this later today.

marjohn56:
typo fix

src/etc/inc/interfaces.inc
```diff
@@ -3237,7 +3237,11 @@ alias {
 
 EOD;
     }
-
+    if (!isset($wancfg['dhcpoveridemtu'])) {
+        $dhclientconf .= <<<EOD
+supersede interface-mtu 0;      
+EOD;
```
fabianfrz: you don't need the heredoc syntax here - just use a double quoted string with \n

fichtner: it's ok either way. no need for further changes as I'll merge this manually

fichtner:
Small adjustments via 28796e83a61c9513e7, you know me :P

Please give it a test. Looks ok from my end in dhcp basic override MTU on / off and dhcp advanced override MTU on / off.

PR #3190, opened 2019-02-05
Add validation check for time period for Dpinger

marjohn56:
Time period needs to be greater than 2.1 times the probe interval.


AdSchellevis:
@marjohn56 thanks!

PR #3204, opened 2019-02-08
Allow explicitly setting Source Hash Key

fredronnv:
This key is randomly assigned unless set every time the ruleset is
loaded. In order to make it persist, allow setting it to a predetermined
value.

### DIFF
--- a/src/opnsense/mvc/app/library/OPNsense/Firewall/SNatRule.php
+++ b/src/opnsense/mvc/app/library/OPNsense/Firewall/SNatRule.php
@@ -51,6 +51,7 @@ class SNatRule extends Rule
         'target' => 'parsePlain, -> ',
         'natport' => 'parsePlain, port ',
         'poolopts' => 'parsePlain',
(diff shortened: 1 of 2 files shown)


fredronnv:
This most likely needs a bit of tidying, which I'm happy to do in case it seems like a worthwhile merge. It's at this stage mostly a quick proof of concept. Perhaps it has other impacts I'm not aware of?

AdSchellevis:
@fredronnv what is the use-case for this this?

fredronnv:
> @fredronnv what is the use-case for this this?

When using a pool as NAT translation, in our case a /23, we like to have a predictable translation for clients, so a particular client will always be translated to the same external address, even if the ruleset is modified. Currently if you choose source-hash, pfctl will generate a random value as the source hash key every time the ruleset is loaded, which means that clients will not maintain their external address translation.

AdSchellevis:
sounds reasonable, let me look at this.

AdSchellevis:
@fredronnv thanks! did a small cleanup while there.

PR #3209, opened 2019-02-09
www/nginx: fix html escaping issue and clean up the file

fabianfrz:
see: opnsense/plugins#1175

### DIFF
--- a/src/opnsense/www/js/opnsense.js
+++ b/src/opnsense/www/js/opnsense.js
@@ -30,12 +30,17 @@
  *
  */
 
+// jQuery window object - cached object is faster
+var $window = $(window);
+
+
  /**
   * html decode text into textarea tag and return decoded value.
   *
   * @param value encoded text
   * @return string decoded text
   */
+
 function htmlDecode(value) {
     return $("<textarea/>").html(value).text();
 }
@@ -51,7 +56,7 @@ function htmlDecode(value) {
 function getFormData(parent) {
 
     var data = {};
-    $( "#"+parent+"  input,#"+parent+" select,#"+parent+" textarea" ).each(function( index ) {
+    $("#"+parent+" input,#"+parent+" select,#"+parent+" textarea" ).each(function() {
         if ($(this).prop('id') === undefined || $(this).prop('id') === "") {
             // we need an id.
             return;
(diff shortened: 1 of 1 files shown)


src/opnsense/www/js/opnsense.js
```diff
@@ -120,32 +125,32 @@ function getFormData(parent) {
...
+                            if (+keyItem["selected"] !== 0) {
                                 opt.attr('selected', 'selected');
                             }
                             targetNode.append(opt);
                         });
-                    } else if (targetNode.prop("type") == "checkbox") {
+                    } else if (targetNode.prop("type") === "checkbox") {
                         // checkbox type
-                        targetNode.prop("checked", node[keypart] != 0);
+                        targetNode.prop("checked", +node[keypart] !== 0);
```
AdSchellevis: `+`node[keypart] ? why the [+]? shouldn't this compare to `"0"` in your case?

src/opnsense/www/js/opnsense.js
```diff
@@ -120,32 +125,32 @@ function getFormData(parent) {
...
                     node = node[keypart];
                 } else {
                     // data node found, handle per type
                     if (targetNode.is("select")) {
                         // handle select boxes
                         targetNode.empty(); // flush
                         $.each(node[keypart],function(indxItem, keyItem){
-                            var opt = $("<option>").val(indxItem).text(keyItem["value"]);
-                            if (keyItem["selected"] != "0") {
+                            var opt = $("<option>").val(htmlDecode(indxItem)).text(keyItem["value"]);
```
AdSchellevis: seems to be working fine, thanks!

src/opnsense/www/js/opnsense.js
fabianfrz: + is a shortcut for convert to a numeric value so it compares 0!==0 instead of "0"!==0.

this prevents some questionable true like comparisons:

```js
""==0
true

+""==0
true

+"a"==0 // the value left is NaN in this case
false
```

src/opnsense/www/js/opnsense.js
fabianfrz: If you don't like the rest, just take that line - This is important not to break the frontend for API values containing HTML characters.

AdSchellevis:
@fabianfrz thanks, I've changed it a bit in https://github.com/opnsense/core/commit/b3a6440aca7c7fb8c26ad5f36985b8e0a1aa07c6 and pulled in the fix

PR #3213, opened 2019-02-10
Shaper: Add DSCP matching

mimugmail:
There was a request on Friday if we support DSCP matching for shaper. @fichtner and me said no, only for `pf` but then I checked ipfw manpage and it looks good.
The beauty of this is, that you can have multiple rules with match any any as you can match for differenct DSCP values, mostly set by SIP phones automatically. Would be a nice addition for enterprise demands.
If pf acts before ipfw in shared forwarding you could also mark rules with pf for DSCP and do the shaping afterwards :)

### DIFF
--- a/src/opnsense/mvc/app/controllers/OPNsense/TrafficShaper/forms/dialogRule.xml
+++ b/src/opnsense/mvc/app/controllers/OPNsense/TrafficShaper/forms/dialogRule.xml
@@ -69,6 +69,13 @@
         <help>Destination port number or well known name (imap, imaps, http, https, ...), for ranges use a dash</help>
     </field>
     <field>
+        <id>rule.dscp</id>
+        <label>DSCP</label>
+        <type>select_multiple</type>
+        <advanced>true</advanced>
+        <help>Match against one or multiple DSCP values.</help>
+    </field>
+    <field>
         <id>rule.direction</id>
         <label>direction</label>
         <type>dropdown</type>
--- a/src/opnsense/mvc/app/models/OPNsense/TrafficShaper/TrafficShaper.xml
+++ b/src/opnsense/mvc/app/models/OPNsense/TrafficShaper/TrafficShaper.xml
@@ -305,6 +305,33 @@
                     <EnableRanges>Y</EnableRanges>
                     <default>any</default>
                 </dst_port>
+                <dscp type="OptionField">
+                    <Required>N</Required>
+                    <Multiple>Y</Multiple>
+                    <OptionValues>
+                        <be>Best Effort</be>
+                        <ef>Expedited Forwarding</ef>
+                        <af11>AF11</af11>
+                        <af12>AF12</af12>
+                        <af13>AF13</af13>
+                        <af21>AF21</af21>
+                        <af22>AF22</af22>
+                        <af23>AF23</af23>
+                        <af31>AF31</af31>
+                        <af32>AF32</af32>
+                        <af33>AF33</af33>
+                        <af41>AF41</af41>
+                        <af42>AF42</af42>
+                        <af43>AF43</af43>
+                        <cs1>CS1</cs1>
+                        <cs2>CS2</cs2>
+                        <cs3>CS3</cs3>
+                        <cs4>CS4</cs4>                
+                        <cs5>CS5</cs5>
+                        <cs6>CS6</cs6>
+                        <cs7>CS7</cs7>
+                    </OptionValues>
+                </dscp>
                 <direction type="OptionField">
                     <Required>N</Required>
                     <default></default>
--- a/src/opnsense/mvc/app/models/OPNsense/TrafficShaper/TrafficShaper.xml
+++ b/src/opnsense/mvc/app/models/OPNsense/TrafficShaper/TrafficShaper.xml
@@ -1,6 +1,6 @@
 <model>
     <mount>//OPNsense/TrafficShaper</mount>
-    <version>1.0.2</version>
+    <version>1.0.3</version>
     <description>
         OPNsense traffic shaper
     </description>
--- a/src/opnsense/service/templates/OPNsense/IPFW/ipfw.conf
+++ b/src/opnsense/service/templates/OPNsense/IPFW/ipfw.conf
@@ -184,8 +184,9 @@ add {{loop.index + 60000}} {{ helpers.getUUIDtag(rule.target) }} {{
     if rule.destination_not|default('0') == '1' %}not {% endif %}{{rule.destination
     }} src-port  {{ rule.src_port }} dst-port {{ rule.dst_port }} recv {{
     physical_interface(rule.interface) }} {%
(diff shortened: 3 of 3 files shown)


muchahagrande:
Would be "THE SOLUTION" I was expecting for.

mimugmail:
Sounds like a volunteer for beta testing :)

muchahagrande:
Hehe... sure... I can do it.

AdSchellevis:
@mimugmail thanks, the ruleset looks decent, I haven't tried the feature itself, but when in master it's easier for others to do more testing.

mimugmail:
@AdSchellevis thanks for merge!
@muchahagrande can you test this with current master?

muchahagrande:
Excuse my ignorance on the subject... may be I'm gonna ask a couple of stupid questions.
As I have OPNSense virtualized on ESXi, I've created a copy of my router to do the testing.
The first thing I did is switching to development branch. Not sure if it is the rigth step.
But I cant find the new dialog in traffic shaper rules menu.
How can I test with the current master?

fichtner:
# opnsense-code core
# cd /usr/core
# make upgrade

muchahagrande:
Ok, in the "make upgrade" it stopped with the following error:

...
...
All repositories are up to date.
pkg: No packages available to install matching 'squid' have been found in the repositories
*** Error code 70

Stop.
make: stopped in /usr/core

I still cant find DSCP options in traffic shapper

mimugmail:
@muchahagrande 
make upgrade CORE_SQUID=3

Then you'll see it :)

muchahagrande:
With "make upgrade CORE_SQUID=3" the upgrade went OK.
But I still can't see DSCP options in shaper menu :-(

mimugmail:
@muchahagrande in rules you have to tick "advanced" to see the dropdown multiselect.

muchahagrande:
It it ticked...
May be something went wrong with the update process.
Going to try it again with a fresh restored virtual machine.
Additionally, after the upgrade and a reboot, I couldn't login with SSH anymore.

muchahagrande:
Great... beggining from a fresh VM backup did the trick.

Before I got "OPNsense\TrafficShaper\TrafficShaper 1.0.2"
Now I see "Migrated OPNsense\TrafficShaper\TrafficShaper from 1.0.2 to 1.0.3"

I can see DSCP option on traffic shaper rules.
I'll do some testing trying to separate my PBX traffic from the ordinary one.

muchahagrande:
Is there an accurate diagnostic for IPFW ?
With the slow refresh of the traffic shaping status I can't be sure of the results of my testing.

mimugmail:
Create 3 pipes, all rule match any any but different dscp. The start a vioce call and you should see matching traffic for this pipe. The switch the value to other pipes to check against

muchahagrande:
I made two pipes one for 8 Mbit and the other for 2 Mbit.
I created two rules on WAN with outgoing direction and matching any any.
The first rule with DSCP=CS3 using the 2 Mbit limiter for VoIP and the second without DSCP using the 8 Mbit pipe for the rest of the traffic.
The first rule is catching all the traffic and sending it through the 2 Mbit pipe like it were ignoring the presence of the DSCP attribute.
I made a speedtest and it's limiting my upload to 2 Mbit when it should be 8 Mbit using the second rule.

mimugmail:
I'll try to reproduce :/

mimugmail:
Sidenote:

@AdSchellevis @fichtner @fabianfrz 
This comes when updating to latest master in system.log:

`Mar  1 12:45:48 OPNsense /configd.py: Handler died on Traceback (most recent call last):   File "/usr/local/opnsense/service/modules/processhandler.py", line 87, in run     config_environment=self.config_environment)   File "/usr/local/opnsense/service/modules/__init__.py", line 37, in getinstance     instances[cls] = cls(*args, **kwargs)   File "/usr/local/opnsense/service/modules/processhandler.py", line 255, in __init__     self.load_config()   File "/usr/local/opnsense/service/modules/processhandler.py", line 272, in load_config     cnf.read(config_filename)   File "/usr/local/lib/python3.6/configparser.py", line 697, in read     self._read(fp, filename)   File "/usr/local/lib/python3.6/configparser.py", line 1066, in _read     lineno) configparser.DuplicateSectionError: While reading from '/usr/local/opnsense/service/conf/actions.d/actions_nginx.conf' [line 35]: section 'log' already exists
`

AdSchellevis:
@mimugmail configd is right here :) https://github.com/opnsense/plugins/blob/127fe3ddfebd8980321f656bfcf90f2fe8a1cc3b/www/nginx/src/opnsense/service/conf/actions.d/actions_nginx.conf#L29-L39, something for @fabianfrz to fix

mimugmail:
@muchahagrande can you fetch latest master and try again .. thx to @AdSchellevis it's working now.
Previously the line in config was:

add 60001 pipe 10000 ip from any to any src-port  any dst-port any    dscp ['af11', 'af21'] via em0 // wan: pipe1

and since this is wrong ipfw just replace it with best effort. Now it's working fine on my end.

fichtner:
nice, this means hotfix nginx plugin -.-

fichtner:
ah, nevermind... this is a new Python 3.6 error it seems so we don't have to hotfix... :)

AdSchellevis:
it's not an issue with python, it could be that python 2 was a bit more forgiving in this regard, but duplicating sections isn't a guarantee to success :)

muchahagrande:
@mimugmail what's the command to fetch the master?

fabianfrz:
opnsense-patch

muchahagrande:
@mimugmail, after a couple of tests I can confirm that is working fine.
I made two pipes for upload, one for 2 Mbit and the other for 8 Mbit. The sum of both is the total upload bandwidth I can get from my ISP.
Then I made two rules to redirect VoIP traffic through the first pipe and the rest through the second one.
The filter criteria to catch the PBX traffic was only DSCP=Expedited Forwarding,AF41,CS3 that corresponds to voice, video and sip packets.
I made some calls and I saw packets passing through the 2 Mbits pipe. After that I made a speed test and the upload was limited to 8 Mbits using the second pipe.
Finally I made a speed test while making a call, and both traffics were conducted through their respective pipes giving a clean voice without interruptions.

I think that ipfw with DSCP is working as expected.

muchahagrande:
@mimugmail 

> If pf acts before ipfw in shared forwarding you could also mark rules with pf for DSCP and do the shaping afterwards :)

Do pf rules already support marking packets with DSCP?

mimugmail:
Yes, just look at the rules, but I'm unsure If marking happens before ipfw. @fichtner ?

muchahagrande:
I think that the marking matched and set on the pf rules page referes to 802.1Q VLAN header that happens at ethernet level, but DSCP is another ToS marking that hapens at IP header level. To use DSCP at the pf rules page there should be some combo boxes to match/set DSCP in addition to the ones that are already there.

PR #3220, opened 2019-02-11
monit: add migration/validation for service/test type dependency

fbrendel:
This PR implements the service/test type dependency.
Conditions are mapped to test types and these test types in turn are mapped to service types.

It prevents the user from adding wrong tests to a service.
That means that you still can select a wrong test for a service but get a validation error.
Furthermore, a test condition can only be changed if the test is not linked to a service or the condition results in 'Custom' type.
Every condition that cannot mapped to a type is a 'Custom' type.

The migration parses the condition field and sets the test type accordingly.
To bypass validation errors I use the post() method instead of run().


src/opnsense/mvc/app/models/OPNsense/Monit/Migrations/M1_0_8.php
```diff
@@ -0,0 +1,44 @@
...
+use OPNsense\Core\Config;
+
+class M1_0_8 extends BaseModelMigration
+{
+    public function post($model)
+    {
+        foreach ($model->getNodeByReference('test')->__items as $test) {
+            $test->type = $model->getTestType($test->condition->getNodeData());
+        }
+        $model->serializeToConfig(false, true);
```
AdSchellevis: nice catch, this would indeed break with the previous validation you added. It would be good to add a note here, the reason is valid, but easily overlooked :)

AdSchellevis:
@fbrendel can you rebase? it seems this pull breaks with the previous one committed.

fbrendel:
I've tried a rebase but that looks a bit strange. :confused: 
@AdSchellevis is that ok or should I make a new PR?

AdSchellevis:
@fbrendel rather a new PR :)

PR #3221, opened 2019-02-11
rspamd: fix constraint

### LINKED ISSUE #3219: Constraint: ComparedToFieldConstraint also need checks when field is empty
For rspamd plugin @fabianfrz introduced ComparedToFieldConstraint. There is a little problem with it, as it also compares when the fields are empty:

https://github.com/opnsense/plugins/blob/master/mail/rspamd/src/opnsense/mvc/app/models/OPNsense/Rspamd/RSpamd.xml#L14-L113

Required = no but the fields are not allowed to be empty :)

fabianfrz:
fixes #3219

See also: https://forum.opnsense.org/index.php?topic=11572.0;topicseen


lucas12433:
Works perfect!!!

PR #3229, opened 2019-02-13
OPNVpn Stale PID fix

marjohn56:
This patch checks that the instnace of OpenVPN that should have shut down has shut down, if not it forces it down.

It then checks to see if the PID is still there and deletes it if required,

### DIFF
--- a/src/etc/inc/plugins.inc.d/openvpn.inc
+++ b/src/etc/inc/plugins.inc.d/openvpn.inc
@@ -1004,6 +1004,18 @@ function openvpn_restart($mode, $settings)
 
     @unlink("/var/etc/openvpn/{$mode_id}.sock");
 
+     // has it shut down? Check the validity of the pid
+    if(isvalidpid("/var/run/openvpn_{$mode_id}.pid")) {
+        //Still running - force it down and wait a second or two
+        killbypid("/var/run/openvpn_{$mode_id}.pid", 'KILL', false); 
+        sleep(2);
+    }
+    // Remove any stray pid file.
+    if(file_exists("/var/run/openvpn_{$mode_id}.pid"))
+    {
+        @unlink("/var/run/openvpn_{$mode_id}.pid");
+    }
+ 
     /* start the new process */
     $fpath = "/var/etc/openvpn/{$mode_id}.conf";
     openvpn_clear_route($mode, $settings);


marjohn56:
I've not had ant feedback on this from my tester who reports he has the issue, seems he has gone AWOL!

However, lets put the patch out there and see what gives.

AdSchellevis:
ok to test, needs some work if this fixes the issue (psr/style cleanup)

src/etc/inc/plugins.inc.d/openvpn.inc
```diff
@@ -1004,6 +1004,18 @@ function openvpn_restart($mode, $settings)
 
     @unlink("/var/etc/openvpn/{$mode_id}.sock");
 
+     // has it shut down? Check the validity of the pid
```
fichtner: this is weird, like 983 already does:

killbypid("/var/run/openvpn_{$mode_id}.pid", 'TERM', true);

src/etc/inc/plugins.inc.d/openvpn.inc
```diff
@@ -1004,6 +1004,18 @@ function openvpn_restart($mode, $settings)
 
     @unlink("/var/etc/openvpn/{$mode_id}.sock");
```
fichtner: maybe we only need the pid unlink for the pid file like here?

marjohn56: Perhaps, but we need feedback to try and nail it down. Just covering all bases to start with.

src/etc/inc/plugins.inc.d/openvpn.inc
marjohn56: I'm not taking prisoners here, hence 'KILL'. :)

BPtLNfxZWo:
Hi, 
Thanks, I tried it, unfortunatly I got an error on the Webgui after installing the patch

`[13-Feb-2019 18:43:28 Europe/Vienna] PHP Parse error:  syntax error, unexpected ';' in /usr/local/etc/inc/plugins.inc.d/openvpn.inc on line 1010`
`
[13-Feb-2019 18:43:50 Europe/Vienna] PHP Parse error:  syntax error, unexpected ';' in /usr/local/etc/inc/plugins.inc.d/openvpn.inc on line 1010`

Crash report has been sent with link to this issue

br

marjohn56:
Sorry... my typo.. Updated try b2cc0ae

BPtLNfxZWo:
Same, no change. Reboot required?

Feb 13 20:20:14 | openvpn[44247]: Exiting due to fatal error
Feb 13 20:20:14 | openvpn[44247]: Cannot open TUN/TAP dev /dev/tun1: Device busy (errno=16)
Feb 13 20:20:14 | openvpn[44247]: TUN/TAP device ovpns1 exists previously, keep at program end

marjohn56:
Yes, reboot and then try it.

BPtLNfxZWo:
Hi,
patched and reboot, issue is partially solved.

I've two OpenVPN Instances running, one TCP one UDP. The first (TCP) is ok now, but the second suffers still the same issue.

`Feb 14 18:11:52 | openvpn[14952]: Exiting due to fatal error`
`Feb 14 18:11:52 | openvpn[14952]: Cannot open TUN/TAP dev /dev/tun3: Device busy (errno=16)`
`Feb 14 18:11:52 | openvpn[14952]: TUN/TAP device ovpns3 exists previously, keep at program end`

br

fichtner:
I've added https://github.com/opnsense/core/commit/f10b710a58 -- the SIGKILL is equal to the first SIGTERM and OpenVPN says SIGINT or SIGTERM should be used:

https://openvpn.net/community-resources/controlling-a-running-openvpn-process/

If the SIGKILL succeeds it means there is a stray process being created after SIGTERM, so someone else is already restarting OpenVPN. We can't fix it by hoping to kill the other caller, the whole call to openvpn_restart() needs to be serialised...

Issue #3089, opened 2019-01-02
Captive portal forwarding and HA/CARP setup


t4skforce:
I stumbled upon a problem when using Captive Portal in a combination with CARP / High Availability. 

The generated `ipfw` rules block access to the CARP IP used.

*Solution:*
Add Virtual IP of the CARP Interface as Allowed addresses
Services -> Captive Portal -> Administration -> Allowed addresses

Can this be added to the Documentation please ;)

AdSchellevis:
I'm not sure what you mean, the carp protocol should be passed, the rest should react the same as a non carp setup by my knowledge. Can you show your ipfw rules and point us to the issue?

t4skforce:
Issue being the default setup uses the local firewall IP not the VIP that is being used in CARP / HA. So the VIP IP is being blocked by the rules per default. The Setup only knows of the Interface not the VIP Address.

CARP VIP: 192.168.1.254
FW1 - VLAN30: 192.168.1.250
FW2 - VLAN30: 192.168.1.251

- ipfw script allows 192.168.1.250 on FW1 and 192.168.1.251 on FW2
- clearly CARP VIP is missing 192.168.1.254

That's why it needs to be added by manually as "Allowed addresses". Haven't found by quickly glancing at the code the exact place where it builds the rules based on the interface IP's.

AdSchellevis:
output of `ipfw show` please.

t4skforce:
Without the "Allowed addresses" 
```00100       0          0 allow pfsync from any to any
00110       0          0 allow carp from any to any
00120       0          0 allow ip from any to any layer2 mac-type ...
00130       0          0 allow ip from any to any layer2 mac-type ...
00140       0          0 allow ip from any to any layer2 mac-type ...
00150       0          0 deny ip from any to any layer2 not mac-type ...
00200       0          0 skipto 60000 ip6 from ::1 to any
00201       0          0 skipto 60000 ip4 from 127.0.0.0/8 to any
00202       0          0 skipto 60000 ip6 from any to ::1
00203       0          0 skipto 60000 ip4 from any to 127.0.0.0/8
01002       0          0 skipto 60000 udp from any to 10.13.1.250 dst-port 53 keep-state :default
01002       0          0 skipto 60000 ip from any to { 255.255.255.255 or 10.13.1.250 } in
01002       0          0 skipto 60000 ip from { 255.255.255.255 or 10.13.1.250 } to any out
[...]

PR #3233, opened 2019-02-14
Dev/ipfwcaptiveportal

fredronnv:
I think this relates to #3089. I noticed that captive portal wasn't working unless I explicitly added the CARP VIPs to the "Allowed addresses" list.
This PR makes all VIPs on captive portal interfaces allowed by default by changing the template for ipfw.conf. I needed an extra array in cp_interface_list to allow checking against the if later.

Only VIPs on Captive Portal interfaces are added, in the same manner as "Allow traffic to this hosts static ip's".

### DIFF
--- a/src/opnsense/service/templates/OPNsense/IPFW/ipfw.conf
+++ b/src/opnsense/service/templates/OPNsense/IPFW/ipfw.conf
@@ -101,13 +101,11 @@ add 203 skipto 60000 ipv4 from any to 127.0.0.0/8
 #======================================================================================
 # Allow traffic to this host
 #======================================================================================
-{% for item in cp_interface_list %}
-add {{loop.index  + 1000}} skipto 60000 udp from any to me dst-port 53 via {{item.if}} keep-state
-add {{loop.index  + 1000}} skipto 60000 ip from any to { 255.255.255.255 or me } in via {{item.if}}
-add {{loop.index  + 1000}} skipto 60000 ip from { 255.255.255.255 or me } to any out via {{item.if}}
-add {{loop.index  + 1000}} skipto 60000 icmp from { 255.255.255.255 or me } to any out via {{item.if}} icmptypes 0
-add {{loop.index  + 1000}} skipto 60000 icmp from any to { 255.255.255.255 or me } in via {{item.if}} icmptypes 8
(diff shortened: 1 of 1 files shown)


AdSchellevis:
@fredronnv thanks, I have to check this, at a first glance the only issue seem to be that the client can't reach the vip interface (for dns / icmp), right? In that case could probably also bind the rules to the physical interface, which reduces the number of rules.

fredronnv:
@AdSchellevis yes, you're right, that's the reason. No DNS resolutions on captive portal if you've configured the VIPs for that purpose. I'll get a commit sorted to bind the rules to the physical interface as well.

fredronnv:
@AdSchellevis actually, looking at it, I think really only DNS is needed to make it work, really, so could limit it to that only. I'll test it out.

AdSchellevis:
you could probably change the `ipaddr` for `if` in the original line

```
add {{loop.index  + 1000}} skipto 60000 udp from any to {{ interface.ipaddr }} dst-port 53 keep-state
```

AdSchellevis:
`to me ` seems to to the trick https://www.freebsd.org/cgi/man.cgi?ipfw(8) and add a `via`  at the end

fredronnv:
Ah I see what you mean, that's probably preferrable. I'll make the necessary changes.

AdSchellevis:
@fredronnv Thanks! very nice to have you around here

AdSchellevis:
we could probably move the full block of rules into your new construction, logically the same also counts for traffic out using the vip in these cases.

fredronnv:
If you'd like I can rework it, but I need a bit of time to make there are no unexpected consequences. I've got some time tomorrow morning (CET) I can dedicate to this.

AdSchellevis:
great! thanks in advance

fredronnv:
Come to think of it, it's unnecessary to loop over interfaces if we use "to me", isn't it?

AdSchellevis:
we should loop, since we only want the captive portal interface, don't want to risk opening udp 53 on wan :)

fredronnv:
Of course, done.

AdSchellevis:
@fredronnv thanks, I'll give it a spin here and pull it in later.

PR #3234, opened 2019-02-14
Sign a cert for CSR on WebUI, fixes #1850

### LINKED ISSUE #1850: Feature Request: Internal CA can sign CSRs in UI
Hi,

Please consider adding to the existing Certificate Authority UI functionality to sign an existing CSR. Presently the UI only permits generating a CSR from scratch or importing existing private key and public cert.

I use OPNsense as my in-house CA and not all devices accept uploads of pre-generated keys/certs.

Thanks!

nhirokinet:
Hello,

I installed OPNsense, and wanted to sign a cert for a CSR file for OpenVPN operation. After writing a code I found the issue, so I added the issue ID on a git comment.

Some part of WebUI had an code to distinguish whether `$cert['prv']` is available, and some part did not. I added some code not to display unavailable Certificate (as Private key), but it may not be enough.

Please check codes and comment what I should add to this Pull Request.

```
% sudo make upgrade CORE_SQUID=3
```
would work now


nhirokinet:
After pushing I found unnecessary comment,  so removed it.

src/etc/inc/certs.inc
```diff
@@ -269,6 +269,48 @@ function cert_create(&$cert, $caref, $keylen, $lifetime, $dn, $digest_alg = 'sha
...
+    $ca_str_key = base64_decode($ca['prv']);
+    $ca_res_crt = openssl_x509_read($ca_str_crt);
+    $ca_res_key = openssl_pkey_get_private(array(0 => $ca_str_key, 1 => ""));
+    if (!$ca_res_key) {
+        return false;
+    }
+    $ca_serial = ++$ca['serial'];
+
+    $args = array(
+        'private_key_type' => OPENSSL_KEYTYPE_RSA,
```
fabianfrz: This may cause trouble if the key type does not match. RSA is not the only option.

src/www/system_certmanager.php
```diff
@@ -89,6 +89,7 @@ function csr_get_modulus($str_crt, $decode = true)
     "import" => gettext("Import an existing Certificate"),
     "internal" => gettext("Create an internal Certificate"),
     "external" => gettext("Create a Certificate Signing Request"),
+    "sign_cert_for_csr" => gettext("Sign for a Certificate Signing Request"),
```
wschlich: Please remove 'for' (in both the array key name and the description) :-)

src/www/system_certmanager.php
```diff
@@ -330,10 +333,15 @@ function () {
         } elseif ($pconfig['certmethod'] == "existing") {
             $reqdfields = array("certref");
             $reqdfieldsn = array(gettext("Existing Certificate Choice"));
+        } elseif ($pconfig['certmethod'] == 'sign_cert_for_csr') {
```
wschlich: Same here (remove 'for').

src/etc/inc/certs.inc
```diff
@@ -269,6 +269,48 @@ function cert_create(&$cert, $caref, $keylen, $lifetime, $dn, $digest_alg = 'sha
     return true;
 }
 
+function sign_cert_for_csr(&$cert, $caref, $csr, $lifetime, $digest_alg, $x509_extensions = 'usr_cert')
```
wschlich: Please remove 'for_' in the function name.

src/www/system_certmanager.php
```diff
@@ -460,6 +471,15 @@ function () {
                     }
                 }
 
+                if ($pconfig['certmethod'] === 'sign_cert_for_csr') {
```
wschlich: Same here (remove 'for_').

src/www/system_certmanager.php
```diff
@@ -460,6 +471,15 @@ function () {
...
+                    if (!sign_cert_for_csr($cert, $pconfig['caref_sign_csr'], $pconfig['csr'], (int) $pconfig['lifetime_sign_csr'], $pconfig['digest_alg_sign_csr'])) {
```
wschlich: Same here (remove 'for_').

src/www/system_certmanager.php
```diff
@@ -682,6 +702,7 @@ function removeRowAltNm() {
             $("#internal").addClass("hidden");
             $("#external").addClass("hidden");
             $("#existing").addClass("hidden");
+            $("#sign_cert_for_csr").addClass("hidden");
```
wschlich: Same here (remove 'for_').

src/www/system_certmanager.php
```diff
@@ -690,6 +711,8 @@ function removeRowAltNm() {
             } else if ($(this).val() == "external") {
                 $("#external").removeClass("hidden");
                 $("#altNameTr").detach().appendTo("#external > tbody:first");
+            } else if ($(this).val() == "sign_cert_for_csr") {
```
wschlich: Same here (remove 'for_').

src/www/system_certmanager.php
```diff
@@ -832,6 +855,66 @@ function internalca_change() {
                 </tr>
               </tbody>
             </table>
+            <!-- sign_cert_for_csr -->
```
wschlich: Same here (remove 'for_').

src/www/system_certmanager.php
```diff
@@ -690,6 +711,8 @@ function removeRowAltNm() {
             } else if ($(this).val() == "external") {
                 $("#external").removeClass("hidden");
                 $("#altNameTr").detach().appendTo("#external > tbody:first");
+            } else if ($(this).val() == "sign_cert_for_csr") {
+                $("#sign_cert_for_csr").removeClass("hidden");
```
wschlich: Same here (remove 'for_').

src/www/system_certmanager.php
```diff
@@ -832,6 +855,66 @@ function internalca_change() {
                 </tr>
               </tbody>
             </table>
+            <!-- sign_cert_for_csr -->
+            <table id="sign_cert_for_csr" class="table table-striped opnsense_standard_table_form">
```
wschlich: Same here (remove 'for_').

src/www/system_certmanager.php
```diff
@@ -832,6 +855,66 @@ function internalca_change() {
...
+              <thead>
+                <tr>
+                  <th colspan="2"><?=gettext("Sign CSR");?></th>
```
wschlich: You got it right here (without 'for') :D

src/etc/inc/certs.inc
wschlich: Exactly. According to http://php.net/manual/en/function.openssl-csr-new.php, these are the other types:
OPENSSL_KEYTYPE_DSA
OPENSSL_KEYTYPE_DH
OPENSSL_KEYTYPE_EC

In real-world scenarios, I'd believe at least OPENSSL_KEYTYPE_EC would be relevant for supporting elliptic curve cryptography certficates.

src/www/system_certmanager.php
```diff
@@ -832,6 +855,66 @@ function internalca_change() {
...
+  <?php
+                    endforeach; ?>
+                    </select>
+                    <div class="hidden" id="no_caref_sign_csr">
+                      <?=sprintf(gettext("No internal Certificate Authorities have been defined. You must %sadd%s an internal CA before creating an internal certificate."),'<a href="system_camanager.php?act=new&amp;method=internal">','</a>');?>
+                    </div>
+                  </td>
+                </tr>
+                <tr>
+                  <td style="width:22%"><a id="help_for_csr_sign" href="#" class="showhelp"><i class="fa fa-info-circle"></i></a> <?=gettext("CSR file");?></td>
```
wschlich: The 'CSR file' shows an orange info icon indicatin there's help text available, but it doesn't appear when clicking on it.
Probably due to 'id="help_for_csr_sign"' vs. 'data-for="help_for_csr"' below (line 888)?

src/etc/inc/certs.inc
wschlich: Also,  $x509_extensions is limited to 'usr_cert' which is bad.

1. It drops any X509v3 subjectAltName attribute values supplied in the CSR while they should indeed be copied 1:1

2. It drops any X509v3 Key Usage attributes while they should indeed be copied 1:1

3. It also adds a meaningless comment: `nsComment = "OpenSSL Generated Certificate"`

Example X509v3 extensions from a CSR I just signed using the code in here:

- CSR
```
        Requested Extensions:
            X509v3 Basic Constraints: 
                CA:FALSE
            X509v3 Key Usage: 
                Digital Signature, Non Repudiation, Key Encipherment
            X509v3 Subject Alternative Name: 
                DNS:my.fqdn.tld
```
- Signed certificate:
```
        X509v3 extensions:
            X509v3 Basic Constraints: 
                CA:FALSE
            Netscape Comment: 
                OpenSSL Generated Certificate
            X509v3 Subject Key Identifier: 
                38:18:41:5C:9C:07:68:6F:DF:99:63:A2:3E:8E:42:3D:94:AB:3D:41
            X509v3 Authority Key Identifier: 
                keyid:38:01:5E:58:A1:1D:C2:4D:E6:62:03:8D:E1:EF:42:4F:57:BE:C2:94
```

src/etc/inc/certs.inc
```diff
@@ -269,6 +269,48 @@ function cert_create(&$cert, $caref, $keylen, $lifetime, $dn, $digest_alg = 'sha
...
+    $ca_str_crt = base64_decode($ca['crt']);
+    $ca_str_key = base64_decode($ca['prv']);
+    $ca_res_crt = openssl_x509_read($ca_str_crt);
+    $ca_res_key = openssl_pkey_get_private(array(0 => $ca_str_key, 1 => ""));
+    if (!$ca_res_key) {
+        return false;
+    }
+    $ca_serial = ++$ca['serial'];
+
+    $args = array(
```
wschlich: A certificate directly created from within OPNsense uses a temporary OpenSSL config based upon the template `/usr/local/etc/ssl/opnsense.cnf` to set a fixed and sane `keyUsage` and to handle `subjectAltName` tags, see `cert_create()` and the call to `create_temp_openssl_config()` therein.

Omitting the 'config' parameter for the `$args` array, as done here, makes OpenSSL use the config `/usr/local/openssl/openssl.cnf` instead. This and the missing handling of `subjectAltName` tags makes the generated certificate have the following issues:

1. Any X509v3 subjectAltName attribute values supplied in the CSR are dropped while they should indeed be copied 1:1

2. Any X509v3 Key Usage attributes supplied in the CSR are dropped while they should indeed be copied 1:1 *or* replaced by the fixed ones as specified in `opnsense.cnf`

Example X509v3 extensions from a CSR I just signed using the code in here:

- CSR
```
        Requested Extensions:
            X509v3 Basic Constraints: 
                CA:FALSE
            X509v3 Key Usage: 
                Digital Signature, Non Repudiation, Key Encipherment
            X509v3 Subject Alternative Name: 
                DNS:my.fqdn.tld
```
- Signed certificate:
```
        X509v3 extensions:
            X509v3 Basic Constraints: 
                CA:FALSE
            Netscape Comment: 
                OpenSSL Generated Certificate
            X509v3 Subject Key Identifier: 
                38:18:41:5C:9C:07:68:6F:DF:99:63:A2:3E:8E:42:3D:94:AB:3D:41
            X509v3 Authority Key Identifier: 
                keyid:38:01:5E:58:A1:1D:C2:4D:E6:62:03:8D:E1:EF:42:4F:57:BE:C2:94
```

wschlich:
@nhirokinet Thanks for writing code to support signing CSRs! Can you please have a look at my comment https://github.com/opnsense/core/pull/3234#pullrequestreview-204127379 ? Thank you!

src/etc/inc/certs.inc
```diff
@@ -40,6 +40,8 @@ $openssl_crl_status = array(
   OCSP_REVOKED_STATUS_CERTIFICATEHOLD       => "Certificate Hold"
 );
 
+$openssl_config_template_file_name = '/usr/local/etc/ssl/opnsense.cnf';
```
fichtner: please don't do this, this introduces side effects

nhirokinet:
Thank you for comments.

I changed codes regarding comments, except for:

- The (probably biggest) problem that SAN is not included in the signiture
- An additional problem that some unexpected error stops signing in some options.

For HTML id, system_certmanager.php had more than 1000 lines without namespace, so I chose longer name.

I will check the correct way to sign in `function sign_cert_csr` and push some change.

wschlich:
@nhirokinet Most important is to actually parse the CSR in PHP and extract the supplied `subjectAltName` values (same could be done for `keyUsage`):
https://gist.github.com/dol/e9f3d682529ae7ee368f0e6862e16a87#file-cert-php-L58

nhirokinet:
Thanks, I will try to make better as much as possible.

First of all I just implemented what just works, and after reading your comments and thinking for a while I understood at least there are more.

It seems at least I have to consider:

- What SSL extensions to support: SAN, and some other properties seems to be considered each, so defining the goal of this pull request seems much more difficult than I thought.
- What should be confirmed a second before clicking "Save" button: first I thought adding signing CSR just the user created, but it seems two below may differ (depends on implementation) much in a way what a "Save" allows the CA:
  - Just "Create an Internal Certificate"
  - "Create a Certificate Signing Request" and "Sign a Certificate Signing Request"

Anyway I will try to do my best. It's my pleasure if you wait for the next my action! I will try to make a next action in a few days.

src/www/system_certmanager.php
```diff
@@ -906,6 +950,7 @@ function internalca_change() {
                   <td style="width:22%"><a id="help_for_csr_sign_csr" href="#" class="showhelp"><i class="fa fa-info-circle"></i></a> <?=gettext("CSR file");?></td>
                   <td style="width:78%">
                     <textarea name="csr" id="csr" cols="65" rows="7"><?=$pconfig['csr'];?></textarea>
+                    <a href="#" class="csr_info_for_sign_csr">Display CSR Content (TODO: better style button)</a>
```
fabianfrz: Try to add classes "btn btn-primary"

nhirokinet: I added, and changed to btn-secondary to make difference from important button. Thanks.

src/etc/ssl/opnsense.cnf
```diff
@@ -292,3 +292,14 @@ extendedKeyUsage       = clientAuth,serverAuth,1.3.6.1.5.5.8.2.2
 keyUsage               = nonRepudiation, digitalSignature, keyEncipherment
 
 ###OPNsense:subjectAltName###
+
+[ sign_csr ]
+
+nsComment              = "OPNsense Generated Combined Client/Server Certificate"
```
wschlich: Seems wrong to me. Why combined client/server certificate?! This is controlled by `keyUsage` and `extendedKeyUsage` instead, I believe.

src/www/system_certmanager.php
```diff
@@ -84,6 +84,38 @@ function csr_get_modulus($str_crt, $decode = true)
...
+    // TODO: implement other extensions
+
+    $ret = array();
+    $ret['parse_success'] = true;
+    $ret['subject'] = openssl_csr_get_subject($csr);
+    if ($ret['subject'] === false) {
+        return array('parse_success' => false);
+    }
+
+    $process = proc_open('/usr/local/bin/openssl req -text -noout | grep "X509v3 Subject Alternative Name:" -A 1 | tail -n 1 | sed -e "s/^ *//g"', array(array("pipe", "r"), array("pipe", "w"), array("pipe", "w")), $pipes);
```
wschlich: This seems like a very ugly solution attempt to me.

I like the solution from https://gist.github.com/dol/e9f3d682529ae7ee368f0e6862e16a87#file-cert-php-L58 much better. 

It uses https://github.com/sop/x509 which in turn depends on a few other libraries. As OPNsense is using X.509 certificates in many places, I'd believe it could be a good idea to add these libraries, but that's more of a question to the core developers.

src/etc/ssl/opnsense.cnf
nhirokinet: Oh, sorry, I did not read well and thought all are the same. I will remove `Combined Client/Server`.

src/www/system_certmanager.php
nhirokinet: Yes, exactly. This is why I commented TODO lines that I will do beforre merging some lines above.

I'm currently surveying how to install some php modules from ports (it seems I also have to change opnsense/ports repository?) and install sop/x509, sop/crypto-tools on vendor directory.

After finding out some set of tasks to prepare environment, I will comment abount it.

nhirokinet: Precisely: not just very ugly but it does not work correctly (at least if CSR has IP address in SAN)

nhirokinet:
To use the sop/x509 library you suggested, I installed these packaged to my OPNsense testing environment by hand, and the minimal code (just in home directory) seems to work.

I installed:

- devel/php-composer
  - php71-intl as dependency
  - many other dependencies like autoconf-related ones, Pgettext, erl 5 internationalization related ones, libutil, gmake
- math/php71-gmp

php71-gmp is required explicitly in README, so adding `math/php71-gmp` is a definitely must.

In my environment composer on Ubuntu without php-intl complainted intl is required. After checking whether this is required by sop/x509 or just by composer, `devel/php71-intl` may be added.

So, I have to install `math/php71-gmp` (and maybe `devel/php71-intl`) additionaly.

Questions:

- I will add `SUBDIR` variables in [Makefile](https://github.com/opnsense/ports/blob/master/math/Makefile) which seems to determine actual installed ports, when this Pull Request is close to merge. Will it successfully install php71-gmp correctly, or do I have other things to do?
-  For now I put just metadata and `composer install` is going to be done each environment, with `composer.lock` which determines exactly one version combination. However, it seems currently building `core` repository does not need `composer`. Should I add something for that?

Note that git commits after that may require additional ports, mentioned in this message.

nhirokinet:
After this comments my testing environments include:

- `math/php71-gmp`
- `devel/php71-composer`
- things obtained by `composer install` in `contrib/x509`, manually put to `/usr/local/opnsense/contrib/x509`

nhirokinet:
This is still Work In Progress. I will do what I wrote in 'TODO' comments for a while within this pull request.

After reading suggestions I thought:

- x509 extensions should be explicitly visible to the user before clicking "Save: button. because:
  - openssl is designed so for security reason
  - we cannot convey "ALL" x509 extensions available according to openssl design, and anyway, user should be able to know what happens

Here I did not make Common Name modifiable to CA admin. This does not have deep reason, just be considered not to be heavy because it is at least visible before clicking "Save button".

UI for x509 extensions is very poor now, I will do something. Lack of validation is a security issue, I will do something. I will do something against any other "TODO" comments added in this pull requests.

If there is a problem with no "TODO" comment around there, it is at least the one I did not notice, so comment is needed.

Thank you for advices.

src/www/system_certmanager.php
AdSchellevis: @nhirokinet @wschlich I'm not exactly sure what we're trying to do here (haven't read all the code), but php has quite some x509 functions built in natively nowadays.  We're always a bit cautious in adding new dependencies to core. If you can paste a simple example of the issue your trying to solve, maybe I can help searching some simple solution.

nhirokinet: @AdSchellevis (At least in my understanding, )

Tasks:

- Sign a CSR
  - Some x509 extensions, which can be included in CSR and certificate, like SANs, are standard and are a must today.

Problem for the task:

- In OpenSSL design, callers must explicitly designate what x509 extensions, like SAN, to be signed. It is same in PHP's x509 functions built, which relies on OpenSSL.
  - [@wschlich commented for detail](https://github.com/opnsense/core/pull/3234#issuecomment-464081639)
- This means that the PHP code (or other method who creates openssl.cnf) must know x509 extensions values, written in CSR.
  - Difference from previous system_certmanager.php here is `written in CSR`. Current "Create an Internal Certificate" does not have to CSR because CSR is just created in same function.
- Despite there are a lot of PHP functions to parse x509 certificates (most starting with `openssl_x509_`), there seems to be a little for CSR (most starting with `openssl_csr_`), none of them seem to be able to parse x509 extension.
  - [PHP: OpenSSL Functions - Manual](http://php.net/manual/en/ref.openssl.php)

nhirokinet:
Thanks to your advices, I added x509 functionality. It may take a little more to implement validation and UI.

For use of additional library sop/x509, there may be a discuss here a little more.

src/www/system_certmanager.php
AdSchellevis: @nhirokinet ok, if there's no glue for this csr task in php, that's a bummer. sop/x509 at first glance seems to be a rather large and complex module, but if there are no better alternatives, we might need to consider it.

nhirokinet:
I wrote down a system to support x509 extensions, and removed WIP.

There are some TODOs in my codes, which should be removed. As I'm not good at English, I'm glad if you watching here give me an example for good English message. Same for keyword expression for extendedKeyUsage.

I think others work. Please comment if you find something that does not work.

nhirokinet:
Questions in a comment yesterday remains: I should do at least something for that

- I will add `SUBDIR` variables in [Makefile](https://github.com/opnsense/ports/blob/master/math/Makefile) which seems to determine actual installed ports, when this Pull Request is close to merge. Will it successfully install php71-gmp correctly, or do I have other things to do?
-  For now I put just metadata and `composer install` is going to be done each environment, with `composer.lock` which determines exactly one version combination. However, it seems currently building `core` repository does not need `composer`. Should I add something for that?
- sop/x509 and related libraries are licensed under MIT license. Is it okay? (I think it's legal as long as license file are properly put, and contrib/base32 library which is already included seems to be MIT )

AdSchellevis:
@nhirokinet We'll look at the dependencies when finished, don't worry :) The MIT license shouldn't be an issue. Thanks for working on this!

nhirokinet:
I did not check on FreeBSD as it is hard to remove installed port, but sop/x509 seems to work in my WSL Ubuntu environment after removing php-intl. So it seems that `math/php71-gmp` is the only additional opnsense/port for sop/x509.

AdSchellevis:
@nhirokinet I'm probably a bit late to the party, but it might be worth to look into https://github.com/phpseclib/phpseclib for the x509 features. The project looks quite active, simple and doesn't seem to come with a lot of dependencies. 

I've tried on my machine to dumps some attributes from a csr and it seems to function quite decently:

```
require __DIR__ . '/phpseclib/vendor/autoload.php';

use phpseclib\File\X509;
$x509 = new X509();
$cert = $x509->loadCSR($my_csr);
print_r($cert['certificationRequestInfo']['attributes']);

```

$my_csr contains my test csr

wschlich:
> @nhirokinet I'm probably a bit late to the party, but it might be worth to look into https://github.com/phpseclib/phpseclib for the x509 features. The project looks quite active, simple and doesn't seem to come with a lot of dependencies.

Also, the lead developer seems to be a PHP committer: 
```
Jim Wigginton <terrafrost@php.net>
```

fichtner:
It would be nice to have full CRL support via the to be selected library as well, but so far I don’t see anything appropriate. We still maintain an unofficial PHP patch. Not sure if this makes selection easier, but it would take pressure from us long term.

wschlich:
> It would be nice to have full CRL support via the to be selected library as well, but so far I don’t see anything appropriate. [...]

http://phpseclib.sourceforge.net/x509/compare.html#createcrl ?

nhirokinet:
Hmm, I tried out phpseclib.
https://github.com/nhirokinet/core/pull/1/files

For my impression,

## Pros for sop/x509

- Strict and well defined interface, no confusion

## Pros for phpseclib

- Short codes needed for OPNsense
- No additional PHP module needed (sop/x509 required gmp)
- more documents when googling (74,800 including Japanese ones ones v.s.27,400 with no Japanese ones in top 10 in Japanese setting)
- Maintainance for old versions are clearly written in README

nhirokinet:
Both seems to have a clear API to designate x509 extensions attributes without creating a file with unclear escaping method.

- https://github.com/sop/x509/blob/035734008e2e30e58f4405310863f888142b72e3/lib/X509/CertificationRequest/CertificationRequestInfo.php#L165-L175
- http://phpseclib.sourceforge.net/x509/compare.html#csrattr

nhirokinet:
Oh, it seems that FuelPHP depends on phpseclib.

- https://fuelphp.com/docs/vendor/phpseclib.html
- https://github.com/fuel/core/blob/0a7931b9acd9b6cbb7abca393c5044b04668f596/composer.json#L15

nhirokinet:
It seems it is better to phpseclib instread of sop/x509, so I switched to phpseclib. Addition to opnsense/port I mentioned in comments is not required anymore.

nhirokinet:
I felt it a little bit hard to manually install PHP files, and tried to make `sudo make upgrade CORE_SQUID=3` work.

In this process I added contrib/phpseclib/vendor, which includes everything of phpseclib, included in git repo. I'm not sure whether this is good, but at least it seems the same as others in `contrib`. I did so because I tried php-composer on my testing OPNsense instance and found it difficult (both pkg and port failed).

Anyway the latest comment can be tried with `sudo make upgrade CORE_SQUID=3`. CORE_SQUID seems to be removed from Makefile recently, so I thought this would be solved after some upgrade.

nhirokinet:
Wow this change made this pull req `+36,389 lines`...

I guess it is better to make contrib/composer.json, add google-api-php-client, simplepie, phpseclib to one conposer.json, and prepare a method for automatically `composer install` just before packaging, but at least this would be by different PR.

fichtner:
We can add a freebsd port for phpseclib. I don’t see value in pulling composer into build and dev duties and making the code harder to maintain and update (and revert).

Squid on master will be available as a package when 19.1.2 is released. For now CORE_SQUID is the correct workaround.


Cheers,
Franco

nhirokinet:
> I don’t see value in pulling composer into build and dev duties and making the code harder to maintain and update (and revert).

I see, unless negative votes come from core deveploers, I'd choose a way to put all codes `contrib/phpseclib/vendor`.

Still there is a problem that one of committers must press merge button to this 36k lines pull req, with confirming that I did not something not acceptable for merging, or even confirming that I did not inject malware to somewhere in this 36k lines.

For this I'm just a passing user, so I would wait for the core deveplopers' comments for how to solve this.

AdSchellevis:
@nhirokinet @fichtner is a core developer :) to limit the size of the commit it would probably be a better idea to strip the dependencies from the pull request and add some steps to how to test the code somewhere in here. Maybe @wschlich could do some testing then, when functionally acceptable, I'll see how/when we can pull this in.

nhirokinet:
@AdSchellevis 

> fichtner is a core developer :)

Well, sorry, I thought so looking at logs, just thought it would be better to wait for at least another core developers, as there may be discussion.

Anyway is it better to make another pull req just adding phpseclib, keeping this pull req small with depending phpseclib pull req?

AdSchellevis:
We don't need a pull request for phpseclib,  but if it helps @wschlich to test your code, you could leave it in there for now and remove it later. A pull this large is a bit difficult to handle.

nhirokinet:
@AdSchellevis Okay, thanks a lot! It would be much easier for me if some of you would add phpseclib, of course. Then I will wait until next comments.

wschlich:
I'd love to continue testing as soon as things are ready (opnsense-patch xxx?) :-)

AdSchellevis:
@nhirokinet maybe you can squash your commits, that would enable @wschlich to use `opnsense-patch` to pull in your changes easily.

nhirokinet:
I squashed and tried `sudo opnsense 8d5c99e` in clean installed environment, but it did not work: phpseclib was put in /usr/local/phpseclib/.
```
sudo opnsense 8d5c99e
sudo cp -R /usr/local/phpseclib /usr/local/opnsense/contrib/
```
worked, so it seems something additional is required to place these files in /usr/local/opnsense/contrib, which seems to be available in `include`.

Where should I add something for that?

AdSchellevis:
@nhirokinet the contrib parts are installed from https://github.com/opnsense/core/tree/master/contrib, but I figure @wschlich can try your code with the two commands provided. Thanks for your contribution, let's wait for some user feedback and take it from there.

wschlich:
I applied the patch, here's two suspicious messages:
```
|diff --git a/contrib/Makefile b/contrib/Makefile
|index 86bfdb305e..652a2b7e92 100644
|--- a/contrib/Makefile
|+++ b/contrib/Makefile
--------------------------
No file to patch.  Skipping...
Hunk #1 ignored at 1.
1 out of 1 hunks ignored--saving rejects to Makefile.rej

--------------------------
|diff --git a/plist b/plist
|index a08cdc4acf..b1d6a4a28b 100644
|--- a/plist
|+++ b/plist
--------------------------
No file to patch.  Skipping...
Hunk #1 ignored at 256.
1 out of 1 hunks ignored--saving rejects to plist.rej
```
@nhirokinet can you elaborate on whether this needs to be taken special care of?

fichtner:
@wschlich not an issue

nhirokinet:
> What purpose does `[x] basicConstraints enabled` serve?!

Like other x509 extensions, I wrote this to disable extension completely.

... thinking on it with writing this comments, I reminded that CA option is the only thing basicConstraints do, so `isCA` will do the job.

nhirokinet:
For hints, it seems multiple lines hints are used some other places, it seems ok.
How do you think about use of ul, li tags in hint area? It may be better to see, but may be a little bit out of place there.

nhirokinet:
When ul and li are used, example is like this.
default

nhirokinet:
I followed review comments currently available. For hints I just preserved a place to write so that I can write it using my local editor.

nhirokinet:
I added some explanations, fixed typos and modified basicConstraints behavior. Also I removed and wrote something which seems better. Comments are welcome.

After previous patch commands, I just edited system_certmanager.php, so after applying previous patch commands just copying system_certmanager.php to /usr/local/www works.

I have wo things I did not decide, both about extendedKeyUsage.

- @wschlich suggested a help text that explains all the listed extendedKeyUsage values, which seems to be a litle bit too long list for hint area, depending on database, should I write then? (Instead, in my current code I wrote `
a help text that explains the typical keyUsage values for common certificate types (server cert, client cert, ca cert etc.` @wschlich mentioned for both keyUsage and extendedKeyUsage.)
- Do you have some suggestions about some available extendedKeyUsage OID database? phpseclib seems to have much less number (6) of supported OIDs compared with sop/x509 (29). Both does not include iKEIntermediate (1.3.6.1.5.5.8.2.2), which OPNsense 'Server Certificate' includes. I googled for that, but I could not find available copiable database dedicated.

AdSchellevis:
@wschlich can you ping me if the feature works as expected, there's probably some work needed for integration and cleanup, but I would like to prevent doing that twice.

src/www/system_certmanager.php
```diff
@@ -1288,7 +1267,7 @@ function internalca_change() {
                     <textarea name="csr" id="csr" cols="65" rows="7"><?=$pconfig['csr'];?></textarea><br/>
                     <a href="#" class="csr_info_for_sign_csr btn btn-secondary"><?=gettext("Show Detail");?></a><br/>
                     <div class="hidden" data-for="help_for_csr_sign_csr">
-                      <?=gettext("Paste the CSR file here. (TODO: good English explanations. x509 extensions that OPNsense knows is automatically copied when clicking Next button and modifiable, and others are ignored)");?>
+                      <?=gettext("Paste the CSR file here.");?>
```
wschlich: It would be awesome if one could also upload a CSR file instead of pasting it :)

AdSchellevis: probably, but I would focus on getting it functional and ready for release first.

wschlich:
> Both does not include iKEIntermediate (1.3.6.1.5.5.8.2.2), which OPNsense 'Server Certificate' includes (for now I added manually). I googled for that, but I could not find available copiable database dedicated.

Supporting ``iKEIntermediate`` seems like a good idea, according to http://javadoc.iaik.tugraz.at/iaik_jce/current/iaik/x509/extensions/ExtendedKeyUsage.html :
> The following purposes have been included in a predecessor draft of RFC 3280 and therefore continue to be registrated by this implementation:
> ipsecEndSystem (1.3.6.1.5.5.7.3.5) -- IP security end system
> ipsecTunnel (1.3.6.1.5.5.7.3.6) -- IP security tunnel termination
> ipsecUser (1.3.6.1.5.5.7.3.7) -- IP security user
> Note that these three key usages are deprecated by the PKIX profile for IKE. Instead the iKEIntermediate extended key usage (1.3.6.1.5.5.8.2.2) may be used in a certificate for an IPsec end entity. However, iKEIntermediate was also only specified by an internet draft (draft-ietf-ipsec-pki-req-05.txt) which has not become an RFC. Nevertheless iKEIntermediate may be required by an IpSec service.

wschlich:
> @wschlich can you ping me if the feature works as expected, there's probably some work needed for integration and cleanup, but I would like to prevent doing that twice.

Since I applied the latest OPNsense update last evening, I had to reapply the last patch and then the subsequent newer ones:
```
opnsense-patch 8d5c99e
pkg install rsync
rsync -avc /usr/local/phpseclib/ /usr/local/opnsense/contrib/phpseclib/
opnsense-patch 477c0ed
opnsense-patch ad65593
opnsense-patch 9cac4c5
opnsense-patch 51f0d98
```
Will test now.

src/www/system_certmanager.php
```diff
@@ -1342,18 +1342,16 @@ function internalca_change() {
...
+                          basic_constraints_enabled = $('#basic_constraints_is_ca_sign_csr').prop('checked');
                           $('#basic_constraints_path_len_sign_csr').prop('disabled', !basic_constraints_enabled)
                       }
                       </script>
-                      <input type="checkbox" name="basic_constraints_enabled_sign_csr"  id="basic_constraints_enabled_sign_csr"  value="true" onchange="basic_constraints_enabled_sign_csr_refresh();" /> <?= gettext('basicConstraints enabled'); ?><br />
-                      <input type="checkbox" name="basic_constraints_is_ca_sign_csr"    id="basic_constraints_is_ca_sign_csr"    value="true" /> <?= gettext('is CA'); ?><br />
+                      <input type="checkbox" name="basic_constraints_is_ca_sign_csr"    id="basic_constraints_is_ca_sign_csr" onchange="basic_constraints_enabled_sign_csr_refresh();" value="true" /> <?= gettext('is CA'); ?><br />
                       <?= gettext('Path Len'); ?>: <input type="text"     name="basic_constraints_path_len_sign_csr" id="basic_constraints_path_len_sign_csr" size="5" value="<?=$pconfig['basic_constraints_sign_csr'];?>"/>
```
wschlich: This should say ```Path Length``` and explain what it is about, see https://blogs.technet.microsoft.com/pki/2014/03/05/constraints-what-they-are-and-how-theyre-used/ and http://www.pkiglobe.org/index.html (I like their wording: ```Path length gives the maximum number of non-self-issued intermediate certificates that may follow this certificate in a valid certification path.```).

wschlich:
``keyUsage`` has ``digitalSignature, nonRepudiation, keyEncipherment`` preselected, effectively representing the ``Client Certificate`` usage as described in the help text.
``extendedKeyUsage``, however, has **no** preselection.

Both defaults should match equal usage examples IMHO.

Either keep both clear, or preselect ``clientAuth`` for ``extendedKeyUsage``.

wschlich:
``keyUsage`` and ``extendedKeyUsage`` have several possible values. Not all of them are explained in the extended help, but they should be, IMHO. Not a showstopper for a merge, but rather an improvement TODO for later on.

wschlich:
> > @wschlich can you ping me if the feature works as expected, there's probably some work needed for integration and cleanup, but I would like to prevent doing that twice.
> 
> Since I applied the latest OPNsense update last evening, I had to reapply the last patch and then the subsequent newer ones:
> 
> ```
> opnsense-patch 8d5c99e
> pkg install rsync
> rsync -avc /usr/local/phpseclib/ /usr/local/opnsense/contrib/phpseclib/
> opnsense-patch 477c0ed
> opnsense-patch ad65593
> opnsense-patch 9cac4c5
> opnsense-patch 51f0d98
> ```
> Will test now.

Besides my previous comments, signing a CSR for a server certificate worked just fine.

nhirokinet:
@wschlich 

> `keyUsage` has `digitalSignature, nonRepudiation, keyEncipherment` preselected, effectively representing the `Client Certificate` usage as described in the help text.
> `extendedKeyUsage`, however, has **no** preselection.

This is not default selections. Both are clear by default, but if CSR includes values, they would add as pre-selected selection.

If extensions in your CSR is missing from the list (which is added in this pull request), then they disappear.

These are examples to try behaviors.

```
-----BEGIN CERTIFICATE REQUEST-----
MIICADCCAaoCAQAwVjELMAkGA1UEBhMCSlAxDjAMBgNVBAgMBVRva3lvMSEwHwYD
VQQKDBhJbnRlcm5ldCBXaWRnaXRzIFB0eSBMdGQxFDASBgNVBAMMC2V4YW1wbGUu
Y29tMFwwDQYJKoZIhvcNAQEBBQADSwAwSAJBAPD7vlnLTcBMe9lylxdr6WP7SA8T
sJGX8C2sLbTOp5zVNR2EYJFn/cShcO2wRJuN9REdMvPIDC4jTBXhcOCX+2kCAwEA
AaCB7jCB6wYJKoZIhvcNAQkOMYHdMIHaMCwGCWCGSAGG+EIBDQQfFh1PcGVuU1NM
IEdlbmVyYXRlZCBDZXJ0aWZpY2F0ZTAdBgNVHQ4EFgQUCqKyodN1Ae6KWQbEB1w2
E/1YVUUwRAYDVR0RBD0wO4ILZXhhcG1sZS5jb22HBMAAAgGBEGhvZ2VAZXhhbXBs
ZS5jb22GFGh0dHBzOi8vZXhhbXBsZS5jb20vMAsGA1UdDwQEAwIF4DAnBgNVHSUE
IDAeBggrBgEFBQcDAgYIKwYBBQUHAwEGCCsGAQUFCAICMA8GA1UdEwQIMAYBAf8C
AQMwDQYJKoZIhvcNAQELBQADQQB7LjlRZBhHGTE9IE8vNVWdgfvaGJkxjJU11uW5
uDkk4kiDkXoIm1FlXO4aEg8PGbgoGYQ6kflIBalJWGGQbLQm
-----END CERTIFICATE REQUEST-----
```

```
-----BEGIN CERTIFICATE REQUEST-----
MIIBejCCASQCAQAwVjELMAkGA1UEBhMCSlAxDjAMBgNVBAgMBVRva3lvMSEwHwYD
VQQKDBhJbnRlcm5ldCBXaWRnaXRzIFB0eSBMdGQxFDASBgNVBAMMC2V4YW1wbGUu
Y29tMFwwDQYJKoZIhvcNAQEBBQADSwAwSAJBAPD7vlnLTcBMe9lylxdr6WP7SA8T
sJGX8C2sLbTOp5zVNR2EYJFn/cShcO2wRJuN9REdMvPIDC4jTBXhcOCX+2kCAwEA
AaBpMGcGCSqGSIb3DQEJDjFaMFgwLAYJYIZIAYb4QgENBB8WHU9wZW5TU0wgR2Vu
ZXJhdGVkIENlcnRpZmljYXRlMB0GA1UdDgQWBBQKorKh03UB7opZBsQHXDYT/VhV
RTAJBgNVHRMEAjAAMA0GCSqGSIb3DQEBCwUAA0EARaSfUnqt0MsNLcJd0BGrVC9K
G7Ac7xPxPACVDEigu4V+YqAvP2EFkpmTlmEcE8gli5pluktrm1uJRLw202MyGg==
-----END CERTIFICATE REQUEST-----
```

src/www/system_certmanager.php
nhirokinet: This wording seems to have come from RFC5280 section 4.2.1.9, so it seems to be good to use it.

https://tools.ietf.org/html/rfc5280#section-4.2.1.9

nhirokinet:
List, which first I copied from phpseclib, was exactly the same as default definitions in RFC5280 section 4.2.1.12, which seems to be a reasonable choice. These does not have extendedKeyUsages defined outsize of RFC5280, but I think this is a good point of start.

iKEIntermediate, which is already used in some other pages of OPNsense, is the only added external extendedKeyUsage.

wschlich:
> @wschlich
> 
> > `keyUsage` has `digitalSignature, nonRepudiation, keyEncipherment` preselected, effectively representing the `Client Certificate` usage as described in the help text.
> > `extendedKeyUsage`, however, has **no** preselection.
> 
> This is not default selections. Both are clear by default, but if CSR includes values, they would add as pre-selected selection.
> 
> If extensions in your CSR is missing from the list (which is added in this pull request), then they disappear.

Aaah, that makes sense :-) Thank you for clearing that up!

wschlich:
I vote for merging this pull-request in the current state.

AdSchellevis:
I'll take a look

AdSchellevis:
@nhirokinet I've manually picked your changes in https://github.com/opnsense/core/commit/2f919443dc6a60352ed24a4a6caffb53c9e4c443, refactored some in https://github.com/opnsense/core/commit/e65669383df0d40e914cb351ee1f382e76376157 

Until @fichtner adds our needed package, https://github.com/opnsense/core/commit/f685abcc9c91805e9f5dad35f8f30bc4bb6c585f disables the feature for now. 

I've tried your examples at my machine, this seems to be working very well, thanks for the work @nhirokinet and the testing @wschlich

AdSchellevis:
It would be good to have some documentation about the certificate management tools in OPNsense, @wschlich if you have the time and spirit, maybe you want to consider contributing to our docs (https://github.com/opnsense/docs) for this subject.

nhirokinet:
@AdSchellevis Thanks a lot for reviewing and adding to master branch, I'll be waiting this feature to be officially released!

fichtner:
https://github.com/opnsense/ports/commit/42f5990f and https://github.com/opnsense/tools/commit/0c7832b9 will allow us to enable the code in 19.1.3 development release ... I'll make the core change after I have a new packages snapshot ready.

fichtner:
not sure what autoload did, used https://github.com/opnsense/core/commit/27d606474f for now

Package for phpseclib:

    # pkg add -f https://pkg.opnsense.org/FreeBSD:11:amd64/snapshots/latest/All/php71-phpseclib-2.0.14.txz

AdSchellevis:
@fichtner minor issue, fixed in https://github.com/opnsense/core/commit/569abf9ac6618dfb312a0ddd5ddf634246acb3ac

fichtner:
@AdSchellevis how can I communicate clearly that autoload.php *does not exist* in the phpseclib package? :D

```
[05-Mar-2019 19:03:03 Europe/Berlin] PHP Warning:  require_once(phpseclib/vendor/autoload.php): failed to open stream: No such file or directory in /usr/local/www/system_certmanager.php on line 32
[05-Mar-2019 19:03:03 Europe/Berlin] PHP Stack trace:
[05-Mar-2019 19:03:03 Europe/Berlin] PHP   1. {main}() /usr/local/www/system_certmanager.php:0
[05-Mar-2019 19:03:03 Europe/Berlin] PHP Fatal error:  require_once(): Failed opening required 'phpseclib/vendor/autoload.php' (include_path='.:/usr/local/etc/inc:/usr/local/www:/usr/local/opnsense/mvc:/usr/local/opnsense/contrib:/usr/local/share/pear:/usr/local/share') in /usr/local/www/system_certmanager.php on line 32
[05-Mar-2019 19:03:03 Europe/Berlin] PHP Stack trace:
[05-Mar-2019 19:03:03 Europe/Berlin] PHP   1. {main}() /usr/local/www/system_certmanager.php:0
```

AdSchellevis:
@fichtner, hmm, something went wrong here. it was stuck on my machine it seems, let me try fixing this, the autoloader was there for a reason...

AdSchellevis:
@fichtner https://github.com/opnsense/core/commit/a98f04372f995bbd3d9f63f28c08868063f18256 it's a hack, but it's not how it's supposed to work..... Composer builds an autoloader for you, we probably should either add one in the package or create something generic (but different paths could complicate life here ;( )

PR #3243, opened 2019-02-18
remove "show certificate info" on certificate list from pending CSR

nhirokinet:
In certificates list, "show certificate info" button for pending CSR file shows empty dialog, because certificate file is not available yet.

Therefore, it is better to remove this button if no certificate is available.


AdSchellevis:
@nhirokinet thanks, good addition

PR #3248, opened 2019-02-24
Some improvements for firewall live filter

opnsenseuser:
I have modified the firewall live filter for gui use a bit. it is now possible to set filter by checkbox. the filter input field is retained. I think this is a helpful modification. It is also clearer. I hope you see it that way too. as far as I could tell, it works fine. What do you think?. thanks, best regards rené

### DIFF
--- a/src/opnsense/mvc/app/views/OPNsense/Diagnostics/fw_log.volt
+++ b/src/opnsense/mvc/app/views/OPNsense/Diagnostics/fw_log.volt
@@ -380,9 +380,9 @@ POSSIBILITY OF SUCH DAMAGE.
                 <fieldset>
                 <input type="radio" id="fw_tcpudp" value="1" name="tcpupd" checked="checked"> tcp/udp</input>
                 </br>
-                <input type="radio" id="fw_tcp" value="2" name="tcp"> tcp</input>
+                <input type="radio" id="fw_tcp" value="2" name="tcpupd"> tcp</input>
                 </br>
-                <input type="radio" id="fw_udp" value="3" name="upd"> udp</input>
+                <input type="radio" id="fw_udp" value="3" name="tcpupd"> udp</input>
                 </br>
                 </fieldset>
                 </div>
--- a/src/opnsense/mvc/app/views/OPNsense/Diagnostics/fw_log.volt
+++ b/src/opnsense/mvc/app/views/OPNsense/Diagnostics/fw_log.volt
@@ -238,16 +238,61 @@ POSSIBILITY OF SUCH DAMAGE.
                 }
             });
         }
-
+        
         // live filter
-        $("#filter").keyup(function(){
-            var search_str = $(this).val().toLowerCase();
+        $('#filter').keyup(function() {
+            var usr_str_block = usr_str_nat = usr_str_pass = usr_str_rdr = usr_str_tcp = usr_str_udp = usr_str_tcpudp = search_str = 'undef';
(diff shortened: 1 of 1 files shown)


opnsenseuser:
grafik

AdSchellevis:
Hi René, 

I'm not to excited yet, the layout mangles at my end, which is solvable, but the code complexity and amount seems to be increasing a bit too much for my taste.

image

If we want to generate filters, we should try to minimise additional code in the frontend.

Maybe it's an idea to limit the pull request to something simple that builds the query first (as a regexp).

`visible_text.match(search_str) ` should likely remain as it is, given the correct regex for your request.

Best regards,

Ad

opnsenseuser:
@AdSchellevis 
Thanks for your fast feedback!
I wanted to leave the code as far as possible unchanged. it was just an idea.
Should I delete the request again?

regards rené

AdSchellevis:
@opnsenseuser I don't mind if you want to work on it some more, although it might take a while before we can pull it in (as I really want to simplify this before hitting our master branch). Just to manage expectations, I'm not agains the idea of simplifying this for the enduser, just cautious with the code underneath.

mimugmail:
What would happen to icmp logs?

opnsenseuser:
I can safely filter separately. but before this I would have to adapt the whole code new so @AdSchellevis said and that seems to me personally currently too much effort.

opnsenseuser:
@mimugmail icmp now included. 
@AdSchellevis layout improvements!
the original filter field still works!

it is now much clearer, because now you can filter roughly and use the original filter field for wan or ip exclusively. it works really well.

opnsenseuser:
@fichtner what do you think?

opnsenseuser:
grafik

mimugmail:
You need a input field above every column to make it scale since there is also esp, ospf, carp and other protocol in use. A radio button for every protocol doesnt scale. Check the Logs page from Fabians Nginx plugin

opnsenseuser:
yes, an input box is certainly better and certainly it is best to overwork the whole live log, but to date no one has done anything to improve the live log and the current results are already much better than the old one.

fbrendel:
Maybe the [filter-control](https://examples.bootstrap-table.com/index.html#extensions/filter-control.html) extension of [bootstrap-table](https://bootstrap-table.com/) is an option?

opnsenseuser:
@all OK. I can not let go of the whole thing. I want to make the live filter better. Real better :-) 
Now I've made myself on the search of the database of opnsense. but somehow the whole thing is too confusing. A line in fw_log.volt (ajaxGet ('/ api / diagnostics / firewall / log /') makes me think, because I can not find this path or how or from where jquery reads the data from the firewall log or filter.log? but why this ajaxGet command? Can someone explain to me where and how the data is stored? I would like to develop the whole thing. Best regards rené

fbrendel:
The GET request _/api/diagnostics/firewall/log_ calls the function _logAction()_ from the firewall API controller  _mvc/app/controllers/OPNsense/Diagnostics/Api/FirewallController.php_.

That function asks the backend _configd_ for the logs
```php
$response = $backend->configdpRun("filter read log", array($limit, $digest));
```
Which runs the command defined in the _read.log_ section of _service/conf/actions.d/actions_filter.conf_.
```
[read.log]
command:/usr/local/opnsense/scripts/filter/read_log.py
parameters: /limit %s /digest %s
type:script_output
message:request filter log output
```
which in turn calls _read_log.py_ with the parameter _$limit_ and _$digest_.
read_log.py reads the _filterlog_ via _clog_ and returns the result back to the GET response.

opnsenseuser:
@fbrendel great thanks for the detailed instructions. 👍 
I have to get used to this.

PR #3267, opened 2019-03-01
monit: add migration/validation for service/test type dependency

fbrendel:
This PR implements the service/test type dependency.
Conditions are mapped to test types and these test types in turn are mapped to service types.

It prevents the user from adding wrong tests to a service.
That means that you still can select a wrong test for a service but get a validation error.
Furthermore, a test condition can only be changed if the test is not linked to a service or the condition results in 'Custom' type.
Every condition that cannot mapped to a type is a 'Custom' type.

The migration parses the condition field and sets the test type accordingly.
To bypass validation errors I use the post() method instead of run().

### DIFF
--- a/src/opnsense/mvc/app/models/OPNsense/Monit/Monit.php
+++ b/src/opnsense/mvc/app/models/OPNsense/Monit/Monit.php
@@ -1,7 +1,7 @@
 <?php
 
 /*
- *  Copyright (C) 2016 EURO-LOG AG
+ *  Copyright (C) 2016-2019 EURO-LOG AG
  *  Copyright (c) 2019 Deciso B.V.
  *  All rights reserved.
  *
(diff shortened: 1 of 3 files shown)


src/opnsense/mvc/app/models/OPNsense/Monit/Monit.php
```diff
@@ -77,26 +167,44 @@ public function performValidation($validateFullModel = false)
                         // test node validations
                         switch ($node->getInternalXMLTagName()) {
                             case 'type':
+                                // only 'Custom' is allowed if test is linked to a service
                                 $testUuid = $parentNode->getAttribute('uuid');
-                                if ($node->isFieldChanged() && $this->isTestServiceRelated($testUuid)) {
+                                if (strcmp((string)$node, 'Custom') != 0 &&
```
AdSchellevis: @fbrendel I'm a bit struggling why this is checked, if it's custom I can move it to something else, when faulty I can't move back. We probably should keep it locked for all, since migration fixes the default types.

AdSchellevis:
@fbrendel made some changes in https://github.com/opnsense/core/commit/eb280f61fccf40e1a45a05de5585a1db890cbdb8, to keep the progress.  if we should re-add the `if` condition for valid reasons, please add it in a new PR.

src/opnsense/mvc/app/models/OPNsense/Monit/Monit.php
fbrendel: The Custom type is meant as a fallback because I'm not sure if getTestType can successfully parse all (valid) variations of the condition string.
Maybe we should remove Custom and print an error message instead if getTestType fails?

AdSchellevis: @fbrendel I've removed it :) if it looks safe enough, we could leave it like this and see what happens if anyone complains

PR #3269, opened 2019-03-01
Proxy: add auth to parent proxy

### LINKED ISSUE #3264: Auth for Parent Proxy
We use opnsense in version OPNsense 19.1.1-amd64

We use the Parent Proxy feature to configure a parent proxy. Our parent needs authentification
 
`
cache_peer <host_parent> <port> <port2> <some parameters> login=<user:pass>
`
It would be nice if these parameters and the credentials would be configurable in the browser.
Now i have to change this line in parentproxy.conf or in the template.
These changes will be overwritten by updates.

mimugmail:
Closes #3264


AdSchellevis:
@mimugmail thanks!

PR #3273, opened 2019-03-01
IDS: High CPU usage when IPS enabled - netmap number of threads option

juliocbc:
Greetings!

Following this forum [thread](https://forum.opnsense.org/index.php?topic=6590.0)

I was wondering a simple way to avoid suricata use all CPU available when IPS enabled.

We've applied this solution in our labs here and the results were quite good, with 50% of CPU usage with heavy load (with the proper number of threads set.)

### DIFF
--- a/src/opnsense/mvc/app/controllers/OPNsense/IDS/forms/generalSettings.xml
+++ b/src/opnsense/mvc/app/controllers/OPNsense/IDS/forms/generalSettings.xml
@@ -12,6 +12,15 @@
         <help><![CDATA[Enable protection mode (block traffic).<br />Before enabling, please disable all hardware offloading first <a href="/system_advanced_network.php">in advanced network</a>.]]></help>
     </field>
     <field>
+        <id>ids.general.NetmapThreads</id>
+        <label>Number of IPS threads</label>
+        <type>text</type>
+        <advanced>true</advanced>
+        <help><![CDATA[With this option, you can set the max number of threads will be started with
+        IPS enabled. If not set, will be set to automatic. Be careful to not
+        use a higher number than the available CPU cores. <a href="/index.php">Check here</a>.]]></help>
+    </field>
+    <field>
         <id>ids.general.promisc</id>
         <label>Promiscuous mode</label>
         <type>checkbox</type>
--- a/src/opnsense/mvc/app/models/OPNsense/IDS/IDS.xml
+++ b/src/opnsense/mvc/app/models/OPNsense/IDS/IDS.xml
@@ -1,6 +1,6 @@
 <model>
     <mount>//OPNsense/IDS</mount>
-    <version>1.0.2</version>
+    <version>1.0.1</version>
     <description>
         OPNsense IDS
     </description>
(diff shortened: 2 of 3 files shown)


AdSchellevis:
hi @juliocbc could you rebase your branch? I see some of our recent changes being reverted in your commit. Just out of curiosity, what does your test setup look like?

juliocbc:
Hi @AdSchellevis I'll do it! I need to redo the PR?

Following the hardware specs from our lab:
OPNsense 18.7.10_4
hw.model: Intel(R) Atom(TM) CPU  C2758  @ 2.40GHz
hw.machine: amd64
hw.ncpu: 8
16GB RAM
Intel i210AT

AdSchellevis:
@juliocbc yes please, I was more interested in how you measured and what variations you've seen with different packet sizes. I kind of expect some other issues when changing the number of threads, since FreeBSD doesn't support RSS at the moment (which is used to spread the flows)

juliocbc:
@AdSchellevis I did some quick tests with iperf, unfortunately, I was limited to a 1Gbps switch, so I can't stress it above this speed. 

I didn't changed the size of datagrams ou test with TCP. 

The iperf tests were with the following args:
`iperf -p 5201 -c 192.168.1.99 -u -b 10m -P 100 -d -t 60`

Full duplex with 1, 10 and 100 threads, IDS mode (without netmap) and IPS (with netmap). With netmap (IPS), all tests results in 100% CPU consumed.

AdSchellevis:
but what was the speed difference in IPS mode using different settings for the threads?

juliocbc:
@AdSchellevis 

Making another tests, we've found an issue:

With netmap threads set to 2:

```
Mar  7 12:04:47 utm01 suricata: [101011] <Error> -- [ERRCODE: SC_ERR_INVALID_VALUE(130)] - Thread count can't be greater than Rx ring count. Configured 2 threads for interface 'igb1_vlan1001' with 1 Rx rings.
Mar  7 12:04:47 utm01 suricata: [101012] <Error> -- [ERRCODE: SC_ERR_INVALID_VALUE(130)] - Thread count can't be greater than Rx ring count. Configured 2 threads for interface 'igb1_vlan1001' with 1 Rx rings.
```
This issue appears when VLAN interfaces are set.

To prevent high load of CPU, maybe the best way is to adjust in detect-thread-ratio param instead of netmap threads.

What do you think about it?

AdSchellevis:
@juliocbc I kind of expected that, usually you do not want to change these defaults. Hence my questions about the gain in performance. My guess is you better leave this as it is, unless there are accurate measurements showing a gain in performance without introducing issues such as packet loss.

juliocbc:
@AdSchellevis the problem is the high CPU usage with IPS enabled. I'll make new tests with `detect-thread-ratio` to see if with it we can prevent 100% of CPU load.

Closing this PR for now.

Thanks!

AdSchellevis:
100% at which packet rate? a c2758 isn't a very fast machine,  it might be totally logical that it's cpu bound... with netmap enabled, it has to process every packet.

juliocbc:
@AdSchellevis at 1Gbps.

AdSchellevis:
it likely won't do more....

PR #3275, opened 2019-03-02
Interfaces Overview - Relinquish DHCP Lease

NOYB:
Release/Renew Options:
Relinquish DHCP Lease - Send gratuitous DHCP release message to server.
Request a Fresh Address - Deletes the DHCP leases DB for the interface before the DHCP request is made.  Essentially this is an option to be used in conjunction with the "Relinquish DHCP Lease" release option.

Cleanup:
Use CSS for layout spacing rather than non-breaking white-spaces.
Correct interface post item naming - Interface description is being posted, so name the posted item accordingly for clarity.


AdSchellevis:
@NOYB I noticed your pull request often come with theme changes, but this makes it harder for us to add, since it infects other theme's out there.

NOYB:
Perhaps themes applied as layered overrides would be more flexible with less maintenance.  A base/main CSS to which theme specific overrides are applied.

NOYB:
Would it be possible to have the 'isc-dhcp44-client-4.4.1' package added to the OPNsense repo? (for '/usr/local/sbin/dhclient')

Have shell script that can send the release packet.  But dhclient is likely to be more robust.  But including a binary blob is not very appropriate in the spirit of open source I think.  It does come from the FreeBSD repo though.

Thanks

******

`root@OPNsense:~ # man -M /usr/local/man man 8 dhclient`

```
root@OPNsense:~ # /usr/local/sbin/dhclient --help
Usage: dhclient [-4|-6] [-SNTPRI1dvrxi] [-nw] [-p <port>] [-D LL|LLT]
                [--dad-wait-time <seconds>] [--prefix-len-hint <length>]
                [--decline-wait-time <seconds>]
                [--address-prefix-len <length>]
                [-s server-addr] [-cf config-file]
                [-df duid-file] [-lf lease-file]
                [-pf pid-file] [--no-pid] [-e VAR=val]
                [-sf script-file] [interface]*
       dhclient {--version|--help|-h}
```
```
root@OPNsense:~ # /usr/local/sbin/dhclient --version
isc-dhclient-4.4.1
```

jeremiah-rs:
If it passes muster, would it be possible to merge just the DHCP release/renew part of this commit?

The current DHCP release/renew function doesn't seem to work well on my router/ISP setup, I have to manually assign a new WAN MAC address in order to get a different IP from the ISP's DHCP server.

AdSchellevis:
timeout

Issue #3274, opened 2019-03-01
FR: Set IP of CSO during OpenVPN authentication using LDAP attributes


EugenMayer:
I am currently, with 18.1 using a custom API to generate local CCDs ( CSO ). Seeing that 19.1 no handles CSO dynamically on auth, i have checked the `IAuthConnector ` implementation of ldap

https://github.com/opnsense/core/blob/master/src/opnsense/mvc/app/library/OPNsense/Auth/LDAP.php#L468 

since that is the API triggered for the CSO attribute mapping using `$authenticator->getLastAuthProperties()`

https://github.com/opnsense/core/blob/master/src/etc/inc/plugins.inc.d/openvpn/auth-user.php#L128

**Question:** 
Am i right that basically all available LDAP attributes on the user being fetched are mapped into the hashmap that is then fetched by e.g. the CSO api with `$authenticator->getLastAuthProperties()`

https://github.com/opnsense/core/blob/master/src/opnsense/mvc/app/library/OPNsense/Auth/LDAP.php#L474

So whatever i have in LDAP, its part of that hashmap. That said, what would be the key i would need to set in ldap to map this to the client ip? Looking up https://github.com/opnsense/core/blob/master/src/etc/inc/plugins.inc.d/openvpn/auth-user.php#L60 i guess it would need to be `Framed-IP-Address` and `Framed-IP-Netmask` , right?

Now i am neither sure ldap would return those values case-sensitive nor i am not sure a field name like `Framed-IP-Address` would be allowed anyway. 

**Suggestion**
This said, i would like to either add a PR to map the "usual/common" ldap based ip/network defintion to https://github.com/opnsense/core/blob/master/src/etc/inc/plugins.inc.d/openvpn/auth-user.php#L57
 or alternatively have an extendable api so one can define the attribute mapping themself.

For the former i know `ipHostNumber` and `ipNetmaskNumber` for the netmask from the `ipHost` schema. Since we hardcoded the values for Radius in here, we might as well do that for ldap (for now) until having either an api like the latter or whatever suits you best. 

I would happily provide a PR for the former, but just wanted to know if that is something you guys would consider / i understood it right / it is desired. Thanks!

EugenMayer:
for an reference I tried to look up if `ipHostName` is an actual standard, which is rfc2307bis:
- https://wiki.archlinux.org/index.php/LDAP_Hosts
- https://docs.microsoft.com/en-us/windows/desktop/adschema/c-iphost
[...]

PR #3277, opened 2019-03-02
Feature/openvpn ldap cso mapping #3274

EugenMayer:
- Added an option, per LDAP server, to allow mapping IP/netmask for a client using the ldap fields
- allow `ipHostNumber` and `ipNetmaskNumber` to be used for this mapping 
- alternatively allow `ipHostNumber` to be in CIDR annotation alternativly
- feature is opt-in only and not active by default (server option)
- never use LDAP based ip mapping if radius already set header for IP mapping


AdSchellevis:
I know it's that simple to bootstrap values, I designed it, question is if this the intended behaviour ( https://github.com/opnsense/core/issues/3274#issuecomment-468899542 ), otherwise it might lead to side affects.

The radius values are intended for provisioning, the intend of ldap in this regard might be different.

EugenMayer:
valid point for sure, what about providing an opt-in switch for this. so either global or per LDAP server - what do you think?

EugenMayer:
@AdSchellevis updated the PR, see the description above. I incorporated your feedback using an opt-in feature per ldap server, of course disabled by default. 

Also i added that if freeradius IP headers are set, ldap is never used ( so it cannot override it ) - since semantically Radius should have the upper hand ( is always more common to be used for this primally )

Also added the docs for that feature to the server help field.


This done, this PR can not have a side-effect for:
 - any Radius or LDAP users
 - only can start impacting if you activate this feature AND does not have Radius setting headers already (anyway )

I think this way we have a 2 way safe-net and this should only be used if "needed and known that it can be used"

What do you think?

EugenMayer:
So for now, i did some testing and made a little lab with opnsense-starter:

You would need `vagrant` and `docker` installed

```
git clone https://github.com/EugenMayer/opnsense-openvpn-ldap-cso
cd opnsense-openvpn-ldap-cso
make start
```

I will start a fresh opnsense 19.1 box, configured with one OpenVPN server with the enabled feature and configure LDAP server which runs in docker and configured in this OPNsense box as an auth server and an auth backend for that OpenVPN server.

Anyway, i do not expect you to run it. Still what i see is

i see this in the logs
```
openvpn: mapped tunnel_network using ldap:172.16.0.101/24
```

corresponding to https://github.com/opnsense/core/pull/3277/files#diff-9c75fb57779afec35f1761aad244f6a1R86 .. so i seems to work because ldap sets
https://github.com/EugenMayer/docker-image-ldapexample/blob/master/testserver.ldif#L47

But the client actually gets

```
2019-03-02 23:10:29: /sbin/ifconfig utun1 172.0.16.2 172.0.16.2 netmask 255.255.255.0 mtu 1500 up
```

So it seems, while the right tunnel configuration is returned, it is not working. Is something else wrong?

EugenMayer:
arg after pasting the configuration refrences i found out that i typo'ed the tunnel network configuration to `172.0.16.0/24` and not `172.16.0.0/24` and that is what it was not applied, fixed that in the opnsense configuration and now it all works like expected

connecting with `included1`

```
2019-03-02 23:46:00: /sbin/ifconfig utun1 172.16.0.101 172.16.0.101 netmask 255.255.255.0 mtu 1500 up
```

which is what is expected due to https://github.com/EugenMayer/docker-image-ldapexample/blob/master/testserver.ldif#L47

Changing to `172.16.0.101/32` in ldap worked too

```
2019-03-02 23:42:27: /sbin/ifconfig utun1 172.16.0.101 172.16.0.101 netmask 255.255.255.255 mtu 1500 up
```

Using Changing to `172.16.0.101`, so without CIDR and without `ipNetmaskNumber` it "failed backup" on regular ip assignment

```
2019-03-02 23:46:56: /sbin/ifconfig utun1 172.16.0.2 172.16.0.2 netmask 255.255.255.0 mtu 1500 up
```

Then i added `ipNetmaskNumber` with `255.255.255.255` and got (leaving `ipHostNumber` at `172.16.0.101` )

```
2019-03-02 23:52:44: /sbin/ifconfig utun1 172.16.0.101 172.16.0.101 netmask 255.255.255.255 mtu 1500 up
```

So also as expected.

Then the important test, went to the server and deactivated `Map IP/Netmask from LDAP fields` and connected again and got

```
2019-03-02 23:54:04: /sbin/ifconfig utun1 172.16.0.2 172.16.0.2 netmask 255.255.255.0 mtu 1500 up
```

So even though configured, the feature is disabled, no IP is mapped.

So i guess all the features above work as expected and also the defensiv opt in works as expected. Anything else?

EugenMayer:
running this is production now, seems to work as expected, assigned 58 client ips

EugenMayer:
There is also something which needs attention here / docs ( how ever ), on needs to also activate

`Force CSO Login Matching`

To be honest i am not entirely sure what this means. i guess, only the common name, so lets say the CN / UID "part" will be matched against the CSO, not the whole "DN" - is that about right? I understand this is the wrong place to ask - but this thing is interlinked with the FR here, since LDAP based "lookup" would return a DN in the first place and AFAICS enabling `Force CSO Login Matching` is a must to have any CSO related features _anyway_ - even if creating custom, local CSO entries ( static ). Right? Maybe this needs a little more notice, right now its just `Use username instead of common name to match client specfic override.` ( would be a different PR i guess?)

AdSchellevis:
image

EugenMayer:
Anything else i could finish this PR with / anything missing?

PR #3279, opened 2019-03-02
System/General - Set default navigation appearance

opnsenseuser:
My starting point for the following consideration was that whenever I log in to the web interface, I get the old menu view even though I almost exclusively use the collapsed menu view.

so i made a new entry in the system - general tab!
You can now select which menu will be displayed automatically after the login.
Saving the selected entry works fine!

grafik

Unfortunately I do not know how to load either the default navigation or the sidebar navigation as standard menu after the login. Can anybody help me further?

Regards
René

### DIFF
--- a/src/www/system_general.php
+++ b/src/www/system_general.php
@@ -148,7 +148,7 @@ if ($_SERVER['REQUEST_METHOD'] === 'GET') {
             unset($config['system']['dnsallowoverride']);
         }
         
-        if (!empty($config['system']['navigation'])) {
+        if (empty($config['navigation'])) {
             $config['system']['navigation'] = 'standard';
         } elseif ($pconfig['navigation'] == "standard") {
             $config['system']['navigation'] = 'standard';
@@ -361,10 +361,10 @@ include("head.inc");
             <tr>
               <td><a id="help_for_navigation" href="#" class="showhelp"><i class="fa fa-info-circle"></i></a> <?=gettext("Navigation"); ?></td>
                 <td>
-                  <input name="navigation" class="navigation" id="navigation_standard" type="radio" value="standard" <?= $pconfig['navigation'] == "standard" ? 'checked="checked"' : '' ?>/>
-                  <?=gettext("Standard"); ?>
-                  <input name="navigation" class="navigation" id="navigation_sidebar" type="radio" value="sidebar" <?= $pconfig['navigation'] == "sidebar" ? 'checked="checked"' : '' ?>/>
-                  <?=gettext("Sidebar"); ?>
+                  <span style="margin-right:10px"><input name="navigation" class="navigation" id="navigation_standard" type="radio" value="standard" <?= $pconfig['navigation'] == "standard" ? 'checked="checked"' : '' ?>/>
+                  <?=gettext("Standard"); ?></span>
+                  <span><input name="navigation" class="navigation" id="navigation_sidebar" type="radio" value="sidebar" <?= $pconfig['navigation'] == "sidebar" ? 'checked="checked"' : '' ?>/>
+                  <?=gettext("Sidebar"); ?></span>
                   <br />
                   <div class="hidden" data-for="help_for_navigation">
                     <?=gettext('Set default navigation menu')?>
--- a/src/www/system_general.php
+++ b/src/www/system_general.php
@@ -47,6 +47,7 @@ if ($_SERVER['REQUEST_METHOD'] === 'GET') {
     $pconfig['language'] = $config['system']['language'];
     $pconfig['prefer_ipv4'] = isset($config['system']['prefer_ipv4']);
     $pconfig['theme'] = $config['theme'];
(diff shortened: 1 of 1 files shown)


src/www/system_general.php
```diff
@@ -348,6 +356,21 @@
                 </div>
               </td>
             </tr>
+            <tr>
+              <td><a id="help_for_navigation" href="#" class="showhelp"><i class="fa fa-info-circle"></i></a> <?=gettext("Navigation"); ?></td>
+                <td>
+                  <input name="navigation" class="navigation" id="navigation_standard" type="radio" value="standard" <?= $pconfig['navigation'] == "standard" ? 'checked="checked"' : '' ?>/>
+                  <?=gettext("Standard"); ?>
+                  &nbsp;&nbsp;&nbsp;
```
fabianfrz: try to avoid non breaking spaces - it is an ugly hack from old days. you could use inline blocks etc. which have a padding and margin or switch to a select box

src/www/system_general.php
```diff
@@ -348,6 +356,21 @@
...
+                  <input name="navigation" class="navigation" id="navigation_sidebar" type="radio" value="sidebar" <?= $pconfig['navigation'] == "sidebar" ? 'checked="checked"' : '' ?>/>
+                  <?=gettext("Sidebar"); ?>
+                  <br />
+                  <div class="hidden" data-for="help_for_navigation">
+                    <?=sprintf(
+                      gettext('Set default navigation menu'));?>
```
fabianfrz: remove the sprintf call here - it has no placeholders

fabianfrz:
the value is currently never used - you need to integrate them into the frontend - otherwise the setting is never applied.

opnsenseuser:
@fabianfrz ok. thx for your support. i will do my best and upload it again! 👍 regards rené

fichtner:
@opnsenseuser this is a good concept, but I'm unsure about the implementation for the moment. let me think about it :)

opnsenseuser:
@fichtner 👍

opnsenseuser:
@fabianfrz i changed the things you said and i made an if !empty command to set a value if it doesn´t exist. I hope that this command aims to set a default value if nothing is set!

AdSchellevis:
@opnsenseuser we probably better save this in your sessionStorage, it's more a user preference than a system preference (same as the expanded advanced section). If you need help storing it in your session storage, I can help you with that.

opnsenseuser:
@AdSchellevis If you could show me how that works, then I would be very grateful. I'm in a learning phase.

AdSchellevis:
@opnsenseuser hmm, while looking at the code, it looks like the intend was already to remember your previous setting:

https://github.com/opnsense/core/blob/master/src/opnsense/www/js/opnsense_theme.js#L200-L202

https://github.com/opnsense/core/blob/master/src/opnsense/www/js/opnsense_theme.js#L209-L212

We probably should have used localstorage in there to outlive the session.
Just to be sure, the intend was to remember my previous dashboard setting after login (on the same device)?

opnsenseuser:
@AdSchellevis I do not use cookies in my browser. for safety reasons. can it be related to that?

opnsenseuser:
@AdSchellevis It would still be great if we could set or save this settings to the system. Do you really not want that? then I would delete the request again. Alright? Regards rene

AdSchellevis:
@opnsenseuser Personally I don't see a lot of value in it, plus not all theme's might support the feature (leading to questions we don't want to answer). Without cookies enabled, local storage probably won't work. Keep it open for now, we'll discuss internally if/how we want to support this.

opnsenseuser:
@AdSchellevis 👍

opnsenseuser:
@AdSchellevis is there any news on this topic? How did you decide?
regards!

AdSchellevis:
@opnsenseuser sorry, not yet, other priorities at the moment. We'll leave it open for now.

opnsenseuser:
@AdSchellevis I'm sorry that I'm talking to you again. Is there any news or could you possibly tell me when you want to talk about the topic? I think this idea is really good, although of course it does not have the highest priority for a firewall. Thanks and best regards, René

AdSchellevis:
@opnsenseuser no problem, we just haven't got time to discuss this internally. We'll let you know as soon as we do.

opnsenseuser:
@AdSchellevis 
@fabianfrz 
@fichtner 
any news on this? 
I'm just asking, because a user has contacted me via pm and asked me if this is still in process. I still think the idea is great. what do you think?  maybe we can implement that for 20.1? regards rené

AdSchellevis:
I don't think we should do this, with cookies enabled in the browser, it already remembers your settings (as stated before)

fichtner:
Upon further reflection the current flexibility is nice to have the sidebar enabled on demand. If we make a setting we lose the flexibility to do this because each page load will reset the previous choice. Plus making this functional we need to inject some sort of JavaScript for this config setting to take effect which isn't ideal. For now we need to shelf this until a better strategy is found.

Issue #3272, opened 2019-03-01
[Feature Request] Create secret key/cert, with downloading secret key and without saving secret key


nhirokinet:
**Imortant notices**
Before you add a new report, we ask you kindly to acknowledge the following:

[v] I have read the contributing guide lines at https://github.com/opnsense/core/blob/master/CONTRIBUTING.md

[v] I have searched the existing issues and I'm convinced that mine is new.

**Is your feature request related to a problem? Please describe.**

OpenVPN for Android seems not to have a menu to create PKI key pair. I (and most Android users, I guess) did not install, or feel hard to install, terminals apps avaiable to use `openssl` commands.

Currently I do in the system **other than OPNsense**:

- Create key pair and CSR on laptop
- Sign a CSR on CA
- Copy private key, CSR, and ovpn files to SD card or equivalent area of Android device
- Import ovpn file in OpenVPN for Android, and remove credentials as this is visible for all apps in the device
- Erase credentials from laptop

This seems to be not too bad, but I think it is not the best for the very nice web UI implemented in OPNsence.

**Describe the solution you'd like**

Procedures like:

- Create private key/CSR pair in web UI (within one page, not saved anywhere)
- Download the private key (credentials saved only in clients)
- Press some button (I guess this is written "Next")  that makes private key downloading link not available anymore (note that stopping here makes nothing; the private key is just a useless file here)
- Seemlessly sign CSR and certificates are added to certmanager, where config.xml is modified first in this procedure (This makes this operation meaning, and this is commited to OPNsense storage)
- Then the secret key is available only for the admin who did this procedures (and the admin who broke into process protection or broke system, if any), and trusted by the CA cert

With conditions that no credentials are written into disk/SSD of OPNsense except for swaps.

**Describe alternatives you've considered**

There might be multiple choices:

- Plan 1: adding "download secret key and fix CSR" button to the screen added in #3234 
- Plan 2: adding some sub-menus to "Create an internal Certificate" menu
- Plan 3: adding additional menu to "Method" in system_certmanager.php
- Plan 4: create certificate like before. Just add "erase private key".
  - First I thought this, but this is not good that this remains in System > Configuration > Backups unless deleting explicitly.

There plans differs in a view point of implementation. Also, there may be some other better opinion.

**Additional context**

I'm willing to write codes and make a pull request after #3234 is merged, as long as no one else does it earlier than me; not preventing someone willing to write a code for this issue, of course.

I'm waiting for comments, especially for which plan seems good.

[...]

PR #3282, opened 2019-03-04
Add option to create key pair without saving onto firewall on "Create Internal Certificate"

nhirokinet:
This is a proposal for #3272 . This enables creating a private only for the device who opens admin web page.

This is not elegant for the point that secret key is sent back to the OPNsense, which produces double traffic, but on situation that OPNsense does not want to save the secret key, I thought significant improvement against this is hard.

### DIFF
--- a/src/www/system_certmanager.php
+++ b/src/www/system_certmanager.php
@@ -366,7 +366,6 @@ if ($_SERVER['REQUEST_METHOD'] === 'GET') {
         echo $result_stderr;
       }
       exit;
-
     } elseif ($act == 'csr_info_json') {
       header("Content-Type: application/json;charset=UTF-8");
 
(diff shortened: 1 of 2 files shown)


AdSchellevis:
I think we can simplify this a lot by just returning the private key data when generated and let the user save the certificate to a file (start with a text area containing the key).

unset the private key here when requested:
https://github.com/opnsense/core/blob/f685abcc9c91805e9f5dad35f8f30bc4bb6c585f/src/www/system_certmanager.php#L738-L740

Fall through after post, ignoring the redirect here
https://github.com/opnsense/core/blob/f685abcc9c91805e9f5dad35f8f30bc4bb6c585f/src/www/system_certmanager.php#L751-L756

Then render the page with the private key, you can probably use `$act` for flow control.

It might still not be the prettiest solution in the world, and to be honest I rather see cleanups to the code first before adding more logic, but it could lead to an acceptable solution.

nhirokinet:
Thank you for comment, I tried to improve my code accoding to your advice.

For newly written HTML I referred some HTML in other places of OPNsense, but I'm not sure whether this is expected in OPNsense.

src/www/system_certmanager.php
```diff
@@ -1087,9 +1108,41 @@ function removeRowAltNm() {
         $("#certmethod").change();
     }
   });
+
+  function refresh_download_link(jquery_key, content, filename) {
+      if (navigator.userAgent.indexOf('MSIE ') !== -1 || navigator.userAgent.indexOf('Trident') !== -1) {
+          // IE11 support; they do not have <a download="">
```
AdSchellevis: I'm not sure we should go through all that trouble trying to support IE11, what if we just make sure the user can copy-paste the content and optionally save to file if supported by the browser?

nhirokinet: Okay, I will remove download link if unsupported.

nhirokinet:
Edited so that check download attribute is available, and if not remove the link.

AdSchellevis:
@nhirokinet cleanups in https://github.com/opnsense/core/commit/fa7e7d878a05b7bc09281974e142420bdeb906f5

AdSchellevis:
thanks!

PR #3285, opened 2019-03-05
Add help for 'Use vt driver'

MichaelDeciso:
Came across it when writing documentation.

### DIFF
--- a/src/www/system_advanced_admin.php
+++ b/src/www/system_advanced_admin.php
@@ -720,10 +720,13 @@ $(document).ready(function() {
                 <td style="width:78%"></td>
               </tr>
               <tr>
-                <td><i class="fa fa-info-circle text-muted"></i> <?= gettext('Console driver') ?></td>
+                <td><a id="help_for_usevirtualterminal" href="#" class="showhelp"><i class="fa fa-info-circle"></i></a> <?= gettext('Console driver') ?></td>
                 <td>
                   <input name="usevirtualterminal" type="checkbox" value="yes" <?= empty($pconfig['usevirtualterminal']) ? '' : 'checked="checked"' ?>  />
                   <?= gettext('Use the virtual terminal driver (vt)') ?>
+                  <div class="hidden" data-for="help_for_usevirtualterminal">
+                    <?=gettext('When unchecked, the older sc driver will be used instead.') ?>
+                  </div>
                 </td>
               </tr>
               <tr>


src/www/system_advanced_admin.php
```diff
@@ -720,10 +720,13 @@ function reloadWaitOld () {
                 <td style="width:78%"></td>
               </tr>
               <tr>
-                <td><i class="fa fa-info-circle text-muted"></i> <?= gettext('Console driver') ?></td>
+                <td><a id="help_for_usevirtualterminal" href="#" class="showhelp"><i class="fa fa-info-circle"></i></a> <?= gettext('Console driver') ?></td>
                 <td>
                   <input name="usevirtualterminal" type="checkbox" value="yes" <?= empty($pconfig['usevirtualterminal']) ? '' : 'checked="checked"' ?>  />
                   <?= gettext('Use the virtual terminal driver (vt)') ?>
+                  <div class="hidden" data-for="help_for_usevirtualterminal">
+                    <?=gettext('When unchecked, the older sc driver will be used instead.') ?>
```
fichtner:
```suggestion
                    <?=gettext('When unchecked, the sc(4) driver will be used instead.') ?>
```

fichtner: Maybe bention vt(4) for consistency as well.... not sure... what do you think?

MichaelDeciso: I think it's important to be consistent, so I think we should either use `vt(4)` and `sc(4)` or `vt` and `sc`.

fichtner: Ok, agreed, then how about killing the "(vt)" annotation and explaining everything in the help text about the vt(4) default and the sc(4) option on disable?

MichaelDeciso: I would do it this way:

```diff
-Use the virtual terminal driver (vt)
+Use the vt(4) driver
```

It seems very unlikely to me that people who don't know what vt(4) and sc(4) are would want to change this setting, so writing it out as "virtual terminal" does not help anyone.

fichtner: hmm, no, I think that is too cryptic, that's why I used "virtual terminal". Unfortunately there doesn't seem to be a good name for "sc" so we are stuck with (4) notation, might be better to only do it in the help text.

MichaelDeciso: I don't think it's too cryptic really - who is going to change this apart from the people who actually know what vt and sc are?

fichtner:
Let's leave this as is then.

MichaelDeciso:
Closing this means that *no* explanation is offered about what this option changes, not even that it will cause the system to use the sc(4) driver.

fichtner:
Your proposal is not good enough for me. The follow-up discussion is a time sink. I'd like to work on something else.

PR #3286, opened 2019-03-05
core/vpn/ipsec: allow easier override of colors in themes

fabianfrz:
see #2700

### DIFF
--- a/src/www/widgets/widgets/ipsec.widget.php
+++ b/src/www/widgets/widgets/ipsec.widget.php
@@ -77,29 +77,46 @@ if (isset($config['ipsec']['phase2'])) {
 ?>
 <script>
     $(document).ready(function() {
-        $(".ipsec-tab").unbind('click').click(function(){
-            $(".ipsec-tab").css('background-color', '#777777');
-            $(".ipsec-tab").css('color', 'white');
-            $(this).css('background-color', '#EEEEEE');
-            $(this).css('color', 'black');
+        $(".ipsec-tab").unbind('click').click(function() {
+            $('.ipsec-tab').removeClass('activetab');
+            $(this).addClass('activetab');
             $(".ipsec-tab-content").hide();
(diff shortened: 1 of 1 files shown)


AdSchellevis:
@fabianfrz thanks!

src/www/widgets/widgets/ipsec.widget.php
```diff
@@ -77,29 +77,46 @@
 ?>
 <script>
     $(document).ready(function() {
-        $(".ipsec-tab").unbind('click').click(function(){
-            $(".ipsec-tab").css('background-color', '#777777');
-            $(".ipsec-tab").css('color', 'white');
-            $(this).css('background-color', '#EEEEEE');
-            $(this).css('color', 'black');
+        $(".ipsec-tab").unbind('click').click(function() {
```
fabianfrz: @AdSchellevis should we remove unbind too - it has been removed from standard jquery?

opnsenseuser:
two questions. 
1. I thought we already defined the classes in the new themes. do we have to change things again for tukan, dark, cicade? 
2. I am not an experienced programmer, but shouldn’t it be better to replace the syntax of .unbind with .off in the source code?

Regards rene

AdSchellevis:
@opnsenseuser if their already defined, they should be preferred if `!important `is set, the unbound is ok, `unbind` could be replaced by `off`, but both seem to doing the same... jQuery is confusing sometimes ;)

src/www/widgets/widgets/ipsec.widget.php
AdSchellevis: https://github.com/opnsense/core/pull/3286#issuecomment-469782949 , but I don't mind.

fabianfrz:
@AdSchellevis unbind does not exist anymore. off should be used now as without the migrate plugin it would crash. The difference is that unbind enforces that the element exists in contrast to on/off

Issue #3295, opened 2019-03-07
[Feature request] VPN IPsec Mobile Clients: Radius: Assign specific virtual IP range to users in specific groups


stumbaumr:
**Is your feature request related to a problem? Please describe.**
I would like to assign IPsec mobile users into separate virtual IP pools. I require this to allow/prohibit access to users in certain groups to specific network resources.

Example: Three groups (Admins, Developers, Users) have separate virtual IP pools. These network ranges are then used in firewall rules on the current IPsec peer or even on another firewall behind a IPsec tunnel in a different location.

The groups and users are configured on a remote Radius server.

**Describe the solution you'd like**
I would like to use the "Group Selection" feature described here:
https://wiki.strongswan.org/projects/strongswan/wiki/EAPRadius

I understand that this is a very specific setup. So maybe it would be best to have a text field "Advanced Configuration" where the content is saved into `/usr/local/etc/ipsec.advanced.conf` and an `include ipsec.advanced.conf` statement in ipsec.conf. Almost same as OpenVPN has it.

For strongswan.conf the only requirement is to add `class_group = yes` to the eap-radius section. It seems it does not hurt in case it is set and no class is passed back from the Radius server. 

**Describe alternatives you've considered**
As far as I understand there is currently no functionality that allows users to be assigned into specific virtual IP ranges.

And from my searches it seems I am not the first person to ask.

Here is my tested configuration I would like in ipsec.advanced.conf

```
conn mobile
  aggressive = no
  fragmentation = yes
  keyexchange = ike
  mobike = yes
  reauth = yes
  rekey = yes
  forceencaps = yes
  installpolicy = yes
  ikelifetime = 28800s
  ike = aes256-aesxcbc-modp2048,aes256-aesxcbc-modp1024,aes256-sha512-modp2048,aes256-sha512-modp1024,aes256-sha384-modp2048,aes256-sha384-modp1024,aes256-sha256-modp2048,aes256-sha256-modp1024!
  dpdaction = clear
  dpddelay = 10s
  dpdtimeout = 60s
  left = XX.XX.XX.XX
  leftid = vpn.example.com
  leftauth = pubkey
  leftcert = /usr/local/etc/ipsec.d/certs/cert-2.crt
  leftsendcert = always
  rightsendcert = never
  right = %any

conn mobile-ops
  also = mobile
  rightauth=eap-radius
  eap_identity = %any
  rightgroups="VPN-OPS" 
[...]

PR #3298, opened 2019-03-08
Implemented IPsec Mobile client expert tunnel configuration

stumbaumr:
Hi,

we have the requirement to assign separate virtual IP pools to different IPsec mobile user groups. I understand that adjusting the WebUI to cater for our specific need is rather complex. I went through comparable requests from other people and realised their requirements are still well different from ours.

So my conclusion was that there are two possible setups
- The existing options are good enough for most people
- The desired options are so complex that a WebUi will never be sufficient

I opened therefore a feature request (https://github.com/opnsense/core/issues/3295 ) to allow for an expert tunnel configuration.

This PR is the result from above feature request.

We had to rearrange some of the code for generation strongswan.conf which is in our opinion backwards compatible.

The only new fixed addition is `class_group = yes` - but that should have no influence as well.

So we are awaiting possible feedback and hope to get that expert tunnel config included.

Best regards
Mark & Rainer

### DIFF
--- a/src/etc/inc/plugins.inc.d/ipsec.inc
+++ b/src/etc/inc/plugins.inc.d/ipsec.inc
@@ -958,42 +958,23 @@ function ipsec_configure_do($verbose = false, $interface = '')
                 $strongswanTree['charon']['plugins']['attr']['28679'] = $a_client['pfs_group'];
             }
 
-            $disable_xauth = false;
+            $radius_enabled = false;
             foreach ($a_phase1 as $ph1ent) {
                 if (!isset($ph1ent['disabled']) && isset($ph1ent['mobile'])) {
                     if ($ph1ent['authentication_method'] == "eap-radius") {
-                        $disable_xauth = true; // disable Xauth when radius is used.
-                        $strongswanTree['charon']['plugins']['eap-radius'] = [];
-                        $strongswanTree['charon']['plugins']['eap-radius']['servers'] = [];
-                        $radius_server_num = 1;
-                        $radius_accounting_enabled = false;
-                        foreach (auth_get_authserver_list() as $auth_server) {
-                            if (in_array($auth_server['name'], explode(',', $ph1ent['authservers']))) {
-                                $server = [
-                                    'address' => $auth_server['host'],
-                                    'secret' => '"' . $auth_server['radius_secret'] . '"',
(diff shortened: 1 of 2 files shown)


AdSchellevis:
maybe we could think of some sort of hooking, but plain text gui input for configuration files is something we don't want to support (impossible to validate, prone to all sorts of errors).

stumbaumr:
So are you going to remove the „Advanced configuration“ from OpenVPN? 
There seems to be no validation as well...

AdSchellevis:
Unfortunately not, no, it's one of the very few leftovers from our legacy base. Although we're not planning on implementing new future issues.

For new components we often try to support template hooks, which allow bootstrapping custom configs, which isn't available for regular users. Maybe we can think of something flexible here as well, if time permits we would like to cleanup some parts in IPsec anyway.

stumbaumr:
So we are just going to go live with our patch then and maintain it until there is a proper solution...

stumbaumr:
So then: How about just always having a `include ipsec.*.conf` in ipsec.conf and a `include ipsec.*.secrets` at the bottom of the generated files?

Would a PR for that be acceptable? It requires anyhow knowledge of the command line and only the knowledgable are creating files manually.

AdSchellevis:
@stumbaumr I think it will, strongswan indeed has an excellent include option there.

stumbaumr:
So is a PR acceptable that only adds the includes for all the time or do you want a checkbox to activate that funktionality specifically?

I can imagine users complaining about magically appearing conn's that are configured on the filesystem and then forgotten about...

fabianfrz:
Our includes are usually hardcoded.

See:
* Squid: https://github.com/opnsense/core/blob/master/src/opnsense/service/templates/OPNsense/Proxy/squid.conf#L281-L291
* nginx (similar as squid): https://github.com/opnsense/plugins/blob/master/www/nginx/src/opnsense/service/templates/OPNsense/Nginx/http.conf#L67, https://github.com/opnsense/plugins/blob/master/www/nginx/src/opnsense/service/templates/OPNsense/Nginx/http.conf#L242, https://github.com/opnsense/plugins/blob/master/www/nginx/src/opnsense/service/templates/OPNsense/Nginx/http.conf#L259

PR #3307, opened 2019-03-10
Not to translate hostname example

nhirokinet:
On System > Settings > General, hostname example is translated.

キャプチャ

(Like: "Example: ファイヤーウォール" for hostname)

Same as here:

キャプチャ

This example contains invalid characters.

キャプチャ

(This alerts says that hostname can contain only a-z, 0-9, and -, which is correct.)

As long as I see, Czech and Japanese are affected. French, on the other hand, seems to have a French text which is valid hostname: "pare-feu".


fichtner:
Nice catch. It would be better to remove the line completely plus the br-tag above to simplify (it's a cluttered sentence and thus hard to translate cleanly)

nhirokinet:
I removed it. I noticed "without domain part" would do a job rather than this example.

fichtner:
Yes, exactly, thank you very much... merged! :)

PR #3318, opened 2019-03-12
Use system log file for rtsold dhcp6c messages

tduboys:
Hi all,

Since Opnsense 19.x, the main log file used by dhcp6c is system.log, not dhcpd.log like on 18.7

I didnt understand where this change was made, but there is another log related to dhcp6c on dhcpd.log

Maybe the good way is to have a dedicated log file, but on the 19.1.3, it's better to have all dhcp6c-related logs on the same place.

Thanks

### DIFF
--- a/src/etc/inc/interfaces.inc
+++ b/src/etc/inc/interfaces.inc
@@ -3001,10 +3001,10 @@ function interface_dhcpv6_prepare($interface = 'wan', $wancfg)
     $rtsoldscript .= "\techo \${2} > /tmp/{$wanif}_defaultgwv6\n";
     $rtsoldscript .= "fi\n";
     $rtsoldscript .= "if [ -f /var/run/dhcp6c_{$wanif}.pid ]; then\n";
-    $rtsoldscript .= "\t/usr/bin/logger -t dhcpd \"RTSOLD script - Sending SIGHUP to dhcp6c for interface {$interface}({$wanif})\"\n";
+    $rtsoldscript .= "\t/usr/bin/logger -t dhcp6c \"RTSOLD script - Sending SIGHUP to dhcp6c for interface {$interface}({$wanif})\"\n";
     $rtsoldscript .= "\t/bin/pkill -HUP -F /var/run/dhcp6c_{$wanif}.pid\n";
     $rtsoldscript .= "else\n";
-    $rtsoldscript .= "\t/usr/bin/logger -t dhcpd \"RTSOLD script - Starting dhcp6 client for interface {$interface}({$wanif})\"\n";
+    $rtsoldscript .= "\t/usr/bin/logger -t dhcp6c \"RTSOLD script - Starting dhcp6 client for interface {$interface}({$wanif})\"\n";
     $rtsoldscript .= "\t$dhcp6ccommand\n";
     $rtsoldscript .= "fi\n";
 


src/etc/inc/interfaces.inc
```diff
@@ -3001,10 +3001,10 @@ function interface_dhcpv6_prepare($interface = 'wan', $wancfg)
     $rtsoldscript .= "\techo \${2} > /tmp/{$wanif}_defaultgwv6\n";
     $rtsoldscript .= "fi\n";
     $rtsoldscript .= "if [ -f /var/run/dhcp6c_{$wanif}.pid ]; then\n";
-    $rtsoldscript .= "\t/usr/bin/logger -t dhcpd \"RTSOLD script - Sending SIGHUP to dhcp6c for interface {$interface}({$wanif})\"\n";
+    $rtsoldscript .= "\t/usr/bin/logger \"RTSOLD script - Sending SIGHUP to dhcp6c for interface {$interface}({$wanif})\"\n";
```
fichtner:
```suggestion
    $rtsoldscript .= "\t/usr/bin/logger -t dhcp6c \"RTSOLD script - Sending SIGHUP to dhcp6c for interface {$interface}({$wanif})\"\n";
```

src/etc/inc/interfaces.inc
```diff
@@ -3001,10 +3001,10 @@ function interface_dhcpv6_prepare($interface = 'wan', $wancfg)
...
     $rtsoldscript .= "\t/bin/pkill -HUP -F /var/run/dhcp6c_{$wanif}.pid\n";
     $rtsoldscript .= "else\n";
-    $rtsoldscript .= "\t/usr/bin/logger -t dhcpd \"RTSOLD script - Starting dhcp6 client for interface {$interface}({$wanif})\"\n";
+    $rtsoldscript .= "\t/usr/bin/logger \"RTSOLD script - Starting dhcp6 client for interface {$interface}({$wanif})\"\n";
```
fichtner:
```suggestion
    $rtsoldscript .= "\t/usr/bin/logger -t dhcp6c \"RTSOLD script - Starting dhcp6 client for interface {$interface}({$wanif})\"\n";
```

let's keep this closer to dhcp6c than dhcpd or rtsold as it is a wrapper script...

tduboys: I'm ok with that, but the idea is to get this log close to the dhcp6c log who uses syslog.

I will look about syslog config to see if I can redirect all to a dedicated log.

tduboys: I added a syslog rule to move all dhcp6c messages to /var/log/dhcp6c.log

Messages are always available on the GUI on system / log files / general

fichtner: No, this is the wrong direction. The logs were moved to system because there is no proper "interface" log where all connectivity stuff is pushed to. System takes care of routing, new IP, but we never see the DHCP (v6 and v6) actions so it was moved there. This will revert back to old behaviour and make it harder for following a reload sequence.

tduboys: Not sure to understand what you mean ?
You prefer to keep all the logs on the system.log but with the good facitily information (dhcp6c in this case) instead of using separated log files for each application ?
In this case, the first suggestion you made will do the job. The only thing I think is important is to have the rtsold and dhcp6c logs in the same place.

fichtner: Yes and no. -t dhcpd was clearly wrong and I thank you for spotting it. If we follow the route of "it's rtsold" it's going to be pushed to the routing log. We don't want this any more than dhcpd log. So now the question is between no -t parameter or -t dhcp6c for which I think that dhcp6c fits better and solves the issue like you originally intended.

tduboys: Yes, I understand now :)
I removed the syslog rule, just fixed the logger facility for this 2 messages.

fichtner:
Thanks, merged!

Issue #3199, opened 2019-02-06
Parsing configuration failed at startup between bridge and dhcp6c on WAN


tduboys:
Hi all,

Following https://github.com/opnsense/dhcp6c/issues/7, I've an issue that is probably not on the dhcp6c client but maybe on config testers during boot sequence.

I'm using OPNSense following this setup : https://wiki.opnsense.org/manual/how-tos/orange_fr_fttp.html

My box is set up as : 
- igb0 as WAN
- igb1, igb2, igb3… as bridge0
- bridge0 as LAN

I'm using a QOTOM hardware with 6x Intel interfaces.

On startup, dhcp6c client parses the config file and see that my LAN interface (bridge0) is tracking the WAN interface to get IPv6 configuration, but bridge0 is not set up at this moment, so the config fails and dhcp6c didnt start.
```
Jan  8 20:19:12 opnsense dhcp6c[38999]: /var/etc/dhcp6c_wan.conf:13 invalid interface (bridge0): Device not configured
Jan  8 20:19:12 opnsense dhcp6c[38999]: called
Jan  8 20:19:12 opnsense dhcp6c[38999]: failed to parse configuration file
```

I need to force a dhcp refresh by saving the WAN page settings and click on « apply », on each startup, to get IPv6 working correctly.

How to reproduce : 
- set up a WAN interface with dhcp6c
- set up a LAN interface with bridge on multiple physical interfaces
- set up id-assoc to the bridge interface 
```
id-assoc pd 0 {
  prefix-interface bridge0 {
    sla-id 0;
    sla-len 8;
  };
};

```
- reboot
- see logs

What expected : 
- no errors during boot, ipv6 working directly

I had this issue on 1.8.7 and also after upgrading to 1.9.1

marjohn56:
Are you still having this issue?

tduboys:
Hi,
I'm on 19.1.1 version, and I rebooted this morning.
The issue is still present with the same message on dhcpd.log

marjohn56:
Can you send me your config.xml file, I'll take a look and see what gives.

marjohn56:
The thing that confuses me is that you say bridge0 is not setup, yet the configuration is finding an interface called bridge0 - You must have entered that somewhere.

tduboys:
I just updated to 19.1.2 but nothing changed.
I've the config.xml (just cleaned my passwords/keys and dhcp auth settings (not other dhcp settings))
But what's the best way to send it to you ?

[...]

PR #3319, opened 2019-03-12
do not merge : configure standard interfaces at last

tduboys:
Hi,
It's about issue #3199 

If you have a tracked interface on a bridge, setup to track a « standard » interface (ie. no grp, lagg, bridge, etc…) the bridge will set up after the standard interface, and the dhcp6c need that the tracked interface up to work.

The idea here is to set up the interfaces with no dependencies to others at the end.

I just tested on my situation and it resolves my issue. But this kind of change needs to be tested a lot.

### DIFF
--- a/src/etc/inc/interfaces.inc
+++ b/src/etc/inc/interfaces.inc
@@ -892,6 +892,7 @@ function interfaces_configure($verbose = false)
     $delayed_list = array();
     $bridge_list = array();
     $track6_list = array();
+    $standard_list = array();
 
     foreach (get_configured_interface_with_descr() as $if => $ifname) {
         $realif = $config['interfaces'][$if]['if'];
@@ -902,7 +903,7 @@ function interfaces_configure($verbose = false)
         } elseif (!empty($config['interfaces'][$if]['ipaddrv6']) && $config['interfaces'][$if]['ipaddrv6'] == 'track6') {
             $track6_list[$if] = $ifname;
         } else {
-            interface_configure($verbose, $if);
+            $standard_list[$if] = $ifname;
         }
     }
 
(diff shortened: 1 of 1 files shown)


src/etc/inc/interfaces.inc
```diff
@@ -902,7 +903,7 @@ function interfaces_configure($verbose = false)
         } elseif (!empty($config['interfaces'][$if]['ipaddrv6']) && $config['interfaces'][$if]['ipaddrv6'] == 'track6') {
             $track6_list[$if] = $ifname;
         } else {
-            interface_configure($verbose, $if);
+            $standard_list[$if] = $ifname;
```
fichtner: looks like the same as adding them to $bridge_list, but I'm quite sure this will cause issues for others from the looks of the amount of reprioritisation done here. no better idea yet, but thanks for looking into it

fichtner:
@tduboys basically the idea here is to move WAN devices who have enabled tracking LAN devices configured to be created last?

tduboys:
yes @fichtner it's the idea, and it's not exactly what I do in this PR, but I didnt understand exactly how the tracked interfaces was managed in this piece of code. At first look, I have the feeling that the tracked6 label is defined at the inverse of the reality.

the interface who have a tracked interface entry in the config need to be set up before the tracked interface.

fichtner:
@tduboys ok, thanks for confirming. I tried to read the bridge code and it has a number of bizzare spots. We'll essentially have to reverse-engineer this code to find out which parts we still need and what can be deleted. The fix may be a couple of lines added on top of the current complexity or breaking it all down and flushing a lot of code in the process... if we do this, we should do it right :)

fichtner:
@tduboys while I agree 100% with your approach I fear the fix will break something so I've started to work on #3199 with a bit more caution. See ticket for details. :)

PR #3322, opened 2019-03-13
Auth/LDAP: Fallback to 'commonName' if 'name' isn't available for full name

indrajitr:
Per [rfc2253](http://www.rfc-editor.org/rfc/rfc2253.txt), `cn` is commonly used to represent LDAP user's full name and is probably supported across standard directory servers.


AdSchellevis:
@indrajitr can you try https://github.com/opnsense/core/commit/9d3002cabbae4fe3cba288f94aec70c025a2f65e ? I've changed the if/else a bit to make things more readable.

indrajitr:
Perfect, thanks @AdSchellevis!

Issue #1831, opened 2017-09-18
squid: Log date/time formatting into gui


monster900:
Hi,
ist it possible to show the squid logs for access an store logs in a more user friendly time format into the gui?
Right now both logs are shown into unix time format.

best regards
Dirk

fichtner:
Hi Dirk,

It's not impossible, but there are far superior tools for Squid Log Analysis than we could ever write / provide. Have you checked out lightsquid?


Cheers,
Franco

Stephanowicz:
Hi,

I've thinking bout this, too - and according to [https://robinward.com/?p=372](https://robinward.com/?p=372) it's quite easy by adding something like
`logformat squid %tl.%03tu %6tr %>a %Ss/%03Hs %`
to the squid config.

This means one needs to add the changes to the proxy template in 

> /usr/local/opnsense/service/templates/OPNsense/Proxy/squid.conf 

I added it to line 337 - this is used when enabling "file extended" in the general settings
**%tl.%03tu %6tr** is added after `logformat opnsense      ` : 
`logformat opnsense      %tl.%03tu %6tr %>a %[ui %>eui %[un [%tl] "%rm %ru HTTP/%rv" %>Hs %<st "%{Referer}>h" "%{User-Agent}>h" %Ss:%Sh
`

After this modification the logfile enties changed from
`1506423311.903     94 192.168.200....`
to
`26/Sep/2017:12:58:12 +0200.969 274530 192.168.200....`

Cheers,

Stephan

monster900:
Stephan,
it's working! But only for the access log. The store log keep showing the unix time format. Is there another line into squid.conf to change the time format for store log too?
What happens after an update? I think that these changes gone after an update, right?

best regards
Dirk

fichtner:
Yes, it should be added to OPNsense else it's overwritten. I don't mind changing the format, but what happens to e.g. lightsquid analysis in these cases?

Stephanowicz:
@fichtner do these depend on the unix time format?

@monster900 as far as I know it's not possible for the store log

probably a custom format field would be helpful... though one needs  to know what one is doing...

fichtner:
I don't know, that's all I'm saying.

lightsquid is not a default so we can get away with using the new timestamp format as a default, but then we need to provide a fallback option for the people who use other tools that expect the "standard" log format.

Stephanowicz:
A combobox with different formatting options would be nice^^

fichtner:
[...]

PR #3326, opened 2019-03-13
Squid log webUI in readable date format #1831

nhirokinet:
This is an idea for #1831.

### DIFF
--- a/src/www/diag_logs_common.inc
+++ b/src/www/diag_logs_common.inc
@@ -33,6 +33,39 @@ if (!isset($logsplit)) {
     $logsplit = 3;
 }
 
+function format_unixtime($unixtime_before_point, $unixtime_after_point) {
+    global $config;
+
+    $datetime_instance = new Datetime('@' . $unixtime_before_point);
+
+    switch ($config['system']['language']) {
+        case 'ja_JP':
+        case 'zh_CN':
+            return $datetime_instance->format('Y/m/d H:i:s') . $unixtime_after_point;
+
+        default:
+            return $datetime_instance->format('D, d M Y H:i:s') . $unixtime_after_point . $datetime_instance->format(' O');
+    }
+}
+
+function format_date_time($date_time) {
+    global $log_datetime_format;
+    if (isset($log_datetime_format) && $log_datetime_format === 'unix') {
+        if (strpos($date_time, '.') !== FALSE) {
+            $unixtime_before_point = substr($date_time, 0, strpos($date_time, '.'));
+            $unixtime_after_point  = substr($date_time, strpos($date_time, '.'));
+        } else {
+            $unixtime_before_point = $date_time;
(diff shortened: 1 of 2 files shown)


src/www/diag_logs_common.inc
```diff
@@ -29,13 +29,43 @@
     POSSIBILITY OF SUCH DAMAGE.
 */
 
+require_once('guiconfig.inc');
```
fichtner: this shouldn't be necessary

src/www/diag_logs_proxy.php
```diff
@@ -7,7 +7,12 @@
 
 $logfile = "/var/log/squid/{$type}.log";
 $logclog = false;
+
 $logsplit = 2;
+if ($type === 'access' || $type === 'store') {
+    $logsplit = 1;
+    $log_datetime_format = 'unix';
```
fichtner: you should declare that variable above so that it is always declared

nhirokinet:
I fixed them.
As I remember I added `require_once('guiconfig.inc');` to load correct timezone, but it seems just my mistake.

fichtner:
I'm ok to try this if @AdSchellevis is also ok with it

AdSchellevis:
@fichtner I haven't tried it out yet, but in principle I'm not agains reformatting unix timestamps, this just looks a bit too overly complicated for my taste. I also think that the default format should match the other log files (YYYY/MM/DD HH24:MI:SS if I'm not mistaken)

nhirokinet:
@AdSchellevis This is just an opinion by just a user, I felt unixtime was better way to store here:

- In #1831 , non-default log format of squid was concern for third-party log analyzer (first reason was this).
- With lack of abstract expression of local time in PHP (and I think some other languages),  I felt unixtime is better storing format, and WebUI can immediately reformat it (and I think is nice if it's according according to users' current preferences).
- Unixtime does not lack an information about DST (but I live in a country without DST, I actually don't know how they treat DST) or timezone.
- Default log format seems to be "Apr 22 23:19:22", whose `$logsplit` should be [3](https://github.com/opnsense/core/blob/715ad3f2534c58b9ec78cdcb062ba90322539188/src/www/diag_logs_common.inc#L33). Squid cache log seems to have `YYYY/MM/DD HH24:MI:SS` format, but it [overrides](https://github.com/opnsense/core/blob/715ad3f2534c58b9ec78cdcb062ba90322539188/src/www/diag_logs_proxy.php#L10) the default.

AdSchellevis:
@nhirokinet you're right about the format, my mistake. The code itself could be simplified, but I don't mind adding this first, since the impact is low and there's not a lot of code involved.

fichtner:
Merged, thanks. @AdSchellevis I am not sure which simplifications you mean. Feel free to pick this up when you have time :)

PR #3333, opened 2019-03-15
Auth/LDAP: More reliable load and syncup for LDAP attributes

indrajitr:
* Load user email when available
* Update descr and email if available when updating existing local user
* Add all the necessary search attributes while making LDAP query

NB: This is a followup of opnsense#3322

### DIFF
--- a/src/www/system_usermanager.php
+++ b/src/www/system_usermanager.php
@@ -626,7 +626,7 @@ $( document ).ready(function() {
                   <tr>
                     <td><a id="help_for_fullname" href="#" class="showhelp"><i class="fa fa-info-circle"></i></a> <?=gettext("Full name");?></td>
                     <td>
-                      <input name="descr" type="text" value="<?=$pconfig['descr'];?>" <?= $pconfig['scope'] == "system" || !empty($pconfig['user_dn']) ? "readonly=\"readonly\"" : "";?> />
+                      <input name="descr" type="text" value="<?=$pconfig['descr'];?>" <?= $pconfig['scope'] == "system" ? "readonly=\"readonly\"" : "";?> />
                       <div class="hidden" data-for="help_for_fullname">
                         <?=gettext("User's full name, for your own information only");?>
                       </div>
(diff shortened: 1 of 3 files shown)


src/www/system_usermanager.php
```diff
@@ -635,7 +635,7 @@ function get_user_privdesc(& $user)
                   <tr>
                     <td><a id="help_for_email" href="#" class="showhelp"><i class="fa fa-info-circle"></i></a> <?=gettext("E-Mail");?></td>
                     <td>
-                      <input name="email" type="text" value="<?= $pconfig['email'] ?>" />
+                      <input name="email" type="text" value="<?= $pconfig['email'] ?>" <?= $pconfig['scope'] == "system" || !empty($pconfig['user_dn']) ? "readonly=\"readonly\"" : "";?> />
```
AdSchellevis: I don't think we should lock the email, if the other end doesn't return it, we likely get support questions about this.

src/www/system_usermanager_import_ldap.php
```diff
@@ -30,14 +30,21 @@
...
-function add_local_user($username, $userdn, $userfullname)
+function add_local_user($username, $userdn, $userfullname, $useremail)
 {
     global $config;
 
     foreach ($config['system']['user'] as &$user) {
         if ($user['name'] == $username && $user['name'] != 'root') {
           // link local user to remote server by updating user_dn
           $user['user_dn'] = $userdn;
+          // also update descr and email if available
```
AdSchellevis: please remove this block `local_user_set()` doesn't use these parameters.

src/www/system_usermanager.php
indrajitr: Agreed! We should probably have the same treatment for `descr` (fullname) as well, shouldn't we?

AdSchellevis: that's debatable, it should be kept locked for system, but for ldap we might strip it off indeed

indrajitr: Sure, so something like this works for `descr` (just removing check for empty `user_dn`)?

```
<input name="descr" type="text" value="<?=$pconfig['descr'];?>" <?= $pconfig['scope'] == "system" ? "readonly=\"readonly\"" : "";?> />
```

indrajitr:
@AdSchellevis, all set. Thank you!

AdSchellevis:
@indrajitr looks good, thanks!

PR #3336, opened 2019-03-15
Monit, remove __items and fix migration issue

### LINKED ISSUE #3334: monit: M1_0_8.php uses __items
Can't see any other core component using it any longer so it should probably be converted.

@fbrendel can you take this? @AdSchellevis can help worst case

fbrendel:
This fixes #3334.

@AdSchellevis I had to revert your commit eb280f6 partially because it brakes the migration.
The BaseModel::runMigrations() function makes a serializeToConfig() after BaseModelMigration::run().
But run() fills the test->type with the default 'Custom' which was not allowed.

### DIFF
--- a/src/opnsense/mvc/app/models/OPNsense/Monit/Migrations/M1_0_8.php
+++ b/src/opnsense/mvc/app/models/OPNsense/Monit/Migrations/M1_0_8.php
@@ -35,7 +35,7 @@ class M1_0_8 extends BaseModelMigration
 {
     public function post($model)
     {
-        foreach ($model->getNodeByReference('test')->__items as $test) {
+        foreach ($model->test->iterateItems() as $test) {
             $test->type = $model->getTestType($test->condition->getNodeData());
         }
         // validation will fail because we want to change the type of tests linked to services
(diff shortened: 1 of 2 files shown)


AdSchellevis:
@fbrendel ok, it's still a bit too obscure what's happing here for my taste, but let's first make sure it's functional again.

fbrendel:
The function BaseModel::runMigrations() calls BaseModelMigration::run() which adds (via BaseModelMigration::checkDefaults()) default test types ('Custom') to the model. Then it calls Monit::serializeToConfig() (Monit is a BaseModel). https://github.com/opnsense/core/blob/9746ebf8f19046cca1a01d4f3277ec7bf02e0595/src/opnsense/mvc/app/models/OPNsense/Base/BaseModel.php#L649

which in turn calls Monit::performValidation(). Here it throws an exception because adding "type" fields to tests which are linked to services was not allowed in Monit::performValidation().
Now it is still not allowed unless it is a 'Custom' type.
https://github.com/opnsense/core/blob/9746ebf8f19046cca1a01d4f3277ec7bf02e0595/src/opnsense/mvc/app/models/OPNsense/Monit/Monit.php#L171-L178

The real problem is that the migration cannot bypass the validation.

PR #3340, opened 2019-03-16
certmanager: Avoid double encoding cert name

indrajitr:
We do not need to wrap `$cert['descr']` with `htmlspecialchars` since everything in `$cert` is already html encoded via `legacy_html_escape_form_data`.

Because of double wrapping, Let's Encrypt issued certificate have an awkward looking name as:
**`my.hostname.com (Let&#039;s Encrypt)`**

### DIFF
--- a/src/www/system_certmanager.php
+++ b/src/www/system_certmanager.php
@@ -1942,7 +1942,7 @@ $( document ).ready(function() {
 <?php
             $i = 0;
             foreach ($a_cert as $cert) :
-                $name = htmlspecialchars($cert['descr']);
+                $name = $cert['descr'];
                 $purpose = null;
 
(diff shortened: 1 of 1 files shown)


indrajitr:
Alternately, we can probably get rid of [these](https://github.com/opnsense/core/blob/a0f55d452954573f3f0ecc6eaf8ee275dc169961/src/www/system_certmanager.php#L777-L778):

```
legacy_html_escape_form_data($a_ca);
legacy_html_escape_form_data($a_cert);
```

It worked in my basic testing but would be too risky without rigorous testing. :)

AdSchellevis:
I'll opt for the generic escaping, which we have now with `legacy_html_escape_form_data()`, thanks for the patch.

Issue #3339, opened 2019-03-15
VPN IPsec Mobile Client: Generating strongswan.conf eap-radius plugin setting from vpn_ipsec_mobile.php instead of Phase1


stumbaumr:
**Is your feature request related to a problem? Please describe.**
Going through my current setup using bespoke include files for strongswan I realised that the information on how to setup the eap-radius plugin in strongswan.conf is already existing in vpn_ipsec_mobile.php. But currently have to set it up using a strongswan.conf include file since I do not create a mobile phase 1.

**Describe the solution you'd like**
Would a PR be acceptable to move the functionality from Tunnel Phase 1 to Mobile Client? Or are there any obstacles?

AdSchellevis:
This issue has been automatically timed-out (after 180 days of inactivity).

For more information about the policies for this repository, 
please read https://github.com/opnsense/core/blob/master/CONTRIBUTING.md for further details.

If someone wants to step up and work on this issue, 
just let us know, so we can reopen the issue and assign an owner to it.

Issue #2833, opened 2018-10-23
Allowing multiple IPSEC Mobile P1 configs


mimugmail:
Reference: https://forum.opnsense.org/index.php?topic=9142.msg44734#msg44734

Today I tried this for a customer. Added a second Mobile P1 via link from forum and set a different authentication mechanism. Only caveat is that I can't user the same network in P2, but this is just a wrong preg validator since it doesn't matter what rightsubnet is when using mobile ipsec.

Last thing is that all users share the same pool, but in principle it's working. What I have setup and tested successfully was a setup mix of IPSEC v1 with PSK+XAuth and a second connection with IKEv2 and EAP-MSCHAPv2 and/or EAP-RADIUS. This is really cool since you can allow all kinds of devices like android with stock VPN since it doesn't support IKEv2 but also Windows7 or Mac (has to be tested yet by my colleague). 

@AdSchellevis Is there a technical reason why only one mobile ipsec is allowed?

mimugmail:
Thanks @AdSchellevis  👍 Can we add a "help wanted" tag? I'll collect some test scenarios, documenting what works and whats not, also to have a reference if there come more FRs from other. Perhaps I can find a moment in post-19.1 where some of you has time to work on this. :)

AdSchellevis:
Finding out what works and what doesn't is probably most of the needed work for options that don't need structural changes, I can probably do the work afterwards depending on the outcome of your testing.

mimugmail:
Mutual PSK + XAuth with ldap as backend, combined with EAP-RADIUS does not work:

strongswan.conf:

```
starter {
    load_warning = no
}

charon {
    # number of worker threads in charon
    threads = 16
    ikesa_table_size = 32
    ikesa_table_segments = 4
    init_limit_half_open = 1000
    ignore_acquire_ts = yes
    syslog {
        daemon {
            ike_name = yes
        }
    }
        cisco_unity = yes

        plugins {
                attr {
                        subnet = 10.0.0.0/8,10.0.0.0/9,10.0.0.0/10
                        split-include = 10.0.0.0/8,10.0.0.0/9,10.0.0.0/10
                }
                xauth-generic {
                        script = /usr/local/etc/inc/plugins.inc.d/ipsec/auth-user.php
                        authcfg = unused
                }
        }
}
```

```
starter {
    load_warning = no
}

charon {
    # number of worker threads in charon
    threads = 16
    ikesa_table_size = 32
    ikesa_table_segments = 4
[...]

PR #3346, opened 2019-03-18
IPsec: Generate strongswan.conf Radius backend configuration from mobile client page

stumbaumr:
Hi,

this is the PR for https://github.com/opnsense/core/issues/3339 .

If there is a mobile phase 1 containing a EAP-Radius configuration the system behaves as before.

If there is no mobile phase 1 but a Radius server configured in the VPN-IPsec-Mobile client page the radius plugin config will be configured from there.

If the plan is to have - at some point in the future - more than only one mobile phase this is the first step  (remove the configuration of the per-user based authentication backends away from the tunnel config into the Mobile client config).

### DIFF
--- a/src/etc/inc/plugins.inc.d/ipsec.inc
+++ b/src/etc/inc/plugins.inc.d/ipsec.inc
@@ -963,42 +963,23 @@ function ipsec_configure_do($verbose = false, $interface = '')
                 $strongswanTree['charon']['plugins']['attr']['28679'] = $a_client['pfs_group'];
             }
 
-            $disable_xauth = false;
+            $radius_enabled = false;
             foreach ($a_phase1 as $ph1ent) {
                 if (!isset($ph1ent['disabled']) && isset($ph1ent['mobile'])) {
                     if ($ph1ent['authentication_method'] == "eap-radius") {
-                        $disable_xauth = true; // disable Xauth when radius is used.
-                        $strongswanTree['charon']['plugins']['eap-radius'] = [];
-                        $strongswanTree['charon']['plugins']['eap-radius']['servers'] = [];
-                        $radius_server_num = 1;
-                        $radius_accounting_enabled = false;
-                        foreach (auth_get_authserver_list() as $auth_server) {
-                            if (in_array($auth_server['name'], explode(',', $ph1ent['authservers']))) {
-                                $server = [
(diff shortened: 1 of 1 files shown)


AdSchellevis:
I'm not sure we should do this now, the code underneath already doesn't win a beauty contest adding more logic without functional use probably doesn't make it better.

Supporting more mobile phase1 connections (using different authentication schemes) in the future would certainly be a nice goal, although likely very difficult to reach with how things work now. 
If possible, we should start with figuring out how the different currently supported authentication options should function when there's more than one (for the old xauth this was not possible as far as I know, with pam in play there might be more options).

It's quite a project to restructure the current code base for ipsec to something more maintainable, not sure when we're going to do so.
There are some ideas, but not concrete yet. Questions like this popup from time to time (https://github.com/opnsense/core/issues/2833), but investments needed to make it happen are quite substantial.

stumbaumr:
So what is your usual approach when doing bigger overhauls like this? Would you accept a PR which fixes everything in one go? As far as I understand @mimugmail is based in Munich, too. If we organise a hackfest that replaces the whole non-PSK/RSA authentications - would the outcome of that be acceptable?

Or do you prefer smaller code changes in a step-by-step fashion? Like the above?

mimugmail:
Please note that I'm not a core member and will never be, as I'm already in partner state.

AdSchellevis:
I might do some work later on this topic, wrap logic in easier to understand classes like we did for the firewall code. Small commits functionally adding something are still being considered. 

Partners help fund the project, which we also take into consideration when prioritising work. Thanks @mimugmail !

stumbaumr:
So what I understand here is "lean back, wait for me to fix it". Fine.

AdSchellevis:
ok, let me rephrase, we do accept functional changes to the current code if it fixes actual issues or adds relevant features. 
We just don't accept partial refactoring pr's as eventually we probably will redo the whole module since it's difficult to maintain as it is and needs a solid plan before doing so. 

Things that help getting there in this case are investigating how current authentication scenarios would fit in proposed new ones (how do they look now, what should be changed, would that be possible within configuration parameters that are available now, etc, etc). Plan->Do->Check->Act

At the end of the day, all kinds of setups are supported with our system, we would like to avoid breakage unless absolutely necessary.

PR #3358, opened 2019-03-21
www/squid: add proxy sortable support for proxy lists (PAC)

### LINKED ISSUE #2979: Proxy Auto-Config rules order of proxy entries
I created two proxy entries and a direct one.

When I use the rules to combine matches and proxies, my proxy order gets messed up when saving the rule.

I want: proxy1, proxy2, direct

When I save, it gets reversed to direct, proxy2, proxy1. The direct "proxy" should act as fall-back. The order is not alphabetical. I tried to name them A, B, C and order is still, C, B, A. I even added them in reverse order, but sort criteria must be something else, even not ip address because proxy1 is lower ip than proxy2.

The resulting wpad.dat creates the saved order which is wrong. The order should be kept the way you add the proxies.

fabianfrz:
closes #2979


fichtner:
Merged, thanks <3

PR #3361, opened 2019-03-22
opnsense/core: automatically remove trailing slash

### LINKED ISSUE #2647: System > Configuration > Backups > Nextcloud URL
Hello!

I'd previously had issues trying to get nextcloud backups configured but kept getting a generic "Communication Error" when testing.  I finally spotted in the logs that I had an extra trailing forward slash in my URL which prevented connectivity.  May I suggest adding "rtrim" where appropriate to find and trim any trailing slashes?  If I remember I'm happy to look at it and do a PR with proposed changes.

fabianfrz:
removes also spaces for more safety.

closes #2647

### DIFF
--- a/src/opnsense/mvc/app/models/OPNsense/Backup/NextcloudSettings.xml
+++ b/src/opnsense/mvc/app/models/OPNsense/Backup/NextcloudSettings.xml
@@ -9,7 +9,7 @@
         </enabled>
         <url type="TextField">
             <Required>N</Required>
-            <mask>/^https?:\/\/.+[^\/]$/</mask>
+            <mask>/^https?:\/\/.*[^\/]$/</mask>
             <ValidationMessage>The url must be valid without a trailing slash. For example: https://nextcloud.example.com or https://example.com/nextcloud</ValidationMessage>
             <Constraints>
                 <check001>
--- a/src/opnsense/mvc/app/models/OPNsense/Backup/NextcloudSettings.xml
+++ b/src/opnsense/mvc/app/models/OPNsense/Backup/NextcloudSettings.xml
@@ -9,7 +9,7 @@
         </enabled>
         <url type="TextField">
             <Required>N</Required>
-            <mask>/^https?:\/\/.*[^\/]$/</mask>
+            <mask>/^https?:\/\/.+[^\/]$/</mask>
             <ValidationMessage>The url must be valid without a trailing slash. For example: https://nextcloud.example.com or https://example.com/nextcloud</ValidationMessage>
             <Constraints>
                 <check001>
--- a/src/opnsense/mvc/app/models/OPNsense/Backup/NextcloudSettings.xml
+++ b/src/opnsense/mvc/app/models/OPNsense/Backup/NextcloudSettings.xml
@@ -9,6 +9,8 @@
         </enabled>
         <url type="TextField">
             <Required>N</Required>
+            <mask>/^https?:\/\/.*[^\/]$/</mask>
+            <ValidationMessage>The url must be valid without a trailing slash. For example: https://nextcloud.example.com or https://example.com/nextcloud</ValidationMessage>
             <Constraints>
                 <check001>
                     <ValidationMessage>An URL for the Nextcloud server must be set.</ValidationMessage>
--- a/src/opnsense/mvc/app/library/OPNsense/Backup/Nextcloud.php
+++ b/src/opnsense/mvc/app/library/OPNsense/Backup/Nextcloud.php
@@ -111,9 +111,6 @@ class Nextcloud extends Base implements IBackupProvider
     public function setConfiguration($conf)
     {
         $nextcloud = new NextcloudSettings();
-        if (array_key_exists('url', $conf) && !empty($conf['url'])) {
-            $conf['url'] = rtrim(trim($conf['url']), '/');
-        }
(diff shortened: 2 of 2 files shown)


fichtner:
Wouldn't it be easier to add this to the validation?

fabianfrz:
@fichtner would be an option too - in this case it would autocorrect which is probably also a benefit for the user as he does not have to change it and save it again.

fichtner:
It's obscure. I'd rather want to go for a hard fail with a clear validation message. the damage is already done so to speak. We can make a not in the release notes...

fabianfrz:
@fichtner, ok then I will change it.

fabianfrz:
@fichtner this should throw an error in case the user does input something stupid

src/opnsense/mvc/app/models/OPNsense/Backup/NextcloudSettings.xml
```diff
@@ -9,6 +9,8 @@
         </enabled>
         <url type="TextField">
             <Required>N</Required>
+            <mask>/^https?:\/\/.*[^\/]$/</mask>
```
fichtner:
```suggestion
            <mask>/^https?:\/\/+*[^\/]$/</mask>
```

"https://" was valid :)

fichtner: hmm ignore for now, it's a bit late

fichtner: should this support sub-directories or not?

fabianfrz: yes, nextcloud is an PHP application which can be put in any directory but it is usually put into the root directory.

fabianfrz: your change is perfectly valid, should have been a + for sure.

src/opnsense/mvc/app/models/OPNsense/Backup/NextcloudSettings.xml
```diff
@@ -9,6 +9,8 @@
         </enabled>
         <url type="TextField">
             <Required>N</Required>
+            <mask>/^https?:\/\/.+[^\/]$/</mask>
```
fichtner:
```suggestion
            <mask>/^https?:\/\/.*[^\/]$/</mask>
```

Sorry for backtracking this, your first one is correct, the trailing [] is already a required atom.

fabianfrz: applied

fichtner:
Merged, thanks!

src/opnsense/mvc/app/models/OPNsense/Backup/NextcloudSettings.xml
opnsenseuser: Incidentally, I use cloudnow. It is an Austrian company. and I post here only my comment because it would be another alternative and it is synonymous of me for months running smoothly. maybe this would be another alternative?

grafik

opnsenseuser: https://www.cloudfiles.app/

opnsenseuser: https://www.cloudfiles.app/sicherheit/

fabianfrz: @opnsenseuser What do you want to tell me with your screenshot? It is very unlikely that the validation fails with your URL.

opnsenseuser: sorry, I have mispronounced myself. I just wanted to say that cloudnow settings work here as well.

opnsenseuser: I'm sure that other cloud services would also be configurable via the input mask. So would not it be advantageous to replace the term nextcloud with a generic term?

opnsenseuser: @fabianfrz I just wanted to make an improvement proposal. I imagine a drop down box, where you can select different cloud services. eg nextcloud, owncloud, cloudnow, etc. the data are then eg already pre-filled but still modifiable. I imagine that as in the dyndns settings Plugin from opnsense. what do you mean?

fabianfrz: @opnsenseuser nextcloud and owncloud are usually on premise solutions, so you have to explicitly configure the URL or the hostname anyway. In my case I have my own nextcloud on my own domain which I used for testing this module against. The reason why it is called nextcloud is that I do not test it against an owncloud instance (even if it very likely works because the same WebDAV library is working under the hood).

SAAS services are different. If it works, that's fine but I don't want to test against all of them (never ending story) and I currently cannot add anything else then the input fields (for example no JS to create something like a template directly in my file but that should be pluggable). Please note that this must follow the paths of nextcloud to work. Other WebDAV variants will not work as the nextcloud paths and responses are hardcoded in the file.

PR #3369, opened 2019-03-25
diag_confbak.php to show correct timestamp

nhirokinet:
When I tried to restore old setting, displayed time on alert was equivalent to unixtime 0. This fix works for me, and this seems to be consistent with the delete line just below modified line.


fichtner:
Nice catch, thanks!

PR #3372, opened 2019-03-26
ipsec: added ikev2 dh group 31 / curve 25519

mmaction:
curve 25519 is supported in strongswan 5.7.1 and newer

### DIFF
--- a/src/www/vpn_ipsec_mobile.php
+++ b/src/www/vpn_ipsec_mobile.php
@@ -524,6 +524,7 @@ endfor; ?>
                             28 => '28 (Brainpool EC 256 bits)',
                             29 => '29 (Brainpool EC 384 bits)',
                             30 => '30 (Brainpool EC 512 bits)',
+                            31 => '31 (Elliptic Curve 25519)',
                         );
                         foreach ($p2_dhgroups as $keygroup => $keygroupname): ?>
                           <option value="<?=$keygroup;
--- a/src/www/vpn_ipsec_phase1.php
+++ b/src/www/vpn_ipsec_phase1.php
@@ -984,6 +984,7 @@ endforeach; ?>
                            28 => '28 (Brainpool EC 256 bits)',
                            29 => '29 (Brainpool EC 384 bits)',
                            30 => '30 (Brainpool EC 512 bits)',
+                           31 => '31 (Elliptic Curve 25519)',
                       );
                       foreach ($p1_dhgroups as $keygroup => $keygroupname):
 ?>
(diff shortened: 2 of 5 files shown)


src/www/vpn_ipsec_mobile.php
```diff
@@ -524,6 +524,7 @@ function print_legacy_box($msg, $name, $value)
                             28 => '28 (Brainpool EC 256 bits)',
                             29 => '29 (Brainpool EC 384 bits)',
                             30 => '30 (Brainpool EC 512 bits)',
+                            31 => '31 (Elliptic Curve 25519)',
```
fabianfrz:
```suggestion
                            31 => '31 (Curve25519)',
```

See https://cr.yp.to/ecdh.html - the page of the developer of this curve -> it is spelled this way.

src/www/vpn_ipsec_phase1.php
```diff
@@ -984,6 +984,7 @@ function ipsec_ikeid_next() {
                            28 => '28 (Brainpool EC 256 bits)',
                            29 => '29 (Brainpool EC 384 bits)',
                            30 => '30 (Brainpool EC 512 bits)',
+                           31 => '31 (Elliptic Curve 25519)',
```
fabianfrz:
```suggestion
                           31 => '31 (Curve25519)',
```

src/www/vpn_ipsec_phase2.php
```diff
@@ -744,6 +744,7 @@ function getIndexByUniqueId($uniqid)
                         28 => '28 (Brainpool EC 256 bits)',
                         29 => '29 (Brainpool EC 384 bits)',
                         30 => '30 (Brainpool EC 512 bits)',
+	                31 => '31 (Elliptic Curve 25519)',
```
fabianfrz:
```suggestion
	                    31 => '31 (Curve25519)',
```

src/www/vpn_ipsec_mobile.php
fichtner: please use the https://wiki.strongswan.org/projects/strongswan/wiki/IKEv2CipherSuites doc to refer to, otherwise it's confusing

fabianfrz: @fichtner ok, then keep it as is since the official documentation of strongswan differs from the official documentation of the cipher.

mmaction:
Hello Fabian and Franco,

thanks for the prompt response.

re: executable bit set
i did some modifications under Windows and copied the files to Linux;
i was not aware of the x bits set.

re: naming "Elliptic Curve 25519" vs. "Curve25519"
according to Fabian it's ok to use "Elliptic Curve 25519"
(as mentioned in the strongswan IKEv2 cipher suites list)

do you expect me to make further modifications or would you rather
take over the current modifications in the 5 files in a future release?

when would the changes appear in an official release?

regards
Peter Stehlin

fichtner:
merged, thanks!

fichtner:
exec bits and indent changes via 1d749b6408ab

I think this can be in 19.1.5 planned for next week.

PR #3388, opened 2019-04-05
CP: support autologin

fabianfrz:
@fichtner this is the code change for this feature: https://forum.opnsense.org/index.php?topic=12298.msg56525#msg56525


AdSchellevis:
I'm ok with moving the login sequence to a function, so people can easily use the url parameters, but we shouldn't allow get request type logins by default in my opinion (the docs would be a better place to describe how one could achieve this)

AdSchellevis:
timeout

Issue #1675, opened 2017-06-08
Transparent proxy blocking Google


kojo1984:
Hi,

I have configured OPNSense several times, using official docs.

I have configured Transparent PROXY with SSL inspection, deployed certificate to all PCs.

I have problem that sometimes, for some users, Google pages are blocked (translate.google.com, www.google.ba).

In Proxy log, I found:

2017/06/08 09:55:55 kid1| 	Error negotiating SSL connection on FD 176: (54) Connection reset by peer
2017/06/08 09:55:55 kid1| 	Error negotiating SSL connection on FD 176: error:14076102:SSL routines:SSL23_GET_CLIENT_HELLO:unsupported protocol (1/-1)
2017/06/08 09:55:55 kid1| 	SECURITY ALERT: on URL: translate.google.com:443
2017/06/08 09:55:55 kid1| 	SECURITY ALERT: Host header forgery detected on local=172.217.18.78:443 remote=192.168.1.93:56166 FD 176 flags=33 (local IP does not match any domain IP)
2017/06/08 09:55:55 kid1| 	SECURITY ALERT: on URL: translate.google.com:443
2017/06/08 09:55:55 kid1| 	SECURITY ALERT: Host header forgery detected on local=172.217.18.78:443 remote=192.168.1.93:56165 FD 176 flags=33 (local IP does not match any domain IP)
2017/06/08 09:55:55 kid1| 	SECURITY ALERT: on URL: translate.google.com:443
2017/06/08 09:55:55 kid1| 	SECURITY ALERT: Host header forgery detected on local=172.217.18.78:443 remote=192.168.1.93:56164 FD 176 flags=33 (local IP does not match any domain IP)
2017/06/08 09:55:51 kid1| 	SECURITY ALERT: on URL: translate.google.com:443
2017/06/08 09:55:51 kid1| 	SECURITY ALERT: Host header forgery detected on local=172.217.18.78:443 remote=192.168.1.93:56162 FD 55 flags=33 (local IP does not match any domain IP)
2017/06/08 09:55:51 kid1| 	SECURITY ALERT: on URL: translate.google.com:443
2017/06/08 09:55:51 kid1| 	SECURITY ALERT: Host header forgery detected on local=172.217.18.78:443 remote=192.168.1.93:56161 FD 55 flags=33 (local IP does not match any domain IP)
[...]

PR #3447, opened 2019-04-28
Proxy: ssl_crtd was removed with Squid 4.X

mimugmail:
On current devel mode you get an error while starting squid.

2019/04/28 18:46:56 kid1| Too few /usr/local/libexec/squid/ssl_crtd -s /var/squid/ssl_crtd -M 4MB processes are running (need 1/5)
2019/04/28 18:46:56 kid1|   Took 0.00 seconds (  0.00 entries/sec).

Filename is now: 

security_file_certgen

Syntax is the same:
https://forum.lissyara.su/freebsd-f8/freebsd-11-squid-4-5-prozrachnoe-proksirovanie-s-p-t45437.html

Should stay in master as long as we are on Squid 3.5

### DIFF
--- a/src/opnsense/service/templates/OPNsense/Proxy/squid.conf
+++ b/src/opnsense/service/templates/OPNsense/Proxy/squid.conf
@@ -52,7 +52,7 @@
 
 {% if helpers.exists('OPNsense.proxy.forward.sslbump') and OPNsense.proxy.forward.sslbump == '1' %}
 # setup ssl re-cert
-sslcrtd_program /usr/local/libexec/squid/ssl_crtd -s /var/squid/ssl_crtd -M {{ OPNsense.proxy.forward.ssl_crtd_storage_max_size|default('4') }}MB
+sslcrtd_program /usr/local/libexec/squid/security_file_certgen -s /var/squid/ssl_crtd -M {{ OPNsense.proxy.forward.ssl_crtd_storage_max_size|default('4') }}MB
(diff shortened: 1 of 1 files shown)


fichtner:
Nice catch, thanks! squid 4 may still be broken for transparent proxy. I haven't tested in a while though.

fichtner:
(broken as in upstream broken, not config settings)

mimugmail:
Switched my home OPN to devel and it seems you're right, GitHub is broken ;)

Franco Fichtner <notifications@github.com> schrieb am So., 28. Apr. 2019,
19:06:

fichtner:
oh noes! I'm not sure, maybe we will have to revert to squid3 for 19.7, but contacting upstream may be the better thing to do in the meantime...

fabianfrz:
@fichtner I had a look at it - it seems like squid is not requesting a certificate because I don't get a line via nc or tee as certd program.

mimugmail:
It works now in SNI log mode, only Avatars from GH missing :)

mimugmail:
Seems I'm hitting this one:
https://github.com/opnsense/core/issues/1675

I'll investigate the next days/weeks at home with multiple devices

fichtner:
There's a reset button for it, we will mention it in the release notes for 19.7... it slipped through 19.7-RC1, sorry

mimugmail:
And sorry for reviewing a merged PR .. I'm on mobile only today :)

fichtner:
no worries, nice catch!

Issue #3394, opened 2019-04-08
Feature request: Squid - SSL-Filtering per Interface


AndyX90:
**Important notices**
Before you add a new report, we ask you kindly to acknowledge the following:

[+] I have read the contributing guide lines at https://github.com/opnsense/core/blob/master/CONTRIBUTING.md

[+] I have searched the existing issues and I'm convinced that mine is new.

**Is your feature request related to a problem? Please describe.**
The SSL-Filtering in Squid can only be done for the whole Machine.

**Describe the solution you'd like**
I would like to have the option to do ssl-bump on LAN for my Windows-Clients and SNI-Filter only on WiFi-Interfaces for Mobile-Devices.

**Describe alternatives you've considered**
An alternative would be to provide a possibility via GUI for multiple Squid-Instances.

**Additional context**
-

AdSchellevis:
This issue has been automatically timed-out (after 180 days of inactivity).

For more information about the policies for this repository, 
please read https://github.com/opnsense/core/blob/master/CONTRIBUTING.md for further details.

If someone wants to step up and work on this issue, 
just let us know, so we can reopen the issue and assign an owner to it.

PR #3455, opened 2019-05-01
Webproxy: Make SNI selectable

AndyX90:
Such a feature would be really helpful.
I need it to make Squid bump my LAN Network (with Windows-Clients) and to splice the Transparent-Proxy-Interfaces with Mobile-Clients (Phones/Tablets) where Certificates cannot be imported easily.
This would resolve my [FR 3394](https://github.com/opnsense/core/issues/3394).
But I don't know if it is the right way to do it.
It would be great if someone could review this.

### DIFF
--- a/src/opnsense/mvc/app/models/OPNsense/Proxy/Proxy.xml
+++ b/src/opnsense/mvc/app/models/OPNsense/Proxy/Proxy.xml
@@ -1,6 +1,6 @@
 <model>
     <mount>//OPNsense/proxy</mount>
-    <version>1.0.3</version>
+    <version>1.0.4</version>
     <description>
         (squid) proxy settings
     </description>
--- a/src/opnsense/mvc/app/models/OPNsense/Proxy/Proxy.xml
+++ b/src/opnsense/mvc/app/models/OPNsense/Proxy/Proxy.xml
@@ -272,9 +272,15 @@
                     </check001>
                 </Constraints>
             </sslbump>
-            <sslurlonly type="BooleanField">
-                <default>0</default>
+            <sslurlonly type="OptionField">
+                <default>off</default>
                 <Required>Y</Required>
+                <OptionValues>
+                    <off>Off</off>
+                    <explicit>Explicit</explicit>
+                    <transparent>Transparent</transparent>
+                    <both>Both (transparent+explicit)</both>
+                </OptionValues>
                 <Constraints>
                     <check001>
                         <reference>sslbump.check001</reference>
--- a/src/opnsense/service/templates/OPNsense/Proxy/squid.conf
+++ b/src/opnsense/service/templates/OPNsense/Proxy/squid.conf
@@ -19,11 +19,11 @@
 
 {% if helpers.exists('OPNsense.proxy.forward.transparentMode') and OPNsense.proxy.forward.transparentMode == '1' %}
 # Setup transparent mode listeners on loopback interfaces
-{{ listener_config('127.0.0.1', OPNsense.proxy.forward.port, 'intercept') }}
-{{ listener_config('[::1]', OPNsense.proxy.forward.port, 'intercept') }}
+{{ listener_config('127.0.0.1', OPNsense.proxy.forward.port, 'name=transparenthttpv4 intercept') }}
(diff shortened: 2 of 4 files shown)


AndyX90:
With the existing solution you can only specify the same behaviour for explicit and transparent Proxy. I try to find a Solution for the Scenario Bump in explicit Mode and splice in transparent Mode.

src/opnsense/mvc/app/models/OPNsense/Proxy/Proxy.xml
```diff
@@ -338,9 +338,15 @@
...
-                <default>0</default>
+            <transparentMode type="OptionField">
+                <default>off</default>
                 <Required>Y</Required>
+                <BlankDesc>Default</BlankDesc>
+                <OptionValues>
+                    <off>Turn off transparent Mode</off>
+                    <bump>Do SSL-Bump in transparent Mode</bump>
+                    <splice>Log SNI only in transparent Mode</splice>
+                </OptionValues>
```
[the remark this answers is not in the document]
fichtner: we should try to avoid the model bump or at least "smart" reuse of the same setting plus migration altogether. changing existing option values and their behaviour at the same time is a recipe for disaster (config restore on older versions, testing and migration time spent in the hopes it'll not break which it ultimately will)

src/opnsense/service/templates/OPNsense/Proxy/squid.conf
```diff
@@ -62,6 +62,9 @@ acl bump_step1 at_step SslBump1
 acl bump_step2 at_step SslBump2
 acl bump_step3 at_step SslBump3
 acl bump_nobumpsites ssl::server_name "/usr/local/etc/squid/nobumpsites.acl"
+{%     if helpers.exists('OPNsense.proxy.forward.sslbump') and OPNsense.proxy.forward.sslbump == '1' and OPNsense.proxy.forward.transparentMode == 'splice' %}
+acl bump_nobumpsites localip 127.0.0.1/255.255.255.255
```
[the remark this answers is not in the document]
fichtner: yep, it's 127/8 indeed although in practice /24 is used most frequently

fichtner:
my preference would be a second checkbox that handles bump vs. splice

fichtner:
or a select with the current default injected and the new optional setting for configurable use

AndyX90:
Btw. for matching on our configured loopback only, the /32 would have been also ok.
But what about ::1?

fichtner:
::1 isn't working as far as I know, but the mask would be ::1/128 -- there's no network allocated to it.

AndyX90:
Okay, the acl which puts the squid-listener of the loopback interface to the no bump sites won't work. Don't know why. 
Maybe someone has another idea how to solve this?

AndyX90:
Hey guys, i had time to look at the problem again and went away from localip thing and did it via port-naming.
I have tested the transparent Proxy with clamav and remote blacklist and it is working!
Maybe someone can test it and verify?
Thanks!

mimugmail:
Doesnt this need a migration script since a already set value of 1 is not handled?

src/opnsense/mvc/app/models/OPNsense/Proxy/Migrations/M1_0_0.php
```diff
@@ -34,4 +34,11 @@
 
 class M1_0_0 extends BaseModelMigration
```
AdSchellevis: @AndyX90 you need to increase the version number and add a matching migration, I recently added some documentation about the subject here https://docs.opnsense.org/development/frontend/models_migrations.html. (people already have 1.0.0 or later installed, so this migration likely won't execute)

AndyX90: I'm sorry, didn't understand that before. I hope this is okay now?

AdSchellevis: @AndyX90 no problem at all, the documentation wasn't in place until this week as well.

AdSchellevis:
timeout

PR #3484, opened 2019-05-15
added failover_split parameter to configure loadbalancing for DHCPD failover

clystron:
While the hardcoded default split of 128 is a perfectly reasonable value there are scenarios where it may be preferrable to only have one peer (eg the primary) give out leases as long as both peers are running. The split-parameter allows to configure what percentage of clients should be handled by either peer in normal mode. It would probably also be an option to implement this as a choice and only allow values of 0, 128 and 256.

there is still an issue with handling of 0 as non-default in this code, I am not sure how to achieve this.
If someone finds the time to review this that would be great, its my first pull-request so I hope I did not screw something up...

### DIFF
--- a/src/etc/inc/plugins.inc.d/dhcpd.inc
+++ b/src/etc/inc/plugins.inc.d/dhcpd.inc
@@ -623,7 +623,11 @@ EOD;
                 $my_port = "519";
                 $peer_port = "520";
                 $type = "primary";
-                $dhcpdconf_pri  = "split 128;\n";
+                if (!empty($dhcpifconf['failover_split'])) {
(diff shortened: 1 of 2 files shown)


AdSchellevis:
@clystron thanks, this https://github.com/opnsense/core/commit/7a15556bad9925cf24e31f9ae885da6a919080cf should fix the 0 handling.

clystron:
Great, thanks!
I should be able to test this using opnsense-patch, right?

AdSchellevis:
should work if O'm not mistaken, the master branch is pretty aligned for these files.

PR #3490, opened 2019-05-19
WIP: JWT crypto code

fabianfrz:
this is not specific for the nginx plugin so it is better placed here.

### DIFF
--- a/src/opnsense/mvc/app/controllers/OPNsense/Base/ApiControllerBase.php
+++ b/src/opnsense/mvc/app/controllers/OPNsense/Base/ApiControllerBase.php
@@ -28,6 +28,8 @@
  */
 namespace OPNsense\Base;
 
+use OPNsense\Auth\JWT\TokenParserFactory;
+use OPNsense\Auth\JWT\WebUIKeyStore;
 use OPNsense\Core\ACL;
 use OPNsense\Auth\AuthenticationFactory;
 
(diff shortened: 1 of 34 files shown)


AdSchellevis:
I'm not sure it should be in core, how does this interact with our authentication framework?

fabianfrz:
Currently not integrated but can theoretically replace be added in addition / replacement to Basic auth.

The header looks like:

```
Authorization: Bearer JWTTOKEN
```

The format of the Token is:

```
TYPE.CLAIMS.SIGNATURE
```

https://jwt.io/

The advantage is, that this tokens act a bit like RSA client certificates. They can have a valid time (from and / or to), and be created for a specif system. It is a way to pass the authentication state for OAUTH2.


In general this is an authentication technique intended for short and long lived tokens in federated environments (SSO).

For example the claims can look like this for usage in core:

```
{
  "typ": "JWT",    // hardcoded requirement
  "alg": "HS256" // signature is HMAC using SHA256
}
{
  "iss": "hostname of opnsense", // issuer - ca reference in case of X509 certificates
  "sub": "root",                             // subject should be the username or id of the user on which the token is valid
  "nbe": 1527226168,                  // not before a specific date
  "exp": 1558762209,                  // expires
  "jti": "jewjto43jeltej3ojewtjow"    // unique string or UUID - can be used to blacklist a token
}
```

AdSchellevis:
ok, we'll see how this pins out before merging. As soon as you have a working solution which integrates with core just let us know.

fabianfrz:
@AdSchellevis you may give it a try or should we merge master into this branch first?

AdSchellevis:
@fabianfrz I'm not going to merge this without documentation, how does this interact with our authentication framework now?

Another question I have is if we want jwt as authentication option, why don't we use one of the existing libraries they reference on their site? Personally I don't think we want to maintain such a library ourselves. 

If this is something you need specifically for your nginx plugin, we could see if we could find a way to detach this from core, I'm not a huge fan of pulling large blobs of authentication code, which needs quite some time to review and test on our end.

fabianfrz:
@AdSchellevis it should just act the same way as the Basic authentication does It gets called the same way and it calls the same code. The difference is that instead of checking the password against the database, the signature is checked against the public key.

AdSchellevis:
Looking at this again, I don't think we should try to include this. 
I'm not against adding other authentication methods, but we should follow standards as close as possible in my opinion.

It might be an option to investigate how to use claims from other identity providers, but trying to implement a full identity provider lies out side the scope of our core project (Auth0, Keycloak, ... are likely more suited for this purpose).

Issue #3069, opened 2018-12-24
Nextcloud file extensions are not supported by windows


opnsenseuser:
I do not know if exactly this exact problem has already been reported, but the file names are not compatible with windows.

![grafik](https://user-images.githubusercontent.com/34602360/50393504-f0cc1800-0756-11e9-8b31-ce0c6bd484f8.png)

marjohn56:
Speak to Microsoft.. :)

fabianfrz:
It is both: A limitation of Microsoft Windows file names in combination which the Nextcloud Client not handling it.

opnsenseuser:
it´s only up to the ":"

opnsenseuser:
Test on windows:
![grafik](https://user-images.githubusercontent.com/34602360/50405468-6070f000-07b5-11e9-8587-78fde5469597.png)

translated means that 
"A filename can not contain any of the following characters: \ /:*?"< >| "

fabianfrz:
A filename can contain such characters in general (except the slash as that is reserved for the directory separator):

![grafik](https://user-images.githubusercontent.com/3713914/50406515-7d61ef00-07c6-11e9-89f2-c1a37e642704.png)

opnsenseuser:
@fabianfrz but not if you use a windows client and integrate "nextcloud" via network drive. 
So you can´t copy the file to your own computer and then windows reports with the above error message.

fabianfrz:
I explain my point of view:

![grafik](https://user-images.githubusercontent.com/3713914/50545815-cd084680-0c1c-11e9-9da8-7ee8ec938b9b.png)

You have two sides:

* Our side, which is OPNsense uploading the file compliant to the WebDAV protocol. If it answers correctly, the backup is OK. If it answers with an error code, you will see a communication failure in the backup page.
* The other side is the client, who can download files from the nextcloud server using WebDAV or the normal web interface (HTTP). Only the client really knows, on which operating system it is running and it can manage the local restrictions. This is why I see that as the job of the software downloading the file. Assume a vendor developing a operating system which only supports digits and dots as file names. Do I really have to name the file 123.456?

opnsenseuser:
@fabianfrz I do not quite understand why this is such a big problem, simply convert the file so that it can also be read by windows without problems. Instead of the ":" simply take a hyphen and the problem is solved for all.

fabianfrz:
[...]

Issue #2498, opened 2018-06-30
Backup file name contains colon, which is not allowed in Windows


astrandb:
I am testing backup to Nextcloud. The file name of the backup is typically config-myrouter.mydomain-2018-06-30_18:55:00.xml . However, the colon used as time separator is not allowed in filenames in Windows. If you download a backup file via the web interface Nextcloud will translate the colon to an underscore. The Nextcloud sync client for Windows refuses to sync the backup files. 
If you want to manipulate or store backup files on Windows file system you will have to rename the file and remember to rename it back again when re-uploading to Nexcloud.
In order to make it possible to move the files between systems I suggest that the filenames are formatted without time separator e.g. config-xxx-2018-06-30_185500.xml or even more compact as config-xxx-20180630_185500.xml

fabianfrz:
This is not a OPNsense bug nor a Nextcloud bug. Columns in filenames are legit characters.
Look for the different file systems in wikipedia. ":" is a legit character.

astrandb:
No,  : is not allowed in filenames in Windows. It is used to distinguish the drive letter e.g. C:
One reference from Microsoft: https://docs.microsoft.com/en-us/windows/desktop/msi/filename

fabianfrz:
So it is Windows issue because the underlying file system (NTFS) supports it (only `NUL` (0x00) and `/` (0x2f) are forbidden regarding to Wikipedia). For handling client OS specific issues, I see the client to be responsible to care, not a completely independent software which does not even have contact with it in that case.

astrandb:
I respect your view, but the problem remains. backup files from OPNsense cannot be stored on Windows systems as long as they contain a colon in the name. This is not unique to OPNSense nor to Nextcloud. You will find tons of examples if you google around. The simplest way to avoid this interoperability issue is to avoid ":" in filenames.

c:myfile would ambiguous if allowed. There is no way  to decide if "c:myfile" is the filename or if it refers to "myfile" on drive c:

fabianfrz:
`c:myfile` is also valid because it does not contain a '/' nor a '\' form the view of the FS but I guess It may be interpreted as a alternative data stream which is probably the root cause. It has no reference to a mount point which is visible as a drive letter.

you may open a ticket for the sync client if you think it should care about windows file name limitations:
https://github.com/nextcloud/desktop

fabianfrz:
@fichtner I created a ticket in the nextcloud client repository.

fichtner:
[...]

PR #3496, opened 2019-05-22
backup/Nextcloud: add datetime formatting field

Alphakilo:
This should help tackle #3069 and #2498 without interfering with existing naming conventions.
A review would be greatly appreciated since this is my first meaningful PR.

### DIFF
--- a/src/opnsense/mvc/app/library/OPNsense/Backup/Nextcloud.php
+++ b/src/opnsense/mvc/app/library/OPNsense/Backup/Nextcloud.php
@@ -84,6 +84,13 @@ class Nextcloud extends Base implements IBackupProvider
                 "type" => "text",
                 "label" => gettext("Directory Name"),
                 "value" => 'OPNsense-Backup'
+            ),
+            array(
+                "name" => "dateformat",
+                "type" => "text",
+                "label" => gettext("Date / Time Format"),
+                "help" => gettext("The date/time format that'll get appended to new backup files. For example: Y-m-d_h-i-s"),
+                "value" => 'Y-m-d_h:m:s'
             )
         );
         $nextcloud = new NextcloudSettings();
@@ -137,7 +144,7 @@ class Nextcloud extends Base implements IBackupProvider
             $backupdir = (string)$nextcloud->backupdir;
             $crypto_password = (string)$nextcloud->password_encryption;
             $hostname = $config->system->hostname . '.' .$config->system->domain;
-            $configname = 'config-' . $hostname . '-' .  date("Y-m-d_H:i:s") . '.xml';
+            $configname = 'config-' . $hostname . '-' .  date((string)$nextcloud->dateformat) . '.xml';
             // backup source data to local strings (plain/encrypted)
             $confdata = file_get_contents('/conf/config.xml');
             if (!empty($crypto_password)) {
--- a/src/opnsense/mvc/app/models/OPNsense/Backup/NextcloudSettings.xml
+++ b/src/opnsense/mvc/app/models/OPNsense/Backup/NextcloudSettings.xml
@@ -52,5 +52,11 @@
             <default>OPNsense-Backup</default>
             <ValidationMessage>The Backup Directory can only consist of alphanumeric characters.</ValidationMessage>
         </backupdir>
+        <dateformat type="TextField">
+            <Required>Y</Required>
+            <mask>/[dDjlNSwzWFmMntLoYyaABgGhHisuveIOPTZcrU\.\-\_\:\(\)]+/</mask>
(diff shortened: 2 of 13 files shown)


src/opnsense/mvc/app/library/OPNsense/Backup/Nextcloud.php
```diff
@@ -137,7 +144,7 @@ public function backup()
             $backupdir = (string)$nextcloud->backupdir;
             $crypto_password = (string)$nextcloud->password_encryption;
             $hostname = $config->system->hostname . '.' .$config->system->domain;
-            $configname = 'config-' . $hostname . '-' .  date("Y-m-d_H:i:s") . '.xml';
```
fichtner: the easiest way is to omit ":" use here and pipe it through gettext() before passing to date(). at least that would be the sane way and language translations can still override it if needed. giving the user this option does not really add to usability especially since we assume it's a bug sitting somewhere between iso date and backup software writers on the other end of our "push".

Alphakilo: That’d be the easiest solution, indeed. There’s already a patch floating around doing just that. 

I’m still disputing that this is a problem of the software client (i.e. Nextcloud Client). I’d expect source.filename == destination.filename from any filetransfer software. It the dest FS can’t handle characters, that’s neither the fault of the source or the transfer utility. Who-/Whatever inserted those characters is.

Having an FAT / NTFS incompatible character hardcoded into a filename is just asking for trouble.

I want to empower users to enjoy their preferred date flavor, while keeping it compatible to their respective infrastructure.

fichtner: Let's keep this simple: we can't brace for all auxiliary impact. If aesthetically pleasing ISO dates are bad for any such reason we might as well use simple unix timestamps. I just don't think we'll ever be able to accommodate 8.3 in all of its restrictiveness so that is not a technical discussion to have in the first place.

Alphakilo: 8.3 is auto generated these days, no reasons to worry about that.
Epoch is fine as well, just not as readable. 

Could you expand on what you mean by „auxiliary impact“?

fichtner: A seemingly simple choice for ISO date with 7bit ASCII characters leading to an incompatibility with a file system, how can anyone account for this... it even degrades readability on both ends and affects users who had no issues with this in the first place. But it is what it is. Would you mind updating and simplifying your PR?

src/opnsense/mvc/app/models/OPNsense/Backup/NextcloudSettings.xml
```diff
@@ -52,5 +52,11 @@
             <default>OPNsense-Backup</default>
             <ValidationMessage>The Backup Directory can only consist of alphanumeric characters.</ValidationMessage>
         </backupdir>
+        <dateformat type="TextField">
+            <Required>Y</Required>
+            <mask>/[dDjlNSwzWFmMntLoYyaABgGhHisuveIOPTZcrU\.\-\_\:\(\)]+/</mask>
+            <default>Y-m-d_h:m:s</default>
```
fabianfrz: The i for minutes must stay, otherwise you break the format. Also in the frontend.

src/opnsense/mvc/app/models/OPNsense/Backup/NextcloudSettings.xml
```diff
@@ -52,5 +52,11 @@
             <default>OPNsense-Backup</default>
             <ValidationMessage>The Backup Directory can only consist of alphanumeric characters.</ValidationMessage>
         </backupdir>
+        <dateformat type="TextField">
+            <Required>Y</Required>
```
fabianfrz: This requires a version bump

fabianfrz:
@fichtner this change will require urlencoding the filename and the "/" still must be rejected.

fabianfrz:
yes url encoding is required: https://en.wikipedia.org/wiki/Percent-encoding#Percent-encoding_reserved_characters

src/opnsense/mvc/app/models/OPNsense/Backup/NextcloudSettings.xml
Alphakilo: Nice catch

Alphakilo:
> this won't lower support efforts.

I have my doubts that hardcoding the next iteration will achieve this, but so be it. I'll prepare another PR.

PR #3501, opened 2019-05-25
core: fix flipped bit for IPv6

fabianfrz:
"allow IPv6" disables the allow rule…

### DIFF
--- a/src/etc/inc/filter.lib.inc
+++ b/src/etc/inc/filter.lib.inc
@@ -207,7 +207,7 @@ function filter_core_rules_system($fw, $defaults)
     // block All IPv6 except loopback traffic
     $fw->registerFilterRule(
         1,
-        array('interface' => 'loopback', 'ipprotocol'=>'inet6', 'disabled' => isset($config['system']['ipv6allow']),
+        array('interface' => 'loopback', 'ipprotocol'=>'inet6', 'disabled' => !isset($config['system']['ipv6allow']),
           'descr' => 'Pass all loopback IPv6', '#ref' => 'system_advanced_firewall.php#ipv6allow'),
         $defaults['pass']
     );


fichtner:
I'm pretty sure this is intended for squid.... it needs IPv6 loopback for startup even if IPv6 is disabled.... please take a look at the actual commit when this was introduced.

fabianfrz:
@fichtner the bug is from 28.10.2016, Caused by @AdSchellevis while initial refactoring the firewall code for plugins:

https://github.com/opnsense/core/commit/3085e4d18f1e8a644b511adf340948b14bb50c47

fichtner:
I said please review the original code... 102443a3e7d86

fabianfrz:
if that's what you want, the disabled flag must be hardcoded to false.

AdSchellevis:
did an extra check to be sure, seems to be doing what it supposed to be doing (and matches the description on top):
IPv6 off:
```
pass in log quick on lo0 inet6 from {any} to {any} label "ff61957b46d85467edab3a09e75a7a18" # Pass all loopback IPv6
block in log quick inet6 from {any} to {any} label "ec6ff90f44157f70c967adb7a6bbd7f6" # Block all IPv6
```

IPv6 on:

```
# pass in log quick on lo0 inet6 from {any} to {any} label "ec255e41800c7c02991730f55697f0b5" # Pass all loopback IPv6
# block in log quick inet6 from {any} to {any} label "89b26ab37bd7eb28f23e277884c2bfb0" # Block all IPv6
```

fabianfrz:
The checkbox info says:

"All IPv6 traffic will be blocked by the firewall unless this box is checked." In my case it is checked and the **PASS** rule is commented out.

AdSchellevis:
The pass rule should be commented, since the quick block isn't there either.

PR #3503, opened 2019-05-27
backup/Nextcloud: remove colons from filenames

Alphakilo:
supersedes #3496


pampersrocker:
Is there any update on this?

fabianfrz:
Y-m-d_His leads to something like 2019-11-14_171600.xml. I do not like that format. It can be an option but should not be the default.

mimugmail:
Hi, can we change this in any way? It doesn't make sense to constantly stick on the colon when this file format is unusable in windows. When you download a backup from nextcloud to windows to check it's content it's not viewable.

AdSchellevis:
@fabianfrz would you be so kind to test if the pr works and explain what possible downsides could be when polling this in? A lot of people are using Windows, that's the reality we live in, if we can help the community by changing are defaults a bit, we might consider doing so. Thanks in advance. (an option could also be to postpone switching defaults to 20.1)

fabianfrz:
This has two downsides:

* translators may not be aware of placeholders of the date functions
* the filename is in my opinion totally ugly

But I can confirm that the patch is working (see the last entry in the screenshot):
Screenshot_20191123_150540

fabianfrz:
BTW: Tested with NextCloud version 17.0.1

fabianfrz:
The change my be breaking for some people. I think if you pull it in then you should not merge it to stable and wait for 20.1. This requires at least a release note.

AdSchellevis:
Ok, I'll keep the PR open for now and merge when we're moving to 20.1 to prevent breakage

Alphakilo:
> 
> 
> I do not like that format. It can be an option but should not be the default.
Me neither @fabianfrz. But this static format was requested in this review: https://github.com/opnsense/core/pull/3496#discussion_r286441871

fabianfrz:
Please resolve the merge conflict

fichtner:
FWIW, omitted ":" could also be replaced by "_" as already used in the format.

oxivanisher:
Does this need a tag or milestone to be merged? Just ran into this problem and almost opened another issue. Since its already fixed, it would be nice if its included in the near future. :)

AdSchellevis:
@oxivanisher we planned to add it in 20.1, but I forgot about it to be honest, which moves this to 20.7, since we don't want to change this in a minor upgrade.

oxivanisher:
Sooo .... 20.1.8 just dropped but this is sadly still not merged. For any one else having this problem, I hacked this "solution" together to [rename (since : is illegal in windows filenames)](https://github.com/opnsense/core/pull/3503#issuecomment-557800867) and move the files to the desired subfolder.

fstab:
```
CLOUD_URL/remote.php/webdav /home/OS_USER/Nextcloud davfs user,auto,uid=OS_USER,file_mode=600,dir_mode=700,_netdev 0 0
```

crontab entries:
```
# move opnsense backups to correct subfolder
10 5 * * * mv /home/OS_USER/Nextcloud/OPNsense-Backup/*.xml /home/OS_USER/Nextcloud/SUB_FOLDER/SUB_SUB_FOLDER/AND_SO_ON/
# rename opnsense backups
15 5 * * * cd /home/OS_USER/Nextcloud/SUB_FOLDER/SUB_SUB_FOLDER/AND_SO_ON/ && find . -type f -name '*:*' -execdir bash -c 'mv "$1" "${1//:/-}"' bash {} \;
```

AdSchellevis:
merged in https://github.com/opnsense/core/commit/651284ea47fa2c5970b63a56418964d4e375b581 without `gettext()` for release in 20.7.

fichtner:
Just a question, why was the gettext() omitted?

AdSchellevis:
@fichtner to be honest, I missed it the first time,  but by my knowledge the `date()` function parameters are not language specific (Y is always year). Trying to translate this as language might lead to unexpected input for `date()` and hard to find issues when someone adds a translation for `Y-m-d_H_i_s` into `my great date format that doesn't exist`

If someone would like the format to be dynamic, he/she should probably add a user parameter (with validation) in the form code.

I should have been more explicit about the reason of omitting it indeed, but I don't think we should pull it back in.

fichtner:
Ok, thanks for clearing it up. 😊

PR #3521, opened 2019-06-07
Renamed mis-leading label 'kill states' to 'disable kill states'

derjohn:


fichtner:
Thanks for the pointer, but I don't see benefit from making the section label longer and bouncing the translation in the process. We usually try to clear up either the options shown in the select or the checkbox label or push things of operational value into the help text or clarify that if needed.

derjohn:
hmmm,
but in the translations en_US is not translated at all:

msgid "Kill states"
msgstr ""

I assumed that for en_US the msgid == msgstr !? And it's really mis-leading. Should I simply add translations for _all_ $LANG ?

fichtner:
en_US is the native language, there are no translations (the are handled as "pass through" like in other languages with the source string when no translation is given). that's also why it's a pot file, not po :)

derjohn:
Yes, understood. But what is the right way to make the visible label in Web UI show "Disable kill state" ?

fichtner:
This patch, but it's superfluous considering that the section label shouldn't redundantly explain the option explained again on the right and help text. Moreover, we may add more settings here that are all relevant to "kill states" so leaving the indicator as is is more appealing than having "disable kill state" "disable voip kill state" "enable rule kill state" on the left for the eye to "parse" through.

derjohn:
It's the annoying unexpected behavior that you "enable" a feature with a checkbox to disable it. Maybe you should change the behavior of the checkbox. BTW: Other options are prefixed like "disable" or "skip", but not this one.
I am closing the PR with a 🤷‍♂

fichtner:
I don't mind flipping the checkbox value actually. We do it in a few places to improve reality vs. expectations. Looking at the help text, it's certainly horrible and in need of a cleanup. Will you provide a new PR?

derjohn:
@fichtner Yes, I'll have a look into that approach.

fichtner:
Thanks. An example was rasend/norasend flipping: https://github.com/opnsense/core/commit/96383c8aa40f779c6fe41942cf5259da67b5425d

The config.xml value stays the same but the GUI intentionally flips the checkbox for the user.

PR #3529, opened 2019-06-11
Update main menu by after installing or deleting plugin.

kekek2:


AdSchellevis:
this looks like an action package [de]install should (already) do

src/opnsense/mvc/app/views/OPNsense/Core/firmware.volt
```diff
@@ -334,6 +334,15 @@
...
+                            type: 'GET',
+                            url: '/ui/core/firmware',
+                            dataType:'html',
+                            contentType: 'text/html',
+                            complete: function(data,status) {
+                                $('#mainmenu').html($('#mainmenu', data.responseText));
+                            }
+                        });
                     } else {
                         $('#updatestatus').html("{{ lang._('Upgrade done.') }}");
```
fichtner: Idea is good, but we need to do it for this line as well or better yet the whole upper else to catch all future cases. Some updates will be spurious, but better than not doing it.

AdSchellevis: It seems I misread in my first response, but replacing the menu like this might have side affects on already registered events. 

Preventing users from going to going to non existing location isn't too hard to handle by requesting the current items from the api and removing the ones that are gone. Ensuring the menu is ajax enabled is a different story when doing it properly (its now a hybrid structure).

AdSchellevis:
a page reload instead of changing `updatestatus` would probably do the trick as well, but I'm not planning to try this out currently.

https://github.com/opnsense/core/blob/2ad9300a32a98289edd7a636fa757b0b57390a9c/src/opnsense/mvc/app/views/OPNsense/Core/firmware.volt#L335-L339

PR #3537, opened 2019-06-15
Sort interface groups in GUI to match firewall rule order

ypid:
This now makes it easy and predictable to add interfaces to multiple groups.

Before this change, the interfaces_groups GUI was sorted by the order that interface groups were added. However, this was not the order that the actual pf rules would then be generated making it unpredictable when adding interfaces to multiple groups.

The filter_rules_sort function already took care of the actual pf rule order.

I also took care of only marking the filter subsystem dirty when needed.  I tested this patch quite a bit on 19.1.8. When updating, users only need to make one change to the interface groups to have them sorted in the GUI.

I hope I did not miss anything because this patch is against master. In it’s current state, the patch can be directly applied to 19.1.x or master.

### DIFF
--- a/src/www/interfaces_groups.php
+++ b/src/www/interfaces_groups.php
@@ -30,6 +30,8 @@
 
 require_once("guiconfig.inc");
 require_once("interfaces.inc");
+require_once("filter.inc");
+require_once("system.inc");
 
 $a_ifgroups = &config_read_array('ifgroups', 'ifgroupentry');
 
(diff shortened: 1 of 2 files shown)


AdSchellevis:
Hi @ypid, I'll have to look into the sorting order, but I'm very in doubt about the filter reload magic. Are you sure changes need a filter reload? (since the pf rule itself is bound on the group, not the interfaces.) or is this only a side affect of your change?

Adding overlapping groups might always be challenging by the way, I'm not sure if presenting the order differently will change this situation a lot.

It could be an option to add/calculate some sort of priority and use that in our usual rule placement

https://github.com/opnsense/core/blob/426061a17b0958463d83ae2b95193c1aa8696544/src/etc/inc/filter.inc#L184-L190


Best regards,

Ad

ypid:
@AdSchellevis Thanks for your feedback.

> It could be an option to add/calculate some sort of priority and use that in our usual rule placement

This is already done implicitly even without my change (when I remember correctly).

> Are you sure changes need a filter reload?

You are right. A filter reload is only needed in a few case for example when the interface group name is changed because then the order of the pf rules could change. My patch takes this into account as mentioned.

> Adding overlapping groups might always be challenging by the way, I'm not sure if presenting the order differently will change this situation a lot.

I would be interested in more details on this. I noticed (because of the not-matching order of UI vs. pf rules) that this use case is new. It can still be useful I think and I am using it already :wink:

AdSchellevis:
I thought changing wasn't allowed when used, but it seems like we're not validating that, so since we seem to replace the (interface)item inside the rules we might as well offer the opportunity to reload here.

The use-case itself can indeed be practical, let me think a bit about the sort order and validate before pulling this in. I'll have to check if forcing the interface here guarantees correct order in all scenario's.

Thanks!

AdSchellevis:
Ok, we should be safe here, thanks!

ypid:
That was quick. Thank you very much! Keep up the amazing work :+1:

PR #3553, opened 2019-06-28
Do not call configdRun('unbound ') if Unbound not enabled.

kekek2:
I'm turning off the Unbound service. I go to the page "Services: Unbound DNS: Statistics". The JS calls more often configdRun('unbound stats') than it works. Web GUI is dead.

### DIFF
--- a/src/opnsense/scripts/unbound/wrapper.py
+++ b/src/opnsense/scripts/unbound/wrapper.py
@@ -33,20 +33,20 @@ import tempfile
 import subprocess
 import argparse
 import json
+import psutil
 
 def unbound_control_reader(action):
+    try:
+        if not psutil.pid_exists(int(open("/var/run/unbound.pid").read().strip())):
+            sys.exit(1)
+    except:
+        sys.exit(1)
     with tempfile.NamedTemporaryFile() as output_stream:
         subprocess.call(['/usr/local/sbin/unbound-control', '-c', '/var/unbound/unbound.conf', action],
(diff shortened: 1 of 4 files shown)


fichtner:
it's probably better to script this in the backend so that it returns if unbound isn't running. if it's enabled but stopped the same issue is there

fichtner:
okay not bad, but you see there is a wrapper script which is perfect for avoiding duplication....

src/opnsense/scripts/unbound/wrapper.py
```diff
@@ -42,6 +42,12 @@ def unbound_control_reader(action):
         for line in output_stream:
             yield line.decode()
 
+try:
```
AdSchellevis: block should be below argument parser to avoid silent exit on `-h` for example

src/opnsense/scripts/unbound/wrapper.py
```diff
@@ -42,6 +42,12 @@ def unbound_control_reader(action):
...
+    if subprocess.call("pgrep -F /var/run/unbound.pid 2>/dev/null >/dev/null", shell=True) != 0:
```
AdSchellevis: I would go for a more native python approach here, for inspiration, this is how the people from psutil seem to fix this : https://github.com/giampaolo/psutil/blob/c10df5aa04e1ced58d19501fa42f08c1b909b83d/psutil/_psposix.py#L32-L51

PR #3555, opened 2019-07-01
Translate LAGG.

kekek2:


src/www/interfaces_lagg_edit.php
```diff
@@ -217,7 +217,7 @@ interface_configure(false, $confif);
 <?php
                       foreach ($laggprotos as $proto):?>
                         <option value="<?=$proto;?>" <?=$proto == $pconfig['proto'] ? "selected=\"selected\"": "";?>>
-                            <?=strtoupper($proto);?>
+                            <?=strtoupper(gettext($proto));?>
```
fabianfrz: this will probably not work because the string cannot be extracted.

PR #3570, opened 2019-07-13
ipsec: Add support for public key authentication

ppmathis:
**Feature Description**
The current IPsec plugin implementation does not support public key authentication, which allows for a more secure mutual authentication than PSK while still not introducing the complexity of X509 certificates. The authentication can easily be set up by generating a bare RSA keypair chain on both machines, followed by exchanging the public keys between the two peers.

This commit introduces public key authentication functionality by adding a new authentication method to phase 1 configuration called "Mutual Public Key" and adding a menu entry "Public Key Authentication", which allows adding public keys (+ optional private keys). It was successfully tested against a Linux virtual machine running Strongswan 5 and the entered RSA keys are automatically verified for correctness.

Useful commands for generating a bare RSA keypair:
```sh
ipsec pki --gen --type rsa --outform pem --size 4096 > private.pem
ipsec pki --pub --outform pem --in private.pem > public.pem
```

**References**
The DN42 project features a Wiki page which explains the benefits of IPsec Public Key Authentication in further detail and also provides documentation for setting it up on various systems: https://dn42.net/howto/IPsec-with-PublicKeys

### DIFF
--- a/src/etc/inc/plugins.inc.d/ipsec.inc
+++ b/src/etc/inc/plugins.inc.d/ipsec.inc
@@ -1,6 +1,7 @@
 <?php
 
 /*
+ * Copyright (C) 2019 Pascal Mathis <mail@pascalmathis.com>
  * Copyright (C) 2016 Deciso B.V.
  * Copyright (C) 2008 Shrew Soft Inc. <mgrooms@shrew.net>
  * Copyright (C) 2008 Ermal Luçi
@@ -86,6 +87,7 @@ function ipsec_p1_authentication_methods()
         'rsa_eap-mschapv2' => array( 'name' => 'Mutual RSA + EAP-MSCHAPV2', 'mobile' => true),
         'eap-radius' => array( 'name' => 'EAP-RADIUS', 'mobile' => true),
         'rsasig' => array( 'name' => 'Mutual RSA', 'mobile' => false ),
+        'pubkey' => array( 'name' => 'Mutual Public Key', 'mobile' => false ),
         'pre_shared_key' => array( 'name' => 'Mutual PSK', 'mobile' => false ),
     );
 }
@@ -542,6 +544,14 @@ function ipsec_mobilekey_sort()
     });
 }
 
+function ipsec_lookup_keypair($uuid)
+{
+    $mdl = new \OPNsense\IPsec\IPsec();
+    $node = $mdl->getNodeByReference('keyPairs.keyPair.' . $uuid);
+
+    return $node ? $node->getNodes() : null;
+}
+
 function ipsec_get_number_of_phase2($ikeid)
 {
     global $config;
@@ -791,14 +801,16 @@ function ipsec_configure_do($verbose = false, $interface = '')
     } else {
         $certpath = "/usr/local/etc/ipsec.d/certs";
         $capath = "/usr/local/etc/ipsec.d/cacerts";
-        $keypath = "/usr/local/etc/ipsec.d/private";
+        $publickeypath = "/usr/local/etc/ipsec.d/public";
+        $privatekeypath = "/usr/local/etc/ipsec.d/private";
 
         mwexec("/sbin/ifconfig enc0 up");
         set_single_sysctl("net.inet.ip.ipsec_in_use", "1");
 
         /* needed directories for config files */
         @mkdir($capath);
-        @mkdir($keypath);
+        @mkdir($privatekeypath);
+        @mkdir($publickeypath);
         @mkdir($certpath);
         @mkdir('/usr/local/etc/ipsec.d');
         @mkdir('/usr/local/etc/ipsec.d/crls');
@@ -1093,7 +1105,7 @@ function ipsec_configure_do($verbose = false, $interface = '')
 
                 @chmod($certpath, 0600);
 
-                $ph1keyfile = "{$keypath}/cert-{$ph1ent['ikeid']}.key";
+                $ph1keyfile = "{$privatekeypath}/cert-{$ph1ent['ikeid']}.key";
                 if (!file_put_contents($ph1keyfile, base64_decode($cert['prv']))) {
                     log_error(sprintf('Error: Cannot write phase1 key file for %s', $ph1ent['name']));
                     continue;
@@ -1117,6 +1129,59 @@ function ipsec_configure_do($verbose = false, $interface = '')
                 if (!empty($peerid_data)) {
                     $pskconf .= $myid . " " . trim($peerid_data) . " : PSK 0s" . base64_encode(trim($ph1ent['pre-shared-key'])) . "\n";
                 }
+            }
+        }
+
+        /* Generate files for key pairs (e.g. RSA) */
+        foreach ($a_phase1 as $ph1ent) {
+            if(isset($ph1ent['disabled'])) {
+                continue;
+            }
+
+            if (!empty($ph1ent['local-kpref'])) {
+                $keyPair = ipsec_lookup_keypair($ph1ent['local-kpref']);
+                if (!$keyPair || empty($keyPair['publicKey']) || empty($keyPair['privateKey'])) {
+                    log_error(sprintf('Error: Invalid phase1 local key pair reference for %s', $ph1ent['name']));
+                    continue;
+                }
+
+                @chmod($publickeypath, 0600);
+
+                $ph1publickeyfile = "${publickeypath}/publickey-local-{$ph1ent['ikeid']}.pem";
+                if (!file_put_contents($ph1publickeyfile, $keyPair['publicKey'])) {
+                    log_error(sprintf('Error: Cannot write phase1 local public key file for %s', $ph1ent['name']));
+                    @unlink($ph1publickeyfile);
+                    continue;
+                }
+                @chmod($ph1publickeyfile, 0600);
+
+                $ph1privatekeyfile = "${privatekeypath}/privatekey-local-{$ph1ent['ikeid']}.pem";
+                if (!file_put_contents($ph1privatekeyfile, $keyPair['privateKey'])) {
+                    log_error(sprintf('Error: Cannot write phase1 local private key file for %s', $ph1ent['name']));
+                    @unlink($ph1privatekeyfile);
+                    continue;
+                }
+                @chmod($ph1privatekeyfile, 0600);
+
+                $pskconf .= " : RSA {$ph1privatekeyfile}\n";
+            }
+
+            if (!empty($ph1ent['peer-kpref'])) {
+                $keyPair = ipsec_lookup_keypair($ph1ent['peer-kpref']);
+                if (!$keyPair || empty($keyPair['publicKey'])) {
+                    log_error(sprintf('Error: Invalid phase1 peer key pair reference for %s', $ph1ent['name']));
+                    continue;
+                }
+
+                @chmod($publickeypath, 0600);
+
+                $ph1publickeyfile = "${publickeypath}/publickey-peer-{$ph1ent['ikeid']}.pem";
+                if (!file_put_contents($ph1publickeyfile, $keyPair['publicKey'])) {
+                    log_error(sprintf('Error: Cannot write phase1 peer public key file for %s', $ph1ent['name']));
+                    @unlink($ph1publickeyfile);
+                    continue;
+                }
+                @chmod($ph1publickeyfile, 0600);
             }
         }
 
@@ -1289,12 +1354,14 @@ function ipsec_configure_do($verbose = false, $interface = '')
                         $authentication = "leftauth = psk\n\trightauth = psk";
                         break;
                     case 'rsasig':
+                    case 'pubkey':
                         $authentication = "leftauth = pubkey\n\trightauth = pubkey";
                         break;
                     case 'hybrid_rsa_server':
                         $authentication = "leftauth = pubkey\n\trightauth = xauth";
                         break;
                 }
+
                 if (!empty($ph1ent['certref'])) {
                     $authentication .= "\n\tleftcert = {$certpath}/cert-{$ph1ent['ikeid']}.crt";
                     $authentication .= "\n\tleftsendcert = always";
@@ -1309,8 +1376,15 @@ function ipsec_configure_do($verbose = false, $interface = '')
                         $authentication .= "\n\trightca = \"/$rightca\"";
                     }
                 }
-                $left_spec = $ep;
 
+                if (!empty($ph1ent['local-kpref'])) {
+                    $authentication .= "\n\tleftsigkey = {$publickeypath}/publickey-local-{$ph1ent['ikeid']}.pem";
+                }
+                if (!empty($ph1ent['peer-kpref'])) {
+                    $authentication .= "\n\trightsigkey = {$publickeypath}/publickey-peer-{$ph1ent['ikeid']}.pem";
+                }
+
+                $left_spec = $ep;
                 if (isset($ph1ent['reauth_enable'])) {
                     $reauth = "reauth = no";
                 } else {
--- a/src/opnsense/mvc/app/controllers/OPNsense/IPsec/Api/KeyPairsController.php
+++ b/src/opnsense/mvc/app/controllers/OPNsense/IPsec/Api/KeyPairsController.php
@@ -1,0 +1,114 @@
+<?php
+/*
+ * Copyright (C) 2019 Pascal Mathis <mail@pascalmathis.com>
+ * All rights reserved.
+ *
+ * Redistribution and use in source and binary forms, with or without
+ * modification, are permitted provided that the following conditions are met:
+ *
+ * 1. Redistributions of source code must retain the above copyright notice,
+ *    this list of conditions and the following disclaimer.
+ *
+ * 2. Redistributions in binary form must reproduce the above copyright
+ *    notice, this list of conditions and the following disclaimer in the
+ *    documentation and/or other materials provided with the distribution.
+ *
+ * THIS SOFTWARE IS PROVIDED ``AS IS'' AND ANY EXPRESS OR IMPLIED WARRANTIES,
+ * INCLUDING, BUT NOT LIMITED TO, THE IMPLIED WARRANTIES OF MERCHANTABILITY
+ * AND FITNESS FOR A PARTICULAR PURPOSE ARE DISCLAIMED. IN NO EVENT SHALL THE
+ * AUTHOR BE LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY,
+ * OR CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF
+ * SUBSTITUTE GOODS OR SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS
+ * INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN
+ * CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE)
+ * ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
+ * POSSIBILITY OF SUCH DAMAGE.
+ */
+
+namespace OPNsense\IPsec\Api;
+
+use OPNsense\Base\ApiMutableModelControllerBase;
+
+/**
+ * Class KeyPairsController
+ * @package OPNsense\IPsec\Api
+ */
+class KeyPairsController extends ApiMutableModelControllerBase
+{
+    protected static $internalModelName = 'ipsec';
+    protected static $internalModelClass = 'OPNsense\IPsec\IPsec';
+
+    /**
+     * Search key pairs
+     * @return array
+     * @throws \ReflectionException
+     */
+    public function searchItemAction()
+    {
+        return $this->searchBase(
+            'keyPairs.keyPair',
+            ['name', 'keyType', 'keySize', 'keyFingerprint']
+        );
+    }
+
+    /**
+     * Update key pair with given properties
+     * @param $uuid
+     * @return array
+     * @throws \OPNsense\Base\UserException
+     * @throws \ReflectionException
+     */
+    public function setItemAction($uuid = null)
+    {
+        $response = $this->setBase('keyPair', 'keyPairs.keyPair', $uuid);
+        if (!empty($response['result']) && $response['result'] === 'saved') {
+            touch('/tmp/ipsec.dirty'); // mark_subsystem_dirty('ipsec')
+        }
+
+        return $response;
+    }
+
+    /**
+     * Add new key pair with given properties
+     * @return array
+     * @throws \OPNsense\Base\UserException
+     * @throws \ReflectionException
+     */
+    public function addItemAction()
+    {
+        $response = $this->addBase('keyPair', 'keyPairs.keyPair');
+        if (!empty($response['result']) && $response['result'] === 'saved') {
+            touch('/tmp/ipsec.dirty'); // mark_subsystem_dirty('ipsec')
+        }
+
+        return $response;
+    }
+
+    /**
+     * Retrieve key pair or return defaults for new one
+     * @param $uuid
+     * @return array
+     * @throws \ReflectionException
+     */
+    public function getItemAction($uuid = null)
+    {
+        return $this->getBase('keyPair', 'keyPairs.keyPair', $uuid);
+    }
+
+    /**
+     * Delete key pair by UUID
+     * @param $uuid
+     * @return array
+     * @throws \OPNsense\Base\UserException
+     * @throws \ReflectionException
+     */
+    public function delItemAction($uuid)
+    {
+        $response = $this->delBase('keyPairs.keyPair', $uuid);
+        if (!empty($response['result']) && $response['result'] === 'deleted') {
+            touch('/tmp/ipsec.dirty'); // mark_subsystem_dirty('ipsec')
+        }
+
+        return $response;
+    }
+}
--- a/src/opnsense/mvc/app/controllers/OPNsense/IPsec/Api/LegacySubsystemController.php
+++ b/src/opnsense/mvc/app/controllers/OPNsense/IPsec/Api/LegacySubsystemController.php
@@ -1,0 +1,79 @@
+<?php
+/*
+ * Copyright (C) 2019 Pascal Mathis <mail@pascalmathis.com>
+ * All rights reserved.
+ *
+ * Redistribution and use in source and binary forms, with or without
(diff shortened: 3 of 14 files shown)


src/www/vpn_ipsec_pubkeys.php
```diff
@@ -0,0 +1,203 @@
+<?php
+
+/*
+ * Copyright (C) 2019 Pascal Mathis <mail@pascalmathis.com>
```
fichtner: You probably copied a file, I’m not going to accept this as is.

src/etc/inc/xmlparse.inc
```diff
@@ -44,7 +44,7 @@ function listtags()
       'member', 'menu', 'mobilekey', 'monitor_type', 'mount',
       'npt', 'ntpserver',
       'onetoone', 'openvpn-server', 'openvpn-client', 'openvpn-csc', 'option',
-      'package', 'passthrumac', 'phase1', 'phase2', 'ppp', 'pppoe', 'priv', 'proxyarpnet', 'pool', 'pages', 'pipe',
+      'package', 'passthrumac', 'phase1', 'phase2', 'ppp', 'pppoe', 'priv', 'proxyarpnet', 'pool', 'pages', 'pipe', 'publickey',
```
fichtner: Is this really necessary? The code is old and in need of replacement and the list trickery in particular is prone to errors.

src/www/vpn_ipsec_pubkeys.php
```diff
@@ -0,0 +1,203 @@
...
+require_once("filter.inc");
+require_once("services.inc");
+require_once("interfaces.inc");
+require_once("plugins.inc.d/ipsec.inc");
+require_once('phpseclib/Math/BigInteger.php');
+require_once('phpseclib/Crypt/Hash.php');
+require_once('phpseclib/Crypt/RSA.php');
+
+config_read_array('ipsec', 'publickey');
+ipsec_publickey_sort();
```
fichtner: Why? Wouldn’t it be better to write them in a sorted fashion?

src/www/vpn_ipsec_pubkeys.php
```diff
@@ -0,0 +1,203 @@
...
+
+function parse_rsa_key($key_str)
+{
+    // Skip parsing of empty key
+    if (!$key_str) {
+        return false;
+    }
+
+    // Attempt loading key as RSA key
+    $rsa_key = new \phpseclib\Crypt\RSA();
```
fichtner: Doesn’t this need require headers to avoid breakage later on?

src/www/vpn_ipsec_pubkeys_edit.php
```diff
@@ -0,0 +1,240 @@
+<?php
+
+/*
+ * Copyright (C) 2019 Pascal Mathis <mail@pascalmathis.com>
```
fichtner: Same here

src/www/vpn_ipsec_pubkeys.php
ppmathis: The required includes for \phpseclib\Crypt\RSA() are included earlier in the same file (see line 33 - 36) and the code was tested on a OPNsense machine running the latest release with `make mount`. phpseclib itself is already a dependency of the existing source code, specifically in `system_certmanager.php`. I made sure to only include files which are actually required and various public and private keys worked fine with the current list of includes.

src/etc/inc/xmlparse.inc
ppmathis: I have noticed that the whole IPsec plugin is based on legacy code, but as multiple keys (RSA public + optional private key) can be stored, I can not think of an array-less way of storing this data. I made sure to follow the current code style / implementation, specifically following along the IPsec Mobile Keys implementation. And no other free / unused arrays existed for configuring this, however I am happy to rename the key to e.g. ipsec-publickey if preferred.

src/www/vpn_ipsec_pubkeys.php
ppmathis: I wanted to avoid introducing too much complexity into the already fairly ugly IPsec implementation code (I have seen the ongoing discussions about this needing a rewrite one day) and oriented myself based on the IPsec Mobile Keys implementation, which also sorts in the same way - which is why I sticked to a similar way. I'm happy to take the different route of sorting keys during saving.

src/www/vpn_ipsec_pubkeys.php
ppmathis: The required includes for \phpseclib\Crypt\RSA() are included earlier in the same file (see line 33 - 36) and the code was tested on a OPNsense machine running the latest release with make mount. phpseclib itself is already a dependency of the existing source code, specifically in system_certmanager.php. I made sure to only include files which are actually required and various public and private keys worked fine with the current list of includes.

src/www/vpn_ipsec_pubkeys.php
ppmathis: Good catch. While I did not copy one-by-one, I based various snippets / methods on the IPsec Mobile Key implementation and forgot to add them to the header. I will add Deciso B.V. and Manuel Kasper to the copyright notice once the other discussion points are finished.

src/www/vpn_ipsec_pubkeys_edit.php
ppmathis: Will also add Deciso B.V. and Manuel Kasper here for the reasons mentioned earlier.

src/www/vpn_ipsec_pubkeys.php
```diff
@@ -0,0 +1,203 @@
...
+ */
+
+require_once("guiconfig.inc");
+require_once("filter.inc");
+require_once("services.inc");
+require_once("interfaces.inc");
+require_once("plugins.inc.d/ipsec.inc");
+require_once('phpseclib/Math/BigInteger.php');
+require_once('phpseclib/Crypt/Hash.php');
+require_once('phpseclib/Crypt/RSA.php');
```
fabianfrz: is this needed or will the openssl functions do it as well?

ppmathis: I have chosen phpseclib as the builtin OpenSSL functions are extremely restrictive about the supported key formats. As an example, OpenSSL only allows PKCS#1 keys (`-----BEGIN PUBLIC KEY-----`) and fails when parsing PKCS#8 keys (`-----BEGIN RSA PUBLIC KEY-----`). As both key formats are fairly common and Strongswan officially supports both of them as well, I did not want to manually restrict the plugin to a single format -or- mangle the key strings manually so that OpenSSL is satisified.

A short example is available here: https://repl.it/repls/GrownWarlikeQueryoptimizer

fabianfrz: In my opinion an automatic conversion to OpenSSL would make sense if it fails in first place. @fichtner What do you think?

src/etc/inc/xmlparse.inc
ppmathis: I double-checked and the only other key which seems suitable would be `key`, however this seems ambiguous to me as the IPsec plugin already makes use of `mobilekey`. I would prefer to introduce `pluginkey` as a list tag, as it does not seem to be in use anywhere in the current OPNsense core (so it should not break anything hopefully, if I understood this function correctly) and it would clearly show which kind of key it is referring to.

src/www/vpn_ipsec_pubkeys.php
ppmathis: I gave this another thought and ended up rewriting parts of my implementation - aside from a few cleanups mentioned above, I moved all the heavy lifting of parsing, validating and sanitizing the RSA keys into a single function within `ipsec.inc` called `ipsec_parse_rsa_key($key_str, $key_type)`.

As every single documentation I could find for public key authentication actually says you should use the `ipsec pki` commands provided by Strongswan itself, I think accepting only X.509 keys is sufficient for 99.9% of use cases, as the Strongswan tool will always output the keys in said format. It will save us from additional complexity of mangling key formats (which tend to get complicated in specific situations) and I was able to fully drop the dependency on phpseclib, replacing it with native OpenSSL functions.

Both keys are now being properly sanitized when saved, so any unnecessary spaces and newlines get removed automatically. Additionally, the check if the public and private key belong to each other is now implemented by comparing the fingerprint, which is more efficient than signing and verifying a dummy message.

Last but not least I also changed the fingerprinting algorithm for the RSA keys to match the `keyid` outputted by `ipsec listall`, as it seemed like a more appropriate choice than the previous ssh-rsa based fingerprinting which does not show up in Strongswan at all.

Thank you for the good inputs and code review so far. I am fairly satisfied with the new implementation, any thoughts or good-to-go?

src/etc/inc/xmlparse.inc
AdSchellevis: Maybe we better move the new functionality to a new mvc based form, it would have the advantage of not increasing the legacy codebase and it will likely be quite easy to use the new (public key) entries in the old codebase.

ppmathis: @AdSchellevis I'm more than happy to look into this as well - I have zero experience with the new MVC model but at a first glance it seems to be more sane. However I am not sure if a mixed solution is appreciated, so I'd wait for agreement / feedback from @fichtner or @fabianfrz before giving this a try.

fichtner: I do trust @AdSchellevis on this, it's a good idea to avoid the early concerns. :)

Access from legacy to MVC is quite easy:

1. Use via MVC model: https://github.com/opnsense/core/blob/master/src/etc/inc/plugins.inc.d/netflow.inc#L31-L33

2. Use via raw config access: https://github.com/opnsense/core/blob/master/src/etc/inc/plugins.inc.d/squid.inc#L35-L36

AdSchellevis: @ppmathis Today I've added a bit of extra documentation on how to use our mvc components for these type of use cases (it was still somewhere on the list of things todo).  https://docs.opnsense.org/development/examples/using_grids.html

ppmathis:
Based on feedback from @AdSchellevis and @fichtner, I have rewritten this pull request to make use of the new OPNsense MVC system and made several changes as I've been unhappy with a few parts of the previous implementation after putting further thought into them:

- The implementation now refers to **Key Pairs**, as the previous term **Public Key Authentication** in the menu has been misleading. After all, the menu item isn't there to configure pubkey-auth but to manage RSA key pairs.
- The implementation only supports RSA keys as of today, but can be easily extended to support further types (e.g. ed25519, if Strongswan should start supporting this)
- The implementation uses the builtin OpenSSL functions and does not rely on any third-party code
- The keys are being automatically stored as a sanitized representation during validation. So e.g. unnecessary newlines or spaces after the end of a key are being trimmed away.
- The key size and fingerprint are being calculated and stored into the model subsystem during validation. This avoids recalculating these stats on every page load for every single key pair, which could waste many CPU cycles.

The legacy code makes use of the new MVC model to access the configuration data. To allow the new configuration page to reload the legacy service,  `LegacySubsystemController` has been added which allows querying the status of the legacy subsystem (dirty / clean) as well as applying the configuration.

Thank you for the advice so far and looking forward to your feedback!

src/opnsense/mvc/app/views/OPNsense/IPsec/key_pairs.volt
```diff
@@ -0,0 +1,123 @@
...
+SUBSTITUTE GOODS OR SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS
+INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN
+CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE)
+ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
+POSSIBILITY OF SUCH DAMAGE.
+#}
+
+<script>
+  $(document).ready(function () {
+    const $applyLegacyConfig = $('#applyLegacyConfig');
```
fabianfrz: be careful with that in `core`. I prefer ES6 too but I don't know how long @AdSchellevis and @fichtner still want to support outdated browsers here. They should respond to this.

src/opnsense/mvc/app/views/OPNsense/IPsec/key_pairs.volt
```diff
@@ -0,0 +1,123 @@
...
+OR CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF
+SUBSTITUTE GOODS OR SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS
+INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN
+CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE)
+ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
+POSSIBILITY OF SUCH DAMAGE.
+#}
+
+<script>
+  $(document).ready(function () {
```
fabianfrz:
```suggestion
  $(function () {
```

src/opnsense/mvc/app/models/OPNsense/IPsec/IPsec.xml
```diff
@@ -0,0 +1,31 @@
...
+                </keyType>
+                <publicKey type="TextField">
+                    <Required>Y</Required>
+                </publicKey>
+                <privateKey type="TextField">
+                    <Required>N</Required>
+                </privateKey>
+
+                <_keySize type="IntegerField"/>
+                <_keyFingerprint type="TextField"/>
```
fabianfrz: Please no, underscores here look awful.

src/opnsense/mvc/app/views/OPNsense/IPsec/key_pairs.volt
```diff
@@ -0,0 +1,123 @@
...
+    const $grid = $('#grid-key-pairs').UIBootgrid({
+      search: '/api/ipsec/key-pairs/searchItem',
+      get: '/api/ipsec/key-pairs/getItem/',
+      set: '/api/ipsec/key-pairs/setItem/',
+      add: '/api/ipsec/key-pairs/addItem/',
+      del: '/api/ipsec/key-pairs/delItem/',
+    });
+
+    // Refresh status of legacy subsystem when grid has changed
+    $grid.on('loaded.rs.jquery.bootgrid', function() {
```
fabianfrz: please don't click apply - you need to add that manually if you want to (due to missing multiline support):
```suggestion
    $grid.on('loaded.rs.jquery.bootgrid', updateLegacyStatus);
```

fabianfrz:
I just had a quick view on the frontend code. Have not looked at the backend.

src/opnsense/mvc/app/models/OPNsense/IPsec/IPsec.xml
ppmathis: I have used underscores as these fields are used for caching/storing the key size and fingerprint and should not be touched manually, which is why the UI also does not offer any method for directly editing these properties.

Any other naming suggestion for this purpose?

src/opnsense/mvc/app/views/OPNsense/IPsec/key_pairs.volt
ppmathis: I manually checked this before making use of `const` and can confirm that it is being used in several places within `core` already, e.g. `src/opnsense/www/js/opnsense_ui.js` and `src/opnsense/www/js/opnsense_bootgrid_plugin.js`. As they are not preprocessed and delivered like that to the browser, I think this should be fine?

src/opnsense/mvc/app/views/OPNsense/IPsec/key_pairs.volt
ppmathis: Sure, makes sense - updated.

src/opnsense/mvc/app/models/OPNsense/IPsec/IPsec.xml
fabianfrz: I just made them looking like regular fields in my code…

If you would download the list via the API for some custom application, you would not care if the server generated it or if it comes from storage - you are just interested in the content.

This may cause some problems because some methods in PHP have special names and also start with an underscore - for example `__constructor`.

src/opnsense/mvc/app/views/OPNsense/IPsec/key_pairs.volt
ppmathis: Updated. This was a leftover from my first tries to get the UIBootgrid example from the docs working: https://wiki.opnsense.org/development/examples/using_grids.html#constructing-the-volt-template

Maybe those should be updated as well if `$(function() { ... })` is preferred over `$(document).ready(function() { ... })`?

fabianfrz: I think so. The others are deprecated according to the jQuery docs:

https://api.jquery.com/ready/

Just read the paragraph after the variants.

fabianfrz: CC @AdSchellevis

src/opnsense/mvc/app/models/OPNsense/IPsec/IPsec.xml
ppmathis: Thank you for your reply, I didn't think about the use case of querying these values through the API - were underscores would indeed be fairly awkward. I have updated the field names to `keySize` and `keyFingerprint`.

src/opnsense/mvc/app/views/OPNsense/IPsec/key_pairs.volt
fabianfrz: They have been removed some time ago and in that case it looks like they came back.
The only issue is that IE yields a syntax error which I do not care about as I don't use it. The bright side is that Windows 7 seems to become EoL in January: https://support.microsoft.com/de-at/help/4057281/windows-7-support-will-end-on-january-14-2020

So please wait for the feedback of a core maintainer.

src/opnsense/mvc/app/controllers/OPNsense/IPsec/Api/KeyPairsController.php
```diff
@@ -0,0 +1,116 @@
...
+ * CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE)
+ * ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
+ * POSSIBILITY OF SUCH DAMAGE.
+ */
+
+namespace OPNsense\IPsec\Api;
+
+use OPNsense\Base\ApiMutableModelControllerBase;
+
+require_once('util.inc');
```
AdSchellevis: legacy imports are not allowed in new style code, to avoid pulling in technical debt.

src/opnsense/mvc/app/controllers/OPNsense/IPsec/Api/KeyPairsController.php
```diff
@@ -0,0 +1,116 @@
...
+     * @param $uuid
+     * @return array
+     * @throws \OPNsense\Base\UserException
+     * @throws \ReflectionException
+     */
+    public function delItemAction($uuid)
+    {
+        $response = $this->delBase('keyPairs.keyPair', $uuid);
+        if (!empty($response['result']) && $response['result'] === 'deleted') {
+            mark_subsystem_dirty('ipsec');
```
AdSchellevis: `touch("/tmp/ipsec.dirty");` will do the trick here.

src/opnsense/mvc/app/controllers/OPNsense/IPsec/Api/LegacySubsystemController.php
```diff
@@ -0,0 +1,80 @@
...
+namespace OPNsense\IPsec\Api;
+
+use OPNsense\Base\ApiControllerBase;
+
+require_once('guiconfig.inc');
+require_once('system.inc');
+require_once('filter.inc');
+require_once('services.inc');
+require_once('interfaces.inc');
+require_once('plugins.inc.d/ipsec.inc');
```
AdSchellevis: remove legacy imports

src/opnsense/mvc/app/controllers/OPNsense/IPsec/Api/LegacySubsystemController.php
```diff
@@ -0,0 +1,80 @@
...
+class LegacySubsystemController extends ApiControllerBase
+{
+    /**
+     * Returns the status of the legacy subsystem, which currently only includes a boolean specifying if the subsystem
+     * is marked as dirty, which means that there are pending changes.
+     * @return array
+     */
+    public function statusAction()
+    {
+        return ['isDirty' => is_subsystem_dirty('ipsec')];
```
AdSchellevis: `file_exists("/tmp/ipsec.dirty)`

src/opnsense/mvc/app/controllers/OPNsense/IPsec/Api/LegacySubsystemController.php
```diff
@@ -0,0 +1,80 @@
...
+    }
+
+    /**
+     * Apply the IPsec configuration using the legacy subsystem and return a message describing the result
+     * @return array
+     */
+    public function applyConfigAction()
+    {
+        if ($this->request->isPost()) {
```
AdSchellevis: Since the code to call doesn't have a new alternative, isolate the logic and call via configd.

Very likely it would already be sufficient to call `pluginctl -c ipsec` and remove the "dirty" file when done. (the filter shouldn't be necessary) 

add action to https://github.com/opnsense/core/blob/master/src/opnsense/service/conf/actions.d/actions_ipsec.conf

then call the usual way, like 
https://github.com/opnsense/core/blob/5689dc8048cfe739f88619291190d029a59ed4b3/src/opnsense/mvc/app/controllers/OPNsense/Core/Api/SystemController.php#L53-L57

src/opnsense/mvc/app/models/OPNsense/IPsec/IPsec.php
```diff
@@ -0,0 +1,188 @@
...
+     */
+    public function parseCryptographicKey($keyString, $keyType)
+    {
+        // Attempt to load key with correct type
+        if ($keyType === 'rsa-public') {
+            $key = openssl_pkey_get_public($keyString);
+        } elseif ($keyType === 'rsa-private') {
+            $key = openssl_pkey_get_private($keyString);
+        } else {
+            throw new \InvalidArgumentException('Unsupported key type: ' . $keyType);
```
AdSchellevis: since the other arguments are passed through `gettext()` I would expect something similar here.

src/opnsense/mvc/app/models/OPNsense/IPsec/IPsec.xml
```diff
@@ -0,0 +1,31 @@
...
+                        <rsa>RSA</rsa>
+                    </OptionValues>
+                </keyType>
+                <publicKey type="TextField">
+                    <Required>Y</Required>
+                </publicKey>
+                <privateKey type="TextField">
+                    <Required>N</Required>
+                </privateKey>
+
```
AdSchellevis: empty newline

src/opnsense/mvc/app/views/OPNsense/IPsec/key_pairs.volt
AdSchellevis: use of `const` is fine indeed, we also have some of them in our libs.

src/etc/inc/plugins.inc.d/ipsec.inc
```diff
@@ -542,6 +544,22 @@ function ipsec_mobilekey_sort()
     });
 }
 
+function ipsec_keypairs()
```
AdSchellevis: it looks like `ipsec_keypairs()` is only used in src/www/vpn_ipsec_phase1.php, in which case I propose to move it to there.

src/opnsense/mvc/app/views/OPNsense/IPsec/key_pairs.volt
ppmathis: Thanks for confirming, @AdSchellevis!

src/etc/inc/plugins.inc.d/ipsec.inc
ppmathis: Updated, this function now resides in `src/www/vpn_ipsec_phase1.php` as proposed.

src/opnsense/mvc/app/models/OPNsense/IPsec/IPsec.xml
ppmathis: Removed.

src/opnsense/mvc/app/models/OPNsense/IPsec/IPsec.php
ppmathis: Updated, this exception now also makes use of the translation system. Good catch!

ppmathis:
Thank you @AdSchellevis for your code review as well. I have implemented the requested changes - the new MVC model code no longer makes use of any legacy code. Reconfiguring IPsec using `pluginctl` through `configd` works fine on my system, so I think this was a great idea AND it got rid of the hacky `error_reporting(...)` workaround. :)

The only other change I've made myself, aside from rebasing to master, was ensuring the "subsystem is dirty" message does not flicker when an exception occurs by checking if there has been a successful message at all before making these DOM changes.

src/opnsense/mvc/app/models/OPNsense/IPsec/Menu/Menu.xml
```diff
@@ -0,0 +1,7 @@
+<menu>
+    <VPN>
+        <IPsec>
+            <PublicKeys order="31" VisibleName="RSA Key Pairs" url="/ui/ipsec/key-pairs" />
```
fichtner: would you mind moving all IPsec menu items to this location? the IPsec ACL can be moved in a similar way.

ppmathis: Done, I have moved all the legacy ACL and menu entries into the MVC code in a separate commit, as other parts than described in this PR have been touched. Additionally, this revealed a 'bug' in the current code, where the ACL `page-status-systemlogs-ppp` has been mislabeled as `IPsec VPN`, which I also fixed in this commit.

AdSchellevis:
@ppmathis it looks good, thanks! I would like to give it a spin on my local machine before merging, at the moment I'm a bit too busy to do this, so I'll merge as soon as I have a bit of time available (maybe at the end of this week).

AdSchellevis:
@ppmathis its merged, small change in the volt template to fix the background, further than that it looked nice. Thanks again!

ppmathis:
Thanks to everyone who was involved with this PR, I am really happy with how the result turned out and the review process felt smooth and effective to me. :+1:

PR #3598, opened 2019-07-25
Don't working migration for Monit on 19.7.

kekek2:


AdSchellevis:
I don't think "validation will fail because we want to change the type of tests linked to services" is actually the case when transitioning from `1.0.6` --> `1.0.7` or from `1.0.7` --> `1.0.8`

PR #3600, opened 2019-07-26
Fix multibyte substring.

kekek2:
substr work incrorrect with UTF-8 description.


AdSchellevis:
almost a duplicate of https://github.com/opnsense/core/pull/3299, solution is the same as well....

Issue #3124, opened 2019-01-12
Feature request: Option to enable logging for all firewall rules at once


JasMan78:
When the firewall ruleset becomes bigger and bigger, and you've found out that an client has access to something that it shouldn't have, it's sometimes difficult to find the rule which allowes the traffic.

In this case it would be great to have an switch, which enables the logging for all rules at once, to identify the rule which allowes the traffic.

AdSchellevis:
closed with https://github.com/opnsense/core/pull/3605 thanks to @johnaheadley, "select all" didn't make it (see PR for discussion), but per item toggle is way easier now.

Nadahar:
I'd say that an "log all" option would be very helpful when trying to debug when something doesn't behave like you want it to. Easier toggling of the logging doesn't really help, you don't want to (at least I don't) "disturb" the normal configuration of what to log and not by turning on a lot of rules you normally don't want logged. You just want to see everything while debugging the issue, so that the "problem" doesn't elude you because some rule you didn't think of come into play.

fichtner:
It's not 2019 anymore and we had a couple of people flooding their logs. From my side that is still a 'no'. :)

Nadahar:
I see this as just as relevant in 2023. Is there another way to "see all that's happening" then? Otherwise I see no other solution than to find pen and paper, write down the current logging status of all the rules, enable them all - and then undo all that from the "paper backup" once done with debugging. It's not a very tempting solution.

fichtner:
"What" is happening is easy to debug with selective rules. The problem isn't the world. It's a sever not working. NAT not responding, etc.

AdSchellevis:
use the packet capture to follow the flow and read the ruleset based on what you're seeing there, always works

Nadahar:
@AdSchellevis I've tried to use the packet capture, but I'm a bit confused as to how that works. I find the packets coming in on one interface, but they never leave on the other. So, I assume that some rule, that isn't being logged, interferes. Maybe that's not the case, but it sure would be helpful to verify.

Is it correct to assume that the packet capture only captures what happens "outside" the firewall (before on incoming, after on outgoing)?

AdSchellevis:
if you're capturing all interfaces, traffic flows in, but never out, a filter rule is one of the most common issues, which you can then easily validate as Franco suggested with selective rules. you're seeking for tools either nog very useful or already available. best use our forum for discussions.

Nadahar:
[...]

PR #3605, opened 2019-07-27
Added logging toggle for firewall rules

johnaheadley:
This is a proposed solution for:

Feature request: Option to enable logging for all firewall rules at once #3124

I propose this as a superior solution to a single button that enables logging on all policies at once.

Having a logging toggle on each firewall rule allows a nice overview of what policies have logging enabled and which don't, as well as giving the admin the ability to quickly enable logging for all policies that don't currently have it enabled.

### DIFF
--- a/src/www/firewall_rules.php
+++ b/src/www/firewall_rules.php
@@ -164,13 +164,6 @@ function firewall_rule_item_icons($filterent)
           );
         }
     }
-    if (isset($filterent['log'])) {
-          $result .= sprintf(
-              "<i class=\"fa fa-info-circle fa-fw %s\"></i>",
-              !empty($filterent['disabled']) ? 'text-muted' : 'text-info'
-          );
-    }
-
     return $result;
 }
 
@@ -188,6 +181,19 @@ function firewall_rule_item_action($filterent)
         return "fa fa-play fa-fw text-success";
     } else {
         return "fa fa-play fa-fw text-muted";
+    }
+}
+
+function firewall_rule_item_log($filterent)
+{
+    if ($filterent['log'] == true && empty($filterent['disabled'])) {
+        return "fa fa-toggle-on text-success";
+    } elseif ($filterent['log'] == false && empty($filterent['disabled'])) {
+        return "fa fa-toggle-off text-danger";
+    } elseif ($filterent['log'] == true && !empty($filterent['disabled'])) {
+        return "fa fa-toggle-on text-muted";
+    } else {
+        return "fa fa-toggle-off text-muted";
     }
 }
 /***********************************************************************************************************
@@ -278,7 +284,18 @@ if ($_SERVER['REQUEST_METHOD'] === 'POST') {
         mark_subsystem_dirty('filter');
         header(url_safe('Location: /firewall_rules.php?if=%s', array($current_if)));
         exit;
-    }
+    } elseif (isset($pconfig['act']) && $pconfig['act'] == 'log' && isset($id)) {
+        // toggle logging
+        if(isset($a_filter[$id]['log'])) {
+            unset($a_filter[$id]['log']);
+        } else {
+            $a_filter[$id]['log'] = true;
+        }
+        write_config();
+        mark_subsystem_dirty('filter');
+        header(url_safe('Location: /firewall_rules.php?if=%s', array($current_if)));
+        exit;
+    }    
 }
 
 $selected_if = 'FloatingRules';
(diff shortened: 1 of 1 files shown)


AdSchellevis:
Hi @johnaheadley,

Thank you for your PR, I have some remarks about the proposed changes.

In stead of using a new icon, I would suggest keeping the original info icon, the info icon is more explicit about the feature that's enabled. 
image
Maybe something like:
image

Looking at https://github.com/opnsense/core/issues/3124, I would expect there's some kind of "enable/disable" selected button as well. Question is how this should look, currently you can enable/disable from the buttons down right, but it's already quite full there:

image

Best regards,

Ad

johnaheadley:
Hey @AdSchellevis,

Thanks for the feedback. The tricky part about keeping the old icon is how to handle differentiating between logging on/off when the rule is disabled. Here are two options I came up with:

Option 1:

Option1

Option 2:

Option2

Option 1 has the downside that the icon does not go text-muted when logging is enabled but rule is disabled.

Also, I can move the logging icon to the be the last icon displayed, like in your photo, if that is what you think looks best.

As far as the enable/disable logging on selected rules option, I agree that it is crowded in the bottom right, but I can't think of a better way to handle that. Here is what it could look like if we were to implement the icons in Option 2.

LogSelected

Once we work through the design choices then I can submit a new commit with the changes.

AdSchellevis:
@johnaheadley I would like to vote for option 1, when disabled, the state of logging is not very relevant anyway. The location of the item (last item, next to action) is less problematic, personally I can live with both and don't have a preference.

johnaheadley:
Alright, so the icons are now back to the fa-info-circle, and the coloring is consistent with the first match icon - regardless of the enable/disable status of the rule, the icon is text-info if logging is enabled or text-muted if not.

Screen Shot 2019-08-04 at 10 49 04 AM

This gets carried over to the newly added icons for the enable/disable logging on selected rules.

Screen Shot 2019-08-04 at 10 13 07 AM

Finally, this is consistent in the legend, where I tried to make the text reflect that the rule status doesn't affect the icon color.

Screen Shot 2019-08-04 at 10 13 59 AM

(Also, sorry for the bit of a mess with the commits. I am still learning git.)

fichtner:
FWIW, the log enable/disable for selected seems displaced.... we can't add buttons for all features here and the only one that is important enough is enable / disable the rule itself

johnaheadley:
Should I remove the enable/disable log for selected buttons and we just leave the toggle on the individual rules to satisfy the feature request?

Otherwise, we can move the buttons somewhere else in the footer bar so that it doesn't wrap around in the last column?

For example, bottom left

Screen Shot 2019-08-04 at 11 59 00 AM

fichtner:
I would opt for removal. While it's nice to enable all you will kill your "normal" logging state and you don't have immediate feedback from here anyway. It would be much cooler if we had a "overlay" live log feature where we can toggle away from the configuration to see what happens and disable logging selectively or collectively so you can "drill down" on your logging. But that's a much larger goal for later perhaps.

AdSchellevis:
Thinking about this, the main issue in the log toggle is the page reload, what if we toggle the icon state async and keep the page as is? I would be more user friendly, and the amount of work toggling a set of rules would be close to the same.

@johnaheadley sorry about the noise, I know I asked about the "toggle selected" thing, in retrospective I agree with @fichtner on this point. 

I can fix the Javascript code for the single toggle action, just let me know if you agree.

fichtner:
I love the "responsive" approach for the toggle. Nice idea.

johnaheadley:
Hey @AdSchellevis, no worries. I agree, please feel free to make any changes you like.

AdSchellevis:
@johnaheadley there you go https://github.com/opnsense/core/commit/c8b57d0bfecd1e654507007b2c39c4a4c1eacacb, thanks for your contribution!

johnaheadley:
@AdSchellevis, you're welcome! Also, I'm testing out the async toggle you added - it is a great user experience, but I see a small issue:

If I enable/disable the log status and/or the rule and then refresh the page, I get the message to "Apply changes". Was your goal to have these two actions not require a "Apply changes" button press?

AdSchellevis:
@johnaheadley hmm, it needs a reload to be effective, so we should probably show some button. Only small issue is that it will scroll your page down when showing the normal button.

AdSchellevis:
@johnaheadley can you try https://github.com/opnsense/core/commit/d2fc3c2567cf0959407b7d1da409110e3784aa46?

johnaheadley:
@AdSchellevis d2fc3c2 looks good!

PR #3610, opened 2019-07-29
Fix IPv4 dhcp-relay 

jayantsahtoe:
DHCP-relay service will fail with the following error if DHCP server IP is know in static route:

dhcrelay: Can't attach interface {ifname} to bpf device /dev/bpf0: Device not configured

This is the only place where get_real_interface( ) isn't used in IPv4 dhcp-relay


AdSchellevis:
@jayantsahtoe looks like a bug indeed, thanks for patching!

Issue #3568, opened 2019-07-12
OpenVPN server wizard always asks to create a new CA


int-red:
When you use the OpenVPN server wizard to create a new server you are always asked to
create a New Certificate Authority (CA) Certificate. 
You can not use an already existing CA. There should be an option to select an existing CA.

johnaheadley:
@fichtner can this be closed now?

AdSchellevis:
fixed in https://github.com/opnsense/core/pull/3614 thanks to @johnaheadley 👍

PR #3614, opened 2019-07-30
To fix: OpenVPN server wizard always asks to create a new CA

johnaheadley:
To fix: OpenVPN server wizard always asks to create a new CA #3568

Note: the wizard (and so my fix) currently restricts CAs for which the firewall has the private key. This is because if you choose to create a new server certificate in the next step, it will utilize the selected CA to sign the server certificate.

However, the CA certificate selected ends up filling the "Peer Certificate Authority" field of the OpenVPN Server, and the Peer Certificate Authority does not need to be a CA certificate with private key when you are not using the wizard. Some logic could be added in the future to the wizard to account for all the possible scenarios.

### DIFF
--- a/src/etc/inc/plugins.inc.d/openvpn/wizard.inc
+++ b/src/etc/inc/plugins.inc.d/openvpn/wizard.inc
@@ -202,8 +202,8 @@ function step6_stepbeforeformdisplay()
     else foreach ($config['ca'] as $ca) {
         if (!empty($ca['prv'])) {
             $no_internal_ca=false;
-        }
-        
+            break;
+        }   
     }
     
     if ($no_internal_ca) {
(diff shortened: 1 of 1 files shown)


src/etc/inc/plugins.inc.d/openvpn/wizard.inc
```diff
@@ -194,10 +194,21 @@ function step5_submitphpaction()
 function step6_stepbeforeformdisplay()
 {
     global $stepid, $config;
+    $no_internal_ca = true;
 
-    if (!empty($config['ca'])) {
+    if (empty($config['ca'])) {
         $stepid++;
     }
+    else foreach ($config['ca'] as $ca) {
```
AdSchellevis: plase check style, psr 2 would be something like: 
```
}  else {
   foreach ($config['ca'] as $ca) { 
   ....
  }
}
```

AdSchellevis:
@johnaheadley looks good, thanks!

johnaheadley:
Hey @AdSchellevis, I found an issue with this merged PR. I fixed [here](https://github.com/johnaheadley/core/commit/4f5d39d43749b38bf11c2e704cc5fd0fb4f3e738), just not sure how to handle it since this is merged already.

PR #3632, opened 2019-08-06
High Availablity: update helptext to reality

mimugmail:
I'd love to rename the label also but this would break all translations so at first only help text :)


mimugmail:
P.S.: of course I've verified it on a live system here ;)

AdSchellevis:
@mimugmail thanks! it does indeed transfer both:

https://github.com/opnsense/core/blob/b87761029672802f3bdfc6bb070a3ca71ff39143/src/etc/rc.filter_synchronize#L267

Issue #3617, opened 2019-07-31
OpenVPN wizard server certificate selection issue


johnaheadley:
**Description**
The OpenVPN Wizard does not filter out non-server certificates from the server certificate drop-down, even though this is a requirement when not using the wizard.

**Reproduction**
1. Begin OpenVPN Wizard (/wizard.php?xml=openvpn)
2. Select Local User Access
3. Proceed through Wizard until VPN: OpenVPN: Servers: Server Certificate Selection, where "Web GUI SSL certificate" will be incorrectly listed (see screenshot 1)
4. Finish Wizard
5. Edit newly created OpenVPN Server, make no changes, then try to save - you will receive error that "certificate Web GUI SSL certificate is not intended for server use" (see screenshot 2)

**Expected Behavior**
Based on the current restriction the OpenVPN Server configuration page, I expect the Wizard to only display certificates that are Server:YES in the selection drop-down.
 
(Though I don't think it should be required in the first place that extendedKeyUsage of server authentication be present in the cert, as long as we have the private key.)

**Screenshots**

![certificate-selection](https://user-images.githubusercontent.com/11711559/62181879-9953e300-b309-11e9-92ce-140ac161cfb9.png)

![save-error](https://user-images.githubusercontent.com/11711559/62181882-9f49c400-b309-11e9-8e61-e481cff56ed8.png)

**Environment**

OPNsense 19.7.1-amd64
FreeBSD 11.2-RELEASE-p11-HBSD
OpenSSL 1.0.2s 28 May 2019

johnaheadley:
It's probably also worth thinking about why the default "Web GUI SSL certificate" is not a server certificate in the first place.

PR #3634, opened 2019-08-07
OpenVPN wizard server cert check

johnaheadley:
This PR fixes the issue I described in #3617.

In the case that both server and non-server certs are present, I chose to have the wizard just filter out non-server certificates from the Server Certificate Selection dropdown.

This is in contrast to the vpn_openvpn_server.php which shows the user all certificates, let's you select a non-server certificate, but then displays an error when trying to save (screenshot of error in issue report.) We can echo this behavior if you think this is a better UX.

### DIFF
--- a/src/etc/inc/plugins.inc.d/openvpn/wizard.inc
+++ b/src/etc/inc/plugins.inc.d/openvpn/wizard.inc
@@ -206,7 +206,7 @@ function step6_stepbeforeformdisplay()
             }
         }
         if ($no_internal_ca) {
-        $stepid++;
+            $stepid++;
         }
     }
 }
@@ -276,7 +276,7 @@ function step8_stepbeforeformdisplay()
             }
         }
         if ($no_server_cert) {
-        $stepid++;
+            $stepid++;
         }
     }
 }
--- a/src/etc/inc/plugins.inc.d/openvpn/wizard.inc
+++ b/src/etc/inc/plugins.inc.d/openvpn/wizard.inc
@@ -205,10 +205,9 @@ function step6_stepbeforeformdisplay()
                 break;
             }
         }
-    }
-
-    if ($no_internal_ca) {
+        if ($no_internal_ca) {
         $stepid++;
+        }
     }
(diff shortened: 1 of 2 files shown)


src/etc/inc/plugins.inc.d/openvpn/wizard.inc
```diff
@@ -205,10 +205,9 @@ function step6_stepbeforeformdisplay()
                 break;
             }
         }
-    }
-
-    if ($no_internal_ca) {
+        if ($no_internal_ca) {
```
AdSchellevis: indent looks a bit odd here, if intended, the next line likely needs 4 spaces as well.

src/etc/inc/plugins.inc.d/openvpn/wizard.inc
```diff
@@ -265,10 +264,20 @@ function step7_submitphpaction()
...
+        $stepid++;
+    } else {
+        foreach ($config['cert'] as $cert) {
+            if (cert_get_purpose($cert['crt'])['server'] == 'Yes') {
+                $no_server_cert=false;
+                break;
+            }
+        }
+        if ($no_server_cert) {
         $stepid++;
```
AdSchellevis: ^^ probably needs an indent too.

AdSchellevis:
@johnaheadley Looks good apart from some minor psr remarks. I agree we can better filter the options in the wizard. The wizard itself isn't very great (code is quite messy), but is difficult to phase out now, it's always good to smoothen the edges a bit :) Thanks!

src/etc/inc/plugins.inc.d/openvpn/wizard.inc
johnaheadley: the if is part of the else now on line 201

AdSchellevis: thanks!

AdSchellevis:
@johnaheadley thanks!

PR #3642, opened 2019-08-13
Set absolute path to root.hints

h-town:
root.hints file is located at /var/unbound not /

Current users may not be leveraging root.hints with unbound deployments.

### DIFF
--- a/src/etc/inc/plugins.inc.d/unbound.inc
+++ b/src/etc/inc/plugins.inc.d/unbound.inc
@@ -282,7 +282,7 @@ chroot: /var/unbound
 username: unbound
 directory: /var/unbound
 pidfile: /var/run/unbound.pid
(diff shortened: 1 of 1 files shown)


h-town:
> Can you state the bug? I was under the impression this works post-chroot

hi fichtner, 
there may not be a bug, per se, but this is the only file in unbound.conf not being referenced with absolute path.  if the full path allows for function i see this as a cleaner approach.

the only concern i had is whether or not unbound continues to operate and function in the absence of root.hints after having ignored a mis-configured entry for the file.

thanks!

fichtner:
Ok, I will have to check before merge. Unbound documentation on chroot best practice is scarce.

fichtner:
Merged, thanks! Consolidation makes sense indeed.

Issue #3495, opened 2019-05-22
Monit: Add raw defives to interface list --> monitor lagg-members/vlan-parents


Kimotu:
[X] I have read the contributing guide lines at https://github.com/opnsense/core/blob/master/CONTRIBUTING.md

[X] I have searched the existing issues and I'm convinced that mine is new.

**Is your feature request related to a problem? Please describe.**
No problem, but monitoring improvment.

**Describe the solution you'd like**
ATM only the assigned interfaces are displayed as choice. lagg interfaces compensate network interruptions and thus can be overseen pretty easy if just one interface in a lagg bonding fails.
So a monitoring of the single members would be nice.

**Describe alternatives you've considered**
Not checked, but if monit supports config includes, then an empty dummy include file where you can write persistent manual checks would be an alternative.

Kimotu:
> INCLUDE FILES
> 
> The Monit control file, monitrc, can include additional configuration files. This feature helps one to organise configuration into separate files instead of having everything in one file, if you like this kind of thing. Include statements can be placed at virtually any place in monitrc though the convention is at the bottom. The syntax is the following:
> 
>   INCLUDE <globstring>
> 
> The globstring is any kind of string as defined in glob(7). Thus, you can refer to a single file or you can load several files at once. If you want to use whitespace in your string the globstring needs to be embedded into quotes (') or double quotes ("). If the globstring matches a directory instead of a file, it is silently ignored.
> 
> Any include statements in an included file are parsed as in the main control file.
> 
> If the globstring matches several results, the files are included in a non sorted manner. If you need to rely on a certain order, you should avoid wild-card globbing and instead specify the full path of files included.
> 
> An example,
> 
>  include /etc/monit.d/*.cfg
> 
> This will load any file matching the globstring. That is, all files in /etc/monit.d that ends with the prefix .cfg.

AdSchellevis:
solved in https://github.com/opnsense/core/pull/3643

PR #3643, opened 2019-08-13
monit: add parent devices to interface list

fbrendel:
Monitoring the physical interface really makes sense.
Solves #3495


AdSchellevis:
@fbrendel thanks!

Issue #3650, opened 2019-08-16
netflow: set active/inactive timeout


fbrendel:
I'd like to set the active/inactive timeouts for the [NetFlow](https://www.freebsd.org/cgi/man.cgi?query=ng_netflow&apropos=0&sektion=4&manpath=FreeBSD+11.2-RELEASE&arch=default&format=html) exporter.
The default active timeout of 30 minutes is to long and 15 seconds for inactive is to short for us.

AdSchellevis:
I'm not sure we could influence this easily with `ng_netflow`, maybe the man page could help you out https://www.freebsd.org/cgi/man.cgi?query=ng_netflow

To fiddle with settings, change `/usr/local/etc/rc.d/netflow` and reload should do the trick.

Netgraph isn't really easy unfortunately.

ps: would you be so kind to use our templates for feature requests in the future?

fbrendel:
> I'm not sure we could influence this easily with `ng_netflow`, maybe the man page could help you out https://www.freebsd.org/cgi/man.cgi?query=ng_netflow

It's easy to implement and I can do that.
I just made this feature request to be sure it's documented and no one is already working on it :wink: 
 
> ps: would you be so kind to use our templates for feature requests in the future?

Yes, of course. Was a bit too lazy to fill in the form for this little request. Sorry.

AdSchellevis:
Added in https://github.com/opnsense/core/pull/3651, for reference, default values remain unchanged with the new code (I couldn't find the defaults in the man page):

https://github.com/opnsense/src/blob/8d1a9170bd4d9d50ffeecc2c8366ec34b8ebdb02/sys/netgraph/netflow/netflow.h#L33-L34

PR #3651, opened 2019-08-16
netflow: set active/inactive timeout

fbrendel:
This PR adds active and inactive timout settings for the NetFlow exporter.
Solves #3650

### DIFF
--- a/src/opnsense/service/templates/OPNsense/Netflow/netflow.conf
+++ b/src/opnsense/service/templates/OPNsense/Netflow/netflow.conf
@@ -25,6 +25,6 @@ netflow_egress_only="{% if OPNsense.Netflow.capture.egress_only %}{% for interfa
 netflow_version="{%if OPNsense.Netflow.capture.version == 'v9' %}9{% else %}5{%endif%}"
 netflow_int_destination="127.0.0.1:2055"
 netflow_destinations="{{OPNsense.Netflow.capture.targets.replace(',', ' ')}}"
-netflow_active_timeout={{OPNsense.Netflow.activeTimeout}}
-netflow_inactive_timeout={{OPNsense.Netflow.inactiveTimeout}}
+netflow_active_timeout={{OPNsense.Netflow.activeTimeout|default(1800)}}
+netflow_inactive_timeout={{OPNsense.Netflow.inactiveTimeout|default(15)}}
 {% endif %}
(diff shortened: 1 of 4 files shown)


src/opnsense/service/templates/OPNsense/Netflow/netflow.conf
```diff
@@ -25,4 +25,6 @@ netflow_egress_only="{% if OPNsense.Netflow.capture.egress_only %}{% for interfa
 netflow_version="{%if OPNsense.Netflow.capture.version == 'v9' %}9{% else %}5{%endif%}"
 netflow_int_destination="127.0.0.1:2055"
 netflow_destinations="{{OPNsense.Netflow.capture.targets.replace(',', ' ')}}"
+netflow_active_timeout={{OPNsense.Netflow.activeTimeout}}
```
AdSchellevis: @fbrendel as a safety precaution it might be better to add `|default()` as well, since netgraph will fail with pretty annoying errors if one of the timeouts is empty. (migration should prevent this, but I'm a bit worried about odd errors down the road if people restore a config without running migrations).

AdSchellevis:
@fbrendel looks good, thanks, one small remark in the commit.

src/opnsense/service/templates/OPNsense/Netflow/netflow.conf
fbrendel: Ok, commit is on the way.

Regarding the "annoying errors", wouldn't it be better then to have single ngctl commands instead of a batch in the rc.d script?
https://github.com/opnsense/core/blob/ef8cd8021dc91f5c05c3808d2e73e1de5c339498/src/etc/rc.d/netflow#L87-L95

AdSchellevis: @fbrendel I haven't tried it myself, but it might be better for traceability indeed. Since this PR is rather small and focused, I rather keep the cleanup in a separate one. If you would like to test separate steps and create a PR afterwards, that would be nice.

fbrendel: @AdSchellevis ok, then we'll see us in another PR :smiley:

AdSchellevis: @fbrendel thanks!, I'll merge this one first.

PR #3658, opened 2019-08-19
netflow: improve error handling in rc.d script

fbrendel:
This PR splits the ngctl batch into single commands to have a better error handling.


AdSchellevis:
@fbrendel thanks, looks good, I'll keep the PR open for a while, so we can first release the other one.

Issue #3659, opened 2019-08-19
monit: default settings missing after OPNsense install


fbrendel:
[X] I have read the contributing guide lines at https://github.com/opnsense/core/blob/master/CONTRIBUTING.md

[X] I have searched the existing issues and I'm convinced that mine is new.

**Describe the bug**
After a fresh install the Monit configuration is empty.
It's entirely missing.

**To Reproduce**
1. Install OPNsense
2. Visit Services -> Monit -> Settings

**Expected behavior**
The Alert, Service and Service Tests Settings should contain predefined entries.

**Additional context**
It seems that the migration code for Monit is not run at install.

**Environment**
OPNsense 19.7

fbrendel:
During the update to 19.7.2 the migration is started but failed:
`config[54379]: [OPNsense\Monit\Monit:test.41a59b6c-a6c7-4e32-8958-2d0875e8f370.condition] Condition would change the type of the test but it is linked to a service.`

That should not happen on an empty config.
I'll take a closer look on the migration code.

fbrendel:
@AdSchellevis on a new install the model functions use the latest model version 1.0.8 for default settings. But the migration from M1_0_0 to M1_0_7 is not compatible due to the additional test type introduced in 1.0.8.
My solution would be to add the missing test types to the M1_0_[0-7] migrations.
What do you think?

AdSchellevis:
@fbrendel that might be the easiest fix, it's not very pretty, but I assume we can't fix the default easily in this case.

AdSchellevis:
closed with https://github.com/opnsense/core/pull/3661

PR #3661, opened 2019-08-20
monit: fix migration

fbrendel:
This PR adds missing test types to the migrations.
The migrations use the `post` method now instead of `run` to
avoid validation errors.
Solves #3659


AdSchellevis:
@fbrendel thanks, should do the trick indeed, I'll double check on my end after merge

PR #3665, opened 2019-08-23
Cosmetic changes to Tunables section

Northguy:
There is a note at the bottom of the tuneables which states that some changes are applied directly and some require a reboot. 

In my opinion a more logical notification would be to show that sentence in the Apply notice box in stead of at the bottom of this page. That location can be easily overlooked.

### DIFF
--- a/src/www/system_advanced_sysctl.php
+++ b/src/www/system_advanced_sysctl.php
@@ -183,7 +183,7 @@ $( document ).ready(function() {
             print_info_box($savemsg);
         }
         if (is_subsystem_dirty('sysctl') && ($act != "edit" )) {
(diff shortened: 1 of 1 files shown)


fichtner:
I agree with this, but we shouldn't kill two perfectly valid translations for a new string that is just a composition of the two. use "gettext(...) . ' ' . gettext(...)"

Northguy:
Hi Franco, my lack of knowledge plays a part here. Is it possible to concatenate two of those strings easily? If so, I will make an effort to provide you with a new commit

Northguy:
@fichtner I think I nailed it. Please check latest commits. Tested on my own system and it works.

fichtner:
Looks good to me, thank you ❤️

PR #3672, opened 2019-08-27
Translate action description.

kekek2:
https://github.com/opnsense/lang/pull/46


AdSchellevis:
not going to do this now, maybe at some point in the future when there's more demand (the lang parser needs work to function properly too)

PR #3676, opened 2019-08-29
Fix tunnable translation.

kekek2:
https://github.com/opnsense/lang/pull/47


fichtner:
It should be fixed for other strings in the future as well now. Editing specific strings or characters isn't a good solution here.

PR #3685, opened 2019-08-31
Add updatefreq for GeoIP aliases.

ErikJStaab:
Add updatefreq for GeoIP, config file is currently generated with no updatefreq; alias.expire() function called from update_tables.py returns False by default if no ttl is present. This creates a condition where unless you manually change the file to trigger alias.change() with no ttl present it will never trigger to update the GeoIP aliases. Tested that it sets updatefreq in config file and now GeoIP aliases will update automatically through update_tables.py.

### DIFF
--- a/src/opnsense/mvc/app/views/OPNsense/Firewall/alias.volt
+++ b/src/opnsense/mvc/app/views/OPNsense/Firewall/alias.volt
@@ -151,6 +151,7 @@
                 case 'geoip':
                     $("#alias_type_geoip").show();
                     $("#alias\\.proto").selectpicker('show');
+                    $("#row_alias\\.updatefreq").show();                
                     break;
                 case 'external':
                     break;


AdSchellevis:
I'm not sure, we either should set a default interval or add this indeed.

https://github.com/opnsense/core/blob/f3dad94b38d3fee09b47e35764a87e522bb5de8c/src/opnsense/service/templates/OPNsense/Filter/filter_tables.conf#L26-L30

Since the lists don't change more than once a day, a default would probably be enough, since specifying short lifetimes would suggest it would be fetched more often as well (In which case we should fix the part below as well).

https://github.com/opnsense/core/blob/3dc5f7150337434d838c79260278b148fa1afda6/src/opnsense/scripts/filter/lib/alias.py#L166

@ErikJStaab what do you think?

ErikJStaab:
Even if they put in a shorter duration than 1 day I believe the alias.py time check against the files would prevent it from fetching more often than once per day regardless of their setting. Once a day should be fine for most people if we just set a default for updatefreq to 1. Maybe someone out there has a use case where they would want it to update less frequently but I cant imagine it would be that big of an issue to download a few MB every day even if they are on a metered connection?

AdSchellevis:
Let's fixate to once a day then, I'll prepare a commit and close this PR.

ErikJStaab:
Great, thank you.

AdSchellevis:
@ErikJStaab can you test https://github.com/opnsense/core/commit/ae5692b477db1701491bce55db5317b0a73728c0 ?

To install, use the following in a console:
```
opnsense-patch ae5692b
```

ErikJStaab:
Rolled back my changes, created a new geoip alias, verified config has no updatefreq, touch file to change modification time to > 1 day old, ran update_tables.py and it looks like it updated, this fix should be good to go.

PR #3699, opened 2019-09-06
Update opnsense.xml

thomasjsn:
mirror.homelab.no has been moved from my basement to a datacenter, with new URL and owner. It is hosted on a 10 Gbit/s connection, with native IPv6, but only supports HTTP method. mirror.homelab.no currently redirects to the proper path on mirror.terrahost.no.


AdSchellevis:
@thomasjsn thanks, I've updated https://opnsense.org/download/ as well.

thomasjsn:
Excellent :)

Issue #2162, opened 2018-02-03
Firewall: Settings: Advanced - 'Aliases Resolve Interval' not working


petererdos:
At least for host(s), the GUI feature is not triggering the resolver for any given values (for example, 60). A cron job is needed for FQDNs to be resolved (as a workaround).

OPNsense 18.1.1-amd64
FreeBSD 11.1-RELEASE-p6
LibreSSL 2.6.4

AdSchellevis:
It should be in the crontab by default, can you grep the following:

` cat /var/cron/tabs/root | grep update_tables.py
`

it should contain:
`10	*	*	*	*	(/usr/local/opnsense/scripts/filter/update_tables.py) > /dev/null
`

petererdos:
It does, but whatever the value I enter in the GUI, it will always execute every 10 minutes, it does not reflect my preference:

10      *       *       *       *       (/usr/local/opnsense/scripts/filter/update_tables.py) > /dev/null

AdSchellevis:
that's correct, the alias update invalidates content after the specified amount of time, so it can run safely every 10 minutes.

fichtner:
Er... 10 means every hours at 10 minutes, not every ten minutes... should it be */10 then?

AdSchellevis:
oops, missed that, every hour could indeed be a bit to long... (worst case an hour interval could update every 2 hours ;) ).

petererdos:
To _really_ use seconds instead of hours brings much more value for this GUI feature, at least for me :)

AdSchellevis:
Where are the seconds? It should be hours, the table heading contains:
![image](https://user-images.githubusercontent.com/9783985/35914406-76f652e2-0c03-11e8-918d-4b819780659f.png)

Usually it's not a very good idea to fetch content very often, for specific cases it might be different, maybe we could support fractions of hours (e.g. 0.25, 0.1). What's your use-case for this?

petererdos:
![screenshot_5](https://user-images.githubusercontent.com/14872721/35916921-be0cb7aa-0c14-11e8-9c81-128cd1880a0d.png)

Use case: 
- (phone  & laptop) dynamic hostname resolution via aliases
-  i'm using mobile data often when i'm on the road, and the service provider changes ips frequently
- i'm using aliases in my firewall rules to permit connections only from those aliases, to allow access to certain services in my lan (port forwards mostly)

AdSchellevis:
[...]

PR #3702, opened 2019-09-07
filter/alias, update_tables.py slow with large alias lists

ErikJStaab:
Avoid reading entire table for url/urltable/geoip output from pfctl on every execution  …
Reading all of the tables every minute this script runs can cause higher CPU utilization with larger tables. Url, urltable, geoip tables will only be updated once when created or if the file modified time hits expiration. They should never be updated manually so no comparison between loaded and on disk tables should be needed unless I am missing something. 

I have looked at Issue #2162 and https://github.com/opnsense/core/commit/c5555b2ebc4c2285fb40d3a1a22c1966820bc64e and I see why update_tables.py is set to run every minute. From the little bit of profiling I did it seems like much of the time is still from python importing libraries so I don't know if there is any better way around that without investing a lot more time. Of course the profiling takes forever compared to normal execution but in that I also went from around 212 seconds down to 82 with both taking around 72 seconds of that to even reach class AliasParser(object).

In real use with ~250k lines in my aliases on an apu2, before this change it takes 11.7s, after it takes 8.7s.

19.7.3 stock: time /usr/local/bin/flock -n -E 0 -o /tmp/filter_update_tables.lock /usr/local/opnsense/scripts/filter/update_tables.py
{"status": "ok"}
8.244u 3.378s 0:11.72 99.0%     69+579k 32+45io 0pf+0w

After changes: time /usr/local/bin/flock -n -E 0 -o /tmp/filter_update_tables.lock /usr/local/opnsense/scripts/filter/update_tables.py
{"status": "ok"}
5.664u 2.978s 0:08.68 99.4%     10+208k 26+0io 0pf+0w

I tested adding/removing/enable/disable urltable and geoip tables as well as touching the files to make them expire and it seems to be working for me. I applied patch 8678567 so I could make changes against what is currently in master. Thoughts?

### DIFF
--- a/src/opnsense/scripts/filter/update_tables.py
+++ b/src/opnsense/scripts/filter/update_tables.py
@@ -149,8 +149,6 @@ if __name__ == '__main__':
                     alias_pf_content.append(line)
             if len(alias_content) != len(alias_pf_content):
                 alias_changed_or_expired = True
-        # Need to check that the url, urltable, or geoip tables are currently loaded. If not loaded for any reason
-        # this should flag the alias_changed_or_expired to cause them to load.
         else:   
             loaded_tables = list()
             sp = subprocess.run(['/sbin/pfctl', '-sT'], capture_output=True, text=True)
--- a/src/opnsense/scripts/filter/update_tables.py
+++ b/src/opnsense/scripts/filter/update_tables.py
@@ -140,14 +140,28 @@ if __name__ == '__main__':
         else:
             alias_content_txt = ""
 
-        alias_pf_content = list()
-        sp = subprocess.run(['/sbin/pfctl', '-t', alias_name, '-T', 'show'], capture_output=True, text=True)
-        for line in sp.stdout.strip().split('\n'):
-            line = line.strip()
-            if line:
-                alias_pf_content.append(line)
+        if alias.get_type() not in ['url', 'urltable', 'geoip']:
+            alias_pf_content = list()
+            sp = subprocess.run(['/sbin/pfctl', '-t', alias_name, '-T', 'show'], capture_output=True, text=True)
+            for line in sp.stdout.strip().split('\n'):
+                line = line.strip()
+                if line:
+                    alias_pf_content.append(line)
+            if len(alias_content) != len(alias_pf_content):
+                alias_changed_or_expired = True
+        # Need to check that the url, urltable, or geoip tables are currently loaded. If not loaded for any reason
+        # this should flag the alias_changed_or_expired to cause them to load.
+        else:   
+            loaded_tables = list()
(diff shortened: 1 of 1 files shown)


AdSchellevis:
@ErikJStaab as I see it, this would functionally change the nature of the function and could falsely assume that content didn't change over time and would likely lead to fussy tickets (missing host changes and new/changed source entries).

Unfortunately it costs a bit of time to process these long lists on small devices, I couldn't find much improvements myself at the time without changing functionality (keep system in sync with defined data). It might be an idea to time how long it would take to dump all alias contents using pfctl in comparison to update_tables, if the difference is large, there might be other options to improve speed. 

I'm a bit hesitant to change the nature of the function at the moment, since in the past we have seen a lot of weird issues when code tries to cut off corners, which often is hard to debug.

ErikJStaab:
pfctl takes around 1.6sec to dump all tables, assuming the line stripping takes nearly the same amount of time that makes up the remainder of the 3s gained in my use case, that looks like about 50/50. I think you are probably right about not being able to do anything else while confirming everything is synchronized. 

What I am gathering is that the main concern is that the alias_changed_or_expired = max(alias.changed(), alias.expired()) may be less than reliable in some cases so you don't want to rely solely on that? Or have there been issues with pf tables consistency and changing over time? I agree that this would put a lot of trust in those two things to work properly.

Only other options I see would require more work to separate update process for url, urltable, geoip separately at a different interval than manual host and network tables. It looks like it used to be every 10 minutes and it was fine except for people using hostnames in aliases taking too long to update. I am not sure that would be worth the work/complexity anyway. I am assuming people that have more/longer alias tables are probably also using faster hardware as you said so it should be less of an issue as well.

Thanks for the feedback and reviewing so quickly.

ErikJStaab:
Another thought, would it be preferred then to run these maintenance/back end scripts and other processes that fire off through cron or config changes with nice to lower the priority? Would rctl also be useful? Wouldn't that keep those processes from contending with more important things like forwarding packets or things that have user interaction to make it feel smoother?

AdSchellevis:
The usual update is triggered using cron, you could try to add a nice here
https://github.com/opnsense/core/blob/29b383c33b40196177e171d1575109ba684b6767/src/etc/inc/plugins.inc.d/pf.inc#L79-L83

Or is you issue that the apply takes longer to process now?

fichtner:
Mentions of nice(1) have been removed over the years as they are not overly effective and tend to mask sloppy behaviour. I would prefer a more sensible approach instead of using circumstantial workarounds which may or may not apply to others now or in the future.

ErikJStaab:
Thanks for your input everyone, closing PR.

PR #3721, opened 2019-09-23
devd does not pick up on laggX_vlanXXXX devices...

stumbaumr:
...since the underscore devices are filtered out.

### DIFF
--- a/src/etc/devd/carp.conf
+++ b/src/etc/devd/carp.conf
@@ -5,7 +5,7 @@
 
 notify 101 {
     match "system"          "CARP";
-    match "subsystem"       "[0-9]+@[0-9a-z]+";
+    match "subsystem"       "[0-9]+@[0-9a-z_]+";
     match "type"            "(MASTER|BACKUP)";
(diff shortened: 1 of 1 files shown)


src/etc/devd/carp.conf
```diff
@@ -5,7 +5,7 @@
 
 notify 101 {
     match "system"          "CARP";
-    match "subsystem"       "[0-9]+@[0-9a-z]+";
+    match "subsystem"       "[0-9]+@[0-9a-z_]+";
```
fichtner:
```suggestion
    match "subsystem"       "[0-9]+@[0-9a-z_.]+";
```

fichtner:
Looks like this was added last year mimicking FreeBSD scripting 0008970cd but you're right about it since we use "_" as an additional character. "." might be interesting sooner or later for VLAN so we should add it as well while here.

stumbaumr:
Good morning Franco and thanks for picking this up so fast.

I can only confirm and test my use case. If you know that the dot is an upcoming issue just add that as well.

fichtner:
Merged, thanks!

PR #3735, opened 2019-09-29
Automatically show hidden advanced values

Northguy:
Advanced features are hidden and can be 'unhidden' when a button is pushed. This is logical to hide all empty/default fields of the advanced features, but does not make sense when Advanced fields are actually filled with data. One would expect that data to be seen at once when opening a page.
 
Suggestion is to unhide hidden field data whenever there is data present in one of the hidden fields. See suggested code. Might be that there is another /easier way to apply this to all hidden fields generically?

### DIFF
--- a/src/www/services_dhcp.php
+++ b/src/www/services_dhcp.php
@@ -929,7 +929,7 @@ include("head.inc");
                       <td><i class="fa fa-info-circle text-muted"></i> <?=gettext("MAC Address Control");?></td>
                       <td>
 <?php
-                        if ((empty($pconfig['mac_allow']) && (empty($pconfig['mac_deny'])) {;?>
+                        if ((empty($pconfig['mac_allow']) && (empty($pconfig['mac_deny'])) {?>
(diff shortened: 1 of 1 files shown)


fichtner:
Some pages do this already. The issue is that the integration and upkeep is time consuming for static pages. For MVC we have a sticky advanced toggle. It might be better to implement this instead in static pages, but I am really not against merging these types of patches.

What do you think?

Northguy:
OK, I figured above solution out by myself with some trial and error, but am not that proficient in finding complete new ways of working. Any ideas how to implement sticky advanced toggle in the static pages? 
If you are willing to merge, my proposal is that I finish this specific html page with proposed solution for all advanced feature boxes, unless there is some clear code snippet I can re-use for an alternative.

AdSchellevis:
timeout

AdSchellevis:
@fichtner I accidentally missed that you where assigned to this issue, but it seems I can't reopen the PR (since parent is gone)

PR #3745, opened 2019-10-03
Allow self signed SSL certificate for Nextcloud backup

deajan:
Since a lot of people use x509 certificates, I added an option to allow self signed certificates.
This could still be improved by adding `CURLOPT__SSL_VERIFYHOST` to line 294 in order to allow non FQDN hostnames, but I guess that's not the same option.

Here's the story behind my PR:

Using OPNsense 19.7.4_1-amd64, my Nextcloud backups suddently stopped working with the following error:
```

Oct 3 09:10:57 | config[87666]: {"url":"https:\/\/cloud.mynextcloud.tld\/remote.php\/webdav\/files\/myuser\/","content_type":null,"http_code":0,"header_size":0,"request_size":0,"filetime":-1,"ssl_verify_result":20,"redirect_count":0,"total_time":7.324464,"namelookup_time":7.192147,"connect_time":7.238916,"pretransfer_time":0,"size_upload":0,"size_download":0,"speed_download":0,"speed_upload":0,"download_content_length":-1,"upload_content_length":-1,"starttransfer_time":0,"redirect_time":0,"redirect_url":"","primary_ip":"XX.XX.XX.XX","certinfo":[],"primary_port":443,"local_ip":"192.168.1.1","local_port":18185}
-- | --
Oct 3 09:10:57 | config[87666]: Error while fetching filelist from Nextcloud
```

After investigation, it seems that the CA list that comes with FreeBSD doesn't like my SSL certificate which I recently changed.
My new certificate is indeed valid in all the browsers I use.

On the opnsense box console, using `curl https://cloud.mynextcloud.tld` fails with`curl: (60) SSL certificate problem: unable to get local issuer certificate`

Of course using `curl https://cloud.mynextcloud.tld -k` works.

My SSL certificate chain is Sectigo > GoGetSSL RSA DV CA (I can provide my certificate chain on demand).
After having checked around, this CA is not only unrecognized FreeBSD, but also current Linux distributions.
Anyway, I could change my SSL certificate to something more "known" to leverage my issue.

Since this could affect a lot of people (using valid certificates without the root certificate being in FreeBSD trust store, or using self signed certificates), here's my PR :)

Best regards to all the OPNsense team.

### DIFF
--- a/src/opnsense/mvc/app/library/OPNsense/Backup/Nextcloud.php
+++ b/src/opnsense/mvc/app/library/OPNsense/Backup/Nextcloud.php
@@ -59,6 +59,12 @@ class Nextcloud extends Base implements IBackupProvider
                 "value" => null
             ),
             array(
+                "name" => "ssl_self_signed",
+                "type" => "checkbox",
+                "label" => gettext("Allow self signed"),
+                "value" => null,
+            ),
+            array(
                 "name" => "user",
                 "type" => "text",
                 "label" => gettext("User Name"),
@@ -132,6 +138,7 @@ class Nextcloud extends Base implements IBackupProvider
         if ($cnf->isValid() && !empty((string)$nextcloud->enabled)) {
             $config = $cnf->object();
             $url = (string)$nextcloud->url;
+            $ssl_self_signed = (string)$nextcloud->ssl_self_signed;
             $username = (string)$nextcloud->user;
             $password = (string)$nextcloud->password;
             $backupdir = (string)$nextcloud->backupdir;
@@ -144,12 +151,13 @@ class Nextcloud extends Base implements IBackupProvider
                 $confdata = $this->encrypt($confdata, $crypto_password);
             }
             try {
-                $directories = $this->listFiles($url, $username, $password, '/');
+                $directories = $this->listFiles($url, $ssl_self_signed, $username, $password, '/');
                 if (!in_array("/$backupdir/", $directories)) {
-                    $this->create_directory($url, $username, $password, $backupdir);
+                    $this->create_directory($url, $ssl_self_signed, $username, $password, $backupdir);
                 }
                 $this->upload_file_content(
                     $url,
+                    $ssl_self_signed,
                     $username,
                     $password,
                     $backupdir,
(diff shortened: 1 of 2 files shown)


jpawlowski:
It might be better to import your local CA or even the self-signed certificate itself to the local trust store. The firewall will then be able to officially trust your Nextcloud instance instead of just ignoring the certificate completely.

deajan:
Indeed importing the root CA into my firewall's trust store is a way to resolve my issue.
Still I think this PR makes it easier doing tests for people with x509 certificates.

jpawlowski:
There is a huge variety of services and everything will need to be adjusted individually.
Using the trust store is not only the safest way, it also works independently.

deajan:
Understood that the trust store is the production way to go.
Nevertheless, an open source firewall is often in homelabs and people won't automagically think of the truststore.

Is this PR worthy or should I just close and forget ? After all, it does enhance functionnality without any backwards compatibility issues or breaks.

jpawlowski:
I am not the maintainer, wait for his reply.

fabianfrz:
I oppose as the original developer but the last words is at the core maintainers.

Reason: People tend to do stupid things.
For example the backup is configured without a password, than you can gain access to almost the entire firewall configuration and extract some passwords. Such a checkbox just asks for failing regarding security and before somebody fixes his/her TLS, they enable that checkbox I guess. If you do not want the core features (ensure that you contacted the right server and then encrypt securely using a secure handshake) of TLS, do not use it and switch to plain HTTP. Just use a trusted CA.

@fichtner @AdSchellevis please add a comment as well.

fichtner:
CURLOPT_SSL_VERIFYPEER is for CN / SAN mismatch, so the whole option and presentation in the GUI is misleading.

fabianfrz:
@fichtner  no, also other validity like the time range or the issuer.

fichtner:
Ah, my mistake. So no checks on the peer certificate at all. :/

deajan:
I understand, OPNSense team does not want self signed certificate support for backups.

For the sake of sanity, I'd suggest OPNsense team also removes support for self signed certificates for Mailer backups on the same configuration page.

AdSchellevis:
@deajan you can always ask the contributor, plugins are less strictly managed than core (and not supported by the core team).

deajan:
@AdSchellevis No problem. I honestly only tried to improve functionnality of the product by providing that PR. Just thought that the product should remain consistent, allowing self signed certificates for mail backups but not for webdav backups makes no sense to me :)
Anyway I'll just close the PR and stick with importing my cert to the CA trust store.

Best regards.

fichtner:
Using 19.7.4 importing a self-signed certificate under system: trust: authorities adds the self-signed cert to the trust store, which in turn should allow you to connect via proper validation. Isn't that so?

deajan:
@fichtner No problem, I stay happy using OPNsense that way :)

PR #3754, opened 2019-10-10
Hide leaking hostname on SSH password auth

sooslaca:
As of today sshd prompts hostname on client side (leaking info):
`Password for root@OPNsense.private.:`

Proposal is to change it to be only:
`Password:`


AdSchellevis:
@sooslaca thanks, sounds reasonable.

PR #3755, opened 2019-10-10
captive portal/ipsec: add description to restart actions

fraenki:
This adds descriptions to the restart action of Captive Portal and IPSec. Both services may use a Let's Encrypt certificate.

This is necessary in order to setup an automation in our Let's Encrypt plugin to automatically restart the service when the certificate was renewed:

https://github.com/opnsense/plugins/blob/a878826c05ff470ab2a5001cf9f26b153a1f7bcf/security/acme-client/src/opnsense/mvc/app/models/OPNsense/AcmeClient/AcmeClient.xml#L860-L866

Without a description the action is not visible in the GUI (for good reasons).

refs https://github.com/opnsense/plugins/issues/1534

### DIFF
--- a/src/opnsense/service/conf/actions.d/actions_captiveportal.conf
+++ b/src/opnsense/service/conf/actions.d/actions_captiveportal.conf
@@ -39,6 +39,7 @@ command:/usr/local/etc/rc.d/captiveportal restart
 parameters:
 type:script
 message:restarting captiveportal services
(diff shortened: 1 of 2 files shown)


AdSchellevis:
@fraenki I don't see any concerns, it would make them visible in cron too, which is ok as far as I'm concerned.

PR #3758, opened 2019-10-11
DHCP Client -> Option Modifiers breaking comma

sooslaca:
When I would like to write an option modifier statement like:
`ignore domain-name-servers, host-name, domain-name, routers;`
it's broken into multiple lines, cause dhclient not start due to wrong configuration.

Result I got in /var/etc/dhclient_wan.conf:
```
  ignore domain-name-servers;
  host-name;
  domain-name;
  routers;;
```

What's expected is to not break on comma, only on semicolon.
hence config will look like as expected:
```
   ignore domain-name-servers, host-name, domain-name, routers;;
```

### DIFF
--- a/src/etc/inc/interfaces.inc
+++ b/src/etc/inc/interfaces.inc
@@ -3344,7 +3344,7 @@ function DHCP_Config_File_Advanced($interface, $wancfg, $wanif)
 
     $option_modifiers = "";
     if ($wancfg['adv_dhcp_option_modifiers'] != '') {
-        $modifiers = preg_split('/\s*,\s*(?=(?:[^"]*"[^"]*")*[^"]*$)/', $wancfg['adv_dhcp_option_modifiers']);
+        $modifiers = preg_split('/\s*;\s*(?=(?:[^"]*"[^"]*")*[^"]*$)/', $wancfg['adv_dhcp_option_modifiers']);
         foreach ($modifiers as $modifier) {
             $option_modifiers .= "\t" . $modifier . ";\n";
         }


NOYB:
When I originally added the DHCP advanced options section way back when. . . The intention was to place each "modifier option declaration" on it's own line in the config file.  Thus the syntax of `[modifier option declaration [, . . .]]`

So for your example it would be...
`ignore domain-name-servers, ignore host-name, ignore domain-name, ignore routers`

Which should produce this in the config file. . .
```
ignore domain-name-servers;
ignore host-name;
ignore domain-name;
ignore routers;
```
Placing each on it's on line is not functionality necessary and I have no preference either way.  But if going to change to the semicolon delimiter then the syntax example should be adjusted accordingly.

Also changing to the semicolon delimiter will potentially break existing configurations that have multiple option modifiers specified.  As they will be placed on a single line like so. `modifier option declaration, modifier option declaration,  . . .`   Which may be a violation of the syntax. `modifier { [option declaration] [, . . . option declaration] }`
  So a config update should also be done.
https://www.freebsd.org/cgi/man.cgi?query=dhclient.conf&sektion=5#OPTION_MODIFIERS

Is `ignore` valid?  Don't see it in the FreeBSD manual page.
https://www.freebsd.org/cgi/man.cgi?query=dhclient.conf&sektion=5

sooslaca:
Interesting, while freebsd dhclient.conf man does not mention it openbsd's mention
[https://man.openbsd.org/dhclient.conf.5](https://man.openbsd.org/dhclient.conf.5)

and is in the source code as well:
[https://github.com/opnsense/src/blob/abe639ab0b0fbaa45518e9a25d8cf2a0cd68995a/sbin/dhclient/dhcpd.h#L170](https://github.com/opnsense/src/blob/abe639ab0b0fbaa45518e9a25d8cf2a0cd68995a/sbin/dhclient/dhcpd.h#L170)

and working for me.

What my goal is to ignore everything except IP address and mask. With `require` it did not work hence I started to use `ignore`.

Unfortunately when I break `ignore` into multiple lines, one line overriding the other, only the last matter.

On the other hand got your point reg existing configurations where people are using it with comma.
Tough one, maybe I just gonna live with patching the code on updates.

NOYB:
Thought that might be the case that only the last of a each modifier would be taken.  The man page example makes no mention of this though and in the example shows send options broken out to multiple lines so was hoping the modifiers would work the same way.  But since they don't it should probably be fixed.

Probably isn't very many existing configs using multiple modifiers and if they are they probably aren't working correctly anyway.  Even so all a config upgrade script should need to do is replace the delimiter commas with semicolons.  The existing regex could probably do that.

AdSchellevis:
timeout

PR #3788, opened 2019-11-02
componentize some things starting with vxlan

fabianfrz:
@AdSchellevis please give it a try.

If we put more functionality in it, we do not need to script everything over and over again.


src/www/csrf.inc
```diff
@@ -96,13 +96,11 @@ class LegacyCSRF
             // csrf token for Ajax type requests
             $script = "
             <script>
-              $( document ).ready(function() {
```
AdSchellevis: is this a requirement for the `customElements.define(....)` ? at a first glance this changes loading behaviour, but maybe I'm mistaken here. If it's not related, we should probably keep it as is now.

src/opnsense/www/js/opnsense_bootgrid_plugin.js
```diff
@@ -455,3 +455,32 @@ $.fn.UIBootgrid = function (params) {
...
+            let val = that.dataset[value];
+            if (val) {
+                opts[value] = base + val;
+            }
+        });
+        $(this).UIBootgrid(opts);
+    }
+}
+
+customElements.define('os-bootgrid-table', OPNsenseBasicBootGridTable, {"extends": 'table'});
```
AdSchellevis: browser support might be problematic, although I haven't tested this (https://developer.microsoft.com/en-us/microsoft-edge/platform/status/customelements/)

AdSchellevis:
@fabianfrz although I'm not against simplifying things here, but we also need to think about future work before changing these things (refactoring current cases, change documentation, perhaps add documentation for these components as well).

One of the main concerns I have at the moment is about the support for the `customElements` tags (the construct is nice) and how the api of our components look like (options are a bit scattered over the place, but that would be fixable eventually if we can remove the old UIBootgrid construct at some point in the future). 

I don't have a lot of time available at the moment, if we are going to add this to core, we should have a plan how we are going to fix the rest of the work needed to keep things consistent.

The advantage of the current method is that it's quite easy to read for people building templates for the first time.

Our javascript codebase is far from perfect, but it's not too inconsistent at the moment.

src/www/csrf.inc
fabianfrz: I get a race condition. The element will initialize when it is ready by itself. At that point the full document is not yet loaded. This way we do not waste time waiting for unnecessary elements being loaded.

Since the boot grid search will be executed before the document is loaded, the post call is missing the csrf token andthe token BTW. does not depend on the DOM being constructed. So it should not be delayed.

src/opnsense/www/js/opnsense_bootgrid_plugin.js
fabianfrz: I thought edgeHTML is/was dropped for the chromium engine?

I tested it under Linux with Firefox 70 which also tells that it is experimental but works quite well.
The Mozilla page says they work with edge 76:
https://developer.mozilla.org/en-US/docs/Web/Web_Components/Using_custom_elements

fabianfrz: The more problematic support would be safari from the Mozilla link.

fabianfrz:
@AdSchellevis I would keep the old constructor for some advanced usage like custom cell formatter and so on but the default use case should work without additional JS coding.

In development the browser helps quite well. It says custom element beside the tag in Firefox and if you click on it, it will jump to the constructor. For a beginner I think it is easier to copy a custom tag than copying HTML resources and hope that the UI will not do strange things for example due to duplicated identifiers.

In theory we could also combine them with templates: https://developer.mozilla.org/en-US/docs/Web/HTML/Element/template

This is also a candidate I would like to get rid of:
https://github.com/opnsense/plugins/blob/master/www/nginx/src/opnsense/mvc/app/views/OPNsense/Nginx/index.volt#L90-L99

AdSchellevis:
at the moment I don't have the time available to refactor all existing usages (including the docs), we'll keep this open for now to avoid queuing a lot of work in master which we're not able to pickup within reasonable time.

AdSchellevis:
Hi Fabian, 

I tried the code, but it doesn't work. bootgrid isn't properly initialised. 
(new code)
image
(original)
image

To be honest, the gain on the bootgrid part isn't very huge either,  the current wrapper already doesn't contain a lot of extra's.

Given our current codebase, it might be a good idea to write a wrapper for the reload action, since there's more to gain in that spot. I guess it would be more consistent to opt for a jQuery prototype definition in that case `$("#my_button").ReloadButton({'endpoint':'...'});`

Best regards,

Ad

fabianfrz:
@AdSchellevis works as well but needs JS to initialize. At least it should avoid code duplication.

PR #3804, opened 2019-11-08
Firewall Rules Edit:  Input validation on interface and ipprotocol fields

stilez:
The current page doesn't validate any of these fields. Additionally, it doesn't check consistency in the different places that the interface can be received as data before acting on it ($_GET['if'], $_POST['floating'] and $_POST['interface']). 

Changes:

- grab valid if's into $interfaces[] so there's a single correct list for the rest of the code, which is also the list displayed in the form
- validate $_GET['if'] is acceptable and note this value applies equally to GET+POST, not just GET
- validate that after POST, $_POST['interface'] is non-empty and contains only valid interfaces
- validate that on POST, $_GET['if'], $_POST['floating'] and $_POST['interface'] are all consistent
- if the edit is to a specific interface (not "floating rules") then by definition we don't need to ask the interface - it's already in the URL. Any change would simply lead to invalid data due to inconsistency. Remove the dropdown box and replace with a hidden field containing the static interface value.
- validate on POST that ipprotocol is one of inet/inet6/inet46.

### DIFF
--- a/src/www/firewall_rules_edit.php
+++ b/src/www/firewall_rules_edit.php
@@ -242,13 +242,13 @@ if ($_SERVER['REQUEST_METHOD'] === 'GET') {
             // Note that the last condition inherently checks $interface_from_url is a valid interface.
             $input_errors[] = gettext("The interface selected has been modified and the page must be reloaded. Please click 'cancel', and re-open this page.");
         }
-        if ($interface_from_url == "FloatingRules" xor empty($pconfig['floating'])) {
+        if ($interface_from_url == "FloatingRules" xor !empty($pconfig['floating'])) {
             // 'if=FloatingRules' in the URL isn't consistent with the 'floating' field in $_POST. Should both be present or both absent.
             $input_errors[] = gettext("The interface selected has been modified and the page must be reloaded. Please click 'cancel', and re-open this page.");
         }
 
         // Validate ipprotocol
-        if (!in_array($ipprotocol, $ipprotocols)) {
+        if (!in_array($pconfig['ipprotocol'], $ipprotocols)) {
             $input_errors[] = gettext('A valid IP Protocol must be selected.');
         }
 
--- a/src/www/firewall_rules_edit.php
+++ b/src/www/firewall_rules_edit.php
@@ -571,7 +571,7 @@ if ($_SERVER['REQUEST_METHOD'] === 'GET') {
         mark_subsystem_dirty('filter');
 
         header(url_safe('Location: /firewall_rules.php?if=%s', array(
-            !empty($pconfig['floating']) ? 'FloatingRules' : $pconfig['interface']
+            !empty($pconfig['floating']) ? 'FloatingRules' : $interface_from_url
         )));
         exit;
     }
@@ -1629,7 +1629,7 @@ include("head.inc");
                       <td>&nbsp;</td>
                       <td>
                         <input name="Submit" type="submit" class="btn btn-primary" value="<?=html_safe(gettext('Save')); ?>" />
-                        <input type="button" class="btn btn-default" value="<?=html_safe(gettext('Cancel'));?>" onclick="window.location.href='/firewall_rules.php?if=<?= !empty($pconfig['floating']) ? 'FloatingRules' : $pconfig['interface'] ?>'" />
+                        <input type="button" class="btn btn-default" value="<?=html_safe(gettext('Cancel'));?>" onclick="window.location.href='/firewall_rules.php?if=<?= !empty($pconfig['floating']) ? 'FloatingRules' : $interface_from_url ?>'" />
                       </td>
                     </tr>
                   </table>
--- a/src/www/firewall_rules_edit.php
+++ b/src/www/firewall_rules_edit.php
@@ -40,8 +40,9 @@ $ostypes = json_decode(configd_run('filter list osfp json'));
 if ($ostypes == null) {
     $ostypes = array();
 }
+$ipprotocols = array('inet', 'inet6', 'inet46');
 $gateways = new \OPNsense\Routing\Gateways(legacy_interfaces_details());
-
+$interfaces = legacy_config_get_interfaces(array("enable" => true));
 
 /**
  * check if advanced options are set on selected element
@@ -74,9 +75,10 @@ function is_posnumericint($arg) {
     return (is_numericint($arg) && $arg[0] != '0' && $arg > 0);
 }
 
-
 $a_filter = &config_read_array('filter', 'rule');
 
+// Whether POST or GET, the URL should still have a valid $_GET['id'] field controlling the interface edited
+$interface_from_url = (empty($_GET['if']) || !is_string($_GET['if'])) ? "" : $_GET['if'];
 
 if ($_SERVER['REQUEST_METHOD'] === 'GET') {
     // input record id, if valid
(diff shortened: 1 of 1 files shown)


stilez:
Screenshot to underline the issue. 

The current code for that page has zero input validation. Literally.  If the POST request contains shell/HTML, that's what'll be stored (unescaped) in the config and used when `pf` restarts.
 
This is saved config after doing just that:
 
opnsense6

This PR fixes it in this page, but raises the question of if this is the only page with unvalidated input. Seems likely not.  I'd like to work on this (input validation/sanitising is a hot interest of mine). If this PR is okay, I'll go look elsewhere and/or try to figure a general cure that fits with the direction of the project, and ideally if able, I'll try to switch at least some legacy code over to mvc once I'm confident enough to do so.

Awaiting thoughts.

src/www/firewall_rules_edit.php
```diff
@@ -40,8 +40,9 @@
 if ($ostypes == null) {
     $ostypes = array();
 }
+$ipprotocols = array('inet', 'inet6', 'inet46');
```
fichtner: basically we have two options: validate *all* pages that allow such input consistently

or leave it unvalidated having XSS issues worst case (if pages are not protected which they normally are) but then again the rule generator underneath should be clever enough to discard garbage so that your rules are not tainted. I'm fairly certain you can't inject rules anymore, which is the only way this would be going. And it would also argue you'd inject a rule in order to be able to change rules what you already can submit in the first place.

validation of drop downs in general is a matter of debate and always happens deliberately so the basic question is what is the thread model?

src/www/firewall_rules_edit.php
```diff
@@ -74,9 +75,11 @@ function is_posnumericint($arg) {
     return (is_numericint($arg) && $arg[0] != '0' && $arg > 0);
 }
 
-
 $a_filter = &config_read_array('filter', 'rule');
 
+// Whether POST or GET, the URL should still have a valid $_GET['id'] field controlling the interface edited
+// It's only validated on POST, but that's safe because only the initial page load is a GET and config only updates upon POST
+$interface_from_url = (empty($_GET['if']) || !is_string($_GET['if'])) ? "" : $_GET['if'];
```
fichtner: this isn't validation as it's assigning any value... benefit vs. cost... and is likely validated below correctly?

stilez: It is validating a value - in an important sense. The value shown to the user, and the interface they understand they are editing, is based on this.  As a general principle it shouldn't be possible to have contradictory values this way. Or put another way, if `&if=` is how the interface being edited is specified, and the POST comes back with $_GET containing `&if=lan` but `$_POST['interface']=wan`, then something's wrong and we shouldn't process the form. 

Also noted that the menu title, **which the user is shown** to confirm the interface being edited, is generated from the URI via $_GET (head.inc/19: `$menu->getItems($_SERVER['REQUEST_URI']);`). But **the interface actually configured** is derived from $_POST. So if the two differ "somehow", the interface presented as being edited and the interface actually edited will silently differ. Very not good.

Cost of checking they;'re consistent? Seems tiny by any standard. Assign it to a variable at the start, use that variable as the authoritative source of "interface being edited" throughout form generation, and check $_POST['interface'] is still consistent with it on form submission. That's all. Trivial to check.

For those reasons I've added this check. It ensures that at least, the interface to be edited, and the interface the user is told will be modified, are the same. Which at best shouldn't hurt and is no bad thing to add if cautious. 

Hopefully you can agree - or at least decide it's not objectionable to check anyhow?

fichtner: It looks like this tries to fix something in the middle. We have five cases:

1. if= not set, means floating
2. if= set, doesn't load page properly
3. if=validinterface set, means floating
4. if=invalidinterface set, doesn't load page properly
5. if=FloatingRules set, means floating

so we "fix" 2 but not 4 ? why? what is the strategy to deal with invalid input other than assuming "floating" ?

src/www/firewall_rules_edit.php
stilez: I'm not sure. I'm not a security specialist in that sense. I do think that if one follows that line, the argument becomes "we prevent a malicious actor entering unwanted data as best we can, but beyond that, junk is OK in any field because it gets blocked at the point of use".  

Which is, I suppose valid but seems pretty unsatisfactory. Do we validate that ports are inputted as 1-65535 only, as well as check at the point we generate pf.conf? What about IP addresses, would we let people enter ";rm *" or whatever as an IP address and accept that into $config, on the basis that it's fine because it'll be caught when IP-based config is built and fields are validated? Or XML characters? Why have $input_errors[] anyway if we catch on use? 

I just... I'm not sure the threat model or how a secuirity specialist would see it. I just feel profoundly disturbed that any form would accept arbitrary text and insert it into the config, whether or not we believe it'll be caught later and rejected upon actual use.

Surely it makes sense to fix if a fix is offered, even if its just one page at present and even if the legacy code will at some time be superseded.

fichtner: In these cases a top down would be best. Set a strategy for invalid input. If case 2 and 4 are bad, maybe even case 1 why not throw an input error right away?

src/www/firewall_rules_edit.php
stilez: - Case 1 shouldn't ever be possible. If floating, the uri will be `&if=FloatingRules` and the same URI will be loaded after POST. It's never missing if things are working correctly.
- Case 2 also incorrect. It loads the page fully (albeit without a breadcrumb but total omission => not obvious). And the page fully works. It will allow a page to be edited and saved, and save it against whatever interface is in $_POST even if &if= is empty.
- Case 4 also incorrect., behaves same as case 2.

This PR fixes all of them, by ensuring &if is valid, the one or multiple interfaces in $_POST['interface'] are valid, and that &if is either FloatingRules or the single (valid) interface in $_POST['interface'].

fichtner: Thanks for explaining what I said. Now that we are on the same page I don't agree with the proposed fix for 1, 2 and 4.

fichtner: To be more precise: there's no use validating only POST if we go so far as to present a faulty form to the user on a faulty GET request, because the accepted GET request is the root of the problem.

stilez:
Thanks Franco. And if I check we are on the same page, it's because I'm new to this, you definitely aren't, and I want to get myself more onto your page on some things.

Both comments above seem to come down to the same thing - if anything is done, it should be a top-down input strategy, and pages should raisde input errors immediately on the GET, not just when they're submitted for processing in the POST. 

I'm happy to work with that (subject to checking we are on the same page on "top down strategy..." and the scope and approach that might be suitable). I do like the idea of validating the data on these forms on principle. Let me know what you feel  best, and I'll have a go at implementing it, on the firewall-rules pages at least for now. 

My own suggestion for this legacy page would be simply to raise an $input _error[] on any bad combinations of data in GET and POST, whether it's being loaded as a result of an initial URI GET or a form submission. The mechanism is there to suppoirt both, including error presentation in the header. If the error is fixable by the user editing the form fields, then as usual the user can edit the fields and submit it. If not, for example bad GET, then the user is told that there has been an error in loading the form, and to cancel and reload the form from the previous page. It's vague, but enough to guide a user. If it happens (which it never should) I'm sure we would hear it in github or forum and investigate then. If these pages are rewritten using phalcon then Im guessing this will probably fix legacy issues as pages are migrated.

fichtner:
Since fixing all pages isn't scalable and static php pages need to be moved to MVC where such validation exists I'm closing this after + 1 year without movement.


Cheers,
Franco

PR #3807, opened 2019-11-10
Firewall Rules: Add a "Detail" toggle button in the button bar, to allow more complete rule display

stilez:
This PR makes a small but crucial change to the display of firewall rules.

At present, for layout purposes, some key information isn't displayed, which really is crucial to the rules. The example I have in mind is that for Floating Rules, you can't see which interfaces the rules apply on, without opening each rule individually in the editor. That's pretty crucial rule information.  But trying to show everything in the table would make the display unusable and ugly. Against this, for many users this info will be needed and many will have wide displays capable of showing full details in a row.  Not everything should be immediately displayed, but the key info needs to be at least easily _displayable_.  

Dilemma.

To overcome this issue, this PR does 2 things.

- Adds a neat unobtrusive button for "details", so extra key details or columns can be displayed.
- Uses the button so that when toggled, an extra column is displayed (in Floating Rules only) to show the rule interfaces, with useful interface hover info - the most immediate missing information at present.

This (1) solves the immediate problem of interface viewing for Floating Rules, and (2) provides a means for other important data to be expanded upon, or displayed, going forward, without disrupting the clean present display, hence generalising the benefits.

As an aside, I've made a minor change to rename the existing button from "Inspect" to "Statistics".  The main reason being that "Inspect" is actually pretty ambiguous in the context of partial info about rules (does it mean inspect rule details, or inspect the stats?).  "Statistics" is much more immediately obvious and is also very accurate, since activating it displays per-rule statistics.

### DIFF
--- a/src/www/firewall_rules.php
+++ b/src/www/firewall_rules.php
@@ -1027,6 +1027,9 @@ $( document ).ready(function() {
                   <tr>
                     <td colspan="2"></td>
                     <td colspan="2" class="view-info"></td>
+<?php if ($selected_if == "FloatingRules"): ?>
+                    <td class="view-fullinfo"></td>
+<?php endif; ?>
                     <td colspan="5" class="view-info hidden-xs hidden-sm"></td>
                     <td colspan="2" class="view-stats hidden-xs hidden-sm"></td>
                     <td colspan="2" class="view-stats"></td>
--- a/src/www/firewall_rules.php
+++ b/src/www/firewall_rules.php
@@ -35,6 +35,30 @@ require_once("system.inc");
 /***********************************************************************************************************
  * format functions for this page
  ***********************************************************************************************************/
+function firewall_rule_item_interfaces($filterent)
+{
+    // construct line interfaces for detailed view
+    if (empty($filterent['interface'])) {
+        return "<span class='text-info' data-toggle='tooltip' title='All'>*</span>";
+    }
+    $interfaces = legacy_config_get_interfaces();
+    $iflist = explode(',', $filterent['interface']);
+    $out = array();
+    foreach ($iflist as $if) {
+        if (!empty($interfaces[$if])) {
+            $ifdata = $interfaces[$if];
+            $name = !empty($ifdata['descr']) ? $ifdata['descr'] : $if;
+            $phy = $ifdata['if'];
+            $hover = $phy . (strcasecmp($if, $name) ? ' (' . $if . ')' : "");
+        } else {
+            $name = $if;
+            $hover = "No details";
+        }
+        $out[] = sprintf("<span class='text-info' data-toggle='tooltip' title='%s'>%s</span>", htmlspecialchars($hover), strtoupper(htmlspecialchars($name)));
+    }
+    return implode(', ', $out);
+}
+
 function firewall_rule_item_proto($filterent)
 {
     // construct line ipprotocol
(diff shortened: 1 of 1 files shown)


AdSchellevis:
If the idea is to show interfaces in floating rules, we might consider adding the field by default on the floating tab, when using a large (-lg) screen size.

The current solution only functions partially (or is at least confusing when in inspect mode) and could probably be simplified.

(detail enabled on floating tab, normal view)
image

(switching to inspect)

image

detail on lan tab:
image

... and switching to inspect:
image

The `Inspect` vs `Statistics` choice is probably a matter of taste, we might consider changing it, but in the original context (which I would suggest we keep), it's quite obvious what it's intend is.

stilez:
Yes, when it switches to inspect/stats, then the detail button could probably lose its 'active' status.  I'll add that to the PR shortly.

The advantage of this is, if a user has a wide display, or at times there is other useful info for rules, we have an elegant way to improve rule detail visibility, without negative effect on the UI for the majority of use. Since rules can have many non-trivial parameters, but the display is limited, this might mean that a selection of other useful parameters (or hoverable  icons representing some) can be made visible in the table, for quick viewing, without compromising the table. 

So it's not just the floating rules interface. Its more that that's an obvious use. If the button is acceptable, then I'd like to expose a few more things, I haven't done it yet because it was late at night. I'll do it shortly today.

AdSchellevis:
It's probably better to use css to automatically show/hide options on different screen size, one less button in need to explain and likely easier to integrate as it is. (if the `-lg` bootstrap tag isn't suitable, there might be options to add an extra viewpoint for even larger screens)

stilez:
What I discovered on tech UI is some people like a lot of detail, even if it's small, others want big fonts and a way to view to detail. With today's 4k monitors being commonplace on everything from 12 inch (30 cm) to 36 inch (90 cm) diagonals, I don't think window pixel size is always a good guide to what should show.  Although we do specify a particular font size, so maybe its okay? But it also isn't hard to add logic to auto-expose extra columns if the width is > some value.  Or to give the user a (remembered) choice of columns, much like the dashboard has a choice of widgets? Multiple approaches, which do you think would please most, and upset the rest least?

AdSchellevis:
I would just  start with adding the interface field for floating and make sure it's visible on `-lg` size displays and see how that looks / reacts. small steps, easy for others to try as well.

stilez:
There are plenty of examples of "accordion" code both bootstrap and css, for that. ([Example1](https://stackoverflow.com/questions/15519047/collapse-div-element-based-on-screen-size) [Example2](https://teamtreehouse.com/community/showhide-elements-depending-on-screen-size)). I'm not sure if the interfaces can be added easily for floating rules, on a small window, without going to 2 lines per rule. But I'll have a go, and come back with whatever I get to. Might take some experimenting as there's a lot to try out.

stilez:
Can you take a look at this? It's screenshots for 3 sizes (smallest and largest `-lg`, and smallest `-xl`), comparing the current layout (on my main router: top row) and the layout I've got so far. 

I haven't incorporated the floating interface yet.  It's clear I'll need to create some screen space for that, so this is how I'm doing it. It looks pretty good to me so far, but before I put more time into it, it's worth checking that you like the general direction.  Usefully, there's almost zero change to the original code, as I'm using output buffers + bootstrap classes. Sample screenshots and code snip below.

The compact version also replaces port=`*` with `Port: *`  for visual ease. (It worked better than bare `* *`).

What's immediately visible to me, is the saved horizontal space, without clutter.  Compare the 2 rows.  There are obviously some overflow and underflow issues, which if this is seen positively, I'll sort out, before a second set of screenshots showing updated layout with+without floating rule interfaces.

Sample of code used to generate compact layout in bootstrap:
```
        <td class="view-info hidden-xs hidden-sm hidden-md">
<?php
            ob_start(); ?>
                 {{ EXISTING CODE TO GENERATE IPADDR COLUMN CONTENT }}
<?php 
            $ipaddr_html = ob_get_contents(); 
            ob_end_flush(); ?>
        </td>
        <td class="view-info hidden-xs hidden-sm hidden-md">
<?php
            ob_start(); ?>
                 {{ EXISTING CODE TO GENERATE PORT COLUMN CONTENT }}
<?php 
            $port_html = ob_get_contents(); 
            ob_end_flush(); ?>
        </td>

        <td class="view-info hidden-lg hidden-xl">                <---- new column
            <?= $ipaddr_html; ?> <br/>
            <?= trim($port_html) == '*' ? gettext("Port") . ": *" : $port_html ; ?>
        </td>
```
opnsense1

AdSchellevis:
timeout, currently out of scope.

PR #3808, opened 2019-11-10
Interfaces.php:  Make interface name control visible when interface disabled

stilez:
Configuring an interface, or viewing its basic config,  just shouldn't be conditional on enabling it.

As the code stands, you can't even give an interface a name, unless it's also enabled. Much less set params that should apply to it. Which often you may want to set, even if the interface isn't yet ready to be enabled.  (Also one often needs to do things between initial creation of the interface and it being ready to enable, and these basics need to be there).

Basics like this should always be accessible for editing, there's no reason they should be behind a visibility toggle and blocked unless enabled.  It's not like the form is long or complex either.  This PR removes the unhelpful show/hide behaviour.

### DIFF
--- a/src/www/interfaces.php
+++ b/src/www/interfaces.php
@@ -1439,6 +1439,19 @@ include("head.inc");
 <body>
 <script>
   $( document ).ready(function() {
+      function toggle_showcfg() {
+          if ($("#enable").prop('checked')) {
+              $(".showcfg").show();
+              $(".hidecfg").hide();
+          } else {
+              $(".showcfg").hide();
+              $(".hidecfg").show();
+          }
+      }
+      // when disabled, hide settings.
+      $("#enable").click(toggle_showcfg);
+      toggle_showcfg();
+
       $("#type").change(function(){
           $('#staticv4, #dhcp, #pppoe, #pptp, #ppp').hide();
           $("#rfc3118").hide();
@@ -1793,25 +1806,33 @@ include("head.inc");
                     </thead>
                     <tbody>
                       <tr>
-                        <td><i class="fa fa-info-circle text-muted"></i> <?= gettext('Enable') ?></td>
-                        <td>
+                        <td style="width:22%"><i class="fa fa-info-circle text-muted"></i> <?= gettext('Enable') ?></td>
+                        <td style="width:78%">
                           <input id="enable" name="enable" type="checkbox" value="yes" <?=!empty($pconfig['enable']) ? 'checked="checked"' : '' ?>/>
                           <strong><?= gettext('Enable Interface') ?></strong>
                         </td>
                       </tr>
                       <tr>
-                        <td><i class="fa fa-info-circle text-muted"></i> <?= gettext('Lock') ?></td>
-                        <td>
+                        <td style="width:22%"><i class="fa fa-info-circle text-muted"></i> <?= gettext('Lock') ?></td>
+                        <td style="width:78%">
                           <input id="lock" name="lock" type="checkbox" value="yes" <?=!empty($pconfig['lock']) ? 'checked="checked"' : '' ?>/>
                           <strong><?= gettext('Prevent interface removal') ?></strong>
+                        </td>
+                      </tr>
+                      <tr>
+                        <td style="width:22%"><a id="help_for_descr" href="#" class="showhelp"><i class="fa fa-info-circle"></i></a> <?=gettext("Description"); ?></td>
+                        <td style="width:78%">
+                          <input name="descr" type="text" id="descr" value="<?=$pconfig['descr'];?>" />
+                          <div class="hidden" data-for="help_for_descr">
+                            <?= gettext("Enter a description (name) for the interface here."); ?>
+                          </div>
                         </td>
                       </tr>
                     </tbody>
                   </table>
                 </div>
               </div>
-              <div id="allcfg">
-                <div class="tab-content content-box col-xs-12 __mb">
+                <div class="tab-content content-box col-xs-12 __mb"><!-- UNINDENT FROM HERE-->
                   <div class="table-responsive">
                     <!-- Section : All -->
                     <table class="table table-striped opnsense_standard_table_form">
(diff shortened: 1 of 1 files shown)


AdSchellevis:
Although it might be datable if you should be able to name the unused interface, we're not planning to remove the current visibility toggle on disabled interfaces.

(The current option helps people to spot un-active interfaces extra quick)

In the last 4 years we never had a question about this, when changing this behaviour, chances are quite large others won't agree with your view on this.

stilez:
Hiding config fields when the interface is disabled, is useful because it helps people spot disabled interfaces extra quickly?? Like, almost everything else, don't we trust the user to notice the checkbox? We don't for example hide the resolver UI because it would help to identify an inactive resolver. 

 I would expect the absence of traffic would be a hint :)  And the name box is hidden  with the other config options.  This behaviour I can agree, is not that important but it does surprise me. I can think that it would very useful to be able to set up config on an interface, without having to enable it to do so.

AdSchellevis:
It might be an option to just move the description one block, in which case you can name the interface you're not planning to use. Simple change, probably won't hurt anyone else.

(believe me, we see all sorts of things while doing support, every change we make has potential impact on our support desk... which we have to be aware of all the time)

stilez:
> It might be an option to just move the description one block, in which case you can name the interface you're not planning to use. Simple change, probably won't hurt anyone else.
 
That does help. I'll look to see if its all settings or really just the name plus maybe just one or 2 others.

> (believe me, we see all sorts of things while doing support, every change we make has potential impact on our support desk... which we have to be aware of all the time)

The extent of this, which is a true and valid point, sometimes gives me nightmares ;-)

fichtner:
can we please keep the diff to a minimum?

src/www/interfaces.php
```diff
@@ -1805,25 +1806,33 @@ function toggle_allcfg() {
...
+                          <div class="hidden" data-for="help_for_descr">
+                            <?= gettext("Enter a description (name) for the interface here."); ?>
+                          </div>
+                        </td>
+                      </tr>
                     </tbody>
                   </table>
                 </div>
               </div>
-              <div id="allcfg" style="display:none">
```
fichtner: embedding dislay:none is an intentional mobile optimisation

fichtner:
this is definitely the wrong direction this PR is going with thousands of lines changed, I'm afraid to point out

stilez:
I've updated and largely reinstated based on @AdSchellevis ' helpful comment.

The updated PR retains the "display:none" and just (1) moves the name outside the hidden section, (2) adds a text note that config requires enabling, as it is "surprising" that there isn't config shown when the interface isn't enabled, so it might be nice to tell the user.

As a consequence, about 1600 lines got unindented by x1, but that's whitespace only and in a single separate commit with no code change, to keep review easy.

If it helps, I can close and resubmit?

src/www/interfaces.php
stilez: now retained

stilez:
Okay, try this?

Everything is pretty much as it was except:

- movement of "name" field to be always visible
- addition of the text line when config section is hidden
- moved `#showcfg` to `tbody` tag, so the text message when shown, has the same appearance and layout as the config section. This resulted in the empty `div` tag which I retained as it's harmless and avoids ~ 1600 unindents.

AdSchellevis:
how about just moving this block 
https://github.com/opnsense/core/blob/b97789b2d9ea717a2e70b1c2f7a42cfbbdbebf0c/src/www/interfaces.php#L1836-L1844

below

https://github.com/opnsense/core/blob/b97789b2d9ea717a2e70b1c2f7a42cfbbdbebf0c/src/www/interfaces.php#L1814-L1820 

?

(both sections have the same description)

stilez:
Done. Try this. As above (see previous comment for changes)

- `Name` field always visible.  
- Text hint shown when config hidden (2nd `tbody` section to ensure no change to layout)
- `Show/hide` attributes moved from `div` to `tbody`. Redundant `div` left but harmless + means no change to indentation.

src/www/interfaces.php
```diff
@@ -3553,8 +3561,7 @@ function toggle_allcfg() {
                 </div>
 <?php
                 endif; ?>
-              <!-- End "allcfg" div -->
-              </div>
+              </div><!-- unused div -->
```
fichtner:
```suggestion
              </div>
```

src/www/interfaces.php
```diff
@@ -1832,16 +1843,13 @@ function toggle_allcfg() {
...
                         <tr>
-                          <td style="width:22%"><a id="help_for_descr" href="#" class="showhelp"><i class="fa fa-info-circle"></i></a> <?=gettext("Description"); ?></td>
-                          <td style="width:78%">
-                            <input name="descr" type="text" id="descr" value="<?=$pconfig['descr'];?>" />
-                            <div class="hidden" data-for="help_for_descr">
-                              <?= gettext("Enter a description (name) for the interface here."); ?>
-                            </div>
-                          </td>
+                          <td style="width:22%"></td>
+                          <td style="width:78%"><?=gettext("The interface must be enabled to configure detailed settings"); ?></td>
```
fichtner:
```suggestion
                          <td style="width:78%"><?=gettext("The interface must be enabled before configuring it further."); ?></td>
```

src/www/interfaces.php
```diff
@@ -1818,11 +1820,20 @@ function toggle_allcfg() {
...
+                            <?= gettext("Enter a description (name) for the interface here."); ?>
+                          </div>
+                        </td>
+                      </tr>
                     </tbody>
                   </table>
                 </div>
               </div>
-              <div id="allcfg" style="display:none">
+              <div><!-- unused div -->
```
fichtner:
```suggestion
              <div>
```

src/www/interfaces.php
stilez: Hesitant (pendantic?) because "before configuring it _further_ " kind of feels like just first time setup, but here it needs to be enabled _every_ time, to even see (not just access) any other config settings. Pedantic I know.  

The suggested edit doesn't quite feel right to me but it's really minor wording and not worth much time. So please pick a wording that you like, and apply it :) I'll be fine with it.

fichtner: "detailed" settings is misleading. we now have two "general configuration" tabs and we unhide the second.

maybe the section headers could be "basic configuration" and "generic configuration".. ."general" seems too, well, general and we already "configure" in the case of the "basic" things like name, lock and operational state.

it would be nice to align the guiding text to the reality of the page descriptions. maybe the suggestion isn't there 100% yet, but that's why we discuss... so what do you think?

stilez: - Top part (enable/lock/name):  `Basic [Interface] Configuration` or `Basic [Interface] Settings`
- Everything else:  `Network Settings` or `Networking`

Would that work? The distinction seems to be that the top (always visible) part is basic configuration/basic settings, such as on/off and locked/unlocked;  the rest are its related network settings such as DHCP/MTU.

PR #3810, opened 2019-11-10
use freebsd built in growfs script

DarkSunOne:
this didn't worked anymore.
tested successfully with the freebsd built in growfs script.
Maybe you want to merge this?


src/etc/rc
```diff
@@ -73,9 +73,7 @@ while read FS_PART FS_MNT FS_TYPE FS_MORE; do
 		FS_DEV=$(glabel status -as | grep ${FS_DEV} | awk 'match($3, /^[a-z]+[0-9]+/) { print substr( $3, RSTART, RLENGTH )}')
 		if [ "${FS_MNT}" = "/" -a -f ${GROWFS_MARKER} ]; then
 			# hammertime!
-			gpart recover ${FS_DEV}
-			gpart resize -i 1 ${FS_DEV}
-			growfs -y ${FS_MNT}
+			/etc/rc.d/growfs onestart
```
fichtner: woks for me and also adds ZFS support... yet.. if we go through all the trouble now to do a one-shot script without arguments wouldn't it be better to move it below the fstab loop? just need to check for GROWFS_MARKER this way. :)

DarkSunOne: Do we have any benefits, moving it outside the loop?
if we let it within the loop, the resize still happens offline, or am i wrong?

I would still prefer to do partition changes with not mounted filesystems ;)

fichtner: the point is to remove obsolete and unused code

the loop here only reads fstab and configures before mounting the file systems

the next loop will mount the file system

simple solution: use the space in between to execute the growfs code :)

DarkSunOne: Well, i changed it now, will update the pr when tests are finished.

Is the whole fstab loop (tunefs, trim and so on) still needed?

fichtner: yes, we need that

DarkSunOne: files are updated and first test worked without problems :)

src/etc/rc
```diff
@@ -71,12 +71,6 @@ while read FS_PART FS_MNT FS_TYPE FS_MORE; do
 	FS_DEV=$(echo ${FS_PART} | awk 'match($0, /^\/dev\/(gpt|ufs)\/.+$/) { print substr( $0, RSTART + 5, RLENGTH - 5 )}')
 	if [ -n "${FS_DEV}" ]; then
```
fichtner: zap

src/etc/rc
```diff
@@ -71,12 +71,6 @@ while read FS_PART FS_MNT FS_TYPE FS_MORE; do
...
 		FS_DEV=$(glabel status -as | grep ${FS_DEV} | awk 'match($3, /^[a-z]+[0-9]+/) { print substr( $3, RSTART, RLENGTH )}')
```
fichtner: zap

src/etc/rc
```diff
@@ -71,12 +71,6 @@ while read FS_PART FS_MNT FS_TYPE FS_MORE; do
...
-		if [ "${FS_MNT}" = "/" -a -f ${GROWFS_MARKER} ]; then
-			# hammertime!
-			gpart recover ${FS_DEV}
-			gpart resize -i 1 ${FS_DEV}
-			growfs -y ${FS_MNT}
-		fi
 	fi
```
fichtner: zap

src/etc/rc
```diff
@@ -95,6 +89,11 @@ while read FS_PART FS_MNT FS_TYPE FS_MORE; do
 		fi
 	fi
 done < /etc/fstab
+		
+if [ -f ${GROWFS_MARKER} ]; then
+	# hammertime!
```
fichtner: thanks for retaining this comment ❤️

src/etc/rc
fichtner: RC will be a good opportunity to test this later. Should also work fine on 11.2 from was I can tell. Nice stuff this.

src/etc/rc
DarkSunOne: had to :smile: :heart:

src/etc/rc
DarkSunOne: what do you mean with zap? ^^

fichtner: all those lines are no longer needed = no code to execute

fichtner: edited PR, will merge now :)

fichtner:
Merged, thanks! ❤️

PR #3812, opened 2019-11-11
Volt: Add a new type "section_title" for related fields on forms without closing+restarting the form table

stilez:
<s>(1) The type field is omitted in the list of valid fields.</s> **PRed and merged separately**

(2) There isn't currently an easy way to add "section headers" to groups of fields. While this doesn't have syntactic importance (it's just a "info"/"label" without "help" and with a tiny extra emphasis), it is useful as a distinct type because in-page sections to demarcate groups of fields are widely used in the UI (eg: `System: Settings: Administration` to add sections headers for Web GUI, Secure Shell, Console and Authentication)*.

**Rationale:**

The main issue is that section headers are like info fields **without** a "help" icon to the left (see `System: Settings: Administration` for thisd). But all other phalcon-defined types do have an active or dimmed help icon, and currentl;y there's no way other than on-load JS to remove it for section titles. Also they need slight emphasis - a tiny amount of spacing and size, compared to normal field labels (see same page).  Hence a new type is needed or worthwhile..

Adding `section_title` to the formal list of types is trivial from a code perspective, and ensures 

- Phalcon generated pages have consistent appearances for section_titles in the UI
- They're easy to add if required without perplexing extension devs as to how to do it within the xml :)

**Testing:**

- Pick any mvc `forms` `.xml` file
- Add an extra field at the top or halfway down (or both):
```
   <field>
      <id>optional.pick.any.id</id>
      <type>section_title</type>
      <label>Security Options:</label>
   </field>
```
- Delete the Phalcon cache and reload the page.

**Note:  The CSS is a "pretty good starting point", but if it needs changing, please change it.**

### DIFF
--- a/src/opnsense/mvc/app/views/layout_partials/form_input_tr.volt
+++ b/src/opnsense/mvc/app/views/layout_partials/form_input_tr.volt
@@ -28,6 +28,16 @@
  # Handle input table row, usage the following parameters (as associative array):
  #
  # id          :   unique id of the attribute
+ # type        :   type of input or field. Valid types are:
+ #                   text               single line of text
+ #                   password           password field for sensitive input. The contents will not be displayed.
+ #                   textbox            multiline text box
+ #                   checkbox           checkbox
+ #                   dropdown           single item selection from dropdown
+ #                   select_multiple    multiple item select from dropdown
+ #                   hidden             hidden fields not for user interaction
+ #                   info               static text (help icon, no input or editing)
+ #                   section_title      label demarcating groups of fields (no help icon, no syntactic significance to mvc)
  # label       :   attribute label (visible text)
  # size        :   size (width in characters) attribute if applicable
  # height      :   height (length in characters) attribute if applicable
@@ -42,10 +52,14 @@
 
 <tr id="row_{{ id }}" {% if advanced|default(false)=='true' %} data-advanced="true"{% endif %}>
     <td>
-        <div class="control-label" id="control_label_{{ id }}">
+            {% if type == "section_title" %}
+                <div class="control-label volt_section_title" id="control_label_{{ id }}">
+            {% else %}
+                <div class="control-label" id="control_label_{{ id }}">
+            {% endif %}
             {% if help|default(false) %}
                 <a id="help_for_{{ id }}" href="#" class="showhelp"><i class="fa fa-info-circle"></i></a>
-            {% elseif help|default(false) == false %}
+            {% elseif help|default(false) == false and type != "section_title" %}
                 <i class="fa fa-info-circle text-muted"></i>
             {% endif %}
             <b>{{label}}</b>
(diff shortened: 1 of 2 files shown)


mimugmail:
I always thought we use type header for this:

https://github.com/opnsense/plugins/blob/master/www/nginx/src/opnsense/mvc/app/controllers/OPNsense/Nginx/forms/location.xml#L226-L230

stilez:
> I always thought we use type header for this

"Header" isn't defined in the code either. That's a decent term as well. But call it what one likes - hesder, section, title, - there's no type actually defined for it in mvc and no way to get the nuances of.formating or uniformity right now. As the above comment says, you can't negate the "info-circle[-muted]" icon on a header/section title to get it left aligned as on other pages, because the code at lines 46-50 always adds an icon. Any emphasis used to guide the eye as on other pages (size, padding etc) isn't uniform either, because it has to be added in the XML via CDATA. 

So regardless what we call it, I think it's a good idea to add some entity for the purpose :) but if already using header, and that's what people like (group?section?) then stay with that?

AdSchellevis:
can you split the PR? The documentation I can pull at the moment, the section is likely already implemented like @mimugmail says.

stilez:
[Found it](https://github.com/opnsense/core/blob/15cbd50b85eec36d161c544a826dadb1398bc871/src/opnsense/mvc/app/views/layout_partials/base_form.volt#L71-L95). Took some digging. 

One lesson from thios is that there needs to be a maintained list of already-defined field types in the wiki.  There are many many files in which field types could be defined, as well as custom types, and variants on similar types.  A new dev doesn't have a hope of finding the type they need, or knowing if it exists, without some kind of reference list. Can you take a look at that code link, and see if it answers the issue in this PR? If so let's close this PR but open an issue for tracking field types, if you think it's useful.

stilez:
> I always thought we use type header for this:

**Tl;dr - yes we use "header". No we shouldn't be using it because it doesn't work and screws up forms if used for that.**

It took a lot of digging to find it [(link)](https://github.com/opnsense/core/blob/15cbd50b85eec36d161c544a826dadb1398bc871/src/opnsense/mvc/app/views/layout_partials/base_form.volt#L71-L95), even knowing it was called "header" and appeared in nginx, because that plugin has numerous stacked tabs in its main tabs :) Thank you for your exceptional skills/memory! 
  
The problem is that type=`header` ends the present table and restarts a new separate table, and the columns won't align because of differing word wrap. 

By contrast the `section/title` type in this PR isn't a table header. It doesn't close the current table and open a new one.  It literally just adds a row in the existing table cointaining a section header text with standard formatting.  So the columns are guaranteed to align, alternate backgrounds guaranteed to remain consistent, and so on.

**Below: Screenshot of the `nginx` HTTP Location dialog noted in @AdSchellevis' previous comment, showing how it actually renders wrongly **because** it uses `header` for the separator.**

opnsense4b

Tangentially, what do you think about keeping a maintained list of already-defined field types in the wiki? There are many many files in which field types could be defined, as well as custom types, and variants on similar types.  A new dev doesn't have a hope of finding the type they need, or knowing if it exists, without some kind of reference list.  Should I add this to the dev docs issue as it fits in there neatly?

AdSchellevis:
It would indeed be good to extend the documentation on this subject, we have added a lot lately (some changes not yet published), but not about the volt partials.

It should probably land somewhere in https://docs.opnsense.org/development/frontend.html 

An alternative for the header might be an option, I haven't tried this PR yet (if we replace it we should remove the old construction as well at some point in time).

AdSchellevis:
timeout

PR #3814, opened 2019-11-12
Volt base forms: Effectively remove unused 3rd colgroup by reducing default mgszone_width=0

stilez:
See discussion about colgroups/msgzone_width [on the forums](https://forum.opnsense.org/index.php?topic=14958.0) (2019-11-12).

**Tl;dr** - dialogs generated by volt define a 3rd colgroup (5/12 = 42% of the total dialog width), and a much narrowed 2nd colgroup. But this 3rd colgroup isn't ever used, other than in 1 form, and even then it's only referred to by setting its width to near zero to try and remove it.  So it has no useful uses right now. Its only actual - as opposed to originally intended - effect is to stop helptext and fields from occupying the right half of dialogs, which is never used anyway and always white spaced. This makes white space poorly used and forms often longer due to excessive line wrapping.

@AdSchellevis ' comment on this issue in the linked forum thread was:

> Originally it [msgzone_width] came from https://github.com/opnsense/core/commit/4c736c65060c926ecc9eb7539b93454559e9d2d4 intended to display "wide" forms, see "Administration -> Intrusion detection -> Rule" for its usage.
> 
> It's not at the top of my priority list, but I wouldn't mind removing the parameter if both form types still look decent. As far as I see, we only use it in IDS at the moment, so it should be easy to refactor and test.

However, actusally removing it would mean changing almost every volt-generated table with a third empty `<td></td>` or `colspan`. in the entire codebase - impractical and pointless, waste of dev time and testing laborious.  It's much easier to simply change the default size of the 3rd colgroup to 0 cols and add trivial CSS to consistently define a zero-width colgroup without L/R padding, than to modify the entire codebase removing it.  It also makes it easily reversible/settable if ever needed.

I've copiously commented this in the edited .volt base files and the css, because it may not be obvious to future devs what's behind the code chosen.

It tests very well here, but as it affects a lot of things, please check to make sure I haven't missed anything!

### DIFF
--- a/src/opnsense/mvc/app/views/OPNsense/IDS/index.volt
+++ b/src/opnsense/mvc/app/views/OPNsense/IDS/index.volt
@@ -928,6 +928,6 @@ POSSIBILITY OF SUCH DAMAGE.
 
 {{ partial("layout_partials/base_dialog_processing") }}
 
-{{ partial("layout_partials/base_dialog",['fields':formDialogRule,'id':'DialogRule','label':lang._('Rule details'),'hasSaveBtn':'true','msgzone_width':1])}}
-{{ partial("layout_partials/base_dialog",['fields':formDialogRuleset,'id':'DialogRuleset','label':lang._('Ruleset details'),'hasSaveBtn':'true','msgzone_width':1])}}
+{{ partial("layout_partials/base_dialog",['fields':formDialogRule,'id':'DialogRule','label':lang._('Rule details'),'hasSaveBtn':'true'])}}
+{{ partial("layout_partials/base_dialog",['fields':formDialogRuleset,'id':'DialogRuleset','label':lang._('Ruleset details'),'hasSaveBtn':'true'])}}
 {{ partial("layout_partials/base_dialog",['fields':formDialogUserDefined,'id':'DialogUserDefined','label':lang._('Rule details'),'hasSaveBtn':'true'])}}
--- a/src/opnsense/mvc/app/views/layout_partials/base_dialog.volt
+++ b/src/opnsense/mvc/app/views/layout_partials/base_dialog.volt
@@ -67,10 +67,15 @@
                 <form id="frm_{{base_dialog_id}}">
                   <div class="table-responsive">
                     <table class="table table-striped table-condensed">
+                        {# See discussion about colgroups/msgzone_width at:                                   #}
+                         # https://forum.opnsense.org/index.php?topic=14958.0 (2019-11-12)                    #}
+                         # Tl;dr it's easier to default 3rd colgroup -> 0 cols, than to modify entire codebase removing 3rd col.
+                         # Also reversible/settable if ever needed. 
+                         # If modified in future, the same code below, and also base_form.volt, will also need updating. #}
                         <colgroup>
                             <col class="col-md-3"/>
-                            <col class="col-md-{{ 12-3-msgzone_width|default(5) }}"/>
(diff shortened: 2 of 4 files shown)


AdSchellevis:
image
^^ 3rd column group (validation messages)

stilez:
So it's used that way.  Okay.  I'm thinking several answers, none difficult. 

- validation messages below  - don't like, as less obvious, especially if help text displayed.
-validation messages above - works, no obvious issue, especially as control will be red outline too.
- add msgzone_width where this is an issue - likely to be many/most dialogs and white space still wasted. Or set default to a smaller number like 2 or 3, not the original 5, and let it wrap. Giving priority to the stuff that's always needing to be seen.

But my preferred answer by far is:

- validation messages in column 2, to the right of the control. Possibly in a 2 x 1 invisible table (no borders or background) or `<ul><li>` tags. If the dialog is wide enough, they'll show to the right as at present. If the dialog is narrow, they'll wrap neatly (CSS/Bootstrap) to below. And the controls and help text can use space if they need it, optimising usage.

stilez:
Would this be an acceptable way to display the same validation message

opnsense-error-on-top

Rationale: 

- main focus should be on controls to be entered, restricting them to ~ 44% of available width after sidebar (4 of (12-3) cols) wastes space that is premium-use, and adds word wrapping that makes forms longer, and hard to read on smaller devices.
- equally when there is an error, it should neatly stand out. The right hand side is a good place to highlight errors.
- Placing the error above the control and an icon to the right (change my choice if you like!) meets both these UI goals.
- If we place the icon just right of the control, it will be (1) ugly because multiple error triangles won't align since width of controls differs, (2) hard to see because other controls and help text might flow into the same horizontal spacespace.

So despite starting with the intent to make a flewxible layout, I've actually kept the 3 column layout, but reduced the right hand column to 1 bootstrap column (3-8-1 instead of 3-4-5). This aligns the icon, prevents the icon space being used by anything else (keeping icons visible) but minimises it so the controls themselves and their help, get the lions share of screen space.

AdSchellevis:
Sorry, I think the current version looks better and won't get "jumpy" when messages are returned (which is the problem with placing it on top). We probably better leave it as is for now.

stilez:
Makes sense (reluctantly). (Probably mutually reluctantly).

What do you think about at least narrowing it a bit, to say 7:2 rather than 4:5? Reserving >> 50% of width for validation messages seems to be going too far in that direction....?

AdSchellevis:
image
Original looks better (below v.s. top). 
image

We have spend quite some time over the years on layout and consistency, feel free to tryout some things, it might just be difficult to find major improvements here.

stilez:
I can see that. 

Fwiw my personal feeling is that having the controls users need 100% of the time, being  allocated just 45% of dialog space, simply on the off-chance that a validation message might occasionally occur on one of them, and further having it that controls and help text can wrap, but the validation space mustn't wrap...  you can see why I am eyebrow-raising.

That said, different perspectives, and years of experience with it, as you say.  So if its best left as-is, so be it. But if a better way to alert validation messages does come to mind, any time, please do ping me. I'd be glad to do something to improve that aspect. 

Tl:;Dr = yes they're important but surely not so important that the mere possibility implies holding 55% of then page blank?

AdSchellevis:
ok, let's stop this conversation for now and keep it as it is. at the moment it doesn't look like we're going to fix any real issues here. close pr?

stilez:
Yes. I was thinking that, but wanted to hold off closing it before you saw the above. Go ahead.

Issue #2336, opened 2018-04-08
DNS and domain options for DHCPv6 and Router Advertisements, multiple issues


maurice-w:
### Parameters in `dhcpdv6.conf`, configured via `services_dhcpv6.php`
1. OPNsense sets the `domain-name` parameter, both as a global option (system domain) as well as a subnet-specific option (_Domain name_ field). However, this is a DHCPv4 option and is ignored by dhcpd6. DHCPv6 uses the `dhcp6.domain-search` option only.
1.1 The global `domain-name` option (set to the system domain) should be renamed to `dhcp6.domain-search`.
1.2 The _Domain name_ field should be removed from `services_dhcpv6.php`.
1.3 The _Domain search list_ field should set the subnet-specific `dhcp6.domain-search` option. Currently it sets the `domain-search` option (which is ignored by dhcpd6 because it's a DHCPv4 option).
2. If no _Range_ is specified, all subnet-specific options are missing. The options should always be there (to allow stateless DHCPv6 mode).
3. There are two additional global DHCPv4 options which should be removed:
`option ldap-server code 95 = text;`
`option domain-search-list code 119 = text;`

Reference (ISC DHCP 4.3 Man pages):
https://www.isc.org/wp-content/uploads/2017/08/dhcp43.html
https://www.isc.org/wp-content/uploads/2017/08/dhcp43options.html

### Parameters in `radvd.conf`, configured via `services_router_advertisements.php`
1. RDNSS
1.1 If addresses are entered into the _DNS servers_ fields, the `RDNSS` option is missing. It should be set to the specified addresses.
1.2 If the _DNS servers_ fields are left blank and Unbound is disabled, the `RDNSS` option is missing. It should be set to the IPv6 DNS servers configured via `system_general.php`.
1.3 Enabling _Use the DNS settings of the DHCPv6 server_ has no effect. It should set `RDNSS` to the DNS servers configured via `services_dhcpv6.php`.
2. DNSSL
2.1 The _Domain search list_ field is supposed to set `DNSSL` to the specified domains, but has no effect. `DNSSL` is always set to the system domain.
2.2 If a _Domain name_(!) is specified via `services_dhcpv6.php`(!), `DNSSL` is set to this domain. This should... not happen.

[...]

PR #3822, opened 2019-11-20
DHCPv6, fix for domain-search option

maurice-w:
The "domain-search" option in dhcpdv6.conf must be prefixed with "dhcp6". This fixes one of the issues from #2336 which are still open.


fichtner:
Nice catch, thanks!

PR #3824, opened 2019-11-20
DHCPv6, replace domain name option with domain search list option

maurice-w:
Should fix the remaining issues from #2336.


fichtner:
Many thanks, looks good ❤️

PR #3837, opened 2019-12-02
Fix stateless DHCPv6 mode

maurice-w:
Always add dhcp6.domain-search and dhcp6.name-servers to dhcpdv6.conf, even if no range6 is specified.


fichtner:
Nice catch, thanks!

PR #3845, opened 2019-12-10
core: add method to get product info in templates

fabianfrz:
It may not be perfect but it does its job.

It is required that @mimugmail can do his patch in opnsense/plugins#1602.

It adds a helpers.systemInfo hash, which contains plugin versions and core metadata (like the release)

### DIFF
--- a/src/opnsense/service/modules/addons/template_helpers.py
+++ b/src/opnsense/service/modules/addons/template_helpers.py
@@ -32,6 +32,7 @@ import os
 import glob
 import collections
 import netaddr
+import ujson
 
 
 class SortKeyHelper:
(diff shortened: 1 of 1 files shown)


AdSchellevis:
I'm not sure we should access the internal version information like this, most tools can use `opnsense-version` inline to abstract specific information about our system.

The code itself could be simplified as well by the way. We'll take it under consideration.

fabianfrz:
@AdSchellevis For the specific case, @mimugmail needs to render the system version into the template so he cannot call a shell script. This way he has a generic access to all system version properties and it can be used as well to check for plugin versions (is installed in templates to check soft dependencies)

AdSchellevis:
we'll take it under consideration, there are likely more ways to fix this, not sure this is the one we prefer. When I see @mimugmail around, I'll ask about his plans. We don't have use-cases for this ourselves at the moment.

AdSchellevis:
In net-snmp's case I expect you can configure the contents of sysDescr in other ways as well, the default seems to be `uname -a`. Changing our ident to `OPNsense` might have other consequences, although it might also be an option (although @fichtner is more of an expert on this subject)

fichtner:
The most descriptive and portable output is from „opnsense-version“ itself, albeit already a composite of build info included in the core json file and subject to change. I don’t see a robust way other than getting the output of the script directly. Sure, there is always more detailed info to be required eventually, but we don’t need to pollute the template with all  sorts of variables that nobody will use.

AdSchellevis:
discussed internally, we're not going to add this now.

Issue #3821, opened 2019-11-18
Allow strongSwan's dpdaction set to be "clear" if the connection method is set to "Start on traffic"


WRMSRwasTaken:
**Important notices**
Before you add a new report, we ask you kindly to acknowledge the following:

[x] I have read the contributing guide lines at https://github.com/opnsense/core/blob/master/CONTRIBUTING.md

[x] I have searched the existing issues and I'm convinced that mine is new.

**Is your feature request related to a problem? Please describe.**
It's (hopefully) not a problem.

**Describe the solution you'd like**
I am running a CARP cluster which is initiating IPsec connections to another peer.
Due to the fact that we have the database clients on our side, initiating the connection to the database server behind the peer Firewall is mandatory for us. Having the connection open constantly is not an option, because the peer closes SAs after 30 minutes of inactivity.
This is why the connection method is set to `Start on traffic` instead of `Respond only`.

Unfortunately, setting the connection method to `Start on traffic` sets strongSwan's `dpdaction` to `restart` and only setting it to `Respoond only` sets it to `clear`.

This is not much of an issue, but when the firewall does a CARP failover, the new master node initiates the connection and the old node runs into a DPD timeout, successfully clearing all SA entries. But it keeps on trying to re-establish the connection and it fails, because the own IPsec peer IP is set to the CARP virtual IP and the new master node receives the response DPD packets.

The previous node is now the standby node, thus not receiving IPsec traffic now, so it should close the IPsec tunnel istead of trying to reconnect it, until it becomes master again, receiving traffic for the IPsec tunnel. For now, I have to manually stop the tunnel on that backup node by hand in the web UI.

I can't tell (for now) if the connection from the master node is going to run into issues, if the opposite peer still receives the DPD packets from the new standby node.

**Describe alternatives you've considered**
Add an Option to configure strongSwan's `dpdaction` config option in the web UI, or set `dpdaction` to `clear` if the connection method is set to `Start on traffic`. Then it should automatically re-establish the tunnel again anyway, when the tunnel receives traffic.

Another option would be to make strongSwan's `inactivity` option configurable. Then it'd close down the tunnel after the timespan set in `inactivity` anyway, because it's the standby node.

This is somewhat related to https://github.com/opnsense/core/issues/3291

AdSchellevis:
If the (new) backup is still sending out messages (successfully) to the other end, there's likely another issue. The carp address should be used to send out traffic, either by binding it to the address or using an outbound nat rule).

[...]

PR #3851, opened 2019-12-21
Make strongSwan's dpdaction configurable

WRMSRwasTaken:
As stated in #3821, making the `dpdaction` configurable comes in handy for IPsec `start on traffic` CARP clusters for a clean failover.

I've added a `default` option, to not break existing setups.


AdSchellevis:
@WRMSRwasTaken this is probably the most sensible way to add it as an option. thanks!

mimugmail:
Nice addition :)

fichtner:
According to the manual there is also "none" and "hold" which could be supported:

https://wiki.strongswan.org/projects/strongswan/wiki/Connsection

AdSchellevis:
none is the default when dpdaction is enabled, hold might be an option, although I've no idea how well it could work.

PR #3860, opened 2019-12-28
PF API

fabianfrz:
Work In Progress

### DIFF
--- a/src/opnsense/mvc/app/models/OPNsense/Base/FieldTypes/FilterProtocolField.php
+++ b/src/opnsense/mvc/app/models/OPNsense/Base/FieldTypes/FilterProtocolField.php
@@ -1,0 +1,145 @@
+<?php
+
+/**
+ *    Copyright (C) 2015-2019 Deciso B.V.
+ *    Copyright (C) 2020 Fabian Franz
+ *    Copyright (C) 2004-2007 Scott Ullrich <sullrich@gmail.com>
+ *    Copyright (C) 2005 Bill Marquette <bill.marquette@gmail.com>
+ *    Copyright (C) 2006 Peter Allgeyer <allgeyer@web.de>
+ *    Copyright (C) 2008-2010 Ermal Luçi
+ *    Copyright (C) 2003-2004 Manuel Kasper <mk@neon1.net>
+ *
+ *    All rights reserved.
+ *
+ *    Redistribution and use in source and binary forms, with or without
+ *    modification, are permitted provided that the following conditions are met:
+ *
+ *    1. Redistributions of source code must retain the above copyright notice,
+ *       this list of conditions and the following disclaimer.
+ *
+ *    2. Redistributions in binary form must reproduce the above copyright
+ *       notice, this list of conditions and the following disclaimer in the
+ *       documentation and/or other materials provided with the distribution.
+ *
+ *    THIS SOFTWARE IS PROVIDED ``AS IS'' AND ANY EXPRESS OR IMPLIED WARRANTIES,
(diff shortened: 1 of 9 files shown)


src/opnsense/service/conf/actions.d/actions_filter.conf
```diff
@@ -112,3 +112,9 @@ command:/usr/local/opnsense/scripts/filter/find_table_references.py
 parameters: %s
 type:script_output
 message:check which aliases match the given IP
+
+[gateways.list]
```
AdSchellevis: duplicate call, already in core (including script) https://github.com/opnsense/core/blob/bb35688f0ea3a88bf6a75923447eab6939a3871f/src/opnsense/service/conf/actions.d/actions_interface.conf#L91-L94

AdSchellevis:
@fabianfrz you better start in plugins, we can always decide if code is suitable for core later (and move/modify if needed). Some of the copy-right notices contain names that don't really have relation to the code if I'm seeing it correctly, you might want to take a look at that too.

src/opnsense/mvc/app/models/OPNsense/Base/FieldTypes/FilterProtocolField.php
```diff
@@ -0,0 +1,145 @@
+<?php
+
+/**
+ *    Copyright (C) 2015-2019 Deciso B.V.
+ *    Copyright (C) 2020 Fabian Franz
+ *    Copyright (C) 2004-2007 Scott Ullrich <sullrich@gmail.com>
+ *    Copyright (C) 2005 Bill Marquette <bill.marquette@gmail.com>
+ *    Copyright (C) 2006 Peter Allgeyer <allgeyer@web.de>
+ *    Copyright (C) 2008-2010 Ermal Luçi
+ *    Copyright (C) 2003-2004 Manuel Kasper <mk@neon1.net>
```
fichtner: Personally, this feels like a wasted opportunity to cleanly rebuild the code. Already nobody knows what these copyrights were for in terms of code and it's historical dead weight for going forward.

src/opnsense/service/conf/actions.d/actions_filter.conf
fabianfrz: thanks, will replace it

src/opnsense/mvc/app/models/OPNsense/Base/FieldTypes/FilterProtocolField.php
fabianfrz: I copied a single function and I cannot identify who is the originator so I have to take all of them with it…

fichtner: So you don't copy the function.

fabianfrz: I just need a variant of getProtocols(), which is from here: get_protocols from src/etc/inc/filter.inc

I need a hash structure instead of an array so I have to convert that anyway (means it is not totally identical to the original).
The original also has some issues with up/downcase handling because the legacy code seems to handle the return value differently.

src/opnsense/service/conf/actions.d/actions_filter.conf
fabianfrz: They are different - I keep it for now and delete it if the other function is enough

Screenshot_20191229_165550

AdSchellevis:
superseded by https://github.com/opnsense/plugins/issues/1720

PR #3865, opened 2019-12-30
Unbound: Add options for logging queries and extended statistics.

Flightkick:
The 'custom options' field is deprecated + does not work for these fields since it places the contents out of the 'server:' context. E.g. adding:
```
log_queries: yes
extended_statistics: yes
```
To the field creates an invalid config which blocks the service from starting.

These boolean fields allow the options to be used and are future proof in case the custom options field gets removed.

### DIFF
--- a/src/www/services_unbound_advanced.php
+++ b/src/www/services_unbound_advanced.php
@@ -358,7 +358,7 @@ include_once("head.inc");
                           <td>
                               <input name="extended_statistics" type="checkbox" id="extended_statistics" value="yes" <?= empty($pconfig['extended_statistics']) ? '' : 'checked="checked"' ?> />
                               <div class="hidden" data-for="help_for_extended_statistics">
-                                  <?= gettext("If  enabled, extended  statistics are printed. Default is off, because keeping track of more statistics takes time.") ?>
+                                  <?= gettext("If enabled, extended statistics are printed.") ?>
                               </div>
                           </td>
                       </tr>
--- a/src/www/services_unbound_advanced.php
+++ b/src/www/services_unbound_advanced.php
@@ -367,7 +367,7 @@ include_once("head.inc");
                           <td>
                               <input name="log_queries" type="checkbox" id="log_queries" value="yes" <?= empty($pconfig['log_queries']) ? '' : 'checked="checked"' ?> />
                               <div class="hidden" data-for="help_for_log_queries">
-                                  <?= gettext("Prints one line per query to the log, with the log timestamp and IP address, name, type and class.  Default is no.  Note that  it takes time to print these lines which makes the server (significantly) slower.  Odd  (nonprintable)  characters  in  names  are printed as '?'.") ?>
+                                  <?= gettext("If enabled, prints one line per query to the log, with the log timestamp and IP address, name, type and class.") ?>
                               </div>
                           </td>
                       </tr>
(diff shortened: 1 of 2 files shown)


src/www/services_unbound_advanced.php
```diff
@@ -350,7 +354,24 @@
...
+                      <tr>
+                          <td><a id="help_for_log_queries" href="#" class="showhelp"><i class="fa fa-info-circle"></i></a> <?= gettext('Log Queries') ?></td>
+                          <td>
+                              <input name="log_queries" type="checkbox" id="log_queries" value="yes" <?= empty($pconfig['log_queries']) ? '' : 'checked="checked"' ?> />
+                              <div class="hidden" data-for="help_for_log_queries">
+                                  <?= gettext("Prints one line per query to the log, with the log timestamp and IP address, name, type and class.  Default is no.  Note that  it takes time to print these lines which makes the server (significantly) slower.  Odd  (nonprintable)  characters  in  names  are printed as '?'.") ?>
+                              </div>
+                          </td>
+                      </tr>
+                      <tr><td></td>
```
fichtner:
```suggestion
                      <tr>
                        <td></td>
```

this should clear up the final diff

src/www/services_unbound_advanced.php
```diff
@@ -350,7 +354,24 @@
                         </td>
                       </tr>
                       <tr>
-                        <td></td>
+                          <td><a id="help_for_extended_statistics" href="#" class="showhelp"><i class="fa fa-info-circle"></i></a> <?= gettext('Extended statistics') ?></td>
+                          <td>
+                              <input name="extended_statistics" type="checkbox" id="extended_statistics" value="yes" <?= empty($pconfig['extended_statistics']) ? '' : 'checked="checked"' ?> />
+                              <div class="hidden" data-for="help_for_extended_statistics">
+                                  <?= gettext("If  enabled, extended  statistics are printed. Default is off, because keeping track of more statistics takes time.") ?>
```
fichtner:
```suggestion
                                  <?= gettext("If enabled, extended statistics are printed.") ?>
```

although the former matches the documentation's wording it is a bit redundant :)

src/www/services_unbound_advanced.php
```diff
@@ -350,7 +354,24 @@
...
+                              </div>
+                          </td>
+                      </tr>
+                      <tr>
+                          <td><a id="help_for_log_queries" href="#" class="showhelp"><i class="fa fa-info-circle"></i></a> <?= gettext('Log Queries') ?></td>
+                          <td>
+                              <input name="log_queries" type="checkbox" id="log_queries" value="yes" <?= empty($pconfig['log_queries']) ? '' : 'checked="checked"' ?> />
+                              <div class="hidden" data-for="help_for_log_queries">
+                                  <?= gettext("Prints one line per query to the log, with the log timestamp and IP address, name, type and class.  Default is no.  Note that  it takes time to print these lines which makes the server (significantly) slower.  Odd  (nonprintable)  characters  in  names  are printed as '?'.") ?>
```
fichtner:
```suggestion
                                  <?= gettext("If enabled, prints one line per query to the log, with the log timestamp and IP address, name, type and class.") ?>
```

src/www/services_unbound_advanced.php
Flightkick: Yes, not part of changed content but nice to clean up anyway. :) well spotted

Flightkick:
👍

fichtner:
Merged, thanks a lot! 😊

PR #3866, opened 2019-12-30
Auto Replace windows DUID dashes

marjohn56:
If adding static mapping v6 addresses then replace the windows dashes with colons automatically. Makes life a lot easier if there's a whole bunch to do.

### DIFF
--- a/src/www/services_dhcpv6_edit.php
+++ b/src/www/services_dhcpv6_edit.php
@@ -85,9 +85,13 @@ if ($_SERVER['REQUEST_METHOD'] === 'GET') {
         $input_errors[] = gettext("A valid IPv6 address must be specified.");
     }
 
-    if (empty($pconfig['duid']) || preg_match('/^([a-fA-F0-9]{2}[:])*([a-fA-F0-9]{2}){1}$/', $pconfig['duid']) !== 1) {
-        $input_errors[] = gettext("A valid DUID Identifier must be specified.");
+    if (!empty($pconfig['duid'])) {
+        $pconfig['duid'] = str_replace("-",":",$pconfig['duid']);
+        if( preg_match('/^([a-fA-F0-9]{2}[:])*([a-fA-F0-9]{2}){1}$/', $pconfig['duid']) !== 1) {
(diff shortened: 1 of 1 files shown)


marjohn56:
Sorry @fichtner - I pushed a new commit without creating a new branch. It's for a different issue. Do you want me to revert it?

src/www/services_dhcp.php
```diff
@@ -87,7 +87,7 @@ function reconfigure_dhcpd()
     dhcp_clean_leases();
     system_hosts_generate();
     clear_subsystem_dirty('hosts');
-    dhcpd_dhcp_configure(false, 'inet');
+    dhcpd_dhcp_configure(false); // and IPv6 else it does not get restarted
```
fichtner: this makes no sense?

src/www/services_dhcpv6.php
```diff
@@ -38,7 +38,7 @@ function reconfigure_dhcpd()
 {
     system_hosts_generate();
     clear_subsystem_dirty('hosts');
-    dhcpd_dhcp_configure(false, 'inet6');
+    dhcpd_dhcp_configure(false); // restart dhcp4
```
fichtner: this makes no sense?

fichtner:
yes, need a clean PR. the current behaviour for restart is intentional...

src/www/services_dhcpv6.php
marjohn56: Try it, if you make a change and save it in dhcp4 then it leaves dhcp6 stopped.

marjohn56:
Restart is OK, if it restarted both.

fichtner:
Haven’t seen or heard this once, but let’s try it.

fichtner:
It’s an issue alright, but just a workaround that doesn’t fix the underlying silliness. Let me fix that. 😊

fichtner:
bbee5f518f96ac should be it

marjohn56:
Nice one... ta muchly.

fichtner:
Merged, thanks!

PR #3870, opened 2020-01-03
Account for case sensitivity of diff function

h-town:
After selecting the 'Synchronize groups' option within 'Servers > LDAP', OPNsense kept stripping me from the 'admins' group. After a great deal of troubleshooting, I determined that the diff function used to examine commonality of group inclusion between LDAP and local is case sensitive, whereas LDAP schemas are not (by default).  It turns out I had added an 'Admins' group to LDAP to conform with the standard capitalization format of LDAP schemas and OPNsense was not recognizing or respecting the sameness between 'Admins' and 'admins' (the latter being OPNsense default).

Noting that duplicate LDAP groups cannot be configured (e.g. attempting to create a group 'admins' when 'Admins' already exists), I've proposed that all user & group variables are normalized with lowercase text prior to analysis.

Although it holds up in various testing scenarios I put it through, this will need to be tested and/or vetted out by developers, as it could be very hacky. Also note that I first installed pkg 'php72-mbstring' in an attempt to address this using mb_strtolower, though I'm not certain this pull request requires it. I'm far from a PHP expert...those 3 "simple" lines took me ~4 hours to arrive at!

### DIFF
--- a/src/opnsense/mvc/app/library/OPNsense/Auth/LDAP.php
+++ b/src/opnsense/mvc/app/library/OPNsense/Auth/LDAP.php
@@ -470,6 +470,11 @@ class LDAP extends Base implements IAuthConnector
             natcasesort($sync_groups);
             natcasesort($user_groups);
             natcasesort($ldap_groups);
+            // convert contents of group arrays to lowercase 
+            $sync_groups = array_change_key_case($sync_groups, CASE_LOWER);
+            $user_groups = array_change_key_case($user_groups, CASE_LOWER);
+            $ldap_groups = array_change_key_case($ldap_groups, CASE_LOWER);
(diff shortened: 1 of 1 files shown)


AdSchellevis:
@h-town https://github.com/opnsense/core/commit/9b72b85e41d4a78d793dbb96e2a3b10e6706e06a should to the trick, since both ends (OPNsense and the ldap provider) need to be matched case-insensitive.

RFC 2307 doesn't seem very clear about case sensitivity (I couldn't find it), but quite some  implementations seem to expect it to be case insensitive (such as AD)

h-town:
fantastic, @AdSchellevis.  thanks for circling back to this!!

PR #3876, opened 2020-01-04
Log access data of proxied end-user

h-town:
At present, visibility of end-user identity is lost within OPNsense when accessing the Administration front-end through reverse proxy. Proposed here is native lighttpd support for capturing these data, which might serve as a baseline for expanded configuration scenarios moving forward. Shown here is the result: https://i.imgur.com/AHwlyTC.png

I have also tested the direct, non-proxied access of OPNsense via the configured lighttpd port (rather than nginx port), which results in the expected logging of end-user IP.

mod_extforward exists in the current version of lighttpd distributed with OPNsense, so no additional components are required beyond this revision to lighty-webConfigurator.conf. 

Of note:
1. I'm not certain that the inclusion of mod_extforward at line 221 is the expected behavior of core developers. Please confirm that this does not belong at line 238, noting that this module **must** be loaded after both mod_openssl and mod_accesslog to function properly.
2. The configuration proposed here supports proxied connections from localhost, which will not meet the needs of all deployments. Although the mod does support _extforward.forwarder = ( "all" => "trust")_ this is clearly a bad path-forward, so consideration may be given to adding a "Support Proxied Connections" option at System: Settings: Administration, wherein the user can select a checkbox and add a list of proxy IPs to allow. The intention of this functionality should be further clarified via help text to avoid user perception of this setting effectively _enabling_ a reverse proxy vs. only supporting connections from them.
3. mod_extforward also supports connections from HAProxy but enabling this configuration is beyond the scope of this pull request. I've acknowledged this observation in the contents of the code.

### DIFF
--- a/src/etc/inc/plugins.inc.d/webgui.inc
+++ b/src/etc/inc/plugins.inc.d/webgui.inc
@@ -218,7 +218,7 @@ fastcgi.server = ( ".php" =>
 
 EOD;
 
-    $lighty_modules = !empty($config['system']['webgui']['httpaccesslog']) ? ', "mod_accesslog"' : "";
+    $lighty_modules = !empty($config['system']['webgui']['httpaccesslog']) ? ', "mod_accesslog", "mod_extforward"' : "";
 
 
     $lighty_config = <<<EOD
@@ -447,6 +447,15 @@ compress.filetype  = ("text/plain","text/css", "text/xml", "text/javascript" )
 {$cgi_config}
 
 expire.url = ( "" => "access 50 hours" )
+
+extforward.params = ("host" => 1, # Overwrite Host with Forwarded value
+                     "remote_user" => 1 # Set REMOTE_USER with Forwarded value
+)
+extforward.forwarder = ("127.0.0.1" => "trust")
+extforward.headers = ("X-Forwarded-For", "Forwarded-For", "X-Real-IP")
+### Acknowledge possible probable need to address HAProxy in a different manner
+# extforward.hap-PROXY = ("enable")
+
 
 EOD;
 


fabianfrz:
I do have something similar in place in the nginx plugin:

https://github.com/opnsense/plugins/blob/master/www/nginx/src/opnsense/service/templates/OPNsense/Nginx/http.conf#L128-L135

Please check out the "HTTP Server" settings. It would be good if the UI and configuration options would match as this functionality is basically the same.

h-town:
Hi @fabianfrz,
I've not been in front of my laptop in the last couple of days so I can check out the full context of the code you linked. From mobile, it appears that's only an NGINX setting that doesn't affect anything on the web-gui / lighttpd side; is that correct? These changes are focused on how the header you're setting there would be captured and logged on the lighttpd side. Unless I'm missing something, lighttpd will still log 127.0.0.1 with that header in place in the absence of mod_extforward.  At least, this is what was occurring in my environment, even after configuring headers within NGINX settings.

fabianfrz:
I do not mean that it has any effect on the web interface directly. I meant that if you want to add the same options to the web interface, then the configuration should look similar to the end user.

In the nginx settings, you can configure trusted proxies, how the headers are named, and so on so if you add the same web interface options under https://github.com/opnsense/core/blob/master/src/www/system_advanced_admin.php, please try to design the UI as close to that as possible.

For example other users may use HAProxy or relayd instead of nginx, then there is no guarantee that the headers are actually set so this feature must be able to be turned off.

h-town:
@fabianfrz In the current build, what is the behavior of web administration server logging when behind HAProxy?  Are proper addresses logged? To reiterate, the proposed change will not affect connectivity or logging of direct or HAProxy/relayd connections--merely revise the scope of what lighttpd looks for to identify real IP.  

They only concern I have with merging without further revision is the (slim) possibility that proxied connections from localhost would _not_ be trusted in some architectures. If your perspective is that it should be a configurable option rather than be hard-coded I'm not sure the UI would be 1:1 between what you have in NGINX (proxy / egress) and lighttpd (ingress)...

AdSchellevis:
Although I'm not against adding features to lighttpd, we should be careful opting-in automatically. closing this for now.

PR #3882, opened 2020-01-13
Update squid.conf for ipv6 listener support

Kimotu:
Add ipv6 addresses to listening interfaces. Also add ipv6 transparent mode listeners:
- Either VIPs if using CARP
- Or interface address if not using CARP

All you need to do is adding a forwarding rule to local interface ip6 address for transparent mode, e.g.

int: lan:
src: lan
dst: any
proto: inet6
dstport: 80
target: lan-ipv6 address

ToDo: Add 'transparentv6_proxy' template in
https://github.com/opnsense/core/blob/master/src/www/firewall_nat_edit.php

### DIFF
--- a/src/opnsense/service/templates/OPNsense/Proxy/squid.conf
+++ b/src/opnsense/service/templates/OPNsense/Proxy/squid.conf
@@ -27,12 +27,41 @@
 {%     endif %}
 {% endif %}
 
+# Setup transparent mode listeners on local scope ipv6 addresses
+{# for CARP use virtual IPs #}
+{%     if helpers.exists('virtualip') %}
+{%         for intf_item in helpers.toList('virtualip.vip') %}
+{%             if intf_item.subnet.find(':') > -1 and intf_item.type == 'single' %}
+{{ listener_config('['+intf_item.subnet+']', OPNsense.proxy.forward.port) }}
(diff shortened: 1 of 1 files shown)


AdSchellevis:
@Kimotu I've pulled the missing ipv6 address in https://github.com/opnsense/core/commit/ae39e9be1773560d9800d42193acde134569e305, transparant mode should use `[::1]` , which is already in there

Kimotu:
@AdSchellevis I am afraid [::1] will not work for transparent proxy due to Freebsd ip6 scopes.

You would forward from global unicast scope to localhost scope which would drop the packet as scope violation. That's the reason I always forward to the global unicast address on the interface where the packet arrived.

The bug report from 2014 (!), updated several times, is still open.

https://bugs.freebsd.org/bugzilla/show_bug.cgi?id=193568

AdSchellevis:
ok, so if I understand you correctly, your suggestion is to drop the `[::1]` in favour of binding the intercept listeners to the configured interfaces as well?

Kimotu:
Right, together with the intercept flag. Because when using transparent ipv6 mode, you have to create separate forwarding rules for each interface that you want to intercept.

AdSchellevis:
sounds good, if you create a new PR removing the `[::1]` and adding the conditional intercept flag, it should be an easy process.

AdSchellevis:
timeout

PR #3892, opened 2020-01-23
- Update maximum MTU for 10Gb NICs

darkfires:
10Gb switches use 9214 MTU on the interfaces vs 9000 for most 1Gb


src/www/interfaces.php
```diff
@@ -908,8 +908,8 @@ interface_sync_wireless_clones($a_interfaces[$if], false);
             $input_errors[] = gettext("A valid MAC address must be specified.");
         }
         if (!empty($pconfig['mtu'])) {
-            if ($pconfig['mtu'] < 576 || $pconfig['mtu'] > 9000) {
-                $input_errors[] = gettext("The MTU must be greater than 576 bytes and less than 9000.");
+            if ($pconfig['mtu'] < 576 || $pconfig['mtu'] > 9214) {
+                $input_errors[] = gettext("The MTU must be greater than 576 bytes and less than 9214.");
```
fichtner: for not running the risk of bouncing translation strings all the time on updates it should use sprintf %s for both the min and the max value

src/www/interfaces_lagg_edit.php
```diff
@@ -115,8 +115,8 @@ function available_interfaces($selected_id=null)
     if (!in_array($pconfig['proto'], $laggprotos)) {
         $input_errors[] = gettext("Protocol supplied is invalid");
     }
-    if (!empty($pconfig['mtu']) && ($pconfig['mtu'] < 576 || $pconfig['mtu'] > 9000)) {
-        $input_errors[] = gettext("The MTU must be greater than 576 bytes and less than 9000.");
+    if (!empty($pconfig['mtu']) && ($pconfig['mtu'] < 576 || $pconfig['mtu'] > 9214)) {
+        $input_errors[] = gettext("The MTU must be greater than 576 bytes and less than 9214.");
```
fichtner: same here

darkfires:
> Please no shell stuff in this PR
I actually didn't mean to do that at all.. don't even understand how that happened

fabianfrz:
Just do a git revert commit I'd and it will be gone again after a push.

However I like the idea of having vtysh in the menu.

AdSchellevis:
cherry-picked in https://github.com/opnsense/core/commit/13f8ba23478e6b3f7cbfc17f594284f2bd3f1b81

PR #3893, opened 2020-01-23
Add a menu option for FRRouting Shell if the frr plugin is installed

darkfires:


darkfires:
> The idea is not bad but as a general rule we do not merge things into core that belong to plugins exclusively.

What about something like this in the plugins.inc.d..
        $services[] = array(
            'description' => gettext('FRRouting Daemon'),
            'configd' => array(
                'restart' => array('quagga restart'),
                'start' => array('quagga start'),
                'stop' => array('quagga stop'),
            ),
            'name' => 'frr',
            'pidfile' => '/var/run/frr/zebra.pid',
            'cli' => '/usr/local/bin/vtysh'
        );

Then have it scan those and add any cli menu options w/ the description?

AdSchellevis:
The way forward would probably be to add some sort of plugin support for menu items, but I expect this won't mature any time soon since there aren’t a lot of other use-cases for it. Closing this now

Issue #3310, opened 2019-03-10
Adding IP Alias to tracking LAN interface breaks tracking


maurice-w:
**Describe the bug**
TL;WR: Adding an IPv6 Alias to a LAN interface which is configured to track an IPv6 WAN interface causes the LAN interface to stop tracking after a reboot. It subsequently only uses the IP Alias and hosts in the LAN lose Internet connectivity.

**To Reproduce**
1. Set the WAN interface IPv6 configuration type to DHCPv6.
2. Configure the LAN interface to track the WAN interface and enable _Manual configuration_. (The ULA prefix which we add in the next step won't be advertised unless we "allow the manual adjustment of DHCPv6 and Router Advertisements", although we don't actually adjust anything manually. This might be considered another bug, but let's not get into this here.)
3. Add a Virtual IP to the LAN interface: Type _IP Alias_, address fd01:2:3:4::1/64 (ULA)
4. On the Dashboard, restart the `radvd` service. This seems to be required to make it pick up the additional interface address and advertise the additional prefix.
5. Check that hosts in the LAN have both GUAs and ULAs and working Internet connectivity. In my tests this worked fine at this point.
5. Reload DHCP on WAN interface or reboot the router.

**Expected behavior**
LAN interface has an auto-generated GUA as well as the manually added ULA and advertises both prefixes. LAN hosts have GUAs and ULAs and working Internet connectivity.

**Actual behavior**
After a WAN DHCP Reload or a router reboot, the LAN interface doesn't track the WAN interface, uses the IP Alias as its only address and only advertises the ULA prefix. Hosts in the LAN only have ULAs and lose Internet connectivity.

**Additional context**
This has been mentioned on the forum [here](https://forum.opnsense.org/index.php?topic=9384) and [here](https://forum.opnsense.org/index.php?topic=10658) but I couldn't find a matching bug report.
This is important because using an IP Alias seems to be the only way to add ULAs to a tracking LAN interface. Using both GUAs and ULAs is recommended when the prefix delegated by the ISP isn't static. (For example, OpenWrt adds ULAs to all LANs by default.)

**Environment**
OPNsense 19.1.3-amd64

**_Update_**
I initially mixed up advertising prefixes and routes. I updated the steps to reproduce the behavior accordingly. This doesn't change anything about the bug itself.

fichtner:
@maurice-w would you mind testing this on 19.1.7's development version (19.7.a_701)?

maurice-w:
[...]

PR #3896, opened 2020-01-28
Vip alias/dhcp6 fix

marjohn56:
Possible fix for issue #3310  - IP Alias & Tracking LAN.

Remove the VIP alias from the interface before dhcp6c applies a new address and restore the VIP alias after.


marjohn56:
@fichtner when you've decided what to do with this and don't want or need it, I'll reset my repo as it looks like its a bit naff, not sure how that Replace windows DUID dashes got onto this commit branch.

AdSchellevis:
I think we should be careful with this, often these kind of constructions tend to have boomerang effects later on (an event removes things it's not really responsible for). I don't have a track interface on our end, but maybe there are other ways to solve the issue.

marjohn56:
OK, only a suggestion.. I'll delete it.

AdSchellevis:
I don't mind keeping it open for a while if it helps pinpoint the issue better, unfortunately I can't reproduce this myself.

marjohn56:
OK, perhaps I didn't explain it very well. What appears to happen is that when the system comes up OR the dhcp6c client gets a change then when a  'dhcp6 configure' fires it's picking the VIP address if its already there, and it is already there, you only have to look at the dhcp6 server interface, it has the VIP address in it. So all this does is remove that alias, and it checks to see if it is an alias and removes it whilst it configures dhcp6c and then puts it back. The result being that dhcp6 then sets itself using the correct addresses and not the alias. It also makes sure the alias is re-applied. There is one extra check I can add, to make sure the alias is in fact a v6 address. If the interfaces is already in a configure state then this should have no side effects. Easy to test, all you need is a second router that slaves off a primary using dhcp6 track, the error shows up instantly as soon as you reboot the slave.

marjohn56:
Cannot re-open this one now as I've cleaned and reset the repo, but I can re-issue it with that further check to make sure the alias is v6.

PR #3898, opened 2020-01-28
VIP alias/dhcp6 fix

marjohn56:
Remove the Interface v6 alias during reconfigure by dhcp6c and re-apply once complete. This commit has a further check that the closed commit I previously issued did not. It now also checks that the alias to be removed and re-added is a v6 address, thus removing the possibility that it could remove a v4 alias on the interface. Whilst the dhcp6c v6 address is being added there should not be an alias on that interface anyway so this makes sure. Ref issue #3310


marjohn56:
Sorry about repeated commits... fubar on my pc!

fichtner:
#3310 should sort this now :)

PR #3906, opened 2020-02-03
Allow for variable debug info from dhcp6c. None, Level 1 or Level 2

marjohn56:
In the current code, debug is either -d or -D, these are both debug levels, Not setting a debug flag at all results in zero debug output. Not totally happy with the way I've implemented the command string in interfaces.inc, but it works, I could not get it to work if the debug was nothing and part of the array. If someone has a better way, go do it.

### DIFF
--- a/src/etc/inc/interfaces.inc
+++ b/src/etc/inc/interfaces.inc
@@ -3044,9 +3044,9 @@ function interface_dhcpv6_prepare($interface = 'wan', $wancfg)
     chmod("/var/etc/dhcp6c_{$interface}_script.sh", 0755);
 
     $dhcp6ccommand = exec_safe(
-        "/usr/local/sbin/dhcp6c %s -c %s -p %s %s",
+        "/usr/local/sbin/dhcp6c {$wancfg['adv_dhcp6_debug']} %s -c %s -p %s %s",
         array(
-            '-' . (empty($wancfg['adv_dhcp6_debug']) ? 'd' : 'D' ) . (!empty($wancfg['dhcp6norelease']) ? 'n' : ''),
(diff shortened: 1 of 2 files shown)


src/etc/inc/interfaces.inc
```diff
@@ -3044,9 +3044,9 @@ function interface_dhcpv6_prepare($interface = 'wan', $wancfg)
     chmod("/var/etc/dhcp6c_{$interface}_script.sh", 0755);
 
     $dhcp6ccommand = exec_safe(
-        "/usr/local/sbin/dhcp6c %s -c %s -p %s %s",
+        "/usr/local/sbin/dhcp6c {$wancfg['adv_dhcp6_debug']} %s -c %s -p %s %s",
```
fichtner: this breaks security by feeding arbitrary values to the command line without escaping

marjohn56:
Yes, that's why I commented I could not get the array way to work, my php skills are limited to 'hello world'

src/etc/inc/interfaces.inc
marjohn56: Yes, that's why I commented I could not get the array way to work, my php skills are limited to 'hello world'

marjohn56: It's a bit screwed then, because with the existing code it's either debug low level or debug high level. you cannot turn off the debug completely because of the '-' at the start if the array string, if both debug and no release are empty then the command fails.

marjohn56:
Sorry, For some unknown reason my core repo was removed from Github. I closed and removed a deleted a branch earlier to do with the RADVD issues, it appears it deleted the repo fork entirely. Whatever I'm closing this commit and issuing a new one.

Issue #3908, opened 2020-02-03
DHCPv6 allows to configure DDNS (Dynamic DNS) but fails to add configuration on config file


trunet:
**Describe the bug**
If you configure Dynamic DNS settings on DHCPv6, only ddns-domainname is being set. No key, no algorithm, nothing.

Would also be nice to add the algorithms drop-down list from https://github.com/opnsense/core/pull/2651 to DHCPv6 settings.

**To Reproduce**
Configure DHCPv6 with Dynamic DNS. Add all 4 settings, only ddns-domainname reaches the configuration file.

**Expected behavior**
If it's on the GUI, this should add the settings to the configuration file.

**Environment**
OPNsense 20.1 (amd64, OpenSSL).

trunet:
BTW, I'll work on that on my fork and will send a PR.

trunet:
The code is very spaghetti :D

Can anybody enlighten me if there's a way to manually (from the CLI) trigger the DHCP config files regeneration and if possible how usually do you debug it?

trunet:
Work in Progress PR: https://github.com/opnsense/core/pull/3909

AdSchellevis:
```
pluginctl dhcp
```

would probably do the trick to restart the dhcp service

trunet:
@AdSchellevis Can you check the PR and see if it's good for you?

AdSchellevis:
At the moment I don't have a lot of time to dig into it. Looking at it briefly I think it could be simplified a bit, the existing legacy code isn't too great, if possible, we try to minimise the amount of new code in there as much as possible.

I'll get back to you on this.

PR #3909, opened 2020-02-03
WIP: Fix for #3908 - DHCPv6 DDNS implementation

trunet:
I would like to leave this PR here as a work in progress.

This DHCP code is very messy (on my opinion) it's being difficult to connect all the dots needed by this change.

If any of you can give me some light, it's more than welcome.

Mainly I would like to know if there's an easy way to call the dhcp config files generation from the CLI and how to get the output of the "print" and "echo" present in the code to "debug".


trunet:
I will open a new final one...

PR #3911, opened 2020-02-05
dhcp: add icons next to online/offline lease status

tyler-ham:
[X] I have read the contributing guide lines at https://github.com/opnsense/core/blob/master/CONTRIBUTING.md

[X] I have searched the existing issues and I'm convinced that mine is new.

**Is your pull request related to a problem? Please describe.**
I find it difficult to distinguish between "offline" and "online" when I look at the list of DHCP Leases. The two words look very similar, and it is hard to scan the column visually.

**Describe the solution**
My commit adds an icon in front of the "online" or "offline" text to make it easier to distinguish the two states from each other at a glance. I chose the `signal` and `ban-circle` glyphicons. I don't have a strong opinion about the particular icons used as long as they make sense and maintain a good enough visual distinction.

**Describe alternatives you've considered**
I played with setting Bootstrap danger/warning/success classes on the status cell (`<td>`) or on the whole row (`<tr>`), but this seemed like overkill and didn't look very good to me.

**Additional context**
Here are before and after screenshots:

without-indicator

with-indicator

### DIFF
--- a/src/www/status_dhcp_leases.php
+++ b/src/www/status_dhcp_leases.php
@@ -421,9 +421,8 @@ legacy_html_escape_form_data($leases);
                   <td><?=$data['descr'];?></td>
                   <td><?= !empty($data['start']) ? adjust_utc($data['start']) : '' ?></td>
                   <td><?= !empty($data['end']) ? adjust_utc($data['end']) : '' ?></td>
-                  <td class="text-nowrap">
-                    <i class="fa fa-<?=$data['online']=='online' ? 'signal' : 'ban';?>"></i>
-                    <?=$data['online'];?>
+                  <td>
+                    <i class="fa fa-<?=$data['online']=='online' ? 'signal' : 'ban';?>" title="<?=$data['online'];?>" data-toggle="tooltip"></i>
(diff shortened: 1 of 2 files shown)


AdSchellevis:
Hi @tyler-ham,

We use font-awesome 4.7 for our icons, but that's probably not to difficult to replace. I think adding the text and the icon is a bit of an overkill, how about adding the label as tooltip and keep the icon? Something like` <i class="fa fa-signal" title="online" data-toggle="tooltip"></i>` would probably do the trick for connected, `fa-times`  is an icon that might be an alternative for the disconnected sign.

image

image

Best regards,

Ad

tyler-ham:
Thanks for the feedback!

The `fa-times` icon looks too clickable in my opinion. I see that font-awesome also has an `fa-ban` icon. What do you think about this?

fa-signal-ban

AdSchellevis:
`fa-ban` works for me too, I see your point, choosing the correct icon is often challenging, we don't have a style guide for it (yet)

AdSchellevis:
@tyler-ham looks good, thanks!

PR #3912, opened 2020-02-06
Make strongSwan's inactivity parameter configurable

WRMSRwasTaken:
Since OPNsense already has a mode to automatically open security associations when needed with the `start on traffic` option, I think it might come in handy for people to automatically tear down security associations when not sending or receiving traffic over a certain timespan (and create those again when needed).

This option is already seen on numerous Firewall vendors, for example on a Cisco ASA.

Luckily, strongSwan offers an option for this via the `inactivity` parameter, so I made it configurable.


AdSchellevis:
@WRMSRwasTaken looks good, thanks!

PR #3914, opened 2020-02-07
Update interfaces.lib.inc

marjohn56:
Quick fix ( maybe ) for GUA addresses in wrong order on widgets and Interface Info.

### DIFF
--- a/src/etc/inc/interfaces.lib.inc
+++ b/src/etc/inc/interfaces.lib.inc
@@ -293,7 +293,11 @@ function legacy_interfaces_details($intf = null)
             }
             if (isset($tmp['subnetbits'])) {
                 $result[$current_interface]["ipv6"][] = $tmp;
-                // sort link local to bottom, leave rest of sorting as-is (primary address on top)
+                // sort link local to the top, this puts GUA addresses first leave rest of sorting as-is (primary address on top)
+                usort($result[$current_interface]["ipv6"], function ($a, $b) {
+                    return  $b['link-local'] - $a['link-local'];
+                });
+                // sort link local to the bottom, leave rest of sorting as-is (GUA primary address on top)
                 usort($result[$current_interface]["ipv6"], function ($a, $b) {
                     return $a['link-local'] - $b['link-local'];
                 });


AdSchellevis:
@marjohn56 isn't that the sort that's already there?

fichtner:
The issue here is we already established we can't trust the ordering of the data, so reordering it to fit our needs will just introduce bugs elsewhere that relies on a certain ordering already short of the spots that we just fixed. :)

AdSchellevis:
I think I see what you mean, but if that's the issue, it's likely easier to capture the order when parsing and sort in one pass using `link-local`, `capture order` next. now it only guarantees link local is bottom. If you want a proposal, just let me know.

fichtner:
@AdSchellevis I wasn't arguing for it, I was arguing against changing this at all actually. the side effect is unpredictable.

AdSchellevis:
@fichtner I know, we partly trust on ordering at the moment indeed, but only to detect if link-local is the only address if I'm not mistaken. Just saying that if we need to know the original position in ifconfig, we might as well add it as an attribute (and use it in sorting).  I don't think ifconfig itself knows the concept of primary, but I might be mistaken there.

fichtner:
That's indeed the issue, the primary is removed and readded, but ifconfig decides to make a VIP alias the primary. It might be that this is dhcp6c's fault for adding an alias, but it's unreliable to fix that (if indeed the case) because other services might do it differently or more or less correct depending on the situation.

AdSchellevis:
I totally agree, primary in this context only means first existing (which doesn't help if you're looking for an entry that's gone ;) ).

marjohn56:
> That's indeed the issue, the primary is removed and readded, but ifconfig decides to make a VIP alias the primary. It might be that this is dhcp6c's fault for adding an alias, but it's unreliable to fix that (if indeed the case) because other services might do it differently or more or less correct depending on the situation.

dhcp6c does not add the alias, the alias is there either before dhcp6c adds the address or its left there when dhcp6c updates and removes the GUA before adding the new one. That's why un my original patch I removed the aliases before dhcp6c assigned the GUA, then put them back.

fichtner:
Alias here means just an additional address, not the primary one like you assign statically. If dhcp6c adds the primary IP as an alias (ifconfig alias notation) as well ifconfig is confused because there is no notion of a primary IP anywhere so the ordering is just random as we are seeing.

fichtner:
Maybe to illustrate: IPv4 DHCP doesn't have this odd behaviour it seems and the primary IP assigned is not flushed out by additional aliases.

marjohn56:
If you manually add a v6 alias, which is the same as add, then remove the GUA, then put back the GUA, you'll find the GUA is at the bottom of the listed addresses when you do ifconfig, so it's just FreeBSD is agnostic about the order of addresses on the interfaces. Do the same thing with an IPv4 address and the same thing happens, add a new address, remove the old address, put back the old address  and it's bottom of the list.

marjohn56:
Interestingly they have the same problem over the road, and have had for several years. 

https://redmine.pfsense.org/issues/5999

fichtner:
d3b5d9ede21b64 might just do the trick for the interface stats

PR #3916, opened 2020-02-10
radvd: don't set Router Address flag

maurice-w:
The Router Address flag "indicates that the Prefix field contains a complete IP address assigned to the sending router" (RFC 6275). This does not apply, we only send a prefix. This flag is only relevant for Mobile IPv6.
It has been there (for unknown reasons) since pfSense switched to radvd 8 years ago: https://github.com/pfsense/pfsense/commit/3f9cc8e44c5b50e588f0f916611ffa37f7ae0bcb


fichtner:
Merged, thanks!

Will give it a go at the office this week, but it looks sane. :)

fichtner:
PS: Looks like this has always been propagated in tutorials... https://tldp.org/HOWTO/Linux+IPv6-HOWTO/ch22s04.html

maurice-w:
Yeah, it seems people copied that from copies from copies from copies without ever questioning what it actually does. And as we know, if you just repeat something often enough it becomes "the truth"...

fichtner:
at least in this case it is a harmless copy cult :)

PR #3918, opened 2020-02-11
Virtual IPv6 prefix advertising, add help text

maurice-w:
As discussed in #3310.

### DIFF
--- a/src/www/firewall_virtual_ip.php
+++ b/src/www/firewall_virtual_ip.php
@@ -244,7 +244,7 @@ $main_buttons = array(
             print_info_box($savemsg);
           else
           if (is_subsystem_dirty('vip'))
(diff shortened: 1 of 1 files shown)


fichtner:
I am not sure what @AdSchellevis thinks but if find this hint too circumstantial to be of use. It may be better suited for an IPv6 tutorial that deals specifically with ULA aw VIP to accompany a tracked WAN prefix.

AdSchellevis:
I agree, better leave these things in the docs, either https://docs.opnsense.org/manual/firewall_vip.html explaining when/why this matters or as @fichtner suggested an additional tutorial page

maurice-w:
This is not specific to ULAs or tracking interfaces, it affects any VIP on any interface (including static ones). They don't fully work until you manually restart radvd. This doesn't seem intuitive. Took me a while to figure it out. A hint in the UI would have really helped.

AdSchellevis:
better keep this in the docs, although I understand it might help some it also triggers questions by others.

PR #3919, opened 2020-02-11
Virtual IPv6 prefix advertising on tracking LAN interfaces, add help …

maurice-w:
As discussed in #3310.


fichtner:
This is relevant for DHCPv6/Router Advertisements, not for interfaces in general. Also better suited in the documentation.

maurice-w:
It just doesn't seem intuitive that VIP functionality depends on this setting, so a hint somewhere in the UI would be helpful. But if you'd rather enable VIPs in automatic mode (as I think you suggest in #3920), I'm all for that.

fichtner:
Added support for VIPs in automatic case in master

maurice-w:
Did a quick test with 20.7.a_43 and it seems to work. Thanks!

fichtner:
Yay, thanks for testing. ❤️

PR #3920, opened 2020-02-11
radvd.conf for tracking LAN interfaces, cleanup

maurice-w:
As discussed in #3310. For tracking LAN interfaces with manual configuration disabled, some odd radvd configuration choices were made:
MinRtrAdvInterval / MaxRtrAdvInterval were set to very low values (3 / 10) for no apparent reason. Now removed so radvd defaults (200 / 600) will be used.
The DHCPv6 server is enabled and configured with a range6, but the Managed flag was not set. Now set to on.
DeprecatePrefix was only set if the IPv6 configuration type of the tracked WAN interface was SLAAC. Now always set to on.

### DIFF
--- a/src/etc/inc/plugins.inc.d/dhcpd.inc
+++ b/src/etc/inc/plugins.inc.d/dhcpd.inc
@@ -480,16 +480,15 @@ function dhcpd_radvd_configure($verbose = false, $blacklist = array())
         $radvdconf .= "# Generated config for {$autotype} delegation from {$trackif} on {$if}\n";
         $radvdconf .= "interface {$realif} {\n";
         $radvdconf .= "\tAdvSendAdvert on;\n";
-        $radvdconf .= "\tMinRtrAdvInterval 3;\n";
-        $radvdconf .= "\tMaxRtrAdvInterval 10;\n";
         $radvdconf .= sprintf("\tAdvLinkMTU %s;\n", !empty($mtu) ? $mtu : 1280);
+        $radvdconf .= "\tAdvManagedFlag on;\n";
         $radvdconf .= "\tAdvOtherConfigFlag on;\n";
         $radvdconf .= "\tprefix {$networkv6} {\n";
         if ($autotype == 'slaac') {
             /* XXX also of interest in the future, see hardcoded prefix above */
             $radvdconf .= "\t\tBase6Interface $realtrackif;\n";
-            $radvdconf .= "\t\tDeprecatePrefix on;\n";
         }
+        $radvdconf .= "\t\tDeprecatePrefix on;\n";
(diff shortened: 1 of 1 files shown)


src/etc/inc/plugins.inc.d/dhcpd.inc
```diff
@@ -480,16 +480,15 @@ function dhcpd_radvd_configure($verbose = false, $blacklist = array())
...
         $radvdconf .= sprintf("\tAdvLinkMTU %s;\n", !empty($mtu) ? $mtu : 1280);
+        $radvdconf .= "\tAdvManagedFlag on;\n";
         $radvdconf .= "\tAdvOtherConfigFlag on;\n";
         $radvdconf .= "\tprefix {$networkv6} {\n";
         if ($autotype == 'slaac') {
             /* XXX also of interest in the future, see hardcoded prefix above */
             $radvdconf .= "\t\tBase6Interface $realtrackif;\n";
-            $radvdconf .= "\t\tDeprecatePrefix on;\n";
         }
+        $radvdconf .= "\t\tDeprecatePrefix on;\n";
```
fichtner: I know we set it for SLAAC when we included this but Is this wise to move to general use? Won't this cause intermittent connectivity issues?

fichtner: Well, according to previous work the choice we have to make is do we want to say we support CARP only for manual tracking use and go ahead with this. It might be worth adding VIPs to automatic case too and be done with the difference.

maurice-w: It's always on in manual mode and on static interfaces, except for CARP(?) where it's always off.
It's always off in automatic mode, except for SLAAC WAN where it's always on.
We can have a discussion whether it should be on or off (or better: configurable), but I suggest it should at least be consistent.

> It might be worth adding VIPs to automatic case too and be done with the difference.

👍

fichtner:
Ok, thanks! So to conclude: automatic tracking mode cannot be used with CARP/HA/LAN combos. You need to set manual for this and set the correct interface in the DHCP/RADVD settings.

marjohn56:
Just a quick comment here on the min max interval, it may have been set low on automatic tracking to allow for ISPs who only have a very short lease time and may change the PD. Does a change in PD cause a re-write of the RADVD conf and restart or SIGHUP?

fichtner:
well, it was kind of tied to the mobile IPv6 "use" where the intervals are shorter by default. we changed it to defaults in manual radvd a while back and there were no complains as far as I can remember.

marjohn56:
OK, only time will tell. :)

fichtner:
just to be sure this will sit on -devel for a while longer, maybe 20.1.2 or 20.1.3

PR #3923, opened 2020-02-11
Allow multiple dhcp6c debugging levels.

marjohn56:
dhcp6c has two levels of debugging or NO debugging,. Specifying -d or -D gives you the two levels. Omitting it completely gives no debug logging, meaning it's quiet.

Due to the escaping of the existing commands in interfaces inc I have had to add a check to see if both debugging and no-release are empty, thus the two exec_safe options.

This replaces the commit I had to close.

#3934 PR replaces this PR and adds multiple dhcp6c WAN. Use of both PRs will require cherry picking.


AdSchellevis:
Is this still active?

marjohn56:
It's built into other updates... IIRC!

PR #3931, opened 2020-02-13
(Auth) fix expiry and validity for vouchers

### LINKED ISSUE #3930: Sessions stay valid after validity ends if expiry date for voucher is set
If we generate Vouchers and set an expiry date, the expiry date will be set as session timeout, instead of the validity time.

This is due to the following part in the code
```php
if ($row['expirytime'] > 0 && $row['expirytime'] > time()) {
    $this->lastAuthProperties['session_timeout'] = $row['expirytime'] - time();
    return true;    
} elseif (time() - $row['starttime'] < $row['validity']) {
    $this->lastAuthProperties['session_timeout'] = $row['validity'] - (time() - $row['starttime']);
    return true;    
}
```
from [src/opnsense/mvc/app/library/OPNsense/Auth/Voucher.php](https://github.com/opnsense/core/blob/2856288bbf1b42a9b9e06959c308b89f4a208acb/src/opnsense/mvc/app/library/OPNsense/Auth/Voucher.php#L424) because of the precedence of expiry time over validity.

**This is security relevant**

**To Reproduce**
1. Go to **Services** -> **Captive Portal** -> **Vouchers**
2. Click on **Create vouchers**
3. Create Voucher with
  * **Validity**: Custom(minutes) 5
  * **Expires in**: 3 month
  * **Number of vouchers**: 1 (should be irrelevant)
  * **Groupname**: whatever
4. Use the generated voucher, Session won't be invalidated/cleaned up after 5 Minutes, instead it will wait for expiry, so validity will be 3 month

**Expected behavior**
Session will be invalidated after validity has reached its end, not after expiry of the voucher itself.

[✓] I have read the contributing guide lines at https://github.com/opnsense/core/blob/master/CONTRIBUTING.md
[✓] I have searched the existing issues and I'm convinced that mine is new.

xx4h:
Always check if voucher is not expired (either
because of never expires or because of expiry
date is in the future) and ensure session timeout
will be the lowest of validity based on the first
usage, the starttime or expiry (if not never expires).
If one of those conditions is not true, reject
authentication.

Fix #3930

### DIFF
--- a/src/opnsense/mvc/app/library/OPNsense/Auth/Voucher.php
+++ b/src/opnsense/mvc/app/library/OPNsense/Auth/Voucher.php
@@ -421,11 +421,15 @@ class Voucher extends Base implements IAuthConnector
                     $row['starttime'] = time();
                     $this->setStartTime($username, $row['starttime']);
                 }
-                if ($row['expirytime'] > 0 && $row['expirytime'] > time()) {
-                    $this->lastAuthProperties['session_timeout'] = $row['expirytime'] - time();
-                    return true;
-                } elseif (time() - $row['starttime'] < $row['validity']) {
-                    $this->lastAuthProperties['session_timeout'] = $row['validity'] - (time() - $row['starttime']);
+                $is_never_expire = ($row['expirytime'] === 0);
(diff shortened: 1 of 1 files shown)


AdSchellevis:
@xx4h thanks!

PR #3934, opened 2020-02-16
dhcp6c multiwan support

marjohn56:
Allow multiwan support for dhcp6c. All interfaces must be scanned to see whether dhcp6c is in use on that interface, the conf file is appended on each iteration. I have allowed for advanced mode too. In  file override mode there can be only one file that must handle all interfaces, thus I need to add a check that if an override file is specified on any interface, all other interfaces must show that the override is in place.. WIP.

However, the override may no longer be needed as you can do everything without the need to override the file. Tested and working on my system. Work may be needed around the firewall, but on my test system it all appears to work fine.


src/etc/inc/interfaces.inc
```diff
@@ -2940,15 +2941,138 @@ function interface_dhcpv6_configure($interface = 'wan', $wancfg)
     mwexec("/var/etc/rtsold_{$realifv6}_script.sh");
 }
 
+
```
fichtner, line 2944:
```suggestion
```

src/etc/inc/interfaces.inc
```diff
@@ -922,7 +922,8 @@ function interfaces_configure($verbose = false)
             continue;
         } elseif ($is_dhcp6c) {
             /* dhcp6c interfaces load last */
-            $dhcp6c[] = $if;
+            $dhcp6c[] = $if;         
+
```
fichtner, line 926:
```suggestion
            $dhcp6c[] = $if;
```

src/etc/inc/interfaces.inc
```diff
@@ -2960,113 +3082,51 @@ function interface_dhcpv6_prepare($interface = 'wan', $wancfg)
...
         if (is_numeric($wancfg['dhcp6-ia-pd-len'])) {
-            $dhcp6cconf .= "  send ia-pd 0; # request prefix delegation\n";
+            $dhcp6cconf .= "  send ia-pd {$count}; # request prefix delegation\n";
         }
 
         $dhcp6cconf .= "  request domain-name-servers;\n";
         $dhcp6cconf .= "  request domain-name;\n";
         $dhcp6cconf .= "  script \"/var/etc/dhcp6c_{$interface}_script.sh\"; # we'd like some nameservers please\n";
-
         $dhcp6cconf .= "};\n";
```
fichtner, line 3094:
```suggestion

        $dhcp6cconf .= "};\n";
```

src/etc/inc/interfaces.inc
```diff
@@ -3092,7 +3153,7 @@ function DHCP6_Config_File_Advanced($interface, $wancfg, $wanif)
         $script = "  script \"{$wancfg['adv_dhcp6_interface_statement_script']}\";\n";
     }
 
-    $interface_statement  = "interface {$wanif} {\n";
+    $interface_statement  = "interface {$realif} {\n";
```
fichtner, line 3156: isn't $wanif $realif already?

src/etc/inc/plugins.inc.d/dhcpd.inc
```diff
@@ -462,13 +462,30 @@ function dhcpd_radvd_configure($verbose = false, $blacklist = array())
...
+            $hasipv6 = true;
+         }
+
+        foreach (config_read_array('virtualip', 'vip') as $vip) {
+            if ($vip['interface'] != $if || !is_ipaddrv6($vip['subnet'])) {
+                continue;
+            } else {
+                $hasipv6 = true;
+            }
+        }
```
fichtner, line 480: interfaces_primary_address  will not emit a VIP so I think this can be removed....

src/etc/inc/plugins.inc.d/dhcpd.inc
```diff
@@ -462,13 +462,30 @@ function dhcpd_radvd_configure($verbose = false, $blacklist = array())
 
         $dnslist = array();
 
+        /* XXX We'll skip this interface unless a valid address can be found for it. */
+
+        $hasipv6 = false; 
         list ($ifcfgipv6, $networkv6) = interfaces_primary_address6($if, $realif, $ifconfig_details);
 
+        if (is_ipaddrv6($ifcfgipv6)) {
```
fichtner, line 470: we can skip if going forward makes no sense. I don't like changing the indent / conditional block stuffing of working code. it introduces edge cases...

src/etc/inc/interfaces.inc
```diff
@@ -2946,9 +2946,131 @@ function interface_dhcpv6_prepare($interface = 'wan', $wancfg)
...
+
+    log_error("dhcp6c command = {$dhcp6ccommand}"); // debug purposes only
+
+    $rtsoldscript = "#!/bin/sh\n";
+    $rtsoldscript .= "# this file was auto-generated, do not edit\n";
+    $rtsoldscript .= "if [ -n \"\${2}\" ]; then\n";
+    $rtsoldscript .= "\techo \${2} > /tmp/{$wanif}_routerv6\n";
+    $rtsoldscript .= "\techo \${2} > /tmp/{$wanif}_defaultgwv6\n";
+    $rtsoldscript .= "fi\n";
+    $rtsoldscript .= "val=`pgrep dhcp6c`\n";
```
fichtner, line 3049: moving blocks, shell code quality decline. If we need to move code fine, if we end up replacing working code, please no.

marjohn56: OK, so if we don't replace code then there's no multwan dhcp6; there can be only one client and we have to make certain that that is the case. However we can have more than rtsold script, thus we have to replace the code.

fichtner: It's what I'm saying. If we don't put work into cleaning this up and/or refactoring step by step there's no multiwan dhcp6. We have tens of thousands of angry IPv6 users already who require us to not mess up single dhcp6 in the process.

marjohn56: No I totally agree, baby steps. I was going to suggest we first refactor the dhcp6c stuff anyway. Shall I create a PR with just the dhcp6c stuff and we can pull that to pieces and make sure it fits the needs now and in the future, possible breaking it out into plugins.inc.d?

fichtner: Yes, please. good idea. I don't mind breaking master with the dhcp6c pid file change. Second WAN doesn't work anyway and more pressure to sync up with 20.7. What do you think?

marjohn56: OK, well I've just taken the interfaces.inc from 20.7 and applied the changes needed to enable multwan there. Just completing testing and it looks good. 

BTW, I'm testing it in the following way. I have multiple VLANs coming off the primary router, each of those VLANs has a different V6 range. I then have my test router with two WANs plugged into separate VLANs and two LANs outputting to further ports on my switch. I can therefore test that the WANs are getting addresses correctly and that the LAN's are also being given the correct addresses by dhcp6c.

I realise that looking at a whole load of code changes is not nice if you don't know the thought behind it and what's related to what. I'll PR the new file now and we'll just work our way through it before moving onto the other stuff. There is t he extra addition of the debug level feature added to interfaces.php, I'll add that as well; I believe that I have worked my way around the things you didn't like when I did the original PR for that.

marjohn56: PR #4082 issued. So let's work our way through the changes step by step so we all understand what each addition/change does and why.

marjohn56: Remember I am on furlough at present, and apart from some webinars etc that I'm doing I have time to play with this, at least until the lockdown is over.

fichtner:
superseded by other recently merged PRs

Issue #3933, opened 2020-02-15
Feature: DNS64 (unbound)


maurice-w:
[x] I have read the contributing guide lines at https://github.com/opnsense/core/blob/master/CONTRIBUTING.md

[x] I have searched the existing issues and I'm convinced that mine is new.

**Is your feature request related to a problem? Please describe.**
Since NAT64 support is now coming (#167, opnsense/plugins#16), we also need DNS64.

**Describe the solution you'd like**
Unbound fully supports DNS64, we only need options in the UI to enable and configure it.

**Additional context**
Two options in `unbound.conf` are required [1]:
* Add `dns64` to `module-config` to enable it.
* Optionally specify a `dns64-prefix` (64:ff9b::/96 will be used if not specified).

I tested this by hacking the required options into the conf and it works. I think I can make the required changes to `services_unbound.php` and `unbound.inc`, but first I'd like some feedback whether such a PR would be accepted.

[1] https://nlnetlabs.nl/documentation/unbound/unbound.conf/

PR #3937, opened 2020-02-16
unbound: remove 192.0.0.0/24 from rebinding prevention list

maurice-w:
192.0.0.0/24 is not a private net. ipv4only.arpa resolves to 192.0.0.170 and 192.0.0.171. This is required for DNS64 prefix detection (RFC7050).
(Quick way to verify: Lookup ipv4only.arpa via Google DNS / Cloudflare DNS / whatever. Resolves to 192.0.0.170 / 192.0.0.171. But doesn't resolve via OPNsense's unbound.)

### DIFF
--- a/src/etc/inc/plugins.inc.d/unbound.inc
+++ b/src/etc/inc/plugins.inc.d/unbound.inc
@@ -128,7 +128,6 @@ private-address: 100.64.0.0/10
 private-address: 127.0.0.0/8     # Loopback Localhost
 private-address: 169.254.0.0/16
 private-address: 172.16.0.0/12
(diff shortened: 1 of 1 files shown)


fichtner:
we do treat it as bogon for technical reasons so I am reluctant to open the flood gates here. is there a bug associated with this?

maurice-w:
Technical reasons?
I'd like to enable DNS64 support in unbound (#3933). This works fine, but DNS64 prefix detection requires ipv4only.arpa to be resolvable. Android does this for example. If it can't resolve ipv4only.arpa, it won't enable its CLAT.

fichtner:
Ok, there's no apparent consensus on this and since its required for a feature I've merged it. Thanks!

PR #3938, opened 2020-02-17
unbound: add option to disable ptr records on Overrides/Aliases

devNan0:
**Problem:**
I don´t always need/want PTR Records for Domain Overrides.

**Solution:**
Add functionality to disable PTR Records for individual Overrides or Aliases.
Defaults to yes. So existing Overrides don’t get broken.


AdSchellevis:
It's probably better to add a type (A or AAAA without PTR), since MX records don't receive PTR records, the option is confusing in that case.

AdSchellevis:
timeout

PR #3943, opened 2020-02-18
Unbound: add DNS64 support

### LINKED ISSUE #3933: Feature: DNS64 (unbound)

maurice-w:
Closes #3933.


maurice-w:
The use case are IPv6-only LANs. These are easier to configure and maintain than Dual-stack LANs, but require DNS64 + NAT64 for accessing IPv4-only services. NAT64 can be provided by the Tayga plugin.

fichtner:
Thanks, nice work. Let's try this then. :)

AdSchellevis:
@maurice-w would you be so kind to add documentation as well? https://docs.opnsense.org/manual/unbound.html  (https://github.com/opnsense/docs/blob/master/source/manual/unbound.rst)

PR #3946, opened 2020-02-20
DHCPv6 static mappings, fix DNS registration

### LINKED ISSUE #3945: 20.1 changed dhcpv6 staticmapping value domain into domainsearchlist --> auto-DNS-registration fails
[X] I have read the contributing guide lines at https://github.com/opnsense/core/blob/master/CONTRIBUTING.md

[X] I have searched the existing issues and I'm convinced that mine is new.

**Describe the bug**
In previous versions before 20.1 you could add static dhcpv6 leases and activate the option 'Register DHCP static mappings' in unbound. No need to add an extra DNS record.

In 20.1 the dhcpv6 static-leases changed the domain entry to domain-searchlist. Thus no DNS entries are created. Entries created with 19.X and migrated by upgrade to 20.1. still have the correct domain value in configuration and get registered.

ATM I manually have to download configuration, add the <domain> XML entry and restore DHCP configuration. 

Previously:
```
<staticmap>
        <duid>00:01:23:45:67:89:01:12:34:56:78:90:aa:bb</duid>
        <ipaddrv6>2002:db::1</ipaddrv6>
        <hostname>www</hostname>
        <descr>Webserver</descr>
        <domain>dmz.domain.tld</domain>
/staticmap>
```
created an AAAA record with www.dmz.domain.tld IN AAAA 2002:db::1

In 20.1:

```
 <staticmap>
        <duid>00:01:23:45:67:89:01:12:34:56:78:90:aa:bb</duid>
        <ipaddrv6>2002:db::1</ipaddrv6>
        <hostname>www</hostname>
        <descr>Webserver</descr>
        <domainsearchlist>dmz.domain.tld;domain.tld</domainsearchlist>
</staticmap>  
```

**To Reproduce**
Create static dhcpv6 lease and activate option to register DHCP leases in unbound.

**Expected behavior**
Possibility to add a domain in static lease configuration and get correct DNS records auto-created by unbound.

**Additional context**
If really needs to set an extra domainsearch-list per host, then at least reactivate the domain option in lease configuration.

**Environment**
OPNsense 20.1.1-amd64
FreeBSD 11.2-RELEASE-p16-HBSD
OpenSSL 1.1.1d 10 Sep 2019

maurice-w:
'domain' was removed in #3824 because it is not used by dhcpdv6. But it is used by unbound and dnsmasq for DNS registration of DHCP static mappings. Just set it to the first entry of the domain search list.
Closes #3945.

### DIFF
--- a/src/etc/inc/plugins.inc.d/dnsmasq.inc
+++ b/src/etc/inc/plugins.inc.d/dnsmasq.inc
@@ -265,8 +265,20 @@ function _dnsmasq_add_host_entries()
                     }
 
                     $domain = $config['system']['domain'];
-                    if ($host['domain']) {
+                    // set domain to first entry of host's domain search list
+                    if ($host['domainsearchlist']) {
+                        $domain_array=preg_split("/[ ;]+/",$host['domainsearchlist']);
+                        $domain = $domain_array[0];
+                    // backward compatibilty: 'domain' was replaced by 'domainsearchlist'
+                    // in dhcpv6 static mappings, but may still exist in older configs
+                    } elseif ($host['domain']) {
                         $domain = $host['domain'];
+                    // set domain to first entry of interface's domain search list
+                    } elseif ($dhcpifconf['domainsearchlist']) {
+                        $domain_array=preg_split("/[ ;]+/",$dhcpifconf['domainsearchlist']);
+                        $domain = $domain_array[0];
+                    // backward compatibilty: 'domain' was removed from dhcpv6
+                    // interface settings, but may still exist in older configs
                     } elseif ($dhcpifconf['domain']) {
                         $domain = $dhcpifconf['domain'];
                     }
--- a/src/etc/inc/plugins.inc.d/unbound.inc
+++ b/src/etc/inc/plugins.inc.d/unbound.inc
@@ -680,8 +680,20 @@ function unbound_add_host_entries()
                     }
 
                     $domain = $config['system']['domain'];
-                    if ($host['domain']) {
+                    // set domain to first entry of host's domain search list
+                    if ($host['domainsearchlist']) {
+                        $domain_array=preg_split("/[ ;]+/",$host['domainsearchlist']);
+                        $domain = $domain_array[0];
+                    // backward compatibilty: 'domain' was replaced by 'domainsearchlist'
+                    // in dhcpv6 static mappings, but may still exist in older configs
+                    } elseif ($host['domain']) {
                         $domain = $host['domain'];
+                    // set domain to first entry of interface's domain search list
+                    } elseif ($dhcpifconf['domainsearchlist']) {
+                        $domain_array=preg_split("/[ ;]+/",$dhcpifconf['domainsearchlist']);
+                        $domain = $domain_array[0];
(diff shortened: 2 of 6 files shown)


AdSchellevis:
Hmm, I think I misunderstood your comment in the issue.  I expected you wanted to re-add the domain as a separate field and keep it as is. This is a bit confusing. if the option was used, the data is "forgotten" and it needs a lot of explaining now that only the first item is used. 

I'm not exactly sure in which scenario you would have more entries in the search list and how that should relate to the dns registrations. Without that I'm a bit worried the next guy asks to add all entries in the search list (if that makes sense?).  If it should always add the first (or all) we better fix unbound (not sure about dnsmasq, need to check) to understand the searchlist property. 

What do you think?

maurice-w:
I considered re-adding a separate domain field. But I'm not aware of a use case where the domain for DNS registration would _not_ be the same as the first (or only) domain in the domain search list (DNSSL). So you would always have to enter the same domain twice.

If someone really wants multiple automatic DNS registrations for a single dynamic mapping... Well, that was never possible and would be a new feature. Out of scope.

About the "forgotten data" issue, I could add some code which fills the domain search list field in the UI with the `<domain>` from the saved config ONLY IF there is no `<domainsearchlist>` in the saved config.

dnsmasq and unbound do the same thing: The `<domain>` of the static mapping has the highest priority. If missing, the `<domain>` of the interface's DHCPv6 configuration is used. If this is missing, too, the system domain is used. Changing this behaviour is possible, but would be more work and I don't really see an advantage.

(Background: The DNSSL provided by DHCPv6 is used by hosts for DNS queries for non-FQDNs. Let's say the DNSSL contains 'example.com' and 'intranet.example.com'. A host wants to know the IP address of 'server1'. It will then perform DNS queries for 'server1.example.com' and 'server1.intranet.example.com'. The first entry will always be considered the system's primary domain, which is why it makes sense to use that for DNS registration, too.)

AdSchellevis:
ok, if the first is the one to use, it's probably better to load `domain` into `domainsearchlist` when the latter isn't set, ditch saving `domain` and fetch the first entry of `domainsearchlist` when set in unbound (and possibly dnsmasq). It doesn't make sense to save the same data twice for the same static entry.

The help text can probably change as suggested.

maurice-w:
> It doesn't make sense to save the same data twice for the same static entry.

Valid point. But requires a lot of changes. I'll try.

Just to be clear, what you suggest is:
* In `services_dhcpv6.php` and `services_dhcpv6_edit.php`, load `domain` into `domainsearchlist` if the latter isn't set and ditch saving `domain`.
* In `unbound.inc` and `dnsmasq.inc`, fetch the first entry of `domainsearchlist` if set, otherwise fetch `domain`.

Oh boy, what have I gotten myself into. ;-)

AdSchellevis:
check, that should be it indeed.

maurice-w:
Okay, seems to work. Please double-check.

AdSchellevis:
@maurice-w thanks, I've simplified the logic a bit in https://github.com/opnsense/core/commit/ea15651f4b5ea7b7dd462a43e23dc74d214fae09

Issue #3852, opened 2019-12-21
IPsec: dual stack virtual address pool for mobile clients


vnxme:
**Important notices**
Before you add a new report, we ask you kindly to acknowledge the following:

[x] I have read the contributing guide lines at https://github.com/opnsense/core/blob/master/CONTRIBUTING.md

[x] I have searched the existing issues and I'm convinced that mine is new.

**Is your feature request related to a problem? Please describe.**
IPsec mobile clients should be able to obtain both IPv4 and IPv6 from the server. These addresses are allocated from virtual address pools specified in **VPN ‣ IPsec ‣ Mobile Clients**. There is currently no way to define IPv6 virtual address pool, as only one IPv4 virtual address pool can be inserted and enabled. However, the backend (strongswan) is confirmed to support the suggested configuration.

**Describe the solution you'd like**
Frontend:
I suggest another block for IPv6 virtual address pool settings should be added to the Client Configuration (mode-cfg) section below the IPv4 virtual address pool settings. In my opinion, the page should have a separate checkbox for IPv6 (_Provide a virtual IPv6 address to clients_) so that mobile clients could be IPv4-only (the existing configuration block), IPv6-only (the configuration block described above) or dual stack (two blocks in question).

Backend:
File _/usr/local/etc/ipsec.conf_ should have `rightsourceip=%config4,%config6`, while it currently provides for something like `rightsourceip=%config4`. E.g. if one enters _192.168.254.0/24_ and _fd00:1:2:3::/64_ as virtual address pools via the frontend, it should be rendered as `rightsourceip=192.168.254.0/24,fd00:1:2:3::/64` in _/usr/local/etc/ipsec.conf_.

**Describe alternatives you've considered**
Before the above mentioned feature is implemented, it is possible to manually adjust _/usr/local/etc/ipsec.conf_ and restart strongswan. However, I have not found a solution to keep such manual changes from being overriden.

**Additional context**
Software Version: OPNsense 19.7.8
IPsec: I tested only IKEv2, as IKEv1 allegedly has some limitations regarding IPv6, but I have not investigated it much.
Wiki: IPv6 is not mentioned in Step 1 - Mobile Clients section of the [relevant wiki page](https://wiki.opnsense.org/manual/how-tos/ipsec-road.html).

vnxme:
Bump. At least 18 new issues have been added and processed by the OPNsense team so far, but this particular issue is ignored (uncommented, untagged, unassigned). Could you explain the reason?

[...]

PR #3949, opened 2020-02-21
IPsec: add virtual IPv6 pool for mobile clients

vnxme:
The feature is described at https://github.com/opnsense/core/issues/3852.

Known OPNsense limitations:
- multiple IKEv2 mobile servers use the same IPv4 and IPv6 pools - ipsec.conf file lets us set individual pools to each mobile server independently, but the current configuration model adopted in OPNsense web UI does not provide for matching tunnel settings to mobile clients profiles - it might be a feature;
- automatic outbound NAT rules (in [filter.lib.inc](https://github.com/opnsense/core/blob/master/src/etc/inc/filter.lib.inc#L196-L199)) include only IPv4 pool, since IPv6 NAT is discouraged - a checkbox can be introduced somewhere to let the admin decide whether to create a NAT rule for IPv6 pool or not - it might also be a feature.

Known protocol and/or client limitations (checked with a latest iOS client):
- IKEv1 does not seem to support dual stack mobile clients when connected via IPv4;
- IKEv1 does not seem to support connections via IPv6 at all.

### DIFF
--- a/src/etc/inc/plugins.inc.d/ipsec.inc
+++ b/src/etc/inc/plugins.inc.d/ipsec.inc
@@ -1349,8 +1349,15 @@ function ipsec_configure_do($verbose = false, $interface = '')
                 }
 
                 $rightsourceip = null;
-                if (!empty($a_client['pool_address']) && isset($ph1ent['mobile'])) {
-                    $rightsourceip = "\trightsourceip = {$a_client['pool_address']}/{$a_client['pool_netbits']}\n";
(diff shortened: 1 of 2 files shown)


AdSchellevis:
@vnxme thanks, within the current legacy module we indeed have some limitations to take into account. This looks like a reasonable addition, within the possibilities we have now. Eventually it would be good to rewrite the IPSec module, but I don't expect that will happen anytime soon (too expensive, migration will be problematic  and some other technologies are gaining traction as well, such as wireguard.)

We should indeed leave the automatic rules as is, if people want to create such rules, they can do so manually.

PR #3955, opened 2020-02-28
Fix deprecated Netscape extensions in OpenSSL config

vnxme:
This PR fixes text inconsistencies in nsComment and disables nsComment and nsCertType extensions as they are deprecated. See [OpenSSL manual](https://www.openssl.org/docs/man1.0.2/man5/x509v3_config.html) for details.

I divided this PR into two independent commits so that you could accept only the first one in case you are against disabling deprecated Netscape extensions for any reason. However, as the use of such extensions is officially discouraged and major authorities (e.g. DigiCert, Let's Encrypt) do not include them into their certificates, it seems reasonable to follow the recommendations. Those who need these extensions may always enable them manually in OpenSSL configuration file.

The tables below illustrate nsComment and nsCertType values depending on certificate type.

###### Before the proposed inconsistency fix:
| Certificate Type | nsCertType | nsComment |
| :---------------- | :----------: | :-------------- |
| CA | none | none |
| CRL | none | OpenSSL Generated Certificate |
| CSR | none | OPNsense Generated Certificate |
| Client | none | OpenSSL Generated Certificate |
| Server | server | OpenSSL Generated Server Certificate |
| Combined | none | OPNsense Generated Combined Client/Server Certificate |

###### After the proposed inconsistency fix, before disabling deprecated extensions:
| Certificate Type | nsCertType | nsComment |
| :---------------- | :----------: | :-------------- |
| CA | none | OPNsense Generated Certificate Authority |
| CRL | none | OPNsense Generated Certificate Revocation List |
| CSR | none | OPNsense Generated Certificate Signing Request |
| Client | none | OPNsense Generated Client Certificate |
| Server | server | OPNsense Generated Server Certificate |
| Combined | none | OPNsense Generated Combined Client/Server Certificate |

###### After disabling deprecated extensions:
| Certificate Type | nsCertType | nsComment |
| :---------------- | :----------: | :-------------: |
| CA | none | none |
| CRL | none | none |
| CSR | none | none |
| Client | none | none |
| Server | none | none |
| Combined | none | none |

### DIFF
--- a/src/etc/ssl/opnsense.cnf
+++ b/src/etc/ssl/opnsense.cnf
@@ -174,7 +174,7 @@ basicConstraints=CA:FALSE
 keyUsage = nonRepudiation, digitalSignature, keyEncipherment
 
 # This will be displayed in Netscape's comment listbox.
-#nsComment			= "OPNsense Generated Client Certificate"
+nsComment			= "OPNsense Generated Client Certificate"
 
 # PKIX recommendations harmless if included in all certificates.
 subjectKeyIdentifier=hash
@@ -198,7 +198,7 @@ keyUsage = nonRepudiation, digitalSignature, keyEncipherment
 # Extensions for a typical CA
 
 # This will be displayed in Netscape's comment listbox.
-#nsComment			= "OPNsense Generated Certificate Authority"
+nsComment			= "OPNsense Generated Certificate Authority"
 
 # PKIX recommendation.
 
@@ -231,7 +231,7 @@ authorityKeyIdentifier=keyid:always
 basicConstraints=CA:FALSE
 
 # This will be displayed in Netscape's comment listbox.
-#nsComment			= "OPNsense Generated Certificate Revocation List"
+nsComment			= "OPNsense Generated Certificate Revocation List"
 
 # PKIX recommendations harmless if included in all certificates.
 subjectKeyIdentifier=hash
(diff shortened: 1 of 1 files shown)


fichtner:
I'm generally ok with the comment changes, but not with commenting them out altogether. They do no harm and provide compatibility with older pieces of equipment / software.

vnxme:
> I'm generally ok with the comment changes, but not with commenting them out altogether. They do no harm and provide compatibility with older pieces of equipment / software.

OK, I see your point. Unless anyone else is for commenting out the deprected extensions, we have only the first commit to accept.

By the way, it is mentioned [here (p. 14, ref. 45)](https://www.ogf.org/documents/GFD.125.pdf), that nsComment should be replaced with certificatePolicies.userNotice.explicitText. I have conducted a few experiments with this extension and found no solution to make it look good in a certificate without specifying OIDs, CPS, etc. If anyone wants to try, please follow a guide in the [official OpenSSL manual, section Certificate Policies](https://www.openssl.org/docs/man1.0.2/man5/x509v3_config.html).

src/etc/ssl/opnsense.cnf
```diff
@@ -229,7 +231,7 @@ authorityKeyIdentifier=keyid:always
 basicConstraints=CA:FALSE
 
 # This will be displayed in Netscape's comment listbox.
-nsComment			= "OpenSSL Generated Certificate"
+nsComment			= "OPNsense Generated Certificate Revocation List"
```
fichtner: this is under [ proxy_cert_ext ], not [ crl_ext ] -- is this correct?

vnxme: You are absolutely right! I didn't notice [ proxy_cert_ext ] at all. Fixed it in the next commit.

fichtner:
Merged, thanks! ❤️

PR #3963, opened 2020-03-03
DynDNS: get_dyndns_ip() Chinese alternative capture regex

NOYB:
Alternative host also requires an alternative capture regex.


AdSchellevis:
currently seems to do the same as it already did, closing this for now.

NOYB:
Um no.  It doesn't do the same as before.  It fixes what was before.  Changing only the host does not work because the current capture regex does not match.  The capture regex also need to be changed.

PR #3971, opened 2020-03-10
DynDNS: Check IP Services - plugin support

NOYB:
Use the Check IP Services plugin if it is installed.

No functional change to the core default.
 * Optional function variable (service name, default null).
 * Capture regex as variable.
 * Only execute if the Check IP plugin is installed.

For Check IP Services - Plugin (opnsense/plugins#1731)

### DIFF
--- a/src/etc/inc/util.inc
+++ b/src/etc/inc/util.inc
@@ -1493,7 +1493,7 @@ function dhcp6c_duid_clear()
     @unlink('/conf/dhcp6c_duid');
 }
 
-function get_dyndns_ip($int, $ipver = 4)
+function get_dyndns_ip($int, $ipver = 4, $checkip = null)
 {
     $ip_address = $ipver == 6 ? get_interface_ipv6($int) : get_interface_ip($int);
     if (empty($ip_address)) {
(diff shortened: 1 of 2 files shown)


AdSchellevis:
It's probably easier to write a new dyndns plugin, `get_dyndns_ip()` is a remnant of two old plugins which aren't very well supported anymore. In core we don't accept dependancies on other plugins, which defuses separation of concerns.

NOYB:
Designed to work with the two plugins, DynDNS and RFC2136, as they currently exist.  To also be completely ignored if the CheckIP plugin is not installed.  Leaving the behavior unchanged.

Being that those two plugins are the only thing that use the get_dyndns_ip function and nothing in core uses it this is pretty insignificant.  None at all for those who do not use the CheckIP plugin.

Creating a new DynDNS plugin that also takes into account RFC2136 and CheckIP would be more work than it may seem.

Would it suffice to separate out the get_dyndns_ip function from etc/inc/util.inc into a  plugins.inc.d include?

AdSchellevis:
I'm afraid not, eventually it should be moved out of core. The code quality of both legacy plugins is quite poor, which isn't a huge issue if it more or less works and no one wants to work on a new and better structured version.
Hard (optional) dependencies in core are definitely a no go area, if hooks are missing, we should consider adding them, in this case the function shouldn't be in core at all.

In theory this doesn't seem like the most complicated software to build, but as core team we don't intend working on this at the moment. 
There might even be other external tools which are being maintained and could be implemented in OPNsense, the looser coupled the better.

NOYB:
Update: Removed in core dependency on other plugins.

AdSchellevis:
as explained in https://github.com/opnsense/core/pull/3971#issuecomment-599967614 , currently we're not planning to pull this in.

PR #3982, opened 2020-03-16
Possibility to add and remove nested alias from and existing alias with the alias_util API

cdeconinck:
Actually the alias_util API, allow only to add address to an existing alias
with this commit, it's possible now to add an existing alias (type = hosts or networks) atomically to another alias
This can be very useful in an environment where many concurrent requests occured because it avoids the logic of retrieving the content of the alias and push the modification with the API (non atomic) where simply adding or deleting a nested alias can be done in one atomic request

### DIFF
--- a/src/opnsense/mvc/app/controllers/OPNsense/Firewall/Api/AliasUtilController.php
+++ b/src/opnsense/mvc/app/controllers/OPNsense/Firewall/Api/AliasUtilController.php
@@ -209,6 +209,38 @@ class AliasUtilController extends ApiControllerBase
             $backend = new Backend();
             $backend->configdpRun("filter delete table", array($alias, $address));
             return array("status" => "done");
+        } elseif ($this->request->isPost() && $this->request->hasPost("alias")) {
+            Config::getInstance()->lock();
+            $aliasToRemove = $this->request->getPost("alias");
+            $cnfAlias = $this->getAlias($alias);
+            if ($cnfAlias == null || !in_array($cnfAlias->type, array('host', 'network'))) {
+                return array("status" => "failed", "status_msg" => sprintf("non existing alias %s", $alias));
+            }
+            // check if the nested alias exist
+            $nestedAlias = $this->getAlias($aliasToRemove);
+            if ($nestedAlias == null || !in_array($nestedAlias->type, array('host', 'network'))) {
+                return array("status" => "failed", "status_msg" => sprintf("non existing alias %s", $aliasToRemove));
+            }
+            $items = !empty((string)$cnfAlias->content) ? explode("\n", $cnfAlias->content) : array();
+            $index = array_search($aliasToRemove, $items);
+            if ($index === false) {
+                return array("status" => "failed", "status_msg" => sprintf("Alias '%s' it not present in %s : %s", $aliasToRemove, $alias, $index));
+            }
+
+            unset($items[$index]);
+            $cnfAlias->content = implode("\n", $items);
+            $this->getModel()->serializeToConfig();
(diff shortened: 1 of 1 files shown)


AdSchellevis:
It would be a good idea to explain a bit what the feature should do, this looks quite complicated, I'm not yet convinced that there isn't an easier solution here. (model content should be validated by the model usually, there might be edge cases, but only with very good reasons)

cdeconinck:
Hello, i've squash my commit and add ad description

For the model part, do you want to check the new content before saving ?
Si vous voulez je peux aussi modifier l'interface web pour prendre en compte cette nouvelle fonctionalité, sachant que ce PR ne casse aucune fonction  éxistant

AdSchellevis:
I'm having difficulties getting my head around this, is the intend here to support aliases as add/remove parameter? If that's the case, I don't think we should do that now. Since aliases change, which makes it quite difficult to guarantee consistency.

cdeconinck:
Yes, the idea is to allow add and remove nested alias from an existing alias with the same format used for the addresses

for exemple if we want to add the alias TEST in the content of the alias MASTER
curl   --header "Content-Type: application/json"   --basic   --user "secret:key"   --request POST   --insecure   --verbose   --data  '{"alias":"TEST4"}'   http://myhost /api/firewall/alias_util/add/MASTER

and if we want to delete it 
curl   --header "Content-Type: application/json"   --basic   --user "secret:key"   --request POST   --insecure   --verbose   --data  '{"alias":"TEST4"}'   http://myhost /api/firewall/alias_util/delete/MASTER

I've done this PR because i'm working in parallele on a terraform provider who will extensively use these feature for provisionig our infrastructure
Actually, for nesting an alias to another alias, we retrieve the content of the alias that we want to modify and we add the desired alias to it then we push an update. This technique works, but if several users want to modify the same alias, then we can end up with competition problems. it seemed to us more logical to implement this functionality rather than to manage the loc of the resources in our tools

AdSchellevis:
You could consider using the normal api endpoints at `/api/firewall/alias`, in which case you would need  some additional calls, but the result should be more less the same. The issue with the current approach is that it won't work for all types of aliases (geoip, url lists, ..).

Adding this functionality to the util endpoint would either lead to an implementation with limitations (difficult for people to understand) or one which can consume quite some time to apply.
You could also build a plugin specifically for your own use-case (and host it yourself), in which case you don't have to worry about certain constraints.

cdeconinck:
I agree that the current PR limit the addition / deletion of aliases on hosts and network type aliases, but it would be quite simple to extend the functionality in the futur to other types of aliases
Besides, currently alias_util can only be used on hosts and network aliases for adding and removing data, so I don't think adding this functionality to the same type of alias can cause a misunderstanding
Our need is just to be able to pass alias instead of address in the alias_util endpoint and work likewise

Regarding the use of the endpoint /api/firewall/alias, this is the one we are currently using. By design, to modify an alias, we first retrieved the content and then update it. So by definition overwriting problems can occur if several requests try to modify the content of the same alias at the same time

And for the performance part, I used the same logic used for adding addresses in the current implémentation, except that this time, instead of adding the address directly in the filter, i retrived the addresses of the alias that I want to add and I loop to add them in the filter

AdSchellevis:
I'm sorry, I gently have to decline this PR for maintenance concerns. The util page and endpoint really isn't intended to add this weight, historically you could always add/drop entries from an existing alias, which is a rather quick operation. These use-cases require full control, which is better served via the standard endpoints as mentioned before (update -> apply also guarantees consistency).

PR #3983, opened 2020-03-16
Check for VPN Interfaces

QDaniel:
Wireguard and TINC Interfaces was defunct

### DIFF
--- a/src/opnsense/mvc/app/library/OPNsense/Routing/Gateways.php
+++ b/src/opnsense/mvc/app/library/OPNsense/Routing/Gateways.php
@@ -84,13 +84,13 @@ class Gateways
     {
         if (!empty($ifcfg['if'])) {
             if ($ipproto == "inet") {
-                if (substr($ifcfg['if'], 0, 5) == "ovpnc") {
+                if (self::IsVPNInterface($ifcfg)) {
                     return "VPNv4";
                 } elseif (in_array(substr($ifcfg['if'], 0, 3), array('gif', 'gre'))) {
                     return "TUNNELv4";
                 }
             } elseif ($ipproto == "inet6" && !empty($ifcfg['if'])) {
-                if (substr($ifcfg['if'], 0, 5) == "ovpnc") {
+                if (self::IsVPNInterface($ifcfg)) {
                     return 'VPNv6';
                 } elseif (in_array(substr($ifcfg['if'], 0, 3), array('gif', 'gre'))) {
                     return 'TUNNELv6';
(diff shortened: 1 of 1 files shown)


AdSchellevis:
The OpenVPN hook is just a leftover from the legacy system, which we rather don't extend unless there's a generic change which functions for all same type interfaces. 

Do you experience issues when you add a gateway manually? As far as I understood so far, you should be able to add manual gateways for these types.

QDaniel:
With the commit 93bbe1ee168bbe0297746c2a2f68a5a4a4cd0972

The dynamic wireguard and tinc Gateways are broken.

AdSchellevis:
I understand that both commits lead to the same result, but I'm not convinced that we need to re-add something similar now (there might be side affects).

What does your setup look like? Is the gateway address unknown?

QDaniel:
I don't have find a model dir the Gateways to add a setting to Override the defunct check

AdSchellevis:
there isn't. Maybe if we go one step back, we could determine what's needed here. Hence the question about your setup.

QDaniel:
Yes, tinc and WG  are dynamic Interfaces without a Gateway ip

AdSchellevis:
In most situations you know one end of the tunnel, which is likely what most people use. In some cases the address to use is collected dynamically. Sending traffic to an interface without a destination usually isn't needed.

You could always ask @mimugmail what he configures for Wireguard.

mimugmail:
https://forum.opnsense.org/index.php?topic=15105.msg71616#msg71616

AdSchellevis:
I don't mind if we do need to reevaluate the choice only support valid addresses in gateways, but I really need more context about when and why we would like to allow this. An extra checkbox might also be an option in this case, or should we do so for all tunnel types.

QDaniel:
> https://forum.opnsense.org/index.php?topic=15105.msg71616#msg71616

I will check it out and test it. Then we need this only for TINC.
Yes a checkbox for no defunct check where the best.


Yes, we have  4 WG VPNs and 1 TINC .
All wg interfaces are allowed 0.0.0.0/0 so we must set routing manual.
With Static Routes and dynamic Gateways.
TINC is in bridge mode , the routes must be only set to the interface.

AdSchellevis:
@QDaniel @mimugmail maybe we better add an option on the interface, like  https://github.com/opnsense/core/commit/dba70c0ead9b55e4be8ac37dad5c2afbfe3209e8, so we can support routes without intermediate gateways, but don't try to hardcode every tunnel type in the world.

src/opnsense/mvc/app/library/OPNsense/Routing/Gateways.php
```diff
@@ -267,9 +281,11 @@ public function getGateways()
...
                 foreach ($intfgws as $gw_arr) {
                     if (!empty($gw_arr)) {
-                        $gw_arr['disabled'] = true;
-                        $gw_arr['defunct'] = true;
-                        unset($gw_arr['gateway']);
+                        if(!self::IsVPNInterface($gw_arr)){
+                            $gw_arr['disabled'] = true;
+                            $gw_arr['defunct'] = true;
+                            unset($gw_arr['gateway']);
+                        }
```
drivera73: In some use cases you may want a VPN interface to be your default gateway, so the IsVPNInterface() check here should probably be removed to maintain functional consistency.

AdSchellevis: @drivera73 this PR won't be included, we're currently opting for this  https://github.com/opnsense/core/commit/dba70c0ead9b55e4be8ac37dad5c2afbfe3209e8

QDaniel:
>  [dba70c0](https://github.com/opnsense/core/commit/dba70c0ead9b55e4be8ac37dad5c2afbfe3209e8)

This is good!

PR #3989, opened 2020-03-17
Gateway Editor Fix

drivera73:
The gateway editor makes an erroneous assumption which results in a code error when the interval isn't set (i.e. allowed to default), but the time period is.

This fix addresses that.

Judging by how the rest of the code is written, it might be a good idea to do give it a once-over so that there's proper input defense is applied.


src/etc/inc/plugins.inc.d/openvpn/ovpn-linkup
```diff
@@ -3,27 +3,33 @@
...
+/usr/bin/touch "/tmp/${1}upv6"
 
-/usr/local/opnsense/service/configd_ctl.py interface newip $1
+/usr/local/opnsense/service/configd_ctl.py interface newip "${1}"
+
+# If there's a setting for "additional --up script", then run that, but
+# within a subshell so we don't mess up the containing environment,
+# and we don't cause problems for the other stuff
 
 exit 0
```
drivera73: This is a stowaway due to poor branch management on my part... it's still useful, but not essential to the spirit of this change request. Cherry pick at your discretion. :)

AdSchellevis:
@drivera73 can you try https://github.com/opnsense/core/commit/0ec7e1f804537c39931848c1522a1c5902df3570 ?

drivera73:
Patch 0ec7e1f fixes the issue! When will it be released? 20.1.3? 20.7?

AdSchellevis:
@drivera73 thanks for confirming, probably in a 20.1.x version, maybe 20.1.5

PR #4019, opened 2020-04-05
PortField: Reduced per-instance ram usage by sharing port lists.

jkellerer:
*Issue*:
While profiling "`ngx_autoblock.php`", I've found that `PortField` instances consume a very large amount of memory. My `Nginx` model contains 72 of them, leading to a RAM consumption of ~300mb for just the `PortField` instances.

Note: I've checked the issue tracker and didn't find any existing entries addressing the problem.

*Reason*:
`PortField` instances use internal port list arrays containing all valid port values. An array in PHP is implemented as LinkedHashMap with lots of overhead per entry, thus this sums up to 4mb per port list, multiplied by the amount of `PortField` instances that a model contains.

*Solution*:
Using static port list arrays that are shared via "copy-on-write" (using normal assignments).
This ensures that existing behaviour isn't changed at all as port lists can still be modified independently per instance (leading to a copy-on-write). All unit tests in `PortFieldTest` still pass after this modification.

(Of course, best solution for current design would be to use array-like objects for numeric-ports but this requires changes to all upstream consumers of `internalOptionList`, thus it is more complex and dangerous)

*Profiling results:*
profiler-before-after

### DIFF
--- a/src/opnsense/mvc/app/models/OPNsense/Base/FieldTypes/PortField.php
+++ b/src/opnsense/mvc/app/models/OPNsense/Base/FieldTypes/PortField.php
@@ -102,14 +102,40 @@ class PortField extends BaseListField
      */
     protected function actionPostLoadingEvent()
     {
-        if ($this->enableWellKown) {
-            foreach (array("any") + self::$wellknownservices as $wellknown) {
-                $this->internalOptionList[(string)$wellknown] = $wellknown;
-            }
-        }
+        // Using static wellknown/numeric port lists (~4mb each) to reduce memory usage per instance.
+        // Arrays are copy-on-write, only instances that modify internalOptionList will get a real copy.
+        static $numericPorts = [];
+        static $wellknownPorts = [];
 
+        if ($this->enableWellKown) {
+            // Building shared list for numeric & wellknown ports
+            if (empty($wellknownPorts)) {
+                foreach (["any"] + self::$wellknownservices as $wellknown) {
+                    $p = strval($wellknown);
+                    $wellknownPorts[$p] = $p;
+                }
+
+                if (empty($numericPorts))
+                    self::addNumericPorts($wellknownPorts);
+                else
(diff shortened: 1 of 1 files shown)


AdSchellevis:
For some of the field types we use static fields, but you always have to take into account which selected options there are, otherwise sets risk interfering. (for example two selectors, one accepts well-known and the other won't). 
Looking at the PR, I think it currently isn't safe yet, but I'll take a look, usually these things are easy to fix.

jkellerer:
Hi @AdSchellevis, the PR is safe since the static sets I introduced are copied, not referenced. Every `PortField` instance can independently change it without interference. The reason why it still saves the memory is that PHP creates a shallow copy (reference) at first as long as the set isn't changed.

AdSchellevis:
@jkellerer you're right, its safe indeed. can you try https://github.com/opnsense/core/commit/fe335024fbcf6685d24072a95b3c938858bc7d60 as well, it's faster than your option, uses a tiny bit more memory, but is more aligned with how other field types handle caching. Jus for fun, try this on both:

```
$starttime = microtime(true);
for ($i =0 ; $i < 500 ; ++$i) {
    $field1 = new PortField();
    $field1->setEnableRanges("Y");
    $field1->eventPostLoading();
    $field1 = new PortField();
    $field1->setEnableRanges("N");
    $field1->eventPostLoading();
}

echo "runtime : " . (microtime(true) - $starttime);
```

jkellerer:
I've also considered this design, for the same reason that it is used in other field types, but I used another approach since I thought this is slightly easier to read. Nevertheless either solution is fine and solves the same problem.

Regarding performance, I've also found that "strval" is slightly slower than casting to string (when creating the PR and reviewing the profiler screenshots :)).

It's totally ok, if you prefer your solution instead of mine. My main concern was that lots of memory is being used every 30 seconds when "ngx_autoblock" runs.

PR #4024, opened 2020-04-09
The alias name must start with the symbol.

kekek2:


AdSchellevis:
needs a reference and a test case, e.g. this are valid aliases `_TEST` , `0TEST`, which seem to work fine on our current version.

NOYB:
Yes I have lots of aliases that begin with numbers.  Especially port number aliases.

PR #4041, opened 2020-04-16
Merge unbound plus

AdSchellevis:
Merge Unbound-plus functionality + cleanups.

All endpoints will change to `/ui/unbound/*` and `/api/unbound/*`, since it's unlikely a lot of people depend on these. Old crontabs should still work (configd call left in place temporary). 
The model stays in the "plus" location, to avoid migration issues and helps enforcing deinstallation of the plugin when installed.

@mimugmail can you check if this still works?


src/opnsense/scripts/unbound/dnsbl.py
```diff
@@ -0,0 +1,192 @@
+#!/usr/local/bin/python3
+
+# DNS BL script
+# Copyright 2020 Petr Kejval <petr.kejval6@gmail.com>
```
fichtner, line 4:
```suggestion
# Copyright (c) 2020 Petr Kejval <petr.kejval6@gmail.com>
```

PR #4058, opened 2020-04-24
avoid log spam #1

hippi-viking:
avoid syslog-ng dumping statistics every 10 minutes

### DIFF
--- a/src/opnsense/service/templates/OPNsense/Syslog/syslog-ng.conf.in
+++ b/src/opnsense/service/templates/OPNsense/Syslog/syslog-ng.conf.in
@@ -11,6 +11,7 @@ options {
     flush_lines(0);
     threaded(yes);
     create_dirs(yes);
+    stats-freq(0);
 };
 
 source s_all {


AdSchellevis:
maybe a stupid question, but does this only influence remote logging? I think I haven't seen those on my local machine yet (could very well be looking in the wrong place)

hippi-viking:
Not a stupid question at all! As I understand OPNSense uses syslogd for local logging and syslog-ng for remote logging (I don't know the reasoning behind, it should be simpler to use only one logging facility for both). I found this option in syslog-ng's online manual, which is specific to syslog-ng - thus you can only see it on the remote log, not the local one.
Would it make sense to switch entirely to syslog-ng (or back to syslogd) in the long run?

AdSchellevis:
circular logs, that's the reason for the extra layer, eventually this should be replaced but it's a longer term goal. I guess you're still using an older version of OPNsense, since 20.1 (if I'm not mistaken) you can filter messages to send to the remote target, which would remove the need to suppress these messages.

hippi-viking:
I have just updated to 20.1.5 from 20.1.4 today. Yes, I saw the filters, but most of them do not seem to do anything for me, or they work kind of inconsistently.

AdSchellevis:
The filters are just plain syslog-ng facilities, it all depends on the inbound log stream. If you have an example I can replicate over here, I can check what could be the issue.

hippi-viking:
Strange thing is I don't see my GUI settings mirrored in the syslog-ng config files - I suspect there could be some discontinuitiy at the config interface which might be the reason for not working quite well for me.
Sure, what should I send? Should I open a ticket about this not to spam this pull-request or send you my config files in a private message?

AdSchellevis:
better open an issue on Github and close the other one, but please, use our template and make sure someone can replicate this step by step.

AdSchellevis:
probably related to https://github.com/opnsense/core/issues/4060, I think we can leave the details as they are now.

Issue #4063, opened 2020-04-25
Feature Request: AdvLinkMTU setting on Router Advertisements Page


linuxgemini:
**Important notices**
Before you add a new report, we ask you kindly to acknowledge the following:

 - [x] I have read the contributing guide lines at https://github.com/opnsense/core/blob/master/CONTRIBUTING.md

 - [x] I have searched the existing issues and I'm convinced that mine is new.

**Is your feature request related to a problem? Please describe.**
Using GIF interfaces through a VPN tunnel that has lower MTU than all other interfaces in OPNsense  (like WireGuard) causes some SSL websites to not load (the usual MTU issue). This can be (usually) mitigated by lowering the advertised MTU on `radvd`. However on `dhcpd_radvd_configure`

https://github.com/opnsense/core/blob/2a216f91c79a469138e5d302b5eb03b3c3743d13/src/etc/inc/plugins.inc.d/dhcpd.inc#L250-L261

the MTU is derived from the `radvd` parent/target interface (and maybe from 6rd tracked interface that is mentioned on #2643 and #2553)

**Describe the solution you'd like**
Addition of AdvLinkMTU to

https://github.com/opnsense/core/blob/45b0d6834cf8cd217711a6119917f8b32d40c6dc/src/www/services_router_advertisements.php#L40

and re-calculating the MTU in `dhcpd.inc` based from the config.

**Describe alternatives you've considered**
I am currently lowering the LAN MTU, however this now causes instability issues on few IPv4 sites.

**Additional context**
The reason I'm doing this "tunnel in a tunnel" is due to me having a dynamic IPv4 address and my VPN mesh in WireGuard is not IPv6 oriented. Plus the server side of the tunnel does not have an update mechanism for IP updates.

linuxgemini:
#4064 is merged and modifeid on https://github.com/opnsense/core/commit/aef9ada7d13ba896de434d4f35268cbe015857f9

PR #4064, opened 2020-04-25
Add AdvLinkMTU to Router Advertisements page

linuxgemini:
As I mentioned in #4063, some edge cases may cause instability issues on IPv6 connectivity by relying to the target interface MTU for RA. This PR adds an override to the AdvLinkMTU.

### DIFF
--- a/src/etc/inc/plugins.inc.d/dhcpd.inc
+++ b/src/etc/inc/plugins.inc.d/dhcpd.inc
@@ -260,6 +260,13 @@ function dhcpd_radvd_configure($verbose = false, $blacklist = array())
             }
         }
 
+        if (!empty($dhcpv6ifconf['AdvLinkMTU'])) {
+            $overridemtu = $dhcpv6ifconf['AdvLinkMTU'];
+            if ($overridemtu < $mtu) {
+                $mtu = $overridemtu;
+            }
+        }
+
         $radvdconf .= "# Generated for DHCPv6 server $dhcpv6if\n";
         $radvdconf .= "interface {$realif} {\n";
         $radvdconf .= "\tAdvSendAdvert on;\n";
--- a/src/etc/inc/plugins.inc.d/dhcpd.inc
+++ b/src/etc/inc/plugins.inc.d/dhcpd.inc
@@ -264,6 +264,8 @@ function dhcpd_radvd_configure($verbose = false, $blacklist = array())
             $overridemtu = $dhcpv6ifconf['AdvLinkMTU'];
             if ($overridemtu < $mtu) {
                 $mtu = $overridemtu;
+            } else {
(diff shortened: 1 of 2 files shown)


AdSchellevis:
could it be that part of the change is missing from the PR? `AdvLinkMTU` doesn't seem to be populated anywhere.

linuxgemini:
> could it be that part of the change is missing from the PR? `AdvLinkMTU` doesn't seem to be populated anywhere.

On the configuration page, the `advanced_options` array is automatically filled to the "Advanced Options" section.
WinCap-2020-04-25-191003

On `dhcpd.inc`, this line is automatically appended; regardless of its source. (The quote below is from master)

https://github.com/opnsense/core/blob/2a216f91c79a469138e5d302b5eb03b3c3743d13/src/etc/inc/plugins.inc.d/dhcpd.inc#L271

linuxgemini:
One other thing to note: This PR does *not* touch the part for `DHCP-PD prefixes and 6RD dynamic interfaces`

AdSchellevis:
ah, sorry, my mistake, now I remember that part https://github.com/opnsense/core/commit/c0fe5ceff4dabced60c84cbec0d1d28110c9636e

src/etc/inc/plugins.inc.d/dhcpd.inc
```diff
@@ -260,6 +260,13 @@ function dhcpd_radvd_configure($verbose = false, $blacklist = array())
             }
         }
 
+        if (!empty($dhcpv6ifconf['AdvLinkMTU'])) {
+            $overridemtu = $dhcpv6ifconf['AdvLinkMTU'];
+            if ($overridemtu < $mtu) {
+                $mtu = $overridemtu;
```
fichtner, line 266: Nice work here. radvd.conf states: "If specified, i.e. not 0, must not be smaller than 1280 and not greater than the maximum MTU allowed for this link (e.g. ethernet has a maximum MTU of 1500. See RFC 4864)."

One small issue though is that we should be logging the case when the overridemtu cannot be used (else condition).

linuxgemini: Added an else condition

https://github.com/opnsense/core/blob/898d34afefdf94f4093490651350483bb5f68d70/src/etc/inc/plugins.inc.d/dhcpd.inc#L267-L269

AdSchellevis:
@linuxgemini  Thanks! I made a small change in https://github.com/opnsense/core/commit/aef9ada7d13ba896de434d4f35268cbe015857f9, functionally it should do the same.

linuxgemini:
Thank you! Now I can close #4063.

AdSchellevis:
@linuxgemini you did the work :) thank you!

PR #4070, opened 2020-04-28
ArrayField - Add sort_flags optional parameter support

NOYB:
e.g. SORT_NATURAL: 9, 20, 80 rather than 20, 80, 9
Default, SORT_REGULAR, is unchanged.

### DIFF
--- a/src/opnsense/mvc/app/controllers/OPNsense/Base/ApiMutableModelControllerBase.php
+++ b/src/opnsense/mvc/app/controllers/OPNsense/Base/ApiMutableModelControllerBase.php
@@ -305,7 +305,7 @@ abstract class ApiMutableModelControllerBase extends ApiControllerBase
      * @return array
      * @throws \ReflectionException when binding to the model class fails
      */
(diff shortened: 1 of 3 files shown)


AdSchellevis:
Adding natural sort is probably a good idea, although I'm not sure about the default if this means we're going to change (almost) all of our endpoints.

NOYB:
I don't follow the part not being sure about the default if means we're changing (almost) all the endpoints.  Could you break that down some for me.

 

Taking a guess at what that means.  We'll there are a lot of end points in the plugins.  Leaving the default as blank/null = SORT_REGULAR allows them to continue as-is if they like.

AdSchellevis:
https://github.com/opnsense/core/pull/4071 seems to replace quite some endpoints, so we need to inspect what's the most logical option when changing this. yes, backwards compatibility is preferable, but if there are only a couple of endpoints unchanged from our own system, we might consider changing defaults to keep the callers more logical.

I'll take a look

NOYB:
Thanks.  Natural sort default is preferable.

PR #4072, opened 2020-04-28
ArrayField - Don't sort unless a sort by field is provided

NOYB:
Avoids default sorting by the 'uuid' so that the default is unsorted.  i.e. config order

### DIFF
--- a/src/opnsense/mvc/app/models/OPNsense/Base/FieldTypes/ArrayField.php
+++ b/src/opnsense/mvc/app/models/OPNsense/Base/FieldTypes/ArrayField.php
@@ -169,11 +169,14 @@ class ArrayField extends BaseField
             $sortedData[$sortKey] = $node;
         }
 
-        // sort by key on ascending or descending order
-        if (!$descending) {
-            ksort($sortedData, $sort_flags);
-        } else {
-            krsort($sortedData, $sort_flags);
+        // proceed with sorting if at least one field name was provided
+        if ($fieldNames[0]) {
+            // sort by key on ascending or descending order
+            if (!$descending) {
+                ksort($sortedData, $sort_flags);
+            } else {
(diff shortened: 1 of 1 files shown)


AdSchellevis:
Which problem are you trying to solve? config order itself isn't safe / guaranteed either.

NOYB:
When adding to a list of whatever, you never know the position in the list where to find it due to sorting by the uuid.  The uuid's do not appear to be chronological.  Especially when combining config from another system.

So the default 'uuid' sorting seems kind of random.  From a human perspective doesn't seem useful.  Leaving it unsorted if no sort field is provided seems a little more deterministic from a chronological perspective.  Seems more useful to me.

 

Why isn't config order safe?  It's the same data, just not sorted.

 

As far as guaranteed.  Well 'uuid' order seem kind of random too anyway.

AdSchellevis:
it's is "random" in some sense, just as the ordering in the file would be, hence it doesn't really seem to make sense to change it. We really discourage using config ordering.

NOYB:
Uuid sort doesn't seem to provide any human usefulness.  While unsorted may not be a huge benefit, or guaranteed to strictly follow the config order. It does seem to at least to some degree.  Which seem to tend toward chronological.  Which is somewhat humanly useful.

Though the upside may not be huge, I'm not seeing any downside.  Net positive.

Would appreciate if the project could accommodate this.

Thanks for your consideration.

AdSchellevis:
I'll take a look, I don't mind not sorting (and save cpu cycles), just want to make sure no one assumes anything by the "natural" order of things.

NOYB:
Thanks for taking a second look at this.
I recall seeing only one core feature that doesn't currently provide a sort by field.  Traffic Shaper in 2 of it's 3 searchBase calls.
Don't know about the plugins.  Other than my own.

PR #4075, opened 2020-04-29
Add sort flag param comment to sortedBy  function header.

NOYB:


AdSchellevis:
@NOYB missed a spot is seems, thanks!

PR #4077, opened 2020-04-30
Nextcloud backup: allow subdir

### LINKED ISSUE #3988: Backup to Nextcloud
At the moment as far as I understand it, it is not possible to backup into a subdirectory of nextcloud. (Short discussion was here https://github.com/opnsense/core/issues/2250, but not really a solution). If I add a path to the Directory Name field I get the error 

> The Backup Directory can only consist of alphanumeric characters.

If I use e.g. `example.com/nextcloud/remote.php/dav/files/USERNAME/Backups`  as update URL i get the error

> Saved settings, but remote backup failed.

**Describe the solution you'd like**
It would be nice to either be able to use one of the two methods for putting the Update into a specific folder.

**Additional context**
I think it would be more convenient and would also support other webdav services if a general "add webdav backup" with the link method would be available.
Based on this one can also implement a short cut for nextcloud (and other webdav servers) where one only has to enter the base url and in the Directory field one can enter the path and in the background the strings will be concatenated according to the Nextcloud requirements.

maxxer:
Fixes #3988


AdSchellevis:
@fabianfrz do you have time to spin this for a test?

fabianfrz:
@AdSchellevis crashes immediately:

Fatal error: Uncaught Error: Call to undefined function OPNsense\Backup\basedir() in /usr/local/opnsense/mvc/app/library/OPNsense/Backup/Nextcloud.php:147 Stack trace: #0 /usr/local/www/diag_backup.php(286): OPNsense\Backup\Nextcloud->backup() #1 {main} thrown in /usr/local/opnsense/mvc/app/library/OPNsense/Backup/Nextcloud.php on line 147

fabianfrz:
log after fixing the issue by replacing it by basename(), opnsense/backup as file name:

config[60397]: Error while fetching filelist from Nextcloud

```json
{
  "url":"https:\/\/nextcloud.test\/remote.php\/dav\/files\/opnsensebkp\/Backup",
  "content_type":"text\/html; charset=UTF-8",
  "http_code":404,
  "header_size":1220,
  "request_size":205,
  "filetime":-1,
  "ssl_verify_result":0,
  ...
}
```

Tested on OPNsense 20.1.r_12 (amd64/LibreSSL) - I don't think somebody changed the nextcloud code inbetween

maxxer:
Hi @fabianfrz thanks for the feedback. 

About the last issue, one question: does `opnsensebkp` directory exist on your Nextcloud? Because indeed the plugin doesn't create all the path to the backup dir, I left the original behaviour that will create just the last one (`Backup` in your example). Is it expected differently?

maxxer:
I'm sorry for the misspelled function name, I meant to use `dirname` but accidentally copied the wrong file to git :(

fabianfrz:
now the php error is gone but the backup still fails.

maxxer:
Does the parent directory exist?

fabianfrz:
nope, but that is what the code is expected to do, create the full directory hierarchy if missing.

maxxer:
Ok, I will implement that. I asked two comments ago but got not answer...

fabianfrz:
Sorry I forgot the answer but if you read the source, there is already MKCOL implemented which is doing this so yes, it is the current behaviour.

maxxer:
I've updated the patch and merge against current master, I hope it's fine now

maxxer:
Any chance to get another review?

hasechris:
Hii,

> Any chance to get another review?
@fabianfrz could you please review this merge?

Thanks in advance

src/opnsense/mvc/app/models/OPNsense/Backup/NextcloudSettings.xml
```diff
@@ -48,9 +48,9 @@
         </password_encryption>
         <backupdir type="TextField">
             <Required>Y</Required>
-            <mask>/[a-z0-9\-]+/i</mask>
+            <mask>/^[\w\-][\w\-\/]+[\w\-]$/</mask>
```
fabianfrz, line 51: I think this regex is wrong. I think you meant something like this:
```suggestion
            <mask>/^([\w%+\-]+\/)*[\w+%\-]+$/</mask>
```
This means the first part is entirely optional (allows names like 'a') and if it is in place, It allows names like 'aa'.

You can test regexes on rubular if you want. Here are some examples;
```
abc
a/b/%20+/aa/xaajsdkjfallsdjfas
a/b/%20+/aa/xaajsdkjfallsdjfas%20
a/b/%20+/aa/xaajsdkjfallsdjfas+
a//b
```
The last one does not match so the double slash is detected as well! Important things: % is used for encoding using hex like %20 for space. Also + is used for spaces so this characters should be allowed as well.

src/opnsense/mvc/app/models/OPNsense/Backup/NextcloudSettings.xml
```diff
@@ -1,6 +1,6 @@
 <model>
     <mount>//system/backup/nextcloud</mount>
-    <version>1.0.0</version>
+    <version>1.0.1</version>
```
fabianfrz, line 3: this is not required

fabianfrz (changes requested):
I found some improvements for the models but if this changes are in the PR, I can give it a test run.

fabianfrz:
Tested now successfully:

Screenshot_20210102_103815
Screenshot_20210102_103916
Screenshot_20210102_103959

AdSchellevis:
@maxxer @fabianfrz thanks! I will merge it to master

PR #4083, opened 2020-05-04
Reporting: Health: Inverse option applied to data series by name

NOYB:
The inverses option is applied to data series on an odd/even basis (every other data series).
This is fine for things like packets and traffic where every other data series is in/out (odd/even).
But for quality and system the data series that are inverted make no sense.

Here the mod method of odd/even is replaced by data series names.
'loss', 'outpass', 'outblock', 'outpass6', 'outblock6', 'interrupt', 'processes'


AdSchellevis:
Although the odd/even might no be as useful for some ranges, its functionality is clearly explained in the docs https://docs.opnsense.org/manual/systemhealth.html?highlight=inverse#inverse

Not going to pull this in at the moment.

PR #4088, opened 2020-05-06
Dhcp6c multiwan

fichtner:
- [x] move global settings to interfaces: settings where DUID resides anyway
- [x] debug level support none, -d, -D #4096 
- [x] loop through existing configs and reload selectively
- [x] rtsold script rewrite is now one script for all

### DIFF
--- a/src/etc/inc/interfaces.inc
+++ b/src/etc/inc/interfaces.inc
@@ -1090,7 +1090,8 @@ function interface_bring_down($interface = "wan", $ifacecfg = false)
     switch ($ifcfg['ipaddrv6']) {
         case 'slaac':
         case 'dhcp6':
-            killbypid("/var/run/dhcp6c_{$realifv6}.pid", 'TERM', true);
+            interface_dhcpv6_prepare($interface, $ifcfg, true);
+            killbypid('/var/run/dhcp6c.pid', 'HUP');
             break;
         case 'track6':
             interface_track6_configure($interface, $ifcfg, false, true);
@@ -2600,8 +2601,7 @@ function interface_track6_configure($interface = 'lan', $lancfg, $linkupevent = 
             case 'slaac':
             case 'dhcp6':
                 interface_dhcpv6_prepare($lancfg['track6-interface'], $trackcfg);
-                $parentrealif = get_real_interface($lancfg['track6-interface'], 'inet6');
-                killbypid("/var/run/dhcp6c_{$parentrealif}.pid", 'HUP');
+                killbypid('/var/run/dhcp6c.pid', 'HUP');
                 break;
             default:
                 break;
@@ -2624,8 +2624,7 @@ function interface_track6_configure($interface = 'lan', $lancfg, $linkupevent = 
         case 'dhcp6':
             if ($linkupevent) {
                 interface_dhcpv6_prepare($lancfg['track6-interface'], $trackcfg);
-                $parentrealif = get_real_interface($lancfg['track6-interface'], 'inet6');
-                killbypid("/var/run/dhcp6c_{$parentrealif}.pid", 'HUP');
+                killbypid('/var/run/dhcp6c.pid', 'HUP');
             }
             break;
     }
@@ -2962,7 +2961,7 @@ function interface_dhcpv6_id($interface)
     return $id;
 }
 
-function interface_dhcpv6_prepare($interface = 'wan', $wancfg)
+function interface_dhcpv6_prepare($interface = 'wan', $wancfg, $linkdownevent = false)
 {
     if (!is_array($wancfg)) {
         return;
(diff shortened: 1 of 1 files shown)


marjohn56:
1095 cannot be a SIGHUP, in the event of an interface being added or removed you have to kill the existing process as the command line carries the interfaces, i.e. igb0, igb1 etc.

fichtner:
nothing can be SIGHUP anymore actually. it's a pity that we have to pass interfaces to dhcp6c at all on the command line.

marjohn56:
It can be sighup if it's a LAN interface that's being added or something on the conf. I wonder why it was done originally using a command line option. Should be possible to change it to using the config, but it could take some work due to the nasty way the conf is processed.

marjohn56:
OK, removed the command line WAN interfaces argument, it now takes the interfaces from the config file, obvious really...

fichtner:
According to manual an interface is mandatory for invoke, but I’m happy if it would be that clever.

src/etc/inc/interfaces.inc
```diff
@@ -2921,8 +2920,9 @@ function interface_dhcpv6_configure($interface = 'wan', $wancfg)
     /* Enable RFC6204w support for IPv6 Customer Edge (CE) router */
     set_single_sysctl("net.inet6.ip6.rfc6204w3", "1");
 
-    /* always kill rtsold in case of reconfigure */
+    /* always kill in case of reconfigure */
     killbypid("/var/run/rtsold_{$realifv6}.pid", 'TERM', true);
+    killbypid('/var/run/dhcp6c.pid', 'TERM', true);
```
fichtner, line 2925: @marjohn56 maybe rtsold_script.sh should always fully reload as long as we can't SIGHUP interface listening. having it removed from the command line is the lesser problem in general

marjohn56: Up to you. The option is there in the PR for dhcp6c if we decide to use it, if not can we just cherry pick the PDINFO bit, that'll be useful.

marjohn56:
pulled this an ran it, appears OK apart from the no-release and debug flags not working. As I said, it's up to you what you want to do with dhcp6c. I'm running with the no interfaces version and its working fine.

src/etc/inc/interfaces.inc
fichtner: rtsold/rtsold_script never comes from trackers, only from true WAN interfaces

fichtner:
@marjohn56 yeah we should rid ourselves of the interfaces on command line option. it has a number of advantages, first and foremost remove the gruesome complexity of core rtsold/dhcp6c scripting around it

src/etc/inc/interfaces.inc
```diff
@@ -1090,7 +1090,8 @@ function interface_bring_down($interface = "wan", $ifacecfg = false)
     switch ($ifcfg['ipaddrv6']) {
         case 'slaac':
         case 'dhcp6':
-            killbypid("/var/run/dhcp6c_{$realifv6}.pid", 'TERM', true);
+            interface_dhcpv6_prepare($interface, $ifcfg, true);
+            killbypid('/var/run/dhcp6c.pid', 'HUP');
```
fichtner, line 1094: HUP works as expected now, rereading the interfaces from the configuration as well as command line. In this proposal we remove the command line use as well.

marjohn56: I've just been driving myself mad trying to find out why the code would not start dhcp6c, it was because I had used a kill -QUIT command to shutdown dhcp6c and not a SIGTERM. QUIT behaves like -9 and leaves the PID behind preventing dhcp6c from starting. This is one of the reasons I added the checks and double checks to make sure this could not happen.. obviously I must have been thinking of myself! We really do need to make sure this cannot happen. If dhcp6c barfs for any reason then it will not restart until the pid is removed.

fichtner: I don't think we ever had problems again, but now that I know the issue it would make sense to prep for this case, i.e. stale pid file => remove it. we shouldn't assume there is another dhcp6c running. better yet, we can lock the start procedure so that only one script will ever be able to start it.

marjohn56: Hence my original PR that checked for a stale PID and also checked the PID matched the process. I know it may have appeared as a total overkill but this dates back to pf*** and the same issues there. 

Even with just one script to start it you still need to make it foolproof... and I am that fool, if it can be broken, I WILL break it. ;)

marjohn56: I've put back my rtsold script routine and it works well. Is there a reason we cannot use it?

fichtner: pgrep can return multiple PIDs, the check isn't so great. I'm pushing a different proposal for now:

check if the pid in the file is responding, if not delete pid file

marjohn56:
Want to take a look at 3063... Methinks if is fi .:)

fichtner:
indeed, thanks!

marjohn56:
array($sysfg['dhcp6_debug']))

dhcp6c_debug is now a value 0,1,2 in the config, I used a switch case to bring it back to the letter in my code. Do you want it back to letters in status_interfaces?

fichtner:
status_interfaces.php ? We don't need the info there I think

marjohn56:
Sorry, I meant system_advanced_network.. doh!

fichtner:
text representation is better... something like none, abbreviated, full ?

marjohn56:
OK, I'll do that..

marjohn56:
How do I commit to this, I cannot fork it and although I create a local clone I cannot push a commit.

fichtner:
# git checkout -b dhcp6c_multiwan_2

push to your repo, create PR against dhcp6c_multiwan branch

fichtner:
nevermind, we still have the interfaces in the rtosold script.... we can merge this now even though devel may not be 100% perfect yet.

fichtner:
@marjohn56 thanks a ton for your work!! ❤️

marjohn56:
Well my de-furlough has been postponed as the clients premises are still closed and will stay so until the UK relaxes the lock down.. So play on. ;)

PR #4091, opened 2020-05-07
Reporting_Health_Resolution_Description_-_Units_Singular

NOYB:


AdSchellevis:
In most cases the output isn't singular, I don't think we should change this.

PR #4096, opened 2020-05-08
Move settings for dhcp6c to system_advanced_network

marjohn56:
To go with #4088


fichtner:
Thanks!

marjohn56:
line 480 in dhcpd, I made the comment that it barfs radvd.. need to look at that.

fichtner:
I remember, bring on the last PR dear Sir.

marjohn56:
It would seem that's already been fixed, so it's some testing that now needs to be done. I suggest we post a forum message looking for users who want to test multiwan dhcp6.

marjohn56:
Ooerr... only one rtsold script... hmm, dat won't work. :)

fichtner:
Why not? Same config, same PID, same debug/norelease.

marjohn56:
Yup sorry... brainfart moment... I'm wondering why I'm not seeing multiple routes... I'll find it.

marjohn56:
interfaces - rtsold should be passed all the interfaces, otherwise it will only get the info from the one given.

fichtner:
Merged via 5ed15ca, thanks! ❤️

PR #4099, opened 2020-05-09
Show ISP Assigned PD in status nterfaces

marjohn56:
uses the new feature within dhcp6c to display the ISPs allocated prefix. We may be able to use this rather than asking the user to enter the prefix size in interfaces.

### DIFF
--- a/src/etc/inc/interfaces.inc
+++ b/src/etc/inc/interfaces.inc
@@ -1,4 +1,4 @@
-<?php
+   <?php
 
 /*
  * Copyright (C) 2015-2020 Franco Fichtner <franco@opnsense.org>
@@ -3007,7 +3007,11 @@ function interface_dhcpv6_prepare($interface = 'wan', $wancfg, $linkdownevent = 
     $dhcp6cscript .= "*)\n";
     $dhcp6cscript .= "\t;;\n";
     $dhcp6cscript .= "esac\n";
-
+    $dhcp6cscript .= "if [ -z \"\${PINFO}\"]; then\n";
+    $dhcp6cscript .= "rm -f /tmp/{$wanif}_pdinfo\n";
+    $dhcp6cscript .= "else\n";
+    $dhcp6cscript .= "echo \$PINFO > /tmp/{$wanif}_pdinfo\n";
+    $dhcp6cscript .= "fi\n";
     @file_put_contents("/var/etc/dhcp6c_{$interface}_script.sh", $dhcp6cscript);
     @chmod("/var/etc/dhcp6c_{$interface}_script.sh", 0755);
 
--- a/src/www/status_interfaces.php
+++ b/src/www/status_interfaces.php
@@ -340,6 +340,14 @@ include("head.inc");
                     </tr>
 <?php
                     endif;
+                    if (file_exists("/tmp/{$ifinfo['if']}_pdinfo")): ?>                    
+                    <tr>
+                      <td><?= gettext("ISP assigned PD") ?></td>
+                      <td><?= file_get_contents("/tmp/{$ifinfo['if']}_pdinfo"); ?></td>
+                      
+                    </tr>  
+<?php
+                    endif;
                     if (!empty($ifinfo['gatewayv6'])): ?>
                     <tr>
                       <td><?= gettext("Gateway IPv6") ?></td>


src/etc/inc/interfaces.inc
```diff
@@ -3007,7 +3007,11 @@ function interface_dhcpv6_prepare($interface = 'wan', $wancfg, $linkdownevent =
     $dhcp6cscript .= "*)\n";
     $dhcp6cscript .= "\t;;\n";
     $dhcp6cscript .= "esac\n";
-
+    $dhcp6cscript .= "if [ -z \$PINFO ]; then\n";
```
fichtner, line 3010:
```suggestion
    $dhcp6cscript .= "if [ -z \"\${PINFO}\"]; then\n";
```
need to quote this, otherwise syntax error on empty expression

src/etc/inc/interfaces.inc
```diff
@@ -3007,7 +3007,11 @@ function interface_dhcpv6_prepare($interface = 'wan', $wancfg, $linkdownevent =
...
+    $dhcp6cscript .= "rm /tmp/{$wanif}_pdinfo\n";
```
fichtner, line 3011:
```suggestion
    $dhcp6cscript .= "rm -f /tmp/{$wanif}_pdinfo\n";
```

suppress error if file not found (we do not care)

src/www/status_interfaces.php
```diff
@@ -338,6 +338,14 @@ interface_configure(false, $interface, true);
                         endforeach;?>
                       </td>
                     </tr>
+<?php
+                    endif;
+                    if (file_exists("/tmp/{$ifinfo['if']}_pdinfo")): ?>
```
fichtner, line 343: in the long run it might be better to pull this through interfaces statistics provider functions, but for now this will suffice except that ifinfo['if'] is probably the wrong interface in some edge cases of IPv6

src/etc/inc/interfaces.inc
marjohn56: We do if it's got a pd the last time and hasn't on this call, it might show a PD that no longer exists.

marjohn56: OK, re-read that... we are deleting the file.. :)

marjohn56:
Updated with your amendments.

fichtner:
Merged via 39a3a912830, thanks!

maurice-w:
@marjohn56, I noticed that the delegated prefix line vanishes from Interfaces / Overview after a few days of uptime (not exactly sure yet). When this happens, the /tmp/_pdinfo file is gone, too. No other side effects though. A manual DHCP reload or system reboot brings it back.

marjohn56:
No idea why that would happen. The pdinfo file only gets deleted on a link down event.

maurice-w:
I checked the log, there were no link state changes since the last reboot (about 10 days ago). But the _pdinfo file definitely was missing when I checked earlier today. And it was back after triggering a WAN DHCP reload in the GUI.

I'll monitor this more closely and will try to spot any correlations when it happens again.

marjohn56:
The joys of mysterious events. :) I had thought of making dhcp6c delete the file at some point but changed my mind and never added that; so the only delete is in interfaces.inc as far as I know. Actually there might be another reason... let me check the dhcp6c script.... There's the problem. The script deletes the pdinfo file even if it's a renew, I'll push a PR to fix that.

marjohn56:
dhcp6c should re-write it even on a renew - very odd.

marjohn56:
Nope... does this look wrong... I think it does,,

if [ -z "${PDINFO}" ]; then
	rm -f /tmp/igb0_pdinfo
else
	echo ${PDINFO} > /tmp/igb0_pdinfo
fi

If I am correct it says if it exists, delete it, otherwise write it.... hence a renew will delete it. :)

marjohn56:
No, it's correct, if the var does not exist delete if it does then write it... it's good. So why it vanished from your system is still a mystery.

maurice-w:
My ISP's DHCPv6 leases have a preferred lifetime of 7 days and a valid lifetime of 14 days. From the logs I can see that OPNsense auto-renews after 3.5 days (half of the preferred lifetime). So I'll go check on Saturday and Sunday and will let you know.
What I can say is that a manual reload writes far more dhcp6c log entries than an automatic renew. But this might be by design?

marjohn56:
Yep, dhcp clients usually renew on half of the valid lifetime,  Here's a nice explanation from jupiter..
**https://www.juniper.net/documentation/en_US/junos/topics/concept/dhcpv6-lease-timers.html#:~:text=valid%2Dlifetime%E2%80%94Length%20of%20time,can%20no%20longer%20be%20used.&text=This%20timer%20is%20also%20called%20the%20rebind%20time.**

maurice-w:
The automatic renew happened and sure enough, the _pdinfo file was gone afterwards. Looking at the dhcp6c log, I might have an idea of what's going wrong:
While the prefix (IA_PD) and WAN address (IA_NA) do have the same lifetimes, they get renewed separately. First, IA_PD gets renewed, then `dhcp6c_wan_script.sh` gets executed. Two seconds later, IA_NA gets renewed and `dhcp6c_wan_script.sh` gets executed again. Might that second run of the script delete the _pdinfo file?

marjohn56:
hmm... nice one, perhaps.... I'll take a gander.

fichtner:
Why are we taking about "_pdinfo" .. the correct file is "em0_pdinfo" or similar... may just be simplification but details matter....

marjohn56:
only if it's a virtual interface, could be igb0 or whatever...

maurice-w:
My intention was to generalize it because the actual name depends on the interface. On my system (Hyper-V) it's actually `/tmp/hn0_pdinfo`.

marjohn56:
I know what you meant.,. :)

marjohn56:
OK, it's possible. if dhcpc6c does not generate the pdinfo env var then it will get deleted. Maybe we should only delete it on an interface down event, which already happens.... 

Thoughts @fichtner ?

marjohn56:
OK, we just need to make some changes to the  dhcp6c script. We'll move the creation of the files to the REQUEST response, rather than global. We can create on REQUEST and delete on RELEASE.

fichtner:
But if the release happens the PD should be gone ok? Or is this a RENEW issue?

marjohn56:
It's only ion RENEW, but I suspect dhcp6c does not generate the PDINFO on RENEW, hence the env var is not there and the script deletes any existing pdinfo file. So I'll move the file creation into the REQUEST|RELEASE case block and create it only on REQUEST and delete it only on release

fichtner:
sounds reasonable and fits the described situation by @maurice-w , nice catch

marjohn56:
@fichtner - What is the purpose of this :  **(!empty($syscfg['dhcp6_norelease']) ? 'EXIT' : 'RELEASE') . ")** ?

If the No RELEASE Flag is set dhcp6c does not respond with a RELEASE env var

fichtner:
d36f0f4f62557e

marjohn56:
So why not just  RELEASE|EXIT ?

fichtner:
I don't know. It's your change.

marjohn56:
It wasn't,  I don't like bourne scripts. :)

fichtner:
Ok, let's dig a bit deeper and reverse roles: me asking you what this means... https://github.com/opnsense/core/pull/2030/files#diff-1332c372788c9e1a8c6c9bae9ebb55a5R3211 ;)

marjohn56:
Well it appears I didn't remember then and I don;t remember now, IF I did do that, which I still find odd then I must have been having a good bourne shell day. However, I see no  reason ( pun intended ) for it and so the fix I've just issued removes that and just uses RELEASE|EXIT. It's still there in the code if we want to revert that bit.          

Fix is #4153

@maurice-w - Can you try it.

marjohn56:
I can check this with my system, my slave systems only get a 30 minute lease.,

marjohn56:
Yes, still there after a renew.

maurice-w:
Thanks guys, I applied 0e44420 and will report back after the next renew! :-)

maurice-w:
I can confirm that it now survives a renew. 👍

PR #4100, opened 2020-05-09
Add multiple interfaces to rtsold call

marjohn56:
We need to get all of the router info from rtsold so need to add multiple interfaces. Something like this, it works but maybe not the best way.

### DIFF
--- a/src/etc/inc/interfaces.inc
+++ b/src/etc/inc/interfaces.inc
@@ -2893,21 +2893,7 @@ function interface_6to4_configure($interface = 'wan', $wancfg)
         link_interface_to_track6($interface, true);
     }
 }
-function get_wan_dhcp6_interfaces()
-{
-    foreach (legacy_config_get_interfaces(['enable' => true, 'virtual' => false]) as $_interface => $_wancfg) {
-        if (empty($_wancfg['ipaddrv6']) || $_wancfg['ipaddrv6'] != 'dhcp6') {
-            continue;
-        }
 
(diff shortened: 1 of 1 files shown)


fichtner:
Geez, why can't these daemons run in parallel? :(

So there is a "-a" option here. If we make sure we only fire dhcp6fc for interfaces that have a valid config around (we can check this as we use it pre-merge for dhcp6c.conf) that would suffice, no?

marjohn56:
The daft thing is that if you read the rtsold(8) it sort of contradicts itself...  I've been going round in circles trying to figure out why I am getting the same route address for both my WAN interfaces... then it struck me, they are both VLANs off the same interface on the primary router!

OK, let me try the -a option, I assume that means we don't spec interfaces in the command?

fichtner:
Yep, this and d8c8129889aa should prevent any non-expected fallout

marjohn56:
-a option works, defaultgw6 and routerv6 are created for both my WAN interfaces.

marjohn56:
I'll do a full test once you pull this in.

fichtner:
Do you mean "pull this in" as in I add -a and close the PR, or do you want to amend this PR to include -a to be pulled in? :)

marjohn56:
:) doing the changes now...

marjohn56:
Argh.. this logging bug is £$%"%!!!! 

I THINK it's working OK. :)

fichtner:
Merged, thanks!

fichtner:
Master is now at bc5405bd54cd, what do you think?

Issue #4068, opened 2020-04-27
Syslog: optionally disable legacy (clog) logging


AdSchellevis:
In our transition to a more versatile logging subsystem, we added syslog-ng in 19.1.10, but to maintain compatibility with the existing installs, we made sure to keep support for circular log files using the non standard clog format.

Now it's time to offer the opportunity to disable the old syslog and switch to syslog-ng only, but keep clog available for existing installs for some time to come. (need to discuss, maybe remove support in 21.1 or 21.7)

Part of this change will be to offer a switch to disable the old syslog and make sure the ecosystem underneath can handle both (in combination).

Design choices / considerations:

- Use syslog-ng functionality to write data to stamped files (https://www.syslog-ng.com/technical-documents/doc/syslog-ng-open-source-edition/3.16/administration-guide/86), avoids the need to rotate (only simple cleanup).
- Log viewers will stitch legacy configuration files behind new style log files, so a legacy -> new log switch will be rather seamless, the other way around will have a wrong sorting.
- Cleanup of old (clog) files should be handled manually, either by letting the user flush his/her files from the existing "reset log files" or on a per topic basis.

todo:
- [x] replace `auth.info;authpriv.info;user.*					|exec /usr/local/sbin/sshlockout_pf 15`
- [x] update documentation
- [x] update development docs
- [x] update filter live log and others to support hybrid logs
- [x] add new templates to plugins
- [x] add cleanup script to remove old logs via cron
- [x] change flush log actions to support new structure

PR #4101, opened 2020-05-09
Syslog: optionally disable legacy (clog) logging

AdSchellevis:
This should be all for https://github.com/opnsense/core/issues/4068, while there also ditched sshlockout_pf for a simpler script.  

See issue for background on this change.


mimugmail:
Really nice, thank you! Businesses needs this for compliance. \o/

src/etc/inc/system.inc
```diff
@@ -644,7 +644,6 @@ news.err;local0.none;local3.none;local4.none			{$log_directive}/var/log/system.l
 local7.none							{$log_directive}/var/log/system.log
 security.*							{$log_directive}/var/log/system.log
 auth.info;authpriv.info;daemon.info				{$log_directive}/var/log/system.log
-auth.info;authpriv.info;user.*					|exec /usr/local/sbin/sshlockout_pf 15
```
fichtner, line 647: if we don't use it anymore please remove dependency from Makefile as well :)

src/opnsense/mvc/app/models/OPNsense/Core/ACL/ACL.xml
```diff
@@ -312,7 +312,8 @@
     <page-diagnostics-logs-firewall-summary>
         <name>Diagnostics: Logs: Firewall: Summary View</name>
         <patterns>
-            <pattern>diag_logs_filter_summary.php*</pattern>
+          <pattern>ui/diagnostics/firewall/stats*</pattern>
```
fichtner, line 315: indent is 2 here, usually 4 like before

src/etc/inc/system.inc
AdSchellevis: @fichtner will do :)

AdSchellevis:
@fichtner I think it's safe to merge, any objections?

fichtner:
Ps: plist?

Issue #23, opened 2015-01-09
System gateway groups edit requires cleanup and bootstrapping


jschellevis:
system_gateway_groups_edit.php needs cleanup.

PR #4102, opened 2020-05-10
Update util.inc

marjohn56:
Pull real pd from interface_pdinfo file.


marjohn56:
It all appears to work.

marjohn56:
@fichtner - If you want me to send you the dhcpdv6.conf and radvd.conf files so you can check the format ask away.

marjohn56:
Just noticed that in RADVD.conf, in auto mode its not setting the 'DepricatePrefix on option', which then took me to this - https://github.com/reubenhwk/radvd/issues/91

If this is true then we need to fix that.

marjohn56:
Additional note from Maurice who reports a failed test, is to migrate the no-release option. Debug cannot easily be migrated I don't think. tested again with a single interface and all good on my system.

marjohn56:
Maurice reports second attempt OK. He's not sure what went wrong with the first attempt but now working fine. There's a a new PR which adds the changes needed for killing and restarting dhcp6c on a norelease, debug level or duid change #4109. If we are happy that RADVD does not crash on deprecate set then I think that's next.

marjohn56:
bump... need these reviewed and pulled in for the next 20.7 beta.

marjohn56:
Looking at interfaces.php, I think we can move the prefix delegation size to become part of the send IPv6 prefix hint, i.e. only visible when the tickbox is checked. We no longer need to use the prefix delegation size itself in creating the dhcpdv6.conf as we pull that from the dhcp6c ENV variable ( that's in this commit ), its only use therefore is as part of the prefix hint. Many ISPs will ignore the prefix hint anyway, but some do.

marjohn56:
Note - THIS PR is obsolete when using dhcp6c-PDINFO---All-Interfaces-PR-#23. An updated PR #4178 replaces this PR.

PR #4108, opened 2020-05-12
Display dest port number in filter log widget

marjohn56:
Display the destination port number. V6 not working, according to existing code comments, but it is...


AdSchellevis:
@marjohn56 thanks, although I'm a bit afraid about the next guy complaining about the width of the content (which seems to be a constant up hill battle with these widgets), it looks useful enough to add.

PR #4109, opened 2020-05-13
Further changes to multiwan dhcp6c

marjohn56:
Move dhcp6 command  creation to re-usable function. Will either return the dhcp6c command or will terminate existing instance and restart. This is used in the system_interfaces_advanced file to restart dhcp6c when debug, log-level or duid have been changed.


fichtner:
I want to solve this via rc.conf script and a configd call to reload, but it has to wait till 20.1.7 is out next week Tuesday (hopefully).

marjohn56:
OK, I'll wait until then before going further.

PR #4111, opened 2020-05-13
Update text in Aliases to be more intuitive

marjohn56:
Change text and add help.

### DIFF
--- a/src/opnsense/mvc/app/views/OPNsense/Firewall/alias.volt
+++ b/src/opnsense/mvc/app/views/OPNsense/Firewall/alias.volt
@@ -613,8 +613,8 @@
                                 <tr id="row_alias.updatefreq">
                                     <td>
                                         <div class="control-label" id="control_label_alias.updatefreq">
-                                            <i class="fa fa-info-circle text-muted"></i>
-                                            <b>{{lang._('Expiration')}}</b>
+                                            <a id="help_for_alias.frequency" href="#" class="showhelp"><i class="fa fa-info-circle"></i></a>
+                                            <b>{{lang._('Refresh Frequency')}}</b>
                                         </div>
                                     </td>
                                     <td>
@@ -633,10 +633,13 @@
                                                 </tr>
                                             </tbody>
                                         </table>
+                                        <div class="hidden" data-for="help_for_alias.frequency">
+                                                               <small>
+                                                {{lang._('The frequency that the list will be refreshed, in days + hours, so 1 day and 8 hours means the alias will be refreshed after 32 hours. ')}}
+                                            </small>
+                                        </div>
                                     </td>
-                                    <td>
-                                        <span class="help-block" id="help_block_alias.updatefreq"></span>
-                                    </td>
+                                                                      
                                 </tr>
                                 <tr id="row_alias.content">
                                     <td>


src/opnsense/mvc/app/views/OPNsense/Firewall/alias.volt
```diff
@@ -633,10 +633,13 @@
                                                 </tr>
                                             </tbody>
                                         </table>
+                                        <div class="hidden" data-for="help_for_alias.frequency">
+                                                               <small>
+                                                {{lang._('The frequency that the list should be updated, where 1 would be daily, 7 would be weekly. Time is the time of day, e.g. 12:00 would be midday')}}
```
AdSchellevis, line 638: I'm not 100% sure we should change it, how it works is days+time equals the validity of the alias, so, if you're choosing 1+12 that would mean every 36 hours it would refresh. 

The mechanism is more comparable to TTL in that regard, hence it translates to alias ttl in our template 
https://github.com/opnsense/core/blob/ae5692b477db1701491bce55db5317b0a73728c0/src/opnsense/service/templates/OPNsense/Filter/filter_tables.conf#L27

So the question is, should we change the label, and if so, what would the phrase most people will understand.

marjohn56: So it doesn't go and refresh the list then?

AdSchellevis: it refreshes the list after that amount of time, not on a specific time (12.00)

marjohn56: So change it to 'Refresh Frequency and change the help to - 'The alias will be refreshed after the given time has expired', or leave it as it is and update the Wiki as it doesn't match what's in the GUI - one or the other.

AdSchellevis: I'm ok with that, good to have this discussion.

marjohn56: :) Yes, but which, change the GUI or Wiki or both?

AdSchellevis: as you proposed it, I would expect both :) The help text triggered me, I can probably live with most variations, as long as it's more clear for others what we mean... the help text pointed to a misunderstanding.

marjohn56: OKey dokey.. will mod accordingly..

src/opnsense/mvc/app/views/OPNsense/Firewall/alias.volt
```diff
@@ -623,7 +623,7 @@
                                             <thead>
                                                 <tr>
                                                     <th>{{lang._('Days')}}</th>
-                                                    <th>{{lang._('Hours')}}</th>
+                                                    <th>{{lang._('Time')}}</th>
```
AdSchellevis, line 626: better leave this hours, you can't input time there :)

marjohn56:
Have done so, and pushed a wiki update...

src/opnsense/mvc/app/views/OPNsense/Firewall/alias.volt
```diff
@@ -633,10 +633,13 @@
...
                                         </table>
+                                        <div class="hidden" data-for="help_for_alias.frequency">
+                                                               <small>
+                                                {{lang._('The frequency that the list will be refreshed, in days + hours, so 1 day and 8 hours means the alias will be refreshed after 32 hours. ')}}
+                                            </small>
+                                        </div>
                                     </td>
-                                    <td>
-                                        <span class="help-block" id="help_block_alias.updatefreq"></span>
-                                    </td>
```
AdSchellevis, line 639: ^^ this looks like a mistake, these `<td/>` tags are used for feedback

marjohn56: ?? The first td is around the help click icon, same as alias name. The second block is around the inputs for day and hour with the hidden help within that block. Looks OK to me and works fine.

AdSchellevis: well, it wouldn't if there where validations set to updatefreq, such as 
image

in the name

marjohn56: So why are they there around the alias.enabled help icon too?

AdSchellevis: same reason, 3 column table, normally our grids are generated automatically (which makes sure the feedback column is always there) https://github.com/opnsense/core/blob/master/src/opnsense/mvc/app/views/layout_partials/form_input_tr.volt#L98-L100

trust me, it should be there....

marjohn56: Well if I remove them it doesn't work. You're the expert show me how it should be done.

AdSchellevis:
ok, done https://github.com/opnsense/core/commit/cfa0b713cdaabe7ca704e491f16fe3ddb79d6646

marjohn56:
I don't understand that at all. :)

AdSchellevis:
but it works, doesn't it?

marjohn56:
Yes it does, but it worked without that addition, THAT's why I don't understand. :)

PR #4112, opened 2020-05-13
Proxy: add connect_timeout

### LINKED ISSUE #4092: Squid: Allow configuration of IPv4 fallback
In my Opnsense configuration the squid proxy connects my IPv4 LAN with the outside IPv6 world in the internet. Both IPV4 and IPV6 internet access work fine, whereas squid automatically prefers IPv6 over IPv4 in case a server is reachable over both protocols.

But, if squid tries to access a server via IPv6 (indicated by related AAAA DNS record) an the server is faulty, it tries to fallback to IPv4 after 60 seconds (timeout) in its standard configuration. This timeout is too long, so that many browsers cancel the initial request.

I solved this issues by adding a manual configuration directive to the squid configuration file (/usr/local/etc/squid/squid.conf):

`connect_timeout 7 seconds`

But manually modifying configuration files is not a good idea, because they get overwritten when reconfiguring or restarting Opnsense.

Fallback test can be performed via https://ipv6-test.com/

**Idea:**
- It would be a great idea to make the above configuration directive configurable via a form field (value in seconds) in the webgui (e.g. Services -> Web proxy -> Administration -> Forward proxy -> General Forward settings).

**Opnsense version:**
 - OPNsense 20.1.6-amd64
 - FreeBSD 11.2-RELEASE-p19-HBSD
 - LibreSSL 3.0.2

mimugmail:
Closes #4092

### DIFF
--- a/src/opnsense/service/templates/OPNsense/Proxy/squid.conf
+++ b/src/opnsense/service/templates/OPNsense/Proxy/squid.conf
@@ -444,6 +444,10 @@ visible_hostname {{OPNsense.proxy.general.VisibleHostname}}
 # Define visible email
 cache_mgr {{OPNsense.proxy.general.VisibleEmail}}
 {% endif %}
+{% if helpers.exists('OPNsense.proxy.general.connecttimeout') and OPNsense.proxy.general.connecttimeout != ''%}
+# Set connection timeout
+connect_timeout {{OPNsense.proxy.general.connecttimeout}} seconds
+{% endif %}
 {%  if helpers.exists('system.language') and system.language != "" %}
 error_directory /usr/local/etc/squid/errors/{{ system.language|lower|replace("_","-")}}
 {% endif %}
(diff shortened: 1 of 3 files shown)


src/opnsense/service/templates/OPNsense/Proxy/squid.conf
```diff
@@ -444,6 +444,10 @@ visible_hostname {{OPNsense.proxy.general.VisibleHostname}}
 # Define visible email
 cache_mgr {{OPNsense.proxy.general.VisibleEmail}}
 {% endif %}
+{% if helpers.exists('OPNsense.proxy.general.connecttimeout') and OPNsense.proxy.general.connecttimeout != '' %}
```
AdSchellevis, line 447: `{% if not helpers.empty('OPNsense.proxy.general.connecttimeout') %}` <--

mimugmail: Didn't know this one, thx :)

AdSchellevis:
@mimugmail thanks!

Issue #4113, opened 2020-05-13
IDS: all enabled snort rules are configured with a double action and will not load


Tra5is:
**Important notices**
Before you add a new report, we ask you kindly to acknowledge the following:

[X] I have read the contributing guide lines at https://github.com/opnsense/core/blob/master/CONTRIBUTING.md

[X] I have searched the existing issues and I'm convinced that mine is new.

**Describe the bug**
Recently noticed that enabled rules are configured with an action that prevents them from being loaded correctly. 

suricata: [100194] <Error> -- [ERRCODE: SC_ERR_INVALID_SIGNATURE(39)] - error parsing signature "**drop alert** tcp $EXTERNAL_NET any -> $HOME_NET $HTTP_PORTS (msg:"SERVER-WEBAPP Western Digital MyCloud jqueryFileTree.php command injection attempt"; flow:to_server,established; content:"/web/addons/jqueryFileTree.php"; fast_pattern:only; http_uri; content:"%26"; http_raw_uri; pcre:"/[?&](host|pwd|user|dir|lang)=[^&]*?%26/Ii"; metadata:policy max-detect-ips drop, policy security-ips drop, service http; reference:url,www.exploitee.rs/index.php/Western_Digital_MyCloud; classtype:web-application-attack; sid:46860; rev:1;)" from file /usr/local/etc/suricata/opnsense.rules/snort_vrt.server-webapp.rules at line 3493

suricata: [100194] <Error> -- [ERRCODE: SC_ERR_UNKNOWN_PROTOCOL(124)] - protocol "alert" cannot be used in a signature. Either detection for this protocol supported yet OR detection has been disabled for protocol through the yaml option app-layer.protocols.alert.detection-enabled

[...]

PR #4115, opened 2020-05-14
IDS: Handle rules prefixed with '# '

Tra5is:
The issue I noticed was that some rules from snort were generating errors during load. Ref: #4113

The snort rulesets prefix some rules that do not have an explicit policy set on them with '# '. It looks like the suricata scripts expected only '#'. 

For example from snort:
    # alert tcp $HOME_NET any -> $EXTERNAL_NET $HTTP_PORTS (msg:"EXPLOIT-KIT Flim exploit kit outbound jar request"; flow:to_server,established; urilen:14; content:".jar"; http_uri; content:" Java/1."; http_header; pcre:"/^\/[a-f0-9]{9}\.jar$/U"; metadata:policy max-detect-ips drop, policy security-ips drop, service http; classtype:trojan-activity; sid:26963; rev:2;)

The rulecache.py was splitting on the ' ' so if a rule started with a space, then the action would be parsed as blank, and the space would be replaced by whatever action the user tried to configure the rule with. This would result in rules being prefixed with something like 'drop alert' resulting in the suricata parsing error during rule loading. 

1. Handle the case of applying a drop filter by being able to match '# alert', not just '#alert'
2. Strip any prefix spaces after discarding the # during rule parsing

### DIFF
--- a/src/etc/inc/filter.lib.inc
+++ b/src/etc/inc/filter.lib.inc
@@ -446,32 +446,31 @@ function filter_core_rules_system($fw, $defaults)
             default:
               if (isset($config['dhcpdv6'][$intf]['enable']) || isset($intfinfo['track6-interface'])
                 || (!empty($config['dhcrelay6']['interface']) && in_array($intf, explode(',', $config['dhcrelay6']['interface'])))) {
-                    $fw->registerFilterRule(5,
+                    $fw->registerFilterRule(1,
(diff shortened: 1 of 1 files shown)


AdSchellevis:
@Tra5is can you try https://github.com/opnsense/core/commit/38ea28d0ad24ce82ac3106e2ee9dc35088fbdb92 ? parsing seems to be a bit sensitive indeed.

Tra5is:
I actually don't have an environment that I can test this in. That's why my changes looked that way. I'll try to set something up this weekend.

Tra5is:
I've created another PR to finish the fix for this issue.

PR #4120, opened 2020-05-19
firewall: Display alias description and alias content in tooltip

devNan0:
**Before:**

**After:**

**Description:**
Sometimes description and content of an alias is quite useful when working on firewall rules.
Currently its only possible to display either description (if set) or the content (if description is empty) of an alias in the tooltip.

With this PR both description (if set) and content (if any) will be shown in the tooltip.

I made the parameter `with_content` optional so existing stuff doesn't break.

### DIFF
--- a/src/etc/inc/legacy_bindings.inc
+++ b/src/etc/inc/legacy_bindings.inc
@@ -152,13 +152,16 @@ function legacy_move_config_list_items($source, $id, $items)
 
 /**
  * Retrieve description text of firewall alias by specifying the alias name.
+ *
  * @param string $name alias name to retrieve the description from
+ * @param bool   $with_content (optional) return description and content if set
+ *
  * @return string|null description of alias, or null if no description was set or alias
  *   does not exist.
  */
-function get_alias_description($name)
+function get_alias_description($name, $with_content = false)
 {
-    return OPNsense\Firewall\Util::aliasDescription($name);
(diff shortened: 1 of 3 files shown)


AdSchellevis:
We should discuss if we want to add content when a description is provided, to me it sounds quite duplicate, but if more people agree that it's a useful addition we should consider changing our code. The parameter addition doesn't look very useful, since the method isn't used anywhere else.


I'll keep the PR open for now to see if we can collect more feedback.

mimugmail:
If code quality is ok I'm in. Especially when troubleshooting customer firewalls where you are not always on it's sometimes confusing to see content (where you'd need a description which is lack of doing) and sometimes you need to know the addresses e.g. for hunting via tcpdump and then you have to edit alias first to see contents. Mix would be great :)

fichtner:
As long as we cap the displayed content. Large lists as tooltips are not useful either.

mimugmail:
It already limited to 10, I think we can keep it like this (10) and just add the description

AdSchellevis:
I'll take a look, we can simplify the code and add the feature, I don't mind.

L1ghtn1ng:
I would like to see this feature added as it is majorly useful

AdSchellevis:
ok, added the feature in https://github.com/opnsense/core/commit/3a20e5a61489afb412af74ac25664322c1041e03

PR #4129, opened 2020-05-24
When enabling rules prefixed with '# ' consume the extra space

Tra5is:
Further to the changes in https://github.com/opnsense/core/commit/38ea28d0ad24ce82ac3106e2ee9dc35088fbdb92

The installRules.py script also needs to handle an extra space in the rule.

### DIFF
--- a/src/opnsense/scripts/suricata/installRules.py
+++ b/src/opnsense/scripts/suricata/installRules.py
@@ -61,12 +61,8 @@ if __name__ == '__main__':
                     and rule_info_record['metadata']['sid'] in rule_updates:
                 # search last comment marker
                 for i in range(len(rule_info_record['rule'])):
-                    if rule[i] != '#':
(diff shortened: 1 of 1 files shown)


src/opnsense/scripts/suricata/installRules.py
```diff
@@ -64,6 +64,10 @@
                     if rule[i] != '#':
                         break
 
+                #consume an extra space before generating altered rule
+                if rule[i] == ' ':
```
AdSchellevis, line 68: it's probably better to move it up to the existing loop, such as `if rule[i]  not in ['#', ' ']:`

Tra5is: Thanks! Much nicer!

AdSchellevis: @Tra5is thanks for taking care of this

PR #4135, opened 2020-05-26
dhcpdv6, fix stateless mode for track6 interfaces

maurice-w:
Only marry live prefix to user-specified range if range is not empty.
Bug report in the forum: https://forum.opnsense.org/index.php?topic=17423.0


marjohn56:
Tested it manually, i.e. clearing the range and restarting dhcpd and it works fine.

fichtner:
Merged, thanks!

PR #4137, opened 2020-05-27
Squid - Allow PURGE on cache

sazb:
Hello

This is a suggestion to enable the PURGE method for handling Squid cache.

Once local disk cache is enabled, the ACL is ready and at this moment on CLI:
squidclient -v -m PURGE http://....
will remove a wrong cached file

Could be available through portal and/or enhanced the current ACL to ask for authentication...

Thanks in advance for your attention.


AdSchellevis:
@sazb your right, purge should be enabled from localhost. thanks!

PR #4152, opened 2020-06-07
dhcp6c server additions

marjohn56:
These additions are to be used for testing the updated dhcp6c serverv6 function, where dhcp6c now generates an ENV VAR containing the address of the server that gave out the lease(s). This can be used where RTSOLD gets no response. Much testing is needed, but on systems where I have disabled rtsdold this works.

ref https://github.com/opnsense/dhcp6c/pull/22

### DIFF
--- a/src/etc/inc/interfaces.inc
+++ b/src/etc/inc/interfaces.inc
@@ -1127,6 +1127,9 @@ function interface_bring_down($interface = "wan", $ifacecfg = false)
 
         if (!empty(trim(@file_get_contents("/tmp/{$realifv6}_routerv6")))) {
             $pfctlflush[$realifv6] = 1;
+        } else if (!empty(trim(@file_get_contents("/tmp/{$realifv6}_serverv6")))) {
+			// XXX - might be a dhcp6c gateway
+            $pfctlflush[$realifv6] = 1;
         }
 
         $ip6 = find_interface_ipv6($realifv6);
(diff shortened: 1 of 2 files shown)


fichtner:
I would rather treat it was the substitute router. Either it works or it doesn't. We do not need to track more 1-off state variables.

marjohn56:
OK, whichever way you want to play it. I was going for the safety approach that we use the time honoured RTSOLD method, if that fails then check to see if there's a dhcp6c server address, which I guess there has to be.

fichtner:
We simply check if a file exists and if it does we do not taint it, otherwise we write the only probably address. I'm not sure this works for all ISPs, but it is worth a try.

marjohn56:
Erm, hold hard. Not sure I follow. If the RTSOLD file exists then I use that, if it does not I check for the dhcp6c file, if that does not exist then it falls back to the existing no file method. What do you mean by             'check if a file exists and if it does we do not taint it', that's really what is happening.

fichtner:
Yes, but without the redirect through serverv6. We write routerv6 directly if we don't have any.

marjohn56:
Ah.. I see said the blind man - So just check for the existence of routerv6 in the dhcp6c script, if not there then write the dhcp6c address to it?

fichtner:
Yup :)

marjohn56:
OK, closing and deleting this, I'll go play with that method.

PR #4168, opened 2020-06-13
Valid IPSec Phase 2 hash config warning raises GUI alert.

BrettMerrick:
When AES-GCM encryption is selected the hashing algorithm should be NULL as AES-GCM has message authentication built in.

The PHP warning raised by supplying in_array() with an empty string results in warning on the dashboard that links to:

[12-Jun-2020 17:08:01 Pacific/Auckland] PHP Warning:  in_array() expects parameter 2 to be array, string given in /usr/local/www/vpn_ipsec_phase2.php on line 713


AdSchellevis:
@BrettMerrick nice catch, thanks!

PR #4178, opened 2020-06-19
dhcpd6-Changes-for-dhcp6c-PDINFO---All-Interfaces-PR-#23

marjohn56:
The changes to dhcp6c mean that the PDINFO returned can contain multiple interfaces when there are multiple tracked interfaces. Thus the changes here handle that.


bsd-source:
Thanks marjohn56 for working on OPNsense IPv6 releted issues. I know a lot of other developement is being done with with other parts of OPNsense but I think IPv6 related functionality is being overlooked. I've tinkered with IPv6 with OPNsense and have found numerous issues with simple coding logic and other issues causing IPv6 to not function properly. Sometimes it works, but most of the time it doesn't.  If you fork OPNsense with all your IPv6 changes I'm wondering if you can release some test **beta** builds not alpha builds. Seems to me you're treading in water and going nowhere with your pull requests.

fichtner:
Think again please. IPv6 is overly complicated and scattered. Users are not able to debit and FreeBSD 12 is giving us further challenges. If IPv6 isn’t progressing fast enough it’s not because nobody else works on it. If you have these insights into all the bugs feel free to provide PRs now rather than forking this work away.

Also one would have to agree that adding features on top of all the alleged bugs is not a sustainable way forward. ;)

marjohn56:
Just to add, I'm not forking anything. My PR's are there for the devs to do with as they see fit. BTW, this is not a feature, if you have multiple dhcp6 WAN then as explained the PDINFO will be incorrect. That only really comes into play when all the other related PRs are pulled together but I realise it's a lot of work and their are more important issues to deal with.

Issue #4132, opened 2020-05-26
Traffic shaper ACK packet selection is too wide and should be narrowed down.


Nimloth:
As discussed here: https://forum.opnsense.org/index.php?topic=17372.0
the drop-down entry 'tcp (ACK packets only)' in the 'Edit rule' dialog of the Traffic Shaper will match all packets with 'tcpflags ack'.

Since this will in effect match almost all traffic, it is not particularly useful.

So far as I could see this would only need changes in service/templates/OPNsense/IPFW/ipfw.conf
and the suggested changes are (so far) to replace 'tcpflags ack' by either 'tcpflags ack,!psh' or 'tcpflags ack iplen 52'.

It might also be important to communicate any modifications as it will change current behavior.

Note: the original issue that prompted the addition of the feature was https://github.com/opnsense/core/issues/528

maxfield-allison:
Have you had a chance to make this change and test? I'm making the commit to my current fork and may set up a VM if no one else has had a chance to try yet.

Nimloth:
> Have you had a chance to make this change and test? I'm making the commit to my current fork and may set up a VM if no one else has had a chance to try yet.

While I have not changed and tested the template, I have run a production firewall with 'tcpflags ack iplen 52' for some time now - I edited the /usr/local/etc/ipfw.rules file directly, though.

Nimloth:
Note: There was no consensus in the discussion in the forum as to what the appropriate rule should be, only that the current rule is to generic and therefore not very useful.

maxfield-allison:
Could you point me to a resource that explains iplen 52? or give me a quick and dirty as to why you chose that instead of !psh?

maxfield-allison:
I just made the change you suggested in my ipfw.rules and reloaded all. looks like it did the trick!

Nimloth:
Sure. From what I can recall, 40 bytes is what you need for tcp+ip(v4) header, e.g. what you need for an ACK or RST packet. An ACK with timestamp needs 52 bytes. (See http://www.faqs.org/rfcs/rfc1323.html - 1.3 Using TCP options).

maxfield-allison:
You're the best, I'm going to change the template on my test device and verify that it works for the drop down option and then submit a pull.

maxfield-allison:
You were correct, The change in the template is tested and working for both tcp ack and non-ack

Nimloth:
[...]

PR #4180, opened 2020-06-19
Update tcp (ACK packets only) to specifically target only ACK packets

maxfield-allison:
Changed `tcpflags ack` to `tcpflags ack iplen 52` in src/opnsense/service/templates/OPNsense/IPFW/ipfw.conf to prevent the option in firewall shaper from capturing all tcp traffic. Tested and confirmed to work on a production system with both `tcp (ACK packets only)` and `tcp (non-ACK packets)`. Change discussed and tested in #4132 with @Nimloth


AdSchellevis:
I've posted my feedback in the ticket https://github.com/opnsense/core/issues/4132#issuecomment-646960369

maxfield-allison:
Feedback addressed in #4181

Issue #528, opened 2015-12-14
Feature Request: Traffic shaper, Ack queue 


MaartenJB:
It would be nice if it was possible to assign a separate Ack queue to the shaper rule, so Ack packages can be prioritized.

In practise:
Make 2 queue's, one high prio and one low prio.
Edit shaper rule, assign the high prio queue to the "Ack queue" option and the low prio queue to "queue" option.

PR #4181, opened 2020-06-23
Add the iplen option to shaper rules

maxfield-allison:
As discussed in #4132 and prompted by #528 
This change enables users to set the maximum packet length a rule should be applied to. Tested on a production system.


AdSchellevis:
@maxfield-allison looks good, thanks!

PR #4182, opened 2020-06-26
Multi-WAN Failover

marjohn56:
An issue has been identified in issue 4160 whereby when dpinger detects a failure the gateway is not switching.  Although the script 10-dpinger is called the call to 'configctl filter reload' does not function, changing this to give the full path to configctl appears to fix the issue.

### DIFF
--- a/src/etc/rc.syshook.d/monitor/10-dpinger
+++ b/src/etc/rc.syshook.d/monitor/10-dpinger
@@ -35,6 +35,6 @@ fi
 /usr/bin/logger -t dpinger "GATEWAY ALARM: ${GATEWAY} (Addr: ${2} Alarm: ${3} RTT: ${4}ms RTTd: ${5}ms Loss: ${6}%)"
 
 echo -n "Reloading filter: "
-configctl filter reload
+/usr/local/sbin/configctl filter reload
 
 exit 0


AdSchellevis:
@marjohn56 since the other syshooks do use the same method, it sounds like there's something wrong with the environment in this case. We probably better examine why the path seems incomplete here to void future issues.

marjohn56:
I had the same thought, but I would really like someone else to confirm my find just to prove its not just my system that this fixes!

fichtner:
Can you try to dump an “env” call to a temporary file?

    # env | sort > /tmp/envdebug

marjohn56:
It's in the path, at least when from ssh it is:

BLOCKSIZE=K
EDITOR=vi
GROUP=wheel
HOME=/root
HOST=OPNsense.queens-park.com
HOSTTYPE=FreeBSD
LOGNAME=root
MACHTYPE=x86_64
MAIL=/var/mail/root
OSTYPE=FreeBSD
PAGER=less
PATH=/sbin:/bin:/usr/sbin:/usr/bin:/usr/local/sbin:/usr/local/bin:/root/bin
PWD=/root
REMOTEHOST=192.168.1.100
SHELL=/usr/local/sbin/opnsense-shell
SHLVL=1
SSH_CLIENT=192.168.1.100 58546 22
SSH_CONNECTION=192.168.1.100 58546 192.168.1.254 22
SSH_TTY=/dev/pts/0
TERM=xterm
USER=root
VENDOR=amd

marjohn56:
Also there if I login directly.

fichtner:
Sure, I mean from dpinger script...

marjohn56:
Doing that now..

marjohn56:
Path is missing.

PATH=/usr/bin:/usr/sbin:/bin:/sbin
PWD=/usr/local/opnsense/service
interface=igb0
new_broadcast_address=10.4.12.255
new_dhcp_lease_time=7200
new_dhcp_message_type=5
new_dhcp_server_identifier=10.4.12.254
new_domain_name=queens-park.com
new_domain_name_servers=10.4.12.254
new_expiry=1593160500
new_interface_mtu=0
new_ip_address=10.4.12.192
new_network_number=10.4.12.0
new_routers=10.4.12.254
new_subnet_mask=255.255.255.0
reason=BOUND

marjohn56:
So in the /etc/rc file ?

fichtner:
Does f6a7df64fe work then?

marjohn56:
testing now

marjohn56:
Yup... that fixes it, nice one. I wonder how many other things that fixes. 👍

marjohn56:
Now can you fix my link-local no route to host problem. :)

fichtner:
Just dpinger since it calls rc.syshook directly. All other facilities use a root shell or the jump pad via configd where we fixed this a long time ago. Still nice catch, thanks!

Err uh, we will see about that next week. I have a birth to attend to any minute now....

mimugmail:
> Err uh, we will see about that next week. I have a birth to attend to any minute now....

Gogogo .. more core dev's needed 🚨

PR #4201, opened 2020-07-08
Blackhawk: Add primary_ip4 jinja macro

lattera:
This macro gets the primary IPv4 address, if it exists, of the requested
interface.

Signed-off-by:	Shawn Webb <swebb@blackhawknest.com>

### DIFF
--- a/src/opnsense/service/templates/OPNsense/Macros/interface.macro
+++ b/src/opnsense/service/templates/OPNsense/Macros/interface.macro
@@ -9,3 +9,15 @@
 else %}{{name}}{%
 endif
 %}{%- endmacro %}
+
+{#
+  macro primary_ip4, return primary IPv4 address of interface (e.g. lan -> 192.168.1.1)
+#}
+{% macro primary_ip4(name) -%}
+{% if helpers.exists('interfaces.'+name+'.ipaddr')
+%}{{
+      helpers.getNodeByTag('interfaces.'+name+'.ipaddr')
+}}{%
+else %}{{name}}{%
+endif
+%}{%- endmacro %}


AdSchellevis:
Wouldn't the one-liner `helpers.getNodeByTag('interfaces.myInterface.ipaddr')` return the same?  since we don't use the macro on our end, it might be rather dead code in core. (I know `physical_interface()` might have the same issue, but it has current use-cases, if I'm not mistaken it was originally added while figuring out if there was more logic to determine the interface)

lattera:
I don't really have any strong feelings either way. I was just trying to stick to current conventions.

Though, I do find this more elegant:

```
primary_ip4(iface)
```

than:

```
helpers.getNodeByTag('interfaces.' + iface + '.ipaddr')
```

in templates needing such functionality. It's a bit more succinct. But, again, I don't have any strong feelings on its inclusion in OPNsense.

fichtner:
Adding a function is problematic if we don't intend to increase the scope of the task. This is especially difficult with "primary" which already exists as a complex runtime concept:

https://github.com/opnsense/core/blob/7c43c8be36fee6438d0221904c70448000b6e3c1/src/etc/inc/interfaces.inc#L4591-L4632

For just reading a single configuration value that can also be an invalid IP address (e.g. "dhcp" or "none") we also have to think about the error cases and fallbacks... and generally we try not to offer system runtime information in the templates.

lattera:
Good catch. In my case, I have a definite need to grab the interface's IP address for a template config file. The config file in question is for a proprietary app that needs to be told which IP to listen on.

What would be the better route to take in this particular case?

lattera:
Actually, I just looked at the Squid template. I think I can base my template's needs off of that.

AdSchellevis:
To avoid addressing issues by the way it's often a good idea to bind to an (additional) loopback interface, it avoids quite some dynamic addressing struggles.

lattera:
This proprietary app only supports binding to a single address. I guess I'll probably need to figure out how to detect when/if an iface's IP changes.

I'll close this PR since I know now it's definitely not the right approach to take. Thanks for the input!

Issue #4209, opened 2020-07-14
Services: Network Time: Status - GPS Clock coordinates and gmaps link incorrect


mikahe:
**Important notices**
Before you add a new report, we ask you kindly to acknowledge the following:

[x] I have read the contributing guide lines at https://github.com/opnsense/core/blob/master/CONTRIBUTING.md
[x] I have searched the existing issues and I'm convinced that mine is new.

**Describe the bug**
Network Time Status page shows incorrect Clock Latitude and Longitude compared to GPS. Formatting and conversions aren't correct. Also Google Maps Link has incorrect coordinates.

**To Reproduce**
1. Configure Network Time with GPS Generic/NMEA serial device
2. Go to 'Services: Network Time: Status'
3. Scroll down to 'Clock Latitude'
Clock Latitude and Clock Longitude
dedd.ddddd (ddmm° mmmm.00000 [N|S|E|W])

In first field I'm seeing that degrees are +1 compared to correct full degrees. Also decimal points and degrees symbol are in incorrect places.

No conversion between decimal degrees, degrees and decimal minutes (or degrees, minutes, decimal seconds) should change the degrees part of the coordinate.

**Expected behavior**
Clock Latitude and Clock Longitude should probably be shown in the below format (at least in parenthesis dd° mm.mmmm is correct location with the shown digits, and that's also in NMEA GPGGA message, i.e. ddmm.mmmm):

Clock Latitude and Clock Longitude
dd.ddddddd (dd° mm.mmmm [N|S|E|W])

Google Maps Link should be based on decimal degrees:
http://maps.google.com/?q=dd.ddddddd,dd.ddddddd

**Environment**
OPNsense 20.1.8_1-amd64

mikahe:
The coordinates are apparently output here: src/www/status_ntpd.php:
Latitude:
<td><?= sprintf("%.5f", $gps_lat); ?> (<?= sprintf("%d", $gps_lat_deg); ?>&deg; <?= sprintf("%.5f", $gps_lat_min*60); ?><?= $gps_lat_dir ?>)</td>

mikahe:
This is how the incorrect Lat sum is calculated:

lat+lon	6240.45
	
gps_lat_deg	6126
gps_lat_min	114.45
	
sum	of above = 6240.45

mikahe:
The root cause of the bug is that degrees and whole minutes are combined in gps_lat_deg, so that's not actually correct. gps_lat_min contains only decimal part of minutes.

[...]

PR #4212, opened 2020-07-17
widgets/ntp, php warnings if no GPS fix and thus lat+lon missing from NMEA msg

mikahe:
In use case of issue #4209 (GPS time source for ntpd), occasionally System Firmware Reporter shows problem reports to send due to these, an example log:

PHP Errors:
[16-Jul-2020 21:23:51 Europe/Helsinki] PHP Warning:  A non-numeric value encountered in /usr/local/www/widgets/widgets/ntp_status.widget.php on line 85
[16-Jul-2020 21:23:51 Europe/Helsinki] PHP Warning:  A non-numeric value encountered in /usr/local/www/widgets/widgets/ntp_status.widget.php on line 87

This commit is using $gps_ok to skip lat+lon calculations with missing coordinates


AdSchellevis:
@mikahe thanks!

PR #4221, opened 2020-07-22
Select better binding address for dpinger

sammie-p:
Check all the interface IPs to preference a binding address that is in the same subnet as the gateway monitor address, rather than picking the first address. This fixes an issue for when CARP is used with a single WAN IP address and the adapter's first address is a private one.

This may not be the most elegant solution but it works well for us.

### DIFF
--- a/src/etc/inc/plugins.inc.d/dpinger.inc
+++ b/src/etc/inc/plugins.inc.d/dpinger.inc
@@ -87,6 +87,32 @@ function dpinger_configure()
     );
 }
 
+function ip_in_range($testip, $ip, $netmask) {
+    if (!empty($netmask) && !empty($ip)) {
+        if (strpos($netmask, '.') !== false) {
+            // $netmask is a 255.255.0.0 format
+            $netmask = str_replace('*', '0', $netmask);
+            $netmask_dec = ip2long($netmask);
+            return ( (ip2long($testip) & $netmask_dec) == (ip2long($ip) & $netmask_dec) );
+        } else {
+            // $netmask is a CIDR size block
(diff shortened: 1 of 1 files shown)


AdSchellevis:
We try to prevent such logic, since it has the tendency to lead to hard to track issues. Athough I agree the current  `find_interface_ip*()` isn't great, we should try to make it explicit if we want to force the use of an (other) address to avoid future similar issues.

fichtner:
How about an interfaces_primary_address6() IPv4 variant to be used here? It has replaced a couple of problematic find_interface_ipv6() calls and seems to work well enough.

marjohn56:
Talking of dpinger gateway issues, Did you see the email I sent you regarding the issue with link local?

sammie-p:
Would a better solution be to provide in the GUI an option to specify or choose a binding address for dpinger? I only wrote the patch quickly to solve the issue for us where we have load-balanced multi-WAN (2 connections) and only 1 public IP on each Internet connection. Works for us but it's not perfect in its current form.

AdSchellevis:
for IPv4 this https://github.com/opnsense/core/commit/5509fabfa3c3df1460c9d32a2397a4e25dbe9e55 should match to a better candidate using the functions already in util.inc.

AdSchellevis:
@sammie-p can you test https://github.com/opnsense/core/commit/5509fabfa3c3df1460c9d32a2397a4e25dbe9e55 on your end?  thanks in advance

sammie-p:
@AdSchellevis the patch works fine for me, thanks.

Issue #4217, opened 2020-07-22
Dashboard Interfaces Widget Table Auto Sizing


gitwer:
[x] I have read the contributing guide lines at https://github.com/opnsense/core/blob/master/CONTRIBUTING.md
I have but cant see how to create a "cleanup" issue.

[x] I have searched the existing issues and I'm convinced that mine is new.
None I could find for this specifically

**Is your feature request related to a problem? Please describe.**
The interfaces widget on the Dashboard has hard coded row widths which  results in lots of space for IP addresses but wrapped interface names quite easily.

**Describe the solution you'd like**
Either
1) let the col's be auto width by removing the `/width:\d+%; /` bits (similar to the Gateways table) OR
2) let most of the col's be auto width with the up/down status arrow having a fixed small width of 30px ish (similar to the Services table and the status col) and either where it is or move it on the left and keep the icons together.

**Describe alternatives you've considered**
The 2 listed above.

**Additional context**
Screenshots:
1) Default
![1_Interfaces_Default](https://user-images.githubusercontent.com/25402147/88137591-37ce0800-cc2f-11ea-9fcd-69b1de1354d1.png)

2) All Auto width
![2_Interfaces_AllAuto](https://user-images.githubusercontent.com/25402147/88137619-4a484180-cc2f-11ea-940f-f4480526baa3.png)

3) up/down col at 30px in place
![3_Interfaces_30px](https://user-images.githubusercontent.com/25402147/88137633-503e2280-cc2f-11ea-9909-bca230b5c504.png)

4) up/down col at 30px on left
![4_Interfaces_30pxOnLeft](https://user-images.githubusercontent.com/25402147/88137642-56cc9a00-cc2f-11ea-9f02-18f4d6b04f77.png)

/src/www/widgets/widgets/interface_list.widget.php
Code: for loop creating rows in table on lines 115 to 172

I prefer number 4 personally, but any of 2-4 would halve the table of the widget on my dashboard (and hopefully not break others).

EDIT: I also think the 5% col (up/down status) is missing and end TD. I will see if I can do one of these magic pull requests...

gitwer:
Hopefully I created the pull request correctly (went for option 4) - https://github.com/opnsense/core/pull/4222

marjohn56:
Now try it with v6 addresses on a triple column lobby,

![image](https://user-images.githubusercontent.com/17962918/91168640-e3a2c180-e6cd-11ea-9210-387acf711535.png)

[...]

PR #4222, opened 2020-07-23
Tweaked layout of interface list widget, see issue 4217.

gitwer:
Took option 4 from https://github.com/opnsense/core/issues/4217. Easy to change if needed...


gitwer:
doesn't work with well with IPv6 addresses :(

PR #4225, opened 2020-07-26
add custom jinja filter "regex_replace"

fraenki:
This adds a new custom filter for the jinja template engine. It makes it possible to do something like this:

```
{{ object.name | regex_replace ("[^A-Za-z0-9]","") }}
```

This will be used in https://github.com/opnsense/plugins/issues/1796, but I'm pretty sure it will prove to be useful in other situations too. FWIW, both Ansible and Salt have their own implementation of this custom filter for their jinja template system.

### DIFF
--- a/src/opnsense/service/modules/template.py
+++ b/src/opnsense/service/modules/template.py
@@ -39,6 +39,7 @@ import traceback
 import copy
 import codecs
 import jinja2
+import re
 from .addons import template_helpers
 
 __author__ = 'Ad Schellevis'
@@ -62,6 +63,7 @@ class Template(object):
         # register additional filters
         self._j2_env.filters['decode_idna'] = lambda x:x.decode('idna')
         self._j2_env.filters['encode_idna'] = self._encode_idna
+        self._j2_env.filters['regex_replace'] = self._regex_replace
 
     @staticmethod
     def _encode_idna(x):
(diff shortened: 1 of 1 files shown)


AdSchellevis:
could be simplified a bit with a lambda, but other than that I'm ok with adding a regex replace option. I'll take a look when I have a bit of time.

fraenki:
@AdSchellevis Do you think we could merge this in time for 20.7.2? I'd love to release a new version of os-haproxy, but it depends on this feature.

AdSchellevis:
@fraenki I still had this on my list somewhere, but it was a bit too busy to pick it up. My plan was to use a lambda in stead of a one line function, https://github.com/opnsense/core/commit/71e1393b73a2cac5433e541d7c3c8a9b974d0d65 should do the same as yours. @fichtner usually handles the merging part, I don't mind adding this if he doesn't

fraenki:
@AdSchellevis I've tested https://github.com/opnsense/core/commit/71e1393b73a2cac5433e541d7c3c8a9b974d0d65 and it works as expected, thank you!

fraenki:
@fichtner Please merge https://github.com/opnsense/core/commit/71e1393b73a2cac5433e541d7c3c8a9b974d0d65 to stable/20.7.

fichtner:
@fraenki only if @AdSchellevis says ok....

AdSchellevis:
@fichtner ok :)

fichtner:
done :)

PR #4234, opened 2020-08-01
Display remote host in IPsec Status Overview

garlic17:
Add a new 'Remote Host' column to the IPsec status overview page.


AdSchellevis:
@garlic17 looks good, thanks

Issue #4240, opened 2020-08-03
system: bump backup encryption standards


fichtner:
Since a while we have headers that store required information on decrypting config.xml backups. In this next phase we need to bump the used standards but let decryption decide on the basis of the headers how the file is going to be decrypted.

This could be done backwards-compatible (prior to header user) or break compatibility. Currently undecided as the impact for the latter is likely quite low.

For reference the headers were added in https://github.com/opnsense/core/issues/2535 and it looks like we will not support importing old backups prior to 18.7 which is fine because these old backups are not kept for forensic purposes...

PR #4265, opened 2020-08-15
system: bump backup encryption standards (#4240)

OliverO2:
This PR provides encryption of configuration files with up-to-date security standards (SHA512, PBKDF2, 100000 iterations).

As current and future standards require information than is currently present in the file header, this code changes the header format, using a new preamble field `Encryption: M1` specifying the encryption method/version used. The header format is extensible for future changes in cryptography and fully backward compatible when decrypting configuration files with the current header format. It also optionally provides backward-compatible encryption (currently unused).

This change may be incompatible with the file format used before commit 94b2df5517f8ed7b3af44c9dd984d88e721798e8, which might be no longer relevant today due to its age.

### DIFF
--- a/src/opnsense/mvc/app/library/OPNsense/Backup/Base.php
+++ b/src/opnsense/mvc/app/library/OPNsense/Backup/Base.php
@@ -36,20 +36,29 @@ namespace OPNsense\Backup;
 abstract class Base
 {
     /**
+     * @var string[] openssl arguments by encryption method
+     */
+    private $openssl_arguments = [
+        "M0" => "-aes-256-cbc -md md5", /* backwards-compatible default */
+        "M1" => "-aes-256-cbc -pbkdf2 -iter 100000 -md SHA512",
+    ];
+
+    /**
      * encrypt+encode base64
      * @param string $data to encrypt
      * @param string $pass passphrase to use
      * @param string $tag
+     * @param int $encryption_method
      * @return string base64 encoded crypted data
      */
-    public function encrypt($data, $pass, $tag = 'config.xml')
+    public function encrypt($data, $pass, $tag = 'config.xml', $encryption_method = "M1")
     {
         $file = tempnam(sys_get_temp_dir(), 'php-encrypt');
         @unlink("{$file}.enc");
 
         file_put_contents("{$file}.dec", $data);
         exec(sprintf(
-            '/usr/local/bin/openssl enc -e -aes-256-cbc -md md5 -in %s -out %s -pass pass:%s',
+            '/usr/local/bin/openssl enc -e ' . $this->openssl_arguments[$encryption_method] . ' -in %s -out %s -pass pass:%s',
             escapeshellarg("{$file}.dec"),
             escapeshellarg("{$file}.enc"),
             escapeshellarg($pass)
@@ -60,8 +69,12 @@ abstract class Base
             $version = trim(shell_exec('opnsense-version -Nv'));
             $result = "---- BEGIN {$tag} ----\n";
             $result .= "Version: {$version}\n";
-            $result .= "Cipher: AES-256-CBC\n";
-            $result .= "Hash: MD5\n\n";
+            if ($encryption_method === "M0") {
+                $result .= "Cipher: AES-256-CBC\n";
+                $result .= "Hash: MD5\n\n";
+            } else {
+                $result .= "Encryption: $encryption_method\n\n";
+            }
             $result .= chunk_split(base64_encode(file_get_contents("{$file}.enc")), 76, "\n");
             $result .= "---- END {$tag} ----\n";
             @unlink("{$file}.enc");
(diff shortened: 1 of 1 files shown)


src/opnsense/mvc/app/library/OPNsense/Backup/Base.php
```diff
@@ -60,8 +69,12 @@ public function encrypt($data, $pass, $tag = 'config.xml')
             $version = trim(shell_exec('opnsense-version -Nv'));
             $result = "---- BEGIN {$tag} ----\n";
             $result .= "Version: {$version}\n";
-            $result .= "Cipher: AES-256-CBC\n";
-            $result .= "Hash: MD5\n\n";
+            if ($encryption_method === "M0") {
+                $result .= "Cipher: AES-256-CBC\n";
```
fichtner, line 73: Please don’t remove this old future proof code. It was paced there to easily upgrade security at a later date which is now. ;)

src/opnsense/mvc/app/library/OPNsense/Backup/Base.php
```diff
@@ -35,21 +35,30 @@
  */
 abstract class Base
 {
+    /**
+     * @var string[] openssl arguments by encryption method
+     */
+    private $openssl_arguments = [
+        "M0" => "-aes-256-cbc -md md5", /* backwards-compatible default */
+        "M1" => "-aes-256-cbc -pbkdf2 -iter 100000 -md SHA512",
```
fichtner, line 43: CBC is still a good idea? We could just bump hash and iterate from there instead of changing all at once.

src/opnsense/mvc/app/library/OPNsense/Backup/Base.php
```diff
@@ -85,10 +98,21 @@ public function decrypt($data, $pass, $tag = 'config.xml')
...
         $data = explode("\n", $data);
+        $encryption_method = "M0"; /* default for backward compatibility */
 
         foreach ($data as $key => $val) {
-            /* XXX remove helper lines for now */
-            if (strpos($val, ':') !== false) {
+            if (strpos($val, "Encryption: ") === 0) {
+                /* honor encryption preamble field */
+                $encryption_method = explode(" ", $val, 2)[1];
+                if (!array_key_exists($encryption_method, $this->openssl_arguments)) {
```
fichtner, line 107: Security is ok by not being able to inject command line arguments but matching via magic constant is a bit counter-intuitive in general.

OliverO2: Is it? Think of version numbers. As it's impossible to know how many and which parameters we might need in the future, using just a version-number like method scheme `M0`, `M1`, ... seems quite natural to me. The map helps to keep the code clean and simple, easy to extend and pretty self-documenting.

src/opnsense/mvc/app/library/OPNsense/Backup/Base.php
OliverO2: To me it looks we could stick with CBC here. AFAIK the only CBC vulnerability is the [padding oracle attack](https://blog.cloudflare.com/padding-oracles-and-the-decline-of-cbc-mode-ciphersuites/), which does not seem to be relevant here as it requires an attacker to inject plain text.

src/opnsense/mvc/app/library/OPNsense/Backup/Base.php
OliverO2: So get rid of this completely?

src/opnsense/mvc/app/library/OPNsense/Backup/Base.php
OliverO2: Ah, and AES(256) (CBC) is still listed as a [recommended strong cipher suite on bettercrypto.org](https://bettercrypto.org/#_configuration_a_strong_ciphers_fewer_clients).

src/opnsense/mvc/app/library/OPNsense/Backup/Base.php
OliverO2: Whats your opinion on using these labels as an alternative to the simple version-like scheme?
* `AES-256-CBC/MD5` instead of `M0`
* `AES-256-CBC/PBKDF2*100000/SHA512` instead of `M1`

OliverO2:
@fichtner Can we proceed on this one?

fichtner:
No, I have code quality concerns and no time to fix this myself at the moment.

OliverO2:
Wow! This sounds remarkably destructive to me ("not invented here, contributions not welcome").

The preceding comments seem like bikeshedding about issues of taste. Why this would be preferable over fixing an actual quality issue of weak encryption in a security product remains a mystery.

fichtner:
Let's not get ahead of ourselves. 02d05f7278c is tedious enough as it is. Please fix the containing bug.

Issue #4257, opened 2020-08-14
IPS: Suricata Complete network "freeze"


rudiservo:
I have Suricata working from the previous version without a problem, but on 20.7 it suddenly it crashes and bogs down the internet and all it's access.
I have to restart both OpnSense dedicated box and the ISP router (that is on bridge mode).

Suricata is bound to WAN only.

I had to turn off suricata and it's stable for now, but I would rather have it on... is this anywhere related with netmap?

WAN interface (wan, igb0) Intel Gigabit.
Rest of interfaces are Vlans on a 10Gbase Intel (ix0)

OPNsense 20.7-amd64
FreeBSD 12.1-RELEASE-p7-HBSD
OpenSSL 1.1.1g  21 Apr 2020

AdSchellevis:
did you turn  off hardware vlan offloading? (https://docs.opnsense.org/troubleshooting/network.html)

rudiservo:
@AdSchellevis  Vlan Hardware Filtering actually no it was the only thing enabled... but WAN does not have a VLAN does that matter and should it affect?
Going to try with it disabled but if you know why it affects Suricata and/or netmap please explain, I am a curious being!

AdSchellevis:
yes, netmap doesn't work when it's enabled on the interface using it. in the previous version it didn't matter, which might indicate that the hardware offloading just didn't work in the driver.

rudiservo:
Right... so it might bog down the interfaces and make it crash, completely unreachable?

AdSchellevis:
it will start working the moment you stop suricata... it's not dead, just capturing packets and sending them to nowhere...

rudiservo:
Ohh... that expains alot... thanks

tomatotoast:
I experience similar problems with Suricata.
igb NIC with LACP Parent and VLANs connected.
The traffic to the VLANs is very slow. Interface filtering is disabled.

OliverO2:
@AdSchellevis 
As people still run into this issue, I had been asked by a forum member a while ago to propose a usability improvement: https://forum.opnsense.org/index.php?topic=13600.msg62833#msg62833

Is this something you'd like to pursue? Anything I could do to help here?

AdSchellevis:
[...]

PR #4271, opened 2020-08-17
Improve IPS mode help to avoid blocking network access, issue #4257

OliverO2:
Current help texts do not contain complete information to prevent users blocking network access if IPS mode **and** hardware accelerations are enabled. See issue discussion and [this posting plus follow-ups](https://forum.opnsense.org/index.php?topic=13600.msg62833#msg62833).

This PR aims to include information for all relevant settings.

### DIFF
--- a/src/opnsense/mvc/app/controllers/OPNsense/IDS/forms/generalSettings.xml
+++ b/src/opnsense/mvc/app/controllers/OPNsense/IDS/forms/generalSettings.xml
@@ -9,7 +9,7 @@
         <id>ids.general.ips</id>
         <label>IPS mode</label>
         <type>checkbox</type>
-        <help><![CDATA[Enable protection mode (block traffic).<br />Before enabling, please disable all hardware offloading first <a href="/system_advanced_network.php">in advanced network</a>.]]></help>
+        <help><![CDATA[Enable protection mode (block traffic).<br />Before enabling, please disable all options under <strong>Hardware Offloading</strong> in <a href="/system_advanced_network.php">Interfaces – Settings</a>.]]></help>
     </field>
     <field>
         <id>ids.general.promisc</id>
--- a/src/opnsense/mvc/app/controllers/OPNsense/IDS/forms/generalSettings.xml
+++ b/src/opnsense/mvc/app/controllers/OPNsense/IDS/forms/generalSettings.xml
@@ -9,7 +9,7 @@
         <id>ids.general.ips</id>
         <label>IPS mode</label>
         <type>checkbox</type>
-        <help><![CDATA[Enable protection mode (block traffic).<br />Before enabling, please disable all hardware offloading first <a href="/system_advanced_network.php">in advanced network</a>.]]></help>
+        <help><![CDATA[Enable protection mode (block traffic).<br />Before enabling, please disable all hardware offloading and disable VLAN hardware filtering first in <a href="/system_advanced_network.php">Interfaces – Settings</a>.]]></help>
     </field>
     <field>
         <id>ids.general.promisc</id>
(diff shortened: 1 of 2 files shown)


AdSchellevis:
I don't think we should repeat the same message everywhere on the system_advanced_network.php  page, a more elegant fix would probably be to move all hardware toggles into their own section (Hardware offloading) so it matches the rest of the warnings all around. Which would keep "ARP Handling" in its own section, which probably should get a more logical name in that case as well.

OliverO2:
Fair enough. This PR contains just the minimum viable set of changes which would prevent people from falling into that freezing trap. Doing as you suggested is even better but would require someone more familiar with the UI framework than me.

If you'd prefer to go the latter route, would you consider switching the checkbox logic from negative (`Disable ...`) to positive language (`Enable ...`)? This would avoid double negatives down the line, e.g. a tutorial stating "_Unchecking 'Disable ...' may freeze your network_".

AdSchellevis:
can we please keep this simple? the page itself is a legacy php page, which shouldn't be too hard to refactor.  Changing the language may sound nice, but has the disadvantage of constantly flipping the underlaying toggle (which is called disable...)

OliverO2:
I have been trying to come up with an improved introductory help text and more precise help texts for all of the options. I found it hard to get to an acceptable quality level when sticking to the current language.

I have also found that the _Disable ..._ settings do more than they say: When turned off, they actively _enable_ the respective interface option (instead of leaving it at its default). Only the _VLAN Hardware Filtering_ has a dedicated `Leave default`, which avoids touching the respective interface options.

I remember having been hit hard by falling into the _VLAN Hardware Filtering_ trap, losing network connectivity, not being able to access the router's web or shell interface, not being able to attach a monitor as the HDMI port would only work with a monitor attached at boot time... So I'd be willing to put in some extra effort if this would improve the user experience but I'm not sure if that would be consistent with your priorities:

Would these changes be acceptable to you if I'd clean up the underlying toggles as well? The code in `interfaces.lib.inc` is already pretty hard to read (`!isset(... ['disablevlanhwfilter'])`). Is there a precedent for dealing with backward compatibility resulting from changes in configuration files?

There are `gettext()` calls on the PHP page and I'm not aware of whether translations are used or not. Can these safely be ignored?

My proposal would be to split the _Network Interfaces_ section into two like this (settings at their current defaults):

----

#### Hardware Offloading

> The following options are disabled by default because some NICs and/or drivers are broken. Enabling a hardware offloading option may completely block network connectivity when IPS mode is enabled in _Services – Intrusion Detection – Administration_.

* Hardware CRC offloading: `[ ]`
    > Choose whether to process checksums in hardware. Note that when this is enabled, a packet capture will see empty (all zero) or flag incorrect packet checksums. This is expected behavior with hardware checksum handling.

* Hardware TSO: `[ ]`
    > Choose whether to process outgoing TCP segmentation in hardware, also known as TCP segmentation offloading (TSO, TSO4, TSO6).

* Hardware LRO: `[ ]`
    > Choose whether to process incoming TCP segmentation in hardware, also known as large receive offloading (LRO).

* VLAN Hardware Processing: `Disable`
    > Choose whether to process VLAN tagging, filtering, outgoing TCP segmentation (TSO) and checksums in hardware. Setting this to 'Default' means that each NIC will retain its own default setting.

#### Logging

* ARP: `[x]` Log ARP address changes
    > Choose whether to make a log entry in the main system log when an IP address moves to a different MAC address.

AdSchellevis:
Since we don't offer config migrations on legacy code, we really try to prevent changes in there. The format itself looks workable from my point of view.

OliverO2:
@AdSchellevis Implemented as suggested.

AdSchellevis:
@OliverO2 The new layout will likely prevent some users from impractical misconfigurations, in this case it won't hurt to add an extra warning  (in most cases I'm not a fan of adding a lot of additional texts, but this particular issue seems to warrant it).  thanks!

OliverO2:
@AdSchellevis Yes, hopefully this will help as intended. One trap still lingering is the IPS settings, as the warning appears in the (initially hidden) help text only.

PR #4279, opened 2020-08-21
Add Traffic Filter for Private IPv4 Networks

kcaj-burr:
Add traffic filter for private IPv4 networks.

This is useful if OPNsense in sitting in-front of a layer 3 switch and the networks are defined in the layer 3 switch and not in OPNsense.

### DIFF
--- a/src/www/status_graph.php
+++ b/src/www/status_graph.php
@@ -149,9 +149,6 @@ include("head.inc");
     margin: 0px;
     margin-bottom: -5px;
 }
-#filter {
-    max-width: none;
-}
 </style>
 <script>
     var graphtable = {};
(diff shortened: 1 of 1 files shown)


src/www/status_graph.php
```diff
@@ -348,6 +362,9 @@ function update_bandwidth_stats() {
                         <option value="local" <?=$pconfig['filter'] == "local" ? " selected=\"selected\"" : "";?>>
                           <?= gettext('Local') ?>
                         </option>
+                        <option value="private" <?=$pconfig['filter'] == "private" ? " selected=\"selected\"" : "";?>>
```
AdSchellevis, line 365: The  `selected=\"selected\""` could go  (on the existing options too), they can't be all selected at the same time so it will pick the first one anyway.

src/www/status_graph.php
```diff
@@ -348,6 +362,9 @@ function update_bandwidth_stats() {
...
+                          <?= gettext('Private') ?>
```
AdSchellevis, line 366: how about "RFC1918 Private Networks"? Private leaves room for interpretation.

AdSchellevis (changes requested):
Thanks for your contribution.

Adding rfc1918 specifically as an option is fine by me, just some small remarks about the change. Eventually we should at least replace the direct call to rate with a wrapper via configd, but functionally this is ok.

kcaj-burr:
Thanks for the comments @AdSchellevis. I am not familiar yet with configd or the complete OPNsense development routine so please bear with me.

I have adjusted as per your comments.

I did have to add a style to allow the filter `select` input to fill its `<td>`, though its just a personal preference on the visuals

AdSchellevis:
@kcaj-burr oops, I think I've misread something, don't reply before coffee ;) can you revert to the previous situation and only change the "private" label?

AdSchellevis:
@kcaj-burr thanks! for inspiration on how to develop new mvc code, just visit https://docs.opnsense.org/develop.html recently I've moved the gateway status widget over to use an api endpoint in https://github.com/opnsense/core/commit/89cfc06d8eb5e0ee1060ba950e89ff66f745c81f

PR #4288, opened 2020-08-23
Updated Project Block Lists Links

gap579137:
These links were updated to reflect the current hosting platform that allows for last-modified header.


AdSchellevis:
As discussed via email, we should use the GitHub locations as primary source (https://github.com/blocklistproject/Lists). Thanks @gap579137 !  

@mimugmail dnscrypt-proxy and Bind seem to be using the other locations as well,  do you want me to replace them?

gap579137:
@AdSchellevis @mimugmail, I created a pull request with proposed changes to the dnscrypt-proxy and Bind scripts

PR #4300, opened 2020-08-26
Fix missing data in dashboard temperature widget.

NigelWhatling:
In certain circumstances (doesn't seem to happen for all), temperature widget on the dashboard does not show temperature values. Temperature readings correctly show in other logging areas.

Cause is grep in temperature plugin API returning extraneous rows (web request logging) that are not device readings. This, in turn, causes an error in the JQuery call on the widget.

Change is to tighten grep selection in the API to only return the appropriate rows.

See [No temperature data on Thermal Sensor widget after upgrade to 20.7.1](https://forum.opnsense.org/index.php?topic=18626).

### DIFF
--- a/src/www/widgets/api/plugins/temperature.inc
+++ b/src/www/widgets/api/plugins/temperature.inc
@@ -32,7 +32,7 @@
 function temperature_api()
 {
     $result = array();
-    exec('/sbin/sysctl -a | grep -E "(dev|hw)[^[:space:]]+temperature:"', $sysctlOutput);
+    exec('/sbin/sysctl -a | grep -E "(dev.cpu|hw.acpi.thermal)\.[^[:space:]]+temperature:"', $sysctlOutput);
     foreach ($sysctlOutput as $sysctl) {
         $parts = explode(':', $sysctl);
         if (count($parts) >= 2) {
--- a/src/www/widgets/api/plugins/temperature.inc
+++ b/src/www/widgets/api/plugins/temperature.inc
@@ -32,7 +32,7 @@
 function temperature_api()
 {
     $result = array();
-    exec("/sbin/sysctl -a | grep temperature", $sysctlOutput);
+    exec("/sbin/sysctl -a | grep temperature:", $sysctlOutput);
     foreach ($sysctlOutput as $sysctl) {
         $parts = explode(':', $sysctl);
         if (count($parts) >= 2) {
--- a/src/www/widgets/api/plugins/temperature.inc
+++ b/src/www/widgets/api/plugins/temperature.inc
@@ -32,7 +32,7 @@
 function temperature_api()
 {
     $result = array();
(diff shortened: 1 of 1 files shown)


fichtner (commented):
I don’t think his is necessary as it does nothing to fix the underlying unrelated issue that should already be fixed once 20.7.2 is out.

Is syslog-no running? Are logging files populated?

NigelWhatling:
What fix is coming out in 20.7.2? syslog-ng (assume that's what you meant) is running. Which logging files am I looking for?

fichtner:
Try to restart syslog-lg once and see if the issue disappears:

    # service syslog-lg restart

Also make sure the log files all have updated (e.g. firewall love log).

NigelWhatling:
> Try to restart syslog-lg once and see if the issue disappears: # service syslog-lg restart Also make sure the log files all have updated (e.g. firewall love log).

I restarted syslog-ng but it did not make any difference. Are you sure you're thinking about missing _temperature_ data? I know there is another existing issue with missing WAN _traffic_ data when IDS is enabled.

The fix you're suggesting would have to do something like change/disable the kern.msgbuf value returned by sysctl. Is that right?

fichtner:
Not entirely, these messages should never appear where they appear. Fixing the temperature widget is one thing by muting the output, but we have to assume this affects other areas as well so we need to find the cause why wrong syslog messages appear in sysctl. This is either FreeBSD 12.1 or syslog(-ng) change related as the issue was never there in 20.1 and below.

Assuming the message is stuck there in the buffer maybe it can’t be easily undone and this particular fix here may be appealing. But it’s worth going further.


Cheers,
Franco

NigelWhatling:
Agree that it's worth looking into further and not forgotten. I don't know why these logs are appearing in kern.msgbuf.

This fix may not resolve the underlying issue, but it does resolve a problem with the way the temperature API collects data. Changing the grep to use regular expressions might be even better, but there might be a slight performance hit with that.

fichtner:
If this should receive specific improvement it should use some sort of strategy for reading only keys and then fetching values fir relevant keys to avoid a potential mismatch in the value part. This might also require changing RRD temperature fetch as well.

marjohn56:
Or take a look at #4304, we only ask for what we want; also reduces the amount of parsing.

NigelWhatling:
Made the search even more precise. Have this running on my box and it's working well. I guess I'll just leave this here.

If I'm not mistaken, RRD fetches the value for CPU temp 0 directly:

```
/* the CPU Temperature gathering function */
$rrdupdatesh .= "CPUTEMP=`$sysctl -n dev.cpu.0.temperature | sed 's/C//g'`\n";
$rrdupdatesh .= "$rrdtool update $rrddbpath$ifname$cputemp N:\${CPUTEMP}\n";
/* end CPU Temp gathering */
```

NigelWhatling:
So ends my first 'contribution' to OSS. Closing in favour of #4306.

PR #4301, opened 2020-08-27
Update rulecache.py

kulikov-a:
catch exception in rule metadata parsing

### DIFF
--- a/src/opnsense/scripts/suricata/lib/rulecache.py
+++ b/src/opnsense/scripts/suricata/lib/rulecache.py
@@ -105,9 +105,17 @@ class RuleCache(object):
                             prop = section[0:sep].strip()
                             value = section[sep+1:].strip(' "')
                             if prop == 'metadata':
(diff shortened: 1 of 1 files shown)


AdSchellevis:
we're not going to catch and silence an exception like this, best open a ticket containing the test case that fails so we can add the proper condition in there. probably all it needs is one line like:
`
record['metadata'][parts[0]] = parts[1] if len(parts) > 1 else None
`

kulikov-a:
ok. thank you. 
"record['metadata'][parts[0]] = parts[1] if len(parts) > 1 else None"
can set very long string in rule details metadata label. 
ticket: 
https://github.com/opnsense/plugins/issues/2005
will post bad formatted PTReasearch rule strings there
Thanks!

AdSchellevis:
ok, better close the issue there, use our template for a bug report in core, I'll take a look.

kulikov-a:
ok, thanks
https://github.com/opnsense/core/issues/4302
closing issue in plugins

PR #4304, opened 2020-08-27
Be specific when using sysctl for temperature calls

marjohn56:
This makes the calls for getting cpu and zone temperatures more specific, thus avoiding junk being returned in the sysctl response

### DIFF
--- a/src/www/widgets/api/plugins/temperature.inc
+++ b/src/www/widgets/api/plugins/temperature.inc
@@ -31,7 +31,7 @@
 function temperature_api()
 {
     $result = array();
-    exec("/sbin/sysctl -a | grep temperature", $sysctlOutput);
+    exec("/sbin/sysctl dev.cpu | grep temperature", $sysctlOutput);
     foreach ($sysctlOutput as $sysctl) {
         $parts = explode(':', $sysctl);
         if (count($parts) >= 2) {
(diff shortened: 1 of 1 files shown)


NigelWhatling:
You're making to system calls instead of one. Surely that's not the most performant way to do this?

Also, since you're capturing specifically, you can simplify this:
```
$tempItem['type'] = strpos($tempItem['device'], 'hw.acpi') !== false ? "zone" : "core";
$tempItem['type_translated'] = $tempItem['type'] == "zone" ? gettext("Zone") : gettext("Core");
```

marjohn56:
Good idea. Strangely enough, the dual call and process seems to run faster on my old APU2 than the original single call, possibly due to the reduction in the amount of data being returned.

NigelWhatling:
> Strangely enough, the dual call and process seems to run faster on my old APU2 than the original single call, possibly due to the reduction in the amount of data being returned.

Really? I would have expected the overhead of the additional calls to be greater.   ¯\_(ツ)_/¯

NigelWhatling:
You could always do this, I guess:

`exec("/sbin/sysctl dev.cpu hw.acpi.thermal | grep temperature", $sysctlOutput);`

marjohn56:
Hence the 'strangely'!

Yes, there's many ways to skin this cat... just something that's reliable and won't require a re-visit anytime is what's the ultimate aim. Ideas have been thrown in. In the end it's up to @fichtner & @AdSchellevis to make the decisions.

fichtner:
can we agree the grep on the key+value is unsafe and best try to avoid it?

NigelWhatling:
> can we agree the grep on the key+value is unsafe and best try to avoid it?

Not sure what a simple alternative to this would be.

marjohn56:
Hmm.. so you suggest iterating through the cores until we get a null response?

marjohn56:
Playing about I have come up with two options, both work and should fit the bill.

 **exec("/sbin/sysctl dev.cpu |egrep -E 'cpu\.[0-9]+\.temp'",$sysctlOutput1);
 **exec("/sbin/sysctl hw.acpi.thermal |egrep -E 'tz[0-9]+\.temp'",$sysctlOutput2);
  **$sysctlOutput = array_merge($sysctlOutput1,$sysctlOutput2);**
    
or

    **exec("sysctl -a | egrep -E 'cpu\.[0-9]+\.temp|tz[0-9]+\.temp'",$sysctlOutput);

NigelWhatling:
`exec("sysctl dev.cpu hw.acpi.thermal | grep -E 'cpu.[0-9]+.temperature|tz[0-9]+.temperature'", $sysctlOutput);`

marjohn56:
even better... hadn't thought of that.

marjohn56:
Closing this in favour of #4306

PR #4306, opened 2020-08-27
Be even more specific in gettting the temperature data

marjohn56:
Only ask for specific data and then filter that response too.

### DIFF
--- a/src/www/widgets/api/plugins/temperature.inc
+++ b/src/www/widgets/api/plugins/temperature.inc
@@ -32,7 +32,8 @@
 function temperature_api()
 {
     $result = array();
-    exec("/sbin/sysctl -a | grep temperature", $sysctlOutput);
+    exec("sysctl sysctl dev.cpu hw.acpi.thermal | egrep -E 'cpu\.[0-9]+\.temp|tz[0-9]+\.temp'",$sysctlOutput);
+
     foreach ($sysctlOutput as $sysctl) {
         $parts = explode(':', $sysctl);
         if (count($parts) >= 2) {


src/www/widgets/api/plugins/temperature.inc
```diff
@@ -32,7 +32,8 @@
 function temperature_api()
 {
     $result = array();
-    exec("/sbin/sysctl -a | grep temperature", $sysctlOutput);
+    exec("sysctl sysctl dev.cpu hw.acpi.thermal | egrep -E 'cpu\.[0-9]+\.temp|tz[0-9]+\.temp'",$sysctlOutput);
```
NigelWhatling, line 35: This should use grep, not egrep.

marjohn56: Why?

NigelWhatling: Apart from GNU comment? `egrep` is the equivalent of `grep -E`. `egrep -E` is redundant. I'd just go with `grep`. Suppose it doesn't really matter.

marjohn56:
It's old school and obsolete... like me. 😉

fichtner:
Thanks for both your work on this topic. We had a larger discussion internally on how to proceed and decided to go ahead with 305d6b35ccfd9feb28ac1dfa4f9ac9b2e30fd31b addressing most of the issues, especially reading only the variable name to avoid pulling in invalid data in the first place which seemed to be the original issue.

There is a bit of doubt about how "temperature" info is collected from the sysctl variables to avoid hardcoding parent nodes. It's unlikely changed from FreeBSD side (new nodes with different names or other purposes) but it would be beneficial to pass through unknown OIDs that seem to fit the pattern and present them to the widget/caller and deal with them in a similar fashion like we did with the reports here.

A side effect of the patch is the easy use of

    # configctl system temp

to return all current temperature readings.


Cheers,
Franco

PR #4316, opened 2020-09-02
Add optional kill sessions with host after reconfigure alias

Actpohomoc:
optional kill sessions with host_to_kill host in reconfigureAction() function

### DIFF
--- a/src/opnsense/mvc/app/controllers/OPNsense/Firewall/Api/AliasController.php
+++ b/src/opnsense/mvc/app/controllers/OPNsense/Firewall/Api/AliasController.php
@@ -227,10 +227,11 @@ class AliasController extends ApiMutableModelControllerBase
     }
 
     /**
-     * reconfigure aliases
+     * reconfigure aliases (optional kill sessions with host_to_kill host)
      */
-    public function reconfigureAction()
+    public function reconfigureAction($host_to_kill = null)
     {
+		$result = array();
         if ($this->request->isPost()) {
             $backend = new Backend();
             $backend->configdRun('template reload OPNsense/Filter');
@@ -239,10 +240,23 @@ class AliasController extends ApiMutableModelControllerBase
             if (!empty($bckresult['messages'])) {
                 throw new UserException(implode("\n", $bckresult['messages']), gettext("Alias"));
             }
-            return array("status" => "ok");
+            $result = array("status" => "ok");
         } else {
-            return array("status" => "failed");
+            $result = array("status" => "failed");
         }
+		// optional kill sessions with host_to_kill host
+		if (!empty($host_to_kill)) { 
+			$output = array();
+			$command = "/sbin/pfctl -k {$host_to_kill}/32 -k 0/0";
+			exec($command . ' 2>&1', $output);
+			$retval = implode("\n", $output);
+			//
+			$command = "/sbin/pfctl -k 0/0 -k {$host_to_kill}/32";
+			exec($command . ' 2>&1', $output);
+			$retval = implode("\n", $output);
+			$result = array("status" => "{$retval}");
+		}
+		return $result;
     }
 
     /**


AdSchellevis:
mvc api's aren't allowed to access privileged tools (https://docs.opnsense.org/development/architecture.html), other than that this also looks like a fix for a custom solution which doesn't really fit our roadmap.

Actpohomoc:
> mvc api's aren't allowed to access privileged tools (https://docs.opnsense.org/development/architecture.html), other than that this also looks like a fix for a custom solution which doesn't really fit our roadmap.
Any other way to drop current connections after reconfiguring aliases?

AdSchellevis:
I don't think we have an endpoint for it at the moment, it would probably need a rewrite to mcv of diag_dump_states.php since the feature itself lives in that corner.

Actpohomoc:
> I don't think we have an endpoint for it at the moment, it would probably need a rewrite to mcv of diag_dump_states.php since the feature itself lives in that corner.

I took the code from somewhere there. I have found that dialog menu of showing current states it is possible to drop it. 
Maybe better to write separate API function?
I have made this changes because after each update I have to rewrite the controller file.

`mvc api's aren't allowed to access privileged tools` 
But why? This is only function that is needed. 
Is it possible somehow to run separately only this privileged function of dropping current connections with host?

I use API to update aliases. But I can not drop current connections that are already not allowed by new alias. This situation is more dangerous.

I am absolutely sure that this API feature will give huge benefit to end users. Everything now are going to be automatic.
I wrote the Telegram bot  that works like "kill switch". All connection with selected host inside my network can be momentary dropped just by pressing one button.

Please write the code that give this possibility in more secure way. Or give me direction how to do it more correctly. I don't really know php. I use code that already was there.

AdSchellevis:
We spend quite some time documenting (https://docs.opnsense.org/develop.html) how new mvc code is intended, you will find all details and examples needed to fulfil the task there. Rewriting the PR lies outside the scope of community support.

Adding the functionality to this endpoint likely won't happen, since both tasks have no real relation.

Actpohomoc:
> We spend quite some time documenting (https://docs.opnsense.org/develop.html) how new mvc code is intended, you will find all details and examples needed to fulfil the task there. Rewriting the PR lies outside the scope of community support.
> 
> Adding the functionality to this endpoint likely won't happen, since both tasks have no real relation.

I am disagree. I read this manual https://docs.opnsense.org/develop.html very careful and found a lot of missing information there. **I can show you places where there is only name of api function without ANY description**.  It looks like someone had no time at all to write details and samples how to use most of the api function.
I read this https://www.ncartron.org/opnsense-api.html blog to get more practical information how to automate process of working with OPNsense.

As I remember the goal of project  OPNsense compare to pfSense to be more dynamic and live to the community needs. 
But who are the people of community? 
How do you know that this function is not public needed? 
It is needed for me and to anybody who wants to build a bot or mobile app with connection dropping firewall's function.

Now everything going to be automatic. Just search "ansible + opnsense". But API function is entry point of an y automation.
We have possibilities to change aliases - this is super importer function! 

**But without dropping current connection is almost useless!**

Everybody who trying to use API to change alias will have this problem!
I see this as **security issue!** 

Look from the point of access inside the network! 
I want to change access as quickly as possible and I do not have this possibility!
I can change the alias but if intruders already there and they continue to be there without any problem.
To drop the connection I have to login to the system and restart the OPNsense!  Nice fast way.

When I searching the subject I have found a lot of requests on forum about how to implement such function. 

Please look to the problem again from security point.

AdSchellevis:
as mentioned, it's not the correct endpoint, the one where the functionality should live doesn't exist. 

Let's agree to disagree on the subject, nobody said the functionality may not exist, shouting of the rooftops that your vision and implementation is the one we should follow usually doesn't work here.

Closing this PR now, since it seems unlikely that it will reach a functional state.

PR #4320, opened 2020-09-03
find_table_references.py: use pfctl for IP check

kulikov-a:
we can use pfctl to check IP against tables
-its exclusions aware
-this should be more accurate because check tables in pf after table loading (with deduplicating, conflict checking etc)


AdSchellevis:
@kulikov-a works like a charm, thanks!

kulikov-a:
very glad, thanks! my first merged PR)

fichtner:
@kulikov-a congrats :)

PR #4325, opened 2020-09-05
Simpllified config restore process

affinityv:


fichtner:
There is nothing to do here. The simplification ignores that we do not want to import things we cannot work with so we export only explicit things we know there are backups of.

affinityv:
If the user had things in their /conf/ directory, why not let them keep them all?
- it lessens the risk of losing files and if it copies more than is needed, then it won't likely cause any harm.

fichtner:
Please let this go. I don't want to retest this basic functionality for the sake of refactoring.

affinityv:
Okay, but can you please accept that there is at least the following error to be fixed...

NB: This is just one output from using shellcheck, ignoring all other outputs:

````
In opnsense-importer line 321:
  for FILE in "$(find /conf/sshd -type f -name '*key')"; do
              ^-- SC2066: Since you double quoted this, it will not word split, and the loop will only run once.
````

If zero files are returned, the chmod will execute with an error.

````sh
# chmod 600
usage: chmod [-fhv] [-R [-H | -L | -P]] mode file ...
````

Using the exec option of find handles odd file naming, such as file names with spaces, which is not an issue with the expected file names in this case.

Would you like me to provide a new pull request which only replaces the broken for-loop with the safer find-exec as follows?

````sh
# Ensure correct permission for keys so that sshd can start successfully
echo -n "Ensure correct permission for sshd keys..."
find /conf/sshd/ -name '*key' -type f -exec chmod -vv -- 600 {} +
echo "done."
````

fichtner:
Fixed the quoting, thanks for the hint.

affinityv:
All good, glad I could help.

affinityv:
FWIW....

````sh
In opnsense-importer line 321:
  for FILE in $(find /conf/sshd -type f -name '*key'); do
              ^-- SC2044: For loops over find output are fragile. Use find -exec or a while read loop.
````

Issue #4328, opened 2020-09-06
radvd: Option to set DeprecatePrefix to off


vlcty:
**Important notices**
Before you add a new report, we ask you kindly to acknowledge the following:

[x] I have read the contributing guide lines at https://github.com/opnsense/core/blob/master/CONTRIBUTING.md

[x] I have searched the existing issues and I'm convinced that mine is new.

**Is your feature request related to a problem? Please describe.**

When I update OPNsense and reboot, radvd sends out a router advertisement setting preferred lifetime to 0 causing all my clients to deprecate the prefix. This is intended behavour, because radvds DeprecatePrefix is set to on in the config.

The config generated by OPNsense:

```
interface vtnet1 {
	AdvSendAdvert on;
	MinRtrAdvInterval 200;
	MaxRtrAdvInterval 600;
	AdvLinkMTU 1500;
	AdvDefaultPreference high;
	prefix 2001:db8::/64 {
		DeprecatePrefix on;
		AdvOnLink on;
		AdvAutonomous on;
	};
	RDNSS 2001:db8::1/32 {
	};
	DNSSL mydomain.de {
	};
};
```

Excerpt from the radvd config man page:

          DeprecatePrefix on|off

	      Upon shutdown, this option will cause  radvd  to	deprecate  the
	      prefix  by  announcing  it  in the radvd shutdown	RA with	a zero
	      preferred	lifetime and a valid lifetime slightly greater than  2
	      hours. This will encourage end-nodes using this prefix to	depre-
	      cate any associated addresses immediately. Note that this	option
	      should  only be used when	only one router	is announcing the pre-
	      fix onto the link, otherwise end-nodes will deprecate associated
	      addresses	 despite  the  prefix  still being valid for preferred
	      use.

	      See RFC4862, section 5.5.3., "Router Advertisement  Processing",
	      part (e).

	      Default: off

I understand that this option is on by default and I'd suggest leaving it that way. It makes the switch to a new prefix when using dynamic prefixes painless.   
I've got a static prefix from my ISP so my LAN prefix never changes. When OPNsense reboots I loose internet access (that's expected) but I also loose LAN connectivity due to the deprecated prefix. I need to fall back to the link local addresses.

**Describe the solution you'd like**

I'd like to have a checkbox in the OPNsense Router Advertisements GUI where I can toggle `DeprecatePrefix` to on or off.

**Describe alternatives you've considered**

Static IPv6 addresses for everything.

**Additional context**

[...]

PR #4329, opened 2020-09-06
radvd: Added option to set DeprecatePrefix to off

vlcty:
I added an option to radvd to disable prefix deprecation. This refers to issue #4328 .

Bildschirmfoto 2020-09-06 um 20 27 20

### DIFF
--- a/src/etc/inc/plugins.inc.d/dhcpd.inc
+++ b/src/etc/inc/plugins.inc.d/dhcpd.inc
@@ -327,7 +327,13 @@ function dhcpd_radvd_configure($verbose = false, $blacklist = array())
 
         foreach ($stanzas as $stanza) {
             $radvdconf .= "\tprefix {$stanza} {\n";
-            $radvdconf .= "\t\tDeprecatePrefix " . (!empty($dhcpv6ifconf['rainterface']) ? "off" : "on") . ";\n";
+
+            if (!empty($dhcpv6ifconf['radontdeprecateprefix'])) {
+                $radvdconf .= "\t\tDeprecatePrefix off;\n";
+            } else {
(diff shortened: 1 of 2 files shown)


AdSchellevis:
There's probably no perfect fix for this, but since DeprecatePrefix is influenced by more toggles, the meaning can be confusing.

https://github.com/opnsense/core/blob/2cff1c5f12fab71e3baa2a03fd82c1bcfeb7d5b3/src/etc/inc/plugins.inc.d/dhcpd.inc#L330

https://github.com/opnsense/core/blob/2cff1c5f12fab71e3baa2a03fd82c1bcfeb7d5b3/src/etc/inc/plugins.inc.d/dhcpd.inc#L497-L501

https://github.com/opnsense/core/blob/2cff1c5f12fab71e3baa2a03fd82c1bcfeb7d5b3/src/etc/inc/plugins.inc.d/dhcpd.inc#L508-L525

Let's try to collect some feedback before pulling this in. Unfortunately some legacy things are hard to change without risking behavioural differences for other users.

fichtner:
I thought about this... if the user knows the prefix is static it should be able to override the deprecation. The patch needs cleanup but there's no harm providing a manual toggle for those who know what they are doing. But this only counts for manual radvd mode anyway.

fichtner:
As another way around this we could just offer "rainterface" with a dynamic and static mode (currently it is only shown as carp interface list). It already does move everything to "sticky" mode, also the route lifetime...

fichtner:
@vlcty does this work for you? b5b2b213fcc429cd1d3

vlcty:
Yeah that'll also work for me. I think it's actually a more elegant way to solve it.

fichtner:
ok, thanks! I'll leave the ticket open to do some GUI option rewording for rainterface. If you have a suggestion there feel free to create PR :)

PR #4357, opened 2020-09-18
Help text implied the opposite of what the checkbox did. Closes #4354

### LINKED ISSUE #4354: "Reverse token order" help text is incorrect
**Problem**
Under System ‣ Access ‣ Servers ‣ Add the `Local + Timebased One Time Password` option offers an option "Reverse token order" which is described as "Require the password in front of the token instead of behind it." 

Reading this description, it sounds like checking the box will require the token in front of the password and leaving the box unchecked will require the token behind the password. However, it's the opposite.

**Proposed Fix**
I propose changing the text to "Checking this box requires the token after the password. Default requires the token in front."

-----

I assume this problem also affects `LDAP + Timebased One Time Password`, but I don't have an LDAP server handy to test with.

bobpaul:


fichtner:
Merged, thanks!

Issue #4356, opened 2020-09-18
DHCP - Dynamic DNS Registration - No zone for DNS domain of additional pool in dhcpd.conf


FromageBlue:
**Describe the bug**
Configured a DHCPv4 service for an interface via the WebUI (services_dhcp.php) and added a Dynamic DNS configuration that updates a DNS zone (mydomain.com) on an external DNS server.
When adding an additional pool which also has a Dynamic DNS configuration that updates a different zone (qa.mydomain.com) on the same external DNS server, DNS updates are done on mydomain.com and not qa.mydomain.com

In /var/dhcpd/etc/dhcpd.conf, the zone declaration for the dynamic DNS domain associated with the Dynamic DNS configuration of the additional pool (i.e. qa.mydomain.com) does not get created. We should see:
zone qa.mydomain.com. {
primary 10.2.0.100;
key ddns-router;
}
but we don't.
To make this explanation simple, I posted in log section my /var/dhcpd/etc/dhcpd.conf from my opensense.

**To Reproduce**

   1. Go to Services: DHCPv4: [interface]
   2. Configure DHCP.
   3. Enable registration of DHCP client names in DNS to mydomain.com
   4. Configure additional pool with registration of DHCP client names in DNS to qa.mydomain.com

**Expected behavior**
Clients that are part of the additional pool should have there domain name registered to the qa.mydomain.com zone.

**Screenshots**

**Relevant log files**
**_/var/dhcpd/etc/dhcpd.conf_**
option domain-name "mydomain.com";
option ldap-server code 95 = text;
option arch code 93 = unsigned integer 16; # RFC4578
option pac-webui code 252 = text;
option custom-opt5-0 code 121 = string;

default-lease-time 7200;
max-lease-time 86400;
log-facility local7;
one-lease-per-client true;
deny duplicates;
ping-check true;
update-conflict-detection false;
authoritative

subnet 10.2.64.0 netmask 255.255.248.0 {
pool {
option domain-name-servers 10.2.0.100,10.2.0.101;
deny members of "A6AAA4";
ddns-update-style interim;
range 10.2.65.1 10.2.71.250;
}

pool {
option domain-name-servers 10.2.0.100,10.2.0.101;
option domain-name "qa.mydomain.com";
ddns-domainname "qa.mydomain.com";
ddns-update-style interim;
range 10.2.71.251 10.2.71.254;
}

option routers 10.2.64.1;
option domain-name "mydomain.com";
[...]

PR #4370, opened 2020-09-23
Add zone declarations of additional pools in dhcpd.conf - Fix for #4356

FromageBlue:
Relating to #4356, when defining additional pools with DDNS values, the corresponding zone and key declarations were not added to /var/dhcpd/etc/dhcpd.conf
This will add zone and key declarations to dhcpd.conf based on DDNS values provided in each additional pool within a subnet.


AdSchellevis:
@FromageBlue looks good, thanks!

PR #4379, opened 2020-09-27
Update picture.widget.php

opnsenseuser:
I do not think, that the <a> link in the integrated image of the picture widget comprehensible. i changed this from "<a href" to "<img scr"

### DIFF
--- a/src/www/widgets/widgets/picture.widget.php
+++ b/src/www/widgets/widgets/picture.widget.php
@@ -89,7 +89,7 @@ if ($config['widgets']['picturewidget_filename'] != "") :?>
   <div id="picture-widgets" style="padding: 5px">
     <img scr="/widgets/widgets/picture.widget.php?getpic=true" target="_blank">
       <img style="border:0px solid; width:100%; height:100%" src="/widgets/widgets/picture.widget.php?getpic=true" alt="picture" />
-    </a>
+    </img>
   </div>
 <?php
 endif ?>
--- a/src/www/widgets/widgets/picture.widget.php
+++ b/src/www/widgets/widgets/picture.widget.php
@@ -87,9 +87,8 @@ if ($_POST) {
 <?php
 if ($config['widgets']['picturewidget_filename'] != "") :?>
   <div id="picture-widgets" style="padding: 5px">
-    <img scr="/widgets/widgets/picture.widget.php?getpic=true" target="_blank">
-      <img style="border:0px solid; width:100%; height:100%" src="/widgets/widgets/picture.widget.php?getpic=true" alt="picture" />
-    </img>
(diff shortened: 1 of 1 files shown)


AdSchellevis:
this is very old, but I think the original intent was to go to (`<a href`) the image on click, if that doesn't work for some reason we might discuss removing the link, adding a duplicate ``) doesn't really make sense to be honest.

AdSchellevis:
@opnsenseuser this won't be merged for reasons mentioned above, please only open PR's explaining what the intent is an why one should want to incorporate a change.

opnsenseuser:
I only changed the source code to the extent, because I don't see any point in being able to click on an image and it will open in a window as a separate image. So I asked myself the question of whether it made sense to make the picture linked.

AdSchellevis:
in which case you would drop the `<a> </a>` to drop the link, which is probably fine, adding another `` doesn't make sense

src/www/widgets/widgets/picture.widget.php
```diff
@@ -87,9 +87,8 @@
 <?php
 if ($config['widgets']['picturewidget_filename'] != "") :?>
   <div id="picture-widgets" style="padding: 5px">
-    <img scr="/widgets/widgets/picture.widget.php?getpic=true" target="_blank">
-      <img style="border:0px solid; width:100%; height:100%" src="/widgets/widgets/picture.widget.php?getpic=true" alt="picture" />
-    </img>
+    <img scr="/widgets/widgets/picture.widget.php?getpic=true" target="_blank" style="border:0px solid; width:100%; height:100%" src="/widgets/widgets/picture.widget.php?getpic=true" alt="picture" />
+      </img>
```
AdSchellevis, line 91: ^^ already closed

opnsenseuser: OK. I haven't tried it now. but i have now shortened the source code to an img tag. would that be okay?

AdSchellevis: you can leave the img, but don't try to close it again. 

Please, pretty please, check and test code before pushing, I don't mind reviewing code from time to time, but a one-liner fix turns into a multitude of work on my end. Just try to keep it simple, a good starting point is to open an issue (for tracking reasons) explaining your goal(s) and an PR solving the issue. only a PR is also ok as long as it's very clear what's being changed, why it should change and what would be the downside of not doing it.

opnsenseuser: yes it works perfect!

AdSchellevis:
it has a duplicate close tag, the original `` was fine as it was, I'll close this PR as it doesn't seem to be maturing to a valid result.

opnsenseuser:
Ok

Issue #4403, opened 2020-10-08
SSHd fails to start if MAX_LISTEN_SOCKS limit exceeded


leprasmurf:
**Important notices**
Before you add a new report, we ask you kindly to acknowledge the following:

[x] I have read the contributing guide lines at https://github.com/opnsense/core/blob/master/CONTRIBUTING.md

[x] I have searched the existing issues and I'm convinced that mine is new.

**Describe the bug**
SSHd has a built-in limit to listening connections.  When selecting interfaces for sshd to listen on, if 8+ interfaces are selected sshd fails to start.  The MAX_LISTEN_SOCKS value is defined upstream (https://github.com/freebsd/freebsd/blob/master/crypto/openssh/sshd.c#L198).  I don't know if this was working in previous versions.

**To Reproduce**
1. Go to '/system_advanced_admin.php'
2. Under 'Listen Interfaces' choose 8+ interfaces
3. Start sshd
4. Filter logging by 'sshd' at '/ui/diagnostics/log/core/system'

**Expected behavior**
SSHd should start and listen on the selected interfaces or prevent the user from selecting more interfaces than MAX_LISTEN_SOCKS allows.

**Screenshots**
![image](https://user-images.githubusercontent.com/195989/95514737-718af100-098a-11eb-8b40-2286dfc7cf26.png)

**Relevant log files**
```
2020-10-08T13:23:44	sshd[14611]	fatal: Too many listen sockets. Enlarge MAX_LISTEN_SOCKS
2020-10-08T13:23:44	sshd[14611]	fatal: Too many listen sockets. Enlarge MAX_LISTEN_SOCKS
2020-10-08T13:23:44	sshd[14611]	Server listening on 192.168.1.3 port 22.
2020-10-08T13:23:44	sshd[14611]	Server listening on 192.168.1.1 port 22.
2020-10-08T13:23:44	sshd[14611]	Server listening on 192.168.2.3 port 22.
2020-10-08T13:23:44	sshd[14611]	Server listening on 192.168.2.1 port 22.
2020-10-08T13:23:44	sshd[14611]	Server listening on 192.168.1.131 port 22.
2020-10-08T13:23:44	sshd[14611]	Server listening on 192.168.1.129 port 22.
2020-10-08T13:23:44	sshd[14611]	Server listening on 192.168.1.195 port 22.
2020-10-08T13:23:44	sshd[14611]	Server listening on 192.168.1.193 port 22.
[...]

PR #4404, opened 2020-10-08
Secure Shell: Describe sshd built-in limit for listening connections

leprasmurf:
Add help text to inform users of sshd default limitation.

Related issue:  https://github.com/opnsense/core/issues/4403

### DIFF
--- a/src/www/system_advanced_admin.php
+++ b/src/www/system_advanced_admin.php
@@ -627,7 +627,7 @@ $(document).ready(function() {
 <?php endforeach ?>
                   </select>
                   <div class="hidden" data-for="help_for_webguiinterfaces">
-                    <?= gettext('Only accept connections from the selected interfaces. Leave empty to listen globally. Use with care.') ?>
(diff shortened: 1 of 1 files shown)


fichtner:
Hmm, we could replace the default in the config or only write a maximum of 16 addresses. The way the feature is intended is not giving it an extended array of interfaces (maybe 1-2 for management purposes) and I can see that is a very ironic way of breaking user expectation for a feature we always said will never work the way people want it to work due to its operational constraints (add more to allow less). In general, firewall rules are fine and effective for all local services.

leprasmurf:
If I understand you correctly you're suggesting to leave sshd listening on 0.0.0.0 and add a floating rule to allow port 22 on the selected interfaces.  That sounds like it would achieve the desired results.

fichtner:
Exactly. That is the way it worked prior to adding management interface style selector.

In any case we better change the socket limit in the config since we know how many sockets we need and then this issue should be solved and no need to document, ok?

PR #4407, opened 2020-10-11
firewall_nat_edit.php display bug fix

safly:
If interface is not set, it will throw warning in POST page, because $pconfig['ipprotocol'] is not predefined as array like GET.
PHP Warning:  in_array() expects parameter 2 to be array, null given in /usr/local/www/firewall_nat_edit.php on line 542

### DIFF
--- a/src/www/firewall_nat_edit.php
+++ b/src/www/firewall_nat_edit.php
@@ -539,7 +539,7 @@ $( document ).ready(function() {
                       <select name="interface[]" class="selectpicker" data-width="auto" data-live-search="true" multiple="multiple">
 <?php
                         foreach (legacy_config_get_interfaces(array("enable" => true)) as $iface => $ifdetail): ?>
-                        <option value="<?=$iface;?>" <?= in_array($iface, $pconfig['interface']) ? "selected=\"selected\"" : ""; ?>>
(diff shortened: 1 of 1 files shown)


AdSchellevis:
@safly  can you provide the exact steps needed to replicate this? `$pconfig['interface']` should contain an array, due to the initialisation. (it doesn't fail on my end, using latest master)

safly:
@AdSchellevis Sure :)
As I mentioned, it will throw warning in POST page if you leave 'Interface' empty.
edit
bug report
As you said, it was set to an array during initialization, but only for GET.
code
When it comes to POST, it only does verification. Personally the best solution is to avoid errors in the UI display so as not to make the code look complicated (than adding another judgement for it).

AdSchellevis:
@safly I'm not sure why I couldn't reproduce it the first time, but we should indeed prevent warnings from happening. As of php 7 we seem to be able to use the `??` operator (https://www.php.net/manual/en/migration70.new-features.php#migration70.new-features.null-coalesce-op).  The old legacy pages are terrible to validate properly, unfortunately some will stay with us for quite some time. https://github.com/opnsense/core/commit/3dc47c23a34341d08aece95023c53b30b872bb4a should fix the issue. Thanks for reporting.

PR #4418, opened 2020-10-17
Auth/LDAP: Fix reading displayName attribute

ServiusHack:
In 6f76b5f the displayName attribute was added as the first camel case
attribute being read from the search result. As various[^1] comments[^2]
for `ldap_search` mention the attribute names must be lower case, even
though the LDAP server might return them differently.

Using all lower case to access the returned attribute results in the
value of displayName actually being used as the full name of the user.

[^1]: https://www.php.net/manual/en/function.ldap-search.php#37317
[^2]: https://www.php.net/manual/en/function.ldap-search.php#28991


AdSchellevis:
On my test windows server vm it does seem to return lower case too, if for some reason other products return camel case, we might have to revise how we extract these values. @ServiusHack Thanks for the PR.

PR #4425, opened 2020-10-22
Add GSuite and Youtube filtering in proxy

juliocbc:
This adds:

- Google Suite personal accounts blocking: when enabled, block any account outside of the specified domain. Very useful to limit access to Google Drive, Gmail inside corporate networks. (ref.: https://support.google.com/a/answer/1668854?hl=en)
- Youtube: Strict and Moderate modes to avoid young people access some videos with violent or sexual content. Very popular in schools networks. (ref.: https://support.google.com/a/answer/6212415?hl=en) 

**Both requires SSL interception enabled!**

### DIFF
--- a/src/opnsense/mvc/app/models/OPNsense/Proxy/Proxy.xml
+++ b/src/opnsense/mvc/app/models/OPNsense/Proxy/Proxy.xml
@@ -389,6 +389,18 @@
                 <mimeType type="CSVListField">
                     <Required>N</Required>
                 </mimeType>
+                <googleapps type="TextField">
+                    <Required>N</Required>
+                    <mask>/^([a-zA-Z0-9]){0,}\.([a-zA-Z0-9].){0,}/</mask>
+                    <ValidationMessage>Please enter a valid domain name here</ValidationMessage>
+                </googleapps>
+                <youtube type="OptionField">
+                    <Required>N</Required>
+                    <OptionValues>
+                        <strict>Strict</strict>
+                        <moderate>Moderate</moderate>
+                    </OptionValues>
+                </youtube>
                 <safePorts type="CSVListField">
                     <default>80:http,21:ftp,443:https,70:gopher,210:wais,1025-65535:unregistered ports,280:http-mgmt,488:gss-http,591:filemaker,777:multiling http</default>
                     <mask>/^([ \-0-9a-zA-Z:,])*/u</mask>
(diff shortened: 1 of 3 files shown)


src/opnsense/mvc/app/models/OPNsense/Proxy/Proxy.xml
```diff
@@ -389,6 +389,18 @@
                 <mimeType type="CSVListField">
                     <Required>N</Required>
                 </mimeType>
+                <googleapps type="TextField">
```
AdSchellevis, line 392: it's probably better to use a `HostnameField` here

src/opnsense/service/templates/OPNsense/Proxy/squid.acl.conf
```diff
@@ -99,6 +99,16 @@ http_access deny blockmimetypes_requests {% if helpers.exists('OPNsense.proxy.fo
 
 {% endif %}
 
+# Google Suite Filter
+{% if helpers.exists('OPNsense.proxy.forward.acl.googleapps') and OPNsense.proxy.forward.acl.googleapps|default('') != '' %}
+request_header_add X-GoogApps-Allowed-Domains {{OPNsense.proxy.forward.acl.googleapps}} localnet
```
AdSchellevis, line 104: is localnet needed here? most access rules don't restrict to localnet, in which case the same policy would also apply for routed networks.

AdSchellevis (changes requested):

src/opnsense/service/templates/OPNsense/Proxy/squid.acl.conf
juliocbc: In this case, what do you recommend to use here?

juliocbc (commented):
revised

src/opnsense/service/templates/OPNsense/Proxy/squid.acl.conf
AdSchellevis: @juliocbc If I'm not mistaken you can omit the acl, in which case it will match on _all_ policies (http://www.squid-cache.org/Doc/config/request_header_add/)

juliocbc: @AdSchellevis you are right! I will change it! Thanks!

AdSchellevis:
@juliocbc looks good, thanks!

PR #4429, opened 2020-10-23
rtadvd_test

marjohn56:
Use this patch for testing rtadvd. Please apply this patch after first applying f2dc854, otherwise debug level settings will not work. There are only three debug levels, level 0 is off, 1 is low and 2 is verbose.

### DIFF
--- a/src/etc/inc/plugins.inc.d/dhcpd.inc
+++ b/src/etc/inc/plugins.inc.d/dhcpd.inc
@@ -206,6 +206,12 @@ function dhcpd_radvd_configure($verbose = false, $blacklist = array())
     global $config;
 
     killbypid('/var/run/radvd.pid', 'TERM', true);
+	
+	// Wait for rtadvd to exit gracefully
+	
+	while(is_process_running("rtadvd")) {
+		sleep(1);	
+	}
 
     if (!dhcpd_radvd_enabled()) {
         return;
@@ -221,6 +227,7 @@ function dhcpd_radvd_configure($verbose = false, $blacklist = array())
 
     /* Process all links which need the router advertise daemon */
     $radvdifs = array();
+    $rtadvd_interface_list = array();
 
     /* handle manually configured DHCP6 server settings first */
     foreach (config_read_array('dhcpdv6') as $dhcpv6if => $dhcpv6ifconf) {
@@ -269,6 +276,7 @@ function dhcpd_radvd_configure($verbose = false, $blacklist = array())
         }
 
         $radvdconf .= "# Generated for DHCPv6 server $dhcpv6if\n";
+        $rtadvd_interface_list[] = $realif;
         $radvdconf .= "interface {$realif} {\n";
         $radvdconf .= "\tAdvSendAdvert on;\n";
         $radvdconf .= sprintf("\tMinRtrAdvInterval %s;\n", !empty($dhcpv6ifconf['ramininterval']) ? $dhcpv6ifconf['ramininterval'] : '200');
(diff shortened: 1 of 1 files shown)


fichtner:
Could be a candidate for 21.1. Happy to leave radvd behind eventually.

marjohn56:
OK, well there's a cock up somewhere I'm trying to sort out, nothing to do with rtadvd, more to do with my patch not applying the way I want it to... trying to apply a patch over a patch for 20.7.4, think I'll bin that idea and and just go with dev, if anyone wants to try it they'll have to switch to dev... what say you?

marjohn56:
Ah.. I'll do it a different way, reverse the first patch and create a whole new one with debug level changes too, that way it will also allow itself to be pulled into 21.1... ,

marjohn56:
@fichtner - Got something weird going on. I have completed ( as far as I can ) all the stuff that creates the rtadvd.conf, added the services so it's separate, all that works and I can start and stop all good. Now the weird bit. If I launch rtadvd from the shell with all the command params, then its working fine, if I launch it from within the GUI, then it shows up on a ps fine, but it will not respond to solicits, from the shell the ps arguments look fine, what's even more strange is that it appears to work fine on my test router, that's just using a single LAN, but on my live router which is using VLANs it doesn't, but as I said, it DOES work fine when launched manually.... very baffled.

fichtner:
Do you use mwexecf()? Try switching to mwexecf_bg() instead.

src/etc/inc/plugins.inc.d/dhcpd.inc
```diff
@@ -535,10 +544,19 @@ function dhcpd_radvd_configure($verbose = false, $blacklist = array())
...
+    if ($config['system']['ipv6_radvd_debug_level'] > 1 ) {
+        $radvddebug = " -D";
+    } else if ($config['system']['ipv6_radvd_debug_level'] = 1 ) {
+        $radvddebug = " -d";
+    } else {
+        $radvddebug = "";
     }
 
+    if (count($radvdifs)) {
+		mwexecf('/usr/sbin/rtadvd %s -c /var/etc/radvd.conf -p /var/run/radvd.pid %s',$radvddebug,$radvd_interfaces);
```
fichtner, line 558: This looks wrong. Second argument is an array with all format arguments.

src/etc/inc/plugins.inc.d/dhcpd.inc
```diff
@@ -535,10 +544,19 @@ function dhcpd_radvd_configure($verbose = false, $blacklist = array())
 
     file_put_contents('/var/etc/radvd.conf', $radvdconf);
 
-    if (count($radvdifs)) {
-        mwexec('/usr/local/sbin/radvd -p /var/run/radvd.pid -C /var/etc/radvd.conf -m syslog');
+    $radvd_interfaces = implode(' ',$rtadvd_interface_list);
```
fichtner, line 547: This is wrong too. You implode but later escape a la “int1 int2” which looks like one single interface that doesn’t exist vs. “int1” “int2”

fichtner:
There is also rtadvctl to add/remove interfaces after restart. It looks like a more flexible approach where we don’t need to prime Interfaces at start time.

marjohn56:
Yes, ignore this commit, its just a conversation holder at the moment. All sorts of stuff added now. I'll try the daemon launcher and let you know.. Will be later on tonight as other things to do for a while.

marjohn56:
#4431

marjohn56:
@fichtner - new commits on PR #4431 - This commit is well obsolete!

PR #4431, opened 2020-10-24
Added rtadvd configuration and service calls

marjohn56:
Added routines to create rtadvd config file; there seem to be some options that do not exist in rtadvd, so we'll need to see if that makes it unopperable. Using the background launch option fixed the issue with no response to RS when launched using mwexecf.

This is a first draft, so obviously there will need to be some fixes.! :)

### DIFF
--- a/src/etc/inc/plugins.inc.d/dhcpd.inc
+++ b/src/etc/inc/plugins.inc.d/dhcpd.inc
@@ -937,9 +937,9 @@ function dhcpd_rtadvd_configure($verbose = false, $blacklist = array())
 
     file_put_contents('/var/etc/rtadvd.conf', $rtadvdconf);
 
-    if ($config['system']['ipv6_radvd_debug_level'] > 1 ) {
+    if ($config['system']['ipv6_rtadvd_debug_level'] == 1 ) {
         $rtadvddebug = "-D";
-    } else if ($config['system']['ipv6_radvd_debug_level'] = 1 ) {
+    } else if ($config['system']['ipv6_rtadvd_debug_level'] == 2 ) {
         $rtadvddebug = "-d";
     } else {
         $rtadvddebug = " ";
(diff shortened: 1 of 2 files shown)


marjohn56:
Need to look at the logging output of rtadvd,  in verbose mode it appears to flood the logging but not all data appears in the logs, in intermediate mode it doesn't say much, except tells you it's exited gracefully. In low mode it says sod all.

marjohn56:
Modded rtadvd to give more output at debug level 1, no shows RS received etc. I'll PR that shortly.

marjohn56:
rtadvd PR https://github.com/opnsense/src/pull/82

fichtner:
I hope for a clean replacement patch so we can see what changes when me migrate vs. adding duplication and overrides.

marjohn56:
Totally agree, but at present, would we not be better to check it out thoroughly first? As I said, there are certain options missing in rtadvd, possibly we don't need them, possibly we do.

marjohn56:
Much the same as when we implemented dpinger, we allowed the option until we were happy there were no issues. Unfortunately in this case we do not have a case where pfs was already running rtadvd, so we have no history in that context.

fichtner:
Should not compare this to dpinger. It worked very differently compared to apinger. rtadvd is mostly a drop-in replacement with eventual caveats and if it works we'll switch service right away. For this we need to be able to review changes and features lost (worst case). Especially the second worries me. And I'm half-sure there will be new bugs later on but we'll have to see when we get there...

marjohn56:
I can issue a separate PR which does just replace radvd with rtadvd. I was also going to the ctrl function so it just forces a config re-read if the daemon is already running; another issue is the command line addition of the interfaces, should we add ALL the interfaces that are not WAN, even if they are down.?

fichtner:
we give rtadvd no interfaces on  the command line and then switch them on and off with rtadvctl how it seems to be intended by the authors

marjohn56:
Yeah, I could not make that work, but it might have been the daemon launch issue spiking me, I'll try it again without the command line interfaces.

marjohn56:
Yes, that must have been the issue, it works fine using the rdadvctl enable ...interfaces. So launch the demon if not running, wait for it then enable the interfaces?

Issue #4419, opened 2020-10-19
OpenVPN client - Gateway misconfigured or missing reply-to


Ulrar:
Hi Opnsense team,

I'm not sure weather it's a bug or if I'm doing something wrong, but after a week of being unsuccessful on the forum and on IRC I'm guessing it must be a bug.

Here's my latest thread for full reference : https://forum.opnsense.org/index.php?topic=19676.0

My issue is that I have an OpenVPN client configured to get a public IP. Whatever I do, the Gateway created when that VPN comes up has "dynamic" in IP the field when editing, but that seems to resolve to the interface's IP when confirming (I'm blacking it out because it's a public IP) : 

![image](https://user-images.githubusercontent.com/873423/96461299-1d9aca80-121c-11eb-98d4-a7b48fc94403.png)
![image](https://user-images.githubusercontent.com/873423/96461323-24294200-121c-11eb-8909-e8144361ef77.png)

That IP itself is routed to the loopback interface, which I'm assuming is added automatically : 

![image](https://user-images.githubusercontent.com/873423/96461348-2f7c6d80-121c-11eb-93c5-96259c518549.png)

(the interface's IP is the red box in the above screenshot, you can see it's routed to lo0).

The problem here is that any incoming connections on that interface, or outgoing policy based routed connections are therefore going through lo0 and never leave the router. If I try to do an http query, I get my own local HAProxy answering since it's bound to port 80, it's never going through the ovpnc1 interface.

If I edit the gateway, remove dynamic and put another IP (I've been using the first IP of the bloc my IP is in, since that has an automatic route) then policy based routing starts to work fine. But I've looked at /tmp/rules.debug and in that case the reply-to bits are missing from the inbound rules, which means the default gateway is used to answer incoming connections (internet -> vpn port 80 -> haproxy -> wan interface -> dropped by my isp).

With an edited vpn gateway, if I edit the /tmp/rules.debug file to add the missing reply-to and re-apply that file then everything works as expected.
I think the bug here is that the gateway doesn't get generated correctly by default, I don't think I should have to edit that by hand to change it's IP, right ? Since the reply-to are there and valid as long as I don't edit the gateway, I think everything would work fine if "dynamic" were to resolve to the correct IP.
An alternative fix would be to be able to force a reply-to on a given rule, but it doesn't look like there's any way to do that short of editing the file after each rule change / reboot.

Sorry for the long text, I hope the issue is clear.
[...]

Issue #4268, opened 2020-08-16
OpenVPN: False gateway address set when topology "subnet" is used


vlcty:
**Describe the bug**
One VPN provider uses net30 as addressing method. In the logs the tunnel address is properly set:

```
openvpn[64918]: /usr/local/etc/inc/plugins.inc.d/openvpn/ovpn-linkup ovpnc5 1500 1557 10.92.0.42 10.92.0.41 init
openvpn[64918]: /sbin/ifconfig ovpnc5 10.92.0.42 10.92.0.41 mtu 1500 netmask 255.255.255.255 up
```

After starting the tunnel the local tunnel address is set to `10.92.0.42` and the automatically generated gateway is set to `10.92.0.41`. 

Another VPN provider I use uses subnet as addressing method with a /24 mask. Here are the logs:

```
openvpn[92855]: /usr/local/etc/inc/plugins.inc.d/openvpn/ovpn-linkup ovpnc3 1500 1557 10.5.81.242 255.255.255.0 init
openvpn[92855]: /sbin/ifconfig ovpnc3 10.5.81.242 10.5.81.1 mtu 1500 netmask 255.255.255.0 up
```

In this case 255.255.255.0 is sent to the ovpn-linkup script. The result: The tunnel local address (`10.5.81.242`) is written to `/tmp/ovpnc3_router` and set as gateway address. The rigth IP would be `10.5.81.1`.

While debugging I found this in the OpenVPN reference manual:

```
Note: Using –topology subnet changes the interpretation of the arguments of –ifconfig to mean “address netmask”, no longer “local remote”.
```

**Environment**

OPNsense 20.7-amd64   
FreeBSD 12.1-RELEASE-p7-HBSD   
OpenSSL 1.1.1g 21 Apr 2020

AdSchellevis:
I expect it would just work if it didn't create a /tmp/xxx_router file, since it probably doesn't need the address to send the traffic to the correct place (should be easy to test). 

https://github.com/opnsense/core/blob/11017fa4ab31d33bcbf6678fd7395762e712d22c/src/opnsense/mvc/app/library/OPNsense/Routing/Gateways.php#L247-L255

Currently I don't have time to test the different options, there already is a ticket about the same topic if I'm not mistaken (https://github.com/opnsense/core/pull/4134). 
But at a first glance I think we're overcomplicating things by trying to find the endpoint (the new "Dynamic gateway policy" knob on the interface without the gateway file likely do the same). 

[...]

PR #4134, opened 2020-05-26
BUGFIX: Patch for https://github.com/opnsense/core/issues/4133

drivera73:
Proposed fix for Issue 4133 that detects the intended gateway address directly from the interface if it's not explicitly set via the environment.

PR #4433, opened 2020-10-26
Fix OpenVPN dynamic gateway detection

hippi-viking:
This is a quick hack to fix OpenVPN dynamic gateway detection for 99% of the cases. It is highly unlikely someone has a VPN gateway on their own local tunnel IP / loopback. The gateway usually is the .1 address on the same subnet as the client, this fix will help in that case. If this is not the case, a proper parser should be written (which could be overkill for simple use cases).

### DIFF
--- a/src/etc/inc/plugins.inc.d/openvpn/ovpn-linkup
+++ b/src/etc/inc/plugins.inc.d/openvpn/ovpn-linkup
@@ -8,7 +8,8 @@ if [ -n "${route_vpn_gateway}" ]; then
 elif [ -n "${ifconfig_remote}" ]; then
 	/bin/echo ${ifconfig_remote} > /tmp/$1_router
 elif [ -n "${ifconfig_local}" ]; then
-	/bin/echo ${ifconfig_local} > /tmp/$1_router
+	vpn_subnet=`/bin/echo ${ifconfig_local} | /usr/bin/cut -d'.' -f 1-3`
+	/bin/echo $vpn_subnet".1" > /tmp/$1_router
 elif [ "${dev_type}" = "tun" ]; then
 	/bin/echo ${5} > /tmp/${1}_router
 fi


AdSchellevis:
this likely won't get merged, different discussions have been going on this subject (like https://github.com/opnsense/core/pull/4134#issuecomment-634461239, https://github.com/opnsense/core/issues/4268#issuecomment-680695490), but in order help this further one will have to prepare test setups for different scenario's and explain what the expected outcome should be when using different options.

hippi-viking:
Thanks, this will be a local patch only then as well for the time being. How can I help with the test setups? I have multiple commercial VPN tunnel clients set up, they all work as per the proposed patch. I cannot imagine a scenario where the local tunnel address might be the gateway (not even for an OpenVPN server), so the proposed patch should not break anything in any case.

AdSchellevis:
If the issues only involve cases where either `route_vpn_gateway` and `ifconfig_remote` are empty, we could probably focus on the `ifconfig_local` setting for now. For policy based routing we need to know the other endpoint (otherwise we could probably just send traffic to the device), but question is how it's defined (why is it always x.x.x.1 for example) and what happens for ipv6. A couple of test scenarios would probably help a lot.

Ulrar:
If a "real" fix is going to take time and effort, could you maybe just add a tickbox "use device as gateway" in the interface settings instead ?
I actually expected ` This interface does not require an intermediate system to act as a gateway ` to do that but it doesn't seem to make any difference when ticked.

I know in my specific case using the first IP in the block works but only because it's on the correct device, my VPN's admin very specifically told me that I shouldn't have an IP at all as the gateway, and I suspect that's true for a lot of people. That's likely why it comes up empty in those variables.

AdSchellevis:
For normal destination based routing you probably shouldn't need an address, people trying to use policy based routing should likely add a manual gateway with the correct endpoint. This is the same discussion as https://github.com/opnsense/core/issues/4268#issuecomment-674529483 , when the gateway can't be predicted reliably the script should probably just remove the router file in this section https://github.com/opnsense/core/blob/a0dad1a3b22de29157db0658cbd1e8bbcc8f5200/src/etc/inc/plugins.inc.d/openvpn/ovpn-linkup#L10-L11

Ulrar:
You can't do policy based routing without a fix, and if you edit the
gateway then reply-to doesn't get added anymore, so you get policy based
routing working but you lose incoming connections.
Either way by default nothing works, the VPN is unusable unfortunately

AdSchellevis:
I don't mind fixing something in this area, the question remains what should we fix. My initial question about preparing different scenarios and their expected outcome is all aimed at collecting the relevant data points.....  if the `${ifconfig_local}` never works (which I assume, but haven't seen proof this covers all scenarios), we can aim at a solution when `${route_vpn_gateway}` and `${ifconfig_remote}` are not provided (same question applies for ipv6). Personally I do like to avoid hot fixing this without a testbed  and a clear description of the issue. If someone wants to help preparing the test setups and detailing the issue, that would be great, if not, it will likely remain for at least a couple of months (I currently can't spare the time to test this properly).

hippi-viking:
Sadly I don't have IPv6 connection (not even my ISP supports it), so I cannot test it easily.
This patch is intended as a last resort fix for cases where `${ifconfig_remote}` and `${route_vpn_gateway}` are not defined, as `${ifconfig_local}` will not work for sure. I would not mind setting up the gateway manually (I have been doing that for months), it is just annoying when you connect to a load-balanced VPN with different IP-ranges (so you have to change the gateway every time it reconnects).
x.x.x.1 seems to be the gateway on most commercial VPNs I have tried, which is of course not a standard but seems like a fairly common setting for it to work out of the box almost every time.

Ulrar:
So I finally got around to testing the patch in this PR and it works fine, except in my case the first IP of the bloc is .129. IMHO if we just added a line to calculate the first IP from the `${ifconfig_local}` + subnet it would work fine for most people, which is likely more than the current behavior.
I'll keep applying it by hand for now I guess, thanks for the fix in the meantime

AdSchellevis:
@Ulrar your probably right, although it would still be practical if more people test different configurations, this https://github.com/opnsense/core/commit/0ad3ec432ff0d1ee45d9969424b7e5b19eb903e2 should probably help for most setups (assuming `ifconfig_netmask`  is being supplied in all cases. (https://github.com/opnsense/core/pull/4134#issuecomment-634665838 seems to suggest this as well)

hippi-viking:
Thanks Ad.
For my cases netmask is indeed seems to be supplied:

(excerpts from the OpenVPN log:)
vpn1: `/sbin/ifconfig ovpnc1 10.18.0.x 10.18.0.1 mtu 1500 netmask 255.255.0.0 up`
vpn1: `/usr/local/etc/inc/plugins.inc.d/openvpn/ovpn-linkup ovpnc1 1500 1585 10.18.0.x 255.255.0.0 init`
vpn2: `/sbin/ifconfig ovpnc2 172.27.0.x 172.27.0.1 mtu 1500 netmask 255.255.252.0 up`
vpn2: `/usr/local/etc/inc/plugins.inc.d/openvpn/ovpn-linkup ovpnc2 1500 1552 172.27.0.x 255.255.252.0 init`

I was thinking either calculating the first IP to be the gateway(?) based on the local IP and netmask OR parsing the routing table (OpenVPN seems to set it up correctly based on the log) OR parsing the openvpn.log file itself to get the gateway IP address.

AdSchellevis:
so, the question is, does the calculation )in https://github.com/opnsense/core/commit/0ad3ec432ff0d1ee45d9969424b7e5b19eb903e2) work for all use-cases known so-far. I rather prevent parsing route tables or ifconfig statements to be honest.

vlcty:
I also stumbled over this problem. I created issue #4268 .
With IPv6 everything is fine, because `ifconfig_ipv6_remote` is always passed via ENV which the ovpn-linkup script parses.

Fo IPv4 the problem only occures with topology `subnet` instead of `net30`. It seems that the first address is always assigned to the server. You COULD change it, however you normally don't do it. At least I have never seen it anywhere, but who am I.   
`ifconfig_netmask` and `ifconfig_local` is also always passed as far as I can see so using it for calculations is In my eyes a good solution.

AdSchellevis:
If https://github.com/opnsense/core/commit/0ad3ec432ff0d1ee45d9969424b7e5b19eb903e2 seems to work for everybody in this thread, I propose to close this PR and probably https://github.com/opnsense/core/issues/4419 as well.

Ulrar:
`[0ad3ec4](https://github.com/opnsense/core/commit/0ad3ec432ff0d1ee45d9969424b7e5b19eb903e2)` does work perfectly, any chance to get that merged at some point ? It'd be nice not to have to go edit the file after each update.

AdSchellevis:
@Ulrar ok, thanks for letting us know, I'll close https://github.com/opnsense/core/issues/4419 and schedule this for the next release or the one after.

kkoshelev:
Not fixed? Would anyone look at my data and help with settings?

I'm having similar problems in my setup. The openvpn gateway doesn't get correct settings. ovpn-linkup didn't get called with valid gateway address. I suspect the problem is with PUSH_REPLY sent by vpn server. I have two cases with exactly the same OPNsense settings, one of them works another doesn't.

**PIA VPN - doesn't work**
PiaVPN ovpnc2 gateway overview
IPv4 address | 10.7.112.24 / 24
Gateway IPv4 | 10.7.112.24  <- wrong ip address, should be 10.7.112.1

routing table
ipv4 | 10.7.112.0/24 | 10.7.112.1 | UGS | 0 | 1500 | ovpnc2 | PiaVPN |   |  
ipv4 | 10.7.112.1 | link#13 | UH | 0 | 1500 | ovpnc2 | PiaVPN |  
ipv4 | 10.7.112.24 | link#13 | UHS | 1098 | 16384 | lo0 | Loopback

logs
2020-12-01T18:37:00 | openvpn[6741] | /usr/local/etc/inc/plugins.inc.d/openvpn/ovpn-linkup ovpnc2 1500 1553 10.7.112.24 255.255.255.0 init
2020-12-01T18:37:00 | openvpn[6741] | /sbin/route add -net 10.7.112.0 10.7.112.1 255.255.255.0
2020-12-01T18:37:00 | openvpn[6741] | PUSH: Received control message: 'PUSH_REPLY,comp-lzo no,redirect-gateway def1,route-ipv6 2000::/3,dhcp-option DNS 10.0.0.243,route-gateway 10.7.112.1,topology subnet,ping 10,ping-restart 60,ifconfig 10.7.112.24 255.255.255.0,peer-id 2,cipher AES-128-GCM'

**ExpressVPN - works**
ovpnc1 gateway 
IPv4 address | 10.158.0.70 / 32
Gateway IPv4 | 10.158.0.69

routing table
ipv4 | 10.158.0.69 | link#12 | UH | 69705 | 1500 | ovpnc1 | ExpressVPN |   |  
ipv4 | 10.158.0.70 | link#12 | UHS | 0 | 16384 | lo0 | Loopback |  

logs
2020-12-01T17:44:02	openvpn[70454]	/usr/local/etc/inc/plugins.inc.d/openvpn/ovpn-linkup ovpnc1 1500 1557 10.158.0.70 10.158.0.69 init
2020-12-01T17:44:02	openvpn[70454]	/sbin/ifconfig ovpnc1 10.158.0.70 10.158.0.69 mtu 1500 netmask 255.255.255.255 up
2020-12-01T17:44:02	openvpn[70454]	PUSH: Received control message: 'PUSH_REPLY,redirect-gateway def1,dhcp-option DNS 10.158.0.1,comp-lzo no,route 10.158.0.1,topology net30,ping 10,ping-restart 60,ifconfig 10.158.0.70 10.158.0.69,peer-id 13,cipher AES-256-GCM'

PR #4434, opened 2020-10-26
Properly halt instead of poweroff

hippi-viking:
When using UPS it is better to halt (as the name suggests) in case of powerloss instead of poweroff. This way the appliance will halt when the power is low, but will be able to restart when power returns. (If it is powered off it will not reboot by itself when power returns.)

### DIFF
--- a/src/etc/rc.halt
+++ b/src/etc/rc.halt
@@ -2,4 +2,4 @@
 
 /usr/local/etc/rc.syshook stop
 
-exec /sbin/shutdown -op now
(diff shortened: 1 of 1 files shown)


AdSchellevis:
our `rc.halt` routes to power off (in the user interface), so I don't think we should change that behaviour for one of the services. If the service wants a different type of shutdown, it should logically being take care of by the service (probably nut in this case).

hippi-viking:
I see, in that case I close this off and keep it as a local change instead. Thanks.

Issue #1434, opened 2017-02-27
Use minus (-) character in alias name


Tragen:
Please allow the minus character in alias names as requested here:
https://forum.opnsense.org/index.php?topic=4626.msg17839#msg17839

fichtner:
I'm late to the party, sorry.

Aliases are macros in pf.conf, the parsing of macro expansion can only match alphanumeric characters and the underscore character. During macro expansion the minus character is used as a breaking delimiter, e.g. for port ranges.

The relevant code line is this: https://github.com/opnsense/src/blob/master/sbin/pfctl/parse.y#L5702

scalfitura:
@fichtner so if I understood correctly this request is technical, not possible, at least not in the never future, right?

fichtner:
Yes and no. There's no easy fix here, but we could use a double underscore to encode the minus in the rules, but it takes a bit of work and care that "__" is not used in an alias already, maybe even deny using this combination.

scalfitura:
@fichtner, Ok, thanks for the explanation. I think it will be really handy if this will be included future (major) releases, the minus is really often used in (server) hostnames, maybe this becomes handy for a lot of people. but again, only if possible of course.

Thanks anyway!

fichtner:
I'll see what we can do, thank you for the feedback :)

reaby:
In case you use UTF-8 systemwide, it's maybe possible to do something like mb_str_replace("-", "‑", $string); which would change the normal "minus-sign" to"non-breaking hyphen". They look graphically same, but are different glyphs in UTF-8 charset.

fichtner:
but that still won't allow use in pf.conf ASCII syntax :(

the best solution is stuffing, e.g. replacing "_" (one) by "__" (two) internally and using "-" as "_". that still requires audit of validation and code paths, also readings and mapping it back.

AdSchellevis:
timeout

Tragen:
Why was this closed?
It's still not fixed.

fabianfrz:
@Tragen because nobody worked on it.

PR #4439, opened 2020-11-03
Allow hyphen '-' in aliases

tafros:
Allow the hyphen '-' in aliases. This has been tested in aliases and nested aliases in firewall rules. This is adding the hyphen and not using HostnameField validation for backwards compatibility with existing aliases.

This feature was requested in forum post below:
https://forum.opnsense.org/index.php?topic=6468.msg27746#msg27746


AdSchellevis:
this doesn't work, see man https://github.com/opnsense/core/issues/1434, https://github.com/opnsense/src/blob/41ba6e29a8d3f862f95f9ab0a1482ef58c4a7cdb/sbin/pfctl/parse.y#L5748

PR #4440, opened 2020-11-03
Feature: Widget with DHCP-leases

Kipjr:
Hi, 

I missed a widget showing me the DHCP-leases on my network, so I made my own. It requires some finetuning but it seems to be working. 

Tested it  on 

OPNsense 20.7.2-amd64
FreeBSD 12.1-RELEASE-p8-HBSD
OpenSSL 1.1.1g 21 Apr 2020

image

### DIFF
--- a/src/www/widgets/include/dhcp_list.inc
+++ b/src/www/widgets/include/dhcp_list.inc
@@ -1,0 +1,4 @@
+<?php
+
+$dhcp_lease_title = gettext('DHCP Leases');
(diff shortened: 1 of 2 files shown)


AdSchellevis:
the parsing code in `status_dhcp_leases.php` should probably be refactored and migrated to mvc first so both components can use the same logic. We don't want to duplicate  a lot of code into a widget.

fichtner:
I agree with @AdSchellevis on this and some of the leases code was already refactored/improved so please feel free to raise a new PR.

PR #4447, opened 2020-11-07
DHCPv4: Removed the need for a static IP being outside of the pool

Gauss23:
As we had the discussion on the forum in the last days, I had a look at the software OPNsense is using as DHCP server. As it turns out it is ISC DHCP, so there is no need to keep the static leases outside of the pool.
Just removed the check from the PHP file and changed one description in line 326 (do I need to do something about translations here?).

Copied the file to my test OPNsense and I was able to set a static lease within the pool. Client was receiving the correct IP.

Old line 144 was not needed anymore as we're already checking this in line 117.

### DIFF
--- a/src/www/services_dhcp_edit.php
+++ b/src/www/services_dhcp_edit.php
@@ -140,27 +140,9 @@ if ($_SERVER['REQUEST_METHOD'] === 'GET') {
 
     $parent_net = find_interface_network(get_real_interface($if));
 
-    /* make sure it's not within the dynamic subnet */
-    if (!empty($pconfig['ipaddr'])) {
-        $dynsubnet_start = ip2ulong($config['dhcpd'][$if]['range']['from']);
-        $dynsubnet_end = ip2ulong($config['dhcpd'][$if]['range']['to']);
-        if (ip2ulong($pconfig['ipaddr']) >= $dynsubnet_start && ip2ulong($pconfig['ipaddr']) <= $dynsubnet_end) {
-            $input_errors[] = sprintf(gettext("The IP address must not be within the DHCP range for this interface."));
-        }
-
-        if (!empty($config['dhcpd'][$if]['pool'])) {
-            foreach ($config['dhcpd'][$if]['pool'] as $pidx => $p) {
-                if (is_inrange_v4($pconfig['ipaddr'], $p['range']['from'], $p['range']['to'])) {
-                    $input_errors[] = gettext("The IP address must not be within the range configured on a DHCP pool for this interface.");
-                    break;
-                }
-            }
-        }
-
-        if (!ip_in_subnet($pconfig['ipaddr'], $parent_net)) {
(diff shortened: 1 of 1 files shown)


AdSchellevis:
I need to test this locally to be sure, but if it was a limitation of isc-dhcp which isn't valid anymore, I'm not against removing the validation.

Gauss23:
Sorry, line 143 is still needed (empty check), if someone doesn't enter a static IP: https://github.com/opnsense/core/commit/dfccde59a742a1730a67a99c85e24cd9450dcab0

Should I create a new pull request?

AdSchellevis:
this one is still open, so need to to open a new one we can just squash two commits into one.

Gauss23:
A little bit off-topic: there are some  menuitems which use the new MVC design pattern like` /ui/ipsec/key-pairs`

But there are still a lot of menuitems running a php file, which mixes everything together. I assume the plan is to shift all those legacy pages to the new MVC model, right?

AdSchellevis:
Ideally yes, but in reality some parts likely won't be migrated anytime soon (no real business case, a proper migration can be quite expensive). When adding new isolated functionality, the shortest path usually is to use our models and integrate into the legacy base, the other way around (using legacy code in new style) isn't allowed to prevent dead-locks in the future.

Some context about this is also written down in our coding guidelines https://docs.opnsense.org/development/guidelines/basics.html#strategy

AdSchellevis:
@Gauss23 seems to work fine indeed, thanks!

Gauss23:
I noticed this merge isn't yet in the current releases. Is there a problem with it?

fichtner:
Standard policy: give a wider test opportunity for the change by adding it to the development version for one cycle. It is now included in the 20.7.7 -devel package. It could land in 20.7.8 depending on feedback, scope and type of change. Worst case it will land in 21.1-RC, but both these things are just 4-5 weeks away anyway.


Cheers,
Franco

fichtner:
PS: may have been in 20.7.6 -devel as well, I’m not sure. Scope for 20.7.7 was to get out LibreSSL/OpenSSL and other security related things.

Gauss23:
It wasn’t meant in a bad way. I was just curious how and when such a merge will find its way into the production tree.

Thanks for clarification.

fichtner:
No worries, I only tried to explain as well :)

dsimoes:
Hi @Gauss23 and everyone

I found an issue with the implementation. I can in fact do static leases within the DHCP range, but once I do, we no longer have the ability to edit and save the DHCP settings. It will not pass the validation:

- The DHCP range cannot overlap any static DHCP mappings.

So, right now I'm unable to do any changes unless I remove all the static leases already configured.

I'm not sure if I'm suppose to open a new bug, I searched and found the original PR for this.

fichtner:
Best to open a new bug report with this reference link.


Thanks,
Franco

dsimoes:
Done. Thanks.

PR #4456, opened 2020-11-12
Change dpinger syslog message to reflect correct RTT and RTTd unit

fhloston:
Dpinger actually reports RTT and RTTd in µs.


AdSchellevis:
@fhloston nice catch, thanks!

PR #4432, opened 2020-10-25
replace radvd with rtadvd

marjohn56:
Clean replacement, no optional radvd. Interfaces->Settings has a debug level option, level 3 seems to flood the logs and misses some data. Please read the comments where I have found no direct replacement, e.g. decremenent, though it appears to do this by default anyway, config file has been created in such a format as to make it easily readable, hence the \x51's. It could all be on one line for each interface, but it's not very user friendly like that.

PR #4461, opened 2020-11-14
Replace radvd with rtadvd - Clean

marjohn56:
Clean version applied against Master source 14-11-2020
Something went amiss with my repo for #4432 which left me unable to merge cleanly, so this PR replaces it..

### DIFF
--- a/src/etc/inc/plugins.inc.d/core.inc
+++ b/src/etc/inc/plugins.inc.d/core.inc
@@ -218,7 +218,7 @@ function core_syslog()
     $logfacilities['ppps'] = array('facility' => array('ppp'));
     $logfacilities['resolver'] = array('facility' => array('unbound'));
     $logfacilities['routing'] = array('facility' => array(
-          'radvd', 'routed', 'rtsold', 'olsrd', 'zebra', 'ospfd', 'bgpd', 'miniupnpd'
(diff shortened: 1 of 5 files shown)


MartB:
So far this has been stable for me, can someone from opnsense check this out?

MartB:
So far this has been stable for me, can someone from opnsense check this out?
@fichtner what do you think about this?

fichtner:
@MartB there is not enough data to indicate rtadvd is the better choice so far. We have some users with a multitude of problems which could range from software to equipment to ISP and some might try rtadvd and say it works, some will not reply because it doesn't and nobody who uses radvd year in year out with no issues is going to try rtadvd. Making a decision here now is impossible.

marjohn56:
I'll maintain and enhance this patch as long as it's needed, But that means users must test out RADVD when updates/changes that could resolve the perceived  radvd issues are made.

s1shed:
> I'll maintain and enhance this patch as long as it's needed, 

Thank you :)

> But that means users must test out RADVD when updates/changes that could resolve the perceived radvd issues are made.

Every time there's a new opnsense update I revert this patch to go back to radvd. So far I've have to go back to rtadvd each time for router advertisements to work. :|

Issue #4480, opened 2020-11-24
Feature: Dashboard CPU usage graph style


tofazzz:
Howdy,

I've posted in the forum and apparently it seems like a feature that users would appreciate; What do you think about changing the CPU usage graph in the dashboard to be the same as the one for State table size/MBUF Usage/Memory usage/Disk usage?

Beside being the same style I think the bars provide more detailed view, always showing the percentage value of the usage.

"Fright" user from the forum kindly provided a proposed code with a practical example. Please find attached the screenshot and the changed code (strings added: 49-52;215-223).

In my opinion, if we add the CPU bar then we can get rid of the current graph as, if I want to have detailed historical CPU data I can add the CPU usage widget.
![widget](https://user-images.githubusercontent.com/64267878/100144439-c7492900-2e64-11eb-8c27-ed5b5c922401.PNG)
[system_information.widget.php.txt](https://github.com/opnsense/core/files/5592332/system_information.widget.php.txt)

fichtner:
Hi,

I'm ok with this replacement, but @AdSchellevis should be too before we go ahead.


Cheers,
Franco

AdSchellevis:
is it a replacement or an addition? the screenshot shows both.

kulikov-a:
Hi
on the forum, I suggested this exactly as an addition. since the graph is convenient in that it displays the history of values
https://forum.opnsense.org/index.php?topic=20003.0

AdSchellevis:
I don't mind adding the current bar as well.

kulikov-a:
ok! should i make a PR?

AdSchellevis:
sure, go ahead!

fichtner:
Done 😊

PR #4482, opened 2020-11-25
SysInfo widget: add current CPU load progress-bar

kulikov-a:
refers to FR: https://github.com/opnsense/core/issues/4480


AdSchellevis:
@kulikov-a works like a charm, thanks!

PR #4486, opened 2020-11-26
DHCP: Add min-secs option for each subnet

vnxme:
The min-secs option allows to configure a delay of 0-255 seconds before a DHCP server responds to a client's request. This can be used to set up a secondary DHCP server as stated in the [manual](https://www.freebsd.org/cgi/man.cgi?query=dhcpd.conf&sektion=5&apropos=0&manpath=FreeBSD+12.2-RELEASE+and+Ports). This PR contains changes for both v4 and v6, however I have not tested the latter much.

image


AdSchellevis:
@vnxme looks good, thanks!

PR #4499, opened 2020-12-03
Add LAGG support to console

sarthurdev:
This PR adds support for LAGG configuration on initial setup and via console.

It helps a lot when dropping a new install into an existing environment that already has LAGGs configured downstream. Would no longer require a temporary config just to access the web interface to then create the LAGGs and VLANs.

### DIFF
--- a/src/etc/inc/console.inc
+++ b/src/etc/inc/console.inc
@@ -132,6 +132,40 @@ EOD;
 
     $ifnames = array_keys($iflist);
 
+        echo <<<EOD
+
+You now have the opportunity to configure LAGGs.  If you don't require LAGGs
+for initial connectivity, say no here and use the GUI to configure LAGGs later.
+
+Do you want to configure LAGGs now? ${yes_no_prompt}
+EOD;
+    if ($interactive) {
+        $key = chop(fgets($fp));
+    } else {
+        $key = 'n';
+        echo $key . PHP_EOL;
+    }
+
+    if (in_array($key, array('y', 'Y'))) {
+        lagg_setup($iflist, $fp);
+    }
+
+    if (isset($config['laggs']['lagg'][0])) {
+        echo "\nLAGG interfaces:\n\n";
+        foreach ($config['laggs']['lagg'] as $lagg) {
+            $members = implode(",", $lagg['members']);
+            echo sprintf(
+                "% -16s%s\n",
+                "{$lagg['laggif']}",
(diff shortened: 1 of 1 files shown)


AdSchellevis:
I'm not sure about this, there already is quite some code duplication in `console.inc`, adding more of roughly the same isn't too great, but functionally I understand that in some cases it can be practical. Let's give this some time to mature and see if more people mis this type of functionality.

Ideally it would be more practical if one could plugin interface types using a more or less standard workflow (from an automation perspective even more practical if one could do that unattended), but short term that's likely not on our radar.

fichtner:
I thought a bit about this and I'm not opposed. The code could use a cleanup. If we can make vlan_setup() an atomic function call we could make it pluggable after lagg_setup() was added in the same manner?

AdSchellevis:
Moving to 21.1 I'm not sure the timing is great to refactor this, I'm also a bit curious how many people we're actually helping with this. In the grand scheme of things it would be good to improve this, but designing a better solution which is more resilient to change costs time. I guess most is about priorities and risk of breaking things when releasing new installer images for 21.1.

So I'm also not against any type of change, but it would be good to know if this is an annoyance shared by thousands or just something a couple of people walk into once in their lifetime. Refactoring the console menu might cost quite some time and if we can combine it with an easier road to automation that would likely help more people than pushing a single use-case. 

An option could be if there are enough users for the feature, to start integrating more or less "as-is" and create a ticket for the change we eventually would like to implement.

sarthurdev:
I will agree that the majority of users won't make use of this, however I think there should be some method put in place to create LAGGs while the Web UI is potentially unavailable.

I have also found a post on the pfSense bug tracker created 11 years ago regarding this: https://redmine.pfsense.org/issues/336

maxfield-allison:
I'd like to throw my hat in on this as well. Definitely would appreciate the functionality despite only needing it in a few cases. It seems to be one of those things that you rarely need but when its available, and you need it, its a life saver.

fraenki:
I'm also interested in this and fully agree to the previous comment. It's definitely something that is common in enterprise/datacenter environments. We heavily rely on LAGG, so having it on the console is a potential life saver in certain situations.

AdSchellevis:
@sarthurdev after some internal discussion, we're going to merge it, the code looks reasonable and fits the existing structure. Thanks for contributing!

PR #4513, opened 2020-12-10
DHCPv6: Fix sorting of IPv6 static mappings

vnxme:
**Issue:**
The DHCPv6 static mappings are shown in the order they were added. Sorting code exists but won't work. 

**Reason:**
The ipcmp() function doesn't seem to return a correct result for IPv6 addresses. 

**Solution:**
Introduce ip6cmp() function and fix sorting of IPv6 static mappings.

**Version:**
OPNsense 20.7.6

### DIFF
--- a/src/etc/inc/util.inc
+++ b/src/etc/inc/util.inc
@@ -1187,20 +1187,8 @@ function check_subnets_overlap($subnet1, $bits1, $subnet2, $bits2)
     return ($sn1 == $sn2);
 }
 
-/* compare two IPv4 addresses */
+/* compare two IP addresses */
 function ipcmp($a, $b)
-{
-    if (ip_less_than($a, $b)) {
-        return -1;
-    } elseif (ip_greater_than($a, $b)) {
-        return 1;
-    } else {
-        return 0;
-    }
-}
-
-/* compare two IPv6 addresses */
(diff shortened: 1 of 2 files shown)


src/etc/inc/util.inc
```diff
@@ -1199,6 +1199,20 @@ function ipcmp($a, $b)
     }
 }
 
+/* compare two IPv6 addresses */
+function ip6cmp($a, $b)
+{
```
fichtner, line 1204: Wouldn’t this work for IPv4 too?

src/etc/inc/util.inc
```diff
@@ -1199,6 +1199,20 @@ function ipcmp($a, $b)
...
+    $na = inet_pton($a);
```
fichtner, line 1205: Sorry, meant this line 😉

vnxme: Yes, I thought above it but wasn't brave enough to change ipcmp() at once. Now it seems secure as this function is used only for sorting of dhcp/dhcpv6 static mappings. See a new commit below.

fichtner: let's be brave 👍

fichtner:
Merged, thanks!

PR #4514, opened 2020-12-10
DHCP: Fix sorting of leases

vnxme:
**Issues:**
1. DHCPv6 leases are shown in a wrong order (e.g. ::00d0 above ::0010)
1. DHCPv6 leases cannot be sorted by another column (e.g. hostname)
1. DHCPv6 leases have an empty interface field (all lines)
1. DHCPv6 leases have an empty interface field when IPv6 address belongs to /120 subnet

**Reasons:**
1. DHCPv4/v6 leases pages use strnatcasecmp() and strcmp() for sorting by IP
1. No UI correspoding to this feature
1. $staticent['ipaddrv6'] should be used instead of $staticent['ipaddr'] when searching a static mapping
1. Incomplete preg pattern used to find possible IPv6 subnets

**Solutions:**
1. Use the updated ipcmp() function for sorting by IPv4/v6
1. Add sorting UI to DHCPv6 leases page
1. Fix a typo
1. Apply a consistent approach of using is_subnet{v4,v6} to find possible subnets

**Version:**
OPNsense  20.7.6

### DIFF
--- a/src/etc/inc/interfaces.inc
+++ b/src/etc/inc/interfaces.inc
@@ -3772,21 +3772,20 @@ function guess_interface_from_ip($ipaddress)
         /* create a route table we can search */
         exec("/usr/bin/netstat -rnWf inet", $output, $ret);
         foreach ($output as $line) {
-            if (preg_match("/^[0-9]+\.[0-9]+\.[0-9]+\.[0-9]+\/[0-9]+[ ]+link[#]/", $line)) {
-                $fields = preg_split("/[ ]+/", $line);
+            $fields = preg_split("/[ ]+/", $line);
+            if (is_subnetv4($fields[0])) {
                 if (ip_in_subnet($ipaddress, $fields[0])) {
                     return $fields[5];
-                }
+                }                
             }
         }
     }
-    /* FIXME: This works from cursory testing, regexp might need fine tuning */
     if (is_ipaddrv6($ipaddress)) {
         /* create a route table we can search */
         exec("/usr/bin/netstat -rnWf inet6", $output, $ret);
         foreach ($output as $line) {
-            if (preg_match("/[0-9a-f]+[:]+[0-9a-f]+[:]+[\/][0-9]+/", $line)) {
-                $fields = preg_split("/[ ]+/", $line);
+            $fields = preg_split("/[ ]+/", $line);
+            if (is_subnetv6($fields[0])) {
                 if (ip_in_subnet($ipaddress, $fields[0])) {
                     return $fields[5];
                 }
--- a/src/www/status_dhcp_leases.php
+++ b/src/www/status_dhcp_leases.php
@@ -246,19 +246,19 @@ if ($_SERVER['REQUEST_METHOD'] === 'GET') {
 
     $order = ( $_GET['order'] ) ? $_GET['order'] : 'ip';
 
-    usort($leases,
-        function ($a, $b) use ($order) {
-            if ($order === 'ip') {
+    if ($order === 'ip') {
+        usort($leases, function ($a, $b) {
+            return ipcmp($a['ip'], $b['ip']);
+        });
+    } else {
+        usort($leases, function ($a, $b) use ($order) {
+            $cmp = strnatcasecmp($a[$order], $b[$order]);
+            if ($cmp === 0) {
                 $cmp = ipcmp($a['ip'], $b['ip']);
-            } else {
-                $cmp = strnatcasecmp($a[$order], $b[$order]);
-                if ($cmp === 0) {
-                    $cmp = ipcmp($a['ip'], $b['ip']);
-                }
             }
             return $cmp;
-        }
-    );
+        });
+    }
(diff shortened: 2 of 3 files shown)


src/www/status_dhcp_leases.php
```diff
@@ -246,15 +246,19 @@ function remove_duplicate($array, $field)
 
     $order = ( $_GET['order'] ) ? $_GET['order'] : 'ip';
 
-    usort($leases,
-        function ($a, $b) use ($order) {
+    if ($order === 'ip') {
```
fichtner, line 249: this looks a bit convoluted. If we pass $order top the usort function we only need one which does inline if/else on the function to use?

src/www/status_dhcpv6_leases.php
```diff
@@ -307,9 +307,19 @@ function parse_duid($duid_string)
         }
     }
 
-    if ($_GET['order']) {
+    $order = ( $_GET['order'] ) ? $_GET['order'] : 'ip';
```
fichtner, line 310: this looks fishy -- some pages have weak handling of input arguments and their defaults

fichtner: hmm, ok it emulates the dhcp v4 leases page. now we have two problems :D

src/etc/inc/interfaces.inc
```diff
@@ -3772,21 +3772,20 @@ function guess_interface_from_ip($ipaddress)
         /* create a route table we can search */
         exec("/usr/bin/netstat -rnWf inet", $output, $ret);
         foreach ($output as $line) {
-            if (preg_match("/^[0-9]+\.[0-9]+\.[0-9]+\.[0-9]+\/[0-9]+[ ]+link[#]/", $line)) {
```
fichtner, line 3775: interfaces.inc changes should probably not be part of this PR as it is used outside of dhcp leases page as well. oddly enough it is not used in the v4 leases page.

src/www/status_dhcp_leases.php
vnxme: Not sure whether I understand your idea right. In commit 2adf500 I put the $order condition inside the usort function and it looked quite nice. Then I realised that we can eliminate this condition being checked multiple times while sorting by means of splitting the single usort function into two ones: a simple one for sorting by IP (no need to even have $order inside the function namespace) and an extra one for sorting by arbitrary field chosen by the user. The latter approach is what I implemented in commit a012f89.

src/www/status_dhcpv6_leases.php
vnxme: Yes, I copied it from the DHCPv4 leases page. Suggested a more secure way for both pages in commit 43427bb.

src/etc/inc/interfaces.inc
vnxme: I have nothing against pulling these changes separately. However, some DHCPv6 leases (e.g. /120 ones) will not be shown correctly unless the interfaces.inc file is fixed.

src/www/status_dhcpv6_leases.php
```diff
@@ -430,22 +453,21 @@ function parse_duid($duid_string)
...
+                    <th class="act_sort" data-field="start"><?=gettext("Start"); ?></th>
+                    <th class="act_sort" data-field="end"><?=gettext("End"); ?></th>
+                    <th class="act_sort" data-field="online"><?=gettext("Online"); ?></th>
+                    <th class="act_sort" data-field="act"><?=gettext("Lease Type"); ?></th>
                     <th class="text-nowrap"></th>
                 </tr>
               </thead>
               <tbody>
 <?php
-              $mac_man = json_decode(configd_run("interface list macdb json"), true);
```
fichtner, line 448: We should probably show the info below the mac address instead of deleting the unused variable.

src/www/status_dhcpv6_leases.php
```diff
@@ -454,7 +476,7 @@ function parse_duid($duid_string)
                     foreach ($config['dhcpdv6'] as $dhcpif => $dhcpifconf) {
                         if (isset($dhcpifconf['staticmap'])) {
                             foreach ($dhcpifconf['staticmap'] as $staticent) {
-                                if ($data['ip'] == $staticent['ipaddr']) {
+                                if ($data['ip'] == $staticent['ipaddrv6']) {
```
fichtner, line 479: looks like a fix... can we separate the $mac_man use and this into a separate PR?

src/etc/inc/interfaces.inc
fichtner: I'm saying its easier to pull in fixes first before working on GUI representation...

src/www/status_dhcp_leases.php
fichtner: This inflates size about a factor of four compared to when the code was when this lease cleanup started. I understand that adding code is necessary to improve functionality but as a leaf page we don't need to prematurely optimise it.

vnxme:
OK, I will prepare a few separate PRs and list them below.

Issue #4518, opened 2020-12-12
unbound: check if a restart is required, reload after dnsbl update


kulikov-a:
**Important notices**
Before you add a new report, we ask you kindly to acknowledge the following:

[+] I have read the contributing guide lines at https://github.com/opnsense/core/blob/master/CONTRIBUTING.md

[+] I have searched the existing issues and I'm convinced that mine is new.

**Is your feature request related to a problem? Please describe.**
Hello everyone.
For now:
- if link on any interface goes down and up, unbound updates the hosts records and restarts. Even if the address has not changed or is static. The restart can take a while if large dnsbl list is in use, so  clients practically lose the Internet for some time.
- unbound does not restart after updating the dnsbl list(s) ( `pluginctl unbound restart` . typo? `-s` missed ?)
https://forum.opnsense.org/index.php?topic=20356.0

**Describe the solution you'd like**
could you consider changes like this:
- check if the host_entries.conf file has actually changed before unbound restart 
 (something like https://github.com/kulikov-a/core/commit/d799447ae87bd722a301c83822cf7671c088ee9e).

- change actions_unbound.conf [dnsbl] command to
`/usr/local/opnsense/scripts/unbound/download_blacklists.py && configctl unbound reload`
( `pluginctl -s unbound restart` works too. but reload is far more fatser than restart. and imho reload is enough after dnsbl update. UPDATE! in my case difference in time for reload vs restart was due to a failed  unbound-anchor launch. not by the procedure itself)
Thanks!

**Describe alternatives you've considered**

**Additional context**

AdSchellevis:
@kulikov-a the dnsbl part looks like a bug indeed, https://github.com/opnsense/core/commit/3facaaa4700a9e5391352ad60733518099746c45 should fix it. Adding more logic in an attempt to "restart unbound" properly I'm not really a fan of. In my opinion most of these restart actions on interface changes shouldn't happen at all, since the service should take care of its own issues. When bound to a dynamic interface one could argue that using a port forward to an internal address would be a safer choice (and is a pattern most other vendors advise as well to prevent cluttering services).

Short term I'm not sure if we can detach unbound from the interface reload actions, there's more weird cluttering of interface addresses in it at the moment.

kulikov-a:
@AdSchellevis thanks! 
not detach, just skip unnecessary action.
I'm not a fan either, but I think that changing host entries through a `unbound-control` is also not a nifty way(
[...]

PR #4520, opened 2020-12-12
unbound.inc: reload after hosts records generate

kulikov-a:
reload, not HUP. 
https://github.com/opnsense/core/issues/4518


AdSchellevis:
@kulikov-a we'll keep this open for now, discussed this a bit internally, but every attempt made to improve reload performance ends up with similar issues (purging cache for example).

kulikov-a:
Ok! Thanks!

fichtner:
This shouldn't fix anything anymore so I'm closing it.

PR #4523, opened 2020-12-14
interfaces.inc: Improve guess_interface_from_ip()

vnxme:
**Issue:**
DHCPv6 leases have an empty interface field when IPv6 address belongs to /120 subnet.

**Reason:**
Incomplete preg pattern used to find possible IPv6 subnets in guess_interface_from_ip() function.

**Solution:**
Apply a consistent approach of using is_subnet{v4,v6} to find possible subnets.

**Other:**
This PR is a part of changes proposed by [#4514](https://github.com/opnsense/core/pull/4514).

### DIFF
--- a/src/etc/inc/interfaces.inc
+++ b/src/etc/inc/interfaces.inc
@@ -3768,7 +3768,7 @@ function guess_interface_from_ip($ipaddress)
     if (is_ipaddrv4($ipaddress)) {
         $family = "inet";
     } elseif (is_ipaddrv6($ipaddress)) {
-    	$family = "inet6";
+        $family = "inet6";
     } else {
     	return false;
     }
--- a/src/etc/inc/interfaces.inc
+++ b/src/etc/inc/interfaces.inc
@@ -3765,33 +3765,25 @@ function get_real_interface($interface = 'wan', $family = 'all')
 
 function guess_interface_from_ip($ipaddress)
 {
-    if (!is_ipaddr($ipaddress)) {
-        return false;
-    }
     if (is_ipaddrv4($ipaddress)) {
-        /* create a route table we can search */
-        exec("/usr/bin/netstat -rnWf inet", $output, $ret);
-        foreach ($output as $line) {
-            $fields = preg_split("/[ ]+/", $line);
-            if (is_subnetv4($fields[0])) {
-                if (ip_in_subnet($ipaddress, $fields[0])) {
-                    return $fields[5];
-                }
+        $family = "inet";
+    } elseif (is_ipaddrv6($ipaddress)) {
+    	$family = "inet6";
+    } else {
+    	return false;
+    }
+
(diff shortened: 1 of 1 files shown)


src/etc/inc/interfaces.inc
```diff
@@ -3772,21 +3772,20 @@ function guess_interface_from_ip($ipaddress)
         /* create a route table we can search */
         exec("/usr/bin/netstat -rnWf inet", $output, $ret);
         foreach ($output as $line) {
-            if (preg_match("/^[0-9]+\.[0-9]+\.[0-9]+\.[0-9]+\/[0-9]+[ ]+link[#]/", $line)) {
-                $fields = preg_split("/[ ]+/", $line);
+            $fields = preg_split("/[ ]+/", $line);
```
fichtner, line 3775:
```suggestion
            $fields = preg_split("/\s+/", $line);
```

We don't know if these are tabs or spaces (should work, but please test)

src/etc/inc/interfaces.inc
```diff
@@ -3772,21 +3772,20 @@ function guess_interface_from_ip($ipaddress)
...
         }
     }
-    /* FIXME: This works from cursory testing, regexp might need fine tuning */
     if (is_ipaddrv6($ipaddress)) {
         /* create a route table we can search */
         exec("/usr/bin/netstat -rnWf inet6", $output, $ret);
         foreach ($output as $line) {
-            if (preg_match("/[0-9a-f]+[:]+[0-9a-f]+[:]+[\/][0-9]+/", $line)) {
-                $fields = preg_split("/[ ]+/", $line);
+            $fields = preg_split("/[ ]+/", $line);
```
fichtner, line 3787:
```suggestion
            $fields = preg_split("/\s+/", $line);
```

Same here I guess

fichtner:
Ok with this going in, but the flow could also be improved: if(ipv4) else (ipv6) else return otherwise we check 3 times if it's an ip

vnxme:
@fichtner Thank you for your feedback! Made changes to both the regex and the flow,

src/etc/inc/interfaces.inc
```diff
@@ -3765,34 +3765,25 @@ function get_real_interface($interface = 'wan', $family = 'all')
...
-                if (ip_in_subnet($ipaddress, $fields[0])) {
-                    return $fields[5];
-                }
-            }
-        }
+        $family = "inet";
+    } elseif (is_ipaddrv6($ipaddress)) {
+        $family = "inet6";
+    } else {
+    	return false;
```
fichtner, line 3773:
```suggestion
        return false;
```

fichtner: all spaces, no tabs :)

fichtner:
Merged, thanks! ❤️

PR #4524, opened 2020-12-14
DHCPv4 leases status: Fix sorting

vnxme:
**Issues in status_dhcp_leases.php for which this PR suggests a solution:**
- Insecure handling of the order argument
- Use of strnatcasecmp() instead of ipcmp() for sorting by ip
- Impossible to sort by interface name

**Other:**
This PR is a part of changes proposed by [#4514](https://github.com/opnsense/core/pull/4514).


fichtner:
Merged, thanks!

PR #4529, opened 2020-12-15
Widget interface_list - Remove hard coded td width

Sebclem:
I have remove hardcoded `td` width in  interface_list Widget. This allow the table to auto fit to content.
This can fix issue with long interface name.
This is my first contribution here so if I'm not doing it right, just tell me.

Before:
image

Final result: 
image

With IPv6:
image

### DIFF
--- a/src/www/widgets/widgets/interface_list.widget.php
+++ b/src/www/widgets/widgets/interface_list.widget.php
@@ -124,7 +124,7 @@ if ($_SERVER['REQUEST_METHOD'] === 'GET') {
     $ifinfo = $ifsinfo[$ifdescr];
     $iswireless = is_interface_wireless($ifdescr); ?>
   <tr id="interface_widget_item_<?= html_safe($ifname) ?>">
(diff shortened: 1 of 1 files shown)


fichtner:
No, this still looks off in most overlong scenarios. There is no fix here.

Sebclem:
@fichtner 
Is this okay if we truncate the name if it's to long ? 
I have done this, if you hover a truncated name, the full name will be shown as title: 

list


I have made the same thing with IPv6 but i don't know if it's the right solution.

Just let me know what you think about this.

marjohn56:
Go take a look at Interfaces_Plus -  #1998 in the plugins commits. Dates back to August but should still work fine. I use it on my system

Sebclem:
> @marjohn56 Go take a look at Interfaces_Plus - #1998 in the plugins commits. Dates back to August but should still work fine. I use it on my system

There is any reason that this plugin as not been merged yet ?

marjohn56:
It's down to the devs to get around to it guess, busy with more important things.

PR #4534, opened 2020-12-19
Use HTTPS everywhere (mechanical edit using util from https-everywhere)

ypid:
Ref: https://github.com/EFForg/https-everywhere

```Shell
~/src/EFForg/https-everywhere/utils/rewriter/rewriter.js .
```

A few changes where reset/fixed manually before the commit.


AdSchellevis:
@ypid thanks!

Issue #4538, opened 2020-12-20
Dnsmasq domain override not used for static hostnames


jmatraszek:
**Important notices**
Before you add a new report, we ask you kindly to acknowledge the following:

[x] I have read the contributing guide lines at https://github.com/opnsense/core/blob/master/CONTRIBUTING.md

[x] I have searched the existing issues and I'm convinced that mine is new.

**Description**
It is possible to specify domain override in Dnsmasq configuration. The specified domain is used only for dynamic DHCP leases, while for static the system domain is used.

**To Reproduce**
1. Go to `Services -> Dnsmasq DNS -> Settings`
2. Configure `Domain override`
3. Apply settings.

**Actual behavior**
```
cat /var/etc/dnsmasq-hosts
192.168.1.2   ap1.home.local ap1
[...]
# dhcpleases automatically entered
192.168.1.13  some-host.home.old some-host            # dynamic entry from dhcpd.leases
```

**Expected behavior**
Both hosts use the same domain (`home.old` in this example).

**Environment**
OPNsense 20.7.5-amd64
FreeBSD 12.1-RELEASE-p10-HBSD
OpenSSL 1.1.1h 22 Sep 2020

**Possible solution**
At first I forked `core` repo and committed [this change](https://github.com/jmatraszek/core/commit/466a71b9097cad947acdf22b46fae875d9a0da470). I didn't want to create a PR without confirming that this is a bug (and I think it is). If my solution looks OK, I am happy to create a PR.

AdSchellevis:
@jmatraszek the link (https://github.com/jmatraszek/core/commit/466a71b9097cad947acdf22b46fae875d9a0da470) returns a 404, if you want to discuss the proposed change from a PR that's also ok by the way.

jmatraszek:
Oh, sorry. I've created a PR now — #4539.

AdSchellevis:
https://github.com/opnsense/core/commit/1adefd7440efc245360290017250be4c8aa6ec2b

PR #4539, opened 2020-12-21
Use domain override for static hosts in dnsmasq-hosts

jmatraszek:
This is a proposed solution to #4538.


AdSchellevis:
@jmatraszek changed it a bit in https://github.com/opnsense/core/commit/1adefd7440efc245360290017250be4c8aa6ec2b, but it looks like you're right indeed.  can you try https://github.com/opnsense/core/commit/1adefd7440efc245360290017250be4c8aa6ec2b

jmatraszek:
@AdSchellevis Applied the patch and it works properly, Dnsmasq overrides the domain for static hosts. Thanks!

AdSchellevis:
@jmatraszek you're welcome, thanks for your help!

PR #4548, opened 2020-12-24
QueryFilter.php: allow underscore in filter string

kulikov-a:
Hi!
it is not forbidden to use the underscore sign in the filter string?
(want to use it in the search in the system log widget and system->log files)
Thanks!


AdSchellevis:
I guess I forgot about the underscore here, patch looks reasonable to me. thanks!

kulikov-a:
Thanks!

Issue #4551, opened 2020-12-26
System log widget: transit to clientside


kulikov-a:
**Important notices**
Before you add a new report, we ask you kindly to acknowledge the following:

- [x] I have read the contributing guide lines at https://github.com/opnsense/core/blob/master/CONTRIBUTING.md

- [x] I have searched the existing issues and I'm convinced that mine is new.

**Is your feature request related to a problem? Please describe.**
Hi!
At this time, the "System log" widget is executed on the server side: there is no auto-update and filtering capabilities

**Describe the solution you'd like**
Could you consider transition to the client side. With auto update and filter? 
example (with filter string validation on submit):
![syslog_wid1](https://user-images.githubusercontent.com/36099472/103157491-3bf4e780-47c4-11eb-93bb-a1f871a8f598.PNG)
![syslog_wid2](https://user-images.githubusercontent.com/36099472/103157495-4adb9a00-47c4-11eb-968b-463b0361e4e8.PNG)
![syslog_wid3](https://user-images.githubusercontent.com/36099472/103157499-53cc6b80-47c4-11eb-85e1-5f18f5732b57.PNG)

Thanks!

**Describe alternatives you've considered**

**Additional context**

PR #4552, opened 2020-12-26
system_log.widget.php: transit to client side

kulikov-a:
example for https://github.com/opnsense/core/issues/4551

### DIFF
--- a/src/www/widgets/widgets/system_log.widget.php
+++ b/src/www/widgets/widgets/system_log.widget.php
@@ -160,11 +160,7 @@ $systemlogentriesfilter = isset($config['widgets']['systemlogentriesfilter']) ? 
        $( document ).ready(function() {
         // needed to display the widget settings menu
         // need document ready to use CSRF Token - not ready at window load
-        // exit if settings button enabled (second call because of .appendTo in widgets moving function)
-
-        if(!($("#system_log-configure").hasClass( "disabled" ))) {
-          return;
-        }
+        
         $("#system_log-configure").removeClass("disabled");
         var rowCount = $("#systemlogfiltercount").val();
         var refresh_interval_ms = parseInt($("#systemlogentriesupdateinterval").val()) * 1000;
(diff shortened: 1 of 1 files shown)


fichtner:
Just a note here: such considerable replacements shouldn’t always be done in the same file. We will be dragging old files around forever even if the change is an effective rewrite.

kulikov-a:
@fichtner 
thanks! got it.
no problem to create a new widget if the team approves the idea and code )

src/www/widgets/widgets/system_log.widget.php
```diff
@@ -28,67 +29,146 @@
...
 }
+$input_errors = array();
 
-if (is_numeric($_POST['logfiltercount'])) {
-   $countReceived =  $_POST['logfiltercount'];
-   $config['widgets']['systemlogfiltercount'] = $countReceived;
-   write_config("Saved Widget System Log Filter Setting");
-   header(url_safe('Location: /index.php'));
-   exit;
+if (isset($_COOKIE['inputerrors'])) {
```
fichtner, line 40: I'm not a fan of this. Broken filter string == no output. While the idea is nice this is probably overly complicating the rework and review.

kulikov-a: Hi! Sorry, I didn't quite understand. Don't like the idea of using a filter? or implementing a check and error return via cookies?

fichtner: Hi :) Second part specifically, yes. I'm unsure about the first part but I think it can be integrated if we discuss further.

kulikov-a: just didn't want to write an arbitrary line to the config, and it will still be sanitized by api when the query is executed (and the filtering result may be unexpected). so trying to inform the user that the filter string does not meet the requirements.
couldn't find a better way to return an error for the widget.  maybe there is a better way i can try?

fichtner: ok, how about this: ideally I'd like to merge this but without the filter and we dicuss filter inclusion in next PR?

kulikov-a: sure ) let's try. I will prepare a request without a filter. thanks!

kulikov-a:
replaced by https://github.com/opnsense/core/pull/5220
(hope to come back to discussing filtering later)

PR #4554, opened 2020-12-28
Add option for webgui and ssh to configure management IP

fraenki:
### Preface

Howdy. In #4541 I've suggested to implement a _workaround_ for a problem. The discussion made me realize that this was not the best approach.

### Description
This PR contains a new solution that finally adresses the underlying issue and adds a simple option to make the management IP configurable. Let me know what you think.

Compared to #4541 this approach...

* is easily understood: users can simply switch between both behaviours
* does not increase code complexity (because it no longer manipulates the behaviour of another option/function)
* is failsafe: it is not visible by default and a clear warning is shown when selected

In case you agree to merge this new option, then I'll happily update [the documentation](https://github.com/opnsense/docs/blob/master/source/manual/settingsmenu.rst) to include a description for the new option.

##### Default: listen on specified interfaces

opn_1

#####  Optional: listen on specified IP

opn_2

(The option was added to both WebGUI and SSH, but they look the same so I've only added screenshots for the WebGUI option.)

### DIFF
--- a/src/etc/inc/plugins.inc.d/webgui.inc
+++ b/src/etc/inc/plugins.inc.d/webgui.inc
@@ -70,7 +70,7 @@ function webgui_configure_do($verbose = false, $interface = '')
     if (isset($config['system']['webgui']['managementaccess']) &&
         $config['system']['webgui']['managementaccess'] === 'ip' &&
         isset($config['system']['webgui']['ip'])) {
-        $listeners[] = $config['system']['webgui']['ip'];
+        $listeners = array($config['system']['webgui']['ip']);
     } else {
         foreach (interfaces_addresses($interfaces) as $tmpaddr => $info) {
             if ($info['scope']) {
--- a/src/etc/inc/plugins.inc.d/webgui.inc
+++ b/src/etc/inc/plugins.inc.d/webgui.inc
@@ -66,13 +66,20 @@ function webgui_configure_do($verbose = false, $interface = '')
 
     $listeners = count($interfaces) ? array() : array('0.0.0.0', '::');
 
-    foreach (interfaces_addresses($interfaces) as $tmpaddr => $info) {
-        if ($info['scope']) {
-            /* link-local does not seem to be supported */
-            continue;
-        }
+    /* bind on (all/selected) interfaces or use specified IP */
+    if (isset($config['system']['webgui']['managementaccess']) &&
+        $config['system']['webgui']['managementaccess'] === 'ip' &&
+        isset($config['system']['webgui']['ip'])) {
+        $listeners[] = $config['system']['webgui']['ip'];
+    } else {
+        foreach (interfaces_addresses($interfaces) as $tmpaddr => $info) {
+            if ($info['scope']) {
+                /* link-local does not seem to be supported */
+                continue;
+            }
 
-        $listeners[] = $tmpaddr;
+            $listeners[] = $tmpaddr;
+        }
     }
 
     chdir('/usr/local/www');
--- a/src/www/system_advanced_admin.php
+++ b/src/www/system_advanced_admin.php
@@ -719,7 +719,7 @@ $(document).ready(function() {
                 <td><a id="help_for_webguiip" href="#" class="showhelp"><i class="fa fa-info-circle"></i></a> <?= gettext("Management IP") ?></td>
                 <td>
                   <input name="webguiip" type="text" value="<?= $pconfig['webguiip'] ?>"/>
-                  <?=gettext("Only accept connections from the specified IP. Use with care."); ?>
+                  <?=gettext("Web GUI only listen on specified IP address. Use with care."); ?>
                   <div class="hidden" data-for="help_for_webguiip">
                     <?= gettext('The WebGUI only listens on the selected IP address. You may need to manually add required firewall rules first. Use with care.') ?>
                   </div>
--- a/src/www/system_advanced_admin.php
+++ b/src/www/system_advanced_admin.php
@@ -834,7 +834,7 @@ $(document).ready(function() {
                 <td><a id="help_for_sship" href="#" class="showhelp"><i class="fa fa-info-circle"></i></a> <?= gettext("Management IP") ?></td>
                 <td>
                   <input name="sship" type="text" value="<?= $pconfig['sship'] ?>"/>
-                  <?=gettext("Only accept connections from the specified IP. Use with care."); ?>
+                  <?=gettext("Only accept connections on the specified IP address. Use with care."); ?>
                   <div class="hidden" data-for="help_for_sship">
                     <?= gettext('The SSH service only listens on the selected IP address. You may need to manually add required firewall rules first. Use with care.') ?>
                   </div>
--- a/src/www/system_advanced_admin.php
+++ b/src/www/system_advanced_admin.php
@@ -836,7 +836,7 @@ $(document).ready(function() {
                   <input name="sship" type="text" value="<?= $pconfig['sship'] ?>"/>
                   <?=gettext("Only accept connections on the specified IP address. Use with care."); ?>
                   <div class="hidden" data-for="help_for_sship">
-                    <?= gettext('The SSH service only listens on the selected IP address. You may need to manually add required firewall rules first. Use with care.') ?>
+                    <?= gettext('The SSH service only listens on the specified IP address. You may need to manually add required firewall rules first. Use with care.') ?>
                   </div>
                 </td>
               </tr>
--- a/src/www/system_advanced_admin.php
+++ b/src/www/system_advanced_admin.php
@@ -806,10 +806,10 @@ $(document).ready(function() {
                 <td>
                   <select id="ssh_managementaccess" name="ssh_managementaccess" class="selectpicker">
                       <option value="interfaces" <?= empty($pconfig['ssh_managementaccess']) || $pconfig['ssh_managementaccess'] === 'interfaces' ? 'selected="selected"' : '';?>>
-                        <?=gettext("Choose Listen Interfaces");?>
+                        <?=gettext("Listen on specified interfaces (recommended)");?>
                       </option>
                       <option value="ip" <?=$pconfig['ssh_managementaccess'] === 'ip' ? 'selected="selected"' : '';?>>
-                        <?=gettext("Specify Management IP");?>
+                        <?=gettext("Listen on specified IP address");?>
                       </option>
                   </select>
                   <div class="hidden" data-for="help_for_ssh_managementaccess">
--- a/src/www/system_advanced_admin.php
+++ b/src/www/system_advanced_admin.php
@@ -408,6 +408,7 @@ $(document).ready(function() {
                      if (!result) {
                          $('#webgui_managementaccess option:selected').prop('selected', false);
                          $('#webgui_managementaccess').selectpicker('refresh');
+                         $("#webgui_managementaccess").change();
                          $.webgui_managementaccess_warned = 0;
                      }
                  }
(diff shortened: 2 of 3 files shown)


src/www/system_advanced_admin.php
```diff
@@ -636,6 +714,16 @@ function reloadWaitOld () {
                   </div>
                 </td>
               </tr>
+              <tr class="webgui_managementaccess_choices webgui_managementaccess_ip">
+                <td><a id="help_for_webguiip" href="#" class="showhelp"><i class="fa fa-info-circle"></i></a> <?= gettext("Management IP") ?></td>
+                <td>
+                  <input name="webguiip" type="text" value="<?= $pconfig['webguiip'] ?>"/>
+                  <?=gettext("Only accept connections from the specified IP. Use with care."); ?>
```
8191, line 721: > Only accept connections _from_ the specified IP.
Reads like the allowed remote IPs (i.e. source IP to connect to OPNsense). I recommend rewording this to something like
* _Only accept connections on the specified IP._
* _Management services only listen on specified IP address._

src/www/system_advanced_admin.php
```diff
@@ -636,6 +714,16 @@ function reloadWaitOld () {
...
+                  <div class="hidden" data-for="help_for_webguiip">
+                    <?= gettext('The WebGUI only listens on the selected IP address. You may need to manually add required firewall rules first. Use with care.') ?>
```
8191, line 723:
```suggestion
                    <?= gettext('The Web GUI only listens on the selected IP address. You may need to manually add required firewall rules first. Use with care.') ?>
```

src/www/system_advanced_admin.php
```diff
@@ -624,6 +686,22 @@ function reloadWaitOld () {
...
+                  <select id="webgui_managementaccess" name="webgui_managementaccess" class="selectpicker">
+                      <option value="interfaces" <?= empty($pconfig['webgui_managementaccess']) || $pconfig['webgui_managementaccess'] === 'interfaces' ? 'selected="selected"' : '';?>>
+                        <?=gettext("Choose Listen Interfaces");?>
+                      </option>
+                      <option value="ip" <?=$pconfig['webgui_managementaccess'] === 'ip' ? 'selected="selected"' : '';?>>
+                        <?=gettext("Specify Management IP");?>
+                      </option>
+                  </select>
+                  <div class="hidden" data-for="help_for_webgui_managementaccess">
+                    <?= gettext('Choose how the WebGUI should be accessible. Use with care.') ?>
```
8191, line 700:
```suggestion
                    <?= gettext('Choose how the Web GUI should be accessible. Use with care.') ?>
```

src/www/system_advanced_admin.php
```diff
@@ -624,6 +686,22 @@ function reloadWaitOld () {
...
               </tr>
               <tr>
+                <td><a id="help_for_webgui_managementaccess" href="#" class="showhelp"><i class="fa fa-info-circle"></i></a> <?=gettext('Management Access') ?></td>
+                <td>
+                  <select id="webgui_managementaccess" name="webgui_managementaccess" class="selectpicker">
+                      <option value="interfaces" <?= empty($pconfig['webgui_managementaccess']) || $pconfig['webgui_managementaccess'] === 'interfaces' ? 'selected="selected"' : '';?>>
+                        <?=gettext("Choose Listen Interfaces");?>
+                      </option>
+                      <option value="ip" <?=$pconfig['webgui_managementaccess'] === 'ip' ? 'selected="selected"' : '';?>>
+                        <?=gettext("Specify Management IP");?>
```
8191, line 696:
```suggestion
                        <?=gettext("Listen on specified interfaces (recommended)");?>
                      </option>
                      <option value="ip" <?=$pconfig['webgui_managementaccess'] === 'ip' ? 'selected="selected"' : '';?>>
                        <?=gettext("Listen on specified IP address");?>
```

Usually one configures on the settings page _what the system shall do_. Therefore one might interpret the text as "the system will chose the listen interface".

src/www/system_advanced_admin.php
```diff
@@ -725,6 +829,16 @@ function reloadWaitOld () {
                   </div>
                 </td>
               </tr>
+              <tr class="ssh_managementaccess_choices ssh_managementaccess_ip">
+                <td><a id="help_for_sship" href="#" class="showhelp"><i class="fa fa-info-circle"></i></a> <?= gettext("Management IP") ?></td>
+                <td>
+                  <input name="sship" type="text" value="<?= $pconfig['sship'] ?>"/>
+                  <?=gettext("Only accept connections from the specified IP. Use with care."); ?>
+                  <div class="hidden" data-for="help_for_sship">
+                    <?= gettext('The SSH service only listens on the selected IP address. You may need to manually add required firewall rules first. Use with care.') ?>
```
8191, line 838:
```suggestion
                    <?= gettext('The SSH service only listens on the specified IP address. You may need to manually add required firewall rules first. Use with care.') ?>
```

src/www/system_advanced_admin.php
```diff
@@ -725,6 +829,16 @@ function reloadWaitOld () {
                   </div>
                 </td>
               </tr>
+              <tr class="ssh_managementaccess_choices ssh_managementaccess_ip">
+                <td><a id="help_for_sship" href="#" class="showhelp"><i class="fa fa-info-circle"></i></a> <?= gettext("Management IP") ?></td>
+                <td>
+                  <input name="sship" type="text" value="<?= $pconfig['sship'] ?>"/>
+                  <?=gettext("Only accept connections from the specified IP. Use with care."); ?>
```
8191, line 836:
```suggestion
                  <?=gettext("Only accept connections on the specified IP address. Use with care."); ?>
```

src/www/system_advanced_admin.php
```diff
@@ -713,6 +801,22 @@ function reloadWaitOld () {
...
               </tr>
               <tr>
+                <td><a id="help_for_ssh_managementaccess" href="#" class="showhelp"><i class="fa fa-info-circle"></i></a> <?=gettext('Management Access') ?></td>
+                <td>
+                  <select id="ssh_managementaccess" name="ssh_managementaccess" class="selectpicker">
+                      <option value="interfaces" <?= empty($pconfig['ssh_managementaccess']) || $pconfig['ssh_managementaccess'] === 'interfaces' ? 'selected="selected"' : '';?>>
+                        <?=gettext("Choose Listen Interfaces");?>
+                      </option>
+                      <option value="ip" <?=$pconfig['ssh_managementaccess'] === 'ip' ? 'selected="selected"' : '';?>>
+                        <?=gettext("Specify Management IP");?>
```
8191, line 811:
```suggestion
                        <?=gettext("Listen on specified interfaces (recommended)");?>
                      </option>
                      <option value="ip" <?=$pconfig['ssh_managementaccess'] === 'ip' ? 'selected="selected"' : '';?>>
                        <?=gettext("Listen on specified IP address");?>
```

src/www/system_advanced_admin.php
```diff
@@ -362,6 +382,40 @@
...
+                     'prevent you from accessing this page if you continue. It is recommended to keep ' .
+                     'this set to the default unless you know what you are doing.')) ?>',
+                 type: BootstrapDialog.TYPE_WARNING,
+                 btnOKClass: 'btn-warning',
+                 btnOKLabel: '<?= html_safe(gettext('I know what I am doing')) ?>',
+                 btnCancelLabel: '<?= html_safe(gettext('Use the default')) ?>',
+                 callback: function(result) {
+                     if (!result) {
+                         $('#webgui_managementaccess option:selected').prop('selected', false);
+                         $('#webgui_managementaccess').selectpicker('refresh');
```
8191, line 410:
```suggestion
                         $('#webgui_managementaccess').selectpicker('refresh');
                         $("#webgui_managementaccess").change();
```

To correctly display currently relevant input field. At least in Firefox the `selectpicker('refresh')` does not trigger the _change_ event.

src/etc/inc/plugins.inc.d/webgui.inc
```diff
@@ -66,13 +66,20 @@ function webgui_configure_do($verbose = false, $interface = '')
...
-    foreach (interfaces_addresses($interfaces) as $tmpaddr => $info) {
-        if ($info['scope']) {
-            /* link-local does not seem to be supported */
-            continue;
-        }
+    /* bind on (all/selected) interfaces or use specified IP */
+    if (isset($config['system']['webgui']['managementaccess']) &&
+        $config['system']['webgui']['managementaccess'] === 'ip' &&
+        isset($config['system']['webgui']['ip'])) {
+        $listeners[] = $config['system']['webgui']['ip'];
```
8191, line 73:
```suggestion
        $listeners = array($config['system']['webgui']['ip']);
```

Otherwise the management IP is actually **added** to the array, which already contains `0.0.0.0` and `::` in case the setting `webguiinterfaces` (which is hidden in the UI when the management IP is configured) is set to "all". This would again result in listening to all IPs of the system.

src/www/system_advanced_admin.php
fraenki: Actually the wording is not consistent in OPNsense:

opnsense/core> grep -Ri 'WebGUI'|wc -l
237
opnsense/core> grep -Ri 'Web GUI'|wc -l
30

"WebGUI" is used more often, so I'd prefer this one.

src/etc/inc/plugins.inc.d/webgui.inc
fraenki: It works as intended. It _adds_ the IP to the array, but there's nothing else in the array.

src/www/system_advanced_admin.php
fraenki: Good catch, thanks.

src/etc/inc/plugins.inc.d/webgui.inc
8191: @fraenki The setting `webguiinterfaces` (which can have one interface or none (none means "all" in the GUI) is still relevant, even when your new setting `managementaccess` is set. That means if you configured "none" on the interfaces, then enable your new setting and configure a management IP, the array contains `['0.0.0.0', '::']` due to line 67.

### Detailed steps to reproduce:
1. Configure _Management Access: Interface_ and _Listen Interfaces: All_
2. Save
3. Change _Management Access_ to _IP_ and configure a _Management IP_
4. Save
5. Result: 
```
$SERVER["socket"] == "0.0.0.0:443" {
    ssl.engine = "enable"
}
$SERVER["socket"] == "[::]:443" {
    ssl.engine = "enable"
}
$SERVER["socket"] == "10.8.16.4:443" {
    ssl.engine = "enable"
}
```
```
root@opnsense-dev ~/core$ sockstat -l -p 443 -P tcp                                                                                         ✹ ✭fraenki-mgmt_ip
USER     COMMAND    PID   FD PROTO  LOCAL ADDRESS         FOREIGN ADDRESS
root     lighttpd   28935 5  tcp4   *:443                 *:*
root     lighttpd   28935 6  tcp6   *:443                 *:*
root     lighttpd   28935 7  tcp4   10.8.16.4:443         *:*
```

8191:
@fraenki Just in case you did not recognize: I added a comment in an already [resolved review comment](https://github.com/opnsense/core/pull/4554#discussion_r553792572)...
Not sure if GitHub sends out notifications for comments in resolved reviews.

src/etc/inc/plugins.inc.d/webgui.inc
fraenki: @8191 Thanks for the detailed explanation.

fraenki:
I've rebased this against current master and resolved the merge conflict.

@8191 Thanks for your review!

fichtner:
@fraenki https://github.com/opnsense/core/issues/5086 is now implemented and you can disable IP aliases for binding :)

PR #4560, opened 2020-12-30
Allow DNS resolver to skip entry on EmptyLabel

jamesgol:
A name like '.example.com' is not a valid name, but should be handled like a non existant name instead of throwing an exception

I was testing using the list from (https://raw.githubusercontent.com/nickspaargaren/no-google/master/categories/youtubeparsed) to attempt to block youtube, but the list would never load.  The first entry is '.youtubeeducation.com', which presumably pihole can deal with but it was causing the following error:

```
Dec 30 13:22:04 OPNsense /update_tables.py: fetch alias url https://raw.githubusercontent.com/nickspaargaren/no-google/master/cat
egories/youtubeparsed (lines: 1823) 
Dec 30 13:22:04 OPNsense /update_tables.py: error fetching alias url https://raw.githubusercontent.com/nickspaargaren/no-google/master/categories/youtubeparsed 
```

### DIFF
--- a/src/opnsense/scripts/filter/lib/alias.py
+++ b/src/opnsense/scripts/filter/lib/alias.py
@@ -123,7 +123,7 @@ class Alias(object):
                 for rdata in self._dnsResolver.query(address, record_type):
                     yield str(rdata)
                 could_resolve = True
-            except (dns.resolver.NoAnswer, dns.resolver.NXDOMAIN, dns.exception.Timeout, dns.resolver.NoNameservers):
+            except (dns.resolver.NoAnswer, dns.resolver.NXDOMAIN, dns.exception.Timeout, dns.resolver.NoNameservers, dns.resolver.EmptyLabel):
                 pass
 
         if not could_resolve:
--- a/src/opnsense/scripts/filter/lib/alias.py
+++ b/src/opnsense/scripts/filter/lib/alias.py
@@ -123,7 +123,7 @@ class Alias(object):
                 for rdata in self._dnsResolver.query(address, record_type):
                     yield str(rdata)
                 could_resolve = True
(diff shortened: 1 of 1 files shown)


AdSchellevis:
did you test this? (and if yes, how?) The python class doesn't seem to have a class named "dns.resolver.EmptyLabel" (https://github.com/rthalley/dnspython/blob/master/dns/resolver.py)

```
>>> import dns.resolver
>>> dns.resolver.EmptyLabel
  File "<stdin>", line 1, in <module>
AttributeError: module 'dns.resolver' has no attribute 'EmptyLabel'
```

jamesgol:
Eeek, sorry about that.  I tested it by going cowboy on my home router but then cloned the branch on another box and typo'd it.

jamesgol:
Is there no chance of getting this change merged?

AdSchellevis:
You can open a new PR, but only when it's actually  functional and tested.

jamesgol:
It was functional and tested with my last commit to the branch 7c841a8

AdSchellevis:
hmm, can you supply a use case that will trigger this exception?  Apparently your last commit was a bit lost in translation, I'll reopen so we can asses later.

jamesgol:
Sorry for the confusion, I should have commented that I fixed the issue with that commit.  I know the flood of github emails can be a bit much.

Using the list: https://raw.githubusercontent.com/nickspaargaren/no-google/master/categories/youtubeparsed
the first (non comment) entry is '.youtubeeducation.com' which causes the Exception to be thrown.

AdSchellevis:
no problem, it's quite a flood indeed, but we're on the same page now. I'll try to test and merge soon.

AdSchellevis:
@jamesgol works like a charm, thanks!

PR #4566, opened 2021-01-03
Wireless: Add WPA Enterprise Configuration for Infrastructure Mode

8191:
- Add WPA Enterprise configuration options using EAP with TLS, TTLS and PEAP
  - New _identity_ property
  - _WPA passphrase_ reworded to _WPA passphrase/EAP password_
  - Select EAP mode: TLS/TTLS/PEAP
  - Select _phase 2 algorithm_: MD5/MSCHAPv2
  - Allow configuration of CA to verify authenticity of authentication server
  - Allow configuration of certificate and private key for TLS/PEAP authentication
- As the wireless interface configuration page is quite overwhelming with all the available options, only relevant settings should be shown. Irrelevant settings are being hidden:
  - Show WPA/WEP/EAP settings only when WPA/WEP/EAP is enabled
  - Show IEEE 802.1X settings only when relevant
  - Hide AP-only/Adhoc-only settings when infrastructure mode is selected
  - Hide infrastucture-only settings when AP/Adhoc mode is selected

### DIFF
--- a/src/etc/inc/interfaces.inc
+++ b/src/etc/inc/interfaces.inc
@@ -2024,7 +2024,57 @@ function interface_wireless_configure($if, &$wl, &$wlcfg)
     /* generate wpa_supplicant/hostap config if wpa is enabled */
     switch ($wlcfg['mode']) {
         case 'bss':
+            $authentication = array();
             if (isset($wlcfg['wpa']['enable'])) {
+                if ($wlcfg['wpa']['wpa_key_mgmt'] == 'WPA-EAP' && isset($wlcfg['wpa']['wpa_eap_method'])) {
+                    switch ($wlcfg['wpa']['wpa_eap_method']) {
+                        case 'PEAP':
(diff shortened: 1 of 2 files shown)


8191:
*ping*

AdSchellevis:
We haven't been able to review this, Wifi has quite a low priority for us as there are more structural issues in there, which raises the question if we do want to expand the issue of the legacy code with new functionality or eventually rewrite it into mvc. 

If we do want to support more (even in the legacy code), it automatically raises the question of what to do with the documentation (https://docs.opnsense.org/manual/wireless.html).

8191:
@fichtner I'm happy to support here. So after you've taken a look at the legacy code and decided how to proceed (rewriting using MVC, etc.) maybe the complete WiFi topic could be split into smaller tasks which can be implemented independently.

fichtner:
Merged, thanks!

PR #4583, opened 2021-01-09
openvpn: use plugin for deferred authentication

### LINKED ISSUE #3992: Openvpn deferred authentication
It seems the same problem exist on Opnsense too...

https://redmine.pfsense.org/issues/7905

marcquark:
fixes #3992

The blocking behaviour of OpenVPN is problematic, especially in larger environments. Authentication stalls traffic for all connected users, every time somebody connects. The impact may be negligible with a small number of users or when local authentication is used, but when OPNsense relies on another server for authentication, the impact becomes worse. Depending on multiple factors, the authentication server may at times be slow to respond or not respond at all. That should not have any negative impact on already-connected users.

This PR introduces the `openvpn-auth-script` plugin to handle user authentication in a non-blocking manner. For reference, this plugin was introduced to pfSense in 2018 and has proven to work well (that's based on my personal experience and the fact that it was basically never touched since then). https://redmine.pfsense.org/issues/7905

Below is a comparison with an artificial 5-second authentication delay.

Before:
before

After:
after

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -104,6 +104,7 @@ CORE_DEPENDS?=		${CORE_DEPENDS_${CORE_ARCH}} \
 			ntp \
 			openssh-portable \
 			openvpn \
+			openvpn-auth-script \
 			opnsense-lang \
 			opnsense-update \
 			pam_opnsense \
--- a/src/etc/inc/plugins.inc.d/openvpn.inc
+++ b/src/etc/inc/plugins.inc.d/openvpn.inc
@@ -726,8 +726,9 @@ function openvpn_reconfigure($mode, $settings, $device_only = false)
                     if ($settings['strictusercn']) {
                         $strictusercn = "true";
                     }
-                    $conf .= "auth-user-pass-verify \"/usr/local/etc/inc/plugins.inc.d/openvpn/ovpn_auth_verify " .
-                        "user '{$settings['authmode']}' '{$strictusercn}' '{$mode_id}'\" via-env\n";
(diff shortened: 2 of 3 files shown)


AdSchellevis:
comments and concerns in https://github.com/opnsense/core/issues/3992#issuecomment-757451356

fichtner:
@marcquark have you had a chance to revisit this on OpenVPN 2.5 in recent OPNsense versions?

marcquark:
@fichtner i haven't, but i did test both 2.4 and 2.5 at the time of writing. In this regard, they're the same. See my comments in https://github.com/opnsense/core/issues/3992#issuecomment-757380876. Deferred Auth support [will be added in 2.6 though](https://github.com/OpenVPN/openvpn/blob/master/Changes.rst)

The plugin should work with 2.5 aswell. I don't know how long it will be until 2.6 is released and makes it into OPNsense and whether or not you'd prefer just "sitting it out" until it's here. Then again, the change isn't that complex, it could be easily reverted and replaced with a GUI switch once 2.6 arrives.

If you decide that you want to add deferred auth as a feature (or, in my humble opinion, it's more of a bugfix), i'll do my part and spin up a new testing environment to try it out and report back. Just lmk :)

AdSchellevis:
We discussed internally and decided to wait for 2.6 and revise then if we do need a deferred auth eventually, as this likely have issues with the way we generate client specific overrides (and we hope the new `auth-user-pass-verify` is backwards compatible in that sense)

PR #4584, opened 2021-01-10
pftop: fix all state value

lucasheld:
The html element of the "all" value in the pftop diagnostics view currently does not match the array value. For this reason, the option has no effect. The number of states is always set to the default value of 200.


AdSchellevis:
@lucasheld nice catch, thanks!

PR #4588, opened 2021-01-10
Firewall ACLs: Add readonly actions to "Aliases" permission

8191:
Add readonly actions to page-firewall-aliases permission:

* api/firewall/alias/export
* api/firewall/alias/getAliasUUID
* api/firewall/alias/getGeoIP


AdSchellevis:
@8191 nice catch, thanks!

8191:
@AdSchellevis Wow! That was super-fast! ;-)
Thanks

Issue #2048, opened 2018-01-03
Enhancement: Advanced OpenVPN configuration via GUI options


ashceryth:
Hi,

It would be great if it would be possible to configure some more advanced settings for OpenVPN directly via options in the GUI rather than using the "Advanced configuration" field. I am thinking of GUI options reflecting the following OpenVPN config options:

- tls-version-min (e.g. allowing only TLS 1.2)
- tls-cipher
- tls-crypt (as an alternative to tls-auth)
- ncp-ciphers
- ecdh-curve
- compress (selection of the desired compression algorithm, e.g. lz4-v2)
- explicit-exit-notify 1 (when using UDP)

Is there anything planned? Also OpenVPN running as unpriviledged user would be superb :)

fichtner:
I’ll work on this as time permits.

imidoriya:
The client export also includes "--comp-lzo" which has been depreciated.  The new method is "--compress [algorithm]".

ky41083:
+1 for this. Please ;-)

Security wise, these are very good controls to have. Some do not have a good way of implementing using the free form advanced settings box. Good as in, fully included in config backups / restores.

trendchiller:
Should the TLA-AUTH feature be replaced by TLS-CRYPT or be choosable as an alternative ?
i need this feature also very desperately, so i perhaps community could help on this (if desired) ?

ky41083:
tls-auth and tls-crypt should coexist as selectable options. One does not deprecate the other. You can only use one of them at a time. If nothing else, they should both remain as options for backwards compatibility with pre-existing setups.

fichtner:
@AdSchellevis maybe you have incentive to work on this earlier, for now I'll have to push to 19.7

AdSchellevis:
@fichtner I have no plans for this at the moment, move further is certainly fine.

Bytechanger:
Hello,
I also need   tls-crypt   option for more security.
At this time I run my openvpn-server on my raspberry!

Keltere:
+1 on tls-crypt, without it it's really a pain

galvanopus:
It would be nice to have this feature implemented.

Btw, there is a workaround:
1. Disable TLS authentication.
2. Paste the following into Advanced section:
```
<tls-crypt>
#
# 2048 bit OpenVPN static key
#
-----BEGIN OpenVPN Static key V1-----
key data here
-----END OpenVPN Static key V1-----
</tls-crypt>
```

[...]

PR #4590, opened 2021-01-11
OpenVPN: Enhance configuration - Compression

vnxme:
This PR deals with OpenVPN compression configurations for both servers and clients. The [official documentation](https://openvpn.net/community-resources/reference-manual-for-openvpn-2-4/) suggests using the option --compress instead of the deprecated --comp-lzo. It allows to use the LZ4/LZ4 v2 algorithms in addition to the LZO. Besides, in view of the [VORACLE attack](https://community.openvpn.net/openvpn/wiki/VORACLE), the stub/stub v2 algorithms are made available.

Before:
openvpn_compression_before

After:
openvpn_compression_after

This is a part of the changes requested in #2048.

### DIFF
--- a/src/etc/inc/plugins.inc.d/openvpn.inc
+++ b/src/etc/inc/plugins.inc.d/openvpn.inc
@@ -133,16 +133,19 @@ function openvpn_verbosity_level()
     );
 }
 
-function openvpn_compression_algos()
+function openvpn_compression_modes()
 {
     return array(
-        '' => gettext('Disabled - No Compression & Framing'),
-        'pfc' => gettext('Partial - Packet Framing for Compression'),
-        'lz4' => gettext('Enabled - LZ4 Algorithm'),
-        'lz4-v2' => gettext('Enabled - LZ4 v2 Algorithm'),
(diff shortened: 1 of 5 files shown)


AdSchellevis:
according to https://community.openvpn.net/openvpn/wiki/DeprecatedOptions  --compress will be deprecated as of 2.5, --comp-lzo as of 2.4, but both are still available. I don't mind supporting --compress as well, but we shouldn't drop features without a clear migration unless absolutely necessary. An option could be to keep the old compress options and start explaining which option to choose, although the general idea from the authors seem to be to stay away from both...

ky41083:
OpenVPN compression still has its use cases. To me, retaining the ability to decide on a case by case basis is important. I like the idea of adding these and briefly explaining which to choose in the GUI.

vnxme:
@AdSchellevis Thank you for your feedback. You are right that the newer option will be deprecated as of 2.5. Modified this PR to keep both options.

AdSchellevis:
@vnxme although the translators might not like the texts being replaced, it's probably a good idea to name the option explicit here since it can be a bit confusing. I'll pull it into master, thanks!

PR #4592, opened 2021-01-12
OpenVPN: Enhance configuration - TLS-Crypt

vnxme:
This PR deals with TLS authentication and encryption options in OpenVPN client/server configurations. It makes available the --tls-crypt option in addition to the --tls-auth option.

According to the [official documentation](https://openvpn.net/community-resources/reference-manual-for-openvpn-2-4/), encrypting (and authenticating) control channel packets:
- provides more privacy by hiding the certificate used for the TLS connection;
- makes it harder to identify OpenVPN traffic as such;
- provides “poor-man’s” post-quantum security, against attackers who will never know the pre-shared key (i.e. no forward secrecy).

Before:
Снимок экрана 2021-01-12 в 13 20 01

After:
Снимок экрана 2021-01-12 в 13 22 32

This is a part of the changes requested in #2048. I would be grateful if someone could test the proposed changes, especially In terms of various configuration export options.


AdSchellevis:
At a first glance this breaks people's setup since the tls option is removed, backwards compatibility is quite important to avoid issues after an upgrade. I currently don't have the time to test/review this, but I'll keep it open in case someone wants to give this a try and share her or hers findings.

ky41083:
@vnxme Does this deal with mapping the legacy checkbox to the new "Enabled - Authentication only" option, preserving backwards compatibility?

vnxme:
@AdSchellevis Agree, improved backwards compatibility.

@ky41083 Yes, the legacy checkbox is mapped to the "Enabled - Authentication only" option. Now, I think, it will not break anything. Please, let me now, if you find it does or anything else is wrong - your feedback is appreciated.

AdSchellevis:
@vnxme I'll try to review as soon as I can find a bit of time, thanks for working on this!

khicks:
Thanks a bunch for the work, @vnxme!

@AdSchellevis, is there a way we can see this in a release soon? Preferably in 21.1.5, but I would understand if you want to wait until 21.7. I'm looking to switch about 10 or so machines to OPNsense, and I would love to not have to mess with existing OpenVPN configs if I can. This is pretty much the last step to making OPNsense a drop-in replacement. Thanks for all the hard work!

trendchiller:
I was also working on this just this moment ...
Because it's also really important to me, too...
My approach is similar and I also really would appreciate this feature in the near future ;-)
I'll stop my work so far until we have a new status...

AdSchellevis:
@khicks no date yet, given the size of the PR it won't be in 21.1.5, but it's on my todo list to review and modify where needed.

vnxme:
Updated the PR to resolve a small conflict after d17e70b39ba23763753805e1db8643820bf5f8a2 (remove OpenSSL engine support #4896).

AdSchellevis:
@vnxme can you try https://github.com/opnsense/core/commit/58186f8c4ed345d0f7a830045b84355ea2247984? It should contain all your changes after some small cleanups.

vnxme:
@AdSchellevis Sorry for not getting back to you sooner. https://github.com/opnsense/core/commit/58186f8c4ed345d0f7a830045b84355ea2247984 looks good, thank you!

AdSchellevis:
@vnxme don't worry, thanks for confirming!

amezin:
@AdSchellevis When I select "Enabled - Authentication & encryption" in OpenVPN client configuration, click "Save", then open client configuration again - I see "Enabled - Authentication only" selected. And openvpn config (/var/etc/openvpn/client2.conf) contains tls-auth, not tls-crypt. OPNsense 21.7.3

fichtner:
@amezin can you try this patch?

    # opnsense-patch 761871d

trendchiller:
applying this patch does not seem to fix the problem since the GUI still switches back to "Enabled -Authentication only"

fichtner:
Great report: now I only have to emulate all that you did on multiple pages?

trendchiller:
Sorry, I do not understand the intention of the post but I try to clarify:
when I select authentication & encryption in the GUI and git save, it switches back to authentication only. (in the GUI).
in the config file that is written there is also no tls-crypt entry, only tis-auth. (as I described in the ticket #5232)
also applying the patch "# opnsense-patch 761871d" does not change anything because the GUI still switches back to authentication only. Perhaps this might be the error ?

fichtner:
For one please add the page you are using to your report. The code is in server and client page... it's easy to talk about the same thing but not do the same thing... and make sure you have the patch applied (maybe you applied it twice which removes it). I tested two times and the values now stick to the config.xml as required. Worst case verify the patch contents is correct.

trendchiller:
ok, sorry...
I use https://xxx/vpn_openvpn_server.php?act=edit&id=2
it seems as if the patch. does not apply correctly.
I'm on OPNsense 21.7.3-amd64 but when I cat vpn_openvpn_server.php my lines look different ...
anything faulty on my. side ?

fichtner:
Ok, I made one for stable/21.7:

    # opnsense-patch 98e6d76d


Cheers,
Franco

trendchiller:
Hi, this one applied successfully and the gui works now as expected. 
vpn functionality test follows when I am at home again

trendchiller:
Thanks a lot, it runs fine now !
You should definitely include this in a hotfix release - I think many people will use it.
nice work !

fichtner:
Thanks for testing. Looks like a merge error to master branch after testing went ok. Can’t explain this otherwise. Will be hotfixed but ETA unclear.

amezin:
I can confirm that 98e6d76d works for me too

Issue #2414, opened 2018-05-18
Feature request - Reset all states if WAN IP Address changes


JosQuelqu1:
Hi,

Can't find this feature on OPNsense. I need this to pull the plug on pfSense...

On pfSense:

Reset All States
Reset all states if WAN IP Address changes
This option resets all states when a WAN IP Address changes instead of only states associated with the previous IP Address.

Thank you!

fichtner:
The implementation is a bit optimistic and only works for DHCP IPv4, not particularly related to "WAN".

Although the functionality is clear, can you state your use case for reference? Maybe we can find a better spot for implementation.

JosQuelqu1:
It's for the VoIP guy here at the office. When the WAN IP change it cause problem with the VoIP service.

He sent me this. It's the exact issue: https://redmine.pfsense.org/issues/1629

topaDev:
Hi,
experiencing the same problem with VoIP, just came across this workaround (search for "resetstates.sh" script):
https://bauschp.wordpress.com/asterisk-projects/asterisk_provider_t_kom/

fichtner:
Please try this patch c5b9f672 via:

    # opnsense-patch c5b9f672


Thanks,
Franco

topaDev:
Thank you Franco, it works like a charm.

fichtner:
Ok, great, will ship this in 18.1.9. Thanks for testing!

PR #4594, opened 2021-01-13
firewall: flush states on specific interface IPv4 change

lucasheld:
This is an extention for https://github.com/opnsense/core/issues/2414.
It allows to select interfaces on which to react with a entire state table reset.
It's helpful if you have vpn clients connected that lose the connection from time to time. With the ip_change_kill_states option in this case all states are reset. Now you can check the wan interface in the ip_change_kill_states_interfaces option to ignore the vpn client reconnects.

### DIFF
--- a/src/etc/rc.newwanip
+++ b/src/etc/rc.newwanip
@@ -153,7 +153,10 @@ if (!is_ipaddr($cacheip) || $ip != $cacheip || !is_ipaddr($configip)) {
 
     system_routing_configure(false, $interface);
     plugins_configure('monitor');
-    filter_configure_sync(false, isset($config['system']['ip_change_kill_states']));
+
+    $kill_states_interfaces = !empty($config['system']['ip_change_kill_states_interfaces']) ? explode(',', $config['system']['ip_change_kill_states_interfaces']) : array();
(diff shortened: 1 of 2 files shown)


AdSchellevis:
I don't think we should do this, the current option is already confusing enough as it is. Filtering which states to reset when the current option is set might be a better idea, but we need to think this through (and test it) before doing so.

lucasheld:
Okay, I see.
`pfctl` also has the `-i interface` (Restrict the operation to the given interface) option. Maybe this is helpful but I don't get any existing states displayed with it.

```
# /sbin/pfctl -s Interfaces
all
enc
enc0
lo
lo0
pflog
pflog0
pfsync
pfsync0
vtnet0
vtnet1
#
```

```
# /sbin/pfctl -i vtnet0 -s states
#
```

AdSchellevis:
you can only do that when states are bound to interfaces, see "Bind states to interface" in firewall->settings->advanced, which isn't our default.

AdSchellevis:
probably related https://github.com/opnsense/core/pull/4726

lucasheld:
Nice, thank you!

Issue #4595, opened 2021-01-13
Unbound Blacklist: Generated dnsbl.conf starts splitting domain names when a list is over a certain amount of lines


defiantmonkey:
**Important notices**

Before you add a new report, we ask you kindly to acknowledge the following:

- [x] I have read the contributing guide lines at https://github.com/opnsense/core/blob/master/CONTRIBUTING.md
- [x] I have searched the existing issues and I am convinced that mine is new.

**Describe the bug**

When generating the dnsbl.conf of a list that contains a certain number of lines/domains (so far from the lists I've tested, seems to be around 100,000 lines), the number of entries in dnsbl.conf and the original list do not match, with some domains seeming getting split in dnsbl.conf (ie: 077ca5eaa2c83c12d5f6dd3389e9016f.redinuid.imrworldwide.com in the orginal becoming two entries in dnsbl.conf - 077ca5eaa2c83c12d5f6dd3389e9016f.redinuid. and imrworldwide.com). I only noticed this due to a bigger list I was working with somehow getting the TLD "org" in the list and did a compare using meld.

**To Reproduce**

1. Enable Unbound service and Blacklist
2. In blacklist configuration, with no other lists enabled, toggle advanced mode and enter a list in "URLs of Blacklists" such as https://raw.githubusercontent.com/lightswitch05/hosts/master/docs/lists/ads-and-tracking-extended.txt
3. Save and let the downloader run
4. Go to logs and filter for download
5. See that lines/records from download are greater in number than the number of domains from the original list
6. Copy both original list and dnsbl.conf, stripped to only the domain names and sorted, into a compare tool like meld then see that some domains from dnsbl.conf are different/not present in the original list

**Expected behavior**

The dnsbl.conf is populated with an equal or lesser number of entries without domains having been split in the process. 

**Screenshots**

N/A

**Relevant log files**

Here's the log output from testing with only this list being downloaded: https://raw.githubusercontent.com/lightswitch05/hosts/master/docs/lists/ads-and-tracking-extended.txt

unbound[87006] | blacklist  download  https://raw.githubusercontent.com/lightswitch05/hosts/master/docs/lists/ads-and-tracking-extended.txt  (lines: 271128 exclude: 2 black: 271102
unbound[87006] | blacklist download done in 1.86 seconds (271102 records)

**Additional context**

N/A

**Environment**

Software version used and hardware type if relevant, e.g.:

OPNsense 20.7.7_1 (amd64, LibreSSL 3.1.5).
[...]

PR #4606, opened 2021-01-15
download_blacklists.py: fix broken lines in large files

kulikov-a:
proposed fix for https://github.com/opnsense/core/issues/4595
also changed "if chop[-1]" to "if parts[-1] != "\n":" (checking an array is more correct than checking an incoming string?)
second "parts.pop()" in line65 just for array  cleanup (so as not to yield "\n" only).
checked on https://raw.githubusercontent.com/lightswitch05/hosts/master/docs/lists/ads-and-tracking-extended.txt and
https://blocklistproject.github.io/Lists/fraud.txt (without ending "\n")

### DIFF
--- a/src/opnsense/scripts/unbound/download_blacklists.py
+++ b/src/opnsense/scripts/unbound/download_blacklists.py
@@ -58,10 +58,11 @@ def uri_reader(uri):
                 break
             else:
                 parts = (prev_chop + chop).split('\n')
-                if len(parts) > chop.find('\n'):
-                     prev_chop = parts.pop()
+                if parts[-1] != "\n":
+                    prev_chop = parts.pop()
                 else:
                     prev_chop = ''
+                    parts.pop()
                 for part in parts:
                     yield part
     else:
--- a/src/opnsense/scripts/unbound/download_blacklists.py
+++ b/src/opnsense/scripts/unbound/download_blacklists.py
@@ -62,7 +62,6 @@ def uri_reader(uri):
                     prev_chop = parts.pop()
                 else:
                     prev_chop = ''
-                    parts.pop()
                 for part in parts:
                     yield part
     else:


src/opnsense/scripts/unbound/download_blacklists.py
```diff
@@ -58,10 +58,11 @@ def uri_reader(uri):
                 break
             else:
                 parts = (prev_chop + chop).split('\n')
-                if len(parts) > chop.find('\n'):
-                     prev_chop = parts.pop()
+                if parts[-1] != "\n":
+                    prev_chop = parts.pop()
                 else:
                     prev_chop = ''
+                    parts.pop()
```
AdSchellevis, line 65: @kulikov-a this looks a bit weird, why do we need to remove the last line here?

kulikov-a: in line 65 parts array last memebr is "\n". do we need to yield it?

AdSchellevis: well, you can chop it if it's empty. e.g.

```
("xxx\nyyy\naaa\n" + "iii\ntttt\n").split("\n")
```
vs.
```
("xxx\nyyy\naaa\n" + "iii\ntttt").split("\n")
```

in the last case it would chop "tttt"  if I'm not mistaken

kulikov-a: in first case we have full string, `parts[-1] == "\n"`, so `prev_chop = ''` and we dont need to yield `"\n"` ?
in second case we do not know if the whole line is received, so we pass this to `prev_chop` and `.pop()` it from `parts`.
line 65 can be can be safely removed.  work well with or without it. `\n` will be filtered at `tmp = line.split('#')[0].split()`.
just didn't want to yield "empty" data to `line`

AdSchellevis: accidentally yielding an empty attribute likely doesn't mind, since the content is filtered anyway. I would remove the pop to be honest, it's less obvious what it's doing from the code.

kulikov-a: checked again. sorry. you're right. its '' not '\n' inf first case (my bad). still works (removing empty member) but absolutely useless.
removing..

AdSchellevis: no problem, thanks for taking care of this

AdSchellevis:
@kulikov-a thanks!

kulikov-a:
Thanks!!

PR #4612, opened 2021-01-17
vpn_openvpn_server.php: netbios options prettify

kulikov-a:
Hi!
NetBIOS options in server config looks a bit ugly (especially in dark themes).
Can we add a couple of breaks to prettify it?
before:
ovpn_nb1

after:
ovpn_nb2

Thanks!

### DIFF
--- a/src/www/vpn_openvpn_server.php
+++ b/src/www/vpn_openvpn_server.php
@@ -1466,7 +1466,7 @@ endif; ?>
                         <div id="netbios_data">
                           <span>
                             <?=gettext("Node Type"); ?>:&nbsp;
-                          </span>
+                          </span><br>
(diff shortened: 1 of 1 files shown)


AdSchellevis:
hmm, I think officially it should have been a label for the input tag, if I'm not mistaken these are formatted properly (have tried yet)

kulikov-a:
in my opinion this is CSS: select's is display: inline-block; inputs is display: block;
so it doesn't look very good with current formatting (input and select in same div)

AdSchellevis:
let's use your fix, the rest of the page handles these issues in similar matter and label has the same issue indeed. thanks!

kulikov-a:
Thanks!
fast as always )

PR #4625, opened 2021-01-22
traffic.volt: remove row by intreface filter val

kulikov-a:
Hi!
just small fix for dont wait TTL to remove filtered interface in top talkers
ref: https://forum.opnsense.org/index.php?topic=20700.msg97995#msg97995


AdSchellevis:
@kulikov-a changed it a bit in https://github.com/opnsense/core/commit/c455ebbe6e1a38b76988cfa7ff407597186b8254, technically we don't need to check the output of /api/diagnostics/traffic/top/  since it already was fed with the interfaces. Better to keep maintenance in the same function. thanks!

kulikov-a:
@AdSchellevis 
thanks!
yes, I had to add a check, because to avoid an unnecessary dom loop (hate dom loops)), i deleted rows on select change (and after that a deferred ajax response with the old set of interfacers is arrived).
I will consider next time
thanks again!

Issue #4025, opened 2020-04-09
IPv6 not working when using 6rd with a /64 prefix


paride:
**Describe the bug**

All what follows is on OPNsense 20.1.4 (i386/OpenSSL).

My ISP supports IPv6 using 6rd but sadly only assigns a single /64. However this should be enough to have radvd/SLAAC on the LAN interface, and indeed it works fine with the router provided by the ISP. But let's proceed by steps.

I can properly set up 6rd on the WAN interface. I need to manually add `wan_stf` as the IPv6 default route to workaround https://github.com/opnsense/core/issues/3903, but then the tunnel works fine:

<pre>
root@net6501:~ # ifconfig wan_stf
wan_stf: flags=4041<UP,RUNNING,LINK2> metric 0 mtu 1280
	inet6 <b>2001:b07:5d2e:58c7:: prefixlen 64</b> 
	nd6 options=1<PERFORMNUD>
	v4net 93.46.88.199/0 -> tv4br 81.208.50.214
	groups: stf 

root@net6501:~ # route -6 get default
   route to: default
destination: default
       mask: default
        fib: 0
  interface: wan_stf
      flags: <UP,DONE,STATIC>
 recvpipe  sendpipe  ssthresh  rtt,msec    mtu        weight    expire
       0         0         0         0      1280         1         0 

root@net6501:~ # ping6 -c1 ipv6.google.com
PING6(56=40+8+8 bytes) 2001:b07:5d2e:58c7:: --> 2a00:1450:4002:805::200e
16 bytes from 2a00:1450:4002:805::200e, icmp_seq=0 hlim=57 time=7.416 ms
</pre>

The LAN interface is configured to track the WAN interface, with IPv6 prefix ID 0x0 (IIUC as I have a single /64 this is the only choice). The interface configuration looks correct: the LAN interface get assigned the `:1` address of the 6rd /64:
<pre>
bridge0: flags=8843<UP,BROADCAST,RUNNING,SIMPLEX,MULTICAST> metric 0 mtu 1500
	ether 02:9d:ee:f2:88:00
	inet 192.168.1.1 netmask 0xffffff00 broadcast 192.168.1.255 
	inet6 <b>2001:b07:5d2e:58c7::1 prefixlen 64</b> 
	inet6 fe80::29d:eeff:fef2:8800%bridge0 prefixlen 64 scopeid 0xa 
	nd6 options=1<PERFORMNUD>
	groups: bridge 
[...]

PR #4635, opened 2021-01-25
fix issue #4025: 6rd with single /64 prefix

mhofer117:
Builds upon the 6rd routing fixes discussed in https://forum.opnsense.org/index.php?topic=20260.0

Instead of setting the calculated /64 subnet length on the _stf interface, I set the original ISP provided subnet length.
And change the gateway to be inside the ISP provided prefix instead of the calculated /64.

wan address will still be the same but the wider subnet solves any routing issues with single /64 prefixes.


fichtner:
Merging as discussed, thank you ❤️

marjohn56:
Got something for you to have a look at. I'll be on IRC tomorrow morning, DHCPDv6 needed a rethink. :)

PR #4636, opened 2021-01-25
Monit: minor bugfixes and UI changes

8191:
* Do not render "stop command" of services to config if not configured
* Add missing service description to model and view
* Clarify various mail server settings in settings view
* Fix spelling in settings view


AdSchellevis:
@8191 looks good, thanks!

PR #4640, opened 2021-01-26
firewall: Fix Icon Toggle for Block

jeffg2k:
When a rule is set to Allow, Block or Reject, the icon is `fa-play`, `fa-times` and `fa-times-circle` respectively.  When you toggle a rule to disable it, the code in `firewall_rules.php` changes the icon to `fa-play` (Allow Icon) regardless of it being Block or Reject, which is confusing.  

Due to the order of how things are defined in the CSS, Reject and Allow work fine, but Block does not.  Since `fa-spinner fa-pulse` appears to be defined later in CSS than the indicators, it takes priority for the icon.  When removed, it restores the proper icon with the appropriate `text-muted` or `text-success`.  I think we can safely remove the `fa-play` from the toggle, which will fix the Block icon.

Example of issue by clicking it a couple times..
Screen Shot 2021-01-26 at 12 05 34 AM

### DIFF
--- a/src/www/firewall_rules.php
+++ b/src/www/firewall_rules.php
@@ -437,7 +437,7 @@ $( document ).ready(function() {
   $(".act_toggle").click(function(event){
       event.preventDefault();
       let target = $(this);
-      target.removeClass('fa-play').addClass('fa-spinner fa-pulse');
+      target.addClass('fa-spinner fa-pulse');
       let id = target.attr("id").split('_').pop(-1);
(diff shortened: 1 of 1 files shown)


jeffg2k:
Alternatively, check which class exists and add remove / add the appropriate class.

AdSchellevis:
@jeffg2k I totally forgot about this one, noticed the same a long time ago but forget. This seems to work fine on my end, so let's pull this in. Thanks!

PR #4653, opened 2021-01-31
Update alias.py to add User-Agent

abraha2d:
This is needed since some web servers (including `os-nginx` in its default configuration) reject requests without a User-Agent specified.

Not sure how to dynamically get the current OPNsense version, or whether that's even a good User-Agent. Feel free to modify to fix that.

### DIFF
--- a/src/opnsense/scripts/filter/lib/alias.py
+++ b/src/opnsense/scripts/filter/lib/alias.py
@@ -140,6 +140,9 @@ class Alias(object):
         req_opts['url'] = url
         req_opts['stream'] = True
         req_opts['timeout'] = self._timeout
+        req_opts['headers'] = {
+            'User-Agent': 'OPNsense/20.7.8_4',
+        }
         if self._ssl_no_verify:
             req_opts['verify'] = False
         # fetch data


AdSchellevis:
when not specified a standard one from python is appended if I'm not mistaken, we should try to pin a specific version of ours in there.

fichtner:
@AdSchellevis read from metadata? "Product/Major version" only I would say

AdSchellevis:
@fichtner maybe better find out why the default user agent is an issue, before you know you end up glueing all scripts with custom user-agents. The default headers should already contain an agent as far as I know.

https://github.com/psf/requests/blob/589c4547338b592b1fb77c65663d8aa6fbb7e38b/requests/utils.py#L821-L826

abraha2d:
Just checking in on this, any updates?

fichtner:
No, as Ad said the default is set so the question is why and where it is stripped of you really don’t have it and still need it.

In my case hard coding doesn’t help. So this PR should rather be a support ticket...

fichtner:
As stated before we cannot hardcode this in any case so I will close the PR.

PR #4657, opened 2021-02-01
Fix sidebar menu collapse

maurice-w:
Sidebar menu collapsed when clicking link to netflow configuration.


fichtner:
Nice catch, thanks!

Issue #24, opened 2015-01-09
Add template engine in check_reload_status and give check_reload_status a more sensible name


AdSchellevis:
One of the main goals for OPNsense is to create a cleaner codebase which is easier to maintain. This feature detaches frontend and backend for template generation and provides a cleaner approach for handling configuration files.
Rename check_reload_status.py to a more sensible name. (as it is the backend service for configuring the firewall)

AdSchellevis:
Template engine in place, rename check_reload_status later on.

PR #4659, opened 2021-02-01
Prefix delegation fix for dhcpdv6

marjohn56:
Currently any PD delegation is incorrectly using the same prefix as the parent interface. For the avergage user who does not need PD on thier LAN this is not an issue, it  will cause issues for those who do want to use dhcpdv6 and give out prefixes. It is quite complex to do as the PDs will vary depending on the prefix obtained from  the ISP and the delegation size being given out, not a simple 'just add '1;

I said this  is quite complex, first we call a new function gen_ipv6_prefix_ranges($interface), the interface being a lan interface, we only need to call this once no matter how many lan interfaces are tracking a particular wan, there is code around this call to make sure it is only called once.
Within the function gen_ipv6_prefix_ranges($interface), we first call another new function get_tracking_interfaces_count($interface) which returns the total number of lan interfaces that are tracking the  wan tracked by the interface in the call to gen_ipv6_prefix_ranges($interface). We need this value to calculate the pd prefixes and ranges. The call returns an array containing the number of interfaces tracking the wan, the prefix delegation supplied by the ISP, this is obtained by using the new call get_isp_pdinfo($wanif, $type = 'len') - (that uses the real prefix obtained from dhcp6c), and the wan interface name.

So now we have the those we can work out the size of the prefix needed to fit the number of lan interfaces plus the number of PDs we wish to give out.

 Once we have this, say for instance we have two lan interfaces tracking a wan, we calculate the number of total prefixes needed, in the case of two lan interfaces we need one prefix for the /64s of the lans themselves and two pd ranges for the delegations, that would make three prefixes total, so the minimum PD prefix would need to support four prefixes  ( we cannot have odd numbers ). Next we call another new function ipv6_prefix_calc($ipaddrv6, $pd_mask, $prefix), this takes the original lan ip, the pd_mask we have calculated from the size of the ISP supplied prefix and the number of PDs we need and the prefix supplied by the ISP. This returns an array containing the prefixes themselves, i.e in the case of us needing a total; of eight prefixes and we have a /56 supplied by the ISP we would need a /59 mask and the prefixes would be aaaa:bbbb:cccc::,   aaaa:bbbb:cccc:20::  aaaa:bbbb:cccc:40:, aaaa:bbbb:cccc:80: etc etc.

I will not go through the whole process, but the result is that the call returns an array indexed by the tracking IDs of the lan interfaces. We then just pass assign the prefixes before it gets written to the conf.

There is the additional bonus that we can use the call ipv6_prefix_calc() to generate a prefix table that could be used to assist a user who is doing manual configuration.

This will require the use of a dhcp6c version that supports returning the PD_INFO in the form wan_interface.prefix, dhcp6c PR #24 does this.

I should add a note that it doesn't matter what size the ISP prefix is, the code will generate the correct prefixes, however I have limited the range to work between /48 and /64. I don't think ISPs will be doing dhcpd6 on anything larger than /48!

### DIFF
--- a/src/etc/inc/plugins.inc.d/dhcpd.inc
+++ b/src/etc/inc/plugins.inc.d/dhcpd.inc
@@ -503,8 +503,9 @@ function dhcpd_radvd_configure($verbose = false, $blacklist = array())
             if ($autotype == 'slaac') {
                 /* XXX also of interest in the future, see hardcoded prefix above */
                 $radvdconf .= "\t\tBase6Interface $realtrackif;\n";
+                $radvdconf .= "\t\tDeprecatePrefix on;\n";
             }
-            $radvdconf .= "\t\tDeprecatePrefix on;\n";
+            /* XXX DeprecatePrefix on crashes radvd on 12.1 */
             $radvdconf .= "\t\tAdvOnLink on;\n";
             $radvdconf .= "\t\tAdvAutonomous on;\n";
             $radvdconf .= "\t};\n";
@@ -1282,6 +1283,10 @@ function dhcpd_dhcp6_configure($verbose = false, $blacklist = array())
         flush();
     }
 
+    $interface_count = 0;
+    $pd_one_pass = 0;
+    $check_wanif_array = array();
+
     /* we add a fake entry for interfaces that are set to track6 another WAN */
     foreach (array_keys($iflist) as $ifname) {
         /* Do not put in the config an interface which is down */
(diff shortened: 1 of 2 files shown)


marjohn56:
Looks like there was a change with deprecate prefix on the radvd section between the time I created the PR and the master pull. So please adjust as neccasary.

marjohn56:
Forgot about multiwan! Now it handles that.

marjohn56:
Additions are available to display prefix allocations for each interface, but I will not commit those until this has been pulled, waste of time otherwise.

PR #4663, opened 2021-02-03
Unbound: fix domain overrides for private address reverse lookup zones

### LINKED ISSUE #4787: Unbound: Issues with domain overrides for private reverse lookup zones

**Describe the bug**

- Domain overrides for IPv6 ULA reverse lookup zones don't work.
- Domain overrides for IPv4 RFC1918 reverse lookup zones only work if DNS rebinding checks are enabled.

Patch in #4663.

**To Reproduce**

1. Go to 'Services: Unbound DNS: Overrides'
2. Add Domain Overrides for ULA and RFC1918 reverse lookup zones like `d.f.ip6.arpa` and `10.in-addr.arpa`, save and apply
3. Perform reverse lookups for addresses in these zones (`drill -x fd00:2001:db8::1` / `drill -x 10.1.2.3`)
4. See error: RFC1918 query gets forwarded to configured server, ULA query doesn't
5. Go to 'System: Settings: General' and disable DNS Rebinding Checks
6. Repeat the reverse lookup for the RFC1918 address
7. See error: Query doesn't get forwarded

**Expected behavior**

Domain overrides should work for all reverse lookup zones and independently of DNS rebinding checks being enabled or not.

None.

**Additional context**

Forum thread: https://forum.opnsense.org/index.php?topic=21208

**Environment**

OPNsense 21.7.a_115 (amd64, OpenSSL)

maurice-w:
Domain overrides for reverse lookup zones add a typetransparent local-zone to unbound.conf. This allows forwarding of reverse lookups of private IP addresses which would otherwise be blocked by ~~DNS rebinding prevention~~ unbound's default AS112 zones.

These entries were only created for IPv4 reverse lookup zones (.in-addr.arpa.), but not for IPv6 reverse lookup zones (.ip6.arpa.).

Closes #4787.

### DIFF
--- a/src/etc/inc/plugins.inc.d/unbound.inc
+++ b/src/etc/inc/plugins.inc.d/unbound.inc
@@ -487,7 +487,7 @@ function unbound_add_domain_overrides($pvt = false)
         if ($pvt == true) {
             $domain_entries .= "private-domain: \"$domain\"\n";
             $domain_entries .= "domain-insecure: \"$domain\"\n";
-            if (preg_match('/.+\.in-addr\.arpa\.?$/', $domain)) {
+            if (preg_match('/.+\.(in-addr|ip6)\.arpa\.?$/', $domain)) {
                 $domain_entries .= "local-zone: \"$domain\" typetransparent\n";
             }
         } else {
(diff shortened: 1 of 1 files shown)


maurice-w:
There is another bug: The 'local-zone' is only added if DNS rebinding checks are enabled. On the other hand, 'private-domain' entries are not required for reverse zones.

The second commit should fix that.

https://forum.opnsense.org/index.php?topic=21208

maurice-w:
I've been using the patches for about a month and haven't seen any side effects. Feedback welcome.

fichtner:
Merged, thanks!

Issue #4500, opened 2020-12-04
firmware: modernize check for update et al


fichtner:
- [x] Extend JSON parsing of opnsense-version for non-core pacakges
- [x] Support console-bound firmware audits (similar to what rc.firmware.subr is doing)
- [x] Provide health and security audits from the console (through option 12)
- [x] ~~Base update can reroot, saving a reboot. Packages install can probably continue without the need for a reboot (unless it is a major OS upgrade maybe).~~ No, it's still buggy.
- [x] Add an overview page for currently installed system information -- the current page gives too little context about its current firmware state
- [x] Scrape business ID from mirror URL
- [x] Plugins are installed without upgrading to the latest version. Maybe we should enforce the latest version check to make sure this is all going fine -- the backend is already wrapped neatly to do this
- [x] Except the backend thinks that install will always run fine, make sure to check if it did indeed go ok.
- [x] UX rework of the missing plugin registration / dismissal
- [x] ~~Small cosmetic bug on the side: reopening the GUI page makes "upgrades" run as "execution" in the blue box which might be misleading~~ The page no longer prints these.
- [x] Improve metadata for plugins as well -- expand to JSON, add product_id at least
- [x] Register misconfigured plugins during resync run when we have JSON data for plugins
- [x] ~~Changing to an unknown release type should fail earlier in opnsense-update~~ The only bug to find was stripping the vital flag when the type target package was not found (pkg-fetch does not error at all)
- [x] Reload the menu after performing an upgrade (\*\*\*DONE\*\*\*) so that we have the new menu structure rendered after e.g. installing a new package.
- [x] api/core/firmware/changelog/N/A lookup is silly during release type transition
- [x] Add business release type and rename production to community
- [x] Changelog URL to business mirror on matching release type
- [x] Firmware connectivity audit (pkg update -f to see for users at least)
- [x] Show configured repositories on status page
- [x] ~~When lighttpd keeps running we get lots of errors while files are being removed and added again. One option would be to take down the server for the time or at least block access to the PHP scripts (a 503 will probably do fine as well)~~ Fixed the timeout trigger instead and we will see how that goes
- [x] Filter for packages and plugins tab
- [x] The check.sh run is prone to timeout errors via configd and we can't see its progress -- interpreting some of the values already but actually seeing them might make more sense. Thus we move the check to the backend and interpret the output via existing call but funnel the check result through the new product status.
[...]

PR #4684, opened 2021-02-07
Sidebar - fix for too long a-link list

opnsenseuser:
if there is not enough space in a list of links upwards, the list moves down by the number of menu entries. so the problem with too many entries should now be solved. I hope that suits you. if so, please test. Thank you

### DIFF
--- a/src/opnsense/www/js/opnsense_theme.js
+++ b/src/opnsense/www/js/opnsense_theme.js
@@ -62,7 +62,7 @@ $(document).ready(function () {
                 close_submenu(this);
                 if (currentHeight > (winHeight - li_itemH)) {
                     var divPos = ((divHeight - li_itemH) > divTop) ? -(((divHeight - li_itemH) - divTop)) : 3;
-		    var viewresult = that.next('div').css('margin-top', -divHeight - (that.is(layer1_a) ? divPos : 0));
+                    var viewresult = that.next('div').css('margin-top', -divHeight - (that.is(layer1_a) ? divPos : 0));
                 }
             }
         },
--- a/src/opnsense/www/js/opnsense_theme.js
+++ b/src/opnsense/www/js/opnsense_theme.js
@@ -45,7 +45,7 @@ $(document).ready(function () {
     footH = $('.page-foot').height(),
     headerH = $('.navbar').height(),
     li_itemH = $('a.list-group-item').height(),
-    navHeight = (countA * li_itemH) + ((footH + headerH) - (20 + countA)),
+    navHeight = (countA * parseInt(li_itemH)) + ((footH + headerH) - (20 + countA)),
     events = {
         mouseenter: function () {
             $('#navigation.col-sidebar-left').css('width', '415px');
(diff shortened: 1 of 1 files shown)


opnsenseuser:
grafik

kulikov-a:
Hi
may i ask which editor are you using? all spaces are replaced with tabs?
as a result, the entire file is perceived as new -  need to read the entire file in search of 2 lines of changes)

kulikov-a:
I have the opposite problem
menu
can the `(currentHeight > winHeight)` condition be slightly changed?

opnsenseuser:
if there is not enough space "down" it should already pop "up" and if there is even not enough space it should now move down by the number of menu entries!

opnsenseuser:
i think 

grafik 

is the problem. you can´t read the last line because the https...... overlaps it.

opnsenseuser:
@kulikov-a i changed this

kulikov-a:
@opnsenseuser 

> you can´t read the last line because the https...... overlaps it.

yes and not only. even when the status line disappears (although this is already inconvenient) the last menu item (in this case Web Proxy) is half hidden behind the browser window. Therefore, I propose to pop the menu up if the "current" and "win" heights differ by less than the height of one menu item

>  i changed this

thanks!!

opnsenseuser:
if I did not miscount, then 34 plugins are now mutually exclusive with a resolution of 1080. that should be enough for now. 
grafik

opnsenseuser:
@fichtner any change to get this into master? Or do you think we need a better solution? Only alternative way is to use a scrollbar. But i want to avoid that because it would be more usability friendly. And compared to freenas or pfsense it would be something different in the operation. What do you think about this? regards rené

opnsenseuser:
I am still of the opinion that the services could also be divided. So make subdivisions in the service folder according to e.g. DNS services, proxy services, tools, etc. Regardless of this, the list at opnsense and also at pfsense is far too long and should be divided. But that's just my point of view.

AdSchellevis:
@opnsenseuser let's try this on master and see how it goes. thanks

opnsenseuser:
For what reason was the update not adopted in the current version of opnsense?

fichtner:
The rule of thumb is core changes go on development release for one cycle.

opnsenseuser:
ok, i see...thx

opnsenseuser:
@fichtner I have a question and please don't be angry. I just wanted to know whether the sidebar will go online with the next release? Please don't get my question wrong. If not, I just wanted to know if I had to change something or made a mistake or the sidebar doesn´t work properly and this is the reason that you didn't release the update. 
i would just like to know, because maybe i can improve something in the future!

Thanks for the info and have a nice day - best regards, rené

fichtner:
We wanted to do the larger rework of #4500 first to finish the proper menu reload even when sidebar is active. A number of vars are attached to the old menu so after reload nothing works anymore as you noted yourself.

opnsenseuser:
@fichtner Is there anything i can do to help you?

Issue #2946, opened 2018-11-20
DHCP server dynamic IP allocation with Host Name and Description fields from Static info


Northguy:
I configured OPNsense as DHCP server and would like to identify my devices in the DHCP leases table a bit more easily, based on a description. Most devices only show generic manufacturer info under the MAC address, but do not provide a Hostname or a more verbose description.

I tried creating a static mapping for the MAC address but am not necessarily looking for a static IP lease. Therefore I did not fill in the IP address in the edit box. According to the help it says "If no IPv4 address is given, one will be dynamically allocated from the pool", but if I follow this way, the MAC address is shown as ' static'  in the leases table without an IP. i.e.: I have created a static lease without a visible dynamic IP?

Is this a bug, or is there another way to create an easily identifiable indicator for the MAC addresses that have obtained a dynamic lease? 

Issue seems (not entirely) related to https://github.com/opnsense/core/issues/1465

zz0rk:
I am in the same boat and would also appreciate some feedback from an expert. I've set up several static DHCP mappings and they show as 'active' when in use under Leases; but no IP address is being populated for each entry row.

ja-finn:
I was looking for the same. In my old Asus router I could just type in a new hostname and I'm guessing it tied it to the MAC. It's really nice to have for devices that you can't set a hostname on (ie IoT).

Further it seems that setting a static entry without defining an IP address leaves you unable to edit the entry afterwards. No edit button or thrascan to remove it. 

OPNsense 19.1.1

![screenshot from 2019-02-08 09-54-46](https://user-images.githubusercontent.com/13936979/52468007-a75d4e00-2b87-11e9-9307-dddae8f346db.png)

teklabania:
Here's some more formal BUG report for the issue.

# Brief description:
What I want to achieve is to translate MAC address to hostname without caring about IP address assigned to the host.

# OPNSense: 19.1.6

# Steps to reproduce
 - go to Services -> Unbound DNS -> General
 - set "Register DHCP static mappings" to true
 - go to Services -> DHCPv4 -> Interface
 - scroll down to section "DHCP Static Mappings for this interface."
 - click "+" to add new Static DHCP Mapping	
 - Under "Static DHCP Mapping" provide:
   - MAC address
   - Hostname
 - click Save
 - go to Services -> DHCPv4 -> Leases

# Results
 - IP from the pool is leased to MAC address
 - No hostname is assigned to IP address
[...]

PR #4689, opened 2021-02-09
DHCP: Update DNS with hostname only static entries

gwjo:
When a DHCP static entry is created with a hostname but no IP address, Unbound DNS is not update when an IP address from the DHCP pool is assigned to the client.

As discussed in #2946 the issue is that the hostname reported in the `client-hostname` field in dhcpd.leases is the hostname reported by the client.  This change adds the configured hostname to a _set_ variable, `hostname-override`, that gets recorded in dhcpd.leases, then use this value if preset to update the DNS server record.

This only works with the Unbound DNS server, but it could be extended to work with dnsmasq by updating sbin/dhcpleases to parse `hostname-override`

### DIFF
--- a/src/www/services_dhcpv6_edit.php
+++ b/src/www/services_dhcpv6_edit.php
@@ -207,7 +207,6 @@ include("head.inc");
                       <input name="hostname" type="text" value="<?=$pconfig['hostname'];?>" />
                       <div class="hidden" data-for="help_for_hostname">
                         <?=gettext("Name of the host, without domain part.");?>
-                        <?=gettext("If no IP address is given above, hostname will not be visible to DNS services with lease registration enabled.");?>
                       </div>
                     </td>
                   </tr>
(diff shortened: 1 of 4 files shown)


AdSchellevis:
@gwjo took some time to review, but looks practical. Thanks!

src/www/services_dhcpv6_edit.php
```diff
@@ -207,7 +207,6 @@
                       <input name="hostname" type="text" value="<?=$pconfig['hostname'];?>" />
                       <div class="hidden" data-for="help_for_hostname">
                         <?=gettext("Name of the host, without domain part.");?>
-                        <?=gettext("If no IP address is given above, hostname will not be visible to DNS services with lease registration enabled.");?>
```
fichtner, line 210: @gwjo @AdSchellevis might not work for dhcpv6 just yet. we don't have parity between newer unbound dhcpv4 reader and the legacy readers dhcpleases and dhcpleases6. Looking closely at dhcpleases6 it looks like it doesn't do anything at all?

https://github.com/opnsense/ports/blob/master/opnsense/dhcpleases/files/dhcpleases6.c

fichtner: Ok it does one thing only...run a given command https://github.com/opnsense/ports/blob/d92b0a3845d4a6441054fd25012e28eb1a11a8b0/opnsense/dhcpleases/files/dhcpleases6.c#L559

Issue #3343, opened 2019-03-16
Router Advertisements: 'Advertise Default Gateway' setting not working


maurice-w:
**Describe the bug**
Enabling 'Advertise Default Gateway' in the 'Router Advertisements' settings adds a `::/0` route to RAs, disabling it removes that route. This is not the correct way of (not) advertising a default gateway: Receipt of **any** RA with Router Lifetime > 0 will cause hosts to install a default route to the RA's source address, so advertising a `::/0` route is redundant at least. This also means that disabling 'Advertise Default Gateway' does not stop hosts from installing a default route (because the Router Lifetime in RAs is still > 0).

**To Reproduce**
1. Go to 'Services' / 'Router Advertisements' / 'LAN'
2. Enable 'Advertise Default Gateway' and save
3. Do a packet capture on the LAN interface and have a look at the RAs. They contain route information for `::/0` (which is redundant).
4. Go to 'Services' / 'Router Advertisements' / 'LAN'
5. Disable 'Advertise Default Gateway' and save
6. Check whether hosts in the LAN still use OPNsense as their default gateway. Spoiler: They do, which means that this setting does not have the desired effect.
7. Do a packet capture on the LAN interface and have a look at the RAs. They don't contain route information for `::/0` (good), but the Router Lifetime is > 0 (bad).

**Expected behavior**
- RAs should never contain route information for `::/0`.
- Disabling 'Advertise Default Gateway' should set the Router Lifetime to 0.

**Additional context**
[RFC 4861 (Neighbor Discovery for IP version 6 (IPv6))](https://tools.ietf.org/html/rfc4861)
> A router might want to send Router Advertisements without advertising itself as a default router. For instance, a router might advertise prefixes for stateless address autoconfiguration while not wishing to forward packets. Such a router sets the Router Lifetime field in outgoing advertisements to zero.

**Environment**
OPNsense 19.1.4-amd64

fichtner:
@maurice-w please double-check e67dade

maurice-w:
Thanks @fichtner, works as expected after applying the patch to 19.7.a_701 and re-saving the RA config.

fichtner:
@maurice-w Thanks for testing. Side question, does the CARP feature mentioned in the patch make sense to you? (using RemoveRoute off / DeprecatePrefix off)

maurice-w:
[...]

PR #4692, opened 2021-02-10
radvd: don't advertise ::/0 route

maurice-w:
As discussed back in #3343: Router Advertisements should never contain a ::/0 route. A default route is indicated by setting the Router Lifetime to > 0. This was originally fixed by @fichtner in e67dade991d631d7edba5b750a87396b38bac2e1, but the ::/0 route is still added for CARP interfaces and "static mode" interfaces (introduced in 66dc0e9b29ec3d7fbe2446faf057116e6a9b6337).

I only noticed this after recently switching to static interface mode.

### DIFF
--- a/src/etc/inc/plugins.inc.d/dhcpd.inc
+++ b/src/etc/inc/plugins.inc.d/dhcpd.inc
@@ -355,12 +355,6 @@ function dhcpd_radvd_configure($verbose = false, $blacklist = array())
             $radvdconf .= "\t};\n";
         }
 
-        if (!empty($dhcpv6ifconf['rainterface'])) {
-            $radvdconf .= "\troute ::/0 {\n";
-            $radvdconf .= "\t\tRemoveRoute off;\n";
-            $radvdconf .= "\t};\n";
-        }
-
         if (!empty($dhcpv6ifconf['raroutes'])) {
             foreach (explode(',', $dhcpv6ifconf['raroutes']) as $raroute) {
                 $radvdconf .= "\troute {$raroute} {\n";


fichtner:
Just to chase this down I looked up the original commit: https://github.com/pfsense/pfsense/commit/dc131dfe2d

To me it looks reasonable as previously discussed. Routes are still sticky in CARP so this shouldn't cause any trouble. Thanks!

maurice-w:
The commit history just before the one you referenced really tells a story. Seems they tried to prevent the default route from flapping in case of a CARP failover.

- First, 'DeprecatePrefix off' was added. Makes sense for CARP, but is unrelated to the default route.
- Then, 'RemoveRoute off' was added to the advertised prefix. Which doesn't work because this is a route option, not a prefix option.
- When realising that, the interface's prefix was additionally advertised as a route and 'RemoveRoute off' moved there. Which makes zero sense (pun intended).
- Finally, that route was changed to ::/0. Which still does nothing, but it seems it was just left like this ever since.

Is there a good English word for "verschlimmbessern"?

fichtner:
https://www.dict.cc/?s=verschlimmbessern

"disimprove" klingt gar nicht mal schlecht :)


Grüsse
Franco

marjohn56:
A word rarely used...  I had to look up the word..

image

I think.. "just made matters worse" is the polite way of saying it.

fichtner:
1820's... those were the days... and "just made matters worse" sounds very rude in British context, doesn't it? ;)

marjohn56:
Not at all, If you had a wound, and I rubbed salt into it, then you might say "That's just made matters worse". Of course I would nit rub salt into an open wound. 😎

PR #4694, opened 2021-02-10
proxy fetchACLs: Change close() semantics

gregveldman:
Under current close() semantics for a DomainSorter() object, more general (e.g. higher level) domains are evaluated after subdomains.  This resulted in a curious situation in which if ads.foo.bar was in a blacklist under one category and foo.bar was in the same blacklist under a different category, the resulting ACL file would have:

```
.ads.foo.bar
foo.bar
```

Which would result in any subdomain other than ads being accessible.  Change the semantics so that higher level domains are considered first and prefixed with `.`, and if any subdomains are subsequently encountered they are just skipped.

### DIFF
--- a/src/opnsense/scripts/proxy/fetchACLs.py
+++ b/src/opnsense/scripts/proxy/fetchACLs.py
@@ -229,7 +229,7 @@ class DomainSorter(object):
                     break
                 else:
                     set_content[line.split('|')[0]] = '|'.join(line.split('|')[1:])
-            for itemkey in sorted(set_content, reverse=True):
+            for itemkey in sorted(set_content, reverse=False):
                 yield set_content[itemkey]
 
     @staticmethod
@@ -262,9 +262,12 @@ class DomainSorter(object):
                         # duplicate, skip
                         continue
                     if self.is_domain(line):
-                        # prefix domain, if this domain is different then the previous one
-                        if prev_line is None or '.%s' % line not in prev_line:
+                        # write prefix for domain, but only if it is not a
+                        # subdomain of the previous entry - if so, skip
+                        if prev_line is None or '.%s' % prev_line not in line:
                             f_out.write(b'.')
+                        else:
+                            continue
                     f_out.write(line.encode())
                     prev_line = line
 


AdSchellevis:
that's the intended behaviour, if a sub domain is defined, it's not going to match all from the top level. (most of these lists don't really distinct between wildcards and explicit matches, which results in unexpected matches for quite some of them).

gregveldman:
The proposed change doesn't add the higher level domain if only a subdomain is blocked.  Rather, it addresses an incomplete condition in the current scheme.  Consider the following actual example.

In the Shalla list, `ads.facebook.com` is blocked as an advertising domain:
```
$ grep facebook.com BL/adv/domains 
ads.ak.facebook.com
ads.facebook.com
```

At the same time, `facebook.com` is blocked as a social networking site:
```
$ grep facebook.com BL/socialnet/domains 
facebook.com
facebook.com.br
facebook.com.tr
```

Using the current algorithm, the resulting ACL file contains:
```
.ads.facebook.com
facebook.com
```

This leaves users behind a filtering Squid, using the ACL as generated, free to visit the site `www.facebook.com` if the ACL is generated using both categories above.  In other words, the generated ACL is incomplete for purpose.

Under the proposed change, the ACL file would contain simply `.facebook.com` (note the preceding `.` here and its absence above), which blocks users from visiting `facebook.com`, `www.facebook.com`, `ads.facebook.com`, and any other subdomain.  In other words, the filter is now complete, and the ACL matches on anything in the facebook.com domain.  If the local admin chose to filter only the `adv` category and not `socialnet`, there would be no change: the resulting ACL would only contain `.ads.facebook.com` and users would still be free to visit `facebook.com`, `www.facebook.com`, or any other subdomains of that site.

AdSchellevis:
The problem is that the original intend of the domain (wildcard or not) isn't clear, which means some higher level domains are being blocked unintended when always assuming wildcard. We might consider this type of change if a lot of people report similar issues and there seems to be a general understanding that all domains are intended as wildcards and there's enough test coverage to prevent future complains. 

It has been quite some time that I build this script and if I remember correctly the above logic has side affects.

gregveldman:
Fair enough.  Feel free to close this then if you like.

PR #4707, opened 2021-02-15
trim carp password. without this the untrimmed password will be used …

fhloston:
…on master firewall and the trimmed password on secondary firewall after config sync - resulting in dual master due to password mismatch.

### DIFF
--- a/src/www/firewall_virtual_ip_edit.php
+++ b/src/www/firewall_virtual_ip_edit.php
@@ -140,6 +140,8 @@ if ($_SERVER['REQUEST_METHOD'] === 'GET') {
                    $input_errors[] = sprintf(gettext("VHID %s is already in use on interface %s. Pick a unique number on this interface."),$pconfig['vhid'], convert_friendly_interface_to_friendly_descr($pconfig['interface']));
                }
             }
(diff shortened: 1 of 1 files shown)


fichtner:
This makes no sense. Never trim passwords. What actually went wrong... the config sync really trimmed the password?

fhloston:
This caused me some hours of debugging, yes.

I do not really care how it gets fixed, as long as it gets fixed and this cannot happen again ;)

`      <password> 1234</password>`
on master leads to
`      <password>1234</password>`
on slave.

AdSchellevis:
We do care how things are fixed..... https://github.com/opnsense/core/commit/a65e793cb48df26be9fa5d9bc5338221e0c081da removing leading/trailing spaces to solve an edge case causes the next person searching for hours why their data is mangled.

fhloston:
> We do care how things are fixed..... [a65e793](https://github.com/opnsense/core/commit/a65e793cb48df26be9fa5d9bc5338221e0c081da) removing leading/trailing spaces to solve an edge case causes the next person searching for hours why their data is mangled.

Thank you! 

Please excuse my bluntness!

PR #4713, opened 2021-02-17
LDAP bugfix: use StartTLS as configured when retrieving authentication containers

### LINKED ISSUE #4453: LDAP server configuration - "Authentication containers" select
Hi there,

I got a small question about LDAP configuration with STARTTLS.

When configuring LDAP Server with STARTTLS, the select "Authentication containers" return nothing
image

This is because OPNSense bind to the LDAP server in TCP standard mode, even if STARTTLS is selected
image

image

Other than that, LDAP authentication works fine for both webui and OpenVPN login with STARTTLS
image

Is it a normal behavior or am I missing something ?

cbrueffer:
Make StartTLS work when retrieving LDAP authentication containers. The code did not set the LDAP connection properties as configured and thus connected as if `TCP - Standard` was selected.

Fixes #4453 (currently erroneously already closed)


AdSchellevis:
@cbrueffer nice catch, thanks!

Issue #4557, opened 2020-12-29
Unbound DNSBL - Add option to use local-zone with refuse instead of redirecting everything to localhost via local-data


defiantmonkey:
**Important notices**
Before you add a new report, we ask you kindly to acknowledge the following:

[x] I have read the contributing guide lines at https://github.com/opnsense/core/blob/master/CONTRIBUTING.md

[x] I have searched the existing issues and I'm convinced that mine is new.

**Is your feature request related to a problem? Please describe.**
Hi, new to OPNsense (and BSD in general), so forgive me if I sound dumb (though I probably am dumb given I'm new to a lot of this).

I mainly got OPNsense installed for a replacement of my local DNS server (plan to do more with it in the future), in particular I wanted to make use of the blocking feature in Unbound.

After testing it out, I noticed that when the dnsbl.conf file is generated everything is redirected to localhost. At a base level this works fine for blocking requests, but I also want to be able to go into the logs and see what was blocked. Currently it will just show that the request succeeded.

**Describe the solution you'd like**

I achieved this locally by modifying line 140 of https://github.com/opnsense/core/blob/master/src/opnsense/scripts/unbound/download_blacklists.py on my install so that it now looks like this:

unbound_outf.write("local-zone: \"%s\" refuse\n" % entry)

which lets me simply search the logs for the word "refuse" to see what is actually being blocked.

As I am quite new at doing this kind of stuff, this may not even be a good way of doing this, but it would be nice if a change could be made so that tracking blocked requests through the logs/statistics is possible on future releases.

Edit: After further discussion, I realized that others may have existing configs working with local-data entries, so instead of a hard change, a toggle/option to use local-zone with refuse instead of local-data redirect (probably through the web UI) would be a better solution.

**Describe alternatives you've considered**
I don't know of any alternative solutions so far.

EDIT: as discussed below, a current workaround for those who wish to stick to the web UI or have existing configs working with the local-data entries is to use the suricata alerts (Intrusion Detection service) and enable/disable the alert as needed. 

**Additional context**
Version: OPNsense 20.7.7_1-amd64

defiantmonkey:
Edited the initial comment, I'm guessing this was marked incomplete because I didn't follow that exact template?

kulikov-a:
@defiantmonkey 
sorry, and what if the client have backup (second) dns server configured?

defiantmonkey:
@kulikov-a 
[...]

PR #4715, opened 2021-02-17
unbound dnsbl: allow to create inform local-zones

kulikov-a:
Hi!
for https://github.com/opnsense/core/issues/4557#issuecomment-780304743
the idea is to add the ability to create inform-type local zones for blacklisted hosts (by default unbound creates transparent zones implicitly in this case). as a result, requests for such hosts are displayed in the log with a record like:
` info: bad.host.com. inform 192.168.0.215@60748 bad.host.com. A IN`
and can be filtered by ' inform ' keyword.
may close https://github.com/opnsense/core/issues/4557
thanks!


fichtner:
@kulikov-a can you respin this patch if you think we should still have it? thanks.

kulikov-a:
@fichtner thanks! 
yes, I'll try. at least worth discussing imho)

PR #4716, opened 2021-02-18
fw_log.volt: show rows with 'or' and empty filter

kulikov-a:
Hi!
table is empty if "Select any of given criteria (or)" selected and filter is empty


AdSchellevis:

kulikov-a:
Thanks!

PR #4723, opened 2021-02-19
Correct static address on an address where dhcpdv6 entry is an override

### LINKED ISSUE #4712: Static DHCPv6 Mapping for dynamic interface reporting wrong on DNS
- [x ] I have read the contributing guide lines at https://github.com/opnsense/core/blob/master/CONTRIBUTING.md
- [x ] I have searched the existing issues and I am convinced that mine is new.

**Describe the bug**
The DHCP6 service is running on a tracking interface. Every now and then the IP provider would assign a new prefix (f.i. 2a02:xxxx:xxxx:xx00/56). Tracking is working (local interface prefix might be 2a02:xxxx:xxxx:xx94/64). Still, I need to assign some static mappings on that interface. The in-page help states, that in that case the IP address can be specified by just giving the static part, i.e. omitting the prefix (f.i. ::40). Btw.: if I'm not mistaken, DHCPv6 is missing from the online documentation.

Indeed, the host will get the IP address assigned (f.i. 2a02:xxxx:xxxx:xx94::40). However, when looking up the hostname, DNS would return the IP6 address exactly as entered into OpnSense (f.i. ::40). The host cannot be looked up and addressed like that, since the prefix is missing on the reported IP address.

**To Reproduce**
On a tracking interface:
1. navigate to Services | DHCPv6 [<ifcname>]
2. for some host add a static mapping
3. specify the hosts IP address in the form ::zzzz, where zzzz should not be in the DHCP server's pool
4. verify that the host is picking up the address after some time (you might want to reduce lease times to few minutes just for the test)
5. nslookup hostname, the returned IP6 address will be just '::zzzz'

**Expected behavior**
The DNS lookup should return a valid IP6 address, i.e. the combination of the interface's prefix + the static mapping address.

With current behaviour there is no other choice that specifying a complete IP6 address in the static mapping. When the provider assigns a new prefix, the static mappings must be updated manually.

**Screenshots**
-/-

**Relevant log files**
-/-

**Additional context**
Dual stack connection.
See also: https://forum.opnsense.org/index.php?topic=14356.msg66214#msg66214

**Environment**
OPNsense 21.1.1-amd64, but I'm facing this problem since I'm using IP6 (about 4 years).

marjohn56:
Correct the omission of the prefix when a dynamic interface v6 dhcpdv6 address is set as a static entry. The result is that only the suffix /64 gets added to the unbound host entries. This patch should sort that so a full address is created for unbound.

Resolves #4712

### DIFF
--- a/src/etc/inc/plugins.inc.d/unbound.inc
+++ b/src/etc/inc/plugins.inc.d/unbound.inc
@@ -681,6 +681,13 @@ function unbound_add_host_entries()
                         continue;
                     }
 
+                   // XXX - When using manual overide on dhcpd6 we need to get the prefix and marry it to the suffix 
+                   // Get the prefix first.
+                   list ($ipaddrv6) = interfaces_primary_address6($dhcpif);
+                   if (!empty($ipaddrv6)) {
+                      $host['ipaddrv6'] = make_ipv6_64_address($ipaddrv6,$host['ipaddrv6']);
+                   }
+
                     $domain = $config['system']['domain'];
                     // XXX: dhcpdv6 domain entries have been superseded by domainsearchlist,
                     //      for backward compatibilty support both here.


src/etc/inc/plugins.inc.d/unbound.inc
```diff
@@ -681,6 +681,16 @@ function unbound_add_host_entries()
                         continue;
                     }
 
+                   // XXX - When using manual overide on dhcpd6 we need to get the prefix and marry it to the suffix
+                   // Check if the static entry begins with '::' , if it does it must be a manual overide entry
+                   if(substr($host['ipaddrv6'],0,2) == '::') {
```
fichtner, line 686: any reason why we are not using make_ipv6_64_address()? :)

marjohn56:
Yes, I don't know every function that exists! :)

Wait till you take a look at my PD correction PR, pretty sure I've duplicated lots of things! :)

fichtner:
didn't you create the function? ;)

In general the construct we already have looks safer than looking for "::"...

https://github.com/opnsense/core/blob/028b129f8715fe72b071a88026599fd7e846a9ea/src/etc/inc/plugins.inc.d/dhcpd.inc#L1560-L1562

marjohn56:
I may well of written it, but I've been to sleep since then.

marjohn56:
Yes, looks like that's a safer bet, but the fact the suffix will always start with ::  when an override is used is also true :)

fichtner:
And ::1  ...

marjohn56:
Yes, but you'd still  be adding the prefix :)

marjohn56:
I'll change the commit, gimme two

fichtner:
Fair point. At least this way we can say it's not an intentional bug.

marjohn56:
Try that.

marjohn56:
I have tested it and it write the correct entries in host_entries.conf

src/etc/inc/plugins.inc.d/unbound.inc
```diff
@@ -681,6 +681,13 @@ function unbound_add_host_entries()
                         continue;
                     }
 
+                   // XXX - When using manual overide on dhcpd6 we need to get the prefix and marry it to the suffix 
+                   // Get the prefix first.
+                   list ($ipaddrv6) = interfaces_primary_address6($dhcpif);
+                   if (!empty($host['ipaddrv6'])) {
```
fichtner, line 687: You probably meant this?

```suggestion
                   if (!empty($ipaddrv6)) {
```

src/etc/inc/plugins.inc.d/unbound.inc
```diff
@@ -681,6 +681,13 @@ function unbound_add_host_entries()
                         continue;
                     }
 
+                   // XXX - When using manual overide on dhcpd6 we need to get the prefix and marry it to the suffix 
+                   // Get the prefix first.
+                   list ($ipaddrv6) = interfaces_primary_address6($dhcpif);
+                   if (!empty($ipaddrv6])) {
```
fichtner, line 687: A stray ] is still in there

marjohn56:
Nuts!

fichtner:
Merged, thanks!

Issue #4733, opened 2021-02-21
No IPv6 address assigned on WAN interface / Router Advertisement ignored


marjohn56:
**Important notices**

Before you add a new report, we ask you kindly to acknowledge the following:

- [ x] I have read the contributing guide lines at https://github.com/opnsense/core/blob/master/CONTRIBUTING.md
- [x ] I have searched the existing issues and I am convinced that mine is new.

**Describe the bug**

IPv6 ICMPv6 RS does not get sent over a PPPoE link.

**To Reproduce**

You'll need to set up a PPPoE server! Reason being the interface needs to come up on PPPoE v4 before you can see anything. Once you have done this then using Wireshark monitor the ICMPv6 traffic being sent to the PPPoE  server, there is no RS being sent.

**Expected behavior**
ICMPv6 RS packets should be sent over the PPPoE link

OPNsense 21.1.r_96-amd64FreeBSD 12.1-RELEASE-p12-
HBSDOpenSSL 1.1.1i 8 Dec 2020
Intel(R) Core(TM) i5-5250U CPU @ 1.60GHz (4 cores)

Solution is to not use the -a flag on RTSOLD as that prohibits RS being sent on point to point interfaces ( according to the man ), but to supply the actual interfaces on the command instead.

fichtner:
https://github.com/opnsense/src/commit/2485dc0351

fichtner:
For the record, the PPP limitation came from KAME in 2000 to FreeBSD via https://github.com/freebsd/freebsd-src/commit/804c83d4bd3 with the "-a" option. Since we force PPPoE interfaces one way or another and the auto-probe just adds devices to the argument list like we would pass it we are good here. Thanks for the pointers to the problem.

PR #4734, opened 2021-02-21
Fix ICMPv6 RS packets not being sent over  PPPoE

marjohn56:
RTSOLD when using the -a flag will not send solicits over a point to point link. You must therefore specify the interfaces to be probed.

Re: #4733

### DIFF
--- a/src/etc/inc/interfaces.inc
+++ b/src/etc/inc/interfaces.inc
@@ -2892,9 +2892,19 @@ function interface_dhcpv6_configure($interface = 'wan', $wancfg)
 
     /* always kill rtsold in case of reconfigure */
     killbypid('/var/run/rtsold.pid', 'TERM', true);
+    /* XXX issue with rtsold not sending on point to point interfaces. We need to get a list of outgoing interfaces and append them */
+    /* to the command rather than using the -a option. Let's get a list of all active outgoing interfaces ( WAN connections ) */
+    /* Might want more than just dhcp6 and slaac ? */
+
+    foreach (config_read_array('interfaces') as $if => $ifconf) {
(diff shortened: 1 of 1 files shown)


fichtner:
We are not going backwards on this. Not after doing all the work to unclog this code for multi WAN DHCPv6.

marjohn56:
It's not going backwards, you have to send a solicit... HTH is it going backwards?????

marjohn56:
To add to my last comment, my test router is set up with multiwan dhcp6, one direct to the primary router and one over a PPPoE link to the Microtik VM, both if them get v6 prefixes and the LANs get the correct prefixes and it works fine. It does NOT work with the -a flag..

glasid:
> 
> 
> We are not going backwards on this. Not after doing all the work to unclog this code for multi WAN DHCPv6.

Seriously? What kind of strange approach is that?

As marjohn56 pointed out, RTSOLD does not send requests to a point-to-point interface when the -a flag is used. Just read the RTSOLD man page!

It has been shown that the current implementation is flawed for users using a PPP WAN link. And now you don't want to fix it?

fichtner:
There is a lot of drama for a silly one line restriction. https://github.com/opnsense/src/commit/2485dc035

marjohn56:
No drama, just trying to make things work as they should.

glasid:
> 
> 
> There is a lot of drama for a silly one line restriction. [opnsense/src@2485dc0](https://github.com/opnsense/src/commit/2485dc035)

Rather than fixing the problem at the point where it arises, we are now tinkering with rtsold's source code.
That's insane!

fichtner:
If you knew anything of the work Martin and me have done over the years you would probably not act in this ignorant stance of yours. Time for you to go with it or just go. I don’t care.

PR #4736, opened 2021-02-22
Correct dhcp6c.conf issue on pppoe link down

marjohn56:
When a PPPoE down event happens the dhcp6c.conf file does not get re-written thus leaving the config section for that link. This results in errors from dhcp6c and possibly delays when the link is re-established,
It happens because the interface_bring_down returns when it's a PPPoE link without calling an update using interface_dhcpv6_prepare() and a HUP to dhcp6c.

### DIFF
--- a/src/etc/inc/interfaces.inc
+++ b/src/etc/inc/interfaces.inc
@@ -1033,6 +1033,25 @@ function interface_bring_down($interface = "wan", $ifacecfg = false)
         kill_hostapd($realif);
     }
 
+    $track6 = link_interface_to_track6($interface);
+    if (count($track6)) {
+        /* bring down radvd and dhcp6 on these interfaces */
+        plugins_configure('dhcp', false, array('inet6', $track6));
+    }
+
+    switch ($ifcfg['ipaddrv6']) {
+        case 'slaac':
+        case 'dhcp6':
+            interface_dhcpv6_prepare($interface, $ifcfg, true);
+            killbypid('/var/run/dhcp6c.pid', 'HUP');
+            break;
+        case 'track6':
+            interface_track6_configure($interface, $ifcfg, false, true);
+            break;
+        default:
+            break;
+    }
+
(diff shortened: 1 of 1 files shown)


marjohn56:
Issue #4737

src/etc/inc/interfaces.inc
```diff
@@ -1050,6 +1050,10 @@ function interface_bring_down($interface = "wan", $ifacecfg = false)
                         break;
                     }
                 }
+                if ( $ifcfg['ipaddrv6'] == 'slaac' || $ifcfg['ipaddrv6'] == 'dhcp6' ) {
+                  interface_dhcpv6_prepare($interface, $ifcfg, true);
+                  killbypid('/var/run/dhcp6c.pid', 'HUP');
+                }
                 return;
```
fichtner, line 1057: what if we remove the return instead and let the code below handle this?

fichtner: second thought: maybe it would be better to kill ipv6 first (switch the blocks)

marjohn56: In answer to the first comment, I'd need to run it up and see what happens; in answer to the second, explain, switch which blocks?

There's another REALLY weird issue I came across this morning. I'm on IRC

fichtner: IPv4 and IPv6 destroy blocks... I mean we set up IPv4 then IPv6, so bring down should kill IPv6 first, then IPv4?

In that case the return wouldn't be harmful because we already called the proper IPv6 destruction code. Unsure which is better, maybe both (switch blocks, no return) would be the most favourable.

I'm chasing down the origin of the "return" now to be sure.

fichtner: Ok, it's one of ours 31282787dfb7 ... need to discuss internally

fichtner:
Merged, thanks!

PR #4738, opened 2021-02-22
Correct help text for Range

marjohn56:
Range help refers to WAN, should be LAN, added ( PD ) to make it even more obvious,

### DIFF
--- a/src/www/services_dhcpv6.php
+++ b/src/www/services_dhcpv6.php
@@ -503,7 +503,7 @@ if (isset($config['interfaces'][$if]['dhcpd6track6allowoverride'])) {
                         </table>
                         <div class="hidden" data-for="help_for_range">
                             <?= gettext("When using a static LAN address, the range should be entered using the full IPv6 address. " .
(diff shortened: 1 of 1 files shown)


marjohn56:
#4739

src/www/services_dhcpv6.php
```diff
@@ -502,8 +502,8 @@ function show_netboot_config() {
                           </tbody>
                         </table>
                         <div class="hidden" data-for="help_for_range">
-                            <?= gettext("When using a static WAN address, the range should be entered using the full IPv6 address. " .
-                            "When using a dynamic WAN address, only enter the suffix part (i.e. ::1:2:3:4)."); ?>
+                            <?= gettext("When using a static LAN address, the range should be entered using the full IPv6 address. " .
+                            "When using a dynamic LAN (PD) address, only enter the suffix part (i.e. ::1:2:3:4)."); ?>
```
fichtner, line 506: Maybe so?

```suggestion
                            "When using a delegated LAN address, only enter the suffix part (i.e. ::1:2:3:4)."); ?>
```

fichtner:
Merged, thanks!

PR #4742, opened 2021-02-23
Unbound: Fix bad dereference when DHCP lease expires

### LINKED ISSUE #4741: Unbound updates break after DHCP expires

**Describe the bug**

My change to fix #4714 (1b530ff38f24e2b2be9a0d70f5b640fdd3e51a51) causes the python script to exit when the DHCP lease expires.  

You can see in the code block below, that on line L134 `cached_leases[address]` is accessed after it was deleted on L133

https://github.com/opnsense/core/blob/1b530ff38f24e2b2be9a0d70f5b640fdd3e51a51/src/opnsense/scripts/dns/unbound_dhcpd.py#L132-L137

I'll open a new PR, but wanted to point this out before someone else runs into it.

**Note**

When confirming this was the cause, I noticed that the `local_data_remove` call on L132 doesn't appear to be correct.  This just uses the client address, and not the full FQDN for the client.  I'll do some more testing, but this doesn't look like it does not delete the RR record.

gwjo:
Move deletion of dhcp cache info when the lease expires to after the
last reference to the cache information.

Fixes #4741

### DIFF
--- a/src/opnsense/scripts/dns/unbound_dhcpd.py
+++ b/src/opnsense/scripts/dns/unbound_dhcpd.py
@@ -130,10 +130,10 @@ def run_watcher(target_filename, domain):
                         "dhcpd expired %s @ %s" % (cached_leases[address]['client-hostname'], address)
                     )
                     unbound_control(['local_data_remove',  cached_leases[address]['client-hostname']])
-                    del cached_leases[address]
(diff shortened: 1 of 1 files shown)


AdSchellevis:
@gwjo oops, missed that one as well, thanks for fixing it.

Issue #4750, opened 2021-02-24
New traffic widget broken in Firefox in Version 21.1.2


borkedporcupine:
- [x] I have read the contributing guide lines at https://github.com/opnsense/core/blob/master/CONTRIBUTING.md
- [x] I have searched the existing issues and I am convinced that mine is new.

**Describe the bug**

After updating to version 21.1.2 the Traffic Graph widget on the dashboard does not display the graphs while using Firefox.  Appears to work without issue in other browsers.  No errors are appearing in the developer console in Firefox. 

**To Reproduce**

1. Open Firefox browser
2. Log into web gui
2. Add traffic widget if it's not already on dashboard
3. View broken widget.

**Expected behavior**

Expect widget to contain graphical information as it does in other browsers. 

**Screenshots**

![Traffic Graph in FF](https://user-images.githubusercontent.com/79605150/109062415-e8b0ef80-76b5-11eb-9576-e62da42c698a.JPG)

**Additional context**

**Environment**

OPNsense 21.1.2-amd64

**Forum Link**

https://forum.opnsense.org/index.php?topic=21726.0

fichtner:
@borkedporcupine thanks for the ticket! I assume 8b9764fa86175 does not help in your case? If so which Firefox version are you using and which dashboard widgets do you have enabled?


Cheers,
Franco

kulikov-a:
changing to  $(window).on("load" to  $( document ).ready works good on my firefox.
with calling on 'load' canvas gets height: 0px on firefox

IsaacFL:
I copied this from Firefox console this morning. Firefox 86 on Windows 10. I don't know if this will helpful or not.

[FireFox 86 Console Output.txt](https://github.com/opnsense/core/files/6051605/FireFox.86.Console.Output.txt)

kulikov-a:
@IsaacFL 
i think that "undefined variables" issued was fixed:
https://forum.opnsense.org/index.php?topic=21726.msg102444#msg102444
after that the console should be quiet imho but widget can be empty some times

PR #4752, opened 2021-02-25
Dashboard: cut scripts from widgets before move

kulikov-a:
Hi.
This can help to get rid of double event bindings in widgets when manipulating DOMs.
Also fixes https://github.com/opnsense/core/issues/4750 on my test vm without changing to $( document ).ready


AdSchellevis:
@kulikov-a seems to work like a charm, the dashboard is a bit hackery, I don't expect we can easily fix this otherwise at the moment. Thanks!

kulikov-a:
@AdSchellevis  as always without holidays and weekends)
Thanks!!

Issue #4753, opened 2021-02-25
IPS alerting disabled rules - part 2


mnaiman:
Continuation of **https://github.com/opnsense/core/issues/4725**

Software version used and hardware type if relevant, e.g.:
![image](https://user-images.githubusercontent.com/2264442/109186535-4f570b80-7791-11eb-9324-e69950434926.png)

1. Install package os-intrusion-detection-content-snort-vrt 
2. Configure rules 
![image](https://user-images.githubusercontent.com/2264442/109185758-98f32680-7790-11eb-8467-8be3a602ddfe.png)
3. Enable all rules
4. Create policy 
![image](https://user-images.githubusercontent.com/2264442/109186051-da83d180-7790-11eb-8429-6cef6b2c1576.png)
5. Verify, that rule is disabled by policy - ok
![image](https://user-images.githubusercontent.com/2264442/109186168-f12a2880-7790-11eb-8abf-cc7d76497a5f.png)
6. Wait for alerts
![image](https://user-images.githubusercontent.com/2264442/109186313-1454d800-7791-11eb-932a-64074fc17dc5.png)

This should not happened. If it is disabled it should not appear at all.

Any idea?

kulikov-a:
can confirm that policies with target_action = disabled not commenting rules in opnsense.rules files out.
rules cache is correct (enabled = 0). quick check replacing `== '0'` with != '1' in
https://github.com/opnsense/core/blob/70f856bf2fa9d2b1e5fc11e0b7e5bbfae04be92e/src/opnsense/scripts/suricata/installRules.py#L69
did the trick

AdSchellevis:
got it, looks like a type error 
https://github.com/opnsense/core/blob/b465a418a0f9e17b7f73cc9dafb1cb4571cd6fb7/src/opnsense/scripts/suricata/lib/rulecache.py#L105-L110

mnaiman:
wow, now it works as expected, thanks

PR #4754, opened 2021-02-26
IDS/IPS: disable rules by Policy Action

kulikov-a:
Hi!
rule_updates[rule_info_record['metadata']['sid']]['enabled'] return 0 for` __manual__` disabled rules and False for disabled rules with matched policy.
fixes https://github.com/opnsense/core/issues/4753 on my test vm


kulikov-a:
Fixed by @AdSchellevis in the only correct way )

Issue #4675, opened 2021-02-05
certs.inc: SAN is appended to DN


kulikov-a:
**Important notices**

Before you add a new report, we ask you kindly to acknowledge the following:

- [x] I have read the contributing guide lines at https://github.com/opnsense/core/blob/master/CONTRIBUTING.md
- [x] I have searched the existing issues and I am convinced that mine is new.

**Is your feature request related to a problem? Please describe.**

Hi!
When CSR or internal cert is created on OPN, the SAN is attached to the certs subject
![SANinDN](https://user-images.githubusercontent.com/36099472/107043743-7777ce00-67d4-11eb-92b7-29c477d636a0.PNG)

![SANinDN](https://user-images.githubusercontent.com/36099472/109256608-5c164680-7807-11eb-82fc-d965ba7b1e67.PNG)

**Describe the solution you like**

imho SAN should not be part of the subject
Is this how it was intended or may be SAN can be unset from dn after creating a config?
at 
https://github.com/opnsense/core/blob/f08b14695ebc3639e12fbb0abd2e89ff53793ec1/src/etc/inc/certs.inc#L729-L730

Thanks!

**Describe alternatives you considered**

any way of correctly composing the subject

**Additional context**

Add any other context or screenshots about the feature request here or links to relevant forum thread or similar

OPNsense 21.1-amd64FreeBSD 12.1-RELEASE-p12-HBSDOpenSSL 1.1.1i 8 Dec 2020

PR #4755, opened 2021-02-26
certs.inc: unset extensions from DN

kulikov-a:
Hi!
fixes https://github.com/opnsense/core/issues/4675 for me
Thanks!

### DIFF
--- a/src/etc/inc/certs.inc
+++ b/src/etc/inc/certs.inc
@@ -727,6 +727,7 @@ function create_temp_openssl_config(&$dn)
             $template_dn = $dnTag . ' = ' . str_replace(array("\r", "\n"), '', $dn[$dnTag]);
             // Overwrite the placeholders for this property
             $template = str_replace('###OPNsense:' . $dnTag . '###', $template_dn, $template);
+            unset($dn[$dnTag]);
         }
     }
     file_put_contents($configFilename, $template);


AdSchellevis:
@kulikov-a that sounds a bit weird, but since the structure seems to be passed by reference, it's definitely not a good idea to change its contents in the function. If the same variable is used later, one might not understand why it's lacking content. At a first glance I would expect there's something wrong with the template if it generates a wrong faulty certificate, but I don't see the issue at a first glance.

The internal certificate should trace back to 

https://github.com/opnsense/core/blob/61e41715f0dd92ca4f51262cfecbce6e930b535a/src/www/system_certmanager.php#L712-L735

After which it's a simple string replace (don't understand why your patch would do anything in this context either to be honest, since it already replaced the content)

kulikov-a:
@AdSchellevis 
Hi! thanks for your time.
if I understood everything correctly:
at the time of creation, the `$dn` array contains only valid DN attributes (dn_country, dn_state, dn_city...dn_commonname).
then SAN values are added to it. 
then this array is used to create a config file in the `create_temp_openssl_config(&$dn)` function and after that it is passed to the functions `openssl_csr_new($dn, $res_key, $args);` to create a CSR.
so imho after passing the values of the extensions from the  `$dn` to the opnsense.cnf file in the `create_temp_openssl_config(&$dn)` function, there is no longer any need to store them in the  `$dn` array. and we can pass `$dn` with only realy DN attributes to the CSR creation.
I agree. we can unset these extensions  from  `$dn` just before the `openssl_csr_new($dn, $res_key, $args);` call if it is safer.

>  I would expect there's something wrong with the template if it generates a wrong faulty certificate, but I don't see the issue at a first glance
I would not say that the received certificate is incorrect or faulty (I have not found any CSR validation tool that would indicate that this is an error), but it seems to me that this is not very elegant and does not quite comply with the standards

AdSchellevis:
@kulikov-a I'm not sure I see the issue here, just to be sure, is the certificate not as you would expect or the presentation in the gui?

kulikov-a:
@AdSchellevis 
yes, I just have not yet see certificates in which the values of the extensions appeared in the subject (and SAN extension can be quite long). in my understanding, the subject usually contains the value of the DN, which in turn contains the values of the RDNs. but not extensions.
And yes, I do not attribute this to a bug or something. Rather FR)
If I am not convinced, then this is not a problem at all. In the end, I can always add this line before creating a request or issuing a certificate on my OPN)
just tell if I should close these PR and corresponding FR

AdSchellevis:
@kulikov-a not saying you should close it, I just don't understand the issue we're trying to fix. If the certificate itself is good, but our presentation in the gui is less optimal, I don't mind changing that (at the right place). If I'm not mistaken the info button will show full details anyway. If the issue/question is with the cert itself, I must be missing something, just not sure what.

kulikov-a:
> the certificate not as you would expect or the presentation in the gui?

not just GUI
the subject field in the certificate actually contains the values of the extensions. and in my understanding of the standard this is not quite correct
I have added a screenshot to the FR. and I can attach a certificate here so that you can check
csr

AdSchellevis:
@kulikov-a got it, `openssl_csr_new()` expects the first parameter to be an array of distinguished names, which shouldn't include the san indeed. so next question is where to fix this. removing attributes from the array isn't very elegant, but we should indeed remove `subjectAltName` before sending it to any of the openssl handles.

kulikov-a:
Hooray! )
It would seem the most correct to me not to add extensions ('subjectAltName', 'keyUsage', 'extendedKeyUsage', 'basicConstraints') to the `$dn` at all. add them to a separate array of extensions and use it to create a config file in a `create_temp_openssl_config(&$dn)`. i can try to test this

AdSchellevis:
@kulikov-a sounds like a solid plan to me :)

kulikov-a:
@AdSchellevis got it!
I will report on the results )

kulikov-a:
@AdSchellevis 
Hi!
replacing this PR with https://github.com/opnsense/core/pull/4761/
tested on all actions (create root and subCA, make internal cer, csr, sign csr). all works for me.
thanks!

PR #4763, opened 2021-02-27
unbound: Fix PTR records for DHCP endpoints

### LINKED ISSUE #4597: DHCP Lease Update script does not add PTR correctly in Unbound
**Important notices**

**Describe the bug**

The [unbound_dhcp.py script](https://github.com/opnsense/core/blob/master/src/opnsense/scripts/dns/unbound_dhcpd.py) for updating DHCP leases in Unbound does not work for the PTR entries. In particular, the live update via `unbound-control` does not work for the PTR (A record works as expected). The `local-data-ptr` entry in `dhcpleases.conf` works correctly, so after a restart of Unbound, the IP is correctly resolved to the right hostname.

I believe the culprit here is the following line in the `unbound_dhcp.py` script:

https://github.com/opnsense/core/blob/bdc443096370efd38e5833eb11858d6dcdb16e5d/src/opnsense/scripts/dns/unbound_dhcpd.py#L154

The problem seems to be the format of the entry for PTR send with `unbound-control local_data`. I tried the command the script would execute while being logged into my OPNSense box with `unbound-control -c /path/here local_data "10.86.5.21 PTR amun.home"`. This returned `ok`, but `nslookup 10.86.5.21` still returned `** server can't find 21.5.86.10.in-addr.arpa: NXDOMAIN`. 

After some different ways of formatting, the following worked: `unbound-control -c /path/here local_data "21.5.86.10.in-addr.arpa IN PTR amun.home"`

**To Reproduce**

In the following reproduction, I've got a host `amun.home` which gets the DHCP IP `10.86.5.21`, dynamically.

Steps to reproduce the actual error I observed:
1. Delete lease in DHCP config
2. Wait until lease expires and is removed from Unbound
3. run `nslookup 10.86.5.21`. This fails as expected with `** server can't find 21.5.86.10.in-addr.arpa: NXDOMAIN`
4. Have the machine get a new lease
5. run `nslookup amun.home`. Observe that you get the correct IP, `10.86.5.21` back
6. run `nslookup 10.86.5.21`. This fails again with `** server can't find 21.5.86.10.in-addr.arpa: NXDOMAIN`
7. Have a look at `/var/unbound/dhcpleases.conf` and observe that the correct entry `local-data-ptr: "10.86.5.21 amun.home"` exists
8. Restart Unbound
9. run `nslookup 10.86.5.21` yet again. Observe that you now get the correct hostname

**Expected behavior**

A reverse DNS lookup on the address of a host added to Unbound via DHCP returns that host's FQDN.

**Environment**

OPNsense 20.7.7 (amd64, OpenSSL).

gwjo:
Changes:
* Add PTR records for DHCP endpoints in the correct format
* Delete PTR records when A records are deleted
* Use bulk add/remove requests to unbound-control to improve performance

Testing:
* Created new endpoint and confirmed hostname and reverse lookup worked
* After DHCP entry timed out and DNS entry is removed lookups for hostname and IP fail.

Fixes #4597

### DIFF
--- a/src/opnsense/scripts/dns/unbound_dhcpd.py
+++ b/src/opnsense/scripts/dns/unbound_dhcpd.py
@@ -41,23 +41,24 @@ from daemonize import Daemonize
 import watchers.dhcpd
 
 
-def unbound_control(commands, input=None, output_stream=None):
+def unbound_control(commands, input=list(), output_stream=None):
     """ execute (chrooted) unbound-control command
         :param commands: command list (parameters)
-        :param input: (optional )text to be sent to input stream
+        :param input: (optional ) list of lines to be sent to input stream
         :param output_stream: (optional)output stream
         :return: None
     """
-    stderr=None
-    if output_stream is None:
-        stderr=subprocess.STDOUT
+    input_string=None
+    if input:
+        nl='\n'
+        input_string = f'{nl.join(input)}{nl}'
 
     subprocess.run(['/usr/sbin/chroot', '-u', 'unbound', '-g', 'unbound', '/',
                     '/usr/local/sbin/unbound-control', '-c', '/var/unbound/unbound.conf'] + commands,
-                   input=input, stdout=output_stream, stderr=stderr,
+                   input=input_string, stdout=output_stream, stderr=subprocess.STDOUT,
                    text=True, check=True)
 
-    if output_stream is not None:
+    if output_stream:
         output_stream.seek(0)
 
 
@@ -127,8 +128,8 @@ def run_watcher(target_filename, domain):
                 cached_leases[lease['address']] = lease
                 dhcpd_changed = True
 
-        remove_rr = ''
-        add_rr = ''
+        remove_rr = list()
+        add_rr = list()
         if time.time() - last_cleanup > cleanup_interval:
             # cleanup every x seconds
             last_cleanup = time.time()
@@ -140,7 +141,8 @@ def run_watcher(target_filename, domain):
                         "dhcpd expired %s @ %s" % (cached_leases[address]['client-hostname'], address)
                     )
                     fqdn = '%s.%s' % (cached_leases[address]['client-hostname'], domain)
-                    remove_rr += f'{fqdn}\n{ipaddress.ip_address(address).reverse_pointer}\n'
+                    remove_rr += [ ipaddress.ip_address(address).reverse_pointer,
+                                   fqdn ]
                     if unbound_local_data.is_equal(address, fqdn):
                         unbound_local_data.cleanup(address, fqdn)
                     del cached_leases[address]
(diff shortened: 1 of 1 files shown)


src/opnsense/scripts/dns/unbound_dhcpd.py
```diff
@@ -119,6 +126,9 @@ def run_watcher(target_filename, domain):
                     and 'client-hostname' in lease and 'address' in lease and lease['client-hostname']:
                 cached_leases[lease['address']] = lease
                 dhcpd_changed = True
+
+        remove_rr = ''
+        add_rr = ''
```
AdSchellevis, line 131: Maybe it's better to replace both for lists, it's slightly faster and saves the \n on each line.

src/opnsense/scripts/dns/unbound_dhcpd.py
```diff
@@ -129,8 +139,8 @@ def run_watcher(target_filename, domain):
                         syslog.LOG_NOTICE,
                         "dhcpd expired %s @ %s" % (cached_leases[address]['client-hostname'], address)
                     )
-                    unbound_control(['local_data_remove',  cached_leases[address]['client-hostname']])
                     fqdn = '%s.%s' % (cached_leases[address]['client-hostname'], domain)
+                    remove_rr += f'{fqdn}\n{ipaddress.ip_address(address).reverse_pointer}\n'
```
AdSchellevis, line 143: shouldn't it drop the A record here too?

AdSchellevis (changes requested):
@gwjo thanks for working on this, I have some small remarks and questions which are in the review.

src/opnsense/scripts/dns/unbound_dhcpd.py
gwjo: Sure.

Each RR record needs to be sent on a separate line, but I can do this inside of `unbound_control()`, keeping this code cleaner.

src/opnsense/scripts/dns/unbound_dhcpd.py
gwjo: It does there are two lines here, the reverse pointer for the PTR record and the FQDN for the A record.

Switching to a list will make this much clearer.

AdSchellevis:
@gwjo thanks, I seem to have missed this https://github.com/opnsense/core/commit/aa6c44e767a8a61b790d4232a478ff767d8c3293 in my last review, can you check master again?

gwjo:
> @gwjo thanks, I seem to have missed this [aa6c44e](https://github.com/opnsense/core/commit/aa6c44e767a8a61b790d4232a478ff767d8c3293) in my last review, can you check master again?

Looks good.  

Good catch, thanks!

PR #4767, opened 2021-02-28
dnsbl: Prefer domain list over host file format

gwjo:
Where the blacklist provider has both host file format and a raw list of
domains to block, download the raw list.  The raw list will always be a
smaller download and require less processing.


AdSchellevis:
@gwjo thanks!

PR #4773, opened 2021-03-03
firewall_rules.php: Icon color after toggle fix

kulikov-a:
Hi.
block\reject rules get wrong color on toggle


AdSchellevis:
@kulikov-a thanks!

kulikov-a:
thanks!!

Issue #4173, opened 2020-06-15
MAC Address Mismatch between ether and hwaddr


TekunoKage:
Hardware MAC address and Ethernet MAC Address mismatch 

I have the same problem on two VMs with different mac address and the MAC is change to the same on both of theme causing an conflict.

MAC Address changes on both systems to: 9e:4a:83:67:7f:b0

To Reproduce

One PHY interface with three vlans on it.
then reboot the system.

On the below output the issue is on the vtnet9 interface and any vlan related to it.

```
root@GUYNPR1-OpnSense-NAF1:~ # ifconfig | grep "vtnet\|hwaddr\|ether"
vtnet0: flags=8843<UP,BROADCAST,RUNNING,SIMPLEX,MULTICAST> metric 0 mtu 1500
        ether d6:3c:9d:14:d1:ed
        hwaddr d6:3c:9d:14:d1:ed
        inet6 fe80::d43c:9dff:fe14:d1ed%vtnet0 prefixlen 64 scopeid 0x1 
vtnet1: flags=8943<UP,BROADCAST,RUNNING,PROMISC,SIMPLEX,MULTICAST> metric 0 mtu 1500
        ether 36:aa:8b:d8:87:2c
        hwaddr 36:aa:8b:d8:87:2c
        inet6 fe80::34aa:8bff:fed8:872c%vtnet1 prefixlen 64 scopeid 0x2 
vtnet2: flags=8943<UP,BROADCAST,RUNNING,PROMISC,SIMPLEX,MULTICAST> metric 0 mtu 1500
        ether de:78:1f:b3:c1:46
        hwaddr de:78:1f:b3:c1:46
        inet6 fe80::dc78:1fff:feb3:c146%vtnet2 prefixlen 64 scopeid 0x3 
vtnet3: flags=8802<BROADCAST,SIMPLEX,MULTICAST> metric 0 mtu 1500
        ether b6:6e:db:cb:b1:91
        hwaddr b6:6e:db:cb:b1:91
vtnet4: flags=8802<BROADCAST,SIMPLEX,MULTICAST> metric 0 mtu 1500
        ether 26:8c:bc:36:94:19
        hwaddr 26:8c:bc:36:94:19
vtnet5: flags=8802<BROADCAST,SIMPLEX,MULTICAST> metric 0 mtu 1500
        ether 7e:43:3c:d1:21:25
        hwaddr 7e:43:3c:d1:21:25
[...]

PR #4774, opened 2021-03-03
Fine-grained MAC address spoofing for VLANs

maurice-w:
If a spoofed MAC address is configured for an interface (physical or VLAN), it gets applied to the physical interface as well as all of its VLANs. If different spoofed MAC addresses are configured for individual VLANs, the one configured last gets applied to all affected interfaces.

This is unexpected and not suitable for all use cases. It was reported in #4173 and there is a recent bug report on the forum: https://forum.opnsense.org/index.php?topic=21705

With this patch, spoofed MAC addresses get applied exactly as configured by the user.

### DIFF
--- a/src/etc/inc/interfaces.inc
+++ b/src/etc/inc/interfaces.inc
@@ -2401,14 +2401,17 @@ function interface_configure($verbose = false, $interface = 'wan', $reload = fal
      * which triggers the interface config again, which attempts to
      * spoof the MAC again which cycles the link again...
      */
-    if (!empty($wancfg['spoofmac']) && strcasecmp($wancfg['spoofmac'], get_interface_mac($realhwif))) {
-        mwexecf('/sbin/ifconfig %s link %s', array($realhwif, $wancfg['spoofmac']));
+    if (!empty($wancfg['spoofmac']) && strcasecmp($wancfg['spoofmac'], get_interface_mac($realif))) {
+        mwexecf('/sbin/ifconfig %s link %s', array($realif, $wancfg['spoofmac']));
 
-        /* All vlans need to spoof their parent mac address, too. */
-        if (isset($config['vlans']['vlan'])) {
+        /* A MAC address change on a parent interface resets all VLANs to the parent MAC, so we have to restore spoofed VLAN MACs. */
+        if (!stristr($realif, "_vlan") && isset($config['vlans']['vlan'])) {
             foreach ($config['vlans']['vlan'] as $vlan) {
                 if ($vlan['if'] == $realhwif) {
-                    mwexecf('/sbin/ifconfig %s link %s', array($vlan['vlanif'], $wancfg['spoofmac']));
+                    $friendly_vlanif = convert_real_interface_to_friendly_interface_name($vlan['vlanif']);
+                    if (!empty($friendly_vlanif) && !empty($config['interfaces'][$friendly_vlanif]['spoofmac'])) {
+                        mwexecf('/sbin/ifconfig %s link %s', array($vlan['vlanif'], $config['interfaces'][$friendly_vlanif]['spoofmac']));
+                    }
                 }
             }
         }


src/etc/inc/interfaces.inc
```diff
@@ -2401,14 +2401,17 @@ function interface_configure($verbose = false, $interface = 'wan', $reload = fal
...
-        if (isset($config['vlans']['vlan'])) {
+        /* A MAC address change on a parent interface resets all VLANs to the parent MAC, so we have to restore spoofed VLAN MACs. */
+        if (!stristr($realif, "_vlan") && isset($config['vlans']['vlan'])) {
             foreach ($config['vlans']['vlan'] as $vlan) {
                 if ($vlan['if'] == $realhwif) {
-                    mwexecf('/sbin/ifconfig %s link %s', array($vlan['vlanif'], $wancfg['spoofmac']));
+                    $friendly_vlanif = convert_real_interface_to_friendly_interface_name($vlan['vlanif']);
+                    if (!empty($friendly_vlanif) && !empty($config['interfaces'][$friendly_vlanif]['spoofmac'])) {
+                        mwexecf('/sbin/ifconfig %s link %s', array($vlan['vlanif'], $config['interfaces'][$friendly_vlanif]['spoofmac']));
+                    }
```
AdSchellevis, line 2414: @maurice-w shouldn't there be an `else` here setting the macaddr to the interface address when not overwritten by the vlan itself? at a first glance I would expect this can break existing setups depending on the mac address being equal on both vlan and interface.

maurice-w: @AdSchellevis Setting the MAC address on a physical interface trickles down to the VLANs automatically. This happens on the OS level, we don't have to do it manually. Why did the original code do it anyway? I have no idea.

So the default behavior doesn't change: Configuring a spoofed MAC on a parent is applied to all of its VLANs, too (unless they have their own spoofed MAC).

AdSchellevis: @maurice-w I guess it's in the order of things, if you set it, apply and remove it, it will stick to the old value...

maurice-w: @AdSchellevis I'm pretty sure removing a spoofed MAC has always required a reboot. We don't do anything if `$wancfg['spoofmac']` is empty so the old spoofed MAC sticks until the interface gets reinitialized on boot.

AdSchellevis: @maurice-w I mean the other way around, set a mac on the vlan, apply and remove it.  In case the parent is reset again, we can't know the old value indeed, but if the vlan changes and the parent is left untouched one might expect after a save it will drop back to the interface mac.

maurice-w: @AdSchellevis, please help me out here, I don't get it. :-)
- This is no longer about breaking existing setups, is it? _"set a mac on the vlan, apply and remove it"_ never worked without a reboot. You're requesting new / improved functionality?
- Since we could only make this work for VLANs, we would shift from "always requires a reboot" to "sometimes works, but sometimes requires a reboot, depending on the type of interface". Do we want that?
- An `else` here would do nothing because we don't get this far if `$wancfg['spoofmac']` is empty. We would have to restructure the entire flow. I intended this change to be limited, but if desired, I guess I can do that.

AdSchellevis: @maurice-w It's about a bit about "breaking", with vlan's supporting custom mac addresses, the remove operation isn't functional anymore with this change. It's easy to explain that you do need a reboot to reset back to factory address, a little less obvious that removing a vlan mac address doesn't bring it back to it's parent one like it previously did. 

I guess my single point is that now you can end up in a state where some mac addresses of the interface are different but configured equal, where in the past they where always the same. I know we can't retrieve the original value if the parent is reconfigured, no problem, I'm just very sure one day someone reports an issue about "not all vlan's are equal, but configured the same" which will be difficult to track.

If the whole intend it to clean the code and always set the mac address when configured and don't try to sync vlan's, this probably isn't the right fix. (removing all vlan code here likely has the same effect)

maurice-w: > removing a vlan mac address doesn't bring it back to it's parent one **like it previously did**

Sorry, but no, it never did. Not without a reboot. That's my point. Please try it. Set a spoofed MAC on a VLAN, apply, remove the spoofed MAC, apply again. The spoofed MAC sticks until you reboot or apply a different spoofed MAC. The patch doesn't change this at all. And a reboot still brings the VLANs back to the parent MAC.

I'm happy to talk about functional improvements, but would like to get the "breaks existing setups" concern out of the way first.

AdSchellevis: I'll keep the PR open to discuss internally later, I see your point, just don't have a lot of time to dig into this now.

maurice-w: Okay. Just to clarify: The patch is meant to fix a specific issue without breaking anything. And I'm convinced it does. It's not meant to solve _all_ MAC address spoofing related issues once and forever. Continual improvement.

AdSchellevis: yes, that's clear, before pulling it in we do need to assess the impact though. The connectivity between the interface and the associated vlan from the system part is just a bit odd. Maybe that changed somewhere over time, I don't know.

maurice-w:
According to feedback on the forum and some quick research, setting the MAC addresses works fine. However, if a VLAN MAC address is different from the parent MAC address, VLAN frames only pass if promiscuous mode is enabled. I see three options:

1. Check whether multiple MAC addresses are configured on the same physical interface. If yes, enable promiscuous mode.
2. Add a checkbox somewhere in the GUI to manually enable promiscuous mode.
3. Decide that it's not supported, document this and put input validations in place to prevent users from entering multiple MAC addresses on the same physical interface.

@AdSchellevis, thoughts?

AdSchellevis:
Hi @maurice-w thanks for gathering feedback, it's a bit of an edge case, the question is how this is supposed to work. 

I can't find a lot of information about the subject to be honest (other than a ticket at pfSense claiming you shouldn't https://redmine.pfsense.org/issues/1337). 

When choosing between your options, I would tend to choose 2 at the moment, automatically setting an interface in promisc mode isn't great, when you're allowed to setup a feature with (documented) constraints we usually don't mind that much. It would help if we had an idea how other vendors implement this (if they do).

maurice-w:
Unfortunately, the promiscuous mode requirement makes this much more complex than anticipated:
- If it were to be a manual setting, where to put it? It would have to exist exactly once per _physical_ interface and I don't see a good spot in the UI for this.
- If only VLANs are used (parent not assigned) and the user wants the same spoofed MAC on all VLANs (or there is only one VLAN), applying the spoofed MAC to the parent would remain the preferred method (no promiscuous mode required). We would have to provide a way to enable this use case.
- There are possible existing setups which would require promiscuous mode to be enabled in order to keep working after applying the patch. If promiscuous mode were to become a manual setting, this would require user intervention to keep their system working.

Detecting the use case and automatically enabling promiscuous mode (or instead applying the spoofed MAC to the parent) might be the better option.

I don't have a good setup for testing this (all virtualized). Maybe the person originally asking for this on the forum is interested in working on it, too.

src/etc/inc/interfaces.inc
marjohn56: Came here by a roundabout method trying to work out why the LAN interface on PC was getting RAs from different VLANs, Checked my switches, no problem there, checked the firewall, removing 134s from filter.lib.inc to narrow it down; problem is definitely on the firewall. So Checked all my rules, they seem fine, checked whether there were any tunables that might be to blame.,, So I set the interface to Promiscuous and installed this patch.. problem solved. Thanks @maurice-w, spent half a day trying to fix  this as I'm pretty certain it never used to happen, but hey, everything appeared to be working fine anyway so I wasn't really looking.

maurice-w: Great to hear @marjohn56! Unfortunately, adding a reliable and user-friendly way to enable promiscuous mode seems harder than expected (see my last comment below). Would you by any chance be interested in helping with that? :-)

marjohn56: Happy to, though my time is sadly limited by real work, only had a bit of time to play today as what was supposed to happen didn't and it was too late to arrange to do anything else!

There's nothing particularly complex on my system. It's a simple Qotom system where I am using one NIC and pushing three VLANs down it, there is no untagged LAN per say, just the three VLANs; connects straight to one of the trunks on my first switch and out to the rest of the switches from there. I'll comment on your last comment now.

marjohn56:
There are a few "gotcha's" here, but basically it boils down to your second point, and I have to ask why would you want a single VLAN and no parent? Also why would you NOT want different MAC addresses on the interfaces, by that I mean physical and virtual, that is why I am here, it was breaking things. BTW, I went as far as re-installing RTADVD just to check it wasn't RADVD, that also should have flagged up something amiss as the rtadvd log was showing:
>  xxxx:xxxx:xxxx:3::/64 from fe80::20e:c4ff:fed2:8142 on igb1_vlan101 is not in our list
and that was there when we were testing rtadvd before, just took no notice of it. Anyway back to the chase. If the user is using more than one interface on the same NIC, then it should use independent macs and promiscuous mode.... IMHO.

maurice-w:
Single VLAN and no parent is not uncommon for the WAN interface (some ISPs require VLAN tagging).

All VLANs having the same MAC address as the parent is the default on most systems I am aware of. Cisco, Juniper, whatever. Using individual MACs is the exception (which we try to enable here).

maurice-w:
What about this:
- We add a promiscuous mode checkbox at a yet to be determined location.
- If disabled (default), we use the current MAC spoofing behaviour (changing any VLAN MAC changes the parent MAC and all other VLAN MACs). So nothing breaks for existing configurations. And the "single VLAN, no parent" use case keeps working without promiscuous mode.
- If enabled, we enable promiscuous mode _and_ switch to the new fine-grained configuration logic.

Thoughts?

maurice-w:
A good location for enabling promiscuous mode might be 'Interfaces: Settings', in the top section amongst the hardware settings. We could use a multi selection drop-down which only contains the physical interfaces (including the non-assigned ones). No VLANs, no virtual / tunnel interfaces without a MAC address.

MWeesenaar:
Hi all,

I opened the topic on the forum. As I tried to describe on the forum, my ISP (Telfort, in the Netherlands) originally delivers a modem which has support for internet, IPTV and VOIP. They all arrive on three different VLANs and therefore I need to specify these (however, I am not using VOIP) on the physical WAN-interface and provide the VLANs with some DHCP-options and so on. If more info or background would help, I am fine with setting up a call to explain.

@maurice-w - I think your suggestions make sense. I cannot oversee the existing configurations and whether the suggested approach covers existing cases.

marjohn56:
> What about this:
> 
> * We add a promiscuous mode setting at a yet to be determined location.
> * If disabled (default), we use the current MAC spoofing behaviour (changing any VLAN MAC changes the parent MAC and all other VLAN MACs). So nothing breaks for existing configurations. And the "single VLAN, no parent" use case keeps working without promiscuous mode.
> * If enabled, we enable promiscuous mode _and_ switch to the new fine-grained configuration logic.
> 
> Thoughts?

That all makes sense to me. However, this morning I am back with my original problem, thought this had fixed it but it hasn't. So back to square one for me.

marjohn56:
OK, I have solved my problem, not Opn at all but the Sky Q devices had turned their WiFi back on, after an update I suppose, which caused massive  loopback issues, it was the last thing I expected. Do you want to me to put together a commit that does the GUI and logic?

MWeesenaar:
I'd love to see and test your proposal, @marjohn56 :)

marjohn56:
I'm waiting on @maurice-w to tell me if he wants to to start looking at the GUI, he might be doing it and there's no point in us both  doing the same thing.. :)

maurice-w:
@marjohn56, would be great if you could look into the promiscuous mode. I can't find a good spot in the config for physical interface settings. Something like `$config['physicalinterfaces']` simply doesn't seem to exist. `$config['interfaces'][$interface]['promiscuous']` would work for assigned physical interfaces. But what if only VLANs are assigned while the parent is unassigned? Where do we store the promiscuous flag?

marjohn56:
I'll take a look over the next few days, probably at the weekend. Probably interfaces, I'll need to play with it in each state to see what works best and is logical.

marjohn56:
I'm not certain myself of the best place, I first thought interfaces, but that's really not quite right, then I thought vlans, that's not right either. I think it must first be in system, such as 'allowautopromiscuous' . So for example, when the system is configuring the vlans, we can scan all vlans and check to see if there is a spoofed IP set. If there is then check to see if the parent has a configuration, if not, then don't enable promiscuous. If there is a parent config and there is a VLAN with a spoofmac then enable promiscuous. If the vlan does not have a spoofmac entered, then carry on as  normal. If a vlan does have a spoof  AND the 'allowautopromiscuous'' entry exists then check to see if spoofs are in place on other vlans, if not then create them automatically,   then enable promiscuous on the parent. The System flag 'allowautopromiscuous' is the key, without that nothing new happens, only if that is set can the rest of the magic happen.
Thoughts....

maurice-w:
That is actually a great idea! Not manual (which doesn't fit OPNsense's strict interface assignment requirement) and not automatic (which might break existing configurations). But manually automatic. I really like that. And it only requires a single new checkbox in the UI. Which I think I can handle. ;-)
So if you haven't already started coding, I'll do so tonight.

marjohn56:
Not coding yet, just re-installing on my test unit, I've been using it to play with RouterOS, Most of the work will be where the interfaces are configured.

maurice-w:
> Most of the work will be where the interfaces are configured.

Indeed. If you don't mind, I'll give it a shot and make a commit later tonight. Which you then could dissect and test. Would you be okay with that?

maurice-w:
Initial version, untested, GUI tomorrow. Feedback welcome. :-)

marjohn56:
Had a quick look... real work today. On a quick scan looks fine, but are you scanning for ALL vlans associated to an interface, i.e. what if there is more than one VLAN, one with a spoof mac, one without but the parent is not configured? Did I miss that?

wolfspyre:
I think we need to think about this as "select a multi-mac-mode for this physical interface" and handle the rest of the interface accordingly. 
 that lets the user select the bounds of the interface promiscuity and gives you their preference so as to configure sub items accordingly. seeing as ultimately all bridges/vlans are children of the phy, that feels the most straight forward and consistent way of thinking about it.

how would this play in a situation of:
```
     lagg0 (  phy0  phy1 )  {
        vlan1(  br_01(  "vlan1_bridge" - front door interface for vlan 1 )  )
        vlan2(  br_02(  "vlan2_bridge" - front door interface for vlan 2 )  )
        vlan3(  br_03(  "vlan3_bridge" - front door interface for vlan 3 )  )
        vlan4(  br_04(  "vlan4_bridge" - front door interface for vlan 4 )  )
        vlan5(  br_05(  "vlan5_bridge" - front door interface for vlan 5 )  )
     }
```

maurice-w:
@wolfspyre, indeed, I did not really consider LAGGs and bridges. I don't even know how MAC address spoofing _currently_ behaves in these scenarios. Do you have experience with that? I'd say let's get this done for plain "VLAN on physical" setups first, then test it thoroughly and see what needs do be done to make it work with LAGGs and bridges (if it's even possible).

@marjohn56, yes, all VLANs are processed. But "more than one VLAN, one with a spoofed mac, one without, parent not assigned" is indeed an edge case where the spoofed VLAN MAC might get applied to the parent and, as a result, to the VLAN without a spoofed MAC. I'll look into this. And I did not forget about the GUI part, just an unscheduled delay... I'll make another commit soon-ish.

marjohn56:
The thought behind my question is that because of 'allowautopromiscuous' being set, when there is more than one VLAN, and only one VLAN has a mac set, then we should auto assign a spoof mac to the VLAN with the missing MAC, there will always be the one user who forgets to add a spoof to the interface and then complains it doesn't work. On the other hand there could just be a 'i' that states that ALL interfaces should have individual spoofs set when 'allowautopromiscuous' is set.

tank0226:
> 
> 
> Single VLAN and no parent is not uncommon for the WAN interface (some ISPs require VLAN tagging).
> 
> All VLANs having the same MAC address as the parent is the default on most systems I am aware of. Cisco, Juniper, whatever. Using individual MACs is the exception (which we try to enable here).

Just to share my use case here as I recently came across this issue while trying to swap out my ISP provided router.

My ISP's single ethernet uplink carries 2 VLAN (Internet and IPTV). I configured OPNsense'a VLAN, MAC and DHCP according to the sniffed traffic. Every time the WAN cable is connected, one of the vlan gets an IP, the interface resets, the other interface gets an IP, the interface resets and this cycle repeats. After reading the interfaces.inc file then I realised OPNsense "attempts to spoof the MAC again which cycles the link again".

My ISP's setup requires both VLAN interface to have different MAC address. If I set both to the same, whoever sends out DHCPDISCOVER first gets offered the respective subnet's IP. The other VLAN gets absolutely no DHCP reply.

Also I want to check if it is sufficient to directly edit the interfaces.inc file on OPNsense so that I can test out this patch and also is "ifconfig wan promisc" the correct way to set promiscuous mode (for now).

Thanks!

tank0226:
Tested the patch by enabling promiscuous mode manually on the parent interface and bypassing the 'allowautopromiscuous' check. VLAN interfaces are stable with no resets and IPTV works.

wolfspyre:
so... How do we move this forward? What are the concerns? without it present, there's not a viable (as far as I know) mechanism to facilitate unique MAC addrs on vlan-interfaces. near as I can tell, with the allowautopromiscuous attr in the config, this patch works as is intended.... what additional validation needs to be performed? 

is this a direction that the powers that be want to go? 
is there a  better mechanism of facilitation that's available?
am I misunderstanding the path forward , or the desired functionality of this patch?

rodeho:
> My ISP's single ethernet uplink carries 2 VLAN (Internet and IPTV). I configured OPNsense'a VLAN, MAC and DHCP according to the sniffed traffic. Every time the WAN cable is connected, one of the vlan gets an IP, the interface resets, the other interface gets an IP, the interface resets and this cycle repeats. After reading the interfaces.inc file then I realised OPNsense "attempts to spoof the MAC again which cycles the link again".


@maurice-w goodmorning Maurice, today I tried to implement my router-on-a-stick setup with IPTV, unfortunately the above behavior occurs. Or the internet vlan gets an IP via DHCP "or" the IPTV vlan gets an IP via DHCP....but not both of them.

Do you have any idea when this code will be implemented, that would be great!

maurice-w:
I'm currently taking a break from all OPNsense related activities for personal reasons. But I have a to do list with several PRs which I will come back to at some point. This is not forgotten, just postponed. No ETA, sorry.

rodeho:
> I'm currently taking a break from all OPNsense related activities for personal reasons. But I have a to do list with several PRs which I will come back to at some point. This is not forgotten, just postponed. No ETA, sorry.

Thanks Maurice, take care of yourself.

calebj:
I had the requirement to spoof the MAC address on my WAN interface for my ISP, so I tried this patch with `allowautopromiscuous` enabled. Unfortunately, this caused my previously stable pfsync setup to stop working, so I reverted and spoofed the MAC using an additional interface on the proxmox side. Inefficient, but at least it worked.

But today I stumbled upon the solution of creating a bridge on top of the VLAN interface in question. An interface assignment has to be created on the VLAN with NO spoofed mac or IP configuration, and with the required MTU if different from the parent interface (mine is). Then create a bridge with that VLAN as the only port assigned. Finally, set the spoofed MAC and IP configuration on a new interface assignment.

I hope this helps someone!

Now, for some details about how it caused pfsync to break. For the vlan parent device, I put the original MAC, so that the VLAN's MAC wouldn't also change the parent's. This was apparently enough to confuse pfsync. The behavior I observed was mixed, and I couldn't identify a pattern. Sometimes sync messages would only come from one node, other times both were sending but one side was ignoring the other (500 vs. 30 states, failover in one direction causing broken connections but not the other). In the latter case, if I recall correctly, the side dropping the updates was showing ~twice the number of pfsync node IDs it should have, like it was taking more than the names into account. Maybe changing the promiscuous setting or manually setting the link address changes the pfsync node ID, even if the new MAC is the same as the factory default. I haven't dug further into the malfunction because I found a workaround.

MWeesenaar:
First of all; best of luck for you and take care, @maurice-w . :) 

@marjohn56 - is there something you might be able to pick up?
Or in a more generic sense: what needs to be done before this pull request can be merged?

danijelt:
I was hit by this issue too. I use Intel X710 4-port SFP+ PCI-e card (i40 driver) and it's even worse with it: I have two LACP groups with two SFP+ ports on the same card, and setting spoofed MAC on one VLAN spanned it across both LACP groups and all VLANs defined on them.

I'll try applying this patch to a spare box that I planned to use for pfSync before I was stopped by this issue to see what happens.

PR #4784, opened 2021-03-06
LDAP minor fixes

kulikov-a:
Hi!
based on https://forum.opnsense.org/index.php?topic=21605.0
- baseDN igonred in search function if ldapauthcontainers is set. help text for "Authentication containers" input fixed
https://github.com/opnsense/core/blob/25959a45a918000c36d092403ac3e3c223c4671a/src/opnsense/mvc/app/library/OPNsense/Auth/LDAP.php#L160-L163

- In select "Authentication containers" modal Close button is actually a Save button (writes changes to the "Authentication containers" input)

- Some debug info added to System: Access: Tester page for LDAP servers
tester1
tester2

thanks!

### DIFF
--- a/src/opnsense/mvc/app/library/OPNsense/Auth/LDAP.php
+++ b/src/opnsense/mvc/app/library/OPNsense/Auth/LDAP.php
@@ -199,7 +199,7 @@ class LDAP extends Base implements IAuthConnector
         $error_string = "";
         if ($this->ldapHandle !== false) {
             ldap_get_option($this->ldapHandle, LDAP_OPT_ERROR_STRING, $error_string);
-            syslog(LOG_ERR, sprintf($message . " [%s,%s]", $error_string, ldap_error($this->ldapHandle)));
+            syslog(LOG_ERR, sprintf($message . " [%s; %s]", str_replace(array("\n","\r","\t"), ' ', $error_string), ldap_error($this->ldapHandle)));
             $this->lastAuthErrors['error'] = $error_string;
             $this->lastAuthErrors['ldap_error'] = ldap_error($this->ldapHandle);
         } else {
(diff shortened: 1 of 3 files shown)


AdSchellevis:
@kulikov-a maybe we better add a method `getLastErrors()` for the  error feedback and move this into a separate array, it makes it more explicit that this is error feedback and not something returned by the authenticator. Other than that it's definitely an improvement to make sure errors are returned to the user here.

kulikov-a:
@AdSchellevis 
thanks! more correct of course
I will try to do so. for now I'm converting it to a draft

kulikov-a:
@AdSchellevis it's kind of done.
moved errors to a separate array.
- slightly tweaked the `logLdapError` function: it turned out that a `LDAP_OPT_ERROR_STRING` can contain not only commas but also new lines and even tabs special characters.
- added Cancel button to "Authentication containers" modal.
- added some debug (and the message for this case in modal window) for "Authentication containers" empty search results
(https://forum.opnsense.org/index.php?topic=21605.msg103543#msg103543)

please take a look when you have time

AdSchellevis:
@kulikov-a looks good, thanks, I'll pull it in to master, haven't tried it myself yet, but don't see any reason why it shouldn't.

kulikov-a:
Wow! thanks!!!

AdSchellevis:
@kulikov-a made a small adjustment after testing (https://github.com/opnsense/core/commit/3277c0beb6f3ecb957fef0e95ae0562cea43eb99), when you can't connect, there's an ldap connect error but no LDAP_OPT_ERROR_STRING, so it's better to omit it in that case.

kulikov-a:
@AdSchellevis thanks! it is always better to add existing ones ) 
I should have noticed it. thank!

Issue #4788, opened 2021-03-09
Feature request: Copy multiselect to clipboard


kulikov-a:
**Important notices**

Before you add a new report, we ask you kindly to acknowledge the following:

- [x] I have read the contributing guide lines at https://github.com/opnsense/core/blob/master/CONTRIBUTING.md
- [x] I have searched the existing issues and I am convinced that mine is new.

**Is your feature request related to a problem? Please describe.**

Hello.
Now it is not possible to copy values from multiselect fields to the clipboard. This can be useful when transferring settings, for editing large lists (eg whitelists or some), or, for example, for checking URLs in aliases.

**Describe the solution you like**

Add "Copy to clipboard" link next to the "Clear All" link for multiselect fields in MVC Views and opnsense_ui.js
![copy1](https://user-images.githubusercontent.com/36099472/110428075-cc4f8280-80b9-11eb-90ed-3b77cac41fc7.PNG)
![copy2](https://user-images.githubusercontent.com/36099472/110428118-dbcecb80-80b9-11eb-9cae-80691ebeaf96.PNG)

Thanks!
**Describe alternatives you considered**

did not find "copy to clipboard" capability in the tokenize2 itself

**Additional context**

Add any other context or screenshots about the feature request here or links to relevant forum thread or similar

Issue #4497, opened 2020-12-03
FR: Support for regex with comma and quantifiers


kulikov-a:
**Important notices**
Before you add a new report, we ask you kindly to acknowledge the following:

- [x] I have read the contributing guide lines at https://github.com/opnsense/core/blob/master/CONTRIBUTING.md

- [x] I have searched the existing issues and I'm convinced that mine is new.

**Is your feature request related to a problem? Please describe.**
Hello!
some fields imply the use of regular expressions (for example, DNSBL exclusions in unbound or black\whitelists in squid acls etc.). regular expressions can contain a comma and  quantifiers.
now it is not possible to use all the features of regexp, such as, for example:
^(?!(?=^.{4,253}$)(^((?!-)[a-zA-Z0-9-_]{0,62}[a-zA-Z0-9]\.)+[a-zA-Z0-9-]{2,63}[\.]?$))
since the gui not allow to insert them and there is no corresponding fieldtype for storage in the config (now the CSVListField is used for this, in which the comma is the separator)

![unbound_regex](https://user-images.githubusercontent.com/36099472/101486146-43fffb00-396d-11eb-9b34-7d57c87238ef.PNG)

**Describe the solution you'd like**
By default, the tokenize2 (used for selects) uses comma, tab and return key as delimiters.
Its possible to remove the comma from the delimiters in GUI
(eg` 'delimiter': '&#13;' `in formatTokenizersUI() in opnsense_ui.js)

To store this values we can use new fieldtype
(eg NLSVListField - same as CSVListField but with `\n` separator)

after these changes, it is easy to change template, model and form XMLs of plugins to fully support regexps

I checked on unbound. it works
(https://forum.opnsense.org/index.php?topic=20284.msg94444#msg94444)
![unbound_dnbsl1](https://user-images.githubusercontent.com/36099472/101486211-59752500-396d-11eb-8da1-4a07331b6bf3.PNG)
![unbound_dnbsl2](https://user-images.githubusercontent.com/36099472/101486226-5f6b0600-396d-11eb-9746-ba2f6b0b388f.PNG)

Thanks!

**Describe alternatives you've considered**

**Additional context**

kulikov-a:
since can not find any other use other than to use for sanitizing blacklists, the request is perfectly closed by https://github.com/opnsense/core/commit/31a0c40e3f503528f3adb05fb1bdd8b139495c38

PR #4789, opened 2021-03-09
MVC View: add copy to mutliselects

kulikov-a:
Hi
possibly closes https://github.com/opnsense/core/issues/4788
I'm not sure about the separator: I made a `\n`  because I use https://github.com/opnsense/core/issues/4497. Before FR review and in case of refusal, we can make a `,`
Thanks!

### DIFF
--- a/src/opnsense/mvc/app/views/OPNsense/Firewall/alias.volt
+++ b/src/opnsense/mvc/app/views/OPNsense/Firewall/alias.volt
@@ -195,6 +195,7 @@
         $("#alias\\.type").change(function(){
             $(".alias_type").hide();
             $("#row_alias\\.updatefreq").hide();
+            $("#copy-options_alias\\.content").hide();
             switch ($(this).val()) {
                 case 'geoip':
                     $("#alias_type_geoip").show();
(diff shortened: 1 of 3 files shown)


fichtner:
Lovely idea. Maybe we can go a bit further and add a paste as well making the operations more like cut/copy/paste everyone knows?

kulikov-a:
@fichtner nice to hear )
tokenize2 supports pasting but only with the specified delimiter (now it's a comma). Using a link, of course, we can check the separator and bring it to the desired value. I'll try to play with this

kulikov-a:
@fichtner played with pasting and can't make compatible code (clipboard security, permissions etc):
on Chrome it works well with the Clipboard API's navigator.clipboard.readText() method but "Firefox only supports reading the clipboard in browser **extensions**" 
https://developer.mozilla.org/en-US/docs/Web/API/Clipboard/readText
`execCommand("paste")` also disabled on most browsers or works in extensions
can you suggest more ways? thanks!

kulikov-a:
@fichtner @AdSchellevis 
Hi!
I have prepared a variant with the pasting possibility.
From the pros: we can insert, for example, a column from the Excel sheet without first changing the comma separator.
The downside is cross-browser compatibility. Checked on Chrome, Opera and FF. Chrome and Opera - everything works well. On FF, I simply do not display the paste link
How will it be more convenient? Do you want me to update the changes in this request or add a new one?
Thanks!
copy-paste

AdSchellevis:
@kulikov-a I'm not sure, but if you post the code I'll take a look when I have some time available. The screenshot looks nice by the way.

PR #4790, opened 2021-03-09
firewall: categories improvements

swhite2:
This commit adds the option of filtering on (No Category) in the firewall rules. A tooltip for visual aid is also added. In order to run static text through gettext(), a separate API endpoint has been created. This is a clean PR and invalidates the old one (https://github.com/opnsense/core/pull/4775) since it was pretty messy.

### DIFF
--- a/src/www/javascript/opnsense_legacy.js
+++ b/src/www/javascript/opnsense_legacy.js
@@ -180,7 +180,7 @@ function window_highlight_table_option()
  */
 function hook_firewall_categories() {
     let cat_select = $("#fw_category");
-    ajaxCall('/api/firewall/category/searchItem', {}, function(data){
+    ajaxCall('/api/firewall/category/searchNoCategoryItem', {}, function(data){
         if (data.rows !== undefined && data.rows.length > 0) {
             let color_map = {};
             for (let i=0; i < data.rows.length ; ++i) {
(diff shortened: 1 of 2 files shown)


src/www/javascript/opnsense_legacy.js
```diff
@@ -242,6 +242,10 @@ function hook_firewall_categories() {
             $(".rule").each(function(){
                 let is_selected = false;
                 $(this).data('category').split(',').forEach(function(item){
+                    if (selected_values.indexOf("(No Category)") > -1 && item === "") {
```
AdSchellevis, line 245: doesn't take translations into account. when "no category" is always on top it's probably better to check if the first element of the selector is checked.

src/www/javascript/opnsense_legacy.js
```diff
@@ -257,5 +261,6 @@ function hook_firewall_categories() {
             $(".opnsense-rules").change();
         });
         cat_select.change();
+        $('[data-toggle="tooltip"]').tooltip();
```
AdSchellevis, line 264: I'm not sure if it matters, but in case existing selectors don't like being triggered twice, you can always add a unique class to the items added and only trigger those,

AdSchellevis (changes requested):

src/www/javascript/opnsense_legacy.js
swhite2: Thanks for the heads up, silly mistake. See new commit.

PR #4793, opened 2021-03-10
LDAP.php: one more error handler

kulikov-a:
Hi!
Can we add just one another LDAP error handler when looking for OUs? It turned out that sometimes it can be a simple time out.
https://forum.opnsense.org/index.php?topic=21605.msg103863#msg103863
Thanks!!

### DIFF
--- a/src/opnsense/mvc/app/library/OPNsense/Auth/LDAP.php
+++ b/src/opnsense/mvc/app/library/OPNsense/Auth/LDAP.php
@@ -444,6 +444,8 @@ class LDAP extends Base implements IAuthConnector
                 $this->logLdapError("LDAP containers search result count: " . $searchResults["count"]);
                 for ($i = 0; $i < $searchResults["count"]; $i++) {
                     $result[] = $searchResults[$i]['dn'];
+                } else {
+                    $this->logLdapError("LDAP containers search returned no results");
                 }
 
                 return $result;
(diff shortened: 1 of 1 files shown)


AdSchellevis:
isn't `$this->logLdapError("LDAP containers search result count: " . 0);` the same as no results?

kulikov-a:
@AdSchellevis it turned out that no )
`$this->logLdapError("LDAP containers search result count: " . 0);` is triggered if any (can be empty) result returned from server
`$this->logLdapError("LDAP containers search returned no results");` is triggered if no result at all. this is the only way we managed to understand that for some reason there is not enough time to complete the search query (we didn’t look for reasons further, as there is no access to server logs. and the topic starter did not change the `timelimit` values in the ldap_search/ldap_list php-functions)

kulikov-a:
@AdSchellevis imho not quite so. the `search` function can return an empty answer, or it can return false (if php ldap_search returns false).
https://github.com/opnsense/core/blob/c3bd32a7fbf74dc14f1ac0862332a5e1cfd2ef11/src/opnsense/mvc/app/library/OPNsense/Auth/LDAP.php#L161-L175
so for ther last case "no results" string will work

AdSchellevis:
@kulikov-a ok, sure, but then your `else` block is in the wrong spot (belongs to the `if` and not to the `for`) if the intend is to to log the lack of results... Thinking about this again, if might be better to remove the result count at all, it's not really an error or warning, just a notice of what's being returned anyway.

https://github.com/opnsense/core/blob/8750a9cea90ed97020daacdac3bb16fa0d98a20b/src/opnsense/mvc/app/library/OPNsense/Auth/LDAP.php#L443-L452

kulikov-a:
Oh! really is the wrong spot!! stupid typo. sorry!

kulikov-a:
@AdSchellevis  thanks alot!

PR #4803, opened 2021-03-11
fix action description

skazi0:


fichtner:
Hmm, correct but I am unsure why we need descriptions here anyway, maybe a historical oversight. Let me check when I have some time available...

skazi0:
@fichtner Sure. I found this by looking at the cron options. There are two with the same description. I'm not sure if there's any use for this one in cron though. Feel free to change it to "remove action description" :o)

src/opnsense/service/conf/actions.d/actions_firmware.conf
```diff
@@ -15,7 +15,7 @@ message:Fetching changelog from remote
 command: /usr/local/opnsense/scripts/firmware/changelog.sh remove
 parameters:
 type:script
-description:Firmware changelog update
+description:Firmware changelog remove
```
fichtner, line 18:
```suggestion
```

Indeed it's better to drop this

fichtner:
Nice catch, thanks!

PR #4807, opened 2021-03-12
default.volt: let translation work on IE11

kulikov-a:
Hi
it turned out that the translation does not work on IE
https://forum.opnsense.org/index.php?topic=21981.0
thanks!


AdSchellevis:
@kulikov-a thanks!

kulikov-a:
@AdSchellevis thanks!!

PR #4811, opened 2021-03-13
OpenVPN: remove checks for NTP servers 3 and 4.

cbrueffer:
OpenVPN: remove checks for NTP servers 3 and 4 since only two optional
NTP servers are exposed through the GUI.


AdSchellevis:
@cbrueffer thanks, there are probably more cleanups possible in this area,  as you noted correctly in the other ticket, the "extra" checkboxes aren't very optimal implemented either.

cbrueffer:
Thanks for merging! Indeed, I'll try to look into some more cleanup as time permits.

PR #4812, opened 2021-03-13
Syslog: Adding support for enabling destination flags

brujoand:
Disclaimer: I currently don't have a development environment setup for OPNsense so I have not tested the contents of the PR. I had one free hour so I thought this would be better that opening an issue asking for someone to implement the support. 

The goal of this PR is to be able to select the flags available on the syslog-ng destination. My use case specifically is to enable the syslog-format flag. I'm currently spending 2 minutes after every upgrade adding this manually so I thought 1 hour now would pay off in a few years time :P

### DIFF
--- a/src/opnsense/mvc/app/controllers/OPNsense/Syslog/Api/SettingsController.php
+++ b/src/opnsense/mvc/app/controllers/OPNsense/Syslog/Api/SettingsController.php
@@ -53,7 +53,7 @@ class SettingsController extends ApiMutableModelControllerBase
     {
         return $this->searchBase(
             "destinations.destination",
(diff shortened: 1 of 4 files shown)


brujoand:
Anything I can do to help move this forward? The lack of this configuration option is making it quite tedious to maintain my logging and monitoring setup for OPNsense.

AdSchellevis:
We need to review and test the PR, at a first glance it looks reasonable. did you get the chance to test the change yourself?

brujoand:
No I couldn't find the time to set up a proper build environment, but I guess it's time to spin up a VM and give it a go.

AdSchellevis:
timeout, let's reopen when someone tested it.

Issue #4815, opened 2021-03-13
Firewall log widget: Selection/filter not working with renamed interface name


wormgat:
**Important notices**

Before you add a new report, we ask you kindly to acknowledge the following:

- [X] I have read the contributing guide lines at https://github.com/opnsense/core/blob/master/CONTRIBUTING.md
- [X] I have searched the existing issues and I am convinced that mine is new.

**Describe the bug**
After renaming "WAN" interface to "WANCBL" the filter is not longer working. Selecting "All" or renaming interface back to "WAN" activates the filter again:
A clear and concise description of what the bug is, including last known working version (if any).

*Tip*: to validate your setup was working with the previous version, use opnsense-revert (https://docs.opnsense.org/manual/opnsense_tools.html#opnsense-revert)

**To Reproduce**

1. Go to 'Lobby / Dashboard / log_title'
2. Click on 'pencil'
3. Click on 'Interfaces to display:'
1.  Select renamed WAN interface

**Expected behavior**
Only entries relating to the selected interface should be shown.

**Screenshots**

Tried to attach the sreenshots from my incomplete ticket #4806, but it shows the following messsage: "Somthing went really wrong, and we can't process that file."
I am sorry for that, I din't learn to operate github, maybe i am to stupid.

**Environment**

Software version used and hardware type if relevant, e.g.:

OPNsense 21.1.3 (amd64, OpenSSL).
APU.2C4 GX-3^1 TC Quad Core
4 GB RAM, 16GB mSATA SSD

wormgat:
Firewall Log Widget still not fixed concerning selection of interface which is renamed.

fichtner:
I’m not sure how you think this works, but if the code didn’t hit the project yet how can something be opportunistically solved in a coincidentally timed release?

wormgat:
I do not know how it works, maybe that's the problem. I just wanted to give feedback. I beg your pardon if I made a wrong impression.

PR #4817, opened 2021-03-14
firewall log widget: get description to compare

kulikov-a:
Hi.
possibly closes https://github.com/opnsense/core/issues/4815
also moved logs fetching to first ajax call completition
can't figure out if a `$("#filterlogentriesinterfaces").val() == intf` comparison is required at all
Thanks!

### DIFF
--- a/src/www/widgets/widgets/log.widget.php
+++ b/src/www/widgets/widgets/log.widget.php
@@ -82,8 +82,13 @@ $nentriesinterfaces = isset($config['widgets']['filterlogentriesinterfaces']) ? 
         const field_type_icons = {'pass': 'fa-play', 'block': 'fa-ban', 'rdr': 'fa-exchange', 'nat': 'fa-exchange'};
 
         var interface_descriptions = {};
+        var selected_description = "";
         ajaxGet('/api/diagnostics/interface/getInterfaceNames', {}, function(data, status) {
             interface_descriptions = data;
+            if ($("#filterlogentriesinterfaces").val() !== "") {
+                selected_description = $("#filterlogentriesinterfaces option:selected").text().trim().toLowerCase();
+            }
+            fetch_log();
         });
         function fetch_log(){
             var record_spec = [];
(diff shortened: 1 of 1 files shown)


fichtner:
I'm kind of tripping over the lowercase match since we try to use uppercase everywhere so it would seem the controller makes some nonconforming naming decisions that probably also affect the firewall live log page. It might be better to move the required fix closer to the source of the problem?

kulikov-a:
@fichtner to be honest, I do not quite understand why the `toLowerCase()` is used before the comparison. as well as why the comparison goes with `val()`. so I just decided not to touch it. I will test the behavior without a `toLowerCase()` .

eventually `#filterlogentriesinterfaces` can be filled on the client side with the same api call
moreover, if I understand correctly, it will not be necessary to convert the interface to a description for each log entry. we can compare directly the interface from the log entry with the `val()` in the `$("#filterlogentriesinterfaces").val()`

kulikov-a:
@fichtner @AdSchellevis 
Hi. sorry, what is the likelihood that I will break something if I make  `strtoupper`  in `                $intfmap[(string)$node->if] = !empty((string)$node->descr) ? (string)$node->descr : $key;` ?:
https://github.com/opnsense/core/blob/7d0b4b539b70681e4126415ea6957dc47a863b95/src/opnsense/mvc/app/controllers/OPNsense/Diagnostics/Api/InterfaceController.php#L45-L54
since `legacy_config_get_interfaces` returns it in uppercase:
https://github.com/opnsense/core/blob/7d0b4b539b70681e4126415ea6957dc47a863b95/src/etc/inc/config.inc#L293
and is there any objection if I try to add the ability to select multiple interfaces for the widget?

AdSchellevis:
@kulikov-a maybe the better question is what's the impact if we remove the upper from the legacy part?

AdSchellevis:
@kulikov-a removed previous reaction, my mistake, the description itself shouldn't break anything in my opinion.

kulikov-a:
@AdSchellevis thanks! widget works well with `strtoupper` in API and client-side `#filterlogentriesinterfaces` rendering (also loopback interfaces showed up). trying to convert it to multiple

kulikov-a:
clearly a workaround, not a solution. replaced by https://github.com/opnsense/core/pull/4825

Issue #4762, opened 2021-02-27
Bug in DHCP server Available range


x1alfa:
Hi, I noticed this in 20.1.1, not sure if the same behavior exist in older version, same is exist in 20.1.2, I was able to reproduce in two different installation.

Interfaces (LAN, OPTx) \ IPv4 Configuration Type (Static IPv4) \ IPv4 address (x.x.x.x/32)
Services \ DHCPv4 \  Available range (x.x.x.2 - x.x.x.0)

With /32 sub-net no IPv4 range should be available for DHCP, plus the range is backward.

I hope reporting this bug help improving OPNSense.

x1alfa:
I'm not a PHP programmer, but looking at the code I think the quickest fix is to add a validation line if subnet is 32 then set start IP and end IP to the same IP of the interface

services_dhcp.php (line 709)

<?php
          if ($config['interfaces'][$if]['subnet']) = 32  {
$range_from = $config['interfaces'][$if]['ipaddr'];
$range_to = $config['interfaces'][$if]['ipaddr'];
} ?>
<?=$range_from;?> - <?=$range_to;?>

<?php
else {
$range_from = ip2long(long2ip32(ip2long($config['interfaces'][$if]['ipaddr']) & gen_subnet_mask_long($config['interfaces'][$if]['subnet']))) + 1;
$range_to = ip2long(long2ip32(ip2long($config['interfaces'][$if]['ipaddr']) | (~gen_subnet_mask_long($config['interfaces'][$if]['subnet'])))) - 1;}?>
<?=long2ip32($range_from);?> - <?=long2ip32($range_to);?>

IShouldHaveBeenAQATester:
This is also an issue with /128 on ipv6.  Also wondering if a /31 or /127 is handled properly.

fichtner:
I'm assuming nobody uses /31 since the validation fails and there are no bugs for it. So /30 is required and will only then show an available range although it won't try to poke a hole into it for the IP address used by the interface.

In IPv6 /128 suffers the same fate, but /127 works since we don't have a network or broadcast address to reserve.

For more details see c7c6299.


Cheers,
Franco

PR #4822, opened 2021-03-15
Add dhcp menu validation, hide interface with subnet mask greater or equal 31

x1alfa:
Proposed fix for bug #4762, a simple validation if subnet mask is greater or equal 31, 31 and 32 simply does not have enough addresses to be used by DHCP, we can expand to subnet mask 30, still 30 can be excluded as it technically has 2 hosts that can be used (including the IP assigned to firewall).
This was tested with virtual OPNSense using the latest master.
Note this is the first time ever I contribute code to GitHub so excuse me if I did something wrong, new learner.


fichtner:
Menu and redirect logic for DHCP pages is already very fragile with unnecessarily strict validation and I don't see why we should expand this. I'm happy to add a warning/error in #4762 because /32 is static, but likely a user mistake and we don't want to keep user guessing why DHCP is not available.

PR #4832, opened 2021-03-16
Add Option for Vendor-Class-ID by pool

marjohn56:
You can do pool dhcp using MAC addresses, but doing it by Vendor-Class is another option, very useful for phones or any devices where the vendor class Id is sent.


AdSchellevis:
@marjohn56 I'm not going to pull this in now, the code can be simplified but before I'm going to spend time on this I do want to wait if more people are actually planning to use this (and perhaps offer to write documentation to explain its use). We can always pull it in at a later time, although at some point we should either decide to rebuild it in mvc (in which case adding features, means adding technical debt) or refactor functions like `dhcpd_dhcp4_configure()` to make sure their readable and extendable.

Issue #4638, opened 2021-01-25
Show client’s real IP when connecting to OPNsense webgui via reverse proxy


Greelan:
**Important notices**

Before you add a new report, we ask you kindly to acknowledge the following:

- [X] I have read the contributing guide lines at https://github.com/opnsense/core/blob/master/CONTRIBUTING.md
- [X] I have searched the existing issues and I am convinced that mine is new.

**Is your feature request related to a problem? Please describe.**

When the OPNsense webgui is accessed through a reverse proxy, OPNsense logs the IP of the reverse proxy rather than the real IP of the client connecting to the reverse proxy. Therefore every logged access to the webgui appears from the same IP, even if different clients are connecting.

**Describe the solution you like**

I would like the webgui to log instead the real IP of the client connecting via the reverse proxy.

This can be achieved by implementing the `mod_extforward` module in lighttpd.

Further details are set out in this forum discussion: https://forum.opnsense.org/index.php?topic=20882.0.

As noted in that discussion (see in particular replies 3 and 4), there are three elements:

- First the `mod_extforward` module needs to be enabled (and in the right order). This could be done by default, or optionally by the user clicking a checkbox if they are using a reverse proxy (which would then add the module to the module list in `webgui.inc`, for example by appending it to the `$lighty_modules` variable, after `mod_accesslog` if that is enabled).
- Second, `extforward.headers` needs to be set, at least until a bugfix that has been recently implemented in the upstream makes it into OPNsense.
- Finally, `extforward.forwarder` needs to be set to the trusted IP(s) of the reverse proxy. This will obviously differ from user to user, and so needs to be able to be set by the user. A data entry field in the webgui would enable this, or failing that at least an ability to manually configure a configuration file on the system.

**Describe alternatives you considered**

None.

**Additional context**

See above link to forum discussion on this, including relevant code for `webgui.inc`.

fichtner:
Hi @Greelan,

Can you ping me again when 21.1 is out? This week is a little busy but as per forum discussion I promised to help out. :)


Thanks,
Franco

Greelan:
Very happy to. Just wasn't sure about protocol for requests such as this, but thought ultimately this was best on GH 😀

Greelan:
Hi @fichtner

Here's that little nudge as requested, for when you finish catching up on sleep after the 21.1 marathon 😀

Greelan:
[...]

PR #4833, opened 2021-03-16
webgui: implement lighttpd mod_extforward module

Greelan:
This is a simple implementation of the mod_extforward module in lighttpd, in response to #4638. It relies on enabling the module by default, and allowing the user to configure a configuration file in an include directory to set the `extforward.forwarder` directive. A more sophisticated implementation would allow the user to enable the module by a checkbox in the webgui (rather than enabling it by default), and also set the trusted IPs in the webgui rather than by a manual configuration file. That is for future work :)


AdSchellevis:
@Greelan it might also be an option to generalise it a bit more and include a directory like `/usr/local/etc/lighttpd_webui/conf.d/` with a readme how to configure additional settings (like extforward).  lighthttp seems to support` server.modules += ("xxx")` in these files and we can document this on our end like we do for some other advanced features as well. What do you think?

fichtner:
It’s what I had in mind when I offered to help on this (sorry btw). Proposal looks promising and with generalisation that is perfect and easy to extend from a future plugin (original concern).

kulikov-a:
@Greelan @AdSchellevis @fichtner 
sorry. do not quite understand. if these are the only directives requested, why not add it directly via the GUI-config-webgui.inc? Or are there even more features planned for the GUI (eg OCSP stapling) via plugin?
thanks!

fichtner:
We've had good experience with the pluggable file inclusion approach (per include directory) and most of the MVC service templating use it as well. It helps to move components to MVC that currently are static, can help split complexity of single configuration files and  also allows users to inject local changes into these files without adding GUI options that are misinterpreted or flagged as bugs for working within their constraints but not user expectations. ;)

kulikov-a:
Ah. got it. thanks!

Greelan:
@fichtner @AdSchellevis thanks for the feedback. I see the sense in your approach. It looks like this is similar in concept to what lighttpd itself is structured around, as seen in `/usr/local/etc/lighttpd/conf.d/`. Indeed with this approach example configs could be added over time to help users.

I should be able to get around to this by the weekend and will aim for submitting a cleaner commit. 😀

@kulikov-a my preference for just the extforward module request was indeed to have this in `webgui.inc`, just for ease of configuration. At the moment that is beyond my ken, but I was hoping down the track I would have time to review similar examples of config options in the existing codebase that I could adapt to get the checkbox, data entry fields and structure to create the `extforward.forwarder` directive.

Greelan:
Superseded by #4837

PR #4837, opened 2021-03-18
webgui: lighttpd include directory for configuration

Greelan:
This PR is to replace #4833 and reflects the feedback as to the preferred approach outlined there. It is intended to address #4638 but in a more generic way that will allow configuration of other lighttpd modules

### DIFF
--- a/src/etc/inc/plugins.inc.d/webgui.inc
+++ b/src/etc/inc/plugins.inc.d/webgui.inc
@@ -261,8 +261,11 @@ EOD;
 ## modules to load
 server.modules              =   (
   "mod_access", "mod_expire", "mod_deflate", "mod_redirect", "mod_setenv",
-  "mod_cgi", "mod_fastcgi","mod_alias", "mod_rewrite", "mod_openssl" {$lighty_modules}
+  "mod_cgi", "mod_fastcgi", "mod_alias", "mod_rewrite", "mod_openssl" {$lighty_modules}
 )
+
+## additional optional modules to load or additional module configurations
+$lighty_config .= "include \"/usr/local/etc/lighttpd_webgui/conf.d/*.conf\"\n";
 
 server.max-keep-alive-requests = 15
 server.max-keep-alive-idle = 30
--- a/src/etc/lighttpd_webgui/conf.d/README
+++ b/src/etc/lighttpd_webgui/conf.d/README
@@ -19,6 +19,10 @@ To use the functionality of this directory:
 
 - specify the relevant configurations for the module, according to the lighttpd documentation
 
-In some cases, the order in which modules are enabled by lighttpd is important, in which case you will need to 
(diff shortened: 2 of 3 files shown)


AdSchellevis:
@Greelan looks good, if @fichtner is also ok with this we can merge it to master

fichtner (approved):
Looks good. We should remove $lighty_modules post-merge using the same trick Ad suggested in the other PR.

Greelan:
Agree, as long as mod_accesslog is enabled in the order that it currently is through $lighty_modules. It needs to be enabled before mod_extforward, for example, or otherwise mod_extforward won't affect IP logging in the access log

Greelan:
One possible solution to the issue I raised in the comment on my last commit is to have three conf files in the include directory that have the default modules grouped into buckets (`01-default` for default modules that should be loaded first, `10-default` for those second, `20-default` for those last). This is in place of loading them directly in `webgui.inc`. Then a user has the option of including their own confs in the right order “between” those default confs, so that the correct order is achieved.

AdSchellevis:
@Greelan Thanks! In principle I'm not against moving more or less static sections to standard configuration files, it removes custom plumbing from the legacy code and is indeed easier to inject custom overwrites in between as needed (only for advanced users, handle with care :) ). The access log settings can easily be migrated to a jinja template.

Greelan:
@AdSchellevis I am happy to continue tinkering with this if I can get some guidance as to the preferred structure for the default modules and the jinja template - eg any similar structures that already exist for other packages that you could point me to?

AdSchellevis:
@Greelan I think syslog-ng sort of has a similar approach, we didn't refactor the ui pages, but did move the logic into Jinja templates (https://github.com/opnsense/core/tree/master/src/opnsense/service/templates/OPNsense/Syslog). There already is a spot for the web gui, since we generate the php requirements as well (https://github.com/opnsense/core/tree/master/src/opnsense/service/templates/OPNsense/WebGui). We could either add a submodule (subdirectory in there for the conf.d directory) or just start moving settings into a single file, although the first option is probably easiest to read and understand. 

So if you want to take a shot at it, go ahead, I'll try to review and comment when there's a PR.

fichtner:
I don’t mind a rework, but do we need to pick the central web GUI for this if we can already handle the drop ins now? ;)

MattElekHarris:
Is there a target version for this to be included in? Thanks for all your work!

Greelan:
I can't answer that, but you can install it (ie the drop-in directory configuration) now with `opnsense-patch bdf7409 f0e26fe`

MattElekHarris:
Awesome, thanks!

Greelan:
Actually, you might be better with `opnsense-patch 77d562f 24cca59`. The commits above were my original commits. I'm going to test just now

PR #4840, opened 2021-03-18
firmware/health.sh: refine missing/invalid signature message

knan-md:
To avoid self-contradictory output:

>>> Check installed kernel version
Version 21.1.3 is correct.
Cannot verify kernel: missing /usr/local/opnsense/version/kernel.mtree.sig
>>> Check for missing or altered kernel files
No problems detected.

Now we will instead get:

>>> Check installed kernel version
Version 21.1.3 is correct.
Cannot verify kernel: missing /usr/local/opnsense/version/kernel.mtree.sig

### DIFF
--- a/src/opnsense/scripts/firmware/health.sh
+++ b/src/opnsense/scripts/firmware/health.sh
@@ -86,8 +86,10 @@ set_check()
 
 	if [ ! -f ${FILE}.sig ]; then
 		echo "Cannot verify ${SET}: missing ${FILE}.sig" | ${TEE} ${LOCKFILE}
+		return
 	elif ! opnsense-verify -q ${FILE}; then
 		echo "Cannot verify ${SET}: invalid ${FILE}.sig" | ${TEE} ${LOCKFILE}
+		return
 	fi
 
 	echo ">>> Check for missing or altered ${SET} files" | ${TEE} ${LOCKFILE}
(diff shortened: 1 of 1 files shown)


fichtner:
In general I agree but the situation is a bit tricky: signature is missing but the mtree file may be there so you still need to run the check but complain about the signature. If there is no mtree file it makes sense to skip it entirely.

fichtner:
Sorry I stand corrected and that code is already there. We can't return here in both cases, but I'm happy to improve the wording.

knan-md:
Slightly disagree - I think running the check doesn't really make sense when you can't verify that the mtree file is from a trusted source. What are you then verifying?

fichtner:
If you have your own build and don't set up keys you can still run the mtree check without the signature. If the signature got corrupted by the file system you can still run your mtree check to see what other files are affected. If the mtree file is screwed so be it, but there is really no way to check for that without a valid signature which means mtree verification is and should still be best effort.

knan-md:
I take your point. How about this?

fichtner:
Looks promising. How about "hashes" -> "consistency check"? It's not just hash comparison, see mtree file snippet:

```
/set type=file uid=0 gid=0 mode=0555
kernel          type=dir mode=0755
    aac.ko      size=88904 \
                sha256digest=39c97b69bb0e78ca73c424f4beadfd249340fd9a99afaff70952aad4f71f8393
    aacraid.ko  size=83568 \
                sha256digest=c69df4c5ff970459c642dcbb3cd4599ad17215f1ca27caa197186d1e4e8daff7
```

knan-md:
Sure. Updated.

fichtner:
Merged, thanks!

PR #4846, opened 2021-03-19
fw categories: error if the category name is a number

kulikov-a:
Hi
javascript throws an error if rule is in one category and category name is a number
Thanks!

### DIFF
--- a/src/www/javascript/opnsense_legacy.js
+++ b/src/www/javascript/opnsense_legacy.js
@@ -191,7 +191,7 @@ function hook_firewall_categories() {
             let category_count = {};
             $(".rule").each(function(){
                 let row = $(this);
-                $(this).data('category').split(',').forEach(function(item){
+                $(this).data('category').toString().split(',').forEach(function(item){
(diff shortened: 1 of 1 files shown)


AdSchellevis:
@kulikov-a nice catch! it seems that statically defined data tags are type casted automatically, so `data-category="X" ` doesn't match type with `data-category="1" ` .

kulikov-a:
@AdSchellevis thanks!!
looks like a jquery featue (not sure if it's that good if the behavior cannot be changed inside the method): https://api.jquery.com/data/#data-html5

kulikov-a:
@AdSchellevis hm. sorry, after reading jquery docs once again..now I doubt, maybe it would be quite correct to replace with `$(this).attr('data-category').split(',').forEach` once $ can repeat it with boolean?

AdSchellevis:
@kulikov-a I think the toString() is safe enough to be honest

kulikov-a:
of course it should still work. thanks again )

PR #4853, opened 2021-03-21
Add new mirror in Russia

xshishkin:
Hi!

I propose to add a new mirror located in Russia.

Available protocols: http, https, rsync
Bandwidth: 200Mb/s
Country: Russia, Nizhny Novgorod
IPv4 — 95.79.25.1
IPv6 — 2a03:1ac0:5f4f:1901::6

Contact email — dmitry@shishkin.email

### DIFF
--- a/src/opnsense/mvc/app/controllers/OPNsense/Core/Api/repositories/opnsense.xml
+++ b/src/opnsense/mvc/app/controllers/OPNsense/Core/Api/repositories/opnsense.xml
@@ -29,6 +29,10 @@
             <description>Cloudfence (HTTPS, Sao Paulo, Brazil)</description>
         </mirror>
         <mirror>
+            <url>https://mirror.surf/opnsense</url>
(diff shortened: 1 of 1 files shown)


AdSchellevis:
@xshishkin we currently only add mirrors to the website, so people can manually add them in OPNsense if they want to use alternative sources. We don't always know if mirrors are being taken offline, which leads to quite some noise on our end keeping the mirrors file up to date. If you want your mirror to be added, just drop us an email at project at opnsense.org containing the name we should use.

xshishkin:
Thanks for the answer. Sent an e-mail request.

I understand that many volunteer mirrors quickly go offline... But I still suggest considering the option of adding a mirror to the general list, which has been active for more than six months or a year.

AdSchellevis:
I've just added it on our website (https://opnsense.org/download/), maybe in time we'll reconsider our mirror policy.

PR #4858, opened 2021-03-22
unbound: add option to validate server certificate

aque:
This PR adds an option for unbound to validate the certificates of DNS-over-TLS servers. Here is the relevant section from `man unbound.conf`:

```
If tls is enabled, then you can append a '#' and a
name, then it'll check the tls authentication certificates with
that name.  If you combine the '@' and '#', the '@' comes first.
```

I also took the opportunity to split the mask regex into IP, port, and name sections - and set a max of 64 entries for the field.

### DIFF
--- a/src/opnsense/mvc/app/controllers/OPNsense/Unbound/forms/miscellaneous.xml
+++ b/src/opnsense/mvc/app/controllers/OPNsense/Unbound/forms/miscellaneous.xml
@@ -13,6 +13,6 @@
         <type>select_multiple</type>
         <style>tokenize</style>
         <allownew>true</allownew>
-        <help>List of nameservers to use for DoT. Use syntax ip@port like 9.9.9.9@853</help>
+        <help>List of nameservers to use for DoT. Use syntax ip@port like 9.9.9.9@853 or ip@port#fqdn like 9.9.9.9@853#dns.quad9.net to validate the server certificate.</help>
     </field>
 </form>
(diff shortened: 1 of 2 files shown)


AdSchellevis:
It might be better to move these items into a grid layout so attributes can be validated separately, the input becomes a bit convoluted if we try to push multiple functional items into a single item. @mimugmail what do you think?

fichtner:
I think Michael and me talked about it. I also think that splitting these up makes sense, but requires migration and other work... it’s probably in the initial PR for DoT.

AdSchellevis:
It the plan is to split the values eventually, I think we should prevent increasing technical debt and leave this for now (convert to a ticket?).

fichtner:
Ok, move this back to core then? https://github.com/opnsense/plugins/issues/1778

AdSchellevis:
To be honest, a ticket describing what the functionality should do in relation to the existing feature would probably be better, but I see the overlap. 

Adding new plugins for unbound is probably not a great idea anyway. Maybe I can spend a bit of time somewhere before 21.7 to move settings around, although it will help to validate where this all fits in and if there's overlap with the exiting legacy pages....

mimugmail:
I believe I have a half-ready branch supporting grids, but it lacks migration script which @fichtner wanted to have a look at.
If you like I can rebase the PR from plugins to core (I guess I started work at plugins level)

AdSchellevis:
@mimugmail sure, the migration part shouldn't be very complicated, only question is if there's overlap with the existing legacy functionality which we should migrate in the same pass.

fichtner:
This was actually fixed in 21.7.

PR #4868, opened 2021-03-26
Firewall Rules: show loopback auto-rules

kulikov-a:
Hi!
ref. https://forum.opnsense.org/index.php?topic=22278.0
Probably more of a question than a request and definitely not urgent.
Now the "Automatically generated rules" 'pass loopback' and 'Pass all loopback IPv6' for the Loopback interface are not displayed in Firewall: Rules: Loopback.
confusing a little..
and it seems that this is due to that this rules is registered with the `loopback` interface, and the GET request goes with the value `if=lo0`.
if, for these rules, only `lo0->loopback` is mapped (not `lo` group)
https://github.com/opnsense/core/blob/ac8bb09c6b912275490aee2720e59009eda9dd8f/src/opnsense/mvc/app/library/OPNsense/Firewall/Plugin.php#L77-L82
then perhaps we can register rules directly with the lo0 interface? then these rules are displayed in the GUI for Loopback if.
rule management is too complex and I'm probably missing something

thanks!

### DIFF
--- a/src/etc/inc/filter.lib.inc
+++ b/src/etc/inc/filter.lib.inc
@@ -211,7 +211,7 @@ function filter_core_rules_system($fw, $defaults)
     // block All IPv6 except loopback traffic
     $fw->registerFilterRule(
         1,
-        array('interface' => 'loopback', 'ipprotocol' => 'inet6', 'disabled' => isset($config['system']['ipv6allow']),
+        array('interface' => 'lo0', 'ipprotocol' => 'inet6', 'disabled' => isset($config['system']['ipv6allow']),
           'descr' => 'Pass all loopback IPv6', '#ref' => 'system_advanced_firewall.php#ipv6allow'),
         $defaults['pass']
     );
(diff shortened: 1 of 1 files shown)


fichtner:
I suspect there is some internal plumbing and rewrite that should die with said transition (loopback => lo0 correction somewhere internally), but avoiding breakage in all edge cases is difficult...

AdSchellevis:
We added loopback last year (https://github.com/opnsense/core/commit/57bd1f2b05b75e0d7b0da1a9cd082ce3a9ce290f), which was at first more or less an experiment to be able to configure direct rules on loopback (partly because additional loopbacks don't receive loopback traffic, https://github.com/opnsense/src/issues/65). 

Now it's in, we should at least try to make sure it's as transparant as possible, will check this out when I have some time available.

kulikov-a:
@fichtner @AdSchellevis thanks for the clarifications! 
I'll leave the changes from the request on my test vm and take a look at rules management behavior (still playing with spinning icons) - maybe some side effect will come out.
is there anything I should do with this request? (convert to draft, close?)

AdSchellevis:
just leave it as is, I'll take it from here

AdSchellevis:
@kulikov-a can you try https://github.com/opnsense/core/commit/bd26a58820716c0f412ed8917f9259d87c5a587b ? the loopback seemed to be redundant indeed.

kulikov-a:
@AdSchellevis works great! thanks!

AdSchellevis:
@kulikov-a thanks for conforming!

PR #4884, opened 2021-03-31
Fix missing ')'

pkejval:
Fix missing ')' at the end of log line:

before fix: blacklist download https://block.energized.pro/bluGo/formats/hosts (lines: 125888 exclude: 5 black: 125814
after fix: blacklist download https://block.energized.pro/bluGo/formats/hosts (lines: 125888 exclude: 5 black: 125814)


AdSchellevis:
@pkejval thanks!

PR #4891, opened 2021-04-02
Potential fix for shutdown order of services

iamperson347:
**Problem:**
It appears that, depending on what services are being utilized, a reboot may hang due to a dependent service being shut down prior to the depending service. The example here is redis and ntopng.

Note - this problem also affects upgrades that require a reboot. It also affects the reboot page within the opnsense GUI.

**Proposed Fix:**
I'm unsure if this fix would have any ill effects. However, when calling rc.freebsd with "stop", if we flip the order of services gathered by rcorder and shutdown services per the new list, it appears that reboots performs cleanly.

**Other details:**

- The current OPNsense version where the bug first appeared
     - Unknown - appears to affect latest release
-  The last OPNsense version where the bug did not exist
    -  Unknown
-  The exact URL of the GUI page involved (if any)
    - N/A
-  A list of steps to replicate the bug
    -  Install redis and ntopng on an opnsense install. Reboots will hang on stopping ntopng due to redis being shut down first

### DIFF
--- a/src/etc/rc.freebsd
+++ b/src/etc/rc.freebsd
@@ -117,6 +117,11 @@ if [ "${1}" == "start" ]; then
 	done
 fi
 
+if [ "${1}" == "stop" ]; then
+	rc_filenames=$(echo "${rc_filenames}" | tail -r)
+	rc_filenames_defer=$(echo "${rc_filenames_defer}" | tail -r)
+fi
+
 # pass all commands to script now
 for rc_filename in ${rc_filenames} ${rc_filenames_defer}; do
 	eval "$(grep "^name[[:blank:]]*=" ${rc_filename})"


fichtner:
Hi David,

Thanks for your patch. tail -r wasn't working on space-separated list so I rewrote it via 7316071. Can you try it on your end?


Thanks,
Franco

iamperson347:
Hi @fichtner,

It looks like the version you put together works as intended. ntopng is stopped before redis (as expected based on the reverse order).

Thank you for getting this in place!

```
root@OPNsense:/usr/local/etc # ./rc.reboot
>>> Invoking stop script 'beep'
>>> Invoking stop script 'freebsd'
Stopping ntopng.
Waiting for PIDS: 37061.
Stopping radiusd.
Waiting for PIDS: 27848.
Stopping redis.
Waiting for PIDS: 57826.
cannot unmount '/var/log': Device busy
cannot unmount '/tmp': Device busy
>>> Invoking stop script 'backup'
>>> Invoking backup script 'captiveportal'
>>> Invoking backup script 'dhcpleases'
>>> Invoking backup script 'duid'
>>> Invoking backup script 'netflow'
>>> Invoking backup script 'rrd'
>>> Invoking stop script 'config'
Shutdown NOW!
shutdown: [pid 36402]
root@OPNsense:/usr/local/etc #
*** FINAL System shutdown message from root@OPNsense.localdomain ***

System going down IMMEDIATELY



System shutdown time has arrived
```

fichtner:
Loos good then. Probably material for 21.1.5. Thank you. ❤️

Issue #4867, opened 2021-03-25
IPv6 Route Advertisement RDNSS lifetime is too short per RFC recommendation.


Napsterbater:
**Important notices**

Before you add a new report, we ask you kindly to acknowledge the following:

- [x] I have read the contributing guide lines at https://github.com/opnsense/core/blob/master/CONTRIBUTING.md
- [x] I have searched the existing issues and I am convinced that mine is new.

**Describe the bug**

IPv6 Route Advertisement RDNSS lifetime is too short per RFC recommendation.
RFC8106 5.1 says RDNSS Lifetime "SHOULD by default be at least 3 * MaxRtrAdvInterval"

additonally

> The rationale of the default value of the Lifetime field is as
>          follows.  The Router Lifetime field, set by AdvDefaultLifetime,
>          has the default of 3 * MaxRtrAdvInterval as specified in
>          [RFC4861], so such a default or a larger default can allow for
>          the reliability of DNS options even under the loss of RAs on
>          links with a relatively high rate of packet loss.  Note that
>          the ratio of AdvDefaultLifetime to MaxRtrAdvInterval is the
>          number of unsolicited multicast RAs sent by the router.  Since
>          the DNS option entries can survive for at most three
>          consecutive losses of RAs containing DNS options, the default
>          value of the Lifetime lets the DNS option entries be resilient
>          to packet-loss environments.

**To Reproduce**

Simply enable Router Advertisements with DNS server configured 

**Expected behavior**

RDNSS Lifetime should be 3 times the Maximum Interval

**Describe alternatives you considered**

OPNsense should add to the radvd config:

`AdvRDNSSLifetime 3060
`
this would go in the same block at the RDNSS changing

`RDNSS 2001:470:****:1::3 2001:470:****:2::8 { };
`
to
```
         RDNSS 2001:470:****:1::3 2001:470:****:2::8 {
         AdvRDNSSLifetime 3060 
         };
```
In this case 3060 is 3 times 1020, my configured Maximum Interval

**Relevant log files**

RA from Packet Cap

> Internet Control Message Protocol v6
>     Type: Router Advertisement (134)
>     Code: 0
>     Checksum: 0xa4b8 [correct]
>     [Checksum Status: Good]
>     Cur hop limit: 64
>     Flags: 0xc0, Managed address configuration, Other configuration, Prf (Default Router Preference): Medium
>     Router lifetime (s): 3060
>     Reachable time (ms): 0
>     Retrans timer (ms): 0
[...]

PR #4893, opened 2021-04-02
radvd advanced options: remove AdvRDNSSLifetime / AdvDNSSLLifetime bounds

maurice-w:
RFC 8106 removes the bound of acceptable values:
https://tools.ietf.org/html/rfc8106#section-5.1
https://tools.ietf.org/html/rfc8106#section-5.2

radvd does not enforce a maximum. 2.19 enforces a minimum (MaxRtrAdvInterval), but does not fail if a lower value is configured. All bounds have been removed in master: radvd-project/radvd@17c43bff200bbc1d4786eb917860a79db4f95c2e

Context: #4867


fichtner:
Merged manually via 3807cf8 -- thanks!

PR #4894, opened 2021-04-02
NPTv6 configuration clean-up

maurice-w:
- The length of the internal and external prefix must match. If non-identical prefix lengths are selected, an error is logged (_'binat' source mask and redirect mask must be the same_) and no pf rules are created. Remove the external prefix length selector and use the internal prefix length for both.
- If either one (or both) of the "Invert" options are enabled, an error is logged (_syntax error_) and no pf rules are created. Remove.
- The "Source" / "Destination" terminology is a little confusing and doesn't match the "Internal" / "External" terminology used on the rule overview page. Rephrase.

### DIFF
--- a/src/www/firewall_nat_npt.php
+++ b/src/www/firewall_nat_npt.php
@@ -229,7 +229,7 @@ $main_buttons = array(
                       <th style="width:2%">&nbsp;</th>
                       <th><?=gettext("Interface"); ?></th>
                       <th><?=gettext("External Prefix"); ?></th>
-                      <th><?=gettext("Internal prefix"); ?></th>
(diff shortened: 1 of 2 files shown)


AdSchellevis:
@maurice-w this makes sense indeed, technically NPTv6 is just a binat rule as you can configure with 1-to-1. I don't think a negate works either with these rules, but if you would want to match specific nets you would need to use a 1-to-1 anyway as you can't specify the actual destination (only the translation target, in which case destination is quite confusing). 


I'll merge this to master, not sure if we addd it to 21.1 yet, but it will be in 21.7. Thanks!

Issue #4859, opened 2021-03-23
Add Ability to Save Filters in Firewall Live Log View 


borkedporcupine:
**Important notices**

Before you add a new report, we ask you kindly to acknowledge the following:

- [ x ] I have read the contributing guide lines at https://github.com/opnsense/core/blob/master/CONTRIBUTING.md
- [ x ] I have searched the existing issues and I am convinced that mine is new.

**Is your feature request related to a problem? Please describe.**

This is a feature request, not related to a problem. 

**Describe the solution you like**

On the firewall live log page, /ui/diagnostics/firewall/log I think it would be useful if we could save filters for common filtering tasks.  I find myself doing repetitive tasks when digging through logs, would like to be able to save the filters for quick filtering. 

**Describe alternatives you considered**

Alternative is a manually choosing filters each time. 

**Additional context**

Something like this is what I would like to see (down toward the bottom) https://campus.barracuda.com/product/webapplicationfirewall/doc/78810008/types-of-logs/

AdSchellevis:
https://github.com/opnsense/core/issues/4731 sounds quite similar, since this template is more complete, I'll close the other one.

kulikov-a:
Hello everyone!
did I get the idea right? (grid-MVC for templates storing. quick manipulation directly on Live View)

https://user-images.githubusercontent.com/36099472/113257088-c290f780-92d2-11eb-9f44-a3d9420967cf.mp4

borkedporcupine:
> Hello everyone!
> did I get the idea right? (grid-MVC for templates storing. quick manipulation directly on Live View)
> 
>  lv-templ.mp4

@kulikov-a This looks fantastic.

kulikov-a:
Thanks! then i continue with this approach)
I did not plan any other functions so as not to clutter up the interface (even editing is not planned: "select->modify->del->add with old name"  are enough imho)
please tell if something is missing

kulikov-a:
Hi everyone.
Me again. All functions are ready, template editing is added. If there are no other suggestions and comments, then I'm ready to clean up the code and prepare a pr.

https://user-images.githubusercontent.com/36099472/113511741-21bc6980-956a-11eb-9ae7-445a275b8fd0.mp4

PR #4902, opened 2021-04-06
Firewall Live View: stored filters templates

kulikov-a:
Hi!
may close https://github.com/opnsense/core/issues/4859
- simple model for storing templates
- managing templates directly on Live View page via api
- MVC view is added but not used in menu. for MVC work debug only
- SORT_FLAG_CASE flag added in API to return case insensitive sorted results (or maybe it makes sense to add this directly to the ApiMutableModelControllerBase?)

### DIFF
--- a/src/opnsense/mvc/app/controllers/OPNsense/Diagnostics/Api/LvtemplatesController.php
+++ b/src/opnsense/mvc/app/controllers/OPNsense/Diagnostics/Api/LvtemplatesController.php
@@ -1,0 +1,68 @@
+<?php
+
+/**
+ *    Copyright (C) 2021 Deciso B.V.
+ *
+ *    All rights reserved.
+ *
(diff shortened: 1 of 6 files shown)


AdSchellevis:
@kulikov-a I've refactored it a bit in https://github.com/opnsense/core/commit/a785f717555e19417a7e999e283565d09799f8b8, may need to check a bit closer later, but generally it looks nice. Thanks!

kulikov-a:
@AdSchellevis  
Hi! wow, Thanks!!!
thanks for the refactoring - I will consider it next time (especially the funcs placement order).
just one thing about `fetchTemplates()` refactor. will comment on commit

AdSchellevis:
@kulikov-a thank you, you did most of the work here :) Maybe there  are still some rough edges, but we'll polish them out sooner or later.

PR #4912, opened 2021-04-10
Live View: keep url-filters applied on first load

kulikov-a:
Hi!
from https://github.com/opnsense/core/commit/a785f717555e19417a7e999e283565d09799f8b8#r49352692
sorry, didn't quite understand how it is more convenient: via pr or direct edit

### DIFF
--- a/src/opnsense/mvc/app/views/OPNsense/Diagnostics/fw_log.volt
+++ b/src/opnsense/mvc/app/views/OPNsense/Diagnostics/fw_log.volt
@@ -28,6 +28,7 @@
 <script>
     'use strict';
 
+    var t_fetched = $.Deferred();
     $( document ).ready(function() {
         var field_type_icons = {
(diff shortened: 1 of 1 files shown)


AdSchellevis:
Hi @kulikov-a, no problem at all, it's probably better to move the deferred into the function by the way, just like https://github.com/opnsense/core/blob/af587e812a7b8d96cf3fd84869dc7e00bbfe09b3/src/opnsense/www/js/opnsense_ui.js#L112-L144

and then you can trigger like :

https://github.com/opnsense/core/blob/af587e812a7b8d96cf3fd84869dc7e00bbfe09b3/src/opnsense/www/js/opnsense_bootgrid_plugin.js#L200

sorry for the noise, I might have misread your previous commit and thought it was similar to this.

kulikov-a:
@AdSchellevis  what noise? - thanks for learning!

kulikov-a:
replaced by https://github.com/opnsense/core/pull/4913

PR #4913, opened 2021-04-10
Live View: set url-filters after templates load

kulikov-a:
Hi!
replaces https://github.com/opnsense/core/pull/4912


AdSchellevis:
@kulikov-a that's it, thanks. Our frontend code just isn't as structured as the backend, which sometimes makes the volt templates a bit harder to read then one would like too. We all learn every day, thanks a lot for your patience and doing to work on this. It's a nice feature that probably quite some people will learn to value.

kulikov-a:
@AdSchellevis thanks alot!!

PR #4915, opened 2021-04-11
unbound dnsbl: yet another change in regex

kulikov-a:
Hi!
started testing unbound with a new regex-check and it looks like we forgot to escape the string for a complete match - allows entries with url on first match. 

-match the entire string to prevent garbage after valid fqdn
-add to skip stat if not match


AdSchellevis:

kulikov-a:
thanks!

PR #4918, opened 2021-04-14
VLANs: add warning not to use tagged + untagged on the same interface

maurice-w:
As recently mentioned by @fichtner [1], you should avoid using a parent interface if it has VLANs. Many users are probably unaware of this limitation (I certainly was), especially since having tagged and untagged frames on the same interface is a very common configuration on other devices.

[1] https://forum.opnsense.org/index.php?topic=22571.msg107332#msg107332

### DIFF
--- a/src/www/interfaces_vlan.php
+++ b/src/www/interfaces_vlan.php
@@ -149,6 +149,9 @@ $main_buttons = array(
                   <tfoot>
                     <tr>
                       <td colspan="5">
+                        <strong><?= gettext("When using VLANs, avoid using their parent interfaces! Firewall rules added to a parent interface may interfere with VLANs and result in unpredictable behaviour.");?></strong>
+                        <br />
+                        <br />
                         <?= gettext("Not all drivers/NICs support 802.1Q VLAN tagging properly. On cards that do not explicitly support it, VLAN tagging will still work, but the reduced MTU may cause problems.");?>
                       </td>
                     </tr>


AdSchellevis:
I'm unsure about this, if I'm not mistaken some providers over here in the Netherlands only tag the the voice or data channel and leave the rest untagged. By my knowledge this works, but we just don't recommend doing so to avoid risk (the parent interface can see all traffic, tagged ones only see the targeted part). 
Broken switches can lead to very odd behaviour and accidental data leakage, where the risk is smaller if one would only used tagged ports on the same physical link. 

I'll leave this open for now, don't mind adding a note that we don't advice combining, but if it technically works we might need to rephrase here.

fichtner:
I thought about this... deriving the complex scenario means a lot of code, but maybe we can push something to the interfaces assignment page and warn users about the choice they made/are making when the parent is already assigned instead of adding a note to a note in bold letters.

IMO this is first and foremost documentation material.

muchahagrande:
Just a comment. I'm using VLANS in combination with the parent interface without any issue sinse years ago.
My parent interface is rarely used, but I need it to access to ESXi management interface in some cases.
DHCP works fine on all interfaces where enabled, including the parent one.

maurice-w:
From how I understood @fichtner, the problem isn't that it "doesn't work" or that you might get in trouble with broken switches. The problem is that it _does_ work, but firewall rules added to the parent might also apply to the VLANs. Which has significant security implications. For example, if the parent is a WAN and you have a VLAN with an 'allow all' rule, your WAN might unintentionally be wide open. And you won't notice ("it works") unless you actually do pen testing.

Did I get that correctly?

Having a dynamic warning on the interface assignment page would be great of course, but I'm currently not up for implementing that. And about having it in the docs: Absolutely, but not exclusively. This is essential, security critical information which belongs in the UI.

fichtner:
It’s the other way around. Untagged interface rules may apply to all tagged ones since pf(4) can read and skip over the VLAN tags on the traffic. It depends on rule ordering and quick/non-quick sequence though.

maurice-w:
@fichtner You're right, had a brain fart, edited it right after posting. The issue remains: You might unknowingly create a configuration which apparently works, but is insecure.

(You might not believe it, but I actually do have a setup where the native (untagged) VLAN is a LAN while the WAN is on a tagged VLAN. Luckily I'm handling VLANs in Hyper-V, not in OPNsense itself. But I'm sure there's someone out there who does...)

muchahagrande:
Of course... I didn't catch the potential implications on pf filtering rules.
I've got WAN VLAN handled as separate WAN at the hypervisor, as @maurice-w does on Hyper-V.

martinseener:
We just bought a new DEC3850 for our office and we have ax0/ax1 (the 10G ifaces) as a LAGG towards a Unifi Aggregation Switch and on top of that several VLANs (tagged) for internal LAN, Guests-WiFi etc. and use that LAGG also untagged for Unifi network itself (for adoption, updates, provisioning). So I guess the same issue with pf applies to us as well.
From my point of view a warning would help but in my opinion this should be finally fixed in pf that it does not simply take rules from untagged iface to all vlan ifaces - it makes no sense and vlan ifaces have different names and thus should be handled independently. I see this clearly as a bug in pf.

Since I haven't heard of this odd behaviour before in my career, does anyone know if this also applies to Linux filtering (iptables, nftables?)

AdSchellevis:
@martinseener I think long story short is that it's not a best practice (https://github.com/opnsense/core/pull/4918#issuecomment-819265246) and although it does work in most cases, it can make your debugging experience pretty painful and in rare cases can lead to side affects (which may also be driver specific). 

It's would however be quite easy to test the general assumption though, just drop all traffic on the parent and allow all on the child. if the child works (what I expect), it's not a generic "issue", merely something we should advise people to avoid.

Issue #4248, opened 2020-08-09
Firewall / aliases - resolve mac addresses


AdSchellevis:
The question to be able to filter on mac addresses is asked from time to time, since `pf(4)` doesn't support this type of functionality the idea is to resolve mac addresses in a similar way as we would when using hostnames. Periodically query arp+ndp to find which addresses are currently attached to a mac address.

https://github.com/opnsense/core/commit/fc38dfaee48c05026a2425067b9fddab6c6798ef added the backend code, our model still needs minor modifications to support this.

Although this is still an experiment and needs documentation to be actually included in our product, it could help to set policies for physical addresses or ranges of addresses (like vendors).

PR #4919, opened 2021-04-14
Dynamic IPv6 prefix handling

marjohn56:
Initial commit of changes that will automatically handle fw aliases when the
 ipv6 prefix changes

### DIFF
--- a/src/etc/inc/filter.inc
+++ b/src/etc/inc/filter.inc
@@ -716,6 +716,16 @@ function filter_generate_aliases()
                 $tmp_ports = implode(" ", filter_core_get_port_alias($aliased['name'], array(), $aliasObject));
                 $aliases .= "{$aliased['name']} = \"{ {$tmp_ports} }\"\n";
                 break;
+            case "dynipv6host":
+                list ($ifcfgipv6) = interfaces_primary_address6($aliased['interface'], null, $ifconfig_details);
+                if (!is_ipaddrv6($ifcfgipv6)) {
+                 break;
+                }
+                $ifcfgipv6 = Net_IPv6::getNetmask($ifcfgipv6, 64);
+                $ifcfgipv6arr = explode(':', $ifcfgipv6);
+                $alias_address = make_ipv6_64_address($ifcfgipv6, $aliased['content']);            
+                $aliases .= "{$aliased['name']} = \"{ {$alias_address} }\"\n";
+                break;
             default:
                 $tblopt = (!empty($aliased['counters']) ? 'counters' : '') . " persist ";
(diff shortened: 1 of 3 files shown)


src/etc/inc/filter.inc
```diff
@@ -716,6 +716,16 @@ function filter_generate_aliases()
                 $tmp_ports = implode(" ", filter_core_get_port_alias($aliased['name'], array(), $aliasObject));
                 $aliases .= "{$aliased['name']} = \"{ {$tmp_ports} }\"\n";
                 break;
+            case "dynipv6host":
```
AdSchellevis, line 719: for addresses we try to prevent adding hooks in filter.inc, since all (non port) alias types are being handled by `update_tables.py` so there's no hard coupling between aliases and rules.

marjohn56:
Ah, OK but now you are in Python territory and I usually avoid snakes. :)
Sort of see in update_tables.py where I THINK it tries to get the alias_content = alias.resolve() at line 131, but to be honest I'm not familiar with python at all, so either I'll need some guidance or someone else will need to take it from there.

AdSchellevis:
maybe it's better to design the wanted feature first in a ticket, if the requirements are clear and straight to the point, it might not be a huge effort to implement it at some point. The mac address addition (https://github.com/opnsense/core/issues/4248) wasn't very complicated either and fits our strategy of making sure components are loosely coupled.

marjohn56:
It's pretty straight forward them ( for someone who works with Python ) we have created an alias with both a partial IPv6 address and an interface. We then merge the  interface prefix with the partial address, which is what happens with this PR but without using the Python routines. I am looking at alias.py at the moment, trying getting my head around it. I can see that with mac there's a arpache.py. In the final instance I would like to be able to be flexible with the suffix, in the context that if the user must enter the full suffix, so for a /64 suffix they would enter ::0:0:0:1000 , however, if they entered ::0:0:0:0:1000 that would be an 80 bit suffix and we would take the first /48 from the prefix and merge on that. However, that 's a wish, and in the initial case we would just want to merge on the /64.

AdSchellevis:
My advise would be to start with an issue, keeping it brief and to the point (I haven't read all of the earlier discussions, to help out I like to prevent reading to a lot of comments in different PR's and issues). Just assume I don't know anything ;) 

Important points are, what kind of information do you expect the user to input, what should it calculate to (and how) and when does it change.

marjohn56:
OK, I'll do that. It will be very interesting to see what you do to create this, should help me understand the Python flow.

marjohn56:
Is done

PR #4941, opened 2021-04-24
Dynamic Ipv6 fw rules handling Mk II

marjohn56:
IPv6 Prefix changes by the ISP  cause any fw rules created for specific clients on the LAN interface(s) to become invalid.

### DIFF
--- a/src/opnsense/mvc/app/models/OPNsense/Firewall/Alias.xml
+++ b/src/opnsense/mvc/app/models/OPNsense/Firewall/Alias.xml
@@ -37,6 +37,7 @@
                         <networkgroup>Network group</networkgroup>
                         <mac>MAC address</mac>
                         <external>External (advanced)</external>
+                        <dynipv6host>IPv6 Dynamic Host</dynipv6host>                                            
                     </OptionValues>
                 </type>
                 <proto type="OptionField">
@@ -46,6 +47,8 @@
                         <IPv6>IPv6</IPv6>
                     </OptionValues>
                 </proto>
+                <dyninterface type="InterfaceField">                                     
+                </dyninterface>
                 <counters type="BooleanField">
                     <default>0</default>
                 </counters>
--- a/src/opnsense/mvc/app/models/OPNsense/Firewall/Alias.xml
+++ b/src/opnsense/mvc/app/models/OPNsense/Firewall/Alias.xml
@@ -37,8 +37,13 @@
                         <networkgroup>Network group</networkgroup>
                         <mac>MAC address</mac>
                         <external>External (advanced)</external>
-                        <dynipv6host>IPv6 Dynamic Host</dynipv6host>                                            
+                        <dynipv6host>IPv6 Dynamic Host</dynipv6host>
                     </OptionValues>
+                    <Constraints>
+                        <check001>
+                            <reference>interface.check001</reference>
+                        </check001>
+                    </Constraints>
                 </type>
                 <proto type="OptionField">
                     <Multiple>Y</Multiple>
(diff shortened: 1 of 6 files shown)


src/opnsense/mvc/app/models/OPNsense/Firewall/FieldTypes/AliasNameField.php
```diff
@@ -63,7 +63,7 @@ public function getValidators()
         $reservedwords = array(
             'all', 'allow-opts', 'altq', 'anchor', 'antispoof', 'any', 'bandwidth', 'binat', 'binat-anchor', 'bitmask',
             'block', 'block-policy', 'buckets', 'cbq', 'code', 'codelq', 'crop', 'debug', 'divert-reply', 'divert-to',
-            'drop', 'drop-ovl', 'dup-to', 'fail-policy', 'fairq', 'fastroute', 'file', 'fingerprints', 'flags',
+            'drop', 'drop-ovl', 'dup-to','dyninterface', 'fail-policy', 'fairq', 'fastroute', 'file', 'fingerprints', 'flags',
```
AdSchellevis, line 66: @marjohn56 these are reserved pf words, you can remove your type from here.

marjohn56: wasn't sure.... covering all bases. Will do and update.

marjohn56:
So a little more testing shows that if I change the prefix, the aliastable entries are not updated. Is the best way to do that to set self._is_changed to true in the def _fetch_dynipv6() function or to write a file that contains the old prefix and compare it to the new one and use the result of that comparison to set the is_changed flag?

AdSchellevis:
@marjohn56 I'm merging your PR locally to test and cleanup, are you planning to push more? The force-push clutters the view from my end, if you do want to keep working on it, better change this PR into draft (in which case I'll just wait).

marjohn56:
Afternoon Ad - Erm, well It's a bit of a learning curve for me, so sort of enjoying the exercise. :) However I'll play locally and not push anything else and leave you to resolve ( no pun intended ) the stuff that needs changing.  The last commit was a copy/paste error that was very basic and should not have made it into the commit in the first place.

AdSchellevis:
good afternoon Martin , no problem, I'll just go ahead then, we all learn :)

marjohn56:
OK, the tables are being created OK, just cannot get them to update. Missing something obvious I suspect, but this is really my fist sojourn into the pf backend and python.

AdSchellevis:
don't worry, we'll figure this out step by step

marjohn56:
OK, found it. The addition of  <ttl>{{ system.aliasesresolveinterval|default('300') }}</ttl>  to filter_tables.conf has sorted that, though 5 minutes after an interface change may not be  way to handle the alias in this case.

AdSchellevis:
a short ttl (30 seconds or so), likely shouldn't be an issue in this case, I'll take a look when I'm there.

marjohn56:
OK, I think that should be fine. I suspect it's mainly a reboot issue where a new prefix is obtained, though some users have said their ISP will suddenly change their prefix. All I have done is this:

<dyninterface>{{ physical_interface(alias.dyninterface)|default('LAN')}}</dyninterface>
      <ttl>{{ system.aliasesresolveinterval|default('300') }}</ttl>      

So I'll change it to 30 seconds and check it all works... thanks for the advice and assistance. I've learned a lot.

marjohn56:
Yes, that works nicely.  Aliastables are updating with the new addresses.

AdSchellevis:
@marjohn56 I think this https://github.com/opnsense/core/commit/56f4243584e833ffc26d967aba83712a9327176e is more or less what you're looking for (branch FR_4923)

marjohn56:
That's working fine. Nice one.

marjohn56:
@AdSchellevis  - Updated, all yours. Thanks again.

AdSchellevis:
@marjohn56 is this the same as my branch or did you change anything? if it's the same then I rather merge the branch so both our commits are referenced properly.

marjohn56:
It's the same. I took your amendments, applied to my system, tested and pushed them back to git on my PR, so they are the same.

AdSchellevis:
@marjohn56 there you go https://github.com/opnsense/core/pull/4959 :)

marjohn56:
Thanks Ad, hopefully that will satisfy a few.

AdSchellevis:
we should still add some documentation about this new type by the way, haven't found time for that yet

marjohn56:
A doc for the Wiki? I'll put something together.

AdSchellevis:
@marjohn56 thanks in advance!

marjohn56:
Is done.. waiting in docs.

fichtner:
Thanks to both of you for this work ❤️

PR #4950, opened 2021-04-27
Unbound: Add description for [reload] action

bradleeedmondson:
Adds description for unbound's configd [reload] action

This causes the reload option to show up as a predefined action in cron and Let's Encrypt UI pages.[1] This is desirable if you want to use your Let's Encrypt certificate to offer DNS-over-TLS via unbound on port 853, so that the LE plugin can reload the unbound config whenever it renews the relevant certificate

[1] https://forum.opnsense.org/index.php?topic=6177.msg25956#msg25956


fichtner:
Restart already has a description on 21.7 and reload has been removed -> closing this PR.

PR #4954, opened 2021-04-28
dhcpd4: support ignore-client-uids

comotion:
A lot of clients boot into two modes one of which doesn't send uids, which means they get different IP addresses when they boot. For instance, if they need a decrypt password at boot time to unlock the root drive. 

Another instance where the uids change is when the machines are vagrant containers, or if they dual boot windows.

The dhcpd server already provides the `ignore-client-uids` configuration for this case. This patch just exposes the function.

Fixes opnsense#3673


AdSchellevis:
@comotion looks good, thanks!

PR #4960, opened 2021-04-29
Menu: Fix order of interfaces in groups

vnxme:
**Issue:**
I have recently noticed that sometimes interfaces in groups are ordered neither by their descriptions nor by their device names (e.g. optXX) making this part of the menu look messy. However, all the other interface lists in the menu (Firewall Rules, DHCPv4/6, Router Advertisements) seem to be properly ordered by the corresponding descriptions.

**Investigation:**
There is a place in the code where all interface lists are ordered before inserting in the menu and it looks ok:
https://github.com/opnsense/core/blob/89a2a8d51bf20f765adf8275de3c4ff66ac04c8c/src/opnsense/mvc/app/models/OPNsense/Base/Menu/MenuSystem.php#L246-L248

Next only groups and non-grouped interfaces are inserted in the menu. It also looks ok:
https://github.com/opnsense/core/blob/89a2a8d51bf20f765adf8275de3c4ff66ac04c8c/src/opnsense/mvc/app/models/OPNsense/Base/Menu/MenuSystem.php#L252-L267

Finally, the remaining (grouped) interfaces are inserted in the menu. The problem here is that the interfaces are inserted in the order of the keys in $ifgroups making the results of natcasesort over $iftargets useless:
https://github.com/opnsense/core/blob/89a2a8d51bf20f765adf8275de3c4ff66ac04c8c/src/opnsense/mvc/app/models/OPNsense/Base/Menu/MenuSystem.php#L269-L288

**Solution:**
The solution this PR proposes is to run the foreach cycle over the $iftargets['if'] instead or $ifgroups to keep the order ensured by natcasesort. ~~I know it looks like there could be only one foreach cycle over $iftargets['if'] instead of two, but it won't work because the underlying logic of the class provides for groups being inserted before nested interfaces.~~
Another solution is to merge two foreach loops into one and use an extra array to trace group items.

**Screenshots:**
Before:
image
After:
image

### DIFF
--- a/src/opnsense/mvc/app/models/OPNsense/Base/Menu/MenuSystem.php
+++ b/src/opnsense/mvc/app/models/OPNsense/Base/Menu/MenuSystem.php
@@ -266,23 +266,23 @@ class MenuSystem
             }
         }
 
-        foreach ($ifgroups as $key => $groupings) {
-            $first = true;
-            foreach ($groupings as $grouping) {
-                if (empty($iftargets['if'][$key])) {
-                    // referential integrity between ifgroups and interfaces isn't assured, skip when interface doesn't exist
-                    continue;
-                }
-                $this->appendItem('Interfaces.' . $grouping, $key, array(
-                    'url' => '/interfaces.php?if=' . $key . '&group=' . $grouping,
-                    'visiblename' => '[' . $iftargets['if'][$key] . ']',
-                ));
-                if ($first) {
-                    $this->appendItem('Interfaces.' . $grouping . '.' . $key, 'Origin', array(
-                        'url' => '/interfaces.php?if=' . $key,
-                        'visibility' => 'hidden',
+        foreach ($iftargets['if'] as $key => $descr) {
+            if (array_key_exists($key, $ifgroups)) {
+                $first = true;
+                foreach ($ifgroups[$key] as $grouping) {
+                    $this->appendItem('Interfaces.' . $grouping, $key, array(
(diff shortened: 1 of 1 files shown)


fichtner:
Hi @vnxme,

Why not sort this one afterwards with an appropriate function?


Cheers,
Franco

fichtner:
Hi Vladimir,

I was thinking about sorting the resulting menu structure since we create an array from recurring objects from somewhere else.


Cheers,
Franco

vnxme:
@fichtner I believe the menu is sorted somewhere inside the class by the 'order' numbers given to appendItem. So here we need to enumerate interfaces in the way they could get proper order numbers. And we effectively do that by natcasesort.

fichtner:
Ok, thanks, let me think about it with a clear head. :)

vnxme:
@fichtner Reworked the solution, now only one foreach loop over $iftargets['if'] is required.

AdSchellevis:
@vnxme how about https://github.com/opnsense/core/commit/d485b40096a50ab03d72651e9042826002ae01ea ?

AdSchellevis:
@vnxme I don't think it's a lot faster to be honest, you can put a profiler on it to test it, but since the cartesian product about the same complexity it's likely costing roughly the same amount of time to process. Since this isn't a real bottleneck at the moment, I prefer a simple solution which is easy to add to a release.

vnxme:
@AdSchellevis Got it, thanks! Hope, it will be released soon.

PR #4964, opened 2021-04-30
Edit Alias fix

marjohn56:
When the Edit panel is opened to edit an existing Dynamic IPv6 Alias, the proto type is visible. If you select another type of Alias and then return to the Dynamic IPv6 type the proto has gone. This PR just adds a hide proto selectpicker.


AdSchellevis:
@marjohn56 thanks!

PR #4976, opened 2021-05-05
lighttpd: set HSTS max-age to 1 year

maurice-w:
An HSTS cache validity period of less than 1 year is no longer considered to be sufficiently secure and might fail various audits.

References:
https://internet.nl/article/new-release-with-improved-tests-for-TLS-and-CSP/
https://www.netsparker.com/web-vulnerability-scanner/vulnerabilities/http-strict-transport-security-hsts-max-age-value-too-low/
https://hstspreload.org/


AdSchellevis:
@maurice-w sounds reasonable, thanks!

PR #4992, opened 2021-05-15
shell: Clarify revert to HTTP

rawtaz:
Before this change, it wasn't clear what one would revert from, as there was no context being given.

### DIFF
--- a/src/opnsense/scripts/shell/setaddr.php
+++ b/src/opnsense/scripts/shell/setaddr.php
@@ -514,7 +514,7 @@ console_configure_dhcpd(4);
 console_configure_dhcpd(6);
 
 if ($config['system']['webgui']['protocol'] == 'https') {
-    if (console_prompt_for_yn('Do you want to revert to HTTP as the web GUI protocol?', 'n')) {
+    if (console_prompt_for_yn('Do you want to change the web GUI protocol from HTTPS to HTTP?', 'n')) {
         $config['system']['webgui']['protocol'] = 'http';
         $restart_webgui = true;
     } elseif (console_prompt_for_yn('Do you want to generate a new self-signed web GUI certificate?', 'n')) {


fichtner:
To be honest, it sounds the same to me and originated from m0n0wall. Such things are better explained in the documentation.

rawtaz:
I'm not sure how it can "sound the same" - the old text mentions X and the new one X and Y. They cannot sound the same :)

I'm guessing that as a developer knowing the ins and outs of this software, it can be very hard to see where small changes like this one are coming from and why they're needed. But from regular users' perspectives it's different. We see and interpret things in the user interface quite a bit differently than those having used this software for a long time.

Yesterday I installed a new OPNsense. After the initial install I had to reconfigure the interface addresses to make it reachable over the network (I couldn't access it by 192.168.1.1), so in the console I went to the option to reconfigure the IP addresses.

After configuring the addresses, I started getting questions about the web GUI and other things, which in itself is pretty unexpected. All I had asked to do was change the IP addresses - I had never asked to reconfigure anything related to the web GUI or similar. So that's the first thing to ponder, but whatever.

Then comes the question "Do you want to revert to HTTP as the web GUI protocol?". At this point in time I have never configured the web GUI at all on this firewall, so I have no idea what it is currently set to. Hence I cannot possibly know what I am reverting *from*, and even though I could go through a thought process an *assume* that it means "from HTTPS", that's far from certain. There's an unknown chance that "reverting to HTTP" includes other changes as well. I simply do not know the context, because there never was a context given here.

That's why this small change is important. In short, it simply removes ambiguity, and that's a very important thing in a user interface. Please let's just merge this - I wouldn't have taken the time to open a PR if this wasn't correcting a flaw in the text.

FWIW, I'm very much aware of an into not complicating user interfaces or text unnecessarily, I work in other projects where there's a ton of user facing change requests, and I'm one of those that push quite hard on not making things too verbose or having too many options unless they're really needed. But as developers we also have to keep in mind that our users do see things that we don't see, at least they interpret them differently in many cases, and most often the way new users interpret a user interface says more about how good the user interface is, than how we interpret it does.

AdSchellevis:
I agree with Franco here, these things are better explained in the docs (https://github.com/opnsense/docs), in an attempt to provide more context, it will never be complete and for most users this change doesn't really clarify more than the original line. Keep in mind that every change proposed costs time from many others (managing the project, documentation, translations). 

It's often good to consider the amount of effort a seemingly simple change costs in reality, for these type of changes we really do seek more justification.

Maybe you do want to write a chapter about the console options for the docs, I don't think that's already there.

rawtaz:
It's two additional words. How can this possibly be any harm, or any more maintenance burden? I'm very well aware of maintenance burdens. This is not one of them.

Do you realize that you are effectively suggesting that people go and read the documentation, to understand what this simple text means? I'm at a loss of words.

Again, how can this small change possibly be a problem for anyone, including you as developers.

AdSchellevis:
let's agree to disagree, time spend in elaborate discussions about changing this sentence (which has been in predecessors of the product for over 15 years), would likely be better spend on improving the documentation.   

My main concern is that this change doesn't really clarify more than it already does, quoting your own comment:

> There's an unknown chance that "reverting to HTTP" includes other changes as well. I simply do not know the context, because there never was a context given here.

If that's the problem, adding a "from https" doesn't really solve these concerns as well.... if context is unclear, it's always a good idea to read the manual.... hence my previous question about extending the docs. 

Consider this my last comment on the discussion by the way, I thought I could be of help providing some direction, which doesn't seem to be the case.

rawtaz:
>> There's an unknown chance that "reverting to HTTP" includes other changes as well. I simply do not know the context, because there never was a context given here.

> If that's the problem, adding a "from https" doesn't really solve these concerns as well.... if context is unclear, it's always a good idea to read the manual.... hence my previous question about extending the docs.

This is simply not an accurate assessment. The old text doesn't say anything about what is being reverted *from*, which means that *something* is being changed to HTTP, but you don't know if it's HTTP and/or something else as well. The new text explains exactly what is being reverted *from*, which means that you know HTTPS is what is being changed to HTTP.

Can you please not just realize that even if you don't agree that this is *needed*, it will definitely not cause any more maintenance burden, not tae up any relevant additional screen estate, and will not cause any harm in any way, and accept it?

Just because things has been the way they are for 15 years doesn't have to mean that they were right. If that was the case, we wouldn't see changes being made at all. So it's a moot point.

I completely agree that we shouldn't be spending time on discussing this.

mimugmail:
IMHO this question should be removed from the dialog, who wants to run this on http in a LAN. Even offering http access nowadays is really risky :)

FingerlessGlov3s:
I get what @rawtaz is getting at. I think the confusion he's speaking of its the word "revert". Since a fresh install is default HTTPS technically there's nothing to revert.

So instead of proposed change "Do you want to revert from HTTPS to HTTP for the web GUI protocol?" maybe it should read something like "Do you wish to use HTTP instead of HTTPS for the web GUI protocol?" this to me reads more clearly.

revert: To go back to a former condition

rawtaz:
Someone with better communication skills put it into words way better than I did. Thank you!

Besides @FingerlessGlov3s suggestion, here's another slightly shorter one in case it's more in line with core developers' intentions: *Do you want to change the web GUI protocol from HTTPS to HTTP?*

I'd love to find a compromise where the text is improved *and* keeps core developers happy :+1:

fichtner:
Yes that last suggestion sounds reasonable. If you change the PR I will merge it.


Cheers,
Franco

rawtaz:
Awesome, I updated+rebased the PR.

fichtner:
Merged, thanks all!

PR #4998, opened 2021-05-19
webgui: make restart action usable in cron jobs

### LINKED ISSUE #4956: Add "webgui restart" to the cron UI by default
**Important notices**

- [ ] I have read the contributing guide lines at https://github.com/opnsense/core/blob/master/CONTRIBUTING.md
- [ ] I have searched the existing issues and I am convinced that mine is new.

I have a primary/backup CARP configuration and I'm using letsencrypt certificates which are pulled down by the primary; this works great, but it does not automatically restart the web GUI on the secondary so when the cert expires I can't access the gui without jumping through annoying hoops.

The "easy way" I can see to fix this is just to make a cron job that restarts the web gui every night -- it's not that hard to do, but it's a bit frustrating that there isn't an easy way to do it by default. All that would be required would be to make the "webgui restart" action visible in the UI by default. (I understand it should be possible to change that pretty easily, though I'm struggling so far. Either way, it should be an easy option from install)

Well, clearly the ideal would be to be able to somehow trigger a restart on the secondary when the certificate is updated, but I'm going to guess that's harder and it's not a hugely common problem, so I'm suggesting the "5 minute fix" instead =]

fraenki:
fixes #4956

### DIFF
--- a/src/opnsense/service/conf/actions.d/actions_webgui.conf
+++ b/src/opnsense/service/conf/actions.d/actions_webgui.conf
@@ -3,3 +3,4 @@ command:/usr/local/etc/rc.restart_webgui
 parameters:%s
 type:script
 message:Restarting web GUI
+description:Restart web GUI service


taxilian:
thanks =]

Issue #2195, opened 2018-02-12
firewall: live log column selector


fichtner:
Especially: flags

Mentioned here: https://forum.opnsense.org/index.php?topic=7277.msg32737#msg32737

tpcr:
I also need to see the flags. Can't tell if a TCP packet in the firewall log is legit or not.
Please provide patch as soon as available.
Even if the flag was added to the info window, that would be a help.

fichtner:
https://github.com/opnsense/core/commit/dddfef4aea4e2ffd41dadf41c7b460126fafaa87 should do the trick for TCP flags in the info window.

fichtner:
I'm unable to filter for anything but SYN packets so I'm closing this now. Filter capabilities can be extended via #4065 still if it makes sense. For now flags are not included for said quirk.

PR #5003, opened 2021-05-20
Add TCP flags to firewall log

maloo1:
Hi, 
Not sure if you would consider this small change, but I find it very useful to have the TCP flags shown in the firewall log...
Thanks!


AdSchellevis:
it would probably be better if one could hide/show columns, preferences differ between people, adding all options will make the screen quite crowded...

fichtner:
There used to be a ticket for this. We should update the contribution template to search for closed tickets as well...

fichtner:
Specifically, please consider this comment https://github.com/opnsense/core/issues/2195#issuecomment-634060680

maloo1:
Yeah that would definitely be a better solution to be able to select which columns to show!! I can have a crack at doing this but my php is pretty average, but if your willing to provide feedback/guide me through the existing opnsense structure ill give it a shot!

fichtner:
@maloo1 sure, we are happy to review. Most of it us UI/JS fiddling in this case so PHP won't get in the way ;)

fichtner:
Closing this waiting for new PR.

PR #5010, opened 2021-05-25
Allow more characters in the "Distinguished name - Organisation" field

jdeluyck:
The "Distinguished name - organisation" field is quite limited in what is allowed in it. Even just having a comma or fullstop was prohibited.

Searching through the docs of openssl I could not find a reason to block those, so here's a PR to allow those characters specifically in the organisation fields.


AdSchellevis:
@jdeluyck thanks! I've cleaned it up a bit in https://github.com/opnsense/core/commit/de09fb4f36051f7d6868b4ecaceb828c1059ef4f

vpx23:
Is it possible to allow the ampersand character '&' so we can use the German type of organization "GmbH & Co. KG"?

PR #5020, opened 2021-05-28
Add route when delegating a prefix

nickg-ca:
I noted in a [thread on the forums](https://forum.opnsense.org/index.php?topic=7719.0) that routes are missing when delegating prefixes. I came across the same behaviour in my use case and had a little time this week to look into it and try to fix it "properly". (Before I had a hack to fix the existing prefix script /usr/local/opnsense/scripts/dhcp/prefixes.php)

This solution adds commit/release/expiry hooks to dhcpv6.conf to add/delete the route. Unfortunately the commit hook is a bit of a hack through a python script, because there's no way to access the link local address of the client requesting a delegated prefix in ISC DHCP as of the latest version.

I think it makes sense to just delete the existing script in dhcp/prefixes.php since it's no longer needed, so I've done that and also removed any references to it. I'm pretty sure I got them all but not 100% certain.

I think the better solution would be to patch ISC DHCP so that the route command with the [requesting client address](https://github.com/mpalmer/isc-dhcp) can be issued directly from the commit hook. (Or to move to Kea which I think might support this already?) While doing all of this within the commit hook would be a better solution, it requires modifying ISC DHCP and it's a bit more work so I haven't pursued that. I'm not sure I like the solution I've come up with, but it does solve the problem of creating routes and so I figured I'd open a pull request so that it can be considered.


AdSchellevis:
looking at `prefixes.php` it does sound like a good idea to retire it, the python code can use some cleanups, but in general I'm not against this. We could use some user feedback, maybe @maurice-w and/or @marjohn56 do want to test this PR.

Eventually it's probably also the question if `dhcpleases[6]` should stay,  at a first glance I don't think it's doing a lot when not triggering `prefixes.php`

fichtner:
+1 on retiring dhcpleases[6] for sure

fichtner:
PS: looks like a very tidy solution for a batch of old glue, nice work @nickg-ca :)

nickg-ca:
I've been testing this a bit more and it needs to be reworked a little bit because the mac address isn't being populated correctly. I'll work on it a bit more and send it back in once I have it working more reliably.

nickg-ca:
I've looked at this a bit more over the past couple days and I realized this patch was more broken than I thought (it actually wasn't working at all and what I thought I was testing was actually based on another hacked up patch I made ages ago). I think the **right** way to do this is to migrate to Kea. I don't have time to do something like that so I'm going to just leave this for now, and hopefully in the future someone can work on it.

PR #5023, opened 2021-05-30
Traffic: Fix long comment preventing ipfw reload

ypid:
It seems that comments are limited in ipfw to 80 chars which is not documented (at least not in the manpage `ipfw(8)`). If users accidentally use long traffic shaper rule descriptions this limit could be reached causing updates via the GUI apply button to be silently ignored which will cause confusion.

`ipfw /usr/local/etc/ipfw.rules` shows the following error:

    Line 77: comment too long (max 80 chars)

and exists with edit code 65.

Interestingly, a slice of up to `0:83` still works. As this is undocumented and the error says max 80 chars I go with something slightly below this max to account for the whitespace.


AdSchellevis:
wow, this limitation seems to be rather old https://github.com/freebsd/freebsd-src/commit/62ff38ae068c4d8a9c58ba952109f8d566733d9a , better to avoid it indeed :) Thanks!

Issue #4988, opened 2021-05-13
Firewall live view filters not filtering correctly subsequent to #4730


upended:
**Describe the bug**
In FW Live View, the filters have been configured to using the "is not" option.  The filtered records include entries which should not be included.

This has also been seen when using other combination of filters / variables. As shown with ticket #4730.

Example1:  port is not 53, will display entries contain traffic with port 53.

Example2: host is not 192.168.1.1, will display entries containing traffic with IP 192.168.1.1

**To Reproduce**
Go to Firewall: Log Files: Live View
select a port or ip you would like to filter out of results
select "is not" and enter value.

**Expected behavior**
For records matching the filtered out criteria to not be displayed.  

**Describe alternatives you considered**
The filter stops operating correctly once the negation has been used and effects the filter when selecting multiple criteria are used.

**Screenshots**
![112042607-f1ee7a00-8b04-11eb-9e15-99afe1644be7](https://user-images.githubusercontent.com/63566363/118174491-bd4fbe00-b3e3-11eb-9400-c9877920706d.JPG)

![image](https://user-images.githubusercontent.com/63566363/118174567-dbb5b980-b3e3-11eb-88b3-f5b4e105d10a.png)

**Relevant log files**
none

**Additional context**

**Environment**

OPNsense 21.1.3_3

upended:
Created new ticket, closing.

upended:
Closed wrong ticket.

fichtner:
To make this even more interesting, the new template filter code in 21.1.6 will not save src,dst (host) and srcport,dstport (port) correctly anyway.

fichtner:
@upended can you try this one top of 21.1.6?

    # opnsense-patch 47d6832ed


Cheers,
Franco

fichtner:
One more issue reported via https://forum.opnsense.org/index.php?topic=23309.0 is the add button click area is odd.

fichtner:
I think the last issue will be addressed by @kulikov-a soon enough. Closing. :)

upended:
This ticket should not have been closed.  Please allow for a testing period and for users to respond.  

**Still experiencing issues after patching.**

 # opnsense-patch 47d6832ed
Fetched 47d6832ed via https://github.com/opnsense/core
Hmm...  Looks like a unified diff to me...
The text leading up to this was:
--------------------------
|From 47d6832edcdb447aa5dd4175b92f730803dd065e Mon Sep 17 00:00:00 2001
[...]

PR #5034, opened 2021-06-05
fw_log.volt: prevent controls overlap

kulikov-a:
Hi!
addresses the https://forum.opnsense.org/index.php?topic=23309.msg110930#msg110930 issue (also mentioned on https://github.com/opnsense/core/issues/4988#issuecomment-850878852)
-templates and records controls are collected in one div (so template controls were not split)
-add more space for filters table
-the filter value field is slightly narrowed so that the filters table does not get out of the div when the screen width decreases


fichtner:
Merged, thanks!

kulikov-a:
thanks!

Issue #5008, opened 2021-05-24
IPv6 ULA/VIP addresses are not properly being compressed before comparing them to the interface address for radvd


jdeluyck:
**Important notices**

Before you add a new report, we ask you kindly to acknowledge the following:

- [x] I have read the contributing guide lines at https://github.com/opnsense/core/blob/master/CONTRIBUTING.md
- [x] I have searched the existing issues and I am convinced that mine is new.

**Describe the bug**

I'm configuring IPv6 in my house:
* ISP offers PD /60, not static
* LAN is static ipv4, track interface ipv6 on PD 0x4

For having some internal stability, I added a VIP ULA on top of the interface, fddf:xxxx:yyyy:34::1/64. 
Forgot to set "Allow manual adjustment of DHCPv6 and Router Advertisements". so re-edited the interface, hit save, hit apply, and suddenly only my VIP was being announced by radvd - no longer the GUA.

**To Reproduce**

1. Configure WAN with ipv6 (PD), save, apply
2. Configure LAN with ipv6, track interface, check "Allow manual adjustment of DHCPv6 and Router Advertisements"
3. Hit save
4. Hit apply

Check that radvd is advertising the primary range, all works, happy campers.

5. Add VIP on the LAN interface, hit save - VIP doesn't show up 
6. Restart radvd - both ranges are announced

7. Re-edit the interface, don't change anything, just hit save
8. Hit apply
9. GUA range disappears from radvd config, only VIP ULA range remains, breaking your network after a while.

**Expected behavior**

Both ranges should keep being announced by radvd. It's probably something that gets overwritten instead of appended.

**Screenshots**

radvd config after applying everything correcty:
```
# cat radvd.conf
# Automatically generated, do not edit
# Generated for DHCPv6 server lan
interface igb1_vlan134 {
        AdvSendAdvert on;
        MinRtrAdvInterval 200;
        MaxRtrAdvInterval 600;
        AdvLinkMTU 1500;
        AdvDefaultPreference medium;
        AdvManagedFlag on;
        AdvOtherConfigFlag on;
        prefix 2a02:xxxx:xxxx:xxxx::/64 {
                DeprecatePrefix on;
                AdvOnLink on;
                AdvAutonomous on;
        };
        prefix fddf:xxxx:xxxx:34::/64 {
                DeprecatePrefix on;
                AdvOnLink on;
                AdvAutonomous on;
        };
[...]

PR #5036, opened 2021-06-06
Properly and forcibly compress IPv6 addresses

jdeluyck:
Properly and forcibly compress ipv6 addresses that we get from the interface and the ones from the VIP.

Fixes issue #5008.
Is a variant of PR #4932

test log:
```
[06-Jun-2021 13:10:42 Etc/UTC] vip[subnet] == fda4:4fab:7fec:0::1
[06-Jun-2021 13:10:42 Etc/UTC] IPV6::compress(vip[subnet]) == fda4:4fab:7fec:0::1
[06-Jun-2021 13:10:42 Etc/UTC] Forced: IPV6::compress(vip[subnet]) == fda4:4fab:7fec::1

[06-Jun-2021 13:10:42 Etc/UTC] addrparts[0] == fda4:4fab:7fec::1
[06-Jun-2021 13:10:42 Etc/UTC] IPV6::compress(addrparts[0]) == fda4:4fab:7fec::1
[06-Jun-2021 13:10:42 Etc/UTC] Forced: IPV6::compress(addrparts[0]) == fda4:4fab:7fec::1
```

without the forced compression the 0 in one of the parts of the ipv6 address remained, causing the test to fail.
I don't know if it's optimal to do it on both ends?

### DIFF
--- a/src/etc/inc/interfaces.inc
+++ b/src/etc/inc/interfaces.inc
@@ -4631,7 +4631,7 @@ function interfaces_primary_address6($interface, $realif = null, $ifconfig_detai
         foreach (config_read_array('virtualip', 'vip') as $vip) {
             if (
                 $vip['interface'] == $interface && $vip['mode'] == 'ipalias' &&
-                Net_IPv6::compress($vip['subnet']) == $addrparts[0]
+                Net_IPv6::compress($vip['subnet'], true) == Net_IPv6::compress($addrparts[0], true)
             ) {
                 /* do not care about alias */
                 continue 2;


fichtner:
Master branch isn’t working for you? Excessive compress I think just hides more bugs.

jdeluyck:
I tested against master, it didn't do it for me, as in the comment it didn't compress my :0: in the vip. Only using the forced compression fixed it.

It feels like a bug in the ipv6 library, but well...

fichtner:
Ok, it could be fixed by uncompressed/compress chain on the VIP side. It feels a bit unfortunate that we don’t uncompress the input on storing a VIP or some other type of config-time normalisation, but it is what it is.

jdeluyck:
That should happen in the subroutines storing the data, I guess? Shouldn't be too hard to implement, but I wonder how many side effects it can have.

fichtner:
Not more than before, but the trouble is the VIP data is already there saved in some twilight form. We can either uncompress both or uncompress/compress the stored one and trust ifconfig. At least some type of preference for a side would be nice :)

jdeluyck:
I guess that decision lies more with you (code maintainers)  ;)

fichtner:
Was afraid of that ;) Can you try 735dbfa7f6e8 ... if we do extended IPv6 transformations we might as well avoid doing it on IPv4 addresses by checking for ":" character.

jdeluyck:
> Was afraid of that ;) Can you try [735dbfa](https://github.com/opnsense/core/commit/735dbfa7f6e8c4e1e143df4643148efc19594fde) ... if we do extended IPv6 transformations we might as well avoid doing it on IPv4 addresses by checking for ":" character.

Just got round to testing it, this works for me.
(small nitpick: there's a spelling error in the comment above it - Matchig should read Matching)

fichtner:
Thanks, fixed typo. Will add the change to next version 21.1.7.


Cheers,
Franco

PR #5042, opened 2021-06-13
Bugfix: Prevent PHP output buffer rewriting from corrupting downloaded packet capture files

arnehaak:
**Issue/how to reproduce:** When creating certain packet capture files via the OPNsense GUI and trying to download them, these files files get corrupted during the download. The corruption happens for packet capture files that contain unencrypted HTML file transfers, e.g., captures of HTTP (not HTTPS) website accesses. The packet capture files can be downloaded without corruption via SSH.

**Root cause:** The CSRF functionality of OPNsense's GUI modifies the PHP output buffer before sending it out to the client. It uses a very simply heuristic to decide whether or not the buffer should be modified (looking for "<html"). This heuristic gives a false positive result for any capture files that contain "<html", which is quite a common case.

**Proposed fix:** Added a new internal state to class LegacyCSRF, which allows to manually control if the output buffer rewriting is enabled. This is then manually disabled before the file download starts and re-stored afterwards.

### DIFF
--- a/src/www/csrf.inc
+++ b/src/www/csrf.inc
@@ -32,6 +32,8 @@ class LegacyCSRF
     private $security = null;
     private $session = null;
     private $is_html_output = false;
+    private $is_rewrite_enabled = true;
+
     public function __construct()
     {
         global $config;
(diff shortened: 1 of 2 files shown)


AdSchellevis:
@arnehaak nice catch, can you try https://github.com/opnsense/core/commit/0a2412f43247b8c36d15c5379cd342e28d946c84, logically this should also remove the handler and doesn't require changes on the csrf class itself.

arnehaak:
Thanks for pointing me at that. I can confirm: That commit also fixes the issue, and is less of a code change.

I thought about that solution too, but I wasn't sure if turning off the output buffer for the transfer of (potentially very large) capture files could have negative side effects, like a relevant decrease in performance...
But I am not a PHP developer, so I guess you can answer that better based on your experience :-)

AdSchellevis:
I don't expect issues, but if it does cause issues, we'll fix those within the context of the page (re-enable buffering without hooks for example). I'll close this PR, thanks for confirming this works as well.

PR #5071, opened 2021-06-25
Fix localization.

kekek2:


fichtner:
Ok, what does it fix?

kekek2:
I built OPNsense with PHP 7.4. On version 7.3. this fix is not needed.

fichtner:
I saw https://github.com/opnsense/core/commit/9e1582ced1689a24a7c934a1a5de3518f350ed6d but not sure what you are seeing that doesn’t work. Please be more precise.

Issue #4454, opened 2020-11-11
LDAP Server configuration - memberOf not returned


GeoCookie:
Hi there,

I'm trying to get the memberOf attribute from my OpenLDAP server but I can't get it to work. I know this is something that has been asked many times, but I have not found any answer.

On the LDAP Server configuration, I have checked "Read Properties" and "Synchronize groups"
![image](https://user-images.githubusercontent.com/43962716/98804439-7c67f400-2416-11eb-9dd8-915e5ccb7aa6.png)

When I test a user, OPNSense retrieves all the attributes except memberOf
![image](https://user-images.githubusercontent.com/43962716/98804478-8ab61000-2416-11eb-9fbd-5d6e5ed7b2f5.png)

If I do a ldapsearch with this user, I can get the memberOf attribute but I have to specifically ask for it
![image](https://user-images.githubusercontent.com/43962716/98804685-bc2edb80-2416-11eb-9e66-c07687f1b846.png)

If I don't add "memberOf" to the query, LDAP will not return it. From what I understand, this is a normal behavior since it is an operational attribute.

Do you know how I can tell OPNSense to request this attribute? Or maybe I'm missing something else?

Edit :
I use the memberOf attribute to filter which accounts OPNSense can retrieve
![image](https://user-images.githubusercontent.com/43962716/98807188-93a8e080-241a-11eb-846c-3c2c8fd973bf.png)

It is working. OPNSense only see members of the group, not all users
![image](https://user-images.githubusercontent.com/43962716/98807283-b5a26300-241a-11eb-87b6-da0a32f86f81.png)

djesionek:
Hi,

I do not see where and how this was resolved. 
Can any of you help me to understand since I have the exact same issue and adding extended query does not change what the tester shows below "This user is a member of these groups: " for me?

memberOf is still not requested by the query looking at my OpenLDAP logs and without sync of groups its not possible to use this feature properly as far as I can see.

PR #5082, opened 2021-07-06
Enable group sync for LDAP servers, that do not return memberOf

rdd2:
The OpenLDAP setup we use does not return functional attributes by default. I understand this behaviour is quite common. One has to query for memberOf separately. So to enable allocating users to groups, I have added an extra ldap_read() that adds memberOf attributes to the lastAuthProperties array. 

This is relevant to issue #4454.

### DIFF
--- a/src/opnsense/mvc/app/library/OPNsense/Auth/LDAP.php
+++ b/src/opnsense/mvc/app/library/OPNsense/Auth/LDAP.php
@@ -578,7 +578,7 @@ class LDAP extends Base implements IAuthConnector
         if ($ldap_is_connected) {
             $this->lastAuthProperties['dn'] = $user_dn;
             if ($this->ldapReadProperties) {
-                $sr = @ldap_read($this->ldapHandle, $user_dn, '(objectclass=*)');
+                $sr = @ldap_read($this->ldapHandle, $user_dn, '(objectclass=*)', array("*", "memberOf"));
                 $info = @ldap_get_entries($this->ldapHandle, $sr);
                 if ($info['count'] != 0) {
                     foreach ($info[0] as $ldap_key => $ldap_value) {
--- a/src/opnsense/mvc/app/library/OPNsense/Auth/LDAP.php
+++ b/src/opnsense/mvc/app/library/OPNsense/Auth/LDAP.php
@@ -578,7 +578,7 @@ class LDAP extends Base implements IAuthConnector
         if ($ldap_is_connected) {
             $this->lastAuthProperties['dn'] = $user_dn;
             if ($this->ldapReadProperties) {
-                $sr = @ldap_read($this->ldapHandle, $user_dn, '(objectclass=*)', array("*", "memberOf"));
+                $sr = @ldap_read($this->ldapHandle, $user_dn, '(objectclass=*)', ['*', 'memberOf']);
                 $info = @ldap_get_entries($this->ldapHandle, $sr);
                 if ($info['count'] != 0) {
                     foreach ($info[0] as $ldap_key => $ldap_value) {


AdSchellevis:
it's probably better to optionally set the attributes to return with an additional parameter, this duplicates most of the code.

rdd2:
I agree that that would be the superior way of doing it, but I don't think it would work. From the [ldapsearch manpage](http://manpages.ubuntu.com/manpages/impish/man1/ldapsearch.1.html):

> If ldapsearch finds one or more entries, the attributes specified by attrs are returned. If * is listed, all user attributes are returned. If + is listed, all operational attributes are returned. If no attrs are listed, all user attributes are returned. If only 1.1 is listed, no attributes will be returned. 

Note that it makes no mention of combining the two wildcards. And from what I could test with a few (Open)LDAP-Servers this is indeed the case. You can query ```+``` and get functional attributes or ```*``` (or empty attribute list) and get user attributes, but you cannot combine them to get all attributes. I'd be interested if someone could test with other LDAP servers how they behave.

It is of course possible to list all attributes by listing them all and tacking ```memberof``` onto the end, but that is not really a good way here. 

Applying this to the situation at hand we definitely need two queries for this type of server. So I considered artificially joining the two result arrays but that would have been a bit clumsy with the ```count``` key and all. 

So I opted to do it this way, to run a separate query in exactly this situation and process the second query exactly like the first one. If you want I could do a slightly more elegant version where I add both results to an array and loop through them with the processing block. I however prefer the existing commit for readability.

Thanks for getting into this so quickly.

Thomas

AdSchellevis:
Hi Thomas,

If I'm not mistaken, the only change in your code is the following line:

```
$sr = @ldap_read($this->ldapHandle, $user_dn, '(objectclass=*)');
$sr_memberOf = @ldap_read($this->ldapHandle, $user_dn, '(objectclass=*)', array("memberOf"));
```

Which seems to suggest that `['memberOf']` could be an optional list of parameters to return, in which case an additional parameter for only this option would be enough to reach your goal, wouldn't break existing setups and doesn't query twice when it doesn't have to.

Best regards,

Ad

rdd2:
Hi Ad, 

yes, this is the change. The thing is, that the second query will ONLY return the ```memberOf``` attribute and value(s) and not all the other attributes. The first query is an implicit wildcard query because it names no attributes. It will only return user attributes (and not ```memberOf``` because it is a operational attribute). AFAIK it is not possibe to combine the wildcards with other attributes.

I my first approach to this was the same but that does not work. 

Best regards

Thomas

P.S. I wrote "functional" instead of operational earlier. Just corrected that.

AdSchellevis:
Hi Stephan,  

Just read my previous comment again and keep in mind that there's no need to do both (fetch all or define what you need).

Best regards,

Ad

fichtner:
> AFAIK it is not possibe to combine the wildcards with other attributes.

To query e.g. hidden attributes createTimestamp and modifyTimestamp my query is [ '*', 'createTimestamp', 'modifyTimestamp' ] and I have to say it works. :)


Cheers,
Franco

rdd2:
Hi Franco, 

you are right. That does work. I just tried. That is of course a much better way of doing it. I'll be back with a new commit. 

Thank you. 

Thomas

fichtner:
wouldn't it be better to pull it unconditionally. what's the worst that could happen?

rdd2:
Since there are LDAP servers which do not support memberOf at all and I can't be sure how they react when queried explicitly, I would make it conditional on users actually activating group sync.

fichtner:
Then you have the same problem on a smaller user base. It just got a lot harder to prove your assumption.

rdd2:
> Then you have the same problem on a smaller user base. It just got a lot harder to prove your assumption.
Please explain. 

As I see it, there are 3 scenarios:
- LDAP server with no memberOf overlay/functionality -> will work for user auth but will not for group sync and may(!) throw an error when queried for memberOf. This can then be remedied by deselecting group sync.
- LDAP with memberOf which answers with memberOf on empty attribute list -> worked before and will work with this patch
- LDAP with memberOf which only returns user attributes on empty attribute list -> did not work before but works with this patch

fichtner:
There is an easy way to test how a server responds to non-existing attributes. My guess is that it will not complain at all.

rdd2:
> There is an easy way to test how a server responds to non-existing attributes. My guess is that it will not complain at all.

Yes, that is also my experience. I have tested with OpenLDAP and Samba4. Samba4 includes memberOf in its wildcard answer and also does not complain on non-existing attributes. I would bet this this also the way AD behaves. There are however quite a few LDAP implementations and I don't have access to them all. If you want I could just add the attributes to the standard query.

src/opnsense/mvc/app/library/OPNsense/Auth/LDAP.php
```diff
@@ -578,7 +578,7 @@ public function authenticate($username, $password)
         if ($ldap_is_connected) {
             $this->lastAuthProperties['dn'] = $user_dn;
             if ($this->ldapReadProperties) {
-                $sr = @ldap_read($this->ldapHandle, $user_dn, '(objectclass=*)');
+                $sr = @ldap_read($this->ldapHandle, $user_dn, '(objectclass=*)', array("*", "memberOf"));
```
fichtner, line 581:
```suggestion
                $sr = @ldap_read($this->ldapHandle, $user_dn, '(objectclass=*)', ['*', 'memberOf']);
```

fichtner (approved):
neat and clean, very nice

rdd2:
Great thank you for all the feedback. This is my first PR in OPNsense. Yay!

fichtner:
Thanks for bearing with us, merged! ❤️

PR #5093, opened 2021-07-13
setaddr.php: allow rfc3021

kulikov-a:
Hi!
ref: https://forum.opnsense.org/index.php?topic=23878.0


fichtner:
Sorry, https://github.com/opnsense/core/commit/b1241aeb74ccf just went in after digging through code history for related reasons.

kulikov-a:
)) thanks!

PR #5100, opened 2021-07-16
Unbound: Update Miscellaneous.xml

mimugmail:
Allow cert verification in transition. 

I know we agreed to move to grid, but maybe for interim we just allow hash sign. We already dont to a validation for private domain field above (which is not an excuse) but at this point, maybe its better to allow this now while still have the option to migrate to grid view in later version


fichtner:
It’s better to migrate now. Already “@@@“ seems to be a valid entry and I have time to play with a grid pattern. 😊

mimugmail:
You now have a PR to start as a base :)

PR #5103, opened 2021-07-17
system: domainsearchlist configuration option

pfevre:
Here is a quick added features to implement teh domain search list for resolv.conf


fichtner (changes requested):
There is no validation here and that should be avoided. We can discuss after rebase on master.

fichtner:
I still don't see a validation of the input, a loose space-separated interpretation and a merge strategy that shuffles entries into the middle of the resulting entry? Do we need more than one single override now that the domain is always added in front when upstream DNS servers are used?

pfevre:
I'm still working on it.
I did update my firewall to last version and I still need to customize the domain search list.
I will implement some validation on the list.
Regards

src/www/system_general.php
```diff
@@ -295,6 +310,16 @@
                 </div>
               </td>
             </tr>
+            <tr>
+              <td><a id="help_for_domainsearchlist" href="#" class="showhelp"><i class="fa fa-info-circle"></i></a> <?=gettext("Domain search list"); ?></td>
+              <td>
+                <input name="domainsearchlist" type="text" value="<?=$pconfig['domainsearchlist'];?>" />
+                <div class="hidden" data-for="help_for_domainsearchlist">
+                  <?=gettext("Enter a custom ordered space separated list of search domain"); ?>
```
8191, line 318:
```suggestion
                  <?=gettext("Search list for host-name lookup. The search list is normally determined from the local domain name; by default, it contains only the local domain name. This may be changed by listing the desired domain search path. Up to 5 names can be separated using a semicolon character. Most resolver queries will be attempted using each component of the search path in turn until a match is found."); ?>
```

Taken from RESOLVER(5) man page.
Assuming the semicolon suggestion from above gets incorporated.

src/www/system_general.php
```diff
@@ -139,6 +153,7 @@
 
     if (count($input_errors) == 0) {
         $config['system']['domain'] = $pconfig['domain'];
+        $config['system']['domainsearchlist'] = $pconfig['domainsearchlist'];
```
8191, line 156: That would be the point to add some sanity checks (e.g. remove spaces).

src/www/system_general.php
```diff
@@ -83,6 +84,19 @@
         $input_errors[] = gettext("The domain may only contain the characters a-z, 0-9, '-' and '.'.");
     }
 
+    /* Domain search list Validation 256 char length max */
+    if (!empty($pconfig['domainsearchlist'])) {
+      if (strlen($pconfig['domainsearchlist']) > 256) {
+        $input_errors[] = gettext("Custom domain search list can not exceed 256 char length");
+      } else {
+        foreach (explode(' ',$pconfig['domainsearchlist']) as $searchdomain) {
+          if (!is_domain($searchdomain)) {
```
8191, line 93: The DHCPv4 services uses semicolon to separate domain names (see /services_dhcp.php). We should aim in consistency.

```suggestion
        foreach (explode(';', $pconfig['domainsearchlist']) as $searchdomain) {
          if (!is_domain(trim($searchdomain))) {
```

src/etc/inc/system.inc
```diff
@@ -186,6 +186,11 @@ function system_resolvconf_generate($verbose = false)
         $resolvconf .= "nameserver 127.0.0.1\n";
     }
 
+    if (isset($syscfg['domainsearchlist'])) {
+        /* add custom search list to search list */
+        $search = array_merge($search, explode(' ',$syscfg['domainsearchlist']));
```
8191, line 191: The DHCPv4 services uses semicolon to separate domain names (see /services_dhcp.php). We should aim in consistency.
```suggestion
        $search = array_merge($search, explode(';', $syscfg['domainsearchlist']));
```

src/www/system_general.php
```diff
@@ -83,6 +84,19 @@
         $input_errors[] = gettext("The domain may only contain the characters a-z, 0-9, '-' and '.'.");
     }
 
+    /* Domain search list Validation 256 char length max */
+    if (!empty($pconfig['domainsearchlist'])) {
+      if (strlen($pconfig['domainsearchlist']) > 256) {
+        $input_errors[] = gettext("Custom domain search list can not exceed 256 char length");
+      } else {
```
8191, line 91: The total length of the _search_ directive cannot exceed 256. As the domain name of the system is also added to the search path and this setting is even on the same page (`$pconfig['domain']`) I'd recommend to take this into account.

```suggestion
      if (strlen($pconfig['domainsearchlist']) + strlen($pconfig['domain']) > 256) {
        $input_errors[] = gettext("Custom domain search list including the domain name must not exceed 256 characters.");
      } else {
```

8191 (changes requested):
Please consider using semicolons as separators instead of spaces, as DHCPv4 for instance uses semicolons as well.

src/www/system_general.php
fichtner: This is tricky overall. We should really have a function in system.inc that validates this so the same check/approach is used in both places (or a way to compile the final list there to check for length here).

src/etc/inc/system.inc
fichtner: I would rather see a tokenizer to hide the complexity, but this one uses a "," to separate. It could be changed, but it doesn't make much sense either way if we aim to hide the presence of the separator more and let it work like all other tokenizers in the GUI?

fichtner (commented):
I'm not against this, but I think we need to talk merge strategies before we include it.

Shall we make it custom-custom so the domain is not appended? Do we want some automatic changes, also give the DHCP-handled search domains? A knob similar to nameservers comes to mind or we need to latch on to this setting actually?

"Allow DNS server list to be overridden by DHCP/PPP on WAN" -- disable search domains from DHCP?

"Do not use the local DNS service as a nameserver for this system" -- disable system domain in search list?

src/www/system_general.php
pfevre: I wasn't aware of this syntax :-/ I always used space separated list. but let stick to the man :-)

src/etc/inc/system.inc
8191: > I would rather see a tokenizer to hide the complexity

:+1: Definitely the best idea. How's it realized in the backend (space vs. comma vs. semicolon) is not important here, I'd say. I just find it confusing when one GUI element uses this and another GUI element that. With the tokenizer this difference is gone anyway.

8191:
> Shall we make it custom-custom so the domain is not appended? Do we want some automatic changes, also give the DHCP-handled search domains? A knob similar to nameservers comes to mind or we need to latch on to this setting actually?

I believe this would be confusing for the end-user. I cannot imagine a situation where the system shall have a domain name, which is actually not part of the search domain list.

> Allow DNS server list to be overridden by DHCP/PPP on WAN

I never understood this setting - what's its purpose? The WAN is normally an untrusted end, i.e. an ISP. Why would I - as an administrator of the trusted side - care about domain names of the uplink?

fichtner:
@pfevre @8191 981373b4cdd903d adds a single domain to the search list as a compromise and first step. I don't feel comfortable messing with the natural order (system domain first) or providing a full override so it's simply appended as the second entry. The field can be easily extended to hold more than one domain but at least for 22.1 we have something that people can use and see how it holds up in practice.


Cheers,
Franco

pfevre:
I decided to put the custom value at the beginning because it precede the system setting in the system resolver. 
For me it make sense to first try a resolve on custom domain and the fallback on system domain.
Regards

fichtner:
In practice you get the same result when assigning the custom one as default domain. What would be the reason not to do this in the first place?

PR #5108, opened 2021-07-20
openvpn: check ipv4 tunnel prefix

kulikov-a:
Hi!
openvpn 2.5 now requires the pool size to be at least 2 (https://github.com/OpenVPN/openvpn/pull/153)
maybe we can verify ipv4 tunnel prefix at server creation/editing?
ref: https://forum.opnsense.org/index.php?topic=23934.0

### DIFF
--- a/src/etc/inc/plugins.inc.d/openvpn/wizard.inc
+++ b/src/etc/inc/plugins.inc.d/openvpn/wizard.inc
@@ -440,6 +440,12 @@ function step10_submitphpaction()
 
     if ($result = openvpn_validate_cidr($_POST['tunnelnet'], gettext('IPv4 Tunnel Network'), false, 'ipv4')) {
         $input_errors[] = $result;
+    } elseif (isset($_POST['tunnelnet'])) {
+        // Check IPv4 tunnelnet pool size. Wizard makes tun mode with net30 server only.
+        list($ipv4tunnel_base, $ipv4tunnel_prefix) = explode('/',trim($_POST['tunnelnet']));
+        if ($ipv4tunnel_prefix > 28) {
+            $input_errors[] = gettext('A prefix longer than 28 cannot be used with a net30 topology.');
+        }
     }
 
     if ($result = openvpn_validate_cidr($_POST['tunnelnetv6'], gettext('IPv6 Tunnel Network'), false, 'ipv6')) {
--- a/src/www/vpn_openvpn_server.php
+++ b/src/www/vpn_openvpn_server.php
@@ -202,6 +202,21 @@ if ($_SERVER['REQUEST_METHOD'] === 'GET') {
 
         if ($result = openvpn_validate_cidr($pconfig['tunnel_network'], gettext('IPv4 Tunnel Network'), false, 'ipv4')) {
             $input_errors[] = $result;
+        } elseif (isset($pconfig['tunnel_network'])) {
+            // Check IPv4 tunnel_network pool size
+            list($ipv4tunnel_base, $ipv4tunnel_prefix) = explode('/',trim($pconfig['tunnel_network']));
+            if ($pconfig['dev_mode'] != "tap") {
+                if ($ipv4tunnel_prefix > 28 && empty($pconfig['topology_subnet'])) {
+                    $input_errors[] = gettext('A prefix longer than 28 cannot be used with a net30 topology.');
+                }
+                if ($ipv4tunnel_prefix > 29 && !empty($pconfig['topology_subnet'])) {
+                    $input_errors[] = gettext('A prefix longer than 29 cannot be used for tunnel network.');
+                }
+            } else {
+                if ($ipv4tunnel_prefix > 29) {
+                    $input_errors[] = gettext('A prefix longer than 29 cannot be used for tunnel network.');
+                }
+            }
         }
 
         if ($result = openvpn_validate_cidr($pconfig['tunnel_networkv6'], gettext('IPv6 Tunnel Network'), false, 'ipv6')) {


src/etc/inc/plugins.inc.d/openvpn/wizard.inc
```diff
@@ -440,6 +440,12 @@ function step10_submitphpaction()
 
     if ($result = openvpn_validate_cidr($_POST['tunnelnet'], gettext('IPv4 Tunnel Network'), false, 'ipv4')) {
         $input_errors[] = $result;
+    } elseif (isset($_POST['tunnelnet'])) {
```
AdSchellevis, line 443: `!empty($_POST['tunnelnet']) && strpos($_POST['tunnelnet'], "/") !== false` might be better.

src/www/vpn_openvpn_server.php
```diff
@@ -202,6 +202,21 @@
 
         if ($result = openvpn_validate_cidr($pconfig['tunnel_network'], gettext('IPv4 Tunnel Network'), false, 'ipv4')) {
             $input_errors[] = $result;
+        } elseif (isset($pconfig['tunnel_network'])) {
+            // Check IPv4 tunnel_network pool size
+            list($ipv4tunnel_base, $ipv4tunnel_prefix) = explode('/',trim($pconfig['tunnel_network']));
+            if ($pconfig['dev_mode'] != "tap") {
```
AdSchellevis, line 208: probably better to use `$pconfig['dev_mode'] == 'tun' `, future device modes (if they would exist) likely have other constraints.

src/www/vpn_openvpn_server.php
```diff
@@ -202,6 +202,21 @@
...
+                if ($ipv4tunnel_prefix > 28 && empty($pconfig['topology_subnet'])) {
+                    $input_errors[] = gettext('A prefix longer than 28 cannot be used with a net30 topology.');
+                }
+                if ($ipv4tunnel_prefix > 29 && !empty($pconfig['topology_subnet'])) {
+                    $input_errors[] = gettext('A prefix longer than 29 cannot be used for tunnel network.');
+                }
+            } else {
```
AdSchellevis, line 215: `elseif ($ipv4tunnel_prefix > 29) { ...` saves a line

src/www/vpn_openvpn_server.php
```diff
@@ -202,6 +202,21 @@
 
         if ($result = openvpn_validate_cidr($pconfig['tunnel_network'], gettext('IPv4 Tunnel Network'), false, 'ipv4')) {
             $input_errors[] = $result;
+        } elseif (isset($pconfig['tunnel_network'])) {
+            // Check IPv4 tunnel_network pool size
+            list($ipv4tunnel_base, $ipv4tunnel_prefix) = explode('/',trim($pconfig['tunnel_network']));
+            if ($pconfig['dev_mode'] != "tap") {
+                if ($ipv4tunnel_prefix > 28 && empty($pconfig['topology_subnet'])) {
+                    $input_errors[] = gettext('A prefix longer than 28 cannot be used with a net30 topology.');
+                }
```
AdSchellevis, line 211: `elseif ($ipv4tunnel_prefix > 29 && !empty($pconfig['topology_subnet'])) {` saves space, easier to read as well (it can't both be true.

src/www/vpn_openvpn_server.php
```diff
@@ -202,6 +202,21 @@
 
         if ($result = openvpn_validate_cidr($pconfig['tunnel_network'], gettext('IPv4 Tunnel Network'), false, 'ipv4')) {
             $input_errors[] = $result;
+        } elseif (isset($pconfig['tunnel_network'])) {
```
AdSchellevis, line 205: better to use `!empty()`, I think it's always set, check for existence of / makes it a bit more safe.

AdSchellevis (changes requested):
@kulikov-a made some remarks inline

src/etc/inc/plugins.inc.d/openvpn/wizard.inc
kulikov-a: i thought strpos($_POST['tunnelnet'], "/") is already checked by openvpn_validate_cidr?

src/www/vpn_openvpn_server.php
kulikov-a: agree

src/etc/inc/plugins.inc.d/openvpn/wizard.inc
AdSchellevis: you're right, didn't read into the function. But if tunnelnet is always there, a simple `else` would have been enough as well then I guess.

src/www/vpn_openvpn_server.php
kulikov-a: sorry. code works  but i'm obviously confused a little about isset() vs. empty() vs. is_null()

kulikov-a:
@AdSchellevis  thanks!! will make a "clean" PR

src/www/vpn_openvpn_server.php
AdSchellevis: I can understand that completely, `isset()` validates if an element exists (the value doesn't matter), `is_null()` (which I personally never use) is more or less the slower version of `$myvar === null` (evaluates to null and the type matches). `empty()` returns true if either the element doesn't exist (same as `!isset()`) or it evaluates to something empty (0, null, "", false)


PHP can be confusing...

kulikov-a: thanks! I'd better test and not guess from the man )

kulikov-a:
replaced by https://github.com/opnsense/core/pull/5114

Issue #5106, opened 2021-07-19
OpenVPN Wizard will create invalid firewall rule resulting in firewall not starting after reboot


sschueller:
**Important notices**

Before you add a new report, we ask you kindly to acknowledge the following:

- [X] I have read the contributing guide lines at https://github.com/opnsense/core/blob/master/CONTRIBUTING.md
- [X] I am convinced that my issue is new after having checked both open and closed issues at https://github.com/opnsense/core/issues?q=is%3Aissue

**Describe the bug**

When using the OpenVPN server wizard and selecting UDP4 from the drop down as well as auto generate inbound rules on WAN will result in the creation of a UDP4 inbound rule which is invalid. Rules are not applied and after restart the following error shows in the logs:

```bash
/usr/local/etc/rc.routing_configure: There were error(s) loading the rules: /tmp/rules.debug:118: unknown protocol udp4 - The line in question reads [118]: pass in quick on igb1 reply-to ( igb1 192.168.8.1 ) proto udp4 from {any} to {(igb1)} label "98a50c7974bb61dbcd84a3f2cb38f58b" # : OpenVPN wizard
```

**To Reproduce**

1. Reset configuration or fresh install
2. Skip setup wizard, go to OpenVPN server and click the wizard to add a server
3. Select local user access
3. Generate CA and certificate
4. Select UDP4 from drop down menu for Protocol, set WAN as interface and set a tunnel network
5. Check both "Traffic from clients to server" and "Traffic from clients to VPN" on the next page
6. Under Firewall/Rules/WAN there should now be a inbound rule with UDP4 set which is not possible when creating one directly or even editing this one
7. If the rule does not appear, delete the openvpn server, remove the other rule under OpenVPN in the firewall and re-run the openvpn server wizard. Select the existing ca and certificate. (for some reason the rule is not always created when I try this)

**Expected behavior**

Correct rule is created

**Describe alternatives you considered**

Editing the rule in the firewall fixes the issue

**Screenshots**

![CropperCapture 438](https://user-images.githubusercontent.com/1180107/126134255-8897a125-03fe-4def-aef8-1bd21235c57f.png)

**Environment**

Software version used and hardware type if relevant, e.g.:

- OPNsense 21.1.8_1-amd64
- FreeBSD 12.1-RELEASE-p19-HBSD
- OpenSSL 1.1.1k 25 Mar 2021
- AMD GX-412TC SOC (4 cores)
- pcengines APU2

Installed Plugins:
[...]

PR #5112, opened 2021-07-21
Openvpn: wizard rules creation fix

kulikov-a:
Hi!
may close https://github.com/opnsense/core/issues/5106 i think
i can add this to https://github.com/opnsense/core/pull/5108


AdSchellevis:
@kulikov-a I haven't tested it on my end, but it looks good. thanks! let's pull this in so others can try as well with opnsense-patch.

kulikov-a:
@AdSchellevis 
Thanks!!

Issue #5127, opened 2021-07-29
GeoIP broken after upgrade to 21.7


Alives:
**Important notices**

Before you add a new report, we ask you kindly to acknowledge the following:

- [x] I have read the contributing guide lines at https://github.com/opnsense/core/blob/master/CONTRIBUTING.md
- [x] I am convinced that my issue is new after having checked both open and closed issues at https://github.com/opnsense/core/issues?q=is%3Aissue

**Describe the bug**

I have a GeoIP floating rule blocking non-USA/non-Canada traffic.  I noticed that nothing is getting through it since upgrading to 21.7.  I removed the rule for now.

1 . Remove the existing GeoIP Alias
2. Updated my maxmind api key
3. Updated and set the maxmind URL to the "GeoLite2 ASN: CSV Format" with my api key in the URI `https://download.maxmind.com/app/geoip_download?edition_id=GeoLite2-Country-CSV&license_key=hunter2&suffix=zip`
4. Click `Apply`
5. Refresh and see:
  * `Last updated: 2021-07-27T00:09:16`
  * `Total number of ranges: 2422331`
5. Click `Aliases` tab
6. Click `+` (Add)
7. Name: `GeoIP_Allow`
8. Type: `GeoIP`
9. `IPv4` and `IPv6` checked
10. Click `America` pulldown
11. Click `Canada`
12. Click `United States`
13. Click `Save`
14. Click `Apply`
15. CPU fan spins up, after about 30 seconds I get an Alias alert: `Invalid argument. [GeoIP_Allow]`

Software version used and hardware type if relevant, e.g.:
OPNsense 21.7-amd64
Intel(R) Core(TM) i7-4500U CPU @ 1.80GHz (4 cores)

AdSchellevis:
For a moment I had the impression that this was new, but `net.pf.request_maxcount` doesn't seem to be changed between versions. The default maximum size for a table is 65535 entries, which is too little for your selection. If you set net.pf.request_maxcount to something high enough in the tunables you should be good.

For the time being I think we should add a reference in the error, not sure if this will be the final solution.

fichtner:
Adjusting maximum table entries is required anyway and automatically sets this sysctl as well. I don’t think there is need to point people to something that another part of the system already takes care of if properly configured in the first place?

AdSchellevis:
@fichtner well, not sure what you mean, but without bogonsv6 configured you're default is likely to end up as 65535

[...]

PR #5142, opened 2021-08-05
Aliases: show available room in table-entries limit

kulikov-a:
Hi!
in continuation of the https://github.com/opnsense/core/issues/5127#issuecomment-892399166 discussion
-show progress bar with overall count (including failed tables length) vs. `table-entries` limit
-send error msg to syslog on table (re)load error
-write stats file ((so api call executes faster)) to aliases db dir with alias name, alias size, alias extra size, pfctl errror message

### DIFF
--- a/src/opnsense/mvc/app/controllers/OPNsense/Firewall/Api/AliasController.php
+++ b/src/opnsense/mvc/app/controllers/OPNsense/Firewall/Api/AliasController.php
@@ -246,6 +246,16 @@ class AliasController extends ApiMutableModelControllerBase
     }
 
     /**
+     * get aliases load stats and table-entries limit
+     */
+    public function getRecordsAction()
+    {
+        $backend = new Backend();
+        $result = $backend->configdRun('filter diag records json');
+        return $result;
+    }
+
+    /**
      * variant on model.getNodes() returning actual field values
      * @param $parent_node BaseField node to reverse
      */
--- a/src/opnsense/mvc/app/views/OPNsense/Firewall/alias.volt
+++ b/src/opnsense/mvc/app/views/OPNsense/Firewall/alias.volt
@@ -99,6 +99,26 @@
         }
 
         /**
+         * show tables limits, counts and alerts
+         **/
+        function get_aliases_stat() {
+            ajaxGet("/api/firewall/alias/getRecords", {}, function(data){
+                overall = data.overall_count;
+                limit = data.table_entries;
+                perc_full = Math.round(100*overall/limit);
+                $('#room_left').attr('aria-valuenow', perc_full + '%').css("width", perc_full + "%");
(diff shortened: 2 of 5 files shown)


Actpohomoc:
Thank you for very good work! I will try to download your files to test and try your very interesting features.

kulikov-a:
@Actpohomoc thanks! 
easier to test with `opnsense-patch -a kulikov-a 72ad80b 6ba8244 76de566 9719a8d f106c8d` (need to restart configd service after patch so new action can work)  and revert with the same command )
and please note that this is a draft pr for discussion only.

fichtner:
Revert only works in reverse order if multiple patches are given! 😉

AdSchellevis:
@kulikov-a changed it a bit in https://github.com/opnsense/core/commit/9b31fad9ed04b7c04423aa007b0ba64ebc9221a0, what do you think? (made it a bit simpler, removed the additional error log and moved the progress bar on top)

kulikov-a:
@AdSchellevis almost complete rework ) 
I like it!
 just a couple of things:
if I understand correctly, now pftablecount.py does not take into account the size of the tables that are not in the /var/db/aliastables/? (bogons, external etc). 
I'm not sure about arithmetic: if the size of the table has decreased, but for some reason there was a loading error, is it correct to choose a larger value?
and small question on html part: 
maybe it is worth to describe both parts of the fraction in the tooltip?

AdSchellevis:
> @AdSchellevis almost complete rework )
> I like it!

Thanks!

> if I understand correctly, now pftablecount.py does not take into account the size of the tables that are not in the /var/db/aliastables/? (bogons, external etc).

Thanks for the fix, that was indeed the intention.

> I'm not sure about arithmetic: if the size of the table has decreased, but for some reason there was a loading error, is it correct to choose a larger value?

The idea is rather simple to be honest, let's say we try to load 2M records in a 1M reservation. If we weren’t able to do so, the actual reservation of the table equals 0 (loading error). So if we always count the intended amount (planned) when it's higher than the actual size, we can easily report the overallocation (used for more than 100% likely needs an increased reservation). 
There are some small downsides to this approach, but I do believe these are less problematic than the issue we're trying to fix, if `pfctl` optimises the table, the actual usage might be slightly lower, manual additions before apply clutter te view a bit as well.

> and small question on html part:
> maybe it is worth to describe both parts of the fraction in the tooltip?

Definitely, I was struggling a bit with the icon yesterday, figured you would take another look as well :) Thanks!

kulikov-a:
@AdSchellevis 

>  but I do believe these are less problematic than the issue we're trying to fix, if pfctl optimises the table, the actual usage might be slightly lower, manual additions before apply clutter te view a bit as well

totally agree that the exact number of planned records can not be obtained by lines counting only but its still a useful visual info)

I think then the request can be considered closed?
Thanks again to @AdSchellevis !

PR #5148, opened 2021-08-07
rulecache.py: skip empty metadata

kulikov-a:
Hi!
again strange metadata in the ptresearch ruleset
`metadata: autosign, id_0,;`
ref: https://github.com/ptresearch/AttackDetection/blob/9bb547bd8411e2398ea6d4972af979d1ee64504e/Telegram/telegram.rules#L2
so mdtag[2] leads to 'index out of range' error for `record['metadata'][parts[0]]`
can we add one more check please?
thanks!


AdSchellevis:
@kulikov-a thanks! empty lists error out indeed

kulikov-a:
@AdSchellevis Thanks!!)

PR #5149, opened 2021-08-08
Aliases progress bar: small suggestions

kulikov-a:
Hi
just small suggestions for https://github.com/opnsense/core/pull/5142#issuecomment-894704862
thanks!

### DIFF
--- a/src/opnsense/scripts/filter/pftablecount.py
+++ b/src/opnsense/scripts/filter/pftablecount.py
@@ -50,7 +50,8 @@ if __name__ == '__main__':
                 planned_size  = tmp.count('\n') + 1 if len(tmp) > 0 else 0
                 # if planned size doesn't fit the table, make sure we report intented size
                 # used size can be divert a bit if pfctl optimizes as well.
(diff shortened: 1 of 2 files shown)


src/opnsense/scripts/filter/pftablecount.py
```diff
@@ -50,7 +50,8 @@
                 planned_size  = tmp.count('\n') + 1 if len(tmp) > 0 else 0
                 # if planned size doesn't fit the table, make sure we report intented size
                 # used size can be divert a bit if pfctl optimizes as well.
-                result['used'] += max(planned_size, table_size)
+                table_size = max(planned_size, table_size)
```
AdSchellevis, line 53: @kulikov-a nice catch, forgot about this one in the rush last night. Thanks!

AdSchellevis:
@kulikov-a looks better indeed, thanks! maybe we should move the title/tooltip to the info widget, but for now this is good enough.

kulikov-a:
@AdSchellevis and users will probably suggest something more)
Thanks!!

PR #5152, opened 2021-08-08
unbound: add configurable dnsbl response

shonjir:
The default DNSBL response of a static NULL A record causes some poorly written sites and mobile apps to spin needlessly on connection failures to 0.0.0.0.  Further, the DNSBL does not block IPv6 requests which utilize different record types.

This PR adds the ability to configure different response types including NXDOMAIN, full IPv4/IPv6 null response and request logging.  It includes the same "log only" functionality as PR https://github.com/opnsense/core/pull/4715 but is more generalized.  The configuration is marked as 'advanced' and retains current OPNsense behavior by default.

Response types added by this PR are 'always_nxdomain', 'always_null', 'redirect', 'inform_redirect', and 'inform'.   Additional types may be easily added should the need arise, although few of the other types are useful in a DNSBL context.

### DIFF
--- a/src/opnsense/mvc/app/models/OPNsense/Unbound/Unbound.xml
+++ b/src/opnsense/mvc/app/models/OPNsense/Unbound/Unbound.xml
@@ -1,7 +1,7 @@
 <model>
     <mount>//OPNsense/unboundplus</mount>
     <description>Unbound configuration</description>
-    <version>1.0.0</version>
+    <version>1.0.1</version>
     <items>
         <service_enabled type="LegacyLinkField">
             <Source>unbound.enable</Source>
@@ -49,6 +49,30 @@
                     <yy>YoYo List</yy>
                 </OptionValues>
             </type>
+            <localzone type="OptionField">
+                <Required>Y</Required>
+                <default>__default__</default>
+                <OptionValues>
+                    <__default__>Return null (0.0.0.0) A record (default)</__default__>
+                    <always_nxdomain>Always return NXDOMAIN</always_nxdomain>
+                    <always_null>Always return null A/AAAA response (matches subdomains)</always_null>
+                    <redirect>Return NXDOMAIN or redirect record (matches subdomains)</redirect>
+                    <inform_redirect>Log and return NXDOMAIN or redirect record (matches subdomains)</inform_redirect>
+                    <inform>Log and resolve normally or return redirect record</inform>
+                </OptionValues>
+            </localzone>
+            <redirect_a type="NetworkField">
+                <Required>N</Required>
+                <netMaskAllowed>N</netMaskAllowed>
+                <version>ipv4</version>
+                <default></default>
+            </redirect_a>
+            <redirect_aaaa type="NetworkField">
+                <Required>N</Required>
+                <netMaskAllowed>N</netMaskAllowed>
+                <version>ipv6</version>
+                <default></default>
+            </redirect_aaaa>
             <lists type="CSVListField">
                 <Required>N</Required>
             </lists>
--- a/src/opnsense/scripts/unbound/blocklists.py
+++ b/src/opnsense/scripts/unbound/blocklists.py
@@ -141,12 +141,38 @@ if __name__ == '__main__':
                     'blocklist download %(uri)s (lines: %(lines)d exclude: %(skip)d block: %(blocklist)d)' % file_stats
                 )
 
+        # fetch configured DNS response type
+        localzone = None
+        redirect_a = None
+        redirect_aaaa = None
+        if cnf.has_section('response'):
+            if cnf.has_option('response', 'localzone'):
+                localzone = cnf.get('response', 'localzone').strip()
+            if cnf.has_option('response', 'redirect_a'):
+                redirect_a = cnf.get('response', 'redirect_a').strip()
+            if cnf.has_option('response', 'redirect_aaaa'):
+                redirect_aaaa = cnf.get('response', 'redirect_aaaa').strip()
+            # null localzone is used to indicate legacy behavior
+            if localzone == '__default__':
+                localzone = None
+
     # write out results
     with open("/usr/local/etc/unbound.opnsense.d/dnsbl.conf", 'w') as unbound_outf:
         if blocklist_items:
             unbound_outf.write('server:\n')
             for entry in blocklist_items:
-                unbound_outf.write("local-data: \"%s A 0.0.0.0\"\n" % entry)
+                if localzone:
+                    # advanced DNSBL custom zone response type
+                    unbound_outf.write("local-zone: \"%s\" %s\n" % (entry, localzone))
+                    # Optional redirect records
+                    if localzone.endswith('redirect') or localzone.startswith('inform'):
+                        if redirect_a:
+                           unbound_outf.write("local-data: \"%s A %s\"\n" % (entry, redirect_a))
+                        if redirect_aaaa:
+                           unbound_outf.write("local-data: \"%s AAAA %s\"\n" % (entry, redirect_aaaa))
+                else:
+                    # legacy OPNsense behavior
+                    unbound_outf.write("local-data: \"%s A 0.0.0.0\"\n" % entry)
 
     syslog.syslog(syslog.LOG_NOTICE, "blocklist download done in %0.2f seconds (%d records)" % (
         time.time() - startup_time, len(blocklist_items)
--- a/src/opnsense/service/templates/OPNsense/Unbound/core/blocklists.conf
+++ b/src/opnsense/service/templates/OPNsense/Unbound/core/blocklists.conf
@@ -52,6 +52,11 @@ custom_{{loop.index}}={{uri}}
 {%      endfor %}
 {%    endif %}
 
+[response]
+localzone={{ OPNsense.unboundplus.dnsbl.localzone }}
+redirect_a={{ OPNsense.unboundplus.dnsbl.redirect_a }}
+redirect_aaaa={{ OPNsense.unboundplus.dnsbl.redirect_aaaa }}
+
 [exclude]
 # exclude localhost entries
 default_pattern_1=.*localhost$
--- a/src/opnsense/mvc/app/controllers/OPNsense/Unbound/forms/dnsbl.xml
+++ b/src/opnsense/mvc/app/controllers/OPNsense/Unbound/forms/dnsbl.xml
@@ -12,6 +12,33 @@
         <help>Select which kind of DNSBL you want to use.</help>
     </field>
     <field>
+        <id>unbound.dnsbl.localzone</id>
+        <label>DNSBL Response Type</label>
+        <type>dropdown</type>
+        <advanced>true</advanced>
+        <help><![CDATA[Select the desired DNSBL response type for blocklist matches. See 'local-zone' in unbound.conf(5) for full details.
+          "Always" types always return the indicated response.
+          "Inform" types generate additional logging that can be used to monitor blocklist matches.
+          "Redirect" types will return NXDOMAIN or local data (if configured).
+          Types that match subdomains can improve DNSBL coverage but may require whitelisting some subdomains.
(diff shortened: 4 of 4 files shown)


kulikov-a:
@shonjir sorry, so (if I understood the code correctly) there is no way to keep the current behavior (static 0.0.0.0) and add logging (inform zone type) as https://github.com/opnsense/core/pull/4715 do?

shonjir:
Hi, @kulikov-a!

With the current code only a single record is generated for each zone, either a static A record (and an implicit transparent zone) or an explicit zone definition with no local data.

For the inform, inform_redirect, and redirect zone types if local data is specified after the zone type definition then that data is returned instead of the default behavior.  The always_nxdomain and always_null zone types do not support local data under the zone and will always return either NXDOMAIN or a null record.

It is possible to add an A 0.0.0.0 record after the zone type definition for some of these types.   This would return an A 0.0.0.0 record for A record requests, and (most likely) NOERROR NODATA for all other record requests.  I can't say for certain whether this would produce the desired DNSBL behavior under all circumstances.

The code could be updated to optionally return a null record for the types that support it but in my opinion this would result in suboptimal client behavior versus NXDOMAIN.  That same code could be extended to support client connection logging by changing the record to point to an arbitrary target address (or addresses).  However, this will most definitely lead to the same suboptimal client connection behavior that I used to experience when I was running pfBlocker-NG on pfSense.

 I honestly think that the suboptimal behavior isn't worth the effort but hey, open source is all about options, right?

kulikov-a:
@shonjir hi!)

> It is possible to add an A 0.0.0.0 record after the zone type definition for some of these types. This would return an A 0.0.0.0 record for A record requests, and (most likely) NOERROR NODATA for all other record requests. I can't say for certain whether this would produce the desired DNSBL behavior under all circumstances.

yes, I'm only talking about adding the "Return static 0.0.0.0 record (default) _and log the request_" option.
in such a way that the current behavior (implicit transparent type with local-data in it) will be preserved but logging will be added (by replacing the type with an explicit inform type and local-data in it). there were repeated questions about the ability to track blocked requests

> but hey, open source is all about options, right?

absolutely! I'm just afraid that the devs will not be happy to return to the discussion of zone types and dnsbl.conf file again and therefore I got into your request, sorry)

shonjir:
I can add the functionality into this PR.  I only created this one because your PR [#4715](https://github.com/opnsense/core/pull/4715/commits) needed to be refactored for current master and it was easier to create a new PR with the features I wanted.

The main difficulty is how best to arrange the UI elements so that the combinations are clearly indicated to the user and invalid combinations are avoided.   I think that the fact that the checkbox does nothing on some zone types should be clearly indicated in the UI.

shonjir:
I've updated the PR to include an option that will add static null A and AAAA records for the redirect and inform response types.  Testing is appreciated.

@kulikov-a, I think you'll get the behavior you're looking for if you select "Log and return local data or resolve normally" and tick the "Static Local Data Record" option.

Here's an example of the output.
```
local-zone: "privetadblock.ru" inform_redirect
local-data: "privetadblock.ru A 0.0.0.0"
local-data: "privetadblock.ru AAAA ::"
```

kulikov-a:
@shonjir I like it, thank you very much!)

> I think you'll get the behavior you're looking for

yep. `inform` type with `local-data` is what I was looking for (although adding an AAAA record makes the `dnsbl.conf` even larger and further increases the unbound load time. but the way to load the blocklist is a separate discussion imho)

thanks again!  remains to wait for the devs opinion )

src/opnsense/service/templates/OPNsense/Unbound/core/blocklists.conf
```diff
@@ -51,6 +51,12 @@ custom_{{loop.index}}={{uri}}
 {%      endfor %}
 {%    endif %}
 
+{%    if not helpers.empty('OPNsense.unboundplus.dnsbl.localzone')%}
+[dnsbl]
```
fichtner, line 55: Probably can have a better designation here: „dnsbl“ is too broad

src/opnsense/mvc/app/models/OPNsense/Unbound/Unbound.xml
```diff
@@ -48,6 +48,22 @@
...
+                    <always_nxdomain>Always return NXDOMAIN</always_nxdomain>
+                    <always_null>Always return null A/AAAA response (matches subdomains)</always_null>
+                    <redirect>Return NXDOMAIN or local data (matches subdomains)</redirect>
+                    <inform_redirect>Log and return NXDOMAIN or local data (matches subdomains)</inform_redirect>
+                    <inform>Log and resolve normally or return local data</inform>
+                </OptionValues>
+            </localzone>
+            <localdata type="BooleanField">
+                <default>0</default>
+                <Required>N</Required>
```
fichtner, line 65: Better set „Y“ here and bump model version for migration

src/opnsense/mvc/app/models/OPNsense/Unbound/Unbound.xml
```diff
@@ -48,6 +48,22 @@
                     <yy>YoYo List</yy>
                 </OptionValues>
             </type>
+            <localzone type="OptionField">
+                <Required>N</Required>
```
fichtner, line 52: „Y“ here as well

src/opnsense/service/templates/OPNsense/Unbound/core/blocklists.conf
```diff
@@ -51,6 +51,12 @@ custom_{{loop.index}}={{uri}}
 {%      endfor %}
 {%    endif %}
 
+{%    if not helpers.empty('OPNsense.unboundplus.dnsbl.localzone')%}
```
fichtner, line 54: When we require both fields this conditional can be removed.

src/opnsense/scripts/unbound/blocklists.py
```diff
@@ -141,12 +141,36 @@ def uri_reader(uri):
...
+        localdata = None
+        localzone = None
+        if cnf.has_section('dnsbl'):
+            if cnf.has_option('dnsbl', 'localdata'):
+                localdata = bool(cnf.getint('dnsbl', 'localdata'))
+            if cnf.has_option('dnsbl', 'localzone'):
+                localzone = cnf.get('dnsbl', 'localzone').strip()
+            # localdata not supported with always_* zone types
+            if localdata and localzone.startswith('always'):
+                localdata = 0
```
fichtner, line 154: This manipulates model data. Not great for maintenance or extensibility. Would be better to add constraints to model itself.

src/opnsense/scripts/unbound/blocklists.py
```diff
@@ -141,12 +141,36 @@ def uri_reader(uri):
                     'blocklist download %(uri)s (lines: %(lines)d exclude: %(skip)d block: %(blocklist)d)' % file_stats
                 )
 
+        # fetch configured DNS response type
+        localdata = None
```
fichtner, line 145: None cannot really happen… tri-state Boolean?

src/opnsense/service/templates/OPNsense/Unbound/core/blocklists.conf
shonjir: Makes sense.  How about '[response]' since this config governs how unbound responds to blocklist matches?

src/opnsense/mvc/app/models/OPNsense/Unbound/Unbound.xml
shonjir: Will be updated in next push

src/opnsense/mvc/app/models/OPNsense/Unbound/Unbound.xml
shonjir: Will be updated in next push

src/opnsense/service/templates/OPNsense/Unbound/core/blocklists.conf
shonjir: Removed in next push

src/opnsense/scripts/unbound/blocklists.py
shonjir: I admit that I was not really happy with this since it results in a UI checkbox that does nothing for some settings but didn't know how to address that in the UI.   I'm more than happy to move this into the model.  I'll look for a similar example in the code for inspiration but would be appreciative if you could point me to a best-practice example that you'd like me to follow.

In the UI I've currently set the localzone option as a dropdown.   Do you think it would be better to expose this as radio buttons since the list is short?

src/opnsense/scripts/unbound/blocklists.py
shonjir: Taking advantage of the fact that False and 0 do the same things.  I've made this explicitly 0 in the next update.

src/opnsense/scripts/unbound/blocklists.py
shonjir: I was looking into the model documentation and examples and I'm not finding any standard constraints that appear to support the required logic.  I need to either ensure that the localdata setting is unset (set to false) if the localzone setting is default, always_null or always_nxdomain, or alternatively the localdata setting should only be enabled if the localzone setting is redirect, inform_redirect or inform, and otherwise set to false.

The SetIfConstraint logic seems mainly intended to ensure a field is set if another field has a given value but there doesn't seem to be a way to specify what value the field should be set to.  I could reverse the logic and use the boolean to disable adding custom data, but I would rather not as I was trying to set this up to be extended with the ability to add user-specified responses later on.

At this point I suspect the logic in blocklist.py might be the easiest way to implement this since adding records is only supported for certain zone types.

@fichtner How would you suggest I proceed?

kulikov-a: @shonjir @fichtner may `DependConstraint` work if we change logic a bit: replace "DNSBL Null Response" checkbox with "Block subdomains" checkbox (or some). and then use this field to constrain the selection of Response (zones) Types?

shonjir: Whether we create two constrained lists or roll it into a single list the net effect is the same - we'd need to embed the boolean flag into the option values (think "inform:null" vs "inform:nxdomain") and add logic to blocklist.py to take action based on the embedded flag value.   The net effect is that we're really just passing the boolean value in a different way and moving the test to a slightly different place, but it does avoid passing invalid combinations to blocklist.py that need to be tested for there. 

It's pretty crufty in my opinion to embed flags in what are otherwise clean option values since you have to make some assumptions as to option format that you didn't need to before.  It would be cleaner if the necessary conditions and/or target values could be specified directly in the constraints.

kulikov-a: agree. since it still need to hide the localdata checkbox at certain localzone option values, maybe it makes sense to add some js-magic to the .volt to set localdata<->localzone dependencies? then the data will get into the model already "clean"

kulikov-a: and we can try to extend model with `AllowIfConstraint`. @fichtner are you talking about it?)

kulikov-a: something like https://github.com/kulikov-a/core/commit/0fb2108f1694bfa445a9a6b5ad127c81abf9b422 ?

shonjir:
I couldn't figure out how to apply the current model capabilities to the logic I had and extending the model is beyond what I want to tackle with this PR.   I have instead removed the code that @fichtner objected to and modified it to accept redirect values supplied by the model.  The blocklist.py script checks for local zone types of 'redirect', 'inform_redirect' or 'inform' since local data records are only valid with these types and if override values have been supplied will generate A and/or AAAA records using the supplied values.  The net effect is that one can now customize the redirect behavior by supplying IPv4 and/or IPv6 records to return instead of returning NXDOMAIN.

@fichtner I hope this addresses your concerns in a way that is acceptable.

AdSchellevis:
partially included in https://github.com/opnsense/core/pull/5736

PR #5160, opened 2021-08-11
Adding additional memory cache options for squid webproxy

Xeroxxx:
memory_cache_mode: 	Controls which objects to keep in the memory cache (cache_mem)

default = always. If always is set, value will not saved to config.

Maximum object size in memory: Objects greater than this size will not be attempted to kept in the memory cache. This should be set high enough to keep objects accessed frequently in memory to improve performance whilst low enough to keep larger objects from hoarding cache_mem.

default = 512KB. If value not set default applies.

Only visible with advanced config. Help text formated and adapted form squid documentation.


AdSchellevis:
@Xeroxxx looks good, thanks!

Issue #4819, opened 2021-03-14
Add "IPv4+IPv6" protocol value for IPSEC mobile VPN


goodomens42:
- [ x] I have read the contributing guide lines at https://github.com/opnsense/core/blob/master/CONTRIBUTING.md
- [ x] I have searched the existing issues and I am convinced that mine is new.

IPv6 is becoming more important day by day as more and more mobile users are using connections with DS-Lite or pure IPv6.
In order to setup a mobile tunnel that works with IPv4 and IPv6 in OPNsense you have to specify the interface as "any".
This approach might have undesired side effects in a configuration with several WAN interfaces that are used for IPSEC.
It would be nice to have a "clean" way to configure the mobile user connection for both IPv4 and IPv6.

I have created a patch, that allows setting the protocol to "any" when editing the mobile user connection.
The generated StrongSwan configuration will then contain something like

`left = 1.2.3.4,2001:1:2:3:4::1`

which works fine for us in production use.
I am aware this will only work for IKEv2, maybe there should be some kind of warning if using protocol=any with IKEv1.

[opnsense_ipsec.patch.txt](https://github.com/opnsense/core/files/6137192/opnsense_ipsec.patch.txt)

AdSchellevis:
@goodomens42 we don't accept patches inside issues, for discussion and feature merges we advise to use the normal GitHub PR process (https://docs.github.com/en/github/collaborating-with-issues-and-pull-requests/about-pull-requests)

jasperroloff:
@goodomens42 Based on your patch, I've created a PR: see #5166

PR #5166, opened 2021-08-17
Add "IPv4+IPv6" protocol value for IPsec mobile

jasperroloff:
This PR adds dual stack support for IPsec mobile VPN, as requested in #4819.

The code I used for this, is based on the [patch](https://github.com/opnsense/core/files/6137192/opnsense_ipsec.patch.txt) provided [here](https://github.com/opnsense/core/issues/4819#issue-831211355).

I've added some checks and comments.

### DIFF
--- a/src/etc/inc/plugins.inc.d/ipsec.inc
+++ b/src/etc/inc/plugins.inc.d/ipsec.inc
@@ -1,6 +1,8 @@
 <?php
 
 /*
+ * Copyright (C) 2021 Jasper Roloff
+ * Copyright (C) 2021 Jürgen Walch
  * Copyright (C) 2019 Pascal Mathis <mail@pascalmathis.com>
  * Copyright (C) 2016 Deciso B.V.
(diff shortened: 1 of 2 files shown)


src/etc/inc/plugins.inc.d/ipsec.inc
```diff
@@ -1,6 +1,8 @@
 <?php
 
 /*
+ * Copyright (C) 2021 Jasper Roloff
+ * Copyright (C) 2021 Jürgen Walch
```
fichtner, line 5: This really does not reach the threshold for originality so copyright claims are void.

jasperroloff: Oh, sorry – I saw other issues where the creators were asked to add a copyright notice.
Nevermind, I've reverted that to the original state now.

jasperroloff:
Hi, are there any news on this?

Kind regards
Jasper

Issue #4911, opened 2021-04-10
syslog-ng destination with RFC5424


muuuh:
Hello!

I've got a feature request concerning adding RFC5424 format to syslog-ng destinations config file.

**Issue:**
I've had a problem logging firewall-events to an Promtail / Loki / Grafana stack (which seems an alternative to ELK) via "syslog-ng". I've set up the syslog destination at "System: Settings: Logging / targets". With my tcp syslog destination (Promtail) for the Application "filter(filterlog)".
Promtail has currently this config:
```
  - job_name: syslog
    syslog:
      listen_address: 0.0.0.0:1514
      idle_timeout: 60s
      label_structured_data: yes
      labels:
        job: "syslog"
    relabel_configs:
      - source_labels: ['__syslog_message_hostname']
        target_label: 'host'
    pipeline_stages:
#IPv4
    - match:
        selector: '{host="OPNsense.localdomain"}'
        stages:
        - regex:
            expression: '^(?P<rule_number>[\d]+),(?P<sub_rule_number>[\d]*),(?P<anchor>[\d]*),(?P<tracker>[\d]*),(?P<interface>[\w\d]*),(?P<reason>[\w]*),(?P<action>[\w]*),(?P<direction>[\w]*),(?P<ip_version>[\d]*),[x\da-f]*,[x\da-f]*,(?P<packet_ttl>[\d]*),(?P<packet_id>[\d]*),(?P<fragment_offset>[\w]*),(?P<ip_flags>.*?),(?P<protocol_id>[\d]*),(?P<protocol_name>[\w]*),(?P<source_port>[\d]*),(?P<source_ip>[\d.]*),(?P<dest_ip>[\d.]*),(?P<dest_port>[\d]*),'
        - labels:
            interface:
            action:
            direction:
            protocol_name:
            source_ip:
            source_port:
            dest_ip:
            dest_port:
```

Data is received, but I got the following error from Promtail:
`promtail_1  | level=warn ts=2021-04-10T15:55:49.704424007Z caller=syslogtarget.go:170 msg="error parsing syslog stream" err="expecting a version value in the range 1-999 [col 5]"`

**Problem**
Promtail seems to only accept RFC5424-format - https://grafana.com/docs/loki/latest/clients/promtail/scraping/#syslog-receiver

**Feature request**
It worked, after I manually patched /usr/local/etc/syslog-ng.conf.d/syslog-ng-destinations.conf to
```
[...]

PR #5175, opened 2021-08-23
syslog-ng, switch our default logging format on disk to rfc5424

AdSchellevis:
**Important notices**

Currently some information is missing in local log files, mainly the severity which would help during debugging. Another problem with the current stored files is that dates don't contain the year in rfc3164 format, which means we do need to calculate them when converting to iso 8601.

Switch local stored format to rfc5424, we don't really need backwards compatibility it we're able to parse both formats. Other non local consumers should use the configurable syslog destinations or our api.

none

**Additional context**

while here, best add optional rfc5424 for external consumers as well (https://github.com/opnsense/core/issues/4911)


AdSchellevis:
manually merged in https://github.com/opnsense/core/commit/6cd8a971821f4382f077f8e4da3c80fc18ca2c9b

PR #5178, opened 2021-08-24
Remove duplicate.

kekek2:


AdSchellevis:
thanks!

src/www/interfaces_assign.php
```diff
@@ -214,7 +214,7 @@ interface_bring_down($id);   /* down the interface */
              * then ensure that we are not running DHCP on the wan which
              * will make a lot of ISP's unhappy.
              */
-            if (!empty($config['interfaces']['lan']) && !empty($config['dhcpd']['wan']) && !empty($config['dhcpd']['wan']) ) {
+            if (!empty($config['interfaces']['lan']) && !empty($config['dhcpd']['wan']) ) {
                 unset($config['dhcpd']['wan']);
             }
```
fichtner, line 219: @AdSchellevis this is a pretty weird conditional all things considered. I don't think this can be happening at all in our code although at some point in time it could have been the case.

fichtner: Also this happens post-config write so what does it do really...

AdSchellevis: @fichtner keeping separation of concerns in mind, dhcpd should check for its own if it makes sense to run, but to be honest, I don't really want to spend time here now finding out if it really does do that. Removing the obvious duplication is fine by me.

fichtner: I’m just saying looking more closely this is weird code, but more obviously dead code that does nothing so it can be safely removed. Can do that next week. 😊

AdSchellevis: well, it's not completely weird, although (a bit) obscure, this should prevent a dhcpd server running on wan and originates from https://github.com/pfsense/pfsense/commit/bda86e8fe6eee631252ef716157a068631bce392. Just not sure its worth the effort at the moment, since if you remove remove it, you would like to be sure dhcpd will do the right thing. The same construct also applies on console assignment.

https://github.com/opnsense/core/blob/b1241aeb74ccfa4ad3ccb3f7cc3b1370850ac955/src/opnsense/scripts/shell/setaddr.php#L541-L552

Hence my advise to leave this as is for now, the impact is managed elsewhere and currently it's not worth the effort.

fichtner: It can’t work anymore in this page after write config and before redirect. I removed this one and left the console intact, although that one should probably be reengineered as well at some point: it makes incorrect assumption at a time where ‘wan’ key could be anything, even a LAN if initial port config was reversed.

AdSchellevis: I missed the write_config part, feel free to ditch it in that case

PR #5184, opened 2021-08-29
CARP: Fix fetching VIP status

vnxme:
Issue:
- Sometimes OPNsense WebUI shows CARP VIP status improperly, i.e. it may report MASTER on [/carp_status.php](https://github.com/opnsense/core/blob/af587e812a7b8d96cf3fd84869dc7e00bbfe09b3/src/www/carp_status.php#L171) while ifconfig reports BACKUP and vice versa.

Reason:
- The grep expression used to fetch CARP VIP status in [etc/inc/interfaces.inc](https://github.com/opnsense/core/blob/af587e812a7b8d96cf3fd84869dc7e00bbfe09b3/src/etc/inc/interfaces.inc#L4503) is not precise enough. When you have CARP VHID 255 in MASTER and CARP VHID 2 in BACKUP, the function may return MASTER for VHID 2 in case ifconfig prints VHID 255 line before VHID 2 line.

Solution:
- Add a space to the grep expression after VHID.

### DIFF
--- a/src/etc/inc/plugins.inc.d/openvpn.inc
+++ b/src/etc/inc/plugins.inc.d/openvpn.inc
@@ -31,7 +31,6 @@
  * POSSIBILITY OF SUCH DAMAGE.
  */
 
-require_once("interfaces.lib.inc");
 
 function openvpn_configure()
 {
--- a/src/etc/inc/plugins.inc.d/openvpn.inc
+++ b/src/etc/inc/plugins.inc.d/openvpn.inc
@@ -903,7 +903,7 @@ function openvpn_restart($mode, $settings, $carp_event = false)
         list ($interface, $vhid) = explode("_vip", $settings['interface']);
         $carp_statuses = legacy_interface_details(get_real_interface($interface))['carp'];
         if (
-            $carp_statuses[array_search($vhid, array_column($carp_statuses, 'vhid'))]['status'] == 'BACKUP'
+            $carp_statuses[$vhid]['status'] == 'BACKUP'
         ) {
             /* do not restart a client if we are a CARP backup instance */
             return;
--- a/src/etc/inc/plugins.inc.d/openvpn.inc
+++ b/src/etc/inc/plugins.inc.d/openvpn.inc
@@ -901,9 +901,10 @@ function openvpn_restart($mode, $settings, $carp_event = false)
         strstr($settings['interface'], '_vip') && $mode == 'client'
     ) {
         list ($interface, $vhid) = explode("_vip", $settings['interface']);
-        $carp_statuses = legacy_interface_details(get_real_interface($interface))['carp'];
+        $interface_details = legacy_interface_details(get_real_interface($interface));
         if (
-            $carp_statuses[$vhid]['status'] == 'BACKUP'
+            !empty($interface_details) && !empty($interface_details['carp'][$vhid]) && 
+            $interface_details['carp'][$vhid]['status'] == 'BACKUP'
         ) {
             /* do not restart a client if we are a CARP backup instance */
             return;
--- a/src/etc/inc/plugins.inc.d/openvpn.inc
+++ b/src/etc/inc/plugins.inc.d/openvpn.inc
@@ -31,7 +31,6 @@
  * POSSIBILITY OF SUCH DAMAGE.
  */
 
-
 function openvpn_configure()
 {
     return array(
--- a/src/etc/inc/plugins.inc.d/openvpn.inc
+++ b/src/etc/inc/plugins.inc.d/openvpn.inc
@@ -31,6 +31,8 @@
  * POSSIBILITY OF SUCH DAMAGE.
  */
 
+require_once("interfaces.lib.inc");
+
 function openvpn_configure()
 {
     return array(
@@ -898,11 +900,16 @@ function openvpn_restart($mode, $settings, $carp_event = false)
     }
 
     if (
-        strstr($settings['interface'], '_vip') && $mode == 'client' &&
-        get_carp_interface_status($settings['interface']) == gettext('BACKUP')
+        strstr($settings['interface'], '_vip') && $mode == 'client'
     ) {
-        /* do not restart a client if we are a CARP backup instance */
-        return;
+        list ($interface, $vhid) = explode("_vip", $settings['interface']);
+        $carp_statuses = legacy_interface_details(get_real_interface($interface))['carp'];
+        if (
+            $carp_statuses[array_search($vhid, array_column($carp_statuses, 'vhid'))]['status'] == 'BACKUP'
+        ) {
+            /* do not restart a client if we are a CARP backup instance */
+            return;
+        }
     }
 
     @unlink("/var/etc/openvpn/{$mode_id}.sock");
--- a/src/www/carp_status.php
+++ b/src/www/carp_status.php
@@ -169,22 +169,26 @@ include("head.inc");
                         continue;
                     }
                     $icon = '';
-                    $intf_carp_statuses = $interfaces_details[get_real_interface($carp['interface'])]['carp'];
-                    $intf_status = $intf_carp_statuses[array_search($carp['vhid'], array_column($intf_carp_statuses, 'vhid'))]['status'];
+                    $carp_statuses = $interfaces_details[get_real_interface($carp['interface'])]['carp'];
+                    $intf_status = $carp_statuses[array_search($carp['vhid'], array_column($carp_statuses, 'vhid'))]['status'];
                     if (($carpcount > 0 && !$status)) {
                         $icon = "fa fa-remove fa-fw text-danger";
                         $intf_status = 'DISABLED';
+                        $intf_status_i18n = gettext('DISABLED');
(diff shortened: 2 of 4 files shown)


fichtner:
Thanks for the find. It should be fixed. But stacking grep here and calling ifconfig without a parser is a waste of resources and precision. We already have a good parser in interfaces.lib.inc and we should use it.

AdSchellevis:
the whole `get_carp_interface_status()` is flawed to begin with,  `gettext()` the results so the other end needs to translate again in order to compare, e.g.

https://github.com/opnsense/core/blob/1b9e263195c82824f172038c7dd154863dfcdae0/src/etc/inc/plugins.inc.d/openvpn.inc#L902

It would likely be better to refactor both status pages to use `legacy_interfaces_details()` directly so it only needs to call ifconfig once as well.  For easier search ability one could argue we best index the carp statuses by vhid, in which case it's almost a one-liner to ditch the function.

https://github.com/opnsense/core/blob/20b1f87d8896f11da030eab09544f80521725a99/src/etc/inc/interfaces.lib.inc#L363-L368

vnxme:
@fichtner @AdSchellevis Thank you for your prompt replies. I couldn't agree more with you when it comes to stacking grep and calling ifconfig multiple times. And you are certainly right proposing many ways of refactoring the code which seems to be far from ideal. 

However, this is the code which has existed in the project for some time now, and whatever it looks like it may give proper or improper results. I found the case when a bad-looking piece of code fails to do what it was intended to, and I suggest a simple way to fix it. This PR is not intended to make the piece of code in question look more attractive or effective. I believe we should should fix the issue as step #1, then we can discuss how to make the code which gives correct results even better as step #2.

AdSchellevis:
If it would be a complicated fix, I might agree, but in this case we better make sure to do it right the first time, better cleanup and fix in one pass to avoid additional noise. If you do want to give it a try, just let us know, otherwise we'll just treat this PR as a ticket for something we should fix.

vnxme:
@AdSchellevis I fully understand your approach and I strive for perfection myself, but this way the issue may remain unfixed for a long time unless someone including you and me has enough time for it. As for me, I would my first priority to WebUI correctness, leaving the second one to efficiency and attractiveness of the code.

Please, see 26bfe20 and bf195f3. What I don't understand is why `legacy_interfaces_details()` is called "legacy" if you propose using it instead of other functions without "legacy" in their names... Having carp_status.php and carp_status.widget.php refactored we have only [openvpn.inc](https://github.com/opnsense/core/blob/1b9e263195c82824f172038c7dd154863dfcdae0/src/etc/inc/plugins.inc.d/openvpn.inc#L902) still calling `get_carp_interface_status()`. Do you think we should keep this function in interfaces.inc for any reason?

AdSchellevis:
Thanks!

> What I don't understand is why legacy_interfaces_details() is called "legacy" if you propose using it instead of other functions without "legacy" in their names...

Understandable, it's in the legacy code base which is what we developed it for, hence the legacy part in its name. Realistically we're not going to refactor it into classes (or python scripts) any time soon.

> Having carp_status.php and carp_status.widget.php refactored we have only openvpn.inc still calling get_carp_interface_status(). Do you think we should keep this function in interfaces.inc for any reason?

I'll take a look what todo with it.

fichtner:
We are talking 1-2 weeks here, not months of going unfixed in any case. The issue has not come up before and it may mean not many are affected to begin with, which makes your patch the perfect intermediate help to apply manually for the time being.

legacy prefix was used for the dreadful pfSense-php module functions which were hard to maintain and even harder to fix quickly (can’t patch C code as easily as opnsense-code). Not a lot of those should still be visibly used by .php pages and the remaining ones can be renamed given that it doesn’t interrupt updates and plugin compatibility too much.


Cheers,
Franco

src/www/carp_status.php
```diff
@@ -29,6 +29,7 @@
 
 require_once("guiconfig.inc");
 require_once("interfaces.inc");
+require_once("interfaces.lib.inc");
```
fichtner, line 32:
```suggestion
```

included by interfaces.inc for portability

src/www/widgets/widgets/carp_status.widget.php
```diff
@@ -29,17 +29,20 @@
 
 require_once("guiconfig.inc");
 require_once("interfaces.inc");
+require_once("interfaces.lib.inc");
```
fichtner, line 32:
```suggestion
```

same here

src/etc/inc/plugins.inc.d/openvpn.inc
```diff
@@ -31,6 +31,8 @@
  * POSSIBILITY OF SUCH DAMAGE.
  */
 
+require_once("interfaces.lib.inc");
```
fichtner, line 34:
```suggestion
```

includes for plugins go to calling pages, not the plugin itself.. likely openvpn page already loads interfaces.inc... if not it should now

vnxme:
@AdSchellevis @fichtner Thanks again for your replies. In 3ae8a2f I eliminated calling `get_carp_interface_status()` by openvpn.inc so that one could safely drop this function. 

By the way, the fact that no one else complained about this CARP status issue doesn't imply no one got problems because of it. I posted a [forum question](https://forum.opnsense.org/index.php?topic=23268.0) about my CARP issues with unfortunately no relevant answers so far. And this find comes from my numerous attempts to debug and understand what's wrong with my CARP setup. Now I have learned I can't fully trust OPNsense WebUI unless this issue is fixed.

src/www/carp_status.php
```diff
@@ -163,26 +163,28 @@ interface_vip_bring_down($vip);
...
                         $icon = "fa fa-play fa-fw text-muted";
-                    } elseif ($intf_status == gettext('INIT')) {
+                    } elseif ($intf_status == 'INIT') {
                         $icon = "fa fa-info-circle fa-fw";
                     }?>
                 <tr>
                   <td><?=convert_friendly_interface_to_friendly_descr($carp['interface']) . "@{$carp['vhid']}" ;?></td>
                   <td><?=$carp['subnet'];?></td>
-                  <td><span class="<?=$icon;?>"></span> <?=$intf_status;?></td>
+                  <td><span class="<?=$icon;?>"></span> <?=gettext($intf_status);?></td>
```
fichtner, line 187: this will be broken, no static text in gettext(), part of why the weirdness existed before

fichtner: (I don't mind dropping the gettext altogether)

src/www/widgets/widgets/carp_status.widget.php
```diff
@@ -52,18 +54,18 @@
...
+      } elseif ($status == "BACKUP") {
           echo "<span class=\"fa fa-play fa-fw text-muted\" title=\"$status\" ></span>";
-      } elseif ($status == gettext("INIT")) {
+      } elseif ($status == "INIT") {
           echo "<span class=\"fa fa-info-circle fa-fw\" title=\"$status\" ></span>";
       }
       if (!empty($carp['subnet'])):?>
         &nbsp;
-        <?=htmlspecialchars($status);?> &nbsp;
+        <?=htmlspecialchars(gettext($status));?> &nbsp;
```
fichtner, line 68: same here...

src/www/carp_status.php
vnxme: > this will be broken, no static text in gettext(), part of why the weirdness existed before

Do you mean `gettext()` doesn't accept dynamic text, only static one? I haven't caught any php warnings/errors on my test system with gettext in place.

fichtner: PHP doesn't care, but translation generator reads all static text from gettext() calls to prepare a translation. You can't read variables from there since the text is never executed/analysed.

vnxme: Got it, thank you. Added a few commits to resolve these translation issues.

src/www/widgets/widgets/carp_status.widget.php
vnxme: Fixed.

AdSchellevis:
I think we can simplify this a bit further with https://github.com/opnsense/core/commit/e44fd48b6ada4c4d569e3de58c36f220f06d97f4, if carp entries are sequenced by vhid's, we can refactor these:

```
$carp_statuses = $interfaces_details[get_real_interface($carp['interface'])]['carp'];
$intf_status = $carp_statuses[array_search($carp['vhid'], array_column($carp_statuses, 'vhid'))]['status'];
```

with something like (including array safeguards):
```
$intf = get_real_interface($carp['interface']);
$intf_status = !empty($interfaces_details[$intf]) && !empty($interfaces_details[$intf]['carp'][$carp['vhid']]) ? $interfaces_details[$intf]['carp'][$carp['vhid']]["status"] : null;
```

(given the length of the statement, the array check might be easier to integrate into an existing if statement)

vnxme:
@AdSchellevis Thanks, done.

AdSchellevis:
@vnxme thanks! merged (after some small cleanups) in https://github.com/opnsense/core/commit/53cd1c7183f44599cf36483bb17775d819ff07bb.

PR #5186, opened 2021-08-30
IPSec Management - Added new identifier type Raw to allow passing My identifier as is without changes.

vityank:
Setup:
My end: OPNSense
Other End: Sophos Hardware Router

Preface:
After upgrade of OPNSense from 21.1.9_1 to 21.7(And then 21.7.1) my P1 authentication no longer works:

I started to receive those in the log:
14[IKE] <con1|17> received AUTHENTICATION_FAILED notify error
14[ENC] <con1|17> parsed IKE_AUTH response 1 [ N(AUTH_FAILED) ]

After digging around a little I came into this:
https://forum.opnsense.org/index.php?topic=22197.0

While the change there is fine, it has a big drawback: while keyid was changed, nothing provided to allow previous behaviour(tag w/o type prefix) and I found quickly by checking ipsec.conf that it was indeed cause of regression on my setup.

I've attached my patch which ads new type 'Raw tag' to the types list, which passes the tag as is, w/o modifications, and yet doesn't break the keyid changes you made before.

The patch is for 21.7/stable and unfortunately doesn't apply to master, due to seemingly refactors done there.

### DIFF
--- a/src/www/vpn_ipsec_phase1.php
+++ b/src/www/vpn_ipsec_phase1.php
@@ -286,6 +286,8 @@ if ($_SERVER['REQUEST_METHOD'] === 'GET') {
         $input_errors[] = gettext("Please enter an address for 'My Identifier'");
     } elseif ($pconfig['myid_type'] == "keyid tag" && $pconfig['myid_data'] == "") {
         $input_errors[] = gettext("Please enter a keyid tag for 'My Identifier'");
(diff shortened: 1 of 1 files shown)


fichtner:
No PRs to stable branches please.

AdSchellevis:
same as https://github.com/opnsense/core/commit/92ea6508887fef8b3fb4457c24163de52ba82fb6

PR #5196, opened 2021-09-05
Update authgui.inc

Greelan:
Adding "OPNsense" to the Title text to distinguish the page from login pages for other services (useful in browser tabs)

### DIFF
--- a/src/etc/inc/authgui.inc
+++ b/src/etc/inc/authgui.inc
@@ -318,7 +318,7 @@ function display_login_form($Login_Error)
     <meta name="copyright" content="" />
     <meta name="viewport" content="width=device-width, initial-scale=1, minimum-scale=1" />
 
-    <title><?=gettext("Login"); ?></title>
+    <title><?=gettext("OPNsense Login"); ?></title>
 
     <link href="<?= cache_safe("/ui/themes/{$themename}/build/css/main.css") ?>" rel="stylesheet">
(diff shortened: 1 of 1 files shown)


fichtner:
This is tricky: for one now one can assert which OPNsense version this is (pre-title change and post-title change) and we really should use the "branding" name here instead of hardcoding OPNsense.

As a general rule we try to avoid hardcoding it in translations. Let's test that theory...

    # git grep 'gettext.*OPNsense'
    src/www/diag_backup.php:    'OPNsense' => gettext('OPNsense Additions'),        /* XXX need specifics */
    src/www/interfaces.php:                                <input name="adv_dhcp_pt_values" type="radio" value="OPNsense"/><?=gettext("OPNsense Default");?>
    src/www/system_authservers.php:                      <?= gettext("Limit the groups which may be used by ldap, keep empty to consider all local groups in OPNsense. ".

Greelan:
Just a suggestion. Feel free to close.

mimugmail:
Yep, I'll take care :)

fichtner:
I don't want to complain or block, just dropped my review thoughts yesterday. I'll take care of this today :)

fichtner:
Had to work around some bitrot in this area thus a bit larger in the final patch. No ETA for inclusion yet. Thanks for the suggestion!

Greelan:
Ah, I see what you mean now by the reference to branding, looking at that commit. Point noted.

PR #5203, opened 2021-09-09
root: add lualoader brand and logo

kevans91:
Note that this is currently dependent on some functionality in upstream e19d9a9b6e44 and ed107537b43c being MFC'd; I'll issue an opnsense/tools PR after I MFC those to stop forcing forthloader.


fichtner:
Nice, thanks!

PR #5206, opened 2021-09-11
Correct the melody database directory in manpage

whc2001:
Blindly copied the path, added my own musical score and wondering why it doesn't work.


fichtner:
Nice spot, thanks!

Issue #5219, opened 2021-09-17
[FR] Quick "log all custom fw rules" switch


kulikov-a:
**Important notices**

Before you add a new report, we ask you kindly to acknowledge the following:

- [x] I have read the contributing guide lines at https://github.com/opnsense/core/blob/master/CONTRIBUTING.md
- [x] I am convinced that my issue is new after having checked both open and closed issues at https://github.com/opnsense/core/issues?q=is%3Aissue

**Is your feature request related to a problem? Please describe.**

Hi!
Since the forum raised the old topic about the possibility of enabling logging for all custom rules at once for traffic filtering debugging with a large number of rules (https://forum.opnsense.org/index.php?topic=10836.0), maybe it makes sense to add such an option outside the rules themselves (so that when the option is switched, the settings inside the rules are remain unchanged), for example, in the Logging settings?

**Describe the solution you like**

Something like

![log_all_1](https://user-images.githubusercontent.com/36099472/133805826-02ce2023-5b12-4709-b652-da59cda48963.PNG)

![log_all_2](https://user-images.githubusercontent.com/36099472/133805883-6b5f5acd-2043-489d-8161-3857e827b865.PNG)

Thanks!
**Describe alternatives you considered**

A clear and concise description of any alternative solutions or features you considered.

**Additional context**

Add any other context or screenshots about the feature request here or links to relevant forum thread or similar

kulikov-a:
considered not useful

Issue #5191, opened 2021-08-31
Allow logging on a specific fw rule to bypass filter.log and directly go to remote syslog server


pallebone:
**Important notices**

Before you add a new report, we ask you kindly to acknowledge the following:

- [ YES ] I have read the contributing guide lines at https://github.com/opnsense/core/blob/master/CONTRIBUTING.md
- [ YES ] I am convinced that my issue is new after having checked both open and closed issues at https://github.com/opnsense/core/issues?q=is%3Aissue

**Is your feature request related to a problem? Please describe.**

Logging in a granular way is very important in day to day, and I have this setup quite nicely. However I would also like to be able to log all traffic on an any any rule that would create too many log entries on the filter.log file when reviewing the live view or parsing the filter.log file.

As a result of this need to log all traffic, but not actually refer to it daily, I would like to be able to set logging on a specific firewall rule to only log to a remote syslog server that can hold a large amount of raw traffic log data for a period of time. This will keep clear the daily live view log and plain log files stored on the filter.log file in /var/log/

At the same time the syslog server will capture all log data that is needed while not hindering the more granular daily use of live log on the firewall itself.

**Describe the solution you like**
I would like an option on a firewall rule where it states "log packets that are handled by this rule" to also have an option "log packets that are handled by this rule only to remote syslog"

This means the firewall will only forward on the logging to the remote syslog for this specific rule and not attempt to write it in the local log file at all.
Under remote syslog settings where one normally sets filter (filterlog) a new entry might need to be created to indicate an option for remote only to monitor appropriatly.

**Describe alternatives you considered**

It would be possible to abandon using live log or the local log files altogether and only use a remote syslog server, but this would reduce ease of administration and the easy to use functionality of the log included on opnsense.

**Additional context**

None.

Many thanks for your consideration.
Kind regards
Pete

AdSchellevis:
It's definitely not very high on our priority list, but in theory you should be able to create more logging interfaces and choose where to send messages to in `pf`.  

In most cases local disc space isn't a huge issue, in which case it doesn't make a lot of sense to add the complexity needed for multiple log targets (it will need additional `pflog` interfaces, listeners to handle the messages {currently handled by `filterlog`} and glue to allow configuration, a lot of work for a small use case).

pallebone:
Ok no problem I understand. Maybe this will change in the future :)
Maybe even just a different log file on the disk for some rules would be helpful.
But no worries, thanks for checking anyway :)

Pete

PR #5221, opened 2021-09-17
Firewall: switch for all custom rules logging

kulikov-a:
Hi!
for possible discussion of the https://github.com/opnsense/core/issues/5219
thanks!

### DIFF
--- a/src/opnsense/mvc/app/library/OPNsense/Firewall/Plugin.php
+++ b/src/opnsense/mvc/app/library/OPNsense/Firewall/Plugin.php
@@ -1,7 +1,7 @@
 <?php
 
 /*
- * Copyright (C) 2016 Deciso B.V.
+ * Copyright (C) 2016-2021 Deciso B.V.
  * All rights reserved.
  *
  * Redistribution and use in source and binary forms, with or without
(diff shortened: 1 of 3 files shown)


fichtner:
Oh,this can and will quickly kill the whole system on large traffic volumes and/or fill up the disk completely. Selective logging per rule is a lot better and safer. Needle in haystack search is another issue here coupled with enormous noise generated.

kulikov-a:
@fichtner completely agree that this opens up a shortcut to killing the system (and using the live log to find a triggered rule will only speed up the process). on the other hand, if an administrator decides to kill the system by debugging traffic this way on a production system during a high load period, he will successfully do it manually by enabling logging on all the rules that he can reach )
perhaps the compromise would be to add a confirmation popup (with an additional warning) before enabling the option?
__
I really don't see it as a lifesaving feature (but only as useful in certain circumstances, like https://github.com/opnsense/core/issues/3124). if you and @AdSchellevis  think this option is too dangerous, I will close requests immediately )

AdSchellevis:
I'm not really a fan of this either to be honest, there's some relation to the discussion we had earlier about being able to feed all to an external log target (https://github.com/opnsense/core/issues/5191), but the scope of doing this properly is a lot wider and not really on our priority list. (one could question if filterlog shouldn't be ditched in the same pass and if the current format is suitable enough for external consumers for example)

kulikov-a:
@fichtner @AdSchellevis understood, thanks)
closing requests as not useful

PR #5231, opened 2021-09-22
diag_testport.php: set verbose

kulikov-a:
Hi!
if "Show Remote Text" (no `-z` flag) set and there is no text to display, `nc` may not output the result at all.
can we add `-v` option to display the connection result in this case?
thanks!


AdSchellevis:
@kulikov-a sure, it shouldn't hurt.

kulikov-a:
thanks!

PR #5235, opened 2021-09-24
configd_ctl.py: catch broken pipe on event handler

kulikov-a:
Hi!
can we discuss `configd_ctl.py` behavior when a pipe breaks? (ref. https://forum.opnsense.org/index.php?topic=24868.0) ?
afaik `readline()` not catching EOFError and just returns empty string. maybe it is worth adding a check for such a condition?

and could you please explain why the `list` is used if only one line is read?

thanks!!

### DIFF
--- a/src/opnsense/service/configd_ctl.py
+++ b/src/opnsense/service/configd_ctl.py
@@ -125,7 +125,11 @@ if args.e:
         rlist, _, _ = select([sys.stdin], [], [], args.t)
         if rlist:
             last_message_stamp = time.time()
-            stashed_lines.append(sys.stdin.readline())
+            r_line = sys.stdin.readline()
+            if len(r_line) == 0:
+                #EOFError. pipe broken?
+                sys.exit(-1)
+            stashed_lines.append(r_line)
 
         if len(stashed_lines) >= 1 and (args.t is None or time.time() - last_message_stamp > args.t):
             # emit event trigger(s) to syslog


AdSchellevis:
> and could you please explain why the list is used if only one line is read?

This one I can easily answer, the PR itself I have to check later. It's a common best practice with python to avoid over excessive object creation, try to time the next code samples:

[1]
```
tmp = ""
for i in range(50000000):
  tmp = tmp + "1234"

print (len(tmp))  
```

[2]
```
tmp = []
for i in range(50000000):
  tmp.append("1234")
  
print (len("".join(tmp)))    
```

remember, everything in python is an object :)

kulikov-a:
very intelligible example, thanks!!)

to the conversation, what I managed to notice when this behavior happened after the upgrade to 21.7.3:
-two `configd_ctl.py -e -t 0.5 system event config_changed` instances one of which was eating up the cpu
-system log file grew in size by 1.7GB in 11 minutes.
-the file is filled with lines like `event @ <epoch timestamp> msg:` 
while the timestamp did not change over time (11 min), msg is empty
then I rebooted the vm

an increase in cpu consumption is easy to reproduce by killing a syslog-ng with `kill -KILL`
can not reproduce the log filling yet

AdSchellevis:
@kulikov-a this looks reasonable indeed, thanks!

kulikov-a:
thanks!

Issue #5147, opened 2021-08-06
monit issue: All events disabled - but still sending alert notifications via mail 


f45tb00t:
**Important notices**

Before you add a new report, we ask you kindly to acknowledge the following:

- [X ] I have read the contributing guide lines at https://github.com/opnsense/core/blob/master/CONTRIBUTING.md
- [X ] I am convinced that my issue is new after having checked both open and closed issues at https://github.com/opnsense/core/issues?q=is%3Aissue

**Describe the bug**
I wanted to monitor a link of the FW itself. I was not very sure which of the "Events" I should use under "Alert settings". Therefore I tried it with trial and error. I ended up with "Do not send alerts for the following events but on all others." enabled and checked all items one by another until all were checked.
Basically I would expect that when I set "Do not send alerts for the...." and I have checked all of them, that no emails about the Link Down would be send in any way. Suprisingly I still get emails. Even with everything "disabled"
I've reloaded the plugin also many times. Result is the same.
As all are disabled, and I'm still getting the mails, I assume some setting must be hidden in the drop down list and I cannot see it.

=> What would be the correct "Event" for this type I'm looking for?

**To Reproduce**

1. Go to Services - monit - settings - Alert Settings - pass your email settings
2. Click on "not on"
3. Select "all" Events
4.  go to Service Settings - add new e.g LINK_DOWN
5. Type = Network, Add IP of one of your interfaces of the FW, Tests; NetworkLink, Rest standard...

**Expected behavior**

With the given settings I would not expect to get an alarm via email.

**Describe alternatives you considered**

Like said, I did trial and eror to find the correct "Event" in fact even when all are disabled due to the "no", I get what I wanted. But I guess this is not the supposed way to be.

**Screenshots**

If screenshots are needed, I can provide some.

**Relevant log files**

No logs so far. 

**Additional context**

none

**Environment**

Software version used and hardware type if relevant, e.g.:

OPNsense 21.7.1 (amd64, OpenSSL).

fbrendel:
The correct event type is `Link down` but it was missing.
Fixed in #5242

AdSchellevis:
closed in https://github.com/opnsense/core/pull/5242, use the following command to try the patch on 21.7.x:

```
opnsense-patch  466ac29
```

f45tb00t:
Thanks a lot :)

PR #5242, opened 2021-09-28
monit: add Link event to alert settings

fbrendel:
The `Link` event for alert settings was missing.
See #5147


AdSchellevis:
@fbrendel thanks!

PR #5247, opened 2021-09-28
radvd/virtual IPs: Implement prefix advertisement switch

vnxme:
**Issue:**

If router advertisement is enable on the interface, now the radvd configuration includes all the prefixes calculated from the addresses assigned to this interface. While it's ok for most users, in some circumstances one may opt for not advertising some of the prefixes.

**Solution:**
Add an IPv6 RA checkbox (on by default for compatibility) to the VIPs edit page and take it into account in the radvd configuration management script.

**To do list (in my tests the solution works fine, but help/advice is appreciated to make it even better):**
1) Radvd configuration isn't updated after applying VIP changes. Thus you must manually restart radvd to take into account the newly added checkboxes (and other changes, to be more precise).
2) The checkbox is shown and its value saved for any interface which you can assign an IPv6 alias / CARP to. But router advertisements configuration is possible for a few "real" interfaces only. E.g. we show the checkbox for an IPv6 alias assigned to a loopback, but it is useless, since such an interface isn't included into radvd configuration. So far I have failed to find the attribute of the interface based on which we can hide/show the checkbox.

### DIFF
--- a/src/etc/inc/plugins.inc.d/dhcpd.inc
+++ b/src/etc/inc/plugins.inc.d/dhcpd.inc
@@ -316,7 +316,7 @@ function dhcpd_radvd_configure($verbose = false, $blacklist = array())
         }
 
         foreach (config_read_array('virtualip', 'vip') as $vip) {
-            if ($vip['interface'] == $dhcpv6if && is_ipaddrv6($vip['subnet'])) {
+            if ($vip['interface'] == $dhcpv6if && is_ipaddrv6($vip['subnet']) && empty($vip['noipv6ra'])) {
                 $subnetv6 = gen_subnetv6($vip['subnet'], $vip['subnet_bits']);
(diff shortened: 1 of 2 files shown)


fichtner:
(it also assumes that you use manual binding of services to interfaces, otherwise :: and 0.0.0.0 overlap anyway and you can't stop anything from binding/blocking a port)

fichtner:
@vnxme I understand your concerns, but I want to avoid going backwards if we go two steps ahead instead of one. I doubt a lot of people running into the state where they require the no-bind VIP solution and also most scenarios are generally avoidable or easily worked around with NAT rules. What is required now, however, is proper documentation of the no-bind  feature and how to set up router advertisements with these non-obvious solutions to excluding networks plus running a successful CARP setup with it. Once we have this and the feature(s) out in the field it's a lot easier to see what needs amendment when being used out in the wild. :)

Closing this then. If you run into any issues with it let me know.


Cheers,
Franco

PR #5259, opened 2021-10-06
system: add option to reload templates after restore

fraenki:
In https://github.com/opnsense/plugins/issues/2474 it was reported that the HAProxy configuration file was missing after restoring the configuration. It looks like a valid complain to me, because the template reload needs to be triggered to generate the configuration file.

This patch adds an option to trigger a template reload after restoring a config backup. What do you think?

(Side note: this does not yet solve the HAProxy startup issue, I'll have to make a small adjustment to `setup.sh` for this to work. But this depends on this PR.)

(Disclaimer: I haven't tested this patch yet, I first like to know whether this approach is good or bad.)

### DIFF
--- a/src/www/diag_backup.php
+++ b/src/www/diag_backup.php
@@ -231,6 +231,9 @@ if ($_SERVER['REQUEST_METHOD'] === 'GET') {
                     if (!empty($pconfig['rebootafterrestore'])) {
                         $do_reboot = true;
                     }
+                    if (!empty($pconfig['templatereloadafterrestore'])) {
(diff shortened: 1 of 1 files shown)


fichtner:
looking at opnsense/plugins#2474 it says "3)XML backup restore of whole device - **reboot**" I don't think that report is correct as on reboot everything is rewritten already?

Besides, we don't start any services on restore either so it's ok to leave the templates out of the picture (it only really applies if reboot is not selected).


Cheers,
Franco

PR #5269, opened 2021-10-11
jquery.bootgrid.js:  convert on append

kulikov-a:
Hi!
since https://github.com/opnsense/core/commit/d641c00528e95e44bd74f7ed232ec2858f6358bf SIZE and RES values in System: Diagnostics: Activity are displayed incorrectly and sorting stopped working
`jquery.bootgrid` forgets to apply conversion on `append`
code is taken from https://github.com/MoveInc/jquery-bootgrid/pull/4
thanks!

### DIFF
--- a/src/opnsense/www/js/jquery.bootgrid.js
+++ b/src/opnsense/www/js/jquery.bootgrid.js
@@ -1446,9 +1446,14 @@ Grid.prototype.append = function(rows)
         var appendedRows = [];
         for (var i = 0; i < rows.length; i++)
         {
-            if (appendRow.call(this, rows[i]))
+            var row = rows[i];
+            for (var j = 0; j < this.columns.length; j++) {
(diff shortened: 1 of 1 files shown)


kulikov-a:
thanks!)

AdSchellevis:
@kulikov-a found a small regression https://github.com/opnsense/core/commit/a68ed580d0484a77cc0639afc7793d9e141cb4ec, it seems that I've worked around this limitation a long time ago and was wondering why there was an "invalid date" now.  Just wanted to let you know in case you trip into similar issues elsewhere.

kulikov-a:
@AdSchellevis thanks! (I would definitely suffer with this. did a search and did not seem to find any more places with double conversion)

AdSchellevis:
@kulikov-a I didn't see any others either, but just in case so we both know what to look for :)

PR #5275, opened 2021-10-12
VPN / IPsec / Advanced settings - add closeaction parameter to UI defaulting to "none"

pmhausen:
Sorry, I messed up the last one. There was a completely unwarranted change to vpn_ipsec.php in there that did not belong to the patch.

Here we go again:

StrongSwan's closeaction parameter is an important option to keep phase 2 SAs up and running in heterogenous enterprise environments. It is not exposed in the UI until now and unfortunately defaults to "none" if not explicitly set.

I expanded the UI to provide an option, defaulting to the current behaviour.

I put this in the phase 1 SA edit form, although it is strictly a phase 2 parameter, because I cannot imagine a situation when you want to set this differently per IPsec tunnel for one and the same IKE peer. Similar to DPD ...

Kind regards,
Patrick


AdSchellevis:
@pmhausen how about https://github.com/opnsense/core/commit/bb9b8820c6a2725730598bd3ee77b11e626b1186 ? I couldn't watch the struggle :) these legacy pages are awful.

PR #5277, opened 2021-10-13
system activity: show all threads and correct WCPU

kulikov-a:
Hi!
just small suggestions for the System: Diagnostics: Activity
-Show all threads on page since we ask for in `activity.py` (for this, the `data-identifier="true"` has been removed so that the bootgrid does not monitor the uniqueness of PID)
-Since the non-interactive `top -n` mode displays one screen by default, and the output of the first display does not give "data about the percentage of time spent by the CPU in every state (https://www.freebsd.org/cgi/man.cgi?top(1) see `-d` option description ) ( `activity.py` output gives zero WCPU for almost all processes), maybe it is worth specifying a `-d` option and parsing the second display? (since this slightly increases the waiting time for a response, the value of the `noResults` label is changed on the page)
imho this can produce more adequate output on the Activity page

Thanks!!

### DIFF
--- a/src/opnsense/mvc/app/views/OPNsense/Diagnostics/systemactivity.volt
+++ b/src/opnsense/mvc/app/views/OPNsense/Diagnostics/systemactivity.volt
@@ -39,9 +39,6 @@ POSSIBILITY OF SUCH DAMAGE.
                 ajax: false,
                 selection: true,
                 multiSelect: true,
-                labels: {
-                    noResults: "{{ lang._('Waiting for data...') }}"
-                }
             };
             if ($("#grid-top").hasClass('bootgrid-table')) {
                 $("#grid-top").bootgrid('clear');


AdSchellevis:
@kulikov-a I'm not sure it's worth the effort at the moment, although I don't mind too much about the extra wait.  Not having a valid primary key isn't too great, but that already was the case, so sounds a bit buggy from my code. 

The challenge is if you would add a primary key, it would be `tid`, which can't be combined with `pid` in a single pass on top, in which case you will need to glue `procstat -ak` to find the corresponding pid. It all starting to look like work....

AdSchellevis:
@kulikov-a it looks like you really want this 🙃, I'll try to take a look and merge somewhere this weekend or next week. I don't have strong objections, just didn't really mis the feature. A primary key in general is always better to have in all datagrids (if possible), I'm not 100% sure if the javascript component cares a lot if it misses, but being able to identify a row uniquely is more or less a best practice. (I just mistakingly assumed the pid was unique ;) )

AdSchellevis:
@kulikov-a minor cleanups in https://github.com/opnsense/core/commit/d45eb09eed6cf43f28434369be5a2f81b6003bde, only thing I'm currently not too fond of is the "Waiting for data..." when no data is found when searching.

kulikov-a:
@AdSchellevis looks good! (I suffered a bit with the records->fieldnames matching  when inserted pids into the beginning of the fieldnames)

> only thing I'm currently not too fond of is the "Waiting for data..." 
Yes, sorry, I also noticed it after I made the pr. I'll try to play with bootgrid and do it right. I will not close the request yet, if you don't mind

thanks!

AdSchellevis:
@kulikov-a looks acceptable, yes, thanks, I'll take it.

kulikov-a:
yay! thanks a lot!!

Issue #4452, opened 2020-11-11
DHCP - Dynamic DNS Registration - DDNS hostname is not set in configuration with static config


rfc4711:
When setting static host-names in dhcp4 configuration, the original client identifier is passed to the dns server, not respecting either the client identifier or the static configured host-name in the dhcp configuration

**To Reproduce**
1. DHCPv4 configuration
2. Setup DDNS to a remote DNS Server
3. Add a static entry for a mac address to add a forward and reverse map of the host name in DNS (Overriding the client-host-name) 
4. The log shows the "original client identifier sent to the DNS server for forward and reverse map 

**Expected behavior**
The static-host name or configured client-identifier should be used to set a ddns-hostname entry in the dhcp configuration files, maybe a feature to provide a checkmark to override client host-name. However, what would be the point of setting a static entry for hostname if that is not updated when ddns is configured set a forward/reverse map in DNS, ie 
```ddns-hostname "my-static-hostname";```

**Log**
```
2020-11-11T02:11:15 | dhcpd[75584] | Added reverse map from 108.131.168.10.in-addr.arpa. to HS200.skynet.net
-- | -- | --
2020-11-11T02:11:15 | dhcpd[75584] | Added new forward map from HS200.skynet.net to 10.168.131.108
```
**Configuration**
```
subnet 10.168.131.0 netmask 255.255.255.0 {
  pool {
    option domain-name-servers 172.17.17.44;
    ddns-update-style interim;
    range 10.168.131.150 10.168.131.250;
  }

  option routers 10.168.131.2;
  option domain-name "skynet.net";
  option domain-search "skynet.net","wifi.net,"dyn.net";
  ddns-domainname "skynet.net";
  option domain-name-servers 172.17.17.44;
  default-lease-time 7200;
  max-lease-time 86400;
  option ntp-servers 172.17.17.7;

}
host s_opt3_22 {
  hardware ethernet 50:c7:bf:ec:bc:87;
  option dhcp-client-identifier "hallway-light";
  fixed-address 10.168.131.108;
  option host-name "hallway-light";
    ddns-update-style interim;
}
```

**Environment**
OPNsense 20.7.4-amd64

**Comment by rfc4711** [comment]
While I am not an expert in coding, I think if we add to the following file:

src/etc/inc/plugins.inc.d/dhcpd.inc  line 1094:
```
$dhcpdconf .= "  option host-name \"{$dhhostname}\";\n";
[...]

PR #5289, opened 2021-10-19
update dhcpd.inc to allow ddns-hostname to be set

rfc4711:
per issue #4452, allow ddns-hostname to be set when dynamic zones are in use.


AdSchellevis:
same https://github.com/opnsense/core/pull/5288 ?

rfc4711:
Yes, I thought I had to submit to the active branches and the master.

PR #5292, opened 2021-10-19
Firewall / Diagnostics add 'rules' to Statistics

kulikov-a:
Hi!
new statistics page looks cool!
Is it possible to add Rules there too?
Imho it would be convenient when debugging the rules (to see the order of the rules after optimization, look at the actual syntax, and look at the statistics, of course etc.)
thanks!


fichtner:
@kulikov-a sorry for breaking this PR... script without suffix made me nervous ;)

PR #5305, opened 2021-10-24
system log widget: add filter

kulikov-a:
Hi!
Can we discuss adding a filter to the widget?
Answering the question posed earlier: yes, it seems to me that being able to save the filter in config might add a bit of convenience. Life scenario: user can setup the monit to monitor events\logs important to him, just save the filter with the "monit" value and always see a filtered log in lobby

and can you please explain why using cookies to pass validation results is not a good idea in this case?
by the way I can try to do this by changing the form submit behavior if it is more correct )

thanks!


kulikov-a:
a little unexpected)
Thanks!

PR #5311, opened 2021-10-27
Prevent Browser Auto Fill Username/Password

NOYB:
Prevent browser from auto filling certain username/password fields with the system login credentials in new forms or if left blank.
 * Don't globally override existing autocomplete attribute on input elements.
 * Apply the attribute autocomplete="new-password" to password input elements.

Examples:

System:
 Access:
  User +
  Servers +
  Tester

 Configuration:
  Backups - Google Drive

 High Availability:
  Settings

Interfaces:
 Point-to-point:
  Devices +

VPN:
 OpenVPN:
  Clients +

Services:
 Dynamic DNS +
 OpenDNS

See also corresponding commit in plugins for DynDNS.


AdSchellevis:
@NOYB looks good, thanks!

PR #5320, opened 2021-10-29
syslog-ng: add empty line to template

kulikov-a:
Hi!
add trailing new line after each file included so as not to control its presence in each inclusion
ref: https://github.com/opnsense/plugins/issues/2587#issuecomment-954873093
(example: `001-local.conf` does not contain trailing new line. so next file will be appended to the last line from it)
thanks!


AdSchellevis:
@kulikov-a thanks, this should ease the plugins implementation indeed.

kulikov-a:
thanks!)

PR #5321, opened 2021-10-29
VPN / IPsec / Advanced settings - add sha256_96 flag

pmhausen:
Hi folks,

another minor enhancement to the IPsec options. Some legacy devices like Forcepoint Sidewinder incorrectly implement HMAC_SHA256 for phase 2.

With the same reasoning as for the "closeaction" parameter I added the field to the phase 1 form, because it applies to all phase 2 entries of a particular peer or not at all.

In general style I followed @AdSchellevis implementation of the "closeaction".

Kind regards,
Patrick

### DIFF
--- a/src/etc/inc/plugins.inc.d/ipsec.inc
+++ b/src/etc/inc/plugins.inc.d/ipsec.inc
@@ -1511,6 +1511,12 @@ function ipsec_configure_do($verbose = false, $interface = '')
                     $ealgosp1 = 'ike = ' . implode(',', array_reverse($list)) . '!';
                 }
 
+                if (!empty($ph1ent['sha256_96'])) {
+                    $sha256_96_line = "sha256_96 = {$ph1ent['sha256_96']}";
+                } else {
+                    $sha256_96_line = '';
+                }
+
                 if (!empty($ph1ent['closeaction'])) {
                     $closeaction_line = "closeaction = {$ph1ent['closeaction']}";
                 } else {
@@ -1675,6 +1681,7 @@ conn con<<connectionId>>
   installpolicy = {$installpolicy}
   type = {$parsed_phase2['type']}
   {$closeaction_line}
+  {$sha256_96_line}
   {$dpdline}
   {$inactivityline}
   {$keyingtriesline}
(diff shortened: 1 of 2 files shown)


src/etc/inc/plugins.inc.d/ipsec.inc
```diff
@@ -1511,6 +1511,12 @@ function ipsec_configure_do($verbose = false, $interface = '')
                     $ealgosp1 = 'ike = ' . implode(',', array_reverse($list)) . '!';
                 }
 
+                if (!empty($ph1ent['sha256_96'])) {
+                    $sha256_96_line = "sha256_96 = {$ph1ent['sha256_96']}";
```
AdSchellevis, line 1515: @pmhausen I would fix this to "yes" to make it more explicit that this is a boolean and can't contain anything else and save the value as a boolean too (I think the other checkboxes use a similar approach).

pmhausen:
Like this?

N.B. I just noticed the standard templating of the `{$foo_line}` entries leads to a lot of unnecessary empty lines in the final config. Which does not hurt unless one examines it manually. This is way better with the Jinja templating in the MVC modules.

AdSchellevis:
@pmhausen I would align it a bit more to the existing toggles, how about https://github.com/opnsense/core/commit/b585c331e5bcf17be1fabe0e83bbc29282da48bf?  The legacy templating (combining logic/storage and functionality into a single block) isn't too great indeed, styling isn't exactly the same in all files, although it has improved a lot over the past 7 years :)

pmhausen:
So it works without being in `$copy_fields`? OK ... thanks for the improvement. Yes, the usage of `!empty()` vs. `isset()` looks quite inconsistent, too. I checked the docs but that made me non the wiser ;)

AdSchellevis:
> So it works without being in $copy_fields? OK ... 

The copy fields is just a shorthand to avoid 1 on 1 copies for everything that really is exactly the same, which is being implemented here 
https://github.com/opnsense/core/blob/b585c331e5bcf17be1fabe0e83bbc29282da48bf/src/www/vpn_ipsec_phase1.php#L95-L103

> Yes, the usage of !empty() vs. isset() looks quite inconsistent, too. I checked the docs but that made me non the wiser ;)

I know, personally I do prefer to use the `empty()` operator as it evaluates an explicit `0` as false, which in our model implementation is also false (and is always stored as such, which in my humble opinion is more explicit)

Issue #2070, opened 2018-01-09
ipsec: what to do with the mobile peer id


fichtner:
https://github.com/opnsense/core/commit/0dd120

via: https://forum.opnsense.org/index.php?topic=6194.0

mimugmail:
Is this only a ike v1 issue? One of the apprentices told me to be bored so I'll set up a test environment.

I also had struggles months ago troubleshooting Mac integrated client .. so this might bei the issue I had too

fichtner:
could be. although for macOS I do not recommend v1 at all.

fichtner:
I'm closing this, again, due to lack of feedback.

wh0else:
Any update on this? IPsec/IKEv2 road warrior VPN with EAP-TLS/Mutual RSA seems to need mobile peer id, at least on clients such as macOS and iOS.
Working in pfSense, as the setting is there

PR #5324, opened 2021-10-30
allow peer id for mobile phase1 & xauth_psk_server

arkste:
Hi,

I've migrated my old pfsense ipsec road warrior config to opnsense and was missing the peer id configuration, which is needed for some ios devices as a group name.

There was already a discussion about it in the forum https://forum.opnsense.org/index.php?topic=6194.0 and in #2070.

I was thinking about reverting 0dd120ff1aac446f0df9a3f82051b95b64eecf27, but dont want to cause issues in some other configurations and instead i've only added an exception for the authentication method "xauth_psk_server", since its definitly needed there and the configuration wont work without it. 

This fix is especially needed, since the official documentation also mentions a peer id here: 
https://docs.opnsense.org/manual/how-tos/ipsec-road.html#phase-1-proposal-authentication

### DIFF
--- a/src/www/vpn_ipsec_phase1.php
+++ b/src/www/vpn_ipsec_phase1.php
@@ -848,7 +848,7 @@ endforeach; ?>
                     </td>
                   </tr>
 <?php
-                  if (empty($pconfig['mobile'])):?>
+                  if ($pconfig['authentication_method'] == "xauth_psk_server" || empty($pconfig['mobile'])):?>
                   <tr class="auth_opt auth_eap_tls auth_psk auth_pubkey">
                     <td><i class="fa fa-info-circle text-muted"></i> <?=gettext("Peer identifier"); ?></td>
                     <td>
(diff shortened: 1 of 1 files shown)


AdSchellevis:
I remember this one, I was pretty sure the value of this field wasn't flushed to the `ipsec.conf` in any way. Can you create a diff between the config with and without the setting (`/usr/local/etc/ipsec.conf`)? It could also be an issue in our documentation.

arkste:
You're right, the value doesnt get written to the config file.

I thought the value is needed, because the ipsec connection wont be established without a group name on the iOS Client.
I thought that the peer id is needed, because its the group name on the clientside, but its not. 

After some more testing, the iOS client DOES need a group name, but it can be whatever... so this PR can be closed then.

AdSchellevis:
@arkste if you have the spirit for it, a PR on the docs would be appreciated :)

PR #5348, opened 2021-11-10
Eliminate duplicate "tmpfs" disk info (sys info widget)

NOYB:


NOYB:
Only been tested on two systems.  One physical zfs with RAM disk (21.7.4), the other VBox zfs without RAM disk (22.1.b).  Though the change appears simple and straight forward, still probably prudent to verify on at least a couple ufs with and without RAM disk before implementing.

### With RAM Disk
System Information Widget - Disk usage - ZFS with RAM Disk

### Without RAM Disk
System Information Widget - Disk usage - ZFS without RAM Disk

fichtner:
Just no.

Issue #4814, opened 2021-03-13
Allow network booting on different architectures within the DHCP gui


anthr76:
**Important notices**

Before you add a new report, we ask you kindly to acknowledge the following:

- [X] I have read the contributing guide lines at https://github.com/opnsense/core/blob/master/CONTRIBUTING.md
- [X] I have searched the existing issues and I am convinced that mine is new.

**Is your feature request related to a problem? Please describe.**

![image](https://user-images.githubusercontent.com/25143877/111034117-23b76080-83e2-11eb-9549-891c90cea0e6.png)

I need to set a arch to netboot from Arm64 which is a different EFI file. Per the [iPXE](https://ipxe.org/cfg/platform) docs the existing if statement can be expanded to include more arch's. 

**Describe the solution you like**

![image](https://user-images.githubusercontent.com/25143877/111034212-8577ca80-83e2-11eb-8da6-72de233a02ea.png)

Add an option to select a bootfile for arm64(aarch64). 

**Describe alternatives you considered**

I had edited `/var/dhcpd/etc/dhcpd.conf` though that gets overwritten when the dhcp service gets restarted.

As a workaround I will utilize the `else` for default bios filename but this doesn't work well if there's non UEFI x86 machines to be network booted as well.

**Additional context**

ARM and RISC are gaining in popularity and it's a requirement in my network to selectively boot a efi based on architecture.

PR #5361, opened 2021-11-18
Allow for ARM architectures in DHCP Network Boot options

keithamus:
I noticed https://github.com/opnsense/core/issues/4814 was raised, requesting the ability to specify alternate files for ARM architectures. It's a shame it was closed for lack of activity, but I presume that doesn't mean I cannot raise a patch for it.

This PR adds the architectures `00:10` (ARM 32 bit) as the field `filename32arm` and `00:11` (ARM 64 bit) as the field `filename64arm`. If not specified these default to `filename32`/`filename64` respectively.

This _does_ change the behavior slightly, as right now the current logic means these architectures fall through to using `filename`. I think it would be fine to try to keep that behaviour, if people feel that's important.

This is my first contribution here, I've read the CONTRIBUTING guidelines and I believe I've done everything necessary. I'm happy to license this work under 2-Clause BSD :slightly_smiling_face:


keithamus:
@fichtner @AdSchellevis thoughts on this PR? Would you like anything more done to this to resolve #4814?

src/etc/inc/plugins.inc.d/dhcpd.inc
```diff
@@ -1085,6 +1085,10 @@ EOD;
                 $dhcpdconf .= "    filename \"{$dhcpifconf['filename64']}\";\n";
                 $dhcpdconf .= "  } else if option arch = 00:09 {\n";
                 $dhcpdconf .= "    filename \"{$dhcpifconf['filename64']}\";\n";
+                $dhcpdconf .= "  } else if option arch = 00:0a {\n";
+                $dhcpdconf .= "    filename \"{$dhcpifconf['filename32arm'] || $dhcpifconf['filename32']}\";\n";
```
AdSchellevis, line 1089: shouldn't this be a `??` and outside the string block?

AdSchellevis:
@keithamus it has been busy, I have some doubts about the dhcp conf, but functionally I don't mind to add it. Can you check if it actually generates the correct config or change the source?

Just leaving a reference here where the numbers come from for future reference : https://www.iana.org/assignments/dhcpv6-parameters/dhcpv6-parameters.xhtml#processor-architecture

src/etc/inc/plugins.inc.d/dhcpd.inc
keithamus: The last time I worked on PHP code, 5.3 was the latest version, so please forgive my rustyness.

keithamus:
@AdSchellevis thanks for the comments, I've adjusted the code and tested it. It generates a suitable config IMO.

AdSchellevis:
@keithamus looks good, thanks!

Issue #5337, opened 2021-11-04
Syslog - remove clog support in 22.1


AdSchellevis:
**Important notices**

Before you add a new report, we ask you kindly to acknowledge the following:

- [x] I have read the contributing guide lines at https://github.com/opnsense/core/blob/master/CONTRIBUTING.md
- [x] I am convinced that my issue is new after having checked both open and closed issues at https://github.com/opnsense/core/issues?q=is%3Aissue

**Is your feature request related to a problem? Please describe.**

Currently we do support two different syslog systems for historic reasons, which given nowadays equipment seems to be less relevant anymore. Clog offers rotating logfiles, which could save log space on embedded systems, but often is misconfigured (setup large target slowing down log handling a lot) or just unneeded ([mem]disk space is less of an issue)

**Describe the solution you like**

Remove clog support, explain the correct usage of our logging (https://github.com/opnsense/docs/issues/357) and stick to a single log system and format.

**Describe alternatives you considered**

keeping clog around, which seems to be less logical.

**Additional context**

none

fichtner:
TODO/Questions:

- [ ] prefixes.php needs a fix for new log format or maybe even a rewrite
- [ ] core.inc references "syslogd" service
- [ ] rename system_syslogd_* to system_syslog_* and/or try to migrate that cruft to rc files
- [ ] diag_logs_settings.php calls killbyname('syslogd');
- [ ] remove async flag from system_syslogd_start()

kulikov-a:
hi!
Considering the upcoming changes and RFC 5424 migration, how silly is the idea to change the System: Log Files: General page a little bit more? change severity filter to something like "maximum severity" (single select) and make it "Warning" by default (and may be storing filter value in localStorage)?
(inspired by https://forum.opnsense.org/index.php?topic=25676.0 (apart from manners))

thanks!

AdSchellevis:
@kulikov-a I'm open for suggestions, I do expect some messages in core will use a wrong level at the moment by the way, but that will sort itself out eventually.

kulikov-a:
@AdSchellevis thanks! then I'll try to play with it and make a pr for discussion if something useful comes out

PR #5370, opened 2021-11-24
Logs: quick severity filter with "Warning" default

kulikov-a:
Hi!
just for possible discussion of  https://github.com/opnsense/core/issues/5337#issuecomment-974886511

Clarifications: Under certain circumstances, some processes and services can create quite a lot of noise in the logs. This is great for debugging, but can be annoying for the administrator during normal operation.

Initial idea: since the OPNSense finally completely switches to syslog-ng, RFC5424 support has been added, and even a severity filter has already been added (!  :thumbsup:), it may be worth considering to make the default log representation of the "warning logs" type (do not show by default lines reflecting normal behavior)?

Details: pr by default displays the logs filtered with "Warning" severity (UPD. badly put it: filtered with a level **below or equal to** "Warning"), the severity filter allows to specify the maximum severity level to fetch and saves the user's choice to localStorage.

thanks!

### DIFF
--- a/src/opnsense/mvc/app/views/OPNsense/Diagnostics/log.volt
+++ b/src/opnsense/mvc/app/views/OPNsense/Diagnostics/log.volt
@@ -158,8 +158,9 @@
           select.selectpicker({ header: header_val });
 
           // attach event handler each time header created
-          $("#exact_severity").on("click", function() {
-              filter_exact = !filter_exact
+          $("#exact_severity").on("click", function(event) {
+              event.stopPropagation();
+              filter_exact = !filter_exact;
               let select = $("#severity_filter");
 
               // set new select value to current value or highest value of multiselect
--- a/src/opnsense/mvc/app/views/OPNsense/Diagnostics/log.volt
+++ b/src/opnsense/mvc/app/views/OPNsense/Diagnostics/log.volt
@@ -26,14 +26,25 @@
 
 <script>
     $( document ).ready(function() {
+      var filter_exact = false;
+      let s_filter_val = "Warning";
       // map available severity values to array
       severities = $('#severity_filter option').map(function(){
           return (this.value ? this.value : null);
       }).get();
 
-      if (window.localStorage && localStorage.getItem('log_max_severity_{{module}}_{{scope}}')) {
-          $('#severity_filter').val(localStorage.getItem('log_max_severity_{{module}}_{{scope}}')).change();
+      if (window.localStorage) {
+          if (localStorage.getItem('log_filter_exact_{{module}}_{{scope}}')) {
+              s_filter_val = localStorage.getItem('log_severity_{{module}}_{{scope}}') ? localStorage.getItem('log_severity_{{module}}_{{scope}}').split(',') : [];
+              filter_exact = true;
+              $("#exact_severity").toggleClass("fa-toggle-on fa-toggle-off").toggleClass("text-success text-danger");
+              switch_mode(s_filter_val);
+          } else {
+              s_filter_val = localStorage.getItem('log_severity_{{module}}_{{scope}}') ? localStorage.getItem('log_severity_{{module}}_{{scope}}').split(',') : "Warning";
+              $('#severity_filter').val(s_filter_val).change();
+          }
       }
+
       let grid_log = $("#grid-log").UIBootgrid({
           options:{
               sorting:false,
@@ -53,7 +64,8 @@
               requestHandler: function(request){
                   if ( $('#severity_filter').val().length > 0) {
                       let selectedSeverity = $('#severity_filter').val();
-                      request['severity'] = severities.slice(0,severities.indexOf(selectedSeverity) + 1);
+                      // get selected severities or severeties below or equal to selected
+                      request['severity'] = filter_exact ? selectedSeverity : severities.slice(0,severities.indexOf(selectedSeverity) + 1);
                   }
                   return request;
               },
@@ -62,7 +74,7 @@
       });
       $("#severity_filter").change(function(){
           if (window.localStorage) {
-              localStorage.setItem('log_max_severity_{{module}}_{{scope}}', $("#severity_filter").val());
+              localStorage.setItem('log_severity_{{module}}_{{scope}}', $("#severity_filter").val());
           }
           $('#grid-log').bootgrid('reload');
       });
@@ -118,21 +130,78 @@
           }
           $('<a></a>').attr('href',download_link).get(0).click();
       });
+
+      $("#exact_severity").on("click", function() {
+          $(this).toggleClass("fa-toggle-on fa-toggle-off");
+          $(this).toggleClass("text-success text-danger");
+          if ($(this).hasClass("fa-toggle-on")) {
+              filter_exact = true;
+              if (window.localStorage) {
+                  localStorage.setItem('log_filter_exact_{{module}}_{{scope}}', 1);
+              }
+          } else {
+              filter_exact = false;
+              if (window.localStorage) {
+                  localStorage.removeItem('log_filter_exact_{{module}}_{{scope}}');
+              }
+          }
+          let select = $("#severity_filter");
+
+          // set new select value to current value or highest value of multiselect
+          let new_val = Array.isArray(select.val()) ? select.val().pop() : select.val();
+
+          // store user choice
+          localStorage.setItem('log_severity_{{module}}_{{scope}}', new_val);
+
+          switch_mode(new_val);
+      });
+
       updateServiceControlUI('{{service}}');
 
       // move filter into action header
       $("#severity_filter_container").detach().prependTo('#grid-log-header > .row > .actionBar > .actions');
     });
+
+    function switch_mode(value) {
+        let select = $("#severity_filter");
+
+        // switch select mode and destroy selectpicker
+        select.prop("multiple", !select.prop("multiple"));
+        select.selectpicker('destroy');
+
+        // remove title option. bug in bs-select. fixed in v1.13.18 https://github.com/snapappointments/bootstrap-select/issues/2491
+        select.find('option.bs-title-option').remove();
+
+        select.selectpicker();
+        select.val(value);
+        // fetch data
+        select.change();
+    }
 </script>
 
 <div class="content-box">
     <div class="content-box-main">
         <div class="table-responsive">
+            <table class="table table-striped table-condensed">
+                <tbody>
+                    <tr>
+                        <td style="text-align:left">
+                            <a href="#"><i class="fa fa-toggle-off text-danger" id="exact_severity"></i></a><small> {{ lang._('exact severity selection') }}</small>
+                            <div class="hidden" data-for="help_for_exact_severity">
+                                <small>{{ lang._('Filter by exact severity level(s) specified. When disabled, include messages with a lower level (more severe).') }}</small>
+                            </div>
+                        </td>
+                        <td style="text-align:right">
+                            <small>full help</small> <a href="#"><i class="fa fa-toggle-off text-danger" id="show_all_help"></i></a>
+                        </td>
+                    </tr>
+                </tbody>
+            </table>
             <div  class="col-sm-12">
                 <div class="hidden">
                     <!-- filter per type container -->
                     <div id="severity_filter_container" class="btn-group">
-                        <select id="severity_filter"  data-title="{{ lang._('Maximum severity level') }}" class="selectpicker" data-width="200px">
+                        <select id="severity_filter"  data-title="{{ lang._('Severity') }}" class="selectpicker" data-width="200px">
                             <option value="Emergency">{{ lang._('Emergency') }}</option>
                             <option value="Alert">{{ lang._('Alert') }}</option>
                             <option value="Critical">{{ lang._('Critical') }}</option>
--- a/src/opnsense/mvc/app/views/OPNsense/Diagnostics/log.volt
+++ b/src/opnsense/mvc/app/views/OPNsense/Diagnostics/log.volt
@@ -176,6 +176,8 @@
                   localStorage.setItem('log_severity_{{module}}_{{scope}}', new_val);
               }
               switch_mode(new_val);
+              // keep it open
+              select.selectpicker('toggle');
           });
 
           select.val(value);
--- a/src/opnsense/mvc/app/views/OPNsense/Diagnostics/log.volt
+++ b/src/opnsense/mvc/app/views/OPNsense/Diagnostics/log.volt
@@ -26,6 +26,11 @@
 
 <script>
     $( document ).ready(function() {
+      // map available severity values to array
+      severities = $('#severity_filter option').map(function(){
+          return (this.value ? this.value : null);
+      }).get();
+
       if (window.localStorage && localStorage.getItem('log_max_severity_{{module}}_{{scope}}')) {
           $('#severity_filter').val(localStorage.getItem('log_max_severity_{{module}}_{{scope}}')).change();
       }
@@ -47,7 +52,6 @@
               },
               requestHandler: function(request){
                   if ( $('#severity_filter').val().length > 0) {
-                      let severities = ["Emergency", "Alert", "Critical", "Error", "Warning", "Notice", "Informational", "Debug"];
                       let selectedSeverity = $('#severity_filter').val();
                       request['severity'] = severities.slice(0,severities.indexOf(selectedSeverity) + 1);
                   }
--- a/src/opnsense/mvc/app/views/OPNsense/Diagnostics/log.volt
+++ b/src/opnsense/mvc/app/views/OPNsense/Diagnostics/log.volt
@@ -28,6 +28,8 @@
     $( document ).ready(function() {
       var filter_exact = false;
       let s_filter_val = "Warning";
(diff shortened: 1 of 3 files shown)


src/opnsense/mvc/app/views/OPNsense/Diagnostics/log.volt
```diff
@@ -44,14 +47,19 @@
               },
               requestHandler: function(request){
                   if ( $('#severity_filter').val().length > 0) {
-                      request['severity'] = $('#severity_filter').val();
+                      let severities = ["Emergency", "Alert", "Critical", "Error", "Warning", "Notice", "Informational", "Debug"];
+                      let selectedSeverity = $('#severity_filter').val();
+                      request['severity'] = severities.slice(0,severities.indexOf(selectedSeverity) + 1);
```
AdSchellevis, line 52: @kulikov-a not sure if I like this part, is there a specific reason to duplicate the levels here?

src/opnsense/mvc/app/views/OPNsense/Diagnostics/log.volt
```diff
@@ -120,12 +128,12 @@
                 <div class="hidden">
                     <!-- filter per type container -->
                     <div id="severity_filter_container" class="btn-group">
-                        <select id="severity_filter"  data-title="{{ lang._('Severity') }}" class="selectpicker" multiple="multiple" data-width="200px">
+                        <select id="severity_filter"  data-title="{{ lang._('Maximum severity level') }}" class="selectpicker" data-width="200px">
                             <option value="Emergency">{{ lang._('Emergency') }}</option>
                             <option value="Alert">{{ lang._('Alert') }}</option>
                             <option value="Critical">{{ lang._('Critical') }}</option>
                             <option value="Error">{{ lang._('Error') }}</option>
```
AdSchellevis, line 135: @kulikov-a you probably want to select error by default as well (maybe Critical and emergency too)

src/opnsense/mvc/app/views/OPNsense/Diagnostics/log.volt
kulikov-a: @AdSchellevis Hi! me neither) i can map select options values to array? (one time at page load).
`var severities = $('#severity_filter option').map(function(){  return this.value; });`
i need array here to make `request['severity']` array from one selected "max severity" value

src/opnsense/mvc/app/views/OPNsense/Diagnostics/log.volt
kulikov-a: but we only limit the maximum level? all levels below the specified (emerg..warning) one will be included in the request during `requestHandler` processing?

AdSchellevis: @kulikov-a oops, missed the most important part in your commit. so that leaves the question if we do want to be able to filter specific levels or a set a cut off maximum. let me think about this for a while

kulikov-a: sure! thanks (I don't think you missed it, but I poorly expressed the idea)

AdSchellevis: @kulikov-a I thought about this for a while, but I think we better keep the filters as is. If for example you're looking for a low priority message you would otherwise be forced to crawl through all higher level items as well. Which is ok for quite some use-cases, but could be annoying when looking for specific messages.

kulikov-a: @AdSchellevis I'm sorry, could you please explain in a little more detail? Perhaps you meant the opposite: will have to see messages with **less** severity? that is, if, for example, the administrator wants to see only Warnings messages and will also see Errors,  Alarms etc? I agree, this is possible.
Simply (and from my experience as an administrator) it seemed to me that in most cases (especially since we cannot know in advance the severity level of the required message) working with logs comes down to switching two modes: normal and debug (normal allows us to concentrate on messages about abnormal behavior , debug - get more information about what is happening in the system). additional filtering in this case can be performed by searching.
but perhaps you are right and I still need to compare to understand which of the methods is more convenient (now at 21.7.6 I use my patches. It's convenient for me)). thanks!

AdSchellevis: @kulikov-a sure, if I'm looking for Informational messages, I would automatically receive Notices, Warnings and Errors too, which is a bit inconvenient. Making sure the default window opens with anything higher than warning is obviously fine, it just feels like a bit of a loss to not being able to filter on an exclusive level anymore.

kulikov-a: @AdSchellevis got it, thanks! if you would be ok with a mode switch checkbox (individual choice <-> maximum level), just let me know) I'll try to do that

AdSchellevis: @kulikov-a that would be defendable indeed, wouldn't mind using maximum level by default in that case.

kulikov-a: great! I'll try )

kulikov-a:
@AdSchellevis Hi! seems to work. Could you look at it when you have time?)

AdSchellevis:
@kulikov-a will do, thanks!

8191:
@kulikov-a This PR looks really nice and useful, as one is very rarely interested in specific levels, but usually only ranges!

I'd like to discuss the "exact severity filter" option a bit further:
I find the toggle box itself a bit misplaced in the UI (top left of the page). I'd recommend putting it below the actual filter elements, as these elements belong contextually to the switch. The help is not needed from my perspective, the name is quite self-explanatory.

Anyway, I cannot imaging any use-case where I'd like to see a precise combination of non-incremental severity levels. The only situation I can imagine is, that a specific like is spamming the log and I'd like to ignore that line for a moment. But even in such a situation, I very likely would loose other interesting logs having the same severity as the spammer... Maybe a negative text-filter would therefore be more useful than a precise selection of severities?

AdSchellevis:
@8191 Ok, maybe I'm complicating it more than needed. I can think of some use-cases when it would be practical to filter only "Informational" when there are a lot of "Notices" as well. @kulikov-a previous version only used max severity, which in hindsight might be better to start with.

kulikov-a:
apologize for the delay (raid on old adaptec drove me nuts all day).
actually, when I started to prepare an update for this request, I finally seem to understand @AdSchellevis . The exact choice of the severity is sometimes useful (for example, when it is necessary to export strings of a certain severity or, for convenience, necessary to remove especially noisy severities).
@8191 you mean to do something like?
log2
@AdSchellevis what do you think? which option is better (I also think the option with a single switch is more compact. help text can be placed in tooltip)?
first variant:
log1

AdSchellevis:
@kulikov-a The new variant looks pretty nice, doesn't draw more attention then needed, nice :) A tooltip might work indeed, but only explaining the use in the documentation is also an option if it involves more text.

8191:
@kulikov-a Thanks for taking my suggestion into account to move the check "box" to the right.
I don't want to be picky, but from a usability point of view, I believe having it exactly above the severity selection would be even better... :hand_over_mouth:

kulikov-a:
@AdSchellevis thanks! I'll try to come up with a concise tooltip and update the code)
@8191 I would try, but I couldn't come up with an organic look for this. could you give an example please?

kulikov-a:
the most compact that I could come up with:

8191:
@kulikov-a sorry, I didn't have time yet to look at this, but your proposal looks good!

AdSchellevis:
@kulikov-a you're very creative :) To be honest I don't have a strong preference on the subject, the small toggle on the left is likely a bit cleaner, but I can live with both.

kulikov-a:
@AdSchellevis  thanks)). updated the code

8191:
@kulikov-a Your proposal looks good to me! :-)

Just for curiosity I've tried something else:
```javascript
select.selectpicker({ header: '<a href="#"><i class="fa fa-toggle-off text-danger" id="exact_severity" title="{{ lang._('Toggle between range (max level) and exact severity filter') }}"></i>Multiselect</a>'});
```

image

kulikov-a:
@8191 looks fun  but i think we will need to (re)bind 'click' each time after bs-select init? not too complex?

8191:
@kulikov-a You're right - maybe it's too much hassle to bring this is.
Your proposal is fine anyway, the only thing which I don't like is that from first sight it looks like to turn on/off the complete filter element. What about changing the icon from `fa-toggle-on`/`fa-toggle-off` to e.g. `fa-bullseye`/`dot-circle`? Like "targeted" or "selected"...

https://user-images.githubusercontent.com/5854959/146633946-57c1d07a-b92e-4cdb-aefd-be0794981e6e.mp4

kulikov-a:
@8191 in this case, I would try a `fa-crosshairs` with color switching :wink:
max_s
but I thought to stick with the general gui style 
I agree to any option that suits @AdSchellevis )

8191:
Yes, we should stick to the general GUI style, but I think there is nothing similar existing already. The "switch" symbol is usually used to switch something on/off. But here we are switching _between_ two functions.

Your proposal is also fine 🙂

AdSchellevis:
I don't have a very strong preference, although Manuel's version is clearer at a first sight.
image

If the current code is functional, I don't mind reviewing and pulling this in and change the layout a bit later by the way.

8191:
Yeah, I'm in favor of the two step approach. We can deal with that detail in a separate PR.

kulikov-a:
I ventured to update the code based on Manuel's idea (in the end, it turned out to be even somewhat shorter). Maybe it will turn out not to return to the question with a separate request )

src/opnsense/mvc/app/views/OPNsense/Diagnostics/log.volt
```diff
@@ -99,18 +129,61 @@
...
+                  if (filter_exact) {
+                      localStorage.setItem('log_filter_exact_{{module}}_{{scope}}', 1);
+                  } else {
+                      localStorage.removeItem('log_filter_exact_{{module}}_{{scope}}');
+                  }
+                  // store user choice
+                  localStorage.setItem('log_severity_{{module}}_{{scope}}', new_val);
+              }
+              switch_mode(new_val);
+          });
```
8191, line 178:
```suggestion
          $("#exact_severity").on("click", function(event) {
              event.stopPropagation();
              filter_exact = !filter_exact;
              let select = $("#severity_filter");

              // set new select value to current value or highest value of multiselect
              let new_val = Array.isArray(select.val()) ? select.val().pop() : select.val();

              if (window.localStorage) {
                  if (filter_exact) {
                      localStorage.setItem('log_filter_exact_{{module}}_{{scope}}', 1);
                  } else {
                      localStorage.removeItem('log_filter_exact_{{module}}_{{scope}}');
                  }
                  // store user choice
                  localStorage.setItem('log_severity_{{module}}_{{scope}}', new_val);
              }
              switch_mode(new_val);
              select.addClass('open');
              select.selectpicker('refresh');
          });
```

`stopPropagation()` allows to to open the selectpicker after recreating it. With `addClass('open')` and `selectpicker('refresh')` you actually open it.

Note: There was also a semicolon missing in line 162.

src/opnsense/mvc/app/views/OPNsense/Diagnostics/log.volt
```diff
@@ -99,18 +129,61 @@
...
+
+          // remove title option. bug in bs-select. fixed in v1.13.18 https://github.com/snapappointments/bootstrap-select/issues/2491
+          select.find('option.bs-title-option').remove();
+
+          let header_val = filter_exact ? m_header : s_header;
+          select.selectpicker({ header: header_val });
+
+          // attach event handler each time header created
+          $("#exact_severity").on("click", function() {
+              filter_exact = !filter_exact
```
8191, line 162:
```suggestion
          $("#exact_severity").on("click", function(event) {
              event.stopPropagation();
              filter_exact = !filter_exact;
```

`stopPropagation()` allows to to open the selectpicker after recreating it. See also comment below at line 177.

_Note: There was also a semicolon missing in line 162._

src/opnsense/mvc/app/views/OPNsense/Diagnostics/log.volt
```diff
@@ -99,18 +129,61 @@
...
+              if (window.localStorage) {
+                  if (filter_exact) {
+                      localStorage.setItem('log_filter_exact_{{module}}_{{scope}}', 1);
+                  } else {
+                      localStorage.removeItem('log_filter_exact_{{module}}_{{scope}}');
+                  }
+                  // store user choice
+                  localStorage.setItem('log_severity_{{module}}_{{scope}}', new_val);
+              }
+              switch_mode(new_val);
```
8191, line 177:
```suggestion
              switch_mode(new_val);
              select.addClass('open');
              select.selectpicker('refresh');              
```

With `addClass('open')` and `selectpicker('refresh')` you actually the selectpicker after recreating it.

8191:
@kulikov-a Looks great! 👍🏻 

One suggestion: keep the selectipicker open after clicking the toggle button. It's a bit tricky, but after some struggle I've accomplished by `addClass('open')` and a `selectpicker('refresh')`. This also requires to stop propagating the click event. - See code comments inline.

src/opnsense/mvc/app/views/OPNsense/Diagnostics/log.volt
kulikov-a: imho 'toggle' method do the same )

kulikov-a: keep select open is great idea. thanks!

src/opnsense/mvc/app/views/OPNsense/Diagnostics/log.volt
```diff
@@ -99,18 +129,64 @@
...
+          });
+
+          select.val(value);
+          // fetch data
+          select.change();
+          // keep it open
+          select.selectpicker('toggle');
```
8191, line 185: This should go into the click event handler, otherwise the menu opens on each page load.

kulikov-a: shoot..wrong place pasted. thanks!

src/opnsense/mvc/app/views/OPNsense/Diagnostics/log.volt
```diff
@@ -99,18 +129,64 @@
...
+                      localStorage.setItem('log_filter_exact_{{module}}_{{scope}}', 1);
+                  } else {
+                      localStorage.removeItem('log_filter_exact_{{module}}_{{scope}}');
+                  }
+                  // store user choice
+                  localStorage.setItem('log_severity_{{module}}_{{scope}}', new_val);
+              }
+              switch_mode(new_val);
+              // keep it open
+              select.selectpicker('toggle');
```
8191, line 180: For some reason _toggle_ does not work. The menu is still opened after each page reload. I've tested it again with _open_ and _refresh_, there it works.😮
I'm not too deep into bootstrap-select to tell why... 

```suggestion
              select.addClass('open');
              select.selectpicker('refresh');
```

src/opnsense/mvc/app/views/OPNsense/Diagnostics/log.volt
8191: For some reason _toggle_ does not work. The menu is still opened after each page reload. I've tested it again with open and refresh, there it works.

src/opnsense/mvc/app/views/OPNsense/Diagnostics/log.volt
kulikov-a: hm. i tested it (except pasted then in wrong place).  I'll try to find out what's going on, sorry

8191: hm... @kulikov-a I've just tested it again and now `toggle` works.... 😄 Maybe my browser cache was somehow not refreshed... 😞 
Sorry for the confusion...

kulikov-a: upd. take whole .volt form pr and tested on chrome, opera, edge, ff and ie: works well everywhere. can you please try to replace whole .volt? (maybe there is something left from previous commits? cannot understand what can trigger a `click` on page load)

kulikov-a: @8191 hooray! so we're ready to show it to Ad?

8191: @kulikov-a yeah, looks good from my side as well! 
Thanks!

kulikov-a:
@AdSchellevis looks like we are ready for review. whenever you have time :smile:

AdSchellevis:
@kulikov-a @8191  works like a charm! thanks!

kulikov-a:
@AdSchellevis  thanks!!
@8191 thanks again for code review and toggle choice! pleasure to participate in this way

PR #5378, opened 2021-11-28
Firewall / Diagnostics / Sessions - typo

kulikov-a:
Hi!
I apologize, did not notice earlier and did not add to the previous pr :worried: .
looks like a typo on the pf_top page. as a result, all sessions are marked as incoming (api returns direction as a `dir`. formatter uses `column.id` to get value ).

thanks!


AdSchellevis:
oO, I didn't notice this either, thanks again!

kulikov-a:
thanks!)

Issue #5382, opened 2021-11-29
Suricata: error installing ids rules (


f45tb00t:
**Important notices**

Before you add a new report, we ask you kindly to acknowledge the following:

- [ ] I have read the contributing guide lines at https://github.com/opnsense/core/blob/master/CONTRIBUTING.md
- [ ] I am convinced that my issue is new after having checked both open and closed issues at https://github.com/opnsense/core/issues?q=is%3Aissue

**Describe the bug**

A clear and concise description of what the bug is, including last known working version (if any).

*Tip*: to validate your setup was working with the previous version, use opnsense-revert (https://docs.opnsense.org/manual/opnsense_tools.html#opnsense-revert)

**To Reproduce**

**Expected behavior**

**Describe alternatives you considered**

A clear and concise description of any alternative solutions or workaround you considered.

**Screenshots**

**Relevant log files**

If applicable, information from log files supporting your claim.

**Additional context**

**Environment**

Software version used and hardware type if relevant, e.g.:

OPNsense 19.1.1 (amd64, OpenSSL).
Intel® Xeon™ E3-1225V5 3.3Ghz Quad Core
Network Intel® I210-AT

f45tb00t:
Note: The interesting part is, I've checked the file status while creation
38596 8389 -rw-r-----  1 root  wheel    30M Nov 29 14:53 rules.sqlite
When it reaches the 30M the error pops up. 

Could it be, that the timeout is set too low somehow? Like the error is shown when applying the policy takes too long?

kulikov-a:
hi
any errors if you run `/usr/local/opnsense/scripts/suricata/installRules.py` manually?

f45tb00t:
Hello @kulikov-a,

no errors at all shown via cli when executing the script manually.

kulikov-a:
@f45tb00t 
sorry for the delay.
yes, the `configdRun()` function contains timouts control (2 min by default). but I had to reduce it to 1 minute for a configuration with snort rules enabled and a 54MB cahe db in order to reproduce your error. and this is on a VM with one Xeon(R) CPU E5-2620 core configured and 4GB of RAM (and with unbound with all blacklists allowed).
so imho the reason is something else. resources are already depleted. Policies applying task is not so resource intensive.

f45tb00t:
Hello,

just let me know what I can provide as logs to track down the issue when you have time. Regardless to the error, it seems the policy is applied. But from where the error comes, I have absolutely no clue.

[...]

PR #5383, opened 2021-11-30
Backend.php: syslog-ng migration

kulikov-a:
Hi!
ref. https://github.com/opnsense/core/issues/5382
it seems that error messages from the `configdRun()` function (https://github.com/opnsense/core/blob/b6ffe6fa17b8ec7838128027b9955697330bacab/src/opnsense/mvc/app/library/OPNsense/Core/Backend.php#L81) do not get into the `syslog-ng` backend log.
changes worked for me:
- `$ident = 'configd.py'` for `f_local_configd` syslog-ng filter matching
- remove `[%date%][%type%] ` from default phalcon message format to avoid duplication in log line

thanks!


AdSchellevis:
@kulikov-a thanks! seems to fix the issue indeed

PR #5395, opened 2021-12-08
IDPS rule-updater: change syslog severity level

kulikov-a:
Hi!
since syslog-ng grabs only `level(notice..emerg)` for system log maybe it is worth changing the severity level so that messages about skipped downloads go to the log?
thanks!


AdSchellevis:
@kulikov-a thanks!

kulikov-a:
thanks!)

PR #5400, opened 2021-12-10
Allow to configure SSH setting PubkeyAcceptedAlgorithms via GUI

8191:
As OpenSSH 8.8 disables RSA signatures using the SHA-1 hash algorithm
by default some older SSH clients might not be able to connect to
OPNsense anymore. Therefore, it might be needed to manually modify the
PubkeyAucceptedAlgorithms sshd config option.

### DIFF
--- a/src/etc/inc/plugins.inc.d/openssh.inc
+++ b/src/etc/inc/plugins.inc.d/openssh.inc
@@ -87,6 +87,21 @@ function openssh_stop()
     mwexecf('/bin/pkill -af %s', '/usr/local/sbin/sshd', true);
 }
 
+function openssh_check_querylist($list, $config) {
+    $supported = json_decode(configd_run("openssh query {$list}"));
+    $selected = [];
+    foreach (explode(',', $config) as $elem) {
+        if (in_array($elem, $supported)) {
+            $selected[] = $elem;
+        }
+    }
+
+    if (count($selected) == 0) {
+        syslog(LOG_WARNING, "OpenSSH: No configured {$list} ({$config}) supported - using defaults.");
+    }
+    return implode(',', $selected);
+}
+
 function openssh_configure_do($verbose = false, $interface = '')
 {
     global $config;
@@ -192,20 +207,36 @@ function openssh_configure_do($verbose = false, $interface = '')
         $sshconf .= "PasswordAuthentication no\n";
     }
     if (!empty($sshcfg['kex'])) {
-        $sshconf .= "KexAlgorithms {$sshcfg['kex']}\n";
+        $kexs = openssh_check_querylist('kex', $sshcfg['kex']);
+        if (strlen($kexs) > 0) {
+            $sshconf .= "KexAlgorithms {$kexs}\n";
+        }
     }
     if (!empty($sshcfg['ciphers'])) {
-        $sshconf .= "Ciphers {$sshcfg['ciphers']}\n";
+        $ciphers = openssh_check_querylist('cipher', $sshcfg['ciphers']);
+        if (strlen($ciphers) > 0) {
+            $sshconf .= "Ciphers {$ciphers}\n";
+        }
     }
     if (!empty($sshcfg['macs'])) {
-        $sshconf .= "MACs {$sshcfg['macs']}\n";
+        $macs = openssh_check_querylist('mac', $sshcfg['macs']);
+        if (strlen($macs) > 0) {
+            $sshconf .= "MACs {$macs}\n";
+        }
     }
     if (!empty($sshcfg['keys'])) {
-        $sshconf .= "HostKeyAlgorithms {$sshcfg['keys']}\n";
+        $keys = openssh_check_querylist('key', $sshcfg['keys']);
+        if (strlen($keys) > 0) {
+            $sshconf .= "HostKeyAlgorithms {$keys}\n";
+        }
     }
     if (!empty($sshcfg['keysig'])) {
-        $sshconf .= "PubkeyAcceptedAlgorithms {$sshcfg['keysig']}\n";
+        $keysigs = openssh_check_querylist('key-sig', $sshcfg['keysig']);
+        if (strlen($keysigs) > 0) {
+            $sshconf .= "PubkeyAcceptedAlgorithms {$keysigs}\n";
+        }
     }
+
     foreach ($keys_all as $name) {
         $file = "/conf/sshd/{$name}";
         if (!file_exists($file)) {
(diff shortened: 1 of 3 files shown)


fichtner:
Older clients should really be updated since RSA can still be used in safe forms with the new server. The clients that can't must be updated.. I think this is unnecessary, but it's up to @AdSchellevis to merge.

AdSchellevis:
let mature this for a while and see if more people will ask for it, to me this feels a bit unnecessary too.

8191:
> Older clients should really be updated since RSA can still be used in safe forms with the new server.

From a security and a policy point of view I totally agree here. But from a real-life point of view, I believe we need to admit that things work differently. :-(

I faced this very situation yesterday as I recognized that one of my development tools ships psftp in version 0.74 which was actually released last year (2020-06-27). The tool is hard-bundled and cannot be changed, so I rely on the upstream project to update PuTTY in there. Support for newer signature methods such as rsa-sha2-256 and rsa-sha2-512 was added to PuTTY in version 0.75 (2021-05-08). That means - as I see it now - that after less than one year of transition period OPNsense is already retiring a quite common SSH client which is not up to date.

I fully agree, that the default setting should not be changed (i.e. ssh-rsa _should be disallowed_ per default), but one should be able to intentionally activate insecure algorithms.

fichtner:
> That means - as I see it now - that after less than one year of transition period OPNsense is already retiring a quite common SSH client which is not up to date.

We do follow the best practices of the bundled software. This isn't our fight and I don't want to start release engineering things like OpenSSH just for portability while also undermining the authors intent for the software's security.

That being said your patch is the best middle ground here for the problem you described. I think we can agree on that.

If we could hide these overrides under an advanced knob that would be my preference as this tries to bridge a gap that is limited in time and I don't want new installs making all these decisions when OpenSSH comes with a sensible default for most users even in the newest releases.

What do you think?


Cheers,
Franco

8191:
> I don't want to start release engineering things like OpenSSH just for portability while also undermining the authors intent for the software's security.

I can understand that at one hand, but at the other hand OPNsense provides a frontend for these tools (such as OpenSSH) and the authors of those tools, such as OpenSSH, make their decisions without the OPNsense frontend in mind. The author's intent was maybe (I'm speculating) to deprecate the RSA-1 algorithm, having in mind that there is a config option enabling it. - Well here we are building that option.

I totally agree with your idea not to provide too many settings on the UI which are only needed by a fraction of users. I've modified this PR to hide all cryptographic settings (also the existing ones for _Key exchange algorithms_, _Ciphers_, _MACs_ and _Host key algorithms_. I believe 99% of the installations should be fine with the provided OpenSSL defaults.

NOYB:
ONPsense 21.7.5 broke several SSH tools.  Short list just to name a few: PuTTY 0.75, Wireshark 3.4.10, GIT Extensions 3.5.4 (PuTTY 0.75), and WinSCP 5.19.4, 

PuTTY 0.76 works
Wireshark 3.6.0 works
GIT Extensions 3.5.4 configured to use PuTTY 0.76 works
WinSCP 5.20 release with PuTTY 0.76 is due any day now and should work.

Trouble with the wait and see how many people complain approach is that for every person who complains there are many more impacted who don't.  "Where there's smoke there's fire."

AdSchellevis:
People will always have to change something (clients, toggles on our end, ..), waiting a bit to see how useful a feature is (that lowers security) usually isn't a very bad idea. It's quite easy to stuff toggles in, removing them later to keep code maintainable is often difficult and comes with a lot of noise. Security is also about keeping code maintainable and making it more difficult to use non-safe options.

Again this doesn't mean we won't add it, just that we wait a bit to see how much it's actually needed.

fichtner:
Well, we heard the side where some utility we don't even see doesn't work with some other utility we happen to provide. And here is the flip side of providing these overrides: https://forum.opnsense.org/index.php?topic=25574.0

8191:
@fichtner Well, even that case is related to these crypto overrides, this basically might happen with each single setting. We'd need to check each single setting after an update of an upstream component where OPNsense generates configurations for.

Before this PR gets merged this learning should be considered and the configd template should take into account that crypto settings might disappear.

fichtner:
This doesn't scale and provides no fallback for options that have gone away even if we can catch this behaviour. Complaints will always be complaints no matter how much effort is spent here.

fichtner:
@8191 I do appreciate your motivation to get the problematic bits out of the way. :) Eventually such changes need a place in the actual template/backend framework to happen automatically for all services, but this is more of a legacy vs. MVC service handling issue and the complexity for OpenSSH is low enough to keep it on the legacy end of the table.

Lockouts can still occur due to settings being removed upstream so the remaining question is how to present this in the GUI? We don't have to solve it now, but it would be interesting what you think about it?

8191:
I actually thought about migrating the SSH config to a configd template, but refrained from it to not blow up this rather simple PR. ;-) Yes, I agree, this is nothing specific for OpenSSH it applies to all 3rd party services.

The presentation on the GUI is a valid point. With my proposed solution startup does not block, but the user never recognizes that the actual SSH configuration differs from the intended configuration. Maybe we could show an exclamation mark next to the setting warning the user to revise the setting. I.e. as long as the config contains values not supported by the system, a warning appears next to the setting in the GUI.
@fichtner Do you think that's something beneficial here? Or just a fancy, but unneeded function?

For the situation that a complete config option vanishes maybe a solution might be to remove the setting itself (e.g. checkbox/dropdown), but leave the label and add an explanation how the removed functionality might be compensated otherwise - at least for some releases.

fichtner:
> ONPsense 21.7.5 broke several SSH tools

Going back to this I think to be fair this should have read "several SSH tools when using RSA keys". The administrator should have issued and authorized complementary ECDSA or ED25519 keys in the meantime. Personally, I did create ED25519 keys for eventual migration away from RSA in March 2020 for my equipment and it works.

fichtner:
@8191 In principle I agree with a markup for bad values, but in the meantime I would not want to do this for static PHP pages because in the end we would have to it twice. How about in the meantime we just log the values that no longer work? The next time the admin opens the page and saves it the values will be gone anyway.

fichtner:
PS: I think we should get this into 22.1... it looks like a good improvement

8191:
@fichtner I've added the log line as a workaround for now. Maybe we can migrate the SSH configuration to MVC for the next major release (22.7?)...

image

L1ghtn1ng:
Just to throw my 2 cents in here for SSH clients, if people need new ssh clients you can look at https://mobaxterm.mobatek.net/ or also https://www.termius.com/ if you need to do sftp you can look at https://cyberduck.io/ moba can do it built in. The reason I say we do not include this is this will be wide spread issue next year which in turn means software will get updated when you have Ubuntu22.04 the next version of fedora and so forth. This will be an issue with arch systems etc already. So I do not see the point in adding this myself. Also if you are on a newer version of win10 you have ssh built in so you can also use that. Hope that helps people

fichtner:
merged, thanks!

PR #5405, opened 2021-12-14
util: add support for terabytes, and petabytes to format_bytes

agh1467:
This just expands the supported range of the formatting to support tera-, and petabytes.

It utilizes the same method as megabytes, and gigabytes, only adding an extra 1024 for the larger sizes.

* This formatted number is used by the Interface Statistics widget,
  and makes the columns go wider than necessary when going above a
  terabyte of data. Add petabytes for good measure.

### DIFF
--- a/src/etc/inc/util.inc
+++ b/src/etc/inc/util.inc
@@ -1106,7 +1106,11 @@ function is_private_ip($iptocheck)
 
 function format_bytes($bytes)
 {
-    if ($bytes >= (1024 * 1024 * 1024)) {
+    if ($bytes >= (1024 * 1024 * 1024 * 1024 * 1024)) {
+        return sprintf("%.2f PB", $bytes / (1024 * 1024 * 1024 * 1024 * 1024));
+    } elseif ($bytes >= (1024 * 1024 * 1024 * 1024)) {
+        return sprintf("%.2f TB", $bytes / (1024 * 1024 * 1024 * 1024));
+    } elseif ($bytes >= (1024 * 1024 * 1024)) {
         return sprintf("%.2f GB", $bytes / (1024 * 1024 * 1024));
     } elseif ($bytes >= 1024 * 1024) {
         return sprintf("%.2f MB", $bytes / (1024 * 1024));
--- a/src/etc/inc/util.inc
+++ b/src/etc/inc/util.inc
@@ -1106,14 +1106,14 @@ function is_private_ip($iptocheck)
 
 function format_bytes($bytes)
 {
-    if ($bytes >= (1024 * 1024 * 1024 * 1024 * 1024)) {
-        return sprintf("%.2f PB", $bytes / (1024 * 1024 * 1024 * 1024 * 1024));
-    } elseif ($bytes >= (1024 * 1024 * 1024 * 1024)) {
-        return sprintf("%.2f TB", $bytes / (1024 * 1024 * 1024 * 1024));
-    } elseif ($bytes >= (1024 * 1024 * 1024)) {
-        return sprintf("%.2f GB", $bytes / (1024 * 1024 * 1024));
-    } elseif ($bytes >= 1024 * 1024) {
-        return sprintf("%.2f MB", $bytes / (1024 * 1024));
+    if ($bytes >= 1024 ** 5) {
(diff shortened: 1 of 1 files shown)


fichtner:
@agh1467 thanks, there's also a similar function in src/opnsense/mvc/app/controllers/OPNsense/Core/Api/FirmwareController.php called formatBytes() ... can you update this one too? The latter mimics the output of the FreeBSD package manager so the output is slightly different.

agh1467:
@fichtner Sure. I'll take a look at it.

agh1467:
@fichtner I created #5417 which addresses the `formatBytes()`.

There is also another function I noticed in that PHP. 

https://github.com/opnsense/core/blob/a984542d37287b6df1001cfc7171430e9c4464d9/src/opnsense/mvc/app/controllers/OPNsense/Core/Api/FirmwareController.php#L132-L150

My guess is that it's iterating through a list of download sizes. I left it alone since I thought it would be unlikely that any individual download size would be larger than gigabyte size.

fichtner:
@agh1467 thanks, yes, there is some glue around the FreeBSD package manager to translate some of its values back to "digestible" values before displaying it again and I agree it's not needed in this particular snipped... for formatBytes(),  however, it may be that this one becomes the authoritative function in the MVC one day

src/etc/inc/util.inc
```diff
@@ -1107,7 +1107,11 @@ function is_private_ip($iptocheck)
 
 function format_bytes($bytes)
 {
-    if ($bytes >= (1024 * 1024 * 1024)) {
+    if ($bytes >= (1024 * 1024 * 1024 * 1024 * 1024)) {
```
fichtner, line 1110: I was thinking just for showing off, err, reducing eye fatigue...should we use "**" instead?

```suggestion
    if ($bytes >= (1024 ** 4)) {
```

see https://www.php.net/manual/en/language.operators.arithmetic.php

agh1467:
It did seems like the long way around, but worked so I didn't change it. I'll update them all to use the arithmetic approach instead as well as add petabytes.

agh1467:
Something else I noticed while taking a look at this was the difference in style between the util function, and the one in the controller API. This one has 2 decimal precision while the other is only one, this one has a space between the number and the bytes designation, this one uses "KB" versus "KiB", and the final one writes out the word "bytes" instead of "B." None of that really matters in the end, but it was just something I noticed.

Anyway, I wrote a quick script to test this function to validate that the output works as expected:
```php
<?php

require_once('/usr/core/src/etc/inc/util.inc');

$tests = array(
    "42705383466540728" => "37.93 PB",
    "1125899906842624" =>   "1.00 PB",
    "723918942673774" =>  "658.40 TB",
    "58274116272128" =>    "53.00 TB",
    "1099511627776" =>      "1.00 TB",
    "72391673774" =>       "67.42 GB",
    "1073741824" =>         "1.00 GB",
    "37633392" =>          "35.89 MB",
    "1048576" =>            "1.00 MB",
    "1024" =>               "1 KB",
    "424" =>              "424 bytes",
    "0" =>                  "0 bytes"
);
print ("Testing format_bytes()\n");
foreach ($tests as $input => $output) {
    printf ("Testing input: %s, expecting output: %s", $input, $output);
    if (format_bytes($input) == $output) {
        print ("  ---PASS\n");
    } else {
        print ("  ---FAIL\n");
    }
}


```

fichtner:
Merged, thanks!

fichtner:
> Something else I noticed while taking a look at this was the difference in style between the util function, and the one in the controller API. This one has 2 decimal precision while the other is only one, this one has a space between the number and the bytes designation, this one uses "KB" versus "KiB", and the final one writes out the word "bytes" instead of "B." None of that really matters in the end, but it was just something I noticed.

The "i" syntax is what pkg tool emits, but the counters in the GUI have always used the other style. Sone people have voiced that they don't like the "i" but on the other hand to avoid excessive conversions between pkg output and the firmware GUI the same style was used (and that includes the "bytes" suffix).


Cheers,
Franco

fichtner:
PS: Same goes for the "%.2f" vs. "%.1F" BTW

Issue #4691, opened 2021-02-10
DHCPv6 server intermittently unresponsive, not responding to solicits


maurice-w:
- [x] I have read the contributing guide lines at https://github.com/opnsense/core/blob/master/CONTRIBUTING.md
- [x] I have searched the existing issues and I am convinced that mine is new (kind of).

**Describe the bug**

OPNsense 20.7 / FreeBSD 12.1 introduced a bug which causes intermittent unresponsiveness of services that listen to multicasts. For Router Advertisements, this was eventually solved by updating radvd (#4338). But the same issue still exists for the DHCPv6 server. The intention behind opening this bug report is keeping it on the agenda after #4338 has been closed.

**To Reproduce**

dhcpd6 intermittently doesn't respond to DHCPv6 solicits. A restart of the dhcpd6 service temporarily restores functionality.

1. Do a packet capture on the LAN interface, filter for DHCPv6 packets.
2. See error: DHCPv6 solicits from clients to multicast address ff02::1:2 with no response from OPNsense.
3. Restart dhcpd6.
4. The next DHCPv6 solicit results in the usual DHCPv6 handshake (solicit - advertise - request - reply).

**Additional context**

Not sure how to approach this. Port the radvd patch to dhcpd6? Fix the underlying FreeBSD issue?

**Environment**

OPNsense 21.7.a_6 (amd64, OpenSSL)

fichtner:
Hi @maurice-w,

This isn't very good news, but could be expected given the radvd adventure we have had. Basically I looked through FreeBSD code to find a fix but if it isn't on 21.1 yet there is not much else there to find for the moment.

For some reason subsequent LEAVE/JOIN sequences  are broken for a single interface and worse yet I haven't been able to write a simple POC binary that triggers this. In isolated code that works without issues?!

If it can be avoided for isc-dhcp binary by patch that would work, but patching upstream software gets messy fast. In the radvd case it was ok-ish because the pieces that broke radvd only ever lived in FreeBSD ports.

No suggestion yet, just trying to offer to get the conversation going hereby.


Cheers,
Franco

marjohn56:
Just to add my few cents, I'm seeing the same thing.

andryyy:
I saw the same thing happening and came to comment on the other thread when I stumble over this issue. :)

I installed a local dhcpy6d in my network. which is pretty straight-forward. It works!

Pinging @martinvogt-tinc 🤜🏼

fichtner:
[...]

PR #5410, opened 2021-12-17
add action to dhcpd v6 to restart dhcpd

Wireheadbe:
workaround to cope with issue #4691 - allows restarting of dhcpd ipv6 via cron


fichtner:
@Wireheadbe thanks. let me take look if we want to add other variants as well and/or rename the action for clarity

fichtner:
@Wireheadbe sorry for dropping the ball here, 80ba04f should do the trick with either empty parameter, "inet" for IPv4 or "inet6" for IPv6.

prodded via https://www.reddit.com/r/opnsense/comments/tycadv/comment/i3rh1k4/?utm_source=reddit&utm_medium=web2x&context=3

PR #5413, opened 2021-12-17
downloader.py updates

btv:
* minor syntax tweaks
* modified `is_supported` to use a tuple and return the result directly.
* changed conditional.

### DIFF
--- a/src/opnsense/scripts/suricata/lib/downloader.py
+++ b/src/opnsense/scripts/suricata/lib/downloader.py
@@ -205,7 +205,7 @@ class Downloader(object):
                         target_filename, fetch_result['filename']
                 ))
                 return None
-            if 'cached' not in fetch_result:
+            if not in fetch_result['cached']:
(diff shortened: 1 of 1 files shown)


src/opnsense/scripts/suricata/lib/downloader.py
```diff
@@ -204,16 +205,13 @@ def download(self, url, url_filename, filename, auth=None, headers=None, version
                         target_filename, fetch_result['filename']
                 ))
                 return None
-            if not fetch_result['cached']:
+            if 'cached' not in fetch_result:
```
AdSchellevis, line 208: `cached` is a boolean, in which case the above statement would return true when false.

btv: Sorry for the delay on this and thanks you for the review. I've updated the code based on your comment.

AdSchellevis: @btv no problem, we're all busy :)

AdSchellevis:
@btv thanks!

PR #5414, opened 2021-12-18
filter/lib/states: Update NAT field to display IPv6

agh1467:
On the Firewall -> Diagnostics -> States page, whenever there is an IPv6 NAT entry in the "Nat" column, the information displayed there is incomplete because it chops up the IPv6 address as though it's an IPv4 address, using the colon as a delimiter.

I added comments for reference to demonstrate what the expected data of the NAT entry might look like. I duplicated the same structure that was there and used it for IPv6, but wrapped both up in a condition dependent on characters which can be used to identify if it's IPv4, or IPv6.

I'm not familiar with the states table, and I had only my own states table to sample from to get an idea of what to expect from the NATs entries.

### DIFF
--- a/src/opnsense/scripts/filter/lib/states.py
+++ b/src/opnsense/scripts/filter/lib/states.py
@@ -139,10 +139,27 @@ def query_states(rule_label, filter_str):
                 'ipproto': parse_address(parts[2])['ipproto']
             }
             if parts[3].find('(') > -1:
-                # NAT enabled
-                record['nat_addr'] = parts[3][1:].split(':')[0]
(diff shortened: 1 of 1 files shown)


AdSchellevis:
@agh1467 can you try https://github.com/opnsense/core/commit/7629c9aa87ed5953f9f96060a3d5bf27af558280 ? should do the same. Thanks!

agh1467:
@AdSchellevis Yes. This works exactly the same. Thanks for taking a look!

AdSchellevis:
@agh1467 you're welcome, thanks for reporting and testing!

PR #5416, opened 2021-12-18
Firewall/Alias: Add striping fix

agh1467:
This is a workaround to fix the color striping of the rows in the dialog box since the hidden rows cause the visible rows to skip the expected alternating colors when the visible rows are at an odd number of rows away. It basically adds one hidden dummy row for each hidden real row so that the rows will always be even, so that the next visible row will be the alternate color.

* Add/Remove dummy rows whenever alias type changes


agh1467:
Closing, as d7a1d9fa7552933f97c9c2dc182e9a4bfee33694 addresses this issue separately.

PR #5420, opened 2021-12-19
Patch traffic shaper texts

Northguy:
Some small textual changes which clarify the use of the shaper


AdSchellevis:
@Northguy looks good, thanks!

Issue #5412, opened 2021-12-17
Automatic Float Rule Logging


NOYB:
Automatic floating rule indicates logging is disabled.  Seems to be logged anyway though.
Whatever the intended behavior it would be nice if logging indicator and behavior matched each other.
Perhaps even ability to toggle the logging same as on the other rules would be nice.
21.7.7

![Automatic Floating Rule](https://user-images.githubusercontent.com/1977521/146530927-80baefdb-aa94-44bf-a3fd-dd5d5c11aad2.jpg)
![Log](https://user-images.githubusercontent.com/1977521/146531260-52d0cbcf-4872-463d-bda1-8c624f1fba8a.jpg)
![Detailed rule info](https://user-images.githubusercontent.com/1977521/146531523-4f45da8c-ec28-4649-b5d5-ac0367b39ce5.jpg)

NOYB:
Also would be nice if source addresses that are blocked at an interface, such as WAN, would not be allowed any degree of ingress.

kulikov-a:
this is a strange log output: the packet data clearly indicates that this is rdr rule, and this is a ip4 tcp syn packet. the name of the rule is outdated, probably pf rules manipulations were carried out without closing the  live log.
Can I ask a question? This is not the first time you create a ticket and continue to delete the template during creation, this has some special meaning (it is much more convenient to read the request within the template)?

agh1467:
Definitely something strange happened. That rule should only apply to ICMP6 + IPv6 packets, and none of the src/dst/port/proto  match that. That rule from your screenshot is not configured to log, and I also see 5 of those floating rules all with the same label. It's not clear which rule it's hitting.

The floating rules on my system aren't configured to log, and I'm assuming that's default. If you were to lookup the rule by id in `pfctl -s rules` I wonder if it would match the info you have in your screenshot. You can also confirm here that the rules aren't logging as well. If a rule is logging, the word "log" will appear after the direction (in/out) in the line for that rule. These floating rules shouldn't have log in them.

NOYB:
Yes should only apply to IPv6 and there is no IPv6 on that WAN interface.  ISP doesn't support IPv6 yet.
All of the automatic float rules are indicated to be logging disabled.
My guess is that it's some other rule that's actually being hit, maybe not even a float, but by some kind of indexing bug or something that rule label is getting picked for the log entry.
There is a port forward NAT rule, that's probably what's actually being hit (rdr).  But the log entry label is from one of those automatic float rules.

[...]

PR #5424, opened 2021-12-22
Firewall/Live View: dont search rid for nat

kulikov-a:
Hi!
For "live view" part of  https://github.com/opnsense/core/issues/5412
-dont check rule number in `running_conf_descr` for nat/rdr rules
-dont show rule lookup link if `rid`  is not rid

thanks!


AdSchellevis:
@kulikov-a looks good, thanks!

kulikov-a:
thanks!!

PR #5428, opened 2021-12-26
GUI: speedup forms load a bit

kulikov-a:
Hi!
noticed when checking the observers performance
I hope not missing anything, if we are looking for element `id` in `validationErrors`, maybe it makes sense to limit `find` only to elements with `id`? (this gave me about 0.4s when loading forms with a lot of elements, like Aliases forms)
Thanks!


AdSchellevis:
@kulikov-a looks good, thanks! an `id` field should be required anyway, no need to touch the rest of the items indeed.

kulikov-a:
@AdSchellevis thanks!!

Issue #3252, opened 2019-02-26
MVC, simplify list field types


AdSchellevis:
The are multiple field types serving lists, it might be a good idea to add a new base class to abstract shared functionality.

The following field types are affected by this change:
- [x] AuthGroupField
- [x] AuthenticationServerField
- [x] CertificateField
- [x] ConfigdActionsField
- [x] CountryField
- [x] InterfaceField
- [x] JsonKeyValueStoreField
- [x] OptionField

PR #5429, opened 2021-12-26
mvc: Add BlankDesc to ModelRelationField

agh1467:
I came across this in my exploration, and thought it was weird that this field didn't support the BlankDesc option with it building a very similar thing to what OptionFields do. I tried it out and it seems to work the same way as OptionFields' BlankDesc. Needs an expert eye though. I just copied the same structures over from the `BaseListField` type and change the code to use the `internalEmptyDescription` instead of the hard coded `none`.

* Add BlankDesc options for ModelRelationField types.
* Copied from BaseListField type.

### DIFF
--- a/src/opnsense/mvc/app/models/OPNsense/Base/FieldTypes/ModelRelationField.php
+++ b/src/opnsense/mvc/app/models/OPNsense/Base/FieldTypes/ModelRelationField.php
@@ -35,37 +35,12 @@ use OPNsense\Base\Validators\CsvListValidator;
  * Class ModelRelationField defines a relation to another entity within the model, acts like a select item.
  * @package OPNsense\Base\FieldTypes
  */
-class ModelRelationField extends BaseField
+class ModelRelationField extends BaseListField
 {
-    /**
-     * @var bool marks if this is a data node or a container
-     */
-    protected $internalIsContainer = false;
-
-    /**
-     * @var bool field may contain multiple data nodes at once
-     */
-    private $internalMultiSelect = false;
-
     /**
      * @var bool field content should remain sort order
      */
     private $internalIsSorted = false;
-
-    /**
-     * @var string default validation message string
-     */
-    protected $internalValidationMessage = "option not in list";
-
-    /**
-     * @var array collected options
-     */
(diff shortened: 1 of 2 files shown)


AdSchellevis:
It's probably beter to derive from `BaseField`, apparently I forgot to refactor this one. Most of the list types should have received a refactor like https://github.com/opnsense/core/commit/10be0e45859723069aa59f39911add181a103241.

agh1467:
@AdSchellevis I understand. I see that there are tests included in your commit as well. I'd like to try my hand at refactoring this to extend `BaseListField` type, and provide updated tests. If that's OK should I commit to this branch, or create a new PR with new commit history?

AdSchellevis:
@agh1467 I think there already are tests for the model type, but feel free to update the PR and inherit from BaseField. If you need help, just let me know.

agh1467:
@AdSchellevis Thanks for the info. There were existing tests and they were paramount! They let me know that my first attempt at the sorting functionality was incorrect.

I noticed that `ModelRelationField` was mentioned in #3252, and it was specifically skipped because of additional logic present. I think I found it, and I think I was able to successfully work around it, primarily in `getNodeData()`, and `getValidators()`. These functions were doing some similar things to the same named functions in parent classes like `BaseListField`. However, I think I got it sorted out by performing the additional logic on the data prior to calling the parent functions for both.

I added comments here and there as I was working through the logic to better understand what was going on. If I got anything incorrect I can update, or if you'd like me to remove them I can do that too. I added comments on the tests as well to make it easier to understand what the tests were at a quick glance.

I also added some more test cases, some which might be uncommon to encounter from the UI, but could possibly happen with an API call.

AdSchellevis:
@agh1467 the refactor was quite some time ago, apparently I skipped this one indeed because of "complications". This looks good at a first glance, maybe a bit too much comments in some areas, but I probably should read and test this one line by line anyway. Thanks for doing the effort, I'll try to review asap.

AdSchellevis:
@agh1467 thanks, merged with some small (style) changes in https://github.com/opnsense/core/commit/59ae021d5aba3c9388a75c51317d4950c8048e85

PR #5430, opened 2021-12-27
fix domain_needed and no_private_reverse help tooltip

ctr49:
This looks like a copy & paste error, retaining the "data-for" from the previous empty.


fichtner:
Merged, thanks!

fichtner:
I had to revert this: there is no way to toggle these that's why they latch on to the toggle for strict_order.

PR #5431, opened 2021-12-27
fix dnsmasq strict-order

ctr49:
Currently, when configured, "strict-order" adds the necessary "strict-order" argument. However, the hardcoded "all-servers" argument overrides this, making "strict-order" pointless.

With this change, "all-servers" and "strict-order" are mutually exclusive.


fichtner:
Merged, thanks! I assume this worked at one point in the past. Can’t see any reason why the code was like this for such a long time otherwise.

PR #5442, opened 2021-12-30
mvc: Add <hint> support for text fields

agh1467:
This is visually similar to select_multiple's placeholder text.

* Add support for hint (placeholder) for text field type
* Re-style similar to select_multiple to shorten line length


AdSchellevis:
@agh1467 reformatted it a bit to look more similar to the dropdown option in https://github.com/opnsense/core/commit/a5dfeb476aa30c14c7c8e661ab99067fa2664080, thanks!

Issue #5439, opened 2021-12-29
Store bootgrid settings in local browser storage


8191:
**Important notices**

Before you add a new report, we ask you kindly to acknowledge the following:

- [x] I have read the contributing guide lines at https://github.com/opnsense/core/blob/master/CONTRIBUTING.md
- [x] I am convinced that my issue is new after having checked both open and closed issues at https://github.com/opnsense/core/issues?q=is%3Aissue

**Is your feature request related to a problem? Please describe.**

Many MVC pages make use of jQuery Bootgrid (by using the OPNsense specific UIBootgrid wrapper). Bootgrid allows the user to configure many things:
- Search filters
- Elements displayed per page
- Column visibility
- Selected page

Unfortunately these settings are lost as soon as the user leaves the page.

It would be a great feature to persist those settings on a per-user basis. When a user opens a settings page they've already modified before (e.g. changed visible columns, number of rows per page, etc.) the previous settings are restored. Settings could be stored in the local browser storage.

**Describe the solution you like**
The PR  rstaib/jquery-bootgrid#392 could be incorporated into OPNsense's jQuery Bootstrap.

_Note: this is a manual task as OPNsense's jQuery Bootgrid already diverged from its upstream._
Question: where is "the master-copy" of jQuery boostrap? In the [opnsense/core](https://github.com/opnsense/core/blob/master/src/opnsense/www/js/jquery.bootgrid.js) or in [opnsense/jquery-bootstrap](https://github.com/opnsense/jquery-bootgrid) repository?

**Describe alternatives you considered**
Implementation of a "workaround" into the `UIBootgrid` function
https://github.com/opnsense/core/blob/6439c039d571519196e4c1cf5733006dcfa1321d/src/opnsense/www/js/opnsense_bootgrid_plugin.js#L59
to simulate restoring the settings, e.g. using
```javascript
var rowCount = localStorage.getItem(grid_id+"_rowCount");
$(this_grid).find('.bootgrid-pagination-count ul.dropdown-menu li a[data-action=' + rowCount + ']').trigger('click');
```

Not really nice and most likely time consuming for things like selecting columns.

**To clarify/discuss**
- Which settings to persist
- Design (implement in opnsense_bootgrid_plugin (UIBootgrid) or in jquery-bootgrid)
- How to uniquely identify a bootgrid (is the HTML ID sufficient or do we need to consider the browser URL as well?)

AdSchellevis:
[...]

PR #5443, opened 2021-12-31
Move storing jQuery Bootgrid settings in browser from core to bootgrid

8191:
Implementation of #5439 to store settings of bootgrids (sort order, visible columns, row count) in local browser storage.

- Update of jquery.bootgrid.js from 1.3.5 to 1.4.0 (opnsense/jquery-bootgrid#5)
- Removing of equivalent functionality from opnsense_bootgrid_plugin.js

### DIFF
--- a/src/opnsense/www/js/jquery.bootgrid.js
+++ b/src/opnsense/www/js/jquery.bootgrid.js
@@ -559,7 +559,7 @@ function renderRowCountSelection(actions)
         {
             var item = $(tpl.actionDropDownItem.resolve(getParams.call(that,
                 { text: getText(value), action: value })))
-                    ._bgSelectAria(value === that.rowCount)
+                    ._bgSelectAria(value.toString() === that.rowCount.toString())
                     .on("click" + namespace, menuItemSelector, function (e)
                     {
                         e.preventDefault();
--- a/src/opnsense/www/js/jquery.bootgrid.js
+++ b/src/opnsense/www/js/jquery.bootgrid.js
@@ -1,6 +1,6 @@
-/*!
- * jQuery Bootgrid v1.3.5 - 03/11/2019
- * Copyright © 2014-2015 Rafael J. Staib; Copyright © 2018-2019 Deciso B.V. (http://www.jquery-bootgrid.com)
+/*! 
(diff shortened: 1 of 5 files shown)


kulikov-a:
@8191 Hi! Huge work,  thanks!
the only flaw i found yet: saved rowCount not highlited on page load (looks like it stores rowCount as a text in localStorage and tries to compare it with a number on page load. so it always `false` at https://github.com/opnsense/core/blob/49101ba03a74c0f514945acc651557ce82ac3cbe/src/opnsense/www/js/jquery.bootgrid.js#L562 and `_bgSelectAria` not setting 'active' class to current value <li>)

8191:
@kulikov-a Thanks for testing this one! I've also recognized that flaw and I'm just working on it. It seems you already found the reason! I'll try to fix it right away. :-)

src/opnsense/www/js/jquery.bootgrid.js
```diff
@@ -559,7 +559,7 @@ function renderRowCountSelection(actions)
         {
             var item = $(tpl.actionDropDownItem.resolve(getParams.call(that,
                 { text: getText(value), action: value })))
-                    ._bgSelectAria(value === that.rowCount)
+                    ._bgSelectAria(value.toString() === that.rowCount)
```
kulikov-a, line 562: less strict comparision (replacing `===` with `==` ) works for me too )

8191: @kulikov-a I also tried this first, but then I'll get a compiler warning:

```
Running "jshint:files" (jshint) task

   dist/jquery.bootgrid.js
    562 |                    ._bgSelectAria(value == that.rowCount)
                                                    ^ Expected '===' and instead saw '=='.

>> 1 error in 1 file
Warning: Task "jshint:files" failed. Use --force to continue.

Aborted due to warnings.
```

Therefore I've changed the left part to string to avoid the warning, as current grunt settings prevent the build with warnings.

kulikov-a: @8191 wow. sorry then) the difference between simple debug and building..

AdSchellevis:
@8191 seems to be working like a charm. thanks!

PR #5445, opened 2021-12-31
dnsmasq: no hosts option; closes #5374

### LINKED ISSUE #5374: dnsmasq: Checkbox to ignore entries in /etc/hosts
**Important notices**

**Describe the bug**

I use a remote BIND server to configure DNS entries in my networks and the local OPNsense dnsmasq to use for the clients. The management IP for the OPNsense is configured in the remote BIND server for the correct interface.

Unfortunately a /etc/hosts entry is generated from the available interfaces - in my case it is using the IP from the  ipsec1000 interface, which is not the management IP/interface.

Therefore I would like the dnsmasq service to ignore the entries in /etc/hosts and only use the DNS entries from the upstream remote BIND server.

My previous work around was to configure "no-hosts" in the custom_options field, which was removed with https://github.com/opnsense/core/commit/5c58dd08c31c68725fe0d8ec454b01982d365573.

Now my work around is to add a config file in /usr/local/etc/dnsmasq.conf.d - which then is not synched to the backup firewall and might be forgotten with a new install.

**To Reproduce**

1.  Install bind DNS service on a remote machine
2. Configure hostname of OPNsense in that bind to a specific IP
3. Activate DNSmasq on OPNsense
4. Do a nslookup from a client using the OPNsense as DNS server and check the returned IP address.

**Expected behavior**

I want to get the IP address configured in bind returned instead of a random interfaces IP address of the OPNsense for the OPNsenses DNS name.

Change the /etc/hosts generation code to only use the IP from the LAN interface to generate the entry for the OPNsense hostname.

agh1467:
Attempting to link this to the issue: closes #5374

* Add checkbox to enable --no-hosts command line option

### DIFF
--- a/src/etc/inc/plugins.inc.d/dnsmasq.inc
+++ b/src/etc/inc/plugins.inc.d/dnsmasq.inc
@@ -184,6 +184,10 @@ function dnsmasq_configure_do($verbose = false)
         $args .= ' --log-queries=extra ';
     }
 
+    if (isset($config['dnsmasq']['no_hosts'])) {
+        $args .= ' --no-hosts ';
+    }
+
     $args .= ' --dns-forward-max=5000 ';
     $args .= ' --cache-size=10000 ';
     $args .= ' --local-ttl=1 ';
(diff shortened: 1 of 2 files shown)


fichtner:
merged, thanks!

PR #5448, opened 2022-01-01
show interface IP address in API

paspo:
**Problem:**
I needed a way to get the currently assigned IP address for a PPPoE interface through API.

At the moment, if you use this command:

```bash
$ curl -sku "${OPNKEY}:${OPNSEC}" https://${OPNSENSE_ADDRESS}/api/routes/gateway/status | jq
```

You get:

```json
{
  "items": [
    {
      "name": "WAN_PPPOE",
      "address": "x.x.x.x",
      "status": "none",
      "loss": "0.0 %",
      "delay": "9.1 ms",
      "stddev": "1.9 ms",
      "status_translated": "Online"
    },
    {
      "name": "WANLTE_GWv4",
      "address": "y.y.y.y",
      "status": "down",
      "loss": "100.0 %",
      "delay": "0.0 ms",
      "stddev": "0.0 ms",
      "status_translated": "Offline"
    }
  ],
  "status": "ok"
}
```

The ip address indicated in the "address" field is the monitored IP address of the gateway.

I found no other suitable API calls to get the currently assigned IP address of an interface, so I modified this one.

**Proposed solution:**

With this modification, with the same command, you get two more fields, the real interface name and the assigned IP address (doesn't matter if it comes from DHCP or is statically assigned).

```json
{
  "items": [
    {
      "name": "WAN_PPPOE",
      "ipaddr": "x.x.x.x",
      "interface": "wan",
      "address": "y.y.y.y",
      "status": "none",
      "loss": "0.0 %",
      "delay": "9.1 ms",
      "stddev": "1.9 ms",
      "status_translated": "Online"
    },
    {
      "name": "WANLTE_GWv4",
      "ipaddr": "x.x.x.x",
      "interface": "opt10",
      "address": "y.y.y.y",
      "status": "down",
      "loss": "100.0 %",
      "delay": "0.0 ms",
      "stddev": "0.0 ms",
      "status_translated": "Offline"
    }
  ],
  "status": "ok"
}
```

### DIFF
--- a/src/opnsense/mvc/app/controllers/OPNsense/Diagnostics/Api/InterfaceController.php
+++ b/src/opnsense/mvc/app/controllers/OPNsense/Diagnostics/Api/InterfaceController.php
@@ -208,6 +208,15 @@ class InterfaceController extends ApiControllerBase
     }
 
     /**
+     * retrieve status/config for each network adapter
+     * @return mixed
+     */
+    public function getInterfaceConfigAction()
+    {
+        return json_decode((new Backend())->configdRun('interface list ifconfig'), true);
+    }
+
+    /**
      * retrieve system-wide socket statistics (merge netstat with sockstat)
      * @return mixed
      */
--- a/src/opnsense/scripts/routes/gateway_status.php
+++ b/src/opnsense/scripts/routes/gateway_status.php
@@ -34,13 +34,8 @@ require_once 'interfaces.inc';
 
 $result = array();
 $gateways_status = return_gateways_status();
-$ifsinfo = get_interfaces_info();
 foreach ((new \OPNsense\Routing\Gateways(legacy_interfaces_details()))->gatewaysIndexedByName() as $gname => $gw) {
     $gatewayItem = array('name' => $gname);
-    $ifdescr = $gw['interface'];
-    $gatewayItem['interface'] = $ifdescr;
-    $ifinfo = $ifsinfo[$ifdescr];
-    $gatewayItem['ipaddr'] = !empty($ifinfo['ipaddr']) ? $ifinfo['ipaddr'] : "~";
     $gatewayItem['address'] = !empty($gw['gateway']) ? $gw['gateway'] : "~";
     if (!empty($gateways_status[$gname])) {
         $gatewayItem['status'] = strtolower($gateways_status[$gname]['status']);
--- a/src/opnsense/scripts/routes/gateway_status.php
+++ b/src/opnsense/scripts/routes/gateway_status.php
@@ -34,8 +34,13 @@ require_once 'interfaces.inc';
 
 $result = array();
 $gateways_status = return_gateways_status();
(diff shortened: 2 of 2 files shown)


AdSchellevis:
Maybe we better add an action in the `ui/diagnostics/interface/` to extract interface statistics. The configd action is already available:

```
configctl interface list ifconfig
```


`get_interfaces_info()` is only used for the legacy interface status and we should try to prevent reuse in my opinion as it tries to do more than it should.

paspo:
We already have ```/api/diagnostics/interface/getInterfaceStatistics``` which gives some statistics, but the data is hard to filter: suppose you need the ipv4 address of an interface (my case and I think the most common one), in the json output you get 3 different array items (link, ipv4 and ipv6) for each interface, without a clear way to differentiate them, sou you pick the one ending with "/32" and hope for the best.

As you stated, the output of ```configctl interface list ifconfig``` is the perfect candidate, clean and well organized.

So we're left with two different paths: 

- we can modify ```/api/diagnostics/interface/getInterfaceStatistics``` and add a "type" field (link, ipv4 and ipv6)
- we can implement a new call like ```/api/diagnostics/interface/getInterfaceStatus``` or ```/api/diagnostics/interface/getInterfaceConfig``` and use the output from ```configctl interface list ifconfig```

This is your call ;)

paspo:
The **show.interfaces** command, used by **getInterfaceStatistics** is directly bound to 
```/usr/bin/netstat -i -n -W -b -d --libxo json```. 
If we want to alter this call, we have to split into ipv4 and ipv6 versions and add the "-4" and "-6" parameters respectively to netstat.

I think the second solution is the cleanest, but it uses ```legacy_interfaces_details()``` which has the word "legacy" in the name, and I have mixed feelings about it.

I'm waiting for your opinion, thanks.

**Comment by AdSchellevis** [comment]
I think I would go for a `/api/diagnostics/interface/getInterfaceConfig` to provide access to `interface list ifconfig`, if at some  time the interface anchors (lan,wan,opt1,..) are needed, we can always extend `/usr/local/opnsense/scripts/interfaces/list_interfaces.php` to merge these in. 

`legacy_interfaces_details()` is indeed within legacy base, but the output is as it should be so it's less likely subjected to change. (I don't expect it will be replaced any time soon to be honest as too many components rely on structured `ifconfig` data)

**Comment by paspo** [comment]
This is now too simple to be true...

**Comment by AdSchellevis** [comment]
@paspo we might need to check the existing ACL's, but yes, it's indeed that simple :)

**Comment by paspo** [comment]
> @paspo we might need to check the existing ACL's, but yes, it's indeed that simple :)

how can I help?

**Comment by AdSchellevis** [comment]
It doesn't seem to be exposed anywhere automatically, so this is safe to merge:
```
git grep "api/diagnostics/interface/" | grep ACL
```


Eventually we might need to add an ACL if non admin users should be able to access this endpoint, but there's no rush.

PR #5449, opened 2022-01-01
IPsec public keys: provide generation function within GUI

8191:
Provide a "Generate" button in the IPsec RSA keys dialog to generate public/private keys:
image


src/opnsense/scripts/ipsec/genkeys.py
```diff
@@ -0,0 +1,74 @@
...
+import subprocess
+import ujson
+
+result = dict()
+if __name__ == '__main__':
+    # parse input parameter
+    keytype = None
+    keysize = None
+    if len(sys.argv) > 2:
+        keytype = sys.argv[1].strip().lower()
```
AdSchellevis, line 41: it might be better to use `argparse` (https://docs.python.org/3/library/argparse.html) for parameter handling, less code :)

AdSchellevis:
@8191 it might also be an option to only add a drop down "generate new key", which by default is set to no, otherwise is using the key size. In which case you can move the responsibility to the controller and keep the frontend implementation simpler.

8191:
@AdSchellevis Thanks for the review, I've incorporated your comments.

Regarding the UI part: yes, having it in the controller might be better from a design point of view, but the downside is, that the user needs to reopen the already closed edit window to see the generated key (which he needs to distribute very likely).

AdSchellevis:
https://github.com/opnsense/core/commit/e395da80b2f2f1d659413030bff6d1b0bcad4f05

PR #5457, opened 2022-01-04
IPsec: Add EC support for public key authentication

8191:
This PR adds elliptic curve support for public key authentication, namely ECDSA, Ed25519 and Ed448.

* Also PR #5449 is part of this, as the functionality depends on this (if this PR gets accepted, other PR can be closed).
* Additional column in key overview: private key (yes/no)
* Renamed menu entry "RSA Key Pairs" to "Key Pairs"

Unfortunately the PHP wrapper for OpenSSL does not support Ed25519 and Ed448 (only ECDSA). Therefore the sanity check of the pasted key needs to be done quite awkwardly by directly calling the openssl binary and parsing the output using preg_match:
https://github.com/8191/opnsense-core/blob/d580772baddcda67d0a7a8bc4658341332bbcea2/src/opnsense/mvc/app/models/OPNsense/IPsec/IPsec.php#L255-L290
Hopefully this could be removed in near future as soon as PHP supports those curves.

The implementation itself was partially tested between two OPNsense installations, meaning it was checked that the authentication succeeds (` 15[IKE] <con4|1> authentication of '10.4.1.1' (myself) with ED25519 successful`). A full end-to-end test was not yet done due to [an open issue with my IPsec setup](https://forum.opnsense.org/index.php?topic=26231.0).

Potential improvements:
* Detect pasted key type instead of requesting the user to select it (the model validator determines the key type anyway to perform sanity checks)
* Move key generation from view to controller (as suggested by @AdSchellevis in #5499)

image

### DIFF
--- a/src/opnsense/mvc/app/controllers/OPNsense/IPsec/Api/KeyPairsController.php
+++ b/src/opnsense/mvc/app/controllers/OPNsense/IPsec/Api/KeyPairsController.php
@@ -29,6 +29,7 @@
 namespace OPNsense\IPsec\Api;
 
 use OPNsense\Base\ApiMutableModelControllerBase;
+use OPNsense\Core\Backend;
 
 /**
  * Class KeyPairsController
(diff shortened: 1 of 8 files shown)


AdSchellevis:
added a simplified version in https://github.com/opnsense/core/commit/e395da80b2f2f1d659413030bff6d1b0bcad4f05, at the moment I  would like to stick to what php has to offer for now.

PR #5463, opened 2022-01-06
Column Spacing

NOYB:
Reduced and consistent column spacing for more efficient display real estate use and less line wrap.


8191:
This _seems_ a quite invasive change as this point is quite global. Also this would require a change in contributed themes (plugins/misc).
What is the initial intention for this change? Can you please explain the precise problem - best with screenshots...

8191:
Just saw the PR for the contributed themes: opnsense/plugins#2743

PR #5465, opened 2022-01-06
IPsec: re-implement status pages using MVC

8191:
Reworked the IPsec status pages
* Security Association Database
* Security Policy Database
* Status Overview

image


8191:
@AdSchellevis I fixed this with minimal effort. Or do you see any advantage in putting some love into the legacy code within _etc/inc/plugins.inc.d/ipsec.inc_?

AdSchellevis:
> @AdSchellevis I fixed this with minimal effort. Or do you see any advantage in putting some love into the legacy code within etc/inc/plugins.inc.d/ipsec.inc?

That probably depends, it's also a simple candidate to wrap into a python script and push it to a configd call

8191:
First version of MVC implementation of SPD and SAD pages. Legacy pages still exist, but are not linked in the menu.

8191:
@AdSchellevis There is now one page showing both database in two different tabs:

image

But currently there are two separate ACLs for SAD and SPD access:
https://github.com/opnsense/core/blob/fb041467bf17155b1736b42c8e65a769f984d0b1/src/opnsense/mvc/app/models/OPNsense/IPsec/ACL/ACL.xml#L64-L75

How to solve this?
1. Merge both ACLs - how to migrate? Does the model migration concept also work for ACLs?
2. Separate the two pages

fichtner:
IMO status, leases and these two are all the same. As for supporting legacy ACL design maybe keep all of them but support all pages for it in each? If people need more fine-grained ACLs they can create them. The GUI should in these cases not render the contents without an error except for displaying no permission or disabling the tab? I don’t think we had this luxury problem before but seems worth solving for future GUI flexibility and MVC conversions.


Cheers,
Franco

fichtner:
PS: in a major iteration we can get away with spitting ACLs in non-compatible way but that would already be 22.7 release since we’re almost done with 22.1.

8191:
> IMO status, leases and these two are all the same.

Agree. I've extended this PR and reworked all these pages, except the leases.

image

8191:
I'd have a general design question: where to put pagination and filter logic? In the API controller or the backend script providing the actual data?

I've seen e.g. in _scripts/filter/list_states.py_ that pagination and filtering is in the backend script, but from my opinion that's something which would better fit in the API controller. Was there a specific reason (e.g. performance when many states are active) to put this functionality into the backend script in that case?

I think having pagination in the script clutters the core functionality of the script. It also duplicates code, as this logic could even be shifted to the `ApiControllerBase`.

8191:
I'm still considering moving the pagination logic to the ApiController to clean up src/opnsense/scripts/ipsec/list_status.py.

AdSchellevis:
@8191 I was away last week, so a bit late in responding on this ticket. 

About the ACL for these edge cases, I wouldn’t worry too much, if we merge two pages into one which makes sense for a functional perspective, we can ditch the old ones and add a new one. If for some reason someone really only wants to grant access to a single topic, we can always discuss options then. Most people probably allow all IPsec or are looking at these pages as admin users anyway.

> I'd have a general design question: where to put pagination and filter logic? In the API controller or the backend script providing the actual data?
> 
> I've seen e.g. in scripts/filter/list_states.py that pagination and filtering is in the backend script, but from my opinion that's something which would better fit in the API controller. Was there a specific reason (e.g. performance when many states are active) to put this functionality into the backend script in that case?

Yes, if the number of entries are limited, it really makes sense to handle filering in the API, we might even consider adding a standard class or wrapper method to handle this for use, as the filter code in IPsec is already quite generic:

https://github.com/opnsense/core/blob/2a970b568a2839f2a73f04fbab2f36a532bcc2d2/src/opnsense/mvc/app/controllers/OPNsense/IPsec/Api/TunnelController.php#L45-L82

Downside however is, you will need to process the same information multiple times, when for example handling a log stream, this will be highly inefficient as the number of source records can easily exceed Gigabytes of memory. In quite some cases the number of results is limited, in which this approach will work like a charm and keeps the `configd` action code simpler. 
A couple Megabytes of expected maximum output usually isn't a huge issue to [de]serialise a couple of times through the stack.

> 
> I think having pagination in the script clutters the core functionality of the script. It also duplicates code, as this logic could even be shifted to the ApiControllerBase.

Yes it does, at the moment I think we only push back when the previous mentioned rule applies (too much data, which could lead to an out of memory).

In this case (SPD/SA's), I would expect the number of results will always be quite limited.

8191:
Yeah, I also doubt that the ipsec status output would be several megabytes, even when having many tunnels configured. 
Will provide a reworked backend script without pagination soon.

8191:
@AdSchellevis Any chance to have a look at this within the next weeks? 😃

AdSchellevis:
@8191 I wasn't sure you where ready with this, I'll try to give it a shot on my end.

src/opnsense/mvc/app/controllers/OPNsense/Diagnostics/IpsecController.php
```diff
@@ -0,0 +1,54 @@
...
+ */
+class IpsecController extends IndexController
+{
+    /**
+     * IPsec status view
+     */
+    public function statusAction()
+    {
+        $acl = new ACL();
+        $this->view->allow_connection = $acl->isPageAccessible($this->session->get("Username"), "/api/diagnostics/ipsec/searchConnection");
```
AdSchellevis, line 48: @8191 I think we better merge the ACL's into a single one as suggested earlier, there likely aren't may use-cases where people only should have access to part of this information and it removes complexity.

src/opnsense/mvc/app/controllers/OPNsense/Base/ApiControllerBase.php
```diff
@@ -68,6 +68,63 @@ private function parseJsonBodyData()
...
+                array_multisort($keys, $order == 'asc' ? SORT_ASC : SORT_DESC, $records);
+            }
+        }
+
+        // Filter entries
+        $entry_keys = array_keys($records);
+        if ($this->request->hasPost('searchPhrase') && $this->request->getPost('searchPhrase') !== '') {
+            $searchPhrase = (string)$this->request->getPost('searchPhrase');
+            $entry_keys = array_filter($entry_keys, function ($key) use ($searchPhrase, $records) {
+                return $this->recursiveFilter($records[$key], $searchPhrase);
```
AdSchellevis, line 110: I'm a bit doubting we should try to parse these records recursively to be honest, my initial thought was to bring the existing code (from TunnelController.php) into here (optionally with a parameter to specify which fields to search). The downside from filtering structures is that results will be returned which are unlikely to be returned in the view.

src/opnsense/scripts/ipsec/list_status.py
```diff
@@ -52,39 +59,215 @@ def parse_sa(in_conn):
...
-        result[sa]['sas'].append(sas[sa])
-
-print (ujson.dumps(result, reject_bytes=False))
+
+def get_conn_description(connection_id):
+    """Parse config.xml for connection description
+
+    :connection_id: IPsec connection ID (e.g. con3)
+    """
+    global ipsec_config
```
AdSchellevis, line 71: better not to use globals, if we do need to collect a list of data better create a function that returns the required map (id -> name) or we should think about putting more of this in a class.

AdSchellevis (commented):
first batch of review comments

src/opnsense/mvc/app/controllers/OPNsense/Base/ApiControllerBase.php
AdSchellevis: moved the generic search code in https://github.com/opnsense/core/commit/2e747acda90d447f4047c5e7bce941fb2f8bf5cb

src/opnsense/mvc/app/controllers/OPNsense/Diagnostics/IpsecController.php
8191: @AdSchellevis Is there any straight-forward migration for such kind of ACL modifications? Or shall we just drop the unneeded and one needs to re-assign ACLs after in upgrade in case the dropped ones were used?

I suggest keeping "page-status-ipsec" and "page-status-ipsec-edit", while dropping "page-status-ipsec-sad" and "-spd".

AdSchellevis: @8191 there's no migration for it, but given the relevancy of these acl's I would go for your suggestion to keep `page-status-ipsec` and be done with it.

AdSchellevis:
closing this in favour of https://github.com/opnsense/core/commit/023d8e4a35939d0cb11cb5ebdfebb7a48fa2f4b8, https://github.com/opnsense/core/commit/245a4b8ca07b45e2e233be05db58703c48696254, https://github.com/opnsense/core/commit/3726f67882c03baa88badb649a617bafcbbafd86 (refactors SAD/SPD). Connection status diagnostics will likely follow using the same approach.

PR #5471, opened 2022-01-12
Translate selectpicker for legacy code.

kekek2:


AdSchellevis:
I don't think we use any of these in legacy pages, skipping this PR to avoid further regressions.

PR #5472, opened 2022-01-12
Logs: handle Firewall Plain View

kulikov-a:
Hi!
can fix regression in https://github.com/opnsense/core/pull/5370 for Firewall Plain View Log (empty with default parameters as it is written with Informational severity)
ref. https://forum.opnsense.org/index.php?topic=26325.0

sorry for the inattention
thanks!


AdSchellevis:
@kulikov-a how about https://github.com/opnsense/core/pull/5472 ? Service binding is a similar issue, which needs some additional metadata, if the amount of "edge-cases" grow, we can feed it in the controller at the same spot and be done with it :)

kulikov-a:
cool, thanks!! of course it looks like a more complete solution (a strange habit of trying to solve everything on the client side)

Issue #5470, opened 2022-01-11
system/ntpd rate limit on by default, restrict: detach limited from kod


ZsZs73:
**Important notices**

Before you add a new report, we ask you kindly to acknowledge the following:

- [x] I have read the contributing guide lines at https://github.com/opnsense/core/blob/master/CONTRIBUTING.md
- [x] I have searched the existing issues and I am convinced that mine is new.

**Is your feature request related to a problem? Please describe.**
I've learned it in the hard way that the rate limit is ON as per default NTPD configuration (both `restrict` and `restrict -6` commands have the flag`limit`). 
`limit` can be enabled/disabled only together with `kod` flag. 
See `/usr/local/etc/inc/plugins.inc.d/ntpd.inc`
```
375     if (empty($config['ntpd']['kod'])) { /*note: this one works backwards */
376         $ntpcfg .= ' kod limited';
377     }
```
IMHO the default rate limit very strict. The default `discard` allows `average` 32s (2^5) between two requests from the same source and the `minimum` between two requests is 4s (2^2) seconds according to the [ntp.conf manual](https://www.freebsd.org/cgi/man.cgi?query=ntp.conf)
Most of the modern clients have `server` parameter with `iburst` configured by default which triggers the default rate limit immediately.

**To Reproduce**

1. Go to 'Services / Network time'
2. Click on 'General'
3. Scroll down to 'Access restriction'
4. Click Advanced button
4. Check 'Enable Kiss-o'-death packets' checkbox
5. Result: 'kod limited' will be added to `restrict` and `restrict -6` commands of `/var/etc/ntpd.conf`
6. Uncheck ' Enable Kiss-o'-death packets' checkbox
7. Result: 'kod limited' will be removed from `restrict` and `restrict -6` commands of `/var/etc/ntpd.conf`

**Expected behavior**
- option 1: `limited` should have its own checkbox on the UI and it should be unchecked by default. Easy to implement.
- option 2: the optimal solution would be to make it possible to add multiple `restrict` and `restrict -6` lines with all possible flags shown. More complex to implement as it needs an array of checkboxes with mandatory (`default`) and optional networks.

Note 1: currently restrictions are getting applied to all clients (`restrict default` means: `restrict address 0.0.0.0  mask 0.0.0.0`, so any clients). The only way to add custom restrictions is to add complete `restrict` commands to the Advanced section of Services / Network Time / General which is not recommended.
[...]

PR #5473, opened 2022-01-12
services/ntp - detach limited from kod, correct help text

ZsZs73:
Hi @fichtner,
This is the PR to address [issue #5470](https://github.com/opnsense/core/issues/5470)
I am not a regular github user so I hope that you can extract something useful from this PR.

### DIFF
--- a/src/www/services_ntpd.php
+++ b/src/www/services_ntpd.php
@@ -401,7 +401,7 @@ include("head.inc");
                       <input type="button" id="showrestrictbox" class="btn btn-default btn-xs" value="<?= html_safe(gettext('Advanced')) ?>" /> - <?=gettext("Show access restriction options");?>
                       </div>
                       <div id="showrestrict" style="display:none">
-                      <?=gettext("These options control access to NTP from all interfaces."); ?>
+                      <?=gettext("These options control access to NTP."); ?>
                       <br /><br />
                       <input name="kod" type="checkbox" id="kod"<?=empty($pconfig['kod']) ? " checked=\"checked\"" : ""; ?> />
                       <?=gettext("Enable Kiss-o'-death packets"); ?>
--- a/src/www/services_ntpd.php
+++ b/src/www/services_ntpd.php
@@ -41,6 +41,7 @@ $copy_fields = [
     'clockstats',
     'interface',
     'kod',
+    'limited',
     'leapsec',
     'logpeer',
     'logsys',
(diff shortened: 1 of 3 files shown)


ZsZs73:
Forgot to ask how to make sure that `limited` is unchecked by default (on new installs).
Note that this would change the default behaviour as previously it was set to `limited`.

src/www/services_ntpd.php
```diff
@@ -400,11 +401,14 @@ function removeRow() {
                       <input type="button" id="showrestrictbox" class="btn btn-default btn-xs" value="<?= html_safe(gettext('Advanced')) ?>" /> - <?=gettext("Show access restriction options");?>
                       </div>
                       <div id="showrestrict" style="display:none">
-                      <?=gettext("These options control access to NTP from the WAN."); ?>
+                      <?=gettext("These options control access to NTP from all interfaces."); ?>
```
fichtner, line 404:
```suggestion
                      <?=gettext("These options control access to NTP."); ?>
```

Might be better to make it shorter?

fichtner:
merged, thanks! Will probably land in 22.1 :)

ZsZs73:
Thank you, however I am still not sure if `limited `will be enabled by default or not.

fichtner:
Did a test and "limited" renders in the config and box is checked by default. Unchecking saves it, removes "limited" from config. Checking it again adds "limited" and removes setting from config... works fine from what I can tell :)

ZsZs73:
I've also tesed it like you did before submitting the PR. 
I meant if the `limited` will be enabled on a fresh install or not. I could not find how the initial `ntpd.conf` gets generated after a fresh install. Default `ntpd.conf` does not exist in the ntp package.
IMHO `limited` should not be enabled on a fresh install like it is not enabled on server OSs.

fichtner:
It is added on a fresh install to retain backwards compatibility. I thought that was your intention.  ;)

ZsZs73:
I realized that there is a rate limit in place because some of my IoT devices could not sync to my OPNsense firewall wich had default NTP config with `limited `enabled.
In my opinion th default ntpd config with `limited `enabled is too restrictive for a local network, espescially for some IoT devices (namley ones running [NTPClient](https://github.com/arduino-libraries/NTPClient))
However I confirm that other linux boxes can sync even with iburst set.

To summarize:
- `limited` enabled: 
  - IoT devices might not be able to sync (two NTP requests immediately after each other)
  - Linux boxes with full blown ntpd running on them sync well even with `iburst` set. (NTP requests with 2 secs pause between them)
- `limited` disabled: anything I checked could sync properly

In meanwhile I'll check if the NTPClient can be fixed, but it will not help to the already deployed IoTs which possibly never get updated.

fichtner:
But leaving limited out of the default configuration will likely back off hardening efforts. I don’t think IoT is a good reason to roll back such protection.

PR #5482, opened 2022-01-17
IDS: 1.0.7 - migrate with policies update

kulikov-a:
Hi!
For IDS 1.0.7 migration possible fix with policies update discussion:
-`filter` field references removed (since field is gone in 1.0.6)
-replace old uuids with new one(s) in policies `rulesets` field
-remove possible duplicates and obsolete references in policies (since the model requires the field `enabled` to be equal to "1")

@AdSchellevis Hi! can we discuss this when you have time?
did i get the idea right?
and question: Please explain why the model has a requirement that the ruleset (to which the policy refers) is enabled? If I understand correctly, in this case the following situation is possible:
-Administrator includes the ruleset in the policy.
-For some reason decides to (perhaps temporarily) disable the ruleset.
-If there is a update\migration or just a policy change, since the policy is touched the model checks the value of the `enabled` field and does not allow references to a disabled (but existing) ruleset.
-The policy silently loses ruleset reference (during migration) or the administrator is need to look for the cause of complaints while saving the policy?

and the second one: may be we need to do something with the enabled `$csets` rulesets? may be its worth to disable them?

Thanks!

### DIFF
--- a/src/opnsense/mvc/app/models/OPNsense/IDS/Migrations/M1_0_7.php
+++ b/src/opnsense/mvc/app/models/OPNsense/IDS/Migrations/M1_0_7.php
@@ -45,19 +45,29 @@ class M1_0_7 extends BaseModelMigration
         if (!isset($cfgObj->OPNsense->IDS->files->file)) {
             return;
         }
-        $csets = array();
-        $nsets = array();
+        $csets = array(); // changed rulesets presented in model
+        $nsets = array(); // new rulesets nodes
+        $uuids = array(); // all enabled sets uuids
+        $c_uuids = array(); // changed sets uuids
+        $n_uuids = array(); // new sets uuids
+        $s_array = array(); // search uuids array for policies update
+        $r_array = array(); // replacement uuids array for policies update
         $changed_sets = ['emerging-current_events.rules', 'emerging-trojan.rules',
                          'emerging-malware.rules',  'emerging-info.rules', 'emerging-policy.rules'];
         $new_sets = ['emerging-ja3.rules', 'emerging-hunting.rules', 'emerging-adware_pup.rules',
                      'emerging-phishing.rules', 'emerging-exploit_kit.rules', 'emerging-coinminer.rules',
                      'emerging-malware.rules'];
         foreach ($model->files->file->iterateItems() as $file) {
+            if ($file->enabled == "1") {
+                $uuids[] = (string)$file->getAttribute('uuid');
+            }
             if (in_array((string)$file->filename, $changed_sets)) {
                 $csets[(string)$file->filename] = $file;
+                $c_uuids[(string)$file->filename] = (string)$file->getAttribute('uuid');
             }
             if (in_array((string)$file->filename, $new_sets)) {
                 $nsets[(string)$file->filename] = $file;
+                $n_uuids[(string)$file->filename] = (string)$file->getAttribute('uuid');
             }
         }
         // add all new to config in deselected state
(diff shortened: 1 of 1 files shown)


AdSchellevis:
@kulikov-a sure, but it looks like I made a mistake here.

> did i get the idea right?

I'm not sure, the original intent was to change activated rulesets into their new "siblings" including the action to perform (alert or drop). But somewhere along the lines something went wrong.....

> and question: Please explain why the model has a requirement that the ruleset (to which the policy refers) is enabled? 

If the ruleset wasn't enabled and seizes to exist, it doesn't really make sense to migrate it either. This migration was more or less a rename action for files that where renamed. for example `emerging-malware.rules` in version 4&5 (https://rules.emergingthreats.net/open/suricata-5.0.0/rules/) is being replaced with `emerging-adware_pup.rules` in version 6 (https://rules.emergingthreats.net/open/suricata-6.0.0/rules/).

But in between, we added a migration for the policies in version 1.0.6. So in order to fix this properly, we should likely
 traverse the files recorded in the model and replace them with the new ones (search `files->file` for items in this migration and rename them to the new names).  Since some are being split, ideally we should add them as a copy of the original content and search the policies for the first one. (e.g. emerging-current_events --> emerging-phishing + emerging-exploit_kit)

Given we already are shipping this, I don't mind shortcuts either knowing not all selections are being converted, this all tends to get far too complex for what we're trying todo (initially it was simple, time made it worse). Shortest fix probably is to remove the filter part and leave the rest as it was, knowing "drop" policies will have to be set again.

In the mean time I also noticed we are shipping the wrong open version which doesn't include the new files (https://github.com/opnsense/core/commit/c30e535ac960e1555391eabd3052f4193c49218c)

Lesson learned, when code needs to survive for almost two years (https://github.com/opnsense/core/commit/09f74fe1ce7b4805c1bb8354e5d0c5b98955497b), chances are reasonable that it doesn't fit anymore.

kulikov-a:
@AdSchellevis thanks! sorry for delay, thinking about answers)

> If the ruleset wasn't enabled and seizes to exist, it doesn't really make sense to migrate it either.

probably i, as always, did not convey the idea very clearly. live example: on 21.7.7 test vm i enabled snort rulesets to test their work and made a policy for them. then i needed to clarify the operation of the "Download&Update rules" button, and to reduce the reaction time, i disabled these rulesets. the system let me do it (no back-reference check?) and that was what was needed. after I fiddled around with the button I just re-enabled those sets - the policies weren't affected.
but (if i understand correctly) if at that moment the migration touched the policies, then the model would not allow to save this policy with a link to the disabled ruleset.

> This migration was more or less a rename action for files that where renamed

looking at the code i would guess that the idea was to add new rulesets to the model and enable those that corresponds the changed ones?

>  Since some are being split, ideally we should add them as a copy of the original content and search the policies for the first one. (e.g. emerging-current_events --> emerging-phishing + emerging-exploit_kit)

sorry, i don't quite understand the "add them as a copy of the original content and search the policies for the first one" part. but wouldn't the proposed method give the same result? (policies referencing new rule sets as per proofpoint rulesets migration)

> Shortest fix probably is to remove the filter part and leave the rest as it

you are probably right (although i liked the idea of ​​updating policies, apart from the possibility of losing refreneces with existing but disabled rulesets): did not see complaints, and as a result of the 1.0.7 migration that worked, new rulesets will simply appear and turn on in accordance with existing ones. 
besides, i could make a mistake in the uuids replacing logic, since I did not find a proofpoint document with a specific migration logic, except for the migration script and the [link](https://www.proofpoint.com/us/corporate-blog/post/emerging-threats-announcing-support-suricata-50)  and they don’t really fit together for me

ready to change pr according to your decision )

AdSchellevis:
@kulikov-a 

> probably i, as always, did not convey the idea very clearly. 
Don't be hard on yourself here, the intention wasn't that clear either.

> live example: on 21.7.7 test vm i enabled snort rulesets to test their work and made a policy for them. then i needed to clarify the operation of the "Download&Update rules" button, and to reduce the reaction time, i disabled these rulesets. the system let me do it (no back-reference check?) and that was what was needed. after I fiddled around with the button I just re-enabled those sets - the policies weren't affected.
> but (if i understand correctly) if at that moment the migration touched the policies, then the model would not allow to save this policy with a link to the disabled ruleset.

Keep in mind we're only looking from the context of the migration here, from a referential integrity stand of view the data is still there, although in this case I don't think we are enforcing integrity either. (like a "safe" delete can do)
https://github.com/opnsense/core/blob/99c9dc3da585f1c70716a53fac2305e26b549fee/src/opnsense/mvc/app/controllers/OPNsense/Base/ApiMutableModelControllerBase.php#L89
 
Let's call the rulesets "eventually consistent", if referenced but non existent it shouldn't break anything either.

> looking at the code i would guess that the idea was to add new rulesets to the model and enable those that corresponds the changed ones?

Yes, more or less since the changes add rulesets in most cases, so we do need to add references anyway. In https://github.com/opnsense/core/commit/09083b2404bb2a99b082b6442777aff50099ec62 I made sure the migration doesn't break anymore and does "the right thing" from the rulesets, knowing we don't add policies.


> sorry, i don't quite understand the "add them as a copy of the original content and search the policies for the first one" part. but wouldn't the proposed method give the same result? (policies referencing new rule sets as per proofpoint rulesets migration)

Well, since the policy migration has already taken place in 1.0.6, logically we should look for policies including the uuid's of the old rulesets and add/replace them with the new/added ones. But currently I don't mind leaving this as it is, since this will take quite some development time for something will only cause issues once.

The intent however would be for a a policy including the uuid of `emerging-trojan.rules` to remove that reference and add the new ones added in the 1.0.7 migration (uuid's of) `emerging-coinminer.rules` + `emerging-malware.rules`

> you are probably right (although i liked the idea of ​​updating policies, apart from the possibility of losing refreneces with existing but disabled rulesets): did not see complaints, and as a result of the 1.0.7 migration that worked, new rulesets will simply appear and turn on in accordance with existing ones.

It's the effort needed at the moment and the fact that we already shipped this which makes this less relevant in my opinion. 

> besides, i could make a mistake in the uuids replacing logic, since I did not find a proofpoint document with a specific migration logic, except for the migration script and the link and they don’t really fit together for me

Unfortunately I know, the new files are part of their Suricata 6 rulesets and I'm afraid the list of changes only lives internally and somewhere in my mailbox.  Let me try to ask on their end if they can publish a new document explaining the sets as they are now since the document you found is about suri 5.

> ready to change pr according to your decision )

I think we can close the PR if you agree.

kulikov-a:
> The intent however would be for a a policy including the uuid of emerging-trojan.rules to remove that reference and add the new ones added in the 1.0.7 migration (uuid's of) emerging-coinminer.rules + emerging-malware.rules

agree. i just thought that this part of pr was doing that

```
        foreach ($model->policies->policy->iterateItems() as $policy) {
            if ((string)($policy->rulesets) != "") {
                var_dump((string)$policy->rulesets);
                // replace changed sets with new one and remove obsolete links (ruleset no exists or not enabled)
                $sets = array_unique(explode(',', str_replace($s_array, $r_array, $policy->rulesets)));
                foreach ($sets as $i => $uuid) {
                    if (!in_array($uuid, $uuids)) {
                        unset($sets[$i]);
                    }
                }
                $policy->rulesets = implode(',', $sets);
            }
        }
```

> I think we can close the PR if you agree.

certainly! perhaps I really paid too much attention to a not particularly significant issue )

AdSchellevis:
> certainly! perhaps I really paid too much attention to a not particularly significant issue )

No problem, if I really don't want to discuss things I will let you know as well. Your efforts are highly appreciated over here, thanks a lot!

PR #5483, opened 2022-01-17
Certificates/Trust: first step towards MVC for certificate management

8191:
Draft PR to indicate that something is going on in that direction. And maybe some first feedback if the change itself is desired.

**Open points:**
- [ ] Legacy implementation also deletes all certificates of a CA when deleting the CA. I personally think that's not really desired and therefore did not implement this in this PR. ➡️ Requires final decision
- [ ] It is (and also was in legacy impl.) possible to delete a CRL which is assigned to an OpenVPN server.
- [ ] Intensive testing

### DIFF
--- a/src/opnsense/mvc/app/controllers/OPNsense/PKI/Api/CertificateController.php
+++ b/src/opnsense/mvc/app/controllers/OPNsense/PKI/Api/CertificateController.php
@@ -1,0 +1,320 @@
+<?php
+
+/*
+ * Copyright (C) 2022 Manuel Faux <mfaux@conf.at>
+ * Copyright (C) 2021 Deciso B.V.
+ * All rights reserved.
+ *
+ * Redistribution and use in source and binary forms, with or without
+ * modification, are permitted provided that the following conditions are met:
+ *
+ * 1. Redistributions of source code must retain the above copyright notice,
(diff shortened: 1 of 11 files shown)


AdSchellevis:
I'm not against refactoring the Trust section, I'm just not sure we can spend enough time to make this happen short time. It might also be easier to start isolating the search functionality by the way so less code is being copied around,  currently we are quite busy preparing for 22.1, but don't let me hold you from progressing this further.

8191:
> I'm just not sure we can spend enough time to make this happen short time

I didn't expect this to be merged Amy soon. It's a crucial functionality, so better longer testing than a broken PKI... 😀 

Maybe a candidate for 22.7.

8191:
@AdSchellevis If you have time to look at this, just give me a hint upfront so that I can rebase onto the current master...

AdSchellevis:
@8191 I'll keep it on the list, but given the amount of time needed to review and test this, it might take a while (it's not on the top of my priority list)

Marvo2011:
Any news to this? I want to get access to certificates (and users) from the API. Looks that is planed for 23.7?

AdSchellevis:
https://github.com/opnsense/core/issues/7248

PR #5484, opened 2022-01-18
DNS Rebind Check Case Sensitivity

NOYB:
in_array() is case sensitive; host and domain names are not.


AdSchellevis:
@NOYB thanks, you're right, missed a spot there

Issue #5392, opened 2021-12-05
API unbound dns override


Thibaut833:
Hi,

i come here to know if there is a solution about the opnsense's api on unbound dns services, particularly to post override host to the firewall.

Is there a functionnality to post an override host in unbound dns service via api rest ?

Is there an alternative ? Via config xml with a reboot to push the modifications ? 

Best regards,
Thibaut.

tabnul:
I am also very interested in this. In fact i think it is a very common use case (automatic provisioning machines, and creating 'dns records' for them.

I tried to debug the GUI, adding the record should be doable (post with some data, do a second APPLY post. Biggest challenge will be avoiding the CSRF.
Removing it is almost impossible since it uses an ID that can only be found in a table in the HTML.

swhite2:
@Thibaut833 @tabnul This is now merged in master: https://github.com/opnsense/core/pull/5488.

PR #5488, opened 2022-01-18
unbound: overrides: migrate to mvc model

swhite2:
This PR moves the legacy Unbound->Overrides section to the new MVC model, enabling it for use with the API. 

A couple of things to note:

- This commit changes the FieldType `HostnameField` to now also accept wildcard entries (*). Default for this is false, so no impact.
- ~~Adds a template helper which retrieves the current configured IP Address of an interface, useful for non-static assignments as they are not stored in config.~~
- Adds a template test in `src/opnsense/service/modules/template.py` to allow for regex checking in a template.
- ~~While most of the relevant code from `unbound.inc` has been deleted, checking the DHCP configured static map (both Ipv4 and Ipv6) and converting this to the relevant local-data sections for Unbound is still done by the `unbound_add_host_entries_dhcp()` function due to stringent IPv6 address-checking. It now writes this to a seperate include (`host_entries_dhcp.conf`).~~ 
To avoid a huge template with too much logic in it, generating the config is still done by `unbound.inc`, it, however, now uses the base model methods to achieve this.
- Provides a migration from host & domain overrides in the legacy config section to the OPNsense/Unbound section.
- Addresses https://github.com/opnsense/core/issues/5392


src/opnsense/mvc/app/models/OPNsense/Unbound/Migrations/M1_0_1.php
```diff
@@ -0,0 +1,113 @@
...
+            return;
+        }
+
+        if (!empty($config->unbound->hosts)) {
+            foreach($config->unbound->hosts as $old_host) {
+                $new_host = $model->hosts->host->add();
+
+                /* Backwards compatibility for records created before introducing RR types. */
+                if (!isset($old_host->rr)) {
+                    require_once 'util.inc';
```
AdSchellevis, line 57: It's not allowed to use legacy functions in MVC, checking for the existence of `:` would probably be enough here.

src/opnsense/service/modules/addons/template_helpers.py
```diff
@@ -31,9 +31,9 @@
 import os
 import glob
 import collections
+import subprocess
```
AdSchellevis, line 34: leftover from the ifconfig code, please delete :)

AdSchellevis (changes requested):
@swhite2 some small remarks, but looks good. I haven't tried it on my end yet, but looking at the code I don't expect huge issues to be honest

swhite2:
@AdSchellevis Thanks for the review, the latest commit addresses all of the above remarks.

src/opnsense/service/templates/OPNsense/Unbound/core/domainoverrides.conf
```diff
@@ -0,0 +1,17 @@
+{% if not helpers.empty('OPNsense.unboundplus.domains.domain') %}
```
fichtner, line 1: personally would prefer leaving the "unboundplus" namespace for default "unbound" if we are going to migrate anyway

swhite2: Wouldn't this require a larger change beyond the scope of this PR? Just my 2 cents.

fichtner: I'm not sure if model renames are a good idea for general model migration behaviour reasons (@AdSchellevis would know better) but except this the change would entail changing a few configuration template lines and that isn't unreasonable to do so during the migration taking place here anyway. M1_0_0.php already contains the code to be used for this. It takes the contents from the old config location and moves it to the new model.

swhite2: Agreed. Impact doesn't seem to be that high. I'd also like to know what @AdSchellevis thinks about model renaming.

AdSchellevis: @swhite2 @fichtner I don't mind that much about renaming the model, but to limit impact it might be better to test and merge this first and add another model version explicitly for the move. race conditions may lead to unintended side affects.

fichtner: Only concern is that templates need to be changed after adding more of them to migrate but the end result is still I can hope for 😊

AdSchellevis:
@swhite2 just tried it on my end, seems to work fine. thanks!

Issue #5487, opened 2022-01-18
Per-rule adaptavie timeouts


ganguin:
**Important notices**

Before you add a new report, we ask you kindly to acknowledge the following:

- [X] I have read the contributing guide lines at https://github.com/opnsense/core/blob/master/CONTRIBUTING.md
- [X] I am convinced that my issue is new after having checked both open and closed issues at https://github.com/opnsense/core/issues?q=is%3Aissue

**Is your feature request related to a problem? Please describe.**

I manage a setup with global adaptative timeout. Additionally, I have a max-src-state directive such that a single source cannot fill the global state table. However, with this configuration the adaptative timeout does not kick in when a user reaches his max-src-state. pf allows to configure the adaptative timeout directive on a rule, alongside with max-src-state, but the OPNsense configuration does not allow to set the adaptative timeout on a firewall rule.

The global adaptative seeting are present in `/system_advanced_firewall.php` under Firewall Adaptive Timeouts (values start and end).

The same options are missing in `/firewall_rules_edit.php?if=xxx` (all interfaces) in the Advanced Options alongside with "Max states" or "Max source states".

**Describe the solution you like**

Add the adaptative timeout start and end fields in the advanced rule configuration (in `/firewall_rules_edit.php?if=xxx`) together with "Max source states" such that the pf rule would end up like this:

    pass from XXX to YYY keep state (max-src-states <nnn>, adaptive.start <mmm>, adaptive.end <ooo>)

With an agressive start and end configuration, a user coming close to his Max source states will likely not experience packet drop because of the timeout becoming really short.

**Describe alternatives you considered**

Globally shortening TCP timeouts is not really an option, as the defaults are sane values. Tuning the global adaptative timeouts comes either with the problem that one src could impact all others on the network or the src with a lot of states experiencing packet losses; there seem not to be a setting were both can be satisfied without the per rule adaptative settings.

kulikov-a:
Hi. (im the one that answered on forum)
Are you sure that the default coefficients used by the `pf` to calculate the  `adaptive.start`, `adaptive.end` depending on the "Max states" value are not enough?
if you set "Max states" = 1000 in GUI it should give something like
`.. keep state (max 1000, adaptive.start 600, adaptive.end 1200) label ..` imho

ganguin:
I'm using max-src-states, not max states, but afaik it doesn't matter.

[...]

PR #5492, opened 2022-01-21
Firewall / Rules - allow per-rule adaptive timeouts

kulikov-a:
Hi!
may close https://github.com/opnsense/core/issues/5487 i hope.
No existing code touched except `FormSetAdvancedOptions` to allow '0' as a value of Advanced fields.
Thanks!

### DIFF
--- a/src/opnsense/mvc/app/library/OPNsense/Firewall/FilterRule.php
+++ b/src/opnsense/mvc/app/library/OPNsense/Firewall/FilterRule.php
@@ -232,6 +232,9 @@ class FilterRule extends Rule
                             $rule['state']['options'][] = $state_tag . " " . $rule[$state_tag];
                         }
                     }
+                    if (is_numeric($rule['adaptivestart']) && is_numeric($rule['adaptiveend'])) {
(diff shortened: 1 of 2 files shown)


AdSchellevis:
@kulikov-a some small cleanups in https://github.com/opnsense/core/commit/32831f9be2e03c44ca5b62014f593b612350c881  (pushed into a separate branch), if it's ok for you I'll merge it to master.

kulikov-a:
@AdSchellevis completely OK )
just two questions if i may: there will be no objections later that 0, strictly speaking, is not positive int?)
for the future PRs, do I understand correctly that it is recommended to validate a field in a single IF statement? (I did separate checks on the assumption that it makes sense to report multiple validation errors if applicable. like setting` adapative.start` to 0 end leave `max` and `adaptive.end` empty)

AdSchellevis:
@kulikov-a the legacy code is still messy when it comes to validations, placing more in `if` `elseif` makes it a bit easier to read, although still not perfect.  you're right about the positive int part, let me change that back. just tried to align the messages a bit, overlooked the zero part here.

AdSchellevis:
merged in https://github.com/opnsense/core/commit/b01da7a918ced0ab53871ce99bb2661db25cb650

kulikov-a:
@AdSchellevis thanks!

PR #5497, opened 2022-01-23
util: use severity levels

kulikov-a:
Hi!
I hope I didn't miss anything and now we can take advantage of using severity levels?
starting with GUI authentication logging )
Thanks!

### DIFF
--- a/src/etc/inc/authgui.inc
+++ b/src/etc/inc/authgui.inc
@@ -127,10 +127,10 @@ function session_auth()
 {
     global $config;
 
-    function auth_log($message)
+    function auth_log($message, $prio = LOG_ERR)
     {
         openlog("audit", LOG_ODELAY, LOG_AUTH);
-        log_error($message);
+        log_error($message, $prio);
         closelog();
     }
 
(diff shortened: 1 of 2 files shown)


AdSchellevis:
@kulikov-a thanks, it was on my list somewhere to look into this. There are probably more spots to inspect at some point, but this is a good starting point.

fichtner:
FWIW, I don’t think a „log_error“ function should be able to change priority of the message. Some log_error calls are warnings or otherwise but in the end touching those lines should be more clear than obfuscating them as log_error($msg, info);

kulikov-a:
@AdSchellevis Thanks! 
@fichtner @AdSchellevis 

> I don’t think a „log_error“ function should be able to change priority

Can we discuss options then? Maybe it would be more correct to add an additional function (say `log_msg()`) with the severity arg, then go through all the `log_error` calls and change it where appropriate?

AdSchellevis:
I agree the naming is indeed a bit unfortunate, but I also don't want to add functions for every level (growing legacy base with limited value). @kulikov-a your suggestion to wrap the current functionality in `log_msg()` and keep `log_error()` functionally equal sounds reasonable to me.

kulikov-a:
@AdSchellevis Great! (actually, i even started with this approach, but then I decided that the less code the better) )
will you revert pr commits or should i start from the current point?

fichtner:
Yes, generic log message function is better is all I’m saying.

AdSchellevis:
> @AdSchellevis Great! (actually, i even started with this approach, but then I decided that the less code the better) )
> will you revert pr commits or should i start from the current point?

Just stack the change on top, no need to revert the previous code

PR #5498, opened 2022-01-23
util/log_error: keep it clear. add a log_msg

kulikov-a:
Hi!
in https://github.com/opnsense/core/pull/5497 continuation
don't obfuscate `log_error`, add `log_msg()`
will try to traverse `log_error` calls and move to `log_msg`  where appropriate in next pr
thanks!

### DIFF
--- a/src/etc/inc/util.inc
+++ b/src/etc/inc/util.inc
@@ -928,12 +928,31 @@ function get_interface_list($only_active = false, $include_dmesg = false)
 * NAME
 *   log_error  - Sends a string to syslog.
 * INPUTS
-*   $msg     - string containing the syslog message.
-*   $prio    - syslog severity level
+*   $error     - string containing the syslog message.
 * RESULT
 *   null
 ******/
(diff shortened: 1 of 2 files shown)


src/etc/inc/util.inc
```diff
@@ -928,12 +928,31 @@ function get_interface_list($only_active = false, $include_dmesg = false)
 * NAME
 *   log_error  - Sends a string to syslog.
 * INPUTS
+*   $error     - string containing the syslog message.
+* RESULT
+*   null
+******/
+function log_error($error)
+{
```
AdSchellevis, line 936: how about just calling `log_msg($msg)` ? the internals are the same :)

kulikov-a: ) thanks!! setting LOG_ERR explicitly in case log_msg default will change

AdSchellevis:
@kulikov-a thanks!

kulikov-a:
@AdSchellevis thanks!!

PR #5505, opened 2022-01-26
pkg: avoid or force upgrades if already done, use VERSIONBIN force flag

NOYB:
pkg: avoid upgrades if already done, use VERSIONBIN

There should at least be a flag for this to continue with forced upgrade anyway.
Pass the flag as -D force


fichtner:
it's called "make install"

PR #5507, opened 2022-01-27
Firewall: Aliases - encode rule name in response

kulikov-a:
Hi!
Accidentally stumbled while playing with the pf table counters issue.
Since html special chars are allowed in the rules descriptons returning plain `sprintf` may lead to possible XSS vul. imho.
example:
Choose any pf rule using aliases, change rule descripton to ``, try to delete Alias associated with this rule

thanks!


AdSchellevis:

kulikov-a:
@AdSchellevis thanks!)

PR #5514, opened 2022-01-28
Fix typo causing error on IPv6 login

ppascher:


fichtner:
Merged, thanks a lot!

PR #5522, opened 2022-01-30
Logs: add backward compatibility

kulikov-a:
Hi.
For logs part of https://github.com/opnsense/plugins/issues/2776 and may close https://github.com/opnsense/plugins/issues/2786

set initial `severity` val to `None `and just add lines with missing severity (not set by parser or not in log) to the output. no fake prio - just empty field

Thanks!

### DIFF
--- a/src/opnsense/scripts/systemhealth/queryLog.py
+++ b/src/opnsense/scripts/systemhealth/queryLog.py
@@ -102,7 +102,7 @@ if __name__ == '__main__':
                             'timestamp': None,
                             'parser': None,
                             'facility': 1,
-                            'severity': 3,
+                            'severity': None,
                             'process_name': '',
                             'pid': None,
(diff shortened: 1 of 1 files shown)


AdSchellevis:
I think we can just change the default facility as suggested, looking at the code, it doesn't look like 3 is converted anywhere else.

kulikov-a:
@AdSchellevis sorry. tested little more (with proxy, dnscrypt etc). updated. how about that? )

AdSchellevis:
@kulikov-a works for me, no level is indeed better than a random one. merge this one? or do you want to play with it more?

kulikov-a:
@AdSchellevis i think im done) thanks!)

AdSchellevis:
@kulikov-a thanks!

kulikov-a:
Thanks!!

PR #5538, opened 2022-02-02
Firewall: Categories - do not select filtered-out rules

kulikov-a:
Hi!
`hook_firewall_categories()` function tries to prevent accidental rules mass actions by disabling `".rule_select"` checkboxes for filtered-out rules. But `prop()` not taking this in to account.

Thanks!


AdSchellevis:
@kulikov-a looks like a proper fix, thanks!

kulikov-a:
@AdSchellevis thanks!

PR #5541, opened 2022-02-04
Log: update severity level where appropriate

kulikov-a:
Hi!
Continuation of https://github.com/opnsense/core/pull/5497
log_msg() sets LOG_WARNING by default as discussed at https://github.com/opnsense/core/commit/279de63e63129e0b6c94fcbf4b944674bc22a38d#comments
change severity level where appropriate

Thanks!

### DIFF
--- a/src/etc/inc/util.inc
+++ b/src/etc/inc/util.inc
@@ -946,7 +946,7 @@ function log_error($error)
 * RESULT
 *   null
 ******/
-function log_msg($msg, $prio = LOG_ERR)
+function log_msg($msg, $prio = LOG_WARNING)
 {
     $page = $_SERVER['SCRIPT_NAME'];
     if (empty($page)) {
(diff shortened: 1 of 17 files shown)


fichtner:
Please let’s avoid the larger change before 1 months prior to 22.7. What we can do now is transform syslog() calls to log_msg() without those in config.inc

AdSchellevis:
@kulikov-a draft is fine, thanks for your patience

src/etc/inc/util.inc
```diff
@@ -946,7 +946,7 @@ function log_error($error)
 * RESULT
 *   null
 ******/
-function log_msg($msg, $prio = LOG_ERR)
+function log_msg($msg, $prio = LOG_WARNING)
```
fichtner, line 949: I think we should get this one single line in now rather than later.

kulikov-a: I don't see any problems ) do I have to make a pr for it or will you make the changes yourself and I'll just keep that in mind when updating this one closer to 22.7?

fichtner: if you have the time please make a PR, better to retain authorship if we can :)

kulikov-a: sure, thanks)

fichtner:
@kulikov-a if you can give this another try, see 931a851e09f 8b078c0ffd a7fd4381a2ee

fichtner:
Too much movement, I think we better close this.

kulikov-a:
@fichtner Sure, I'll try to find some time )

PR #5542, opened 2022-02-04
Unbound: overrides: fixup UI and validation due to model change

swhite2:
Addresses the change in https://github.com/opnsense/core/commit/308b31dfa84943c0d3487f23c53e8bb10a819222


AdSchellevis:
It seems I missed something in the review as well, the `rr` field needs to link it's dependancies to prevent changes to that field leading to inconsistencies. like for example:

https://github.com/opnsense/core/blob/ce656d076bdebb2d53688f877472b1c0e915c5db/src/opnsense/mvc/app/models/OPNsense/Firewall/Alias.xml#L42-L46

swhite2:
Good one, I don't think this is mentioned in the documentation though. Can I add multiple references inside a single check as a catch-all solution? As it stands I require 4 in total.

Issue #5513, opened 2022-01-28
MVC page using legacy config generation does not properly update service configuration


swhite2:
**Important notices**

Before you add a new report, we ask you kindly to acknowledge the following:

- [x] I have read the contributing guide lines at https://github.com/opnsense/core/blob/master/CONTRIBUTING.md
- [x] I am convinced that my issue is new after having checked both open and closed issues at https://github.com/opnsense/core/issues?q=is%3Aissue

**Describe the bug**

Right now the legacy restart/reconfigure(apply) buttons in the webgui do not function the same as the MVC restart/configure(apply) buttons. The MVC status widget actions and the reconfigure button assume that all configuration is generated via the template system, while the legacy restart/reconfigure actions configure *both* legacy and templated configuration data. Since it is sometimes necessary to prevent large pieces of logic cluttering up a template or minimizing impact, legacy config generation might be utilized instead, while the MVC model is used to do model validation and page rendering. Currently, this is the case for Unbound -> Overrides -> Host entries, which I migrated last week. 

The overall effect is that both applying and restarting the service from this MVC page (in this example Host Overrides) does not actually update the configuration in /var/unbound/* unless (from a users' perspective):
1. You restart the service from any legacy page.
2. You reboot the system.

Which in both case activate the relevant pluginctl hooks.

The question now becomes how we might do this correctly.

It seems https://github.com/opnsense/core/blob/master/src/opnsense/scripts/unbound/start.sh must always be called to properly start Unbound, which is called in both the MVC and legacy cases (https://github.com/opnsense/core/blob/master/src/etc/inc/plugins.inc.d/unbound.inc#L419 and https://github.com/opnsense/core/blob/master/src/opnsense/mvc/app/controllers/OPNsense/Base/ApiMutableServiceControllerBase.php#L187), so I suspect interference here is not necessary. I would say that overriding the reconfigureAction() method and calling a pluginctl hook in Unbounds' ServiceController might be a solution. This provides the added benefit of having a single legacy hook to signal that a transition in the future might be necessary. The downside is that it's not a generic solution - meaning one could add a host entry, switch to another MVC page for the same service and restart it from there, which would also not properly update the configuration.

Please correct me if I'm wrong in any of these statements @AdSchellevis @fichtner.

**To Reproduce**

1. Go to Unbound->Overrides->Host entries (default page)
2. Add a host entry
3. Click Apply and/or restart the service
4. Observe /var/unbound/host_entries.conf has not changed
[...]

PR #5553, opened 2022-02-08
unbound: host overrides: use legacy reconfigure and reduce configd calls

swhite2:
Fixes https://github.com/opnsense/core/issues/5513

Makes sure the 'Apply' button actually does something, since the host overrides are configured via the legacy section.

### DIFF
--- a/src/etc/inc/plugins.inc.d/unbound.inc
+++ b/src/etc/inc/plugins.inc.d/unbound.inc
@@ -416,7 +416,7 @@ function unbound_configure_do($verbose = false, $interface = '')
         flush();
     }
 
-    configd_run('unbound start', true);
+    mwexecf('/usr/local/bin/flock -n -E 0 -o /tmp/unbound_start.lock /usr/local/opnsense/scripts/unbound/start.sh');
 
     if (isset($config['unbound']['regdhcp'])) {
         $domain = $config['system']['domain'];
--- a/src/etc/inc/plugins.inc.d/unbound.inc
+++ b/src/etc/inc/plugins.inc.d/unbound.inc
@@ -416,7 +416,7 @@ function unbound_configure_do($verbose = false, $interface = '')
         flush();
     }
 
-    mwexecf('/usr/local/bin/flock -n -E 0 -o /tmp/unbound_start.lock /usr/local/opnsense/scripts/unbound/start.sh');
+    mwexec('/usr/local/bin/flock -n -E 0 -o /tmp/unbound_start.lock /usr/local/opnsense/scripts/unbound/start.sh');
 
     if (isset($config['unbound']['regdhcp'])) {
         $domain = $config['system']['domain'];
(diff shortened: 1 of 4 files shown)


src/opnsense/service/conf/actions.d/actions_unbound.conf
```diff
@@ -83,3 +83,10 @@ command:pgrep -q -nF /var/run/unbound.pid && echo "unbound is running" || echo "
 parameters:
 type:script_output
 message:request Unbound status
+
+[reconfigure]
+command:pluginctl -s unbound restart
+parameters:
+type:script
+message:Configuring and restarting Unbound
+description:Configure and restart Unbound via legacy code
```
fichtner, line 92: any reason we need a cron job (description set)?

swhite2: We don't, misinterpretation on my end :)

src/etc/inc/plugins.inc.d/unbound.inc
```diff
@@ -416,7 +416,7 @@ function unbound_configure_do($verbose = false, $interface = '')
         flush();
     }
 
-    configd_run('unbound start', true);
+    mwexecf('/usr/local/bin/flock -n -E 0 -o /tmp/unbound_start.lock /usr/local/opnsense/scripts/unbound/start.sh');
```
fichtner, line 419:
```suggestion
    mwexec('/usr/local/bin/flock -n -E 0 -o /tmp/unbound_start.lock /usr/local/opnsense/scripts/unbound/start.sh');
```
FWIW ;)

swhite2: :neutral_face: I need a coffee.

swhite2:
@AdSchellevis The latest commit makes a lot more sense.

src/opnsense/service/conf/actions.d/actions_unbound.conf
```diff
@@ -41,21 +41,19 @@ type:script
...
+command:pluginctl -s unbound stop
 parameters:
 type:script
 message:Stopping Unbound
 
 [restart]
-command:
-  /usr/local/sbin/unbound-control -c /var/unbound/unbound.conf stop &&
-  /usr/local/bin/flock -n -E 0 -o /tmp/unbound_start.lock /usr/local/opnsense/scripts/unbound/start.sh
+command:pluginctl -s unbound stop && pluginctl -s unbound start
```
fichtner, line 56: We want to start even if stop fails

fichtner: Ps don’t we have “restart” too?

swhite2: You're right

fichtner:
what I like about this is that unbound_start.lock is now only used from within unbound.inc and only once at that 👍

fichtner:
@swhite2 now after this unbound takes a long time to become available on boot. disabled my blocklist to see if that is part of the cause or not. can you take a closer look too? ideally unbound should be available much sooner.

swhite2:
> @swhite2 now after this unbound takes a long time to become available on boot. disabled my blocklist to see if that is part of the cause or not. can you take a closer look too? ideally unbound should be available much sooner.

Interesting. I'll take a look.

PR #5566, opened 2022-02-11
dns: fix resolving DNS names

oshogbo:
The drill package is installed by ldns packages.
Because of that it is in /usr/local/bin not in /usr/bin.

Sponsored by: DynFi


AdSchellevis:
If I'm not mistaken we also ship `/usr/bin/drill` as part of base as this seems to be working on our end.  `diag_dns.php` is quite messy so if it works, we better aim for replacing it someday.

fichtner:
ldns is actually just a coincidence for having openssh-portable installed so we can't rely on /usr/local/bin/drill to be there at the moment as it might change. The obvious fix is to add ldns to dependencies but since we already have a binary in the base I think that isn't needed at all. :)


Cheers,
Franco

PR #5568, opened 2022-02-11
Fix implode() call

clementmoulin:
According to https://www.php.net/manual/en/function.implode.php and PHP warning since 7.4.0:
Exception: Error at /usr/local/etc/inc/plugins.inc.d/dhcpd.inc:2012 - implode(): Passing glue string after array is deprecated. Swap the parameters (errno=8192) in /usr/local/opnsense/mvc/app/controllers/OPNsense/Base/ApiControllerBase.php:96


AdSchellevis:
@clementmoulin thanks! looks like a typo indeed.

PR #5570, opened 2022-02-12
sidebar - optimization

opnsenseuser:
Removed unnecessary code for better performance


AdSchellevis:
@opnsenseuser less is more :) thanks!

Issue #3657, opened 2019-08-18
DHCPv6: static mapping with prefix tracking leads to wrong DNS AAAA entry


klada:
**Important notices**
Before you add a new report, we ask you kindly to acknowledge the following:

[*] I have read the contributing guide lines at https://github.com/opnsense/core/blob/master/CONTRIBUTING.md

[*] I have searched the existing issues and I'm convinced that mine is new.

**Describe the bug**

According to the DHCPv6 settings page a static mapping for a network with IPv6 prefix tracking can be entered in the format `::1:2:3.4`.

The static record itself is working for DHCPv6, but the resulting AAAA-DNS records are wrong - the (tracked) IPv6 prefix is missing from them:

```bash
$ host test.lan
test.lan has address 192.168.1.5
test.lan has IPv6 address ::1:2:3:4
```

This essentially breaks all communication with the host, as IPv6 is usually preferred.

**To Reproduce**
1. Enable IPv6 prefix tracking for an interface
1. Enable Unbound DNS (with option "register DHCP static mappings")
2. Add a static DHCPv6 entry for "test.lan", omitting the prefix (just as suggested by the documentation)
3. Apply changes
4. This results in a broken DNS AAAA record for "test.lan"

**Expected behavior**

The DNS-AAAA record for the static host should contain the current prefix and the suffix configured in the static DHCPv6 mapping.

**Environment**
Software version used and hardware type if relevant.
e.g.:

OPNsense OPNsense 19.7.2 (amd64, OpenSSL).
APU2C4
Network Intel® I210-AT

fichtner:
Hmm, this is an Unbound/Dnsmasq issue then. Problem is that prefix merging works in DHCPv6 service but there is no great way of knowing without creating external plumbing.

klada:
@fichtner Wouldn't a script which runs on each prefix change be enough? This script could generate a correct unbound/dnsmasq config containing the full IPv6 addresses of static host mappings.

This would also be an important step for proper IPv6 firewall rules in environments with dynamic prefixes, as then we could use DNS aliases for IPv6 hosts.

poggenpower:
Interestingly the `/var/dhcpd/etc/dhcpdv6.conf` contains the full address, but the `core/src/etc/inc/plugins.inc.d/unbound.inc` reads the leases from `$config['dhcpdv6']` which doesn't seem complete. 
If there is information about the interface in the `$config['dhcpdv6']` tree, than the address could be reconstructed from the interface. 
[...]

Issue #2544, opened 2018-07-14
firewall: global prefix in rules needs to be changed manually when IPv6 upstream prefix changes


RyuunoAelia:
While playing a bit with NPTv6 I found a a workflow issue. The main and only reason to use NPTv6 is to avoid changing your local configuration (firewall rules, etc) when your upstream changes.

However, as things are done in opnsense right now, you need to manually update the NPTv6 rule each time the upstream changes.

This is fine for cases where you have complete control over when the prefix change occurs (multi-homed setup, etc). However, when the prefix may change without your knowledge (say a 6rd setup, or ISP providing ipv6 connectivity with short lived DHCP leases), having to manually update the rule painful.

I understand selecting the IPv6 might be problematic to implement in general. An interface may have multiple IPv6 at a point in time (even if you ignore link-local addresses). So maybe having a way to select on an interface if this interface will have dynamic NPTv6 rules might be a more straightforward flow for setting up things. But since I do not know much about the internals of the UI, I cannot give a pertinent insight.

My full context is described in #2538.

fichtner:
This is also useful for DHCPv6 and SLAAC, but will require more tinkering.

fichtner:
Not for the rules currently. It doesn't allow tracking the interface prefix change.

marjohn56:
So maybe we try splitting the rules into prefix/suffix..,, Alias for Prefix+suffix. It does make it complex for the user too, but that's the problem with a changing prefix, it's not for the faint-hearted.

fichtner:
We already have a larger change in that area to make 6RD usable for "inet6" type rules and need to see if that is alright before we stack another layer on top, but generally an automatic prefix adaption option should be possible

chris42:
Having similar troubles discussed with marjohn56 in https://forum.opnsense.org/index.php?topic=9438.0

Being completely obvious on how technically the DHCPv6, SLAAC, RADV and the firewall works: Since the clients in the network negotiate their IPv6 couldn't Opnsense create a central register in the DNS, that is build during the negotiation process? This would keep the changing IP information centralized and up to date in realtime. (This would cover prefix changes as well as changes by e.g. privacy extensions)
Then you could use that register with Hostname in the firewall rules?

Basically using the Alias, but not having a 5min delay between resolving the hostname, but accessing the register (which should be up-to-date) of IPv6s in the network?

fichtner:
[...]

PR #5574, opened 2022-02-13
unbound: Add support for native IPv6 prefix tracking in host overrides

n-thumann:
This PR add support for native tracking of dynamic IPv6 prefixes in Unbound host overrides.
It allows one to be using SLAAC provide the clients with a network prefix, let them auto-configure on their own and only add the client's IID and the interface to track to Unbound.

It is similar to Unbounds DHCP Static Mapping, but without requiring DHCPv6. Personally, I don't run DHCPv6 in my network and hence cannot use the static mappings and would also like to avoid cross-dependencies between DHCPv6 and Unbound.

With this patch applied you can optionally add an interface to track to an AAAA record:
Screenshot of Editing host override

The table also shows the interface, `LAN` in this example, because it's configured as a Track Interface of WAN.
Screenshot of host overrides

The host overrides as seen above result in the following `host_entries.conf` (see last two lines):
```
root@OPNsense:~ # cat /var/unbound/host_entries.conf
[... snip ...]
local-data-ptr: "2003:[redacted]:5076:9bff:fe04:e801 OPNsense.localdomain"
local-data: "OPNsense.localdomain AAAA 2003:[redacted]:5076:9bff:fe04:e801"
local-data: "OPNsense AAAA 2003:[redacted]:5076:9bff:fe04:e801"
local-data: "OPNsense.localdomain AAAA fe80::5076:9bff:fe04:e801"
local-data: "OPNsense AAAA fe80::5076:9bff:fe04:e801"
local-data-ptr: "1.1.1.1 some-a-host.domain"
local-data: "some-a-host.domain IN A 1.1.1.1"
local-data-ptr: "2606:4700:4700::1111 some-aaaa-host.domain"
local-data: "some-aaaa-host.domain IN AAAA 2606:4700:4700::1111"
local-data-ptr: "2003:[redacted]::1337 tracked-prefix-host.domain"
local-data: "tracked-prefix-host.domain IN AAAA 2003:[redacted]::1337"
```

### DIFF
--- a/src/etc/inc/plugins.inc.d/unbound.inc
+++ b/src/etc/inc/plugins.inc.d/unbound.inc
@@ -552,13 +552,21 @@ function unbound_add_host_entries($ifconfig_details = null)
                     switch ($host->rr) {
                         case 'A':
                         case 'AAAA':
+                            if (!empty((string) $host->interface) && strpos($host->server, "::") === 0) {
+                                list ($address) = interfaces_primary_address6((string) $host->interface, $ifconfig_details);
+                                $server = make_ipv6_64_address($address, $host->server);
+                                $server = Net_IPv6::compress(Net_IPv6::uncompress($server));
+                            } else {
+                                $server = $host->server;
+                            }
+
                             /* Handle wildcard entries which have "*" as a hostname. Since we added a . above, we match on "*.". */
                             if ($alias['hostname'] == '*.') {
                                 $unbound_entries .= "local-zone: \"{$alias['domain']}\" redirect\n";
-                                $unbound_entries .= "local-data: \"{$alias['domain']} IN {$host->rr} {$host->server}\"\n";
+                                $unbound_entries .= "local-data: \"{$alias['domain']} IN {$host->rr} {$server}\"\n";
                             } else {
-                                $unbound_entries .= "local-data-ptr: \"{$host->server} {$alias['hostname']}{$alias['domain']}\"\n";
-                                $unbound_entries .= "local-data: \"{$alias['hostname']}{$alias['domain']} IN {$host->rr} {$host->server}\"\n";
+                                $unbound_entries .= "local-data-ptr: \"{$server} {$alias['hostname']}{$alias['domain']}\"\n";
+                                $unbound_entries .= "local-data: \"{$alias['hostname']}{$alias['domain']} IN {$host->rr} {$server}\"\n";
                             }
                             break;
                         case 'MX':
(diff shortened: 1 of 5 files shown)


AdSchellevis:
To be honest, I'm not really a fan of trying to glue more interface address dependancies in the unbound code as it's usually quite flaky and prone to errors.

If we have an option to cleanly feed this to unbound without the need of a restart (loosely coupled), it might be different, but at the moment I don't think we should try to merge this. @swhite2 is doing some more work around Unbound, parking this with him for now to see if we can come up with a design change at some point for these dynamic types.

fichtner (commented):
Just from the looks of it the interface selection feels really odd here. It's not even the prefix merge that trips me off. I agree this needs some sort of scope to work but I would not favour Unbound integration to be the one to pioneer it.

swhite2:
> If we have an option to cleanly feed this to unbound without the need of a restart (loosely coupled), it might be different, but at the moment I don't think we should try to merge this. @swhite2 is doing some more work around Unbound, parking this with him for now to see if we can come up with a design change at some point for these dynamic types.

Agreed. This especially pertains to a requirement for dynamic types in the mvc model.

n-thumann:
Thanks for your feedback!

> To be honest, I'm not really a fan of trying to glue more interface address dependancies in the unbound code as it's usually quite flaky and prone to errors.

For that reason I used the same logic as in https://github.com/opnsense/core/blob/0b453ea98dc21d5c8f7ab2946e4867b5cd4528ba/src/etc/inc/plugins.inc.d/unbound.inc#L474, because if it works at that point, we can probably assume that it works in this patch as well.

> If we have an option to cleanly feed this to unbound without the need of a restart (loosely coupled), it might be different [...]

I definitely agree that this should be the ultimate goal! But given the fact that the problem in general has been around for quite some time (e.g. https://github.com/opnsense/core/issues/2544#issuecomment-405078997 or https://github.com/opnsense/core/issues/3657#issuecomment-560150755), this may be a good solution in the meantime.
Also, it doesn't require any additional restarts of Unbound, because it's already being restarted when the upstream prefix changes to update its own DNS record.

AdSchellevis:
The problem with adding more similar problems that are already in there makes it eventually harder to fix the real problem (which is the entanglement of the service). I do agree that similar constructs do exist within the same module, but that doesn't mean we should increase the maintenance  problem which someone will have to fix at some point in time. 

As @fichtner already noted, it's also not certain we are trying to fix the right thing here. (I only looked at the technical implications so far)

froogl:
BTW: If you need prefix-aware DNS-hosts you could just use dnsmasq. It has native support for this (see `--dynamic-host` feature). It's just lacking a GUI option in OPNsense right now so maybe it would be wiser to put more energy into the dnsmasq UI for this particular use-case.

TheOfficialMrBlah:
Hi,

what is a good and easy workaround, as a replacement for this PR?
So how do I create a host override with a Dynamic Prefix in Unbound?

I face the same problem that I have a new prefix every day.

AdSchellevis:
closing, currently not a core priority and doesn't fit unbound's architecture very well.

PR #5583, opened 2022-02-17
unbound: background potentially long-running operation

swhite2:
See https://github.com/opnsense/core/pull/5553


fichtner:
Merged, thanks!

PR #5587, opened 2022-02-21
Logs: prevent possible racing

kulikov-a:
Hi!
ref. https://forum.opnsense.org/index.php?topic=26867.msg131115#msg131115
it seems that the user rapidly changing the severity filter may cause some race condition, especially if a large rowCounter value is selected.
pr disable severity filter change until the results of the previous query are received

and I have another question: since now the log sizes have increased, choosing the value "All" for row count can lead to receiving such an amount of data that the uibootgrid plugin is not able to pass. perhaps it makes sense to limit the choice to some reasonable value (5K?)?

thanks!

### DIFF
--- a/src/opnsense/mvc/app/views/OPNsense/Diagnostics/log.volt
+++ b/src/opnsense/mvc/app/views/OPNsense/Diagnostics/log.volt
@@ -74,6 +74,7 @@
           search:'/api/diagnostics/log/{{module}}/{{scope}}'
       });
       $("#severity_filter").change(function(){
+          $('#severity_filter').attr('disabled',true);
           if (window.localStorage) {
               localStorage.setItem('log_severity_{{module}}_{{scope}}', $("#severity_filter").val());
           }
@@ -97,6 +98,10 @@
                   $("#severity_filter").val("Debug").change();
               }
           });
+          setTimeout(function() {
+              $('#severity_filter').attr('disabled',false);
+              $('#severity_filter_container').find('a[role="option"]').css('cursor', '');
+          }, 500);
       });
 
       $("#flushlog").on('click', function(event){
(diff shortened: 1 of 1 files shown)


AdSchellevis:
Hi Alex,

I'm not sure if the template changes are needed if we limit the maximum results, which is likely a good idea anyway. The export might have a bit higher limit, but that shouldn't be an issue if we lower the maximum here:

https://github.com/opnsense/core/blob/784ebd8d27874ddc09a99726ca0cbe41cc081611/src/opnsense/mvc/app/controllers/OPNsense/Diagnostics/Api/LogController.php#L62

Best regards,

Ad

kulikov-a:
Hi!)
I was thinking about a simple "-1" to "5000" replacement in 
[              rowCount:[20,50,100,200,500,1000,-1],](https://github.com/opnsense/core/blob/f64dfb8976ad3a87a6c402571a0810cb12f7a7f7/src/opnsense/mvc/app/views/OPNsense/Diagnostics/log.volt#L54)

it should not affect export imho (since it not setting row count and returns result as attachment. 96MB/500Krows file downloaded wihout issues)

queryLog.py script works very efficiently with huge log files (tried to limit RLIMIT_CPU for it and it needs veeery litle value to trigger signal)

AdSchellevis:
it's probably safer to limit -1 as the stored settings may overwrite the list items if I'm not mistaken.

kulikov-a:
ah, yes, localStorage, thanks ) should i add this to this pr or make add new one for rowCount limits?

AdSchellevis:
I think only forcing a limit should be enough, so a clean PR would be easier

kulikov-a:
replaced by https://github.com/opnsense/core/pull/5589

PR #5589, opened 2022-02-21
Logs: limit rowCount

kulikov-a:
Hi!
replaces https://github.com/opnsense/core/pull/5587
Hope I got the idea right.
Although I still see some race condition when changing the severity filter quickly (without locking it on change), at least it doesn't completely block gui and doesn't drain php memory.
Thanks!


AdSchellevis:
@kulikov-a thanks! this should prevent the endless waits indeed.

kulikov-a:
@AdSchellevis Thanks!!

Issue #5138, opened 2021-08-02
Unbound custom forwarders when Forwarding Mode Enabled.


FingerlessGlov3s:
**Important notices**

Before you add a new report, we ask you kindly to acknowledge the following:

- [X] I have read the contributing guide lines at https://github.com/opnsense/core/blob/master/CONTRIBUTING.md
- [X] I am convinced that my issue is new after having checked both open and closed issues at https://github.com/opnsense/core/issues?q=is%3Aissue

**Is your feature request related to a problem? Please describe.**

Back in 21.1 and prior there was a custom options option in unbound, this had now been removed. If you want forwards all DNS requests of unbound to say DNSCrypt for an example, this can no longer be done from within the WebUI. I think you should be able to do this without the requirement for a user to create a custom .conf file in "/usr/local/etc/unbound.opnsense.d/" to enable use of different DNS server(s) upstream when using unbound. You may not always want to use the same System DNS as you do for upstream unbound.

Example of contents of old custom options field, now required in custom file via CLI, when using DNSCrypt 
```
forward-zone:
  name: "."
  forward-addr: 127.0.0.1@15353
```

**Describe the solution you like**

Very simple extra field adding to "DNS Query Forwarding".
A text entry box, when left empty it'll use the DNS servers list from "System: Settings: General" when filled in it'll use those IPs listed for forwarding, making sure custom DNS ports are useable as well. So then you can specify "127.0.0.1@15353" as a DNS server to forward to, which could would solve the issue of any custom options required, making it much more user friendly.

**Describe alternatives you considered**

Using custom .conf file in "/usr/local/etc/unbound.opnsense.d/", work but I think this extra option would benefit many users.

**Additional context**

Screenshot of the idea
![image](https://user-images.githubusercontent.com/3495647/127926180-44a0e3b3-e6af-43ad-ba1b-a655edf1b9d0.png)

FingerlessGlov3s:
One to make note of is if 127.0.0.1 is trying to be forwarded too `do-not-query-localhost: no` is required to allow this.

fichtner:
@FingerlessGlov3s 2fa400beca is already in 21.7.1 and 580e6c7cc8107 is on the way

I still have trouble wrapping my head around a simple way to provide a forwarder interface that supports DoT, internal forwards and manual ones... it's a lot of modes that do almost the same (especially config-wise)

fichtner:
To be shipped in 22.1.5 via bab7883c5674

PR #5606, opened 2022-03-01
unbound: implement custom forwarders over current dot setup

swhite2:
resolves https://github.com/opnsense/core/issues/5138

This PR pulls query forwarding over the current dot setup, so visually nothing changes.

All API calls are redirected to new Forward functions, which slightly modifies what is returned based on whether "Query Forwarding" or "DNS over TLS" is selected from the menu. This way backwards compatibility is preserved.

As an addition, a user is now able to specify a specific domain for a forward zone as well. Meaning that queries for this specific domain will skip a catch-all (".") domain (if specified), and instead use the server specified for this domain.

Entering a forward zone with a catch-all domain (".") in both Query Forwading and DNS over TLS is considered a duplicate by Unbound, so a static warning for this has been attached in the grid - however, it might be possible for a user to be warned dynamically over this.

### DIFF
--- a/src/opnsense/mvc/app/controllers/OPNsense/Unbound/DotController.php
+++ b/src/opnsense/mvc/app/controllers/OPNsense/Unbound/DotController.php
@@ -35,6 +35,7 @@ class DotController extends IndexController
     public function indexAction()
     {
         $this->view->selected_forward = "";
+        $this->view->forwardingForm = $this->getForm('forwarding');
         $this->view->formDialogEdit = $this->getForm('dialogDot');
         $this->view->pick('OPNsense/Unbound/dot');
     }
--- a/src/opnsense/mvc/app/controllers/OPNsense/Unbound/DotController.php
+++ b/src/opnsense/mvc/app/controllers/OPNsense/Unbound/DotController.php
@@ -32,9 +32,9 @@ use OPNsense\Base\IndexController;
 
 class DotController extends IndexController
 {
-    public function indexAction($selected = null)
+    public function indexAction()
     {
-        $this->view->selected_forward = $selected;
+        $this->view->selected_forward = "";
         $this->view->formDialogEdit = $this->getForm('dialogDot');
         $this->view->pick('OPNsense/Unbound/dot');
     }
--- a/src/opnsense/mvc/app/models/OPNsense/Unbound/Migrations/M1_0_2.php
+++ b/src/opnsense/mvc/app/models/OPNsense/Unbound/Migrations/M1_0_2.php
@@ -1,7 +1,7 @@
 <?php
 
 /*
- * Copyright (C) 2021 Michael Muenz <m.muenz@gmail.com>
+ * Copyright (C) 2022 Deciso B.V.
  * All rights reserved.
  *
  * Redistribution and use in source and binary forms, with or without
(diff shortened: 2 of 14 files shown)


src/opnsense/mvc/app/models/OPNsense/Unbound/Migrations/M1_0_2.php
```diff
@@ -0,0 +1,56 @@
...
+    public function run($model)
+    {
+        $config = Config::getInstance()->object();
+
+        $node = $config->OPNsense->unboundplus;
+
+        if (!empty($node->dots) && !empty($node->dots->dot)) {
+            foreach ($model->dots->dot->iterateItems() as $dot) {
+                $dot->type = "dot";
+                $dot->domain = "";
```
AdSchellevis, line 52: should be empty by default.

src/opnsense/mvc/app/controllers/OPNsense/Unbound/DotController.php
```diff
@@ -32,8 +32,9 @@
 
 class DotController extends IndexController
 {
-    public function indexAction()
+    public function indexAction($selected = null)
```
AdSchellevis, line 35: looking at this again, it might be cleaner to use two controllers `DotController.php` and `ForwardController.php` so the url's are a bit more logical (`/ui/unbound/dot/` and `/ui/unbound/forward/` ads you can omit the index part)

AdSchellevis (changes requested):
looks good, with some small remarks. we might need to merge this in a new branch for release engineering, but let's ask @fichtner when ready.

src/opnsense/mvc/app/controllers/OPNsense/Unbound/DotController.php
swhite2: Makes sense, trying to keep the change as minimal as possible was the reason I did it this way with this drawback as a result.

src/opnsense/mvc/app/models/OPNsense/Unbound/Migrations/M1_0_2.php
swhite2: see https://github.com/opnsense/core/pull/5606/commits/2b63edd076871569671591707debc97d47c96c44

src/opnsense/mvc/app/controllers/OPNsense/Unbound/DotController.php
swhite2: see https://github.com/opnsense/core/pull/5606/commits/2b63edd076871569671591707debc97d47c96c44

swhite2:
@AdSchellevis Currently "DNS Query Forwarding" in Unbound -> General is always preferred over "Query Forwarding" and DoT. If we keep it this way, it might make sense to mention this statically in the forwading and DoT pages. This, in my opinion, also removes the need for a mention in the General page. 

What do you think? We can just as easily change the order - this makes sense if we were to prefer DoT over anyting else. 

Also, to prevent general confusion it might make sense to rename "Query forwarding" to "Custom forwarding". This hints more to the fact that specific domains can be specified (which are always used regardless of place in config relative to the "catch-all" forward zone).

AdSchellevis:
> @AdSchellevis Currently "DNS Query Forwarding" in Unbound -> General is always preferred over "Query Forwarding" and DoT. If we keep it this way, it might make sense to mention this statically in the forwading and DoT pages. This, in my opinion, also removes the need for a mention in the General page.

I'm not exactly sure which mention you mean in the General page, but since "DNS Query Forwarding" effectively seems to render both forwarding options useless it does seem to make sense to add a note.

> What do you think? We can just as easily change the order - this makes sense if we were to prefer DoT over anyting else.

I would leave the order as is for now, since you can configure the same options manually with your new forwarding page.

> Also, to prevent general confusion it might make sense to rename "Query forwarding" to "Custom forwarding". This hints more to the fact that specific domains can be specified (which are always used regardless of place in config relative to the "catch-all" forward zone).

Sounds like a good suggestion indeed.

swhite2:
Added documentation: https://github.com/opnsense/docs/commit/ba466f2c36f7167eafcf75b2d1103fde0c1df89f (also includes documentation for DoT)

src/opnsense/scripts/system/nameservers.php
```diff
@@ -0,0 +1,26 @@
+#!/usr/local/bin/php
+<?php
+
```
fichtner, line 3: needs a copyright

fichtner:
Merged, thanks!

PR #5615, opened 2022-03-05
Improve configctl help output - cosmetic change

juliocbc:
A cosmetic change. It will reflect the configctl (instead of configd_ctl.py) command output while invoking  /usr/local/sbin/configctl -h for example.

### DIFF
--- a/src/sbin/configctl
+++ b/src/sbin/configctl
@@ -5,4 +5,4 @@ if [ ! -e /var/run/configd.pid ]; then
 	exit 1
 fi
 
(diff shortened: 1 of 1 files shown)


fichtner:
This isn’t a great idea as it tampers with the final return code and probable output. It can be fixed within configd_ctl.py I think.

juliocbc:
OK! I avoided altering the configd_ctl.py because its output must keep the same as the script (configd_ctl.py). In this case, I will close this PR.

Thank you @fichtner

PR #5618, opened 2022-03-07
Add a cron task for scrubbing zfs pools

iain-henderson:
Easily manage ZFS maintenance from the GUI.

### DIFF
--- a/src/opnsense/service/conf/actions.d/actions_zfs.conf
+++ b/src/opnsense/service/conf/actions.d/actions_zfs.conf
@@ -1,0 +1,7 @@
+[scrub]
+description:Scrub ZFS Pool 
+command:/sbin/zpool scrub 
+parameters:%s
+type:script
+message:Scrubbing ZFS Pool %s
+


AdSchellevis:
Probably better to add as a plugin as from core we're not planning to document the use-case, in which case it would at some point raise support questions we don't plan to answer.

gsuberland:
Was this ever added as a plugin? I don't see it as an option in the cron menu, and I can't see any matching plugins. ZFS volumes will degrade without a periodic scrub, so this is quite important.

fichtner:
You can create any sort of cron task you like. ZFS related additions are being discussed but need time to be designed and implemented.

https://docs.opnsense.org/development/backend/configd.html

iain-henderson:
Adding cron tasks isn't quite as simple as it could be, but I never circled back around to set this up as a plug-in.

src/opnsense/service/conf/actions.d/actions_zfs.conf
```diff
@@ -0,0 +1,7 @@
+[scrub]
+description:Scrub ZFS Pool 
+command:/sbin/zpool scrub 
+parameters:%s
+type:script
+message:Scrubbing ZFS Pool %s
+
```
fichtner, line 7:
```suggestion
message:Scrubbing ZFS Pool %s
```

fichtner:
Merged to core after internal discussion. Thanks!

PR #5619, opened 2022-03-07
dhcpd: add support for client-updates option

Jacob-Coyle:
Adds support for the client-updates option for DDNS updates in dhcpd.

Can specify whether client updates should be allowed (default), denied, or ignored. dhcpd defaults to "allow client-updates" if nothing is specified, so that was chosen as the default to ensure any existing configs with DDNS continue to work as expected.

If DDNS updates are already configured on the DHCP subnet, then no changes will be made until the next time the DHCP config is save for that interface.


AdSchellevis:
closing for now, due to `isc-dhcp` deprecating we prefer a feature freeze on this component for now.

Issue #5622, opened 2022-03-09
OpenVPN - Please add option to change keepalive checks, change inactivity timeouts and increase buffer sizes.


tomsparks:
- [X] I have read the contributing guide lines at https://github.com/opnsense/core/blob/master/CONTRIBUTING.md
- [X] I am convinced that my issue is new after having checked both open and closed issues at https://github.com/opnsense/core/issues?q=is%3Aissue

**Is your feature request related to a problem? Please describe.**
I'm supporting a number of users across the US and one datacenter has issues that are remedied with these fixes.
I'm a user of both PFSense and OPNsense and PFSense supports most of these settings.

Disabling compression is unclear. It should be denoted as "Disabled". 
There is an option for disabling LZO - "Legacy - Disabled LZO algorithm"
I only found this out after reading the code. I believe I've found the correct way to remedy this.

Changing the keepalive check period is needed for lossier networks, PFSense supports this and it
remedies the problem. 
The current default is:
keepalive 10 60
I would like the ability to change it to:
keepalive 5 30

Increasing the buffer sizes helps maintain thruput, PFSense supports this and adds the following configuration:
sndbuf 2097152
rcvbuf 2097152

There is an additional tweak which increases the buffer settings on the client:
push "sndbuf 512000"
push "rcvbuf 512000"

Currently there is no inactivity timeout option, I'd like to be able to set it to 20 minutes, this is the config output:
inactive 1200

**Describe alternatives you considered**
Currently I can change some of these options via the "Advanced" textbox, however this is un ideal since typos in the config can cause problems/outages.

tomsparks:
I just sent in a pull request that will add some of this.
Glad I was able to do that.

PR #5623, opened 2022-03-09
Pr5622

tomsparks:
This adds several of the features I requested in Issue #5622.
I've tested this on one of my opnsense boxes and confirmed it works.
I have not added the inactivity timer or the client push since I need to do more testing against those features.

### DIFF
--- a/src/etc/inc/plugins.inc.d/openvpn.inc
+++ b/src/etc/inc/plugins.inc.d/openvpn.inc
@@ -161,6 +161,14 @@ function openvpn_buffer_settings()
     );
 }
 
+function openvpn_keepalive_settings()
+{
+    return array(
+        'standard' => gettext('Standard'),
+        'short' => gettext('Short'),
+        'long' => gettext('Long'),
+    );
+}
 
 function openvpn_get_protocols()
 {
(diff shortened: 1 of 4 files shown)


src/etc/inc/plugins.inc.d/openvpn.inc
```diff
@@ -150,6 +151,25 @@ function openvpn_compression_modes()
     );
 }
 
+function openvpn_buffer_settings()
+{
+    return array(
```
AdSchellevis, line 156: Do you have some performance metrics about the impact of this feature? In my experience sofar on FreeBSD increasing socket send and receive buffers don't impact performance as one would expect. Adding fine grained options like these only make sense if we are able to explain in which circumstances changing the options helps.

When there's a generic issue around send and receive buffers, the sysctl parameters (net.inet.tcp.recvspace, net.inet.tcp.sendspace) might be more suitable to increase (https://www.freebsd.org/cgi/man.cgi?query=tuning)

AdSchellevis (changes requested):
some concerns about the usefulness of sndbuf/recvbuf configuration and some issues about input type and minor bugs.

src/etc/inc/plugins.inc.d/openvpn.inc
mimugmail: I had a similar case where company is transfering video files heavily. At first the tunnel was set to TCP and no client was able to transfer more than 15Mbps, then we added a second instance with UDP, transfer rates grew to 25Mbps. After this we added to advanced field:
```
sndbuf 393216
rcvbuf 393216
```
Now clients are able to transfer at around 70Mbps .. which is quite the max of vdsl you get in our area. 
It was tested with 21.7 as the base, the system is not updated to 22.1 yet.

AdSchellevis:
unmergable, missing feedback

Issue #5577, opened 2022-02-14
unbound-anchor: prioritize system-wide DNS servers over public roots


marinbernard-pep06:
**Important notices**

Before you add a new report, we ask you kindly to acknowledge the following:

- [x] I have read the contributing guide lines at https://github.com/opnsense/core/blob/master/CONTRIBUTING.md
- [x] I am convinced that my issue is new after having checked both open and closed issues at https://github.com/opnsense/core/issues?q=is%3Aissue

**Is your feature request related to a problem? Please describe.**

We use the unbound resolver in forwarding mode on several OPNsense boxes hosted in internal networks. These unbound instances are used as caching DNS servers, and have no connectivity to the root DNS servers living on the Internet.

At unbound startup, `unbound-anchor` is run to update DNS roots. By default, unbound-anchor tries to contact DNS root servers. Since the box has no connectivity to foreign DNS servers, unbound startup is delayed until unbound-anchor gives up (timeout).

In such a context, running `unbound-anchor` with the `-Rf /etc/resolv.conf` argument is appropriate, as it would prioritize system-wide DNS servers from `resolv.conf` (which typically are upstream servers) over public roots to update the keys, while still allowing falling back to public roots on failure.

This is the only way to perform successful anchor updates at startup in offline network environments such as ours, and avoid a 60-second delay each time the unbound process is restarted.

**Describe the solution you like**

Modify unbound startup script at `/usr/local/opnsense/scripts/unbound/start.sh` to make unbound-anchor use the servers set in `resolv.conf`, still allowing a fallback to public roots on resolution failure:

```
--- /usr/local/opnsense/scripts/unbound/start.sh        2022-01-27 11:25:21.000000000 +0100
+++ start.sh    2022-02-14 13:45:50.232509000 +0100
@@ -30,7 +30,7 @@
 
 # prepare and startup unbound, so we can easily background it
 
-chroot -u unbound -g unbound / /usr/local/sbin/unbound-anchor -a /var/unbound/root.key
+chroot -u unbound -g unbound / /usr/local/sbin/unbound-anchor -a /var/unbound/root.key -Rf /etc/resolv.conf
 
 if [ ! -f /var/unbound/unbound_control.key ]; then
     chroot -u unbound -g unbound / /usr/local/sbin/unbound-control-setup -d /var/unbound
```

**Describe alternatives you considered**

Replace the startup script with our own amended version. But it'd be more future-proof to commit it upstream.

**Additional context**

[...]

PR #5625, opened 2022-03-10
unbound: prefer system nameservers if forwarding is enabled and only run unbound-anchor when necessary

swhite2:
resolves https://github.com/opnsense/core/issues/5577


src/opnsense/scripts/unbound/start.sh
```diff
@@ -26,11 +26,26 @@
...
-chroot -u unbound -g unbound / /usr/local/sbin/unbound-anchor -a /var/unbound/root.key
+# if the root.key file is missing or damaged, run unbound-anchor
+/usr/local/sbin/unbound-checkconf /var/unbound/unbound.conf 2> /dev/null 
+
+if [ $? -eq 1 ]; then
+	# unbound-anchor has undefined behaviour if file is corrupted, start clean 
+	rm -f /var/unbound/root.key
+
+	# if we are in forwarding mode, prefer to use the configured system nameservers
+	if [ $(/usr/local/sbin/pluginctl -g unbound.forwarding) ]; then
```
fichtner, line 39: that should read

```suggestion
	if [ -n "$(/usr/local/sbin/pluginctl -g unbound.forwarding)" ]; then
```

but honestly I don't think we can trust unbound.forwarding as it may be set but empty historically. don't we have a new forwarding knob as of forwarding MVC restructure?

src/opnsense/scripts/unbound/start.sh
```diff
@@ -26,11 +26,26 @@
...
 
-set -e
-
 # prepare and startup unbound, so we can easily background it
 
-chroot -u unbound -g unbound / /usr/local/sbin/unbound-anchor -a /var/unbound/root.key
+# if the root.key file is missing or damaged, run unbound-anchor
+/usr/local/sbin/unbound-checkconf /var/unbound/unbound.conf 2> /dev/null 
+
+if [ $? -eq 1 ]; then
```
fichtner, line 34:
```suggestion
if ! /usr/local/sbin/unbound-checkconf /var/unbound/unbound.conf 2> /dev/null; then
```

assuming "1" is returned is strange for all sorts of error behaviour that could be embedded here or later.. better to assume ! 0 and avoid going through return variable if we don't need it again

src/opnsense/scripts/unbound/start.sh
```diff
@@ -26,11 +26,26 @@
...
+	# unbound-anchor has undefined behaviour if file is corrupted, start clean 
+	rm -f /var/unbound/root.key
+
+	# if we are in forwarding mode, prefer to use the configured system nameservers
+	if [ $(/usr/local/sbin/pluginctl -g unbound.forwarding) ]; then
+		OPT_RESOLVE="-Rf /etc/resolv.conf"
```
fichtner, line 40: small note: we don't have to strictly point to /etc/resolv.conf and could populate our ow resolv.conf for bootstrapping (if it makes sense)

src/opnsense/scripts/unbound/start.sh
swhite2: > don't we have a new forwarding knob as of forwarding MVC restructure?
We don't. Query forwarding here refers to the legacy forwarding option as listed in General, using the system nameservers - forwarding as part of MVC restructure doesn't touch this.
 
> but honestly I don't think we can trust unbound.forwarding as it may be set but empty historically
any suggestions? I have no idea how this was registered in old versions.

src/opnsense/scripts/unbound/start.sh
swhite2: Fair enough, much better.

src/opnsense/scripts/unbound/start.sh
swhite2: This is actually a neat suggestion to integrate it with the MVC custom forwarding page. If we could do this earlier in the chain we can also use this to migrate the entire legacy query forwarding settings to MVC. The current hitch is the fact that system nameservers are awkward to fetch in MVC.

src/opnsense/scripts/unbound/start.sh
fichtner: for pluginctl this doesn't work if it was stored as `<forwarding/>` ... I have a similar issue with dns server registering at the moment and worked around it by moving logic from shell scripts  to php backend altogether.

we cound write a /var/unbound/resolv.conf.root file if we have that situation and just include it if it exists? I'm not 100% sure trusting /etc/resolv.conf is good because it may have other servers in there not used by unbound forwarding in particular.

swhite2: Are we certain this was the case historically? Current legacy doesn't do this - though it is better to assume it happens I suppose. 

As far as I'm aware the only servers populating /etc/resolv.conf are manually assigned nameservers in system->settigns->general, as well as dynamically added DHCP DNS servers. This falls within expectations for forwarding purposes (but only for query forwarding in unbound->general). Honestly, I'd really like to move this legacy section to MVC so every nameserver can be included here as well.

src/opnsense/scripts/unbound/start.sh
fichtner: you mean system: settings: general? if so we are not quite there yet

src/opnsense/scripts/unbound/start.sh
```diff
@@ -26,11 +26,24 @@
 # OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF
 # SUCH DAMAGE.
 
-set -e
-
 # prepare and startup unbound, so we can easily background it
 
-chroot -u unbound -g unbound / /usr/local/sbin/unbound-anchor -a /var/unbound/root.key
+# if the root.key file is missing or damaged, run unbound-anchor
+if ! /usr/local/sbin/unbound-checkconf /var/unbound/unbound.conf 2> /dev/null; then
```
fichtner, line 32: now that this is guarded by "if !" the set -e could remain where it originally was. if we really worry about unbound-anchor returning errors we may as well use "if !" with it to emit a log message in that case or use "|| true" and move the error return comment above the invoke where it belongs. it's easy to miss this otherwise later on

src/opnsense/scripts/unbound/start.sh
swhite2: Ok, I'll convert the PR to a draft state and figure out a solution to this and keep you updated.

src/opnsense/scripts/unbound/start.sh
swhite2: unbound-anchor will return non-zero on failover, which is still valid for bootstrapping.

src/opnsense/scripts/unbound/start.sh
swhite2: @fichtner changes regarding system: settings: general addressed in https://github.com/opnsense/core/pull/5606

src/opnsense/scripts/unbound/start.sh
```diff
@@ -30,7 +30,25 @@ set -e
...
+	# unbound-anchor has undefined behaviour if file is corrupted, start clean 
+	rm -f /var/unbound/root.key
+
+	# if we are in forwarding mode, prefer to use the configured system nameservers
+	if [ -n "$(/usr/local/sbin/pluginctl -g unbound.forwarding)" ]; then
+	  if [ ! -f /var/unbound/resolv.conf.root ]; then
+	    # fall back to regular resolv.conf file
+		  OPT_RESOLVE="-Rf /etc/resolv.conf"
```
fichtner, line 42: is there any benefit to fall back in that case?

src/opnsense/scripts/unbound/start.sh
```diff
@@ -30,7 +30,25 @@ set -e
...
 
-chroot -u unbound -g unbound / /usr/local/sbin/unbound-anchor -a /var/unbound/root.key
+# if the root.key file is missing or damaged, run unbound-anchor
+if ! /usr/local/sbin/unbound-checkconf /var/unbound/unbound.conf 2> /dev/null; then
+	# unbound-anchor has undefined behaviour if file is corrupted, start clean 
+	rm -f /var/unbound/root.key
+
+	# if we are in forwarding mode, prefer to use the configured system nameservers
+	if [ -n "$(/usr/local/sbin/pluginctl -g unbound.forwarding)" ]; then
+	  if [ ! -f /var/unbound/resolv.conf.root ]; then
```
fichtner, line 40: maybe we could just check if the file is non empty, in that case use it avoiding to use pluginctl -g here (of course only filling the file with contents on this setting.

src/etc/inc/plugins.inc.d/unbound.inc
```diff
@@ -282,9 +283,17 @@ EOD;
...
+                    /* Generate a custom resolv.conf file for use by unbound-anchor.
+                     * These servers all use port 53 so exclude localhost from being queried for bootstrapping
+                     * in our custom resolv.conf file as Unbound doesn't exist yet.
+                     */
+                    $resolv .= "nameserver $dnsserver\n";
                 }
                 $forward_conf .= "\tforward-addr: $dnsserver\n";
             }
+
+            file_put_contents("/var/unbound/resolv.conf.root", $resolv, LOCK_EX);
```
fichtner, line 296: are we always writing the file (empty in worst case)? to make sure it is reset if settings changed

swhite2: good point. it should be removed or emptied out if forwarding is disabled.

src/etc/inc/plugins.inc.d/unbound.inc
```diff
@@ -257,6 +257,7 @@ EOF;
 
     $forward_conf = '';
     $forward_local = '';
+    $resolv = '';
```
fichtner, line 260: last nit... I use grep like `grep resolv.conf.root` which can light up this variable too for easy code checks...

```suggestion
    $resolv_conf_root = '';
```

src/opnsense/scripts/unbound/start.sh
```diff
@@ -30,7 +30,20 @@ set -e
...
 
-chroot -u unbound -g unbound / /usr/local/sbin/unbound-anchor -a /var/unbound/root.key
+# if the root.key file is missing or damaged, run unbound-anchor
+if ! /usr/local/sbin/unbound-checkconf /var/unbound/unbound.conf 2> /dev/null; then
+	# unbound-anchor has undefined behaviour if file is corrupted, start clean 
+	rm -f /var/unbound/root.key
+
+	# if we are in forwarding mode, prefer to use the configured system nameservers
+	if [ -s /var/unbound/resolv.conf.root ]; then
+	  OPT_RESOLVE="-Rf /var/unbound/resolv.conf.root"
```
fichtner, line 40:
```suggestion
		OPT_RESOLVE="-Rf /var/unbound/resolv.conf.root"
```

fichtner:
Merged, thanks! ❤️

PR #5642, opened 2022-03-22
unbound: fix model description validation

swhite2:
We are being too strict with the Unbound host & domain descriptions, this fixes the resulting migration issues.


fichtner:
Merged, thanks!

PR #5648, opened 2022-03-25
get_nameservers(): also return manual dns entries to simplify code elsewhere

swhite2:


fichtner:
Minor issues here... let me deal with them in the merged code. Thanks!

PR #5658, opened 2022-03-30
Use random_bytes() from PHP 7.0+

oittaa:
PHP has had native [random_bytes()](https://www.php.net/manual/en/function.random-bytes.php) since version 7.0. Internally on FreeBSD it uses [getrandom()](https://github.com/php/php-src/blob/PHP-7.4.0/ext/standard/random.c#L104-L122) call which is probably a lot faster and more secure than `openssl_random_pseudo_bytes()`.

While making this pull request I also noticed that voucher generation is quite complicated. Should I make an additional pull request that simplifies the voucher generation with `random_int()` instead of those huge byte arrays? It could be optimized by using a string with valid characters and then just looping with `random_int()` like in the example below.

```php
// ...
if ($this->simplePasswords) {
    $characterMap = '23456789ABC...';
} else {
   // the other case here
  $characterMap = '...';
}
// ...
                $generatedUsername = '';
                for ($j = 0; $j < $this->usernameLength; $j++) {
                    $generatedUsername .= $characterMap[random_int(0, strlen($characterMap) - 1)];
                }
// ... and so on ...
```

### DIFF
--- a/src/etc/inc/auth.inc
+++ b/src/etc/inc/auth.inc
@@ -433,7 +433,7 @@ function local_user_set_password(&$user, $password = null)
 {
     if ($password == null) {
         /* generate a random password */
-        $bytes = openssl_random_pseudo_bytes(50);
+        $bytes = random_bytes(50);
         $password = pack('H*', bin2hex($bytes));
     }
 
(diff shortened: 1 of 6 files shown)


oittaa:
Fixed in #5659

PR #5660, opened 2022-03-30
Use password_verify() which is safer against timing attacks

oittaa:
> This function is safe against timing attacks.

https://www.php.net/manual/en/function.password-verify.php


AdSchellevis:
@oittaa that's practical, I had it on my list somewhere to look for a `crypt()` replacement, missed the `password_verify()` somehow. thanks!

PR #5661, opened 2022-03-31
Don't leak backup password to `ps`

oittaa:
https://www.openssl.org/docs/man1.1.1/man1/openssl.html

```
pass:password
    The actual password is password. Since the password is visible to utilities (like 'ps' under
    Unix) this form should only be used where security is not important.
```

Another issue was that secure temporary files were created, but never used or deleted.
```php
$file = tempnam(sys_get_temp_dir(), 'php-encrypt');
```
The files actually used were just created insecurely by writing to a known file after the creation of the previously mentioned `$file`.

```php
file_put_contents("{$file}.dec", $data);
```

### DIFF
--- a/src/opnsense/mvc/app/library/OPNsense/Backup/Base.php
+++ b/src/opnsense/mvc/app/library/OPNsense/Backup/Base.php
@@ -40,7 +40,7 @@ abstract class Base
      * @param string $params parameters to openssl command
      * @param string $pass passphrase to use
      * @param string $input stdin to openssl process
-     * @return string base64 encoded output from openssl
+     * @return string output from openssl
      */
     private function opensslEnc($params, $pass, $input)
     {
--- a/src/opnsense/mvc/app/library/OPNsense/Backup/Base.php
+++ b/src/opnsense/mvc/app/library/OPNsense/Backup/Base.php
@@ -36,6 +36,44 @@ namespace OPNsense\Backup;
 abstract class Base
 {
     /**
+     * openssl enc command
+     * @param string $params parameters to openssl command
+     * @param string $pass passphrase to use
+     * @param string $input stdin to openssl process
+     * @return string base64 encoded output from openssl
+     */
+    private function opensslEnc($params, $pass, $input)
+    {
+        $tmpFile = tempnam(sys_get_temp_dir(), 'php-encrypt');
+        if (!$tmpFile) {
+            return null;
+        }
+        $result = null;
+        file_put_contents($tmpFile, $pass);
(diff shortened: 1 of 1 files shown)


oittaa:
I really like that the current encrypted config.xml can be decrypted with just OpenSSL without OPNsense, but there's one relatively big caveat. It's not protected against accidental or malicious corruption. AES in GCM mode would prevent that. I made [a tiny wrapper](https://github.com/oittaa/aes-gcm-php/blob/main/src/AESGCM.php) as an example what a new "aes-256-gcm" mode could look like. It wouldn't even need any `openssl` invocation outside of PHP since it uses the native `openssl_encrypt`.

So the question is, do you think that the integrity checking would be worth the trouble of updating the scheme?

AdSchellevis:
@oittaa I still need to test this locally, but the code looks good.

> I really like that the current encrypted config.xml can be decrypted with just OpenSSL without OPNsense, but there's one relatively big caveat. It's not protected against accidental or malicious corruption. AES in GCM mode would prevent that. I made [a tiny wrapper](https://github.com/oittaa/aes-gcm-php/blob/main/src/AESGCM.php) as an example what a new "aes-256-gcm" mode could look like. It wouldn't even need any openssl invocation outside of PHP since it uses the native openssl_encrypt.
> 
> So the question is, do you think that the integrity checking would be worth the trouble of updating the scheme?

The problem might be backwards compatibility, which has the risk of adding more logic and the need to keep both approaches for a very long time. We can always discuss in PR, these are just my main concerns.

AdSchellevis:
@oittaa I seems like I was a bit too optimistic  here and had to revert the commit as it seems to hang the backup on my end. Maybe we best keep the `exec()` and only go for your idea to flush the password in between execution? It's easier to test and less likely to regress on other peoples installs.

oittaa:
Have you tested https://github.com/opnsense/core/pull/5665?

That one doesn't need any `openssl` shell invocation or temporary files at all.

oittaa:
https://soatok.blog/2020/07/12/comparison-of-symmetric-encryption-methods/#aes-gcm-vs-aes-cbc

TLDR:
> AES-GCM vs. AES-CBC
> Just use AES-GCM. No contest.
> If you must use AES-CBC, then you must also MAC your ciphertext (and the initialization vector–IV for short).

oittaa:
@AdSchellevis I don't have older MD5 based configuration file backups, but at least the "aes-256-cbc sha512" ones work perfectly fine with that newer pull request.

AdSchellevis:
> Have you tested https://github.com/opnsense/core/pull/5665?

I haven't, but at the moment I don't have enough time available to review it in full either. Let's keep the PR open for a while and see if other people chime in for some additional testing and feedback. I'll try to fix the password leak in the current code later today.

PR #5662, opened 2022-04-01
Dynamically generated bcrypt cost

oittaa:
If I understood correctly, bcrypt cost was set to `10` in 2016. While bcrypt is still an excellent algorithm, the computers get faster every year. This pull request increases bcrypt cost dynamically until it takes at least 0.2 seconds. This shouldn't make the authentication process horrible on slower machines, but especially faster machines will benefit from better security. If 0.2 seconds is too long, maybe it could be lowered to 0.1s...

For example on my cheap and couple of years old laptop this would mean at least `$cost = 12`.

### DIFF
--- a/src/etc/inc/auth.inc
+++ b/src/etc/inc/auth.inc
@@ -439,9 +439,8 @@ function local_user_set_password(&$user, $password = null)
     $cost = 10;
     do {
         $timer = hrtime(true);
-        $hash = password_hash($password, PASSWORD_BCRYPT, [ 'cost' => $cost ]);
+        $hash = password_hash($password, PASSWORD_BCRYPT, [ 'cost' => $cost++ ]);
         $timer = (hrtime(true) - $timer) / 10 ** 9;
-        $cost++;
     } while ($timer < 0.2 && $cost < 32);
     if ($hash !== false) {
         $user['password'] = $hash;
--- a/src/etc/inc/auth.inc
+++ b/src/etc/inc/auth.inc
@@ -438,6 +438,7 @@ function local_user_set_password(&$user, $password = null)
 
     $cost = 10;
     do {
+        /* increase bcrypt cost until it takes at least 0.2 seconds */
         $timer = hrtime(true);
         $hash = password_hash($password, PASSWORD_BCRYPT, [ 'cost' => $cost++ ]);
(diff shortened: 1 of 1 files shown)


fichtner:
It’s a reasonable approach but the code itself ends up overthinking the problem. How about we just increase the cost moderately to compensate for faster CPUs?

oittaa:
To my understanding ` pack('H*', bin2hex($bytes))` is a non-operation so I removed it.

You can test it like this

```php
$bytes = random_bytes(50);
$password = pack('H*', bin2hex($bytes));
var_dump($bytes === $password);
```

oittaa:
> It’s a reasonable approach but the code itself ends up overthinking the problem. How about we just increase the cost moderately to compensate for faster CPUs?

I wouldn't be completely against it, but this approach is quite future proof unless bcrypt gets broken. :)

fichtner:
Well, it has a built-in bit rot cost of 32 upper limit and it iterates from cheapest to desired value. It tries to be too clever.

pack() was likely used to introduce a readable password without hidden binary characters for better or worse. It shouldn’t do any harm and increases resulting password length.

oittaa:
> pack() was likely used to introduce a readable password without hidden binary characters for better or worse. It shouldn’t do any harm and increases resulting password length.

I don't understand this. Can you show an example?

fichtner:
Alright, fine, originated from here 66d982fd16 but doesn't do anything as you said. Looking further it seems to be introduced in a160f3a2fa1e a few years earlier. Not sure what the intent was then by @AdSchellevis as I only preserved this code.

AdSchellevis:
@fichtner I'm pretty sure I choose it for readability at the time, looking at it now it looks like @oittaa is right it doesn't change anything to the output.

fichtner:
@oittaa pick a new standard cost or scrap this PR please. We can verify a reasonable fixed cost on older hardware easily but cannot live with dynamic cost that can be lowered by inducing system load.

oittaa:
> @oittaa pick a new standard cost or scrap this PR please. 

Ok, I will scrap this.

> We can verify a reasonable fixed cost on older hardware easily but cannot live with dynamic cost that can be lowered by inducing system load.

I don't know if you noticed, but this never returned anything lower than the current value of 10.

PR #5663, opened 2022-04-01
unbound: clean up references to legacy code, use mvc instead. also minor style fix

swhite2:
Migrating to MVC means we can use the existing get/set API calls without having to manually manipulate the legacy configuration.

This PR also adds a migration for the legacy unbound->forwarding configuration.

I'm not sure what the original intent behind https://github.com/opnsense/core/blob/master/src/wizard/system.xml#L158-L159 was, but i'm removing it here.

### DIFF
--- a/src/etc/inc/plugins.inc.d/unbound.inc
+++ b/src/etc/inc/plugins.inc.d/unbound.inc
@@ -255,10 +255,13 @@ EOF;
         $include_dhcpleases = '';
     }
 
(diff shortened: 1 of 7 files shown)


fichtner:
Just quickly up front: the wizard tries to set non-forward mode on reconfigure to ensure consistency with a default configuration. If we need it or simply discard I’m not sure, but it’s more work than necessary most likely to emulate it.

fichtner:
merged, thanks!

Issue #5667, opened 2022-04-03
GitHub Actions and unit testing


oittaa:
**Important notices**

Before you add a new report, we ask you kindly to acknowledge the following:

- [x] I have read the contributing guide lines at https://github.com/opnsense/core/blob/master/CONTRIBUTING.md
- [x] I am convinced that my issue is new after having checked both open and closed issues at https://github.com/opnsense/core/issues?q=is%3Aissue

Having unit tests and linter run on every pull request would be a helpful addition. Maybe it would even encourage people to write more unit tests under https://github.com/opnsense/core/tree/master/src/opnsense/mvc/tests

**Describe the solution you like**

Add some Actions under `.github/workflows`

**Describe alternatives you considered**

A clear and concise description of any alternative solutions or features you considered.

**Additional context**

Add any other context or screenshots about the feature request here or links to relevant forum thread or similar

AdSchellevis:
it would be practical to offer some integration, not sure what's needed for this to happen to be honest.

fichtner:
Everything is run via nightly.opnsense.org

oittaa:
I've never used GitHub Actions in FreeBSD before, but there seems to be a way to run virtual machines with it. No idea how long the full build would take. https://github.com/vmactions/freebsd-vm

oittaa:
Seems to be working. https://github.com/oittaa/core/runs/5805822964?check_suite_focus=true

I think it would make sense to first run linter and phpunit on Ubuntu to get faster results, if there's a simple typo or other errors.

oittaa:
[Here](https://github.com/oittaa/core/runs/5809463371?check_suite_focus=true)'s an example of `phpunit` run on Ubuntu. I had to copy couple of files to /usr/local, but otherwise it seems to be working out of the box. The same tests are failing as on my real FeeBSD OPNsense installation.

AdSchellevis:
> The same tests are failing as on my real FeeBSD OPNsense installation.

I noticed the same thing recently as well, we have to figure out why that is at some point in time.

PR #5673, opened 2022-04-04
PHPunit testing with GitHub Actions

oittaa:
Initial support for continuous integration. The same tests fail on Ubuntu as on my FreeBSD-based installation.

More at https://github.com/opnsense/core/issues/5667

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,0 +1,33 @@
+name: CI
+
+on:
+  # Triggers the workflow on push or pull request events but only for the master branch
+  push:
+    branches: [ master ]
+  pull_request:
+    branches: [ master ]
+
+jobs:
+  unit-tests:
+    runs-on: ubuntu-20.04
+    strategy:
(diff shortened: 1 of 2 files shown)


williamdes:
@AdSchellevis I'd like to add a static php code analyser like phpstan, is there something blocking this PR so I can use it as a base afterwards ?

AdSchellevis:
@williamdes at the moment we only use coverty (https://scan.coverity.com/projects/opnsense-core) for periodic static analysis, but don't have plans to support GitHub actions also due to the time it would take from our team to make it a valuable addition (we do run unit tests from our build infrastructure as described here https://github.com/opnsense/tools#regression-tests-and-ports-audit). 

Our base principle is that we only add/accept code that we are actually using as these things tend to lead to unmaintained bitrot quite easily.

AdSchellevis:
at the moment no plans in using GitHub actions from our end, the voucher part might be mergeable, but needs a different PR.

PR #5676, opened 2022-04-04
Fix UUID generation

oittaa:
mt_rand() is in not suitable for UUID generation.

### DIFF
--- a/src/opnsense/mvc/app/models/OPNsense/Base/FieldTypes/BaseField.php
+++ b/src/opnsense/mvc/app/models/OPNsense/Base/FieldTypes/BaseField.php
@@ -141,17 +141,10 @@ abstract class BaseField
      */
     public function generateUUID()
     {
-        return sprintf(
-            '%04x%04x-%04x-%04x-%04x-%04x%04x%04x',
-            mt_rand(0, 0xffff),
-            mt_rand(0, 0xffff),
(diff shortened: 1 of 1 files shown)


AdSchellevis:
I'm not sure the rfc agrees to be honest https://datatracker.ietf.org/doc/html/rfc4122.html#section-4.4 , unless there are other good reasons for replacing this, I think we should be careful with replacing random pieces of code to avoid unexpected regressions.

oittaa:
The wording isn't super clear, but just take a look at any real UUIDv4 implementation. They all use cryptographically secure random number generators.

oittaa:
For example https://github.com/python/cpython/blob/3.10/Lib/uuid.py#L715

AdSchellevis:
I know, most (not all) use the system source from `/dev/urandom`, but again predictability isn't an issue for these sequences, uniqueness is. The RFC clearly states "pseudo" in which case Mersenne Twister should qualify as well.

oittaa:
It probably won't cause huge problems here even though Mersenne Twister has [issues](https://en.wikipedia.org/wiki/Mersenne_Twister#Disadvantages), but these  kind of things usually end up being footguns. In the future someone might use this UUID for something security critical and then it's trivially easy to break down. Even if you don't think this implementation is good enough for you, you should have a plan to migrate to something that has better properties.

PR #5685, opened 2022-04-08
unbound: overrides: mvc: case sort order

NOYB:
Unbound host and domain sort order.

Host and domain names are case insensitive.  Should be handled case insensitively.  myHost comes before YourHost.  Always.

Sort order should be:
myHost
YourHost

Not:
YourHost
myHost


swhite2:
@NOYB this change would also mean that `AnotherHost` would come before `myHost`, this is by definition case insensitive and if you feel this is correct I will accept this change. Otherwise we would have to remove the `SORT_FLAG_CASE` and somehow force sorting in descending order by default.

swhite2:
@NOYB this change would also mean that `AnotherHost` would come before `myHost`, this is by definition case insensitive and if you feel this is correct I will accept this change. Otherwise we would have to remove the `SORT_FLAG_CASE` and somehow force sorting in descending order by default.

NOYB:
Yes that is correct.  "A" and "a" are equivalent in case insensitivity.  So both come before "M" and "m".  Or any other letter irregardless of upper/lower case.
Thanks

Issue #5493, opened 2022-01-21
DHCP client doesn't set fargw


jplitza:
When a DHCP server replies with netmask /32, OPNsense doesn't correctly set the gateway (which is outside the "subnet"). Configuring this manually works fine.

Manually inserting `$thisconf['fargw'] = 1;` into this block makes it work via DHCP too:
https://github.com/opnsense/core/blob/68775a50e8980cfa4d88f476260e321e7008ee45/src/opnsense/mvc/app/library/OPNsense/Routing/Gateways.php#L242-L248

Obviously setting fargw should only happen when needed - which is why this isn't a PR... yet.

fichtner:
Hi @jplitza,

That’s an interesting observation. I suppose detecting this is useful regardless of how it could be carried out (and when).

One thing to note about the far gateway setting is that it will stick to the configuration once saved (automatic gateway becoming a manual gateway). In that case if DHCP should lease a normal IP with same subnet gateway later on the setting would stick to it and maybe break connectivity (although I’m not sure).

Maybe dynamic gateways must overwrite this setting according to the detected situation?


Cheers,
Franco

jplitza:
@fichtner TBH, I'm not even a user of OPNsense, so I'm not at all familiar with the concepts you mention ("automatic gateway becoming a manual gateway", "dynamic gateway") and the implications of the proposed change on them. I'm just looking into rolling out such a DHCP setup in our ISP network, and one of our customers is using OPNsense, so I dug into the code.

fichtner:
Ok, no problem. Let me assign this and try to work on an out of the box solution.

Bit at least setting far gateway manually for your customers is an easy workaround to get it running.


Cheers,
Franco

fichtner:
Proposal is in there to be released in 22.7 at the latest.

PR #5703, opened 2022-04-12
system: detect far gateway situation for #5493

fichtner:
The far gateway flag has some benefits for configuration runs
and validation purposes on the GUI but in the end after lots
of reworks we are able to reliably get a network from the interface
to put the default route on so that we can detect if we are in
need of a far gateway or not.  This is required for automatic
gateways on DHCP that hand out these situations while the
gateway code should not be in charge of flipping on the fargw
bit as it does pertain to runtime interface configuration.

We don't modify the configuration: if fargw is set it is used,
if not we force fargw if it is not set but required (automatic
case describe in ticket).

(cherry picked from commit 9cf2b221d8be339b609258b970e89542a2b37f7c)
(cherry picked from commit 9486473b2ff449318c9b1bf451aa9387f7a097c9)
(cherry picked from commit 00a86f74dbbefe3abbb926d98251067df53dfe8e)
(cherry picked from commit 7fa1f8b8bee0e307ed7e2c24e649147167c363a4)
(cherry picked from commit f6551c982d75feed016f1bb33458629915f2fe48)
(cherry picked from commit fc51b1541a8af8ef4ee9059b783bdb1aeb7027b5)

### DIFF
--- a/src/etc/inc/system.inc
+++ b/src/etc/inc/system.inc
@@ -526,7 +526,12 @@ function system_default_route($gateway, $family, $interface, $far = false)
         @file_put_contents("/tmp/{$realif}_defaultgw", $gateway);
 
         if (!$far) {
-            $realif = null;
+            list (, $network) = interfaces_primary_address($interface);
+            if (ip_in_subnet($gateway, $network)) {
+                $realif = null;
+            } else {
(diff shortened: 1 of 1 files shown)


fichtner:
We will let this blend into 22.7 naturally and see what happens.

PR #5707, opened 2022-04-12
Unbound overrides: fix handling of wildcard aliases

devin122:
Wildcard functionality of HostnameField was refactored
See f110c988d4c5722308d0ec7a4ef5b86c0b3f767c

### DIFF
--- a/src/opnsense/mvc/app/models/OPNsense/Unbound/Unbound.xml
+++ b/src/opnsense/mvc/app/models/OPNsense/Unbound/Unbound.xml
@@ -199,7 +199,7 @@
                             <check></check>
                         </check001>
                     </Constraints>
-                    <WildcardAllowed>Y</WildcardAllowed>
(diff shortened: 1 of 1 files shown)


AdSchellevis:
this change is part of the master branch, so we should be good there.

devin122:
I don't follow. This PR fixes a bug introduced via f110c988d4. In that commit the wildcard functionality was broken up, but the model for aliases was not updated (ie the hostname of aliases still refers to the no-longer existent `WildcardAllowed`)

AdSchellevis:
Sorry, my mistake, it seems I replaced one but forgot the other.
https://github.com/opnsense/core/blob/0e3a02bfd2d3af68fc329fc6a33c9e717270eb78/src/opnsense/mvc/app/models/OPNsense/Unbound/Unbound.xml#L101-L104

AdSchellevis:
@devin122 thanks!

PR #5711, opened 2022-04-15
Phalcon5 migration: provide compatibility layer between v4 and v5

swhite2:
This PR does the following:

- ~~Provide php code to generate classes that provide most of the heavy lifintg for operating between the namespace changes in v4 -> v5.~~ 
- ~~The generated classes are included for check-in here as well to fuel the discussion for whether we should dynamically generate these or just provide the from the source repository. Main argument is to keep the diff reduced, but generated code might be fragile to corruption.~~
- Check-in files that handle all the namespaces affected by the changes in Phalcon5.
- Phalcon's AbstractValidator was a problem and has been worked around using @AdSchellevis 's implementation of a Validation and BaseValidator class. These act as a wrapper around the "problem zone". 
- The bulk of the changes include namespace changes from Phalcon\* to OPNsense\Phalcon\*. The OPNsense\Phalcon namespace takes care of the namespace issues created by Phalcon v5.
- ~~The new Validator() class does not return a Messages() object, but an array. This means Messages()->Count() occurrences do not work anymore. i've changed these to use a normal array count.~~
- ~~Following the same logic, the expectations of certain unit tests have changed as well.~~

All of this should be thoroughly tested before merging it into master.


AdSchellevis:
@swhite2 changed the return type of `performValidation()` to stay backwards compatible for now. Changing the return type would also influence some of the plugins. See  https://github.com/opnsense/core/commit/6814f32755f239a362a6153abdf287bac8f9cc33

swhite2:
@AdSchellevis Thanks, adjusted the PR to account for this change. Unit tests cannot expect `null` now, since a Messages() object is always created in the constructor. So these check for a zero or non-zero count now.

Makefile
```diff
@@ -132,7 +132,6 @@ CORE_DEPENDS_amd64?=	beep \
 # transition helpers for PHP 8/Phalcon 5 migration
 CORE_DEPENDS_PHP74=	php74-json php74-openssl php74-phalcon${CORE_PHALCON}
```
fichtner, line 133:
```suggestion
CORE_DEPENDS_PHP74=	php74-json php74-openssl php74-phalcon
```

to align with below change

PR #5716, opened 2022-04-18
Alias Tables Natural Sort Order

NOYB:
Sort IP addresses in natural order.

FIREWALL: DIAGNOSTICS: ALIASES

i.e.
sort as
 2.3.4.5
12.3.4.5

rather than
12.3.4.5
 2.3.4.5


AdSchellevis:
@NOYB can you try https://github.com/opnsense/core/commit/fcef2d65d216f2f43aad98152d6baa5588a5dbb6 ?  should do the trick as well and cleans up some duplicated code.

NOYB:
Yes this works.  Well except for IPv6 which sort/rsort doesn't work either.
Seems alpha is placed ahead of numbers (2a00: comes before 2000:).
Don't know if PHP has any built in function for IPv6 natural sort though.

NOYB:
Perhaps not sorting at all and expect/rely on the source order to be best.

PR #5736, opened 2022-04-23
Add ability to set a custom dnsbl target address

lfuelling:
Hi,

this PR adds the ability to set a custom target address for the unbound DNS blocklist.
The address can be set using a text field in the advanced blocklist settings of the unbound plugin and will be used in the `blocklists.py` script instead of `0.0.0.0` when matching an ip regex. (I'd prefer validating the ip in the textfield but I was unable to find examples for this in the source code)

My motivation for this is that I want to host a service that shows a blank page for any request to prevent blocked stuff from generating error pages and get some statistics about what stuff actually got blocked.

### DIFF
--- a/src/opnsense/mvc/app/controllers/OPNsense/Unbound/forms/dnsbl.xml
+++ b/src/opnsense/mvc/app/controllers/OPNsense/Unbound/forms/dnsbl.xml
@@ -21,6 +21,12 @@
         <help>List of domains from where blocklist will be downloaded.</help>
     </field>
     <field>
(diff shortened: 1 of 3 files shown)


AdSchellevis:
@lfuelling I think you're PR was incomplete (missed model modifications), can you try https://github.com/opnsense/core/commit/43958b00cc698928b3127b3477868890e626642f ?

lfuelling:
@AdSchellevis thanks for the hint, I have added another commit to the branch.
Could you please reopen the PR as I don't have access to do that?

PR #5747, opened 2022-04-28
unbound: make blocklist additions/removals dynamic to prevent a restart

swhite2:
This PR contains an optimization for the way unbound can be updated in its' configuration
without the need for restarting Unbound and by extension a loss of internet connectivity.

Hopefully this will pave the way for optimizations on other aspects of Unbounds' configuration as well.
We could for example look into making the provided wrapper function more generic.

The GUI has also been modified slightly to inform the user of the amount of RRs added/removed.

### DIFF
--- a/src/opnsense/mvc/app/controllers/OPNsense/Unbound/Api/ServiceController.php
+++ b/src/opnsense/mvc/app/controllers/OPNsense/Unbound/Api/ServiceController.php
@@ -44,7 +44,11 @@ class ServiceController extends ApiMutableServiceControllerBase
         $this->sessionClose();
         $backend = new Backend();
         $backend->configdRun('template reload ' . escapeshellarg(static::$internalServiceTemplate));
-        $response = $backend->configdRun(static::$internalServiceName . ' dnsbl');
-        return array('status' => $response);
+        $response = json_decode(trim($backend->configdRun(static::$internalServiceName . ' dnsbl')), true);
+        if ($response !== null) {
+            $response['status'] = "OK";
+            return $response;
+        }
+        return array('message' => 'unable to run configd action');
     }
 }
--- a/src/opnsense/mvc/app/controllers/OPNsense/Unbound/Api/ServiceController.php
+++ b/src/opnsense/mvc/app/controllers/OPNsense/Unbound/Api/ServiceController.php
@@ -47,6 +47,11 @@ class ServiceController extends ApiMutableServiceControllerBase
         $response = json_decode(trim($backend->configdRun(static::$internalServiceName . ' dnsbl')), true);
         if ($response !== null) {
             $response['status'] = "OK";
+            $response['status_msg'] = sprintf(
+              gettext("Added %d and remove %d  resource records."),
+              $response['additions'],
+              $response['removals']
+            );
             return $response;
         }
         return array('message' => 'unable to run configd action');
--- a/src/opnsense/mvc/app/controllers/OPNsense/Unbound/Api/ServiceController.php
+++ b/src/opnsense/mvc/app/controllers/OPNsense/Unbound/Api/ServiceController.php
@@ -48,7 +48,7 @@ class ServiceController extends ApiMutableServiceControllerBase
         if ($response !== null) {
             $response['status'] = "OK";
             $response['status_msg'] = sprintf(
-              gettext("Added %d and remove %d  resource records."),
+              gettext("Added %d and removed %d  resource records."),
               $response['additions'],
               $response['removals']
             );
--- a/src/opnsense/scripts/unbound/blocklists.py
+++ b/src/opnsense/scripts/unbound/blocklists.py
@@ -147,7 +147,6 @@ if __name__ == '__main__':
     # write out results
     with open("/usr/local/etc/unbound.opnsense.d/dnsbl.conf", 'w') as unbound_outf:
         if blocklist_items:
-            unbound_outf.write('server:\n')
             for entry in blocklist_items:
                 unbound_outf.write("local-data: \"%s A %s\"\n" % (entry, destination_address))
 
--- a/src/opnsense/scripts/unbound/wrapper.py
+++ b/src/opnsense/scripts/unbound/wrapper.py
@@ -1,7 +1,7 @@
 #!/usr/local/bin/python3
 
 """
-    Copyright (c) 2017 Ad Schellevis <ad@opnsense.org>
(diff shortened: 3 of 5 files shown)


src/opnsense/mvc/app/controllers/OPNsense/Unbound/Api/ServiceController.php
```diff
@@ -44,7 +44,11 @@ public function dnsblAction()
         $this->sessionClose();
         $backend = new Backend();
         $backend->configdRun('template reload ' . escapeshellarg(static::$internalServiceTemplate));
-        $response = $backend->configdRun(static::$internalServiceName . ' dnsbl');
-        return array('status' => $response);
+        $response = json_decode(trim($backend->configdRun(static::$internalServiceName . ' dnsbl')), true);
+        if ($response !== null) {
+            $response['status'] = "OK";
```
fichtner, line 49: what sort of output are we looking for here? seems like a lot to emulate "OK". Better to add a separate configd action for it?

swhite2: specifically, the amount of RRs added/deleted from the running Unbound instance. It would be nice to include the file stats generated in the blocklists as well, but this is all from a UX perspective and up for debate.

I agree the emulation is a bit out of place, either we scrap showing the user anything and make it a simple script or include the status key in the wrapper output since the SimpleActionButton always requires one.

swhite2:
> > Hopefully this will pave the way for optimizations on other aspects of Unbounds' configuration as well.
> > We could for example look into making the provided wrapper function more generic.
> 
> Yes and no. What would be nice would be ACL updates but that isn't supported by unbound-control. This likely applies to other situations as well where it would be nice but wasn't implemented. ;)

True, but unbound-control contains a wealth of other functionality which theoretically allows us to drastically reduce the number of restarts Unbound requires. This matters especially when millions of DNSBL records need to be loaded in every time. On top of that, we could trim the legacy codebase of functionality (such as overrides) we deemed too complex to put in a template and make it the responsibility of a configd action. 

A downside I see would be the fact that 'Apply' wouldn't really apply the entire configuration anymore, but may be addressed through either documentation or a general 'force restart' option. This is also the reason I renamed the dnsbl button to `Download & Apply`.

AdSchellevis:
@swhite2 feedback and proposed changes in https://github.com/opnsense/core/tree/unbound_PR5747 (https://github.com/opnsense/core/commit/052fc62f91925a6233fe05e507273a9e5ce8e10d)

swhite2:
@AdSchellevis Thanks for the review. Cherry-picked it and changed the following:

1. Small spelling fix
2. Strip more aggressively
3. Do removals before additions, to prevent case insensitive duplicates from being added first, only to be removed next.

src/opnsense/mvc/app/controllers/OPNsense/Unbound/Api/ServiceController.php
fichtner: > scrap showing the user anything and make it a simple script

well you can leave the output but I would change "scritp_output" to "script" in the template to get rid of this workaround

src/opnsense/scripts/unbound/wrapper.py
```diff
@@ -1,7 +1,7 @@
 #!/usr/local/bin/python3
 
 """
-    Copyright (c) 2017 Ad Schellevis <ad@opnsense.org>
+    Copyright (c) 2017-2022 Ad Schellevis <ad@opnsense.org>
```
fichtner, line 4:
```suggestion
    Copyright (c) 2017 Ad Schellevis <ad@opnsense.org>
```

identity crisis :)

src/opnsense/scripts/unbound/blocklists.py
```diff
@@ -147,7 +147,6 @@ def uri_reader(uri):
     # write out results
     with open("/usr/local/etc/unbound.opnsense.d/dnsbl.conf", 'w') as unbound_outf:
         if blocklist_items:
-            unbound_outf.write('server:\n')
```
fichtner, line 150: what's the story here?

src/opnsense/scripts/unbound/wrapper.py
fichtner: nevermind :)

src/opnsense/scripts/unbound/blocklists.py
swhite2: server: is a no-op. It does nothing, the original thought was that a `server:` statement was necessary to parse the local-data entries. This makes parsing easier.

src/opnsense/mvc/app/controllers/OPNsense/Unbound/Api/ServiceController.php
swhite2: how would I fetch the amount of records then? It's returned by the script itself.

src/opnsense/scripts/unbound/blocklists.py
fichtner: thanks for clarification :)

src/opnsense/mvc/app/controllers/OPNsense/Unbound/Api/ServiceController.php
fichtner: ok fair enough, but shouldn't we at least try to make the return consistent? status + status_message vs. message at the moment (see below)

swhite2: We can move "status":"OK" to the script output, but "status_msg" needs a gettext() so it needs a modification anyway.

AdSchellevis:
@swhite2 thanks!

src/opnsense/service/conf/actions.d/actions_unbound.conf
```diff
@@ -68,9 +68,9 @@ message:Checking Unbound configuration
 [dnsbl]
 command:
     /usr/local/opnsense/scripts/unbound/blocklists.py &&
-    /usr/local/sbin/pluginctl -s unbound restart
+    /usr/local/opnsense/scripts/unbound/wrapper.py -b
 parameters:
-type:script
+type:script_output
 message:Updating Unbound DNSBLs
 description:Update Unbound DNSBLs
```
fichtner, line 75: @swhite2 @AdSchellevis due to review for 22.1.7 I saw that this is also used for a cron job due to the attached description. will it even continue to work? if not we need to split into dnsbl and dnsbl.cron ...

PR #5769, opened 2022-05-13
Translate for privileges.

kekek2:


AdSchellevis:
I rather keep it as is for now as we're not feeding these into the translation pipeline anyway

PR #5770, opened 2022-05-13
improve diag_dns

zerwes:
During some posts in the opnsense forum, I realized that many users are mostly overwhelmed if you ask for some debugging information for dns problems from the console.
So I decidet to extend the diag_dns script with the option to add the option for selecting a source IP for the queries (required for example for queries passing ipsec VPNs etc...).
While looking at the script I realized some disagreements, so the PR is about a little bit more ...

 * allow setting a source address for the query
 * allow specifying the DNS server to be queried
 * collect and display more results like rcode and answer per server (not just the query time)
 * refactored the DNS diag process

Well, my PHP foo is in fact from the time of PHP 3/4, but I hope the code fulfills your requirements.

Some examples of the resulting UI:
 * https://ibb.co/k4WvzF1
 * https://ibb.co/QPpXwss

### DIFF
--- a/src/www/diag_dns.php
+++ b/src/www/diag_dns.php
@@ -28,38 +28,90 @@
  */
 
 require_once("guiconfig.inc");
+require_once("system.inc");
+require_once("interfaces.inc");
 
 $resolved = array();
-$dns_speeds = array();
-if (!empty($_REQUEST['host'])) {
+
+if ($_SERVER['REQUEST_METHOD'] === 'GET') {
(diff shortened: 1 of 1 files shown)


AdSchellevis:
I've just replaced the legacy diagnostics tool with an mvc implementation in https://github.com/opnsense/core/commit/f159f68f9728348d35841721251325d613f4c053 as we try to get rid of the old code. The new version should be easier to extend.

PR #5771, opened 2022-05-13
Bypass cache with timestamp in upgradestatus call

gibwar:
Even though the server responds with proper cache headers to avoid
browser caching, it seems that Safari (and maybe others) may still cache
repeated calls to `/api/core/firmware/upgradestatus`. This change adds
in the time-old tradition of appending the current timestamp in
milliseconds to each request which should force the browser to fetch the
resource.

Ref: https://www.reddit.com/r/opnsense/comments/umhx04/opnsense_2217_released/i83bp36/?context=5

### DIFF
--- a/src/opnsense/mvc/app/views/OPNsense/Core/firmware.volt
+++ b/src/opnsense/mvc/app/views/OPNsense/Core/firmware.volt
@@ -260,7 +260,9 @@
     }
 
     function trackStatus() {
-        ajaxGet('/api/core/firmware/upgradestatus', {}, function(data, status) {
+        // append Date.now() (epoch time in millisecods) to the end of the URL to force browsers to
+        // bypass aggressive caching. The use of the "_" param is to mimic jQuery's cache: false.
+        ajaxGet('/api/core/firmware/upgradestatus', { _: Date.now() }, function(data, status) {
             if (status != 'success') {
                 // recover from temporary errors
                 setTimeout(trackStatus, 1000);
--- a/src/opnsense/mvc/app/views/OPNsense/Core/firmware.volt
+++ b/src/opnsense/mvc/app/views/OPNsense/Core/firmware.volt
@@ -330,7 +330,6 @@
             $("#statustab_progress").removeClass("fa fa-spinner fa-pulse");
 
             if (reset === true) {
-                // bust aggressive caching with adding a timestamp to the request.
                 ajaxGet('/api/core/firmware/upgradestatus', { _: Date.now() }, function(data, status) {
                     if (data['log'] != undefined && data['log'] != '') {
                         $('#update_status').html(data['log']);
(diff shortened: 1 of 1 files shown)


gibwar:
Manually tested by opening up a console window on the OPNsense management page and pasted and ran the following:

```js
ajaxGet('/api/core/firmware/upgradestatus', { _: Date.now() }, console.log)
```
```text
Output:
<< Object { readyState: 1, getResponseHeader: getResponseHeader(e), getAllResponseHeaders: getAllResponseHeaders(), setRequestHeader: setRequestHeader(e, t), overrideMimeType: overrideMimeType(e), statusCode: statusCode(e), abort: abort(e), state: state(), always: always(), catch: catch(e), … }
---
Object { status: "done", log: "***GOT REQUEST TO CHECK FOR UPDATES***\n{...snip...}\n***DONE***" }
 success
```

fichtner:
Thanks for adding a PR WRT our discussion 😊 I think there’s a second upgradestatus call.

still not sure if the time stamp is seconds or milliseconds. We need millisecond resolution since we call multiple times per second (200 milliseconds?)

Cheers,
Franco

gibwar:
Yes, [`Date.now()`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date/now) returns the Unix epoch in milliseconds, so it can run multiple times per second without issue. I've confirmed this all the way down to IE11. 😄 If we're calling more than once per millisecond it'll break. 🤭

Copied from MDN for completeness:
> [Return value](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date/now#return_value)
>
> A [Number](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Number) representing the milliseconds elapsed since the UNIX epoch.

gibwar:
> I think there’s a second upgradestatus call.

Indeed there is a second call! I searched the code base for all references to `upgradestatus` and found the second one just below and no other references!

src/opnsense/mvc/app/views/OPNsense/Core/firmware.volt
```diff
@@ -328,7 +330,8 @@
             $("#statustab_progress").removeClass("fa fa-spinner fa-pulse");
 
             if (reset === true) {
-                ajaxGet('/api/core/firmware/upgradestatus', {}, function(data, status) {
+                // bust aggressive caching with adding a timestamp to the request.
```
fichtner, line 333:
```suggestion
```

let's ditch these comments

src/opnsense/mvc/app/views/OPNsense/Core/firmware.volt
```diff
@@ -260,7 +260,9 @@
     }
 
     function trackStatus() {
-        ajaxGet('/api/core/firmware/upgradestatus', {}, function(data, status) {
+        // append Date.now() (epoch time in millisecods) to the end of the URL to force browsers to
+        // bypass aggressive caching. The use of the "_" param is to mimic jQuery's cache: false.
```
fichtner, line 264:
```suggestion
```

fichtner:
Merged, thanks!

Issue #5795, opened 2022-05-30
Unbound: Support Additional Serve Expired Settings


iamperson347:
**Important notices**

Before you add a new report, we ask you kindly to acknowledge the following:

- [X] I have read the contributing guide lines at https://github.com/opnsense/core/blob/master/CONTRIBUTING.md
- [X] I am convinced that my issue is new after having checked both open and closed issues at https://github.com/opnsense/core/issues?q=is%3Aissue

**Is your feature request related to a problem? Please describe.**

No

**Describe the solution you like**

Adjust the Unbound Advanced Settings page (/services_unbound_advanced.php) to support the following new Unbound settings:
- serve-expired-ttl
- serve-expired-client-timeout
- serve-expired-reply-ttl

**Describe alternatives you considered**

Setting these settings in `/usr/local/etc/unbound.opnsense.d/` (which does work)

**Additional context**

dacioromero:
@swhite2 I'm admittedly new to OPNsense, but it doesn't seem like the additional serve expired settings were hooked up to `src/opnsense/service/templates/OPNsense/Unbound/core/advanced.conf` in #6053

swhite2:
@dacioromero Must've slipped through, https://github.com/opnsense/core/commit/9966db9946b28b15b3df49c9be009c4641c93238

PR #5796, opened 2022-05-30
MR for Feature Request "Unbound: Support Additional Serve Expired Settings"

iamperson347:
This MR adds new fields in allow setting new unbound settings outlined in the linked issue.

#5795 

NOTE: I'm not familiar with the opnsense codebase - so this should be reviewed carefully to ensure I didn't miss any other areas that may need updates. Initial testing on my VM seems ok.


AdSchellevis:
Maybe we better rewrite this page to MVC first and add these options next, as the current "advanced" page lacks validation and doesn't come with a huge amount of options. Adding more options now would push the target for cleaning this up further away I'm afraid.

AdSchellevis:
I suppose the running version isn't the same as you're trying to build. The docs is a good place to start, probably combined with examples in the plugins.

PR #5812, opened 2022-06-11
mvc: parseFormNode extension

agh1467:
I don't expect this to be merged, but I'd like to understand if this is a good approach to making these improvements. I use this function extensively in one of my plugins and would like to have the functionality in core to aid the expansion of the layout_partials which hasn't had any significant improvements or added functionality for several years.

While working on one of my plugins I encountered limitations with the information that could be pulled out of the form XMLs. Almost everything I wanted to do in Volt I wasn't able to do because the information I'd put into the XML (nested elements, element attributes) wasn't getting into the environment. Without massively expanding the structure of the XML, and hijacking elements to store what should be element attribute values, there wasn't another way that I knew to accomplish what I was trying to do.

This change expands the functionality of the parseFormNode function which originally performs only a single recursion, and stores any further elements as a key/value (which doesn't work for nested elements). This extended version does primarily two things, recurses infinitely (risky?), and also grabs element attributes and stores them in the array along side the string value (or child arrays).

One of the challenges I encountered while working on this was the fact that the structure allows the possibility that a given object may either be a key/value pair (variable), or an array. This ambiguity is sometimes troublesome when writing the Volt templates because any conditions have to be written in such a way that tests if the object is an array or not before performing any evaluation of the object.

Could anyone provide to me background on why the SimpleXML object isn't just translated to an array wholesale, rather than what is done here with walking the tree and manually re-creating it with a slightly alternative structure? I can only speculate that maybe a limited outcome was desired.

I've included extensive comments for clarity only and wouldn't expect them to be necessary if something like this were to be merged. The comments include XML->Array examples for additional clarification. Note that the "field" case was removed because the functionality provided by that case is provided in the default case instead (though not by name, instead by virtue of the `<field>` element existing in the form XML). The base_form.volt needed a slight modification as the "field" array is no longer assumed as all node types are recursed equally, but instead the array for the fields is called "field" as that's the name of the XML element `<field>`.

I looked around for tests for this PHP but didn't see anything in the tests directory. I did a cursory click through in the UI without any trouble. I've used this function in my plugin for a while, and the complexity of the plugin's XML covers most of what would be included in test cases. That's no substitution for proper testing though.

This added functionality goes beyond what's simply described above. It enables additional structures to be defined and handled in the volt templates via functions, and I've even used them to build Javascript functions as well as defining Javascript attachments. It allows for the possibility of significantly extending the volt templates, and eases plugin development by shifting the burden away from writing static HTML/Javascript. Below are some examples from my plugin which utilize this function:

Defining the entirety of a bootgrid along with definition of the dialog for that bootgrid within the same form XML instead of defining these things manually as HTML and Javascript, and in separate form XML files. This _significantly_ reduces code footprint.
``` xml
<field>
    <target>allowed_ips_internal.entries</target>
    <label>Internal Allowed IPs List</label>
    <help><![CDATA[
        The allow list is managed entirely in OPNsense. This allows
        for adding comments, editing individual entries, searching,
        sorting, as well as quick additions and removals along with
        importing and exporting the entire list.
        ]]>
    </help>
    <type>bootgrid</type>
    <edit>true</edit>
    <columns>
        <select>true</select>
        <column id="expression" width="" size="" type="string" visible="true" data-formatter="">Expression</column>
        <column id="comment" width="" size="" type="string" visible="true" data-formatter="">Comment</column>
    </columns>
    <api>
        <search>/api/dnscryptproxy/settings/grid/search</search>
        <get>/api/dnscryptproxy/settings/grid/get</get>
        <set>/api/dnscryptproxy/settings/grid/set</set>
        <add>/api/dnscryptproxy/settings/grid/add</add>
        <del>/api/dnscryptproxy/settings/grid/del</del>
        <toggle>/api/dnscryptproxy/settings/grid/toggle</toggle>
        <import>/api/dnscryptproxy/settings/import</import>
        <export>/api/dnscryptproxy/settings/export</export>
    </api>
    <row_count>7,14,20,50,100,-1</row_count>
    <dialog>
        <label>Edit Entry</label>
        <field>
            <id>entries.enabled</id>
            <label>Enabled</label>
            <type>checkbox</type>
            <help><![CDATA[
                This will enable or disable the allow list entry.
                ]]>
            </help>
        </field>
        <field>
            <id>entries.expression</id>
            <label>Expression</label>
            <type>text</type>
            <help><![CDATA[
                Set the domain, IP or expression to allow list,
                e.g. ads.* or *.example.com
                ]]>
            </help>
            <hint>*.example.com</hint>
        </field>
        <field>
            <id>entries.comment</id>
            <label>Comment</label>
            <type>text</type>
            <help><![CDATA[
                Optional comment.
                ]]>
            </help>
        </field>
    </dialog>
</field>
```

Defining radio buttons:
``` xml
<field>
    <id>settings.blocked_names_type</id>
    <label>Blocked Names List type</label>
    <type>radio</type>
    <help><![CDATA[
        Select which blocklist type to use:<br>
        <b>Internal</b> - blocklist entries are managed within
        OPNsense via the list on this page.<br>
        <b>External</b> - Specify the filesystem path to a blocklist
        file external to dnscrypt-proxy.<br>
        <b>Manual</b> - Manage the blocklist manaully by uploading
        a blocklist.
        ]]>
    </help>
    <buttons>
        <button value="external">External</button>
        <button value="manual">Manual</button>
        <button value="internal">Internal</button>
    </buttons>
</field>
```

Defining actions to control fields when specific pre-defined events occur:
``` xml
<field>
    <id>settings.nx_log.enabled</id>
    <type>checkbox</type>
    <hidden>true</hidden>
    <control>
        <action on_set="true" type="tab" do_state="visible">tab_header_dnscryptproxy_logs_nx</action>
        <action on_set="false" type="tab" do_state="hidden">tab_header_dnscryptproxy_logs_nx</action>
    </control>
</field>
```

Adding a button (which has an additional two sub-buttons defined), which includes definitions for some style, and identifier attributes.
``` xml
<button type="group" icon="fa fa-floppy-o" label="Save All Settings" id="save_actions">
    <dropdown action="save_all" icon="fa fa-floppy-o">Save Only</dropdown>
    <dropdown action="save_apply_all" icon="fa fa-floppy-o">Save and Apply</dropdown>
</button>
```

Here are some volt templates from my plugin which showcase the reduced codebase required for a plugin:
Simple example: [diagnostics.volt](https://github.com/agh1467/plugins/blob/dnscrypt-proxy-ng/dns/dnscrypt-proxy-ng/src/opnsense/mvc/app/views/OPNsense/Dnscryptproxy/diagnostics.volt)
More complex: [settings.volt](https://github.com/agh1467/plugins/blob/dnscrypt-proxy-ng/dns/dnscrypt-proxy-ng/src/opnsense/mvc/app/views/OPNsense/Dnscryptproxy/settings.volt)

### DIFF
--- a/src/opnsense/mvc/app/controllers/OPNsense/Base/ControllerBase.php
+++ b/src/opnsense/mvc/app/controllers/OPNsense/Base/ControllerBase.php
@@ -51,6 +51,11 @@ class ControllerBase extends ControllerRoot
     {
         $result = array();
         foreach ($xmlNode as $key => $node) {
+            $element = array();
+            $nodes = $node->children();
+            $nodes_count = $nodes->count();
+            $attributes = $node->attributes();
+
             switch ($key) {
                 case "tab":
                     if (!array_key_exists("tabs", $result)) {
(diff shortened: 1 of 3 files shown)


AdSchellevis:
You can easily ship your own version of the controller deriving from ControllerBase (extend) or ControllerRoot (ancestor of all controllers).  I don't think we currently have much more need for further nesting as our tab navigation doesn't offer more layers either, which is why we added these in the first place if memory serves right (https://github.com/opnsense/core/commit/f730946c036aca7f529139abc09187228b5afbc3).

Adding xml templates which simplify plugin development isn't a bad idea at all, but is just quite some work to make it easy yet still extendable and requires work on documentation as well (plus refactoring existing pages so people are more likely to use them). Currently I don't expect we can spare the time for that to be honest, existing patterns can easily be copied and we moved more functionality into easy to use wrappers (for the Javascript additions) to ease development as well.

agh1467:
Thanks for the feedback and taking a look! I had no idea the previous method was to write out an entire array inside the volt which is definitely much more cumbersome than the XML approach.

Currently I provide a version of the [ControllerBase.php](https://github.com/agh1467/plugins/blob/dnscrypt-proxy-ng/dns/dnscrypt-proxy-ng/src/opnsense/mvc/app/controllers/OPNsense/Dnscryptproxy/ControllerBase.php) inside my plugin to use `getForm()` which calls `parseFormNode()`. My initial concern with that is drift from core's version of that file (one more thing to maintain). It currently sits along side the other controller PHP classes.

I'm not sure if it's the correct, or preferred way to do this, but it was the first way I found that worked. This is how I call it within another class:
https://github.com/agh1467/plugins/blob/52ca20d47721532b2498acf4f6acd44180e09447/dns/dnscrypt-proxy-ng/src/opnsense/mvc/app/controllers/OPNsense/Dnscryptproxy/SettingsController.php#L52-L62

I'll go ahead and close this.

Issue #5478, opened 2022-01-15
Bootgrid: command-info requires get instead of info URL


8191:
In the OPNsense UIBootgrid, the "info" action is only wired when a "get" URL is set:

https://github.com/opnsense/core/blob/88334b838a13c7083f5ad29044ff65d4bb5f14ba/src/opnsense/www/js/opnsense_bootgrid_plugin.js#L83-L86

Is this intended; or should it require the "info" URL, which is later actually used in the event handler?

https://github.com/opnsense/core/blob/88334b838a13c7083f5ad29044ff65d4bb5f14ba/src/opnsense/www/js/opnsense_bootgrid_plugin.js#L351-L354

8191:
"command-info" is not used at all so far, as I can see. So should have no impact to change.

AdSchellevis:
It's used in
https://github.com/opnsense/plugins/blob/16f3522d08d30919b17e66bdec38352ef4c75208/net/zerotier/src/opnsense/mvc/app/views/OPNsense/Zerotier/index.volt#L48
 https://github.com/opnsense/plugins/blob/16f3522d08d30919b17e66bdec38352ef4c75208/net/zerotier/src/opnsense/mvc/app/views/OPNsense/Zerotier/index.volt#L104 


Maybe also in some private repo's but I'm not 100% sure about that.

8191:
> It's used in
> https://github.com/opnsense/plugins/blob/16f3522d08d30919b17e66bdec38352ef4c75208/net/zerotier/src/opnsense/mvc/app/views/OPNsense/Zerotier/index.volt#L48
> https://github.com/opnsense/plugins/blob/16f3522d08d30919b17e66bdec38352ef4c75208/net/zerotier/src/opnsense/mvc/app/views/OPNsense/Zerotier/index.volt#L104

You're totally right. I grep'ed for `command-info` which is not really useful in that case. ;-)

What's your opinion in changing the `requires: ['get']` to `requires: ['info']` in
https://github.com/opnsense/core/blob/88334b838a13c7083f5ad29044ff65d4bb5f14ba/src/opnsense/www/js/opnsense_bootgrid_plugin.js#L83-L86

AdSchellevis:
> What's your opinion in changing the requires: ['get'] to requires: ['info'] in

Sounds like a good plan indeed.

agh1467:
@AdSchellevis I forgot to use the 'closes' tag in the commit. Can you close this issue manually instead?

[...]

PR #5829, opened 2022-06-25
ui: bootgrid, correct required api for command-info #5478

agh1467:
* command-info uses this.command_info method, which expects
  params['info'] to be populated, not 'get'.

I stumbled across this issue ticket and saw that this wasn't addressed yet. I looked into it and I think it should be changed to info, as the method does use 'info', and not the 'get' param. If 'info' is not provided, the method still executes without error, but ends up with a literal `undefined` as the URL in the ajaxGet call. Which gets caught by the API error handler as action not found.

https://github.com/opnsense/core/blob/91b840d5d2ae0d274c3b7b105d2a76bef2209598/src/opnsense/www/js/opnsense_bootgrid_plugin.js#L333-L347

I expect anyone that's using `commandsWithInfo` data-formatter will have an info URL defined, since the button won't do anything without it.

### DIFF
--- a/src/opnsense/www/js/opnsense_bootgrid_plugin.js
+++ b/src/opnsense/www/js/opnsense_bootgrid_plugin.js
@@ -82,7 +82,7 @@ $.fn.UIBootgrid = function (params) {
             },
             "command-info": {
                 method: this_grid.command_info,
-                requires: ['get']
+                requires: ['info']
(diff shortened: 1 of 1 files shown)


AdSchellevis:
@agh1467 thanks! this looks reasonable indeed.

PR #5830, opened 2022-06-26
MVC - model: throw when no mount found

agh1467:
* Return a useful message to the UI if no mount is found in the model XML.
  Otherwise this condition results in a SimpleXMLElement error, trying to
  parse an empty string in toXml(), `$xml = new SimpleXMLElement($xml_root_node);`

I followed the pattern of error handling of the two throws above it. I'm not sure about the message, but it should be clear enough.

Message looks like:
```
/usr/local/opnsense/mvc/app/models/OPNsense/Base/BaseModel.php:323: model xml /usr/local/opnsense/mvc/app/models/OPNsense/Myplugin/Settings.xml missing mount definition
```


AdSchellevis:
@agh1467 looks reasonable, thanks!

Issue #2435, opened 2018-06-01
MVC / 404 (page not found) not functional when namespace or controller routed


AdSchellevis:
When the namespace or the controller is routed, but the destination is not found, an exception occurs which seem to be hard to catch and forward to our usual not_found page.

First the symptoms, the errors below are the ones we should fix to avoid people getting to a page which is difficult to navigate out (not menu system): 
```
XXXXController handler class cannot be loaded
```

```
Action 'xxx' was not found on handler 'xxx'
```

Original idea was to use the exceptions described at phalcon https://docs.phalconphp.com/en/latest/dispatcher#handling-not-found-exceptions , which we also use for the api endpoints (https://github.com/opnsense/core/blob/18.1.8/src/opnsense/mvc/app/config/services_api.php#L95-L124). 

For some odd reason, it seem to be impossible to `pick()` a view and render it, as it always leads to an empty page. An escape could be to trap and `redirect()` (in stead of `forward()`) to a page everyone can reach, but it adds more complexity and you loose the url in the navigation.

AdSchellevis:
test

AdSchellevis:
This issue has been automatically timed-out.

For more information about the policies for this repository, 
please read `CONTRIBUTING.md` which is located in the root of this repository.

If someone wants to step up and work on this issue, 
just let us know, so we can reopen the issue and assign an owner to it.

AdSchellevis:
This issue has been automatically timed-out (after 180 days of inactivity).

For more information about the policies for this repository, 
please read https://github.com/opnsense/core/blob/master/CONTRIBUTING.md which is located in the root of this repository.

If someone wants to step up and work on this issue, 
just let us know, so we can reopen the issue and assign an owner to it.

AdSchellevis:
This issue has been automatically timed-out (after 180 days of inactivity).

For more information about the policies for this repository, 
please read https://github.com/opnsense/core/blob/master/CONTRIBUTING.md for further details.

If someone wants to step up and work on this issue, 
just let us know, so we can reopen the issue and assign an owner to it.

AdSchellevis:
This issue has been automatically timed-out (after 180 days of inactivity).

For more information about the policies for this repository, 
please read https://github.com/opnsense/core/blob/master/CONTRIBUTING.md for further details.

If someone wants to step up and work on this issue, 
just let us know, so we can reopen the issue and assign an owner to it.

PR #5831, opened 2022-06-26
MVC: revise route to default

agh1467:
This logic was originally for an issue #2435

* Original change on commit 1bdef70

* Adjust logic so that exceptions are visible to the user.
  Use some style so that it's easier to read.

I'm leery of changing this logic without additional scrutiny. This might not even be the correct or best way to handle it. The intent here is to present the exceptions that are thrown inside views (and controllers) to the user. Right now, the logic will always catch the first `if`, and simply draw the default page, because `$application` will always be set if the controller was able to load properly. The exception data however is not included in the output.

I found this through digging, and found that a typo in my form XML was causing the issue. However, I thought it was weird that an exception is thrown accordingly, but nothing appears to happen with it. For example, these throws here are never seen:
https://github.com/opnsense/core/blob/59c2059eb206440ab6a7ffa1921c221839e908c5/src/opnsense/mvc/app/controllers/OPNsense/Base/ControllerBase.php#L95-L114

I'd guess that _any_ throws occurring will never get seen when loaded by `index.php`.

It would be nice to actually show a page, with a dialog box like when an API exception occurs, but it's a bit out of scope for me.

**Maybe related**, attempting to load a non-existent page like "/ui/noexist" doesn't appear throw at all. So the comment doesn't sound like it correlate with what's happening: `// Render default UI page when controller or action wasn't found` As that describes the situation of no controller found but the catch doesn't happen with that condition.

### DIFF
--- a/src/opnsense/www/index.php
+++ b/src/opnsense/www/index.php
@@ -72,14 +72,18 @@ try {
     echo $application->handle($_SERVER['REQUEST_URI'])->getContent();
 } catch (\Exception $e) {
     if (
-        isset($application) || (
-          stripos($e->getMessage(), ' handler class cannot be loaded') !== false ||
-          stripos($e->getMessage(), ' was not found on handler ') !== false
+        stripos($e->getMessage(), ' handler class cannot be loaded') !== false ||
+        stripos($e->getMessage(), ' was not found on handler ') !== false
         )
-    ) {
+    {
         // Render default UI page when controller or action wasn't found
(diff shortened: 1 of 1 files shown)


AdSchellevis:
With the migration to phalcon 5 I'll rather leave it as is for now. Long term I'm not sure we keep using Phalcon for these functions as quite some of the logic might be easier to rewrite than migrate on every upgrade. Keeping this open for now so we can decide what to do later.

Issue #5808, opened 2022-06-07
Aliases "Enable / Disable" checkbox no longer functions since update to 22.1.8_1


copperheadclgp:
**Important notices**

Before you add a new report, we ask you kindly to acknowledge the following:

- [X] I have read the contributing guide lines at https://github.com/opnsense/core/blob/master/CONTRIBUTING.md
- [X] I am convinced that my issue is new after having checked both open and closed issues at https://github.com/opnsense/core/issues?q=is%3Aissue

**Describe the bug**
Since upgrading to 22.1.8_1-amd64, the ability to enable and disable Aliases using a combination of checkbox + Apply button no longer functions. There is a new column that appears next to the checkbox that shows either 1 (enabled) or 0 (disabled) but this value does not change.

**To Reproduce**

1. Go to 'Firewall > Aliases'
2. Click on the checkbox for an alias
3. Click 'Apply'
4. Observe that the alias does not enable or disable as expected.

**Expected behavior**

Should behave as in previous releases, wherein the administrator can select one or more aliases and enable / disable them en-mass. 

**Describe alternatives you considered**

Aliases can be enabled / disabled singularly as expected via the 'Edit' button.

**Screenshots**

[image](https://user-images.githubusercontent.com/51468137/172453374-254b98c0-c636-45fe-83e2-65abc9677efc.png)

**Environment**

OPNsense 22.1.8_1 (amd64, OpenSSL).
AMD GX-412TC SOC 
PCEngines APU2

fichtner:
3dac44b77a maybe?

    # opnsense-patch 3dac44b77a


Cheers,
Franco

PR #5833, opened 2022-06-27
ui: bootgrid override formatters

agh1467:
Keep default formatters, but allow them to be overridden with options.

I'm not an expert in Javascript, so this may not be the best way to get the intended result.

This intends to allow plugin authors to define new formatters without also having to re-define the default ones. In the case where a custom formatter is desired along with the desire to use one of the default formatters at the same time.

This has manifested recently in #5808, and was fixed by re-defining the `rowtoggle` formatter in 3dac44b within the plugin. This should address this type of issue in the future.

This should allow for two things:
1. Define new formatters without wiping out the default ones.
2. Override default formatters without wiping out the other default ones.

I looked at the other params/options, and formatters seemed to be the only one where this kind of accommodation would be necessary.

### DIFF
--- a/src/opnsense/www/js/opnsense_bootgrid_plugin.js
+++ b/src/opnsense/www/js/opnsense_bootgrid_plugin.js
@@ -140,7 +140,12 @@ $.fn.UIBootgrid = function (params) {
         // merge additional options (if any)
         if (params['options'] !== undefined) {
             $.each(params['options'],  function(key, value) {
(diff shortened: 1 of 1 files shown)


AdSchellevis:
@agh1467 good point, I think this https://github.com/opnsense/core/commit/79f5d8f3fa5c484ab2e20797bd9586a585bd592a should do the trick as well but is slightly more generic.

PR #5837, opened 2022-06-29
fix IntegerValidator returning false for negative numbers

xbb:
Fixes validation for valid negative integers

Uses FILTER_VALIDATE_INT which should cover all the cases including min/max int value


AdSchellevis:
@xbb looks good, thanks!

PR #5838, opened 2022-06-29
fix IntegerField minimum value

xbb:
Set the minimum value using `PHP_INT_MIN` which  should be equal to `-PHP_INT_MAX - 1`


agh1467:
I found the commit where this was introduced, there is a short conversation there which might add context to the reason why it was that way.
https://github.com/opnsense/core/commit/06ff13943b3cddc378005bcd8995c105bb8057a0

xbb:
> I found the commit where this was introduced, there is a short conversation there which might add context to the reason why it was that way. [06ff139](https://github.com/opnsense/core/commit/06ff13943b3cddc378005bcd8995c105bb8057a0)

I see, `PHP_INT_MIN` was introduced in PHP 7.0 and opnsense at that time was using, I think, PHP 5.6

AdSchellevis:
@xbb thanks!

PR #5839, opened 2022-06-29
Sort Order - Firewall : Aliases

NOYB:
Sort Order - Firewall : Aliases

Aliases are sorted case sensitively.  Causing un-natural sort order.  myAlias should come before YourAlias.  Always.

Sort order should be:
my Alias
Your Alias

Not:
YourAlias
myAlias


AdSchellevis:
not sure about this, if it should be standard behaviour, the alias controller might not be the right spot

NOYB:
Whether it should be the behavior; It's a human interface.  It should accommodate humans in the way that humans work.  As humans we don't typically sort lists in a case sensitive manner unless there is a specific need/reason to do so.  When looking in a sorted list for "beUserFriendly" we don't typically expect to find it after "Zebra".  Yes I missed an alias because of the case sensitive sort order.

This is where it is being done here until something else comes along.

AdSchellevis:
https://github.com/opnsense/core/commit/c929c73a50de525e6b8a192b11b6b8ad169f2850

Issue #5845, opened 2022-07-01
Opendns has wrong ipv6 dns server address


jrylander:
**Important notices**

Before you add a new report, we ask you kindly to acknowledge the following:

- [x] I have read the contributing guide lines at https://github.com/opnsense/core/blob/master/CONTRIBUTING.md
- [x] I am convinced that my issue is new after having checked both open and closed issues at https://github.com/opnsense/core/issues?q=is%3Aissue

**Describe the bug**

The OpenDNS plugin uses array('2620:0:ccc::2', '2620:0:ccd::2') but should be using: 
array('2620:119:35::35', '2620:119:53::53') according to https://www.opendns.com/about/innovations/ipv6/

I have tried with the changed settings and verified that it works whereas it did not with existing addresses.

**To Reproduce**

Activate OpenDNS and try lookup a blocked domain, e.g. "sex.com". Blocking will work with suggested ipv6 servers but not with existing servers.

**Expected behavior**

Blocked domain should return ip for OpenDNS blocked page.

**Describe alternatives you considered**

-

**Screenshots**

-

**Relevant log files**

-

**Additional context**

-

**Environment**

  |Version | 22.1.9 |  
  | Architecture | amd64 |  
  | Flavour | OpenSSL |  
  | Commit | 986c2a83b |  
  | Mirror | https://pkg.opnsense.org/FreeBSD:13:amd64/22.1 |  
  | Repositories | OPNsense |  
  | Updated on | Sun Jun 26 21:51:26 CEST 2022

jrylander:
https://github.com/opnsense/core/pull/5846

Issue #434, opened 2015-10-15
Request to add IPv6 OpenDNS resolvers to OpenDNS configuration


StopTrain:
Setting up OpenDNS for the first time on OPNsense and all is working, but I noticed it is only showing their IPv4 resolvers.   Researched and found they do have IPv6 addresses as well.

Link to the article:
https://support.opendns.com/entries/26056194-Does-OpenDNS-support-IPv6-

The addresses are:

2620:0:ccc::2 
2620:0:ccd::2

Thank you!

sjorge:
:+1: on this

StopTrain:
Totally agree.  IPv6 is really making people realize they are behind :+1:

PR #5846, opened 2022-07-01
Update OpenDNS ipv6 servers

jrylander:
https://github.com/opnsense/core/issues/5845

The OpenDNS plugin uses array('2620:0:ccc::2', '2620:0:ccd::2') but should be using:
array('2620:119:35::35', '2620:119:53::53') according to https://www.opendns.com/about/innovations/ipv6/

I have tried with the changed settings and verified that it works whereas it did not with existing addresses.


AdSchellevis:
Looks good, referencing https://github.com/opnsense/core/issues/434 as the unfiltered dns (https://support.opendns.com/hc/en-us/articles/227986667-Does-OpenDNS-Support-IPv6-) was added there. Thanks!

PR #5848, opened 2022-07-04
MVC: SimpleActionButton, add catch undefined TypeError

agh1467:
* Add condition for data existing before others
  catches Uncaught TypeError: data.status is undefined

I encountered this when I was using SimpleActionButton with a custom API that doesn't have 'status' in the response.

Adding a condition that requires `status` key existing in `data` appears to mitigate the error. I assume that it doesn't evaluate the rest of the conditions with this one failing. Since `data['status']` is used later on, this will require that it exists first.


AdSchellevis:
@agh1467 thanks!

PR #5863, opened 2022-07-10
Fix "Allow service binding" for multiple aliases per interface

ADawidowski:
When using multiple aliases per interface, disabling binding on one could potentially disable binding for other aliases on the same interface, depending on the order of the VIPs in the config. The 'alias' setting was evaluated regardless of whether the subnet matched, so if a previous VIP for a matching interface had matched on subnet, the current VIP's 'bind' setting would be applied to the interface address even though the current VIP's subnet didn't match.

### DIFF
--- a/src/etc/inc/interfaces.inc
+++ b/src/etc/inc/interfaces.inc
@@ -4322,18 +4322,24 @@ function interfaces_addresses($interfaces, $as_subnet = false, $ifconfig_details
                 continue;
             }
 
+            $matched_subnet = false;
             if ($info['family'] == 'inet' && strpos($vip['subnet'], ':') === false) {
-                $info['alias'] = $info['alias'] || $vip['subnet'] == $info['address'];
+                $matched_subnet = $vip['subnet'] == $info['address'];
             } elseif ($info['family'] == 'inet6' && strpos($vip['subnet'], ':') !== false) {
                 /*
                  * Since we do not know what subnet value was given by user
                  * uncompress/compress to match correctly compressed system
                  * value.
                  */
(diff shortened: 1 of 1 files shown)


fichtner:
I know we had an issue before and something was fixed so I need to make really sure this is correct…

ADawidowski:
I only started looking at the code yesterday, so I'm not implying this fix is 100% correct. Please, take your time and let me know if I can help in any way.

fichtner:
Maybe you have an example with two aliases where this overlaps. That would certainly help.

Thanks,
Franco

ADawidowski:
I'm running OPNsense with CARP and IP aliases on my LAN interface - both IPv4 and IPv6, but judging by what the last part of interfaces_addresses is doing, it seems pretty easy to reproduce using a simpler config:
- configure 2 IPv4 aliases on LAN ( I dunno: 10.190.0.0/24 and 10.191.0.0/24)
- configure Unbound DNS
- examine Services -> Unbound DNS -> Access Lists - both aliases should be there.
- disable "Allow service binding" on the first alias and check Access List again
- enable "Allow service binding" on the first alias, disable on the second one and check Access List again.
At least one of the last two checks will yield unexpected results: both addresses will be missing from the list instead of only the one that has binding disabled.

In my somewhat complicated case I had an IPv6 alias set up on an interface where IPv4 CARP is also configured. Disabling binding on the alias prevented Unbound from listening on the CARP address as well.

fichtner:
Committed via aa86b9d43d as a slightly modified version (style changes). It's correct that if we iterate multiple times through info we might cross a virtual IP that has nobind on and it triggers on the wrong condition:

https://github.com/opnsense/core/blob/b8c6c182025eb49f93f419a2be4ad9b914c2710c/src/etc/inc/interfaces.inc#L4336


Thanks,
Franco

PR #5872, opened 2022-07-14
interfaces: Fix unable to bring up multiple loopback

johnnyslee:
This fixes a bug:
* Loopback interfaces other than the one being configured/created
  are all going to be removed/destroyed.


johnnyslee:
To reproduce the bug, simply add and assign multiple loopback interfaces, and you'll see only the last one (or the one just being modified) would be brought up.

fichtner:
Nice catch, thanks a lot!

fichtner:
FYI VXLAN had the same issue ddb4af9040d4 since VXLAN was copied from loopback code back in the day :)

PR #5875, opened 2022-07-14
System Status: replace old notices system with a global one

swhite2:
- Remove the old notices system.
- Add a single clickable status icon which reflects the current system status.
- Remove the old "A problem has been detected." banner from the dashboard and instead notify the user globally with the aforementioned icon.
- Currently the crash reporter and the firewall are implemented. Like the old notice system, should PF get stuck or any filter rule fail to load, the user will be warned via this system instead.
- Extensible by nature, more subsystems can be added in the future to accommodate notification and referral needs.
- Implemented in both legacy and MVC code (duplicated JS code is necessary in this case).
- System status can be fetched via an API call.

Things to be discussed:
- ~~ACL check in the SystemController (a user should only be able to see notifications for areas it has access to)~~
- Locale in dynamically generated JQuery objects.


src/etc/inc/config.inc
```diff
@@ -36,7 +36,6 @@ openlog($g['product_id'], LOG_ODELAY, LOG_USER);
 register_shutdown_function('closelog');
 
 require_once("xmlparse.inc");
-require_once("notices.inc");
```
fichtner, line 39: "make plist-fix" :)

src/etc/inc/filter.inc
```diff
@@ -487,13 +487,13 @@ function filter_configure_sync($verbose = false, $load_aliases = true)
         if (strstr("$rules_error[0]", "busy")) {
             exec('/sbin/pfctl -d; /sbin/pfctl -e; /sbin/pfctl -f /tmp/rules.debug');
             log_error('PF was wedged/busy and has been reset.');
-            file_notice(gettext('PF was wedged/busy and has been reset.'));
+            file_put_contents('/tmp/rules.warning', '/ui/diagnostics/log/core/system');
```
fichtner, line 490: Although it's nice to point to a page I think this is somehow losing context about what exactly threw the notice here. Maybe a syslog message prefix wold make sense to filter for this when opening the page?

swhite2: It could also make sense to split off and create a new log facility, requiring less technical details since the current state of the log system doesn't implement passing data to it (to able to filter) from a referring page. Anything else in the firewall section can refer to this as well (such as aliases). Maybe call it "General"?

fichtner: May be better to try and see this in action first. I'm unsure what you mean here.

src/opnsense/mvc/app/controllers/OPNsense/Core/Api/SystemController.php
```diff
@@ -56,4 +57,59 @@ public function rebootAction()
...
+
+        return json_encode($statuses);
+    }
+
+    public function dismissStatusAction()
+    {
+        if ($this->request->isPost() && $this->request->hasPost("subject") && $this->request->hasPost("acl")) {
+            $acl = new ACL();
+            $aclCheck = $this->request->getPost("acl");
+            if (!$acl->isPageAccessible($this->getUserName(), $aclCheck)) {
```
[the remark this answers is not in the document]
swhite2: Hopefully https://github.com/opnsense/core/pull/5875/commits/f531822fcebead6ff381b857b65d5a2104d0ef1c is acceptable.

src/opnsense/mvc/app/controllers/OPNsense/Core/Api/SystemController.php
```diff
@@ -56,4 +57,65 @@ public function rebootAction()
...
+        sort($statusCodes);
+        $statuses['System'] = [
+            'status' => $order[$statusCodes[0]]
+        ];
+
+        return json_encode($statuses);
+    }
+
+    public function dismissStatusAction()
+    {
```
fichtner, line 95:
```suggestion
    {
        $this->sessionClose();
    
```

src/opnsense/mvc/app/controllers/OPNsense/Core/Api/SystemController.php
```diff
@@ -56,4 +57,65 @@ public function rebootAction()
...
+        if ($this->request->isPost() && $this->request->hasPost("subject")) {
+            $acl = new ACL();
+            $backend = new Backend();
+            $subsystem = $this->request->getPost("subject");
+            $system = json_decode(trim($backend->configdRun('system status')), true);
+            if (!array_key_exists($subsystem, $system)) {
+                return ["status" => "status type does not exist"];
```
fichtner, line 102: information disclosure before ACL check

src/opnsense/mvc/app/controllers/OPNsense/Core/Api/SystemController.php
```diff
@@ -56,4 +57,65 @@ public function rebootAction()
...
+            }
+            if (isset($system[$subsystem]['logLocation'])) {
+                $aclCheck = $system[$subsystem]['logLocation'];
+                if (!$acl->isPageAccessible($this->getUserName(), $aclCheck) ||
+                    $acl->hasPrivilege($this->getUserName(), 'user-config-readonly')) {
+                    return ["status" => "user not allowed to dismiss status"];
```
fichtner, line 108: same here, security issue

src/opnsense/mvc/app/controllers/OPNsense/Core/Api/SystemController.php
```diff
@@ -56,4 +57,65 @@ public function rebootAction()
...
         ];
     }
+
+    public function statusAction()
+    {
+        $backend = new Backend();
+        $statuses = json_decode(trim($backend->configdRun('system status')), true);
+
+        if (!$statuses) {
+            return array("message" => "Unable to run configd action");
```
fichtner, line 67: also security issue

fichtner:
as a general note the acl/status workflow is probably similar for display and dismiss and since this is security relevant should be refactored into a single private function called from both actions

src/opnsense/mvc/app/controllers/OPNsense/Core/Api/SystemController.php
```diff
@@ -56,4 +57,66 @@ public function rebootAction()
...
+            $statusCodes = array_map(function($v) {
+                return $v['statusCode'];
+            }, array_values($statuses));
+            sort($statusCodes);
+            $statuses['System'] = [
+                'status' => $order[$statusCodes[0]]
+            ];
+
+            return json_encode($statuses);
+        }
```
fichtner, line 90: seem to be missing a default return

AdSchellevis:
@swhite2 added my modifications as promised in https://github.com/opnsense/core/pull/5875/commits/66949d2017e4f57268d5ffdf73b62f8dc0e4a256

PR #5880, opened 2022-07-15
Disable IPv6 inside 4in6 and 4in4 gif tunnels

maurice-w:
When setting up an IPv4-over-IPv6 or IPv4-over-IPv4 gif tunnel, IPv6 should be disabled inside the tunnel to prevent the automatic configuration of an IPv6 link-local address. From man gif(4):

> Note that IPv6 link-local addresses (those that start with fe80::) will be automatically configured whenever possible. You may need to remove IPv6 link-local addresses manually	using ifconfig(8), if you want to disable the use of IPv6 as the inner header (for example, if you need a pure IPv4-over-IPv6 tunnel).


fichtner:
Sounds good, thanks!

PR #5894, opened 2022-07-21
Firewall: redirect alias syslog messages to new general firewall log

swhite2:
Facilitates https://github.com/opnsense/core/pull/5875 to notify users in a better defined context.

### DIFF
--- a/src/etc/inc/filter.inc
+++ b/src/etc/inc/filter.inc
@@ -334,6 +334,8 @@ function filter_configure_sync($verbose = false, $load_aliases = true)
         }
     }
 
+    openlog("firewall", LOG_DAEMON, LOG_LOCAL4);
+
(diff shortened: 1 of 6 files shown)


AdSchellevis:
@swhite2 I think you missed a spot :) https://github.com/opnsense/core/blob/master/src/etc/inc/filter.inc, but this is basically what I meant indeed. we can discuss about the naming, but further than that it should be as simple as this :)

fichtner (approved):
sure :)

menu ordering needs some adjustments but can do that later

AdSchellevis:
ohw, yes, and a syslog registration, almost forgot :) similar to 
https://github.com/opnsense/core/blob/216e489564711afceb6c3bda515d64e6091a3b15/src/etc/inc/plugins.inc.d/core.inc#L249-L269

PR #5920, opened 2022-08-02
unbound: do not start DHCP watcher immediately after daemonizing unbound

swhite2:


src/opnsense/scripts/unbound/start.sh
```diff
@@ -61,3 +63,7 @@ chown -R unbound:unbound /var/unbound
 
 /usr/local/sbin/unbound -c /var/unbound/unbound.conf
 /usr/local/opnsense/scripts/unbound/cache.sh load
+
+if [ ! -z "${DOMAIN}" ]; then
```
fichtner, line 67:
```suggestion
if [ -n "${DOMAIN}" ]; then
```

;)

src/opnsense/scripts/unbound/start.sh
```diff
@@ -61,3 +63,7 @@ chown -R unbound:unbound /var/unbound
...
+  /usr/local/opnsense/scripts/dhcp/unbound_watcher.py --domain ${DOMAIN}
```
fichtner, line 68: might want to check tabs vs spaces here, I "try" to use tabs for shell scripts and it looks mixed in here

fichtner (approved):
ok minus comments :)

Issue #5919, opened 2022-08-02
DNSSL set in radvd.conf although sending DNS is disabled in UI


pmhausen:
- [X] I have read the contributing guide lines at https://github.com/opnsense/core/blob/master/CONTRIBUTING.md
- [X] I am convinced that my issue is new after having checked both open and closed issues at https://github.com/opnsense/core/issues?q=is%3Aissue

**Describe the bug**

Even when sending DNS info via RA is explicitly disabled, radvd is configured with DNSSL. This leads to (at least) TrueNAS 13 dropping its static DNS configuration after each reboot.

**To Reproduce**

1. Manually configure router advertisements on LAN like in screenshot.
2. Check /var/etc/radvd.conf

```
interface igb0 {
	AdvSendAdvert on;
	MinRtrAdvInterval 200;
	MaxRtrAdvInterval 600;
	AdvLinkMTU 1500;
	AdvDefaultPreference medium;
	prefix 2003:a:827:2a00::/64 {
		DeprecatePrefix on;
		AdvOnLink on;
		AdvAutonomous on;
	};
	DNSSL punkt.de {
	};
```

**Expected behavior**

The DNSSL section should not be there.

**Screenshots**
![Bildschirmfoto 2022-08-02 um 09 19 01](https://user-images.githubusercontent.com/27051716/182316595-a5a0b52c-1ab2-4e47-a164-830292bbf17c.png)

**Environment**

OPNsense 22.1.10

pmhausen:
This is what lead to this issue:

https://ixsystems.atlassian.net/browse/NAS-117458

pmhausen:
Thanks for the quick reply. Does the milestone setting imply this is not going into 22.7.x as a bug fix?

fichtner:
Means we can't retrofit 22.7 release but 22.7.x is a possibility :)

maurice-w:
The current behaviour of 'Do not send DNS settings to clients' is consistent with the 'Use the DNS settings of the DHCPv6 server' feature. This only grabs the DNS servers from the DHCPv6 config, not the search list. The placement of these settings in the UI also implies this, although the wording doesn't.

I agree with extending 'Do not send DNS settings to clients' to also apply to DNSSL. But for consistency, 'Use the DNS settings of the DHCPv6 server' should then be modified accordingly. And a UI layout change might be required to reflect the new behaviour (both settings applying to RDNSS and DNSSL).

fichtner:
[...]

PR #5921, opened 2022-08-02
core/radvd - do not advertise DNS domain when DNS advertisments are disabled

pmhausen:
https://github.com/opnsense/core/issues/5919

### DIFF
--- a/src/etc/inc/plugins.inc.d/dhcpd.inc
+++ b/src/etc/inc/plugins.inc.d/dhcpd.inc
@@ -369,6 +369,7 @@ function dhcpd_radvd_configure($verbose = false, $blacklist = array())
             }
         }
 
+        /* do not advertise any DNS servers via RA (RFC 8106) if disabled in UI */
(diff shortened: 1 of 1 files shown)


fichtner:
@pmhausen looks reasonable, maybe @maurice-w can weigh in as well

fichtner:
Actually under the hood this is implemented as radisablerdnss which is correct in that it ignores DNSSL. it probably makes sense to merge this or alternatively make the setting more explicit in the GUI and add one for DNSSL as well. Both work for me, just need to make sure we document this in the code if we opt for this solution here (number 1).

pmhausen:
Comments added, squashed commit.

fichtner:
Merged, thanks!

pmhausen:
Geez ... you are quick. I just did the UI reorganising as suggested by @maurice-w ...

New pull-request?

fichtner:
sure but please wait for my dhcpd.inc change

PR #5922, opened 2022-08-02
system: change UX for new notices

fichtner:
o Switch to shared circle with coloring (no visual jumping)
o Switch SUCCESS circle to muted (muted keeps us calm)
o Switch dialog color to primary (keep one consistent "neutral" layout)
o Switch dialog close button to plain (it doesn't do anything)
o Add translation in legacy template (because that's nice)
o Add muted circle to page load (to avoid shifts and jumps)
o Check error level for crash report in production (probably better as warning)
o Rearrange dismiss and error presentation, inline "circle" indicator (perhaps)

Behavioural changes thought of but discarded:

o ~~Open error dialog right away~~ (it implies action to be taken but might be annoying especially if the reporting page is already open)
o ~~Dismiss on redirect also?~~ (can't do this, clears crash report)

### DIFF
--- a/src/opnsense/mvc/app/controllers/OPNsense/Core/Api/SystemController.php
+++ b/src/opnsense/mvc/app/controllers/OPNsense/Core/Api/SystemController.php
@@ -90,6 +90,64 @@ class SystemController extends ApiControllerBase
                 'status' => $order[$statusCodes[0] ?? 2]
             ];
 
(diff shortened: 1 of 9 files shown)


fichtner:
@AdSchellevis @swhite2 I found a little more time and I think this is my proposal waiting for review :)

PR #5923, opened 2022-08-02
core/radvd Slightly Reorganise UI for disabling of DNS advertisements

pmhausen:
https://github.com/opnsense/core/issues/5919

### DIFF
--- a/src/www/services_router_advertisements.php
+++ b/src/www/services_router_advertisements.php
@@ -420,10 +420,7 @@ include("head.inc");
                       </div>
                       <br />
                       <input id="rasamednsasdhcp6" name="rasamednsasdhcp6" type="checkbox" value="yes" <?=!empty($pconfig['rasamednsasdhcp6']) ? "checked='checked'" : "";?> />
-                      <?= gettext('Use the DNS settings of the DHCPv6 server') ?>
-                      <br />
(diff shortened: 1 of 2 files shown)


src/www/services_router_advertisements.php
```diff
@@ -420,10 +420,7 @@ function addRow() {
                       </div>
                       <br />
                       <input id="rasamednsasdhcp6" name="rasamednsasdhcp6" type="checkbox" value="yes" <?=!empty($pconfig['rasamednsasdhcp6']) ? "checked='checked'" : "";?> />
```
fichtner, line 422: I think we should move both checkboxes into a new row with "DNS options" ?

Their help text is correct after b504bc8 (dns settings includes search list and servers

fichtner:
mind if I take this over?

fichtner:
@pmhausen thanks, if you can please have a brief look at https://github.com/opnsense/core/commit/68894e989dc890954746507fe85cd4f5c087d94a to see if it makes sense (help texts)

Ok, I'll try to place the plugin issue in the appropriate queue.

pmhausen:
I find the wording "If a setting here is used the information below is ignored" a bit clumsy.

How about "If checked no DNS information is sent and the next setting concerning DHCPv6 is ignored."

fichtner:
I tend to agree. The text, however, is trying to explain both settings at the same time. The below configuration is not DHCPv6 -- only the checkbox option above is.

fichtner:
@pmhausen how about 1c322498c90 ? might be a little over the top but now the checkboxes do what the configuration actually does (disable other checkbox, hide custom dns options)

Issue #5477, opened 2022-01-15
Unbound reverse PTR records


lsbbs:
When I create multiple A records for the same IP address like

`name1.example.domain.   A  123.123.123.123`
`name2.example.domain.   A  123.123.123.123`

For each name also a PTR record is created.

`123.123.123.123.in-addr.arpa  PTR   name1.example.domain.`
`123.123.123.123.in-addr.arpa  PTR   name2.example.domain.`

There should be only one reverse record for each IP.
A checkbox if a PTR record should be created would be fine solution for this.

OPNsense 21.7.7-amd64 with unbound 1.13.2

Starkstromkonsument:
@lsbbs Is this a duplicate of #4420 ?

lsbbs:
Yes, somehow.

x390:
Would be nice to see ntopng show the correct address instead of some random alias.

x390:
@fichtner Is this a intended function?

maciekb:
I strongly support that it should be possible to manually indicate whether a PTR record should be generated for a given override record. The problem is quite trivial, but annoying. Many web services rely on reverse name recognition. The first example that comes to mind is Zenarmor (Sensei), which presents inconsistent data in its reports in the current situation. Multiple PTR records for the same IP address cause the name to be retrieved randomly, causing the reports to show several hosts as the source despite the data coming from a single IP address.
In addition, I noticed that Unbound in OPNsense generates just as many IN A records as both PTR records for the local network gateway IP address. This is a problem if we have vlans, for example, and are trying to access the router by name. If the resolver gets an IP address in response from a vlan address that we don't have access to from the source network, then access is obviously not possible.

swhite2:
@maciekb While I agree that manual PTR records as a separate type would be ideal; scope, migration and impact all limit the viability of pushing this in the short term. For now I'm proposing https://github.com/opnsense/core/pull/5925 in order to mitigate the issues discussed here. Let me know what you think.

maciekb:
@swhite2 I understand that a more detailed control may require rebuilding more of the internal logic and adapting the user interface to it. It seems to me that the proposed restriction on the generation of PTR records is a reasonable solution for the moment. 
Thank you very much for understanding the problem and offering a solution.

PR #5925, opened 2022-08-02
unbound: restrict creation of PTR records for both the system domain and host overrides

swhite2:
In order to prevent unpredictable behaviour from something that is not explicitly prohibited in RFC1035, it is best to restrict the creation of PTR records from every single host and alias (except for wildcard entries, no PTR records are created here), to only non-alias overrides (edit: the exception here is an alias whose parent does not create a PTR record). We also further restrict it to unique IP addresses so there can be no confusion in how to maintain the entries within the running Unbound instance.

Hopefully this can pave the way for adding PTR records as a separate type instead of generating them under the hood, as is done currently.

This change should at least address inconsistencies regarding random PTR records being returned as mentioned in https://github.com/opnsense/core/issues/5477

### DIFF
--- a/src/etc/inc/plugins.inc.d/unbound.inc
+++ b/src/etc/inc/plugins.inc.d/unbound.inc
@@ -463,40 +463,35 @@ function unbound_add_host_entries($ifconfig_details = null)
         $interfaces = array_keys(get_configured_interface_with_descr());
     }
 
-    sort($interfaces);
     foreach ($interfaces as $interface) {
         if ($interface == 'lo0' || substr($interface, 0, 4) == 'ovpn') {
             continue;
         }
 
         list ($laddr) = interfaces_primary_address($interface, $ifconfig_details);
-        if (!empty($laddr)) {
-            $domain = $config['system']['domain'];
-            if (isset($config['dhcpd'][$interface]['enable']) && !empty($config['dhcpd'][$interface]['domain'])) {
-                $domain = $config['dhcpd'][$interface]['domain'];
-            }
-            if ($interface === $interfaces[0] && !in_array($laddr, $ptr_records, true)) {
-                $unbound_entries .= "local-data-ptr: \"{$laddr} {$config['system']['hostname']}.{$domain}\"\n";
-                $ptr_records[] = $laddr;
-            }
-            $unbound_entries .= "local-data: \"{$config['system']['hostname']}.{$domain} A {$laddr}\"\n";
-            $unbound_entries .= "local-data: \"{$config['system']['hostname']} A {$laddr}\"\n";
-        }
         list ($laddr6) = interfaces_primary_address6($interface, $ifconfig_details);
-        if (!empty($laddr6)) {
+
+        foreach (['4' => $laddr, '6' => $laddr6] as $ip_version => $addr) {
+            if (empty($addr)) {
+                continue;
+            }
+
             $domain = $config['system']['domain'];
-            if (isset($config['dhcpdv6'][$interface]['enable']) && !empty($config['dhcpdv6'][$interface]['domain'])) {
-                $domain = $config['dhcpdv6'][$interface]['domain'];
+            $dhcpd = $ip_version == '4' ? 'dhcpd' : 'dhcpd6';
+            $record = $ip_version == '4' ? 'A' : 'AAAA';
+            if (isset($config[$dhcpd][$interface]['enable']) && !empty($config[$dhcpd][$interface]['domain'])) {
+                $domain = $config[$dhcpd][$interface]['domain'];
             }
-            if ($interface === $interfaces[0] && !in_array($laddr6, $ptr_records, true)) {
-                $unbound_entries .= "local-data-ptr: \"{$laddr6} {$config['system']['hostname']}.{$domain}\"\n";
-                $ptr_records[] = $laddr6;
+            if ($interface === get_primary_interface_from_list($interfaces) && !in_array($addr, $ptr_records, true)) {
+                $unbound_entries .= "local-data-ptr: \"{$addr} {$config['system']['hostname']}.{$domain}\"\n";
             }
-            $unbound_entries .= "local-data: \"{$config['system']['hostname']}.{$domain} AAAA {$laddr6}\"\n";
-            $unbound_entries .= "local-data: \"{$config['system']['hostname']} AAAA {$laddr6}\"\n";
+            $unbound_entries .= "local-data: \"{$config['system']['hostname']}.{$domain} {$record} {$addr}\"\n";
+            $unbound_entries .= "local-data: \"{$config['system']['hostname']} {$record} {$addr}\"\n";
+
+            $ptr_records[] = $addr;
         }
+
         if (empty($config['unbound']['noreglladdr6'])) {
-            list ($lladdr6) = interfaces_scoped_address6($interface, $ifconfig_details);
             if (!empty($lladdr6)) {
                 /* cannot embed scope */
                 $lladdr6 = explode('%', $lladdr6)[0];
--- a/src/etc/inc/plugins.inc.d/unbound.inc
+++ b/src/etc/inc/plugins.inc.d/unbound.inc
@@ -439,7 +439,7 @@ function unbound_add_host_entries($ifconfig_details = null)
     global $config;
 
     $local_zone_type = 'transparent';
-    $ptr_records = [];
+    $ptr_records = ['127.0.0.1', '::1'];
 
     openlog("unbound", LOG_DAEMON, LOG_LOCAL4);
 
--- a/src/etc/inc/plugins.inc.d/unbound.inc
+++ b/src/etc/inc/plugins.inc.d/unbound.inc
@@ -463,44 +463,46 @@ function unbound_add_host_entries($ifconfig_details = null)
         $interfaces = array_keys(get_configured_interface_with_descr());
     }
 
-    foreach ($interfaces as $interface) {
-        if ($interface == 'lo0' || substr($interface, 0, 4) == 'ovpn') {
-            continue;
-        }
-
-        list ($laddr) = interfaces_primary_address($interface, $ifconfig_details);
-        list ($laddr6) = interfaces_primary_address6($interface, $ifconfig_details);
-
-        foreach (['4' => $laddr, '6' => $laddr6] as $ip_version => $addr) {
-            if (empty($addr)) {
+    if (empty($config['unbound']['noregrecords'])) {
+        foreach ($interfaces as $interface) {
+            if ($interface == 'lo0' || substr($interface, 0, 4) == 'ovpn') {
                 continue;
             }
 
-            $domain = $config['system']['domain'];
-            $dhcpd = $ip_version == '4' ? 'dhcpd' : 'dhcpd6';
-            $record = $ip_version == '4' ? 'A' : 'AAAA';
-            if (isset($config[$dhcpd][$interface]['enable']) && !empty($config[$dhcpd][$interface]['domain'])) {
-                $domain = $config[$dhcpd][$interface]['domain'];
-            }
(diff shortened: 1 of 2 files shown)


fichtner:
fair enough but my main concern was
https://github.com/opnsense/core/blob/2115c45e593caf9f266a9df191c148ac96228525/src/etc/inc/plugins.inc.d/unbound.inc#L473
and
https://github.com/opnsense/core/blob/2115c45e593caf9f266a9df191c148ac96228525/src/etc/inc/plugins.inc.d/unbound.inc#L483

swhite2:
> fair enough but my main concern was
> 
> https://github.com/opnsense/core/blob/2115c45e593caf9f266a9df191c148ac96228525/src/etc/inc/plugins.inc.d/unbound.inc#L473
> 
> and
> 
> https://github.com/opnsense/core/blob/2115c45e593caf9f266a9df191c148ac96228525/src/etc/inc/plugins.inc.d/unbound.inc#L483

Also very relevant, I suppose the same logic makes the most sense here

src/etc/inc/plugins.inc.d/unbound.inc
```diff
@@ -470,7 +473,10 @@ function unbound_add_host_entries($ifconfig_details = null)
             if (isset($config['dhcpd'][$interface]['enable']) && !empty($config['dhcpd'][$interface]['domain'])) {
                 $domain = $config['dhcpd'][$interface]['domain'];
             }
-            $unbound_entries .= "local-data-ptr: \"{$laddr} {$config['system']['hostname']}.{$domain}\"\n";
+            if ($interface === $interfaces[0] && !in_array($laddr, $ptr_records, true)) {
+                $unbound_entries .= "local-data-ptr: \"{$laddr} {$config['system']['hostname']}.{$domain}\"\n";
+                $ptr_records[] = $laddr;
+            }
             $unbound_entries .= "local-data: \"{$config['system']['hostname']}.{$domain} A {$laddr}\"\n";
```
fichtner, line 480: Maybe we should widen the conditional to guard A and AAAA as well?

src/etc/inc/plugins.inc.d/unbound.inc
```diff
@@ -480,7 +486,10 @@ function unbound_add_host_entries($ifconfig_details = null)
             if (isset($config['dhcpdv6'][$interface]['enable']) && !empty($config['dhcpdv6'][$interface]['domain'])) {
                 $domain = $config['dhcpdv6'][$interface]['domain'];
             }
-            $unbound_entries .= "local-data-ptr: \"{$laddr6} {$config['system']['hostname']}.{$domain}\"\n";
+            if ($interface === $interfaces[0] && !in_array($laddr6, $ptr_records, true)) {
```
fichtner, line 489: lo0 or similar may throw this one off

src/etc/inc/plugins.inc.d/unbound.inc
```diff
@@ -560,7 +569,13 @@ function unbound_add_host_entries($ifconfig_details = null)
                                 $unbound_entries .= "local-zone: \"{$alias['domain']}\" redirect\n";
                                 $unbound_entries .= "local-data: \"{$alias['domain']} IN {$host->rr} {$host->server}\"\n";
                             } else {
-                                $unbound_entries .= "local-data-ptr: \"{$host->server} {$alias['hostname']}{$alias['domain']}\"\n";
+                                if (($alias === $tmp_aliases[0] || $tmp_aliases[0]['hostname'] === '*') && !in_array($host->server, $ptr_records, true)) {
```
fichtner, line 572: Perhaps log an error here when overlapping PTR exists?

src/etc/inc/plugins.inc.d/unbound.inc
swhite2: Isn't this caught by https://github.com/opnsense/core/blob/2e42838751cf2c60aab605b00bf315452afe8744/src/etc/inc/plugins.inc.d/unbound.inc#L466-L468?

fichtner: fair point, let me check which interfaces could end up there or if this is already good

src/etc/inc/plugins.inc.d/unbound.inc
swhite2: We should, but I'm not entirely sure if we can rely on the sorted list of interfaces and pick the first one. Shouldn't we at the very least determine if the picked interface is used for e.g. the webgui?

src/etc/inc/plugins.inc.d/unbound.inc
swhite2: Ideally this should be handled by extending the front-end validation, a warning seems more fitting in the grand scheme of things.

fichtner: yeah, the log is cheaper here, warning is ok (still speaking of "log_error" as log error but we should be adhering to severity correctness more)

src/etc/inc/plugins.inc.d/unbound.inc
fichtner: So I've made e76f15c357a4645f for you to make this easier, maybe even move it out of the interfaces loop.
I tend to agree, but the metrics for the web GUI are difficult (what interfaces does the user want to listen on and what sort of NAT is he doing to change it to something else). So what about trying to aim for the same logic as anti-lockout would?  That's a little further but not too complicated.

src/etc/inc/plugins.inc.d/unbound.inc
fichtner: ipsec devices throw this off as well, see other comment and commit

src/etc/inc/plugins.inc.d/unbound.inc
swhite2: @fichtner https://github.com/opnsense/core/pull/5925/commits/9a5d8f851f36876cdd236dd461c17fcef2efbded FYI

src/etc/inc/plugins.inc.d/unbound.inc
swhite2: @fichtner https://github.com/opnsense/core/pull/5925/commits/e5633b81de3bbb7ee8c00ce1b841446b3e18cbb5. What do you think?

src/etc/inc/plugins.inc.d/unbound.inc
```diff
@@ -459,42 +462,47 @@ function unbound_add_host_entries($ifconfig_details = null)
...
+
                 $domain = $config['system']['domain'];
-                if (isset($config['dhcpdv6'][$interface]['enable']) && !empty($config['dhcpdv6'][$interface]['domain'])) {
-                    $domain = $config['dhcpdv6'][$interface]['domain'];
+                $dhcpd = $ip_version == '4' ? 'dhcpd' : 'dhcpd6';
+                $record = $ip_version == '4' ? 'A' : 'AAAA';
+                if (isset($config[$dhcpd][$interface]['enable']) && !empty($config[$dhcpd][$interface]['domain'])) {
+                    $domain = $config[$dhcpd][$interface]['domain'];
+                }
+                if ($interface === get_primary_interface_from_list($interfaces) && !in_array($addr, $ptr_records, true)) {
```
fichtner, line 486: shouldn't this be handled first to avoid a wrong PTR or forcefully clear the offending one?

src/etc/inc/plugins.inc.d/unbound.inc
```diff
@@ -439,6 +439,9 @@ function unbound_add_host_entries($ifconfig_details = null)
     global $config;
 
     $local_zone_type = 'transparent';
+    $ptr_records = [];
```
fichtner, line 442:
```suggestion
    $ptr_records = ['127.0.0.1', '::1'];
```

perhaps? people are strange...

src/etc/inc/plugins.inc.d/unbound.inc
swhite2: https://github.com/opnsense/core/pull/5925/commits/34633e9a894e215f032dd06aff1f8aec5c063b33 ok? :)

fichtner: yep yep

src/etc/inc/plugins.inc.d/unbound.inc
swhite2: good idea, added.

fichtner (approved):

fpehla:
The last commit (https://github.com/opnsense/core/commit/547c8b1c7b4ffe3b79691e9c81a6b523c6d61a67) broke the creation of ptr records for host overrides that are not aliases. No ptr records are generated at all for host entries even if the entry is not an alias and the ip address is not already referenced by another ptr record. 

The new condition in unbound.inc
` if (($alias === $tmp_aliases[0] || $tmp_aliases[0]['hostname'] === '*') && !in_array($host->server, $ptr_records, true))`
evaluates to false for regular host overrides. 

Rewriting the condition like this fixes this issue for me (note the negation '!' inserted):
` if (!($alias === $tmp_aliases[0] || $tmp_aliases[0]['hostname'] === '*') && !in_array($host->server, $ptr_records, true))`

This results in 
* PTR records being created for host overrides
* no PTR records being created for aliases

efahl:
> Rewriting the condition like this fixes this issue for me (note the negation '!' inserted): ` if (!($alias === $tmp_aliases[0] || $tmp_aliases[0]['hostname'] === '*') && !in_array($host->server, $ptr_records, true))`

Tested and works for me, thanks.

guttermonk:
I'm still receiving warnings for "PTR record already exists". I see the fix above, but it looks like an update has been deployed. I'm running Unbound version 1.16.2

Was the fix for this issue merged to master on Aug 16th? Is there a way to tell which version of unbound master is on?

efahl:
@guttermonk Shouldn't matter what version of unbound is running, this is in OPNsense's code that generates the unbound config.  Looking at master, I don't see the fix that @fpehla proposes above.  See https://github.com/opnsense/core/blob/master/src/etc/inc/plugins.inc.d/unbound.inc#L573

guttermonk:
Should this issue be flipped from Merged to Open for visibility?

swhite2:
> I'm still receiving warnings for "PTR record already exists". 

This is completely normal for setups where multiple hostnames/domain names point to the same IP address (like when using an alias), this is a warning for this very reason: it is nonsensical to have multiple PTR records all pointing to the same IP, the system will just return a random host/domain name.

> No ptr records are generated at all for host entries even if the entry is not an alias and the ip address is not already referenced by another ptr record.

@fpehla I don't think the negation is the solution here, it's a solution by chance since the equality comparison `$alias === $tmp_aliases[0]` excludes the host override IF a hostname is specified as the object gets modified when this is the case: 
```
if ($alias['hostname'] != '') {
    $alias['hostname'] .= '.';
}
```

Host overrides which do not have a hostname specified will still generate a PTR record, I can easily reproduce this on my end.

I'll prepare a patch if you wish to test this.

swhite2:
@fpehla `# opnsense-patch 2351f17c3`

ref: https://github.com/opnsense/core/commit/2351f17c33be28cd147a0081fd8129876f6394ae

maciekb:
@swhite2 I I confirm that your patch works, all PTR records are generating correctly. Without the patch, the PTR records for the overwritten hosts did not generate.

swhite2:
> @swhite2 I I confirm that your patch works, all PTR records are generating correctly. Without the patch, the PTR records for the overwritten hosts did not generate.

@maciekb Good to hear. Thanks for the feedback.

@fpehla Thanks for catching this!

fpehla:
Thanks for the patch, this fixes the issue for me. PTR records are generated again for overrides. 

I was not sure if the negation I introduced was the correct fix or if it just solved the issue for me by chance and broke something for others. So thanks again for looking into this!

PR #5927, opened 2022-08-03
Unbound: handle host/domain overrides dynamically

swhite2:
Since DNS blocklists are now done dynamically, it would make sense for other areas of Unbound to do the same in an effort to reduce the amount of restarts.

One such possibility of a restart is the 'hosts' `pluginctl/plugins_configure`, as well as CRUD operations from the WebGUI - which is addressed here.

The dnsbl code has also been refactored to be shorter, as most of this (future) functionality looks more or less the same.

### DIFF
--- a/src/opnsense/service/conf/actions.d/actions_unbound.conf
+++ b/src/opnsense/service/conf/actions.d/actions_unbound.conf
@@ -74,6 +74,14 @@ type:script_output
 message:Updating Unbound DNSBLs
 description:Update Unbound DNSBLs
 
+[overrides]
+command:
+    /usr/local/sbin/pluginctl -c hosts &&
+    /usr/local/opnsense/scripts/unbound/wrapper.py -o
+parameters:
+type:script_output
+message:Updating Unbound overrides
+
 [status]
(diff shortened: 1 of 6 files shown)


src/opnsense/service/conf/actions.d/actions_unbound.conf
```diff
@@ -74,6 +74,14 @@ type:script_output
 message:Updating Unbound DNSBLs
 description:Update Unbound DNSBLs
 
+[overrides]
+command:
+    /usr/local/sbin/pluginctl -c hosts &&
```
fichtner, line 79: 'hosts' also messes with Dnsmasq and we should avoid one way or another

swhite2: would it make sense to split these up on top of the existing `hosts` hook? e.g. `hosts_unbound`. I know it's not particularly clean but I'm missing a simple alternative.

fichtner:
Leave this open for now, maybe we need this later, but Unbound should be a lot faster with blocklist out of the way.

PR #5928, opened 2022-08-03
Fixed a bug in BaseModel.php where raising a meaningful Error did not work

enohka:
In line 153 $field_rfcls could not be resolved because it is declared in the scope of "if (class_exists($classname))"


AdSchellevis:
Doesn't this just swap the logic around? 

```
if (!class_exists($classname)) {
raise ...
}
... do logic
```

looks equal to the previous construction:
```
if (class_exists($classname)) {
 ... do logic
} else {
  raise ..
}
```

Issue #5862, opened 2022-07-09
interfaces: SLAAC WAN improvements


maurice-w:
**Important notices**

Before you add a new report, we ask you kindly to acknowledge the following:

- [x] I have read the contributing guide lines at https://github.com/opnsense/core/blob/master/CONTRIBUTING.md
- [x] I am convinced that my issue is new after having checked both open and closed issues at https://github.com/opnsense/core/issues?q=is%3Aissue

**Describe the bug**

If the WAN IPv6 Configuration Type is set to SLAAC, `/tmp/$if_routerv6` doesn't get created - _unless_ a (stateless) DHCPv6 server exists upstream. Without `/tmp/$if_routerv6`, features like gateway monitoring, gateway groups and policy-based routing don't work.
This might have never worked and is probably not a recently introduced issue.

**To Reproduce**

1. Set up two OPNsense instances. Connect the OPNsense2 WAN interface to the OPNsense1 LAN interface.
2. OPNsense1 LAN interface: IPv6 Configuration Type 'Static IPv6', IPv6 address '2001:db8::1/64', Router Advertisements mode 'Unmanaged', DHCPv6 server disabled
3. OPNsense2 WAN interface: IPv6 Configuration Type 'SLAAC'
4. See error on OPNsense2: WAN interface gets an IPv6 address, but `/tmp/$if_routerv6` doesn't exist and 'System / Gateways / Single / WAN_SLAAC' doesn't show a gateway address.
5. On OPNsense1, change the Router Advertisements mode to 'Stateless' and enable the DHCPv6 server (you don't have to specify an address range).
6. See result on OPNsense2: `/tmp/$if_routerv6` exists and 'System / Gateways / Single / WAN_SLAAC' does show a gateway address.

**Expected behavior**

For SLAAC WAN interfaces, an upstream DHCPv6 server should not be required for proper functionality.

**Describe alternatives you considered**

Configure the gateway address statically, but this isn't always possible.

**Additional context**

SLAAC without DHCPv6 is common on mobile (4G / 5G) connections. I encountered this issue when trying to set up a dual WAN system where WAN2 is tethered to an Android phone.

**Environment**

OPNsense 22.1.10 (amd64, OpenSSL).
Hyper-V Gen2

maurice-w:
[...]

PR #5937, opened 2022-08-07
SLAAC WAN: add support for requesting DNS info via stateless DHCPv6

maurice-w:
There was some dormant code to request DNS servers + search list via stateless DHCPv6, but this was not actually merged into dhcp6c.conf. Also, there was no code to process stateless DHCPv6 replies in the dhcp6c script. Context: [#5862](https://github.com/opnsense/core/issues/5862#issuecomment-1207473812)

### DIFF
--- a/src/etc/inc/interfaces.inc
+++ b/src/etc/inc/interfaces.inc
@@ -2786,7 +2786,11 @@ REQUEST)
         ARGS="\${ARGS} -a \${NAMESERVER}"
     done
     /usr/local/sbin/ifctl -i ${wanif} -6nd \${ARGS}
-    /usr/local/sbin/ifctl -i ${wanif} -6sd \${new_domain_name:+"-a \${new_domain_name}"}
+    ARGS=
+    for DOMAIN in \${new_domain_name}; do
+        ARGS="\${ARGS} -a \${DOMAIN}"
+    done
+    /usr/local/sbin/ifctl -i ${wanif} -6sd \${ARGS}
     /usr/local/sbin/ifctl -i ${wanif} -6pd \${PDINFO:+"-a \${PDINFO}"}
     /usr/local/sbin/configctl -d interface newipv6 {$wanif}
     ;;
@@ -2797,7 +2801,11 @@ INFOREQ)
         ARGS="\${ARGS} -a \${NAMESERVER}"
     done
     /usr/local/sbin/ifctl -i ${wanif} -6nd \${ARGS}
-    /usr/local/sbin/ifctl -i ${wanif} -6sd \${new_domain_name:+"-a \${new_domain_name}"}
+    ARGS=
+    for DOMAIN in \${new_domain_name}; do
+        ARGS="\${ARGS} -a \${DOMAIN}"
+    done
+    /usr/local/sbin/ifctl -i ${wanif} -6sd \${ARGS}
     /usr/local/sbin/configctl -d interface newipv6 {$wanif}
     ;;
 EXIT|RELEASE)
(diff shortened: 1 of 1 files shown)


src/etc/inc/interfaces.inc
```diff
@@ -2786,10 +2786,28 @@ REQUEST)
...
+    for NAMESERVER in \${new_domain_name_servers}; do
+        ARGS="\${ARGS} -a \${NAMESERVER}"
+    done
+    /usr/local/sbin/ifctl -i ${wanif} -6nd \${ARGS}
+    ARGS=
+    for DOMAIN in \${new_domain_name}; do
+        ARGS="\${ARGS} -a \${DOMAIN}"
+    done
+    /usr/local/sbin/ifctl -i ${wanif} -6sd \${ARGS}
+    /usr/local/sbin/configctl -d interface newipv6 {$wanif}
```
fichtner, line 2809:
```suggestion
    /usr/local/sbin/configctl -d dns reload
```

only reload dns here?

src/etc/inc/interfaces.inc
```diff
@@ -2786,10 +2786,28 @@ REQUEST)
...
-    /usr/local/sbin/ifctl -i ${wanif} -6sd \${new_domain_name:+"-a \${new_domain_name}"}
+    ARGS=
+    for DOMAIN in \${new_domain_name}; do
+        ARGS="\${ARGS} -a \${DOMAIN}"
+    done
+    /usr/local/sbin/ifctl -i ${wanif} -6sd \${ARGS}
     /usr/local/sbin/ifctl -i ${wanif} -6pd \${PDINFO:+"-a \${PDINFO}"}
     /usr/local/sbin/configctl -d interface newipv6 {$wanif}
     ;;
+INFOREQ)
```
fichtner, line 2797: Nice. As far as I can see this was never once implemented and also absent from pfSense.

src/etc/inc/interfaces.inc
```diff
@@ -2828,7 +2846,7 @@ EOF;
 
     /* merge configs and prepare single instance of dhcp6c for startup */
     foreach (legacy_config_get_interfaces(['enable' => true, 'virtual' => false]) as $_interface => $_wancfg) {
-        if (empty($_wancfg['ipaddrv6']) || $_wancfg['ipaddrv6'] != 'dhcp6') {
+        if (empty($_wancfg['ipaddrv6']) || ($_wancfg['ipaddrv6'] != 'dhcp6' && $_wancfg['ipaddrv6'] != 'slaac')) {
```
fichtner, line 2849: nice catch

src/etc/inc/interfaces.inc
maurice-w: Not sure to be honest. For SLAAC interfaces, we need to invoke rc.newwanip(v6) after an address was autoconfigured, correct? Is this already handled somewhere else?

fichtner: Oh, ok, let me check how often this is being called.

src/etc/inc/interfaces.inc
maurice-w: Checking $REASON was introduced in https://github.com/opnsense/core/commit/d36f0f4f62557eb7caa150c7bd1611c08667990f. Not sure if it might have worked before.

fichtner: INFOREQ seems to be emitted when "information-only" is used so that REQUEST isn't called so that seems to be correct here indeed.

PR #5943, opened 2022-08-11
CRL management small adjustments

kulikov-a:
Hi!
ref. https://forum.opnsense.org/index.php?topic=29567.0
it turned out that phpseclib2.0 does not support some algos and does not throw exceptions on errors (hope that with phpseclib3.0 it will be possible to do everything more correctly).
https://github.com/phpseclib/phpseclib/blob/master/CHANGELOG.md#300---2020-12-16
https://github.com/phpseclib/phpseclib/blob/master/CHANGELOG.md#303---2021-01-16

PR:
add some debug info
dont add cert to crl if cert_revoke() / crl_update() failed
check if CA can sign anything before CRL create
set "method" param to set Method select value

Thanks!

### DIFF
--- a/src/etc/inc/certs.inc
+++ b/src/etc/inc/certs.inc
@@ -612,6 +612,10 @@ function crl_update(&$crl)
     $crl['serial']++;
     $ca_str_crt = base64_decode($ca['crt']);
     $ca_str_key = base64_decode($ca['prv']);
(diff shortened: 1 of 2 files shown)


swhite2:
This looks good, and yes this section is going to need a rework as soon as phpseclib 3.0 is included, as it's API has largely changed and supposedly makes things easier.

kulikov-a:
Thanks!!

PR #5945, opened 2022-08-11
Unbound / DNS64: add AAAA-only mode

### LINKED ISSUE #5944: Feature: Unbound: AAAA-only mode
**Important notices**

Some poorly implemented devices (IoT) do have IPv6 support, but prefer IPv4 whenever possible. These devices do perform DNS queries for A + AAAA records. If both exist, they should use IPv6, but actually use IPv4. Some devices don't even "fall back" to IPv6 in case IPv4 connections fail. The solution to this is only sending AAAA records in DNS responses, which successfully forces these devices to use IPv6.

I would like an option in the Unbound settings which removes all A records from responses. Since this only makes sense if all host names have AAAA records, this option should only apply if DNS64 is enabled.

**Additional context**

This is also useful for testing which devices / applications brake if there are no A records in DNS at all.

maurice-w:
Adds on option which removes all A records from the answer section of all responses. Closes #5944.

### DIFF
--- a/src/etc/inc/plugins.inc.d/unbound.inc
+++ b/src/etc/inc/plugins.inc.d/unbound.inc
@@ -119,9 +119,15 @@ function unbound_generate_config()
 
     $optimization = unbound_optimization();
 
+    $respip = '';
     $dns64prefix = '';
     if (isset($config['unbound']['dns64'])) {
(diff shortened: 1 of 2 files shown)


fichtner:
Merged, thanks!

PR #5965, opened 2022-08-20
CRL: decode stored crl data before display

kulikov-a:
Hi!
ref. https://forum.opnsense.org/index.php?topic=29937.0
Turned out that administrator may accidentally break the **imported** CRL by clicking Save when viewing crl data: `CRL data` field shows encoded string.
Thanks!

### DIFF
--- a/src/www/system_crlmanager.php
+++ b/src/www/system_crlmanager.php
@@ -453,7 +453,7 @@ include("head.inc");
               <tr>
                 <td><a id="help_for_crltext" href="#" class="showhelp"><i class="fa fa-info-circle"></i></a> <?=gettext("CRL data");?></td>
                 <td>
-                  <textarea name="crltext" id="crltext" cols="65" rows="7"><?=$thiscrl['text'];?></textarea>
+                  <textarea name="crltext" id="crltext" cols="65" rows="7"><?=base64_decode($thiscrl['text']);?></textarea>
                   <div class="hidden" data-for="help_for_crltext">
                     <?=gettext("Paste a Certificate Revocation List in X.509 CRL format here.");?>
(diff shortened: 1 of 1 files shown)


fichtner:
This is all a bit strange... if we save base64 encoded data and expect users to paste base64 encoded data (they certainly cannot paste binary data into a textarea very well) where is the need for any sort of decode/encode internally?

kulikov-a:
@fichtner Hi! hm..if we encode user input for imorted/existing crl at
https://github.com/opnsense/core/blob/f782ccb8c4a16c8b194f8114822beb77132718e0/src/www/system_crlmanager.php#L219-L221
i think we need to decode it before display (or we will double encode it later on save? ;)

swhite2:
> This is all a bit strange... if we save base64 encoded data and expect users to paste base64 encoded data (they certainly cannot paste binary data into a textarea very well) where is the need for any sort of decode/encode internally?

Do we expect base64 encoded data though?

kulikov-a:
since all other certs data is encoded in conf (keys, internal crl's etc) i found it quiet logical )

fichtner:
I don't think we double-encode all PEM related input. It is what it is (and there could be more bugs).

kulikov-a:
yep. and, for example, on the `system_camanager.php` page, config certificates are pre-decoded:
https://github.com/opnsense/core/blob/05922760bd05dca42a34fae7094eb798ba14bca5/src/www/system_camanager.php#L168-L179

Or did I miss the thread somewhere?

swhite2:
@kulikov-a Thanks!

kulikov-a:
Thanks!)

Issue #5974, opened 2022-08-22
Interfaces Overview: DHCPv6 Reload / Release / Renew buttons missing


maurice-w:
**Important notices**

Before you add a new report, we ask you kindly to acknowledge the following:

- [x] I have read the contributing guide lines at https://github.com/opnsense/core/blob/master/CONTRIBUTING.md
- [x] I am convinced that my issue is new after having checked both open and closed issues at https://github.com/opnsense/core/issues?q=is%3Aissue

**Describe the bug**

When the IPv6 configuration type of an interface is set to DHCPv6, but the IPv4 configuration type is _not_ set to DHCP, no 'DHCP Reload / Release / Renew' buttons are shown in 'Interfaces: Overview'.

**To Reproduce**

1. Go to 'Interfaces: WAN', set 'IPv4 Configuration Type' to DHCP and 'IPv6 Configuration Type' to DHCPv6, save & apply.
2. Go to 'Interfaces: Overview', click on 'DHCP Release'.
3. See result: Both IPv4 and IPv6 addresses get released.
4. Go to 'Interfaces: WAN', set 'IPv4 Configuration Type' to 'Static IPv4' or 'None', save & apply.
5. Go to 'Interfaces: Overview', see error: 'DHCP Reload / Release / Renew' buttons are missing.

**Expected behavior**

DHCPv6 reload / release / renew feature should be available for interfaces without DHCPv4.

**Additional context**

There is code in `status_interfaces.php` to display dedicated buttons for DHCPv6:
https://github.com/opnsense/core/blob/f609a9cbb1a814023fc4d019a8abc65955a02d8e/src/www/status_interfaces.php#L141-L153
But there is no code in `get_interfaces_info()` which would actually return 'dhcp6link' information.

**Environment**

OPNsense 23.1.a_102 (amd64, OpenSSL)
Hyper-V Gen2 Ver10.0

fichtner:
Caveat as per commit message 0592276. We could move the thing "up" to the status and just show a link type thing (dhclient and dhcp6c up indicator). SLAAC also missing. Maybe you want to give it a try.

maurice-w:
Thanks @fichtner, mostly works. Except for the DHCPv6 "Release" button not switching to "Renew" after a release. We don't actually kill dhcp6c, so get_interfaces_info() always returns dhcp6link = up. Might have to find another way to determine DHCPv6 state.

[...]

PR #5990, opened 2022-08-24
Interfaces Overview, merge DHCPv4 / DHCPv6 buttons

maurice-w:
Having two sets of buttons implies that they only act on the respective protocol, but this is not the case.

https://github.com/opnsense/core/issues/5974#issuecomment-1225222923


fichtner:
Merged, thanks!

PR #5992, opened 2022-08-24
DHCPv6: add / delete null route for delegated prefix

### LINKED ISSUE #3304: No null route is created for delegated prefix
When a DHCPv6 WAN interface gets delegated a prefix, a null route should automatically be created for this prefix. This is best practice to prevent routing loops and is implemented in most routers.

Example:
ISP delegates prefix 2001:db8:1:a00::/56. OPNsense uses prefix ID 1 (2001:db8:1:a01::/64) for its tracking LAN interface. A host in this LAN sends packets to an address in any other (unused) subnet of the delegated prefix (e.g. 2001:db8:1:a02::/64). OPNsense routes these packets to the ISP who routes them back to OPNsense which routes them to the ISP... and so on. This can easily be reproduced using traceroute.

If you have a static prefix you can manually add a null route, but this is not possible with dynamic prefixes.

Just for reference, the RFCs which actually mandate this:

[RFC 7084 (Basic Requirements for IPv6 Customer Edge Routers)](https://tools.ietf.org/html/rfc7084):
> Any packet received by the CE router with a destination address in the prefix(es) delegated to the CE router but not in the set of prefixes assigned by the CE router to the LAN **must** be dropped.  In other words, the next hop for the prefix(es) delegated to the CE router should be the null destination.  This is necessary to prevent forwarding loops when some addresses covered by the aggregate are not reachable [RFC4632].

[RFC 4632 (Classless Inter-domain Routing (CIDR))](https://tools.ietf.org/html/rfc4632):
> A router that generates an aggregate route for multiple, more-specific routes **must** discard packets that match the aggregate route, but not any of the more-specific routes. In other words, the "next hop" for the aggregate route should be the null destination.  This is necessary to prevent forwarding loops when some addresses covered by the aggregate are not reachable.

maurice-w:
Null routes should be added for delegated prefixes to prevent possible routing loops.

Closes #3304 (finally).

### DIFF
--- a/src/etc/inc/interfaces.inc
+++ b/src/etc/inc/interfaces.inc
@@ -916,6 +916,12 @@ function interface_bring_down($interface = 'wan', $ifacecfg = false)
         }
     }
 
+    /* delete null route for delegated prefix */
+    $oldpd = trim(shell_exec(exec_safe('/usr/local/sbin/ifctl -i %s -6p', $realifv6)));
+    if (!empty($oldpd)) {
+        mwexecf('/sbin/route -6 delete %s', $oldpd);
+    }
+
     /* clear stale state associated with this interface */
     mwexecf('/usr/local/sbin/ifctl -4c -i %s', $realif);
     mwexecf('/usr/local/sbin/ifctl -6c -i %s', $realifv6);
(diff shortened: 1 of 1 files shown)


fichtner:
Aha! How about we add this to ifctl instead when adding a prefix and clearing the interface address family data?

maurice-w:
Interesting idea... Sure, why not! Are you up for this? I don't really dare to touch ifctl. ;-)

PR #5993, opened 2022-08-26
webgui: match cipher suites and openssl config commands

kulikov-a:
Hi!
ref. https://forum.opnsense.org/index.php?topic=29284
ref. https://forum.opnsense.org/index.php?topic=30048.0
It turned out that the administrator can break access to the gui by selecting  TLSv1.3 ciphersuites only.
OpenSSL uses different config file options and commands to define ciphersuites for TLS1.2 and below (`CipherString`) and for TLS1.3 suites (`Ciphersuites`): 
https://www.openssl.org/docs/manmaster/man3/SSL_CONF_cmd.html#SUPPORTED-CONFIGURATION-FILE-COMMANDS
PR tries to follow this logic and use matching command for defined ciphers.
RFC8446 Compliance Requirements (https://www.rfc-editor.org/rfc/rfc8446#section-9.1) is also fulfilled when using TLS1.3 suites: the presence of the TLS_AES_128_GCM_SHA256 is controlled.
Thanks!

UPD. RFC8446 Compliance Requirements check implemented in https://github.com/opnsense/core/commit/d31bff3c3975fe8c043538bcccbe30674068eae8 and removed from pr

### DIFF
--- a/src/etc/inc/plugins.inc.d/webgui.inc
+++ b/src/etc/inc/plugins.inc.d/webgui.inc
@@ -411,14 +411,21 @@ ssl.openssl.ssl-conf-cmd = (
 
 EOD;
         } else {
-            $lighty_config .= <<<EOD
-
-ssl.openssl.ssl-conf-cmd = (
-    "MinProtocol" => "TLSv1",
-    "CipherString" => "{$config['system']['webgui']['ssl-ciphers']}"
-)
-
-EOD;
+            // use the same supported ciphers source as system_advanced_admin.php page do (its not a full list. but its openssl defaults)
+            $sys_ciphers = json_decode(configd_run("system ssl ciphers"), true);
+            $tls13_suites = array_keys(array_filter($sys_ciphers, function($val) { return $val['version'] == "TLSv1.3"; }));
+            $suites_selected = explode(":", $config['system']['webgui']['ssl-ciphers']);
+            $tls_suites_selected = array_diff($suites_selected, $tls13_suites);
+            $tls13_suites_selected = array_intersect($tls13_suites,$suites_selected);
+            $lighty_config .= "ssl.openssl.ssl-conf-cmd = (\n";
+            $lighty_config .= "    \"MinProtocol\" => \"TLSv1\"";
+            if ($tls13_suites_selected) {
+                $lighty_config .= ",\n    \"Ciphersuites\" => \"" . implode(":", $tls13_suites_selected) . "\"";
+            }
+            if ($tls_suites_selected) {
+                $lighty_config .= ",\n    \"CipherString\" => \"" . implode(":", $tls_suites_selected) . "\"";
+            }
+            $lighty_config .= "\n)\n";
         }
 
         if (!empty($config['system']['webgui']['ssl-hsts'])) {


src/etc/inc/plugins.inc.d/webgui.inc
```diff
@@ -450,6 +465,9 @@ server.error-handler-404   = "/ui/404"
 ## to help the rc.scripts
 server.pid-file            = "/var/run/lighty-webConfigurator.pid"
 
+## virtual directory listings
+server.dir-listing         = "disable"
```
fichtner, line 469: This one is gone, no longer supported.

kulikov-a: yes, sorry. fixed already. copy/paste (

src/etc/inc/plugins.inc.d/webgui.inc
```diff
@@ -411,14 +411,29 @@ ssl.openssl.ssl-conf-cmd = (
...
         } else {
-            $lighty_config .= <<<EOD
-
-ssl.openssl.ssl-conf-cmd = (
-    "MinProtocol" => "TLSv1",
-    "CipherString" => "{$config['system']['webgui']['ssl-ciphers']}"
-)
-
-EOD;
+            $tls13_suites = array("TLS_AES_128_GCM_SHA256", "TLS_AES_256_GCM_SHA384", "TLS_CHACHA20_POLY1305_SHA256", "TLS_AES_128_CCM_SHA256", "TLS_AES_128_CCM_8_SHA256");
```
fichtner, line 414: Personally hardcoding RFCs here is what we wanted to avoid with these advanced selectors. I’m not a fan of the setting as a whole, but it also seems we can’t do much else here other than fix it like this.

fichtner:
Assigning @AdSchellevis since he integrated this and likely has an opinion on integration.

src/etc/inc/plugins.inc.d/webgui.inc
kulikov-a: agree. and i thought about ` openssl ciphers -s -tls1_3 `. but it returns only 3 of 5 suits.
although with a direct request it gives all 5
`openssl ciphers -tls1_3 -v -s -ciphersuites TLS_AES_256_GCM_SHA384:TLS_CHACHA20_POLY1305_SHA256:TLS_AES_128_GCM_SHA256:TLS_AES_128_CCM_8_SHA256:TLS_AES_128_CCM_SHA256`

fichtner: That’s slightly better, hopefully LibreSSL doesn’t screw this up. Still have to deal with this as well until 23.1.

kulikov-a: actualy, system_advanced_admin.php doesn't show these two suites (`TLS_AES_128_CCM...`) either. so we realy can try to use `configd_run("system ssl ciphers")` (like system_advanced_admin.php do)  instead of static array (will this call work on system start)?

kulikov-a:
sorry, is 
https://github.com/kulikov-a/core/commit/41043fce8d4deb0dd0f59a8f2041c925fbd78b66
what we talking about?

AdSchellevis:
@kulikov-a When it comes to ensuring the correct contents, I would suggest we move that to the "model" (knowing we don't have one here). As the model is responsible for proper data. This https://github.com/opnsense/core/commit/d31bff3c3975fe8c043538bcccbe30674068eae8 would improve the validation. This just leaves us with writing the correct config.

kulikov-a:
@AdSchellevis true! removed this check

kulikov-a:
updated (no static array) and made shorter

AdSchellevis:
@kulikov-a looks good, thanks! it would have been nice if we could easily determine if a cipher belongs to 1.3 so we shouldn't have to ask openssl, unfortunately that doesn't appear to be the case (unless injecting the list statically).

kulikov-a:
@AdSchellevis  Thanks!! (agree, openssl is rather obscure in some places)

PR #6012, opened 2022-09-07
system: give configd a chance to load

kulikov-a:
Hi!
ref. https://forum.opnsense.org/index.php?topic=30197.0
Turned out that the templates may not be reloaded at system start.
(returns "configd socket missing..")
Perhaps the call to the service is going too fast and it has not yet been loaded?
May be we can give a second chance? )
This works and with `w 1`.  `w 3` just in case

Thanks!

### DIFF
--- a/src/etc/rc.syshook.d/early/15-templates
+++ b/src/etc/rc.syshook.d/early/15-templates
@@ -1,4 +1,4 @@
 #!/bin/sh
 
 echo -n "Generating configuration: "
-configctl template reload \*
+configctl -w 3 template reload \*


fichtner:
@kulikov-a after discussing with @AdSchellevis we went the other way with https://github.com/opnsense/core/commit/e2c182bd4b51ba59e6f6c1631b97238052a618e0 as this could also be an issue during +POST_INSTALL phase in both core and plugins and we don't want to miss these, even if just happening for a fraction of users (slow box, too much load, who knows).

PR #6014, opened 2022-09-10
update pattern to zst for the Arch packages

gacekjk:
Arch Linux moved from `xz` compression to `zst` some time ago: [Now using Zstandard instead of xz for package compression](https://archlinux.org/news/now-using-zstandard-instead-of-xz-for-package-compression/).

The change simply replaces the pattern for the `xz` compressed package names for Arch Linux with the `zst` in the `squid.conf`.


AdSchellevis:
@gacekjk thanks!

Issue #4185, opened 2020-06-26
IPs in ARP table not sorted numerically


xorbital:
**Describe the bug**
The ARP table sorts the IPs in this order:
- 192.168.1.1
- 192.168.1.11
- 192.168.1.2

**To Reproduce**
Go to Interfaces - Diagnostics - ARP table
Click on "IP" to sort via IP.

**Expected behavior**
The IPs should be sorted in numeric ascending / descending order

**Screenshots**
![Screenshot_2020-06-26 ARP Tabelle Diagnose Schnittstellen OPNsense localdomain](https://user-images.githubusercontent.com/53773184/85863441-70f79180-b7c3-11ea-89f8-9c5e572294fd.png)

**Relevant log files**
(not applicable)

**Additional context**
It seems like the IPs are interpreted as a string while there should probably be some number sorting going on.

**Environment**
OPNsense 20.1.7-amd64

Thank you very much for your work on this project, I really appreciate it :heart:

pakair:
arp -n | sort -V produces ip-sorted output.

PR #6016, opened 2022-09-12
ARP Table: hostname uses reverse DNS lookup as fallback

soif:
Arp table only relies on dhcp-leases entries to get the hostname. So when using an external DHCP server, or for IP addresses that are not assigned by DHCP, the displayed hostnames are always blank.

Features added:
- if hostame is not found in the dhcp-leases, the hostname is fetched from a DNS PTR query, (using the system's DNS server).
- use_dns variable can allow/disallow this behaviour (useful is someone would like to move this to a system preference)

HTH

### DIFF
--- a/src/opnsense/mvc/app/controllers/OPNsense/Diagnostics/Api/InterfaceController.php
+++ b/src/opnsense/mvc/app/controllers/OPNsense/Diagnostics/Api/InterfaceController.php
@@ -70,7 +70,12 @@ class InterfaceController extends ApiControllerBase
     public function getArpAction()
     {
         $backend = new Backend();
-        $response = $backend->configdRun('interface list arp json');
+        if (strtolower($this->request->get('resolve')) == 'yes') {
+            $response = $backend->configdRun('interface list arp -r json');
+        } else {
+            $response = $backend->configdRun('interface list arp json');
+        }
+
         $arptable = json_decode($response, true);
 
         $intfmap = $this->getInterfaceNames();
--- a/src/opnsense/mvc/app/views/OPNsense/Diagnostics/arp.volt
+++ b/src/opnsense/mvc/app/views/OPNsense/Diagnostics/arp.volt
@@ -41,7 +41,12 @@ POSSIBILITY OF SUCH DAMAGE.
                     multiSelect: false
                 });
             }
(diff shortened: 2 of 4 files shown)


AdSchellevis:
A similar construction as used for the routes might be acceptable, putting the burden to resolve (with timeouts for everyone) isn't, so if you're up for it, please change the PR to include the option to the endpoint, either as uri part or as parameter (latter might make more sense at the moment).

For some inspiration, the following code blocks are being used in the route dialog.

https://github.com/opnsense/core/blob/745d46c4eeefb33566f5fe25024d6235846660fd/src/opnsense/mvc/app/views/OPNsense/Diagnostics/routes.volt#L62

https://github.com/opnsense/core/blob/745d46c4eeefb33566f5fe25024d6235846660fd/src/opnsense/mvc/app/controllers/OPNsense/Diagnostics/Api/InterfaceController.php#L137-L142

https://github.com/opnsense/core/blob/745d46c4eeefb33566f5fe25024d6235846660fd/src/opnsense/scripts/routes/show_routes.py#L42

ps:  removing the `-n` from arp would resolve in the same pass.

soif:
Hi thanks for the fast review!

You're perfectly right. I wrongly assumed that the ARP table should typically be only filled with LAN IP addresses where DNS timeout is usually not an issue for in-addr.arpa zones. But for cases where OPNsense is used mainly with  publically routables ip addresses, that obviously would create timeout issues.

**Thanks** for the inspiration examples (helps a LOT):  I'm gonna fix the PR to make DNS resolving optionnal, like in routes.

soif:
Here it is. 

Notice that:

* i used the -r as additional parameter, that triggers the Reverse DNS. 
Without -r the commands behaves exactly as in the latest release. 

* I've added a natural sort of IP addresses (separate commit, to be easily reverted), as requested in #4185.
This is visible when using the command line:
`configctl "interface list arp -r"`
When using the GUI, and because Javascript sort the results, it might still be usefull, when for example sorting by Interface or Interface name.

HTH

AdSchellevis:
@soif thanks, I've made some small cleanups in https://github.com/opnsense/core/commit/1548ae0eb79bde1f059f58c0a61658b5e33a0c9f before merging. The natural sort has been left out for now as it will likely lead to new tickets since it doesn't fix the complete issue.

Issue #5912, opened 2022-07-29
Internal CA's created in OPNSense have no KeyUsage Extension


goodomens42:
- [x ] I have read the contributing guide lines at https://github.com/opnsense/core/blob/master/CONTRIBUTING.md
- [x ] I am convinced that my issue is new after having checked both open and closed issues at https://github.com/opnsense/core/issues?q=is%3Aissue

**Description**

When creating a new CA in "System -> Trust -> Authorities" using "Create an internal Certificate Authority", the resulting CA certificate has no KeyUsage Extension. It should have one of type "critical" with values "Certificate Sign, CRL Sign" as otherwise third party software might not accept the CA as valid.

**To Reproduce**

1. Go to "System -> Trust -> Authorities"
2. Click "Create an internal Certificate Authority" and fill in the form
3. Export the generated CA certificate
4. Examine it with a suitable certificate viewer

**Expected behavior**

See Description

**Environment**

OPNsense 22.1.10-amd64
FreeBSD 13.0-STABLE
OpenSSL 1.1.1q 5 Jul 2022

kulikov-a:
Hi
could you please give an example when such a cert is not considered valid? I thought that this extension is optional (although it is marked critical if present) and is intended to limit the scope of the key (if there is no need to limit the scope, then the extension may be absent)

kulikov-a:
@fichtner @AdSchellevis @goodomens42 @barzog
I must apologize if my previous post misled anyone, after taking the time to read the slightly ambiguous text of the [rfc5280](https://www.rfc-editor.org/rfc/rfc5280.html#section-4.2.1.3) again and digging around a bit, I agree with the users that the text of the standard is rightly treated  keyUsage Extension as necessary for (root) CA's that used to validate other certs or crls
`
   Conforming CAs MUST include this extension in certificates that
   contain public keys that are used to validate digital signatures on
   other public key certificates or CRLs.  When present, conforming CAs
   SHOULD mark this extension as critical
`
a possible fix is mentioned at https://github.com/opnsense/core/issues/5629

sorry!

some links
https://www.truenas.com/community/threads/openvpn-ca-from-opnsense-root-ca-must-have-keyusage-extension-set.102686/
https://kb.vmware.com/s/article/74756
https://www.ibm.com/docs/en/external-auth-server/2.4.2?topic=extensions-keyusage-extension

PR #6017, opened 2022-09-12
OpenSSL: add keyUsage extension in CA config

kulikov-a:
Hi!
ref. and hopefully closes https://github.com/opnsense/core/issues/5912
Thanks!

### DIFF
--- a/src/etc/ssl/opnsense.cnf
+++ b/src/etc/ssl/opnsense.cnf
@@ -211,6 +211,10 @@ authorityKeyIdentifier=keyid:always,issuer
 #basicConstraints = critical,CA:true
 # So we do this instead.
 basicConstraints = CA:true
+# keyUsage (rfc5280): Conforming CAs MUST include this extension in certificates that contain public keys
+# that are used to validate digital signatures on other public key certificates or CRLs.
+# When present, conforming CAs SHOULD mark this extension as critical.
+keyUsage = critical, cRLSign, digitalSignature, keyCertSign
 
 ###OPNsense:subjectAltName###
 


AdSchellevis:
@kulikov-a thanks, I think this would be a good addition indeed, since this only affects new CA's I don't think this has a very high impact, as this seems to be the advised setting anyway. Looking at https://openssl-ca.readthedocs.io/en/latest/intermediate-configuration-file.html and some other references, apparently the basicConstraints should append critical too (`basicConstraints = critical, CA:true`), what's your take on that?

kulikov-a:
@AdSchellevis Thanks!
yes, reading (https://datatracker.ietf.org/doc/html/rfc5280#section-4.2.1.9):
`
  Conforming CAs MUST include this extension in all CA certificates
   that contain public keys used to validate digital signatures on
   certificates and MUST mark the extension as critical in such
   certificates.
`
Im pretty sure that `basicConstraints = critical, CA:true` is the right choice. but there is a comment:
`# This is what PKIX recommends but some broken software chokes on critical
# extensions.` 
so i decided it was intentional and well reasoned

AdSchellevis:
@kulikov-a I read the same note as well, no clue if this is still relevant to be honest, but let's keep it as is for now until someone opens a ticket.

AdSchellevis:
@kulikov-a I'm not surprised, if we are going to change defaults anyway, we might as well change this one while there.

kulikov-a:
@AdSchellevis yes, i would change it now if we try to follow the rfc in other places anyway and there are no reasoned objections. should i make a pr for this?

AdSchellevis:
@kulikov-a yes please :)

PR #6019, opened 2022-09-13
Suggestion to generalize the use of "set"

antipiot:
Every option is formatted as "if set - option do" but this one, which may be uncomfortable. I'd suggest to change this, either by changing the title as pulled or by changing the effect of the option description if it's not clear enough.


fichtner:
The page is going away mid-term and the change is incomplete. Since we also change documentation and translations we might as well skip this.

Issue #6008, opened 2022-09-06
states.py - problems with long running sessions


mlht:
- [x ] I have read the contributing guide lines at https://github.com/opnsense/core/blob/master/CONTRIBUTING.md
- [ x] I am convinced that my issue is new after having checked both open and closed issues at https://github.com/opnsense/core/issues?q=is%3Aissue

**Describe the bug**
My WebGUI (Firewall:Diagnostics:Sessions) shows today "No results found!", which can't be true.
pftop changes the output format on **long running sessions** (even if you set the terminal width to 1000) for Age and Bytes at some point.

![pftop](https://user-images.githubusercontent.com/35572091/188737864-41c2d7b8-31da-4519-a7a3-283dba8be7a1.PNG)

The screenshot shows:
1. The age value changed the format from xx:xx:xx to an integer
2. The bytes value changed from bytes to kilobytes

/usr/local/opnsense/scripts/filter/lib/states.py
The first problem (age) throws an exception at line 206

           for timefield in ['age', 'expire']:
                tmp = record[timefield].split(':')
                record[timefield] = int(tmp[0]) * 3600 + int(tmp[1]) * 60 + int(tmp[2]) if len(tmp) > 2 else 0

The second problem (line 197) returns 0 if the bytes value ends with K
`record['bytes'] = int(parts[idx+4]) if parts[idx+4].isdigit() else 0`

A short test with this line returns the right value
`record['bytes'] = int(parts[idx+4][0:-1])*1024 if (parts[idx+4] == "K") else int(parts[idx+4])`

OPNsense 22.1.10 (amd64, OpenSSL).
FreeBSD 13.0-STABLE

kulikov-a:
Hi
didn't check it and it's probably not quite technically correct, but since the format of the specified fields seems to depend on the value of the `-w` argument ([you can check it in the source code](https://github.com/araujobsd/pftop/blob/ab98de9ce24f0b15da5ab677acb50db187ad1094/engine.c)), have you tried to significantly increase it in the `states.py`? (otherwise, to parse large values in these fields, it's necessary to mimic the pftop's conversion logic?)

mlht:
At line 166 is the call for pftop in states.py:
sp = subprocess.run(['/usr/local/sbin/pftop', '-w', '1000', '-b','-v', 'long','9999999999999'], capture_output=True, text=True)

-w is set to 1000

[...]

PR #6020, opened 2022-09-13
Firewall/Diagnostics/Sessions: parse pftop internal data conversion

kulikov-a:
Hi!
ref. https://github.com/opnsense/core/issues/6008
Turned out that for sessions with large values of Age or Bytes, the `pftop` utility can change the display format of this fields even with a large value of the `-w` parameter. So internal Age value conversion can break  Firewall/Diagnostics/Sessions page display

[print_fld_sdiv(field_def *fld, u_int64_t size, int div)](https://github.com/araujobsd/pftop/blob/ab98de9ce24f0b15da5ab677acb50db187ad1094/engine.c#L661)
[print_fld_age(field_def *fld, unsigned int age)](https://github.com/araujobsd/pftop/blob/ab98de9ce24f0b15da5ab677acb50db187ad1094/engine.c#L607)

PR tries to take this into account and returns the values to a single format.
The value of the multiplier for converting units on the frontend has also been changed to align with the `pftop` logic.
Thanks!

ps. not so happy with the if-elif-elif-else expression for age/expire fields, but at least tried to make it according to the probability of a match

### DIFF
--- a/src/opnsense/mvc/app/views/OPNsense/Diagnostics/fw_pftop.volt
+++ b/src/opnsense/mvc/app/views/OPNsense/Diagnostics/fw_pftop.volt
@@ -64,9 +64,9 @@
                             bytes: function(column, row) {
                                 if (!isNaN(row[column.id]) && row[column.id] > 0) {
                                     let fileSizeTypes = ["", "K", "M", "G", "T", "P", "E", "Z", "Y"];
-                                    let ndx = Math.floor(Math.log(row[column.id]) / Math.log(1000) );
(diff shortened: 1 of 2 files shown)


AdSchellevis:
@kulikov-a thanks, I've made some changes in https://github.com/opnsense/core/commit/f22c05ec61fa429b29e5d16e289ec3945d3ac78c, the `any(time_mark in record[timefield] for time_mark in list(time_marks.keys()))` was a bit  mind boggling, maybe you can check if my additions have the intended result.

kulikov-a:
@AdSchellevis yes, all works well!
Thanks!!

PR #6023, opened 2022-09-14
OpenVPN widget: bring back the buttons action

kulikov-a:
Hi!
seems to have overlooked a small regression after https://github.com/opnsense/core/commit/405395264ffb4f77e298a3bfa5d670b9a53eeda6 at openvpn widget (script with event binding stripped before execute).
(a little strange that there were no complaints)

thanks!


AdSchellevis:
@kulikov-a I guess it's hardly used, thanks for the fix!

kulikov-a:
thanks!

PR #6031, opened 2022-09-18
Backend: allow custom response on timeout

kulikov-a:
Hi!
For discussion (so marked as a draft).
What if the backend allowed to control the response in case of a timeout (e.g.,  tasks for which this is expected. like unbound dnsbl)?
This would allow a more accurate result message to be returned to the user?

Thanks!


AdSchellevis:
lots of changes in that area, if still valid and useful, best rebase the branch or open a new PR

kulikov-a:
absolutely agree, completely forgot about this request. the initial reason (unbound lazy load) also gone

Issue #6026, opened 2022-09-16
Applying Unbound blacklist fails with error


haarp:
- [x] I have read the contributing guide lines at https://github.com/opnsense/core/blob/master/CONTRIBUTING.md
- [x] I am convinced that my issue is new after having checked both open and closed issues at https://github.com/opnsense/core/issues?q=is%3Aissue

**Describe the bug**

After adding a new entry in the Unbound blocklist whitelist (going from 2 to 3 elements) and pressing Download&Apply, it waits a while and then I am greeted with an error box stating

```
Error updating blocklists
An error occurred during script execution. Check the logs for details
```

Backend log shows

```
Timeout (120) executing : unbound dnsbl
Script action failed with Command ' /usr/local/opnsense/scripts/unbound/blocklists.py && /usr/local/opnsense/scripts/unbound/wrapper.py -b ' returned non-zero exit status 99. at Traceback (most recent call last): File "/usr/local/opnsense/service/modules/processhandler.py", line 482, in execute subprocess.check_call(script_command, env=self.config_environment, shell=True, File "/usr/local/lib/python3.9/subprocess.py", line 373, in check_call raise CalledProcessError(retcode, cmd) subprocess.CalledProcessError: Command ' /usr/local/opnsense/scripts/unbound/blocklists.py && /usr/local/opnsense/scripts/unbound/wrapper.py -b ' returned non-zero exit status 99.
```

Subsequent attempts at pressing Download&Apply immediately return with the error box. I am currently unable to save at all.

**To Reproduce**

Steps to reproduce the behavior (I am not sure how reproducible it is on other installs):
1. Have Unbound blocklist active with Blocksite.list Fraud,Malware,Phishing,Ransomware,Scam; EasyList; EasyPrivacy; WindowsSpyblocker (spy)'
2. Have two elements in whitelist: `foo.com`, `.foo.com`
3. Download&Apply
4. Use for a while
5. Later, add third element `bar.com` to whitelist
6. Press Download&Apply

**Expected behavior**

Applies without trouble

**Describe alternatives you considered**

Rebooting the box perhaps? Not an option until at least the weekend. And I'd be workarounding, nox fixing things :)

**Screenshots**

-

**Relevant log files**

See above for log

**Additional context**

**Environment**

Software version used and hardware type if relevant, e.g.:

OPNsense 22.7.4-amd64
Intel(R) Celeron(R) J4125 CPU @ 2.00GHz (4 cores, 4 threads)

[...]

PR #6035, opened 2022-09-19
unbound: catch exception on blocklist reading

kulikov-a:
Hi!
another thing noticed when trying to reproduce https://github.com/opnsense/core/issues/6026
if the connection or server is very (i made it 5 kbit/s) slow (and we also reduced the timeout value), the update may fail with a non-obvious `Command ' /usr/local/opnsense/scripts/unbound/blocklists.py && /usr/local/opnsense/scripts/unbound/wrapper.py -b ' returned non-zero exit status 1` error in backend log. Perhaps it makes sense to catch exceptions when reading a file? Then a more understandable message like
 ```blocklist download : error reading file from https://blocklistproject.github.io/Lists/alt-version/malware-nl.txt (error : HTTPSConnectionPool(host='blocklistproject.github.io', port=443): Read timed out.)```  will get into the unbound log, and script execution will continue
Thanks!

### DIFF
--- a/src/opnsense/scripts/unbound/blocklists.py
+++ b/src/opnsense/scripts/unbound/blocklists.py
@@ -52,19 +52,24 @@ def uri_reader(uri):
         req.raw.decode_content = True
         prev_chop  = ''
         while True:
-            chop = req.raw.read(1024).decode()
-            if not chop:
-                if prev_chop:
-                    yield prev_chop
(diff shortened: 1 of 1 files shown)


AdSchellevis:
@kulikov-a thanks!

kulikov-a:
thanks!

PR #6040, opened 2022-09-19
Fireewall : Aliases - PHP8 error fix

kulikov-a:
Hi!
ref https://forum.opnsense.org/index.php?topic=30368.0
`abs()` "no longer accepts internal objects which support numeric conversion." [(ref)](https://www.php.net/manual/en/function.abs.php)

Thanks!


AdSchellevis:
@kulikov-a thanks!

Issue #6043, opened 2022-09-20
Dashboard breaks with PHP Fatal error if RSS widget rssfeed field is used


haarp:
- [x] I have read the contributing guide lines at https://github.com/opnsense/core/blob/master/CONTRIBUTING.md
- [x] I am convinced that my issue is new after having checked both open and closed issues at https://github.com/opnsense/core/issues?q=is%3Aissue

**Describe the bug**

Hello,

The RSS widget on the dashboard has an edit button, which gives me some dropdown boxes and a large unlabelled empty box. It's HTML id is `rssfeed`, so my assumption is that you can put feed URLs into this box. Upon doing so, e.g. by entering `https://www.reddit.com/r/opnsense/.rss` and saving, the dashboard becomes unable to load. The rest of the web UI also doesn't load properly anymore, leaving one wondering what to do :)

To fix this, one merely needs to remove the `<rssfeed>` line from config.xml again.

**To Reproduce**

1. Add `https://www.reddit.com/r/opnsense/.rss` to unlabelled field in RSS widget
2. Save
3. Try to reload dashboard

**Expected behavior**

Dashboard able to catch errors and continue working. A label for the large field in the widget would probably help too :)

**Describe alternatives you considered**

-

**Screenshots**

-

**Relevant log files**

The firmware reporter shows this error:

```
PHP Fatal error:  Uncaught TypeError: Unsupported operand types: string / int in /usr/local/opnsense/contrib/simplepie/library/SimplePie/Parse/Date.php:693
Stack trace:
#0 /usr/local/opnsense/contrib/simplepie/library/SimplePie/Parse/Date.php(602): SimplePie_Parse_Date->date_w3cdtf('2022-09-07T09:5...')
#1 /usr/local/opnsense/contrib/simplepie/library/SimplePie/Item.php(773): SimplePie_Parse_Date->parse('2022-09-07T09:5...')
#2 /usr/local/opnsense/contrib/simplepie/library/SimplePie.php(2890): SimplePie_Item->get_date('U')
#3 /usr/local/opnsense/contrib/simplepie/library/SimplePie.php(3020): SimplePie->get_items(0, 0)
#4 /usr/local/opnsense/contrib/simplepie/library/SimplePie.php(2832): SimplePie::merge_items(Array, 0, 0, 0)
#5 /usr/local/www/widgets/widgets/rss.widget.php(154): SimplePie->get_items()
#6 /usr/local/www/index.php(409): include('/usr/local/www/...')
[...]

PR #6044, opened 2022-09-20
simplepie: type cast

kulikov-a:
Hi
should close https://github.com/opnsense/core/issues/6043 imho
Thanks!


AdSchellevis:
@kulikov-a missed yours, looks like an upstream issue (https://github.com/simplepie/simplepie/issues/686), but I don't mind pulling a patch in in the mean time.  To prevent future breakage it looked like a good idea to catch the "Error" as well https://github.com/opnsense/core/commit/9a956d065a72153fbb36d3d6d6ff0c9d5b95b001

kulikov-a:
Thanks! And error catching is always good imho )

Issue #5978, opened 2022-08-22
unbound: move advanced settings page to MVC/API, log-reply option etc.


fichtner:
**Important notices**

Before you add a new report, we ask you kindly to acknowledge the following:

- [x] I have read the contributing guide lines at https://github.com/opnsense/core/blob/master/CONTRIBUTING.md
- [x] I am convinced that my issue is new after having checked both open and closed issues at https://github.com/opnsense/core/issues?q=is%3Aissue

**Is your feature request related to a problem? Please describe.**

Static pages should go away eventually. For Unbound to move further find smallest impact page(s).

**Describe the solution you like**

Advanced page migration is easy and gives opportunity to add more features.

**Describe alternatives you considered**

Moving all to MVC is too much work, too many side-effects.

**Additional context**

https://github.com/opnsense/core/issues/5795 https://github.com/opnsense/core/issues/5256

swhite2:
While at it, let's move https://github.com/opnsense/core/issues/5256 to this page as well.

Include `log-reply` conf option as well.

Issue #5256, opened 2021-10-03
Private Domains setting is hard to find under non-DNSBL use cases


delan:
**Important notices**

Before you add a new report, we ask you kindly to acknowledge the following:

- [x] I have read the contributing guide lines at https://github.com/opnsense/core/blob/master/CONTRIBUTING.md
- [x] I am convinced that my issue is new after having checked both open and closed issues at https://github.com/opnsense/core/issues?q=is%3Aissue

**Is your feature request related to a problem? Please describe.**

On my network, the search domains are home.daz.cat (containing private IPv4 addresses) followed by daz.cat (containing public addresses). For example, storage.home.daz.cat is 172.19.42.3, and storage.daz.cat is 2403:5800:c200:8600:5054:ff:fe01:4018.

But home.daz.cat is served publicly, which annoys unbound’s DNS rebinding prevention:

```
[17554:2] debug: sanitize: removing public name with private address <storage.home.daz.cat.> 172.19.42.3#53
```

**Describe solutions and alternatives**

At first I thought the only solution was to add a domain override (**Services** > **Unbound DNS** > **Overrides**), which happens to add private-domain and domain-insecure in addition to the forward-zone, but eventually I found **Blocklist** > **Private Domains**. Thankfully the setting works regardless of whether DNSBL is enabled.

Given my use case, I think this is a confusing place to put the setting. What do you think about moving it to **Advanced** or similar, and updating the help text to be less “DNSBL only”?

**Additional context**

* [blog post about the pfSense counterpart](https://blog.jenningsga.com/pfsense-dns-resolver-and-private-domains/)

fichtner:
Yes, it needs to be moved. And no it won’t be done fast and easy because a lot of work is still needed to modernise the code base we inherited.

I have this on my radar, but it’s a bigger issue than keeping track of when a single option is moved somewhere else regardless of if it’s properly documented or not. That’s another story.


Cheers,
Franco

delan:
No worries, thanks for the quick reply!

fichtner:
For a bit more context where private-domain setting came from https://github.com/opnsense/core/issues/5104

PR #6050, opened 2022-09-27
Unbound: convert advanced page to MVC

swhite2:
- adds the `Log Reply` and `Tag Queries and Replies` options to enhance logging capabilities as well.

PR #6053, opened 2022-09-28
Unbound: addition to advanced rework.

swhite2:
Small addition to https://github.com/opnsense/core/pull/6050:

- Move private/insecure domains to advanced as its intended use has a larger scope than DNSBLs only. Fixes https://github.com/opnsense/core/issues/5256
- Extends the migration to also include these domains.
- leftover cleanup of legacy settings in migration.
- Adds the `serve-expired-reply-ttl`, `serve-expired-ttl`, `serve-expired-ttl-reset`, `serve-expired-client-timeout` options. These options are hidden until the `serve-expired` checkbox is checked, and indented to signify a relationship.
- Removes all dropdowns and instead provides numeric fields to input raw values for more control and less "guessing" of what is acceptable.
- Removes default settings to prevent mismatches with upstream in the future. It's probably best to refer to the Unbound documentation in our own documentation.
- Previously, `rrset-cache-size` and `outgoing-range` were implicitly set. These are now input fields. The migration code will take care of legacy setting assumptions.

Fixes https://github.com/opnsense/core/issues/5978
Fixes https://github.com/opnsense/core/issues/5795


swhite2:
Seems https://github.com/opnsense/core/issues/5795 is missing, I'll update the PR

src/opnsense/service/templates/OPNsense/Unbound/core/+TARGETS
```diff
@@ -3,6 +3,5 @@ blocklists.conf:/tmp/unbound-blocklists.conf
 dot.conf:/usr/local/etc/unbound.opnsense.d/dot.conf
 private_domains.conf:/var/unbound/private_domains.conf
 domainoverrides.conf:/usr/local/etc/unbound.opnsense.d/domainoverrides.conf
-miscellaneous.conf:/usr/local/etc/unbound.opnsense.d/miscellaneous.conf
```
fichtner, line 6: this will let the file /usr/local/etc/unbound.opnsense.d/miscellaneous.conf linger in installs.we need to ship an empty template until 23.1 is out.

swhite2: got it, thanks for the heads-up

swhite2: https://github.com/opnsense/core/pull/6053/commits/b45a4b1609cd508764f0a9adb9608181c46e1f7c

fichtner (approved):

Issue #6047, opened 2022-09-24
Unbound API - Query forwarding


ansibleguy:
**Important notices**

Before you add a new report, we ask you kindly to acknowledge the following:

- [x] I have read the contributing guide lines at https://github.com/opnsense/core/blob/master/CONTRIBUTING.md
- [x] I am convinced that my issue is new after having checked both open and closed issues at https://github.com/opnsense/core/issues?q=is%3Aissue

**Describe the bug**

The API management of 'Query Forwarding' is not working as expected/documented.

API References:
- https://docs.opnsense.org/development/api/core/unbound.html
- https://github.com/opnsense/core/blob/master/src/opnsense/mvc/app/models/OPNsense/Unbound/Unbound.xml

There is a 'type' switch that does nothing - it will always use 'dot' (_DNS-over-TLS_) as type.

```xml
<type type="OptionField">
  <Required>Y</Required>
  <default>dot</default>
  <OptionValues>
    <dot>DNS over TLS</dot>
    <forward>Forward</forward>
  </OptionValues>
</type>
```
Here is the type as seen in the 'get' response:

```json
{"type":{"dot":{"value":"DNS over TLS","selected":0},"forward":{"value":"Forward","selected":1}}
```

If one does try to set the 'type' to 'forward' - as seen in the documentation - it does create a 'dot' item.

```
REQUEST: POST | HEADERS: '{'Content-Type': 'application/json'}' | URL: https://FW/api/unbound/settings/addForward | DATA: {'dot': {'type': 'forward', 'enabled': 1, 'domain': 'tmp.at', 'server': '1.1.1.1', 'port': 53}}
RESPONSE: {'status_code': 200,... '_content': b'{"result":"saved","uuid":"..."}'}
```

The commit this behaviour changed seems to be the following one: https://github.com/opnsense/core/commit/6832fd75a0b41e376e80f287f8ad3cfe599ea3d1#diff-a3952cd78b5ddaabed637d6331fbd16ec8f4d04bba06f6689b991493156fb66e

Currently the 'type' can only be set using the '**Referer**' http-header.

**BUT** it does only work when targeting the undocumented '[add|set|del]Dot' API endpoint.

Per example:
```
# this request creates an item of type 'dot'
[...]

PR #6054, opened 2022-09-28
unbound: allow setting forwarder type via API

kulikov-a:
Hi!
ref. https://github.com/opnsense/core/issues/6047
It turned out that the unbound api does not allow adding a simple (non-DoT) forwarder in the usual way (have to emulate the browser behavior by setting the HTTP Referer header value).
PR tries to add the ability to set the forwarder type by explicitly specifying the type in the request.
Thanks!

### DIFF
--- a/src/opnsense/mvc/app/controllers/OPNsense/Unbound/Api/SettingsController.php
+++ b/src/opnsense/mvc/app/controllers/OPNsense/Unbound/Api/SettingsController.php
@@ -63,7 +63,8 @@ class SettingsController extends ApiMutableModelControllerBase
         if (substr($method, -6) == 'Action') {
             $fn = preg_replace('/Dot/', 'Forward', $method);
             if (method_exists(get_class($this), $fn)) {
-                if (preg_match("/forward/i", $this->request->getHTTPReferer())) {
+                // set the type to forward if the request came from the corresponding gui page or is explicitly set in the API call
+                if (preg_match("/forward/i", $this->request->getHTTPReferer()) || $this->request->getPost('dot')['type'] == 'forward') {
                     $this->type = "forward";
                 }
                 return $this->$fn(...$args);


swhite2:
Good idea, thanks! :)

kulikov-a:
Thanks!

ansibleguy:
Hi.

@kulikov-a 
As I understand it this change allows the usage of the 'type'-switch for the undocumented [add|set|del|trigger]Dot api-endpoints. (_as '\_\_call' only catches the non-existing method-calls_)

Please correct me if I'm wrong. (:

Per example:
```
# this request creates an item of type 'dot'
POST: https://FW/api/unbound/settings/addForward | DATA: {'dot': {'type': 'forward', ...}}
# this one successfully creates an item of 'type' forward
POST: https://FW/api/unbound/settings/addDot | DATA: {'dot': {'type': 'forward', ...}}
```

Would it be possible to add that switch also for the documented [add|set|del|trigger]Forward endpoints?

- AnsibleGuy

kulikov-a:
@ansibleguy in my understanding of the model and api,  [add|set|del|trigger]Dot calls should be treated as a right API endpoints (since all forwarders (simple and DoT) settings lives in `dots`. and and gui uses them) as it was before https://github.com/opnsense/core/commit/6832fd75a0b41e376e80f287f8ad3cfe599ea3d1 migration.
so imho this is only a matter of docs

ansibleguy:
@kulikov-a I get your point.
If the documentation can be updated to **only include `Dot`** endpoints the usage would be clear to the api-endusers.
Else it still does not make sense that one could/would per example call the addForward endpoint and could NOT add an item of type `forward`.

But as mentioned in [the comment](https://github.com/opnsense/core/issues/6047#issuecomment-1257150343) of #6047 it seems the docs can only include the `Forward` endpoints..

\- AnsibleGuy

PR #6071, opened 2022-10-06
Default subnet mask of 24 instead of 32 if value unset

codiflow:
**Improvement**
Default value for IPv4 subnet mask on interfaces should be 24 not 32

Currently the default value for a subnet mask is 32 if the value was not set before. This for example is the default if you are changing an interface from DHCP to Static IPv4. The downside is that someone who oversees the value might lock himself out from the machine.

Example:
Someone (indeed this someone was me 😜) is connected to the firewall through VPN and changes the WAN interface from DHCP to Static IPv4. After entering the right IPv4 address the person forgets to set the corresponding subnet mask from 32 to 24 (where the latter is mostly standard on local networks). After applying the changes the person gets locked out from the firewall and has to fix it locally because no traffic can get in and out.

A default setting of 24 instead of 32 would have prevented this situation. A subnet mask of 32 also doesn't make any sense on a WAN interfaces IP address from my perspective.

### DIFF
--- a/src/etc/inc/certs.inc
+++ b/src/etc/inc/certs.inc
@@ -824,7 +824,7 @@ function is_openvpn_server_crl($crlref)
 
 function is_crl_internal($crl)
 {
-    return (!(!empty($crl['text']) && empty($crl['cert'])) || ($crl["crlmethod"] ?? '' == "internal"));
+    return $crl["crlmethod"] ?? '' == "internal";
 }
 
 function cert_get_cn($crt, $isref = false)
--- a/src/www/interfaces.php
+++ b/src/www/interfaces.php
@@ -2113,7 +2113,7 @@ include("head.inc");
 
 <?php
                                     for ($i = 32; $i > 0; $i--):?>
-                                    <option value="<?=$i;?>" <?=$i == $pconfig['subnet'] ? "selected=\"selected\"" : "";?>><?=$i;?></option>
+                                    <option value="<?=$i;?>" <?=($i == $pconfig['subnet'] || ($pconfig['subnet'] == "" && $i == 24)) ? "selected=\"selected\"" : "";?>><?=$i;?></option>
 <?php
                                     endfor;?>
                                   </select>
--- a/src/www/system_crlmanager.php
+++ b/src/www/system_crlmanager.php
@@ -237,6 +237,7 @@ if ($_SERVER['REQUEST_METHOD'] === 'GET') {
                 $crl['cert'] = array();
             }
 
+            crl_update($crl);
             if (!isset($id)) {
                 $a_crl[] = $crl;
             }


fichtner:
A mismatch of IP mask always has consequences one way or another. The fix might be nicer to handle with JS also but I've never heard this complaint in all the years before...

codiflow:
Another idea would be to leave the subnet mask blank after switching to IPv4 for the first time and let the browser handle the field which could be marked as required. So someone has to interact with the field just once before the setting is stored.

fichtner:
Unfortunately I'm not convinced that /24 is the right preselection. select doesn't have an empty value, adding and validating it adds even more code and makes the workflow longer than necessary. People can still lock themselves out choosing a number of values, not only /32.

codiflow:
> Unfortunately I'm not convinced that /24 is the right preselection.

What would your recommendation be? I think we are aligned that 32 is not a good preselection, are we?

> People can still lock themselves out choosing a number of values, not only /32.

You are right. Nevertheless the procedure to switch from Static IPv4 to DHCP or vice versa is a common task and should prevent situations like that by having proper default values. We have lots of ways to provide the user a experience where he has to overthing given values twice or he has to set wrong values actively – not passively.

> select doesn't have an empty value, adding and validating it adds even more code and makes the workflow longer than necessary.

Theres another idea we could use to prevent this without setting a default value beforehand:
https://stackoverflow.com/a/46710230

The downside of those techniques would be that you would need to test them on most browsers – I don't know if all of them would respect it.

codiflow:
@AdSchellevis May I know why you just closed this PR? There is neither a proper solution for the issue I'm refering too, nor are there good arguments to just close the issue.

I understand if there might be more "important" issues to solve. But please keep in mind that just closing a PR without any comment or explanation for the closure does not help keeping the motivation high to contribute to a project.

fichtner:
We discussed this at length yesterday. The problem you state is that you haven’t set the proper netmask. Your patch changes the situation to a more favourable way, but it’s not a fix of the problem class. Secondly the problem here was never raised in the bug tracker before and the real world value is in doubt. Thirdly interfaces.php is a legacy page and we would like to avoid unnecessary or lengthy additions. So lastly a proper fix would require more logic and code, possibly in JS conflicting with other points here and it’s doubtful the complexity is worth doing something in this regard.


Cheers,
Franco

codiflow:
Hi Franco, 

thanks for the answer.

I don't get the argument that you won't implement it just because nobody raised the issue ever before. That doesn't implies that the problem does not exist. I asked you if we are consent that the default value for the subnet mask of /32 doesn't make sense for most of the WAN cases but I unfortunately did not get an answer on this. I still think it is not a good default value.

> So lastly a proper fix would require more logic and code, possibly in JS conflicting with other points here and it’s doubtful the complexity is worth doing something in this regard. 

Another idea would be to simply **replace the subnet dropdown with a text input** in general which gets checked by a regex before sending the form. This way the field could be left empty by default (like the IP address which is empty by default too) and people have to interact with it at least once. This would not need anything in addition which is not already there and solves the problem.

> but it’s not a fix of the problem class.

A final fix could also be to **combine the fields for IP address and subnet mask into one field** which would reduce overall complexity at the end because only one field has to be validated / propagated instead of two.

Another **personal** thing:
This is my first PR for OPNsense and it doesn't feel well how things went. It would have been nice to get an answer like yours before your collegue just closes the PR without any comment. I don't think that's a good way to threat people who are contributing and looking for a solution in a constructive way. Your arguments sound to me a bit like "we don't want to change anything here" and I'm missing the will of finding a solution together for the issue I raised.
If you personally don't want to implement a fix or don't see the necessity of a fix I'm fine with that. **You are the maintainer, you decide.** But please don't mask your personal decision with arguments which you are not willing to discuss on.

I got from you that OPNsense is a system where you should think twice and check values again before you apply them. That is true and I fully back that. Thats what's a sysadmin's job. But I think all of us know that reality sometimes is a bit different and we are happy if technical systems "assist" us by avoiding potential lockout situations, isn't it?

Exactly this kind of avoidance is what I tried to implement with my patch.

fichtner:
Text input is probably the most elegant way here, drop-down breaks flow and an empty drop down value as an alternative seems silly since the point of drop-down is to offer a preselection. Merging fields is out of the question for complexity.

but now consider changing UX for one single component and leaving out the fact that drop downs are everywhere for subnet size.

Try to explain to others why any single change here is better than before. Others are far likely to complain about inconsistency as a whole or changes in particular that come out of nowhere without a technical reason underneath.

if you feel you have been treated unjustly that fair since everyone is entitled to their unique experience. For me I’m happy to stop responding because too much time went into this already.


Happy holidays,
Franco

AdSchellevis:
> It would have been nice to get an answer like yours before your collegue just closes the PR without any comment.

While preparing for 23.1 we (as a team) assumed there had been enough discussion on this PR, I should have probably mentioned that while closing the issue. 

Long story short, at the moment we don't feel like changing this area as whatever fix we implement it will likely always raise similar issues. Enforcing a mask (don't split the address and mask) would likely be the best option, but given the complexity already in the page and the burden of maintenance for legacy components we simply don't think we should aim for that. 

I can certainly understand that this wasn't the best first experience in contributing code, a way to prevent these things from happening is starting with a ticket first and asking for feedback.  If an issue is more wide spread, more people feel alike, it is more likely changes will be incorporated at some point. Unfortunately we still have the burden of some legacy code to maintain which in the long run we would rather replace, but we do have to prioritise.  

Pull requests on core are weighted heavily, as in our experience the number of regressions we have to fix after merging code from others often out weights the amount of time needed to make the initial PR.

Looking at this PR for example we normally wouldn't pin a default in the select but move it on top of the page as an explicit default, in which case the select remains simple in terms of logic. For MVC pages these type of issues don't exist as the model and form functions handle this automatically. 

When it comes to closing tickets and pull-requests, we automated the ticket part as mentioned in the contribution guidelines, pull requests we inspect manually before release preparation. 
A lot of projects just keep them open endlessly which avoids these discussions, but also provides a false view on the reality in my humble opinion, priorities change and people do need to spend time.

I hope this explains things a bit.

codiflow:
@AdSchellevis I really thank you for your detailed answer and the explanations. This helped me to understand the background.

> Pull requests on core are weighted heavily, as in our experience the number of regressions we have to fix after merging code from others often out weights the amount of time needed to make the initial PR.

> A lot of projects just keep them open endlessly which avoids these discussions, but also provides a false view on the reality in my humble opinion, priorities change and people do need to spend time.

I totally understand what you are pointing out and I think we can just leave the PR closed according to this explanation. I may find a different way to fix this for me and others in the future.

I dislike the way @fichtner was communicating here – especially according to what I wrote about my personal experience. So we leave this as my personal "unique experience". No further discussion needed from my side.

Happy holidays too
Christian

PR #6074, opened 2022-10-07
interfaces: add VlanDeviceField and proper population of text field

fichtner:
This allows to add "labels" to internal values for text fields. In this case VLAN device names can be associated with their assigned interface since when these are assigned one cannot rename them. Which is not obvious while trying to do so.

Small issue remains that "option not in list" is returned for selecting a value not yet known to the configuration.


AdSchellevis:
@fichtner how about https://github.com/opnsense/core/commit/a31a21bd93b04bf1917ee8e976bbadf0370b4268 ?

fichtner:
@AdSchellevis I'll take it, thanks! ❤️

Issue #5636, opened 2022-03-20
VPN / IPsec - migrate ipsec.conf to swanctl.conf


AdSchellevis:
**Important notices**

Before you add a new report, we ask you kindly to acknowledge the following:

- [x] I have read the contributing guide lines at https://github.com/opnsense/core/blob/master/CONTRIBUTING.md
- [x] I am convinced that my issue is new after having checked both open and closed issues at https://github.com/opnsense/core/issues?q=is%3Aissue

**Is your feature request related to a problem? Please describe.**

In time Strongswan will remove support for `ipsec.conf` (could't find a clear path on their end, if there is, it might be good to add a dead line here), in which case the newer `swanctl.conf` should be used.

**Describe the solution you like**

Migrate current legacy code to output new style configuration data instead of the current ipsec.conf file.

Currently the ipsec.conf is written here:

https://github.com/opnsense/core/blob/c3bae88bf5195e3f54b6e7bfba8cafce633164c5/src/etc/inc/plugins.inc.d/ipsec.inc#L1754

A challenging part might be the current support for custom includes in :
https://github.com/opnsense/core/blob/c3bae88bf5195e3f54b6e7bfba8cafce633164c5/src/etc/inc/plugins.inc.d/ipsec.inc#L1752

**Describe alternatives you considered**

None

**Additional context**

- Migration note from Strongswan https://wiki.strongswan.org/projects/strongswan/wiki/Fromipsecconf

AdSchellevis:
Rebased on master, apply https://github.com/opnsense/core/issues/4614 first on 22.7.4 to test
```
opnsense-patch 9448b0c
```

AdSchellevis:
merged

fichtner:
@AdSchellevis marked this "roadmap" as well :)

fichtner:
FWIW, I still get these warnings on bootup:

```
[10-Oct-2022 20:37:13 Europe/Berlin] PHP Warning:  Undefined array key "ikeid" in /usr/local/etc/inc/plugins.inc.d/ipsec.inc on line 244
[10-Oct-2022 20:37:16 Europe/Berlin] PHP Warning:  Undefined array key "ikeid" in /usr/local/etc/inc/plugins.inc.d/ipsec.inc on line 844
[10-Oct-2022 20:37:16 Europe/Berlin] PHP Warning:  Undefined array key "ikeid" in /usr/local/etc/inc/plugins.inc.d/ipsec.inc on line 390
[...]

PR #6076, opened 2022-10-08
VPN / IPsec - migrate ipsec.conf to swanctl.conf

AdSchellevis:
VPN/IPsec - work in progress refactoring ipsec.conf to swanctl.conf for https://github.com/opnsense/core/issues/5636

Wrap the following blocks in functions:

o generation of strongswan.conf
o write and cleanup IPsec's CA database
o write certificates used by IPsec
o write RSA Key Pairs
o construct ipsec.secrets for swanctl, combining different sources, phase1 PSK's, certificates, keypairs
o replace VPN: IPsec: Lease Status with new mvc version that uses swanctl underneath

Remap ipsec.conf to swanctl.conf attributes using https://wiki.strongswan.org/projects/strongswan/wiki/Fromipsecconf

Remove lifetime defaults from phase 1 and 2 so new connections will use Strongswan's defaults and move the option in phase above Margintime/Rekeyfuzz as these combined are being migrated (see margintime in Fromipsecconf doc).

Restructure used path directives so we can use the default structure as descried in https://docs.strongswan.org/docs/5.9/swanctl/swanctlDir.html and installed via ports.

IPsec status overview needs a small fix as well (passthrough network seems to be returned by vici, but doesn't contain a class (which crashes list_status.py).)

While here, also remove some php warnings.

Most tests performed are described in my repo https://github.com/AdSchellevis/OPNsense_IPSec_testsuite


AdSchellevis:
@fichtner I think this is it for the master branch, documentation needs updates, but there's quite some time until 23.1 is released

mimugmail:
Happy to test when in. Currently again investing time regarding EAP-TLS so I have some Labs around :)

AdSchellevis:
@mimugmail thanks, it was quite some work refactoring this, I can certainly use an extra pair of eyes on the results.

src/opnsense/service/conf/actions.d/actions_ipsec.conf
```diff
@@ -59,13 +65,13 @@ type:script
 message:IPsec service start
 
 [stop]
-command:/usr/local/sbin/ipsec stop
+command:/usr/local/etc/rc.d/strongswan stop
```
fichtner, line 68: should this not use `onestop`?

src/opnsense/service/conf/actions.d/actions_ipsec.conf
```diff
@@ -59,13 +65,13 @@ type:script
...
 parameters:
 type:script
 message:IPsec service stop
 
 [restart]
-command:/usr/local/sbin/ipsec stop; /usr/local/sbin/pluginctl -c ipsec
```
fichtner, line 68: `onestop`?

src/etc/inc/plugins.inc.d/ipsec.inc
```diff
@@ -843,820 +812,756 @@ function ipsec_configure_spd()
...
-                $ph1publickeyfile = "${publickeypath}/publickey-local-{$ph1ent['ikeid']}.pem";
-                if (!file_put_contents($ph1publickeyfile, $keyPair['publicKey'])) {
-                    log_error(sprintf('Error: Cannot write phase1 local public key file for %s', $ph1ent['name']));
-                    @unlink($ph1publickeyfile);
-                    continue;
-                }
-                @chmod($ph1publickeyfile, 0600);
+    if (!isset($ipseccfg['enable'])) {
+        /* try to stop charon */
+        mwexec('/usr/local/etc/rc.d/strongswan stop');
```
fichtner, line 1247: `onestop`?

src/etc/inc/plugins.inc.d/ipsec.inc
```diff
@@ -843,820 +812,756 @@ function ipsec_configure_spd()
...
+            echo 'Configuring IPsec VPN...';
+        }
 
-                $ph1publickeyfile = "${publickeypath}/publickey-peer-{$ph1ent['ikeid']}.pem";
-                if (!file_put_contents($ph1publickeyfile, $keyPair['publicKey'])) {
-                    log_error(sprintf('Error: Cannot write phase1 peer public key file for %s', $ph1ent['name']));
-                    @unlink($ph1publickeyfile);
-                    continue;
+        // cleanup legacy ipsec.conf files
+        if (is_file("/usr/local/etc/ipsec.conf") && is_dir("/usr/local/etc/ipsec.d")) {
```
fichtner, line 1267: ipsec.conf is possibly sampled by strongswan package so always there even if we delete it (but need to double-check)

src/etc/inc/plugins.inc.d/ipsec.inc
```diff
@@ -843,820 +812,756 @@ function ipsec_configure_spd()
...
+            foreach (new RecursiveIteratorIterator(new RecursiveDirectoryIterator("/usr/local/etc/ipsec.d")) as $x) {
+                if (is_file($x)) {
+                    @unlink($x);
                 }
-                @chmod($ph1publickeyfile, 0600);
             }
+            @unlink("/usr/local/etc/ipsec.conf");
+            @unlink("/usr/local/etc/ipsec.secrets");
```
fichtner, line 1274: same for this file

src/etc/inc/plugins.inc.d/ipsec.inc
```diff
@@ -1786,6 +1691,7 @@ function ipsec_configure_vti($verbose = false, $device = null)
         }
         if (
             empty($configured_intf[$intf])
+            || empty($intf_details['tunnel'])
```
fichtner, line 1694: what's the context here?

AdSchellevis: the two lines below,  if `tunnel` doesn't exist, this `$local_configured != $intf_details['tunnel']['src_addr']` is always true, but will fail with a php error (at least as of php 8).

src/opnsense/service/conf/actions.d/actions_ipsec.conf
AdSchellevis: fixed in next commit

src/etc/inc/plugins.inc.d/ipsec.inc
fichtner: 👍

fichtner:
Merged, thanks!

PR #6082, opened 2022-10-12
Add description to filter sync action

jan-win1993:
To automate the configuration synchronization to the backup firewall with a cronjob a description in the "filter sync" action is needed.

### DIFF
--- a/src/opnsense/service/conf/actions.d/actions_filter.conf
+++ b/src/opnsense/service/conf/actions.d/actions_filter.conf
@@ -8,6 +8,7 @@ message:Reloading filter
 command:/usr/local/etc/rc.filter_synchronize
 parameters:
 type:script
+description:Synchronize configuration to backup firewall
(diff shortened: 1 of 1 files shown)


src/opnsense/service/conf/actions.d/actions_filter.conf
```diff
@@ -8,6 +8,7 @@ message:Reloading filter
 command:/usr/local/etc/rc.filter_synchronize
 parameters:
 type:script
+description:Synchronize configuration to backup firewall
 message:Syncing firewall %s
```
fichtner, line 12: side note, this `%s` is probably spurious

fichtner:
I'd probably opt for a better description but the decision here is for @AdSchellevis to include this in general

jan-win1993:
@AdSchellevis thanks for the recommendatio :)
Sorry my bad, it's also already in the documentation.

Issue #6027, opened 2022-09-16
Unbound blocklist (DNSBL): Add support for returning NXDOMAIN instead of 0.0.0.0


cellsplicer:
**Important notices**

Before you add a new report, we ask you kindly to acknowledge the following:

- [ x ] I have read the contributing guide lines at https://github.com/opnsense/core/blob/master/CONTRIBUTING.md
- [ x ] I am convinced that my issue is new after having checked both open and closed issues at https://github.com/opnsense/core/issues?q=is%3Aissue

**Is your feature request related to a problem? Please describe.**

There was a previous feature request and subsequent work for adding NXDOMAIN support in Unbound-plus - see https://github.com/opnsense/plugins/issues/1842

This is important as 0.0.0.0 causes connection delays and poor performance on Apple devices running third party software. Apple has forced third party developers to use Network Extensions since 2020.

**Describe the solution you like**

I would like a check box on the ui/unbound/dnsbl/index page for returning NXDOMAIN instead of "Destination Address".

**Describe alternatives you considered**

Using piHole which has the functionality to configure returning NXDOMAIN or any other destination IP address for the blocklist.

**Additional context**

N/A

haarp:
0.0.0.0 is an invalid IP address. I was always under the impression that most libraries would almost instantly fail when trying to connect to it. Does it work differently on Apple devices?

cellsplicer:
It can be replicated on my M1 Macbook Pro running Big Sur with corporate internet filtering:
`time curl 0.0.0.0 -s > /dev/null`
`curl 0.0.0.0 -s > /dev/null  0.00s user 0.01s system 1% cpu 1.024 total`

`time curl 127.0.0.1 -s > /dev/null`
`curl 127.0.0.1 -s > /dev/null  0.00s user 0.01s system 69% cpu 0.012 total`

As can be seen above the request to 0.0.0.0 blocks for a full second.

According to an old bug report for Firefox https://bugzilla.mozilla.org/show_bug.cgi?id=1678446 it occurs when there are any user space applications hooking into MacOS Network Extension framework such as a VPN client or content filter. I happen to have both as enforced by corporate IT policy.

refuseduser:
Hello,
NXDOMAIN is better than redirecting to a wrong address, for example, but it is still confusing because the domain exists but you don't want to resolve it.
Response code 5 (REFUSED, RFC1035) would be better, where it is clear that the response to the request is refused due to a policy.
[...]

PR #6083, opened 2022-10-12
unbound: rework DNSBL implementation to python module

swhite2:
This PR replaces the current blocklist implementation to use python instead of relying on `unbound-control`. The latter had the drawback of a very long execution time to administrate the local-data entries both locally and in Unbound. The memory footprint was also considerably larger due to unbound internals, while the python module keeps it all in memory in a simple dictionary - reducing the total amount of memory consumption by more than a factor of 10. A drawback is a potential decrease in performance of ~15%, although most setups shouldn't be affected by this as most hardware which is capable of running this should be scaled towards its intended use.

The option of returning NXDOMAIN has also been added (fixes https://github.com/opnsense/core/issues/6027), which in this implementation is a lot easier than what we would have to do if `local-data` were to be used.


src/opnsense/mvc/app/views/OPNsense/Unbound/dnsbl.volt
```diff
@@ -27,10 +27,21 @@
...
+               $('input[id="unbound.dnsbl.address"]').attr('disabled', 'disabled') :
+               $('input[id="unbound.dnsbl.address"]').removeAttr('disabled');
+       }
+
        var data_get_map = {'frm_dnsbl_settings':"/api/unbound/settings/get"};
        mapDataToFormUI(data_get_map).done(function(data){
            formatTokenizersUI();
            $('.selectpicker').selectpicker('refresh');
+
+           $('input[id="unbound.dnsbl.nxdomain"]').click(function() {
```
AdSchellevis, line 41: it's easier to bind to `change()`, like :

https://github.com/opnsense/core/blob/b6f95cdea429d243f1e70f55ce817407264eb60f/src/opnsense/mvc/app/views/OPNsense/Firewall/alias.volt#L227

src/opnsense/mvc/app/views/OPNsense/Unbound/dnsbl.volt
```diff
@@ -27,10 +27,21 @@
 
 <script>
    $(document).ready(function() {
+       function nxdomain_toggle() {
+           $('input[id="unbound.dnsbl.nxdomain"]').is(':checked') ?
+               $('input[id="unbound.dnsbl.address"]').attr('disabled', 'disabled') :
```
AdSchellevis, line 32: shouldn't be needed to disable, the current value is in the form anyway.

src/opnsense/service/templates/OPNsense/Unbound/core/+TARGETS
```diff
@@ -5,3 +5,4 @@ private_domains.conf:/var/unbound/private_domains.conf
 domainoverrides.conf:/usr/local/etc/unbound.opnsense.d/domainoverrides.conf
 root.min.hints:/var/unbound/root.hints
 unbound_dhcpd.conf:/usr/local/etc/unbound_dhcpd.conf
+dnsbl_module.py:/var/unbound/dnsbl_module.py
```
AdSchellevis, line 8: One could debate if code should be stashed in templates, but I'm not sure about any better location either at the moment. We can always move it later on if needed.

AdSchellevis (changes requested):
Looks good, left some small comments

src/opnsense/mvc/app/views/OPNsense/Unbound/dnsbl.volt
swhite2: Removed it as per your next comment for simplicity's sake.

src/opnsense/service/templates/OPNsense/Unbound/core/+TARGETS
swhite2: Agreed, it does not happen often (if at all). If the use case would turn out bigger we could always define a separate mechanism.

src/etc/inc/plugins.inc.d/unbound.inc
```diff
@@ -107,7 +107,9 @@ function unbound_generate_config()
         return;
     }
 
-    $dirs = array('/dev', '/etc', '/lib', '/run', '/usr', '/usr/local/sbin', '/var/db', '/var/run');
+    $pythonv = readlink('/usr/local/bin/python3');
```
fichtner, line 110: ❤️

src/etc/inc/plugins.inc.d/unbound.inc
```diff
@@ -325,6 +334,9 @@ include: /var/unbound/etc/*.conf
 
 {$forward_conf}
 
+python:
```
fichtner, line 337: what if dnsbl is disabled?

src/opnsense/mvc/app/models/OPNsense/Unbound/Migrations/M1_0_4.php
```diff
@@ -0,0 +1,43 @@
...
+
+use OPNsense\Base\BaseModelMigration;
+use OPNsense\Core\Backend;
+
+class M1_0_4 extends BaseModelMigration
+{
+    public function post($model)
+    {
+        /* ensure an existing blocklist is migrated during version upgrade */
+        if (!file_exists('/var/unbound/data/dnsbl.json')) {
```
fichtner, line 39: if this happens on a major upgrade it might not happen at all or stall the install... :(

src/etc/inc/plugins.inc.d/unbound.inc
swhite2: python is used unconditionally, the scripts itself picks up on dnsbl being enabled/disabled. a common code path makes more sense if we're looking towards the future regarding statistics.

src/etc/inc/plugins.inc.d/unbound.inc
swhite2: :)

src/opnsense/mvc/app/models/OPNsense/Unbound/Migrations/M1_0_4.php
swhite2: I don't have another suggestion apart from informing users to manually reapply their blocklist settings. In any case, some form of migration is necessary

fichtner: We can add an empty dnsbl.json sample file in unbound.opnsense.d also because we shouldn't anchor the file origin inside the chroot as we might want to clear it on each reboot perhaps. /var MFS is gone but I think for chroots we should treat it as if volatile.
Then we could add a late dnsbl call to rc.syshook "start" facility if we are concerned about "timely" updates.

fichtner: FWIW, base migration starts prior to bootstrapping the chroot for unbound in a new install / image.

swhite2: @fichtner https://github.com/opnsense/core/pull/6083/commits/9205a0fd868aba9a58fea6ecfb3785aea2d2efd7

src/etc/inc/plugins.inc.d/unbound.inc
```diff
@@ -117,6 +119,13 @@ function unbound_generate_config()
         mwexecf('/sbin/mount -t devfs devfs %s', '/var/unbound/dev');
     }
 
+    $python_dir = "/usr/local/lib/{$pythonv}";
+    $chroot_python_dir = "/var/unbound{$python_dir}";
+
+    if (mwexec("/sbin/mount | grep {$chroot_python_dir}")) {
```
fichtner, line 125:
```suggestion
    if (mwexec("/sbin/mount | grep {$chroot_python_dir}", true)) {
```

Because we can ignore error since we handle it here

src/opnsense/scripts/unbound/blocklists.py
```diff
@@ -145,11 +155,16 @@ def uri_reader(uri):
...
-                unbound_outf.write("local-data: \"%s A %s\"\n" % (entry, destination_address))
+            json.dump(blocklist_items, unbound_outf)
+
+    # atomically replace the current dnsbl so unbound can pick up on it
+    os.replace('/var/unbound/data/dnsbl.json.new', '/var/unbound/data/dnsbl.json')
 
     syslog.syslog(syslog.LOG_NOTICE, "blocklist download done in %0.2f seconds (%d records)" % (
         time.time() - startup_time, len(blocklist_items)
     ))
+
```
fichtner, line 170:
```suggestion
```

fichtner (approved):
ok for me with the two code suggestions

didn't you want to remove wrapper.py code as well?

swhite2:
> didn't you want to remove wrapper.py code as well?

Thanks for the heads up. Done

Issue #4831, opened 2021-03-16
The filesystem full error makes flowd_aggregate die while there is a lot of space left


vnxme:
**Important notices**

Before you add a new report, we ask you kindly to acknowledge the following:

- [X] I have read the contributing guide lines at https://github.com/opnsense/core/blob/master/CONTRIBUTING.md
- [X] I have searched the existing issues and I am convinced that mine is new.

**Describe the bug**

The flowd_aggregate.py dies from time to time on my firewall. According to the general log, database or disk is full, but there is obviously a lot of space left. Unfortunately, the script doesn't restart without me logging in and pressing the start button on the dashboard. Please refer to the screenshots below for details.

**To Reproduce**

Not sure how to reproduce this behaviour. I would understand if the script died when the disk was really full.

**Expected behavior**

I guess, there are 3 options to think about:
1. There might be a bug in the script or somewhere else which could be corrected;
2. A watchdog (e.g. cron job) might be needed to manually restart the script if it dies;
3. A file rotation policy might be implemented to always keep some free space.

**Screenshots**

![image](https://user-images.githubusercontent.com/46669194/111272357-91ae9280-8643-11eb-97bf-cde436becbe2.png)
![image](https://user-images.githubusercontent.com/46669194/111272872-34671100-8644-11eb-8f4d-d3de2bfe45b9.png)
![image](https://user-images.githubusercontent.com/46669194/111272924-41840000-8644-11eb-8bf3-0a8e51a024f5.png)

**Environment**

OPNsense 21.1.3_3 (amd64, OpenSSL), but I remember the script died even before 21.1.

PR #6085, opened 2022-10-13
Customize amount of Netflow aggregator history kept

n-thumann:
This PR introduces the ability for a user to customize the amount of history that is kept by the Netflow aggregator instead of using hardcoded values. For the source address details there's always 62 days of history kept, even though a user might want to keep more / less.

Why is this needed?
I regularly ran into the same issue as https://github.com/opnsense/core/issues/4831, https://forum.opnsense.org/index.php?topic=30071.msg145214 or https://forum.opnsense.org/index.php?topic=20489.msg95251. In these cases the SQLite DBs of the aggregator take up a lot of space and cause the vacuum to fail.

I therefore moved the hardcoded values to a config file, which is written by configd. I kept the old history values as defaults. The newly added fields can be access via Reporting -> Netflow after enabling the advanced settings.


AdSchellevis:
I'll try to take a look, but at a first glance I would rather try to focus on less toggles facing the user.

n-thumann:
> focus on less toggles facing the user

Yeah, i was thinking about this as well, but am struggling to come up with a good idea. Any ideas and how to incorporate them in the UI?

AdSchellevis:
I'm not sure yet either to be honest, we could try to focus on the most problematic files in size first and uniform it a bit, the more aggregated ones are likely less problematic.  How large are the files on your installation?

AdSchellevis:
Maybe it's enough to just offer the maximum numbers of days to keep given the daily summary sizes are the problematic ones. 

If we  choose to configure the `src_addr_details_086400.sqlite` to a relative number, we only need a single setting (1/6 th of specified days for example).

n-thumann:
Good point! How about slightly extending this to all resolution categories: Instead of adding a field for each aggregator, we only add one for each resolution available (i.e. 30, 300, ...)? This would reduce the number of fields added from eleven down to four.

AdSchellevis:
I was more thinking in the direction of configuring the max and stick source details to a relative factor e.g 1/6 x configured would be roughly the same as it is now.

AdSchellevis:
not enough priority on our end, we would like to rewrite the storage at some point, but the amount of time clashes with priorities.

Issue #6088, opened 2022-10-15
Dynamic state reset is too overbearing


haarp:
Before you add a new report, we ask you kindly to acknowledge the following:

- [x] I have read the contributing guide lines at https://github.com/opnsense/core/blob/master/CONTRIBUTING.md
- [x] I am convinced that my issue is new after having checked both open and closed issues at https://github.com/opnsense/core/issues?q=is%3Aissue

**Is your feature request related to a problem? Please describe.**

Hello!

The Firewall→Settings→Advanced→Dynamic state reset option is useful. However, it will reset *every* state. This includes innocent connections unrelated to the affected interface. e.g.:

- Connections to a remote server over primary WAN die when failover WAN hiccups.
- SSH sessions into the firewall itself die when WAN changes.

This poses an unnecessary danger to long-lived connections, and makes working with and testing/debugging things involving connection resets a lot harder and frustrating.

**Describe the solution you like**

**Idea 1:** The simple approach, previously proposed in #5367: Only reset the states going through the interface in question. This would cut down on unnecessary state killing.

*However*, this doesn't cover the failover scenario: If the primary WAN goes down and comes back, states on the backup WAN will not be killed. If backup WAN is metered or otherwise undesirable (e.g. cell networks), this is a problem. The *current* behavior of Dynamic state reset covers this case.

**Idea 2:** Whitelist LAN and loopback. This would keep e.g. SSH sessions into the firewall alive. But it would still unnecessarily kill primary WAN states if e.g. the backup WAN changes.

**Idea 3:** Per-interface configuration options. Thus you could set it so changes to Primary WAN1 would kill states on WAN1 and WAN2, But changes on backup WAN2 would only kill states on WAN2.

Nothing is ever easy :/ But I think idea 3 would give the best flexibility and cover all use cases.

**Describe alternatives you considered**

Someone might say "just disable State killing". That's not a solution.

**Additional context**

Failover scenario previously also discussed in #3516.

Thanks a lot!

haarp:
It could be an idea to connect state killing to gateway groups. When the active gateway moves up a tier (tier number decreases), the previous tier's states may be killed, but not the other way around.

AdSchellevis:
> Someone might say "just disable State killing". That's not a solution.

In most cases it probably is as our states aren't bound to interfaces by default and in most cases the client is unaware of the removal of the state (and hangs for that reason).

[...]

PR #6089, opened 2022-10-15
Add info that states get flushed either way to Dynamic state reset help text

haarp:
This information should be visible. From a comment by @AdSchellevis in #6088:

https://github.com/opnsense/core/blob/20a60050ee40a096ab2c86e13411b4485a89b7db/src/etc/rc.newwanip#L170-L178

Cheers! :)


fichtner:
As stated in the comment we might as well remove ip_change_kill_states switch for 23.1. I’ve never heard a case where this was deployed specifically because of its behaviour. The default seems to cover the generic use case nowadays. I think it was changed to this way back in 22.1.

haarp:
Yes, I suppose removing Dynamic state reset is reasonable. The only use case for it is failover recovery, and that can be solved differently ([like this](https://github.com/opnsense/core/pull/6089#issuecomment-1279905802))

Feel free to close.

haarp:
Obsoleted by bb9abf8.

PR #6094, opened 2022-10-17
dnsmasq: DNSSEC: Remove expired root trust anchor

johnnyslee:
See: https://data.iana.org/root-anchors/root-anchors.xml


fichtner:
So, since 2019... nice catch, thanks!

PR #6095, opened 2022-10-18
DHCP static mapping custom host options

paspo:
# the problem

I needed a way to customize DHCP static mapping, primarily for iPXE handling.
See [here](https://ipxe.org/howto/dhcpd) for examples.

# proposed solution

This PR adds a "custom" textarea in which the user can enter (non-parsed) host options. These options goes straight into `dhcpd.conf`, so I added a warning about the danger of breaking all DHCP.

### DIFF
--- a/src/etc/inc/plugins.inc.d/dhcpd.inc
+++ b/src/etc/inc/plugins.inc.d/dhcpd.inc
@@ -1176,6 +1176,12 @@ EOD;
                     }
                 }
 
+                // custom static map host config
+                if (!empty($sm['custom'])) {
+                    $dhcpdconf .= $sm['custom'];
+                    $dhcpdconf .= "\n";
+                }
+
                 $dhcpdconf .= "}\n";
             }
         }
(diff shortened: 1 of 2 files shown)


paspo:
> We don't accept pull requests for undefined custom configuration data, more info https://docs.opnsense.org/development/guidelines/basics.html#reality-overdue-maintenance

I understand that, but this guideline will force me to basically rewrite the whole DHCP module to make it fit to configd.
A module that'll probably be rewritten anyway when migrating away from ISC DHCP (see below)

> In the light of https://www.isc.org/blogs/isc-dhcp-eol/, the future of isc is unclear as well.

The future of ISC DHCP is very clear: it is EOL. It has been replaced by [kea DHCP](https://www.isc.org/kea/). And we have alternatives.

> Override directories exist: https://docs.opnsense.org/manual/dhcp.html#advanced-settings

Override directories doesn't allow to add a custom option to an existing static mapping: redefining host options is not allowed.
If I want to use an override file, I have to use it for the whole static mapping, losing the ability to edit from web interface.

fichtner:
With the override directory you can add an MVC/API plugin with the ability to manage your static mappings.

From a core perspective, however, this doesn't make any sense and you could just add the missing options with proper support.

PR #6101, opened 2022-10-21
unbound: add overview dashboard

swhite2:
Adds a reporting backend for Unbound as well as a dashboard showing an overview of interesting statistics:

- Total DNS queries (over a maximum period of 7 days)
- Blocked queries
- Cache hits
- local-data hits
- A traffic graph with variable intervals
- top N passed/blocked domains

There are a couple of hardcoded values:
- The underlying database is kept up-to-date for the last 7 days.
- The ring buffer used to decouple unbound from the logging backend has a size of 100000 entries.
- In very busy networks, db commits are done for a max of 4000 entries, which according to testing seems to be a good break-even point for DNS throughput vs. insert performance. Everything below this is flushed to the database every 5 seconds.
- To facilitate the variable interval traffic graph, multiple views with varying detail in time slots are created: 1, 5 and 30 minute buckets. These can easily be extended should different intervals be necessary.


src/opnsense/scripts/unbound/logger.py
```diff
@@ -0,0 +1,139 @@
...
+        """.format(intv=interval, min=(interval//60), lim=((60//(interval//60))*24))
+        self.cursor.execute(bucket)
+
+    def rotate_db(self, interval=7):
+        t = datetime.date.today() - datetime.timedelta(days=interval)
+        query = """
+            DELETE
+            FROM query
+            WHERE DATETIME(time, 'unixepoch') < DATETIME('{time}');
+        """.format(time=t)
```
AdSchellevis, line 72: best use bind parameters in these case, e.g. `self.cursor.execute(sql, {'prop', val})` (https://docs.python.org/3/library/sqlite3.html#sqlite3-placeholders)

src/etc/inc/plugins.inc.d/unbound.inc
```diff
@@ -369,6 +370,11 @@ function unbound_configure_do($verbose = false, $unused = '')
         }
     }
 
+    @unlink('/var/unbound/data/stats');
```
AdSchellevis, line 373: beter move to else, no need to remove it when enabled

AdSchellevis (changes requested):
Looks good, some small notes and I'll probably have to check `dnsbl_module.py` as I'm not sure what (all) the difference are between master and this version in terms of functionality.

src/opnsense/mvc/app/controllers/OPNsense/Unbound/OverviewController.php
```diff
@@ -0,0 +1,39 @@
+<?php
+
+/*
+ * Copyright (C) 2018 Deciso B.V.
```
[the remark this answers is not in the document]
swhite2: indeed it does ;)

src/etc/inc/plugins.inc.d/unbound.inc
```diff
@@ -127,7 +128,7 @@ function unbound_generate_config()
     $python_dir = "/usr/local/lib/{$pythonv}";
     $chroot_python_dir = "/var/unbound{$python_dir}";
 
-    $dirs = ['/dev', '/etc', '/lib', '/run', '/usr', '/usr/local/sbin', '/var/db', '/var/run', $python_dir];
+    $dirs = ['/dev', '/etc', '/lib', '/run', '/usr', '/usr/local/sbin', '/var/db', '/var/run', $python_dir, '/data'];
```
fichtner, line 131:
```suggestion
    $dirs = ['/data', '/dev', '/etc', '/lib', '/run', '/usr', '/usr/local/sbin', '/var/db', '/var/run', $python_dir];
```

more order ;)

src/etc/inc/plugins.inc.d/unbound.inc
```diff
@@ -368,6 +369,12 @@ function unbound_configure_do($verbose = false, $unused = '')
         }
     }
 
+    if (isset($config['unbound']['stats'])) {
+        @touch('/var/unbound/data/stats');
```
fichtner, line 373: this is for deciding to start the logger in start.sh?

swhite2: correct, but also to determine if logging should take place from the perspective of unbound

src/opnsense/scripts/unbound/start.sh
```diff
@@ -68,3 +68,7 @@ chown -R unbound:unbound /var/unbound
 if [ -n "${DOMAIN}" ]; then
 	/usr/local/opnsense/scripts/dhcp/unbound_watcher.py --domain ${DOMAIN}
 fi
+
+if [ -f /var/unbound/data/stats ]; then
```
fichtner, line 72: ok perhaps we solve this via second script parameter?

src/opnsense/service/templates/OPNsense/Unbound/core/dnsbl_module.py
```diff
@@ -11,8 +50,19 @@ def __init__(self, env):
         self.dnsbl_mtime_cache = 0
         self.dnsbl_update_time = 0
         self.dnsbl_available = False
+        self.log_update_time = time.time()
+        self.dnssec_enabled = 'validator' in self.env.cfg.module_conf
+        self.stats_enabled = os.path.exists('/data/stats')
```
fichtner, line 55: ah here we are, I see :)

swhite2: leave it as a file check in start.sh then? seems like more code than necessary if we still need the file check anyway

fichtner: yep, nothing to do here

fichtner (approved):
fWIW, works on my end after minor adjustments done yesterday

PR #6114, opened 2022-11-04
Don't sort IP address tables.  Though allow to be reversed.

NOYB:
Don't sort IP address tables.  Though allow to be reversed.

If the table is IP addresses (IPv4 or IPv6) then don't sort, but still allow to be reversed.
If the table is not IP addresses then sort natural case insensitive.

For background see: https://github.com/opnsense/core/pull/5716

### DIFF
--- a/src/opnsense/mvc/app/controllers/OPNsense/Firewall/Api/AliasUtilController.php
+++ b/src/opnsense/mvc/app/controllers/OPNsense/Firewall/Api/AliasUtilController.php
@@ -121,12 +121,18 @@ class AliasUtilController extends ApiControllerBase
             $sortcolumn = array_key_first($this->request->getPost('sort'));
             $sort_order = $this->request->getPost('sort')[$sortcolumn];
             if (!empty(array_column($formatted_full, $sortcolumn))) {
-                array_multisort(
-                    array_column($formatted_full, $sortcolumn),
-                    $sort_order == 'asc' ? SORT_ASC : SORT_DESC,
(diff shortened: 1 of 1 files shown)


AdSchellevis:
Disallowing sort wouldn't be complicated (can be set from the grid), but this will likely lead to new tickets. When there is a very good reason to avoid the standards, there should be more background about it, trusting the source is usually not the best idea.

NOYB:
Sorting makes a mess of IPv6 addresses.  Though using the source list order is most likely already correctly sorted.

An IP list source that is not already sorted, likely few, is probably in an appropriate order of some kind already.  Even so ctrl-f provides direct access to address of interest.  Probably going to do browser search anyway for large lists.

NOYB:
May not actually be trusting the third-party source.  Appears to use the table as loaded in the packet filter for the source.  "pfctl -t <table> -vT show"; which appears to be in sorted order for both IPv4 and IPv6.
So don't see need for redundant PHP sorting that messes up IPv6 order.

Original is comment is poorly worded.
FW: Diag: Aliases
Don't re-sort on IP Address column.  Reverse without re-sorting instead.

AdSchellevis:
To do this properly, we would need to have an ip[v4|v6] we simply can't trust the output of pfctl. If one would like to sort on the original order, it would likely be better to add a sequence number to the output which one could sort on then (in which case it might be sorted on addresses or not, depending on the shell tool underneath). Not going merge this at the moment.

PR #6123, opened 2022-11-06
system: wait 1 second for configd daemon

kulikov-a:
Hi!
It turned out that unfortunately there is no luck with https://github.com/opnsense/core/commit/9b0d87261d6a49f50019f151852990602fb6d6df
(still "configd socket missing.." on boot and no templates reload)
if I understand correctly, after forking, the parent process exits and, accordingly, `sleep` is not performed
https://github.com/opnsense/core/blob/0ae0c2fc9307d89c449a61605a1af50e35c54fdd/src/opnsense/service/modules/daemonize.py#L108-L110

pr moves `sleep` to daemonizer
worked for me, however I'm not sure if it's the correct way
Thanks!

### DIFF
--- a/src/opnsense/service/configd.py
+++ b/src/opnsense/service/configd.py
@@ -170,6 +170,4 @@ else:
                        logger=loghandle
                        )
     daemon.start()
-    # wait for the socket to come up before yielding to the caller
-    time.sleep(1)
 sys.exit(0)
--- a/src/opnsense/service/modules/daemonize.py
+++ b/src/opnsense/service/modules/daemonize.py
@@ -9,6 +9,7 @@ import signal
 import resource
 import logging
 import atexit
+import time
 from logging import handlers
 
 
@@ -107,6 +108,8 @@ class Daemonize(object):
                 sys.exit(1)
             elif process_id != 0:
                 # This is the parent process. Exit.
+                # Wait a second (so fork can do a deed) and exit
+                time.sleep(1)
                 sys.exit(0)
             # This is the child process. Continue.
 


mimugmail:
@AdSchellevis maybe this was also the reason last time we looked at my system.

fichtner:
@mimugmail I doubt this as templates are generated per service in either case. This is more of a cosmetic thing.

AdSchellevis:
we probably better move the "sleep" into the starter script in that case https://github.com/opnsense/core/blob/master/src/etc/rc.syshook.d/early/10-configd#L4, it's not really the daemonizers fault it shouldn't exit instantly.

fichtner:
rc.d script then would be better why this should have been closer to the daemon anyway. the strange thing is the delay was there while testing "service configd start" so I'm a bit at a loss what happened oO

kulikov-a:
@AdSchellevis 
its actualy already there - in configd_poststart (https://github.com/opnsense/core/blob/7989daa6acd3106d567b77d787c7bbeb0844379b/src/etc/rc.d/configd#L43)
but apparently the delay is not enough (works well with a delay of 2 sec on the test vm)
it just seemed to me that @fichtner  wanted to see it in the daemonizer
https://forum.opnsense.org/index.php?topic=30197.msg145862#msg145862

 so I'm a bit at a loss too ))

fichtner:
@kulikov-a how about 4710ec0cdc64dc then?

fichtner:
sorry, typos...

fichtner:
# opnsense-patch 4710ec0cdc64 b69696c5fb

kulikov-a:
@fichtner tested and works great. thanks!

PR #6131, opened 2022-11-10
Create actions_curl.conf

Choff3:
Added a cURL action. This is useful for scheduling cURL cron jobs for pinging uptime services such as Uptime Kuma, UptimeRobot or HealthChecks.io.

### DIFF
--- a/src/opnsense/service/conf/actions.d/actions_curl.conf
+++ b/src/opnsense/service/conf/actions.d/actions_curl.conf
@@ -1,0 +1,5 @@
+[start]
+command: curl
+parameters: %s
+type: script
+description: Curl


fichtner:
Structurally this has no overarching purpose and we would encourage to keep this as a local customisation as otherwise we would not be able to get rid of it ever again even if for security reasons. 😉


Cheers,
Franco

Choff3:
Makes sense. I would love to be able to make custom cron jobs from the web GUI.

AdSchellevis:
Both are not going to happen, shell injections are against our guidelines (https://docs.opnsense.org/development/guidelines/basics.html#safeguard-user-input) and a cause of privilege escalations.

PR #6145, opened 2022-11-19
Update base tunings to bring it inline with other Router projects

failedrequest:
The following tunings are bumpped in src/etc/inc/system.inc . This is sourced from BSDRP, FreeBSD.org and pfsense.

  - kern.ipc.nmbclusters="1000000"
  - kern.ipc.nmbjumbop="524288"
  - kern.ipc.nmbjumbo9="524288"
  - net.link.ifqmaxlen="128"

### DIFF
--- a/src/etc/inc/system.inc
+++ b/src/etc/inc/system.inc
@@ -75,6 +75,9 @@ function system_sysctl_defaults()
         'hw.syscons.kbd_reboot' => [ 'default' => '0' ],
         'hw.uart.console' => [ 'default' => 'io:0x3f8,br:' . system_console_speed(), 'type' => 't' ], /* XXX support comconsole_port if needed */
         'kern.ipc.maxsockbuf' => [ 'default' => '4262144' ],
+        'kern.ipc.nmbclusters' => [ 'default' => '1000000' ],
+        'kern.ipc.nmbjumbop' => [ 'default' => '524288' ],
+        'kern.ipc.nmbjumbo9' => [ 'default' => '524288' ],
         'kern.randompid' => [ 'default' => '1' ],
         'net.enc.in.ipsec_bpf_mask' => [ 'default' => '2', 'required' => true ], /* after processing */
         'net.enc.in.ipsec_filter_mask' => [ 'default' => '2', 'required' => true ], /* after processing */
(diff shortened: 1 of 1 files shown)


failedrequest:
Ad this pull request is in regards to https://forum.opnsense.org/index.php?topic=30841.msg148741#msg148741 . Let me know if this is ok. This is stuff that I have been adjusting out of the box , none of this major changes but it will improve this use case and a number of others out of the box ,

AdSchellevis:
Hi Mark, I think it's better to leave (most of) this in the docs to be honest, best performant settings relate to the use-case and the equipment at hand, for example `nmbclusters` by default is a calculated value which scales to the amount of memory (and probably can exceed the proposed limit as well).

https://github.com/freebsd/freebsd-src/blob/440217b0af5ffa08086829f5902b01a92e7a6790/sys/kern/kern_mbuf.c#L155-L205

Some network cards advise very different settings than standard, but usually it's not a one size fits all in my experience. 

The `net.link.ifqmaxlen` setting I'm not entirely sure about,  the best practice seems to be calculated from the total amount of tx rings, but that's not something we can do easily.  I couldn't find very scientific documentation about it to be honest, a lot of people fiddle with settings, which is ok, but doesn't automatically warrant new defaults in my opinion. In this case a higher value might be a good idea, but what it should be, I don't know.

Our documentation spot for these kind of settings is https://docs.opnsense.org/troubleshooting/performance.html

Best regards,

Ad

failedrequest:
Ad
  I was not aware of how each NIC driver used ifqmaxlen . Looking at some of the drivers I can see where setting this to large will cause issues, forget about that . 

As for kern.ipc.nmbclusters setting this to 1M has been a starting point for me , for a long time. Considering this is a amd64 only release, I doubt there are many install with low ram where this would be an issue.  

Also since kern.ipc.nmbjumbop and kern.ipc.nmbjumbo9 are based off what the value for kern.ipc.nmbclusters is. I would ask you to consider if just bumping kern.ipc.nmbclusters .

AdSchellevis:
Mark,

I don't think we should manually try to set these values as argumented in my previous statement, looking at some of our own machines, `kern.ipc.nmbclusters` already has a value over 1M on machines offering 16GB of memory using the calculated kernel defaults.  If documentation is missing for some (specific) type of setups, our documentation is the better place to explain other defaults, forcing a value which was calculated by default only "helps" small installs at the price of larger installs not offering what they used to have.

Best regards,

Ad

PR #6153, opened 2022-11-23
certs: typo in is_crl_internal

kulikov-a:
Hi!
ref. https://forum.opnsense.org/index.php?topic=31109.0
(external (imported) CRLs cannot be viewed\edited)
since comparision have highest precedence then coalescing, `is_crl_internal` returns string if array member exists.
so all CRLs starts treated as internal
Thanks!


kulikov-a:
Thanks!!)

PR #6166, opened 2022-12-02
tokenize2: unicode content in CSS

kulikov-a:
Hi!
not a big deal but can annoy sometimes)
after some work in dev console, chrome may return tokenized field content with a garbage like

css

pr uses escaped unicode number

thanks!

### DIFF
--- a/src/opnsense/www/css/tokenize2.css
+++ b/src/opnsense/www/css/tokenize2.css
@@ -120,7 +120,7 @@
 }
 
 .tokenize > .tokens-container > .token > .dismiss:after {
-    content: "×";
+    content: "\00D7";
 }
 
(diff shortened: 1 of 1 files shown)


AdSchellevis:
@kulikov-a thanks for the fix.

Thought to check upstream first (https://github.com/dragonofmercy/Tokenize2), but apparently that was deprecated in favour of https://tom-select.js.org/ . The JavaScript library collection remains very fluent unfortunately, maybe in time we should try to upgrade to the next iteration if it's better.

kulikov-a:
Thanks! also first checked the upstream, but did not pay attention to the successor. would be interesting to look at it, thanks

PR #6172, opened 2022-12-05
webgui: enable compression

kulikov-a:
Hi!
Noticed that the gui returns the content uncompressed, although the `mod_deflate` module is enabled.
I think lighty expects `deflate.allowed-encodings` to select a method from a match with the browser request.
after adding the option, content is returned gziped.

perhaps building lighty with `--with-brotli` would be good  but I did not find it in the ports..

ref. https://redmine.lighttpd.net/projects/lighttpd/wiki/Mod_deflate

thanks!

ps. I'm not sure about the need to add a cron for old\stale cache entries purge.

### DIFF
--- a/src/etc/inc/plugins.inc.d/webgui.inc
+++ b/src/etc/inc/plugins.inc.d/webgui.inc
@@ -446,9 +446,10 @@ debug.log-response-header  = "disable"
 debug.log-request-handling = "disable"
 debug.log-file-not-found   = "disable"
 
-# gzip compression
+# compression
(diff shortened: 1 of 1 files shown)


gstrauss:
Depending on the characteristics of the site, one option is to prune or clear the cache each time lighttpd is started, or to do it on shutdown.

Unless the content being returned is larger and mostly static (not frequently changed), the difference between brotli and gzip is often small.  YMMV.

kulikov-a:
thanks!

PR #6173, opened 2022-12-05
unbound: always use python first

kulikov-a:
Hi
ref. https://forum.opnsense.org/index.php?topic=31336.0
It turned out that when the DNS64 module is enabled, the blocklists do not work. Looks like the python module always has to be first for blocklists to work
thanks!


kulikov-a:
Thanks!)

PR #6177, opened 2022-12-06
Firewall: Log Files: Live View - improve table slicing

kulikov-a:
Hi!
ref. https://forum.opnsense.org/index.php?topic=31360.0
it turned out that a simplified table cleanup can leave hidden rows and thus cut off those that match the filter ahead of time
pr removes hidden rows first.
to speed up DOM work: class added for all rows, `gt` migrated to `slice` [(jquery docs)](https://api.jquery.com/gt-selector/#gt1)

thanks!

upd. `last_digest` handling updated (no dependency on filter matches)
upd2. deffering log polling added


AdSchellevis:
discussed internally, we'll keep this open in case more reports come in. All modifications in this area have downsides (removing searchable data within current slice v.s. keeping old results visible for a longer time), not sure if this is worth the effort at the moment.

kulikov-a:
yes, sure!
If I understand correctly, it all comes down to the choice of logic: show records that match filters among the latest `max_rows`  or show` max_rows` of the last records that match the filters.
user expectations may vary

src/opnsense/mvc/app/views/OPNsense/Diagnostics/fw_log.volt
```diff
@@ -313,7 +321,8 @@
                     apply_filter();
 
                     // limit output, try to keep max X records on screen.
-                    $("#grid-log > tbody > tr:gt("+max_rows+")").remove();
+                    $(".log_tr").filter(":hidden").remove();
+                    $(".log_tr").slice(max_rows).remove();
```
AdSchellevis, line 325: @kulikov-a if you are going to remove all outside of visible range, you might need to reset the grid on filter changes as well. Not sure if we are going to implement this, but I certainly like the `$.Deferred()` in `fetch_log()`. The `grid_reset` might be avoided by  setting `last_digest` to `false` by the way, saves a variable which more or less has the same function anyway.

kulikov-a:
@AdSchellevis , hi! thanks for the review. made changes.
If you have time and mood, could you please comment a little on the decision to change the logic of displaying the results? I expressed my concerns at https://forum.opnsense.org/index.php?topic=31568.msg152388#msg152388 (i give the link so as not to chatter here as well). not so much trying to convince as to understand )
Thanks!

AdSchellevis:
@kulikov-a The most important reason to change the previous logic was to prevent gaps when the number of records per time interval increased, leading to the situation where the live view couldn't keep up. So it's both performance and functionality. The amount of feedback so far has been quite low, which is also the reason why we're a bit cautious in changing it again, which doesn't mean we won't change it, but it might take some time.

kulikov-a:
@AdSchellevis thanks for your time!
became a little clearer )
I agree that so far there is not enough feedback for this..
just in case, made one more change: the filter is applied only to a fresh data chunk (and only as long as it makes sense).

AdSchellevis:
@kulikov-a we'll just keep this one open for now, maybe merge it later in some state. Thanks for working on it anyway

AdSchellevis:
with https://github.com/opnsense/core/commit/854043daa1e1565f40b938f1dd9da2c03a1381f3 the situation seems to be stable, let's close this now.

PR #6182, opened 2022-12-07
unbound / overview: migrate to duckdb

swhite2:
Moves the reporting backend from sqlite to duckdb.

Since, unlike sqlite, duckdb does not support concurrency, a wrapper has been written to synchronize access to the database. Refer to the class itself for further documentation.

If you were on the master branch and had unbound statistics enabled prior to this commit, you should `rm -f /var/unbound/data/unbound.sqlite` and restart Unbound.

### DIFF
--- a/src/www/services_unbound.php
+++ b/src/www/services_unbound.php
@@ -40,7 +40,6 @@ if ($_SERVER['REQUEST_METHOD'] === 'GET') {
     $pconfig = array();
     // boolean values
     $pconfig['enable'] = isset($a_unboundcfg['enable']);
-    $pconfig['stats'] = isset($a_unboundcfg['stats']);
     $pconfig['enable_wpad'] = isset($a_unboundcfg['enable_wpad']);
     $pconfig['dnssec'] = isset($a_unboundcfg['dnssec']);
     $pconfig['dns64'] = isset($a_unboundcfg['dns64']);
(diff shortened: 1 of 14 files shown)


swhite2:
> I'm missing package dependencies here...

That's the main reason this PR is still in draft state.

> also what other python do we directly depend on? pandas and numpy or more?

If we're talking direct dependencies, only duckdb, pandas and numpy.

src/www/services_unbound.php
```diff
@@ -40,7 +40,6 @@
     $pconfig = array();
     // boolean values
     $pconfig['enable'] = isset($a_unboundcfg['enable']);
-    $pconfig['stats'] = isset($a_unboundcfg['stats']);
```
fichtner, line 43: me like :)

fichtner (approved):
looks good on paper, can give it a try tomorrow

swhite2:
> @swhite2 added deps in [afa13df](https://github.com/opnsense/core/commit/afa13df34d30a4fe56a9996a1cfae7f39f14f5d6) 

Thanks!

> -- looking at it "DNS" might be too vague so I'd suggest "Unbound DNS" as reporting entry. Also the Unbound control 
> widget should probably removed if we have this anchored in reporting section. People are probably confused what it will do > (reload stats? reload stats collector? but not unbound itself).

Agreed, within the same context I named it "Unbound DNS" in the reporting settings toggle for the same reason, will change it accordingly for the menu item as well. Thanks for the feedback :)

fichtner:
PS: works fine from here, nice work!

PR #6187, opened 2022-12-11
VPN/IPsec add new MVC module

AdSchellevis:
Add new component to manage IPsec connections in a similar format as `swanctl.conf` is defined (https://docs.strongswan.org/docs/5.9/swanctl/swanctlConf.html).  As this needs to work in conjunction with the legacy IPsec module, some minor changes are needed to the current state. 

o VPN/IPsec/Pre-Shared Keys - add optional remote identifier (merges in `ipsec.inc`)
o VPN/IPsec/Virtual Tunnel Interfaces - new component to show existing VTI's and add new ones (as these are separate entities)
o VPN/IPsec/Connections [new] - configuration tool to build `swanctl.conf` 
o Integrate MVC generated `swanctl.conf` into `ipsec.inc` (legacy overlays)
o Integrate manually configured VTI's into `ipsec.inc` (`array_merge(ipsec_get_configured_vtis(), (new \OPNsense\IPsec\Swanctl())->getVtiDevices())`)
o fix minor php warning when changing reqid's (`$local|remote_configured` initialisation when `$configured_intf[$intf]` not found)

### DIFF
--- a/src/etc/inc/plugins.inc.d/ipsec.inc
+++ b/src/etc/inc/plugins.inc.d/ipsec.inc
@@ -176,7 +176,8 @@ function ipsec_interfaces()
             }
         }
 
-        foreach (ipsec_get_configured_vtis() as $intf => $details) {
+        $vtis = array_merge(ipsec_get_configured_vtis(), (new \OPNsense\IPsec\Swanctl())->getVtiDevices());
+        foreach ($vtis as $intf => $details) {
             $interfaces[$intf] = [
                 'enable' => true,
                 'descr' => preg_replace('/[^a-z_0-9]/i', '', $details['descr']),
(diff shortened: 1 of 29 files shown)


ansibleguy:
Greetings.

As I'm curious - will those IPSec changes/updates enable us to API-manage tunnels (_create/update/delete P1+P2_) in the future?

Thanks (:

\- AnsibleGuy

fichtner:
Hello @ansibleguy,

Yes it will. Biggest caveat is that we won’t migrate legacy data to new format (modelled around swanctl.conf) so with 23.1 you have a full API but existing P1/P2 can’t be edited.

The plan forward is not fully decided. We might just remove the legacy pages in 2024. Whether an automatic migration makes sense depends on user feedback and funding in 2023. But we would much rather keep working on API additions for other lacking components.


Cheers,
Franco

ansibleguy:
Awesome! :D

Thanks for the information.
I'm looking forward to implementing it as Ansible module.

\-AnsibleGuy

AdSchellevis:
@fichtner and initial usage https://github.com/opnsense/core/commit/12363f5f3cf9e3d1b4bc813fbc7687c7c829972d

ansibleguy:
> > I'm looking forward to implementing it as Ansible module.
> 
> That was my secret hope :D Thanks for your work!

FYI: The Ansible modules for IPSec are available => [Docs](https://opnsense.ansibleguy.net/en/latest/modules/ipsec.html)

\- AnsibleGuy

PR #6194, opened 2022-12-14
Add base-uri and report-sample csp directives

L1ghtn1ng:
The base-uri directive prevents base elements from injected as a way to bypass the csp. Also adding report-sample which helps with debugging what is being blocked with the CSP so can be easier fixed

### DIFF
--- a/src/opnsense/mvc/app/controllers/OPNsense/Base/ControllerBase.php
+++ b/src/opnsense/mvc/app/controllers/OPNsense/Base/ControllerBase.php
@@ -282,6 +282,7 @@ class ControllerBase extends ControllerRoot
         // set security policies
         $policies = array(
             "default-src" => "'self'",
+            "base-uri" => "'self'",
             "img-src" => "'self'",
             "script-src" => "'self' 'unsafe-inline' 'unsafe-eval'",
             "style-src" => "'self' 'unsafe-inline' 'unsafe-eval'");
--- a/src/www/guiconfig.inc
+++ b/src/www/guiconfig.inc
@@ -41,7 +41,7 @@ require_once("config.inc");
 /* CSRF BEGIN: CHECK MUST BE EXECUTED FIRST; NO EXCEPTIONS */
 require_once('csrf.inc');
 // hardening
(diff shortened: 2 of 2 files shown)


AdSchellevis:
@L1ghtn1ng I suppose this PR relates to https://github.com/opnsense/core/issues/6195 or am I missing something?

L1ghtn1ng:
Yes and no, the report-sample will help with fixing things for sure as it will give better error messages in the browser console. But this PR is just a recommended addition and a slight tidy up.

On 14 Dec 2022, at 08:28, Ad Schellevis ***@***.***> wrote:

﻿

@L1ghtn1ng<https://github.com/L1ghtn1ng> I suppose this PR relates to #6195<https://github.com/opnsense/core/issues/6195> or am I missing something?

AdSchellevis:
I'll put it on my list of things todo to review and test this one.

L1ghtn1ng:
Hi @AdSchellevis any updates on this?

AdSchellevis:
not really, too busy with other work and not enough context on why we should do this.

L1ghtn1ng:
In a nutshell, stops a csp bypass vector using the base-uri include and the report-sample is useful for seeing what code violates a csp directive so helps with fixing any csp issues

For more information please see https://book.hacktricks.xyz/pentesting-web/content-security-policy-csp-bypass

On 5 Sep 2023, at 15:19, Ad Schellevis ***@***.***> wrote:

﻿

not really, too busy with other work and not enough context on why we should do this.

AdSchellevis:
I don't mind improving the situation, but someone needs to test this thoroughly, the modification should be consistent between legacy and mvc parts and should follow best practices (I can already see the next PR coming in to remove `unsafe-inline` for example https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Security-Policy/base-uri). 

This needs work to do it right, which unfortunately isn't high enough on our priority list. If you're willing to finish it and document test results, we can reevaluate if we should move it up the list. In our experience changes like these have a tendency to haunt us later.

PR #6197, opened 2022-12-17
unbound: change working dir before unbound-checkconf on start

kulikov-a:
Hi!
ref. https://forum.opnsense.org/index.php?topic=31475.0 and https://github.com/opnsense/core/commit/13ea70af049dabcec3f7fe907240d3a8aaeef4dc
just another place where the unbound-checkconf messes things up
( false-error triggers trust anchor update on every unbound (re)start)

Thanks!

### DIFF
--- a/src/opnsense/scripts/unbound/start.sh
+++ b/src/opnsense/scripts/unbound/start.sh
@@ -35,6 +35,7 @@ for FILE in $(find /var/unbound/etc -depth 1); do
 done
 
 # if the root.key file is missing or damaged, run unbound-anchor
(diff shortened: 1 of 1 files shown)


fichtner:
Merged, thanks!

fichtner:
In ref to forum discussion: the problem in the setup is still there, just not that frequent anymore.

kulikov-a:
@fichtner thanks! to admit, I didn’t fully understand what was meant by three launches)
for me the patch works on the test vm. restarted several times - roots never poked after the fix

PR #6205, opened 2022-12-21
Unbound / Blocklists: add exact domain blocking and integrate into overview page

swhite2:
This includes the ability to whitelist a domain from the overview page as well. Relevant to both
the top passed/blocked domains, as well as the detailed query grid.

blocklists.py has been modified in such a way that it will detect whether it needs to
start the download process or simply administrate locally. The latter currently only happens
when custom domains for blocking have been added/removed by a user. The reasoning is
that we can easily extend/shrink the current blocklist when it comes to blocking exact
domains as this is handled on the incoming side. However, while we can modify the current
list to accomodate a new whitelist entry (which can be regex), we (currently) cannot know which domains
were skipped in the process of retrieving them in the first place if a user explicitly removes
a whitelist entry. Therefore we decide to re-run the download on a whitelist action.

furthermore, the updateBlocklistAction in the controller administrates how the model is updated
(e.g. when a blocked item is whitelisted, it should be removed from the blocklist model entry and
added to the whitelist)

In the future we could optimize the whole process by checking if a remote file has changed in
date or size.

### DIFF
--- a/src/opnsense/mvc/app/controllers/OPNsense/Unbound/Api/SettingsController.php
+++ b/src/opnsense/mvc/app/controllers/OPNsense/Unbound/Api/SettingsController.php
@@ -40,6 +40,58 @@ class SettingsController extends ApiMutableModelControllerBase
 
     private $type = 'dot';
 
+    public function updateBlocklistAction()
+    {
+        $result = ["status" => "failed"];
+        if ($this->request->isPost() && $this->request->hasPost('domain') && $this->request->hasPost('type')) {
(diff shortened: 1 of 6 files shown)


fichtner:
Sorry browser crashes this morning while writing a comment. Does this already enable regex input? If yes we should consider some form of regex field type for inclusion in 23.1.x and/or improve error reporting on faulty input while running the python module?

fichtner:
PS: happy new year \o/

PR #6213, opened 2022-12-25
Monit starttimeout

spoutin:
The PR is to add the ability via the web to configure Service Start Timeout.

This could be extended to include Service Stop Timeout as well but I wanted to see if this was going to be accepted first.

This is my first PR so take it easy, I willing to accept any pointers to feedback.

Thanks

### DIFF
--- a/src/opnsense/service/templates/OPNsense/Monit/monitrc
+++ b/src/opnsense/service/templates/OPNsense/Monit/monitrc
@@ -135,7 +135,11 @@ check {{ service.type }} {{ service.name }} {{ path }}
 {%      endfor %}
 {%     endif %}
 {%     if service.start|default('') != '' %}
-   start program = "{{ service.start }}"
+{%       set srv_start_timeout = '' %}
+{%       if service.starttimeout|default('') != '' %}
+{%         set srv_start_timeout = "with timeout " ~ service.starttimeout ~ " seconds" %}
+{%       endif %}
+   start program = "{{ service.start }}" {{ srv_start_timeout }}
 {%     endif %}
 {%     if service.stop|default('') != '' %}
    stop program = "{{ service.stop }}"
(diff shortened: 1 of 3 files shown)


src/opnsense/service/templates/OPNsense/Monit/monitrc
```diff
@@ -135,7 +135,11 @@ check {{ service.type }} {{ service.name }} {{ path }}
 {%      endfor %}
 {%     endif %}
 {%     if service.start|default('') != '' %}
-   start program = "{{ service.start }}"
+{%       set srv_start_timeout = '' %}
+{%       if service.starttimeout|default('') != '' %}
+{%         set srv_start_timeout = "with timeout " ~ service.starttimeout ~ " seconds" %}
+{%       endif %}
+   start program = "{{ service.start }}" {{ srv_start_timeout }}
```
AdSchellevis, line 142: If we are always going to set the timeout, it might be better to replace the above with:

```
start program = "{{ service.start }}" with timeout {{service.starttimeout|default('30')}} seconds
```

As timeout is 30 seconds by default anyway, we should be safe here.

AdSchellevis:
@spoutin one small remark, but other than that this looks ok. If you adjust the PR, I'll merge it.

src/opnsense/service/templates/OPNsense/Monit/monitrc
spoutin: I've made your recommended change. 
I hesitated doing it the way you recommended because it would hard code Monit's default value in Opnsense's source code and if it were to change, it would also need to change in Opnsense.  On the other hand, its less code which is nice.

Thanks for the feedback.

AdSchellevis: @spoutin setting the default in the model would lead to the same effect as it persists the "30" after a save. In this case I don't think it matters that much to force it to default.

AdSchellevis:
@spoutin thanks!

PR #6225, opened 2023-01-03
Name update in system_advanced_firewall description text

SZim92:


fichtner:
Hi, „pf“ is the name of the firewall component in FreeBSD being used. I don’t want to change that and invalidate the translation to embed a project name here with no added value for the reader. 😉

cheers,
Franco

SZim92:
Apologies, my mistake. I misparsed the sentence as referring to the overall project.

Issue #3903, opened 2020-02-02
[6rd][stf] default route not setup automatically


RyuunoAelia:
**Describe the bug**

When setting up the 6rd config, the default inet6 route is not set. Setting it up manually works.

```
# ifconfig
[...]
wan_stf: flags=4041<UP,RUNNING,LINK2> metric 0 mtu 1480
	inet6 2a02:120X:XXXX:XXX0:: prefixlen 60 
	nd6 options=1<PERFORMNUD>
	v4net X.X.X.X/0 -> tv4br 193.5.29.1
	groups: stf 
```

```
netstat -rn
[...]
Internet6:
Destination                       Gateway                       Flags     Netif Expire
::1                               link#3                        UH          lo0
2a02:120X:XXXX:XXX0::              link#9                        UHS         lo0
2a02:120X:XXXX:XXX0::/60           link#9                        U       wan_stf
2a02:120X:XXXX:XXX1::/64           link#1                        U           re0
2a02:120X:XXXX:XXX1::1             link#1                        UHS         lo0
fdb9:86fc:f74d:1945::             link#1                        UHS         lo0
fdb9:86fc:f74d:1945::/64          link#1                        U           re0
```

If I setup the correct route using:
```
# ping6 google.com
ping6: UDP connect: No route to host
# route -6 add default 2a02:120X:XXXX:XXX0::c105:1d01
add net default: gateway 2a02:120X:XXXX:XXX0::c105:1d01
# ping6 google.com
PING6(56=40+8+8 bytes) 2a02:120X:XXXX:XXX0:: --> 2a00:1450:400a:802::200e
16 bytes from 2a00:1450:400a:802::200e, icmp_seq=0 hlim=56 time=5.517 ms
[...]
```
c105:1d01 being the IP address of the border gateway 193.5.29.1 in hex (for those who have the same issue, and are wondering how I worked around the issue).

**To Reproduce**
1. Setup a working 6rd config
2. ping6 google.com
3. get no route to host error

Note: I am using NPTv6 and a local ipv6 address as Virtual IP on the internal network side. Though if I setup the route manually I get full connectivity without issue.

**Expected behavior**
6rd setup should set the default route instead of having to set it up manually.

**Environment**
OPNsense 20.1 (amd64, OpenSSL).

RyuunoAelia:
[...]

PR #6227, opened 2023-01-05
Fixed route-to interface not being set correctly in firewall rules when using 6rd/6to4 IPv6 gateways with gateway groups.

fjelliott:
Hello. First-time contributor here.

I've been battling with an annoying problem on my OPNsense box (great software, btw). So, I have a dual-WAN setup and one of the WANs uses 6rd for IPv6. I followed the guide to set up a gateway group and then configured the right firewall rules. However, OPNsense was not routing IPv6 traffic at all. I was annoyed by this, so I spent a day debugging the issue and coding up a fix (which is this pull request). This problem ONLY occurs when you use a gateway group that contains a gateway that uses either 6rd or 6to4 for IPv6 connectivity. Here is what I discovered:

## Root Cause
The root cause ended up being that OPNsense wasn't correctly writing the firewall rule to set the gateway policy for traffic entering the gateway. To be more precise, instead of using the correct `wan_stf` interface (which in my case is called `opt1_stf`), it would use the interface for the underlying IPv4 gateway (which in my case is `pppoe0`). This would result in the gateway trying to send un-encapsulated IPv6 packets on the underlying IPv4 WAN interface instead of first adding 6rd or 6to4 headers to the packets. Since my ISP doesn't do native IPv6, this is roughly equivalent to sending all IPv6 packets to `/dev/null`. I was able to discover this by inspecting a packet capture of my `pppoe0` interface and pinging an IP address that isn't normally queried by my network (to make filtering the packet capture easier). When I set up my firewall rules to use the default gateway, IPv6 packets would be correctly encapsulated in a 6rd-compliant IPv4 header. When I set up the firewall rules to point to my gateway group, I would see raw IPv6 packets sent to `pppoe0`.

After this, I inspected the firewall rules that OPNsense was actually producing, and noticed that when I pointed it at my gateway group, it issued these entries (perhaps I changed a couple things because privacy):

```
pass in quick on vlan0.2 route-to {( pppoe0 2608::207.171.2.60 )} inet6 from {any} to {any} keep state label "81498130b9824e4f784fe38058fcbcfa" # : IPv6 Gateway Policy
pass in quick on lagg1 route-to {( pppoe0 2608::207.171.2.60 )} inet6 from {any} to {any} keep state label "81498130b9824e4f784fe38058fcbcfa" # : IPv6 Gateway Policy
```

Oh yeah, vlan0.2 is my guest network and lagg1 is my primary LAN.

You'll notice that it's routing packets to `pppoe0`, where they're not supposed to go. This is the root cause of the problems.

## Fix
As you might guess, the fix is getting OPNsense to issue the correct rule for 6rd and 6to4 setups. As it turns out it already does this correctly for dynamic gateways (which is why the issue is only present for gateway groups). You can see this a little further down in `getGateways()` on line 229. So I simply copied this code up above to where it handles statically-defined gateways and it fixed my problems. I refreshed my firewall rules and all of a sudden I had IPv6 connectivity!

Of course, that's probably not enough proof for you. So I took copies of /tmp/rules.debug before and after I applied my fix and ran them through `diff` to see if anything else got changed. The only stuff that got changed was what I wanted to be changed. Attached are the before and after `/tmp/rules.debug` files for you to inspect. I may or may not have fiddled with some IP addresses (again, because privacy). I also had to rename them because it seems GitHub is picky about file extensions.
[rules-before.txt](https://github.com/opnsense/core/files/10349881/rules-before.txt)
[rules-after.txt](https://github.com/opnsense/core/files/10349882/rules-after.txt)

## Testing
While I have tested that it fixes my problem, I haven't tested to make sure my change doesn't break stuff elsewhere. Although, it seems like a fairly safe change considering that this exact same change is already done for dynamic gateways in the same function.

### DIFF
--- a/src/opnsense/mvc/app/library/OPNsense/Routing/Gateways.php
+++ b/src/opnsense/mvc/app/library/OPNsense/Routing/Gateways.php
@@ -195,7 +195,10 @@ class Gateways
                             // default address family
                             $gw_arr['ipprotocol'] = 'inet';
                         }
-                        $gw_arr["if"] = $definedIntf[$gw_arr["interface"]]['if'];
+                        // Make sure to use the tunnel interface for 6rd and 6to4 setups.
+                        $ifname = $gw_arr["interface"];
+                        $ifcfg = $definedIntf[$ifname];
+                        $gw_arr["if"] = $gw_arr['ipprotocol'] == 'inet6' && in_array($ifcfg['ipaddrv6'] ?? null, ['6to4', '6rd']) ? "{$ifname}_stf" : $ifcfg['if'];
                         $gw_arr["attribute"] = $i++;
                         if (Util::isIpAddress($gateway->gateway)) {
                             if (empty($gw_arr['monitor_disable']) && empty($gw_arr['monitor'])) {


fjelliott:
I think this fixes #3903 and probably a few others.

AdSchellevis:
The whole `stf(4)` situation is unfortunately a design flaw from the past, it highly likely should have been a separate interface like `gif(4)` is in which case you could actually select the right target and the configuration logic would match reality. 

At a first glance it looks ok to align the interface magic with the dynamic selection for consistency reasons, a proper fix for this whole mess is likely difficult to achieve. 

Thanks for creating taking the time to create a PR and an elaborate explanation of the issue , I'll try to review and merge this soon.

fichtner:
Looks reasonable. Additionally I think that „Use IPv4 connectivity“ also plays tricks on IPv6 gateway interface selection present here. I’ll have to investigate next week.

As for #3903 the noise has subsided since 2021 so I suspect it was already fixed (or part of it). Vehement reports are often abandoned without a trace once fixes arrive because reporters cannot pinpoint/reproduce steps correctly.


Cheers,
Franco

fichtner:
PS: as expected 21.1 shipped an improvement, see https://github.com/opnsense/changelog/blob/a9bd70fb5ede27345b9ecae9e3d4e7f78b82b33f/community/21.1/21.1#L60

fichtner:
Merged, thanks! I think I will change this further and add a get_real_interface() equivalent in MVC to make sure the lookup goes according to what the system actually expects (for better or worse).


Cheers,
Franco

fichtner:
@fjelliott I reworked the state of the gateway code to always deliver the correct interface (according to existing code elsewhere) in https://github.com/opnsense/core/commit/5b7575b9d1d5

Would you mind trying this one? It requires to scrub the manual changes on your end with `opnsense-revert` and should work on the latest 22.7.x releases:

    # opnsense-revert opnsense
    # opnsense-patch 5b7575b9d1d5


Cheers,
Franco

fjelliott:
I'll try it out when I get home from work.

fjelliott:
Yes, that patch does seem to emit the correct firewall rules. I applied it and it doesn't seem to have broken anything, either. Quick question: do the effects of the `opnsense-patch` command persist across system updates? As in, if I download an update for OPNsense will opnsense-patch re-apply itself?

fichtner:
Thanks for testing! This will got into RC2 for sure. ❤️

Updates will revert patch states always. The patch is cached on the disk for later use, but not applied automatically anymore.

Two examples:

1. 22.1.8 + patch -> 22.1.9 does not offer the patch as a change so the patch will be gone unless reapplied.
2. 22.1.8 + patch -> 22.1.10 offers included patch so no manual patching necessary.

One caveat is that once 22.1.10 is patched running opnsense-patch again will *revert* the patch. That's a "feature" of the patch(1) utility to be able to reverse patches. In that sense running opnsense-patch on the same twice will apply and remove the patch (or remove it and reapply).

It's a handy little tool ;)


Cheers,
Franco

Issue #3574, opened 2019-07-16
python3: remove subprocess.call in favour of subprocess.run


AdSchellevis:
https://docs.python.org/3/library/subprocess.html#subprocess.run

ErikJStaab:
Is this going to be merged in any time soon? I have been noticing with update_tables.py running every 1 minute it reads in all of the aliases and compares them which can take a while depending on how long your url tables/geo ip are. I am curious if this is faster will it make it a non issue? (Currently with 250k lines or of aliases so it takes about 50% of 1 core on my router for 5s every minute) It does seem to affect the threading somewhat since it is a good size load to be running off/on every minute.

As an alternative I am wondering if there is a reason it needs to read/compare loaded pf tables against for the file tables url/urltable/geoip types every time the script runs? Have there been consistency issues in the past?

fichtner:
Is this done or is there still work for plugins to do? :)

AdSchellevis:
I think there's one left in configd and some in the plugins, the plugins should be easy to replace, configd I rather leave for now. Let me look at the plugins and close the issue when done.

PR #6229, opened 2023-01-06
configd: capture scripts outputs

kulikov-a:
Hi!
For now if the configd action fails, the output of the command can only be seen in the shell.
pr tries to help with this with a migration from `subrocess.call` to `subprocess.run` so that we can grab the commands output.
This can help, for example, with services startup errors debugging (configuration errors, socket conflicts, etc.) when such errors do not get into the plugin log.
configd

(as a bonus, get rid of pgrep false-errors like https://forum.opnsense.org/index.php?topic=31079.0).

catch 'parameter mismatch' for inline actions also (ref. https://forum.opnsense.org/index.php?topic=30908.msg149703#msg149703)

Thanks!
ps. it seems that it was even in the plans, but for some reason did not get around to it? ) (https://github.com/opnsense/core/issues/3574#issuecomment-539113436)

### DIFF
--- a/src/opnsense/service/modules/processhandler.py
+++ b/src/opnsense/service/modules/processhandler.py
@@ -470,6 +470,11 @@ class Action(object):
                         script_output = output_stream.read()
                         script_error_output = error_stream.read()
                         if result.returncode == 0:
+                            if len(script_error_output) > 0 and self.type.lower() == 'script_output' and self.command.endswith('exit 0') and not self.command.startswith('pgrep'):
+                                # backward compatibility with subprocess.check_call version: exit 0 forced for script_output action and stderr is not empty. log stderr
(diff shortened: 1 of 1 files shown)


kulikov-a:
not relevant to the code in master branch anymore obviously )
will try to play with updated configd first

AdSchellevis:
@kulikov-a I've only moved it around to make room, error logging could still be improved

kulikov-a:
Hi, Ad!
yes, I'm just afraid to do it without testing. Will make a new pr when I test it and if it still be relevant, if you don't mind )

AdSchellevis:
Hi Alex,

If you want to take a stab at it, go ahead, I think we can simplify this even more by merging some functionality, but there's no rush, I do agree logging could be better here (shouldn't be too difficult).

Best regards,

Ad

PR #6248, opened 2023-01-17
unbound: log startup errors

kulikov-a:
Hi!
there still seems to be an issue with the lack of diagnostic messages if an unbound  error occurs before logging is switched to syslog. (ref. https://forum.opnsense.org/index.php?topic=31865.0)

pr passes the output of `/usr/local/sbin/unbound` to the syslog if it fails and aborts script execution.

unbound_err

thanks!

### DIFF
--- a/src/opnsense/scripts/unbound/start.sh
+++ b/src/opnsense/scripts/unbound/start.sh
@@ -63,7 +63,12 @@ rm -f /usr/local/etc/unbound.opnsense.d/dnsbl.conf
 
 chown -R unbound:unbound /var/unbound
 
-/usr/local/sbin/unbound -c /var/unbound/unbound.conf
+if ! start_output=$(/usr/local/sbin/unbound -c /var/unbound/unbound.conf 2>&1); then
+        msg="$(echo "$start_output" | tr '\n' ';')"
+	logger -s -t unbound -p user.error "unbound start failed. output was: $msg"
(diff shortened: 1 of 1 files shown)


fichtner:
I like the idea, but I'm a little concerned about the impractical advice Unbound is giving our users here. Perhaps it's better to start unbound-check and output that if Unbound did not start? However, the address in use is probably hard for unbound-check to produce.

kulikov-a:
I agree, a little more talkative than would like ) but it would also be a shame to lose the opportunity to catch a startup error (and not just a config one). I'll see if I can come up with something more )

fichtner:
Sure, I don't mind all that much, but users are unpredictable. Just sharing early feedback in the hope it's helpful :)

kulikov-a:
@fichtner how about that? (a little more code than expected, but I think more logic)
moved up obsolete file deleting and inclusions copying so that the configuration check was already on a fully prepared config; if the config check is unsuccessful and the anchor key is mentioned in the output - try to update (plus this excludes attempts to update the key if it is not used in the config), if not - abort the execution with a configuration error.
I did not remove the startup error logging - it can still be useful imho (in this case, I "accidentally" bind nginx to this port)

unbound_err2

fichtner:
I think we can go without overcomplicating this for the time being.

Issue #5635, opened 2022-03-19
firmware: iOS 16 layout issues with bootstrap components


MainServerSpace:
The audit button dropdown does not work on mobile devices but can verify it works on Desktop. Any ideas?

EmperorsLoungeVPN:
https://user-images.githubusercontent.com/86547540/159139009-67d1b468-14ea-47c1-b6be-1f72f21bce97.MOV

EmperorsLoungeVPN:
Physically have to close out the side menu for it to go to status page also.

MainServerSpace:
Any update on a fix

MainServerSpace:
Would be nice to see the mobile webui work properly again

fichtner:
Maybe if you give device information or any sort of direction where this happens that would help?

MainServerSpace:
I gave all the details it’s happening on my iOS device using any browser I’m currently using the newest version and all the steps to reproduce is to login to mobile and go the the firmware and check area and perform a security audit or health audit and then the buttons become stuck and don’t allow for the security or health audit to be performed. I shared a video of it happening but can’t really see the clicks

fichtner:
> I gave all the details

Ok, but...

> it’s happening on my iOS device using any browser

That's something you haven't said before. ;)

I've tried iOS and the layout is a bit broken but the buttons do work... layout issues possibly since iOS 16, but you haven't said that either ;)


Cheers,
Franco

MainServerSpace:
You do have a point but this was working previously but since router upgrade its been doing this. I did put a broad term of Mobile devices in my first comment message in light of others and not just my self.

fichtner:
We removed Internet Explorer support in 22.7 but I doubt that iOS would complain about that specifically. Problem with phones is that getting to the JS error is tedious... if the button doesn't work the whole page may be broken. The check for updates button works or not?

fichtner:
BTW, can you try the standard theme? What is odd to me is that the audit button has a primary color but it's not a primary button...

MainServerSpace:
> We removed Internet Explorer support in 22.7 but I doubt that iOS would complain about that specifically. Problem with phones is that getting to the JS error is tedious... if the button doesn't work the whole page may be broken. The check for updates button works or not?

Yes the check for updates for OS work the others don't

MainServerSpace:
> BTW, can you try the standard theme? What is odd to me is that the audit button has a primary color but it's not a primary button...

Can try that using the custom theme

[...]

PR #6253, opened 2023-01-19
firmware.volt: fix regression in previous. typo

kulikov-a:
fixes Updates tab behavior after checking for updates.
ref. https://github.com/opnsense/core/issues/5635#issuecomment-1396823926
sorry for the inattention (


fichtner:
Merged, thanks!

PR #6286, opened 2023-01-30
Lobby/Dashboard/System Log widget - add severity filter

kulikov-a:
Hi!
FR ref (sort of) : https://forum.opnsense.org/index.php?topic=31560.msg152374#msg152374
System Log widget can be used to monitor and react to important events. Probably an additional Severity filter would be useful (especially since the data is already coming from the backend)?
PR tries to add the severity filter.
plus:
fixed weird typos in tags.
use selectpicker for selects
a uniform code is used to fill the selects.
empty result message is aligned with the same in log grids

Thanks!


fichtner:
Merged, thanks!

PR #6288, opened 2023-01-31
MVC/Firewall/Util - prevent possible infinite loop in port aliases parsing

kulikov-a:
Hi!
possible fix for https://github.com/opnsense/core/commit/2ba91361a7e27f616fa8f1b46250267ffac34e74#commitcomment-98713264
Prevent possible infinite loop in getPortAlias if alias references itself somehow (add current (or nested) alias to "parsed aliases" array before parsing)

### DIFF
--- a/src/opnsense/mvc/app/library/OPNsense/Firewall/Util.php
+++ b/src/opnsense/mvc/app/library/OPNsense/Firewall/Util.php
@@ -208,6 +208,7 @@ class Util
         $result = array();
         foreach (self::$aliasObject->aliasIterator() as $node) {
             if (!empty($name) && (string)$node['name'] == $name && $node['type'] == 'port') {
+                $aliases[] = $name;
                 foreach ($node['content'] as $address) {
                     if (Util::isAlias($address)) {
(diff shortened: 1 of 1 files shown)


AdSchellevis:
ok, now I feel really stupid, in my head I added the deadlock prevention, actually added the plumbing for it in https://github.com/opnsense/core/commit/d3f245b6466202280e402ea8998b21a1f10e9559 and then moved on.... 

Looking at your addition, I kind of assume we only need to add one `$aliases[] = $name;` between 

https://github.com/opnsense/core/blob/31973bee54dc4e6f85b549c019611811750aa726/src/opnsense/mvc/app/library/OPNsense/Firewall/Util.php#L211-L212

To avoid the alias being resolved again while still resolving.

kulikov-a:
I was also a little surprised at the lack of filling the array with an explicit preparation for this)
yes, it looks like I'm a bit overprotective. Of course, one addition should be enough, thanks!

AdSchellevis:
@kulikov-a thanks!

PR #6289, opened 2023-01-31
Firewall/NAT/Port Forward - dont calc local port range for alias

kulikov-a:
Hi!
Sorry, me again )
If port alias is specified as a Redirect target port, range is not calculated for rdr rule (local port is omitted or first port in alias is used): https://github.com/opnsense/core/blob/fdcd17cd5712f8a166cc76af9ee49cded3e617ff/src/opnsense/mvc/app/library/OPNsense/Firewall/ForwardRule.php#L123-L136
thanks!


kulikov-a:
Thanks!)

PR #6290, opened 2023-01-31
hostname/nameservers can be regenerated from cron

syserr0r:
As suggested in https://github.com/opnsense/plugins/issues/3090

### DIFF
--- a/src/opnsense/service/conf/actions.d/actions_dns.conf
+++ b/src/opnsense/service/conf/actions.d/actions_dns.conf
@@ -3,7 +3,7 @@ command:/usr/local/sbin/pluginctl -c dns_reload
 parameters:
 type:script_output
 message:rewriting /etc/resolv.conf and /etc/hosts
-description:Regenerate fixed hostnames (/etc/hosts) and nameservers (/etc/resolv.conf)
+description:Reload local DNS resolver configuration
 
 [query]
 command:/usr/local/opnsense/scripts/dns/query_dns.py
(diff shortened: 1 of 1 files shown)


src/opnsense/service/conf/actions.d/actions_dns.conf
```diff
@@ -3,6 +3,7 @@ command:/usr/local/sbin/pluginctl -c dns_reload
 parameters:
 type:script_output
 message:rewriting /etc/resolv.conf and /etc/hosts
+description:Regenerate fixed hostnames (/etc/hosts) and nameservers (/etc/resolv.conf)
```
fichtner, line 6: Can we simplify this a bit perhaps like the following:

```suggestion
description:Reload local DNS configuration
```

syserr0r: I feared it might be too long -- my concern with 'local DNS' is it might be confused with reloading unbound or dnsmasq.

Does `Reload local DNS resolver configuration` work?

fichtner: Perhaps it’s best to use the message approach and say rewrite xxx and yyy explicitly? Though the problem is it could change and it has before. The job was always called reload_dns internally but it’s more of an apply on a system level collecting from all plugged interface configurations.

src/opnsense/service/conf/actions.d/actions_dns.conf
```diff
@@ -3,6 +3,7 @@ command:/usr/local/sbin/pluginctl -c dns_reload
 parameters:
 type:script_output
 message:rewriting /etc/resolv.conf and /etc/hosts
+description:Reload local DNS resolver configuration
```
fichtner, line 6:
```suggestion
description:Reload system DNS resolver configuration
```

related change for clarity: ed6cd638ac7e24

fichtner:
Merged, thanks!

PR #6304, opened 2023-02-04
Make description field show for all types of VIP

FingerlessGlov3s:
Currently description only shows for CARP type, but the description should show for all.


AdSchellevis:
@FingerlessGlov3s thanks!

PR #6323, opened 2023-02-12
ntpd.conf: Remove "disable monitor" to get rid of log warnings 

meyergru:
There is a log message "2023-02-12T14:33:48	Warning	ntpd	restrict: 'monitor' cannot be disabled while 'limited' is enabled" ever so often when rate limiting is enabled. Disabling rate limiting is not advisable and even then, there will be another warning because certain combinations of rate limiting and kiss-of-death are chosen. ntpd options should probably be overhauled anyway.

However, according to the referenced https://www.cisa.gov/uscert/ics/advisories/ICSA-14-051-04, this issue has been fixed long ago. The current version 4.2.8 of ntpd is not longer vulnerable to this, such that "disable monitor" is no longer neccessary.


fichtner:
merged, thanks!

PR #6331, opened 2023-02-16
Unbound: let the logger wait for the pipe a bit

kulikov-a:
Hi!
ref. https://forum.opnsense.org/index.php?topic=32494.0
looks like logger db without index initializes so much faster that pipe may not be ready yet due to json loading?
thanks!

### DIFF
--- a/src/opnsense/scripts/unbound/logger.py
+++ b/src/opnsense/scripts/unbound/logger.py
@@ -185,14 +185,22 @@ class DNSReader:
 
         self._setup_db()
 
-        try:
-            # open() will block until a query has been pushed down the fifo
-            self.fd = open(self.target_pipe, 'r')
-        except InterruptedError:
-            self.close_logger()
-        except OSError:
-            syslog.syslog(syslog.LOG_ERR, "Unable to open pipe. This is likely because Unbound isn't running.")
-            sys.exit(1)
+        r_count = 0
+        pipe_ready = False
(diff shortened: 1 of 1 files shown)


AdSchellevis:
Hi Alex, am I right in assuming the reader (the other end) is to fast in trying to access the pipe and doesn't re-try? I'm not sure the actual issue in the dnsbl_module.py script in that case.

kulikov-a:
Hi Ad!  )
I agree that adding a second poke to the `logger.py` is the most correct solution. it just seems to me that creating a pipe before json loading does not spoil anything and never failed during testing (before fixing the db speed, creating\loading the index in `logger.py` gave the `dnsbl_module.py` time to create the pipe?).
i'll try to add a retry on `OSError` exception in `logger.py`

AdSchellevis:
I'm not worried about changing the order of setup, just want to avoid masquerading the real issue. As we're currently refactoring some code in the python unbound module, I rather keep the order as is to avoid it being set back after merge anyway :)

kulikov-a:
@AdSchellevis got it ) done
it just opening pipe on second try now

AdSchellevis:
@kulikov-a looks good, maybe increase the wait to 1 second so it has a bit more wiggle room. the module caches anyway in the meantime

kulikov-a:
done ;)

AdSchellevis:
@kulikov-a looks good, thanks!

PR #6339, opened 2023-02-19
Changes for issue 5569

kevinfason:
Per issue 5569 I made some proposed changes to the two files around NTP network servers. I broke out the changes in two phases. One is to simply parse the GUI entered servers to determine if they are a public pool server or not and set them correctly in the ntpd.conf file. Secondly to add setting in GUI to set maxcount. This tells ntpd how many concurrent servers to maintain. Due to the pool directive, status page will not show only the servers declared on the network tab (GPS,PPS unimpacted).  With the use of pool servers I felt it would be useful for users to set this if they wish.

### DIFF
--- a/src/etc/inc/plugins.inc.d/ntpd.inc
+++ b/src/etc/inc/plugins.inc.d/ntpd.inc
@@ -335,9 +335,9 @@ function ntpd_configure_do($verbose = false)
     foreach (explode(' ', $config['system']['timeservers']) as $ts) {
         /* Determine if Network Time Server is from the NTP Pool or not */
         if (preg_match("/\.pool\.ntp\.org$/", $ts)) {
-        $ntpcfg .= "pool {$ts}";
+            $ntpcfg .= "pool {$ts}";
         } else {
-        $ntpcfg .= "server {$ts}";
+            $ntpcfg .= "server {$ts}";
         }
         if (in_array($ts, $iburst)) {
             $ntpcfg .= ' iburst';
--- a/src/etc/inc/plugins.inc.d/ntpd.inc
+++ b/src/etc/inc/plugins.inc.d/ntpd.inc
@@ -323,7 +323,12 @@ function ntpd_configure_do($verbose = false)
     $ntpcfg .= "\n\n# Upstream Servers\n";
     /* foreach through ntp servers and write out to ntpd.conf */
     foreach (explode(' ', $config['system']['timeservers']) as $ts) {
+        /* Determine if Network Time Server is from the NTP Pool or not */
+        if (preg_match("/\.pool\.ntp\.org$/", $ts)) {
+        $ntpcfg .= "pool {$ts}";
+        } else {
         $ntpcfg .= "server {$ts}";
+        }
         if (in_array($ts, $iburst)) {
             $ntpcfg .= ' iburst';
         }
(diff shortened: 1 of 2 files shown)


fichtner:
@kevinfason sorry I lost track of this.. is this tested/finished?

src/etc/inc/plugins.inc.d/ntpd.inc
```diff
@@ -200,7 +200,17 @@ function ntpd_configure_do($verbose = false)
         $ntpcfg .= '12';
     }
     $ntpcfg .= "\n";
-
+
```
fichtner, line 203:
```suggestion

```

src/etc/inc/plugins.inc.d/ntpd.inc
```diff
@@ -200,7 +200,16 @@ function ntpd_configure_do($verbose = false)
         $ntpcfg .= '12';
     }
     $ntpcfg .= "\n";
-
+    /* Add Max Clock Count */
```
fichtner, line 203:
```suggestion

    /* Add Max Clock Count */
```

src/etc/inc/plugins.inc.d/ntpd.inc
```diff
@@ -323,7 +333,12 @@ function ntpd_configure_do($verbose = false)
     $ntpcfg .= "\n\n# Upstream Servers\n";
     /* foreach through ntp servers and write out to ntpd.conf */
     foreach (explode(' ', $config['system']['timeservers']) as $ts) {
+        /* Determine if Network Time Server is from the NTP Pool or not */
+        if (preg_match("/\.pool\.ntp\.org$/", $ts)) {
+        $ntpcfg .= "pool {$ts}";
+        } else {
         $ntpcfg .= "server {$ts}";
```
fichtner, line 340:
```suggestion
            $ntpcfg .= "pool {$ts}";
        } else {
            $ntpcfg .= "server {$ts}";
```

src/etc/inc/plugins.inc.d/ntpd.inc
```diff
@@ -412,6 +427,26 @@ function ntpd_configure_do($verbose = false)
     }
     if (empty($config['ntpd']['notrap'])) { /*note: this one works backwards */
         $ntpcfg .= ' notrap';
+    }
+            /* Access restrictions for Pool */
```
fichtner, line 431: is this necessary? for mixed scenarios?

src/www/services_ntpd.php
```diff
@@ -91,6 +92,9 @@
     if (!empty($pconfig['orphan']) && ($pconfig['orphan'] < 0 || $pconfig['orphan'] > 15 || !is_numeric($pconfig['orphan']))) {
         $input_errors[] = gettext("Orphan mode must be a value between 0..15");
     }
+    if (!empty($pconfig['maxclock']) && ($pconfig['maxclock'] < 0 || $pconfig['maxclock'] > 99 || !is_numeric($pconfig['maxclock']))) {
```
fichtner, line 95: "0" will not pass !empty() check

kevinfason:
> @kevinfason sorry I lost track of this.. is this tested/finished?

It is ready to go IMO with some potential cleanup by someone better skilled than I. I have been using it on 4 systems since I made the proposed changes.

fichtner:
I've taken your code and improved it a bit: 7366d785a62d

PR #6340, opened 2023-02-19
unbound: fix typo in logger. create a pipe early in dnsbl_module.py

kulikov-a:
Hi!
I'm very ashamed, perhaps the strangest typo on my end (I've never even used unary operators in real code).
as a result, the logger waits forever for the pipe ..
typo fix shows a dependence of the required waiting time on the size of the dnsbl (on a test vm with all lists enabled, it can reach up to 6 sec, so I set it to 10).
Besides of changing the order of actions in the `dnsbl_module.py` (first create a pipe, then load the json), using `Thread` to call load_dnsbl showed a good result.

### DIFF
--- a/src/opnsense/scripts/unbound/logger.py
+++ b/src/opnsense/scripts/unbound/logger.py
@@ -188,7 +188,7 @@ class DNSReader:
         r_count = 0
         pipe_ready = False
         # give dnsbl_module.py some time to create a pipe
-        while r_count < 3 and not pipe_ready:
+        while not pipe_ready:
             try:
                 # open() will block until a query has been pushed down the fifo
                 self.fd = open(self.target_pipe, 'r')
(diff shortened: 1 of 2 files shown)


AdSchellevis:
the `+=` vs `=+` I also missed, we should likely also change the order in the `dnsbl_module.py` to make sure the pipe doesn't have to wait for the list

kulikov-a:
@AdSchellevis sorry, should i add  `dnsbl_module.py` changes now? (when we were discussing https://github.com/opnsense/core/pull/6331, I thought it was better to wait for the refactoring and make changes if this is not taken into account)

AdSchellevis:
@kulikov-a maybe it's better to add the change in master as well indeed, the refactor might take a bit longer to enter a release and changing the order should be harmless. sorry for the noise.

kulikov-a:
@AdSchellevis understood, thanks! I hope at least this time I did not plant typos (need to do something with this really..)

AdSchellevis:
@kulikov-a thanks!

PR #6344, opened 2023-02-20
Unbound: implement wildcard blocking and refactor dnsbl_module

swhite2:
Since the implementation has changed a bit, the same logic as https://github.com/opnsense/core/pull/6340 for the dnsbl_module has been included, but adapted to the new situation.


src/opnsense/mvc/app/models/OPNsense/Unbound/Unbound.xml
```diff
@@ -173,6 +173,9 @@
             </whitelists>
             <blocklists type="CSVListField">
                 <Required>N</Required>
+                <Mask>/^((\*\.)?((?:(?:[a-z0-9]|[a-z0-9][a-z0-9\-]*[a-z0-9])\.){1,})(?:[a-z0-9]|[a-z0-9][a-z0-9\-]*[a-z0-9]))$/i</Mask>
```
fichtner, line 176: *.example.co.uk will fail?

src/opnsense/mvc/app/controllers/OPNsense/Unbound/forms/dnsbl.xml
```diff
@@ -41,7 +41,11 @@
         <type>select_multiple</type>
         <style>tokenize</style>
         <allownew>true</allownew>
-        <help>List of domains to blocklist. Only exact matches are supported.</help>
+        <help>
+            List of domains to blocklist. Only exact matches and wildcards are supported.
+            Wildcards are supported 2 TLDs down and only at the beginning of a domain,
+            e.g. "*.example.com", but not "*", "*.com" or "example.*.com".
```
fichtner, line 47: Just as an idea: wildcards are pretty powerful and thus restricted, but perhaps adding a separate field for wildcard blocks is easier for users and code taking the same type of input.

src/opnsense/service/templates/OPNsense/Unbound/core/dnsbl_module.py
```diff
@@ -1,5 +1,5 @@
 """
```
fichtner, line 1: no strong feelings on this end of the scripting ;)

src/opnsense/mvc/app/models/OPNsense/Unbound/Unbound.xml
swhite2: Good one, this slipped my mind.

src/opnsense/mvc/app/controllers/OPNsense/Unbound/forms/dnsbl.xml
swhite2: Sure, makes sense :)

src/opnsense/service/templates/OPNsense/Unbound/core/dnsbl_module.py
swhite2: I was already wondering whether it would be relevant here. Maybe it's better to omit it to stay in line with the examples as well as signify it's not really meant to be executed directly for it's main purpose.

swhite2:
Due to https://github.com/opnsense/core/pull/6344/commits/358c938c7e99413553573e5bfc56cacbe067e30a, wildcards are currently not taking effect since it requires a small parsing change on the blocklist preprocessing side, which will be addressed in https://github.com/opnsense/core/pull/6350

fichtner (approved):

src/opnsense/mvc/app/controllers/OPNsense/Unbound/forms/dnsbl.xml
```diff
@@ -41,7 +41,22 @@
         <type>select_multiple</type>
         <style>tokenize</style>
         <allownew>true</allownew>
-        <help>List of domains to blocklist. Only exact matches are supported.</help>
+        <help>
+            List of domains to blocklist. Only exact matches and wildcards are supported.
```
fichtner, line 45: Wait, this help text is wrong now with nbound.dnsbl.wildcards addition, right?

src/opnsense/mvc/app/models/OPNsense/Unbound/Unbound.xml
```diff
@@ -174,6 +174,12 @@
             <blocklists type="CSVListField">
                 <Required>N</Required>
             </blocklists>
+            <wildcards type="CSVListField">
+                <Required>N</Required>
+                <Mask>/^((?:(?:[a-z0-9]|[a-z0-9][a-z0-9\-]*[a-z0-9])\.){1,})(?:[a-z0-9]|[a-z0-9][a-z0-9\-]*[a-z0-9])$/i</Mask>
```
fichtner, line 179: isn't this a simple FQDN now?

src/opnsense/mvc/app/controllers/OPNsense/Unbound/forms/dnsbl.xml
swhite2: You're right, missed a spot. Thanks:)

src/opnsense/mvc/app/models/OPNsense/Unbound/Unbound.xml
swhite2: Yes, as far as I can see the regex fits the requirement (particularly: not accepting a TLD). Would you suggest something else?

fichtner: I was just wondering if we could use an existing FQDN instead of hand rolling this. No TLD means at least one dot? That would be a simpler regex then combined with FQDN built in. But I don’t know. Also co.uk is a TLD in some sense.

swhite2: > No TLD means at least one dot? That would be a simpler regex then combined with FQDN built in. But I don’t know. Also co.uk is a TLD in some sense.

At least one dot, yes. The problem with considering cases like `co.uk` is that there are a lot more to account for: https://publicsuffix.org/list/public_suffix_list.dat. So the intent here is to at least prevent blocking everything for the most common cases.

>  I was just wondering if we could use an existing FQDN instead of hand rolling this.

How would that look for CSVListField validation?

Issue #6337, opened 2023-02-16
Allow adding custom input to configuration of Monit daemon


codiflow:
**Important notices**

Before you add a new report, we ask you kindly to acknowledge the following:

- [X] I have read the contributing guide lines at https://github.com/opnsense/core/blob/master/CONTRIBUTING.md
- [X] I am convinced that my issue is new after having checked both open and closed issues at https://github.com/opnsense/core/issues?q=is%3Aissue

**Is your feature request related to a problem? Please describe.**

The configuration regarding the Monit daemon using the OPNsense GUI is very limited. So one can't use / configure all features of the daemon.

**Describe the solution you like**

Instead of adding input and all the other additional stuff for every single feature it would be more sustainable to have a **text area** where you can add a **custom monit configuration** which automatically gets added to the **end** of the configuration file (and therefore overwrites / extends all the other options).

**Describe alternatives you considered**

As the configuration file gets overwritten with every configuration change it is not sufficient to add the changes manually to `/usr/local/etc/monitrc`. So currently adding custom stuff  to the Monit daemon in a sustainable way is not possible at all.

**Additional context**

An example use case would be to be able to increase the limits of the Monit daemon with this feature:

```
## Set limits for various tests. The following example shows the default values:
##
set limits {
     programOutput:     5120 B,      # check program's output truncate limit
#     sendExpectBuffer:  256 B,      # limit for send/expect protocol test
     fileContentBuffer: 5120 B,      # limit for file content test
#     httpContentBuffer: 1 MB,       # limit for HTTP content test
#     networkTimeout:    5 seconds   # timeout for network I/O
#     programTimeout:    300 seconds # timeout for check program
#     stopTimeout:       30 seconds  # timeout for service stop
     startTimeout:      120 seconds  # timeout for service start
#     restartTimeout:    30 seconds  # timeout for service restart
}
```

A missconfiguration in this field would not need additional checking as you can already "misconfigure" Monit with the service checks and the changes will not be applied then. The log will show whats wrong and the issue can easily be solved.

AdSchellevis:
If monit support import hooks, that might be an option. Unvalidated user input is not going to happen as explained in https://docs.opnsense.org/development/guidelines/basics.html#safeguard-user-input

codiflow:
Oh thats an interesting point thanks for that background information.

Of course monit supports those file hooks (I hope I understood that right):
https://mmonit.com/monit/documentation/monit.html#INCLUDE-FILES

[...]

PR #6346, opened 2023-02-20
Added permanent include statement for custom Monit files

codiflow:
Referencing https://github.com/opnsense/core/issues/6337

### DIFF
--- a/src/etc/monit.opnsense.d/00-example.conf
+++ b/src/etc/monit.opnsense.d/00-example.conf
@@ -1,1 +1,1 @@
-# This is a placeholder file. You can add your custom and advanced Monit configuration here.
+# This is a placeholder file. You can add your advanced Monit configuration in this directory.
--- a/src/etc/monit.opnsense.d/00-example.conf
+++ b/src/etc/monit.opnsense.d/00-example.conf
@@ -1,0 +1,1 @@
+# This is a placeholder file. You can add your custom and advanced Monit configuration here.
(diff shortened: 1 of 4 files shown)


codiflow:
I checked the paths twice (in PR and in docs) and `/usr/local/etc/monit.opnsense.d` with monit.**opnsense**.d should be fine.

We could also use monit.**conf**.d but I think having the opnsense here is not a problem at all 😎

src/opnsense/service/templates/OPNsense/Monit/+TARGETS
```diff
@@ -1,2 +1,3 @@
 monitrc:/usr/local/etc/monitrc
 rc.conf.d:/etc/rc.conf.d/monit
+00-example.conf:/usr/local/etc/monit.opnsense.d/00-example.conf
```
fichtner, line 3: Can we do this without the templates? Just add the proper git file directly

codiflow: I'm sorry but I don't get what you mean. Could you explain me what exactly are you expecting and where should this file be located then?

fichtner: remove this entry and move 00-example.conf to src/etc/monit.opnsense.d/00-example.conf and done

codiflow:
Would appreciate if you could check again if the PR is now ready for being merged. Thanks!

src/etc/monit.opnsense.d/00-example.conf
```diff
@@ -0,0 +1 @@
+# This is a placeholder file. You can add your custom and advanced Monit configuration here.
```
fichtner, line 1:
```suggestion
# This is a placeholder file. You can add your advanced Monit configuration in this directory.
```

fichtner:
I suppose @AdSchellevis agrees... looks simple enough now, thanks!

PR #6350, opened 2023-02-21
Unbound: blocklists: allow scripts to extend blocklist functionality

swhite2:


src/opnsense/scripts/unbound/blocklists/__init__.py
```diff
@@ -0,0 +1,250 @@
...
+        with domains as keys and a dictionary of metadata as values
+        """
+        pass
+
+    def _write_cache(self):
+        """
+        Cache so each derived class can decide whether a download is necessary
+        """
+        if self.cnf:
+            with open('/tmp/unbound-blocklists.conf.cache', 'w') as cache_config:
```
AdSchellevis, line 77: @swhite2 I don't think this file is generic as `unbound-blocklists.conf` belongs to `DefaultBlocklistHandler` in order to be able to reuse files, it might even be better to define a "cache" directory and set a ttl on these files to download. This way we could loose part of the complexity of determining if we should download and offer reusability of files when they are downloaded from the same location.

src/opnsense/scripts/unbound/blocklists/__init__.py
```diff
@@ -0,0 +1,250 @@
...
+        for bl in list(blocklists.values())[1:]:
+            for key, value in bl.items():
+                if key not in first:
+                    # no collision, merge
+                    first[key] = value
+                else:
+                    # a handler with a lower priority has provided a policy
+                    # on a domain that already exists in the blocklist,
+                    # add it for debugging purposes
+                    first[key].setdefault('collisions', []).append(value)
```
AdSchellevis, line 217: I'm having some difficulties wrapping my head around this, how is priority handled? If for example the first handler offers a list with networks and the next doesn't, which one prevails?

swhite2: Priority is determined by a simple `priority` attribute in the plugged in class

AdSchellevis: ah, got it

src/opnsense/scripts/unbound/blocklists/__init__.py
swhite2: > @swhite2 I don't think this file is generic as `unbound-blocklists.conf` belongs to `DefaultBlocklistHandler` 

This is an oversight, it should be `self.config`, not `/tmp/unbound-blocklists.conf.cache`. Each derived class defines its own config file.

src/opnsense/scripts/unbound/blocklists.py
```diff
@@ -1,7 +1,7 @@
 #!/usr/local/bin/python3
 
 """
-    Copyright (c) 2020 Ad Schellevis <ad@opnsense.org>
+    Copyright (c) 2023 Ad Schellevis <ad@opnsense.org>
```
AdSchellevis, line 4: 2020-2023 :)

src/opnsense/scripts/unbound/blocklists/default_bl.py
```diff
@@ -0,0 +1,112 @@
...
+            if self.cnf.has_option('settings', 'rcode'):
+                cfg['rcode'] = self.cnf.get('settings', 'rcode')
+        return cfg
+
+    def get_blocklist(self):
+        result = {}
+        for blocklist, bl_shortcode in self._blocklists_in_config():
+            per_file_stats = {'uri': blocklist, 'skip': 0, 'blocklist': 0, 'lines': 0}
+            for entry in self._domains_in_blocklist(blocklist):
+                domain = entry.lower()
```
AdSchellevis, line 55:
```
per_file_stats['lines'] += 1
```
^^ missing (always reporting 0), but `_domains_in_blocklist()` already cleanses the data, it might be less relevant here to report.

AdSchellevis:
@swhite2 some small remarks, but looks good!

src/opnsense/scripts/unbound/blocklists/default_bl.py
swhite2: Nice catch, moved it to `_blocklist_reader()` as a separate log line.

src/opnsense/scripts/unbound/blocklists.py
swhite2: Woops, thanks :)

PR #6355, opened 2023-02-23
GUI and SSH listener recommended on all interfaces in system_advanced_admin.php

steljwagh:
In security terms it is not recomended to open the gui/ssh port on all interfaces, altough you still need a ACL entry to let traffic pass, we should not recommend this listener on ALL interfaces.

I strongly advice not to suggest opening admin port on ALL interfaces on a firewall product.

### DIFF
--- a/src/www/system_advanced_admin.php
+++ b/src/www/system_advanced_admin.php
@@ -690,7 +690,7 @@ $(document).ready(function() {
               <tr>
                 <td><a id="help_for_webguiinterfaces" href="#" class="showhelp"><i class="fa fa-info-circle"></i></a> <?=gettext('Listen Interfaces') ?></td>
                 <td>
-                  <select id="webguiinterface" name="webguiinterfaces[]" multiple="multiple" class="selectpicker" title="<?= html_safe(gettext('All (recommended)')) ?>">
+                  <select id="webguiinterface" name="webguiinterfaces[]" multiple="multiple" class="selectpicker" title="<?= html_safe(gettext('All (not recommended)')) ?>">
 <?php foreach ($interfaces as $iface => $ifacename): ?>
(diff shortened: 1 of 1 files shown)


fichtner:
Besides @AdSchellevis's point (which is easily testable BTW), the recommendation is for reliability, not security. You may end up locking yourself out of the firewall when selecting specific interfaces with different modes of connectivity that are non-static.

steljwagh:
Then replace the "recommended" to "default".

I do understand the choice made out of support also you guys are probably dyed in the wool here. But i fell in disbelieve seeing such an recommendation.

Also it all starts default on the WAN/LAN interface, and the default anti-lockout rule helps with... well, not locking yourself out.
Even more crazy is that it presents a mgmt interface on outside, yes no acl permit, but still the risk.

The risks starts with a dmz, guest, opt interfaces, then also those interfaces listens on 443 and 22, with the 

Give it a thought not to recommend it please, that's all, default ok.

fichtner:
One could just as easily add a doc on how to properly set up a DMZ to avoid overly open firewall policies. For most of what we do we care about same default, the rest is the admins choice and responsibility.

steljwagh:
If you care about sane default, you can't put management listeners on the WAN interface if you have a security perspective. You could easility add a doc for everything, I don't see your point there.

I feel sane providing the community with a safe product, if you have to read a doc to avoid open ports it's not the right product for you maybe. But mistakes can be made, even quicker when you recommend open mgmt ports on the WAN interface (or all in this matter)

fichtner:
To be frank, the listener is 0.0.0.0 and we don’t even know if the user has a WAN. We also established it’s not accessible in that case (if it’s not a WAN-only setup where you want to be able to access it). 😉

PR #6360, opened 2023-02-25
Unbound: typo

kulikov-a:
Hi!
ref. https://forum.opnsense.org/index.php?topic=32597.0
Type casted a bit early imho: may skip updateBlocklist action if model node exists but empty.
Thanks!

### DIFF
--- a/src/opnsense/mvc/app/controllers/OPNsense/Unbound/Api/SettingsController.php
+++ b/src/opnsense/mvc/app/controllers/OPNsense/Unbound/Api/SettingsController.php
@@ -48,7 +48,7 @@ class SettingsController extends ApiMutableModelControllerBase
             $domain = $this->request->getPost('domain');
             $type = $this->request->getPost('type');
             $mdl = $this->getModel();
-            $item = (string)$mdl->getNodeByReference('dnsbl.' . $type);
+            $item = $mdl->getNodeByReference('dnsbl.' . $type);
 
             if ($item != null) {
                 $remove = function ($csv, $part) {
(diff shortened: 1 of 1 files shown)


AdSchellevis:
Hi Alex,

The location of the cast looks off indeed, but reading briefly through the code, I don't expect the `if` will be skipped when the node exists in the model as `getNodeByReference()` returns `null` when not found, both "blocklists" and "whitelists" probably always exist in this case. The current fix is good in my opinion, as it implements the proper "does exist" check, I'm just not sure if this was supposed to fix anything else. 

Best regards,

Ad

kulikov-a:
Hi Ad!
i think you are right but it should be without type juggling `if ($item !== null)` for this?

AdSchellevis:
Hi Alex,

An empty string will indeed also match on `!=`, but in this case I do expect an empty object of the FieldType class being returned, which logically doesn't match on `null`.  In this case I don't think the `!=` and `!==` make an actual difference, but both work as far as I'm concerned. Looking at the code, I expect this is only intended to make sure someone doesn't try to inject "hutsefluts" as type which obviously doesn't exist. 

I do believe your current fix is the right one for the intend of the code, if the current issue is that it doesn't want to add a domain when the list is empty, this is likely the proper fix as well.

Best regards,

Ad

kulikov-a:
@AdSchellevis 
> if the current issue is that it doesn't want to add a domain when the list is empty, this is likely the proper fix as well

yes, thanks! )

kulikov-a:
@AdSchellevis Thanks!

PR #6364, opened 2023-02-26
Web UI Interfaces: Re-order Overview to be at top

haarp:
I find it very un-intuitive that the "main" entry in the interfaces sub-menu, **Overview**, isn't at the top. Especially since some entries, like **Diagnostics** are consistent in their position (always at the bottom).

This change puts **Overview** at the top of interfaces.

I do some re-ordering so that this will work well until an user reaches over 60 interfaces, at which point ordering might become weird. But who has 60 interfaces?

thanks! :)

### DIFF
--- a/src/opnsense/mvc/app/models/OPNsense/Base/Menu/MenuSystem.php
+++ b/src/opnsense/mvc/app/models/OPNsense/Base/Menu/MenuSystem.php
@@ -248,7 +248,7 @@ class MenuSystem
         }
 
         // add groups and interfaces to "Interfaces" menu tab...
-        $ordid = 0;
+        $ordid = 901;
         foreach ($iftargets['if'] as $key => $descr) {
             if (array_key_exists($key, $iftargets['gr'])) {
                 $this->appendItem('Interfaces', $key, array(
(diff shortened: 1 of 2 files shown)


fichtner:
As there are no other complaints I’d like to leave this as is. Usually these changes end up raising more “but why” questions afterwards.

Cheers,
Franco

haarp:
Fair enough. Looks like I'll have to live with this unique piece of weirdness :)

fichtner:
At some time in the future overview and assignment page are likely going to be merged into one MVC page. That should address your concern perhaps. But it might still be 1-2 years away.

haarp:
I have found an alternative solution using browser-side CSS. :D

Ublock Origin has the option of injecting CSS into a page. Add these two rules (adapt the IP to your OPNsense IP/hostname):

```
10.0.0.1##div#Interfaces.in:style(display: flex !important; flex-direction: column !important;)
10.0.0.1##div#Interfaces a:has-text(Overview):style(order:-1 !important;)
```

It sets the Interface parent (when expanded, i.e. `in` class set) to `flex`. This allows the `order` directive to work on individual list elements.

Issue #6389, opened 2023-03-03
mvc: Phalcon framework dependency


fichtner:
**Important notices**

Before you add a new report, we ask you kindly to acknowledge the following:

- [x] I have read the contributing guide lines at https://github.com/opnsense/core/blob/master/CONTRIBUTING.md
- [x] I am convinced that my issue is new after having checked both open and closed issues at https://github.com/opnsense/core/issues?q=is%3Aissue

**Is your feature request related to a problem? Please describe.**

Given the state of Phalcon, it’s probably a good idea to start planning for losing more dependencies on the framework.

**Describe the solution you like**

A good start would probably be to start with our model implementation, which seems to depend on the following classes for which we should build native php alternatives in that case:

* ~~Phalcon\Filter\Validation\Validator\Email~~
* ~~Phalcon\Filter\Validation\Validator\ExclusionIn~~
* ~~Phalcon\Filter\Validation\Validator\InclusionIn~~
* ~~Phalcon\Filter\Validation\Validator\Numericality~~
* ~~Phalcon\Filter\Validation\Validator\PresenceOf~~
* ~~Phalcon\Filter\Validation\Validator\Regex~~
* ~~Phalcon\Filter\Validation\Validator\Url~~
* ~~OPNsense\Phalcon\Logger\Logger~~
* ~~Phalcon\Filter\Validation\Exception~~
* Phalcon\Di\FactoryDefault
* ~~Phalcon\Logger\Adapter\Syslog~~
* ~~Phalcon\Messages\Message~~
* ~~Phalcon\Messages\Messages~~
* ~~Phalcon\Config\Config~~

**Describe alternatives you considered**

N/A

**Additional context**

N/A

PR #6410, opened 2023-03-11
Demo of volt framework using Dnsmasq

agh1467:
**Disclaimer: Only for demonstration purposes for the front-end volt template framework. Not ready for merge, looking for feedback regarding improvements, or different ways to do any part of this framework. It still needs lots of polish. There's comments for a lot of things, some of the comments may be incorrect. Some are just comments/questions/ideas for myself.**

I wanted to see if a conversation could be started around the Phalcon volt framework, there hasn't been any significant work on the framework for many years. Since MVC work is planned for the upcoming roadmap, this might be an appropriate time to talk about it. In working on my plugin for dnscrypt-proxy, I've built a framework based on the original one in Core. I've included the framework adapted to Core and an example of usage targeting the Dnsmasq plugin.

The main goal I had in building this framework was to reduce the need to write HTML, PHP, or Javascript directly. I didn't like the fact that I had to do this with early iterations of the dnscrypt-proxy plugin, jumping from file to file changing variable names and DOM element ids and attributes so everything matched was a huge headache, so I strived to eliminate the need to do it at all. I think this framework heavily embraces the spirit of MVC. It's intended to make it easier to rapidly develop a plugin, taking the labor out of rendering a page to look right (with the advantage of keeping styles uniform across the OPNsense UI). It keeps everything in sync, as well as makes it easy to expand the framework to render different kinds of things (e.g. rows/fields) on the page.

Here is a summary of some of the things that are done differently from the current Core framework:
1. The most major change is that the form XML is not converted to an array. After much deliberation and contemplation, I saw little value is this because it was incredibly inhibiting to flexibility of rending content on the page based on content/structure of the XML. With each new element or concept, the array needed to be changed to accommodate what was already in the XML structure. This is entirely skipped, and the XML is used as a SimpleXMLElement object instead.
2. Instead of using just partials to render, there is extensive utilization of several features of Phalcon, namely includes/partials/macros/extends/blocks. These allow for the framework to lay a solid foundation, without a plugin developer re-building from scratch each page. All that needs to happen is inherit the right files, and the plugin will build itself. The use of blocks allow for overriding specific sections of the base templates with their own blocks. Additionally, the extensive use of partials allows for easy expansion of the framework by simply adding a new file.
3. Instead of manually designing a bootgrid in HTML, creating the UIBootgrid call, defining the APIs, etc. Bootgrids are rendered entirely automatically. Bootgrid dialogs are rendered automatically. It is even possible to render bootgrids within bootgrid dialogs (see the Host Overrides dialog box). There are some logistical hurdles to overcome for this to be a successful application, but just as a proof of concept it is possible and easy to do.
4. It adds support for different kinds of 'containers' beyond tabs/subtabs, to include boxes. It also supports rendering without any containers, just fields (it just puts them in a box currently). It supports multiple tabs, multiple sets of tabs, multiple models, multiple boxes, etc. It's designed to be driven by the XML in a dynamic way with some specific restrictions (related mostly to the model or tabs, etc).
5. Also, there is some flexibility for where things are defined. In the Dnsmasq form XML, there is extensive use of element attributes, instead of defining named sub-elements for these things. This is flexible and backwards compatible. There are two macros which are mostly used to get these: get_xml_prop(), and get_field_id()
6. Also introduced is throwing in the volt for error handling. If, for example, a field has no id defined (as either an attribute or sub-element), it will throw, and return the element missing the id so it's easy to see where and what the problem is. (Kudos to whoever build the crash reporter, it's amazing, 10/10 would recommend throwing).
7. Not demonstrated here, but is noted in some of the comments for the fields, is the concept of what I'm kind of calling "field control," where activities with a field can have actions. For example, checking a box for a setting may enable or disable another field (or fields) on the page. It's flexible and dynamic and would work for fields which have a known state like checkboxes, dropdowns, radio buttons, etc.
8. Includes an "apply changes" box which can appear when saving a config without applying it. I think this will be very helpful in making the user experience uniform across the UI since currently this apply button may appear in multiple places throughout using OPNsense. If this box were to be used universally it would always appear in the same place. It would need much consideration to be universally usable though as noticed in the comment with it.

In this example for Dnsmasq, there is 0 static HTML, PHP, or Javascript. Everything is written automatically by the framework from the form XML.

To see more of the framework in action, I use most features of it in the dnscrypt-proxy plugin (I do need to update it though, what's in the repo isn't the latest), since it was what drove the need for many of them. This is just an example of the front-end parts, there are more parts for the back-end such as API functionality which I also built with the intention of making them available in Core for all plugins to use.

You should be able to try it out with the following command, then browse to Services->Dnsmasq DNS->Settings:
`opnsense-patch -a agh1467 443dd9fc9dd7845f9a4b403fbc0e64bb0c3aa402`

This will _not_ clobber the existing partials, and shouldn't impact any other functionality. It lives alongside the current framework, and is only used if called directly by using the build_xml() macro.

Future ideas:
* Making it possible for plugins to override, and include new, fields or rows (or even containers) of their own.

### DIFF
--- a/src/opnsense/mvc/app/controllers/OPNsense/Dnsmasq/Plugin.php
+++ b/src/opnsense/mvc/app/controllers/OPNsense/Dnsmasq/Plugin.php
@@ -1,0 +1,144 @@
+<?php
+
+/*
+ * Copyright (C) 2015 Deciso B.V.
+ * All rights reserved.
+ *
+ * Redistribution and use in source and binary forms, with or without
+ * modification, are permitted provided that the following conditions are met:
+ *
+ * 1. Redistributions of source code must retain the above copyright notice,
+ *    this list of conditions and the following disclaimer.
+ *
+ * 2. Redistributions in binary form must reproduce the above copyright
+ *    notice, this list of conditions and the following disclaimer in the
+ *    documentation and/or other materials provided with the distribution.
+ *
+ * THIS SOFTWARE IS PROVIDED ``AS IS'' AND ANY EXPRESS OR IMPLIED WARRANTIES,
+ * INCLUDING, BUT NOT LIMITED TO, THE IMPLIED WARRANTIES OF MERCHANTABILITY
+ * AND FITNESS FOR A PARTICULAR PURPOSE ARE DISCLAIMED. IN NO EVENT SHALL THE
+ * AUTHOR BE LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY,
+ * OR CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF
+ * SUBSTITUTE GOODS OR SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS
+ * INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN
+ * CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE)
+ * ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
+ * POSSIBILITY OF SUCH DAMAGE.
+ */
+
+namespace OPNsense\Dnsmasq;
+
+/**
+ * Class PluginIndexController extends Core's OPNsense\Base\IndexController.
+ *
+ * All controllers for this plugin, should extend from this one.
+ *
+ *
+ *
+ *
+ *
+ * @package OPNsense\Dnsmasq
+ */
+class Plugin
+{
+    /**
+     * Array for this plugin's settings.
+     *
+     * @var array $settings
+     */
+    protected $settings = array();
+
+    /**
+     * Array for this plugin's settings.
+     *
+     * @var array $settings
+     */
+    protected $plugin_ini = '/usr/local/opnsense/data/dnsmasq/plugin.ini';
+
+    /**
+     * Array for this plugin's settings.
+     *
+     * @var array $settings
+     */
+    public function __construct()
+    {
+        // Set values for the protected settings array.
+        $this->settings = [
+            // The common name of the plugin.
+            'plugin_name' => 'dnsmasq',
+            // A safe name to use in DOM id's.
+            'plugin_safe_name' => 'dnsmasq',
+            // A common (stylized) label for the plugin.
+            'plugin_label' => 'Dnsmasq DNS',
+            // The service name that this plugin has.
+            'plugin_service_name' => 'dnsmasq',
+            // The name of the configd module for this plugin.
+            'plugin_configd_name' => 'dnsmasq',
+            // The name of the log directory, assumed to be in /var/log
+            'plugin_log_dir_name' => 'dnsmasq'
+        ];
+    }
+
+    /**
+     * Function to get the settings out of the class.
+     *
+     * @var array $settings
+     */
+    public function getSettings()
+    {
+        return $this->settings;
+    }
+
+    /**
+     * Returns the name of the configd for this plugin.
+     *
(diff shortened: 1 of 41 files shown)


AdSchellevis:
To me it's not entirely sure what the goal is here. When it comes to the Phalcon framework itself, we aim to limit our exposure in the (near) future (https://github.com/opnsense/core/issues/6389).  

If we miss components in our current framework as documented in our development manual (https://docs.opnsense.org/develop.html), I don't mind discussing them on a per topic base (what to add and to which means). When preparing for structural changes, it should be very clear how it helps developers and how we're going to make sure any new practice will be incorporated by developers using the framework in place.  Building standard form based applications is already rather simple as it is.

agh1467:
Maybe I'm doing a poor job of explaining it.

If I were to specify specific goals:
1. Make it easier to unify the look and feel of the UI. It should make it easier to add new things, and to make changes universally from a single place.
2. Make it easier for plugin developers with little web development experience to build plugins. Define what you want in the XML, and the framework builds it.

For a well versed web developer, building a standard form based application may be simple, but for others like myself it doesn't come so easily. I'd like to make it easier for people like me to build plugins for OPNsense.

I'm not sure what you mean by when talking about structure. The structure of this framework is effectively the same as it is today (what was already there was basically great and well thought out). Only the components are broken out, and dynamically called via partials instead of living in a single file (ex: form_input_tr.volt).

If the plan is to step away from Phalcon, will Volt continue to be used or replaced with a different templating engine? If that were the case I wouldn't really want to figure any ways to make this one better, but I'd want to see similar functionality in a future one. I have definitely seen myself some of the troubles that the validation changes have made. Though not mentioned in the linked, rendering isn't immune to issues (https://github.com/phalcon/cphalcon/issues/16019#issuecomment-1196399232).

AdSchellevis:
In an ideal world the frontend would be a bit more dynamic (browser generating forms instead of backend), but given our current codebase, that's difficult to achieve. I do expect Volt to be there for a while.

The problem with the current suggested simplification is in my opinion that it adds way to much plumbing all around the codebase.  In order to build a simple form application at the moment, there's really not much one has to arrange, it basically consists of:

- A model (let's take a simple example, https://github.com/opnsense/core/blob/master/src/opnsense/mvc/app/models/OPNsense/Interfaces/Loopback.xml)
- An API controller, which inherits from ApiMutableModelControllerBase (for example https://github.com/opnsense/core/blob/master/src/opnsense/mvc/app/controllers/OPNsense/Interfaces/Api/LoopbackSettingsController.php)
- Optionally an API as service handler (inherited from ApiMutableServiceControllerBase to avoid code duplication)
- A controller + view which binds a form (same example https://github.com/opnsense/core/blob/master/src/opnsense/mvc/app/controllers/OPNsense/Interfaces/LoopbackController.php , https://github.com/opnsense/core/blob/master/src/opnsense/mvc/app/controllers/OPNsense/Interfaces/forms/dialogLoopback.xml, https://github.com/opnsense/core/blob/master/src/opnsense/mvc/app/views/OPNsense/Interface/loopback.volt)

Quite clean and simple, sure there are (always) possibilities to simplify things, but then we first have to establish where we can gain and what would be needed for adoption. Adding more layers of abstraction also adds complexity when trying to debug issues at some point. 

There was a time when we thought it would make sense to add a script to generate the boilerplate, but in practice, it's another component that someone has to maintain and probably won't be used that much.

In my humble opinion it's better to extend documentation of topics when it's not clear what they do then trying to add the next layer on top of it hoping that people won't get lost on the next abstraction. Most of the components are there for a reason, as in practice the simple forms are easy, but someone at some point will need a hook to extend it somewhere. 

So if you do want to extend templating in some way, my advise would be to limit the scope and start with a description of the problem you're trying to solve (try to keep steps small, so people are able to review and offer feedback). 

Don't get me wrong, I'm not saying nothing can or will be simplified, just mentioning that the current direction is likely not going to mature.

AdSchellevis:
We're planning to migrate dnsmasq in core, probably for 24.1. closing this for now.

PR #6412, opened 2023-03-11
ipsec: Check ipsec config exists before using, closes #6411

### LINKED ISSUE #6411: PHP Warning:  Undefined array key "ipsec" in /usr/local/etc/inc/plugins.inc.d/ipsec.inc on line 1273
**Important notices**

**Describe the bug**

An error is displayed on the console (and reported in crash reporter) when no ipsec configuration is present, as in when a new install is performed.
`PHP Warning:  Undefined array key "ipsec" in /usr/local/etc/inc/plugins.inc.d/ipsec.inc on line 1273`

**To Reproduce**

1. Install OPNsense.
2. Minimally configure interfaces.
3. Boot.

**Expected behavior**

No message appears in the console for this condition.

None

**Screenshots**

image

**Relevant log files**

**/tmp/PHP_errors.log**
```
[11-Mar-2023 18:10:39 Etc/UTC] PHP Warning:  Undefined array key "ipsec" in /usr/local/etc/inc/plugins.inc.d/ipsec.inc on line 1273
```

**Environment**

OPNsense 23.1.3 (amd64, OpenSSL).

agh1467:
Looks like `$ipseccfg` is only used in the if condition just after this definition, and it is also wrapped in an `isset()`. This should catch when ipsec node doesn't exist in the config, set to an empty array, and then the following isset() will be looking for the named index in an empty array which will not be set.

### DIFF
--- a/src/etc/inc/plugins.inc.d/ipsec.inc
+++ b/src/etc/inc/plugins.inc.d/ipsec.inc
@@ -1270,7 +1270,7 @@ function ipsec_configure_do($verbose = false, $interface = '')
         set_single_sysctl('net.key.preferred_oldsa', '0');
     }
 
-    $ipseccfg = isset($config['ipsec']) ? $config['ipsec'] : [];
+    $ipseccfg = $config['ipsec'] ?? [];
     $a_phase1 = isset($config['ipsec']['phase1']) ? $config['ipsec']['phase1'] : [];
     $a_phase2 = isset($config['ipsec']['phase2']) ? $config['ipsec']['phase2'] : [];
     $a_client = isset($config['ipsec']['client']) ? $config['ipsec']['client'] : [];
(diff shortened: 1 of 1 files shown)


src/etc/inc/plugins.inc.d/ipsec.inc
```diff
@@ -1270,7 +1270,7 @@ function ipsec_configure_do($verbose = false, $interface = '')
         set_single_sysctl('net.key.preferred_oldsa', '0');
     }
 
-    $ipseccfg = $config['ipsec'];
+    $ipseccfg = isset($config['ipsec']) ? $config['ipsec'] : [];
```
fichtner, line 1273: We prefer

```suggestion
    $ipseccfg = $config['ipsec'] ?? [];
```

fichtner: But the question is if we have to go any further if it’s not there/empty.

agh1467: That's neat! I didn't know that shortcut. It's definitely better. I'm not familiar with this file so I'm not sure about other options. Seems like maybe it's trying to do ipsec things even when ipsec isn't configured at all.

williamdes (approved):

PR #6414, opened 2023-03-12
Only require the common name when generating certificates

### LINKED ISSUE #2317: Certificate generation requires entries for optional fields

chstaudte:
This pull request fixes #2317. When generating certificates, now only the "common name" field is required. The whole internet works like this, and I would like to see this feature here, also.

I tested it by generating CAs, intermediate CAs, client certs and CSRs.

### DIFF
--- a/src/etc/ssl/opnsense.cnf
+++ b/src/etc/ssl/opnsense.cnf
@@ -53,7 +53,7 @@ serial		= $dir/serial		# The current serial number
 crlnumber	= $dir/crlnumber	# the current crl number
 					# must be commented out to leave a V1 CRL
 crl		= $dir/crl.pem		# The current CRL
-private_key	= $dir/private/cakey.pem# The private key
+private_key	= $dir/private/cakey.pem	# The private key
 RANDFILE	= $dir/private/.rand	# private random number file
 
 x509_extensions	= usr_cert		# The extensions to add to the cert
@@ -127,17 +127,17 @@ string_mask = utf8only
 
 [ req_distinguished_name ]
 countryName			= Country Name (2 letter code)
-countryName_default		= AU
+#countryName_default		= AU
 countryName_min			= 2
 countryName_max			= 2
 
 stateOrProvinceName		= State or Province Name (full name)
-stateOrProvinceName_default	= Some-State
+#stateOrProvinceName_default	= Some-State
 
 localityName			= Locality Name (eg, city)
 
 0.organizationName		= Organization Name (eg, company)
-0.organizationName_default	= Internet Widgits Pty Ltd
+#0.organizationName_default	= Internet Widgits Pty Ltd
 
 # we can do this but it is not needed normally :-)
 #1.organizationName		= Second Organization Name (eg, company)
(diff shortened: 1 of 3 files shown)


fichtner:
4 years later, I think your argument deconstructs itself. I don’t mind, but I don’t want to spend time review and improve to make sure the page works afterwards. We’ve seen a lot of breakage via these files.

Cheers,
Franco

chstaudte:
Why do you think my argument deconstructs itself? Do you have any arguments? Besides from the legacy certificate authorities, there is no reason to require more than the common name field. All ACME-compliant CAs just use this field, and if I want to generate an internal OPNsense PKI there also is no reason to require them.
But my main use case is for OpenVPN certificates. When generating one for each of our employees, it makes admin life easier to just state the employee's name as common name and that's it.

Regarding testing: As I already wrote, I tested as good as I could, and everyone else is invited to also do so. This change is running on our production setup now.

fichtner:
> The whole internet works like this

4 years without another PR, 20 years without the feature.

kulikov-a:
@chstaudte 
sorry, technically speaking, in that case why is CN a required RDN? why not require only a "non-empty" DN?
then it will be more in line with the rfc

>  life easier to just state the employee's name as common name and that's it.

once again sorry, doesn't gui fill the fields with the values ​​of the issuing certs by default? or i missing some

chstaudte:
So I checked the RFC (and also the ITU-T Rec. X.501) and if I haven't overlooked something, then just no single field is required. But at least one field must be given, quote from the RFC:
> Where it is non-empty, the subject field MUST contain an X.500
   distinguished name (DN).  The DN MUST be unique for each subject
   entity certified by the one CA as defined by the issuer field.  A CA
   MAY issue more than one certificate with the same DN to the same
   subject entity.

This corresponds to what @kulikov-a stated. And that also means, that the implementation of OPNsense (and pfSense) was wrong for 20 years now.

But in my opinion, also making the CN optional wouldn't be a good idea. That would allow to generate certificates with for example only a country name set. While that is covered by the RFC, it's not really intuitive, right?

kulikov-a:
> While that is covered by the RFC, it's not really intuitive, right?
CN omitting is not so uncommon practice though
https://www.websecurity.digicert.com/theme/roots

> was wrong for 20 years now.
IMHO this is not necessarily `wrong`, it can be called a PKI policy: it can be useful, for example, when comparing DNs (more fields - more accurate result)

chstaudte:
Well, I don't really feel comfortable with doing such a change. As @fichtner wrote, it is easy to break things in these files, and I don't have a deep understanding of the source code. This pull request may not be the best solution, but I'm sure it is way better than before.

AdSchellevis:
closing, currently we want to avoid risk of regressions and the Trust module should be rewritten at some point anyway.

PR #6418, opened 2023-03-14
Unbound: migrate General page to MVC

swhite2:


src/etc/inc/plugins.inc.d/dhcpd.inc
```diff
@@ -574,6 +576,8 @@ function dhcpd_dhcp4_configure($verbose = false)
 
     killbypid('/var/dhcpd/var/run/dhcpd.pid');
 
+    $unbound_enabled = !empty((string)(new \OPNsense\Unbound\Unbound())->general->enabled);
```
fichtner, line 579: probably want to do something about this mid-term through a more straightforward "run" plugin hook. invoking models everywhere will not help the legacy code.

fichtner: (thinking about it it makes sense so we can actually make plugins like bind act in the same way if both unbound and dnsmasq are off)

src/opnsense/mvc/app/models/OPNsense/Unbound/ACL/ACL.xml
```diff
@@ -2,7 +2,7 @@
     <page-services-dnsresolver>
         <name>Services: Unbound DNS: General</name>
         <patterns>
-            <pattern>services_unbound.php*</pattern>
+            <pattern>ui/unbound/general/*</pattern>
```
fichtner, line 5: api missing? ;)

src/opnsense/mvc/app/models/OPNsense/Unbound/FieldTypes/UnboundInterfaceField.php
```diff
@@ -0,0 +1,67 @@
...
+            if ((empty($node->virtual) || $key == 'lo0') && !empty($node->enable)) {
+                $this->internalOptionList[$key] = !empty($node->descr) ? (string)$node->descr : strtoupper($key);
+            }
+        }
+
+        foreach ($config->openvpn->children() as $mode => $setting) {
+            if (!isset($setting->disable) && !empty($setting)) {
+                $key = 'ovpn' . substr($mode, 8, 1) . (string)$setting->vpnid;
+                $type = substr($mode, 8, 6);
+                $this->internalOptionList[$key] = "OpenVPN {$type} (" . (!empty($setting->description) ?
```
fichtner, line 60: Not sure this is still needed in its entirety if we want to kill special ACLs anyway.

src/www/services_dnsmasq.php
```diff
@@ -82,9 +82,10 @@
         if (isset($pconfig['local_ttl']) && $pconfig['local_ttl'] !== '' && !is_numericint($pconfig['local_ttl'])) {
             $input_errors[] = gettext("You must specify a valid TTL for local DNS");
         }
-        $unbound_port = empty($config['unbound']['port']) ? "53" : $config['unbound']['port'];
+        $unbound_mdl = new \OPNsense\Unbound\Unbound();
+        $unbound_port = $unbound_mdl->general->port;
```
fichtner, line 86: not a string?

fichtner (approved):
ok for master minus comments 👍

src/www/services_dnsmasq.php
swhite2: nice catch :)

src/opnsense/mvc/app/controllers/OPNsense/Unbound/Api/ServiceController.php
```diff
@@ -47,4 +47,18 @@ public function dnsblAction()
...
+     * Only used on the general page to account for resolver_configure and dhcp hooks
+     * since these check if unbound is enabled.
+     */
+    public function reconfigureGeneralAction()
+    {
+        $this->sessionClose();
+        $backend = new Backend();
+        $backend->configdRun('dns reload');
+        $result = $this->reconfigureAction();
+        $backend->configdRun('dhcpd restart');
```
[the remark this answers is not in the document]
swhite2: maybe so for dhcpd, but `dns_reload` handles `/etc/revolv.conf`. We could however narrow this down to an `isFieldChanged` on `enabled`.

src/opnsense/mvc/app/models/OPNsense/Unbound/FieldTypes/UnboundInterfaceField.php
swhite2: Not sure how we're able to remove this unless we account for it in the migration, since users won't be able to deselect it anymore. But removing something that's explicitly configured is probably a no-go anyway.

src/opnsense/mvc/app/controllers/OPNsense/Unbound/Api/ServiceController.php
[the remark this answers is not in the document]
fichtner: yep, dns_reload is unavoidable

src/opnsense/mvc/app/models/OPNsense/Unbound/FieldTypes/UnboundInterfaceField.php
fichtner: you are right, it's also fudged into interfaces and outgoing interfaces, but we will still be able to let go off the ACL part once that is being worked on.

swhite2: absolutely :)

Issue #6343, opened 2023-02-20
DHCP - Setting "Gateway" to "None" On a Specific Device Does Not Work


NateroniPizza:
**Important notices**

Before you add a new report, we ask you kindly to acknowledge the following:

- [X ] I have read the contributing guide lines at https://github.com/opnsense/core/blob/master/CONTRIBUTING.md
- [X ] I am convinced that my issue is new after having checked both open and closed issues at https://github.com/opnsense/core/issues?q=is%3Aissue

I see that this was previously brought up a couple years back (https://github.com/opnsense/core/issues/4843), but that was when this was initially added (because it was not available). At this point, the feature looks to no longer be working.

**Describe the bug**

I am attempting to set a static DHCP mapping for one of my clients without a gateway (this is on a dual-NIC device, and to ensure anything outside of the NICs' local subnets get routed out a specific NIC, I want the secondary NIC to not receive a gateway when pulling DHCP). When I set the "gateway" of that device's static mapping to "none," as specified in the "gateway" description, it still gives it a gateway. If I set this same setting to "none" on the DHCP interface context, it is correctly applied (i.e. no gateway assigned when pulling DHCP on client).

**The current OPNsense version where the bug first appeared:** Currently running OPNsense 23.1_6-amd64
**The last OPNsense version where the bug did not exist:** Don't know. I've only been running it since the previous version, and it wasn't working there too.
**The exact URL of the GUI page involved (if any):** https://10.0.20.1/services_dhcp_edit.php?if=opt3&id=0 (installation-specific. To get there go to Services>DHCPv4>[select interface]>Edit existing DHCP client).

*Tip*: to validate your setup was working with the previous version, use opnsense-revert (https://docs.opnsense.org/manual/opnsense_tools.html#opnsense-revert)

**To Reproduce**

1. Service>DHCPv4>[your interface]
2. Set up new static DHCP mapping.
3. Set "Gateway" for the static mapping to "none," per the Gateway description.
4. Save and renew IP on DHCP client.

**Expected behavior**

No gateway assigned to DHCP client.

**Describe alternatives you considered**

None available, aside from setting static IP.

**Screenshots**

![image](https://user-images.githubusercontent.com/14255967/220013829-9dc0d4e5-50b2-46d8-abdf-d7b52becc1b2.png)

**Relevant log files**

n/a

**Additional context**

n/a

**Environment**

Software version used and hardware type if relevant, e.g.:

Client: Windows 11
[...]

PR #6425, opened 2023-03-18
Update dhcpd.inc to Allow Static Mapping to Not Assign Gateway to Client

NateroniPizza:
Apologies if there is anything wrong with this pull request - this is the first pull request I've done. Also note that I don't really know PHP (I just have a basic understanding of programming logic, and was googling stuff to figure out the syntax), so please double-check to ensure this doesn't mess anything unintended up.

See issue #6343 , particularly this post for an explanation of what I'm looking to do: https://github.com/opnsense/core/issues/6343#issuecomment-1437483299

**The problem's TL;DR:**
When creating a DHCP static mapping, the description of "Gateway" states that if you enter "none," it will not assign a gateway to that client (from the description: "Type "none" for no gateway assignment.") However, what this actually does is the same thing as leaving it blank - it omits the value from the client's static mapping in /var/dhcpd/etc/dhcpd.conf, which causes it to reference the pool's "option routers" entry, resulting in it assigning a gateway to the client anyway.

**Proposed solution:**
When looking through the dhcpd documentation, I did not see a valid value to put for "option routers" in a static mapping that would cause it to override the pool's "option routers" value to just leave it blank. In testing, though, I found that by putting an invalid hostname in for "option routers," it will not assign a hostname to the client pc (as it would need to resolve to an IP, then send that IP for the Gateway field on the client - since it cannot resolve it, it does not send an IP).

The change proposed in this pull request will make it so that if someone enters "none" for the Gateway value in a static mapping, it will add the line "option routers no-gateway-assigned;" to the static mapping's dhcpd.conf entry. As the chances of "no-gateway-assigned" resolving to an IP address in someone's environment would be essentially 0, I can't think of any issues with this. If you would like, this can be changed to any other equally-or-less-likely-to-already-be-used "hostname."

I have briefly (successfully) tested this on both a Windows-based and a Debian-based client, with OPNsense 23.1.3_4-amd64. Note that I did run into a problem with the "option domain-name-servers" being missing from pools at one point, but this appears to only have been when curl'd the dhcpd.inc from the Master branch - when I curl from the 23.1 branch and add just this change in, I do not have any issues.

### DIFF
--- a/src/etc/inc/plugins.inc.d/dhcpd.inc
+++ b/src/etc/inc/plugins.inc.d/dhcpd.inc
@@ -1126,6 +1126,8 @@ EOD;
 
                 if (!empty($sm['gateway']) && $sm['gateway'] != "none" && (empty($dhcpifconf['gateway']) || $sm['gateway'] != $dhcpifconf['gateway'])) {
                     $dhcpdconf .= "  option routers {$sm['gateway']};\n";
+                } elseif ($sm['gateway'] == "none") {
+                    $dhcpdconf .= "  option routers no-gateway-assigned;\n";
                 }
 
                 $smdnscfg = "";


fichtner:
Huh, why not change the validation instead to make this viable input? A magic value for an impossible isc-dhcp configuration is probably going to cause more reports like yours. Also, you can add override files with your required configuration already.

NateroniPizza:
> 

"None" is the input that should cause there to be no gateway assigned to a DHCP client, according to the description of "Gateway" I referenced in the linked issue. Are you proposing removing the "none" value as valid input, and instead change it to "no-gateway-assigned" and putting it in the config instead?

For what reason would it cause more reports? The only way I've been able to think of it causing issues would be if someone had "no-gateway-assigned" resolving to an IP, but I know that there's a lot more to any of this than I am aware of.

What override files? Are they accessible via the webGUI? This functionality is supposed to be present in the webGUI, so would like to avoid having to access the command-line to make this type of change.

fichtner:
Why remove anything? „none“ seems like either special case or plain host name to me.

Not all things need a GUI. It really depends on the demand and I don’t extrapolate much demand from the general lack of requests in this area.

Again, it was wrong to support none in the overrides, because it doesn’t work in isc-dhcp if you already have a router assigned globally.


Cheers,
Franco

kulikov-a:
@NateroniPizza 
Hi
have you tested setting the static lease gateway address to the lease address?
maybe this will do the trick without "trick"?)

mdbetancourt:
i had this issue my computer has 2 NIC one has internet access the another one is connected to a private network i'm keep getting the gateway messing with my internet connection i tried to assign to me a static ip and set the gateway as none (so the others device in the networks keep getting the gateway correctly) but it is just ignoring the "none" i have to assign it globally or use a dedicated vlan or interface for my computer to prevent this

PR #6432, opened 2023-03-22
Date/Time Display Short Format (Jan 01 00:00:00)

NOYB:
Fractional seconds is not meaningful for human consumption in these pages.  Uselessly takes up display space that is more useful for other fields.
Leave in logging but remove from these display pages.

System log
Firewall log (and widget)
Firewall alias last updated (trim micro seconds)


AdSchellevis:
not going to change data formats as iso timestamps are easier to parse for machines, changing the display format however isn't a problem https://github.com/opnsense/core/commit/38dfe47258afd078f071b330fff150906f2ae026

PR #6436, opened 2023-03-23
Can we please leave out the time delimiter ('T') too?  Better human readability.

NOYB:
Leave out delimiter ('T') from UI last updated time display for better readability.  Run together strings are less readable for we mere mortals.

### DIFF
--- a/src/opnsense/mvc/app/views/OPNsense/Firewall/alias.volt
+++ b/src/opnsense/mvc/app/views/OPNsense/Firewall/alias.volt
@@ -87,7 +87,7 @@
                     },
                     timestamp: function (column, row) {
                         if (row[column.id] && row[column.id].includes('.')) {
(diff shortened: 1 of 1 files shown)


src/opnsense/mvc/app/views/OPNsense/Firewall/alias.volt
```diff
@@ -87,9 +87,9 @@
                     },
                     timestamp: function (column, row) {
                         if (row[column.id] && row[column.id].includes('.')) {
-                            return row[column.id].split('.')[0];
+                            return row[column.id].split('.')[0].replace('T', ' ');
                         }
-                        return row[column.id];
+                        return row[column.id].replace('T', ' ');;
```
AdSchellevis, line 92: @NOYB this could be null, better leave this line as is, don't mind replacing the `T` when we know it's a time.

PR #6443, opened 2023-03-24
Log line timestamp (e.g. boot log)

NOYB:
If "timestamp" is empty and "line" begins with timestamp.  Use the line timestamp.

e.g. boot log

### DIFF
--- a/src/opnsense/mvc/app/controllers/OPNsense/Diagnostics/Api/LogController.php
+++ b/src/opnsense/mvc/app/controllers/OPNsense/Diagnostics/Api/LogController.php
@@ -81,6 +81,15 @@ class LogController extends ApiControllerBase
                     $severities
                 ]);
                 $result = json_decode($response, true);
+                foreach ($result['rows'] as &$record) {
+                    if (!$record['timestamp'] && $record['line']) {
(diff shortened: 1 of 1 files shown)


AdSchellevis:
boot log looks normal on my end, no need to add more hard to understand glue here.

NOYB:
Don't know what you are seeing on your end.  But it is not normal on this end.
SystemBootLogA

Appearing like the other logs with a date field would look normal.  Also easier to read with the date separated from the message. 
SystemBootLogB

As for hard to understand glue.  It's quite simple.  If timestamp field not provided by the log.  See if it's available from the line field.  A single straight forward simple "If" completely bypasses.  Leaving any existing timestamp untouched.

Seens that a date field would be better for external log systems also.

Boot log file snippet.  It is all stuffed in the "line" field.  This patch grabs the date from the "line" field and puts it in the proper, "timestamp", field where it belongs.
```
2023-03-21T20:43:41-07:00 rrd_configure[256] done.
2023-03-21T20:43:41-07:00 system_powerd_configure[256] Starting power daemon...
2023-03-21T20:43:41-07:00 system_powerd_configure[256] done.
```

PR #6447, opened 2023-03-26
Fix typo: "second-level subdir" in helptext

olmari:
Fix help-text error, should describe "second-level" instead "first level".

Reported-by: gecko_x2 (irc)
Signed-off-by: Sami Olmari <sami@olmari.fi>


fichtner:
Nice catch, thanks!

Issue #6348, opened 2023-02-20
Schedule - Time Range Description Replaced with URL Encoded String


kuya1284:
**Important notices**

Before you add a new report, we ask you kindly to acknowledge the following:

- [X] I have read the contributing guide lines at https://github.com/opnsense/core/blob/master/CONTRIBUTING.md
- [X] I am convinced that my issue is new after having checked both open and closed issues at https://github.com/opnsense/core/issues?q=is%3Aissue

**Describe the bug**

When saving a Time Range for a Schedule, the Time Range Description gets replaced with a URL encoded string.  For example, when I attempt to save the description as "9:00 PM to 11:59 PM", the string gets saved and replaced with "9%3A00%20PM%20to%2011%3A59%20PM", making the description very difficult to read. I'm new to OPNsense and first noticed this issue with 22.7.

**To Reproduce**

1. Go to **Firewall > Settings > Schedule** (https://[OPNsense IP]/firewall_schedule.php)
2. Click on the **Add** button (or edit an existing entry)
3. Configure the time range 
4. In **Time Range Description**, enter a description containing characters like spaces, colons, etc.
5. Click the **Add Time** button
6. The new entry will appear on the **Configured Ranges** row with the **Time Range Description** re-saved as a URL encoded string
7. Click the **Save** button
8. The new schedule will appear on the list of schedule entries, along with the **Time Range Description** as a URL encoded string

**Expected behavior**

The **Time Range Description** should appear just as I entered it. If a URL encoded string needs to be saved on the backend, that's fine. But when rendering the description, it must be URL decoded to not only make the description more readable, but also to prevent double-encoding in case the description were to be re-saved.

**Describe alternatives you considered**

I have no known workarounds or alternatives.

**Screenshots**

![image](https://user-images.githubusercontent.com/121985/220165404-87211c15-3a83-41fd-8eff-b4ca1bbd88fa.png)
![image](https://user-images.githubusercontent.com/121985/220165466-f1395727-d5bc-429c-9bf6-a25f82b93203.png)

**Relevant log files**

N/A

**Additional context**

N/A

**Environment**

OPNsense 23.1.1_2 (amd64, OpenSSL)
Protectli VP2410
Intel® Celeron® J4125 2.0GHz Quad Core

AdSchellevis:
It's unfortunately a known omission with very low priority due to the low usage and complexity of the legacy page, more info in https://github.com/opnsense/core/issues/3357

kuya1284:
[...]

PR #6460, opened 2023-03-31
Firewall Schedules - Patch for #6348

kuya1284:
* ~Adds logic to decode the time range description that was previously encoded via `escape()`~
* Improvements to leverage **jQuery** to insert schedule items in the DOM


src/www/firewall_schedule.php
```diff
@@ -97,6 +97,12 @@
                 }]
       });
     });
+
+    // Decode time range descriptions previously encoded via escape()
+    $('.range-description').each(function(i, description) {
```
AdSchellevis, line 102: I think we shouldn't try to present data differently, it might have side affects on "new" data and there's no way of knowing we should present differently.

AdSchellevis:
@kuya1284 looks good, one remark on the overview page, I think we should keep that as it was, the rest could be merged in my opinion. Its another step in simplifying the code and it seems to work fine.

src/www/firewall_schedule.php
kuya1284: Hi @AdSchellevis , thank you for reviewing the code. I see what you're saying. So if someone intentionally wanted to save a string such as this,

> 12%3A00%20AM%20to%208%3A00%20AM

it should be left as-is. No problem. I'll remove those lines of code and re-commit.  I'll also remove the same code that leverages `decodeURIComponent()` in `firwall_schedule_edit.php` as well.

Thank you.

kuya1284:
Hi @AdSchellevis , the change has been committed in [dec8d53](https://github.com/opnsense/core/pull/6460/commits/dec8d53a5700bc0025016cf13ed72ce10ce030d9). Thank you again for the review and the feedback.

AdSchellevis:
@kuya1284 thanks! I'll merge right away

kuya1284:
> @kuya1284 thanks! I'll merge right away

tyvm, @AdSchellevis

PR #6462, opened 2023-03-31
system: Handle empty DNS server gateway

n-thumann:
This PR fixes adds handling for empty DNS server gateway to fix ` /system_general.php: ROUTING: not a valid host gateway address: ''`.

In System -> Settings -> General a user can _optionally_ set a gateway for each DNS server. However, if no gateway is set for a DNS server the error above is thrown. This is caused by `system_host_route` being always called, even though there is no gateway set. If none is set, there's no need to setup a static route.


fichtner:
Merged, thanks!

PR #6474, opened 2023-04-04
Unbound: ACLs: Migrate to MVC

swhite2:
Things to note:

- The "internally generated" ACL entries have been removed in favor of a wildcard (well, a wildcard and localhost)
- Users can now change the default action used by the wildcard: either allow, deny or refuse.
- The legacy implementation wasn't very strict (you could add an entry without any input), this does at least require a name and at least one network. To facilitate the migration, if the 'name' property was empty, it will be set to 'Unnamed ACL'. Entries without defined networks will be discarded without notice.


src/opnsense/mvc/app/models/OPNsense/Unbound/Migrations/M1_0_6.php
```diff
@@ -0,0 +1,105 @@
...
+            'allow' => 'allow',
+            'deny' => 'deny',
+            'refuse' => 'refuse',
+            'allow snoop' => 'allow_snoop',
+            'deny nonlocal' => 'deny_non_local',
+            'refuse nonlocal' => 'refuse_non_local'
+        ];
+
+        if (!empty($config->unbound->acls)) {
+            foreach ($config->unbound->acls as $acl) {
```
AdSchellevis, line 50: @swhite2 it might be good to check for broken data, `continue` if `!isset($legacy_format[(string)$acl->aclaction])`.

AdSchellevis:
@swhite2 one small remark, but other than that can be squashed/merged into master. As it changes behavior, we will keep it there for quite some time anyway.

src/opnsense/mvc/app/models/OPNsense/Unbound/Migrations/M1_0_6.php
swhite2: @AdSchellevis Thanks, good point :)

PR #6477, opened 2023-04-05
proxy: allow more signs for username and password

Bi0T1N:
- eliminate the artificial limit of length and signs which does not exist in squid
- allow in particular `@` to support `first.last@mycompany.com` as username
- all other characters can be used via URL escaping (i.e. `%20` for spaces) but this also means `%` must be written as `%%`

see http://www.squid-cache.org/Doc/config/cache_peer/


AdSchellevis:
@Bi0T1N thanks!

PR #6480, opened 2023-04-06
New IPSec configuration interface: XAuth PAM (group) constraints

GottemHams:
With the legacy interface we could restrict mobile users (or more likely, PAM auth in general) to certain groups. Since the new interface has you enabling XAuth on an individual remote authentication round, it makes sense to also have related settings somewhere in the vicinity. I chose a separate subtab under `Connections [new]`, since these **do affect all auth attempts** that go through Strongswan's `xauth-pam` plugin (`libcharon`'s actually though), similar to what it originally does. I did notice `swanctl.conf` allows a `groups` directive for the remote auth round, which could allow for round-specific group constraints. Unfortunately though, after messing around quite a bit with OPNsense's PAM module it seems like XAuth in general doesn't really support groups on its own. ;_;

The new-style VPN settings will take precedence over legacy ones (in some places) to try and avoid conflict. Of course this isn't 100%, since if you configure a legacy phase 1 or phase 2 then I've allowed OPNsense to still insert some config in `swanctl.conf`. But due to the overlap with the legacy mobile client settings I did make it skip a fair portion of it for **strongswan.conf**. It should still be possible to configure a VPN connection through the new interface in such a way that mobile clients can connect too. The original mobile settings are kinda insecure anyways I think, so this might even help move people to more secure settings (the new method doesn't insert any phase config itself, it only configures `charon` to interact with PAM).

Furthermore, the constraints shouldn't conflict with auth methods that aren't XAuth, since `charon` only uses OPNsense's `IPsec` auth service specifically when it's told to use the `xauth-pam` plugin. RADIUS, EAP, etc all have their own plugins so no problems there. :>

**Do note though** that I also mirrored the auth database setting, although I'm not entirely sure what the implications of it are. In my case I only have the local DB, but choosing something different might not even be relevant?

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -11,6 +11,8 @@
 /src/etc/pkg.conf
 /src/etc/pkg/**
 /src/etc/ssh/**
+/src/etc/strongswan.conf
+/src/etc/swanctl/**
 /src/lib/**
 /src/opnsense/changelog
 /src/www/wpad.dat
--- a/.gitignore
+++ b/.gitignore
@@ -4,6 +4,8 @@
 *.pyc
 *.sass-cache
 *.volt.php
+/src/etc/filter_geoip.conf
+/src/etc/filter_tables.conf
 /src/etc/php.ini
 /src/etc/php/**
 /src/etc/pkg.conf
--- a/src/etc/inc/plugins.inc.d/ipsec.inc
+++ b/src/etc/inc/plugins.inc.d/ipsec.inc
@@ -1,6 +1,7 @@
 <?php
 
 /*
+ * Copyright (C) 2023 GottemHams
  * Copyright (C) 2019 Pascal Mathis <mail@pascalmathis.com>
  * Copyright (C) 2016 Deciso B.V.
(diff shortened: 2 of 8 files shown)


AdSchellevis:
Have you tried to just use the settings in "VPN: IPsec: Mobile Clients"? I don't really see a reason why the existing mobile settings wouldn't apply here (if they don't, we should probably aim for investigating that first).

GottemHams:
They do, but the main problem with those is that you can't **just** enable the XAuth settings, it also adds phase configuration. The way the new interface is set up kind of makes those settings redundant, so I figured it was meant to replace the legacy phase configuration plus the accompanying mobile settings in their entirety.

AdSchellevis:
ok, then we better fix the issue with the legacy page first, both will be here for some time, better to aim for consistency here. If you open a ticket, I don't mind taking a look at how we can do that in a most practical way (minimizing code)

GottemHams:
Works for me, by ticket I assume you mean a regular issue here on GitHub? :>

AdSchellevis:
yes please, I'll try to look at it asap, it really helps if there's someone to test in these cases

Issue #6491, opened 2023-04-16
If change the VLAN interface MAC address, the VLAN interface only works in `Promiscuous mode`


steelywing:
**Important notices**

Before you add a new report, we ask you kindly to acknowledge the following:

- [X] I have read the contributing guide lines at https://github.com/opnsense/core/blob/master/CONTRIBUTING.md
- [X] I am convinced that my issue is new after having checked both open and closed issues at https://github.com/opnsense/core/issues?q=is%3Aissue

**Describe the bug**

If change the VLAN interface MAC address, after a few minutes, the VLAN interface can not receive packet, but if set the VLAN interface to `Promiscuous mode`, it works again.

I think the problem is the parent (trunk) interface think that MAC address is not for itself, so it drop if the trunk interface is not in `Promiscuous mode`.

Usually using same MAC addresses in VLAN interfaces is fine, but I have 2 internet line from 1 ISP, it will flip flap when I use same MAC addresses 😥

**To Reproduce**

1. Create 2 VLAN interface
2. Change 1 of the VLAN interface MAC address
3. The interface that changed MAC address can not receive packet
4. The interface that does not change MAC address is works fine
5. Enable `Promiscuous mode`, the VLAN interface works again

**Environment**

OPNsense 23.1.5 (amd64, OpenSSL).

fichtner:
This is true. That’s why promiscuous setting exists in the GUI.

steelywing:
> This is true. That’s why promiscuous setting exists in the GUI.

So this is expected? Can we have a doc for it? 😀

fichtner:
Yes it’s expected. I don’t consider this a documentation issue as the limitation is within FreeBSD kernel and/or NIC drivers. It’s a fringe setup with the attached grain of salt.

We have made it possible to be configured unlike pfSense where this doesn’t even work today. So you see there is not much real world impact having documentation or not (and it should be documented somewhere within FreeBSD manual pages).

steelywing:
> Yes it’s expected. I don’t consider this a documentation issue as the limitation is within FreeBSD kernel and/or NIC drivers. It’s a fringe setup with the attached grain of salt.

But not everyone is FreeBSD expert, for the people like me not use FreeBSD at all except using OPNsense, I almost use 1 day debug for this issue, at first I don't know it is the MAC address issue (because my setup have several switch and ACL between OPNsense and the ISP device).

fichtner:
[...]

PR #6492, opened 2023-04-16
Help tip for enable "Promiscuous mode" for changing MAC address in VLAN interface

steelywing:
Is this commit acceptable?

#6491


fichtner:
I do think visibility is an issue here as well as bouncing a very old translation. Let me think about it.

fichtner:
I've modified it slightly and added more help to it.

fichtner:
Thanks!

PR #6499, opened 2023-04-17
webgui: prevent PHP session gc from running early

lin-xianming:
With the default PHP settings, there is a 1% chance for the session gc to trigger upon the start of a session, removing session files with modification times older than 24 minutes. This can cause sessions to timeout earlier than indicated by the webgui settings.


AdSchellevis:
@lin-xianming thanks! Apparently it doesn't happen very often, but the documentation certainly suggests that it may happen.

PR #6517, opened 2023-04-23
Firewall Log Widget: Header Tweak

NOYB:
"Interface" header makes column way wider then need be.
Abbreviate "Interface" with "I/F" to reduce unnecessary column width and line wrap.


NOYB:
Wrong one.

PR #6518, opened 2023-04-23
Firewall Log Widget: Header Tweak

NOYB:
"Interface" header makes column way wider then need be.
Abbreviate "Interface" with "I/F" to reduce unnecessary column width and line wrap.

### DIFF
--- a/src/www/widgets/widgets/log.widget.php
+++ b/src/www/widgets/widgets/log.widget.php
@@ -234,7 +234,7 @@ $nentriesinterfaces = isset($config['widgets']['filterlogentriesinterfaces']) ? 
       <th data-column-id="__digest__" data-type="string" class="hidden"><?= gettext('Hash') ?></th>
       <th data-column-id="action" data-type="icon" class="text-center"><?= gettext('Act') ?></th>
       <th data-column-id="__timestamp__" data-type="time"><?= gettext('Time') ?></th>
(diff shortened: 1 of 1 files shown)


AdSchellevis:
We tend to be cautious with these changes as experience learns the next person will open a ticket for "I/F" not being clear, so we rather keep it as is for now.

NOYB:
Grasping at straws.  Target audience, anyone capable of managing an OPNsense firewall, knows what the I/F column is, not only by the header but column contents is obvious.  Especially after selecting the interfaces in settings.

As obvious as which pedal is gas and break to anyone who is capable of driving a car.  Even the bad drivers out there know which pedal is gas and break.

Anyone opening a ticket for "I/F" not being clear in that widget context is being deliberately nitpickish.

fichtner:
Try to understand translators and users of other languages having to deal with "i/f" (not translated) or "if"... worst case already being translated to "when" so the whole thing is futile to start with.

NOYB:
"I/F" Google search.  First result: "What is an Interface (I/F)? - Definition from Techopedia"
There is an abundance of context with-in the widget.  Not being used in a vacuum.

> "if"... worst case already being translated to "when"

"I/F" not "if".  Is there a language that translates "I/F" to "when"?

fichtner:
Sure ok, you have me not being interested anymore. Have a nice day.

PR #6519, opened 2023-04-23
Firewall Log Widget: Date/Time display locale format selector

NOYB:
+ OPNsense language locale format (Jan 01 00:00)
 + Log Short format (MM-DD hh:mm)
 + Log Long format (YYYY-MM-DD-Thh:mm:ss+/-hh:mm)


AdSchellevis:
at the moment I expect we rather keep it as is, micromanaging formats may lead to more reports for little gain.

NOYB:
ISO 8601 format may be nice for computers.  But atrocious for humans.  A human friendly date format should be displayed in the web GUI.

PR #6554, opened 2023-05-11
unbound: Added cache-max-negative-ttl setting

hp197:
See: https://forum.opnsense.org/index.php?topic=23747.0

Maximum negative TTL for cached records.
Useful if the cloud isn't as fast as you are.

For example, if you add records to CloudFlare it may take a few moments before the records are propagated.
If the negative TTL on that domain 1h, you need to flush your local cache (or wait an hour).

The setting forces a max TTL on negative responses (NXDOMAIN responses).

before:
```
> dig test.tcdn.nl     

...
;; QUESTION SECTION:
;test.tcdn.nl.			IN	A

;; AUTHORITY SECTION:
tcdn.nl.		3600	IN	SOA	donna.ns.cloudflare.com. dns.cloudflare.com. 2300657040 10000 2400 604800 3600
...
```

after added setting set to 30:
```
>dig test.tcdn.nl     

...
;; QUESTION SECTION:
;test.tcdn.nl.			IN	A

;; AUTHORITY SECTION:
tcdn.nl.		30	IN	SOA	donna.ns.cloudflare.com. dns.cloudflare.com. 2300657040 10000 2400 604800 3600
...
```

locally tested and working against 23.1.6

### DIFF
--- a/src/opnsense/mvc/app/controllers/OPNsense/Unbound/forms/advanced.xml
+++ b/src/opnsense/mvc/app/controllers/OPNsense/Unbound/forms/advanced.xml
@@ -326,6 +326,17 @@
         </help>
     </field>
     <field>
+        <id>unbound.advanced.cachemaxnegativettl</id>
+        <label>Maximum Negative TTL for RRsets and messages</label>
+        <type>text</type>
+        <help>
+            Configure a maximum Negative Time to live in seconds for RRsets and messages in the cache.
+            When the internal TTL expires the negative response cache item is expired.
+            This can be configured to force the resolver to query for data more often in case you wont
+            get a valid answer.
+        </help>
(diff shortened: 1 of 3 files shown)


hp197:
I'm totally in for a freeform textfield in unbound advanced config, where we could (also) put this kind of custom config.
What do the core devs think about this?

fichtner:
@hp197 thanks for the PR. @swhite2 will review and merge at some point.

You can read about our custom config policy here https://docs.opnsense.org/development/guidelines/basics.html#safeguard-user-input

Exception to this policy are file-based overrides which Unbound already offers: https://docs.opnsense.org/manual/unbound.html#advanced-configurations


Cheers,
Franco

PR #6558, opened 2023-05-13
unbound: Add support for DoH and DoT

schreibubi:
Added additional values to the unbound statistics page which allow to monitor DoH and DoT queries.

Updated PR to work against latest master where the Unbound General page is implemented using MVC

### DIFF
--- a/src/etc/inc/plugins.inc.d/unbound.inc
+++ b/src/etc/inc/plugins.inc.d/unbound.inc
@@ -30,6 +30,19 @@
  * POSSIBILITY OF SUCH DAMAGE.
  */
 
+require_once('certs.inc');
+
+function export_pem_file($filename, $data, $post_append = null)
+{
+    $pem_content = trim(str_replace("\n\n", "\n", str_replace(
+        "\r",
+        "",
+        base64_decode((string)$data)
+    ) . ($post_append == null ? '' : "\n" . $post_append)));
+    file_put_contents($filename, $pem_content);
+    chmod($filename, 0640);
+}
+
 function unbound_enabled()
 {
     $mdl = new \OPNsense\Unbound\Unbound();
@@ -176,6 +189,10 @@ function unbound_generate_config()
         }
     }
 
+    $port = $general['port'];
+    $port_doh = $general['port_doh'];
+    $port_dot = $general['port_dot'];
+
     $bindints = '';
     if (!empty($general['active_interface'])) {
         $active_interfaces = explode(',', $general['active_interface']);
@@ -196,12 +213,25 @@ function unbound_generate_config()
         }
 
         foreach ($addresses as $address) {
-            $bindints .= "interface: $address\n";
+            $bindints .= "interface: $address@$port\n";
+            if (!empty($general['enable_dot'])) {
+                $bindints .= "interface: $address@$port_dot\n";
+            }
+            if (!empty($general['enable_doh'])) {
+                $bindints .= "interface: $address@$port_doh\n";
+            }
         }
     } else {
-        $bindints .= "interface: 0.0.0.0\n";
-        $bindints .= "interface: ::\n";
-        $bindints .= "interface-automatic: yes\n";
+        $bindints .= "interface: 0.0.0.0@$port\n";
+        $bindints .= "interface: ::@$port\n";
+        if (!empty($general['enable_dot'])) {
+            $bindints .= "interface: 0.0.0.0@$port_dot\n";
+            $bindints .= "interface: ::@$port_dot\n";
+        }
+        if (!empty($general['enable_doh'])) {
+            $bindints .= "interface: 0.0.0.0@$port_doh\n";
+            $bindints .= "interface: ::@$port_doh\n";
+        }
     }
 
     $outgoingints = '';
@@ -223,7 +253,7 @@ function unbound_generate_config()
 
     unbound_add_host_entries($ifconfig_details);
 
-    $port = $general['port'];
+
 
     /* do not touch prefer-ip6 as it is defaulting to 'no' anyway */
     $do_ip6 = isset($config['system']['ipv6allow']) ? 'yes' : 'no';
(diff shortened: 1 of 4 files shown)


schreibubi:
Still applies and works with 23.7. Needed to re-enable DoH/DoT and select the correct certificate after the upgrade though.

AdSchellevis:
@schreibubi The part I don't understand in this is, if Android contacts the dns server provided, why should it trust the certificate it received? I would expect this can only work with client configuration as well, which would make it less useful in common scenario's. Maybe (probably) I'm missing something, but without understanding the lay of the land, it would be difficult to incorporate a feature like this and describe it on our end.

doktornotor:
> @schreibubi The part I don't understand in this is, if Android contacts the dns server provided, why should it trust the certificate it received? 

It works just fine with proper certificates issued by system-trusted CAs (notably Lets Encrypt). It currently does **not** work with user installed CAs and Google's effort to fix it is about non-existent. See 
[Private DNS over TLS is ignoring user installed CA](https://issuetracker.google.com/issues/241147758
)

schreibubi:
@AdSchellevis Take a look at this page from google: https://developers.google.com/speed/public-dns/docs/dns-over-tls 
What I was referring to was the opportunistic privacy profile mentioned there. It protects against a passive attacker, but not an active one. Using DoT that way at least raises the bar for an attacker, so is better than doing nothing (i.e. unencrypted DNS). The client in this case can only verify that the certificate is signed by a valid CA, but not that it really talks to the server the user wants it to talk to (because the user did not specify the server name to use and it just used the DNS server specified by the DHCP response). There is BTW an IETF draft where the DHCP response will contain the DoT server name: https://www.ietf.org/archive/id/draft-ietf-add-ddr-10.html. Then the client will be able to verify the certificate the DoT server presents. Though DHCP will then become the weakest link in the chain....

When the user explicitly configures the DoT server, that will raise the bar even higher, though a capable enough active attacker will still be able to do a MITM attack when he can get a certificate in the DoT's server name. Preventing that attack vector would then require something like certificate pinning. 

As always with security, it is about raising the bar. But a capable enough attacker with enough resources will always be able to circumvent your security...

AdSchellevis:
With a browser trusted certificate, it might work, I'm a bit doubting if it remains functional in the long run as the trust situation from the clients perspective hasn't really improved  in my humble opinion (a change of strategy from Google's end might change the world dramatically). 

Although I certainly miss some context, if there would be a man-in-the-middle, why couldn't he/she do the exact same trick in this case by using a let's encrypt cert and sniff traffic in between? If trust is based on accepting any certificate signed by a trusted party, that shouldn't be too hard. Again, I likely miss some context which might explain why this wouldn't be possible when our end accepts tls from the client. 

The main issue I do have with these additions is that there are multiple constraints involved before people are able to use the feature for which no documentation exists on our end, which has a high risk of creating a false sense of security when not properly used. 

> As always with security, it is about raising the bar. But a capable enough attacker with enough resources will always be able to circumvent your security...

I certainly agree, I'm just missing the part which bar is being raised, other then eaves dropping without intercepting traffic.

doktornotor:
HTTPS == encrypted browser traffic. DoT == encrypted DNS traffic. There's nothing else those certificates have **ever** ensured,

DoT is no less secure than the unencrypted DNS over 53, if anything. Anything else is up to the network admin to implement properly, like with any other feature.  If the firewall admin does not wish to implement this, they simply won't enable the feature. 

I really do not get the point of discussion here.

AdSchellevis:
It's quite simple, if we're not able to explain why a certain feature would be useful and how an administrator should be able to utilize it, we should not consider adding it. Our product is being used by all sorts of people and we aim to keep it accessible.

Unclear features lead to future support nightmares. In order to incorporate this (or any other feature in core), someone from our team will likely need to spend time on it in some way (refactoring code, writing documentation and answering questions). 
People seeking to build advanced setups not supported by our gui are often able to use custom hooks, which is also the case for unbound (https://docs.opnsense.org/manual/unbound.html#advanced-configurations).

So that's the discussion, nothing more, nothing less. It's currently not a priority on our end to work on this, that might change if someone is able to explain suitable use-cases and constraints, you never know. 

So let's try to keep the discussion constructive and try to find valid use-cases in stead of complaining the discussion isn't clear.

doktornotor:
Well, all I can say is that this patch works perfectly fine with DoT-capable clients - as long as the certificate can be validated by the client's OS resolver. That is very much expected and works exactly as designed. Android bugs ignored by Google should not prevent admins from configuring useful DNS resolver features which are working just fine on other OSes (such as Windows 11) via the GUI. 

(As for the hooks, I cannot see an easy way to pass a selected certificate to those.)

definitio:
There is another use case. Firefox doent't support [ECH](https://datatracker.ietf.org/doc/draft-ietf-tls-esni/) without DoH. The only way to use ECH with local DNS resolver is to connect to it via DoH.

jeffreyflynt:
I appreciate your concern about feature clarity and support burden. Let me provide specific use cases and address why GUI support is needed for this existing Unbound capability.

### The Feature Already Exists in Unbound

Unbound natively supports DoH/DoT serving via these configuration options:

- tls-service-key / tls-service-pem / tls-port: 853
- https-port: 443

Users can configure this through custom unbound.conf includes, but this is precisely what creates support issues.

### Primary Use Case: Securing DNS on Untrusted Local Networks

**The Problem**: In multi-tenant buildings (apartments, offices), WiFi traffic can be intercepted by neighbors or other users on the same network segment. DNS queries in plaintext reveal browsing habits, internal hostnames, and can be manipulated.
**The Solution**: Clients configure DoT/DoH to their OPNsense resolver, encrypting queries from device to firewall.
**Administrator Benefit**: Simple security improvement - enable the service and point clients to it. Uses existing OPNsense certificates (Let's Encrypt, etc.).

### Why GUI Support Reduces Your Support Burden

**Current situation**: Users are already trying to do this via custom configs, leading to:

- Support tickets when configurations break after updates
- Certificate path issues and permission problems
- Confusion about which certificates to use
- Custom configs that conflict with GUI-managed settings

**With GUI support**:

- Straightforward checkbox: "Enable DoT/DoH serving"
- Certificate dropdown: Use existing certificate infrastructure
- Clear validation and error messages
- Survives updates without breaking
- Clear documentation of a supported feature path

### Additional Use Cases

**Multi-site deployments**: Remote locations using central OPNsense for DNS filtering without full VPN tunnels. Encrypted queries protect against ISP surveillance.
**Defense-in-depth**: Encrypting DNS even on trusted networks prevents metadata collection by compromised devices on the same VLAN.

### Documentation Would State

Purpose: Encrypt DNS queries from clients to OPNsense (distinct from upstream DoH/DoT)
When to use: Untrusted local networks, remote sites, enhanced privacy

schreibubi:
I've created a PR which adds support for DNR in Kea (#9341). This allows a client to understand where to find an encrypted DNS server.

Issue #6220, opened 2022-12-30
OpenVPN IPv6 Redirect Gateway


tabacha:
**Important notices**

Before you add a new report, we ask you kindly to acknowledge the following:

- [x ] I have read the contributing guide lines at https://github.com/opnsense/core/blob/master/CONTRIBUTING.md
- [x ] I am convinced that my issue is new after having checked both open and closed issues at https://github.com/opnsense/core/issues?q=is%3Aissue

**Is your feature request related to a problem? Please describe.**

We want to get all IPv6 Traffic though OpenVPN. But there is only one Option in the GUI:
"Redirect Gateway" which will only work for IPv4. The option will add a 
`push "redirect-gateway def1"` to openvpn config file.

**Describe the solution you like**

Rename "Redirect Gateway"  to "Redirect IPv4 Gateway" 

Add a new GUI Option: "Redirect IPv6 Gateway" 
which will set `push "redirect-gateway ipv6"` to the openvpn config file.

**Describe alternatives you considered**

Alternative is to add the string: `push "redirect-gateway ipv6"`  to the advanced options. This works for us.

**Additional context**

no

AdSchellevis:
For now the advanced options would do the trick indeed.

We are discussing a re-write of the openvpn legacy module at the moment to offer api support and more align it with the actual configuration logic (https://openvpn.net/community-resources/reference-manual-for-openvpn-2-5/). 

For 23.1 we rewrote the IPsec module, if a similar strategy works for OpenVPN we might consider that as well for a future version ( in which case it's practical to know which options are practical to offer).

AdSchellevis:
https://github.com/opnsense/core/pull/6584

Issue #6135, opened 2022-11-14
OpenVPN auth-gen-token


monkz:
**Important notices**

Before you add a new report, we ask you kindly to acknowledge the following:

- [x] I have read the contributing guide lines at https://github.com/opnsense/core/blob/master/CONTRIBUTING.md
- [x] I am convinced that my issue is new after having checked both open and closed issues at https://github.com/opnsense/core/issues?q=is%3Aissue

**Is your feature request related to a problem? Please describe.**

To allow clients to authenticate with an OTP as part of their password, reneg-sec has to be set to 0.
This allows users to stay connected longer than 1 hour.

But to allow ping-restart auth-gen-token is needed, as the OTP is invalid at this time.

Basically: https://forum.opnsense.org/index.php?topic=25124.0

**Describe the solution you like**

Add auth-gen-token to OpenVPN Server Configuration

**Describe alternatives you considered**
Setting the option in the advanced configuration - but the field says it will be deprecated.

prtrx:
This would be really nice to have. Especially in combination with `auth-gen-token-secret`. This way we could use roaming between multiple OPNsense devices while using passwords that contain OTP tokens.

monkz:
Thanks 🙂

Reiner030:
Hi, 
this feature is not completely implemented.
Because my answer in forum was just one day after this unknown patch I checked my interface why I have overseen this still in last months... 
=>  compared to `renec-sec` it's missing in some PHP web files to get configured:

```
root@fw01:~ # grep -rl reneg-sec /usr/local/
/usr/local/sbin/openvpn
/usr/local/opnsense/mvc/app/models/OPNsense/OpenVPN/OpenVPN.php
/usr/local/opnsense/mvc/app/models/OPNsense/OpenVPN/OpenVPN.xml
/usr/local/opnsense/mvc/app/controllers/OPNsense/OpenVPN/forms/dialogInstance.xml
/usr/local/opnsense/mvc/app/library/OPNsense/OpenVPN/TheGreenBow.php
/usr/local/opnsense/mvc/app/library/OPNsense/OpenVPN/PlainOpenVPN.php
/usr/local/include/openvpn-plugin.h
/usr/local/www/vpn_openvpn_server.php
/usr/local/www/vpn_openvpn_client.php
grep: /usr/local/openssl/misc/tsget: No such file or directory
/usr/local/etc/inc/plugins.inc.d/openvpn.inc
```
[...]

PR #6584, opened 2023-05-29
VPN: OpenVPN: Instances (MVC)

AdSchellevis:
VPN: OpenVPN: Instances - add new module using the same approach as introduced for IPsec in 23.1. Since we likely can't easily migrate the old cruft, we better focus on offering the correct options for openvpn following upstream documentation.

- Implement a solution to keep vpnid's unique so device creation for legacy and mvc can function in similar ways.  (`OpenVPN->usedVPNIds()`)
- Add some of the main "helper" options for clients and servers (`getInstanceById`, `isEnabled`)
- Implement certificate logic, selecting a certificate also implies an authority (which we validate) 
- Hook CRL generation into the exising openvpn_refresh_crls() event 
- Attach already refactored authentication to new MVC as well, `OpenVPN->getInstanceById()` is responsible for feeding the data needed during authentication and overwrite generation. 
- When in client mode and in need for a username+password combination, flush these to file and link in "auth-user-pass" 
- Routes (remote) and push routes (local), combine IPv4 and IPv6 for ease of administration, 
- Keep alive [push] ping-[restart] defined as separate fields for validation
- Add various "push" to client options in Miscellaneous section 
- Add "auth-gen-token" lifetime for https://github.com/opnsense/core/issues/6135 
- Allow selection of redirect-gateway type for https://github.com/opnsense/core/issues/6220 
- Move tls-auth/crypt into separate static keys objects (tab in instances page) 
- Hook existing events (ovpn_event.py) and make sure they locate the server using getServerById() when needed 
- Use `getInstanceById` in openvpn_prepare() to return both legacy as MVC device configuration 
- Add ovpn_service_control.php for service control [stop|start|restart|configure] and glue this in openvpn_services() via configd 
- Change `openvpn_interfaces()` to use `isEnabled()` method on the model to query if any (legacy/mvc) instances are enabled 
- Move `openvpn_config()` from `openvpn.inc` to widget and extend with MVC instances 
- Extend `ovpn_status.py` to parse "instance-" sockets as well, since the filename doesn't explain the role, we're using the status call to figure out the use. uuid's are keys in this case 
- server_id type to str in kill_session.py so we can match either legacy or mvc sockets 
- Hook `ExportController` to `OpenVPN` model using `getInstanceById()` to glue the Client Export utility to both components 
- Extend connection status with mvc sessions (descriptions)

Which should almost complete the new module, things still on the todo list:

- handle `CARP` for client type instances (stop/start?)
- testing 
- documentation

### DIFF
--- a/src/opnsense/mvc/app/models/OPNsense/OpenVPN/Export.xml
+++ b/src/opnsense/mvc/app/models/OPNsense/OpenVPN/Export.xml
@@ -5,7 +5,7 @@
     <items>
         <servers>
             <server type="ArrayField">
-                <vpnid type="IntegerField">
+                <vpnid type="TextField">
                     <Required>Y</Required>
                 </vpnid>
                 <hostname type="HostnameField">
--- a/src/opnsense/scripts/openvpn/ovpn_service_control.php
+++ b/src/opnsense/scripts/openvpn/ovpn_service_control.php
@@ -1,0 +1,183 @@
+#!/usr/local/bin/php
+<?php
+
+/*
+ * Copyright (C) 2023 Deciso B.V.
+ * All rights reserved.
+ *
+ * Redistribution and use in source and binary forms, with or without
+ * modification, are permitted provided that the following conditions are met:
+ *
+ * 1. Redistributions of source code must retain the above copyright notice,
+ *    this list of conditions and the following disclaimer.
+ *
+ * 2. Redistributions in binary form must reproduce the above copyright
+ *    notice, this list of conditions and the following disclaimer in the
+ *    documentation and/or other materials provided with the distribution.
+ *
+ * THIS SOFTWARE IS PROVIDED ``AS IS'' AND ANY EXPRESS OR IMPLIED WARRANTIES,
+ * INCLUDING, BUT NOT LIMITED TO, THE IMPLIED WARRANTIES OF MERCHANTABILITY
+ * AND FITNESS FOR A PARTICULAR PURPOSE ARE DISCLAIMED. IN NO EVENT SHALL THE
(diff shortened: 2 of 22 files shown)


src/opnsense/mvc/app/models/OPNsense/OpenVPN/Export.xml
```diff
@@ -5,7 +5,7 @@
     <items>
         <servers>
             <server type="ArrayField">
-                <vpnid type="IntegerField">
+                <vpnid type="TextField">
```
fichtner, line 8: What's the idea here?

src/opnsense/scripts/openvpn/ovpn_service_control.php
```diff
@@ -0,0 +1,183 @@
+#!/usr/local/bin/php
```
fichtner, line 1: this script looks a bit convoluted (new/different start,stop,etc) but you probably meant to extract bootstrap/setup code from legacy side

src/opnsense/mvc/app/models/OPNsense/OpenVPN/Export.xml
AdSchellevis: mvc uses uuid's

src/opnsense/scripts/openvpn/ovpn_service_control.php
AdSchellevis: yes, I don't want to mangle the existing hooks as this will likely lead to more noise and the "configure" action is required to cleanly apply pending changes (adding the start/[re]start doesn't increase the boilerplate)

PR #6597, opened 2023-06-03
src: remove noodp and noydir, they do not serve any SEO purpose now

williamdes:
Ref: https://searchfacts.com/noodp-noydir-meta-tags/

I did learn this right now, and found all occurrences to remove them as the projects are long gone

### DIFF
--- a/src/etc/inc/authgui.inc
+++ b/src/etc/inc/authgui.inc
@@ -320,7 +320,7 @@ function display_error_form($text)
     <meta charset="UTF-8" />
     <meta http-equiv="X-UA-Compatible" content="IE=edge">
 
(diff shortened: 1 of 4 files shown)


AdSchellevis:
The whole meta tag section might be rather useless for a firewall, I'll leave this one for @fichtner to assess.

williamdes:
> The whole meta tag section might be rather useless for a firewall, I'll leave this one for @fichtner to assess.

IMO it's not because it could be quite easily exposed to internet, so to search engines and end up on Google for people to find instances to bruteforce in an even easier way

right ?

williamdes:
Do I have something to change for this one @fichtner ?

fichtner:
LGTM, thanks!

PR #6598, opened 2023-06-03
Fix some typos

williamdes:
I read quickly all www files with vscode spell checker enabled and found some minor typos. One is in a translation string


AdSchellevis:
@williamdes thanks!

PR #6600, opened 2023-06-04
fix: un-needed I/O

williamdes:
After this change there is still some files that resist because the data is ordered in some other way.
I think that for now I am satisfied and will not fight with the ordering

```
Contents: 10.10.11.0/23
fdba::/120
New contents: fdba::/120
10.10.11.0/23
```

```text
M	/var/db/aliastables/bogonsv6.self.txt
M	/var/db/aliastables/__lan_network.self.txt
M	/var/db/aliastables/__lo0_network.self.txt
M	/var/db/aliastables/bogons.self.txt
```


src/opnsense/scripts/filter/lib/alias/__init__.py
```diff
@@ -181,10 +181,13 @@ def resolve(self, force=False):
                                 self._resolve_content.add(address)
                         # resolve hostnames (async) if there are any in the collected set
                         self._resolve_content = self._resolve_content.union(address_parser.resolve_dns())
-                    # Always save last recorded content to disk, also when we're not responsible for the alias
-                    # so we can use cached results when reloading a single alias.
-                    with open(self._filename_alias_content, 'w') as f_out:
-                        f_out.write('\n'.join(self._resolve_content))
+                    contents_to_write = '\n'.join(self._resolve_content)
+                    # Do not overwrite unchanged contents
+                    if undo_content != contents_to_write:
```
AdSchellevis, line 186: Looking briefly at the code, there might be an issue when the file is not there and the new code is empty, it might be safer to ensure the file is created when it doesn't exist yet (`or not os.path.isfile(self._filename_alias_content)`).

AdSchellevis:
@williamdes left one remark in the alias code, the two other changes I'll leave for @fichtner to assess. Personally I don't mind that much wrapping an `file_exists()` as the `product` instance will likely be called quite regularly. The ipsec pinger I haven't looked at.

williamdes:
> Personally I don't mind that much wrapping an `file_exists()` as the `product` instance will likely be called quite regularly.

Well, I did not find usages of `product_booting`, that did trouble be. Also the code is not that much explicit at what it really does. Else I would love to move this into a dedicated method out of the constructor

> The ipsec pinger I haven't looked at.

I did not have any handy way to disable the hole block if ipsec is not used, and what purpose does serve the file `currentipsecpinghosts` ?
Ref: https://github.com/search?q=org%3Aopnsense%20currentipsecpinghosts&type=code

If someone can explain, maybe we can document it in the code and improve the if conditions
Since IDK deciding if it should be written is complicated ^^


@AdSchellevis Should I extract the change for `/__init__.py` into a new PR for the stable branch ?
If an alias list is quite large in data, the impact could be important. For example multiple MB alias list.
Do you have an idea on how to solve the ordering problem ?
Even after one write it does not remain in the same order, making my check useless for some files
Anyway, I think there is more debug for me to do on this subject, some files seem to escape the check

src/etc/inc/config.inc
```diff
@@ -49,7 +49,11 @@ final class product
 
         /* boot detection */
         $bootflag = '/var/run/booting';
-        @touch($bootflag);
+
+        if (! file_exists($bootflag)) {
```
fichtner, line 53: can you try to adhere to PSR12 style? If you have os-debug(or os-debug-devel) installed you can enforce it via:

# make style-fix

from core repository root dir.

src/sbin/ping_hosts.sh
```diff
@@ -13,9 +13,11 @@
 #  Field 7:  Wan ping time threshold
 #  Field 8:  Address family
 
+IPSECHOSTS="/var/db/ipsecpinghosts"
+
 # Read in ipsec ping hosts and check the CARP status
-if [ -f /var/db/ipsecpinghosts ]; then
-	IPSECHOSTS="/var/db/ipsecpinghosts"
+# Check the file exists and is not empty
+if [ -f $IPSECHOSTS ] && [ -s $IPSECHOSTS ]; then
```
fichtner, line 20: I'd prefer moving this to enabled block below instead:

https://github.com/opnsense/core/blob/a7b1a7053356d82626f1848dde1b57b804286bf1/src/etc/inc/plugins.inc.d/ipsec.inc#L835

src/etc/inc/config.inc
williamdes: Sure, sorry about that
It's coding style habits from the phpMyAdmin coding standard

AdSchellevis:
> @AdSchellevis Should I extract the change for /__init__.py into a new PR for the stable branch ?
> If an alias list is quite large in data, the impact could be important. For example multiple MB alias list.
> Do you have an idea on how to solve the ordering problem ?
> Even after one write it does not remain in the same order, making my check useless for some files
> Anyway, I think there is more debug for me to do on this subject, some files seem to escape the check

Maybe it is better to split it up indeed.
The hash calculation sorts the keys, the downside is that it will take up more computing power for large lists if you will have to compute it twice (which should be avoidable by changing the order of things, but may come at a cost of transparency in the code).

https://github.com/opnsense/core/blob/23dc326bb3e89b95b43a3dfd4e34fd41fef504a3/src/opnsense/scripts/filter/lib/alias/__init__.py#L110

Calculating a hash in all cases is probably not an issue, but types without a parser just may not be written to disk (managed external or via pre_process). As these are usually quite small I don't expect it issues in terms of processing. 

https://github.com/opnsense/core/blob/23dc326bb3e89b95b43a3dfd4e34fd41fef504a3/src/opnsense/scripts/filter/lib/alias/__init__.py#L196

We spend quite some time improving the performance of this component, some of these lists may grow quite large as you noticed as well. Usually the storage part isn't very important as newer nvme/m2 drives survive full disk write cycles for many years nowadays (the once we use specify around 2 full cylces per day for 3 years if I'm not mistaken).

src/etc/inc/config.inc
williamdes: Hi @fichtner 
I do not understand how this should work, it throws missing separator `Makefile:29: *** séparateur manquant. Arrêt.`.
It complains at the line where there is an include (`Mk/defaults.mk`), if I comment it the next line starting with a dot is unhappy.

So I tried to run `../../../mariadb-mysql-kbs/vendor/bin/phpcs --standard=ruleset.xml /mnt/Dev/@williamdes/@forked-repos/@opnsense/core/src/etc/inc/config.inc` and it did output some invalid rules errors

After more research I found out that on Debian you need to install `php-codesniffer`.

So I ran `phpcs --standard=ruleset.xml src/etc/inc/config.inc`

And nothing did pop out for the lines you mentioned. `phpcbf` even said `No fixable errors were found`.

I am using a fresh Debian 12.

I bet the operators spacing could be wrong, but phpcs seems to not report it.

Can you help me ?

fichtner, line 53: might not pick it up indeed... the style is (or at least that is what we use everywhere)

```suggestion
        if (!file_exists($bootflag)) {
```

williamdes: Thanks, I did open a PR to the stable version: https://github.com/opnsense/core/pull/6604

Do I need to include it here too ?
Or do you merge stable into dev ?

fichtner: yes, master please. Sorry for the trouble. Could commit this myself but in the open source spirit external authors are welcome and encouraged :)

williamdes: Well, I force pushed this PR. Still have to fix the comment about ping hosts. But you can probably cherry-pick my commit for the booting file, I will rebase ;)

fichtner: ok done :) 97343e3aae

williamdes: Awesome, thank you !
Will it go to the stable release, is there a procedure ?

src/sbin/ping_hosts.sh
williamdes: Done, I tried to keep the script doing the same behavior than before. Still not knowing what consumes `currentipsecpinghosts`

src/etc/inc/plugins.inc.d/ipsec.inc
```diff
@@ -833,6 +833,8 @@ function ipsec_setup_pinghosts()
     /* get the automatic ping_hosts.sh ready */
     @unlink('/var/db/ipsecpinghosts');
     @touch('/var/db/ipsecpinghosts');
```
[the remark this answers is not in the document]
williamdes: Okay, I pushed another version now that I better understood what you did mention.
It should reduce I/O in general for the ping hosts script, but also when the function is called.
It will stop re-creating the file for nothing. But the ping hosts script still has to cope with an empty file, since from what I did understand the code will not be called to remove it

src/sbin/ping_hosts.sh
williamdes: Okay, I was also thinking that the file put contents would create the file

fichtner: I still think something like https://github.com/opnsense/core/commit/893a7ab5612c25f901c375ead8bd6e627f981fe9 would fix your issue without modifying anything else

williamdes: Ohh, okay, got it

fichtner:
Don't take this the wrong way but I think with the other changes here it's not going to be merged. We would like to avoid regressions and assume the rest of the script works. Modifying anything above the file creation time as 893a7ab561 requires reason and testing with ipsec which is too much for the issue at hand you try to fix.

fichtner:
> Do you know what consumes currentipsecpinghosts or not ?

apparently nothing, but again if you want to fix i/o you can't start messing with ipsec behaviour

williamdes:
> > Do you know what consumes currentipsecpinghosts or not ?
> 
> apparently nothing, but again if you want to fix i/o you can't start messing with ipsec behaviour

got it, and 100% understood
is there anything I should apply ?

fichtner:
Ok, thanks for your work! :)

Issue #6596, opened 2023-06-03
Useless I/O eating up disks


williamdes:
**Important notices**

Before you add a new report, we ask you kindly to acknowledge the following:

- [x] I have read the contributing guide lines at https://github.com/opnsense/core/blob/master/CONTRIBUTING.md
- [x] I am convinced that my issue is new after having checked both open and closed issues at https://github.com/opnsense/core/issues?q=is%3Aissue

**Describe the bug**

A clear and concise description of what the bug is, including last known working version (if any).

*Tip*: to validate your setup was working with the previous version, use opnsense-revert (https://docs.opnsense.org/manual/opnsense_tools.html#opnsense-revert)

**To Reproduce**

1. Setup a normal OPNsense on ZFS
2. Enable logs and stuff into tmpfs
3. Have firewall aliases

**Expected behavior**

Stop re-writing my tables when they did not change

**Describe alternatives you considered**

Fix/reduce the useless writes.

**Screenshots**

The writes are constant
![Sélection_2023060320:55:15001](https://github.com/opnsense/core/assets/7784660/a48d633f-a07f-485d-bcc1-a513f95fb0cb)

The reads are okay
![Sélection_2023060320:55:35001](https://github.com/opnsense/core/assets/7784660/9cba47fc-02c8-431e-b94d-3c1e566f79b0)

**Additional context**

```sh
   	18:32	zfs snapshot zroot/ROOT/default@06jun2023t20h32
   	18:40	zfs snapshot zroot/ROOT/default@06jun2023t20h40
   	18:46	zfs diff zroot/ROOT/default@06jun2023t20h32 zroot/ROOT/default@06jun2023t20h40
```

```text
M	/root
M	/var/db/aliastables/Emails.self.txt
M	/var/db/aliastables/Emails.md5.txt
M	/var/db/aliastables/Emails.txt
M	/var/db/entropy/saved-entropy.3
M	/var/db/aliastables/Bogus_destinations.self.txt
M	/var/db/aliastables/Bogus_destinations.md5.txt
M	/var/db/aliastables/Bogus_destinations.txt
M	/var/db/aliastables/SAFE_HOSTS.self.txt
M	/var/db/aliastables/SAFE_HOSTS.md5.txt
M	/var/db/aliastables/SAFE_HOSTS.txt
M	/var/db/aliastables/xxx_Ambrose.self.txt
M	/var/db/aliastables/xxx_Ambrose.md5.txt
[...]

PR #6632, opened 2023-06-23
fix: I/O on aliases tables

williamdes:
This is not complete, but for now at least the .md5 files are not re-written each time

I am still debugging this to stop all the writes to be done when nothing changes. But not all works

I open this as draft for comments

Ref: https://github.com/opnsense/core/pull/6600#issuecomment-1580757616
Fixes https://github.com/opnsense/core/issues/6596

### DIFF
--- a/src/opnsense/scripts/filter/lib/alias/__init__.py
+++ b/src/opnsense/scripts/filter/lib/alias/__init__.py
@@ -164,6 +164,7 @@ class Alias(object):
             :return: string
         """
         if not self._resolve_content:
+
             if self.expired() or self.changed() or force:
                 alias_file_exists = os.path.isfile(self._filename_alias_content)
                 if alias_file_exists:
@@ -182,6 +183,7 @@ class Alias(object):
                 else:
                     hash_contents = ""
                 unique_id = self.uniqueid()
+                syslog.syslog(syslog.LOG_ERR, 'add %s (%s, %s, %s/%s)' % (self._name, self.expired(), self.changed(), force, hash_contents, unique_id))
                 try:
                     self._resolve_content = self.pre_process()
                     if address_parser:
@@ -190,8 +192,8 @@ class Alias(object):
                                 self._resolve_content.add(address)
                         # resolve hostnames (async) if there are any in the collected set
                         self._resolve_content = self._resolve_content.union(address_parser.resolve_dns())
-                    # Do not overwrite unchanged contents
-                    # Do not overwrite if the hashes are the same
+                    # Do not overwrite if the hashes are the same but write if the file is empty
+                    # TODO: handle length 0 and empty contents
                     if (undo_content == "" or hash_contents != unique_id) or not alias_file_exists:
                         # Always save last recorded content to disk, also when we're not responsible for the alias
                         # so we can use cached results when reloading a single alias.
--- a/src/opnsense/scripts/filter/lib/alias/__init__.py
+++ b/src/opnsense/scripts/filter/lib/alias/__init__.py
@@ -165,35 +165,48 @@ class Alias(object):
         """
         if not self._resolve_content:
             if self.expired() or self.changed() or force:
-                if os.path.isfile(self._filename_alias_content):
+                alias_file_exists = os.path.isfile(self._filename_alias_content)
+                if alias_file_exists:
                     try:
                         undo_content = open(self._filename_alias_content, 'r').read()
                     except UnicodeDecodeError:
                         undo_content = ""
                 else:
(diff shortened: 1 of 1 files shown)


src/opnsense/scripts/filter/lib/alias/__init__.py
```diff
@@ -164,36 +164,51 @@ def resolve(self, force=False):
...
                     syslog.syslog(syslog.LOG_ERR, 'alias resolve error %s (%s)' % (self._name, e))
                     # parse issue, keep data as-is, flush previous content to disk
                     with open(self._filename_alias_content, 'w') as f_out:
                         f_out.write(undo_content)
                     self._resolve_content = set(undo_content.split("\n"))
-                if self.get_parser():
-                    # flush md5 hash to disk
-                    open(self._filename_alias_hash, 'w').write(self.uniqueid())
+                if address_parser:
+                    if hash_contents != unique_id:
```
williamdes, line 209: this works great

src/opnsense/scripts/filter/lib/alias/__init__.py
```diff
@@ -164,36 +164,51 @@ def resolve(self, force=False):
...
+                address_parser = self.get_parser()
+                if os.path.isfile(self._filename_alias_hash):
+                    try:
+                        hash_contents = open(self._filename_alias_hash, 'r').read()
+                    except UnicodeDecodeError:
+                        hash_contents = ""
+                else:
+                    hash_contents = ""
+                unique_id = self.uniqueid()
+                syslog.syslog(syslog.LOG_ERR, 'add %s (%s, %s, %s/%s)' % (self._name, self.expired(), self.changed(), force, hash_contents, unique_id))
```
williamdes, line 186:
```suggestion
```

src/opnsense/scripts/filter/lib/alias/__init__.py
AdSchellevis: should be safe indeed when moving the `address_parser = self.get_parser()` up as you did.

src/opnsense/scripts/filter/lib/alias/__init__.py
```diff
@@ -164,36 +164,51 @@ def resolve(self, force=False):
...
+                alias_file_exists = os.path.isfile(self._filename_alias_content)
+                if alias_file_exists:
                     try:
                         undo_content = open(self._filename_alias_content, 'r').read()
                     except UnicodeDecodeError:
                         undo_content = ""
                 else:
                     undo_content = ""
+                address_parser = self.get_parser()
+                if os.path.isfile(self._filename_alias_hash):
```
AdSchellevis, line 178: I think I would move this into the `try..except` block on top, if content isn't usable, it's safe to assume the hash has no value too. The files we write are also highly unlikely to fail on a unicode decode error.

src/opnsense/scripts/filter/lib/alias/__init__.py
```diff
@@ -164,36 +164,51 @@ def resolve(self, force=False):
...
                                 self._resolve_content.add(address)
                         # resolve hostnames (async) if there are any in the collected set
                         self._resolve_content = self._resolve_content.union(address_parser.resolve_dns())
-                    # Always save last recorded content to disk, also when we're not responsible for the alias
-                    # so we can use cached results when reloading a single alias.
-                    with open(self._filename_alias_content, 'w') as f_out:
-                        f_out.write('\n'.join(self._resolve_content))
+                    # Do not overwrite if the hashes are the same but write if the file is empty
+                    # TODO: handle length 0 and empty contents
+                    if (undo_content == "" or hash_contents != unique_id) or not alias_file_exists:
```
AdSchellevis, line 197: when initializing `hash_contents` with an empty string, I think we only need to compare `hash_contents` with `unique_id` here (as an empty string will also return an md5 value). Which in my opinion would lead to a structure like:

```
hash_contents = ''
alias_file_exists = os.path.isfile(self._filename_alias_content)
if alias_file_exists:
    try {...} except {}  # reading undo_content followed by hash_contents
```

In which case `hash_contents` is always empty if content or hash are either not valid or not available.

williamdes:
Still on my list, did a couple of upgrades since then. Will have to take this from zero and update this PR

AdSchellevis:
@williamdes can you retry OPNsense 24.1? We made some changes in zfs handling, see  https://github.com/opnsense/core/commit/d766ae211c0ab268a038c92c2fedde3bcda96017 .

williamdes:
Sure, I will upgrade my firewalls and report back on writes

PR #6643, opened 2023-06-30
dhcp: migrate leases page to MVC

swhite2:
- reworks the `get_leases.py` script to accomodate the "last declaration is the current one" logic from `dhcpd.leases(5)`. (neither core nor plugins is currently using this action)
- includes `binding` states for failover setups in `dhcpd.py` lease file parsing. These are included in the leases page if `show inactive` is selected.
- adds small utility function in `Firewall/Utility` to check wheter a given IP is within the bounds of a given CIDR block. Needed to bind a lease declaration to an interface in the front end.
- remove `dhcpleaseinlocaltime` property from `services_dhcp.php`. `dhcpd.py` works with epochs, so we only need to convert to a sensible local date in the controller.
- create standard MVC boilerplate to accomodate a very small migration which unsets `dhcpleaseinlocaltime` so it's properly cleaned up.
- implements multiselect lease deletion
- implements multiselect interface filtering
- does not yet hook into `ApiMutableServiceControllerBase`, since the legacy configuration makes it a bit tricky.

@AdSchellevis @fichtner I've skipped `gethostbyaddr` as an alternative for clients that do not specify a hostname (https://github.com/opnsense/core/blob/master/src/www/status_dhcp_leases.php#L176), but this can easily be included. Not sure if this is the way to go here as I'm not able to specify a timeout.


src/opnsense/mvc/app/models/OPNsense/DHCP/ACL/ACL.xml
```diff
@@ -0,0 +1,28 @@
+<acl>
+    <page-services-dhcprelay>
```
fichtner, line 2: movement seems a bit inconsistent: either move all dhcp related things or just the new MVC pages here

src/opnsense/mvc/app/models/OPNsense/DHCP/DHCP.xml
```diff
@@ -0,0 +1,6 @@
+<model>
+    <mount>//OPNsense/DHCP</mount>
+    <description>DHCP configuration</description>
+    <version>1.0.1</version>
```
fichtner, line 4: Why 1.0.1? :)

src/opnsense/mvc/app/models/OPNsense/DHCP/Menu/Menu.xml
```diff
@@ -0,0 +1,9 @@
+<menu>
+    <Services>
+        <DHCPv4 cssClass="fa fa-bullseye fa-fw">
+            <Relay order="200" url="/services_dhcp_relay.php"/>
```
fichtner, line 4: same here about inconsistency

src/opnsense/scripts/dhcp/get_staticmap.php
```diff
@@ -0,0 +1,34 @@
...
+ * CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE)
+ * ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
+ * POSSIBILITY OF SUCH DAMAGE.
+ */
+
+require_once 'config.inc';
+require_once 'interfaces.inc';
+require_once 'plugins.inc.d/dhcpd.inc';
+
+echo json_encode(dhcpd_staticmap("not.found", legacy_interfaces_details(), false, 4), JSON_PRETTY_PRINT) . PHP_EOL;
```
fichtner, line 34: better not hardcode "4" here if we plan to reuse it

src/opnsense/service/conf/actions.d/actions_dhcpd.conf
```diff
@@ -1,9 +1,15 @@
...
+command:/usr/local/opnsense/scripts/dhcp/get_leases.py
+parameters:/inactive %s /duplicates %s
 type:script_output
 message:list dhcp leases %s
 
+[list.static]
+command:/usr/local/opnsense/scripts/dhcp/get_staticmap.php
+parameters:
+type:script_output
+message: list dhcp static mappings %s
```
fichtner, line 11: `%s` only if param which doesn't exist yet

src/opnsense/mvc/app/models/OPNsense/DHCP/Migrations/M1_0_0.php
```diff
@@ -0,0 +1,43 @@
...
+ use OPNsense\Base\BaseModelMigration;
+ use OPNsense\Core\Config;
+
+ class M1_0_0 extends BaseModelMigration
+ {
+    public function post($model)
+    {
+        $config = Config::getInstance()->object();
+        foreach ($config->dhcpd->children() as $dhcpif => $dhcpifconf) {
+            unset($dhcpifconf->dhcpleaseinlocaltime);
```
fichtner, line 40: if this is all it's doing I'd vote in favour of ignoring this in the config like we do for most other things that are not being used anymore from static pages

src/opnsense/mvc/app/models/OPNsense/DHCP/DHCP.xml
fichtner: Ok I get it...this is a DHCPv4 model. Probably going to be confusing later on.. DHCP... DHCPv6 ... DHCPv4New ... DHCPv6New... we should be careful with the name as we can't reuse it later when we rewrite DHCP based on some other software.

src/opnsense/service/conf/actions.d/actions_dhcpd.conf
```diff
@@ -1,9 +1,15 @@
 [list.leases]
-command:/usr/local/opnsense/scripts/dhcp/get_leases.py /inactive %s
-parameters:%s
+command:/usr/local/opnsense/scripts/dhcp/get_leases.py
+parameters:/inactive %s /duplicates %s
 type:script_output
 message:list dhcp leases %s
 
+[list.static]
+command:/usr/local/opnsense/scripts/dhcp/get_staticmap.php
```
fichtner, line 8: BTW, I'm quite sure `pluginctl -r static_mapping` is shorter. we can figure out a solution for only grabbing ipv4. the rest ist standard arguments.

fichtner (approved):
ok, but might be more than necessary and namespace needs to be future-proof. see individual comments

src/opnsense/mvc/app/controllers/OPNsense/DHCP/Api/LeasesController.php
```diff
@@ -0,0 +1,214 @@
...
+                $leases[$idx]['ends'] = date('Y/m/d H:i:s', $lease['ends']);
+            }
+
+            if (array_key_exists('client-hostname', $lease)) {
+                $leases[$idx]['hostname'] = $lease['client-hostname'];
+            }
+        }
+
+        $macs = [];
+        foreach ($leases as $idx => $lease) {
```
AdSchellevis, line 117: move up to save a loop ?

src/opnsense/mvc/app/controllers/OPNsense/DHCP/Api/LeasesController.php
```diff
@@ -0,0 +1,214 @@
...
+                case 'backup':
+                    $leases[$idx]['state'] = 'backup';
+                    break;
+            }
+
+            if (array_key_exists('hardware', $lease)) {
+                $mac = $lease['hardware']['mac-address'];
+                $leases[$idx]['mac'] = $mac;
+                $leases[$idx]['status'] = $onlineStatus($lease['address']);
+                unset($leases[$idx]['hardware']);
```
AdSchellevis, line 100: `$macs[$mac][] = $idx;`  (see comment in next loop)

AdSchellevis (changes requested):

src/opnsense/service/conf/actions.d/actions_dhcpd.conf
```diff
@@ -1,9 +1,15 @@
 [list.leases]
-command:/usr/local/opnsense/scripts/dhcp/get_leases.py /inactive %s
-parameters:%s
+command:/usr/local/opnsense/scripts/dhcp/get_leases.py
+parameters:/inactive %s /duplicates %s
```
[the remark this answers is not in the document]
swhite2: Woops, leftover argument. This is a bit of a 50/50 call to be honest, don't expect there to be a lot of callers for this but I'd also prefer cutting logic in the controller where possible. Also active/inactive is fairly generalized information coming from the leases file so I don't think it's out of place.

src/opnsense/mvc/app/controllers/OPNsense/DHCP/Api/LeasesController.php
```diff
@@ -0,0 +1,167 @@
...
+        }
+
+        /* get configured static leases */
+        $sleases = json_decode($backend->configdRun('dhcpd list static'), true);
+
+        /* include inactive leases if requested */
+        $leases = json_decode($backend->configdpRun('dhcpd list leases', [$inactive]), true);
+        foreach ($leases as $idx => $lease) {
+            /* set defaults */
+            $leases[$idx]['type'] = 'dynamic';
```
swhite2, line 63: @AdSchellevis should `type` and `state` be wrapped in a `gettext()`?

src/opnsense/service/conf/actions.d/actions_dhcpd.conf
swhite2: The current construction is fairly simple and explicit, If this has the potential of being shorter it would likely require some combination of named parameters and a single argument to https://github.com/opnsense/core/blob/master/src/etc/inc/plugins.inc.d/dhcpd.inc#L44,  not entirely sure how this would fit into `pluginctl`.

fichtner: Moving family arg of dhcpd_staticmap to first would be an option. Also family is pretty easy to distinguish between returned ipaddr vs. ipaddrv6

src/opnsense/mvc/app/library/OPNsense/Firewall/Util.php
```diff
@@ -339,6 +339,22 @@ public static function calcRuleHash($rule)
         return md5(json_encode($rule));
     }
 
+    /**
+     * returns whether a given IP is in a CIDR block
+     */
+    public static function isIPInCIDR($ip, $cidr)
```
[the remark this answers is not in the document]
swhite2: @AdSchellevis How about https://github.com/opnsense/core/pull/6643/commits/c10b366fc0ff657a1c34e8d2c10c0b2625de5a94

src/opnsense/mvc/app/controllers/OPNsense/DHCP/Api/LeasesController.php
AdSchellevis: @swhite2 a bit late to the party, small issue with my mail, but I don't mind keeping it untranslated for now, from an API perspective its better to keep values untranslated.

src/opnsense/mvc/app/library/OPNsense/Firewall/Util.php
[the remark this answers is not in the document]
AdSchellevis: @swhite2 looks good

src/opnsense/service/conf/actions.d/actions_dhcpd.conf
swhite2: @fichtner how about https://github.com/opnsense/core/pull/6643/commits/d38ab331d0f745e734358ccf01b200d8480046f3?

fichtner: Looks good, thanks!

PR #6653, opened 2023-07-06
dhcp6: migrate leases page to MVC

swhite2:
WIP for dhcpv6 leases.


src/opnsense/scripts/dhcp/get_leases6.py
```diff
@@ -0,0 +1,186 @@
...
+    lease['iaid'] = iaid
+    lease['duid'] = duid
+
+    for line in lines:
+        parts = line.split()
+
+        cltt = None
+        if parts[0] == 'cltt' and len(parts) >= 3:
+            cltt = parse_date(parts[2], parts[3])
+        if cltt is not None:
```
AdSchellevis, line 126: indent and ditch the `cltt = None`?

src/opnsense/scripts/dhcp/get_leases6.py
```diff
@@ -0,0 +1,186 @@
...
+import sys
+sys.path.insert(0, "/usr/local/opnsense/site-python")
+import ujson
+import calendar
+import datetime
+import params
+import time
+
+def parse_date(ymd, hms):
+    result = None
```
AdSchellevis, line 40: unused, but when returning from the `try..except` you could save some lines too

src/opnsense/scripts/dhcp/get_leases6.py
```diff
@@ -0,0 +1,186 @@
...
+    """
+    parse either an iaaddr or iaprefix segment. return a tuple
+    containing the type parsed and the corresponding segment
+    """
+    seg_type = input[0].split()[0]
+    segment = dict()
+    segment[seg_type] = input[0].split()[1]
+    for line in input[1:]:
+        parts = line.split()
+        field_name = parts[0]
```
AdSchellevis, line 59: not sure if this can happen, but an empty string would crash out.  to prevent this from happening, a construct like the following might help:

```
field_name = parts[0] if len(parts) > 0 else ''
```

src/opnsense/scripts/dhcp/get_leases6.py
```diff
@@ -0,0 +1,186 @@
...
+                parsed.append("%02x" % octal_to_decimal)
+                i += 3
+        else:
+            parsed.append("%02x" % ord(c))
+        i += 1
+
+    iaid = int(''.join(reversed(parsed[0:4])))
+    duid = ":".join([str(a) for a in parsed[4:]])
+    iaid_duid = ":".join([str(a) for a in parsed])
+    return (iaid_duid, iaid, duid)
```
AdSchellevis, line 103: I think I would return a `dict` here, so you can update the lease directly (`lease.update()`)

src/opnsense/scripts/dhcp/get_leases6.py
```diff
@@ -0,0 +1,186 @@
...
+    if addresses:
+        lease['addresses'] = addresses
+    if prefixes:
+        lease['prefixes'] = prefixes
+
+    return (iaid_duid, lease)
+
+if __name__ == '__main__':
+    app_params = {'inactive': '0'}
+    params.update_params(app_params)
```
AdSchellevis, line 152: I would opt for `argparse` here, I know I added the "params" a long time ago, but argparse is cleaner and removes the need for the custom `sys.path.insert(0, "/usr/local/opnsense/site-python")`

AdSchellevis (changes requested):

src/opnsense/scripts/dhcp/get_leases6.py
swhite2: i agree, this was quick & dirty - but argparse prefers boolean interpretations as either '--value' or nothing, which doesn't really translate well to configd action parameters. I'll update it to a string argument in argparse.

src/opnsense/scripts/dhcp/get_leases6.py
swhite2: unlikely, but you know what they say about assumptions..

src/opnsense/mvc/app/controllers/OPNsense/DHCPv6/Api/LeasesController.php
```diff
@@ -0,0 +1,245 @@
...
+                    $intf = (string)$if->descr;
+                    $leases[$idx]['if_descr'] = $intf;
+                    $leases[$idx]['if'] = $lease['if'];
+
+                    if (!array_key_exists($lease['if'], $interfaces)) {
+                        $interfaces[$lease['if']] = $intf;
+                    }
+                }
+            } else {
+                foreach ($config->dhcpdv6->children() as $dhcpif => $dhcpifconf) {
```
AdSchellevis, line 170: As the number of `dhcpdv6` servers is highly likely limited, this shouldn't be an issue, but otherwise you would like to store the interface ranges in a separate loop to prevent a cartesian product.

src/opnsense/mvc/app/controllers/OPNsense/DHCPv6/Api/LeasesController.php
```diff
@@ -0,0 +1,245 @@
...
+                $prefix['prefix'] = $prefix_raw['iaprefix'];
+                if (array_key_exists('ends', $prefix_raw)) {
+                    $prefix['ends'] =  date('Y/m/d H:i:s', $prefix_raw['ends']);
+                }
+                $prefix['state'] = $prefix_raw['binding'] == 'free' ? 'expired' : $prefix_raw['binding'];
+                $prefixes[] = $prefix;
+            }
+        }
+
+        $response = $this->searchRecordsetBase($prefixes, null, 'prefix', null, SORT_REGULAR);
```
AdSchellevis, line 226: `return $this->searchRecordsetBase()` ? storing the value likely doesn't make much sense here.

AdSchellevis (approved):

src/opnsense/mvc/app/controllers/OPNsense/DHCPv6/Api/ServiceController.php
```diff
@@ -0,0 +1,101 @@
...
+
+namespace OPNsense\DHCPv6\Api;
+
+use OPNsense\Base\ApiControllerBase;
+use OPNsense\Core\Backend;
+
+class ServiceController extends ApiControllerBase
+{
+    /**
+     * XXX most of this logic can be replaced when appropriate start/stop/restart/status
```
fichtner, line 37: Since we have all these configd hooks now it would make sense to add this to the mutable service controller as an option similar to force restart override. It doesn’t get more generic than this and we can use it for other parts pretty soon as well.

fichtner (approved):
Thanks, neat!

src/opnsense/mvc/app/controllers/OPNsense/DHCPv6/Api/ServiceController.php
swhite2: Good idea, though we would likely have to conditionally drop https://github.com/opnsense/core/blob/8dc8d39d4bdae3b8cfaf5e99be26af1258ef0f95/src/opnsense/mvc/app/controllers/OPNsense/Base/ApiMutableServiceControllerBase.php#L72-L87

swhite2: As this isn't exclusive to DHCP, I'll implement this after the merge. Thanks :)

Issue #6614, opened 2023-06-13
Unbound can´t start due to special characters in the hostname


Tharguard:
**Important notices**

Before you add a new report, we ask you kindly to acknowledge the following:

- [x] I have read the contributing guide lines at https://github.com/opnsense/core/blob/master/CONTRIBUTING.md
- [x] I am convinced that my issue is new after having checked both open and closed issues at https://github.com/opnsense/core/issues?q=is%3Aissue

**Describe the bug**

An IoT-device (Brother sewing machine Innovis NV880E) has special characters in it´s hostname: 56;2;1;1.53
I think its the ";" sign
If i switch on the device on it gets an IP-address from dhcp and then unbound service works until i restart the service, 
the restart fails then.

**To Reproduce**

Turn device on

**Expected behavior**

Unbound works flawless ;)

**Describe alternatives you considered**

Assigning a hostname via DHCP is not the solution.

**Screenshots**

**Relevant log files**

2023-06-11T13:19:11 Critical unbound [45340:0] fatal error: Could not set up local zones
2023-06-11T13:19:11 Error unbound [45340:0] error: Bad local-data RR 56;2;1;1.60.local IN A 10.10.10.74
2023-06-11T13:19:11 Error unbound [45340:0] error: error parsing local-data at 34 '56;2;1;1.60.local IN A 10.10.10.74': Syntax error, could not parse the RR's TTL

**Additional context**

If i delete in /var/unbound/dhcpleases.conf this two lines unbound service starts again:

local-data-ptr: "10.10.10.74 56;2;1;1.53.local"
local-data: "56;2;1;1.53.local IN A 10.10.10.74"

**Environment**

Software version used and hardware type if relevant, e.g.:

OPNsense 23.1.9-amd64
FreeBSD 13.1-RELEASE-p7
OpenSSL 1.1.1t 7 Feb 2023

swhite2:
Nice catch, invalid hostnames exist out in the wild and we should ignore them.

Can you try `# opnsense-patch 02aae56f0 e43c23fce`?

Tharguard:
> Nice catch, invalid hostnames exist out in the wild and we should ignore them.
> 
> Can you try `# opnsense-patch 02aae56f0 e43c23fce`?

Thanks, this patch works ;)

fichtner:
@swhite2 neat... would you mind applying something similar to dnsmasq watcher too?

swhite2:
@fichtner on it

PR #6665, opened 2023-07-13
Allow underscores in DNS names from DHCP leases

bugfixin:
Currently DHCP leases with DNS names including underscores cause a warning "dhcpd leases: XXX_XXX not a valid hostname, ignoring" 

While these may not be true internet-valid hostnames according to IETF, many valid DNS names include underscores (e.g. _acme-challenge for ACME/letsencrypt validation, microsoft active directory names, SRV records, and manufacturer-default IP camera names, though I'm sure there are others). 

Blindly denying underscores for DNS records is likely to cause unforeseen issues, and I don't believe it's in our best interest to police names to that degree.

### DIFF
--- a/src/opnsense/scripts/dhcp/dnsmasq_watcher.py
+++ b/src/opnsense/scripts/dhcp/dnsmasq_watcher.py
@@ -49,7 +49,7 @@ def run_watcher(target_filename, default_domain, watch_file, service_pid):
     # initiate lease watcher and setup cache
     dhcpdleases = watchers.dhcpd.DHCPDLease(watch_file)
     cached_leases = dict()
-    hostname_pattern = re.compile("(?!-)[A-Z0-9-]*(?<!-)$", re.IGNORECASE)
+    hostname_pattern = re.compile("(?!-)[A-Z0-9-_]*(?<!-)$", re.IGNORECASE)
 
     # start watching dhcp leases
     last_cleanup = time.time()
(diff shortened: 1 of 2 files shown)


AdSchellevis:
The current constraints might be a bit stricter then needed (introduced in https://github.com/opnsense/core/commit/02aae56f05f313e7a5bc59d1e2551864dbea9a12 due to https://github.com/opnsense/core/issues/6614), but someone does need to validate if allowing those will not lead to similar issues as those we started with. 

Strictly speaking rfc1123 does not allow underscores in hostnames if I'm not mistaken (https://datatracker.ietf.org/doc/html/rfc1123), so if the applications underneath (unbound/dnsmasq) are strict, they might lead to issues, in which case we should not try to feed those.

bugfixin:
Per RFC 2181 section 11: https://www.rfc-editor.org/rfc/rfc2181#section-11

> Occasionally it is assumed that the Domain Name System serves only
   the purpose of mapping Internet host names to data, and mapping
   Internet addresses to host names.  This is not correct, the DNS is a
   general (if somewhat limited) hierarchical database, and can store
   almost any kind of data, for almost any purpose.

 >  The DNS itself places only one restriction on the particular labels
   that can be used to identify resource records.  That one restriction
   relates to the length of the label and the full name.

In other words, if built to the RFCs, DNS itself doesn't care if a DNS record has underscores or whether it's a valid "hostname" per se - just that it isn't too long. I have tested unbound and dnsmasq and they happily serve records with underscores but I'll update this PR to include the same change to dnsmasq's sync script.

AdSchellevis:
@bugfixin thanks for testing and updating both. The restriction is there for good reasons, but that certainly doesn't mean we should not allow underscores if both applications do not mind. We'll take it under consideration.

bugfixin:
Thank you - while per RFC 1123 extensibility section, DNS *should* allow any label containing up to 63 8-bit octets without care for validity, I agree that it makes sense to prevent certain characters that could cause problems.

I guess my host named 💩 will have to wait.

AdSchellevis:
@bugfixin discussed internally, we'll merge this one. Thanks!

PR #6683, opened 2023-07-22
interfaces: allow multiple ip addresses in dhcp reject from validation and improve help text

csabakos:
Fixes an issue with the help text for `Reject Leases From` in the Interfaces page, where according to the help text, the `Reject Leases From` value can be an IP address or a subnet, however in reality subnets are not supported by the dhclient version that opnsense uses (but lists of IP addresses are).

This MR updates the help text and the validation to match what dhclient accepts.

### DIFF
--- a/src/www/interfaces.php
+++ b/src/www/interfaces.php
@@ -327,6 +327,28 @@ function get_wireless_channel_info($interface)
     return $wireless_channels;
 }
 
+/**
+ * @return true if $value is a comma separated list of valid dotted IPv4
+ * addresses, or false otherwise.
+ *
+ * Spaces around the IPv4 addresses are allowed.
+ */
+function is_ipaddrlistv4($value)
+{
+    if (!is_string($value)) {
+        return false;
+    }
+
+    $list = explode(',', $value);
+    foreach ($list as $item) {
+        $trimmed_item = trim($item);
+        if (!is_ipaddrv4($trimmed_item)) {
+            return false;
+        }
+    }
+    return true;
+}
+
 $ifdescrs = legacy_config_get_interfaces(['virtual' => false]);
 $hwifs = array_keys(get_interface_list());
 
@@ -845,8 +867,8 @@ if ($_SERVER['REQUEST_METHOD'] === 'GET') {
         if (!empty($pconfig['alias-subnet']) && !is_numeric($pconfig['alias-subnet'])) {
             $input_errors[] = gettext("A valid alias subnet bit count must be specified.");
         }
-        if (!empty($pconfig['dhcprejectfrom']) && !is_ipaddrv4($pconfig['dhcprejectfrom'])) {
-            $input_errors[] = gettext("A valid alias IP address must be specified to reject DHCP Leases from.");
+        if (!empty($pconfig['dhcprejectfrom']) && !is_ipaddrlistv4($pconfig['dhcprejectfrom'])) {
+            $input_errors[] = gettext("A valid IP address or comma separated IP address list must be specified to reject DHCP Leases from.");
         }
 
         if ($pconfig['gateway'] != "none" || $pconfig['gatewayv6'] != "none") {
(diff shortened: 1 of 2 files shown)


fichtner:
And you did test this? Because if you actually use the FreeBSD man page you will see it doesn’t.

csabakos:
> And you did test this? Because if you actually use the FreeBSD man page you will see it doesn’t.

Thanks for pointing that out. I only verified that the `.conf` file was written as expected, but you are right that the version included in opnsense doesn't support subnets, only a list of IP addresses ([reference](https://github.com/opnsense/src/blob/stable/23.1/sbin/dhclient/clparse.c#L930)).

I'll rework this to only accept an IP list and update the help text accordingly.

csabakos:
Updated

src/etc/inc/util.inc
```diff
@@ -720,6 +720,28 @@ function is_subnetv6($subnet)
     return (is_subnet($subnet) == 6);
 }
 
+/**
+ * @return true if $value is a comma separated list of valid dotted IPv4
+ * addresses, or false otherwise.
+ *
+ * Spaces around the IPv4 addresses are allowed.
+ */
+function is_ipaddrlistv4($value)
```
fichtner, line 729: Can we avoid this and just move to handle this in interfaces.php? The MVC framework supports CSV natively so we would want to avoid increasing "library" load for functions that are never going to be reused.

src/www/interfaces.php
```diff
@@ -2280,7 +2280,7 @@ function toggle_wirelesscfg() {
                           <td>
                             <input name="dhcprejectfrom" type="text" id="dhcprejectfrom" value="<?=htmlspecialchars($pconfig['dhcprejectfrom']);?>" />
                             <div class="hidden" data-for="help_for_dhcprejectfrom">
-                              <?=gettext("If there is a certain upstream DHCP server that should be ignored, place the IP address or subnet of the DHCP server to be ignored here."); ?>
+                              <?=gettext("If there are certain upstream DHCP servers that should be ignored, place the comma separated list of one or more IP addresses of the DHCP server(s) to be ignored here."); ?>
```
fichtner, line 2283: A bit to uncertain about itself here with conditionals and "(s)" use. It's probably better to rephrase this in a snappier way.

fichtner (changes requested):
I had to check dhclient(8) in src.git since the man page is not specific about lists but `parse_reject_statement()` indeed supports CSV input. However, I have change requests here to avoid code bloat.

src/etc/inc/util.inc
csabakos: Thank you for the feedback!

I could easily move this to `interfaces.inc` but not to `interfaces.php` since the function is used from `interfaces.inc` too. Does that sound good?

> The MVC framework supports CSV natively so we would want to avoid increasing "library" load for functions that are never going to be reused.

Could you elaborate on what you mean by the MVC framework supporting CSV natively? Is there a better way to check that something is a comma separated list of IPv4 addresses than the `explode()` and loop solution below?

fichtner: No extra functions, no validation in interfaces.inc -- this is far too much effort than useful. I've mentioned the MVC pages because all static PHP pages are going to be converted at some point so most extra effort on static PHP pages will either lead to unnecessary work or perhaps also regressions.

fichtner: > no validation in interfaces.inc 

to clarify the old code does this but it should not

csabakos: Updated

src/www/interfaces.php
csabakos: Simplified

fichtner:
I merged a modified versions due to consistency concerns https://github.com/opnsense/core/commit/1f3311f5dd45784fe6802a2d150776dc75a6c591

PR #6686, opened 2023-07-26
Makefile: move packages from CORE_DEPENDS_amd64 to CORE_DEPENDS

yrzr:
The four packages including py${CORE_PYTHON}-duckdb,  py${CORE_PYTHON}-numpy , py${CORE_PYTHON}-pandas , and suricata-devel, can be built on aarch64. So no longer need to put them under amd64 only.


fichtner:
this will probably break the 32 bit arm build if anyone is still doing that

Makefile
```diff
@@ -126,10 +126,6 @@ CORE_COPYRIGHT_WWW?=	https://www.deciso.com/
 CORE_COPYRIGHT_YEARS?=	2014-2023
 
 CORE_DEPENDS_amd64?=	beep \
-			py${CORE_PYTHON}-duckdb \
```
fichtner, line 129: not sure which CORE_ARCH is exposed on your end, but we could move them to this CORE_DEPENDS_xxx variant and then inline this variant into CORE_DEPENDS_amd64

fichtner:
Merged, thanks!

maurice-w:
Will this make it into 23.7 or 23.7.x?

fichtner:
Will add this to stable/23.7 tomorrow before building the final 23.7

PR #6705, opened 2023-08-04
Update opnsense.xml

jeremiah-rs:
Updated LeaseWeb and NYC BUG mirrors to use HTTPS


fichtner:
Will require testing before shipping but these mirrors now have stable HTTPS indeed. That wasn't always the case. ;) Thanks!

PR #6709, opened 2023-08-04
Allow use of seperate CA for validation of OpenVPN clients

Skywalker-11:
With the new OpenVPN interface of 23.7 the option to use a specific CA for verification of the client certificates is no longer available. 
This patch add the option to specify a specific user CA that should be used for authentication of the clients tls certificates. The setting is optional and if absent it falls back to using the CA of the server certificate for the OpenVPN `<ca>` option. If verification of client certificates is enabled and the server certificate doesn't include a CA and the CA client certificate doesn't exist an error is displayed. The CA server certificate which is currently in 23.7 is used as the `<ca>` option is instead defined as `<extra-certs>` to be able to complete its certificate chain.

Since the webpki develops to require or at least recommend separate CAs for server certificates and user certificates this option should still be available.

### DIFF
--- a/src/opnsense/mvc/app/controllers/OPNsense/OpenVPN/forms/dialogInstance.xml
+++ b/src/opnsense/mvc/app/controllers/OPNsense/OpenVPN/forms/dialogInstance.xml
@@ -140,6 +140,12 @@
         <help>Select a certificate to use for this service.</help>
     </field>
     <field>
+        <id>instance.client_ca</id>
+        <label>CA Client Authentication</label>
+        <type>dropdown</type>
+        <help>Select a ca certificate for client validation</help>
+    </field>
+    <field>
         <id>instance.crl</id>
         <label>Certificate Revocation List</label>
         <type>dropdown</type>
--- a/src/opnsense/mvc/app/models/OPNsense/OpenVPN/OpenVPN.php
+++ b/src/opnsense/mvc/app/models/OPNsense/OpenVPN/OpenVPN.php
@@ -74,6 +74,7 @@ class OpenVPN extends BaseModel
                     );
                 }
             }
+
             if (!empty((string)$instance->cert)) {
                 if ($instance->cert->isFieldChanged() || $validateFullModel) {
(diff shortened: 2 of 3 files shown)


AdSchellevis:
We left it out for good reasons, I don't mind re-assessing earlier choices,  but clear references are needed why one should do this.

Skywalker-11:
Our users generally get a client certificate that is trusted by a CA that is technically constrained to only issue client/user certificates. The users currently use this certificate for different authentication purposes (VPN, Wifi, Webapps etc.). As such certificates from that CA cannot issue server certificates for the vpn server and also using a self signed server certificate generated by the Opnsense firewall is not an option with the currently available interface. There is also a new S/MIME Baseline Requirement for public CAs beeing introduced that probably will further lead CAs to differentiate between server certificates and user certificates.

It is bad practice to use the same certificate for different devices/users. The current GUI of Opnsense is not designed in a way that would allow it be a portal where users would be able to generate or renew their own access certificates. As such every such certificate would have to be generated and then somehow distributed to the clients. Especially since the VPN connection would be the way to do this securely via automation this is not a feasable option for us.

AdSchellevis:
Let's agree to disagree, but with clear references I don't mind reevaluating previous choices as mentioned earlier, looking at OpenVPN's own documentation both should normally be using the same CA (https://openvpn.net/vpn-server-resources/ca-certificate-management-with-openvpn-access-server/ , https://openvpn.net/community-resources/how-to/)

We haven't seen other valid options in years, if there are, we can always reassess, but only when it's very clear which problem the additional selection would be a fix for (best practice documentation from well known institutes or rfc's would help in that case).

fichtner:
For reference ovpn.com seemed to indicate that they need this https://www.reddit.com/r/opnsense/comments/15z4uvv/comment/jxiqvyk/?utm_source=reddit&utm_medium=web2x&context=3 although the wording/discussion there is a bit fuzzy

AdSchellevis:
https://github.com/opnsense/core/commit/27e723a00ed8e89c55d060c4be72ecf88780061b

PR #6724, opened 2023-08-09
ntpd: fix 'Soliciting pool server'

aque:
From ntp.conf man page:
if you want to use servers from a pool directive and also want to use nopeer by default, you'll want a restrict source ... line as well that does not include the nopeer directive.

This PR reorders ntpd.inc to remove `nopeer` from the `restrict source` line.

[Forum discussion](https://forum.opnsense.org/index.php?topic=35311.0)


fichtner:
Merged, thanks! Will hotfix.

Issue #6744, opened 2023-08-14
Suricata - please add "bypass" parameter for User defined rules


Monviech:
**Important notices**

Before you add a new report, we ask you kindly to acknowledge the following:

- [x] I have read the contributing guide lines at https://github.com/opnsense/core/blob/master/CONTRIBUTING.md
- [x] I am convinced that my issue is new after having checked both open and closed issues at https://github.com/opnsense/core/issues?q=is%3Aissue

**Is your feature request related to a problem? Please describe.**

Local routed traffic in the opnsense is double inspected by suricata. When NAT is used, the official docs state that the "Interfaces" in "Services: Intrusion Detection: Administration: Settings" should be internal networks. "WAN" shouldn't be selected.
This works well if there is only "WAN" and "LAN".
As soon as there is "WAN", "LAN" and "DMZ", and "LAN", "DMZ" are selected as "Interfaces", the traffic between "LAN" and "DMZ" is inspected when it leaves "LAN" and when it enters "DMZ". This impacts the transmission speed during both inspections.
Creating options for the user to bypass local routed traffic if they only want IPS from/to "$INTERNAL_NET" to/from "$EXTERNAL_NET" would create a huge speed benefit and reduce wasted CPU time.

https://docs.suricata.io/en/suricata-6.0.1/rules/bypass-keyword.html
https://docs.suricata.io/en/suricata-6.0.0/performance/ignoring-traffic.html
https://www.stamus-networks.com/blog/2016/09/28/suricata-bypass-feature

**Describe the solution you like**

I would like an input field in `Services: Intrusion Detection: Administration: User defined` which would add `bypass;` to `/usr/local/etc/suricata/opnsense.rules/OPNsense.rules`.
Rules would look like this:
`pass ip   10.16.1.0/24  any ->  172.16.0.0/24  any (msg:"bypass from 10.16.1.0 to 172.16.0.0"; bypass; sid:1; rev:1;)`

The option would add the `bypass;` parameter to a custom `User defined` rule.

**Describe alternatives you considered**

In `Services: Intrusion Detection: Administration: Settings` there could be an option in the `advanced mode` below `Home networks` which reads `Bypass local traffic from inspection`. The help text would read `Only inspect traffic between "WAN" and "Home networks", bypass inspection between networks defined in "Home networks" completely`.

The setting would read the networks in "Home networks" and create automatic rules between them.

[...]

PR #6752, opened 2023-08-16
"bypass" keyword in suricata user defined rules

Monviech:
Ref: #6744

Adds a new input field in `Services: Intrusion Detection: Administration: User defined` which adds `bypass;` keyword to `/usr/local/etc/suricata/opnsense.rules/OPNsense.rules`.

grafik

Rules are generated with the template `src/opnsense/service/templates/OPNsense/IDS/OPNsense.rules `and look like this when the checkbox is enabled:
`pass ip   10.0.0.0/8  any ->  10.0.0.0/8  any (msg:"bypass 10.0.0.0/8 to 10.0.0.0/8"; bypass; sid:1; rev:1;)`

### DIFF
--- a/src/opnsense/mvc/app/models/OPNsense/IDS/IDS.xml
+++ b/src/opnsense/mvc/app/models/OPNsense/IDS/IDS.xml
@@ -1,6 +1,6 @@
 <model>
     <mount>//OPNsense/IDS</mount>
-    <version>1.0.7</version>
+    <version>1.0.8</version>
     <description>
         OPNsense IDS
     </description>
@@ -115,7 +115,7 @@
                 </action>
                 <bypass type="BooleanField">
                         <default>0</default>
-                        <Required>N</Required>
+                        <Required>Y</Required>
                 </bypass>
             </rule>
         </userDefinedRules>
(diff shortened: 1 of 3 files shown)


src/opnsense/mvc/app/models/OPNsense/IDS/IDS.xml
```diff
@@ -113,6 +113,10 @@
                         <pass>Pass</pass>
                     </OptionValues>
                 </action>
+                <bypass type="BooleanField">
+                        <default>0</default>
+                        <Required>N</Required>
```
AdSchellevis, line 118: better to make this required and upgrade the model version, migration will make sure it's set to 0 after upgrade.

AdSchellevis (changes requested):
one small remark, almost ready for merge. Nice work!

src/opnsense/mvc/app/models/OPNsense/IDS/IDS.xml
Monviech: I have changed the version and the requirement. Do I have to change something in a migration script, too? I'm unsure about it.

AdSchellevis: nope, it's as simple as this. defaults are part of the standard migration, so adding a new field with a default will enforce an automatic migration. Going to merge this, thanks for the effort!

Monviech: Thank you too for being so supportive.

PR #6756, opened 2023-08-18
Unbound: fixed configuration when custom blocks are used

Karlson2k:
When any custom block is used, it may change the scope from `server:` which is expected by some configuration files. 
Added explicit `server:` scope to such file for to avoid ambiguity,

Found this because custom block for forwarding to Dnscrypt-proxy changes the scope to `forward-zone:` and this produces error in `access_lists.conf` parsing.


fichtner (approved):
makes sense, but handing over to @swhite2

swhite2:
What about the `private_domains.conf` file? I see some clauses in there that are also normally grouped in `server:`. You could perhaps also test with `include-toplevel:` instead of `include` in `unbound.inc`. With a bit of luck the `server` clause is the default one anyway, which would prevent this issue for new configuration files.

Karlson2k:
@swhite2 Correct, missed this one.

I've updated the PR.

Karlson2k:
I'd stay with simple bug-fix in this PR.
`include-toplevel:` is a good idea, but it is more like improvement and should go in another PR.

swhite2:
Looks good, thanks!

PR #6765, opened 2023-08-21
Add an API endpoint to configure interfaces

levkohimins:
1. Added the ability to explicitly specify interface names for Vlan's, by default is assigned automatically, as it was.
2. Added an API endpoint for configuring interfaces. It is only possible to configure a static IP or DHCP connection, other types are not available via API. I decided not to update the xml schema for interfaces in the configuration file in order to avoid issues, and make it fully backwards compatible without changing the current logic. This is most likely a temporary solution. Even if the decision is not made to merge these change to the master, perhaps this pull request will be some intermediate solutions as it is for me, where such functionality is critically needed.

#### Resources (SettingsController.php)
Method | Module | Controller | Command | Parameters
-- | -- | -- | -- | --
POST | interfaces | settings | assignItem
POST | interfaces | settings | setItem | $uuid
POST | interfaces | settings | delItem | $uuid
GET | interfaces | settings | get |  
GET | interfaces | settings | getItem | $uuid=null
\* | interfaces | settings | searchItem |
POST | interfaces | settings | reconfigure |


AdSchellevis:
Unfortunately we can not merge this as it will require an enormous effort from the core team to complete it and change the gui accordingly, otherwise we end up with two parts of complicated code to maintain.  

Smaller steps might be possible (adding a mac validator to core, discuss change to vlan's that might be practical, etc).

levkohimins:
@AdSchellevis, I understand you. Unfortunately, I also do not have enough time to fully implement interfaces with GUI to MVC. Therefore I am closing this PR. If you would like, I can create a new PR with just the VLAN changes.

xilmen:
Hi, has the new PR been made ? Do we have another way to configure the intreface (VLAN and PHY) with api?

levkohimins:
Hi @xilmen, there are forked repos https://github.com/biptec/opnsense-* with this feature, but you need to compile an opnsense image on your own. 
You might also be interested in the ansible library, which supports interface configuration (VLAN and PHY) using the API https://github.com/biptec/ansible-opnsense.

xilmen:
Thanks. Atm, i check vyos, I feel that api / ansible module implementation is complete.

PR #6785, opened 2023-08-27
Configure SSH Host Certificates

### LINKED ISSUE #6007: Add support for SSH Certificates
**Important notices**

As far as I can tell, it doesn't appear that opnsense provides the ability to use custom SSH keys that are signed with SSH certificates. (see [here](https://goteleport.com/blog/how-to-ssh-properly/) on SSH certificates)

Specifically, I would like to be able to add something like this to the sshd_config (and also remove the existing `HostKey` entries):

```
HostKey /custom/dir/ssh_host_ed25519_key
HostKey /custom/dir/ssh_host_rsa_key
HostCertificate /custom/dir/ssh_host_ed25519_key-cert.pub
HostCertificate /custom/dir/ssh_host_rsa_key-cert.pub
TrustedUserCAKeys /custom/dir/trusted-user-ca-keys.pem
```

What would be really nice is some way to change or hook into the sshd_config before it is written. I'm not familiar enough with opnsense to know if/how this problem is solved with other configuration files, but, ideally, it would be consistent with that method.

Another possibility would be to put the keys and certificates in `/conf/ssh/` and assume a specific format (i.e. ``*-cert.pub` for certificates) and automatically add the `HostCertificate` line if that file exists in `/conf/ssh/`. This doesn't solve the problem of removing unwanted `HostKey` entries (as removed keys appear to be automatically regenerated) or adding `TrustedUserCAKeys`.

**Additional context**

I'm happy to contribute, but I would like input on a good solution before doing anything.

Thanks!

frenchie:
Allows HostCertificates to be defined via the web interface (Settings > System > Administration > Show cryptographic overrides) which are then dynamically deployed at runtime.

Note minimal checking is required on these entries as OpenSSH fails safe in all of the likely failure modes

- If the certificates do not match the key they are ignored with a warning in the log
- Any HostKeys without a matching certificate function as per normal
- Any non-key data present in these files is simply ignored by the daemon.
- SSH matches certificates to keys by fingerprint, the menu names are more to give sane filenames on disk

Does not currently set up TrustedUserCAKeys as this can already be achieved simply by adding the CA certificate to the authorized_keys file with the option `cert-authority`.

fixes #6007


AdSchellevis:
We don't plan to add more features to the legacy page, but have implemented and documented options to have local overwrites (https://docs.opnsense.org/manual/settingsmenu.html#secure-shell-advanced-settings). closing this now

Issue #6775, opened 2023-08-24
CARP: Can not use the same link-local address for multiple CARP VIPs


bimbar:
**Important notices**

Before you add a new report, we ask you kindly to acknowledge the following:

- [ x ] I have read the contributing guide lines at https://github.com/opnsense/core/blob/master/CONTRIBUTING.md
- [ x ] I am convinced that my issue is new after having checked both open and closed issues at https://github.com/opnsense/core/issues?q=is%3Aissue

**Describe the bug**

I'd like to use fe80::1/64 as CARP address on all of my networks, but there is a unique constraint on the field, so it fails to validate.

**To Reproduce**

Configure 2 CARP VIPs with fe80::1 as address (on different interfaces, of course).

**Expected behavior**

As a rule, a unique constraint is a good idea here, but it should exclude link-local addresses on different interfaces.

**Describe alternatives you considered**

Just omit the check, as it was in earlier versions.

**Environment**

Software version used and hardware type if relevant, e.g.:

OPNsense 23.4.1-amd64
FreeBSD 13.1-RELEASE-p7
OpenSSL 1.1.1u 30 May 2023

fichtner:
It's not that easy. The interface "@intX" is usually added for that reason to make then unique, but the model system is a little rigid here. But it's true that this worked before.

bimbar:
That was my workaround in the meantime, I removed the unique constraint from Vip.xml .

fichtner:
Which is fine as long as you don't add the same global address to multiple interfaces. :)

bimbar:
Sure, but maybe it is better to let the user make this mistake in this case.

fichtner:
It's never a good idea to let users run into these issues knowingly. It also prolongs support session time and effort spent on both ends.

bimbar:
So, what's the way forward here?

fichtner:
Patience. At the moment there are some other things that have priority.

bimbar:
Can we relabel this as a bug?

fichtner:
Not in my view.

bimbar:
So why is being unable to enter a valid IP address not a bug?

AdSchellevis:
it's an unsupported feature, you want to make it supported..... doesn't sound too complicated.

bimbar:
That seems like a perfectly serviceable fix.

fichtner:
No, it’s more complicated than discussed before. Doing an exception would allow to add the same link-local IP to the same interface multiple times.

[...]

PR #6808, opened 2023-08-31
Gateways: Single: Migrate to MVC

swhite2:
- `setaddr.php` and `system.xml` still needs work. This will likely require some CRUD actions on the model itself, but it is unclear whether a different strategy is more feasible.
- migrates `gateways->gateway_item` to `OPNsense->Gateways->gateway_item`, inline migration not possible due to the `gateway_groups`.

One validation has been dropped: whether a gateway IP falls within the subnet of the configured static IP address. Due to the complexity of the legacy configuration which may or may not contain outdated settings, migrating this would've been too risky. Instead, if a non-far gateway falls outside the configured subnet, the front-end will mark this one as `Misconfigured`.

The inline gateway creation option in the interfaces configure page allowed for non-unique names to be passed to the configuration. This means we have to assume that duplicate gateways can exist, which needs handling in the migration.


src/etc/inc/plugins.inc.d/dpinger.inc
```diff
@@ -428,7 +430,6 @@ function dpinger_status()
             $status[$gwname]['status'] = $r['status'];
         }
     }
-
```
fichtner, line 431: I enjoy the solitude of return statements ;)

src/opnsense/mvc/app/controllers/OPNsense/Routes/forms/dialogEditRoute.xml
```diff
@@ -15,7 +15,7 @@
     <id>route.gateway</id>
     <label>Gateway</label>
     <type>dropdown</type>
-    <help><![CDATA[Choose which gateway this route applies to or <a href="/system_gateways_edit.php">add a new one</a>.]]></help>
+    <help><![CDATA[Choose which gateway this route applies to or <a href="/ui/gateways">add a new one</a>.]]></help>
```
fichtner, line 18: strings like this are never translated due to embedded html. better to ditch the link for long-term prosperity

src/opnsense/mvc/app/models/OPNsense/Core/Menu/Menu.xml
```diff
@@ -49,10 +49,7 @@
             <LogFile order="80" VisibleName="Log File" url="/ui/diagnostics/log/core/pkg"/>
         </Firmware>
         <Gateways cssClass="fa fa-location-arrow fa-fw">
-            <Single order="10" url="/system_gateways.php">
-                <List url="/system_gateways.php*" visibility="hidden"/>
-                <Edit url="/system_gateways_edit.php*" visibility="hidden"/>
-            </Single>
+            <Single order="10" url="/ui/gateways"></Single>
```
fichtner, line 52: the naming on this sucks (mea culpa)... any other way to name this? this is our chance

src/opnsense/mvc/app/models/OPNsense/Base/Constraints/UniqueConstraint.php
```diff
@@ -56,7 +56,6 @@ public function validate($validator, $attribute): bool
             if ($allowEmpty && empty((string)$node)) {
                 return true;
             }
-
```
fichtner, line 59: spurious change

src/opnsense/service/conf/actions.d/actions_gateways.conf
```diff
@@ -0,0 +1,26 @@
+[start]
```
fichtner, line 1: is all this boilerplate really necessary? you also can't start a specific instance here (parameters:%s missing)

src/wizard/system.xml
```diff
@@ -420,11 +420,9 @@ if (!empty($input_errors)) {
             </field>
         </fields>
         <stepbeforeformdisplay><![CDATA[
-if (isset($config['gateways']['gateway_item'])) {
-    foreach ($config['gateways']['gateway_item'] as $gw) {
-        if (empty($config['wizardtemp']['wangateway']) && $gw['name'] == 'WAN_GW') {
-            $config['wizardtemp']['wangateway'] = $gw['gateway'];
-        }
+foreach ((new Gateway())->gatewayIterator() as $gw) {
```
fichtner, line 423: this might only work by chance... full path would be better.

src/etc/inc/plugins.inc.d/dpinger.inc
swhite2: woops :)

src/opnsense/service/conf/actions.d/actions_gateways.conf
swhite2: It's not pretty, i agree.. But for us to hook into `updateServiceControlUI();`, the assumption of `<service> <start/stop/restart/status>` is necessary. The generic `actions_service.conf` assumes the reverse: `<start/stop/restart/status> <service>`.

This would also make sense from the perspective of modularization, though if we can make this more generic in general we also need to make sure other modules follow the same pattern.

fichtner: ok let's try to address this at a different time

src/etc/inc/plugins.inc.d/dhcpd.inc
```diff
@@ -1826,9 +1827,10 @@ function dhcpd_dhcrelay6_configure($verbose = false)
         }
 
         if (!isset($destif)) {
-            if (is_array($config['gateways']['gateway_item'])) {
-                foreach ($config['gateways']['gateway_item'] as $gateway) {
-                    if (isset($gateway['defaultgw'])) {
+            $gateways = iterator_to_array((new \OPNsense\Gateways\Gateway)->gatewayIterator());
+            if (is_array($gateways)) {
```
fichtner, line 1831: I'm pretty sure this is always an array now and safe to feed to foreach directly

src/etc/inc/plugins.inc.d/dhcpd.inc
```diff
@@ -1718,9 +1718,10 @@ function dhcpd_dhcrelay4_configure($verbose = false)
         }
 
         if (!isset($destif)) {
-            if (is_array($config['gateways']['gateway_item'])) {
-                foreach ($config['gateways']['gateway_item'] as $gateway) {
-                    if (isset($gateway['defaultgw'])) {
+            $gateways = iterator_to_array((new Gateway)->gatewayIterator());
```
fichtner, line 1721: that one is still missing the path btw

swhite2: woops, seems i missed the last grep entry

src/etc/inc/plugins.inc.d/dhcpd.inc
```diff
@@ -1718,9 +1718,10 @@ function dhcpd_dhcrelay4_configure($verbose = false)
         }
 
         if (!isset($destif)) {
-            if (is_array($config['gateways']['gateway_item'])) {
-                foreach ($config['gateways']['gateway_item'] as $gateway) {
-                    if (isset($gateway['defaultgw'])) {
+            $gateways = iterator_to_array((new \OPNsense\Gateways\Gateway())->gatewayIterator());
+            if (is_array($gateways)) {
```
fichtner, line 1722:
```suggestion
            if (!empty($gateways)) {
```

src/opnsense/mvc/app/models/OPNsense/Core/Menu/Menu.xml
swhite2: what about "primary"?

fichtner: "what's your primary gateways?" is probably confusing for different reasons... more like "individual" but also strange word for a single "thing" :/

swhite2: well, the alternative are the groups - which are not functional until explicitly configured for PBR, in contrast to the "primary" gateways.

Other than that, we could also take inspiration from something like the routes and name it "Configuration" - it suggests being the "main" window as well as the benefit of consistency with other submenus.

src/opnsense/mvc/app/library/OPNsense/Routing/Gateways.php
```diff
@@ -30,6 +30,7 @@
 
 use OPNsense\Core\Config;
 use OPNsense\Firewall\Util;
+use OPNsense\Gateways\Gateway;
```
AdSchellevis, line 33: See note in model class

src/opnsense/mvc/app/library/OPNsense/Routing/Gateways.php
```diff
@@ -234,41 +235,38 @@ public function getGateways()
...
-                        }
-                        if (empty($gw_arr['priority'])) {
-                            // default priority
-                            $gw_arr['priority'] = 255;
-                        }
-                        if (empty($gw_arr['ipprotocol'])) {
-                            // default address family
-                            $gw_arr['ipprotocol'] = 'inet';
+            foreach ((new Gateway())->gatewayIterator() as $gw_arr) {
+                if (!empty($gw_arr)) {
```
AdSchellevis, line 239: Is it possible for `gatewayIterator()` to return empty arrays ? That looks a bit problematic

src/opnsense/mvc/app/models/OPNsense/Core/ACL/ACL.xml
```diff
@@ -632,30 +632,32 @@
             <pattern>api/diagnostics/log/core/pkg/*</pattern>
         </patterns>
     </page-system-firmware-manualupdate>
+    <page-system-gateways>
```
AdSchellevis, line 635: up for discussion how useful it is to add two acl's. I think we dropped one in the other refactors.

src/opnsense/mvc/app/models/OPNsense/Gateways/Gateway.php
```diff
@@ -0,0 +1,224 @@
...
+        return $messages;
+    }
+
+    private function validateDpingerSettings($tag, $parent, $messages, $ref)
+    {
+        switch ($tag) {
+            case 'latencylow':
+            case 'latencyhigh':
+                if ((string)$parent->latencylow > (string)$parent->latencyhigh) {
+                    $messages->appendMessage(new Message("The high latency threshold needs to be higher than the low latency threshold.", $ref.".".$tag));
```
AdSchellevis, line 100: Line wraps and missing `gettext()` (same on the other Messages)

src/opnsense/service/conf/actions.d/actions_gateways.conf
AdSchellevis: Can probably go, when removing the service controller and opt for an "apply" action only

src/www/interfaces.php
```diff
@@ -851,14 +851,10 @@ interface_sync_wireless_clones($a_interfaces[$if], false);
...
-                foreach($config['gateways']['gateway_item'] as $gateway) {
-                    if (in_array($pconfig['gateway'], $gateway)) {
-                        $match = true;
-                    }
-                }
-                foreach($config['gateways']['gateway_item'] as $gateway) {
-                    if (in_array($pconfig['gatewayv6'], $gateway)) {
+            $gateways = iterator_to_array((new \OPNsense\Gateways\Gateway())->gatewayIterator());
+            if (!empty($gateways)) {
+                foreach ($gateways as $gateway) {
```
AdSchellevis, line 856: `foreach ((new \OPNsense\Gateways\Gateway())->gatewayIterator() as $gateway)`

src/www/interfaces.php
```diff
@@ -2156,8 +2064,9 @@ function toggle_wirelesscfg() {
                             <select name="gateway" class="selectpicker" data-style="btn-default" data-size="10" id="gateway">
                               <option value="none"><?= gettext('Auto-detect') ?></option>
 <?php
-                              if (!empty($config['gateways']['gateway_item'])):
-                                foreach ($config['gateways']['gateway_item'] as $gateway):
+                              $gateways = iterator_to_array((new \OPNsense\Gateways\Gateway())->gatewayIterator());
+                              if (!empty($gateways)):
```
AdSchellevis, line 2068: same as above

src/www/interfaces.php
```diff
@@ -2717,8 +2581,9 @@ function toggle_wirelesscfg() {
                             <select name="gatewayv6" class="selectpicker" data-size="10" data-style="btn-default" id="gatewayv6">
                               <option value="none"><?= gettext('Auto-detect') ?></option>
 <?php
-                              if (!empty($config['gateways']['gateway_item'])):
-                                foreach ($config['gateways']['gateway_item'] as $gateway):
+                              $gateways = iterator_to_array((new \OPNsense\Gateways\Gateway())->gatewayIterator());
```
AdSchellevis, line 2584: same as above

AdSchellevis (changes requested):

src/opnsense/mvc/app/models/OPNsense/Base/Constraints/UniqueConstraint.php
```diff
@@ -70,9 +70,13 @@ public function validate($validator, $attribute): bool
                 // collect (additional) key fields
                 $keyFields = array($nodeName);
                 $keyFields = array_unique(array_merge($keyFields, $this->getOptionValueList('addFields')));
+                $exclusions = array_unique($this->getOptionValueList('exclusions'));
```
AdSchellevis, line 73: @swhite2 It might be an option to change this to a regex, I think that solves this one https://github.com/opnsense/core/issues/6775 assigned to @fichtner too

fichtner: This is a sort of "allowEmpty" but for a different value.... I'm dealing with a small uphill battle with an exception mask for the nodeValue... we can disable the unqiue constraint, but that doesn't make sense... we need to add more fields (in my case the interface field) if we encounter a special value... otherwise I'm able to create the same link-local address twice to the same interface. I'll finish this on Monday.

AdSchellevis: When combinations of fields should exclude a specific pattern, it might be better to fix the validation in the model itself in stead of using the constraint (not generic enough). In which case my comment might not be relevant for this field type. (I think you can add fields, but combining with a regen has a high chance of getting obscure in that case)

fichtner: Yeah, I'll think a bit more about it. The constraint would make more sense as it uses less code in the long run, but if the options are unreadable that doesn't make it any better.

src/etc/inc/plugins.inc.d/dhcpd.inc
```diff
@@ -1718,9 +1718,10 @@ function dhcpd_dhcrelay4_configure($verbose = false)
         }
 
         if (!isset($destif)) {
-            if (is_array($config['gateways']['gateway_item'])) {
-                foreach ($config['gateways']['gateway_item'] as $gateway) {
-                    if (isset($gateway['defaultgw'])) {
+            $gateways = iterator_to_array((new \OPNsense\Gateways\Gateway())->gatewayIterator());
+            if (!empty($gateways)) {
+                foreach ($gateways as $gateway) {
```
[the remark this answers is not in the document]
swhite2: Applied the suggest change, but accounted for https://github.com/opnsense/core/commit/5e9f114ea0975bc5df1f98cc340407f010bff27a

src/etc/inc/plugins.inc.d/dhcpd.inc
```diff
@@ -1822,8 +1822,8 @@ function dhcpd_dhcrelay6_configure($verbose = false)
         }
 
         if (empty($destif)) {
-            foreach ($a_gateways as $gateway) {
-                if (isset($gateway['defaultgw']) && $gateway['ipprotocol'] == 'inet6') {
+            foreach ((new \OPNsense\Gateways\Gateway())->gatewayIterator() as $gateway) {
```
fichtner, line 1825: a_gateways is already using `gatewaysIndexedByName()`. Not sure if a new class and different approach makes sense here

src/etc/inc/plugins.inc.d/dhcpd.inc
```diff
@@ -1715,8 +1715,8 @@ function dhcpd_dhcrelay4_configure($verbose = false)
         }
 
         if (empty($destif)) {
-            foreach ($a_gateways as $gateway) {
-                if (isset($gateway['defaultgw']) && $gateway['ipprotocol'] == 'inet') {
+            foreach ((new \OPNsense\Gateways\Gateway())->gatewayIterator() as $gateway) {
```
fichtner, line 1718: Same here

src/etc/inc/plugins.inc.d/dhcpd.inc
fichtner: FWIW, this will also iterate over disabled gateways, which the former code did not (I think that should be the way).

src/opnsense/mvc/app/models/OPNsense/Gateways/Gateway.php
```diff
@@ -0,0 +1,224 @@
...
+use Phalcon\Messages\Message;
+use OPNsense\Core\Config;
+
+class Gateway extends BaseModel
+{
+
+    public static function getDpingerDefaults()
+    {
+        return [
+            'alert_interval' => 1,
```
fichtner, line 43: alert_interval was removed

src/opnsense/mvc/app/models/OPNsense/Gateways/Gateway.php
```diff
@@ -0,0 +1,224 @@
...
+            'data_length' => 0,
+            'interval' => 1,
+            'latencyhigh' => 500,
+            'latencylow' => 200,
+            'loss_interval' => 2,
```
fichtner, line 48: loss_interval changed to 4 (4 * interval internall).

Rendering it to model has the disadvantage of it not being computed by dpinger anymore.

src/opnsense/mvc/app/models/OPNsense/Gateways/Gateway.php
```diff
@@ -0,0 +1,224 @@
+<?php
```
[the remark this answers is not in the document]
swhite2: @AdSchellevis Do we want to move the controller and view part to the `Routing` namespace as well?

src/opnsense/mvc/app/controllers/OPNsense/Gateways/forms/dialogEditGateway.xml
```diff
@@ -0,0 +1,150 @@
...
+        <advanced>true</advanced>
+        <type>text</type>
+        <help>The time period over which results are averaged. Default is 60.</help>
+    </field>
+    <field>
+        <id>gateway_item.loss_interval</id>
+        <label>Loss Interval</label>
+        <advanced>true</advanced>
+        <type>text</type>
+        <help>Time interval before packets are treated as lost. Default is 2.</help>
```
fichtner, line 141:
```suggestion
        <help>Time interval before packets are treated as lost. Default is 4 (four times the probe interval).</help>
```

src/opnsense/mvc/app/controllers/OPNsense/Gateways/forms/dialogEditGateway.xml
```diff
@@ -0,0 +1,150 @@
...
+        <type>text</type>
+        <help>How often that an ICMP probe will be sent in seconds. Default is 1.</help>
+    </field>
+    <field>
+        <id>gateway_item.alert_interval</id>
+        <label>Alert Interval</label>
+        <advanced>true</advanced>
+        <type>text</type>
+        <help>Time interval between alerts. Default is 1.</help>
+    </field>
```
fichtner, line 128:
```suggestion
```

src/opnsense/mvc/app/models/OPNsense/Routing/Gateways.php
```diff
@@ -47,9 +45,200 @@ class Gateways
...
     public function __construct(array $unused = [])
     {
+        parent::__construct();
         $this->configHandle = Config::getInstance()->object();
     }
 
+    public static function getDpingerDefaults()
+    {
+        return [
+            'alert_interval' => 1,
```
fichtner, line 55:
```suggestion
```

src/opnsense/mvc/app/models/OPNsense/Routing/Gateways.php
```diff
@@ -47,9 +45,200 @@ class Gateways
...
+            'data_length' => 0,
+            'interval' => 1,
+            'latencyhigh' => 500,
+            'latencylow' => 200,
+            'loss_interval' => 2,
```
fichtner, line 60:
```suggestion
            'loss_interval' => 4,
```

src/opnsense/mvc/app/models/OPNsense/Gateways/Gateway.php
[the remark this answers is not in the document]
AdSchellevis: @swhite2 for consistency that's likely a good idea

src/opnsense/mvc/app/models/OPNsense/Routing/Gateways.xml
```diff
@@ -0,0 +1,160 @@
...
+                </Constraints>
+            </losshigh>
+            <interval type="IntegerField">
+                <Required>N</Required>
+                <MinimumValue>1</MinimumValue>
+            </interval>
+            <alert_interval type="IntegerField">
+                <Required>N</Required>
+                <MinimumValue>1</MinimumValue>
+            </alert_interval>
```
fichtner, line 145:
```suggestion
```

src/opnsense/mvc/app/models/OPNsense/Routing/Gateways.php
```diff
@@ -28,14 +28,12 @@
 
 namespace OPNsense\Routing;
 
-use OPNsense\Core\Config;
+use OPNsense\Base\BaseModel;
 use OPNsense\Firewall\Util;
+use Phalcon\Messages\Message;
+use OPNsense\Core\Config;
```
fichtner, line 34: would be nicer to sort these alphabetically

src/opnsense/mvc/app/controllers/OPNsense/Gateways/Api/SettingsController.php
```diff
@@ -0,0 +1,259 @@
...
+                }
+            }
+        }
+
+        return $this->searchRecordsetBase($gateways);
+    }
+
+    public function getGatewayAction($uuid = null)
+    {
+        if ($uuid != null) {
```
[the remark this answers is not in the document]
swhite2: @AdSchellevis Added https://github.com/opnsense/core/pull/6808/commits/ee5758d2691e4c30831a6a4171ed26ba6e74ef2c#diff-4db3c29903c593a27a8c7583cc8791a5e16ae565e126586afb271a4dcf569a28R81 to facilitate this. We could move this up to `ApiControllerBase`, though it is unlikely the `ServiceControllerBase` will have any use for this.

src/opnsense/mvc/app/controllers/OPNsense/Gateways/Api/SettingsController.php
```diff
@@ -0,0 +1,259 @@
...
+            $gateways = array_values($this->gw->gatewaysIndexedByName(true, false, true));
```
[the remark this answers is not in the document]
swhite2: Opted to not bail here, as the view relies on getAction to provide an empty node in cases of new additions.

src/opnsense/mvc/app/library/OPNsense/Routing/Gateways.php
swhite2: No, not possible :)

src/opnsense/mvc/app/controllers/OPNsense/Gateways/Api/SettingsController.php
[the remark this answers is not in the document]
AdSchellevis: @swhite2 `ApiMutableModelControllerBase` looks like the right place for this refactor indeed.

src/etc/inc/interfaces.inc
```diff
@@ -1057,15 +1057,13 @@ default:
...
-    if (isset($config['gateways']['gateway_item'])) {
-        foreach ($config['gateways']['gateway_item'] as $gateway) {
-            if ($interface == $gateway['interface'] && isset($gateway['defaultgw'])) {
-                $setdefaultgw = true;
-                break;
-            } elseif (isset($gateway['defaultgw']) && !empty($gateway['interface'])) {
-                $founddefaultgw = true;
-                break;
-            }
+    foreach ((new \OPNsense\Gateways\Gateway())->gatewayIterator() as $gateway) {
```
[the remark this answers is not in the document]
swhite2: https://github.com/opnsense/core/commit/7a2f3f62d91fea0a2216c3d4d0d9b95ea26e2f7f

src/opnsense/mvc/app/models/OPNsense/Routing/Migrations/M1_0_0.php
```diff
@@ -0,0 +1,88 @@
...
+                }
+
+                foreach ($gateway as $key => $value) {
+                    if ($key === 'gateway') {
+                        // change all occurences of "dynamic" to empty string
+                        $node->gateway = str_replace('dynamic', '', (string)$value);
+                        continue;
+                    }
+
+                    if ($key === 'alert_interval') {
```
fichtner, line 67: can't we match this against model nodes?

swhite2: how about https://github.com/opnsense/core/pull/6808/commits/373a48c274b6a2f1ed1f3a42e8ee37ae70181e05?

fichtner (approved):
LGTM 👍

src/opnsense/mvc/app/models/OPNsense/Routing/Migrations/M1_0_0.php
```diff
@@ -0,0 +1,89 @@
...
+                        // change all occurences of "dynamic" to empty string
+                        $node->gateway = str_replace('dynamic', '', (string)$value);
+                        continue;
+                    }
+
+                    $node->$key = (string)$value;
+                }
+
+                // apply dpinger defaults if old model didn't have them set
+                foreach ((new Gateways)::getDpingerDefaults() as $key => $value) {
```
fichtner, line 78: does this have to be constructed to be used? The function is declared as static.

swhite2: You're right, oversight on my end - at some point during development it was a regular public function

swhite2:
@fichtner I cleaned up the model a bit to be in line with your other mvc changes https://github.com/opnsense/core/pull/6808/commits/6b55eb975c5a8a3e007bdc85c4cc6cae95a5b81b. Any objections to the current default/required properties in the model in general?

fichtner:
model looks good. maybe you can try "make lint-model" if you haven't already.

not sure about "monitoring was on when no node present, exception to the rule above" monitoring was specifically never enforced by default after 2016 or so. It was once on by default but almost nobody should still have this "valid" configuration (unchanged since then) and turning it off just means it's always online so risk of breakage is low.

swhite2:
monitoring is still off by default. the legacy config could have either a `monitor_disable` with 0 or no node at all - in both cases this means monitoring on. the line of code in question simply applies this, otherwise it will turn monitoring off for all gateways which have it enabled explicitly.

fichtner:
>  the line of code in question simply applies this

true, ok if you want to keep it that way. just meant to say it's probably not necessary. :)

src/etc/inc/plugins.inc.d/dpinger.inc
```diff
@@ -320,7 +320,7 @@ function dpinger_status()
         $status[$gwname] = $report = [
             'status' => !empty($gwitem['force_down']) ? 'force_down' : $gwstatus,
             /* grab the runtime monitor from the instance if available */
-            'monitor' => $gwitem['monitor'] ?? '~',
+            'monitor' => !empty($gwitem['monitor']) ? $gwitem['monitor'] : '~',
```
fichtner, line 323: don't mind moving this to master so we don't have to though dpinger.inc from this PR again :)

swhite2: done :)

AdSchellevis (approved):

src/opnsense/mvc/app/models/OPNsense/Routing/Gateways.php
```diff
@@ -1,7 +1,7 @@
 <?php
 
 /*
- * Copyright (C) 2019-2023 Deciso B.V.
+ * Copyright (C) 2023 Deciso B.V.
```
fichtner, line 4:
```suggestion
 * Copyright (C) 2019-2023 Deciso B.V.
```

src/etc/inc/plugins.inc.d/core.inc
```diff
@@ -426,7 +426,7 @@ function core_xmlrpc_sync()
     $result[] = array(
         'description' => gettext('Static Routes'),
         'help' => gettext('Synchronize the Static Route configuration and Gateways to the other HA host.'),
-        'section' => 'staticroutes,gateways',
+        'section' => 'staticroutes,gateways,OPNsense.Gateways',
```
fichtner, line 429: @swhite2 @AdSchellevis this is marginally related to the PR, but I have to ask :)

Here we keep "gateways" because the code can deal with it.... but what do we actually do in the backup import now with these sections all being marked as "importable"? Because I've been thinking of the alias situation... Should we clear the sections not given in the backup? It would allow a direct old alias import since it forces migration... I suppose the same is true for "gateways" key from an older install when "OPNsense.Gateways" was already created by the default migration on a newer version..

src/opnsense/mvc/app/models/OPNsense/Core/Menu/Menu.xml
```diff
@@ -47,10 +47,7 @@
             <LogFile order="80" VisibleName="Log File" url="/ui/diagnostics/log/core/pkg"/>
         </Firmware>
         <Gateways cssClass="fa fa-location-arrow fa-fw">
-            <Single order="10" url="/system_gateways.php">
-                <List url="/system_gateways.php*" visibility="hidden"/>
-                <Edit url="/system_gateways_edit.php*" visibility="hidden"/>
-            </Single>
+            <Configuration order="10" url="/ui/routing/configuration"></Configuration>
```
fichtner, line 50:
```suggestion
            <Configuration order="10" url="/ui/routing/configuration"/>
```

src/opnsense/scripts/shell/setaddr.php
```diff
@@ -259,26 +269,26 @@ function add_gateway_to_config($interface, $gatewayip, $inet_type, $is_in_subnet
...
-    if (!$is_in_subnet) {
-        $item['fargw'] = 1;
-    }
+        'fargw' => !$is_in_subnet ? 1 : 0
+    ];
 
-    $a_gateways[] = $item;
+    $gw->createOrUpdateGateway($item, $uuid);
+    /* Serialize back to global config */
+    $config = OPNsense\Core\Config::getInstance()->toArray(listtags());
```
fichtner, line 291: ugh :)

fichtner: doesn't this attempt to do "write_config()" later on?

src/wizard/system.xml
```diff
@@ -605,33 +605,33 @@ if (!empty($input_errors)) {
...
-        }
-        if ($gw['name'] == 'WAN_GW') {
-            $found = true;
-            $gw['gateway'] = $config['wizardtemp']['wangateway'];
-            $config['interfaces']['wan']['gateway'] = $gw['name'];
-        }
+$found = false;
+$defaultgw_found = false;
+$gw = new \OPNsense\Routing\Gateways();
+foreach ($gw->gatewayIterator() as $gw) {
```
fichtner, line 611: $gw->xxx as $gw is asking for trouble?

src/opnsense/scripts/shell/setaddr.php
swhite2: yes, but it isn't the same instance, result is that the latter `write_config()` will overwrite the changes made here

fichtner: I know, maybe the comment needs to be improved. it explains what but not why

swhite2: sure, and no, it isn't pretty

fichtner: to be fair the problem is real, but before write_config() would actually serialize all of the changes at once. Maybe the code doing the config sync could be moved closer together or the $config is rewritten in memory but not on the disk for the gateway flush?

src/opnsense/scripts/shell/setaddr.php
```diff
@@ -259,26 +269,28 @@ function add_gateway_to_config($interface, $gatewayip, $inet_type, $is_in_subnet
...
-    if (!$is_in_subnet) {
-        $item['fargw'] = 1;
-    }
-
-    $a_gateways[] = $item;
+        'fargw' => !$is_in_subnet ? 1 : 0
+    ];
+
+    $gw->createOrUpdateGateway($item, $uuid);
+    /* Serialize back to global config, capturing the changes made here.
```
fichtner, line 290:
```suggestion
    /* 
     * Serialize back to global config, capturing the changes made here.
```

multiline comments be like this :D

src/opnsense/scripts/shell/setaddr.php
fichtner: sorry maybe it's doing that. still on my first coffee

src/etc/inc/plugins.inc.d/core.inc
fichtner: pew pew? https://github.com/opnsense/core/commit/b8bbb00da7

works for me on legacy "aliases" backup ;)

PR #6815, opened 2023-09-01
Fix misleading help text

8191:
Confusing punctation would imply that the field is only relevant for certificate authentication.


AdSchellevis:
@8191 thanks!

PR #6817, opened 2023-09-02
Fix IPsec key pair generator for 256 EC and add properer naming…

8191:
Key generation for the "256" EC does not work as the curve used in the generator code "secp256r1", does not exist in OpenSSL; it's called prime256v1 there.
Source: https://neuromancer.sk/std/nist/P-256

Also the names shown in the UI are ambiguous as there are several curves with 256, 384 or 521 bits. Renaming the drop-down list to "NIST P-256", "NIST P-384" and "NIST P-521" makes them uniquely identifiable.


AdSchellevis:
@8191 looks good, thanks!

PR #6822, opened 2023-09-05
OpenVPN Client Export: fix certificate list when CA specified

8191:
When a different CA is specified for client certificate validation than the server certificate's CA, a wrong list of certificates is shown in the Client Export dialog.


AdSchellevis:
@8191 thanks!

Issue #6818, opened 2023-09-04
Add local_port and remote_port to VPN: IPsec: Connections [new]


Monviech:
**Important notices**

Before you add a new report, we ask you kindly to acknowledge the following:

- [x] I have read the contributing guide lines at https://github.com/opnsense/core/blob/master/CONTRIBUTING.md
- [x] I am convinced that my issue is new after having checked both open and closed issues at https://github.com/opnsense/core/issues?q=is%3Aissue

**Is your feature request related to a problem? Please describe.**

It's for specific situations where UDP 500 for NAT detection in IKEv2 fails.
Example: OPNsense is behind a router which NATs. This other router also has an IPSec daemon listening on port UDP 500. The IKE response will always be caught by the router in front of the OPNsense. So Phase 1 will fail because the IKE response packet won't be received by the OPNsense.

**Describe the solution you like**

I would like local_port and remote_port added to "VPN: IPsec: Connections [new]" in "General settings" of a "Connection" when "advanced mode" is enabled.

If both are set to 4500, the IKE_SA_INIT request will be sent from UDP 4500 to UDP 4500 right away.

https://docs.strongswan.org/docs/5.9/swanctl/swanctlConf.html#_connections
<conn>.local_port
<conn>.remote_port

https://docs.strongswan.org/docs/5.9/features/natTraversal.html#_custom_server_ports

Thank you very much for reading and considering this feature request.

**Describe alternatives you considered**

Using IPv6. Sadly it's not always a choice.

**Additional context**

Here's swanctl connections with Default Port and local/remote port set to 4500

**Default Port (500)**

```
[IKE] initiating IKE_SA site5-to-opn01[135] to 91.XXX.XXX.XXX
[ENC] generating IKE_SA_INIT request 0 [ SA KE No N(NATD_S_IP) N(NATD_D_IP) N(FRAG_SUP) N(HASH_ALG) N(REDIR_SUP) ]
[NET] sending packet: from 10.169.172.207[500] to 91.XXX.XXX.XXX[500] (464 bytes)
[NET] received packet: from 91.XXX.XXX.XXX[500] to 10.169.172.207[500] (472 bytes)
[ENC] parsed IKE_SA_INIT response 0 [ SA KE No N(NATD_S_IP) N(NATD_D_IP) N(FRAG_SUP) N(HASH_ALG) N(CHDLESS_SUP) N(MULT_AUTH) ]
[...]

PR #6835, opened 2023-09-08
Add local_port and remote_port to VPN: IPsec: Connections [new]

Monviech:
[#6818](https://github.com/opnsense/core/issues/6818)

I created new dropdown menus in the IPsec Connections Form. 
Users can select between port 500 (default) and port 4500 (NAT-T). I didn't create a free input field since other values have to be actually supported by the backend, and I didn't see the usecase for free choice.

I didn't upgrade the model version since I used the `elseif ((string)$attr == '')` in the Swanctl.php in order to skip the default value. I chose the default value as empty, because I didn't want to add local_port = 500 and remote_port = 500 explicitly to the configuration file, it's not needed since its the default even if its not specified. Thus I could maintain a non required field that won't need a migration since no config changes will be made when its implemented.

When a user enables advanced mode and chooses "4500 (NAT-T), the port gets added in the swanctl.conf.

### DIFF
--- a/src/opnsense/mvc/app/models/OPNsense/IPsec/Swanctl.xml
+++ b/src/opnsense/mvc/app/models/OPNsense/IPsec/Swanctl.xml
@@ -46,7 +46,6 @@
                 </local_addrs>
                 <local_port type="OptionField">
                     <Required>N</Required>
-                    <Default></Default>
                     <OptionValues>
                         <port500 value="">500 (default)</port500>
                         <port4500 value="4500">4500 (NAT-T)</port4500>
@@ -57,7 +56,6 @@
                 </remote_addrs>
                 <remote_port type="OptionField">
                     <Required>N</Required>
-                    <Default></Default>
                     <OptionValues>
                         <port500 value="">500 (default)</port500>
                         <port4500 value="4500">4500 (NAT-T)</port4500>
--- a/src/opnsense/mvc/app/models/OPNsense/IPsec/Swanctl.xml
+++ b/src/opnsense/mvc/app/models/OPNsense/IPsec/Swanctl.xml
@@ -44,9 +44,25 @@
                 <local_addrs type=".\IKEAdressField">
                     <Required>N</Required>
                 </local_addrs>
+                <local_port type="OptionField">
+                    <Required>N</Required>
+                    <Default></Default>
+                    <OptionValues>
+                        <port500 value="">500 (default)</port500>
+                        <port4500 value="4500">4500 (NAT-T)</port4500>
+                    </OptionValues>
+                </local_port>
                 <remote_addrs type=".\IKEAdressField">
                     <Required>N</Required>
                 </remote_addrs>
+                <remote_port type="OptionField">
+                    <Required>N</Required>
+                    <Default></Default>
+                    <OptionValues>
+                        <port500 value="">500 (default)</port500>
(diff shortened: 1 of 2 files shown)


AdSchellevis:
@Monviech the `Swanctl.php` file seems to be missing in your PR

Monviech:
@AdSchellevis 
I didnt need to change it. The php file already polled for local and skipped empty strings so it all worked without a change to the code.

AdSchellevis:
@Monviech my mistake :) needed coffee probably ;)

src/opnsense/mvc/app/models/OPNsense/IPsec/Swanctl.xml
```diff
@@ -44,9 +44,25 @@
                 <local_addrs type=".\IKEAdressField">
                     <Required>N</Required>
                 </local_addrs>
+                <local_port type="OptionField">
+                    <Required>N</Required>
+                    <Default></Default>
```
AdSchellevis, line 49: you can drop the default clause here

src/opnsense/mvc/app/models/OPNsense/IPsec/Swanctl.xml
```diff
@@ -44,9 +44,25 @@
...
+                        <port500 value="">500 (default)</port500>
+                        <port4500 value="4500">4500 (NAT-T)</port4500>
+                    </OptionValues>
+                </local_port>
                 <remote_addrs type=".\IKEAdressField">
                     <Required>N</Required>
                 </remote_addrs>
+                <remote_port type="OptionField">
+                    <Required>N</Required>
+                    <Default></Default>
```
AdSchellevis, line 60: same here

AdSchellevis (changes requested):

src/opnsense/mvc/app/models/OPNsense/IPsec/Swanctl.xml
Monviech: I removed the default and tested it. Still works as expected. Thank you.

AdSchellevis:
@Monviech looks good, thanks!

gongoscho:
Hello,

please make an empty input field / possibility to change port for IPSec.

Because there are facis firewalls that block the default IPsec ports.

Custom ports are supported: https://docs.strongswan.org/docs/5.9/features/natTraversal.html#_custom_server_ports

Need a feature request for IPSec Custom Port?

grEEtZ

Monviech:
Wouldn't it make more sense to use OpenVPN on TCP 443 or something?

gongoscho:
Maybe an example: [The Android app stops receiving packets from the gateway](https://github.com/strongswan/strongswan/discussions/2234)

gongoscho:
> Wouldn't it make more sense to use OpenVPN on TCP 443 or something?

II don't think so, because nothing comes close to IPSec (strongSwan). Integrated almost everywhere in the OS, stable, very good AES performance.

End-to-end military grade encryption with e.g. Curve448 or Curve25519

Monviech:
Please create a new feature request.

I just think that changing ports here will not give the desired benefit since UDP encapsulated ESP packets can be easily detected in their pattern by a traffic matcher and blocked.

Changing ports these days is no guarantee. Thats why OpenVPN on TCP/443 with additional protocol scrambling is the best choice to get through any firewall. And if security is an issue, IPsec can be established through the OpenVPN tunnel additionally.

PR #6843, opened 2023-09-11
mvc: regex conditional for UnqiueConstraint to fix #6775

### LINKED ISSUE #6775: CARP: Can not use the same link-local address for multiple CARP VIPs

fichtner:
Solution for unique constraint extension based on `addFields` extension with conditional regex matching. If this works for @AdSchellevis I will add a test case as well.

### DIFF
--- a/src/opnsense/mvc/app/models/OPNsense/Base/Constraints/BaseConstraint.php
+++ b/src/opnsense/mvc/app/models/OPNsense/Base/Constraints/BaseConstraint.php
@@ -73,11 +73,11 @@ abstract class BaseConstraint extends BaseValidator
      */
     protected function getOptionValueList($fieldname)
     {
-        $result = array();
+        $result = [];
         $options = $this->getOption($fieldname);
         if (!empty($options)) {
-            foreach ($options as $option) {
-                $result[] = $option;
+            foreach ($options as $tag => $option) {
+                $result[$tag] = $option;
             }
         }
         return $result;
--- a/src/opnsense/mvc/app/models/OPNsense/Base/Constraints/UniqueConstraint.php
+++ b/src/opnsense/mvc/app/models/OPNsense/Base/Constraints/UniqueConstraint.php
@@ -62,8 +62,17 @@ class UniqueConstraint extends BaseConstraint
             }
             if ($containerNode != null && $level == 2) {
                 // collect (additional) key fields
-                $keyFields = array($nodeName);
+                $keyFields = [$nodeName];
                 $keyFields = array_unique(array_merge($keyFields, $this->getOptionValueList('addFields')));
+                $regexFields = $this->getOptionValueList('addFieldsRegex');
+                // remove fields if regex was specified matching on constraint anchor value
+                foreach ($regexFields as $tag => $regex) {
+                    if (!isset($keyFields[$tag])) {
+                        continue;
+                    } elseif (!preg_match($regex, (string)$parentNode->{$keyFields[0]})) {
+                        unset($keyFields[$tag]);
+                    }
+                }
                 // calculate the key for this node
                 $nodeKey = '';
                 foreach ($keyFields as $field) {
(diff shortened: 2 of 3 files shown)


fichtner:
@AdSchellevis if you say know I have time to rework :)

AdSchellevis:
@fichtner How about changing the exclude format to something like:

`<filter><subnet> /^fe[89ab][0-9a-f]:/i </subnet><interface>.*</interface></filter>`

Although not totally sure what the regex for interface should be, but it would make the filters very explicit.

fichtner:
Conceptually the match adds the field so the regex is a conditional extension for addFields. The match should either be on the anchor element (subnet value) or the full unconditional unique key of addFields -- but the latter is hard to control with complex unique keys if that ever happens. I've tried to use XML attributes but they disappear quickly in the constraints handling. Maybe that would be the most explicit solution. Rewriting the handling for addFields would also be possible to next filters inside: `field1` as general container.

AdSchellevis:
The addFields construction didn’t take this kind of filters into account. I guess the issue is that in some cases you just want to exclude a specific value from the list (Stephan’s original issue if I’m not mistaken) and sometimes the constraint has a hard relationship with the fields added. 

My thought was if we make the target explicit for the exclusion, it would be much easier to read. Though I am still struggling a bit with what this filter should offer, in theory for this example it should add a token (interface) for all subnets matching the regex to determine uniqueness. In which case it would be more logical if addField could contain a condition inside the same construct. I’m just not sure if this could fix the other issue as well.

Looking at the code, I think it’s more about how the user/developer should define this than what it’s currently doing (which is probably fine, just need to check if Stephan’s use-case fits as well)

fichtner:
Ok, Stephan’s issue was entirely different: treating a special value as the blank value (“dynamic”) in this case. We discussed it this week and the best way to avoid the problem is migrate the “dynamic” value to a blank value “” in line with the model. That makes an extension unnecessary.

What I’m trying to do is to extend the construction of the primary key based on real word condition (link-local multiple addresses allowed but need @intX appended to be unique). If it’s too special maybe better to do the model validation manually, but the unique constraint handles 80% of the code already it doesn’t make much sense not to extend it and copy+even extend it.

Yep, it’s a presentation issue first and foremost. Either we can try to allow attributes for each field or we restructure addFields to make it easier to extend:

`<addField><name>interface</name><parentMatch></parentMatch></addField>`

and support multiple addField nodes, maybe even in addition to keeping addFields. And I’m not tied to addField or name or parentMatch at all… whatever makes the most sense and is easy to use. 😊

AdSchellevis:
That sounds good indeed, stick the field to the condition, this should be relatively easy to fix and explain.

fichtner:
Hmm, I can't get getOption() to deal with "array" values in XML so multiple fields of `addField` are flattened to the last one. I'm doubting a bit if this is all worth it. The constructor comes all the way from BaseValidator.

AdSchellevis:
I'm also fine by just fixing this in the vip model, but all things considered, all of this has quite a low priority and there are not a lot of similar issues we can reuse the same construct (that I'm aware of).

fichtner:
Yeah, let's fix it in the model validation then.

PR #6844, opened 2023-09-13
unbound: fix problematic migration of interfaces

fichtner:
This pertains to migration of `active_interface` and `outgoing_interface` from legacy configuration to model. It's easy to reproduce by adding a nonexistent interface there and trying to migrate it, because...

Interfaces may not be available or disabled.  We can still give a hint with the tokenizer/type-ahead but we have to accept the old configuration or else the migration will always fail as long as bad values are stored.  CSV field will allow these arbitrary values and deletion thereof if needed even if we do not follow a strict UX pattern here.

Also remove the Loopback from the selection.  It is always added automatically to prevent the user breaking local resolving.

All ideas welcome.

### DIFF
--- a/src/opnsense/mvc/app/controllers/OPNsense/Unbound/forms/general.xml
+++ b/src/opnsense/mvc/app/controllers/OPNsense/Unbound/forms/general.xml
@@ -14,6 +14,8 @@
         <id>unbound.general.active_interface</id>
         <label>Network Interfaces</label>
         <type>select_multiple</type>
+        <style>tokenize</style>
+        <allownew>true</allownew>
         <help>
             Interface IP addresses used for responding to queries from clients.
             If an interface has both IPv4 and IPv6 IPs, both are used.
@@ -115,6 +117,8 @@
         <id>unbound.general.outgoing_interface</id>
         <label>Outgoing Network Interfaces</label>
         <type>select_multiple</type>
+        <allownew>true</allownew>
+        <style>tokenize</style>
         <advanced>true</advanced>
         <help>
             Utilize different network interfaces that Unbound will use to send queries to authoritative servers and receive their replies.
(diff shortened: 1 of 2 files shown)


AdSchellevis:
Sounds like an issue in the migration to me, when values don’t exist, the migration should omit them. Adding arbitrary values might not be the best option here.

AdSchellevis:
The other option is to allow disabled interfaces as well, both work, the latter is easier to keep consistent, the first one more strict (and maybe more accurate).

swhite2:
Should disabled interfaces be allowed, it's still possible for unbound to bind to a VIP on that interface as neither the unbound code nor `interfaces_addresses` checks for disabled interfaces, only if an ipaddr exists. This was possible in legacy scenarios so we will be changing this behavior if we discard disabled interfaces. Something to consider.

fichtner:
I don't see what you mean. interfaces_addresses() only looks for existing addresses. As long as they are applied they are used (and can be used). That's one of the reasons to allow disabled interfaces here because the code doesn't care as much as the selector does (and discards selection and hard-fails on migration).

swhite2:
that's precisely my point, an argument in favor of allowing disabled interfaces :)

AdSchellevis:
It might still need a fix in the migration as well, but allowing disabled sounds like the best option.

fichtner:
Yeah, still need to discard all options not available.

fichtner:
Ok first part is https://github.com/opnsense/core/commit/fe17f4ff48 as discussed. I'll do a migration for spurious 'lo0' setting just to be sure.

The big question: should a BaseListField be able to drop faulty values during migration via a separate function? Something like` discardInvalid()`? I reckon this will be coming up every now and then?

AdSchellevis:
It might make sense to move this to the model indeed, but for safety, it’s likely best to make sure it’s in the abstract base class as well. The implementation may then choose to implement it when needed (base list is likely a generic merge as you proposed)

fichtner:
> It might make sense to move this to the model indeed

Do you really mean "model" or should this be "field" here?

fichtner:
Ah ok now I think I understand ;)

PR #6845, opened 2023-09-13
Firewall / Aliases - Don't skip renamed aliases on cleanup

kulikov-a:
Hi!
ref. https://forum.opnsense.org/index.php?topic=35903.0
reproducible

I apologize in advance if I missed something (quite complex logic ;) ).
Cleanup worked for renamed aliases on 22.7. I think registered_aliases did  not include aliases from 'pfctl -sT' with corresponding hash-files but not presented in config

Thanks!

### DIFF
--- a/src/opnsense/scripts/filter/update_tables.py
+++ b/src/opnsense/scripts/filter/update_tables.py
@@ -125,7 +125,7 @@ if __name__ == '__main__':
 
     # cleanup removed aliases when reloading all
     if to_update is None:
-        registered_aliases = [alias.get_name() for alias in aliases if alias.is_managed()]
+        registered_aliases = [alias.get_name() for alias in aliases if (alias.is_managed() and alias.get_type() != 'external')]
         to_remove = list()
         to_remove_files = dict()
         for filename in glob.glob('/var/db/aliastables/*.txt'):


AdSchellevis:
Could be some race, but I'm quite sure the type can't be used to drop the old aliases (as dynamic ones are external).

The `is_managed()` call should check if there is still a file created by us, so it knows it was "static" at some point.

https://github.com/opnsense/core/blob/f890334f9a35082b4dc0c51e842a848817f143d7/src/opnsense/scripts/filter/lib/alias/__init__.py#L145

The only downside is, that if it is not able to remove the alias for some reason, it will not try again later. My best guess is the ruleset somehow locks the alias and the removal  doesn't happen (haven't seen this yet to be honest):

https://github.com/opnsense/core/blob/f890334f9a35082b4dc0c51e842a848817f143d7/src/opnsense/scripts/filter/update_tables.py#L143

When it did try to delete the alias, there should be a notice like "remove old alias XX" in the logs.

kulikov-a:
Hi!

> When it did try to delete the alias, there should be a notice like "remove old alias XX" in the logs.

no, this does not happen with the current script (and is starting to appear with the proposed changes)

It still seems to me that the reason is in the changed logic of the update_tables script (or rather in determining the search range for removed tables). if I'm not missing anything in the sequence: we rename the alias and hit the Apply button,  the config changes accordingly, new rules.debug generated and pf (re)loads config. Reloading a config file by itself does not remove tables from memory (only redefines matching ones and creates missing ones). So we have "OriginalAlias" in pf memory at this point. Please correct me if im wrong.
In 22.7 `registered_aliases` contained only tables with  known parser (so it is **only from the source_tree** for which there was a parser). and thus the "OriginalAlias" table did not end up in `registered_aliases` and successfully cleared at the end of the script.
Now `registered_aliases` also contains tables from the pf memory with the corresponding hash-file (because of **or** in `is_managed(self)`). So we skip the  "OriginalAlias" from the cleanup? 
or am I missing something obvious?
thanks)

AdSchellevis:
@kulikov-a can you try https://github.com/opnsense/core/commit/6393c5b5776de7dd93830c9377c50944d9dee5d2 ?

kulikov-a:
@AdSchellevis Thanks!

PR #6853, opened 2023-09-15
dhcpd6: add scope to link-local DHCPv6 static mapping when creating route for delegated prefix

maurice-w:
Allows prefix delegation to downstream routers which have an unnumbered WAN interface.
As discussed on the forum: https://forum.opnsense.org/index.php?topic=35724.msg174257#msg174257

### DIFF
--- a/src/opnsense/scripts/dhcp/prefixes.php
+++ b/src/opnsense/scripts/dhcp/prefixes.php
@@ -87,7 +87,7 @@ foreach (plugins_run('static_mapping') as $map) {
 
         if (!empty($host['ipaddrv6'])) {
             /* we want static mapping to have a higher priority */
-            $duid_arr[$host['duid']]['address'] = $host['ipaddrv6'];
+            $duid_arr[$host['duid']]['address'] = is_linklocal($host['ipaddrv6']) ? $host['ipaddrv6'] . '%' . get_real_interface($host['interface']) : $host['ipaddrv6'];
         }
     }
 }


fichtner:
There is a little error here but I'll test and fix post-merge tomorrow. Thanks!

fichtner:
@maurice-w Can you try with d86702262bf4 and 7fcbb220943 on top? looks good on my end

maurice-w:
3582242 + d867022 looks good here, too.

7fcbb22 has a side effect though when used with a track6 interface: The PD range doesn't get merged anymore. I tested this with a /62 delegated from upstream (2001:db8:5678:9abc::/62), address range ::1 - ::ff, PD range ::1 - ::3 and PD size 64.

Without 7fcbb22, this gets properly merged:
```
range6 2001:db8:5678:9abc::1 2001:db8:5678:9abc::ff;
prefix6 2001:db8:5678:9abd:: 2001:db8:5678:9abf::/64;
```

With 7fcbb22, only the address range gets merged:
```
range6 2001:db8:5678:9abc::1 2001:db8:5678:9abc::ff;
prefix6 ::1 ::3/64;
```

Interestingly, if I don't specify an address range, the prefix range does get merged:
```
prefix6 2001:db8:5678:9abd:: 2001:db8:5678:9abf::/64;
```

fichtner:
Thanks for testing. I guess we will have to provide a bit of an override in the edge case there. I’ll do that and ship in 23.7.6.

fichtner:
@maurice-w likely just 40955ddaee1960

maurice-w:
@fichtner Yes, works with 40955ddaee1960e885294d908a1098c8c1cbfaba!

fichtner:
Ok, mea culpa. Thanks a lot!

Issue #6860, opened 2023-09-18
VPN: IPsec: Connections: Pools add IP4_DNS and IP6_DNS Configuration Payloads


Monviech:
**Important notices**

Before you add a new report, we ask you kindly to acknowledge the following:

- [x] I have read the contributing guide lines at https://github.com/opnsense/core/blob/master/CONTRIBUTING.md
- [x] I am convinced that my issue is new after having checked both open and closed issues at https://github.com/opnsense/core/issues?q=is%3Aissue

**Is your feature request related to a problem? Please describe.**

With the current implementation of pools in the OPNsense GUI it's not possible to define  DNS Servers as "Configuration Payload" (rfc4306 and rfc7296 3.15).

The DNS Server Configuration Payloads are defined as:
```
                                                        
Attribute Type           Value     Multi-Valued   Length

INTERNAL_IP4_DNS         3     YES                    0 or 4 octets
INTERNAL_IP6_DNS        10    YES                    0 or 16 octets
```
https://docs.strongswan.org/docs/5.9/swanctl/swanctlConf.html#_pools
The strongswan docs defines all additional attributes that are accepted in pools. ``<name>.<attr>`` allows the use of the ``Value`` as numeric number.

The Values 3 and 10 are mapped in ``vici`` as ``dns`` and allow multiple values seperated by a comma as delimiter.

An example configuration ``swanctl.conf`` ``pools{}`` section would look like this:
```
pools {
    pool-roadwarrior-ipv4 {
        addrs = 172.16.203.0/24
        dns = 192.168.1.1,192.168.1.2
    }
    pool-roadwarrior-ipv6 {
        addrs = 2001:db8:1234:ec::/120
        dns = 2001:db8:1234:1::1,2001:db8:1234:1::2
    }
}
```
It looks like there is no limit to 2 dns servers. You can input several with comma as delmiter and they're all pushed.
```
dns = 192.168.1.1,192.168.1.2,192.168.1.3,2001:db8:1234:1::1,2001:db8:1234:1::2,2001:db8:1234:1::3
```

**Describe the solution you like**

I would like additional configuration fields in VPN: IPsec: Connections Pools tab that allow to set DNS IPv4 (Value 3) and DNS IPv6 (Value 10) Servers. That way clients will be autoconfigured with the Virtual IP and DNS Servers.

**Describe alternatives you considered**

Configuring DNS manually in all roadwarrior configurations.

**Additional context**

[...]

PR #6864, opened 2023-09-19
VPN: IPsec: Connections: Pools add IP4_DNS and IP6_DNS Configuration Payloads

Monviech:
https://github.com/opnsense/core/issues/6860

Small explanation what I did:
- I assessed that I needed to tokenize the input and to turn it into an array and to check if the input is an ipv4 or ipv6 address. I didn't know how to do it though, so I searched for clues how the Connections ``Local addresses`` did it, because I needed the same behavior.
- I found the ``IKEAdressField.php`` which almost did what I needed, and copied it to ``PoolsDNSField.php`` and changed it a little so it only accepts IP Addresses.
- Then I added a new field ``pools.dns`` to ``dialogPool.xml`` to accept the new DNS inputs in the WEB GUI.
- Last I added  ``<dns type=".\PoolsDNSField"/>`` to the ``<Pools>`` definition of ``Swanctl.xml``

At that point it just worked as expected, I can add multiple IPv4 and IPv6 addresses into each individual pool and they get saved as array in the swanctl.conf. On connection, they're pushed to the client as IKE configuration payload.

### DIFF
--- a/src/opnsense/mvc/app/models/OPNsense/IPsec/Swanctl.xml
+++ b/src/opnsense/mvc/app/models/OPNsense/IPsec/Swanctl.xml
@@ -328,6 +328,7 @@
                     <NetMaskRequired>Y</NetMaskRequired>
                     <ValidationMessage>Please specify a valid CIDR subnet.</ValidationMessage>
                 </addrs>
+                <dns type=".\PoolsDNSField"/>
             </Pool>
         </Pools>
         <VTIs>
--- a/src/opnsense/mvc/app/controllers/OPNsense/IPsec/forms/dialogPool.xml
+++ b/src/opnsense/mvc/app/controllers/OPNsense/IPsec/forms/dialogPool.xml
@@ -18,5 +18,14 @@
           Accepts a single CIDR subnet defining the pool to allocate addresses from
         </help>
     </field>
-
+    <field>
+        <id>pool.dns</id>
+        <label>DNS</label>
+        <type>select_multiple</type>
+        <style>tokenize</style>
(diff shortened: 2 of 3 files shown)


Monviech:
@AdSchellevis This works, I'm not sure why though and if it was the right way to do it. :)

src/opnsense/mvc/app/models/OPNsense/IPsec/Swanctl.xml
```diff
@@ -328,6 +328,7 @@
                     <NetMaskRequired>Y</NetMaskRequired>
                     <ValidationMessage>Please specify a valid CIDR subnet.</ValidationMessage>
                 </addrs>
+                <dns type=".\PoolsDNSField"/>
```
AdSchellevis, line 331: @Monviech The `NetworkField` type (https://docs.opnsense.org/development/frontend/models_fieldtypes.html#networkfield) with some options set should be enough in this case, I don't think we need anything special here, other than that, this should indeed do the trick.

Monviech: Thank you for your feedback, my solution was way too complicated. xD
I added the NetworkField instead with the minimum of options needed to get the same result as before.

AdSchellevis: @Monviech yep, this should be all indeed. thanks!

Monviech:
@AdSchellevis 
Great, I will update the new roadwarrior doc I wrote to incorporate this soon.

PR #6867, opened 2023-09-20
Multiple text corrections

criadoperez:
Multiple text corrections


src/opnsense/service/templates/OPNsense/IDS/suricata.yaml
```diff
@@ -784,7 +784,7 @@ app-layer:
       mime:
         # Decode MIME messages from SMTP transactions
         # (may be resource intensive)
-        # This field supercedes all others because it turns the entire
+        # This field supersedes all others because it turns the entire
```
fichtner, line 787: the whole file is upstream and typos re owned there, but I'm happy to fix them on our end for now

fichtner:
Merged, thanks!

PR #6872, opened 2023-09-21
mvc: improve field validation message handling

fichtner:
* defaultValidationMessage() provides the field default (because gettext() doesn't work on constants)
* $internalValidationMessage only provides the XML value if given (translated via lang.git)
* getValidationMessage() returns the preferred validation message correctly translated


fichtner:
There are more fields to adjust but I tested these and the rest is trivial enough now that the framework has been established.

swhite2 (approved):
LGTM!

PR #6876, opened 2023-09-21
interfaces: make link-local vips unique per interface #6775

fichtner:
Had a few more minutes. I think the cleanup doesn't strip multiple LL addresses, but can deal with that later. Looks reasonable to me now since model validation does all sorts of things already.

### DIFF
--- a/src/opnsense/mvc/app/models/OPNsense/Interfaces/Vip.php
+++ b/src/opnsense/mvc/app/models/OPNsense/Interfaces/Vip.php
@@ -41,21 +41,33 @@ class Vip extends BaseModel
     public function performValidation($validateFullModel = false)
     {
         $messages = parent::performValidation($validateFullModel);
-        $vips = [];
+
+        $unqiue_addrs = [];
         $carp_vhids = [];
+        $vips = [];
 
         // collect changed VIP entries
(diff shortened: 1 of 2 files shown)


fichtner:
@AdSchellevis if you have some time this works on my end

src/opnsense/mvc/app/models/OPNsense/Interfaces/Vip.php
```diff
@@ -41,26 +41,43 @@ class Vip extends BaseModel
...
             }
+
             $vhid_key = sprintf("%s_%s", $parentNode->interface, $parentNode->vhid);
+
             if ((string)$parentNode->mode == 'carp' && !isset($carp_vhids[$vhid_key])) {
                 $carp_vhids[$vhid_key] = $parentNode;
             }
         }
 
+        foreach ($this->getFlatNodes() as $key => $node) {
```
AdSchellevis, line 68: @fichtner I think I would combine this with the `foreach ($this->getFlatNodes() as $key => $node) {` loop above, so we can reuse some of it and only need to iterate once. Other than that, this looks nice :)

fichtner: fair enough. I moved it out of the loop because at first I tried to fit it into `if ($validateFullModel || $node->isFieldChanged()) {` which wasn't the case.

PR #6879, opened 2023-09-22
openvpn: add up and down scripts to vpn instance configs

### LINKED ISSUE #6868: Gateway behind OpenVPN client instance does not get monitored (dpinger does not start)
**Important notices**

**Describe the bug**

When starting an OpenVPN client instance (`VPN > OpenVPN > Instances [new]`) OPNsense 23.7.x does not start a `dpinger` instance to monitor the corresponding gateway.

The client instance is connecting to a remote server on the default OpenVPN port using a certificate and a TLS static key. No settings were provided for the section `Authentication`. For `Routing` only `Remote Network` was provided as `10.9.0.0/24`. For `Miscellaneous` the option `route-nopull` is enabled.

The VPN Client is assigned as interface (`Interfaces > Assignments`) and a gateway is configured for the interface (`System > Gateways > Single`). The gateway has `10.9.0.1` specified as `Monitor IP`.

When starting the VPN Client instance the connection is established, but the gateway remains `Offline`. Checking for running `dpinger` instances using `ps auxww | grep dpinger` there is no instance monitoring the gateway of the VPN client. Manually disabling and re-enabling the gateway causes an instance of `dpinger` for the gateway to spawn and the gateway gets a valid status.

This used to be no problem back on 23.1: when using a VPN Client on OPNsense 23.1 the `dpinger` instance gets started automatically and the gateway get monitored. Disabling the VPN client keeps the `dpinger` instance around and upon reconnecting the `dpinger` instance gets removed and a new one gets started to replace it.

On a side note: back on 23.1 it was not even required to specify a `Monitor IP` for the gateway, the gateway IP would automatically be used if left blank (as the help text implies it should still do), but trying to disable / enable the gateway does not spawn an instance of `dpinger` if no explicit `Monitor IP` is specified on 23.7.4 - this might be another bug.

**Additional context**

This issue might be related to https://github.com/opnsense/core/issues/6862, considering that issue seems to be related to `dpinger` not properly getting restarted/reloaded when it should be, leading to `dpinger` delivering incorrect data. So there might be a few places (like VPN connections) missing a restart/reload. I stumbled upon this VPN issue during tests related to the other issue, but I had already noticed the gateway monitoring not working during a quick test on 23.7.1, but assumed it was related to https://github.com/opnsense/core/issues/6728 at the time.

**Expected behavior**

A gateway that is associated with a VPN instance should automatically be monitored when the VPN instance establishes a connection.

**Environment**

Versions:
 - OPNsense 23.7.4-amd64
 - FreeBSD 13.2-RELEASE-p3
 - OpenSSL 1.1.1v 1 Aug 2023

CPU type:
 - Intel(R) Xeon(R) CPU D-1518 @ 2.20GHz (4 cores, 8 threads)

Daggolin:
OpenVPN configs generated for new instances did not include instructions to call `ovpn-linkup` and `ovpn-linkdown`. Due to this `newip` would not trigger and `rc.newwanip` would not be called. For example this would lead to gateway monitors not starting.

Closes #6868

### DIFF
--- a/src/opnsense/mvc/app/models/OPNsense/OpenVPN/OpenVPN.php
+++ b/src/opnsense/mvc/app/models/OPNsense/OpenVPN/OpenVPN.php
@@ -499,6 +499,8 @@ class OpenVPN extends BaseModel
                 $options['proto'] = (string)$node->proto;
                 $options['verb'] = (string)$node->verb;
                 $options['verify-client-cert'] = (string)$node->verify_client_cert;
+                $options['up'] = '/usr/local/etc/inc/plugins.inc.d/openvpn/ovpn-linkup';
+                $options['down'] = '/usr/local/etc/inc/plugins.inc.d/openvpn/ovpn-linkdown';
 
                 foreach (
                     [


AdSchellevis:
I remember doubting about this one, but looking at it again, it does make sense to hook both events for this purpose. @Daggolin thanks!

Issue #6893, opened 2023-09-29
Interfaces: Other Types: VXLAN needs additional constraints to prevent user error


Monviech:
**Important notices**

Before you add a new report, we ask you kindly to acknowledge the following:

- [x] I have read the contributing guide lines at https://github.com/opnsense/core/blob/master/CONTRIBUTING.md
- [x] I am convinced that my issue is new after having checked both open and closed issues at https://github.com/opnsense/core/issues?q=is%3Aissue

**Describe the bug**

In the GUI in ``Interfaces: Other Types: VXLAN`` in the mask ``Edit VxLan`` the fields ``Remote address``, ``Multicast group`` and ``Device`` can all be populated, even though that prevents the operating system from creating a working vxlan interface.

If you populate ``Remote address`` and ``Device (other than none)`` at the same time, ``dmesg`` shows this:
``vxlan2: cannot initialize interface: can only specify interface with a group address``

If you populate ``Remote address`` and``Multicast group`` and ``Device`` at the same time, the vxlan interface comes up, but it's in multicast mode, ignoring the "Remote address" parameter:
``vxlan vni 42 local 172.16.0.254:4789 group 239.0.0.1:4789``

**To Reproduce**

1. Go to "Interfaces: Other Types: VXLAN" in "Edit VxLan"
- VNI: 42
- Source address: 172.16.0.254     (IP address of a firewall bound interface)
- Remote address: 172.16.0.186
- Multicast group: **empty**
- Device: LAN

2. Go to "Interfaces: Other Types: VXLAN" in "Edit VxLan"
- VNI: 43
- Source address: 172.16.0.254     (IP address of a firewall bound interface)
- Remote address: 172.16.0.186
- Multicast group: 239.0.0.1
- Device: LAN

**Expected behavior**

- When a ``Multicast group`` is set, it should be required to have ``Remote address`` empty, and the ``Device`` has to be NOT ``none``
- When a ``Remote address`` is set, it should be required to have ``Multicast group`` empty, and the ``Device`` has to be ``none``

**Additional context**

Heres the precise troubleshooting of my user error:

https://forum.opnsense.org/index.php?topic=36205.0

It was pretty hard to troubleshoot, so I think the additional constraints could be a valid addition. 

**Environment**

Software version used and hardware type if relevant, e.g.:

OPNsense 23.7.4-amd64
FreeBSD 13.2-RELEASE-p3
OpenSSL 1.1.1v 1 Aug 2023

[...]

PR #6899, opened 2023-09-30
Update VxLan.php - Add input validations to model

Monviech:
[#6893](https://github.com/opnsense/core/issues/6893)

That was rather challenging for me. But I think I got it working the way it should.

Here's some screenshots with the results:

Requirement for either Remote address or Multicast group NOT to be empty:
grafik

Requirement for both Remote address or Multicast group NOT to be populated at the same time:
grafik

Requirement for Device being set when Multicast group is populated:
grafik

Requirement for Device being NONE when Remote address is populated:
grafik

### DIFF
--- a/src/opnsense/mvc/app/models/OPNsense/Interfaces/VxLan.php
+++ b/src/opnsense/mvc/app/models/OPNsense/Interfaces/VxLan.php
@@ -1,7 +1,7 @@
 <?php
 
 /*
- * Copyright (C) 2019 Deciso B.V.
+ * Copyright (C) 2023 Deciso B.V.
  * All rights reserved.
  *
  * Redistribution and use in source and binary forms, with or without
@@ -28,8 +28,56 @@
 
 namespace OPNsense\Interfaces;
 
+use Phalcon\Messages\Message;
 use OPNsense\Base\BaseModel;
 
 class VxLan extends BaseModel
 {
+    public function performValidation($validateFullModel = false)
+    {
+        $messages = parent::performValidation($validateFullModel);
+
+        foreach ($this->vxlan->iterateItems() as $vxlan) {
+
+            $vxlangroup = (string) $vxlan->vxlangroup;
+            $vxlanremote = (string) $vxlan->vxlanremote;
+            $vxlandev = (string) $vxlan->vxlandev;
+
+            // Validation 1: At least one of vxlangroup and vxlanremote must be populated, but not both.
+            if ((!empty($vxlangroup) && !empty($vxlanremote)) ||
+                (empty($vxlangroup) && empty($vxlanremote))) {
+                $messages->appendMessage(new Message(
+                    "Remote address -or- Multicast group has to be specified",
+                    "vxlan.vxlanremote",
+                    "GroupOrRemote"
+                ));
+                $messages->appendMessage(new Message(
+                    "Multicast group -or- Remote address has to be specified",
+                    "vxlan.vxlangroup",
+                    "GroupOrRemote"
+                ));
+            }
+
+            // Validation 2: If vxlanremote is populated, vxlandev must be an empty string.
+             if (!empty($vxlanremote) && !empty($vxlandev)) {
+                $messages->appendMessage(new Message(
+                    "Remote address is specified, Device must be None",
+                    "vxlan.vxlandev",
+                    "DeviceRequirementForRemote"
+                    
+                ));
+            }
+
+            // Validation 3: If vxlangroup is populated, vxlandev must not be an empty string.
+             if (!empty($vxlangroup) && empty($vxlandev)) {
+                $messages->appendMessage(new Message(
+                    "Multicast group is specified, a Device must also be specified",
+                    "vxlan.vxlandev",
+                    "DeviceRequirementForGroup"
+                ));
+            }
+        }
+
+        return $messages;
+    }
 }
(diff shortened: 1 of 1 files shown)


AdSchellevis:
hehe, it's certainly a good starting point, in general I think there are a couple of small things missing. The messages should be passed though `gettext()` and you want to make sure the record inspected is actually changed (to prevent old invalid data not presented locking your save operation)

I can take a look later and offer you the commit, the weekend is a bit busy, but we're probably not in a hurry :)

Monviech:
Thank you very much for your feedback. This is by far the most challenging thing I ever changed in the OPNsense. And I'm not really in a hurry here. I really wanted to see if I could get a result. Maybe you've seen the debugging I did in the forum that resulted in this PR, I'm more firm as network or linux admin. Trying to actually change something in code is pretty new to me. xD

AdSchellevis:
I missed the forum part, but it has been extremely busy the last week, I just like it when people are making an effort with small and focussed changes. 

If you're interested, I can complete this one and annotate the changes in a separate commit so the rational between choices might be easier to understand.

Out of curiosity, how are you using OPNsense? seen quite some useful efforts from your end, thanks for that.

Monviech:
I will try and implement the gettext() - I've read that its for translation between languages, so I'm interested in seeing how that works. Also I want to try to "prevent old invalid data not presented locking your save operation" - But I would need a hint here how to actually test if something like this works. I have seen in the VIP.php file that at the beginning it loads all changed data. I might be able to do the same, so I wanne try it at least before passing the baton :)

Also I'm using OPNsense DEC Hardware with the business edition for customers and the company I work at. And of course the community edition for home use and tinkering.

Hope you'll have a nice weekend still. ^^

AdSchellevis:
Let me add some review notes, I can understand that you want to give it a shot, with some pointers it should be easier 🙂

I'm certainly enjoying my weekend, hope you are enjoying yours too.

src/opnsense/mvc/app/models/OPNsense/Interfaces/VxLan.php
```diff
@@ -28,8 +28,56 @@
...
+
+            $vxlangroup = (string) $vxlan->vxlangroup;
+            $vxlanremote = (string) $vxlan->vxlanremote;
+            $vxlandev = (string) $vxlan->vxlandev;
+
+            // Validation 1: At least one of vxlangroup and vxlanremote must be populated, but not both.
+            if ((!empty($vxlangroup) && !empty($vxlanremote)) ||
+                (empty($vxlangroup) && empty($vxlanremote))) {
+                $messages->appendMessage(new Message(
+                    "Remote address -or- Multicast group has to be specified",
```
AdSchellevis, line 50: If you wrap this in (e.g. `gettext('"Remote address -or- Multicast group has to be specified"')`), it should be collected using our crawler for translations. You wouldn't notice a difference yet, but after the phrase is being translated via our translation project (https://translate.opnsense.org/), it would return the correct language.

src/opnsense/mvc/app/models/OPNsense/Interfaces/VxLan.php
```diff
@@ -28,8 +28,56 @@
...
+use Phalcon\Messages\Message;
 use OPNsense\Base\BaseModel;
 
 class VxLan extends BaseModel
 {
+    public function performValidation($validateFullModel = false)
+    {
+        $messages = parent::performValidation($validateFullModel);
+
+        foreach ($this->vxlan->iterateItems() as $vxlan) {
```
AdSchellevis, line 40: here comes the tricky part in terms of what to validate. When `$validateFullModel` is set to `true`, everything should be handled, but the default is to only validate what has changed. That's why in most of our code you see constructions like these:

```
foreach ($this->getFlatNodes() as $key => $node) {
    if ($validateFullModel || $node->isFieldChanged()) {
       // XXX : some logic
    }
```

In this case, since our validation only covers 3 fields, you could also consider to traverse using the `iterateItems()` as you're already doing, but just check if any of the related fields has changed. Something like 

`$validateFullModel || $vxlan->vxlangroup->isFieldChanged() || ...`  When that's the case, it's safe to proceed.

When always enforcing validation, someone with two broken records in the set would not be able to fix his issue anymore due to being locked (on screen is one record, the other one will also report being broken),. Just try it out and you see what I mean.

AdSchellevis:
ok, the relevant hints should be in there now, I'm going offline now, but if you do run into issues, just make a note and I'll take a look later. 

Enjoy the rest of your weekend!

src/opnsense/mvc/app/models/OPNsense/Interfaces/VxLan.php
Monviech: Oh now I understand. I did test it now. I added 3 broken records, afterwards I added the validation and I tried to create a new entry. But I couldn't since the whole model was validated, and not only the changed records. Thanks! Now I know what I have to do. :)

src/opnsense/mvc/app/models/OPNsense/Interfaces/VxLan.php
Monviech: I added the gettext function to all messages.

Monviech:
I have tested this, and now I'm able to save old broken records because no fields have been changed. And I'm also able to create new records that are validated because fields have been changed.

AdSchellevis:
@Monviech nice! should work like a charm indeed. Thanks!

Issue #6888, opened 2023-09-27
Support for blocklists containing wildcard domains or in Unbound format.


nva14:
**Important notices**

Before you add a new report, we ask you kindly to acknowledge the following:

- [x] I have read the contributing guide lines at https://github.com/opnsense/core/blob/master/CONTRIBUTING.md
- [x] I am convinced that my issue is new after having checked both open and closed issues at https://github.com/opnsense/core/issues?q=is%3Aissue

**Is your feature request related to a problem? Please describe.**

Currently DNSBL feature shipped with Unbound in Opnsense only support list of FQDNs. Predefined lists in Opnsense are either not good enough or no longer maintained (Adguard and Easylist from Justdomains). Alternative sources like [https://oisd.nl](https://oisd.nl) will soon stop updating domain lists, they do however provide list in wildcard format of but it does not work in Opnsense.

There's also list in Unbound format but ironically doesn't work with Unbound in Opnsense: https://small.oisd.nl/unbound

I'm aware i can pull that file and put it in Unbound custom config directory but Opnsense config file wouldn't have that change.

**Describe the solution you like**

DNSBL should be able to handle wildcard domain, or in Unbound format.

--------
Thanks for considering.

fichtner:
While I can see this is a drop in config file I'm not sure if we can integrate those as is. My guess is they need to be parsed and rewritten. That might be more work than helpful to the cause, but @swhite2 would know more.

swhite2:
supporting the wildcard notation provided by oisd (or other providers) shouldn't be too hard. The Unbound notation sounds logical, but the blocklist implementation has moved away from this, and multiple providers often present multiple formats for the same blocklist, such as oisd - so keeping it generic seems like the better option.

Also, it seems this blocklist is rather active, I'll include it in the options list in the GUI.

kjkent:
@swhite2 just wanted to say thank you for including OISD.nl in the OPNsense GUI. Especially now that the former has dropped FQDN lists, it's a great convenience. 🙏🏻

swhite2:
@kjkent You're very welcome, glad to hear this has been a valuable addition for you :)

PR #6900, opened 2023-09-30
parsing custom blocklist with wildcard domains

nva14:
Pull request to implement a small change to accommodate the feature i suggested in issue [#6888](https://github.com/opnsense/core/issues/6888).

Currently blocklist handler is already able tag a domain as wildcard with the property `'wildcard': True` for Unbound to pick up. I think it's trivial to add support for wildcard domains in the format `*.example.com` from custom blocklist with minimal change in logic.

New code only attempt to alter the parsing of custom blocklist and does not touch other inputs.


handfreezer:
Hello, as this pull request was closed, how does wildcard is working into personnal URL blocklist ?

PR #6905, opened 2023-10-02
unbound: add support for wildcard domain lists

swhite2:
Also add 3 lists that contain wildcard domains. Removed the "exclude non domain entries" regex pattern since all domains are filtered through a larger domain pattern that now includes the wildcard symbol as well.


AdSchellevis (approved):

AdSchellevis:
@swhite2 looks good! thanks

PR #6908, opened 2023-10-03
dpinger: improve validation for 'time period'

### LINKED ISSUE #6907: OPNsense does not validate a gateway's 'Time Period' against its 'Loss Interval'
**Important notices**

**Describe the bug**

A user can set a gateway's 'Time Period' to a value that is too short (i.e. less than 2 times the sum of the 'Probe Interval' and 'Loss Interval'). This results in `dpinger` being unable to start, with no relevant log messages available anywhere.

**To Reproduce**

1. Go to 'System > Gateways > Single'
2. Edit a gateway
   - Set 'Time Period' to '6'
   - Make sure that 'Probe Interval' and 'Loss Interval' are unset (or set to '1' and '4', respectively)
   - Make sure that 'Disable Gateway Monitoring' is _not_ checked
3. Click 'Save', then 'Apply changes'
   - See that the gateway is now marked as 'Offline'
4. Go to 'System > Diagnostics > Services'
   - See that `dpinger` for that gateway is stopped
5. Attempt to start `dpinger` for that gateway
   - See that it does not start
   - See that no reason is given anywhere as to why it does not start

**Expected behavior**

The user is prompted that the 'Time Period' is too short, and not allowed to save changes to the gateway.

**Additional context**

The relevant check in `dpinger` can be seen at https://github.com/dennypage/dpinger/blob/v3.3/dpinger.c#L1080

OPNsense currently only checks that the 'Time Period' is greater than 2.1 times the 'Probe Interval', ignoring the 'Loss Interval' completely (see https://github.com/opnsense/core/blob/23.7.5/src/www/system_gateways_edit.php#L263).

Apologies if the resulting `dpinger` error is actually logged somewhere... I couldn't find it, though.

**Environment**

OPNsense 23.7.5-amd64

abraha2d:
FIxes #6907.

### DIFF
--- a/src/opnsense/mvc/app/models/OPNsense/Routing/Gateways.php
+++ b/src/opnsense/mvc/app/models/OPNsense/Routing/Gateways.php
@@ -93,7 +93,7 @@ class Gateways extends BaseModel
                 if (in_array($tagName, ['ipprotocol', 'gateway'])) {
                     $this->validateDynamicMatch($parent, $messages, $ref);
                 }
-                if (in_array($tagName, ['latencylow', 'latencyhigh', 'losslow', 'losshigh', 'time_period', 'interval'])) {
+                if (in_array($tagName, ['latencylow', 'latencyhigh', 'losslow', 'losshigh', 'time_period', 'interval', 'loss_interval'])) {
                     $this->validateDpingerSettings($tagName, $parent, $messages, $ref);
                 }
             }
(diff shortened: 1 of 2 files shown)


fichtner:
This is bad timing. The page is going to be replaced, but the code is not in the master branch yet.

abraha2d:
I've updated the PR to point to (what seems like) the branch for the new page.

swhite2:
Fair enough, do make sure to trigger the validation on `loss_interval` change as well: https://github.com/opnsense/core/blob/gateways/src/opnsense/mvc/app/models/OPNsense/Routing/Gateways.php#L96.

We likely need to account for this in the migration and revert to the defaults if necessary, assuming people have configured but ignored this.

abraha2d:
Done. Also added a bit to the migration, but not a fan of the way it duplicates the validation logic. Thoughts?

swhite2:
The migration code is always custom to the issue at hand and should ideally only have to be run once. I see no benefit in trying to reduce duplication here as it helps to be more explicit on what it's trying to achieve.

swhite2 (approved):
LGTM :)

PR #6927, opened 2023-10-11
ipsec: make Description in Connections Required

mimugmail:
Hi,

it seems that when leaving the field empty, you have an empty dropdown list in authentication and children, it's a bit weird when you forget to fill in the field it's quite hard to find the error why you cannot link the newly created connection.

Happy if you find an alternative to this PR :)

### DIFF
--- a/src/opnsense/mvc/app/models/OPNsense/IPsec/Swanctl.xml
+++ b/src/opnsense/mvc/app/models/OPNsense/IPsec/Swanctl.xml
@@ -1,6 +1,6 @@
 <model>
     <mount>//OPNsense/Swanctl</mount>
-    <version>1.0.0</version>
+    <version>1.0.1</version>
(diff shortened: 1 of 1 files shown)


src/opnsense/mvc/app/models/OPNsense/IPsec/Swanctl.xml
```diff
@@ -1,6 +1,6 @@
 <model>
     <mount>//OPNsense/Swanctl</mount>
-    <version>1.0.0</version>
+    <version>1.0.1</version>
```
fichtner, line 3: No need for this, the value cannot be set by migration /default
```suggestion
    <version>1.0.0</version>
```

fichtner (approved):

AdSchellevis:
@mimugmail thanks!

Issue #6986, opened 2023-11-05
Unbound DNS Access Lists - remove the localhost automatic handling


doktornotor:
**Important notices**

Before you add a new report, we ask you kindly to acknowledge the following:

- [x] I have read the contributing guide lines at https://github.com/opnsense/core/blob/master/CONTRIBUTING.md
- [x] I am convinced that my issue is new after having checked both open and closed issues at https://github.com/opnsense/core/issues?q=is%3Aissue

**Describe the bug**

The default list that includes 127.0.0.1/8 and ::1/128 produces silly resulting configuration when you want to set up something more restricted. Note that it does not actually block the localhost queries anyway but it is confusing in general and makes users wonder what actually will be used. [Localhost is allowed by default in Unbound](https://unbound.docs.nlnetlabs.nl/en/latest/manpages/unbound.conf.html#unbound-conf-access-control), there is no need to set it up explicitly in the "default allow" mode, and the behaviour is generally not wanted when you change the default action.

> By default only localhost is allowed, the rest is refused. The default is refused, because that is protocol-friendly. The DNS protocol is not designed to handle dropped packets due to policy, and dropping may result in (possibly excessive) retried queries.

**To Reproduce**

1. Go to 'Services: Unbound DNS: Access Lists
2. Select 'Block/Refuse' as Default action and apply.

```
# cat /var/unbound/etc/access_lists.conf
server:
access-control: 0.0.0.0/0 refuse
access-control: ::/0 refuse
access-control: 127.0.0.1/8 refuse
access-control: ::1/128 refuse

```

**Expected behavior**
No entries for localhost.

**Describe alternatives you considered**

Ignoring the issue and resulting confusion. 😉

**Additional context**

Consider the following example, hopefully it makes the potential confusion more clear:

```
# cat /var/unbound/etc/access_lists.conf
server:
access-control: 0.0.0.0/0 refuse
access-control: ::/0 refuse
access-control: 127.0.0.1/8 refuse
access-control: ::1/128 refuse

# thisfirewall
access-control: 127.0.0.1/32 allow_snoop
access-control: ::1/128 allow_snoop
# localnets
access-control: 192.168.255.0/24 allow
access-control: 2001:dead:beef:db00::/56 allow

```

**Environment**

Software version used and hardware type if relevant, e.g.:

OPNsense 23.7.7_3 (amd64)

AdSchellevis:
[...]

PR #6987, opened 2023-11-05
Remove localhost from automatically created Unbound ACL (#6986)

doktornotor:
Fixes https://github.com/opnsense/core/issues/6986


AdSchellevis:
@doktornotor thanks!

Issue #6971, opened 2023-10-31
dhcp: offer KEA as DHCPv4 alternative


AdSchellevis:
**Important notices**

Before you add a new report, we ask you kindly to acknowledge the following:

- [x] I have read the contributing guide lines at https://github.com/opnsense/core/blob/master/CONTRIBUTING.md
- [x] I am convinced that my issue is new after having checked both open and closed issues at https://github.com/opnsense/core/issues?q=is%3Aissue

**Is your feature request related to a problem? Please describe.**

With isc-dhcp being eol'd (https://www.isc.org/blogs/isc-dhcp-eol/), we should start aiming for a successor on our end. The number of realistic options is relatively small unfortunately.  DNSmasq looked nice in terms of options, but lacks support for high availability, which in the long run might be problematic. 

**Describe the solution you like**

Offer KEA DHCP (https://www.isc.org/kea/) as optional alternative for the existing DHCPv4 server. Since the dhcp code is rather old, we shouldn't try to aim for a migration at the moment to prevent wasting precious time in trying to solve past inconsistencies.  

**Describe alternatives you considered**

Mostly DNSmasq, small and simple.

**Notes and findings**

- Removing leases manually is likely not the greatest idea in the world, kea doesn't seem to support manual removal of already offered leases. Cleanups of the local file are being handled via https://kea.readthedocs.io/en/kea-2.0.2/man/kea-lfc.8.html , the current approach is to remove entries and restart the daemon which is very disruptive and also not recommended.

bueste:
Please add the feautre to add custom dhcp ranges (which aren't interfaces on the opensense) for the dhcp. I dont want to have separate dhcp servers anymore in a setup witch L3 Switches and a transit route to opensense.

rschell:
Need to add "ip_address" to the reservation table, currently just lists "subnet", "MAC", "Description" and "Commands".

rschell:
Understand it’s a work in progress.  Any plans for opnsense to convert the isc-dhcp settings to this new service?

A hostname field would be a nice addition to the table as well.

rschell:
Had used Kea-dhcp before in plus-sense land before my transition to OPNsense.  The migration was painless and didn’t have any issues with it.

Only had it briefly enabled in OPNsense.  Noticed that I had to populate my 30+ static assignments first.  That’s done, not sure the next testing steps now.   An active lease table maybe needed to track its operation.  I assume that simultaneous operation of both dhcp services would not be desirable.

[...]

PR #6999, opened 2023-11-10
Kea / Dhcp - DHCPv4 replacement [#6971], Add "IP Address" to KEA reservations Table

rschell:
The device's IP address is missing from the KEA reservations table. See #6971


AdSchellevis:
@rschell thanks, that is quick, kea is still a work in progress on our end, already receiving a PR to fix an issue is nice :)

PR #7001, opened 2023-11-10
www/services_dhcp_edit: Allow for Cisco-style MAC addresses.

jurajlutter:
Allow for Cisco-style (XXXX.XXXX.XXXX-formatted) addresses.

### DIFF
--- a/src/www/services_dhcp_edit.php
+++ b/src/www/services_dhcp_edit.php
@@ -101,6 +101,17 @@ if ($_SERVER['REQUEST_METHOD'] === 'GET') {
     }
 
     /* normalize MAC addresses - lowercase and convert Windows-ized hyphenated MACs to colon delimited */
+    /* Also convert cisco-style MAC address to colon-separated form */
+    f (preg_match("/^(..)(..)\.(..)(..)\.(..)(..)$/", $pconfig['mac'],
+	    $_macmatches)) {
+	    $pconfig['mac'] = sprintf("%02s:%02s:%02s:%02s:%02s:%02s",
+		    $_macmatches[1],
+		    $_macmatches[2],
+		    $_macmatches[3],
+		    $_macmatches[4],
+		    $_macmatches[5],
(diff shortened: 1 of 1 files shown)


AdSchellevis:
I'm not sure this is worth the effort for a service that is eol (https://www.isc.org/blogs/isc-dhcp-eol/), pushing more in there will make it harder in the long run. We're currently working on adding kea as an alternative.

jurajlutter:
Then this could also be implemented for KEA (I'm not sure whether KEA does support this format out of the box).

AdSchellevis:
By my knowledge both don't support these (https://kea.readthedocs.io/en/kea-2.2.0/arm/dhcp4-srv.html#host-reservations-in-dhcpv4), translations only increase the codebase without much value in my humble opinion.

fichtner:
I wouldn't mind allowing this format as long as we would normalize it in the model. Obviously that's easier done in the MVC code where we have a MAC field type now I think. Otherwise this would be too much code. The actual question is if that would be acceptable @jurajlutter  (your proposal would indicate it is)

jurajlutter:
This feature is beneficial in situation where you log into a switch, do like `show mac address-table` and copypasta the mac address directly into OPNsense UI.

AdSchellevis:
To do this properly, data should probably be cleansed in the browser (javascript), so you don't risk validation messages about X where the user entered Y (due to cleansing before validation), but since we don't do that anywhere in our mvc code at the moment, I don't think it's a good idea to organize our next issue. 

(keep in mind this is a very small use-case, even Cisco expects regular formatted addresses in quite some cases, for example: https://content.cisco.com/chapter.sjs?uri=/searchable/chapter/content/en/us/td/docs/routers/nfvis/switch_command/b-nfvis-switch-command-reference/address_table_commands.html.xml )

jurajlutter:
Thank you for the constructive approach.

PR #7019, opened 2023-11-17
interfaces: revamp overview page

swhite2:
(eventually) replaces the old overview page, with the intent of quicker debugging by presenting the most relevant information that could indicate faulty configuration.

last remaining bit is cleaning up `get_interfaces_info()`, but the `interface_list` widget is still using a callout, this should be replaced by the already present and adjusted callback mechanism in `src/www/widgets/api/plugins/interfaces.inc`.

This commit omits some of the link-specific information such as ppp uptime, disconnect/release mechanism, as well as wireless and bridge information. Since there is more of this type of information available than was originally being handled by `get_interfaces_info()`, perhaps it makes more sense to extend the backend script with the relevant bits in time.


src/opnsense/scripts/interfaces/ifinfo.php
```diff
@@ -0,0 +1,110 @@
+#!/usr/local/bin/php
```
fichtner, line 1: Would be nicer to fold this into pluginctl, very similar functionality already exists there.

src/www/widgets/api/plugins/interfaces.inc
```diff
@@ -30,37 +30,31 @@
...
-                $interfaceItem['status'] = $ifinfo['status'];
-                break;
-            case 'associated':
-                $interfaceItem['status'] = 'up';
-                break;
-            default:
-                $interfaceItem['status'] = '';
-                break;
+    $result = [];
+    $ifsinfo = json_decode((new OPNsense\Core\Backend())->configdpRun('interface list info', [null, true]), true);
```
fichtner, line 34: Idiom is „ifinfos“ => „ifinfo“ 😉

src/www/widgets/api/plugins/interfaces.inc
```diff
@@ -30,37 +30,31 @@
...
+        if (empty($ifinfo['config']) || (!empty($ifinfo['config']['virtual']))) {
+            /* not configured */
+            continue;
         }
-        $interfaceItem['ipaddr'] = empty($ifinfo['ipaddr']) ? "" : $ifinfo['ipaddr'];
-        $interfaceItem['media'] = empty($ifinfo['media']) ? "" : $ifinfo['media'];
+        $config = $ifinfo['config'];
+        $stats = $ifinfo['statistics'];
+        $interfaceItem = [
+            'inpkts' => $stats['packets received'],
```
fichtner, line 43: Can we make the widget use flatter by using the internal names?

src/www/widgets/include/interface_list.inc
```diff
@@ -1,4 +1,4 @@
 <?php
 
 $interface_list_title = gettext('Interfaces');
-$interface_list_title_link = 'status_interfaces.php';
+$interface_list_title_link = 'ui/interfaces/overview';
```
fichtner, line 4: Seeing they both point to the same page wouldn’t it make more sense to point this to the assignments page?

src/www/widgets/api/plugins/interfaces.inc
swhite2: yes, the widget needs a bit more work as it's the last remaining caller of `get_interfaces_info()`.

src/opnsense/scripts/interfaces/ifinfo.php
swhite2: sure :)

src/www/widgets/include/interface_list.inc
swhite2: Good idea :)

swhite2:
note to self: `get_interfaces_info()` and `format_bytes` (`util`) can be cleaned up once `status_overview.php` has been removed.

The requirements to format bytes has appeared at least 3 times now in javascript code, it probably makes sense to move this one level up to the default includes

src/www/widgets/widgets/interface_list.widget.php
```diff
@@ -112,61 +145,27 @@ function interface_widget_update(sender, data)
...
+?>
+  <tr id="interface_widget_item_<?= html_safe($ident) ?>">
+    <td style="width:5%; word-break: break-word;" id="link_status"></td>
+    <td style="width:15%; word-break: break-word;" id="if_name"></td>
+    <td style="width:5%; word-break: break-word;" id="if_status"></td>
+    <td style="width:32%; word-break: break-word;" id="if_media"></td>
     <td style="width:43%; word-break: break-word;">
-      <?= html_safe($ifinfo['ipaddr']) ?>
-      <?= !empty($ifinfo['ipaddr']) ? '<br/>' : '' ?>
-      <?= html_safe(interfaces_has_prefix_only($ifdescr) ? $ifinfo['linklocal'] : $ifinfo['ipaddrv6']) ?>
```
swhite2, line 169: @fichtner this one is annoying, any particular opinion about it?

fichtner: interfaces_primary_address6() should do exactly this to be honest

fichtner (approved):
Feel free to merge this to progress and make it easier to test / review UI/UX

AdSchellevis (approved):

ohartl:
This implements #7019 doesn't it?

AdSchellevis:
@ohartl it does, together with https://github.com/opnsense/core/commit/4d495ea6cc3d3c993662867bf3b3a072069e1fc8 to delete the old cruft.

Issue #1347, opened 2017-01-24
Feature Request: Bind Web GUI / SSH to specific interfaces or IPs


willjones:
It would be really nice to have an option to easily configure what interfaces the web GUI and SSH processes listen on.

By default they are active on all IPs on the firewall, this potentially exposes management access to networks that people would maybe not want to, whilst we can obviously solve this with a firewall rule it strikes me as maybe being better to allow folks to specify an interface or IP that these services bind to.

Potentially this saves a lot of firewall rules too in situations with a large number of internal networks where exposing access to the GUI or SSH would not be desirable at all.

fichtner:
(definitely maybe)

rptxcosmo:
+1
Same need.

fichtner:
will be done on our way to 18.1, the code preparations have been done for 17.7.

fraenki:
@fichtner Is this still on the roadmap for 18.1?

fichtner:
It's not on the roadmap per se, but I still have this in my priority list for 18.1.

fichtner:
Alright, so we're going to do this:

* Interface selection for SSH and web GUI separately
* Selected interfaces will be expanded in primary IP addresses and all alias / carp configurations automatically, so no manual IP stuff...
* The default will be none, which is nicely backwards-compatible and the least worse in terms of operational impact ;)
* Dynamic reload for DHCP-configured interfaces will be supported
* There will not be a lot of sanity checking. To stress this point, if all manually configured interfaces do not have a single IP listening address, the service will refuse to start as opposed to falling back to listen on all interfaces...
* Use at your own risk. It's hard to recover without other precautions like console access, auto-console login, etc.

Thoughts? :)

fraenki:
@fichtner All I've ever wanted! :D I'm absolutely OK with having no sanity checking and defaulting to a backwards-compatible config.

fichtner:
Thanks, full speed ahead then. :) For recovery, I'm planning this as people will run into issues for sure. https://github.com/opnsense/core/issues/1966

fraenki:
> Dynamic reload for DHCP-configured interfaces will be supported

After thinking about this a little longer I have to ask: A reload will _only_ be triggered for DHCP, right? It would be cumbersome if the reload would happen if a VPN tunnel (IPsec, OpenVPN) is reconfigured/reconnected.

fichtner:
[...]

PR #7032, opened 2023-11-25
webgui + openssh: listen on IPv6 link-local addresses

maurice-w:
WebGUI and SSH connections to link-local addresses only work if the listen interfaces are set to "All (recommended)". This PR allows link-local connections for specific listen interfaces, too. While selecting specific interfaces is only recommended for some use cases, explicitly excluding link-local addresses doesn't seem appropriate. Maybe there once was a bug which required this?

### DIFF
--- a/src/etc/inc/plugins.inc.d/openssh.inc
+++ b/src/etc/inc/plugins.inc.d/openssh.inc
@@ -202,11 +202,6 @@ function openssh_configure_do($verbose = false, $interface = '')
     $listeners = array();
 
     foreach (interfaces_addresses($interfaces) as $tmpaddr => $info) {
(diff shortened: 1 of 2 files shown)


fichtner:
Yes, there was a bug with these not working after the fact, but the commit history doesn’t say a lot more than that. Let’s try this again assuming you tested it 😉

fichtner:
Merged, thanks!

fichtner:
A bit of history here https://github.com/opnsense/core/issues/1347#issuecomment-351916705

maurice-w:
Yes, I tested this for both ssh and webgui on two systems. Can't guarantee that nothing breaks under obscure circumstances of course...

PR #7048, opened 2023-12-03
Vxlan - added vxlanlocalport and vxlanremoteport

Monviech:
I've chosen a pragmatic approach for this:

- Require the new fields not to be empty
- Fill them with the default value 4789
- Hide the "vxlanlocalport" and "vxlanremoteport" with advanced
- Always add them to all vxlan interfaces that are created
- Bumped the model version so existing configurations will be migrated to be with port numbers

If this approach is alright, I have tested it with unicast configuration and multicast group. Here's how it looks like in ifconfig:

```
vxlan1: flags=8802<BROADCAST,SIMPLEX,MULTICAST> metric 0 mtu 1450
        options=80020<JUMBO_MTU,LINKSTATE>
        ether 58:9c:fc:00:6d:6d
        groups: vxlan
        vxlan vni 12 local 192.168.3.1:4789 remote 192.168.3.2:4789
        media: Ethernet autoselect (autoselect <full-duplex>)
        status: active
vxlan2: flags=8802<BROADCAST,SIMPLEX,MULTICAST> metric 0 mtu 1450
        options=80020<JUMBO_MTU,LINKSTATE>
        ether 58:9c:fc:10:ff:e0
        groups: vxlan
        vxlan vni 13 local 192.168.3.1:4785 remote 192.168.3.4:4785
        media: Ethernet autoselect (autoselect <full-duplex>)
        status: active
vxlan3: flags=8802<BROADCAST,SIMPLEX,MULTICAST> metric 0 mtu 1450
        options=80020<JUMBO_MTU,LINKSTATE>
        ether 58:9c:fc:10:6e:52
        groups: vxlan
        vxlan vni 14 local 192.168.3.1:4785 group 227.0.0.1:4785
        media: Ethernet autoselect (autoselect <full-duplex>)
        status: active
```

### DIFF
--- a/src/etc/inc/plugins.inc.d/vxlan.inc
+++ b/src/etc/inc/plugins.inc.d/vxlan.inc
@@ -97,7 +97,9 @@ function vxlan_configure_do($verbose = false, $device = null)
         $current_settings = [
             'vxlanid' => null,
             'vxlanlocal' => null,
+            'vxlanlocalport' => null,
             'vxlanremote' => null,
+            'vxlanremoteport' => null,
             'vxlangroup' => null,
         ];
 
@@ -109,7 +111,9 @@ function vxlan_configure_do($verbose = false, $device = null)
             $isChanged = false;
             $current_settings['vxlanid'] = $interfaces_details[$device_name]['vxlan']['vni'];
             $current_settings['vxlanlocal'] = explode(":", $interfaces_details[$device_name]['vxlan']['local'])[0];
+            $current_settings['vxlanlocalport'] = explode(":", $interfaces_details[$device_name]['vxlan']['localport'])[0];
             $current_settings['vxlanremote'] = explode(":", $interfaces_details[$device_name]['vxlan']['remote'])[0];
+            $current_settings['vxlanremoteport'] = explode(":", $interfaces_details[$device_name]['vxlan']['remoteport'])[0];
             if (!empty($interfaces_details[$device_name]['vxlan']['group'])) {
                 $current_settings['vxlangroup'] = explode(":", $interfaces_details[$device_name]['vxlan']['group'])[0];
             }
@@ -117,7 +121,7 @@ function vxlan_configure_do($verbose = false, $device = null)
         // gather settings, detect changes
         $ifcnfcmd = '/sbin/ifconfig %s';
         $ifcnfcmdp = array($device_name);
-        foreach (array('vxlanid', 'vxlanlocal', 'vxlanremote', 'vxlangroup', 'vxlandev') as $param) {
+        foreach (array('vxlanid', 'vxlanlocal', 'vxlanlocalport', 'vxlanremote', 'vxlanremoteport', 'vxlangroup', 'vxlandev') as $param) {
             $value = '';
             if ($param == 'vxlandev') {
                 $intfnm = (string)$vxlan->$param;
--- a/src/etc/inc/plugins.inc.d/vxlan.inc
+++ b/src/etc/inc/plugins.inc.d/vxlan.inc
@@ -97,8 +97,8 @@ function vxlan_configure_do($verbose = false, $device = null)
         $current_settings = [
             'vxlanid' => null,
             'vxlanlocal' => null,
-            'vxlanlocalport' => null,
             'vxlanremote' => null,
+            'vxlanlocalport' => null,
             'vxlanremoteport' => null,
             'vxlangroup' => null,
         ];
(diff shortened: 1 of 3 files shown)


src/etc/inc/plugins.inc.d/vxlan.inc
```diff
@@ -109,15 +111,17 @@ function vxlan_configure_do($verbose = false, $device = null)
             $isChanged = false;
             $current_settings['vxlanid'] = $interfaces_details[$device_name]['vxlan']['vni'];
             $current_settings['vxlanlocal'] = explode(":", $interfaces_details[$device_name]['vxlan']['local'])[0];
+            $current_settings['vxlanlocalport'] = explode(":", $interfaces_details[$device_name]['vxlan']['localport'])[0];
```
fichtner, line 114: This looks suspicious. The port should be at [1] of the previous “:” explode() line.

src/etc/inc/plugins.inc.d/vxlan.inc
```diff
@@ -109,15 +111,17 @@ function vxlan_configure_do($verbose = false, $device = null)
...
             $current_settings['vxlanremote'] = explode(":", $interfaces_details[$device_name]['vxlan']['remote'])[0];
+            $current_settings['vxlanremoteport'] = explode(":", $interfaces_details[$device_name]['vxlan']['remoteport'])[0];
```
fichtner, line 116: Same here

src/opnsense/mvc/app/controllers/OPNsense/Interfaces/forms/dialogVxlan.xml
```diff
@@ -20,6 +20,13 @@
           When the interface is configured in unicast mode, the listening socket is bound to this address.
         </help>
     </field>
+    <field>
+        <id>vxlan.vxlanlocalport</id>
+        <label>Source port</label>
+        <type>text</type>
+        <help>Define the source port. Default is 4789.</help>
```
fichtner, line 27: If the goal is to default to that value automatically I don’t think it works. We should not break prior setups so leaving empty must be allowed or migrated, but I see no reason to migrate and force people to enter a default port from now on especially when hidden under advance d.

fichtner (changes requested):

src/opnsense/mvc/app/controllers/OPNsense/Interfaces/forms/dialogVxlan.xml
Monviech: Thanks for the feedback. I will change the approach to it.

fichtner: Thanks. Don’t mind me nagging. You are doing good 👍

Monviech: Thank you very much. I don't mind it at all, because you are right about these things. I'm very open minded in general, and I always think that what I try can be improved.  :+1:

Monviech:
@fichtner Can you give this another look? I have removed the explode statements because they're not necessary in my opinion. I have changed the model to make the port fields non required and empty by default.

I have tested all combinations in the GUI, and the ifconfig interfaces all get created and destroyed upon deletion. The isChanged flag also correctly detects changed ports when they're modified and the ifconfig vxlan interfaces are changed accordingly.

src/etc/inc/plugins.inc.d/vxlan.inc
```diff
@@ -108,16 +110,18 @@ function vxlan_configure_do($verbose = false, $device = null)
         } else {
             $isChanged = false;
             $current_settings['vxlanid'] = $interfaces_details[$device_name]['vxlan']['vni'];
-            $current_settings['vxlanlocal'] = explode(":", $interfaces_details[$device_name]['vxlan']['local'])[0];
-            $current_settings['vxlanremote'] = explode(":", $interfaces_details[$device_name]['vxlan']['remote'])[0];
+            $current_settings['vxlanlocal'] = $interfaces_details[$device_name]['vxlan']['local'];
+            $current_settings['vxlanlocalport'] = $interfaces_details[$device_name]['vxlan']['localport'];
```
AdSchellevis, line 114: @Monviech this looks a bit odd `legacy_interfaces_details()` returns `local` and `remote` as `address:port` combination. Where should the `localport` come from in this case?

Monviech: You are totally right that it is odd. I made some php debug statements and it shows exactly what you pointed out. I have to look more into this.

```
[11-Dec-2023 20:03:24 Etc/UTC] vxlan_configure_do started. Verbose: true, Device: null
[11-Dec-2023 20:03:24 Etc/UTC] Processing device: vxlan1
[11-Dec-2023 20:03:24 Etc/UTC] Before setting current_settings for vxlan1
[11-Dec-2023 20:03:24 Etc/UTC] interfaces_details for vxlan1: Array
(
    [vni] => 50
    [local] => 192.168.3.1:5000
    [group] =>
    [remote] => 192.168.1.2:5000
)
[11-Dec-2023 20:03:24 Etc/UTC] After setting current_settings for vxlan1
[11-Dec-2023 20:03:24 Etc/UTC] Current settings: Array
(
    [vxlanid] => 50
    [vxlanlocal] => 192.168.3.1:5000
    [vxlanlocalport] =>
    [vxlanremote] => 192.168.1.2:5000
    [vxlanremoteport] =>
    [vxlangroup] =>
)
```

fichtner: See my original comment on the explode() use and how the port should be on [1] for local and remote.

Monviech: @fichtner I should have looked into this more thoroughly. Sorry.

src/etc/inc/plugins.inc.d/vxlan.inc
```diff
@@ -110,14 +112,29 @@ function vxlan_configure_do($verbose = false, $device = null)
             $current_settings['vxlanid'] = $interfaces_details[$device_name]['vxlan']['vni'];
             $current_settings['vxlanlocal'] = explode(":", $interfaces_details[$device_name]['vxlan']['local'])[0];
             $current_settings['vxlanremote'] = explode(":", $interfaces_details[$device_name]['vxlan']['remote'])[0];
-            if (!empty($interfaces_details[$device_name]['vxlan']['group'])) {
-                $current_settings['vxlangroup'] = explode(":", $interfaces_details[$device_name]['vxlan']['group'])[0];
+            $current_settings['vxlanlocalport'] = explode(":", $interfaces_details[$device_name]['vxlan']['local'])[1];
```
AdSchellevis, line 115: @Monviech at a first glance the assessment in the PR comment looks correct, but we can simplify this quite a bit in my opinion. Looking at the ticket we're trying to set "local" and "remote" so "group" would be good as it was, although we can simplify it in the same loop.

Something like this might save some code:

```
foreach (['local', 'remote', 'group'] as $target) {
    if (!empty($current_settings[$interfaces_details[$device_name]['vxlan'][$target]]) {
        $tmp = explode(':', $interfaces_details[$device_name]['vxlan'][$target]]);
        $current_settings['vxlan'.$target] = $tmp[0];
        $current_settings['vxlan'.$target.'port'] = $tmp[1];
    }
}
```

AdSchellevis:
there you go https://github.com/opnsense/core/commit/dea5f295ac3161e47966328ff2611fb3f51a72ff

PR #7054, opened 2023-12-05
core / log views - use built-in selectpicker event to handle async init

kulikov-a:
Hi!
Can we try using built-in event to handle async `selectpicker` init (should be faster and more reliable)?
sorry, i’m also not very clear about class check part. if this is about https://github.com/snapappointments/bootstrap-select/issues/2756 , do we really need to care about the select class if we assign the `selectpicker` not to the class, but to the id?

Thanks!
upd. one moment..

### DIFF
--- a/src/opnsense/mvc/app/views/OPNsense/Diagnostics/log.volt
+++ b/src/opnsense/mvc/app/views/OPNsense/Diagnostics/log.volt
@@ -151,12 +151,14 @@
           let select = $("#severity_filter");
           let header_val = filter_exact ? m_header : s_header;
 
+          // switch select mode and reinit selectpicker
           select.prop("multiple", filter_exact);
-          // switch select mode and destroy selectpicker
-          if (!select.hasClass('severity_filter_init')) {
-              select.addClass('severity_filter_init');
-          } else {
-              select.selectpicker('destroy');
+
+          // destroy, reinit & reopen select after load if called from mode switch
+          if (event && (event.currentTarget.id == 'exact_severity')) {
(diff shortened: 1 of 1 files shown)


src/opnsense/mvc/app/views/OPNsense/Diagnostics/log.volt
```diff
@@ -151,12 +151,14 @@
...
+          // switch select mode and reinit selectpicker
           select.prop("multiple", filter_exact);
-          // switch select mode and destroy selectpicker
-          if (!select.hasClass('severity_filter_init')) {
-              select.addClass('severity_filter_init');
-          } else {
-              select.selectpicker('destroy');
+          select.selectpicker('destroy');
+          // keep select open after reinit if called from mode switch
+          if (event && (event.currentTarget.id = 'execat_severity')) {
```
AdSchellevis, line 158: `exact_severity` ?

kulikov-a: sorry, sure )
funny but that was copy/paste from chrome console (i thought)

kulikov-a:
@AdSchellevis hm. When re-init was removed from page load, the interference appeared of auto-init (by 'selectpicker' class) and select mode application  on page loading. I removed the `selectpicker` class to eliminate auto-init.

AdSchellevis:
@kulikov-a there are more cases where auto-init is problematic, removing the class isn't an issue. If this works as it should, I'll test on my end and merge. At a first glance this looks good.

kulikov-a:
Thanks!

PR #7057, opened 2023-12-06
Firewall: Aliases - read before write in alias table update to spare unneeded disk writes

5kft:
Hi @AdSchellevis , this change relates to issue https://github.com/opnsense/core/issues/6596, and could perhaps be considered an additional optimization relating to the changes made in https://github.com/opnsense/core/commit/0e3d660e18c3b7e7fdefc6a9046ced6b60400500.

In the existing implementation, when the "Alias Resolve Interval" expires, `update_tables.py` will unconditionally rewrite affected alias entries, even if they haven't changed.  These modifications optimize this behavior so alias files will only be rewritten if the content is different than that on disk.  (This actually saves a significant number of extraneous writes on my system.)

I hope that this is helpful!

### DIFF
--- a/src/opnsense/scripts/filter/update_tables.py
+++ b/src/opnsense/scripts/filter/update_tables.py
@@ -94,8 +94,20 @@ if __name__ == '__main__':
         # only try to replace the contents of this alias if we're responsible for it (know how to parse)
         if alias.get_parser():
             # when the alias or any of it's dependencies has changed, generate new
-            if alias_changed_or_expired or not os.path.isfile('/var/db/aliastables/%s.txt' % alias_name):
-                open('/var/db/aliastables/%s.txt' % alias_name, 'w').write('\n'.join(sorted(alias_content)))
+            alias_filename = '/var/db/aliastables/%s.txt' % alias_name
+            alias_file_exists = os.path.isfile(alias_filename)
+            if alias_changed_or_expired or not alias_file_exists:
+                new_alias_content = '\n'.join(sorted(alias_content))
+                if alias_file_exists:
+                    try:
+                        current_alias_content = open(alias_filename, 'r').read()
+                    except UnicodeDecodeError:
+                        current_alias_content = ""
+                else:
+                     current_alias_content = ""
+
+                if current_alias_content != new_alias_content:
+                    open(alias_filename, 'w').write(new_alias_content)
 
             # list current alias content when not trying to update a targetted list
             alias_pf_content = list(PF.list_table(alias_name)) if to_update is None else alias_content


src/opnsense/scripts/filter/update_tables.py
```diff
@@ -94,8 +94,20 @@
         # only try to replace the contents of this alias if we're responsible for it (know how to parse)
         if alias.get_parser():
             # when the alias or any of it's dependencies has changed, generate new
-            if alias_changed_or_expired or not os.path.isfile('/var/db/aliastables/%s.txt' % alias_name):
```
AdSchellevis, line 97:
```
if alias.changed() or not os.path.isfile('/var/db/aliastables/%s.txt' % alias_name):
```

should likely do the trick already.  `changed()` returns `True` if the calculated hash differs and is "sticky" throughout the session. Hashes only exist when there is a parser, which is the same condition as here.  For reference: 

https://github.com/opnsense/core/blob/0e3d660e18c3b7e7fdefc6a9046ced6b60400500/src/opnsense/scripts/filter/lib/alias/__init__.py#L115-L125

https://github.com/opnsense/core/blob/0e3d660e18c3b7e7fdefc6a9046ced6b60400500/src/opnsense/scripts/filter/lib/alias/__init__.py#L195-L205

Writing again on expire when unchanged should indeed not need an additional write, but adding too much glue in already complex code is also something we should prevent. Can you change accordingly and test on your end?

5kft:
Thanks! I just made the change and tested, and there is an unfortunate side effect:  When updating an alias, the alias text file that normally would get updated no longer gets written.  To explain:

With my change, changing and saving+applying the alias "Test_Servers" will result in all alias files being written correctly:

    1 -rw-r-----   1 root  wheel       32 Dec  7 05:55 Test_Servers.md5.txt
    1 -rw-r-----   1 root  wheel       26 Dec  7 05:55 Test_Servers.self.txt
    1 -rw-r-----   1 root  wheel       26 Dec  7 05:55 Test_Servers.txt

With your suggestion above - i.e., changing just `alias_changed_or_expired` to `alias.changed()` in the original code, and performing the same test will result in `Test_Servers.txt` no longer getting written:

    1 -rw-r-----   1 root  wheel       32 Dec  7 06:04 Test_Servers.md5.txt
    1 -rw-r-----   1 root  wheel       39 Dec  7 06:04 Test_Servers.self.txt
    1 -rw-r-----   1 root  wheel       26 Dec  7 05:55 Test_Servers.txt

The `Test_Servers.self.txt` file isn't updated on expiry timeout, so that optimization works.  But isn't it a problem that the original alias file no longer gets written?

AdSchellevis:
something does look off indeed, although our change should not have affected the self.txt file. what does this test scenario look like?

5kft:
> something does look off indeed, although our change should not have affected the self.txt file. what does this test scenario look like?

It's pretty simple - I have an existing alias that is a simple list of a couple of host IP addresses.  I edit the list and add another IP address, save the alias, and apply on the parent alias list page.  The new IP address is added correctly to the "self.txt" file, but the non-"self.txt" file is not updated.

AdSchellevis:
I need to think about this, maybe we end up with a read+compare after all, but then we should probably move the read file logic inside the class to avoid duplication.

AdSchellevis:
@5kft can you test https://github.com/opnsense/core/commit/935c64295b3af04a17e96621f09909febb25b749 ? it does roughly what your original proposal did, but avoids some code duplication.

5kft:
Hi @AdSchellevis, these changes appear to work great.  New aliases are written correctly, modified aliases are updated correctly, and expired and unchanged aliases are not re-written upon expiry timeout.  Thanks!

AdSchellevis:
@5kft thanks for reporting back, I changed the last bit in https://github.com/opnsense/core/commit/bb4cddda35906a68e02eaf7b8b0af33bbef7038c

PR #7058, opened 2023-12-06
system: read before write of /etc/hosts and /etc/resolv.conf to spare unneeded disk writes

5kft:
This change relates to issue https://github.com/opnsense/core/issues/6596, and is a simple optimization that eliminates numerous unnecessary rewrites of the `/etc/resolv.conf` and `/etc/hosts` files.

My system receives lots of "new WAN IP" events (via both gateway router DHCP renewals, IPv6 RAs, etc.), when nothing actually changes (e.g., "No IP change detected").  OPNsense processing for these events end up calling `system_resolver_configure()`, which regenerates these files each time.

This commit simply compares the hashes of the contents of the existing `/etc/resolv.conf` and `/etc/hosts` files to hashes of the new file content, and will only write the files to disk if they don't exist, or if the new content is different.

### DIFF
--- a/src/etc/inc/system.inc
+++ b/src/etc/inc/system.inc
@@ -222,11 +222,20 @@ function system_resolvconf_generate($verbose = false)
         $resolvconf .= "search {$result}\n";
     }
 
-    $tempfile = tempnam('/tmp', 'resolv.conf');
-    file_put_contents($tempfile, $resolvconf);
-    chmod($tempfile, 0644);
-
-    rename($tempfile, '/etc/resolv.conf');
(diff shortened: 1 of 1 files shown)


fichtner:
I admire the effort but we only have a few cases where changing content signifies something more than just avoiding a write and if we want to reuse that pattern in potentially hundreds of places we need a helper function although on the practical side I'm not sure it's worth the increased power consumption if calculating hashes all the time vs. writing all the time unconditionally.

5kft:
Understood.  Thanks for the consideration!

PR #7078, opened 2023-12-16
Make the GUI interfaces selection actually honored by ntpd

doktornotor:
Without the "interface ignore wildcard" directive, the service still listens on wildcard.


fichtner:
Funky... thanks!

doktornotor:
Well yes, this annoying behaviour has been a source of complaints probably since forever - e.g. [Ubuntu bug from 2011](https://bugs.launchpad.net/ubuntu/+source/ntp/+bug/858493). Don't know why upstream insists on this. 🤦 🤷‍♂️ 

There's some write-up and discussion about this [at their support website](https://support.ntp.org/Dev/NtpdAndNetworkSockets)

Thanks.

PR #7082, opened 2023-12-17
f/u to #6838 : implement ocsp check in tls_verify.php

zerwes:
f/u to #6838

This is a draft for implementing a ocsp check for openvpn in the `tls_verify.php` script.
The config and form part ist not yet covered here. For now I just used a placeholder `ocsp_uri` in the openvpn server config.

... and here the tl;dr part ...

First of all: sorry for the delay on this topic.

As discussed in #6838 , a ocsp support would be preferred instead of (I have to admit it) some obscure twists and turns around CRLs (despite the fact that I still consider a CRL support a valid option)

Well, I started with a modified script from the openvpn repo: https://gist.github.com/zerwes/5987cc037b0a1935a2b7a5cb32027463

As the `tls-verify` cfg setting for openvpn is already in use; I implemented a first draft for the ocsp check in `tls_verify.php` (called from `ovpn_event.py`).

The current draft of the implementation is overflowing with debug messages and is expecting the not unimplemented openvpn server config setting `ocsp_uri`.

The question is:
should this be set via the GUI (if this is the desired way, juts let me know if you like to have this implemented in the pr here).
Or would it be a viable option to check if the CA cert has a OCSP URI set in the `authorityInfoAccess` field and use it, as long as the config is not explicitly denying the OCSP check? (Well, then the cfg and ui have to be extended for this too.)
This would mean a increase of the security by default and could be even carried out in several stages:
 1. as long as the server config is **not** explicitly denying to check if a client cert is revoked,
 2. use OCSP if the `authorityInfoAccess` field is present (here first the ca should be checked, and if the settings is not available, the server cert)
    `openssl x509 -in /var/etc/openvpn/server${SERVER_ID}.[ca|cert] -noout -ocsp_uri`
 3. if not, use the CRL if configured in `crlDistributionPoints` (again, first testing the ca, and if not available, the server cert)
    `openssl x509 -in /var/etc/openvpn/server${SERVER_ID}.[ca|cert] -noout-ext crlDistributionPoints`
    (here the procedure to fetch and updated the CRL must be discussed and implemented)

Additionally there might be some more settings that should be at least considered if they should be configurable:
 * `nonce` usage (https://www.openssl.org/docs/man3.0/man3/OCSP_check_nonce.html)
 * currently, if the ocsp server is not available, the check fails and the connection is rejected; this is the desired behavior from my point of view, since the availability and correct functionality of the ocsp server should a task for the monitor and not for the openvpn service, but opinions might differ.

Disclaimer: I have tested the implementation with a simple test CA setup without intermediate signing cert!

### DIFF
--- a/src/opnsense/scripts/openvpn/tls_verify.php
+++ b/src/opnsense/scripts/openvpn/tls_verify.php
@@ -30,7 +30,7 @@
 require_once("config.inc");
 
 /**
- * verify certificate depth
+ * verify certificate depth and optional against a ocsp server
  * @param string $serverid server identifier
  * @return string|bool an error string or true when properly authenticated
  */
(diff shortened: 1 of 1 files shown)


AdSchellevis:
Hi Klaus,

Can you share the steps to generate the test ca and certificates with me? I don't have anything available to test at the moment and haven't looked into setting up anything myself yet.

Best regards,

Ad

zerwes:
Hello @AdSchellevis 
Yes, of course.
I can try today to create something generic and put it in a repo. Will just take some short while, as I have currently some other stuff in my pipeline.
I'll keep you updated.
Greetings
Klaus

zerwes:
Hello again Ad.
And now the steps to test the ocsp stuff ... (sorry, was in a hurry before)

 1. checkout the repo mentioned above
    * adjust the cfg
    * create the CA
    * create the ocsp signing cert
    * create a server cert for opnsense / openvpn
    * create 2 vpn certs
 2. opnsense:
    * Trust:
      * Authorities: create ca w/ the ca cert `CA/ca.crt` (key is not required)
      * Certificates: import cert and key of the server cert generated (`CA/certs/nosense.zero-sys.net.key` `CA/certs/nosense.zero-sys.net.pem`)
    * VPN -> openvpn:
      * create a server instance
        * Remote Access (SSL/TLS)
        * TLS Authentication enabled
        * Peer Certificate Authority set to the created authority
        * Certificate Depth One
        * Verb 3
      * do a *Client Export* for the server instance
      * extract the export and copy the ca cert and the 2 vpn certs and keys into the extracted directory
      * copy the `*.ovpn` to `vpn1.ovpn` and `vpn2.ovpn`
      * edit the `vpn*.ovpn` files and set (instead of the `pkcs12` entry if prresent or if not, just append the neww entries) (see diff below)
        the second file should point to the vpn2 cert/key pair
      * run the openvpn client against both cfg files and test if they work
        `openvpn --config vpn[12].ovpn`
      * if both work as expected ssh into the opnsense box and
      * apply the patch from the pr
        `opnsense-patch -s https://github.com -a zerwes -r opnsense-core e09e8d9d3a18fa682ce94e810b086a1e4d50fbcc`
      * check again both cfg work
      * edit `/usr/local/opnsense/scripts/openvpn/tls_verify.php` and set the `$ocsp_uri` (line 52)
    * in the CA start the ocsp server (the `ocsp_uri` should point there)
    * check both vpn cfg work
    * in the CA revoke vpn2 cert
    * check only vpn1 cfg should work now

Hope I did not miss a step
If I can give some more assistance, juts let me know

Greetings
Klaus

```diff
-pkcs12 test_ocsp_zerosense_zero_sys_net.p12
+ca ca.crt
+cert vpncert1.zero-sys.net.pem
+key vpncert1.zero-sys.net.key
```

AdSchellevis:
Hi Klaus,

No problem, I need some time to try this myself (quite busy at the moment), but this is certainly helpful in progressing this matter. 

Best regards,

Ad

zerwes:
> No problem, I need some time to try this myself (quite busy at the moment), but this is certainly helpful in progressing this matter.

Hello Ad,
ditto no problem, it took me ages to find the time ...
In fact it is not that complicated as my description sounds.
Just let me know if I can contribute something.

Greetings
Klaus

AdSchellevis:
Hi Klaus,

I think you have left quite some pointers to help this progress, I'm trying things out step by step on my end, if I need additional help, I'll let you know. Thanks for spending the time on this.

Best regards,

Ad

AdSchellevis:
Hi Klaus,

I think I have all the components I need for OCSP and integrated some missing parts in our gui as well for easier testing (https://github.com/opnsense/core/commit/5cf738f6e7a3cd481c8c009d9e3108f96f318228 + https://github.com/opnsense/core/commit/52fe119e0c82cd4b9613dd80263d6290c3d62358 )

Was able to revoke a cert on my end as well, only thing I couldn't find, how do I "unrevoke" one while testing? 

Best regards,

Ad

zerwes:
Hello Ad
`unrevoke` is not a valid action in terms of a certificate, as the serial is registered in the rev db.
you must generate a new cert (new serial), but you can use the same CN etc.
technically you can even use the old key, but usually a revoke is done due to a compromised key, so you should start all over (afair my rev script deletes the key).

PS: if I find the time, I'll try the new UI settings.

Best regards,
Klaus

zerwes:
As it is only for test purposes, the following idea may help: put the CA in a local git repo and perform the revocation of the cert in a branch; then you can switch between the branches with the valid / revoked cert... (nothing one should do on a real CA, but for testing this should work)
Greetings
Klaus

AdSchellevis:
@zerwes can you try the latest code in master? The new `ocsp_validate()` should ease implementation for serves that would like to use OCSP on our end (https://github.com/opnsense/core/commit/ff75146f73eee563afd516fa60b6bfca29991584)

PR #7090, opened 2023-12-19
dhcpd6: Fix help text for prefix delegation range syntax

### LINKED ISSUE #7046: Wrong "help" in DHCPv6 for Prefix Delegation Range
**Important notices**

**Describe the bug**

The help in DHCPv6 "Prefix Delegation Range" says:

> When using a tracked interface then please only enter the range itself, i.e. ::xxxx:0:0:0:0. For example, for a /56 delegation from ::100;0:0:0:0 to ::f00:0:0:0:0.

But this is wrong.

My LAN interface is set to "track interface" and I chose to delegate /60, so I set the range as ::10:0:0:0:0 to ::f0:0:0:0:0, because it says to only enter the range. But this was not working. OPNsense RA is Unmanaged and my subrouter was only requesting a prefix and not an address. It got the prefix (`f` was the first assigned) but its LAN addresses were `::f0:xxxx:xxxx:xxxx:xxxx/128` instead of `2001:(...)f0:xxxx:xxxx:xxxx:xxxx/128`. I had to set the range with the public routing prefix too for it to work.

This also caused the route to be created with `0:0:0:f0/60` instead of `2001:(...)f0:xxxx:xxxx:xxxx:xxxx/60`.

So either the help is wrong, or the prefix delegation doesn't work as expected.

**To Reproduce**

1. Add a subrouter (another OPNsense if you want) behind the main OPNsense.
2. Enable DHCPv6 and set the Prefix Delegation Size to 60 and delegate from ::10:0:0:0:0 to ::f0:0:0:0:0
3. Set the subrouter WAN interface to DHCPv6 and its LAN to "track interface"
4. Check the LAN address

**Expected behavior**

The prefix delegation should delegate the public routing prefix as well as the prefix itself.

Add the public routing prefix in the range.

**Screenshots**

Untitled

**Relevant log files**

If applicable, information from log files supporting your claim.

**Additional context**

**Environment**

OPNsense 23.7.9-amd64 (virtualized)
Ryzen 5 3600
Linux bridge

maurice-w:
Closes #7046.

### DIFF
--- a/src/www/services_dhcpv6.php
+++ b/src/www/services_dhcpv6.php
@@ -534,8 +534,8 @@ include("head.inc");
                            <?= gettext("Ensure that any prefix delegation range does not overlap the LAN prefix range."); ?>
 <?php if (isset($config['interfaces'][$if]['dhcpd6track6allowoverride'])): ?>
                           <br/><br/>
-                          <?= gettext('When using a tracked interface then please only enter the range itself, i.e. ::xxxx:0:0:0:0. ' .
-                            'For example, for a /56 delegation from ::100:0:0:0:0 to ::f00:0:0:0:0. ' .
+                          <?= gettext('When using a tracked interface then please only enter the range itself, i.e. ::xxxx. ' .
+                            'For example, for /56 delegations from ::100 to ::f00 or for /60 delegations from ::10 to ::f0. ' .
                             'Also make sure that the desired prefix delegation size is not longer than the available size shown above.') ?>
 <?php endif ?>
                           <br/><br/>


fichtner:
I’m unsure this is the right fix. Did anyone check that the compatibility glue with the old layout breaks this? It even logs a message when doing so. As I remember the help text is derived from isc-dhcp manual itself.

maurice-w:
I doubt you'll find anything about this in the isc-dhcp manual. It can't perform the prefix merge itself, I think this has always been done by OPNsense.

fichtner:
See https://github.com/opnsense/core/issues/7046#issuecomment-1863307667 -- you are using the legacy way of specifying the prefix and it should end up with a warning message in the log...

maurice-w:
Thanks for letting me know, wasn't aware this syntax is considered legacy. So I'll close this PR and switch my configs to the new way - once it works. ;)

PR #7091, opened 2023-12-19
Sunnyvalley.repo repo dynamic update process

ureyni:
copy SunnyValley.conf.shadow to SunnyValley.conf file.


fichtner:
this goes to your plugin, but with the NODE_UUID additions that you want in the conf file...

Issue #6973, opened 2023-11-01
ipsec: rearrange "enable" checkbox to make it more obvious and fail-safe


fichtner:
**Important notices**

Our forum is located at https://forum.opnsense.org , please consider joining discussions there in stead of using GitHub for these matters.

Before you ask a new question, we ask you kindly to acknowledge the following:

- [x] I have read the contributing guide lines at https://github.com/opnsense/core/blob/master/CONTRIBUTING.md
- [x] I am convinced that my issue is new after having checked both open and closed issues at https://github.com/opnsense/core/issues?q=is%3Aissue

It was mentioned to me that this is easy to overlook just last week and today this was also the case on the forum:

https://forum.opnsense.org/index.php?topic=36751.0

This also happened in the past and maybe we can do something about it.

AdSchellevis:
it's more or less the same place for both options, no idea what a better place would be, its always easily overlooked (unless you hide the grids when disabled, but that has other downsides)

fichtner:
Yep, just something to discuss for now.

Monviech:
Can't IPsec be automatically enabled the first time swanctl.conf is populated with a connection?

fichtner:
It could, but a global on/off has been proven helpful both in operation and during development

PR #7095, opened 2023-12-21
Move the Enable checkbox for legacy tunnels to a visible place

doktornotor:
Prevents time wasted with debugging why IPsec does not work at all.


AdSchellevis:
https://github.com/opnsense/core/issues/6973

fichtner:
I'd rather leave this as is now as it aligns with the connections page https://github.com/opnsense/core/commit/7ed57ae22ff a bit annoying the tunnel page doesn't have an apply button down there. Thoughts?

AdSchellevis:
closing this now as the legacy components are reaching their end of life on our end.

PR #7117, opened 2024-01-05
MVC: remove Phalcon migration glue

swhite2:
The v4->v5 compatibility layer is not needed anymore, so remove any cruft that made it possible. Some notable changes:

- The autoloader workaround isn't necessary anymore, this is provided by Phalcon.
- the Validation() constructor does not need a phalcon version check anymore
- The `Validation` class as a wrapper for Phalcon's `Validation` class remains, as this might ease dropping the dependency as mentioned in https://github.com/opnsense/core/commit/bc881b9c11500ee57a7448e94794c5e9827d7afe

All other changes are simply namespace changes that refer to the Phalcon v5 implementation.


fichtner (approved):

AdSchellevis:
@swhite2 looks good, thanks!

PR #7146, opened 2024-01-19
WIP: OpenVPN CSC: Add iroute

mimugmail:
Hi,

a draft to add an iroute field to CSC, I'm unsure about the file generation as it's quite dynamic now and I can't check the result. 
Any hint how to debug this better?

Thx
Michael

### DIFF
--- a/src/etc/inc/plugins.inc.d/openvpn.inc
+++ b/src/etc/inc/plugins.inc.d/openvpn.inc
@@ -1014,6 +1014,14 @@ function openvpn_csc_conf_write($settings, $server, $target_filename = null)
     ) {
         $conf .= openvpn_gen_routes($settings['remote_network'], 'ipv4', false, true);
     }
+
+    if (
+        !empty($settings['iroute_network']) &&
+        openvpn_validate_cidr($settings['iroute_network'], '', true, 'ipv4') === false
+    ) {
+        $conf .= openvpn_gen_routes($settings['iroute_network'], 'ipv4', false, true);
+    }
+
     if (
         !empty($settings['remote_networkv6']) &&
         openvpn_validate_cidr($settings['remote_networkv6'], '', true, 'ipv6') === false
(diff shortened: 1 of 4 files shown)


AdSchellevis:
@mimugmail at a first glance I don’t expect this is needed, but I need to double check when I’m around a pc

mimugmail:
@AdSchellevis thx for having a look. It's for dynamic routing, when you have 2 OPNsense OpenVPN servers and both have a server instance and you put the remote network field, the process already creates a route in routing table, no matter if the client is connected or not. My "guess" would be to put the iroute in CSC which will create the kernel route when client connect and the on OPNsense instance can send a signal to the other one that the network of client X is available via X.

AdSchellevis:
But what happens if you leave the routes from the instance and only define local and remote in the cso?

mimugmail:
It doesnt place a route in the kernel, which is essential for dynamic routing. I can demonstrate in a lab if you are interested

AdSchellevis:
I’m pretty sure iroutes aren’t what your looking for as they influence the OpenVPN internal routes. If client connect does not install the remote route, it might be worth checking the logs. In most cases you don’t want the client to influence the server routes, so it might ignore them if not installed already.

if that’s the case, it might be possible to use the connect/disconnnect events, but that is likely more complicated.

ping me online next week when I’m on irc so you can show me your setup

mimugmail:
When I read this, CSO iroute is exactly what I need :)

https://forums.openvpn.net/viewtopic.php?t=34572

mimugmail:
Hm, ok, maybe a routing protocol is needed. 

https://forums.openvpn.net/viewtopic.php?t=30836

With IPsec this works just fine, but no option for a backup IP peer :)

AdSchellevis:
I’m pretty sure the iroute is already there, the forum thread seems to confirm what I thought…. If you want to add/delete kernel routes on (dis)connect, you need to hook a script.

mimugmail:
Lets close this one, I'll do this with dynamic routing.

Issue #6748, opened 2023-08-15
Monit: E-mail server password with special chars ! and " won't work 


TheNightmanX:
**Important notices**

Before you add a new report, we ask you kindly to acknowledge the following:

- [x] I have read the contributing guide lines at https://github.com/opnsense/core/blob/master/CONTRIBUTING.md
- [x] I am convinced that my issue is new after having checked both open and closed issues at https://github.com/opnsense/core/issues?q=is%3Aissue

**Describe the bug**
Monit alert e-mail password with special chars ! (exclamation mark) and " (double quotes) won't work 

**To Reproduce**
1. Go to opnsense Web: Services - Monit - Settings
2. Click on page 'General Settings'
3. enter the data and an email server password containing a special characters ! (exclamation mark) and " (double quotes)
6. save and apply
7. service does not start
8. see error in monit log: Error monit  /usr/local/etc/monitrc:12: syntax error '!'

**Expected behavior**
Service should accept the email password with special characters ! and " to succesfully start the service

**Describe alternatives you considered**
none possible, I am not able to change the password of the email service account

**Screenshots**

**Relevant log files**

If applicable, information from log files supporting your claim.

**Additional context**

**Environment**

Software version used and hardware type if relevant, e.g.:

OPNsense 23.7.1_3-amd64
FreeBSD 13.2-RELEASE-p2
OpenSSL 1.1.1v 1 Aug 2023
Intel(R) Celeron(R) J4125 CPU @ 2.00GHz (4 cores, 4 threads)

fbrendel:
Hi @TheNightmanX,
monit already accepts special characters except double quotes.

This is not an OPNsense limitation but from Monit.
See [THE MONIT CONTROL FILE](https://www.mmonit.com/monit/documentation/monit.html#THE-MONIT-CONTROL-FILE)  

> Strings can be either quoted or unquoted. A quoted string is bounded by **double quotes** ...

Escaping the double quote is also not possible.

MAkcanca:
Here is the related part.
https://github.com/opnsense/core/blob/4d383db1d2a5f91eb3b5b8d521d58d731a269f92/src/opnsense/service/templates/OPNsense/Monit/monitrc#L57


If user inputs a password that contains `"` rc file will fail with syntax error '}'
We can't blame a 15 year old monitoring tool since they probably won't accept new MRs. https://github.com/arnaudsj/monit

But since OPNsense generates the rc file using jinja, we can add a small line like 
[...]

PR #7156, opened 2024-01-22
Monit: support quotes in passwords

fbrendel:
Passwords containing single quotes must be bounded by double quotes and vice versa.
So passwords with double quotes needs to be bounded by single quotes.
See #6748

### DIFF
--- a/src/opnsense/service/templates/OPNsense/Monit/monitrc
+++ b/src/opnsense/service/templates/OPNsense/Monit/monitrc
@@ -52,9 +52,13 @@ set eventqueue basedir {{ OPNsense.monit.general.eventqueuePath }} {{ slots }}
 {%    set port = "port " ~ OPNsense.monit.general.port if OPNsense.monit.general.port|default('') != '' %}
 {%    set username = '' %}
 {%    set password = '' %}
+{%    set quotechar = '"' %}
 {%    if helpers.exists('OPNsense.monit.general.username') and helpers.exists('OPNsense.monit.general.password') %}
 {%      set username = "username " ~ OPNsense.monit.general.username %}
-{%      set password = "password " ~ '"' ~ OPNsense.monit.general.password ~ '"' %}
+{%      if '"' in OPNsense.monit.general.password %}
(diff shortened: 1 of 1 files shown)


fichtner:
And if it has both? ;)

MAkcanca:
Highly unlikely but could happen. However I think it's better than the case before, where we don't handle double quotes. I had rough couple of hours trying to figure out why my monit setup don't send anything out then I discovered my password contains some quotes and it broke monit config. However if I was informed or got prevented from saving that password in the web page setup form, none of this would've happened.
Can we modify the web form so it prevents certain stuff?

fichtner:
I want to be very honest: security "researchers" don't give a damn about probability and this is clearly problematic from a technical point of view. Touch this once and fix all the issues is the best approach or else you will end up spending more time and more time again and processing a CVE eventually...

That being said single and double quotes have different escaping rules. In the shell you can use different quotes to write problematic quotes. I don't know how this works for monit but I also don't want to investigate.

`"'` can be escaped as `'"'"'"` in the shell, but bison probably uses escapes here so it needs to be checked which escape sequences are less for which quote and pick that as a standard and only deal with the remaining offending quote char separately (and escape the escape character by default, so `\` becomes `\\`.


Cheers,
Franco

MAkcanca:
@fichtner But monit doesn't use shell to parse? Maybe I'm missing the point.
See
https://github.com/arnaudsj/monit/blob/aa491f6ca2d2dd79abebf1e97b6f3a14fd4453e8/util.c#L396
https://github.com/arnaudsj/monit/blob/aa491f6ca2d2dd79abebf1e97b6f3a14fd4453e8/l.l#L408

But if you can make it work, that would be awesome. Good luck!

fichtner:
I'm only explaining my point because the approach so far is not cutting it.

fbrendel:
Ok, then, to be on the safe site, we have to encode all charachters.
@fichtner is it possible to do that via a new field type e.g. `HexField`?

AdSchellevis:
no progress at the moment

PR #7179, opened 2024-01-30
Add Uptime API Endpoint

Neozlag:
The purpose of this pull request is to add the uptime of OPNsense accessible with the API Endpoint /api/core/system/uptime

### DIFF
--- a/src/opnsense/mvc/app/controllers/OPNsense/Core/Api/SystemController.php
+++ b/src/opnsense/mvc/app/controllers/OPNsense/Core/Api/SystemController.php
@@ -185,4 +185,13 @@ class SystemController extends ApiControllerBase
 
         return ["status" => "failed"];
     }
+
+    public function uptimeAction()
+    {
+        if ($this->request->isGet()) {
+            $backend = new Backend();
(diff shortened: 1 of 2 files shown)


AdSchellevis:
We try to prevent endpoints not being used by the gui, maybe when we're working on refactoring the dashboard, we can offer an endpoint that also offers (parsed) uptime content. 

You can always deploy an endpoint in your local setup to reach the same effect.

Neozlag:
Perhaps we can make an exception until the dashboard refactoring ?
I've got a hundred or so boxes for which I'd like to have the uptime, and making the changes locally could take a long time ^^
Regards,

AdSchellevis:
I don't expect as removing it later will cost us even more time in discussions. Uptime only as a string is also not a very useful endpoint for other machines to process.

swhite2:
Since the dashboard refactor is in progress on master, here is the commit: https://github.com/opnsense/core/commit/d267e33de4fb5629c476341d90050fcd5698075d. The output is a bit more detailed than the `uptime` command (seconds included), and adjusted for locale. Endpoint is `/api/core/system/systemTime`

PR #7186, opened 2024-01-31
alias_util.volt: make select width more consistent

kulikov-a:
Hi!
may close https://github.com/opnsense/plugins/issues/3782 i hope.
'.selectpicker' auto-init may result a too narrow alias choice select. and it not handles well by 'refresh' or 'render' methods.
reproduces quite stable in the Edge browser.
PR inits bs.select after all options have been added so that bs.select knows the required width.
Thanks!


fichtner:
Merged, thanks!

kulikov-a:
thanks!)

Issue #6622, opened 2023-06-18
Possibility to change color scheme of graphs in "Reporting: Traffic"


vgdh:
**Important notices**

Before you add a new report, we ask you kindly to acknowledge the following:

- [v] I have read the contributing guide lines at https://github.com/opnsense/core/blob/master/CONTRIBUTING.md
- [v] I am convinced that my issue is new after having checked both open and closed issues at https://github.com/opnsense/core/issues?q=is%3Aissue

**Is your feature request related to a problem? Please describe.**

Possibility to change color scheme of graphs in "Reporting: Traffic"
Sometimes it choose very similar color and it become very inconvenient to use this tool.

**Describe the solution you like**

Possibility to change color scheme of graphs in "Reporting: Traffic"
It may be not permanent color scheme but when i need to diagnostic something i would change it to what i want and what is convenient for me.
So theoretically it will be not so hard to implement, it will only affect web UI. So no change to backend.

**Describe alternatives you considered**

__

**Additional context**

__

swhite2:
The color-scheme we use (`brewer.Paired12`) is in my opinion the one that differentiates datasets best. Which set of schemes would you have in mind? https://nagix.github.io/chartjs-plugin-colorschemes/

vgdh:
I am not a UI designer but something like this would definitely work for me (button to change color).
Because some colors look similar to each other, it's hard to tell which interface you're looking at.
![colorScheme1](https://github.com/opnsense/core/assets/15322782/5e5868f3-d731-48fe-b40e-11028bc8df16)

vgdh:
one more example of similar colors
![11](https://github.com/opnsense/core/assets/15322782/5c9993e2-f715-4376-b843-37a070863af4)

swhite2:
We have little control over the colors assigned to individual datasets, this is handled by the plugin as referred to in my first comment. Changing this logic would be too much effort for too little gain. 

We have a choice of dataset colors, which shouldn't be hard to make slightly more configurable, though we would need to pick from a select few, which is where your opinion could come in. If you are interested you can scroll down to the page I referred to earlier and check out the different available options.

vgdh:
It would be cool to have a choice from a variety of color schemes from **chartjs-plugin-colorschemes**

PR #7217, opened 2024-02-08
update traffic graph colors to be contrast and consistent in widget and traffic diagnostic

brotherla:
default `brewer.Paired12` color scheme contains pairs very similar colors, so every two interfaces in list look very similar to each other, see image, where dataset1 and dataset2 are blue, dataset3 and dataset4 are green

image

to increase contrast between interface graphs I suggest to use `tableau.Classic10` which contains 10 very distinctive colors:

image

This should partially fix: [#6622](https://github.com/opnsense/core/issues/6622)

Another problem is that traffic widget and traffic diagnostic graphs are using different color schemes and indexes for the same interface, which is inconsistent and confusing, it is better to apply the same rules in both sections


swhite2:
I agree this scheme has much better color differentiation.

Looks good, thanks!

PR #7234, opened 2024-02-12
Kea: allow setting domain for DHCP clients

miken32:
I know Kea is still in the very early stages, but this is a fairly important option to have for continuing evaluation.

I noticed while testing, the validation of `HostnameField` types will allow pretty much anything to be passed that matches `[0-9a-z.-]+` except really outrageous problems like 2 dots in a row. In particular, the `IpAllowed` flag [does nothing](https://github.com/opnsense/core/blob/master/src/opnsense/mvc/app/models/OPNsense/Base/Validators/HostValidator.php#L75) because `filter_var('10.1.2.3', FILTER_VALIDATE_DOMAIN)` does not return false.

### DIFF
--- a/src/opnsense/mvc/app/controllers/OPNsense/Kea/forms/dialogSubnet4.xml
+++ b/src/opnsense/mvc/app/controllers/OPNsense/Kea/forms/dialogSubnet4.xml
@@ -34,6 +34,12 @@
         <help>DNS servers to offer to the clients</help>
     </field>
     <field>
+        <id>subnet4.option_data.domain_name</id>
+        <label>Domain name</label>
+        <type>text</type>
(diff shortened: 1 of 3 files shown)


AdSchellevis:
@miken32 reimplemented in https://github.com/opnsense/core/commit/8278fbe944fede0ad6f0934d997ed95b129a1bf5 , since the firewall always has a static domain name configured, we don't have to persist the value in this case.

AdSchellevis:
by the way, I still plan to take a look at the `HostnameField` type, the allow ip address feature is correctly a bit suboptimal as  `FILTER_VALIDATE_DOMAIN` seems to think it's a valid domain as well.

PR #7254, opened 2024-02-18
Add init7 mirror

thgoebel:
This PR adds the mirror that init7, a Swiss ISP, is running. See https://mirror.init7.net/.
I'm an init7 customer and not affiliated with them.


AdSchellevis:
We stopped adding predefined mirrors to the list due to maintenance concerns, the ones that report to us are being added to https://opnsense.org/download/ (and removed when someone reports one being non functional).

thgoebel:
Ok, thanks for the info.

PR #7271, opened 2024-02-22
IDS: Add checkbox for exception-policy

mimugmail:
Hi,

Suricata 7 added exeception policies:
https://docs.suricata.io/en/latest/configuration/exception-policies.html#exception-policies

When just upgrading to 24.1.2 there is no entry in suricata.yml which seems to activate this feature:
https://forum.suricata.io/t/my-traffic-gets-blocked-after-upgrading-to-suricata-7/3745

Two confirmations that when adding it to custom.yml it seems to work again in Suricata 6 behavior:
https://forum.opnsense.org/index.php?topic=38961.0
https://forum.opnsense.org/index.php?topic=38944.0

There might arise more problems e.g. when running openvpn on port 443 as it also does not speak https.

Nonetheless, in the official documentation it states that the default is ignore but it only seems to act like v6 when adding it explicitly, so maybe we can also set it to enabled by default, or just activate it in suricata.yml without a knob and let the decision up to you guys :)
https://docs.suricata.io/en/latest/configuration/exception-policies.html#exception-policies

### DIFF
--- a/src/opnsense/mvc/app/controllers/OPNsense/IDS/forms/generalSettings.xml
+++ b/src/opnsense/mvc/app/controllers/OPNsense/IDS/forms/generalSettings.xml
@@ -92,6 +92,13 @@
         <help>With this option, you can set the size of the packets on your network. It is possible that bigger packets have to be processed sometimes. The engine can still process these bigger packets, but processing it will lower the performance.</help>
     </field>
     <field>
+        <id>ids.general.ignorePolicy</id>
+        <label>Ignore Policy Errors</label>
+        <type>checkbox</type>
+        <advanced>true</advanced>
+        <help>With this option, when a protocol parser fails, it will just ignore the error which will mimic the default behavior in Suricata version 6 and below. Starting with Suricata 7, the default changed to drop affcted packets which might break e.g. services running on port 443 but not offering HTTPS protocol.</help>
+    </field>
+    <field>
         <id>ids.general.AlertLogrotate</id>
         <label>Rotate log</label>
         <type>dropdown</type>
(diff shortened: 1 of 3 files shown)


AdSchellevis:
defaults in their documentation seem to be rather convoluted this release, not sure what happened there, but for now I prefer setting these values to their advertised defaults to prevent issues and additional user interaction. If at one point we do want to offer access to these parameters, we should add all possible options instead of a checkbox anyway.

For reference, the sample yaml also seems to suggest ignore as being default (an not needed to set):
https://github.com/OISF/suricata/blob/69f0e85785e03ef344e56c16929a9e803c2894ab/suricata.yaml.in#L858-L862

I'll close this ticket with a commit to set it to the default, next it might also be a good idea to compare differences between our default yaml and theirs, but that's probably for a next minor release.

AdSchellevis:
@kulikov-a I'm not sure that's needed now as we also added https://github.com/opnsense/core/commit/ce87c2f68c40682c8dba5a047c504d4d4ab91b7e and https://github.com/opnsense/core/commit/0d676c7bf6494a77a078a8b6e0a2aec444428341 to ignore midstream exceptions. Are there know cases where this https://github.com/opnsense/core/commit/9b82093ed2d9d6a67b08b4ef3c1afbd7664b930f doesn't fix the current issues?

kulikov-a:
@AdSchellevis 

> Are there know cases where this https://github.com/opnsense/core/commit/9b82093ed2d9d6a67b08b4ef3c1afbd7664b930f doesn't fix the current issues?

no actually. just a precaution in case the memcap hit

AdSchellevis:
@kulikov-a Maybe we should go in phases then, https://github.com/opnsense/core/commit/4cf6870b03533098b52a8bc561d67f3eee0ea5b2 aligns the `suricata.yaml` to upstream, https://github.com/opnsense/core/commit/36b2b66a33c546b3d40e92ffbf508e46d747720e sets it to ignore as suggested as a precaution. Both should probably go in to a minor release, but not the next one....

kulikov-a:
@AdSchellevis sounds logical ) Thanks!

PR #7277, opened 2024-02-25
Fix some help text

lin-xianming:


src/opnsense/mvc/app/controllers/OPNsense/Firewall/forms/groupEdit.xml
```diff
@@ -3,7 +3,7 @@
         <id>group.ifname</id>
         <label>Name</label>
         <type>text</type>
-        <help>No numbers or spaces are allowed. Only characters in a-zA-Z</help>
+        <help>Only letters, digits and underscores are allowed. Cannot end in a digit.</help>
```
AdSchellevis, line 6: how about?

> Only texts containing letters, digits and underscores with a maximum length of 15 characters are allowed and the name may not end with a digit.

AdSchellevis:
@lin-xianming thanks

PR #7281, opened 2024-02-26
Unbound: root-hints and aggressive-nsec switches

kulikov-a:
Hi!
Just a couple of suggestions:
~~- Add a `root-hints` switch:
  ref. https://forum.opnsense.org/index.php?topic=37840.msg187795#msg187795
  Looks like there may be a strange thread race on unbound start, leading to an error loading root hints, high CPU load and unbound inoperability. Adding a switch should not cause harm imho since the internal hints list is up-to-date and unbound is used in forwarding mode in many configurations.~~
- Add an `aggressive-nsec` switch:
  ref. https://forum.opnsense.org/index.php?topic=38888.0
  ref. https://github.com/NLnetLabs/unbound/issues/824
  ref. https://unbound.docs.nlnetlabs.nl/en/latest/topics/privacy/aggressive-nsec.html
  Bad nsec records may lead to name resolution problems with  `aggressive-nsec` (`yes` by default).

Thanks!

### DIFF
--- a/src/opnsense/mvc/app/models/OPNsense/Unbound/Unbound.xml
+++ b/src/opnsense/mvc/app/models/OPNsense/Unbound/Unbound.xml
@@ -54,6 +54,7 @@
                 <Multiple>Y</Multiple>
             </outgoing_interface>
             <enable_wpad type="BooleanField"/>
+            <builtin_hints type="BooleanField"/>
         </general>
         <advanced>
             <hideidentity type="BooleanField"/>
--- a/src/opnsense/mvc/app/models/OPNsense/Unbound/Unbound.xml
+++ b/src/opnsense/mvc/app/models/OPNsense/Unbound/Unbound.xml
@@ -62,6 +62,10 @@
             <prefetch type="BooleanField"/>
             <prefetchkey type="BooleanField"/>
             <dnssecstripped type="BooleanField"/>
+            <aggressivensec type="BooleanField">
+                <Required>Y</Required>
+                <Default>1</Default>
+            </aggressivensec>
             <serveexpired type="BooleanField"/>
             <serveexpiredreplyttl type="NumericField"/>
             <serveexpiredttl type="NumericField"/>
--- a/src/opnsense/mvc/app/models/OPNsense/Unbound/Unbound.xml
+++ b/src/opnsense/mvc/app/models/OPNsense/Unbound/Unbound.xml
@@ -1,7 +1,7 @@
 <model>
     <mount>//OPNsense/unboundplus</mount>
     <description>Unbound configuration</description>
-    <version>1.0.8</version>
+    <version>1.0.9</version>
     <items>
         <general>
             <enabled type="BooleanField">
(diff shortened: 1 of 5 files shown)


fichtner:
To be honest I'd like to solve the root hints thing differently, perhaps with something along the lines of https://github.com/opnsense/core/commit/845fbd384 Using the copy should always work unless unbound has larger issues and adding an opt-out knob for less controllable results has no benefits.

kulikov-a:
@fichtner Hi!)

> unless unbound has larger issues

thats what im talking about). from what I saw during experiments (https://forum.opnsense.org/index.php?topic=37840.msg187795#msg187795) i don’t understand how to solve this on the OPNsense side: it doesn't look like it's a matter of using templates

fichtner:
(I'd like to avoid two workardounds for this now at once, maybe you can remove that bit from your PR to merge the other thing)

src/opnsense/mvc/app/models/OPNsense/Unbound/Unbound.xml
```diff
@@ -54,13 +54,18 @@
...
+            <builtin_hints type="BooleanField"/>
         </general>
         <advanced>
             <hideidentity type="BooleanField"/>
             <hideversion type="BooleanField"/>
             <prefetch type="BooleanField"/>
             <prefetchkey type="BooleanField"/>
             <dnssecstripped type="BooleanField"/>
+            <aggressivensec type="BooleanField">
+                <Required>Y</Required>
```
fichtner, line 66: this requires a model version bump

kulikov-a: oops. sure, I'll fix it now. so should I remove the hints part?

fichtner: for now yes but we should add it if the problem continues and upstream is unresponsive

kulikov-a: done!)

fichtner:
Merged, thanks!

kulikov-a:
@fichtner
Thanks!
Although I like my theory about thread safety and reading the file on each thread init, I'm afraid that my knowledge is not enough to create a proper request at upstream...
But I'll try to dig into the code some more )

fichtner:
efc9b09c769 then

fichtner:
technically it's probably still not perfectly safe, ouch... let me do that later

kulikov-a:
@fichtner no no. I talked about race and thread safety on unbound side itself.
(https://github.com/NLnetLabs/unbound/blob/352245160058e9419565f922d62ce01634280b9d/daemon/worker.c#L2278 on each thread with https://github.com/NLnetLabs/unbound/blob/master/iterator/iter_hints.c#L307)
so I don't expect that even locking the entire seup/start procedure will help to prevent this.
I talked about making a request in unbound repo )
sorry if I expressed unclearly

fichtner:
@kulikov-a ok but how can this screw up reading a file? unless the threads use a shared structure which is... err... bad?

kulikov-a:
@fichtner but the errors content: sudden attempts to parse a RR type in a comment line or a mismatch between the record type and the ip type (for the A-record suddenly there is an ipv6 address ?) leads to the assumption that .... yes :wink:

fichtner:
hmm, I am starting to not understand why people in the forum keep insisting on fixing this with us

PR #7283, opened 2024-02-26
firewall/rules: fix floating rule display

lin-xianming:
1. For a floating rule with a single interface group, the interface count and tooltip did not show when viewed from an interface rules page.

2. For a floating rule with an interface group and at least one other interface or interface group, the entire rule did not show when viewed from an interface rules page for which the interface is contained in one of the interface groups.


AdSchellevis:
@lin-xianming thanks, unfortunately this legacy code has more complications then intended, which given the constraints is very hard to fix now. Your changes look good, thanks for polishing this one out.

PR #7361, opened 2024-04-05
Allow setting UniFi Controller IP in Kea DHCP.

reitermarkus:
Tested using WireShark for my controller IP, 10.0.1.77 (`0x0a00014d`):

```wireshark
Option: (43) Vendor-Specific Information
  Length: 6
  Value: 01040a00014d
```

### DIFF
--- a/src/opnsense/mvc/app/controllers/OPNsense/Kea/forms/dialogSubnet4.xml
+++ b/src/opnsense/mvc/app/controllers/OPNsense/Kea/forms/dialogSubnet4.xml
@@ -86,4 +86,10 @@
         <type>text</type>
         <help>Boot filename to request</help>
     </field>
+    <field>
+        <id>subnet4.option_data.unifi_controller</id>
+        <label>UniFi Controller</label>
+        <type>text</type>
+        <help>IP address of your UniFi Controller</help>
+    </field>
 </form>
--- a/src/opnsense/mvc/app/models/OPNsense/Kea/KeaDhcpv4.xml
+++ b/src/opnsense/mvc/app/models/OPNsense/Kea/KeaDhcpv4.xml
@@ -83,6 +83,9 @@
                     <boot_file_name type="TextField">
                         <Mask>/^([^\n"])*$/u</Mask>
                     </boot_file_name>
+                    <unifi_controller type="TextField">
+                        <Mask>/^([^\n"])*$/u</Mask>
+                    </unifi_controller>
                 </option_data>
                 <pools type=".\KeaPoolsField">
                 </pools>
--- a/src/opnsense/mvc/app/models/OPNsense/Kea/KeaDhcpv4.xml
+++ b/src/opnsense/mvc/app/models/OPNsense/Kea/KeaDhcpv4.xml
@@ -83,8 +83,9 @@
                     <boot_file_name type="TextField">
                         <Mask>/^([^\n"])*$/u</Mask>
                     </boot_file_name>
-                    <unifi_controller type="TextField">
-                        <Mask>/^([^\n"])*$/u</Mask>
+                    <unifi_controller type="NetworkField">
+                        <NetMaskAllowed>N</NetMaskAllowed>
+                        <AddressFamily>ipv4</AddressFamily>
                     </unifi_controller>
                 </option_data>
                 <pools type=".\KeaPoolsField">
--- a/src/opnsense/service/templates/OPNsense/Kea/kea-dhcp4.conf
+++ b/src/opnsense/service/templates/OPNsense/Kea/kea-dhcp4.conf
@@ -49,6 +49,10 @@
 {% for od_attr in (subnet.option_data|list + option_data_defaults|list)|unique if subnet.option_data[od_attr]|length > 1 or od_attr in option_data_defaults %}
                     {
                         "name": "{{od_attr.replace('_','-')}}",
+{% if od_attr == 'unifi_controller' %}
+                        "space": "vendor-encapsulated-options-space",
+                        "code": 43,
+{% endif %}
                         "data": {{subnet.option_data[od_attr]|default(option_data_defaults[od_attr])|tojson}}
                     }{% if not loop.last %},{% endif +%}
 {% endfor %}
--- a/src/opnsense/service/templates/OPNsense/Kea/kea-dhcp4.conf
+++ b/src/opnsense/service/templates/OPNsense/Kea/kea-dhcp4.conf
@@ -47,6 +47,12 @@
                 "subnet": "{{subnet.subnet}}",
                 "option-data": [
 {% for od_attr in (subnet.option_data|list + option_data_defaults|list)|unique if subnet.option_data[od_attr]|length > 1 or od_attr in option_data_defaults %}
+{% if od_attr == 'unifi_controller' %}
+                    {
+                        "name": "vendor-encapsulated-options",
+                        "always-send": true
+                    },
+{% endif %}
                     {
                         "name": "{{od_attr.replace('_','-')}}",
 {% if od_attr == 'unifi_controller' %}
--- a/src/opnsense/service/templates/OPNsense/Kea/kea-dhcp4.conf
+++ b/src/opnsense/service/templates/OPNsense/Kea/kea-dhcp4.conf
@@ -33,6 +33,13 @@
                 "severity": "INFO"
             }
         ],
+        "option-def": [{
+          "name": "unifi-controller",
+          "code": 43,
+          "space": "vendor-encapsulated-options-space",
+          "type": "string",
+          "array": false
+        }],
         "subnet4": [
 {% for subnet in helpers.toList('OPNsense.Kea.dhcp4.subnets.subnet4') %}
             {
--- a/src/opnsense/service/templates/OPNsense/Kea/kea-dhcp4.conf
+++ b/src/opnsense/service/templates/OPNsense/Kea/kea-dhcp4.conf
@@ -35,9 +35,9 @@
         ],
         "option-def": [{
           "name": "unifi-controller",
-          "code": 43,
+          "code": 1,
           "space": "vendor-encapsulated-options-space",
-          "type": "string",
+          "type": "ipv4-address",
           "array": false
         }],
         "subnet4": [
@@ -57,9 +57,11 @@
                         "name": "{{od_attr.replace('_','-')}}",
 {% if od_attr == 'unifi_controller' %}
                         "space": "vendor-encapsulated-options-space",
-                        "code": 43,
-{% endif %}
+                        "code": 1,
+                        "data": "{{subnet.option_data[od_attr]}}"
+{% else %}
                         "data": {{subnet.option_data[od_attr]|default(option_data_defaults[od_attr])|tojson}}
+{% endif %}
                     }{% if not loop.last %},{% endif +%}
 {% endfor %}
                 ],


AdSchellevis:
We need something more generic and better maintainable for this, option 43 equals vendor-encapsulated-options (https://kea.readthedocs.io/en/latest/arm/dhcp4-srv.html) and can be used for multiple purposes, when 10 different vendors use the same option, we rather don't want to add all of them (and possibly cause incompatible variations of options). 

Leaving this open as feature request and room for further discussion.

mimugmail:
Its the same as my previous request, option 43 should be some kind of grid as there could be multiple 43 option entries

AdSchellevis:
@mimugmail but if I'm not mistaken, they are mutually exclusive (so only one applies for that specific configuration)

AdSchellevis:
> Based on 9.3 of https://www.isc.org/docs/2023kea_custom_options.pdf, there is a way to apply different options for different clients.

But I would like to keep it as simple as possible.... these type of constructs are almost impossible to generalize.

reitermarkus:
I have made this more generic now, by adding option definitions



and options



and allowing to specify the options in subnets

Bildschirmfoto 2024-04-07 um 00 39 26

AdSchellevis:
@reitermarkus although it feels a bit like using a cannon to kill a mosquito, maybe it's something we can't prevent for these custom options (I was hoping to keep this simple). I'll take a look as soon as I can.

AdSchellevis:
@reitermarkus what if we merge the definitions and the data? from a data perspective the split might be cleaner, but from a user input perspective the question is how often this relation is actually `1-N`. We can polish the input a bit to prevent non-relevant options being shown as well (e.g. the record type inputs). thoughts?

reitermarkus:
I have merged the definitions and data. Can you let me know how to hide e.g. the record types field based on the type field?

AdSchellevis:
I don't mind changing that for you,  but if you want to give it a shot, usually we add a style to the field and hook an event. For example in OpenVPN: 

https://github.com/opnsense/core/blob/a8e329b90529bac6ac3682d92fd6d7dec89b3cae/src/opnsense/mvc/app/controllers/OPNsense/OpenVPN/forms/dialogInstance.xml#L130

https://github.com/opnsense/core/blob/a8e329b90529bac6ac3682d92fd6d7dec89b3cae/src/opnsense/mvc/app/views/OPNsense/OpenVPN/instances.volt#L63-L76

But if we agree on the direction, I'm more than fine polishing the final bits when merging.

reitermarkus:
I have now also added a field to specify a client class test condition.

I'll leave the polishing up to you. Ideally `record_types` should work like `select_multiple` but with duplicates allowed instead of being a plain string.

AdSchellevis:
@reitermarkus I don't think we should add the client class now as the relation between both entries is not the same (usually a client class contains a set of options). If in the future we do want to add client-classes, we should be able to reuse the custom options.

reitermarkus:
I have tried to figure this out, but it doesn't seem that simple after all. I think the definitions have to be split again, as there are currently a bunch of problems:

- It is currently not possible to specify the same option with different data for different subnets.
- Pre-defined options cannot be re-defined, so these need to be specified without a definition.
- If client classes are added, sub-options need to be defined globally with a custom `space` in case they conflict with other client classes, while the parent-option needs to be defined in the client class, see the example here: https://kea.readthedocs.io/en/latest/arm/dhcp4-srv.html#dhcpv4-private-options

AdSchellevis:
Maybe a stupid question, but can't we use `uuids` to bind the definitions and payload together? Personally I have a hard time finding definitions for kea fields in different area's, but something like this doesn't seem to crash:

```
         "option-def": [
                {
                        "name": "11635360-0eaf-4c63-9d30-a192c8e59168",
                        "code": 1,
                        "type": "string",
                        "space": "11635360-0eaf-4c63-9d30-a192c8e59168"
                }
        ],
....
        "option-data": [
                    {
                        "name": "11635360-0eaf-4c63-9d30-a192c8e59168",
                        "space": "11635360-0eaf-4c63-9d30-a192c8e59168",
                        "data": "0.0.0.0"
                    }
        ]
```

While writing this, it still feels like we're sending a rocket to the moon, where in practice, most issues only circle around code 43 as this option isn't well defined. 

Personally I think offering fine grained controls with client classes adds way more complexity than 99% of our users are looking for, when someone needs almost programatic control of the dhcp process, a manual configuration might be a better alternative.

reitermarkus:
> but can't we use `uuids` to bind the definitions and payload together?

This will only work for sub-options. But then you need to also re-define the parent option with `"encapsulate": "11635360-0eaf-4c63-9d30-a192c8e59168"`, which can only be done in a client class, i.e. you cannot re-define a pre-defined option globally.

AdSchellevis:
hmm, that's annoying. Partly this is difficult in terms of templating (overcomplicated `jinja` templates are hard to maintain) and to some degree the user input is also problematic. We could opt for moving the `json` file generation to the model, which does make it easier to de-duplicate certain information into a single container.

If we change the config generation handling, the next question is, what should the user input look like and how do we prevent invalid entries being generated. Do you have thoughts about that? 

I do think it makes sense to first figure out what we want to build and generate code after that, there are some loose ends here, currently I do have some difficulties seeing all constraints that apply.

I don't mind spending some time on this to ease future additions, this was also a bit of the goal from my end to not start with ipv6 yet until ew have enough feedback for the ipv6 variant.

AdSchellevis:
@reitermarkus can you try the code currently in master? https://github.com/opnsense/core/commit/3f184a695fdbcbe4071f61da54227a437da94bcd adds the "vendor-encapsulated-options-space" with a minimal set of options. I haven't tested this myself, but the generated configuration on my end is valid and the ui offers validations to prevent duplicate sub-codes being inserted with different definitions.

AdSchellevis:
@mimugmail tried it out, didn't seem to work, but config looks as expected. I've moved the feature into its own branch for now implementing https://github.com/opnsense/core/commit/00acf13e081f5bbc8a55b8740699ea94c42fb2ad (kea_custom_opt_pr7361)

MeneerThijs:
The decision to move the `json` file generation to the model breaks the ability to use target overwrites to create custom kea-dhcp4.conf templates. This change limits the use of Kea for users who want to use features that are not (yet) available in the GUI. Please reconsider this change.

AdSchellevis:
not reconsidering (due to the long term maintainability of the json file), but if you wish to disable kea from the gui and use a custom configuration instead just open a ticket. It's not a priority for us, but I'm open for suggestions.

fichtner:
@MeneerThijs the basic approach here is to ask for inclusion of file-based includes if kea supports it...

MeneerThijs:
@fichtner Thank you for your suggestion. Kea indeed support file-based inclusions; however, my use case requires inclusions at multiple locations. I opted to create a custom template for `keactrl.conf` which references a custom `kea-dhcp4.conf` file in an alternate directory. (This is necessary because the filename must remain `kea-dhcp4.conf`, as it is used in the pid filename and the plugin definition expects a pid file named `/var/run/kea/kea-dhcp4.kea-dhcp4.pid`.) I realize this is a bit of a hack but at least it does not require patching any code..

fichtner:
@MeneerThijs @AdSchellevis I know it is redundant but if we chain the config generation in the right order this becomes possible with the kea template override subfolder:

    kea_setup="/usr/local/sbin/pluginctl -c kea_sync"

changed to

    kea_setup="/usr/local/sbin/pluginctl -c kea_sync; /usr/local/sbin/configctl template reload OPNsense/Kea"


Cheers,
Franco

AdSchellevis:
I don’t think we should as we’re not offering a suitable template it will just overwrite the one just created if I’m not mistaken. The standard overwrite always has an example (our template) which you can revert to

fichtner:
Maybe I don't understand the problem scope. If someone wants an override and the system delivers that should be working as intended. Neither `kea-dhcp4.conf` nor `kea-ctrl-agent.conf` are in the `+TARGETS` now but if someone wanted to use them the template system can't deliver this at the moment because it's rewritten later.

AdSchellevis:
Before writing my response. I do agree the problem scope is not clear enough. 

My concerns are that (if I’m not mistaken) we render the templates twice for everybody else and people trying to make a modification to our template can usually do that by cloning ours and ship a new version. 

If we don’t offer the base template, people will start opening tickets like these when future changes are incompatible. The previous template will break when we progress in time, which means building your own json with parts of the configuration will be a less viable option in the near future anyway. There are very good reasons why in this case we don’t use the template system (template complexity growing rapidly leading to hard to debug edge cases)

In my opinion the only viable extension here is if you don’t want to use our code at all and ship your own config, for which there are likely cleaner solutions. …. This also loops back to problem scope.

long story short, it all start with a ticket explaining goals for which we can agree or disagree if they match ours.

mimugmail:
@AdSchellevis I just tried "make mount" with a current system and I can set the values. Is this way to go to test?
Currently having a problem with a customer and also with hex in isc dhcp, would be a good fit to continue testing :)

AdSchellevis:
@mimugmail this probably needs a rebase, the ticket has been stale for months I'm afraid. The branch I opened is https://github.com/opnsense/core/tree/kea_custom_opt_pr7361, I can try to rebase to master, but that has a lot of other breaking changes.

mimugmail:
I did: 

```
opnsense-code core
cd /usr/core
git checkout kea_custom_opt_pr7361
make mount 
```

Sure, Dashboard is messy but for testing it's fine. No need to rebase right now, let me test first :)

mimugmail:
Shawn would say: Because reasons :)

mimugmail:
Hi, I spent some time on this turning some rounds and it seems the PR of @reitermarkus is correct. Kea requirest the use of client-classes for code 43, as sub classes are intended with this one. 

This is working when editing the file directly:

```
        "subnet4": [
            {
                "id": 1,
                "subnet": "10.1.1.0\/24",
                **"client-class": "ubnt",**
                "option-data": [
                    {
                        "data": "192.168.1.1,192.168.1.2"
                    },
                    {
                        "name": "domain-search",
                        "data": "domain"
                    },
                    {
                        "name": "routers",
                        "data": "10.1.1.1"
                    },
                    {
                        "name": "domain-name",
                        "data": "domain"
                    },
                    {
                        "name": "ntp-servers",
                        "data": "192.168.1.1"
                    },
                    {
                        "name": "time-servers",
                        "data": "192.168.1.1"
                    }
                ],
                "pools": [
                    {
                        "pool": "10.101.1.100-10.101.1.200"
                    }
                ],
                "reservations": [
                    {
                        "hw-address": "ff:ff:ff:ff:ff:ff",
                        "ip-address": "10.1.1.200",
                        "hostname": "WOL-Broadcast"
                    }
                ]
            }
        ],
        "option-def": [
            {
                **"name": "unifi-address",
                "code": 1,
                "type": "ipv4-address",
                "space": "ubnt",**
            }
        ],
    **"client-classes": [
        { "name": "ubnt",
          "test": "(option[vendor-class-identifier].text == 'ubnt')",
          "option-def": [
                { "name": "vendor-encapsulated-options",
                  "type": "empty",
                  "encapsulate": "ubnt",
                  "code": 43 }
          ],
          "option-data": [
              { "name": "unifi-address",
                "space": "ubnt",
                "data": "16.1.1.1" },
              { "name": "vendor-encapsulated-options" }
          ]
        }
    ]**
    }
}

```

mimugmail:
Btw. space name needs to be ubnt, vendor-encapsulated-options-space_43 will be ignored by the dhcp device.

AdSchellevis:
@mimugmail it has been quite some time since I made the branch, not sure what it needs to offer this functionality without offering impossible to validate input. Can you ping me in a couple of weeks? With the example you offered I can have another look when I have some time

AdSchellevis:
I did rebase https://github.com/opnsense/core/tree/kea_custom_opt_pr7361 , but realistically this doesn't bring it much further at the moment. It doesn't look very feasible to offer an interface to `client-classes` which is easy to use and is possible to validate in a uniform way.

pmhausen:
Came to this issue while checking the state of vendor option support.

> @mimugmail but if I'm not mistaken, they are mutually exclusive (so only one applies for that specific configuration)

Experience at least firmly says yes, although that might depend on how "intelligent" the client devices are.

DHCP provided unifi controller address confused the heck out of our SIP phones here which tried to parse it as their firmware bootstrap URL (also option 43). We ended up keeping the unifi option in place and configuring the phones statically through their UI.

Any ETA/target release for at least a rudimentary implementation? I need that unifi thing before ISC DHCP is removed ;) Given how far this issue has gone I will not try to hack my own solution and submit it.

Kind regards,
Patrick

AdSchellevis:
> Any ETA/target release for at least a rudimentary implementation? I need that unifi thing before ISC DHCP is removed ;) Given how far this issue has gone I will not try to hack my own solution and submit it.

no, not likely to appear either for reasons mentioned earlier. dnsmasq might be a better target to aim for or use a completely custom kea config (see also https://opnsense.org/roadmap/)

pmhausen:
What do you think of "prepackaged" UI options for the more common vendor options? Given the ubiquity (*grin*) of Unifi gear we could start with that and leave it to users to request more - mutually exclusive and with easy fail safe setup and proper constraints like "valid IP address" for Unifi for example.

mimugmail:
Isnt there a config include folder for such setups with kea?

AdSchellevis:
> Isnt there a config include folder for such setups with kea?

Unfortunately kea doesn't support includes, which either ends up in some hacky construct or not offering it at all, we choose the later for good reasons (see also https://github.com/opnsense/core/issues/7822). Somewhere down the line we came to the conclusion kea might not be the best option for small and medium sized networks (which often has other requirements than provider grade ones), which is what triggered addition of dnsmasq for dhcp on our end. 

I'm open for maintainable pull-requests, the time we have spend so-far on kea is a bit staggering to be honest, there are just limits in the amount of time I am willing to invest.

(maybe the best example so-far is about adding kea for v6, implementing the example as documented https://kea.readthedocs.io/en/stable/arm/dhcp6-srv.html#introduction, leading to the conclusion this can't work for "reasons" https://lists.isc.org/pipermail/kea-users/2018-January/001581.html)

AdSchellevis:
It won't offer a consistent state, you can merge json files, but the outcome is unpredictable (increasing the chances kea won't start, which as a result triggers unsolvable reports).

kea in my experience is quite fragile, which means data validation is crucial. When people want custom configurations, just ship the full config, which is possible since  https://github.com/opnsense/core/issues/7822

pmhausen:
> Also, implementing a specific variant for Unifi does not stand in the way of other vendor-specific option 43 use cases for other manufacturers, AFAIK.

Correct. As I found out with vendor classes you can have as many of them as you need. Will give unifi a try over the long weekend.

Monviech:
@AdSchellevis

https://help.ui.com/hc/en-us/articles/204909754-Remote-Adoption-Layer-3

https://forum.opnsense.org/index.php?topic=47370.msg238856#msg238856

```
You'll need to configure your DNS server to resolve unifi to your remote UniFi Network Application.
There are two methods of specifying the machine running the Network Application:
IP Address: [http://ip-address:8080/inform](http://xyz:8080/inform)
Fully Qualified Domain Name (FQDN): http://fqdn:8080/inform
```

Using this dhcp option 43 is optional for unifiy, not a strict requirement. 

Just leaving this here for those who come after.

mimugmail:
Yep, but there are other business solutions around too, using option 43 to send configuration to clients like VoIP phones

Kampfwurst:
It would be great if there would be a quick solution. I would need the option 138 for my TP AP and the Controller.

Wireheadbe:
> newer equipment is probably less likely to use option 43, but with dnsmasq it's likely a lot easier to match clients option 60 request to a proper 43 response anyway.

DNSmasq doesn't really play well with HA setups though, as there's no lease sync. I had to go to KEA for that very reason, only to then get stuck with missing DHCP options, so it's a bit of chicken/egg story. Being able to add DHCP options (to the user's discretion) would be handy though...

Monviech:
@Wireheadbe

Read this comment and use DNS instead:
https://github.com/opnsense/core/pull/7361#issuecomment-2919635609

Wireheadbe:
> @Wireheadbe
> 
> Read this comment and use DNS instead: [#7361 (comment)](https://github.com/opnsense/core/pull/7361#issuecomment-2919635609)

Doesn't work in my situation. My NVR is on "Unifi" but my network controller is on a different host (l3 adoption); due to having a large number of sites to handle ;) So I need option 43...

Monviech:
If it's a requirement you currently have no other choice but to do a manual KEA configuration (which is supported). If you read the general issues that comes with this option (its not generic in KEA, it is used by different vendors, implementation is a pain), it will not likely be added anytime soon.

Wireheadbe:
Why not just offer the possibility to add JSON snippets? And then have a button (just like on the HA-proxy page) to validate the config?
That would leave it to the users discretion to add advanced settings, while still having proper visibility over the config in the GUI and proper HA sync?

I realize that this isn't perfect, but for support cases you could simply say: make sure you haven't got anything in the "includes"-fields.

Monviech:
Adding raw configuration snippets like that (through the GUI) goes against our coding guidelines as it cannot be validated (when being serialized to the model/configuration, not the service validating the generated config). Snippets like that could also collide with future expansions. Also KEA likes to take any opportunity possible to crash if the config is not 100% correct.

All in all that also won't happen.

AdSchellevis:
https://github.com/opnsense/core/pull/7361#issuecomment-2919635609

Wireheadbe:
> [#7361 (comment)](https://github.com/opnsense/core/pull/7361#issuecomment-2919635609)

Hi Ad, with all due respect, but that is a workaround, and not a full solution. This is a workaround using DNS. There's plenty other equipment that does need this properly set.

AdSchellevis:
dnsmasq supports options like these out of the box on our end, with kea currently there doesn't seem to be a viable solution at this time. keeping his PR open longer just doesn't make much sense for that reason.

Wireheadbe:
Hi Ad,
I get that, but where would people use this? In a corp environment where Kea isn't optional (supporting dhcp failover and lease sync). DNSmasq doesn't cater for that failover and thus corp environments.

This is a chicken/egg situation, i get it, but it's a feature of Kea that's sadly missing in Opnsense.

Regards,
Jeffrey

miken32:
I hope ISC isn't going anywhere, because if opnsense isn't going to add support for custom configs to Kea we're not able to migrate. I don't care about unifi, just need to be able to set any DHCP option to any value, like you're able to do with every other DHCP server ever!

AdSchellevis:
to plugins (https://github.com/opnsense/core/issues/9155) and next step out of support, which from isc's perspective has been the case for quite some time already (https://www.isc.org/blogs/isc-dhcp-eol/). 

you can manage your own settings with kea, just not via the gui. has been the case for quite some time already (https://github.com/opnsense/core/issues/7822).

Holladiewal:
I would really like to get at least some basic option 43 support into the GUI. 
I did some basic testing with Kea and setting the option 43 data is possible without client classes (if I understand the above discussion right, this was one of the complication concerns raised) with just the following snippet within the subnet configuration:

```
"option-data": [
    {
        "name": "vendor-encapsulated-options",
        "csv-format": false,
        "data": "01040A3B3714"
    }
]
```

The data is the raw TLV-data that is delivered to the client, therefore only "requiring" minimal validation, if any.

Is this something that would have any chance of being merged?
If so, I would be willing to look into making a PR with some inspiration from the work done in this PR.

Monviech:
Most likely fixed via:

Issue #7307, opened 2024-03-08
Register KEA DHCP Mappings with Unbound


drinn:
**Important notices**

Before you add a new report, we ask you kindly to acknowledge the following:

- [X] I have read the contributing guide lines at https://github.com/opnsense/core/blob/master/CONTRIBUTING.md
- [X] I am convinced that my issue is new after having checked both open and closed issues at https://github.com/opnsense/core/issues?q=is%3Aissue

**Is your feature request related to a problem? Please describe.**

No problem reported. Just a new feature request.

**Describe the solution you like**

I would like to have KEA DHCP mappings be resolvable via Unbound. This has been a mainstay with ISC DHCP for as long as I've been using opnsense. I understand after reading through #7237 that this doesn't have priority at the moment, but I didn't want it to get lost as a requested item in the shuffle.

Having all hosts on the network easily identifiable by DNS lookups is important to me on the networks that I manage. It would be nice for this functionality to be extended in KEA if that is the go forward solution to DHCP services.

Totally understand that it may not happen in the near future and am happy using ISC DHCP until this is possible.

**A clear and concise description of what you want to happen.**

To be able to have static DHCP mappings in KEA be fully resolvable in Unbound.

**Describe alternatives you considered**

My alternative for the time being is to continue to use ISC DHCP. If there is a scenario down the road where ISC DHCP disappears and KEA still does not have this feature, I may have to manually enter all static mappings into Unbound manually. In that case, it would be nice to have an export hosts from KEA tool to the import into unbound?

PR #7362, opened 2024-04-05
Support static DNS mappings using Kea DHCP.

reitermarkus:
Closes https://github.com/opnsense/core/issues/7307.


fichtner:
I've added https://github.com/opnsense/core/commit/14cc9a1c2f and https://github.com/opnsense/core/commit/b53fe7c1d9 to avoid tainting private static_mapping consumers after discussion with @AdSchellevis.

I'm not against this, but this needs careful crafting because the current PR is (in light of the impending breakage it would have created above) not complete. Here are a few thoughts:

- For one the action is not required and shouldn't be added until its needed.
- You should check if the static mapping's server is actually enabled.
- Given the lack of an IPv6 kea implementation we shouldn't mess with looping over IPv6 entries that don't exist.

Cheers,
Franco

reitermarkus:
> For one the action is not required and shouldn't be added until its needed.

So the action is not needed for the plugin function to work?

> You should check if the static mapping's server is actually enabled.

This is now checked.

> Given the lack of an IPv6 kea implementation we shouldn't mess with looping over IPv6 entries that don't exist.

I have removed the useless loop.

reitermarkus:
@fichtner, can you have another look here?

src/etc/inc/plugins.inc.d/kea.inc
```diff
@@ -45,6 +45,68 @@ function kea_services()
...
+                    $hostname = null;
+                }
+                $description = (string)$reservation->description;
+                if (empty($description)) {
+                    $description = null;
+                }
+
+                $subnet_node = $keav4->getNodeByReference("subnets.subnet4.{$reservation->subnet}");
+
+                $domain = $domain_fallback;
```
fichtner, line 86: we should maybe considering removing domain_fallback or at least not implementing it for kea.. it was only a workaround for unbound when it wasn't capable of setting the right domain

src/etc/inc/plugins.inc.d/kea.inc
```diff
@@ -45,6 +45,68 @@ function kea_services()
...
+                if ($subnet_node) {
+                    if (!empty((string)$subnet_node->option_data->domain_search)) {
+                        $domain = (string)$subnet_node->option_data->domain_search;
```
fichtner, line 89: I'm not sure why this is a good idea?

src/etc/inc/plugins.inc.d/kea.inc
```diff
@@ -45,6 +45,68 @@ function kea_services()
...
+                    } else if (!empty((string)$subnet_node->option_data->domain_name)) {
+                        $domain = (string)$subnet_node->option_data->domain_name;
+                    }
+                }
+                $domain = preg_split('/[ ;]+/', $domain)[0];
```
fichtner, line 94: probably only added because of domain search...

fichtner: ok this is because of the isc dhcpv6 complictatios... let's just scratch that.

src/etc/inc/plugins.inc.d/kea.inc
```diff
@@ -45,6 +45,68 @@ function kea_services()
...
+                    continue;
+                }
+                $hostname = (string)$reservation->hostname;
+                if (empty($hostname)) {
+                    $hostname = null;
+                }
+                $description = (string)$reservation->description;
+                if (empty($description)) {
+                    $description = null;
+                }
```
fichtner, line 82: this feels a bit overcomplicated, empty strings should be fine. you end up thowing a PHP warning now or later because empty() should be the first sanity check then which can mute the warning. Here it does nothing.

src/etc/inc/plugins.inc.d/kea.inc
reitermarkus: I just copied what `dhcpd_staticmap` does, but I guess that only applies to IPv6 it seems.

src/etc/inc/plugins.inc.d/kea.inc
```diff
@@ -45,6 +45,68 @@ function kea_services()
...
+                    }
+                }
+                $domain = preg_split('/[ ;]+/', $domain)[0];
+
+                $entry = [
+                    'descr' => $description,
+                    'domain' => $domain,
+                    'hostname' => $hostname,
+                    'ipaddr' => $ip_address,
+                ];
```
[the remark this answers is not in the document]
reitermarkus: Where to get the interface from? Or just set it to `null`?

src/etc/inc/plugins.inc.d/kea.inc
reitermarkus: > you end up thowing a PHP warning now or later because empty() should be the first sanity check then which can mute the warning

I'm not quite following, should the `empty` check be removed completely or moved somewhere else?

src/etc/inc/plugins.inc.d/kea.inc
reitermarkus: So it should not use the system domain as fallback? 

https://github.com/opnsense/core/blob/24.7.a/src/etc/inc/plugins.inc.d/unbound.inc#L586

dMopp:
I would love to see it in the next release :)

fichtner (approved):
good for merge! all relevant parts to make this happen are in 24.1.7 now

fichtner:
Merged, thanks!

dMopp:
Just for myself: that means it will be part of the NEXT release, right?

fichtner:
Next development version. Final release in 24.1.9 most likely.

dMopp:
opnsense-patch 139a3ad Works well :). Thanks. Next stop: dynamic leases into unbound 👯‍♂️

fichtner:
It's not the full patching. I'd say it works but I'd rather have confirmation on latest development state because there it works without the obvious issues it introduces into the wider system.

dMopp:
Yeah, all fine, for "me" it seems to work fine. And i can always revert if in case of any issues

fichtner:
Well, the point was that if the actual development state is tested it helps us more to ship the feature than cherry-picking the PR commit because mistakes can happen. I know it works as a concept because it was reviewed and merged, but it doesn't reflect the full state of changes necessary. That's all.

dMopp:
Ah, okay I understood. Might consider it, but it might affect the woman acceptance factor 😅

fichtner:
We all have our reasons, no problem. I just wanted to get that point across. In any case kudos for using opnsense-patch to subvert the WAF. ;)

tim-oe:
what 24.7.x release will this be part of?
running 24.7.8 and not seeing any option under unbounded general settings

fichtner:
> community/24.1/24.1.9:o kea-dhcp: support static DNS mappings (contributed by Markus Reiter)

so... any 24.7 release?

PR #7363, opened 2024-04-05
dashboard: widgets init next update after response

DerMistkaefer:
I'm having the problem that my OPNSENSE DEC2685 is running at 100% CPU usage every time I open the main web dashboard page.
I ignored the problem for a while, but now I'm starting to investigate it.

The main problem is that the widgets continuously request the APIs even if there is no response due to CPU overload.
Therefore, I think the best solution is to request new data after the previous request is answered.

I checked all widgets in the /src/www/widgets/widgets folder and made the change.
This behavior is already the case in the Wiregard widget.

image

### DIFF
--- a/src/www/widgets/widgets/gateways.widget.php
+++ b/src/www/widgets/widgets/gateways.widget.php
@@ -114,8 +114,8 @@ $gateways = (new \OPNsense\Routing\Gateways())->gatewaysIndexedByName();
                         }
                     });
                 }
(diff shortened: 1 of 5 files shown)


fichtner:
Due to risk of uncontrollable regression and since the dashboard is being replaced for 24.7 I’d recommend not adjusting this in 24.1.x.

DerMistkaefer:
Okay, thanks for the information than I hope this will get fixed in the new dashboard.

PR #7369, opened 2024-04-09
WireGuard: Add description for restart

mimugmail:
Sometime we need to restart wireguard during the night cause of ISP disruptions.
Can we have description to restart wireguard via cron?


AdSchellevis:
Hi Michael,

I don't mind adding a description, although it does sound a bit odd a restart is needed, I would expect a keep alive should also awake the tunnel.

Best regards,

Ad

mimugmail:
Same here, there is a hickup during multiwan and failover back and forth. I wasn't able to reproduce locallly (cause of stable ISP) :)

PR #7375, opened 2024-04-11
FW Log Live View - Layout consistency & space use

NOYB:
Make controls area layout more consistent with other log pages.
Reduce unused/inefficient vertical space use.


AdSchellevis:
only works on wide screens, we keep it as is.

Issue #7372, opened 2024-04-10
Anitlockout rule opens ports on WAN interface


Curly060:
**Important notices**

Before you add a new report, we ask you kindly to acknowledge the following:

- [X] I have read the contributing guide lines at https://github.com/opnsense/core/blob/master/CONTRIBUTING.md
- [X] I am convinced that my issue is new after having checked both open and closed issues at https://github.com/opnsense/core/issues?q=is%3Aissue

**Describe the bug**

Under some circumstances the anti lockout firewall rule chooses a WAN interface and then WebGUI (ports 80,443) and SSH (port 22, if enabled) are accessible on the WAN! That is for sure not desired behaviour!

**To Reproduce**

1. install a fresh OPNsense on a host with 4 NICs
2. create a 2nd WAN interface (WAN_NEW)
3. create a 2nd LAN interface (LAN_NEW) and an appropriate "allow" firewall rule
4. at this point verify that both WANs have no open ports
5. delete the old WAN and LAN interfaces
6. reload all services
7. check the WAN again for open ports

=> Now port 80 and 443 are open on WAN (and 22 if SSH access was enabled)

**Note:** The order of creation of WAN_NEW and LAN_NEW is important: WAN_NEW first, then LAN_NEW

**Expected behavior**
The anti lockout rule should *never* apply to a WAN interface.

**Describe alternatives you considered**
Enabling the setting "Disable anti-lockout" in Firewall/Settings/Advanced prevents this from happening, of course. This will be my default setting from now on.

**Environment**

Software version used and hardware type if relevant, e.g.:
OPNsense 24.1.3_1-amd64
FreeBSD 13.2-RELEASE-p10

Curly060:
And that is exactly why the machine shouldn't guess!

I found out the hard way, suddenly my firewall was open, even though I did not make any changes to the Firewall rules. It was simply because I made changes to my interfaces and then a default(!) setting kicked in opening ports on WAN.

That is quite unexpected. Choosing **any** interface is actually unexpected. The setting could easily be made more robust:
Instead of a checkbox it could be a select box where one can select the interface that the rule shall apply for. By default it is LAN, nothing selected means disabled and if someone tries to delete the selected interface a warning could appear thatg this will disable the anti lockout rule.

No more guessing, no surprises. So what do you think?

I know I can disable it (I wrote that) and I have now. But this is a default setting and it was in its default state all those years I have been using OPNsense...

P.S.: What does the "support" label mean?

[...]

PR #7381, opened 2024-04-14
Firewall: fix issues with antilockout firewall rule

Curly060:
This commit fixes two issues with the antilock firewall rule:
- the rule may randomly open ports on the WAN under certain circumstances
- the rule may select inactive interfaces

The issue is fixed by giving the user a choice for which interface the rule should apply for, default being LAN. The user is also being prevented from deleting/unassigning/disabling the currently selected interface.

### DIFF
--- a/src/etc/inc/console.inc
+++ b/src/etc/inc/console.inc
@@ -383,6 +383,7 @@ EOD;
         config_read_array('interfaces', 'lan');
         $config['interfaces']['lan']['if'] = $lanif;
         $config['interfaces']['lan']['enable'] = true;
+        $config['interfaces']['lan']['antilockout'] = true;
 
         if ($new) {
             $config['interfaces']['lan']['ipaddr'] = '192.168.1.1';
--- a/src/etc/inc/filter.lib.inc
+++ b/src/etc/inc/filter.lib.inc
@@ -94,9 +94,12 @@ function filter_core_get_antilockout()
         return [];
     }
 
-    $lockout_if = get_primary_interface_from_list(array_keys(legacy_config_get_interfaces(['virtual' => false])));
-    if (empty($lockout_if)) {
+    $lockout_ifs = array_keys(legacy_config_get_interfaces(['virtual' => false, 'enable' => true, 'antilockout' => true]));
+    if (empty($lockout_ifs)) {
+      $lockout_ifs = [get_primary_interface_from_list(array_keys(legacy_config_get_interfaces(['virtual' => false])))];
+      if (empty($lockout_ifs)) {
         return [];
+      }
     }
 
     /*
@@ -132,8 +135,7 @@ function filter_core_get_antilockout()
 
     sort($lockout_ports);
 
-    /* return a convenient one-entry array to iterate over for our callers */
-    return [$lockout_if => $lockout_ports];
+    return array_combine($lockout_ifs, array_fill(0, count($lockout_ifs), $lockout_ports));
 }
 
 /**
--- a/src/opnsense/scripts/shell/setaddr.php
+++ b/src/opnsense/scripts/shell/setaddr.php
@@ -542,6 +542,12 @@ if (console_prompt_for_yn('Restore web GUI access defaults?', 'n')) {
         unset($config['system']['webgui']['noantilockout']);
         $restart_webgui = true;
     }
+    if (isset($config['interfaces']['lan'])) {
+        $config['interfaces']['lan']['antilockout'] = true;
+        $restart_webgui = true;
+    } else {
+        echo "Could not enable anti lockout rule because no LAN interface was found.\n";
+    }
     if (isset($config['system']['webgui']['interfaces'])) {
         unset($config['system']['webgui']['interfaces']);
         $restart_webgui = true;
--- a/src/etc/config.xml.sample
+++ b/src/etc/config.xml.sample
@@ -269,6 +269,7 @@
       <mediaopt/>
       <track6-interface>wan</track6-interface>
       <track6-prefix-id>0</track6-prefix-id>
+      <antilockout>true</antilockout>
     </lan>
   </interfaces>
   <dhcpd>
--- a/src/www/interfaces.php
+++ b/src/www/interfaces.php
@@ -420,6 +420,7 @@ if ($_SERVER['REQUEST_METHOD'] === 'GET') {
     }
     $pconfig['enable'] = isset($a_interfaces[$if]['enable']);
     $pconfig['lock'] = isset($a_interfaces[$if]['lock']);
+    $pconfig['antilockout'] = isset($a_interfaces[$if]['antilockout']);
     $pconfig['blockpriv'] = isset($a_interfaces[$if]['blockpriv']);
     $pconfig['blockbogons'] = isset($a_interfaces[$if]['blockbogons']);
     $pconfig['gateway_interface'] = isset($a_interfaces[$if]['gateway_interface']);
@@ -596,6 +597,11 @@ if ($_SERVER['REQUEST_METHOD'] === 'GET') {
     } elseif (empty($pconfig['enable'])) {
         if (isset($a_interfaces[$if]['enable'])) {
             unset($a_interfaces[$if]['enable']);
+        }
+        if (!empty($pconfig['antilockout'])) {
+            $a_interfaces[$if]['antilockout'] = true;
+        } elseif (isset($a_interfaces[$if]['antilockout'])) {
+            unset($a_interfaces[$if]['antilockout']);
         }
         if (!empty($pconfig['lock'])) {
             $a_interfaces[$if]['lock'] = true;
@@ -1013,6 +1019,7 @@ if ($_SERVER['REQUEST_METHOD'] === 'GET') {
             $new_config['descr'] = preg_replace('/[^a-z_0-9]/i', '', $pconfig['descr']);
             $new_config['enable'] = !empty($pconfig['enable']);
             $new_config['lock'] = !empty($pconfig['lock']);
+            $new_config['antilockout'] = !empty($pconfig['antilockout']);
             $new_config['spoofmac'] = $pconfig['spoofmac'];
 
             $new_config['blockpriv'] = !empty($pconfig['blockpriv']);
@@ -1759,6 +1766,19 @@ include("head.inc");
                         <td>
                           <input id="lock" name="lock" type="checkbox" value="yes" <?=!empty($pconfig['lock']) ? 'checked="checked"' : '' ?>/>
                           <?= gettext('Prevent interface removal') ?>
+                        </td>
+                      </tr>
+                      <tr>
+                        <td><a id="help_for_antilockout" href="#" class="showhelp"><i class="fa fa-info-circle"></i></a> <?= gettext('Antilockout') ?></td>
+                        <td>
+                          <input id="antilockout" name="antilockout" type="checkbox" value="yes" <?=!empty($pconfig['antilockout']) ? 'checked="checked"' : '' ?>/>
+                          <?= gettext('Enable antilockout firewall rule on this interface') ?>
+                          <div class="hidden" data-for="help_for_antilockout">
+                            <?= gettext(<<<EOD
+                                When this is checked, an anti lockout rule is generated which always permits access to the web GUI and SSH on this interface,
+                                regardless of the user-defined firewall rule set, unless the anti lockout feature is disabled in the advanced firewall settings.
+                                EOD); ?>
+                          </div>
                         </td>
                       </tr>
                       <tr>


src/etc/inc/filter.lib.inc
```diff
@@ -90,13 +90,23 @@ function filter_core_get_antilockout()
 {
     global $config;
 
+    // Migration from old noantilockout option
```
AdSchellevis, line 93: Assumes the default (new) configuration contains a fixed choice, which circles back to a static `lan` in `config.xml.sample`,  when not persisting lan, initial boot after install will likely always persist `lan` here anyway, which might then be altered by an interface mismatch later on. 

When removing the side affect `write_config();`, a migration also will assume a static situation, which past experiences learn has its limitations.

It's also good to remember, you're only able to cause your issue by removing existing interfaces, which almost sound like if we didn't let you remove "WAN", you didn't have an issue either..... (probably not true, but similar to the changes proposed)

Just my two cents on the subject.

Curly060: Yes, that's what I [wrote in the related issue of this PR](https://github.com/opnsense/core/issues/7372#issuecomment-2054016915): The place and the way of the migration doesn't feel right, which is due to my lack of knowledge about the internal workings of the config system. A one-time migration is most likely the better choice, but I'd need guidance as how to do that.

And it is not the removal of the WAN interface that causes trouble, it is the removal of the LAN interface (more precisely: the interface that in the config is named "lan"). Once that is gone, the opt with the lowest number is chosen:

If that opt happens to act as a WAN => antilockout rule opens ports on that WAN
If that opt happens to be disabled => antilockout rule doesn't work at all

src/etc/inc/console.inc
```diff
@@ -384,6 +384,7 @@ EOD;
         $config['interfaces']['lan']['if'] = $lanif;
         $config['interfaces']['lan']['enable'] = true;
 
+        $config['system']['webgui']['antilockout_interface'] = 'lan';
```
fichtner, line 387: as said in the issue I'd prefer a per-interface field instead of a loose list maintained somewhere else.

src/www/system_advanced_firewall.php
```diff
@@ -139,11 +139,7 @@
             unset($config['system']['lb_use_sticky']);
         }
 
-        if (!empty($pconfig['noantilockout'])) {
-            $config['system']['webgui']['noantilockout'] = true;
-        } elseif (isset($config['system']['webgui']['noantilockout'])) {
-            unset($config['system']['webgui']['noantilockout']);
-        }
+        $config['system']['webgui']['antilockout_interface'] = $pconfig['antilockout_interface'];
```
fichtner, line 142: historically these constructs exist to avoid PHP warnings and notices (check development mode)

src/www/system_advanced_firewall.php
```diff
@@ -732,19 +728,31 @@
                   </div>
                 </td>
               </tr>
-              <tr>
-                <td><a id="help_for_noantilockout" href="#" class="showhelp"><i class="fa fa-info-circle"></i></a> <?=gettext("Disable anti-lockout"); ?></td>
+              <tr id="antilockout_interface">
```
fichtner, line 731: To be frank adding a better way but removing the old way to make it "safe" is questionable at best.

fichtner (changes requested):
Ok there is the thing:

Redesigning the feature and implementing it in a complimentary way is probably easier, but I see potential for hours of work and going to the drawing board to actually come up with a better system that prevents a similar amount of lockout-protection that the current system is offering.

A flaw in the system doesn't make the whole system worthless and that's why removing it will just create friction with existing users that I'm not going to debate including.

Maybe complications are unavoidable, but we better know all of them. See my "user view" questions.

src/etc/inc/console.inc
fichtner: User view: I only use WAN and/or OPTX and I still want anti-lockout. What do I do on the console? A single WAN will also lock itself out...

src/etc/inc/filter.lib.inc
```diff
@@ -90,13 +90,23 @@ function filter_core_get_antilockout()
...
     if (isset($config['system']['webgui']['noantilockout'])) {
-        return [];
+      unset($config['system']['webgui']['noantilockout']);
+    }
+    if (!isset($config['system']['webgui']['antilockout_interface'])) {
+      $lockout_if = get_primary_interface_from_list(array_keys(legacy_config_get_interfaces(['virtual' => false, 'enabled' => true])));
```
fichtner, line 98: User view: Why are we adding compatibility but don't provide a way to turn this off anymore? I just want the old setting to remain and work as before.

src/www/interfaces_assign.php
```diff
@@ -51,6 +51,13 @@ function link_interface_to_group($int)
     return $result;
 }
 
+function link_interface_to_antilockout($int)
```
fichtner, line 54: Can we please avoid legacy glue constructs like this?

src/www/interfaces_assign.php
```diff
@@ -155,6 +162,8 @@ interface_sync_wireless_clones($config['interfaces'][$newifname], false);
             $input_errors[] = gettext("The interface is part of a gre tunnel. Please delete the tunnel to continue");
         } elseif (link_interface_to_gif($id)) {
             $input_errors[] = gettext("The interface is part of a gif tunnel. Please delete the tunnel to continue");
+        } elseif (link_interface_to_antilockout($id)) {
+            $input_errors[] = gettext("The interface is used by the anti lockout firewall rule. Please select a different one to continue");
```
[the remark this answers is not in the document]
fichtner: User view: ok now I need to go to some other page to set a new anti-lockout interface to be able to delete this and then I'll just forget to set anti-lockout I just want to create a new interface...

Curly060:
> Ok there is the thing:
> 
> Redesigning the feature and implementing it in a complimentary way is probably easier, but I see potential for hours of work and going to the drawing board to actually come up with a better system that prevents a similar amount of lockout-protection that the current system is offering.
> 
> A flaw in the system doesn't make the whole system worthless and that's why removing it will just create friction with existing users that I'm not going to debate including.
> 
> Maybe complications are unavoidable, but we better know all of them. See my "user view" questions.

Yes, I do like this idea much better than the old way. Also, it was much easier to implement this than the old way. So I have updated ([force-pushed](https://github.com/opnsense/core/compare/89179231ab68c457f17508a54cd3b0e016fbed31..5beb223077bf6fae960c7da3eede1c0474c973d2)) this PR implementing the way you proposed, see commit https://github.com/opnsense/core/commit/5beb223077bf6fae960c7da3eede1c0474c973d2.

Screenshots:
The setting:
grafik
Prevent deletion (popup is only blurry in this screenshot, not in reality):
grafik

Resetting to defaults will enable the anti lockout on the LAN interface.

One thing is missing, because I couldn't find out how to do it:
The migration of the configuration from the old logic to the new one. https://docs.opnsense.org/development/frontend/models_migrations.html do not seem to apply here. The interfaces do not have a model (or do they and I was blind?).

src/www/system_advanced_firewall.php
```diff
@@ -43,7 +43,6 @@
     $pconfig['maximumfrags'] = isset($config['system']['maximumfrags']) ? $config['system']['maximumfrags'] : null;
     $pconfig['adaptivestart'] = isset($config['system']['adaptivestart']) ? $config['system']['adaptivestart'] : null;
     $pconfig['adaptiveend'] = isset($config['system']['adaptiveend']) ? $config['system']['adaptiveend'] : null;
-    $pconfig['noantilockout'] = isset($config['system']['webgui']['noantilockout']);
```
fichtner, line 46: Not sure if I said this earlier, but it's not a good idea to aim for a replacement at this point without a migration strategy.

Curly060: Ah, now I got what you meant. You still want to keep a global on/off switch for the whole feature. Yes, that makes sense.

So I have updated the PR once again and now the changes are as minimal as possible:
Interfaces can have an attribute "antilockout". Interfaces with this attribute take precedence when trying to find interfaces to activate the anti lockout rule for. If no such interfaces are found, the old (buggy) logic applies.

If the "Disable anti-lockout" setting in Firewall/Settings/Advanced is checked, no anti lockout rules will be generated, regardless of the new attribute.

Resetting to defaults (or generating the default configuration) will set the new antilockout attribute on the LAN interface.

This way there is also no need to migrate anything for now.

This doesn't fully solve the issue, but it makes it even less likely and now there is actually an option to define the interfaces for which the anti lockout rule shall apply for.

src/www/system_advanced_firewall.php
```diff
@@ -738,13 +738,13 @@
...
                   <?= gettext('Disable administration anti-lockout rule') ?>
                   <div class="hidden" data-for="help_for_noantilockout">
-                    <?= sprintf(gettext("When this is unchecked, access to the web GUI or SSH " .
-                                "on the %s interface is always permitted, regardless of the user-defined firewall " .
-                                "rule set. Check this box to disable the automatically added rule, so access " .
-                                "is controlled only by the user-defined firewall rules. Ensure you have a firewall rule " .
-                                "in place that allows you in, or you will lock yourself out."),
-                                count($config['interfaces']) == 1 && !empty($config['interfaces']['wan']['if']) ?
-                                gettext('WAN') : gettext('LAN')) ?>
+                    <?= gettext(<<<EOD
```
fichtner, line 741: this breaks translations. not sure why that's a good idea

src/opnsense/scripts/shell/setaddr.php
```diff
@@ -542,6 +542,10 @@ function console_configure_dhcpd($version = 4)
         unset($config['system']['webgui']['noantilockout']);
         $restart_webgui = true;
     }
+    if (empty($config['interfaces'][$interface]['antilockout'])) {
```
fichtner, line 545: eh, huh, why? you can use it to configure a WAN too...

src/etc/inc/filter.lib.inc
```diff
@@ -556,7 +558,7 @@ function filter_core_rules_system($fw, $defaults)
                 'to' => '(self)',
                 'to_port' => implode(' ', $lockoutprts),
                 'descr' => 'anti-lockout rule',
-                '#ref' => 'system_advanced_firewall.php#noantilockout'
+                '#ref' => 'interfaces.php?if=' . $lockoutif
```
fichtner, line 561: I don't think that makes sense as both could be valid.

src/etc/inc/console.inc
```diff
@@ -383,6 +383,7 @@ EOD;
         config_read_array('interfaces', 'lan');
         $config['interfaces']['lan']['if'] = $lanif;
         $config['interfaces']['lan']['enable'] = true;
+        $config['interfaces']['lan']['antilockout'] = true;
```
fichtner, line 386: config.xml.sample needs this too, but you have to consider if people skip LAN the unfavourable (as per your view) implementation wins and keeps doing what it always did.

it might just be better to stay clear and leave the new antilockout as an opt-in for the forseeable future.

src/www/system_advanced_firewall.php
Curly060: reverted

src/www/interfaces_assign.php
```diff
@@ -433,11 +433,23 @@ interface_configure(false, $ifname, true);
                           </select>
                         </td>
                         <td>
-<?php if (empty($iface['lock'])): ?>
-                          <button title="<?= html_safe(gettext('Delete')) ?>" data-toggle="tooltip" data-id="<?=$ifname;?>" class="btn btn-default act_delete" type="submit">
-                            <i class="fa fa-trash fa-fw"></i>
-                          </button>
-<?php endif ?>
+<?php
```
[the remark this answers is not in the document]
Curly060: reverted

src/opnsense/scripts/shell/setaddr.php
Curly060: I assumed the user would not choose "yes" here when configuring a WAN interface, but yes, the user could. It will configure antilockout for the LAN interface now and if no LAN is present, print a warning about it.

src/etc/inc/filter.lib.inc
Curly060: true, it was a left over from previous work. reverted

src/etc/inc/console.inc
fichtner: think more about it it would make sense to simply ask the user if they want to participate in the anti-lockout group with this interface. It should be safe, because in this assignment run we assign all ports so there is no ambiguity about it.

src/opnsense/scripts/shell/setaddr.php
```diff
@@ -542,6 +542,12 @@ function console_configure_dhcpd($version = 4)
         unset($config['system']['webgui']['noantilockout']);
         $restart_webgui = true;
     }
+    if (isset($config['interfaces']['lan'])) {
```
fichtner, line 545: What this code tries to achieve is go back to the factory defaults. We are not the LAN interface... we are any interface here and reverting to defaults attempts to restore GUI access, but not for a definitive "lan" interface. This script allows the user to select the interface to set up prior (and asking for anti-lockout there might be more sensible).

fichtner: (more to your original point the 'lan' key could have been repurposed as the WAN interface in the meantime)

src/etc/inc/filter.lib.inc
```diff
@@ -94,9 +94,12 @@ function filter_core_get_antilockout()
         return [];
     }
 
-    $lockout_if = get_primary_interface_from_list(array_keys(legacy_config_get_interfaces(['virtual' => false])));
-    if (empty($lockout_if)) {
+    $lockout_ifs = array_keys(legacy_config_get_interfaces(['virtual' => false, 'enable' => true, 'antilockout' => true]));
+    if (empty($lockout_ifs)) {
```
fichtner, line 98: I'm still split on this fallback. It could also be considered to move `noantilockout` to only refer to the automatic anti-lockout behaviour and never mess with the interface-based setting.

Curly060:
Uh, now I feel this has become like a moving target and I don't really know anymore where this should go. I'd like a clear path forward. So I propose the following strategy to continue on this topic:

1. Define clear goals which actually truly fix the problem and that can be reached with a minimum amount of work and impact on existing installations
2. Once agreed on those goals, define measures to reach these goals
3. Once agreed on the measures, implement them

I am happy to do all these steps, so here is my proposal for step 1, the goals:
**Goal 1:** The router shall not accidentally open ports on undesired interfaces.
**Goal 2:** The end user shall not accidentally close ports on desired interfaces.
**Goal 3:** Existing configurations shall not change their current behavior due to the necessary changes.

Can we agree on this strategy to go forward and on these goals? If not, please propose other clear and reachable goals that fix the problem at hand.

fichtner:
Skipping the design phase of this is exactly what makes this a moving target and I said so initially:

> Redesigning the feature and implementing it in a complimentary way is probably easier, but I see potential for hours of work and going to the drawing board to actually come up with a better system that prevents a similar amount of lockout-protection that the current system is offering.

Your goals are ok, but you have to keep regressions in mind, offering a complimentary system, namely:

> Goal 1: The router shall not accidentally open ports on undesired interfaces.

Yes, but now you already manoeuvred yourself into a dead-end because this will not work without breaking other people's setup. You will lock someone out and I don't think it's worth it.

> Goal 2: The end user shall not accidentally close ports on desired interfaces.

This is unrealistic because the user can delete interfaces (if not from the GUI then from the console) and then he/she locked himself/herself out.

> Goal 3: Existing configurations shall not change their current behavior due to the necessary changes.

This clearly conflicts with goal 1.

The per interface setting presented here is looking promising, but it's not done if you want people to accept and use the new way of doing it. You've discarded a number of my concerns and requests that were in line with your desire for everyone to use it (adding the new setting to the default config.xml) and you touched code where otherwise not relevant or dangerous because you missed the scope of the script (like setaddr.sh).

I know this stuff is hard, but I've done it for 10 years and if something doesn't look good enough it will certainly not be for the users when this will be included in a release it must be as good as you can get it and it has to make sense for everyone who didn't follow this discussion too. I think for now it just makes sense for you and that's what you aim for, which is a fine enough motivation to get started, but not something that can be included without severe repercussions, post-release fixes and maybe 1-2 years of further tinkering and the bad feeling that it should not have been merged permaturely.

Take it for what it is. I'm only here to guide and shape the submission if you are willing.


Cheers,
Franco

Curly060:
You said (highlighting by me):

> > Redesigning the feature and implementing it in a **complimentary** way is probably easier, 

And this is exactly what I have in mind. A complimentary feature. So either the old logic is active, or the new one. Never both. The user can choose. But that's already discussing step 2 (measures). I'd like to discuss step 1 (goals, and only the goals) first before going into step 2.

> > Goal 1: The router shall not accidentally open ports on undesired interfaces.
> 
> Yes, but now you already manoeuvred yourself into a dead-end because this will not work without breaking other people's setup. You will lock someone out and I don't think it's worth it.

I totally disagree. I do not want to break existing configurations (goal 3) and I do not want to lock anyone out (goal 2).
The goal is: "The router shall not accidentally open ports on undesired interfaces."
Independent of the implications this goal may have: Is this a goal we can agree on, yes or no?

> > Goal 2: The end user shall not accidentally close ports on desired interfaces.
> 
> This is unrealistic because the user can delete interfaces (if not from the GUI then from the console) and then he/she locked himself/herself out.

Then let's prevent that from happening! But that's already discussing an implementation detail. I want to focus on the goal itself:
"The end user shall not accidentally close ports on desired interfaces."
Can we agree on this goal, yes or no?

> > Goal 3: Existing configurations shall not change their current behavior due to the necessary changes.
> 
> This clearly conflicts with goal 1.

No. See above, complimentary feature.
Again, this is a simple goal and again, I do not (yet) want to discuss implications, so:
Can we agree on this goal, yes or no?

> Take it for what it is. I'm only here to guide and shape the submission if you are willing.

I am willing if we can reach an agreement on the goals. I kept them very simple on purpose. If they are not clear enough, let me know. Only in the next step I would like to discuss measures how to reach them. And then we will see what implications those measures are bringing in and if those implications are acceptable or not. But not before we have an agreement (or not) on the goals. So let's sort that out first, please.

Cheers, Curly060 =;->

fichtner:
> And this is exactly what I have in mind. A complimentary feature.

Is that why you removed the other feature in the first submission? Sorry I don't want to nitpick but you have to be realistic.

Goal 1 is not well-defined. "The router" in the factory config will do this in the current state of the PR. Hence I'm pressing on the right approach for the factory reset, whatever it is going to be. The goal is not reached.

Goal 2 also needs to predict user mistakes. We don't have to prevent anything and I don't think we can with the per-interface default. If the user wants to delete an interface he can and will. If the strategy for new installs as per Goal 1 is to not offer the old fallback (no anti-lockout enabled) then there is no way we can add another interface in anti-lockout.

Goal 3 fair enough but now you need to make sure the expectation matches the reality. There is a knob to turn of anti-lockout but you've bound your new feature to it. The user using the new anti-lockout will lock itself out by enabling this option. I don't think that is expected or wanted because it fails to adhere to a simple and fool-proof approach.

Together we defined a per interface feature that works quite well from the looks of it, but you haven't exposed it to the user in all the cases in the right manner where it matters:

1. factory reset configuration
2. setports.php
3. setaddr.php (asking for anti-lockout yes/no for the active interface, resetting default access needs to be discussed because we have not established what the default is going to be)
4. dealing with the proper strategy between anti-lockout disable, the automatic anti-lockout and the manual anti-lockout points.


Cheers,
Franco

Curly060:
> Is that why you removed the other feature in the first submission? Sorry I don't want to nitpick but you have to be realistic.

I removed stuff because I think you did not want them. It was not clear to me what you actually want, because except for the good idea to make it interface based you only ever said what you don't want. That's why I suggested to start from scratch. But now the same thing happens again:
I propose goals and you say they are not clear/well defined. Then please, make them clear/well defined. 

> Goal 1 is not well-defined.

How is that not well defined? I want a firewall to never ever open ports in an unexpected way, esp not on unexpected interfaces. That's one of the most basic expectations I have on a firewall. How else to formulate it?

> Goal 2 also needs to predict user mistakes. We don't have to prevent anything and I don't think we can with the per-interface default. If the user wants to delete an interface he can and will. If the strategy for new installs as per Goal 1 is to not offer the old fallback (no anti-lockout enabled) then there is no way we can add another interface in anti-lockout.

Then let's not make the "per interface" the default, see below.

> Goal 3 fair enough but now you need to make sure the expectation matches the reality. There is a knob to turn of anti-lockout but you've bound your new feature to it. The user using the new anti-lockout will lock itself out by enabling this option. I don't think that is expected or wanted because it fails to adhere to a simple and fool-proof approach.

I will not bind the new feature to that switch.

Ok, even though we haven't agreed on the goals yet, here is my proposal for step 2, the measures to reach the goals I had proposed. Maybe that makes the goals more clear to you:

Measures:
1. Add possibility to configure an anti lockout rule explicitly on each interface (goal 1)
2. Prevent removal of interface where the anti lockout flag is set (goal 2)
3. Set all flags per default to "not set" (goal 3)
4. The old heuristic procedure remains in place. However, the old heuristic procedure can be switched off with a switch (goal 1)
5. There is already a switch to switch the old procedure on and off. If the semantics of this switch is defined in such a way that it only affects the old procedure there is no need for an additional switch (goal 1 and 3)
6. As both control options will now be offered to the user independently of each other, the labeling of the help texts will be selected in such a way that there can be no misunderstanding when using these options.

The advantage of this is:
With a minimum of effort this will fix the problem already now. There is no need to go through the "cans and worms" of the code. Factory reset will just work, because it is like it has always been. And from a default configuration this bug is much less likely than from a configuration that has grown over the years.
So the average user will never notice a difference. But the concerned user now has an option to get rid of the old behavior. Eventually the old logic can be removed but then yes, one needs to go through all those "cans and worms".

That's my proposal. If you don't like it, then please come up with a different proposal (clear goals and measures).

> Together we defined a per interface feature that works quite well from the looks of it, but you haven't exposed it to the user in all the cases in the right manner where it matters:

I think I have. But maybe not in the way you had expected it. The mistake was to set the anti lockout flag as default.

Cheers, Curly060 =;->

Monviech:
Isn't the feature that is tried to be added here, actually just the same as this:
- Go to "Firewall - Settings - Advanced" and check "Disable administration anti-lockout rule"
- Create your own firewall rules.

- Goal 1 is reached - The firewall won't open ports accidentally anymore.
- Goal 2 is reached - The user won't accidentally close ports anymore. They have opted in on managing their own firewall rules.
- Goal 3 is reached - Existing configurations will not change.

And all without adding any new code.

Curly060:
> Isn't the feature that is tried to be added here, actually just the same as this:

No, not at all, see below.

>
>     1. Go to "Firewall - Settings - Advanced" and check "Disable administration anti-lockout rule"
>     2. Create your own firewall rules.
> 
>     * Goal 1 is reached - The firewall won't open ports accidentally anymore.
>     * Goal 2 is reached - The user won't accidentally close ports anymore. They have opted in on managing their own firewall rules.

But I do not want to opt in to create my own firewall rules for this! I quite like the anti lockout feature, esp. when refactoring the rules. In fact, I would like to always have access on LAN, no matter how badly I mess up my firewall rules. I just don't want the system to guess and eventually guess wrong.

That's why I like the new interface based approach very much:
No need to disable the anti lockout system and I can define exactly on which interfaces it should be created for. An no accidental opening/closing ports on wrong interfaces.

Cheers, Curly060 =;->

Curly060:
There is a huge difference between:
- I need to take care to never lock myself out
- The system takes care to never lock me out

The whole point of the anti lockout feature is that it works **despite** of any user defined rules! It has to be able to always override whatever the user does. And it currently does, except in a few rare cases.

And these rare cases come from the fact that the current system is trying to "guess" the interface to work on, based on how they are named inside the config. And sometimes it "guesses" wrong.

The proposal to set a checkbox per interface removes the need to guess. But if the "guess" way shall be fully replaced by the new way, then yes indeed, that's not so easy. Hence my latest proposal.

AdSchellevis:
The problem is that in terms of actions for the user it’s roughly the same as @Monviech mentioned, the outcome in terms of rules is the same. 

I think I already mentioned this a long time ago, but we’re really trying to over-engineer a relatively small feature here, where in reality this is all about visibility. (Something we may be able to further improve in the future, but not now)

AdSchellevis:
We have had lengthy internal discussions about this topic, but currently it's just not worth our time and effort. In order to improve the feature, more work is needed for very little gain as people can just disable the anti-lockout anyway.  All of these auto-generated firewall rules have downsides unfortunately.

PR #7398, opened 2024-04-19
Misspelling of address in wireguard peer generator

ryanwinter:
"Address" was incorrectly spelt as "adress"


fichtner:
Merged, thanks!

PR #7420, opened 2024-04-30
wireguard: update dialogEditWireguardClient.xml for minor spelling mistake

ryanwinter:


src/opnsense/mvc/app/controllers/OPNsense/Wireguard/forms/dialogEditWireguardClient.xml
```diff
@@ -29,7 +29,7 @@
         <style>tokenize</style>
         <type>select_multiple</type>
         <allownew>true</allownew>
-        <help>List of networks allowed to pass trough the tunnel adapter. Use CIDR notation like 10.0.0.0/24.</help>
+        <help>List of networks allowed to pass through the tunnel adapter. Use CIDR notation like 10.0.0.0/24.</help>
```
fichtner, line 32: If we are here can we do this?

```suggestion
        <help>List of networks to route through the tunnel adapter. Use CIDR notation like 10.0.0.0/24.</help>
```

fichtner: (all we do is run `route add`)

AdSchellevis: it's pass and route when "Disable routes" is unchecked, otherwise it's only pass 

https://github.com/opnsense/core/blob/100aac2fb88c9ceeed83fe7355ff3b9755c585a4/src/opnsense/service/templates/OPNsense/Wireguard/wireguard-server.conf#L37

fichtner: Ok, my main motivation is to get rid of explaining „allowed“ terminology with „allowed“ in the description.

AdSchellevis: It's a bit confusing indeed, maybe better to only explain the type of expected input, something like:

> Use CIDR notation like 10.0.0.0/24, when routes are enabled one will be added for this network as well.

PR #7429, opened 2024-05-01
fix typo in message about "outboud nat rule"

lyricnz:


src/opnsense/mvc/app/models/OPNsense/Interfaces/Vip.php
```diff
@@ -42,7 +42,6 @@ public function performValidation($validateFullModel = false)
     {
         $messages = parent::performValidation($validateFullModel);
 
-        $unqiue_addrs = [];
```
fichtner, line 45: the var exists as `$unique_addrs` so it would make more sense to fix that typo instead of removing.

lyricnz: True! I didn't notice that. Will restore and fix.

fichtner:
Merged, thanks!

PR #7445, opened 2024-05-07
interfaces: tweak wording in interface gateway selection

fichtner:
@AdSchellevis what do you think about this wording tweak?

### DIFF
--- a/src/www/interfaces.php
+++ b/src/www/interfaces.php
@@ -2590,7 +2590,7 @@ include("head.inc");
                             </select>
                             <div class="hidden" data-for="help_for_gatewayv6">
                               <?= gettext('Select a gateway from the list to reply the incoming packets to the proper next hop on their way back. ' .
-                                          'This is typically disabled for LAN type connections.') ?>
+                                          'This is typically disabled for LAN type interfaces.') ?>
                             </div>
                           </td>
                         </tr>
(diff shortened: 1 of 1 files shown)


src/www/interfaces.php
```diff
@@ -2075,7 +2075,7 @@ function toggle_wirelesscfg() {
                             </select>
                             <div class="hidden" data-for="help_for_gateway">
                               <?= gettext('Select a gateway from the list to reply the incoming packets to the proper next hop on their way back and apply source NAT when configured. ' .
-                                          'Usually LAN type connections leave this to default.');?>
+                                          'This is typically disabled for LAN type connections.') ?>
```
fichtner, line 2078:
```suggestion
                                          'This is typically disabled for LAN type interfaces.') ?>
```

src/www/interfaces.php
```diff
@@ -2590,7 +2590,7 @@ function toggle_wirelesscfg() {
                             </select>
                             <div class="hidden" data-for="help_for_gatewayv6">
                               <?= gettext('Select a gateway from the list to reply the incoming packets to the proper next hop on their way back. ' .
-                                          'Usually LAN type connections leave this to default.');?>
+                                          'This is typically disabled for LAN type connections.') ?>
```
fichtner, line 2593:
```suggestion
                                          'This is typically disabled for LAN type interfaces.') ?>
```

PR #7455, opened 2024-05-12
mvc - larger overhaul replacing Phalcon Application #6389 

AdSchellevis:
(https://github.com/opnsense/core/issues/6389)

This removes most phalcon code currently being used on our end, except the Volt templates (which are re-wrapped) and a translation class we can easily replace later in a separate commit. Consumers of our controller classes shouldn't notice a difference as the used objects and methods are named the same.

The most notable changes are the following ones:

* Exceptions about not being able to find a requested path now break down into different exceptions inheriting from DispatchException, which makes it easier from the entrypoint (api.php, index.php) to catch and handle accordingly.
* When not in development mode, raw exceptions are not being returned anymore, which increases security
* The Dispatcher class is reponsible for object construction and mapping validation (valid uri, but no object found)
* The Router class replaces previous Application class, it disects offered uri's into namespaces, classnames and methods to call.

In the long run there should be a seperate controller for controllers using volt templates or api calls, but as the existing ones don't distinct between this and the output handling is different now, we can park this for a later moment in time (the performance penalty should be rather low).

Some unused functionality has been removed, for example support for the  X-HTTP-Method-Override header in Request->getMethod() (see https://github.com/phalcon/cphalcon/blob/44243c07658d060cd8a21761743b0f4fc01641aa/phalcon/Http/Request.zep#L599-L609).


src/opnsense/mvc/app/library/OPNsense/Mvc/Dispatcher.php
```diff
@@ -0,0 +1,178 @@
...
+        return $this->returnedValue;
+    }
+
+    /**
+     * XXX: remove call from ControllerBase, seems like a workaround for a specific phalcon issue back in 2015
+     * @return bool
+     */
+    public function wasForwarded(): bool
+    {
+        return false;
```
fichtner, line 176: hehe, funny :)

src/opnsense/mvc/app/library/OPNsense/Mvc/Request.php
```diff
@@ -0,0 +1,136 @@
...
+        $value = isset($_REQUEST[$name]) ? $_REQUEST[$name] : $defaultValue;
+        if ($filter !== null) {
+            $value = (new SanitizeFilter())->sanitize($value, $filter);
+        }
+        return $value;
+    }
+
+    public function getJsonRawBody(bool $associative = false): stdClass| array| bool
+    {
+        return json_decode($this->getRawBody(), $associative) ?? false;
```
fichtner, line 134: Feels like this can be simplified in the future. I've never seen us use JSON as an object and likely will not happen anytime soon.

src/opnsense/mvc/app/library/OPNsense/Mvc/Router.php
```diff
@@ -0,0 +1,181 @@
...
+            "action" => null,
+            "parameters" => []
+        ];
+        foreach ($defaults as $key => $val) {
+            $result[$key] = $val;
+        }
+
+        $count = 0;
+        foreach($pathElements as $element){
+            if($count == 0) {
```
fichtner, line 167: the foreach-if-count construct is a little over-engineered. maybe this could be flattened using `array_shift()` in the future as well

src/opnsense/mvc/app/library/OPNsense/Mvc/Security.php
```diff
@@ -0,0 +1,99 @@
...
+        $this->request = $request;
+    }
+
+    /**
+     * @param bool $new when not found, create new
+     * @return string|null token
+     */
+    public function getToken(bool $new=true): ?string
+    {
+        $token = $this->session->get('$PHALCON/CSRF$');
```
fichtner, line 62: Maybe want to make `PHALCON` a constant for later rename?

src/opnsense/www/api.php
```diff
@@ -1,56 +1,37 @@
...
-    if ($unbuffered) {
-        /**
-         * if buffer has been deleted already, we must assume the implementation
-         * has already sent headers and we cannot send them again.
-         */
-        return;
+function error_output($http_code, $e,  $user_message)
+{
+    $response = [];
+    if (OPNsense\Core\Config::getInstance()->object()->system->deployment != 'development'){
```
fichtner, line 6: Checking existence of `/var/run/development` is quicker (and a workaround for pre-config-readability in legacy parts)

src/opnsense/www/api.php
```diff
@@ -1,56 +1,37 @@
...
         $response['errorTitle'] = $e->getTitle();
-    } else {
-        $response['errorTitle'] = gettext('An API exception occurred');
-        $response['errorMessage'] = $e->getFile() . ':' . $e->getLine() . ': ' . $response['errorMessage'];
     }
-
-    header('HTTP', true, 500);
+    header('HTTP', true, $http_code);
     header("Content-Type: application/json;charset=utf-8");
+    echo htmlspecialchars(json_encode($response), ENT_NOQUOTES);
```
fichtner, line 17: was this unquoted before? oO

src/opnsense/www/api.php
```diff
@@ -1,56 +1,37 @@
...
+
 
-    echo json_encode($response, JSON_UNESCAPED_SLASHES);
+try {
+    $config = include __DIR__ . "/../mvc/app/config/config.php";
+    include __DIR__ . "/../mvc/app/config/loader.php";
+
+    $router = new OPNsense\Mvc\Router('/api/', 'Api');
+    $response = $router->routeRequest($_SERVER['REQUEST_URI']);
+    if (!$response->isSent()) {
```
fichtner, line 27: Feels like sending a rocket to the moon. Who could send() before? Mid-term maybe just preventing send after send() is called and not checking isSent() seems more straightforward.

src/opnsense/www/index.php
```diff
@@ -58,27 +58,23 @@ function view_html_safe($text)
...
-    $application = new \Phalcon\Mvc\Application($di);
+    $router = new OPNsense\Mvc\Router('/ui/');
+    try {
+        $response = $router->routeRequest($_SERVER['REQUEST_URI'], [
+            'controller' => 'indexController',
+            'action' => 'indexAction'
+        ]);
+    } catch (\OPNsense\Mvc\Exceptions\DispatchException) {
+        // unroutable (page not found), present page not found controller
+        $response = $router->routeRequest('/ui/core/index/index');
```
fichtner, line 72: `/ui/core/index/index` could be misplaced here... looks too specific and hardcoded for this level

fichtner (approved):
I like it very much, great work! left some notes for future changes perhaps.

src/opnsense/mvc/app/library/OPNsense/Mvc/Request.php
AdSchellevis: only caller seems to be ApiControllerBase, I'll remove the parameter as it's highly unlikely anyone else uses this (a grep on plugins shows nothing)

src/opnsense/mvc/app/library/OPNsense/Mvc/Router.php
AdSchellevis: some style issues in that block which are better fixed now, I'll fix those, use the `$idx`  from the array itself and be done with it :)

src/opnsense/mvc/app/library/OPNsense/Mvc/Security.php
AdSchellevis: at some point I rather rename this, but to lower risk, this isn't the time yet.

src/opnsense/www/api.php
AdSchellevis: thanks, I'll change that one

src/opnsense/www/api.php
AdSchellevis: Unfortunately there are more quirks in phalcon, in practice the controller already has called `send()`, which means this shouldn't be needed, but to keep backwards compatibility I've left the old behavior here (send if nobody did)

fichtner:
Merged, thanks! 🥳

kriansa:
@fichtner I understand this is probably not the best place to ask this, so I apologize upfront -- but why is Phalcon being removed? I'm not really aware of what's happening in PHP community but I'm curious to know why it's being considered frowned upon.

AdSchellevis:
@kriansa lowering the dependency chain and ease of maintenance in the long run.  For us Phalcon doesn't add much value and over the years we spend quite some time handling quirks which would have been easier to solve without the framework in between. We will keep some of it (the volt templates), due to lack of choices in that area.

fichtner:
@kriansa fair question, don't worry. I think at some point we might even break out volt into its own module so we are not bogged down waiting for Phalcon to support a newer PHP version in the long-term https://github.com/phalcon/volt

kriansa:
Thanks guys, I appreciate the transparency :+1:

PR #7460, opened 2024-05-16
src: resolve deprecation warnings for sre_constants

MaxXor:
On boot with python 3.11 a new deprecation warning appears:

```
configd.py	[e3c1a4e0-1dec-4667-a300-0ceff7926539] Script action stderr returned "b"/usr/local/opnsense/scripts/syslog/queryLog.py:36: DeprecationWarning: module 'sre_constants' is deprecated\n import sre_constants""
```

It's only used to have an exception class `sre_constants.error` it can be safely replaced with `re.error` since it is also imported from regex compiler there.

### DIFF
--- a/src/opnsense/scripts/suricata/queryAlertLog.py
+++ b/src/opnsense/scripts/suricata/queryAlertLog.py
@@ -33,7 +33,6 @@
 import sys
 import os.path
 import re
-import sre_constants
 import shlex
 import ujson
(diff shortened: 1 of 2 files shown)


AdSchellevis:
@MaxXor thanks!

John-AZ:
Thanks.
I had this error too and it was fixed by changing the code listed https://github.com/opnsense/core/commit/e21159a973

PR #7461, opened 2024-05-17
dashboard: add page reload on successful save

Monviech:
When trying out the new dashboard, the save button didn't give me any feedback on successful save. So I have added a page reload.

### DIFF
--- a/src/opnsense/www/css/dashboard.css
+++ b/src/opnsense/www/css/dashboard.css
@@ -2,6 +2,43 @@
     font-size: 2em;
 }
 
+#save-grid {
+    width: inherit;
+    height: inherit;
+    display: inline-flex;
+    align-items: center;
+    justify-content: center;
+    transition: opacity 0.3s ease;
+    position: relative;
+    width: 50px;
+    height: 30px;
+    margin-right: 10px;
+}
+
+#save-btn-text, #icon-container {
+    position: absolute;
+}
+
+#icon-container {
+    display: inline-flex;
+    align-items: center;
+    justify-content: center;
+}
+
+#save-spinner, #save-check {
+    transition: opacity 0.3s ease, transform 0.3s ease;
+}
+
+.hide {
+    opacity: 0;
+    transform: scale(0);
+}
+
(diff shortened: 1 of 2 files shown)


swhite2:
While this would certainly work with the least amount of code, the intent of the original design is to keep everything as dynamic as possible. This action will interrupt the keep-alive sessions collecting data.

I think this is fine for now, though perhaps there is a less invasive way we can think of to let the user know the action was succesful. Any opinion? :)

Monviech:
Hello, we could also display a success message in a HTML message area, or make the save button vanish upon successful save, e.g. only reloading the area with the buttons?

I just replicated the behavior of the Restore Defaults button with the full reload of the whole page.

swhite2:
Yes, the restore defaults was the only exception here, although in theory only the widget positions and width/height need to update with the current logic, this has a lower priority.

> we could also display a success message in a HTML message area, or make the save button vanish upon successful save, e.g. only reloading the area with the buttons?

If you can spare the time to propose something in this PR that'd be great - otherwise I'll add it to my todo list and merge this one for the time being.

Monviech:
Sure I can try.

Monviech:
@swhite2 What do you think of this idea? I think it fits to the general responsiveness of the new dashboard. Though, if it's not what you want, I'm open to change things, or pass you the torch on this. ^^

Monviech:
Thanks for the feedback. Yeah that makes sense. 

I cant continue here until next week. If you need to work on the file I suggest you merge or close the PR, and we can fix the action and timeout sequence later.

swhite2:
I'll change it in this PR in a bit and merge it after. Many thanks for the initial code.

FWIW, the dashboard is under active development so if you have any other feedback or changes, all welcome :)

swhite2:
@Monviech Got a bit carried away with the styling part here.. sorry about that ;)

Monviech:
@swhite2 Hey it looks interesting and like you had fun. :D Im going to try it out on the weekend. Thanks ^^

PR #7500, opened 2024-05-31
mvc: $timestamp is undefined in getRevisionContext()

fichtner:
As spotted by coverity... we could also ditch the condition but it looked like we could start passing the timestamp around, maybe even omitting the need for empty check since it's only called once?


fichtner:
The offending line was this one:

https://github.com/opnsense/core/blob/6a541aff16cfce74882b702580eff832bb1e2159/src/opnsense/mvc/app/library/OPNsense/Core/Config.php#L489

AdSchellevis:
@fichtner I think I rather ditch the parameter in this case. I'll do that later and close this PR, thanks!

fichtner:
Ok 👍

PR #7501, opened 2024-05-31
system: restrict PHP further

fichtner:
Dipping the toe in the php.ini best practices...

* max execution time is recommended as 300 at most via coverity
* We don't handle multiple `$_FILES` except maybe in external backup providers who could create multiple type=file fields.
* having fopen() not read URLs is probably sane enough


src/etc/rc.subr.d/recover
```diff
@@ -199,6 +200,7 @@ variables_order = "GPCS"
 file_uploads = On
 upload_tmp_dir = /tmp
 upload_max_filesize = 200M
+max_file_uploads = 1
```
AdSchellevis, line 203: for some plugins this might be challenging, in our code this shouldn't matter.

fichtner: I couldn't find a `$_FILES` use in plugins. The default is 20, assuming 200 MB per file I'd say more than 5 is already pretty exhaustive

swhite2:
can't see any weirdness that would be restricted by this, LGTM

src/etc/rc.subr.d/recover
fichtner: @muratbalaban43 could you check this? Are you using file uploads extensively? Maybe in Phalcon using `getUploadedFiles()`

fichtner, line 203:
```suggestion
max_file_uploads = 5
```

src/opnsense/service/templates/OPNsense/WebGui/php.ini
```diff
@@ -14,6 +15,7 @@ variables_order = "GPCS"
 file_uploads = On
 upload_tmp_dir = /tmp
 upload_max_filesize = 200M
+max_file_uploads = 1
```
fichtner, line 18:
```suggestion
max_file_uploads = 5
```

src/etc/rc.subr.d/recover
fichtner: @muratbalaban43 going with 5 for now, probably not before 24.7

muratbalaban43: @fichtner thanks for the heads-up. looking into this now.

Issue #7189, opened 2024-01-31
KEA-DHCP needs network-booting-options like ISC-DHCP


mayerthomas:
**Important notices**

- [x] I have read the contributing guide lines at https://github.com/opnsense/core/blob/master/CONTRIBUTING.md
- [X] I am convinced that my issue is new after having checked both open and closed issues at https://github.com/opnsense/core/issues?q=is%3Aissue

**Description of my problem**

In order to PXE-boot my clients to my FOG-cloning-server, I had a successful setup with ISC-DHCP:
- Enabled Network Booting
- Set next-server IP
- Set default bios filename

This setup only worked for me with the ISC-DHCP-Network-Booting-Settings, but never with the ISC-DHCP-TFTP-Server-Settings.
At the moment, KEA-DHCP in OPNSense only ships with TFTP-Server-Settings, which do not work for me.

**Possible Solution**

Is there a possibility to add Network-Booting-Settings to KEA-DHCP?

ndom91:
You should be able to edit the `/usr/local/etc/kea/kea-dhcp4.conf` file to add these kinds of options while they're not exposed in the UI.

mayerthomas:
Thanks for the tip! In terms of user-friendliness for all OPNsense-users, I would prefer appropriate setting-options in the GUI.

Darkfella91:
I tried adding those options manually in the config but after reload of Kea DHCP service they get removed.

h3krn:
Maybe I'm getting the question wrong, but the's a "TFTP server" and "TFTP bootfile name" entry box in the "Edit Subnet" dialog.
Isnt' that what you are looking for?
 
 "Kea DHCPv4" -> "Subnets" -> <edit range> -> "DHCP option data"

mayerthomas:
I need the setting "next-server" like in the former ISC-DHCP-Server under "Network Booting" > "Set next-server IP"

h3krn:
Thnx for explaining @mayerthomas , I now realize I need this as well. (same use case being PXE booting from FOG).
Last night I was assuming I just had broken hardware. But looking up the next-server parameter and comparing to my old dhcpd config this fixes my FOG setup as well.

I'm not a programmer, but I took a swing ad adding the parameter myself and created a PR hoping it can be included.

bcatubig:
> You should be able to edit the `/usr/local/etc/kea/kea-dhcp4.conf` file to add these kinds of options while they're not yet exposed in the UI.

[...]

PR #7505, opened 2024-06-01
Add next-server option to Kea

h3krn:
This PR aims to add the ``next-server`` (a.k.a. ``siaddr``) option to support PXE network booting with Kea
Fixes: #7189

### DIFF
--- a/src/opnsense/mvc/app/controllers/OPNsense/Kea/forms/dialogSubnet4.xml
+++ b/src/opnsense/mvc/app/controllers/OPNsense/Kea/forms/dialogSubnet4.xml
@@ -78,7 +78,7 @@
         <id>subnet4.next_server</id>
         <label>Next server</label>
         <type>text</type>
-        <help>Next server address or fqdn</help>
+        <help>Next server IP address</help>
     </field>
     <field>
         <id>subnet4.option_data.tftp_server_name</id>
--- a/src/opnsense/mvc/app/controllers/OPNsense/Kea/forms/dialogSubnet4.xml
+++ b/src/opnsense/mvc/app/controllers/OPNsense/Kea/forms/dialogSubnet4.xml
@@ -75,6 +75,12 @@
         <help>Specifies a list of RFC 868 time servers available to the client.</help>
     </field>
     <field>
+        <id>subnet4.next_server</id>
+        <label>Next server</label>
+        <type>text</type>
+        <help>Next server address or fqdn</help>
+    </field>
+    <field>
         <id>subnet4.option_data.tftp_server_name</id>
         <label>TFTP server</label>
         <type>text</type>
--- a/src/opnsense/mvc/app/models/OPNsense/Kea/KeaDhcpv4.php
+++ b/src/opnsense/mvc/app/models/OPNsense/Kea/KeaDhcpv4.php
@@ -141,6 +141,7 @@ class KeaDhcpv4 extends BaseModel
             $record = [
                 'id' => $subnet_id++,
                 'subnet' => (string)$subnet->subnet,
+                'next-server' => (string)$subnet->next_server,
                 'option-data' => [],
                 'pools' => [],
                 'reservations' => []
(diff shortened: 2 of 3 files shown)


src/opnsense/mvc/app/models/OPNsense/Kea/KeaDhcpv4.xml
```diff
@@ -36,6 +36,9 @@
                     <AddressFamily>ipv4</AddressFamily>
                     <Required>Y</Required>
                 </subnet>
+                <next_server type="TextField">
```
AdSchellevis: move inside `option_data` section.  When next server is a hostname (or address), the `HostnameField` is likely a better type or when only addresses are allowed `NetworkField`.

src/opnsense/mvc/app/models/OPNsense/Kea/KeaDhcpv4.php
```diff
@@ -141,6 +141,7 @@ private function getConfigSubnets()
             $record = [
                 'id' => $subnet_id++,
                 'subnet' => (string)$subnet->subnet,
+                'next-server' => (string)$subnet->next_server,
```
AdSchellevis, line 144: when inside  `option_data` , you can omit this.

src/opnsense/mvc/app/controllers/OPNsense/Kea/forms/dialogSubnet4.xml
```diff
@@ -74,6 +74,12 @@
         <allownew>true</allownew>
         <help>Specifies a list of RFC 868 time servers available to the client.</help>
     </field>
+    <field>
+        <id>subnet4.next_server</id>
```
AdSchellevis, line 78: see  `option_data`  comment in the model.

src/opnsense/mvc/app/controllers/OPNsense/Kea/forms/dialogSubnet4.xml
```diff
@@ -74,6 +74,12 @@
...
+        <label>Next server</label>
+        <type>text</type>
+        <help>Next server address or fqdn</help>
```
AdSchellevis, line 81: are we sure an fqdn is valid here? https://kea.readthedocs.io/en/kea-2.2.0/arm/dhcp4-srv.html#setting-fixed-fields-in-classification suggests an address.

AdSchellevis (changes requested):

src/opnsense/mvc/app/models/OPNsense/Kea/KeaDhcpv4.xml
h3krn: I've just updated the type to be a NetworkField.
As for the "option_data". Kea does not seem to accept the ``next_server`` field ?
```
 ERROR [kea-dhcp4.dhcp4.0x833e8f000] DHCP4_PARSER_FAIL failed to create or run parser for configuration element subnet4: definition for the option 'dhcp4.next-server' does not exist (/usr/local/etc/kea/kea-dhcp4.conf:142:33)
 ```

AdSchellevis: when inside the option_data block the underscores should be replaced automatically https://github.com/opnsense/core/blob/1e948bfc2f7c69d064d19af5e4e376b143443baa/src/opnsense/mvc/app/models/OPNsense/Kea/KeaDhcpv4.php#L150

AdSchellevis: @h3krn totally my mistake, I was expecting `next-server` to be in the option data block. So the current state of your code looks good.

h3krn: No worries. I was trying to figure out for myself why Kea need to have it this way as the exmaples were not 100% clear. 

But I think I figured it out. The ``next-server`` config does not have a dhcp option number. (Like ``tftp-server-name`` which is dhcp option 66.) It is also absent in https://github.com/isc-projects/kea/blob/master/doc/examples/kea4/all-options.json

 Instead it's defined as ``siaddr``.  If finally found an example here where it's also defined outside of the ``option-data`` block: https://github.com/isc-projects/kea/blob/master/doc/examples/kea4/all-keys.json#L573

AdSchellevis: @h3krn it's easy to get lost in the parameters, this looks good, I've merged it to master. Thanks!

PR #7507, opened 2024-06-03
system: do not create an interface route without an address

fichtner:
Similar to what default route handling already does.  In this case it's not logged as an error, because the condition is likely normal.


fichtner:
Will push after test.

PR #7511, opened 2024-06-05
Update main.scss iconspacer width

insilen:
Solve the problem of misaligned display of Chinese sidebar

**Before repair**
image

**Repair effect**
image

### DIFF
--- a/src/opnsense/www/themes/opnsense/assets/stylesheets/main.scss
+++ b/src/opnsense/www/themes/opnsense/assets/stylesheets/main.scss
@@ -403,7 +403,7 @@ main.page-content.col-lg-12 {
             font-size: 20px;
           }
           &.__iconspacer {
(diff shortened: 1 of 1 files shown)


fichtner:
Thanks for the report. I think we can solve this without the CSS, but it requires a little bit of tinkering... brb :)

insilen:
> @insilen can you try [0e54376](https://github.com/opnsense/core/commit/0e543767d5ee899eb2d3e5adcb6a51aacbbeea1c) ?
> 
> ```
> # opnsense-patch 0e543767d
> ```
> 
> Cheers, Franco

So Cool, it works!!!

Truly impressive!

fichtner:
Nice, the report is highly appreciated. It also fixes languages where the words are too long so it truncates them with "..." to fit the menu.

PR #7512, opened 2024-06-05
unbound: discard-timeout

### LINKED ISSUE #7493: Unbound: allow discard-timeout to be configured
**Important notices**

In Unbound DNS, if RFC 8767 support is configured with defaults (1800ms), replies that arrive >1900ms (discard-timeout) do not update the cache. This severely impacts the usability of RFC 8767 which can be invaluable when dealing with slow resolvers.

I would like to be able to configure 'discard-timeout' on the UI (Advanced settings), probably within 'serve expired settings'

The help should guide the user to make this value higher than the client expired response timeout and warn that otherwise these late responses will not update the cache

An alternative would be to allow custom settings, However this was previously removed, and does not help/guide the user who's using serve-expired to get a valid configuration.

I tried manually editing some of the unbound configuration, but these of course get overwritten on service start....

**Additional context**

See initial discussion at https://forum.opnsense.org/index.php?topic=40738.0

Note the reference to being a bit larger than server-expired-client-timeout in the official docs at https://unbound.docs.nlnetlabs.nl/en/latest/manpages/unbound.conf.html

_discard-timeout: <msec>_
_The wait time in msec where recursion requests are dropped. This is to stop a large number of replies from accumulating. They receive no reply, the work item continues to recurse. It is nice to be a bit larger than [serve-expired-client-timeout](https://unbound.docs.nlnetlabs.nl/en/latest/manpages/unbound.conf.html#unbound-conf-serve-expired-client-timeout) if that is enabled. A value of 1900 msec is suggested. The value 0 disables it._

_Default: 1900_

I am planning to experiment with a value around 3000 (since there is a related timeout for all tcp queries against auth servers that defaults to 3000)

The biggest cause of slow replies in my case was 'far away' servers. I am using quad9 DoT, but some chinese servers in particular can on occasion take 2,3,4 seconds to reply. serve-expired can work well here, but I want to catch the responses that finally come back so that I keep the cache as fresh as possible

Note that this particular code path results in the request queue exceeded counter increasing. A little misleading, as really this is a 'requests dropped' counter. However that's a base unbound issue... This was how I originally noticed the issue, questioning why the counter was increasing when I was nowhere near queue size limits.

Finally if it helps I could consider trying to build a PR. I'm an experienced dev & have worked with open source, but never opnsense.

planetf1:
Adds support for 'discard-timeout' to Unbound configuration.

This can be useful when using the option to serve expired responses.

The new option is added just below 'Jostle Timeout' within the advanced panel. 
I did consider adding it within the serve-expired section, however it's not strictly as tightly related and has more general applicability (as per unbound docs). I also removed once sentence (for help) which made no sense from the unbound docs, and clarified the exact labels used for related settings in the UI.

I tested with a 'live' install to ensure the panel worked, and the value was then placed into /var/unbound/advanced.conf

Based off master

Fixes #7493

### DIFF
--- a/src/opnsense/mvc/app/controllers/OPNsense/Unbound/forms/advanced.xml
+++ b/src/opnsense/mvc/app/controllers/OPNsense/Unbound/forms/advanced.xml
@@ -106,6 +106,18 @@
         </help>
     </field>
     <field>
+        <id>unbound.advanced.discardtimeout</id>
+        <label>Discard Timeout</label>
+        <type>text</type>
+        <help>
+	    The wait time in msec where recursion requests are dropped. This is to stop a large number of replies
+            from accumulating.
+            If 'Serve Expired Responses' is enabled it should be set greater than 'Client Expired Response Timeout'.
+            Otherwise, a value of 1900 msec is suggested.
+            The value 0 disables it. Default 1900 msec.
+        </help>
+    </field>
+    <field>
         <id>unbound.advanced.privatedomain</id>
         <label>Private Domains</label>
         <type>select_multiple</type>
--- a/src/opnsense/mvc/app/controllers/OPNsense/Unbound/forms/advanced.xml
+++ b/src/opnsense/mvc/app/controllers/OPNsense/Unbound/forms/advanced.xml
@@ -111,10 +111,9 @@
         <type>text</type>
         <help>
 	    The wait time in msec where recursion requests are dropped. This is to stop a large number of replies
-            from accumulating.
-            If 'Serve Expired Responses' is enabled it should be set greater than 'Client Expired Response Timeout'.
-            Otherwise, a value of 1900 msec is suggested.
-            The value 0 disables it. Default 1900 msec.
+            from accumulating. If 'Serve Expired Responses' is enabled this field should be set greater than
+            'Client Expired Response Timeout', otherwise, these late responses will not update the cache.
+            The value 0 disables it. Default 1900. This setting may increase the "request queue exceeded" counter.
         </help>
     </field>
     <field>
--- a/src/opnsense/mvc/app/models/OPNsense/Unbound/Unbound.xml
+++ b/src/opnsense/mvc/app/models/OPNsense/Unbound/Unbound.xml
@@ -1,7 +1,7 @@
 <model>
     <mount>//OPNsense/unboundplus</mount>
     <description>Unbound configuration</description>
-    <version>1.0.10</version>
+    <version>1.0.9</version>
     <items>
         <general>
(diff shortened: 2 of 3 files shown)


src/opnsense/mvc/app/models/OPNsense/Unbound/Unbound.xml
```diff
@@ -1,7 +1,7 @@
 <model>
     <mount>//OPNsense/unboundplus</mount>
     <description>Unbound configuration</description>
-    <version>1.0.9</version>
+    <version>1.0.10</version>
```
fichtner, line 4: Don't need this unless we deal with a new default value, but the new value is not required.
```suggestion
    <version>1.0.9</version>
```

src/opnsense/mvc/app/controllers/OPNsense/Unbound/forms/advanced.xml
```diff
@@ -105,6 +105,18 @@
...
+    <field>
+        <id>unbound.advanced.discardtimeout</id>
+        <label>Discard Timeout</label>
+        <type>text</type>
+        <help>
+	    The wait time in msec where recursion requests are dropped. This is to stop a large number of replies
+            from accumulating.
+            If 'Serve Expired Responses' is enabled it should be set greater than 'Client Expired Response Timeout'.
+            Otherwise, a value of 1900 msec is suggested.
+            The value 0 disables it. Default 1900 msec.
```
fichtner, line 117: "0" could be problematic which may be indistinguishable from "" (the empty default), but I leave that for @swhite2 to check

planetf1: I know serve-expired-ttl (which we use) has a similar use of 0 as per the [unbound doc](https://nlnetlabs.nl/documentation/unbound/unbound.conf/)

I just checked - and if I leave the value blank in the opnsense UI, no value is written to the configuration for discard-timeout. Once I set a value, it is written. Finally, if I specifically put 0 there, that does get written.

So I think in all cases it matches the unbound docs ?

swhite2:
Small nitpick, but looks good!

swhite2:
Oh, one more thing: in the ticket you said

> The help should guide the user to make this value higher than the client expired response timeout and warn that otherwise these late responses will not update the cache

Perhaps this can be included in the help text as well

dinguz:
I would like to have this functionality. Is there any planned timeline for inclusion?

swhite2:
@dinguz I'll finish up the PR today, it'll probably be included in a near minor release

swhite2 (approved):

PR #7523, opened 2024-06-11
API: reverse getArp resolve parameter behavior

kpelzel:
Currently the `resolve` parameter for the [getArp](https://docs.opnsense.org/development/api/core/diagnostics.html#id5) API call performs the opposite behavior you'd expect. When the `resolve=yes` parameter is supplied `list_arp.py` _will not_ resolve hostnames, while `resolve=no` (the default) _will_ resolve them. 

If this behavior is intentional, disregard this PR.

### DIFF
--- a/src/opnsense/scripts/interfaces/list_arp.py
+++ b/src/opnsense/scripts/interfaces/list_arp.py
@@ -38,7 +38,7 @@ import watchers.dhcpd
 
 if __name__ == '__main__':
     # do we use reverse DNS lookup ?
-    arp_arg = '-an' if '-r' in sys.argv else '-a'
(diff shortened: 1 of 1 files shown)


AdSchellevis:
This does look a bit strange indeed, but may require changes in the endpoint as well. I’ll put it on the list of things to check before the 24.7 release

AdSchellevis:
Just for reference, the controller seems to pass the expected parameter, so it smells like a bug.

https://github.com/opnsense/core/blob/b68fe69aa09096dbbfaab35fb734a9e88897f945/src/opnsense/mvc/app/controllers/OPNsense/Diagnostics/Api/InterfaceController.php#L74-L78

Issue #7527, opened 2024-06-13
interfaces: DHCPv6 incoming requirements for random server origin port


antipiot:
**Important notices**

Before you add a new report, we ask you kindly to acknowledge the following:

- [X] I have read the contributing guide lines at https://github.com/opnsense/core/blob/master/CONTRIBUTING.md
- [X] I am convinced that my issue is new after having checked both open and closed issues at https://github.com/opnsense/core/issues?q=is%3Aissue

**Describe the bug**

My opnsense installation does not seems to use / apply the prefix delegated by my internet service provider router

**To Reproduce**

Setup IPV6 with DHCP on the router side: i have a/48 delegated
On Opnsense, enable DHCPv6 -  / Send IPv6 prefix hint / Prefix delegation size /57 (or whatever does the same) / Request only an IPv6 prefix
Check what is the delegated network using else:
 cat `ifctl -6 -l -p` or the system - route - status - and look for loopback addresses that match your ipv6 ip range.
- Nothing shows up.

**Expected behavior**

A network with the delegated subnet is applied / used by opnsense
The command and routes status show's and address that looks like 1:1:1:1000/52

**Describe alternatives you considered**

Tested on 2 similar installations (same provider / different bare metla hardware for opnsense) with opnsense v23 and v24 with the same behavior

**Relevant log files**

I've wiresharked the thing and it looks like the delagation is correctly sendt to opnsense:
![image](https://github.com/opnsense/core/assets/26579988/0f57a7d0-3fcb-4d28-bf20-0619ba2b054c)

**Other informations**
I've already deployed IPv6 with success with another internet provider.

**Environment**

opnsense bare metal 24.1.8

Thanks for any hint!

Regards

fichtner:
Hi @antipiot,

I see your point, but no configuration or system output we can trace this by. There's also the ability to set debug log level for dhcp6c but for some reason I expect this to be configuration related.


Cheers,
Franco

antipiot:
Hi @fichtner !

Thanks for your help. For sure it can be configuration related, but... where...

If i'm not mistaken the only needed configuration to get the delegation is on the interface settings. Thoses are my settings.
![image](https://github.com/opnsense/core/assets/26579988/de976816-83c1-4367-9575-6e8d954ef45d)

I've indeed played around here to try to find a sweet spot without luck. Is there anything elses i can try?

I will try to enable debug log level for DHCPv6

antipiot:
[...]

PR #7538, opened 2024-06-18
interfaces: change IPv6 requirements #7527

fichtner:
The origin of the link-local allow rule for DHCPv6 traffic is a bit weird and the rule itself is probably dysfunctional, see https://github.com/pfsense/pfsense/commit/dbcddabcdf7e -- It has never been edited again and remains the same in OPNsense and pfSense today. Typically server client traffic exchanges exclusively over port 546 and 547 so the original one may have been a typo.

Now as witnessed by #7527 the server port could be random, but should always come from link-local so we can merge both rules into one without causing much problems. Works fine locally on my network too.

### DIFF
--- a/src/etc/inc/filter.lib.inc
+++ b/src/etc/inc/filter.lib.inc
@@ -355,7 +355,7 @@ function filter_core_rules_system($fw, $defaults)
                 'to' => 'fe80::/10',
                 'to_port' => 546,
                 'from' => 'fe80::/10',
-                /* omit from_to as the server may use a different source port */
+                /* omit from_port as the server may use a different source port */
             ], $defaults['pass']);
             $dhcpv6_opts = [
                 'descr' => 'allow dhcpv6 client out ' . $intfinfo['descr'],
--- a/src/etc/inc/filter.lib.inc
+++ b/src/etc/inc/filter.lib.inc
@@ -347,29 +347,27 @@ function filter_core_rules_system($fw, $defaults)
     foreach ($fw->getInterfaceMapping() as $intf => $intfinfo) {
         // allow DHCPv6 client out, before adding bogons (sequence 1, bogons @ 5)
         if (isset($config['system']['ipv6allow']) && in_array($intfinfo['ipaddrv6'], array("slaac","dhcp6"))) {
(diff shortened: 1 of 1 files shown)


src/etc/inc/filter.lib.inc
```diff
@@ -347,29 +347,27 @@ function filter_core_rules_system($fw, $defaults)
...
-            $dhcpv6_opts = array(
+            $fw->registerFilterRule(1, [
                 'descr' => 'allow dhcpv6 client in ' . $intfinfo['descr'],
                 '#ref' => 'system_advanced_network.php#ipv6allow',
+                'interface' => $intf,
+                'protocol' => 'udp',
+                'to' => 'fe80::/10',
+                'to_port' => 546,
+                'from' => 'fe80::/10',
+                /* omit from_to as the server may use a different source port */
```
AdSchellevis, line 358: shouldn't `from_to` be `from_port` ?

fichtner: sure

fichtner, line 358:
```suggestion
                /* omit from_port as the server may use a different source port */
```

src/etc/inc/filter.lib.inc
```diff
@@ -347,29 +347,27 @@ function filter_core_rules_system($fw, $defaults)
...
-            $dhcpv6_opts = array(
+            $fw->registerFilterRule(1, [
                 'descr' => 'allow dhcpv6 client in ' . $intfinfo['descr'],
                 '#ref' => 'system_advanced_network.php#ipv6allow',
+                'interface' => $intf,
+                'protocol' => 'udp',
+                'to' => 'fe80::/10',
+                'to_port' => 546,
+                'from' => 'fe80::/10',
+                /* omit from_port as the server may use a different source port */
```
fichtner, line 358:
```suggestion
                'from' => 'fe80::/10',
                /* omit from_port as the server may use a different source port */
                'to' => 'fe80::/10',
                'to_port' => 546,
```

swhite2 (approved):

AdSchellevis (approved):

PR #7559, opened 2024-06-26
use zpool.cache when importing pools at boot

nt1:


fichtner:
Thanks. Ideally hardcoding cache files should be avoided. It just looks like a maintenance spot further down the road. As said I’ll take a look when I’m back.

nt1:
Minimally, I'll move those string constants to a better place for the sake of maintainability.  This would also be a good place for a link to openzfs zpool cache documentation.

src/etc/rc
```diff
@@ -151,9 +153,19 @@ while [ ${attempts} -lt 3 ]; do
 done
 
 if kldstat -qm zfs; then
+
```
fichtner, line 156:
```suggestion
```

fichtner:
To be frank I'd prefer something like

```
ZPOOL_IMPORT_PATH="/dir1:/dir2"
export ZPOOL_IMPORT_PATH
```

The docs say:

> ZPOOL_IMPORT_PATH The search path for devices or files  to use with the pool. This is a colon-separated list of directories in which zpool looks for device nodes and files. Similar to the -d option in zpool import.

But the fun part is nothing really says what the compatible directory default is nor does the code make an effort to bootstrap a default.


Cheers,
Franco

fichtner:
Ok from the old code "/dev" should suffice so if you can let me know if this works as well... e7d80522e3

fichtner:
Shipping 701dff45b2f can't hurt so I will.

Issue #7521, opened 2024-06-11
Kea DHCPv4 not saving 'DHCP option data' when creating new subnet (first save)


ETES-Stuttgart:
**Important notices**

Before you add a new report, we ask you kindly to acknowledge the following:

- [X] I have read the contributing guide lines at https://github.com/opnsense/core/blob/master/CONTRIBUTING.md
- [X] I am convinced that my issue is new after having checked both open and closed issues at https://github.com/opnsense/core/issues?q=is%3Aissue

**Describe the bug**

When creating a new Subnet in Kea DHCPv4 in Services -> Kea DHCP [new] -> Kea DHCPv4 -> Subnets, after clicking "Save", all my settings under "DHCP option data" vanish (this means all, except the first ones "Subnet" and "Pools").
The fields are empty after clicking the edit-button of the entry.

**To Reproduce**

1. Go to 'Kea DHCPv4 in Services -> Kea DHCP [new] -> Kea DHCPv4 -> Subnets'
2. Fill in some of the options in 'DHCP option data'
3. Click on 'Save'
4. Click on 'Edit' of the created Subnet
5. See no data in 'DHCP option data'
6. Fill in the options in "DHCP option data' again
7. Click on 'Save'
8. Click on 'Edit' of the created Subnet
9. The data is now saved

**Expected behavior**

The options in "'DHCP option data' should be saved the first time I click on 'Save'

**Screenshots**

Creating a new Subnet with some options in 'DHCP option data':
![kea01](https://github.com/opnsense/core/assets/87643186/94ca2e71-29d6-453b-b15f-eb7402525463)
After clicking on 'Save' and then editing the created Subnet again:
![kea02](https://github.com/opnsense/core/assets/87643186/840ec4c9-e314-4404-9536-09cc45bf45c2)

**Environment**

Software version used and hardware type if relevant, e.g.:

OPNsense 24.4_8 (amd64).
AMD EPYC 3101 4-Core Processor (4 cores, 4 threads)

--

Thank you in advance & kind regards
Patrick

Monviech:
Maybe it has to do with the ``option_data_autocollect``

Not all of the option fields have that style set to hide them when it is 1.

I'm not sure, but maybe these fields aren't supposed to be able to be used when that option is 1.

[...]

PR #7564, opened 2024-06-27
KEA: custom DNS for reservation #7310 - basic GUI and conf implementa…

Reiner030:
…tion.

First implementation try for https://github.com/opnsense/core/issues/7310

NOT FOR PRODUCTION USAGE yet.

Basic GUI / conf implementation works "as usual" ... with several "copied" bugs where I am wondering where QA/community checks were failing to find this basic misbehavior yet since months:
* if you copy entry similar to subnet definition with "Auto collect option data" unchecked
  the content of ALL option_data is GONE after save 
* if you check again "Auto collect option data" then you got still the option_data entries set previously
(this commit can be yet the only "stable" one)

I was wondering why in subnet4 a dedicated "data-option" array was setup.
Especially for "plain" CSV  export/import and also "array dump to json" it doesn't make much sense.
But maybe it's a workaround for the "not nested arrays implemented" code checks in which I trapped while trying to implement export/import CSV enhancments for reservations.

The import and exports handles modifies basic model templates so it must be used very carefully/checked against all other usages of this models. But I am not a PHP dev but more administrator with basic programming skills... ^^


7AXEL:
Check out this project: https://github.com/7AXEL/Web2APK

AdSchellevis:
> if you copy entry similar to subnet definition with "Auto collect option data" unchecked
the content of ALL option_data is GONE after save

https://github.com/opnsense/core/issues/7521

Reiner030:
Thanks for hint - wasn't checking last 3 weeks for updates of this problem.
Just tested and also works for reservations fine like for subnet4.

AdSchellevis:
timeout

PR #7570, opened 2024-06-27
OpenVPN, rename 'pin' to 'otp_token' similar as in token based authen…

Reiner030:
…tication used.


AdSchellevis:
@Reiner030 would you be so kind to stop spamming our GitHub?

PR #7588, opened 2024-07-01
firewall rules page accessibility fixes

jfayre:
Added aria-label attribute to buttons on firewall rules table. Also removes aria-hidden attribute from modal dialogs. 
Using aria-hidden makes dialogs invisible to screen readers.

### DIFF
--- a/src/www/firewall_rules.php
+++ b/src/www/firewall_rules.php
@@ -367,7 +367,6 @@ $( document ).ready(function() {
                     $("#iform").submit()
                 }
               }]
-        
     });
     } else {
       // delete selected
@@ -387,7 +386,6 @@ $( document ).ready(function() {
                     $("#iform").submit()
                 }
               }]
-
       });
     }
   });
--- a/src/www/firewall_rules.php
+++ b/src/www/firewall_rules.php
@@ -366,11 +366,7 @@ $( document ).ready(function() {
                     $("#action").val("del");
                     $("#iform").submit()
                 }
-              }],
-              // remove aria-hidden
-              onshown: function(dialogRef){
-                  dialogRef.getModal().removeAttr('aria-hidden');
-              }
+              }]
         
     });
     } else {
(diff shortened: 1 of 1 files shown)


src/www/firewall_rules.php
```diff
@@ -366,7 +366,12 @@ function filter_rule_item_alias_tooltip($alias_name)
                     $("#action").val("del");
                     $("#iform").submit()
                 }
-              }]
+              }],
+              // remove aria-hidden
+              onshown: function(dialogRef){
+                  dialogRef.getModal().removeAttr('aria-hidden');
```
AdSchellevis, line 372: I think we should remove this in the bootstrap dialog itself.

https://github.com/opnsense/core/blob/6355a3debd3052526f26c03c029de6de6ed84f2f/src/opnsense/www/js/bootstrap-dialog.js#L392

(By the way, as https://github.com/nakupanda/bootstrap3-dialog seems to be stale, we can update the source on our end.)

I think the original bootstrap 3 examples contained the `aria-hidden` at some time, which might explain why a lot of these forms contain them.

AdSchellevis (changes requested):
If you remove the `onshow` events, we can merge this as well

jfayre:
Awesome! I really wasn’t looking forward to updating all of these dialogs. The onshown events should be removed now.

src/www/firewall_rules.php
```diff
@@ -367,6 +367,7 @@ function filter_rule_item_alias_tooltip($alias_name)
                     $("#iform").submit()
                 }
               }]
+
```
AdSchellevis, line 370: can you remove the extra line break as well? thanks

src/www/firewall_rules.php
```diff
@@ -386,6 +387,7 @@ function filter_rule_item_alias_tooltip($alias_name)
...
```
AdSchellevis, line 390: same here

AdSchellevis:
@jfayre thanks!

PR #7589, opened 2024-07-01
MVC: remove aria-hidden from base dialogs

jfayre:
Aria-hidden attribute makes dialogs inaccessible for screen readers.


AdSchellevis:
@jfayre thanks!

PR #7594, opened 2024-07-03
Implement router forwarding for paths containing hyphens

ureyni:
- Implement router forwarding for paths containing hyphens
- Add getQuery() function to handle GET request parameters


fichtner:
I believe @AdSchellevis wanted to avoid this. Changing URLs to underscore should fix all of this easily.

AdSchellevis:
yes, we had only one of these by mistake, they don't follow the camelcase rules and worked by accident in earlier versions.

ureyni:
Ok @AdSchellevis . We can update our zenarmor codes. what is your say about getQuery function?

AdSchellevis:
@ureyni I'll take a look, don't expect issues with adding a `getQuery()` method which is roughly the same as before.

AdSchellevis:
@ureyni https://github.com/opnsense/core/commit/9d6d7f70031ab4e9ee3ab0a02700cd394a4053bb should do the trick for the getQuery part

PR #7599, opened 2024-07-04
firewall: accessibility fixes for NAT page

jfayre:
Add aria-label to buttons and links in table.
Added role="cell" in last column of table to prevent screen readers from reading action buttons when navigating to other table rows.

### DIFF
--- a/src/www/firewall_nat.php
+++ b/src/www/firewall_nat.php
@@ -339,21 +339,21 @@ $( document ).ready(function() {
                       <th><?=gettext("IP");?></th>
                       <th><?=gettext("Ports");?></th>
                       <th><?=gettext("Description");?></th>
-                      <th class="text-nowrap">
-                        <a href="firewall_nat_edit.php" class="btn btn-primary btn-xs" data-toggle="tooltip" title="<?= html_safe(gettext('Add')) ?>">
+                      <th role="cell" class="text-nowrap">
+                        <a href="firewall_nat_edit.php" class="btn btn-primary btn-xs" data-toggle="tooltip" title="<?= html_safe(gettext('Add')) ?>" aria-label="<?= html_safe(gettext('Add')) ?>"">
                           <i class="fa fa-plus fa-fw"></i>
                         </a>
 <?php if (count($a_nat)): ?>
-                        <button id="move_<?= count($a_nat) ?>" name="move_<?= count($a_nat) ?>_x" data-toggle="tooltip" title="<?=html_safe(gettext("Move selected rules to end"))?>" class="act_move btn btn-default btn-xs">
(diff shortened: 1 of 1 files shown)


src/www/firewall_nat.php
```diff
@@ -339,21 +339,21 @@ function delete_id($id, &$array)
                       <th><?=gettext("IP");?></th>
                       <th><?=gettext("Ports");?></th>
                       <th><?=gettext("Description");?></th>
-                      <th class="text-nowrap">
-                        <a href="firewall_nat_edit.php" class="btn btn-primary btn-xs" data-toggle="tooltip" title="<?= html_safe(gettext('Add')) ?>">
+                      <th role="cell" class="text-nowrap">
+                        <a href="firewall_nat_edit.php" class="btn btn-primary btn-xs" data-toggle="tooltip" title="<?= html_safe(gettext('Add')) ?>" aria-label="<?= html_safe(gettext('Add')) ?>"">
```
AdSchellevis, line 343:
```suggestion
                        <a href="firewall_nat_edit.php" class="btn btn-primary btn-xs" data-toggle="tooltip" title="<?= html_safe(gettext('Add')) ?>" aria-label="<?= html_safe(gettext('Add')) ?>">
```

AdSchellevis (changes requested):

AdSchellevis:
@jfayre 
While reading this PR, I'm wondering if we are trying to apply the right fix here.

Looking at `bootstrap.js` on our end, on rendering the title is moved to `data-original-title`:

https://github.com/opnsense/core/blob/93304298abf4dc48a47766b2fa30729489ec6db8/src/opnsense/www/js/bootstrap.js#L1784-L1789

Newer versions seem to add `aria-label` here too:

https://github.com/twbs/bootstrap/blob/7c392498fabed7b4327e7c798c6af152b5071d29/dist/js/bootstrap.js#L3554-L3564

Maybe we better backport this change on our end so all tooltips are equal and we don't need to copy titles. 

What do you think?

fichtner:
I would prefer the deduplication too. The risk for regressions is also a factor here changing all these lines. Things happen.

jfayre:
I completely agree if you can do that. There are a lot of these types of issues and it would be great if we could fix them all in one go.Sent from my iPhoneOn Jul 8, 2024, at 5:27 AM, Ad Schellevis ***@***.***> wrote:﻿
@jfayre
While reading this PR, I'm wondering if we are trying to apply the right fix here.
Looking at bootstrap.js on our end, on rendering the title is moved to data-original-title:
https://github.com/opnsense/core/blob/93304298abf4dc48a47766b2fa30729489ec6db8/src/opnsense/www/js/bootstrap.js#L1784-L1789
Newer versions seem to add aria-label here too:
https://github.com/twbs/bootstrap/blob/7c392498fabed7b4327e7c798c6af152b5071d29/dist/js/bootstrap.js#L3554-L3564
Maybe we better backport this change on our end so all tooltips are equal and we don't need to copy titles.
What do you think?

AdSchellevis:
I'll take a look

AdSchellevis:
https://github.com/opnsense/core/commit/b0cab3bdca1180826ffe230ccca7e56280529759 should do the trick

PR #7603, opened 2024-07-07
Added configured host name to the page title

brettski74:
When using password managers like KeePass, the page title is helpful in allowing the password manager to correctly identify the correct credentials to autofill the login form. By not including the host name in the page title, it is impossible for password managers to differentiate between multiple different OPNsense routers. This change adds the host name to the login page title in a format that is consistent with other pages throughout OPNsense and allows password managers to correctly identify the credentials required at login.


fichtner:
It’s also helpful to security breaches so sorry this is not acceptable.

PR #7640, opened 2024-07-22
Dashboard: New dashboard will not work with themes that use custom fonts, potential fix (not the best option)

Jackysi:
Through changing/designing the new template, I noticed fonts on new Dashboard Gauges do not work with the pre-set from the body. So, I modified the code to take the heading (H1, which is always present on the page) and body font for canvas rendering.  Works with 96% of browsers according to caniuse.com. 

This isn't the best solution, but would love to see this functionality in the new release. 

image


AdSchellevis:
@Jackysi looks good, thanks

PR #7643, opened 2024-07-24
template: motd: drop X/Twitter for Reddit

AntiEngineer:
It's quite controversial but things kind of change.

I recently pulled the 24.1 update, and noticed it was still mentioned as Twitter. 
Here's a patch to correctly refer to the product as of now, which is X.

### DIFF
--- a/src/opnsense/service/templates/OPNsense/Auth/motd.in
+++ b/src/opnsense/service/templates/OPNsense/Auth/motd.in
@@ -5,5 +5,5 @@
 | Handbook:	https://docs.opnsense.org/   |       ))))))))   ((((((((
 | Forums:	https://forum.opnsense.org/  |         @@@///   \\\@@@
 | Code:		https://github.com/opnsense  |        @@@@         @@@@
-| Twitter:	https://twitter.com/opnsense |         @@@@@@@@@@@@@@@
+| X:		https://x.com/opnsense       |         @@@@@@@@@@@@@@@
 ----------------------------------------------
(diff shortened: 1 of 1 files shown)


fichtner:
I'd rather drop this for Reddit (if it fits)

AntiEngineer:
What about now? :)

fichtner:
thanks for your effort, but you need to keep the alignment otherwise you break console rendering

fichtner:
(I mean not shifting everything further right)

AntiEngineer:
Well, in such case, what about if we name the Reddit handle instead of the full URL? I don't think someone would copy the URL from the console and then paste it in a browser. Something like "r/opnsense" or so on?

fichtner:
there should be enough space between the link and the label

AntiEngineer:
Screenshot 2024-07-24 at 3 21 57 PM

This is another feasible option, but to my personal opinion it looks bad without spacing at the right side. If you are good to go with it I'll do another commit.

AntiEngineer:
We can also get that space back if we downgrade the URL to http, but that will look weird considering the other URLs use HTTPS.

fichtner:
how about aa2fc872e3 ?!

AntiEngineer:
Well, that seems to be an option.
The thing is I believed you wouldn't want to remove the TAB/CONTROL
characters, else I'd have replaced them with spaces :P.

But yeah, your commit makes the trick and hence then, this pull request can
be closed.

fichtner:
yes, well, it served a lazy purpose when we didn't have to salvage these characters for URLs. I'll merge and close then. Appreciate the PR and discussion!

PR #7646, opened 2024-07-25
Add monitor IP to API results.

hb0nes:
While creating a Gateway Prometheus Exporter for OPNSense, I was interested in which IP was being monitored by dping, not just the name of the gateway. This PR exposes that information through the API.


AdSchellevis:
@hb0nes looks good, thanks

PR #7664, opened 2024-07-26
Remove dots from traffic chart

fuchsg:
Just a minor cosmetic suggestion to remove the dots from the traffic chart so they only show when hovering.


swhite2:
Sure: https://github.com/opnsense/core/commit/31380e4f6de758aebda45db5e66ff6a1e62849bc :)

PR #7667, opened 2024-07-27
Update suricata.yaml and remove two deprecated settings and add a few useful settings

j0nny55555:
Removes deprecated -http and -tls from types/alert, and adds useful default features for output

Example suricata.yaml config used for reference, it is version 7.x:
https://github.com/OISF/suricata/blob/main-7.0.x/suricata.yaml.in

I have successfully tested these (and more) settings, but I was using the /usr/local/etc/suricata/custom.yaml file and restarting the service to see it not error output about the deprecated settings. It is my understanding that the environment replaced the suricata.yaml setting with the custom.yaml setting.

The custom.yaml file does not work completely as expected btw, see here: https://forum.opnsense.org/index.php?topic=40955.msg200771

If that means I did not complete a successful test, please let me know, otherwise this should mean fewer errors and useful logs by default.

### DIFF
--- a/src/opnsense/service/templates/OPNsense/IDS/suricata.yaml
+++ b/src/opnsense/service/templates/OPNsense/IDS/suricata.yaml
@@ -337,6 +337,13 @@ outputs:
       identity: "suricata"
       facility: local5
       level: Info
+      community-id: true
+      community-id-seed: 0
+      xff:
+        enabled: yes
+        mode: extra-data
+        deployment: reverse
+        header: X-Forwarded-For
       types:
         - alert:
 {%     if not helpers.empty('OPNsense.IDS.general.LogPayload') %}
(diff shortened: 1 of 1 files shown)


AdSchellevis:
if the alert options are deprecated, I don't mind removing them, but sending non alert type events to the log by default is likely not a good idea as it might cause floods of unexpected data for others.

j0nny55555:
> if the alert options are deprecated, I don't mind removing them, but sending non alert type events to the log by default is likely not a good idea as it might cause floods of unexpected data for others.

understood, I've removed all non-alert related elements

My Suricata is configured to TCP Syslog the evelog syslog to my Logstash, these will add useful content to the individual alerts but would only align alerts to expected outputs especially with the "tagged" option, no event explosion.

AdSchellevis:
@j0nny55555 looks reasonable, thanks!

fichtner:
@AdSchellevis please be more carful as this went to stable/24.7 so I'm going to scrub it right now.

AdSchellevis:
@fichtner oops, totally my mistake indeed. it should go to master as usual.

fichtner:
d2ef070687d1 then :)

j0nny55555:
thank you!!

PR #7689, opened 2024-07-30
OpenVPN instances: Add float option to support roaming clients

chrko:
In the classic, legacy server variant there was an option called "dynamic_ip" which injected the options `persist-remote-ip` and `float` into the generated configuration file. In the modern variant `persist-remote-ip` is already available, but `float` isn't available anymore.

The OpenVPN `float` option is used to allow client to change IP addresses during an ongoing connection to allow roaming e.g. in mobile networks or changes in the client WiFi.

See https://github.com/opnsense/core/blob/d2ef070687d1a4077a818211db4ffd6a5eab69f6/src/etc/inc/plugins.inc.d/openvpn.inc#L872-L875

### DIFF
--- a/src/opnsense/mvc/app/models/OPNsense/OpenVPN/OpenVPN.xml
+++ b/src/opnsense/mvc/app/models/OPNsense/OpenVPN/OpenVPN.xml
@@ -277,9 +277,9 @@
                         <float>float</float>
                         <passtos>passtos</passtos>
                         <persist-remote-ip>persist-remote-ip</persist-remote-ip>
-                        <route-nopull>route-nopull</route-nopull>
-                        <route-noexec>route-noexec</route-noexec>
                         <remote-random>remote-random</remote-random>
+                        <route-noexec>route-noexec</route-noexec>
+                        <route-nopull>route-nopull</route-nopull>
                     </OptionValues>
                 </various_flags>
                 <various_push_flags type="OptionField">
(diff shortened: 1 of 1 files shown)


src/opnsense/mvc/app/models/OPNsense/OpenVPN/OpenVPN.xml
```diff
@@ -274,11 +274,12 @@
...
                         <client-to-client>client-to-client</client-to-client>
                         <duplicate-cn>duplicate-cn</duplicate-cn>
+                        <float>float</float>
                         <passtos>passtos</passtos>
                         <persist-remote-ip>persist-remote-ip</persist-remote-ip>
-                        <route-nopull>route-nopull</route-nopull>
-                        <route-noexec>route-noexec</route-noexec>
                         <remote-random>remote-random</remote-random>
+                        <route-noexec>route-noexec</route-noexec>
+                        <route-nopull>route-nopull</route-nopull>
```
AdSchellevis, line 282: why move these options around? adding `float` is fine, but let's try to keep the diff as small as possible unless there is a good reason to move things.

fichtner: sorting them alphabetically can help users when these lists grow beyond a certain size. certainly worth having (and consistent with out approach)

chrko: I agree with both comments. Despite the "sort" commit not matching the PR expectations, I've favored consistency afterwards over as small as possible as the diff itself is still small and less than a page.

fichtner: It's also two commits. Using rebase option here on GitHub should make that crystal clear in the code history.

AdSchellevis:
@chrko merged as discussed, thanks!

PR #7692, opened 2024-07-31
interfaces: require PPP interface to be in up state

### LINKED ISSUE #7691: default route gets lost when "Periodic interface reset" is run
**Important notices**

**Description**

I stumbled upon this problem recently. I can't really tell when it started, the issue existed in previous versions as well.

I have a cron job running, that executes "Periodic interface reset" for "wan" at 02:00 in the night.
Every morning I notice, that the internet connection is not working. Looking at the web gui, it is obvious that the default route is missing (system / routes / status). Apart of that, everything else seems to be fine - gateway is up and running.

I'm using a rather standard German DSL setup (pppoe for IPv4, DHCPv6 for IPv6)

**To Reproduce**

1. run a cronjob with "Periodic interface reset"
or
1. run "rc.configure_interface wan" in a shell

**Expected behavior**

Expected behavior is that internet connection is working again after command completes.

As a workaround it is possible to restart the "Gateway monitor watcher" - the default route will then be re-added and everything starts to work again.

**Screenshots**

There are no applicable screenshots.

**Relevant log files**

From what I could tell, there is nothing in the logs that reveals this issue. Can post logs, if needed.

**Additional context**

At the beginning, I thought this was related to me adding a second gateway using the third network port on the APU2. I added a Huawei LTE route as backup. But I disabled everything and the problem still persisted.

**Environment**

Running on a PCEngines APU2 ([apu2e4](https://pcengines.ch/apu2e4.htm))

opnsense version info:

```
Type: opnsense	
Version: 24.7_5	
Architecture: amd64	
Commit: 6578c5a18	
Mirror: https://pkg.opnsense.org/FreeBSD:14:amd64/24.7	
Repositories: OPNsense	
Updated on: Sun Jul 28 23:53:49 CEST 2024
```

**Analysis**

From what I can tell the "Periodic interface reset" cron action runs the same code as the script `rc.configure_interface`.

This script basically only runs a single function `configure_interface`:

https://github.com/opnsense/core/blob/4501c593ecd2ca9424ab1dc309875b5cc43641bf/src/etc/rc.configure_interface#L41

During the course of this function, there is a call to `interface_ppps_configure()`, which is executed, because I use pppoe:
https://github.com/opnsense/core/blob/4501c593ecd2ca9424ab1dc309875b5cc43641bf/src/etc/inc/interfaces.inc#L2340-L2353

Inside `interface_ppps_configure()` there is a loop, that is supposed to wait, until the interface shows up:
https://github.com/opnsense/core/blob/4501c593ecd2ca9424ab1dc309875b5cc43641bf/src/etc/inc/interfaces.inc#L1251-L1260

Here seems to be the culprit:
On my system, the shell return code for the first loop iteration is 0 and the output is as follow:

```
Array
(
    [0] => pppoe0: flags=8890<POINTOPOINT,NOARP,SIMPLEX,MULTICAST> metric 0 mtu 1500
    [1] =>      options=0
    [2] =>      nd6 options=23<PERFORMNUD,ACCEPT_RTADV,AUTO_LINKLOCAL>
)
```

Thus, the interface is there, but not UP, yet.

Back in `configure_interface`, there is a call to `system_routing_configure` that should set the default route:
https://github.com/opnsense/core/blob/4501c593ecd2ca9424ab1dc309875b5cc43641bf/src/etc/inc/interfaces.inc#L2414

For setting the default route, the gateways are fetched:
https://github.com/opnsense/core/blob/4501c593ecd2ca9424ab1dc309875b5cc43641bf/src/etc/inc/system.inc#L664

But, if there are no gateways, the function `system_default_route()` won't get called:
https://github.com/opnsense/core/blob/4501c593ecd2ca9424ab1dc309875b5cc43641bf/src/etc/inc/system.inc#L698-L708

Analyzing further, at the time of creation the object `OPNsense\Routing\Gateways` does not return a gateway for the pppoe interface which results in the default route not to be set, thus leaving my system without a default route.

Now, if I add just some seconds more wait time for the loop that waits for the interface to be visible using ifconfig, the output changes to this:

```
Array
(
    [0] => pppoe0: flags=10088d1<UP,POINTOPOINT,RUNNING,NOARP,SIMPLEX,MULTICAST,LOWER_UP> metric 0 mtu 1492
    [1] =>      options=0
    [2] =>      inet6 fe80::20d:b9ff:fe4f:b17c%pppoe0 prefixlen 64 tentative scopeid 0x8
    [3] =>      inet6 2001:9e8:4040:ddec:20d:b9ff:fe4f:b17c prefixlen 64 tentative autoconf pltime 172800 vltime 259200
    [4] =>      nd6 options=23<PERFORMNUD,ACCEPT_RTADV,AUTO_LINKLOCAL>
)
```

It seems that it just takes some time for the interface to be UP and have a gateway. Before, the call to `getDefaultGW()` does not return anything (i.e. the `gateway` entry in the return array is missing).

ifconfig returns a 0 on the command line even if the interface is not up - which is to be expected. What works for me is adding the `-u` flag to ifconfig, thus only listing UP interfaces and comparing the exec output to make sure that the interface is actually listed.

I'll link a PR in a few minutes.

nscheer:
Adds `-u` flag to ifconfig check to ensure only interfaces that are UP are reported. Furthermore adds a check for the output of ifconfig to ensure that the interface in question in fact is listed.

Otherwise the check will "approve" an interface that is not yet UP, and subsequent calls to fetch the gateway of that interface might fail if issued to quickly thereafter.

Resolves #7691

### DIFF
--- a/src/etc/inc/interfaces.inc
+++ b/src/etc/inc/interfaces.inc
@@ -1239,6 +1239,9 @@ EOD;
         @unlink('/var/spool/lock/LCK..' . basename($port));
     }
 
+    /* precaution for post-start 'up' check */
+    legacy_interface_flags($interface, 'down');
+
     /* fire up mpd */
     mwexecf(
         '/usr/local/sbin/mpd5 -b -d /var/etc -f %s -p %s -s ppp %s',
--- a/src/etc/inc/interfaces.inc
+++ b/src/etc/inc/interfaces.inc
@@ -1250,7 +1250,6 @@ EOD;
 
     /* wait for functional device */
     $i = 0;
-    $interface_is_up = false;
     while ($i < 30) {
         if (does_interface_exist($ppp['if'], 'up')) {
             $interface_is_up = true;
--- a/src/etc/inc/interfaces.inc
+++ b/src/etc/inc/interfaces.inc
@@ -1251,10 +1251,10 @@ EOD;
     /* wait for functional device */
     $i = 0;
     while ($i < 30) {
+        sleep(1);
         if (does_interface_exist($ppp['if'], 'up')) {
             break;
         }
-        sleep(1);
         $i++;
     }
 
--- a/src/etc/inc/interfaces.inc
+++ b/src/etc/inc/interfaces.inc
@@ -1252,7 +1252,6 @@ EOD;
     $i = 0;
     while ($i < 30) {
         if (does_interface_exist($ppp['if'], 'up')) {
-            $interface_is_up = true;
             break;
         }
         sleep(1);
--- a/src/etc/inc/interfaces.inc
+++ b/src/etc/inc/interfaces.inc
@@ -1259,7 +1259,7 @@ EOD;
         $i++;
     }
 
-    if ($i >= 30) {
+    if ($i >= $max) {
         log_msg("interface_ppps_configure() waiting threshold exceeded - device {$interface} is still not up", LOG_WARNING);
     }
 
--- a/src/etc/inc/interfaces.inc
+++ b/src/etc/inc/interfaces.inc
@@ -1261,7 +1261,7 @@ EOD;
     }
 
     if (!$interface_is_up) {
-        log_msg("Waiting threshold exceeded - interface $interface is still not UP in interface_ppps_configure().", LOG_WARNING);
+        log_msg("interface_ppps_configure() waiting threshold exceeded - device {$interface} is still not up", LOG_WARNING);
     }
 
     switch ($ppp['type']) {
--- a/src/etc/inc/interfaces.inc
+++ b/src/etc/inc/interfaces.inc
@@ -1248,15 +1248,20 @@ EOD;
     /* appease netgraph by setting the correct node name */
     shell_safe('/usr/sbin/daemon -f /usr/local/opnsense/scripts/interfaces/ppp-rename.sh %s %s %s', [$interface, $ifcfg['if'], $ppp['type']]);
 
-    /* wait for up to 10 seconds for the interface to appear (ppp(oe)) */
+    /* wait for up to 30 seconds for the interface to appear in UP state (ppp(oe)) */
     $i = 0;
-    while ($i < 10) {
-        exec("/sbin/ifconfig " . escapeshellarg($ppp['if']) . " 2>&1", $out, $ret);
-        if ($ret == 0) {
+    $interface_is_up = false;
+    while ($i < 30) {
+        if (does_interface_exist($ppp['if'], 'up')) {
+            $interface_is_up = true;
             break;
         }
         sleep(1);
         $i++;
+    }
+
+    if (!$interface_is_up) {
+        log_msg("Waiting threshold exceeded - interface $interface is still not UP in interface_ppps_configure().", LOG_WARNING);
     }
 
     switch ($ppp['type']) {
--- a/src/etc/inc/interfaces.inc
+++ b/src/etc/inc/interfaces.inc
@@ -1249,6 +1249,7 @@ EOD;
     shell_safe('/usr/sbin/daemon -f /usr/local/opnsense/scripts/interfaces/ppp-rename.sh %s %s %s', [$interface, $ifcfg['if'], $ppp['type']]);
 
     /* wait for functional device */
+    $max = 30;
     $i = 0;
     while ($i < 30) {
         sleep(1);
--- a/src/etc/inc/interfaces.inc
+++ b/src/etc/inc/interfaces.inc
@@ -1248,7 +1248,7 @@ EOD;
     /* appease netgraph by setting the correct node name */
     shell_safe('/usr/sbin/daemon -f /usr/local/opnsense/scripts/interfaces/ppp-rename.sh %s %s %s', [$interface, $ifcfg['if'], $ppp['type']]);
 
-    /* wait for up to 30 seconds for the interface to appear in UP state (ppp(oe)) */
+    /* wait for functional device */
     $i = 0;
     $interface_is_up = false;
     while ($i < 30) {
--- a/src/etc/inc/interfaces.inc
+++ b/src/etc/inc/interfaces.inc
@@ -1240,7 +1240,7 @@ EOD;
     }
 
     /* precaution for post-start 'up' check */
-    legacy_interface_flags($interface, 'down');
+    legacy_interface_flags($interface, 'down', false);
 
     /* fire up mpd */
     mwexecf(
--- a/src/etc/inc/interfaces.inc
+++ b/src/etc/inc/interfaces.inc
@@ -1258,7 +1258,7 @@ EOD;
         $i++;
     }
 
-    if (!$interface_is_up) {
+    if ($i >= 30) {
         log_msg("interface_ppps_configure() waiting threshold exceeded - device {$interface} is still not up", LOG_WARNING);
     }
 
--- a/src/etc/inc/interfaces.inc
+++ b/src/etc/inc/interfaces.inc
@@ -1251,7 +1251,7 @@ EOD;
     /* wait for functional device */
     $max = 30;
     $i = 0;
-    while ($i < 30) {
+    while ($i < $max) {
         sleep(1);
         if (does_interface_exist($ppp['if'], 'up')) {
             break;


fichtner:
@nscheer thanks, safe to say this code wasn't changed in a decade which is good and bad at the same time.changing this can easily break behaviour for others as well. I need to dig into the code history first.

nscheer:
@fichtner Thanks for looking into this!
Yes, there might be side effects. Question is if this ever should continue without an interface being UP.
And there still is the 10 second "timeout" when it will continue, no matter what.

Let me know, if I can help or test anything :)

fichtner:
Ok so first things first this was Seth in pfSense via https://github.com/pfsense/pfsense/commit/302d646e9c25 and unfortunately his work seldomly mentions tickets or problem scope. The behaviour changed very little except that they now wait around 30 seconds for this to complete which tells us how problematic this implementation is to this day... yes I know it just continues after that time anyway ;)

src/etc/inc/interfaces.inc
```diff
@@ -1248,11 +1248,12 @@ EOD;
...
 
-    /* wait for up to 10 seconds for the interface to appear (ppp(oe)) */
+    /* wait for up to 10 seconds for the interface to appear in UP state (ppp(oe)) */
     $i = 0;
     while ($i < 10) {
-        exec("/sbin/ifconfig " . escapeshellarg($ppp['if']) . " 2>&1", $out, $ret);
-        if ($ret == 0) {
+        $out = [];
+        exec("/sbin/ifconfig -u " . escapeshellarg($ppp['if']) . " 2>&1", $out, $ret);
+        if ($ret == 0 && strpos(implode("\n", $out), $ppp['if']) !== false) {
```
fichtner, line 1256: using aadc34d6a0b this can be:

```suggestion
        if (does_interface_exist($ppp['if'], 'up')) {
```

fichtner:
Modernizing the existence check out of the way I don't mind making this wait up to 30 seconds as well. The only thing that I think is worth changing is that when 30 seconds passed without the interface being up a log message should be emitted for traceability?

fichtner:
Just a thought: for further consistency we could consider bringing the interface down if it exists before starting mpd5 again?

fichtner:
@nscheer is it ok to offload this to you or do you want me to take this over? just because I'd like to get this into the code rather sooner than later.

nscheer:
> @nscheer is it ok to offload this to you or do you want me to take this over? just because I'd like to get this into the code rather sooner than later.

Yes, let me update my branch accordingly. Will get back to you, if something’s unclear.
Might be rather late today, tough :)

fichtner:
No problem. Let me know if I can help 👍

nscheer:
@fichtner I rebased on master and added the minimal changes I thought were needed.
Usually I'd like to mention the threshold value (i.e. 30 seconds) in the log message - but I'd rather introduce a constant for that than having the literals spread all over the place. Didn't know if there where any project specific style guides about that.
Let me know if I should adjust that.

nscheer:
Since we are talking about modernizing:
You just added the flag parameter to `does_interface_exist` and it has a default value.
Is it correct that this function is „newer“ than the legacy list function?
What do you think of adding an enum of interface states and require this in `does_interface_exist`? This way, we’d enforce correct use and have it self documenting. The enum can be backed by the string values the legacy function expects.

This would clean up the api a bit and we’d already use the enum for the first time.

fichtner:
1. We try to modifications in these parts of the code unless there is a fundamental technical issue that needs fixing. I just adopted the same style as `legacy_interface_listget()` as you can see this code is rarely touched over the years because risks of regressions are real.
2. There is probably very little benefit from sinking time here into a once-used API. Eventually when the MVC/API efforts shift to interface pages this whole code base is in need of a major overhaul anyway. But even then I doubt there is much need of checking if an interface exists and is up with this instance being the first in the code after many years without it.


Cheers,
Franco

nscheer:
Agreed, thanks for the clarification!

src/etc/inc/interfaces.inc
```diff
@@ -1248,17 +1248,22 @@ EOD;
...
+        if (does_interface_exist($ppp['if'], 'up')) {
+            $interface_is_up = true;
             break;
         }
         sleep(1);
         $i++;
     }
 
+    if (!$interface_is_up) {
+        log_msg("Waiting threshold exceeded - interface $interface is still not UP in interface_ppps_configure().", LOG_WARNING);
```
fichtner, line 1264:
```suggestion
        log_msg("interface_ppps_configure() waiting threshold exceeded - device {$interface} is still not up", LOG_WARNING);
```

src/etc/inc/interfaces.inc
```diff
@@ -1248,17 +1248,22 @@ EOD;
     /* appease netgraph by setting the correct node name */
     shell_safe('/usr/sbin/daemon -f /usr/local/opnsense/scripts/interfaces/ppp-rename.sh %s %s %s', [$interface, $ifcfg['if'], $ppp['type']]);
 
-    /* wait for up to 10 seconds for the interface to appear (ppp(oe)) */
+    /* wait for up to 30 seconds for the interface to appear in UP state (ppp(oe)) */
```
fichtner, line 1251:
```suggestion
    /* wait for functional device */
```

src/etc/inc/interfaces.inc
```diff
@@ -1248,17 +1248,22 @@ EOD;
     /* appease netgraph by setting the correct node name */
     shell_safe('/usr/sbin/daemon -f /usr/local/opnsense/scripts/interfaces/ppp-rename.sh %s %s %s', [$interface, $ifcfg['if'], $ppp['type']]);
 
-    /* wait for up to 10 seconds for the interface to appear (ppp(oe)) */
+    /* wait for functional device */
     $i = 0;
-    while ($i < 10) {
-        exec("/sbin/ifconfig " . escapeshellarg($ppp['if']) . " 2>&1", $out, $ret);
-        if ($ret == 0) {
+    $interface_is_up = false;
```
fichtner, line 1253:
```suggestion
```

src/etc/inc/interfaces.inc
```diff
@@ -1248,17 +1248,21 @@ EOD;
...
 
-    /* wait for up to 10 seconds for the interface to appear (ppp(oe)) */
+    /* wait for functional device */
     $i = 0;
-    while ($i < 10) {
-        exec("/sbin/ifconfig " . escapeshellarg($ppp['if']) . " 2>&1", $out, $ret);
-        if ($ret == 0) {
+    while ($i < 30) {
+        if (does_interface_exist($ppp['if'], 'up')) {
+            $interface_is_up = true;
```
fichtner, line 1255:
```suggestion
```

src/etc/inc/interfaces.inc
```diff
@@ -1248,17 +1248,20 @@ EOD;
...
-        if ($ret == 0) {
+    while ($i < 30) {
+        if (does_interface_exist($ppp['if'], 'up')) {
             break;
         }
         sleep(1);
         $i++;
     }
 
+    if (!$interface_is_up) {
```
fichtner, line 1261:
```suggestion
    if ($i >= 30) {
```

src/etc/inc/interfaces.inc
```diff
@@ -1248,17 +1248,20 @@ EOD;
...
+    /* wait for functional device */
     $i = 0;
-    while ($i < 10) {
-        exec("/sbin/ifconfig " . escapeshellarg($ppp['if']) . " 2>&1", $out, $ret);
-        if ($ret == 0) {
+    while ($i < 30) {
+        if (does_interface_exist($ppp['if'], 'up')) {
             break;
         }
         sleep(1);
```
fichtner, line 1257:
```suggestion
        sleep(1);
        if (does_interface_exist($ppp['if'], 'up')) {
            break;
        }
```

src/etc/inc/interfaces.inc
```diff
@@ -1248,17 +1248,20 @@ EOD;
     /* appease netgraph by setting the correct node name */
     shell_safe('/usr/sbin/daemon -f /usr/local/opnsense/scripts/interfaces/ppp-rename.sh %s %s %s', [$interface, $ifcfg['if'], $ppp['type']]);
 
-    /* wait for up to 10 seconds for the interface to appear (ppp(oe)) */
+    /* wait for functional device */
```
fichtner, line 1251:
```suggestion
    /* wait for functional device */
    $max = 30;
```

src/etc/inc/interfaces.inc
```diff
@@ -1248,17 +1248,21 @@ EOD;
...
+    $max = 30;
     $i = 0;
-    while ($i < 10) {
-        exec("/sbin/ifconfig " . escapeshellarg($ppp['if']) . " 2>&1", $out, $ret);
-        if ($ret == 0) {
+    while ($i < 30) {
```
fichtner, line 1254:
```suggestion
    while ($i < $max) {
```

src/etc/inc/interfaces.inc
```diff
@@ -1248,17 +1248,21 @@ EOD;
...
+    while ($i < $max) {
+        sleep(1);
+        if (does_interface_exist($ppp['if'], 'up')) {
             break;
         }
-        sleep(1);
         $i++;
     }
 
+    if ($i >= 30) {
```
fichtner, line 1262:
```suggestion
    if ($i >= $max) {
```

src/etc/inc/interfaces.inc
```diff
@@ -1239,6 +1239,9 @@ EOD;
         @unlink('/var/spool/lock/LCK..' . basename($port));
     }
 
+    /* precaution for post-start 'up' check */
+    legacy_interface_flags($interface, 'down');
```
fichtner, line 1243:
```suggestion
    legacy_interface_flags($interface, 'down', false);
```

fichtner:
@nscheer merged, thanks! I modified it a bit. Can you make sure it still works as intended?

fichtner:
@nscheer FYI https://github.com/opnsense/core/commit/7c9b97ab74c

nscheer:
@fichtner Nice! Yes, still works as intended!

fichtner:
Thanks, will be in 24.7.2. 😊

Issue #7669, opened 2024-07-27
Monit status widget does not update properly on new dashboard


mgittelman:
**Important notices**

Before you add a new report, we ask you kindly to acknowledge the following:

- [x ] I have read the contributing guide lines at https://github.com/opnsense/core/blob/master/CONTRIBUTING.md
- [ ] I am convinced that my issue is new after having checked both open and closed issues at https://github.com/opnsense/core/issues?q=is%3Aissue

**Describe the bug**

The green circles on the left next to each item do not update if an alert has failed. 

**To Reproduce**

1.  Add the monit widget
2.  Add a monit test service that works such as ping external site
3.  Check dashboard to make sure the monit widget item is green
4.  Create condition where test service fails, such as changing ping IP to one that is un=pingable
5.  Check monit dashboard again to see if the status has changed.

**Expected behavior**

The green circles next to each item being monitored by monit on the dashboard should turn red in the case of a failure in the service being monitored. Additionally, when I hover my mouse over the green circles, if an alert has failed it just says "undefined" instead of "ok." The hover over text should say "failed"

**Screenshots**
In the screenshot, the item "Test_Ping_Check" has failed and you can see the mouseover is showing "undefined"

![image](https://github.com/user-attachments/assets/7260a4b7-8f6c-41b7-a02d-aa275afe220b)

**Relevant log files**

N/A

**Additional context**

N/A

**Environment**

OPNsense 24.7_5-amd64 (amd64).
Intel i3-9100f
Supermicro X11SCL-IF

Monviech:
When the ping does not work, its status in the API shows as ``"status":"16384"``

Since this is not mapped in the widget, it goes to undefined.

PR #7705, opened 2024-08-01
dashboard: monit widget: Invert the logic of success and failed

Monviech:
When a status code is not mapped, it will always fall back to failed in the tooltip, and text-danger as color. Fixes issue with ping check, the status is 16384 when failed. Since it can be assumed that all statuses other than 0 are bad, this change in logic seems pragmatic.

Fixes: https://github.com/opnsense/core/issues/7669

### DIFF
--- a/src/opnsense/www/js/widgets/Monit.js
+++ b/src/opnsense/www/js/widgets/Monit.js
@@ -54,7 +54,7 @@ export default class Monit extends BaseTableWidget {
         };
 
         this.statusColors = {
-            "1": "text-danger",
+            "0": "text-success",
             "2": "text-warning",
             "3": "text-warning"
         };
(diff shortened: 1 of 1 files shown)


Monviech:
Bildschirmfoto 2024-08-01 um 15 04 39

src/opnsense/www/js/widgets/Monit.js
```diff
@@ -54,7 +54,7 @@ export default class Monit extends BaseTableWidget {
         };
 
         this.statusColors = {
-            "1": "text-danger",
+            "0": "text-success",
             "2": "text-warning",
             "3": "text-warning"
```
fichtner, line 59: @AdSchellevis this mapping looks over-engineered from the start, but factually the fix looks good

AdSchellevis: @fichtner to be honest, I think the problem actually starts with the api (https://github.com/opnsense/core/blob/master/src/opnsense/mvc/app/controllers/OPNsense/Monit/Api/StatusController.php). It would be better if the endpoint would also send a readable status back, in which case you could just use it without further conversion (only prefix `text-`) .

fichtner: Improving the API would also be ok, though keeping it minimal/compatible could be challenging

swhite2 (approved):

fichtner:
@AdSchellevis merge now, add a ticket for API review?

AdSchellevis:
@fichtner sounds good, thanks

fichtner:
Merged, thanks!

Issue #7657, opened 2024-07-26
Cores and threads appear to be swapped in the CPU and thermal sensors widgets


NOYB:
Reporting 4 cores and 2 threads for 11th Gen Intel Core i3-1115G4.
Should be 2 cores and 4 threads.

Similar in thermal sensors widget reporting 4 cores and 2 zones.

24.7

Reference source:
https://ark.intel.com/content/www/us/en/ark/products/208652/intel-core-i3-1115g4-processor-6m-cache-up-to-4-10-ghz.html

![CPU Cores   Threads](https://github.com/user-attachments/assets/8f1fa825-270c-4878-ac84-68e41225f338)    ![Thermal Sensors Cores   Threads](https://github.com/user-attachments/assets/d0613f0f-d106-4a89-a0f8-2b5e80d6bc37)

fichtner:
fixed 6 hours ago 678eaf2fb99afc

NOYB:
Fixes CPU widget but not Thermal Sensors widget.  Thermal Sensors widget still shows 4 cores (0,1,2,3) and 2 zones (0,1).

fichtner:
I did a tweak for thermal sensors but af74aa42ab from the looks of it it just reports what you have. Might be worth checking sysctls:

    # configctl system temp


Cheers,
Franco

NOYB:
Doesn't resolve the issue and seems to randomly inflict some strange behavior.
Mostly displays this:
![Snap1](https://github.com/user-attachments/assets/e4bd1faa-2cd1-4e07-b35a-c6f38b7cf81a)
Infrequently displays this:
![Snap2](https://github.com/user-attachments/assets/f491454d-8352-4f8e-a5a7-1fe78b7de767)

The configctl system temp command seems to report threads as cores.  Every thread pair is a single core though.

How about if hyperthreading enabled return only every other "core"?  i.e. the even or odd numbered "cores".

11th Gen Intel(R) Core(TM) i3-1115G4 @ 3.00GHz
(2 cores, 4 threads)

fichtner:
> The configctl system temp command seems to report threads as cores. Every thread pair is a single core though.

Dude, just post the actual output. I don’t have time for this.

NOYB:
```
configctl system temp
dev.cpu.0.temperature=49.0C
dev.cpu.1.temperature=49.0C
dev.cpu.2.temperature=57.0C
dev.cpu.3.temperature=57.0C
[...]

PR #7723, opened 2024-08-05
Thermal Sensors Widget ("cpu"/threads per core)

NOYB:
Report only every other "cpu" temperature if hyper-thread enabled.  i.e. 2 threads per core.


fichtner (changes requested):
As said in https://github.com/opnsense/core/issues/7657#issuecomment-2268712821 we are not mangling the backend output. My last message here.

fichtner:
Now showing the backend output. No need to manipulate.

Issue #7727, opened 2024-08-06
Thermal sensors elevated highly fluctuating core temperatures


NOYB:
Thermal sensors widget reporting elevated and highly fluctuating Core temperatures by about 10 to 20 degrees Celsius.

Typically reporting mid 50's to mid 60's Celsius with a high amount of fluctuation.

Shell sysctl typically reports low to mid 40's Celsius and much more stable.

NOYB:
From this:
![Snap1](https://github.com/user-attachments/assets/e9454c75-b7f8-4996-883b-7bd85c2852b0)

To this:
![Snap2](https://github.com/user-attachments/assets/10591cf8-cb76-4b33-99d0-e7b8e3a7b86a)

fichtner:
#7730

PR #7728, opened 2024-08-06
Thermal Sensors: Avoid reporting elevated and highly fluctuating "CPU" (core) temperatures

NOYB:
Thermal Sensors: Avoid reporting elevated and highly fluctuating "CPU" (core) temperatures

Running sysctl ${SYSCTLS} immediately after sysctl -aN | grep temperature results in elevated and highly fluctuating "CPU" (core) temperatures by about 10 to 20 degrees Celsius being returned.  Sleep for a second to avoid.

CPU: 11th Gen Intel(R) Core(TM) i3-1115G4

Dependent on
system: increase widget timeout to 5 seconds
0d38c780459117afb045d20c9e0ee8edbfc18c3b

Fixes Thermal sensors elevated highly fluctuating core temperatures #7727
https://github.com/opnsense/core/issues/7727#issue-2449698175

### DIFF
--- a/src/opnsense/scripts/system/temperature.sh
+++ b/src/opnsense/scripts/system/temperature.sh
@@ -46,6 +46,7 @@ else
 	# as long as we can find something that matches our search.
 	SYSCTLS=$(sysctl -aN | grep temperature)
 	if [ -n "${SYSCTLS}" ]; then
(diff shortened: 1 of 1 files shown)


fichtner:
As a POC that's not even required I agree with this. In practice stuffing arbitrary delays is worst practice.

Issue #7649, opened 2024-07-25
WebGUI doesn't start on 24.1.10 and 24.7


CatFoxVoyager:
Issue Summary

Problem: The WebGUI in OPNsense versions 24.1.10 and later (including 24.7) fails to work properly. Reverting to version 24.1.8 resolves the issue.

Steps to Reproduce

    Upgrade OPNsense to version 24.1.10 or later.
    Attempt to access the WebGUI.
    Observe that the WebGUI does not load
    

OPNsense 24.1.8-amd64
FreeBSD 13.2-RELEASE-p11
OpenSSL 3.0.14

    
   Log :
   
   `2024-07-25 15:32:48: (/usr/obj/usr/ports/www/lighttpd/work/lighttpd-1.4.76/src/mod_openssl.c.1453) SSL: inactive/expired X509 certificate '/var/etc/cert.pem'
2024-07-25 15:32:48: (/usr/obj/usr/ports/www/lighttpd/work/lighttpd-1.4.76/src/network.c.604) bind() [fe80::200:b1ff:fe4d:0]:443: Can't assign requested address
...
2024-07-25T15:32:48-04:00 opnsense 296 - [meta sequenceId="305"] <118>Starting web GUI...failed.
`

After doing `configctl webgui restart renew`

`event @ 1721936304.36 exec: system event config_changed response: OK  
<171>1 2024-07-25T15:38:44-04:00 opn.exemple.com config 454 - [meta sequenceId="4"] /usr/local/etc/rc.restart_webgui: The command '/usr/local/bin/flock -ne /var/run/lighty-webConfigurator.pid /usr/local/sbin/lighttpd -f /var/etc/lighty-webConfigurator.conf' returned exit code '255', the output was '2024-07-25 15:38:44: (/usr/obj/usr/ports/www/lighttpd/work/lighttpd-1.4.76/src/network.c.604) bind() [fe80::200:b1ff:fe4d:0]:443: Can't assign requested address'`

configt of lighttpd : [lighty-webConfigurator.txt](https://github.com/user-attachments/files/16383071/lighty-webConfigurator.txt)

I've Suricata, got some error after upgrading to 24.1.10 and 24.7

'<13>1 2024-07-25T15:14:37-04:00 opn.exemple.com  kernel - - [meta sequenceId="419"] <6>vlan100: permanently promiscuous mode enabled
[...]

PR #7731, opened 2024-08-06
Skip tentative IPv6 addresses for binding of webgui. Fixes: #7649

tionu:
IPv6 addresses might be tentative. lighttpd will fail when trying to bind to a tentative address. This commit skips tentative IPv6 addresses for binding.

### DIFF
--- a/src/etc/inc/plugins.inc.d/webgui.inc
+++ b/src/etc/inc/plugins.inc.d/webgui.inc
@@ -83,7 +83,7 @@ function webgui_configure_do($verbose = false, $interface = '')
     $listeners = count($interfaces) ? [] : ['0.0.0.0', '::'];
 
     foreach (interfaces_addresses($interfaces) as $tmpaddr => $info) {
-        if (!$info['bind']) {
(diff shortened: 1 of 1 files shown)


fichtner:
Merged, thanks! Do you have interest in adjusting openssh.inc in the same way?

tionu:
OpenSSH does not fail when trying to bind to a tentative IPv6 address. It just skips them.
The equivalant block would be at [openssh.inc#L197](https://github.com/opnsense/core/blob/bd3faedadb0384b099bb1bb739835fd0767605c8/src/etc/inc/plugins.inc.d/openssh.inc#L197).
I've dumped the interface info to log (`service_log(json_encode($info), $verbose);`) and the link-local IPv6 addresses show up as tentative. OpenSSH still only binds to non-tentative addresses (checked with `sockstat -4 -6 -p 22` after boot).

I'm hesitant to submit a pull request because it ain't broken.

fichtner:
Fair enough. The funny thing is this may further complicate what the actual issue is and how it can be handled generically. :/

Issue #7734, opened 2024-08-06
Change title of "Swap Usage" widget to "Swap"


pmhausen:
**Important notices**

Before you add a new report, we ask you kindly to acknowledge the following:

- [X] I have read the contributing guide lines at https://github.com/opnsense/core/blob/master/CONTRIBUTING.md
- [X] I am convinced that my issue is new after having checked both open and closed issues at https://github.com/opnsense/core/issues?q=is%3Aissue

**Is your feature request related to a problem? Please describe.**

It is purely cosmetic but this display is driving me nuts:
<img width="308" alt="Bildschirmfoto 2024-08-06 um 20 08 50" src="https://github.com/user-attachments/assets/23d6957a-94ee-4b98-94af-2ea4927999d0">

It would be great if all of these were the same size.

PR #7736, opened 2024-08-07
dashboard: Make titles of Swap and Mbuf shorter 

Monviech:
...so there is no line break that increases the size of the widget.

Fixes: https://github.com/opnsense/core/issues/7734


fichtner (approved):
Nice, you addressed the user concern and anticipated the next user concern for the same problem scope 👍

swhite2 (approved):

pmhausen:
@Monviech Hero! 🙂

PR #7740, opened 2024-08-07
Captive portal logoff issue fixed

Vivekmauli14:
Captive portal > user unable to logoff via the user form 

I found the issue that actions.d > captive portal > disconnect command is mismatched parameters that is why it is not getting logoff.

either we can modify the actions.d command for captive portal with parameter or as I raised this PR, remove zoneid from the accesscontroller.


AdSchellevis:
looks like a regression in https://github.com/opnsense/core/commit/8ba454ab04886156a29c1a7314937597b88b34ff , thanks!

PR #7743, opened 2024-08-07
System: Boot Environments - Addtional cleanups and restructures

sheridans:
check boot environment to delete is not in any active state


AdSchellevis:
@sheridans thanks, that's helpful. We're almost there :)

PR #7745, opened 2024-08-08
replace mail with unique id

jameskimmel:
If I am not mistaken, there is no e-mail based auth anymore and it got replaced by unique ID of the services account.

### DIFF
--- a/src/opnsense/mvc/app/library/OPNsense/Backup/GDrive.php
+++ b/src/opnsense/mvc/app/library/OPNsense/Backup/GDrive.php
@@ -51,9 +51,9 @@ class Gdrive extends Base implements IBackupProvider
             "value" => null
         );
         $fields[] = array(
-            "name" => "GDriveEmail",
+            "name" => "GDriveUID",
             "type" => "text",
-            "label" => gettext("Email Address"),
+            "label" => gettext("Unique ID"),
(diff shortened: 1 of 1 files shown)


src/opnsense/mvc/app/library/OPNsense/Backup/GDrive.php
```diff
@@ -192,7 +192,7 @@ public function backup()
                 try {
                     $client = new \Google\API\Drive();
                     $client->login(
-                        (string)$config->system->remotebackup->GDriveEmail,
+                        (string)$config->system->remotebackup->GDriveUID,
```
AdSchellevis, line 195: this will break existing setups, as the address/id is currently stored here.

jameskimmel: But only changing the label would be fine? Unfortunately I am not a coder.

AdSchellevis: if their documentation refers to uid now instead of email, I'm fine with changing the label, I haven't checked myself yet to be honest.

jameskimmel: I also did a pull request for the docs here: https://github.com/opnsense/docs/pull/586

Issue #6911, opened 2023-10-04
CPU temperature reporting in dashboard


dimitrijp:
**Important notices**

Before you add a new report, we ask you kindly to acknowledge the following:

- [x ] I have read the contributing guide lines at https://github.com/opnsense/core/blob/master/CONTRIBUTING.md
- [x ] I am convinced that my issue is new after having checked both open and closed issues at https://github.com/opnsense/core/issues?q=is%3Aissue

Im testing new Intel N200 box, and came across interesting issue with CPU temperature reporting. My fanless box is practically cool to touch, but CPU temp in OPNSense is reporting 45 degrees, that it seemd to me too much. So I did some digging.

If I log in console and issue sysctl dev.cpu | grep temperature, I got more realistic CPU temp

sysctl dev.cpu | grep temperature
dev.cpu.3.temperature: 35.0C
dev.cpu.2.temperature: 35.0C
dev.cpu.1.temperature: 36.0C
dev.cpu.0.temperature: 35.0C

I checked what command is OPNSense using for temp sensors and its sysctl -a | grep temperature, this temp is also reported in dashboard.

sysctl -a | grep temperature
hw.acpi.thermal.tz0.temperature: 27.9C
dev.cpu.3.temperature: 45.0C
dev.cpu.2.temperature: 45.0C
dev.cpu.1.temperature: 44.0C
dev.cpu.0.temperature: 43.0C

If I stress test the machine this 2 temps come together at 55 degrees after 30 min witch is ok.

I installed HTOP and check the temp there, the lower values are reported, with occasional spike above 40 but average at 36 - 37 degrees.

I live booted linux on the same machine and checked linux temp sensors .. they are also reporting average of 36 - 38 degrees.

Out of curiosity I also spin up pfsense, and temp was also ok .. no spikes above 40 degrees. - but pfsense is using different command to read cpu temp then opnsense, I think sysctl dev.cpu | grep temperature

Overall this is not life threatening situation 😊 😊, but only my observation I decided to post it here. According to my testing both temps are correct higher temp is actually achieved by CPU when short load hits it, but it lasts less then a second and it falls to 36 degrees. It looks to me that the command sysctl -a | grep temperature is displaying max values in certain period, averaging in higher temp in dashboard that it really is.

I also searched the forum, and few similar post have already been posted;

https://forum.opnsense.org/index.php?topic=34395.msg166556#msg166556

[...]

Issue #7730, opened 2024-08-06
system: add a python script that monitors and perhaps averages temperature readings


fichtner:
**Important notices**

Before you add a new report, we ask you kindly to acknowledge the following:

- [x] I have read the contributing guide lines at https://github.com/opnsense/core/blob/master/CONTRIBUTING.md
- [x] I am convinced that my issue is new after having checked both open and closed issues at https://github.com/opnsense/core/issues?q=is%3Aissue

**Is your feature request related to a problem? Please describe.**

System temperatures are reported in such a fine-grained fashion that reading system information including temperatures reports higher temperatures because the CPU is busier at that particular point in time. In fact, neither the temperature reported is wrong nor is the system doing the wrong thing. The problem is merely that we're looking at the spikes but miss the bigger picture that the CPU is generally cooler.

**Describe the solution you like**

Python backend script capturing all known temperature sources for easy consumption from the backend.

**Describe alternatives you considered**

N/A

**Additional context**

N/A

fichtner:
A pre-cache may be nice, but not really relevant for functional implementation.

PR #7758, opened 2024-08-11
Improve CPU temperature measurement

stuart-mclaren:
There can be a mis-match between the CPU temperature reported by the operating system and the CPU temperature shown in the OPNsense dashboard.

This happens for N100, N200 and possibly other CPU types.

The issue seems to be that the sysctl -a command (which returns around 14,000 lines of output) causes a short-lived spike in CPU temperature. This spike causes an elevated temperature value to be reported in the dashboard.

Avoiding "sysctl -a" means that the dashboard temperature matches the OS reported temperature.

Current script:

```
 # /usr/local/opnsense/scripts/system/temperature.sh
 dev.cpu.0.temperature=46.0C
 .
 .
 .
```

With this change:

```
 # ./temperature.sh
 dev.cpu.0.temperature=40.0C
 .
 .
 .
```

This issue was previously reported here:
 https://github.com/opnsense/core/issues/6911

### DIFF
--- a/src/opnsense/scripts/system/temperature.sh
+++ b/src/opnsense/scripts/system/temperature.sh
@@ -32,6 +32,56 @@ hw.acpi.thermal.tz0.temperature
 hw.temperature.CPU
 "
 
+# count_cpu gets the number of cpu
+count_cpu() {
+	local n
+	n=$(sysctl -n hw.ncpu)
+
+	if [ $? -ne 0 ]
+	then
+		n=0
+	fi
+	eval "$1=$n"
+}
+
+read_cpu_temperatures() {
+	local ncpu=0
+	count_cpu ncpu
+	if [ $ncpu -gt 0 ]
+	then
+		sysctl -i -e $(jot -w 'dev.cpu.%d.temperature' $ncpu 0)
+	fi
+	sysctl -i -e hw.temperature.CPU
+}
+
+# count_thermal_zones gets the number of thermal zones
+count_thermal_zones() {
+	local n=0
+	while true
+	do
+		sysctl -n hw.acpi.thermal.tz${n}.temperature > /dev/null 2>/dev/null
+		if [ $? -ne 0 ]
+		then
+			break
+		fi
+		n=$((n + 1))
+	done
+	eval "$1=$n"
+}
+
+read_thermal_zone_temperatures() {
+	local ntz=0
+	count_thermal_zones ntz
+	if [ $ntz -gt 0 ]; then
+		sysctl -i -e $(jot -w 'hw.acpi.thermal.tz%d.temperature' $ntz 0)
+	fi
+}
+
+read_temperatures() {
+	read_cpu_temperatures
+	read_thermal_zone_temperatures
+}
+
 if [ "${CMD}" = 'rrd' ]; then
 	for SYSCTL in ${SYSCTLS}; do
 		TEMP=$(sysctl -i -n ${SYSCTL} | sed 's/C//g')
@@ -41,11 +91,5 @@ if [ "${CMD}" = 'rrd' ]; then
 		fi
 	done
 else
-	# The grep is opportunistic, but at least we only grep the
-	# variable names and not their content at the same time and
-	# as long as we can find something that matches our search.
-	SYSCTLS=$(sysctl -aN | grep temperature)
-	if [ -n "${SYSCTLS}" ]; then
-		sysctl -e ${SYSCTLS} | sort
-	fi
+	read_temperatures
 fi


fichtner:
The idle temps are wrong too if you follow that logic. Changing the script slightly does nothing but skew the result and the next one will complain later. We need some sort of averaging strategy and sanely placed intervals to fetch data evenly spaced out.

fichtner:
Besides you are probably going to destroy readings for some platforms by assuming there are only two possibilities.

stuart-mclaren:
> The idle temps are wrong too if you follow that logic.

Can you clarify what you mean here? Thanks.

fichtner:
The temperature reading is real time. You get a different reading on idle, pseudo-idle and busy and pseudo-busy. The pseudo values are wrong readings because you either miss or hit the computation peak temperature.

NOYB:
The issue actually extends further upstream than just the sysctl -a command.
The call to "temperature.sh" is a "new backend".  There is a lot of processing going on even before sysctl -a is run.

The sysctl -a can be avoided pretty easily by saving/retrieving from ramdisk (/tmp).
But the new backend not so much.  A very short delay seems to do the trick though.

temperature.sh:
```
else
	# The grep is opportunistic, but at least we only grep the
	# variable names and not their content at the same time and
	# as long as we can find something that matches our search.
-	SYSCTLS=$(sysctl -aN | grep temperature)
+
+	# Save to file instead of constructing every time.
+	SYSCTLS_FILE='/tmp/SYSCTLS_Thermal_Sensors_Widget'
+	if [ ! -s "${SYSCTLS_FILE}" ]; then
+		sysctl -aN | grep temperature" > "${SYSCTLS_FILE}"
+	fi
+
+	SYSCTLS=$(cat "${SYSCTLS_FILE}")
+
	if [ -n "${SYSCTLS}" ]; then
+		sleep 0.5	# Avoid returning elevated and highly fluctuating "CPU" (core) temperatures
		sysctl -e ${SYSCTLS} | sort
	fi
fi

```

fichtner:
https://github.com/opnsense/core/issues/7730

stuart-mclaren:
> The temperature reading is real time. You get a different reading on idle, pseudo-idle and busy and pseudo-busy. The pseudo values are wrong readings because you either miss or hit the computation peak temperature.

Thanks for clarifying.

I think what's happening is slightly different.

> The issue actually extends further upstream than just the sysctl -a command.

Yes, I can see that effect too. (I'd experimented with "sleep 1".)

But I think there's something else going on not random/that's specific to "sysctl -a".
 
Consider this script:

```
#!/bin/sh
i=0
while [ "$i" -lt 4 ]
do
  i=$(($i + 1))
  sysctl dev.cpu.0.temperature
done
sysctl -a | grep dev.cpu.0.temperature
```

Running it (on some hardware) can give something like this:

dev.cpu.0.temperature: 49.0C
dev.cpu.0.temperature: 49.0C
dev.cpu.0.temperature: 49.0C
dev.cpu.0.temperature: 49.0C
dev.cpu.0.temperature: 57.0C <<< "sysctl -a" reading

This jump doesn't seem to be random variation due to cpu activity. It's
more like a large "observer effect" [1]. The effect goes away when "sysctl -a" is
not used.

Even if you want to calculate average numbers, you will need to get
numbers that exclude this effect.

A side note: cpus can shutdown if their temperature exceeds a certain threshold.
It's *extremely* unlikely, but in principle, the temperature spike caused by "sysctl -a" could cause an outage. 

Would the following be reasonable?

I can update the existing patch to add "sleep 0.5".

This gives the following benefits (with no downside?):

1. Removes the inefficient filtering of 14,000 lines to extract
 ~5 lines of output (pfsense made an equivalent change in 2017 [2])
2. Avoids the perception that opnsense boxes are running hotter than they really are

Thanks.

[1] https://en.wikipedia.org/wiki/Observer_effect_(physics)
[2] https://github.com/pfsense/pfsense/pull/4475

AdSchellevis:
the high fluctuations are usually just an effect of an inefficient cooling solution, in which case the higher boundary (doing something) might even be closer to reality anyway when looking at averages.

fichtner:
We could do something like f473d9a5c74 and be done with it. It does have the benefit to be able to ignore the mocked per-CPU values at least for AMD it seems a sane sensor point exists, but needs better numbering.

NOYB:
re: Ad comment
The dev.cpu temperatures are from on-die sensors that are very fast reacting.  The peaks are very momentary.  Like a few hundred milliseconds.  Cooling system is not going to have much impact on that.  Like a transistor heat sink vs. transient junction temperatures.

fichtner:
I'm beginning to understand some people would like the monitoring to lie to them showing a lower temperature than reality, but also not trusting in averages. It's almost funny if it wasn't so exhausting to talk about it.

Monviech:
I want to understand why this is all such a big deal. Shouldn't it all be okay as long as the temperatures stay below the hardware threshold?

For an N100 for example the T Junction is 105°C. https://ark.intel.com/content/www/us/en/ark/products/231803/intel-processor-n100-6m-cache-up-to-3-40-ghz.html

Why does it matter if it now reads 47 or 50°C? It does not affect anything.

Most Server boards have an ambient temperature sensor so the component temperature can be put into proper perspective (mostly on the PSU air inlet)

AdSchellevis:
In the grand scheme of things this is quite irrelevant. If the issue is different cores report different values, averaging for the same cpu might work, but there certainly are bigger fish to fry than this.

doktornotor:
Can't we just do something like `echo(rand(20,100));`, or - bonus points - perhaps let the user choose whether they prefer their CPUs raw, medium or well done and adjust the range accordingly?

Monviech:
I have just tested with an IR thermometer on my DEC 740 at home. Widget shows 53°C, case temperature on heatsink measures 49.7°C. So its pretty accurate.

stuart-mclaren:
> some people would like the monitoring to lie to them

Can you let me know if that was comment was directed at me? Thanks.

What would people like to do here?

The idea that the number in the dashboard should match the number reported by the operating system doesn't seem to be something that we can get consensus on. 

Putting that aside: the patch does still prevent 'overmatching' of the oids (ie avoids generating thousands of lines that aren't needed). Equivalent to https://github.com/pfsense/pfsense/pull/4475.

Is that something that people would see as a (small) improvement?

Let me know. If not I can close the PR.

fichtner:
I wrote f473d9a5c74ed7466 but unless there is *some* consensus over what we can achieve and should and shouldn't do it's pointless to continue with any sort of "solution to the "problem".

1. Personally I want to get all the available temperature readings from the kernel now or in the future. There is no database, so some sort of data collections needs to take place. The current code tries to do it, the commit above does it with 100% certainty.
2. Temperature readings should reflect the system state and we need to assume temperature readings are correct all the time from a processor perspective. That means no side-stepping and idle-forcing of readings prolonging the measurement fetch. That also means we can't avoid the GUI heating the CPU while collecting data from the dashboard.
3. Stick with a solution instead of always criticising the reported CPU temperature in edge cases (as done in the above commit).
4. Bonus point: not hiding shell command injection in the code. Another PR is guilty of this.

Maybe point two is futile. We could also use averages but I doubt we can settle for this as it probably also needs a backend service to do that polling and averaging. Point 3 seems equally out of reach at the moment.


Cheers,
Franco

stuart-mclaren:
I'll close this PR.

If anyone revisits temperature measurement in the future, I'd encourage them to
to take into account that calling "sysctl -a" appears to skew the temperature readings, introducing an "observer effect", at least on some systems.

Again, this script:

```
#!/bin/sh
i=0
while [ "$i" -lt 4 ]
do
  i=$(($i + 1))
  sysctl dev.cpu.0.temperature
done
sysctl -a | grep dev.cpu.0.temperature
```

Can produce output like this:

```
dev.cpu.0.temperature: 50.0C
dev.cpu.0.temperature: 50.0C
dev.cpu.0.temperature: 50.0C
dev.cpu.0.temperature: 50.0C
dev.cpu.0.temperature: 59.0C  <<< skewed "sysctl -a" reading
```

NOYB:
A probe that significantly alters what it is probing/measuring is typically undesirable.

The sysctl -a command significantly alters (increases) the temperature of the on-die sensors being measured.  Grepping all that output adds insult to injury so to speak.  As does instantiating a backend.

Though the temperature being reported is correct for a very brief moment while the probe is doing it's thing (maybe a couple hundred milliseconds).  It is not the temperature absent the probe (many thousands of milliseconds interval between the probe events).

Like taking a person's temperature by a means that gives them a fever.  Sure, okay, the person has a fever.  But it was the taking of their temperature that gave them the fever.

Don't think anyone is suggesting that the temperature reading is not accurate.  But that it is being significantly influenced by the "probe" for the very brief time that the temperature is being read.

fichtner:
Ok, how do we offer temperature readings if we cannot query for eligible OIDs even once because that's a "hot" subject?

fichtner:
You can clearly see this was an issue at least in 2016 when 90a31b13da54fd as written. But the old code also did a `sysctl -a` .. so maybe FreeBSD 14.1 is "wrong" now creating all this heat? Or where does the sudden change come from.

NOYB:
Pre create them so that the probe is not dependent on a "sysctl -a" command.

There can be a variety of reasons for an issue not being highlighted for a period of time and doesn't necessarily mean that it didn't exist.

stuart-mclaren:
> But you're also not disputing the fact that the change from 24.1 to 24.7 suddenly made a decade old behaviour more obvious? ;)

I'm on 24.1

```
# opnsense-version 
OPNsense 24.1.10_3
```

NOYB:
The cache-ttl was on the right track and I thought it was/is a positive thing.  But there are other pieces that come into play with it too.
Such as doing the probe right on it's heals the first time.  Also right on the heals of instantiating the backend every time.

Perhaps the OIDs could be put into an environment var at OPNsense startup?  Then it's just the backend instantiation.

Actually I'm neither disputing nor acknowledging that the change from 24.1 to 24.7 suddenly made a decade old behavior more obvious.  I simply don't know if it has been made more obvious or not.  As for it's previous behavior I never looked.  So would know.  What I do know is that it exists.

Issue #6923, opened 2023-10-10
Unable to set NTP Pools other than subdomains of pool.ntp.org


CJ-Ross:
**Important notices**

Before you add a new report, we ask you kindly to acknowledge the following:

- [x ] I have read the contributing guide lines at https://github.com/opnsense/core/blob/master/CONTRIBUTING.md
- [ x] I am convinced that my issue is new after having checked both open and closed issues at https://github.com/opnsense/core/issues?q=is%3Aissue

**Describe the bug**

This commit introduced the ability of using ntp pools but restricts it to subdomains of pool.ntp.org.

https://github.com/opnsense/core/commit/7366d785a62d86de5f7eb7bfab15039492eea9bf

https://github.com/opnsense/core/issues/5569

**To Reproduce**

1. Go to Network Time Service General page
2. Add a pool such as pool.ntp.org or time.nist.gov and save
3. Go to Network Time Service Status page
4. See that added entry resolves to just one server.

**Expected behavior**

Ability to designate time servers as pools.

**Describe alternatives you considered**

Add a server/pool toggle to the time server UI along with the existing Prefer, Iburst, and Do not use.

daerSeebaer:
Hi,
so I tried to implement this as a checkbox that switches a given time server from "server" to "pool". The plan was to take the checkbox value, if the checkbox is set the config line begins with "pool", if not it begins with "server".

So far I have managed to implement the checkbox itself, and if it is set the server gets written into the OPNSense config file unter ntpd->pool as expected:

```
<ntpd>
    <logsys>on</logsys>
    <iburst>pool.ntp.org</iburst>
    <pool>pool.ntp.org</pool>
  </ntpd>
```

However, when the code tries to load the config entry, it crashes and claims that it got an array instead of a string:
`Fatal error: Uncaught TypeError: explode(): Argument #2 ($string) must be of type string, array given in /usr/local/etc/inc/plugins.inc.d/ntpd.inc:332`

The offending lines (the "$pool" line is 332):
```
    $noselect = isset($config['ntpd']['noselect']) ? explode(' ', $config['ntpd']['noselect']) : [];
    $pool = isset($config['ntpd']['pool']) ? explode(' ', $config['ntpd']['pool']) : [];
    $prefer = isset($config['ntpd']['prefer']) ? explode(' ', $config['ntpd']['prefer']) : [];
    $iburst = isset($config['ntpd']['iburst']) ? explode(' ', $config['ntpd']['iburst']) : [];
```

[...]

PR #7760, opened 2024-08-11
Ntp: Add pool checkbox to timeserver configuration

daerSeebaer:
This PR solves #6923 while preserving the current default behavior.

## Technical description
I more or less just copied one of the other checkboxes and replaced everything with "pool_server". I also tested it by copying the impacted files into a running OPNSense, and then successfully using the new configuration option and verifying the result by inspecting the generated ntp.conf file.

## Reasoning
This change makes it more transparent how the entries in the ntp config are configured, and gives more options to administrators to either set a NTP Pool server as "server" type or use the "pool" keyword for pooled servers besides the NTP Pool.

## Impact
I extended the default configuration file to set the pool config for the default timeservers, so new installations will not change their behavior.

For existing installations, since the option is "opt in", the missing pool option will set all confgurations for NTP Pool servers back to a "server" type association. If there is a possibility to migrate existing configurations and keep the existing configuration during installation of this fix (`pool` for all NTP Pool servers, `server` for anything else), that would be preferrable but I don't know how to do that. One alternative idea was to make the feature "opt out", setting the pool type unless a timeserver gets explicitely marked as "not pool", but that also would impact existing installations by setting the pool type for manually configured not-NTP Pool servers.

### DIFF
--- a/src/etc/config.xml.sample
+++ b/src/etc/config.xml.sample
@@ -323,6 +323,7 @@
     <enable/>
   </rrd>
   <ntpd>
+    <pool_server>0.opnsense.pool.ntp.org 1.opnsense.pool.ntp.org 2.opnsense.pool.ntp.org 3.opnsense.pool.ntp.org</pool_server>
     <prefer>0.opnsense.pool.ntp.org</prefer>
   </ntpd>
   <widgets>
--- a/src/etc/inc/plugins.inc.d/ntpd.inc
+++ b/src/etc/inc/plugins.inc.d/ntpd.inc
@@ -329,17 +329,18 @@ function ntpd_configure_do($verbose = false)
 
     $noselect = isset($config['ntpd']['noselect']) ? explode(' ', $config['ntpd']['noselect']) : [];
     $prefer = isset($config['ntpd']['prefer']) ? explode(' ', $config['ntpd']['prefer']) : [];
+    $pool_server = isset($config['ntpd']['pool_server']) ? explode(' ', $config['ntpd']['pool_server']) : [];
     $iburst = isset($config['ntpd']['iburst']) ? explode(' ', $config['ntpd']['iburst']) : [];
 
     $ntpcfg .= "\n\n# Upstream Servers\n";
     /* foreach through ntp servers and write out to ntpd.conf */
     foreach (explode(' ', $config['system']['timeservers']) as $ts) {
-        /* Determine if Network Time Server is from the NTP Pool or not */
-        if (preg_match("/\.pool\.ntp\.org$/", $ts)) {
-            $ntpcfg .= "pool {$ts}";
+        if (in_array($ts, $pool_server)) {
+            $ntpcfg .= "pool";
         } else {
-            $ntpcfg .= "server {$ts}";
+            $ntpcfg .= "server";
         }
+        $ntpcfg .= " {$ts}";
         if (in_array($ts, $iburst)) {
             $ntpcfg .= ' iburst';
         }
--- a/src/www/services_ntpd.php
+++ b/src/www/services_ntpd.php
@@ -75,10 +75,12 @@ if ($_SERVER['REQUEST_METHOD'] === 'GET') {
 
     // parse timeservers
     $pconfig['timeservers_host'] = array();
+    $pconfig['timeservers_pool'] = array();
     $pconfig['timeservers_noselect'] = array();
     $pconfig['timeservers_prefer'] = array();
     $pconfig['timeservers_iburst'] = array();
     if (!empty($config['system']['timeservers'])) {
+        $pconfig['timeservers_pool'] = !empty($a_ntpd['pool_server']) ? explode(' ', $a_ntpd['pool_server']) : array();
         $pconfig['timeservers_noselect'] = !empty($a_ntpd['noselect']) ? explode(' ', $a_ntpd['noselect']) : array();
         $pconfig['timeservers_prefer'] = !empty($a_ntpd['prefer']) ? explode(' ', $a_ntpd['prefer']) : array();
         $pconfig['timeservers_iburst'] = !empty($a_ntpd['iburst']) ? explode(' ', $a_ntpd['iburst']) : array();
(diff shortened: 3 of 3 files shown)


doktornotor:
I'm probably getting lazy and this static PHP code needs MVC/API conversion anyway. What's the issue with keeping the current behaviour? I mean:

```
if (in_array($ts, $pool_server) || preg_match("/\.pool\.ntp\.org$/", $ts)) {
	...
} else {
	...
}
```

Requires no migration, or?

daerSeebaer:
> I'm probably getting lazy and this static PHP code needs MVC/API conversion anyway. What's the issue with keeping the current behaviour? I mean:
> 
> ```
> if (in_array($ts, $pool_server) || preg_match("/\.pool\.ntp\.org$/", $ts)) {
> 	...
> } else {
> 	...
> }
> ```
> 
> Requires no migration, or?

One of my goals was to make it configurable if NTP Pool hostnames should be handled as "pool" or "server". But yes, if we keep the old behavior that would preserve all current configurations across the change.

If that is the way forward, I would suggest that for entries matching the pool-rule, the frontend should indicate the backend behavior by showing a checked, greyed out, uneditable checkbox in the "pool" column. And the help text could be expanded to also explain the behavior.


> this static PHP code needs MVC/API conversion anyway

Is that planned for the near future?

fichtner:
>> this static PHP code needs MVC/API conversion anyway
>
> Is that planned for the near future?

Not for 25.1 anyway.

I'm all for small scope patches first. I agree this is an issue, but considering what's been going on lately not the most pressing one for sure.

src/etc/config.xml.sample
```diff
@@ -323,6 +323,7 @@
     <enable/>
   </rrd>
   <ntpd>
+    <pool_server>0.opnsense.pool.ntp.org 1.opnsense.pool.ntp.org 2.opnsense.pool.ntp.org 3.opnsense.pool.ntp.org</pool_server>
```
fichtner, line 326: cleaning up the `timerserver` mess being anchored in `system` instead of `ntpd` would be much more appealing, but it requires MVC code to make this work appealing and future-proof.

fichtner: I mean the whole structure of the data is mind boggling. You want to create a server entry and add all options like also `pool_server` as  a property, but now you have all properties as global arrays that refer to `timeservers` which isn't even anchored here. This also negatively impacts partial config imports as it is BTW.

doktornotor:
> One of my goals was to make it configurable if NTP Pool hostnames should be handled as "pool" or "server". But yes, if we keep the old behavior that would preserve all current configurations across the change.

Is there some advantage to not treating pool as pool? (Beyond the note below).

> If that is the way forward, I would suggest that for entries matching the pool-rule, the frontend should indicate the backend behavior by showing a checked, greyed out, uneditable checkbox in the "pool" column. And the help text could be expanded to also explain the behavior.

Yes, it's already inconsistent with "Do not use". You [cannot use `noselect` with pools](https://www.ntp.org/documentation/4.2.8-series/confopt/)

Can do some PR if we agree how to treat `*pool.ntp.org` - FWIW "upstream" has this hardcoded as pool as well. 

```
noselect
Marks the server or peer to be ignored by the selection algorithm as unreachable, but visible to the monitoring program. 
This option is valid only with the server and peer commands.
```

From the experience with adding `noquery` to be enabled by default, I'd seriously rather avoid any configuration migrations here in this code mess.

src/etc/config.xml.sample
daerSeebaer: Yes, the config is all over the place. I tried to "fit in" with the other options like `<prefer>` and just put the `<pool_server>` next to them. For this PR that seemed like the best thing to do, since my goal was to introduce the ability for pool configurations, not to refactor the entire thing.

Do you have alternative ideas where and how to store this data? One option that comes to mind would be to start the migration to a more sensible model and introduce a list of timeserver-tags that for now only contains a `<pool_server>[true|false]</pool_server>`, however the naming conventions for that structure should be agreed upon since they should last into the MVC age.

daerSeebaer:
> > One of my goals was to make it configurable if NTP Pool hostnames should be handled as "pool" or "server". But yes, if we keep the old behavior that would preserve all current configurations across the change.
> 
> Is there some advantage to not treating pool as pool? (Beyond the note below).

Not really. All NTP Pool domains can reasonably be expected to return multiple, changing IP addresses. I initially wanted to make it configurable to bring configurability close to the level that an administrator handling the config file itself would have, but on second thought there is no real benefit. Together with the drawbacks of a migration I agree that the combined option of handling all `(.*\.)?pool.ntp.org` entries as pool and additionally being able to "opt in" additional servers would be best.

I'm going to add that functionality to this PR, but keep it as draft until https://github.com/opnsense/core/pull/7841 is merged. After that I can utilize the new JS functions and finalize the PR.

doktornotor:
> I'm going to add that functionality to this PR, but keep it as draft until #7841 is merged. After that I can utilize the new JS functions and finalize the PR.

Yeah, you'd better wait till someone sanitizes my garbage JS. 😂 (Though, I wrote it so that it would easily handle the additional "pool" checkboxes.)

src/etc/inc/plugins.inc.d/ntpd.inc
```diff
@@ -329,17 +329,18 @@ function ntpd_configure_do($verbose = false)
...
     $noselect = isset($config['ntpd']['noselect']) ? explode(' ', $config['ntpd']['noselect']) : [];
     $prefer = isset($config['ntpd']['prefer']) ? explode(' ', $config['ntpd']['prefer']) : [];
+    $pool_server = isset($config['ntpd']['pool_server']) ? explode(' ', $config['ntpd']['pool_server']) : [];
     $iburst = isset($config['ntpd']['iburst']) ? explode(' ', $config['ntpd']['iburst']) : [];
 
     $ntpcfg .= "\n\n# Upstream Servers\n";
     /* foreach through ntp servers and write out to ntpd.conf */
     foreach (explode(' ', $config['system']['timeservers']) as $ts) {
-        /* Determine if Network Time Server is from the NTP Pool or not */
-        if (preg_match("/\.pool\.ntp\.org$/", $ts)) {
```
fichtner, line 338: @daerSeebaer looking at this again what pool servers do not conform with this address? what servers are you using? Also `noselect` poses an issue when adding an arbitrary pool switch as shown by #7841

daerSeebaer: > @daerSeebaer looking at this again what pool servers do not conform with this address? what servers are you using?

The pool under pool.ntp.org is the capital P "NTP Pool". However, anyone that operates two or more ntp servers at different IP addresses can make them reachable under the same host name. Notable examples: `time.google.com`, `time.cloudflare.com` and `time.apple.com` all return multiple IP addresses and could each be used as a single entry with the `pool` option set. Personally, I also operate multiple time servers that are all reachable under a common host name that I don't want to disclose.

As discussed above, in order to not affect existing setups the current idea is to handle any `pool.ntp.org` entries **and** any entries where the pool flag is set as `pool` in the ntp config.

> Also `noselect` poses an issue when adding an arbitrary pool switch as shown by #7841

I'm fully aware of this and changed the status of this PR to draft when #7841 was opened. I was hoping that the other PR would be merged first and I could then finalize my PR to reflect the `noselect` behavior in the UI.

src/etc/config.xml.sample
stumbaumr: > cleaning up the `timerserver` mess being anchored in `system` instead of `ntpd` would be much more appealing, but it requires MVC code to make this work appealing and future-proof.

Since I use the os-chrony plugin as ntpd was unreliable I would prefer "Services - Network time" to become a plugin - so I can uninstall it. Same for Dnsmasq...

fichtner:
closing this one in favour of #9862

PR #7762, opened 2024-08-12
Thermal Sensors SYSCTLS Save/Retrieve

NOYB:
Eliminate "sysctl -aN" and "grep temperature" impact on Thermal Sensors widget readings.

Save the "SYSCTRLS" to /tmp file on first use after boot up for subsequent ongoing sensor readings.

### DIFF
--- a/src/opnsense/scripts/system/temperature.sh
+++ b/src/opnsense/scripts/system/temperature.sh
@@ -44,7 +44,15 @@ else
 	# The grep is opportunistic, but at least we only grep the
 	# variable names and not their content at the same time and
 	# as long as we can find something that matches our search.
-	SYSCTLS=$(sysctl -aN | grep temperature)
+
+	# Save/retrieve file instead of reconstructing SYSCTLS every time.
+	SYSCTLS_FILE='/tmp/SYSCTLS_Thermal_Sensors_Widget'
+	if [ ! -s "${SYSCTLS_FILE}" ]; then
+		sysctl -aN | grep temperature > "${SYSCTLS_FILE}"
+	fi
+
+	SYSCTLS=$(cat "${SYSCTLS_FILE}")
+
 	if [ -n "${SYSCTLS}" ]; then
 		sysctl -e ${SYSCTLS} | sort
 	fi


fichtner:
This is what #7730 would achieve without storing temporary data somewhere. We still have no way to average and people will keep complaining therefore.

NOYB:
Processor temperature averaging is overrated IMO.  Smooths out and obscures the peaks and lows.  Also, an average of what time period?  seconds, minutes, hours, ...???
Moment in time snapshots capture more peaks if there are more peaks occurring, and more lows if there are more lows occurring.  Think it makes for a a better gauge than an average.
Would people like an averaging fuel gauge their car?  Very much doubt it. ;)
Snap1

fichtner:
> Processor temperature averaging is overrated IMO.

So why do talk about temperatures are too high? Frankly, this is all a bit of a waste of time.

NOYB:
Because the sensors being measured are being significantly impacted by the measurement tool.

The measurement reading being too high is not the actual issue.  It's a symptom.  The issue is the measurement tool is significantly increasing the temperature.  Resulting in skewed measurements.

Typical case of test/measurement tool being a too significant part of the circuit.
The more sensitive a circuit is the more sensitive the test/measurement equipment needs to be to not significantly alter and impact the circuit under test/measurement.

fichtner:
> Because the sensors being measured are being significantly impacted by the lack of a measurement tool.

Fixed it for you.

src/opnsense/scripts/system/temperature.sh
```diff
@@ -44,7 +44,15 @@ else
...
 	# as long as we can find something that matches our search.
-	SYSCTLS=$(sysctl -aN | grep temperature)
+
+	# Save/retrieve file instead of reconstructing SYSCTLS every time.
+	SYSCTLS_FILE='/tmp/SYSCTLS_Thermal_Sensors_Widget'
+	if [ ! -s "${SYSCTLS_FILE}" ]; then
+		sysctl -aN | grep temperature > "${SYSCTLS_FILE}"
+	fi
+
+	SYSCTLS=$(cat "${SYSCTLS_FILE}")
```
fichtner, line 54: Added a shell injection BTW.

NOYB:
There is not a lack of a measurement tool.  The tool currently exists and is in use by the system.  It is just not "sensitive" enough to do the job without significant impact to the device being measured.

fichtner:
I see the good intentions here but this is not the way.

PR #7775, opened 2024-08-15
IDS: Support configurating eve-log for HTTP and TLS

tobyc11:
Add the appropriate front end controls and backend template for setting up TLS and HTTP monitoring for Suricata's eve-log.

### DIFF
--- a/src/opnsense/mvc/app/controllers/OPNsense/IDS/forms/generalSettings.xml
+++ b/src/opnsense/mvc/app/controllers/OPNsense/IDS/forms/generalSettings.xml
@@ -110,4 +110,53 @@
         <help>Send package payload to the log for further analyses.</help>
         <advanced>true</advanced>
     </field>
+    <field>
+        <id>ids.general.eveLog.http.enable</id>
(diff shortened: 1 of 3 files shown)


tobyc11:
Hi maintainers,

Currently there exists `include: custom.yaml` by default in OPNsense, but Suricata does not know how to merge lists. Once `outputs` is defined in `custom.yaml`, the whole outputs list is overriden. There exists no convenient way to specify additional outputs. This is why I really need these controls defined in the OPNsense UI.

Please take a look. Thanks!

AdSchellevis:
@tobyc11 I'll put it on my list of things to look at, we may need to reformat the form a bit if the number of options grow (not an issue, just an observation)

tobyc11:
Hi @AdSchellevis, any update on this? I can move the new items under a new tab if you would like.

AdSchellevis:
@tobyc11 very busy, but tried it and looks good. We should only reorganise the settings page a bit with headers, which I will do in the next commit. Thanks

PR #7785, opened 2024-08-18
Sidebar - submenu expand problem

opnsenseuser:
Fixes the problem that the submenu does not remain expanded if the focus remains on the menu item.

### DIFF
--- a/src/opnsense/www/js/opnsense_theme.js
+++ b/src/opnsense/www/js/opnsense_theme.js
@@ -51,7 +51,9 @@ $(document).ready(function () {
             var nextDiv = that.next('div');
 
             navigation.css('width', '415px');
-            if (nextDiv.not('in')) {
+            if (that.next('div').hasClass('in')) {
+                /* no action needed */
+            } else {
                 /* calculate coordinates for submenu */
                 var winHeight = $(window).height(),
                     offsetTop = that.offset().top,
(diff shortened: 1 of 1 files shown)


opnsenseuser:
@fichtner  it would be really important to merge this. I'm sorry for the inconvenience
regards rené

src/opnsense/www/js/opnsense_theme.js
```diff
@@ -51,7 +51,9 @@ $(document).ready(function () {
             var nextDiv = that.next('div');
 
             navigation.css('width', '415px');
-            if (nextDiv.not('in')) {
+            if (that.next('div').hasClass('in')) {
+                /* no action needed */
+            } else {
```
fichtner, line 56: Shorter is better?

```suggestion
            if (!that.next('div').hasClass('in')) {
```

opnsenseuser: ok

fichtner:
Merged, thanks!

opnsenseuser:
@fichtner you never stop learning. I tested it. It works perfectly. Thanks for the tip

fichtner:
@opnsenseuser after taking a closer look a5a109f9bf0b6c makes sense too. the variable is defined so the main if should use it too just in case the variable is modified later on.

PR #7788, opened 2024-08-18
Fix Data Length help text to reflect the current default value

doktornotor:
Changed in  https://github.com/opnsense/core/commit/044b0d9ad909b7b8ed85d57996693292d999a907


AdSchellevis:
@doktornotor nice catch, thanks

PR #7801, opened 2024-08-21
system: replace confirm() use for consistency

fichtner:


src/opnsense/www/js/opnsense_widget_manager.js
```diff
@@ -564,18 +565,38 @@ class WidgetManager  {
         return $panel;
     }
 
-    _restoreDefaults(confirmDialog = true) {
-        if (confirmDialog) {
-            if (!confirm(this.gettext.restoreconfirm)) {
-                return;
-            }
-        }
+    __restoreDefaults(dialog) {
```
fichtner, line 568: only C programmers appreciate this ;)

src/opnsense/www/js/opnsense_widget_manager.js
```diff
@@ -322,6 +322,7 @@ class WidgetManager  {
                 message: $content,
                 buttons: [{
                     label: this.gettext.add,
+                    cssClass: 'btn-primary',
```
fichtner, line 325: in theory this is nicer but primary color for header and button doesn't match which is annoying. we could also pivot to "warning" dialogs here? not sure if "yellow" is better.. at least better than "red" (danger)

fichtner: I removed the primary hint in the modals so this is now just a consistency fix for the `confirm()` replacement.

swhite2 (approved):

src/opnsense/www/js/opnsense_widget_manager.js
```diff
@@ -564,18 +565,38 @@ class WidgetManager  {
...
+        BootstrapDialog.show({
+            title: this.gettext.restore,
+            draggable: true,
+            animate: false,
+            message: this.gettext.restoreconfirm,
+            buttons: [{
+                label: this.gettext.ok,
+                cssClass: 'btn-primary',
+                hotkey: 13,
+                action: async (dialog) => {
```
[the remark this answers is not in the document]
fichtner: done, works fine

PR #7802, opened 2024-08-21
Update MenuSystem.php

sfrings:
Hi guys,

I bet you meant 2015 instead of 2050 in your copyright information, right?


fichtner:
nice catch, that was stuck there since 2020 🥇

Issue #7301, opened 2024-03-06
duckduckgo.com -> external content not loading


thojo0:
**Describe the bug**

If you use UnboundDNS in OPNsense by default `external-content.duckduckgo.com` always returns 404.
This results in not loading images, etc.
This makes complete sense because in `/usr/local/etc/unbound.opnsense.d/safesearch.conf` the local-zone is set to redirect, but `save.duckduckgo.com` is not the same server as `external-content.duckduckgo.com`.
I temporary fixed it by adding a new conf file with content:
```
server:
local-zone: "external-content.duckduckgo.com" always_transparent
```
But I think a transparent zone would be better in the `safesearch.conf`.

**To Reproduce**

1. Install OPNsense
2. Configure UnboundDNS
3. Set DNS server to OPNsense Ip
4. Search on `duckduckgo.com` and go to the images tab

**Expected behavior**

Images load

AdSchellevis:
this https://github.com/opnsense/core/commit/64ec6dec9a8c2b007558945ef6a373ae05eeabb8 should be it then, right?

ajkblue:
@AdSchellevis looks correct to me 👍

PR #7815, opened 2024-08-26
Services: Unbound DNS: Blocklist - CNAME and A record on query fix

thojo0:
With the current zone settings, Unbound returns both, the A and CNAME (to it self) record on different safe search subdomains.

Affected subdomains:
- `safe.duckduckgo.com`
- `strict.bing.com`
- `safesearch.pixabay.com`
- `safeapi.qwant.com`

This commit fixes this issue.
I also checked this on official documentations to be as accurate as possible, so nothing else breaks again.

### DIFF
--- a/src/opnsense/service/templates/OPNsense/Unbound/core/safesearch.conf
+++ b/src/opnsense/service/templates/OPNsense/Unbound/core/safesearch.conf
@@ -59,15 +59,14 @@ local-data: "www.{{ domain }} CNAME forcesafesearch.google.com"
 {%   endfor%}
 
 # DuckDuckGo
-local-zone: "duckduckgo.com" redirect
+local-zone: "duckduckgo.com" transparent
 local-data: "duckduckgo.com CNAME safe.duckduckgo.com"
 local-zone: "duck.com" redirect
 local-data: "duck.com CNAME safe.duckduckgo.com"
-local-zone: "external-content.duckduckgo.com" always_transparent
 
 # Bing
-local-zone: "bing.com" redirect
-local-data: "bing.com CNAME strict.bing.com"
+local-zone: "www.bing.com" redirect
+local-data: "www.bing.com CNAME strict.bing.com"
 
 # YouTube
 local-zone: "www.youtube.com" redirect
@@ -82,10 +81,10 @@ local-zone: "www.youtube-nocookie.com" redirect
 local-data: "www.youtube-nocookie.com CNAME restrictmoderate.youtube.com"
 
 # Pixabay
-local-zone: "pixabay.com" redirect
+local-zone: "pixabay.com" transparent
 local-data: "pixabay.com CNAME safesearch.pixabay.com"
 
 # Qwant
-local-zone: "qwant.com" redirect
-local-data: "qwant.com CNAME safeapi.qwant.com"
+local-zone: "api.qwant.com" redirect
+local-data: "api.qwant.com CNAME safeapi.qwant.com"
 {% endif %}
--- a/src/opnsense/service/templates/OPNsense/Unbound/core/safesearch.conf
+++ b/src/opnsense/service/templates/OPNsense/Unbound/core/safesearch.conf
@@ -61,8 +61,6 @@ local-data: "www.{{ domain }} CNAME forcesafesearch.google.com"
 # DuckDuckGo
 local-zone: "duckduckgo.com" transparent
 local-data: "duckduckgo.com CNAME safe.duckduckgo.com"
-local-zone: "duck.com" redirect
-local-data: "duck.com CNAME safe.duckduckgo.com"
 
 # Bing
 local-zone: "www.bing.com" redirect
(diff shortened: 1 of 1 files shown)


AdSchellevis:
I don't mind merging, but can you share the documentation that you are referring to?

thojo0:
# DuckDuckGo

https://duckduckgo.com/duckduckgo-help-pages/features/safe-search/

> For network administrators, you can force strict safe search for everyone on your network by mapping duckduckgo.com to safe.duckduckgo.com. Mapping to safe.duckduckgo.com will guarantee that safe search is enabled for all DuckDuckGo queries on the network, and that client safe search controls are disabled.

# Bing

https://support.microsoft.com/en-us/topic/blocking-adult-content-with-safesearch-or-blocking-chat-946059ed-992b-46a0-944a-28e8fb8f1814

> At a network level, map www.bing.com to strict.bing.com.

# Pixabay

https://pixabay.com/blog/posts/block-adult-content-on-pixabay-at-your-school-or-w-140/

> Set the DNS entry for pixabay.com to be a CNAME for safesearch.pixabay.com.

# Qwant

I didn't find an official docs/blog but because the same problem was there I used the same way like on the other ones.

AdSchellevis:
but this doesn't explain why we are changing the redirect to transparent in 

```
local-zone: "duckduckgo.com" transparent
```

thojo0:
Ah sorry, I meant I checked the exact domains again.
the transparent zone I pot there because of the CNAME+A record problem.

> With the current zone settings, Unbound returns both, the A and CNAME (to it self) record on different safe search subdomains.

After some tests, this was the best solution to fix it and also the problem mentioned in #7301 without an explicit "whitelisting".

src/opnsense/service/templates/OPNsense/Unbound/core/safesearch.conf
```diff
@@ -59,15 +59,14 @@ local-data: "www.{{ domain }} CNAME forcesafesearch.google.com"
 {%   endfor%}
 
 # DuckDuckGo
-local-zone: "duckduckgo.com" redirect
+local-zone: "duckduckgo.com" transparent
 local-data: "duckduckgo.com CNAME safe.duckduckgo.com"
 local-zone: "duck.com" redirect
 local-data: "duck.com CNAME safe.duckduckgo.com"
```
wetono, line 65:
```suggestion
```
Not necessary, because duck.com simply redirects to duckduckgo.com and then the "duckduckgo.com CNAME safe.duckduckgo.com" kicks in anyway.

src/opnsense/service/templates/OPNsense/Unbound/core/safesearch.conf
```diff
@@ -82,10 +81,10 @@ local-zone: "www.youtube-nocookie.com" redirect
 local-data: "www.youtube-nocookie.com CNAME restrictmoderate.youtube.com"
 
 # Pixabay
-local-zone: "pixabay.com" redirect
+local-zone: "pixabay.com" transparent
 local-data: "pixabay.com CNAME safesearch.pixabay.com"
 
 # Qwant
```
wetono, line 87:
```suggestion
```
Safesearch on qwant isn't enabled/enforced for me, neither with the original config, nor with the change you recommend. I also couldn't find any documentation from qwant about enforcing safesearch, so maybe this feature has been removed? Because of that it can be removed imo.

src/opnsense/service/templates/OPNsense/Unbound/core/safesearch.conf
```diff
@@ -82,10 +81,10 @@ local-zone: "www.youtube-nocookie.com" redirect
...
-local-zone: "qwant.com" redirect
-local-data: "qwant.com CNAME safeapi.qwant.com"
+local-zone: "api.qwant.com" redirect
+local-data: "api.qwant.com CNAME safeapi.qwant.com"
```
wetono, line 89:
```suggestion
```
Safesearch on qwant isn't enabled/enforced for me, neither with the original config, nor with the change you recommend. I also couldn't find any documentation from qwant about enforcing safesearch, so maybe this feature has been removed? Because of that it can be removed imo.

wetono (changes requested):
Since I can't access duckduckgo with safesearch enabled because of the problem you described (I'm using a Windows DNS server because of AD and have the OPNsense as the upstreams DNS, when the Windows DNS server caches the duckduckgo.com entry it only gives the client the safe.duckduckgo.com CNAME safe.duckduckgo.com and not the A-Record, so the client can't access duckduckgo), I'm glad to see there's a pull request that fixes this! I did some testing and have two small suggestions for improvement, but otherwise it works as intended!

src/opnsense/service/templates/OPNsense/Unbound/core/safesearch.conf
thojo0: Yeah, I also think that makes sense, else it suggests, that it would/should work, when it doesn't.

PR #7827, opened 2024-08-30
system: handle stale "pfsyncinterfaces" and improve workflow

fichtner:
PR: https://forum.opnsense.org/index.php?topic=42549.0

I did not test this yet so for immediate review only.

### DIFF
--- a/src/opnsense/mvc/app/controllers/OPNsense/Core/forms/hasyncSettings.xml
+++ b/src/opnsense/mvc/app/controllers/OPNsense/Core/forms/hasyncSettings.xml
@@ -16,16 +16,10 @@
         <help>When this device is configured as CARP backup it will disconnect all PPP type interfaces and try to reconnect them when becoming master again.</help>
     </field>
     <field>
-        <id>hasync.pfsyncenabled</id>
-        <label>Synchronize States</label>
-        <type>checkbox</type>
-        <help>pfsync transfers state insertion, update, and deletion messages between firewalls.</help>
-    </field>
-    <field>
         <id>hasync.pfsyncinterface</id>
-        <label>Synchronize Interface</label>
+        <label>Synchronize all states via</label>
         <type>dropdown</type>
-        <help>If Synchronize States is enabled, it will utilize this interface for communication. Best choose a dedicated interface for this type of communication to prevent manipulation of states causing security issues.</help>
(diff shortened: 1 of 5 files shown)


src/opnsense/mvc/app/controllers/OPNsense/Core/forms/hasyncSettings.xml
```diff
@@ -15,17 +15,11 @@
...
-    <field>
-        <id>hasync.pfsyncenabled</id>
-        <label>Synchronize States</label>
-        <type>checkbox</type>
-        <help>pfsync transfers state insertion, update, and deletion messages between firewalls.</help>
-    </field>
     <field>
         <id>hasync.pfsyncinterface</id>
-        <label>Synchronize Interface</label>
+        <label>Synchronize States</label>
```
fichtner, line 20: @AdSchellevis the code is tested and ready. I'm just not sure here. It's better workflow but the explanation that pfsync interface is the transport interface, not the interface's states to be synced is very unclear. We know which is which but just reading it makes little sense. Something like "Synchronize all states via" seems to be the closest I can get it sound right in my head, but it's really not ideal still.

FWIW, I'll do a bit more refactoring like https://github.com/opnsense/core/commit/82397315c because it just makes sense using the correct naming also regarding function name and:

https://github.com/opnsense/core/blob/82397315c0db24e406b92dc985ce8b7948108812/src/opnsense/scripts/interfaces/reconfigure_pfsync.php#L33-L34

;)

AdSchellevis: @fichtner the label is a bit confusing in most cases indeed, it feels like "Synchronize all states via" is the most implicit phrase in explaining what the intention is. Maybe we should just opt for that.

AdSchellevis: o yes, and `interfaces_carp_setup()` is better renamed as well indeed, the current name doesn't make sense at all.

fichtner: ok done. will leave this review open till Monday but extract some of the renaming directly to master branch

PR #7830, opened 2024-08-31
Enable restrict noquery by default for ntpd

doktornotor:
See https://github.com/opnsense/core/issues/7832

### DIFF
--- a/src/etc/inc/plugins.inc.d/ntpd.inc
+++ b/src/etc/inc/plugins.inc.d/ntpd.inc
@@ -396,7 +396,7 @@ function ntpd_configure_do($verbose = false)
     if (empty($config['ntpd']['nomodify'])) { /*note: this one works backwards */
         $ntpaccess .= ' nomodify';
     }
-    if (empty($config['ntpd']['noquery'])) { /*note: this one works backwards */
+    if (empty($config['ntpd']['noquery2'])) { /*note: this one works backwards */
         $ntpaccess .= ' noquery';
     }
     if (empty($config['ntpd']['notrap'])) { /*note: this one works backwards */
--- a/src/etc/inc/plugins.inc.d/ntpd.inc
+++ b/src/etc/inc/plugins.inc.d/ntpd.inc
@@ -396,7 +396,7 @@ function ntpd_configure_do($verbose = false)
     if (empty($config['ntpd']['nomodify'])) { /*note: this one works backwards */
         $ntpaccess .= ' nomodify';
     }
-    if (!empty($config['ntpd']['noquery'])) {
+    if (empty($config['ntpd']['noquery'])) { /*note: this one works backwards */
         $ntpaccess .= ' noquery';
     }
     if (empty($config['ntpd']['notrap'])) { /*note: this one works backwards */
@@ -411,6 +411,10 @@ function ntpd_configure_do($verbose = false)
     }
     $ntpcfg .= "\nrestrict default {$ntpaccess}";
     $ntpcfg .= "\nrestrict -6 default {$ntpaccess}";
+    /* Do not break status_ntpd.php page with restrict noquery */
+    $ntpaccess = str_replace("noquery", "", $ntpaccess);
+    $ntpcfg .= "\nrestrict 127.0.0.1 {$ntpaccess}";
+    $ntpcfg .= "\nrestrict ::1 {$ntpaccess}";
     $ntpcfg .= "\n";
 
     /* A leapseconds file is really only useful if this clock is stratum 1 */
--- a/src/www/services_ntpd.php
+++ b/src/www/services_ntpd.php
@@ -99,7 +99,7 @@ if ($_SERVER['REQUEST_METHOD'] === 'GET') {
     }
 
     // swap fields, really stupid field usage which we are not going to change now....
-    foreach (array('kod', 'limited', 'nomodify', 'nopeer', 'notrap') as $fieldname) {
+    foreach (array('kod', 'limited', 'nomodify', 'nopeer', 'notrap', 'noquery') as $fieldname) {
         $pconfig[$fieldname] = empty($pconfig[$fieldname]);
     }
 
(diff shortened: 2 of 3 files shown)


fichtner:
@doktornotor looks good at first glance. let me know when it's ready

doktornotor:
Well, pretty much ready - sans any migration bits as noted on https://github.com/opnsense/core/issues/7832#issuecomment-2323236749

The "logic" is just really - WTH. 🤐

src/etc/inc/plugins.inc.d/ntpd.inc
```diff
@@ -396,7 +396,7 @@ function ntpd_configure_do($verbose = false)
     if (empty($config['ntpd']['nomodify'])) { /*note: this one works backwards */
         $ntpaccess .= ' nomodify';
     }
-    if (!empty($config['ntpd']['noquery'])) {
+    if (empty($config['ntpd']['noquery'])) { /*note: this one works backwards */
```
fichtner, line 399: On a vanilla 24.7.3 the checkbox on the GUI does what I expect it to do WRT ntpd.conf but something here is definitely not right.

src/www/services_ntpd.php
```diff
@@ -425,7 +425,7 @@ function removeRow() {
                       <input name="nomodify" type="checkbox" id="nomodify"<?=empty($pconfig['nomodify']) ? " checked=\"checked\"" : ""; ?> />
                       <?=gettext("Deny state modifications (i.e. run time configuration) by ntpq and ntpdc"); ?>
                       <br />
-                      <input name="noquery" type="checkbox" id="noquery"<?=!empty($pconfig['noquery']) ? " checked=\"checked\"" : ""; ?> />
+                      <input name="noquery" type="checkbox" id="noquery"<?=empty($pconfig['noquery']) ? " checked=\"checked\"" : ""; ?> />
```
fichtner, line 428: hmm I don't think we should reverse this? WRT the config render

src/www/services_ntpd.php
```diff
@@ -99,7 +99,7 @@
     }
 
     // swap fields, really stupid field usage which we are not going to change now....
-    foreach (array('kod', 'limited', 'nomodify', 'nopeer', 'notrap') as $fieldname) {
+    foreach (array('kod', 'limited', 'nomodify', 'nopeer', 'notrap', 'noquery') as $fieldname) {
```
fichtner, line 102: or am I missing this one? I mean I agree straightening out the GUI but we cannot change the config render for existing entries in user configs as that would flip their default (unless that is the intention here).

doktornotor: Note the missing migration note. If users have `noquery` currently set in config.xml, it should be unset via migration. Then these fields appear ticked and "enabled by disabling" 

That is exactly the braindead way all those other tripple-inverted enabled by disabling fields are (that `/*This one works backwards */` code in `ntpd.inc` is - and I did not want to make any special case for this one. Would confuse matters even further.

src/www/services_ntpd.php
doktornotor: These go along with the `// really stupid fields usage`  - if your head already hurts, I'm not surprised. Without this additional commit which I originally missed, it does not work properly.

doktornotor:
And yes, flipping the default is what's intended purpose of these commits here. I suspect pretty much everyone is unaware of the info leak. Also, this is sort of an DDoS amplification feature, as noted e.g in https://www.ncsc.gov.ie/emailsfrom/Shadowserver/DoS/NTP-Version/

src/etc/inc/plugins.inc.d/ntpd.inc
doktornotor: Yep, see my additional comment. It works properly now and to remain properly working after switching the defaults, users who enabled it need to have the configuration unset in config.xml and all of this stuff needs to go backwards to remain consistently inconsistent with the rest. 🤣Those who did not enable it should be ok without doing anything and without any need to do any migration for them and with sane new default. . If they want to re-enable the query "feature", then they need to change (untick) the checkbox after upgrade themselves (release notes item probably).

src/www/services_ntpd.php
fichtner: since we don't have a migration in place here can we do the unfathomable and call this field 'noquery2'?

src/etc/inc/plugins.inc.d/ntpd.inc
fichtner: perhaps the best course of action is park this on devel for eventual 25.1 inclusion with a "migration note" in the release notes that the old "noquery" (opt-in) is no longer in effect and we switched to a new default (opt-out)

src/etc/inc/plugins.inc.d/ntpd.inc
```diff
@@ -411,6 +411,10 @@ function ntpd_configure_do($verbose = false)
     }
     $ntpcfg .= "\nrestrict default {$ntpaccess}";
     $ntpcfg .= "\nrestrict -6 default {$ntpaccess}";
+    /* Do not break status_ntpd.php page with restrict noquery */
```
fichtner, line 414: FWIW, this is a good addition indeed

src/www/services_ntpd.php
doktornotor: And keep the rest of the _really stupid_, _backwards_ etc. as is in this PR? Tempting indeed. Just need to check whether there are not other references to the field elsewhere.

fichtner: Well, kill the old `noquery` with fire. We might get away with this during a minor change due to security concerns.

src/etc/inc/plugins.inc.d/ntpd.inc
```diff
@@ -396,7 +396,7 @@ function ntpd_configure_do($verbose = false)
     if (empty($config['ntpd']['nomodify'])) { /*note: this one works backwards */
         $ntpaccess .= ' nomodify';
     }
-    if (!empty($config['ntpd']['noquery'])) {
+    if (empty($config['ntpd']['noquery2'])) { /*note: this one works backwards */
```
fichtner, line 399: looking back this could probably be `query` for the sake of being entirely correct

fichtner: Not from the config end. Let me merge as is then and deal with that later.

fichtner:
Merged, thanks!

PR #7839, opened 2024-09-03
Use FontAwesome for menu entries

0xThiebaut:
Glyphicon is not compatible with Apple's [Lockdown Mode](https://support.apple.com/en-us/105120) resulting in broken icons on iOS. This PR switches the Firewall menu entry's icon over to Font Awesome, as is the case for the other menu entries.

image


fichtner:
The fa icon being ugly is my two humble cents.

AdSchellevis:
https://github.com/opnsense/core/commit/cc97dd579c572c6c628ac968896868bdb8548e8b

PR #7841, opened 2024-09-04
NTPD configuration cleanups

doktornotor:
Remove some options for `restrict source` template that make no sense there.. Also, do not use `noselect` with pools.

### DIFF
--- a/src/etc/inc/plugins.inc.d/ntpd.inc
+++ b/src/etc/inc/plugins.inc.d/ntpd.inc
@@ -330,6 +330,7 @@ function ntpd_configure_do($verbose = false)
     $noselect = isset($config['ntpd']['noselect']) ? explode(' ', $config['ntpd']['noselect']) : [];
     $prefer = isset($config['ntpd']['prefer']) ? explode(' ', $config['ntpd']['prefer']) : [];
     $iburst = isset($config['ntpd']['iburst']) ? explode(' ', $config['ntpd']['iburst']) : [];
+    $have_pools = false;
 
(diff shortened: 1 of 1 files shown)


doktornotor:
Incomplete, at least would like to make the `noselect` checkbox no-op in `services_ntpd.php` for pools in this round.

src/etc/inc/plugins.inc.d/ntpd.inc
```diff
@@ -330,13 +330,15 @@ function ntpd_configure_do($verbose = false)
     $noselect = isset($config['ntpd']['noselect']) ? explode(' ', $config['ntpd']['noselect']) : [];
     $prefer = isset($config['ntpd']['prefer']) ? explode(' ', $config['ntpd']['prefer']) : [];
     $iburst = isset($config['ntpd']['iburst']) ? explode(' ', $config['ntpd']['iburst']) : [];
+    $have_pools = false;
```
doktornotor, line 333: Needs to go to the `foreach()`, originally contemplated using it for something more global for https://github.com/opnsense/core/pull/7760  but my brain got hurt once again, notably for the reasons mentioned in https://github.com/opnsense/core/pull/7760#discussion_r1738053007

That PR will need to wait for later.

doktornotor:
Well, there goes your daily dose of crappy JS... 🙄 

At least it seems to do what's expected, for me at least. 😁

AdSchellevis:
eventually we need to rewrite ntpd to mvc, so in the meantime it's better to have a code freeze.

PR #7862, opened 2024-09-13
Add OpenVPN option for case insensitive Strict UserCN check

tsdogs:
Using an LDAP authentication like MS Active Directory permits the username to match independently from char case.
When generating certificates for openvpn on users, the certificate CN is generated matching the user original case set in AD (for example Mario.Rossi)
Right now if the "Strict User/CN Matching" is set and the user uses "mario.rossi" as the authentication user name, this check results in a failed authentication even though the password is correct.
Currently the only solution is either forcing the user to write the username matching the original case, or disabling this option.

This change adds an option on OpenVPN "Strict User/CN Matching" to be done also ignoring the char case of the certificate CN.


AdSchellevis:
this https://github.com/opnsense/core/commit/c345e01de27f196d69a275b33c223ee4f0fd6374 should do the trick, but only for instances. There's a code freeze on legacy components.

PR #7865, opened 2024-09-15
Add DHCPv4 Leases widget

wogglenet:
This adds a DHCPv4 Leases widget.

The widget has 4 items per row: an icon to indicate online/offline, an IP address, a hostname, and MAC. It is resizable and scrollable.

It is based on the code from the dyndns widget, which I have modified significantly. It works great for me, but please review since I haven't done javascript in 10 years.

### DIFF
--- a/src/opnsense/www/js/widgets/Leases.js
+++ b/src/opnsense/www/js/widgets/Leases.js
@@ -1,0 +1,129 @@
+/*
+ * Copyright (C) 2024 Wogglesoft
+ * Copyright (C) 2024 Deciso B.V.
+ * All rights reserved.
+ *
+ * Redistribution and use in source and binary forms, with or without
+ * modification, are permitted provided that the following conditions are met:
+ *
+ * 1. Redistributions of source code must retain the above copyright notice,
+ *    this list of conditions and the following disclaimer.
+ *
+ * 2. Redistributions in binary form must reproduce the above copyright
+ *    notice, this list of conditions and the following disclaimer in the
+ *    documentation and/or other materials provided with the distribution.
+ *
+ * THIS SOFTWARE IS PROVIDED ``AS IS'' AND ANY EXPRESS OR IMPLIED WARRANTIES,
+ * INCLUDING, BUT NOT LIMITED TO, THE IMPLIED WARRANTIES OF MERCHANTABILITY
+ * AND FITNESS FOR A PARTICULAR PURPOSE ARE DISCLAIMED. IN NO EVENT SHALL THE
+ * AUTHOR BE LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY,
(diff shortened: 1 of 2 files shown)


swhite2:
A lot of networks have to deal with a lot of leases, which can grow the widget extremely fast without any mechanism to organize/group/search. The Services -> ISC DHCPv4 -> Leases page already offers all of this functionality so it doesn't make much sense to duplicate this to a widget.

In my humble opinion it would make more sense to display information which immediately tells you something about the state of the system; e.g. the number of online/offline clients (inactive included/excluded) etc.

If there are other things we would like to know that aren't included by default in the controller, it may be an idea to add a separate endpoint for this.

FWIW, https://github.com/opnsense/core/commit/9f565344b59d7c60a5e57b4177ef31d0fcfdef90 also introduced breakout link buttons which positions all of the aforementioned functionality a single click away.

wogglenet:
The beauty of the widget system is that it is optional and configurable - in this case this widget can be any size you want, and the data is scrollable inside it. Having the widget included allows people who want it to use it - you are not forced to use it. There's very little downside to including something like this. For this reason I think it makes sense to have as many widgets as possible available to add, so that the end user is not faced with the frustrating situation of wanting something that is completely unavailable or difficult to access or requires manually adding files.

The widget only shows non-expired leases. It shows if they are online or offline. It is valuable information at-a-glance.

By comparison, the full page is much more to sort through.

The information is very helpful if you are adding devices to the network which you then need to know the IP/mac of but the devices themselves do not report this information.

fichtner:
Just by your argument having it searchable via text input would make more sense than showing a blob of info verbatim.

wogglenet:
I will see if I can add that functionality, although that would make it awfully similar to the full page

fichtner:
Which is the original point. 😉

I’m all for these types of additions but the scope needs to be clear and complete, because otherwise people just ask for more and when you’re not around someone else will have to deal with it because requests will keep coming and eventually it needs maintenance.

swhite2:
> The beauty of the widget system is that it is optional and configurable - in this case this widget can be any size you want, and the data is scrollable inside it.

> The information is very helpful if you are adding devices to the network which you then need to know the IP/mac of but the devices themselves do not report this information. To be blunt, the lack of this dashboard functionality nearly caused me to abandon opnsense after initial configuration.

That does not mean it is free in terms of processing or rendering, nor does it help you find anything specific any faster than referring to the original page will (which, again, is a single click away). Instead, showing the newest N leases makes much more sense in that case.

> The widget only shows non-expired leases. It shows if they are online or offline. It is valuable information at-a-glance.

This can and will be in the order of thousands for any decent sized network and can easily be conveyed in a limited set instead.

> By comparison, the full page is much more to sort through.

But it is sortable, searchable and groupable by interface.

I also am certainly not against this addition, but I'm against rendering dense amounts of information which is easily managed in the system elsewhere.

AdSchellevis:
isc will be deprecated, dnsmasq already has a widget.

PR #7867, opened 2024-09-16
system: make dpinger_configure_do map aware

swhite2:
Same as https://github.com/opnsense/core/commit/a85bd197c2d2179d80b9f528ddda5f8799351748 really, this callout also seems to gobble up a lot of unnecessary time on bigger systems.

### DIFF
--- a/src/etc/inc/plugins.inc.d/dpinger.inc
+++ b/src/etc/inc/plugins.inc.d/dpinger.inc
@@ -69,7 +69,7 @@ function dpinger_services()
 function dpinger_configure()
 {
     return [
-        'monitor' => ['dpinger_configure_do:3'],
+        'monitor' => ['dpinger_configure_do:2'],
     ];
 }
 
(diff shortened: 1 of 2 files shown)


src/etc/inc/plugins.inc.d/dpinger.inc
```diff
@@ -69,7 +69,7 @@ function dpinger_services()
 function dpinger_configure()
 {
     return [
-        'monitor' => ['dpinger_configure_do:3'],
+        'monitor' => ['dpinger_configure_do:2'],
```
fichtner, line 72: nice catch, remnant from the $bootup flag :)

fichtner (approved):

src/etc/inc/plugins.inc.d/dpinger.inc
swhite2: :)

PR #7870, opened 2024-09-17
mvc: Remove bold text from base_tabs_header.volt

Monviech:
This change makes the tabs that use the `base_tabs_header.volt` `layout_partial` more in line with all the other tabs in the GUI that do not use bold text.


fichtner (approved):
ok after a bit discussion and research into the inconsistencies here

PR #7876, opened 2024-09-20
VPN: IPsec: Mobile Clients - explicit split-include - charon attribute 

falkevik:
Enabling more than one split network to be used with both vpnc and macos cisco vpn. Phase2 networks are set to 0.0.0.0/0 but the split-include is explicit.

Manual SPDs couldn't be added with mobile clients.
Using the IPsec Phase2 for more than one network didn't work as the SPD for the first network was the only one setup.
Rendering the subsequent networks in the split-include to be dropped on the way back to the vpn client IP address.

### DIFF
--- a/src/etc/inc/plugins.inc.d/ipsec.inc
+++ b/src/etc/inc/plugins.inc.d/ipsec.inc
@@ -964,7 +964,11 @@ function ipsec_write_strongswan_conf()
         if (!empty($net_list)) {
             $net_list_str = implode(",", $net_list);
             $strongswanTree['charon']['plugins']['attr']['subnet'] = $net_list_str;
-            $strongswanTree['charon']['plugins']['attr']['split-include'] = $net_list_str;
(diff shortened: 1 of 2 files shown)


AdSchellevis:
Eventually we should migrate this to MVC, the latest versions already contains "VPN: IPsec: Advanced Settings", which is likely also the spot where most of these settings should migrate to in some form. 

Keeping this here for now to prevent complicating next steps more than needed.

falkevik:
Ok, understood. Thanks for considering the additional configuration option.
For us we couldn't find any other way to solve the problem with multiple split networks.

I can have a look at adding this to the MVC pattern once that is applicable.

AdSchellevis:
@falkevik I'll keep it on the list, when refactoring code, it shouldn't be an issue to add it as well.

AdSchellevis:
should be implemented in https://github.com/opnsense/core/pull/8380

Issue #7878, opened 2024-09-21
UpdateOnlyTextField is not compatible with DependConstraint


kumy:
**Important notices**

Before you add a new report, we ask you kindly to acknowledge the following:

- [x] I have read the contributing guide lines at https://github.com/opnsense/core/blob/master/CONTRIBUTING.md
- [x] I am convinced that my issue is new after having checked both open and closed issues at https://github.com/opnsense/core/issues?q=is%3Aissue

**Describe the bug**

Using `DependConstraint` with `UpdateOnlyTextField` fields always trigger the validation message

**To Reproduce**
In a custom plugin…

Given such xml (simplified) model, the `aws_secret_access_key` field validation always trigger an error.
```
<model>
    <mount>//foo/bar</mount>
    <description>Foobar</description>
    <items>
        <enabled type="BooleanField">
            <default>0</default>
            <Required>Y</Required>
            <Constraints>
                <check003>
                    <reference>aws_secret_access_key.check001</reference>
                </check003>
            </Constraints>
        </enabled>
        <aws_secret_access_key type="UpdateOnlyTextField">
            <Required>N</Required>
            <Constraints>
                <check001>
                    <ValidationMessage>An AWS_SECRET_ACCESS_KEY is required.</ValidationMessage>
                    <type>DependConstraint</type>
                    <addFields>
                        <field1>enabled</field1>
                    </addFields>
                </check001>
            </Constraints>
        </aws_secret_access_key>
    </items>
</model>
```

**Expected behavior**

Should not trigger an error when the user enter a value, and should not trigger error on further saves

**Describe alternatives you considered**

Switching the field type to `TextField` don't trigger intepestive error.

**Screenshots**
We can see that a value is really posted.
![image](https://github.com/user-attachments/assets/7b77f128-abcd-4fb6-a26e-d6e129b8037c)
![image](https://github.com/user-attachments/assets/2bc60226-7054-4e03-8175-7321ce40cca5)

**Relevant log files**

N/A

**Additional context**

[...]

PR #7879, opened 2024-09-21
fix: UpdateOnlyTextField incompatibility with DependConstraint

kumy:
Closes: #7878

- create a public getter for accessing field current value
- use the getter in BaseConstraint check

@AdSchellevis is it what you expected? Is there some unit tests to run/adjust? Do you think of any other place the same change should happen?


AdSchellevis:
@kumy that's probably part of the solution, only issue is more constraints use `(string)x` which would suffer from the same issue as this one.  For example:

https://github.com/opnsense/core/blob/a69b5a9408564686628cb85a5bcb0b99a85da231/src/opnsense/mvc/app/models/OPNsense/Base/Constraints/UniqueConstraint.php#L54

src/opnsense/mvc/app/models/OPNsense/Base/Constraints/ComparedToFieldConstraint.php
```diff
@@ -61,16 +61,16 @@ public function validate($validator, $attribute): bool
...
                 return true;
             }
 
             if (
                 !$this->is_constraint_fulfilled(
                     $operator,
-                    floatval((string)$node),
-                    floatval((string)$other_node_content)
+                    floatval((string)$node->getCurrentValue()),
+                    floatval((string)$other_node_content->getCurrentValue())
```
kumy, line 73: Should we create a dedicated `asFloat()` property in `BaseConstraint`?

kumy: I did it in a further commit. feel free to ask a revert if you don't want it

src/opnsense/mvc/app/models/OPNsense/Base/Constraints/AllOrNoneConstraint.php
```diff
@@ -53,7 +53,7 @@ public function validate($validator, $attribute): bool
             $countEmpty = 0;
             $countNotEmpty = 0;
             foreach ($Fields as $fieldname) {
-                if (empty((string)$parentNode->$fieldname)) {
+                if ($this->isEmpty($parentNode->$fieldname)) {
```
kumy, line 56: I'm sorry I'm a little confused about this one. What kind of object do we get by `$parentNode->$fieldname`?

src/opnsense/mvc/app/models/OPNsense/Base/Constraints/SingleSelectConstraint.php
```diff
@@ -53,7 +53,7 @@ public function validate($validator, $attribute): bool
             $Fields = array_unique(array_merge(array($nodeName), $this->getOptionValueList('addFields')));
             $countNotEmpty = 0;
             foreach ($Fields as $fieldname) {
-                if (!empty((string)$parentNode->$fieldname)) {
+                if (!$this->isEmpty($parentNode->$fieldname)) {
```
kumy, line 56: same

src/opnsense/mvc/app/models/OPNsense/Base/Constraints/BaseConstraint.php
```diff
@@ -43,7 +43,7 @@ public function isEmpty($node)
         $node_class = get_class($node);
         if ($node_class == "OPNsense\Base\FieldTypes\BooleanField") {
             return empty((string)$node);
-        } elseif (empty((string)$node) || (string)$node == "0") {
+        } elseif (empty((string)$node->getCurrentValue()) || (string)$node->getCurrentValue() == "0") {
             return true;
         }
         return false;
```
kumy, line 49: I BTW would be tempted to move this function to `BaseField`, WDYT?

Function usage:
image

kumy:
@AdSchellevis I've found potential other places. Please double-check :pray: (new commits are untested sorry)

I'm also confused by a code part, please see my [comment in code](https://github.com/opnsense/core/pull/7879/files/b5cef2598b2efd644f9de597b7c27983c0b75311#diff-83b1386f39d4edc3287f6c60e97b8311c848e33e80b8bea0b2c006e5689133ee)

src/opnsense/mvc/app/models/OPNsense/Base/Constraints/BaseConstraint.php
AdSchellevis: I agree, but that means that `BooleanField` has a different implementation, which might even be cleaner anyway.

AdSchellevis (changes requested):

AdSchellevis:
@kumy left some notes, but I like the direction this is going. Nice work.

src/opnsense/mvc/app/models/OPNsense/Base/FieldTypes/BaseField.php
```diff
@@ -472,7 +499,25 @@ public function getChild($name)
...
+    public function isEmpty(): bool
+    {
+        return empty($this->getCurrentValue()); // "0" is already considered false by `empty()`
+    }
+
+    /**
+     * check if current value is NOT empty  (either boolean field as false or an empty field)
+     * @return bool
+     */
+    public function isNotEmpty(): bool
```
fichtner, line 515: FWIW, I’m not a fan of inversions such as this one being anchored as a separate function. Since the logic of isEmpty can be overridden I don’t see how isNotEmpty would not adhere to !isEmpty ever. If anything it leaves room for future mistakes. Otherwise I’m quite happy with this effort. Thanks!

kumy: @fichtner hum :thinking: ok then I can remove `isEmpty()` and keep only `isNotEmpty()` as this is the key to use the [Nullsafe methods and properties](https://www.php.net/manual/en/language.oop5.basic.php#language.oop5.basic.nullsafe) from PHP 8.0.0 and dealing with `null` value at the same time. eg like in

https://github.com/opnsense/core/pull/7879/files#diff-83b1386f39d4edc3287f6c60e97b8311c848e33e80b8bea0b2c006e5689133eeR56-R59


---
Sample test code for the logic
```php
<?php

class Test {
  var $foo=null;
}

class Test2 {
  var $bar=null;
  function isNotEmpty() {
    return !empty(trim((string)$this->bar));
  }
}

function dotest($f) {
  #var_dump($f->foo?->bar);
  #var_dump($f->foo?->isNotEmpty());

  if ($f->foo?->isNotEmpty()) {
    return 'countNotEmpty';
  }
  return 'countEmpty';
}

$t = new Test();
$t->foo = null;
var_dump(assert(dotest($t) == 'countEmpty'));

# expect: countEmpty
$t->foo = new Test2();
var_dump(assert(dotest($t) == 'countEmpty'));

# expect: countEmpty
$t->foo->bar = "";
var_dump(assert(dotest($t) == 'countEmpty'));

# expect: countEmpty
$t->foo->bar = "   ";
var_dump(assert(dotest($t) == 'countEmpty'));

# expect: countNotEmpty
$t->foo->bar = "yallah";
var_dump(assert(dotest($t) == 'countNotEmpty'));


if (null) {
  throw new Exception("SHOULD NOT HAPPEN");
}
```

kumy: New commit pushed for the `isEmpty()` removal

fichtner: I’m sorry for causing this change here. We would prefer to keep the isEmpty because !isNotEmpty is going to be much harder to read.

kumy: > much harder to read

agree that was why I defined both. moving to `isEmpty()` would imply to change 
```php
if ($parentNode->$fieldname?->isNotEmpty()) {
    $countNotEmpty++;
} else {
    $countEmpty++;
}
```
to something like
```php
if (is_null($parentNode->$fieldname) || $parentNode->$fieldname->isEmpty()) {
    $countEmpty++;
} else {
    $countNotEmpty++;
}
```

A bit longer, but probably easier to read for everyone :smirk: 
go for that?

kumy: @fichtner done with latest commit...

(we should not forget to squash all commits once we all agree on the code)

src/opnsense/mvc/app/models/OPNsense/Base/Constraints/ComparedToFieldConstraint.php
```diff
@@ -54,23 +54,23 @@ public function validate($validator, $attribute): bool
...
+        if ($node && !($node->isEmpty() || empty($operator) || empty($field_name))) {
             $parent_node = $node->getParentNode();
             $other_node_content = $parent_node->$field_name;
 
             // if the other field is not set, or invalid type -> ignore this constraint
             if (
-                $this->isEmpty($other_node_content) ||
-                    !is_numeric((string)$node) && !is_numeric((string)$other_node_content)
+                is_null($other_node_content) || $other_node_content->isEmpty() ||
+                !$node->isNumeric() && !$other_node_content->isNumeric()
```
kumy, line 64: hum, probably `!$other_node_content->isNumeric()` may have issue when `$other_node_content` is `null`. Wondering what would happen with the `&&` in the middle here

fichtner: It’s a bit of a trick question. Null is not numeric, but it’s certainly empty. But ‘0’ is numeric, but still empty. The practicality of the checks either needs to be guarded or we need earlier checks I discard handling invalid values or just assume null is either ‘’ or 0

src/opnsense/mvc/app/models/OPNsense/Base/FieldTypes/BaseField.php
fichtner: I think we never used the `?->` operator before and there are certainly some risk involved in using that in conditional statements

src/opnsense/mvc/app/models/OPNsense/Base/Constraints/ComparedToFieldConstraint.php
kumy: The more I read it the more I'm confused :exploding_head:  . I'm now thinking I spoke too fast :expressionless: .

If you think it should be changed, would you be kind to provide a bit of code for the required change :pray:

src/opnsense/mvc/app/models/OPNsense/Base/Constraints/BaseConstraint.php
kumy: done

src/opnsense/mvc/app/models/OPNsense/Base/Constraints/AllOrNoneConstraint.php
kumy: thanks

src/opnsense/mvc/app/models/OPNsense/Base/FieldTypes/BaseField.php
```diff
@@ -481,13 +518,9 @@ public function isRequired()
...
      */
-    public function isEmptyAndRequired()
+    public function isEmptyAndRequired(): bool
     {
-        if ($this->internalIsRequired && ($this->internalValue == "" || $this->internalValue == null)) {
-            return true;
-        } else {
-            return false;
-        }
+        return $this->internalIsRequired && $this->isEmpty();
```
AdSchellevis, line 523: this will be a bit tricky, when we're going to overwrite `isEmpty` in the boolean field, this will lead to migration issues as the empty intended in this scope is actually "not used" or not provided.

src/opnsense/mvc/app/models/OPNsense/Base/Constraints/ComparedToFieldConstraint.php
fichtner: I'm not sure. Just throwing ideas out at the moment. :) Could we maybe start with merging a new test case as a separate PR first so we can add the visibility for it being broken?

src/opnsense/mvc/app/models/OPNsense/Base/FieldTypes/BaseField.php
kumy: I don't really get what you mean. maybe you are referring to code that doesn't exist yet... I can revert that part if you prefer :shrug:

src/opnsense/mvc/app/models/OPNsense/Base/Constraints/ComparedToFieldConstraint.php
kumy: > new test case

Do you have Unit Tests :heart: ? Where?

src/opnsense/mvc/app/models/OPNsense/Base/FieldTypes/BaseField.php
kumy: hum, should be related to `0`and `false` beeing considered by the `empty()`.
Reverted the test :heavy_check_mark:

src/opnsense/mvc/app/models/OPNsense/Base/Constraints/ComparedToFieldConstraint.php
fichtner: Hmm, so in general:

    # pkg install ps-debug && cd /usr/core && make test

The actual code for the FieldTypes and Constraints in particular:

https://github.com/opnsense/core/tree/master/src/opnsense/mvc/tests/app/models/OPNsense/Base

src/opnsense/mvc/app/models/OPNsense/Base/FieldTypes/BaseField.php
AdSchellevis: No problem, I just realized I might have explained it the wrong way around anyway, by example:

We are moving the following block into the base field type, which was the starting point of this journey:

https://github.com/opnsense/core/blob/ec23ffc0ef65ee3fe1f25d23c9690b709edb384c/src/opnsense/mvc/app/models/OPNsense/Base/Constraints/BaseConstraint.php#L41-L47

Now, since "0" is also empty when the type isn't a boolean, we would prevent a migration from saving `0` when the default was `0`, due to:  

https://github.com/opnsense/core/blob/ec23ffc0ef65ee3fe1f25d23c9690b709edb384c/src/opnsense/mvc/app/models/OPNsense/Base/BaseModelMigration.php#L49-L51

The problem here is that "empty" in the constraint has a different meaning than empty in this case, which is also used to make sure unfilled fields are being processed.

Hope this explains the issue a bit, reverting this part of the change for this reason would indeed be better.

src/opnsense/mvc/app/models/OPNsense/Base/Constraints/ComparedToFieldConstraint.php
fichtner: sorry, small typo in previous, should have been "os-debug"

kumy: Really missed that folder thanks.

`pkg install os-debug` I'm on Linux, will try to get tests working in some way. Any specific doc I should read, https://github.com/opnsense/core/blob/master/CONTRIBUTING.md doesn't talk about that :shrug:

AdSchellevis: `phpunit --configuration PHPunit.xml` should do the trick on other platforms as well, you can find it in `src/opnsense/mvc/tests/PHPunit.xml`.

Out of sheer curiosity, what are you building?

kumy: Should I upgrade my opnsense VM to some `main` branch equivalent? how? Sorry for my silly questions :( 
I'm on `OPNsense 24.7.4_1-amd64`

image

src/opnsense/mvc/app/models/OPNsense/Base/FieldTypes/BaseField.php
kumy: Done :heavy_check_mark: 
image

src/opnsense/mvc/app/models/OPNsense/Base/Constraints/ComparedToFieldConstraint.php
fichtner: The tests expect some persistent installed state from the core package so it's only possible to run it in an opnsense VM at the moment. In theory you only need PHPUnit but since our tests run during a nightly build that is not a pressing thing to do (also because PHP versions may differ in results). 

Just for reference these are some debug tools the plugin pulls in https://github.com/opnsense/plugins/blob/b8699cde8cf210e864e4bfab7329dcaaf40ec995/devel/debug/Makefile#L4-L8

kumy: @AdSchellevis 
> phpunit --configuration PHPunit.xml

That made it!
image


> Out of sheer curiosity, what are you building?

I first manually installed [`restic`](https://restic.readthedocs.io/en/stable/030_preparing_a_new_repo.html) to backup my configuration, but found i need to create some action to make it scheduled and needed some script and password... This wasn't much pratical to restore in case of disaster recovery and having a plugin would be better. plus it would certainly help others

I'm currently having pages and configuration saving working
image
image
image
also templates and a basic script
image

I will next work on  creating cron/schedules like in acme plugin. And also wish to integrate it in some way to trigger an automatic backup after each config change.

kumy: Please find PR #7883 for the preliminary Test Case

src/opnsense/mvc/app/models/OPNsense/Base/Constraints/BaseConstraint.php
```diff
@@ -33,21 +33,6 @@
...
-    public function isEmpty($node)
-    {
-        $node_class = get_class($node);
-        if ($node_class == "OPNsense\Base\FieldTypes\BooleanField") {
-            return empty((string)$node);
-        } elseif (empty((string)$node) || (string)$node == "0") {
-            return true;
-        }
-        return false;
-    }
```
fichtner, line 36:
```suggestion
```

src/opnsense/mvc/app/models/OPNsense/Base/Constraints/ComparedToFieldConstraint.php
fichtner: So I still think `!$other_node_content->isNumeric()` may overlap with `$other_node_content->isEmpty()` in the "0" case. I'm not sure what the best way to untangle this is. But I admit I also don't know if that is actually an issue. @AdSchellevis what do you think?

src/opnsense/mvc/app/models/OPNsense/Base/FieldTypes/BaseField.php
```diff
@@ -363,6 +363,33 @@ public function __toString()
         return (string)$this->internalValue;
     }
 
+    /**
+     * return field current value
+     * @return null|string field current value
+     */
+    public function getCurrentValue(): ?string
+    {
+        return trim((string)$this->internalValue);
```
fichtner, line 372: I'm pretty sure trim() is not a good idea even if the impact is subtle. We had trim() in password processing before which isn't ideal for example.

src/opnsense/mvc/tests/app/models/OPNsense/Base/Constraints/ComparedToFieldConstraintTest.php
```diff
@@ -78,6 +78,55 @@ public function test_if_it_validates_number_ranges_correctly_with_gt_and_error()
         $this->assertEquals(true, $ret);
     }
 
+    // Empty values
```
fichtner, line 81: ah so I guess these are the tests I'm missing? :)

kumy: They are the tests from the other PR, they will be removed from this one after rebase.

src/opnsense/mvc/app/models/OPNsense/Base/FieldTypes/BaseField.php
kumy: no worries, let's remove it

src/opnsense/mvc/app/models/OPNsense/Base/Constraints/ComparedToFieldConstraint.php
kumy: @fichtner I think the case you are thinking of is `node=2`, `othernode=0` `operator=lt`, and you are expecting an error. (Current code in `master` don't raise any error message for that - neither "next" one)

Test would be:
```php
    public function test_if_it_validates_zero_number_ranges_correctly_with_lt_and_error()
    {
        $validator = new \OPNsense\Base\Validation();
        $validate = $this->make_validator(2, 0, 'test', 'lt');
        $ret = $validate->validate($validator, '');
        $messages = $validator->getMessages();
        $this->assertEquals(1, count($messages));
        $this->assertEquals(true, $ret);
    }
```

kumy: I pushed new commit https://github.com/opnsense/core/pull/7879/commits/1651ff82f10cc548e1a19f8efdd3dc5288fbad54 to fix that specific case. It introduces a new `isEmptyButZero()` function. Let me know what you think.

AdSchellevis:
@kumy after a couple of busy weeks, I thought to take another look at this. Looking at the code, I think it's mostly ready to merge, I'm just not sure about  the naming of `isEmptyButZero`,  how about `isEmptyString`? Ideally the `isEmpty` should be more context aware (boolean expects 0, string ''), but given the risk of regressions, that's not a valid path at the moment.

kumy:
Thanks! Ok I will rename `isEmptyButZero` to `isEmptyString`

> Ideally the `isEmpty` should be more context aware (boolean expects 0, string ''), but given the risk of regressions, that's not a valid path at the moment.

Agree, but such change should be separated from this PR 🙂

kumy:
@AdSchellevis Thanks for the review :nerd_face: 

- PR rebased
- function `isEmptyButZero` renamed to `isEmptyString`
- Unit Tests still passing

```
# phpunit --configuration PHPunit.xml 
PHPUnit 9.6.20 by Sebastian Bergmann and contributors.

Runtime:       PHP 8.2.23
Configuration: PHPunit.xml

...............................................................  63 / 219 ( 28%)
............................................................... 126 / 219 ( 57%)
............................................................... 189 / 219 ( 86%)
..............................                                  219 / 219 (100%)

Time: 00:00.139, Memory: 50.88 MB

OK (219 tests, 456 assertions)
```

---

BTW, if the `PHPunit.xml` file name is all lowercase then it is automatically picked by `phpunit`
```
# cd /usr/core/src/opnsense/mvc/tests

# phpunit
PHPUnit 9.6.20 by Sebastian Bergmann and contributors.

Usage:
  phpunit [options] UnitTest.php
  phpunit [options] <directory>

[…]

# mv PHPunit.xml phpunit.xml
# phpunit
PHPUnit 9.6.20 by Sebastian Bergmann and contributors.

Runtime:       PHP 8.2.23
Configuration: /usr/core/src/opnsense/mvc/tests/phpunit.xml

...............................................................  63 / 219 ( 28%)
............................................................... 126 / 219 ( 57%)
............................................................... 189 / 219 ( 86%)
..............................                                  219 / 219 (100%)

Time: 00:00.138, Memory: 50.88 MB

OK (219 tests, 456 assertions)
```

AdSchellevis:
@kumy not sure why I made it uppercase back in 2016, but if the default is lowercase, I don't mind changing that. Thanks for taking the time to finish this PR

kumy:
Thanks! :man_dancing:

> not sure why I made it uppercase back in 2016, but if the default is lowercase, I don't mind changing that.

If you want to change it, I let you do it - it'll probably be easier like that. I bet the "uppercase" one lives in many places/repositories...

Any chance you have a bit of time for the other ones?
- https://github.com/opnsense/plugins/issues/4259
- https://github.com/opnsense/tools/issues/434

Thanks

AdSchellevis:
renamed the phpunit file in https://github.com/opnsense/core/commit/84437b3812d959a7aca7f1af8d6c528d8683607d

It looks like xymon-client is not very big and seems to build without issues on my end, I've added it to the build (https://github.com/opnsense/tools/commit/88e547d70eefde02137660f313ba079db4aeae68) assuming @fichtner is ok with it since it builds on my end.

It has been pretty busy lately, so it probably slipped our attention.

PR #7883, opened 2024-09-22
tests: Add a unit test for ComparedToFieldConstraintTest

kumy:
Add unit tests for a specific case as we had [concerns](https://github.com/opnsense/core/pull/7879#discussion_r1770370436) with this specific code when `$other_node_content` is `null` (x-link #7879)

image

The tests `test_if_it_validates_constraint_if_other_is_null_and_no_error`/`test_if_it_validates_other_node_empty_values_correctly_with_gt_and_no_error` will specifically trigger that case.

/cc @fichtner :pray:

### DIFF
--- a/src/opnsense/mvc/tests/app/models/OPNsense/Base/Constraints/ComparedToFieldConstraintTest.php
+++ b/src/opnsense/mvc/tests/app/models/OPNsense/Base/Constraints/ComparedToFieldConstraintTest.php
@@ -78,6 +78,55 @@ class ComparedToFieldConstraintTest extends \PHPUnit\Framework\TestCase
         $this->assertEquals(true, $ret);
     }
 
+    // Empty values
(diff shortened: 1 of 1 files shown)


fichtner:
Merged, thanks!

fichtner:
@kumy finally back home so I  tried this and all tests pass. I thought you would add the test to it that would break under the current code?

fichtner:
@kumy sorry for the delay. To reiterate: I would expect a test case to fail because you raised #7879 so there has to be a test case that isn't covered and would break without the PR merged and fixed when the PR is merged?

kumy:
They are here. Do you wanna them in a distinct PR?

https://github.com/opnsense/core/pull/7879/files#diff-ddb4f140a20a4a3d87274cf0793601132f7fa90b69f2e494cbe39be49c597237R81-R118

fichtner:
Yes, please, that would be great.

PR #7889, opened 2024-09-23
System/Logging - handle lines with no timestamp

### LINKED ISSUE #7888: queryLog.py breaks trying to parse timestamp from multi-line log entries
`queryLog.py` fails in cases where there are log entries split across multiple lines - it tries to parse a timestamp from the addional lines. This results in the log being unviewable in the web UI. This can be triggered by Kea DHCP in 24.7.4 (at least), which produces entries like:

```
<132>1 2024-09-23T00:01:04+00:00 opnsense kea-dhcp4 63216 - [meta sequenceId="3"] WARN  [kea-dhcp4.alloc-engine.0x82d4d90d000] ALLOC_ENGINE_V4_DISCOVER_ADDRESS_CONFLICT [hwtype=1 68:39:43:d8:11:4a], cid=[01:00:00:00:00:00:00], tid=0x16e2f536: conflicting reservation for address 10.1.1.83 with existing lease Address:       10.1.1.83
Valid life:    4000
Cltt:          1727048850
Hardware addr: 68:39:43:d8:11:4a
Client id:     01:68:39:43:d8:11:4a
Subnet ID:     2
Pool ID:       0
State:         default
Relay ID:      (none)
Remote ID:     (none)
```

which fails like:

```
root@opnsense:~ # /usr/local/opnsense/scripts/syslog/queryLog.py --module 'core' --filename 'kea' --valid_from '1726930874.967'
  File "/usr/local/opnsense/scripts/syslog/queryLog.py", line 75, in <module>
    if valid_from and isoparse(record['timestamp']).timestamp() < valid_from:
                      ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.11/site-packages/dateutil/parser/isoparser.py", line 37, in func
    return f(self, str_in, *args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.11/site-packages/dateutil/parser/isoparser.py", line 134, in isoparse
    components, pos = self._parse_isodate(dt_str)
                      ^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.11/site-packages/dateutil/parser/isoparser.py", line 208, in _parse_isodate
    return self._parse_isodate_common(dt_str)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.11/site-packages/dateutil/parser/isoparser.py", line 213, in _parse_isodate_common
    len_str = len(dt_str)
              ^^^^^^^^^^^
TypeError: object of type 'NoneType' has no len()
root@opnsense:~ #
```

when attempting to parse:

`{'timestamp': None, 'parser': None, 'facility': 1, 'severity': None, 'process_name': '', 'pid': None, 'rnum': 4, 'line': 'Remote ID:     (none)'}`

Arguably kea should not be producing multi-line log entries, but `queryLog.py` should handle it more gracefully - a quick fix might be to just pass over "entries" (lines) that have no timestamp ("just keep swimming!") when looking for a stopping point.

dseven:
Recent feature addition to apply time constraint fails in cases where a log line does not contain a timestamp, as in the case of multi-line log entries. This change just moves on when a line contains no timestamp, as we will eventually find the line that contains the start of the log entry (and the timestamp).

fixes #7888

### DIFF
--- a/src/opnsense/scripts/syslog/queryLog.py
+++ b/src/opnsense/scripts/syslog/queryLog.py
@@ -72,7 +72,7 @@ if __name__ == '__main__':
                 break
             # exit when data found is older than provided valid_from
             try:
-                if valid_from and isoparse(record['timestamp']).timestamp() < valid_from:
+                if valid_from and record.get('timestamp') and isoparse(record['timestamp']).timestamp() < valid_from:
                     break
             except ValueError:
                 pass


fichtner:
@dseven looks reasonable to me, but leaving it for @swhite2 or @AdSchellevis since they worked on the code directly.

AdSchellevis:
@dseven thanks, should do the trick indeed. The other option would have been to catch `TypeError` as well, but the result is more or less the same.

PR #7908, opened 2024-09-25
firmware: remove escaped slashes workaround, works since 24.7.5

fichtner:
Parking this as a PR for next week.


fichtner:
24.7.5 is out, let's just do this now instead of forgetting it next week ;)

PR #7914, opened 2024-09-28
Update default tunables

cpalv:
Update two default tunables.  There are a few other `kern.elf*.alsr` that I think might be worthwhile, but I thought starting with these two first and getting some feedback would be best

### DIFF
--- a/src/etc/config.xml.sample
+++ b/src/etc/config.xml.sample
@@ -170,6 +170,16 @@
       <value>default</value>
     </item>
     <item>
+      <descr><![CDATA[Allow pages to be mapped simultaneously writable and executable]]></descr>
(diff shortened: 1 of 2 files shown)


AdSchellevis:
If I'm not mistaken these have quite a risk of introducing regressions, we shouldn't change these overnight and probably follow upstream for their defaults.

cpalv:
For what its worth I have these tunables running on two different OPNsense boxes with different hardware and have not noticed any system instability

fichtner:
Hmm, what’s the larger idea here? More security? Usually this will introduce subtle breakage in third party software and I would much rather follow FreeBSD defaults.

Cheers,
Franco

Issue #7911, opened 2024-09-26
Sensible to hide OTP seed in System > Access > Users > User Edit > OTP seed


andrew-ma:
**Important notices**

Before you add a new report, we ask you kindly to acknowledge the following:

- [X] I have read the contributing guide lines at https://github.com/opnsense/core/blob/master/CONTRIBUTING.md
- [X] I am convinced that my issue is new after having checked both open and closed issues at https://github.com/opnsense/core/issues?q=is%3Aissue

**Is your feature request related to a problem? Please describe.**

For the purpose of managing users, I think it is better to hide the OTP seed by default because it is sensitive information and is easily captured by screen recording/screen sharing.

I have to regenerate the OTP seed and re-add it to my OTP Authenticator App every time I 

**Describe the solution you like**

I think it would be sensible to hide the OTP seed either with: 
1) similar "Click to unhide" button that is used with the OTP QR code, or
2) make it hidden with asterisks "************" and displayed when a "Show Seed" checkbox next to the input field is toggled on.  And make it blank "" when a OTP seed is not generated yet.

**Describe alternatives you considered**

There are no alternatives.

**Additional context**

In System > Access > Users > User Edit > OTP seed

cpalv:
this seems pretty reasonable to me.  I can put something together an open a PR if there are no objections

fichtner:
I don't mind too much either way, but which security issue are we fixing? An admin allowed to edit all users on the system not being allowed to see the OTP seed he can click on to see anyway?


Cheers,
Franco

cpalv:
> is easily captured by screen recording/screen sharing. 

shoulder surfers?

fichtner:
Is that the technical term? What about network internals on an admin screen? Passwords on Slack? ;)

Well be my guest to raise a PR but take note that #7904 will rewrite the page for 25.1 anyway and additions may disappear for lack of core team time while converting the pages. The self-service Lobby: Password use case probably also needs fixing.

That being said in MVC we would have the "advanced" toggle which may be all you need here so it's hidden by default. It may just be better to wait this out.


Cheers,
Franco

cpalv:
Not sure I understand the "Lobby: Password" use case.  The sensitive information is already hidden by the "password" input type.

[...]

PR #7926, opened 2024-10-01
hide otp

cpalv:
#7911 hide the OTP seed after it has been generated.  Tested in a vm.  Not a big deal if this gets swept up in migration for #7904.  I'd be happy to help chip away on that

### DIFF
--- a/src/www/system_usermanager.php
+++ b/src/www/system_usermanager.php
@@ -434,6 +434,19 @@ $( document ).ready(function() {
         $('#otp_qrcode').show();
     });
 
+    // Show OTP
+    var otp_hidden = true;
+    $('#show_otp').on("click", function () {
+        if (otp_hidden) {
+            var hidden = document.getElementById("otp_seed").value;
+            document.getElementById("otp_seed_text").textContent = hidden;
+            otp_hidden = false;
+        } else {
+            document.getElementById("otp_seed_text").textContent = '****************';
+            otp_hidden = true;
+        }
+    });
+
     // remove user
     $(".act-del-user").click(function(event){
       var userid = $(this).data('userid');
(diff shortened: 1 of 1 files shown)


fichtner:
Wouldn't it be easier to move the OTP seed to the QR code unhide button by clustering the sections together?

src/www/system_usermanager.php
```diff
@@ -434,6 +434,19 @@ function get_user_privdesc(& $user)
         $('#otp_qrcode').show();
```
cpalv, line 434: cluster the sections together as in bundle the "unhide" action into this method??

fichtner: yes, so you don't have to manipulate the existing input / add the textarea

cpalv:
I tried to move the OTP seed into the otp_qrcode div

https://github.com/opnsense/core/blob/b49b935bb9f80e34358f13dc380e63e0a54eff27/src/www/system_usermanager.php#L933-L945

like so 

```
<tr>
    <td><a id="help_for_otp_code" href="#" class="showhelp"><i class="fa fa-info-circle"></i></a> <?= gettext('OTP QR code') ?></td>
     <td>
        <label class="btn btn-primary" id="otp_unhide"><?= gettext('Click to unhide') ?></label>
        <div style="display:none;" id="otp_qrcode">
           <td rowspan="3"><a id="help_for_otp_seed" href="#" class="showhelp" style="display;none;"><i class="fa fa-info-circle"></i></a> <?= gettext('OTP seed') ?></td>
            <td>
               <input id="otp_seed" name="otp_seed" value="<?=$pconfig['otp_seed'];?>"/>
               <small ><?= gettext('Generate new secret (160 bit)') ?></small>
                  <div class="hidden" data-for="help_for_otp_seed"><?=gettext("OTP (base32) seed to use when a one time password authenticator is used");?><br/></div>
            </td>
         </div>
         <script>
                $('#otp_qrcode').qrcode('<?= $otp_url ?>');
          </script>
          <div class="hidden" data-for="help_for_otp_code">
                    <?= gettext('Scan this QR code for easy setup with external apps.') ?>
           </div>
   </td>
</tr>
```

but i guess when `$('#otp_qrcode').qrcode()`  creates the QR code image and inserts a canvas into the same otp_qrcode div all of the OTP seed nodes get moved outside the div.

There is another problem.  When a new user is created the OTP field is gone entirely from the form and can no longer be generated.  The only way I see around this is to allow generating an OTP seed at user creation, but then the field would only appear IFF the user has an OTP seed at creation.

fichtner:
I’ll take a look on Monday. Thanks for looking into it.

AdSchellevis:
I think we can close this PR, our new usermanager (https://github.com/opnsense/core/pull/8046) hides the seed by default:

image

Issue #7953, opened 2024-10-07
OpenVPN: Add --ccd-exclusive option


swhite2:
**Important notices**

Before you add a new report, we ask you kindly to acknowledge the following:

- [x] I have read the contributing guide lines at https://github.com/opnsense/core/blob/master/CONTRIBUTING.md
- [x] I am convinced that my issue is new after having checked both open and closed issues at https://github.com/opnsense/core/issues?q=is%3Aissue

**Is your feature request related to a problem? Please describe.**

OpenVPN defines the `--ccd-exclusive` boolean that requires a connecting client to have a corresponding file in the `--client-config-dir` of the server for added security.

In theory it should be enough to define client-specific overrides, enable this option and observe that only clients with the same common name as the CSOs are allowed to connect.

**Describe the solution you like**

Just a simple checkbox in the Connections (server) form that allows to set this option.

**Describe alternatives you considered**

N/A

**Additional context**

N/A

Issue #4293, opened 2020-08-25
OpenVPN ccd-exclusive not working


Lightning1984:
Hi,

I'm currently trying to migrate from PFSense (Version 2.3.4)  to an OPNSense platform (20.7.1).
Unfortunately i have encountered an issue with OpenVPN which doesn't allow me finish migration.

In general I'm using certificate only based authentication + TLS authentication.
For additional security I'm using the "ccd-exclusive" directive which prevents any client without a "Client Specific Override" configured from connecting. 
On PFSense each entry on the "Client Specific Overrides" admin page creates a file in the relevant csc folder, i.e. "/var/etc/openvpn-csc/server2".
On OPNSense the csc folder visible in the OpenVPN config remains empty (/var/etc/openvpn-csc/4), instead AFAIK the overrides are dynamically created when a client connects by invoking some script: client-connect "/usr/local/etc/inc/plugins.inc.d/openvpn/ovpn_setup_cso.php server4"

Without "ccd-exclusive" configured everything works normal, the client can connect and also the client specific settings defined in the web gui for the client in question get pushed to the client (i.e. IP address, routes, DNS, etc.)

When ccd-exclusive directive is present in the OpenVPN configuration clients can no longer connect due to the following error:
Tue Aug 25 10:09:08 2020 a.b.c.d:10524 TLS Auth Error: --client-config-dir authentication failed for common name 'XXX-MOBILE' file='/var/etc/openvpn-csc/4/XXX-MOBILE'

When manually creating the file in the csc folder the connection works again, two lines show up in the log:
Tue Aug 25 11:01:58 2020 XXX-MOBILE/a.b.c.d:10462 OPTIONS IMPORT: reading client specific options from: /var/etc/openvpn-csc/4/XXX-MOBILE
Tue Aug 25 11:01:58 2020 XXX-MOBILE/a.b.c.d:10462 OPTIONS IMPORT: reading client specific options from: /tmp/openvpn_cc_6473f5202ef9d246c77bc1221621b4a.tmp

So it looks like the "client-connect" script is generating the file too late in the connection setup sequence.
Is there a way to revert back to the old file based Client Specific Overrides or any other trick to make ccd-exclusive work again.

regards
Rupert

kosli:
I have the need too to use the **ccd-exclusive** option, whereas obviously this would only work if the client specific configs are stored in the **client-config-dir**. 

[...]

PR #7956, opened 2024-10-09
OpenVPN: Instances: add Require Client Provisioning option

swhite2:
Fixes https://github.com/opnsense/core/issues/7953

`ccd-exclusive` doesn't work in our case, as explained in https://github.com/opnsense/core/issues/4293.

This is logical, as we use our own hook to dynamically generate the CSOs to allow for multiple authentication backends to dictate the contents of the CSO. Therefore, the `ccd-exclusive` option on it's own doesn't make much sense from the perspective of e.g. Radius, as there is no common name to match to and as such the original option would not add as much benefit in terms of security.

This PR matches both `common-name` and `tunnel_network`/`tunnel_networkv6`. In all cases an ipv4/ipv6 network must be provisioned with this option enabled.

### DIFF
--- a/src/opnsense/mvc/app/controllers/OPNsense/OpenVPN/forms/dialogInstance.xml
+++ b/src/opnsense/mvc/app/controllers/OPNsense/OpenVPN/forms/dialogInstance.xml
@@ -318,11 +318,8 @@ Set to 0 to disable, remember to change your client as well.
         <style>role role_server</style>
         <advanced>true</advanced>
         <help>
-        Require, as a condition of authentication, that a connecting client must be provisioned from an authentication backend.
-        If no authentication method or the local database has been selected, a client-specific override must be present
-        with at least the common name and a tunnel network specified (IPv4/IPv6). For other backends, the backend must
-        at least provide the client IP address. This is similar to OpenVPN's 'ccd-exclusive' option, but stricter
-        as the authentication backend must supply an IP address for the client.
+        Require, as a condition for authentication, that a tunnel address will be provisioned either from a local defined client-specific override or offered by an authenticator (such as RADIUS) .
+        This is similar to OpenVPN's 'ccd-exclusive' option, but stricter as we expect the client to receive a tunnel address for the protocol used.
         </help>
     </field>
     <field>
--- a/src/opnsense/mvc/app/controllers/OPNsense/OpenVPN/forms/dialogInstance.xml
+++ b/src/opnsense/mvc/app/controllers/OPNsense/OpenVPN/forms/dialogInstance.xml
@@ -312,6 +312,20 @@ Set to 0 to disable, remember to change your client as well.
         </help>
     </field>
     <field>
+        <id>instance.provision_exclusive</id>
+        <label>Require Client Provisioning</label>
+        <type>checkbox</type>
+        <style>role role_server</style>
+        <advanced>true</advanced>
+        <help>
+        Require, as a condition of authentication, that a connecting client must be provisioned from an authentication backend.
+        If no authentication method or the local database has been selected, a client-specific override must be present
+        with at least the common name and a tunnel network specified (IPv4/IPv6). For other backends, the backend must
+        at least provide the client IP address. This is similar to OpenVPN's 'ccd-exclusive' option, but stricter
+        as the authentication backend must supply an IP address for the client.
+        </help>
+    </field>
+    <field>
         <type>header</type>
         <label>Routing</label>
     </field>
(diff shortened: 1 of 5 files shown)


src/opnsense/mvc/app/controllers/OPNsense/OpenVPN/forms/dialogInstance.xml
```diff
@@ -311,6 +311,20 @@ Set to 0 to disable, remember to change your client as well.
...
+        <label>Require Client Provisioning</label>
+        <type>checkbox</type>
+        <style>role role_server</style>
+        <advanced>true</advanced>
+        <help>
+        Require, as a condition of authentication, that a connecting client must be provisioned from an authentication backend.
+        If no authentication method or the local database has been selected, a client-specific override must be present
+        with at least the common name and a tunnel network specified (IPv4/IPv6). For other backends, the backend must
+        at least provide the client IP address. This is similar to OpenVPN's 'ccd-exclusive' option, but stricter
+        as the authentication backend must supply an IP address for the client.
```
AdSchellevis, line 325: I think something like the following is more accurate:
```
Require, as a condition for authentication, that a tunnel address will be provisioned either from a local defined client-specific override or offered by an authenticator (such as RADIUS) .
This is similar to OpenVPN's 'ccd-exclusive' option, but stricter as we expect the client to receive a tunnel address for the protocol used.
```

src/opnsense/mvc/app/models/OPNsense/OpenVPN/OpenVPN.php
```diff
@@ -234,6 +235,29 @@ public function getOverwrite($server_id, $common_name)
...
+        }
+
+        // overlay is fed by authentication backends and takes precedence
+        $result = array_merge($result, $overlay);
+
+        // check if provisioning by authentication backend is mandatory
+        foreach ($this->Instances->Instance->iterateItems() as $node_uuid => $node) {
+            if (
+                (!empty((string)$node->enabled)) &&
+                ($server_id == $node_uuid) &&
```
AdSchellevis, line 249:
```suggestion
                $server_id == $node_uuid &&
```

src/opnsense/mvc/app/models/OPNsense/OpenVPN/OpenVPN.php
```diff
@@ -234,6 +235,29 @@ public function getOverwrite($server_id, $common_name)
...
+            $result['common_name'] = $common_name;
+        }
+
+        // overlay is fed by authentication backends and takes precedence
+        $result = array_merge($result, $overlay);
+
+        // check if provisioning by authentication backend is mandatory
+        foreach ($this->Instances->Instance->iterateItems() as $node_uuid => $node) {
+            if (
+                (!empty((string)$node->enabled)) &&
```
AdSchellevis, line 248:
```suggestion
                !empty((string)$node->enabled) &&
```

src/opnsense/mvc/app/models/OPNsense/OpenVPN/OpenVPN.php
```diff
@@ -234,6 +235,29 @@ public function getOverwrite($server_id, $common_name)
...
+                ($server_id == $node_uuid) &&
+                ((string)$node->role == 'server') &&
```
AdSchellevis, line 250:
```suggestion
                (string)$node->role == 'server' &&
```

src/opnsense/mvc/app/models/OPNsense/OpenVPN/OpenVPN.php
```diff
@@ -234,6 +235,29 @@ public function getOverwrite($server_id, $common_name)
...
+                (!empty((string)$node->provision_exclusive))
```
AdSchellevis, line 251:
```suggestion
                !empty((string)$node->provision_exclusive)
```

src/opnsense/mvc/app/models/OPNsense/OpenVPN/OpenVPN.xml
```diff
@@ -314,6 +314,10 @@
                 <auth-gen-token type="IntegerField">
                     <MinimumValue>0</MinimumValue>
                 </auth-gen-token>
+                <provision_exclusive type="BooleanField">
```
AdSchellevis, line 317: don't forget to update the version number :)

src/opnsense/mvc/app/models/OPNsense/OpenVPN/OpenVPN.php
```diff
@@ -170,9 +170,10 @@ public function performValidation($validateFullModel = false)
      * Retrieve overwrite content in legacy format
      * @param string $server_id vpnid
      * @param string $common_name certificate common name (or username when specified)
+     * @param array $overlay overwrite CSO properties
      * @return array legacy overwrite data
```
AdSchellevis, line 174:
```suggestion
     * @return array legacy overwrite data, empty when failed
```

AdSchellevis (changes requested):

PR #7961, opened 2024-10-11
Fix crt condition in setCRL

richierg:
Fix check empty condition before populate the variable.


fichtner (approved):
nice catch

AdSchellevis:
@richierg thanks!

Issue #7868, opened 2024-09-16
mvc: cleanup "this->sessionClose" since it is non optional now


Monviech:
**Important notices**

Before you add a new report, we ask you kindly to acknowledge the following:

- [x] I have read the contributing guide lines at https://github.com/opnsense/core/blob/master/CONTRIBUTING.md
- [x] I am convinced that my issue is new after having checked both open and closed issues at https://github.com/opnsense/core/issues?q=is%3Aissue

**Is your feature request related to a problem? Please describe.**

https://github.com/opnsense/core/blob/630d380728c7a7e4a87a82c9a53e17bafa890e26/src/opnsense/mvc/app/controllers/OPNsense/OpenVPN/Api/ServiceController.php#L221

**Describe the solution you like**

All of these instances can be cleaned up:
`$this->sessionClose()`

**Describe alternatives you considered**

Leaving everything as it is.

**Additional context**

https://github.com/opnsense/plugins/pull/4131

PR #7969, opened 2024-10-14
mvc: cleanup "this->sessionClose" since it is non optional now

Monviech:
WIP

Fixes: https://github.com/opnsense/core/issues/7868

One commit per changed file as requested by @fichtner


Monviech:
@fichtner Okay I think I got every spot.

PR #7981, opened 2024-10-18
Traffic ipv6 translation support

Ashus:
This adds a support for resolving IPv6 reverse addresses.
Second commit is a bit more controversial, but handy as well. Imagine you have static IPv6 prefixes delegated in your network. For that you need static IPv6 routes to properly configured home routers. This feature allows to translate a dynamically created IP via SLAAC into last part of IPv6 address@GatewayName. eg. **fc00:106:4:0:fc82:c2ff:ffff:a0cb** would be translated to **a0cb@ClientsName**.

### DIFF
--- a/src/opnsense/scripts/routes/ipv6_gateways.php
+++ b/src/opnsense/scripts/routes/ipv6_gateways.php
@@ -1,0 +1,21 @@
+#!/usr/local/bin/php
+<?php
+
+/*
+ * Returns static IPv6 routes with gateway names in TSV format.
+ * It is used for approximate IPv6 translation of dynamic addresses in traffic view (traffic_top.py).
+ */
+require_once 'config.inc';
+require_once 'util.inc';
+
+$allRoutes = get_staticroutes();
+$allSubnets = [];
+foreach ($allRoutes as $route) {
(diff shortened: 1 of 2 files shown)


AdSchellevis:
I don't mind taking a look at https://github.com/opnsense/core/pull/7981/commits/de690719fb1314b55d2ca74adb3a91f417de7aea, but the other commit looks to convoluted to me. can you open a PR for the reverse dns part only?

Ashus:
Here you go. Although I hope you give the second commit a second look later.

src/opnsense/scripts/routes/ipv6_gateways.php
```diff
@@ -0,0 +1,21 @@
+#!/usr/local/bin/php
```
fichtner, line 1: we would rather fold this special case return into an existing script. duplicating functionality for a single IP family is not ideal and we can easily build a plugin-based wrapper that can be called with `pluginctl`.

Ashus: Please point me in the right direction, where you want it, I'm not familiar with this.
I was trying to fetch the info on gateways directly from system, but gateway names are not there. So they are extracted from possibly non-applied config instead.

PR #7985, opened 2024-10-18
Traffic ipv6 translation support (clean)

Ashus:


src/opnsense/scripts/interfaces/traffic_top.py
```diff
@@ -94,15 +95,32 @@ async def request_ittr(self, addresses):
...
-            tasks.append(dnsResolver.resolve_address(address))
+            try:
+                if ":" in address:
+                    ipv6_obj = ipaddress.IPv6Address(address)
+                    reverse_ip = '.'.join(reversed(ipv6_obj.exploded.replace(":", ""))) + ".ip6.arpa"
+                else:
+                    reverse_ip = '.'.join(reversed(address.split("."))) + ".in-addr.arpa"
+
+                tasks.append(dnsResolver.resolve(reverse_ip, "PTR"))
+            except Exception as e:
```
AdSchellevis, line 106: better to catch explicit exception types, unexpected errors are better be thrown so you know what to look for later

src/opnsense/scripts/interfaces/traffic_top.py
```diff
@@ -94,15 +95,32 @@ async def request_ittr(self, addresses):
...
+                return
```
AdSchellevis, line 107: continue? the next address might be a valid one according to the loop

src/opnsense/scripts/interfaces/traffic_top.py
```diff
@@ -94,15 +95,32 @@ async def request_ittr(self, addresses):
         dnsResolver.timeout = 2
         tasks = []
         for address in addresses:
-            tasks.append(dnsResolver.resolve_address(address))
```
AdSchellevis, line 97: Why are you trying to re-implement the function of `resolve_address()` in the code below? According to the upstream documentation (https://dnspython.readthedocs.io/en/latest/resolver-class.html), `resolve_address()` should already return a valid response for both address families.

Can you offer some example addresses where the code below offers your expected outcome and `resolve_address()` doesn't?

AdSchellevis (changes requested):
We probably need test cases (examples) first to assess what your current issue is, looking in the upstream documentation of the library it should already work

src/opnsense/scripts/interfaces/traffic_top.py
Ashus: I don't really write in python, imo any exception here is acceptable to skip that address in the lookup process.

src/opnsense/scripts/interfaces/traffic_top.py
Ashus: Good catch. I fixed this in my next commit.

src/opnsense/scripts/interfaces/traffic_top.py
Ashus: I wasn't trying to re-implement it. But the following part was hard-coded to work only with IPv4 (.in-addr.arpa). My goal was only to allow translation for any IP address that comes up.
I was surprised to find that low-level translation to PTR there and only expanded on it. If there is a better solution, I'll gladly accept it.

Ashus:
Static local IPv6 addresses from **ISC DHCPv6** service were not translating and that was the initial reason to do any changes.

src/opnsense/scripts/interfaces/traffic_top.py
AdSchellevis: Please offer examples of addresses where your code offers different outcomes than the existing code, https://github.com/rthalley/dnspython/blob/740ae72d28e5bdadafbda80169eea255bbde16ac/dns/asyncresolver.py#L113-L139 suggests it does work on both, digging deeper in the code (https://github.com/rthalley/dnspython/blob/740ae72d28e5bdadafbda80169eea255bbde16ac/dns/reversename.py#L31-L66) doesn't appear to do anything else than support both either....

Ashus: I guess you mean why not use this: 

```python
    async def request_ittr(self, addresses):
        self._results = {}
        try:
            dnsResolver = Resolver()
        except dns.resolver.NoResolverConfiguration:
            return
        dnsResolver.timeout = 2
        tasks = []
        for address in addresses:
            tasks.append(dnsResolver.resolve_address(address))

        responses = await asyncio.gather(*tasks, return_exceptions=True)

        for response, address in zip(responses, addresses):
            if isinstance(response, dns.resolver.Answer):
                if ":" in address:
                    addr = str(ipaddress.IPv6Address(address))
                else:
                    addr = ".".join(reversed(response.canonical_name.to_text().replace('.in-addr.arpa.', '').split('.')))

                for item in response.response.answer:
                    if isinstance(item, dns.rrset.RRset) and len(item.items) > 0:
                        self._results[addr] = str(list(item.items)[0])

        return self._results

``` 

I tried it, and in console it works fine. But on the web it just loops with an empty JSON response.

AdSchellevis: with the risk of repeating myself, please offer example addresses for comparison.

Ashus: There isn't any difference that I know of.
I made a partial revert, would you like me to squash it?

AdSchellevis: This isn't going to be merged, these commits look highly suspicious, without test examples (which I repeatedly asked for), we will stop our efforts in these PR's.

PR #7986, opened 2024-10-18
Add SFP module details to interface details

Kreeblah:
I noticed that the interface details link (clicking on the magnifying glass on an interface in the Interfaces > Overview page) wasn't showing some of the information about the SFP transceivers I have plugged in to those ports.  It turns out that OPNsense is already grabbing that data (it's retrieved via `ifconfig -v`), but it just isn't parsing it when it's available.

So, this adds a few entries to the interface info in order to display those things.  In my case, these look like this:

field_name | field_value
-- | --
sfp_transceiver_plugged | SFP/SFP+/SFP28 Unknown (RJ45)
sfp_transceiver_vendor | FS
sfp_transceiver_part_number | SFP-10GM-T
sfp_transceiver_serial_number | SOMESERIALNUMBER
sfp_transceiver_manufacturing_date | 2000-01-31

In theory, there are also other things that could show up like module temperature or TX/RX strength (https://forums.freebsd.org/threads/network-card-sfp-module-status.68666/ has an example of this), but my modules don't report those, so I left them out since I wouldn't have a way of testing them.

### DIFF
--- a/src/etc/inc/interfaces.lib.inc
+++ b/src/etc/inc/interfaces.lib.inc
@@ -397,6 +397,13 @@ function legacy_interfaces_details($intf = null)
             $result[$current_interface]['media_raw'] = substr(trim($line), 7);
         } elseif (preg_match("/media (.*)/", $line, $matches)) {
             $result[$current_interface]["supported_media"][] = str_replace(" mediaopt ", " ", trim($matches[1]));
+        } elseif (preg_match("/plugged: (.*)/", $line, $matches)) {
+            $result[$current_interface]['sfp_transceiver_plugged'] = $matches[1];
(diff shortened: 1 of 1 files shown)


AdSchellevis:
This was still on my todo list after @swhite2 added the info in our `axgbe` driver.  Let me rearrange this a bit and add it to master.

AdSchellevis:
@Kreeblah can you double check https://github.com/opnsense/core/commit/cca972c05b22a28e2b076a4297cbe139ae7519af on your end? Thanks!

Kreeblah:
Yep, that commit looks like it's working for me.  I'm seeing a row labeled `sfp` with the parsed data as columns in it.  Thanks!

AdSchellevis:
@Kreeblah thank you for working on it

PR #8001, opened 2024-10-22
Unbound DNS: Overrides: Add note

mr-manuel:
Add a note for the user, until this is fixed in a better way.

### DIFF
--- a/src/opnsense/mvc/app/views/OPNsense/Unbound/overrides.volt
+++ b/src/opnsense/mvc/app/views/OPNsense/Unbound/overrides.volt
@@ -181,6 +181,10 @@ $( document ).ready(function() {
     #infosection {
         margin: 1em;
     }
+    #infosection_aliases {
+        margin: 1em;
+        padding: 0;
+    }
 </style>
 
 <ul class="nav nav-tabs" data-tabs="tabs" id="maintabs">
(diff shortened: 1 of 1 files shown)


fichtner:
Improving UX would be preferable over notes that people will not read wherever they are anyway.

In fact this comes up a few times a year, but given the amount of users I would not say it's in immediate need of a rewrite.


Cheers,
Franco

fichtner:
Nobody is disputing the UX situation. The issue I see is that every now and then somebody thinks this should be fixed, proposes something, nothing happens and then this happens again quite soon with or without a note.


Cheers,
Franco

mr-manuel:
Ok, so what could be the solution in your eyes? One we know the direction, I could try to fix that.

AdSchellevis:
We improved the situation a bit with implementing the table system, not a lot of issues have been reported in the meantime.

mr-manuel:
@AdSchellevis can you please at least add a column in the host table, so that it's visible, if a host has an alias or not? I have a lot of entries and only a few have aliases. Now I have to click one after another to check which host has aliases and which not.

AdSchellevis:
@mr-manuel extending the top view is likely not an issue, just open a ticket and I'll see what I can do when I have some time.

mr-manuel:
For reference: https://github.com/opnsense/core/issues/9540

PR #8007, opened 2024-10-23
Default checksum offloading for the vtnet(4) driver to disabled

pmhausen:
Most cloud providers based on KVM have either a bug or a common misconfiguration in their VirtIO network interface implementation which leads to FreeBSD falsely probing the interface as checksum offload capable while the emulated hardware in fact does not perform the offload.

The root cause of the problem is currently investigated by various developers in the FreeBSD community with support from at least Vultr and possibly Digitalocean last that I heard of the current state.

The simple and effective workaround is to disable checksum offloading globally for vtnet(4) interfaces.

I therefore argue this sysctl value should be the default for OPNsense installations.

See this forum post for more background and links:
https://forum.opnsense.org/index.php?topic=43583.msg216918#msg216918

I can of course refer you to people involved - or you simply join the bhyve production users call taking place every Thursday evening (european time).

I am not quite sure the proposed changes are everything that is necessary but I *am* fully convinced changing the tunable by default is the reasonable thing to do. The POLA violation when people get only a couple of 100 k/s of throughput far outweighs the minimal performance penalty with the tunable set to 1.

Kind regards,
Patrick

### DIFF
--- a/src/etc/config.xml.sample
+++ b/src/etc/config.xml.sample
@@ -135,11 +135,6 @@
       <value>default</value>
     </item>
     <item>
-      <descr><![CDATA[Disable receive and send checksum offload for vtnet(4). Recommended for most KVM based cloud providers.]]></descr>
-      <tunable>hw.vtnet.csum_disable</tunable>
-      <value>default</value>
-    </item>
-    <item>
       <descr><![CDATA[Enable TCP extended debugging]]></descr>
       <tunable>net.inet.tcp.log_debug</tunable>
       <value>default</value>
--- a/src/etc/config.xml.sample
+++ b/src/etc/config.xml.sample
@@ -135,6 +135,11 @@
       <value>default</value>
     </item>
     <item>
+      <descr><![CDATA[Disable receive and send checksum offload for vtnet(4). Recommended for most KVM based cloud providers.]]></descr>
+      <tunable>hw.vtnet.csum_disable</tunable>
+      <value>default</value>
+    </item>
+    <item>
       <descr><![CDATA[Enable TCP extended debugging]]></descr>
       <tunable>net.inet.tcp.log_debug</tunable>
       <value>default</value>
--- a/src/etc/inc/system.inc
+++ b/src/etc/inc/system.inc
@@ -73,7 +73,9 @@ function system_sysctl_defaults()
         'debug.kassert.warn_only' => [ 'default' => '1', 'required' => true, 'description' => 'KASSERT triggers a panic (0) or just a warning (1)', 'type' => 'w' ],
         'hw.ibrs_disable' => [ 'default' => '0' ],
         'hw.ixl.enable_head_writeback' => [ 'default' => '0', 'required' => true ],
+        'hw.vtnet.csum_disable' => [ 'default' => '1', 'required' => true ],
         'hw.syscons.kbd_reboot' => [ 'default' => '0' ],
+        'hw.vtnet.csum_disable' => [ 'default' => '1', 'required' => true, 'description' => 'Disable receive and send checksum offload for vtnet(4). Recommended for most KVM based cloud providers.' ],
         'hw.uart.console' => [ 'default' => 'io:0x3f8,br:' . system_console_speed(), 'type' => 't' ], /* XXX support comconsole_port if needed */
         'kern.coredump' => [ 'default' => '0', 'required' => true ],
         'kern.ipc.maxsockbuf' => [ 'default' => '4262144' ],
--- a/src/etc/inc/system.inc
+++ b/src/etc/inc/system.inc
@@ -75,7 +75,7 @@ function system_sysctl_defaults()
         'hw.ixl.enable_head_writeback' => [ 'default' => '0', 'required' => true ],
         'hw.vtnet.csum_disable' => [ 'default' => '1', 'required' => true ],
         'hw.syscons.kbd_reboot' => [ 'default' => '0' ],
-        'hw.vtnet.csum_disable' => [ 'default' => '1', 'required' => true, 'description' => 'Disable receive and send checksum offload for vtnet(4). Recommended for most KVM based cloud providers.' ],
+        'hw.vtnet.csum_disable' => [ 'default' => '1', 'required' => true ],
         'hw.uart.console' => [ 'default' => 'io:0x3f8,br:' . system_console_speed(), 'type' => 't' ], /* XXX support comconsole_port if needed */
         'kern.coredump' => [ 'default' => '0', 'required' => true ],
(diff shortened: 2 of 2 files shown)


src/etc/inc/system.inc
```diff
@@ -73,7 +73,9 @@ function system_sysctl_defaults()
         'debug.kassert.warn_only' => [ 'default' => '1', 'required' => true, 'description' => 'KASSERT triggers a panic (0) or just a warning (1)', 'type' => 'w' ],
         'hw.ibrs_disable' => [ 'default' => '0' ],
         'hw.ixl.enable_head_writeback' => [ 'default' => '0', 'required' => true ],
+        'hw.vtnet.csum_disable' => [ 'default' => '1', 'required' => true ],
```
fichtner, line 76: you really only need this line

I'm not overly enthusiastic about having two spots and the default already being disabled, but having an automatic sysctl would be ok

src/etc/inc/system.inc
```diff
@@ -73,7 +73,9 @@ function system_sysctl_defaults()
...
         'hw.syscons.kbd_reboot' => [ 'default' => '0' ],
+        'hw.vtnet.csum_disable' => [ 'default' => '1', 'required' => true, 'description' => 'Disable receive and send checksum offload for vtnet(4). Recommended for most KVM based cloud providers.' ],
```
fichtner, line 78: remove this (description is auto-parsed from sysctl)

pmhausen: Done.

src/etc/inc/system.inc
pmhausen: > I'm not overly enthusiastic about having two spots and the default already being disabled, but having an automatic sysctl would be ok

The default of hardware offloading is enabled - "disabled" sysctl == 0 - I am suggesting to disable it by default for new OPNsense installs. Reason being that it doesn't hurt but helps a lot with the majority of common cloud providers.

fichtner, line 76:
```suggestion
```

src/etc/config.xml.sample
```diff
@@ -134,6 +134,11 @@
       <tunable>hw.syscons.kbd_reboot</tunable>
       <value>default</value>
     </item>
+    <item>
+      <descr><![CDATA[Disable receive and send checksum offload for vtnet(4). Recommended for most KVM based cloud providers.]]></descr>
+      <tunable>hw.vtnet.csum_disable</tunable>
+      <value>default</value>
+    </item>
```
fichtner, line 141:
```suggestion
```

src/etc/inc/system.inc
pmhausen: If I commit both your suggested changes, what's left? What is your intention?

fichtner: 1 single line that is needed for this to work, see my first comment.

pmhausen: But that is already the case: https://github.com/opnsense/core/pull/8007/files

Your new suggested change removes even that one so nothing is left.

fichtner: Look at the diff in GitHub. It was added twice.

pmhausen: Done :) My bad.

fichtner:
I'll allow this experiment, but not in 24.7.x to avoid unforeseen regressions. Basically all it does is scrub the capabilities here:

https://github.com/opnsense/src/blob/d4799c2b7bbfa9a26316838daa7ee420dc7d028c/sys/dev/virtio/network/if_vtnet.c#L635-L645

Which should already be disable by default. So this failsafe will prevent later misconfiguration, but will also be non-obvious to the user. If CSUM/TSO/LRO do not disable properly on vtnet(4) that should be a better area of work in the future.

Merged, thanks!


Cheers,
Franco

pmhausen:
Thanks. Actually to my reading of the code and my experience in a default installation without any tunables set these features are negotiated with the virtual hardware.

The problem I am addressing here is that the hypervisor host announces them as available and the driver does indeed enable them. See my VM at Vultr:

```
root@us:~ # freebsd-version 
14.1-RELEASE-p4
root@us:~ # ifconfig vtnet0
vtnet0: flags=1008843<UP,BROADCAST,RUNNING,SIMPLEX,MULTICAST,LOWER_UP> metric 0 mtu 1500
	options=4c07bb<RXCSUM,TXCSUM,VLAN_MTU,VLAN_HWTAGGING,JUMBO_MTU,VLAN_HWCSUM,TSO4,TSO6,LRO,VLAN_HWTSO,LINKSTATE,TXCSUM_IPV6>
	ether 56:00:04:d6:14:ae
	inet 66.135.24.72 netmask 0xfffffe00 broadcast 66.135.25.255
	inet6 fe80::5400:4ff:fed6:14ae%vtnet0 prefixlen 64 scopeid 0x1
	inet6 2001:19f0:1000:13c9:5400:4ff:fed6:14ae prefixlen 64 autoconf pltime 604800 vltime 2592000
	media: Ethernet autoselect (10Gbase-T <full-duplex>)
	status: active
```

See? Enabled. And bad performance because the host does not actually perform what it is supposed to.

Kind regards,
Patrick

AdSchellevis:
The better question might be why it is enabled, we disable all offloading by default (none of m `vtnet` interfaces  I'm using have any offloading enabled)

https://github.com/opnsense/core/blob/d67f9b4ab80e2b30d8fe0b4d2c2c4b884f302d1d/src/etc/config.xml.sample#L226-L229

https://github.com/opnsense/core/blob/d67f9b4ab80e2b30d8fe0b4d2c2c4b884f302d1d/src/etc/inc/interfaces.lib.inc#L570-L594

Output of a clean installed proxmox box:
```
# ifconfig vtnet1
vtnet1: flags=1008843<UP,BROADCAST,RUNNING,SIMPLEX,MULTICAST,LOWER_UP> metric 0 mtu 1500
	description: WAN (wan)
	options=800a8<VLAN_MTU,JUMBO_MTU,VLAN_HWCSUM,LINKSTATE>
	ether 99:99:99:99:99:99
	inet x.x.x.x netmask 0xffffff00 broadcast 10.37.132.255
	inet6 fe80::be24:11ff:fea7:9980%vtnet1 prefixlen 64 scopeid 0x2
	media: Ethernet autoselect (10Gbase-T <full-duplex>)
	status: active
```

Checked a box with default config on Vultr as well, which has slightly different options, but certainly no rx/tx offloading enabled.

AdSchellevis:
better revert it then I guess...

pmhausen:
It definitely does not hurt. This much I know from the ongoing discussion upstream.

Issue #8009, opened 2024-10-23
Firewall - skip filtering loopback instead of allowing all


AdSchellevis:
**Important notices**

Before you add a new report, we ask you kindly to acknowledge the following:

- [x] I have read the contributing guide lines at https://github.com/opnsense/core/blob/master/CONTRIBUTING.md
- [x] I am convinced that my issue is new after having checked both open and closed issues at https://github.com/opnsense/core/issues?q=is%3Aissue

**Is your feature request related to a problem? Please describe.**

Currently loopback traffic is pushed through pf, but always allowed. 

https://github.com/opnsense/core/blob/2c4c5cf09bb7472ba4292696c30159e8c118cc4b/src/etc/inc/filter.lib.inc#L527

While looking into https://github.com/opnsense/core/issues/8008, it turned out that our current stateless `lo0` rule caused issues when combined with syncookies. Further reading in OpenBSD's documentation does seem to recommend skipping loopback as a best practice (https://www.openbsd.org/faq/pf/filter.html).

As far as I could find, filtering on loopback has never been a conscious choice and around since the start of the codebase
(https://github.com/pfsense/pfsense/blame/349aa28eb1bb7cddc1c369f1536c70f139c4bae3/etc/inc/filter.inc#L876-L880)

**Describe the solution you like**

Skip loopback processing using:
```
set skip on lo0
```

And remove all loopback references in interface selections.

**Describe alternatives you considered**

Short term, this https://github.com/opnsense/core/commit/e7e7e57dc16c9d705cab4091844530b588f3e3c3 "solves" part of the problem.

**Additional context**

https://freebsdfoundation.org/resource/an-introduction-to-packet-filter-pf/

PR #8010, opened 2024-10-24
Firewall - adhere to best practices 

AdSchellevis:
(https://www.openbsd.org/faq/pf/filter.html) and skip lo0 from processing. closes https://github.com/opnsense/core/issues/8009

Since I needed to skim the filter.* files anyway, also add the following minor improvements:

* change array() to []
* skip ipv6 only rules when ipv6 is disabled
* make carp rules conditional

### DIFF
--- a/src/etc/inc/filter.lib.inc
+++ b/src/etc/inc/filter.lib.inc
@@ -206,13 +206,6 @@ function filter_core_rules_system($fw, $defaults)
     $dhcrelay6_interfaces = plugins_run('dhcrelay_interfaces', ['inet6']);
     $dhcrelay6_interfaces = !empty($dhcrelay6_interfaces['dhcrelay']) ? $dhcrelay6_interfaces['dhcrelay'] : [];
 
-    // block All IPv6 except loopback traffic
-    $fw->registerFilterRule(
-        1,
-        array('interface' => 'lo0', 'ipprotocol' => 'inet6', 'disabled' => isset($config['system']['ipv6allow']),
-          'descr' => 'Pass all loopback IPv6', '#ref' => 'system_advanced_network.php#ipv6allow'),
-        $defaults['pass']
-    );
     $fw->registerFilterRule(
         1,
         ['ipprotocol' => 'inet6','descr' => 'Block all IPv6', 'disabled' => !$ipv6_disabled,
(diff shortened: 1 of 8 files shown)


src/etc/inc/filter.lib.inc
```diff
@@ -201,90 +201,87 @@ function filter_core_get_default_nat_outbound_networks()
 function filter_core_rules_system($fw, $defaults)
 {
     global $config;
+    $ipv6_allow = isset($config['system']['ipv6allow']);
 
     $dhcrelay6_interfaces = plugins_run('dhcrelay_interfaces', ['inet6']);
     $dhcrelay6_interfaces = !empty($dhcrelay6_interfaces['dhcrelay']) ? $dhcrelay6_interfaces['dhcrelay'] : [];
 
     // block All IPv6 except loopback traffic
```
fichtner, line 209:
```suggestion
```

it's obvious now :)

src/etc/inc/filter.lib.inc
```diff
@@ -201,90 +201,87 @@ function filter_core_get_default_nat_outbound_networks()
...
         1,
-        array('interface' => 'lo0', 'ipprotocol' => 'inet6', 'disabled' => isset($config['system']['ipv6allow']),
-          'descr' => 'Pass all loopback IPv6', '#ref' => 'system_advanced_network.php#ipv6allow'),
-        $defaults['pass']
-    );
-    $fw->registerFilterRule(
-        1,
-        array('ipprotocol' => 'inet6','descr' => 'Block all IPv6', 'disabled' => isset($config['system']['ipv6allow']),
-            '#ref' => 'system_advanced_network.php#ipv6allow'),
+        ['ipprotocol' => 'inet6','descr' => 'Block all IPv6', 'disabled' => $ipv6_allow,
```
fichtner, line 212: is this inverted now? it would be nicer to rename the variable and flip the logic to make this read easily... disabled => ipv6_disable

AdSchellevis: I just kept the naming, I'm just afraid it will be confusing either way around. If we name `isset($config['system']['ipv6allow'])` `$ipv6_disable`, that's technically correct for what the rule should do (disable), but not what the ipv6 state is (enabled).....

AdSchellevis: flipped in https://github.com/opnsense/core/pull/8010/commits/a1138e57f4b4fb9db8d6b5745e95b2f449572b52, it feels easier on the eyes, so I'm ok with it.

src/etc/inc/filter.lib.inc
AdSchellevis: hehe, right https://github.com/opnsense/core/pull/8010/commits/a1138e57f4b4fb9db8d6b5745e95b2f449572b52

fichtner (approved):
it may be better to split the ipv6 disable and the lo0 skip, otherwise let's try this

AdSchellevis:
@fichtner I'll split it into two commits and merge it to master, thanks

Issue #8023, opened 2024-10-29
CPU Temperature graphs not working in 24.7.7


awilkin87:
**Important notices**

Before you add a new report, we ask you kindly to acknowledge the following:

- [X] I have read the contributing guide lines at https://github.com/opnsense/core/blob/master/CONTRIBUTING.md
- [X] I am convinced that my issue is new after having checked both open and closed issues at https://github.com/opnsense/core/issues?q=is%3Aissue

**Describe the bug**

After upgrading to 24.7.7, I found that the reporting graphs were not working. I reset the RRD data and restarted the system. The graphs re-populated, except for the cputemp graph.

After some troubleshooting, I tracked the issue to the recent RRD refactoring (https://github.com/opnsense/core/pull/7836). I found that the sysctl command to get CPU temps was returning an error. Also, the RRD filename is not defined in Types/Temperature.php.

**To Reproduce**

1. Install OPNsense 24.7.6
2. Ensure Thermal Sensors are configured correctly under System -> Settings -> Miscellaneous
3. Ensure "Enables the RRD graphing backend" is checked under Reporting -> Settings.
4. Ensure non-zero values are displayed in the System/Cputemp graph under Reporting -> Health.
5. Upgrade to OPNsense 24.7.7
6. Go to Reporting -> Settings
7. Click "Reset RRD Data" 
8. Wait over 1 minute

**Actual behavior**

- The system-cputemp RRD is not recreated
- There is no System/Cputemp graph under Reporting -> Health, or if there is a graph it does not update.

**Expected behavior**

- The system-cputemp RRD should be listed under Reporting -> Settings
- The System/Cputemp graph under Reporting -> Health should appear with non-zero values.

**Describe alternatives you considered**

N/A

**Screenshots**

N/A

**Relevant log files**

N/A

**Additional context**

N/A

**Environment**

Software version used and hardware type if relevant, e.g.:

OPNsense 24.7.7 (amd64).
Intel® Pentium® J3710 2.64Ghz Quad Core (using coretemp)
Network Intel® I225-V

awilkin87:
I've submitted a pull request with a fix: https://github.com/opnsense/core/pull/8024

PR #8024, opened 2024-10-29
Fix cputemp rrd

awilkin87:
This PR addresses Issue https://github.com/opnsense/core/issues/8023, where the CPU temperature graphs stop working after the the recent refactoring.

### DIFF
--- a/src/opnsense/scripts/health/library/OPNsense/RRD/Stats/Temperature.php
+++ b/src/opnsense/scripts/health/library/OPNsense/RRD/Stats/Temperature.php
@@ -33,10 +33,10 @@ class Temperature extends Base
     public function run()
     {
         $data = $this->shellCmd(
-            '/sbin/sysctl -ni dev.cpu.0.temperature hw.acpi.thermal.tz0.temperature hw.temperature.CPU | sed s/C//g'
+            '/sbin/sysctl -ni dev.cpu.0.temperature hw.acpi.thermal.tz0.temperature hw.temperature.CPU'
         );
         if (!empty($data)) {
-            return [$data[0]];
+            return [preg_replace('/[^0-9,.]/', '', $data[0])];
         }
         return [];
     }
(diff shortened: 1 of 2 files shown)


AdSchellevis:
@awilkin87 can you dump the output of both commands?

```
/sbin/sysctl -n dev.cpu.0.temperature hw.acpi.thermal.tz0.temperature hw.temperature.CPU
```

```
/sbin/sysctl -ni dev.cpu.0.temperature hw.acpi.thermal.tz0.temperature hw.temperature.CPU
```

awilkin87:
Sure, here are outputs and exit codes for my system:
```
root@opnsense:~ $ /sbin/sysctl -n dev.cpu.0.temperature hw.acpi.thermal.tz0.temperature hw.temperature.CPU
54.0C
sysctl: unknown oid 'hw.acpi.thermal.tz0.temperature'
sysctl: unknown oid 'hw.temperature.CPU'
root@opnsense:~ $ echo $?
2
root@opnsense:~ $ /sbin/sysctl -ni dev.cpu.0.temperature hw.acpi.thermal.tz0.temperature hw.temperature.CPU
49.0C
root@opnsense:~ $ echo $?
0
```

AdSchellevis:
@awilkin87 can you revert the changes to `Stats/Temperature.php` and try again with only the filename `shellCmd()` should take care of the output.

awilkin87:
I will test it later tonight, but I believe the `-i` flag is needed since `shellCmd()` requires a zero exit code in order to return the output.

AdSchellevis:
@awilkin87 looks like you're right, only strip the sed in that case and we should be good. thanks!

awilkin87:
@AdSchellevis I believe the `sed` is required as well. After removing it, the table view of cputemp is reporting zero values. Using `rrdtool dump /var/db/rrd/system-cputemp.rrd`, those entries are stored as NaN.

image

Just to be sure, I reset the rrd. While the file was recreated, it did not contain any temperature readings.

AdSchellevis:
@awilkin87 can you try https://github.com/opnsense/core/commit/8e3b4b7229e487e844cbc2dbaf58369b3ffcd081 ? the data needs to be cleansed indeed, but better keep it in php to prevent changing multiple shell commands.

awilkin87:
@AdSchellevis I tested with preg_replace instead of sed. It works as expected.

AdSchellevis:
@awilkin87 thanks for confirming, I'll ask @fichtner to slip the change into the next release.

PR #8025, opened 2024-10-29
firewall: groups: Fix wording of 9425ca7

Monviech:
Reverts to shall to keep translation the same.

Use same style for changed validation.


fichtner (approved):
Thanks!

Issue #7990, opened 2024-10-20
Regression: Update DNS with static entries without IP


Vaarlion:
**Important notices**

Before you add a new report, we ask you kindly to acknowledge the following:

- [X] I have read the contributing guide lines at https://github.com/opnsense/core/blob/master/CONTRIBUTING.md
- [X] I am convinced that my issue is new after having checked both open and closed issues at https://github.com/opnsense/core/issues?q=is%3Aissue

**Describe the bug**

Creating a static lease with a hostname and a mac but without an ip doesn't register it in the DNS server when "Register DHCP Static Mappings" is checked, and "Register ISC DHCP4 Leases" is disable.

This was introduced by this https://github.com/opnsense/core/pull/4689, who also removed the documentation warning about it not working, yet as far as i can tell, it doesn't work.

**To Reproduce**

1. Under Services: ISC DHCPv4, select a network and create a static lease for a known device. Do NOT set an ip, Set a Hostname and the MAC address.
2. Under Services: Unbound DNS: General, make sure "Register DHCP Static Mappings" is checked, and "Register ISC DHCP4 Leases" is NOT.
3. Connect the device, and check that it got a lease under Services: ISC DHCPv4: Leases. You should see 2 line for your device (weirdly) the static one without an ip, and the dynamic one with an ip.
4. Under Interfaces: Diagnostics: DNS Lookup, query for your device and see that it does not show up.

**Expected behavior**
Either the device host-name and it's dynamic ip is added to the DNS server automatically, or the documentation saying that this doesn't work need to be brought back.

**Describe alternatives you considered**

N/A

**Screenshots**

N/A

**Relevant log files**

Request if needed, i don't know which one would be useful.

**Additional context**

I have no idea if this ever worked, i've moved to Opnsense recently and can only see that the documentation about this feature not working was removed, yet i can't get this to work. 

**Environment**

Software version used and hardware type if relevant, e.g.:

I'm using ISC DHCPV4 and Unbound DNS, on Opensense 24.1.10_8, waiting for 24.7 to have a couple more month under the belt before upgrading.

AdSchellevis:
The feature depends on "Register ISC DHCP4 Leases", I don't mind if someone opens a PR for the documentation, but without collecting addresses, it's just impossible to tell which address the lease belongs to. As `isc-dhcp` is sunsetting, I don't think we should spend much time on this code to be honest.

Vaarlion:
[...]

PR #8028, opened 2024-10-30
Docs/clarify static lease register

Vaarlion:
Add some documentation to clarify #7990

I could not run the test on my computer, the makefile being for BSD system.
I'm hopping for an internal CI on your side ?

### DIFF
--- a/src/opnsense/mvc/app/controllers/OPNsense/Unbound/forms/general.xml
+++ b/src/opnsense/mvc/app/controllers/OPNsense/Unbound/forms/general.xml
@@ -63,6 +63,7 @@
         <type>checkbox</type>
         <help>
             <![CDATA[If this option is set, then DHCP static mappings will be registered in Unbound, so that their name can be resolved.
+            DHCP lease without a specified static ip cannot be register.
             You should also set the domain in <a href="/system_general.php">System: Settings: General</a> to the proper value.]]>
         </help>
     </field>


src/opnsense/mvc/app/controllers/OPNsense/Unbound/forms/general.xml
```diff
@@ -63,6 +63,7 @@
         <type>checkbox</type>
         <help>
             <![CDATA[If this option is set, then DHCP static mappings will be registered in Unbound, so that their name can be resolved.
+            DHCP lease without a specified static ip cannot be register.
```
fichtner, line 66: Static mappings are not leases. IP addresses are more or less implied for DNS to resolve but clarifying it in the first sentence wouldn’t hurt I suppose for all those who do read…

Vaarlion: Does `DHCP static mapping which doesn't contain a static ip will not be registered` sound more accurate ?

fichtner: You want to make it affirmative:

… static mappings with an IP address will be registered…

Vaarlion:
My bad sorry, i had other thing rise and completely forgot about this....
It look like i was just being dumb anyway :)

fichtner:
@Vaarlion no worries. appreciate the PR and discussion yet sometimes we also have to discuss and wait for similar reports for these types of suggestions.. sometimes documentation also makes sense.. here it looks like it's ok as is.

Issue #7949, opened 2024-10-05
Kia Leases


simonmcnair:
**Important notices**

Before you add a new report, we ask you kindly to acknowledge the following:

- [yes] I have read the contributing guide lines at https://github.com/opnsense/core/blob/master/CONTRIBUTING.md
- [yes] I am convinced that my issue is new after having checked both open and closed issues at https://github.com/opnsense/core/issues?q=is%3Aissue

**Is your feature request related to a problem? Please describe.**
It would be great if this could follow the same format as the ISC dhcp V4 page, especially giving the cool and useful mac address vendor. going from 10 fields to 6 seems like a reduction in helpfulness.

It would also be helpful to understand if /how/wy kia supports ipv6 as I see it for DHCP and I am pretty sure Kia does both IPV4 and IPV6 but the fact that IPV4 is specified everywhere is perhaps misleading or inaccurate ?

Thanks :-)

A clear and concise description of what the problem is including your motivation for the request,
i.e. "For the purpose of [...] I am missing a solution that will [...]."

**Describe the solution you like**

I would like the Kia dhcp leases page to look like the ISC leases page.

**Describe alternatives you considered**

A clear and concise description of any alternative solutions or features you considered.

**Additional context**

Add any other context or screenshots about the feature request here or links to relevant forum thread or similar

cpalv:
is this something that the internal team finds useful?  I think this and #7950 could help the transition from the ISC dhcp EOL.  I don't think the Kea leases and ISC leases page need to look exactly the same, but I find some of the missing fields: status, state, and lease type do provide some good information.  I am also interested in using Kea for IPv6.  If there isn't anybody looking into it I could take a crack at it

AdSchellevis:
@cpalv I don't mind adding information if it is of relevance, it hasn't been an explicit choice not to add mac vendor for example, just not a priority on our end. If you open a PR, I'll take a look at it. I also don't think we should aim in making things the same by the way.

cpalv:
just opened PR, the lease file clean up setting should also address this [forum post](https://forum.opnsense.org/index.php?topic=43485.msg216357#msg216357)

AdSchellevis:
https://github.com/opnsense/core/pull/9409

Issue #7950, opened 2024-10-05
[Feature request] Kia leases - create a reservation


simonmcnair:
**Important notices**

Before you add a new report, we ask you kindly to acknowledge the following:

- [ yes] I have read the contributing guide lines at https://github.com/opnsense/core/blob/master/CONTRIBUTING.md
- [ yes] I am convinced that my issue is new after having checked both open and closed issues at https://github.com/opnsense/core/issues?q=is%3Aissue

**Is your feature request related to a problem? Please describe.**

It's not a problem, it improves usability

**Describe the solution you like**

It would be nice if, within kia leases, you could add a a button in the Kia leases page to create a reservation so that you don't manually have to type in the host and mac address for known clients.

**Describe alternatives you considered**

A clear and concise description of any alternative solutions or features you considered.

**Additional context**

Add any other context or screenshots about the feature request here or links to relevant forum thread or similar

PR #8039, opened 2024-11-05
Kea dhcp4

cpalv:
for #7949


AdSchellevis:
It might be better to start with a simple PR for adding the mac vendor as mentioned in the original ticket, I do have doubts about trying to glue everything around reservations together in this page as we try to prevent side affects.

Making the `lfc-interval` optional (and changing the default) as part of a cleanup is also not great in terms of traceability.

cpalv:
`lfc-interval` is a required field? I was attempting to address this [issue](https://forum.opnsense.org/index.php?topic=43485.msg216357#msg216357) mentioned in the forums.  The kea documentation also says that the LFC process is [necessary](https://kea.readthedocs.io/en/kea-2.6.1/arm/dhcp4-srv.html#why-is-lease-file-cleanup-necessary).  

There is already a setting for `Periodic DHCP Leases Backup` in `System > Settings > Miscellaneous` that when disabled I assume will clear the current lease file on poweroff, but this probably doesn't mean much for devices that do not reboot.  I am also not  sure how this setting works w/ Kea or if it only effects ISC leases.

Kea LFC does run as a separate process though and I'm not sure where or if it logs anything about the lease file cleanup.  So it is kind of magical as far as traceability goes. 

Gluing together the leases and reservations seemed like the most straightforward way to handle #7950 along with getting reservation info/status for #7949.

I may have tried to do too much in one PR.  I can tear this one down and try just the MAC vendor information if that's preferable

AdSchellevis:
> `lfc-interval` is a required field?

And set by default, from the manual (https://kea.readthedocs.io/en/kea-2.0.0/arm/dhcp4-srv.html):

> ... The default value of the lfc-interval is 3600. A value of 0 disables the LFC.

>There is already a setting for Periodic DHCP Leases Backup in System > Settings > Miscellaneous that when disabled I assume will clear the current lease file on poweroff, but this probably doesn't mean much for devices that do not reboot. I am also not sure how this setting works w/ Kea or if it only effects ISC leases.

That's legacy isc only, handled via https://github.com/opnsense/core/blob/master/src/etc/rc.syshook.d/backup/20-dhcpleases

> I may have tried to do too much in one PR. I can tear this one down and try just the MAC vendor information if that's preferable

One step at a time would be better indeed, my time is limited and reviewing large blobs of code often cost more time than adding features ourselves.

cpalv:
Ah, I must've glossed over that section.  I'll go ahead and close this PR and open another

Issue #6230, opened 2023-01-06
Auth Base has LDAP specific group handling, which is incompatible with the Radius Auth provider


peter-dolkens:
**Important notices**

Before you add a new report, we ask you kindly to acknowledge the following:

- [ ] I have read the contributing guide lines at https://github.com/opnsense/core/blob/master/CONTRIBUTING.md
- [ ] I am convinced that my issue is new after having checked both open and closed issues at https://github.com/opnsense/core/issues?q=is%3Aissue

**Describe the bug**

The following code in `Auth/Base.php` is LDAP specific, and therefore fails when called by the `Auth/Radius.php` provider, as the RADIUS `Class` attribute format is incompatible with the LDAP format

https://github.com/opnsense/core/blob/a8bd3379b1a8c922ba181648626431c543b378b3/src/opnsense/mvc/app/library/OPNsense/Auth/Base.php#L163-L172

**To Reproduce**

1. Setup a RADIUS auth server that passes one or more groups using the `Class` attribute
2. Use the Tester to test a valid set of credentials
3. Notice the groups fail to come through

**Expected behavior**

Groups should be parsed. An alternative delimiter `;` or `,` should also be supported, as I don't believe RADIUS will like the current `\n` delimiter.

**Describe alternatives you considered**

n/a

**Screenshots**

![image](https://user-images.githubusercontent.com/469666/211098673-a2c6424a-cc34-452d-acd8-8f746a72c60e.png)

**Relevant log files**

n/a

**Additional context**

n/a

**Environment**

Software version used and hardware type if relevant, e.g.:

OPNsense 22.7.10

AdSchellevis:
I've recently added https://github.com/opnsense/core/commit/ef0da3ea59ffaeaa70f7baef8245b347d3097589, if radius specifies a different delimiter, it should be changed here 

https://github.com/opnsense/core/blob/ef0da3ea59ffaeaa70f7baef8245b347d3097589/src/opnsense/mvc/app/library/OPNsense/Auth/Radius.php#L462-L467

Some standard or RFC to refer to would be practical. The `\n` delimiter is just an internal convention, as implemented here for ldap 

https://github.com/opnsense/core/blob/ef0da3ea59ffaeaa70f7baef8245b347d3097589/src/opnsense/mvc/app/library/OPNsense/Auth/LDAP.php#L510

peter-dolkens:
Thanks @AdSchellevis - I actually got called away while I was filing this, but wanted to get it noted down.

[...]

PR #8055, opened 2024-11-12
Tooltip for Radius group sync

adn77:
This tries to better explain how the Synchronize groups feature works with RADIUS servers.

https://github.com/opnsense/core/issues/6230#issuecomment-2463987192


AdSchellevis:
documentation https://docs.opnsense.org/manual/how-tos/user-radius.html#access-servers-radius

Issue #3290, opened 2019-03-06
[feature request] OpenVPN with OTP and Challenge/Response


tbandixen:
- [x] I have read the contributing guide lines at https://github.com/opnsense/core/blob/master/CONTRIBUTING.md

- [x] I have searched the existing issues and I'm convinced that mine is new.

**Is your feature request related to a problem? Please describe.**
The OpenVPN client has a nice option to add a challenge/response input box to enter a OTP, however I can't figure out how this should work in OPNsense. 

I configured OpenVPN with google authenticator (which works great), but it requires to enter the number in combination with the password.

**Describe the solution you'd like**
I feel it would be much more user friendly if we could use the "static-challenge" option in the client.

![image](https://user-images.githubusercontent.com/2293689/53862172-c57f6800-3fe6-11e9-8a1c-17ed3cfbf268.png)

**Additional context**
Just setting the static-challenge option in the client gives a "SIGUSR1[soft,auth-failure] received, process restarting".

The thread was opened by [olivierfaber](https://forum.opnsense.org/index.php?action=profile;u=19321)  in the [forum](https://forum.opnsense.org/index.php?topic=10082.0).
I just ported it to github, that we can discuss thing a bit nearer to the code 😄

tbandixen:
The Viscosity Client has this nice feature too.
By setting `static-challenge "Please enter your OTP code" 1` in the client settings you'll get this nice window after you entered your credentials.
![image](https://user-images.githubusercontent.com/2293689/53864259-53118680-3fec-11e9-8243-4659469fd2dc.png)

I dont know yet, what needs to be changed server side. I'm totaly new in the OpenVPN topic.

AdSchellevis:
looks doable, it seems to encode the password differently when set:
```
SCRV1:dGVzdA==:OTk5OTk5OTk5OQ==
```
is the equivalent for password `test` token `9999999999`

tbandixen:
Is this related to this docu?
https://github.com/OpenVPN/openvpn/blob/master/doc/management-notes.txt#L1120

AdSchellevis:
yes, it seems so, specifically this:
```
password "Auth" "SCRV1:<password_base64>:<response_base64>"
```

tbandixen:
With this request, I hope that I can save my credentials in the vpn client and only have to enter the OTP...

AdSchellevis:
There will be one limitation, it won't support "reverse token order"
[...]

PR #8058, opened 2024-11-13
Openvpn authentication with otp as suffix

tsdogs:
Authentication server as Local and LDAP with TOTP have an option which permits to enter the TOTP code after the password (default is as prefix)

In this case if you enable the TOTP question in Openvpn Client (static-challenge) the authentication fails as the PIN is presented as prefix and not as a suffix.

I currently have many configurations where static-challenge is not enabled and thus the users write the password and  the PIN in the password input on the client.
Migrating to support the static-challenge would require them to be changed all at the same time.

I propose the move of the check (and how the password is composed) based on the Authentication method settings.

### DIFF
--- a/src/opnsense/mvc/app/library/OPNsense/Auth/TOTP.php
+++ b/src/opnsense/mvc/app/library/OPNsense/Auth/TOTP.php
@@ -54,7 +54,7 @@ trait TOTP
     /**
      * @var bool token after password
      */
-    private $passwordFirst = false;
+    public $passwordFirst = false;
 
     /**
      * use graceperiod and timeWindow to calculate which moments in time we should check
--- a/src/opnsense/scripts/openvpn/user_pass_verify.php
+++ b/src/opnsense/scripts/openvpn/user_pass_verify.php
@@ -80,16 +80,15 @@ function do_auth($common_name, $serverid, $method, $auth_file)
     if (empty($username) || empty($password)) {
         return "username or password missing ({$method} - {$auth_file})";
     }
+    $pinpassword = false;
+    $pin = false;
     if (strpos($password, 'SCRV1:') === 0) {
         // static-challenge https://github.com/OpenVPN/openvpn/blob/v2.4.7/doc/management-notes.txt#L1146
-        // validate and concat password into our default pin+password
+        // validate and split password into pin and password (used below based on authenticato settings)
         $tmp = explode(':', $password);
         if (count($tmp) == 3) {
-            $pass = base64_decode($tmp[1]);
+            $pinpassword = base64_decode($tmp[1]);
             $pin = base64_decode($tmp[2]);
-            if ($pass !== false && $pin !== false) {
-                $password = $pin . $pass;
-            }
         }
     }
     $a_server = $serverid !== null ? (new OPNsense\OpenVPN\OpenVPN())->getInstanceById($serverid, 'server') : null;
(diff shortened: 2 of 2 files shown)


src/opnsense/scripts/openvpn/user_pass_verify.php
```diff
@@ -80,16 +80,15 @@ function do_auth($common_name, $serverid, $method, $auth_file)
...
+    $pinpassword = false;
+    $pin = false;
     if (strpos($password, 'SCRV1:') === 0) {
         // static-challenge https://github.com/OpenVPN/openvpn/blob/v2.4.7/doc/management-notes.txt#L1146
-        // validate and concat password into our default pin+password
+        // validate and split password into pin and password (used below based on authenticato settings)
         $tmp = explode(':', $password);
         if (count($tmp) == 3) {
-            $pass = base64_decode($tmp[1]);
+            $pinpassword = base64_decode($tmp[1]);
```
fichtner, line 90: all of this just obscures the intention...

src/opnsense/scripts/openvpn/user_pass_verify.php
```diff
@@ -121,7 +120,15 @@ function do_auth($common_name, $serverid, $method, $auth_file)
...
-            if ($authenticator->authenticate($username, $password)) {
+            $checkpassword = $password;
+            if ($pin!=false) { 
+                if (property_exists($authenticator,'passwordFirst') && $authenticator->passwordFirst) {
+                    $checkpassword = $pinpassword . $pin;
+                } else {
+                    $checkpassword = $pin . $pinpassword;
+                }
+            }
+            if ($authenticator->authenticate($username, $checkpassword)) {
```
fichtner, line 131: if we had a third argument for the pin this might be possible to handle gracefully inside the authenticator but I am not sure of the benefit of al of this CC @AdSchellevis

src/opnsense/mvc/app/library/OPNsense/Auth/TOTP.php
```diff
@@ -54,7 +54,7 @@ trait TOTP
     /**
      * @var bool token after password
      */
-    private $passwordFirst = false;
+    public $passwordFirst = false;
```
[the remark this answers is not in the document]
tsdogs: This is needed to know if the authenticator has the pin as suffix.
There's a readonly option but then it needs to be initialized in the creator of the class.

src/opnsense/scripts/openvpn/user_pass_verify.php
tsdogs: I thought also about that, but it would impact more code. If needed I might just change the Auth backends to accept the pin.

tsdogs: About #3290 that's exactly the point. Enabling the pin request does not handle the reverse option enabled.

tsdogs:
I can understand you not accepting this, as the solution for you would simply be to remove the "reverse" option.
Thing is that when you have many users out there it takes time to migrate all of them.

src/opnsense/scripts/openvpn/user_pass_verify.php
tsdogs: We might move this also inside the authenticator loop for clarity, just it didn't seem programmatically correct to me.

src/opnsense/scripts/openvpn/user_pass_verify.php
tsdogs: Should we maybe move the "SCRV1:" handling code inside the authenticator itself?

AdSchellevis: easiest option is just to add 2 authenticators if you need both orders, use one for openvpn, the other for the ui. I don't think we need more glue to be honest, a note in the documentation about limitations should be more than enough.

AdSchellevis:
this https://github.com/opnsense/core/commit/cc16ba7b33d09f3fc4e1b6f673a9ad4d89b669ff should do the trick, @tsdogs  if you have time to test that would be helpful.

PR #8061, opened 2024-11-13
Remote syslog localip

jahlers:
We ran into a problem where remote syslog from branch offices won't be sent through an IPSEC tunnel to the central log server. 
Solution to this issue is binding syslog-ng's source to a specific ip using syslog-ng's "localip" option.


AdSchellevis:
just add a stick route and you should be fine, this is a quite common and easy to fix issue (you need to make sure traffic is send into the trap). See also the note in https://docs.opnsense.org/manual/diagnostics_interfaces.html#ping

Issue #8035, opened 2024-11-03
KEA DHCP Lease to Reservation Wizard.


LostPoE:
The idea is to have a button next to all the leased IP address in the KEA DHCP leases section that would start a wizard to migrate that IP to a reservation. Things to choose form in the wizard like subnet and subnet description, IP, description, hostname. This would save a lot on copy/paste. Thanks!

PR #8073, opened 2024-11-19
Kea: Add direct reservation from leases page

pieewiee:
This PR adds the ability to create DHCP reservations directly from the leases page in Kea DHCP.

Ref: #8035

While this PR doesn't implement the full wizard functionality requested in the issue, it provides a first step towards that goal by allowing users to:
- Convert an existing lease to a reservation with a single click
- See which leases are already reserved
- Apply changes through an integrated Apply button

The current implementation automatically matches the subnet based on the IP address and transfers the existing hostname. This could be extended in the future to include a full wizard with more configuration options as suggested in the original issue.

Current features:
- Added addReservationAction to Leases4Controller
- Updated leases view with reservation button and Apply functionality
- Added proper validation and subnet matching for new reservations
- Visual indication of which leases are already reserved

This provides a foundation that could be expanded to include the full wizard functionality in the future.

Changes:
- Added addReservationAction to handle reservation creation
- Updated leases view with reservation button
- Added Apply button for service reconfiguration
- Added proper validation and subnet matching

lease

reserved

### DIFF
--- a/src/opnsense/mvc/app/controllers/OPNsense/Kea/Api/Leases4Controller.php
+++ b/src/opnsense/mvc/app/controllers/OPNsense/Kea/Api/Leases4Controller.php
@@ -1,30 +1,30 @@
 <?php
 
 /*
- * Copyright (C) 2023 Deciso B.V.
- * All rights reserved.
- *
- * Redistribution and use in source and binary forms, with or without
- * modification, are permitted provided that the following conditions are met:
- *
- * 1. Redistributions of source code must retain the above copyright notice,
- *    this list of conditions and the following disclaimer.
- *
- * 2. Redistributions in binary form must reproduce the above copyright
- *    notice, this list of conditions and the following disclaimer in the
- *    documentation and/or other materials provided with the distribution.
- *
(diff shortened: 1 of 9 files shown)


pieewiee:
please give me your honest feedback - i would apricate your improvements!!!

AdSchellevis:
I have to think about this, the feature overlaps with an earlier PR (https://github.com/opnsense/core/pull/8039), one of the main concerns I see is that it's generally not a good idea to assign addresses inside an existing pool as these are free for usage when the machine in question is not connected (at least with `isc-dhcp`). 

Ones we started to accept these in the legacy (isc-dhcp) server, we have had several reports about this behavior, which isn't ours to fix.

pieewiee:
i could have checked other PRs before spending time on this. You maybe got another issue for me that i could support (something easy)?

honestly the other PR (https://github.com/opnsense/core/pull/8039) is better in terms of how he handled the add reservation

AdSchellevis:
> i could have checked other PRs before spending time on this.

no problem, things happen.

> You maybe got another issue for me that i could support (something easy)?

Any particular topic in mind you would like to spend time on? There likely aren't a lot of easy ones, but I can search the list to see if something might be worth giving a try.

PR #8085, opened 2024-11-25
monit: wrap exec in double quotes to allow arguments

uvNikita:
Currently, if you want to execute command with arguments, you have to create a script somewhere on filesystem, then specify path to it in the Execute parameter.
With this patch, it's possible to pass command line arguments as part of Execute string, e.g. `/sbin/ifconfig wg0 down`.

From [monit documentation](https://mmonit.com/monit/documentation/monit.html#ACTION):

> EXEC can be used to execute an arbitrary program and send an alert. If you choose this action you must state the program to be executed and if the program requires arguments you must enclose the program and its arguments in a quoted string

### DIFF
--- a/src/opnsense/service/templates/OPNsense/Monit/monitrc
+++ b/src/opnsense/service/templates/OPNsense/Monit/monitrc
@@ -147,7 +147,7 @@ check {{ service.type }} {{ service.name }} {{ path }}
 {%       if test.condition is defined and test.action is defined %}
 {%        set epath = '' %}
 {%        if test.action == 'exec' and test.path|default('') != '' %}
-{%         set epath = test.path %}
+{%         set epath = "\"" ~ test.path ~ "\"" %}
 {%        endif %}
    if {{ test.condition }} then {{ test.action }} {{ epath }}
 {%       endif %}


uvNikita:
@AdSchellevis could you look at this fix? 🙂 It's a pretty trivial change.

AdSchellevis:
to be very honest, adding monit to core has been a past mistake as it's impossible to validate the input here. I'll take this change under consideration, but given it currently just passes the input, there might be unintended side affects.

maurice-w:
@AdSchellevis  This does indeed have unintended side effects. It breaks the monit configuration if double quotation marks or options like `repeat every` are used in the _Path_ field. Some of my tests which worked in 24.7 don't work in 25.1 (syntax error).

I suggest to revert this.

@uvNikita Your use case can be solved without this patch, for example:
`/bin/sh -c "ifconfig wg0 down"`

AdSchellevis:
@maurice-w reverted the patch in https://github.com/opnsense/core/commit/bef9066195d61b71638cb3d4d593684020b0f4d6 , it's not worth my time to look into this further. The monit feature is unfortunately full of gaps like this, from a validation perspective, this shouldn't have entered our core branch to be very honest. In time we might consider moving it out of core.

PR #8090, opened 2024-11-26
Update temperature.sh for smaller CPU footprint 

meyergru:
Update temperature.sh for smaller CPU footprint
This patch eliminates many lines of output from sysctl and searches it much faster, because there is no regex matching.

This results in a smaller CPU footprint, thus reducing the impact of temperature monitoring itself on the reported temps, without changing the outcome.

See also: https://forum.opnsense.org/index.php?topic=36234.0

### DIFF
--- a/src/www/interfaces.php
+++ b/src/www/interfaces.php
@@ -1316,10 +1316,6 @@ if (!interface_ppps_capable($a_interfaces[$if], $a_ppps)) {
     /* do not offer these raw types as a transition back from PPP */
     $types4['staticv4'] = gettext('Static IPv4');
     $types4['dhcp'] = gettext('DHCP');
-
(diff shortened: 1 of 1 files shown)


fichtner:
Sorry, no. It just discards values for other people.

meyergru:
Like which? The code analysing the output even filter on hw.acpi....

fichtner:
For reference: f473d9a5c74e

fichtner:
> Like which? The code analysing the output even filters on hw.acpi....

It does not filter, it separates the Zone and CPU cases (which is also too optimistic).

meyergru:
I never saw any temperatures in sysctl besides hw.acpi and dev.cpu - though that does not mean they do not exist.

Issue #7898, opened 2024-09-24
system: add a certificate widget to allow visibility for upcoming expiries


fichtner:
**Important notices**

Before you add a new report, we ask you kindly to acknowledge the following:

- [x] I have read the contributing guide lines at https://github.com/opnsense/core/blob/master/CONTRIBUTING.md
- [x] I am convinced that my issue is new after having checked both open and closed issues at https://github.com/opnsense/core/issues?q=is%3Aissue

**Is your feature request related to a problem? Please describe.**

People have always asked for certificate expiration notifications, but these are not very reliable if ignored and when they should notify which leaves a lot of configuration glue. But expiry is a problem that needs to be dealt with.

**Describe the solution you like**

Build a dashboard widget to list the certificates to be expired soon with delete and renew functions.

**Describe alternatives you considered**

N/A

**Additional context**

N/A

Monviech:
The delete and renew functions are out of scope for the widget, they should be handled in their respective UI pages.

Issue #7895, opened 2024-09-24
dashboard: which plugins may require inline edit functionality for their data?


fichtner:
**Important notices**

Our forum is located at https://forum.opnsense.org , please consider joining discussions there in stead of using GitHub for these matters.

Before you ask a new question, we ask you kindly to acknowledge the following:

- [x] I have read the contributing guide lines at https://github.com/opnsense/core/blob/master/CONTRIBUTING.md
- [x] I am convinced that my issue is new after having checked both open and closed issues at https://github.com/opnsense/core/issues?q=is%3Aissue

This would extend the idea of a general edit page link being able to make changes in a modal directly on the dashboard.  Also collect the widgets from the plugins if applicable.

Monviech:
@fichtner

I gave this a quick look and it does not seem to provide much benefit anymore, now that every widget has the quick link button to its respective page.

I also couldn't find widgets that would clearly benefit from this. Some widgets already have more selective behavior:
- e.g. Interfaces goes to the correct interfaces page
- e.g. Firewall opens Livelog with correct rid selected

For other widgets, it is much easier to find and edit data in the respective core or plugin pages after clicking the referral link. I would rather give this functionality a pass.

Monviech:
Since I found a working simple solution for this I gonna reopen it and add it to some other widgets, e.g. the VPN widgets.

PR #8105, opened 2024-12-03
dashboard: Add certificate widget that displays CAs and Certs sorted by expiration date

Monviech:
Fixes: https://github.com/opnsense/core/issues/7898

Adds a widget that pulls all CA and Certificates from the trust store API and displays them sorted by expiration date.

- Certificates with more than 14 days show a green lock
- with less than 14 days show yellow lock
- with no days left show red log and expired

An additional self populating dropdown allows to configure the plugin, hiding certificates.
The default is displaying all existing certificates.

### DIFF
--- a/src/opnsense/www/js/widgets/Certificates.js
+++ b/src/opnsense/www/js/widgets/Certificates.js
@@ -109,6 +109,10 @@ export default class Certificates extends BaseTableWidget {
                 const statusText = remainingDays === 0 ? this.translations.expired : this.translations.valid;
                 const iconClass = remainingDays === 0 ? 'fa fa-unlock' : 'fa fa-lock';
 
+                const expirationText = remainingDays === 0
+                    ? `${this.translations.expiredon} ${validTo.toLocaleString()}`
+                    : `${this.translations.expiresin} ${remainingDays} ${this.translations.days}, ${validTo.toLocaleString()}`;
+
                 const row = `
                     <div>
                         <i class="${iconClass} ${colorClass} certificate-tooltip" style="cursor: pointer;"
@@ -118,8 +122,7 @@ export default class Certificates extends BaseTableWidget {
                         <span><b>${item.descr}</b></span>
                         <br/>
                         <div style="margin-top: 5px; margin-bottom: 5px;">
-                            <i>${this.translations.expires}</i> ${remainingDays} ${this.translations.days},
-                            ${validTo.toLocaleString()}
+                            ${expirationText}
                         </div>
                     </div>`;
                 rows.push({ html: row, expirationDate: validTo });
--- a/src/opnsense/www/js/widgets/Certificates.js
+++ b/src/opnsense/www/js/widgets/Certificates.js
@@ -49,10 +49,31 @@ export default class Certificates extends BaseTableWidget {
         return $container;
     }
 
+    // Do not trigger danger modal if endpoint does not exist
+    async endpointExists(url) {
+        return new Promise((resolve) => {
+            $.ajax({
+                url: url,
+                type: 'HEAD',
+                success: function() {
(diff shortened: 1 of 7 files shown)


src/opnsense/www/js/widgets/Certificates.js
```diff
@@ -0,0 +1,177 @@
...
+                    ? 'text-danger'
+                    : remainingDays < 14
+                        ? 'text-warning'
+                        : 'text-success';
+
+                const statusText = remainingDays === 0 ? this.translations.expired : this.translations.valid;
+
+                const row = `
+                    <div>
+                        <i class="fa fa-lock ${colorClass} certificate-tooltip" style="cursor: pointer;"
```
fichtner, line 90: can we do fa-unlock for expired? it will be so much cooler :D

src/opnsense/www/js/widgets/Certificates.js
```diff
@@ -0,0 +1,177 @@
...
+                            data-tooltip="${type}-${item.descr}" title="${statusText}">
+                        </i>
+                        &nbsp;
+                        <span><b>${item.descr}${type === 'ca' ? ' (CA)' : ''}</b></span>
```
fichtner, line 94: Not sure this is needed, in all sane cases the subject should be clear

fichtner: (and a CA could also be a self-signed certificate)

fichtner:
general question: are we able to open the edit page for a certificate via link yet?

Monviech:
I saw that in the user manager there is some kind of link that can open the certificates bootgrid with a preapplied filter to only show a single certificate (for the user certificate). Maybe that could be added? But I have to see how it works first.

src/opnsense/www/js/widgets/Certificates.js
Monviech: Sure ^^

src/opnsense/www/js/widgets/Certificates.js
Monviech: Yeah you are right.

Monviech:
> general question: are we able to open the edit page for a certificate via link yet?

@fichtner I handled this with a trick that will populate the existing search-field with the common name of the certificate. This leverages existing things without creating too much overhead.

If a better solution can be found, I'm all for it, yet this piece of code is universal and can be dropped into any page that needs this functionality.

fichtner:
might be worth asking @AdSchellevis for an opinion, for me it's more difficult, but in general it would be nice opening the edit page so the save would immediately rewnew (I found out the hard way that it works this way).

AdSchellevis:
the "hacky" option of editing a uuid when jumping into the form:

https://github.com/opnsense/core/blob/f4b9017cd9290a74854dee653b8c9c135d4ca4ba/src/opnsense/mvc/app/views/OPNsense/Firewall/alias.volt#L179-L186

but if we do want to use this more often, we likely better implement it in the bootgrid for reuse

Monviech:
@AdSchellevis I have pushed the current code of what I showed you earlier. If you like it, we could consolidate it to a different spot so it is not duplicated. I would need a pointer where to put it.

AdSchellevis:
@Monviech maybe keep it in the trust templates for now, if we are reusing the same pattern more often, it might get more clear where it belongs. This does work nice by the way.

Monviech:
@AdSchellevis Hehe thank you. This kinda fixes this indirectly:

https://github.com/opnsense/core/issues/7895

Maybe if we want to use it in more widgets we could put it in a different place.

swhite2 (approved):
Looks good to me :)

fichtner:
Nice work 👍

PR #8115, opened 2024-12-08
dashboard: Improved widget links with automatic edit and/or search

Monviech:
Fixes: https://github.com/opnsense/core/issues/7895

- Introduces a generalized function to handle "search" and "edit" in url hashes. 
- Can handle changing a tab before firing a search and/or edit.

Implementations:
- [x] Certificate widget (edit uuid)
- [x] Gateway widget (edit either name or uuid)
- [x] Wireguard widget (change tab to peers and search uuid)
- [x] IPsec Tunnels (search a session)
- [x] CARP (search an interface)
- [x] OpenVPN Client Connections (search a session)
- [x] OpenVPN Server Connections (search a session)

### DIFF
--- a/src/opnsense/mvc/app/views/OPNsense/IPsec/sessions.volt
+++ b/src/opnsense/mvc/app/views/OPNsense/IPsec/sessions.volt
@@ -116,6 +116,15 @@
         });
 
         updateServiceControlUI('ipsec');
+
+        /* for ipsec tunnels widget */
+        $('#grid-phase1').on("loaded.rs.jquery.bootgrid", function () {
+            handleSearchAndEdit('#grid-phase1');
+        });
+        $(window).on('hashchange', function () {
+            handleSearchAndEdit('#grid-phase1');
+        });
+
     });
 
 </script>
--- a/src/opnsense/mvc/app/views/OPNsense/IPsec/sessions.volt
+++ b/src/opnsense/mvc/app/views/OPNsense/IPsec/sessions.volt
@@ -121,9 +121,6 @@
         $('#grid-phase1').on("loaded.rs.jquery.bootgrid", function () {
             handleSearchAndEdit('#grid-phase1');
         });
-        $(window).on('hashchange', function () {
-            handleSearchAndEdit('#grid-phase1');
-        });
 
     });
 
--- a/src/opnsense/mvc/app/views/OPNsense/IPsec/sessions.volt
+++ b/src/opnsense/mvc/app/views/OPNsense/IPsec/sessions.volt
@@ -117,11 +117,8 @@
 
         updateServiceControlUI('ipsec');
 
-        /* for ipsec tunnels widget */
-        $('#grid-phase1').on("loaded.rs.jquery.bootgrid", function () {
-            handleSearchAndEdit('#grid-phase1');
-        });
-
+        /* for ipsec widget */
+        handleSearchAndEdit('#grid-phase1');
     });
 
 </script>
--- a/src/opnsense/www/js/opnsense_ui.js
+++ b/src/opnsense/www/js/opnsense_ui.js
@@ -730,14 +730,18 @@ $.fn.SimpleFileUploadDlg = function (params) {
  *      }
  * @param {*} params data structure to use for the select picker
  */
-$.fn.replaceInputWithSelector = function (data) {
+$.fn.replaceInputWithSelector = function (data, multiple=false) {
     let that = this;
     this.new_item = function() {
         let $div = $("<div/>");
         let $table = $('<table style="max-width: 348px"/>');
+        let $select = $('<select data-live-search="true" data-size="5" data-width="348px"/>');
+        if (multiple) {
+            $select.attr('multiple', 'multiple');
+        }
         $table.append(
             $("<tr/>").append(
-                $("<td/>").append($('<select data-live-search="true" data-size="5" data-width="348px"/>'))
+                $("<td/>").append($select)
             )
         );
         $table.append(
@@ -773,6 +777,9 @@ $.fn.replaceInputWithSelector = function (data) {
         $this_select.attr('for', $(this).attr('id')).selectpicker();
         $this_select.change(function(){
             let $value = $(this).val();
+            if (Array.isArray($value)) {
+                $value = $value.filter(value => value !== '').join(',');
+            }
             if ($value !== '') {
                 $this_input.val($value);
                 $this_input.hide();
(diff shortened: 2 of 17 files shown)


src/opnsense/www/js/widgets/OpenVPNClients.js
```diff
@@ -116,7 +116,7 @@ export default class OpenVPNClients extends BaseTableWidget {
                                    title="${client.status}">
                                 </i>
                                 &nbsp;
-                                <a href="/ui/openvpn/status">${client.common_name}</a>
+                                <a href="/ui/openvpn/status#search=${client.common_name}">${client.common_name}</a>
```
swhite2, line 119: The client connections are grouped by server, which has a "no clients connected" clickable link case as well. One could argue this server may also be filtered on.

https://github.com/opnsense/core/pull/8115/files#diff-d318e5384c1c7d42653b13d0fb1e444dd7a2c7251e587989a52364082cac953eR147

src/opnsense/www/js/opnsense_ui.js
```diff
@@ -809,3 +809,53 @@ $.fn.replaceInputWithSelector = function (data, multiple=false) {
...
+        const decodedPhrase = rawPhrase ? decodeURIComponent(rawPhrase.trim()) : null;
+        if (!decodedPhrase) return;
+
+        const searchField = $('.search-field');
+        if (searchField.val() !== decodedPhrase) {
+            searchField.val(decodedPhrase).trigger('keyup');
+        }
+
+        if (prefix === 'edit') {
+            $(gridSelector).one("loaded.rs.jquery.bootgrid", function () {
```
swhite2, line 851: Don't think it will happen often, but in theory attaching the event handler after we are triggering the condition for `loaded.rs.jquery.bootgrid` to occur (`trigger('keyup')`) might mean we miss the event. This should probably be moved up.

src/opnsense/www/js/widgets/Gateways.js
```diff
@@ -65,15 +65,15 @@ export default class Gateways extends BaseTableWidget {
             gateways: {
                 title: this.translations.title,
                 type: 'select_multiple',
-                options: this.cachedGateways.map((name) => {
+                options: this.cachedGateways.map(({ name, uuid }) => {
                     return {
-                        value: name,
+                        value: uuid,
```
swhite2, line 70: I suppose this implicates that current configurations are now invalid? or does `name == uuid` the the old vs new api call?

If invalid, good thing we now have https://github.com/opnsense/core/commit/b52055af0b2ae2851905c132b9597dbc7d07c28f + https://github.com/opnsense/core/commit/7373985f3b2b0344c1e2596bdbbb5b0870cadb57

src/opnsense/mvc/app/views/OPNsense/IPsec/sessions.volt
```diff
@@ -116,6 +116,15 @@
         });
 
         updateServiceControlUI('ipsec');
+
+        /* for ipsec tunnels widget */
+        $('#grid-phase1').on("loaded.rs.jquery.bootgrid", function () {
+            handleSearchAndEdit('#grid-phase1');
+        });
+        $(window).on('hashchange', function () {
```
swhite2, line 124: since we always seem to handle this case after the event registration, can this not be moved into `handleSearchAndEdit`? The identifier is the same

src/opnsense/www/js/widgets/Gateways.js
```diff
@@ -48,12 +48,12 @@ export default class Gateways extends BaseTableWidget {
     }
 
     async _fetchGateways() {
-        const data = await this.ajaxCall('/api/routes/gateway/status');
-        if (!data.items || !data.items.length) {
+        const data = await this.ajaxCall('/api/routing/settings/searchGateway/');
```
swhite2, line 51: the `status` symbol is now undefined, since the API changed we probably need to recheck the keys

src/opnsense/www/js/widgets/Gateways.js
Monviech: Yeah the configuration will be invalid since it will be mapped to uuid now. If its okay to push this as new default with your commits you made that ensure a default, we can leave it like this.

If not, I can revert this back to name but having uuid here would be better in the long run I think.

src/opnsense/www/js/widgets/Gateways.js
Monviech: I have fixed the status.

src/opnsense/www/js/widgets/OpenVPNClients.js
Monviech: Added :)

src/opnsense/www/js/opnsense_ui.js
Monviech: Thanks I think I fixed this.

src/opnsense/www/js/opnsense_ui.js
```diff
@@ -809,3 +809,59 @@ $.fn.replaceInputWithSelector = function (data, multiple=false) {
...
+            }
+        }
+
+        if (action) {
+            const [prefix, rawPhrase] = action.includes('=') ? action.split('=') : [null, null];
+            const decodedPhrase = rawPhrase ? decodeURIComponent(rawPhrase.trim()) : null;
+            if (!decodedPhrase) return;
+
+            if (prefix === 'edit') {
+                const editButton = $(`${gridSelector} .command-edit[data-row-id="${decodedPhrase}"]`);
```
swhite2, line 846: the edit button may not exist yet, as it is dependent on the front-end having rendered the row, which can only be guaranteed if the `loaded.rs.jquery.bootgrid` triggers again after `trigger('keyup')`.

Perhaps there's a better approach by talking to the bootgrid implementation directly. The edit button fires the bootgrid instance's `show_edit_dialog` function, maybe we can do something like this?

```
function handleSearchAndEdit(gridSelector, getEndpoint) {
let bootgridInstance = gridSelector.data("ui.bootgrid");
bootgridInstance.show_edit_dialog(null, getEndpoint);
}
```

Don't know if this works (.data() might contain a different key or nothing at all). But this allows us to skip any dependency on the front-end and just call the partial template directly.

swhite2: small correction: https://github.com/opnsense/core/blob/master/src/opnsense/www/js/opnsense_bootgrid_plugin.js#L392 is the correct function

Monviech:
As discussed with @swhite2 I will hand this over, since the most future proof spot is most likely integrated directly into the bootgrid plugin so that it handles the url hash functionality.

src/opnsense/www/js/opnsense_ui.js
swhite2: https://github.com/opnsense/core/pull/8115/commits/3091b3645cb1cf3f94e24fcc97590a617d4008c8

src/opnsense/www/js/widgets/Gateways.js
fichtner: trailing slash? besides it made the lint-acl unhappy:

```
Registered widget endpoints do not match:
<<<<<<< src/opnsense/www/js/widgets/Gateways.js
/api/routing/settings/searchGateway/
========
/api/routes/gateway/status
>>>>>>> src/opnsense/www/js/widgets/Metadata/Core.xml
```

swhite2: https://github.com/opnsense/core/commit/23ec00f641b300347124b89c27873975db029f20

Issue #8155, opened 2024-12-21
system: shells are now static in the model, but must be read from /etc/shells


fichtner:
**Important notices**

Our forum is located at https://forum.opnsense.org , please consider joining discussions there in stead of using GitHub for these matters.

Before you ask a new question, we ask you kindly to acknowledge the following:

- [x] I have read the contributing guide lines at https://github.com/opnsense/core/blob/master/CONTRIBUTING.md
- [x] I am convinced that my issue is new after having checked both open and closed issues at https://github.com/opnsense/core/issues?q=is%3Aissue

The new model makes shells static but that is not correct:

> Even though bash is installed (via pkg install bash) and appears in /etc/shells, it doesn't appear as a shell choice in the UI.

As a side quest the user says:

> If I choose another valid shell it still gets written as nologin in /etc/passwd and so I can't log in as that user via SSH.

https://forum.opnsense.org/index.php?topic=44771.0

AdSchellevis:
I know, but realistically there are not a lot of valid choices, I think I rather add some static options with a note (bash, when installed) than more glue for options hardly anyone uses.

bensmithurst:
I have created https://github.com/opnsense/core/pull/8156 which solves both problems for me, I will let you decide how to proceed.

Thanks

fichtner:
I am a bit in doubt that deviating from the previous, OS aligned, unproblematic behaviour gains much other than future reports.

Also, not having opnsense-shell as an explicit choice anymore seems strange. Another side effect from removing auth_get_shells().

AdSchellevis:
> Also, not having opnsense-shell as an explicit choice anymore seems strange.

I also expected opnsense-shell in the option list, but while looking at the code again, it is only offered for root... 

https://github.com/opnsense/core/blob/908f3ee73edc10ba147c8dddd6c6b946b1d01229/src/etc/inc/auth.inc#L221-L223

In which case the default choice will be non explicit the same:

https://github.com/opnsense/core/blob/908f3ee73edc10ba147c8dddd6c6b946b1d01229/src/etc/inc/auth.inc#L348-L349

In which case I don't think it's much of a loss here, unless we would like to offer it to non root admins as well (which is something we can do without issues).

[...]

PR #8156, opened 2024-12-21
Fix shell configuration and show installed shells in dropdown

bensmithurst:
- Add config.inc include into auth.inc, this is needed to verify user's permissions when setting a shell for a non-root user.  Without this the code always thinks they are not an admin, and defaults to nologin.
- Add available shells to shell dropdown by reading /etc/shells
- ~~Add help text noting that shells not installed by default will be installed if needed.~~
- ~~Add code to install bash if needed.~~

ref #8155

### DIFF
--- a/src/etc/inc/auth.inc
+++ b/src/etc/inc/auth.inc
@@ -33,6 +33,7 @@
 
 require_once("interfaces.inc");
 require_once("util.inc");
+require_once("config.inc");
 
 function auth_log($message, $prio = LOG_ERR)
 {
--- a/src/etc/inc/auth.inc
+++ b/src/etc/inc/auth.inc
@@ -345,7 +345,7 @@ function local_user_set(&$user, $force_password = false, $userattrs = null)
                 $backend = new OPNsense\Core\Backend();
                 $backend->configdpRun("firmware install", array('bash'), true);
             }
-	}
+        }
     }
 
     // XXX: primary group id can only be wheel or nobody, otherwise we should map the correct numbers for comparison
--- a/src/etc/inc/auth.inc
+++ b/src/etc/inc/auth.inc
@@ -338,6 +338,14 @@ function local_user_set(&$user, $force_password = false, $userattrs = null)
         $user_shell = $is_admin && !empty($user['shell']) ? $user['shell'] : '/usr/sbin/nologin';
         $user_group = $is_admin ? 'wheel' : 'nobody';
         $user_home = "/home/{$user_name}";
+
+        if (!file_exists($user_shell)) {
+            // If the shell doesn't exist and it's one we know how to install, then install it.
+            if ($user_shell === '/usr/local/bin/bash') {
+                $backend = new OPNsense\Core\Backend();
+                $backend->configdpRun("firmware install", array('bash'), true);
+            }
+	}
     }
 
     // XXX: primary group id can only be wheel or nobody, otherwise we should map the correct numbers for comparison
--- a/src/etc/inc/auth.inc
+++ b/src/etc/inc/auth.inc
@@ -33,7 +33,6 @@
 
 require_once("interfaces.inc");
 require_once("util.inc");
-require_once("config.inc");
 
 function auth_log($message, $prio = LOG_ERR)
 {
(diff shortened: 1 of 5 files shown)


src/etc/inc/auth.inc
```diff
@@ -337,6 +338,14 @@ function local_user_set(&$user, $force_password = false, $userattrs = null)
         $user_shell = $is_admin && !empty($user['shell']) ? $user['shell'] : '/usr/sbin/nologin';
         $user_group = $is_admin ? 'wheel' : 'nobody';
         $user_home = "/home/{$user_name}";
+
+        if (!file_exists($user_shell)) {
```
AdSchellevis, line 342: introduces side affects, we're most certainly not going to do this, also nog guaranteed to work when versions don't match for example (in which case more glue might be needed).

https://github.com/opnsense/core/blob/e0e7e0b638afae2b1e71d0e53ea8c41cb9bc3348/src/opnsense/scripts/firmware/install.sh#L41

src/opnsense/mvc/app/controllers/OPNsense/Auth/forms/dialogUser.xml
```diff
@@ -64,6 +64,7 @@
         <id>user.shell</id>
         <label>Login shell</label>
         <type>dropdown</type>
+        <help>Shells which are not part of the base install will be installed if necessary.</help>
```
AdSchellevis, line 67: see above.

src/opnsense/mvc/app/models/OPNsense/Auth/User.xml
```diff
@@ -30,6 +30,7 @@
                     <o0 value="/bin/csh">/bin/csh</o0>
                     <o1 value="/bin/sh">/bin/sh</o1>
                     <o2 value="/bin/tcsh">/bin/tcsh</o2>
+                    <o3 value="/usr/local/bin/bash">/usr/local/bin/bash</o3>
```
AdSchellevis, line 33: my suggestion was `/usr/local/bin/bash (when installed)`

AdSchellevis (changes requested):

src/etc/inc/auth.inc
bensmithurst: Fair enough, it was just an idea, but I did wonder if it was too risky.

src/etc/inc/auth.inc
```diff
@@ -33,6 +33,7 @@
 
 require_once("interfaces.inc");
 require_once("util.inc");
+require_once("config.inc");
```
fichtner, line 36: This looks wrong. The other too, historically, are wrong as well. Which script misses what?

src/opnsense/mvc/app/models/OPNsense/Auth/User.xml
```diff
@@ -30,6 +30,7 @@
                     <o0 value="/bin/csh">/bin/csh</o0>
                     <o1 value="/bin/sh">/bin/sh</o1>
                     <o2 value="/bin/tcsh">/bin/tcsh</o2>
+                    <o3 value="/usr/local/bin/bash">/usr/local/bin/bash (when installed)</o3>
```
fichtner, line 33: I’d really much rather have this implicit. It will be an unpluggable maintenance hotspot.

src/etc/inc/auth.inc
bensmithurst: Looks like sync_user.php was including auth.inc and config.inc in the wrong order - auth.inc calls functions when loaded which depend on the config already being loaded.  I've updated the PR now, maybe that's less wrong?

fichtner: Yes, nice catch ❤️

src/opnsense/mvc/app/models/OPNsense/Auth/User.xml
bensmithurst: Would it look something like this..?  Fumbling in the dark a bit as I'm not overly familiar with all of the intricacies of this MVC framework.

https://github.com/bensmithurst/opnsense-core/commit/0bb18c5badd39b05eed54c8e8edd6bbe7fad2d52

AdSchellevis:
@bensmithurst I implemented this a bit different in https://github.com/opnsense/core/commit/96a1145c3e453af7ac5cfdf6d035667b5b58bc9c , which is more inline with the other fields, the import order mismatch I've added separately in  https://github.com/opnsense/core/commit/adecb9ddf1ac00e6d4f47557cce717acedd7eb4b . Thanks for taking the time looking at this.

Issue #8161, opened 2024-12-22
GUI value of GRE tunnel config does not match the configured value


codiflow:
**Important notices**

Before you add a new report, we ask you kindly to acknowledge the following:

- [x] I have read the contributing guide lines at https://github.com/opnsense/core/blob/master/CONTRIBUTING.md
- [x] I am convinced that my issue is new after having checked both open and closed issues at https://github.com/opnsense/core/issues?q=is%3Aissue

**Describe the bug**

I have two GRE tunnels on my multi-WAN machine. The GUI shows the primary WAN as "local address" but for both tunnels the secondary WAN has been configured. Looks like being just a display issue.

**To Reproduce**

1. Create GRE tunnel on multi-WAN machine
2. Configure each tunnel with a different WAN
3. All GRE tunnels will have the same (primary) WAN shown in the list overview

**Expected behavior**
The really configured value is being shown.

**Screenshots**
![grafik](https://github.com/user-attachments/assets/795d1fcf-4315-4a52-a1a5-b5ac99dfde86)

![grafik](https://github.com/user-attachments/assets/a02d6e6d-8375-48ad-9605-4f9f0336e81a)

![grafik](https://github.com/user-attachments/assets/e2d53877-a563-4578-bb04-ee920f012b59)

**Environment**

Software version used and hardware type if relevant, e.g.:

`OPNsense 24.7.11_2-amd64`
`FreeBSD 14.1-RELEASE-p6`
`OpenSSL 3.0.15`

fichtner:
Not sure how to prove your theory: we don’t know the actual configured config.xml value and your interface names plus internal identifiers (which should match the config.xml value in the GRE).

Cheers,
Franco

codiflow:
I recognized the same behaviour while creating a GIF tunnel yesterday.

Maybe these values from config.xml are helping to proof this?

```
    <opt5>
      <if>gre0</if>
      <descr>GRE1</descr>
      <enable>1</enable>
      <lock>1</lock>
      <spoofmac/>
      <blockpriv>1</blockpriv>
      <blockbogons>1</blockbogons>
    </opt5>

[...]

    <opt11>
      <if>igc1</if>
      <descr>WAN2_O2</descr>
      <enable>1</enable>
      <lock>1</lock>
      <spoofmac/>
      <blockbogons>1</blockbogons>
      <ipaddr><REDACTED></ipaddr>
      <subnet>24</subnet>
      <gateway>WAN2_GWv4</gateway>
[...]

PR #8162, opened 2024-12-23
interfaces: fix value lookup in LinkAddressField #8161

fichtner:
As described in #8161 "WAN" is displayed because the default in the XML is "wan". `internalValue` stays "wan" for the lifetime of the field, which makes me question its use in `__toString()` and perhaps `getValidators()`. Thoughts?

### DIFF
--- a/src/opnsense/mvc/app/models/OPNsense/Interfaces/FieldTypes/LinkAddressField.php
+++ b/src/opnsense/mvc/app/models/OPNsense/Interfaces/FieldTypes/LinkAddressField.php
@@ -93,17 +93,16 @@ class LinkAddressField extends BaseField
     public function getValidators()
     {
         $validators = parent::getValidators();
-        if ($this->internalValue != null) {
-            $validators[] = new CallbackValidator(["callback" => function ($data) {
-                $messages = [];
-                if (isset(self::$known_addresses[$data])) {
-                    return $messages;
-                } elseif (!Util::isIpAddress($data)) {
-                    $messages[] = gettext('A valid network address is required.');
-                }
+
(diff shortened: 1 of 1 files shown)


src/opnsense/mvc/app/models/OPNsense/Interfaces/FieldTypes/LinkAddressField.php
```diff
@@ -113,10 +113,13 @@ public function getValidators()
      */
     public function getDescription()
     {
-        if (isset(self::$known_addresses[$this->internalValue])) {
-            return self::$known_addresses[$this->internalValue];
+        $value = (string)$this;
```
fichtner, line 116: we could make this explicit, since it relies on `__toString()` to work the magic here:

```suggestion
        $value = $this-->toString();
```

AdSchellevis: @fichtner we can use `$this->getCurrentValue()`, the string cast seemed to be missing indeed.

fichtner: Ah, ok, shouldn't basefield __toString also use getCurrentValue() then? Right now we need to override __toString and getCurrentValue because the current value is not internalValue :/

fichtner: (feel free to take this over, would probably take longer for me)

fichtner:
note to self: check post value validation, then merge

fichtner:
Will merge via git

PR #8163, opened 2024-12-23
mvc: fields should implement getCurrentValue() rather than __toString()

fichtner:
To ease this effort avoid having derived classes implement both `__toString` and `getCurrentValue` for consistency across value return. Also only ever return strings (empty if not otherwise possible).

### DIFF
--- a/src/opnsense/mvc/app/models/OPNsense/Base/FieldTypes/BaseField.php
+++ b/src/opnsense/mvc/app/models/OPNsense/Base/FieldTypes/BaseField.php
@@ -356,11 +356,11 @@ abstract class BaseField
 
     /**
      * return string interpretation of this field
-     * @return null|string string interpretation of this field
+     * @return string string interpretation of this field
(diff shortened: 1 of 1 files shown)


fichtner:
Included in #8162

AdSchellevis:
I think this is ok, the tricky/risky one is `UpdateOnlyTextField` if I'm not mistaken, but as long as the child class overwrites `__String()` we should be good.

https://github.com/opnsense/core/blob/1e788857a59204ec349e869d6a6d9babb0f445f2/src/opnsense/mvc/app/models/OPNsense/Base/FieldTypes/UpdateOnlyTextField.php#L41-L44

Although I do expect some impact in a couple of constraints (likely introduced here https://github.com/opnsense/core/commit/85721719af21a1c2bebf851de2621d1ecc894542)

fichtner:
Thanks, let this sit till January when I can test better.

Issue #8179, opened 2025-01-03
Third party software warning message is user-unfriendly


niklasfemerstrand:
When installing plugins from 3rd party repos [a warning message](https://github.com/opnsense/core/blob/master/src/opnsense/mvc/app/views/OPNsense/Core/firmware.volt#L581) is displayed to the user:

![opnsense1](https://github.com/user-attachments/assets/410dee1a-8cc8-4f5d-aea0-5e7a6e1154c0)

Because the warning message doesn't specify which software the user has clicked to install, they are at mercy of clicking the correct icon in a list that looks like the image below. This can be a little bit interesting on wider screens and it is impossible to tell from the warning message what is being installed and who the vendor is. The user does not know for certain what they are agreeing to install.

I propose that the warning message is changed to: "{{ package_name }} is provided by an external vendor, for more information contact {{ the vendor }}."

Apologies for not submitting a PR, but fixing this involves localization changes as well.

![opnsense2](https://github.com/user-attachments/assets/07683f25-5e29-4a11-83dc-800397998a6d)

Monviech:
The step of adding an additional repository to the firewall implies an advanced user knowing the implications and the contents of this repository.

Additionally the "Repository" is shown in the "Firmware: Plugins" table.

niklasfemerstrand:
The removal confirmation displays the plugin name. It would be good if the installation confirmation showed it too.

![Screenshot 2025-01-06 at 14 57 23](https://github.com/user-attachments/assets/9a2da5a9-cf11-469b-a6d1-40f5c67c39a7)

Monviech:
I'll look into it and see if it can be improved.

niklasfemerstrand:
@Monviech Hang on, I'm 1 min away from a PR that does it the same for installation as for removal. :-)

niklasfemerstrand:
This is a good compromise I think because it doesn't require any large localization changes.

PR #8188, opened 2025-01-06
Adds plugin_name to plugin removal confirmation window, closes https://github.com/opnsense/core/issues/8179

niklasfemerstrand:


src/opnsense/mvc/app/views/OPNsense/Core/firmware.volt
```diff
@@ -579,7 +579,7 @@
                     BootstrapDialog.show({
                         type:BootstrapDialog.TYPE_INFO,
                         title: "{{ lang._('Third party software') }}",
-                        message: "{{ lang._('This software package is provided by an external vendor, for more information contact the author')}}",
+                        message: "{{ lang._('This software package is provided by an external vendor, for more information contact the author')}}" + " <strong>" + plugin_name + "</strong>",
```
Monviech, line 582:
```suggestion
                        message: "{{ lang._('This software package is provided by an external vendor, for more information contact the author')}}" + " <strong>" + package_name + "</strong>",
```

Monviech (changes requested):

Monviech:
I don't see any harm. I'll leave it up to @fichtner

fichtner:
Unfortunately I'm not a fan of this. The removal dialog at least has a question and the attached package name.... and also because nobody asked for improvement here for many years. Completely reworking this may be better, but we cannot pull much information from the metadata (like author email or URL) and we don't want to anyway.

PR #8194, opened 2025-01-08
configctl service reload all hangs

bensmithurst:
While testing some changes I ran `configctl service reload all` and got a timeout error.  After doing some digging I found that this relates to netflow data collection and doesn't happen if netflow collection is disabled.

After a bit of poking around it is due to the daemon/samplicate processes keeping a pipe open shared with the reload_all script, which then never termintes.  Adding the `daemon -f` parameter fixes.


fichtner:
Merged, thanks!

Issue #7690, opened 2024-07-30
OpenVPN crashes when <auth-gen-token/> is not empty


stuart-edge:
|| || |Version|24.7_9|

If I configure anything in VPN --> OpenVPN --> Instances --> Instance --> Auth Token Lifetime
The OpenVPN service will not start or restart.
If I remove the value (blank) the OpenVPN service will start.
Also fails with the value set to 0
The value looks OK in the backup file.
<auth-gen-token>36000</auth-gen-token>

OpenVPN will not start with this config
          <maxclients/>
          <keepalive_interval/>
          <keepalive_timeout/>
          <reneg-sec>0</reneg-sec>
          <auth-gen-token>60</auth-gen-token>
          <redirect_gateway/>
          <route_metric/>

OpenVPN will run with this config
          <maxclients/>
          <keepalive_interval/>
          <keepalive_timeout/>
          <reneg-sec>0</reneg-sec>
          <auth-gen-token/>
          <redirect_gateway/>
          <route_metric/>

opnsense
Version | 24.7_9 |  
Architecture | amd64 |  
Commit | 0d38c7804 |  
Mirror | https://pkg.opnsense.org/FreeBSD:14:amd64/24.7 |  
Repositories | OPNsense |  
Updated on | Tue Jul 30 13:51:49 AEST 2024

fichtner:
Same thing I posted here and got no response https://www.reddit.com/r/opnsense/comments/1efq0cg/openvpn_crashes_when_authgentoken_is_not_empty/

> This might clash with another setting that the OpenVPN daemon complains about and refuses to start. Can you check your log?

stuart-edge:
I thinks this is the error. I gave up on OpenVPN and started using WireGuard.
/usr/local/opnsense/scripts/openvpn/ovpn_service_control.php: The command '/usr/local/sbin/openvpn --config '/var/etc/openvpn/instance-226a3474-1caa-4f0a-9cf1-e0cb671bb1d3.conf'' returned exit code '1', the output was ''

fichtner:
I think we know the service didn't start. The question is what OpenVPN service itself logged when it decided not to start ;)

hsiewert:
I have read the contributing guide lines at https://github.com/opnsense/core/blob/master/CONTRIBUTING.md
 I am convinced that my issue is new after having checked both open and closed issues at https://github.com/opnsense/core/issues?q=is%3Aissue

I guess, the problem is probably the same as described before:

### Environment
[...]

PR #8203, opened 2025-01-11
VPN: OpenVPN: Instances - fix "auth-gen-token" and "reneg-sec" value validation

devopsoffice-ivan:
Allow saving config when reneg-sec is set to zero and auth-gen-token has no value, but prevents saving config when both parameters are set to zero.
Setting both parameters to zero results invalid configuration with error message:
Options error: --auth-gen-token needs a non-infinite --renegotiate_seconds setting

### DIFF
--- a/src/opnsense/mvc/app/models/OPNsense/OpenVPN/OpenVPN.php
+++ b/src/opnsense/mvc/app/models/OPNsense/OpenVPN/OpenVPN.php
@@ -133,9 +133,15 @@ class OpenVPN extends BaseModel
                         $key . ".verify_client_cert"
                     ));
                 }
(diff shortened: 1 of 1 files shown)


fichtner:
Just for further reference this is related to #7690

devopsoffice-ivan:
Yes, #7690 fixes a bad config value combination for reneg-sec and auth-gen-token (reneg-sec == 0 and auth-gen-token > 0) but blocks setting reneg-sec == 0 and empty auth-gen-token value.
Setting both reneg-sec and auth-gen-token to zero results an invalid configuration also, so an empty auth-gen-token value should be used when reneg-sec is set to zero. In this case auth-gen-token is skipped from generated configuration, just reneg-sec is included and the result is a valid generated configuration.
This patch allows setting empty value for auth-gen-token when reneg-sec is set to zero.

PR #8206, opened 2025-01-11
Allow custom additions to sshd_config

ngreatorex:
## The problem
On my OPNsense router, I wish to make modifications to the local `sshd_config` file and have them persist across restarts and upgrades. At the moment, this is not possible as the `/usr/local/etc/ssh/sshd_config` file is generated and controlled by OPNsense and doesn't allow custom configuration options to be set anywhere.

## The solution
Modify the generation of the `/usr/local/etc/ssh/sshd_config` file to include any files in a local directory `/usr/local/etc/ssh/sshd_config.d` if that directory exists. 

## Testing
I have tested on my local OPNsense install that it works both with and without the directory `/usr/local/etc/ssh/sshd_config.d` existing.


fichtner:
I'm ok with this since this fits our options override policies, but:

1. needs documentation (docs) because not obvious otherwise.
2. we should create the directory through core package, best with a readme file explaining why the directory exists -- but I'm not sure how OpenSSH reacts to a non-SSH config file, perhaps with comments such as https://github.com/opnsense/core/blob/master/src/etc/dnsmasq.conf.d/README

ngreatorex:
@fichtner Thanks. I've added the directory. `/src/etc/ssh/**` was in the `.gitignore` file. As the `sshd_config.d` is a directory, I can't just override it with a `!/src/etc/ssh/sshd_config.d`, so I've replaced it with ignore lines for keys and configs. Happy to adjust further if there are other files it's necessary to ignore.

ngreatorex:
Will work on a documentation PR shortly

fichtner:
I like this approach outlined here by some Linux:

https://www.reddit.com/r/linuxquestions/comments/1600kte/comment/jxjn5hp/

     /etc/ssh/sshd_config.d/*.conf  files  are  included at the start of the
    configuration file,  so  options  set  there  will  override  those  in
    /etc/ssh/sshd_config.

I don't think we need to modify the gitignore as you can force files into ignored directories. The reason it is ignored is due to the "make mount" feature which can snatch your on disk files when you happened to mount /usr/local to the  core source directory. I think in src/etc/pkg it's the same now.

AdSchellevis:
In the long run we might want to add some (uniform) way to show there are overwrites in play. In terms of support these things are quite annoying, which doesn’t mean we shouldn’t do this, but in a later phase might want to add “something” (not sure what at the moment, just a brain dump)

fichtner:
@AdSchellevis Based on ACLs we could add persistent banners now that implement is-there-a-non-empty override directory. Making this work with multiple services is probably the most trick part. The nice thing is you are only going to see this in the GUI on the relevant page.

ngreatorex:
> I like this approach outlined here by some Linux:
> 
> 
> 
> https://www.reddit.com/r/linuxquestions/comments/1600kte/comment/jxjn5hp/
> 
> 
> 
>     /etc/ssh/sshd_config.d/*.conf  files  are  included at the start of the
> 
>     configuration file,  so  options  set  there  will  override  those  in
> 
>     /etc/ssh/sshd_config.
> 
> 

I have moved the include to the top of the file. Do you want me to also change it to be only *.conf that gets included?

> 
> I don't think we need to modify the gitignore as you can force files into ignored directories. The reason it is ignored is due to the "make mount" feature which can snatch your on disk files when you happened to mount /usr/local to the  core source directory. I think in src/etc/pkg it's the same now.

OK I have reverted the change to .gitignore and force included the README.

src/etc/inc/plugins.inc.d/openssh.inc
```diff
@@ -158,6 +158,11 @@ function openssh_configure_do($verbose = false, $interface_map = null)
     $sshport = isset($sshcfg['port']) ? $sshcfg['port'] : 22;
 
     $sshconf = "# This file was automatically generated by /usr/local/etc/inc/plugins.inc.d/openssh.inc\n";
+
+    if (is_dir('/usr/local/etc/ssh/sshd_config.d')) {
+        $sshconf .= 'Include /usr/local/etc/ssh/sshd_config.d/*' . PHP_EOL;
+    }
```
fichtner, line 164:
```suggestion
    $sshconf .= 'Include /usr/local/etc/ssh/sshd_config.d/*.conf' . PHP_EOL;
```

src/etc/inc/plugins.inc.d/openssh.inc
```diff
@@ -158,6 +158,11 @@ function openssh_configure_do($verbose = false, $interface_map = null)
...
+
```
fichtner, line 165:
```suggestion
```

nevermind this, quick and easy change here :)

src/etc/inc/plugins.inc.d/openssh.inc
```diff
@@ -158,6 +158,8 @@ function openssh_configure_do($verbose = false, $interface_map = null)
     $sshport = isset($sshcfg['port']) ? $sshcfg['port'] : 22;
 
     $sshconf = "# This file was automatically generated by /usr/local/etc/inc/plugins.inc.d/openssh.inc\n";
+
```
fichtner, line 161:
```suggestion
```

fichtner:
Merged, thanks!

fichtner:
@AdSchellevis @swhite2 how about something along the lines of 65fbf0b

fichtner:
That did not commit as planned... it's [515e56f](https://github.com/opnsense/core/commit/515e56fc8f65cd13e2b3bfc492862daafc1f4c06)

AdSchellevis:
@fichtner In principle I think that would do the trick, but I have to try it out. Question is if this also works in a similar way for Unbound and IPsec as these are the most likely other targets. Last minor thing is file access, in some (most?) cases we likely need to detach the call to test for the pattern as we’re not always allowed to search the path (when running as www user in the future)

can you dump this in a PR so we don’t forget about it?

PR #8222, opened 2025-01-17
acl: prioritize dashboard and ui routes for landing page

swhite2:
With the new user manager (and probably in the old situation as well), privileges can be assigned in any order, which means the ACL can start with an `api` pattern, which would redirect the user to "nothing", practically locking them out. The first ACL may also point to anything, where in reality it would make more sense to sort this to something that's prioritized - if selected.

Just a little PoC for discussion.


AdSchellevis:
@swhite2 https://github.com/opnsense/core/commit/5a9a551425afeadc19b7039dc8a6c152d0d36c23 should do the trick for what we want to achieve.

Issue #8139, opened 2024-12-17
Delegated DHCPv6 Static Reservations Convert Decimal to Hexidecimal in Lease


Ryushin:
**Important notices**

Before you add a new report, we ask you kindly to acknowledge the following:

- [ X] I have read the contributing guide lines at https://github.com/opnsense/core/blob/master/CONTRIBUTING.md
- [X ] I am convinced that my issue is new after having checked both open and closed issues at https://github.com/opnsense/core/issues?q=is%3Aissue

Running version: 24.7.11

When setting a reserved IP address in a delegated prefix, input is received is as a decimal, but the outputted reserved lease is in hexadecimal.

![DHCPv6_Static_Mappings](https://github.com/user-attachments/assets/0b355fbf-d3cd-4408-ad39-955a803686de)

![DHCPV6_Leases](https://github.com/user-attachments/assets/d1e3ef15-d7d6-418c-9636-6c888b27d943)

Both should be the same.  If ::0.0.0.10 is entered for the static lease the output should also be .10 and not .a.

fichtner:
In practice that’s impossible as a general rule. An IPv6 is just a 128 bit number and this information gets los somewhere, especially with compression in play.

Ryushin:
I agree the IPv6 addresses are in hex.  I'm saying that when you create the reserved IP address, the input is in decimal and the actual lease is converted to hex.

fichtner:
No, they are not `::0.0.0.28` or `::0.0.0.a` because the leases page does **not** show IPv4-mapped IPv6 address format.

To reiterate: you are using IPv4-mapped IPv6 address format. It has colons AND dots at the end which mimics an IPv4 address (mapped in IPv6). This is a convenience way of setting an IPv6, but the actual adress displayed as a normal IPv6 is what you see...

Ryushin:
This is under the ISC DHCPv6 Leases.  I'm not using any IPv4-mapped to IPv6.  The IPv6 /64 prefix is handed to the LAN from the WAN /56 prefix requested from my ISP.

Since the LAN has a /64 delegated prefix, the first four sets of numbers are already specified and I just specify the last four sets of numbers, so ::0.0.0.28 or whatever.

I've worked around this by converting the hexadecimal number I want as a static mapping to decimal number in the mappings.  So I wanted ::0.0.0.10 as the lease I entered ::0.0.0.16 in the static mapping.

I think this is a bug since this is related just to to DHCPv6 and is not related to DHCPv4.

[...]

PR #8228, opened 2025-01-21
mvc: fix NetworkValidator for IPv4-mapped addresses with netmask

jfieber:
The NetworkValidator fails on [IPv4-mapped addresses](https://www.rfc-editor.org/rfc/rfc4291#section-2.2), for example `::ffff:198.51.100.0/120`.  Inferring IPv4 from the presence of a period (.) fails as these IPv6 address forms contain periods. Inferring IPv6 from the presence of a colon (:) should be more robust.

### DIFF
--- a/src/opnsense/mvc/app/models/OPNsense/Base/Validators/NetworkValidator.php
+++ b/src/opnsense/mvc/app/models/OPNsense/Base/Validators/NetworkValidator.php
@@ -96,14 +96,14 @@ class NetworkValidator extends BaseValidator
                     } else {
                         $mask = $parts[1];
                         $value = $parts[0];
-                        if (strpos($parts[0], ".")) {
-                            // most likely ipv4 address, mask must be between 0..32
-                            if ($mask < 0 || $mask > 32) {
+                        if (strpos($parts[0], ":") !== false) {
+                            // probably ipv6, mask must be between 0..128
+                            if ($mask < 0 || $mask > 128) {
                                 $result = false;
                             }
                         } else {
-                            // probably ipv6, mask must be between 0..128
-                            if ($mask < 0 || $mask > 128) {
+                            // most likely ipv4 address, mask must be between 0..32
+                            if ($mask < 0 || $mask > 32) {
                                 $result = false;
                             }
                         }


AdSchellevis:
The question is if all consumers of `NetworkField` types actually support mapped addresses..... changing this in general might not be a good idea otherwise.

fichtner:
I think the motivation of this change is correct if the validator then proceeds to accept the mapped address.

The mapped use, though, is just for giving IPv6 only implementations an IPv4 address to use. Plus we already know the mapping is inconsistent as with the wrong prefix it will just take the IPv4 address and use its bits for IPv6 addressing. Depending on the application you may end up with an error passing such an address to the service via configuration, but that's not really our problem either.

AdSchellevis:
I'm just worried about the various users of the actual field type, if all do accept mapped addresses where they accept ipv6 addresses, it's not an issue, but given the number of field involved, someone has some validation todo before this is a good idea (e.g. https://github.com/search?q=repo%3Aopnsense%2Fcore%20NetworkField&type=code). 

If only one of our "consumers" doesn't accept it, we end up needing to refactor this again to make it optional....

fichtner:
This isn't a fix or feature for mapped addresses. It's a fix for IPv6 detection in an edge case.

fichtner:
BTW, the legacy pages already accepted this for years... without much noise except #8139 recently...

AdSchellevis:
I know, a lot of legacy validations aren't great, which is also a good reason to be careful changing generic behavior without testing the result.

jfieber:
I ran into this working on opnsense/plugins#4031 (because I'm tired of patching my named.conf template by hand after every update).  The Bind documentation expresses the default value for the dns64 `excluded` option in this format. To allow a user to the default plus another value, it seems reasonable to expect the validator to accept the form used in the documentation.

However, I've also determined I need a custom validator anyway for a better implementation of the fairly ubiquitous ` { <address_match_element>; ... };` construct is Bind, so for my purposes, it isn't critical that it be allowed in the Base NetworkValidator.

However, in the interest of consistency, note that only the netmask, eg `::ffff:198.51.100.0/120` is rejected while the non-netmask form, eg `::ffff:198.51.100.1` passes.  If there are consumers where this serialization can't be used, maybe those can be handled with a validator option to explicitly disallow the form?

AdSchellevis:
Sometimes the custom field types are more practical indeed. I'll take a look at the inconsistent matching when I have some time.

AdSchellevis:
@jfieber you're right, the current address validation does allow mapped addresses, I've tried a couple of services to make sure the value is accepted, which seems to be the case. So let's allow subnets as well to make it consistent.  Thanks!

Issue #8187, opened 2025-01-06
mvc: Create a method to allow the use of a bootgrid partial


Monviech:
**Important notices**

Before you add a new report, we ask you kindly to acknowledge the following:

- [x] I have read the contributing guide lines at https://github.com/opnsense/core/blob/master/CONTRIBUTING.md
- [x] I am convinced that my issue is new after having checked both open and closed issues at https://github.com/opnsense/core/issues?q=is%3Aissue

**Is your feature request related to a problem? Please describe.**

I would like to implement a solution for automating bootgrid tables: https://github.com/opnsense/core/tree/partial-bootgrid-table
It does need a new method and other considerations: https://github.com/opnsense/plugins/pull/4388#issuecomment-2539887327

**Describe the solution you like**

Right now, the partial requires a data structure like this:

https://github.com/opnsense/core/blob/85f562e4b5090eebf630af9312487bcf6fc92228/src/opnsense/mvc/app/views/layout_partials/bootgrid_table.volt#L35-L44

Here is what I imagine a `getFormGrids()` method to do:

There are a few fields with defaults derived from existing data in a form:
```
data-column-id: (string) Default is content of <form><field><id> - e.g. "peer.enabled" must become "enabled"
data-column-label: (string) Default is content of <form><field><label>
data-type: (string) Default is content of <form><field><type>; if <form><field><type> is 'checkbox' the content should become the string "boolean"
data-formatter: (string) Default is empty, if if <form><field><type> is 'checkbox' the content should become the string "boolean"
```
Additional fields that emit defaults:
```
data-column-visible: (boolean) Default is "true"
data-visible: (boolean) Default is "true"
data-width: (string) Default is empty
data-sortable: (boolean) Default is "true"
data-identifier: (boolean) Default is "false"
```
Special field to allow columns to be sorted, the method must sort this:
```
sequence: (integer) Default is 0, sorting is based on lower number = higher priority. If numbers are duplicate it will be first come first serve.
```

Each of these fields can be specified in <form><field><column> to overwrite their defaults and set their content directly.

Maybe a container could look like this:

```
<form> 
     <field> 
         <id>peer.name</id> 
         <label>Name</label> 
         <type>text</type> 
         <help>Peer name, there should be one entry matching this machines "This server name"</help>
         <column>
             <sequence>1</sequence>
             <data-visible>false</data-visible>
             <data-width>10em</data-width>
         </column>
     </field>
</form>
```
**Describe alternatives you considered**

Not doing this.

[...]

PR #8229, opened 2025-01-21
dns/unbound: Integrate layout_partials/base_bootgrid_table

Monviech:
Requires: https://github.com/opnsense/core/issues/8187

Will be one commit per changed volt page.

- [x] acl.volt
- [x] dot.volt (Shared between DotController and ForwardController)
- [x] overrides.volt

### DIFF
--- a/src/opnsense/mvc/app/controllers/OPNsense/Unbound/forms/dialogAcl.xml
+++ b/src/opnsense/mvc/app/controllers/OPNsense/Unbound/forms/dialogAcl.xml
@@ -4,6 +4,11 @@
         <label>Enabled</label>
         <type>checkbox</type>
         <help>Enable or disable this ACL.</help>
+        <grid_view>
+            <width>6em</width>
+            <type>boolean</type>
+            <formatter>rowtoggle</formatter>
+        </grid_view>
     </field>
     <field>
         <id>acl.name</id>
@@ -42,6 +47,9 @@
         <help>
             List of networks in CIDR notation to apply this ACL on. For example: 192.168.1.0/24.
         </help>
+        <grid_view>
+            <visible>false</visible>
+        </grid_view>
     </field>
     <field>
         <id>acl.description</id>
--- a/src/opnsense/mvc/app/views/OPNsense/Unbound/dot.volt
+++ b/src/opnsense/mvc/app/views/OPNsense/Unbound/dot.volt
@@ -1,5 +1,5 @@
 {#
- # Copyright (c) 2014-2024 Deciso B.V.
+ # Copyright (c) 2014-2025 Deciso B.V.
  # All rights reserved.
  #
  # Redistribution and use in source and binary forms, with or without modification,
@@ -70,12 +70,7 @@
             }
         }
 
-
-        /*************************************************************************************************************
-         * link grid actions
-         *************************************************************************************************************/
-
-        $("#grid-dot").UIBootgrid(
+        $("#{{formGridDot['table_id']}}").UIBootgrid(
                 {   'search':'/api/unbound/settings/search'+this_page+'/',
                     'get':'/api/unbound/settings/get'+this_page+'/',
                     'set':'/api/unbound/settings/set'+this_page+'/',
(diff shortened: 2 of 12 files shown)


src/opnsense/mvc/app/controllers/OPNsense/Unbound/forms/dialogAcl.xml
```diff
@@ -42,6 +47,9 @@
         <help>
             List of networks in CIDR notation to apply this ACL on. For example: 192.168.1.0/24.
         </help>
+        <grid_view>
+            <visible>false</visible>
```
Monviech, line 51: Something to find out: 
The network row was not rendered before in the grid. It's somehow empty when rendered even if it contains networks.

AdSchellevis: @Monviech I don't see an obvious reason why it doesn't return anything, but does it change when you alter the model field to not return a list? 

https://github.com/opnsense/core/blob/5d596601d46666a16641fe54dbb510fa47f035d9/src/opnsense/mvc/app/models/OPNsense/Unbound/Unbound.xml#L170

AdSchellevis:
@Monviech I haven't tried the result, but the code looks nice and tidy :)

src/opnsense/mvc/app/controllers/OPNsense/Unbound/forms/dialogAcl.xml
Monviech: 'networks' was missing here. I should make sure that all bootgrid columns are in the searchBase()

https://github.com/opnsense/core/blob/5d596601d46666a16641fe54dbb510fa47f035d9/src/opnsense/mvc/app/controllers/OPNsense/Unbound/Api/SettingsController.php#L275

src/opnsense/mvc/app/views/OPNsense/Unbound/overrides.volt
```diff
@@ -26,121 +26,91 @@
...
-                $("div.actionBar").each(function(){
-                    if ($(this).closest(".bootgrid-header").attr("id").includes("alias")) {
-                        $(this).parent().prepend($('<td id="heading-wrapper" class="col-sm-2 theading-text">{{ lang._('Aliases') }}</div>'));
-                        $(this).removeClass("col-sm-12");
-                        $(this).addClass("col-sm-10");
-                        heading_appended = true;
+                    return row.server;
+                },
+                /* commands and rowtoggles added here since adding a custom formatter removes these by default for some reason */
+                "commands": function (column, row) {
```
swhite2, line 49: both the `"commands"` and `"rowtoggle"` formatter have been copied from `opnsense_bootgrid_plugin.js`. The reason for this was that specifying a custom formatter overwrites the entire `formatters` object.

I don't know if this is still the case (I imagine it is), but this probably needs a second look (i.e. merging objects). This is outside the scope of this PR, but wanted to mention it as we are moving the code around now.

AdSchellevis (changes requested):
some small remarks, but this cleans up quite some code :)

src/opnsense/mvc/app/views/OPNsense/Unbound/dot.volt
```diff
@@ -173,6 +136,9 @@
         <br/><br/>
     </div>
 </div>
+<div id="DotChangeMessage" class="alert alert-info" style="display: none" role="alert">
```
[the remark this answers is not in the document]
Monviech: I wanted to avoid the displacement of the "Apply" button. It's moved down and up whenever the Change message appears. In my opinion its more user friendly if the Apply button is static at the same spot and does not move around.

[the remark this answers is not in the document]
AdSchellevis: I agree on the static part, but as most of them now don't leave the button static, it might be better to discuss that for a future change. This change stretches the lower box which also doesn't look very nice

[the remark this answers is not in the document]
Monviech: Okay I'll clean this up and put it back in the old spot, and also do the same in my other currently open PRs.

PR #8234, opened 2025-01-22
system: upgrade Chart.js to v4.4.1

swhite2:
A couple of known limitations:

- There is a bug in v4 that causes charts to not scale up with a widget being expanded. This commit includes a CSS workaround, however, this causes dead spacing for certain widgets.
- Dataset-local `backgroundColor` properties seem to be ignored, even though there are no migration notes around this area. Currently the default color maps are used.

- The plugins we use aren't officially support in v4, see inner commit message.

There may be other issues.


fichtner:
thanks! 👍

PR #8241, opened 2025-01-23
Added 2FA logic on login screen

smitty-nieto:
PR : 6310

Added 2FA to Login screen

### DIFF
--- a/src/www/authgui.inc
+++ b/src/www/authgui.inc
@@ -128,6 +128,15 @@ function check_security_http_referer_enforcement()
     return false;
 }
 
+function isMFARequired($config) {
+    global $config;
+    $authMode = $config['system']['webgui']['authmode'] ?? 'Local Database';
+    $authServer = array_filter($config['system']['authserver'] ?? [], function($server) use ($authMode) {
+        return $server['name'] === $authMode && $server['type'] === 'totp';
+    });
+    return !empty($authServer);
+}
+
 function session_auth()
 {
     global $config;
@@ -145,6 +154,8 @@ function session_auth()
         session_set_cookie_params(["SameSite" => "Lax"]);
         session_start();
     }
+
+    $showMFAField = isMFARequired($config);
 
     // Detect protocol change
     if (!isset($_POST['login']) && !empty($_SESSION['Username']) && $_SESSION['protocol'] != $config['system']['webgui']['protocol']) {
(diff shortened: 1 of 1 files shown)


src/www/authgui.inc
```diff
@@ -409,6 +442,22 @@ function display_login_form($Login_Error)
           <label for="passwordfld"><?=gettext("Password:"); ?></label>
           <input id="passwordfld" type="password" name="passwordfld" class="form-control pwd" tabindex="2" />
         </div>
+<?php
+    // Check configuration for authmode and authserver type
```
rajiv, line 446: Instead of repeating this logic here, can you call `isMFARequired()` from above?

smitty-nieto: Absolutely. I have adjusted and pushed up the change.

smitty-nieto:
When can we expect this to merge upstream ? :)

src/www/authgui.inc
```diff
@@ -409,6 +442,16 @@ function display_login_form($Login_Error)
...
           <input id="passwordfld" type="password" name="passwordfld" class="form-control pwd" tabindex="2" />
         </div>
+<?php
+    // Determine if we need to display the MFA code field
+    $showMFAField = isMFARequired($config);
+
+    if ($showMFAField): ?>
+        <div class="form-group">
+            <label for="mfacode"><?= gettext("MFA Code:"); ?></label>
+            <input type="text" class="form-control" name="mfacode" id="mfacode" value="" />
```
0xThiebaut, line 452: Are there cases where MFA wouldn't be a number? Otherwise we might want to improve browser hinting for the auto-completion (e.g., password manager) by adding `inputmode="numeric"`, `autocomplete="one-time-code"` as well as `pattern="\d*"`.

0xThiebaut (changes requested):

danohn:
Any updates on this PR 8 months later? Would love to see this implemented.

fichtner:
As said elsewhere we're against a third field on the same page. The next person doesn't want it for valid reasons and the token is best asked after the password was entered which doesn't fit our local authentication flow. OpenIDC is coming to business 25.10 release as well which means that people are redirected for login anyway.

fichtner:
As mentioned we need a separate page after login to ask for the second factor. This requires reworking the login sequence and that's not on the projects agenda for the foreseeable future.


Cheers,
Franco

PR #8275, opened 2025-02-02
opnsense: Add support for forward-first when configuring forwarders

planetf1:
Description:

- Adds support for 'forward-first' configuration for unbound forwarders (regular & TLS)

Based on/tested on 24.10 (with an in-situ make mount to run master live)

Rationale:

When using TLS, or other, forwarding, if all servers become unavailable, unbound will return SERVFAIL.
By [setting 'forward-first' ](https://unbound.docs.nlnetlabs.nl/en/latest/manpages/unbound.conf.html)unbound will instead fallback to recursive resolution.

In my case I typically use quad9 (9.9.9.9) but very very occasionally they have a glitch and resolution fails. For example this [happened last week](https://www.reddit.com/r/Quad9/comments/1ieb266/problems_in_uk/) in parts of the UK

Implementation:
- Added new checkbox to UI (controller) when setting up a forwarding server. I did try to set the advanced flag, however for DoT the advanced settings never appeared. I therefore removed the advanced option - but if reviewers think this a better choice please advise how to set this?
- Added new option in model
- Updated the generator for dot.conf so that if this value is set on any forwarder configured for a domain, it is set for the domain in the generated unbound configuration file.

Test:
Verified setting on/off flag for both regular forwarded and dot

Example config with the flag set:
```
# Forward zones
forward-zone:
  name: "google.com"
  forward-addr: 9.8.7.6
  forward-first: yes

# Forward zones over TLS
server:
  tls-cert-bundle: /usr/local/etc/ssl/cert.pem
forward-zone:
  name: "."
  forward-tls-upstream: yes
  forward-addr: 8.4.1.9
  forward-first: yes
```

These are invalid nameservers. With the option set, DNS resolution is slow - until the SERVFAIL timeout occurs, but then resolution does succeed. This can therefore add a level of resilience in the situation where a primary resolver network is misbehaving.

Note that I haven't not tested timeouts etc in detail, for example if multiple resolvers are setup. This is getting into the actual functionality supported by unbound. The intent here is just to expose this configuration in opnsense

Comments/feedback welcome. I hope this makes sense and could be a useful enhancement

### DIFF
--- a/src/opnsense/mvc/app/controllers/OPNsense/Unbound/forms/dialogDot.xml
+++ b/src/opnsense/mvc/app/controllers/OPNsense/Unbound/forms/dialogDot.xml
@@ -48,6 +48,19 @@
         </grid_view>
     </field>
     <field>
+        <id>dot.forward_first</id>
+        <label>Forward first</label>
+        <type>checkbox</type>
+        <help>
+            If a forwarded query is met with a SERVFAIL error, and this option is enabled, Unbound will fall back to normal recursive resolution for this query as if no query forwarding had been specified. The fallback will only occur after a delay, so consider refining any server timeouts as needed. Please note this setting applies to the domain, so when multiple forwarders are defined for the same domain, all are assumed to use this setting. 
+        </help>
+        <grid_view> 
+            <type>boolean</type> 
+            <formatter>boolean</formatter> 
+            <visible>false</visible> 
+        </grid_view>
+    </field>
+    <field>
         <id>dot.verify</id>
         <label>Verify CN</label>
         <type>text</type>
--- a/src/opnsense/service/templates/OPNsense/Unbound/core/dot.conf
+++ b/src/opnsense/service/templates/OPNsense/Unbound/core/dot.conf
@@ -23,16 +23,20 @@ server:
 
 # Forward zones
 {%  for domain, forwards in all_forwards|groupby("domain", default=".") %}
-{% set domain_opts = namespace(forward_tcp_upstream=False) %}
+{% set domain_opts = namespace(forward_tcp_upstream=False, forward_first=False) %}
 forward-zone:
   name: "{{ domain }}"
 {%    for forward in forwards %}
   forward-addr: {{ forward.server }}{% if forward.port %}@{{ forward.port }}{% endif %}
 {%    set domain_opts.forward_tcp_upstream = domain_opts.forward_tcp_upstream or forward.forward_tcp_upstream == '1' %}
+{%    set domain_opts.forward_first = domain_opts.forward_first or forward.forward_first == '1' %}
 
 {%    endfor %}
 {%    if domain_opts.forward_tcp_upstream %}
   forward-tcp-upstream: yes
+{%    endif %}
+{%    if domain_opts.forward_first  %}
+  forward-first: yes
 {%    endif %}
 {%  endfor %}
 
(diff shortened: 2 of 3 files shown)


src/opnsense/mvc/app/controllers/OPNsense/Unbound/forms/dialogDot.xml
```diff
@@ -37,6 +37,16 @@
             all are assumed to use tcp only.
         </help>
     </field>
+    <field>
+        <id>dot.forward_first</id>
+        <label>Forward first</label>
+        <type>checkbox</type>
+        <help>
```
Monviech, line 44: Please use one string for the help text. It helps in translations.

src/opnsense/mvc/app/controllers/OPNsense/Unbound/forms/dialogDot.xml
```diff
@@ -37,6 +37,16 @@
...
+            If a forwarded query is met with a SERVFAIL error, and this option is enabled, Unbound will fall back to normal recursive resolution for this query as if no query forwarding had been specified.  
+            The fallback will only occur after a delay, so consider refining any server timeouts as needed.
+            Please note this setting applies to the domain, so when multiple forwarders are defined for the same domain, all are assumed to use this setting. 
+        </help>
```
Monviech, line 48: Please add this below help:

https://github.com/opnsense/core/blob/a3bdccd5e4f19a27c5ee37039e5abb02a5b90050/src/opnsense/mvc/app/controllers/OPNsense/Unbound/forms/dialogDot.xml#L44-L48

It controls how this new option will be rendered in the grid. (Its a recent addition so it was not there when you started the PR :)

src/opnsense/mvc/app/controllers/OPNsense/Unbound/forms/dialogDot.xml
planetf1: Ok - have concatenated to single string.

src/opnsense/mvc/app/controllers/OPNsense/Unbound/forms/dialogDot.xml
planetf1: done

planetf1:
Thanks for the review @Monviech . After making those changes I did also rebase off master (I saw the other grid settings). However a 'make mount' in /usr/core didn't see to give me a live UI preview as I had working before - but I think that's a local issue of mine.

Monviech (approved):
I think this is okay. I'll keep this open so @AdSchellevis can give a second opinion.

src/opnsense/service/templates/OPNsense/Unbound/core/dot.conf
```diff
@@ -29,26 +29,34 @@ forward-zone:
 {%    for forward in forwards %}
   forward-addr: {{ forward.server }}{% if forward.port %}@{{ forward.port }}{% endif %}
 {%    set domain_opts.forward_tcp_upstream = domain_opts.forward_tcp_upstream or forward.forward_tcp_upstream == '1' %}
+{%    set domain_opts.forward_first = domain_opts.forward_first or forward.forward_first == '1' %}
```
AdSchellevis, line 32: best initialize to False in `namespace()` for consistency

AdSchellevis: @planetf1 no problem, GitHub doesn't let you comment on unchanged lines, but my proposed change is in line 26:

```
{% set domain_opts = namespace(forward_tcp_upstream=False) %}
```

should probably be changed to the following for consistency:
 
```
{% set domain_opts = namespace(forward_tcp_upstream=False, forward_first=False) %}
```

planetf1: @AdSchellevis thanks for explaining!

Updated, checked locally (I needed to remove .mount_done). rebased & pushed.

AdSchellevis:
@planetf1 thanks for contributing!

PR #8286, opened 2025-02-04
Expose Kea option v6-only-preferred in the GUI

darses:
Pull requests to expose the Kea option `v6-only-preferred` in the GUI.

I have set the MinimumValue to MIN_V6ONLY_WAIT (300 sec) as defined in https://www.rfc-editor.org/rfc/rfc8925.html#name-constants-and-configuration. I have not set the default value as this would always include the `v6-only-preferred` option in the configuration.

I have tested the config generation using OPNSense 25.1 and can confirm that a working config is generated both with and without value defined.

### DIFF
--- a/src/opnsense/mvc/app/controllers/OPNsense/Kea/forms/dialogSubnet4.xml
+++ b/src/opnsense/mvc/app/controllers/OPNsense/Kea/forms/dialogSubnet4.xml
@@ -145,7 +145,7 @@
             <visible>false</visible>
         </grid_view>
     </field>
-        <field>
+    <field>
         <id>subnet4.option_data.v6_only_preferred</id>
         <label>IPv6-only Preferred</label>
         <type>text</type>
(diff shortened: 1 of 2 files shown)


src/opnsense/mvc/app/controllers/OPNsense/Kea/forms/dialogSubnet4.xml
```diff
@@ -145,4 +145,13 @@
...
         </grid_view>
     </field>
+    <field>
+        <id>subnet4.option_data.v6_only_preferred</id>
+        <label>IPv6-only Preferred</label>
+        <type>text</type>
+        <help>The number of seconds for which the client should disable DHCPv4</help>
+        <grid_view>
+            <visible>false</visible>
+        </grid_view>
```
AdSchellevis, line 155: perhaps add advanced here, using the following tag:

```
<advanced>true</advanced>
```

I doesn't feel like a very common option at a first glance, other than that, I'm fine with adding it

bluefire-fx:
Oh this is a wonderful PR. I've been experimenting with option 108 by directly editing /usr/local/etc/kea/kea-dhcp4.conf and then running keactrl reload as this bypasses the config generator.

AdSchellevis:
@darses looks good, thanks!

PR #8289, opened 2025-02-05
Expose radvd option nat64prefix (PREF64) in GUI 

### LINKED ISSUE #7487: Support PREF64 for Router Advertisements
**Important notices**

This request is relevant to IPv6-only networking, as this is one of the prerequisites for deploying IPv6-mostly network.

I would like to be able to specify PREF64 (RFC 8781) option for Router Advertisements.

N/A

**Additional context**

Relevant is [RFC 8781](https://datatracker.ietf.org/doc/html/rfc8781)

darses:
Expose the `nat64prefix` option in Router Advertisements in the GUI. This is an alternative to the existing DNS64 support in Unbound, which breaks DNSSEC. Support for this option appeared in radvd 2.20, which was added in OPNSense 25.1.

The option is added in the Router Advertisements advanced options section. I kept the lowercase naming to match the option in the config, but it does look a bit out of place in the GUI this way.

I did not add the `AdvValidLifeTime` option within as it conflicts with an existing advanced option with the same name. I also did not add any input validation, but can add some if this is requested.

Fixes #7487

### DIFF
--- a/src/etc/inc/plugins.inc.d/dhcpd.inc
+++ b/src/etc/inc/plugins.inc.d/dhcpd.inc
@@ -387,6 +387,11 @@ function dhcpd_radvd_configure($verbose = false, $blacklist = [])
             $radvdconf .= "\t};\n";
         }
 
+        if (!empty($dhcpv6ifconf['nat64prefix'])) {
+            $radvdconf .= "\tnat64prefix {$dhcpv6ifconf['nat64prefix']} {\n";
+            $radvdconf .= "\t};\n";
+        }
+
         $radvdconf .= "};\n";
     }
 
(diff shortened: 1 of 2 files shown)


fichtner:
Ideally we want to do this but we need to move the RA page to MVC/API first, which is going to be on the roadmao for 25.7 most likely.

darses:
@fichtner Now that the MVC migration is no longer in scope (per https://github.com/opnsense/core/issues/8351), could you take another look into this PR with the hope of merging it in 25.7?

fichtner:
Added in 6f47a0a with proper validation (requires MVC/API conversion code from #8351)

PR #8302, opened 2025-02-09
firewall/rules: Display fa icon if advanced features in firewall rules have been set

Monviech:
This could improve UX if many rules are present and a few of them have advanced options set.

### DIFF
--- a/src/etc/inc/filter.inc
+++ b/src/etc/inc/filter.inc
@@ -791,3 +791,28 @@ function get_tos_values($blank_label = '')
 
     return $ret;
 }
+
+/**
+ * check if advanced options are set in firewall_rules_edit.php or firewall_rules.php
+ */
+function FormSetAdvancedOptions(&$item) {
+    foreach (array("max", "max-src-nodes", "max-src-conn", "max-src-states","nopfsync", "statetimeout", "adaptivestart"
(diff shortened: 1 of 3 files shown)


src/etc/inc/filter.inc
```diff
@@ -791,3 +791,28 @@ function get_tos_values($blank_label = '')
 
     return $ret;
 }
+
+/**
+ * check if advanced options are set in firewall_rules_edit.php or firewall_rules.php
+ */
+function FormSetAdvancedOptions(&$item) {
```
AdSchellevis, line 798: I'm not a fan of moving this to a shared library as eventually both php files should go away at some point in time.

Monviech: Should we rather copy paste the function? Or would it be better to discard the idea? What do you suggest? :)

AdSchellevis: I'm not sure yet to be honest, in some cases it can be practical to know "other" options are used, but given the layout of the grid, that might even count for some of the non-advanced options. 

If you know "advanced" options are used, the next question would be, which then? and before you know it you're trying to pull everything into a view that's certainly not capable of serving that many viewpoints.

Monviech:
I'll close this one since the points being made do make sense. This does not fix the main issue that exists in the firewall grid in general.

PR #8307, opened 2025-02-11
Reporting: Health: Touch Up

NOYB:


fichtner:
Breaking the API compatibility on a stable branch that cannot be merged to avoid direct regressions like these? My kind of humour.

PR #8310, opened 2025-02-11
js/opnsense_ui: Add checkmark to SimpleActionButton on success

Monviech:
The Apply button does not give any feedback on success.

This PR makes a checkmark appear on success. It is consistent with how the dashboard handles its Save button, where the spinner and checkmark also appear on success.

The width of the Apply button stays consistent when the spinner and checkmark are added or removed.

### DIFF
--- a/src/opnsense/www/js/opnsense_ui.js
+++ b/src/opnsense/www/js/opnsense_ui.js
@@ -578,31 +578,43 @@ stdDialogRemoveItem.defaults = {
 $.fn.SimpleActionButton = function (params) {
     let this_button = this;
     this.construct = function () {
-        let label_content = '<b>' + this_button.data('label') + '</b> <i class="reload_progress">';
+        const label_content = '<b>' + this_button.data('label') + '</b> <i class="reload_progress" style="display:inline-block;"></i>';
         this_button.html(label_content);
         this_button.on('click', function () {
-            this_button.find('.reload_progress').addClass("fa fa-spinner fa-pulse");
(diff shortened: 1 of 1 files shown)


fichtner:
Frankly I'm not a fan of this behaviour: the spinner should spin while work is in progress. On abort the spinner should stop and show an API error or other message if relevant. The default outcome is success and indicating this further seems spurious. Just my 2 cents here.

swhite2:
> On abort the spinner should stop and show an API error or other message if relevant.

This, in my opinion, goes both ways for success or failure. Assuming "default" behavior might as well be just as spurious.

I think the checkmark approach strikes a nice balance between little visual noise and positive feedback, but it indeed requires that no checkmark is shown on failure in combination with an explicit error message (not sure if this is always possible), which is no different than the current situation.

The main concern I have here is that the definition for "success" from the API side might not be as consistent as we'd like, which may cause confusion down the line, which works in favor of @fichtner 's argument.

swhite2:
I'd argue this isn't for no reason, it's to indicate success. I do agree that this could feel like more time being taken up, but this would logically fade as soon as a user is used to it.

Frankly I don't mind either way.

fichtner:
I'm only asking myself if we change it because we want to fix it and if we fix it who had an issue with the current approach. With advancements like this it's much easier to talk UX fine print, but it also shows we don't have stringent UX guidelines (and no consistent current behaviour for that matter).

Monviech:
Let's keep things the way they are until somebody complains.

Issue #8318, opened 2025-02-12
mvc: change most frontend to base_bootgrid_table and base_apply_button


Monviech:
**Important notices**

Before you add a new report, we ask you kindly to acknowledge the following:

- [x] I have read the contributing guide lines at https://github.com/opnsense/core/blob/master/CONTRIBUTING.md
- [x] I am convinced that my issue is new after having checked both open and closed issues at https://github.com/opnsense/core/issues?q=is%3Aissue

**Is your feature request related to a problem? Please describe.**

Implement these two features in most pages in core:
https://github.com/opnsense/core/issues/8284
https://github.com/opnsense/core/issues/8187

Views:
`/usr/local/opnsense/mvc/app/views/OPNsense # find . -type f -name "*.volt"`

- [x] OpenVPN https://github.com/opnsense/core/pull/8332
    - [x] ./OpenVPN/instances.volt
    - [x] ./OpenVPN/cso.volt
- [x] TrafficShaper https://github.com/opnsense/core/pull/8524
    - [x] ./TrafficShaper/index.volt
- [x] DHCRelay https://github.com/opnsense/core/pull/8254 https://github.com/opnsense/core/pull/8314
    - [x] ./DHCRelay/relay.volt
- [x] Wireguard https://github.com/opnsense/core/pull/8313
    - [x] ./Wireguard/general.volt
- [x] Interface https://github.com/opnsense/core/pull/8257
    - [x] ./Interface/neighbor.volt
    - [x] ./Interface/vxlan.volt
    - [x] ./Interface/gif.volt
    - [x] ./Interface/vip.volt
    - [x] ./Interface/loopback.volt
    - [x] ./Interface/gre.volt
    - [x] ./Interface/vlan.volt
    - [x] ./Interface/lagg.volt
- [x] Auth https://github.com/opnsense/core/pull/8529
    - [x] ./Auth/priv.volt
    - [x] ./Auth/group.volt
    - [x] ./Auth/user.volt
- [x] CaptivePortal https://github.com/opnsense/core/pull/8846
    - [x] ./CaptivePortal/index.volt
- [x] Unbound https://github.com/opnsense/core/pull/8229 https://github.com/opnsense/core/pull/8315
    - [x] ./Unbound/dnsbl.volt
    - [x] ./Unbound/acl.volt
    - [x] ./Unbound/advanced.volt
    - [x] ./Unbound/overrides.volt
    - [x] ./Unbound/dot.volt
[...]

PR #8332, opened 2025-02-14
vpn/openvpn: Implement base_bootgrid_table and base_apply_button

Monviech:
For: https://github.com/opnsense/core/issues/8318

The whole sequence thing in the grid was unavoidable due to the way the old grid displayed some of its options in a different sequence than its form in the grid.

### DIFF
--- a/src/opnsense/mvc/app/views/OPNsense/OpenVPN/instances.volt
+++ b/src/opnsense/mvc/app/views/OPNsense/OpenVPN/instances.volt
@@ -94,6 +94,15 @@
 
 </script>
 
+<style>
+    /* The instances grid column dropdown has many items */
+    .actions .dropdown-menu.pull-right {
+        max-height: 400px;
+        min-width: max-content;
+        overflow-y: auto;
+        overflow-x: hidden;
+    }
+</style>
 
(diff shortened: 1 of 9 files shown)


Monviech:
Gonna think a bit more about this one if it can be streamlined a bit more.

src/opnsense/mvc/app/views/OPNsense/OpenVPN/instances.volt
```diff
@@ -94,78 +94,33 @@
 
 </script>
 
+<style>
+    /* The instances grid column dropdown has many items */
+    .actions .dropdown-menu.pull-right {
```
Monviech, line 99: @AdSchellevis Could you tell me where this style would need to go so it matches these grid dropdown-menus centrally? I think it would benefit other views too if they overflow like this per default.

AdSchellevis: @Monviech I'm not sure yet, if we add it to the theme it will invalidate existing ones. maybe keep it here for now and see if we need the same pattern more often.

AdSchellevis (approved):

Issue #8335, opened 2025-02-17
dashboard: wireguard.js - Change online/offline calculation of tunnels


Monviech:
**Important notices**

Before you add a new report, we ask you kindly to acknowledge the following:

- [x] I have read the contributing guide lines at https://github.com/opnsense/core/blob/master/CONTRIBUTING.md
- [x] I am convinced that my issue is new after having checked both open and closed issues at https://github.com/opnsense/core/issues?q=is%3Aissue

**Is your feature request related to a problem? Please describe.**

The wireguard widget could just use the last keepalive directly and compare it to a constant in seconds.
That avoids issues like these:

https://forum.opnsense.org/index.php?topic=45816.0

PR #8337, opened 2025-02-17
vpn/wireguard: Change tracking of wg peer status, improve widget and diagnostics

Monviech:
...to calculate if tunnel is online in backend. Implement it in wireguard.js widget and diagnostics.volt

Fixes: https://github.com/opnsense/core/issues/8335

### DIFF
--- a/src/opnsense/mvc/app/controllers/OPNsense/Wireguard/Api/ServiceController.php
+++ b/src/opnsense/mvc/app/controllers/OPNsense/Wireguard/Api/ServiceController.php
@@ -98,8 +98,10 @@ class ServiceController extends ApiMutableServiceControllerBase
 
             if (!empty($record['latest-handshake'])) {
                 $record['latest-handshake-age'] = time() - (int)$record['latest-handshake'];
+                $record['latest-handshake-epoch'] = date('Y-m-d H:i:s', (int)$record['latest-handshake']);
             } else {
                 $record['latest-handshake-age'] = null;
+                $record['latest-handshake-epoch'] = null;
             }
 
             // Peer is considered online if handshake was within 240s, wg handshakes approx every 120s.
--- a/src/opnsense/mvc/app/controllers/OPNsense/Wireguard/Api/ServiceController.php
+++ b/src/opnsense/mvc/app/controllers/OPNsense/Wireguard/Api/ServiceController.php
@@ -104,13 +104,15 @@ class ServiceController extends ApiMutableServiceControllerBase
                 $record['latest-handshake-epoch'] = null;
             }
 
-            // Peer is considered online if handshake was within 240s, wg handshakes approx every 120s.
-            if ($record['type'] === 'peer' &&
-                !is_null($record['latest-handshake-age']) &&
-                $record['latest-handshake-age'] <= 240) {
-                $record['peer-connected'] = true;
+            // Peer is considered online if handshake was within 300s, wg handshakes approx every 120s.
+            if ($record['type'] === 'peer' && !is_null($record['latest-handshake-age'])) {
+                if ($record['latest-handshake-age'] <= 300) {
+                    $record['peer-status'] = 'online';
+                } elseif ($record['latest-handshake-age'] > 300) {
+                    $record['peer-status'] = 'stale';
+                }
             } else {
-                $record['peer-connected'] = false;
+                $record['peer-status'] = 'offline';
             }
 
             $record['ifname'] = $ifnames[$record['if']];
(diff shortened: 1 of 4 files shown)


src/opnsense/www/js/widgets/Wireguard.js
```diff
@@ -77,17 +77,43 @@ export default class Wireguard extends BaseTableWidget {
...
+
+            latest_handshake_fmt: row['latest-handshake']
+                ? moment.unix(row['latest-handshake']).local().format('YYYY-MM-DD HH:mm:ss')
+                : null,
+
+            /**
+             *  Considered online if handshake was within 180s, wg does a handshake approx every 120s.
+             *  latest-handshake-age can be a valid 0 when it just happened.
+             */
+            connected: row['latest-handshake-age'] !== null
```
swhite2, line 109: probably best to do the comparison on the controller side as well, maybe include both `connected` and `latest-handshake-age`. At least this way 180 seconds is constant as well for any consumer of the API.

src/opnsense/www/js/widgets/Wireguard.js
```diff
@@ -77,17 +77,43 @@ export default class Wireguard extends BaseTableWidget {
...
+
+            tx: row['transfer-tx']
+                ? this._formatBytes(row['transfer-tx'])
+                : this.translations.notavailable,
+
+            // No fallback since we handle if 0
+            latest_handshake: row['latest-handshake'],
+
+            latest_handshake_fmt: row['latest-handshake']
+                ? moment.unix(row['latest-handshake']).local().format('YYYY-MM-DD HH:mm:ss')
```
swhite2, line 102: should also push to the controller for the same time consistency reasons

swhite2 (changes requested):

src/opnsense/www/js/widgets/Wireguard.js
Monviech: sounds good to me :)

src/opnsense/www/js/widgets/Wireguard.js
Monviech: https://github.com/opnsense/core/pull/8337/commits/46ae71f59cc665fd885fd0d7cbc16f9e0d69ff39

src/opnsense/www/js/widgets/Wireguard.js
Monviech: https://github.com/opnsense/core/pull/8337/commits/daec5e3a8e471662357fea0755038da54a832a5d

src/opnsense/mvc/app/controllers/OPNsense/Wireguard/Api/ServiceController.php
```diff
@@ -95,6 +95,24 @@ public function showAction()
...
+                $record['latest-handshake-epoch'] = date('Y-m-d H:i:s', (int)$record['latest-handshake']);
+            } else {
+                $record['latest-handshake-age'] = null;
+                $record['latest-handshake-epoch'] = null;
+            }
+
+            // Peer is considered online if handshake was within 240s, wg handshakes approx every 120s.
+            if ($record['type'] === 'peer' &&
+                !is_null($record['latest-handshake-age']) &&
+                $record['latest-handshake-age'] <= 240) {
```
Monviech, line 110: @fichtner Does this sound like a sane number? There is some randomness to the 120s, sometimes peers can take around 300s or more too for their handshakes I've seen in some of my tests.

If it's hard to pin down an exact number, could we make this configurable?

fichtner: Well, between 2 and 5 minute sounds about ok. Implying that it's a binary decision can be difficult.. "recently active" seems to resemble what it was intended to do so you can show green and grey instead of red and green.. "not connected" is much more problematic than shifting the time it takes to send the handshake again.

Monviech: So to reiterate:
- No handshake (handshake null): - Not connected (red)
- There was a handshake at some point: - Stale Connection (gray)
- There was a handshake within 5 minutes: - Active Connection (green)

fichtner: more or less, yes

note the UDP does not have a "connection" per se. I would show a cross, checkmark and perhaps question mark (or something more fitting)

Monviech: https://github.com/opnsense/core/pull/8337/commits/10933a660adbbc628f4ba000bcf4be7b150d0dd7

Sorry for the bigger diff, the ``let tunnel = newTunnels`` started at the wrong indentation.

fichtner (approved):
looks sane, thanks!

swhite2 (approved):
Looks good :)

Issue #8329, opened 2025-02-13
dnsmasq: offer DHCP functionality


AdSchellevis:
**Important notices**

Before you add a new report, we ask you kindly to acknowledge the following:

- [x] I have read the contributing guide lines at https://github.com/opnsense/core/blob/master/CONTRIBUTING.md
- [x] I am convinced that my issue is new after having checked both open and closed issues at https://github.com/opnsense/core/issues?q=is%3Aissue

**Is your feature request related to a problem? Please describe.**

In an attempt to deprecate ISC-DHCP we implemented KEA, which for enterprise environments is likely a good choice, but for smaller setups not always as practical, certainly when people are seeking a tight integration between `dhcp` and `dns` services (which can only be reached by implementing all sorts of hacks).

As we are  not planning to "bridge" unbound and kea, but would also like to support the small and medium sized setups, it would be practical to offer dhcp services using `dnsmasq` as this is what most systems use in similar cases.

**Describe the solution you like**

Add dhcp options to `dnsmasq` so we can switch to this option as default and deprecate isc-dhcp.

**Describe alternatives you considered**

Just drop isc in favor of KEA and dropping some of the current use-cases.

**Additional context**

* https://thekelleys.org.uk/dnsmasq/docs/dnsmasq-man.html

fichtner:
Roadmap discussion notes:

* MVC as is in early 25.1.x
* DHCP in later 25.1.x
* switch default DHCP in 25.7
* Look over DHCPv6/RADVD replacement options

AdSchellevis:
> 1) Will the complete — dhcp-host option be supported to manage static IPv4/IPv6?

not likely ipv4 and ipv6 are very different in that regard, please check https://thekelleys.org.uk/dnsmasq/docs/dnsmasq-man.html before opening a feature request.

> 2) Are you also planning to support --dhcp-hostsfile?
No, it doesn't look relevant from a ui perspective

gspannu:
> > 1. Will the complete — dhcp-host option be supported to manage static IPv4/IPv6?
> 
> not likely ipv4 and ipv6 are very different in that regard, please check https://thekelleys.org.uk/dnsmasq/docs/dnsmasq-man.html before opening a feature request.
> 
> > 2. Are you also planning to support --dhcp-hostsfile?
> >    No, it doesn't look relevant from a ui perspective

Thanks for the quick response… and _just to clarify Point (1)_.

[...]

PR #7852, opened 2024-09-09
Update ICANN Trust Anchor to include the new one

loganaden:
It will become active in 2026.

Signed-off-by: Jagveer Loky (jagveer@cyberstorm.mu)

PR #8355, opened 2025-02-19
dnsmasq: Backend migration and add dhcp support 

AdSchellevis:
for https://github.com/opnsense/core/issues/8329

This rather large commit implements most relevant dhcp options and rewrites dnsmasq's backend.

By default dnsmasq is disabled, eventually we do want dnsmasq enabled for dhcp services by default, but dns itself disabled. For this reason we support port "0" as implemented at dnsmasq (not listening for dns).

For cases where users want to integrate dns and dhcp services, the advise is to make dnsmasq listen on a non standard port and point unbound to the zones where dnsmasq is responsible for. This has the advantage of a direct connection between dhcp registered hosts and the requesting service. In these cases dnsmasq's dns service acts like a "connector".

In the long run we should deprecate `regdhcpstatic` and `regdhcp` as these either belong to legacy isc-dhcp or hook kea entries (which are better served via unbound).

The first mvc migration phase implemented IndexController.php, which we rename to SettingsController.php now as these results in more logical ui endpoints.

Since we don't bind to addresses directly (unless specifically configured and adviced only for static setups), we can skip the newwanip event which means we don't restart the service on interface changes. dnsmasq is able to filter the relevant networks on the fly, which is the advised scenario and can cope more easily with changes.

When different clients need to receive different options, we can use "tags" now. Requests can add tags to filter options which will be offered to the client, in the most simple scenario one would tag on a range or a host reservation, but more advanced choices can also be achieved using match statements (for example architecture [client-arch])


src/etc/inc/plugins.inc.d/dnsmasq.inc
```diff
@@ -103,102 +145,12 @@ function dnsmasq_configure_do($verbose = false)
...
-    } else {
-        $args .= ' --local-ttl=1 ';
-    }
-
-    $args .= ' --conf-dir=/usr/local/etc/dnsmasq.conf.d,\*.conf ';
-
     _dnsmasq_add_host_entries();
 
-    mwexec("/usr/local/sbin/dnsmasq {$args}");
+    mwexec("/usr/local/sbin/dnsmasq");
```
fichtner, line 150: can we consider rc.d scripting?

src/opnsense/mvc/app/models/OPNsense/Dnsmasq/Dnsmasq.xml
```diff
@@ -29,21 +32,45 @@
...
+        </dhcp_default_fw_rules>
+        <no_ident type="BooleanField">
+            <Required>Y</Required>
+            <Default>1</Default>
+        </no_ident>
         <hosts type="ArrayField">
             <host type="HostnameField">
                 <IsDNSName>Y</IsDNSName>
                 <IpAllowed>N</IpAllowed>
-                <Required>Y</Required>
```
fichtner, line 36: should we drop this in 25.1.2 already? same for domain

src/opnsense/service/templates/OPNsense/Dnsmasq/dnsmasq.conf
```diff
@@ -0,0 +1,133 @@
+# DO NOT EDIT THIS FILE -- OPNsense auto-generated file
```
fichtner, line 1: all the more reason for rc.d since the backend doesn't see this file and it may be missing on startup / using the old one?

src/opnsense/service/templates/OPNsense/Dnsmasq/trust-anchors.conf
```diff
@@ -0,0 +1,6 @@
+# The root DNSSEC trust anchor, valid as at 11/01/2019
```
fichtner, line 1: I like to avoid templating static files and create them directly at the target location -- this way it gets removed if we do not need it

AdSchellevis: @fichtner usually I would as well, but dnsmasq ships it already, but without the one we recently added.

fichtner: then this will flag in the health audit, yikes

src/opnsense/mvc/app/models/OPNsense/Dnsmasq/Dnsmasq.xml
AdSchellevis: @fichtner it's only not required when a mac address is offered (which is not in 25.1.2)

src/etc/inc/plugins.inc.d/dnsmasq.inc
fichtner: we have tricks for that: 

    dnsmasq_skip="YES"

and we keep steering through rc.d conf files but keep some of the legacy glue... it solves the pid vs disable issue BTW

src/opnsense/service/templates/OPNsense/Dnsmasq/dnsmasq.conf
AdSchellevis: we should be safe here, https://github.com/opnsense/core/blob/master/src/etc/rc.syshook.d/early/15-templates generates the file before anything wakes up.

src/opnsense/service/templates/OPNsense/Dnsmasq/trust-anchors.conf
fichtner: the last update of this file was 5 years ago

https://github.com/imp/dnsmasq/commits/master/trust-anchors.conf

I'm asking myself whether we're not trusting this file instead of fudging the anchors (I get why we did it from the command line). Now we don't need to if we trust them to update in time?

src/opnsense/service/templates/OPNsense/Dnsmasq/dnsmasq.conf
fichtner: there was a bug with configd regarding early start. it could still be happening. worst case I'll find out when testing images ;(

src/opnsense/service/templates/OPNsense/Dnsmasq/trust-anchors.conf
AdSchellevis: I can live with both options, it just means we will revert https://github.com/opnsense/core/pull/7852

fichtner: I'm not attached to it, the old one probably doesn't expire in 2026 either

src/opnsense/service/templates/OPNsense/Dnsmasq/dnsmasq.conf
AdSchellevis: I haven't seen the bug in ages to be honest, not sure if it makes sense to force another template reload just in case, but it's not the end of the world either if it makes you feel better.

src/opnsense/service/templates/OPNsense/Dnsmasq/trust-anchors.conf
AdSchellevis: ok, will trash this then. no problem, rather use defaults too, doubted a bit here.

fichtner (approved):
industry legend--dary ;)

src/opnsense/mvc/app/views/OPNsense/Dnsmasq/settings.volt
```diff
@@ -0,0 +1,207 @@
...
+        <hr/>
+        {{ partial('layout_partials/base_bootgrid_table', formGridDHCPboot)}}
+    </div>
+    <!-- Tab: DHCP Options / Match -->
+    <div id="dhcpmatches" class="tab-pane fade in">
+        {{ partial('layout_partials/base_bootgrid_table', formGridDHCPmatch)}}
+    </div>
+</div>
+
+<section class="page-content-main">
```
Monviech, line 185: Perfect spot for the apply button partial :)

AdSchellevis: @Monviech it is, forgot about it :) so, go ahead

PR #8364, opened 2025-02-20
mvc/view: Ensure fields stay aligned relatively to another when headers are used in forms

Monviech:
This fixes an issue where introducing headers in forms will botch the alignment of fields because they all spawn their own table. Using relative sizes in the table and colgroup fixes this issue.

This also improves responsiveness of the modal.

Bonus: This also improved how form modals render on mobile devices.

### DIFF
--- a/src/opnsense/mvc/app/views/layout_partials/base_dialog.volt
+++ b/src/opnsense/mvc/app/views/layout_partials/base_dialog.volt
@@ -1,5 +1,5 @@
 {#
- # Copyright (c) 2014-2015 Deciso B.V.
+ # Copyright (c) 2014-2025 Deciso B.V.
  # All rights reserved.
  #
  # Redistribution and use in source and binary forms, with or without modification,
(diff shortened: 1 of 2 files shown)


fichtner:
I'm a bit skeptical having dealt with % width whac-a-mole code changes over the last decade :/

Monviech:
I would like the opportunity to whack the mole too :D

Monviech:
Needed to force push this since I accidentally pushed something else in here that belongs to another branch.

src/opnsense/mvc/app/views/layout_partials/base_dialog.volt
```diff
@@ -66,11 +66,17 @@
             <div class="modal-body">
                 <form id="frm_{{base_dialog_id}}">
                   <div class="table-responsive">
-                    <table class="table table-striped table-condensed">
+                    <table class="table table-striped table-condensed" style="table-layout: fixed; width: 100%;">
                         <colgroup>
+                        {% if msgzone_width is defined %}
```
swhite2, line 71: Do we even need the `msgzone`? Seems it's only used by IDS

swhite2 (commented):
This does seem to improve modal/form consistency a lot :)

src/opnsense/mvc/app/views/layout_partials/base_dialog.volt
Monviech: I wanted to keep the change backwards compatible. I dont know the reason behind this in IDS.

swhite2: these properties should ideally live inside a form metadata tag (or if they're unique cases, handled by javascript), but the scope for that is larger than this change so not worth the effort to look into it here.

swhite2 (approved):
I don't mind this change, although considering @fichtner's reservations best if you sign off on this as well.

fichtner (approved):
we'll just have to see I guess

Monviech:
Let's give this a try then, if there are complains we can always revert it.

PR #8366, opened 2025-02-21
Update php.ini to add dynamic session.name value

### LINKED ISSUE #8365: Change PHP session name to be dynamic
**Important notices**

When running multiple OPNSense instances on the same hostname (e.g. SSH tunnel port forwards), you can only be logged into only one OPNSense at a time due to PHP using the default `PHPSESSID` session name.

As a workaround, I have added the following to the last line of `src/opnsense/service/templates/OPNsense/WebGui/php.ini`:

`session.name={{system.hostname}}`

1. Obviously, using different hostnames or different browsers, however this adds unnecessary overhead for simple use-cases.
2. Set the session name in the web UI, however that was beyond the scope of my needs.

**Additional context**

Steps to reproduce:

1. Have multiple instances of OPNSense running
2. SSH into two or more instances with port forward tunneling enabled (`ssh -L 8443:127.0.0.1:443 ..`)
3. Connect to all instances in the same web browser
4. Login to multiple instances and attempt to navigate around all of them
5. The last logged-in instance is the only one you will be able to continue using

Kage:
This PR adds `session.name={{system.hostname}}` to the last line of `src/opnsense/service/templates/OPNsense/WebGui/php.ini` so that OPNSense sessions do not use the PHP default `PHPSESSID` cookie name.  This PR permits multiple OPNSense instances to operate on the same hostname (e.g. via localhost from SSH tunneled sessions) without interfering with each other.

Fixes #8365

### DIFF
--- a/src/opnsense/service/templates/OPNsense/WebGui/php.ini
+++ b/src/opnsense/service/templates/OPNsense/WebGui/php.ini
@@ -35,3 +35,4 @@ error_log=/tmp/PHP_errors.log
 date.timezone="{{system.timezone|default('Etc/UTC')}}"
 session.save_path=/var/lib/php/sessions
 session.gc_maxlifetime={{system.webgui.session_timeout|default(240)|int * 60}}
+session.name={{system.hostname}}


AdSchellevis:
I'm not sure we should change the session names for all users (and leak hostnames in the process) for a single use-case. We'll leave this one open for now for further discussion.

Issue #8326, opened 2025-02-13
Captive Portal: replace ipfw implementation with pf


swhite2:
**Important notices**

Before you add a new report, we ask you kindly to acknowledge the following:

- [x] I have read the contributing guide lines at https://github.com/opnsense/core/blob/master/CONTRIBUTING.md
- [x] I am convinced that my issue is new after having checked both open and closed issues at https://github.com/opnsense/core/issues?q=is%3Aissue

**Is your feature request related to a problem? Please describe.**

With the recent efforts of integrating shaping options with pf rules we can start reducing the need for ipfw.

**Describe the solution you like**

Starting off, the captive portal ipfw dependency should be replaced by pf.

**Describe alternatives you considered**

None

**Additional context**

None

mimugmail:
Please consider mac bypass, not sure if pf support it already

mimugmail:
You got me wrong, I was just pointing not to forget about the existing mac bypass. Isnt this implemented via ipfw?

AdSchellevis:
ah, no plans to deprecate features, mac addresses are resolved, so no special features are used

swhite2:
https://github.com/opnsense/core/commit/d08addc25c42b0d53e9fd229e2ea642438184b69

AdSchellevis:
@swhite2 looks like theres a minor regression in this change, https://github.com/opnsense/core/commit/00e3d6d384754ac61f238b79ac79f79ee5b8148a implemented `interfaces_inbound`, which still exists as an option but isn't wired to anything anymore.

At a first glance I would expect the option should disappear, but I'll leave this for you.

swhite2:
> [@swhite2](https://github.com/swhite2) looks like theres a minor regression in this change, [00e3d6d](https://github.com/opnsense/core/commit/00e3d6d384754ac61f238b79ac79f79ee5b8148a) implemented `interfaces_inbound`, which still exists as an option but isn't wired to anything anymore.
> 
> At a first glance I would expect the option should disappear, but I'll leave this for you.

https://github.com/opnsense/core/commit/4567021437d75283951ee02a32e8f16781d20f1e should do it. As the commit message states, existing pf rules should now be enough to provide the same functionality.

PR #8368, opened 2025-02-21
Captive Portal: migrate to pf (https://github.com/opnsense/core/issues/8326)

swhite2:
Same logic as `ipfw`, in this case using transparent `match` rules inside anchors (one anchor per zone) to do the accounting. Since statistics for accounting can be fetched per anchor, the parsing overhead is relatively small.

With this change the following rules become redundant as they are generated automatically:
- access to DNS on the firewall
- access to captive portal on ports 8xxx/9xxx

A manual pass rule is still necessary to allow traffic from authenticated clients, however, non-authenticated clients will always be blocked regardless of rule setup.

Also updated the OPNsense logo in the default captive portal landing page.

Marking as draft until testing is completed.


src/opnsense/service/templates/OPNsense/Captiveportal/lighttpd-zone.conf
```diff
@@ -168,7 +168,7 @@ ssl.engine = "disable"
 
 #### compress module
 deflate.cache-dir        = "/tmp/"
-deflate.mimetypes        = ("text/plain", "text/html", "text/css", "image/png")
```
AdSchellevis, line 171: maybe better add svg, custom template may use png images still

src/opnsense/scripts/OPNsense/CaptivePortal/cp-background-process.py
```diff
@@ -50,18 +51,57 @@ def __init__(self):
...
+                        'last_accessed_hash': cur_hash,
+                        'in_pkts': 0,
+                        'in_bytes': 0,
+                        'out_pkts': 0,
+                        'out_bytes': 0
+                    }
+
+                    # merge counters as they are known in pf, last_accessed is
+                    # determined on next update to save us the work of querying the db stats
+                else:
```
AdSchellevis, line 96: we likely want to exclude all zero's here, a reload flushes the counters, which doesn't change access time. Had to read all of this a couple of times, but given the boundaries that exist, this does seem like a nice trick to offer a decent approximation of last accessibility.

swhite2: @AdSchellevis I updated the logic such that the decision for an accounting sync is now also determined in the same function, based on the state of the db (this state is always correct after a zone sync). This means I moved the accounting update after the zone sync.

If an accounting sync is necessary, all counters are flushed as you mentioned, therefore we can provide zeroes for the initial addition in memory and bootstrap it with the last known access time from the db (the next iteration will pick up on last access change). Since we now sync the accounting (insert match rules) after the accounting update to the db, the counters will reset immediately after collecting the last numbers which serves to minimize the amount of data we miss.

src/etc/inc/plugins.inc.d/captiveportal.inc
```diff
@@ -0,0 +1,190 @@
...
+ * ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
+ * POSSIBILITY OF SUCH DAMAGE.
+ */
+
+function captiveportal_services()
+{
+    $services = [];
+
+    $cp = new \OPNsense\CaptivePortal\CaptivePortal();
+    if ($cp->isEnabled()) {
```
AdSchellevis, line 34:
```suggestion
    if ((new \OPNsense\CaptivePortal\CaptivePortal())->isEnabled()) {
```

src/etc/inc/plugins.inc.d/captiveportal.inc
```diff
@@ -0,0 +1,190 @@
...
+}
+
+function captiveportal_firewall($fw)
+{
+    global $config;
+
+    $cp = new \OPNsense\CaptivePortal\CaptivePortal();
+    if ($cp->isEnabled()) {
+        foreach ($cp->zones->zone->iterateItems() as $zone) {
+            if (empty((string)$zone->enabled)) {
```
AdSchellevis, line 83:
```suggestion
            if ($zone->enabled->isEmpty()) {
```

we added these recently, practical to use in these conditions

src/etc/inc/plugins.inc.d/captiveportal.inc
```diff
@@ -0,0 +1,190 @@
...
+
+                foreach (['80', '443'] as $to_port) {
+                    $rdr_port = $to_port === '443' ? (8000 + (int)$zoneid) : (9000 + (int)$zoneid);
+
+                    // XXX the forwarding rules can be merged if
+                    // source inversion is supported by registerForwardRule
+                    // for now, skip rdr on the relevant table
+
+                    // disable rdr for authenticated clients, jump to filter pass rule
+                    $fw->registerForwardRule(
```
AdSchellevis, line 118: doesn't the `_not` suffix already work? (e.g. from_not)
https://github.com/opnsense/core/blob/03f96eb008eae0de6ba848511bfd6e3f4edabe47/src/opnsense/mvc/app/library/OPNsense/Firewall/Rule.php#L98-L100

all rules types use `reader()` to handle generics.

src/opnsense/mvc/app/controllers/OPNsense/CaptivePortal/Api/AccessController.php
```diff
@@ -256,18 +259,76 @@ public function logoffAction($zoneid = 0)
...
+                if ((time() - (int)$clientSession['startTime']) < (string)$zone->hardtimeout * 60) {
+                    $result['seconds-remaining'] = (string)$zone->hardtimeout * 60 - ((time() - (int)$clientSession['startTime']));
+                }
+            }
+
+            $this->response->setRawHeader("Cache-Control: private");
+            $this->response->setContentType("application/captive+json");
+
+            $result["captive"] = $captive;
+            $result["user-portal-url"] = "https://{$host}/index.html"; // XXX
```
AdSchellevis, line 323: out of curiosity, why the XXX?

src/opnsense/mvc/app/library/OPNsense/Firewall/Plugin.php
```diff
@@ -233,16 +233,26 @@ public function anchorToText($types = "fw", $placement = "tail")
...
+                        if (!empty($ifs)) {
+                            $result .= " on {" . implode(', ', $ifs) . "}";
+                        }
+                    }
                     $result .= "\n";
                 }
             }
         }
-        return $result;
+        return $result . "\n";
```
AdSchellevis, line 255: Not sure why we need an extra line break here, if we want extra spacing for readability, it might be better to add the `\n` to the `$result` line above (`\n\n` instead of `\n`)

src/opnsense/scripts/OPNsense/CaptivePortal/cp-background-process.py
```diff
@@ -34,11 +34,12 @@
 import traceback
 import subprocess
 import sqlite3
+import hashlib
```
AdSchellevis, line 37: unused?

AdSchellevis (changes requested):
@swhite2 a couple of small comments, but rest is looking good. nice work!

src/etc/inc/plugins.inc.d/captiveportal.inc
swhite2: neat :)

src/etc/inc/plugins.inc.d/captiveportal.inc
swhite2: @AdSchellevis Just tested it, seems `from_not` isn't quite there yet for both filter and forwarding rules.

```
'from_not' => '192.168.1.1/32' --> results in "any".
'from_not' => "__captiveportal_zone_{$zoneid}" --> skips parsing for fwd rule, "any" for filter rule
'from_not' => "<__captiveportal_zone_{$zoneid}>" --> same as above
```

The following does work for both filter and forward rules:
```
'source' => ['not' => true, 'address' => "<__captiveportal_zone_{$zoneid}>"]
```

AdSchellevis: @swhite2 `from_not` is a boolean, so it's `from` => `192.168.1.1/32` + `from_not` >=1

PR #8372, opened 2025-02-22
Extend Suricata support

0xThiebaut:
Extend Suricata support in OPNsense to increase the monitoring capabilities (e.g., integrate with Malcolm)

Features:
- [x] Extend support for EVE log, supporting all available protocols (RDP, DNS, ...).
- [x] Add support for PCAP logs.

Subsequent changes:
- [x] Update configuration template to align with Suricata 7.0.8 (e.g., add new fields such as `ja4`).
- [x] Disable `stats.log` (stats can be enabled in EVE).
- [x] Add migration logic.


src/opnsense/service/templates/OPNsense/IDS/suricata.yaml
```diff
@@ -254,121 +263,126 @@ outputs:
...
+        {{ '#' if helpers.empty('OPNsense.IDS.general.eveLog.krb5.enable') else '' }}- krb5
+        {{ '#' if helpers.empty('OPNsense.IDS.general.eveLog.bittorrentDht.enable') else '' }}- bittorrent-dht
+        {{ '#' if helpers.empty('OPNsense.IDS.general.eveLog.snmp.enable') else '' }}- snmp
+        {{ '#' if helpers.empty('OPNsense.IDS.general.eveLog.rfb.enable') else '' }}- rfb
+        {{ '#' if helpers.empty('OPNsense.IDS.general.eveLog.sip.enable') else '' }}- sip
+{%      if not helpers.empty('OPNsense.IDS.general.eveLog.quic.enable') %}
+        - quic:
+            # ja4 hashes in quic records will never be logged unless
+            # the following is set to on. (Default off)
+            ja4: {{ 'on' if helpers.empty('OPNsense.IDS.general.eveLog.tls.custom') or 'ja4' in OPNsense.IDS.general.eveLog.tls.custom else 'off' }}
```
0xThiebaut, line 336: QUIC's JA4 support is enabled if JA4 is enabled for TLS. We should consider whether a standalone setting for QUIC is preferred.

src/opnsense/mvc/app/controllers/OPNsense/IDS/forms/generalSettings.xml
```diff
@@ -88,87 +88,352 @@
...
+    </field>
+    <field>
+        <id>ids.general.eveLog.files.enable</id>
+        <label>Enable EVE's file logging</label>
+        <type>checkbox</type>
+        <help>Log file events to EVE log.</help>
+        <advanced>true</advanced>
+    </field>
+    <field>
+        <id>ids.general.eveLog.files.forceMagic</id>
```
Monviech, line 195: There seem to be a log of checkboxes. Maybe it would make more sense for one multi select option dropdown to save lots of code:
https://github.com/opnsense/plugins/blob/58c646121f2babb6f66ac47e2fb63c869af9924f/www/caddy/src/opnsense/mvc/app/models/OPNsense/Caddy/Caddy.xml#L113-L122

https://github.com/opnsense/plugins/blob/58c646121f2babb6f66ac47e2fb63c869af9924f/www/caddy/src/opnsense/mvc/app/controllers/OPNsense/Caddy/forms/general.xml#L35-L40

0xThiebaut: Pushed a new variant using multi-selects; it indeed looks a lot more user-friendly. Would this work for you? If so, I'll update the migration logic.

src/opnsense/mvc/app/controllers/OPNsense/IDS/forms/generalSettings.xml
```diff
@@ -5,76 +5,73 @@
     </field>
     <field>
         <id>ids.general.enabled</id>
-        <label>Enabled</label>
+        <label>Enable</label>
         <type>checkbox</type>
-        <help>Enable intrusion detection system.</help>
+        <help>Enable IDS (Intrusion Detection System).</help>
```
Monviech, line 10: Every help or label string you change bounces the translations of other languages. Could you reduce the scope of your PR to only the functionality you would like to add/change?

Monviech:
If you limit this PR only to the new functionality you want to add it will be simpler to review.

If you want to change existing language strings it could be in a separate PR after this one.

src/opnsense/mvc/app/models/OPNsense/IDS/IDS.xml
```diff
@@ -174,22 +174,6 @@
                 </Model>
                 <ValidationMessage>Related cron not found.</ValidationMessage>
             </UpdateCron>
-            <AlertLogrotate type="OptionField">
```
Monviech, line 177: Is it really necessary to alter the model like that? Can't we go for a minimal impact approach first?

src/opnsense/mvc/app/controllers/OPNsense/IDS/forms/generalSettings.xml
```diff
@@ -27,38 +27,35 @@
         <type>select_multiple</type>
         <help>Select interface(s) to use. When enabling IPS, make sure the (virtual) driver supports this feature.</help>
     </field>
-    <field>
```
Monviech, line 30: Why is this header deleted?

0xThiebaut: The header was deleted as I marked the `ids.general.MPMAlgo` options advanced, leading to an empty section when advanced options are not visible.

If we do consider this option as basic, I'll roll back its advanced tag and reinstate the section.

src/opnsense/mvc/app/models/OPNsense/IDS/Migrations/M1_1_1.php
```diff
@@ -0,0 +1,117 @@
...
+ */
+
+namespace OPNsense\IDS\Migrations;
+
+use OPNsense\Base\FieldTypes\BaseField;
+use OPNsense\Core\Config;
+use OPNsense\Base\BaseModelMigration;
+use OPNSense\IDS\IDS;
+
+class M1_1_1 extends BaseModelMigration
```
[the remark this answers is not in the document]
0xThiebaut: We definitely could add it without a migration; I focused on having a reasonable code-base with pleasant UI but if avoiding a migration is a priority I can add the new features without requiring a migration.

Could you confirm you'd prefer this? This would involve having a drop-down for new protocol types (e.g.,DNS) while retaining check boxes for already-supported log types (i.e., HTTP & TLS).

This wouldn't be consistent but would avoid the need for a migration.

src/opnsense/mvc/app/controllers/OPNsense/IDS/forms/generalSettings.xml
Monviech: You can add advanced to the header too. :)

0xThiebaut: Oh! I should have guessed that :) Will do!

src/opnsense/mvc/app/models/OPNsense/IDS/Migrations/M1_1_1.php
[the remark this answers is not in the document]
fichtner: We don't have this policy in place (yet_ but it would be a very good idea to either only introduce new changes or only move existing data, but not both in the same PR. This is beneficial in reducing risk of regression and considering backporting. On top of that suricata.yaml has broken a few times already so you have that risk as well on top.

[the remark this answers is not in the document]
0xThiebaut: Alright, with that guideline I'll make the necessary modifications to have this PR introduce the required functionality without it involving a migration.

I'll leave the refactoring for better UX for a later time.

0xThiebaut:
Will create a new PR once this has been fixed.

Issue #8367, opened 2025-02-21
firewall: automation: Explore UI possibilities (POC)


Monviech:
**Important notices**

Before you add a new report, we ask you kindly to acknowledge the following:

- [x] I have read the contributing guide lines at https://github.com/opnsense/core/blob/master/CONTRIBUTING.md
- [x] I am convinced that my issue is new after having checked both open and closed issues at https://github.com/opnsense/core/issues?q=is%3Aissue

**Is your feature request related to a problem? Please describe.**

Since the legacy firewall UI pages are at some point, well, um, legacy, there has to be a replacement at some point. The most valid replacement right now is "Firewall: Automation: Filter", which uses the modern mvc framework to generate and display its rules.

The only issue with that, is the User Experience. 10 or 20 firewall rules should not be an issue, but if the view scales up to 100 or 1000 elements even, there could be usability issues.

**Describe the solution you like**

Brainstorm the possibilites of improving UX in that view to prepare for the future:

Prime directives:
- The consistent sequence of rules in the view at all times
- Do not stray too far from what the framework established
    - (There is some hacky stuff in the FilterController to inject the legacy internal rules into searchRuleAction to show them in the bootgrid. Though the actual API output has not been altered, these rules can be added to the output via type in the request.)

General UX:
- [x] Move the current code to the reusable grid structure to expose all options
- [x] Ensure firewall rules cannot be sorted (would destroy consistent sequence)
- [x] Create custom formatters to replicate the old legacy firewall view as first step
- [x] Custom commands to change the sequence of a rule relative to the other rules, this needs some controller magic
- [x] Show change message when custom command was used.
- [x] Analyze the basic needs of a firewall rule, and selectively hide more fields in advanced view or auto collapsed headers inside the main form. For maintainability, having all fields in the same general form is the best.
- [x] Restrict size of bootgrid dropdown like here: https://github.com/opnsense/core/blob/7129728488b988b85a89e7d6d52b715b5cb4d7bf/src/opnsense/mvc/app/views/OPNsense/OpenVPN/instances.volt#L97
- [x] ~~Create AutoNumberField that automatically increases by 10, and not 1. This will help in larger rulesets where you want to set rules between other rules later on. If the sequence is always unbroken, youd need to move a rule up like 500 times or so after creation at the bottom of the ruleset. Increase allowed sequence to 99.999.~~
- [x] ~~Maybe increment by 100 per default and increase max sequence to 999.999 to leave more space inbetween rules.~~
[...]

PR #8377, opened 2025-02-25
Firewall: Automation filter poc

Monviech:
Fixes: https://github.com/opnsense/core/issues/8367

Opening this as draft to have a better overview of my current state with commits and diff.


Monviech:
@AdSchellevis I want to hand this over to you at the current state it is in.

AdSchellevis:
@Monviech I think the branch is almost in a functional state, in case you want to take a peek, just checkout the current branch.

Changes sofar:

- use a single checkbox to request extended information.
- remove savepoint and revert buttons as we likely don't want to offer them in the final ui anyway
- always generate a "sort_order" field and enforce this as ordering in the ui. Although it's not a very presentable number, it warrants consistency between the actual ruleset and the one presented.
- add a "legacy" flag to all records offered via "filter list non_mvc_rules"
- exclude pf_firewall rules in "filter list non_mvc_rules" to prevent duplication
- change the result format of getAdvancedIds() to ease the volt templates

Items still in need of attention:

- searchRuleAction() trickery around request handling to make sure we always receive the full set of items from searchBase()
- formatting of categories, this likely needs some trickery in the backend to pass the colors instead of the names
- better label for "Show all rules" and maybe also relocate the checkbox to the left
- move rules action, trickery around the renumbering, haven't looked at that yet, but might need some work
- check performance and memory requirements when there are more items in the mvc model.

We might want to fiddle a bit further with the layout as well, but let's first try to make it "feature complete".

src/opnsense/mvc/app/controllers/OPNsense/Firewall/Api/FilterController.php
```diff
@@ -27,17 +27,131 @@
...
     protected static $categorysource = "rules.rule";
 
+    private function getFieldMap()
+    {
+        $result = ['category' => [], 'interface' => []];
+        foreach ((new Category())->categories->category->iterateItems() as $key => $category) {
+            $result['category'][$key] = (string)$category->name;
+        }
+        foreach ((Config::getInstance()->object())->interfaces->children() as $key => $ifdetail) {
+            $descr = htmlspecialchars(!empty($ifdetail->descr) ? $ifdetail->descr : strtoupper($ifname));
```
Monviech, line 47: This line crashes for me:

[09-Mar-2025 18:29:43 Etc/UTC] ErrorException: Undefined variable $ifname in /usr/local/opnsense/mvc/app/controllers/OPNsense/Firewall/Api/FilterController.php:47

Maybe like this?

```suggestion
            $descr = htmlspecialchars(!empty($ifdetail->descr) ? $ifdetail->descr : strtoupper($key));
```

src/opnsense/mvc/app/controllers/OPNsense/Firewall/Api/FilterController.php
```diff
@@ -27,17 +27,131 @@
...
+
+        /* only fetch internal and legacy rules when 'show_all' is set */
+        if ($show_all) {
+            $otherrules = json_decode((new Backend())->configdRun("filter list non_mvc_rules") ?? [], true);
+        } else {
+            $otherrules = [];
+        }
+
+        $_REQUEST = $ORG_REQ; /* XXX:  fix me ?*/
+        return $this->searchRecordsetBase(array_merge($otherrules, $filterset), null, "sort_order", $filter_funct_rs);
```
Monviech, line 154: When enabling the "Show all rules" checkbox, I get this error:

[09-Mar-2025 18:35:18 Etc/UTC] TypeError: array_merge(): Argument #1 must be of type array, null given in /usr/local/opnsense/mvc/app/controllers/OPNsense/Firewall/Api/FilterController.php:154
Stack trace:
#0 /usr/local/opnsense/mvc/app/controllers/OPNsense/Firewall/Api/FilterController.php(154): array_merge(NULL, Array)
#1 /usr/local/opnsense/mvc/app/library/OPNsense/Mvc/Dispatcher.php(166): OPNsense\Firewall\Api\FilterController->searchRuleAction()
#2 /usr/local/opnsense/mvc/app/library/OPNsense/Mvc/Router.php(156): OPNsense\Mvc\Dispatcher->dispatch(Object(OPNsense\Mvc\Request), Object(OPNsense\Mvc\Response), Object(OPNsense\Mvc\Session))
#3 /usr/local/opnsense/mvc/app/library/OPNsense/Mvc/Router.php(139): OPNsense\Mvc\Router->performRequest(Object(OPNsense\Mvc\Dispatcher))
#4 /usr/local/opnsense/www/api.php(36): OPNsense\Mvc\Router->routeRequest('/api/firewall/f...', Array)
#5 {main}

src/opnsense/mvc/app/controllers/OPNsense/Firewall/Api/FilterController.php
AdSchellevis: @Monviech got it, thanks, the `htmlspecialchars()` shouldn't be needed either. I'll push a fix in the next commit

src/opnsense/mvc/app/controllers/OPNsense/Firewall/Api/FilterController.php
AdSchellevis: @Monviech probably configd not reloaded, but it shouldn't crash, the default seems to be in the wrong spot, thanks!

Monviech: Thanks for your work on this Ad. Im looking at it tomorrow. :)

AdSchellevis: No rush, just wanted to get it off my list :)

Monviech:
@AdSchellevis 

I have addressed these points:

- formatting of categories, this likely needs some trickery in the backend to pass the colors instead of the names
   -> Works now
- better label for "Show all rules" and maybe also relocate the checkbox to the left
   -> Works and there is some new trickery surrounding it, e.g. only showing statistic columns while its active 
- move rules action, trickery around the renumbering, haven't looked at that yet, but might need some work
   -> Logic has been improved:
   -> There are gaps again, the auto number field increments by 100, there is a hook to the clone action so the sequence is always rendered unique on cloning.
   -> When rules are moved, the delta between rules is used to calculate the new spot. If no delta exists anymore, then as fallback the whole ruleset is renumbered with base 100 and increment 100
   -> Using sort_order as a trick, e.g., an interface rule cannot be moved before a floating rule or group rule, it emits an error

I've also fiddled around with the layout, making everything more neat with new icons (e.g. an icon for the rule type), Improved the interface select (adding headers and group the elements inside) and made the whole layout more responsive with CSS so it works and looks fine on small devices.

What I need from you now is that you look into these points:

- searchRuleAction() trickery
- check performance and memory requirements
- click around a bit in the current branch and try things out to see if you like them

Thanks :)

Monviech:
Wooooo here we go :) thanks!

AdSchellevis:
@Monviech it might need some additional checks on the data, but when adding ~1000 rules divided over 4 interfaces, it works like a charm

Issue #8352, opened 2025-02-19
wizard: reimplement system setup and remove openvpn server one


fichtner:
**Important notices**

Before you add a new report, we ask you kindly to acknowledge the following:

- [x] I have read the contributing guide lines at https://github.com/opnsense/core/blob/master/CONTRIBUTING.md
- [x] I am convinced that my issue is new after having checked both open and closed issues at https://github.com/opnsense/core/issues?q=is%3Aissue

**Is your feature request related to a problem? Please describe.**

The wizard works on XML files which we have tolerated for OpenVPN server and initial setup, but never did more than fix bugs and avoid the code as much as possible.

**Describe the solution you like**

Remove the wizard.php and related glue from the project and build a lightweight form for the wizard steps.

**Describe alternatives you considered**

No. This needs to go.

**Additional context**

#8347

Issue #2751, opened 2018-09-23
system: dnsallowoverride isset() vs !empty() and shell script XML parsing fails


fichtner:
https://twitter.com/l33tname/status/1043783662410616832

fliiiix:
Unfortunately I don't have steps how to reproduce it.

fichtner:
@fliiiix no worries, I found two spots that don't seem to do the right thing. You can try the patch via:

    # opnsense-patch a07dca9

If you go to System: Settings: General and save once it should no longer happen :)

I just have to see if all readers of the setting do the right thing as well otherwise the write fix is pointless.

fichtner:
Surprisingly, the wizard does the right thing also so the patch will be shipped in 18.7.4 and that's it. :)

fliiiix:
Amazing :tada:

fichtner:
You know how they say... just a normal day for bicycle repair man.

https://www.youtube.com/watch?v=5Lwjv_VRM7A

;)

PR #8378, opened 2025-02-26
wizard: reimplement system setup

AdSchellevis:
closes https://github.com/opnsense/core/issues/8352

This commit implements our replacement for the setup wizard. The questions are roughly the same as in the legacy version. Some less relevant options have been removed (pppoe ondemand for example) and isc-dhcpd has been replaced with dnsmasq.

Only standard tools have been used, a memory model to validate the data and simple input forms in tabs.

The in memory model acts as a wrapper around a legacy configuration data and a couple of component models to apply the requested settings.

Some legacy settings using isset() have been altered to use their empty() equivalent.


src/etc/config.xml.sample
```diff
@@ -93,18 +93,19 @@
...
-      </range>
-    </lan>
-  </dhcpd>
   <unbound>
     <enable>1</enable>
   </unbound>
+  <dnsmasq>
+      <enable>1</enable>
+      <port>0</port>
+      <interface>lan</interface>
```
fichtner, line 102: if people don't assign a "lan" interface in console assignment this can be a problem. we also need to make sure migrations do the normalize dance then on this field and the one in dhcp_ranges ;(

src/etc/inc/system.inc
```diff
@@ -328,7 +328,7 @@ function get_searchdomains()
         $master_list[] = $syscfg['dnssearchdomain'];
     }
 
-    if (isset($syscfg['dnsallowoverride'])) {
+    if (!empty($syscfg['dnsallowoverride'])) {
```
fichtner, line 331: we did have issues with one of these before

src/opnsense/mvc/app/models/OPNsense/Core/ACL/ACL.xml
```diff
@@ -365,7 +365,8 @@
     <page-wizard-system>
         <name>System Setup Wizard</name>
```
fichtner, line 366: Menu says System: Wizard, I wonder if can bury it in System: Settings

src/etc/config.xml.sample
AdSchellevis: @fichtner but isn't this the same for isc at the moment?

https://github.com/opnsense/core/blob/3e299b29b9d0b1aaa17d1f950a6ef62676541c40/src/etc/config.xml.sample#L96-L104

In practice this likely means we need to replace the dhcpd sections in console.inc, which I haven't looked at yet (but should do indeed)

src/etc/inc/system.inc
AdSchellevis: @fichtner not sure what it was, but for simplicity, I prefer the `empty()` variants, so if it doesn't work as it should, we might have to figure out why that is.

src/opnsense/mvc/app/models/OPNsense/Core/ACL/ACL.xml
AdSchellevis: I don't have a strong preference, can change the acl description or the menu item.

src/etc/inc/system.inc
fichtner: this was https://github.com/opnsense/core/issues/2751 and looks like it was fixed in the way we want, but then we should move all spots to !empty, not just this one

src/etc/config.xml.sample
fichtner: dhcp is disabled when lan is not used, yes, this ties back to the console.inc madness

as far as isc doing it, also yes, but it's not an mvc migration-breaks-on-validation case

AdSchellevis: the console part should be handled implemented via https://github.com/opnsense/core/pull/8378/commits/70b1672e8dc64e52c7cbe1f9c3266006bc194a97 now.

src/etc/inc/system.inc
AdSchellevis: @fichtner let's change the rest of them as well indeed, https://github.com/opnsense/core/commit/588566fba8eb5eb170777e22a4b7fd51c5ea952a should do the trick.

src/etc/inc/xmlparse.inc
```diff
@@ -51,6 +51,7 @@ function listtags()
       'user',
       'vip', 'virtual_server', 'vlan',
       'winsserver', 'wolentry', 'widget',
+      'dhcp_ranges'
```
fichtner, line 54: would be nice to add this sorted

src/opnsense/mvc/app/models/OPNsense/Core/ACL/ACL.xml
fichtner: system: settings or system: configuration, decide after merge

src/www/system_general.php
```diff
@@ -200,7 +200,6 @@
             $config['system']['dnsallowoverride'] = true;
             $config['system']['dnsallowoverride_exclude'] = implode(',', $pconfig['dnsallowoverride_exclude']);
         } elseif (isset($config['system']['dnsallowoverride'])) {
```
fichtner, line 202:
```suggestion
        } else {
```

src/opnsense/service/conf/actions.d/actions_system.conf
```diff
@@ -148,6 +148,13 @@ type:script_output
 message:list locales
 cache_ttl:300
 
+[list.timezones]
+command:/usr/local/opnsense/scripts/system/get_timezones.php
+parameters:
+type:script_output
+message:list timezones
+
+
```
fichtner, line 157:
```suggestion

```

fichtner (approved):
minor comments

AdSchellevis:
@fichtner fixed in https://github.com/opnsense/core/pull/8378/commits/c1ef2a5e413c67926be0d9888ed40829047df38b , merging now

PR #8396, opened 2025-03-04
OpenVPN: Add support for various options

mhalden:
Add support for auth-gen-token renawal time.

Add support for auth-gen-token-secrets to allow failover between servers.

Add support for pushing inactive to clients to have them disconnect after being idle for a set time.

Add support for explicit-exit-notify and for pushing it. This will allow UDP connection to notify peers that they are going away.

Add support for ifconfig-pool-persist, which will allow smoother roaming combined with auth-gen-token since client will keep their ip address.

Add support from compress migrate. This option will conditionally use stub compression for clients announcing they have compression enabled while leaving it of for all other clients.

### DIFF
--- a/src/opnsense/mvc/app/controllers/OPNsense/OpenVPN/forms/dialogInstance.xml
+++ b/src/opnsense/mvc/app/controllers/OPNsense/OpenVPN/forms/dialogInstance.xml
@@ -433,6 +433,28 @@ Set to 0 to disable, remember to change your client as well.
         </grid_view>
     </field>
     <field>
+        <id>instance.auth-gen-token-renewal</id>
+        <label>Auth Token Renewal</label>
+        <type>text</type>
+        <advanced>true</advanced>
+        <style>role role_server</style>
+        <help>How often the auth token will be renewed, token expire after 2 * renewal time.</help>
+        <grid_view>
+            <visible>false</visible>
+        </grid_view>
+    </field>
+    <field>
+        <id>instance.auth-gen-token-secret</id>
+        <label>Auth Token secret</label>
+        <type>textbox</type>
+        <advanced>true</advanced>
+        <style>role role_server</style>
+        <help>Optional secret for use with auth-gen-token. This is useful to allow failover between multiple servers without user interaction.</help>
+        <grid_view>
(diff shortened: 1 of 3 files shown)


src/opnsense/mvc/app/controllers/OPNsense/OpenVPN/forms/dialogInstance.xml
```diff
@@ -432,6 +432,28 @@ Set to 0 to disable, remember to change your client as well.
...
+        <type>text</type>
+        <advanced>true</advanced>
+        <style>role role_server</style>
+        <help>How often the auth token will be renewed, token expire after 2 * renewal time.</help>
+        <grid_view>
+            <visible>false</visible>
+        </grid_view>
+    </field>
+    <field>
+        <id>instance.auth-gen-token-secret</id>
```
AdSchellevis, line 447: @mhalden If I understand the option correctly, these should be generated via `openvpn --genkey auth-token`, in which case it't probably practical if we add a button there (and maybe validate the input as well as this seems to be pretty strict).

AdSchellevis (commented):
@mhalden looks good, one question which might need a bit of work from my end, but further than that this should be smooth sailing. Thanks!

src/opnsense/mvc/app/controllers/OPNsense/OpenVPN/forms/dialogInstance.xml
mhalden: Yup, that's correct. It would definitely be useful with a button to generate the secret, but for our case we're fine without since it's basically set it and forget it.

AdSchellevis: @mhalden ok, thanks, let me fiddle a bit with this then and merge the batch.

AdSchellevis:
and added https://github.com/opnsense/core/commit/e7ad0a4fcb7afc4455acd99034d2dee0819388fe as these only seem to be relevant for servers as well.

mhalden:
Awesome, thanks :)

PR #8405, opened 2025-03-05
dnsmasq: When dhcp-fqdn is set, there must be a domain without an address set as default

Monviech:

`--dhcp-fqdn`

``To ensure that all names have a domain part, there must be at least --domain without an address specified when --dhcp-fqdn is set.``

Fixes the following error:

```
2025-03-05T15:44:42 | Critical | dnsmasq | FAILED to start up
2025-03-05T15:44:42 | Critical | dnsmasq | there must be a default domain when --dhcp-fqdn is set
```

### DIFF
--- a/src/opnsense/mvc/app/controllers/OPNsense/Dnsmasq/Api/SettingsController.php
+++ b/src/opnsense/mvc/app/controllers/OPNsense/Dnsmasq/Api/SettingsController.php
@@ -42,7 +42,7 @@ class SettingsController extends ApiMutableModelControllerBase
     public function getAction()
     {
         $data = parent::getAction();
-        $data[self::$internalModelName]['dhcp']['domain'] = (string)Config::getInstance()->object()->system->domain;
+        $data[self::$internalModelName]['dhcp']['this_domain'] = (string)Config::getInstance()->object()->system->domain;
 
         return $data;
     }
--- a/src/opnsense/mvc/app/controllers/OPNsense/Dnsmasq/Api/SettingsController.php
+++ b/src/opnsense/mvc/app/controllers/OPNsense/Dnsmasq/Api/SettingsController.php
@@ -29,11 +29,23 @@
 namespace OPNsense\Dnsmasq\Api;
 
 use OPNsense\Base\ApiMutableModelControllerBase;
+use OPNsense\Core\Config;
 
 class SettingsController extends ApiMutableModelControllerBase
 {
     protected static $internalModelName = 'dnsmasq';
     protected static $internalModelClass = '\OPNsense\Dnsmasq\Dnsmasq';
+
+    /**
+     * @inheritdoc
+     */
+    public function getAction()
+    {
+        $data = parent::getAction();
+        $data[self::$internalModelName]['dhcp']['domain'] = (string)Config::getInstance()->object()->system->domain;
+
+        return $data;
+    }
 
     /* hosts */
     public function searchHostAction()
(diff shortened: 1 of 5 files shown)


src/opnsense/service/templates/OPNsense/Dnsmasq/dnsmasq.conf
```diff
@@ -33,6 +33,10 @@ dhcp-lease-max={{dnsmasq.dhcp.lease_max}}
 dhcp-fqdn
 {% endif %}
 
+{% if dnsmasq.dhcp.domain %}
```
AdSchellevis, line 36: we could consider using the default one when not specified, but that's also something we can do later if needed. This is a good addition, thanks!

AdSchellevis (approved):

src/opnsense/service/templates/OPNsense/Dnsmasq/dnsmasq.conf
Monviech: I pushed a new commit with a sensible default.

https://en.wikipedia.org/wiki/.internal

AdSchellevis: @Monviech I was thinking more about the firewall's default domain to be honest, just need some trickery for the hint part in that case. kea has something similar for the hostname:

https://github.com/opnsense/core/blob/6b4c98726d583c8cb2c083030e67d9d9c8a61a67/src/opnsense/mvc/app/views/OPNsense/Kea/dhcpv4.volt#L30-L42

Monviech: Looks cool I try that out. :)

src/opnsense/mvc/app/controllers/OPNsense/Dnsmasq/Api/SettingsController.php
```diff
@@ -29,12 +29,24 @@
...
     protected static $internalModelName = 'dnsmasq';
     protected static $internalModelClass = '\OPNsense\Dnsmasq\Dnsmasq';
 
+    /**
+     * @inheritdoc
+     */
+    public function getAction()
+    {
+        $data = parent::getAction();
+        $data[self::$internalModelName]['dhcp']['domain'] = (string)Config::getInstance()->object()->system->domain;
```
AdSchellevis, line 45: you need to place this into a separate container to prevent the value from being persisted.

AdSchellevis (changes requested):

src/opnsense/mvc/app/controllers/OPNsense/Dnsmasq/Api/SettingsController.php
Monviech: I misunderstood the concept. I have fixed it now.

src/opnsense/service/templates/OPNsense/Dnsmasq/dnsmasq.conf
```diff
@@ -31,7 +31,7 @@ dhcp-lease-max={{dnsmasq.dhcp.lease_max}}
 
 {% if dnsmasq.dhcp.fqdn == '1' %}
 dhcp-fqdn
-domain={{dnsmasq.dhcp.domain|default('internal')}}
+domain={{dnsmasq.dhcp.domain}}
```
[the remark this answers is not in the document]
Monviech: I've used ``system.domain`` instead

AdSchellevis (approved):
Nice!

PR #8407, opened 2025-03-06
reporting: granularity default item label tweak

NOYB:
Granularity default selector item consistency with other items.


fichtner:
Merged, thanks.

PR #8408, opened 2025-03-06
dnsmasq: Use interface directly as tag in dhcp options

Monviech:
This one makes simple setups simpler since we do not need to create any tags manually if we want to match custom DHCP options with a DHCP range.

### DIFF
--- a/src/opnsense/service/templates/OPNsense/Dnsmasq/dnsmasq.conf
+++ b/src/opnsense/service/templates/OPNsense/Dnsmasq/dnsmasq.conf
@@ -115,7 +115,7 @@ conf-dir=/usr/local/etc/dnsmasq.conf.d,\*.conf
 {% if dhcp_range.start_addr and ':' not in dhcp_range.start_addr %}
 {# IPv4 range #}
 dhcp-range={%if dhcp_range.interface -%}
-tag:{{helpers.physical_interface(dhcp_range.interface)}},
+{% if dhcp_range.interface and dhcp_range.set_tag -%}tag:{%- endif -%}{{helpers.physical_interface(dhcp_range.interface)}},
 {%- endif -%}
 {%- if dhcp_range.set_tag -%}
 set:{{dhcp_range.set_tag|replace('-','')}},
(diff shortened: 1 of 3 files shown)


src/opnsense/service/templates/OPNsense/Dnsmasq/dnsmasq.conf
```diff
@@ -115,7 +115,7 @@ conf-dir=/usr/local/etc/dnsmasq.conf.d,\*.conf
 {% if dhcp_range.start_addr and ':' not in dhcp_range.start_addr %}
 {# IPv4 range #}
 dhcp-range={%if dhcp_range.interface -%}
-tag:{{helpers.physical_interface(dhcp_range.interface)}},
+{% if dhcp_range.interface and dhcp_range.set_tag -%}tag:{%- endif -%}{{helpers.physical_interface(dhcp_range.interface)}},
```
AdSchellevis, line 118: why is this? I was under the impression that we can always prepend `tag:` for interfaces as all requests are tagged by interface name, but maybe I'm missing something.

src/opnsense/service/templates/OPNsense/Dnsmasq/dnsmasq.conf
```diff
@@ -131,7 +131,7 @@ set:{{dhcp_range.set_tag|replace('-','')}},
 {% else %}
 {# IPv6 range #}
 dhcp-range={%if dhcp_range.interface -%}
-{{helpers.physical_interface(dhcp_range.interface)}},
+{% if dhcp_range.interface and dhcp_range.set_tag -%}tag:{%- endif -%}{{helpers.physical_interface(dhcp_range.interface)}},
```
AdSchellevis, line 134: same

AdSchellevis (commented):
some questions, but rest looks good!

src/opnsense/service/templates/OPNsense/Dnsmasq/dnsmasq.conf
AdSchellevis: let's prefix the `tag:` as this is more consistent. assuming the interface always receives a tag (which is also what I could find), setting it again shouldn't make much sense anyway.

Issue #8399, opened 2025-03-04
Add the ability to reorder domain overrides in dnsmasq


airforceixi:
**Important notices**

Before you add a new report, we ask you kindly to acknowledge the following:

- [X] I have read the contributing guide lines at https://github.com/opnsense/core/blob/master/CONTRIBUTING.md
- [X] I am convinced that my issue is new after having checked both open and closed issues at https://github.com/opnsense/core/issues?q=is%3Aissue

**Is your feature request related to a problem? Please describe.**

Overrides in /ui/dnsmasq in the Domain tab cannot be reordered. This is problematic when strict-order ("Query DNS servers sequentially") is enabled and there are several overrides in place. Adding an override in a specific spot requires you to remove and re-add overrides.

**Describe the solution you like**

Similar to firewall rules, add the ability to re-order domain overrides in /ui/dnsmasq.

**Describe alternatives you considered**

Skipping the UI and adding the overrides in /usr/local/etc/dnsmasq.conf.d/ is a workaround to provide the ability to reorder them, as needed, but isn't ideal.

**Additional context**

Using strict-order and multiple entries per domain I am able to define the order in which forwarders are queried, based on physical location. Example: 1x AD local and 1x AD remote. Both AD servers provide DNS for AD. AD domains are defined in dnsmasq to forward queries to the AD servers, local being first. In the event an additional domain is added, or another AD server is added, adjusting the order requires a partial rebuild of the domain override list.

PR #8411, opened 2025-03-06
dnsmasq: Allow domain overrides to be optionally sorted by sequence number to support strict-order

Monviech:
Fixes: https://github.com/opnsense/core/issues/8399

### DIFF
--- a/src/opnsense/service/templates/OPNsense/Dnsmasq/dnsmasq.conf
+++ b/src/opnsense/service/templates/OPNsense/Dnsmasq/dnsmasq.conf
@@ -59,8 +59,14 @@ bind-interfaces
 bogus-priv
 {% endif %}
 
-{% for override in helpers.toList('dnsmasq.domainoverrides') %}
-server=/{{override.domain}}/{{override.ip}}{%if override.port %}#{{override.port}}{% endif %}{%if override.srcip %}@{{override.srcip}}{% endif %}
+{# Domain Overrides can be sorted by sequence to support strict order #}
+{% set unsorted_domainoverrides = helpers.toList('dnsmasq.domainoverrides') %}
+{% for item in unsorted_domainoverrides %}
+{%     set _ = item.update({'sequence': item.get('sequence', '0') | int}) %}
+{% endfor %}
+{% set sorted_domainoverrides = unsorted_domainoverrides | sort(attribute='sequence') %}
+{% for override in sorted_domainoverrides %}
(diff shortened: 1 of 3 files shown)


src/opnsense/service/templates/OPNsense/Dnsmasq/dnsmasq.conf
```diff
@@ -59,8 +59,14 @@ bind-interfaces
 bogus-priv
 {% endif %}
 
-{% for override in helpers.toList('dnsmasq.domainoverrides') %}
-server=/{{override.domain}}/{{override.ip}}{%if override.port %}#{{override.port}}{% endif %}{%if override.srcip %}@{{override.srcip}}{% endif %}
+{# Domain Overrides can be sorted by sequence to support strict order #}
+{% set unsorted_domainoverrides = helpers.toList('dnsmasq.domainoverrides') %}
+{% for item in unsorted_domainoverrides %}
```
AdSchellevis, line 64:
```suggestion
{% for override in helpers.toList('dnsmasq.domainoverrides', 'sequence') %}
```

is likely all this needs :)

https://github.com/opnsense/core/blob/b7ed45db20d67d1e93fa67d373894bcb6202dd4e/src/opnsense/service/modules/addons/template_helpers.py#L99

gspannu:
@Monviech 
You may (or may not) be aware that the current implementation of dnsmasq in 25.1.1/2 has a ‘possible’ bug that the strict-order functionality doe snot work as expected.

OPNsense respects the strict-order and forwards the dns request to the 1st listed server (rather than all in parallel); however, if the 1st server does not respond, dnsmasq does not forward the request to the next server; and instead waits for a long time and then the query times out.

This was not an issue in 24.x (although I cant pinpoint in which release this issue got introduced).

I have a custom dnsmasq.conf with the following entries (and the GUI servers are blank) and this issue is exhibited.

```
server=192.168.1.1#53035
server=1.1.1.1
add-mac
add-subnet=32,128

```

I have BlockyDNS running on port 53035.

_Behaviour seen:_
If I enable the strict-order option in the GUI, my requests are only forwarded to 192.168.1.1:53035 and all works, but if I disable my ‘Blocky DNS Instance’ running on port 53035, then the query just times out after a long time, rather than being forwarded to 1.1.1.1 as expected.

**In addition, may i request that these features be added to the dnsmasq GUI**
- GUI option to specify a port number along with the dns server address (default 53 if unspecified)
- GUI options for add-mac
- GUI options for add-subnet

Issue #8273, opened 2025-01-31
unbound: improve whitelist handling


piracyquest:
**Important notices**

Before you add a new report, we ask you kindly to acknowledge the following:

- [x] I have read the contributing guidelines at https://github.com/opnsense/core/blob/master/CONTRIBUTING.md
- [x] I am convinced that my issue is new after having checked both open and closed issues at https://github.com/opnsense/core/issues?q=is%3Aissue

**Describe the bug**

Allowlists (whitelists) fail to bypass blocklists in DNS filtering. Explicitly allowed domains remain blocked if they match an active blocklist entry.

Example: Whitelisting `0.beer` (or any other) while using Hagezi Ultimate blocklist still prevents domain resolution (nslookup fails) unless the blocklist is disabled.

**To Reproduce**

1. Go to `Services > Unbound > Blocklist`
2. Enable "Hagezi Multi Ultimate" blocklist
3. Go to `Whitelist`
4. Add domain `0.beer` (or any other) and save (use regex or regular domains)
5. Apply changes and restart the service. Wait all you want :)
6. Execute from client: `nslookup 0.beer` or `dig 0.beer`
7. Observe DNS resolution failure

**Expected behavior**

Allowlisted domains should override blocklist entries, permitting resolution of specified domains.

**Actual behavior**

Allowlisted domains remain blocked when present in active blocklist.

**Workarounds tested**

- Success: Disabling entire blocklist (confirming whitelist works without blocklist)
- Failed: Whitelist entry alone while blocklist is active

**Relevant log files**

- Unbound logs register a red entry, signaling it's blocking the domain even though the unblock button turns into a "block" button (giving me the option to block the domain, meaning it should be already unblocked)

**Environment**

Software version:
OPNsense 25.x (amd64)
Hagezi Blocklist Ultimate

**Additional context**

- Verified with Unbound resolver
- Reproducible after service restarts
- Test domain `0.beer` confirmed present in Hagezi blocklist

JasMan78:
I can confirm that whitelisting e.g. 00fun.com does not work in 24.7.12_4 and 25.1.
Whitelisting other domains that does not start with a zero work in both versions.

piracyquest:
Not in my case: no domain can be whitelisted.

samham77:
try .*beer

JasMan78:
> Not in my case: no domain can be whitelisted.

Most times I need to "Apply" and restart Unbound to whitelist a domain. Have you tried that?
[...]

PR #8415, opened 2025-03-07
Make exceptions to unbound blocklists more intuitive

oyxnaut:
This changes the way the exceptions for unbound DNS blocking work, in order to make it more intuitive. (see #8273)
Currently, this only writes all exceptions into a new list element in `dnsbl.json`. This only works for non-regex domains. It could be changed to use `re.findall`, but I suspect that will be slow.

### DIFF
--- a/src/opnsense/scripts/unbound/blocklists/__init__.py
+++ b/src/opnsense/scripts/unbound/blocklists/__init__.py
@@ -67,6 +67,12 @@ class BaseBlocklistHandler:
         """
         pass
 
+    def get_excludelist(self):
+        """
+        Overridden by derived classes to produce a formatted exludelist. Returns a set of domains
+        """
+        return set()
+
     def _blocklist_reader(self, uri):
         """
         Used by a derived class to define a caching and/or download routine.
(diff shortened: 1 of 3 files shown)


AdSchellevis:
@oyxnaut at a first glance I expect we should move the whole whitelist handling to the unbound module which would also ease different matching strategies. I'll take a look when I have some time, thanks for starting this PR.

AdSchellevis:
@oyxnaut can you try https://github.com/opnsense/core/commit/aa2cff3e665fd0fb2b22b69be4334d365ee066eb ? 

To install, use:
```
opnsense-patch aa2cff3
```

Make sure to fetch the list again and stop/start unbound

BestDistress:
> @oyxnaut can you try [aa2cff3](https://github.com/opnsense/core/commit/aa2cff3e665fd0fb2b22b69be4334d365ee066eb) ?
> 
> To install, use:
> 
> ```
> opnsense-patch aa2cff3
> ```
> 
> Make sure to fetch the list again and stop/start unbound

@AdSchellevis this resolved it for me, although it didn't seem to show up in reporting as a successful query when on the white list.

VartanS:
Hot to initiate list update manually ?

AdSchellevis:
@BestDistress maybe it wasn't processed yet, it should log the same thing as before when the item was just not in the list (it works on my end at least).

oyxnaut:
> @oyxnaut can you try [aa2cff3](https://github.com/opnsense/core/commit/aa2cff3e665fd0fb2b22b69be4334d365ee066eb) ?

This works. Thank you.

Issue #7324, opened 2024-03-13
Add interface assignments to API


djryanj:
**Important notices**

Before you add a new report, we ask you kindly to acknowledge the following:

- [X] I have read the contributing guide lines at https://github.com/opnsense/core/blob/master/CONTRIBUTING.md
- [X] I am convinced that my issue is new after having checked both open and closed issues at https://github.com/opnsense/core/issues?q=is%3Aissue

**Is your feature request related to a problem? Please describe.**

Currently, the interface assignments aren't (or don't seem to be) accessible via the API to either get or set. This makes complete programmatic bootstrapping (e.g., IaC) of a device difficult.

Even the lack of being able to identify which interface assignment (e.g., "opt1") is assigned to (e.g.,) a VLAN makes further programmatic configuration difficult, such as trying to build firewall rules as the interface (not the device or description as used in the GUI) is required to do that.

**Describe the solution you like**

Implement a complete API around interface assignments.

**Describe alternatives you considered**

- Do this manually
- Do it using fancy bespoke code via the web interface

fichtner:
It’s on the work queue but it will not be done this year having the roadmap already up for 24.7 and the next one being 25.1.

Cheers,
Franco

djryanj:
@fichtner - thanks for your reply.

I assume a PR would be accepted even if it wasn't in line with the planned/roadmapped date?

fichtner:
Likely not for a number of reasons. I think a PR was proposed already and the impact was discussed there.

djryanj:
Thanks very much.

fichtner:
For reference that was https://github.com/opnsense/core/pull/6765

swhite2:
> Even the lack of being able to identify which interface assignment (e.g., "opt1") is assigned to (e.g.,) a VLAN makes further programmatic configuration difficult, such as trying to build firewall rules as the interface (not the device or description as used in the GUI) is required to do that.

FWIW, we do have `/api/interfaces/overview/interfacesInfo` nowadays, as used by Interfaces -> Overview.

djryanj:
@swhite2  - Yes, and I believe that the Terraform module I'm leveraging (https://registry.terraform.io/providers/browningluke/opnsense/latest/docs) uses that.

The issue is that that endpoint doesn't return interface assignment details (e.g., "opt1"). If it could be extended to include that simply enough, that would be a huge help.

swhite2:
@djryanj 

[...]

PR #8436, opened 2025-03-14
Add interface assignments to API

black3x3:
Although the [issue](https://github.com/opnsense/core/issues/7324) is closed, I have worked on it as I also need this functionality. Additionally, I have opened two pull requests in the following repositories:
	- https://github.com/browningluke/terraform-provider-opnsense 
	- https://github.com/browningluke/opnsense-go repositories

I would appreciate a review and would be happy to improve the quality of the changes.


AdSchellevis:
Unfortunately this is unlikely to mature as it will need a lot of time from our end to bring this to the finish line. In order to include an api on a component on our end, the ui parts should also be finished. I'll keep the ticket open for when we start planning the work on our end.

vpereira:
> Unfortunately this is unlikely to mature as it will need a lot of time from our end to bring this to the finish line. In order to include an api on a component on our end, the ui parts should also be finished. I'll keep the ticket open for when we start planning the work on our end.

Hello, thank you for your answer!

> In order to include an api on a component on our end, the ui parts should also be finished

Could you please take a look in this PR? The "Assignment" UI is already in place, so i fail to see which UI has to be written here. I can however extend the UI too, if needed

>  I'll keep the ticket open for when we start planning the work on our end.

It is a working PR, which together with the other two related PRs (see the description) that allow users to access it programmatically via terraform 


@fichtner moooin if we cannot get this code into the mainline, would be a possible/good idea to write it as plugin? I am wondering if it doesn't get merged, how to keep my opnsense up-to-date and still apply my custom code ❓

AdSchellevis:
> .. Could you please take a look in this PR? The "Assignment" UI is already in place,...

It's not as there's no user interface using the exact same endpoints (which is a requirement).

> moooin if we cannot get this code into the mainline, would be a possible/good idea to write it as plugin?

best keep the plugin local at your end, no need to push it to our repository.


Eventually we will refactor the assignments as well, it's just not yet on the list.

vpereira:
> @vpereira hi there long time no see :)

Hey @fichtner yeah, too long actually :)

> What's your current use case?

We are in the pursuit of automating stuff with Terraform, and with this PR in combination with https://github.com/browningluke/terraform-provider-opnsense/pull/60 is a huge leap forward. Our goal is to handle OPNsense change requests in a GitOps style, so we can deploy them via CI/CD and all that jazz. It's what the cool kids do these days 😄

szymczag:
I'll just leave it here: https://github.com/opnsense/core/issues/7324#issuecomment-2830694222

M4jx:
Would love to see that implemented. I’m currently developing a plugin that manages WireGuard instances and peers, but I’m currently unable to continue the automation because the interface assignment functionality lacks an API :(

fichtner:
Implemented since 26.7. Bringing all the interface settings to the page will take some time, but work has already been started.

vpereira:
> Implemented since 26.7. Bringing all the interface settings to the page will take some time, but work has already been started.

Thank you @fichtner and cheers 🍻

fichtner:
@vpereira no problem, hope you're doing good out there my old friend! :)

PR #8441, opened 2025-03-15
Suricata: Add JA4 support

0xThiebaut:
Adds Suricata's JA4 option.

### DIFF
--- a/src/opnsense/service/templates/OPNsense/IDS/suricata.yaml
+++ b/src/opnsense/service/templates/OPNsense/IDS/suricata.yaml
@@ -254,6 +254,9 @@ outputs:
             # output TLS transaction where the session is resumed using a
             # session id
             session-resumption: {{ 'no' if helpers.empty('OPNsense.IDS.general.eveLog.tls.sessionResumption') else 'yes' }}
+            # ja4 hashes in tls records will never be logged unless
+            # the following is set to on. (Default off)
+            ja4: {{ 'on' if not helpers.empty('OPNsense.IDS.general.eveLog.tls.extended') or 'ja4' in OPNsense.IDS.general.eveLog.tls.custom.split(',') else 'off' }}
             # custom controls which TLS fields that are included in eve-log
 {%          if not helpers.empty('OPNsense.IDS.general.eveLog.tls.custom') %}
             custom: [{{ OPNsense.IDS.general.eveLog.tls.custom }}]
@@ -923,9 +926,10 @@ app-layer:
       detection-ports:
         dp: 443
 
-      # Generate JA3 fingerprint from client hello. If not specified it
+      # Generate JA3/JA4 fingerprints from client hello. If not specified it
       # will be disabled by default, but enabled if rules require it.
       ja3-fingerprints: yes
+      ja4-fingerprints: yes
 
       # What to do when the encrypted communications start:
       # - default: keep tracking TLS session, check for protocol anomalies,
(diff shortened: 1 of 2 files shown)


src/opnsense/service/templates/OPNsense/IDS/suricata.yaml
```diff
@@ -254,6 +254,9 @@ outputs:
             # output TLS transaction where the session is resumed using a
             # session id
             session-resumption: {{ 'no' if helpers.empty('OPNsense.IDS.general.eveLog.tls.sessionResumption') else 'yes' }}
+            # ja4 hashes in tls records will never be logged unless
+            # the following is set to on. (Default off)
+            ja4: {{ 'on' if helpers.empty('OPNsense.IDS.general.eveLog.tls.custom') or 'ja4' in OPNsense.IDS.general.eveLog.tls.custom else 'off' }}
```
AdSchellevis, line 259: this defaults to `on` (empty custom property)

src/opnsense/service/templates/OPNsense/IDS/suricata.yaml
```diff
@@ -923,9 +926,10 @@ app-layer:
       detection-ports:
         dp: 443
 
-      # Generate JA3 fingerprint from client hello. If not specified it
+      # Generate JA3/JA4 fingerprints from client hello. If not specified it
       # will be disabled by default, but enabled if rules require it.
       ja3-fingerprints: yes
+      ja4-fingerprints: yes
```
AdSchellevis, line 932: question is if we actually need to enable this, unless no rules actually use the option in your use-case. There might be a performance penalty, so just wanted to double check.

AdSchellevis (changes requested):

src/opnsense/service/templates/OPNsense/IDS/suricata.yaml
0xThiebaut: Updated to be enabled if TLS extended logging is enabled or JA4 is explicitly listed in the custom fields given JA4 is considered a default field part of the TLS extended logging: https://github.com/OISF/suricata/blob/66e47a1983bee06fb85418d8ca053d0d9a372474/src/output-json-tls.c#L101-L114

src/opnsense/service/templates/OPNsense/IDS/suricata.yaml
0xThiebaut: I've enabled it given it behaves like JA3. From the [documentation](https://docs.suricata.io/en/latest/output/eve/eve-json-format.html#id10):
> JA3 and JA4 must be enabled in the Suricata config file (set 'app-layer.protocols.tls.ja3-fingerprints'/'app-layer.protocols.tls.ja4-fingerprints' to 'yes').

Technically, we could for each JA3/JA4:
- Set it to `yes` if either extended TLS logging is enabled or if the field is explicitly mentioned in the custom TLS fields.
- Not specify it otherwise, letting Suricata determine whether it is required if a rule needs it.

I however did not take this approach as it is technically a "breaking" change for JA3 (i.e., changing the default).

src/opnsense/service/templates/OPNsense/IDS/suricata.yaml
AdSchellevis: @0xThiebaut ok, next question, where is `ja4` specified in suricata for tls? I couldn't find it in their reference configuration or their docs myself.

https://github.com/OISF/suricata/blob/66e47a1983bee06fb85418d8ca053d0d9a372474/suricata.yaml.in#L281-L288

https://docs.suricata.io/en/latest/output/eve/eve-json-output.html#tls

AdSchellevis: @0xThiebaut ok, that's confusing, latest isn't "latest" apparently in the documentation, thanks for clearing that up :)

AdSchellevis:
@0xThiebaut merged it now, thanks!

PR #8442, opened 2025-03-15
Suricata: Add support for all available EVE types

0xThiebaut:
Add support for all available Suricata EVE types.

This PR avoids migrations for existing types (e.g., `http`, `tls`, `alert`, `ssh`, ...) and hence does not offer the best UX. A migration will be proposed at a later stage.

### DIFF
--- a/src/opnsense/service/templates/OPNsense/IDS/suricata.yaml
+++ b/src/opnsense/service/templates/OPNsense/IDS/suricata.yaml
@@ -220,34 +220,6 @@ outputs:
             dump-all-headers: {{OPNsense.IDS.general.eveLog.http.dumpAllHeaders}}
 {%          endif %}
 {%       endif %}
-#        - dns:
-            # This configuration uses the new DNS logging format,
-            # the old configuration is still available:
-            # https://docs.suricata.io/en/latest/output/eve/eve-json-output.html#dns-v1-format
-
-            # As of Suricata 5.0, version 2 of the eve dns output
-            # format is the default.
-            #version: 2
-
-            # Enable/disable this logger. Default: enabled.
-            #enabled: yes
-
-            # Control logging of requests and responses:
-            # - requests: enable logging of DNS queries
-            # - responses: enable logging of DNS answers
-            # By default both requests and responses are logged.
-            #requests: no
-            #responses: no
-
-            # Format of answer logging:
-            # - detailed: array item per answer
-            # - grouped: answers aggregated by type
-            # Default: all
-            #formats: [detailed, grouped]
-
-            # DNS record types to log, based on the query type.
-            # Default: all.
-            #types: [a, aaaa, cname, mx, ns, ptr, txt]
 {%      if not helpers.empty('OPNsense.IDS.general.eveLog.tls.enable') %}
         - tls:
             extended: {{ 'no' if helpers.empty('OPNsense.IDS.general.eveLog.tls.extended') else 'yes' }}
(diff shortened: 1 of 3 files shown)


src/opnsense/service/templates/OPNsense/IDS/suricata.yaml
```diff
@@ -230,7 +230,7 @@ outputs:
             #version: 2
 
             # Enable/disable this logger. Default: enabled.
-            #enabled: yes
+            enabled: {{ 'yes' if 'dns' in OPNsense.IDS.general.eveLog.types.enabled else 'no' }}
```
AdSchellevis, line 233: you probably want to store the types before usage as `OPNsense.IDS.general.eveLog.types.enabled` is actually a string, in which case two overlapping options will have side affects (e.g. `'yes' in 'x,b,yesy,k'`)

src/opnsense/service/templates/OPNsense/IDS/suricata.yaml
```diff
@@ -282,43 +285,48 @@ outputs:
...
-        #- dcerpc
-        #- krb5
-        #- bittorrent-dht
-        #- snmp
-        #- rfb
-        #- sip
-        #- quic
-        #- dhcp:
-            #enabled: yes
+        {{ '#' if 'dnp3' not in OPNsense.IDS.general.eveLog.types.enabled else '' }}- dnp3
```
AdSchellevis, line 290: if we only assign valid options in `OPNsense.IDS.general.eveLog.types.enabled` we can iterate over them here:

```
{% for opt in OPNsense.IDS.general.eveLog.types.enabled.split(',') %}
    - {{opt}}
{% endfor %}
```

Use an if statement inside the blob to decorate the option when set.

AdSchellevis (changes requested):

src/opnsense/service/templates/OPNsense/IDS/suricata.yaml
```diff
@@ -282,43 +260,18 @@ outputs:
...
-            #totals: yes       # stats for all threads merged together
-            #threads: no       # per thread stats
-            #deltas: no        # include delta values
-        # bi-directional flows
-        #- flow
-        # uni-directional flows
-        #- netflow
+            extended: yes
+{%          elif opt == 'mqtt' and opt in eveLog_types_extended %}
+            passwords: yes           # enable output of passwords
```
AdSchellevis, line 270: this feels like a possible security concern...

src/opnsense/service/templates/OPNsense/IDS/suricata.yaml
```diff
@@ -942,7 +895,7 @@ app-layer:
       #encryption-handling: default
 
     pgsql:
-      enabled: no
+      enabled: {{ 'yes' if 'pgsql' in eveLog_types_enabled else 'no' }}
```
AdSchellevis, line 898: shouldn't this also move into the loop? I expect it duplicates the item now, or am I mistaken?

src/opnsense/service/templates/OPNsense/IDS/suricata.yaml
```diff
@@ -998,14 +951,14 @@ app-layer:
...
-        body-md5: no
+        body-md5: {{ 'yes' if 'smtp' in eveLog_types_extended else 'no' }}
       # Configure inspected-tracker for file_data keyword
       inspected-tracker:
         content-limit: 100000
         content-inspect-min-size: 32768
         content-inspect-window: 4096
     imap:
-      enabled: detection-only
+      enabled: {{ 'yes' if 'imap' in eveLog_types_enabled else 'detection-only' }}
```
AdSchellevis, line 961: same here

src/opnsense/service/templates/OPNsense/IDS/suricata.yaml
```diff
@@ -1171,7 +1124,7 @@ app-layer:
 
     # DNP3
     dnp3:
-      enabled: no
+      enabled: {{ 'yes' if 'dnp3' in eveLog_types_enabled else 'no' }}
```
AdSchellevis, line 1127: same here

src/opnsense/service/templates/OPNsense/IDS/suricata.yaml
```diff
@@ -1192,7 +1145,7 @@ app-layer:
       enabled: yes
 
     sip:
-      #enabled: yes
+      enabled: {{ 'yes' if 'sip' in eveLog_types_enabled else 'no' }}
```
AdSchellevis, line 1148: same here

AdSchellevis (changes requested):

src/opnsense/service/templates/OPNsense/IDS/suricata.yaml
0xThiebaut: I agree it is a security concern but due to the fact it transits in plaintext over the wire, not due to its logging. A similar already existing risk is the logging of HTTP headers which might contain bearer tokens and/or cookies.

If you nonetheless object to this specific logging, I can remove it from the fields supporting extended logging.

src/opnsense/service/templates/OPNsense/IDS/suricata.yaml
AdSchellevis: @0xThiebaut ok, thanks, I'll need to try this on my end before merging. these yaml files can be sensitive in my experience.

src/opnsense/service/templates/OPNsense/IDS/suricata.yaml
AdSchellevis: @0xThiebaut that's true, let me think about this, adding an extra warning in the help text might be enough and more generic anyway.

src/opnsense/service/templates/OPNsense/IDS/suricata.yaml
0xThiebaut: Thanks, I really appreciate the time and care you're taking for community contributions!

AdSchellevis: @0xThiebaut you're welcome, we can't pull-in everything, but do make an effort when possible.

0xThiebaut:
Closing as stale.

Issue #8440, opened 2025-03-14
[dnsmasq] Make additional dnsmasq options available in GUI


gspannu:
**Important notices**

Before you add a new report, we ask you kindly to acknowledge the following:

- [x] I have read the contributing guide lines at https://github.com/opnsense/core/blob/master/CONTRIBUTING.md
- [x] I am convinced that my issue is new after having checked both open and closed issues at https://github.com/opnsense/core/issues?q=is%3Aissue

**Is your feature request related to a problem? Please describe.**
Earlier issue reported here: https://github.com/opnsense/core/issues/8416

**Describe the solution you like**

Make the following **2 dnsmasq options** available in the GUI under OPNsense > dnsmasq 

These are options available in dnsmasq and explained in manpages [here](https://dnsmasq.org/docs/dnsmasq-man.html).

1. `--add-mac=` 
From dnsmasq manpage
_--add-mac[=base64|text]
Add the MAC address of the requestor to DNS queries which are forwarded upstream. This may be used to DNS filtering by the upstream server._

2. `--add-subnet`
From dnsmasq manpage
_--add-subnet[[=[<IPv4 address>/]<IPv4 prefix length>][,[<IPv6 address>/]<IPv6 prefix length>]]
Add a subnet address to the DNS queries which are forwarded upstream. If an address is specified in the flag, it will be used, otherwise, the address of the requestor will be used. The amount of the address forwarded depends on the prefix length parameter: 32 (128 for IPv6) forwards the whole address, zero forwards none of it but still marks the request so that no upstream nameserver will add client address information either. The default is zero for both IPv4 and IPv6. For example, —add-subnet=32,128 will add the /32 and /128 subnets of the request or for IPv4 and IPv6, --add-subnet=24,96 will add the /24 and /96 subnets of the requestor for IPv4 and IPv6 requestors, respectively. --add-subnet=1.2.3.4/24 will add 1.2.3.0/24 for IPv4 requestors and ::/0 for IPv6 requestors. --add-subnet=1.2.3.4/24,1.2.3.4/24 will add 1.2.3.0/24 for both IPv4 and IPv6 requestors._

**Why these features are required?**

- OPNsense users who run an external Adblocker (e.g. PiHole, etc.) require the **mac-address/ip address of clients** requesting dns queries to better display statistics and manage adblocking.
- Without these directives, all queries forwarded to external servers appear as coming from OPNsense itself (rather than the actual client ip/mac-address)

**Way to implement**
[...]

PR #8447, opened 2025-03-17
dnsmasq: Add add-mac, add-subnet and strip-subnet options

Monviech:
Fixes: https://github.com/opnsense/core/issues/8440

The "add-subnet" option would be pretty hard to validate without a custom field type, or splitting the field up into like 2 integer and 2 Network fields (4 fields). Thus only the most probable use-case of adding the IPv4 and IPv6 client addresses was added. Even that would be hard to split up, as the documentation is a little fuzzy about only setting IPv6 for example.

This here is the simplest fix without spending too much time.


AdSchellevis (approved):
@Monviech I doubted a bit about the `add-subnet` part, but also don't expect a lot of other valid uses-cases than sending the client ip.

PR #8448, opened 2025-03-17
dnsmasq: Add option6 to support DHCPv6 options

Monviech:
From man page:

`IPv6 options are specified using the option6: keyword, followed by the option number or option name. The IPv6 option name space is disjoint from the IPv4 option name space. IPv6 addresses in options must be bracketed with square brackets, eg. --dhcp-option=option6:ntp-server,[1234::56] For IPv6, [::] means "the global address of the machine running dnsmasq", whilst [fd00::] is replaced with the ULA, if it exists, and [fe80::] with the link-local address.`


AdSchellevis (approved):
@Monviech makes sense, thanks!

PR #8449, opened 2025-03-17
dnsmasq: Add router advertisement options

Monviech:
Fixes: https://github.com/opnsense/core/issues/8455


Monviech:
This PR has matured and works as expected in my environment. I'm using it as my primary RA daemon right now.

@AdSchellevis @fichtner it's ready to discuss if this is worth being added.

src/opnsense/mvc/app/models/OPNsense/Dnsmasq/Dnsmasq.xml
```diff
@@ -180,19 +181,34 @@
                 <IpAllowed>N</IpAllowed>
             </domain>
             <nosync type="BooleanField"/>
+            <ra_mode type="OptionField">
+                <OptionValues>
+                    <ra-only>ra-only</ra-only>
```
fichtner, line 186: Human readable labels would be nicer than duplicating the config file switches

fichtner (approved):
I'm going to start unwinding the radvd code now, but it will force further work here down the line. The autmatic setup isn't great when we have multiple things doing the same but from a user perspective it's much better than managing this by hand so I sort of expect we need to roll with usable defaults including DHCPv6 in Dnsmasq too.

src/opnsense/mvc/app/models/OPNsense/Dnsmasq/Dnsmasq.xml
Monviech: I decides this so they can be searched for in the dnsmasq man page. They all have pretty distinct features.

I thought about preparing premade options like in radvd that combine settings like these, but who knows which new switches will come in the future.

What would you suggest here?

fichtner: Not sure, the documentation is sparse on this as well. To be frank what bothers me most is why we cannot write <ra-only/> etc ;)

Monviech: I'll leave it like it is for now then.

Monviech:
@fichtner Can we merge this to master or still wait?

src/opnsense/mvc/app/controllers/OPNsense/Dnsmasq/forms/dialogDHCPrange.xml
```diff
@@ -15,7 +15,7 @@
         <id>range.start_addr</id>
         <label>Start address</label>
         <type>text</type>
-        <help>Start of the range, e.g. 192.168.1.100, 2000::1 or when constructor is used a partial like ::1.</help>
+        <help>Start of the range, e.g. 192.168.1.100 for DHCPv4, 2000::1 for DHCPv6 or when a constructor is used a partial like ::1. To reveal IPv6 related options, enter a IPv6 address. When using router advertisements, it is possible to use a constructor with :: as the start address and no end address.</help>
```
fichtner, line 18:
```suggestion
        <help>Start of the range, e.g. 192.168.1.100 for DHCPv4, 2000::1 for DHCPv6 or when a constructor is using a suffix like ::1. To reveal IPv6 related options, enter a IPv6 address. When using router advertisements, it is possible to use a constructor with :: as the start address and no end address.</help>
```

src/opnsense/mvc/app/controllers/OPNsense/Dnsmasq/forms/dialogDHCPrange.xml
```diff
@@ -27,19 +27,81 @@
         <id>range.constructor</id>
         <label>Constructor</label>
         <type>dropdown</type>
+        <style>selectpicker style_dhcpv6</style>
         <help>Interface to use to calculate the proper range, when selected, a range maybe specified as partial (e.g. ::1, ::400)</help>
```
fichtner, line 31:
```suggestion
        <help>Interface to use to calculate the proper range, when selected, a range may be specified as a suffix (e.g. ::1, ::400)</help>
```

I'd rather not reinvent the wheel with "partial", the address isn't partial per-se, but can be considered a sufix (with padded leading zeroes).

fichtner (approved):
otherwise ok, you own all bugs here in the forseeable future ;)

Monviech:
> otherwise ok, you own all bugs here in the forseeable future ;)

😅

PR #8450, opened 2025-03-17
Suricata: Add support for PCAP logging

0xThiebaut:
Another commonly used feature in network security monitoring is the capturing of all traffic (e.g., Malcolm, Security Onion, ...). This PR adds support for this feature through Suricata.

Notably:
- This PR changes the default number of retained PCAP files from `2000` to `5` in order to avoid accidental storage exhaustion.
- Some comments are furthermore updated to align with the [reference configuration](https://github.com/OISF/suricata/blob/suricata-7.0.8/suricata.yaml.in#L368-L432).
- The PR relies on the default `normal` PCAP logging mode to ease storage estimations , `multi` relying on the number of threads.


AdSchellevis:
As mentioned in the other ticket as well, we already have a packet capture option included in OPNsense, when functionality is missing there, it might be better to investigate what that is.....

PR #8452, opened 2025-03-18
Thermal Sensors Widget (core/cpu)

NOYB:


AdSchellevis:
let's try not to overcomplicate things...

Issue #7844, opened 2024-09-05
captive portal does not send Calling Station ID to the client during the connection


ITMPGIT:
**Important notices**

Before you add a new report, we ask you kindly to acknowledge the following:

- [ ] I have read the contributing guide lines at https://github.com/opnsense/core/blob/master/CONTRIBUTING.md
- [x] I am convinced that my issue is new after having checked both open and closed issues at https://github.com/opnsense/core/issues?q=is%3Aissue

**Describe the bug**

I used the captive portal with authentication on an external freeradius and mysql server. In my tests I noticed that the opnsense captive portal does not send all the fields that in theory would be standard (rfc3580) for accounting storage, such as:
- Calling-Station-Id
- Called-Station-Id
among others

*Tip*: to validate your setup was working with the previous version, use opnsense-revert (https://docs.opnsense.org/manual/opnsense_tools.html#opnsense-revert)

**To Reproduce**

standard captive configuration with default template and freeradius authentication
**Expected behavior**

record client and nas mac addresses

**Relevant log files**

using tcpdump it is possible to check that the fields containing the mac address are not passed during authentication via captive portal.

```
tcpdump: listening on em0, link-type EN10MB (Ethernet), snapshot length 262144 bytes
13:21:58.620479 IP (tos 0x0, ttl 64, id 40615, offset 0, flags [none], proto UDP (17), length 110)
    192.168.1.72.52807 > x.x.x.x.radius: [udp sum ok] RADIUS, length: 82
        Access-Request (1), id: 0x43, Authenticator: xxxxxx6ab5b4888c75
          User-Name Attribute (1), length: 5, Value: eee
            0x0000:  6565 65
          Service-Type Attribute (6), length: 6, Value: Login
            0x0000:  0000 0001
          Framed-Protocol Attribute (7), length: 6, Value: #15
            0x0000:  0000 000f
          NAS-Identifier Attribute (32), length: 15, Value: 66d20a8ae1a23
            0x0000:  3636 6432 3061 3861 6531 6132 33
          NAS-Port Attribute (5), length: 6, Value: 0
            0x0000:  0000 0000
          NAS-Port-Type Attribute (61), length: 6, Value: Ethernet
            0x0000:  0000 000f
          User-Password Attribute (2), length: 18, Value:
            0x0000:  3e25 757f xxxx 0e4e 0480 88bd a88e 7ada
[...]

Issue #6712, opened 2023-08-05
Enable Radius Attribute


K-razy:
**Important notices**

Before you add a new report, we ask you kindly to acknowledge the following:

- [X] I have read the contributing guide lines at https://github.com/opnsense/core/blob/master/CONTRIBUTING.md
- [X] I am convinced that my issue is new after having checked both open and closed issues at https://github.com/opnsense/core/issues?q=is%3Aissue

Hello, hope you are fine.

I'm working on a OPNsense using freeradius to authenticate and do accounting during the visitor's connection regardless if it is Captive Portal or not.
However attributes **acct-input-octets** and **acct-input-octets**  avec 32-bit counters and are reset when 4 Go of data are reached.

Could you please implement Radius Extension, in particular **Acct-Input-Gigawords**, **Acct-Output-Gigawords** and their mechanisms to be able to monitor without losing the total amount of traffic ?
Link to the RFC linked to the request : https://datatracker.ietf.org/doc/html/rfc2869#page-22

Thanks in advance

Best Regards

AdSchellevis:
duplicate of https://github.com/opnsense/core/issues/6682 ?

K-razy:
Hello, 

Thanks for your answer, i've seen this request but in this case, he wanted a non Radius attribute

In my case, is to implement extension which complete existing monitoring in place.
My case is not for a commercial case, but only to monitor and applies some restrictions after that.

Perhaps it may exists, but i really don't know how to enable Gigawords in OPNsense.

Thanks for your time.

AdSchellevis:
If it's a question about Freeradius, we should probably move this to the plugin section.

K-razy:
Unfortunately no, i'm not using the freeradius plugin because some features are still experimental. I'm using an external FreeRadius server

I was looking about the accounting sent by OPNsense and i noticed that Gigawords are unfortunately not sent by OPNsense.

AdSchellevis:
ok, but then you're only looking at captive portal, there is no other process sending accounting messages...

PR #8458, opened 2025-03-19
radius: improved RFC alignment

swhite2:
- Adds proper Termination Cause handling, requires addition of `delete_reason` column in the captiveportal sqlite database.
- Implements `Acct-Input|Output-Gigawords` accounting handling (Fixes https://github.com/opnsense/core/issues/6712)
- implements `Called|Calling-Station-Id`. (Fixes https://github.com/opnsense/core/issues/7844)
	- `Called-Station-Id` implemented via additional optional input field
	- `Calling-Station-Id` requires an extra preauth hook, which in general is useful to handle per-session authentication properties.
- While here, silence output of pf during state kills when client is disconnected
- `NAS-IP-Address` isn't added, RFC2865 states it should be either `NAS-Identifier` of `NAS-IP-Address`.

Since it touches this area, perhaps we can remove the deprecated `$zoneid` param in the captive portal SessionController.

### DIFF
--- a/src/opnsense/mvc/app/controllers/OPNsense/CaptivePortal/Api/AccessController.php
+++ b/src/opnsense/mvc/app/controllers/OPNsense/CaptivePortal/Api/AccessController.php
@@ -39,6 +39,8 @@ use OPNsense\CaptivePortal\CaptivePortal;
  */
 class AccessController extends ApiControllerBase
 {
+    private static $arp = [];
+
     /**
      * request client session data
      * @param string $zoneid captive portal zone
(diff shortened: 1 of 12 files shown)


src/opnsense/scripts/OPNsense/CaptivePortal/process_accounting_messages.php
```diff
@@ -82,14 +83,14 @@
                 $stmt->bindParam(':sessionid', $row['sessionid']);
                 $stmt->execute();
                 if (method_exists($authenticator, 'stopAccounting')) {
-                    $time_spend = time() - $row['created'];
-                    $authenticator->stopAccounting($row['username'], $row['sessionid'], $time_spend, $row['bytes_in'], $row['bytes_out'], $row['ip_address']);
+                    $time_spend = time() - intval($row['created']);
```
swhite2, line 86: FWIW, this was done to fix `PHP Deprecated:  Implicit conversion from float 459.6189193725586 to int loses precision in /usr/local/opnsense/mvc/app/library/OPNsense/Auth/Radius.php on line 383`

src/opnsense/mvc/app/controllers/OPNsense/CaptivePortal/Api/AccessController.php
```diff
@@ -102,6 +100,25 @@ private function getClientIp()
...
+        $mac = null;
+
+        if (empty(self::$arp)) {
+            self::$arp = json_decode((new Backend())->configdRun("interface list arp json"), true);
+        }
+
+        if (self::$arp != null) {
+            foreach (self::$arp as $arp) {
+                if (!empty($arp['ip'] && $arp['ip'] == $ip)) {
+                    $mac = $arp['mac'];
```
AdSchellevis, line 114: may short circuit here (`return $arp['mac']`) in which case we can remove `$mac`

AdSchellevis (approved):
one small remark, but looks good, thanks!

PR #8461, opened 2025-03-20
dnsmasq: Change dhcptags position in menu to align with c9b1af5

Monviech:


AdSchellevis (approved):

fichtner (commented):
I'm not sure there's a need to flush all of this into the menu in the first place, but if we do we will have to micro-manage like this.

AdSchellevis:
I'm also ok by leaving them out, the titel should already avoid confusions (which is why I added them originally)

Monviech:
I kinda like these somehow. But if the plan is removal I'll close this.

AdSchellevis:
I can live with both options, so maybe merge this first and discuss if we want to keep the details later?

Issue #8457, opened 2025-03-19
bootgrid: add clear handle to change columns


AdSchellevis:
**Important notices**

Before you add a new report, we ask you kindly to acknowledge the following:

- [x] I have read the contributing guide lines at https://github.com/opnsense/core/blob/master/CONTRIBUTING.md
- [x] I am convinced that my issue is new after having checked both open and closed issues at https://github.com/opnsense/core/issues?q=is%3Aissue

**Is your feature request related to a problem? Please describe.**

The bootgrid plugin lacks a clear way to set/unset options, which was needed in the new filter ui. For now we disabled the functionality as it keeps pushing refresh to the backend:

https://github.com/opnsense/core/blob/62536e40eee7a61b46dd9d4096e42bf82b39a692/src/opnsense/mvc/app/views/OPNsense/Firewall/filter_rule.volt#L562-L573

While looking at this, it might also be good to take a look at a reset to defaults in a more generic way, there was already a PR on our end, which likely slipped our minds. The filter page now executes:

https://github.com/opnsense/core/blob/62536e40eee7a61b46dd9d4096e42bf82b39a692/src/opnsense/mvc/app/views/OPNsense/Firewall/filter_rule.volt#L60C12-L65

Which would also be solvable in a more generic way using https://github.com/opnsense/jquery-bootgrid/pull/6

**Describe the solution you like**

Some minor feature enhancements in the library we use, so the amount of glue in our implementation could be less. A clear `set_colums()` and `get_columns()` might be the cleanest option so the implementation may merge that with the set it needs.

**Describe alternatives you considered**

more glue in the volt templates nobody wants :)

**Additional context**

-

PR #8462, opened 2025-03-20
bootgrid: improve UX and extend bootgrid behavior

swhite2:
Added options:

- `columnSelectForceReload` (default false). Changes current behavior for all bootgrids (currently adding a new column re-fetches the data, which is unnecessary in most cases). Caches response internally, thereby assuming the data for a newly added column is already present.
- `headerFormatters` object. Can be explicitly set via `data-headerFormatter-<identifier>` or implicitly linked via the row id.
- `setColumns` function (`grid.bootgrid("setColumns", ['colA', 'colB' ...])`). Marks passed columns for addition. Requires either a `reload` or `softreload` to apply.
- `unsetColumns` function (`grid.bootgrid("unsetColumns", ['colA', 'colB' ...])`). Marks passed columns for removal. Requires either a `reload` or `softreload` to apply.
- `softreload` function (`grid.bootgrid("softreload")`).

UX changes:

- `headerFormatters` now makes sure that if column headers require styling, the styling doesn't flash and is applied from the beginning / during reloads.
- The "Loading..." status has been replaced with a transparent overlay containing a spinner. This prevents unnecessary style flashing when data is reloaded, i.e. when scrolling through pages, setting columns, forced refreshes etc.
- Added "reset to defaults" button, resetting the sort, visiblity and rowcount options to the controller defaults (removes them from localstorage).

Fixes https://github.com/opnsense/core/issues/8457

### DIFF
--- a/src/opnsense/mvc/app/views/OPNsense/Firewall/filter_rule.volt
+++ b/src/opnsense/mvc/app/views/OPNsense/Firewall/filter_rule.volt
@@ -57,13 +57,6 @@
         // Test if the UUID is valid, used to determine if automation model or internal rule
         const uuidRegex = /^[0-9a-f]{8}-[0-9a-f]{4}-[1-5][0-9a-f]{3}-[89ab][0-9a-f]{3}-[0-9a-f]{12}$/i;
 
-        // XXX: Clear any stored column visibility settings for the firewall filter page
-        // Synergizes with the Inspect Button since it unhides certain columns that should be hidden on page load
-        // Prevents messed up views, we always control the intentional default of the page
-        Object.keys(localStorage)
-            .filter(key => key.startsWith("visibleColumns[/ui/firewall/filter/"))
-            .forEach(key => localStorage.removeItem(key));
-
         // Initialize grid
         const grid = $("#{{formGridFilterRule['table_id']}}").UIBootgrid({
             search:'/api/firewall/filter/search_rule/',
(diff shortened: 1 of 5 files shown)


Monviech:
@swhite2 
Since we talked about removing the select checkbox for internal rules:

It's added here:

https://github.com/opnsense/core/blob/ec634601fece2fea243b7f731acda941c6aa9454/src/opnsense/www/js/jquery.bootgrid.js#L608-L612

They are added due to this default:

https://github.com/opnsense/core/blob/ec634601fece2fea243b7f731acda941c6aa9454/src/opnsense/www/js/opnsense_bootgrid_plugin.js#L128

Can you turn this checkbox into a formatter?

swhite2:
> Can you turn this checkbox into a formatter?

@Monviech let's keep this out of the scope of this PR for now, there is some more complexity involved. For now a request handler can be used to validate correct selections

src/opnsense/mvc/app/views/OPNsense/Firewall/filter_rule.volt
```diff
@@ -556,23 +546,11 @@
 
         $("#internal_rule_selector").detach().insertAfter("#type_filter_container");
         $('#all_rules_checkbox').change(function(){
-            grid.bootgrid('reload');
+            const isChecked = $('#all_rules_checkbox').is(':checked');
+            grid.bootgrid(isChecked ? "setColumns" : "unsetColumns", ['evaluations', 'states', 'packets', 'bytes'])
```
AdSchellevis, line 550:
```suggestion
            grid.bootgrid(isChecked ? "setColumns" : "unsetColumns", ['evaluations', 'states', 'packets', 'bytes']);
```

AdSchellevis (changes requested):
@swhite2  missed one semicolon, but other than that this looks good, thanks!

PR #8465, opened 2025-03-21
dnsmasq: Add filter function for interfaces and tags with multiselect

Monviech:
I really want something like this.

It's a combined selectpicker for interfaces and tags, and you can quickly filter seamlessly through multiple tabs to get a quick overview of the current configuration.

It improves UX quite a bit if the configuration is larger, especially if multiple tags and interfaces are used.

@AdSchellevis It might need some cleanup but the concept works.

### DIFF
--- a/src/opnsense/mvc/app/controllers/OPNsense/Dnsmasq/Api/SettingsController.php
+++ b/src/opnsense/mvc/app/controllers/OPNsense/Dnsmasq/Api/SettingsController.php
@@ -40,13 +40,12 @@ class SettingsController extends ApiMutableModelControllerBase
      * Tags and interface filter function.
      * Interfaces are tags too, in the sense of dnsmasq.
      *
-     * @param array $filterValues List of values to filter against (e.g. UUIDs, interface names).
-     * @param array $fieldNames   List of field names to extract values from in each record.
-     *
      * @return callable|null
      */
-    private function buildFilterFunction(array $filterValues, array $fieldNames): ?callable
+    private function buildFilterFunction(): ?callable
     {
+        $filterValues = $this->request->get('tags') ?? [];
+        $fieldNames = ['interface', 'set_tag', 'tag'];
         if (empty($filterValues)) {
             return null;
         }
@@ -58,11 +57,8 @@ class SettingsController extends ApiMutableModelControllerBase
                     continue;
                 }
 
-                // Some fields allow multi selection
-                $fieldValues = array_map('trim', explode(',', (string)$record->{$fieldName}));
-
                 // Match field values against filter list
-                foreach ($fieldValues as $value) {
+                foreach (array_map('trim', explode(',', (string)$record->{$fieldName})) as $value) {
                     if (in_array($value, $filterValues, true)) {
                         return true;
                     }
@@ -87,9 +83,7 @@ class SettingsController extends ApiMutableModelControllerBase
     /* hosts */
     public function searchHostAction()
     {
-        $filters = $this->request->get('tags') ?? [];
-        $filter_funct = $this->buildFilterFunction($filters, ['set_tag']);
-        return $this->searchBase('hosts', null, null, $filter_funct);
+        return $this->searchBase('hosts', null, null, $this->buildFilterFunction());
     }
 
     public function getHostAction($uuid = null)
@@ -212,9 +206,7 @@ class SettingsController extends ApiMutableModelControllerBase
     /* dhcp ranges */
     public function searchRangeAction()
     {
-        $filters = $this->request->get('tags') ?? [];
-        $filter_funct = $this->buildFilterFunction($filters, ['interface', 'set_tag']);
-        return $this->searchBase('dhcp_ranges', null, null, $filter_funct);
+        return $this->searchBase('dhcp_ranges', null, null, $this->buildFilterFunction());
     }
 
     public function getRangeAction($uuid = null)
@@ -240,9 +232,7 @@ class SettingsController extends ApiMutableModelControllerBase
     /* dhcp options */
     public function searchOptionAction()
     {
-        $filters = $this->request->get('tags') ?? [];
-        $filter_funct = $this->buildFilterFunction($filters, ['interface', 'tag']);
-        return $this->searchBase('dhcp_options', null, null, $filter_funct);
+        return $this->searchBase('dhcp_options', null, null, $this->buildFilterFunction());
     }
 
     public function getOptionAction($uuid = null)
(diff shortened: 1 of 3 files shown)


src/opnsense/mvc/app/controllers/OPNsense/Dnsmasq/Api/SettingsController.php
```diff
@@ -36,6 +36,24 @@ class SettingsController extends ApiMutableModelControllerBase
...
 
+    /**
+     * Create a lookup map of tag UUIDs to tag names
+     *
+     * @return array uuid => name
+     */
+    private function getTagUuidMap(): array
+    {
+        $map = [];
+        foreach ($this->getModel()->dhcp_tags->iterateItems() as $tag) {
```
AdSchellevis, line 47:
```suggestion
        foreach ($this->getModel()->dhcp_tags->iterateItems() as $uuid => $tag) {
```

src/opnsense/mvc/app/controllers/OPNsense/Dnsmasq/Api/SettingsController.php
```diff
@@ -50,7 +68,21 @@ public function getAction()
     /* hosts */
     public function searchHostAction()
     {
-        return $this->searchBase('hosts');
+        $filters = $this->request->get('tags');
+        $tagMap = $this->getTagUuidMap();
+
+        $filter_funct = function ($record) use ($filters, $tagMap) {
```
AdSchellevis, line 74: function could be generalised, the fields to choose from are `set_tag`, `tag` and `interface` when they exist.

src/opnsense/mvc/app/controllers/OPNsense/Dnsmasq/Api/SettingsController.php
```diff
@@ -50,7 +68,21 @@ public function getAction()
...
+            if (empty($filters)) {
```
AdSchellevis, line 75: we don't need to hook a filter function when none is set (reduce call overhead)

src/opnsense/mvc/app/controllers/OPNsense/Dnsmasq/Api/SettingsController.php
```diff
@@ -36,6 +36,24 @@ class SettingsController extends ApiMutableModelControllerBase
     protected static $internalModelName = 'dnsmasq';
     protected static $internalModelClass = '\OPNsense\Dnsmasq\Dnsmasq';
 
+    /**
+     * Create a lookup map of tag UUIDs to tag names
+     *
+     * @return array uuid => name
+     */
+    private function getTagUuidMap(): array
```
AdSchellevis, line 44: for simplicity I would only create a single method to collect options, a tag or an interface are technically the same anyway (it's either a `uuid` or an interface name)

src/opnsense/mvc/app/controllers/OPNsense/Dnsmasq/Api/SettingsController.php
```diff
@@ -50,7 +68,21 @@ public function getAction()
     /* hosts */
     public function searchHostAction()
     {
-        return $this->searchBase('hosts');
+        $filters = $this->request->get('tags');
+        $tagMap = $this->getTagUuidMap();
```
AdSchellevis, line 72: always filter on keys so you don't need to remap values on their way back.

src/opnsense/mvc/app/controllers/OPNsense/Dnsmasq/Api/SettingsController.php
```diff
@@ -262,4 +371,54 @@ public function delBootAction($uuid)
...
+            $interfaces[] = [
+                'value' => $key,
+                'label' => empty($intf->descr) ? strtoupper($key) : (string)$intf->descr
+            ];
+        }
+        usort($interfaces, fn($a, $b) => strcasecmp($a['label'], $b['label']));
+
+        // Tags
+        $tags = [];
+        foreach ($this->getModel()->dhcp_tags->iterateItems() as $tag) {
```
AdSchellevis, line 398: see above

src/opnsense/mvc/app/controllers/OPNsense/Dnsmasq/Api/SettingsController.php
```diff
@@ -262,4 +371,54 @@ public function delBootAction($uuid)
...
+                ];
+            }
+        }
+        usort($tags, fn($a, $b) => strcasecmp($a['label'], $b['label']));
+
+        // Assemble result
+        return [
+            'tags' => [
+                'label' => gettext('Tags'),
+                'icon'  => 'fa fa-tag text-primary',
```
AdSchellevis, line 413: Although the icons are funny, if we do want to use them more often we might want to consider adding a wrapper for the pattern. (see also further below)

src/opnsense/mvc/app/views/OPNsense/Dnsmasq/settings.volt
```diff
@@ -136,13 +157,55 @@
             }
         });
 
+        // Populate tag selectpicker
+        ajaxCall('/api/dnsmasq/settings/getTagList', {}, function (data) {
```
AdSchellevis, line 161: Since we tend to populate more dropdowns lately, it might be an idea to add a wrapper in the `opnsense_ui.js` file.
In previous projects, I've used something like:

```
/**
 * fetch option list for remote url
 * @param url
 * @param key_attr
 * @param val_attr
 * @param params
 * @param store_data store data in data attribute
 */
$.fn.ajax_fetch_options = function(url, key_attr, val_attr, params, store_data) {
    var deferred = $.Deferred();
    var obj = $(this);
    obj.empty();

    ajaxGet(url, params, function(data,status){
        if (store_data !== undefined) {
            obj.data("store", data);
        }

        data.map(function (item) {
            obj.append($("<option/>").attr({"value": item[key_attr]}).text(htmlDecode(item[val_attr])));

        });
        if (obj.hasClass('selectpicker')) {
            obj.selectpicker('refresh');
        }
        obj.change();
        deferred.resolve();
    });

    return deferred.promise();
};
```

In which case you can chain them easily and only need something like this in the implementation:

```
$("#..").ajax_fetch_options(..);
```

AdSchellevis (changes requested):
@Monviech I'm ok with the idea, but the code needs work. Currently I think you're overthinking some things here. There's also some room for optimization in general (not related to this PR, but just a pattern I see emerging more often now)

src/opnsense/mvc/app/views/OPNsense/Dnsmasq/settings.volt
Monviech: Having something generalized for this sounds like a good idea for the future.

AdSchellevis:
@Monviech there you go :)

Monviech:
@AdSchellevis Oh really nice, I see what you did in the filter function. That makes total sense now that I see it. Thank you. And as bonus even got a generalized helper for the UI :)

PR #8468, opened 2025-03-22
dnsmasq: Add internalModelUseSafeDelete to prevent deletion of currently used tags

Monviech:
While testing I added and deleted tags, and then my configuration became inconsistent because tags are used in model relation fields.

This should prevent inconsistent configuration states as it disallows tags in use from being deleted.


AdSchellevis (approved):
@Monviech nice catch, thanks!

PR #8476, opened 2025-03-24
Display OpenVPN virtual IPv6 addresses for OpenVPN clients in dashboard widget

cs-1:
Due to the re-organisation of code the widget for OpenVPN clients lost the ability to display virtual IPv6 addresses for clients. This patch fixes this issue.

### DIFF
--- a/src/opnsense/www/js/widgets/OpenVPNClients.js
+++ b/src/opnsense/www/js/widgets/OpenVPNClients.js
@@ -130,7 +130,7 @@ export default class OpenVPNClients extends BaseTableWidget {
                                 </span>
                             </div>
                             <div>
-                                ${client.real_address} | ${client.virtual_address}
+                                ${client.real_address} | ${client.virtual_address} | ${client.virtual_ipv6_address}
                             </div>
                             <div>
                                 ${client.connected_since}
--- a/src/opnsense/www/js/widgets/OpenVPNClients.js
+++ b/src/opnsense/www/js/widgets/OpenVPNClients.js
@@ -25,112 +25,89 @@
  */
 
 export default class OpenVPNClients extends BaseTableWidget {
-  constructor(config) {
-    super(config);
+    constructor(config) {
+        super(config);
 
-    this.locked = false;
-  }
+        this.locked = false;
+    }
 
-  getGridOptions() {
-    return {
-      sizeToContent: 650,
-    };
-  }
+    getGridOptions() {
+        return {
+            sizeToContent: 650
+        };
+    }
 
-  getMarkup() {
-    let $container = $('<div id="ovpn-client-table-container"></div>');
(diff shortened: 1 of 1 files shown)


src/opnsense/www/js/widgets/OpenVPNClients.js
```diff
@@ -130,7 +130,7 @@ export default class OpenVPNClients extends BaseTableWidget {
                                 </span>
                             </div>
                             <div>
-                                ${client.real_address} | ${client.virtual_address}
+                                ${client.real_address} | ${client.virtual_address} | ${client.virtual_ipv6_address}
```
Monviech, line 133: Can you handle the case if a client does not have a virtual_ipv6_address?

I assume right now it would looke like this:

```10.0.0.1:12345 | 10.8.0.6 | ```

**Review by Monviech — CHANGES_REQUESTED** [review]

File: src/opnsense/www/js/widgets/OpenVPNClients.js [hunk 2]
```diff
@@ -130,7 +130,7 @@ export default class OpenVPNClients extends BaseTableWidget {
                                 </span>
                             </div>
                             <div>
-                                ${client.real_address} | ${client.virtual_address}
+                                ${client.real_address} | ${client.virtual_address}${client.virtual_ipv6_address ? ` | ${client.virtual_ipv6_address}` : ""}
```
fichtner, line 133: I'd prefer a join(' | ', $values) type of statement here for an array that was fed earlier so this doesn't grow in ugliness when the next edge case or refactor is due

cs-1: Understood. JS noob here, getting @lucaspalomodevelop to help me. :)

Monviech: Maybe try something like this (not tested just an idea)

`${[client.real_address, client.virtual_address, client.virtual_ipv6_address].filter(Boolean).join(' | ')}`

cs-1: @lucaspalomodevelop has proposed a patch, we're in progress of testing it.

fichtner: @Monviech separating the view from the logic would be better, else you're still fiddling changes into the view scope

cs-1: We'll try to move the logic outside `$clients.append(...)`.

cs-1: We moved the logic within the widget's source. Could you guys be so kind and check if the new structure is what you had in mind? We tested the patch against OPNsense 25.1.3-amd64 and all seems fine.

fichtner (approved):
LGTM, thank you @cs-1 and @lucaspalomodevelop

Issue #8479, opened 2025-03-24
dnsmasq: ra-params template issue


Monviech:
**Important notices**

Before you add a new report, we ask you kindly to acknowledge the following:

- [x] I have read the contributing guide lines at https://github.com/opnsense/core/blob/master/CONTRIBUTING.md
- [x] I am convinced that my issue is new after having checked both open and closed issues at https://github.com/opnsense/core/issues?q=is%3Aissue

**Describe the bug**

Right now, `ra_param` uses `constructor` in the template to add the mandatory interface to the option:

`--ra-param=<interface>,[mtu:<integer>|<interface>|off,][high,|low,]<ra-interval>[,<router lifetime>]`

Though, if you don't want to use the constructor and set the range statically while slaac is enabled, this will make the template fail because constructor is not defined.

**Expected behavior**

Router Advertisements can be enabled with and without constructor without failing the template.

or

constructor is a requirement for RA, the man page always uses constructor in all RA examples too.

**Proposed Fix**

Add validation that constructor is mandatory when setting an RA mode.

PR #8482, opened 2025-03-25
dnsmasq: Require interface when ra_mode is selected due to ra-param interface requirement.

Monviech:
There were a few choices how to solve this, but this seemed to be the best in my opinion:

- Constructor is not a requirement for using RA now, as you can also use SLAAC with static IPv6 e.g. ``2003:aaaa:bbbb:cc00::`` without using a constructor, if you do not want to set a GUA on the interface.
- ``Default`` ra_mode with enable_ra does not use custom ra_params, but use implicit global defaults from enable_ra
- When a custom ra_mode is selected, the interface becomes a requirement, but thats how SLAAC works anyway, it must use an interface.

This prevents any new checkboxes or type selectpickers from being introduced.

Validation prevents misconfiguration and style visibility hides options automatically that are not executed in the template for the current choice.

Fixes: https://github.com/opnsense/core/issues/8479

### DIFF
--- a/src/opnsense/mvc/app/controllers/OPNsense/Dnsmasq/forms/dialogDHCPrange.xml
+++ b/src/opnsense/mvc/app/controllers/OPNsense/Dnsmasq/forms/dialogDHCPrange.xml
@@ -28,7 +28,7 @@
         <label>Constructor</label>
         <type>dropdown</type>
         <style>selectpicker</style>
-        <help>Interface to use to calculate a DHCPv6 and/or RA range. Start address can then be specified as a suffix (e.g. ::, ::1 or ::400).</help>
+        <help>Interface to use to calculate a DHCPv6 or RA range. Start address can then be specified as a suffix (e.g. ::, ::1 or ::400).</help>
         <grid_view>
             <visible>false</visible>
         </grid_view>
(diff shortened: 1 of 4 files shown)


src/opnsense/mvc/app/controllers/OPNsense/Dnsmasq/forms/dialogDHCPrange.xml
```diff
@@ -27,8 +27,8 @@
         <id>range.constructor</id>
         <label>Constructor</label>
         <type>dropdown</type>
-        <style>selectpicker style_dhcpv6</style>
-        <help>Interface to use to calculate the proper range, when selected, a range may be specified as a suffix (e.g. ::1, ::400)</help>
+        <style>selectpicker</style>
+        <help>Interface to use to calculate a DHCPv6 and/or RA range. Start address can then be specified as a suffix (e.g. ::, ::1 or ::400).</help>
```
fichtner, line 31:
```suggestion
        <help>Interface to use to calculate a DHCPv6 or RA range. Start address can then be specified as a suffix (e.g. ::, ::1 or ::400).</help>
```

src/opnsense/mvc/app/models/OPNsense/Dnsmasq/Dnsmasq.php
```diff
@@ -194,6 +203,7 @@ public function performValidation($validateFullModel = false)
                     );
                 }
             }
+
```
fichtner, line 206:
```suggestion
```

fichtner (approved):
ok minus comments

PR #8483, opened 2025-03-25
auth/user: Remove styles conflicting with 387c381

Monviech:
Some dropdowns in dialogUser did not stay in line due to the important width style.

Also, the col-md-* do not exist anymore, they have been replaced with % width in the base dialog.


fichtner (approved):
And so it begins 😉

Monviech:
> And so it begins 😉

😇

Issue #8472, opened 2025-03-22
DHCPv6 clients requesting both an address and a prefix delegation show partially incorrect in Leases tab


agoodkind:
## Acknowledgements

* I have read the contributing guide lines at https://github.com/opnsense/core/blob/master/CONTRIBUTING.md
* I am convinced that my issue is new after having checked both open and closed issues at https://github.com/opnsense/core/issues?q=is%3Aissue

## Description

DHCPv6 clients requesting both an address and a prefix delegation (with the same DUID) will appear in the "Services: ISC DHCPv6: Leases" UI display either in the "Leases" tab or in the "Delegated Prefixes" tab, but not both.

# Reproducing

Have a client request an address and a prefix delegation (Ubuntu 24.04 seems to do this by default?). Check the Ubuntu logs:

```
$ journalctl -b -u systemd-networkd.service | grep DHCP
Mar 06 11:47:29 <redacted> systemd-networkd[293916]: eno1: DHCPv6 address <redacted>:82a1::2000/128 (valid for 1h 59min 59s, preferred for 1h 14min 59s)
Mar 06 11:47:29 <redacted> systemd-networkd[293916]: eno1: DHCP: received delegated prefix <redacted>:82f0::/62
```

Then check the dhcpd leases file:

```
$ sudo cat /var/dhcpd/var/db/dhcpd6.leases | grep -E 'iaaddr|iaprefix' | sort -u
  iaaddr <redacted>:82a0::1288 {
  iaaddr <redacted>:82a1::2000 {
  iaprefix <redacted>:82f0::/62 {
```

Then query the OPNsense API for leases (and note the absence of the `82a1::2000` address):

```
% curl -s --cookie 'PHPSESSID=<redacted>' https://<redacted>/api/dhcpv6/leases/searchLease/ | python3 -m json.tool | grep '"address"'
            "address": "<redacted>:82a0::1288",
```

Then query the OPNsense API for prefixes (where we will indeed see the delegated prefix):

```
% curl -s --cookie 'PHPSESSID=<redacted>' https://<redacted>/api/dhcpv6/leases/searchPrefix/ | python3 -m json.tool | grep '"prefix"'
            "prefix": "<redacted>:82f0::/62",
```

## Expected Behavior

I'd expect to see the address appear on the "Leases" tab, and the delegated prefix in the "Delegated Prefixes" tab.

## Environment

OPNsense 25.1.3-amd64
FreeBSD 14.2-RELEASE-p2

[...]

PR #8492, opened 2025-03-27
DHCP6: use `lease_type` to key lease map in addition to `iaid_duid`

agoodkind:
## Description

Currently, `get_leases6` misreports clients that request an address *and* a prefix delegation. Adding the lease type to the map key prevents this.

## Testing

Request an address and a PD from the same client (same DUID) and verify that the address appeared in the output of `/api/dhcpv6/leases/searchLease/` and that the delegated prefix appeared in `/api/dhcpv6/leases/searchPrefix/`.

Fixes: #8472

### DIFF
--- a/src/opnsense/scripts/dhcp/get_leases6.py
+++ b/src/opnsense/scripts/dhcp/get_leases6.py
@@ -148,7 +148,7 @@ if __name__ == '__main__':
                 elif len(line) > 1 and line[0] == '}' and len(cur_lease) > 0:
                     cur_lease.append(line)
                     parsed_lease = parse_lease(cur_lease)
-                    last_leases[parsed_lease['iaid_duid']] = parsed_lease
+                    last_leases[parsed_lease['lease_type'], parsed_lease['iaid_duid']] = parsed_lease
                     cur_lease = []
                 elif len(cur_lease) > 0:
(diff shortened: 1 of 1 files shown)


agoodkind:
@AdSchellevis

src/opnsense/scripts/dhcp/get_leases6.py
```diff
@@ -148,7 +148,7 @@ def parse_lease(lines):
                 elif len(line) > 1 and line[0] == '}' and len(cur_lease) > 0:
                     cur_lease.append(line)
                     parsed_lease = parse_lease(cur_lease)
-                    last_leases[parsed_lease['iaid_duid']] = parsed_lease
+                    last_leases[parsed_lease['lease_type'], parsed_lease['iaid_duid']] = parsed_lease
```
AdSchellevis, line 151:
```suggestion
                    last_leases["%s,%s" % (parsed_lease['lease_type'], parsed_lease['iaid_duid'])] = parsed_lease
```

I prefer to use single value keys for debugging, the suggested change should have the same result as you're looking for.

AdSchellevis (changes requested):

agoodkind:
> @agoodkind thanks!

No problem ☺️

PR #8493, opened 2025-03-27
firewall/filter: Use fetch_options from opnsense_ui.js to build interface_select selectpicker

Monviech:


src/opnsense/mvc/app/controllers/OPNsense/Firewall/Api/FilterController.php
```diff
@@ -342,7 +342,8 @@ public function getInterfaceListAction()
                 'items' => [
                     [
                         'value' => '',
-                        'label' => gettext('Any')
+                        'label' => gettext('Any'),
+                        'selected' => true
```
fichtner, line 346:
```suggestion
                        'selected' => true,
```

Reduces the next diff when adding another option ;)

AdSchellevis (approved):

PR #8496, opened 2025-03-27
bootgrid: resizable columns

swhite2:
Bootgrid option: `resizableColumns`: `true` (default false)

### DIFF
--- a/src/opnsense/mvc/app/views/OPNsense/Firewall/filter_rule.volt
+++ b/src/opnsense/mvc/app/views/OPNsense/Firewall/filter_rule.volt
@@ -66,6 +66,7 @@
             del:'/api/firewall/filter/del_rule/',
             toggle:'/api/firewall/filter/toggle_rule/',
             options: {
+                resizableColumns: true,
                 triggerEditFor: getUrlHash('edit'),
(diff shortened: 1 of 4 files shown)


Monviech:
I gave this a quick try right now and I like it.

src/opnsense/mvc/app/views/OPNsense/Firewall/filter_rule.volt
```diff
@@ -473,6 +474,7 @@
             $('[data-toggle="tooltip"]').tooltip();
         });
 
+
```
fichtner, line 477:
```suggestion
```

AdSchellevis (approved):

AdSchellevis:
@swhite2 we might need to revert this one, when a `<td>` text is longer than the column, it doesn't apply the `ellipsis` anymore. Looking at the styles it should:

```
.bootgrid-table td {
    overflow: hidden;
    -ms-text-overflow: ellipsis;
    -o-text-overflow: ellipsis;
    text-overflow: ellipsis;
    white-space: nowrap;
} 
```

I'm not sure if it's fixable, but when it's difficult and we're already working on an alternative implementation, it might be safer to drop this for now. 

current view:

image

previous: 

image

Issue #8487, opened 2025-03-26
dnsmasq: Support dhcp-host options


gspannu:
**Important notices**

Before you add a new report, we ask you kindly to acknowledge the following:

- [x] I have read the contributing guide lines at https://github.com/opnsense/core/blob/master/CONTRIBUTING.md
- [x] I am convinced that my issue is new after having checked both open and closed issues at https://github.com/opnsense/core/issues?q=is%3Aissue

**Is your feature request related to a problem? Please describe.**

Ability in dnsmasq GUI to assign _static DHCP IP addresses_ to clients as _per their MAC address_. 
- This feature is quite well implemented in dnsmasq as per the `dhcp-host` option, well documented in the [manpage](https://dnsmasq.org/docs/dnsmasq-man.html#lbAE). 
- This feature **already exists** in OPNsense in **ISC** v4/v6 and **KEA** implementations through GUI. 
- The `dhcp-host` DHCP static address option in dnsmasq is well optioned as it allows IPV4/IPv6/IPv6 range, lease time, tag settings to a client in a single module; and is quite a powerful **implementation for multiple use cases**.

dnsmasq allows static DHCP mapping for a client using MAC Hardware address to be mapped with following parameters 
`--dhcp-host=[<hwaddr>][,id:<client_id>|*][,set:<tag>][,tag:<tag>][,<ipaddr>][,<hostname>][,<lease_time>][,ignore]
`

**Describe the solution you like**
- Mechanism in the dnsmasq GUI (perhaps a tab section) where the dhcp-host option can be configured for each client.
- Maybe make the GUI similar to the Services > Kea DHCP > Kea DHCPv4 > Reservations
- but extend it further by making additional entry boxes for IPv6 address/ IPv6 address prefix, least time, ignore checkbox, etc..

**Alternative Solution**
- dnsmasq also supports the `--dhcp-hostsfile=<path>` and this maybe considered a temporary workaround to allow defining the above dhcp-hosts configuration in a separate custom file.
- Have a GUI option to specify the name of this custom file.
I also admit that this not a neat/elegant solution, but perhaps maybe considered until the _dhcp-host GUI_ is not implemented.

**Additional context**

- This feature has been tested in dnsmasq (Debian Bookworm) and is working as documented.
- dnsmasq manpage context

> --dhcp-host=[<hwaddr>][,id:<client_id>|*][,set:<tag>][,tag:<tag>][,<ipaddr>][,<hostname>][,<lease_time>][,ignore]
[...]

PR #8497, opened 2025-03-27
dnsmasq: Add full dhcp-host support for IPv4 and IPv6

Monviech:
Fixes: https://github.com/opnsense/core/issues/8487

Directive:
```
--dhcp-host=[<hwaddr>][,id:<client_id>|*][,set:<tag>][,tag:<tag>][,<ipaddr>][,<hostname>][,<lease_time>][,ignore]
```

This should almost fully satisfy the requirements of the man page regarding this directive.

Test dataset:

```
    <hosts uuid="9e5eb2a8-0ab2-4ed9-af5b-0c2787ddbfab">
      <host>hostabc</host>
      <domain>example.com</domain>
      <ip>172.16.1.200,fd12:3456:789a:1::2</ip>
      <client_id>*</client_id>
      <hwaddr>aa:aa:aa:aa:aa:aa,bb:bb:bb:bb:bb:bb</hwaddr>
      <lease_time/>
      <ignore>1</ignore>
      <set_tag/>
      <descr/>
      <comments/>
      <aliases>abc1.example.com,abc2.example.com</aliases>
    </hosts>
    <hosts uuid="883bd818-2885-443d-a699-90fb55fe3466">
      <host>hostd</host>
      <domain>example.com</domain>
      <ip>172.16.1.199,fd12:3456:789a:1::1</ip>
      <client_id>12121-3243242-222</client_id>
      <hwaddr/>
      <lease_time>4040</lease_time>
      <ignore>0</ignore>
      <set_tag>09e84129-c07b-4133-8251-f17c4d00f675</set_tag>
      <descr/>
      <comments/>
      <aliases>abc1.example.com,abc2.example.com</aliases>
    </hosts>
    <hosts uuid="af0c1d49-cc38-4468-9af4-61c31933dc79">
      <host>hostggg</host>
      <domain>example.com</domain>
      <ip>192.168.4.5</ip>
      <client_id/>
      <hwaddr/>
      <lease_time/>
      <ignore>0</ignore>
      <set_tag/>
      <descr/>
      <comments/>
      <aliases/>
    </hosts>
    <hosts uuid="04b21c12-6a51-4e45-b219-54b8052c9a17">
      <host>hostfff</host>
      <domain>example.com</domain>
      <ip>192.168.4.6</ip>
      <client_id/>
      <hwaddr>fe:fe:fe:fe:fe:fe</hwaddr>
      <lease_time/>
      <ignore>0</ignore>
      <set_tag/>
      <descr/>
      <comments/>
      <aliases/>
    </hosts>
```

Result:

```/usr/local/etc/dnsmasq.conf```
```
dhcp-host=id:*,aa:aa:aa:aa:aa:aa,bb:bb:bb:bb:bb:bb,172.16.1.200,[fd12:3456:789a:1::2],hostabc,86400,ignore
dhcp-host=id:12121-3243242-222,set:09e84129c07b41338251f17c4d00f675,172.16.1.199,[fd12:3456:789a:1::1],hostd,4040
dhcp-host=fe:fe:fe:fe:fe:fe,192.168.4.6,hostfff,86400
```

```/var/etc/dnsmasq-hosts```
```
172.16.1.200	hostabc.example.com hostabc
172.16.1.200	abc1.example.com
172.16.1.200	abc2.example.com
fd12:3456:789a:1::2	hostabc.example.com hostabc
fd12:3456:789a:1::2	abc1.example.com
fd12:3456:789a:1::2	abc2.example.com
172.16.1.199	hostd.example.com hostd
172.16.1.199	abc1.example.com
172.16.1.199	abc2.example.com
fd12:3456:789a:1::1	hostd.example.com hostd
fd12:3456:789a:1::1	abc1.example.com
fd12:3456:789a:1::1	abc2.example.com
192.168.4.5	hostggg.example.com hostggg
192.168.4.6	hostfff.example.com hostfff
```

### DIFF
--- a/src/opnsense/mvc/app/models/OPNsense/Dnsmasq/Dnsmasq.xml
+++ b/src/opnsense/mvc/app/models/OPNsense/Dnsmasq/Dnsmasq.xml
@@ -85,7 +85,10 @@
                 <FieldSeparator>,</FieldSeparator>
                 <AsList>Y</AsList>
             </ip>
-            <client_id type="TextField"/>
+            <client_id type="TextField">
+                <Mask>/^(?:\*|(?:[0-9A-Fa-f]{2}(?::[0-9A-Fa-f]{2})+))$/</Mask>
+                <ValidationMessage>Value must be a colon-separated hexadecimal sequence (e.g., 01:02:f3) or "*".</ValidationMessage>
+            </client_id>
             <hwaddr type="MacAddressField">
                 <FieldSeparator>,</FieldSeparator>
                 <AsList>Y</AsList>
--- a/src/opnsense/mvc/app/models/OPNsense/Dnsmasq/Dnsmasq.xml
+++ b/src/opnsense/mvc/app/models/OPNsense/Dnsmasq/Dnsmasq.xml
@@ -82,8 +82,18 @@
             <ip type="NetworkField">
                 <Required>Y</Required>
                 <NetMaskAllowed>N</NetMaskAllowed>
+                <FieldSeparator>,</FieldSeparator>
+                <AsList>Y</AsList>
             </ip>
-            <hwaddr type="MacAddressField"/>
+            <client_id type="TextField"/>
+            <hwaddr type="MacAddressField">
+                <FieldSeparator>,</FieldSeparator>
+                <AsList>Y</AsList>
+            </hwaddr>
+            <lease_time type="IntegerField">
+                <MinimumValue>1</MinimumValue>
+            </lease_time>
+            <ignore type="BooleanField"/>
             <set_tag type="ModelRelationField">
                 <Model>
                     <tag>
--- a/src/opnsense/service/templates/OPNsense/Dnsmasq/dnsmasq.conf
+++ b/src/opnsense/service/templates/OPNsense/Dnsmasq/dnsmasq.conf
@@ -200,13 +200,9 @@ id:{{ host.client_id }},
 {%-     if host.set_tag -%}
 set:{{ host.set_tag|replace('-', '') }},
 {%-     endif -%}
-{%-     if host.ip -%}
-{%-         set ip_list = host.ip.split(',') -%}
-{%-         for ip in ip_list -%}
-{%-             set ip = ip.strip() -%}
-{{ '[' if ':' in ip }}{{ ip }}{{ ']' if ':' in ip }}{%- if not loop.last %},{% endif %}
-{%-         endfor -%}
-{%-     endif -%}
+{%-     for ip in host.ip.split(',')|map('trim')  if host.ip -%}
+{{ '[' ~ ip ~ ']' if ':' in ip else ip  }}{%- if not loop.last %},{% endif %}
+{%-     endfor -%}
 {%-     if host.host -%}
 ,{{ host.host }}
 {%-     endif -%}
--- a/src/opnsense/service/templates/OPNsense/Dnsmasq/dnsmasq.conf
+++ b/src/opnsense/service/templates/OPNsense/Dnsmasq/dnsmasq.conf
@@ -186,9 +186,37 @@ ra-param={{ helpers.physical_interface(dhcp_range.interface) }}
 {% endfor %}
 
 {% for host in helpers.toList('dnsmasq.hosts') %}
-{%    if host.hwaddr and host.hwaddr.find(':') == -1%}
-dhcp-host={{host.hwaddr}}{% if host.set_tag%},set:{{host.set_tag|replace('-','')}}{% endif %},{{host.ip}}
-{%    endif %}
+{#     Skip if MAC is missing or invalid (no colon), unless client_id is present #}
+{%     if not host.client_id and (not host.hwaddr or host.hwaddr.find(':') == -1) %}
+{%         continue %}
+{%     endif %}
+dhcp-host=
+{%-     if host.client_id -%}
+id:{{ host.client_id }}
+{%-     endif -%}
+{%-     if host.hwaddr -%}
+{% if host.client_id %},{% endif %}{{ host.hwaddr }}
+{%-     endif -%}
+{%-     if host.set_tag -%}
+,set:{{ host.set_tag|replace('-', '') }}
+{%-     endif -%}
+{%-     if host.ip -%}
+,
+{%-         set ip_list = host.ip.split(',') -%}
+{%-         for ip in ip_list -%}
+{%-             set ip = ip.strip() -%}
+{{ '[' if ':' in ip }}{{ ip }}{{ ']' if ':' in ip }}{%- if not loop.last %},{% endif %}
+{%-         endfor -%}
+{%-     endif -%}
+{%-     if host.host -%}
+,{{ host.host }}
+{%-     endif -%}
+{%-     if host.lease_time -%}
+,{{ host.lease_time }}{% else %},86400
+{%-     endif -%}
+{%-     if host.ignore|default('0') == '1' -%}
+,ignore
+{%-     endif +%}
 {% endfor %}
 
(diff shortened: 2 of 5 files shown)

### CODE REVIEW

**Comment by gspannu** [comment]
> Fixes: #8487
> 
> Directive: https://thekelleys.org.uk/dnsmasq/docs/dnsmasq-man.html
> 
> This should almost fully satisfy the requirements of the man page regarding this directive.

First of all, thanks a ton for implementing this so quickly.

Quick question: 
Can we safely assume that the implemented `dhcp-host` option for **_constructor style IPv6_** will work as defined in the manpage?

<hr>
_From the manpage.._

A single --dhcp-host may contain an IPv4 address or one or more IPv6 addresses, or both. IPv6 addresses must be bracketed by square brackets thus: --dhcp-host=laptop,[1234::56] **IPv6 addresses may contain only the host-identifier part**: --dhcp-host=laptop,[::56] in which case they act as wildcards in **constructed DHCP ranges**, with the appropriate network part inserted. **For IPv6, an address may include a prefix length**: --dhcp-host=laptop,[1234:50/126] which (in this case) specifies four addresses, 1234::50 to 1234::53. This (an the ability to specify multiple addresses) is useful when a host presents either a consistent name or hardware-ID, but varying DUIDs, since it allows dnsmasq to honour the static address allocation but assign a different adddress for each DUID.

File: src/opnsense/mvc/app/models/OPNsense/Dnsmasq/Dnsmasq.xml [hunk 1]
```diff
@@ -82,8 +82,18 @@
             <ip type="NetworkField">
                 <Required>Y</Required>
                 <NetMaskAllowed>N</NetMaskAllowed>
+                <FieldSeparator>,</FieldSeparator>
+                <AsList>Y</AsList>
             </ip>
-            <hwaddr type="MacAddressField"/>
+            <client_id type="TextField"/>
```
AdSchellevis, line 88: will need a validation

src/opnsense/service/templates/OPNsense/Dnsmasq/dnsmasq.conf
```diff
@@ -186,9 +186,37 @@ ra-param={{ helpers.physical_interface(dhcp_range.interface) }}
 {% endfor %}
 
 {% for host in helpers.toList('dnsmasq.hosts') %}
-{%    if host.hwaddr and host.hwaddr.find(':') == -1%}
-dhcp-host={{host.hwaddr}}{% if host.set_tag%},set:{{host.set_tag|replace('-','')}}{% endif %},{{host.ip}}
-{%    endif %}
+{#     Skip if MAC is missing or invalid (no colon), unless client_id is present #}
```
AdSchellevis, line 189: I need some help with this part, which output are we expecting when a user enters something like `192.168.1.1,192.168.1.2,2000:1234::1`

AdSchellevis (changes requested):

src/opnsense/service/templates/OPNsense/Dnsmasq/dnsmasq.conf
Monviech: Before changing the logic here, it did not render the dhcp-host directive if no MAC address was present.

I think this was before I touched it so that users could create host overrides without adding a dhcp-host at the same time.

Now this has been expanded to also process the dhcp host when either client_id or hwaddr are populated.

So if a user just inputs a string of comma separated IP addresses, without client_id or hwaddr set, just the normal host overrides for each IP address+hostname+fqdns will be generated in the path that the php script populates. Without the dhcp-host directive appearing in the dnsmasq.conf file.

src/opnsense/mvc/app/models/OPNsense/Dnsmasq/Dnsmasq.xml
Monviech: I think I would need help with a validation here, since there are some DUID types.

DUID-UUID
DUID-LLT
DUID-EN
DUID-LL: https://www.rfc-editor.org/rfc/rfc6355.html#section-2

There's also a "normal" DUID?:

https://datatracker.ietf.org/doc/html/rfc3315#section-9

This sounds like a validation nightmare, maybe we can do a very light input validation by using a "DescriptionField"?

Monviech:
@gspannu As ```::56``` is a valid IPv6 address it can be input just fine. Just no prefix allowed right now, gotta input multiple addresses if you want multiple in the pool.

fichtner:
Strictly speaking it’s a reserved/deprecated IPv6 range that doesn’t exist in the real world😉 https://www.iana.org/assignments/ipv6-address-space/ipv6-address-space.xhtml

src/opnsense/mvc/app/models/OPNsense/Dnsmasq/Dnsmasq.xml
gspannu: I agree with _light validation_. Any user who is expecting to use this feature would be expected to be reasonably network savvy to know DUID validations, etc…

AdSchellevis: @Monviech isn't just a hexadecimal string with separators? (e.g. ` --dhcp-host=id:01:02:03:04,.....` for dnsmasq)

Monviech: The man page specifies

```Thus: --dhcp-host=id:01:02:03:04,..... refers to the host with client identifier 01:02:03:04.
It is also allowed to specify the client ID as text, like this: --dhcp-host=id:clientidastext```

And then theres ```--dhcp-host=id:*```  allowed too.

Which means I'll build a regex that allows alphanumerical characters, allow some special characters as separators and a max length.

AdSchellevis: I know what the man page specifies, question is what the desired input is... If I would like to know my hostid, I would ask the client (osx, windows, ..), which probably communicates in a more uniform way than "any"

src/opnsense/service/templates/OPNsense/Dnsmasq/dnsmasq.conf
Monviech: I removed this logic and added a checkbox that clearly states that a DHCP reservation will be created.

Since DHCP reservations do not need a Hardware address or a Client id to be set, this was the most logical and UX friendly choice, to distinguish between simple host override and host override + dhcp-host.

Monviech: I rolled the last changes back as discussed.

src/opnsense/mvc/app/models/OPNsense/Dnsmasq/Dnsmasq.xml
Monviech: Validation added as discussed.

Monviech:
All dhcp-host entries must have only unique IP addresses. Otherwise dnsmasq has a hard error that prevents it to start.

I will add a draft for an additional validation that triggers when:

- An entry has either a ``client_id`` or ``hwaddr`` set, thus creating a dhcp-host
- The IP addresses of all of these entries are compared
- The validation triggers if the whole dhcp-host dataset is not unique IP wise.

dns overrides do not have to be unique, so the validation specifically targets only dhcp-host

AdSchellevis:
@Monviech I've added https://github.com/opnsense/core/pull/8497/commits/9a0f3c528a1656c97067e5c96b54dc99ff9b2de4, if you're ok with it, you can merge it.

Monviech:
@AdSchellevis Just tested it and everything still works. So in it goes. Thank you for your help :)

Issue #8499, opened 2025-03-28
dnsmasq: Change add-mac default to a different value


Monviech:
**Important notices**

Before you add a new report, we ask you kindly to acknowledge the following:

- [x] I have read the contributing guide lines at https://github.com/opnsense/core/blob/master/CONTRIBUTING.md
- [x] I am convinced that my issue is new after having checked both open and closed issues at https://github.com/opnsense/core/issues?q=is%3Aissue

**Is your feature request related to a problem? Please describe.**

make lint does not like that this is called default even though its just an option in an OptionField:

https://github.com/opnsense/core/blob/b2dc6fed7cd1f3b1335b4aa889f6033c5a62fc67/src/opnsense/mvc/app/models/OPNsense/Dnsmasq/Dnsmasq.xml#L38

So changing the name of the option is the easiest fix as this is not released yet.

PR #8500, opened 2025-03-28
dnsmasq: Change add_mac OptionValue from default to standard to satisfy lint

Monviech:
Fixes: https://github.com/opnsense/core/issues/8499


fichtner (approved):
thank you

PR #8515, opened 2025-04-02
Config settings to define the web GUI session cookie name

### LINKED ISSUE #8365: Change PHP session name to be dynamic

Kage:
This PR makes the PHP session cookie name configurable so that OPNsense sessions are not locked to the PHP default `PHPSESSID`.  This allows multiple OPNsense web GUIs to operate on the same hostname or IP without interfering with each other (e.g. multiple OPNsense VMs with an SSH tunnel to the vhost to access them).

The configuration will only permit alphanumeric names.  If a session name setting is not provided, it defaults to `opnsense`.

Fixes #8365

### DIFF
--- a/src/opnsense/service/templates/OPNsense/WebGui/php.ini
+++ b/src/opnsense/service/templates/OPNsense/WebGui/php.ini
@@ -35,3 +35,4 @@ error_log=/tmp/PHP_errors.log
 date.timezone="{{system.timezone|default('Etc/UTC')}}"
 session.save_path=/var/lib/php/sessions
 session.gc_maxlifetime={{system.webgui.session_timeout|default(240)|int * 60}}
+session.name={{system.webgui.session_name|default('opnsense')}}
--- a/src/www/system_advanced_admin.php
+++ b/src/www/system_advanced_admin.php
@@ -124,8 +124,9 @@ if ($_SERVER['REQUEST_METHOD'] === 'GET') {
     }
 
     if (!empty($pconfig['session_name'])) {
-        if (!ctype_alnum($pconfig['session_name'])) {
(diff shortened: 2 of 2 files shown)


darkk:
> This allows multiple OPNsense web GUIs to operate on the same hostname or IP

I would strongly advice against that. Every OPNsense instance will get all the cookies in this case as you suggest to replace `PHPSESSID` with something that is possibly unique for every node.

Imagine one of that nodes getting compromised. It'll allow lateral movement as the attacker will get access to possibly-active sessions on other nodes with the very first `GET /` without any password entry.

If you have multiple OPNsense web GUIs on the same IP, you'd rather use different ports. Speaking of `ssh -L lll:a.b.c.d:rrr` case &mdash; pick an unique `${lll}` port for each instance.

If you have multiple OPNsense web GUIs on the same hostname, I'd suggest to use subdomains & CNAMEs. It'll be more reliable solution.

PR #8516, opened 2025-04-02
dnsmasq: Unify dhcp-option and dhcp-match in the same grid

Monviech:
Both dhcp-option and dhcp-match use almost the same fields in the form and concern the same options. Unifying them could make sense to remove some duplicate code and potentially improving the UX.

@AdSchellevis This is just a suggestion I made in code so the intent is clear. If we do not want this, its okay and we can close it. :)


AdSchellevis (approved):

AdSchellevis:
@Monviech this crossed my mind, but didn't make it to the implementation. Looking at the current state, I think it makes sense to merge these. Thanks!

Monviech:
@AdSchellevis Thanks for the feedback, I'll merge this then :)

PR #8520, opened 2025-04-02
dnsmasq: Implement a light validation for option6 values

Monviech:
As they require to be wrapped inside square brackets if they contain an IPv6 address.

Since the value field gets shared by all options, whatever they might contain, it might not be the best choice to automate this in the template. Maybe this validation is a good choice?

Man page:

```
IPv6 options are specified using the option6: keyword, followed by the option number or option name. The IPv6 option name space is disjoint from the IPv4 option name space. IPv6 addresses in options must be bracketed with square brackets, eg. --dhcp-option=option6:ntp-server,[1234::56].
```


src/opnsense/mvc/app/models/OPNsense/Dnsmasq/Dnsmasq.php
```diff
@@ -250,6 +250,22 @@ public function performValidation($validateFullModel = false)
                     )
                 );
             }
+
+            if (
+                !$option->value->isEmpty() &&
+                !$option->option6->isEmpty() &&
+                preg_match(
+                    '/::|([A-F0-9]{1,4}:){2,}/i',
```
fichtner, line 258: looks a bit too magical for my taste.. are we verifying a link-local address? or a normal IPv6?

Monviech: @fichtner 

Since ``value`` can be literally anything right now, and I dont know if there is e.g., an option that would use something like ``example.com:443`` as value, I went for either at least two ``::`` or at least two hex groups like ``abcd:1234:`` which would prevent that a MAC address would match.

But this is by no means great, this ``value`` field is pretty fragile, and depending on the ``option`` or ``option6`` value that you choose, dnsmasq can always hard fail with errors. Its got nothing to do with IPv6 really.

The issue is broader.

I guess to fix this completely, the value field would probably need to become its own field type, and context sensitive validations added for all ``option`` and ``option6`` types that exist.

So the question here is, is it fine if Kea does not start with the log saying whats wrong, or does this value field require a full validation for every option that exists?

https://github.com/opnsense/core/blob/ef1c4e07c86f5f03f643c89218bacf9e91956243/src/opnsense/mvc/app/models/OPNsense/Dnsmasq/Dnsmasq.xml#L246

fichtner: maybe I'm missing something but if you want to ensure it's an IPv6 address with brackets you need to check for IPv6 address first, then assert if brackets are found and only of no brackets are found you are allowed to emit the validation message.

Right now this code is guessing and I don't beliebe it detects the brackets?

Monviech: Oh right that makes sense.

Monviech: I gave it another shot but I would like somebody else to look at what we can do here, and if we should do anything here. There are quite some edge cases I cannot catch.

src/opnsense/mvc/app/models/OPNsense/Dnsmasq/Dnsmasq.php
```diff
@@ -259,6 +260,28 @@ public function performValidation($validateFullModel = false)
...
+            if (
+                !$option->value->isEmpty() &&
+                !$option->option6->isEmpty()
+            ) {
+                $values = array_map('trim', explode(',', (string)$option->value));
+                foreach ($values as $value) {
+                    $address = trim($value, '[]');
+
+                    if (
+                        Util::isIpv6Address($address) &&
```
fichtner, line 273: I'd just test on $value now and then you know it doesn't have brackets

Monviech: isIPv6Address() returns true:

`[fe80::`

isIPv6Address() returns false:

`[fe80::]`

I think its because of `filter_var()`

https://github.com/opnsense/core/blob/207bddcad5c9854e5dd803c394c7221b65c01bd8/src/opnsense/mvc/app/library/OPNsense/Firewall/Util.php#L72

So if I trim `value` beforehand it works better.

fichtner: that's not very reassuring but makes sense then

Monviech: I logged it again now and that was not the issue:

[03-Apr-2025 15:00:51 Etc/UTC] IPv6 Check: value='fe99::]' => isIpv6=false
[03-Apr-2025 15:00:41 Etc/UTC] IPv6 Check: value='[fe99::' => isIpv6=false
[03-Apr-2025 15:00:46 Etc/UTC] IPv6 Check: value='fe99::' => isIpv6=true

I guess my first logging was flawed in a way.

Monviech:
I'd like to leave this one up to @AdSchellevis , I feel like its still too fragile.

src/opnsense/mvc/app/models/OPNsense/Dnsmasq/Dnsmasq.php
```diff
@@ -259,6 +260,28 @@ public function performValidation($validateFullModel = false)
...
+                        !(strpos($value, '[') === 0 && substr($value, -1) === ']')
```
AdSchellevis, line 274:
```suggestion
                        !(str_starts_with($value, '[') && str_ends_with($value, ']'))
```

src/opnsense/mvc/app/models/OPNsense/Dnsmasq/Dnsmasq.php
AdSchellevis, line 273:
```suggestion
                        Util::isIpv6Address(trim($value, '[]')) &&
```

src/opnsense/mvc/app/models/OPNsense/Dnsmasq/Dnsmasq.php
```diff
@@ -259,6 +260,28 @@ public function performValidation($validateFullModel = false)
...
                 );
             }
+
+            if (
+                !$option->value->isEmpty() &&
+                !$option->option6->isEmpty()
+            ) {
+                $values = array_map('trim', explode(',', (string)$option->value));
+                foreach ($values as $value) {
+                    $address = trim($value, '[]');
```
AdSchellevis, line 270: since we're not using `$address` anywhere else, I would combine the statement.

AdSchellevis (changes requested):
some small remarks, but further ok.

src/opnsense/mvc/app/models/OPNsense/Dnsmasq/Dnsmasq.php
```diff
@@ -259,6 +260,26 @@ public function performValidation($validateFullModel = false)
...
             }
+
+            if (
+                !$option->value->isEmpty() &&
+                !$option->option6->isEmpty()
+            ) {
+                $values = array_map('trim', explode(',', (string)$option->value));
+                foreach ($values as $value) {
+                    if (
+                        Util::isIpv6Address(trim($value, '[]')) &&
```
fichtner, line 271: [[fe80::1]] probably doesn't validate so that's why I suggested to avoid the trim and just check for IPv6 ;)

PR #8524, opened 2025-04-03
TrafficShaper: Implement base_bootgrid_table and base_apply_button

Monviech:


fichtner (approved):
looks reasonable. I usually do a final test after merge so let's go!

PR #8526, opened 2025-04-04
dashboard: Certificates.js adjustments for f90e544

Monviech:
A few minor adjustments for: https://github.com/opnsense/core/commit/f90e5445db69de5177f97048e440400a1393b424

The current hidden certificate configuration will be invalidated due to this change as the key/value pair was description, an oversight by my earlier me.

### DIFF
--- a/src/opnsense/www/js/widgets/Certificates.js
+++ b/src/opnsense/www/js/widgets/Certificates.js
@@ -1,5 +1,5 @@
 /*
- * Copyright (C) 2024 Cedrik Pischem
+ * Copyright (C) 2024-2025 Cedrik Pischem
  * All rights reserved.
  *
  * Redistribution and use in source and binary forms, with or without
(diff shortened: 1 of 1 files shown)


src/opnsense/www/js/widgets/Certificates.js
```diff
@@ -161,13 +173,13 @@ export default class Certificates extends BaseTableWidget {
 
         if (caResponse.rows) {
             caResponse.rows.forEach(ca => {
-                hiddenItemOptions.push({ value: `${ca.descr}`, label: ca.descr });
+                hiddenItemOptions.push({ value: `${ca.uuid}`, label: ca.descr });
```
swhite2, line 176: This probably doesn't need string formatting, rather `ca.uuid` directly, be sure to test that

src/opnsense/www/js/widgets/Certificates.js
```diff
@@ -161,13 +173,13 @@ export default class Certificates extends BaseTableWidget {
...
             });
         }
 
         if (certResponse.rows) {
             certResponse.rows.forEach(cert => {
-                hiddenItemOptions.push({ value: `${cert.descr}`, label: cert.descr });
+                hiddenItemOptions.push({ value: `${cert.uuid}`, label: cert.commonname || cert.descr });
```
swhite2, line 182: same here

src/opnsense/www/js/widgets/Certificates.js
```diff
@@ -73,7 +73,7 @@ export default class Certificates extends BaseTableWidget {
 
     processItems(items, type, hiddenItems, rows) {
         items.forEach(item => {
-            if (!hiddenItems.includes(item.descr)) {
+            if (!hiddenItems.includes(item.uuid)) {
```
swhite2, line 76: hiddenItems is set elsewhere in the code as `config.hiddenItems || []`. Best use `getWidgetConfig().hiddenItems` instead so the default is set based on what's returned by `setWidgetConfig()`. In this case the result is still an empty array, but at least it adheres to the proper logic.

swhite2 (changes requested):
Minor things, otherwise looks good to me :)

src/opnsense/www/js/widgets/Certificates.js
Monviech: Yup works, good eyes thanks. :)

swhite2 (approved):

Issue #7193, opened 2024-02-01
Change default domain from OPNsense.local to OPNsense.internal


vpx23:
**Important notices**

Before you add a new report, we ask you kindly to acknowledge the following:

- [x] I have read the contributing guide lines at https://github.com/opnsense/core/blob/master/CONTRIBUTING.md
- [x] I am convinced that my issue is new after having checked both open and closed issues at https://github.com/opnsense/core/issues?q=is%3Aissue

Please consider changing the default domain for new installations to OPNsense.internal following the IANA proposition.

https://itp.cdn.icann.org/en/files/root-system/identification-tld-private-use-24-01-2024-en.pdf

I know the decision for the domain is not final by the IANA but maybe you can already do preparations or identify possible problems with this change.

fichtner:
For future reference this should be a single line pull request:

https://github.com/opnsense/core/blob/5220c4a1a684adbf8b6b74431a3a8693f685478a/src/etc/config.xml.sample#L207

But I suggest waiting for this to be actually a standard or at least gain a wider recognition first.


Cheers,
Franco

PR #5898, opened 2022-07-23
Change system domain name suggestions

da2x:
* Don't suggest using unreserved top-level domains.
* Use the reserved example.com|net domains for other examples.
* Encourage adoption of the special-purpose internal/site home.arpa (RFC 8375) domain.

PR #8535, opened 2025-04-09
Change default DHCP domain to `internal`

Self-Hosting-Group:
as current domain is blocked by DNSSEC validating clients

This PR changes the domain, updates the help and also adds it to the new setup wizard. And don't just suggest the name as in #5898, update the default as previously suggested in #7193.

systemd-resolved has `internal` on its DNSSEC negative trust anchor list since 2016, where `home.arpa` was added later:
https://github.com/systemd/systemd/commit/30c778094b90a637c6691c462a66df81eeb865b5#diff-23206c93f79c419c8a911a58d533e3ca9a6103ff215a5fcc887aaecb59021276R155

**.internal history**

Internal was first mentioned in RFC 6762 (2013) within multicast DNS:
https://datatracker.ietf.org/doc/html/rfc6762#appendix-G

IANA assessment (2024-01):
https://itp.cdn.icann.org/en/files/root-system/identification-tld-private-use-24-01-2024-en.pdf

ICANN board resolution (2024-07):
https://www.icann.org/en/board-activities-and-meetings/materials/approved-resolutions-special-meeting-of-the-icann-board-29-07-2024-en#section2.a

IETF RFC draft (2025-03):
https://datatracker.ietf.org/doc/html/draft-davies-internal-tld
According to the last IETF 122 meeting there is `Almost no work` to do:
https://datatracker.ietf.org/doc/minutes-122-dnsop/

**.home.arpa (former aproch/name)**
https://datatracker.ietf.org/doc/html/rfc8375

### DIFF
--- a/src/opnsense/mvc/app/controllers/OPNsense/Core/forms/wizard_general_info.xml
+++ b/src/opnsense/mvc/app/controllers/OPNsense/Core/forms/wizard_general_info.xml
@@ -8,6 +8,7 @@
         <id>wizard.domain</id>
         <label>Domain</label>
         <type>text</type>
+        <help>Do not use 'local' as your internal domain name. It is reserved for and will interfere with mDNS (avahi, bonjour, etc.). Instead, use the special-use domain internal or home.arpa, also to avoid DNSSEC validation issues. E.g. internal, name.internal.</help>
     </field>
     <field>
         <id>wizard.language</id>
--- a/src/etc/config.xml.sample
+++ b/src/etc/config.xml.sample
@@ -6,7 +6,7 @@
   <system>
     <optimization>normal</optimization>
     <hostname>OPNsense</hostname>
-    <domain>localdomain</domain>
+    <domain>internal</domain>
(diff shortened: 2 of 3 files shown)


src/opnsense/mvc/app/controllers/OPNsense/Core/forms/wizard_general_info.xml
```diff
@@ -8,6 +8,7 @@
         <id>wizard.domain</id>
         <label>Domain</label>
         <type>text</type>
+        <help><![CDATA[Do not use 'local' as your internal domain name. It is reserved for and will interfere with mDNS (avahi, bonjour, etc.). Use the special-purpose internal or home.arpa domain instead.<br>e.g. <code>internal</code>, <code>name.internal</code>, <code>example.net</code>, <code>branch.example.com</code>, etc.]]</help>
```
fichtner, line 11: no markup please. it will just end up not being translated ever.

fichtner:
Merged, thanks!

PR #8542, opened 2025-04-13
Added Button to select all in policy edit modal - policy.volt

vzeller:
Added Button to select all in policy edit modal - policy.volt 
So users can quickly select all options and deselect what they do not want. 
Saves a lot of time when creating custom policies.

### DIFF
--- a/src/opnsense/mvc/app/views/OPNsense/IDS/policy.volt
+++ b/src/opnsense/mvc/app/views/OPNsense/IDS/policy.volt
@@ -32,6 +32,41 @@ POSSIBILITY OF SUCH DAMAGE.
        * update service status
        */
       updateServiceControlUI('ids');
+      // Add Select All Options button to DialogPolicy modal when it's shown
+      $(document).on('shown.bs.modal', '#DialogPolicy', function() {
+          // Check if button already exists to avoid duplicates
+          if ($('#selectAllOptions').length === 0) {
+              // Add button at the top of the form
+              $('#DialogPolicy .modal-body').prepend('<button type="button" class="btn btn-primary" id="selectAllOptions" style="margin-bottom: 10px;">Select All Options</button>');
+              
+              // Add click event handler
+              $('#selectAllOptions').on('click', function() {
+                  // Check and enable all checkboxes
+                  $('input[type="checkbox"]').each(function() {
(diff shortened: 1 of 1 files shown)


AdSchellevis:
@vzeller do you have a concrete example where this would save time? I haven't seen many cases where you would like to have all options except some.

vzeller:
> @vzeller do you have a concrete example where this would save time? I haven't seen many cases where you would like to have all options except some.

Not sure if im using Policys the intended or right way. However, i have added Snort rules via oink code and ET Pro via the Telemetry Plugin and enabled all rules. To make the IDS use the rules work, except some things i want to have passed. It is my workflow to add a Policy that has everything enabled. And change all Alerts to be Dropped. Then i deselect the the category i want to have passed.
Before i came up with the Javascript i would select all categories CVE's and Threat Groups .. everything manually. Which takes a lot of time if you have many entries populated from Snort, ET Pro and maybe others. 
Makes sense?

vzeller:
Also, with every update ( i have daily updates for IDS rules via cron job). It would change names and add entries in First-Seen or CVE's etc. So i have to update the policy frequently.

vzeller:
Privacy enhancing techniques used in a surveilled network / surveillance state.

AdSchellevis:
I'm only using ET's rules, but in practice you really only want to filter either full sets (files) or things that are clearly specified (and have limited options), such as deployment type. Micromanaging on things that change daily is usually not a great strategy in my humble opinion.

vzeller:
Des, ideally it would add the new IDS rules enabled. Vor give the Option to add them either disabled or enabled, keeping previous selections. Until that, this is a step in the right direction.

AdSchellevis:
Maybe it's better to start with a ticket explaining the issue you want to solve, either I don't understand the use-case (which is perfectly possible) or this option doesn't add much value and we're trying to fix the wrong thing here. 

These metadata selections are intended to match more or less static things like type of attacks (from a limited list) or deployments. When trying to more or less specify single sids by there metadata, this seems overly complicated.

vzeller:
If the code is not harmfull and adds functionality, why argue against if?

vzeller:
Well, after all I'm just a user. And this helps me a lot, so I wanted to share it 💞

Issue #8556, opened 2025-04-15
Expose boot time via API


eopo:
**Important notices**

Before you add a new report, we ask you kindly to acknowledge the following:

- [X] I have read the contributing guide lines at https://github.com/opnsense/core/blob/master/CONTRIBUTING.md
- [X] I am convinced that my issue is new after having checked both open and closed issues at https://github.com/opnsense/core/issues?q=is%3Aissue

**Is your feature request related to a problem? Please describe.**
For statistics, I read the API Endpoint System/SystemTime. I want to track the system's boot time in third party systems.
The endpoint does not return the boot time, but the uptime as formatted string with second precision. It is not possible to reliably calculate the system's boot time from that, depending on the exact timing of requests the result may vary by a second, which leads to unwanted side effects on my side.

**Describe the solution you like**
A field boottime (or other naming) that returns the systems boot time similar to System Time. It could be included in the same route, therefore only needing minimal changes and be non-breaking.

**Describe alternatives you considered**

PR #8557, opened 2025-04-15
Add field boottime to api/system/systemTime

eopo:
This PR implements the feature request in #8556
Additionally to the calculated and formatted uptime, it exposes the systems boot time as Date-Time-String alongside the Current System time. This makes it easier for API consumers to work with the value, instead of having to try to calculate the boot time from the uptime.


AdSchellevis:
@eopo thanks

alexdelprete:
> @eopo thanks

can you please tell us in what version this will be available?

thanks a lot.

AdSchellevis:
not yet planned, but a future minor

alexdelprete:
> not yet planned, but a future minor

thank you. is the plan published somewhere so we can check every once in a while?

AdSchellevis:
usually our changelogs

Issue #8528, opened 2025-04-06
Interface tracking and separation of concerns (discussion)


AdSchellevis:
**Important notices**

Before you add a new report, we ask you kindly to acknowledge the following:

- [x] I have read the contributing guide lines at https://github.com/opnsense/core/blob/master/CONTRIBUTING.md
- [x] I am convinced that my issue is new after having checked both open and closed issues at https://github.com/opnsense/core/issues?q=is%3Aissue

**Is your feature request related to a problem? Please describe.**

As a starting point to isolate `dhcp` and `router advertisements` on ipv6 for client (lan) networks, we likely need a clearer separation of concerns. Certainly if we want `dnsmasq` to offer (part of) the functionality.

In order to understand the role of "tracking" we need to follow the execution flow when configuring these interfaces and their parents.

When a "lan" type interface has configured ipv6 to "track", the following path is traversed:

* `interface_configure()` is called to configure the interface { or `link_interface_to_track6()` via events, `interface_reset()` when reset} 
    * `interface_track6_configure()` then evaluates the parent, depending on type, the following happens:
        * [6to4] —> `interface_track6_6to4_configure()`
            * calculates an address for the (lan) interface and applies it (`/sbin/ifconfig {$lanif} inet6 {$sixto4lan} prefixlen 64`)
        * [6rd] —> `interface_track6_6rd_configure`
            * calculates an address for the (lan) interface and applies it (`/sbin/ifconfig {$lanif} inet6 {$rd6lan} prefixlen 64`)
        * [slaac]
            * enable accepting ICMPv6 Router Advertisement messages (`/sbin/ifconfig {$lanif} inet6 accept_rtadv`)
        * [slaac, dhcp6] —> `interface_dhcpv6_prepare($lancfg['track6-interface'], $trackcfg);`  —> create /var/etc/rtsold_script.sh (no exec) for the parent interface

From the parents perspective (wan), the tracking "children" are relevant to generate a dhcp client config when auto configured as such:

* `interface_configure()` is called to configure the interface, relevant types are:
    * [slaac, dhcp6] 
         * accept router advertisements (`/sbin/ifconfig {$wanif} inet6 accept_rtadv -ifdisabled up`) 
         * `interface_dhcpv6_prepare()` uses  `sla-id`/`sla-len` to generate dhcp client config, which depends on:
             * `link_interface_to_track6()` to collect its “children” 
[...]

PR #8576, opened 2025-04-23
Services: ISC DHCPv6 - show "tracking" interfaces when enabled and offer an explicit disable

AdSchellevis:
Needed in case someone would like to use dnsmasq or kea instead.

To avoid large changes, we opt for a  minimal set here. In services_dhcpv6.php, we add a separate form and handler in case tracking (without dhcpd6track6allowoverride) is set, which either flushes the unused isc-dhcpv6 server configuration when enabled (default) or writes a small section only including ['enabled' => -1]. For visibility, we show the calculated range as would be set by dhcpd_dhcp6_configure() when tracking is used.

The backend code then double checks the services which er explicitly disabled (-1) and skip processing for these (not enabled).

In order to make people aware of the fact that an isc-dhcpv6 server could be running, make sure the menu system also reflects reality.

Since router advertisements are stored within the same container and will need a toggle as well, keep the value of ramode so we have a way to intervene in a similar way as for dhcpv6.
One small side affect of this commit is that it will show "Services: Router Advertisements" for the tracking interface, which we need to implement later.

One of the building blocks for: https://github.com/opnsense/core/issues/8528

### DIFF
--- a/src/www/services_dhcpv6.php
+++ b/src/www/services_dhcpv6.php
@@ -1,7 +1,7 @@
 <?php
 
 /*
- * Copyright (C) 2014-2022 Deciso B.V.
+ * Copyright (C) 2014-2025 Deciso B.V.
  * Copyright (C) 2003-2004 Manuel Kasper <mk@neon1.net>
  * Copyright (C) 2010 Seth Mos <seth.mos@dds.nl>
  * All rights reserved.
@@ -58,7 +58,7 @@ function show_track6_form($if)
       $range['to'] = Net_IPv6::compress(implode(':', $ifcfgipv6arr));
   }
 
-  if (!empty($config['dhcpdv6']) && !empty($config['dhcpdv6']['lan']) && isset ($config['dhcpdv6']['lan']['enable']) && $config['dhcpdv6']['lan']['enable'] == '-1') {
+  if (!empty($config['dhcpdv6']) && !empty($config['dhcpdv6'][$if]) && isset($config['dhcpdv6'][$if]['enable']) && $config['dhcpdv6'][$if]['enable'] == '-1') {
       /* disabled */
       $enable_input = '';
   }
--- a/src/www/services_dhcpv6.php
+++ b/src/www/services_dhcpv6.php
@@ -174,7 +174,7 @@ if ($if === null) {
     } elseif ($_SERVER['REQUEST_METHOD'] === 'POST') {
         process_track6_form($if);
     }
-    die;
+    exit;
 }
 /* default form processing */
 
--- a/src/www/services_dhcpv6.php
+++ b/src/www/services_dhcpv6.php
@@ -34,6 +34,104 @@ require_once("system.inc");
 require_once("interfaces.inc");
 require_once("plugins.inc.d/dhcpd.inc");
 
+
+function show_track6_form($if)
+{
+  global $config;
+  $service_hook = 'dhcpd6';
+  include("head.inc");
+  include("fbegin.inc");
+  $enable_label = gettext("Enable");
+  $range_label = gettext("Range");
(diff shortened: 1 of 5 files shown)


src/www/services_dhcpv6.php
```diff
@@ -34,24 +34,120 @@
...
 if ($_SERVER['REQUEST_METHOD'] === 'GET') {
     // handle identifiers and action
     if (!empty($_GET['if']) && !empty($config['interfaces'][$_GET['if']]) &&
         isset($config['interfaces'][$_GET['if']]['enable']) &&
         (is_ipaddr($config['interfaces'][$_GET['if']]['ipaddrv6']) ||
-        !empty($config['interfaces'][$_GET['if']]['dhcpd6track6allowoverride']))) {
+        !empty($config['interfaces'][$_GET['if']]['track6-interface']))) {
         $if = $_GET['if'];
-    } else {
-        /* if no interface is provided this invoke is invalid */
```
fichtner, line 52: wouldn't remove this at first glance, the page shows garbage when invoked this way?

src/www/services_dhcpv6.php
```diff
@@ -60,11 +156,28 @@ function reconfigure_dhcpd()
...
 }
 
+/**
+ * XXX: In case of tracking, show different form and only handle on/off options.
+ *      this code injection is intended to keep changes as minimal as possible and avoid regressions on existing isc-dhcp6 installs,
+ *      while showing current state for tracking interfaces.
+ **/
+if ($if === null) {
+    /* if no interface is provided this invoke is invalid */
+    header(url_safe('Location: /index.php'));
```
fichtner, line 169: ah ok you moved this here

src/www/services_dhcpv6.php
```diff
@@ -60,11 +156,28 @@ function reconfigure_dhcpd()
...
+    exit;
+} elseif (!empty($config['interfaces'][$if]['track6-interface']) && !isset($config['interfaces'][$if]['dhcpd6track6allowoverride'])) {
+    if ($_SERVER['REQUEST_METHOD'] === 'GET') {
+        show_track6_form($if);
+    } elseif ($_SERVER['REQUEST_METHOD'] === 'POST') {
+        process_track6_form($if);
+    }
+    die;
```
fichtner, line 177:
```suggestion
    exit;
```

I would reserve "die" for actual errors or aborts

src/www/services_dhcpv6.php
```diff
@@ -34,24 +34,120 @@
...
+function process_track6_form($if)
+{
+    $dhcpdv6cfg = &config_read_array('dhcpdv6');
+    $this_server = [];
+    if (isset($dhcpdv6cfg[$if]) && isset($dhcpdv6cfg[$if]['ramode'])) {
+        /* keep ramode for router advertisements so we can use this field to disable the service when in tracking mode */
+        $this_server['ramode'] = $dhcpdv6cfg[$if]['ramode'];
+    }
+    if (empty($_POST['enable'])) {
+        $this_server['enable'] = '-1';
```
fichtner, line 125: the downside of this is when the interface is switched to static the value will be '' or '0' again and it re-enabled the auto-mode which potentially leads to more code and all the 'enable' use flags need to be checked/adjusted too empty() doesn't work anymore. Maybe you have done it, i just want to mention it here.

AdSchellevis: I don't think the interfaces page mangles with `dhcpv6` section, the only reference I could find was this one :

https://github.com/opnsense/core/blob/39d5ff1a26ae306abc83c9ae1c70c8c4c07ce997/src/www/interfaces.php#L624-L626


which is merely a validation (and might need a change, but haven't looked at that yet)

fichtner: yeah stuff like that, just to keep in mind

fichtner (approved):

fichtner (approved):
ok, let's just see how this behaves on master

PR #8581, opened 2025-04-25
Add ability to specify pipe and queue masking using the *-ip6 specifiers.

tangrs:
Currently, dynamic pipe and queues based on masking don't work (well) for IPv6. IPv6 flows appear to all end up bucketed into single `::/0` flow. This isn't particularly desirable when attempting to use the traffic shaper to implement the "share bandwidth equally among all users" use-case, because there's no way to distinguish between IPv6 flows.

From experimentation, it seems like configuring a mask using `*-ip` specifiers always buckets IPv6 flows into a single bucket, and vice versa for `*-ip6` specifiers.

This patch allows the user to decide which mask specifiers to use. As far as I can tell, this change should be backwards compatible with existing configuration.


AdSchellevis:
@tangrs thanks! ipv6 needs different qualifiers indeed. Although I haven't tried the ipv6 part, this commit shouldn't influence existing behavior.

Issue #8266, opened 2025-01-30
ISC DHCPv4 static mapping with domain starting with dot breaks Unbound DNS


dseven:
See also: https://forum.opnsense.org/index.php?topic=45457.0

Reproducible in v25.1. Unknown if this was ever not an issue.

The OPNsense Web UI allows addition of an ISC DHCPv4 static mapping with a domain name starting with a dot ("."). The resulting Unbound DNS config (assuming "Register DHCP Static Mappings" is enabled) looks like:

```
local-data-ptr: "192.168.1.254 fakehostname..fakedomain"
local-data: "fakehostname..fakedomain IN A 192.168.1.254"
```

which results in Unbound failing to start:

```
<27>1 2025-01-30T16:00:08+00:00 opntest1.localdomain unbound 8416 - [meta sequenceId="24"] [8416:0] error: error parsing local-data at 15 'fakehostname..fakedomain IN A 192.168.1.254': Empty label
<27>1 2025-01-30T16:00:08+00:00 opntest1.localdomain unbound 8416 - [meta sequenceId="25"] [8416:0] error: Bad local-data RR fakehostname..fakedomain IN A 192.168.1.254
<26>1 2025-01-30T16:00:08+00:00 opntest1.localdomain unbound 8416 - [meta sequenceId="26"] [8416:0] fatal error: Could not set up local zones
```

**To Reproduce**

1. Enable "Register DHCP Static Mappings" for Unbound DNS
2. Create an ISC DHCPv4 static mapping (for any DHCP-enabled interface) with a fake MAC address (e.g. "aa:bb:cc:dd:ee:ff"), hostname "fakehostname" and domain ".fakedomain"
3. (try to) restart the Unbound DNS service (it will fail)
4. Observe error in resolver log

Note: It appears that neither ISC DHCPv6 static mappings nor Kea reservations allow specification of a domain name, so this issue seems to pertain to ISC DHCPv4 only.

**Expected behavior**

Perhaps OPNsense should strip any leading dots from domain names when combining them with hostnames to produce DNS records, which would allow Unbound to start.

If that's not feasible, the UI should refuse to create a static mapping where the domain starts with a dot.

PR #8588, opened 2025-04-26
unbound: fix invalid FQDN generation from DHCPv4 static map domains

SZim92:
Hi OPNsense Team,

This PR resolves Issue #8266, addressing an Unbound startup failure when "Register DHCP Static Mappings" is enabled and an ISC DHCPv4 static mapping specifies a domain name beginning with a dot.

---

## Problem

The `unbound_add_host_entries` function reads domain values directly from ISC DHCPv4 static mappings in `config.xml` without validation.  
When the domain field begins with a dot (e.g., `.domain.local`), concatenation during FQDN generation (`hostname`.`domain`) produces invalid entries like `hostname..domain.local`.  
This results in invalid `local-data` records written to `/var/unbound/host_entries.conf`, causing Unbound to fail parsing and preventing service startup.

---

## Solution

This patch introduces targeted sanitation in `unbound_add_host_entries`:
- Leading dots are stripped from the domain field using `ltrim()`.
- Each correction is logged via `syslog(LOG_WARNING)` for administrative traceability.
- The cleaned domain is used for all subsequent FQDN generation.

---

## Scope and Risk

- Change is limited to Unbound configuration generation.
- Does not affect DHCP runtime services or static mapping storage.
- No configuration migration required.
- Fully backward compatible.
- Matches coding and logging conventions within `unbound.inc`.

---

## Validation Notes

This patch addresses the traced failure path identified in #8266 with a targeted, code-level correction.  
The change is isolated to domain string sanitation during Unbound configuration generation and does not impact DHCP runtime behavior or core system services.

This patch has been fully reviewed at the code level for scope and safety.  
Would greatly appreciate assistance from someone familiar with the OPNsense build process to help validate runtime behavior and move this over the line.  
Appreciate your support!

---

Fixes: #8266

Thanks for your time and review.

### DIFF
--- a/src/etc/inc/plugins.inc.d/unbound.inc
+++ b/src/etc/inc/plugins.inc.d/unbound.inc
@@ -604,6 +604,22 @@ function unbound_add_host_entries($ifconfig_details)
                     /* cannot register without a hostname */
                     continue;
                 }
+
+                if (isset($host['domain'])) {
+                    $original_domain = $host['domain'];
+                    if (!empty($original_domain) && is_string($original_domain) && strpos($original_domain, '.') === 0) {
+                        // Strip all leading dots
+                        $sanitized_domain = ltrim($original_domain, '.');
+                        syslog(LOG_WARNING, sprintf(
+                            "Unbound: Sanitized invalid domain '%s' to '%s' for ISC DHCPv4 static map host '%s'",
+                            $original_domain,
+                            $sanitized_domain,
+                            $host['hostname']
+                        ));
+                        $host['domain'] = $sanitized_domain;
+                    }
+                }
+
                 if (empty($host['domain'])) {
                     $host['domain'] = $config['system']['domain'];
                 }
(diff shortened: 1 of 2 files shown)


AdSchellevis:
Although I'm not a huge fan of trying to keep isc alive much longer, if we do want to fix it's validation, we should probably fix it where it originated from, which is somewhere around :

https://github.com/opnsense/core/blob/3c2ad5d6b3c53f27d747bbde7b6ff81dc4bb7e5f/src/www/services_dhcp_edit.php#L106-L116

SZim92:
Good call. Added validation in services_dhcp_edit.php to block leading-dot domains at save time.

Thanks for pointing it out. Happy to tweak if needed!

SZim92:
Thanks, and fully agreed. I’ve already removed the runtime sanitation from Unbound in the latest commits after adding the input validation.

Appreciate the guidance. Should be clean now!

AdSchellevis:
@SZim92 looks good, thanks!

PR #8404, opened 2025-03-05
shaper: move pipe & queue configuration to dnctl service

swhite2:
if no shaper (`ipfw`) rules are present, or these rules are disabled, `ipfw` will be disabled as well (`firewall_enable="NO"` and `rc.ipfw onestop`).

Traffic shaped via `pf` will not show up in the stats output of `dnctl pipe|queue|sched show`. Also, there is currently no logic to associate pipes/queues with `pf` rules

PR #8589, opened 2025-04-26
Feature: Add Alias Support for Traffic Shaper Rules (host/network types)

SZim92:
Add backend support for using Firewall Aliases in Traffic Shaper (ipfw) rules, allowing source and destination fields to reference Aliases instead of literal IPs or CIDRs.

---

## Problem

Traffic Shaper rules previously required manual IP/CIDR inputs, duplicating firewall configurations and limiting dynamic management when aliases were already available.

---

## Solution

- Added `Alias::resolveAliasToNetworks()` in PHP to expand aliases recursively into IPs/CIDRs.
- Added configd action `alias resolve_networks` to expose alias resolution through configctl.
- Added Python helper `resolve_shaper_address()` to bridge configctl calls inside templates.
- Updated `ipfw.conf` template:
  - Dynamically expands aliases into multiple `ipfw` rules.
  - Uses robust sequence numbering (base 61000 + rule offset) to avoid conflicts.
- Warnings are logged for circular references, missing aliases, and expansion limit violations.

No runtime behavior changes outside Traffic Shaper config generation.
No configuration migration required.

---

## Current Scope

- Backend code infrastructure only. No runtime validation performed yet.
- Alias types fully supported: `host` and `network`.
- Alias types recognized but skipped: `urltable`, `geoip`, and `external` (warnings logged).
- Frontend validation for alias syntax is deferred to a follow-up PR.

### DIFF
--- a/src/etc/inc/plugins.inc.d/unbound.inc
+++ b/src/etc/inc/plugins.inc.d/unbound.inc
@@ -604,6 +604,22 @@ function unbound_add_host_entries($ifconfig_details)
                     /* cannot register without a hostname */
                     continue;
                 }
+
+                if (isset($host['domain'])) {
+                    $original_domain = $host['domain'];
+                    if (!empty($original_domain) && is_string($original_domain) && strpos($original_domain, '.') === 0) {
+                        // Strip all leading dots
+                        $sanitized_domain = ltrim($original_domain, '.');
+                        syslog(LOG_WARNING, sprintf(
+                            "Unbound: Sanitized invalid domain '%s' to '%s' for ISC DHCPv4 static map host '%s'",
+                            $original_domain,
+                            $sanitized_domain,
+                            $host['hostname']
+                        ));
+                        $host['domain'] = $sanitized_domain;
+                    }
+                }
+
                 if (empty($host['domain'])) {
                     $host['domain'] = $config['system']['domain'];
                 }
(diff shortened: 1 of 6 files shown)


AdSchellevis:
@SZim92 we thought about similar constructs in the past, but they always have downsides. As of 25.1.x it's also not really needed anymore as you can use pipes and queues in firewall rules now (https://github.com/opnsense/core/commit/ed7a90194cc9726ed5a1cc1ab9b48d5f5b53400b).

SZim92:
Hi @AdSchellevis,

Thanks, that makes sense. Inline shaping through firewall rules is absolutely the right long-term direction. Appreciate you flagging the commit.

This PR (#8589) was scoped purely to close a small usability gap for users still relying on the legacy Traffic Shaper, since that path remains active and exposed in the UI today.

No intention to expand the legacy model. Just a targeted patch to align behavior and ease the transition while the inline shaping model matures.

If it's helpful to legacy users in the interim, happy to keep this available as a contained fix. If not, happy to close to stay fully aligned with the project direction.

Thanks again for the feedback and for pushing the platform forward!

AdSchellevis:
@SZim92 the separate rule option will stay for some time in the product as we can't automatically migrate the rules to pf ones. What we did change recently is the move to `dnctl` (https://github.com/opnsense/core/pull/8404) for the pipe and queue configuration, which skips `ipfw` when not used (less overhead)

SZim92:
Understood. Thanks for the additional context and for explaining the direction around `dnctl` and future shaping integration.

Closing this PR to keep things clean and aligned with the project's focus. Appreciate the feedback and the opportunity to contribute!

Issue #6776, opened 2023-08-24
Interface:"Block private networks" and "Block bogon networks" does not work properly with ipv6.


IsaacFL:
**Important notices**

Before you add a new report, we ask you kindly to acknowledge the following:

- [X ] I have read the contributing guide lines at https://github.com/opnsense/core/blob/master/CONTRIBUTING.md
- [X ] I am convinced that my issue is new after having checked both open and closed issues at https://github.com/opnsense/core/issues?q=is%3Aissue

**Describe the bug**

Blocking bogons blocks ipv6 private ULA addresses even if "Block private networks" is not checked. This is not consistent with behavior of ipv4 block bogons.

The problem is in the script located at:

/usr/local/opnsense/scripts/filter/update_bogons.sh
Line 77         egrep -iv "^fc00::/7" ${WORKDIR}/fullbogons-ipv6.txt > ${DESTDIR}/bogonsv6

The script is trying to remove the ULA address from the downloaded bogon list, like it does for ipv4 bogons, But this will not work for ipv6 as the source bogon list has a single entry of 8000::/1.  This entry blocks the entire upper half of the address space (including ULA), instead of individual ranges like fc00::/7.

So effectively the ipv6 Private ULA address space is always blocked with bogons and is not consistent with ipv4.

Instead of grepping for fc00::7, I would recommend just adding "!fd00::/8" to the end of the bogon list. The lower half, fc00::/8, is currently not defined so it truly is a bogon.

This issue is primarily about bringing consistency between ipv4 and ipv6 in the ULA space, but I would also recommend adding the link local address "!fe80::/64" as an exception.  I don't see how the link local should be treated as a bogon, when it is allowed router to router and causes a lot of probably not expected blocks if you block bogons. The rest of the fe80::/10 is a bogon, since it is reserved but undefined.

**To Reproduce**

If you enable block bogons, but not block private on an interface with logging on, you will see the using ULA is getting blocked by block bogons rule.

**Expected behavior**

You should be able to block bogons, but still use ipv6 private addresses unless Block Private is also enabled. Just as in ipv4.

**Describe alternatives you considered**

Currently I create my own alias, that is just the bogonsv6 with added !!fd00::/8, !fe80::/64 and create floating rule instead of using block bogons on the interface.  This does work, but you now have duplicated a very large table, so is not optimal solution.

**Screenshots**

NA

**Relevant log files**

NA

**Additional context**

NA

**Environment**

Software version used and hardware type if relevant, e.g.:

[...]

PR #8591, opened 2025-04-26
Interfaces / firewall: Improve private and bogon network filters

### LINKED ISSUE #8587: Disabling 'Block private networks' doesn't unblock ULAs
**Important notices**

**Describe the bug**

Disabling 'Block private networks' unblocks RFC1918 source addresses, but not ULAs. While the `block drop in [...] from fc00::/7` rule _does_ get removed, this has no effect since the IPv6 bogons contain 8000::/1.

So to unblock ULAs, you have to disable _both_ "Block private networks" _as well as_ "Block bogon networks". This is not intuitive.

For IPv4, private addresses are explicitly excluded from the bogons:
https://github.com/opnsense/core/blob/3c2ad5d6b3c53f27d747bbde7b6ff81dc4bb7e5f/src/opnsense/scripts/firmware/bogons.sh#L59-L61

For IPv6, there is no such exception. This inconsistency leads to unexpected behaviour.

**To Reproduce**

1. Go to 'Interfaces: [WAN]'
2. Disable 'Block private networks', save & apply
3. Go to 'Firewall: Log Files: Live View'
4. See error: ULAs get blocked by the 'Block bogon IPv6 networks' rule

**Expected behavior**

Disabling 'Block private networks' should allow ULAs without also having to disable "Block bogon networks".

Open to discussion. If we want to keep the current behaviour, it should at least be explained in the help text in the GUI. This currently doesn't mention IPv6 at all.

**Environment**

OPNsense 25.1.5_5-amd64

maurice-w:
Disabling 'Block private networks' now allows link-local addresses and ULAs without having to disable 'Block bogon networks'.

Draft, open for comments. If accepted, I'll update the help texts in `interfaces.php`.

Tests performed:
'Block private networks' enabled, 'Block bogon networks' enabled or disabled: Link-locals and ULAs are blocked.
'Block private networks' disabled, 'Block bogon networks' enabled or disabled: Link-locals and ULAs pass.

Inspired by #6776. Closes #8587.

### DIFF
--- a/src/etc/inc/filter.lib.inc
+++ b/src/etc/inc/filter.lib.inc
@@ -302,7 +302,7 @@ function filter_core_rules_system($fw, $defaults)
     // block bogons and private nets
     $bogontmpl = ['type' => 'block', 'log' => !isset($config['syslog']['nologbogons']), 'disablereplyto' => 1];
     $privtmpl = ['type' => 'block', 'log' => !isset($config['syslog']['nologprivatenets']),
-      'from' => '10.0.0.0/8,127.0.0.0/8,100.64.0.0/10,172.16.0.0/12,192.168.0.0/16',
+      'from' => '10.0.0.0/8,127.0.0.0/8,100.64.0.0/10,172.16.0.0/12,169.254.0.0/16,192.168.0.0/16',
       'disablereplyto' => 1];
     foreach ($fw->getInterfaceMapping() as $intf => $intfinfo) {
         $fw->registerFilterRule(
@@ -333,7 +333,7 @@ function filter_core_rules_system($fw, $defaults)
             5,
             ['direction' => 'in', 'interface' => $intf, 'ipprotocol' => 'inet6',
             '#ref' => "interfaces.php?if=" . $intf . "#blockpriv",
-            'descr' => "Block private networks from " . $intfinfo['descr'], 'from' => 'fc00::/7',
+            'descr' => "Block private networks from " . $intfinfo['descr'], 'from' => 'fd00::/8,fe80::/64',
             'disabled' => $ipv6_disabled || empty($intfinfo['blockpriv'])],
             $privtmpl
         );
--- a/src/opnsense/scripts/firmware/bogons.sh
+++ b/src/opnsense/scripts/firmware/bogons.sh
@@ -58,7 +58,7 @@ LINES_V4=`wc -l ${WORKDIR}/fullbogons-ipv4.txt | awk '{ print $1 }'`
 if [ $ENTRIES_MAX -gt $((2*ENTRIES_TOT-${ENTRIES_V4:-0}+LINES_V4)) ]; then
     # private and pseudo-private networks will be excluded
     # as they are being operated by a separate GUI option
-    egrep -v "^100.64.0.0/10|^192.168.0.0/16|^172.16.0.0/12|^10.0.0.0/8" ${WORKDIR}/fullbogons-ipv4.txt > ${DESTDIR}/bogons
+    egrep -v "^169.254.0.0/16|^100.64.0.0/10|^192.168.0.0/16|^172.16.0.0/12|^10.0.0.0/8" ${WORKDIR}/fullbogons-ipv4.txt > ${DESTDIR}/bogons
     RESULT=`/sbin/pfctl -t bogons -T replace -f ${DESTDIR}/bogons 2>&1`
     echo "$RESULT" | awk '{ print "Bogons V4 file updated: " $0 }' | logger
 else
@@ -73,7 +73,11 @@ LINES_V6=`wc -l ${WORKDIR}/fullbogons-ipv6.txt | awk '{ print $1 }'`
 if [ $BOGONS_V6_TABLE_COUNT -gt 0 ]; then
     ENTRIES_V6=`pfctl -vvsTables | awk '/-\tbogonsv6$/ {getline; print $2}'`
     if [ $ENTRIES_MAX -gt $((2*ENTRIES_TOT-${ENTRIES_V6:-0}+LINES_V6)) ]; then
-        egrep -iv "^fc00::/7" ${WORKDIR}/fullbogons-ipv6.txt > ${DESTDIR}/bogonsv6
+        # private and pseudo-private networks will be excluded
+        # as they are being operated by a separate GUI option
+        cat ${WORKDIR}/fullbogons-ipv6.txt > ${DESTDIR}/bogonsv6
+        echo "!fd00::/8" >> ${DESTDIR}/bogonsv6
+        echo "!fe80::/64" >> ${DESTDIR}/bogonsv6
         RESULT=`/sbin/pfctl -t bogonsv6 -T replace -f ${DESTDIR}/bogonsv6 2>&1`
         echo "$RESULT" | awk '{ print "Bogons V6 file updated: " $0 }' | logger
     else
@@ -81,7 +85,11 @@ if [ $BOGONS_V6_TABLE_COUNT -gt 0 ]; then
     fi
 else
     if [ $ENTRIES_MAX -gt $((2*ENTRIES_TOT+LINES_V6)) ]; then
-        egrep -iv "^fc00::/7" ${WORKDIR}/fullbogons-ipv6.txt > ${DESTDIR}/bogonsv6
+        # private and pseudo-private networks will be excluded
+        # as they are being operated by a separate GUI option
+        cat ${WORKDIR}/fullbogons-ipv6.txt > ${DESTDIR}/bogonsv6
+        echo "!fd00::/8" >> ${DESTDIR}/bogonsv6
+        echo "!fe80::/64" >> ${DESTDIR}/bogonsv6
         echo "Not updating IPv6 bogons table because IPv6 Allow is off" | logger
     else
         echo "Not saving IPv6 bogons table (IPv6 Allow is off and table-entries limit is potentially too low)" | logger


src/etc/inc/filter.lib.inc
```diff
@@ -302,7 +302,7 @@ function filter_core_rules_system($fw, $defaults)
     // block bogons and private nets
     $bogontmpl = ['type' => 'block', 'log' => !isset($config['syslog']['nologbogons']), 'disablereplyto' => 1];
     $privtmpl = ['type' => 'block', 'log' => !isset($config['syslog']['nologprivatenets']),
-      'from' => '10.0.0.0/8,127.0.0.0/8,100.64.0.0/10,172.16.0.0/12,192.168.0.0/16',
+      'from' => '10.0.0.0/8,127.0.0.0/8,100.64.0.0/10,172.16.0.0/12,169.254.0.0/16,192.168.0.0/16',
```
fichtner, line 305: This IPv4 addition isn’t overly popular. I thought about making this list configurable via GUI, but disabling it is always an option. To some degree this is private but it’s also a bit more public than it isn’t and easily configurable.

src/opnsense/scripts/firmware/bogons.sh
```diff
@@ -73,15 +73,23 @@ LINES_V6=`wc -l ${WORKDIR}/fullbogons-ipv6.txt | awk '{ print $1 }'`
 if [ $BOGONS_V6_TABLE_COUNT -gt 0 ]; then
     ENTRIES_V6=`pfctl -vvsTables | awk '/-\tbogonsv6$/ {getline; print $2}'`
     if [ $ENTRIES_MAX -gt $((2*ENTRIES_TOT-${ENTRIES_V6:-0}+LINES_V6)) ]; then
-        egrep -iv "^fc00::/7" ${WORKDIR}/fullbogons-ipv6.txt > ${DESTDIR}/bogonsv6
+        # private and pseudo-private networks will be excluded
+        # as they are being operated by a separate GUI option
+        cat ${WORKDIR}/fullbogons-ipv6.txt > ${DESTDIR}/bogonsv6
+        echo "!fd00::/8" >> ${DESTDIR}/bogonsv6
```
fichtner, line 79: I was wondering does it have to be the first entry in order to exclude a competing fd00::/8 entry if it ever comes into the list. In that case it would make sense to strip it properly first. This also doesn’t exclude smaller matches if they ever happen in the file (it should match from highest netmask first to lowest last like routes?).

maurice-w: Good questions. I performed some more tests.

This does indeed block ULAs:
```
# cat /usr/local/etc/bogonsv6
fd00::/8
!fd00::/8
```
But this doesn't:
```
# cat /usr/local/etc/bogonsv6
!fd00::/8
fd00::/8
```
This does block fd03::/16:
```
# cat /usr/local/etc/bogonsv6
fd03::/16
!fd00::/8
```
And so does this:
```
# cat /usr/local/etc/bogonsv6
!fd00::/8
fd03::/16
```
So I agree with making !fd00::/8 and !fe80::/10 the first entries, just in case these should ever get added explicitly.
Regarding ULA prefixes longer than /8 getting added: Seems very unlikely to me.

src/etc/inc/filter.lib.inc
fichtner: Sorry I didn't pinpoint the subnet when I wrote this. I'm ok with a separate PR. I was reminded of earlier requests to include CG-NAT subnet here.

src/etc/inc/filter.lib.inc
```diff
@@ -333,7 +333,7 @@ function filter_core_rules_system($fw, $defaults)
             5,
             ['direction' => 'in', 'interface' => $intf, 'ipprotocol' => 'inet6',
             '#ref' => "interfaces.php?if=" . $intf . "#blockpriv",
-            'descr' => "Block private networks from " . $intfinfo['descr'], 'from' => 'fc00::/7',
+            'descr' => "Block private networks from " . $intfinfo['descr'], 'from' => 'fd00::/8,fe80::/64',
```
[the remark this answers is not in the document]
fichtner: agreed 👍

src/opnsense/scripts/firmware/bogons.sh
fichtner: To me these results look like we do not need to modify the table (grep -v) and just make sure our exclusions come first? I'm thinking it discards the second entry because it has the same key and in doing so keeps the negation property of the first one in place.

> Regarding ULA prefixes longer than /8 getting added: Seems very unlikely to me.

I meant as a general issue here, but it's less likely given the small scope of the modifications here.

src/opnsense/scripts/firmware/bogons.sh
```diff
@@ -73,15 +73,23 @@ LINES_V6=`wc -l ${WORKDIR}/fullbogons-ipv6.txt | awk '{ print $1 }'`
...
         echo "Not saving or updating IPv6 bogons (increase table-entries limit)" | logger
     fi
 else
     if [ $ENTRIES_MAX -gt $((2*ENTRIES_TOT+LINES_V6)) ]; then
-        egrep -iv "^fc00::/7" ${WORKDIR}/fullbogons-ipv6.txt > ${DESTDIR}/bogonsv6
+        # private and pseudo-private networks will be excluded
+        # as they are being operated by a separate GUI option
+        cat ${WORKDIR}/fullbogons-ipv6.txt > ${DESTDIR}/bogonsv6
+        echo "!fd00::/8" >> ${DESTDIR}/bogonsv6
+        echo "!fe80::/64" >> ${DESTDIR}/bogonsv6
```
fichtner, line 92:
```suggestion
        : > ${WORKDIR}/bogonsv6
        echo "!fd00::/8" >> ${WORKDIR}/bogonsv6
        echo "!fe80::/10" >> ${WIRKDIR}/bogonsv6
        cat ${WORKDIR}/fullbogons-ipv6.txt >> ${WORKDIR}/bogonsv6
        mv ${WORKDIR}/bogonsv6 ${DESTDIR}/bogonsv6
```

This then? We can remove a few sins of the past here too by preparing the file and moving it into place after it's ready for an atomic replace.

src/opnsense/scripts/firmware/bogons.sh
```diff
@@ -58,7 +58,7 @@ LINES_V4=`wc -l ${WORKDIR}/fullbogons-ipv4.txt | awk '{ print $1 }'`
 if [ $ENTRIES_MAX -gt $((2*ENTRIES_TOT-${ENTRIES_V4:-0}+LINES_V4)) ]; then
     # private and pseudo-private networks will be excluded
     # as they are being operated by a separate GUI option
-    egrep -v "^100.64.0.0/10|^192.168.0.0/16|^172.16.0.0/12|^10.0.0.0/8" ${WORKDIR}/fullbogons-ipv4.txt > ${DESTDIR}/bogons
+    egrep -v "^169.254.0.0/16|^100.64.0.0/10|^192.168.0.0/16|^172.16.0.0/12|^10.0.0.0/8" ${WORKDIR}/fullbogons-ipv4.txt > ${DESTDIR}/bogons
```
fichtner, line 61: once we have IPv6 figured out (see last suggestion) we can adjust this the same way in a separate PR? :)

src/opnsense/scripts/firmware/bogons.sh
maurice-w: > To me these results look like we do not need to modify the table (grep -v) and just make sure our exclusions come first? I'm thinking it discards the second entry because it has the same key and in doing so keeps the negation property of the first one in place.

Exactly.

src/etc/inc/filter.lib.inc
maurice-w: Your comment was about 100.64.0.0/10? Block it in bogons instead of in private networks? Or not block it at all? Either way, I agree this should be a separate PR.

You're okay with blocking 169.254.0.0/16 in private networks and removing it from the bogons here in this PR?

src/opnsense/scripts/firmware/bogons.sh
maurice-w: Good idea. So for this PR, adding 169.254.0.0/16 here is okay with you?

fichtner: Let's break this into two PRs for each address family, but both also changing their egrep -v into the method proposed for IPv6

src/opnsense/scripts/firmware/bogons.sh
```diff
@@ -73,15 +73,27 @@ LINES_V6=`wc -l ${WORKDIR}/fullbogons-ipv6.txt | awk '{ print $1 }'`
...
+        mv ${WORKDIR}/bogonsv6 ${DESTDIR}/bogonsv6
         RESULT=`/sbin/pfctl -t bogonsv6 -T replace -f ${DESTDIR}/bogonsv6 2>&1`
         echo "$RESULT" | awk '{ print "Bogons V6 file updated: " $0 }' | logger
     else
         echo "Not saving or updating IPv6 bogons (increase table-entries limit)" | logger
     fi
 else
     if [ $ENTRIES_MAX -gt $((2*ENTRIES_TOT+LINES_V6)) ]; then
-        egrep -iv "^fc00::/7" ${WORKDIR}/fullbogons-ipv6.txt > ${DESTDIR}/bogonsv6
+        # private and pseudo-private networks will be excluded
```
fichtner, line 90: best to refactor this double usage as well but I can do it post-merge

src/opnsense/scripts/firmware/bogons.sh
maurice-w: I reverted all IPv4 changes.

src/opnsense/scripts/firmware/bogons.sh
maurice-w: > I can do it post-merge

Thanks!

maurice-w:
What about the GUI update (help texts)?

fichtner:
IPv6 private networks wasn't documented before but since we have automatic rule visibility this auto-documents already.

PR #8594, opened 2025-04-29
captive portal: allow disabling automatic firewall rules

swhite2:


src/opnsense/mvc/app/controllers/OPNsense/CaptivePortal/forms/dialogZone.xml
```diff
@@ -17,6 +17,17 @@
...
     </field>
+    <field>
+        <id>zone.disableRules</id>
+        <type>checkbox</type>
+        <label>Disable firewall rules</label>
+        <help>
+            <![CDATA[If this option is set, no automatic firewall rules for portal redirection and traffic
+            blocking will be generated. This option allows you to override the default portal behavior for advanced use cases,
+            such as redirections for DNS. See the documentation to see which rules you should implement in this scenario.]]>
+        </help>
```
fichtner, line 28:
```suggestion
        <help>If this option is set, no automatic firewall rules for portal redirection and traffic blocking will be generated. This option allows you to override the default portal behavior for advanced use cases, such as redirections for DNS. See the documentation to see which rules you should implement in this scenario.</help>
```

only looking out for translators ;)

src/opnsense/mvc/app/controllers/OPNsense/CaptivePortal/forms/dialogZone.xml
```diff
@@ -18,7 +18,10 @@
         <help><![CDATA[Select interface(s) to enable for captive portal.]]></help>
     </field>
     <field>
-        <id>zone.authservers</id>
+        <id>zone.disableRules</id>
```
fichtner, line 21: it looks like this landed in another field?

swhite2: Yeah, github seems a bit slow today, applied your suggestion twice ...

fichtner (approved):
FWIW: LGTM :)

PR #8598, opened 2025-04-29
Refactor IPv4 bogons and move link-locals to private networks

maurice-w:
Same as #8591, but for IPv4.

@fichtner How do you want to proceed with 100.64.0.0/10?

### DIFF
--- a/src/etc/inc/filter.lib.inc
+++ b/src/etc/inc/filter.lib.inc
@@ -302,7 +302,7 @@ function filter_core_rules_system($fw, $defaults)
     // block bogons and private nets
     $bogontmpl = ['type' => 'block', 'log' => !isset($config['syslog']['nologbogons']), 'disablereplyto' => 1];
     $privtmpl = ['type' => 'block', 'log' => !isset($config['syslog']['nologprivatenets']),
-      'from' => '10.0.0.0/8,127.0.0.0/8,100.64.0.0/10,172.16.0.0/12,169.254.0.0/16,192.168.0.0/16',
+      'from' => '10.0.0.0/8,172.16.0.0/12,192.168.0.0/16,127.0.0.0/8,100.64.0.0/10,169.254.0.0/16',
       'disablereplyto' => 1];
     foreach ($fw->getInterfaceMapping() as $intf => $intfinfo) {
         $fw->registerFilterRule(
--- a/src/etc/inc/filter.lib.inc
+++ b/src/etc/inc/filter.lib.inc
@@ -302,7 +302,7 @@ function filter_core_rules_system($fw, $defaults)
     // block bogons and private nets
     $bogontmpl = ['type' => 'block', 'log' => !isset($config['syslog']['nologbogons']), 'disablereplyto' => 1];
     $privtmpl = ['type' => 'block', 'log' => !isset($config['syslog']['nologprivatenets']),
-      'from' => '10.0.0.0/8,127.0.0.0/8,100.64.0.0/10,172.16.0.0/12,192.168.0.0/16',
+      'from' => '10.0.0.0/8,127.0.0.0/8,100.64.0.0/10,172.16.0.0/12,169.254.0.0/16,192.168.0.0/16',
       'disablereplyto' => 1];
     foreach ($fw->getInterfaceMapping() as $intf => $intfinfo) {
         $fw->registerFilterRule(
--- a/src/opnsense/scripts/firmware/bogons.sh
+++ b/src/opnsense/scripts/firmware/bogons.sh
@@ -76,6 +76,7 @@ LINES_V4=`wc -l ${WORKDIR}/fullbogons-ipv4.txt | awk '{ print $1 }'`
 
 if [ $ENTRIES_MAX -gt $((2*ENTRIES_TOT-${ENTRIES_V4:-0}+LINES_V4)) ]; then
     update_bogons fullbogons-ipv4.txt bogons 100.64.0.0/10 169.254.0.0/16 192.168.0.0/16 172.16.0.0/12 10.0.0.0/8
+
     RESULT=`/sbin/pfctl -t bogons -T replace -f ${DESTDIR}/bogons 2>&1`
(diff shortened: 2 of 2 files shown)


src/opnsense/scripts/firmware/bogons.sh
```diff
@@ -73,10 +73,9 @@ ENTRIES_MAX=`pfctl -s memory | awk '/table-entries/ { print $4 }'`
 ENTRIES_TOT=`pfctl -vvsTables | awk '/Addresses/ {s+=$2}; END {print s}'`
 ENTRIES_V4=`pfctl -vvsTables | awk '/-\tbogons$/ {getline; print $2}'`
 LINES_V4=`wc -l ${WORKDIR}/fullbogons-ipv4.txt | awk '{ print $1 }'`
+
 if [ $ENTRIES_MAX -gt $((2*ENTRIES_TOT-${ENTRIES_V4:-0}+LINES_V4)) ]; then
-    # private and pseudo-private networks will be excluded
-    # as they are being operated by a separate GUI option
-    egrep -v "^100.64.0.0/10|^192.168.0.0/16|^172.16.0.0/12|^10.0.0.0/8" ${WORKDIR}/fullbogons-ipv4.txt > ${DESTDIR}/bogons
+    update_bogons fullbogons-ipv4.txt bogons 100.64.0.0/10 169.254.0.0/16 192.168.0.0/16 172.16.0.0/12 10.0.0.0/8
```
fichtner, line 78:
```suggestion
    update_bogons fullbogons-ipv4.txt bogons 100.64.0.0/10 169.254.0.0/16 192.168.0.0/16 172.16.0.0/12 10.0.0.0/8

```

src/opnsense/scripts/firmware/bogons.sh
```diff
@@ -73,10 +73,10 @@ ENTRIES_MAX=`pfctl -s memory | awk '/table-entries/ { print $4 }'`
...
```
fichtner, line 78:
```suggestion
    update_bogons fullbogons-ipv4.txt bogons 10.0.0.0/8 172.16.0.0/12 192.168.0.0/16 127.0.0.0/8 100.64.0.0/10 169.254.0.0/16  
```

reorder by type, add 127/8 even if it isn't in the bogons for consistency

src/etc/inc/filter.lib.inc
```diff
@@ -302,7 +302,7 @@ function filter_core_rules_system($fw, $defaults)
     // block bogons and private nets
     $bogontmpl = ['type' => 'block', 'log' => !isset($config['syslog']['nologbogons']), 'disablereplyto' => 1];
     $privtmpl = ['type' => 'block', 'log' => !isset($config['syslog']['nologprivatenets']),
-      'from' => '10.0.0.0/8,127.0.0.0/8,100.64.0.0/10,172.16.0.0/12,192.168.0.0/16',
+      'from' => '10.0.0.0/8,127.0.0.0/8,100.64.0.0/10,172.16.0.0/12,169.254.0.0/16,192.168.0.0/16',
```
fichtner, line 305:
```suggestion
      'from' => '10.0.0.0/8,172.16.0.0/12,192.168.0.0/16,127.0.0.0/8,100.64.0.0/10,169.254.0.0/16',
```

fichtner:
@maurice-w leave CG-NAT as is. Merged, thanks!

fichtner:
@maurice-w I'm doing a bit of auditing I'm wondering why we de-bogonized 127.0.0.0/8 here? Cannot find more context.

maurice-w:
@fichtner You said it in your comment when [suggesting this change](https://github.com/opnsense/core/pull/8598#discussion_r2067853558): "add 127/8 even if it isn't in the bogons for consistency"

Loopback isn't in fullbogons-ipv4.txt so de-bogonizing does nothing. "For consistency" seems to mean that everything which is defined as a private network should be explicitly de-bogonized, which makes sense to me.

fichtner:
@maurice-w ok thanks. so the weird part is excluding ::1/128 for IPv6 bogons causes a bit of splatter in the new bogons published since it splits out all the networks to exclude it specifically.

I'm just looking for consistency also because https://github.com/opnsense/core/issues/10694 probably proves all of this does nothing :)

maurice-w:
@fichtner Maybe instead of adding ::1/128 to the IPv6 private addresses and bogon exclusions, we could remove 127.0.0.0/8 from the IPv4 ones?
Loopback addresses are not in the published bogons so excluding them is not required. And unlike everything else we defined as "private addresses", they should never appear on the wire. But I may be missing something.

While we're revisiting bogons: #10125 is patiently waiting. ;)

fichtner:
Ok then we'll keep loopback for IPv6 in the bogons to simplify and perhaps consider removing loopback from private again. The situation seems unclear but then again loopback shouldn't be accepted from the wire anyway.

fichtner:
> Loopback addresses are not in the published bogons so excluding them is not required. And unlike everything else we defined as "private addresses", they should never appear on the wire. But I may be missing something.

127.0.0.0/8 is in https://team-cymru.org/Services/Bogons/fullbogons-ipv4.txt or maybe I'm misunderstanding. We currently exclude it, but maybe we should not.

maurice-w:
@fichtner Neither ::1/128 nor 127.0.0.0/8 are in https://pkg.opnsense.org/FreeBSD:15:amd64/26.7/sets/bogons.txz. How these lists get created I don't know. Are these based on the "Team Cymru" lists, but with additional processing? I can't find any information / documentation about this. A link would be appreciated as a foundation for further discussion.

fichtner:
@maurice-w yes we changed it this week. We still get the Cymru upstream file and preprocess it like bogons.sh did it (but correctly splitting the subnets). I'm simply pondering leaving ::1/128 and 127.0.0.0/8 in the file again because the exclusion seems like futile micromanagement (especially in the IPv6 case).

fichtner:
PS: that's what https://github.com/opnsense/core/commit/17ca47489f is about

maurice-w:
> I'm simply pondering leaving ::1/128 and 127.0.0.0/8 in the file again

Makes sense to me.

Issue #6703, opened 2023-08-03
openvpn: CSO support for previously used freeform custom settings


fichtner:
**Important notices**

Before you add a new report, we ask you kindly to acknowledge the following:

- [x] I have read the contributing guide lines at https://github.com/opnsense/core/blob/master/CONTRIBUTING.md
- [x] I am convinced that my issue is new after having checked both open and closed issues at https://github.com/opnsense/core/issues?q=is%3Aissue

**Is your feature request related to a problem? Please describe.**

Custom options are no longer available which requires us to add new fields which can be validated properly.

**Describe the solution you like**

Add requested fields:

* ~`sndbuf 524288`~ (increase system values if this is required, documentation isn't very enthusiastic about this)
* ~`rcvbuf 524288`~ (increase system values if this is required, documentation isn't very enthusiastic about this)
* `push "sndbuf 524288"`
* `push "rcvbuf 524288"`

Fields done:

* `route-gateway`

**Describe alternatives you considered**

File based overrides for CSOs are not very elegant and a nightmare to handle for the user.

**Additional context**

https://openvpn.net/community-resources/reference-manual-for-openvpn-2-6/
https://forum.opnsense.org/index.php?topic=35149.0

smeretech:
As reffered to #6801, necessity to push "**route-gateway <gateway>**".

smeretech:
Hello
Sorry, I just wanted to know if it will be considered to be able to put the **route-gateway** option back in again. Unfortunately, I cannot upgrade from 23.1 to 23.7 as the CSO is also modified for those using the legacy part of openVPN.

smeretech:
Thanks for your reply.
the context in which I use it is to deliver a single OpenVPN instance/server on multi projects that need to be separated.
In my specific case, the server adopts a /16 internal network.
Through CSOs, I would assign different sub-networks of this /16 to each workgroup of users but the only way to set a gateway to clients that falls within this sub-network was to use **route-gateway**. otherwise, OpenVPN assigns the first IP of the /16 as the gateway resulting in the error of not being able to reach it.

example:
Tunnel network: 10.50.0.0/16

Group1 (CSO) IPv4 Tunnel Network: 10.50.0.0/28 (gateway 10.50.0.14)
Group2 (CSO) IPv4 Tunnel Network: 10.50.0.16/28 (gateway 10.50.0.30)
Group3 (CSO) IPv4 Tunnel Network: 10.50.0.32/27 (gateway 10.50.0.62)

[...]

PR #8603, opened 2025-05-06
OpenVPN: Add support for verify-x509-name/sndbuf/rcvbuf options

laozhoubuluo:
Add support for verify-x509-name for improve security in some scenarios (especially point-to-point scenarios).

Add support for sndbuf/rcvbuf for optimize the performance of OpenVPN in different scenarios.

### DIFF
--- a/src/opnsense/mvc/app/controllers/OPNsense/OpenVPN/forms/dialogInstance.xml
+++ b/src/opnsense/mvc/app/controllers/OPNsense/OpenVPN/forms/dialogInstance.xml
@@ -680,4 +680,34 @@ Set to 0 to disable, remember to change your client as well.
             <visible>false</visible>
         </grid_view>
     </field>
+    <field>
+        <id>instance.verify-x509-name</id>
+        <label>Verify X.509 name</label>
+        <type>text</type>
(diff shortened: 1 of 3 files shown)


laozhoubuluo:
> we looked at sndbuf and rcvbuf before, but didn't look very useful at the time, see also #6703.
> 
> `net.inet.tcp.sendspace` and `net.inet.tcp.recvspace`, `net.inet.udp.recvspace` should show the current defaults if I'm not mistaken.

sndbuf and rcvbuf are just inherited from the original configuration habits. I am not sure if there will be any other side effects if they are changed at the operating system level.
In addition, the logic of empty = system default value is valid. The existing framework does not seem to be very convenient for real-time reading of operating system values.

AdSchellevis:
while converting configurations, it's often also a good idea to reassess habits I suppose. During our testing, these values didn't turn out to be very relevant compared to the defaults.

laozhoubuluo:
> while converting configurations, it's often also a good idea to reassess habits I suppose. During our testing, these values didn't turn out to be very relevant compared to the defaults.

OK. I have removed the two corresponding settings parameters.

Issue #7228, opened 2024-02-11
OpenVPN generates invalid configuration for net30 [non-legacy]


joni1993:
**Important notices**

Before you add a new report, we ask you kindly to acknowledge the following:

- [x] I have read the contributing guide lines at https://github.com/opnsense/core/blob/master/CONTRIBUTING.md
- [x] I am convinced that my issue is new after having checked both open and closed issues at https://github.com/opnsense/core/issues?q=is%3Aissue

**Describe the bug**
When transitioning from legacy to the new configuration interface, the generated config is invalid for net30 configurations and won't start

**To Reproduce**

1. Go to 'Generate OpenVPN Server config in new UI'
2. Use 'net30' as 'Topology'
3. Apply

**Expected behavior**

OpenVPN instance starts normally 

**Relevant log files**

```
/usr/local/opnsense/scripts/openvpn/ovpn_service_control.php: The command '/usr/local/sbin/openvpn --config '/var/etc/openvpn/instance-84b64fa6-83fe-41e9-9862-9d7816065e4b.conf'' returned exit code '1', the output was ''
Options error: --client-connect requires --mode server
```

**Environment**

OPNsense 24.1.1-amd64
FreeBSD 13.2-RELEASE-p9
OpenSSL 3.0.13

AdSchellevis:
https://github.com/opnsense/core/commit/66fd0e4699ae06bc9a1038ecdb441a6c6b71a875

joni1993:
Thank you - I did not find this.

But one problem remains: the generated ifconfig-pool section causes "IPv4 pool size is too small (1), must be at least 2" on startup. Removing that config line at least brings up the instance (the config line is also not generated by the legacy interface)

joni1993:
Also setting verify-client-certificate to none results in `Options error: --verify-client-cert none|optional must be used with --management-client-auth, an --auth-user-pass-verify script, or plugin`

AdSchellevis:
I'm not sure we can fix that, it depends on the exact settings being configured (which aren't in the ticket).

joni1993:
Yeah the second issue is more of like a validation issue probably.

The config is the most basic s2s one:
```
Role: Server
Protocol: UDP
Port: 1194
Type: tun
Server: x.x.x.8/30
Topology: net30
Certificate: <selected>
Certificate Depth: Do not check
TLS Statis key: [auth] <selected>
Auth: SHA256
Data Cipher: AES-256-GCM
[...]

PR #8604, opened 2025-05-06
OpenVPN: Use OpenVPN's official p2p mode, consistent with standard OpenVPN behavior

laozhoubuluo:
OpenVPN peer-to-peer tunnels in legacy mode can support traffic from any address instead of a specific peer-to-peer address. This is used for dynamic routing-based networking or scenarios where NAT is not allowed. This is achieved using OpenVPN's native p2p mode (mode p2p).

In the new version of the OPNsense's OpenVPN configuration interface, server mode (mode server) is hard-coded in the p2p scenario of configuring the /30 network segment. The server mode cannot support this scenario because it involves multiple clients. This causes traffic from non-specified peer-to-peer addresses to be dropped.

This PR restores OpenVPN's native p2p mode in peer-to-peer mode and removes client parameters that are invalid for real peer-to-peer requirements, thereby resolving this issue.

Ref: #7228

### DIFF
--- a/src/opnsense/mvc/app/models/OPNsense/OpenVPN/OpenVPN.php
+++ b/src/opnsense/mvc/app/models/OPNsense/OpenVPN/OpenVPN.php
@@ -584,11 +584,8 @@ class OpenVPN extends BaseModel
                             $masklong = ip2long($mask);
                             $ip1 = long2ip32((ip2long32($parts[0]) & $masklong) + ($masklong == 0xfffffffe ? 0 : 1));
                             $ip2 = long2ip32((ip2long32($parts[0]) & $masklong) + ($masklong == 0xfffffffe ? 1 : 2));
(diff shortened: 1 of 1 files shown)


AdSchellevis:
this will undo https://github.com/opnsense/core/commit/66fd0e4699ae06bc9a1038ecdb441a6c6b71a875 and reintroduce issues reported earlier (for both net30 as p2p).

AdSchellevis:
I'm not sure what the question is, but if the p2p serves a larger net (<=29), the address calculation wouldn't be used in our case.  Best go back to the drawing board and start with a ticket explaining what we are trying to fix and which circumstances.

laozhoubuluo:
> I'm not sure what the question is, but if the p2p serves a larger net (<=29), the address calculation wouldn't be used in our case. Best go back to the drawing board and start with a ticket explaining what we are trying to fix and which circumstances.

I'm create a new ticket https://github.com/opnsense/core/issues/8622 for details.

PR #8608, opened 2025-05-08
captive portal: redirect on direct access as well

swhite2:
The filter rule was redundant, as the zone server wasn't listening on (self), only localhost, which requires a redirect regardless of pass rule.

Since this one is `pass`, no filter rule is necessary anymore either.


swhite2:
Scratch this, the zone server is listening on `0.0.0.0`, so the current firewall rule is correct for direct access.

PR #8609, opened 2025-05-08
Fix help message for Kea static routes

liujed:
Using semicolons to separate destination–router pairs is a syntax error. Kea expects the `static-routes` option to be a list of IP addresses: https://github.com/isc-projects/kea/blob/60222843a6b2c4e7a6c7d4e63f88d0dcfa0d5b97/doc/examples/kea4/all-options.json#L415-L430


AdSchellevis:
@liujed thanks!

Issue #8620, opened 2025-05-09
DNSmasq DHCP options incomplete, untyped and confusing


meyergru:
**Important notices**

Before you add a new report, we ask you kindly to acknowledge the following:

- [x] I have read the contributing guide lines at https://github.com/opnsense/core/blob/master/CONTRIBUTING.md
- [x] I am convinced that my issue is new after having checked both open and closed issues at https://github.com/opnsense/core/issues?q=is%3Aissue

**Is your feature request related to a problem? Please describe.**

Just three observations:

1. It is unfortunate that the "DHCP options" are that complex to set up. Effectively, you need at least 6 lines for every (V)LAN / subnet, which soon multiplies up and gets confusing:

- router/gateway
- netmask
- DNS server (which, according to the resulting DNSmasq config file, has a default of 0.0.0.0, thus could be left out in most cases, yet you see that only when you inspect the resuling config)
- NTP server
- DNS domain
- DNS search list
...

plus their IPv6 counterparts!

2. There are some options completely missing, without the possibility to even configure them as custom options, like:

- time-server (4)
- wpad (252)
- Unifi controller (43 with a specific format for the IP)

These options should be added to the list of available DHCP options. Also, a "custom" option like in ISC DHCP would be appreciated for cases like these.

Call this lack of feature-parity for some obscure options, but was this not the reason for going this route when Kea was not up to par with ISC DHCP?

3. I used the "log-server (7)" option with a DNS name instead of an IP address, only to find that DNSmasq would fail at that. This works with ISC DHCP by the "IP address or host" type in the custom options. This is an evident example of lacking strong types.

**Describe the solution you like**

For point 1 from above:

This was much easier in ISC DHCP, because often-used options like these (and some more) were summarized nicely on one page per interface. Maybe tabs for each interface would be a nice addition. It is also easier to have strong types for the most-often fields. As an example, consider the IPv6 DNSSL festure of RADVD (domain-search (24)), which should give DNS search domains in this syntax as seen with radvdump:

        DNSSL lan iot guest dmz
        {
                AdvDNSSLLifetime 86400;
        }; # End of DNSSL definition

where the original GUI input was "lan;iot;guest;dmz" and which is correct for the "domain-search (119)" option for DHCPv4.
You have to specifiy "lan;iot;guest;dmz" for IPv4 but "lan iot guest dmz" for IPv6, or else you will end up with:

        DNSSL lan;iot;guest;dmz
        {
                AdvDNSSLLifetime 86400;
[...]

PR #8626, opened 2025-05-09
Update dnsmasq_dhcp_options.py for all IANA and add custom options

meyergru:
See issue: https://github.com/opnsense/core/issues/8620, especially the missing IANA options and custom options parts (i.e. problem #2).

It does not fix the lack of strong types, but allows to use most IANA DHCPv4 options while preserving DNSmasq names, plus it adds all custom options in both DHCPv4 and DHCPv6 as "option-N [N]" via the GUI.


Monviech:
If that is really the way to fix missing options by copying the iana defined ones for ipv4 and ipv6, we could theoretically just cut this script out completely and use static json files for options4 and options6.

JsonKeyValueStoreField - SourceFile
https://docs.opnsense.org/development/frontend/models_fieldtypes.html

cc @AdSchellevis

meyergru:
It is not quite that easy. The disadvantage of that approach would be:

1. that new options, which will build in later into new DNSmasq versions would not appear under their newly assigned names.
2. that you would have no generic/custom options available that neither IANA list nor DNSmasq have built in yet.

As said, this is only to address the "incompleteness" aka "feature parity" problem, not the other two. **You** wanted to break the issue down into smaller parts, remember? ;-)

AdSchellevis:
I rather stick to the approach of using `/usr/local/sbin/dnsmasq --help dhcp` and overlay extra's, which has two advantages, the naming of "known" attributes matches dnsmasqs and when options are added, we automatically include them.

In terms of maintenance, it's likely better to use the upstream data as offered (e..g https://www.iana.org/assignments/bootp-dhcp-parameters/options.csv and possibly others for different rfcs) as its easier to explain where data originates from for future reference (and requires less code).

If someone can add the missing rfc's to this ticket (e.g. option 252 and probably some others), I don't mind adding the references somewhere next week. I expect iana offers similarly structured data for other options as well.

Monviech:
Sounds like a good plan. Thanks.

meyergru:
I would be fond of the idea to use a better source for the IANA options. You will have to rename them in order ot be lowercase and replace blanks by dashes in order to keep the DNSmasq naming scheme (although that is not a strict requirement, since these names are used only on the UI presentation layer.

However, when you actaully look at the IANA list (and the CSV as well), you will find that above option 125, it sometimes gets fishy, see options 0,126, 128, 130, 163-174, 175, 224-254, i.e. this list gets more and more informal and less likely to be useable without some processing, which would involve some intelligent coding to skip or enumerate ranges

Thus, I did that manually to the best of my ability, knowing that "fishy" options can be used via "option-N" by my approach.

I added in option 252, because that is in use by the WPAD option in ISC DHCP right now. In the IANA list, it is listed as "Reserved (Private Use)", yet I found it to be more user-friendly the way I did it. I found no RFC, but only this: https://datatracker.ietf.org/doc/html/draft-ietf-wrec-wpad-01

For Unifi, I did not invent a specific name like "unifi-server", because I know of other real-world uses of option 43, so there is no real standard, it stayed at "vendor-specific" from IANA. Also, I know of no RFC for any of these use cases.

meyergru:
Oh, BTW: There is a similar list that could/should probably be used for DHCPv6 initialisation: https://www.iana.org/assignments/dhcpv6-parameters/dhcpv6-parameters.xhtml. I do not use DHCPv6 apart from "domain-search [24]", which DNSmasq automagically converts to a DNSSL RA option even without DHCPv6 enabled.

AdSchellevis:
this https://github.com/opnsense/core/commit/ea157394fe31df938243bacc486da84b8039c00d should do the trick then

meyergru:
Really? I have not tried it, but it does not offer custom options as discussed, so the wpad url (option-252) cannot be used.

Monviech:
I guess the script splits ranges into individual entries in the result:

https://github.com/opnsense/core/blob/ea157394fe31df938243bacc486da84b8039c00d/src/opnsense/scripts/dns/dnsmasq_dhcp_options.py#L48

meyergru:
Ah, right. Missed that. So as long as the CSV contains the range, it will be in the output. WIth 151-65535 for DHCPv6, that will be a log of options....

AdSchellevis:
yep, we expand the range. From earlier lists I haven't seen option codes being used without an iana assignment, so leaving the unassigned ones out is likely the safest option (limiting chances of misconfigurations).

razza-guhl:
Unfortunately, the codes described as “unassigned” by IANA cannot be selected in the GUI (the script filters them out).
Some of these unassigned codes (example range 178-207 and 214-219) are required and used in practice.

My suggestion is that all options from 1-254 can be used without restriction within dnsmasq/OPNsense.

The risk of misconfiguration is low. Options that are not required by a DHCP client are discarded.

PR #8644, opened 2025-05-13
Adjust a few instances of dnsmasq labeling/description primarily for the services widget display

xconverge:
Not sure what the roadmap/plans for this are but I noticed this after migrating from KEA to dnsmasq for my DHCP usage today (especially since I use 0 for the DNS port to disable it, which makes the service name/label of `Dnsmasq DNS` even more confusing for me). I figured it would be easier to open a PR instead of an issue for such a small thing...

image


fichtner (approved):
Nice, thanks

Monviech (approved):
I'm fine with this, thank you :)

fichtner:
Merged, thanks!

Issue #8642, opened 2025-05-12
dnsmasq: Partial IPv6 addresses from static hosts polluting DNS


eamaclean:
**Important notices**

Before you add a new report, we ask you kindly to acknowledge the following:

- [x] I have read the contributing guide lines at https://github.com/opnsense/core/blob/master/CONTRIBUTING.md
- [ ] I am convinced that my issue is new after having checked both open and closed issues at https://github.com/opnsense/core/issues?q=is%3Aissue

**Describe the bug**

When static hosts are assigned with partial v6 addresses, to be completed by the constructor setting on a range, /var/etc/dnsmasq-hosts is created containing the partial address, causing invalid DNS records containing a partial address to be returned.

**To Reproduce**

Create a host definition with an partial IPv6 address such as ::1234 and attempt to resolve the hostname through dnsmasq

**Expected behavior**

Only complete IP addresses to be added to hosts files, or partial addresses not to be accepted when defining static hosts.

**Describe alternatives you considered**

Selecting an IPv6 range when defining static hosts to allow partial addresses to be automatically completed.

**Additional context**

This was previously reported in #8623 but the original reporter has walked away from dnsmasq and closed the issue.

**Environment**

OPNsense 25.1.6 (amd64).

Monviech:
We have to see if that can be solved via adjusting the configuration of dnsmasq.

fichtner:
@Monviech how does the dnsmasq.conf look like in this case?

Monviech:
@fichtner 

From what I remember without double checking:

I think this section in the code is from when dhcp-host did not exist yet  for static host overrides.

Im not sure that if a host has the dhcp-host directive, if it still needs this too.

It could be validated by checking if hwaddr or client_identifier exists, and only if not it is assumed to be a normal dns host override. And if it exists we skip because its a dhcp-host.

Monviech:
Hey there, can you test if this fixes it for you? Execute this command in a root shell and afterwards press Apply in dnsmasq.

`opnsense-patch https://github.com/opnsense/core/commit/13df9c5bf`

eamaclean:
That does the trick. One less stumbling block. Thanks!

PR #8653, opened 2025-05-14
dnsmasq: Do not create entries in dnsmasq-hosts file for dhcp-host entries

Monviech:
Fixes: https://github.com/opnsense/core/issues/8642

As the form allows both Host Overrides and dhcp-host entries to be created, there can be wrong entries in the /var/etc/dnsmasq-host file.

Dnsmasq registers the names of leases dynamically. We do not need to additionally add them statically.
If there is already something in this host file, registering the name will fail, as also shown here in the forum:

https://forum.opnsense.org/index.php?topic=47135.msg237342#msg237342

It also prevents partial IPv6 addresses to be resolved with their full prefix and registered, returning just partial entries.

### DIFF
--- a/src/etc/inc/plugins.inc.d/dnsmasq.inc
+++ b/src/etc/inc/plugins.inc.d/dnsmasq.inc
@@ -180,6 +180,10 @@ function _dnsmasq_add_host_entries()
     }
 
     foreach ($dnsmasqcfg['hosts'] as $host) {
+        // Do not create entries in dnsmasq-hosts file for dhcp-host entries
+        if (!empty($host['client_id']) || !empty($host['hwaddr'])) {
+            continue;
+        }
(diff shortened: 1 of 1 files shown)


src/etc/inc/plugins.inc.d/dnsmasq.inc
```diff
@@ -180,6 +180,10 @@ function _dnsmasq_add_host_entries()
     }
 
     foreach ($dnsmasqcfg['hosts'] as $host) {
+        // Do not create entries in dnsmasq-hosts file for dhcp-host entries
```
fichtner, line 183:
```suggestion
        /* do not create entries for dhcp-host types */
```

just for clarification

fichtner (approved):
minor comment tweak, otherwise very reasonable

Issue #8647, opened 2025-05-13
mvc: apply button message on successful apply


fichtner:
**Important notices**

Our forum is located at https://forum.opnsense.org , please consider joining discussions there in stead of using GitHub for these matters.

Before you ask a new question, we ask you kindly to acknowledge the following:

- [x] I have read the contributing guide lines at https://github.com/opnsense/core/blob/master/CONTRIBUTING.md
- [x] I am convinced that my issue is new after having checked both open and closed issues at https://github.com/opnsense/core/issues?q=is%3Aissue

Having talked about checkmarks and other means of visually showing the finished apply it would probably be best to use the space at the right of the apply button to indicate the apply finished with a message there.

This is also important because on remote sessions the video encoding may hide the spinner or not show it spinning.

Only do this on the shared apply button template. Plugins need to be extended with it later.

PR #8655, opened 2025-05-14
ui: Add success message in base_apply_button

Monviech:
Fixes: https://github.com/opnsense/core/issues/8647


fichtner:
looks good so far -- not sure if the style-block is very elegant but no need to rush this into a release either

Monviech:
As discussed I don't like this either right now so this needs some more polishing.

Issue #8671, opened 2025-05-16
Dnsmasq configuration won't accept `infinite` as a lease time value


wrobelda:
**Important notices**

- [x] I have read the contributing guide lines at https://github.com/opnsense/core/blob/master/CONTRIBUTING.md
- [x] I am convinced that my issue is new after having checked both open and closed issues at https://github.com/opnsense/core/issues?q=is%3Aissue

**Describe the bug**

The 'infinite' lease time is an acceptable vale for dhcp-range and dhcp-host options in dnsmasq, however, it is not possible to configure that via the UI, as it expects an integer value.

https://thekelleys.org.uk/dnsmasq/docs/dnsmasq-man.html

1. Edit a dnmasq range or a host 
2. Type in 'infinite' into Lease time field 
3. See error

**Expected behavior**

The `infinite` (or possibly a `-1`) value should be acceptable in this field.

**Environment**

OPNsense 25.1.6_4-amd64
FreeBSD 14.2-RELEASE-p3
OpenSSL 3.0.1

meyergru:
The discussion was from here: https://forum.opnsense.org/index.php?topic=47245

It could most easily done by changing in Dnsmasq.xml like:

```
            <lease_time type="TextField">
                <ValidationMessage>Should be an integer value or "infinite"</ValidationMessage>
                <Mask>/^([1-9]\d*)|(infinite)$/</Mask>
            </lease_time>
```

(beware, there are 2 occurrences!) and dialogDHCPrange.xml like:

```
    <field>
        <id>range.lease_time</id>
        <label>Lease time</label>
        <type>text</type>
        <hint>86400</hint>
        <help>Defines how long the addresses (leases) given out by the server are valid (in seconds or 'infinite')</help>
        <grid_view>
            <visible>false</visible>
        </grid_view>
    </field>
```

and dialogHostOverride.xml like:

```
    <field>
        <id>host.lease_time</id>
        <label>Lease time</label>
        <type>text</type>
        <hint>86400</hint>
        <help>Defines how long the addresses (leases) given out by the server are valid (i
n seconds or 'infinite')</help>
        <grid_view>
            <visible>false</visible>
        </grid_view>
    </field>
```

from: https://docs.opnsense.org/development/frontend/models_fieldtypes.html#textfield

By doing it like so, no further code changes are neccessary. I changed it locally, did a "configctl webgui restart" and it works.

fichtner:
Can we please stop doing regexes 🤷‍♂️

[...]

PR #8678, opened 2025-05-17
dnsmasq: Allow lease_time 0 to set it infinite

Monviech:
Fixes: https://github.com/opnsense/core/issues/8671


fichtner (approved):

AdSchellevis (approved):
I don't think there's another functional meaning of `0` at the moment, which would make this the easiest choice with the least amount of noise. For consistency we should indeed add this in hosts and ranges, although I highly doubt adding `infinite` in a range is a good idea (and could easily deplete your pool)/

PR #8688, opened 2025-05-18
Firewall: Aliases - allow setting a custom auth http header

nox-404:
It is currently only possible to use `Authorization: Bearer ${token}` form of header authentication mechanism.

This PR allow to set a specific header in aliases such as `Key: ${token}`


nox-404:
I tested this by applying a git patch directly on my instance of opnsense (`OPNsense 25.1.6_4-amd64`)
I don't have a freebsd on hand to build this unfortunately

DominicTWHV (commented):
Password validation logic looks good, consider adding `$username` validation as per RFC 7230? 

eg: `/^[A-Za-z0-9-]+$/`

nox-404:
I added a length and charset limit on the key ($username) too

DominicTWHV (commented):
Looks good to me, validation logic seems fine

AdSchellevis:
@nox-404 merged, thanks!

Issue #8694, opened 2025-05-19
25.1.7-Dnsmasq-Hosts-Aliases not working (also applies to static reservations that had no DHCP registration yet)


meyergru:
**Important notices**

Before you add a new report, we ask you kindly to acknowledge the following:

- [x] I have read the contributing guide lines at https://github.com/opnsense/core/blob/master/CONTRIBUTING.md
- [x] I am convinced that my issue is new after having checked both open and closed issues at https://github.com/opnsense/core/issues?q=is%3Aissue

**Describe the bug**

OPNsense v25.1.7 host aliases no longer working:

**To Reproduce**

1. Upgrade to 25.1.7
2. Create a DNSmasq host entry with a MAC, an IPv4, a name and an alias.
3. Immediately after creation, try to resolve the alias via DNS.
4. There is no IP answer.

Obviously, the DNS entry is written only upon a DHCP lease being given out. So, if the host with that IP is statically configured, it cannot get DNS resolution any more. This worked before 25.1.7.

**Expected result**

A static reservation DNS name and its aliases should be resolved, no matter if a DHCP lease has been handed out.

**Describe alternatives you considered**

None.

**Screenshots**

None.

**Relevant log files**

None.

**Additional context**

See https://forum.opnsense.org/index.php?topic=47274.0

As it seems, /var/etc/dnsmasq-hosts is not filled any more, so all the aliases are being ignored.
This applies to any static reservation with a MAC or a client ID. They are not being written any more in order to allow for partial IPv6 addresses like "::3", because otherwise, their DNS names cannot be registered once their 64 bit prefix is determined.

**Environment**

Software version used and hardware type if relevant, e.g.:

OPNsense 25.1.7 (amd64).

fichtner:
In the .7 release DHCP leases are not written into dnsmasq-hosts. They only resolve when the lease is served now. Otherwise Dnsmasq breaks on handing out IPv6 constructor leases.

meyergru:
But this breaks static IPv4 reservations with aliases completely.

meyergru:
The are not offline. Such "reservations" sometimes are not even used via DHCP at all, like for statically configured web servers, where the only purpose of these entries is to have internal DNS aliases like 'www.internal', not have DHCP assign them an IP.

meyergru:
[...]

PR #8698, opened 2025-05-20
dnsmasq: Set aliases of dhcp-host as cname directive

Monviech:
Fixes: https://github.com/opnsense/core/issues/8694

A minimal suggestion for a fix. Will need validation for the cnames and maybe some UX.


Monviech:
It needs 2 validations:

- cnames must be unique
- cnames must not overlap with dhcp-host hostnames

In the UX:
- Hide the domain field if hwaddr or client_id is set, or validate that it should not be filled out
- Change the description of aliases (FQDN is not applicable when using cnames in dhcp-host)

swhite2:
Looks ok, yes, but I'll back off the review as the validations seem quite important. Perhaps https://github.com/opnsense/core/issues/8694#issuecomment-2893201704 can be included here as well.

PR #8699, opened 2025-05-20
dnsmasq: Create static host entries for all except partial IPv6 addresses

Monviech:
Fixes: https://github.com/opnsense/core/issues/8694

Another possible minimal fix for the issue referred to in that ticket.

### DIFF
--- a/src/etc/inc/plugins.inc.d/dnsmasq.inc
+++ b/src/etc/inc/plugins.inc.d/dnsmasq.inc
@@ -179,13 +179,19 @@ function _dnsmasq_add_host_entries()
         $dnsmasqcfg['hosts'] = [];
     }
 
+    /**
+    * XXX: We want to create these entries for dhcp-host as well.
+    *      Hostnames must be resolvable regardless of their DHCP lease status.
+    *      This overrides the dhcp-host behavior, as hostnames of leases will not be registered if they appear in this file.
+    *      One edge case are partial IPv6 addresses (::1) as they must be constructed by dnsmasq internally.
+    */
     foreach ($dnsmasqcfg['hosts'] as $host) {
-        /* do not create entries for dhcp-host types */
-        if (!empty($host['client_id']) || !empty($host['hwaddr'])) {
-            continue;
-        }
         /* The host.ip field supports multiple IPv4 and IPv6 addresses */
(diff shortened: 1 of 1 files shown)


fichtner (commented):
If we do this:

https://github.com/opnsense/core/blob/dabbc65045e35466f81310b8791d50c0f8641e4d/src/opnsense/service/templates/OPNsense/Dnsmasq/dnsmasq.conf#L184-L189

but do something else here we're just asking for future trouble

AdSchellevis:
@fichtner these are two different functions of the same service, you are allowed to feed static entries, as long as you can "guarantee" them to be static. Which is fine for normal ipv4 and ipv4 addresses, but when they are dynamic, you simply don't know which address they will receive when being requested.  I don't expect future trouble on these particular entries, we might miss some validations in the model at some point, but that's more about mixing types (something that can be fixed later when needed).

`_("not giving name %s to the DHCP lease of %s because the name exists in %s with address %s")`  is merely a warning about the fact that one host can not be bound twice (which would lead to round robin behavior from the clients perspective)

fichtner:
We should just make this explicit with a checkbox, check the boolean for dhcp-host inclusion in the dnsmasq.conf and add a migration excluding "::" prefixed entries by default. I know it's work but we will not have to come back to this later and can validate in all directions to appease dnsmasq's internal validation.

Or we park this and I'll simply push a hotfix revert of the offending commit in 25.1.7.

AdSchellevis:
Which problem are we trying to solve with a checkbox?

src/etc/inc/plugins.inc.d/dnsmasq.inc
```diff
@@ -179,13 +179,19 @@ function _dnsmasq_add_host_entries()
         $dnsmasqcfg['hosts'] = [];
     }
 
+    /**
+    * XXX: We want to create these entries for dhcp-host as well.
+    *      Hostnames must be resolvable regardless of their DHCP lease status.
+    *      This overrides the dhcp-host behavior, as hostnames of leases will not be registered if they appear in this file.
+    *      One edge case are partial IPv6 addresses (::1) as they must be constructed by dnsmasq internally.
+    */
```
fichtner, line 187: no need for this

fichtner (approved):

PR #8706, opened 2025-05-21
Use CIDR notation for specifying masks to dnctl.

tangrs:
Using the same format for both IPv4 and IPv6 mask specifications makes the template easier to understand at a glance.

### DIFF
--- a/src/opnsense/service/templates/OPNsense/Shaper/dnctl.conf
+++ b/src/opnsense/service/templates/OPNsense/Shaper/dnctl.conf
@@ -7,8 +7,8 @@ pipe {{ pipe.number }} config bw {{ pipe.bandwidth }}{{ pipe.bandwidthMetric }}/
  if pipe.queue %} queue {{ pipe.queue }}{%
  if pipe.queueMetric != 'slots' %}{{pipe.queueMetric}}{% endif %}{% endif
  %}{% if pipe.buckets %} buckets {{ pipe.buckets }}{% endif
- %}{% if pipe.mask in ['src-ip', 'dst-ip'] %} mask {{ pipe.mask }} 0xffffffff {% endif
- %}{% if pipe.mask in ['src-ip6', 'dst-ip6'] %} mask {{ pipe.mask }} 128 {% endif %}{%
+ %}{% if pipe.mask in ['src-ip', 'dst-ip'] %} mask {{ pipe.mask }} /32 {% endif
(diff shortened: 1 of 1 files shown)


src/opnsense/service/templates/OPNsense/Shaper/dnctl.conf
```diff
@@ -7,8 +7,8 @@ pipe {{ pipe.number }} config bw {{ pipe.bandwidth }}{{ pipe.bandwidthMetric }}/
  if pipe.queue %} queue {{ pipe.queue }}{%
  if pipe.queueMetric != 'slots' %}{{pipe.queueMetric}}{% endif %}{% endif
  %}{% if pipe.buckets %} buckets {{ pipe.buckets }}{% endif
- %}{% if pipe.mask in ['src-ip', 'dst-ip'] %} mask {{ pipe.mask }} 0xffffffff {% endif
- %}{% if pipe.mask in ['src-ip6', 'dst-ip6'] %} mask {{ pipe.mask }} 128 {% endif %}{%
+ %}{% if pipe.mask in ['src-ip', 'dst-ip'] %} mask {{ pipe.mask }} /32 {% endif
+ %}{% if pipe.mask in ['src-ip6', 'dst-ip6'] %} mask {{ pipe.mask }} /128 {% endif %}{%
```
AdSchellevis, line 11: does this mean the previous change (128 instead of /128) as actually incorrect?

tangrs: No, I don't believe so. dnctl accepts mask specifications in many different formats, and both produce the same results when I tested the dnctl command manually.

`128` triggers the following code path through [sbin/ipfw/dummynet.c:1556](https://github.com/freebsd/freebsd-src/blob/c3c816173d25fcb908a94c34b75fca98c5860f18/sbin/ipfw/dummynet.c#L1556), while `/128` triggers [sbin/ipfw/dummynet.c:1552](https://github.com/freebsd/freebsd-src/blob/c3c816173d25fcb908a94c34b75fca98c5860f18/sbin/ipfw/dummynet.c#L1552). At the end of the if/else block, `a` is always set to `128`, thus no net change.

AdSchellevis: @tangrs sounds logical, thanks, I'll merge this one.

Issue #8708, opened 2025-05-21
Dnsmasq: prevent client timeouts by returning NODATA for unknown hostnames.


swhite2:
**Important notices**

Before you add a new report, we ask you kindly to acknowledge the following:

- [x] I have read the contributing guide lines at https://github.com/opnsense/core/blob/master/CONTRIBUTING.md
- [x] I am convinced that my issue is new after having checked both open and closed issues at https://github.com/opnsense/core/issues?q=is%3Aissue

**Is your feature request related to a problem? Please describe.**

https://forum.opnsense.org/index.php?topic=47278.0 describes a typical Windows client "issue", in which manual queries done with nslookup without an FQDN dot suffixed, will result in nslookup suffixing the requested host/domain name with the DHCP DNS domain (option 15, default for both Dnsmasq and ISC set to the system domain name). It is unclear to me if this is a larger client issue that causes sluggish DNS resolution behavior in general, but with this specific (incorrect) test case, DNS resolution will take multiple seconds, as the client needs to figure out that it isn't getting a response for the suffixed request before it starts trying the original request. 

The above behavior has not changed over our releases, but since our documentation now recommends an example setup with dynamic DNS registration using Dnsmasq, any such requests forwarded by Unbound to Dnsmasq for the internal domain that are not known within Dnsmasq, will try to use `/etc/resolv.conf` to forward those internal queries. This means it may either use Unbound again, looping back around, or use one of the configured system nameservers in System -> Settings -> General. In all cases it will produce a timeout, as no such domain exists.

To be clear: Normal DNS usage requires clients to terminate a domain to signify the FQDN, this can easily be verified by doing a packet capture on port 53 on the LAN side, where each request contains a "." at the end of each domain.

**Describe the solution you like**

It would be practical for Dnsmasq to either:

- Act as both DHCP and DNS server for static/dynamic (known) entries, but disabling forwarding entirely, or
- always return `NODATA` for hostnames (but within the known domain) that are not known to Dnsmasq, instead of trying to forward them.

Both solutions will prevent a timeout.

**Describe alternatives you considered**

N/A

**Additional context**

N/A

**Logs showing use case**

The following logs are from: https://forum.opnsense.org/index.php?topic=47278.msg237815#msg237815

```
2025-05-19T13:53:40-04:00    Informational    unbound    [5329:2] info: 192.168.30.2 opnsense.org. AAAA IN NOERROR 0.081485 0 58   
[...]

PR #8709, opened 2025-05-21
dnsmasq: Add no-resolv option to prevent use of system defined dns servers

Monviech:
Fixes: https://github.com/opnsense/core/issues/8708

### DIFF
--- a/src/opnsense/mvc/app/controllers/OPNsense/Dnsmasq/forms/general.xml
+++ b/src/opnsense/mvc/app/controllers/OPNsense/Dnsmasq/forms/general.xml
@@ -104,7 +104,7 @@
         <id>dnsmasq.no_resolv</id>
         <label>Do not forward to system defined DNS servers</label>
         <type>checkbox</type>
-        <help>If this option is set, DNS forwarding to system nameservers (defined in System: General Setup: DNS Servers) will be disabled. Upstream servers defined in Services: Dnsmasq DNS & DHCP: Domains will still be used. This option is recommended when Unbound forwards local domain queries to Dnsmasq, so that all queries terminate without further lookups if they are unknown.</help>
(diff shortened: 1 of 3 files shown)


swhite2:
Couple of points:

- It's not clear that "upstream servers from the dnsmasq configuration" refer to the domain overrides.
- Perhaps the description can hint that this is **recommended** in scenarios where dnsmasq shouldn't be the active dns server.
- Is there no way for Dnsmasq to actively return something if no record is found (within a specific domain)?

Note: this should be in the docs for the default unbound + dnsmasq setup

Monviech:
I think the main issue why dnsmasq tries to forward can be seen in the original issue:

`opnsense.org.h1.home.arpa.` is queried.

Dnsmasq knows `h1.home.arpa.`, it does not know `org.h1.home.arpa.`.

So it asks the upstream server 127.0.0.1#53 (which is Unbound) for `opnsense.org.h1.home.arpa.` because it is in no domain dnsmasq has registered itself.

I do not know if NODATA can be returned.

src/opnsense/mvc/app/controllers/OPNsense/Dnsmasq/forms/general.xml
```diff
@@ -100,6 +100,12 @@
         <type>checkbox</type>
         <help>If this option is set, we will not forward A or AAAA queries for plain names, without dots or domain parts, to upstream name servers. If the name is not known from /etc/hosts or DHCP then a "not found" answer is returned.</help>
     </field>
+    <field>
+        <id>dnsmasq.no_resolv</id>
+        <label>Do not forward to system defined DNS servers</label>
+        <type>checkbox</type>
+        <help>If this option is set, do not use (System: General Setup: DNS Servers). Only use upstream servers defined in (Services: Dnsmasq DNS & DHCP: Domains). This option is recommended when Unbound forwards local domain queries to Dnsmasq, so that all queries terminate without further lookups if they are unknown.</help>
```
swhite2, line 107:
```suggestion
        <help>If this option is set, DNS forwarding to system nameservers (defined in System: General Setup: DNS Servers) will be disabled. Upstream servers defined in Services: Dnsmasq DNS & DHCP: Domains will still be used. This option is recommended when Unbound forwards local domain queries to Dnsmasq, so that all queries terminate without further lookups if they are unknown.</help>
```

swhite2 (approved):

PR #8710, opened 2025-05-22
dnsmasq: When dhcp-fqdn is active, set all dhcp domains as local

Monviech:
to prevent queries to upstream servers, this makes dnsmasq solely responsible for them. Otherwise it could accidentally ask its upstream servers if it acts as a resolver at the same time.

```
-S, --local, --server=[/[<domain>]/[domain/]][<server>[#<port>]][@<interface>][@<source-ip>[#<port>]]
Also permitted is a -S flag which gives a domain but no IP address; this tells dnsmasq that a domain is local and it may answer queries from /etc/hosts or DHCP but should never forward queries on that domain to any upstream servers. --local is a synonym for --server to make configuration files clearer in this case.
```

Log file example:

```
2025-05-22T07:05:26Informationaldnsmasqusing only locally-known addresses for ad.pischem.com
2025-05-22T07:05:26Informationaldnsmasqusing only locally-known addresses for test.pischem.com
2025-05-22T07:05:26Informationaldnsmasqusing only locally-known addresses for 3.test.pischem.com
2025-05-22T07:05:26Informationaldnsmasqusing only locally-known addresses for 4.test.pischem.com
```

Fixes: https://github.com/opnsense/core/issues/8708

### DIFF
--- a/src/opnsense/mvc/app/models/OPNsense/Dnsmasq/Dnsmasq.xml
+++ b/src/opnsense/mvc/app/models/OPNsense/Dnsmasq/Dnsmasq.xml
@@ -1,6 +1,6 @@
 <model>
     <mount>/dnsmasq</mount>
-    <version>1.0.5</version>
+    <version>1.0.6</version>
     <items>
         <enable type="BooleanField"/>
         <regdhcp type="BooleanField"/>
@@ -47,7 +47,10 @@
             <no_interface type="InterfaceField">
                 <Multiple>Y</Multiple>
             </no_interface>
-            <fqdn type="BooleanField"/>
+            <fqdn type="BooleanField">
+                <Required>Y</Required>
+                <Default>1</Default>
+            </fqdn>
             <domain type="HostnameField">
                 <IsDNSName>Y</IsDNSName>
                 <IpAllowed>N</IpAllowed>
--- a/src/opnsense/service/templates/OPNsense/Dnsmasq/dnsmasq.conf
+++ b/src/opnsense/service/templates/OPNsense/Dnsmasq/dnsmasq.conf
@@ -32,6 +32,17 @@ dhcp-lease-max={{dnsmasq.dhcp.lease_max}}
 {% if dnsmasq.dhcp.fqdn == '1' %}
 dhcp-fqdn
 domain={{dnsmasq.dhcp.domain|default(system.domain)}}
+# This tells dnsmasq that a domain is local and it may answer queries from /etc/hosts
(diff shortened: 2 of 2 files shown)


AdSchellevis (approved):

src/opnsense/mvc/app/models/OPNsense/Dnsmasq/Dnsmasq.xml
```diff
@@ -47,7 +47,10 @@
             <no_interface type="InterfaceField">
                 <Multiple>Y</Multiple>
             </no_interface>
-            <fqdn type="BooleanField"/>
+            <fqdn type="BooleanField">
+                <Required>Y</Required>
+                <Default>1</Default>
```
swhite2, line 52: @AdSchellevis  @Monviech  Any apply without an explicit user toggle on the option will have set this value to "0". Should we flip this with a migration?

fichtner:
not sure this will cover all the complexity in a real world setup. basically this tries to prevents DNS issues when Dnsmasq serves its own DHCP, no? If yes, adding the options verbatim as in the dnsmasq docs and adding more glue to fix side effects will lead to more edge cases from a user perspective. Maybe the setting could be more use-case tailored or latched on to enabled DHCP (when DNS is also enabled) automatically?

Monviech:
We can check:
- if DNS is enabled if the port is not 0.

I would imply DHCP is enabled if ranges are configured because thats what we collect here.

fichtner:
something like that, yes. people are definitely scared by DHCP complexity (and bitten by it)

src/opnsense/service/templates/OPNsense/Dnsmasq/dnsmasq.conf
```diff
@@ -29,9 +29,20 @@ no-dhcp-interface={{helpers.physical_interfaces(dnsmasq.dhcp.no_interface.split(
 dhcp-lease-max={{dnsmasq.dhcp.lease_max}}
 {% endif %}
 
-{% if dnsmasq.dhcp.fqdn == '1' %}
+{% if helpers.exists('dnsmasq.dhcp_ranges') and dnsmasq.dhcp.fqdn == '1' %}
```
fichtner, line 32: well this looks a bit odd now. you always want to enable it, move the first conditional into the loop below...

fichtner: (the idea mentioned played on the fact that maybe you don't want to offer dhcp.fqdn setting at all)

Issue #8703, opened 2025-05-21
Dnsmasq DHCP Reservations: Permit empty "IP address" field


skybob22:
**Important notices**

Our forum is located at https://forum.opnsense.org , please consider joining discussions there in stead of using GitHub for these matters.
[Current thread](https://forum.opnsense.org/index.php?topic=47296.0): Was directed to open an issue

Before you ask a new question, we ask you kindly to acknowledge the following:

- [x] I have read the contributing guide lines at https://github.com/opnsense/core/blob/master/CONTRIBUTING.md
- [x] I am convinced that my issue is new after having checked both open and closed issues at https://github.com/opnsense/core/issues?q=is%3Aissue

**Describe the bug**
Trying to add a reservation to Dnsmasq, I am unable to create an entry consisting of only a hardware (MAC) address and hostname.

This is something that Dnsmasq supports, but is not permitted by the OPNSense web interface

**To Reproduce**
1. Log in to OPNSense
2. Navigate to `Services ->  Dnsmasq DNS & DHCP -> Hosts`
3. Use `+` to add new host entry
4. Attempt to add entry while providing data for only the `Host` and `Hardware addresses` field

Unable to add entry due to message "A value is required" for `IP addresses` field

**Expected behavior**
Adds host entry to give client specific hostname, while getting IP from DHCP pool

**Describe alternatives you considered**
Not using Dnsmasq and continuing to use ISC v4, however that is being deprecated in favor of Dnsmasq for small/medium networks
Kea is impractical as it lacks certain features that ISC/Dnsmasq have

**Additional context**
This seems fairly similar to issue #8247 regarding the same kind of "hostname-only" configuration for Kea DHCP which raised concerns about DNS/hostname lookups due to Unbound potentially being unaware of leases that Kea handed out.

I would imagine that in this context it should be less of a concern as Dnsmasq does both DHCP and DNS where it would be aware of its own current leases. Even if Dnsmasq is not used as the primary DNS resolver, Unbound can be configured to forward lookup requests to it. ([As detailed in the OPNsense docs](https://docs.opnsense.org/manual/dnsmasq.html#dhcpv4-with-dns-registration))

**Environment**
OPNsense 25.1.7-amd64
Intel N6005 CPU

PR #8713, opened 2025-05-22
dnsmasq: Allow either empty IP or empty hostname for dhcp-host entries.

Monviech:
Fixes: https://github.com/opnsense/core/issues/8703

Please take a good look at the validation.

Please improve it if necessary, and test it.

I have tested it, and generated the correct configuration file with a wide range of inputs, but this feels like a delicate change that needs some thoughts, also if we want it.


Monviech:
@AdSchellevis Thank your the review. I understand the intention of the required changes. If you have time, could you fix these and merge it in? Thank you :)

AdSchellevis (approved):

PR #8714, opened 2025-05-24
dhcp/dhcrelay: Make dhcrelay carp-aware

AndyX90:
This commit adds carp-awareness to the dhcrelay service.

### DIFF
--- a/src/etc/inc/plugins.inc.d/dhcrelay.inc
+++ b/src/etc/inc/plugins.inc.d/dhcrelay.inc
@@ -26,6 +26,12 @@
  * POSSIBILITY OF SUCH DAMAGE.
  */
 
+function dhcrelay_carp_enabled()
+{
+    $model = new \OPNsense\DHCRelay\DHCRelay();
+    return (string)$model->general->enablecarp == '1';
+}
+
 function dhcrelay_configure()
 {
     return [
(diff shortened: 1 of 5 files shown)


Monviech:
Wouldn't a DHCP Discover broadcast be received by any running dhcrelay instance in the same broadcast domain?

AdSchellevis:
Totally my mistake, had `relayd` in my head while reading the PR,  for dhcp relay there might be scenario's where you don't want to forward the discoveries indeed.

AndyX90:
Actually this is ready and working if you only consider the carp functionality.
The case which is not covered atm is if you are on carp backup and apply service settings which calls reconfigure.
Then it would also start on the backup.

AdSchellevis:
it might be good to fix the startup on backup first, but further than that I have no objections.

AndyX90:
@AdSchellevis I think this is the better approach, as we don't have interfaces to depend on for a single service.
It now starts only on configured interfaces which have CARP vips and are in MASTER state.
Interfaces where no vips are configured are untouched and start anyway.

Monviech:
https://forum.opnsense.org/index.php?topic=48131

AdSchellevis:
@AndyX90 in case you want to test, this https://github.com/opnsense/core/commit/6e7e2a3225c3ad910b9bbfb04a539f25909d1e25 should offer carp vhid tracking per relay, similar to openvpn and wireguard.

PR #8729, opened 2025-05-27
trust: check private key matches provided certificate data

swhite2:
Probably other spots that need checking, but putting this one up for discussion purposes


AdSchellevis (approved):
@swhite2 looks good to me, at a first glance I don't expect other entry points here as csr imports already validate their response. Maybe CA's could use a similar check, but this is likely less important anyway.

Issue #8731, opened 2025-05-27
dnsmasq: Fix wildcard handling in HostOverride and dhcp-host


Monviech:
**Important notices**

Before you add a new report, we ask you kindly to acknowledge the following:

- [x] I have read the contributing guide lines at https://github.com/opnsense/core/blob/master/CONTRIBUTING.md
- [x] I am convinced that my issue is new after having checked both open and closed issues at https://github.com/opnsense/core/issues?q=is%3Aissue

**Is your feature request related to a problem? Please describe.**

A Host Override can be:

`*.example.com`

But a dhcp-host cannot be:

`dhcp-host=fe:fe:fe:fe:fe:fe,192.168.1.1,*,86400`

`2025-05-27T15:08:39Criticaldnsmasqbad DHCP host name at line 67 of /usr/local/etc/dnsmasq.conf`

**Describe the solution you like**

Add validation to prevent the use of "*" as hostname when a dhcp-host is created, but continue to allow it for normal Host Overrides.

**Describe alternatives you considered**

None

**Additional context**

None

AdSchellevis:
@Monviech the asterisk was allowed by mistake (https://github.com/opnsense/core/commit/1c52cb5a596a95e1ebd97d48eb9839083a25e6a3), same issue as https://github.com/opnsense/core/issues/8733

The php validation is pretty useless in this case, but since RFC2181 allows almost anything, excluding the real harmful characters might be the best option for now.

Monviech:
@AdSchellevis This here is finished, tested and pretty useful.

Though I can close it if we do not want this.

PR #8734, opened 2025-05-28
dnsmasq: Fix wildcard host handling

Monviech:
Fixes: https://github.com/opnsense/core/issues/8731


Monviech:
Still have to check one more thing so draft until later.

Issue #8738, opened 2025-05-29
Kea DHCP: Missing validation of "static routes" input field


meyergru:
**Important notices**

Before you add a new report, we ask you kindly to acknowledge the following:

- [X] I have read the contributing guide lines at https://github.com/opnsense/core/blob/master/CONTRIBUTING.md
- [X] I am convinced that my issue is new after having checked both open and closed issues at https://github.com/opnsense/core/issues?q=is%3Aissue

**Describe the bug**

The input for Kea DHCP IPv4 subnets feature a "static route" input, which is not validated.

You can enter an IP address on its own there, which should be at least a pair of (destination, ip), according to RFC2132.
The help text also says so. If you do that, Kea will stil start, but produce defective DHCP answers, which cannot be used by Windows 11 clients. They disregard the whole DHCP offer, which effectively renders Kea useless, without the user knowing.

**To Reproduce**

1. Services: Kea DHCP: Kea DHCPv4 -> Subnets
2. Enter an IP like 1.2.3.4 into a subnet's "Static routes" input
3. Save the entry
4. Wait for leases to be prolonged.
5. Windows clients will no longer pick up leases. 

**Expected behavior**

The input field should be validated according to RFC2132, prohibiting such inputs.

**Describe alternatives you considered**

None.

**Screenshots**

None.

**Relevant log files**

A defective DHCP offer looks like this:

```
17:33:47.852214 IP (tos 0x10, ttl 128, id 0, offset 0, flags [DF], proto UDP (17), length 394)
    OPNsense.bootps > pcmg.mgsoft.bootpc: [udp sum ok] BOOTP/DHCP, Reply, length 366, xid 0x949cb5db, Flags [none] (0x0000)
          Your-IP pcmg.mgsoft
          Server-IP docker.mgsoft
          Client-Ethernet-Address 80:3f:5d:fa:xx:xx (oui Unknown)
          Vendor-rfc1048 Extensions
            Magic Cookie 0x63825363
            DHCP-Message (53), length 1: Offer
            Subnet-Mask (1), length 4: 255.255.255.0
            Default-Gateway (3), length 4: OPNsense
            Domain-Name-Server (6), length 4: OPNsense
            Hostname (12), length 4: "pcmg"
            Domain-Name (15), length 6: "mgsoft"
            Static-Route (33), length 4:
          trailing data length 4
            Lease-Time (51), length 4: 7200
            Server-ID (54), length 4: OPNsense
[...]

PR #8742, opened 2025-05-30
Kea static route GUI validation

meyergru:
This implements https://github.com/opnsense/core/issues/8738.

I know a regex is ugly, yet I have found no easy way to use NetworkField, because the list may only contain pairs of destination/router ips. Otherwise, clients will reject the whole DHCP offer.


meyergru:
I will try that later. Question: Do I need to / can I transform the output? In this case, it would be better if the single elements would be specified like 0.0.0.0,1.2.3.4 and grouped like in a NetworkField, but the output has to concatenate them together, separated by commas.

BTW: The "Unifi" option is a similar case where an IP must be transformed to a binary string.

meyergru:
@Monviech: I think this works now - at least for me, it does.

I have not really looked at migrations (if people specified multiple routes earlier, they must edit the innput manually) and I found nothing in the API as a blueprint for this, either. I think that the API code misses the options altogether, right?

Monviech (commented):
This is all I can add, custom field types are more in @AdSchellevis domain.

PR #8750, opened 2025-06-01
dnsmasq: Update help on multiple tags usage (all tags must match)

wrobelda:
This should be self-explanatory — I was confused with how tags would work, I somehow assumed any of the selected tags would match, but that's not the case, as all tags must match if more than one is selected, per dnsmasq documentation: 

```
Each configuration line containing one or more tag:<tag> constructs applies when all its tags exist on the request. That is:

Configuration tag:A applies to a request set:tagged A.

Configuration tag:B applies to a request set:tagged B.

Configuration tag:A+B doesn't apply to a request set:tagged A.

Configuration tag:A+B doesn't apply to a request set:tagged B.

Configurations tag:A+B, tag:A, tag:B apply to a request set:tagged A+B.
```

### DIFF
--- a/src/opnsense/mvc/app/controllers/OPNsense/Dnsmasq/forms/dialogDHCPboot.xml
+++ b/src/opnsense/mvc/app/controllers/OPNsense/Dnsmasq/forms/dialogDHCPboot.xml
@@ -9,7 +9,7 @@
         <id>boot.tag</id>
         <label>Tag</label>
         <type>select_multiple</type>
-        <help>Only offer this boot image to the clients matched by the given tag. Can be optionally combined with an interface tag.</help>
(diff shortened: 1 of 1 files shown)


AdSchellevis:
for the translations we might best use the same phrase for options and boot options, options already state:

https://github.com/opnsense/core/blob/9f421222296dff8f2f74da46da0c43864935caaa/src/opnsense/mvc/app/controllers/OPNsense/Dnsmasq/forms/dialogDHCPoption.xml#L31

Which should be clear as well.

AdSchellevis:
@wrobelda thanks!

wrobelda:
> Which should be clear as well.

Fair enough, I updated the string.

> @wrobelda thanks!

Thanks as well!

Issue #8670, opened 2025-05-16
WebUI automation/filter: Enable rule logging toggle via interactive icon


dstapa:
**Important notices**

Before you add a new report, we ask you kindly to acknowledge the following:

- [x ] I have read the contributing guide lines at https://github.com/opnsense/core/blob/master/CONTRIBUTING.md
- [x ] I am convinced that my issue is new after having checked both open and closed issues at https://github.com/opnsense/core/issues?q=is%3Aissue

**Describe the bug**

Automation rules in the new UI don't currently support quick-toggle of logging by clicking on the logging icon, as normal firewall rules do.

**To Reproduce**

Create an Automation filter rule and observe the non-interactive logging icon.

**Expected behavior**

Feature parity with other firewall rules to quick-toggle logging from the rules list view.

**Screenshots**

![Image](https://github.com/user-attachments/assets/d83201a5-0e28-4fb0-a17c-295c96ce0365)

**Environment**

OPNsense 25.1.6_4 (amd64).

PR #8757, opened 2025-06-02
automation/filter: Add toggle log command for parity with firewall/rules

Monviech:
Fixes: https://github.com/opnsense/core/issues/8670

### DIFF
--- a/src/opnsense/mvc/app/controllers/OPNsense/Firewall/Api/FilterController.php
+++ b/src/opnsense/mvc/app/controllers/OPNsense/Firewall/Api/FilterController.php
@@ -247,7 +247,7 @@ class FilterController extends FilterBaseController
 
         if ($node === null) {
             throw new UserException(
-                gettext("This rule is not managed with this component"),
+                gettext("Rule not found"),
                 gettext("Filter")
             );
         }


src/opnsense/mvc/app/controllers/OPNsense/Firewall/Api/FilterController.php
```diff
@@ -230,6 +230,35 @@ public function toggleRuleAction($uuid, $enabled = null)
...
+        foreach ($mdl->rules->rule->iterateItems() as $item) {
+            if ((string)$item->getAttribute('uuid') === $uuid) {
+                $node = $item;
+                break;
+            }
+        }
+
+        if ($node === null) {
+            throw new UserException(
+                gettext("This rule is not managed with this component"),
```
swhite2: this is an assumption, probably easier to just state rule not found

swhite2 (approved):
Looks good, minus the `validateFullModel` as discussed and the nitpick

src/opnsense/mvc/app/controllers/OPNsense/Firewall/Api/FilterController.php
Monviech: thanks, changed it

Monviech:
> Looks good, minus the `validateFullModel` as discussed and the nitpick

I'll leave this with the default params for now as its the safest. If we want to disable validation we can always add this later.

https://github.com/opnsense/core/blob/f0c32264da4b7a6f957b25119c4d40d4581faa0f/src/opnsense/mvc/app/models/OPNsense/Base/BaseModel.php#L789

PR #8764, opened 2025-06-04
interfaces: drop redundant updates in rtsold_resolvconf.sh

### LINKED ISSUE #8759: IPv6 RA processing results in excessive load and pointless churn
**Important notices**

**Describe the bug**

I have a WAN interface (ax0) configured using SLAAC, with DHCPv6 only for prefix delegation. My ISP (Init7) sends an IPv6 RA (including DNS servers and prefix info) every 9-10s. As best I can tell, this causes the following:
 1. rtsold runs `/var/etc/rtsold_script.sh`, which in turn sends SIGHUP (twice?) to dhcp6c
 2. rtsold runs `/usr/local/opnsense/scripts/interfaces/rtsold_resolvconf.sh`, which:
   a. deletes and re-adds all the ipv6 nameservers 
   b. invokes `configctl -d interface newipv6` aka `/usr/local/etc/rc.newwanipv6` which does a *lot* including `system_resolver_configure` which causes `/etc/hosts` and `/etc/resolv.conf` to get rewritten
 3. dnsmasq notices that /etc/resolv.conf changed, and noisily reconfigures itself

Since this happens frequently, yet every RA contains the *same* configuration information, this feels awfully wasteful, both in terms of CPU load and SSD write wear for all the log messages. Couldn't there be a check fairly early (I'd imagine even in rtsold!) that discards redundant updates?

**To Reproduce**

1. Sign up with init7.net :)
2. Configure the WAN interface with DHCP + DHCPv6 (basic config, with "Request prefix only" checked and 
Prefix delegation size of 48)
3. Enable dnsmasq + unbound following https://docs.opnsense.org/manual/dnsmasq.html#dhcpv4-with-dns-registration
4. Inspect log files and top

**Expected behavior**

Frequent RAs do not induce log entries and system load.

I could hack a short-circuit to avoid most of this into rtsold_resolvconf.sh myself, but I'd probably break something.

**Relevant log files**

https://forum.opnsense.org/index.php?topic=47469.0 was my initial debugging, and shows some of the log spam this induces. I'm happy to gather more info if needed.

**Environment**

OPNsense 25.1.7_4-amd64 on DEC740

0xabu:
This reduces system churn (e.g. resolv.conf rewrites and dnsmasq reloads) from WAN RAs.

Fixes #8759

### DIFF
--- a/src/opnsense/scripts/interfaces/rtsold_resolvconf.sh
+++ b/src/opnsense/scripts/interfaces/rtsold_resolvconf.sh
@@ -65,6 +65,19 @@ if [ "${1}" = "-a" ]; then
 		fi
 	done
 
+	# compare to the existing configuration and drop redundant updates
+	for value in $(/usr/local/sbin/ifctl -i ${ifname} -6n); do
+		cur_nameservers="${cur_nameservers} -a ${value}"
+	done
+	for value in $(/usr/local/sbin/ifctl -i ${ifname} -6s); do
+		cur_searchlist="${cur_searchlist} -a ${value}"
+	done
+	cur_rasrca="$(/usr/local/sbin/ifctl -i ${ifname} -6r)"
+	if [ "${nameservers}" = "${cur_nameservers}" -a "${searchlist}" = "${cur_searchlist}" -a "${rasrca}" = "${cur_rasrca}" ]; then
+		echo "Redundant update ignored."
+		exit 0
+	fi
+
 	/usr/local/sbin/ifctl -i ${ifname} -6nd ${nameservers}
 	/usr/local/sbin/ifctl -i ${ifname} -6sd ${searchlist}
 	/usr/local/sbin/ifctl -i ${ifname} -6rd -a ${rasrca}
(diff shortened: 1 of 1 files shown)


fichtner (commented):
ifctl is relatively helpful here, perhaps we can reduce the code to do an update instead of delete+add asking ifctl to return false if the update was not needed so we can steer the DNS reload.

fichtner:
@0xabu do you want to take a shot at it using 980d0c0 ?

0xabu:
@fichtner ok, looks straightforward, but just to double check: this version ifctl should be safe to drop onto a 25.1 router?

fichtner:
@0xabu I tested what I could think of but we usually let it sit on master for a release to wait for edge case reports and to have time for 1-2 more internal audits while doing releases and release related testing.. long story short: probably a "no" today

0xabu:
Updated to use the new `ifctl -u` flag. This seems to work for me.

fichtner (approved):

fichtner:
Merged, thanks! I will refactor a bit more as the
else case will benefit as well.

PR #8769, opened 2025-06-07
Unbound: Add "ISC" to "Register DHCP Static Mappings"

MeganerdNL:
To sync it with the docs https://docs.opnsense.org/manual/unbound.html It's not entirely clear that this is for ISC DHCP only.


AdSchellevis:
I think it's exactly the other way around, the docs should remove the ISC part (https://github.com/opnsense/core/commit/139a3add4bb4360e2dda8f3251283e0173b0f980)

MeganerdNL:
Ah, that was my misunderstanding how this works then. I see now this is for Kea too. Thanks.

AdSchellevis:
a PR for the docs would be highly appreciated :)

MeganerdNL:
> a PR for the docs would be highly appreciated :)

I will do this.

PR #8778, opened 2025-06-09
Fix filename in kea dhcp6 manual help.

sharmay:
Wrong filename used in kea dhcp6 Manual config file help.


AdSchellevis:
@sharmay nice catch, thanks!

PR #8790, opened 2025-06-12
dnsmasq: Allow host wildcards in domain overrides again

Monviech:
Regression from: https://github.com/opnsense/core/commit/1c52cb5a596a95e1ebd97d48eb9839083a25e6a3


fichtner:
Is there a report for this?

EDIT: sorry, we are upstream ;)

Monviech:
@fichtner Im the reporter because I need this in my setup and it stopped working. Otherwise you cannot create a wildcard domain entry fir query forwarding (aka forward all queries).

Issue #4145, opened 2020-05-30
Wildcard Support in Firewall Alias - Type Host(s)


Mer-idium:
**Important notices**
Before you add a new report, we ask you kindly to acknowledge the following:

[-] I have read the contributing guide lines at https://github.com/opnsense/core/blob/master/CONTRIBUTING.md

[-] I have searched the existing issues and I'm convinced that mine is new.

Hi All,

Is it possible to add the feature to include wildcard support in Firewall Alias groups? At the moment rather than going *.somedomain.com to white list all subdomains i need to manually add each one like so: first.somedomain.com, second.somedomain.com etc...

This can get irritating when i have many subdomains to add which are near identical. In some instances the only way to obtain all domains is to run through Unbound logs to find the single sub domain i missed which drags out the process. 

Thanks

AdSchellevis:
long story short, you can't (easily) request all existing records for a (top) level domain. https://stackoverflow.com/questions/19322962/how-can-i-list-all-dns-records

Our aliases resolve on intervals and capture the associated addresses in a table, which only works if you can determine addresses up front.

You can add domain overrides in unbound and prevent the clients from resolving, which is what most of the blacklist solutions use as well. A more complex solution would be to capture unbounds responses and add those to a table, this is however outside the scope of our project.

pfeiflelukas:
If the issue was for blacklisting i wouldn't mind as i would use unbound. This is more for traffic direction to negate a proxy and then pass the traffic through a VPN.

agh1467:
This is becoming a more common feature in firewalls. Several vendors already have support for it. As cloud usage expands, DNS names are becoming more dynamic by nature.

Here are several examples of the feature from other vendors:
https://www.sonicwall.com/support/knowledge-base/dns-resolution-of-wildcard-fqdn-address-objects/170505295458240/
https://www.watchguard.com/help/docs/help-center/en-US/Content/en-US/Fireware/policies/fqdn_about_c.html
https://docs.fortinet.com/document/fortigate/6.2.0/cookbook/217973/using-wildcard-fqdn-addresses-in-firewall-policies

[...]

PR #8791, opened 2025-06-12
dnsmasq: Add ipset support

Monviech:
Fixes: https://github.com/opnsense/core/issues/8823

Implements the following:

```
--ipset=/<domain>[/<domain>...]/<ipset>[,<ipset>...]
Places the resolved IP addresses of queries for one or more domains in the specified Netfilter IP set.
```

There have been quite some user reports who use this successfully in the past on github and the forum and it works with pf correctly.

This is the set I tested it with:

```
root@opn03:/usr/local/etc/dnsmasq.conf.d # cat ipset.conf
ipset=/google.com/dnsmasq_google
ipset=/google.de/dnsmasq_google
```

Result of alias:

```
root@opn03:/usr/local/etc/dnsmasq.conf.d # pfctl -t dnsmasq_google -T show
   142.251.36.174
   142.251.36.195
```

### DIFF
--- a/src/opnsense/mvc/app/models/OPNsense/Dnsmasq/Dnsmasq.xml
+++ b/src/opnsense/mvc/app/models/OPNsense/Dnsmasq/Dnsmasq.xml
@@ -140,7 +140,7 @@
                         <items>aliases.alias</items>
                         <display>name</display>
                         <filters>
-                            <type>/^External.*/</type>
+                            <type>/^[Ee]xternal.*/</type>
                             <name>/^(?!bogons$|bogonsv6$|virusprot$|sshlockout$|__captiveportal.*).*/</name>
                         </filters>
                     </aliases>


fichtner:
Interesting! Isn't this emulating what the hosts aliases do?

Also I want to mention https://forum.opnsense.org/index.php?topic=47582.0 as a side note, where "alias" is used as terminology already so maybe that should be renamed while at it to avoid more confusion.

Monviech:
@fichtner As far as I read in the past it can be used to create wildcard host entries, which you cannot do with a normal host alias.

So e.g. you could add *.cloudflare.com as ipset and it would track all IP addresses that clients request of subdomains under cloudflare.com.

https://github.com/opnsense/core/issues/4145

fichtner:
Ah ok nice

swhite2:
~~Has https://github.com/opnsense/core/issues/4145#issuecomment-1209391936 been considered?~~

~~> An apparent limitation of this dnsmasq functionality is that the TTL for DNS records is not considered, and IPs added to the set remain in the set forever. Used with a CDN service, the table may become huge. I'd guess it may have a negative impact on performance at some point depending on hardware, the number of records in the table, and the frequency and number of requests made by clients.~~

Edit: mentioned in the description

src/opnsense/mvc/app/models/OPNsense/Dnsmasq/Dnsmasq.php
```diff
@@ -112,6 +112,19 @@ public function performValidation($validateFullModel = false)
             }
         }
 
+        foreach ($this->domainoverrides->iterateItems() as $domain) {
+            if (!$validateFullModel && !$domain->isFieldChanged()) {
+                continue;
+            }
+            $key = $domain->__reference;
+
+            if ($domain->domain == '*') {
```
swhite2, line 133: How does dnsmasq respond to `.*` or `*.` inputs?

If we look at wildcard matching in Unbound (different functionality), we also do not allow `*.com` or any other TLD. Should we do the same here?

src/opnsense/mvc/app/controllers/OPNsense/Dnsmasq/forms/dialogDomainOverride.xml
```diff
@@ -33,6 +33,12 @@
         <type>text</type>
         <help>Source IP address for queries to the DNS server for the override domain. Best to leave empty</help>
     </field>
+    <field>
+        <id>domainoverride.ipset</id>
+        <label>Ipset</label>
+        <type>dropdown</type>
+        <help>Choose an "external (advanced)" type alias from "Firewall - Aliases". Whenever a client successfully resolves the domain in this override, the resolved IP addresses will be automatically added to the chosen alias. You can create multiple domain overrides to add multiple domains to the same alias. Adding a wildcard domain will also add all IP addresses of resolved subdomains. Please note that domain TTL is not evaluated; once an IP address is added, it will stay permanently, or until manually flushed in "Firewall - Diagnostics - Aliases".</help>
```
swhite2:
```suggestion
        <help>Choose an "external (advanced)" type alias from "Firewall - Aliases". Whenever a client successfully resolves the domain in this override, the resolved IP addresses will be automatically added to the chosen alias. You can create multiple domain overrides to add multiple domains to the same alias. Adding a wildcard domain will also add all IP addresses of resolved subdomains. Please note that DNS record TTL is not evaluated; once an IP address is added, it will stay permanently, or until manually flushed in "Firewall - Diagnostics - Aliases".</help>
```

Monviech:
@swhite2 - @AdSchellevis told me he wants to implement:

"pfctl -t xxx -T expire 86400"

This could prevent ipsets that are note quite as selective to grow indefinitely.

Though the usecase for this alias type should be more selective, as its for a very specific usecase.

swhite2:
> @swhite2 - @AdSchellevis told me he wants to implement:
> 
> "pfctl -t xxx -T expire 86400"
> 
> This could prevent ipsets that are note quite as selective to grow indefinitely.
> 
> Though the usecase for this alias type should be more selective, as its for a very specific usecase.

https://github.com/opnsense/core/issues/4145#issuecomment-1209648932 mentions:

> While pfctl can remove old entries (-T expire 86400 for over 24 hours old), these are not safe to remove. These "old" entries may have been recently served by dnsmasq even seconds ago. After dnsmasq ipset adds an IP address, the table timestamp will not be updated again on subsequent queries.

I might be interpreting this incorrectly, but I assume dnsmasq fully backs off on responsibility for the entry in the table after insertion?

src/opnsense/mvc/app/models/OPNsense/Dnsmasq/Dnsmasq.php
Monviech: Right now the validation in the model will only allow explicit "*". *. or *.com are not allowed.

I'm still in the process of finding out if "google.com" in an ipset already includes all subdomains below that, or not.

I'm still unsure any "*." wildcards are needed.

Monviech: Looks like *.google.com is not needed to add maps.google.com to the ipset alias:

```
root@opn03:/usr/local/etc/dnsmasq.conf.d # cat ipset.conf
ipset=/google.com/dnsmasq_google
```

```
root@opn03:/usr/local/etc/dnsmasq.conf.d # pfctl -t dnsmasq_google -T show
   142.251.36.174
   142.251.36.195
```

```
root@opn03:/usr/local/etc/dnsmasq.conf.d # dig @127.0.0.1 -p 53053 maps.google.com

; <<>> DiG 9.20.8 <<>> @127.0.0.1 -p 53053 maps.google.com
; (1 server found)
;; global options: +cmd
;; Got answer:
;; ->>HEADER<<- opcode: QUERY, status: NOERROR, id: 20019
;; flags: qr rd ra; QUERY: 1, ANSWER: 1, AUTHORITY: 0, ADDITIONAL: 1

;; OPT PSEUDOSECTION:
; EDNS: version: 0, flags:; udp: 1232
;; QUESTION SECTION:
;maps.google.com.		IN	A

;; ANSWER SECTION:
maps.google.com.	130	IN	A	142.251.36.206

;; Query time: 28 msec
;; SERVER: 127.0.0.1#53053(127.0.0.1) (UDP)
;; WHEN: Fri Jun 13 09:56:27 CEST 2025
;; MSG SIZE  rcvd: 60
```

```
root@opn03:/usr/local/etc/dnsmasq.conf.d # pfctl -t dnsmasq_google -T show
   142.251.36.174
   142.251.36.195
   142.251.36.206
```

AdSchellevis:
ok, funny, apparently Unbound's `ipset` option also supports `pf` since last year (https://github.com/NLnetLabs/unbound/blob/master/doc/Changelog , https://github.com/NLnetLabs/unbound/blob/master/doc/README.ipset.md)

ItsHarper:
I would recommend testing this feature with the latest unreleased Dnsmasq commits. [98189ff988](https://thekelleys.org.uk/gitweb/?p=dnsmasq.git;a=commit;h=98189ff988d01d48929057037060d8cb2b4a22a6) seems like a significant change to how this feature works, and if I'm right about the fact that it breaks building on FreeBSD (see [my forum post](https://forum.opnsense.org/index.php?topic=47594.0) ), it clearly wasn't implemented with much thought to how it will work on FreeBSD, much less has been tested there.

If this feature has been broken upstream, it seems much better to get ahead of that and make sure it's fixed before the next Dnsmasq release.

I think fixing the build should be as simple as restoring the deleted preprocessor checks, so testing this shouldn't be too hard.

ItsHarper:
It's interesting that dnsmasq's `add_to_ipset()` function has a parameter to make it remove entries rather than add them, but it's always set to 0. I wonder how much work it would be to make it auto-remove entries for expired leases.

ItsHarper:
> I wonder how much work it would be to make it auto-remove entries for expired leases.

I actually think it should be pretty straightforward. See my notes in this commit: https://github.com/ItsHarper/dnsmasq/commit/00cafb53f799d4943b3495ee95b90b5ccb56c727

I think it might actually be a mistake to add this feature without the auto-removal support. If auto-removal were to be implemented, I think it would become pretty important that Dnsmasq was the  _only_ modifier of any Aliases that it touches (because otherwise it might remove things it shouldn't), and that those Aliases had names prefixed with `dnsmasq-` or something. That way, if a user migrates from Dnsmasq to another DHCP server like Kea, it will become clear that their firewall rules are referencing something that only works in the old setup.

This seems like a case where it's way better to wait to ship until you can ship an actually coherent solution (I'm aware this PR is just a draft and not about to ship, I'm just sharing my thoughts).

src/opnsense/mvc/app/controllers/OPNsense/Dnsmasq/forms/dialogDomainOverride.xml
```diff
@@ -33,6 +33,12 @@
         <type>text</type>
         <help>Source IP address for queries to the DNS server for the override domain. Best to leave empty</help>
     </field>
+    <field>
+        <id>domainoverride.ipset</id>
+        <label>Ipset</label>
+        <type>dropdown</type>
+        <help>Choose an "external (advanced)" type alias from "Firewall - Aliases". Whenever a client successfully resolves the domain in this override, the resolved IP addresses will be automatically added to the chosen alias. You can create multiple domain overrides to add multiple domains to the same alias. Adding a wildcard domain will also add all IP addresses of resolved subdomains. Please note that DNS record TTL is not evaluated; once an IP address is added, it will stay permanently, or until manually flushed in "Firewall - Diagnostics - Aliases".</help>
```
ItsHarper: I think this should include a note that this feature should not be used to define security-critical firewall rules, as they can be bypassed merely by configuring a static IP address.

Monviech: How can it be bypassed via setting a static IP on a client. Its the domains IP addresses that get added to the alias. Usecase is allowing domains hosted in CDNs for example.

Monviech:
For the upstream related issued, I'll leave these up to @fichtner to sight as he has way more experience than me with them. Thank you @ItsHarper for looking into it so far.

src/opnsense/mvc/app/controllers/OPNsense/Dnsmasq/forms/dialogDomainOverride.xml
ItsHarper: I believe the IP address won't get added to the alias until the client actually makes a DHCP request and a lease is created. If the client uses a static IP address, the lease won't get created (and they can always try to set a static IP that's different from the one the DHCP server would give them).

This admittedly becomes even more of a concern if the IP address is automatically removed when the lease expires. 

Also, if you don't have auto-removal and you are only configuring a hostname assignment without assigning a fixed IP address to give out to the client, this feature can slowly keep adding one IP address after another to the firewall alias, which could cause both security and functionality issues depending on the rules referencing the alias.

Monviech: I think there is some sort of misunderstanding here. Check here how an IP address is added:

https://github.com/opnsense/core/pull/8791#discussion_r2144431744

src/opnsense/mvc/app/models/OPNsense/Dnsmasq/Dnsmasq.php
swhite2: Shouldn't we at least be able to differentiate between `*.google.com` and `google.com`? I might only be interested in `a.b.c`, but this then includes all other subdomains as well. Just leaving this here for reference.

Monviech: I guess if you are only interested in exactly ```google.com```, a normal host alias is the best choice?

Monviech:
> ok, funny, apparently Unbound's `ipset` option also supports `pf` since last year (https://github.com/NLnetLabs/unbound/blob/master/doc/Changelog , https://github.com/NLnetLabs/unbound/blob/master/doc/README.ipset.md)

@AdSchellevis 

https://github.com/opnsense/tools/pull/481
https://github.com/opnsense/ports/pull/231

The build works for me so I gonna try an implementation there too so dnsmasq and unbound both have ipset support.

Monviech:
@fichtner This one would be ready for master. I rebased it without the Unbound commits.

AdSchellevis (approved):

Issue #8789, opened 2025-06-12
Unable to delete Dnsmasq aliases


ItsHarper:
- [x] I have read the contributing guide lines at https://github.com/opnsense/core/blob/master/CONTRIBUTING.md
- [x] I am convinced that my issue is new after having checked both open and closed issues at https://github.com/opnsense/core/issues?q=is%3Aissue

### Describe the bug

In OPNsense `25.1.8_1`, once an alias is added to a Dnsmasq host configuration, it's impossible to remove it through the web UI.

I never tried to do this in previous versions.

### To reproduce

1. Log into a system that has Dnsmasq configured as the DHCP server
2. Go to `Services: Dnsmasq DNS & DHCP: Leases`
3. Click the plus button next to a dynamic lease
4. Put `tempAlias` in the `Aliases` field
5. Click `Save`
6. Click the edit button for the host configuration you just created
7. Click the `x` icon on the `tempAlias` tag
8. Click `Save`
9. Click `Apply` (it does the same thing whether you do this or not)
10. Click the edit button for the new host configuration again
11. See that `tempAlias` is still listed under `Aliases`

#### Expected behavior

The `tempAlias` alias should have been removed

### Describe alternatives you considered

None

### Environment

OPNsense 25.1.8_1 (amd64).
ODroid H4+

Monviech:
I can replicate it, I'll look into it.

PR #8792, opened 2025-06-13
dnsmasq: Allow AliasesField to be cleared

Monviech:
Fixes: https://github.com/opnsense/core/issues/8789


AdSchellevis:
better kill the endif above :)

Monviech:
okay I changed it to remove the *elseif

PR #8795, opened 2025-06-13
tabulator: backwards compatible split for stable/25.1

swhite2:
See first commit description for relevant cherry-picks. Second commit encapsulates all changes providing backwards compatibility.

### DIFF
--- a/src/opnsense/mvc/app/controllers/OPNsense/Base/ControllerBase.php
+++ b/src/opnsense/mvc/app/controllers/OPNsense/Base/ControllerBase.php
@@ -71,7 +71,6 @@ class ControllerBase extends ControllerRoot
           '/ui/js/opnsense_bootgrid_plugin.js',
 
           '/ui/js/tabulator.min.js',
-        //   '/ui/js/opnsense_bootgrid.js', /* include for 25.7 */
 
           // Bootstrap type ahead
           '/ui/js/bootstrap3-typeahead.min.js',


AdSchellevis (approved):
@swhite2 I've checked various forms and so far all seems to be working as before, when 25.7 is out, it might be practical to cleanup some of the `*Initialized` flags and move them to the jquery handle if possible (something like `$("#my-target").isInitialized()`), but that's not something we have to worry about now.

src/opnsense/mvc/app/controllers/OPNsense/Base/ControllerBase.php
```diff
@@ -68,13 +68,17 @@ protected function templateJSIncludes()
           '/ui/js/tokenize2.js',
           // Bootgrid (grid system from http://www.jquery-bootgrid.com/ )
           '/ui/js/jquery.bootgrid.js',
+          '/ui/js/opnsense_bootgrid_plugin.js',
+
+          '/ui/js/tabulator.min.js',
+        //   '/ui/js/opnsense_bootgrid.js', /* include for 25.7 */
```
fichtner: since stable/25.1 never meets stable/25.7 you can omit this

fichtner (approved):
nice thanks (only checked scope of backport)

swhite2:
> @swhite2 I've checked various forms and so far all seems to be working as before, when 25.7 is out, it might be practical to cleanup some of the `*Initialized` flags and move them to the jquery handle if possible (something like `$("#my-target").isInitialized()`), but that's not something we have to worry about now.

In most cases this can even be simplified without an initialized flag, but I'll leave it like this for now.

Issue #8834, opened 2025-06-20
Can't Download CSR


Zewwy:
**Important notices**
No existings Tickets from what I've found.

**Describe the bug**

Attempting to Download a CSR, does nothing.

**To Reproduce**

1. Go to System -> Trust -> Certificate -> + -> Type CSR -> Save
2. Click on the newly created CSR line item's Cloud Icon (The download icon)
3. Get prompt for what to download (PCKS,Prive Key,Cert)
4. Pick Cert -> Nothing Happens

**Expected behavior**

In step 3, you shouldn't even get a prompt for Type, as there is no full cert for the PCKS, The Private Key is useless at this point cause again no full cert is created, it just creates the Priv Key and await the signed Cert from the external CA.

Step 4, when picking Certificate you should get the CSR downloaded, but nothing happens.

**Describe alternatives you considered**

A clear and concise description of any alternative solutions or workaround you considered.

I've had this on 25.1.7_4 (prod system), as well as 25.1.9_2.. I even downloaded the latest ISO from the site, loaded it live on a VM, and same issue. Browser is Latest Edge on Windows 10/11

N/A, it's a silent fail. Open Dev Tools in Browser look at console.. NOTHING

If applicable, information from log files supporting your claim.

It's really easy to reproduce.

OPNsense hosted on ESXi 7.0.3

Software version used and hardware type if relevant, e.g.:

OPNsense 25.1.9_2-amd64
FreeBSD 14.2-RELEASE-p3
OpenSSL 3.0.16
Multiple hardware

Zewwy:
Thank you Bot! I edited the top of the template thus it doesn't meet the template need got it.. closing creating new one

Issue #8835, opened 2025-06-20
Download CSR not working


Zewwy:
**Important notices**

Before you add a new report, we ask you kindly to acknowledge the following:

- [x] I have read the contributing guide lines at https://github.com/opnsense/core/blob/master/CONTRIBUTING.md
- [x] I am convinced that my issue is new after having checked both open and closed issues at https://github.com/opnsense/core/issues?q=is%3Aissue

Important notices
No existings Tickets from what I've found.

Describe the bug

Attempting to Download a CSR, does nothing.

To Reproduce

Go to System -> Trust -> Certificate -> + -> Type CSR -> Save
Click on the newly created CSR line item's Cloud Icon (The download icon)
Get prompt for what to download (PCKS,Prive Key,Cert)
Pick Cert -> Nothing Happens
Expected behavior

In step 3, you shouldn't even get a prompt for Type, as there is no full cert for the PCKS, The Private Key is useless at this point cause again no full cert is created, it just creates the Priv Key and await the signed Cert from the external CA.

Step 4, when picking Certificate you should get the CSR downloaded, but nothing happens.

Describe alternatives you considered

A clear and concise description of any alternative solutions or workaround you considered.

I've had this on 25.1.7_4 (prod system), as well as 25.1.9_2.. I even downloaded the latest ISO from the site, loaded it live on a VM, and same issue. Browser is Latest Edge on Windows 10/11

N/A, it's a silent fail. Open Dev Tools in Browser look at console.. NOTHING

If applicable, information from log files supporting your claim.

It's really easy to reproduce.

OPNsense hosted on ESXi 7.0.3

Software version used and hardware type if relevant, e.g.:

OPNsense 25.1.9_2-amd64
FreeBSD 14.2-RELEASE-p3
OpenSSL 3.0.16
Multiple hardware

g-a-c:
> In step 3, you shouldn't even get a prompt for Type, as there is no full cert for the PCKS, The Private Key is useless at this point cause again no full cert is created, it just creates the Priv Key and await the signed Cert from the external CA.

This is _probably_ fixable, but the UI elements are pretty static and don't care what "type" of thing you're trying to download, it's kind of on the user to know what they're doing at least a little and pick the right type.

g-a-c:
> N/A, it's a silent fail. Open Dev Tools in Browser look at console.. NOTHING

[...]

PR #8836, opened 2025-06-20
#8834/#8835: allow CSR to be downloaded from Sytem/Trust/Certificates

g-a-c:
Although you can create a CSR in the OPNsense web UI, you cannot download one. User @Zewwy from IRC was trying and getting silent failures because the UI doesn't have an option for it. You can choose the following options

- Certificate (silent failure, the CSR has no certificate body)
- PKCS12 (silent failure, the CSR has no certificate body)
- Private key (works fine, the private key is generated alongside the CSR)

This PR adds the "Certificate Signing Request" option into the download box, and when this option is used the API serves up the CSR payload

Fixes: #8834, #8835


AdSchellevis:
@g-a-c thanks, I’ll double check on my en when I’m near a pc.

AdSchellevis:
@g-a-c fixed a minor issue in https://github.com/opnsense/core/commit/8dd68b36d0c0128bb46195823bdc48410720e4a1, should be ok now. thanks!

PR #8848, opened 2025-06-24
dns/unbound: Add ipset support

Monviech:
Requires plumbing from: https://github.com/opnsense/core/pull/8791

Cherry picked into its own PR so the dnsmasq one can move along until Unbound is decided.


Monviech:
Since it only supports two aliases, its not as good as the dnsmasq implementation and for now we do not want it.

PR #8855, opened 2025-06-24
firewall/category: Add HTML based standardized color input to form

Monviech:
Just as an idea since the javascript dependency looks unmaintained.

This lets the browser render the color input.

It gets appended to body which is nice.

https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Elements/input/color
Chrome on Macbook:
image

Link: https://github.com/opnsense/core/pull/8858

### DIFF
--- a/src/opnsense/mvc/app/controllers/OPNsense/Firewall/Api/CategoryController.php
+++ b/src/opnsense/mvc/app/controllers/OPNsense/Firewall/Api/CategoryController.php
@@ -60,24 +60,6 @@ class CategoryController extends ApiMutableModelControllerBase
         return $result;
     }
 
-    /* The model validates the color for without # */
-    private function getCategoryOverlay(): array
-    {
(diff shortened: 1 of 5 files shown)


AdSchellevis:
@Monviech interesting option, but if we can't ditch the hash in the frontend (which would be my preferende), we should probably just add a simple `ltrim($str, '#')` in the controller

Monviech:
@AdSchellevis Thanks for the feedback. Ill see if I can get rid of # somehow in the frontend, I chose the controller first to test it.

swhite2:
At a first glance I don't see any issues using this (it's a lot cleaner in the UI, depending on platform 😉). The only functionality we would be dropping is locally cached colors for reuse elsewhere, but I don't see the use case for this to be honest.

Monviech:
Thanks, since there is no opposition I'll merge this to master for now.

PR #8859, opened 2025-06-25
mvc: extend base_bootgrid_table with grid_commands

swhite2:
Since the rendering of the table is now asynchronous, and because of general backwards incompatibility, custom buttons that were injected after UIBootgrid initialization don't show up anymore in with the new wrapper.

This commit attempts to generalize the problem by moving custom grid commands to the `base_bootgrid_table` partial, while keeping the responsibility of interaction local to the page at hand.

Because constructs such as `SimpleFileUploadDlg` cannot be delegated, move this into an event handler that fires when we know the UI has rendered.

In the long term it would make more sense to move this logic to the tabulator wrapper instead, but this requires the UI to be fully decoupled from the template.

### DIFF
--- a/src/opnsense/mvc/app/views/OPNsense/Kea/dhcpv4.volt
+++ b/src/opnsense/mvc/app/views/OPNsense/Kea/dhcpv4.volt
@@ -58,7 +58,13 @@
                 add:'/api/kea/dhcpv4/add_reservation/',
                 del:'/api/kea/dhcpv4/del_reservation/'
             }
-        );
+        ).on('load.rs.jquery.bootgrid', function() {
+            $("#upload_reservations").SimpleFileUploadDlg({
+                onAction: function(){
+                    grid_reservations.bootgrid('reload');
+                }
+            });
+        })
 
         $("#{{formGridPeer['table_id']}}").UIBootgrid(
             {   search:'/api/kea/dhcpv4/search_peer',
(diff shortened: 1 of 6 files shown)


src/opnsense/mvc/app/views/OPNsense/Kea/dhcpv4.volt
```diff
@@ -58,7 +58,13 @@
                 add:'/api/kea/dhcpv4/add_reservation/',
                 del:'/api/kea/dhcpv4/del_reservation/'
             }
-        );
+        ).on('load.rs.jquery.bootgrid', function() {
```
fichtner, line 61: it would be beneficial to pop the function into UIBootgrid() as this otherwise duplicates the same code on all required pages?

src/opnsense/mvc/app/views/OPNsense/Kea/dhcpv4.volt
```diff
@@ -172,7 +142,31 @@
...
+                            'toggle': 'tooltip'
+                        }
+                    },
+                    'download_reservations': {
+                        'title': lang._('Export as csv'),
+                        'class': 'btn btn-xs',
+                        'icon_class': 'fa fa-fw fa-table',
+                        'data': {
+                            'tooltip': 'toggle'
+                        }
```
fichtner: I wouldn't say this is ideal but much better than the custom hooking done before

swhite2: It certainly isn't ideal, but as mentioned in the description, we can't move this to the js logic yet without breaking backwards compatibility with the jquery bootgrid. Even if it were backwards compatible, we would still need the structure defined in HTML, which is now restricted to the partial.

If the UI were to be rendered based on the result of an API call this whole thing can move to UIBootgrid.

fichtner: I like the change, but I also wanted to affirm your further direction thoughts :)

src/opnsense/mvc/app/views/OPNsense/Kea/dhcpv4.volt
swhite2: That would require changing both `opnsense_bootgrid_plugin` and `opnsense_bootgrid`. We also can't assume what's executed here

fichtner: is there going to be wide spread use of load.rs.jquery.bootgrid doing this change for all other pages? aliases are using it and this commit indicates that there will be more spots, but maybe I misunderstood

swhite2: I would say more an exception to the rule for cases where the UI footer element must be present to bind something to it.

fichtner: ok thanks, marking as resolved

fichtner (approved):
implications of e57c325018 were discussed, worth a try

swhite2:
> implications of [e57c325](https://github.com/opnsense/core/commit/e57c32501862d780419cbf06f9f00d5ccf4b8e8a) were discussed, worth a try

For a bit more context: calling `$('[data-toggle="tooltip"]').tooltip();` will activate tooltips rendered at this point in time. The replacement construct is a delegation and will therefore steer UI elements rendered dynamically as well. There may be performance implications, but logically this only affects tooltips that have not yet been initialized. In these cases every mouseenter, hover etc. has to be checked for a target match. However, in most cases tooltips have been initialized on a specific selector in our code base - so the default tooltip acts as a fallback.

PR #8872, opened 2025-06-26
snapshots: Implement base_bootgrid_table

Monviech:
Delete button is hidden with a new parameter in the base_bootgrid_table


swhite2:
> Single selection is allowed now since the base bootgrid always renders the delete button

Not sure if this should be the case (hasn't always been the case, right?). Anyway, perhaps this can also be parameterized in the partial, we may not want to present the deleteSelected in all cases.

Edit: in fact, if we start looking at pages such as `sessions.volt`, the add button should also not be there.

Monviech:
Good idea, I'll work on that.

Monviech:
@swhite2 Okay I implemented your idea.

PR #8874, opened 2025-06-26
mvc: field types getValues() abstraction

fichtner:
The idea here is to remove `explode($char, (string)$field)` type constructs especially for cases where we iterate over.  As a liberty taken here we avoid emitting empty values as they do not make much sense to me.

CSVListField loses its internal separator since it does not offer to change it and the field should go away eventually.  It also adds a rudimentary unit test for the feature introduced here.

Now, things get a bit ugly.  BaseListField is our shared class for multi-value fields, but NetworkField, HostnameField, IPPortField and MacAddressField all implement their own separator handling.  I've also gone ahead and changed NetworkField since it is used by the example refactor in DHCRelay, but we should refactor this to avoid diverging implementation and being able to add an easy field type unit test that checks this only once (same as only once implemented ideally).

I'd also like to get rid of `getCurrrentValue()` where possible through `isEqual()` or `getValues()` and rename the remaining spots to `getValue()` for brevity.

First things first: discuss and merge the change proposed here and then work on another base class and the remaining refactors.

### DIFF
--- a/src/opnsense/mvc/app/models/OPNsense/Base/FieldTypes/BaseField.php
+++ b/src/opnsense/mvc/app/models/OPNsense/Base/FieldTypes/BaseField.php
@@ -393,7 +393,7 @@ abstract class BaseField
      * return field current value(s) as array (empty strings are omitted)
      * @return array field current values
      */
-    public function getValues()
+    public function getValues(): array
     {
         $value = $this->getCurrentValue();
         return strlen($value) ? [$value] : [];
--- a/src/opnsense/mvc/app/models/OPNsense/Base/FieldTypes/BaseListField.php
+++ b/src/opnsense/mvc/app/models/OPNsense/Base/FieldTypes/BaseListField.php
@@ -191,7 +191,7 @@ abstract class BaseListField extends BaseField
     /**
      * {@inheritdoc}
      */
-    public function getValues()
+    public function getValues(): array
     {
         return array_values(array_filter(explode(',', $this->internalValue), function ($k) {
             return !!strlen($k);
--- a/src/opnsense/mvc/app/models/OPNsense/Base/FieldTypes/CSVListField.php
+++ b/src/opnsense/mvc/app/models/OPNsense/Base/FieldTypes/CSVListField.php
@@ -176,7 +176,7 @@ class CSVListField extends BaseField
     /**
      * {@inheritdoc}
      */
-    public function getValues()
+    public function getValues(): array
     {
         return array_values(array_filter(explode(',', $this->internalValue), function ($k) {
             return !!strlen($k);
--- a/src/opnsense/mvc/app/models/OPNsense/Base/FieldTypes/NetworkField.php
+++ b/src/opnsense/mvc/app/models/OPNsense/Base/FieldTypes/NetworkField.php
@@ -283,7 +283,7 @@ class NetworkField extends BaseField
     /**
      * {@inheritdoc}
      */
-    public function getValues()
+    public function getValues(): array
     {
         return array_values(array_filter(explode($this->internalFieldSeparator, $this->internalValue), function ($k) {
             return !!strlen($k);


fichtner:
A bit of illustration:

```
# git grep 'explode.*getCurrentValue'
src/opnsense/mvc/app/controllers/OPNsense/Auth/Api/PrivController.php:            foreach (explode(',', $user->priv->getCurrentValue()) as $priv) {
src/opnsense/mvc/app/controllers/OPNsense/Auth/Api/PrivController.php:            foreach (explode(',', $group->priv->getCurrentValue()) as $priv) {
src/opnsense/mvc/app/controllers/OPNsense/Auth/Api/PrivController.php:                    in_array($id, explode(',', $user->priv->getCurrentValue())) &&
src/opnsense/mvc/app/controllers/OPNsense/Auth/Api/PrivController.php:                    in_array($id, explode(',', $group->priv->getCurrentValue())) &&
src/opnsense/mvc/app/controllers/OPNsense/Auth/Api/PrivController.php:                    $uuids = explode(',', $mdl->users->getCurrentValue());
src/opnsense/mvc/app/controllers/OPNsense/Auth/Api/PrivController.php:                    $uuids = explode(',', $mdl->groups->getCurrentValue());
src/opnsense/mvc/app/controllers/OPNsense/Auth/Api/PrivController.php:                    $privlist = array_filter(explode(',', $item->priv->getCurrentValue()));
src/opnsense/mvc/app/controllers/OPNsense/Auth/Api/UserController.php:            $members = explode(',', $group->member->getCurrentValue());
src/opnsense/mvc/app/models/OPNsense/Auth/FieldTypes/ApiKeyField.php:        foreach (array_filter(explode("\n", $this->getCurrentValue())) as $line) {
src/opnsense/mvc/app/models/OPNsense/Auth/FieldTypes/ApiKeyField.php:        foreach (array_filter(explode("\n", $this->getCurrentValue())) as $line) {
src/opnsense/mvc/app/models/OPNsense/Auth/FieldTypes/ApiKeyField.php:        foreach (array_filter(explode("\n", $this->getCurrentValue())) as $line) {
src/opnsense/mvc/app/models/OPNsense/Auth/FieldTypes/ApiKeyField.php:            array_filter(explode("\n", $this->getCurrentValue())),
src/opnsense/mvc/app/models/OPNsense/Interfaces/Bridge.php:            $members = explode(',', $bridge->members->getCurrentValue());
src/opnsense/mvc/app/models/OPNsense/Interfaces/Bridge.php:                foreach (explode(',', $bridge->$section->getCurrentValue()) as $if) {
src/opnsense/mvc/app/models/OPNsense/Kea/KeaDhcpv6.php:        $this_interfaces = explode(',', $this->general->interfaces->getCurrentValue());
```

```
# git grep 'explode.*(string).*->' | wc -l
      78
```

swhite2:
> and the field should go away eventually

I imagine there was an original intent here. If that doesn't hold up anymore, why not ditch it everywhere and move getValues() to the base class?

fichtner:
I'm looking at CSVListField as a rapid prototype field of the early days. We've moved on to fine-grained field types for various reasons. It does not have to go away, but it certainly could considering replacement with more on-point field types and hopefully also reducing code and regex use.

getValues() depends on the separator implicitly defined in BaseListField, previously internally defined in CSVListField and interchangeably defined in NetworkField et al. and the `internalMultiSelect` flag only found in BaseListField so that the BaseField doesn't know what it should do. NetworkField et al on top of that use `internalAsList` instead.

fichtner:
Talking further with @swhite2 we should rename BaseListField to BaseOptionField,BaseKeyValueField,BaseDictField,BaseSelectField or anything nice to make room for a new BaseListField for CSVListField, NetworkField and others to derive their functionality from. Luckily BaseListField isn't used directly in models.

Best case BaseXXXField could derive from BaseListField then, but that's just wishful thinking ;)

PR #8877, opened 2025-06-26
shaper: combine dnctl/ipfw setup and fix setup order (continuation of…

swhite2:
… eae56a725dc97c1d8fb0d368072a579e695fb186)

this has the added benefit that a call through configd won't be necesary anymore.

while here, dnctl_setup seems to be executed before the dummynet module load on bootup. fix this by deferring the sysctl values.

### DIFF
--- a/src/opnsense/scripts/shaper/start.sh
+++ b/src/opnsense/scripts/shaper/start.sh
@@ -24,9 +24,17 @@
 # ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
 # POSSIBILITY OF SUCH DAMAGE.
 
-# sysctl settings, use sysctl.kld.d since this setup script is called
-# before module load through rc, meaning the sysctls don't exist yet.
-cat <<EOF > /etc/sysctl.kld.d/dummynet.conf
-net.inet.ip.dummynet.io_fast=1
-net.inet.ip.dummynet.hash_size=256
-EOF
+SERVICE=${1}
+DNCTL="/etc/rc.d/dnctl start"
+IPFW="/etc/rc.d/ipfw enabled && /etc/rc.d/ipfw start || ( /etc/rc.d/ipfw onestop || true ); /usr/local/etc/rc.ipfw.post || true"
+
+if [ $# -eq 0 ]; then
+    ${DNCTL}
+    ${IPFW}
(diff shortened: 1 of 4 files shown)


src/opnsense/scripts/shaper/start.sh
```diff
@@ -0,0 +1,40 @@
...
+# OR CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF
+# SUBSTITUTE GOODS OR SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS
+# INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN
+# CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE)
+# ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
+# POSSIBILITY OF SUCH DAMAGE.
+
+SERVICE=${1}
+DNCTL="/etc/rc.d/dnctl start"
+IPFW="/etc/rc.d/ipfw enabled && /etc/rc.d/ipfw start || ( /etc/rc.d/ipfw onestop || true ); /usr/local/etc/rc.ipfw.post || true"
```
fichtner: general note: Nice spotting this so far!

We don’t need the `true` and id like to suggest unwinding this as a real script below checking if [ -z “${SERVICE}” -o “ ${SERVICE}” = “dnctl” ] and the other one below

swhite2: @fichtner Thanks :) How about https://github.com/opnsense/core/pull/8877/commits/60f1becbc045aa10f8a0f57648dff22d15a227f3?

fichtner: Yes, almost, don’t need the variables anymore and transform || to if else for readability

fichtner: Btw we should either exit 0 at the end or set errors:no just to be safe

swhite2: @fichtner Attempt no. 2: https://github.com/opnsense/core/pull/8877/commits/eefa0ffde43b9e911a291665ec082025e086b25d

PR #8883, opened 2025-06-27
monit: Remove dirty subsystem checks as base_apply_button does not track that

Monviech:
Follow up for: https://github.com/opnsense/core/commit/f4bef1f530d76d23a5a485d52c0548a9e75ba979


src/opnsense/mvc/app/models/OPNsense/Monit/Monit.php
```diff
@@ -315,39 +315,6 @@ public function performValidation($validateFullModel = false)
         return $messages;
     }
 
-    /* return the dirty marker file location */
-    private function dirtyFile()
-    {
-        return (new AppConfig())->application->tempDir . '/monit.dirty';
```
fichtner, line 321: since I refactored this recently it means you can remove use AppConig line at the beginning of the file

PR #8885, opened 2025-06-27
ui: Add checkmark to SimpleActionButton as additional indicator

Monviech:
Persist checkmark long enough to ensure a new keyframe in remote desktop streaming tools can update the change.

Fixes: https://github.com/opnsense/core/issues/8647

### DIFF
--- a/src/opnsense/www/js/opnsense_ui.js
+++ b/src/opnsense/www/js/opnsense_ui.js
@@ -602,11 +602,9 @@ $.fn.SimpleActionButton = function (params) {
             let pre_action = function () {
                 return (new $.Deferred()).resolve();
             }
-
             if (params && params.onPreAction) {
                 pre_action = params.onPreAction;
             }
-
             pre_action().done(function () {
                 ajaxCall(this_button.data('endpoint'), {}, function (data, status) {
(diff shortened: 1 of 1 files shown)


src/opnsense/www/js/opnsense_ui.js
```diff
@@ -577,11 +577,28 @@ stdDialogRemoveItem.defaults = {
...
         this_button.html(label_content);
+
+        let hideCheckTimeout;
+
         this_button.on('click', function () {
-            this_button.find('.reload_progress').addClass("fa fa-spinner fa-pulse");
+            const icon = this_button.find('.reload_progress');
+
+            // prevent icon issues with multiple rapid clicks on the button
+            clearTimeout(hideCheckTimeout);
```
swhite2, line 598: this throws no error on undefined?

swhite2 (approved):
Ok, small comment

src/opnsense/www/js/opnsense_ui.js
Monviech: https://developer.mozilla.org/en-US/docs/Web/API/Window/clearTimeout

`Passing an invalid ID to clearTimeout() silently does nothing; no exception is thrown.`

I think thats fine right?

PR #8887, opened 2025-06-27
ipfw: combine service logic

swhite2:
Reduce the complexity a bit


src/opnsense/scripts/shaper/flush_all.sh
```diff
@@ -1,5 +1,5 @@
 #!/bin/sh
-#    Copyright (c) 2016 Deciso B.V.
+#    Copyright (c) 2025 Deciso B.V.
```
fichtner: Time travel? 😊

fichtner (approved):

PR #8901, opened 2025-07-02
automation/filter: Do not reload the grid during reconfigureAct

Monviech:
Since we call functions to repopulate the selectpickers during the reconfigureAct they triggered the changed events to reload the bootgrid and change the url hash. A new variable was introduced that tracks the reconfigureActInProgress and prevents spurious reloads during it. After the selectpickers have been repopulated with their preserved values, a promise is returned which sets the reconfigureActInProgress to false, unlocking the normal change event behavior during selecpicker usage.

Essentially this prevents the grid reloading and the url hash resetting, each apply would boot the user back to Floating rules with resetted selectpickers.

Refines this: https://github.com/opnsense/core/pull/8898

### DIFF
--- a/src/opnsense/mvc/app/views/OPNsense/Firewall/filter_rule.volt
+++ b/src/opnsense/mvc/app/views/OPNsense/Firewall/filter_rule.volt
@@ -509,46 +509,49 @@
 
         // Populate category selectpicker
         function populateCategoriesSelectpicker(callback) {
-            ajaxCall('/api/firewall/filter/list_categories', {}, function (data) {
-                if (!data.rows) return;
+            const currentSelection = $("#category_filter").val();
 
-                const $categoryFilter = $("#category_filter");
-                const currentSelection = $categoryFilter.val();
+            $("#category_filter").fetch_options(
+                '/api/firewall/filter/list_categories',
+                {},
+                function (data) {
+                    if (!data.rows) return [];
 
-                // Sort used categories first, then alphabetically
-                data.rows.sort((a, b) => {
-                    const aUsed = a.used > 0 ? 0 : 1;
-                    const bUsed = b.used > 0 ? 0 : 1;
+                    // Sort used categories first, then alphabetically
+                    data.rows.sort((a, b) => {
+                        const aUsed = a.used > 0 ? 0 : 1;
+                        const bUsed = b.used > 0 ? 0 : 1;
 
-                    if (aUsed !== bUsed) {
-                        return aUsed - bUsed;
-                    }
+                        if (aUsed !== bUsed) return aUsed - bUsed;
+                        return a.name.localeCompare(b.name);
+                    });
 
-                    return a.name.localeCompare(b.name);
-                });
-
-                $categoryFilter.empty().append(
-                    data.rows.map(row => {
+                    return data.rows.map(row => {
                         const optVal = $('<div/>').text(row.name).html();
                         const bgColor = row.color || '31708f';
 
-                        return $("<option/>", {
+                        return {
                             value: row.uuid,
-                            html: row.name,
+                            label: row.name,
                             id: row.used > 0 ? row.uuid : undefined,
-                            "data-content": row.used > 0
+                            'data-content': row.used > 0
                                 ? `<span><span class="badge badge-sm" style="background:#${bgColor};">${row.used}</span> ${optVal}</span>`
                                 : undefined
-                        });
-                    })
-                );
+                        };
+                    });
+                },
+                false,
+                function () {
+                    if (currentSelection?.length) {
+                        $("#category_filter").val(currentSelection).selectpicker('refresh');
+                    }
 
-                $categoryFilter.val(currentSelection).selectpicker('refresh');
-
-                if (typeof callback === "function") {
-                    callback();
-                }
-            });
+                    if (typeof callback === "function") {
+                        callback();
+                    }
+                },
+                true  // render_html
+            );
         }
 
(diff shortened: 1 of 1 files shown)


fichtner:
@swhite2 is better at this than me ;)

src/opnsense/mvc/app/views/OPNsense/Firewall/filter_rule.volt
```diff
@@ -543,14 +544,21 @@
                 );
 
                 $categoryFilter.val(currentSelection).selectpicker('refresh');
```
swhite2: Isn't this happening twice now with the callback as well?

I feel like the `withPreservedSelection` logic isn't really necessary if the refresh behavior is dictated by the populate functions. Refreshing with the current selection active, but falling back to a default if the new fetched list doesn't contain the entry anymore should perhaps be the default behavior? It's also ok to have this logic duplicated.

TL:DR; `withPreservedSelection` can probably be dropped, unless I'm misunderstanding something.

src/opnsense/mvc/app/views/OPNsense/Firewall/filter_rule.volt
```diff
@@ -507,7 +507,8 @@
             $('[data-toggle="tooltip"]').tooltip({ container: 'body' });
         });
 
-        function populateCategoriesSelectpicker() {
+        // Populate category selectpicker
+        function populateCategoriesSelectpicker(callback) {
```
swhite2: As a general note, why isn't the categories selectpicker using the `fetch_options`?

Monviech: Good point, I added it.

src/opnsense/mvc/app/views/OPNsense/Firewall/filter_rule.volt
Monviech: You were totally right, thank you.

src/opnsense/mvc/app/views/OPNsense/Firewall/filter_rule.volt
```diff
@@ -507,50 +507,61 @@
             $('[data-toggle="tooltip"]').tooltip({ container: 'body' });
         });
 
-        function populateCategoriesSelectpicker() {
-            ajaxCall('/api/firewall/filter/list_categories', {}, function (data) {
-                if (!data.rows) return;
+        // Populate category selectpicker
+        function populateCategoriesSelectpicker(callback) {
```
swhite2: Callback isn't necessary anymore when moving to `fetch_options`

src/opnsense/mvc/app/views/OPNsense/Firewall/filter_rule.volt
```diff
@@ -507,50 +507,61 @@
...
+                function () {
+                    if (currentSelection?.length) {
+                        $("#category_filter").val(currentSelection).selectpicker('refresh');
+                    }
 
-                $categoryFilter.val(currentSelection).selectpicker('refresh');
-            });
+                    if (typeof callback === "function") {
+                        callback();
+                    }
```
swhite2: Can be dropped

src/opnsense/mvc/app/views/OPNsense/Firewall/filter_rule.volt
```diff
@@ -593,10 +604,17 @@
...
+                            if (!reconfigureActInProgress) {
+                                grid.bootgrid('reload');
+                            }
                         }
                     }
                     interfaceInitialized = true;
+
+                    if (typeof callback === "function") {
+                        callback();
+                    }
```
swhite2: Can be dropped

src/opnsense/mvc/app/views/OPNsense/Firewall/filter_rule.volt
```diff
@@ -706,9 +726,15 @@
...
         $("#reconfigureAct").SimpleActionButton({
-            onAction: function (data, status) {
-                populateInterfaceSelectpicker();
-                populateCategoriesSelectpicker();
+            onAction(data, status) {
+                reconfigureActInProgress = true;
+
+                Promise.all([
+                    new Promise(populateInterfaceSelectpicker),
+                    new Promise(populateCategoriesSelectpicker)
```
swhite2: Both functions are going to run sync here. If constructed using a Promise, you need to account for `resolve` (and optionally `reject`) in both functions as params. Technically you could even return a promise from `fetch_options`, but probably better to do it as is, and call `resolve()` from the completion callback

swhite2 (changes requested):

src/opnsense/mvc/app/views/OPNsense/Firewall/filter_rule.volt
```diff
@@ -706,9 +719,16 @@
         $('#category_filter').parent().find('.dropdown-toggle').prepend('<i class="fa fa-tag" style="margin-right: 6px;"></i>');
 
         $("#reconfigureAct").SimpleActionButton({
-            onAction: function (data, status) {
-                populateInterfaceSelectpicker();
-                populateCategoriesSelectpicker();
+            onAction(data, status) {
+                reconfigureActInProgress = true;
```
swhite2, line 723: technically, this should already be true on `onPreAction`, as `onAction` fires after the request is done.

`onPreAction` expects a promise to resolve though, so do `return Promise.resolve()` there.

swhite2 (approved):
Looks good, minus small comment

PR #8908, opened 2025-07-03
dnsmasq: Allow DomainIPField to be cleared

Monviech:
Same issue as here:

https://github.com/opnsense/core/pull/8792


fichtner:
May be worth grepping for “update only” and see if other fields have the same issue.

Monviech:
https://github.com/opnsense/core/blob/05ddadf5a137a2231820d2399fa8c00e1e84c802/src/opnsense/mvc/app/models/OPNsense/Auth/FieldTypes/ApiKeyField.php#L58-L61

I also grepped for something like:

`git grep -F 'elseif (' | grep 'value'`

But I did not find much more on first glance.

Issue #8821, opened 2025-06-18
Dnsmasq: Alias should be CNAME an not A-Record


olmo1501:
**Important notices**

Before you add a new report, we ask you kindly to acknowledge the following:

- [X] I have read the contributing guide lines at https://github.com/opnsense/core/blob/master/CONTRIBUTING.md
- [X] I am convinced that my issue is new after having checked both open and closed issues at https://github.com/opnsense/core/issues?q=is%3Aissue

**Describe the bug**

If you add an Alias to a Dnsmasq Host entry it is configured as an A-Record. As such it resolves only to it's IPv4 address instead of all it's addresses.

**Expected behavior**

CNAME that points to the host.
**Additional context**

See also: https://forum.opnsense.org/index.php?topic=47513.msg239445#msg239445

**Environment**

OPNsense 25.1.8_1 (amd64).
Intel® N6005 Quad Core
Network Intel® i226

Monviech:
This one is quite tricky

The aliases are currently written into the file when they are either:
- static IPv4 addresses
- static IPv6 addresses

But partial IPv6 addresses are skipped as they must be constructed internally by DNSmasq.

As people want to resolve their hostnames and aliases /before/ a dynamic dns register has been made, this file exists. But partial IPv6 addresses cannot do that, they can only be resolved after registration.

This means either all aliases become CNAMES, or partial Ipv6 addresses are not supported as aliases.

Monviech:
I want a second opinion from @fichtner.

Monviech:
Adding an additional grid for CNAMEs could be a possibility, as we could use the unique constraint of the model (cnames must be unique). This could also be used for other record types later if they become a requirement.

Another alternative would be an additional cname field in the current host grid, though this might be messy.

But before something new is added here, it is best to wait for @AdSchellevis how we should continue with new features in dnsmasq right now.

Monviech:
For now we put this on hold and wait for more requests for this feature.

olmo1501:
I'm completely fine with that. Thanks for considering this request.

The longer I think about it the more I think that Unbound would be the better place to add CNAMEs. For now I did this via the Unbound Advanced Configuration which works perfectly fine.

stelb:
just switched to dnsmasq and tried to add a cname alias.

I have a host with dynamic IP and services running on it, so sth like this:

someservice.int.mydomain.de -> app02.int.mydomain.de
anotherservice.int.mydomain.de -> app02.int.mydomain.de

[...]

PR #8915, opened 2025-07-06
dnsmasq: Add CNAME configuration option to host overrides

Monviech:
It is more useful than expected:

- Services that are IPv6 only, use DHCPv6 via dynamic partial IPv6 addresses can only have alternative names with CNAMEs
- Services that are dynamic IPv4, have a dhcp-host without IPv4 address set, can only have an alias with CNAMEs

Hosting virtual hosts on a webserver with dynamic IPv6 addresses need CNAMEs.

Fixes: https://github.com/opnsense/core/issues/8821

known error states (crash):

```
1.
cname=example4.com
 
dnsmasq: bad CNAME at line 12 of /usr/local/etc/dnsmasq.conf
 
2.
cname=example2.com,example.com
cname=example2.com,example.com
 
dnsmasq: duplicate CNAME at line 11 of /usr/local/etc/dnsmasq.conf
 
3.
cname=example3.com,example.com
cname=example.com,example3.com
 
dnsmasq: CNAME loop involving example.com
```

Allowed (no crash):

```
1.
host-record=example.com,192.168.1.1
cname=example.com,example3.com

2.
host-record=example.com,192.168.1.1
cname=example3.com,example.com

### DIFF
--- a/src/opnsense/mvc/app/models/OPNsense/Dnsmasq/Dnsmasq.php
+++ b/src/opnsense/mvc/app/models/OPNsense/Dnsmasq/Dnsmasq.php
@@ -137,7 +137,7 @@ class Dnsmasq extends BaseModel
                 }
             }
 
-            foreach (array_filter(explode(',', (string)$host->cnames)) as $cname) {
+            foreach ($host->cnames->getValues() as $cname) {
                 if ($usedHostCnames[$cname] > 1) {
                     $messages->appendMessage(
                         new Message(
--- a/src/opnsense/mvc/app/models/OPNsense/Dnsmasq/Dnsmasq.php
+++ b/src/opnsense/mvc/app/models/OPNsense/Dnsmasq/Dnsmasq.php
@@ -74,10 +74,6 @@ class Dnsmasq extends BaseModel
                 $usedHostFqdns[$fqdn] = true;
             }
 
-            foreach (array_filter(explode(',', (string)$host->aliases)) as $alias) {
-                $usedHostFqdns[$alias] = true;
-            }
-
             foreach (array_filter(explode(',', (string)$host->cnames)) as $cname) {
                 $usedHostCnames[$cname] = ($usedHostCnames[$cname] ?? 0) + 1;
             }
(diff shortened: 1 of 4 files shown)

### CODE REVIEW

**Comment by AdSchellevis** [comment]
how about adding a checkbox to register aliases as cnames instead of A/AAAA records? it would save quite some code and both are not likely used combined.

**Comment by Monviech** [comment]
I'm not sure:

- I would still need to validate that cnames do not overlap with aliases and host/domain
- I would still need to validate that cnames are unique
- I would have to implement exclusion logic here for such a checkbox enabled: https://github.com/opnsense/core/blob/38ca7f9fbed869aff3a55396f8a21b9c8904c83a/src/etc/inc/plugins.inc.d/dnsmasq.inc#L182
- From a UX standpoint its now cleanly labeled what each field does.

I think overall the additional field has slightly more benefits as opposed to a dropdown or checkbox.

File: src/opnsense/mvc/app/models/OPNsense/Dnsmasq/Dnsmasq.php [hunk 1]
```diff
@@ -122,6 +136,27 @@ public function performValidation($validateFullModel = false)
                     $messages->appendMessage(new Message($messageText, $key . ".ip"));
                 }
             }
+
+            foreach (array_filter(explode(',', (string)$host->cnames)) as $cname) {
```
AdSchellevis: might consider `getValues()` which @fichtner recently implemented.

AdSchellevis (approved):
one small remark due to a recent change (https://github.com/opnsense/core/commit/bb5e03a3e37d95c938cb82d85140ed323c54cbc8)

src/opnsense/mvc/app/models/OPNsense/Dnsmasq/Dnsmasq.php
fichtner:
```suggestion
            foreach ($host->cnames->getValues() as $cname) {
```

yep, no cast and no filter needed

Issue #8918, opened 2025-07-07
SystemResources vs system_resources API endpoints require different permissions


Snuffy2:
**Important notices**

Before you add a new report, we ask you kindly to acknowledge the following:

- [x] I have read the contributing guide lines at https://github.com/opnsense/core/blob/master/CONTRIBUTING.md
- [x] I am convinced that my issue is new after having checked both open and closed issues at https://github.com/opnsense/core/issues?q=is%3Aissue

**Describe the bug**

I maintain https://github.com/travisghansen/hass-opnsense. Based on https://github.com/opnsense/docs/commit/83d9aa86c468721fab054d209cbc3ed211284a6d, I changed all of the API endpoints to use the snake_case endpoint.

`/api/diagnostics/system/systemResources` works with just the `Lobby: Dashboard` permission. However, `/api/diagnostics/system/system_resources` does not work with `Lobby: Dashboard` and gives a 403.

With the snake case endpoints, only the `All pages` (aka. Admin) permission works. No other permissions I tested allowed them to work. 

See comment below, but this applies to most (all?) of the /api/diagnostics/system/ endpoints.

**To Reproduce**

Steps to reproduce the behavior: See above

**Expected behavior**

Permissions should be the same between the old camel case and new snake case api endpoints and shouldn't require only admin permissions to use. It would be even better if the API documentation also listed the required permission.

**Screenshots**

Test user with just the `Lobby: Dashboard` permission in Postman:

<img width="859" alt="Image" src="https://github.com/user-attachments/assets/a8e938a7-3808-4cab-b46b-45a98799938d" />

<img width="863" alt="Image" src="https://github.com/user-attachments/assets/af80ce61-529d-4c6f-ad2f-4554dace5c58" />

**Relevant log files**
None

**Additional context**

I have not fully tried the other endpoints that changed from camel case to snake case to see if their permissions changed as well.

**Environment**

OPNsense 25.1.10

Snuffy2:
The same issue applies to:

| Old camel case | New snake case | Finding |
| --- | --- | --- |
| /api/diagnostics/system/systemResources | /api/diagnostics/system/system_resources | Camel case worked with `Lobby: Dashboard` permission. **Snake case does not.** | 
| /api/diagnostics/system/systemTemperature | /api/diagnostics/system/system_temperature | Camel case worked with `Lobby: Dashboard` permission. **Snake case does not.** | 
[...]

PR #8920, opened 2025-07-08
mvc: Turn camelCase into snake_case in javascript, ACL and volt files

Monviech:
Found some more camelCase patterns in:

- Dashboard javascript and ACL
- missed volt file
- system ACLs

Fixes: https://github.com/opnsense/core/issues/8918


Monviech:
Still finding a bit more, I gonna work on this for a while longer.

PR #8923, opened 2025-07-08
Make "Reset zoom" translatable

tobiasdegen:
This pull request updates the "Reset zoom" button label in the Health module to support internationalization.

Change summary:
Replaced the hardcoded string "Reset zoom" with {{ lang._('Reset zoom') }} in the Volt template


AdSchellevis:
@tobiasdegen thanks, looks like we missed a spot there.

PR #8931, opened 2025-07-10
mvc: use getNodeContent to gather grid data

swhite2:
This allows us to include keyable data that is unaffected by translations, making sure that we can format grid cells properly based on the actual value.

Builds on https://github.com/opnsense/core/commit/23c09fc4d

fwiw, `$` as a prefix may not be optimal so leaving this open for discussion

### DIFF
--- a/src/opnsense/mvc/app/library/OPNsense/Base/UIModelGrid.php
+++ b/src/opnsense/mvc/app/library/OPNsense/Base/UIModelGrid.php
@@ -126,23 +126,14 @@ class UIModelGrid
                 $row = [];
                 $row['uuid'] = $record->getAttributes()['uuid'];
                 $content = $record->getNodeContent();
-                foreach ($content as $field => $val) {
-                    $key = substr($field, 1);
-                    if (str_starts_with($field, '$') && in_array($key, $fields)) {
-                        $row[$key] = $val;
-                        if ($content[$key] !== $val) {
-                            /**
-                             * getNodeContent() returns "$<key>" for description values
-                             * this would break API compat if we're returning this directly,
-                             * so swap the two around:
-                             *
-                             * [key] => "Descriptive (translated) value"
-                             * [$key] => "value used in config"
-                             *
-                             * only do this if the two values don't match, unconditionally including it
-                             * would double API response size.
-                             */
-                            $row[$field] = $content[$key];
+                foreach ($fields as $fieldname) {
+                    if (array_key_exists($fieldname, $content)) {
+                        $row[$fieldname] = $content[$fieldname];
+
+                        $descr = '$' . $fieldname;
+                        if (array_key_exists($descr, $content) && $content[$descr] !== $content[$fieldname]) {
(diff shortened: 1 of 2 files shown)


src/opnsense/mvc/app/library/OPNsense/Base/UIModelGrid.php
```diff
@@ -125,9 +125,24 @@ public function fetch(
                 // before searching.
                 $row = [];
                 $row['uuid'] = $record->getAttributes()['uuid'];
-                foreach ($fields as $fieldname) {
-                    if ($record->$fieldname != null) {
-                        $row[$fieldname] = $record->$fieldname->getDescription();
+                $content = $record->getNodeContent();
+                foreach ($content as $field => $val) {
+                    if (str_starts_with($field, '$') && in_array($key = substr($field, 1), $fields)) {
```
fichtner: hiding the substitution in the function call calls for refactoring trouble later

src/opnsense/mvc/app/library/OPNsense/Base/UIModelGrid.php
```diff
@@ -125,9 +125,24 @@ public function fetch(
...
+                        $row[$key] = $val;
+                        if ($content[$key] !== $val) {
+                            /**
+                             * getNodeContent() returns "$<key>" for description values
+                             * this would break API compat if we're returning this directly,
+                             * so swap the two around:
+                             *
+                             * [key] => "Descriptive (translated) value"
+                             * [$key] => "value used in config"
```
fichtner: I'm not opposed to the idea as already discussed. The question is why we cannot / should not switch those values while implementing this. The previous value (descriptive) doesn't have immediate operational value? A breaking change in e.g. 26.1 would seem reasonable?

swhite2: Sure, it will need some handling on the grid implementation though to always default to presenting a `$<key>` value if available. Let me cook something up

src/opnsense/mvc/app/library/OPNsense/Base/UIModelGrid.php
swhite2: https://github.com/opnsense/core/pull/8931/commits/697bea5db8677a0ce7f2e5a2b453dbf0a8049ef6

src/opnsense/mvc/app/library/OPNsense/Base/UIModelGrid.php
swhite2: @fichtner https://github.com/opnsense/core/pull/8931/commits/6ec225664868dbad7723f09e81bf63f972f064bc should do the trick

swhite2:
> Personally I'm ok with this, although we need to talk about this in the next meeting, especially the '$' character (or a replacement) use which would be best synced between model cache and API return before 25.7 is out.

Yep, there's already implicit escaping happening on the js side:

```
key: value
"$key": value
```

PR #8932, opened 2025-07-10
lang: Make more strings translateable

tobiasdegen:
Updated strings to be translateable...

The string "Starting from " is still not translateable. I don't know how to do it, because it is dynamically loading after volt templating.


fichtner:
Merged, thanks!

Translations in JavaScript are trickier, but should also work, see:

https://github.com/opnsense/core/blob/f85bd4f751ca3344362c893f1e8767f45313af34/src/opnsense/mvc/app/views/OPNsense/CaptivePortal/vouchers.volt#L123

Issue #8913, opened 2025-07-05
mvc: increase default row count


TheColetrain:
by default, lists only show 7 rows.  Add an option to increase the default.

Yes, I know you can adjust each page, but I don't want to do it EVERY TIME!!! 
yes, I realize the page will load slower, but it is faster than me having to select a higher number every page load!

I would be happy with any number larger, 7 is just terrible.  10, or 20 ?  Or let us set a default.

AT-StephenDetomasi:
I think there is an issue open for this already. Agreed though, 7 is simply not enough and really is a silly default setting. 25 would be much more suitable, or at least allow us to edit the default somehow.

sopex:
I agree, preferably the user choice should just be saved. But 7 is very little and I believe arbitrary for today's computing power.

fichtner:
Maybe @Monviech can look at it for his base grid template and change the default there? 20-25 makes sense screen wise.

Monviech:
@fichtner I looked at this briefly and the default comes from here:

https://github.com/opnsense/core/blob/f85bd4f751ca3344362c893f1e8767f45313af34/src/opnsense/www/js/opnsense_bootgrid_plugin.js#L130

This means I guess I would need to make this plugin configurable, e.g. by feeding it a configuration api endpoint like the dashboard uses:

https://github.com/opnsense/core/blob/f85bd4f751ca3344362c893f1e8767f45313af34/src/opnsense/www/js/opnsense_widget_manager.js#L124

cc @swhite2

Monviech:
I could also just do this probably:

`rowCount:[20,7,14,50,100,-1], `

swhite2:
Given that the updated grid system now scrolls in-place and adjusts for the page height the quickest win here would be to change the default for any page, I'd say 20 or 50.

Not every page has the same pagination values, so one could argue it could even be "the largest value before 'all'". For example, the log pages would be set to 1000 by default.

Monviech:
I take this one again, it was discussed that the simplest fix is removing 7 from the default, and add it back on pages where two grids are displayed by default.

These are not many instances of double grids (maybe 5-6 in core) so this will spawn the least new code while coming close to the expected result.

bbcoursera:
I want to chime in as a new Opnsense user.  

[...]

PR #8933, opened 2025-07-11
bootgrid: Double default row count

Monviech:
This would be the lowest hanging fruit I think.

We have to be at least a bit reservative since there are pages with two grids below each other.

Fixes: https://github.com/opnsense/core/issues/8913

### DIFF
--- a/src/opnsense/www/js/opnsense_bootgrid.js
+++ b/src/opnsense/www/js/opnsense_bootgrid.js
@@ -124,7 +124,7 @@ class UIBootgrid {
         this.options = {
             sorting: true,
             selection: true,
-            rowCount: [7, 14, 20, 50, 100, true],
+            rowCount: [14, 7, 20, 50, 100, true],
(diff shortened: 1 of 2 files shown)


swhite2:
This changes the rendering order as well, not that great in terms of UI

swhite2:
We can change the index to 1 here:

https://github.com/opnsense/core/blob/ffdea7f2bb292d03dbde7b013e6193b34b8f9fbe/src/opnsense/www/js/opnsense_bootgrid.js#L963

However, these values can be defined by the template, in which case index 1 might not exist - then again, 0 might also not exist if an empty array is provided.

probably best to add a `defaultRowCountIndex` option, default it to `1` so it can be overridden from templates and do some checks and balances such that the index cannot be out of bounds.

fichtner:
20-25 may be a better fit with the grid filling the screen but appt still visible underneath. The default could be labeled “Default (xx)” which would make the ordering less wonky.

Monviech:
I'll leave this one up to @swhite2

PR #8939, opened 2025-07-15
ui: flatten nested containers for grid inclusion

swhite2:
Continuation of https://github.com/opnsense/core/commit/a0bc7adda95876937ea0ba9c3d5542bfb741ffdb and https://github.com/opnsense/core/commit/66b0488a99a6dc6bf763cb440c65b5d37c847c27.

Because https://github.com/opnsense/core/blob/66b0488a99a6dc6bf763cb440c65b5d37c847c27/src/opnsense/mvc/app/controllers/OPNsense/Base/ControllerBase.php#L260 always picks the last key regardless of depth, and because our model does not allow nested ArrayFields, this should be safe to do.

### DIFF
--- a/src/opnsense/mvc/app/controllers/OPNsense/Base/ControllerBase.php
+++ b/src/opnsense/mvc/app/controllers/OPNsense/Base/ControllerBase.php
@@ -256,12 +256,11 @@ class ControllerBase extends ControllerRoot
                             $record['label'] = gettext((string)$item);
                             break;
                         case 'id':
+                            $parts = explode('.', (string)$item);
                             if (!empty($root)) {
-                                $parts = explode('.', (string)$item);
                                 $record['column-id'] = implode('.', array_slice($parts, array_search($root, $parts) + 1));
                             } else {
-                                $tmp = explode('.', (string)$item);
-                                $record['column-id'] = end($tmp);
+                                $record['column-id'] = end($parts);
                             }
                             break;
                     }
--- a/src/opnsense/mvc/app/controllers/OPNsense/Base/ControllerBase.php
+++ b/src/opnsense/mvc/app/controllers/OPNsense/Base/ControllerBase.php
@@ -235,7 +235,7 @@ class ControllerBase extends ControllerRoot
      * Extract grid fields from form definition
      * @return array
      */
-    public function getFormGrid($formname, $grid_id = null, $edit_alert_id = null)
+    public function getFormGrid($formname, $grid_id = null, $edit_alert_id = null, $root = null)
     {
         /* collect all fields, sort by sequence */
         $all_data = [];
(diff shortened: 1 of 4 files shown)


src/opnsense/mvc/app/library/OPNsense/Base/UIModelGrid.php
```diff
@@ -121,9 +121,27 @@ public function fetch(
                     continue;
                 }
 
+                $flatten = function($input) {
```
AdSchellevis: probably better to move this into the class as method, each loop (record) will reconstruct the function here. 

With a bit of magic, we should be able to save some space, how about something like:


```
function flatten($node, $path='')
{
    if (is_array($node)) {
        foreach ($node as $key => $value) {
            yield from flatten($value, ltrim($path . '.'. $key, '.'));
        }
    } else {
       yield $path => $node;
    }
}
```

In which case we can merge the payload in a one-liner `iterator_to_array(flatten($mydata))`

AdSchellevis (changes requested):

src/opnsense/mvc/app/library/OPNsense/Base/UIModelGrid.php
swhite2: @AdSchellevis Clever, neat :)

Unfortunately, the front-end can't really deal with "model.container.key" constructs without more structural changes (backwards incompatibility for non-`base_bootgrid` tables). The simplest solution is still to only move the key to the root, at the risk of collisions. Thoughts?

swhite2: something like https://github.com/opnsense/core/pull/8939/commits/75050d14e91b259bd18c334727f31e833424bf85

swhite2:
@AdSchellevis https://github.com/opnsense/core/pull/8939/commits/59d1d25dde3100f44c0d3b1cef6d106e48c62679 should do the trick per our discussion

src/opnsense/mvc/app/controllers/OPNsense/Base/ControllerBase.php
```diff
@@ -256,8 +256,13 @@ public function getFormGrid($formname, $grid_id = null, $edit_alert_id = null)
                             $record['label'] = gettext((string)$item);
                             break;
                         case 'id':
-                            $tmp = explode('.', (string)$item);
-                            $record['column-id'] = end($tmp);
+                            if (!empty($root)) {
+                                $parts = explode('.', (string)$item);
```
AdSchellevis: can move this one line up

src/opnsense/mvc/app/controllers/OPNsense/Base/ControllerBase.php
```diff
@@ -256,8 +256,13 @@ public function getFormGrid($formname, $grid_id = null, $edit_alert_id = null)
...
+                                $record['column-id'] = implode('.', array_slice($parts, array_search($root, $parts) + 1));
+                            } else {
+                                $tmp = explode('.', (string)$item);
+                                $record['column-id'] = end($tmp);
```
AdSchellevis: and use `$parts` here :)

AdSchellevis (approved):
one small remark, saves a line :)

PR #8940, opened 2025-07-15
interface/vip: Add generalized grid formatter that uses existing style declarations

Monviech:
… in dialogVip to hide and show the same fields based on selected type, e.g. CARP only shows carp related fields in the grid all else will be empty string.

### DIFF
--- a/src/opnsense/mvc/app/controllers/OPNsense/Interfaces/VipController.php
+++ b/src/opnsense/mvc/app/controllers/OPNsense/Interfaces/VipController.php
@@ -35,5 +35,41 @@ class VipController extends \OPNsense\Base\IndexController
         $this->view->pick('OPNsense/Interface/vip');
         $this->view->formDialogVip = $this->getForm("dialogVip");
         $this->view->formGridVip = $this->getFormGrid("dialogVip");
+        $this->view->modeStyleFlat = $this->getModeStyleFlatMap($this->view->formDialogVip);
     }
+
+    /**
+     * Return a flat map string of field names to allowed modes based on their styles.
+     * Type is used to decide between boolean and text formatter in volt template
+     * ("vhid=carp,ipalias|checkbox;advbase=carp|text")
+     * Consumed by bootgrid formatter to automatically hide column fields based on mode
+     */
+    protected function getModeStyleFlatMap(array $form): string
+    {
+        $entries = [];
+
+        foreach ($form as $field) {
+            if (!isset($field['style'], $field['id'], $field['type'])) {
+                continue;
+            }
+
+            $parts = explode('.', $field['id']);
+            $fieldName = end($parts);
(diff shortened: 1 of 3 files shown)


src/opnsense/mvc/app/controllers/OPNsense/Interfaces/VipController.php
```diff
@@ -35,5 +35,39 @@ public function indexAction()
...
+            }
+
+            if (!empty($modes)) {
+                $result[$fieldName] = ['modes' => $modes, 'type' => $type];
+            }
+        }
+
+        return $result;
+    }
+
```
fichtner:
```suggestion
```

fichtner:
I'm not against this, but it feels like over-engineering a minor model issue/quirk from the frontend especially.

Monviech:
@fichtner Do you have an idea how to do it differently maybe? I thought doing it like this was pretty smart as you can add more fields with styles in the future if needed, and you don't have to touch that code anymore.

fichtner:
I don't have strong opinions. My basic concern is we're hiding API-returned data in the frontend, because we know that data is not relevant. Question is if we should return the data the way we are from the API or not.

Monviech:
I would like to use some centralized metadata that shows which data is relevant for which mode, right now the dialog has that information in the form of styles.

I think that metadata is only related to the view as its in the view controller? That was why I hide things in the view right now.

Maybe there is a different approach without duplicating this metadata or declaring it at a different spot?

swhite2:
As far as I can see, nothing is really generalized here except the agreement that this information should come from the view, even though all relevant data for determining rendering mode is contained fully in each row.

I would personally prefer to keep it simple and just determine what to render in the formatter, if we feel this starts to clutter the code in the future we may reconsider a different mechanism, but I don't expect it to be honest.

fichtner:
> I would personally prefer to keep it simple and just determine what to render in the formatter

If you mean exactly what's done in this PR then alright.

VIPs are convoluted so this likely won't come up often, but if it does this needs to be addressed differently IMO.

Monviech:
Since grid_view defines the metadata I need for the grid, I have added the modes there. This duplicates them but overall its less code and it seems simpler.

fichtner:
Yes, that looks better 👍

swhite2:
> VIPs are convoluted so this likely won't come up often, but if it does this needs to be addressed differently IMO.

After some more thought it is indeed better to define this in the grid_view portion so I think the current state is a nice middle ground.

> If you mean exactly what's done in this PR then alright.

No, I was referring to the glue between the form metadata and grid metadata.

swhite2 (approved):

fichtner:
> No, I was referring to the glue between the form metadata and grid metadata.

Thanks for the clarification. :)

PR #8943, opened 2025-07-16
ui: Improve the picture widget on the dashboard

sopex:
This is a quality-of-life/visual improvement.

Essentially, it allows the picture to fill the whole widget size, removing the title, divider bar, and settings link.

While one could argue that the setting link is useful, I would counter that it is seldom used, and keeping the text/settings link when no picture is supplied is sufficient.

Thank you for your time!

Screenshot 2025-07-16 145532
Screenshot 2025-07-16 145449
Screenshot 2025-07-16 144443

### DIFF
--- a/src/opnsense/www/js/opnsense_widget_manager.js
+++ b/src/opnsense/www/js/opnsense_widget_manager.js
@@ -424,6 +424,9 @@ class WidgetManager  {
                 $('#restore-defaults').hide();
                 $('.title-invisible').hide();
             }
+
+            // expect layout to have shifted
+            this._updateGrid();
         });
 
         $('#edit-grid').mouseup(function() {


fichtner:
What happens on edit now? And is it still movable?

sopex:
1. Not sure what you mean
2. Yes of course, it's movable.

Monviech:
When you go into the Dashboard Edit mode, can you still remove the picture widget?

sopex:
> When you go into the Dashboard Edit mode, can you still remove the picture widget?

Of course 👍... You now can. Nice catch!

fichtner:
Thanks, the only issue I see now is that when the header is shown the height is different, which makes aligning widgets with a "headless" widget before it harder. Maybe we only want to do this optionally anyway, because otherwise default layouts change (subtly but some will notice).

Monviech:
Doing that optionally would require the widget being configurable like this:

https://github.com/opnsense/core/blob/bc82c17e53d2e8b3344fc0b4098afdf0d52b6cda/src/opnsense/www/js/widgets/Certificates.js#L32

This would make it more complicated, dunno if its worth it.

Monviech:
Maybe it would be best to have a general setting to hide all headers of all widgets or something?

sopex:
> Thanks, the only issue I see now is that when the header is shown the height is different, which makes aligning widgets with a "headless" widget before it harder. Maybe we only want to do this optionally anyway, because otherwise default layouts change (subtly but some will notice).

I don't see the problem with this, even now, the picture widget cannot align 100% with any other. But I am obviously open to your thoughts and knowledge.

I also believe that the ones that "would notice" (as you mentioned), also notice the off-center logo/picture with a description that exists today 😂

sopex:
> Maybe it would be best to have a general setting to hide all headers of all widgets or something?

Some titles are useful. Most wouldn't disable them all to fix one.

swhite2:
I think that based on the discussion a neat middle ground would be to add a property in the `baseWidget` (perhaps something like `titleVisible`), default it to true, create a public getter and use this to determine if we should render it in the widget manager. The Picture widget can then set this property to false in its constructor so, like much of the design, the decision is made in the right spot :)

If edit mode is on, the title should always be visible to ensure consistency.

@sopex if you agree, do you want to give that a try?

sopex:
@swhite2 Sure, I will try to cook something up.

sopex:
@swhite2 Do you think this would work?

swhite2:
Aren't the `titleElement` and `linkHandle` non-existent for the picture here?

If so, `$('.widget-title').show();` won't show anything in this case. Probably best to control the visiblity through `display: none` or `hide()` so the elements exist in all cases.

src/opnsense/www/js/opnsense_widget_manager.js
```diff
@@ -582,21 +588,28 @@ class WidgetManager  {
...
+        const linkHandle = widget && widget.isTitleVisible() ? link : '';
         let $header = $(`
-            <div class="widget-header">
+            <div class="widget-header"${headerStyle}>
                 <div class="widget-header-left"></div>
-                <div id="${identifier}-title" class="widget-title"><b>${title}</b></div>
+                ${titleElement}
                 <div class="widget-command-container">
-                    ${link}
+                    ${linkHandle}
```
swhite2: Aren't the `titleElement` and `linkHandle` non-existent for the picture here?

If so, `$('.widget-title').show();` won't show anything in this case. Probably best to control the visiblity through display: none or hide() so the elements exist in all cases.

sopex: Something like this?

swhite2:
@sopex This can be simplified a little further, let me push something to this PR.

swhite2:
That should be it, care to test it and see if you agree?

sopex:
> That should be it, care to test it and see if you agree?

Sure, would love to. Does it work for you, though? opnsense-patch fails for me

swhite2:
> > That should be it, care to test it and see if you agree?
> 
> Sure, would love to. Does it work for you, though? opnsense-patch fails for me

Works for me :)

FWIW, you can do `# make upgrade` in your core directory to apply the changes (assuming you are on development)

sopex:
> > > That should be it, care to test it and see if you agree?
> > 
> > 
> > Sure, would love to. Does it work for you, though? opnsense-patch fails for me
> 
> Works for me :)
> 
> FWIW, you can do `# make upgrade` in your core directory to apply the changes (assuming you are on development)

That was easier. Yes, I like it very much! Exactly what I was hoping for!
Thank you

swhite2:
Merged, thanks :)

PR #8946, opened 2025-07-17
ids: Add banner when custom.yaml is not the default stub

Monviech:
Banner can be hidden by either deleting the custom.yaml or reverting it to its original state.

Fixes: https://github.com/opnsense/core/issues/8945

### DIFF
--- a/src/opnsense/mvc/app/library/OPNsense/System/Status/IDSOverrideStatus.php
+++ b/src/opnsense/mvc/app/library/OPNsense/System/Status/IDSOverrideStatus.php
@@ -47,7 +47,7 @@ class IDSOverrideStatus extends AbstractStatus
 
     public function collectStatus()
     {
-        $fileHash = @hash_file('sha256', '/usr/local/etc/suricata/custom.yaml');
+        $fileHash = @hash_file('sha256', '/usr/local/opnsense/service/templates/OPNsense/IDS/custom.yaml');
         $sampleHash = @hash_file('sha256', '/usr/local/opnsense/service/templates/OPNsense/IDS/custom.yaml.sample');
 
         if ($fileHash && $sampleHash && $fileHash !== $sampleHash) {


fichtner:
> deleting the custom.yaml

Not that it's mission-critical, but wouldn't that block suricata from starting?

src/opnsense/mvc/app/library/OPNsense/System/Status/IDSOverrideStatus.php
```diff
@@ -0,0 +1,63 @@
...
+        $this->internalScope = [
+            '/ui/ids',
+            '/ui/ids/policy',
+        ];
+    }
+
+    public function collectStatus()
+    {
+        $customFile = '/usr/local/etc/suricata/custom.yaml';
+        $expectedHash = '0cd7bf5ab0aa582303cfdebb3d4071366b232f300ef285101467533fd54c95d7';
```
fichtner: better hash_file($customFile . 'sample')

src/opnsense/mvc/app/library/OPNsense/System/Status/IDSOverrideStatus.php
```diff
@@ -0,0 +1,63 @@
...
+
+        if (is_file($customFile)) {
```
fichtner: you probably want to prevent an error on hash_file, how about using `@hash_file` instead on both

src/opnsense/mvc/app/library/OPNsense/System/Status/IDSOverrideStatus.php
Monviech: Do you mean I should provide a custom.yaml.sample in the same path and compare its hash against custom.yaml?

src/opnsense/mvc/app/library/OPNsense/System/Status/IDSOverrideStatus.php
Monviech: sounds good :)

Monviech:
> > deleting the custom.yaml
> 
> Not that it's mission-critical, but wouldn't that block suricata from starting?

I moved and deleted the file but suricata still starts. Maybe it only crashes in very specific circumstances when it is missing?

src/opnsense/mvc/app/library/OPNsense/System/Status/IDSOverrideStatus.php
fichtner: The sample is supplied by the core package. Removing the custom.yaml will eventually bring a .sample copy back on a package update. There is a "risk" that user edits both custom.yaml and custom.yaml.sample locally but it beats triggering the banner if we ever modify custom.yaml.sample and have to change the hardcoded hash here.

fichtner:
> I moved and deleted the file but suricata still starts. Maybe it only crashes in very specific circumstances when it is missing?

No, I was just wondering about it. Nice that it doesn't crash and burn.

src/opnsense/mvc/app/library/OPNsense/System/Status/IDSOverrideStatus.php
```diff
@@ -47,17 +47,14 @@ public function __construct()
 
     public function collectStatus()
     {
-        $customFile = '/usr/local/etc/suricata/custom.yaml';
-        $expectedHash = '0cd7bf5ab0aa582303cfdebb3d4071366b232f300ef285101467533fd54c95d7';
+        $fileHash = @hash_file('sha256', '/usr/local/etc/suricata/custom.yaml');
+        $sampleHash = @hash_file('sha256', '/usr/local/opnsense/service/templates/OPNsense/IDS/custom.yaml.sample');
```
fichtner: almost, we check the template location exclusively, /usr/local/etc/suricata/custom.yaml will and should be overwritten anyway

fichtner (approved):

PR #8948, opened 2025-07-17
Feat: added IPv6 bracket helpers for templates

BPplays:
some programs want IPv6 addresses in brackets especially if it's an address and a port like `[2001:db8::1]:8080` so i made a helper for that it takes in a string and puts it it brackets if it's an IPv6 address

### DIFF
--- a/src/opnsense/service/modules/addons/template_helpers.py
+++ b/src/opnsense/service/modules/addons/template_helpers.py
@@ -151,6 +151,23 @@ class Helpers(object):
             result.append(self.getNodeByTag('interfaces.'+name+'.if'))
         return list(filter(None, result))
 
+
+    def getBracketedIp(self, str):
+        """ returns ipv6 addresses with brackets else returns original
+        :param tag: tag in dot notation (section.item)
+        :return: boolean
+        """
+        if self.is_ipv6(str):
+            return "[" + str + "]"
+        return str
+
+    @staticmethod
+    def is_ipv6(str):
+        try:
+            return ipaddress.ip_address(str).version == 6
+        except ValueError:
+            return False
+
     @staticmethod
     def getIPNetwork(network):
         """
--- a/src/opnsense/service/modules/addons/template_helpers.py
+++ b/src/opnsense/service/modules/addons/template_helpers.py
@@ -165,7 +165,7 @@ class Helpers(object):
     def is_ipv6(str):
         try:
             return ipaddress.ip_address(str).version == 6
-        except ValueError:
+        except:
             return False
 
     @staticmethod
--- a/src/opnsense/service/modules/addons/template_helpers.py
+++ b/src/opnsense/service/modules/addons/template_helpers.py
@@ -152,7 +152,7 @@ class Helpers(object):
         return list(filter(None, result))
 
 
-    def getBracketedIp(self, str):
+    def get_bracketed_ip(self, str):
         """ returns ipv6 addresses with brackets else returns original
         :param tag: tag in dot notation (section.item)
         :return: boolean


fichtner:
IMO this only happens for IP/port combinations so passing the port as well would make more sense?

src/opnsense/service/modules/addons/template_helpers.py
```diff
@@ -151,15 +151,27 @@ def physical_interfaces(self, names):
...
+        :param port: setting this < 0 will disable it's output and just output the ip formatted for inclusion with a separate formatted port
         :return: string
         """
-        if self.is_ipv6(str):
-            return "[" + str + "]"
-        return str
+        if self.is_ipv6(host):
+            host = "[" + host + "]"
+
+        if self.to_int(port) < 0:
```
fichtner: you don't want to put brackets on IPv6 if no port is added so I thin this needs to move up

src/opnsense/service/modules/addons/template_helpers.py
```diff
@@ -151,15 +151,27 @@ def physical_interfaces(self, names):
             result.append(self.getNodeByTag('interfaces.'+name+'.if'))
         return list(filter(None, result))
 
-
-    def get_bracketed_ip(self, str):
-        """ returns ipv6 addresses with brackets else returns original
-        :param str: any string that may be an IPv6 address in a place where an IPv6 address would need brackets
+    def get_host_port(self, host: str, port: str | int):
```
fichtner: Since the usage is in a plugins PR I've made a comment there:

https://github.com/opnsense/plugins/pull/4823/files/4022932e935be6606694290868218c4961db4468..cb1ee952723967a0ddd5582bd23f2d1a72fb0f71#r2215689666

src/opnsense/service/modules/addons/template_helpers.py
BPplays: @fichtner i made it like this because in case it's needed to have just an IP formatted like this

for example [Nut/upsmon.conf](https://github.com/opnsense/plugins/blob/06ea26bbb5fdfad0b392a93c4b439fcb3121ae87/sysutils/nut/src/opnsense/service/templates/OPNsense/Nut/upsmon.conf) checks if 'OPNsense.Nut.netclient.port' exists before putting a port, it doesn't format for IPv6 addresses currently but it probably wants the same address format even without the port,

also it might be easier it some cases to just get the formatted IP and add the port separately,
there are probably also other situations that IPs need to be formatted like this without a port, and i don't see a reason to use this func if your IP doesn't need this kind of formatting

src/opnsense/service/modules/addons/template_helpers.py
```diff
@@ -151,6 +151,61 @@ def physical_interfaces(self, names):
...
+        port = self.getNodeByTag(port_tag)
+
+        if port is None:
+            port = -1
+
+        return self.get_host_port(host, port, no_brackets_on_bare_ip)
+
+
+
+    def get_host_port(self, host: str, port: str | int = 0, no_brackets_on_bare_ip: bool = False):
```
fichtner: We don't want to this: `no_brackets_on_bare_ip` because the risk for syntax errors is higher wrapping an IPv6 into brackets without a port than the other way around. Per general style inverted flags are trickier and having said the first argument the risk is higher with the current default here.

I would also try to avoid doing more than one function. This library is very likely to get edited much.

Lastly, get_host_port could offer a return value based on the presence of host_tag as well returning empty if that is not found. With this we could use it for preprocessing the result and only writing results conditionally when not empty much like you handle the optional port now internally.

fichtner: also as far as names go the library here is not very consistent, but for brevity `helper.host_port` would be preferable. The other option would be `helper.getHostPort` but that style is used less frequently and apparently only used by squid plugin these days.

BPplays: @fichtner is there a reason why i should avoid adding more than one function? i feel like get_host_port being for formatting the text if you just need that and get_host_port_tag being a helper if you're using tags directly works fine

for example if you wanted to loop over list of IPs and ports in the template wouldn't you need to pass the text directly?

fichtner: Since this is tailored for address and port needing to be joined because they are separate config entities in models the one or the other may not be required in which case the template needs to do more juggling to see if the values are available and how the resulting string needs to be constructed. If there is a helper it should do this on a single line which only works by passing tags like helpers.exists does.

@Monviech noted that caddy uses this simple construct for verbatim values (caused here by a CSV due to multiple addresses and only one port)

https://github.com/opnsense/plugins/blob/06ea26bbb5fdfad0b392a93c4b439fcb3121ae87/www/caddy/src/opnsense/service/templates/OPNsense/Caddy/includeLayer4#L30-L31

I see value in reducing template noise around checking address and port, but verbatim construction is quite portable in Jinja.

BPplays: @fichtner i merged `get_host_port` and `get_host_port_tag` so it's now just 1 function that takes in 2 tags, is it more like what you wanted now?

src/opnsense/service/modules/addons/template_helpers.py
```diff
@@ -151,6 +151,50 @@ def physical_interfaces(self, names):
...
+            return host
+
+        return '{}:{}'.format(host, port)
+
+    @staticmethod
+    def to_int(s):
+        try:
+            return int(s)
+        except:
+            return -1
```
fichtner: We should trust the model data so the only cases to be handled are None and "" and this removes the need for this used-once conversion function

src/opnsense/service/modules/addons/template_helpers.py
```diff
@@ -151,6 +151,50 @@ def physical_interfaces(self, names):
             result.append(self.getNodeByTag('interfaces.'+name+'.if'))
         return list(filter(None, result))
 
+    def get_host_port(self, host_tag: str, port_tag: str, brackets_on_bare_ip: bool = False):
```
fichtner: thanks for flipping this. note: from experience I think nobody will ever use True here and the flag and documentation is still wrong because it's only doing this for IPv6 which is only required for port append to avoid parsing mistakes between ":" characters

src/opnsense/service/modules/addons/template_helpers.py
```diff
@@ -151,6 +151,50 @@ def physical_interfaces(self, names):
...
+            return -1
+
+    @staticmethod
+    def get_ip_version(ip: str) -> int:
+        """ returns an IP version or -1 if not an IP
+        """
+        try:
+            return ipaddress.ip_address(ip).version
+        except:
+            return -1
```
fichtner: I would prefer detecting ":" character for IPv6. Again we're supposed to trust the model data and rather break the config than mask the mistake in the model.

src/opnsense/service/modules/addons/template_helpers.py
BPplays: @fichtner i know but i it included because:
i assume there are some programs with a separate IP and Port field in the config but it assumes they can just be joined with no further formatting so an IP field must be bracketed even if it's just an IP address.

also some configs may also use colons for something different
it's not a config but for example `scp` uses `:` to separate the remote host and directory so you need [] for IPv6 even without a port.

even if it's not used often it seems like it might still be useful

fichtner: true. then we can argue we don't need a third parameter if we set port_tag: str = '' and the user can simply call get_host_port('::1') to get the desired behaviour?

BPplays: @fichtner im not really sure what that means, without `brackets_on_bare_ip` putting in a tag with `::1` would just return '::1' not '[::1]' and no brackets if there's no port is the more standard default

do you mean you'd rather it just default to always brackets even when there is no port specified?

fichtner: You are right. After thinking it over, how about this: a second function with only the `host_tag` as a parameter and internally it calls get_host_port(self, host_tag, '', True) -- that should be as user friendly as it gets but both functions need better naming then, perhaps `host_with_port` and `host_for_port`.

The only templating issue I see is that for the user to decide they need to call host_for_port is knowing that they definitely want a port rendered after. We could also allow the brackets parameter as a second one for the new function this time set to true. But as mentioned I would think these would be very rarely used explicitly.

fichtner:
Merged, thank you very much!

Issue #8960, opened 2025-07-20
Wireguard status permission for observers


hboetes:
**Important notices**

Before you add a new report, we ask you kindly to acknowledge the following:

- [X] I have read the contributing guidelines at https://github.com/opnsense/core/blob/master/CONTRIBUTING.md
- [X] I am convinced that my issue is new after having checked both open and closed issues at https://github.com/opnsense/core/issues?q=is%3Aissue

**Is your feature request related to a problem? Please describe.**

There is the permission for checking the status of IPSEC and OpenVPN, but no such permission exists for Wireguard

**Describe the solution you like**

To implement granting the permission to check the Wireguard status.

PR #8968, opened 2025-07-21
vpn/wireguard: Add diagnostics and log file ACL

Monviech:
Fixes: https://github.com/opnsense/core/issues/8960

There is a little overlap between the config and diagnostics ACL in UI paths.
That could be tweaked as well by limiting the ui path to `ui/wireguard/general/*` but it would change current behavior of the `page-wireguard-config` ACL.


fichtner (commented):
not my battle

@AdSchellevis should decide if we want this and "historically" we try to avoid micromanagement of MVC ACLs

Monviech:
Even if we do not want this one, the log ACL was missing so that should be added at least.

AdSchellevis:
I don't mind that much, feel free to merge

PR #8985, opened 2025-07-24
firewall/automation: Adjust firewall_rule_lookup to open correct interface and rule from firewall live log

Monviech:
The automation page has an interface hash since the last overhaul.

This change selects the correct interface, and opens the correct rule to edit.

### DIFF
--- a/src/www/firewall_rule_lookup.php
+++ b/src/www/firewall_rule_lookup.php
@@ -30,6 +30,8 @@ require_once("guiconfig.inc");
 require_once("filter.inc");
 require_once("system.inc");
 
+use OPNsense\Firewall\Filter;
+
 $a_filter = &config_read_array('filter', 'rule');
 
 $fw = filter_core_get_initialized_plugin_system();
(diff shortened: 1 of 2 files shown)


src/www/firewall_rule_lookup.php
```diff
@@ -53,7 +55,22 @@
                         header(url_safe('Location: /%s?if=%s&id=%s', [$parts['path'], $query['if'], $query['id']]));
                     }
                 } else {
-                    header(sprintf('Location: /%s', $rule->getRef()));
+                    // search model firewall rule
+                    $interface = '';
+                    foreach ((new Filter())->rules->rule->iterateItems() as $ruleItem) {
```
AdSchellevis: can be simplified into something like:

```
$node = (new Filter())->getNodeByReference('rules. rule.'.$rid);
```

src/www/firewall_rule_lookup.php
```diff
@@ -53,7 +55,22 @@
...
+                            if ($interfaceValue !== '' && strpos($interfaceValue, ',') === false) {
+                                $interface = $interfaceValue;
+                            }
+                            break;
+                        }
+                    }
+
+                    $base = strtok($rule->getRef(), '#');
+                    $hash = url_safe('interface=%s&edit=%s', [$interface, $rid]);
+                    header("Location: /{$base}#{$hash}");
```
AdSchellevis, line 71: if this is a trampoline into the mvc code, it should probably just go there with a uuid, it feels like the interface is a bit irrelevant here.

AdSchellevis (changes requested):

src/www/firewall_rule_lookup.php
Monviech: It can indeed open the UUID directly, but then the current interface the rule belongs to will not be selected, so the grid will default to the "Floating" ruleset as entry point.

AdSchellevis: I see your point, in that case this might indeed be the shortest path (combined with a direct fetch of the node)

src/www/firewall_rule_lookup.php
Monviech: I have used the above, thanks for the hint.

AdSchellevis (approved):

PR #8993, opened 2025-07-25
bootgrid: add default row count field, and increase default max rows

BPplays:
this PR:
- adds a `defaultRowCount` option to bootgrid
  - with logic to add `defaultRowCount` to `rowCount` if it's not there
- changes the default values for `rowCount`
- adds `_getFriendlyRowCount(count)` to be used when displaying row count

Fixes: #8913

### DIFF
--- a/src/opnsense/www/js/opnsense_bootgrid.js
+++ b/src/opnsense/www/js/opnsense_bootgrid.js
@@ -197,6 +197,17 @@ class UIBootgrid {
     }
 
     initialize() {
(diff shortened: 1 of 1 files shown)


fichtner:
@swhite2 will take care of this, but likely not via PR; also see https://github.com/opnsense/core/pull/8933

Monviech:
We opted for a simpler solution as a first step:

https://github.com/opnsense/core/commit/baa1730b1a9f1d8c9f0bb81a0ff1521636824231

Thank you for your effort in this PR.

PR #9015, opened 2025-07-29
firewall/automation: tabulator tree view poc

Monviech:
image

Fixes: https://forum.opnsense.org/index.php?topic=2851

The general logic of this is simple yet effective:

- The sequence of rendered rules stays completely intact
- Rules can have one or multiple categories
- Before each rule, a bucket row will generate with the category(s) of its first child copied
- Each rule from now on that changes categories will render a new bucket row, if not it will appear below the previous bucket row as additional rule
- Rules without a category will appear under a default bucket row
- Since bucket rows are normal data-rows, they adhere to the sequence, and can be duplicates without any hacks

This means, all "rule separators" are automatically generated, and only influenced by two properties:

- The category name(s) of a rule
- The sequence number of a rule
- The view is optional

### DIFF
--- a/src/opnsense/mvc/app/views/OPNsense/Firewall/filter_rule.volt
+++ b/src/opnsense/mvc/app/views/OPNsense/Firewall/filter_rule.volt
@@ -64,6 +64,7 @@
             $('#toggle_tree_button').addClass('active btn-primary');
         }
 
+        // Lives outside the grid, so the logic of the response handler can be changed after grid initialization
         function dynamicResponseHandler(resp) {
             // convert the flat rows into a tree view (if enabled)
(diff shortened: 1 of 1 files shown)


swhite2:
One of the things that's nice about this is the little amount of code necessary to do this.

Note to self: I'm not a big fan of how the direct Tabulator options now also need to pass through the compatibility layer (this is the only way curently), but this should be rather easy to fix.

Monviech:
@swhite2 It takes so little code because it leverages all the work you did with tabulator over many months :)

Monviech:
It's good enough for master, its optional, no bugs I noticed so far. Everybody saw what it does.

Issue #8857, opened 2025-06-25
OpenVPN Instance addresses allocation vs Client Specific Overrides


borysohnsorge:
**Important notices**

Before you add a new report, we ask you kindly to acknowledge the following:

- [X] I have read the contributing guide lines at https://github.com/opnsense/core/blob/master/CONTRIBUTING.md
- [X] I am convinced that my issue is new after having checked both open and closed issues at https://github.com/opnsense/core/issues?q=is%3Aissue

Similar (closed without resolution) issue: https://github.com/opnsense/core/issues/8600

**Describe the bug**

I have an OpenVPN instance configured with the network 10.255.248.0/22. I’ve set up client-specific overrides assigning IPs from 10.255.248.2 to 10.255.248.190. 

Client "A" connects and has the address set to 10.255.248.2 in client-specific overrides, then Client "B" connects that doesn't have reservation in client-specific overrides and gets as well 10.255.248.2 address etc.

We have a situation where two users connected to OpenVPN Instance have same ip address and that leads to connection conflict/problem.

**To Reproduce**

1. Configure OpenVPN Instance (for eg with 192.168.10.0/24 network). Serwer itself will take 192.168.10.1
2. Configure 2 users
3. Make "reservations" for first user in client-specific overrides and set IP to 192.168.10.2
4. Connect User with this ip reservation
5. Connect second User
6. Look at VPN->OpenVPN->Connection Status, both Users will have the same ip address.

**Expected behavior**

Second User will get "next free" address from defined network

**Screenshots**

**Environment**

Software version used and hardware type if relevant, e.g.:

OPNsense version:
OPNsense 25.4.1-amd64 (Business Edition)
FreeBSD 14.2-RELEASE-p3
OpenSSL 3.0.16
Licensed until 2026-01-26
OPNsense Business Support Subscription

System type (physical, virtual, cloud)
Physical server

System manufacturer/model (or virtualization software / cloud stack used)
2x Supermicro 1U Server e3-1230v5, x11-ssl, 32GB Ram, 4x 3TB HDD (RAID 10), NIC 2x 25Gbps (Intel)

Monviech:
Hello @borysohnsorge 

I have recreated the issue that you have, two clients received the same virtual IP address from the pool.
Server was 192.168.101.1/24
Client-A had CSO for 192.168.101.2
Client-B had no CSO

When first Client-A connected, and afterwards Client-B connected, they both received 192.168.101.2.

[...]

PR #9025, opened 2025-08-01
vpn/openvpn: Add nopool directive

Monviech:
A mix of dynamic and cso enabled clients can cause IP duplication in the pool. With nopool, only CSO will manage all addresses, clients without a CSO will not receive an IP address.

Fixes: https://github.com/opnsense/core/issues/8857

### DIFF
--- a/src/opnsense/mvc/app/models/OPNsense/OpenVPN/OpenVPN.php
+++ b/src/opnsense/mvc/app/models/OPNsense/OpenVPN/OpenVPN.php
@@ -600,7 +600,7 @@ class OpenVPN extends BaseModel
                             $options['ifconfig'] = "{$ip1} {$ip2}";
                             $options['ifconfig-pool'] = "{$ip2} {$ip3}";
                         } else {
-                            $options['server'] = $parts[0] . " " . $mask;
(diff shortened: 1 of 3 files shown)


src/opnsense/mvc/app/models/OPNsense/OpenVPN/OpenVPN.php
```diff
@@ -600,7 +600,7 @@ public function generateInstanceConfig($uuid = null)
                             $options['ifconfig'] = "{$ip1} {$ip2}";
                             $options['ifconfig-pool'] = "{$ip2} {$ip3}";
                         } else {
-                            $options['server'] = $parts[0] . " " . $mask;
+                            $options['server'] = $parts[0] . " " . $mask . ($node->nopool == '1' ? ' nopool' : '');
```
fichtner:
```suggestion
                            $options['server'] = $parts[0] . " " . $mask;
                            if ($node->nopool->isEqual('1')) {
                                $options['server'] .=  ' nopool';
                            }
```

something like this is easier to maintain and avoids implicit casting

fichtner (approved):
only edited the code in github, make sure it works

Monviech:
Tested and works. Thanks for your help.

PR #9043, opened 2025-08-04
dnsmasq: Add tabulator groupBy functionality to group by interfaces and tags

Monviech:
Groups by interface on first level, and by tags on second level (if they exist)

image


Monviech:
Note to me:

Grouping tags in level 2 adds a lot of noise. Just grouping interfaces will be enough to improve the consistency of ranges and options, as most people use interfaces.

swhite2 (approved):

Issue #8952, opened 2025-07-18
unbound: per-subnet host overrides using views


stuart-mclaren:
**Important notices**

Before you add a new report, we ask you kindly to acknowledge the following:

- [x] I have read the contributing guide lines at https://github.com/opnsense/core/blob/master/CONTRIBUTING.md
- [x] I am convinced that my issue is new after having checked both open and closed issues at https://github.com/opnsense/core/issues?q=is%3Aissue

Add ability to override hosts per interface.

This has mentioned before, eg here:

https://github.com/opnsense/core/issues/6789

I'm going to suggest an approach that (hopefully) does not change the existing code/UI
too much.

**Describe the solution you like**

Current behaviour:

In unbound->Overrides->Host Overrides
you can currently 

1. override individual hosts

for example:

foo.example.com -> 10.0.0.1

2. wildcard overrides

for example:

*.example.com -> 10.0.0.2
(this means anything.example.com gets resolved to 10.0.0.2)

These end up as the following unbound config:

1. foo.example.com ->10.0.0.1

```
local-data: "example.com A 192.168.1.100"
```

2. *.example.com -> 10.0.0.2

```
local-zone: "example.com" redirect
local-data: "example.com A 10.0.0.2"
```

The proposal would add the following

3. bar.example.com -> 10.0.0.3 interface LAN

In the UI, an additional drop down box would list current interfaces
when creating a host override. Here the user has selected LAN from
the drop down (default would be "all interfaces", existing behaviour).

the generated unbound config would be (approximately):

```
# "eth0" deduced from LAN, not entered by user
interface-tag: eth0 "LAN"
interface-tag-data:  eth0 "LAN" "bar.example.com A 10.0.0.3"
```

4. *.example.com -> 10.0.0.4 interface OPT

Here the user has selected OPT and entered * into the host field.

The generated unbound config would be (approximately):

```
local-zone: "example.com" redirect
interface-tag: eth1 "OPT"
interface-tag-data: eth1 "OPT" "example.com A 10.0.0.4"
```

**Describe alternatives you considered**

unbound 'views' - hard to isolate more general purpose 'view' configuration.
Feels like adding it to existing host overrides would not be the way to go.

**Additional context**

 I think most people currently do this with views, but tags seem a potentially
easier way to extend the current host overrides without changing too much.

Note: I have not yet tested the tag based configuration - I've just read the unbound man pages.

Would need to gracefully handle the cases where an interface is deleted or 
[...]

PR #9055, opened 2025-08-05
unbound: add basic split view DNS support

stuart-mclaren:
Add the ability to return different DNS records based on the client's subnet.

For example, for host `foo.example.com` clients from `10.0.0.0/8` may get `1.1.1.1` where clients from
`192.168.0.0/16` may get `2.2.2.2`, etc.

The new implementation is kept as separate as possible from existing Unbound code. Eg the model section, controllers and new UI tabs, are all separated. This means that

1. Existing code/state is minimally affected
2. Users who are not interested in (more advanced) view functionality see no UI changes to existing, familiar menus, and do not have to spend time thinking about new view options.
3. Corner cases are removed - the separate model section avoids solving for these cases in logic
4. Support is easier: because view logic does not overlap with existing logic (e.g. separate unbound configuration file generation logic, separate API controllers, separate xml state) fault isolation should be easier.

Note that unbound's view local-data (host) records are slightly different than standard local-data (host) entries:

```
view:
  local-data: "ccc.example.com. A 3.3.3.3"
```

compared with

```
server:
  local-data: "ttt.example.com  IN A 7.7.7.7"
```

The new `Split View DNS` menu under unbound has three UI tabs:

 "Views", "Subnets", "Hosts".

These match the following flow:
Users create a View, create a Subnet to scope the view to, then assign Hosts to that View.

A Host can be assigned to multiple Views. A Subnet can be assigned to a single View; multiple Subnets can be assigned to the same View. This behaviour is reflected in the model (and validation).

Subnets cannot overlap - this behaviour is recommended and means that View order does not matter.

If a View is deleted, Hosts are no longer assigned to that view. If the Host was assigned to multiple Views it remains assigned to the remaining existing views. Otherwise it it is unassigned, and can be associated with a different View if desired. (It does not become part of the "server:" section - that would be a potentially dangerous side effect.)

The Hosts table has a Views dropdown which can be used to filter Hosts based on View. This dropdown and all tables update automatically if changes are made (eg if a View name is changed that will be reflected automatically in the Hosts table, the Hosts filter, and the Subnets table).

There is no support for interface based views, but the design would easily support adding that. (Interfaces would be assigned to views in the same way as subnets.)

Reverse records are created automatically. Only A/AAAA record types are supported. Wildcards are supported using `*`.

Updates are saved atomically. E.g. If a view is deleted, it is removed from the Host state and the view state atomically, so there should be no stale references.

For some design discussion, and UI screenshots see:

 https://github.com/opnsense/core/issues/8952


fichtner (commented):
I don't want to stop you from doing this but I have to point out that the sheer volume of this PR makes swift inclusion unlikely. Personally this feels like overkill and puts the burden on the project when bugs are reported here, which is likely going to happen (same for the code that we introduce).

eriklogtenberg:
+1 for this feature though 👍

stuart-mclaren:
Should I close this?

stuart-mclaren:
Closing because this looks like it is not going to happen.

PR #9058, opened 2025-08-07
dnsmasq: Swap Hosts and Domains tab for consistency reasons

Monviech:
Hosts is more related to DHCP than Domains.

In Hosts, a selectpicker for tags is shown, in Domains hidden.

This is just something that personally bugs me when clicking around.

Clicking from the Tab "Hosts" to "Domain" to "DHCP ranges" just feels kinda inconsistent.

PR can be ignored if too much micro management.


src/opnsense/mvc/app/models/OPNsense/Dnsmasq/Menu/Menu.xml
```diff
@@ -2,8 +2,8 @@
     <Services>
         <Dnsmasq VisibleName="Dnsmasq DNS &amp; DHCP" cssClass="fa fa-tags fa-fw">
           <General order="10" url="/ui/dnsmasq/settings#general"/>
+          <Domains order="15" url="/ui/dnsmasq/settings#domains"/>
```
fichtner: Can you shift to 20 and 30 as before?

PR #9062, opened 2025-08-08
dnsmasq: Allow disabling local for DHCP domains

Monviech:
If dnsmasq is the first DNS server in line, and you do not want to use a unique DHCP domain, forcing the DHCP domain to be local breaks all upstream queries for it. This might be wanted if you forward from Unbound to dnsmasq since it fixes DNS loops, but if dnsmasq is your primary forwarding resolver, it knows what to forward and what to answer.

Mostly from reddit and opnsense forum feedback, it caught more than one user offguard.

Most recent example: https://www.reddit.com/r/opnsense/comments/1mk16y1/dnsmasq_not_forwarding_local_domain_queries/


Monviech:
Default behavior stays the same, just with optional checkbox to disable the magic.

PR #9068, opened 2025-08-10
Improve live mode greeting for first-time user clarity

lilchancep:
Added the default installation password to the live media display. This change addresses repeated instances of user uncertainty during installation by making the password easily visible.

And by "Users" I mean me..

### DIFF
--- a/src/etc/rc.subr.d/livemode
+++ b/src/etc/rc.subr.d/livemode
@@ -63,7 +63,7 @@ if (is_install_media()) {
 
         $greeting = "Welcome!  {$namex} is running in live mode from install media.  Please\n" .
                     "login as 'root' to continue in live mode, or as 'installer' to start the\n" .
-                    "installation.  Use the default or previously-imported root password for\n" .
+                    "installation.  Use the default password 'opnsense' or previously-imported root password for\n" .
                     "both accounts.";
 
         if (!isset($config['system']['ssh']['noauto']) && is_process_running('sshd')) {


fichtner:
Hardcoding makes branding harder. Softcoding requires to hardcode this information somewhere. We only have a salted hash in the default config.xml and it can be changed for whatever reason we know or don’t know about. If it were this easy it would have been done much earlier.

JTM-rootstorm:
> Hardcoding makes branding harder. Softcoding requires to hardcode this information somewhere. We only have a salted hash in the default config.xml and it can be changed for whatever reason we know or don’t know about. If it were this easy it would have been done much earlier.

I'm usually against "what's one more small thing that has to be changed when X has to be done" kind of stuff, especially when I'm not familiar with what that entails with a particular codebase, but I figure being able to tell the end user the default password (or that there is no password) without them having to scrawl the vendor site would be a decent QoL addition. 

I don't think another bit of text on the installer is going to be too much for someone who's going out of their way to rebrand an _entire OS_, especially if they're going to change/remove the password, as long as it's documented that a change has to be made in this file (and they may be already wanting to change the greeter anyways to align with branding).

I've been in a tier 2 support role long enough to realize that you could have all the neon signs pointing to basic information or knowledge articles on your site and the user will still blaze past it or ignore it for what they're trying to get to (case in point: I'm one of them, hypocrite that I am). 

Apologies if this comes off as rambling, I tend to be a very stream-of-consciousness typist.

**TL;DR** I can get behind extensibility/flexibility for re-branding but I think that drive would be going too far as to sacrifice user accessibility/ease-of-use to gain "not having to hard-change another bit of text in the install environment if the password changes". The rest of the install and initial configuration isn't that difficult -- starting that process should be straightforward as well

fichtner:
I've given this more thought and I think this isn't as straight forward as it seems. We've done a lot of well-intentioned changes that caused issues later on and I think this will be one of those.

1. In decades of project history I've never heard someone say the default password was not known to them. When we sell hardware we add a quick start guide sheet and I checked https://docs.opnsense.org/manual/install.html which mentions the default password. The document is linked in every release announcement we do offer images with.
2. A lot of thought went into crafting this tiny sequence of words to make it as concise as possible while not implying something to be true like what the default password may be. This goes beyond simple "but when the user doesn't know" as the next one may claim "but it said it had _that_ password" when it had not for whatever reason.
3. Users have varying reasons not to use OPNsense and the default password is definitely not one of them. In reality this doesn't affect adoption rate as you may imply.

It happens that people mix up passwords or cannot type the actual password correctly for reasons beyond our scope. So this adds nothing of value.


Cheers,
Franco

lilchancep:
Thank you for your detailed response. I realize my suggestion came from not fully understanding the full picture involved in what I thought was making "A one word change". 

It was based on a personal experience in a less than ideal situation. The first time I installed on my own hardware I had not read the manual, had already taken my internet offline, and had to look things up on my phone. That is really the only “complaint,” if it can be called that.

Credit where it is due, the installation process for OPNsense is very straightforward.

Thanks again for your time.

PR #9069, opened 2025-08-10
Fix OpenVPN client config for MikroTik OpenVPN server

fabianfrz:
MikroTik provides OpenVPN with a very old OpenSSL version. If it is used as a server, the OPNsense version cannot be configured to connect to it as the required cipher cannot be selected on the web interface.

Also the form hides those entries from the user in the client configuration.

Also improve the UI slightly. Data Ciphers and Data Ciphers Fallback use different select styles currently, which makes it look strange.

### DIFF
--- a/src/opnsense/mvc/app/controllers/OPNsense/OpenVPN/forms/dialogInstance.xml
+++ b/src/opnsense/mvc/app/controllers/OPNsense/OpenVPN/forms/dialogInstance.xml
@@ -337,7 +337,7 @@
         <label>Data Ciphers</label>
         <type>select_multiple</type>
         <advanced>true</advanced>
-        <style>selectpicker role role_server</style>
+        <style>selectpicker</style>
         <help>Restrict the allowed ciphers to be negotiated to the ciphers in this list.</help>
         <grid_view>
             <visible>false</visible>
@@ -346,9 +346,9 @@
     <field>
         <id>instance.data-ciphers-fallback</id>
         <label>Data Ciphers Fallback</label>
-        <type>dropdown</type>
+        <type>select_multiple</type>
         <advanced>true</advanced>
-        <style>selectpicker role role_server</style>
+        <style>selectpicker</style>
         <help>
             Configure a cipher that is used to fall back to if we could not determine which cipher the peer is willing to use.
             This option should only be needed to connect to peers that are running OpenVPN 2.3 or older versions,
--- a/src/opnsense/mvc/app/models/OPNsense/OpenVPN/OpenVPN.xml
+++ b/src/opnsense/mvc/app/models/OPNsense/OpenVPN/OpenVPN.xml
@@ -238,6 +238,7 @@
                     <OptionValues>
                         <AES-256-GCM>AES-256-GCM</AES-256-GCM>
                         <AES-128-GCM>AES-128-GCM</AES-128-GCM>
+                        <AES-256-CBC>AES-256-CBC</AES-256-CBC>
                         <CHACHA20-POLY1305>CHACHA20-POLY1305</CHACHA20-POLY1305>
                     </OptionValues>
                 </data-ciphers>
@@ -245,6 +246,7 @@
                     <OptionValues>
                         <AES-256-GCM>AES-256-GCM</AES-256-GCM>
                         <AES-128-GCM>AES-128-GCM</AES-128-GCM>
+                        <AES-256-CBC>AES-256-CBC</AES-256-CBC>
                         <CHACHA20-POLY1305>CHACHA20-POLY1305</CHACHA20-POLY1305>
                     </OptionValues>
                 </data-ciphers-fallback>


src/opnsense/mvc/app/controllers/OPNsense/OpenVPN/forms/dialogInstance.xml
```diff
@@ -337,7 +337,7 @@
         <label>Data Ciphers</label>
         <type>select_multiple</type>
         <advanced>true</advanced>
-        <style>selectpicker role role_server</style>
+        <style>selectpicker</style>
```
AdSchellevis: you may omit the selectpicker as its default (when no style is offered)

src/opnsense/mvc/app/controllers/OPNsense/OpenVPN/forms/dialogInstance.xml
```diff
@@ -346,9 +346,9 @@
     <field>
         <id>instance.data-ciphers-fallback</id>
         <label>Data Ciphers Fallback</label>
-        <type>dropdown</type>
+        <type>select_multiple</type>
         <advanced>true</advanced>
-        <style>selectpicker role role_server</style>
+        <style>selectpicker</style>
```
AdSchellevis: same

src/opnsense/mvc/app/controllers/OPNsense/OpenVPN/forms/dialogInstance.xml
```diff
@@ -346,9 +346,9 @@
     <field>
         <id>instance.data-ciphers-fallback</id>
         <label>Data Ciphers Fallback</label>
-        <type>dropdown</type>
+        <type>select_multiple</type>
```
AdSchellevis: when using a multi select, the backend needs to be changed as (model and probably implementation)

sopex:
Why don't you use the aes256-gcm that mikrotik provides and openvpn recommends as default. 

CBC doesn't provide authentication and has to be handled by mikrotik...

fabianfrz:
@sopex depends on the ROS version maybe. The version of ROS 6 I am contacting, does not support GCM at all.

sopex:
> @sopex depends on the ROS version maybe. The version of ROS 6 I am contacting, does not support GCM at all.

Yes, you are correct, it was added 2-3 years ago in 7 something. ;)

src/opnsense/mvc/app/controllers/OPNsense/OpenVPN/forms/dialogInstance.xml
fabianfrz: It is probably already implemented that way:

https://github.com/opnsense/core/blob/c756be09707aad972dd32a302d2a6150f9d77e2c/src/opnsense/mvc/app/models/OPNsense/OpenVPN/OpenVPN.php#L732

However now since I tested it, it looks like this ciphers in the config should be called cipher because it does not support multiple ciphers as it seems.

Making it a standard dropdown is the best way to handle thing ui wise then.

src/opnsense/mvc/app/models/OPNsense/OpenVPN/OpenVPN.xml
```diff
@@ -244,13 +238,15 @@
                     <OptionValues>
                         <AES-256-GCM>AES-256-GCM</AES-256-GCM>
                         <AES-128-GCM>AES-128-GCM</AES-128-GCM>
+                        <AES-256-CBC>AES-256-CBC</AES-256-CBC>
```
fichtner:
```suggestion
                        <AES-256-CBC>AES-256-CBC</AES-256-CBC>
                        <AES-128-GCM>AES-128-GCM</AES-128-GCM>
                        <AES-256-GCM>AES-256-GCM</AES-256-GCM>
```

src/opnsense/mvc/app/models/OPNsense/OpenVPN/OpenVPN.xml
```diff
@@ -242,15 +236,17 @@
...
                         <AES-128-GCM>AES-128-GCM</AES-128-GCM>
+                        <AES-256-GCM>AES-256-GCM</AES-256-GCM>
                         <CHACHA20-POLY1305>CHACHA20-POLY1305</CHACHA20-POLY1305>
                     </OptionValues>
                 </data-ciphers>
                 <data-ciphers-fallback type="OptionField">
                     <OptionValues>
                         <AES-256-GCM>AES-256-GCM</AES-256-GCM>
                         <AES-128-GCM>AES-128-GCM</AES-128-GCM>
+                        <AES-256-CBC>AES-256-CBC</AES-256-CBC>
```
fichtner:
```suggestion
                        <AES-256-CBC>AES-256-CBC</AES-256-CBC>
                        <AES-128-GCM>AES-128-GCM</AES-128-GCM>
                        <AES-256-GCM>AES-256-GCM</AES-256-GCM>
```

fichtner:
Merged, thanks!

Issue #9070, opened 2025-08-10
OTP QR code contains &amp; instead of & in URI, causing some authenticator apps (e.g., 2FAS) to reject the code


plmarcus:
When enabling or resetting OTP (Two-Factor Authentication) for a user in OPNsense, the QR code generated on the System ‣ Access ‣ Users page contains HTML-escaped ampersands (&amp;) in the otpauth:// URI instead of literal &.

While Google Authenticator still works, some apps — such as 2FAS — fail to parse the code correctly or display incorrect account details because they interpret the &amp; literally.

### To Reproduce
Log in to OPNsense 25.7.1_1 as an administrator.

Navigate to System ‣ Access ‣ Users.

Edit or create a user account.

Enable OTP seed and click "Generate new seed".

Scan the displayed QR code in the 2FAS authenticator app.

Observe that OTP setup fails or is misconfigured.
(Google Authenticator will still accept the QR code.)

### Expected behavior
The QR code should encode a valid otpauth:// URI containing unescaped ampersands (&) between parameters.
Example expected format:

`otpauth://totp/Example:user?secret=ABC123&issuer=Example`
### Actual behavior
The QR code encodes the URI with HTML entities (&amp;) instead of literal ampersands:

`otpauth://totp/Example:user?secret=ABC123&amp;issuer=Example`
Authenticator apps that strictly parse the URI (like 2FAS) interpret the entity literally, breaking OTP setup.

### Affected versions
OPNsense 25.7.1_1 (confirmed)

Likely affects other recent versions using /usr/local/opnsense/mvc/app/views/OPNsense/Auth/user.volt

### Temporary workaround
I modified user.volt to decode &amp; before generating the QR code:

```
const otp_raw = data.otp_uri_template
    .replace('|USER|', $("#user\\.name").val())
    .replace(/&amp;/g, '&');
$('#otp_qrcode').empty().qrcode(otp_raw);

$('#otp_qrcode').qrcode($("#user\\.otp_uri").val().replace(/&amp;/g,'&'));
```
Patch file:

```
--- /usr/local/opnsense/mvc/app/views/OPNsense/Auth/user.volt
+++ /usr/local/opnsense/mvc/app/views/OPNsense/Auth/user.volt
@@ -146,7 +146,10 @@
                 ajaxGet('/api/auth/user/new_otp_seed', {}, function(data){
                     if (data.seed) {
                         $("#user\\.otp_seed").val(data.seed);
[...]

PR #9071, opened 2025-08-10
Fix OTA QR generation for non lenient apps

sopex:
Fixes: https://github.com/opnsense/core/issues/9070#issuecomment-3172787596


sopex:
Already fixed at: dfa428d606260303dd2e6618ccbc8049f83808a6

Issue #9078, opened 2025-08-12
vpn/ipsec: Make before break is default in strongswan 6.0.0


Monviech:
**Important notices**

Before you add a new report, we ask you kindly to acknowledge the following:

- [x] I have read the contributing guide lines at https://github.com/opnsense/core/blob/master/CONTRIBUTING.md
- [x] I am convinced that my issue is new after having checked both open and closed issues at https://github.com/opnsense/core/issues?q=is%3Aissue

**Is your feature request related to a problem? Please describe.**

https://github.com/strongswan/strongswan/releases/tag/6.0.0

make_before_break is default in strongswan 6.0.0

**Describe the solution you like**

https://github.com/opnsense/core/blob/b2a2059572049d559aba51caffd9f8d7e5214b39/src/opnsense/mvc/app/models/OPNsense/IPsec/IPsec.xml#L63

We can probably drop this, or change the logic so it can be disabled.

PR #9080, opened 2025-08-12
vpn/ipsec: Add default value to make_before_break that retains original behavior before strongswan 6.0.0

Monviech:
Fixes: https://github.com/opnsense/core/issues/9078


fichtner:
sounds logical, but @AdSchellevis should also weigh in

AdSchellevis (approved):

PR #9087, opened 2025-08-13
dnsmasq: Add leases widget that shows latest leases

Monviech:


src/opnsense/www/js/widgets/DnsmasqLeases.js
```diff
@@ -0,0 +1,171 @@
...
+    async onWidgetTick() {
+        $('[data-toggle="tooltip"]').tooltip({container: 'body'});
+
+        const settingsResponse = await this.ajaxCall(`/api/dnsmasq/settings/${'get'}`);
+        if (settingsResponse?.dnsmasq?.enable !== '1') {
+            this.displayError(this.translations.unconfigured, '/ui/dnsmasq/settings#general');
+            return;
+        }
+
+        const leasesResponse = await this.ajaxCall(`/api/dnsmasq/leases/${'search'}`);
```
swhite2: we can probably push the `limit` logic back to the controller (`POST` + `rowCount` property)

Issue #9084, opened 2025-08-12
"advanced mode" and "full help" toggle from keyboard


dkeiht83:
**Important notices**

Before you add a new report, we ask you kindly to acknowledge the following:

- [x ] I have read the contributing guide lines at https://github.com/opnsense/core/blob/master/CONTRIBUTING.md
- [ x] I am convinced that my issue is new after having checked both open and closed issues at https://github.com/opnsense/core/issues?q=is%3Aissue

**Is your feature request related to a problem? Please describe.**

I Use Keyboard to interact with the web interface as much as possible.
The toggles for "advanced mode" and "full help" are accessible with tab to the item, however there appears to be no way to toggle with the keyboard.
I use a program that replays keyboard actions  to program the firewall.

**Describe the solution you like**

"advanced mode" and "full help" toggles  should be togglable with spacebar

**Describe alternatives you considered**

Use my mouse to click it.

sopex:
I did some testing of this and have two different PRs ready. 
Essentially, you need these attributes to every element: tabindex="0" role="switch" aria-checked="false"

opnsense.js can do this automatically, or we need to "manually" (with a script), add them to every 
"advanced mode" and "full help" elements of the core and plugins.

The problem with opnsense.js is that you need to tab two times before you can toggle it with the spacebar, one to go there and one to focus on it.

Any other ideas?

fichtner:
(a) and (h) may be useful shortcuts, especially (a). I've thought about this a few times too. Introducing this for MVC pages is probably pretty easy. We don't have to fix every legacy page (and of those there were other concepts of "advanced" toggles).

As long as the tabindex is not landing in an input box by default defanging a potential (a) and (h) toggle, or we can leverage a key combination with CTRL, although that would quickly overlay a system or browser default.

sopex:
> (a) and (h) may be useful shortcuts, especially (a). I've thought about this a few times too. Introducing this for MVC pages is probably pretty easy. We don't have to fix every legacy page (and of those there were other concepts of "advanced" toggles).
> 
> As long as the tabindex is not landing in an input box by default defanging a potential (a) and (h) toggle, or we can leverage a key combination with CTRL, although that would quickly overlay a system or browser default.

Much better idea. Will try to implement it.

sopex:
@dkeiht83 I went with the a and h shortcuts. Feel free to test the patch and let me know if it doesn't cover your needs.

PR #9091, opened 2025-08-13
gui: keyboard shortcuts for advanced and help

sopex:
Closes: https://github.com/opnsense/core/issues/9084

This would be useful both in no mouse situations and in boosting productivity, enabling users to toggle quickly between states.

"a" = advanced mode

"h" = full help

### DIFF
--- a/src/opnsense/www/js/opnsense_ui.js
+++ b/src/opnsense/www/js/opnsense_ui.js
@@ -496,14 +496,17 @@ function initGlobalOpenShortcuts() {
         const tag = (t.tagName || '').toLowerCase();
         if (tag === 'input' || tag === 'textarea' || tag === 'select' || t.isContentEditable) return;
 
+        const $context = $('.modal:visible, .ui-dialog:visible, [role="dialog"]:visible').last();
+        const searchContext = $context.length > 0 ? $context : $(document);
+
         if (e.key === 'a' || e.key === 'A') {
-            const $adv = $('[id*="show_advanced"]').first();
+            const $adv = searchContext.find('[id*="show_advanced"]').first();
             if ($adv.length) {
                 $adv.click();
                 e.preventDefault();
             }
         } else if (e.key === 'h' || e.key === 'H') {
-            const $help = $('[id*="show_all_help"]').first();
+            const $help = searchContext.find('[id*="show_all_help"]').first();
             if ($help.length) {
                 $help.click();
                 e.preventDefault();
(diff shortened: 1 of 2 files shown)


src/opnsense/www/js/opnsense_ui.js
```diff
@@ -486,6 +486,32 @@ function initFormAdvancedUI() {
...
+        if (tag === 'input' || tag === 'textarea' || tag === 'select' || t.isContentEditable) return;
+
+        if (e.key === 'a' || e.key === 'A') {
+            const $adv = $('[id*="show_advanced"]').first();
+            if ($adv.length) {
+                $adv.click();
+                e.preventDefault();
+            }
+        } else if (e.key === 'h' || e.key === 'H') {
+            const $help = $('[id*="show_all_help"]').first();
```
swhite2: I wonder if this will open the full help of the main form when a modal is open on top of it (when the user meant to open the full help of the modal). Case in point: unbound -> query forwarding -> open dialog for a new entry, press `h`

sopex: At the current form, the answer is both. That's kind of a unique scenario actually. Give me some time to try and cook something up.

sopex:
Done @swhite2

swhite2:
@sopex Thanks!

sopex:
@fichtner Can we see this shipped? Thanks :)

fichtner:
@sopex planning on shipping this in .6 -- Kea 3.0.1, Suricata 8.0.1 and pkg 2.3.1 as well as 25.10 have taken priority.

This also doesn't work on the legacy pages yet and I wanted to see if I can add it there as well without a lot of code.


Cheers,
Franco

sopex:
> @sopex planning on shipping this in .6 -- Kea 3.0.1, Suricata 8.0.1 and pkg 2.3.1 as well as 25.10 have taken priority.
> 
> This also doesn't work on the legacy pages yet and I wanted to see if I can add it there as well without a lot of code.
> 
> 
> Cheers,
> Franco

Obviously it's not a priority. Keep up the great releases! 

Can you provide a page example that it doesn't work?

fichtner:
System: Settings: Administration for example. Basically all .php pages I presume. But that's only for help -- there is no unified advanced concept there.

sopex:
> System: Settings: Administration for example. Basically all .php pages I presume. But that's only for help -- there is no unified advanced concept there.

Would you prefer to load opnsense_ui.js on all legacy pages? Fix it only for those that load it? Or "duplicate" the logic on opnsense_legacy.js ? :)

fichtner:
A copy is preferable. Eventually the copy will be removed with the last legacy files.

PR #9097, opened 2025-08-14
firewall/automation: Automation tree view improvement, mostly persistence handling and style

Monviech:
Improves upon the concept introduced here:

https://github.com/opnsense/core/pull/9015

Current changes:

- Bucket rows get unique UUIDs based on UUID of first child
- Persistence handling makes sure open rows are remembered
- A count snapshot system opens a row if the amount inside it changes during the lifetime of the document
- Fix persistence handling of Inspect and Tree Button, do not enable by default if key does not exist in local storage yet
- Improved style of bucket rows
- Expand/Collapse All button added when Tree View is active

Still an issue, but needs to be addressed later:

- Whenever the table rerenders the scroll position changes to top.

### DIFF
--- a/src/opnsense/www/js/opnsense_bootgrid.js
+++ b/src/opnsense/www/js/opnsense_bootgrid.js
@@ -1824,6 +1824,8 @@ class UIBootgrid {
                         } else {
                             $("#" + editDlg).change();
                         }
+                        this.expandGroupBy(editDlg);
+                        this.expandDataTree();
                         this._reload(true);
                         this.showSaveAlert(event);
                         saveDlg.find('i').removeClass("fa fa-spinner fa-pulse");


src/opnsense/www/js/opnsense_bootgrid.js
```diff
@@ -843,8 +855,24 @@ class UIBootgrid {
...
+                    row.treeExpand?.(); // no-op if already expanded
+                }
+                const kids = row.getTreeChildren?.();
+                if (kids && kids.length) queue.push(...kids);
+            }
+        }
+
         // backwards compat
         this.$element.trigger("loaded.rs.jquery.bootgrid");
+
```
Monviech:
```suggestion
```

src/opnsense/www/js/opnsense_bootgrid.js
```diff
@@ -843,6 +855,21 @@ class UIBootgrid {
             $last_btn.show();
         }
 
+        // Tree does not have a groupStartOpen style hook, so we must restore expansion after data is loaded
+        if (this.table?.options?.dataTree && this.rememberedTreeIds?.size) {
```
swhite2: some of the optional chaining operators can be scrapped here, overuse of this may hide mistakes in code we cannot easily spot otherwise (i.e. `rememberedTreeIds.size` should always exist, if it doesn't, it's not a `Set` - which wouldn't be valid here. )

src/opnsense/www/js/opnsense_bootgrid.js
```diff
@@ -843,6 +855,21 @@ class UIBootgrid {
...
+            const queue = this.table.getRows();
+            while (queue.length) {
+                const row = queue.shift();
+                const data = row?.getData?.();
+                const id = data?.[this.dataIdentifier] ?? data?.[this.options.datakey];
+                if (id && this.rememberedTreeIds.has(id)) {
+                    row.treeExpand?.(); // no-op if already expanded
+                }
+                const kids = row.getTreeChildren?.();
```
swhite2: Here for example, we unconditionally try to access `row.getTreeChildren`, but `row` may be `undefined` according to `row?.getData?.()`.

If we expect that a row may be `undefined`, we may want to deal with this explicitly by bailing this loop iteration. Otherwise it's not necessary to use the chaining operator on a row and just let it error out sooner rather than later.

Also, `getTreeChildren` is a documented feature of Tabulator - if it isn't there, we want to know about it instead of silencing it by defaulting to `undefined`. (same goes for `row.getData` for example)

src/opnsense/www/js/opnsense_bootgrid.js
```diff
@@ -1184,6 +1211,41 @@ class UIBootgrid {
...
+            return rowMap;
+        };
+
+        const beforeSnapshot = snapshot();
+
+        const expandOnceAfterReload = () => {
+            this.table.off('dataProcessed', expandOnceAfterReload);
+            const afterSnapshot = snapshot();
+            afterSnapshot.forEach((childCount, rowId) => {
+                if (childCount > (beforeSnapshot.get(rowId) || 0)) {
```
swhite2: Want to quickly point out that while this code is valid, these types of constructs are often better dealt with through null coalescing (`??`). If the fallback was supposed to be `|| 42`, this would always overwrite an intended`0` inserted into the Map.

No problem here but just general feedback :)

swhite2 (approved):

PR #9109, opened 2025-08-18
reporting/health: Selective RRD reports collection

EddieHyun:
Current reporting UI provides every RRD graphs collected from predefined sources behind the scene, without exception. I'm suggesting an improvement to make them selective.
Here's what I've done:
- One-click UI for enabling/disabling report collection for each source
- Disabled reports no longer appear in the Reporting>Health page
- RRD selection config is persisted as a comma-separated string at rrd->exclude element
- Slightly reduces disk writes

healthss
Items with green play button are enabled. :wink:

### DIFF
--- a/src/opnsense/mvc/app/controllers/OPNsense/Diagnostics/Api/SystemhealthController.php
+++ b/src/opnsense/mvc/app/controllers/OPNsense/Diagnostics/Api/SystemhealthController.php
@@ -1,6 +1,7 @@
 <?php
 
 /*
+ * Copyright (C) 2025 Eddie Hyun <edwards.hyun@protonmail.ch>
  * Copyright (C) 2023 Deciso B.V.
  * Copyright (C) 2015 Jos Schellevis <jos@opnsense.org>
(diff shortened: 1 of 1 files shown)


EddieHyun:
All the changes are tested and worked fine on my latest opnsense environment.
Any suggestion would be appreciated. Thanks.

AdSchellevis:
I'm not necessarily against configurability of the RRD graphs, but we shouldn't try to mold this into the legacy cleanup page (reporting_settings.php) as it will stash technical dept someone has to fix later.

As the files are not the actual configuration items, the configuration items should cover `Types` and `Stats` and be persisted in the configuration similar to other settings of the system.

There's also the question if skipping the parts of the collection saves a lot of processing time, in which case micromanaging the stats might be more of a visibility issue than a processing one.

EddieHyun:
Totally agreeable, as well as other opinions on this issue. Also solved my curiosity. Actually, I was wondering why that page alone had that ancient code. Thanks.
I'm closing this PR for the moment.

PR #9114, opened 2025-08-19
firewall: live log: refactor to ringbuffer approach

swhite2:


src/opnsense/mvc/app/views/OPNsense/Diagnostics/fw_log.volt
```diff
@@ -12,7 +12,7 @@
  #    this list of conditions and the following disclaimer in the documentation
  #    and/or other materials provided with the distribution.
  #
- # THIS SOFTWARE IS PROVIDED ``AS IS'' AND ANY EXPRESS OR IMPLIED WARRANTIES,
```
fichtner: doesn't make sense to adjust this ;) eventually I think we should move to https://spdx.dev/learn/handling-license-info/#how at some point which helps reduce the bulk of the header

swhite2: to be honest I didn't notice this, as the header was copied from somewhere else before replacing the original `fw_log.volt`

fichtner (approved):
can't review the bulk of the js code but seeing it in action in master branch would be great, thanks!

Issue #9094, opened 2025-08-14
Add kea DHCP server lease affinity options to OPNsense 25.7 GUI options to configure kea 2.6.3 DHCP lease affinity


hharry06:
recently deployed a new OPNsense 25.7.1_1-amd64 network using kea 2.6.3 as the DHCP server, and customer wishes to configure DHCP affinity options, which kea 2.6 supports, but not yet exposed in OPNsense GUI

reference for kea DHCP lease affinity options here -> https://kea.readthedocs.io/en/kea-2.6.0/arm/lease-expiration.html

Will these already available kea configuration options, be available in OPNsense GUI front end, in upcoming releases ? If yes when ?

Customer is currently using Cisco IOS DHCP server, which already does DHCP lease affinity, and wishes to move to OPNsense Kea DHCP...with same DHCP lease affinity functionality.

sopex:
I agree, it should be implemented.

sopex:
I would appreciate feedback on this:

`opnsense-patch -a sopex -c core c6d7453d5809d948df6f29eba32b6aee39b02227`

fichtner:
@sopex can you link your commits to this issue url or post the link? That would help with audits and review.

sopex:
> [@sopex](https://github.com/sopex) can you link your commits to this issue url or post the link? That would help with audits and review.

Sure, everything is public either way. I just haven't made a PR yet because I don't want to add to your backlog without being 100% sure. The conf file is created correctly, but I don't run KEA to test. I believe affinity is important enough to be added to the GUI.

https://github.com/sopex/core/commit/c6d7453d5809d948df6f29eba32b6aee39b02227

Also:
https://forum.opnsense.org/index.php?topic=48516.0

fichtner:
That's fine and fair. I just don't want to train users to get in the habit of applying patches from external accounts without looking at the contents of the patch. I try to post patch URLs where I can, opnsense-patch even supports reading URLs although github likes to mange URLs and external accounts still need -a option for safety reasons.

I think the best thing is to drop an URL into the test commit and it should show here as a clickable link automatically.

sopex:
> That's fine and fair. I just don't want to train users to get in the habit of applying patches from external accounts without looking at the contents of the patch. I try to post patch URLs where I can, opnsense-patch even supports reading URLs although github likes to mange URLs and external accounts still need -a option for safety reasons.
> 
[...]

PR #9120, opened 2025-08-21
Kea: enable affinity through gui

sopex:
Closes: https://github.com/opnsense/core/issues/9094

Affinity is a useful option that enables clients to reclaim their lease. Εven though KEA has a multitude of options, and manual setup is recommended for most of them, this is useful to enough people both in the business and home verticals to warrant a GUI entry.

https://kea.readthedocs.io/en/stable/arm/lease-expiration.html

@fichtner Now you are officially 4 PRs behind :)

### DIFF
--- a/src/opnsense/mvc/app/controllers/OPNsense/Kea/forms/generalSettings4.xml
+++ b/src/opnsense/mvc/app/controllers/OPNsense/Kea/forms/generalSettings4.xml
@@ -33,6 +33,18 @@
         <help>Defines how long the addresses (leases) given out by the server are valid (in seconds)</help>
     </field>
     <field>
+        <id>dhcpv4.general.kaffinity</id>
+        <label>Lease Affinity</label>
+        <type>checkbox</type>
+        <help>Provides the means to assign the same lease to a returning client after its lease has expired</help>
+    </field>
+    <field>
+        <id>dhcpv4.general.hold_reclaimed_time</id>
+        <label>Affinity lifetime</label>
+        <type>text</type>
+        <help>Defines for how long a returning client will be able to keep the same lease (in seconds)</help>
+    </field>
+    <field>
         <id>dhcpv4.general.fwrules</id>
         <label>Firewall rules</label>
         <type>checkbox</type>
--- a/src/opnsense/mvc/app/models/OPNsense/Kea/KeaDhcpv4.php
+++ b/src/opnsense/mvc/app/models/OPNsense/Kea/KeaDhcpv4.php
@@ -207,6 +207,19 @@ class KeaDhcpv4 extends BaseModel
         return $result;
     }
 
+    private function getExpiredLeasesProcessingConfig()
+    {
+            if (empty((string)$this->general->kaffinity)) {
+            return null;
+        }
+
+        return [
+            'reclaim-timer-wait-time' => 10,
+            'hold-reclaimed-time' => (int)$this->general->hold_reclaimed_time->__toString(),
+            'flush-reclaimed-timer-wait-time' => 25
+        ];
+    }
+
     public function generateConfig($target = '/usr/local/etc/kea/kea-dhcp4.conf')
     {
         $cnf = [
@@ -238,6 +251,10 @@ class KeaDhcpv4 extends BaseModel
                 'subnet4' => $this->getConfigSubnets(),
             ]
         ];
+        $expiredLeasesConfig = $this->getExpiredLeasesProcessingConfig();
+            if ($expiredLeasesConfig !== null) {
+                $cnf['Dhcp4']['expired-leases-processing'] = $expiredLeasesConfig;
+            }
         if (!(new KeaCtrlAgent())->general->enabled->isEmpty()) {
             $cnf['Dhcp4']['hooks-libraries'] = [];
             $cnf['Dhcp4']['hooks-libraries'][] = [
--- a/src/opnsense/mvc/app/models/OPNsense/Kea/KeaDhcpv4.xml
+++ b/src/opnsense/mvc/app/models/OPNsense/Kea/KeaDhcpv4.xml
@@ -1,6 +1,6 @@
 <model>
     <mount>//OPNsense/Kea/dhcp4</mount>
-    <version>1.0.4</version>
+    <version>1.0.5</version>
     <description>Kea DHCPv4 configuration</description>
     <items>
         <general>
@@ -28,6 +28,15 @@
                     <raw>raw</raw>
                 </OptionValues>
             </dhcp_socket_type>
+            <kaffinity type="BooleanField">
+                <Default>0</Default>
+                <Required>Y</Required>
+            </kaffinity>
+            <hold_reclaimed_time type="IntegerField">
+                <default>28800</default>
+                <Required>Y</Required>
+                <MinimumValue>0</MinimumValue>
+            </hold_reclaimed_time>
         </general>
         <ha>
             <enabled type="BooleanField">


src/opnsense/mvc/app/models/OPNsense/Kea/KeaDhcpv4.php
```diff
@@ -207,6 +207,19 @@ private function getConfigSubnets()
...
     }
 
+    private function getExpiredLeasesProcessingConfig()
+    {
+            if (empty((string)$this->general->kaffinity)) {
+            return null;
+        }
+
+        return [
+            'reclaim-timer-wait-time' => 10,
```
Monviech: I would make all of these properties configurable, that are specified in the manual.

You never know when the next user comes who wants a slightly different value here for /some/ reason.

sopex: It's up to you of course, but then the settings page will balloon very quickly.

Even under "expired-leases-processing" there are 8 (plus these 2) more settings that someone might need.

I personally believe that "manual config" is more of a feature than a bug.

Monviech: I only mean those that you are implementing right now.

src/opnsense/mvc/app/controllers/OPNsense/Kea/forms/generalSettings4.xml
```diff
@@ -32,6 +32,18 @@
         <type>text</type>
         <help>Defines how long the addresses (leases) given out by the server are valid (in seconds)</help>
     </field>
+    <field>
+        <id>dhcpv4.general.kaffinity</id>
```
Monviech: You could show the lease affinity specific options (e.g. hold_reclaimed_time) only when kaffinity is true. This would unclog the GUI if more options are introduced as well.

https://github.com/opnsense/core/blob/fb374975f46de015abc1aa84c9e11249b750f041/src/opnsense/mvc/app/views/OPNsense/Dnsmasq/settings.volt#L177
https://github.com/opnsense/core/blob/fb374975f46de015abc1aa84c9e11249b750f041/src/opnsense/mvc/app/controllers/OPNsense/Dnsmasq/forms/dialogDHCPrange.xml#L86

Another code wise simpler solution could be a header thats collapsed per default, and just drop all lease affinity specific properties into it (also the checkbox) e.g.:

```
    <field>
        <type>header</type>
        <label>Lease Affinity</label>
        <collapse>true</collapse>
    </field>
```

src/opnsense/mvc/app/models/OPNsense/Kea/KeaDhcpv4.php
AdSchellevis: just out of curiosity, aren't these just the defaults when not set (https://kea.readthedocs.io/en/latest/arm/lease-expiration.html#lease-reclamation-configuration-parameters)?

src/opnsense/mvc/app/models/OPNsense/Kea/KeaDhcpv4.xml
```diff
@@ -28,6 +28,15 @@
...
             </dhcp_socket_type>
+            <kaffinity type="BooleanField">
+                <Default>0</Default>
+                <Required>Y</Required>
+            </kaffinity>
+            <hold_reclaimed_time type="IntegerField">
+                <default>28800</default>
+                <Required>Y</Required>
+                <MinimumValue>0</MinimumValue>
+            </hold_reclaimed_time>
```
fichtner: The lightweight approach would be just to offer hold_reclaimed_time, make it optional to retain the current default and let the default values be set implicitly. If defaults change on Keas end for good reasons we will not be able to change them, especially hard for values that are not presented on the GUI then.

sopex:
After reading all your comments and returning from the sea :)

I believe there are two optimal ways forward.

1) I only keep hold_reclaimed_time exposed in the GUI and let the other two options be in the default state and not hardcoded to default.

2) Expand on what @Monviech suggested and implement the whole lease expiration collection under a header. In that case, affinity will be front and center but all other options will also be exposed in a combination of normal and advanced settings.

This can also work as a template to eventually implement most of the KEA options under different headers. (budget for an extra salary?  ;) 

Option 1 can also be a intermediate addition until 26.1 that could have the whole of KEA exposed? 

Thank you for your time, let me know your thoughts and what you would prefer.

src/opnsense/mvc/app/models/OPNsense/Kea/KeaDhcpv4.xml
sopex: That would eliminate the enable toggle which is probably for the best. I think the options I provided bellow would work great.

src/opnsense/mvc/app/models/OPNsense/Kea/KeaDhcpv4.php
sopex: > just out of curiosity, aren't these just the defaults when not set (https://kea.readthedocs.io/en/latest/arm/lease-expiration.html#lease-reclamation-configuration-parameters)?

Yes, they are. I hardcoded them since they won't change, to help the user change them when editing the file for manual configuration. Anyway, the options I provided below would work best IMO.

sopex:
Sorry for pestering you @fichtner

sopex:
I know you had a difficult day or two with the hotfixes. We really appreciate your dedication.
When you get a chance, take a look here. I ended up going with option 2, and in my testing, it works flawlessly. 

Best,
sopex

src/opnsense/mvc/app/models/OPNsense/Kea/KeaDhcpv4.php
```diff
@@ -207,6 +207,25 @@ private function getConfigSubnets()
...
+    $fields = [
+        'reclaim-timer-wait-time' => $this->lexpire->reclaim_timer_wait_time->__toString(),
+        'hold-reclaimed-time' => $this->lexpire->hold_reclaimed_time->__toString(),
+        'flush-reclaimed-timer-wait-time' => $this->lexpire->flush_reclaimed_timer_wait_time->__toString(),
+        'max-reclaim-leases' => $this->lexpire->max_reclaim_leases->__toString(),
+        'max-reclaim-time' => $this->lexpire->max_reclaim_time->__toString(),
+        'unwarned-reclaim-cycles' => $this->lexpire->unwarned_reclaim_cycles->__toString(),
+    ];
+    $config = [];
+    foreach ($fields as $key => $value) {
```
AdSchellevis: we should be able to simplify this by iterating over the fields, maybe even use the original names in the model (e.g. `reclaim-timer-wait-time`), in which case an `iterator_to_array($this->lexpire>iterateItems())` likely offers the output you're looking for

src/opnsense/mvc/app/models/OPNsense/Kea/KeaDhcpv4.xml
```diff
@@ -29,6 +29,38 @@
                 </OptionValues>
             </dhcp_socket_type>
         </general>
+        <lexpire>
+            <hold_reclaimed_time type="IntegerField">
+                <Default></Default>
```
AdSchellevis: empty `Defaults` have no meaning, you can omit those.

src/opnsense/mvc/app/controllers/OPNsense/Kea/forms/generalSettings6.xml
```diff
@@ -38,9 +38,60 @@
...
+        <id>dhcpv6.lexpire.hold_reclaimed_time</id>
+        <label>Affinity lifetime</label>
+        <type>text</type>
+        <help>Defines for how long a returning client will be able to keep the same lease (in seconds)</help>
+    </field>
+    <field>
+        <id>dhcpv6.lexpire.reclaim_timer_wait_time</id>
+        <label>Reclamation delay</label>
+        <type>text</type>
+        <hint>10</hint>
```
AdSchellevis, line 56: if we move the hints to the defaults, it would simplify the code, with the small price of always persisting the current defaults (which are likely quite static)

sopex: If there is a need to simplify the for loop, sure I can do it with required and default.

I just did it like this because a few weeks back it was mentioned that you prefer not to hardcode these.

And to also keep the config files cleaner.

AdSchellevis: lesser code equals lesser maintenance ;) I'm also ok with omitting the defaults, there are arguments for both strategies.

src/opnsense/mvc/app/models/OPNsense/Kea/KeaDhcpv4.php
sopex: I decided against original names to maintain the consitency with all the other values in the model/ controllers and used the `str_replace` that you have used for all the other field names.

AdSchellevis: @sopex that's ok, I see your point there.

src/opnsense/mvc/app/models/OPNsense/Kea/KeaDhcpv4.php
```diff
@@ -207,6 +207,20 @@ private function getConfigSubnets()
...
 
+    private function getExpiredLeasesProcessingConfig()
+    {
+        $config = [];
+        $lexpireFields = iterator_to_array($this->lexpire->iterateItems());
+        foreach ($lexpireFields as $fieldName => $fieldValue) {
+            $value = $fieldValue->__toString();
+            if ($value !== '' && $value !== null) {
+                $keaFieldName = str_replace('_', '-', $fieldName);
+                $config[$keaFieldName] = (int)$value;
```
fichtner: it's better to use the builtins here:

```suggestion
            if (!$fieldValue->isEqual('')) {
                $keaFieldName = str_replace('_', '-', $fieldName);
                $config[$keaFieldName] = (int)$fieldValue->getValue();
```

src/opnsense/mvc/app/models/OPNsense/Kea/KeaDhcpv4.xml
```diff
@@ -29,6 +29,32 @@
                 </OptionValues>
             </dhcp_socket_type>
         </general>
+        <lexpire>
+            <hold_reclaimed_time type="IntegerField">
+                <Required>N</Required>
```
fichtner: You don't need to set Required=N

src/opnsense/mvc/app/models/OPNsense/Kea/KeaDhcpv6.php
```diff
@@ -194,6 +194,20 @@ private function getConfigSubnets()
         return $result;
     }
 
+    private function getExpiredLeasesProcessingConfig()
+    {
+        $config = [];
+        $lexpireFields = iterator_to_array($this->lexpire->iterateItems());
+        foreach ($lexpireFields as $fieldName => $fieldValue) {
+            $value = $fieldValue->__toString();
```
fichtner, line 202: same here, see previous suggestion

src/opnsense/mvc/app/models/OPNsense/Kea/KeaDhcpv6.xml
```diff
@@ -21,6 +21,32 @@
                 <Default>1</Default>
             </fwrules>
         </general>
+        <lexpire>
+            <hold_reclaimed_time type="IntegerField">
+                <Required>N</Required>
```
fichtner: Same

src/opnsense/mvc/app/models/OPNsense/Kea/KeaDhcpv4.xml
```diff
@@ -1,6 +1,6 @@
 <model>
     <mount>//OPNsense/Kea/dhcp4</mount>
-    <version>1.0.4</version>
+    <version>1.1.0</version>
```
fichtner: The model version doesn't need a change because no Required=Y field was added

src/opnsense/mvc/app/models/OPNsense/Kea/KeaDhcpv6.xml
```diff
@@ -1,6 +1,6 @@
 <model>
     <mount>//OPNsense/Kea/dhcp6</mount>
-    <version>1.0.0</version>
+    <version>1.1.0</version>
```
fichtner: same here, no change needed

src/opnsense/mvc/app/models/OPNsense/Kea/KeaDhcpv4.xml
sopex: Sure, I just see a lot of empty defaults and required=N in existing code.

fichtner: In plugins maybe, try "make lint-model" in core root on your modifications ;)

src/opnsense/mvc/app/models/OPNsense/Kea/KeaDhcpv6.xml
```diff
@@ -1,6 +1,6 @@
 <model>
     <mount>//OPNsense/Kea/dhcp6</mount>
-    <version>1.0.0</version>
+    <version>1.0.4</version>
```
fichtner:
```suggestion
    <version>1.0.0</version>
```

src/opnsense/mvc/app/controllers/OPNsense/Kea/forms/generalSettings6.xml
```diff
@@ -38,9 +38,60 @@
...
+        <help>The interval between the completion of the previous reclamation cycle and the start of the next one. (in seconds)</help>
```
fichtner: this is looking good, but one last request: the help texts should end in a dot to force proper sentences. I'm ok with parenthesis some times, but the principle I learned was that either it's important enough to remove the parenthesis around the remark or the note can be removed completely because it's not overly important. This would help to ensure proper easy to read sentence structure.

fichtner: Example suggestion:

```suggestion
        <help>The interval in seconds between the completion of the previous reclamation cycle and the start of the next one.</help>
```

src/opnsense/mvc/app/models/OPNsense/Kea/KeaDhcpv6.xml
sopex: I apologize profusely

src/opnsense/mvc/app/controllers/OPNsense/Kea/forms/generalSettings6.xml
sopex: Shall I keep it in parentheses for 0? Parentheses are not proper, but let the user quickly glance over.

src/opnsense/mvc/app/models/OPNsense/Kea/KeaDhcpv6.xml
fichtner: no worries at all :)

src/opnsense/mvc/app/controllers/OPNsense/Kea/forms/generalSettings6.xml
fichtner: I would prefer something like: Use "0" for no limit.

src/opnsense/mvc/app/controllers/OPNsense/Kea/forms/generalSettings4.xml
```diff
@@ -44,9 +44,60 @@
...
+        <advanced>true</advanced>
+        <help>This parameter controls in seconds how often the server initiates the lease reclamation procedure.</help>
+    </field>
+    <field>
+        <id>dhcpv4.lexpire.max_reclaim_time</id>
+        <label>Maximum Reclamation time</label>
+        <type>text</type>
+        <hint>10</hint>
+        <advanced>true</advanced>
+        <help>Defines in milliseconds an upper limit to the length of time a lease reclamation procedure can take. Using 0 disables the time limit</help>
```
fichtner:
```suggestion
        <help>Defines an upper limit in milliseconds to the length of time a lease reclamation procedure can take. Using "0" disables the time limit.</help>
```

fichtner: You can commit them here from GitHub, that's easier in this phase. There may be more, one sec.

src/opnsense/mvc/app/controllers/OPNsense/Kea/forms/generalSettings4.xml
```diff
@@ -44,9 +44,60 @@
...
+    </field>
+    <field>
+        <id>dhcpv4.lexpire.max_reclaim_leases</id>
+        <label>Maximum reclamation leases</label>
+        <type>text</type>
+        <hint>100</hint>
+        <advanced>true</advanced>
+        <help>Defines the maximum number of reclaimed leases that can be processed at one time. Use 0 to set it to unlimited.</help>
```
fichtner:
```suggestion
        <help>Defines the maximum number of reclaimed leases that can be processed at one time. Use "0" to set it to unlimited.</help>
```

src/opnsense/mvc/app/controllers/OPNsense/Kea/forms/generalSettings6.xml
```diff
@@ -30,17 +30,68 @@
         <id>dhcpv6.general.valid_lifetime</id>
         <label>Valid lifetime</label>
         <type>text</type>
-        <help>Defines how long the addresses (leases) given out by the server are valid (in seconds)</help>
+        <help>Defines in seconds how long the addresses (leases) given out by the server are valid.</help>
```
fichtner:
```suggestion
        <help>The valid life time of the leases given out by the server in seconds.</help>
```

You can try to incorporate the option names into the help text. Most of the time they are very clearly crafted to reflect their purpose and you don't have to explain the world from scratch again. :)

src/opnsense/mvc/app/controllers/OPNsense/Kea/forms/generalSettings6.xml
```diff
@@ -30,17 +30,68 @@
...
+        <hint>10</hint>
+        <help>The interval in seconds between the completion of the previous reclamation cycle and the start of the next one.</help>
+    </field>
+    <field>
+        <id>dhcpv6.lexpire.flush_reclaimed_timer_wait_time</id>
+        <label>Reclamation initiation</label>
+        <type>text</type>
+        <hint>25</hint>
+        <advanced>true</advanced>
+        <help>This parameter controls in seconds how often the server initiates the lease reclamation procedure.</help>
```
fichtner:
```suggestion
        <help>This parameter controls the server wait time in seconds between each lease reclamation procedure.</help>
```

fichtner: edit: I reworded this a bit more

src/opnsense/mvc/app/controllers/OPNsense/Kea/forms/generalSettings6.xml
```diff
@@ -30,17 +30,68 @@
...
+    </field>
+    <field>
+        <id>dhcpv6.lexpire.max_reclaim_time</id>
+        <label>Maximum Reclamation time</label>
+        <type>text</type>
+        <hint>10</hint>
+        <advanced>true</advanced>
+        <help>Defines in milliseconds an upper limit to the length of time a lease reclamation procedure can take. Using 0 disables the time limit.</help>
```
fichtner:
```suggestion
        <help>Defines an upper limit to time in milliseconds that a lease reclamation procedure may take. Using "0" disables the time limit.</help>
```

src/opnsense/mvc/app/controllers/OPNsense/Kea/forms/generalSettings6.xml
```diff
@@ -30,17 +30,68 @@
...
+    </field>
+    <field>
+        <id>dhcpv6.lexpire.max_reclaim_leases</id>
+        <label>Maximum reclamation leases</label>
+        <type>text</type>
+        <hint>100</hint>
+        <advanced>true</advanced>
+        <help>Defines the maximum number of reclaimed leases that can be processed at one time. Use 0 to set it to unlimited</help>
```
fichtner:
```suggestion
        <help>Defines the maximum number of reclaimed leases that can be processed at one time. Use "0" to set it to unlimited.</help>
```

src/opnsense/mvc/app/controllers/OPNsense/Kea/forms/generalSettings6.xml
sopex: This is not one of mine. It's in general not lexpire (I just edited the help text for uniformity with the parentheses). 

But judging from the language they copied the KEA documentation descriptions like I did. I thought it would be better, but I can rewrite them, whatever you prefer.

Lease expiry names are not that self explanatory unlike the example you gave.

fichtner: As OPNsense we have an obligation to the user to provide help texts. The upstream help texts may be great but sometimes are not or too lengthy. The help text here aims to explain the effect of the option in 1-2 sentences. The docs.git is for complex setups and tutorials. For everything else people should refer to the upstream documentation, but that also only works if users can infer from our GUI options the upstream equivalent in the config file (and for "Valid lifetime" here that is already the case).

TL;DR: I'd prefer our documentation to be consistent with itself, because users come for us asking for help more often than they will go to the upstream and that's perfectly fine. In that case a lot of ambiguity can already be spared.

src/opnsense/mvc/app/controllers/OPNsense/Kea/forms/generalSettings4.xml
```diff
@@ -44,9 +44,60 @@
...
+        <advanced>true</advanced>
+        <help>This parameter controls the server wait time in seconds between each lease reclamation procedure.</help>
+    </field>
+    <field>
+        <id>dhcpv4.lexpire.max_reclaim_time</id>
+        <label>Maximum Reclamation time</label>
+        <type>text</type>
+        <hint>10</hint>
+        <advanced>true</advanced>
+        <help>Defines an upper limit in milliseconds to the length of time a lease reclamation procedure may take. Using "0" disables the time limit.</help>
```
fichtner:
```suggestion
        <help>Defines an upper limit in milliseconds to the length of time a lease reclamation procedure may take. Use "0" to disable the time limit.</help>
```

I'll do this from GitHub and merge

src/opnsense/mvc/app/controllers/OPNsense/Kea/forms/generalSettings6.xml
```diff
@@ -30,17 +30,68 @@
...
+        <advanced>true</advanced>
+        <help>This parameter controls the server wait time in seconds between each lease reclamation procedure.</help>
+    </field>
+    <field>
+        <id>dhcpv6.lexpire.max_reclaim_time</id>
+        <label>Maximum Reclamation time</label>
+        <type>text</type>
+        <hint>10</hint>
+        <advanced>true</advanced>
+        <help>Defines an upper limit in milliseconds to the length of time a lease reclamation procedure may take. Using "0" disables the time limit.</help>
```
fichtner:
```suggestion
        <help>Defines an upper limit in milliseconds to the length of time a lease reclamation procedure may take. Use "0" to disable the time limit.</help>
```

fichtner:
Thanks a lot, merged!

sopex:
> Thanks a lot, merged!

I thank you a lot! :)

PR #9131, opened 2025-08-25
Update Disk.js - add used/total GB info

MagicFab:
Add used / total GB disk space used info to Disk widget

### DIFF
--- a/src/opnsense/www/js/widgets/Disk.js
+++ b/src/opnsense/www/js/widgets/Disk.js
@@ -57,7 +57,19 @@ export default class Disk extends BaseGaugeWidget {
 
         return size * units[unit];
     }
+    
+    /**
+     * Converts bytes to gigabytes using SI units (base-10).
+     * This returns a numeric value rounded to one decimal place without the "GB" label,
+     * useful for displaying raw numbers in a compact format.
+     * Note: Uses 1000-based conversion to match user-facing expectations
+     * (e.g. disk labels and UI conventions), not binary (1024-based) units.
+     */
 
+    _formatGB(bytes) {
+        return (bytes / (1000 * 1000 * 1000)).toFixed(1);
+    }
+    
     getMarkup() {
         return $(`
             <div class="${this.id}-chart-container">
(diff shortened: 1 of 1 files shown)


MagicFab:
image

sopex:
I like it :)

src/opnsense/www/js/widgets/Disk.js
```diff
@@ -83,6 +95,19 @@ export default class Disk extends BaseGaugeWidget {
...
+            // Displays used and total disk space in GB below the percentage,
+            // similar to the Memory widget layout. Uses base-10 formatting for clarity,
+
+            secondaryText: (data, chart) => {
+                const usedBytes = chart.config.data.datasets[0].data[0];
+                const freeBytes = chart.config.data.datasets[0].data[1];
+                const totalBytes = usedBytes + freeBytes;
+                const usedGB = this._formatGB(usedBytes);
+                const totalGB = this._formatGB(totalBytes);
+                return `(${usedGB}/${totalGB}) GB`;
```
fichtner, line 108: The () looks misplaced before the GB, may be better to just omit and free the space used by it.

src/opnsense/www/js/widgets/Disk.js
```diff
@@ -57,7 +57,19 @@ export default class Disk extends BaseGaugeWidget {
...
+    /**
+     * Converts bytes to gigabytes using SI units (base-10).
+     * This returns a numeric value rounded to one decimal place without the "GB" label,
+     * useful for displaying raw numbers in a compact format.
+     * Note: Uses 1000-based conversion to match user-facing expectations
+     * (e.g. disk labels and UI conventions), not binary (1024-based) units.
+     */
+
+    _formatGB(bytes) {
+        return (bytes / (1000 * 1000 * 1000)).toFixed(1);
```
fichtner, line 70: I think this is too simplistic as it should adapt to k M G T...

Monviech: This kind of formatter could be added to the BaseWidget I think.

E.g. here is one the UI uses in multiple places.

https://github.com/opnsense/core/blob/5875c18758853a6f1eaf2941bf83b443442bb623/src/opnsense/www/js/opnsense_ui.js#L37

src/opnsense/www/js/widgets/Disk.js
MagicFab: It looks the same as the memory widget, but I agree the parenthesis aren't useful. Then the memory widget would have to be adjusted (easily done).

src/opnsense/www/js/widgets/Disk.js
MagicFab: Yes, that would make more sense. My reasoning was that storage for such a device would be within XX GB and XXX GB - but we know how that ends up :)

swhite2:
Used in and superseded by https://github.com/opnsense/core/commit/7f3bcf5b787145b7cf185854758c3cddefdf1971

PR #9132, opened 2025-08-25
fix: null parameter passing in PHP 8.1 is depreciated

sopex:


src/opnsense/mvc/app/models/OPNsense/Firewall/FieldTypes/AliasField.php
```diff
@@ -84,7 +84,7 @@ protected function actionPostLoadingEvent()
         }
         if (self::$current_stats === null) {
             self::$current_stats = [];
-            $stats = json_decode((new Backend())->configdRun('filter diag table_size'), true);
+            $stats = json_decode((string)(new Backend())->configdRun('filter diag table_size'), true);
```
fichtner, line 87: Normally we use `?? ''` these days but it would be nicer if configdRun would just return an empty string in the error case

There is no strict return type for it but the documentation says `@return string` but then uses null in some cases

CC @AdSchellevis

AdSchellevis: @fichtner as far as I know, we don't validate strictly on `null` anywhere, let me change the `Backend` class accordingly

AdSchellevis:
should be fixed in https://github.com/opnsense/core/commit/32d24c13583c29cd61311fbd93408606cf1f32d5

fichtner:
@AdSchellevis @sopex thanks I cleaned up the ones I could find a7c0c93c396e

AdSchellevis:
@fichtner nice! thanks!

PR #9156, opened 2025-08-27
Monit toptalker.php script

ednt:
During a training program a few participants asked for the possibility to get toptalkers via e-mail.

So I searched for an acceptable and easy way to do this.
I'm using Monit, because of the easy SMTP configuration and the possibility of using cron like execution timeplan.

NetFlow with local capture needs to be active.
I think this is a very usefull script and it should be public available.
It would be also nice if someone creates a migration file to create a disabled service entry in Monit, maybe with '1 cycles' as
default poll time. The interface is not needed at the first time, because it is written in the alert that it is needed.
Also running of NetFlow is checked.

Of course you can test it 'by hand': ./toptalker.php vtnet0

Best regards

### DIFF
--- a/src/opnsense/scripts/monit/toptalker.php
+++ b/src/opnsense/scripts/monit/toptalker.php
@@ -1,0 +1,81 @@
+#!/usr/local/bin/php
+<?php
+
+/*
+ * Copyright (c) 2025 EDNT GmbH
+ * All rights reserved.
+ *
+ * Redistribution and use in source and binary forms, with or without
+ * modification, are permitted provided that the following conditions are met:
+ *
+ * 1. Redistributions of source code must retain the above copyright notice,
+ *    this list of conditions and the following disclaimer.
+ *
+ * 2. Redistributions in binary form must reproduce the above copyright
+ *    notice, this list of conditions and the following disclaimer in the
+ *    documentation and/or other materials provided with the distribution.
+ *
+ * THIS SOFTWARE IS PROVIDED ``AS IS'' AND ANY EXPRESS OR IMPLIED WARRANTIES,
+ * INCLUDING, BUT NOT LIMITED TO, THE IMPLIED WARRANTIES OF MERCHANTABILITY
+ * AND FITNESS FOR A PARTICULAR PURPOSE ARE DISCLAIMED. IN NO EVENT SHALL THE
+ * AUTHOR BE LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY,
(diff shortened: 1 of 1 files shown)


AdSchellevis:
It’s probably better to add this as an example to the docs combined with documentation about how people should use your script.

our core repository isn’t the right place for examples

ednt:
Aren't the other scripts in this directory not also 'examples' ?
They are not enabled by default.
And 70% will never use the carp_status.php, because they don't use CARP.
But traffic is always on the interfaces.

And all of the current 3 examples on https://docs.opnsense.org/manual/monit.html does not use any custom scripts.

So if you want to add a 4th example you need to place the sourcecode in the documentation (the wrong place in my opinion).
Also many users are not familar with the CLI and how to place own scripts and make them executable.
In our trainings it becomes clear that many of the students wants to avoid the contact with the CLI

I think it would be a good compromise, to add the file in the repository and write in an Example 4, how to use/enable it in the GUI.

For other 3rd party stuff you have the plugin solution, but for such scripts this is not the way to go.
Maybe an other soultion is needed to add 3rd party scripts.

AdSchellevis:
To some degree these are examples, but with little to no maintenance needs as they literally only return a yes/no situation without further content. We imported Monit in core a long time ago, and in all fairness, we shouldn't have done that as the script execution hooks don't fit our standards.

If you're using these scripts in your training program, it might be better to host this part of the material on your end with a small instruction for people how to use them. In which case it's also not an issue to add marketing text in the email as is the case in this script (2BSD clause copyright claims are always fine, adding company names and websites and output obviously isn't).

In time someone need to document and maintain these scripts as they are technical dept waiting to happen, which we unfortunately can not be responsible for.

ednt:
I don't want to dispute with you, but ... 

OPNsense (c) 2014-2025 Deciso B.V.

with link is directly on the main page of the GUI. ;)
according to your writing this is not allowed, because Deciso is a selling company.
Also some plugins points to companies where you can buy licenses for extended features.

I thought, since users requested such a feature, I make it public for all OPNsense users.
As a goody which is already included and usable, no need to pay for it.
But it is no problem for us to provide this only for users who join our trainings.

So ...
do you want to add/use the script or not?
If not I simply delete the pull request and all is fine.

AdSchellevis:
> I don't want to dispute with you, but ...
> OPNsense (c) 2014-2025 Deciso B.V.

I don't think I need to explain ownership of OPNsense and it's trademark. 

> Also some plugins points to companies where you can buy licenses for extended features.

for plugins that's allowed to some extend, usually there's a clear free version of the functionality in question, in some cases the plugin merely points to the place where to register for an account and isn't even maintained by said company.

> So ...
> do you want to add/use the script or not?
> If not I simply delete the pull request and all is fine.

We're not planning to use it, so feel free to close the PR.

PR #9162, opened 2025-08-29
bootgrid: Raise rowCount default to 50 as y-overflow handles this gracefully

Monviech:
ensure low initial rowcount for master detail grids, clean up some scattered rowcounts in views

### DIFF
--- a/src/opnsense/www/js/opnsense_bootgrid.js
+++ b/src/opnsense/www/js/opnsense_bootgrid.js
@@ -129,7 +129,7 @@ class UIBootgrid {
             disableScroll: false,
             sorting: true,
             selection: true,
-            rowCount: [7, 14, 20, 50, 100, true],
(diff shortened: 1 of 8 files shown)


src/opnsense/www/js/opnsense_bootgrid.js
```diff
@@ -129,7 +129,7 @@ class UIBootgrid {
             disableScroll: false,
             sorting: true,
             selection: true,
-            rowCount: [7, 14, 20, 50, 100, true],
+            rowCount: [50, 100, 200, 500, 1000, true], // tabulator y-overflows, no need for low defaults
```
fichtner: what's the difference between -1 and true?

fichtner: FWIW, I don't think we should add the comment

fichtner (approved):

PR #9171, opened 2025-09-01
firewall/automation: Improved moveRuleBeforeAction() sequence handling

Monviech:
Shift: when there is a distance ≤ 2 between the previous and target sequence, it increases the target and all following rules in the same prio_group by +100 to make space. The selected rule then takes the target’s old spot but also with a 100 space to ensure the shift does not happen right again.

Midpoint: when there is a distance > 2, it inserts the selected rule at the midpoint between previous and target, without changing the target or later rules.

-----

Also hide the sequence and sort_order in advanced mode, unclogs the mind.

### DIFF
--- a/src/opnsense/mvc/app/controllers/OPNsense/Firewall/Api/FilterController.php
+++ b/src/opnsense/mvc/app/controllers/OPNsense/Firewall/Api/FilterController.php
@@ -321,10 +321,8 @@ class FilterController extends FilterBaseController
             $uuid = $record->getAttribute('uuid');
             if ($prev_record != null && (string)$prev_record->prio_group == (string)$record->prio_group) {
                 $prev_sequence = $prev_record->sequence->asFloat();
-                // raw gap between this and the previous sequence
-                $raw_gap = (int)$record->sequence->asFloat() - (int)$prev_record->sequence->asFloat();
-                // distance will be averaged in push-forward logic, which is why the minimum should be at least 2 (half is 1)
-                $distance = max($raw_gap, 2);
+                /* distance will be averaged, which is why the minimum should be at least 2 (half is 1) */
+                $distance = max($record->sequence->asFloat() - $prev_record->sequence->asFloat(), 2);
             } elseif ($selected_node !== null && $target_node !== null) {
                 /* group processed */
                 break;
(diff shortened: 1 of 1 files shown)


src/opnsense/mvc/app/controllers/OPNsense/Firewall/Api/FilterController.php
```diff
@@ -334,12 +336,34 @@ public function moveRuleBeforeAction($selected_uuid, $target_uuid)
...
+                 *
+                 * Example:
+                 *   prev=100, target=101 → raw gap=1, but distance=2.
+                 *   Using $distance would suggest room for a midpoint (101),
+                 *   colliding with the target’s sequence.
+                 *
+                 * With raw_gap we correctly detect "no gap" and shift later rules
+                 * forward to create space.
+                 */
+                if ($raw_gap > 1) {
```
AdSchellevis: `$raw_gap` could be undefined

src/opnsense/mvc/app/controllers/OPNsense/Firewall/Api/FilterController.php
```diff
@@ -334,12 +336,34 @@ public function moveRuleBeforeAction($selected_uuid, $target_uuid)
...
+                    // normal case, safe gap
+                    $selected_id = (int)$prev_sequence + intdiv($raw_gap, 2);
+                } else {
+                    // no gap, shift target and everything after it forward
+                    $target_seq = (int)$record->sequence->asFloat();
+                    $shift = $target_seq + 1;
+                    foreach ($mdl->rules->rule->sortedBy(['prio_group', 'sequence']) as $rec2) {
+                        if ((string)$rec2->prio_group == (string)$record->prio_group &&
+                            (int)$rec2->sequence->asFloat() >= $target_seq) {
+                            $rec2->sequence = (string)$shift++;
```
AdSchellevis: doesn't this automatically cause all future moves to have no room between sequences?

src/opnsense/mvc/app/controllers/OPNsense/Firewall/Api/FilterController.php
```diff
@@ -321,8 +321,10 @@ public function moveRuleBeforeAction($selected_uuid, $target_uuid)
             $uuid = $record->getAttribute('uuid');
             if ($prev_record != null && (string)$prev_record->prio_group == (string)$record->prio_group) {
                 $prev_sequence = $prev_record->sequence->asFloat();
-                /* distance will be averaged, which is why the minimum should be at least 2 (half is 1) */
-                $distance = max($record->sequence->asFloat() - $prev_record->sequence->asFloat(), 2);
```
AdSchellevis: I would keep the `$distance` variable as it prevents the need of calculating things twice. when `$distance < 2` triggers different behavior, we can deal with that later (also prevents the `$raw_gap` undefined from happening later)

src/opnsense/mvc/app/controllers/OPNsense/Firewall/Api/FilterController.php
Monviech: removed it, everything is decided via `$distance`

src/opnsense/mvc/app/controllers/OPNsense/Firewall/Api/FilterController.php
Monviech: Fixed that by ensuring there is 100 space after move

src/opnsense/mvc/app/controllers/OPNsense/Firewall/Api/FilterController.php
Monviech: Only `$distance` variable is used now.

Monviech:
Fixed here: https://github.com/opnsense/core/commit/f3a5e200f4678d9647bfbe0075b0944f3101257f

PR #9174, opened 2025-09-02
mvc: account for option container and passthrough value

swhite2:
Fixes minor regression caused by https://github.com/opnsense/core/commit/5875c18758853a6f1eaf2941bf83b443442bb623


AdSchellevis:
@swhite2 other consumers of getValues() will have the same issue ;)

AdSchellevis:
totally misread from my end, this should be all indeed

Issue #8239, opened 2025-01-23
Input Field for 2FA Code During Login


smitty-nieto:
**Important notices**

Before you add a new report, we ask you kindly to acknowledge the following:

- [X] I have read the contributing guide lines at https://github.com/opnsense/core/blob/master/CONTRIBUTING.md
- [X] I am convinced that my issue is new after having checked both open and closed issues at https://github.com/opnsense/core/issues?q=is%3Aissue

**Is your feature request related to a problem? Please describe.**

The current implementation of 2FA in OPNsense requires users to manually prepend or append their 2FA token to their password before submitting the login form. This approach is unintuitive and can be confusing, especially for users unfamiliar with this method.

For the purpose of improving user experience and ensuring a smoother, more intuitive authentication process, I propose a change to how 2FA codes are entered during login process.

**Describe the solution you like**

The solution involves creating a dedicated input field for the 2FA code on the login page, separate from the password field. The system will handle the logic for concatenating the 2FA code with the password server-side during authentication, based on the authserver configuration.

This proposal supports both:

Prepending the MFA code to the password (default TOTP Authserver).
Appending the MFA code to the password (TOTP Reverse token order is checked).

The server-side implementation ensures:

No manual appending/prepending is required by the user.
Error messages for missing or invalid MFA codes are displayed inline on the login form for a better user experience.

**Describe alternatives you considered**

1. Retaining the current manual appending method: This method is error-prone and not user-friendly.
2. Adding client-side logic: While possible, it is less secure and relies on browser functionality.

**Additional context**

Add any other context or screenshots about the feature request here or links to relevant forum thread or similar

![Image](https://github.com/user-attachments/assets/665332c0-6845-428c-a339-0847260dd639)
![Image](https://github.com/user-attachments/assets/48759924-201e-46b0-9875-22b30d832752)

smitty-nieto:
The code change to accomplish this is trivial...This could certainly be improved, but this DOES work for me.

`130a131,139
> function isMFARequired($config) {
>     global $config;
>     $authMode = $config['system']['webgui']['authmode'] ?? 'Local Database';
>     $authServer = array_filter($config['system']['authserver'] ?? [], function($server) use ($authMode) {
>         return $server['name'] === $authMode && $server['type'] === 'totp';
>     });
>     return !empty($authServer);
> }
> 
148a158,159
>     $showMFAField = isMFARequired($config);
> 
169a181,202
> 
[...]

PR #9179, opened 2025-09-04
added optional OTP field to login form (issue #8239)

jullrich:
Adding a simple numeric OTP input field to the login form (Issue #8239)


danohn:
Is this essentially the same as #8241 ?

jullrich:
yes. looks like it is the same

PR #9184, opened 2025-09-08
KEA DHCPv6 static mappings DNS registration domain name

NOYB:
KEA DHCPv6 static mappings DNS registration domain name

DHCPv6 static mappings being DNS registered with the routers domain name is limiting.  Use the first search list domain for DNS registration.

Unbound requires reload.


Monviech:
@AdSchellevis I do not like the magic in this one. I think it would be better to close this and refer to using the Kea DDNS server implementation instead which has more flexibility regarding the forward zone.

PR #9186, opened 2025-09-09
KEA DHCPv4: honor client specific reservation domain name option

NOYB:
KEA DHCPv4: honor client specific reservation domain name option

Register in DNS with the client specific reservation domain name when specified.

Unbound requires reload.

### DIFF
--- a/src/opnsense/mvc/app/controllers/OPNsense/Kea/forms/dialogSubnet4.xml
+++ b/src/opnsense/mvc/app/controllers/OPNsense/Kea/forms/dialogSubnet4.xml
@@ -78,7 +78,7 @@
         <id>subnet4.option_data.domain_name</id>
         <label>Domain name</label>
         <type>text</type>
-        <help>The domain name to offer to the client, set to client's reservation domain name or this firewall's domain name when both left empty</help>
+        <help>A custom domain name to offer to the clients.</help>
         <grid_view>
             <visible>false</visible>
         </grid_view>


src/etc/inc/plugins.inc.d/kea.inc
```diff
@@ -115,6 +115,9 @@ function kea_staticmap($proto = null, $valid_addresses = true, $ifconfig_details
                     if (!empty((string)$subnet_node->option_data->domain_name)) {
                         $domain = (string)$subnet_node->option_data->domain_name;
                     }
+                    if (!empty((string)$reservation->option_data->domain_name)) {
```
AdSchellevis, line 115: no? will overwrite the subnets one specified above...

NOYB:
A client specific reservation domain name should override the subnet specified domain name should it not?

AdSchellevis:
then it's an `else if` the other way around...

NOYB:
Works as is for me.
If subnet domain name option is not empty set the domain name to that.
If client specific reservation domain name option is not empty set the domain to that (override subnet domain name option value).
Have you tested it and gotten a different result?
Sure, the flow logic can be done a number of ways.  But the result should be the same. i.e. override subnet domain option and default router domain name with client specific reservation domain name option.

AdSchellevis:
fix the flow order and you have a chance of it being merged.

NOYB:
flow order changed with else if
thanks

src/opnsense/mvc/app/controllers/OPNsense/Kea/forms/dialogReservation4.xml
```diff
@@ -68,7 +68,7 @@
         <id>reservation.option_data.domain_name</id>
         <label>Domain name</label>
         <type>text</type>
-        <help>The domain name to offer to the client, set to this firewall's domain name when left empty</help>
+        <help>The domain name to offer to the client, set to interface's domain name or this firewall's domain name when both left empty</help>
```
fichtner: I think explaining the code here in the help text will both be confusing and prone to bitrot

```suggestion
        <help>A custom domain name to offer to the client.</help>
```

src/opnsense/mvc/app/controllers/OPNsense/Kea/forms/dialogSubnet4.xml
```diff
@@ -78,7 +78,7 @@
         <id>subnet4.option_data.domain_name</id>
         <label>Domain name</label>
         <type>text</type>
-        <help>The domain name to offer to the client, set to this firewall's domain name when left empty</help>
+        <help>The domain name to offer to the client, set to client's reservation domain name or this firewall's domain name when both left empty</help>
```
fichtner:
```suggestion
        <help>A custom domain name to offer to the clients.</help>
```

fichtner (approved):
Looks good, thanks

PR #9188, opened 2025-09-09
unbound: add support for TXT records in host overrides

swhite2:


src/etc/inc/plugins.inc.d/unbound.inc
```diff
@@ -632,6 +632,8 @@ function unbound_add_host_entries($ifconfig_details)
                         case 'MX':
                             $unbound_entries .= "local-data: \"{$alias['hostname']}{$alias['domain']} {$host->ttl} IN MX {$host->mxprio} {$host->mx}\"\n";
                             break;
+                        case 'TXT':
+                            $unbound_entries .= "local-data: '{$alias['hostname']}{$alias['domain']} {$host->ttl} IN TXT \"" . addslashes($host->txtdata) . "\"'\n";
```
fichtner, line 636:
```suggestion
                            $unbound_entries .= "local-data: '{$alias['hostname']}{$alias['domain']} {$host->ttl} IN TXT \"" . addslashes($host->txtdata) . "\"'\n";
                            break;
```

swhite2:
> I'm a bit sceptical about the addslashes() doing exactly what unbound requires for escaping, but these things are always difficult to ensure in practice.

Agreed, but at least the `txtsupport` option doing the same thing has withstood the test of time

PR #9204, opened 2025-09-12
mvc: disable Dnsmasq/Unbound template generation; closes #8888

### LINKED ISSUE #8888: services: inconsistent template reconfigure behavior
**Important notices**

**Describe the bug**

Both Dnsmasq and Unbound use the standard reconfigure action in `ApiMutableServiceControllerBase`. This stops the service, reloads the template and starts it. These services use `pluginctl` to start, which then does the same sequence.

Both services therefore stop and (possibly) reload their templates twice.

**To Reproduce**

1. Click 'apply' on any of the above services.
2. Logs show service attempting to be stopped, but is already stopped.

**Expected behavior**

Services are stopped once

N/A

**Screenshots**

N/A

**Relevant log files**

```
 /usr/local/sbin/pluginctl: The command '/usr/local/etc/rc.d/dnsmasq stop' returned exit code '1', the output was 'dnsmasq not running? (check /var/run/dnsmasq.pid).'
```

fichtner:
There are two things to consider:

1. Stopping does not alter the templates because the (re)start code is in charge of doing the templates, which is not called when a service is disabled.

2. That actually lead to an Unbound bug on stop where the file system is unmounted, but that should only happen when the service is disabled.  This runs into issue 1. again, but it's better to leave the system mounted until the backend kicks in again unmounting it anyway.

### DIFF
--- a/src/etc/inc/plugins.inc.d/dnsmasq.inc
+++ b/src/etc/inc/plugins.inc.d/dnsmasq.inc
@@ -178,7 +178,6 @@ function dnsmasq_configure_do($verbose = false)
 
     _dnsmasq_add_host_entries();
 
-    /* XXX: cleanup, should have been run, but interface changes during boot or wizard are messing things up */
     configd_run('template reload OPNsense/Dnsmasq');
 
     mwexec('/usr/local/etc/rc.d/dnsmasq start');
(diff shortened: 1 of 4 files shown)


fichtner:
Going with the smallest implementation here trading a side effect for another more minor side effect.

PR #9220, opened 2025-09-18
Allow source NAT to use interface networks

stevekerrison:
The old interface only lists interface addesses, not nets, and the rule parser parser only looks for interface names suffixed by `ip`. In pf, interface address is used dynamically by being surrounded by parenthesis and includes the ":0" modifier to exclude aliases. By combining this with the ":network" modifier, source nat can also use an interface's network, dynamically, i.e. "($iface:network:0)".

This is different to one-to-one nat because the source and target networks do not need to be the same size. Any of pf's pool options are valid in this situation.

Noticing PR #8495, I think updating the refactoring done there to include support for this would be fairly simple, with an update around https://github.com/opnsense/core/pull/8495/files#diff-fa864e1de379b514cb12d89d3e3c738b31c6c340fba099be98ba03520035c676R79

I made these changes because I wanted to source NAT onto a prefix delegated subnet, where the source subnet is larger than the target prefix and so masking/NPTv6 won't work, and the prefix is dynamic (DHCP assigned by ISP) so so must be based on an interface reference.


AdSchellevis:
The legacy outbound nat page is on its way out (https://github.com/opnsense/core/issues/9515), which means there's a feature freeze.

stevekerrison:
> The legacy outbound nat page is on its way out (#9515), which means there's a feature freeze.

When I was initially working on this the new page was read-only, so I suppose if that's no longer the case and it's usable on 16.1 then I'll see what needs updating there instead.

fichtner:
Not sure if the new page supports it, but we're not going to add it to the old one being feature-freezed as mentioned.

Issue #9216, opened 2025-09-17
captive portal: mac bypass is case sensitive


mimugmail:
**Important notices**

Before you add a new report, we ask you kindly to acknowledge the following:

- [x] I have read the contributing guide lines at https://github.com/opnsense/core/blob/master/CONTRIBUTING.md
- [x] I am convinced that my issue is new after having checked both open and closed issues at https://github.com/opnsense/core/issues?q=is%3Aissue

**Describe the bug**

It seems that when you use mac bypass for authentication, case sensitive is a thing. 
We have a script collecting MAC addresses and write them in 00:FF:CC:11:22.BB, but when you look at the arp table it's in lower case. Traffic is blocked until I add the MAC in lower case. 

Is this expected behaviour? :)

*Tip*: to validate your setup was working with the previous version, use opnsense-revert (https://docs.opnsense.org/manual/opnsense_tools.html#opnsense-revert)

**To Reproduce**

See above

**Expected behavior**

**Describe alternatives you considered**

**Screenshots**

**Relevant log files**

**Additional context**

**Environment**

Software version used and hardware type if relevant, e.g.:

OPNsense 25.7.3_7-amd64
FreeBSD 14.3-RELEASE-p2
OpenSSL 3.0.17

PR #9202, opened 2025-09-11
mvc: proposal for https://github.com/opnsense/core/issues/9199

swhite2:

PR #9223, opened 2025-09-18
captive portal: case insensitive MAC parsing

swhite2:
Fixes https://github.com/opnsense/core/issues/9216

There shouldn't be any existing entries in the database, as capitalized MACs would never have worked.

Putting this under review as this may also be fixed on the model layer.

Related to https://github.com/opnsense/core/pull/9202


AdSchellevis (approved):
in terms of amount of code to be added, this is likely the quickest option. 

The second loop has a disadvantage for being inside the list_clients() loop, but given the amount of entries is likely small, I think we should you just merge this.

mimugmail:
Does this mean loading a list of 1500 mac addresses will slow down?

AdSchellevis:
could be the case, but also easy to test.

Issue #9229, opened 2025-09-20
Question mark in host name allowed, but crashed dnsmasq.


AndreasAanestad:
Version: OPNsense 25.7.3_7-amd64

When writing a host name in /ui/dnsmasq/settings#hosts you do do not get a warning when you try to apply a host name containing a question mark.

This will cause the service to crash:

Logs:

2025-09-20T09:44:50Criticaldnsmasqbad DHCP host name at line 53 of /usr/local/etc/dnsmasq.conf
2025-09-20T09:40:25CriticaldnsmasqFAILED to start up

You do get a warning if you use a space, so there must be a regex or something that checks this. 

Steps to reproduce:
1. Services -> DnsMasq -> hosts
2. Change name of a host to test?
3. save

Monviech:
https://github.com/imp/dnsmasq/blob/770bce967cfc9967273d0acfb3ea018fb7b17522/src/option.c#L3810-L3814

https://github.com/imp/dnsmasq/blob/770bce967cfc9967273d0acfb3ea018fb7b17522/src/util.c#L187-L218

Legal chars in hostnames in dnsmasq:

https://github.com/imp/dnsmasq/blob/770bce967cfc9967273d0acfb3ea018fb7b17522/src/util.c#L188

```
a-z A-Z 0-9 - _ .
```
And a restriction that a label cannot start with ```- _```

https://github.com/imp/dnsmasq/blob/770bce967cfc9967273d0acfb3ea018fb7b17522/src/util.c#L207C7-L207C44

PR #9232, opened 2025-09-22
dnsmasq: strict hostname and domain validation

Monviech:
Fixes: https://github.com/opnsense/core/issues/9229

Unit test succeeds:

```
root@opnsense-dev:/usr/local/opnsense/mvc/tests # phpunit app/models/OPNsense/Dnsmasq/FieldTypes/LegalHostnameFieldTest.php
PHPUnit 9.6.23 by Sebastian Bergmann and contributors.

Runtime:       PHP 8.3.23
Configuration: /usr/local/opnsense/mvc/tests/phpunit.xml

.........                                                           9 / 9 (100%)

Time: 00:00.002, Memory: 24.89 MB

OK (9 tests, 17 assertions)
```


src/opnsense/mvc/app/models/OPNsense/Dnsmasq/FieldTypes/LegalHostnameField.php
```diff
@@ -0,0 +1,92 @@
...
+ * ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
+ * POSSIBILITY OF SUCH DAMAGE.
+ */
+
+namespace OPNsense\Dnsmasq\FieldTypes;
+
+use OPNsense\Base\FieldTypes\HostnameField;
+use OPNsense\Base\Validators\CallbackValidator;
+
+class LegalHostnameField extends HostnameField
```
swhite2: `parent::getValidators()` isn't used, which (if intentional) means we're not interested in `HostnameField` validation, better extend from `BaseSetField` directly then

src/opnsense/mvc/app/models/OPNsense/Dnsmasq/Dnsmasq.xml
```diff
@@ -80,11 +80,7 @@
...
         </no_ident>
         <hosts type="ArrayField">
-            <host type="HostnameField">
-                <HostWildcardAllowed>Y</HostWildcardAllowed>
-                <IsDNSName>Y</IsDNSName>
-                <IpAllowed>N</IpAllowed>
-            </host>
+            <host type=".\LegalHostnameField"/>
             <domain type="TextField">
                 <Mask>/^(?:(?:[a-z0-9]|[a-z0-9][a-z0-9\-]*[a-z0-9])\.)*(?:[a-z0-9]|[a-z0-9][a-z0-9\-]*[a-z0-9])$/i</Mask>
```
fichtner: Getting rid of the mask would be nice ;)

Monviech: It crossed my mind as well, I can clean that up too.

src/opnsense/mvc/app/models/OPNsense/Dnsmasq/FieldTypes/LegalHostnameField.php
Monviech: Thank you, fixed that.

Monviech:
@fichtner I wasnt really inventing any name I just copied it from here xD

https://github.com/imp/dnsmasq/blob/770bce967cfc9967273d0acfb3ea018fb7b17522/src/util.c#L191

Can use a different name if theres a better suggestion.

fichtner:
If it's upstream nomenclature discard my concern. Easy to explain this and the link is already in it :)

swhite2:
Do we expect these field types to be used elsewhere in the dnsmasq model? Since it's been marked as a `BaseSetField` and so it can be a list type.

src/opnsense/mvc/app/models/OPNsense/Dnsmasq/FieldTypes/LegalDomainField.php
```diff
@@ -0,0 +1,106 @@
...
+     * but with some differences as individual labels are processed here.
+     * IPv4 addresses are valid, "." and numbers are allowed in labels.
+     * IPv6 addresses are rejected implicitly as ":" is not allowed.
+     * https://github.com/imp/dnsmasq/blob/770bce967cfc9967273d0acfb3ea018fb7b17522/src/util.c#L191
+     *
+     * @return array list of validators for this field
+     */
+    public function getValidators()
+    {
+        $sender = $this;
```
AdSchellevis:
```suggestion
        $validators = parent::getValidators();
        $sender = $this;
```

When inheriting a field type, best keep its existing generic validators in place for clarity.

src/opnsense/mvc/app/models/OPNsense/Dnsmasq/FieldTypes/LegalDomainField.php
```diff
@@ -0,0 +1,106 @@
...
+
+        return [
+            new CallbackValidator([
```
AdSchellevis: in which case we can add the additional requirements as well.

```
$validators[] = new CallbackValidator(....
```

src/opnsense/mvc/app/models/OPNsense/Dnsmasq/FieldTypes/LegalDomainField.php
```diff
@@ -0,0 +1,106 @@
...
+ * ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
+ * POSSIBILITY OF SUCH DAMAGE.
+ */
+
+namespace OPNsense\Dnsmasq\FieldTypes;
+
+use OPNsense\Base\FieldTypes\BaseSetField;
+use OPNsense\Base\Validators\CallbackValidator;
+
+class LegalDomainField extends BaseSetField
```
AdSchellevis: Is `BaseSetField` the proper type here to inherit from? both host and domain seem to be single entries

src/opnsense/mvc/app/models/OPNsense/Dnsmasq/FieldTypes/LegalDomainField.php
```diff
@@ -0,0 +1,106 @@
...
+        $sender = $this;
+
+        return [
+            new CallbackValidator([
+                "callback" => function ($data) use ($sender) {
+                    $response = [];
+
+                    foreach ($sender->iterateInput($data) as $value) {
+                        // Allow empty domain (combination of all labels)
+                        if ($value === '') {
```
AdSchellevis: we can then move this to the xml spec for being required.

src/opnsense/mvc/app/models/OPNsense/Dnsmasq/FieldTypes/LegalDomainField.php
```diff
@@ -0,0 +1,106 @@
...
+                            continue;
+                        }
+
+                        // Split domain into labels
+                        $labels = explode('.', $value);
+
+                        foreach ($labels as $label) {
+                            // Empty labels are not valid
```
AdSchellevis: I think we can replace the whole block below with a single regex (and one error label), something like:

```
^(?![_-])([a-zA-z0-9\-_]){1,63}
```

The loop over the domain parts still helps to simplify the regex, but this will shrink the code quite significantly.

src/opnsense/mvc/app/models/OPNsense/Dnsmasq/FieldTypes/LegalDomainField.php
```diff
@@ -0,0 +1,106 @@
...
+use OPNsense\Base\Validators\CallbackValidator;
+
+class LegalDomainField extends BaseSetField
+{
+    /**
+     * {@inheritdoc}
+     */
+    protected function defaultValidationMessage()
+    {
+        return gettext("Please enter a valid domain name.");
```
AdSchellevis: If we could think of one phrase that explains the boundaries, that might even be better, but I'm not sure we can.

AdSchellevis (changes requested):
@Monviech I've checked `LegalDomainField`, but the same comments likely apply to hostname as well.

src/opnsense/mvc/app/models/OPNsense/Dnsmasq/FieldTypes/LegalDomainField.php
Monviech: Wouldn't shrinking the code sacrifice the user experience? It throws whats wrong more granularly right now which makes it easier for the user to fix the validation error.

AdSchellevis: maybe a bit, but not that much, certainly not if we offer an error message that explains the constraints. more code eventually is more maintenance burden which we need to weigh in the equation.

AdSchellevis (approved):
@Monviech nice! easy to read and to the point.

Monviech:
@AdSchellevis Thanks. I still want to draft this some more as I can still use a BaseSetField so I can also use it for the Alias and Cname fields. They might not crash dnsmasq on invalid input but the story seems better if all fields in that dialog have the same validation.

AdSchellevis:
@Monviech take your time, there's no rush in merging the fix.

fichtner (approved):
yay for unit test! :)

src/opnsense/mvc/app/models/OPNsense/Dnsmasq/FieldTypes/LegalHostnameField.php
```diff
@@ -0,0 +1,172 @@
...
+ * ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
+ * POSSIBILITY OF SUCH DAMAGE.
+ */
+
+namespace OPNsense\Dnsmasq\FieldTypes;
+
+use OPNsense\Base\FieldTypes\BaseSetField;
+use OPNsense\Base\Validators\CallbackValidator;
+
+class LegalHostnameField extends BaseSetField
```
AdSchellevis: @Monviech Maybe we should just use this `HostnameField`, after all it's just a dnsmasq specific hostname implementation, legal sounds a bit like "see you in court" ;) The rest looks good.

Monviech: Okay I will update the name :)

Issue #9006, opened 2025-07-28
bootgrid: column properties behavior on localStorage persistence


swhite2:
**Important notices**

Before you add a new report, we ask you kindly to acknowledge the following:

- [x] I have read the contributing guide lines at https://github.com/opnsense/core/blob/master/CONTRIBUTING.md
- [x] I am convinced that my issue is new after having checked both open and closed issues at https://github.com/opnsense/core/issues?q=is%3Aissue

**Is your feature request related to a problem? Please describe.**

Currently grids where either columns are added/removed or the width is changed manually/dynamically, there is only a hard toggle to flip the persistence to "on", locking the column properties in place through localStorage on the next page load. This behavior works mostly fine in default setups, but may lead to visual problems in more complicated pages such as the Automation -> Filter rule, where "Inspect" will dynamically add columns.

**Describe the solution you like**

While the Tabulator implementation contains quite some hooks to extend this behavior, none of them are quite viable to properly fix this - as we either hook in too late when determining if we should flip on persistence, or we don't get the required properties at all based on the combination of other settings in the table.

The only proper way to fix this would be to inject a custom persistence module, which can be largely based on the original implementation, but simpler: https://github.com/olifolkerd/tabulator/blob/master/src/js/modules/Persistence/Persistence.js

**Describe alternatives you considered**

N/A

**Additional context**

N/A

swhite2:
Note to self: some custom persistence logic has also been added through https://github.com/opnsense/core/commit/e2ed3138a5ad3d2f207501954a9193b1ce061fa2

Monviech:
This one might be able to get parked because here we can fix the only spot it (reoccuring) happens: https://github.com/opnsense/core/pull/9272

swhite2:
Turns out we're able to work around this problem with https://github.com/opnsense/core/commit/be35c9f276fc5e583525bba398991d06b393de26. Simply deleting properties before they're stored will make sure their auto-calculation applies on a page reload.

This also solves page reloads on different resolution windows/screens showing tables that are smaller than the page width.

Issue #7777, opened 2024-08-15
dashboard: switch to static loading of JS base class assets


Monviech:
**Important notices**

Our forum is located at https://forum.opnsense.org , please consider joining discussions there in stead of using GitHub for these matters.

Before you ask a new question, we ask you kindly to acknowledge the following:

- [x] I have read the contributing guide lines at https://github.com/opnsense/core/blob/master/CONTRIBUTING.md
- [x] I am convinced that my issue is new after having checked both open and closed issues at https://github.com/opnsense/core/issues?q=is%3Aissue

Please reconsider the use of import maps for the new dashboard, as it breaks it completely for alternative browsers that do not support that feature (yet).

It seems like there are some OPNsense users who like to use privacy focused open source browsers.

An example is Palemoon (an open source Firefox alternative).

PR: https://forum.opnsense.org/index.php?topic=42231.0

Commit in question: https://github.com/opnsense/core/commit/c73f63c3fa7271f28192d66f8a6f1280e0ca0c2c

Feature: https://github.com/WICG/import-maps#readme

fichtner:
Privacy on the firewall GUI? I know for a fact that some people have different or dedicated browsers to administrate systems. The real benefit of the feature is to avoid other caching workarounds. I don’t think it will be easy to argue against the bulk of the users not using these browsers here.

Just a few thoughts for discussion.


Cheers,
Franco

fichtner:
The compatibility list is quite extensive BTW. This one is even from Mozilla. Not sure where the Chrome hat is at. https://developer.mozilla.org/en-US/docs/Web/HTML/Element/script/type/importmap

wolfbeast:
Browsers not necessarily on a commercial development path in terms of how quickly they can implement new features will be behind in support for otherwise unannounced changes to the HTML living standard (that changes daily). Even Mozilla indicated that this feature was complex even for their large team at time of implementation consideration:

> From our side: at present, there isn't bandwidth to eagerly implement this proposal at the cost of other high priority work. Coupled with the experimental status, the shifting space of complex multi-resource applications, and the implementation complexity which may interfere with ongoing work, it doesn't make sense to implement immediately.

Goanna/Pale Moon/Basilisk/... is in a similar situation.
[...]

PR #9239, opened 2025-09-25
bootgrid: allow usage of Javascript Modules

swhite2:
This PR attempts to generically solve the usage of JS modules. In the specific case of bootgrid, this is necessary for https://github.com/opnsense/core/issues/9006. Specifically, we want to extend the ESM-aware library code of Tabulator, but cannot do so cleanly without import statements.

To use such modules, one can specify:

```
<script type="module">
import {Something} from "Module";
</script>
```

In the UIBootgrid case, the single `import` in `default.volt` will expose both `UIBootgrid` and `$(<element>).UIBootgrid()` for all sub-templates.

The ControllerBase and default volt template have been adjusted to allow for new inclusions as well.

I am aware of the discussion in https://github.com/opnsense/core/issues/7777 and the associated risk with using `importmap`, which is why I'm opening this as a PR.

Note that this PR switches the Tabulator UMD build to an ESM one.


swhite2:
Update: polyfilling using https://github.com/guybedford/es-module-shims does not seem to work for Pale Moon, so including it to combat incompatibility is not an option

Monviech:
I'm okay with using import maps, its a HTML standard now and there are enough browser choices that support it.

https://html.spec.whatwg.org/multipage/webappapis.html#import-maps
 
https://github.com/WICG/import-maps?tab=readme-ov-file#readme

https://caniuse.com/import-maps

fichtner:
> there are enough browser choices that support it

Yet!

Issue #9238, opened 2025-09-25
firewall/automation: Improve responsiveness of grid and action bar


Monviech:
**Important notices**

Before you add a new report, we ask you kindly to acknowledge the following:

- [x] I have read the contributing guide lines at https://github.com/opnsense/core/blob/master/CONTRIBUTING.md
- [x] I am convinced that my issue is new after having checked both open and closed issues at https://github.com/opnsense/core/issues?q=is%3Aissue

**Is your feature request related to a problem? Please describe.**

media css for devices with small screens for action bar.

MinWidth for colums for better fit:
https://tabulator.info/docs/6.3/layout#fittodata

**Describe the solution you like**

Great responsiveness also for smaller screens or viewport sizes, tablets, and smartphones. 

**Describe alternatives you considered**

None

**Additional context**

None

Monviech:
Fixed via https://github.com/opnsense/core/pull/9240

PR #9240, opened 2025-09-25
firewall/automation: Improve responsiveness via minWidth() and @media for action row

Monviech:
Fixes: 

https://github.com/opnsense/core/issues/9238

Now it works nice on small screens too (e.g. tablets)

minWidth() makes sure no column collapses below a certain floor, making it scrollable vertically while maintaining the information level needed to make it usable.

image

### DIFF
--- a/src/opnsense/mvc/app/controllers/OPNsense/Firewall/forms/dialogFilterRule.xml
+++ b/src/opnsense/mvc/app/controllers/OPNsense/Firewall/forms/dialogFilterRule.xml
@@ -29,6 +29,7 @@
             <!-- The sequence order of firewall rules is absolute, no other field shall be sorted -->
             <order>asc</order>
             <sortable>false</sortable>
+            <min-width>120</min-width>
         </grid_view>
     </field>
     <field>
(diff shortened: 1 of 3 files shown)


swhite2:
I'm seeing the same type of change in https://github.com/opnsense/core/commit/efed0ea116349f3f7b650175f69bd9918a3a3242, so perhaps move this to the bootgrid CSS? This way if any action bar is modified (and there are a couple of places), we can at least expect them to neatly stack.

Monviech:
Sure sounds like a plan I will do that.

fichtner:
Sounds like a good approach. I would try to keep this file without theme support. Potentially, everyone can use an existing theme file to override?

AdSchellevis:
In theory we could first import our "fixed" style and let the theme overlap what they want, in most cases you can overwrite styles anyway. It's not perfect, but prevent us updating all themes again (which would be the second time for these grids)

Monviech:
@fichtner @AdSchellevis 

How about this? https://github.com/opnsense/core/pull/9240/commits/f176a54035e744ea2d2c15ff970ac4880ff88145

swhite2 (approved):

PR #9243, opened 2025-09-26
interfaces/overview: Fix flags in interface details, do not show hex code but parsed flags

Monviech:
Before patch:

image

After patch:

image


AdSchellevis (approved):

AdSchellevis:
@Monviech I had to read it twice, but this is indeed the proper fix :)

Monviech:
@AdSchellevis Thank you :)

PR #9249, opened 2025-09-29
Include `authType` in Captive Portal logged-in status response

agoodkind:
When calling `/api/captiveportal/access/status/` while a session is already logged in to the captive portal, the field `authType` is not returned, whereas when logged out, it is returned. This PR adds the field in the logged-in response as well.

Logged in:
Screenshot 2025-09-28 at 21 28 19

Logged out:
Screenshot 2025-09-28 at 21 28 23


AdSchellevis:
there is a performance penalty in doing this as authenticated clients will fetch data from the configuration as well, which strictly speaking doesn't have a lot of value (client was authenticated, the type is a bit redundant anyway)

PR #9255, opened 2025-10-01
diagnostics/packetcapture: Fix permission of capture zip file for wwwonly strict security mode

Monviech:
Otherwise the packet capture is zipped as root and download fails since its not owned by wwwonly. (When strict security mode is enabled)

### DIFF
--- a/src/opnsense/scripts/interfaces/capture.py
+++ b/src/opnsense/scripts/interfaces/capture.py
@@ -234,5 +234,6 @@ if __name__ == '__main__':
 
         # always set file ownership for strict security mode
         os.chown(result['filename'], pwd.getpwnam("wwwonly").pw_uid, grp.getgrnam("wheel").gr_gid)
+        os.chmod(result['filename'], 0o640)
 
     print (ujson.dumps(result))


src/opnsense/scripts/interfaces/capture.py
```diff
@@ -230,4 +232,14 @@ def pcap_reader(filename, eargs=None):
                 zfh.write(filename, os.path.basename(filename))
         zfh.close()
 
+        # decide ownership based on sentinel file for strict security mode (wwwonly)
```
fichtner: wwwonly/wheel permission is quicker and more robust

fichtner: Also keep in mind AppConfig holds the correct ownership data.

Monviech: So the zip file should just always and unconditionally become wwwonly:wheel ?

fichtner:
PS make sure permission is 644, otherwise root will have issues reading it under 600.

Monviech:
> PS make sure permission is 644, otherwise root will have issues reading it under 600.

Do you mean 640? Or do you want it world readable?

fichtner:
640, 660, 644, doesn’t matter too much 😊

Issue #8537, opened 2025-04-10
Support `net_gateway` route exceptions in OpenVPN Instances mode


patphobos:
**Important notices**

Before you add a new report, we ask you kindly to acknowledge the following:

- [X] I have read the contributing guide lines at https://github.com/opnsense/core/blob/master/CONTRIBUTING.md
- [X] I am convinced that my issue is new after having checked both open and closed issues at https://github.com/opnsense/core/issues?q=is%3Aissue

**Is your feature request related to a problem? Please describe.**

In OpenVPN "Instances" configuration mode (VPN > OpenVPN > Instances), it is not currently possible to push custom routes to clients using the special OpenVPN keyword `net_gateway`.

For the purpose of excluding specific IP addresses from being routed through the VPN tunnel — while keeping the default `redirect-gateway def1` behavior — I am missing a solution that will allow me to push exceptions using `net_gateway`.

This functionality is supported by OpenVPN itself and works by pushing routes such as:

push "route 1.2.3.4 255.255.255.255 net_gateway"

The former "Server" configuration mode allowed adding custom advanced directives. The current "Instances" mode does not offer any equivalent, making it impossible to define these route exceptions from the web interface.

**Describe the solution you like**

I would like the OpenVPN "Instances" configuration interface to support pushing routes with `net_gateway` as the gateway.

**Describe alternatives you considered**

Currently, no alternatives are viable within the supported configuration system in OPNsense.

**Additional context**

Two real-world examples where this feature is essential:

1. A client site connects to two different OpenVPN servers from the same provider. When VPN-A pushes a default route, it must exclude the server IP of VPN-B so that the second connection is not routed into the first tunnel.

2. A site uses OpenVPN for general traffic but needs to maintain a fallback/failsafe connection (e.g. monitoring or admin access) to a public IP outside the VPN. That destination must be excluded from the tunnel using a `net_gateway` route.

This feature is natively supported by OpenVPN and is critical in advanced routing and high-availability VPN scenarios.

patphobos:
up ?

HiDef888:
Would be nice to allow an area for user configuration.  I edited my .conf file and it just continues to get overwritten, so even that is not possible it seems.

r000m5:
+1

PR #9258, opened 2025-10-03
OpenVPN: add support for pushing excluded routes via net_gateway

patphobos:
This patch introduces a new `push_excluded_routes` option in the OpenVPN instance dialog. It allows administrators to define routes that should be excluded from the VPN tunnel and instead use the client’s existing default gateway (`net_gateway`).

Changes include:
- Added `instance.push_excluded_routes` field to dialogInstance.xml.
- Extended OpenVPN model to process `push_excluded_routes` and append `net_gateway` to the pushed route options.
- Defined `push_excluded_routes` as a NetworkField in OpenVPN.xml.

This provides a convenient way to push split-exclusion routes, ensuring specific traffic is routed via the client’s pre-existing default gateway.

Fixes: #8537


AdSchellevis:
@patphobos thanks! merged to be included in a future release.

Issue #9263, opened 2025-10-05
Allow `*` domain in combination with `#` IPset/Firewall alias wildcard to store all resolved IPs


narub-s:
**Important notices**

Before you add a new report, we ask you kindly to acknowledge the following:

- [x] I have read the contributing guide lines at https://github.com/opnsense/core/blob/master/CONTRIBUTING.md
- [x] I am convinced that my issue is new after having checked both open and closed issues at https://github.com/opnsense/core/issues?q=is%3Aissue

**Is your feature request related to a problem? Please describe.**

dnsmasq allows to set `#` [wildcard](https://lists.thekelleys.org.uk/pipermail/dnsmasq-discuss/2013q1/006882.html): `ipset=/#/resolved` catches all resolved domains and stores their IPs in `resolved` IPset. This e.g. is useful for domain-based whitelisting.

It already is possible to set `*` as domain, as described in [OPNsense docs](https://docs.opnsense.org/manual/dnsmasq.html#dnsmasq-as-primary-dns-resolver). It would be great to store all resolved domains in a firewall alias/IPset as well. But selecting an alias in combination with `*` domain triggers error
> Top level wildcard entries are not allowed for Ipset.

**Describe the solution you like**

Allow to use domain `*` with a firewall alias `myalias`, which adds `ipset=/#/myalias` to config.

**Describe alternatives you considered**

Manually maintaining dnsmasq config file

**Additional context**

https://forum.opnsense.org/index.php?topic=49161.0

Monviech:
Im not a fan of storing that much in an alias, but since the default table sizes are increased, maybe?

Maybe @AdSchellevis can decide on this one.

AdSchellevis:
Also not a fan, this can easily overflow and will likely cause unexpected issues when it does.

Monviech:
The domain field already accepts '#' so nothing prevents this configuration for ipset right now.

narub-s:
Fair point about the default table size.
1000000 entries is more than enough for me, but it certainly depends on environment/orga.
But I am bit a bit surprised to hear about those unexpected side effects  - shouldn't OPNsense handle table overflows more gracefully, like in this case?

Btw: If you set `#` as domain, config is created as follows:
```
server=/#/127.0.0.1#5335
ipset=/#/dns_resolved_ips
rebind-domain-ok=/#/
```
ipset `/#/` would be OK, but `#` in `server=` specifically has a different meaning:

[...]

Issue #9262, opened 2025-10-05
Allow to set IPSet/Firewall alias with empty IP address and port for Dnsmasq domain override


narub-s:
**Important notices**

Before you add a new report, we ask you kindly to acknowledge the following:

- [x] I have read the contributing guide lines at https://github.com/opnsense/core/blob/master/CONTRIBUTING.md
- [x] I am convinced that my issue is new after having checked both open and closed issues at https://github.com/opnsense/core/issues?q=is%3Aissue

**Is your feature request related to a problem? Please describe.**

`ipset=` in dnmasq in general [does not rely](https://thekelleys.org.uk/dnsmasq/docs/dnsmasq-man.html) on a forwarding server and port, just the domain. But adding domain overrides in Dnsmasq > Domains > Add/Edit in OPNsense currently requires a forwarding server ("IP address") and its port always to be specified. 

This is unneeded and a bit repetitive work for multiple IPsets given same server/port. Relation between forwarding server, domain and IPset might become unclear for users, causing confusion.

When a new domain override with same server/port is created for purpose of adding a new IPset/firewall alias, redundant `server=` entries are currently added to dnsmasq config file `/usr/local/etc/dnsmasq.conf`. Multiple `server=` entries in general might be problematic, as dnsmasq per default sends DNS queries to all upstream servers concurrently (not sure about the case, where servers are same IP).

**Describe the solution you like**

Allow to keep IP address/port empty and to just fill domain in combination with IPset/Firewall alias. This will create an `ipset=`, but no `server=` entry in config.

Note that an empty IP in `server=/<domain>/<ip>` for dnsmasq means, that this domain should never be forwarded by dnsmasq to upstream. If OPNsense supports this case, then it would be a good idea to make this feature explicit in the webform (e.g. a toggle check box "Never forward to upstream"), also to not cause ambiguities.

**Describe alternatives you considered**

Former preferred solution: Separate grid/tab "IPSet" under Services > Dnsmasq & DHCP, which allows to create a domain/IPset mapping, similar to dnsmasq notation `ipset=/<domain>/<ipset_name>`, but I got feedback, this is out of scope.

**Additional context**
As discussed in https://forum.opnsense.org/index.php?topic=49161.0

Monviech:
This is probably fine to allow with little changes needed.

PR #9269, opened 2025-10-07
dnsmasq: Improve Firewall alias (ipset) validations

Monviech:
Closes: https://github.com/opnsense/core/issues/9263

- domain in domainoverrides is changed to the strict HostnameField introduced in 2d2781c to disallow the use of "#" which would add all resolved domains to an ipset alias. The overall strictness of the field now matches hosts.

Fixes: https://github.com/opnsense/core/issues/9262

- dnsmasq.conf template does not render server and rebind-domain-ok for entries anymore which are just a domain and the Firewall alias (ipset). Before this change, it would generate a server with an empty IP, which means blocking forwarding. Firewall alias (ipset) needs forwarding to function, otherwise it fails for the chosen domain.

Unrelated to the above tickets, but fixed regardless:

- ipset ModelRelationField got a name filter adjustment, to exclude all system defined external aliases that start with __

### DIFF
--- a/src/opnsense/mvc/app/models/OPNsense/Dnsmasq/Dnsmasq.xml
+++ b/src/opnsense/mvc/app/models/OPNsense/Dnsmasq/Dnsmasq.xml
@@ -130,11 +130,10 @@
                 <Required>Y</Required>
                 <ValidationMessage>Please enter a value between 1 and 99999.</ValidationMessage>
             </sequence>
-            <domain type="HostnameField">
-                <HostWildcardAllowed>Y</HostWildcardAllowed>
-                <IsDNSName>Y</IsDNSName>
-                <IpAllowed>N</IpAllowed>
+            <domain type=".\HostnameField">
                 <Required>Y</Required>
+                <IsDomain>Y</IsDomain>
+                <IsWildcardAllowed>Y</IsWildcardAllowed>
             </domain>
             <ipset type="ModelRelationField">
                 <Model>
@@ -144,7 +143,7 @@
                         <display>name</display>
                         <filters>
                             <type>/^[Ee]xternal.*/</type>
-                            <name>/^(?!bogons$|bogonsv6$|virusprot$|sshlockout$|__captiveportal.*).*/</name>
+                            <name>/^(?!bogons$|bogonsv6$|virusprot$|sshlockout$|__.*).*/</name>
                         </filters>
                     </aliases>
                 </Model>
--- a/src/opnsense/service/templates/OPNsense/Dnsmasq/dnsmasq.conf
+++ b/src/opnsense/service/templates/OPNsense/Dnsmasq/dnsmasq.conf
@@ -125,12 +125,16 @@ bogus-priv
 
 {# Domain Overrides can be sorted by sequence to support strict order #}
 {% for override in helpers.toList('dnsmasq.domainoverrides', sortBy='sequence', sortAs='int') %}
+{#    Empty IP means to block forwarding for that domain, but forwarding is a requirement for ipset #}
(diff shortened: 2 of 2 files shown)


src/opnsense/service/templates/OPNsense/Dnsmasq/dnsmasq.conf
```diff
@@ -125,13 +125,17 @@ bogus-priv
 
 {# Domain Overrides can be sorted by sequence to support strict order #}
 {% for override in helpers.toList('dnsmasq.domainoverrides', sortBy='sequence', sortAs='int') %}
+{#    Empty IP means to block forwarding for that domain, but forwarding is a requirement for ipset #}
+{#    That is why generating the server is skipped for entries with ipset set but empty IP. #}
+{%    if override.ip or not override.ipset %}
```
AdSchellevis, line 130: probably more a question, but is `/my.domain/` without an address valid data?

Monviech:
```
-S, --local, --server=[/[<domain>]/[domain/]][<server>[#<port>]][@<interface>][@<source-ip>[#<port>]]

Also permitted is a -S flag which gives a domain but no IP address; this tells dnsmasq that a domain is local and it may answer queries from /etc/hosts or DHCP but should never forward queries on that domain to any upstream servers.
```

https://github.com/opnsense/core/blob/d1042bb65e2f170780040e1aff635eda3c14e10b/src/opnsense/mvc/app/controllers/OPNsense/Dnsmasq/forms/dialogDomainOverride.xml#L22

It matches with our existing help text.

src/opnsense/mvc/app/models/OPNsense/Dnsmasq/Dnsmasq.xml
```diff
@@ -130,11 +130,10 @@
                 <Required>Y</Required>
                 <ValidationMessage>Please enter a value between 1 and 99999.</ValidationMessage>
             </sequence>
-            <domain type="HostnameField">
-                <HostWildcardAllowed>Y</HostWildcardAllowed>
-                <IsDNSName>Y</IsDNSName>
-                <IpAllowed>N</IpAllowed>
+            <domain type=".\HostnameField">
```
Monviech, line 133: @AdSchellevis I'm unsure if I really want to prevent "#", its also valid in the server directive:

```
-S, --local, --server=[/[<domain>]/[domain/]][<server>[#<port>]][@<interface>][@<source-ip>[#<port>]]

The special server address '#' means, "use the standard servers", so --server=/google.com/1.2.3.4 --server=/[www.google.com](http://www.google.com/)/# will send queries for google.com and its subdomains to 1.2.3.4, except [www.google.com](http://www.google.com/) (and its subdomains) which will be forwarded as usual.
```

Though for ipset it means to store all resolved domains, I tested that.

```
--ipset=/<domain>[/<domain>...]/<ipset>[,<ipset>...]
(.....) Domains and subdomains are matched in the same way as --address.
```

```
-A, --address=/<domain>[/<domain>...]/[<ipaddr>]
The domain specification works in the same way as for --server, with the additional facility that /#/ matches any domain.
```

Unsure if the strict validation is the right way, or if it should just be left alone.

AdSchellevis: I don't mind that much adding `#` for `ipset`, but without boundaries someone will wreck his machine and come back complaining

Monviech: I rather constrain the use of # then, if somebody really wants it they can add an override in their dnsmasq configuration. I dont want hard to debug errors with tables.

PR #9272, opened 2025-10-08
firewall/automation: Consolidate the stats column into a combined always visible one, implement cache invalidation via refresh button

Monviech:
Proposal to fix the issue with showing and hiding columns in the most minimal way.

- Stat column always shows
- It shows icons for each type of stat, but only if that stat has data
- A refresh button in the same column can invalidate the cache via https://github.com/opnsense/core/commit/27bd8125

Reference to alternative: https://github.com/opnsense/core/pull/9239

### DIFF
--- a/src/opnsense/mvc/app/controllers/OPNsense/Firewall/Api/FilterController.php
+++ b/src/opnsense/mvc/app/controllers/OPNsense/Firewall/Api/FilterController.php
@@ -458,4 +458,17 @@ class FilterController extends FilterBaseController
 
         return $result;
     }
+
+    public function flushInspectCacheAction()
+    {
+        if (!$this->request->isPost()) {
+            return ['status' => 'error', 'message' => gettext('Invalid request method')];
+        }
+
+        // The "!" prelude flushes any cached script_output before execution
+        (new Backend())->configdRun('!filter rule stats');
+
+        return ['status' => 'ok'];
+    }
+
 }
--- a/src/opnsense/mvc/app/views/OPNsense/Firewall/filter_rule.volt
+++ b/src/opnsense/mvc/app/views/OPNsense/Firewall/filter_rule.volt
@@ -41,11 +41,6 @@
             });
         }
 
-        // Show statistics columns when inspect button is checked
-        function updateStatisticColumns() {
-            grid.bootgrid(inspectEnabled ? "setColumns" : "unsetColumns", ['evaluations', 'states', 'packets', 'bytes']);
-        }
-
         // Get all advanced fields, used for advanced mode tooltips
         const advancedFieldIds = "{{ advancedFieldIds }}".split(',');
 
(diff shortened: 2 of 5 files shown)


src/opnsense/mvc/app/controllers/OPNsense/Firewall/Api/FilterController.php
```diff
@@ -458,4 +458,17 @@ public function getInterfaceListAction()
...
         return $result;
     }
+
+    public function flushInspectCacheAction()
+    {
+        if (!$this->request->isPost()) {
+            return ['status' => 'error', 'message' => gettext('Invalid request method')];
+        }
+
+        // The "!" prelude flushes any cached script_output before execution
```
fichtner: I don't think we need to explain it like this, especially since the action is called "flush" :)

```suggestion
```

src/opnsense/service/conf/actions.d/actions_filter.conf
```diff
@@ -98,7 +98,7 @@ message:request pf rules
 [diag.top]
 command:/usr/local/opnsense/scripts/filter/pftop.py
 type:script_output
-cache_ttl:30
+cache_ttl:600
```
fichtner: maybe 5 minutes? the time of a coffee break

```suggestion
cache_ttl:300
```

src/opnsense/mvc/app/controllers/OPNsense/Firewall/Api/FilterController.php
```diff
@@ -458,4 +458,16 @@ public function getInterfaceListAction()
...
+    {
+        if (!$this->request->isPost()) {
+            return ['status' => 'error', 'message' => gettext('Invalid request method')];
+        }
+
+        (new Backend())->configdRun('!filter rule stats');
+
+        return ['status' => 'ok'];
+    }
+
```
fichtner:
```suggestion
```

and before the style eater gets it ;)

fichtner (approved):

src/opnsense/mvc/app/views/OPNsense/Firewall/filter_rule.volt
```diff
@@ -715,8 +768,15 @@
         });
 
         grid.on('loaded.rs.jquery.bootgrid', function () {
-            updateStatisticColumns();  // ensures inspect columns are consistent after reload
-            $("#{{formGridFilterRule['table_id']}}").toggleClass("tree-enabled", treeViewEnabled);
+            const $table = $("#{{formGridFilterRule['table_id']}}");
+            $table.toggleClass("tree-enabled", treeViewEnabled);
+
+            // Bind click event for cache flush icon in statistics header
+            $table.find('.inspect-cache-flush').off('click').on('click', function () {
```
swhite2: If I'm not mistaken, Tabulator now supports returning an HTML element with an attached click event handler directly from the formatter as well, in which case it's not necessary to bind it to a reload.

Example: https://github.com/opnsense/core/blob/master/src/opnsense/mvc/app/views/OPNsense/Diagnostics/log.volt#L65-L79.

Though not entirely sure it's supported by headerFormatter

Monviech: I have moved the click handler in there and it works

src/opnsense/mvc/app/views/OPNsense/Firewall/filter_rule.volt
```diff
@@ -187,6 +182,31 @@
...
+                        return `
+                            <span data-toggle="tooltip"
+                                title="{{ lang._('Statistics') }}">
+                                <i class="fa-solid fa-fw fa-eye"></i>
+                            </span>
+
+                            <span style="cursor:pointer; margin-left:4px;"
+                                data-toggle="tooltip"
+                                title="{{ lang._('Refresh') }}"
+                                onclick="
```
swhite2: please refrain from using inline event handlers due to tight coupling of logic and markup (and other reasons). You can also access the [0] element of any created jquery object to get it's HTML element.

swhite2 (changes requested):

src/opnsense/mvc/app/views/OPNsense/Firewall/filter_rule.volt
Monviech: Thank you, that worked.

Issue #9279, opened 2025-10-10
OpenVPN split DNS with multiple VPN domains


devopsoffice-ivan:
**Important notices**

- [x] I have read the contributing guide lines at https://github.com/opnsense/core/blob/master/CONTRIBUTING.md
- [x] I am convinced that my issue is new after having checked both open and closed issues at https://github.com/opnsense/core/issues?q=is%3Aissue

**Describe the solution you like**

Allow setting multiple domains in OpenVPN instance setting window / Miscellaneous / DNS Default Domain:
https://192.168.11.254/ui/openvpn/instances

Allow setting multiple domains in OpenVPN Client Config Overrides / Client Settings / DNS Default Domain:
https://192.168.11.254/ui/openvpn/client_overwrites

Alternative option:
- "Additional domains" input box after DNS Default Domain input box

This request is related to this issue: https://github.com/opnsense/core/issues/4313

If redirect gateway option isn't used, Linux NetworkManager and some MacOS OpenVPN client implementations only assign the DNS Default Domain to the tunnel interface which is set in config, and there is no way to serve other domains by DNS server set by OpenVPN server.
Sometimes you need to serve an IP address from your own DNS server to a given domain when client is connected to VPN, e.g. an internal URL contains such a domain which is not resolvable by public DNS servers but URL domain differs from VPN domain.
On linux there's a workaround with up scripts which set search domains to interface as domain, but it's a workaround, because search domains have a different purpose and it's not used by NetworkManager. (Command line openvpn use it.)

Related documentations:
https://openvpn.net/connect-docs/dns-servers-mobile-behavior.html
https://openvpn.net/community-docs/community-articles/openvpn-2-6-manual.html

OpenVPN manual doesn't mention repeated dhcp-option DOMAIN usage but iOS example page use multiple DOMAIN option.

My experience is that DNS option without DOMAIN option doesn't catch DNS request at all when I use NetworkManager, so iOS client implementation differs from Gnome NetworkManager.

I tested multiple DOMAIN option by adding a new row after row 667 in src/opnsense/mvc/app/models/OPNsense/OpenVPN/OpenVPN.php (master commit id was: f469cf85296e3da8fe0bb49f4aaee58617d5e065):
$options['push'][] = "\"dhcp-option DOMAIN fake.xyz\"";

After connecting to VPN server, resolvectl status ouput:
DNS Domain: fake.xyz office.devopsoffice.net

[...]

PR #9281, opened 2025-10-12
OpenVPN: allow multiple domains settings for client connection

devopsoffice-ivan:
The change related to issue https://github.com/opnsense/core/issues/9279

It allows multiple domain settings to client connection so multiple DNS domain can be served by DNS server set by OpenVPN server.

### DIFF
--- a/src/etc/inc/plugins.inc.d/openvpn.inc
+++ b/src/etc/inc/plugins.inc.d/openvpn.inc
@@ -365,7 +365,9 @@ function openvpn_validate_cidr($value, $name, $multiple = false, $ipproto = 'ipv
 function openvpn_add_dhcpopts(&$settings, &$conf)
 {
     if (!empty($settings['dns_domain'])) {
-        $conf .= "push \"dhcp-option DOMAIN {$settings['dns_domain']}\"\n";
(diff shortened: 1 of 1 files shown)


AdSchellevis:
@devopsoffice-ivan according to the documentation this should work indeed, just noticed you mis the implementation for client specific overrides.

I expect you need a look here as well:

https://github.com/opnsense/core/blob/f469cf85296e3da8fe0bb49f4aaee58617d5e065/src/etc/inc/plugins.inc.d/openvpn.inc#L368

devopsoffice-ivan:
> @devopsoffice-ivan according to the documentation this should work indeed, just noticed you mis the implementation for client specific overrides.
> 
> I expect you need a look here as well:
> 
> https://github.com/opnsense/core/blob/f469cf85296e3da8fe0bb49f4aaee58617d5e065/src/etc/inc/plugins.inc.d/openvpn.inc#L368

Thanks for the suggestion. It's updated in this commit: https://github.com/opnsense/core/pull/9281/commits/6a4dbbc4e19613736107d4a55a550d4f3696129b

AdSchellevis:
@devopsoffice-ivan looks good, merged it into the master branch for a future release. thanks

PR #9284, opened 2025-10-13
firewall/automation: Hide and unhide statistics without triggering tabulator persistence event

Monviech:
Add transient `_silentToggle` marker to Tabulator columns so `columnVisibilityChanged` events triggered by scripted show/hide operations (e.g. inspect toggle) do not set the persistence flag.

### DIFF
--- a/src/opnsense/www/js/opnsense_bootgrid.js
+++ b/src/opnsense/www/js/opnsense_bootgrid.js
@@ -2040,22 +2040,22 @@ class UIBootgrid {
         this._destroyTable();
     }
 
-    setColumns(columns, silent = false) {
+    setColumns(columns) {
         this.table.getColumns().forEach((col) => {
             const def = col.getDefinition();
             if (columns.includes(def.field)) {
-                col._silentToggle = silent;
+                col._silentToggle = true;
                 col.show();
                 delete col._silentToggle;
             }
         });
     }
 
-    unsetColumns(columns, silent = false) {
+    unsetColumns(columns) {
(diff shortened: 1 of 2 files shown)


src/opnsense/www/js/opnsense_bootgrid.js
```diff
@@ -2038,22 +2040,26 @@ class UIBootgrid {
         this._destroyTable();
     }
 
-    setColumns(columns) {
+    setColumns(columns, silent = false) {
```
AdSchellevis: @Monviech  named parameters in javascript looks a bit suspicious, are you sure this is correct?

Monviech: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Functions/Default_parameters

I wanted it to be an optional parameter with a default set. But since the firewall page is the only consumer right now(of set and unsetColumns), it could also be a default in that function without adding a new parameter, since I dont think its expected that hiding and unhiding colums programatically should trigger persistence.

If I got something wrong here feel free to correct.

fichtner: I vaguely remember crashlanding with this „works as defined“ assumption on parameter defaults a few years back in the firmware page. But I don’t know how many years ago and if it actually matters that we should avoid it anyway. 😉

Monviech: I removed the parameter due to the reason stated above.

Issue #9136, opened 2025-08-26
unbound: block list source selection support for community edition


fichtner:
**Important notices**

Before you add a new report, we ask you kindly to acknowledge the following:

- [x] I have read the contributing guide lines at https://github.com/opnsense/core/blob/master/CONTRIBUTING.md
- [x] I am convinced that my issue is new after having checked both open and closed issues at https://github.com/opnsense/core/issues?q=is%3Aissue

**Is your feature request related to a problem? Please describe.**

Source selection is a business edition feature, but we want to offer a similar functionality for the community edition as well now.

**Describe the solution you like**

Add a GUI using the same available pluggable framework, but likely should not interfere with the business feature.

**Describe alternatives you considered**

Use an external solution for this, which complicates management. Use the business edition, which is not an option for some users.

**Additional context**

N/A

PR #9301, opened 2025-10-20
unbound: merge extended blocklists into core (https://github.com/opnsense/core/issues/9136)

swhite2:
TODO:

- Discuss migration strategy, including `post()` usage to clean up config.
- Improve domain tester policy matching for better traceability.

### DIFF
--- a/src/etc/inc/plugins.inc.d/unbound.inc
+++ b/src/etc/inc/plugins.inc.d/unbound.inc
@@ -190,6 +190,7 @@ function unbound_generate_config()
             RecursiveIteratorIterator::CHILD_FIRST)
         )
     );
+    @rmdir($targetdir . "lib/__pycache__");
 
     $basedir = '/usr/local/opnsense/scripts/unbound-dnsbl/';
     foreach (new RecursiveIteratorIterator(new RecursiveDirectoryIterator($basedir)) as $filename) {


src/opnsense/mvc/app/controllers/OPNsense/Unbound/forms/dnsbl.xml
```diff
@@ -28,23 +26,23 @@
         <help>List of domains from where blocklist will be downloaded.</help>
     </field>
     <field>
-        <id>unbound.dnsbl.whitelists</id>
+        <id>blocklist.whitelists</id>
```
fichtner: while we are here perhaps we could get rid of whitelist. difficult topic with needless complication but from this code there is only one instance of "black" remaining so that side has progressed more already

fichtner (approved):
looks good on the surface

swhite2:
One thing to note here is that with many blocklists, the matching algorithm has quite a performance penalty due to the subnet coverage calculation. This doesn't necessarily have to be a problem, as it's properly locked and is only for pre-processing, but it's slower than before.

src/opnsense/mvc/app/controllers/OPNsense/Unbound/forms/dnsbl.xml
swhite2: done: https://github.com/opnsense/core/pull/9301/commits/045954afa8b195ff3a213662b036581a29cfe32b

swhite2:
Just to ease review a bit, the data path is as follows:

- Input:

```
[blocklist:1]
description=default
rcode=NOERROR
address=0.0.0.0
source_nets=
cache_ttl=72000
excludes.default=.*localhost$
excludes.patterns=
includes.patterns=
includes.wildcards=

[blocklist:2]
description=<second>
rcode=NOERROR
address=0.0.0.0
source_nets=192.168.2.0/24,192.168.1.0/24
cache_ttl=72000
blocklists.<shortcode>=<URL> <-- only exists if one or more types set
blocklists.custom<loop.index>=<URL> <-- only exists if one or more lists set
excludes.default=.*localhost$
excludes.patterns=
includes.patterns=
includes.wildcards=
```

- Output is a single JSON file with the following structure:
```
{
	"data": {
		"<domain>": [
			{"idx": <index into the first logical policy it should check},
			{"idx": ...}
			...
		],
	},
	"config": {
		"<idx>": {
			"passlist": <compiled passlist>,
			"source_nets": [
				"192.168.1.0/24",
				...
			]
			...
		}
	}
}
```

- There may be multiple policies associated with a single domain, as source nets can differ, as well as the allowlist (passlist here) regex.
- Config output is used directly in `dnsbl_module.py` and is returned when queried in test mode.
- To keep the resulting output small, the indices are simple integers - we assume other blocklist handlers may be injected, therefore the index must be managed by the `BlocklistParser` class.
	- We first merge all configurations, separated by index
	- The handlers return their blocklists, which is an array of dictionaries, all of which may contain domains that overlap in the final blocklist.
	- `_merge_items_with_config` merges the blocklists together and decides in what order the policies are attached to the domains -> more specific subnets have priority, and in that set the bigger passlists also have priority.

AdSchellevis:
My changes are in, but can use an extra pair of eyes on the current state.

src/etc/inc/plugins.inc.d/unbound.inc
```diff
@@ -176,11 +176,19 @@ function unbound_generate_config()
     $python_dir = "/usr/local/lib/{$pythonv}";
     $chroot_python_dir = "/var/unbound{$python_dir}";
 
-    $dirs = ['/data', '/dev', '/etc', '/lib', '/run', '/usr', '/var/db', '/var/run', $python_dir];
+    $dirs = ['/data', '/dev', '/etc', '/lib', '/run', '/usr', '/var/db', '/var/run', '/unbound-dnsbl/lib', $python_dir];
 
     foreach ($dirs as $dir) {
         mwexecf('/bin/mkdir -p %s', "/var/unbound{$dir}");
     }
+    /* deploy python module (without test suite) */
```
fichtner, line 185:
```suggestion

    /* deploy python module (without test suite) */
```

src/opnsense/mvc/app/models/OPNsense/Unbound/Unbound.php
```diff
@@ -60,6 +60,30 @@ public function performValidation($validateFullModel = false)
...
+                /* Extract all subnets (eg x.x.x.x/24 --> 24) and protocol families */
+                $sizes = array_unique(
+                    array_map(fn($x) => explode("/", $x)[1] ?? null, explode(",", $node->source_nets))
+                );
+                $ipproto = array_unique(
+                    array_map(fn($x) => strpos($x, ':') == false ? 'inet' : 'inet6' , explode(",", $node->source_nets))
+                );
+                if (count($sizes) > 1) {
+                    $messages->appendMessage(new Message(
+                        gettext("All offered networks should be equally sized to avoid overlaps"),
```
fichtner:
```suggestion
                        gettext('All offered networks should be equally sized to avoid overlaps.'),
```

Mostly about missing dot in sentence.

src/opnsense/mvc/app/models/OPNsense/Unbound/Unbound.php
```diff
@@ -60,6 +60,30 @@ public function performValidation($validateFullModel = false)
...
+                        $node->source_nets->__reference
+                    ));
+                }
+                if (count($ipproto) > 1) {
+                    $messages->appendMessage(new Message(
+                        gettext("All offered networks should use the same IP protocol"),
```
fichtner:
```suggestion
                        gettext('All offered networks should use the same IP protocol.'),
```

swhite2:
Apart from the extra review, points of interest here are:

- We need to make sure we skip validation for items coming from the extended blocklists that do not fit the new validations.
- The `dnsbl.json` file needs to be regenerated at some point. Likely needs a simple check in the Unbound code to reject the old list.

src/etc/inc/plugins.inc.d/unbound.inc
```diff
@@ -176,11 +176,19 @@ function unbound_generate_config()
...
+    $basedir = '/usr/local/opnsense/scripts/unbound-dnsbl/';
+    foreach (new RecursiveIteratorIterator(new RecursiveDirectoryIterator($basedir)) as $filename) {
+        $target = "/var/unbound/unbound-dnsbl/".substr($filename, strlen($basedir));
+        if (str_ends_with($filename, '.py') && is_dir(dirname($target))) {
+            @copy($filename, $target);
```
fichtner, line 199: I don't think we're making sure to clean /var/unbound/unbound-dnsbl so our future selves will be happy?

Issue #9135, opened 2025-08-26
kea-dhcp: automatic route support for PD leases


fichtner:
**Important notices**

Before you add a new report, we ask you kindly to acknowledge the following:

- [x] I have read the contributing guide lines at https://github.com/opnsense/core/blob/master/CONTRIBUTING.md
- [x] I am convinced that my issue is new after having checked both open and closed issues at https://github.com/opnsense/core/issues?q=is%3Aissue

**Is your feature request related to a problem? Please describe.**

ISC DHCPv6 has prefixes.php to deal with this. We need a similar mechanism for Kea.

**Describe the solution you like**

Minimal scripting using available Kea hooks.

**Describe alternatives you considered**

Kea supporting this in a canonical way already.. ;)

**Additional context**

N/A

Monviech:
KEA leases file with IA_NA and IA_PD:

cat /var/db/kea/kea-leases6.csv 

*Primary HA node*

```
address,duid,valid_lifetime,expire,subnet_id,pref_lifetime,lease_type,iaid,prefix_len,fqdn_fwd,fqdn_rev,hostname,hwaddr,state,user_context,hwtype,hwaddr_source,pool_id
2003:ffff:ffff:ff01::100,00:01:00:01:30:2f:4a:3c:00:15:5d:00:ad:23,4000,1761047783,1,2500,0,0,128,0,0,,00:15:5d:00:ad:23,0,,1,2,0
2003:ffff:ffff:ff11::,00:01:00:01:30:2f:4a:3c:00:15:5d:00:ad:23,4000,1761047783,1,2500,2,0,64,0,0,,00:15:5d:00:ad:23,0,,1,2,0
```

*Secondary HA node*

```
address,duid,valid_lifetime,expire,subnet_id,pref_lifetime,lease_type,iaid,prefix_len,fqdn_fwd,fqdn_rev,hostname,hwaddr,state,user_context,hwtype,hwaddr_source,pool_id
2003:ffff:ffff:ff01::101,00:01:00:01:30:2f:4a:3c:00:15:5d:00:ad:23,4000,1761047783,1,2500,0,0,128,0,0,,00:15:5d:00:ad:23,0,,1,0,0
2003:ffff:ffff:ff11::,00:01:00:01:30:2f:4a:3c:00:15:5d:00:ad:23,4000,1761047783,1,2500,2,0,64,0,0,,00:15:5d:00:ad:23,0,,1,0,0
```

Example configuration:

```
"subnet6": [
[...]

PR #9306, opened 2025-10-22
dhcp/kea: Install prefix delegation routes targeting IA_NA address

Monviech:
Fixes: https://github.com/opnsense/core/issues/9135

This uses the same logic and reuses most code of the existing `prefixes.php` and  `prefixes.sh`.

I tested it and it works on my box that has a static delegated prefix from my ISP, the service also works after a reboot.

The lease files are synchronized between primary and secondary kea nodes in HA, meaning both will install the same static routes.

A static route can only be installed if there is an interface with an IPv6 address in the same prefix as the gateway address (IA_NA) of the IA_PD.

This means currently there is no support for LLA routes, as we do not want to assume anything. IA_NA and IA_PD are both in the lease file == zero assumptions.

Theoretically if one would use KEA hooks, there is an env variable with the LLA of the requesting client, which could theoretically be used for routing, if one would be bold enough to trust KEAs hooks.

### DIFF
--- a/src/etc/inc/plugins.inc.d/kea.inc
+++ b/src/etc/inc/plugins.inc.d/kea.inc
@@ -163,7 +163,7 @@ function kea_configure_do($verbose = false)
         if ($keaDhcpv6->isEnabled()) {
             mwexecf(
                 '/usr/sbin/daemon -m0 -f -p %s %s',
-                ['/var/run/kea-pd-routes.pid', '/usr/local/opnsense/scripts/dhcp/prefixeskea.sh']
+                ['/var/run/kea-pd-routes.pid', '/usr/local/opnsense/scripts/kea/prefixes.sh']
             );
         }
         service_log("done.\n", $verbose);


src/opnsense/scripts/kea/prefixes.php
```diff
@@ -0,0 +1,103 @@
...
+            $routes[$prefix] = $entry['address'];
+        }
+    }
+}
+
+/* expire all first */
+foreach (array_keys($routes) as $prefix) {
+    mwexecf('/sbin/route delete -inet6 %s', [$prefix], true);
+}
+
```
fichtner: the expire derived from the log. if we don't collect expiry we don't have to do a separate loop on the same data

src/opnsense/scripts/kea/prefixes.sh
```diff
@@ -0,0 +1,39 @@
+#!/bin/sh
+
+# Copyright (c) 2025 Deciso B.V.
```
fichtner, line 3: that's cheating ;)

src/opnsense/scripts/kea/prefixes.php
```diff
@@ -0,0 +1,103 @@
...
+require_once 'util.inc';
+
+$leases_file = '/var/db/kea/kea-leases6.csv';
+if (!file_exists($leases_file)) {
+    exit(1);
+}
+
+$duid_arr = [];
+$now = time();
+if (($fh = fopen($leases_file, 'r')) !== false) {
```
fichtner: SplFileObject was used to speed up the reading, so this may be slow again?

src/opnsense/scripts/kea/prefixes.php
```diff
@@ -0,0 +1,103 @@
...
+}
+
+/* expire all first */
+foreach (array_keys($routes) as $prefix) {
+    mwexecf('/sbin/route delete -inet6 %s', [$prefix], true);
+}
+
+/* active route apply */
+foreach ($routes as $prefix => $address) {
+    if (!empty($address)) {
```
fichtner: looking at the other script this can never be empty due to no expiry

src/opnsense/service/conf/actions.d/actions_kea.conf
```diff
@@ -34,3 +34,9 @@ command:/usr/local/opnsense/scripts/dhcp/get_kea_leases.py
 parameters:--proto inet6
 type:script_output
 message:list kea inet dhcp leases
+
+[update.prefixes]
+command:/usr/local/opnsense/scripts/dhcp/prefixeskea.php
```
fichtner: as mentioned I would prefer scripts/kea and keeping the copied file names the same as before

src/etc/inc/plugins.inc.d/kea.inc
```diff
@@ -157,7 +157,16 @@ function kea_configure_do($verbose = false)
...
+        if ($keaDhcpv6->isEnabled()) {
+            killbypid('/var/run/prefixeskea.pid');
+            mwexecf(
+                '/usr/sbin/daemon -m0 -f -p %s %s',
+                ['/var/run/prefixeskea.pid', '/usr/local/opnsense/scripts/dhcp/prefixeskea.sh']
+            );
+        }
         service_log("done.\n", $verbose);
+    } else {
+        killbypid('/var/run/prefixeskea.pid');
```
fichtner: this trailing else is wrong

it's better to always stop and then do a conditional start

src/etc/inc/plugins.inc.d/kea.inc
```diff
@@ -157,7 +157,16 @@ function kea_configure_do($verbose = false)
             $keaDhcpv6->generateConfig();
         }
         (new \OPNsense\Kea\KeaCtrlAgent())->generateConfig();
+        if ($keaDhcpv6->isEnabled()) {
+            killbypid('/var/run/prefixeskea.pid');
+            mwexecf(
+                '/usr/sbin/daemon -m0 -f -p %s %s',
+                ['/var/run/prefixeskea.pid', '/usr/local/opnsense/scripts/dhcp/prefixeskea.sh']
```
fichtner: `/var/run/kea-pd-routes.pid` is probably much nicer when looking at /var/run later on wondering what all of this is

src/opnsense/scripts/kea/prefixes.php
```diff
@@ -0,0 +1,99 @@
...
+        foreach ($entry['prefix'] as $prefix) {
+            /* new or reassigned takes priority */
+            $routes[$prefix] = $entry['address'];
+        }
+    }
+}
+
+/* delete and re-add routes */
+foreach ($routes as $prefix => $address) {
+    mwexecf('/sbin/route delete -inet6 %s', [$prefix], true);
```
fichtner, line 97: We could avoid the delete and add, but we’d need to check if it exists for semantics reasons in ifconfig. Maybe later, not for now.

Monviech:
@AdSchellevis I wait for your ok before merging this one.

AdSchellevis:
@Monviech I'm afraid the reality is a bit more complicated when it comes to kea. Ideally we would have a script reading the lease database and then tail the active one for changes. This also has the advantage of not flushing the routes all the time for all known PD's as we only see all at init or when they change.

The misery starts here with kea https://github.com/isc-projects/kea/blob/ef1f878f5272d/src/lib/dhcpsrv/memfile_lease_mgr.h#L1039-L1051

Since we now trigger a reload on file changes, we will remove and re-add all routes found in the lease file on every change, which I believe is also one of the issues in the isc version. This has the downside of constantly killing routes when clients sign-on for something else than a prefix (on any of the networks).

Conceptually, we need a single script that does this in my opinion:

```
-> Read kea database (which unfortunately isn't a single file), full, and keep reading for changes
  --> filter relevant records
     --> update route when found         
```

It has been some time ago that I wrote https://github.com/opnsense/core/blob/master/src/opnsense/scripts/dhcp/get_kea_leases.py, so I can't remember if after a full read we can always trust the `/var/db/kea/kea-leases6.csv` to receive the latest updates, but kind of expect it does.

Food for thought, let's think about this more next week before merging.

Monviech:
Do you think it would already help to write a cache file to /tmp so we can keep track of some state, and keep the "lease files (.1 .2 .x) have changed" logic lean?

We write routes that have been already installed into the /tmp/kea-pd-routes-cache.log or something, so we can compare the current reality with this cache? Since its in tmp every reboot would start with a full reinstall of all routes.

EDIT: Could also read netstat -rn for these routes?

fichtner:
System routing already reads route table before adding a route that already exists so I think that’s better. The bigger question though is if we can keep track of expired routes or not.

Monviech:
Theoretically I could outsource the lease file reading to:

configctl kea list leases6
That logic was already done in the https://github.com/opnsense/core/blob/master/src/opnsense/scripts/dhcp/get_kea_leases.py, and I can trust its output.

I can mdp5 the output of that command, and process the json, expire unix timestamp vs time.now reveals which leases are expired.

```
--> mdp5 configctl kea list leases6 in a loop (prefixes.sh)
  --> When changes are detected, process json output in (prefixes.php), use time to tag expired and valid routes.
     --> Read "configctl interface routes list" and compare with valid and expired routes
        --> Delete routes that exist and are expired
           --> Add routes that do not exist and are valid
```

AdSchellevis:
The disadvantage is that you always have to read the whole file, which is why some sort of tail would be better / less intrusive.

Monviech:
What I don't understand: if only tailing the lease file for changes, it would never "re-read" the expired leases in that file, since they will not change once they have been written (their unix timestamp is when they will expire).

It must read the full leases set for the decision which routes to delete.

Only new leases are appended after all:

https://github.com/isc-projects/kea/blob/ef1f878f5272d2f9c9d27b2dd592055e330ac887/src/lib/dhcpsrv/memfile_lease_mgr.h#L1055-L1056

Monviech:
Hmm wait only tailing might work. When a lease expires its also appended:

https://github.com/isc-projects/kea/blob/ef1f878f5272d2f9c9d27b2dd592055e330ac887/src/lib/dhcpsrv/memfile_lease_mgr.h#L59-L65

fichtner:
We're almost back at the infamous dhcpleases6 point in time that never worked reliably with listening for file changes.

A good way forward would be to add route reading like system routing already does (mainly to avoid clobbering default routes that are already set) and make the initial md5 more clever, maybe by grepping for relevant data only.

AdSchellevis:
Not sure what you mean with `dhcpleases6` (I assume this is about unbound+isc), but in order to collect changes reliably at scale, you do need to filter the relevant lease entries before processing anything. Keep in mind we're not trying to sync two services, only collect and fire events, which on service restart you can always reprocess. Tailing files is usually not an issue.

Monviech:
I mean if we only want to act on events, we could also circle back to this, it also would have the link local address:

https://github.com/opnsense/core/issues/9135#issuecomment-3425230064

`QUERY6_REMOTE_ADDR=fe80::215:5dff:fe00:ad24`

Which means I do not have to install the worse choice, the IA_NA targeting route. This saves a processing step.

fichtner:
I just mean for complexity and maintainability reasons we removed dhcpleases(6) daemons a few years ago. Daemons in Python are a possibility but I expect them to take a bit of maintenance over the course of their rollout. If the Kea hooks work that would be the simplest approach. The ISC PD routes prefixes.php is more or less what it always was, previously triggered by dhcpleases6 if I recall correctly.

AdSchellevis:
You might be able to use the hooks, but they very likely have the issue of "forgetting" things after transitions, which was the main reason to stay away from them. Having a shell script polling constantly or a python script is basically the same, the advantage of using the python script is only needing one simple entry point (in stead of two in the current approach).

Having the shell script tail and filter is also an option, but with the same argument against it, why not keep it simple with one script instead of two.

So far the experiences with kea are far from great, if we can guarantee safe transitions in all failover and failback scenarios, I'm fine with them as well. I just have a strong feeling we will spend a lot of time on something that will not work (as nobody in the real world seems to be using these in similar cases).

Monviech:
I think all arguments so far have pros and cons which is why making a decision is so hard.

A unified python script that is daemonized sounds good too. I think trying that is a good step forward.

Monviech:
About KEA events, do not work reliably for this usecase:

- When primary KEA responds, it emits events there, and if there is a failover (0 unacked clients + stopping primary service), the backup responds and events are emitted there.
- leases6_committed -> Only emitted on the current KEA instance that processed the lease
- lease6_expire -> Somehow that is emitted on both...

All in all we have to use a script that parses the lease files in some way and /forget/ about these hook events.

Issue #9296, opened 2025-10-17
remove "unassigned" dhcp option filter from dnsmasq_dhcp_options.py


mimugmail:
**Important notices**

Before you add a new report, we ask you kindly to acknowledge the following:

- [X] I have read the contributing guide lines at https://github.com/opnsense/core/blob/master/CONTRIBUTING.md
- [X] I am convinced that my issue is new after having checked both open and closed issues at https://github.com/opnsense/core/issues?q=is%3Aissue

**Is your feature request related to a problem? Please describe.**

We have some DELL Wyse clients using DHCP option 165, 166 and 199 which are stated as "unassigned in IANA list.
It seems that dnsmasq_dhcp_options.py is excluding them:

https://github.com/opnsense/core/blob/master/src/opnsense/scripts/dns/dnsmasq_dhcp_options.py#L49

Is there a specific reason to exclude them?

**Describe the solution you like**

(e.g. I would like an input field in the `/ui/firewall/alias` which would add .... to ....)

**Describe alternatives you considered**

A clear and concise description of any alternative solutions or features you considered.

**Additional context**

Add any other context or screenshots about the feature request here or links to relevant forum thread or similar

Monviech:
Essentially it does not matter to dnsmasq if an option does exist or not, it will happily serve any option with any option string. So I wouldn't mind showing them all.

Before changing anything though lets wait for more than one opinion.

cc @AdSchellevis

AdSchellevis:
be careful what you wish for? how much options are we going to add when removing the filter? (large lists may confuse others)

Monviech:
its 256 options in total if there is no filter.

https://github.com/opnsense/core/blob/master/contrib/iana/dhcpv4-options.csv

Maybe optgroups could sort them better?

PR #9309, opened 2025-10-22
dhcp/dnsmasq: Add optgroup support to option fields, expose all dhcpv4 options

Monviech:
Fixes: https://github.com/opnsense/core/issues/9296

- Common is what dnsmasq exposes in the help text
- Assigned, Unassigned is deduplicated what IANA has left to offer
This ensures we do not assume anything, thus we will not have to adjust the script if the future changes in some way

DHCPv4 options are all exposed now (256), DHCPv6 options are still filtered since there is a silly amount of them (65535). Lets just hope nobody needs DHCPv6 Option 65432 or something.

JsonKeyValueStoreField now supports OPTGROUPS.

### DIFF
--- a/src/opnsense/mvc/app/models/OPNsense/Dnsmasq/Dnsmasq.xml
+++ b/src/opnsense/mvc/app/models/OPNsense/Dnsmasq/Dnsmasq.xml
@@ -269,7 +269,7 @@
             <option6 type="JsonKeyValueStoreField">
                 <ConfigdPopulateAct>dnsmasq list dhcp_options6</ConfigdPopulateAct>
                 <OptionValues>
-                     <!-- mock option groups returned by JSON -->
+                    <!-- mock option groups returned by JSON -->
                     <optgroup1>Common</optgroup1>
                     <optgroup2>Assigned</optgroup2>
                     <optgroup3>Unassigned</optgroup3>


Monviech:
> I like the idea. Just a simple question: how and when are "Common" etc translated? ;)

You found the crux of it. Got any idea how to fix this?

(though its not like the dns option names are translated either)

Monviech:
@fichtner I tried something out that seems to work for translations, can you check if thats what you expect, or if there is a different strategy that is preferred?

fichtner:
not sure how @AdSchellevis feels about this but looks like extending the existing JSON field to translate optgroups is more useful in the long run. If optgroups are not being used in configuration we could argue these are merely labels, not more. But you'd also have to consider that current optgroups may also lack a translation? Unsure about it.

AdSchellevis:
If we expect option groups to usually be language specific (sounds logical), it would mean that the json field type should at least offer a hook to offer the translations, when always statically typed, it would be possible to push them into the xml for the translation script to pickup. 

The options themselves are usually (often) language agnostic.

fichtner:
@AdSchellevis sounds like a 'yes'. @Monviech and me will try to extend the field tomorrow before advancing this one

src/opnsense/mvc/app/models/OPNsense/Dnsmasq/Dnsmasq.xml
```diff
@@ -257,10 +257,10 @@
                 <Default>set</Default>
                 <Required>Y</Required>
             </type>
-            <option type="JsonKeyValueStoreField">
+            <option type=".\DhcpOptionField">
```
fichtner:
```suggestion
            <option type="JsonKeyValueStoreField">
            <OptionValues>
                <!-- mock option groups returned by JSON -->
                <optgroup1>Common</optgroup1>
                <optgroup2>Assigned</optgroup2>
                <optgroup3>Unassigned </optgroup3>
            </OptionValues>
```

something like this to be able to extract the values for translation

src/opnsense/www/js/opnsense.js
```diff
@@ -184,10 +184,13 @@ function setFormData(parent,data) {
                                 optgroups[optgroup].push(opt);
                             });
                         }
+                        // Render ungrouped items first (so None selection is first)
```
fichtner: Not entirely sure about this yet, looks ripe for regressions

src/opnsense/scripts/dns/dnsmasq_dhcp_options.py
```diff
@@ -41,20 +41,47 @@
...
+        recommended[key] = val
+        # Deduplicate from other groups
+        assigned.pop(key, None)
+        unassigned.pop(key, None)
 
-print(json.dumps(result))
+print(json.dumps({
+    "common": recommended,        # Options directly supported by dnsmasq (help output)
+    "assigned": assigned,
+    "unassigned": unassigned
```
fichtner:
```suggestion
    "Common": recommended,        # Options directly supported by dnsmasq (help output)
    "Assigned": assigned,
    "Unassigned": unassigned
```

src/opnsense/mvc/app/models/OPNsense/Base/FieldTypes/JsonKeyValueStoreField.php
```diff
@@ -178,7 +178,24 @@ protected function actionPostLoadingEvent()
...
+            foreach ($data as $key => $value) {
+                if (!is_array($value)) {
+                    // flat form: { "id": "Label" }
+                    $this->internalOptionList[$key] = $value;
+                } else {
+                    // grouped form: { "Group": { "id": "Label", ... } }
+                    foreach ($value as $subkey => $subval) {
+                        $this->internalOptionList[$subkey] = [
+                            'value' => $subval,
+                            'optgroup' => $key
```
fichtner:
```suggestion
                            'optgroup' => gettext($key),
                            'value' => $subval,
```

src/opnsense/www/js/opnsense.js
Monviech: I removed it, None last isnt that bad (only small annoyance) better than regressions

src/opnsense/mvc/app/models/OPNsense/Dnsmasq/Dnsmasq.xml
```diff
@@ -259,9 +259,20 @@
...
+                <OptionValues>
+                    <!-- mock option groups returned by JSON -->
+                    <optgroup1>Common</optgroup1>
+                    <optgroup2>Assigned</optgroup2>
+                    <optgroup3>Unassigned</optgroup3>
+                </OptionValues>
             </option>
             <option6 type="JsonKeyValueStoreField">
                 <ConfigdPopulateAct>dnsmasq list dhcp_options6</ConfigdPopulateAct>
+                <OptionValues>
```
fichtner, line 271:
```suggestion
                <OptionValues>
                     <!-- mock option groups returned by JSON -->
```

fichtner (approved):
LGTM

PR #9321, opened 2025-10-25
fix squid blacklist categories not showing

raoulblt:
JsonKeyValueStoreField SourceField not being set from BaseModel.php while parent node not set

Resolve squid black categories list not showing:

- ref opnsense/plugins/issues/4392
- ref opnsense/plugins/issues/4929

### DIFF
--- a/src/opnsense/mvc/app/models/OPNsense/Base/BaseModel.php
+++ b/src/opnsense/mvc/app/models/OPNsense/Base/BaseModel.php
@@ -159,7 +159,7 @@ abstract class BaseModel
      * @throws ModelException when unable to parse field type
      * @throws ReflectionException when unable to create class
      */
-    private function getFieldObject($xmlNode, $thisModelPath, $new_ref)
+    private function getFieldObject($xmlNode, $thisModelPath, $new_ref, $internal_data)
     {
         if (self::$internalCacheReflectionClasses === null) {
             self::$internalCacheReflectionClasses = [];
@@ -238,6 +238,7 @@ abstract class BaseModel
             $fieldObject->setInternalIsVolatile();
         }
         if (!$fieldObject->isContainer()) {
+            $internal_data->addChildNode($xmlNode->getName(), $fieldObject);
             foreach ($field_methods as $method_name => $payload) {
                 $fieldObject->$method_name($payload);
             }
@@ -266,12 +267,11 @@ abstract class BaseModel
             $thisModelPath = empty($model_path) ? get_class($this) . "." . $tagName : $model_path . '.' . $tagName;
             // generate full object name ( section.section.field syntax ) and create new Field
             $new_ref = $internal_data->__reference == "" ? $tagName : $internal_data->__reference . "." . $tagName;
-            $fieldObject = $this->getFieldObject($xmlNode, $thisModelPath, $new_ref);
+            $fieldObject = $this->getFieldObject($xmlNode, $thisModelPath, $new_ref, $internal_data);
 
             // now add content to this model (recursive)
             $config_section_data = $config_data?->$tagName ?? null;
             if (!$fieldObject->isContainer()) {
-                $internal_data->addChildNode($tagName, $fieldObject);
                 if ($config_section_data !== null) {
                     /* do not set initial data, default will take care */
                     $fieldObject->setValue($config_section_data);


AdSchellevis:
This can't be the proper fix, since no other consumer of `JsonKeyValueStoreField` uses `SourceFile` + `SourceField`, there might be an issue in the field type when these specific construct is used, but this doesn't warrant a change that influences all fieldtypes. 

There are other (custom) field types back referencing the row, which means it should be possible to keep the same behavior.

If you search for `getParentNode` you will likely find some examples.

oulra:
If there is a field type issue I have difficulty to see how this can be working (for blacklist), since the only modification here is to call `addChildNode` before the specific construct. Any suggestion to look further?

AdSchellevis:
I'm not entirely sure where the issue is (not using squid myself), but if the initialization in the setter has an issue, moving the logic might help.

https://github.com/opnsense/core/blob/5c51c5a7c06072604702d7aa781be88e34ca0b1d/src/opnsense/mvc/app/models/OPNsense/Base/FieldTypes/JsonKeyValueStoreField.php#L80

First step is figuring out if `setSourceField` is indeed populated with the wrong data, which can be done by building a small script which only loads the model and output the data at that spot (or a debugger breakpoint).

oulra:
The issue is that blacklist categories are not shown while index file `/usr/local/etc/squid/acl/UT1.index` is well generated :

```
root@opnsense-test:~ # ls -la /usr/local/etc/squid/acl/
total 56638
drwxr-x---  2 root squid         4 Oct 26 15:19 .
drwxr-xr-x  8 root squid        23 Oct 26 09:02 ..
-rw-r-----  1 root squid 139610293 Oct 26 15:19 UT1
-rw-r-----  1 root squid      1802 Oct 26 15:19 UT1.index
```

BL_categories_not_ok

It's seems that `setSourceField` is not populated because the condition line 79 is not checked :

A simple test by hardcoding `UT1` :

```php
 77     public function setSourceField($value)                                      
 78     {                                                                           
 79         if (isset($this->internalParentNode->$value)) {                         
 80             $this->internalSourceField = basename($this->internalParentNode->$value);
 81             $this->internalSourceField = 'UT1';                                 
 82         }                                                                                                        
 83     }
```
Edit blacklist again : no change, blacklist categories are not displayed.

Now setting `SourceField` with `UT1` out of condition :

```php
 77     public function setSourceField($value)                                      
 78     {                                                                           
 79         if (isset($this->internalParentNode->$value)) {                         
 80             $this->internalSourceField = basename($this->internalParentNode->$value);
 81         }                                                                       
 82         $this->internalSourceField = 'UT1';                                     
 83     }
```

Edit blacklist again : categories are displayed :

BL_categories_ok

So `$this->internalParentNode->$value` is not set and I'm trying to debug further.

Thank you for your advices I will try to make a debugger breakpoint.

oulra:
> This can't be the proper fix, since no other consumer of `JsonKeyValueStoreField` uses `SourceFile` + `SourceField`, there might be an issue in the field type when these specific construct is used, but this doesn't warrant a change that influences all fieldtypes.
> 
> There are other (custom) field types back referencing the row, which means it should be possible to keep the same behavior.
> 
> If you search for `getParentNode` you will likely find some examples.

`JsonKeyValueStoreField` seems to be an exception for back referencing his parent node at setters level. It could be a regression starting from fec3f0b (`addChildNode` was called before setters). @AdSchellevis could you consider that?

AdSchellevis:
@raoulblt it does sound logical that the setter can't reliably back reference, which means a fix `JsonKeyValueStoreField` might be required indeed.

oulra:
Can you fix it?

whbingham:
I am running opnsense with Squid Web Proxy.  Upgraded to 25.7.6 with os-squid 1.4 and Squid Web Proxy would not work, kept crashing.  After modifying the latest BaseModel.php file with the same modifications as suggested by @raoulblt, it does seem to fix the issues with Squid Remote ACL categories list and it now appears to work normally without crashing.

AdSchellevis:
@raoulblt https://github.com/opnsense/core/commit/99ee648132e9a114477d0bf78fc3b4a0d46b5e61 should do the trick.

oulra:
Thank you @AdSchellevis it's working.

whbingham:
Thank you @raoulblt and @AdSchellevis.  I reverted the changes suggested by @raoulblt and modified the JsonKeyValueStoreField.php and all is working correctly with Squid web proxy remote access control lists under Opnsense 25.7.6 and os-squid 1.4.

PR #9327, opened 2025-10-29
Update interfaces.php

Neozlag:
We found that if we change the MTU of an interface (ex 2000), it works, but if, after that, we remove the MTU and leave the field empty, it does nothing and it left the old value (in our case, 2000). To change that, I had a control to check if MTU is empty, and in that case, we set the value to 1500, the default value

### DIFF
--- a/src/www/interfaces.php
+++ b/src/www/interfaces.php
@@ -844,7 +844,11 @@ if ($_SERVER['REQUEST_METHOD'] === 'GET') {
         if (!empty($pconfig['spoofmac']) && !is_macaddr($pconfig['spoofmac'])) {
             $input_errors[] = gettext("A valid MAC address must be specified.");
         }
-
+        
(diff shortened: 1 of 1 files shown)


fichtner:
Although that is a bit of an edge case it's impossible to get the system to get back to the "default" MTU state. A reboot would do it, but this PR would never bring back the system default again.


Cheers,
Franco

Neozlag:
You mean if the default value of the system isn't 1500 ?

Regards,

fichtner:
Yes, for HW ethernet NICs that may be true, but not for a lot of other network devices.


Cheers,
Franco

Neozlag:
Ok thanks, my bad, I didn't see that point.

Regards,

fichtner:
No problem. A note in the docs may make sense but the question is who will really see it when it matters. It's a difficult edge cause as said.

Close this one?

PR #9329, opened 2025-10-30
Allow multiple ICMP6 type to be selected in firewall rules

pfoo:
This webui patch allows to create firewall rules for IPv6-ICMP with multiples ICMP6 type selected. This was already working for IPv4 ICMP.


AdSchellevis:
maybe better to implement this in the new rule ui when missing, we try to freeze the legacy code as much as possible.

AdSchellevis:
implemented for the new interface in https://github.com/opnsense/core/commit/5f19a2617d805ddc638735e5cb4546de806be802

Issue #9155, opened 2025-08-27
isc-dhcp: move to plugins and change DHCPv6 default to Dnsmasq (no more automatic PD support)


fichtner:
**Important notices**

Before you add a new report, we ask you kindly to acknowledge the following:

- [x] I have read the contributing guide lines at https://github.com/opnsense/core/blob/master/CONTRIBUTING.md
- [x] I am convinced that my issue is new after having checked both open and closed issues at https://github.com/opnsense/core/issues?q=is%3Aissue

**Is your feature request related to a problem? Please describe.**

Since ISC DHCP is EOL it needs to be moved to plugins. Kea and Dnsmasq both fill the void left by it.

**Describe the solution you like**

Move ISC DHCP to plugins and change the automatic IPv6 setup to Dnsmasq (factory reset and wizard default).

**Describe alternatives you considered**

Use Kea, but it does not offer dynamic prefix tracking.

**Additional context**

N/A

PR #9335, opened 2025-10-31
dhcp: Move kea and dnsmasq scripts out of dhcp folder

Monviech:
Needed for preparation of turning ISC DHCP into a plugin.

Reference: https://github.com/opnsense/core/issues/9155

### DIFF
--- a/src/opnsense/service/conf/actions.d/actions_kea.conf
+++ b/src/opnsense/service/conf/actions.d/actions_kea.conf
@@ -24,7 +24,7 @@ type:script_output
 message:get kea daemon status
 
 [list.leases4]
-command:/usr/local/opnsense/scripts/dkea/get_kea_leases.py
+command:/usr/local/opnsense/scripts/kea/get_kea_leases.py
 parameters:--proto inet
 type:script_output
 message:list kea inet dhcp leases


plist
```diff
@@ -1087,17 +1087,15 @@
 /usr/local/opnsense/scripts/captiveportal/strip_template.py
 /usr/local/opnsense/scripts/dhcp/cleanup_leases4.php
 /usr/local/opnsense/scripts/dhcp/cleanup_leases6.php
-/usr/local/opnsense/scripts/dhcp/dnsmasq_watcher.py
```
fichtner, line 1090: Unsure about this one. It’s only needed when isc is used so theoretically it can move to plugin.

Monviech: Okay skipped that one.

src/opnsense/service/conf/actions.d/actions_kea.conf
```diff
@@ -24,13 +24,13 @@ type:script_output
 message:get kea daemon status
 
 [list.leases4]
-command:/usr/local/opnsense/scripts/dhcp/get_kea_leases.py
+command:/usr/local/opnsense/scripts/dkea/get_kea_leases.py
```
fichtner:
```suggestion
command:/usr/local/opnsense/scripts/kea/get_kea_leases.py
```

Issue #9149, opened 2025-08-27
firewall/automation: Rename feature and place it in a more prominent spot


Monviech:
**Important notices**

Before you add a new report, we ask you kindly to acknowledge the following:

- [x] I have read the contributing guide lines at https://github.com/opnsense/core/blob/master/CONTRIBUTING.md
- [x] I am convinced that my issue is new after having checked both open and closed issues at https://github.com/opnsense/core/issues?q=is%3Aissue

**Is your feature request related to a problem? Please describe.**

The new filter UI page should be placed in a spot where users are more eager to find and use it over the traditional rules view.

**Describe the solution you like**

Place it somewhere nice in menu.xml
Rename it

PR #9344, opened 2025-11-03
firewall/automation: Move menu items to firewall/rules/filter and firewall/nat/sourcenat

Monviech:
Fixes: https://github.com/opnsense/core/issues/9149

image

### DIFF
--- a/src/opnsense/mvc/app/models/OPNsense/Firewall/ACL/ACL.xml
+++ b/src/opnsense/mvc/app/models/OPNsense/Firewall/ACL/ACL.xml
@@ -1,13 +1,13 @@
 <acl>
     <page-filter-api>
-        <name>Firewall: Automation: Filter</name>
+        <name>Firewall: Rules [new]</name>
         <patterns>
             <pattern>ui/firewall/filter/*</pattern>
             <pattern>api/firewall/filter/*</pattern>
         </patterns>
     </page-filter-api>
     <page-filter-snat-api>
-        <name>Firewall: Automation: Source NAT</name>
+        <name>Firewall: NAT: Source NAT</name>
         <patterns>
             <pattern>ui/firewall/source_nat/*</pattern>
             <pattern>api/firewall/source_nat/*</pattern>


Monviech:
Oops I misclicked on copilot...

src/opnsense/mvc/app/models/OPNsense/Firewall/Menu/Menu.xml
```diff
@@ -1,17 +1,17 @@
 <menu>
     <Firewall>
-        <Automation cssClass="fa fa-gear">
-            <Filter order="50" url="/ui/firewall/filter/">
+        <Rules>
+            <Filter order="0" VisibleName="Filter [new]" url="/ui/firewall/filter/">
```
fichtner: I think this already starts at 0, so making a bit of room over there is better:

https://github.com/opnsense/core/blob/5256bed259b194c5f3d16f00eb215fbc78bb11e6/src/opnsense/mvc/app/models/OPNsense/Base/Menu/MenuSystem.php#L333

src/opnsense/mvc/app/models/OPNsense/Firewall/Menu/Menu.xml
```diff
@@ -1,17 +1,17 @@
...
-            <SourceNat order="100" VisibleName="Source NAT" url="/ui/firewall/source_nat/">
-                <FilterRef url="/ui/firewall/source_nat#*" visibility="hidden"/>
-            </SourceNat>
-        </Automation>
+        </Rules>
         <NAT>
             <OneToOne order="200" VisibleName="One-to-One" url="/ui/firewall/one_to_one/">
                 <FilterRef url="/ui/firewall/one_to_one#*" visibility="hidden"/>
             </OneToOne>
+            <SourceNat order="300" VisibleName="Source NAT" url="/ui/firewall/source_nat/">
```
fichtner: Outbound is 300 ;)

https://github.com/opnsense/core/blob/5256bed259b194c5f3d16f00eb215fbc78bb11e6/src/opnsense/mvc/app/models/OPNsense/Core/Menu/Menu.xml#L98

fichtner:
This also needs ACL adjustment.

Also "Filter" seems a bit blant and perhaps confusing: We could use Global, Unified, Aggregate, Combined, Consolidated. Global is perhaps too close to Floating, "ed" sounds a little retro but can work, aggregate is a bit edgy. Another batch: cluster, bundle or collection, accumulated or full.

Just some ideas for discussion.

mimugmail:
If automation will replace the whole legacy rules in some time, Global would make sense as Floating would be replaced too? 

( but please don't touch the api endpoints :) )

fichtner:
Eventually the whole GUI will be under Firewall: Rules without fancy names (floating and interface specific ones will all be gone). But no actual migration strategy has been discussed and it's also not a focus for 26.1.

mimugmail:
I didn't follow the latest changes to the code, but I had in mind that from priority perspective, first there is floating, then automation, then groups, then interface rules. If still correct, maybe sort Filter [new] below Floating?

fichtner:
To get people to use it the first entry would probably be the best choice given the fact that it can be to be the standard (and only) GUI entry point in a few major versions.

Monviech:
I would kinda go with

Firewall - Rules [new]
Firewall - Rules - Floating/Interface etc...

fichtner:
Not bad, although it asserts a bit of pressure on us to leave the [new] in there for quite a long time ;)

Monviech:
Firewall Reloaded - The final frontier.

fichtner:
Merged, thanks!

PR #9351, opened 2025-11-04
Firewall: Fix delete_selected firewall states

AlexanderS:
Without the datakey option the delete_selected command will try to use the non-existing property "uuid" of the selected rows.


AdSchellevis:
@AlexanderS thanks! looks like a regression introduced with the new grid system

PR #9353, opened 2025-11-04
Allow inline # comments in all alias types (network/host/port)

I-G-1-1:
Summary:
This PR makes it possible to use inline comments (with '#') on every alias type (host/network/port) both in the WebUI and during runtime rule generation.

Modified:
- src/opnsense/mvc/app/models/OPNsense/Firewall/FieldTypes/AliasContentField.php : WebUI validation
- src/opnsense/scripts/filter/lib/alias/__init__.py : backend Python, ensures addresses/networks ignore comments
- src/opnsense/mvc/app/library/OPNsense/Firewall/Util.php : PHP alias expansion
- src/etc/inc/filter.lib.inc : legacy/compat port alias expansion, strips comments from port lines

All patches use preg_replace('/\s*#.*$/', '', $line) logic per line, and ignore empty/commented lines.

How to test:
1. Create an alias (any type) and add lines like '192.168.1.1 # test' or '1234 # http'.
2. Save, Apply, and verify ports/addresses/networks are present in live rules and diagnostics, but comment text is ignored.

Fixes longstanding community requests for better in-place documentation of firewall configs.

### DIFF
--- a/src/opnsense/mvc/app/models/OPNsense/Firewall/FieldTypes/AliasContentField.php
+++ b/src/opnsense/mvc/app/models/OPNsense/Firewall/FieldTypes/AliasContentField.php
@@ -95,11 +95,10 @@ class AliasContentField extends BaseField
     private function getItems($data)
     {
         foreach (explode($this->separatorchar, $data) as $row) {
-            // Remove all after first #
-            if (strpos($row, '#') !== false) {
-                $row = preg_replace('/\s*#.*$/', '', $row);
-            }
             foreach (explode(',', $row) as $value) {
+                if (strpos($value, '#') !== false) {
+                    $value = preg_replace('/\s*#.*$/', '', $value);
+                }
                 $value = trim($value);
                 if ($value !== '') {
                     yield $value;
--- a/src/etc/inc/filter.lib.inc
+++ b/src/etc/inc/filter.lib.inc
@@ -151,7 +151,6 @@ function filter_core_get_port_alias($aliasname, $aliases = [], $aliasObject = nu
     $aliasObject = $aliasObject == null ? (new \OPNsense\Firewall\Alias(true)) : $aliasObject;
     foreach ($aliasObject->aliasIterator() as $aliased) {
         if ($aliasname == $aliased['name'] && $aliased['type'] == 'port' && !empty($aliased['enabled'])) {
-
             foreach ($aliased['content'] as $address) {
             /* Original
                 if (in_array($address, $all_aliases)) {
@@ -166,11 +165,11 @@ function filter_core_get_port_alias($aliasname, $aliases = [], $aliasObject = nu
                     $response[] = $address;
                 }
                 */
-                // Remove all after first #
-                if (strpos($address, '#') !== false) {
-                    $address = preg_replace('/\s*#.*$/', '', $address);
-                }
+                // Split on comma and Remove all after first #
                 foreach (explode(',', $address) as $value) {
+                    if (strpos($value, '#') !== false) {
+                        $value = preg_replace('/\s*#.*$/', '', $value);
+                    }
                     $value = trim($value);
                     if ($value === '') {
                         continue;
@@ -187,7 +186,7 @@ function filter_core_get_port_alias($aliasname, $aliases = [], $aliasObject = nu
                         $response[] = $value;
                     }
                 }
-               // Remove all after first #
+               // Split on comma and Remove all after first #
             }
         }
     }
--- a/src/opnsense/mvc/app/library/OPNsense/Firewall/Util.php
+++ b/src/opnsense/mvc/app/library/OPNsense/Firewall/Util.php
@@ -286,8 +286,8 @@ class Util
         foreach (self::$aliasObject->aliasIterator() as $node) {
             if (!empty($name) && (string)$node['name'] == $name && $node['type'] == 'port') {
                 $aliases[] = $name;
-                /* Original
                 foreach ($node['content'] as $address) {
+                /* Original
                     if (Util::isAlias($address)) {
                         if (!in_array($address, $aliases)) {
                             foreach (Util::getPortAlias($address, $aliases) as $port) {
@@ -301,12 +301,11 @@ class Util
                     }
                 }
                 */
-                foreach ($node['content'] as $address) {
-                    // Remove all after first #
-                    if (strpos($address, '#') !== false) {
-                        $address = preg_replace('/\s*#.*$/', '', $address);
-                    }
+                    // Split on comma and Remove all after first #
                     foreach (explode(',', $address) as $value) {
+                        if (strpos($value, '#') !== false) {
+                            $value = preg_replace('/\s*#.*$/', '', $value);
+                        }
                         $value = trim($value);
                         if ($value === '') {
                             continue;
@@ -323,8 +322,8 @@ class Util
                             $result[] = $value;
                         }
                     }
+                    // /Split on comma and Remove all after first #
                 }
-                // /Remove all after first #
             }
         }
 
--- a/src/opnsense/scripts/filter/lib/alias/__init__.py
+++ b/src/opnsense/scripts/filter/lib/alias/__init__.py
@@ -193,16 +193,14 @@ class Alias(object):
                     address_parser = self.get_parser()
                     if address_parser:
                         for item in self.items():
-                            # Remove all after first #
-                            if '#' in item:
-                                item = item.split('#', 1)[0]
+                            # Split on comma and Remove all after first #
                             for value in item.split(','):
+                                if '#' in value:
+                                    value = value.split('#', 1)[0]
                                 value = value.strip()
                                 if not value:
                                     continue
-                                for address in address_parser.iter_addresses(value):
-                                    self._resolve_content.add(address)
-                            # /Remove all after first #
+                            # /Split on comma and Remove all after first #
                             for address in address_parser.iter_addresses(item):
                                 self._resolve_content.add(address)
                         # resolve hostnames (async) if there are any in the collected set


AdSchellevis:
I don't expect we will merge this to be honest, we removed similar input styles years ago for good reasons, aliases can be nested easily to document separate items.

I-G-1-1:
I know, I already saw that also on the forum ([here](https://forum.opnsense.org/index.php?topic=24037.msg167254#msg167254) for example) every time someone ask for the ability to comment on each line of aliases the solution proposed it's always "simply use nested aliases".
But I feel that this approach is extremely limiting and for this reason, although I don't know much about coding, I spent the afternoon creating and testing on my installation this PR in the hope of convincing the maintainers of this project to include this function.

> aliases can be nested easily to document separate items

Sorry, but I think that "easily" it's a bit subjective in this case. Nested aliases method really don't scale well at all.

Here is an example: I need to nat 50 different random ports for 50 services on different servers.

Using the `nested aliases` method:
1) on Firewall: Aliases click +
2) select Type Ports
3) add the port in Content
4) add in the Description the service and the server to which the port refers
5) click save
Now I just created the first alias for port 1 on server 1.
I'm at 1/50 of my journey, now I need to **make this same things 50x more times**. 
After this _time consuming task_ I can finally nest these 50 aliases.
6) create an alias that contains all these 50 aliases.

At this point I just completed 50 ports for 50 services only on server 1

Now start all over again for server 2 that has same services on different ports so you need to make 50 more aliases to document what service and server these port refers to and you cannot re-use the aliases from server 1 because equal services uses different ports on the 2 server.
(3*50*2= 300 clicks)

Using the `inline # comments` method:
1) on Firewall: Aliases click +
2) select Type Ports
3) in Content change to text
4) add 50 lines with `port_number #comment`
5) save
Done (3 clicks)

I came from OPNsense 18 that had this user interface with the ability of commenting each line:
OPNsense 18

I understand that for some technical reason this kind of data input has been abandoned in favor of one line data.
I'm ok with this "new" interface also because the new "text" input method let me simply copy/paste port, ip address or network. But the inability to document each line is really a draw back for my use case.
OPNsense 25

Sorry but I don't understand this rigidity and why we cannot have a piece of software that can accomodate different use cases of different user. 
Please try to put yourself in the workflow of others who probably do things differently from you, and don't restrict the potential of this awesome software.

I want also to mention that after 100+ aliases to create nested aliases the drop menu with its 5 lines becomes really clogged and hard to consult (yes I know I can use search in the drop menu, but this imply that someone know what to search that's different from have a list under the eyes).
OPNsense 25 - 2

AdSchellevis:
If the problem is bulk input, changing the data format is still not the right direction. Data entry is just not a data model issue. (p.s. try to dump a comma in your comment for fun)

I-G-1-1:
The problem is a dull and obtuse webUI/backend that doesn't adapt to different workflow and instead of being resilient and give the user the flexibility he needs it forces the user to make repetitive tasks and waste time.
It remembers me the stupid webUI of the cheap ISP Router/Wifi access point that doesn't allow you to change anything. That same cheap device that pushes people to buy a mini pc and install OPNsense (or equivalent).

Please explain what are the technical issue that avoid the adoption of this PR instead of philosophizing that data input problems are not solved by changing the data format or data model.

Neither the style "I perfectly know how you should use my software, I don't want anyone to use it in a way that deviates from my vision" is the right direction.

> try to dump a comma in your comment for fun

As I said I'm not a programmer, so I trusted your input to try comma "for fun" expecting some kind of issue like the ability to inject and let OPNsense run malicious code as consequence of the changes I made. Instead it seems nothing bad happens when I put commas (or '):
OPNsense 25 - comma 01
 
OPNsense 25 - comma 02

src/opnsense/mvc/app/models/OPNsense/Firewall/FieldTypes/AliasContentField.php
```diff
@@ -81,12 +81,30 @@ public function getNodeData()
...
         }
     }
+    */
+
+    private function getItems($data)
+    {
+        foreach (explode($this->separatorchar, $data) as $value) {
+            // Strip inline comments starting with #
+            if (strpos($value, '#') !== false) {
+                $value = preg_replace('/\s*#.*$/', '', $value);
```
fichtner: this alone hinges on the fact that `$separatorchar` is not changed

in all fairness it would make more sense to provide a separate alias type that handles this cleanly instead of wedging it into the existing one, because this will lead to regressions

I-G-1-1: Thanks for pointing out this issue. I'm not a programmer and I don't have the ability to make the changes you suggested, the best I could do is let me help by Copilot. 
I really hoped that someone with more experience inspired by this PR could pick the idea of allowing inline # comments and code it in a more professional way, knowing exactly what he is doing an making it secure and compatible with the current OPNsense code.

I-G-1-1:
Tried to change the logic to avoid issue with "," commas.

24455 port should not appear as it is after #commenting of port 24454.

Or do you expect this to work? I thought that on original OPNsense only a port for line was allowed in "text" mode.

OPNsense 25 - comma 03

OPNsense 25 - comma 04

as result:

OPNsense 25 - comma 05

Monviech:
From a different perspective this looks like some kind of Macro actions are missing that would turn repetitive actions into one automatic flow.

E.g. "Create a nested alias with these 1000 ports".

Though if you look at Macros in other software (e.g. excel) they are programmed via VBS or another script language.

The aliases do have an API endpoint, so you could automate your repetitive workflows via a script language as well.

I-G-1-1:
Thanks for the answer.

> E.g. "Create a nested alias with these 1000 ports".

then if you want to delete/edit all these aliases? what I'm supposed to do? write a Macro every time I need to make some bulk change? And the clogged drop menu with these 1000 aliases?
Only because this project cannot overcome the "no comment on aliases dogma" a simply copy/paste workflow becomes an endless loop?

I could make a file excel with the ports on a column and the description in another column and copy/paste the first column in an alias in OPNsense. But this way the documentation of my ports is only on my pc and not in the OPNsense machine. This approach wants me to be able to access that file, and I cannot do any change on the fly without that excel. And when someone else will need to make some changes will be hit by a wall of not documented ports.

> The aliases do have an API endpoint, so you could automate your repetitive workflows via a script language as well.

I could also configure all OPNsense using the shell and some script but would be time consuming and with a learning curve, so I'd prefer not to do that and rely on a flexible webUI. Like in this case, I would prefer to makes these changes on the webUI.

Monviech:
I'm curious why are are not using a port range if this is not a purely theoretical problem.

image

Or if you want to skip one port:

image

I-G-1-1:
> why are are not using a port range

each port is a different service, if I use a range how can I document in the alias description to wich service a specific port of the port range refers to? 
I could use the description field whit long long string like this "service1 - service 2 - service 3 - service n°", but it seem to me to be not well human readable or intelligible.

Monviech:
But this work has already been done by IANA:

https://www.iana.org/assignments/service-names-port-numbers/service-names-port-numbers.xhtml?&page=1

I-G-1-1:
> But this work has already been done by IANA

But if I want to use random ports on a service instead of the one proposed by IANA for example to make it harder for automated bot/scripts to check what service I'm running? Like for example openVPN not on the 1124 port..

However the "50 different random ports for 50 services on different servers" was only one of the possible example of why someone would want to comment a bunch of entries in the alias "content" field instead of having to nest aliases.

I-G-1-1:
Changed again the logic to support multiple ports with comments for each port separated by "," comma.

now aliases commenting can handle:
- port1, port2 #comment
- port1 #comment1, port2 #comment2

to make this possible now it's **NOT** allowed:
- port1 #comment, with, commas

OPNsense 25 - comma v3 1

as result:
OPNsense 25 - comma v3 2

As expected behaviour you cannot use commas in the comment:
OPNsense 25 - comma v3 3

OPNsense 25 - comma v3 4

Monviech:
It's unlikely this will make it in because it's still a validation nightmare with regression potential and introduced technical debt. Also the importance looks more like "affects only me".

I-G-1-1:
It's not true, it doesn't affect only me. I found a lot of people that in the years asked on the OPNsense forum or on github the ability to comment the aliases, but from the maintainers of this project they got only: "use nested aliases instead".

This changes are less than 10 lines of code for each file for 4 files. If there is a regression you can simply go back to the previous code, I'm pretty sure github allows that.
No one even tested this PR and you already speak of "validation nightmare" "potential regression" and "technical debt".
Simply you don't want.

At least can this PR become some kind of plug-in?

fichtner:
Good question.

> At least can this PR become some kind of plug-in?

Let me answer with what I already wrote.

> in all fairness it would make more sense to provide a separate alias type that handles this cleanly instead of wedging it into the existing one, because this will lead to regressions

I-G-1-1:
as I already told the context was: 

> I'm not a programmer and I don't have the ability to make the changes you suggested, the best I could do is let me help by Copilot.
> I really hoped that someone with more experience inspired by this PR could pick the idea of allowing inline # comments and code it in a more professional way, knowing exactly what he is doing an making it secure and compatible with the current OPNsense code.

Tried to make a plugin: https://github.com/I-G-1-1/opnsense_plugins/tree/master/security/advancedalias
In 2 days I only got a menu in firewall that call a page that doesn't work.
It's beyond my abilities to make a functional plugin.

AdSchellevis:
discussed this with the core team, we won't implement this.

I-G-1-1:
Yes I understand that it was a milestone too ambitious to achieve and the technical debt would have been enormous with your current programming knowledge.

Good luck.

PR #9354, opened 2025-11-04
Interfaces: Neighbors: Automatic Discovery - add new hostdiscovery feature

AdSchellevis:
This splits the Neighbors menu item in an existing static assignments and adds the automatic discovery feature which includes logging and insights into the currently known hosts on the network. When automatic discovery is disabled, it drops back to arp/ndp.

By default automatic mode is enabled on all interfaces, but configuration options exists (to limit interfaces and networks).

Other consumers which require host mappings can use the /usr/local/opnsense/scripts/interfaces/list_hosts.py  script to either fetch the sqlite database or arp/ndp data for the selected pprotocol(s).

More information about the hostwatch daemon is available at https://github.com/opnsense/hostwatch

### DIFF
--- a/src/opnsense/mvc/app/views/OPNsense/Hostdiscovery/settings.volt
+++ b/src/opnsense/mvc/app/views/OPNsense/Hostdiscovery/settings.volt
@@ -42,9 +42,6 @@ POSSIBILITY OF SUCH DAMAGE.
                     dfObj.resolve();
                 });
                 return dfObj;
-            },
-            onAction: function(data, status) {
-                updateServiceControlUI('hostdiscovery');
             }
         });
 
@@ -116,4 +113,4 @@ POSSIBILITY OF SUCH DAMAGE.
     </div>
 </div>
 
-{{ partial('layout_partials/base_apply_button', {'data_endpoint': '/api/hostdiscovery/service/reconfigure'}) }}
+{{ partial('layout_partials/base_apply_button', {'data_endpoint': '/api/hostdiscovery/service/reconfigure', 'data_service_widget': 'hostdiscovery'}) }}
--- a/src/opnsense/scripts/interfaces/setup_hostwatch.sh
+++ b/src/opnsense/scripts/interfaces/setup_hostwatch.sh
@@ -1,6 +1,6 @@
 #!/bin/sh
 
-mkdir /var/db/hostwatch 2>/dev/null
+mkdir -p /var/db/hostwatch 2>/dev/null
 mkdir /var/run/hostwatch 2>/dev/null
 chown -R hostd:hostd /var/db/hostwatch
 chown -R hostd:hostd /var/run/hostwatch


src/etc/inc/plugins.inc.d/hostwatch.inc
```diff
@@ -0,0 +1,64 @@
...
+function hostwatch_services()
+{
+    $services = [];
+    $template = [
+        'configd' => [
+            'restart' => ['kea restart'],
+            'start' => ['kea start'],
+            'stop' => ['kea stop'],
+        ],
+        'name' => 'kea-dhcp',
```
Monviech: Why is kea here?

src/opnsense/mvc/app/models/OPNsense/Hostdiscovery/Menu/Menu.xml
```diff
@@ -0,0 +1,8 @@
+<menu>
+    <Interfaces>
+        <Neighbors>
+            <Hostdiscovery VisibleName="Automatic Discovery" cssClass="fa fa-network-wired fa-fw"  url="/ui/hostdiscovery/settings"/>
+            <LogFile order="100" VisibleName="Discovery log" url="/ui/diagnostics/log/core/hostwatch"/>
```
Monviech, line 5: The log is missing in the ACL

src/etc/inc/plugins.inc.d/hostwatch.inc
fichtner: Nice catch 😊

src/opnsense/mvc/app/views/OPNsense/Hostdiscovery/settings.volt
```diff
@@ -0,0 +1,119 @@
...
+        $("#reconfigureAct").SimpleActionButton({
+            onPreAction: function() {
+                const dfObj = new $.Deferred();
+                saveFormToEndpoint("/api/hostdiscovery/settings/set", 'frm_settings', function(){
+                    dfObj.resolve();
+                });
+                return dfObj;
+            },
+            onAction: function(data, status) {
+                updateServiceControlUI('hostdiscovery');
```
Monviech: This shouldn't be needed if "data_service_widget" inside the base_apply_button is used.

https://github.com/opnsense/core/blob/fff33db13608ade8952fe57bf1267e9b6b72fe69/src/opnsense/mvc/app/views/OPNsense/Unbound/acl.volt#L74

src/etc/inc/plugins.inc.d/hostwatch.inc
AdSchellevis: face-palm, forgot to remove $template

src/opnsense/scripts/interfaces/setup_hostwatch.sh
```diff
@@ -0,0 +1,6 @@
+#!/bin/sh
+
+mkdir /var/db/hostwatch 2>/dev/null
```
Monviech: Wouldn't `mkdir -p` be safer here?

src/opnsense/scripts/interfaces/list_hosts.py
```diff
@@ -0,0 +1,92 @@
...
+        return open(rc_file,'r').read().find('hostwatch_enable="YES"') > -1
+    return False
+
+
+if __name__ == '__main__':
+    parser = argparse.ArgumentParser(
+        description='Dump known host entries',
+        epilog='Outputs a list of entries containing [ifname, ether, ip_address <,vendor>]'
+    )
+    parser.add_argument('-d', '--discover', help='Disable host discovery data', action="store_false", default=True)
```
swhite2, line 46:
```suggestion
    parser.add_argument('-d', '--discover', help='Use host discovery data', action="store_false", default=True)
```

Or perhaps I'm misinterpreting this

PR #9367, opened 2025-11-07
Update fw_log.volt

pallebone:
Add additional needed live log value amounts for 30k and 80k respectively to traverse a longer time in the past. Please consider.

In the past versions there were more options to go back in time much further in opnsense. Since the new version this has been reduced, but the new interface is much faster. Can you consider these values so I dont need to manually edit the file myself after every update?

Kind regards
Peter

### DIFF
--- a/src/opnsense/mvc/app/views/OPNsense/Diagnostics/fw_log.volt
+++ b/src/opnsense/mvc/app/views/OPNsense/Diagnostics/fw_log.volt
@@ -1430,6 +1430,8 @@
                     <option value="50">50</option>
                     <option value="75">75</option>
                     <option value="100">100</option>
+                    <option value="30000">30000</option>
+                    <option value="80000">80000</option>
                 </select>
                 <label>{{ lang._('Table size') }}</label>
             </div>
@@ -1438,6 +1440,7 @@
                     <option value="10000">10000</option>
                     <option value="20000">20000</option>
                     <option value="30000">30000</option>
+                    <option value="80000">80000</option>
                 </select>
                 <label>{{ lang._('History size') }}</label>
             </div>


AdSchellevis:
The new interface uses a larger queue underneath, which removes the need for very large tables (which will be extremely slow to render). I don't think we should do this, knowing the search window is larger anyway.

If I'm not mistaken, the total window size is specified here:

https://github.com/opnsense/core/blob/cbd6ea95a759e61bb9a8475b877ebdabf59141e6/src/opnsense/mvc/app/views/OPNsense/Diagnostics/fw_log.volt#L874-L874

pallebone:
How do I display 30k records on the screen that match a specific label on the screen with the current setup? Seems to show only 100?

AdSchellevis:
it shows 100 out of maximum 30k, filters can be used to limit the active set.

pallebone:
> it shows 100 out of maximum 30k, filters can be used to limit the active set.

Yes I understand. But 100 isn't enough and before the change it was many many many more. I believe at least 20k on the old version... 

Is there a problem to leave the choice up to the end user? I understand you many not use this but for those of us who do…

AdSchellevis:
we'll keep it under consideration, when offering choices beyond the specification of the used components, it usually leads to complains from others at a later point in time. A bit larger is likely not an issue, but nobody has a screen that first 30k rows anyway

pallebone:
Ok thank you for considering the issue I have.

swhite2:
@pallebone https://github.com/opnsense/core/commit/07e0ba8b98a5fa8dc294e7dccf3ed1c8d089e05b should make this a little better. We can go with higher numbers, but given the constraint mentioned in the PR of this commit we should stop polling on larger table sizes to prevent rendering issues (perhaps already at > 100). The history size (total queue) should also be automatically matched and have higher values. Feel free to take a stab at this if you want in a new PR and I'll have a look. Closing this one in the meantime.

PR #9393, opened 2025-11-14
backend: exec_safe() improvement and Shell class consolidation

fichtner:


src/opnsense/mvc/app/library/OPNsense/Core/Safe.php
```diff
@@ -0,0 +1,71 @@
...
+ * SUBSTITUTE GOODS OR SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS
+ * INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN
+ * CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE)
+ * ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
+ * POSSIBILITY OF SUCH DAMAGE.
+ */
+
+namespace OPNsense\Core;
+
+class Safe
```
AdSchellevis: @fichtner it might be better to replace the current `Shell` class, it's only used in a single spot as far as I know and Shell is likely a more logical name for the things we want to do here.

fichtner: True. I'm a bit experimenting with where to take this from here at the moment. The util.inc wrapping works well already. :)

This is a bit weird btw:

https://github.com/opnsense/core/blob/67de8a646c691bfd11ba703dfc236010dc0c5461/src/opnsense/mvc/app/library/OPNsense/Trust/Store.php#L645

fichtner: Speaking of Shell class: do we really need to instantiate it?

AdSchellevis: > True. I'm a bit experimenting with where to take this from here at the moment. The util.inc wrapping works well already. :)
> 
> This is a bit weird btw:
> 
> https://github.com/opnsense/core/blob/67de8a646c691bfd11ba703dfc236010dc0c5461/src/opnsense/mvc/app/library/OPNsense/Trust/Store.php#L645

That does look weird indeed, doesn't look intentional.



> Speaking of Shell class: do we really need to instantiate it?

Not really, it's old, static in these cases are fine. Monit can probably live without the current callouts.

fichtner: Ok, done. The monit thing is easy, see fa16cd7b9d08fd703!

I'll fiddle a bit with what's next. execSingle will probably become what mwexecf was, and we don't need the exec for the multiple commands I'm sure (plus array of format is now a single command anyway)

AdSchellevis: @fichtner nice, thanks!

fichtner:
@AdSchellevis I haven't tested every one of these, but the most important during operation all seem to work in a VM now. So IMO this can go to master and the fully compatible changes in util.inc + Shell.php should go into .8 along with a few selected execution changes as discussed.

fichtner:
@AdSchellevis found one more bug https://github.com/opnsense/core/pull/9393/commits/871cd1cc7d and tested most of this. It's ready for master if you agree.

AdSchellevis:
@fichtner Although I haven't reviewed al code, the parts that I did check look good. As discussed we should think about release timing, backwards compatible groundworks first (and fast), the rest maybe in a couple of cycles. 

In terms of documentation we might also consider documenting some of the convenience php functions we have now as they are likely with us for the long run.
This includes the file helpers (e.a. `File::file_put_contents` and friends) and the now cleaned up `Shell` class (which can be used both by the frontend as the backend, as long as one considers no special access rights should be required).

fichtner:
@AdSchellevis ok, I'll keep the parent ticket open and add a reminder for the docs part

PR #9401, opened 2025-11-16
Kea: Add DHCP-DDNS Support

### LINKED ISSUE #9359: dhcp/kea:  Improve DDNS support
**Important notices**

When migrating off of isc-dhcp onto kea automatic update of DNS records is lost, kea does support this though a helper utility,  opnsense should implement the configuration gui to enable this.

https://kea.readthedocs.io/en/kea-2.2.0/arm/ddns.html

ruifung:
This PR adds full support for enabling and configuring the Kea DHCP-DDNS server.

This will enable Kea to perform Dynamic DNS updates over RFC2136 for both IPv4 and IPv6.

Are there any docs changes that should be done for this too?

Closes #9359

### DIFF
--- a/src/opnsense/mvc/app/models/OPNsense/Kea/KeaDhcpDdns.php
+++ b/src/opnsense/mvc/app/models/OPNsense/Kea/KeaDhcpDdns.php
@@ -29,7 +29,6 @@ namespace OPNsense\Kea;
 
 use OPNsense\Core\File;
 use OPNsense\Base\BaseModel;
-use OPNsense\Base\Messages\Message;
 
 class KeaDhcpDdns extends BaseModel
 {
@@ -38,26 +37,40 @@ class KeaDhcpDdns extends BaseModel
         return $this->general->enabled->isEqual('1');
     }
 
-    private function getTsigKeys() {
-        return array_values($this->tsig_keys->getNodes());
+    private function getTsigKeys()
+    {
+        $tsig_keys = [];
+        foreach ($this->tsig_keys->iterateItems() as $key) {
+            $tsig_keys[] = [
+                'name' => $key->name->getValue(),
+                'algorithm' => $key->algorithm->getValue(),
+                'secret' => $key->secret->getValue(),
+            ];
+        }
+        return $tsig_keys;
     }
 
-    private function buildDomains($domainsNode) {
-        $tsigKeys = array_column(array_values($this->tsig_keys->getNodes()), 'name', '__uuid__');
-
-        return array_map(function ($domain) use ($tsigKeys) {
+    private function buildDomains($domainsNode)
+    {
+        $domains = [];
+        $tsigNameMap = [];
+        foreach ($this->tsig_keys->iterateItems() as $uuid => $key) {
+            $tsigNameMap[$uuid] = $key->name->getValue();
+        }
+        foreach ($domainsNode->iterateItems() as $domain) {
             $server = [
                 'ip-address' => $domain->ip_address->getValue(),
                 'port' => $domain->port->asInt()
             ];
-            if (!empty($tsigKeys[$domain->key_name->getValue()])) {
-                $server['key-name'] = $tsigKeys[$domain->key_name->getValue()];
+            if (!empty($tsigNameMap[$domain->key_name->getValue()])) {
+                $server['key-name'] = $tsigNameMap[$domain->key_name->getValue()];
             }
-            return [
+            $domains[] = [
                 'name' => $domain->name->getValue(),
                 'dns-servers' => [$server]
             ];
-        }, iterator_to_array($domainsNode->iterateItems()));
+        }
+        return $domains;
     }
 
     public function generateConfig($target = '/usr/local/etc/kea/kea-dhcp-ddns.conf')


src/opnsense/mvc/app/controllers/OPNsense/Kea/Api/DhcpddnsController.php
```diff
@@ -0,0 +1,151 @@
...
+ */
+
+namespace OPNsense\Kea\Api;
+
+use OPNsense\Base\ApiMutableModelControllerBase;
+
+class DhcpddnsController extends ApiMutableModelControllerBase
+{
+    protected static $internalModelName = 'dhcp_ddns';
+    protected static $internalModelClass = 'OPNsense\\Kea\\KeaDhcpDdns';
```
fichtner: IMO this shouldn't need escaping. None of the other code does this.

src/opnsense/mvc/app/models/OPNsense/Kea/KeaDhcpDdns.xml
```diff
@@ -0,0 +1,135 @@
+<model>
+    <mount>//OPNsense/Kea/dhcp_ddns</mount>
+    <version>1.0.7</version>
```
fichtner:
```suggestion
    <version>1.0.0</version>
```

Let's start at the beginning ;)

src/opnsense/mvc/app/models/OPNsense/Kea/KeaDhcpv4.php
```diff
@@ -90,6 +90,34 @@ public function performValidation($validateFullModel = false)
             }
         }
 
+        // Require DDNS service to be enabled when DHCPv4's DDNS is enabled
+        // Skip this check when manual configuration is enabled for DHCPv4
+        if (!empty((string)$this->general->enable_ddns) && $this->general->manual_config->isEmpty()) {
```
fichtner: I don't know if these crisscross validations are helpful in the big picture. They always bogged legacy pages down locking people from doing what they wanted to do in the first place.

src/opnsense/mvc/app/controllers/OPNsense/Kea/Api/DhcpddnsController.php
```diff
@@ -0,0 +1,151 @@
+<?php
+
+/*
+ * Copyright (C) 2025 Deciso B.V.
```
fichtner: I bet someone else wrote this code actually :)

src/etc/inc/plugins.inc.d/kea.inc
```diff
@@ -159,13 +168,19 @@ function kea_configure_do($verbose = false)
             /* skip kea-dhcp6.conf when configured manually */
             $keaDhcpv6->generateConfig();
         }
+        if ($keaDhcpDdns->isEnabled() && $keaDhcpDdns->general->manual_config->isEmpty()) {
+            $keaDhcpDdns->generateConfig();
```
fichtner, line 172: if this is a separate service I don't see it being started?

fichtner:
I made a few comments to move this along. In general this looks promising although I have to say test coverage will be tricky with all the options already present here.


Thanks,
Franco

src/etc/inc/plugins.inc.d/kea.inc
ruifung: It's started by keactrl (enabled in it's config) just like Dhcp4 and Dhcp6, they get started/stopped together. Not sure if the existing code is already doing additional steps to start/stop kea in general though.

EDIT: yep. looked in actions_kea.conf, it just calls /usr/local/etc/rc.d/kea which uses keactrl to start/stop everything together.

src/opnsense/mvc/app/controllers/OPNsense/Kea/Api/DhcpddnsController.php
ruifung: Eeeh, I just copied the copyright headers, not quite sure what do to here...? 

EDIT: nevermind, looked at the COPYRIGHT file and some other code files, I'll update it.

src/opnsense/mvc/app/models/OPNsense/Kea/KeaDhcpv4.php
ruifung: I mean, fair enough, it doesn't break anything if it's not included, it's just something that might potentially be missed when enabling ddns.

i.e. enabled ddns in the v4/v6 subnets, but forgot to enable the ddns server or something.

src/opnsense/mvc/app/controllers/OPNsense/Kea/Api/DhcpddnsController.php
```diff
@@ -0,0 +1,151 @@
...
+    protected static $internalModelName = 'dhcp_ddns';
+    protected static $internalModelClass = 'OPNsense\\Kea\\KeaDhcpDdns';
+
+    /**
+     * Return general model data for DDNS
+     * @return array
+     */
+    public function getAction()
+    {
+        // default implementation from ApiMutableModelControllerBase is sufficient
```
AdSchellevis: ditch the method in that case (when not defined, it falls back to the existing anyway)

src/opnsense/mvc/app/models/OPNsense/Kea/KeaDhcpDdns.xml
ruifung: Was just wondering how a running system would react to this being reset (my own installations), but I tested it and it seems fine.

src/opnsense/mvc/app/controllers/OPNsense/Kea/Api/DhcpddnsController.php
ruifung: Do I need to update the COPYRIGHT file too?

src/opnsense/mvc/app/controllers/OPNsense/Kea/Api/DhcpddnsController.php
ruifung: Removed redundant stub method.

ruifung:
Rebased on latest state of master

ruifung:
> I made a few comments to move this along. In general this looks promising although I have to say test coverage will be tricky with all the options already present here.
> 
> Thanks, Franco

Anything I can do to help on that?

Monviech:
(this is just a comment, do not act upon it (yet))

Reading the scope of the code it feels like theres a lot of feature bloat in there. Couldn't this be stripped down to a more minimal viable thingy with only whats essential for the forward and reverse zones? Eg. these regex and replacement fields are most likely not needed by 99% of the users.

I also feel like theres too much "is this service enabled" validations going on, that could create circular dependencies, I would also cut this down to essentials (aka only validate what would hard crash KEA with an error).

I know that this is rather complete, but adding too much initially will add all the bloat that is unknown if anybody actually needs it, but then its around for good. Adding is easy, removing is very hard, so adding less and only adding more upon request is the better route to go imo.

I have the feeling if we would trim this down to the essentials it would fit at least 90% of usecases and safe quite some potentially dead code.

ruifung:
I suppose the hostname regex/replacement fields aren't likely to be needed by most users yeah. I'm not actually sure under what scenario would you want to uh, regex characters out of a client hostname, given that I think hostnames should already be valid as a DNS record anyway (?).

Although I don't think the same can be said of any other fields right now though. Except maybe the override no-update / override client update options, I'm not sure what options most DHCP clients can/will set with regards to those preference flags in their DHCP requests.

And I suppose if the criteria for "should this be validated?" is "will this crash kea?" then yeah, the service enabled validations are not needed at all. The worst that will happen is kea dhcp4/dhcp6 just logging warnings every time it tries to update if the ddns service isn't running.

src/opnsense/mvc/app/controllers/OPNsense/Kea/forms/generalSettings6.xml
```diff
@@ -38,6 +38,12 @@
         <type>checkbox</type>
         <help>Automatically add a basic set of firewall rules to allow dhcp traffic, more fine grained controls can be offered manually when disabling this option.</help>
     </field>
+    <field>
+        <id>dhcpv6.general.enable_ddns</id>
+        <label>Enable DDNS</label>
```
Monviech: Why can you enable this at three spots:
- the service itself
    - inside the dhcpv4/6 settings
        - inside each pool

src/opnsense/mvc/app/models/OPNsense/Kea/KeaDhcpDdns.php
```diff
@@ -0,0 +1,250 @@
...
+{
+    /**
+     * {@inheritdoc}
+     */
+    public function performValidation($validateFullModel = false)
+    {
+        // Run default field-level validators first
+        $messages = parent::performValidation($validateFullModel);
+
+        // Prevent disabling the DDNS service while DHCPv4 or DHCPv6 has DDNS enabled
```
Monviech: Remove as much validation as possible that does not crash the service, just help the user by limiting choices. If there arent so many enabled checkboxes things cant be misconfigured.

src/opnsense/mvc/app/controllers/OPNsense/Kea/forms/dialogDdnsDnsServer.xml
```diff
@@ -0,0 +1,33 @@
+<form>
+    <field>
+        <id>dns_servers.name</id>
```
Monviech: Wouldnt a description work better for this? If the name isnt needed why store it?

src/opnsense/mvc/app/controllers/OPNsense/Kea/forms/dialogSubnet4.xml
```diff
@@ -28,6 +28,106 @@
...
+        <id>subnet4.ddns_options.send_updates</id>
+        <label>Send updates</label>
+        <type>checkbox</type>
+        <help>Enable Dynamic DNS updates for this subnet.</help>
+        <grid_view>
+            <visible>false</visible>
+        </grid_view>
+    </field>
+    <field>
+        <id>subnet4.ddns_options.override_no_update</id>
```
Monviech: Not sure if this is needed right away, could be added later?

src/opnsense/mvc/app/controllers/OPNsense/Kea/forms/dialogSubnet4.xml
```diff
@@ -28,6 +28,106 @@
...
+        <label>Override no-update</label>
+        <type>checkbox</type>
+        <help>Override client request to disable DNS updates.</help>
+        <advanced>true</advanced>
+        <grid_view>
+            <visible>false</visible>
+        </grid_view>
+    </field>
+    <field>
+        <id>subnet4.ddns_options.override_client_update</id>
```
Monviech: Same here?

src/opnsense/mvc/app/controllers/OPNsense/Kea/forms/dialogSubnet4.xml
```diff
@@ -28,6 +28,106 @@
...
+        <label>Override client update</label>
+        <type>checkbox</type>
+        <help>Server performs updates regardless of client preference.</help>
+        <advanced>true</advanced>
+        <grid_view>
+            <visible>false</visible>
+        </grid_view>
+    </field>
+    <field>
+        <id>subnet4.ddns_options.replace_client_name</id>
```
Monviech: Same here?

src/opnsense/mvc/app/controllers/OPNsense/Kea/forms/dialogSubnet4.xml
```diff
@@ -28,6 +28,106 @@
...
+        <label>Replace client name</label>
+        <type>dropdown</type>
+        <help>Controls when the server replaces the client supplied hostname.</help>
+        <grid_view>
+            <visible>false</visible>
+        </grid_view>
+    </field>
+    <field>
+        <id>subnet4.ddns_options.generated_prefix</id>
```
Monviech: Same here?

src/opnsense/mvc/app/controllers/OPNsense/Kea/forms/dialogSubnet4.xml
```diff
@@ -28,6 +28,106 @@
...
+        <help>Send DNS updates when leases are renewed.</help>
+        <grid_view>
+            <visible>false</visible>
+        </grid_view>
+    </field>
+    <field>
+        <id>subnet4.ddns_options.qualifying_suffix</id>
+        <label>Qualifying suffix</label>
+        <type>text</type>
+        <help>Suffix appended to hostnames to form FQDNs. Must end with a dot.</help>
```
Monviech, line 58: Same here? Also if a hostname is registered inside a zone wouldnt it become an FQDN automatically?

src/opnsense/mvc/app/controllers/OPNsense/Kea/forms/dialogSubnet4.xml
```diff
@@ -28,6 +28,106 @@
...
+        <grid_view>
+            <visible>false</visible>
+        </grid_view>
+    </field>
+    <field>
+        <id>subnet4.ddns_options.hostname_char_set</id>
```
Monviech: Same here?

src/opnsense/mvc/app/controllers/OPNsense/Kea/forms/dialogSubnet4.xml
```diff
@@ -28,6 +28,106 @@
...
+        <label>Hostname invalid char set</label>
+        <type>text</type>
+        <help>POSIX Extended Regex Expression that matches invalid characters. Default: [^A-Za-z0-9.-]</help>
+        <advanced>true</advanced>
+        <grid_view>
+            <visible>false</visible>
+        </grid_view>
+    </field>
+    <field>
+        <id>subnet4.ddns_options.hostname_char_replacement</id>
```
Monviech: Same here?

src/opnsense/mvc/app/controllers/OPNsense/Kea/forms/dialogSubnet4.xml
```diff
@@ -28,6 +28,106 @@
             <visible>false</visible>
         </grid_view>
     </field>
+    <field>
+        <type>header</type>
+        <label>Dynamic DNS (DDNS)</label>
+        <collapse>true</collapse>
+    </field>
+    <field>
+        <id>subnet4.ddns_options.send_updates</id>
```
Monviech, line 37: As far as I can see you can configure multiple servers and forward and reverse zones, but you cant choose any of these here explicitely.

I dont see a one to many relationship.

I feel like theres something wrong here.

src/opnsense/mvc/app/controllers/OPNsense/Kea/forms/dialogDdnsDnsServer.xml
ruifung: Mostly because its entirely possible to have different tsig / server pairs here, the name is entirely for display / user facing purpose.

Although I guess description is probabaly a better name for it.

src/opnsense/mvc/app/controllers/OPNsense/Kea/forms/generalSettings6.xml
ruifung: Because they correspond to different settings in the kea config.

1. Enables the ddns server compnent
2. Enables sending ddns updates for the dhcp server components
3. Controls if a subnet scope should send ddns updates at all.

src/opnsense/mvc/app/controllers/OPNsense/Kea/forms/dialogSubnet4.xml
ruifung: This is useful on IoT networks to force replacement of the hostnames crap devices like to send (or not send) sometimes with a unique generated one.

src/opnsense/mvc/app/controllers/OPNsense/Kea/forms/dialogSubnet4.xml
ruifung: Related to the hostname replacement option above

src/opnsense/mvc/app/controllers/OPNsense/Kea/forms/dialogSubnet4.xml
ruifung: No, kea doesnt do that automatically, you need to tell it what to append to turn single hostnames into fqdns.

It will happily try to register 'hostname.' (And fail because the zone doesnt exist.)

src/opnsense/mvc/app/controllers/OPNsense/Kea/forms/dialogSubnet4.xml
ruifung: Yeah, this one I agree with.

src/opnsense/mvc/app/controllers/OPNsense/Kea/forms/dialogSubnet4.xml
ruifung: Same here too.

src/opnsense/mvc/app/controllers/OPNsense/Kea/forms/dialogSubnet4.xml
ruifung: Kea does this automatically based on the fully qualified hostname either that's sent by the client in a FQDN option or created by prepending the qualifying suffix above to the hostname.

Multiple servers are used to configure the zones, and kea-dhcp-ddns selects the most specific zone to use based on the fqdn being updated.

And a subnet is really the only scope that is supported in kea by opnsense.

src/opnsense/mvc/app/controllers/OPNsense/Kea/forms/dialogSubnet4.xml
ruifung: Fair enough for this one, I really don't know if clients sending options to skip ddns updates are common.

src/opnsense/mvc/app/controllers/OPNsense/Kea/forms/dialogSubnet4.xml
ruifung: Same with this, not sure how common are clients that send the option that this is overriding.

src/opnsense/mvc/app/controllers/OPNsense/Kea/forms/dialogSubnet4.xml
Monviech: That sounds stupid but I dont expect any less from Kea :)

src/opnsense/mvc/app/controllers/OPNsense/Kea/forms/dialogSubnet4.xml
Monviech: So if multiple servers are configured that all have the same zones, kea will send updates to all of them if the client option matches? I guess that was the relation I was missing here.

ruifung: It's the other way around I think, each zone can have one or more servers assigned, strictly speaking, the split between zones and servers here is mostly to reduce the amount of repetition needed since in the kea config, each forward/reverse zone specifies it's own set of dns servers... with IP, port and optionally per-server TSIG.

That was quite tedious to set up, so I opted to split it here to reduce duplicated effort when setting up the things.

And yeah, AFAIK the dhcp4/dhcp6 servers just send the FQDN/IP/DHCID to the DDNS server and it handles it based on that.

Monviech: Okay makes sense. My angle right now is to reduce some of the initial bloat so the essense of the feature reveals itself. That will make a positive review and merge far more likely. The less there is to worry about the less effort it takes to construct the mental model, and also to take the support burden long term.

src/opnsense/mvc/app/controllers/OPNsense/Kea/forms/generalSettings6.xml
Monviech: isnt 2 unnecessary? I could derive that from 3.

As soon as one of point 3 exists, it should start sending updates.

ruifung: Probably yeah, I initially just modelled the whole thing almost 1 to 1 with the kea configuration structure, but I guess the only use of (2) is to be able to turn off ddns updates for only the dhcp 4/6 server... which now that I think about it, probably isn't something anyone will use frequently at all.

Monviech: Yeah, I assume if users want to turn it off globally due to some reason they can just turn off the ddns server. So 2 is just an extra step without merit.

src/opnsense/mvc/app/models/OPNsense/Kea/KeaDhcpDdns.php
ruifung: I'll remove the enable validation, but will keep the FQDN-ends-with-dot validation because that's kind of a very easy footgun to hit.

Monviech: Essentially if its field specific it would be better to have a custom field type for that as it can also be unit tested. Compare to here:
https://github.com/opnsense/core/pull/9232

You could extend the HostnameField (https://github.com/opnsense/core/blob/master/src/opnsense/mvc/app/models/OPNsense/Base/FieldTypes/HostnameField.php) and add the validation you need on top.

src/opnsense/mvc/app/models/OPNsense/Kea/KeaDhcpv4.php
ruifung: Removed criscross validations

src/opnsense/mvc/app/models/OPNsense/Kea/KeaDhcpDdns.php
Monviech: Or a simple alternative, a Mask with a regex that verifies the input ends with a dot (.) in the model.

https://github.com/opnsense/core/blob/bc1a33057879119842b02e46705c798e549d346d/src/opnsense/mvc/app/models/OPNsense/Kea/KeaDhcpv4.xml#L58

Im not sure though if it works in the HostnameField, you have to test that.

src/opnsense/mvc/app/controllers/OPNsense/Kea/forms/generalSettings6.xml
ruifung: Removed the enable_ddns boolean field for Dhcp4/Dhcp6

src/opnsense/mvc/app/controllers/OPNsense/Kea/forms/dialogDdnsDnsServer.xml
ruifung: Renamed to description, still marked required though, because it's used to display in the Forward/Reverse zones

src/opnsense/mvc/app/controllers/OPNsense/Kea/forms/dialogSubnet4.xml
ruifung: Removed field

src/opnsense/mvc/app/controllers/OPNsense/Kea/forms/dialogSubnet4.xml
ruifung: Removed this field too

src/opnsense/mvc/app/controllers/OPNsense/Kea/forms/dialogSubnet4.xml
ruifung: Removed this field

src/opnsense/mvc/app/controllers/OPNsense/Kea/forms/dialogSubnet4.xml
ruifung: Removed this field

ruifung:
I've made several changes addressing further feedback

src/opnsense/mvc/app/models/OPNsense/Kea/KeaDhcpDdns.php
ruifung: By Extend the HostnameField, do you mean extend HostnameField in a new field class or did you mean to add additional logic into HostnameField.

Honestly, for an FqdnField it would, probably, be closer to a TextField just validated against FILTER_VALIDATE_DOMAIN and checked for a trailing dot.

or integrating the split/regex loop as seen in that dnsmasq pr

Monviech: If the mask doesnt work lets keep it as it is right now. We can always mess with the field types later. Sorry for rambling :)

ruifung: Yeah, HostnameField doesn't look like it has a mask field. I'll leave it for now then.

src/opnsense/mvc/app/models/OPNsense/Kea/KeaDhcpDdns.xml
```diff
@@ -0,0 +1,134 @@
...
+        <general>
+            <enabled type="BooleanField">
+                <Default>0</Default>
+                <Required>Y</Required>
+            </enabled>
+            <manual_config type="BooleanField"/>
+        </general>
+        <!-- shared DNS servers, can be referenced by domains -->
+        <dns_servers type="ArrayField">
+            <description type="TextField">
```
Monviech:
```suggestion
            <description type="DescriptionField">
```

https://docs.opnsense.org/development/frontend/models_fieldtypes.html#descriptionfield

src/opnsense/mvc/app/models/OPNsense/Kea/KeaDhcpDdns.xml
```diff
@@ -0,0 +1,134 @@
...
+                <Required>Y</Required>
+            </description>
+            <ip_address type="NetworkField">
+                <NetMaskAllowed>N</NetMaskAllowed>
+                <Required>Y</Required>
+            </ip_address>
+            <port type="IntegerField">
```
Monviech:
```suggestion
            <port type="PortField">
```

https://docs.opnsense.org/development/frontend/models_fieldtypes.html#portfield

src/opnsense/mvc/app/models/OPNsense/Kea/KeaDhcpDdns.xml
```diff
@@ -0,0 +1,134 @@
...
+                </Model>
+            </key_name>
+        </dns_servers>
+        <forward_ddns>
+            <ddns_domains type="ArrayField">
+                <name type="HostnameField">
+                    <IpAllowed>N</IpAllowed>
+                    <ValidationMessage>Please specify a valid domain name.</ValidationMessage>
+                    <Constraints>
+                        <uniq001>
```
Monviech:
```suggestion
                        <check001>
```

Just for brevity https://docs.opnsense.org/development/frontend/models_constraints.html#uniqueconstraint

src/opnsense/mvc/app/models/OPNsense/Kea/KeaDhcpv4.php
```diff
@@ -90,6 +90,23 @@ public function performValidation($validateFullModel = false)
             }
         }
 
+        // Enforce that ddns qualifying suffix ends with a dot when set
+        foreach ($this->subnets->subnet4->iterateItems() as $subnet) {
+            if (!$validateFullModel && !$subnet->isFieldChanged()) {
+                continue;
+            }
+            $send_updates = !empty((string)$subnet->ddns_options->send_updates);
```
Monviech: Why is this important? The input should always be correct regardless of if this is enabled or not?

src/opnsense/mvc/app/models/OPNsense/Kea/KeaDhcpv4.php
```diff
@@ -90,6 +90,23 @@ public function performValidation($validateFullModel = false)
...
+            $suffix = trim((string)$subnet->ddns_options->qualifying_suffix);
```
Monviech: Why do you trim here? The existing input validation should prevent whitespaces.

Monviech:
```suggestion
            $suffix = $subnet->ddns_options->qualifying_suffix->getValue());
```

https://github.com/opnsense/core/blob/6d1ff4c975f8468dacd0c8416c09b0e8ba49a0d0/src/opnsense/mvc/app/models/OPNsense/Base/FieldTypes/BaseField.php#L386

src/opnsense/mvc/app/models/OPNsense/Kea/KeaDhcpv4.php
```diff
@@ -90,6 +90,23 @@ public function performValidation($validateFullModel = false)
...
+            if ($send_updates && $suffix !== '' && substr($suffix, -1) !== '.') {
```
Monviech: Since you only check one condition now you can inline it here without variables.

src/opnsense/mvc/app/models/OPNsense/Kea/KeaDhcpv4.php
```diff
@@ -175,6 +192,28 @@ private function getConfigSubnets()
                 'pools' => [],
                 'reservations' => []
             ];
+
+            // Conditionally include DDNS settings only when send-updates is enabled,
+            // and only include fields that have meaningful values.
+            $ddns_send_updates = !empty((string)$subnet->ddns_options->send_updates);
```
Monviech:
```suggestion
            $ddns_send_updates = !($subnet->ddns_options->send_updates->isEmpty());
```

https://github.com/opnsense/core/blob/6d1ff4c975f8468dacd0c8416c09b0e8ba49a0d0/src/opnsense/mvc/app/models/OPNsense/Base/FieldTypes/BaseField.php#L542

src/opnsense/mvc/app/models/OPNsense/Kea/KeaDhcpv4.php
```diff
@@ -175,6 +192,28 @@ private function getConfigSubnets()
...
+            if ($ddns_send_updates) {
+                $record['ddns-send-updates'] = true;
+                if ((string)$subnet->ddns_options->replace_client_name !== '') {
```
Monviech: Can also use isEmpty()

src/opnsense/mvc/app/models/OPNsense/Kea/KeaDhcpv4.php
```diff
@@ -175,6 +192,28 @@ private function getConfigSubnets()
...
+                    $record['ddns-replace-client-name'] = (string)$subnet->ddns_options->replace_client_name;
+                }
+                if ((string)$subnet->ddns_options->generated_prefix !== '') {
```
Monviech: Can also use isEmpty()

The gist is do not cast manually here if the BaseField offers something that does it for you.
There are more spots to fix in this file but I will not reference all.

src/opnsense/mvc/app/models/OPNsense/Kea/KeaDhcpv4.php
```diff
@@ -175,6 +192,28 @@ private function getConfigSubnets()
...
                 'reservations' => []
             ];
+
+            // Conditionally include DDNS settings only when send-updates is enabled,
+            // and only include fields that have meaningful values.
+            $ddns_send_updates = !empty((string)$subnet->ddns_options->send_updates);
+            if ($ddns_send_updates) {
+                $record['ddns-send-updates'] = true;
+                if ((string)$subnet->ddns_options->replace_client_name !== '') {
+                    $record['ddns-replace-client-name'] = (string)$subnet->ddns_options->replace_client_name;
```
Monviech: Can also use getValue()

etc...

src/opnsense/mvc/app/models/OPNsense/Kea/KeaDhcpv4.xml
```diff
@@ -130,6 +130,32 @@
...
+                        <Default>never</Default>
+                        <Required>Y</Required>
+                        <OptionValues>
+                            <never>never</never>
+                            <always>always</always>
+                            <when-present>when-present</when-present>
+                            <when-not-present>when-not-present</when-not-present>
+                        </OptionValues>
+                    </replace_client_name>
+                    <generated_prefix type="TextField"/>
```
Monviech:
```suggestion
                    <generated_prefix type="HostnameField"/>
```

Maybe rather this as it looks like this is kinda hostname related stuff?

src/opnsense/mvc/app/models/OPNsense/Kea/KeaDhcpv4.xml
```diff
@@ -130,6 +130,32 @@
...
+                    <update_on_renew type="BooleanField"/>
+                    <qualifying_suffix type="TextField"/>
```
Monviech: Same here, maybe HostnameField?

src/opnsense/mvc/app/models/OPNsense/Kea/KeaDhcpv6.php
```diff
@@ -72,6 +72,23 @@ public function performValidation($validateFullModel = false)
             }
         }
 
+        // Enforce that ddns qualifying suffix ends with a dot when set
+        foreach ($this->subnets->subnet6->iterateItems() as $subnet) {
+            if (!$validateFullModel && !$subnet->isFieldChanged()) {
+                continue;
+            }
+            $send_updates = !empty((string)$subnet->ddns_options->send_updates);
```
Monviech: same comments as v4

src/opnsense/mvc/app/models/OPNsense/Kea/KeaDhcpv6.php
```diff
@@ -129,6 +146,28 @@ private function getConfigSubnets()
                 'pd-pools' => [],
                 'reservations' => []
             ];
+
+            // Conditionally include DDNS settings only when send-updates is enabled,
+            // and only include fields that have meaningful values.
+            $ddns_send_updates = !empty((string)$subnet->ddns_options->send_updates);
```
Monviech: same comments as v4

src/opnsense/mvc/app/models/OPNsense/Kea/KeaDhcpv6.xml
```diff
@@ -99,6 +99,32 @@
...
+                        <Default>never</Default>
+                        <Required>Y</Required>
+                        <OptionValues>
+                            <never>never</never>
+                            <always>always</always>
+                            <when-present>when-present</when-present>
+                            <when-not-present>when-not-present</when-not-present>
+                        </OptionValues>
+                    </replace_client_name>
+                    <generated_prefix type="TextField"/>
```
Monviech: same comments as v4

src/opnsense/mvc/app/models/OPNsense/Kea/KeaDhcpDdns.php
```diff
@@ -0,0 +1,230 @@
...
+    {
+        // Run default field-level validators first
+        $messages = parent::performValidation($validateFullModel);
+
+        // Explicitly validate that forward and reverse domain names end with a dot (FQDN)
+        foreach ($this->forward_ddns->ddns_domains->iterateItems() as $domain) {
+            if (!$validateFullModel && !$domain->isFieldChanged()) {
+                continue;
+            }
+            $name = trim((string)$domain->name);
```
Monviech: why trim? also getValue()

src/opnsense/mvc/app/models/OPNsense/Kea/KeaDhcpDdns.php
```diff
@@ -0,0 +1,230 @@
...
+                        $domain->__reference . '.name'
+                    )
+                );
+            }
+        }
+        foreach ($this->reverse_ddns->ddns_domains->iterateItems() as $domain) {
+            if (!$validateFullModel && !$domain->isFieldChanged()) {
+                continue;
+            }
+            $name = trim((string)$domain->name);
```
Monviech: same as earlier

src/opnsense/mvc/app/models/OPNsense/Kea/KeaDhcpDdns.php
```diff
@@ -0,0 +1,230 @@
...
+                    $entry['key-name'] = $tsigNameMap[$kn];
+                }
+            }
+
+            // dns-servers referenced via ModelRelationField (comma-separated UUIDs)
+            $servers = [];
+            $refs = isset($domain->dns_servers) ? (string)$domain->dns_servers : '';
+            if (!empty($refs)) {
+                foreach (array_filter(explode(',', $refs)) as $uuid) {
+                    $uuid = trim($uuid);
```
Monviech: Why trim an uuid?

Monviech (changes requested):
Now for the next round, just some general coding things that can be streamlined and or changed.

I don't know if everything I suggested is correct, but extrapolate from this.

src/opnsense/mvc/app/models/OPNsense/Kea/KeaDhcpv4.php
ruifung: I'm pretty sure i saw other spots using this style to check boolean flags, so I copied that... exactly.

ruifung: I'm pretty sure i saw other spots using this style to check boolean flags, so I copied that... exactly.

ruifung: I'm pretty sure i saw other spots using this style to check boolean flags, so I copied that... exactly.

Monviech: EDIT:

root@opn03:/src/git/core/src/opnsense/mvc/app/models # grep -r "isEmpty()"
./OPNsense/Kea/KeaDhcpv4.php:                if (!$reservation->hw_address->isEmpty()) {
./OPNsense/Kea/KeaDhcpv4.php:        if (!(new KeaCtrlAgent())->general->enabled->isEmpty()) {

Monviech: You are doing fine, with grep you can see these specific patterns used in many other files, functions like isEmpty(), isEqual(), getValue(), etc. They just help you with the type casting, so they improve your code.

This is shaping up good, I hope you do not get discouraged. Thanks for keeping it up :)

src/opnsense/mvc/app/models/OPNsense/Kea/KeaDhcpv4.php
ruifung: This pattern was used for the reservations validation above. So this was, more or less, copied from there.

I think the intention is to ensure that should invalid data be in the model, the validation doesn't fail when the user is saving another dialog with no way to correct the existing values...?

src/opnsense/mvc/app/models/OPNsense/Kea/KeaDhcpv4.php
ruifung: I mean, sure, but I mostly copied the code style from other parts of the files I was touching soo...

ruifung: Probably force of habit to trim everything from user input, I'll remove the trim then.

src/opnsense/mvc/app/models/OPNsense/Kea/KeaDhcpv4.php
ruifung: > EDIT:
> 
> root@opn03:/src/git/core/src/opnsense/mvc/app/models # grep -r "isEmpty()" ./OPNsense/Kea/KeaDhcpv4.php: if (!$reservation->hw_address->isEmpty()) { ./OPNsense/Kea/KeaDhcpv4.php: if (!(new KeaCtrlAgent())->general->enabled->isEmpty()) {

But at the same time... 
image

src/opnsense/mvc/app/models/OPNsense/Kea/KeaDhcpv4.php
ruifung: There are many spots to fix in this file, but I'll fix the bits I added.

src/opnsense/mvc/app/models/OPNsense/Kea/KeaDhcpDdns.xml
ruifung: Changed to DescriptionField

src/opnsense/mvc/app/models/OPNsense/Kea/KeaDhcpDdns.xml
ruifung: Changed

src/opnsense/mvc/app/models/OPNsense/Kea/KeaDhcpv4.php
ruifung: At any rate, I've did a full pass over all the php code I added and updated them to use those methods. Though I'll note that the code style I've observed in the files seem rather inconsistent about that =P

src/opnsense/mvc/app/models/OPNsense/Kea/KeaDhcpv4.xml
ruifung: That works.

src/opnsense/mvc/app/models/OPNsense/Kea/KeaDhcpDdns.php
ruifung: Yeah, I think I missed this from my initial prototyping of the implementation.

At any rate, I've did another pass over all the code I've added to make sure there's no late-night misses left around.

src/opnsense/mvc/app/models/OPNsense/Kea/KeaDhcpv4.php
ruifung: I assume you mean $send_updates, at any rate, I've reworked this using the methods from BaseField too.

ruifung:
Made several changes in-line with the latest round of suggestions, and also went over the DDNS model php file to change it to use the methods from BaseField as suggested.

There really is a lot of inconsistency in the kea dhcp4/dhcp6 code though, there's lots of field casting, and some use of isEmpty().

It seems like a git blame on BaseField says that at least the isEmpty() and getValue() methods were added somewhat recently, and parts of the kea code predates their existence?

Monviech:
Hello, yes some things have been added recently, that means if we use them in new code additions right away we dont have to clean up. Thanks I'll look at this tomorrow :)

src/opnsense/mvc/app/controllers/OPNsense/Kea/forms/dialogDdnsForwardDomain.xml
```diff
@@ -0,0 +1,27 @@
...
+            <label>TSIG key</label>
+        </grid_view>
+    </field>
+    <field>
+        <id>ddns_domains.dns_servers</id>
+        <label>DNS servers</label>
+        <type>select_multiple</type>
+        <help>Select one or more shared DNS servers to update for this domain.</help>
+        <grid_view>
+            <label>DNS servers</label>
```
Monviech: you only need to define a label for the grid column if it should be different than from the dialog form.

src/opnsense/mvc/app/controllers/OPNsense/Kea/forms/dialogDdnsForwardDomain.xml
```diff
@@ -0,0 +1,27 @@
...
+            <visible>true</visible>
```
Monviech: visible true is the default, you only need to define visible false if you want to hide the column from the grid.

Monviech: This means you can remove the whole grid view key almost everywhere if you dont need to change defaults in it.

src/opnsense/mvc/app/models/OPNsense/Kea/KeaDhcpDdns.xml
```diff
@@ -0,0 +1,132 @@
...
+                <Required>Y</Required>
+                <Constraints>
+                    <check001>
+                        <ValidationMessage>Duplicate TSIG key name exists.</ValidationMessage>
+                        <type>UniqueConstraint</type>
+                    </check001>
+                </Constraints>
+            </name>
+            <algorithm type="OptionField">
+                <Required>Y</Required>
```
Monviech, line 86: Needs a default if required.

EDIT: Disregard this comment, I tested via "make lint-model" and there is no issue.

src/opnsense/mvc/app/controllers/OPNsense/Kea/forms/dialogDdnsForwardDomain.xml
ruifung: Removed redundant grid_view keys except on the subnet4/subnet6 ones because those are hidden.

Monviech:
Without the intent to totally derail this PR, here is the whole of what ISC had:
DHCPv4
image
DHCPv6
image

It was configured per range.

Why do I need more than that now in KEA, if the above was all that was ever needed without known complains about lacking in functionality?

I still have the feeling the current PR is a bit overdesigned, even though it seems to be quite generic. Most could be implicitly generated just from a few new fields in the ranges.

Challenge my assumption if you want to, this does not necessarily mean that ISC was designed correctly around this feature, but it has been awfully quiet around it since years (and I read a lot of forum, reddit, github issues).

ruifung:
> Without the intent to totally derail this PR, here is the whole of what ISC had:
> DHCPv4
> image
> DHCPv6
> image
> 
> It was configured per range.
> 
> Why do I need more than that now in KEA, if the above was all that was ever needed without known complains about lacking in functionality?
> 
> I still have the feeling the current PR is a bit overdesigned, even though it seems to be quite generic. Most could be implicitly generated just from a few new fields in the ranges.
> 
> Challenge my assumption if you want to, this does not necessarily mean that ISC was designed correctly around this feature, but it has been awfully quiet around it since years (and I read a lot of forum, reddit, github issues).

1. Not needing to adjust every single subnet (x2 for both dhcp4 and 6) if you change your DNS servers / rotate TSIG keys is great. This is exactly what I had to do for ISC DHCP. Not something you do often, but its definitely less effort to configure when you don't need to adjust the DDNS server options on every subnet.

3. Also, DDNS zone and subnet domain is not necessarily always the same, I believe the official RFC2136 specification has DDNS updates having a zone and one or more RRSets, implying that you can have a.b.x.y.z FQDN updated into x.y.z zone. Some DNS servers (Like earlier versions of Technitium DNS) would very strictly follow the zone specified in the request and fail to process the update if you sent an update with b.x.y.z as the zone to update.

4. Other then the split to separate out the DDNS service configs, some of which I've already addressed the use cases for:

- ~~hostname generation/replacement ( to replace badly behaved IoT clients that all send the -same- or no hostname, like a bunch of smart plugs I have all send p110 as the hostname)~~
  - Actually taking a second look at the kea docs, these look quite useless since it generates a client name using (prefix)-(address text). This is silly and serves no real purpose beyond just giving it a rDNS PTR record. It's not even a stable generated name using the DUID or MAC address since it's tied to the *lease*, and if you set reservations, you can override the hostname there.
  - I'm not even sure why would you want to clutter your DNS Zone with prefix-ip-address.domain records because if you know the domain you already know the IP address.
  - It seems like the only way to have somewhat useful hostname generation for this use case is to use their hook-ddns-tuning which provides a hostname-expr [parameter](https://kea.readthedocs.io/en/latest/arm/hooks.html#procedural-host-name-generation) to do this. But this is most likely beyond the scope of this PR.

- conflict resolution - this controls how kea uses the DHCID records, and is basically mandatory to tweak if you want to point dhcpv4 and dhcpv6 at the same zone. Because many clients do not send their DUID as a client identifier in dhcp4 requests. So they get different DHCIDs between dhcp4 and dhcp6 resulting in either dhcp4 or dhcp6 not updating DDNS because of the detected conflict. This -never- worked for me in ISC dhcp6, likely because of the conflict detection.

- the qualifying name is just that, a qualifying name. Kea (and ISC DHCP) does actually respect the FQDN  for clients that sends that. Naming the fields like this makes it more obvious what the behaviour is.

- Its also handy to have kea able to send DDNS updates on lease renew instead of on just lease creation. Mostly when tweaking DDNS settings.

4. I know there's the manual config option, but it's really suboptimal because now you can't rely on config backup and XMLRPC HA synchronization for Kea.

Monviech:
The added convenience inflates this feature though. I think some inconvenience is okay if it could cut down most of the code.

Essentially this is a set and forget feature for most so I do not see the need for so many new grids that relate to another.

Though lets keep it like this for now and see what others have to say, overall this feature would have it far easier in the review process if the LOC footprint would be less. That would also reduce future maintainance load/regression risk if it has less turning knobs.

ruifung:
I've further removed the generated prefix and client name replacement options as these result in generating a hostname with the assigned IP address, which is, of questionable usefulness for most cases where you want the DHCP server to register clients into DNS.

What's the point of having a DNS entry where you need to know the IP to get it's DNS entry?

Monviech:
I still have a feeling that if we would put the "tsig key" and the "dns server" into the existing "Dhcpv4" and "Dhcpv6" pools model, we could cut half of the LOC footpring of this code. And we could generate the config without having these reference maps in php.

The config could be built from information of the v4 and v6 models from flat relations.

Sure it would be more inconvenient for the user (if they have to change the key or server entry in a few pools), but the tradeoff would be maintainability and a clear scope of the offered feature (its very easy to configure, and very hard to do a mistake that way).

Other than the inconvenience, the scope of the ISC feature has seemed to be just right for /most/ people (not all of them).

ruifung:
Are you also suggesting to move the forward and reverse zone definitions into there? Because if not you'd need to add even more code just for that, to match the subnets to the forward and reverse zones (which IMO, adds fragility and additional complexity, for some LOC reduction)

Also, you cannot strictly make any assumptions about what the forward and reverse zones in use are, I'm pretty sure I've already seen some other issues talking about classless rdns zones for ipv4.

Neither, again, is the assumption that the forward zone(s) are exactly one level below the hostname. (I.e. depending on dns server imementation, this may be required to be specified as an exact zone that exists.)

That being said, if there are even more objections, the server and tsig keys could be moved to the zone definitions. (Which actually more closely aligns with the kea config format)

But for now I'll leave it as it is.

Wrong-Code:
I would like to leave a small note of support from a user perspective.

In my setup, OPNsense handles routing and DHCP while DNS is provided externally (Technitium DNS). ISC DHCP's DDNS support integrates well with that arrangement, and it has been reliable for a long time.

With the announced EOL of ISC DHCP, I've tried to switch to Kea, but once I discoverd it doesn't support DDNS I had to go back to ISC. DNS registration for DHCP lease in an important feature in my environment, and this PR seems to address that gap. It would make the migration path much easier for setups like mine.

I completely understand the need to keep things maintainable and not overload the UI. Having this functionality available, even in an advanced or optional section, would already be very helpful for users who rely on DDNS.

Thank you to everyone involved for the work so far - I appreciate the discussion and hope this can move forward at whatever pace feels right for the project.

bkshogunlord:
@Monviech , would it be acceptable to add a toggle in the Kea DHCP service for this DDNS feature?

What are the remaining blockers that @ruifung need to address before this PR is acceptable and can be merged?

Monviech:
In my opinion the scope is still too large with too many knobs exposed. I would start with feature parity with the ISC GUI since thats where most users want to migrate from.

KEA features need to be very strict in scope because it is so brittle in general. The more knobs are exposed the higher the chance of misconfiguring something thats hard to validate. This puts maintainance burden on the core code.

The less code there is, and the tighter the scope is, the higher the chance it will make it in (without promising anything).

ruifung:
And I've made my opinions clear on the last bit where @Monviech wanted the tsig key and DNS zone in the DHCPv4/v6 subnet configuration, that it makes far too many assumptions about the way the DNS zones are being used.

For example, generating the reverse zones from the subnet, which is often not the same as the zone set in the DNS server (1.168.192.in-addr.arpa vs 168.192.in-addr.arpa), which MAY be an issue for certain DNS servers.

Also, most of the kea code more or less, aligns with the kea config structure, this really doesn't belong in subnet configuration in my opinion.

AdSchellevis:
We'll just keep the PR open for now, but given the amount of code involved, it's unlikely we can merge it any time soon. 

A smaller PR has more chances of being included (as already mentioned by @Monviech ) given the time needed to review, test and document these. If there are options to substantially limit the amount of code needed to add the feature, feel free to ping us again. 

There's always a hidden cost in adding (lots of) code in terms of long term maintenance, in which case we have to decide where to spend our time and money.

bkshogunlord:
Since the Kea DHCP Server is integrating with DNS via DDNS, it seems valid to create fields for TSIG key and DNS Reverse zone within the Kea DHCP subnet configurations, regardless of any users' DNS zone setups. The section could be a collapsible DDNS subsection.

It also seems necessary to elaborate on what is meant by "knobs" in this PR context to help with addressing those concerns.

djagoo:
We are currently planning a new deployment and need the ability to use DDNS to update an existing DNS server at the customer's site. Could anyone provide an estimated timeframe for when this might be implemented in Kea? Alternatively, is there a workaround to deploy Kea and inject custom lines into the generated config? I would really prefer to start with Kea immediately rather than deploying ISC DHCP now and migrating a few months later.

AdSchellevis:
@djagoo https://github.com/opnsense/core/pull/9401#issuecomment-3706886327 should be pretty clear.

ruifung:
I'll see what I can do to reduce the LOC count, but I don't think folding tsig keys into the subnet config is a good idea for one reason:

The kea config expects a top level tsig-keys config element to define every tsig key used. This poses additional complexity for validation because the keys are identified by key name, which MUST match the key name on the DNS server.

Allowing the tsig key to be specified at the subnet level means that all tsig keys from all subnets need to be collected, deduplicated by key name, and possibly checked for mismatches in the secret value (or some other defined behaviour for this scenario, which should be quite common). I can only see this as making the validation of tsig fields and the generation of kea config more complicated.

Also, if the kea-ddns page is removed, where would I stick the "manual configuration" checkbox for kea-dhcp-ddns?

P.S. I am, indeed, taking my time with this, because 
(a) I did this initially for my own use, which has already been satisfied since I can run my modified code on my own installations. 
(b) I'm doing this in my limited free time, and my free time has much competition for it. 
(c) Stripping out any further "knobs" beyond what has already been done really kills my motivation to work on this because those so-called knobs, are knobs I actively use. (i.e. I would be making this something I would actively be annoyed by.)

@djagoo you need to modify the keactrl config template (in /usr/local/opnsense/service/templates/OPNsense/Kea) in order to enable the kea-dhcp-ddns service, then you can put in your own kea-ddns configuration, and under kea advanced settings, you can disable config generation for kea to modify the kea config files directly.

ruifung:
By removing the shared DNS server configuration (And including it into zone config), as well as adding a custom KeaFQDNField type to handle the end-with-dot validation, I've somewhat reduced the size of the PR from the initial state of +961 lines down to the current state of +706 lines.

This is an approximately 26.5~% reduction in LOC count of the PR.

@AdSchellevis Does this constitute a significant enough reduction in the size of this PR?

AdSchellevis:
it's still pretty huge, which means quite some review time, which currently isn't on the top of our priority list. If this is in your opinion the minimal viable product, just leave it open and maybe we can review it at some point in time.

brendanbank:
Thanks @ruifung for the work on this — I've been running Kea DDNS in production on my own fork and I'm happy to  abandon that fork if we can make progress in an other way.

@AdSchellevis @Monviech — would you be open to a `plugins_run()` hook approach to keep the core change minimal? The idea: a tiny core PR (~25 lines) that adds generic hook points, with all the DDNS logic in a separate plugin (`os-kea-ddns`).

The core change would be two hook points:

In `KeaDhcpv4.php::generateConfig()` (before `json_encode`):
```php
foreach (plugins_run('kea_dhcpv4_ddns_config') as $overlay) {
    if (isset($overlay['global'])) {
        $cnf['Dhcp4'] = array_merge($cnf['Dhcp4'], $overlay['global']);
    }
    if (isset($overlay['subnets'])) {
        foreach ($cnf['Dhcp4']['subnet4'] as &$subnet) {
            if (isset($overlay['subnets'][$subnet['subnet']])) {
                $subnet = array_merge($subnet, $overlay['subnets'][$subnet['subnet']]);
            }
        }
        unset($subnet);
    }
}
```

In `kea.inc::kea_configure_do()` (for kea-dhcp-ddns config generation + keactrl toggle):
```php
$ddns_overlays = plugins_run('kea_ddns_generate');
$ddnsEnabled = !empty($ddns_overlays);
```

That's it for core. Everything else — the DDNS model, controllers, views, TSIG key management, kea-dhcp-ddns config generation — lives in the plugin using the existing `plugins_run` pattern. This is the same mechanism already used in core: Kea exports `static_mapping` via `kea_run()` which Unbound and Dnsmasq consume via `plugins_run('static_mapping')` to register DHCP hostnames, and dpinger exports `return_gateways_status` / `host_routes` the same way. The plugin's model references core subnets via `ModelRelationField` (cross-model, same pattern already used elsewhere).

This way core gets a ~25 line generic hook that's useful beyond just DDNS, and the DDNS feature ships as an installable plugin with no core LOC burden.

I'd be happy to create a proof of concept based on @ruifung's code if you think this approach could work.

Best,

 - Brendan

ruifung:
@brendanbank You missed one hook point for dhcpv6.

But this is an intresting idea? If I understand you correctly, this is a hook point for plugins to be able to inject additional properties into the kea configuration?

I'm not certain how generic the hook will be if the keactrl config enables it based on the presence of any overlays.

AdSchellevis:
@brendanbank 

> would you be open to a plugins_run() hook approach to kea..

Not at the moment as it has a high risk of causing difficult to track support issues and doesn't align well with the object oriented code in this scope.

brendanbank:
@AdSchellevis 

> Not at the moment as it has a high risk of causing difficult to track support issues and doesn't align well with the object oriented code in this scope.

Thanks for the feedback — understood, we'll wait for the right time.

@ruifung Good point about the DHCPv6 hook. I went ahead and built the proof of concept to validate the approach — it's working in production on my firewall with 7 subnets doing forward and reverse DDNS updates.

The core hooks are at [brendanbank/opnsense-core `feature/kea-ddns-hooks`](https://github.com/brendanbank/opnsense-core/tree/feature/kea-ddns-hooks) (~25 lines changed) and the plugin is at [brendanbank/opnsense-plugins `feature/kea-ddns`](https://github.com/brendanbank/opnsense-plugins/tree/feature/kea-ddns/net/kea-ddns).

dBitech:
Reworking this as a plugin as @brendanbank suggests sounds like good middle ground until such time as a final version is accepted.  I've been running the plugin version for a couple weeks now and besides a few minor items related to logging etc, it seems to be stable, and covers off my use case perfectly.

brendanbank:
Thanks @dBitech — really glad to hear it's working for your setup.

@AdSchellevis @Monviech — small update. After hitting a real-world issue where the earlier plugin version called `mwexec()` during boot (removed in 26.1.3, which bricked a firewall), I reworked the plugin to use `plugins_configure('kea_sync')` instead. This requires zero core changes — the plugin just post-processes the Kea JSON configs after core generates them.

That boot failure was a humbling reminder of exactly the kind of support risk you flagged when I proposed the `plugins_run()` hooks. You were right to push back on that.

The rewritten plugin is at [banzai-plugins](https://github.com/brendanbank/banzai-plugins) with signed packages for anyone who needs DDNS in the meantime. It's no substitute for a proper core integration like @ruifung's work here, but it fills the gap for now.

ruifung:
Out of curiosity, does it actually work with HA sync,  or does it post-process on the other end.

Regardless, this PR -may- appear inactive, but thats because I'm not going to rebase it that often (it's really annoying to sign so many commits with a securoty key!)

That being said, I still am tracking opnsense stable updates on a separate branch (where I've squashed everything) and I'm still actively running this code on my own installations.

brendanbank:
Thanks for bringing this up @ruifung. I looked into this and it does work with HA sync, though it required a small addition.

v2.0_2 (just released) adds `kea_ddns_xmlrpc_sync()` which registers `OPNsense.KeaDdns` for XMLRPC sync. "Kea DDNS" now appears as a checkbox in the HA sync settings alongside "Kea DHCP".

The flow on the backup: config XML syncs over, then when Kea restarts, the rc.conf.d `kea_setup` calls `pluginctl -c kea_sync`, which fires both `kea_configure_do()` (core) and `kea_ddns_configure_do()` (plugin) — so the post-processing runs on the backup the same as on the primary.

One thing I noticed: core Kea itself seems to have the same limitation — config sync alone doesn't regenerate the JSON files, services need to restart for configs to take effect. Is that your understanding as well, or am I missing something?

fichtner:
After a lot of back and forth externally and internally we decided to offer a minimal implementation matching isc-dhcp's features in OPNsense via https://github.com/opnsense/core/pull/9923

All help testing and extending this starting point is welcome.

I'm closing this PR since it's conflicting now. Thanks @ruifung for submitting and everyone else participating.


Cheers,
Franco

Issue #9411, opened 2025-11-17
Login fails silently on multiple PHPSESSID cookies


darkk:
**Important notices**

- [x] I have read the contributing guide lines at https://github.com/opnsense/core/blob/master/CONTRIBUTING.md
- [x] I am convinced that my issue is new after having checked both open and closed issues at https://github.com/opnsense/core/issues?q=is%3Aissue

**Describe the bug**

If the browser sends two `PHPSESSID` cookies (coming from two different scopes), OPNsense is doomed as session can't be initialized properly. The login fails silently.

I've observed that several times. One box was `opnsense: 24.7.12_4 -> 25.1`, another one was updated more than once. The browser I used today is a reasonably fresh Firefox 145.0.

**To Reproduce**

1. Get two `PHPSESSID` cookies in the jar. How? That's another issue to investigate :)
2. Go to `https://10.0.2.1/` or `https://[fd41:bb86:195d:2241:1614:4bff:feb3:5680]/`
3. Enter login & password
4. `[Press enter]`
5. `POST` is redirected back to /index.php
6. ✔️ `audit 9712 - [meta sequenceId="105190"] user root authenticated successfully for WebGui [using OPNsense\Auth\Services\WebGui + OPNsense\Auth\Local]`
7. ✔️ `audit 9712 - [meta sequenceId="105191"] /index.php: Successful login for user 'root' from: 10.91.66.130`
8. 💥 `/index.php` shows login form once again, as if nothing happened at all

**Describe alternatives you considered**

- Private browsing works just fine.
- Cleaning `/var/lib/php/sessions` does not help.

**Screenshots**

`POST /` sets one cookie on top of existing ones:

<img width="1289" height="784" alt="Image" src="https://github.com/user-attachments/assets/015be4f5-370f-49b4-bb69-2dc0033e3043" />

`GET /index.php` overrides another cookie to avoid session fixation, but the browser still sends two of them.

<img width="1289" height="784" alt="Image" src="https://github.com/user-attachments/assets/39e4b6e8-d5d8-45f1-9e66-e1a815b2b5de" />

**Additional context**

I don't know what exactly triggers two `PHPSESSID` cookies to be stored by Firefox. That's another bug worth investigating, however my friend reported similar issue with their OPNsense box for road-warrior-to-office VPNs, so it's probably not just my machine 😆 

[...]

PR #9412, opened 2025-11-17
Warn about multiple PHPSESSID being sent by the browser

darkk:
That improves UX in #9411 cases: warning is better than silence.

### DIFF
--- a/src/www/authgui.inc
+++ b/src/www/authgui.inc
@@ -129,9 +129,36 @@ function check_security_http_referer_enforcement()
     return false;
 }
 
+function session_id_cookies_count()
+{
+    // If the browser persisted two PHPSESSID cookies due to scope difference and sends us both,
+    // we're kinda doomed.  That's a real header from Firefox 145.0:
+    // Cookie: \
+    //   PHPSESSID=98f5c2560301e96be907155747cd115c; \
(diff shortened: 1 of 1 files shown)


src/www/authgui.inc
```diff
@@ -129,10 +129,37 @@ function check_security_http_referer_enforcement()
...
+}
+
 function session_auth()
 {
     global $config;
 
+    // Two+ session_id() cookies confuse _SESSION management. They should go before we continue.
+    // It's hard to reset them from the server-side as we don't know the right scope with confidence.
+    if (($c = session_id_cookies_count()) >= 2) {
+        auth_log(sprintf("%d %s cookies from %s, no way to proceeed", $c, session_name(), $_SERVER['REMOTE_ADDR']), LOG_NOTICE);
```
darkk, line 159: One one hand, it opens a way to flood logs of the OPNsense box with curl. So, maybe, these log lines should be added to `/usr/local/opnsense/scripts/syslog/lockout_handler`.

On another hand, `lockout_handler` does nothing about log lines like `sshd-session 79000 - [meta sequenceId="135327"] Connection closed by 10.91.66.130 port 34902`, so it seems to be focused on nothing but authentication attempts. There is already a way to flood audit log with TCP connection attempts to OpenSSH daemon anyway. :slightly_smiling_face:

Issue #9396, opened 2025-11-15
Make udp.* timeout confgurable per-rule, like tcp.established


darkk:
- [x] I have read the contributing guide lines at https://github.com/opnsense/core/blob/master/CONTRIBUTING.md
- [x] I am convinced that my issue is new after having checked both open and closed issues at github;
  - #5543 is related, but it's focused on default timeouts, not per-service ones

**Is your feature request related to a problem? Please describe.**

[Our hackerspace](https://0x08.in/) runs [one of ntppool's](https://www.ntppool.org/scores/83.243.68.157) Stratum 1 NTP servers. OPNsense acts as a "default gateway" and NATs external NTP traffic to a Stratum 1 box named `timekeeper`.  Every 15min NTP server gets a traffic spike from ~the thundering herd~ misconfigured SNTP clients. The spike leads to connectivity hiccup that manifests itself as `[zone: pf states] PF states limit reached` in `dmesg`. The spike is not huge, it's just 35'000 pps for a couple of seconds. The router has 4 GiB of RAM and allocates 393'400 states for the state table.

PF timeout settings are the default ones: `{ udp.first 60, udp.single 30, udp.multiple 60 }`.

Each NTP exchange allocates two state entries those linger for 60s. 35'000 pps spike overflows the table quickly and the hiccup lasts for a minute.

The traffic spike is a "normal" thing due to cron-based SNTP clients ~mis~configured to run every `*/15 * * * *`. It's not something we can realistically amend.

**Describe the solution you like**

OPNsense should provide ability to tune UDP timeouts per-rule as it already does for `tcp.established`. Having `udp.{first,single,multiple}` timeouts lowered down to 2…3 seconds for NTP traffic would not provide a bulletproof solution, but will make migration between configurations easier.

**Describe alternatives you considered**

Per-port stateless one-to-one NAT is not available in PF to the best of my knowledge. I mean "nat" rule can't have "no state" per `man pf.conf`.

Port forwarding with "no state" is not an option as well. I mean "rdr" rule can't have "no state" either.

Possible option is to have `dummy` interface at the internal `timekeeper` box and `route-to` at OPNsense. It'll kill NAT for NTP traffic at OPNsense altogether, but it'll make services at the hair-pinned IP unreachable from the `timekeeper` without extra effort.

I'd also be glad to hear about other options I have to handle the traffic spike.

**Additional context**

Software version used and hardware type if relevant:
 - OPNsense 25.7.7_4 (amd64).
[...]

PR #9414, opened 2025-11-18
Firewall: Rules [new]: Add per rule state timeouts for udp.first, udp.multiple and udp.single

Monviech:
Fixes: https://github.com/opnsense/core/issues/9396

### DIFF
--- a/src/opnsense/mvc/app/controllers/OPNsense/Firewall/forms/dialogFilterRule.xml
+++ b/src/opnsense/mvc/app/controllers/OPNsense/Firewall/forms/dialogFilterRule.xml
@@ -285,7 +285,7 @@
     </field>
     <field>
         <type>subheader</type>
-        <label>Policy</label>
+        <label>State options</label>
         <advanced>true</advanced>
     </field>
     <field>
@@ -326,7 +326,7 @@
     </field>
     <field>
         <type>subheader</type>
-        <label>Timeouts</label>
+        <label>State timeouts</label>
         <advanced>true</advanced>
     </field>
     <field>
@@ -391,7 +391,7 @@
     </field>
     <field>
         <type>subheader</type>
-        <label>Limits</label>
+        <label>State limits</label>
         <advanced>true</advanced>
     </field>
     <field>


src/opnsense/mvc/app/controllers/OPNsense/Firewall/forms/dialogFilterRule.xml
```diff
@@ -309,9 +309,39 @@
     </field>
     <field>
         <id>rule.statetimeout</id>
-        <label>State timeout</label>
+        <label>TCP established</label>
```
fichtner, line 334: It would be nice to have headers for different types of advanced options since these all tweak timeouts for example, but I don't know how doable that is given this is a selection of misc options.

Monviech: So far these are all under the "Stateful Firewall" header, maybe subheaders could be something to think about though I don't know how convoluted that would become.

fichtner: maybe not a real subheader with collapsible functionality but just a subheading basically?

Monviech: how about something like this

image

image

src/opnsense/mvc/app/models/OPNsense/Firewall/Filter.xml
```diff
@@ -169,6 +169,15 @@
...
                 </statetimeout>
+                <udp-first type="IntegerField">
+                    <MinimumValue>1</MinimumValue>
+                </udp-first>
+                <udp-multiple type="IntegerField">
+                    <MinimumValue>1</MinimumValue>
+                </udp-multiple>
+                <udp-single type="IntegerField">
+                    <MinimumValue>1</MinimumValue>
+                </udp-single>
```
AdSchellevis, line 184: We might want to set reasonable max values here (65535?), although I haven't tried setting a php max value in it, I can image this runs into issues somewhere.

Monviech: I think timeouts are `u_int32_t` but I can do 16bit as well.

darkk: > 65535

Default `tcp.established` is 86400. "Conservative" `tcp.established` is 432000.

So, 65535 feels like an sub-optimal for form validation purposes.

Just my ¢¢.

AdSchellevis: @darkk good point, not sure what a decent upper bound would look like for these, but limiting it to u_int32_t might be good enough. I which case the max should be 4294967295 (which I believe is still smaller than `PHP_INT_MAX`)

darkk: I'd suggest to use `0x7FFFFFFF` (2147483647, `PHP_INT_MAX` for 32-bit builds), not to `0xFFFFFFFF` &mdash; just to be on a safe side with regards to signed/unsigned int. Anyway, timeout of 68 years makes as little sense as timeout of 136 years. :slightly_smiling_face:

src/opnsense/mvc/app/controllers/OPNsense/Firewall/forms/dialogFilterRule.xml
fichtner: Thanks, that looks good. The header has room for an icon and proper alignment with the options and some other things that bother me but I have to bring that up in the next meeting if I don’t forget (it’s not a super important style issue with the fact that everything is using “strong” making the headings much less prominent).

Monviech: This should make the subheader distinct enough, I really like this FA-Icon, it makes it look like its part of the header, while also being at the exact line the arrow points down.

image

I'll merge this state and we can adjust it later if needed.

Issue #9422, opened 2025-11-19
firewall_rule_lookup fails for rules with id=0


darkk:
**Important notices**

- [x] I have read the contributing guide lines at https://github.com/opnsense/core/blob/master/CONTRIBUTING.md
- [x] I am convinced that my issue is new after having checked both open and closed issues at https://github.com/opnsense/core/issues?q=is%3Aissue
  - #5126 is the bug of "redirect generator" issue not firewall_rule_lookup
  - #7609 has certain similarity from the user's point of view, but it's still a different one

**Describe the bug**

Click the "lookup rule" link, get the blank page. Sounds frustrating.

**To Reproduce**

1. Create some floating rule.
2. Go to Firewall / Rules / wan
3. Open [Floating rules] dropdown
4. Click the very first rule in the list
5. Be faced with a blank "200 OK" page of `content-length: 0`

**Expected behavior**

Opening `https://10.0.2.1/firewall_rules_edit.php?if=FloatingRules&id=0` would be nice.

**Additional context**

See the PR #9423. It's an ugly fix, but it's unclear if `/firewall_rules_edit.php` deserves more given ongoing migration to MVC.

**Environment**

OPNsense 25.7.7_4 (amd64)

PR #9423, opened 2025-11-19
Fix firewall_rule_lookup for fw rules with id=0

darkk:
See #9422

### DIFF
--- a/src/www/firewall_rule_lookup.php
+++ b/src/www/firewall_rule_lookup.php
@@ -50,9 +50,16 @@ if ($_SERVER['REQUEST_METHOD'] === 'GET') {
                     parse_str($parts['query'], $query);
                     if (!empty($parts['fragment'])) {
                         header(url_safe('Location: /%s?if=%s#%s', [$parts['path'], $query['if'], $parts['fragment']]));
-                    } elseif (!empty($query['if']) && !empty($query['id'])) {
+                    } elseif (strlen($query['if']) && strlen($query['id'])) {
                         // firewall index reference
-                        header(url_safe('Location: /%s?if=%s&id=%s', [$parts['path'], $query['if'], $query['id']]));
+                        $loc = url_safe('Location: /%s?if=%s&id=%s', [$parts['path'], $query['if'], $query['id']]);
+                        // `"0"` is a valid rule `"id"`, empty() check in url_safe() converts it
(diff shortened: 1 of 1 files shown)


AdSchellevis:
we'll keep it simple in https://github.com/opnsense/core/commit/35dc4de80ce1017702eb2ef77191231462ff6fde, not sure what the future is for this script, but long term we may have to come up with something else anyway.

darkk:
I suspect https://github.com/opnsense/core/commit/35dc4de80ce1017702eb2ef77191231462ff6fde is not enough as `url_safe()` will make `""` out of `"0"` down the road :-(

https://github.com/opnsense/core/blob/7ee3b2c516590bae918532ba406960ef2bf426db/src/etc/inc/util.inc#L919

darkk:
Unfortunately, due to PHP idiosyncratic handling of types, that calls straight to WATMAN, just like [JavaScript does](https://www.destroyallsoftware.com/talks/wat).

```
root@uxie:~ # echo '<? print(empty("1") ? "TRUE" : "FALSE"); ?>.' | /usr/local/bin/php-cgi
Content-type: text/html; charset=UTF-8

FALSE.

root@uxie:~ # echo '<? print(empty(0) ? "TRUE" : "FALSE"); ?>.' | /usr/local/bin/php-cgi
Content-type: text/html; charset=UTF-8

TRUE.

root@uxie:~ # echo '<? print(empty("0") ? "TRUE" : "FALSE"); ?>? WUT!' | /usr/local/bin/php-cgi
Content-type: text/html; charset=UTF-8

TRUE? WUT!
```

<details>
  <summary>WATMAN.jpg</summary>

WATMAN

</details>

AdSchellevis:
ok, let's try to fix the actual issue then ;) https://github.com/opnsense/core/commit/7fe42e78c5bd4fdfbb3c95bbe29caf9856a6da43

darkk:
@AdSchellevis thanks! I was tadbit afraid of touching `url_safe()` given the number of callers of the function :-)

AdSchellevis:
@darkk I understand, there might be some small side affects, although I don't expect a lot of callers try to push empty values into the function anyway.

fichtner:
This appears to date back to 7f230df702420fb and hasn't caused much issues since then although it has masqueraded the "0" case to no detrimental effect to UX. The new fix is much more appropriate and will go into 25.7.8.

Issue #9425, opened 2025-11-19
firewall: Rules with inverted interfaces in Firewall: Automation: Filter display in the wrong group


ruifung:
**Important notices**

Before you add a new report, we ask you kindly to acknowledge the following:

- [X] I have read the contributing guide lines at https://github.com/opnsense/core/blob/master/CONTRIBUTING.md
- [X] I am convinced that my issue is new after having checked both open and closed issues at https://github.com/opnsense/core/issues?q=is%3Aissue

**Describe the bug**

In the Firewall: Automation: Filter view, rules with a Invert Interface checked still get displayed under the interface that it's not supposed to be in.

**To Reproduce**

1. Go to 'Firewall: Automation: Filter'
2. Add a rule
3. Check 'Invert Interface'
4. Select an interface or group
5. See that it's displayed under the interface or group it's not supposed to apply to.

**Expected behavior**

Expected it to display under 'Floating' which is consistent with rules that have multiple interfaces selected.

**Describe alternatives you considered**

N/A

**Screenshots**

<img width="580" height="247" alt="Image" src="https://github.com/user-attachments/assets/48ecc3b4-1428-4bf1-822b-142c4ac870e7" />

**Relevant log files**

N/A

**Additional context**

N/A

**Environment**

Software version used and hardware type if relevant, e.g.:

OPNsense 25.7.7 (amd64).

PR #9426, opened 2025-11-19
Firewall: Rules [new]: Fix handling of interfacenot, evaluate as floating rules in correct prio_group

Monviech:
Fixes: https://github.com/opnsense/core/issues/9425

A bit unsure if I got all spots, but this seems to work.

@AdSchellevis please correct this one as needed.

### DIFF
--- a/src/opnsense/mvc/app/controllers/OPNsense/Firewall/Api/FilterController.php
+++ b/src/opnsense/mvc/app/controllers/OPNsense/Firewall/Api/FilterController.php
@@ -108,10 +108,13 @@ class FilterController extends FilterBaseController
                 $is_if = count($rule_interfaces) != 1 || !$record->interfacenot->isEmpty();
             } elseif ($show_all) {
                 $is_if = $if_intersects || empty($rule_interfaces);
+            } elseif (!$record->interfacenot->isEmpty()) {
+                // Exclude as it should only be returned with show_all
+                $is_if = false;
             } else {
-                $is_if = $if_intersects;
+                // Include only an exact match, not a partial overlap
+                $is_if = $if_intersects && (count($interfaces) == count($rule_interfaces));
             }
-
             return $is_cat && $is_if;
         };
 


Monviech:
I fixed something small and now it reflects reality in my test:

Empty Interface, Inverted Group, Inverted Interface, Multiple Interface -> All in floating
Single group -> Is in group
Single interface -> Is in interface

image

@fichtner Should I wait on this one or is it okay for master?

src/opnsense/mvc/app/controllers/OPNsense/Firewall/Api/FilterController.php
```diff
@@ -178,6 +178,10 @@ public function searchRuleAction()
             $is_cat = empty($categories) || array_intersect(explode(',', $record->categories), $categories);
             $rule_interfaces = array_filter(explode(',', (string)$record->interface));
 
+            if ((string)$record->interfacenot === "1") {
+                $rule_interfaces = [];
```
fichtner: if we're strict we need to get a full list of interfaces now excluding the ones that are in there. This way nothing else needs to be changed, but the list could be long...

EDIT: things are a bit simpler due to restrictions in GUI and validations of `interfacenot`. A comment would be best.

fichtner:
```suggestion
                /*
                 * XXX Block the search from matching the interface in this
                 * case.  It's not entirely correct, but at the moment we are
                 * not able to search for !interface either.
                 */
                $rule_interfaces = [];
```

AdSchellevis: basically this:

https://github.com/opnsense/core/blob/c6eaefc1e26550085b09abf1cca73fdc8f76db13/src/www/firewall_rules.php#L802-L804

When "floating" these should aways be visible, when an interface (or a group of interfaces) is selected, all but the ones specified. In our cases (with `$interfaces`) this means `array_intersect` on both sets when inverse is set.

src/opnsense/mvc/app/models/OPNsense/Firewall/FieldTypes/FilterRuleField.php
```diff
@@ -108,8 +108,13 @@ public function getPriority()
...
         $configObj = Config::getInstance()->object();
         $interface = (string)$this->interface;
-        if (strpos($interface, ",") !== false || empty($interface)) {
-            // floating (multiple interfaces involved)
+
+        if (
+            // floating: empty, multiple, or inverted interface
+            (string)$this->interfacenot === "1" ||
+            (strpos($interface, ",")) !== false ||
+            empty($interface)
```
AdSchellevis: this doesn't seem to match our actual sorting

https://github.com/opnsense/core/blob/c6eaefc1e26550085b09abf1cca73fdc8f76db13/src/etc/inc/filter.lib.inc#L675-L681

src/opnsense/mvc/app/controllers/OPNsense/Firewall/Api/FilterController.php
```diff
@@ -221,6 +225,8 @@ public function searchRuleAction()
 
             if (empty($interfaces)) {
                 $is_if = empty($record['interface']) || count(explode(',', $record['interface'])) > 1;
+            } elseif ((string)$record['interfacenot'] === "1") {
+                $is_if = true;
```
fichtner:
```suggestion
                /* XXX given that $interfaces is deduplicated multiple interfaces unmute the match */
                $is_if = empty($record['interface']) || count($interfaces > 1) || $record['interface'] != $interfaces[0];
```

fichtner: FWIW, I'd like to restructure the logic checks in this block since they are a mostly upside down by checking the same things in conditionals.

src/opnsense/mvc/app/models/OPNsense/Firewall/FieldTypes/FilterRuleField.php
fichtner: https://github.com/opnsense/core/commit/be0b18930f36f9 says it's a floating rule pinned to one interface so this seems correct?

src/opnsense/mvc/app/controllers/OPNsense/Firewall/Api/FilterController.php
```diff
@@ -221,6 +230,9 @@ public function searchRuleAction()
 
             if (empty($interfaces)) {
                 $is_if = empty($record['interface']) || count(explode(',', $record['interface'])) > 1;
+            } elseif ((string)$record['interfacenot'] === "1") {
+                /* XXX given that $interfaces is deduplicated multiple interfaces unmute the match */
+                $is_if = empty($record['interface']) || count($interfaces > 1) || $record['interface'] != $interfaces[0];
```
fichtner: did I write this? 😆 


```suggestion
                $is_if = empty($record['interface']) || count($interfaces) > 1 || $record['interface'] != $interfaces[0];
```

Monviech: I didnt see this either xd

src/opnsense/mvc/app/models/OPNsense/Firewall/FieldTypes/FilterRuleField.php
fichtner: I changed the file a bit in 37e5056446a3f 51005211e1d6

For now we should limit floating to this condition and keep sorting single-floating rules to their interface-specific ordering location

The whole floating concept needs to die and adapting it further will just play ping pong with rules that work how they work given their current ordering (if it's correct or not is debatable)

src/opnsense/mvc/app/controllers/OPNsense/Firewall/Api/FilterController.php
```diff
@@ -401,10 +491,10 @@ public function getInterfaceListAction()
         // Count rules per interface
         $ruleCounts = [];
         foreach ((new \OPNsense\Firewall\Filter())->rules->rule->iterateItems() as $rule) {
-            $interfaces = $rule->interface->getValues();
+            $interfaces = array_filter(explode(',', (string)$rule->interface));
 
-            if (count($interfaces) !== 1) {
-                // floating: empty or multiple interfaces
+            if ((string)$rule->interfacenot === "1" || count($interfaces) !== 1) {
```
fichtner:
```suggestion
            if ($rule->interfacenot->isEqual(1) || count($interfaces) !== 1) {
```

src/opnsense/mvc/app/models/OPNsense/Firewall/FieldTypes/FilterRuleField.php
Monviech: There are no single interface floating rules possible here. The condition right now should be correct.

fichtner: `(string)$this->interfacenot === "1" ||` but this adds single interface ones ;)

Monviech: Kinda, but only inverted single interface ones, which counts as multiple interfaces implicitely?

I guess inverting "LAN" makes it an infinite set of interfaces that exludes LAN, which counts as multiple?

fichtner: true. given that the situation is what it is right now and we don't have any reports -- fixing a spot and missing several others is going to make it worse likely.

Issue #9274, opened 2025-10-08
OpenVPN: Feature request (new Instances) add Option fast-io


mdten:
**Important notices**

Before you add a new report, we ask you kindly to acknowledge the following:

- [ x ] I have read the contributing guide lines at https://github.com/opnsense/core/blob/master/CONTRIBUTING.md
- [ x ] I am convinced that my issue is new after having checked both open and closed issues at https://github.com/opnsense/core/issues?q=is%3Aissue

**Is your feature request related to a problem? Please describe.**

In the process of migrating the openVPN config from "legacy" to the "new Instances", I am missing the option "fast-io" that I had entered under "Advanced configuration" - which brought us a significant performance gain.

![Image](https://github.com/user-attachments/assets/e98c2e88-b668-42d9-9c39-778cf603f7c6)

**Describe the solution you like**

Add the option "fast-io" to Miscellaneous > Options in openVPN config of new Instances.

![Image](https://github.com/user-attachments/assets/1317e60a-52cb-4bad-ba7a-d499c356c75a)

**Describe alternatives you considered**

A clear and concise description of any alternative solutions or features you considered.

**Additional context**

https://openvpn.net/community-docs/community-articles/openvpn-2-6-manual.html

![Image](https://github.com/user-attachments/assets/fd836b5f-b3a4-413d-b82c-60c6456637c8)

AdSchellevis:
https://github.com/opnsense/core/pull/9430

PR #9430, opened 2025-11-20
Update OpenVPN.xml add Option fast-io

mdten:
OpenVPN  new Instances add Option fast-io 

add Feature #9274

### DIFF
--- a/src/opnsense/mvc/app/models/OPNsense/OpenVPN/OpenVPN.xml
+++ b/src/opnsense/mvc/app/models/OPNsense/OpenVPN/OpenVPN.xml
@@ -281,6 +281,7 @@
                         <route-noexec>route-noexec</route-noexec>
                         <route-nopull>route-nopull</route-nopull>
                         <explicit-exit-notify>explicit-exit-notify</explicit-exit-notify>
+                        <fast-io>fast-io</fast-io>
                     </OptionValues>
                 </various_flags>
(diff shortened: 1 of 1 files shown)


AdSchellevis:
@mdten thanks

AdSchellevis:
hmm, I was a bit too fast to merge it seems. according to the manual it should validate on:

> This option can only be used on non-Windows systems, when --proto udp is specified, and when --shaper is NOT specified.


@mdten do you plan to add that as well?

AdSchellevis:
@mdten nevermind, I've added https://github.com/opnsense/core/commit/5f3f2b2b6a3ba70120fcb99a1042da31e1c5c807 on top, if you can test and confirm that would be appreciated.

mdten:
@AdSchellevis Thanks.
It would have taken me forever to find the right file and create the check function.

But I gladly tested it.

I installed the 2 patchs and cloned my instance.

opnsense-patch effe0aa 
opnsense-patch 5f3f2b2

If I have selected TCP or DCO, or both, I get the messages.

fast-io_tcp
fast-io_dco
fast-io_tcp_dco


If the instance is on UDP, I can save it with the option fast-io.

AdSchellevis:
@mdten ok, thanks, sounds like we have a winner then :)

fichtner:
@mdten since you were nice enough to test the code I've added the patch to the upcoming 25.7.8 release

PR #9450, opened 2025-11-24
dhcp/kea: Refactor KeaDhcpv4.php and KeaDhcpv6.php to use BaseField helper methods

Monviech:


src/opnsense/mvc/app/models/OPNsense/Kea/KeaDhcpv6.php
```diff
@@ -199,7 +199,6 @@ private function getExpiredLeasesProcessingConfig()
         $config = [];
         $lexpireFields = iterator_to_array($this->lexpire->iterateItems());
         foreach ($lexpireFields as $fieldName => $fieldValue) {
-            $value = $fieldValue->__toString();
```
Monviech, line 202: This appears to be unused.

src/opnsense/mvc/app/models/OPNsense/Kea/KeaDhcpv4.php
```diff
@@ -95,7 +95,7 @@ public function performValidation($validateFullModel = false)
 
     public function isEnabled()
     {
-        return (string)$this->general->enabled == '1' && !empty((string)(string)$this->general->interfaces);
```
fichtner, line 98: what's a (string)(string) :)

Monviech:
I'll thoroughly test this before merging, it's not in any rush.

I touched these files to improve consistency since thats something I noticed in this PR: https://github.com/opnsense/core/pull/9401

src/opnsense/mvc/app/models/OPNsense/Kea/KeaDhcpv4.php
```diff
@@ -260,7 +260,7 @@ public function generateConfig($target = '/usr/local/etc/kea/kea-dhcp4.conf')
             $cnf['Dhcp4']['hooks-libraries'][] = [
                 'library' => '/usr/local/lib/kea/hooks/libdhcp_lease_cmds.so'
             ];
-            if (!empty((string)$this->ha->enabled)) {
+            if (!$this->ha->enabled->isEmpty()) {
```
ruifung, line 263: Just out of curiosity (and for my open PR) is there a specific case where you'd want to use !(field->isEmpty()) vs field->isEqual('1') when checking if a boolean field is checked?

Because I've seen both being used and it's mildly confusing as to which is the preferred method for checking if a checkbox is checked.

fichtner: isEqual is a newer than isEmpty. isEqual(‘1’) or isEqual(1) is always safe, but isEqual(‘0’) had some issues in the past but is now always safe too. Though for booleans the best check to support perhaps would be isOn() and isOff() but that requires more brain smarts as it doesn’t really fit the base field. Just a tangent here since you asked 😊

Monviech:
Since I tested the before and after via these tests: https://github.com/opnsense/core/pull/9455

I can say with some certainty that they will not cause huge regressions in config generation and other vital parts.

PR #9455, opened 2025-11-26
dhcp/kea: Add unit tests, big smoke test via snapshots, a few smaller tests to catch regressions early

Monviech:
The snapshots use only real data I input to the GUI and is fully validated.

A config.xml is sent through the current model, and the resulting json is compared to the snapshot samples. If they do not match, a detailed error message is generated.

Other than that, some smaller tests are for firewall rules, enabled logic, and interface parsing.

With that in place, I can test for major regressions in here: https://github.com/opnsense/core/pull/9450

And it is also handy for the future, if somebody changes something in KEA, the tests will trigger (most likely), catching most regressions early.

If new features are added, the tests need to be maintained and expanded.


Monviech:
With the latest changes this is already outdated. A unit test approach is better, lets scrap this for now.

PR #9484, opened 2025-12-03
isc-dhcp: interalize interfaces_staticarp_configure(); closes #9476

### LINKED ISSUE #9476: ISC static ARP is too sticky, can create issues with other successor DHCP servers
**Important notices**

If Kea or Dnsmasq do not work, users tend to quickly blame the new DHCP servers. Though most of the time, an ISC setting is to blame:

https://github.com/opnsense/core/blob/e2d8d25df53b8c4869b5b204dac36670f40c746f/src/www/services_dhcp.php#L879

https://github.com/opnsense/core/blob/d0e53868c5bf5de46aaad1b9bebe5816c6988c36/src/etc/inc/interfaces.inc#L3851

https://github.com/opnsense/core/blob/d0e53868c5bf5de46aaad1b9bebe5816c6988c36/src/etc/inc/plugins.inc.d/dhcpd.inc#L170-L172

https://github.com/opnsense/core/blob/d0e53868c5bf5de46aaad1b9bebe5816c6988c36/src/etc/inc/plugins.inc.d/dhcpd.inc#L231

-> The static ARP entries should vanish as soon as ISC is disabled.

fichtner:
Instead of making the interface code pluggable, push the code that causes the persistent side effect to the ISC DHCP plugin which then gets to fix the stuck static ARP flag after disable/deinstall plus a reboot.  The situation isn't ideal, but much better than before.

### DIFF
--- a/src/etc/inc/interfaces.inc
+++ b/src/etc/inc/interfaces.inc
@@ -2454,12 +2454,10 @@ function interface_configure($verbose = false, $interface = 'wan', $reload = fal
     }
 
     interfaces_vips_configure($interface);
+    interfaces_neighbors_configure($interface, $ifconfig_details);
 
     /* XXX device member plugin hook */
     link_interface_to_bridge($interface, $device, $ifconfig_details);
-
-    /* XXX may be called twice since also tied to dhcp configure */
-    interfaces_staticarp_configure($interface, $ifconfig_details);
 
     service_log("done.\n", $verbose);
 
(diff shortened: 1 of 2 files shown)


Monviech:
I just tried it (applied the patch, deactivated ISC, rebooted)

and the static assignment seems to stick around, to be precise this one:
https://github.com/opnsense/core/blob/9e8058033a7f06dd756df1372b565b46658a06b4/src/www/services_dhcp_edit.php#L221

fichtner:
The neighbours are being added, yes. But the static interface flag should not be as well as that the arp cache should not be flushed. That's because I've left this one for future use. Otherwise neighbours stop populating completely. It's debatable.

https://github.com/opnsense/core/blob/e24d3cdac47a307a95bb3a950c0ec3e5b5b0a77d/src/etc/inc/interfaces.inc#L2457

Monviech:
The ```staticarp``` flag is gone indeed so I think this is good.

PR #9491, opened 2025-12-04
unbound: clean up marker and size files, closes #9489

### LINKED ISSUE #9489: unbound: old blocklist active marker does not get deleted when migration removed all blocklists from user
**Important notices**

Granted this is an edge case but makes our instructions to apply not work.

The workaround is to add a blocklist and apply again.

It would probably help if we remove the marker file when no blocklists are going to be rendered?

Noticed while testing a blocklist migration from 25.7 o 25.7.9.

swhite2:
This turned ino a larger shift, as there were some nitpicks:

- the size file didn't update to 0 in cases where the blocklist wasn't available
- the `self.dnsbl` variable wasn't cleaned up in cases where the state changed to unavailable

### DIFF
--- a/src/opnsense/scripts/unbound-dnsbl/lib/dnsbl.py
+++ b/src/opnsense/scripts/unbound-dnsbl/lib/dnsbl.py
@@ -66,7 +66,7 @@ class DNSBL:
             self._load_dnsbl()
 
     def _load_dnsbl(self):
-        last_state = self.dnsbl is not None
+        last_state = (self.dnsbl is not None)
 
         if not self._dnsbl_exists():
             self.dnsbl = None
@@ -96,7 +96,7 @@ class DNSBL:
         if os.path.exists(self.warn_file):
             os.remove(self.warn_file)
 
-        if last_state != self.dnsbl is not None:
+        if last_state != (self.dnsbl is not None):
             with open(self.size_file, 'w') as sfile:
                 sfile.write(str(len(self.dnsbl['data'])) if self.dnsbl else '0')
 


src/opnsense/scripts/unbound-dnsbl/lib/dnsbl.py
```diff
@@ -64,40 +64,46 @@ def _update_dnsbl(self):
         t = time.time()
         if (t - self.dnsbl_update_time) > 60:
             self.dnsbl_update_time = t
-            if not self._dnsbl_exists():
-                self.dnsbl_available = False
-                return
+            self._load_dnsbl()
+
+    def _load_dnsbl(self):
+        last_state = self.dnsbl_available
```
AdSchellevis: I think I would gather all logic around `self.dnsbl` in which case this would be something like

```
last_state = self.dnsbl is not None
```

src/opnsense/scripts/unbound-dnsbl/lib/dnsbl.py
```diff
@@ -64,40 +64,46 @@ def _update_dnsbl(self):
...
-                    return
-                else:
-                    log_err("dnsbl_module: error parsing blocklist: %s, reusing last known list" % e)
+                with open(self.dnsbl_path, 'r') as f:
+                    try:
+                        self.dnsbl = json.load(f)
+                        if self._context and type(self.dnsbl.get('config')) is dict:
+                            if not self.dnsbl['config'].get('general'):
+                                # old format, needs blocklist reload
+                                raise ValueError("incompatible blocklist")
```
AdSchellevis, line 85: before the raise `self.dnsbl = None`

src/opnsense/scripts/unbound-dnsbl/lib/dnsbl.py
```diff
@@ -64,40 +64,46 @@ def _update_dnsbl(self):
...
+                            self._context.set_config(self.dnsbl['config'])
+                        self.dnsbl_available = True
```
AdSchellevis: there is data, so it's enabled.

src/opnsense/scripts/unbound-dnsbl/lib/dnsbl.py
```diff
@@ -64,40 +64,46 @@ def _update_dnsbl(self):
...
+                            open(self.warn_file, "a").close()
+                            return
+                        else:
+                            log_err("dnsbl_module: error parsing blocklist: %s, reusing last known state" % e)
 
         if os.path.exists(self.warn_file):
             os.remove(self.warn_file)
 
-        self.dnsbl_available = True
+        if last_state != self.dnsbl_available:
```
AdSchellevis:
```suggestion
        if last_state != self.dnsbl is not None:
```

Just to bind to the single version of the truth

src/opnsense/scripts/unbound-dnsbl/lib/dnsbl.py
```diff
@@ -64,40 +64,46 @@ def _update_dnsbl(self):
...
+            with open(self.size_file, 'w') as sfile:
+                sfile.write(str(len(self.dnsbl['data'])) if self.dnsbl_available else '0')
```
AdSchellevis:
```suggestion
                sfile.write(str(len(self.dnsbl['data'])) if self. dnsbl else '0')
```

Which is actually the reason for my suggestions, if someone drops a pen at some point, there's always a chance `dnsbl_available` and `dnsbl` are not aligned anymore.

src/opnsense/scripts/unbound-dnsbl/lib/dnsbl.py
```diff
@@ -64,40 +64,46 @@ def _update_dnsbl(self):
...
+
+        if not self.dnsbl_available:
+            self.dnsbl = None
```
AdSchellevis: this can go as being reset already.

swhite2:
> some small suggestions, mainly to ease future maintenance

@AdSchellevis fully agree, thanks for having a look. https://github.com/opnsense/core/pull/9491/commits/0b86124894583874e545b2655c993e9fd212301c should do it

src/opnsense/scripts/unbound-dnsbl/lib/dnsbl.py
```diff
@@ -64,40 +63,42 @@ def _update_dnsbl(self):
...
+                            open(self.warn_file, "a").close()
+                            return
+                        else:
+                            log_err("dnsbl_module: error parsing blocklist: %s, reusing last known state" % e)
 
         if os.path.exists(self.warn_file):
             os.remove(self.warn_file)
 
-        self.dnsbl_available = True
+        if last_state != self.dnsbl is not None:
```
AdSchellevis:
```suggestion
        if last_state != (self.dnsbl is not None):
```

It might be correct as it was, I'm just doubting here.

swhite2: good catch, as it was it would be equivalent to `if last_state != self.dnsbl and self.dnsbl is not None`, the original `last_state` should be changed as well

PR #9492, opened 2025-12-05
nat/one-to-one: Refactor one-to-one nat view, refactor two controller methods as base methods

Monviech:
Note: The volt templates have quite some boilerplate copied, yet there are subtle differences between these views. Generalizing them would make maintenance hard. Best is to resist premature abstraction, as its safer to change little things about individual views without causing side-effects in other views.

The same view boilerplate will most likely be copied a few more times in the future, yet abstracting it now will bite later in my opinion.

image

This makes the One-to-One NAT view the same as DNat and Filter.

### DIFF
--- a/src/opnsense/mvc/app/controllers/OPNsense/Firewall/Api/OneToOneController.php
+++ b/src/opnsense/mvc/app/controllers/OPNsense/Firewall/Api/OneToOneController.php
@@ -41,7 +41,7 @@ class OneToOneController extends FilterBaseController
 
         $filter_funct = function ($record) use ($category) {
             /* categories are indexed by name in the record, but offered as uuid in the selector */
-            $catids = !empty((string)$record->categories) ? explode(',', (string)$record->categories) : [];
+            $catids = !$record->categories->isEmpty() ? $record->categories->getValues() : [];
             return empty($category) || array_intersect($catids, $category);
         };
 


src/opnsense/mvc/app/controllers/OPNsense/Firewall/Api/OneToOneController.php
```diff
@@ -65,4 +88,15 @@ public function toggleRuleAction($uuid, $enabled = null)
...
+    {
+        return $this->moveRuleBeforeBase($selected_uuid, $target_uuid, 'onetoone.rule.', 'sequence', 'One-to-One');
+    }
+
+    public function toggleRuleLogAction($uuid, $log)
+    {
+        return $this->toggleRuleLogBase($uuid, $log, 'onetoone.rule.', 'One-to-One');
+    }
+
+}
```
fichtner:
```suggestion
}

```

;)

src/opnsense/mvc/app/controllers/OPNsense/Firewall/forms/dialogOneToOneRule.xml
```diff
@@ -1,86 +1,160 @@
...
         <advanced>true</advanced>
         <help>Use this option to invert the sense of the match.</help>
+        <grid_view>
+            <ignore>true</ignore>
+        </grid_view>
     </field>
     <field>
         <id>rule.source_net</id>
-        <label>Source / Internal</label>
+        <label>Source (Internal)</label>
```
fichtner: Maybe we can lose the internal annotation here? it's already in the help text

src/opnsense/mvc/app/models/OPNsense/Firewall/Menu/Menu.xml
```diff
@@ -7,7 +7,7 @@
             <DNat order="100" VisibleName="Destination NAT" url="/ui/firewall/d_nat/">
                 <FilterRef url="/ui/firewall/d_nat#*" visibility="hidden"/>
             </DNat>
-            <OneToOne order="200" VisibleName="One-to-One" url="/ui/firewall/one_to_one/">
+            <OneToOne order="200" VisibleName="One-to-One NAT" url="/ui/firewall/one_to_one/">
```
fichtner, line 10: this is what they call peer pressure and now we have NAT : XXX NAT NAT NAT in the menu. Honestly it may make more sense to remove the NAT menu folder altogether.

fichtner (approved):
very nice all things considered, thanks!

src/opnsense/mvc/app/models/OPNsense/Firewall/Menu/Menu.xml
Monviech: It's all for NATing :)

src/opnsense/mvc/app/controllers/OPNsense/Firewall/forms/dialogOneToOneRule.xml
Monviech: Now that it is inside its own "Mapping" header, which clearly distincs "External network" and "Source" from Destination, I think thats ok to strip.

src/opnsense/mvc/app/controllers/OPNsense/Firewall/Api/FilterBaseController.php
```diff
@@ -293,4 +293,98 @@ public function revertAction($revision)
...
+     * Rules will be renumbered within the selected range to prevent movements causing overlaps,
+     * but try to keep the changes as minimal as possible.
+     *
+     * @param string $selected_uuid     The UUID of the rule to be moved.
+     * @param string $target_uuid       The UUID of the target rule (the rule before which the selected rule is to be placed).
+     * @param string $node_reference    Node reference prefix, e.g. "onetoone.rule".
+     * @param string $sort_key          Sort key field, e.g. "sequence".
+     * @return array Returns ["status" => "ok"] on success, throws a UserException otherwise.
+     */
+    public function moveRuleBeforeBase($selected_uuid, $target_uuid, $node_reference, $sort_key)
```
AdSchellevis, line 310: @Monviech I would have expected `moveRuleBeforeAction()` in `FilterController` to be refactored as well here, or am I missing something?

Monviech: The filter controller has some custom logic like the priority groups so I didnt touch it.

AdSchellevis: ok, clear, maybe something to refactor in the future then.

PR #9500, opened 2025-12-07
add and refactor host template helpers

BPplays:
- add new functions for use with strings instead of tags
- moved tag based functions to be based on string functions


fichtner:
I don't think we should revisit this, especially considering what I mentioned here https://github.com/opnsense/plugins/pull/5074

PR #9504, opened 2025-12-08
nat/npt: Refactor network prefix translation view

Monviech:
Uses two methods (toggle log and move rule) that I generalized into base helper methods in this PR: https://github.com/opnsense/core/pull/9492

image


src/opnsense/mvc/app/controllers/OPNsense/Firewall/Api/NptController.php
```diff
@@ -25,24 +25,37 @@
...
 
     public function searchRuleAction()
     {
-        $category = $this->request->get('category');
+        $category = (array)$this->request->get('category');
+
         $filter_funct = function ($record) use ($category) {
-            return empty($category) || array_intersect(explode(',', $record->categories), $category);
+            /* categories are indexed by name in the record, but offered as uuid in the selector */
+            $catids = !empty((string)$record->categories) ? explode(',', (string)$record->categories) : [];
```
fichtner: string? :D explode? :D

getValues()

src/opnsense/mvc/app/controllers/OPNsense/Firewall/Api/NptController.php
```diff
@@ -25,24 +25,37 @@
...
+            return empty($category) || array_intersect($catids, $category);
```
fichtner, line 45: if you cast to array please use count()

Monviech: Can I fix this in a separate PR, because the same pattern is used in 3 or 4 other controllers.

src/opnsense/mvc/app/controllers/OPNsense/Firewall/Api/NptController.php
```diff
@@ -25,24 +25,37 @@
...
-            $filter_funct
-        );
+
+        $results = $this->searchBase("npt.rule", null, "sequence", $filter_funct);
+
+        /* carry results */
+        foreach ($results['rows'] as &$record) {
+            /* offer list of colors to be used by the frontend */
+            $record['category_colors'] = $this->getCategoryColors(
+                !empty($record['categories']) ? explode(',', $record['categories']) : []
```
fichtner, line 54: why are we getting array keys here instead of objects? these constructs handling multi value conversions are not good as they bloat the controller code

src/opnsense/mvc/app/controllers/OPNsense/Firewall/Api/NptController.php
fichtner: okay

src/opnsense/mvc/app/controllers/OPNsense/Firewall/Api/NptController.php
AdSchellevis: `$results` serialized via `searchBase()` and decorated with `category_colors` here, so that's ok

AdSchellevis (changes requested):
let's only fix the `getValues()` as proposed by @fichtner, eventually I think we should cleanup the volt template a bit as well, but that's not a priority (we can do that later, also keeping in mind the different templates have quite some shared functionality now).

PR #9507, opened 2025-12-09
nat/snat: Refactor source nat view

Monviech:
Uses two methods (toggle log and move rule) that I generalized into base helper methods in this PR: https://github.com/opnsense/core/pull/9492

Things that are left (maybe for later)
- An endpoint to control the "Manual, hybrid etc..." behavior in that page as well.
- A second grid that shows the automatic Snat rules? I kinda feel like I want to avoid the trickery of the filter page that funnels it all in the same grid.


fichtner:
(discussion contines in #9515)

Monviech:
I won't merge this before the other PRs. I'll wait until I can merge all of them (its incremental in some way). There are some overlaps (like controller methods still missing in master) and the filter.volt file.

Merge order:
- 1. One to One NAT
- 2. nptv6
- 3. Source NAT

PR #9510, opened 2025-12-10
bootgrid: extend commands structure for generalized usage

swhite2:
Continuation of:
https://github.com/opnsense/core/commit/3fe0cc4a28c37b874a53807abdb3a605db41d46e
https://github.com/opnsense/core/commit/3736489db27e0f5a1406ba5d0bad0c4bf71153d7
https://github.com/opnsense/core/commit/1356068da4043874475f6ed0a20b611d05ca6a70

- primary change: allow commands to be defined in javascript to be rendered in the footer of the table, reducing dependency on the templates.
- provide a PoC of the above in the dhcpv4.volt page. To achieve the same functionality, some command properties have been added as documented in the code: title, footer, filter, onRendered
- The linking of footer actions to rendered footer buttons has been moved to `_renderFooter()` as it was previously tied to data load.
- Tooltip handling was unnecessarily aggressive and scanned the whole grid for each rendered element, this has now been moved to the onRendered handler of each cell, which saves a significant amount of execution time. Some cleanups here as well.
- The `_renderFooterCommands()` function deals with the special cases of 'add' and 'deleteSelected' to maintain the same DOM layout as before.
- We may consider dropping the `this.options.addButton` and `this.options.deleteSelectedButton` logic which ties the `<tfoot>` elements to the add/deleteSelected actions. We can instead render these if all `requires` have been met. Should allows us to clean up quite some code in the templates.

### DIFF
--- a/src/opnsense/www/js/opnsense_bootgrid.js
+++ b/src/opnsense/www/js/opnsense_bootgrid.js
@@ -827,19 +827,20 @@ class UIBootgrid {
         // since we know the footer is rendered at this point,
         // but we don't want to wait on the data before swapping
         // the UI elements around.
+        if (this.options.navigation) {
+            let a = $(`#${this.id} .tabulator-page-counter`)[0];
+            let b = $(`#${this.id} .tabulator-paginator`)[0];
+            var aparent = a.parentNode;
+            var asibling = a.nextSibling === b ? a : a.nextSibling;
+            b.parentNode.insertBefore(a, b);
+            aparent.insertBefore(b, asibling);
 
-        let a = $(`#${this.id} .tabulator-page-counter`)[0];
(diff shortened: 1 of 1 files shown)


src/opnsense/www/js/opnsense_bootgrid.js
```diff
@@ -812,6 +812,7 @@ class UIBootgrid {
 
     _renderFooter() {
         this._renderFooterCommands();
+        this._wireFooterCommands();
```
fichtner: granted this looks a bit strange after refactor side by side, but I get where this is coming from ;)

swhite2: Agreed, moved it :)

https://github.com/opnsense/core/pull/9510/commits/e257b63f07a90cee2218c85f44d7f4bcd3194e3c

PR #9512, opened 2025-12-11
dns/unbound: Fix condition in safesearch template

Monviech:
PR: https://www.reddit.com/r/opnsense/comments/1pjjs7e/force_safesearch_is_broken/

### DIFF
--- a/src/opnsense/service/templates/OPNsense/Unbound/core/safesearch.conf
+++ b/src/opnsense/service/templates/OPNsense/Unbound/core/safesearch.conf
@@ -50,7 +50,7 @@
         "google.cat"
     ]
 %}
-{% if not helpers.empty('OPNsense.unboundplus.dnsbl.safesearch') %}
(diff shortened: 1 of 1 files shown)


src/opnsense/service/templates/OPNsense/Unbound/core/safesearch.conf
```diff
@@ -50,7 +50,7 @@
         "google.cat"
     ]
 %}
-{% if not helpers.empty('OPNsense.unboundplus.dnsbl.safesearch') %}
+{% if OPNsense.unboundplus.general.safesearch|default("0") == "1" %}
```
AdSchellevis: I would prefer the not empty as it's a safer operation.

Monviech: okay changed it again and tested

fichtner:
removed bug tag, it's for tickets, not PRs (since they do not track as issues)

PR #9547, opened 2025-12-19
Firewall: Rules [new]: Add multiselect icmp6type options

Monviech:


src/opnsense/mvc/app/models/OPNsense/Firewall/Filter.xml
```diff
@@ -106,6 +106,37 @@
                         <maskrep>Address Mask Reply (Deprecated)</maskrep>
                     </OptionValues>
                 </icmptype>
+                <icmp6type type="OptionField">
+                    <Multiple>Y</Multiple>
+                    <OptionValues>
+                        <unreach>Destination unreachable</unreach>
```
fichtner: I wonder if it wouldn't be better to use the numbers as option values. The short forms are useful, but only if you don't have a GUI with (translatable) labels.

Issue #9141, opened 2025-08-26
interfaces: convert settings page to MVC/API


fichtner:
**Important notices**

Our forum is located at https://forum.opnsense.org , please consider joining discussions there in stead of using GitHub for these matters.

Before you ask a new question, we ask you kindly to acknowledge the following:

- [x] I have read the contributing guide lines at https://github.com/opnsense/core/blob/master/CONTRIBUTING.md
- [x] I am convinced that my issue is new after having checked both open and closed issues at https://github.com/opnsense/core/issues?q=is%3Aissue

The page is relatively easy and contained, but settings need to be migrated. Test out this page before going to more complex settings pages.

We need our batch of MVC/API conversions for this iteration because even though it's over 80% done that's still a bit left. :)

fichtner:
49bb68282f9fbfd5e8 should be taken into account when adding the model

PR #9569, opened 2025-12-30
interfaces: settings: convert to MVC

swhite2:
- Includes a migration.
- The LLT and LL DUID generation used to calculate based on the MAC of the remote client currently connected to the webgui. I believe this is wrong (and in a lot of cases it won't be in the ARP/NDP table), as the DUID is supposed to be the system-wide unique identifier, so it now picks the system primary interface:

```
$primary_if = get_primary_interface_from_list();
$system_mac = get_interface_mac(get_real_interface($primary_if));
```

- I haven't touched the `config.xml.sample`, as it doesn't have the `OPNsense` namespace yet, is it enough to just add it there?
- Includes the `dhcp6_ratimeout` as @fichtner mentioned in https://github.com/opnsense/core/issues/9141#issuecomment-3693958314

Closes https://github.com/opnsense/core/issues/9141

### DIFF
--- a/src/etc/inc/interfaces.inc
+++ b/src/etc/inc/interfaces.inc
@@ -2772,8 +2772,8 @@ function interface_dhcpv6_configure($interface, $wancfg)
     $settings = new OPNsense\Interfaces\Settings();
 
     /* write DUID if override was set */
-    if (!$settings->ipv6duid->isEmpty()) {
-        dhcp6c_duid_write($settings->ipv6duid->getValue());
+    if (!$settings->dhcp6_duid->isEmpty()) {
+        dhcp6c_duid_write($settings->dhcp6_duid->getValue());
     /* clear DUID if it is faulty */
     } elseif (empty(dhcp6c_duid_read())) {
         dhcp6c_duid_clear();
--- a/src/opnsense/mvc/app/controllers/OPNsense/Interfaces/Api/SettingsController.php
+++ b/src/opnsense/mvc/app/controllers/OPNsense/Interfaces/Api/SettingsController.php
@@ -39,7 +39,7 @@ class SettingsController extends ApiMutableModelControllerBase
     public function getAction()
     {
         $duids = json_decode((new Backend())->configdRun('interface settings gen_duid'), true);
-        if (!$this->getModel()->ipv6duid->isValidDuid($duids['current'])) {
+        if (!$this->getModel()->dhcp6_duid->isValidDuid($duids['current'])) {
             $duids['current'] = $duids['default'];
         }
         $result = parent::getAction();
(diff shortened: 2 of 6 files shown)


src/etc/inc/interfaces.inc
```diff
@@ -2770,10 +2770,11 @@ function interface_6to4_configure($interface, $wancfg, $update = false)
 function interface_dhcpv6_configure($interface, $wancfg)
 {
     $syscfg = config_read_array('system');
```
AdSchellevis: unused in this function now?

src/etc/inc/interfaces.inc
```diff
@@ -2842,6 +2844,7 @@ function interface_dhcpv6_prepare($interface, $wancfg, $cleanup = false)
 
     $id = interface_dhcpv6_id($interface);
     $syscfg = config_read_array('system');
```
AdSchellevis: unused in this function now?

src/etc/inc/interfaces.lib.inc
```diff
@@ -584,7 +584,11 @@ function configure_interface_hardware($ifs, $intf_details = null)
     foreach ($config['interfaces'] as $iface => $ifconf) {
         if (!empty($ifconf['if']) && $ifconf['if'] == $ifs) {
             if (!empty($ifconf['hw_settings_overwrite'])) {
-                $hwsettings = $ifconf;
```
AdSchellevis, line 587:
```
$hwsettings = $config['OPNsense']['Interfaces']['settings'];
```

This likely isn't the correct change. I understand the issue (isset/empty), but we should add a small note here and always initialize `$hwsettings` again to make it stick.

AdSchellevis (changes requested):
@swhite2 I made a couple of small remarks, but most of it looks fine. Eventually we might want to replace config access for model usages, but the current changes are likely less risky and better in the short term.

swhite2:
@fichtner Can you merge this next week if you're ok with the PR? (since I'm unavailable)

fichtner:
Can’t answer in depth today, but the spoof MAC idea is from the perspective of putting OPNsense in front another routing device and mock its MAC address, not necessarily the client one. I’ve heard no complaints about this ever. The approach originates from interfaces.php

I can take care of the rest and merge next week.

happy new year BTW 😊


Cheers,
Franco

swhite2:
> I’ve heard no complaints about this ever

Fair enough, I'll leave it up to you, if we do decide to change it back the whole thing should probably move to the controller, which simplifies it.

Happy new year! :)

swhite2:
> bring back MAC read from connected client

Fixed in https://github.com/opnsense/core/commit/b652d69dfa52aa472c40ee58bc41bf6494fd1bcb

PR #9573, opened 2026-01-01
Fix GPS coordinate display on NTP status page

brotherla:
It looks like this controller was refactor and new format returns data in top level object instead of gps property

image

### DIFF
--- a/src/opnsense/mvc/app/views/OPNsense/Ntpd/status.volt
+++ b/src/opnsense/mvc/app/views/OPNsense/Ntpd/status.volt
@@ -62,7 +62,7 @@
         });
 
         ajaxGet('/api/ntpd/service/gps', {}, function (data, status) {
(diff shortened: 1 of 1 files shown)


AdSchellevis:
@brotherla  I think it's better to drop the `gps` in the controller:

https://github.com/opnsense/core/blob/8e044aca0de05b841ebb1c4b0b4ad1462d13f54d/src/opnsense/mvc/app/controllers/OPNsense/Ntpd/Api/ServiceController.php#L114


Can you try that on your end? I don't have a gps module available at the moment.

brotherla:
> @brotherla I think it's better to drop the `gps` in the controller:
> 

@AdSchellevis yes, it works as well, updated PR

AdSchellevis:
@brotherla thanks!

Issue #9374, opened 2025-11-09
GeoIP download doesn't work in 25.7.7


SeryaOne:
**Important notices**

Before you add a new report, we ask you kindly to acknowledge the following:

- [ x] I have read the contributing guide lines at https://github.com/opnsense/core/blob/master/CONTRIBUTING.md
- [ x] I am convinced that my issue is new after having checked both open and closed issues at https://github.com/opnsense/core/issues?q=is%3Aissue

**Describe the bug**

The GeoIP settings don't work, and downloaded nothing until I put the MaxMind or IPinfo URL in it.

**To Reproduce**

1. Go to 'Firewall'
2. Click on 'Aliases'
3. Click on 'GeoIP settings'
4.1 Put url in 'https://download.maxmind.com/app/geoip_download?edition_id=GeoLite2-Country-CSV&license_key=MY_TOKEN&suffix=zip'
4.2 Put url in 'https://ipinfo.io/data/ipinfo_lite.csv.gz?token=MY_TOKEN'
5. Click on 'Apply'
6. See nothing happens, and the total number of ranges remains 0

I expected that the total number of ranges > 0

**Screenshots**

<img width="2257" height="743" alt="Image" src="https://github.com/user-attachments/assets/58fcf5c2-dbc9-47d0-97dd-ed15600943b8" />

**Environment**

Software version used and hardware type if relevant, e.g.:

OPNsense 25.7.7_4 (amd64).
Intel(R) Core(TM) i7-4510U CPU

Works on OPNsense 25.7.6 and 25.7.4 fine with the same URLs.

SeryaOne:
Addendum: I can't see any requests at MaxMind either. Furthermore, I have this phenomenon on two different devices, one newly installed today, the other somewhat older and frequently updated.

Monviech:
Did you create any geo ip alias yet? I think it downloads when an alias has been created.

SeryaOne:
Yes, I have, and OPNsense always displays a message that I configure the source first.

<img width="1871" height="397" alt="Image" src="https://github.com/user-attachments/assets/b35a36bb-b2cb-4c7d-aa56-3f8dd8a2ec31" />

AdSchellevis:
most likely an invalid license key, but it looks like the code doesn't send that to the log yet., https://github.com/opnsense/core/commit/0304ab66efc7e5f7081b8fe1676db0812ab75dd8 should fix the log part

SeryaOne:
[...]

PR #9581, opened 2026-01-05
Implement alias.stats file validation function

jrichman-r7:
Added a function to validate the alias.stats file for existence and required fields.

### DIFF
--- a/src/opnsense/scripts/filter/lib/alias/geoip.py
+++ b/src/opnsense/scripts/filter/lib/alias/geoip.py
@@ -224,6 +224,7 @@ class GEOIP(BaseContentParser):
                         except (ipaddress.AddressValueError, ValueError):
                             pass
 
+    @staticmethod
     def is_valid_alias_stats(path):
     """
     Validates the alias.stats file.


AdSchellevis:
@jrichman-r7 it might be better to start with a ticket explaining the issue and the proposed fix.

jrichmancodes:
Hey @AdSchellevis , got a little ahead of myself. I was going to submit an issue caused by outdated documentation (https://docs.opnsense.org/manual/how-tos/maxmind_geo_ip.html) that caused a malformed alias.stats file full of null values due to an incorrect or blank URL, and the file validator ```if not os.path.isfile('/usr/local/share/GeoIP/alias.stats')``` would prevent further download attempts even after updating the URL.

Even when I got the correct URL, I had to force the download (python3 /usr/local/opnsense/scripts/filter/download_geoip.py) to bypass the logic preventing it.

jrichmancodes:
https://github.com/opnsense/core/pull/9583 and https://github.com/opnsense/docs/pull/828 are related. I apologize for the additional PR noise; I haven't worked with forks before and am doing this all through the GitHub UI at the moment.

AdSchellevis:
@jrichman-r7 no problem, the documentation PR is much appreciated, the other one might be caused earlier in the sequence, in which case we might need a different fix (so the faulty content (all nulls) helps in that case)

jrichmancodes:
@AdSchellevis Let me know if there's anything else I can do for you on this!

AdSchellevis:
@jrichman-r7 I think the doc's one is finished, right? I that case you can flip the draft to ready for review :)

AdSchellevis:
> Hey @AdSchellevis , got a little ahead of myself. I was going to submit an issue caused by outdated documentation (https://docs.opnsense.org/manual/how-tos/maxmind_geo_ip.html) that caused a malformed alias.stats file full of null values due to an incorrect or blank URL, and the file validator if not os.path.isfile('/usr/local/share/GeoIP/alias.stats') would prevent further download attempts even after updating the URL.

> Even when I got the correct URL, I had to force the download (python3 /usr/local/opnsense/scripts/filter/download_geoip.py) to bypass the logic preventing it.

hmm, this is a bit weird, as it only flushes `_stats_output`, but never uses it to determine age, which is handled by the following block:

https://github.com/opnsense/core/blob/9092d60b38bedb3a8a157dbdde2f47717b77481d/src/opnsense/scripts/filter/lib/alias/geoip.py#L205-L208

It always updates unless the geoip target file is found and valid, changing the source is part of the equation. 

Any idea how we can reproduce your sequence of events?

jrichmancodes:
Absolutely!

First via SSH, clear out your `/usr/local/share/GeoIP/` directory of the alias dir, md5 file, and alias.stats file. I did this with `sudo -i` permissions.

In the Web UI, navigate to your geoip settings under Firewall -> Aliases, clear out your Url, and click 'Apply'.

My understanding is that this saves a configuration file and triggers update_tables.py to download the geoip data, and geoip.py writes the default 'result' variable values to alias.stats in the _update function.

On subsequent Url updates via web UI and clicking 'Apply', the download is not triggered in update_tables.py as alias.stats exists.

jrichmancodes:
related thread :) 

https://www.reddit.com/r/opnsense/comments/1pm1ckb/cant_configure_maxmind_geoip_tables/

AdSchellevis:
> On subsequent Url updates via web UI and clicking 'Apply', the download is not triggered in update_tables.py as alias.stats exists.

Well, that part is weird, as the update doesn't depend on that file. Let me try to explain the flow and see if we can reconstruct the issue in some way:

`update_tables.py` is indeed the starting point, after which each alias detects if it has changed or expired:

https://github.com/opnsense/core/blob/9092d60b38bedb3a8a157dbdde2f47717b77481d/src/opnsense/scripts/filter/update_tables.py#L85

when it has changed, we will try to resolve its contents here:

https://github.com/opnsense/core/blob/9092d60b38bedb3a8a157dbdde2f47717b77481d/src/opnsense/scripts/filter/update_tables.py#L105

which dives into:

https://github.com/opnsense/core/blob/9092d60b38bedb3a8a157dbdde2f47717b77481d/src/opnsense/scripts/filter/lib/alias/__init__.py#L199

and eventually ends up in:

https://github.com/opnsense/core/blob/9092d60b38bedb3a8a157dbdde2f47717b77481d/src/opnsense/scripts/filter/lib/alias/geoip.py#L207


My best guess is the `changed()` doesn't report properly now as it assumes empty is the proper content, if that's the case, you could retry your test and only remove the contents of the alias cache file in `/var/db/aliastables/<alias_name>.*`

If that fixes the issue on your end, I think I know what we should modify.

jrichmancodes:
I believe my issue is specific to the initial file download on clicking 'Apply', not a later update. If not, I'll have to take another look at your suggestion.

jrichmancodes:
A screenshot of my directory, for reference: 
image


I haven't configured aliases to utilize the geoip db yet.

And contents of the malformed alias.stats
```{"address_count":0,"file_count":0,"timestamp":null,"locations_filename":null,"address_sources":{"IPv4":null,"IPv6":null}}```

**Comment by AdSchellevis** [comment]
> I believe my issue is specific to the initial file download on clicking 'Apply', not a later update. If not, I'll have to take another look at your suggestion.

well, the regular update (apply button) always uses the same sequence, it determines if it needs to fetch files itself based on its type.  so, if there's at least one alias requiring geoip, it will try to download it. (when no aliases are configured, there's also no need to download)

**Comment by jrichmancodes** [comment]
Ah, and here I was just trying to make sure the link worked :)

**Comment by jrichmancodes** [comment]
From a user standpoint, I think it's important to confirm that someone has a valid link regardless of it being used as an alias, esp. with the documentation that was outdated (and the UI that doesn't confirm anything in the current state).

**Comment by AdSchellevis** [comment]
The documentation is one part indeed,  but if there is still an issue when the proper url is offered with the test I described earlier, just let me know.

It's not really possible to validate the url other than during processing as this operation will take too long (you need to download and unpack in order to know if its valid). The log reports issues when the url can't be used.

**Comment by jrichmancodes** [comment]
I tried finding relevant logs, but couldn't. Can you tell me which logs/where to find them?

Also, I'm not sure about other datasets, but the download from maxmind was only 5.2MB and took seconds to download/process on my router, so I'm wondering if it's not feasible as other datasets are much larger?

I'll get back to you later about the other test.

**Comment by AdSchellevis** [comment]
> I tried finding relevant logs, but couldn't. Can you tell me which logs/where to find them?

The process should log in "Firewall: Log Files: General" 

> Also, I'm not sure about other datasets, but the download from maxmind was only 5.2MB and took seconds to download/process on my router, so I'm wondering if it's not feasible as other datasets are much larger?

For small machines both download and unpack can take much longer, it's common practice to validate the input (uri in this case) but not the result (file download).

**Comment by jasonpcrowley** [comment]
> My best guess is the `changed()` doesn't report properly now as it assumes empty is the proper content, if that's the case, you could retry your test and only remove the contents of the alias cache file in `/var/db/aliastables/<alias_name>.*`

I tried removing `/var/db/aliastables/<alias_name>.*`, and it fixed the problem. In less than a minute, these lines showed up in `/var/log/firewall/latest.log`. The GeoIP aliases were populated with IP ranges, and the filter updated to use the aliases correctly. 
```
<165>1 2026-01-24T06:41:02-06:00 firewall.mydomain.com firewall 23282 - [meta sequenceId="1"] found .gz format, process
<165>1 2026-01-24T06:41:14-06:00 firewall.mydomain.com firewall 23282 - [meta sequenceId="2"] geoip updated (files: 496 lines: 4031215)
```

I have some thoughts:
1. There might be a logic problem here. You're checking for the existence of `/usr/local/share/GeoIP/alias/NL-IPv4` then trying to get stats from it, but it doesn't exist yet. It seems to me like `do_update` should stay true, but I don't see the associated log entries, at least not for 24 hours. I'm not sure what kicks off this function, so I'm not sure how often it should get executed. 
   https://github.com/opnsense/core/blob/9092d60b38bedb3a8a157dbdde2f47717b77481d/src/opnsense/scripts/filter/lib/alias/geoip.py?plain=1#L204-L213
2. If the above logic is not the problem, maybe forcing a retry when the GeoIP database has 0 files or 0 lines would help the aliases get populated faster. As it stands now, it takes 24 hours. 
3. Adding some log stating that we're downloading GeoIP aliases prior to or on the same line with `found .gz format, process` would make that more intuitive to the reader.

AdSchellevis:
@jasonpcrowley thanks for confirming my suspicion, the issue should be in the `changed()` logic as it doesn't account for a missing database. I'll reopen the other issue.

AdSchellevis:
should be fixed with https://github.com/opnsense/core/commit/7bb40089867079e75857823d26ecdee5b12075d8 (https://github.com/opnsense/core/issues/9374#issuecomment-3795378720)

PR #9593, opened 2026-01-06
Services: Dnsmasq DNS & DHCP: Add automatic RDNSS option when none is configured

Monviech:
For: https://github.com/opnsense/core/issues/9155

Dnsmasq uses option6 23 for both Router Advertisements via RDNSS option and in DHCPv6 as dns-server option. For the default configuration, it is beneficial to handle these the same as the DHCPv4 default to align with user expectations and ease configuring the defaults for Router Advertisements.

### DIFF
--- a/src/opnsense/service/templates/OPNsense/Dnsmasq/dnsmasq.conf
+++ b/src/opnsense/service/templates/OPNsense/Dnsmasq/dnsmasq.conf
@@ -289,7 +289,8 @@ set:{{ host.set_tag|replace('-', '') }},
 {%-     endif +%}
 {% endfor %}
 
-{% set has_default=[] %}
+{% set has_default_v4=[] %}
+{% set has_default_v6=[] %}
 {% for option in helpers.toList('dnsmasq.dhcp_options') %}
 {%     if option.type == "set" %}
 {%     set all_tags = [] %}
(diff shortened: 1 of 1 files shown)


src/opnsense/service/templates/OPNsense/Dnsmasq/dnsmasq.conf
```diff
@@ -321,10 +325,14 @@ option6:{{ option.option6 }}{% if option.value %},{{ option.value }}{% endif %}
...
 
-{% if not has_default %}
-# default dns mapped to this server (0.0.0.0)
+{% if not has_default_v4 %}
+# default IPv4 DNS mapped to this server (0.0.0.0)
 dhcp-option=6,0.0.0.0
 {% endif %}
+{% if not has_default_v6 %}
+# default IPv6 DNS mapped to this server (::)
+dhcp-option=option6:23,[::]
```
fichtner, line 334: is `[::]` dnsmasq magic?

Monviech: Yes, it will automatically track the IPv6 address(es) of the interface it sends the option from. GUA and ULA, it skips LLA.

fichtner: 👍

Issue #9334, opened 2025-10-31
radvd: (Re)add Base6Interface option


Monviech:
**Important notices**

Before you add a new report, we ask you kindly to acknowledge the following:

- [x] I have read the contributing guide lines at https://github.com/opnsense/core/blob/master/CONTRIBUTING.md
- [x] I am convinced that my issue is new after having checked both open and closed issues at https://github.com/opnsense/core/issues?q=is%3Aissue

**Is your feature request related to a problem? Please describe.**

```
       Base6Interface name

	      If  this options is specified, this prefix will be combined with
	      the IPv6 address of the interface	specified by  name.   The  re-
	      sulting prefix length will be 64.

```

Example configuration:

eth0 = LAN
eth1 = WAN

```
# LAN
interface eth1 {
    prefix ::/64 {
        Base6Interface eth0;
    };
};
```
or

```
# LAN
interface eth1 {
    prefix ::/64 {
        Base6Interface eth1;
    };
};
```

**Describe the solution you like**

Adding this again at some point.

**Describe alternatives you considered**

It's mostly needed when an NDP proxy is used and you want to send your own RAs with radvd. Dnsmasq is not an alternative, its constructor cannot use an interface with a dynamic (SLAAC) address as base.

**Additional context**

https://man.freebsd.org/cgi/man.cgi?query=radvd.conf&apropos=0&sektion=5&manpath=FreeBSD+14.3-RELEASE+and+Ports&arch=default&format=html

Monviech:
```
root@opn-dev-02:/src/git/radvd # cat /var/etc/radvd.conf
# Automatically generated, do not edit
# Generated RADVD config for manual assignment on lan
interface hn0 {
	AdvSendAdvert on;
	MinRtrAdvInterval 10;
	MaxRtrAdvInterval 30;
	AdvDefaultLifetime 9000;
	AdvLinkMTU 1500;
	AdvDefaultPreference medium;
	AdvManagedFlag off;
	AdvOtherConfigFlag on;
	prefix ::/64 {
                Base6Interface hn0;
		DeprecatePrefix on;
		AdvOnLink on;
		AdvAutonomous on;
	};
	RDNSS fd80::1:1 {
	};
	DNSSL ad.pischem.com {
	};
};
```

It looks like when the IP changes that radvd does /not/ need a restart or SIGHUP:

```
root@opn-dev-02:/src/git/radvd # ifconfig hn0
[...]

PR #9595, opened 2026-01-06
interfaces: make Dnsmasq the default IPv6 provider

Monviech:
For: https://github.com/opnsense/core/issues/9155

An additional DHCP range constructs RA from the LAN prefix. LAN requires track interface for this to work. ra-stateless will set the A bit, allowing clients to use SLAAC and optionally receive another address other options via DHCPv6. The RDNSS option will be offered via eaf7630.

This is just a starting point so we can see whats needed.

### DIFF
--- a/src/etc/config.xml.sample
+++ b/src/etc/config.xml.sample
@@ -95,8 +95,8 @@
     </dhcp_ranges>
     <dhcp_ranges>
       <interface>lan</interface>
-      <start_addr>::100</start_addr>
-      <end_addr>::199</end_addr>
+      <start_addr>::1000</start_addr>
+      <end_addr>::2000</end_addr>
       <constructor>lan</constructor>
       <ra_mode>slaac</ra_mode>
     </dhcp_ranges>


fichtner:
I think this requires to remove track6 from the default lan configuration. But which mode should be used? The new link-local one?

src/etc/config.xml.sample
```diff
@@ -85,11 +85,20 @@
...
+    </dhcp>
     <dhcp_ranges>
       <interface>lan</interface>
       <start_addr>192.168.1.100</start_addr>
       <end_addr>192.168.1.199</end_addr>
     </dhcp_ranges>
+    <dhcp_ranges>
+      <interface>lan</interface>
+      <start_addr>::</start_addr>
+      <constructor>wan</constructor>
```
fichtner: is this even correct? shouldn't we use "lan" for the right /64 subnet? but then this requires track6 again

Monviech: You are right, it's not correct, I made a typo there. It should be ``lan`` as well.

And yes currently this requires track interface, because otherwise dnsmasq has no concept of different prefixes, that could be set by DHCPv6-PD.

fichtner: @Monviech @AdSchellevis then maybe the best course of action is to add a new "track6" mode that doesn't have the legacy attachment to isc-dhcp. if you think that's good idea I can try to come up with a proposal

Monviech: https://docs.opnsense.org/manual/dnsmasq.html#dhcpv6-and-router-advertisements

Monviech: Judging from the man page, dnsmasq does not need any restarts/reloads when the address changes:

--dhcp-range=::1,::400,constructor:eth0

will look for addresses on eth0 and then create a range from <network>::1 to <network>::400. If the interface is assigned more than one network, then the corresponding ranges will be automatically created, and then deprecated and finally removed again as the address is deprecated and then deleted. The interface name may have a final "*" wildcard. **Note that just any address on eth0 will not do: it must not be an autoconfigured or privacy address, or be deprecated.**

If a --dhcp-range is only being used for stateless DHCP and/or SLAAC, then the address can be simply ::

--dhcp-range=::,constructor:eth0

Monviech: I think a track interface mode that does not hook into any RA daemon and dhcpv6 server service sounds like a good idea.

Monviech: A tangent

I think essentially radvd has the same option like dnsmasqs constructor, Base6Interface: https://github.com/opnsense/core/issues/9334

So I dont think it needs anything special either?

fichtner: These constructors are useful especially for shifting prefixes but they may have edge cases too and differ in implementation details. Since we can do divide-and-conquer here let's focus on Dnsmasq-only mode for default configurations.

Looking at the problem scope I want to call the new mode `ia6` or `idassoc6` and offer the same implementation that `track6` has at the moment and all of the dhcpd.inc stuff is then automatically off when the plugin is not installed.

src/etc/config.xml.sample
```diff
@@ -85,11 +85,20 @@
...
     <dhcp_ranges>
       <interface>lan</interface>
       <start_addr>192.168.1.100</start_addr>
       <end_addr>192.168.1.199</end_addr>
     </dhcp_ranges>
+    <dhcp_ranges>
+      <interface>lan</interface>
+      <start_addr>::</start_addr>
+      <constructor>lan</constructor>
+      <ra_mode>ra-stateless</ra_mode>
```
Monviech: Just putting a comment here that this needs discussion.

It currently sets A and O bit (Slaac address only, stateless DHCPv6) which means there will not be 2 addresses (Slaac + DHCPv6) but only 1 address (Slaac only).

That differs from current default behavior it seems? But its more correct since Slaac is the universal truth of all implementations, and DHCPv6 addresses are best effort only.

Monviech: I wonder in practice if the DHCPv6 address (IA_NA) would even be used by client implementations for outbound (source address selection), if there is also a SLAAC address available:

https://datatracker.ietf.org/doc/html/rfc6724#section-5

An interpretation of rule 5.5 could be:
SLAAC addresses are preferred in practice because they are associated with the advertising next-hop (Rule 5.5), while DHCPv6 IA_NA addresses are not.

fichtner: We can try it, yet in the real world who knows what somebody did and keeping closer to what we had before wouldn't be the worst thing (especially since Dnsmasq can do it). I'm not advocating for any direction, but I have to comment on guessing what a best practice might be.

In my install for example IA-NA is the same as SLAAC -- I'm not sure if this is really what the RFC wants but it is what the Fritzbox does.

Monviech: Okay I changed it to a range, so the mode now works like this:

`"slaac" tells dnsmasq to offer Router Advertisement on this subnet and to set the A bit in the router advertisement, so that the client will use SLAAC addresses. When used with a DHCP range or static DHCP address this results in the client having both a DHCP-assigned and a SLAAC address.`

src/etc/config.xml.sample
```diff
@@ -85,11 +85,21 @@
...
+    </dhcp>
     <dhcp_ranges>
       <interface>lan</interface>
       <start_addr>192.168.1.100</start_addr>
       <end_addr>192.168.1.199</end_addr>
     </dhcp_ranges>
+    <dhcp_ranges>
+      <interface>lan</interface>
+      <start_addr>::100</start_addr>
+      <end_addr>::199</end_addr>
```
fichtner: for reference ISC did ::1000 - ::2000

Monviech:
That's all that's needed for the default as far as I know, let's merge it.

Self-Hosting-Group:
When 26.1 is newly installed and the setup wizard is not launched, IPv6 works on the LAN as in previous versions, but when I complete the wizard, the Dnsmasq IPv6 range is deleted. Is this intentional? How about if only the IPv4 range is deleted by the wizard? PS: I have seen that there are plans to extend the wizard for IPv6 in 26.7, which is of course the optimal solution. See: https://github.com/opnsense/core/blob/master/src/opnsense/mvc/app/models/OPNsense/Core/InitialSetup.php#L383-L385

fichtner:
At the moment this is intended historic behaviour since the wizard does not (and never did) cover IPv6: https://github.com/opnsense/core/issues/9653

Ran into the same issue but wedging it in felt wrong here so this will be a larger piece for 26.7.


Cheers,
Franco

Self-Hosting-Group:
IMHO, this is a regression as IPv6 does not work out-of-the-box with 26.1. Why not add a line and not delete the IPv6 ranges that are part of config.xml.sample?

fichtner:
Because it would have shifted the release from the targeted day to a later day which wasn't possible for other reasons.

fichtner:
FWIW, fixed in 80d4a70

Self-Hosting-Group:
> FWIW, fixed in [80d4a70](https://github.com/opnsense/core/commit/80d4a70f2ac5c916f62c434bcb5c9113cd75d6a8)

Great. Thank you.

PR #9596, opened 2026-01-06
System: rename sudoers file to make it more sortable

### LINKED ISSUE #8930: /usr/local/etc/sudoers.d/opnsense should be 50-opnsense
The sudoers file `/usr/local/etc/sudoers.d/opnsense` should be renamed to `/usr/local/etc/sudoers.d/50-opnsense` to make it easy to put rules, both before and after using number prefixes like `00-` and `99-`.

Alternatively one could use letter prefixes like `aa-` and `xx-`, but the general convention is to use numbers.

davidolrik:
Fixes #8930

Just a simple rename, to make it easier to add more "sudoers" files before and after.

### DIFF
--- a/src/opnsense/service/templates/OPNsense/Auth/+TARGETS
+++ b/src/opnsense/service/templates/OPNsense/Auth/+TARGETS
@@ -1,6 +1,7 @@
 ipsec.pam:/etc/pam.d/ipsec
 motd:/var/run/motd
 sshd.pam:/etc/pam.d/sshd
-sudoers:/usr/local/etc/sudoers.d/opnsense
+sudoers:/usr/local/etc/sudoers.d/20-opnsense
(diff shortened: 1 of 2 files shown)


fichtner:
Now you have two problems instead of one: the old file is permanently left with the last known content.

davidolrik:
Is there a way to migrate the old?

fichtner:
You need to keep writing a file with a single comment for documentation purposes.

BTW, we use “20” for such things from core perspective and “50” for plugins.

davidolrik:
And it would be ok to just overwrite the old non numbered file?

davidolrik:
Changed numbering to 20, and added deprecation notice.

fichtner:
Merged, thanks!

PR #9603, opened 2026-01-08
router advertisements: migrate to MVC/API

fichtner:
TODO:

- [x] Additional validations that were in place in the old www file
- [x] Settle on a debug level (currently 2 from 1-5) to gain a few syslog messages that could help with debugging: -d1
- [x] Test migration scenarios
- [x] ~~What to do with "tracking" configurations now that the old page is gone? Mock in model?~~ We can migrate the disabled entries and can add them in the MVC code manually. Only visibility is missing compared to the old page but that's not a PR merge blocker

### DIFF
--- a/src/www/services_dhcpv6.php
+++ b/src/www/services_dhcpv6.php
@@ -116,10 +116,6 @@ function process_track6_form($if)
 {
     $dhcpdv6cfg = &config_read_array('dhcpdv6');
     $this_server = [];
-    if (isset($dhcpdv6cfg[$if]) && isset($dhcpdv6cfg[$if]['ramode'])) {
-        /* keep ramode for router advertisements so we can use this field to disable the service when in tracking mode */
-        $this_server['ramode'] = $dhcpdv6cfg[$if]['ramode'];
-    }
     if (empty($_POST['enable'])) {
         $this_server['enable'] = '-1';
     }


fichtner:
This is ready for review pending more testing and final polish. Probably ready for merging on Monday.

src/etc/inc/plugins.inc.d/radvd.inc
```diff
@@ -80,6 +85,17 @@ function radvd_services()
     return $services;
 }
 
+function radvd_xmlrpc_sync()
+{
+    return [[
+        'help' => gettext('Synchronize router advertisements.'),
+        'description' => gettext('Router Advertisements'),
+        'section' => 'OPNsense.radvd',
```
Monviech, line 93: Theoretically this needs a migration since users who had dhcp6 section enabled now wont have the new radvd section enabled.

Monviech:
I'm handling the model validations right now.

src/etc/inc/plugins.inc.d/radvd.inc
fichtner: if you want to do it, sure

Monviech:
@fichtner I've added the validations now, please give them a look.

src/etc/inc/plugins.inc.d/radvd.inc
Monviech: I'm not sure, it could be a comment in the patch notes. Won't break any running system if its not auto migrated. Just wanted to leave a pointer here.

fichtner: As discussed internally: no migration for this at the moment because the scope (and fix) is small.

Monviech (approved):
It's good for master.

PR #9606, opened 2026-01-08
Firewall: Rules:  Migration assistant - add export option and guidance for migrations to the new mvc system.

### LINKED ISSUE #9579: Firewall: Rules [new] - assist migrations with import/export functionality
**Important notices**

As migrating rules from the legacy format to the new MVC format has some risks, we should offer some handles to ease the process for our users when they want to migrate to the new user interface in a control manner. Plus side is that these import/exports can also be used to bulk modify data for advanced users.

A simple import/export function to dump all rules in a simple (csv) file and a way to merge these back into the configuration using their unique ids (`uuid`), these should be unique across mvc and legacy due to a code change from a couple of years ago (https://github.com/opnsense/core/commit/cebea865c98692937fd706ae4dfb892b86e2b20d)

Automatic migration, forcing the users into the new at some point, but given such a key element of the system, better offer some control over the process.

**Additional context**

none

AdSchellevis:
Add new "Firewall: Rules: Migration assistant" to help people moving to our new mvc based firewall system by offering a couple of simple steps for exporting the old rules into a csv file and importing them into the new system. When all rules are migrated the user may drop all old ones using the ConfigMaintenance module used in "defaults".

There's one small issue remaining in fbegin.inc as it doesn't render the icon correctly due to VisibleName being escaped currently (which isn't the case in our mvc template).

closes https://github.com/opnsense/core/issues/9579

### DIFF
--- a/src/opnsense/service/conf/actions.d/actions_filter.conf
+++ b/src/opnsense/service/conf/actions.d/actions_filter.conf
@@ -54,6 +54,13 @@ parameters:
 type:script_output
 message:request active rule ids and descriptions
 
+[list.legacy_rules]
+command:/usr/local/opnsense/scripts/filter/list_legacy_rules.php
+parameters:
+type:script_output
+message:dump legacy rules in mvc format
+
+
 [list.pfsync]
 command:/usr/local/opnsense/scripts/filter/list_pfsync.py
 parameters: %s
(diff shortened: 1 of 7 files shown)


src/opnsense/service/conf/actions.d/actions_filter.conf
```diff
@@ -54,6 +54,13 @@ parameters:
 type:script_output
 message:request active rule ids and descriptions
 
+[list.legacy_rules]
+command:/usr/local/opnsense/scripts/filter/list_legacy_rules.php
+parameters:
+type:script_output
+message:dump legacy rules in mvc format
+
```
fichtner, line 62:
```suggestion
```

PR #9608, opened 2026-01-09
dashboard: ui improvements of the Services widget

### LINKED ISSUE #7887: system: compact services widget
**Important notices**

The service widget could avoid wasted empty space. Sometimes the list is too long and you need to scroll down inside the widget.

Make a more compact tile-based service widget.

N/A

**Additional context**

N/A

sopex:
A good start.
Closes: https://github.com/opnsense/core/issues/7887

Screenshot 2026-01-09 182920

### DIFF
--- a/src/opnsense/www/js/widgets/Services.js
+++ b/src/opnsense/www/js/widgets/Services.js
@@ -38,17 +38,16 @@ export default class Services extends BaseTableWidget {
     }
 
     getMarkup() {
-        let $table = this.createTable('services-table', {
-            headerPosition: 'left',
-            headerBreakpoint: 270
-        });
-        return $(`<div id="services-container"></div>`).append($table);
+        // Return a flex container instead of a table for compact tile layout
+        return $(`<div id="services-container" style="display: flex; flex-wrap: wrap; padding: 5px; gap: 5px; justify-content: space-between;"></div>`);
     }
 
     serviceControl(actions) {
         return actions.map(({ action, id, title, icon }) => `
             <button data-service_action="${action}" data-service="${id}"
-                  class="btn btn-xs btn-default srv_status_act2" style="font-size: 10px;" title="${title}" data-toggle="tooltip">
+                  class="btn btn-xs btn-default srv_status_act2"
+                  style="font-size: 10px; padding: 2px 6px;"
+                  title="${title}" data-toggle="tooltip">
                 <i class="fa fa-fw fa-${icon}"></i>
             </button>
         `).join('');
@@ -62,13 +61,16 @@ export default class Services extends BaseTableWidget {
             return;
         }
 
+        const $container = $('#services-container');
+        $container.empty();
+
         $('.service-status').tooltip('hide');
         $('.srv_status_act2').tooltip('hide');
 
-        for (const service of data.rows) {
-            let name = service.name;
-            let $description = $(`<div style="font-size: 12px;">${service.description}</div>`);
+        // Sort by description for better readability
+        data.rows.sort((a, b) => a.description.localeCompare(b.description));
 
+        for (const service of data.rows) {
             let actions = [];
             if (service.locked) {
                 actions.push({ action: 'restart', id: service.id, title: this.translations.restart, icon: 'refresh' });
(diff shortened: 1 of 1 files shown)


fichtner:
@sopex thanks! The schedule for 26.1 is a bit right. I need to wait for merge until stable/26.1 is branched which should happen early next week.


Cheers,
Franco

sopex:
> @sopex thanks! The schedule for 26.1 is a bit right. I need to wait for merge until stable/26.1 is branched which should happen early next week.
> 
> Cheers, Franco

Sure. Let me know if you have any changes in mind. I understand that these days are difficult!

Keep up the awesome work!

swhite2 (commented):
Don't think this needs to be a `BaseTableWidget` anymore?

swhite2:
Franco also mentioned the idea of placing the name on the left side of the tile, the buttons on the right side and letting the tile background color reflect the current status, instead of the "play" button. The tooltip should then be triggered on the entire tile.

sopex:
I personally like them being centered. 

Another idea would be eliminating the buttons. Have the tile be green for On/ Blue for restarting/ red for off. And make the whole tile a button. But you are losing the restart functionality for the ones that support turning off.

sopex:
But honestly, I like the result of this PR. It fits with the whole aesthetic much more, while modernizing the old list and making it look like a control panel.

sopex:
@fichtner @swhite2 Any new thoughts here?

swhite2:
The WebGUI tile seems a bit off here, any chance this can be equally sized?

sopex:
> The WebGUI tile seems a bit off here, any chance this can be equally sized?

I suppose it could. It just resizes because it's the last one, and there is space for it. If you prefer it to leave empty space, I don't believe this would be a problem.

swhite2:
> > The WebGUI tile seems a bit off here, any chance this can be equally sized?
> 
> I suppose it could. It just resizes because it's the last one, and there is space for it. If you prefer it to leave empty space, I don't believe this would be a problem.

Let's go for that then.

sopex:
Option 1:

Screenshot 2026-03-24 174944

Screenshot 2026-03-24 174902

Option 2:

Screenshot 2026-03-24 175001

Screenshot 2026-03-24 174122

Sure about option 1? @swhite2

sopex:
@swhite2 If you have an input here too.

swhite2:
Yes, option 1 looks better.

Testing this it seems the grid of controls isn't scaling the widget. when made small enough, you can only see the first column.

sopex:
> Yes, option 1 looks better.

Ok, I am 50-50


> Testing this it seems the grid of controls isn't scaling the widget. when made small enough, you can only see the first column.

Do you want it to rescale or just disable the resizing? It's not like it was resizing great before, and 3 columns seem a good idea.

sopex:
@swhite2 This way it works much better on mobile devices too.

sopex:
@swhite2 Hi Stephan, kind reminder.

src/opnsense/www/js/widgets/Services.js
```diff
@@ -79,21 +79,46 @@ export default class Services extends BaseTableWidget {
...
+                        font-weight: bold;
+                        font-size: 12px;
+                        margin-bottom: 5px;
+                        white-space: nowrap;
+                        overflow: hidden;
+                        text-overflow: ellipsis;
+                        width: 100%;
+                        text-align: center;
+                        color: #555;
+                    " title="${service.description}">${service.description}</div>
```
swhite2: This seems to be missing a tooltip:

Image

swhite2:
@sopex Thoughts on aligning the buttons consistently?

image

src/opnsense/www/js/widgets/Services.js
sopex: Isn't the tooltip the thing that shows up when you hover above the text?
This is Chrome but tried it with firefox that you seem to use and it works. 

https://github.com/user-attachments/assets/04a2b8d7-119e-4fff-8521-38349e9f2c11

sopex:
@swhite2 If this counts, then yes. You break them to two rows, the left row is aligned right with a 10px padding and the right row is aligned left.

image

src/opnsense/www/js/widgets/Services.js
swhite2: @sopex Yep! however, "our" tooltip is more direct and prominent and triggered through `data-toggle="tooltip"`. Most implementations of tooltip activation (like `$('.service-status').tooltip({container: 'body'});`) also aren't needed anymore due to https://github.com/opnsense/core/blob/master/src/opnsense/mvc/app/views/layouts/default.volt#L182

swhite2:
> @swhite2 If this counts, then yes. You break them to two rows, the left row is aligned right with a 10px padding and the right row is aligned left.

I think this is a bit easier on the eyes, let's go for that.

sopex:
> I think this is a bit easier on the eyes, let's go for that.

Done!

src/opnsense/www/js/widgets/Services.js
```diff
@@ -62,13 +60,12 @@ export default class Services extends BaseTableWidget {
             return;
         }
 
-        $('.service-status').tooltip('hide');
-        $('.srv_status_act2').tooltip('hide');
+        const $container = $('#services-container');
```
swhite2, line 65: One small thing left: Before we do anything, we probably do want to hide existing tooltips before we update the UI each tick. For example: https://github.com/opnsense/core/blob/master/src/opnsense/www/js/widgets/IpsecTunnels.js#L93

This is because the UI renders the tooltip, but we re-render the tiles, decoupling the tooltip from the original tile, locking it in place.

As of yet I haven't found a proper generic solution for this

swhite2 (approved):
Looks good to me apart from the small nitpick above. There's something to be said still about the spacing in the tiles, but I'll discuss that with @fichtner

sopex:
> Looks good to me apart from the small nitpick above. There's something to be said still about the spacing in the tiles, but I'll discuss that with @fichtner

I think with [9c99154](https://github.com/opnsense/core/pull/9608/commits/9c99154529602591aef73bd36e0051c75c04bdf4) this is almost perfect. 1.2 or 1.25 should do the trick.

image

swhite2:
Looks good, thanks @sopex :)

ronin3510:
> @sopex Thoughts on aligning the buttons consistently?
> 
> image

Multiple Gateway Monitors where the actual name describing what the monitor is for is mostly hidden are not particularly useful.

To me this looks like the corner case where using two lines would be justified:

Line 1 - widget function - Gateway Monitor 

Line 2 - widget scope - WAN_Primary-Vodafone

================

And another bit of feedback for everyone to consider: 

The words “service” “daemon” and “server” could simply be removed as there’s no value added by their presence randomly and inconsistently in the widgets. In this particular case less is more.


My €0.02 🙂

fichtner:
Yes, both are basically a problem of overly verbose and directionless service name descriptions. While these could be changed the GUI should be a bit more open to showing them instead of truncating them (as before). There's still #10178 open for further tweaking.

sopex:
The gateway monitor X can probably become X monitor too.

AdSchellevis:
We probably want to anchor the colors to something already in the theme file.... below an example using `opnsense-dark`

image

Something like the following might do the trick here (as long as we can "pick" the color from the current DOM).

```
getComputedStyle(document.querySelector('.navbar-default')).backgroundColor
```

sopex:
> We probably want to anchor the colors to something already in the theme file.... below an example using `opnsense-dark`
> 
> image
> Something like the following might do the trick here (as long as we can "pick" the color from the current DOM).
> 
> ```
> getComputedStyle(document.querySelector('.navbar-default')).backgroundColor
> ```

Yes I caught it a few weeks back, but @fichtner is experimenting with #10178 so I am not sure there is a need yet.

Issue #8912, opened 2025-07-04
Hidden fields in the "add alias" form are being stored (which may lead to exposed passwords from browser autofill)


t3t3:
**Important notices**

Before you add a new report, we ask you kindly to acknowledge the following:

- [X] I have read the contributing guide lines at https://github.com/opnsense/core/blob/master/CONTRIBUTING.md
- [X] I am convinced that my issue is new after having checked both open and closed issues at https://github.com/opnsense/core/issues?q=is%3Aissue

**Describe the bug**

When adding an alias, the popup form contains fields which are just hidden, but not disabled. Circumstances may have the browser filling them in (including passwords!), plus they get stored and will be part of a json export (as clear text).

**To Reproduce**

1. Have OPNsense running on an IP address (or URL) for which i.e. Chrome has stored a password (which in my case happend during setup, when running on a non-production IP)
2. Add a host alias
3. Export the aliases as json
4. See the stored password in the "password" field of that alias

**Expected behavior**

I would expect the GUI to not expose unused input fields, or at least to disable them, and also to not store entries from hidden fields (nowhere actually)

**Describe alternatives you considered**

Can't think of any easy workaround, apart from using a browser/profile with no autofill data when interacting with the OPNsense GUI

fichtner:
Wasn’t it considered impractical for browsers to fill hidden fields quite some time ago because of data scraping issues?

t3t3:
... can't tell. I'm using an up-to-date Chrome version, and it happened. But it would be (IMO) better to prohibit that from the GUI – on a Firewall at least ;)

fichtner:
Did you ever switch to JSON type or was this without interaction? And which password did it auto-fill, the OPNsense firewall password?

t3t3:
Without interaction; just clicked on the '+' button, filled in name & IP address & clicked Save. Exported the json. Repeated – same result.

I'm currently running it on the default IP (192.168.1.1, over https), just for doing preliminary setup, and had a password stored in Chrome for that particular IP. So it wasn't a password for any actual OPNsense user, in my case not even anything "production-grade", but understandably still had me raising an eyebrow, when I found a password (I knew), which shouldn't be there, in that export file...

zerwes:
I can confirm this.
[...]

PR #9610, opened 2026-01-11
disable autocomplete for alias password form field

zerwes:
fix for #8912


AdSchellevis:
@zerwes thanks, used the new-password tag in https://github.com/opnsense/core/commit/1ddc63e402564c2a3a4f1271e820ea0115caee08 based on your suggestion.

PR #9627, opened 2026-01-19
Enable TRIM support for non-(S)ATA disks

maurice-w:
Currently, the rc script invokes `camcontrol identify` to check whether a disk supports TRIM and if so, enables TRIM for the UFS rootfs.

However, `camcontrol identify` only supports (S)ATA disks - it sends an ATA identify command to the disk.
If the disk is SCSI / SAS, VirtIO etc., the test fails. As a result, TRIM remains disabled on all non-(S)ATA disk.

This patch replaces `camcontrol identify` with `diskinfo -v`, which should work on all disks.


fichtner:
LGTM, thanks a lot!

Issue #9266, opened 2025-10-06
HA issue: Manual Sync starts OpenVPN Client on slave


ednt:
**Important notices**

Before you add a new report, we ask you kindly to acknowledge the following:

- [x] I have read the contributing guide lines at https://github.com/opnsense/core/blob/master/CONTRIBUTING.md
- [x] I am convinced that my issue is new after having checked both open and closed issues at https://github.com/opnsense/core/issues?q=is%3Aissue

**Describe the bug**

Manually triggering 'Synchronize and reconfigure all' in a HA Setup with CARP starts OpenVPN Client on Slave.

**To Reproduce**

1. Setup High Availability with 2 OPNsense (Master/Slave)
2. Setup CARP
3. Setup OpenVPN Client using Instances with 'Depend on (CARP)' active
4. Manually trigger 'Synchronize and reconfigure all' under 'System -> High Availability -> Status'

**Expected behavior**

Nothing should happen in regard to the OpenVPN Client.

**Describe alternatives you considered**

Remove OpenVPN from 'Services to synchronize'
Result: Didn't have any effect. Even with all services disabled.

Use Cron for the HA-Sync
Result: OpenVPN Client on Slave doesn't start. Only a manual sync triggers the bug.

**Environment**
2x OPNsense (25.7.4) as VM
HA-Setup (Master/Slave) + CARP on all interfaces
OpenVPN Client (Instance) as S2S Tunnel with 'Depend on (CARP)' active

welcomereality:
Can confirm im seeing simillar behavior 25.7

sat-o-matic:
I see the same behavior on 25.10.1_2.

Monviech:
Im investigating on this as I have another report as well.

swhite2:
0b7c06f should do the trick. If anyone wants to test:

```
# opnsense-patch 0b7c06f
```

Monviech:
I have feedback that the above patch fixes the issue.

sat-o-matic:
I've tested too, the patch seems to have fixed the problem. Nice work!

sat-o-matic:
When will this patch make it into the business edition 25.10 series? Sorry if I'm asking in the wrong place, please help me pointing out where I should be posting instead if that is the case.

AdSchellevis:
@sat-o-matic if I'm not mistaken it's in the latest (25.10.2)

Monviech:
Hello this should already be in 25.10.2 if Im not mistaken.

[...]

PR #9634, opened 2026-01-21
openvpn: account for CARP status in start and restart cases as well

swhite2:
This also inverts the logic such that DISABLED is considered UP as well.

The original logic was likely a deliberate choice, so putting this up as a PR to see if this is valid.

Fixes https://github.com/opnsense/core/issues/9266

### DIFF
--- a/src/opnsense/scripts/openvpn/ovpn_service_control.php
+++ b/src/opnsense/scripts/openvpn/ovpn_service_control.php
@@ -165,8 +165,7 @@ if (isset($opts['h']) || empty($args) || !in_array($args[0], ['start', 'stop', '
             $destroy_if = !empty($instance_stats['dev_type']) && $instance_stats['dev_type'] != $node->dev_type;
             $carp_down = false;
             if ((string)$node->role == 'client' && !empty($vhids[(string)$node->carp_depend_on])) {
-                $status = $vhids[(string)$node->carp_depend_on];
-                $carp_down = $status == 'BACKUP' || $status == 'INIT'; /* consider up if DISABLED */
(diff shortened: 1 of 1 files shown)


swhite2:
To be more specific: when applying from the instances GUI, `openvpn configure` is executed, while a HA sync executes `openvpn restart` only, skipping the CARP check. In the worst case, this hijacks the MASTER openvpn session (and confusing the server, producing TLS errors). We could also fire a `configure` during the sync instead, but it seems to me that if the CARP tracking is enabled on the node, it should apply to all service control options.

src/opnsense/scripts/openvpn/ovpn_service_control.php
```diff
@@ -163,22 +163,27 @@ function get_vhid_status()
         if (flock($statHandle, LOCK_EX)) {
             $instance_stats = ovpn_instance_stats($node, $statHandle);
             $destroy_if = !empty($instance_stats['dev_type']) && $instance_stats['dev_type'] != $node->dev_type;
+            $carp_down = false;
+            if ((string)$node->role == 'client' && !empty($vhids[(string)$node->carp_depend_on])) {
+                $status = $vhids[(string)$node->carp_depend_on];
+                $carp_down = $status == 'BACKUP' || $status == 'INIT'; /* consider up if DISABLED */
```
AdSchellevis:
```suggestion
                $carp_down = $vhids[(string)$node->carp_depend_on] != 'MASTER';
```

disabled isn't a carp status, hence the !MASTER (https://github.com/opnsense/src/blob/98ad27755186ef6a45aa57cab4d2994814e9b3ab/sys/netinet/ip_carp.h#L129)

Issue #9640, opened 2026-01-22
26.1RC1 impossible Edit/Clone Firewall rules if the rule contains protocol "TCP/UDP"


I-G-1-1:
26.1RC1 impossible Edit/Clone Firewall rules if the rule contain protocol "TCP/UDP"

Tried on Safari, Chrome, Firefox.

1. Go to 'ui/firewall/filter/#interface=wan or ui/firewall/filter/#interface=lan'
2. Add a firewall rules with protocol "TCP/UDP"
3. Apply
4. try to Edit/Clone the rule
5. nothing happen.

**Expected behavior**

I can edit or clone the rule

**Describe alternatives you considered**

If the rule hasn't the protocol "TCP/UDP" but only "TCP" there is no issue and the rule can be edited/cloned

**Environment**

Software version used and hardware type if relevant, e.g.:

OPNsense 26.1RC1 (amd64).

fichtner:
I didn't have issues with TCP/UDP while trying a clone but there is an issue with the old value of `tcp` which is not selected in the new GUI and saving will remove it. It seems to want `TCP` in the config now. Not sure how to solve it. Briefly looking at the old page the value was lowercase by default:

https://github.com/opnsense/core/blob/3668d87f5dbc0abe2268e91bf121e2b95bc157a0/src/www/firewall_nat_edit.php#L51

https://github.com/opnsense/core/blob/3668d87f5dbc0abe2268e91bf121e2b95bc157a0/src/www/firewall_nat_edit.php#L145

CC @Monviech @AdSchellevis

fichtner:
Basically all of them were...

https://github.com/opnsense/core/blob/3668d87f5dbc0abe2268e91bf121e2b95bc157a0/src/www/firewall_nat_edit.php#L597

fichtner:
(will fix the unrelated dnat page rant from earlier separately, didn't notice this was about firewall rules)

I-G-1-1:
I switched to development and I'm on 26.1.r_9 now. Still have the same issue. 

It was supposed to be solved in 26.1.r_9?

I found there is also an issue on how the webui shows the TCP/UPD selection:

<img width="896" height="752" alt="Image" src="https://github.com/user-attachments/assets/aea9ca9c-ee3e-46c2-b0fa-70059551c4cf" />

For example If the rules has as Protocol "TLSP" and I select "TCP/UPD" then the dropdown menu shows "TLSP" and doesn't update to "TCP/UPD".
[...]

PR #9643, opened 2026-01-23
mvc: add ChangeCase support to ProtocolField for DNAT special case

fichtner:
References the comments here https://github.com/opnsense/core/issues/9640#issuecomment-3786257696

### DIFF
--- a/src/opnsense/mvc/tests/app/models/OPNsense/Base/FieldTypes/ProtocolFieldTest.php
+++ b/src/opnsense/mvc/tests/app/models/OPNsense/Base/FieldTypes/ProtocolFieldTest.php
@@ -1,7 +1,7 @@
 <?php
 
 /*
- * Copyright (C) 2020 Deciso B.V.
+ * Copyright (C) 2020-2026 Deciso B.V.
  * All rights reserved.
  *
  * Redistribution and use in source and binary forms, with or without


src/opnsense/mvc/app/models/OPNsense/Base/FieldTypes/ProtocolField.php
```diff
@@ -58,27 +60,49 @@ public function setAddOptions($value)
...
+            }
             foreach (explode("\n", file_get_contents('/etc/protocols')) as $line) {
                 if (substr($line, 0, 1) != "#") {
                     $parts = preg_split('/\s+/', $line);
                     if (count($parts) >= 4 && $parts[1] > 0) {
-                        $protocol = trim(strtoupper($parts[0]));
-                        if (!in_array($protocol, $ipv6_ext) && !isset(self::$internalStaticOptionList[$protocol])) {
-                            self::$internalStaticOptionList[$opt_hash][$protocol] = $protocol;
+                        $protocol = $this->applyChangeCase($parts[0]);
+                        if (!in_array(strtoupper($protocol), $ipv6_ext) && !isset(self::$internalStaticOptionList[$protocol])) {
```
fichtner: shouldn't this have been

```suggestion
                        if (!in_array(strtoupper($protocol), $ipv6_ext) && !isset(self::$internalStaticOptionList[$opt_hash][$protocol])) {
```

fichtner: Since we don't repopulate anyway it's probably also ok to drop the second condition

```suggestion
                        if (!in_array(strtoupper($protocol), $ipv6_ext)) {
```

Monviech:
I created multiple rules inside Source NAT, Destination NAT, and Filter Rule, and when checking config.xml the protocols all had either the correct upper case (SNAT, Filter), or lower case (DNAT)

I did only test, I did not evaluate the code.

src/opnsense/mvc/app/models/OPNsense/Firewall/DNat.xml
```diff
@@ -35,8 +35,9 @@
                 <!-- XXX: currently optional, need to decide what to do with it -->
                 <!-- <Required>Y</Required>
                 <Default>any</Default> -->
+                <ChangeCase>lower</ChangeCase>
                 <AddOptions>
-                    <opt1 value="tcp/udp">TCP/UDP</opt1>
+                    <opt1 value="TCP/UDP">TCP/UDP</opt1>
```
fichtner, line 40: this looks odd btw, but is intentional: so that these optional values are also lowercased correctly. Works from the GUI but let me also add a test for this :)

fichtner:
@AdSchellevis I've made sure that 'any' will now be the first option always. The tests for the changes here are (and were already) pretty extensive.

The biggest change here is that 'any' will not be appended when not a required field which is the case for DNat setting a blank 'any' label to not have it in there twice and extra handling  for 'any' being always lowercase.

The second biggest change is that values pushed through save are always forced so 'TcP' is accepted in both cases.

The `internalInitialValue` will also follow the case selection now, something that appears to have been overlooked before.

AdSchellevis:
@fichtner should do the trick indeed, thanks!

src/opnsense/mvc/app/models/OPNsense/Base/FieldTypes/BaseField.php
```diff
@@ -920,17 +919,21 @@ public function setConstraints($constraints)
     }
 
     /**
-     * apply change case to this node, called by setValue
+     * apply configured change case to this node
```
fichtner:
```suggestion
     * apply configured change case to a given value
```

src/opnsense/mvc/app/models/OPNsense/Base/FieldTypes/BaseField.php
```diff
@@ -897,17 +897,16 @@ public function setRequired($value)
...
     public function setChangeCase($value)
     {
         if (strtoupper(trim($value)) == 'UPPER') {
             $this->internalChangeCase = 'UPPER';
         } elseif (strtoupper(trim($value)) == 'LOWER') {
             $this->internalChangeCase = 'LOWER';
-        } else {
-            $this->internalChangeCase = null;
         }
+        /* do not allow changing back to the "null" default */
```
fichtner:
```suggestion
        /* do not allow changing back to the "null" default via typo */
```

src/opnsense/mvc/app/models/OPNsense/Base/FieldTypes/ProtocolField.php
```diff
@@ -58,27 +60,70 @@ public function setAddOptions($value)
...
+
         $this->internalOptionList = self::$internalStaticOptionList[$opt_hash];
     }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function setValue($value)
+    {
+        $new_value = empty($this->internalChangeCase) || strtolower((string)$value) == 'any' ?
```
fichtner:
```suggestion
        $new_value = strtolower((string)$value) == 'any' ?
```

the empty change-case option is prohibited by the protocol field (enforcing uppper by default)

src/opnsense/mvc/tests/app/models/OPNsense/Base/FieldTypes/ProtocolFieldTest.php
```diff
@@ -1,31 +1,29 @@
 <?php
 
-/**
- *    Copyright (C) 2020 Deciso B.V.
+/*
+ * Copyright (C) 2020 Deciso B.V.
```
fichtner:
```suggestion
 * Copyright (C) 2020-2026 Deciso B.V.
```

Issue #9619, opened 2026-01-16
kea: prefix watcher improvements


fichtner:
**Important notices**

Our forum is located at https://forum.opnsense.org , please consider joining discussions there in stead of using GitHub for these matters.

Before you ask a new question, we ask you kindly to acknowledge the following:

- [x] I have read the contributing guide lines at https://github.com/opnsense/core/blob/master/CONTRIBUTING.md
- [x] I am convinced that my issue is new after having checked both open and closed issues at https://github.com/opnsense/core/issues?q=is%3Aissue

1. When the lease file is not yet provided the daemon exits with a traceback. https://forum.opnsense.org/index.php?topic=50303.0
2. Since the daemon uses NDP we could consider a hostwatch-based source for IP addresses.

PR #9648, opened 2026-01-26
kea: Use hostwatch as ndp source in kea_prefix_watcher script

Monviech:
Fixes: https://github.com/opnsense/core/issues/9619

### DIFF
--- a/src/opnsense/scripts/kea/kea_prefix_watcher.py
+++ b/src/opnsense/scripts/kea/kea_prefix_watcher.py
@@ -100,6 +100,7 @@ class Hostwatch:
             return self._def_local_db[mac]
 
 
+
 if __name__ == '__main__':
     parser = argparse.ArgumentParser()
     parser.add_argument('filename',


src/opnsense/scripts/kea/kea_prefix_watcher.py
```diff
@@ -69,24 +70,34 @@ def yield_log_records(filename, poll_interval=5):
...
-            if idx > 0 and len(parts) > 1:
-                try:
-                    addr = parts[0].split('%')[0]
-                    if ipaddress.ip_address(addr).is_link_local:
-                        self._def_local_db[parts[1]] = parts[0]
-                except (ValueError, IndexError):
-                    pass
+        self._def_local_db.clear()
+
+        out = subprocess.run(['/usr/local/sbin/configctl', 'hostwatch', 'dump_full'], capture_output=True, text=True).stdout
```
AdSchellevis: you can skip `configd` here as you don't really need a stepping stone to access the data, a call to `/usr/local/opnsense/scripts/interfaces/list_hosts.py` should do the trick.

src/opnsense/scripts/kea/kea_prefix_watcher.py
```diff
@@ -1,7 +1,7 @@
 #!/usr/local/bin/python3
 
 """
-    Copyright (c) 2025 Ad Schellevis <ad@opnsense.org>
+    Copyright (c) 2025-2026 Ad Schellevis <ad@opnsense.org>
```
fichtner: well, technically you can add yours, but not modify Ad's ;)

src/opnsense/scripts/kea/kea_prefix_watcher.py
```diff
@@ -69,24 +70,34 @@ def yield_log_records(filename, poll_interval=5):
...
+        try:
+            rows = json.loads(out).get("rows", [])
+        except (ValueError, TypeError):
+            return
 
+        for row in rows:
+            try:
+                # [ifname, mac, ip, vendor]
+                mac = row[1]
+                addr = row[2]
```
AdSchellevis: I would just use row[1] and row[2] below as it removes quite some cruft.

src/opnsense/scripts/kea/kea_prefix_watcher.py
```diff
@@ -69,24 +70,34 @@ def yield_log_records(filename, poll_interval=5):
...
+
+                if ipaddress.ip_address(addr).is_link_local:
+                    self._def_local_db[mac] = addr
+            except (ValueError, IndexError):
```
AdSchellevis: the indexerror looks a bit suspicious here, how can we trip into that in this case? (faulty data perhaps, but you probably don't want to skip silently in that case as `list_hosts` is pretty strict in its output)

src/opnsense/scripts/kea/kea_prefix_watcher.py
```diff
@@ -28,6 +28,7 @@
 import argparse
 import csv
 import ipaddress
+import json
```
AdSchellevis: probably doesn't matter much, but `ujson` is faster

src/opnsense/scripts/kea/kea_prefix_watcher.py
AdSchellevis: I'm also ok by replacing it for Deciso B.V. (my name is a bit of a gimmick here)

Monviech: I'll leave it as it is for now.

AdSchellevis (approved):
@Monviech looks good, thanks!

PR #9651, opened 2026-01-26
Services: Kea DHCP: Kea DHCPv6 / Reservations: Lease6 view must distinguish between duid and hwaddr

Monviech:
Api changes, is_reserved is not boolean anymore, but a string containing reference key

For: https://github.com/opnsense/core/commit/b9b9336975fa589c23b2dfc2309b615330fcf7ad

### DIFF
--- a/src/opnsense/mvc/app/controllers/OPNsense/Kea/Api/LeasesController.php
+++ b/src/opnsense/mvc/app/controllers/OPNsense/Kea/Api/LeasesController.php
@@ -105,7 +105,7 @@ abstract class LeasesController extends ApiControllerBase
                 } else {
                     $mac = strtolower($record['hwaddr'] ?? '');
                     if (isset($resv4[$mac])) {
-                        $record['is_reserved'] = $resv6[$mac];
+                        $record['is_reserved'] = $resv4[$mac];
                     }
                 }
             }


src/opnsense/mvc/app/controllers/OPNsense/Kea/Api/LeasesController.php
```diff
@@ -87,13 +93,25 @@ public function searchAction()
                 $mac = strtoupper(substr(str_replace(':', '', $record['hwaddr']), 0, 6));
                 $record['mac_info'] = isset($mac_db[$mac]) ? $mac_db[$mac] : '';
                 // Reservation
+                $record['is_reserved'] = '0';
+                $record['is_reserved_key'] = '';
```
AdSchellevis: for some reason I think we better opt for a single result field, either `is_reserved` with 3 states (e.g. 0,1[mac],2[duid]) or only the key being used, it saves quite some fluff which is only relevant when trying to search the record.

src/opnsense/mvc/app/controllers/OPNsense/Kea/Api/LeasesController.php
```diff
@@ -60,16 +60,22 @@ public function searchAction()
...
+            $resv4[strtolower($reservation->hw_address->getValue())] = 'hwaddr';
         }
 
         foreach ((new KeaDhcpv6())->reservations->reservation->iterateItems() as $reservation) {
-            $resv6[strtolower((string)$reservation->duid)] = true;
+            // At least one of these is required in the model
+            if (!$reservation->duid->isEmpty()) {
+                $resv6[strtolower($reservation->duid->getValue())] = 'duid';
+            }
+            if (!$reservation->hw_address->isEmpty()) {
```
AdSchellevis:
```suggestion
            elseif (!$reservation->hw_address->isEmpty()) {
```

src/opnsense/mvc/app/controllers/OPNsense/Kea/Api/LeasesController.php
Monviech: I will model around the key then. The boolean representation was from the assumption that IPv6 would always only be DUID, and IPv4 would only be HWADDR. With keys it's simpler (for me) to understand the logic than with magic numbers.

PR #9656, opened 2026-01-27
Bootgrid: allow multi word tooltips

byWambo:
Since version 26.x, it hasn't been possible to have command tooltips contain content with multiple words. Here, my command is set up with the title 'Flush aliases'.
```js
                'flush-aliases': {
                    'title': "{{ lang._('Flush aliases') }}",
                    'classname': "fa fa-fw fa-trash",
                    'sequence': 500,
                },
```
image

You can only see 'Flush'.

Adding quotes to the title resolving lines will resolve the issue.
image


swhite2:
Merged, thanks :)

Issue #9655, opened 2026-01-27
[26.1] PF rejects UUID overload table names in 26.1 (table name too long)


daygle:
**Important notices**

Before you add a new report, we ask you kindly to acknowledge the following:

- [X] I have read the contributing guide lines at https://github.com/opnsense/core/blob/master/CONTRIBUTING.md
- [X] I am convinced that my issue is new after having checked both open and closed issues at https://github.com/opnsense/core/issues?q=is%3Aissue

**Describe the bug**

After upgrading to OPNsense 26.1, PF is refusing to load the ruleset whenever a firewall rule uses rate‑limit / max‑src‑conn‑rate options.
The overload table names appear to be UUIDs, which exceed PF's maximum table‑name length.
This results in PF rejecting the entire ruleset.

Error output:
There were error(s) loading the rules: /tmp/rules.debug:317:
table name 'cc63f2df-3dc0-4fe5-a002-b8e7a2d5ade1' too long

The line in question reads [317]:
pass in quick on igc0 inet proto tcp from {any} to $ssh_ipv4 port {22}
keep state ( max 100 max-src-nodes 50 max-src-conn 20 max-src-states 3
tcp.established 300 max-src-conn-rate 2 /60,
overload <cc63f2df-3dc0-4fe5-a002-b8e7a2d5ade1> flush global )
label "4622edd3-7c20-497c-ba73-8c044b3cfcca" # SSH/RL/IPv4

Multiple similar UUID‑style table names are generated for other rules with rate‑limit settings, and PF rejects all of them.

**To Reproduce**

1. Create a firewall rule (e.g., SSH on WAN)
2. Open Advanced Options
3. Enable - Max src‑conn‑rate and Overload table alias.
4. Apply changes
5. PF fails to load ruleset with "table name too long"

**Expected behavior**

Rule to load with overload table.

**Describe alternatives you considered**

You can remove the overload alias from the rule until a fix has been applied.

**Screenshots**

N/A

**Relevant log files**

N/A

**Additional context**

N/A

**Environment**

Software version used and hardware type if relevant, e.g.:

OPNsense 26.1.r_31-amd64
Intel(R) Celeron(R) N5105 @ 2.00GHz (4 cores, 4 threads)

daygle:
Also related, the migration firewall rules import failed due to the same issue. Export of old firewall rules produced alias names rather than uuids. The only way I was able to import was to remove the overload table alias names from the csv.

Monviech:
[...]

PR #9657, opened 2026-01-27
Firewall: Rules [new]: normalize overload table between uuid and name

Monviech:
Fixes: https://github.com/opnsense/core/issues/9655

### DIFF
--- a/src/opnsense/mvc/app/controllers/OPNsense/Firewall/Api/FilterController.php
+++ b/src/opnsense/mvc/app/controllers/OPNsense/Firewall/Api/FilterController.php
@@ -483,12 +483,11 @@ class FilterController extends FilterBaseController
             foreach ((new Alias())->aliases->alias->iterateItems() as $key => $alias) {
                 $aliases[$alias->name->getValue()] = $key;
             }
-            $that = $this;
             return $this->importCsv(
                 'rules.rule',
                 $this->request->getPost('payload'),
                 ['@uuid'],
-                function (&$record) use ($categories, $aliases, $that) {
+                function (&$record) use ($categories, $aliases) {
                     if (!empty($record['categories'])) {
                         /* only map what we know, ignore the rest */
                         $cats = [];
@@ -501,11 +500,6 @@ class FilterController extends FilterBaseController
                     }
                     if (!empty($record['overload']) && isset($aliases[$record['overload']])) {
                         $record['overload'] = $aliases[$record['overload']];
-                    }
-                    if (!empty($record['@uuid']) && !$that->isValidUUID($record['@uuid'])) {
-                        throw new \Exception(
-                            sprintf("Invalid UUID offered (%s)", $record['@uuid'])
-                        );
                     }
                 }
             );
--- a/LICENSE
+++ b/LICENSE
@@ -28,7 +28,7 @@ Copyright (c) 2021 Kyle Evans <kevans@FreeBSD.org>
 Copyright (c) 2015 Manuel Faux <mfaux@conf.at>
 Copyright (c) 2003-2006 Manuel Kasper <mk@neon1.net>
 Copyright (c) 2012 Marcello Coutinho
-Copyright (c) 2018 Martin Wasley <mjwasley@gmail.com>
+Copyright (c) 2017-2020 Martin Wasley <mjwasley@gmail.com>
 Copyright (c) 2022 Maurice Walker <maurice@walker.earth>
 Copyright (c) 2010-2015 Michael Bostock
 Copyright (c) 2018-2021 Michael Muenz <m.muenz@gmail.com>
--- a/src/etc/config.xml.sample
+++ b/src/etc/config.xml.sample
@@ -32,8 +32,8 @@
     </webgui>
     <disablenatreflection>yes</disablenatreflection>
     <usevirtualterminal>1</usevirtualterminal>
-    <ipv6allow>1</ipv6allow>
     <disableconsolemenu/>
+    <ipv6allow>1</ipv6allow>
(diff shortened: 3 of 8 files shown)


src/opnsense/mvc/app/library/OPNsense/Firewall/FilterRule.php
```diff
@@ -254,6 +256,13 @@ private function parseFilterRules()
                     }
                     if (!empty($rule['max-src-conn-rate']) && !empty($rule['max-src-conn-rates'])) {
                         $otbl = !empty($rule['overload']) ? $rule['overload'] : "virusprot";
```
fichtner, line 256: now the migration code falls back to "virusprot" default but this code doesn't because we've already assigned the uuid here. I don't mind it going to virusprot then too at least it works mostly as the user intended without a syntax error

src/opnsense/mvc/app/library/OPNsense/Firewall/FilterRule.php
```diff
@@ -254,6 +256,13 @@ private function parseFilterRules()
...
+                        // If alias name cannot be resolved, assume it is an UUID and resolve its name
+                        if (!empty($otbl) && !Util::isAlias($otbl)) {
+                            $alias = (new Alias())->getNodeByReference('aliases.alias.' . $otbl);
+                            if ($alias !== null && !$alias->name->isEmpty()) {
+                                $otbl = $alias->name->getValue();
+                            }
+                        }
```
AdSchellevis: it's a bit more complicated than this I'm afraid, the rule parser should be able to trust its input, which means we should map uuid to name in 

https://github.com/opnsense/core/blob/b2a376cece20eb7dced96c6cffc40fa1d8b01088/src/opnsense/mvc/app/models/OPNsense/Firewall/FieldTypes/FilterRuleField.php#L51

Next challenge is arranging the input part as people expect the input to understand names as well, which would be around here

https://github.com/opnsense/core/blob/b2a376cece20eb7dced96c6cffc40fa1d8b01088/src/opnsense/mvc/app/controllers/OPNsense/Firewall/Api/FilterController.php#L490

Similar to how categories work.

daygle:
I put together a pull request for this issue and would appreciate your review:

https://github.com/opnsense/core/pull/9658

Monviech:
@daygle But why, we are already working on this issue here in this PR.

daygle:
@Monviech Just wanted to help. Didn't think it would cause any issues.

Monviech:
It's not any issue, its just splitting efforts. Usually when an issue is assigned and there is also already a PR open it means it's being worked on. This here just takes a bit because its a delicate spot.

daygle:
@Monviech Close my pull request if you wish.

src/opnsense/mvc/app/controllers/OPNsense/Firewall/Api/FilterController.php
```diff
@@ -458,6 +458,12 @@ function ($node, $record) use ($categories) {
                         }
                         $record['categories'] = implode(',', $cats);
                     }
+                    if (!empty($record['overload'])) {
+                        $alias = (new Alias())->getNodeByReference('aliases.alias.' . $record['overload']);
```
AdSchellevis: push out of the loop (similar to categories) to avoid performance issues.

src/opnsense/mvc/app/controllers/OPNsense/Firewall/Api/FilterController.php
```diff
@@ -488,6 +494,12 @@ function (&$record) use ($categories) {
                         }
                         $record['categories'] = implode(',', $cats);
                     }
+                    if (!empty($record['overload'])) {
```
AdSchellevis: same, cache and resolve

src/opnsense/mvc/app/models/OPNsense/Firewall/FieldTypes/FilterRuleField.php
```diff
@@ -88,6 +89,15 @@ public function serialize()
                 $result['to_port'] = (string)$this->destination_port;
             }
         }
+        if (!$this->overload->isEmpty()) {
+            $alias = (new Alias())->getNodeByReference('aliases.alias.' . $this->overload->getValue());
```
AdSchellevis: we need to cache this in some way, either add a static to this class or add a function in the util class to resolve a name by uuid (https://github.com/opnsense/core/blob/master/src/opnsense/mvc/app/library/OPNsense/Firewall/Util.php)

src/opnsense/mvc/app/models/OPNsense/Firewall/Filter.xml
```diff
@@ -229,7 +229,7 @@
                 </max>
                 <max-src-conn-rate type="IntegerField">
                     <MinimumValue>1</MinimumValue>
-                    <MaximumValue>2147483647</MaximumValue>
+                    <MaximumValue>4294967</MaximumValue>
```
AdSchellevis, line 232: It's probably ok, but where does this number come from?

AdSchellevis (approved):
@Monviech just one question left, but rest looks good, thanks!

src/opnsense/mvc/app/models/OPNsense/Firewall/Filter.xml
Monviech: I got this per accident during testing and documented it in this commit message:

https://github.com/opnsense/core/pull/9657/changes/3de9452662ca8e9045cdd5a6b707b6135c51a800

AdSchellevis: @Monviech ok, thanks, rebase and merge then :)

PR #9667, opened 2026-01-30
Firewall: Rules [new]: Remember the last selected interface and restore it inside the selectpicker automatically on load

Monviech:
Up for discussion


Monviech:
The volt page already adds a URL hash for the interface which also selects the interface filter on a new page load. That means the groundwork from the view is already laid out to use these for links to the page.

fichtner:
The hash is a bit impractical for the menu idea but that's a minor point for now.

Monviech:
I feel like I want to skip this one for now. The fix here won't help much in practice if the default is already sane.
Closing in favor of: https://github.com/opnsense/core/issues/9691

Issue #9664, opened 2026-01-29
PHP Fatal Error when saving DNAT rule with port range


smart7324:
**Important notices**

Before you add a new report, we ask you kindly to acknowledge the following:

- [X] I have read the contributing guide lines at https://github.com/opnsense/core/blob/master/CONTRIBUTING.md
- [X] I am convinced that my issue is new after having checked both open and closed issues at https://github.com/opnsense/core/issues?q=is%3Aissue

**Describe the bug**

Applying a Destination NAT (Port Forward) rule with a destination port range and firewall rule set to `Register rule` causes a PHP fatal error and crashes the filter reload.
The error is triggered by a call to a non-existent PHP function `sprint()` in `filter.lib.inc`.

```
[29-Jan-2026 20:45:34 Europe/Berlin] PHP Fatal error:  Uncaught Error: Call to undefined function sprint() in /usr/local/etc/inc/filter.lib.inc:661
Stack trace:
#0 /usr/local/etc/inc/filter.lib.inc(46): filter_core_rules_system(Object(OPNsense\Firewall\Plugin), Array)
#1 /usr/local/etc/inc/filter.inc(187): filter_core_bootstrap(Object(OPNsense\Firewall\Plugin))
#2 /usr/local/etc/rc.filter_configure(40): filter_configure_sync(true, true)
#3 {main}
  thrown in /usr/local/etc/inc/filter.lib.inc on line 661
```

**To Reproduce**

1. Go to Firewall → NAT → Destination NAT
2. Create or edit a DNAT rule
3. Set Destination Port to a range (e.g. 51021-51051)
4. Set Redirect target port to a single port (e.g. 51021)
5. Set Firewall rule to `Register rule`
6. Save and apply the rule

**Expected behavior**

No error and successful apply.

**Environment**

OPNsense 26.1 (amd64).

fichtner:
How about [476ad93](https://github.com/opnsense/core/commit/476ad93d6f380c18ee766353c15235e462864398) ?

    # opnsense-patch 476ad93


Cheers,
Franco

smart7324:
Yes, working! Thanks:)

smart7324:
@fichtner one more thing regarding the port range for the DNAT rule translation redirect target port. 

Shouldn't the target port only be the start port of the range as the end gets calculated automatically? The validation also doesn't let me save it when entering a range, but the selection and error message does not make it clear as it suggests to specify a valid port range:

[...]

PR #9668, opened 2026-01-30
Firewall: NAT: Destination NAT: The local-port field does not support range and well-known name

Monviech:
Fixes: https://github.com/opnsense/core/issues/9664

### DIFF
--- a/src/opnsense/mvc/app/views/OPNsense/Firewall/dnat_rule.volt
+++ b/src/opnsense/mvc/app/views/OPNsense/Firewall/dnat_rule.volt
@@ -493,11 +493,17 @@
 
         ajaxGet('/api/firewall/d_nat/list_port_select_options', [], function (data) {
             if (!data || !data.single) return;
-            // XXX: There are multiple port fields, but one of them does not support a range, so we replace the label
+            // local-port does not support port ranges, so we replace the label for clarity
             const singlePortOnly = $.extend(true, {}, data);
             singlePortOnly.single.label = "{{ lang._('Single port') }}";
-            $('#row_rule\\.local-port .port_selector').replaceInputWithSelector(singlePortOnly, false);
-            $('.port_selector').not('#row_rule\\.local-port .port_selector').replaceInputWithSelector(data, false);
+
+            $(".port_selector").each(function () {
+                const opts = $(this).is('#row_rule\\.local-port .port_selector')
+                    ? singlePortOnly
+                    : data;
+
+                $(this).replaceInputWithSelector(opts, false);
+            });
         });
 
         $('#category_filter').parent().find('.dropdown-toggle').prepend('<i class="fa fa-tag" style="margin-right: 6px;"></i>');


src/opnsense/mvc/app/models/OPNsense/Firewall/DNat.xml
```diff
@@ -68,7 +68,7 @@
             <target type="NetworkAliasField"/>
             <local-port type="PortField">
                 <EnableAlias>Y</EnableAlias>
-                <ValidationMessage>Please specify a valid portnumber, name, alias or range.</ValidationMessage>
+                <ValidationMessage>Please specify a valid portnumber or alias.</ValidationMessage>
```
fichtner: name refers to port name I suppose?

Monviech: well-known name, its *not allowed in the model here (e.g. http, or dns)

src/opnsense/mvc/app/views/OPNsense/Firewall/dnat_rule.volt
```diff
@@ -493,9 +493,11 @@
 
         ajaxGet('/api/firewall/d_nat/list_port_select_options', [], function (data) {
             if (!data || !data.single) return;
-            $(".port_selector").each(function () {
```
fichtner, line 496: you should keep this and do a second replace with the modified "data" on the last one.. this might be cleaner

fichtner: or at least use the above loop and modify based on the id?

src/opnsense/mvc/app/models/OPNsense/Firewall/DNat.xml
fichtner: ok fine, let's just not make "portnumber" a thing
```suggestion
                <ValidationMessage>Please specify a valid port number or alias.</ValidationMessage>
```

Monviech: Okay changed all instances of this.

src/opnsense/mvc/app/views/OPNsense/Firewall/dnat_rule.volt
Monviech: I changed it to selection inside the loop

Monviech:
Yeah it's weird still, maybe this needs to be solved at a different layer, aka where the label is generated:

https://github.com/opnsense/core/blob/c264c905042ec7e06b96a1f1d0f082826ec2ccd7/src/opnsense/mvc/app/controllers/OPNsense/Firewall/Api/FilterBaseController.php#L213

fichtner:
this isn't about the label IMO but some other JS hiding the input field that should be shown (or the select needs to show a different value on page load because it's empty)

Monviech:
I just tried, that behavior existed before this PR, so should we look at it separately?

Issue #9650, opened 2026-01-26
Feature Request: Adding an option "Group member attribute" to LDAP Authentication Servers, to allow use of servers which dont use memberOf


jakobsen-lrz:
**Important notices**

Before you add a new report, we ask you kindly to acknowledge the following:

- [x ] I have read the contributing guide lines at https://github.com/opnsense/core/blob/master/CONTRIBUTING.md
- [x ] I am convinced that my issue is new after having checked both open and closed issues at https://github.com/opnsense/core/issues?q=is%3Aissue

**Is your feature request related to a problem? Please describe.**
Our LDAP Servers dont use the memberOf attribute to specify group memberships. It uses a custom Attribute for this purpose
We are currently looking in to migrating to opnsense, but this would make the migration quite a bit more difficult, as the workaround would involve having a Authserver for each group, and specifying that group as the default group for each individual server.

**Describe the solution you like**

I would like an option "Group member attribute",  in the config of LDAP Authentication Servers, that allows specifying the LDAP Attribute to check for group memberships.

**Describe alternatives you considered**

I cant think about a different way to implement it that would work in our case.

**Additional context**

I have implemented this option myself and it seems to be working fine on my Test-Firewall. So i would like to open a PR to have it merged into OPNsense.

jerryhze:
I have a related issue here.  
https://github.com/opnsense/core/issues/9286

If a work around is presented in this post, then I can work around my issue as well, so just cross posting here for reference.

PR #9675, opened 2026-01-30
auth: adding option "group member attribute" to ldap authservers

jakobsen-lrz:
Implements #9650

### DIFF
--- a/src/opnsense/mvc/app/library/OPNsense/Auth/LDAP.php
+++ b/src/opnsense/mvc/app/library/OPNsense/Auth/LDAP.php
@@ -75,7 +75,10 @@ class LDAP extends Base implements IAuthConnector
      * @var null|string user attribute
      */
     private $ldapAttributeUser = null;
-
+    /**
+     * @var null|string group member attribute
+     */
+    private $ldapAttributeMember = 'memberof';
     /**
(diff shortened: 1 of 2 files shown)


jakobsen-lrz:
A little explanation on what this is about. LDAP Authservers currently assume that the memberof attribute is used to specify group memberships. But some LDAP Servers/Systems use a different Attribute for this purpose. This change adds a new Option, allowing to specify which attribute to check for group memberships.

I hope this change gets accepted into OPNsense. And i will fix problems with my changes if necessary.

AdSchellevis:
@jakobsen-lrz it's just a bit busy, but it's on my list to make this configurable as suggested. I'll review the PR later, thanks for doing the work

jakobsen-lrz:
Sorry, im just a bit nervous, it isnt urgent at all. Thanks for your effort

AdSchellevis:
@jakobsen-lrz I implemented it slightly different, but this https://github.com/opnsense/core/commit/c7650c6167b4ff9948d91811f3e4bba41720e6a5 should do the trick. can you try on your end?

jakobsen-lrz:
Looks and works fine on my instance. Thanks for your work and effort.

Issue #9674, opened 2026-01-30
Rules [new] allow Statistic section to resize into single row on section expansion


SeimusS:
**Important notices**

Before you add a new report, we ask you kindly to acknowledge the following:

- [X] I have read the contributing guide lines at https://github.com/opnsense/core/blob/master/CONTRIBUTING.md
- [X] I am convinced that my issue is new after having checked both open and closed issues at https://github.com/opnsense/core/issues?q=is%3Aissue

**Is your feature request related to a problem? Please describe.**
Each field in the new Rule tab is capable to resize into single row if its section is expanded. This is however not working for the Statistic section (column). It would be great if this section could do that too, so we can resize the Rule view into single rows for better visibility.

<img width="379" height="47" alt="Image" src="https://github.com/user-attachments/assets/7ce3ddbd-b4ad-4dcc-9bcf-432d1a13f323" />

**Describe the solution you like**

Possibility to have the Statistic section in a single row if I expand its section.

**Describe alternatives you considered**

To live with it. But honestly the double row and single row combinations kinda destroy the workflow.

**Additional context**

I would like to have something like this on the section expansion.

<img width="378" height="22" alt="Image" src="https://github.com/user-attachments/assets/05d6963f-62dc-4d5d-a5a0-4f19aa98dfe4" />

Monviech:
The decision is made here

https://github.com/opnsense/core/blob/c264c905042ec7e06b96a1f1d0f082826ec2ccd7/src/opnsense/mvc/app/views/OPNsense/Firewall/filter_rule.volt#L553-L562

I think when I made the decision, the icons were volatile without this, not being in the exact same spot all the time. But I can revisit it and see what can be done.

SeimusS:
That would be awesome!

Many thanks in advance.

Monviech:
You can try this, if it becomes more like what you imagined:

`opnsense-patch https://github.com/opnsense/core/commit/cd653dd9afc18eec05e1e0cd04d8ccbcfab67545`

Just execute the patch again to revert it.

Dont forget to hit the grid reset button once. <img width="56" height="55" alt="Image" src="https://github.com/user-attachments/assets/8b2400d9-3ccc-4f54-9aa4-e60fc6ee6d54" />

SeimusS:
Chefs kiss

<img width="377" height="761" alt="Image" src="https://github.com/user-attachments/assets/a7c5c4b5-17b7-46bc-a567-2eccdf723950" />


[...]

PR #9679, opened 2026-01-30
Firewall: Rules [new]: Statistics column is responsive now

Monviech:
This keeps the icon all neatly below each other in 4 rows, but if enough horizontal space is given they are allowed to flex out and take the full space.

- Safes additional space on small screens
image

- Allows large screens to show more
image

Fixes: https://github.com/opnsense/core/issues/9674

### DIFF
--- a/src/opnsense/mvc/app/views/OPNsense/Firewall/filter_rule.volt
+++ b/src/opnsense/mvc/app/views/OPNsense/Firewall/filter_rule.volt
@@ -550,14 +550,9 @@
                             return "";
                         }
 
-                        // Split into two vertical rows
-                        const firstGroup  = parts.slice(0, 2).join(" ");
-                        const secondGroup = parts.slice(2).join(" ");
-
                         return `
                             <div class="stats-cell">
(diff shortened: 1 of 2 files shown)


src/opnsense/mvc/app/views/OPNsense/Firewall/filter_rule.volt
```diff
@@ -1045,11 +1040,20 @@
...
+        container-type: inline-size;
+    }
+
+    .stats-cell > span {
+        white-space: nowrap;
     }
 
-    .stats-cell div {
-        gap: 6px;
+    @container (max-width: 160px) {
```
fichtner, line 1053: we can't we just set it to flex and group each element? if there is space for all on one line that's ok, if one breaks down or two or three it's ok?

Monviech: I just want to prevent visual noise, that all icons are cleanly in a familiar pattern in the default view.

PR #9688, opened 2026-01-31
radvd: allow user controlled hop limit

BPplays:
this PR allows the user to set the `AdvCurHopLimit` option in radvd

### DIFF
--- a/src/etc/inc/plugins.inc.d/radvd.inc
+++ b/src/etc/inc/plugins.inc.d/radvd.inc
@@ -187,9 +187,8 @@ function radvd_configure_do($verbose = false, $ignorelist = [])
         if ($entry->AdvDefaultLifetime->isSet()) {
             $radvdconf .= "    AdvDefaultLifetime {$entry->AdvDefaultLifetime->getValue()};\n";
         }
-        $radvdconf .= "    AdvCurHopLimit {$entry->AdvCurHopLimit->getValue()};\n";
-
         $radvdconf .= sprintf("    AdvLinkMTU %s;\n", !empty($mtu) ? $mtu : 0);
+        $radvdconf .= "    AdvCurHopLimit {$entry->AdvCurHopLimit->getValue()};\n";
         $radvdconf .= "    AdvDefaultPreference {$entry->AdvDefaultPreference->getValue()};\n";
 
         switch ($entry->mode->getValue()) {


src/opnsense/mvc/app/models/OPNsense/Radvd/Radvd.xml
```diff
@@ -127,6 +127,11 @@
                 <NetMaskRequired>Y</NetMaskRequired>
                 <AddressFamily>ipv6</AddressFamily>
             </nat64prefix>
+            <AdvCurHopLimit type="IntegerField">
```
fichtner, line 130: Looks good so far... I wonder if we should make it required and set the 64 default value so we can save the conditional in the radvd.inc file similar to what MinRtrAdvInterval is doing?

src/opnsense/mvc/app/controllers/OPNsense/Radvd/forms/dialogEntry.xml
```diff
@@ -138,6 +138,16 @@
             <visible>false</visible>
         </grid_view>
     </field>
+    <field>
+        <id>entries.AdvCurHopLimit</id>
+        <label>Hop limit</label>
+        <type>text</type>
+        <help>Controls the maximum number of hops a packet can traverse before being discarded</help>
```
fichtner: We should add a sentence that the value "0" is unspecified to match what radvd.conf(5) says about it.

src/opnsense/mvc/app/models/OPNsense/Radvd/Radvd.xml
fichtner:
```suggestion
    <version>1.0.1</version>
```

src/etc/inc/plugins.inc.d/radvd.inc
```diff
@@ -195,6 +187,8 @@ function radvd_configure_do($verbose = false, $ignorelist = [])
         if ($entry->AdvDefaultLifetime->isSet()) {
             $radvdconf .= "    AdvDefaultLifetime {$entry->AdvDefaultLifetime->getValue()};\n";
         }
+        $radvdconf .= "    AdvCurHopLimit {$entry->AdvCurHopLimit->getValue()};\n";
+
         $radvdconf .= sprintf("    AdvLinkMTU %s;\n", !empty($mtu) ? $mtu : 0);
         $radvdconf .= "    AdvDefaultPreference {$entry->AdvDefaultPreference->getValue()};\n";
```
fichtner:
```suggestion
        $radvdconf .= sprintf("    AdvLinkMTU %s;\n", !empty($mtu) ? $mtu : 0);
        $radvdconf .= "    AdvCurHopLimit {$entry->AdvCurHopLimit->getValue()};\n";
        $radvdconf .= "    AdvDefaultPreference {$entry->AdvDefaultPreference->getValue()};\n";
```

fichtner:
merged, thanks!

Issue #9687, opened 2026-01-31
Radvd Config is not generated with Link Local.


Babola123:
**Important notes**

Before adding a new report, please confirm the following:

- [X] I have read the guidelines for submissions at https://github.com/opnsense/core/blob/master/CONTRIBUTING.md.
- [X] I am confident that my issue is new after checking both open and closed issues at https://github.com/opnsense/core/issues?q=is%3Aissue.

**Describe the error**

The os-ndp-proxy-go plugin is supposed to pass on the IPV6 addresses. Radvd should be used for more flexibility. When the interfaces are set to Link Local, no configuration for the interfaces is generated in Radvd.
Radvd cannot be used with link local IPv6 addresses. 

https://www.reddit.com/r/opnsense/comments/1qbzz2r/comment/o2ryk6p/?utm_source=share&utm_medium=web3x&utm_name=web3xcss&utm_term=1&utm_content=share_button

**How to reproduce the error**

Steps to reproduce the problem:

Interfaces:
WAN SLAAC
LAN Link Local

os-ndp-proxy-go
Upstream WAN
Downstream LAN
Proxy router advertisements disabled

Radvd
Interface LAN
Constructor WAN
Mode Stateless

**Expected behavior**

Adjust the configuration for the interface under /var/etc/radvd.conf

**Describe alternatives you have considered**

If the interface, e.g., LAN, is assigned a static IP, it works and a corresponding config is generated.

**Environment**

OPNsense 26.1_4-amd64

fichtner:
You probably forgot to set the constructor to WAN under advanced settings.


Cheers,
Franco

Babola123:
No, unfortunately not.  In the advanced settings for radvd, WAN was selected for Constructor. 
The behavior can be reproduced.

Monviech:
@fichtner It trips into this https://github.com/opnsense/core/blob/70629923bb6aef9a9f3a983725c830f021bd4178/src/etc/inc/plugins.inc.d/radvd.inc#L155

fichtner:
I assume you need an exception for your constructor mode then ;)

Monviech:
I missed that during testing. I'll handle it.

Monviech:
@Babola123 You can try

`opnsense-patch https://github.com/opnsense/core/commit/3bcdae70f742771a4e71e7029009778935d86362`

Afterwards the radvd config should also generate when link-local is used on the advertising interface.

Babola123:
many thanks for your support. It works.

PR #9689, opened 2026-01-31
radvd: When Base6Interface constructor is used, use its primary address for ifcfgipv6

Monviech:
There will not always be a primary address on the advertising interface when Base6Interface is used.

Since this new case only executes if ifcfgipv6 is not already containing an AdvRASrcAddress, it's a minimal change that should not mess with any existing setups.

The automatic RDNSS option will become the Base6Interface IPv6 address this way.

Fixes: https://github.com/opnsense/core/issues/9687

### DIFF
--- a/src/etc/inc/plugins.inc.d/radvd.inc
+++ b/src/etc/inc/plugins.inc.d/radvd.inc
@@ -149,7 +149,6 @@ function radvd_configure_do($verbose = false, $ignorelist = [])
 
         $carp_mode = false;
         $src_addr = false;
-
         $ifcfgipv6 = null;
 
         if (!$entry->AdvRASrcAddress->isEmpty()) {


src/etc/inc/plugins.inc.d/radvd.inc
```diff
@@ -151,7 +151,9 @@ function radvd_configure_do($verbose = false, $ignorelist = [])
         $src_addr = false;
 
         $ifcfgipv6 = !$entry->AdvRASrcAddress->isEmpty() ? $entry->AdvRASrcAddress->getValue() : get_interface_ipv6($dhcpv6if);
-        if (!is_ipaddrv6($ifcfgipv6)) {
+        if (!is_ipaddrv6($ifcfgipv6) && !$entry->Base6Interface->isEmpty()) {
+            $ifcfgipv6 = explode('%', get_interface_ipv6($dhcpv6if, null, 'scoped'), 2)[0];
```
fichtner: I thought about this but loopholing this isn’t a great plan. Either we pivot the primary address to wan or we ignore the ip restriction and leave the condition with null. Maybe WAN IP is the better idea since the proxy needs a global address to work anyway.

Monviech: Does the backend notice if the WAN SLAAC IPv6 address changes its prefix? I would use it as its technically more correct, but I'm worried that e.g., the RDNSS option goes stale.

If that's not noticed then my priority would be either the LLA like proposed now, or null as last choice.

fichtner: on 26.1 radvd is now restarted when the prefix shifted if you mean that.. although from experience that was a rather rare thing with some ISPs

Monviech: Okay then the primary IPv6 from the WAN is good for me. Thanks for the feedback.

Monviech:
@fichtner I think this is good now, it works for me.

src/etc/inc/plugins.inc.d/radvd.inc
```diff
@@ -150,7 +150,16 @@ function radvd_configure_do($verbose = false, $ignorelist = [])
         $carp_mode = false;
         $src_addr = false;
```
fichtner:
```suggestion
```

just because I'm old-fashioned ;)

fichtner (approved):

PR #9699, opened 2026-02-01
Normalize case when checking dnsmasq is_reserved

vilhalmer:
Static hosts with uppercase hardware addresses were erroneously reported as dynamic in the Leases UI. Querying leases from dnsmasq returns them normalized to lowercase, but reading directly from the config reports whatever case the user entered.

I tested this on my live box. I'm very unfamiliar with both PHP and opnsense's codebase, but I think this change is safe since the reservedKeys array isn't used for anything but this comparison and thus nothing else is depending on this matching the case in the configuration.

### DIFF
--- a/src/opnsense/mvc/app/controllers/OPNsense/Dnsmasq/Api/LeasesController.php
+++ b/src/opnsense/mvc/app/controllers/OPNsense/Dnsmasq/Api/LeasesController.php
@@ -83,7 +83,7 @@ class LeasesController extends ApiControllerBase
             if (!empty($host->hwaddr)) {
                 foreach (explode(',', (string)$host->hwaddr) as $hwaddr) {
                     if (!empty($hwaddr)) {
-                        $reservedKeys[] = $hwaddr;
+                        $reservedKeys[] = strtolower($hwaddr);
                     }
                 }
             }


fichtner:
@vilhalmer thanks, we'll discuss this tomorrow

Monviech:
I think this is okay, but we have to do the same with client_id as well:
https://github.com/opnsense/core/blob/9aaf67569488cdd045d7e1963ea4b9e1cd87bd1c/src/opnsense/mvc/app/controllers/OPNsense/Dnsmasq/Api/LeasesController.php#L80

`$reservedKeys[] = strtolower($host->client_id->getValue());`

Monviech:
And I think here we can use getValues() now?

https://github.com/opnsense/core/blob/9aaf67569488cdd045d7e1963ea4b9e1cd87bd1c/src/opnsense/mvc/app/controllers/OPNsense/Dnsmasq/Api/LeasesController.php#L84

I feel like I might clean this up myself, would that be okay?

vilhalmer:
Yep, that's fine! Thanks for looking.

Monviech:
Thank you for showing us the issue, I fixed it with this: https://github.com/opnsense/core/commit/5276f51dc08f5bd145a3ef92ddb2b2849c260513

Issue #9693, opened 2026-01-31
OPNsense FW Rules GUI - simple but critical visibility enhancement


wyae:
**Important notices**

Before you add a new report, we ask you kindly to acknowledge the following:

- [x ] I have read the contributing guide lines at https://github.com/opnsense/core/blob/master/CONTRIBUTING.md
- [ x] I am convinced that my issue is new after having checked both open and closed issues at https://github.com/opnsense/core/issues?q=is%3Aissue

**Is your feature request related to a problem? Please describe.**

In the firewall rules GUI(s):

For people with handicapped eyesight or simply bad monitors it is sometimes hard  to differentiate enabled vs disabled states, especially the logging (which only differs between petrol and gray with similar darkness value). 

**Describe the solution you like**

Replace the icon letter (x and i) in the icons for disabled block/reject or log options with a minus sign, but keep the (background) icon shape and colours.

**Describe alternatives you considered**

Change "disabled" icon colours to be visibly very different from "enabled" ones. But depending on colour handicap that still might be a problem - unless choosing extremely contrasting colours, which probably will break GUI and colour scheme design. 
Or keep the current scheme and symbols - which really is hard to distinguish, especially with respect to whether logging is enabled or disabled.

Monviech:
I have also experienced this, I can see if it can be improved by also changing the fa-icon when toggling the log.

PeterVanDresden:
I also experience this, About 5% (or 9% ?) of male persons have green color blindness (daltonism).

Monviech:
Its already fixed in 26.1.1 with two distinct icons instead of colors.

PR #9704, opened 2026-02-02
firewall: Change toggle_log icon to help visibility of enabled/disabled status

Monviech:
Fixes: https://github.com/opnsense/core/issues/9693

I also had issues seeing enabled and disabled after using it for a while, had to always use the tooltip.

I think this is the best solution, though open for feedback for other fa-icons.

image


fichtner:
I'm mostly ok with this but visually we shouldn't colour/mute action buttons.

Monviech:
Sure I remove the color. I already had feedback from @swhite2 that he is okay with the icons as well.

PR #9708, opened 2026-02-02
Firewall: Rules [new]: Remove schedule formatter from group and automatic rule rows

Monviech:


Monviech:
I had to adjust it slightly, now the following is true:
- Schedule formatter is not evaluated on Tree Group Rows (group row check)
- Schedule formatter is not "Undefined" on automatic rules (typeof string check)
- Schedule formatter still shows up in legacy rules inside automation rules
- Schedule formatter will not show when empty (empty string check)

PR #9713, opened 2026-02-03
Firewall - Rules [new]: Add all rules option to interface selectpicker and make it default selection

Monviech:
I thought about it a lot, a single select "Any" is the best. I could create that quite cleanly.
 
A multiselect would play too much with user expectations ("Why doesnt it show my rules for WAN and LAN, I selected both"), also it opens a can of worms with "Floating" as that selection cannot be easily merged with single interfaces. 
That would argubly make the pretty "strong" GUI a lot worse.

So here we are, explicit choice only:

```interface=__any```, but the ```__any``` only exists in the frontend, the interface parameter is ```null``` in the controller.
```interface=__floating```, but the ```__floating``` only exists in the frontend, the interface parameter is ```""``` in the controller.

image

Fixes: https://github.com/opnsense/core/issues/9691

### DIFF
--- a/src/opnsense/mvc/app/controllers/OPNsense/Dnsmasq/Api/LeasesController.php
+++ b/src/opnsense/mvc/app/controllers/OPNsense/Dnsmasq/Api/LeasesController.php
@@ -76,22 +76,20 @@ class LeasesController extends ApiControllerBase
         $reservedKeys = [];
 
         foreach ((new Dnsmasq())->hosts->iterateItems() as $host) {
-            if (!empty($host->client_id)) {
-                $reservedKeys[] = (string)$host->client_id;
+            if (!$host->client_id->isEmpty()) {
+                $reservedKeys[] = strtolower($host->client_id->getValue());
             }
 
-            if (!empty($host->hwaddr)) {
-                foreach (explode(',', (string)$host->hwaddr) as $hwaddr) {
-                    if (!empty($hwaddr)) {
-                        $reservedKeys[] = $hwaddr;
-                    }
+            if (!$host->hwaddr->isEmpty()) {
+                foreach ($host->hwaddr->getValues() as $hwaddr) {
+                    $reservedKeys[] = strtolower($hwaddr);
                 }
             }
         }
 
         foreach ($records as &$record) {
             $is_ipv6 = Util::isIpv6Address($record['address'] ?? '');
-            $key = $is_ipv6 ? ($record['client_id'] ?? '') : ($record['hwaddr'] ?? '');
+            $key = strtolower($is_ipv6 ? ($record['client_id'] ?? '') : ($record['hwaddr'] ?? ''));
             $record['is_reserved'] = in_array($key, $reservedKeys, true) ? '1' : '0';
         }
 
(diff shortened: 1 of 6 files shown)

### CODE REVIEW

File: src/opnsense/mvc/app/controllers/OPNsense/Firewall/Api/FilterController.php [hunk 1]
```diff
@@ -426,6 +441,9 @@ public function getInterfaceListAction()
             }
         }
 
+        // ALL rules
+        $result['any']['items'][] = $makeItem('__*', gettext('All rules'), $totalRules, 'any');
```
fichtner: not sure if we need magic values for both floating and any at the same time - your code already distinguishes between no selection and any selection but they are the same

I think it's easy to switch the select picker to preselect the any value as ""  on page load even if it is the last item.

Monviech: I had explicit issues with the URL hash being flakey if there is "". I would rather want this to be explicit if I'm allowed to, I have a feeling it will reduce future pain.

fichtner: fair, at least consider dropping the asterisk here if you want to go with underscore prefix... there's no added value from these characters and they might force the browser to do URL encoding when it's not really necessary

Monviech: sure, would you prefer something like ``__`` or ``__any``

Monviech:
If we aim to include this in 26.1.1 I require 2 passing reviews and at least one other tester.

I'd like to include this rather sooner than later because it fixes a few pain points.

Monviech:
> As far as I can tell, this uses an empty string for "floating" and `null` for "any/all rules" on the controller now, sacrificing readability a bit and limiting options in the future. Ideally I'd like to look at a request and spot the intent immediately

It's fair to use something like ``__any`` in the request and not ``null``. I want another opinion by @fichtner before changing the code in this regard.

I chose ``null`` because I reasoned the absence of an interface parameter means you don't want it filtered by interfaces at all. And you also cannot pass things like ``__any, opt1, opt2`` to the controller, which could become a bit weird here most likely:
https://github.com/opnsense/core/blob/c3dd6d56f1d2b64eab3ed761014c1cb836f6dcf2/src/opnsense/mvc/app/controllers/OPNsense/Firewall/Api/FilterController.php#L85

Though I'm open for changes.

kevinchalet:
While you're tweaking the picker, any chance you could also update it so that it uses the description of groups instead of their internal name? Using the description of the virtual interface would be the most logical and would be consistent with the other - non-group - entries listed by the selector.

(you guys did a great job with the new rules UI! ❤️)

Monviech:
@kevinchalet Thank you, but could you create a new ticket for that. The changes in this PR will stay focused on fixing the issue.

kevinchalet:
@Monviech sure: https://github.com/opnsense/core/issues/9715 👍🏻

fichtner:
Tested and looks ok, some small mods and remarks in a separate branch: e4b7d6552909fb1

I'm not sure if the sticky selection is a good step at the moment. It kind of defeats the purpose of the any selection when opening the page. We might learn less about user behaviour if we ship both of these changes at the same time.

fichtner:
Made a ticket from the notes on the commit I mentioned #9719

Issue #9709, opened 2026-02-02
New Rules UI: Edit/Clone Rule popup shows “Nothing selected” after switching interfaces


fuba82:
**Important notices**

- [x] I have read the contributing guide lines at https://github.com/opnsense/core/blob/master/CONTRIBUTING.md
- [x] I am convinced that my issue is new after having checked both open and closed issues at https://github.com/opnsense/core/issues?q=is%3Aissue

---

**Describe the bug**

The new Rules UI generally works and firewall rules are applied correctly.
However, the *Edit / Clone Rule popup* behaves inconsistently when switching between different interfaces.

When changing interfaces, the **Source** and **Destination** fields (and occasionally **Port** fields) randomly show **"Nothing selected"**, even though the underlying values are still present and the rule itself remains correct.

Refreshing the page restores the correct values until interfaces are switched again.

The issue does **not** occur when using the legacy Rules UI.

Last known working version: unknown  
(This issue was not observed with the legacy Rules UI.)

---

**To Reproduce**

1. Open **Firewall → Rules**
2. Select an interface (e.g. *Interface1*)
3. Edit an existing rule  
   → All fields (Source, Destination, Ports) are displayed correctly
4. Switch to a different interface
5. Open or inspect rules  
   → Some edit boxes or combo boxes show **"Nothing selected"**
6. Switch back to the original interface
7. Observe that even more fields now show **"Nothing selected"**
8. Refresh the page without switching interfaces  
   → All fields are displayed correctly again until switch Interface again!

---

**Expected behavior**

When switching between interfaces, all rule fields (Source, Destination, Ports, etc.) should consistently display their configured values in the Edit / Clone Rule popup.

---

**Describe alternatives you considered**

- Refreshing the page temporarily resolves the issue
- Using the legacy Rules UI avoids the problem entirely

---

**Screenshots / Video**

A video demonstrating the issue is available here:  
https://www.youtube.com/watch?v=i9aolMH_kXw

---

**Relevant log files**

No relevant log entries found.  
The issue appears to be UI-related only.

---

**Additional context**

- The issue occurs regardless of browser or operating system
- Tested browsers:
  - Firefox (Windows)
  - Edge (Windows)
  - Brave (Windows)
  - Firefox (CachyOS / Linux)
- Tested after clearing all browser data
- Tested from multiple client machines

---

**Environment**

OPNsense version: 26.1_4 
Architecture: amd64

Hardware:
- CPU: Intel® N100
[...]

PR #9717, opened 2026-02-03
Firewall: Rules [new]: Desync during selectpicker decoration

Monviech:
Fixes: https://github.com/opnsense/core/issues/9709

I verified with logs that a desync happens that makes the mapping fail in some cases. This here fixes the behavior completely, I could not log a desync anymore.

Though this one seems like it needs a closer look from @swhite2 since this might be the wrong spot to fix this, but I don't know.

### DIFF
--- a/src/opnsense/mvc/app/views/OPNsense/Firewall/filter_rule.volt
+++ b/src/opnsense/mvc/app/views/OPNsense/Firewall/filter_rule.volt
@@ -864,6 +864,28 @@
 
         // Hook into add event
         $('#{{formGridFilterRule["edit_dialog_id"]}}').on('opnsense_bootgrid_mapped', function(e, actionType) {
+            if (actionType === 'edit' || actionType === 'clone') {
+                /*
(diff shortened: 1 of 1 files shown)


AdSchellevis:
This feels like fixing a side effect, the usual spot for selectpickers to refresh is :

https://github.com/opnsense/core/blob/40cb82128dc8ba5f8262654a693a9b273d716eb6/src/opnsense/www/js/opnsense_bootgrid.js#L1796

As we overwrite the style here, it currently needs plumbing in the template itself:

https://github.com/opnsense/core/blob/40cb82128dc8ba5f8262654a693a9b273d716eb6/src/opnsense/mvc/app/controllers/OPNsense/Firewall/forms/dialogFilterRule.xml#L245C16-L245C29

My current best guess is we (I probably) forgot to add a selectpicker class in :

https://github.com/opnsense/core/blob/40cb82128dc8ba5f8262654a693a9b273d716eb6/src/opnsense/www/js/opnsense_ui.js#L935

In which case it likely never receives a refresh at the moment (which apparently often works).

PR #9718, opened 2026-02-03
Fix PTR Record Generation for Unbound Host Aliases

daygle:
Unbound was generating PTR records for all aliases when the parent host override used a wildcard, causing "PTR record already exists" warnings. Per documentation, only the first alias should receive a PTR when wildcards are used.

- Only generate PTR for main host override (when not wildcard)
- When wildcard is used, only first alias gets PTR record
- Subsequent aliases no longer generate PTR records
- Only show warning for aliases that shouldn't get PTR records

### DIFF
--- a/src/etc/inc/plugins.inc.d/unbound.inc
+++ b/src/etc/inc/plugins.inc.d/unbound.inc
@@ -610,8 +610,9 @@ function unbound_add_host_entries($ifconfig_details)
                     }
                 }
 
-                foreach ($tmp_aliases as $alias) {
+                foreach ($tmp_aliases as $alias_idx => $alias) {
                     $override_is_main = $alias === $tmp_aliases[0];
+                    $is_first_alias = $alias_idx === 1;
 
                     if ($alias['hostname'] != '') {
                         $alias['hostname'] .= '.';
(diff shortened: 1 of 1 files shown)


fichtner (commented):
FWIW, I think this complicates the code too much to make a prediction about edge cases and possible regressions. It would be nicer to restructure this in full. Perhaps it requires two loops?

AdSchellevis:
@daygle can you share some sample input that currently renders faulty output? given the complexity of this code, it might be better to offer a checkbox with a validation if a PTR record should be generated, which moves part of the problem to where it belongs (user input).

daygle:
@AdSchellevis 

Here’s the warning I’m seeing. It’s reproducible by creating an alias that uses a different domain but the same IP as its parent domain. In my setup, the parent is mail.daygle.net and the alias is mail.dmz.daygle.net.

```
2026-02-05T00:12:23
Warning
unbound
PTR record already exists for mail.dmz.daygle.net(192.168.10.100)
```

I’m completely fine with the idea of a PTR checkbox as you suggested - but the current behaviour definitely needs a fix.

AdSchellevis:
@daygle I mean the data used which results in that error.

daygle:
> @daygle I mean the data used which results in that error.

AdSchellevis:
ok, but logically, we should never create a PTR record automatically for an alias as this will always overlap, which leaves a duplicate host override for the same ip as validation. 

Since it's not a hard error, we can also opt for just adding a checkbox "add PTR" so you can at least disable them if you don't want them (preventing the duplicates as well). 

The only question is which PTR record is currently used by unbound, if it's just the override (and not the alias), we can safely drop them for the aliases.

daygle:
> ok, but logically, we should never create a PTR record automatically for an alias as this will always overlap, which leaves a duplicate host override for the same ip as validation. 
> 
> Since it's not a hard error, we can also opt for just adding a checkbox "add PTR" so you can at least disable them if you don't want them (preventing the duplicates as well). 
> 
> The only question is which PTR record is currently used by unbound, if it's just the override (and not the alias), we can safely drop them for the aliases.

Correct, the alias should not have a PTR.

Issue #9705, opened 2026-02-02
26.1 missed Option TCP any Flag


leebrand:
until version 26.1 it was possible to set TCP Flags to any in Firewall Rules.

This was helpful for solving false Default Deny (DF Flag) in combination with Vlan or asymmetric Routing.

Please bring this Option back.

Thank you.

Monviech:
I think when setting "State Type - Sloppy State" also prevents the default deny from happening when traffic is asymmetric.

https://github.com/opnsense/core/blob/0f6d82af34c24246b3f29928f0bbc045e4e4103d/src/opnsense/mvc/app/models/OPNsense/Firewall/Filter.xml#L18

leebrand:
Sloppy state was trying, but it isnt help.

Monviech:
I'll look into whats missing for parity with the old GUI.

PR #9720, opened 2026-02-04
Firewall: Rules [new]: Add tcpflags_any for parity with legacy rules

Monviech:
Fixes: https://github.com/opnsense/core/issues/9705

### DIFF
--- a/src/opnsense/mvc/app/library/OPNsense/Firewall/FilterRule.php
+++ b/src/opnsense/mvc/app/library/OPNsense/Firewall/FilterRule.php
@@ -198,7 +198,7 @@ class FilterRule extends Rule
             }
             // restructure flags
             if (isset($rule['protocol']) && $rule['protocol'] == "tcp") {
-                if (isset($rule['tcpflags_any'])) {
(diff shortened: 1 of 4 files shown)


src/opnsense/mvc/app/library/OPNsense/Firewall/FilterRule.php
```diff
@@ -198,7 +198,7 @@ private function parseFilterRules()
             }
             // restructure flags
             if (isset($rule['protocol']) && $rule['protocol'] == "tcp") {
-                if (isset($rule['tcpflags_any'])) {
```
fichtner, line 201: this needs double checking

fichtner: and it seems ok to me, the legacy page also uses !empty() to write the property into the config

Monviech:
@AdSchellevis I had the exact same thoughts, yet then I thought about migration between old and new and the more complicated validation, and I thought better keep it exactly like this.

Issue #9730, opened 2026-02-05
Feature Request: Implement missing ICMP-Types


jakobsen-lrz:
**Important notices**

Before you add a new report, we ask you kindly to acknowledge the following:

- [x] I have read the contributing guide lines at https://github.com/opnsense/core/blob/master/CONTRIBUTING.md
- [x] I am convinced that my issue is new after having checked both open and closed issues at https://github.com/opnsense/core/issues?q=is%3Aissue

**Is your feature request related to a problem? Please describe.**

Some ICMP-Types, e.g. trace, cant be selected when creating a filter rule.

**Describe the solution you like**

Adding missing ICMP-Types

fichtner:
I think you need to be more precise: which GUI page and which input is missing (e.g. type number would help).


Cheers,
Franco

jakobsen-lrz:
This is about editing/adding rules under `/ui/firewall/filter/` or `Firewall/Rules (new)`

I wanted to implement the following ICMP-Types:

```
     30	  trace		  Traceroute
     31	  dataconv	  Data conversion problem
     32	  mobredir	  Mobile host redirection
     35	  mobregreq	  Mobile registration request
     36	  mobregrep	  Mobile registration reply
     39	  skip		  SKIP
     40	  photuris	  Photuris
```

PR #9731, opened 2026-02-05
Firewall/Filter: Implementing missing ICMPTYPES

jakobsen-lrz:
Implement #9730

Hello, this implements some missing ICMP-Types for the New Rules API/GUI. It seems that it is only necessary to add them to the xml model of filter rules. I hope these changes are fine. Thanks.

### DIFF
--- a/src/opnsense/mvc/app/models/OPNsense/Firewall/Filter.xml
+++ b/src/opnsense/mvc/app/models/OPNsense/Firewall/Filter.xml
@@ -89,22 +89,35 @@
                 <icmptype type="OptionField">
                     <Multiple>Y</Multiple>
                     <OptionValues>
-                        <echoreq>Echo Request</echoreq>
-                        <echorep>Echo Reply</echorep>
-                        <unreach>Destination Unreachable</unreach>
-                        <squench>Source Quench (Deprecated)</squench>
-                        <redir>Redirect</redir>
-                        <althost>Alternate Host Address (Deprecated)</althost>
-                        <routeradv>Router Advertisement</routeradv>
-                        <routersol>Router Solicitation</routersol>
-                        <timex>Time Exceeded</timex>
-                        <paramprob>Parameter Problem</paramprob>
-                        <timereq>Timestamp</timereq>
-                        <timerep>Timestamp Reply</timerep>
-                        <inforeq>Information Request (Deprecated)</inforeq>
-                        <inforep>Information Reply (Deprecated)</inforep>
-                        <maskreq>Address Mask Request (Deprecated)</maskreq>
-                        <maskrep>Address Mask Reply (Deprecated)</maskrep>
+                        <Common>
+                            <echoreq>Echo Request</echoreq>
+                            <echorep>Echo Reply</echorep>
+                            <unreach>Destination Unreachable</unreach>
+                            <redir>Redirect</redir>
+                            <routeradv>Router Advertisement</routeradv>
+                            <routersol>Router Solicitation</routersol>
+                            <timex>Time Exceeded</timex>
+                            <paramprob>Parameter Problem</paramprob>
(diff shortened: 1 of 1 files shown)


Monviech:
Does this complete the list of types as referenced here?

https://man.freebsd.org/cgi/man.cgi?icmp

fichtner:
Well it says "The abbreviations for the types and codes may be used in rules in `pf.conf(5)`." ;)

Monviech:
I meant more like, if we touch this, is this complete now? e.g. why no

ipv6-where, ipv6-here?

fichtner:
fair. I'll say let's just sync the list and migrate the text to numbers :>

jakobsen-lrz:
I left those out on purpose because it seemed to me that OPNsense does some stuff in the background when using ICMP (Not IPV6-ICMP) Protocol with inet6. But i can absolutely implement both of these as well, if you want me to. Else all other ICMP and ICMPv6 Types seem to be implemented.

jakobsen-lrz:
@Monviech I implemented both now.

jakobsen-lrz:
@fichtner Sorry, but what am i supposed to do? Change the icmptypes in Filter.xml to be like the icmp6types? Wouldnt this require some other changes to handle preexisting rules?

fichtner:
Sorry that part as a joke because @Monviech doesn't like it ;)

src/opnsense/mvc/app/models/OPNsense/Firewall/Filter.xml
```diff
@@ -105,6 +105,15 @@
                         <inforep>Information Reply (Deprecated)</inforep>
                         <maskreq>Address Mask Request (Deprecated)</maskreq>
                         <maskrep>Address Mask Reply (Deprecated)</maskrep>
+                        <trace>Traceroute</trace>
```
fichtner: I wonder which of these are deprecated as well and if we should move the deprecated ones to the bottom of the list?

jakobsen-lrz: Not all of them are deprecated some are just obscure and unused. I think they are ordered by type id, so i continued with that.

jakobsen-lrz: I just looked into it. Almost all of the ICMPTypes i added are actually deprecated.

jakobsen-lrz: I edited my changes again, so that the newly added deprecated types, are marked as such. I hope this PR can go forward now.

Monviech:
@fichtner how about we optgroup between deprecated and non deprecated instead?

Monviech:
@jakobsen-lrz If you could do something like this I would prefer that. And remove the deprecated from the label

```
                <icmptype type="OptionField">
                    <Multiple>Y</Multiple>
                    <OptionValues>
                        <Common>
                            <echoreq>Echo Request</echoreq>
                            <echorep>Echo Reply</echorep>
                        </Common>
                        <Deprecated>
                            <squench>Source Quench</squench>
                        </Deprecated>
                    </OptionValues>
                </icmptype>
```

That will pool them automatically in the GUI and we don't have to write deprecated into each one.

fichtner:
Yes please. MVC magic is nice.

jakobsen-lrz:
@Monviech @fichtner I just updated it again. Seems to work fine on my opnsense instance.

Monviech (approved):
Looks good to me and works

Monviech:
@jakobsen-lrz Merged thank you :)

fichtner:
nice one, thanks a lot! ❤️

jakobsen-lrz:
Thanks.

Issue #9680, opened 2026-01-30
26.1 (GM) Firewall Group Name Change Affects Internet Connectivity


2tqkfqv2yy-cmyk:
**Important notices**

Before you add a new report, we ask you kindly to acknowledge the following:

- [x] I have read the contributing guide lines at https://github.com/opnsense/core/blob/master/CONTRIBUTING.md
- [x] I am convinced that my issue is new after having checked both open and closed issues at https://github.com/opnsense/core/issues?q=is%3Aissue

**Describe the bug**

Changing firewall group names in OpnSense 26.1.0 (GM) web GUI causes internet connectivity to be lost, with traffic being caught by the catch-all rule instead of allowing access based on the updated group rules.

**To Reproduce**

1. Ensure you have an OpnSense firewall version 26.1.0 (GM) with groups configured for firewall rules.
2. Create and configure three groups:
   - `all_internal`: Sequence 11, containing general internet access rules
   - `priv_internal`: Sequence 9, containing internal rules specific to certain VLANs
   - `iot_internal`: Sequence 9, containing internal rules specific to certain VLANs
3. Edit the group names in the web GUI:
   - Change `all_internal` to `GRP_all_int`
   - Change `priv_internal` to `GRP_priv_int`
   - Change `iot_internal` to `GRP_iot_int`
4. Click `[Apply]` to save changes
5. Attempt to access the internet (e.g., visit google.com)
6. Observe that traffic is being caught by the catch-all rule instead of allowing internet access

**Expected behavior**

After changing firewall group names, internet access should still function correctly based on the configured rules within those groups.

**Describe alternatives you considered**

*None*

**Additional context**

* This issue was observed after changing group names and clicking `[Apply]`.
* Internet access is restored after rolling back group name changes and clicking `[Apply]`.
* Mentioned in forum post here: https://forum.opnsense.org/index.php?topic=50596.msg258436#msg258436

**Environment**

Software version used and hardware type if relevant, e.g.:

OPNsense 26.1 GM (amd64).

2tqkfqv2yy-cmyk:
As mentioned in the forum, one other user has been able to independently reproduce this so far in their environment.

Monviech:
I'll look into this. I assume some references are not updated.

AdSchellevis:
best check the ruleset before and after, easily to do by creating a copy of `/tmp/rules.debug` before the rename and using diff to compare.

```
cp  /tmp/rules.debug /tmp/rules.debug.mytest
[...]

PR #9734, opened 2026-02-05
Firewall: Rules [new]: Fix group rename in source_net, destination_net and SNAT/DNAT target fields

Monviech:
Fixes: https://github.com/opnsense/core/issues/9680

### DIFF
--- a/src/opnsense/mvc/app/models/OPNsense/Firewall/Group.php
+++ b/src/opnsense/mvc/app/models/OPNsense/Firewall/Group.php
@@ -82,8 +82,7 @@ class Group extends BaseModel
             $interfaces = explode(",", (string)$node->interface);
             // interface works the same for legacy and mvc rules
             if (in_array($oldname, $interfaces)) {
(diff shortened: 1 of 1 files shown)


src/opnsense/mvc/app/models/OPNsense/Firewall/Group.php
```diff
@@ -80,17 +80,36 @@ public function refactor($oldname, $newname)
...
                     $has_changed = true;
                 }
+                // mvc rules (source_net...)
+                $field = $net . '_net';
+                $value = (string)$node->$field;
+                if (!empty($value)) {
+                    $nets = explode(',', $value);
+                    if (in_array($oldname, $nets)) {
+                        unset($nets[array_search($oldname, $nets)]);
+                        $nets[] = $newname;
```
fichtner: it might be nicer to do this in place btw


```suggestion
                        $nets[array_search($oldname, $nets)] = $newname;
```

PR #9738, opened 2026-02-05
Firewall: Rules [new]: The SKIP icmp type is not a valid name for pf, most likely a bug. Number would work.

Monviech:
Doesnt work:
```There were error(s) loading the rules: /tmp/rules.debug:230: syntax error - The line in question reads [230]: pass in quick inet proto icmp from {any} to {any} icmp-type {skip} keep state label "4bcf8573-1db7-428f-80c5-6bd25d6ba04b"```

Works:
```pass in quick inet proto icmp from {any} to {any} icmp-type {39} keep state label "e552e61c-5054-4daa-b2be-c288ec0b6e45"```

### DIFF
--- a/src/opnsense/mvc/app/models/OPNsense/Firewall/Filter.xml
+++ b/src/opnsense/mvc/app/models/OPNsense/Firewall/Filter.xml
@@ -116,7 +116,7 @@
                             <ipv6-here>IPv6 i-am-here</ipv6-here>
                             <mobregreq>Mobile registration request</mobregreq>
                             <mobregrep>Mobile registration reply</mobregrep>
-                            <skip>SKIP</skip>
+                            <!-- <skip>SKIP</skip> is valid but pf does not like the name, number would work -->
                         </Deprecated>
                     </OptionValues>
                 </icmptype>


Monviech:
@jakobsen-lrz Just as info, don't worry though, happens. The spec is correct, but pf doesn't like this one. xD

jakobsen-lrz:
@Monviech 
ok. Should i do something about this? So should i open a PR with a fix?

fichtner:
I think this is a parser error since "skip" keyword exists

The list is here including skip: https://github.com/opnsense/src/blob/c961e158e2727ab1418ab434ec040c1f8eed964a/sbin/pfctl/pfctl_parser.c#L82-L108

fichtner:
Huh but why... are we ignoring our own code? 79c6142803f5f

Monviech:
> @Monviech ok. Should i do something about this? So should i open a PR with a fix?

No, I just wanted to CC you so you can see that when adding something its a good idea to test all options once. It's what I did after merging, and what I usually do when something is added. But it's okay, it's just fyi, happens to everybody. :)

fichtner:
@Monviech this looks more feasible 48d3a73343eae9fd .. it's a tiny bit risky but we don't have to rush it either

Monviech:
@fichtner I don't think we should do anything risky for a deprecated icmp option only. I would just have commented it out explicitely, maybe I'm thinking too pragmatic?

fichtner:
@Monviech talking to @AdSchellevis we're doing this and keeping it on master which gives me some time to add tests for rules rendering :)

Issue #8761, opened 2025-06-03
Captive Portal: IPv6 support


swhite2:
**Important notices**

Before you add a new report, we ask you kindly to acknowledge the following:

- [x] I have read the contributing guide lines at https://github.com/opnsense/core/blob/master/CONTRIBUTING.md
- [x] I am convinced that my issue is new after having checked both open and closed issues at https://github.com/opnsense/core/issues?q=is%3Aissue

**Is your feature request related to a problem? Please describe.**

It would be nice for IPv6-only or dual-stack networks to be able to use the captive portal system as well. Currently only IPv4 is supported.

**Describe the solution you like**

No functional changes on the user-facing Captive Portal administration end, just automatic compatibility with IPv6.

Technically, the impact would be as follows:

- The captive portal web interface should bind to `::1` as well.
- Redirection should take place to `::1`.
- The current firewall rules already handle IPv4 and IPv6 correctly (except for redirection).
- Authentication should account for NDP entries as well, matching on MAC address and inserting the relevant IPv6 address(es) in the zone pf table.
- The background process synchronizing the zones must have an accurate reflection of the NDP table at all times, which may be quite expensive as this table can grow a lot larger than the ARP table. In an ideal scenario one should be able to subscribe to NDP state changes, but it as yet unclear whether this is possible.
- Ethernet accounting rules must match on `0x86DD`. Initial experiments show some IPv6 traffic matching on `0x0800` (IPv4) as well, producing some garbage entries which may unnecessarily grow (and slow down) the accounting output.
- Usage of multiple IPv6 addresses on clients can be problematic, see note below.
- Other unkowns.

Enabling users to use multiple IP addresses on the same client should be considered as well (although perhaps optionally as a toggle). This is relevant for both IPv4 (virtual IPs) and IPv6, however, this scenario is more common in IPv6 with the option of privacy extensions on SLAAC, which may keep old IPv6 addresses active and actively used by clients.

**Describe alternatives you considered**

N/A

**Additional context**

Initial work provided in https://github.com/opnsense/core/tree/cp_ipv6

agoodkind:
+1 on this

agoodkind:
[...]

Issue #5966, opened 2022-08-20
No IPv6 Web GUI access if LAN interface is set to 'Track Interface'


maurice-w:
**Important notices**

Before you add a new report, we ask you kindly to acknowledge the following:

- [x] I have read the contributing guide lines at https://github.com/opnsense/core/blob/master/CONTRIBUTING.md
- [x] I am convinced that my issue is new after having checked both open and closed issues at https://github.com/opnsense/core/issues?q=is%3Aissue

**Describe the bug**

If a LAN interface is tracking a WAN interface and the Web GUI is configured to only listen on this LAN interface, the Web GUI is inaccessible via IPv6.

**To Reproduce**

1. Make sure the WAN and LAN interfaces have default settings (WAN: DHCPv4 + DHCPv6 with prefix delegation, LAN: Static IPv4 + 'Track Interface' IPv6).
3. Go to 'System: Settings: Administration', set the 'Web GUI Listen Interfaces' to 'LAN', save and reboot.
4. See error: The Web GUI is not accessible via the LAN IPv6 address.
5. Access the Web GUI via the LAN IPv4 address, go to 'System: Diagnostics: Services' and restart the webgui service.
6. See result: The Web GUI is now accessible via the LAN IPv6 address.

**Expected behavior**

The Web GUI should be accessible via IPv6 without having to manually restart the service after a reboot.

**Relevant log files**

Excerpt from `/var/etc/lighty-webConfigurator.conf` after a reboot:
```
## bind to port (default: 80)
server.bind  = "127.0.0.1"
server.port  = 80
$SERVER["socket"] == "127.0.0.1:80" {
}
$SERVER["socket"] == "[::1]:80" {
}
$SERVER["socket"] == "192.168.1.1:80" {
}
```

After webgui service restart:
```
## bind to port (default: 80)
server.bind  = "127.0.0.1"
server.port  = 80
$SERVER["socket"] == "127.0.0.1:80" {
}
$SERVER["socket"] == "[::1]:80" {
}
$SERVER["socket"] == "192.168.1.1:80" {
}
$SERVER["socket"] == "[2001:db8:1234:5600:215:5dff:fed2:761b]:80" {
}
```

**Additional context**

It works when not selecting specific listen interfaces (`$SERVER["socket"] == "[::]:80"`).

**Environment**

OPNsense 23.1.a_102 (amd64, OpenSSL)
Hyper-V Gen2 Ver10.0

maurice-w:
[...]

PR #9745, opened 2026-02-06
Captive portal: IPv6 support

swhite2:
Closes https://github.com/opnsense/core/issues/8761

### DIFF
--- a/src/opnsense/mvc/app/controllers/OPNsense/CaptivePortal/Api/AccessController.php
+++ b/src/opnsense/mvc/app/controllers/OPNsense/CaptivePortal/Api/AccessController.php
@@ -88,26 +88,17 @@ class AccessController extends ApiControllerBase
      */
     protected function getClientIp()
     {
-        $clientAddress = $this->request->getClientAddress();
-        $forwardedFor = $this->request->getHeader('X-Forwarded-For');
-        $realIp = $this->request->getHeader('X-Real-Ip');
-
         // determine original sender of this request
         if (
-            $forwardedFor != "" &&
-            (explode('.', $clientAddress)[0] == '127' || $clientAddress === '::1')
+            $this->request->getHeader('X-Forwarded-For') != "" &&
+            explode('.', $this->request->getClientAddress())[0] == '127'
         ) {
             // use X-Forwarded-For header to determine real client
-            return $forwardedFor;
+            return $this->request->getHeader('X-Forwarded-For');
+        } else {
+            // client accesses the Api directly
+            return $this->request->getClientAddress();
         }
-
-        if ($realIp != "" && $clientAddress === '::1') {
-            // API dispatcher may proxy via IPv6 loopback while preserving the client in X-Real-Ip
-            return $realIp;
-        }
-
-        // client accesses the Api directly
-        return $clientAddress;
     }
 
     protected function getClientMac($ip)
@@ -305,7 +296,7 @@ class AccessController extends ApiControllerBase
                                 (string)$cpZone->zoneid,
                                 $userName,
                                 $clientIp,
-                                $authServerName,
+                                $authServerName
                             ]
                         );
                         $CPsession = json_decode($CPsession, true);
--- a/src/opnsense/scripts/captiveportal/lib/pf.py
+++ b/src/opnsense/scripts/captiveportal/lib/pf.py
@@ -51,10 +51,6 @@ class PF(object):
     @staticmethod
     def add_to_table(zoneid, address):
         subprocess.run(['/sbin/pfctl', '-t', f'__captiveportal_zone_{zoneid}', '-T', 'add', address], capture_output=True)
-        # Clear pre-auth states so newly authorized sibling addresses can use the updated table match immediately.
-        subprocess.run(['/sbin/pfctl', '-k', f'{address}'], capture_output=True)
-        wildcard = '::/0' if PF._is_ipv6(address) else '0.0.0.0/0'
-        subprocess.run(['/sbin/pfctl', '-k', wildcard, '-k', f'{address}'], capture_output=True)
 
     @staticmethod
     def remove_from_table(zoneid, address):
--- a/src/opnsense/service/templates/OPNsense/Captiveportal/lighttpd-zone.conf
+++ b/src/opnsense/service/templates/OPNsense/Captiveportal/lighttpd-zone.conf
@@ -5,21 +5,8 @@
         {# found zone, search for interface ip #}
         {% for intf_tag in item.interfaces.split(',') %}
             {% for conf_key, conf_inf in interfaces.items() %}
-                {% if conf_key == intf_tag %}
-                    {% if conf_inf.ipaddrv6 and helpers.is_ipv6(conf_inf.ipaddrv6) %}
-                        {% do item.update({'interface_hostaddr_v6': conf_inf.ipaddrv6}) %}
-                    {% endif %}
-                    {% if conf_inf.ipaddr and conf_inf.ipaddr != 'dhcp' %}
-                        {% do item.update({'interface_hostaddr_v4': conf_inf.ipaddr}) %}
-                    {% endif %}
-                    {# prefer IPv6 if available, fallback to IPv4 #}
-                    {% if conf_inf.ipaddrv6 and helpers.is_ipv6(conf_inf.ipaddrv6) %}
-                        {% do item.update({'interface_hostaddr':conf_inf.ipaddrv6, 'interface_hostaddr_family':'ipv6'}) %}
-                    {% endif %}
-                    {# fallback to IPv4 if IPv6 not found #}
-                    {% if not item.interface_hostaddr and conf_inf.ipaddr and conf_inf.ipaddr != 'dhcp' %}
-                        {% do item.update({'interface_hostaddr':conf_inf.ipaddr, 'interface_hostaddr_family':'ipv4'}) %}
-                    {% endif %}
+                {% if conf_key == intf_tag and conf_inf.ipaddr and conf_inf.ipaddr != 'dhcp' %}
+                    {% do item.update({'interface_hostaddr': conf_inf.ipaddr}) %}
                 {% endif %}
             {% endfor %}
         {% endfor %}
@@ -30,49 +17,21 @@
 
 {% if cp_zone_item != None %}
     {% if cp_zone_item.servername|default("") != "" %}
-      {% set servername_family = 'ipv6' if helpers.is_ipv6(cp_zone_item.servername) else 'ipv4' %}
-      {% do cp_zone_item.update({
-        'interface_hostaddr': cp_zone_item.servername,
-        'interface_hostaddr_family': servername_family
-      }) %}
+      {% do cp_zone_item.update({'interface_hostaddr':cp_zone_item.servername}) %}
     {% endif %}
 
     {# generate zone redirect address #}
     {% if not cp_zone_item.interface_hostaddr %}
-      # interface address not found    {% else %}
-      {% set redirect_scheme = 'https://' if cp_zone_item.certificate|default('') != '' else 'http://' %}
-      {% set ipv4_host = cp_zone_item.interface_hostaddr_v4|default('') %}
-      {% set ipv6_host = cp_zone_item.interface_hostaddr_v6|default('') %}
-      {% set fallback_host = cp_zone_item.interface_hostaddr %}
-      {% set fallback_family = cp_zone_item.interface_hostaddr_family %}
-      {% if cp_zone_item.servername|default('') != '' %}
-        {% if fallback_family == 'ipv6' %}
-          {% set fallback_formatted = '[' + fallback_host + ']' %}
-          {% set fallback_match = '\\[' + fallback_host.replace(':','\\:') + '\\]' %}
-        {% else %}
-          {% set fallback_formatted = fallback_host %}
-          {% set fallback_match = fallback_host.replace('.','\.') %}
-        {% endif %}
-        {% do cp_zone_item.update({
-          'redirect_host_v4': redirect_scheme + fallback_formatted + ':' ~ (cp_zone_item.zoneid|int + 8000) ~ '/index.html',
-          'redirect_host_v6': redirect_scheme + fallback_formatted + ':' ~ (cp_zone_item.zoneid|int + 8000) ~ '/index.html',
-          'redirect_host_match_v4': fallback_match ~ ':' ~ (cp_zone_item.zoneid|int + 8000),
-          'redirect_host_match_v6': fallback_match ~ ':' ~ (cp_zone_item.zoneid|int + 8000)
-        }) %}
-      {% else %}
-        {% set host_v4 = ipv4_host if ipv4_host else fallback_host %}
-        {% set host_v6 = ipv6_host if ipv6_host else fallback_host %}
-        {% set formatted_v4 = host_v4 %}
-        {% set formatted_v6 = '[' + host_v6 + ']' if helpers.is_ipv6(host_v6) else host_v6 %}
-        {% set match_v4 = host_v4.replace('.','\.') if not helpers.is_ipv6(host_v4) else '\\[' + host_v4.replace(':','\\:') + '\\]' %}
-        {% set match_v6 = '\\[' + host_v6.replace(':','\\:') + '\\]' if helpers.is_ipv6(host_v6) else host_v6.replace('.','\.') %}
-        {% do cp_zone_item.update({
-          'redirect_host_v4': redirect_scheme + formatted_v4 + ':' ~ (cp_zone_item.zoneid|int + 8000) ~ '/index.html',
-          'redirect_host_v6': redirect_scheme + formatted_v6 + ':' ~ (cp_zone_item.zoneid|int + 8000) ~ '/index.html',
-          'redirect_host_match_v4': match_v4 ~ ':' ~ (cp_zone_item.zoneid|int + 8000),
-          'redirect_host_match_v6': match_v6 ~ ':' ~ (cp_zone_item.zoneid|int + 8000)
-        }) %}
-      {% endif %}
+      # interface address / servername not found
+    {% elif cp_zone_item.certificate|default("") != "" %}
+      # ssl enabled, redirect to https
+      {% do cp_zone_item.update({'redirect_host': 'https://' ~ cp_zone_item.interface_hostaddr ~ ':' ~ (cp_zone_item.zoneid|int + 8000) ~ '/index.html'}) %}
+    {% else %}
+      # ssl disabled, redirect to http
+      {% do cp_zone_item.update({'redirect_host': 'http://' ~ cp_zone_item.interface_hostaddr ~ ':' ~ (cp_zone_item.zoneid|int + 8000) ~ '/index.html'}) %}
+    {% endif %}
+    {% if cp_zone_item.interface_hostaddr %}
+        {% do cp_zone_item.update({'redirect_host_match':cp_zone_item.interface_hostaddr.replace('.','\.') ~ ':' ~ (cp_zone_item.zoneid|int + 8000) }) %}
     {% endif %}
 
 #############################################################################################
@@ -122,44 +81,38 @@ server.bind              = "0.0.0.0"
 #### bind to port
 server.port              = {{ cp_zone_item.zoneid|int + 8000  }}
 
-$SERVER["socket"] == "[::]:{{ cp_zone_item.zoneid|int + 8000 }}" {
-	$HTTP["host"] !~ "(.*{{cp_zone_item.redirect_host_match_v6}}.*)" {
-		$HTTP["host"] =~ "([^:/]+)" {
-			url.redirect = ( "^(.*)$" => "{{cp_zone_item.redirect_host_v6}}?redirurl=%1$1")
-		}
-	}
-{% if cp_zone_item.certificate|default("") != "" %}
-	ssl.engine = "enable"
-	ssl.pemfile = "/var/etc/cert-cp-zone{{cp_zone_item.zoneid}}.pem"
-{# set ca-file if ca is provided #}
-{% for certItem in helpers.toList('cert') %}
-{%    if certItem.refid == cp_zone_item.certificate and certItem.caref %}
-	ssl.ca-file = "/var/etc/ca-cp-zone{{cp_zone_item.zoneid}}.pem"
-{%    endif %}
-{% endfor %}
-	ssl.openssl.ssl-conf-cmd = (
-	    "MinProtocol" => "TLSv1.2",
-	    "Options" => "-ServerPreference",
-	    "CipherString" => "EECDH+AESGCM:AES256+EECDH:CHACHA20:!SHA1:!SHA256:!SHA384"
-	)
-{% else %}
-	ssl.engine = "disable"
-{% endif %}
-}
-
 ## Redirect response code to use
 url.redirect-code        = 302
+
+$HTTP["host"] !~ "(.*{{ cp_zone_item.redirect_host_match }}.*)" {
+	$HTTP["host"] =~ "([^:/]+)" {
+		url.redirect = ( "^(.*)$" => "{{ cp_zone_item.redirect_host }}?redirurl=%1$1")
+	}
+}
+
+$SERVER["socket"] == "[::]:{{ cp_zone_item.zoneid|int + 8000 }}" {
+	$HTTP["host"] !~ "(.*{{cp_zone_item.redirect_host_match}}.*)" {
+		$HTTP["host"] =~ "([^:/]+)" {
+			url.redirect = ( "^(.*)$" => "{{cp_zone_item.redirect_host}}?redirurl=%1$1")
+		}
+    }
+    {% if cp_zone_item.certificate|default("") != "" %}
+    ssl.engine = "enable"
+    {% else %}
+    ssl.engine = "disable"
+    {% endif %}
+}
 
 ## redirect http traffic to http(s) main target
 $SERVER["socket"] == ":{{ cp_zone_item.zoneid|int + 9000 }}" {
 	$HTTP["host"] =~ "([^:/]+)" {
-		url.redirect = ( "^(.*)$" => "{{cp_zone_item.redirect_host_v4}}?redirurl=%1$1")
+		url.redirect = ( "^(.*)$" => "{{ cp_zone_item.redirect_host }}?redirurl=%1$1")
 	}
 	ssl.engine = "disable"
 }
 $SERVER["socket"] == "[::]:{{ cp_zone_item.zoneid|int + 9000 }}" {
 	$HTTP["host"] =~ "([^:/]+)" {
-		url.redirect = ( "(.*)" => "{{cp_zone_item.redirect_host_v6}}?redirurl=%1$1")
+		url.redirect = ( "^(.*)$" => "{{ cp_zone_item.redirect_host }}?redirurl=%1$1")
 	}
 	ssl.engine = "disable"
 }
@@ -167,8 +120,6 @@ $SERVER["socket"] == "[::]:{{ cp_zone_item.zoneid|int + 9000 }}" {
 setenv.set-request-header = ("zoneid" => "{{cp_zone_item.zoneid}}")
 proxy.server = ( "/api/captiveportal/access" => (
 		( "host" => "127.0.0.1",
-		  "port" => 8999 ),
-		( "host" => "::1",
 		  "port" => 8999 )
 	)
 )
--- a/src/opnsense/service/templates/OPNsense/Captiveportal/lighttpd-api-dispatcher.conf
+++ b/src/opnsense/service/templates/OPNsense/Captiveportal/lighttpd-api-dispatcher.conf
@@ -38,12 +38,9 @@ url.access-deny             = ( "~", ".inc" )
 
 ######### Options that are good to be but not necessary to be changed #######
 
-## bind the default listener to loopback only, otherwise lighttpd falls back to *:80
+## bind to port (default: 80)
 server.bind               = "127.0.0.1"
 server.port               = 8999
-
-$SERVER["socket"] == "[::1]:8999" {
-}
 
 ## to help the rc.scripts
 server.pid-file            = "/var/run/lighttpd-api-dispatcher.pid"


src/opnsense/mvc/app/controllers/OPNsense/CaptivePortal/Api/ServiceController.php
```diff
@@ -52,6 +52,9 @@ protected function serviceEnabled()
 
     protected function invokeFirewallReload()
     {
+        $backend = new Backend();
+        $backend->configdRun('template reload OPNsense/IPFW');
+        $backend->configdRun("ipfw reload");
```
fichtner: is this side effect a good idea? feels a bit too clever :)

src/opnsense/mvc/app/models/OPNsense/CaptivePortal/CaptivePortal.xml
```diff
@@ -33,6 +33,10 @@
                     <Multiple>Y</Multiple>
                     <Service>CaptivePortal</Service>
                 </authservers>
+                <allowVirtualIps type="BooleanField">
```
fichtner: maybe we can find a better name for this, virtual and alias are already taken terminology-wise

src/opnsense/mvc/app/controllers/OPNsense/CaptivePortal/Api/ServiceController.php
swhite2: I expected comments here ;)

To be honest this was the quickest win here, the IPFW reload is a requirement though, and missing since the re-introduction of it and I didn't spot any other obvious hooks in the right place.

src/opnsense/mvc/app/models/OPNsense/CaptivePortal/CaptivePortal.xml
swhite2: Agreed, suggestions?

src/opnsense/mvc/app/controllers/OPNsense/CaptivePortal/Api/ServiceController.php
swhite2: for context: the spot where this code resided previously was cleaned up in https://github.com/opnsense/core/commit/d8519a06a8a5b4fd3dd7991ef33d833163dcfff6

src/opnsense/mvc/app/controllers/OPNsense/CaptivePortal/forms/dialogZone.xml
```diff
@@ -37,10 +37,10 @@
         </grid_view>
     </field>
     <field>
-        <id>zone.allowVirtualIps</id>
+        <id>zone.roaming</id>
         <type>checkbox</type>
-        <label>Allow multiple client IPs</label>
-        <help>Allow a connecting client to use multiple IPs (bound to the same MAC). For IPv4, these can be virtual IPs on the client. For IPv6, this option is needed for maximum compatibility because a client may actively use multiple IPv6 addresses.</help>
+        <label>Client roaming</label>
+        <help>Allow a connecting client to use multiple IPs (bound to the same MAC) over the course of its session. For IPv4, these can be virtual IPs on the client. For IPv6, this option is needed for maximum compatibility because a client may actively use multiple IPv6 addresses.</help>
```
fichtner: strictly speaking the more common term is "alias" for an extra address (as per ifconfig). Both could still be confusing. Perhaps we can remove the IPv4 explanation or change it to a minor note "This also affects IPv4."

Wouldn't that also allow login from multiple IPs? Didn't we have a setting for that? Concurrent something?

fichtner: (if it's strictly tied to the MAC/DUID that's something else and please ignore me)

swhite2: "roaming" keys clients (with IP aliases) to a MAC address, if they connect from a different device the "concurrent user logins" setting still applies, different MAC.

src/etc/inc/plugins.inc.d/captiveportal.inc
```diff
@@ -133,7 +133,30 @@ function captiveportal_firewall($fw)
...
+                            'pass' => true,
+                            'nordr' => false,
+                            'ipprotocol' => 'inet6',
+                            'protocol' => 'tcp',
+                            'from' => "<__captiveportal_zone_{$zoneid}>",
+                            'from_not' => true,
+                            'to' => "<__captiveportal_zone_{$zoneid}>",
+                            'to_not' => true,
+                            'to_port' => $to_port,
+                            'target' => $intf . 'ip',
```
Monviech, line 149: Is the rdr working in an IPv6 only network?

swhite2: With the rdr now bound to the primary IPv6 address of the specified interface, yes :)

src/opnsense/service/templates/OPNsense/Captiveportal/lighttpd-zone.conf
```diff
@@ -5,8 +5,20 @@
         {# found zone, search for interface ip #}
         {% for intf_tag in item.interfaces.split(',') %}
             {% for conf_key, conf_inf in interfaces.items() %}
-                {% if conf_key == intf_tag and conf_inf.ipaddr and conf_inf.ipaddr != 'dhcp' %}
-                    {% do item.update({'interface_hostaddr':conf_inf.ipaddr}) %}
+                {% if conf_key == intf_tag %}
+                    {# prefer IPv6 if available, fallback to IPv4 #}
+                    {% if conf_inf.ipv6 and conf_inf.ipv6|length > 0 %}
```
AdSchellevis: this looks highly suspicious, there is no `interface.XXX.ipv6` in our configuration.

src/etc/inc/plugins.inc.d/captiveportal.inc
```diff
@@ -91,7 +91,7 @@ function captiveportal_firewall($fw)
             }
 
             foreach ($zone->interfaces->getValues() as $intf) {
-                // allow DNS
+                // allow DNS (IPv4 + IPv6)
```
AdSchellevis: I think we can just kill the whole comment, it doesn't add much to be honest

src/etc/inc/plugins.inc.d/captiveportal.inc
```diff
@@ -133,7 +133,30 @@ function captiveportal_firewall($fw)
...
+                            'target' => $intf . 'ip',
+                            'localport' => $rdr_port,
+                            'natreflection' => 'disable',
+                            'log' => true,
+                            'descr' => "Redirect to Captive Portal (zone {$zoneid})",
+                            '#ref' => "ui/captiveportal#edit={$uuid}"
+                        ]
+                    );
+
+                    // Allow access to the captive portal (IPv4 + IPv6)
```
AdSchellevis: same is the case for this one I guess, the description is already pretty clear.

src/etc/inc/plugins.inc.d/captiveportal.inc
```diff
@@ -152,7 +175,7 @@ function captiveportal_firewall($fw)
                     );
                 }
 
-                // block all non-authenticated users
+                // block all non-authenticated users (IPv4 + IPv6)
```
AdSchellevis: same here

src/opnsense/mvc/app/controllers/OPNsense/CaptivePortal/Api/AccessController.php
```diff
@@ -50,21 +50,41 @@ class AccessController extends ApiControllerBase
...
             foreach ($allClients as $connectedClient) {
-                if ($connectedClient['ipAddress'] == $this->getClientIp()) {
+                if ($connectedClient['ipAddress'] == $clientIp) {
                     // client is authorized in this zone according to our administration
                     $connectedClient['clientState'] = 'AUTHORIZED';
                     return $connectedClient;
                 }
             }
+
+            // IP-based lookup failed - try MAC-based lookup for MAC-authenticated sessions
```
AdSchellevis: This whole block is a bit questionable in my opinion, `list clients` should be responsible for tracking the changes via the sqlite database. When we query specifically on mac here we might end with the situation where the frontend reports "AUTHORIZED", but in reality I'm still reflected to the login page for not being in the list yet.

If we do want to wait for the "accounting department" to be finished with the registration, we can make the frontend query again in the background to see if the "connected" state changed after 30 seconds or so. But this would be more a default template thing....

src/opnsense/mvc/app/controllers/OPNsense/CaptivePortal/Api/AccessController.php
```diff
@@ -87,30 +107,34 @@ protected function clientSession(string $zoneid)
      */
     protected function getClientIp()
     {
```
AdSchellevis, line 90: Strictly speaking I don't think we need to change this method, although the old one returned possible extra (proxy) addresses, I can't remember a ticket describing the issue and `strpos($clientAddress, '127') === 0` would match on more than the `127.x.x.x` as specified before.

src/opnsense/mvc/app/controllers/OPNsense/CaptivePortal/Api/AccessController.php
```diff
@@ -263,7 +287,12 @@ public function logonAction($zoneid = 0)
 
                             // check group when group enforcement is set
                             if ($isAuthenticated && (string)$cpZone->authEnforceGroup != "") {
-                                $isAuthenticated = $authServer->groupAllowed($userName, $cpZone->authEnforceGroup);
+                                /** @var \OPNsense\Auth\Base $authServer */
+                                if ($authServer instanceof \OPNsense\Auth\Base) {
```
AdSchellevis: If we're suddenly afraid the factory returns the wrong type, it might be better to add `groupAllowed()` to the interface(`IAuthConnector`) as no other implementations than `Base` exist at the moment.

```
# grep -r IAuthConnector src/ | grep extends
src/opnsense/mvc/app/library/OPNsense/Auth/LDAP.php:class LDAP extends Base implements IAuthConnector
src/opnsense/mvc/app/library/OPNsense/Auth/API.php:class API extends Base implements IAuthConnector
src/opnsense/mvc/app/library/OPNsense/Auth/Local.php:class Local extends Base implements IAuthConnector
src/opnsense/mvc/app/library/OPNsense/Auth/Radius.php:class Radius extends Base implements IAuthConnector
src/opnsense/mvc/app/library/OPNsense/Auth/Voucher.php:class Voucher extends Base implements IAuthConnector
```

src/opnsense/mvc/app/controllers/OPNsense/CaptivePortal/Api/ServiceController.php
AdSchellevis: Looks duplicate now, as in the following line both `dnctl` and `ipfw` are triggered when no argument is specified:

https://github.com/opnsense/core/blob/36b17adf3155181ba4b09e39ba2b15635a2f7d31/src/etc/inc/filter.inc#L179

src/opnsense/scripts/captiveportal/lib/db.py
```diff
@@ -100,71 +100,126 @@ def create(self, force_recreate=False):
...
+                    , ip_address VARCHAR NOT NULL
+                    , PRIMARY KEY (zoneid, sessionid, ip_address)
+                    , FOREIGN KEY (zoneid, sessionid)
+                        REFERENCES cp_clients(zoneid, sessionid)
+                        ON DELETE CASCADE
+                )
+            """)
+
+            # indexes used when allow_roaming is turned on for a zone
+            cur.execute("CREATE INDEX IF NOT EXISTS cp_client_ips_ip   ON cp_client_ips(ip_address)")
```
AdSchellevis: can move this outside the `if` block to prevent duplication

src/opnsense/scripts/captiveportal/lib/db.py
```diff
@@ -100,71 +100,126 @@ def create(self, force_recreate=False):
...
+            cur.execute("CREATE INDEX IF NOT EXISTS cp_client_ips_zone ON cp_client_ips(zoneid)")
+
+            self._connection.commit()
```
AdSchellevis: in which case the commit is also always there...

src/opnsense/scripts/captiveportal/lib/db.py
```diff
@@ -100,71 +100,126 @@ def create(self, force_recreate=False):
...
+        else:
+            # Table exists: ensure indexes are present
+            cur.execute("CREATE INDEX IF NOT EXISTS cp_client_ips_ip   ON cp_client_ips(ip_address)")
+            cur.execute("CREATE INDEX IF NOT EXISTS cp_client_ips_zone ON cp_client_ips(zoneid)")
+            self._connection.commit()
+
         cur.close()
 
     def sessions_per_address(self, zoneid, ip_address=None, mac_address=None):
```
AdSchellevis:
```suggestion
    def sessions_per_address(self, zoneid, ip_address='', mac_address=''):
```

src/opnsense/scripts/captiveportal/lib/db.py
```diff
@@ -100,71 +100,126 @@ def create(self, force_recreate=False):
...
-        """ fetch session(s) per (mac) address
-        :param zoneid: cp zone number
-        :param ip_address: ip address
-        :return: active status (boolean)
+        """ fetch session(s) per (ip/mac) address
+        Primary IP is stored in cp_clients.ip_address; roaming IPs in cp_client_ips.
         """
+        # Nothing to match on
+        if (ip_address is None or str(ip_address).strip() == '') and (mac_address is None or str(mac_address).strip() == ''):
```
AdSchellevis:
```suggestion
        if ip_address.strip() == '' and mac_address.strip() == '':
```

src/opnsense/scripts/captiveportal/lib/db.py
```diff
@@ -100,71 +100,126 @@ def create(self, force_recreate=False):
...
-                                or
-                                cc.mac_address = :mac_address
-                             )""", request)
 
-        result = []
-        for row in cur.fetchall():
-            result.append({'sessionId': row[0], 'authenticated_via': row[1]})
-        return result
+        clauses = []
+        if ip_address is not None and str(ip_address).strip() != '':
```
AdSchellevis:
```suggestion
        if ip_address.strip() != '':
```

src/opnsense/scripts/captiveportal/lib/db.py
```diff
@@ -100,71 +100,126 @@ def create(self, force_recreate=False):
...
+            # Match either primary IP or any roaming IP
+            clauses.append("(cc.ip_address = :ip_address OR ci.ip_address = :ip_address)")
+        if mac_address is not None and str(mac_address).strip() != '':
```
AdSchellevis:
```suggestion
        if mac_address.strip() != '':
```

src/opnsense/scripts/captiveportal/lib/db.py
```diff
@@ -100,71 +100,126 @@ def create(self, force_recreate=False):
...
-        # add new session
-        cur.execute("""INSERT INTO cp_clients(zoneid, authenticated_via, sessionid, username,  ip_address, mac_address, created)
-                       VALUES (:zoneid, :authenticated_via, :sessionId, :userName, :ipAddress, :macAddress, :startTime)
-                    """, response)
+        try:
+            cur.execute("BEGIN")
+
+            # set cp_client as deleted in case there's already a user logged-in at this ip address
+            # (match both primary IP and roaming IPs)
+            if ip_address is not None and str(ip_address).strip() != '':
```
AdSchellevis:
```suggestion
            if ip_address.strip() != '':
```

src/opnsense/scripts/captiveportal/lib/db.py
```diff
@@ -180,6 +235,80 @@ def update_client_ip(self, zoneid, sessionid, ip_address):
                     """, {'zoneid': zoneid, 'sessionid': sessionid, 'ip_address': ip_address})
         self._connection.commit()
 
+    def update_roaming_ips(self, zoneid, sessionid, ip_addresses):
+        """Update roaming IP addresses for a session to exactly match the provided list.
+        Returns the current set of IP addresses stored in the database.
+        """
+
+        if isinstance(sessionid, bytes):
```
AdSchellevis: is this possible? looking at the code there's always a session string (bytes) unless its None and the following would crash

src/opnsense/scripts/captiveportal/lib/db.py
```diff
@@ -180,6 +235,80 @@ def update_client_ip(self, zoneid, sessionid, ip_address):
                     """, {'zoneid': zoneid, 'sessionid': sessionid, 'ip_address': ip_address})
         self._connection.commit()
 
+    def update_roaming_ips(self, zoneid, sessionid, ip_addresses):
```
AdSchellevis:
```suggestion
    def update_roaming_ips(self, zoneid, sessionid, ip_addresses=[]):
```

src/opnsense/scripts/captiveportal/lib/db.py
```diff
@@ -180,6 +235,80 @@ def update_client_ip(self, zoneid, sessionid, ip_address):
...
+        """Update roaming IP addresses for a session to exactly match the provided list.
+        Returns the current set of IP addresses stored in the database.
+        """
+
+        if isinstance(sessionid, bytes):
+            sessionid = sessionid.decode()
+
+        if not ip_addresses:
```
AdSchellevis: can ditch this when specifying the default.

src/opnsense/scripts/captiveportal/lib/db.py
```diff
@@ -180,6 +235,80 @@ def update_client_ip(self, zoneid, sessionid, ip_address):
...
+            ip_addresses = []
+
+        # clean + deduplicate + remove empty values
+        new_ips = {
+            str(ip).strip()
```
AdSchellevis: `get_all_addresses_by_mac()` seems to be responsible to feed this, I would expect we enforce proper typing there if that isn't already the case.

src/opnsense/scripts/captiveportal/lib/db.py
```diff
@@ -180,6 +235,80 @@ def update_client_ip(self, zoneid, sessionid, ip_address):
...
+            for ip in ip_addresses
+            if ip is not None and str(ip).strip() != ''
```
AdSchellevis: same here

src/opnsense/scripts/captiveportal/lib/db.py
```diff
@@ -180,6 +235,80 @@ def update_client_ip(self, zoneid, sessionid, ip_address):
...
+                AND sessionid = :sessionid
+            """, {
+                'zoneid': zoneid,
+                'sessionid': sessionid
+            })
+
+            current_ips = {row[0] for row in cur.fetchall()}
+
+            # if identical (order-independent), do nothing
+            if current_ips == new_ips:
```
AdSchellevis, line 229: nice touch using a set here

src/opnsense/scripts/captiveportal/lib/db.py
```diff
@@ -243,24 +373,66 @@ def list_clients(self, zoneid=None):
...
 
+    
+    def list_session_ips(self, zoneid, sessionid):
+        """
+        Return primary + roaming IPs for a session.
+        - Primary IP is cp_clients.ip_address
+        - roaming IPs are cp_client_ips.ip_address
+        Returns a de-duplicated set[str] (order not guaranteed).
+        """
+        if isinstance(sessionid, bytes):
```
AdSchellevis: same as in `update_roaming_ips()`, how?

src/opnsense/scripts/captiveportal/lib/db.py
```diff
@@ -292,84 +464,173 @@ def find_concurrent_user_sessions(self, zoneid):
...
+        """)
+        for zoneid, sessionid, ip in cur.fetchall():
+            key = (zoneid, sessionid)
+            if key in sessions:
+                sessions[key]['ips'].add(ip)
+
+        # Ensure primary IP is included for each session
+        for rec in sessions.values():
+            pip = rec.get('primary_ip')
+            if pip is not None and str(pip).strip() != '':
```
AdSchellevis: going further through the code, I notice a lot of "consumers" `strip()` their results, it feels a bit like we're either doing something wrong with storing the data or these are not needed at all.

src/opnsense/scripts/captiveportal/sql/init.sql
```diff
@@ -18,6 +18,20 @@ create table cp_clients (
 create index cp_clients_ip ON cp_clients (ip_address);
 create index cp_clients_zone ON cp_clients (zoneid);
 
+-- multiple IPs per session
```
AdSchellevis, line 21: we might consider always calling `init.sql` so we can change the `create table`  statements to `create table if not exists` and remove the create when not there yet code earlier.

src/opnsense/scripts/captiveportal/listClients.py
```diff
@@ -37,5 +37,5 @@
 parser.add_argument('-z', help='optional zoneid to filter on', type=str)
 args = parser.parse_args()
 
-response = DB().list_clients(int(args.z) if str(args.z).isdigit() else None)
+response = DB().list_clients(int(args.z) if str(args.z).isdigit() else None, True)
```
AdSchellevis, line 40: `DB().list_clients(...,True)` doesn't seem to have an actual meaning, so either the default `None` is an issue or adding the bool here was not needed.

agoodkind:
Thanks for taking this over!

src/opnsense/mvc/app/controllers/OPNsense/CaptivePortal/Api/AccessController.php
swhite2: Since we now expose `ipAddresses` in all cases ("roaming" or not), I've adjusted the code to look at all known IP addresses here - that should be enough. For IPv6 there is still a delay after logging in where the status action doesn't yet reflect reality but if we want to fix that better to do in the the template as you mentioned.

src/opnsense/scripts/captiveportal/listClients.py
swhite2: I stripped code a little too vigorously in https://github.com/opnsense/core/pull/9745/changes/8297c7d54ad135a45c6f95629316d747c618c018. The `AccessController` (also the only consumer) needs this now to have a full picture of each client.

swhite2:
@agoodkind If you have the setup... care to give this a spin?

src/opnsense/scripts/captiveportal/sql/init.sql
swhite2: @AdSchellevis https://github.com/opnsense/core/pull/9745/changes/17587f64860846a83114ae7b33f099b770b6761f should take care of this. The background process is now responsible for creating the db, all other `DB()` callers now skip migration code, which is slightly more efficient anyway.

src/opnsense/mvc/app/controllers/OPNsense/CaptivePortal/Api/AccessController.php
swhite2: @AdSchellevis I've gone for https://github.com/opnsense/core/pull/9745/changes/bedaadc6d2ef92525693e07c8c66a77e6191cc94 instead, which updates the db as soon as `allow.py` is triggered preventing the need for hardcoded delays on the front-end.

src/opnsense/mvc/app/controllers/OPNsense/CaptivePortal/Api/ServiceController.php
swhite2: @AdSchellevis Coming back to this as it's related to https://github.com/opnsense/core/issues/9902.. the `start.sh` script does not trigger the template generation for https://github.com/opnsense/core/blob/master/src/opnsense/service/templates/OPNsense/IPFW/rc.conf.d, which means that configuring a zone in a clean install will likely not kickstart `ipfw` until a reboot.

In this case I'd opt for

```
$backend->configdRun('template reload OPNsense/IPFW');
```

To remain in here.

src/opnsense/scripts/captiveportal/sql/init.sql
AdSchellevis: @swhite2 I like the try/except here and moving the creation to the backend process in full, although the previous construct might be slightly more forgiving if something went wrong with the main process, I do believe this is much better and indeed more efficient.

src/opnsense/scripts/captiveportal/allow.py
```diff
@@ -41,16 +41,28 @@
...
     authenticated_via=args.authenticated_via,
     username=args.username,
     ip_address=args.ip_address,
     mac_address=arp_entry['mac'] if arp_entry is not None else None
 )
-PF.add_to_table(zoneid=args.zoneid, address=args.ip_address)
+
+session_ips = {args.ip_address}
+if args.roaming:
+    session_ips = db.update_roaming_ips(args.zoneid, response.sessionId, arp.get_all_addresses_by_mac(arp_entry['mac']))
```
AdSchellevis: as discussed we likely better only add the connected address and let the background process take care of the rest.

src/opnsense/service/conf/actions.d/actions_captiveportal.conf
```diff
@@ -6,7 +6,7 @@ message:list registered clients
 
 [allow]
 command:/usr/local/opnsense/scripts/captiveportal/allow.py
-parameters:--zoneid=%s --username=%s --ip_address=%s --authenticated_via=%s
+parameters:--zoneid=%s --username=%s --ip_address=%s --authenticated_via=%s --roaming=%s
```
AdSchellevis: when roaming isn't relevant, we can ditch it from here as well.

src/opnsense/service/templates/OPNsense/Captiveportal/lighttpd-zone.conf
```diff
@@ -5,8 +5,15 @@
         {# found zone, search for interface ip #}
         {% for intf_tag in item.interfaces.split(',') %}
             {% for conf_key, conf_inf in interfaces.items() %}
-                {% if conf_key == intf_tag and conf_inf.ipaddr and conf_inf.ipaddr != 'dhcp' %}
-                    {% do item.update({'interface_hostaddr':conf_inf.ipaddr}) %}
+                {% if conf_key == intf_tag %}
+                    {# prefer IPv6 if available, fallback to IPv4 #}
```
AdSchellevis: what happens with dual stack and an ipv4 only client here?

agoodkind:
> @agoodkind If you have the setup... care to give this a spin?

I had some free time and did a matrix of testing and made a PR that tries to address all of @AdSchellevis comments along with bugs I discovered: https://github.com/opnsense/core/pull/9927

src/opnsense/mvc/app/controllers/OPNsense/CaptivePortal/Api/AccessController.php
swhite2: Ditched for reasons as described in https://github.com/opnsense/core/pull/9745#discussion_r2899365752

agoodkind: During testing I confirmed a narrow version of the concern @AdSchellevis raised in https://github.com/opnsense/core/pull/9745#discussion_r2836094805. The very first post-login request from a DHCPv6 `/128` source address would still receive one `302` redirect before traffic flowed, even though the login API had already returned `AUTHORIZED` and the address was already in the PF allow table.

The cause is a PF state ordering issue: `allow.py` adds the address to the table and calls `pfctl -k` to flush existing states, but the DHCPv6 `/128` address seems to have a slightly different state lifecycle than SLAAC or privacy addresses. SLAAC and privacy address traffic worked on the first request after login without the same race.

The `pf.py` change in this PR (flushing states on `add_to_table`) does help in the common cases. The `/128` edge is a narrower residual.

Separately, @swhite2 was also working in the same direction with https://github.com/opnsense/core/commit/bedaadc6d2ef92525693e07c8c66a77e6191cc94, which updated the DB immediately when `allow.py` is triggered to avoid hardcoded delays on the frontend. That approach and the eager expansion were both subsequently backed out. If revisiting the timing issue is ever on the table, that commit is probably a useful reference point for what was tried.

agoodkind:
Tested with my setup against https://github.com/opnsense/core/pull/9745/commits/9df582aba41c14ddf3e3e64f06ac058691643c9b

### Setup

`opnsense-dev`
- Captive Portal test zone on `lan`
- Portal hosts:
  - IPv4: `10.240.240.1:8000`
  - IPv6: `3d06:bad:b01:240::1:8000`
- Standard outbound NAT66 for `3d06:bad:b01:240::/64`
- `DNS64` enabled in Unbound
- `dns64prefix`: `3d06:bad:b01:2464::/96`
- `radvd` advertises `nat64prefix 3d06:bad:b01:2464::/96`
- DHCPv4 option `108` present on `lan`

`ubuntu-vm-test`
- Single NIC behind `opnsense-dev`
- `systemd-networkd`
- Address types exercised:
  - IPv4 DHCP
  - DHCPv6 IA_NA `/128`
  - SLAAC / privacy `/64`
  - extra temporary `/64`
- Two guest modes:
  - dual-stack
  - IPv6-only preferred (`IPv6OnlyMode=yes`)

### Current behavior

The login address works immediately.

Sibling addresses do not always work immediately on current upstream. Addresses that are already visible to the firewall may still need the background reconciliation pass before they are consistently usable.

### Matrix

| Case | Result | Notes |
|---|---|---|
| IPv4-only pre-auth | Pass | Redirects to IPv4 portal host |
| IPv4-only auth via IPv4 | Pass | IPv4 egress works immediately |
| Dual-stack pre-auth via IPv4 | Pass | Redirects to IPv4 portal host |
| Dual-stack pre-auth via IPv6 | Pass | Redirects to IPv6 portal host |
| Dual-stack auth via IPv4, same IPv4 address | Pass | Login address works immediately |
| Dual-stack auth via IPv4, DHCPv6 `/128` sibling, cold cache | Delayed | Not immediately* |
| Dual-stack auth via IPv4, already-known IPv6 siblings, warm cache | Mixed | Better once visible, not consistently immediate |
| Dual-stack auth via IPv6, same DHCPv6 `/128` address | Partial | First request can still see one `302`, retry works |
| Dual-stack auth via IPv6, IPv4 sibling, cold cache | Delayed | Not immediately* |
| Dual-stack auth via IPv6, already-known IPv4 and IPv6 siblings, warm cache | Mixed | Better once visible, not consistently immediate |
| IPv6-only preferred pre-auth | Pass | Redirects to IPv6 portal host |
| IPv6-only preferred auth via DHCPv6 `/128` | Partial | Same narrow first-request race as above |
| IPv6-only preferred auth via privacy / SLAAC `/64` | Pass | Immediate post-auth IPv6 egress worked |
| NAT64 / DNS64 / PREF64 after auth | Pass | Synthesized `3d06:bad:b01:2464::101:101` returned a real HTTP response |
| Multi-address IPv6 convergence | Pass | Newly observed sibling addresses join later through background reconciliation |

### Main takeaway

Current upstream is usable for the core portal flow, but it still has a convergence model rather than an immediate warm-start model.

The address used for login works immediately. Other addresses on the same client join later as they become visible to the firewall and the background reconciliation loop processes them.

The narrowest remaining rough edge is the DHCPv6 `/128` source case, where the first immediate post-login request can still receive one redirect before the next request succeeds.

*Sibling addresses usually became usable within about 5 to 12 seconds, which lined up with the 5 second background reconciliation loop plus the time needed for the address to become visible in neighbor discovery.

src/opnsense/mvc/app/controllers/OPNsense/CaptivePortal/forms/dialogZone.xml
agoodkind, line 121: Maybe could change to:

Hostname to redirect the login page to. Leave blank to use the interface IP address. When using an SSL certificate, the hostname and certificate name must match. **Note:** For IPv6-only client compatibility, either set a hostname with both A and AAAA records, or ensure the selected interface has a static IPv6 address configured. If only a dynamic DHCPv6 address is present and no hostname is set, IPv6-only clients may be redirected to an IPv4 address they cannot reach.

```suggestion
        <help><![CDATA[Hostname to redirect the login page to. Leave blank to use the interface IP address. When using an SSL certificate, the hostname and certificate name must match. Note: For IPv6-only client compatibility, either set a hostname with both A and AAAA records, or ensure the selected interface has a static IPv6 address configured. If only a dynamic DHCPv6 address is present and no hostname is set, IPv6-only clients may be redirected to an IPv4 address they cannot reach.]]></help>
        
```

Monviech: I advice to avoid HTML inside language strings as this will break during translation.

swhite2:
@agoodkind Thanks for your further testing. Your results seem to align with what is to be expected given the constraints. After some further discussion internally, the best strategy to address the rather complicated lighty template seems to be to stop guessing for IPv6 addresses (which are going to be static less often in comparison to IPv4 anyway) and require that a user sets a proper hostname + associated AAAA records (and possibly DNS64 support in Unbound) to make IPv6 work.

src/opnsense/mvc/app/controllers/OPNsense/CaptivePortal/forms/dialogZone.xml
agoodkind: Fair. I took out the HTML formatting in the string suggestion

agoodkind:
> @agoodkind Thanks for your further testing. Your results seem to align with what is to be expected given the constraints. After some further discussion internally, the best strategy to address the rather complicated lighty template seems to be to stop guessing for IPv6 addresses (which are going to be static less often in comparison to IPv4 anyway) and require that a user sets a proper hostname + associated AAAA records (and possibly DNS64 support in Unbound) to make IPv6 work.

Thanks for the follow-up. I understand the appeal of simplifying the template, but requiring a hostname + AAAA for IPv6 captive portal to function at all would be quite a gap IMO. Although it is true that most network admins _should_ set up some type of hostname to handle this, I still feel it is a gap. 

`track6` is the documented way to deploy DHCPv6-PD on LAN, DHCPv6-PD is required by [RFC 7084](https://www.rfc-editor.org/rfc/rfc7084.html) for CE routers, and [RIPE-690](https://www.ripe.net/publications/docs/ripe-690/) documents it as the standard operator practice for prefix assignment to end-users.

For example, I've encountered two different ISPs where you are required to use DHCP-PD otherwise the traffic won't route (AT&T Fiber Business, and Webpass Business). A captive portal that silently falls back to an unreachable IPv4 address in that configuration is a real bug, not a corner case.

The fix itself is fairly contained: runtime address resolution for `track6` interfaces The configd template system was designed around `config.xml`, which holds static values. With `track6`, the LAN address is runtime-derived from DHCPv6-PD and isn't in `config.xml` at all. The web GUI (https://github.com/opnsense/core/blob/master/src/etc/inc/plugins.inc.d/webgui.inc#L36), Unbound (https://github.com/opnsense/core/blob/master/src/etc/inc/plugins.inc.d/unbound.inc#L45), and OpenSSH (https://github.com/opnsense/core/blob/master/src/etc/inc/plugins.inc.d/openssh.inc#L44) solved this same problem via `newwanip` hooks and PHP-side config regeneration (https://github.com/opnsense/core/issues/5966). The captive portal lighttpd config uses the Jinja2 template system instead, so the address has to enter the template somehow. I opened a PR with both: https://github.com/opnsense/core/pull/9973

src/opnsense/mvc/app/controllers/OPNsense/CaptivePortal/Api/AccessController.php
```diff
@@ -87,30 +88,40 @@ protected function clientSession(string $zoneid)
      */
     protected function getClientIp()
     {
+        $clientAddress = $this->request->getClientAddress();
+        $forwardedFor = $this->request->getHeader('X-Forwarded-For');
+        $realIp = $this->request->getHeader('X-Real-Ip');
```
AdSchellevis: Can we please just go back to the original situation, I have a very strong feeling we're trying to fix an edge case while introducing new security concerns in the process. It's debatable if by default any of these headers should be processed, but since that's already the case (introduced a very long time ago https://github.com/opnsense/core/commit/a7033f22a1d7a8f2577189a0f59c0ae07ae3c1ba), let's try not to increase the impact so we can discuss improvements on this topic in a separate ticket.

src/opnsense/mvc/app/controllers/OPNsense/CaptivePortal/Api/AccessController.php
```diff
@@ -294,7 +305,7 @@ public function logonAction($zoneid = 0)
                                 (string)$cpZone->zoneid,
                                 $userName,
                                 $clientIp,
-                                $authServerName
+                                $authServerName,
```
AdSchellevis:
```suggestion
                                $authServerName
```

src/opnsense/scripts/captiveportal/lib/pf.py
```diff
@@ -44,10 +51,16 @@ def list_table(zoneid):
     @staticmethod
     def add_to_table(zoneid, address):
         subprocess.run(['/sbin/pfctl', '-t', f'__captiveportal_zone_{zoneid}', '-T', 'add', address], capture_output=True)
+        # Clear pre-auth states so newly authorized sibling addresses can use the updated table match immediately.
+        subprocess.run(['/sbin/pfctl', '-k', f'{address}'], capture_output=True)
+        wildcard = '::/0' if PF._is_ipv6(address) else '0.0.0.0/0'
+        subprocess.run(['/sbin/pfctl', '-k', wildcard, '-k', f'{address}'], capture_output=True)
```
AdSchellevis: this looks rather suspicious and may result in unexpected locked sessions to already accepted destinations.

Issue #9542, opened 2025-12-18
Unbound Aliases - Not Updating


daygle:
Hi,

I have noticed that with version OPNsense 25.7.9_7-amd64 Aliases under Unbound DNS > Overrides is no longer working.

When attempting to create an alias nothing happens. Record is not saved.

fichtner:
This may be related to https://github.com/opnsense/core/issues/9540 but it's mostly how the page works. It works on my end.

mr-manuel:
I tested with OPNsense `25.7.8`, `25.7.9_7` and `26.1.b_2` with and without the changes of https://github.com/opnsense/core/issues/9540 and it works without problems.

fichtner:
Well, this is about not selecting the one that you're adding so it looks like nothing was added. With #9540 the table should update after adding so you can see the count increase but OTOH if the override is not currently visible there's little to be done. The same goes for a full alias list underneath.


Cheers,
Franco

daygle:
@fichtner is any more detail needed? I'm a little worried as mention you cannot replicate?

fichtner:
Sure have you selected the corresponding override first?

daygle:
Something isn't correct, have a look at the attached. I am now on version OPNsense 25.7.10-amd64.

host_entries.conf is containing entries that do not exist.

[resolver.log](https://github.com/user-attachments/files/24247684/resolver.log)
<img width="1577" height="870" alt="Image" src="https://github.com/user-attachments/assets/8884101d-287f-4afc-9949-ba3a111a9750" />

[host_entries.conf.txt](https://github.com/user-attachments/files/24247704/host_entries.conf.txt)

daygle:
For some reason I see old/unrelated/deleted entries in the host_entries file. These cannot be seen in the GUI

local-data: "test.daygle.net  IN AAAA 2403:5802:4645:2::100"
local-data: 'test.daygle.net TXT "test"'
local-data: "test.daygle.net  IN AAAA 2403:5802:4645:2::100"
local-data: 'test.daygle.net TXT "test"'
local-data: "test.daygle.net  IN AAAA 2403:5802:4645:2::100"
local-data: 'test.daygle.net TXT "test"'

daygle:
Oh, hang on I think these entries were causing issues:

<img width="1432" height="101" alt="Image" src="https://github.com/user-attachments/assets/8d1d3e00-703a-48e8-a953-44e375eb6d11" />

daygle:
[...]

PR #9753, opened 2026-02-07
Fix PTR Record Generation for Unbound

daygle:
1. Model Update (Unbound.xml)

    Added createptr BooleanField to the aliases section with default value 0 (unchecked)
    Updated model version from 1.0.13 to 1.0.14

2. Form Update (dialogHostAlias.xml)

    Added a new checkbox field labeled "Create PTR Record"
    Positioned after the "Enabled" checkbox
    Includes helpful text explaining the functionality

3. Logic Update (unbound.inc)

    Modified alias processing to include the createptr field value
    Updated PTR generation logic with strict comparison (===)
    Added clear comments explaining the conditional behavior
    Preserved existing behavior for main host overrides

### DIFF
--- a/src/opnsense/mvc/app/models/OPNsense/Unbound/Unbound.xml
+++ b/src/opnsense/mvc/app/models/OPNsense/Unbound/Unbound.xml
@@ -1,7 +1,7 @@
 <model>
     <mount>//OPNsense/unboundplus</mount>
     <description>Unbound configuration</description>
-    <version>1.0.13</version>
+    <version>1.0.14</version>
     <items>
         <general>
             <enabled type="BooleanField">
@@ -386,6 +386,10 @@
                     <Default>1</Default>
                     <Required>Y</Required>
                 </enabled>
+                <createptr type="BooleanField">
+                    <Default>0</Default>
+                    <Required>Y</Required>
+                </createptr>
                 <host type="ModelRelationField">
                     <Model>
                         <host>


daygle:
@AdSchellevis I have created an improved pull request, would love your thoughts about this one.

https://github.com/opnsense/core/issues/9542

AdSchellevis:
@daygle well, I think we can simplify this quite a bit if we just mark the new option as default selected and remove most of the logic. I don't mind giving a stab at it, maybe later this weekend. Having the outcome more visible is something I like so a user knows what to expect from the ui.

daygle:
@AdSchellevis I'll leave it with you then. Hope to see a fix soon.

AdSchellevis:
@daygle this https://github.com/opnsense/core/commit/766c6ce8e70858619651af332522fbe84162f65e should basically do the trick, it exposes the PTR option to the override where you can disable it if you don't want it. The alias in this case isn't the proper spot as the alias doesn't come with its own address (logically the primary address may have the PTR). Functionally I believe this change shouldn't impact existing installations as the first one was always the one to attach the PTR to (but if there are more pointing to the same address, one can choose now).

daygle:
Thanks @AdSchellevis, look forward to seeing the change in the next release. Thanks for looking in to this one.

daygle:
@AdSchellevis Am I correct in assuming that the fix is not in 26.1.2?

fichtner:
Strictly speaking all feature additions should get a round on the development version so .2 doesn't have the change indeed.

daygle:
@fichtner Fair enough. I'll continue to wait.

PR #9755, opened 2026-02-07
Revert "Services: ISC DHCPv4: hide menu items when dnsmasq is enabled…

fichtner:
… to improve "out of the box" experience. closes https://github.com/opnsense/core/issues/8329"

We are not hiding v6 now and since ISC-DHCP is not in the core system we can go back to the normal way.

This reverts commit 0d6aa56527c60be14e6e626e5d0728108110cf2f.

PR: https://www.reddit.com/r/opnsense/comments/1qxqru9/comment/o42nx7v/

### DIFF
--- a/src/opnsense/mvc/app/models/OPNsense/Base/Menu/MenuSystem.php
+++ b/src/opnsense/mvc/app/models/OPNsense/Base/Menu/MenuSystem.php
@@ -172,27 +172,6 @@ class MenuSystem
     }
 
     /**
-     * temporary legacy glue to remove isc dhcp4 settings when not enabled and Dnsmasq is configured as dhcp
-     * @return boolean
-     */
-
-    private function isc_v4_enabled($if)
-    {
-        $config = Config::getInstance()->object();
-        if (isset($config->dhcpd) && isset($config->dhcpd->$if) && !empty((string)$config->dhcpd->$if->enable)) {
-            /* still configured on interface */
-            return true;
-        } elseif (isset($config->dnsmasq) && empty((string)$config->dnsmasq->enable)) {
-            /* dnsmasq not configured at all */
-            return true;
-        } elseif (isset($config->dnsmasq) && !empty((string)$config->dnsmasq->interface)) {
-            /* dnsmasq configured, but only on selected interfaces */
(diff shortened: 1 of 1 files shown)


AdSchellevis:
Maybe we better hook this on the existence of the configuration data, a bit similar to what we recently discussed for the firewall rules (hide when filter.rules empty and mvc has items) . It would prevent this from happening and hides the sections when irrelevant.

fichtner:
After thinking a bit... ISC-DCHP(v6) still has a valid use case. I wouldn't want to fuzz too much about it being in the menu now that it's a plugin since it can be removed. The risk of someone asking why it's not there or cannot be used is not outweighing the benefit of having people avoid misconfiguration.

fichtner:
For context 99c810931a0f8 hides the plugin menu entries in the core-bound files when plugin is not installed and it's no longer in the default installation with 26.1 so doesn't even show using that code alone.

fichtner:
Also relevant to https://forum.opnsense.org/index.php?topic=51079.0

fichtner:
> Maybe we better hook this on the existence of the configuration data, a bit similar to what we recently discussed for the firewall rules (hide when filter.rules empty and mvc has items) . It would prevent this from happening and hides the sections when irrelevant.

In principle I'm ok with this at the price of people not being able to use the plugin in new installs at all. Yet I don't see a pressing reason to actually lock it down like this and fear more of these forum posts in the future.


Cheers,
Franco

AdSchellevis:
ok, let's lock it to "plugin installed" for now and evaluate later then.

fichtner:
These are already being handled further down here

https://github.com/opnsense/core/blob/3c28be916d40db28327cb5a68b8e4f329b6834bd/src/opnsense/mvc/app/models/OPNsense/Base/Menu/MenuSystem.php#L386

and here

https://github.com/opnsense/core/blob/3c28be916d40db28327cb5a68b8e4f329b6834bd/src/opnsense/mvc/app/models/OPNsense/Base/Menu/MenuSystem.php#L409

so merge this PR as is and done for now?

AdSchellevis:
let's do that, yes

PR #9765, opened 2026-02-09
Firewall: Rules [new] - hide legacy option after migration 

AdSchellevis:
(when there's at least a new rule and no legacy rules exist).

This cleans up the menu and prevents existing users from loosing access to their rules.

### DIFF
--- a/src/opnsense/mvc/app/models/OPNsense/Base/Menu/MenuSystem.php
+++ b/src/opnsense/mvc/app/models/OPNsense/Base/Menu/MenuSystem.php
@@ -329,14 +329,28 @@ class MenuSystem
         }
 
         // add interfaces to "Firewall: Rules" menu tab...
-        $this->appendItem('Firewall.Rules', 'Migration', [
-                'url' => '/ui/firewall/migration',
-                'fixedname' => sprintf("<i class='fa fa-fw fa-gears'> </i> %s", gettext('Migration assistant')),
-                'order' => 0,
-        ]);
-        $iftargets['fw'] = array_merge(['FloatingRules' => gettext('Floating')], $iftargets['fw']);
+        $has_legacy_fw = !empty($config->filter?->rule?->count());
+        $has_mvc_fw = !empty($config->OPNsense?->Firewall?->Filter?->rules?->count());
+        if ($has_legacy_fw) {
+            $this->appendItem('Firewall.Rules', 'Migration', [
+                    'url' => '/ui/firewall/migration',
+                    'fixedname' => sprintf("<i class='fa fa-fw fa-gears'> </i> %s", gettext('Migration assistant')),
+                    'order' => 0,
+            ]);
+            $iftargets['fw'] = array_merge(['FloatingRules' => gettext('Floating')], $iftargets['fw']);
+        }
         $ordid = 1;
         foreach ($iftargets['fw'] as $key => $descr) {
+            if ($has_mvc_fw && !$has_legacy_fw) {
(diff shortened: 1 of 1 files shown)


fichtner:
I think this is ok to introduce for that specific purpose although in practise it avoids people going back to using it. In the DHCPv4 case this has already caused issues (but also because implemented differently).

Still I'm asking myself if we park the legacy rules in a plugin it would be better to not revert to other magic other than testing if the plugin is installed or not. And the prerequisite question of whether it should be a plugin or not -- I don't see a reason why not).

src/opnsense/mvc/app/models/OPNsense/Base/Menu/MenuSystem.php
```diff
@@ -329,37 +329,51 @@ public function __construct()
...
-                'order' => $ordid++,
-            ]);
-            $this->appendItem('Firewall.Rules.' . $key, 'Select' . $key, [
-                'url' => '/firewall_rules.php?if=' . $key . '&*',
-                'visibility' => 'hidden',
-            ]);
-            if ($key == 'FloatingRules') {
-                $this->appendItem('Firewall.Rules.' . $key, 'Top' . $key, [
-                    'url' => '/firewall_rules.php',
+            if ($has_mvc_fw && !$has_legacy_fw) {
```
fichtner, line 344: can we have this block end with `continue` and change the indent of the others? It would clean up the diff a lot

fichtner (approved):
after some back and forth with @AdSchellevis: since it's in the menu all the time it's ok for me

Monviech:
I tested it and logged the variables has_legacy_fw and has_mvc_fw and they reflected the reality of my production firewall. So when I removed all legacy rules, the menu disappeared. I think this is good.

Monviech (approved):
I tested it and it works.

AdSchellevis:
@fichtner @Monviech I think it's ready to merge now, objections?

Monviech:
@AdSchellevis I like this, its exactly what some users asked for.

src/opnsense/mvc/app/models/OPNsense/Base/Menu/MenuSystem.php
```diff
@@ -329,14 +329,28 @@ public function __construct()
...
+                    'order' => 0,
+            ]);
+            $iftargets['fw'] = array_merge(['FloatingRules' => gettext('Floating')], $iftargets['fw']);
+        }
         $ordid = 1;
         foreach ($iftargets['fw'] as $key => $descr) {
+            if ($has_mvc_fw && !$has_legacy_fw) {
+                /* only search */
+                $this->appendItem('Firewall.Rule', $key, [
+                    'url' => '/ui/firewall/filter/#interface=' . $key,
```
fichtner, line 347: ok so this is the link in the new, not the old. in that case you need to split $has_mvc_fw and $has_legacy_fw conditions (first for the entries second for the continue)

AdSchellevis: well, I think we should only "unlock" the reference when there's only one rule set to point to. as soon as we start removing the "new" thingy things get a bit obfuscated.

fichtner: Hmm, I don't mind trying this but I'm feeling it will not be the last adjustment to be made here either.

fichtner (approved):

PR #9768, opened 2026-02-09
bootgrid: batch enable/disable-selected toggle by default, clean up IDS page

swhite2:
The footer commands weren't taking into account the `required` properties, rendering unlinked buttons where they shouldn't be (e.g. in dnsmasq DHCP ranges etc.).

Take this opportunity to clean up and improve some front-end code, including batching of enable/disable-selected API calls. The arbitrary default limit of "40" has been set based on (40 * 36-character UUID + ~150-character URI out of a 2k URI character limit).

Reference: https://forum.opnsense.org/index.php?topic=50806.0

### DIFF
--- a/src/opnsense/mvc/app/controllers/OPNsense/Base/ApiMutableModelControllerBase.php
+++ b/src/opnsense/mvc/app/controllers/OPNsense/Base/ApiMutableModelControllerBase.php
@@ -606,35 +606,40 @@ abstract class ApiMutableModelControllerBase extends ApiControllerBase
         $result = ['result' => 'failed'];
         if ($this->request->isPost()) {
             Config::getInstance()->lock();
-            $mdl = $this->getModel();
-            if ($uuids != null) {
-                foreach (explode(",", $uuids) as $uuid) {
-                    $node = $mdl->getNodeByReference($path . '.' . $uuid);
-                    if ($node != null) {
-                        $result[$uuid]['changed'] = true;
-                        if ($enabled == "0" || $enabled == "1") {
-                            $result[$uuid]['result'] = !empty($enabled) ? "Enabled" : "Disabled";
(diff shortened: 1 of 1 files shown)


src/opnsense/mvc/app/controllers/OPNsense/Base/ApiMutableModelControllerBase.php
```diff
@@ -594,40 +594,45 @@ public function setBase($post_field, $path, $uuid, $overlay = null)
...
+                            $result[$uuid]['result'] = "Disabled";
+                            $node->enabled = "0";
+                        } else {
+                            $result[$uuid]['result'] = "Enabled";
+                            $node->enabled = "1";
+                        }
+                        // if item has toggled, serialize to config and save
+                        if ($result[$uuid]['changed']) {
+                            // something changed, function used to be a single-fire action.
+                            // this preserves some notion of API backwards compatibility
```
swhite2: @AdSchellevis the API result has changed in format, for your consideration

AdSchellevis:
@swhite2 how about https://github.com/opnsense/core/pull/9768/changes/ce8360550c4c41af86ba48567a3febf04a8d6f96? it's slightly more efficient and only accepts multi row toggles with an explicit value.

swhite2:
@AdSchellevis much better, but the API `$result` still doesn't mean much in this way (`changed` and `result` showing state change for the last entry in the list of uuids). This is fine if we don't consider it important.

AdSchellevis:
@swhite2 for the list of updates it doesn't indeed, I was doubting about adding a counter in these cases, but for some reason I don't expect many will use the output here. We can always add these easily later was my thought.

swhite2:
@AdSchellevis Fair enough, if the rest of the PR is ok I'll merge it

AdSchellevis:
@swhite2 be my guest, rest looks good, simplifies quite some code, nice.

Issue #9757, opened 2026-02-07
Add direct link to jump to Live search filterin one rule


adrianhiller:
**Important notices**

Before you add a new report, we ask you kindly to acknowledge the following:

- [ x ] I have read the contributing guide lines at https://github.com/opnsense/core/blob/master/CONTRIBUTING.md
- [ x ] I am convinced that my issue is new after having checked both open and closed issues at https://github.com/opnsense/core/issues?q=is%3Aissue

**Is your feature request related to a problem? Please describe.**

In order to debug a firewall rule, it would be helpful to have an action button at the end of a rule line in the new rule UI to jump to the live log file directly with the filter set to the rule id.

**Describe the solution you like**

Have an action button at the right side of a rule in the new rule UI.
Target: /ui/diagnostics/firewall/log filter to rule_id={{rule_id}}

**Describe alternatives you considered**

Copy rule id on rule editing and building the filter manually

**Additional context**

Monviech:
Hmm this sounds cool and useful. Not sure about another command though. More than 5 commands will make things cumbersome especially on small screens.

I had the idea to maybe put a link into the log command's selectpicker that would say "Search Livelog" or something, but I have to think about feasability first, if there is already a facility that can work with such a link. I'll assign this to me and give it a look, but no promises yet.

Monviech:
Hello, we have implemented it. You can test it via:

`opnsense-patch https://github.com/opnsense/core/commit/63e0b92278ad4c5e4b0db9746599903089bf14dd`

(Apply the patch from a root shell, you can revert it with the same command if you have any issues.)

PR #9770, opened 2026-02-10
Firewall: Rules [new]: Add a command button to open the live log with prefilled rule ID

Monviech:
Fixes: https://github.com/opnsense/core/issues/9757

Uses a url hash inside fw_log.volt to set a filter when opening it through a link from filter_rule.volt

The url hash can set any type of a single filter, so it can be reused in other pages as well.

### DIFF
--- a/src/opnsense/mvc/app/views/OPNsense/Diagnostics/fw_log.volt
+++ b/src/opnsense/mvc/app/views/OPNsense/Diagnostics/fw_log.volt
@@ -887,15 +887,14 @@
         });
 
         table.on('tableBuilt', () => {
-            const rawUrlHash = getUrlHash('filter');
-            if (rawUrlHash) {
-                try {
-                    const filter = JSON.parse(decodeURIComponent(rawUrlHash));
-                    filterVM.addFilter(filter);
-                    history.replaceState(null, '', location.pathname);
-                } catch (e) {
-                    // ignore malformed hashes
-                }
+            const params = new URLSearchParams(window.location.hash.slice(1));
+            const field = params.get('field');
+            const operator = params.get('operator');
+            const value = params.get('value');
+
+            if (field && operator && value) {
+                filterVM.addFilter({ field, operator, value });
+                history.replaceState(null, '', location.pathname);
             }
         });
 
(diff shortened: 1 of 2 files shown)


src/opnsense/mvc/app/views/OPNsense/Diagnostics/fw_log.volt
```diff
@@ -882,12 +882,21 @@
...
 
         $reset.on('click', function() {
             tableWrapper.bootgrid('setPersistence', false);
             location.reload();
         });
 
+        const raw = getUrlHash('filter');
+        if (raw) {
+            const [field, operator, ...rest] = decodeURIComponent(raw).split(',');
+            const value = rest.join(',');
```
swhite2: the `value` can't contain multiple entries, however, the `field` can (e.g. `['src', 'dst']`), which adds a "combined filter":

In the case of "contains" or "is", this is `src OR dst`, in the case of "does not contain" or "is not", this is `src AND dst`. This logic happens internally, but currently the API provides `addFilter()` and `addCombinedFilter`, logically this should be reduced to the single function call `addFilter()` and detect array usage.

There will be no callers for this yet, however, the "filter" URI component suggests any sort of filter can be set.

src/opnsense/mvc/app/views/OPNsense/Firewall/filter_rule.volt
```diff
@@ -235,29 +235,48 @@
...
+                        const rowId = row.uuid || "";
+                        const hasUuid = rowId.includes("-");
+
+                        const logSearchCommand = (rid, log) => {
+                            const loggingEnabled = log === '1' || log === true;
+                            if (!loggingEnabled) return '';
+
+                            return `
+                                <a href="/ui/diagnostics/firewall/log#filter=${encodeURIComponent(
+                                    JSON.stringify({ field: 'rid', operator: '=', value: rid })
```
swhite2: I'm still not 100% convinced this is the way to go as it produces very large strings.

Perhaps we should stick to `URLSearchParams()`

swhite2 (changes requested):

src/opnsense/mvc/app/views/OPNsense/Firewall/filter_rule.volt
Monviech: Okay I changed it to URLSearchParams()

PR #9773, opened 2026-02-11
Firewall: Rules [new]: Reduce complexity in url hash handling and when using firewall_rule_lookup

Monviech:
Links from the log will now open the "__any" view and search the UUID of the rule via the searchPhrase.

This is more user friendly and uses less code.

Requires: https://github.com/opnsense/core/pull/9780 (but won't break without it either)


swhite2 (approved):
Looks good to me.

`firewall_rule_lookup` currently doesn't handle references to other pages very well (for example: https://github.com/opnsense/core/blob/master/src/etc/inc/plugins.inc.d/captiveportal.inc#L132). This will open the captive portal page, but open an empty zone as I think the UUID and path is misinterpreted.

As you do here, it may be better to switch that behavior to a page switch + UUID search instead of a direct edit as well.

Monviech:
@swhite2 Yeah I think searching an UUID will be always preferred because then the user can decide what to do on the page, e.g. just inspecting what the item is, deleting it, or toggling another command.

Issue #9729, opened 2026-02-04
HTML encoding issue in Firewall Dashboard widget


ZeroByteGitter:
**Important notices**

Before you add a new report, we ask you kindly to acknowledge the following:

- [x] I have read the contributing guide lines at https://github.com/opnsense/core/blob/master/CONTRIBUTING.md
- [x] I am convinced that my issue is new after having checked both open and closed issues at https://github.com/opnsense/core/issues?q=is%3Aissue

**Describe the bug**

After reading the bug regarding the export/import issue ( New rules UI [ 26.1], HTML encoding issue on rules export. #9694 ) I wanted to bring up a similar problem, that happens in the same way on the **Firewall Dasboard widget with the circle diagram, converting <,> into HTML "&lt;" and "&gt;"**

As far as I'm aware this problem exists since OPNsense V25.1, but I can't precisely track back the first time I've seen it. I also searched back related issues but couldn't find any other mention.

**To Reproduce**

1. Having rules with < or >
2. Open the Firewall Dashboard widget
4. See the wrong sign representation

**Expected behavior**

Convert > as >, here seen on a old screenshot from OPNsense V24.7.9 below.

**Describe alternatives you considered**

N/A

**Screenshots**

OPNsense V26.1.1
<img width="785" height="471" alt="Image" src="https://github.com/user-attachments/assets/13dffa05-b93d-4098-9f2f-b314ffbc82ec" />

Old screenshot from OPNsense V24.7.9:
<img width="688" height="409" alt="Image" src="https://github.com/user-attachments/assets/3db62ae7-9756-4542-b36d-0b9ab757226d" />

**Environment**

OPNsense 26.1.1 (amd64)

PR #9776, opened 2026-02-11
Fix encoding issue

sopex:
Fix: https://github.com/opnsense/core/issues/9729


AdSchellevis:
@sopex thanks, move it around a bit in https://github.com/opnsense/core/commit/0b5f58511339e46c165b66e70e0533be8c3b2282

PR #9779, opened 2026-02-12
bootgrid: batch delete-selected by default

swhite2:
Putting this up as a PR due to the `checkAndThrowValueInUse` modification.

Functionally, if a user selects multiple items and deletes them, and if the "in use" error is thrown, it helps to know which to-be-deleted item in particular caused this error, include this in the exception.

### DIFF
--- a/src/opnsense/mvc/app/controllers/OPNsense/Base/ApiMutableModelControllerBase.php
+++ b/src/opnsense/mvc/app/controllers/OPNsense/Base/ApiMutableModelControllerBase.php
@@ -537,9 +537,11 @@ abstract class ApiMutableModelControllerBase extends ApiControllerBase
                 return $result;
             }
             foreach ($uuids as $uuid) {
+                $n = $rootnode?->$uuid;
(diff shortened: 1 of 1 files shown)


src/opnsense/mvc/app/controllers/OPNsense/Base/ApiMutableModelControllerBase.php
```diff
@@ -119,8 +119,17 @@ protected function checkAndThrowValueInUse($token, $contains = true, $only_mvc =
             $xpath = "//*[text() = '{$token}']";
         }
         $usages = [];
+        $cfg = Config::getInstance()->object();
+
+        $used_by = $cfg->xpath("//*[@uuid and @uuid='{$token}']");
```
fichtner: I wouldn't mind making this a proper Config function instead of going for the xpath and hardcoded expression

swhite2: A little bird told me that this expression may not even be necessary, commit incoming from @AdSchellevis ;)

Issue #9777, opened 2026-02-12
mvc: Allow searching an item via UUID in searchPhrase POST


Monviech:
**Important notices**

Before you add a new report, we ask you kindly to acknowledge the following:

- [x] I have read the contributing guide lines at https://github.com/opnsense/core/blob/master/CONTRIBUTING.md
- [x] I am convinced that my issue is new after having checked both open and closed issues at https://github.com/opnsense/core/issues?q=is%3Aissue

**Is your feature request related to a problem? Please describe.**

I need to find a specific rule in the firewall rule grid, e.g. if it is linked via url hash. (#search=UUID)

Right now, UUIDs are not found, so there is no unique key to find a specific rule with.

Reference: https://github.com/opnsense/core/pull/9773

**Describe the solution you like**

If the searchPhrase is a valid UUID, search only through the UUIDs and return the item.

If the searchPhrase is not a valid UUID, do not match partially on it, but only on the other items.

**Describe alternatives you considered**

Something fishy like this:

https://github.com/opnsense/core/blob/29a74a292f7ef989b10fd44ccb72460efdc6d958/src/opnsense/mvc/app/controllers/OPNsense/Firewall/Api/FilterController.php#L201-L208

**Additional context**

It would be nice if it could work generically in all searches.
I think it's decided here:
https://github.com/opnsense/core/blob/bc6035e10192753f6074f2af0880f4be2c946e80/src/opnsense/mvc/app/controllers/OPNsense/Base/ApiControllerBase.php#L67-L71

AdSchellevis:
fixed in https://github.com/opnsense/core/pull/9780

PR #9780, opened 2026-02-12
mvc: collect uuid field so it can be searched

Monviech:
But only if the searchPhrase contains a valid uuid to avoid matching partial strings

Idea for: https://github.com/opnsense/core/issues/9777

### DIFF
--- a/src/opnsense/mvc/app/library/OPNsense/Core/Type.php
+++ b/src/opnsense/mvc/app/library/OPNsense/Core/Type.php
@@ -44,7 +44,7 @@ class Type
     }
 
     /**
-     * ckeck if an array contains a valid UUID
+     * check if an array contains a valid UUID
      */
     public static function containsUUID(array $tokens): bool
     {


Monviech:
@AdSchellevis It's an initial idea, it works this way, and prevents partial matches. Though I'm sure there are some considerations to make here, so I'm leaving this with you as an example of what I need/want.

src/opnsense/mvc/app/library/OPNsense/Base/UIModelGrid.php
```diff
@@ -48,6 +48,18 @@ public function __construct($DataField)
         $this->DataField = $DataField;
     }
 
+    public function isValidUUID($uuid)
```
fichtner: we already have a function with the same name for this... a static util might be good ;)

Monviech: I know I copy pastad from here:
https://github.com/opnsense/core/blob/c0569f86d5538b4312dd7fd8f8613664db8dbed7/src/opnsense/mvc/app/controllers/OPNsense/Base/ApiMutableModelControllerBase.php#L94

I kept it simple for now as it's just to show an idea.

AdSchellevis: a static might be good indeed, after this, we probably have 3 copies of the logic.

My suggestion would be to add a class named `Types` in https://github.com/opnsense/core/tree/master/src/opnsense/mvc/app/library/OPNsense/Core

src/opnsense/mvc/app/library/OPNsense/Base/UIModelGrid.php
```diff
@@ -77,6 +89,13 @@ public function fetchBindRequest(
         }
 
         $searchPhrase = $request->get('searchPhrase', 'string', '');
+        // we want the uuid field added, but not search through partial ones to limit false results
```
AdSchellevis: technically, we should move the "collect field when" to somewhere around here:

https://github.com/opnsense/core/blob/c0569f86d5538b4312dd7fd8f8613664db8dbed7/src/opnsense/mvc/app/controllers/OPNsense/Base/ApiMutableModelControllerBase.php#L429

Which logically means we should also move `$searchPhrase` outside of the block as we are going to tokenize it later (and we need the parts to know if one of them matched a uuid). 

https://github.com/opnsense/core/blob/c0569f86d5538b4312dd7fd8f8613664db8dbed7/src/opnsense/mvc/app/library/OPNsense/Base/UIModelGrid.php#L129

The only annoying thing is `fetchBindRequest()` is used in some legacy plugins (and captiveportal), which means we need to take backwards compatibility into account.

src/opnsense/mvc/app/library/OPNsense/Base/UIModelGrid.php
```diff
@@ -77,6 +89,13 @@ public function fetchBindRequest(
...
+        if (!empty($fields) && $this->isValidUUID($searchPhrase)) {
```
AdSchellevis: when moving `isValidUUID()` into a static, we could also add `containsUUID(array $tokens)` to simplify the caller a bit.

src/opnsense/mvc/app/library/OPNsense/Core/Type.php
```diff
@@ -35,7 +35,8 @@ class Type
      * @param string $uuid
      * @return boolean
      */
-    public static function isValidUUID(string $uuid) {
+    public static function isValidUUID(string $uuid)
```
fichtner: just for clarity I think `isUUID` is enough if we touch all of this anyway

Monviech: But does being an UUID also make it a valid one?

fichtner: This matches the style in the Util class, e.g. `isIpAddress`

fichtner: (the answer to your question is likely "yes")

Monviech:
Thanks, follow up ticket here: https://github.com/opnsense/core/issues/9797

PR #9786, opened 2026-02-13
Update system_gateway_groups.php, Fatal Error on Gateway Group Deletion

s4saken:
When deleting gateway group i was getting fatal error on multiple devices. I dont know when it started. This involves opnsense installation from v22 updated all the way to v26.1.2. Unable to delete gateway groups as a result.

Fatal error: Uncaught TypeError: Cannot access offset of type string on string in /usr/local/www/system_gateway_groups.php:41 Stack trace: #0 {main} thrown in /usr/local/www/system_gateway_groups.php on line 41 

Did some research, appeared to be an empty filter rule this code could not handle and resulted in fatal error.

Rewrote logic.

### DIFF
--- a/src/www/system_gateway_groups.php
+++ b/src/www/system_gateway_groups.php
@@ -38,7 +38,7 @@ $a_gateways = (new \OPNsense\Routing\Gateways())->gatewaysIndexedByName();
 if ($_SERVER['REQUEST_METHOD'] === 'POST') {
     if (isset($_POST['act']) && $_POST['act'] == "del" ) {
         if (!empty($a_gateway_groups[$_POST['id']])) {
-            if (!empty($config['filter']['rule'])) {
+            if (isset($config['filter']['rule']) && is_array($config['filter']['rule'])) {
                 foreach ($config['filter']['rule'] as $idx => $rule) {
                     if (isset($rule['gateway']) && $rule['gateway'] == $a_gateway_groups[$_POST['id']]['name']) {
                         unset($config['filter']['rule'][$idx]['gateway']);


src/www/system_gateway_groups.php
```diff
@@ -59,6 +61,7 @@
     }
 }
 
+
```
fichtner:
```suggestion
```

src/www/system_gateway_groups.php
```diff
@@ -38,9 +38,11 @@
 if ($_SERVER['REQUEST_METHOD'] === 'POST') {
     if (isset($_POST['act']) && $_POST['act'] == "del" ) {
         if (!empty($a_gateway_groups[$_POST['id']])) {
-            foreach ($config['filter']['rule'] as $idx => $rule) {
-                if ($rule['gateway'] == $a_gateway_groups[$_POST['id']]['name']) {
-                    unset($config['filter']['rule'][$idx]['gateway']);
+            if (isset($config['filter']['rule']) && is_array($config['filter']['rule'])) {
```
fichtner, line 41:
```suggestion
            if (!empty(i$config['filter']['rule'])) {
```

src/www/system_gateway_groups.php
```diff
@@ -38,9 +38,11 @@
 if ($_SERVER['REQUEST_METHOD'] === 'POST') {
     if (isset($_POST['act']) && $_POST['act'] == "del" ) {
         if (!empty($a_gateway_groups[$_POST['id']])) {
-            foreach ($config['filter']['rule'] as $idx => $rule) {
-                if ($rule['gateway'] == $a_gateway_groups[$_POST['id']]['name']) {
-                    unset($config['filter']['rule'][$idx]['gateway']);
+            if (!empty(i$config['filter']['rule'])) {
```
fichtner:
```suggestion
            if (!empty($config['filter']['rule'])) {
```

fichtner:
Did you migrate your rules to the new GUI and removed the old ones? Looks like it. The fix is ok in general but a) there could be more spots like this now, rules were never really empty before and b) I'd like to make these shorter and try something else instead of these if-foreach constructs that riddle the backend code for historic reasons.

fichtner:
Note to self:

> Fatal error: Uncaught TypeError: Cannot access offset of type string on string

Caused by a single `<filter/>` XML entry

Quick test shows that this catches the error:

`foreach ($config['filter']['rule'] ?? [] as $rule)`

s4saken:
> Did you migrate your rules to the new GUI and removed the old ones? Looks like it. The fix is ok in general but a) there could be more spots like this now, rules were never really empty before and b) I'd like to make these shorter and try something else instead of these if-foreach constructs that riddle the backend code for historic reasons.

Compact is always good! thanks.

Since more people will be facing this, after migration. I guess a patch wont hurt.

fichtner:
There are a few combinations of errors that could happen here. Need to test all of them and I think not all will be happy with the ?? operator but it’s almost 11 pm so maybe tomorrow 🙃

fichtner:
I think I want to go with https://github.com/opnsense/core/commit/a971a6088 and we can clean up a few more spots which also makes code coverage better of such a change (although in practice it only adds the new mode to `config_read_array()` so we do not have to deal with regressions in places where the new `$insert_mode` is true for backwards compatibility.

I'll test test, add more refactors and commit on Monday.


Cheers,
Franco

fichtner:
Tested and works:

     # opnsense-patch feddc1f


Thanks,
Franco

PR #9790, opened 2026-02-15
mvc: add additional dnsmasq server IP validation

Greelan:
Avoids dnsmasq failing to start due to invalid config options for DNS servers

### DIFF
--- a/src/opnsense/mvc/app/models/OPNsense/Dnsmasq/Dnsmasq.php
+++ b/src/opnsense/mvc/app/models/OPNsense/Dnsmasq/Dnsmasq.php
@@ -403,7 +403,7 @@ class Dnsmasq extends BaseModel
 
             if (
                 !$option->value->isEmpty() &&
-                in_array($option->option6->getValue(), ['21','23','31','34'])
+                in_array($option->option6->getValue(), ['22','23','31','34'])
             ) {
                 $values = preg_split('/\s*,\s*/', trim($option->value->getValue()), -1, PREG_SPLIT_NO_EMPTY);
                 foreach ($values as $value) {


Greelan:
Note that my focus is DNS server validation since that is what tripped me up, but if desired the PR can be extended to other dhcp-options that configure IPs as per RFC 2132

src/opnsense/mvc/app/models/OPNsense/Dnsmasq/Dnsmasq.php
```diff
@@ -385,12 +385,46 @@ public function performValidation($validateFullModel = false)
 
             if (
                 !$option->value->isEmpty() &&
-                !$option->option6->isEmpty()
+                !$option->option->isEmpty() &&
+                (string)$option->option->getValue() === '6'
```
AdSchellevis:
```suggestion
                $option->option->getValue() == '6'
```

src/opnsense/mvc/app/models/OPNsense/Dnsmasq/Dnsmasq.php
```diff
@@ -385,12 +385,46 @@ public function performValidation($validateFullModel = false)
...
             ) {
                 $values = array_map('trim', $option->value->getValues());
+
+                foreach ($values as $value) {
+                    if (!Util::isIpv4Address(trim($value, '[]'))) {
```
AdSchellevis:
```suggestion
                    if (!Util::isIpv4Address($value)) {
```
I expect the brackets only apply for the ipv6 case

src/opnsense/mvc/app/models/OPNsense/Dnsmasq/Dnsmasq.php
```diff
@@ -385,12 +385,46 @@ public function performValidation($validateFullModel = false)
 
             if (
                 !$option->value->isEmpty() &&
-                !$option->option6->isEmpty()
+                !$option->option->isEmpty() &&
```
AdSchellevis:
```suggestion
```
can ditch as '6' <> ''

AdSchellevis (changes requested):

src/opnsense/mvc/app/models/OPNsense/Dnsmasq/Dnsmasq.php
Greelan: I suspect you are right. I was thinking of needing to sanitise the input for testing but if the input has square brackets it wouldn't be a valid IPv4 address anyway xD

src/opnsense/mvc/app/models/OPNsense/Dnsmasq/Dnsmasq.php
Greelan: Not sure I follow this, sorry. I'm trying to test that the DHCP v4 option field is not empty.

Greelan: Ah, nvm, I understand - it is redundant xD

src/opnsense/mvc/app/models/OPNsense/Dnsmasq/Dnsmasq.php
```diff
@@ -385,12 +385,45 @@ public function performValidation($validateFullModel = false)
...
+                !$option->value->isEmpty() &&
+                !$option->option6->isEmpty()   
+            ) {
+                $values = array_map('trim', $option->value->getValues());
+
                 foreach ($values as $value) {
+                    $raw = trim($value, '[]');
+                    if (
+                        !Util::isIpv6Address($raw) &&
+                        (string)$option->option6->getValue() === '23'
```
Monviech:
```suggestion
                        $option->option6->getValue() === '23'
```

src/opnsense/mvc/app/models/OPNsense/Dnsmasq/Dnsmasq.php
```diff
@@ -385,12 +385,45 @@ public function performValidation($validateFullModel = false)
...
-                !$option->option6->isEmpty()
+                $option->option->getValue() == '6'
             ) {
                 $values = array_map('trim', $option->value->getValues());
+
+                foreach ($values as $value) {
+                    if (!Util::isIpv4Address($value)) {
+                        $messages->appendMessage(
+                            new Message(
+                                gettext("Only IPv4 addresses are allowed for DHCP option 6."),
```
Monviech:
```suggestion
                                gettext("Only IPv4 addresses are allowed for this DHCP option."),
```

Then it can be extended more easily later if more values should be added to the validation I think.

src/opnsense/mvc/app/models/OPNsense/Dnsmasq/Dnsmasq.php
```diff
@@ -385,12 +385,45 @@ public function performValidation($validateFullModel = false)
...
+
                 foreach ($values as $value) {
+                    $raw = trim($value, '[]');
+                    if (
+                        !Util::isIpv6Address($raw) &&
+                        (string)$option->option6->getValue() === '23'
+                    ) {
+                        $messages->appendMessage(
+                            new Message(
+                                gettext("Only IPv6 addresses are allowed for DHCPv6 option 23."),
```
Monviech:
```suggestion
                                gettext("Only IPv6 addresses are allowed for this DHCPv6 option."),
```

src/opnsense/mvc/app/models/OPNsense/Dnsmasq/Dnsmasq.php
Greelan: Thanks, see changes made

src/opnsense/mvc/app/models/OPNsense/Dnsmasq/Dnsmasq.php
Greelan: Thanks, see changes made

src/opnsense/mvc/app/models/OPNsense/Dnsmasq/Dnsmasq.php
Greelan: Thanks, see changes made

src/opnsense/mvc/app/models/OPNsense/Dnsmasq/Dnsmasq.php
```diff
@@ -385,12 +385,45 @@ public function performValidation($validateFullModel = false)
...
+                        );
+                    }
+                }
+            }
+
+            if (
+                !$option->value->isEmpty() &&
+                !$option->option6->isEmpty()   
+            ) {
+                $values = array_map('trim', $option->value->getValues());
```
fichtner: Looks like excessive trimming to me. We try to bail with white spaces in normal validation. Normally getValues() should be trusted validated input, also below trimming input for compare.

src/opnsense/mvc/app/models/OPNsense/Dnsmasq/Dnsmasq.php
```diff
@@ -385,12 +385,45 @@ public function performValidation($validateFullModel = false)
...
+
                 foreach ($values as $value) {
+                    $raw = trim($value, '[]');
```
fichtner: In which cases are brackets allowed here? Brackets should only be needed if ports are appended

src/opnsense/mvc/app/models/OPNsense/Dnsmasq/Dnsmasq.php
```diff
@@ -385,12 +385,45 @@ public function performValidation($validateFullModel = false)
 
             if (
                 !$option->value->isEmpty() &&
-                !$option->option6->isEmpty()
+                $option->option->getValue() == '6'
```
fichtner: At this point I’m definitely bike shedding but ->isEqual(‘6’) is a thing too

src/opnsense/mvc/app/models/OPNsense/Dnsmasq/Dnsmasq.php
Monviech: The brackets are needed around all ipv6 addresses or dnsmasq fails to load. I added that a while ago as validation.

src/opnsense/mvc/app/models/OPNsense/Dnsmasq/Dnsmasq.php
Greelan: I've got no dog in this fight, this was the existing code. Happy if you want to change it

src/opnsense/mvc/app/models/OPNsense/Dnsmasq/Dnsmasq.php
Greelan: I adopted Ad's suggestion, will leave it to you guys to adopt the preferred xD. But now the test is of values in an array anyway

Greelan:
Now added additional DHCP option IP types for validation

src/opnsense/mvc/app/models/OPNsense/Dnsmasq/Dnsmasq.php
Greelan: Done

Greelan:
Hmm, this is now no longer successfully parsing comma-separated lists of otherwise valid IPs. Will need to investigate further

Greelan:
OK, I realised now that the `value` field is a `TextField`, not an array, and so `getValues()` returns the entire value as one array element. (It makes sense it is a `TextField` since it is a multi-purpose value field, and so a particular format can't be assumed.)

Hence trimming, and splitting on commas, is necessary to ensure proper validation. Incidentally this means the current approach of validating starting and ending square brackets for IPv6 addresses also doesn't work as intended, since it validates on the entire string, not each individual IPv6 address.

Note that the output to dnsmasq doesn't appear to need trimming for whitespace, as dnsmasq handles that (I tested using spaces at least).

Greelan:
@Monviech, arguably both IPv6 validation tests can be applied only for DHCPv6 option 23? Was there another option that you wanted IPv6 address validation for?

22, 23, 31, 34 seem the relevant list.

Greelan:
Duly implemented xD

src/opnsense/mvc/app/models/OPNsense/Dnsmasq/Dnsmasq.php
```diff
@@ -385,20 +385,45 @@ public function performValidation($validateFullModel = false)
...
-                !$option->option6->isEmpty()
+                in_array($option->option->getValue(), ['3','4','5','6','7','8','9','10','11'])
             ) {
-                $values = array_map('trim', $option->value->getValues());
-                foreach ($values as $value) {
-                    if (
-                        Util::isIpv6Address(trim($value, '[]')) &&
-                        !(str_starts_with($value, '[') && str_ends_with($value, ']'))
-                    ) {
+                $ips = preg_split('/\s*,\s*/', trim($option->value), -1, PREG_SPLIT_NO_EMPTY);
```
Monviech:
```suggestion
                $values = preg_split('/\s*,\s*/', trim($option->value), -1, PREG_SPLIT_NO_EMPTY);
```

Could we keep the ``values`` terminology, we don't know yet whats inside here, could also be ``192.168.1.1, foo, bar, 8.8.8.8``

Monviech: Also why aren't we using ``$option->value->getValue()`` here for cast safety?

Greelan: OK

Greelan: Agreed

src/opnsense/mvc/app/models/OPNsense/Dnsmasq/Dnsmasq.php
```diff
@@ -385,20 +385,44 @@ public function performValidation($validateFullModel = false)
...
+                            )
+                        );
+                        break;
+                    }
+                }
+            }
+
+            if (
+                !$option->value->isEmpty() &&
+                in_array($option->option6->getValue(), ['21','23','31','34'])
```
Monviech: ``21``is ``OPTION_SIP_SERVER_D`` (which looks like its a Domain list), shouldn't it be ``22`` for ``OPTION_SIP_SERVER_A`` ?

Greelan: Correct, my bad. I thought 22. Wrote 21

PR #9798, opened 2026-02-16
Feature: Added support for tls-crypt-v2 for OpenVPN (client) static keys

ninjatrek2891:
I made some small changes to a couple of PHP files to support a tls-crypt-v2 key in the config.

After deploying, the menu option appears under OpenVPN static keys and is added properly to the instance conf file in /var/etc/openvpn/

I also made a good connection to my openvpn instance.

### DIFF
--- a/src/opnsense/mvc/app/controllers/OPNsense/OpenVPN/forms/dialogInstance.xml
+++ b/src/opnsense/mvc/app/controllers/OPNsense/OpenVPN/forms/dialogInstance.xml
@@ -314,7 +314,7 @@
         <type>dropdown</type>
         <help>
             Add an additional layer of HMAC authentication on top of the TLS control channel to mitigate DoS attacks and attacks on the TLS stack.
-            The prefixed mode determines if this measurement is only used for authentication (--tls-auth) or includes encryption (--tls-crypt).
+            The prefixed mode determines if this measurement is only used for authentication (--tls-auth) or includes encryption (--tls-crypt) and (--tls-crypt-v2).
         </help>
         <grid_view>
             <visible>false</visible>
--- a/src/opnsense/mvc/app/models/OPNsense/OpenVPN/OpenVPN.xml
+++ b/src/opnsense/mvc/app/models/OPNsense/OpenVPN/OpenVPN.xml
@@ -399,7 +399,7 @@
                     <OptionValues>
                         <auth>auth (Authenticate control channel packets)</auth>
                         <crypt>crypt (Encrypt and authenticate all control channel packets)</crypt>
-                        <crypt-v2>crypt-v2(Encrypt and authenticate all control channel packets)</crypt-v2>
+                        <crypt-v2>crypt-v2 (Encrypt and authenticate all control channel packets)</crypt-v2>
                     </OptionValues>
                 </mode>
                 <key type="TextField">


src/opnsense/mvc/app/controllers/OPNsense/OpenVPN/forms/dialogStaticKey.xml
```diff
@@ -10,7 +10,7 @@
         <label>Mode</label>
         <width>220px</width>
         <type>dropdown</type>
-        <help>Define the use of this key, authentication (--tls-auth) or authentication and encryption (--tls-crypt)</help>
+        <help>Define the use of this key, authentication (--tls-auth) or authentication and encryption (--tls-crypt) or (--tls-crypt-v2)</help>
```
AdSchellevis, line 13: should probably add `--tls-crypt-v2` here too:

https://github.com/opnsense/core/blob/dfa65e1589b2665f73ba2f6f10e596fa9070b5a8/src/opnsense/mvc/app/controllers/OPNsense/OpenVPN/forms/dialogInstance.xml#L317

ninjatrek2891:
I completely overlooked that one! I added the context on that line and fixed a missing space in the dropbox menu context.

AdSchellevis:
@ninjatrek2891 no problem, looking at our key-generation button, I think we're missing some logic to generate the proper key type as well. How did you test this feature? (I expect you pasted your own key in there)

ninjatrek2891:
I tested the changes by changing the php files directly on the firewall via the shell. It all looks okay on my end, the openvpn instance is connecting with the tls-crypt-v2 key.

I've parsed the /var/etc/openvpn/instance.conf file and it all looks okay, the key <tls-crypt-v2> looks parsed properly.

To answer your question, I used my own key. I did not use the key generation as I did have my own key.

Also tls-crypt v2 requires a server and client key. The logic for it to build could be builtin Opnsense. But I think, it might be wiser to turn the button off or disabled it when crypt-v2 is chosen.

AdSchellevis:
let me keep this open a small while as I believe the generate button needs a fix too as it seems to only generate a single type of key.

ninjatrek2891:
We could disable the keygen button when cryptv2 is choosen.

crypt-v2 is client and server key pair. It might not belong under Static Keys as I did here.

`openvpn --tls-crypt-v2 server.key --genkey tls-crypt-v2-client --secret client.key` is needed to get a client key.

Monviech:
The main difference in tls-crypt-v2 is that each client (and their server) should get their own unique key.

The normal static key in tls-crypt was shared (=the same) between all clients and server.

ninjatrek2891:
Yes, I completely understand that the crypt v2 might not be well suited descriptive wise at Static Keys.

Except for the key-gen button, it worked. Is it the right philosophy to put it at Static Keys, I don't know. But it worked (for now) and I don't have to rely on the legacy plugin.

Monviech:
I added this now it should be simpler to add the new mode to the key generation:

https://github.com/opnsense/core/commit/242fc74ccbdb738fc34ac3c9efbbf51dd8aab042

Monviech:
The remaining challenge would be:
- generate a client tls-crypt-v2 key key out of the instance (server) tls-crypt-v2 key (that has been added to the instance)
- add this unique key to the client export profiles when they are downloaded (unique per profile)
- I assume each profile download can have a new key, even when redownloading, so we do not need to keep track of each generated key and just generate a new one on each download

Issue #9797, opened 2026-02-16
mcv: Collect direct callers of UIModelGrid() to remove backwards compat in fetchBindRequest()


Monviech:
**Important notices**

Before you add a new report, we ask you kindly to acknowledge the following:

- [x] I have read the contributing guide lines at https://github.com/opnsense/core/blob/master/CONTRIBUTING.md
- [x] I am convinced that my issue is new after having checked both open and closed issues at https://github.com/opnsense/core/issues?q=is%3Aissue

**Is your feature request related to a problem? Please describe.**

https://github.com/opnsense/core/blob/d976ae7b6b0f06c235d22de087e4904c28b1286c/src/opnsense/mvc/app/library/OPNsense/Base/UIModelGrid.php#L83-L86

Introduced in: https://github.com/opnsense/core/commit/7738d7e865025022035025c8b65578a57c94feab

**Describe the solution you like**

List of all callers in core and plugins to clean this up and remove backwards compat of fetchBindRequest by using the new ``search_tokens`` parameter or moving the caller up to [searchBase()](https://github.com/opnsense/core/blob/d976ae7b6b0f06c235d22de087e4904c28b1286c/src/opnsense/mvc/app/controllers/OPNsense/Base/ApiMutableModelControllerBase.php#L400)

**Callers**

https://github.com/opnsense/core/blob/f399b33df924e2c5b8a09f580b15ee696549414d/src/opnsense/mvc/app/controllers/OPNsense/CaptivePortal/Api/ServiceController.php#L168-L169

Monviech:
git grep -r "UIModelGrid"

src/opnsense/mvc/app/controllers/OPNsense/IDS/Api/SettingsController.php:use OPNsense\Base\UIModelGrid;
src/opnsense/mvc/app/controllers/OPNsense/Monit/Api/SettingsController.php:use OPNsense\Base\UIModelGrid;
src/opnsense/mvc/app/controllers/OPNsense/Syslog/Api/SettingsController.php:use OPNsense\Base\UIModelGrid;

Monviech:
@fichtner 
Plugins will be quite a battle as it seems, not sure if it should be mine :D

Commonly used plugins like AcmeClient seem to use it as well

```
net-mgmt/zabbix-agent/src/opnsense/mvc/app/controllers/OPNsense/ZabbixAgent/Api/SettingsController.php:use OPNsense\Base\UIModelGrid;
[...]

PR #9799, opened 2026-02-16
Services: Captive Portal: Move template actions out of the ServiceController into its own TemplateController, so it can use the ApiMutableModelControllerBase methods

Monviech:
For: https://github.com/opnsense/core/issues/9797

Please note that this breaks the current API for template actions since it moves the calls to a new controller.

It has to be decided if this is okay, or if there should be stubs in ServiceController for backwards compatibility.

### DIFF
--- a/src/opnsense/mvc/app/controllers/OPNsense/CaptivePortal/Api/TemplateController.php
+++ b/src/opnsense/mvc/app/controllers/OPNsense/CaptivePortal/Api/TemplateController.php
@@ -125,16 +125,22 @@ class TemplateController extends ApiMutableModelControllerBase
         }
 
         $template->name = $templateName;
-
-        $validation = $this->validate();
-
-        if (!empty($validation['validations'])) {
-            return $validation;
+        $errorMsg = [];
+        foreach ($this->getModel()->performValidation() as $validation_message) {
+            $errorMsg[] = (string)$validation_message;
         }
 
-        $this->save(false, true);
+        if (!empty($errorMsg)) {
+            return [
+                "name" => (string)$template->name,
+                "error" => implode("\n", $errorMsg)
+            ];
+        }
 
-        return ["result" => "saved"];
+        $this->getModel()->serializeToConfig();
+        Config::getInstance()->save();
+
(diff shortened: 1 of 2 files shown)


src/opnsense/mvc/app/controllers/OPNsense/CaptivePortal/Api/TemplateController.php
```diff
@@ -0,0 +1,146 @@
...
+
+        Config::getInstance()->lock();
+
+        $content = $this->request->getPost("content", "striptags", "");
+        $templateName = $this->request->getPost("name", "striptags");
+
+        if ($this->request->hasPost("uuid")) {
+            $uuid = $this->request->getPost("uuid", "striptags");
+            $template = $this->getModel()
+                ->getNodeByReference('templates.template.' . $uuid);
```
fichtner: these line breaks break my brain :D

src/opnsense/mvc/app/controllers/OPNsense/CaptivePortal/Api/TemplateController.php
```diff
@@ -0,0 +1,146 @@
...
+            if ($template === null) {
+                return ["error" => "node not found"];
+            }
+        } else {
+            $template = $this->getModel()
+                ->getTemplateByName($templateName);
+        }
+
+        if (strlen($content) > 20 || strlen((string)$template->content) === 0) {
+            $tempFile = (new AppConfig())->application->tempDir
```
fichtner: as a style nit the operator should be last on a multi-line break so that would be "." here

src/opnsense/mvc/app/controllers/OPNsense/CaptivePortal/Api/TemplateController.php
Monviech: oops yeah this is weird

Monviech:
Assigned to @swhite after internal discussion

Monviech:
I think it was decided that this can be on master for now, so I'll merge it.

PR #9803, opened 2026-02-16
Fix typo

henryreed:
The installation wizard has a typo wherein Unbound is referred to as "Unboud" (missing the 'n' character). This commit fixes this typo.


AdSchellevis:
@henryreed thanks!

PR #9807, opened 2026-02-17
Firewall: Schedule: Fix is_schedule_inuse so it works for legacy and new rules

Monviech:
I couldn't get it to work with config_read_array(), I might need help understanding it.

EDIT: Was just because I wasn't on latest master, all okay :)

### DIFF
--- a/src/www/firewall_schedule_edit.php
+++ b/src/www/firewall_schedule_edit.php
@@ -34,19 +34,15 @@ require_once("filter.inc");
 
 function is_schedule_inuse($schedule)
 {
-    global $config;
-
     if ($schedule == '') {
         return false;
     }
 
-    // XXX: legacy and mvc rule logic coexist here
-    $rules = array_merge(
-        $config['filter']['rule'] ?? [],
-        $config['OPNsense']['Firewall']['Filter']['rules']['rule'] ?? []
-    );
+    /* loop through firewall rules looking for schedule in use */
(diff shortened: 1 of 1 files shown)


src/www/firewall_schedule_edit.php
```diff
@@ -34,13 +34,20 @@
...
 function is_schedule_inuse($schedule)
 {
+    global $config;
+
     if ($schedule == '') {
         return false;
     }
 
-    /* loop through firewall rules looking for schedule in use */
-    foreach (config_read_array('filter', 'rule', false) as $rule) {
```
fichtner, line 42: please don't revert this... use the model for the new rules?

Monviech: With the above logic I get this when editing a schedule:

```
Fatal error: Uncaught TypeError: Cannot access offset of type string on string in /usr/local/www/firewall_schedule_edit.php:43 Stack trace: #0 /usr/local/www/firewall_schedule_edit.php(708): is_schedule_inuse('OtherSchedule') #1 {main} thrown in /usr/local/www/firewall_schedule_edit.php on line 43
```

fichtner: sound have been addressed by https://github.com/opnsense/core/commit/feddc1f28 IMO

Monviech: You are right I was dumb and forgot to make upgrade today

PR #9821, opened 2026-02-19
Fix SNAT isEnabled check

aaronkrogers:
This change fixes the isEnabled check for the Source NAT table to check that the "enabled" field of each SNAT row is equal to 1 instead of 0. This prevents misleading checkbox and tooltip values on the "firewall>NAT>Source Nat" page and makes their behavior match that of other rule tables.

Before:
Screenshot From 2026-02-18 19-27-23

After:
Screenshot From 2026-02-18 19-27-36

### DIFF
--- a/src/opnsense/mvc/app/views/OPNsense/Firewall/snat_rule.volt
+++ b/src/opnsense/mvc/app/views/OPNsense/Firewall/snat_rule.volt
@@ -172,7 +172,7 @@
                         if (row.isGroup || !rowId.includes('-')) {
                             return '';
                         }
-                        const isEnabled = row[column.id] === "0";
+                        const isEnabled = row[column.id] === "1";
                         return `
                             <span class="fa fa-fw ${isEnabled ? 'fa-check-square-o' : 'fa-square-o text-muted'} bootgrid-tooltip command-toggle"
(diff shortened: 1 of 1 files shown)


fichtner:
I'm not sure but it looks like this is not entirely consistent either way

```
# git grep const.isEnabled
src/opnsense/mvc/app/views/OPNsense/Firewall/dnat_rule.volt:                        const isEnabled = row[column.id] === "0";
src/opnsense/mvc/app/views/OPNsense/Firewall/filter_rule.volt:                        const isEnabled = row[column.id] === "1";
src/opnsense/mvc/app/views/OPNsense/Firewall/npt_rule.volt:                        const isEnabled = row[column.id] === "0";
src/opnsense/mvc/app/views/OPNsense/Firewall/onat_rule.volt:                        const isEnabled = row[column.id] === "0";
src/opnsense/mvc/app/views/OPNsense/Firewall/snat_rule.volt:                        const isEnabled = row[column.id] === "0";
```

fichtner:
FWIW, I think your change is correct so let me merge it and look at the others separately. Thanks!

fichtner:
65b3a98 should take care of the others :)

Issue #9789, opened 2026-02-15
Failover process kills states when no viable failover target exists


justinship714:
What caused this to start happening randomly was my ISPs gateway monitor IP started to have latency/loss. 

**Important notices**

- [X] I have read the contributing guide lines at https://github.com/opnsense/core/blob/master/CONTRIBUTING.md
- [X] I am convinced that my issue is new after having checked both open and closed issues at https://github.com/opnsense/core/issues?q=is:issue

**Describe the bug**

When dpinger declares a gateway down, OPNsense runs the full failover process — killing all states for that gateway and reconfiguring routing — even when no viable failover target exists. The secondary interface and its gateway are both disabled in OPNsense, yet the system still destroys active connections on the only working gateway, causing self-inflicted outages.

I started experiencing intermittent DNS hangs and internet stalling across all devices. Switching to cellular bypassed the issue, pointing to the firewall. After debugging, I traced it to the gateway logs — OPNsense was repeatedly declaring WAN down, killing all states, and reconfiguring routing every 1-3 minutes. My ISP gateway had become unreliable in responding to ICMP, which exposed this bug.

Last known working version: Unknown — the issue was exposed when the ISP gateway started responding slowly to ICMP. The failover logic itself may have always behaved this way but was never triggered until now.

**To Reproduce**

1. Configure a gateway group with one active gateway (Tier 1) and a second gateway whose interface and gateway are both disabled in OPNsense
2. Set trigger level to "Member Down"
3. Have a gateway that occasionally responds slowly to ICMP (>200ms RTT)
4. Observe dpinger escalating from delay to down
5. Observe states being killed and routing reconfigured with no failover target available

**Expected behavior**

The failover process should not kill states when no viable failover target exists. If all other members of the gateway group are disabled or down, destroying states on the only working gateway serves no purpose. The system already logs `ROUTING: ignoring down gateways: WAN_DHCP, USB5GPHONE_DHCP` — it knows there's nowhere to go, yet still kills states.

**Relevant log files**

dpinger declaring the gateway down with 0% packet loss:

```
ALERT: WAN_DHCP (Addr: 47.147.221.1 Alarm: none -> delay RTT: 291.2 ms RTTd: 723.2 ms Loss: 0.0 %)
ALERT: WAN_DHCP (Addr: 47.147.221.1 Alarm: delay -> down RTT: 502.7 ms RTTd: 871.4 ms Loss: 0.0 %)
```

System log showing states killed with no failover target:

```
[...]

PR #9823, opened 2026-02-19
monitor: skip state kill when no viable failover target exists

internetsecureORG:
## Problem

When `monitor_killstates` triggers in `20-recover` for a down gateway, states are killed
unconditionally via `pfctl -k gateway`. If the only other gateway in the group is disabled
(or no other gateway exists), this destroys all active connections with nowhere to fail over
to — causing a self-inflicted outage every time dpinger reports the gateway down.

## Reproducer

1. Gateway group with WAN_DHCP (Tier 1, active) and a backup gateway (Tier 2, disabled)
2. Trigger level: Member Down
3. ISP gateway responds slowly to ICMP (exceeds latency threshold)
4. dpinger declares gateway down
5. 20-recover kills all states for WAN_DHCP
6. No failover target available — all connections destroyed
7. ISP recovers moments later, cycle repeats every 1-3 minutes

## Fix

Added `has_viable_failover_target()` before killing states. Uses the same status evaluation
logic as `Gateways::getGroups()` to check:

1. Gateway groups: is another member of the same group UP per the group's trigger settings?
2. Default gateway switching (`gw_switch_default`): is another gateway of the same protocol family UP?

When no viable target exists, the kill is skipped and logged:
`ROUTING: skipping state kill for gateway WAN_DHCP, no viable failover target`

## Related

- #9789

### DIFF
--- a/src/etc/rc.syshook.d/monitor/20-recover
+++ b/src/etc/rc.syshook.d/monitor/20-recover
@@ -32,13 +32,122 @@ require_once 'util.inc';
 require_once 'system.inc';
 require_once 'interfaces.inc';
 
+/**
+ * Check if a down gateway has a viable failover target using the same logic
+ * as Gateways::getGroups(). Without a viable target, killing states only
+ * causes a self-inflicted outage with no benefit.
+ *
+ * Checks two failover paths:
+ *   1. Gateway groups: another member of the same group is UP (per trigger settings)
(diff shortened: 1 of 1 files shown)


fichtner:
Thanks for the submission. This works around a non-issue with so much code it’s not useful to continue discussing again. See #9789.

PR #9825, opened 2026-02-19
wireguard: add command hooks

Greelan:
Intent is to mimic the wg-quick functionality for PreUp, PostUp, PreDown and PostDown

Context: https://forum.opnsense.org/index.php?topic=50985.0

### DIFF
--- a/src/opnsense/mvc/app/controllers/OPNsense/Wireguard/forms/dialogEditWireguardServer.xml
+++ b/src/opnsense/mvc/app/controllers/OPNsense/Wireguard/forms/dialogEditWireguardServer.xml
@@ -116,6 +116,46 @@
         </grid_view>
     </field>
     <field>
+        <id>server.preup</id>
+        <label>Pre-up commands</label>
+        <type>textbox</type>
+        <advanced>true</advanced>
+        <help>Commands to execute before setting up the interface. Specify each command on a new line. The special string %i is expanded to the interface name. Use with care.</help>
+        <grid_view>
+            <visible>false</visible>
+        </grid_view>
+    </field>
+    <field>
+        <id>server.postup</id>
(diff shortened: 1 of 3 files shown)


fichtner:
Well, first of all mwexec() doesn't exist on master anymore as per release/migration notes for 26.1 and in general this is against our basic security policy https://docs.opnsense.org/development/guidelines/basics.html#safeguard-user-input

Greelan:
I did wonder whether guardrails would be wanted on this...

Even with the "full admin" restriction being applied like the other legacy components, this would not be acceptable?

In that case, what options are there for implementing the commands functionality of wg-quick in OPNsense? Do users simply have to roll their own solution outside OPNsense?

Greelan:
I've implemented a solution via devd, but I was sorta hoping for a more user-friendly option xD

fichtner:
> Even with the "full admin" restriction being applied like the other legacy components, this would not be acceptable?

The full-admin thing is a stopgap measure to legacy OpenVPN due to a CVE (not entirely sure but it was equally grave) assigned to this particular fun subject. We avoid it for all new components.

> In that case, what options are there for implementing the commands functionality of wg-quick in OPNsense? Do users simply have to roll their own solution outside OPNsense?

What are you seeking to automate? Feature additions as well-scoped features generally make (more) sense since everyone can use and verify them.


Cheers,
Franco

Greelan:
> What are you seeking to automate? Feature additions as well-scoped features generally make (more) sense since everyone can use and verify them.

https://github.com/opnsense/core/pull/9825#issue-3961870128

fichtner:
Ok sorry I missed the link in the forum the first time I saw the thread. With a custom binary running this isn't a good base for a feature and the devd solution is clever.


Cheers,
Franco

PR #9829, opened 2026-02-19
openvpn: Adding options for legacy ciphers

jakobsen-lrz:
This adds some ciphers as options for OpenVPN Servers, and groups the options into recommended and legacy. It adds 128/192/256 Bit Versions of AES-x-CBC/CFB/CFB1/CFB8/OFB under Legacy, and AES-192-GCM under Recommended.

All of these Ciphers are still supported by the OpenVPN Server on OPNsense.

Note: Even though AES-256-CBC was added previously in commit 62d0c4e, it was added to the Legacy group. As according to the commit message, it was added for legacy compatibility.

I tested these changes on my OPNsense Instance and they all seem to work fine.


AdSchellevis:
@jakobsen-lrz thanks!

PR #9833, opened 2026-02-20
monitor: skip state killing when no viable failover target exists

internetsecureORG:
**Important notices**

- [X] I am convinced that my issue is new after having checked both open and closed issues at https://github.com/opnsense/core/issues?q=is:issue

**Describe the change**

When a gateway goes down, the 20-recover monitor hook kills all states for that gateway — even when no other gateway is available to failover to. When the only other gateway and its interface are both disabled, this destroys active connections on the only working gateway with zero benefit, causing self-inflicted outages.

The system already acknowledges this situation in its own logs:

    ROUTING: ignoring down gateways: WAN_DHCP, USB5GPHONE_DHCP

Yet it still kills states for WAN_DHCP immediately after.

**The fix**

Two lines added before the existing loop to pre-scan `return_gateways_status()` for any non-down gateway. One `if` wrapper around the existing kill call to skip it when no viable target exists. No new functions, no new imports, no structural changes.

When states are preserved, the decision is logged:

    ROUTING: preserving states for WAN_DHCP (no viable failover target)

When a viable target does exist, behavior is completely unchanged.

**Ref:** #9789 (closed), previous PR #9823 (closed — this is a simplified resubmission with minimal changes)

### DIFF
--- a/src/etc/rc.syshook.d/monitor/20-recover
+++ b/src/etc/rc.syshook.d/monitor/20-recover
@@ -37,8 +37,10 @@ $affected_gateways = !empty($argv[1]) ? explode(',', $argv[1]) : [];
 
 $metered_found_prios = ['inet' => 256, 'inet6' => 256];
 $metered_gws = ['inet' => [], 'inet6' => []];
+$all_statuses = return_gateways_status();
+$has_viable_target = !empty(array_filter($all_statuses, function ($s) { return strpos($s['status'], 'down') === false; }));
 
-foreach (return_gateways_status() as $status) {
+foreach ($all_statuses as $status) {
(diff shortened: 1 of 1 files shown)


fichtner:
The code itself is snappy now. Nice job.

In the grand scheme of things I don't know if this is wise as it prevents configured state killing in case the gateway really goes down and connections get stuck over an old gateway. But that was already mentioned.

Spent a number of years cleaning up the code removing side effects. Feels like a step in the wrong direction to me especially since you stated your gateway is technically considered down by the default settings you did not want to change in #9789 as suggested.


Cheers,
Franco

internetsecureORG:
You're right, close/cancel this. I will regroup and see if I can come up with anything else.

Issue #9816, opened 2026-02-18
mvc: BaseListField: shared implementation of $internalStaticOptionList


fichtner:
**Important notices**

Our forum is located at https://forum.opnsense.org , please consider joining discussions there in stead of using GitHub for these matters.

Before you ask a new question, we ask you kindly to acknowledge the following:

- [x] I have read the contributing guide lines at https://github.com/opnsense/core/blob/master/CONTRIBUTING.md
- [x] I am convinced that my issue is new after having checked both open and closed issues at https://github.com/opnsense/core/issues?q=is%3Aissue

The tests nowadays run decoupled from the normal system configuration in /conf which means that some tests may launch classes for e.g. field types inside PHPUnit's lifetime before a configuration file was given. The caching inside $internalStaticOptionList is strong so that it must be purged using `resetStaticOptionList()` but this needs to be added to every file type because BaseListField doesn't know about `$internalStaticOptionList`.

Moving will be a bit noisy but can probably remove quite a bit of code duplication. The tests should be able to catch refactoring issues too so it's worth giving it a try.

Relevant commits are 870692c154f and 104d5f02f89.

AdSchellevis:
@fichtner small POC to fix the subject https://github.com/opnsense/core/commit/64dbf01fc23c2842317b3ae96e803afd4df42253

PR #9839, opened 2026-02-21
mvc: BaseListField: shared implementation of $internalStaticOptionList

AdSchellevis:
Proof of concept for https://github.com/opnsense/core/issues/9816

Wrap static access in protected functions which ensures content is static per inherited class:

hasStaticOptions()
getStaticOptions()
setStaticOptions(array)
resetStaticOptions()

### DIFF
--- a/src/opnsense/mvc/app/models/OPNsense/Base/FieldTypes/AuthGroupField.php
+++ b/src/opnsense/mvc/app/models/OPNsense/Base/FieldTypes/AuthGroupField.php
@@ -1,7 +1,7 @@
 <?php
 
 /*
- * Copyright (C) 2015-2019 Deciso B.V.
+ * Copyright (C) 2015-2026 Deciso B.V.
  * All rights reserved.
  *
  * Redistribution and use in source and binary forms, with or without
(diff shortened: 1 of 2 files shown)


AdSchellevis:
@fichtner ok, merged, thanks for having a look. When I have a bit of time, I'll refactor the others too. This might also be a better pattern for `$internalStaticChildren` in `ArrayField` field as well as it's less fragile than trusting the user to overwrite the protected value. I also need to double check `$internalStaticOptList`, forgot before pressing merge, I think a private is more suitable here.

fichtner:
Wondered a bit about the `[static::class]` if it's protected every class should have its own instance or does it not? But easy to test with two conversions :)

AdSchellevis:
> Wondered a bit about the `[static::class]` if it's protected every class should have its own instance or does it not? But easy to test with two conversions :)

yep `static::class` is the actual classname (of the descendant)

fichtner:
I see but if its scope is the descendant there may only ever one entry there (protected vs. private) but I don't actually know. I have test data in a bit.

fichtner:
Ok statics are weird, confirmed via

```
array(3) {
  ["OPNsense\Base\FieldTypes\InterfaceField"]=>
  array(3) {
    ["*"]=>
    array(3) {
      ["lan"]=>
      string(3) "LAN"
      ["opt1"]=>
      string(4) "OPT1"
      ["wan"]=>
      string(3) "WAN"
    }
    ["eafbc20e60b92337fa1789a8b8dc3440"]=>
    array(6) {
      ["lan"]=>
      string(3) "LAN"
      ["opt1"]=>
      string(4) "OPT1"
      ["wan"]=>
      string(3) "WAN"
      ["if004"]=>
      string(7) "[if004]"
      ["if005"]=>
      string(7) "[if005]"
      ["if006"]=>
      string(7) "[if006]"
    }
    ["af1befeb1c240cf471da494dac60e1d6"]=>
    array(2) {
      ["lan"]=>
      string(3) "LAN"
      ["opt1"]=>
      string(4) "OPT1"
    }
  }
  ["OPNsense\Base\FieldTypes\AuthGroupField"]=>
  array(3) {
    [100]=>
    string(7) "group 1"
    [101]=>
    string(7) "group 2"
    [102]=>
    string(7) "group 3"
  }
  ["OPNsense\Base\FieldTypes\CertificateField"]=>
  array(2) {
    ["cert"]=>
    array(2) {
      ["5537f364ba123"]=>
      string(6) "cert 1"
      ["55c382ee6020f"]=>
      string(6) "cert 2"
    }
    ["ca"]=>
    array(2) {
      ["1537f364ba123"]=>
      string(4) "ca 1"
      ["15c382ee6020f"]=>
      string(4) "ca 2"
    }
  }
}
```

PR #9843, opened 2026-02-22
VPN: OpenVPN: Instances - debounce learn-address calls

AdSchellevis:
To help us limit the number of alias updates to a minimum.

When a lot of clients connect at the exact same time, it doesn't really make sense to keep calling the alias update procedure as logically we are only interested in the last one. This change simply opens a file, keeps track of its modification stamp and if another updated it after us, assume that caller is responsible for making the magic happen.

Although there is a small downside to this (clients connecting constantly, nobody responsible for accounting), the gain is likely much bigger as it prevents these calls from floodin the system in all cases (saving a lot of load).

There seems to be a possibility to trigger an excessive amount of calls to learn-address, but during my tests I have never seen connects from one client triggering updates for another. If this would happen, the number of calls would grow substantially. 

@swhite2 @Monviech

### DIFF
--- a/src/opnsense/scripts/openvpn/ovpn_event.py
+++ b/src/opnsense/scripts/openvpn/ovpn_event.py
@@ -1,7 +1,7 @@
 #!/usr/local/bin/python3
 
 """
-    Copyright (c) 2023 Ad Schellevis <ad@opnsense.org>
+    Copyright (c) 2023-2026 Ad Schellevis <ad@opnsense.org>
     All rights reserved.
 
     Redistribution and use in source and binary forms, with or without
@@ -31,9 +31,13 @@ import argparse
 import sys
 import os
 import subprocess
+import syslog
+import time
+import pathlib
 
 
 def main(params):
+    syslog.openlog('openvpn', facility=syslog.LOG_AUTH)
     if params.common_name:
         os.environ['common_name'] = params.common_name
     if params.auth_control_file:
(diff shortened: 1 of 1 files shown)


src/opnsense/scripts/openvpn/ovpn_event.py
```diff
@@ -1,7 +1,7 @@
 #!/usr/local/bin/python3
 
 """
-    Copyright (c) 2023 Ad Schellevis <ad@opnsense.org>
+    Copyright (c) 2023-2025 Ad Schellevis <ad@opnsense.org>
```
fichtner: Almost the right year 😉

AdSchellevis: @fichtner time travel ;) stupid me.

src/opnsense/scripts/openvpn/ovpn_event.py
```diff
@@ -67,10 +71,19 @@ def main(params):
...
     elif params.script_type == 'learn-address':
-        if os.fork() == 0:
+        # Highly simplified debouncer.
+        # Gathers events for 2 seconds, if more are triggered in the same slot, execute the last one.
+        # This moves events until we have at least 2 seconds of "silence" to process them
+        debounce_ref = pathlib.Path('/tmp/tmp_ovpn_update_tables.tmp')
+        debounce_ref.touch()
+        dbounce_start = debounce_ref.stat().st_mtime
+        time.sleep(2)
+        if debounce_ref.stat().st_mtime == dbounce_start and os.fork() == 0:
```
swhite2: `os.fork() == 0` should be in its own `if` statement, the `else` is always executed since we split into two processes.

src/opnsense/scripts/openvpn/ovpn_event.py
```diff
@@ -67,10 +71,19 @@ def main(params):
...
         sys.exit(subprocess.run("%s/client_disconnect.sh" % cmd_path).returncode)
     elif params.script_type == 'learn-address':
-        if os.fork() == 0:
+        # Highly simplified debouncer.
+        # Gathers events for 2 seconds, if more are triggered in the same slot, execute the last one.
+        # This moves events until we have at least 2 seconds of "silence" to process them
+        debounce_ref = pathlib.Path('/tmp/tmp_ovpn_update_tables.tmp')
+        debounce_ref.touch()
+        dbounce_start = debounce_ref.stat().st_mtime
+        time.sleep(2)
```
swhite2: After some testing, OpenVPN seems to seralize this. Meaning: if a client connects, openvpn sleeps for 2 seconds. If a different client connects in the meantime, this `ovpn_event.py` will only trigger after the sleep.

The debounce should happen after the fork

swhite2 (changes requested):

AdSchellevis:
@swhite2 merged in https://github.com/opnsense/core/commit/d74dc5acc14c402b482842b588c267acc365bd00, which moves the debouncer inside the fork.

PR #9847, opened 2026-02-23
Trim whitespace from tokens before tokenising

### LINKED ISSUE #9438: Could you please add .trim() to the FW alias network/host tokenizer?
**Important notices**

- [x ] I have read the contributing guide lines at https://github.com/opnsense/core/blob/master/CONTRIBUTING.md
- [x ] I am convinced that my issue is new after having checked both open and closed issues at https://github.com/opnsense/core/issues?q=is%3Aissue

The tokenizer is 'dumb' in a sense that it tokenizes any input even if it includes tabs/spaces but later throws validation errors on save.

If the input is trimmed before it is tokenized, user experience will improve because we will no longer have to go back, search for invalid tokens, delete them, and re-add them w/o blank spaces.

N/A

**Additional context**

N/A

mintsoft:
Previously if you pasted an IP or similar in whitespace on either side it will automatically put it into a token box and then error on save.

The only way to fix that is to paste into an editor, fix the data and then copy-paste back into opnsense. This should trim before tokenizing the data

Fixes: https://github.com/opnsense/core/issues/9438

### DIFF
--- a/src/opnsense/www/js/tokenize2.js
+++ b/src/opnsense/www/js/tokenize2.js
@@ -282,6 +282,8 @@
      */
     Tokenize2.prototype.tokenAdd = function(value, text, force){
 
+        text = text.trim();
+        value = value.trim();
         text = text || value;
         selector_value = $.escapeSelector(value);
         force = force || false;


fichtner:
It's a good idea on the surface, but it doesn't fix much and leads to other problems. My favourite example is that the code used to trim a password for no good reasons. It's asking for trouble later than just making sure the pasted data is clean.


Cheers,
Franco

mintsoft:
Hmm, you make a good point regarding the password and I understand your concerns however my reading of the code is that this won't affect that anyway, it's only in the situation where the tokenizer is hooked into the list (which seems to be multi-select only).

WRT to significance, I honestly couldn't disagree more strongly.  This is an absolutely infuriating experience for any experienced person using opnsense. It's a super common problem to have when you're setting up VPNs to copy and paste out of browsers which frequently will crowbar whitespace before and after values and there is no way to resolve it in-ui because the tokenizer kicks-in before you're able to fix it. 

An alternative place to do this (which is more targeted, https://github.com/opnsense/core/pull/9847/changes#diff-31373581f35eaa592e76f126a2d240ffdd432a076ec95cb5d8ae1f047716edceR321) would be in the last step when it wraps the values in the `li`. At that point it is about to write the invalid data into the DOM, so it is inevitable that the next interaction that the user has with the box is to delete the value when the save validation fails.

fichtner:
No changes to current tokenizer planned. It's in need of a replacement. Closing this.

Issue #9818, opened 2026-02-18
Hardcoded color in firewall (new rules) should be deleted


opnsenseuser:
<img width="438" height="331" alt="Image" src="https://github.com/user-attachments/assets/871760e1-1eff-4f10-860c-c3fc52d3ec03" />

<img width="543" height="175" alt="Image" src="https://github.com/user-attachments/assets/91cec504-df1f-4257-aa06-6038f2758709" />

fa fa-fw fa-tag has already a color and has a global impact!
so no need for this hardcoded color.

regards
rené

Monviech:
A category needs a color so this is a fallback if no color has been set.

https://github.com/opnsense/core/blob/36b17adf3155181ba4b09e39ba2b15635a2f7d31/src/opnsense/mvc/app/models/OPNsense/Firewall/Category.xml#L18

https://github.com/opnsense/core/blob/36b17adf3155181ba4b09e39ba2b15635a2f7d31/src/opnsense/mvc/app/controllers/OPNsense/Firewall/Api/FilterBaseController.php#L65

In this case the automatic rules also set the color here:

https://github.com/opnsense/core/blob/36b17adf3155181ba4b09e39ba2b15635a2f7d31/src/opnsense/mvc/app/controllers/OPNsense/Firewall/Api/FilterController.php#L184

opnsenseuser:
@Monviech

You're right! But wouldn't it be better to hide the icon completely? For two reasons. First, you can create your own black categories, and then you wouldn't be able to distinguish your own category from the automated ones. Second, if you're using dark themes like Opnsense Dark, etc., readability isn't great anymore. Is hiding the icon completely an option for you? Thanks, René

Monviech:
There are a few more hardcoded colors in the firewall template that have be pushed to a style sheet. I have to think about it, Im sure we can find some solution.

Monviech:
@opnsenseuser I did a larger cleanup and I think all hardcoded colors are gone now. Can you check if this is sufficient for your theming work?

opnsenseuser:
@Monviech From what I can see in the code, it should work. Thanks.

PR #9848, opened 2026-02-23
Firewall: Rules [new]: Remove hardcoded colors where possible

Monviech:
Fixes: https://github.com/opnsense/core/issues/9818

- [x] Turn hardcoded badge colors into inherited label colors, use chip class for badges in tabulator rows
- [x] Remove hardcoded fallback category colors
- [x] Adjust the NAT templates with the same category logic
- [x] Lets first unify the NAT templates so that the same logic doesn't have to be changed at different spots: https://github.com/opnsense/core/issues/9849


Monviech:
@fichtner This should clean up all color issues, by using the ``label`` class and changing its shape to a pill via css, I avoid having to change any css files as I would have to when continuing to use ``badge``. That's better since there is no rat tail here with adjusting themes.

fichtner (approved):

PR #9860, opened 2026-02-25
Keav4: Allow setting option 43 data for subnets

Holladiewal:
This idea is based on #7361, and I am aware of concerns and issues regarding complexity and client-classes raised in that PR.

However I think offering the ability to set option 43 data is still important, even if there is no hand-holding by the UI to get the right values and not caring about client-classes.
If a client wants some option 43 data, there is only on set of data to choose from _per subnet_.

This PR allows just that: Setting the value delivered to the clients.
Nothing more, nothing less.

This could be expanded by offering help encoding values (for example encoding an ASCII-String into a hex-string).

Below is a picture of config and outcome (I had to set "always-send" so ubuntu would get the option. That should probably be exposed as a checkbox to the user):
grafik

### DIFF
--- a/src/opnsense/mvc/app/controllers/OPNsense/Kea/forms/dialogReservation4.xml
+++ b/src/opnsense/mvc/app/controllers/OPNsense/Kea/forms/dialogReservation4.xml
@@ -143,4 +143,14 @@
             <visible>false</visible>
         </grid_view>
     </field>
+    <field>
+        <id>reservation.option_data.vendor_encapsulated_options</id>
+        <label>Raw Option 43 data</label>
+        <type>text</type>
+        <help>Raw TLV-Data for option 43. This will be delivered to the client as is. Specify as hex value without leading "0x"</help>
+        <grid_view>
+            <visible>false</visible>
+        </grid_view>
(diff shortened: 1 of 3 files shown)


AdSchellevis:
Preferably I need some people to test this before merging, @Holladiewal can you ask people in the other PR to test yours? I'll keep this open for a bit to collect feedback.

Holladiewal:
Had the chance to do some testing of my own:

Contrary to most things, Kea actually seems to be quite content with getting some malformed data in its config.

Options that are too short are not sent:
option-too-short

Options that are longer than specified are truncated to the proper length:
option-too-long

Options specified out of order are being reordered (config order is "f1, f3, f2" and options are sent as "f1,f2,f3"):
two-options-reordered

So far, I haven't found a way to make Kea not start up based on the user input.
It might not work as expected if the operator makes a mistake, but at least the service doesn't fall over and refuse to work.

Monviech:
@Holladiewal The hint about `csv-format false` is very good. I think this could be generalized for all options.

We might be able to solve the whole subset of problems with such an approach, but since KEA is weird we also might not. But I will investigate this here:

PR #9862, opened 2026-02-25
Small NTP fix

sopex:
Fixes: https://github.com/opnsense/core/issues/6923

This commit fixes a longstanding issue while minimizing the extra code to be converted to MVC.

### DIFF
--- a/src/etc/config.xml.sample
+++ b/src/etc/config.xml.sample
@@ -136,6 +136,6 @@
   </rrd>
   <ntpd>
     <prefer>0.opnsense.pool.ntp.org</prefer>
-    <ispool>0.opnsense.pool.ntp.org</ispool>
+    <ispool>0.opnsense.pool.ntp.org 1.opnsense.pool.ntp.org 2.opnsense.pool.ntp.org 3.opnsense.pool.ntp.org</ispool> 
   </ntpd>
 </opnsense>


src/www/services_ntpd.php
```diff
@@ -281,6 +285,9 @@ function removeRow() {
                             <td>
                               <input name="timeservers_prefer[]" class="ts_checkbox" type="checkbox" value="<?=$timeserver;?>" <?= !empty($pconfig['timeservers_prefer']) && in_array($timeserver, $pconfig['timeservers_prefer']) ? 'checked="checked"' : '' ?>/>
                             </td>
+                            <td>
+                              <input name="timeservers_ispool[]" class="ts_checkbox" type="checkbox" value="<?=$timeserver;?>" <?= !empty($pconfig['timeservers_ispool']) && in_array($timeserver, $pconfig['timeservers_ispool']) ? 'checked="checked"' : '' ?>/>
```
fichtner, line 289: if you mark them as checked here this will render in the config... which is mostly ok given the constraints but then it should also reflect that reality in the src/etc/config.xml.sample file

src/www/services_ntpd.php
```diff
@@ -305,6 +312,8 @@ function removeRow() {
                         <br />
                         <?= gettext('The "prefer" option indicates that NTP should favor the use of this server more than all others.') ?>
                         <br />
+                        <?= gettext('The "pool" option marks this entry as a pool hostname, allowing NTP to resolve and use multiple servers from it. Entries with ".pool.ntp.org" syntax, are automatically resolved as pools') ?>
```
fichtner:
```suggestion
                        <?= gettext('The "pool" option marks this entry as a pool hostname, allowing NTP to resolve and use multiple servers from it.') ?>
```

sooner or later the automatic handling will disappear. It doesn't make sense to document the oddity.

sopex: Sure

src/etc/config.xml.sample
```diff
@@ -136,5 +136,6 @@
   </rrd>
   <ntpd>
     <prefer>0.opnsense.pool.ntp.org</prefer>
+    <ispool>0.opnsense.pool.ntp.org</ispool>
```
fichtner:
```suggestion
    <ispool>0.opnsense.pool.ntp.org 1.opnsense.pool.ntp.org 2.opnsense.pool.ntp.org 3.opnsense.pool.ntp.org</ispool>
```

it's all of them :)

fichtner:
Merged, thanks!

Issue #9842, opened 2026-02-22
[Firewall:Rules [new]]: setting 'reply-to' on floating rule with VPN gateway causes syntax error


dstapa:
**Important notices**

Before you add a new report, we ask you kindly to acknowledge the following:

- [x] I have read the contributing guide lines at https://github.com/opnsense/core/blob/master/CONTRIBUTING.md
- [x] I am convinced that my issue is new after having checked both open and closed issues at https://github.com/opnsense/core/issues?q=is%3Aissue

**Describe the bug**

I have a few "force gw" type rules to route firewall traffic from the VPN interface to the appropriate VPN gateway.  Until now those did not have an explicit 'reply-to' option set.  I tried to enable it and got a syntax error in my pf ruleset.

**To Reproduce**

1. Create a VPN interface and a corresponding VPN gateway to a provider (e.g. Mullvad)
2. Create a floating rule to force internet traffic from the VPN interface source IP to the VPN gateway.
3. Observe everything working up to this point.
4. Edit the floating rule and enable 'reply-to' with the VPN gateway IP in advanced options.
5. Observe syntax error in OPNsense UI after saving.

**Screenshots**

<img width="601" height="290" alt="Image" src="https://github.com/user-attachments/assets/7bc1ba77-14c4-4bcf-9f9a-e19f91056ebb" />

<img width="1945" height="174" alt="Image" src="https://github.com/user-attachments/assets/7f6cef3d-168f-4042-91dc-513b6508b92c" />

**Relevant log files**

Diff of `/tmp/rules.debug` before and after applying the 'reply-to' (addresses masked) below.

Line number doesn't match screenshot because I reproduced it a second time after some other rule changes to capture the log.

```
root@firewall:/tmp # diff rules.debug.old rules.debug
692c692
< pass out log route-to ( wg1 10.66.xxx.xx ) inet from {(wg1)} to {!(wg1:network)} keep state label "b3a58ac6-3e14-402b-a320-2c9077d9250b" # Route WAN_VPN0 interface traffic through VPN (force gw)
---
> pass out log route-to ( wg1 10.66.xxx.xx ) reply-to ( wg1 10.66.xxx.xx ) inet from {(wg1)} to {!(wg1:network)} keep state label "b3a58ac6-3e14-402b-a320-2c9077d9250b" # Route WAN_VPN0 interface traffic through VPN (force gw)
```

**Environment**

OPNsense 26.1.2_5 (amd64)

Monviech:
[...]

PR #9863, opened 2026-02-26
mvc: Add validation for using both gateway and reply-to in rules [new]

TomWalraven:
Added validation to rules [new] to ensure a user can't select both a gateway and reply-to destination when adding/editing a firewall rule.

Fixes: #9842

### DIFF
--- a/src/opnsense/mvc/app/models/OPNsense/Firewall/Filter.php
+++ b/src/opnsense/mvc/app/models/OPNsense/Firewall/Filter.php
@@ -267,10 +267,10 @@ class Filter extends BaseModel
                                 $rule->{'divert-to'}->__reference
                             ));
                         }
-                        if (!$rule->{'gateway'}->isEmpty() && !$rule->{'replyto'}->isEmpty()) {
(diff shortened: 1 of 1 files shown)


src/opnsense/mvc/app/models/OPNsense/Firewall/Filter.php
```diff
@@ -267,6 +267,12 @@ public function performValidation($validateFullModel = false)
                                 $rule->{'divert-to'}->__reference
                             ));
                         }
+                        if (!$rule->{'gateway'}->isEmpty() && !$rule->{'replyto'}->isEmpty()) {
```
Monviech: I think the curly braces are not required here are there are no special characters.


```suggestion
                        if (!$rule->gateway->isEmpty() && !$rule->replyto->isEmpty()) {
```

Same below.

I think looks good already otherwise, nice :)

TomWalraven: Thanks for the suggestion! I've removed the curly braces in line 270 and 273.

Monviech:
@TomWalraven Thank you, works just fine in my tests :)

Issue #9850, opened 2026-02-23
OpenVPN: Username as CN depends on authmode being set


swhite2:
**Important notices**

Before you add a new report, we ask you kindly to acknowledge the following:

- [x] I have read the contributing guide lines at https://github.com/opnsense/core/blob/master/CONTRIBUTING.md
- [x] I am convinced that my issue is new after having checked both open and closed issues at https://github.com/opnsense/core/issues?q=is%3Aissue

**Describe the bug**

If enabling `Username as CN` while leaving `Authentication` empty, OpenVPN will refuse to start, since there is no authentication, there will also not be a username to match on.

**To Reproduce**

1. Create an OpenVPN instance
2. Enable `Username as CN`
3. Do not set any `Authentication`
4. Start OpenVPN, see error in log:

```
--username-as-common-name must be used with --management-client-auth, an --auth-user-pass-verify script, or plugin
```
Core only uses `auth-user-pass-verify`.

**Expected behavior**

Validation error if no `Authentication` set while `Username as CN` is enabled.

**Describe alternatives you considered**

N/A

**Screenshots**

N/A

**Relevant log files**

N/A

**Additional context**

N/A

**Environment**

OPNsense 26.7.a_180 (amd64).

PR #9878, opened 2026-03-02
mvc: Add validation for selecting username as CN without setting any authentication

TomWalraven:
Added validation to openvpn to ensure a user can't select username as CN without setting any authentication when adding/editing an openvpn instance.

Fixes: #9850

### DIFF
--- a/src/opnsense/mvc/app/models/OPNsense/OpenVPN/OpenVPN.php
+++ b/src/opnsense/mvc/app/models/OPNsense/OpenVPN/OpenVPN.php
@@ -135,7 +135,7 @@ class OpenVPN extends BaseModel
                 }
                 if (!$instance->username_as_common_name->isEmpty() && $instance->authmode->isEmpty()) {
                     $messages->appendMessage(new Message(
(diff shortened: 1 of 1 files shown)


src/opnsense/mvc/app/models/OPNsense/OpenVPN/OpenVPN.php
```diff
@@ -133,6 +133,12 @@ public function performValidation($validateFullModel = false)
                         $key . ".verify_client_cert"
                     ));
                 }
+                if (!$instance->username_as_common_name->isEmpty() && $instance->authmode->isEmpty()) {
+                    $messages->appendMessage(new Message(
+                        gettext('Username as CN requires an authentication mode.'),
```
swhite2: I'd go for plural here since multiple auth modes can be set, e.g. `requires one or more authentication modes`.

TomWalraven: The message has been updated, thanks!

swhite2:
Merged, thanks :)

PR #9880, opened 2026-03-02
Firewall: NAT: Destination NAT: Allow well known ports in local-port

Monviech:
Seems to work for me:

```
root@opn-dev-02:/src/git/core # pfctl -s nat

rdr inet proto tcp from 127.0.0.1 port = https to 127.0.0.1 port = ftp -> 127.0.0.1 port 993
```

image


fichtner:
either way is fine for now

Monviech:
@fichtner Your call, I'm fine with either as well.

Issue #9879, opened 2026-03-02
Firewall: NAT: Destination NAT: Missing validation for TCP/UDP + port


Monviech:
**Important notices**

Before you add a new report, we ask you kindly to acknowledge the following:

- [x] I have read the contributing guide lines at https://github.com/opnsense/core/blob/master/CONTRIBUTING.md
- [x] I am convinced that my issue is new after having checked both open and closed issues at https://github.com/opnsense/core/issues?q=is%3Aissue

**Is your feature request related to a problem? Please describe.**

in DNAT, setting a port should require a protocol like TCP, UDP or TCP/UDP.

Right now it's possible to save without setting a protocol but setting ports.

**Describe the solution you like**

Same as here:
https://github.com/opnsense/core/blob/b165e23001beec365cb61b8a3c6f81dfe5e4ac66/src/opnsense/mvc/app/models/OPNsense/Firewall/Filter.php#L150-L165

Since most validations that we want are inside the Filter.php model, and we have a DNAT model, I'm not sure how to continue here without duplicating these validations.

https://github.com/opnsense/core/blob/master/src/opnsense/mvc/app/models/OPNsense/Firewall/DNat.php

**Describe alternatives you considered**

None

**Additional context**

None

Monviech:
Tagging @AdSchellevis for hints how to solve this.

Monviech:
I looked into creating a trait for the NAT rules validation, while technically possible it's introducing complexity because the models of SNAT and DNAT have different data structures in some spots. So it would need helpers that normalize them inside the trait. For around 4 needed validations it's easier and better maintainable to just duplicate them into the DNAT model. Thats the approach I will take.

PR #9885, opened 2026-03-03
Firewall: NAT: Destination NAT: Add model validations for common errors

Monviech:
Fixes: https://github.com/opnsense/core/issues/9879

### DIFF
--- a/src/opnsense/mvc/app/models/OPNsense/Firewall/DNat.php
+++ b/src/opnsense/mvc/app/models/OPNsense/Firewall/DNat.php
@@ -45,56 +45,27 @@ class DNat extends BaseModel
 
         foreach ($this->rule->iterateItems() as $rule) {
             if ($validateFullModel || $rule->isFieldChanged()) {
-
-                if (!empty($rule->source->port->getValue()) && !in_array($rule->protocol->getValue(), $port_protos)) {
-                    $messages->appendMessage(new Message(
-                        gettext("Ports are only valid for TCP or UDP type rules."),
-                        $rule->source->port->__reference
-                    ));
+                $ports = [$rule->source->port, $rule->destination->port, $rule->{'local-port'}];
+                foreach ($ports as $port) {
+                    if (!empty($port->getValue()) && !in_array($rule->protocol->getValue(), $port_protos)) {
+                        $messages->appendMessage(new Message(
+                            gettext("Ports are only valid for TCP or UDP type rules."),
+                            $port->__reference
+                        ));
+                    }
                 }
 
-                if (!empty($rule->destination->port->getValue()) && !in_array($rule->protocol->getValue(), $port_protos)) {
-                    $messages->appendMessage(new Message(
-                        gettext("Ports are only valid for TCP or UDP type rules."),
-                        $rule->destination->port->__reference
-                    ));
-                }
+                $addresses = [$rule->source->network, $rule->destination->network, $rule->target];
+                foreach ($addresses as $address) {
+                    $is_addr = Util::isSubnet($address->getValue()) || Util::isIpAddress($address->getValue());
+                    $proto = strpos($address->getValue(), ':') === false ? "inet" : "inet6";
 
-                if (!empty($rule->{'local-port'}->getValue()) && !in_array($rule->protocol->getValue(), $port_protos)) {
-                    $messages->appendMessage(new Message(
-                        gettext("Ports are only valid for TCP or UDP type rules."),
-                        $rule->{'local-port'}->__reference
-                    ));
-                }
-
-                $src_is_addr = Util::isSubnet($rule->source->network->getValue()) || Util::isIpAddress($rule->source->network->getValue());
-                $src_proto = strpos($rule->source->network->getValue(), ':') === false ? "inet" : "inet6";
-
-                if ($src_is_addr && in_array($rule->ipprotocol->getValue(), ['inet', 'inet6']) && !$rule->ipprotocol->isEqual($src_proto)) {
-                    $messages->appendMessage(new Message(
-                        gettext("Address type should match selected TCP/IP protocol version."),
-                        $rule->source->network->__reference
-                    ));
-                }
-
-                $dest_is_addr = Util::isSubnet($rule->destination->network->getValue()) || Util::isIpAddress($rule->destination->network->getValue());
-                $dest_proto = strpos($rule->destination->network->getValue(), ':') === false ? "inet" : "inet6";
-
-                if ($dest_is_addr && in_array($rule->ipprotocol->getValue(), ['inet', 'inet6']) && !$rule->ipprotocol->isEqual($dest_proto)) {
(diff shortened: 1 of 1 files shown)


src/opnsense/mvc/app/models/OPNsense/Firewall/DNat.php
```diff
@@ -29,11 +29,76 @@
...
+    {
+        $messages = parent::performValidation($validateFullModel);
+        $port_protos = ['tcp', 'udp', 'tcp/udp'];
+
+        foreach ($this->rule->iterateItems() as $rule) {
+            if ($validateFullModel || $rule->isFieldChanged()) {
+
+                if (!empty($rule->source->port->getValue()) && !in_array($rule->protocol->getValue(), $port_protos)) {
+                    $messages->appendMessage(new Message(
+                        gettext("Source ports are only valid for tcp or udp type rules."),
```
fichtner:
```suggestion
                        gettext("Ports are only valid for TCP or UDP type rules."),
```

Try to avoid "source" et al in strings too, so these can be reused below...

src/opnsense/mvc/app/models/OPNsense/Firewall/DNat.php
```diff
@@ -29,11 +29,76 @@
...
+                    $messages->appendMessage(new Message(
+                        gettext("Target ports are only valid for tcp or udp type rules."),
+                        $rule->{'local-port'}->__reference
+                    ));
+                }
+
+                $src_is_addr = Util::isSubnet($rule->source->network->getValue()) || Util::isIpAddress($rule->source->network->getValue());
+                $src_proto = strpos($rule->source->network->getValue(), ':') === false ? "inet" : "inet6";
+
+                if ($src_is_addr && in_array($rule->ipprotocol->getValue(), ['inet', 'inet6']) && !$rule->ipprotocol->isEqual($src_proto)) {
```
fichtner: perhaps this is overzealous. it depends on whether pf.conf accepts mismatches or not

Monviech: If we doubt it here we should doubt it here too:

https://github.com/opnsense/core/blob/29ac6a0ca2b2940de47ee3cce5c9b455125ba618/src/opnsense/mvc/app/models/OPNsense/Firewall/Filter.php#L65-L81

src/opnsense/mvc/app/models/OPNsense/Firewall/DNat.php
```diff
@@ -29,11 +29,76 @@
...
+                    $messages->appendMessage(new Message(
+                        gettext("Destination address type should match selected TCP/IP protocol version."),
+                        $rule->destination->network->__reference
+                    ));
+                }
+
+                $target_is_addr = Util::isSubnet($rule->target->getValue()) || Util::isIpAddress($rule->target->getValue());
+                $target_proto = strpos($rule->target->getValue(), ':') === false ? "inet" : "inet6";
+
+                if ($target_is_addr && in_array($rule->ipprotocol->getValue(), ['inet', 'inet6']) && !$rule->ipprotocol->isEqual($target_proto)) {
```
fichtner: I think you're also getting in trouble if ipprotocol is inet46 or empty...

Monviech: Could be, have to test this, thanks

Monviech: `""` and `inet46` get skipped here because its not included in_array(), so it will not fail validation

src/opnsense/mvc/app/models/OPNsense/Firewall/DNat.php
Monviech: That makes sense, fixed, thanks.

src/opnsense/mvc/app/models/OPNsense/Firewall/DNat.php
```diff
@@ -29,11 +29,76 @@
...
+                        $rule->destination->port->__reference
+                    ));
+                }
+
+                if (!empty($rule->{'local-port'}->getValue()) && !in_array($rule->protocol->getValue(), $port_protos)) {
+                    $messages->appendMessage(new Message(
+                        gettext("Ports are only valid for TCP or UDP type rules."),
+                        $rule->{'local-port'}->__reference
+                    ));
+                }
```
fichtner: I think you can put all of these in a loop now

src/opnsense/mvc/app/models/OPNsense/Firewall/DNat.php
fichtner: if it's the same everywhere that's ok but pair of eyes via @AdSchellevis wouldn't hurt

src/opnsense/mvc/app/models/OPNsense/Firewall/DNat.php
fichtner: 👍

src/opnsense/mvc/app/models/OPNsense/Firewall/DNat.php
Monviech: 2026-03-05T09:29:55ErrorfirewallThere were error(s) loading the rules: /tmp/rules.debug:96: no translation address with matching address family found. - The line in question reads [96]: rdr inet6 from {any} to {192.168.1.1} -> 192.168.1.2

src/opnsense/mvc/app/models/OPNsense/Firewall/DNat.php
Monviech: Good catch, done

src/opnsense/mvc/app/models/OPNsense/Firewall/DNat.php
```diff
@@ -29,11 +29,47 @@
...
  */
 class DNat extends BaseModel
 {
+    public function performValidation($validateFullModel = false)
+    {
+        $messages = parent::performValidation($validateFullModel);
+        $port_protos = ['tcp', 'udp', 'tcp/udp'];
+
+        foreach ($this->rule->iterateItems() as $rule) {
+            if ($validateFullModel || $rule->isFieldChanged()) {
```
fichtner, line 47: just a note here: isFieldChanged() checks if any field in the rule changed, not only the fields we're trying to validate

Monviech: That's the same everywhere though, right? I don't think you can do "rule->desiredfield->isFieldChanged()"?

fichtner: Yes you can, but it would be bit of overkill here because we can say that the rule was edited.

fichtner: maybe @AdSchellevis has a thought here too

PR #9887, opened 2026-03-03
Update Opnsense Sidebar

opnsenseuser:
**Fix:**

Upon loading, $(window).height() and $(window).width() are immediately checked – the same condition as in the resize handler (navHeight / 760px).

If the viewport is too small → the sidebar is hidden, mouse events are disabled, and the toggle button is hidden.
If the viewport is large enough → normal behavior as usual (including the localStorage preset).

**Readability & Structure**

var consistently replaced with const/let
Helpful functions like isSidebarLeft() and isSidebarHidden() extracted to avoid duplicate hasClass calls transition_duration → setTransitionDuration, mouse_events_off → offMouseEvents (clearer names)

**Performance**

jQuery selectors are cached only once (e.g., toggle_btn, allLayers, aLayers, divLayers)

$.each(document.styleSheets, ...) replaced with modern Array.from(...).some(...) – shorter and terminates early toggle_sidebar_loaded check with early return instead of deeply nested if block

**Redundancy Reduction**

Duplicate calls consolidated in event handlers Repeated offMouseEvents()/transition_duration() calls in Resize handler consolidates winWidth directly from $(window).width() instead of via win variable

@fichtner I've taken the sidebar to a whole new level. Tested a thousand times. Works flawlessly. It would be great if we could release this version with the next update, as it also fixes a problem that's currently present. See description. Thanks, Rene

### DIFF
--- a/src/opnsense/www/js/opnsense_theme.js
+++ b/src/opnsense/www/js/opnsense_theme.js
@@ -179,7 +179,7 @@ $(document).ready(function () {
     $(window).on('resize', function () {
         const winHeight = $(window).height();
         const winWidth  = $(window).width();
-        const tooSmall  = winHeight < navHeight || winWidth < 1000;
+        const tooSmall  = winHeight < navHeight || winWidth < 760;
 
         if (tooSmall && !isSidebarHidden()) {
             navigation.addClass('col-sidebar-hidden');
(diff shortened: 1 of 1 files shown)


opnsenseuser:
@fichtner I would ask you to revert the last two push commands where I increased the width to 1000, as this doesn't fix the problem. However, it would be important to merge the sidebar update before the next release, as it fixes a major display bug when switching from the standard menu to the sidebar. Thank you very much, best regards, Rene

fichtner:
can you fix your branch for this PR? I'd squash so there's nothing to revert.

opnsenseuser:
@fichtner ready. Changed it to the org

opnsenseuser:
@fichtner would be great if you can merge this. regards rené

PR #9896, opened 2026-03-05
Add HA/CARP safety for PPP link startup

### LINKED ISSUE #10663: CARP PPPoE bring-up on backup node startup, leaving it permanently redialing
**Important notices**

- [ x] I have read the contributing guidelines at https://github.com/opnsense/core/blob/master/CONTRIBUTING.md
- [ x] I am convinced that my issue is new after having checked both open and closed issues at https://github.com/opnsense/core/issues?q=is%3Aissue
- [ x] AI tools were used to create at least part of the text submitted herewith.

- Model used: Claude Sonnet 5
- Extent of AI involvement: Debugging assistance, text generation, log parsing, German->English translation

**Describe the bug**

I'm on opnsense 26.07 but I had problems with PPPoE failover for ~2 years, when I switched from double-NAT to PPPoE. Now Claude actually presented me with a workaround, which solves that. It would be beneficial for others to include the fix upstream.

In a two-node CARP HA cluster with a PPPoE WAN interface, "Disconnect dialup interfaces" (`disconnectppps`) is correctly enabled on both nodes, and the CARP virtual IP is correctly placed on the same interface the PPPoE link runs over (per the documented requirement for `/usr/local/etc/rc.syshook.d/carp/20-ppp` to work).

When the backup node boots, `20-ppp` fires as expected with the correct BACKUP status very early in the boot sequence — but at that point the PPP subsystem (`pppoe0`/`ng0`) has not been created yet, so the suspend is effectively a no-op (`/usr/local/sbin/ifinfo pppoe0` fails because the interface doesn't exist). Several seconds later, the regular boot-time interface bring-up (`rc.bootup` / `interface_ppps_configure()`) configures WAN/PPPoE unconditionally, with no regard for the current CARP state. Since CARP does not transition again after this point (it remains stable in BACKUP), `20-ppp` never re-fires, and the PPPoE session on the backup node is left permanently attempting to redial — indefinitely, until an administrator manually forces a CARP state transition (e.g. via "Temporarily Disable CARP" on the master, or a reboot of the master).

This results in continuous log spam ("PPPoE connection timeout after 9 seconds", thousands of reconnection attempts over hours/days) and an unauthenticated PPPoE session that never becomes active — which is mostly cosmetic/log-noise in this case since the ISP will only authenticate one session at a time, but is a functional violation of the "Disconnect dialup interfaces" contract and could plausibly cause session/auth conflicts with some ISPs depending on how they handle duplicate PPPoE discovery from the same line.

**To Reproduce**

1. Set up two OPNsense nodes in CARP HA with a PPPoE WAN interface, following the documented pattern: CARP VIP on the same interface (`vtnet0` in our case) that is selected as the PPP "Link interface(s)" for the PPPoE connection (assigned as `opt10`/WAN).
2. Enable "Disconnect dialup interfaces" under System → High Availability → Settings on **both** nodes (confirmed present in `/conf/config.xml` as `<disconnectppps>1</disconnectppps>` on both).
3. Reboot the backup node while the master remains active/MASTER.
4. Observe `/var/log/system/latest.log` on the backup node after boot.

**Expected behavior**

After boot, if the node's CARP status is BACKUP, the PPPoE interface should not be left in a continuous reconnect loop — either `20-ppp` should re-check/retry after `rc.bootup` completes, or `rc.bootup`/`interface_ppps_configure()` should itself respect the current CARP state for interfaces flagged with `disconnectppps`.

**Log evidence**
(backup node, `fwint2`, timestamps from `/var/log/system/latest.log`)

01:25:33 carp: 9@vtnet0: MASTER -> INIT
   (hardware interface down) # node begins shutdown/reboot
01:26:17 rc.syshook.d/carp/20-ppp: The command /usr/local/sbin/ifinfo 'pppoe0'' failed to execute       
   (repeated 5x — pppoe0 does not exist yet, this fires during early boot) 
01:31:03  vtnet0: <VirtIO Networking Adapter> on virtio_pci2
   (interface reattached post-boot) 
01:31:04  carp: 9@vtnet0: INIT -> BACKUP
   (initialization complete) 
01:31:04  rc.syshook.d/carp/20-ppp: vtnet0 is connected to ppp interface opt10 set new status BACKUP 
01:31:04  rc.syshook.d/carp/20-ppp: The command /usr/local/sbin/ifinfo 'pppoe0'' failed to execute
    (repeated 8x — pppoe0 still does not exist, suspend is a no-op)
01:31:25 rc.bootup: interface_ppps_configure() waiting threshold exceeded - device pppoe0 is still not up
   (regular boot-time bring-up starts dialing PPPoE unconditionally, ignoring CARP state)
01:31:29 setup vtnet0 [egress only]
01:31:43 DSLmodemManagement (vtnet0) -> v4: 192.168.200.11/24

From this point on, the node remains CARP BACKUP indefinitely (confirmed via `ifconfig | grep carp` hours later), yet `ppp.log` continues to show:

Notice · ppp · [opt10_link0] PPPoE connection timeout after 9 seconds
Notice · ppp · [opt10_link0] Link: reconnection attempt 2216

We confirmed this is purely a boot-timing issue and not a general problem with `20-ppp`/`ifinfo`/`interface_suspend()`: manually forcing a real CARP state transition later (via "Temporarily Disable CARP" on the master) correctly triggered `20-ppp`, and `interface_suspend('opt10')` succeeded, tearing down `pppoe0` cleanly on the (now) backup node. So the underlying suspend logic works correctly — it's specifically the boot sequence where `20-ppp` fires too early (before `pppoe0` exists) and is never retried once `rc.bootup` unconditionally brings the PPPoE interface up afterward.

**Environment**

- OPNsense 26.7 (FreeBSD 15.1)
- 2x OPNsense VMs (KVM/Proxmox, `vtnet0` virtio NIC) in CARP HA
- WAN: PPPoE over VLAN-tagged transfer network, CARP VIP on the same underlying interface (`vtnet0`) as required for `20-ppp` to function
- `disconnectppps` confirmed enabled in `/conf/config.xml` on both nodes

**Workaround in use**

We added a custom syshook script (`/usr/local/etc/rc.syshook.d/start/95-ppp-carp-recheck`) that re-checks the CARP state of the WAN parent interface after boot completes and calls `interface_suspend()` again if still BACKUP. This resolves the symptom but obviously isn't a proper fix, since it duplicates logic that should live in `20-ppp` or in the boot sequence itself.

RedVortex:
**Important notices**

- [x] I have read the contributing guidelines

- Model used: Claude Opus 5 (via Claude Code)
- Extent of AI involvement: analysis of the ordering and startup races, drafting
  the patch and its comments, and a test harness for the CARP decision logic.
  Reviewed, tested on my HA pair and submitted by me.

---

**Describe the problem**

With `disconnectppps` enabled the PPP link should only run on the CARP MASTER.
Two things prevent that today:

1. `interfaces_configure()` orders interfaces by queue and then by description.
   Nothing in that order knows a PPP interface depends on the CARP VIPs of its
   parent, so depending on which side wins, either the BACKUP node starts PPP
   and keeps it, or it starts PPP and the CARP hook suspends it moments later
   -- the short window where both nodes hold a session that @fichtner raised.

2. During boot the link has two possible starters: `interfaces_configure()` and
   the CARP hook. On a node that becomes MASTER while still booting, both can
   fire, so the link is started twice or killed and redialled right after
   coming up.

The same gap applies outside boot: Save/Apply on a BACKUP node starts the link
there, and since `20-ppp` only fires on a state transition, nothing brings it
back down again.

One file, `src/etc/inc/interfaces.inc`, +70/-2.

- `interface_ppps_carp_master()` reports whether CARP controls the link and
  whether this node is MASTER. A VIP configured on the parent but not up yet
  counts as not MASTER, which makes the decision independent of interface
  order. It uses the resolved `$ports`, so a PPP link on top of an assigned
  interface works (see the existing XXX in `interface_ppps_hardware()`), and it
  requires a MASTER rather than the absence of a BACKUP, so a parent with
  several vhids is judged correctly.
- The decision is made before mpd5 is set up at all, so a link is never started
  only to be stopped again.
- While booting, a CARP-controlled link is left to the CARP hook, the only
  caller that knows the election is over. That gives it a single starter, so no
  locking is needed. `interface_ppps_configure()` gained an optional
  `$defer_to_carp` argument for this, passed from `interface_configure()`;
  existing callers are unaffected and `rc.syshook.d/carp/20-ppp` is untouched.
- After the link came up the state is checked once more, because a teardown
  arriving before mpd5 wrote its pid file could not have stopped it.

Behaviour without `disconnectppps` is unchanged: `interface_ppps_carp_master()`
returns null before touching ifconfig or the VIP configuration, and every
branch is gated on it.

**Testing**

Static: `php -l` clean. The CARP decision logic was extracted from the patched
file and run against stubbed config/ifconfig lookups -- 22 cases across three
suites covering HA off, HA on without CARP, MASTER/BACKUP/INIT, VIP configured
but not yet up, VIP on an unrelated interface, ipalias vs carp, missing `mode`
key, multiple vhids, MLPPP, serial modem device nodes, and my real topology.
All pass under E_ALL with no notices.

On hardware, an HA pair where the two nodes do not even share a parent device
for the link -- master is `ix0` (physical), backup is `vlan040` -- both on
vhid 2:

| Test | Result |
|------|--------|
| Reboot BACKUP | link stays down, gate logs the skip, mpd config not regenerated |
| Maintenance failover | new MASTER starts the link once, mpd5 start time matches the CARP event to the second |
| `configctl interface reconfigure wan` on MASTER | link restarts cleanly, address restored |
| Reboot demoted (maintenance) node | link stays down, live WAN on the peer completely undisturbed |
| Failback | both nodes transition in the same second, exactly one session |
| Reboot MASTER without maintenance | WAN moves to the peer and back, single start on each, never two sessions |
| Forced ordering: PPP interface configured before its parent | link stays down through the configuration fallback |
| Save/Apply WAN on BACKUP from the GUI | no session started, skip logged from `/interfaces.php` |

Every transition ended with exactly one mpd5 across the pair.

For the forced-ordering case I set the PPP interface's IPv6 type to None and
renamed the parent so it sorts after it. The skip is then logged *before* the
CARP event for the parent's VIP, so the decision came from the configuration
rather than from live CARP state -- the case that could not be decided before.
Reproduced on 26.7.1_1; master resolves VLAN devices differently since
137513cc25, so the same trick may not provoke that ordering there.

**Not covered by these tests**

- The deferral path needs a node that becomes MASTER *during* boot, which only
  happens when the peer is absent (cold start of a single node, or both nodes
  booting together after a power cut). Reasoning and unit tests only.

**Known limitation**

Teardown on the old MASTER and startup on the new one are driven by devd
independently on each node, so a brief window remains on failover. Split brain
gives two MASTERs and therefore two sessions by definition. Neither is
addressable from a single node.

### DIFF
--- a/src/etc/inc/interfaces.inc
+++ b/src/etc/inc/interfaces.inc
@@ -1338,6 +1338,39 @@ EOD;
         @unlink('/var/spool/lock/LCK..' . basename($port));
     }
 
+    /*
+     * HA/CARP safety for PPP links:
+     * If "Disconnect dialup interfaces" is enabled, only allow PPP startup
+     * when at least one underlying parent interface is currently CARP MASTER.
+     *
+     * This prevents a rebooted BACKUP node from starting PPPoE later in boot
+     * after the CARP hook has already tried (too early) to suspend it.
+     */
+    if (!empty($config['hasync']['disconnectppps'])) {
+      $carp_master = false;
+
+      foreach ($ports as $port) {
+        $ifconfig_out = shell_exec('/sbin/ifconfig ' . escapeshellarg($port) . ' 2>/dev/null');
+        if (!is_string($ifconfig_out)) {
+          continue;
+        }
+
+        if (preg_match('/\bcarp:\s+MASTER\b/m', $ifconfig_out)) {
+          $carp_master = true;
+          break;
+        }
+      }
+
+      if (!$carp_master) {
+        log_msg(
+          "interface_ppps_configure() skipped for {$interface} " .
+          "(disconnectppps enabled, no parent interface is CARP MASTER)",
+          LOG_NOTICE
+        );
+        return;
+      }
+    }
+
     /* precaution for post-start 'up' check */
     legacy_interface_flags($ifcfg['if'], 'down', false);
 
--- a/src/etc/inc/interfaces.inc
+++ b/src/etc/inc/interfaces.inc
@@ -1339,36 +1339,36 @@ EOD;
     }
 
     /*
-     * HA/CARP safety for PPP links:
-     * If "Disconnect dialup interfaces" is enabled, only allow PPP startup
-     * when at least one underlying parent interface is currently CARP MASTER.
+     * In HA mode with disconnectppps enabled, only start PPP when at least
+     * one underlying parent interface is currently CARP MASTER.
      *
-     * This prevents a rebooted BACKUP node from starting PPPoE later in boot
-     * after the CARP hook has already tried (too early) to suspend it.
+     * During boot, the CARP hook may already have processed a BACKUP event
+     * before the PPP pseudo-interface exists. Without this guard, the later
+     * normal startup path can still start PPP on a BACKUP node.
      */
     if (!empty($config['hasync']['disconnectppps'])) {
-      $carp_master = false;
+        $carp_master = false;
 
-      foreach ($ports as $port) {
-        $ifconfig_out = shell_exec('/sbin/ifconfig ' . escapeshellarg($port) . ' 2>/dev/null');
-        if (!is_string($ifconfig_out)) {
-          continue;
+        foreach ($ports as $port) {
+            $ifconfig_out = shell_exec('/sbin/ifconfig ' . escapeshellarg($port) . ' 2>/dev/null');
+            if (!is_string($ifconfig_out)) {
+                continue;
+            }
+
+            if (preg_match('/\bcarp:\s+MASTER\b/m', $ifconfig_out)) {
+                $carp_master = true;
+                break;
+            }
         }
 
-        if (preg_match('/\bcarp:\s+MASTER\b/m', $ifconfig_out)) {
-          $carp_master = true;
-          break;
+        if (!$carp_master) {
+            log_msg(
+                "interface_ppps_configure() skipped for {$interface} " .
+                "(disconnectppps enabled, no parent interface is CARP MASTER)",
+                LOG_NOTICE
+            );
+            return;
         }
-      }
-
-      if (!$carp_master) {
-        log_msg(
-          "interface_ppps_configure() skipped for {$interface} " .
-          "(disconnectppps enabled, no parent interface is CARP MASTER)",
-          LOG_NOTICE
-        );
-        return;
-      }
     }
 
     /* precaution for post-start 'up' check */
--- a/src/etc/inc/interfaces.inc
+++ b/src/etc/inc/interfaces.inc
@@ -1008,6 +1008,44 @@ function interface_ppps_hardware($device)
     return $devices;
 }
 
+function interface_ppps_carp_master($ports)
+{
+    /*
+     * Whether CARP considers this node the one to run a PPP link on: null when
+     * CARP does not own the link, true when a parent device is MASTER and false
+     * when the link must not run here.  A VIP that is configured but not up yet
+     * counts as false, because interfaces are configured without any notion of a
+     * PPP link depending on the VIPs of its parent and being unable to tell is
+     * not a reason to start.
+     */
+    if (empty(config_read_array('hasync', false)['disconnectppps'])) {
+        return null;
+    }
+
+    $owned = false;
+
+    foreach ($ports as $port) {
+        foreach (legacy_interface_details($port)['carp'] ?? [] as $carp) {
+            if ($carp['status'] == 'MASTER') {
+                return true;
+            }
+            $owned = true;
+        }
+    }
+
+    if (!$owned) {
+        /* the parent interface may not be configured yet, so ask the configuration */
+        foreach (config_read_array('virtualip', 'vip', false) as $vip) {
+            if (($vip['mode'] ?? '') == 'carp' && in_array(get_real_interface($vip['interface']), $ports)) {
+                $owned = true;
+                break;
+            }
+        }
+    }
+
+    return $owned ? false : null;
+}
+
 function interface_ppps_reset($interface, $suspend, $ifcfg, $ppps)
 {
     if (!interface_ppps_capable($ifcfg, $ppps)) {
@@ -1026,7 +1064,7 @@ function interface_ppps_reset($interface, $suspend, $ifcfg, $ppps)
     }
 }
 
-function interface_ppps_configure($interface)
+function interface_ppps_configure($interface, $defer_to_carp = false)
 {
     global $config;
 
@@ -1133,6 +1171,26 @@ function interface_ppps_configure($interface)
                 log_msg("Unknown {$ppp['type']} configured as PPP interface.", LOG_ERR);
                 break;
         }
+    }
+
+    $carp_master = interface_ppps_carp_master($ports);
+
+    if ($carp_master === false) {
+        /* decide before mpd5 is set up at all so a link is never started to be stopped again */
+        log_msg("interface_ppps_configure() skipped for {$interface}, no parent device is CARP MASTER", LOG_NOTICE);
+        killbypid("/var/run/{$ppp['type']}_{$interface}.pid");
+        return;
+    }
+
+    /*
+     * The interface configuration does not know the outcome of the election
+     * while booting, but the CARP event that follows it does.  Let the event
+     * own the link until then so it has a single starter instead of two racing
+     * ones and cannot be brought up twice.
+     */
+    if ($carp_master && $defer_to_carp && product::getInstance()->booting()) {
+        log_msg("interface_ppps_configure() deferred for {$interface} to the CARP event", LOG_NOTICE);
+        return;
     }
 
     // Construct the mpd.conf file
@@ -1365,6 +1423,16 @@ EOD;
         default:
             legacy_interface_mtu($ifcfg['if'], $mtus[0]);
             break;
+    }
+
+    /*
+     * CARP may have moved on while the link was coming up.  A teardown that
+     * arrived before mpd5 wrote its pid file could not stop it, so check the
+     * state once more now that there is a pid file to act on.
+     */
+    if (interface_ppps_carp_master($ports) === false) {
+        log_msg("interface_ppps_configure() stopping {$interface}, no parent device is CARP MASTER", LOG_NOTICE);
+        killbypid("/var/run/{$ppp['type']}_{$interface}.pid");
     }
 }
 
@@ -2484,7 +2552,7 @@ function interface_configure($verbose = false, $interface = 'wan', $reload = fal
     }
 
     /* since the connectivity can include IPv4 and/or IPv6 let the function decide */
-    interface_ppps_configure($interface);
+    interface_ppps_configure($interface, true);
 
     switch ($wancfg['ipaddr'] ?? 'none') {
         case 'dhcp':


RedVortex:
This patch fixes a long-standing issue affecting PPPoE interfaces in OPNsense HA deployments using CARP with the `disconnectppps` option enabled.

Problem
In an HA setup, when a backup node boots, CARP can correctly transition the interface to `BACKUP` and trigger `rc.syshook.d/carp/20-ppp`, which attempts to suspend the PPP interface. However, at that point in the boot sequence, the PPP pseudo-interface may not exist yet, so the suspend happens too early to be effective.

Later in the normal boot process, `interface_ppps_configure()` is still called and starts PPP anyway, even though the node is already in CARP `BACKUP` state.

This causes an unwanted PPPoE session to be established on the backup node during boot.

Impact
In real HA environments, this can lead to multiple problems:

* the backup node may briefly or fully establish the PPPoE session while the master is already active
* the active session on the master may be disrupted or conflicted
* the backup may temporarily receive a public IP it should never own
* Dynamic DNS or other WAN-triggered services may publish or act on the wrong IP
* the issue only self-corrects later if a new CARP transition occurs, which means the backup can remain in a bad state until failover/failback happens

In short, the existing CARP hook reacts to the state change, but the later PPP startup path can override that decision during boot.

Goal of this patch
The goal of this patch is to prevent PPP from starting at all when the node is not the CARP master.

This patch adds a guard inside `interface_ppps_configure()` so that when `hasync.disconnectppps` is enabled, PPP startup is allowed only if at least one of the underlying parent interfaces is currently in `carp: MASTER` state.

If no parent interface is `MASTER`, the function returns before launching `mpd5`.

Why this approach
This issue is specifically a boot-order / startup sequencing issue in HA environments.

The existing `20-ppp` CARP hook is still useful for runtime CARP transitions, but it is not sufficient during boot because it can run before the PPP interface exists. The correct place to enforce this safely is therefore in the PPP startup path itself.

This makes the behavior fail closed:

* `MASTER` -> PPP is allowed to start
* `BACKUP` / `INIT` / not yet settled -> PPP is not started

This is safer for HA because it prevents the backup node from briefly establishing an unwanted PPP session.

Testing
I tested this patch in my own OPNsense HA environment using CARP with PPPoE on WAN and `disconnectppps` enabled.

In my test case, before this patch:

* the backup node would boot
* CARP would correctly transition to `BACKUP`
* `20-ppp` would run
* but later in boot, PPPoE would still start on the backup node

After this patch:

* the backup node no longer starts PPPoE during boot
* the active/master node continues to operate normally
* the behavior now matches the intended HA expectation for `disconnectppps`

Request for review
This patch has been validated in my own HA environment and resolves the issue for me, but I would appreciate review from maintainers to confirm:

* that this change does not introduce side effects in other PPP configurations
* that the logic is correct for other HA / CARP scenarios
* that the implementation style and placement are aligned with the surrounding code in `interfaces.inc`
* and whether there is a preferred alternative location or helper for this check within the core codebase

I believe the current behavior is a real HA bug, but I would like reviewer feedback to ensure the fix is correct and consistent with the project’s expectations before it is merged.

src/etc/inc/interfaces.inc
```diff
@@ -1338,6 +1338,39 @@ EOD;
...
+            $ifconfig_out = shell_exec('/sbin/ifconfig ' . escapeshellarg($port) . ' 2>/dev/null');
+            if (!is_string($ifconfig_out)) {
+                continue;
+            }
+
+            if (preg_match('/\bcarp:\s+MASTER\b/m', $ifconfig_out)) {
+                $carp_master = true;
+                break;
+            }
+        }
```
fichtner: we can adjust this for better use of helper functions but for now my question is if we're applying the right metric here for multi-link PPP and general startup.

1. Shouldn't $carp_master be true unless one or all ports are not master?
2. Perhaps "BACKUP || INIT is a better metric? Looking at https://github.com/opnsense/core/blob/43c933d6a67cac11841ddc9803565bb279f13219/src/etc/rc.syshook.d/carp/20-ppp#L54
3. We're kinda too far into the configuration at this point do to anything drastic such as interface_suspend() in contrast to 20-ppp so I'm wondering where the best spot is to run this code in this function or elsewhere as a preamble. It's only called in 3 places after all.

fichtner: The last thought is why not defer to 20-ppp completely. I suspect the master doesn't come up then on a reboot or both machines think they are master for a bit? I'm talking about a hard return of the code, but that would be ouside the function kinda like 3.) implies

fichtner: the funny thing is 20-ppp isn't using MLPPP 

https://github.com/opnsense/core/blob/43c933d6a67cac11841ddc9803565bb279f13219/src/etc/rc.syshook.d/carp/20-ppp#L50

Maybe we should emulate this here and annotate it so the behaviour matches (more or less point 2)

RedVortex: Yeah I really wanted and tried to use 20-ppp but unfortunately it doesn't work since it is called early in the bootup. It tries and fails to bring down pppoe because it is not even up yet, we even see errors logs in the bootup regarding this.

And some time during bootup, after 20-ppp was called, the bootup brings up all interfaces and the code that handles bringing up all interfaces actually doesn't seem to care if it should or not bring up pppoe (because of CARP status) and since CARP is stable at this point (BACKUP since this is the BACKUP CARP node, the other node is MASTER), 20-ppp is never called again. Furthermore, even if it would be called later on, the damage is done during bootup since it started a pppoe session but it shouldn't have (thus creating issues since we now have 2 nodes having a public WAN on the same PPPoE user and also breaking Dynamic DNS thus flapping the Dynamic DNS public IP between the 2 nodes now).

This issue really happens only during a reboot of opnsense (during bootup), never otherwise in my case (except if I edit the pppoe interface and save it, then it also brings it up (but it shouldn't since the node is CARP BACKUP) but I haven't tried that (save the pppoe interface on the backup to see if my patch also fixes/handle this, I'll give it a shot later tonight).

Thus why I tried to put this code in this akward place to handle this situation (pppoe coming up on bootup whether or not in CARP BACKUP mode). This patch also only applies when the HA setting is configured to bring down ppp interfaces when CARP becomes BACKUP.

Also, to be honest I never considered if it would have an impact on MLPPP but I assume the logic is the same, I'm not sure it is handled at the same place in the script though.

I thought also about creating a custom 99-scriptsomething that would trigger and bring down pppoe later or a cronjob checking the state of CARP and bring it down in case it came up after bootup but it is very bad that the pppoe came up at all. Some providers with block you or rate limit you if you try to bring up 2 pppoe on the same user or link or even create conflicts if, for example at my work, we have a static IP on our PPPoE username, thus resulting in conflict between the 2 nodes. So it definitely should never come up at all if the node is BACKUP even if bring it down later, it creates important problems.

fichtner:
I've just gone over the code bits and reduce your solution. Can you confirm this one works on top? 2d30a94f5ba916a707b9fbbc76692fa26c12705e

I want do take this further but not without getting some basics out of the way first.


Thanks,
Franco

RedVortex:
> I've just gone over the code bits and reduce your solution. Can you confirm this one works on top? [2d30a94](https://github.com/opnsense/core/commit/2d30a94f5ba916a707b9fbbc76692fa26c12705e)

Sorry about the delay getting back to you, I needed to make sure nobody would be complaining if I brought the network down, hehehe. And I also wanted to test this properly for you.

Also, I had some issues applying the patch, seems like I had to apply everything in between 26.1.3 and this patch for both files

My 2 patches after 26.1.3 on src/etc/inc/interfaces.inc

```
opnsense-patch 0c1404e
opnsense-patch 673acfb
```

2 patches you did on 20-ppp after 26.1.3 on src/etc/rc.syshook.d/carp/20-ppp

```
opnsense-patch b7166dc
opnsense-patch 81a0db2
```

And finally your last one on top of those 4 for both src/etc/inc/interfaces.inc and src/etc/rc.syshook.d/carp/20-ppp

`opnsense-patch 2d30a94`

I'm happy to report that this works very well for me.

I tested:

- Manual Failover and Failback (Enter Persistent CARP Maintenance Mode) and it worked without any issue. pppoe got enabled on the BACKUP once it became MASTER and MASTER brought pppoe DOWN when it became BACKUP
- Reboot of the BACKUP node and pppoe remained DOWN as it should be (that's the fix we want, yesss !)
`/usr/local/etc/rc.bootup: interface_ppps_configure() skipped for wan (disconnectppps enabled, no parent interface is CARP MASTER)`

- Went into the interface on the BACKUP node, saved it and applied it and it remained DOWN as it should be (bonus fix, great !)
`/interfaces.php: interface_ppps_configure() skipped for wan (disconnectppps enabled, no parent interface is CARP MASTER)`

- Went into the interface on the MASTER node, saved it and applied it and pppoe restarted the pppoe properly as it should

So far so good to me, looks great on my side, no issue to report from my point of view and setup. And I definitely prefer the way you did that versus my way too looong code in the interfaces includes...

Thanks,
Patrick

fichtner:
@RedVortex thanks for your tests. I'm not really satisfied with the code for historic reasons. Did a futile refactor in 5914ce73981aa3 but realized we're depending on CARP when we start before CARP is even configured (the first time). This needs more brain cells and time, but I will continue.


Cheers,
Franco

RedVortex:
@fichtner Thanks.

This is definitely not urgent at all, I've been living with this little issue for years so I'm used to it, there is absolutely no reason to rush this in any way.

This is indeed a "touchy" patch to do, especially in those include files. I totally understand that we have to play it very safe here and not burn ourselves and especially not burn everyone else either. Especially to fix something that is as small as this bug. Pretty sure not a lot of people have this issue and/or they don't bother to try to fix or debug it, lol, I took a chance.

I'll be monitor this thread closely in case you need me for anything, in the meantime I'll use our little patches that are good enough for me until we get something more robust.

Pat

RedVortex:
I decided to burn a few tokens, lol, with the help of Claude and updated this PR to reflect the current state of 26.1.7:

- The `20-ppp` hook cleanup from this PR already landed on master as `547ae9d10` — removed from here.
- Replaced the original `shell_exec`/regex approach in `interfaces.inc` with the `legacy_interface_details` version from your `ppp_carp` branch (`825836f73`), which you wrote and I confirmed working during testing.
- Single clean commit, one file changed, no conflicts with current master.

Your `ppp_carp` branch already has this fix — this PR just puts it in a mergeable state if you'd like to close the loop.

I'll test it properly against the latest hotfixes for 26.1.7 and let you know if it still works for me.

RedVortex:
So, I updated to 26.1.7_3 and rebooted. After reboot, on the CARP backup, the pppoe was up (our current bug, it should not have come up since this is a backup).

I then patched this PR

```
root@opnsense-backup:~ # opnsense-patch 05dc998
Fetched 05dc998 via https://github.com/opnsense/core
Hmm...  Looks like a unified diff to me...
The text leading up to this was:
--------------------------
|From 05dc9985e440a2da0d834d14461d79877d5a9ff0 Mon Sep 17 00:00:00 2001
|From: xxxxxxxxxx
|Date: Sat, 9 May 2026 16:54:50 -0400
|Subject: [PATCH] interfaces: only start PPP on CARP MASTER when disconnectppps
| is enabled
|
|---
| src/etc/inc/interfaces.inc | 24 ++++++++++++++++++++++++
| 1 file changed, 24 insertions(+)
|
|diff --git a/src/etc/inc/interfaces.inc b/src/etc/inc/interfaces.inc
|index deecbea627..9a1344ad85 100644
|--- a/src/etc/inc/interfaces.inc
|+++ b/src/etc/inc/interfaces.inc
--------------------------
Patching file etc/inc/interfaces.inc using Plan A...
Hunk #1 succeeded at 1338 (offset 18 lines).
done
All patches have been applied successfully.  Have a nice day.
```

Then rebooted again and pppoe remained down, as it should.

Tried a failover (rebooted the primary) to make sure backup would become master and bring up pppoe and it did and when primary came back online, the backup became backup again and pppoe was brought down (as it should) and primary brought up pppoe when it became primary again.

RedVortex:
As a follow-up, I also just tested the same patch against 26.1.8_5 and it still works well.

RedVortex:
Also tested and working well against OPNsense 26.1.9-amd64

RedVortex:
Also tested and working well against OPNsense 26.1.10-amd64

RedVortex:
Also tested and working well against OPNsense 26.1.11_5-amd64

RedVortex:
Also tested and working well against OPNsense 26.7-amd64

fichtner:
Hey Patrick, I'm so sorry I dropped the ball here. I'm still a bit conflicted about the code in general because:

https://github.com/opnsense/core/blob/09b63a4bf2b4388051b65bfc4b500cb46e6ecb50/src/etc/inc/interfaces.inc#L2478

We init PPP here, but further down start adding VIPs:

https://github.com/opnsense/core/blob/09b63a4bf2b4388051b65bfc4b500cb46e6ecb50/src/etc/inc/interfaces.inc#L2533

So the issue is that we don't have CARP information the first time we pass the code and we'd start PPP on both machines, even if just for a brief period of time.

So this needs a better approach, but I'm unsure what it is.


Cheers,
Franco

RedVortex:
Hi Franco,

no worries, I totally understand and there is no rush fixing this. For now it's easy for me to simply re-patch after a release. 

Also, I tested this over the last week or so while troubleshooting the gateway issues post 26.7 a few of us discovered which highly impacted my setup. At first I thought this was my patch causing me issues and thus I did not report back success yet with 26.7.1 but after some more troubleshooting I found out the gateway issue wasn't the patch but the opnsense sense update itself.

Btw, thanks for fixing it so rapidly, I'll be able to remove my temporary cronjob that was reloading gateways that I had put in place to address this until a patch was released 👍 

I can confirm my patch still works very well with OPNsense 26.7.1_1-amd64 now.

Also, I'll be investigating what you mentioned here since I don't believe that I see or experience this:

`So the issue is that we don't have CARP information the first time we pass the code and we'd start PPP on both machines, even if just for a brief period of time.
`

I'll try to see if I can replicate it and maybe see if we can adapt the patch in some way to address it. For now, even if the patch surely may not be perfect, it still solves the bug of pppoe coming up after reboot when it shouldn't and is much better than not having it. It is also small and precise enough that it does not create new bugs (from my tests and setup at least), that was my only goal and thus why I kept it so small after your original comments.

I'll let you know if I can replicate what you report and if I can find a way to also address it.

Thanks for the follow-up

Pat

RedVortex:
Franco, I'm replacing my last few comments with this one — they were a running log and got hard to follow.

**Your concern was right, and it's handled now.** I couldn't reproduce it at first: my CARP VIP sits on the PPPoE parent, and the parent always happened to be configured first — the PPP interface lands in the `$dhcp6c` queue, and "BellLAN" sorts before "BellWAN". That's luck of naming and settings, not something the code guarantees, so the ordering you described is reachable. I just never hit it.

**What the patch does** — one file, +70/-2, every branch gated on `disconnectppps`:

- When live CARP state isn't available yet, it falls back to the VIP configuration and treats "VIP configured on the parent but not up" as *not* MASTER. The decision no longer depends on interface order.
- While booting, a CARP-controlled link is left to the CARP hook — the only caller that knows the election is over — so there's one starter instead of two racing.
- It re-checks once the link is up, since a teardown arriving before mpd5 wrote its pid file couldn't have stopped it.

**I did manage to reproduce your case**, by forcing the reverse ordering on the backup (PPP interface's IPv6 type set to None so it lands in `$hardware`, parent renamed so it sorts after it). The skip is then logged *before* the CARP event for the parent's VIP, and no session starts. On every normal boot those two are the other way round.

**Tested** across seven reboot/failover scenarios on my pair, plus Save/Apply on the backup from the GUI. Every transition ended with exactly one PPPoE session across the pair. The boot-deferral path is reasoning and unit tests only — it needs a node that becomes MASTER while booting, i.e. the peer down.

Full detail is in the PR description. The branch was force-pushed, so commit hashes in older comments no longer resolve.

RedVortex:
Tested and still working well against OPNsense 26.7.3_8-amd64

src/etc/inc/interfaces.inc
```diff
@@ -1008,6 +1008,44 @@ function interface_ppps_hardware($device)
...
+    $owned = false;
+
+    foreach ($ports as $port) {
+        foreach (legacy_interface_details($port)['carp'] ?? [] as $carp) {
+            if ($carp['status'] == 'MASTER') {
+                return true;
+            }
+            $owned = true;
+        }
+    }
```
fichtner, line 1034: Hi! :) I wonder if we could make this stateless by omitting ifconfig MASTER state and simply letting the CARP hooks take the lead? In this case no PPP would be opened during boot or other normal reconfiguration through interface_configure().

src/etc/inc/interfaces.inc
```diff
@@ -1135,6 +1173,26 @@ function interface_ppps_configure($interface)
         }
     }
 
+    $carp_master = interface_ppps_carp_master($ports);
+
+    if ($carp_master === false) {
+        /* decide before mpd5 is set up at all so a link is never started to be stopped again */
+        log_msg("interface_ppps_configure() skipped for {$interface}, no parent device is CARP MASTER", LOG_NOTICE);
+        killbypid("/var/run/{$ppp['type']}_{$interface}.pid");
+        return;
```
fichtner, line 1182: These returns are weird in hindsight: we're not creating the pppoe interface so later code in interfaces_configure() is practically useless.

Also, do you know what happens when we simply DOWN the pppoe interface? If we could get it into a hibernation state that would simplify some things since the interface would exist and interfaces_configure() could do its job regardless.

src/etc/inc/interfaces.inc
```diff
@@ -1366,6 +1424,16 @@ EOD;
             legacy_interface_mtu($ifcfg['if'], $mtus[0]);
             break;
     }
+
+    /*
+     * CARP may have moved on while the link was coming up.  A teardown that
+     * arrived before mpd5 wrote its pid file could not stop it, so check the
+     * state once more now that there is a pid file to act on.
+     */
+    if (interface_ppps_carp_master($ports) === false) {
```
fichtner, line 1433: yes so if the hook could decide on its own we'd never end up here

src/etc/inc/interfaces.inc
fichtner: if it's not a hallucination there's room for a control socket and we could omit the "open" directive and operate the control socket from the carp script that simply issues `set link max-redial -1` + `close` and `set link max-redial 0` + `open`. It would eliminate all side effects WRT interface_configure() and we don't even need the extra function parameter as interface_ppps_configure9) won't be called from the CARP hook.

fichtner: (fixed some hallucination)

RedVortex:
Hi Franco,

Thanks for picking this back up, and no rush at all on my side.

I agree with where you're heading — one owner for the link is cleaner than what I have. Before changing anything, let me write down why it ended up this shape, since most of it came out of testing rather than design.

One note up front: this has been going on for a few months and the details were no longer fresh in my head, so I went back over the whole thread, the commit history and my own test notes with Claude's help to reconstruct why the patch looks the way it does. What follows is that summary. The reasoning and the testing are mine, and I've checked each claim against the code, the git history or the mpd5 manual, but the digging back through it all was assisted.

**Why the CARP state check is where it is**

- Both nodes carry the same VIP in config by definition — the same vhid on both is what makes CARP work — so nothing in config.xml can tell a node whether it is the one that should dial. Live CARP state is the only per-node signal there is. That's why the config fallback in the patch can only ever answer "don't start", never "start": it knows the link is CARP-controlled, not that this node is MASTER.
- 20-ppp only fires on CARP transitions. Save/Apply on the WAN, or `configctl interface reconfigure wan`, produces no CARP event. If `interface_ppps_configure()` stops consulting CARP state, a reconfigure on the MASTER takes the link down and nothing brings it back until the next failover, which on a healthy pair can be a long time. Restart-on-MASTER is a row in my test table; the mirror case, Save/Apply on the BACKUP starting a session, is a bug this patch fixed as a side effect.
- The ifconfig check itself is yours — the `legacy_interface_details()` version from 825836f73 that you wrote in March as "a starting point". I kept it over my original `shell_exec`/regex because it was cleaner and I had tested it. The only change was requiring a MASTER rather than the absence of BACKUP/INIT, so a parent with several vhids is judged correctly.

**Boot versus failover**

On a failover everything already exists, and the hook does its job — that path was never the problem. Boot is the asymmetric one: the BACKUP's CARP event lands before the PPP interface exists, `interface_suspend()` has nothing to act on, and since there is no further transition the hook never runs again. That is the original bug, and it is why the decision ended up in the startup path rather than in the hook. So the question I'd have about the stateless version is what opens the link on a node whose MASTER event arrives before mpd5 exists — same timing, other direction. If the answer is that the hook is re-triggered or replayed, that solves my second point above too.

**On the specific things you raised**

- The interface not being created on the BACKUP: that is also where the existing path ends up — `interface_suspend()` → `interface_reset(..., true)` → `interface_ppps_reset()` → `killbypid()`, and the interface goes away. So the patch makes the boot state match the post-failover steady state rather than inventing a new one. Not saying that's ideal, only that it isn't a state the backup doesn't otherwise reach.
- Downing the pppoe interface: I'd expect the session to survive, since it lives in netgraph rather than in the interface flags, but I have not tested it. I can if it's useful.
- The control socket: mpd5's console is TCP only — `set console self <ip> <port>` plus `set user <name> <pass> <priv>`; the manual describes open/close as creating and closing "the listening TCP socket", and there's no unix-domain option. core has emitted `set console close` and `set web close` since the initial import. Bound to loopback it isn't network-exposed, but it is still a new local control channel with a password on every PPP-capable install, so that felt like your call rather than mine to assume.
- Omitting the `open` directive: one thing to watch is that `interface_ppps_configure()` blocks up to 20s waiting for `does_interface_exist($ifcfg['if'], 'up')` and then logs a warning, so with no `open` that becomes a 20s stall per PPP interface on the backup on every boot and reconfigure. I'd also want to check how `ondemand` links behave there, since `set iface enable on-demand` dials on traffic — whether an `open` is still required in that mode I don't know, and I haven't tested it.
- The post-up recheck: it guards a teardown that lands while mpd5 is starting. Between the fork and the pid file being written, `killbypid()` has nothing to act on, so a BACKUP event in that window leaves a session running on a node that is no longer MASTER. A resident mpd5 with a control channel would close that window properly, which would be a real gain over what I have.

**Where I stand**

I saw your note on #10663 — agreed on side effects in the interface code, that's the part that worries me too, and it's why every branch here is gated on `disconnectppps` and nothing outside `interfaces.inc` is touched. Worth adding that biggaeh's confirmation on that issue is a different topology from mine — KVM/virtio, PPPoE over a VLAN-tagged transfer network, `opt10` rather than `wan` — so there are two independent setups behind it now rather than just mine.

The notes above are meant as input for your experiments rather than pushback, and I'm glad to keep tuning this. The one thing I'd rather not do is author a redesign that reaches paths I have no way to test — MLPPP, ondemand, 3G modem, PPTP/L2TP WANs, non-HA installs. If you push a branch I'll test it on a live HA pair against a real PPPoE ISP and report back per scenario, same as before. And if it's simpler for you to carry it yourself, that's completely fine by me — the bug getting fixed matters more than whose patch does it.

Pat

RedVortex:
Tested and still working well with OPNsense 26.7.3_11-amd64

RedVortex:
I'll be checking the new conflict likely tonight or this weekend

RedVortex:
Conflict resolved — merged current master in. The only collision was the comment line above the `interface_ppps_configure()` call that 7cd941d20c rewrote, so the resolution just keeps your comment and the new `interfaces_wlan_configure()` line alongside the existing call.

I also noticed where the interface code seems to be heading — the pluggable daemon start/reset hooks sketched in the XXX comments next to the WLAN path. If that lands, it should absorb a fair part of what this patch currently does by hand. I'll keep tracking master as it moves toward 27.1 and keep this branch aligned, and I'm happy to rework it onto whatever shape that takes.

For anyone testing on 26.7.x, `opnsense-patch aa1df88dd3` still applies cleanly — that commit is untouched.

RedVortex:
Rebased onto current master as a single commit, replacing the earlier merge. No code change, so the review comments are still on the same lines.

Works on 26.7.4_1. I upgraded and patched both nodes of my HA pair, then ran a BACKUP reboot, a maintenance failover, a reboot of the demoted node and a failback. Every step ended with exactly one PPP session across the pair.

It also handled a quick CARP flap correctly. Saving a pfsync change briefly demotes CARP, so the backup's MASTER event reached 20-ppp after the backup was already BACKUP again, and the live state check skipped the start.

To try it: `opnsense-patch 94eb5b1125` on 26.7.4 / 26.7.4_1 (it also applies to master), or `opnsense-patch aa1df88dd3` on 26.7.3.

AdSchellevis:
it might be good to double check if the pfsync glitch is in the same corner as https://github.com/opnsense/core/issues/10858 , @swhite2 is looking into that, but the behavior sounds a bit similar.

RedVortex:
@AdSchellevis yes, same corner. Reproduced it and traced it to the pfsync bulk END check in the kernel, details in https://github.com/opnsense/core/issues/10858#issuecomment-5738452810. The patch behaved correctly throughout: one PPP session across the pair on every run.

- Model used: Claude Opus 5 (via Claude Code)
- Extent of AI involvement: the analysis behind the linked report and drafting this comment, reviewed and submitted by me.

AdSchellevis:
When unchanged the interface shouldn’t try to reconfigure from our end, which logically should fix most of the issue.

RedVortex:
> When unchanged the interface shouldn’t try to reconfigure from our end, which logically should fix most of the issue.

I definitely agree on this part, it really should not do that unless there are changes. Then again, when there are changes, we can also encounter a much worse issue (not always but often) that "breaks" CARP and triggers a failover for a few minutes which is quite bad, as I described in https://github.com/opnsense/core/issues/10858#issuecomment-5738452810.

But even if it does not encounter the larger bug, it still triggers a "flash CARP demotion event" which is not ideal but might not be avoidable since there are changes but yeah, I believe this behaviour could also be improved, maybe, but it would be a very rare occurence that someone does this anyways unless they are setting up HA (or tuning it or upgrading it from 1400 to 1500 like I was doing - first time in a long time we had a new sync version update), likely why nobody ever complained or saw it much really before now.

aeble:
Thanks for keeping this going. I'd like to raise a design question rather than report a problem, because I think the answer shapes where the decision belongs.

**Should the availability of the uplink depend on CARP?**

Today it does not. If CARP misbehaves, the line still comes up, possibly on the wrong node, but it comes up. With the link deferred to a CARP decision, any CARP fault becomes an internet outage. Cases where no node cleanly reaches MASTER are not exotic:

- `net.inet.carp.demotion` can get stuck at a multiple of 240 after a lost `interface up`,  and the node then looks healthy while being unable to take over
- a hung pfsync bulk update demotes permanently
- [#10781](https://github.com/opnsense/core/issues/10781) describes a split brain on exactly the VHID of a PPPoE carrier, present since 26.7.2_2 and still without a comment

In each of those, "wait for MASTER" means "nobody dials".

**Two reasons why a VIP on the carrier is not always available**

First, some OLTs limit the number of MAC addresses behind the ONT. A CARP VIP on the carrier adds `00:00:5e:00:01:<vhid>` as a third address next to the two node MACs, and [a well-known write-up of this exact setup](https://www.mayrhofer.eu.org/post/opnsense-pppoe-ha/) describes an OLT that allowed only two, so the CARP MAC had to be filtered on the switch. FreeBSD offers no way to suppress the virtual MAC; the full `net.inet.carp.*` sysctl set has no option for it, unlike keepalived's `use_vmac` on Linux.

Second, #10781 makes adding one a risk of its own, and its trigger is disabling and re-enabling CARP, which is exactly what the handbook prescribes before changing a VIP.

**Our topology**, OPNsense 26.7.4_1, two nodes: PPPoE runs on `vlan0.7.2`, a VLAN device that is not assigned as an interface; only `pppoe1` is assigned. All nine CARP VIPs live on internal segments, all on VHID 112. `disconnectppps` is `1`. `20-ppp` therefore never matches, both nodes dial, and whoever dialled last wins. Measured on 2026-09-19: an apply on node 1 restarted its PPP daemon, node 2 authenticated one second later, and the CARP master
then sent PADI for ten minutes. IPv6 for the whole site routes over that line.

**A suggestion: a window instead of a condition.** Rather than "start only when MASTER", wait up to N seconds for a CARP decision and start anyway if none arrives, preferring the node with the lower `advskew`. In the normal case that costs a few seconds at boot. In the degraded case it keeps the uplink independent of CARP, which is where it is today. It would also cover setups where the carrier holds no VIP at all, since there the current approach can only ever answer "don't start".

I have not tested the patch; this is read from the description and from your comment of 2026-09-04. I'd rather not patch a production node for this, but I'm happy to supply any detail about this topology that is useful, and to confirm behaviour once it lands in a release.

PR #9907, opened 2026-03-07
firewall.js should keep coloring after reload, use sematic groups blo…

GunnarLieb:
- Color slices by semantic group: block/deny (red), pass/allow (green), redirect/nat (blue)
- Use label-based matching (CrowdSec, haproxy, etc.) with action fallback                                                  
- Hash rule ID for deterministic shade selection across reloads                                                            
- Merge duplicate labels into single slices with summed counts                                                             
- Disable colorschemes plugin to use explicit color assignment

### DIFF
--- a/src/opnsense/www/js/widgets/Firewall.js
+++ b/src/opnsense/www/js/widgets/Firewall.js
@@ -58,8 +58,8 @@ export default class Firewall extends BaseTableWidget {
             let factor = ((hash % 10) - 5) * 0.08;
             return this._shadeColor(base, factor);
         }
-        let rids = this.chart.data.datasets[0].rids;
-        return colors[rids.length % colors.length];
+        const i = index ?? this.chart.data.datasets[0].rids.length;
+        return colors[i % colors.length];
     }
 
     async getWidgetOptions() {


GunnarLieb:
looks like this now:
CleanShot 2026-03-07 at 09 48 43@2x

src/opnsense/www/js/widgets/Firewall.js
```diff
@@ -31,6 +31,39 @@ export default class Firewall extends BaseTableWidget {
...
+            group = 'redirect';
+        } else if (action === 'block') {
+            group = 'block';
+        } else if (action === 'pass') {
+            group = 'pass';
+        } else {
+            group = 'redirect';
+        }
+        let palette = this.palettes[group];
+        return palette[this._hashRid(rid) % palette.length];
```
swhite2: A fw rid is already a hex string hash.

`const idx = parseInt(rid.slice(0, 8), 16);`

src/opnsense/www/js/widgets/Firewall.js
```diff
@@ -31,6 +31,39 @@ export default class Firewall extends BaseTableWidget {
         this.counters = {};
         this.chart = null;
         this.rotation = 5;
+        this.palettes = {
+            block:    ['#e15759', '#c44e52', '#a83a3d', '#ff6b6b', '#d94f4f', '#b33636', '#e87272', '#cf5c5c', '#a04040', '#ff8585'],
+            pass:     ['#59a14f', '#4c8c43', '#3d7a36', '#6fbf65', '#52a347', '#448c3a', '#7acc70', '#5db352', '#4a9c42', '#88d680'],
```
swhite2: Are these color codes arbitrary or picked from the existing colorschemes plugin?

src/opnsense/www/js/widgets/Firewall.js
```diff
@@ -31,6 +31,39 @@ export default class Firewall extends BaseTableWidget {
...
+        } else if (/haproxy|rdr rule|nat rule/i.test(label)) {
+            group = 'redirect';
+        } else if (action === 'block') {
+            group = 'block';
+        } else if (action === 'pass') {
+            group = 'pass';
+        } else {
+            group = 'redirect';
+        }
+        let palette = this.palettes[group];
```
swhite2:
```suggestion
        let palette = this.palettes[action];
```

But only if all actions are defined in `palettes`. We are however not going to match on arbitrary label strings

src/opnsense/www/js/widgets/Firewall.js
```diff
@@ -31,6 +31,39 @@ export default class Firewall extends BaseTableWidget {
         this.counters = {};
         this.chart = null;
         this.rotation = 5;
+        this.palettes = {
```
swhite2: better to match on https://github.com/opnsense/core/blob/master/src/opnsense/www/js/widgets/Firewall.js#L88-L91

src/opnsense/www/js/widgets/Firewall.js
```diff
@@ -147,25 +180,37 @@ export default class Firewall extends BaseTableWidget {
...
         } else {
-            data[idx] = count;
+            // sum counts for all rids that share this label
+            let total = 0;
+            for (let r in this.ridToLabel) {
+                if (this.ridToLabel[r] === decodedLabel && this.counters[r]) {
+                    total += this.counters[r].count;
+                }
+            }
+            dataset.data[idx] = total;
```
swhite2: Not sure I understand this code. Isn't there a 1-to-1 relationship between rid <-> label?

swhite2 (changes requested):

GunnarLieb:
I hope I resolved everything, please review, here is what I changed:
- Hash simplification — Removed the custom _hashRid() method entirely. Now uses parseInt(rid.slice(0, 8), 16)
  directly since rid is already a hex hash string.
- Color palette origin — Added a comment clarifying the palettes are derived from Tableau 10 (a standard,
  well-known chart color scheme), with shade variants generated per semantic group.
 - Action matching — Removed all regex label-pattern matching (crowdsec, haproxy, let out, etc.). Now groups purely
  by the action field using the backend's known actions: block/reject → red, pass → green, everything else
  (rdr/nat/binat) → blue. The label parameter was removed from _getColor() entirely.
 - rid/label relationship — The dedup logic in _updateChart remains because the relationship is not 1:1 (multiple
  rids can share a label, e.g. same rule on different interfaces). This is unchanged but is now clearer since the
  color logic is simpler.

src/opnsense/www/js/widgets/Firewall.js
swhite2: @GunnarLieb As stated above, the action is simply a string. We don't have to translate it to a string of exactly the same value to finx the index into `palettes`. See this comment https://github.com/opnsense/core/pull/9907/changes#r2903816331 for the right palettes array keys.

src/opnsense/www/js/widgets/Firewall.js
```diff
@@ -201,16 +232,17 @@ export default class Firewall extends BaseTableWidget {
                 parsing: false,
                 onClick: (event, elements, chart) => {
                     const i = elements[0].index;
-                    const rid = chart.data.datasets[0].rids[i];
-                    window.open(`/ui/diagnostics/firewall/log?rid=${rid}`);
+                    const clickedLabel = chart.data.labels[i];
+                    const rid = Object.keys(this.ridToLabel).find(r => this.ridToLabel[r] === clickedLabel);
```
swhite2: This is going to be an arbitrary match on the first of potentially duplicated labels.

src/opnsense/www/js/widgets/Firewall.js
```diff
@@ -147,25 +166,37 @@ export default class Firewall extends BaseTableWidget {
...
-    _updateChart(rid, label, count) {
+    _updateChart(rid, label, count, action) {
+        let dataset = this.chart.data.datasets[0];
         let labels = this.chart.data.labels;
-        let data = this.chart.data.datasets[0].data;
-        let rids = this.chart.data.datasets[0].rids;
+        let decodedLabel = $("<textarea/>").html(label).text();
+
+        // track which rids belong to which chart label
+        if (!this.ridToLabel) this.ridToLabel = {};
```
swhite2: I don't think we should start matching on labels as there is in fact a 1-to-1 relationship between rid and label. Even if labels are the same, we cannot just start summing counts based on the label, they're different rules.. This also causes a problem in the rule click handler, see https://github.com/opnsense/core/pull/9907/changes#r2905670181.

If we do want to differentiate the labels in some way, better include the interface in the label, then this function doesn't need to be changed.

swhite2 (changes requested):

src/opnsense/www/js/widgets/Firewall.js
Monviech: What happens to themes if the colors are set like this?

GunnarLieb: colors are taken out of Chart.js plugin with tableau.Classic10, palettes is defined for easier lookup logic

GunnarLieb:
clicking on chart to having filtering on live log doesn't work for me anyhow. Any advice how to fix that?

swhite2:
> clicking on chart to having filtering on live log doesn't work for me anyhow. Any advice how to fix that?

The click handler is bound to the pie chart segments

src/opnsense/www/js/widgets/Firewall.js
swhite2: @Monviech The colors are the same across themes.

@GunnarLieb FWIW, the tableau colors can be fetched directly from `Chart.colorschemes.tableau.Classic10`. You can probably shorten the code by using a function to define different indexing patterns into that namespace instead of harcoding the colors here.

GunnarLieb:
> > clicking on chart to having filtering on live log doesn't work for me anyhow. Any advice how to fix that?
> 
> The click handler is bound to the pie chart segments

sure, currently I get something  like this: /ui/diagnostics/firewall/log?rid=aa668c65edf34cc160b01aca25b69c2f but no filtering applies on Live View. (I reverted all my changes to be sure that I didn't break anything. Still doesn't work)

swhite2:
> sure, currently I get something like this: /ui/diagnostics/firewall/log?rid=aa668c65edf34cc160b01aca25b69c2f but no filtering applies on Live View. (I reverted all my changes to be sure that I didn't break anything. Still doesn't work)

I see what you mean, seems to be a separate issue as the URL params have changed slightly

@GunnarLieb https://github.com/opnsense/core/blob/master/src/opnsense/mvc/app/views/OPNsense/Firewall/filter_rule.volt#L253 is the correct pattern, care to add this in the PR?

GunnarLieb:
> > sure, currently I get something like this: /ui/diagnostics/firewall/log?rid=aa668c65edf34cc160b01aca25b69c2f but no filtering applies on Live View. (I reverted all my changes to be sure that I didn't break anything. Still doesn't work)
> 
> I see what you mean, seems to be a separate issue as the URL params have changed slightly
> 
> @GunnarLieb https://github.com/opnsense/core/blob/master/src/opnsense/mvc/app/views/OPNsense/Firewall/filter_rule.volt#L253 is the correct pattern, care to add this in the PR?

I tried to fix it, still doesn't work. My guess is:

1.   tableBuilt fires and calls filterVM.addFilter({ field, operator, value }), this adds the filter to the internal filter store
2. history.replaceState(null, '', location.pathname) strips the hash from the URL
3. But renderFilters() is never called, so the filter chip never appears in the UI

This is just a first idea, I can investigate, but would make it a separate PR

Monviech:
I'm pretty sure the url hash log filter works, we have recently added it. I just tested it too on 26.1.3:

image
image

GunnarLieb:
> I'm pretty sure the url hash log filter works, we have recently added it. I just tested it too on 26.1.3:
> 
> image image

Sorry I can't test really without setting up another machine (We are on Business latest)

Monviech:
Business doesnt have this change yet. For development it would be a good choice to set up a VM set to the Development firmware branch. Otherwise things like these happen.

GunnarLieb:
> @GunnarLieb The benefit of the already existing approach is a lot of contrast, which makes the chart segments easier to differentiate. I expect people are going to complain about such a change, as such I think it's best to default to the current color scheme, and present the semantic groups as a separate option. See https://docs.opnsense.org/development/frontend/dashboard.html#configurable-widgets or https://github.com/opnsense/core/blob/master/src/opnsense/www/js/widgets/Cpu.js#L60-L81 for a concrete example

@swhite2 so what about making it configuarable?
CleanShot 2026-03-10 at 10 14 44@2x
CleanShot 2026-03-10 at 10 16 33@2x
CleanShot 2026-03-10 at 10 17 17@2x
default would be: cycles through all Classic10 palette colors based on rule ID hash, maximizing visual differentiation

let me know what you think

swhite2:
@GunnarLieb Perfect, precisely what I had in mind :)

src/opnsense/www/js/widgets/Firewall.js
```diff
@@ -31,6 +31,65 @@ export default class Firewall extends BaseTableWidget {
...
+        }
+        let factor = ((hash % 10) - 5) * 0.08;
+        return this._shadeColor(base, factor);
+    }
+
+    async getWidgetOptions() {
+        return {
+            colorscheme: {
+                title: this.translations.colorscheme,
+                type: 'select_multiple',
```
swhite2:
```suggestion
                type: 'select',
```

src/opnsense/www/js/widgets/Firewall.js
```diff
@@ -31,6 +31,65 @@ export default class Firewall extends BaseTableWidget {
...
+                    { value: 'semantic', label: this.translations.semantic }
+                ],
+                default: ['contrast']
+            }
+        }
+    }
+
+    async onWidgetOptionsChanged(options) {
+        const config = await this.getWidgetConfig();
+        this.colorScheme = config.colorscheme.includes('semantic') ? 'semantic' : 'contrast';
```
swhite2:
```suggestion
        this.colorScheme = options.colorscheme;
```

src/opnsense/www/js/widgets/Firewall.js
```diff
@@ -175,17 +242,22 @@ export default class Firewall extends BaseTableWidget {
         const data = await this.ajaxCall('/api/diagnostics/interface/get_interface_names');
         this.ifMap = data;
 
+        const config = await this.getWidgetConfig();
+        this.colorScheme = config.colorscheme.includes('semantic') ? 'semantic' : 'contrast';
```
swhite2:
```suggestion
        this.colorScheme = config.colorscheme;
```

src/opnsense/www/js/widgets/Firewall.js
```diff
@@ -31,6 +31,65 @@ export default class Firewall extends BaseTableWidget {
...
+
+    _getColor(action, rid) {
+        let colors = Chart.colorschemes.tableau.Classic10;
+        let hash = parseInt(rid.slice(0, 8), 16);
+        let base;
+        if (this.colorScheme === 'semantic') {
+            let idx = this.actionColorIndex[action] ?? this.defaultColorIndex;
+            base = colors[idx];
+        } else {
+            base = colors[hash % colors.length];
```
swhite2: I'd like to preserve the old behavior (incremental colors), as using predictable colors like this will cause an uneven spread of picked colors.

Is it possible to dynamically enable the colorschemes plugin instead?

GunnarLieb: the plugin assigns colors at chart creation time and doesn't reassign for data points added later via chart.update(), so I would manual assign colors, would still use tableau.Classic10, look is almost identical, please check (colors after reload would be only stable in sematic mode).

src/opnsense/www/js/widgets/Firewall.js
```diff
@@ -31,6 +31,64 @@ export default class Firewall extends BaseTableWidget {
...
+                    { value: 'semantic', label: this.translations.semantic }
+                ],
+                default: 'contrast'
+            }
+        }
+    }
+
+    async onWidgetOptionsChanged(options) {
+        const config = await this.getWidgetConfig();
+        this.colorScheme = config.colorscheme;
```
swhite2:
```suggestion
    async onWidgetOptionsChanged(options) {
        this.colorScheme = options.colorscheme;
```

src/opnsense/www/js/widgets/Firewall.js
```diff
@@ -31,6 +31,64 @@ export default class Firewall extends BaseTableWidget {
...
+        let colors = Chart.colorschemes.tableau.Classic10;
+        if (this.colorScheme === 'semantic') {
+            let idx = this.actionColorIndex[action] ?? this.defaultColorIndex;
+            let base = colors[idx];
+            let hash = parseInt(rid.slice(0, 8), 16);
+            let factor = ((hash % 10) - 5) * 0.08;
+            return this._shadeColor(base, factor);
+        }
+        let rids = this.chart.data.datasets[0].rids;
+        return colors[rids.length % colors.length];
```
swhite2: if switching from `semantic` -> `contrast`, `rids.length` and `colors.length` are going to be the same for every invocation of `_getColor()`, this means every pie segment will be the same color.

src/opnsense/www/js/widgets/Firewall.js
```diff
@@ -31,6 +31,72 @@ export default class Firewall extends BaseTableWidget {
...
+            rids.forEach((rid, i) => {
+                if (this.colorScheme === 'semantic') {
+                    let idx = this.actionColorIndex[actions[i]] ?? this.defaultColorIndex;
+                    let base = palette[idx];
+                    let hash = parseInt(rid.slice(0, 8), 16);
+                    let factor = ((hash % 10) - 5) * 0.08;
+                    colors[i] = this._shadeColor(base, factor);
+                } else {
+                    colors[i] = palette[i % palette.length];
+                }
```
swhite2:
```suggestion
                colors[i] = this._getColor(actions[i], rid, i);
```

src/opnsense/www/js/widgets/Firewall.js
```diff
@@ -31,6 +31,72 @@ export default class Firewall extends BaseTableWidget {
...
+        let colors = Chart.colorschemes.tableau.Classic10;
+        if (this.colorScheme === 'semantic') {
+            let idx = this.actionColorIndex[action] ?? this.defaultColorIndex;
+            let base = colors[idx];
+            let hash = parseInt(rid.slice(0, 8), 16);
+            let factor = ((hash % 10) - 5) * 0.08;
+            return this._shadeColor(base, factor);
+        }
+        let rids = this.chart.data.datasets[0].rids;
+        return colors[rids.length % colors.length];
```
swhite2:
```suggestion
        const i = index ?? this.chart.data.datasets[0].rids.length;
        return colors[i % colors.length];
```

src/opnsense/www/js/widgets/Firewall.js
```diff
@@ -31,6 +31,72 @@ export default class Firewall extends BaseTableWidget {
...
+        let r = parseInt(hex.slice(1, 3), 16);
+        let g = parseInt(hex.slice(3, 5), 16);
+        let b = parseInt(hex.slice(5, 7), 16);
+        r = Math.min(255, Math.max(0, Math.round(r + (factor > 0 ? (255 - r) : r) * factor)));
+        g = Math.min(255, Math.max(0, Math.round(g + (factor > 0 ? (255 - g) : g) * factor)));
+        b = Math.min(255, Math.max(0, Math.round(b + (factor > 0 ? (255 - b) : b) * factor)));
+        return `#${((1 << 24) + (r << 16) + (g << 8) + b).toString(16).slice(1)}`;
+    }
+
+    _getColor(action, rid) {
```
swhite2:
```suggestion
    _getColor(action, rid, index = null) {
```

GunnarLieb:
Is there anything I should still do?

swhite2:
Hi, @GunnarLieb Sorry, other things came up, I'll have a look today

swhite2:
Merged, thank you! :)

PR #9912, opened 2026-03-09
Kea: add multi-arch boot-file-name support

h3krn:
This PR intends to add multi-arch boot-file-name support to kea.

I've tested it manually, by netbooting a x86_64 BIOS, x86_64 OVMF (UEFI64) and aarch64 OVMF Qemu virtual machine.

### DIFF
--- a/src/opnsense/mvc/app/models/OPNsense/Kea/KeaDhcpv4.php
+++ b/src/opnsense/mvc/app/models/OPNsense/Kea/KeaDhcpv4.php
@@ -147,10 +147,18 @@ class KeaDhcpv4 extends BaseModel
                 if ($key == 'static_routes') {
                     $value = implode(',', array_map('trim', explode(',', $value->getValue())));
                 }
-                $result[] = [
-                    'name' => $target_fieldname,
-                    'data' => (string)$value
-                ];
+                if (substr($key, 0, 15) == 'boot_file_name_') {
+                    $result[] = [
+                        'name'           => 'boot-file-name',
+                        'data'           => (string)$value,
+                        'client-classes' => [ substr($key, 15) ]
+                    ];
+                } else {
+                    $result[] = [
+                        'name' => $target_fieldname,
+                        'data' => (string)$value
+                    ];
+                }
             } elseif ($key == 'domain_name' && $defaults) {
                 $result[] = [
                     'name' => $target_fieldname,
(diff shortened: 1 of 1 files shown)


src/opnsense/mvc/app/models/OPNsense/Kea/KeaDhcpv4.xml
```diff
@@ -119,9 +119,21 @@
                     <tftp_server_name type="TextField">
                         <Mask>/^([^\n"])*$/u</Mask>
                     </tftp_server_name>
-                    <boot_file_name type="TextField">
```
Monviech, line 122: What happened to the boot_file_name here? What if somebody uses it already?

Monviech:
Can `client_classes` be made generic somehow by using this new framework for options? I could imagine it could be it's own grid, to be able to match any option a client sends.

To be more specific I want the same generic behavior as when using dnsmasq currently with set and match options.
https://docs.opnsense.org/manual/dnsmasq.html#dhcp-boot

src/opnsense/mvc/app/models/OPNsense/Kea/KeaDhcpv4.xml
h3krn: Good question! I did not think of that.
Maybe I should revert `boot_file_name_bios` to `boot_file_name` and try to use that as a fallback for non-matching archs?

h3krn:
> Can `client_classes` be made generic somehow by using this new framework for options? I could imagine it could be it's own grid, to be able to match any option a client sends.

I don't know. I'll have a  look.

Monviech:
I can help with the coding, I just want to see the kind of goal we want to reach.

Eg:
1. User defines client class
2. User defines option <- already possible
3. User can select client class(es) in an option to attach them to it
4. User attaches the option to the subnet <- already possible

Result:
Option is only sent out when client class matches.

Everything hex only for now.

Monviech:
I think I will whip a small concept up for this and see how usable this is.

Monviech:
@h3krn Do you think this would help you? https://github.com/opnsense/core/pull/9988

Right now only hex is supported but I feel like this can solve the issue generically. More encoding types can be added later to help.

Monviech:
Additionally to the last PR that I've merged, this one will address the rest of it:

https://github.com/opnsense/core/pull/9995

So you'll be able to match and send any option and also have usability with it.

Monviech:
Fixed via:
https://github.com/opnsense/core/pull/9995
and documentation:
https://github.com/opnsense/docs/pull/860

Issue #9872, opened 2026-02-28
Add dnsmasq dhcp no-ping option to reduce assignment delays


buckweet1980:
Looking for a UI option to add the no-ping option for dnsmasq dhcp services.

By default it tries to ping the address being assigned, causing a few second delay generally. 

In my use-case ICMP is blocked so ping will never succeed, would like an option to set this to reduce DHCP delays.

--no-ping is the option, disabled by default as defined in the docs below.

https://thekelleys.org.uk/dnsmasq/docs/dnsmasq-man.html

PR #9915, opened 2026-03-09
Add no-ping on dnsmasq

sopex:
Closes: https://github.com/opnsense/core/issues/9872

### DIFF
--- a/src/opnsense/mvc/app/controllers/OPNsense/Dnsmasq/forms/general.xml
+++ b/src/opnsense/mvc/app/controllers/OPNsense/Dnsmasq/forms/general.xml
@@ -193,7 +193,7 @@
         <help>Setting this will enable Router Advertisements for all configured DHCPv6 ranges with the managed address bits set, and the use SLAAC bit reset. To change this default, select a combination of the possible options in the individual DHCPv6 ranges. Keep in mind that this is a global option; if there are configured DHCPv6 ranges, RAs will be sent unconditionally and cannot be deactivated selectively. Setting Router Advertisement modes in DHCPv6 ranges will have no effect without this global option enabled.</help>
     </field>
     <field>
-        <id>dnsmasq.dhcp.no_ping</id>
+        <id>dnsmasq.dhcp.host_ping</id>
         <label>ICMP ping</label>
         <type>checkbox</type>
         <advanced>true</advanced>
--- a/src/opnsense/mvc/app/models/OPNsense/Dnsmasq/Dnsmasq.xml
+++ b/src/opnsense/mvc/app/models/OPNsense/Dnsmasq/Dnsmasq.xml
@@ -73,10 +73,10 @@
                 <MaximumValue>60</MaximumValue>
             </reply_delay>
             <enable_ra type="BooleanField"/>
-            <no_ping type="BooleanField">
+            <host_ping type="BooleanField">
                 <Required>Y</Required>
                 <Default>1</Default>
-            </no_ping>
+            </host_ping>
             <nosync type="BooleanField"/>
             <log_dhcp type="BooleanField">
                 <Constraints>
--- a/src/opnsense/service/templates/OPNsense/Dnsmasq/dnsmasq.conf
+++ b/src/opnsense/service/templates/OPNsense/Dnsmasq/dnsmasq.conf
@@ -95,7 +95,7 @@ dhcp-reply-delay={{dnsmasq.dhcp.reply_delay}}
 enable-ra
 {% endif %}
 
-{% if dnsmasq.dhcp.no_ping != '1' %}
+{% if dnsmasq.dhcp.host_ping != '1' %}
 no-ping
 {% endif %}
 


src/opnsense/mvc/app/controllers/OPNsense/Dnsmasq/forms/general.xml
```diff
@@ -192,6 +192,13 @@
         <type>checkbox</type>
         <help>Setting this will enable Router Advertisements for all configured DHCPv6 ranges with the managed address bits set, and the use SLAAC bit reset. To change this default, select a combination of the possible options in the individual DHCPv6 ranges. Keep in mind that this is a global option; if there are configured DHCPv6 ranges, RAs will be sent unconditionally and cannot be deactivated selectively. Setting Router Advertisement modes in DHCPv6 ranges will have no effect without this global option enabled.</help>
     </field>
+    <field>
+        <id>dnsmasq.dhcp.no_ping</id>
+        <label>Disable ICMP ping</label>
```
fichtner: Reversed options to enable the disable are notoriously know for being confusing/annoying. It's not a big deal in this case but since we have a model we can fix things like this right at the start of adding the feature. Not sure what @Monviech thinks about it.

src/opnsense/mvc/app/models/OPNsense/Dnsmasq/Dnsmasq.xml
```diff
@@ -1,7 +1,7 @@
 <model>
     <mount>/dnsmasq</mount>
     <description>Dnsmasq DNS and DHCP</description>
-    <version>1.0.8</version>
+    <version>1.0.9</version>
```
fichtner, line 4: We only need this change if we force a default/required one e.g. for inverted use.

Leaving the suggestion here for now depending on the flag being verbatim (don't need it) or inverted (need the version change).

```suggestion
    <version>1.0.8</version>
```

src/opnsense/mvc/app/controllers/OPNsense/Dnsmasq/forms/general.xml
Monviech: I want enabled properties wherever possible.

setting default to 1 in the model is simple.

sopex: Ok, I think reversing this would be even more confusing for someone looking for the specific setting from prior dnsmasq knowledge. On the other hand, it will be better for someone reading the available settings for the first time. You win some, you lose some. Let me know of your preference.

sopex: > I want enabled properties wherever possible.
> 
> setting default to 1 in the model is simple.

Will do

src/opnsense/service/templates/OPNsense/Dnsmasq/dnsmasq.conf
```diff
@@ -95,7 +95,7 @@ dhcp-reply-delay={{dnsmasq.dhcp.reply_delay}}
 enable-ra
 {% endif %}
 
-{% if dnsmasq.dhcp.no_ping == '1' %}
+{% if dnsmasq.dhcp.no_ping == '0' %}
```
fichtner: I'd prefer != '1' since it's harder to break if a migration fails although behaviour could be a bit undefined in practice.

sopex: Sure

src/opnsense/mvc/app/controllers/OPNsense/Dnsmasq/forms/general.xml
```diff
@@ -192,6 +192,13 @@
         <type>checkbox</type>
         <help>Setting this will enable Router Advertisements for all configured DHCPv6 ranges with the managed address bits set, and the use SLAAC bit reset. To change this default, select a combination of the possible options in the individual DHCPv6 ranges. Keep in mind that this is a global option; if there are configured DHCPv6 ranges, RAs will be sent unconditionally and cannot be deactivated selectively. Setting Router Advertisement modes in DHCPv6 ranges will have no effect without this global option enabled.</help>
     </field>
+    <field>
+        <id>dnsmasq.dhcp.no_ping</id>
```
fichtner: we need to fix the option name too ;)

sopex: Nope. This is the name of a specific dnsmasq command, leave it alone  ;)

src/opnsense/service/templates/OPNsense/Dnsmasq/dnsmasq.conf
Monviech: @fichtner if you want to guard against edge cases why not

`{% if dnsmasq.dhcp.no_ping|default('1') == '1' %}`

src/opnsense/service/templates/OPNsense/Dnsmasq/dnsmasq.conf
```diff
@@ -95,7 +95,7 @@ dhcp-reply-delay={{dnsmasq.dhcp.reply_delay}}
 enable-ra
 {% endif %}
 
-{% if dnsmasq.dhcp.no_ping != '1' %}
+{% if dnsmasq.dhcp.no_ping|default('1') == '1' %}
```
Monviech: Wait now I'm confused.

sopex: I just copied your logic, which was wrong, my bad. Will fix it

Monviech: I just asked a question hehe I didn't think too hard about it.

sopex: :)

fichtner: required fields don't really need `default()`, that's why they are required

src/opnsense/mvc/app/models/OPNsense/Dnsmasq/Dnsmasq.xml
```diff
@@ -73,6 +73,10 @@
                 <MaximumValue>60</MaximumValue>
             </reply_delay>
             <enable_ra type="BooleanField"/>
+            <no_ping type="BooleanField">
```
fichtner: ok last but not least "no_ping" makes no sense using it inverted here so it should be "host_ping" perhaps

@Monviech isn't sure we should merge it just yet. may take a bit of time to discuss internally.

sopex: I disagree for the same reason as above, force it if need be. 

Take your time.

fichtner: At some point we're likely going to write policies about the use of properties starting with disable/no. The model doesn't have to reflect configuration reality. The template is clear about the intent.

We can't fix other project's approaches and opinions but we know exactly why such disable/no flags come from and why they should (and can be) be avoided for our models (and forms) in particular.

Monviech: Here a reference where I was my own upstream, also developed "no-*" flags into that daemon, but in the model chose to use enabled and alter the names

https://github.com/opnsense/plugins/blob/78e3906a3ec5ea7e157cd33c17012fa0995ed85a/net/ndp-proxy-go/src/opnsense/service/templates/OPNsense/NdpProxy/ndp_proxy_go#L21-L26

What I and @fichtner want seems to align, yet in the dnsmasq model that grew over time there is not always the same consistency.

sopex: I understand your argument, and I am not defending the dnsmasq naming scheme. My argument is solely that since this is just the UI for a dnsmasq option, it's not an OPNsense option or an OPNsense DHCP, there should be some naming consistency between dnsmasq and OPNsense UI.

Anyway, it's obviously up to you, so I am changing it. Keep up the awesome work! Thank you! :)

Monviech (approved):
Looks good, thank you.

Monviech:
Merged thanks you :)

PR #9927, opened 2026-03-12
Follow up for dual-stack captive portal authorization in `CaptivePortal`

agoodkind:
Upstream PR: https://github.com/opnsense/core/pull/9745
Issue: https://github.com/opnsense/core/issues/8761

## What
This follow-up finishes the dual-stack Captive Portal path on top of `captive-portal-ipv6`.

It fixes client IP detection behind the local API proxy, restores IPv6 neighbor discovery for session expansion, and warm-starts roaming by immediately authorizing sibling addresses that are already known for the same MAC at login time.

## Why
I had some free time to run a fuller test matrix on top of the upstream branch, and that turned up a few gaps that were still visible in practice.

This branch tries to address the remaining review points from @AdSchellevis and the bugs that showed up during testing, especially around dual-stack login, IPv6-first login, and secondary IPv6 addresses after authentication.

## How
The API now resolves the real client address correctly when requests arrive through the local dispatcher. The `allow` path authorizes the connected address immediately, also authorizes sibling addresses that are already visible for the same MAC, and still leaves the background sync path in place for later discovery and cleanup.

## Test Matrix

| Case | Pre-auth | Auth path | Post-auth result |
|---|---|---|---|
| IPv4-only | Redirects to IPv4 portal host | IPv4 portal host | IPv4 egress works |
| Dual-stack, cold cache | IPv4 redirects to IPv4 portal host, IPv6 redirects to IPv6 portal host | IPv4 portal host | The login address works immediately. Sibling addresses join as they become visible. |
| Dual-stack, warm cache | IPv4 redirects to IPv4 portal host, IPv6 redirects to IPv6 portal host | IPv4 portal host | The login address works immediately. Already-known sibling IPv6 addresses work immediately. |
| Dual-stack, warm cache | IPv4 redirects to IPv4 portal host, IPv6 redirects to IPv6 portal host | IPv6 portal host | The login address works immediately. Already-known sibling IPv4 and IPv6 addresses work immediately. |
| IPv6-only preferred (DHCPv4 option 108) | Redirects to IPv6 portal host | IPv6 portal host | The login address works immediately. Already-known sibling IPv6 addresses work immediately. |
| NAT64 / DNS64 / PREF64 | IPv6-only preferred client uses DNS64 and RA `nat64prefix` | IPv6 portal host | IPv6-native and NAT64 HTTP access work after authentication. |
| Multi-address IPv6 | Session starts on one address | Background reconciliation | Newly observed SLAAC, privacy, and temporary IPv6 addresses join the same session later. |

Note: Right after link-up, clients often authenticate before they have exercised every candidate IPv6 source address. The login path can only warm-start what is already visible for the MAC, and the background reconciliation loop still picks up additional addresses later as the client starts using them. In CLI testing there was still one edge case where the first immediate post-login request from the DHCPv6 `/128` source needed a retry, while the privacy or SLAAC-style sibling addresses and the NAT64 path behaved as expected.

During testing I also confirmed that standard outbound NAT66 on `opnsense-dev` is required for downstream guest IPv6 egress to work consistently.

### DIFF
--- a/src/opnsense/mvc/app/controllers/OPNsense/CaptivePortal/Api/AccessController.php
+++ b/src/opnsense/mvc/app/controllers/OPNsense/CaptivePortal/Api/AccessController.php
@@ -306,7 +306,6 @@ class AccessController extends ApiControllerBase
                                 $userName,
                                 $clientIp,
                                 $authServerName,
-                                $cpZone->roaming->getValue()
                             ]
                         );
                         $CPsession = json_decode($CPsession, true);
--- a/src/opnsense/scripts/captiveportal/allow.py
+++ b/src/opnsense/scripts/captiveportal/allow.py
@@ -41,30 +41,19 @@ parser.add_argument('--username', help='username', type=str, required=True)
 parser.add_argument('--zoneid', help='zone number to allow this user in', type=str, required=True)
 parser.add_argument('--authenticated_via', help='authentication source', type=str)
 parser.add_argument('--ip_address', help='source ip address', type=str)
-parser.add_argument('--roaming', help='roaming allowed for this client', type=str)
 args = parser.parse_args()
 
-arp = ARP()
-db = DB()
 
-arp_entry = arp.get_by_ipaddress(args.ip_address)
-response = db.add_client(
+arp_entry = ARP().get_by_ipaddress(args.ip_address)
+response = DB().add_client(
     zoneid=args.zoneid,
     authenticated_via=args.authenticated_via,
     username=args.username,
     ip_address=args.ip_address,
     mac_address=arp_entry['mac'] if arp_entry is not None else None
 )
-
-session_ips = {args.ip_address}
-if args.roaming and arp_entry is not None:
-    known_ips = set(arp.get_all_addresses_by_mac(arp_entry['mac']))
-    known_ips.add(args.ip_address)
-    session_ips = db.update_roaming_ips(args.zoneid, response['sessionId'], list(known_ips))
-
-for ip_address in session_ips:
-    PF.add_to_table(zoneid=args.zoneid, address=ip_address)
-    IPFW.add_accounting(ip_address)
+PF.add_to_table(zoneid=args.zoneid, address=args.ip_address)
+IPFW.add_accounting(args.ip_address)
 
 response['clientState'] = 'AUTHORIZED'
 print(ujson.dumps(response))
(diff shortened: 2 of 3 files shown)


swhite2:
@agoodkind FWIW, the logic in `allow.py` will be ditched in the original PR as there are concerns that pfctl locks up during mass login attempts, especially if the background process is not serialized with this process. Better keep it serialized in the background process and accept the potential delay: https://github.com/opnsense/core/pull/9745/changes/5114d1e0576c482b5bc2740415c59d9a4d20e574

src/opnsense/scripts/captiveportal/lib/arp.py
```diff
@@ -44,7 +44,7 @@ def reload(self):
 
         # fetch addresses, no IPv6 if hostwatch disabled
         out = ujson.loads(subprocess.run(
-            ['/usr/local/opnsense/scripts/interfaces/list_hosts.py','--last-seen-window', '86400', '-v'],
+            ['/usr/local/opnsense/scripts/interfaces/list_hosts.py', '--ndp', '--last-seen-window', '86400', '-v'],
```
swhite2, line 47: `--ndp` was left out intentionally for performance reasons

swhite2:
I'll start testing with a slightly modified version of this patch (see comments above)

src/opnsense/service/templates/OPNsense/Captiveportal/lighttpd-zone.conf
```diff
@@ -98,11 +122,10 @@ server.bind              = "0.0.0.0"
 #### bind to port
 server.port              = {{ cp_zone_item.zoneid|int + 8000  }}
 
-#### IPv6 socket binding for main HTTPS port
 $SERVER["socket"] == "[::]:{{ cp_zone_item.zoneid|int + 8000 }}" {
-	$HTTP["host"] !~ "(.*{{cp_zone_item.redirect_host_match}}.*)" {
+	$HTTP["host"] !~ "(.*{{cp_zone_item.redirect_host_match_v6}}.*)" {
```
swhite2, line 126: If there is no static IPv6 configured and no Hostname set, but there is IPv6 through e.g. Identity Association, this is going to render the IPv4 address, where possibly clients are IPv6-only.

It's pretty difficult to validate this through the template, but the question is if this is a practical problem or this should be documented instead (i.e. use a static address OR make sure you provide AAAA records and use a hostname)

PR #9929, opened 2026-03-12
firewall: fix regression in 8554581eac so alias content summary is shown

fichtner:
The "description" is a summary so change the underlying code accordingly to avoid future misinterpretations.


Monviech (approved):
Thank you that looks great and works :)

Monviech:
PR: https://forum.opnsense.org/index.php?topic=51246.0

Issue #9861, opened 2026-02-25
Firewall: Import/Export add header check to .csv


Daniii438:
**Important notices**

Before you add a new report, we ask you kindly to acknowledge the following:

- [x] I have read the contributing guide lines at https://github.com/opnsense/core/blob/master/CONTRIBUTING.md
- [x] I am convinced that my issue is new after having checked both open and closed issues at https://github.com/opnsense/core/issues?q=is%3Aissue

**Is your feature request related to a problem? Please describe.**

When I exported the firewall rules to import them into the firewall [new] rules, I accidentally selected a previously exported backup-config.xml file, which I had created before upgrading to 26.x, and imported that into the firewall rules.

This was unfortunately accepted as valid, and OPNsense imported about 80,000 rules instead of aborting the process.

As a result, the firewall became unresponsive due to the excessive number of rules. I had to restore the system from a backup, as it could not handle 80,000 rules.

A clear and concise description of what the problem is including your motivation for the request,
i.e. "For the purpose of [...] I am missing a solution that will [...]."

This lagged my firewall, and I had to recover from a backup, as it could not handle 80,000 rules.

(e.g. I would like an input field in the `/ui/firewall/alias` which would add .... to ....)

Add a check before importing to verify that the header of the .csv firewall rules file is present in the uploaded file.
Otherwise, show an error indicating that the required header is missing and/or add a checkmark to "override this check".

@uuid;enabled;statetype;state-policy;sequence;action;quick;interfacenot;interface;direction;ipprotocol;protocol;icmptype;icmp6type;gateway;replyto;disablereplyto;log;allowopts;nosync;nopfsync;statetimeout;max-src-nodes;max-src-states;max-src-conn;max;max-src-conn-rate;max-src-conn-rates;overload;adaptivestart;adaptiveend;prio;set-prio;set-prio-low;tag;tagged;tcpflags1;tcpflags2;categories;sched;tos;shaper1;shaper2;description;source_not;source_net;source_port;destination_not;destination_net;destination_port

A clear and concise description of any alternative solutions or features you considered.

Being more attentive and not selecting the wrong file would be a good start on my end.

Add any other context or screenshots about the feature request here or links to relevant forum thread or similar

sopex:
Sounds fun :)

PR #9933, opened 2026-03-12
Fix csv imports' carelessness 

sopex:
Fixes: https://github.com/opnsense/core/issues/9861
And many more CSV imports.

In my testing, it works well with all the CSV imports. It might be a bit simplistic, but it is great for blocking careless mistakes.

### DIFF
--- a/src/opnsense/mvc/app/controllers/OPNsense/Base/ApiMutableModelControllerBase.php
+++ b/src/opnsense/mvc/app/controllers/OPNsense/Base/ApiMutableModelControllerBase.php
@@ -697,7 +697,6 @@ abstract class ApiMutableModelControllerBase extends ApiControllerBase
                         sprintf(gettext("Missing mandatory fields: %s"), implode(", ", $missing_keys)), gettext("CSV Import Failed")
                     );
                 }
-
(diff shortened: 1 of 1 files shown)


AdSchellevis:
let's not cut off partial imports, it's perfectly fine to only offer required values as long as it passes validations.

sopex:
@AdSchellevis 

IMO, validations should not allow the config backup to be imported as dhcp entries or firewall rules. And firewall rules should not be allowed to be imported as dhcp entries :) 

Thus, a more comprehensive validation could be created per instance to catch more things but this PR prevents the more crude errors without over-validating everything, in a sense, promoting what you suggest.

Also, the error message tells you what you are missing. If you know what you are doing you just add them manually.

AdSchellevis:
If the model supports empty records to be added, it's easy to render new (empty) records indeed, which might be less practical, but also a bit part of the deal I'm afraid. I don't mind looking at options to at least require something being inserted, the question just is what that should be. requiring keys doesn't really make it less fragile as in time a lot of these things have similar keys like a uuid.

sopex:
If you believe that more than one CSV uses exactly the same keys then this doesn't make much sense. In my experience, they don't, but haven't really looked it up.

Requiring something to be inserted is not a bad idea, but it will break many already saved files :(

Issue #9728, opened 2026-02-04
Adding a notes / notepad / scratchpad type widget on the dashboard?


Jas0n99:
**Important notices**

Before you add a new report, we ask you kindly to acknowledge the following:

- [X] I have read the contributing guide lines at https://github.com/opnsense/core/blob/master/CONTRIBUTING.md
- [X] I am convinced that my issue is new after having checked both open and closed issues at https://github.com/opnsense/core/issues?q=is%3Aissue

**Is your feature request related to a problem? Please describe.**

I find myself sometimes needing to make small notes using opnsense and right now I end up just saving a text file on my PC desktop. It would be nice if on the dashboard there was a way to add a widget that you could enter basic plain text notes (that saves on the disk somewhere) so that they are always available for future reference.

**Describe the solution you like**

I've found the Proxmox "Notes" section extremely helpful on the web dashboard summary for containers / vms just to jot down important bits of info. Having something similar on the opnsense dashboard would be a great feature and likely require minimal effort to implement.

**Describe alternatives you considered**

Right now I just save a text document on my PC, but that is not optimal as I often forget about it or what it contains. Even though it's just a few small snippets, they are immense time savers or important file locations.

**Additional context**

Example of how Proxmox implements a notes feature:
<img width="362" height="350" alt="Image" src="https://github.com/user-attachments/assets/a6ca40ee-bc14-448f-b9f2-7d7a4659f855" />

PR #9916, opened 2026-03-10
Auto light-dark mode

sopex:
Closes: https://github.com/opnsense/core/issues/8264
Closes: https://github.com/opnsense/core/issues/9805

In my testing, it works great!

Let me know of any ideas for improvement.

PR #9936, opened 2026-03-12
dashboard: add Notepad widget

sopex:
Closes: https://github.com/opnsense/core/issues/9728

If anyone wants to test it :)

image

### DIFF
--- a/src/opnsense/www/js/widgets/Notes.js
+++ b/src/opnsense/www/js/widgets/Notes.js
@@ -27,7 +27,6 @@
 export default class Notes extends BaseWidget {
     constructor(config) {
         super(config);
-        this.titleVisible = false;
         this.configurable = true;
     }
 
--- a/src/opnsense/www/js/widgets/BaseWidget.js
+++ b/src/opnsense/www/js/widgets/BaseWidget.js
@@ -106,6 +106,10 @@ class BaseWidget {
         return {};
     }
 
+    get dialogTitle() {
+        return null;
+    }
+
     getMarkup() {
         return $("");
     }


Monviech:
It is helpful and admirable that you want to work on so many things.
But please slow down a bit, if somebody has something assigned it would be courteous to ask first in the issue what the strategy is.

sopex:
> It is helpful and admirable that you want to work on so many things. But please slow down a bit, if somebody has something assigned it would be courteous to ask first in the issue what the strategy is.

Under no circumstances should you take my commits as a pressure to act on them. I understand you have your own, much more logical priorities. If there is a similar issue/ request I just open a PR. 

Anyway, I have missed the mark here with @swhite2 I apologize. It's just that due to the way of our communication, pestering team members on issues/ PR pages IMO is also meh meh so I try to avoid it.

Monviech:
Don't worry, you are not pestering when you ask questions. It helps with collaboration and planning. :)

sopex:
@swhite2 I once again apologize for this. How do you want to play it? Shall I finish it and tag you for review? Shall I let you do it from the begining your way? Do you have any input? Thanks!

swhite2:
@sopex No worries, to be honest this slipped my mind a bit. Given the scope, go ahead and finish it and tag me for review, I don't expect a lot of code to shift afterwards, but we'll see.

sopex:
@swhite2 Let me know :)

swhite2:
Couple of points:

- Let's rename it "Notes"
- I don't want a save button in there, so it's best to combine this with the edit mode of the widgets. Move the editable textbox to the rendered options form when "edit" is clicked, and the "notes" will just be rendered as text - no textbox around it. This will require a new type in https://github.com/opnsense/core/blob/master/src/opnsense/www/js/opnsense_widget_manager.js#L779-L817 (much like is the case with our form xmls).
- `this.titleVisible = false;` will probably work better for this widget, much like the picture widget.

sopex:
> Couple of points:
> 
> * Let's rename it "Notes"
> * I don't want a save button in there, so it's best to combine this with the edit mode of the widgets. Move the editable textbox to the rendered options form when "edit" is clicked, and the "notes" will just be rendered as text - no textbox around it. This will require a new type in https://github.com/opnsense/core/blob/master/src/opnsense/www/js/opnsense_widget_manager.js#L779-L817 (much like is the case with our form xmls).
> * `this.titleVisible = false;` will probably work better for this widget, much like the picture widget.

Thank you for your comments.

After some extensive testing, I am afraid that removing the save button makes this counterintuitive to use, especially on mobile.

That being said, I don't have any better ideas for removing the save button, only auto-saving, which might have issues of its own, so I will try to have something working as soon as possible.

sopex:
@swhite2 Ready

You were absolutely right, it's much better this way!

image

image

swhite2:
@sopex This does look a lot better indeed :)

Neat how it now streamlines the content into the dashboard config with the new option type.

sopex:
> @sopex This does look a lot better indeed :)
> 
> Neat how it now streamlines the content into the dashboard config with the new option type.

Indeed it's cool :)

src/opnsense/www/js/widgets/Notes.js
```diff
@@ -0,0 +1,68 @@
...
+ * INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN
+ * CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE)
+ * ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
+ * POSSIBILITY OF SUCH DAMAGE.
+ */
+
+export default class Notes extends BaseWidget {
+    constructor(config) {
+        super(config);
+        this.titleVisible = false;
```
swhite2: It seems this causes the side effect of not showing the title + edit symbol as soon as the widget is newly added to the dashboard, I'll have to look into that.

src/opnsense/www/js/widgets/Notes.js
```diff
@@ -0,0 +1,68 @@
...
+        this.configurable = true;
+    }
+
+    get dialogTitle() {
+        return this.translations.titleedit;
+    }
+
```
swhite2, line 42:
```suggestion

    getGridOptions() {
        return {
            minH: 150,
            minW: 4
        }
    }

```

Something like this may also be practical to not present a tiny box as soon an emptynotes widget is rendered. For context: gridstack seems to interpret the unit value for W and H differently

src/opnsense/www/js/widgets/Notes.js
sopex: Haven't seen that. I will check it later.

src/opnsense/www/js/widgets/Notes.js
sopex: I agree on principle but the width seems excessive. Probably a minW 2 will be fine.

swhite2: Sure, this was merely a suggestion on how, not what :)

sopex:
@swhite2 Resolved

src/opnsense/www/js/widgets/Notes.js
```diff
@@ -0,0 +1,74 @@
...
+        this.configurable = true;
+    }
+
+    getGridOptions() {
+        return {
+            minH: 82,
+        }
+    }
+
+    get dialogTitle() {
```
swhite2, line 39: This pattern isn't used anywhere else so let's keep it `getDialogTitle()`, though I'd prefer a more descriptive name such as `getOptionsDialogTitle()`. Also, specify it in the `BaseWidget` which returns `null`.

src/opnsense/www/js/widgets/Notes.js
```diff
@@ -0,0 +1,74 @@
...
+ * INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN
+ * CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE)
+ * ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
+ * POSSIBILITY OF SUCH DAMAGE.
+ */
+
+export default class Notes extends BaseWidget {
+    constructor(config) {
+        super(config);
+        this.titleVisible = false;
```
swhite2: Tried out the PR, I think keeping the title visible may be better in hindsight.

src/opnsense/www/js/opnsense_widget_manager.js
```diff
@@ -414,6 +414,7 @@ class WidgetManager  {
                                     $('.link-handle').hide();
                                     $('.close-handle').show();
                                     $('.edit-handle').show();
+                                    $('.title-invisible').css('display', '');
```
swhite2, line 417: nice one

sopex: Thank you!

src/opnsense/www/js/widgets/Notes.js
sopex: 😡😡😡
No worries

src/opnsense/www/js/widgets/Notes.js
sopex: Done

sopex:
@swhite2 
No worries, merge my other 2 PRs today, and I will forgive you. #10063 #9608 :)
Bonus points if you have any good ideas for the opnsense-auto problem #9916

Have a great rest of your day!

Issue #9140, opened 2025-08-26
unbound: revamp overrides grid using Tabulator Group-by


fichtner:
**Important notices**

Before you add a new report, we ask you kindly to acknowledge the following:

- [x] I have read the contributing guide lines at https://github.com/opnsense/core/blob/master/CONTRIBUTING.md
- [x] I am convinced that my issue is new after having checked both open and closed issues at https://github.com/opnsense/core/issues?q=is%3Aissue

**Feature**

Master-detail views have UX limits and can be difficult for new users. Now that we have Tabulator it's possible to build a better version for a mostly drop-in replacement.

While here discuss the name. ;)

PR #9954, opened 2026-03-13
Unbound: consolidate override aliases into tree view

swhite2:
This also adds the `aliases` volatile field, which renders a tokenized list of aliases into the host override form. "quick" additions/deletions in this field are possible and reflected to the model through some custom controller logic. "quick" aliases are reflected back in the grid in the tree view and one can edit them there to i.e. add descriptions.

Fixes https://github.com/opnsense/core/issues/9140

### DIFF
--- a/src/opnsense/mvc/app/views/OPNsense/Unbound/overrides.volt
+++ b/src/opnsense/mvc/app/views/OPNsense/Unbound/overrides.volt
@@ -28,7 +28,7 @@
 $( document ).ready(function() {
     let selectedHostOverride = null;
     const commandOverride = {
-        dialog: "{{formGridHostAlias['edit_dialog_id']}}",
+        dialog: "DialogHostAlias",
         get: '/api/unbound/settings/get_host_alias/',
         set: '/api/unbound/settings/set_host_alias/',
         add: '/api/unbound/settings/add_host_alias/',
@@ -203,4 +203,4 @@ $( document ).ready(function() {
 </div>
 {{ partial('layout_partials/base_apply_button', {'data_endpoint': '/api/unbound/service/reconfigure', 'data_service_widget': 'unbound'}) }}
 {{ partial("layout_partials/base_dialog",['fields':formDialogHostOverride,'id':formGridHostOverride['edit_dialog_id'],'label':lang._('Edit Host Override')])}}
-{{ partial("layout_partials/base_dialog",['fields':formDialogHostAlias,'id':formGridHostAlias['edit_dialog_id'],'label':lang._('Edit Host Override Alias')])}}
+{{ partial("layout_partials/base_dialog",['fields':formDialogHostAlias,'id':'DialogHostAlias','label':lang._('Edit Host Override Alias')])}}
--- a/src/opnsense/mvc/app/controllers/OPNsense/Unbound/OverridesController.php
+++ b/src/opnsense/mvc/app/controllers/OPNsense/Unbound/OverridesController.php
@@ -40,6 +40,5 @@ class OverridesController extends IndexController
         $this->view->formGridHostOverride = $this->getFormGrid('dialogHostOverride');
 
         $this->view->formDialogHostAlias = $this->getForm('dialogHostAlias');
-        $this->view->formGridHostAlias = $this->getFormGrid('dialogHostAlias');
     }
 }


src/opnsense/mvc/app/controllers/OPNsense/Unbound/Api/SettingsController.php
```diff
@@ -40,6 +40,105 @@ class SettingsController extends ApiMutableModelControllerBase
...
+        ];
+    }
+
+    private function updateHostAliases($hostUUID)
+    {
+        $result = ['status' => 'failed'];
+
+        $mdl = $this->getModel();
+        $host = $this->request->getPost('host') ?? [];
+        $host['uuid'] ??= $hostUUID;
```
fichtner: "finally, some respect" *ziodberg noise*

src/opnsense/mvc/app/controllers/OPNsense/Unbound/forms/dialogHostOverride.xml
```diff
@@ -79,8 +88,19 @@
...
+        <style>tokenize</style>
+        <allownew>true</allownew>
+        <help>Quickly add aliases</help>
+        <grid_view>
+            <visible>false</visible>
+        </grid_view>
+    </field>
+    <field>
+        <id>host.aliascount</id>
+        <label>Alias count</label>
```
fichtner: this nice idea seems almost useless now?

src/opnsense/mvc/app/models/OPNsense/Unbound/FieldTypes/AliasReflector.php
```diff
@@ -26,25 +26,35 @@
...
  */
 
-namespace OPNsense\Kea\Api;
+namespace OPNsense\Unbound\FieldTypes;
 
-use OPNsense\Base\ApiMutableModelControllerBase;
+use OPNsense\Base\FieldTypes\CSVListField;
 
-class DdnsController extends ApiMutableModelControllerBase
+class AliasReflector extends CSVListField
```
fichtner: Why CSVListField?

src/opnsense/mvc/app/controllers/OPNsense/Unbound/Api/SettingsController.php
swhite2: zoidburg-woop

src/opnsense/mvc/app/controllers/OPNsense/Unbound/forms/dialogHostOverride.xml
swhite2: Yep, it's gone

src/opnsense/mvc/app/models/OPNsense/Unbound/FieldTypes/AliasReflector.php
swhite2: It seems I've forgotten about `BaseSetField`

fichtner: yes please

src/opnsense/mvc/app/models/OPNsense/Unbound/FieldTypes/AliasReflector.php
```diff
@@ -28,26 +28,33 @@
...
     public function actionPostLoadingEvent()
     {
-        if ($this->refcountcache === null) {
-            $this->refcountcache = [];
-            foreach ($this->getParentModel()->aliases->alias->iterateItems() as $node) {
-                $uuid = $node->host->getValue();
-                if (!isset($this->refcountcache[$uuid])) {
-                    $this->refcountcache[$uuid] = 0;
+        $uuid = $this->getParentNode()->getAttribute('uuid') ?? '';
+        $this->setValue(implode(',', array_map(function ($node) {
```
fichtner: at the risk of making this a side project it looks like `setValues()` might be useful for BaseSetField at least because the separator is configurable and you're hardcoding it here. A simple fix is using the variable for it.

fichtner: setValues also has the benefit of dealing with empty values more gracefully -> ['', ''] becomes [] / '' not ','

swhite2: @fichtner how about https://github.com/opnsense/core/pull/9954/commits/54ed928af3b5badaf6658d9636c5f79cc58877da?

fichtner: good enough for now.. I can adjust this post-merge (and also move to BaseField) for avoiding code breaking changes when we move fields and the backend code stays at setValues (similar to getValues).

src/opnsense/mvc/app/controllers/OPNsense/Unbound/Api/SettingsController.php
```diff
@@ -40,6 +40,103 @@ class SettingsController extends ApiMutableModelControllerBase
...
+
+    private function updateHostAliases($hostUUID)
+    {
+        $result = ['status' => 'failed'];
+
+        if (!$this->request->isPost()) {
+            return $result;
+        }
+
+        $mdl = $this->getModel();
```
AdSchellevis: Introduces locking issues, but I think we can simplify this whole construct a lot to be honest. Ideally you want to move the set operation to the model, so others can reuse the behavior easily when needed, question in that case is just where we can do that with least impact.

The trust model fixes something similar in the controller layer using the `setBaseHook` construct, but the main reason for that is that the operation is quite expensive when generating new certs:

https://github.com/opnsense/core/blob/770480715b15e1645c98a3eabade878e8f757e74/src/opnsense/mvc/app/controllers/OPNsense/Trust/Api/CaController.php#L55

Some inspiration we may collect from the DNSmasq model, which detects the initial set from our config xml via a type switch:

https://github.com/opnsense/core/blob/770480715b15e1645c98a3eabade878e8f757e74/src/opnsense/mvc/app/models/OPNsense/Dnsmasq/FieldTypes/DomainIPField.php#L40

Although I haven't given it much thought yet, I do expect if we can add some simple "magic" to know it's being set as part of new data, in which case we need to track the children.

src/opnsense/mvc/app/controllers/OPNsense/Unbound/Api/SettingsController.php
```diff
@@ -186,7 +283,26 @@ public function toggleForwardAction($uuid, $enabled = null)
...
+            $host['isAlias'] = false;
+            $uuid = $host['uuid'];
+            foreach ($mdl->getHostAliases($uuid) as $node) {
+                $child['uuid'] = $node->getAttribute('uuid');
+                $child['isAlias'] = true;
+                foreach (['enabled', 'hostname', 'domain', 'description'] as $key) {
+                    $child[$key] = $node->$key->getValue();
+                }
+                unset($child['aliases']);
+                $host['_children'] ??= [];
```
AdSchellevis, line 203: nice one

src/opnsense/mvc/app/controllers/OPNsense/Unbound/Api/SettingsController.php
```diff
@@ -196,7 +312,16 @@ public function getHostOverrideAction($uuid = null)
 
     public function addHostOverrideAction()
     {
-        return $this->addBase('host', 'hosts.host');
+        /* overlay empty aliases here so they're not reflected back in the volatile 'aliases' field */
+        $result = $this->addBase('host', 'hosts.host', ['aliases' => '']);
+        if ($result['result'] === 'saved' && !empty($result['uuid'])) {
+            $aliasUpdate = $this->updateHostAliases($result['uuid']);
```
AdSchellevis: if we are able to push it back in the model, we can stick with the default `addBase`, but same is the case if we're hooking `setBaseHook`

src/opnsense/mvc/app/models/OPNsense/Base/FieldTypes/BaseSetField.php
```diff
@@ -112,4 +112,9 @@ public function getValues(): array
             return !!strlen($k);
         }));
     }
+
+    public function setValues($values)
```
AdSchellevis:
```suggestion
    public function setValues(array $values)
```

Might be better to make this strict in this case.

src/opnsense/mvc/app/views/OPNsense/Unbound/overrides.volt
```diff
@@ -26,19 +26,87 @@
...
+        commands: {
+            "edit": {filter: (cell) => !cell.getData()['isAlias']},
+            "copy": {filter: (cell) => !cell.getData()['isAlias']},
+            "delete": {filter: (cell) => !cell.getData()['isAlias']},
+            "add-alias": {
+                filter: (cell) => !cell.getData()['isAlias'],
+                method: function (event, cell) {
+                    const data = cell.getData();
+                    selectedHostOverride = data;
+                    grid_hosts.bootgrid('command_add', event, cell, commandOverride);
```
AdSchellevis, line 55: Would it be possible to refer to something like `this` here? The `grid_hosts` looks a bit suspicious as this is all inside a block that assigns `grid_hosts` for the first time.

swhite2: this would require ditching the jquery click handler logic to properly bind the class instance in https://github.com/opnsense/core/blob/master/src/opnsense/www/js/opnsense_bootgrid.js#L964-L966, which we can't do because of backwards compatibility.

`grid_hosts` is guaranteed to exist here since both method binding and execution, triggered by the data fetch, happen after grid initialization.

src/opnsense/mvc/app/controllers/OPNsense/Unbound/Api/SettingsController.php
swhite2: > Introduces locking issues

Isn't this lazy-loaded? https://github.com/opnsense/core/blob/master/src/opnsense/mvc/app/controllers/OPNsense/Base/ApiMutableModelControllerBase.php#L222-L228

swhite2: @AdSchellevis https://github.com/opnsense/core/pull/9954/changes/64268210feb73350e5fb96f0cdf85373676cc79e should roughly be what you had in mind :)

src/opnsense/mvc/app/models/OPNsense/Unbound/FieldTypes/AliasReflector.php
```diff
@@ -0,0 +1,117 @@
...
+            $concat = !empty($hostname) ? $hostname . '.' : '';
+            $concat .= !empty($domain) ? $domain : $this->getParentNode()->domain->getValue();
+
+            return $concat;
+        }, $this->getParentModel()->getHostAliases($uuid));
+        $this->setValues($this->postLoadValue);
+    }
+
+    public function setValue($value)
+    {
```
AdSchellevis, line 71: @swhite2 a bit late to the party, but I suddenly remembered you might want to bail here with something like this:

```
if (is_a($value, \SimpleXMLElement::class)) {
     return parent::setValue($value);
}
```

During construction phase, you probably don't want this logic to execute.

swhite2: As discussed internally, this can't happen for volatile fields, however, this did trigger me to realize there's a cheaper bail-out option here: https://github.com/opnsense/core/commit/dde6f27f31840b4ec2462bfaf499b6fab96f7efd

Issue #9942, opened 2026-03-13
kea: MVP for custom options


fichtner:
**Important notices**

Before you add a new report, we ask you kindly to acknowledge the following:

- [x] I have read the contributing guide lines at https://github.com/opnsense/core/blob/master/CONTRIBUTING.md
- [x] I am convinced that my issue is new after having checked both open and closed issues at https://github.com/opnsense/core/issues?q=is%3Aissue

**Is your feature request related to a problem? Please describe.**

Feature partity with isc-dhcp and dnsmasq requires a "minimum viable product" for Kea. Previous attempts are

https://github.com/opnsense/core/pull/7361
https://github.com/opnsense/core/tree/kea_custom_opt_pr7361

but have so far failed to meet project goals in terms of maintenance and ease of use (also in relation to how Kea configuration works).

**Describe the solution you like**

Implement a small framework that doesn't get lost in the details and is maintainable and extendable going forward.

**Describe alternatives you considered**

N/A

**Additional context**

N/A

Monviech:
To avoid the complexity of the typed option model used by Kea DHCP, DHCP options can be serialized as raw hexadecimal payloads.

When csv-format is set to false, Kea does not interpret the option value according to its internal type definitions. Instead, it converts the hexadecimal string directly into the raw byte payload of the DHCP option.

This allows a completely generic model:

`user input (hex) → Kea → DHCP option payload`

No option typing or conversion logic is required in the GUI.

It should be something for advanced usecases after all and pushes the whole issue to the user to know what they want to send.

Common options already have the easy input in the GUI, everything more advanced can just be hex.

https://kea.readthedocs.io/en/kea-2.2.0/arm/dhcp4-srv.html#standard-dhcpv4-options

section about ``csv-format``

```
Options can also be configured as hexadecimal values. If csv-format is set to false, option data must be specified as a hexadecimal string. The following commands configure the domain-name-servers option for all subnets with the following addresses: 192.0.3.1 and 192.0.3.2. Note that csv-format is set to false.

"Dhcp4": {
    "option-data": [
        {
            "name": "domain-name-servers",
            "code": 6,
            "space": "dhcp4",
            "csv-format": false,
            "data": "C0 00 03 01 C0 00 03 02"
        },
        ...
    ],
    ...
}
Kea supports the following formats when specifying hexadecimal data:

[...]

PR #9958, opened 2026-03-13
Services: Kea: DHCPv4: allow sending any DHCP option as raw hex or string payload

Monviech:
Fixes: https://github.com/opnsense/core/issues/9942

https://kea.readthedocs.io/en/kea-2.0.1/arm/dhcp4-srv.html

This uses ```csv-false``` to force kea to send option data as hexadecimal value, instead of enforcing types on them.

This means theoretically you can now send any byte coded option out, meaning the user can construct whatever value they need.

```
            {
                "id": 3,
                "subnet": "192.168.50.0\/24",
                "next-server": "",
                "match-client-id": false,
                "option-data": [
                    {
                        "data": "192.168.30.1"
                    },
                    {
                        "name": "routers",
                        "data": "192.168.30.1"
                    },
                    {
                        "name": "domain-name",
                        "data": "ad.pischem.com"
                    },
                    {
                        "name": "ntp-servers",
                        "data": "192.168.30.1"
                    },
                    {
                        "code": 6,
                        "csv-format": false,
                        "data": "01010101",
                        "always-send": false
                    },
                    {
                        "code": 4,
                        "csv-format": false,
                        "data": "08080808",
                        "always-send": false
                    }
                ],
                "pools": [
                    {
                        "pool": "192.168.50.100-192.168.50.110"
                    }
                ],
                "reservations": []
            }
```

There is also another weird possibility, a very specific format of ```data``` can automatically convert strings into hex (bytes), kea handles this conversion internally.

```"data": "'convert this text to binary'"```

```
                    {
                        "code": 1,
                        "csv-format": false,
                        "data": "'example.com'",
                        "always-send": false
                    },
```

### DIFF
--- a/src/opnsense/mvc/app/controllers/OPNsense/Kea/forms/dialogOption4.xml
+++ b/src/opnsense/mvc/app/controllers/OPNsense/Kea/forms/dialogOption4.xml
@@ -9,7 +9,7 @@
         <id>option.encoding</id>
         <label>Encoding</label>
         <type>dropdown</type>
-        <help>Choose between auto converting the option data to binary, or providing binary data yourself. For encapsulated and other complex options, constructing your own binary payload can be a requirement.</help>
+        <help>Choose between auto-converting string data to binary, or providing hexadecimal data yourself. For encapsulated and other complex options, constructing your own binary payload can be a requirement.</help>
     </field>
     <field>
         <id>option.data</id>
--- a/src/opnsense/mvc/app/models/OPNsense/Kea/KeaDhcpv4.php
+++ b/src/opnsense/mvc/app/models/OPNsense/Kea/KeaDhcpv4.php
@@ -204,8 +204,8 @@ class KeaDhcpv4 extends BaseModel
                     }
                     // Kea autoconverts strings to binary when providing 'data' => "'data to convert'"
                     $data = $option->data->getValue();
-                    if ($option->encoding->getValue() === 'string') {
-                        $data = "'" . $data . "'";
+                    if ($option->encoding->isEqual('string')) {
+                        $data = "'{$data}'";
                     }
 
                     $optdata[] = [
@@ -230,8 +230,8 @@ class KeaDhcpv4 extends BaseModel
 
                 // Kea autoconverts strings to binary when providing 'data' => "'data to convert'"
                 $data = $option->data->getValue();
-                if ($option->encoding->getValue() === 'string') {
-                    $data = "'" . $data . "'";
+                if ($option->encoding->isEqual('string')) {
+                    $data = "'{$data}'";
                 }
 
                 $entry = [

### CODE REVIEW

**Comment by Holladiewal** [comment]
The fact that this can be used for every option makes this so much better than my approach for option 43 in #9860. 
I gave it a quick test just now:

It is currently possible to select two entries that modify the same option code, so config data for the same option code is generated twice:
grafik

This doesn't stop Kea from starting up however and only the _first_  defined option is actually served.

**Comment by Monviech** [comment]
Thanks for giving this a quick look, the behavior of KEA is pretty much ideal with this.

The named/typed options will take precedence this way before the untyped raw byte options if there are duplicates.

This means we won't need complex validations.

The hex values also support any encapsulated option since you can construct anything you need.

I think overall this approach is the most promising overall:
- accessible typed options for dns server etc... exist in subnet/reservation for "normal" users
- an escape hatch via generic byte code exists for "power" users who need full payload control

Maintainance of this almost non existent since it is so generic and RFC conform.

File: src/opnsense/mvc/app/controllers/OPNsense/Kea/forms/dialogOption4.xml [hunk 1]
```diff
@@ -0,0 +1,37 @@
...
+        <id>option.code</id>
+        <label>Code</label>
+        <type>dropdown</type>
+        <help>DHCPv4 option to offer to the client.</help>
+    </field>
+    <field>
+        <id>option.encoding</id>
+        <label>Encoding</label>
+        <type>dropdown</type>
+        <help>Choose between auto converting the option data to binary, or providing binary data yourself. For encapsulated and other complex options, constructing your own binary payload can be a requirement.</help>
```
fichtner:
```suggestion
        <help>Choose between auto-converting the option data to binary, or providing binary data yourself. For encapsulated and other complex options, constructing your own binary payload can be a requirement.</help>
```

src/opnsense/mvc/app/models/OPNsense/Kea/KeaDhcpv4.php
```diff
@@ -196,12 +196,53 @@ private function getConfigSubnets()
...
+            /* append raw options */
+            foreach ($subnet->option->getValues() as $uuid) {
+                $option = $this->getNodeByReference("options.option.$uuid");
+                if ($option === null) {
+                    continue;
+                }
+
+                // Kea autoconverts strings to binary when providing 'data' => "'data to convert'"
+                $data = $option->data->getValue();
+                if ($option->encoding->getValue() === 'string') {
```
fichtner:
```suggestion
                if ($option->encoding->isEqual('string')) {
```

src/opnsense/mvc/app/models/OPNsense/Kea/KeaDhcpv4.php
```diff
@@ -196,12 +196,53 @@ private function getConfigSubnets()
...
+                    $data = "'" . $data . "'";
```
fichtner:
```suggestion
                    $data = "'{$data}'";
```

little more readable IMO

PR #9969, opened 2026-03-15
fix(ntpd): remove stale /dev/pps0 symlink when PPS is disabled

### LINKED ISSUE #2219: Setting the NTP stratum level has no effect
When setting up a GPS or PPS ntp reference, setting the stratum level has no effect. In fact, nothing changes in the `/var/etc/ntpd.conf` autogenerated file.

There is actually an effect, but an unwanted one. When setting up a PPS reference, if the "Serial port" is set to `none` (PPS disabled), but "Stratum" is not left empty, a PPS entry is still written in `ntpd.conf`. Not sure if it is the same for the GPS reference, but it's worth checking.

### LINKED ISSUE #9968: PPS config still written to ntpd.conf after disabling — /dev/pps0 symlink persists (ref #2219)
I have read the contributing guidelines
I am convinced that my issue is new after checking open and closed issues. Issue https://github.com/opnsense/core/issues/2219 was closed but the underlying problem is not fully resolved.

Describe the bug
When PPS serial port is set to "none" and stratum field is empty in Services > Network Time > PPS, PPS lines are still written to /var/etc/ntpd.conf if /dev/pps0 exists as a symlink. This causes ntpd to crash on startup with refclock_atom: /dev/pps0: Is a directory. Last known working state: PPS was never functional due to this bug. Related to closed issue https://github.com/opnsense/core/issues/2219.

To Reproduce
Go to Services > Network Time > PPS (/ui/services/ntpd/pps)
Set Serial port to "none", leave Stratum empty, save
Go to Services > Network Time > GPS, configure a GPS device, save
Observe ntpd crashes with refclock_atom: /dev/pps0: Is a directory in NTP log

Expected behavior
With PPS serial port set to "none", no PPS lines should appear in ntpd.conf and ntpd should start cleanly.

Describe alternatives you considered
Manually deleting /dev/pps0 symlink via shell (rm /dev/pps0) prevents the PPS lines from being written. This is the workaround until the bug is fixed.

Relevant log files
refclock_atom: /dev/pps0: Is a directory
GPS_NMEA(0) read EOF

Additional context
Root cause: ntpd_configure_pps() in plugins.inc.d/ntpd.inc line 218 uses file_exists() which returns true on symlinks. The symlink /dev/pps0 -> /dev/cuaU1 is created by the same function and persists after PPS is disabled. Fix submitted in PR https://github.com/opnsense/core/pull/9969.

Environment
OPNsense 26.1.2 (amd64)
Intel Core i3-1115G4
Adafruit Ultimate GPS PA1616D (PA1616D/MTK3333) on USB

Original issue
Issue #2219 was closed but the underlying problem is not fully resolved. Even with PPS serial port set to "none" and stratum field empty, PPS lines are still written to /var/etc/ntpd.conf if /dev/pps0 exists as a symlink.

Root cause: ntpd_configure_pps() in plugins.inc.d/ntpd.inc line 218 uses file_exists() to check for the port device before writing PPS config. file_exists() returns true on symlinks, so even with PPS disabled in the GUI the config generator still writes the PPS block if the symlink exists.

The symlink /dev/pps0 -> /dev/cuaU1 was originally created by ntpd_configure_pps() and persists after PPS is disabled in the GUI. Nothing removes it on disable.

Symptom: ntpd crashes on startup with refclock_atom: /dev/pps0: Is a directory

**Workaround:** rm /dev/pps0

Proper fix: Remove the symlink when PPS is disabled in the GUI, or check if the configured port is empty/none before calling file_exists().

Tested on OPNsense 26.1 with Adafruit Ultimate GPS PA1616D (PA1616D/MTK3333) on USB.

### LINKED ISSUE #9970: NTP PPS config written to ntpd.conf when disabled due to stale /dev/pps0 symlink
I have read the contributing guidelines
I am convinced that my issue is new after checking open and closed issues. Issue https://github.com/opnsense/core/issues/2219 was closed but the underlying problem is not fully resolved.

Describe the bug
When PPS serial port is set to "none" and stratum field is empty in Services > Network Time > PPS, PPS lines are still written to /var/etc/ntpd.conf if /dev/pps0 exists as a symlink. This causes ntpd to crash on startup with refclock_atom: /dev/pps0: Is a directory. Last known working state: PPS was never functional due to this bug. Related to closed issue #2219.

To Reproduce
Go to Services > Network Time > PPS (/ui/services/ntpd/pps)
Set Serial port to "none", leave Stratum empty, save
Go to Services > Network Time > GPS, configure a GPS device, save
Observe ntpd crashes with refclock_atom: /dev/pps0: Is a directory in NTP log

Expected behavior
With PPS serial port set to "none", no PPS lines should appear in ntpd.conf and ntpd should start cleanly.

Describe alternatives you considered
Manually deleting /dev/pps0 symlink via shell (rm /dev/pps0) prevents the PPS lines from being written. This is the workaround until the bug is fixed.

Relevant log files
refclock_atom: /dev/pps0: Is a directory
GPS_NMEA(0) read EOF

Additional context
Root cause: ntpd_configure_pps() in plugins.inc.d/ntpd.inc line 218 uses file_exists() which returns true on symlinks. The symlink /dev/pps0 -> /dev/cuaU1 is created by the same function and persists after PPS is disabled. Fix submitted in PR #9969.

Environment
OPNsense 26.1.2 (amd64)
Intel Core i3-1115G4
Adafruit Ultimate GPS PA1616D (PA1616D/MTK3333) on USB

justgosh:
When PPS serial port is set to none, the /dev/pps0 symlink created by ntpd_configure_pps() persists and causes file_exists() to return true, resulting in PPS lines being written to ntpd.conf even when PPS is disabled. This causes ntpd to crash on startup with: refclock_atom: /dev/pps0: Is a directory

Fix: explicitly remove the symlink when port is empty, and add port emptiness check to guard the PPS config block.

Fixes #2219
Fixes #9968
Fixes #9970

### DIFF
--- a/src/etc/inc/plugins.inc.d/ntpd.inc
+++ b/src/etc/inc/plugins.inc.d/ntpd.inc
@@ -212,9 +212,15 @@ function ntpd_configure_do($verbose = false)
     }
     $ntpcfg .= "\n";
 
+    /* Clean up PPS symlink if PPS is disabled or port is not set */
+    if (empty($config['ntpd']['pps']['port'])) {
+        @unlink('/dev/pps0');
+    }
+
     /* Add PPS configuration */
     if (
         !empty($config['ntpd']['pps'])
+		&& !empty($config['ntpd']['pps']['port'])
         && file_exists('/dev/' . $config['ntpd']['pps']['port'])
         && ntpd_configure_pps($config['ntpd']['pps']['port'])
     ) {


AdSchellevis:
@fichtner missed your assignment, my bad, but https://github.com/opnsense/core/commit/770480715b15e1645c98a3eabade878e8f757e74 seems like the shortest path to close the loop here. Feel free to amend. 

@justgosh thanks for reporting and opening a PR, could you test https://github.com/opnsense/core/commit/770480715b15e1645c98a3eabade878e8f757e74 on your end? thanks!

fichtner:
@AdSchellevis no worries, works for me

justgosh:
@AdSchellevis Thank you for maintaining this project. I understood my code and it worked in testing. You deleted any reference to me in what looks like a superior implementation and you are asking me to test something I don't completely understand in a production environment that recently had an outage due to this bug. I'll wait for it to work its way through the update process or sooner if I understand more of the implications of your change before I do. Again, grateful for your work here.

Also worth noting for anyone searching for a solution to PPS issues: PPS over USB on FreeBSD is unreliable regardless of this fix. Users may enable it expecting microsecond accuracy and instead file bug reports wondering why their stratum 1 NTP is still showing 50ms jitter. USB polling latency is the cause regardless of kernel options. For accurate PPS, connect the PPS signal to a hardware serial port (cuau) via the DCD pin. Note that PPS_SYNC is not compiled into the FreeBSD GENERIC kernel, which is a separate issue affecting hardware UART PPS accuracy.

There is already a note on the GUI but a more explicit warning might have been: "Note: PPS via USB serial devices is unreliable on FreeBSD due to USB polling latency and is not recommended. For accurate PPS, connect the PPS signal converted to RS-232 levels to a hardware serial port (cuau) via the DCD pin. Additionally, hardware UART PPS is not optimized in OPNsense due to PPS_SYNC not being compiled into the FreeBSD kernel, so PPS will not be as precise as other Linux distributions. That said, hardware UART PPS remains the most accurate timing method available on OPNsense."

For those considering chrony as an alternative: the os-chrony plugin on OPNsense doesn't support GPS refclock because gpsd isn't available in the OPNsense package repo. Without gpsd, chrony can't read the GPS module since it relies on gpsd's SHM interface. ntpd has its own native NMEA serial driver that bypasses gpsd entirely, which is why ntpd works directly with the GPS module on OPNsense while chrony doesn't. Sideloading gpsd wouldn't fix the PPS issue anyway since PPS over USB on FreeBSD doesn't work reliably in practice due to USB polling latency so you are back to hardware serial port (cuau) via the DCD pin.

For anyone researching PPS accuracy on OPNsense: hardware UART PPS on FreeBSD without PPS_SYNC will see accuracy in the 10-100 microsecond range rather than the sub-10 microsecond range available with PPS_SYNC enabled distros. UART PPS is still significantly better than USB NMEA only at 10-100ms. A Raspberry Pi with GPIO-connected PPS running chrony on a standard Raspberry Pi OS distro achieves sub-microsecond accuracy at the GPS source, but NTP over LAN to OPNsense adds 100-500us of latency, making it less accurate at OPNsense than a directly wired hardware UART PPS signal but better than a USB NMEA attached to OPNsense. Sub-microsecond accuracy is relevant for specialized use cases such as financial trading, telecom infrastructure, or scientific instrumentation which sync to GPS disciplined oscillators (GPSDO), Rubidium clocks, or PTP (IEEE 1588) networks.

PR #9973, opened 2026-03-15
Captive portal: fix `track6` IPv6 redirect host and bracketed `redirurl`

agoodkind:
Upstream PR: https://github.com/opnsense/core/pull/9745

### Summary

This follow-up to the comment: https://github.com/opnsense/core/pull/9745#issuecomment-4054896550 and fixes two IPv6 redirect bugs that are still present on top of
- When the portal interface uses `track6`, the template currently reads the literal
  config value `track6` and falls back to the IPv4 portal host. An IPv6-only client
  then receives a redirect to an unreachable IPv4 address.
- When a client reaches the portal using a bracketed IPv6 literal, the current
  host-match regex truncates the value at the first `:`, so `redirurl` becomes
  incomplete.

### Why

`track6` is a normal OPNsense IPv6 deployment mode, not a corner case. DHCPv6-PD is required for CE routers by [RFC 7084](https://www.rfc-editor.org/rfc/rfc7084.html) (WPD-1), and [RIPE-690](https://www.ripe.net/publications/docs/ripe-690/) documents it as the standard operator practice for prefix assignment to end-users. OPNsense documents `track6` as the normal way to place a delegated `/64` on a LAN.
 
Captive portal users are often deploying on ISP-managed or upstream-managed connections where the firewall receives a delegated prefix rather than a static IPv6 assignment. The portal needs the live routed IPv6 address, not the static config token.
 
RIPE's deployment survey reports that 37.5% of ISPs still rotate prefixes, so the routed LAN IPv6 can legitimately change underneath the portal. OPNsense already uses runtime address handling in other places for `track6`, notably the web GUI binding work in <https://github.com/opnsense/core/issues/5966>. The captive portal faces the same problem and should use the same pattern.

The underlying issue is that the configd template system was designed around `config.xml`, which holds static values. With IPv6 and track6, interface addresses are runtime-derived: the delegated prefix comes from DHCPv6-PD and can change, so the LAN address isn't known until the interface is configured. This isn't unique to the captive portal. The web GUI (https://github.com/opnsense/core/blob/master/src/etc/inc/plugins.inc.d/webgui.inc#L36), Unbound (https://github.com/opnsense/core/blob/master/src/etc/inc/plugins.inc.d/unbound.inc#L45), and OpenSSH (https://github.com/opnsense/core/blob/master/src/etc/inc/plugins.inc.d/openssh.inc#L44) all solved the same problem via `newwanip` hooks and PHP-side config regeneration (originally driven by https://github.com/opnsense/core/issues/5966).

More broadly, this is probably going to keep coming up as IPv6 support continues to mature, since IPv6's end-to-end model means runtime-derived addresses appear in places that IPv4 never required. It might be worth considering a general-purpose mechanism for passing runtime interface data into templates, rather than solving it service-by-service.
 
The regex fix is independent and corrects IPv6 literal handling directly.
 
### Lab overview
 
Reproduces two captive portal IPv6 bugs on OPNsense. kea delegates a `/60` via DHCPv6-PD, opnsense tracks it on LAN with `track6`, IPv6-only client hits captive portal.
 
### Hosts
 
| Host | Interfaces | Role |
|------|------------|------|
| kea-pd-test | WAN: `3d06:bad:b01:200::1/64` | DHCPv6-PD server. Delegates `/60`s (`::210::/60`–`::2f0::/60`). |
| opnsense-dev | WAN: `3d06:bad:b01:200::11/128`, LAN: `10.240.240.1/24` + `3d06:bad:b01:210:be24:11ff:fef5:c938/64` (track6) | Firewall under test. Captive portal enabled. |
| ubuntu-test-client | LAN: `3d06:bad:b01:210::902/64` | IPv6-only client behind opnsense-dev. |
 
### Addressing
 
| Segment | Range |
|---------|-------|
| WAN | `3d06:bad:b01:200::/64` |
| Delegated PDs | `3d06:bad:b01:210::/60`–`::2f0::/60` |
| LAN | `3d06:bad:b01:210::/64` |
 
### Test plan
- Configure a captive portal zone on a `track6` LAN with no hostname override
- From an IPv6-only client, request `http://[lan-ipv6]:8000/`
- Confirm the redirect host uses the live bracketed IPv6 portal address
- Confirm `redirurl` preserves the full bracketed IPv6 literal
### Testing matrix for the two fixes
| Case | Upstream result | This PR |
|------|-----------------|---------|
| Dynamic IPv6 portal host on `track6` LAN | Fails. Redirect is `http://10.240.240.1:8000/...` | Redirect to `http://[3d06:bad:b01:210:be24:11ff:fef5:c938]:8000/...` |
| Bracketed IPv6 literal preserved in `redirurl` | Fails. `redirurl=[2601/` | `redirurl=[2606:4700:4700::1111]/` |

### DIFF
--- a/src/opnsense/service/templates/OPNsense/Captiveportal/lighttpd-zone.conf
+++ b/src/opnsense/service/templates/OPNsense/Captiveportal/lighttpd-zone.conf
@@ -9,8 +9,6 @@
                     {% set runtime_hostaddr_v6 = helpers.interface_routed_address6(conf_key) %}
                     {% if runtime_hostaddr_v6 and helpers.is_ipv6(runtime_hostaddr_v6) %}
                         {% do item.update({'interface_hostaddr_v6': runtime_hostaddr_v6}) %}
-                    {% elif conf_inf.ipaddrv6 and helpers.is_ipv6(conf_inf.ipaddrv6) %}
-                        {% do item.update({'interface_hostaddr_v6': conf_inf.ipaddrv6}) %}
                     {% endif %}
                     {% if conf_inf.ipaddr and conf_inf.ipaddr != 'dhcp' %}
                         {% do item.update({'interface_hostaddr_v4': conf_inf.ipaddr}) %}
@@ -18,8 +16,6 @@
                     {# prefer IPv6 if available, fallback to IPv4 #}
                     {% if runtime_hostaddr_v6 and helpers.is_ipv6(runtime_hostaddr_v6) %}
                         {% do item.update({'interface_hostaddr':runtime_hostaddr_v6, 'interface_hostaddr_family':'ipv6'}) %}
-                    {% elif conf_inf.ipaddrv6 and helpers.is_ipv6(conf_inf.ipaddrv6) %}
-                        {% do item.update({'interface_hostaddr':conf_inf.ipaddrv6, 'interface_hostaddr_family':'ipv6'}) %}
                     {% endif %}
                     {# fallback to IPv4 if IPv6 not found #}
                     {% if not item.interface_hostaddr and conf_inf.ipaddr and conf_inf.ipaddr != 'dhcp' %}


src/opnsense/scripts/interfaces/get_interface_address.php
```diff
@@ -0,0 +1,51 @@
+#!/usr/local/bin/php
```
fichtner, line 1: There’s pluginctl -6 interface for that kind of operation.

src/opnsense/service/modules/addons/template_helpers.py
```diff
@@ -151,6 +153,35 @@ def physical_interfaces(self, names):
...
+            check=False
+        )
+        address = result.stdout.strip() if result.returncode == 0 else ''
+        self._runtime_interface_address_cache[cache_key] = address
+        return address
+
+    def interface_routed_address4(self, name: str) -> str:
+        return self._runtime_interface_address(name, 'inet')
+
+    def interface_routed_address6(self, name: str) -> str:
```
fichtner: I would advise not to start doing this in templates too.

src/opnsense/scripts/interfaces/get_interface_address.php
agoodkind: `pluginctl -6` calls `interfaces_primary_address6()`, which returns the link-local address (`fe80::...`) for `track6` interfaces. For captive portal redirects we need the routed GUA, which is what `interfaces_routed_address6()` returns.

src/opnsense/service/modules/addons/template_helpers.py
agoodkind: We need this here because captive portal's lighttpd config is generated through the Jinja-based `configd` path, and that path only sees the static config model unless runtime interface state is exposed explicitly.

For `track6`, the correct routed IPv6 address is not present in that static model. It only exists in live interface state after DHCPv6-PD has delegated a prefix and the interface has been configured. So without a helper at this layer, the template has no way to emit the correct IPv6 redirect target.

This is not an edge case. DHCPv6-PD with `track6` is a normal IPv6 deployment model, and OPNsense has already had to solve the same runtime-address problem elsewhere. `webgui.inc`, `openssh.inc`, and `unbound.inc` already handle live interface address changes through PHP-side regeneration and `newwanip` hooks. The difference here is only that captive portal renders its final config through Jinja, so the same class of runtime information has to be exposed differently.

src/opnsense/scripts/interfaces/get_interface_address.php
fichtner: Which is the reason I think this is over-engineered and premature to address now without getting the IPv6 support out first on dependable scope.

src/opnsense/service/modules/addons/template_helpers.py
AdSchellevis: This really needs a different fix for the captive portal template, in practice, when CP is reflecting, it doesn't care which ipv6 address you use as long as its routed and passes the firewall. A manual input might be an option, best practice is and remains adding domain names. Let's focus on the important parts first to avoid getting lost in piles of code which is likely not needed.

agoodkind: Fair. Thanks for reviewing!

Issue #9290, opened 2025-10-15
Add "update now" to firewall GeoIP settings


Taomyn:
**Important notices**

Before you add a new report, we ask you kindly to acknowledge the following:

- [x] I have read the contributing guide lines at https://github.com/opnsense/core/blob/master/CONTRIBUTING.md
- [x] I am convinced that my issue is new after having checked both open and closed issues at https://github.com/opnsense/core/issues?q=is%3Aissue

**Is your feature request related to a problem? Please describe.**

When setting or changing the GeoIP URL in settings, there is little indication that it's working until the next automated update check. This prevents diagnosing any issues without needing to wait some random amount of time.

**Describe the solution you like**

A button that forces an update and some kind of status message. Ideally the ability to set the time for when the update happens would also be useful or a link to the appropriate log viewer.

**Additional context**

Born from this discussion: https://forum.opnsense.org/index.php?topic=49180

abdullahdevrel:
I work for IPinfo. I am subscribing to this ticket in case there are any issues with our data. Unfortunately, cannot help with the software side of things.

AdSchellevis:
@Taomyn  Not sure about an extra button, but from the console you can run at all times:


```
/usr/local/opnsense/scripts/filter/download_geoip.py
```

Monviech:
Fixed via: https://github.com/opnsense/core/pull/9974

fichtner:
@TomWalraven wanted to make another PR for cron job use

PR #9974, opened 2026-03-16
aliases: Add force GeoIP update button

TomWalraven:
Added a Update GeoIP button in Firewall: Aliases: GeoIP settings. This button downloads GeoIP data from a given URL with status messages to indicate whether it worked correctly.

Fixes: #9290

### DIFF
--- a/src/opnsense/service/conf/actions.d/actions_filter.conf
+++ b/src/opnsense/service/conf/actions.d/actions_filter.conf
@@ -206,5 +206,5 @@ message: fetch geoip statistics
 [update.geoip]
 command:/usr/local/opnsense/scripts/filter/download_geoip.py
 parameters:
-type:script_output
+type:script
 message:force update geoip


fichtner:
I don't want to derail this but I need to point out "update bogons" button in in alias diagnostics. Maybe we can unify this?

TomWalraven:
How would you like to see these buttons unified?

src/opnsense/mvc/app/views/OPNsense/Firewall/alias.volt
```diff
@@ -594,6 +594,24 @@
         }
         loadSettings();
 
+        // update geoip data
```
AdSchellevis: you can simplify the button handling using a `SimpleActionButton` (https://docs.opnsense.org/development/frontend/view_js_helpers.html#simpleactionbutton)

which adds a spinner to the button and throws a dialog on failure.

src/opnsense/mvc/app/views/OPNsense/Firewall/alias.volt
```diff
@@ -721,6 +739,9 @@
     </div>
     <div id="geoip" class="tab-pane fade in">
       {{ partial("layout_partials/base_form",['fields':formGeoIPSettings,'id':'frm_GeopIPSettings'])}}
+        <div>
```
AdSchellevis: if you move this in a separate tab ("Actions"?), you can also add the bogons button current defined here 

https://github.com/opnsense/core/blob/57015c28435e82f41902328dd75e0e8aee7d6f33/src/opnsense/mvc/app/views/OPNsense/Firewall/alias_util.volt#L78-L84

https://github.com/opnsense/core/blob/57015c28435e82f41902328dd75e0e8aee7d6f33/src/opnsense/mvc/app/views/OPNsense/Firewall/alias_util.volt#L200-L207

(which can also use the `SimpleActionButton` option)

TomWalraven: I've added a separate tab "Actions" in Firewall/Aliases and put both buttons there. Let me know if there's anything that needs to be added or modified. a9239b0

src/opnsense/mvc/app/views/OPNsense/Firewall/alias.volt
```diff
@@ -722,6 +728,16 @@
     <div id="geoip" class="tab-pane fade in">
       {{ partial("layout_partials/base_form",['fields':formGeoIPSettings,'id':'frm_GeopIPSettings'])}}
     </div>
+    <div id="actions" class="tab-pane fade in">
+        <div class="row">
+            <section class="col-xs-12">
+                <div class="content-box">
+                    <button id="geoip_update_btn" type="button" class="btn btn-default" data-endpoint="/api/firewall/alias/update_geo_ip" data-label="{{ lang._('Update GeoIP') }}" data-error-title="{{ lang._('Error updating GeoIP') }}"></button>
```
fichtner:
```suggestion
                    <button id="geoip_update_btn" type="button" class="btn btn-default" data-endpoint="/api/firewall/alias/update/geoip" data-label="{{ lang._('Update GeoIP') }}" data-error-title="{{ lang._('Error updating GeoIP') }}"></button>
```

perhaps a shared endpoint for "update" would work like so? allowing the parameter "geoip" or "bogons"

TomWalraven: Something like this? 68d362b

src/opnsense/mvc/app/controllers/OPNsense/Firewall/Api/AliasController.php
```diff
@@ -493,4 +493,16 @@ public function getGeoIPAction()
...
         return $result;
     }
+
+    /**
+     * update specified alias type
+     */
+    public function updateAction($action)
+    {
+        if ($this->request->isPost() && in_array($action, ['geoip', 'bogons'])) {
+            (new Backend())->configdRun('filter update ' . $action);
```
fichtner: you need configdpRun() here, action is potentially unsafe (not now but in the future)

if $action is null you may also get an API error here?

src/opnsense/mvc/app/views/OPNsense/Firewall/alias_util.volt
```diff
@@ -75,15 +75,6 @@
 
         });
 
-        // update bogons
-        $("#update_bogons").on('click', function(event){
-            event.preventDefault();
-            $("#update_bogons_progress").addClass("fa fa-spinner fa-pulse");
-            ajaxCall("/api/firewall/alias_util/update_bogons", {}, function(){
```
fichtner, line 82: updateBogonsAction() is still defined -- normally we don't touch API but in this case the risk of removal is low -- @AdSchellevis can decide

TomWalraven: Wasn't this already removed in 68d362b?

fichtner: In 919559a35 -- I was viewing a partial diff it seems, sorry for that

src/opnsense/service/conf/actions.d/actions_filter.conf
```diff
@@ -202,3 +202,9 @@ parameters:
 errors:no
 type:script_output
 message: fetch geoip statistics
+
+[update.geoip]
+command:/usr/local/opnsense/scripts/filter/download_geoip.py
+parameters:
+type:script_output
+message:force update geoip
```
fichtner:
```suggestion
type:script
message:force update geoip

```

fichtner: there is no useable output for the GUI, `script_output` hands it over but the GUI doesn't use it so `script` is enough

src/opnsense/mvc/app/controllers/OPNsense/Firewall/Api/AliasController.php
TomWalraven: I've changed it in 7f496b4. It doesn't seem like I'm getting an API error while testing $action being null

TomWalraven:
Great! Is there still anything that should be changed about the frontend? I didn't put much priority in the styling

fichtner:
I'd leave it as is. Perhaps discuss at the meeting, but in reality it's just two buttons and not much to be done with it :)

src/opnsense/mvc/app/controllers/OPNsense/Firewall/Api/AliasController.php
fichtner: probably cast to empty string '' in these cases, but that's ok :)

fichtner:
Merged, thanks!

PR #9975, opened 2026-03-16
Fix False “PTR record already exists” Warnings in Unbound DNS Overrides

daygle:
Unbound in OPNsense incorrectly emits a “PTR record already exists” warning when multiple DNS overrides share the same IP address, even if only one of those overrides actually requests a PTR record.

When two overrides point to the same IP - one with  enabled and one without - the current logic evaluates the second override as if it were attempting to create a PTR, causing Unbound to log:

`Warning unbound PTR record already exists for <hostname>(<ip>)`

This warning is misleading because no duplicate PTR was requested. The warning should only be generated when multiple overrides explicitly request PTR creation (addptr = true) for the same IP address.

### DIFF
--- a/src/etc/inc/plugins.inc.d/unbound.inc
+++ b/src/etc/inc/plugins.inc.d/unbound.inc
@@ -613,39 +613,57 @@ function unbound_add_host_entries($ifconfig_details)
                 }
 
                 foreach ($tmp_aliases as $alias) {
-                    if ($alias['hostname'] != '') {
-                        $alias['hostname'] .= '.';
-                    }
-
-                    switch ($host->rr) {
-                        case 'A':
-                        case 'AAAA':
-                            /* Handle wildcard entries which have "*" as a hostname. Since we added a . above, we match on "*.". */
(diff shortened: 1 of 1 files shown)


AdSchellevis:
your replacing and `AND` operator with an `if` in an `if`, which is the same....

daygle:
@AdSchellevis - The original code logs a warning whenever the combined condition fails - including when `addptr` is false.

My change ensures the warning is only emitted when `addptr = true` and the PTR already exists.

This prevents false warnings when multiple overrides share an IP but only one requests a PTR.

So the logic is not equivalent - the original `else` is too broad, and that’s the bug being fixed

AdSchellevis:
in that case it's better to change the `else` for an `ifelse`

daygle:
@AdSchellevis Ok, any better?

AdSchellevis:
@daygle I was aiming for a one line fix, here you go https://github.com/opnsense/core/commit/b9a5d5c5cd608738987187bbdb57a9b97363c6dd

PR #9979, opened 2026-03-16
Services: Kea: DHCPv4: Remove option auto-conversion of string to binary

Monviech:
Cleanup previous: https://github.com/opnsense/core/commit/8350fcb73b9dd44e8b1e00d2ea03ced71e0f71ac

It's not as it seems, it only supports flat options and is very picky. It will create wrong expectations, rather than what hexadecimal byte pairs do - which are really generic.

I was lucky when I evaluated it by picking the 1 or 2 options it actually worked with, but it fails for most others, unlike hex.

Better not go into there right away, it will attract the wrong user reactions.

### DIFF
--- a/src/opnsense/mvc/app/controllers/OPNsense/Kea/forms/dialogOption4.xml
+++ b/src/opnsense/mvc/app/controllers/OPNsense/Kea/forms/dialogOption4.xml
@@ -9,7 +9,7 @@
         <id>option.encoding</id>
         <label>Encoding</label>
         <type>dropdown</type>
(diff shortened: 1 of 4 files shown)


AdSchellevis:
An alternative might be to add specific "convertors" similar to how kea seems to implement the "standard" types (https://kea.readthedocs.io/en/kea-3.1.6/arm/dhcp4-srv.html#id5)

If the user specifies which input it expects, we can convert that data to the hex format it expects.

Monviech:
Sounds like a good idea for future expansion, but I won't jump into it right away (if that's okay).

I want to see first if users are able to create most configurations now that would have needed the custom configuration files before.

AdSchellevis:
I do expect people will have challenges converting their data into a hex string, but there's certainly no reason to rush anything.

Issue #7688, opened 2024-07-30
Redirect to login screen in all open browser tabs when session of WebGUI becomes invalid


Monviech:
**Important notices**

Before you add a new report, we ask you kindly to acknowledge the following:

- [x] I have read the contributing guide lines at https://github.com/opnsense/core/blob/master/CONTRIBUTING.md
- [x] I am convinced that my issue is new after having checked both open and closed issues at https://github.com/opnsense/core/issues?q=is%3Aissue

**Is your feature request related to a problem? Please describe.**

- Open a Browser
- Log in, go to `/ui/core/dashboard` or any other page (but dashboard shows it clearer)
- Open another browser tab, go to "Lobby/Logout"
- You get logged out in that tab, but the `/ui/core/dashboard` in the other tab remains open.
- The CSRF Token is now invalid and all widgets fail to load while the dashboard (or any other site) remains open. All user interaction that depends on the API fail until a refresh and a log in.

**Describe the solution you like**

When the session of a user becomes invalid, there should be a forced refresh or some other mechanism to redirect them back to the login screen, in all Browser Tabs the user has open.

**Describe alternatives you considered**

Leaving it as it is since its not that big of a deal, it can just be confusing when it happens. Maybe its more obvious when people want to run the Dashboard on a big screen for long periods of times and suddenly it starts to fail.
A refresh reveals the session has become invalid.

AdSchellevis:
We might think of a generic solution by trapping requests, but there will be downsides of redirecting to the login page automatically as well. So far this hasn't been a huge issue, but it's good to keep a ticket open to discuss options.

Monviech:
I don't have experience how these things usually work.

But I imagine some kind of watchdog. It will try to reach an API every minute or so that is always accessable for any user, and when that fails it retries a few times before it catches the error and shows a bootstrap dialog.

"The Login session has expired."
With a button that will refresh the current page when pressed, and a button to cancel the dialog. (So there is no forceful redirect.)

AdSchellevis:
poll works indeed, but creates a constant load, using the requests being fired is likely also an option, a bit like how error handlers currently work (`ajaxError`, which is just a hook inside jQuery to trap all ajax events fired using jQuery):

https://github.com/opnsense/core/blob/d2ef070687d1a4077a818211db4ffd6a5eab69f6/src/opnsense/mvc/app/views/layouts/default.volt#L52-L67

[...]

PR #9980, opened 2026-03-17
Logout logic

sopex:
This PR implements automatic logout logic when the session timeout is reached. 

Once the session expires, the pages/ tabs automatically reload and redirect the user to the login screen. 

After the user logs in in one tab, all tabs log in to the page they were before the logout.

As an unintended consequence, it seems to log out users with much greater consistency than before. https://forum.opnsense.org/index.php?topic=51305.0 (I found that out when I reduced the timeout session)

### DIFF
--- a/src/opnsense/mvc/app/controllers/OPNsense/Base/ControllerBase.php
+++ b/src/opnsense/mvc/app/controllers/OPNsense/Base/ControllerBase.php
@@ -397,6 +397,7 @@ class ControllerBase extends ControllerRoot
         $this->view->session_username = !empty($_SESSION['Username']) ? $_SESSION['Username'] : '(unknown)';
         $this->view->system_hostname = $cnf->object()->system->hostname;
         $this->view->system_domain = $cnf->object()->system->domain;
+        $this->view->session_timeout = $this->session_timeout;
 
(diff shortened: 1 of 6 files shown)


Monviech:
I remember I made an issue for this once:
https://github.com/opnsense/core/issues/7688

AdSchellevis:
@Monviech thanks for the reference, at the moment I don't want to merge this as it solves part of the problem (and it's not on our roadmap to spend time on this), when automatically logging out , we probably need to make sure we have some visual feedback to the user as well.

sopex:
> I remember I made an issue for this once: #7688

Hadn't seen it due to naming differences, what you are describing there is a subproblem addressed in this PR.

I see your conclusion there but the session timeout process is already significantly unreliable, its just that almost no one knows it because the 4 hour default makes it rarely triggered.

sopex:
> @Monviech thanks for the reference, at the moment I don't want to merge this as it solves part of the problem (and it's not on our roadmap to spend time on this), when automatically logging out , we probably need to make sure we have some visual feedback to the user as well.

I would appreciate if you could share what is the other part of the problem is. Also, the visible feedback is that they got reloaded to the login screen. That way, they log in and continue what they are working on, in contrast to now that they realize when they eventually hit apply.

PR #9983, opened 2026-03-17
auth/user: Added option to API to set the user password as a hash

jakobsen-lrz:
This adds a boolean option hashed_password to the API. When set to true, it means that the password is already hashed, and therefore doesnt need to be hashed anymore.

This is useful for automatically importing users.

I hope this can be merged like this. Thanks.

### DIFF
--- a/src/opnsense/mvc/app/controllers/OPNsense/Auth/Api/UserController.php
+++ b/src/opnsense/mvc/app/controllers/OPNsense/Auth/Api/UserController.php
@@ -98,7 +98,11 @@ class UserController extends ApiMutableModelControllerBase
             } else {
                 $password = $node->password->getValue();
             }
-            $hash = $this->getModel()->generatePasswordHash($password);
+            if (empty((string)$node->scrambled_password) && !empty((string)$node->hashed_password)) {
+                $hash = $password;
+            } else {
+                $hash = $this->getModel()->generatePasswordHash($password);
+            }
             if ($hash !== false && strpos($hash, '$') === 0) {
                 $node->password = $hash;
                 $node->pwd_changed_at = microtime(true);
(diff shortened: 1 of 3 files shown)


fichtner:
Perhaps that's a good idea in principle but this doesn't validate input at all.

jakobsen-lrz:
The problem with verifying the hash, is that there isnt a simple function that can do that.

I know that password_get_info exists, but that can only identify hashes that are created using password_hash. And not hashes that can be verified against using password_verify. Password verify also doesnt throw any error when given an invalid hash.

The only way i can think of is using a simple regex like: `^\$(1|apr1|2|2a|2b|2y3|5|6)\$[a-zA-Z0-9./+=_-]+\$[a-zA-Z0-9./+=_-]+$`
But this only catches some invalid hashes, so i dont know if this is satisfactory.

fichtner:
Yeah, I see that it's not easy. The risk of locking the user out is small but real. Least we have to do is discuss it. I'm sure @AdSchellevis will chime in too.

AdSchellevis:
For import/export purposes we should at least limit the behavior to these calls in my opinion, to avoid some endpoint doing something weird and suddenly locking people out. 

Currently this field is ignored, which might be an option to change, but also has (limited) security risks.

jakobsen-lrz:
I tried to edit the UI so that it would explicitly set the flag to false. But i couldnt figure out how to do that.

fichtner:
A hidden read-only field that overlays to password field might be a bit more convenient. It avoids exposing this on the form and having to set it although this wouldn't solve the risky part just yet.

jakobsen-lrz:
I closed this PR, as i wouldnt be able to do these more advanced changes.

PR #9988, opened 2026-03-17
Services: Kea: DHCPv4/v6: Add single client_class support to DHCP options

Monviech:
Should theoretically fix issues like these generically: https://github.com/opnsense/core/pull/9912

For: https://github.com/opnsense/core/issues/9942

Matching a DHCP option is optional.

This should match 90% of the usecases, and makes it easy to reason with (if this -> then that, in the same dialog), no complex hierarchies possible.

Send option always:
image

Send option only to clients who send that option 93 (client arch) and it's a certain data value
image

image

Attach to any subnet4/subnet6 and reservation
image

Configuration example:

```
{
    "Dhcp4": {
        "subnet4": [
            {
                "id": 1,
                "subnet": "192.168.50.0\/24",
                "next-server": "",
                "match-client-id": false,
                "option-data": [
                    {
                        "data": "192.168.30.1"
                    },
                    {
                        "name": "routers",
                        "data": "192.168.30.1"
                    },
                    {
                        "name": "domain-name",
                        "data": "ad.pischem.com"
                    },
                    {
                        "name": "ntp-servers",
                        "data": "192.168.30.1"
                    },
                    {
                        "code": 3,
                        "csv-format": false,
                        "data": "06060606",
                        "always-send": false
                    },
                    {
                        "code": 4,
                        "csv-format": false,
                        "data": "08080808",
                        "always-send": false,
                        "client-classes": [
                            "b5ba1864-bc3d-43e7-9e8d-b31b34ad59f6"
                        ]
                    }
                ],
                "pools": [
                    {
                        "pool": "192.168.50.100-192.168.50.110"
                    }
                ],
                "reservations": []
            }
        ],
        "client-classes": [
            {
                "name": "b5ba1864-bc3d-43e7-9e8d-b31b34ad59f6",
                "test": "option[7].hex == 0x01010101"
            }
        ]
    }
}
```

### DIFF
--- a/src/opnsense/mvc/app/models/OPNsense/Kea/KeaDhcpv4.xml
+++ b/src/opnsense/mvc/app/models/OPNsense/Kea/KeaDhcpv4.xml
@@ -327,7 +327,7 @@
                     </OptionValues>
                 </encoding>
                 <data type=".\KeaOptionDataField">
-                    <EncodingField>encoding</EncodingField>
+                    <encoding_source>encoding</encoding_source>
                     <Required>Y</Required>
                 </data>
                 <force type="BooleanField"/>
@@ -363,7 +363,7 @@
                     </Constraints>
                 </match_encoding>
                 <match_data type=".\KeaOptionDataField">
-                    <EncodingField>match_encoding</EncodingField>
+                    <encoding_source>match_encoding</encoding_source>
                     <Constraints>
                         <check001>
                             <ValidationMessage>Match option and match data must either both be set or left empty.</ValidationMessage>
@@ -376,6 +376,12 @@
                 </match_data>
                 <description type="DescriptionField">
                     <Required>Y</Required>
+                    <Constraints>
+                        <check001>
+                            <ValidationMessage>Description must be unique.</ValidationMessage>
+                            <type>UniqueConstraint</type>
+                        </check001>
+                    </Constraints>
                 </description>
             </option>
         </options>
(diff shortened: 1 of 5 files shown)


src/opnsense/mvc/app/models/OPNsense/Kea/FieldTypes/KeaMatchOptionDataField.php
```diff
@@ -0,0 +1,61 @@
...
+
+    public function getValidators()
+    {
+        $validators = parent::getValidators();
+        if (!empty($this->internalValue)) {
+            $validators[] = new CallbackValidator([
+                "callback" => function ($data) {
+
+                    $messages = [];
+                    $encoding = $this->getParentNode()->match_encoding->getValue();
```
Monviech: Its a bit unfortunate I had to copy the Field Type just to change this one field. I will fix this with configurability soon.

src/opnsense/mvc/app/models/OPNsense/Kea/KeaDhcpv4.xml
```diff
@@ -327,9 +327,53 @@
                     </OptionValues>
                 </encoding>
                 <data type=".\KeaOptionDataField">
+                    <EncodingField>encoding</EncodingField>
```
fichtner: this looks a bit weird: a property that is named like a field type

It would be better to use a different name or spacing

Monviech: Changed the name to encoding_source as discussed, thanks.

PR #9989, opened 2026-03-17
fix: enforce POST-only on captiveportal logoffAction

oliverjueguen:


AdSchellevis:
@oliverjueguen thanks!

PR #9993, opened 2026-03-17
dnsmasq: prefer client_id over hwaddr for IPv4 reservation detection

### LINKED ISSUE #9972: UI Regression and Terminology Inconsistency in DHCP Leases & Static Mappings (Dnsmasq)
**Important notices**

**Describe the bug**
Since the transition to Dnsmasq as the default DHCP/RA provider, OPNsense has moved toward a more unified view for DHCPv4 and DHCPv6. However, the current implementation suffers from significant terminology mismatches and functional UI regressions. The labels used in the frontend do not accurately reflect the underlying protocol data, leading to a confusing user experience.

### 1. Misleading Terminology in Leases Overview
**Issue:** For DHCPv4 leases, the column labeled "DUID" actually displays the full RFC 4361 Client Identifier (Option 61).

**Defect:** Labeling this strictly as "DUID" is technically incorrect for IPv4. Furthermore, the existing "IAID" column remains empty for IPv4, even though the IAID is clearly part of the string displayed in the "DUID" column.

### 2. Confusing Static Mapping Form (v6 DUID vs. v4 Identifier)

**Issue:** When creating a static IPv6 mapping from an active lease (clicking the '+' icon), the UI pre-fills the DUID into a field labeled "Client identifier".

**Defect:** "Client identifier" is specifically a DHCPv4 term (Option 61). Using this label for a DHCPv6 DUID (RFC 3315) is highly confusing. The UI essentially forces two different protocol standards into a single, incorrectly labeled input field without clarifying which format is expected.

### 3. Faulty "Static" Status Detection (IPv4)

**Issue:** If an administrator creates an IPv4 static mapping using the Client Identifier (copying the string from the "DUID" column) while leaving the MAC field blank, the backend correctly assigns the static IP.

**Defect:** However, the Leases Overview continues to flag this lease as "dynamic". The UI's status logic appears to rely exclusively on MAC address matching for IPv4, ignoring valid identifier-based mappings. This makes the "Status" column untrustworthy for modern, ID-based setups.

### 4. Ambiguous Help Texts

**Issue:** The help text for the static mapping field "CLient Identifier" attempts to cover both worlds but fails at both: "Match the identifier of the client, e.g., DUID for DHCPv6..."

**Defect:** It does not explain that for IPv4, the full Option 61 string (including the IAID and prefix) is required, while for IPv6, it expects a DUID but currently ignores the IAID.

### Summary

The old ISC and KEA both have a clear separation of DHCPv4 and DHCPv6 in the UI. This would probably be best for dnsmasq, too. If it shall stay like it is, please at lest get the terminology right or clarify the current situation in the docs. 

P.S. Regarding IAID in Static Mappings:
While the primary issue is terminology, it should be noted that the current Static Mapping form for DHCPv6 lacks a dedicated field for the IAID. Since DHCPv6 uniquely identifies an interface via the (DUID, IAID) tuple, omitting the IAID makes deterministic assignment difficult for multi-interface hosts

**Environment**
OPNsense 26.1.4-amd64
FreeBSD 14.3-RELEASE-p9

Curly060:
Fixes #9972

IPv4 leases are currently matched only by hwaddr, ignoring client_id (option 61). This causes leases with client identifier based mappings to not be marked as reserved.

Prefer client_id when present and fall back to hwaddr for IPv4 to retain backward compatibility.

### DIFF
--- a/src/opnsense/mvc/app/controllers/OPNsense/Dnsmasq/Api/LeasesController.php
+++ b/src/opnsense/mvc/app/controllers/OPNsense/Dnsmasq/Api/LeasesController.php
@@ -88,9 +88,27 @@ class LeasesController extends ApiControllerBase
         }
 
         foreach ($records as &$record) {
-            $is_ipv6 = Util::isIpv6Address($record['address'] ?? '');
-            $key = strtolower($is_ipv6 ? ($record['client_id'] ?? '') : ($record['hwaddr'] ?? ''));
-            $record['is_reserved'] = in_array($key, $reservedKeys, true) ? '1' : '0';
+            $is_ipv4 = Util::isIpv4Address($record['address'] ?? '');
+
+            $is_reserved = false;
+
+            // Always prefer client_id for both, IPv4 and IPv6
+            if (!empty($record['client_id'])) {
+                $key = strtolower($record['client_id']);
+                if (in_array($key, $reservedKeys, true)) {
+                    $is_reserved = true;
+                }
+            }
+
+            // Fallback to hwaddr for IPv4 if not matched yet
+            if (!$is_reserved && $is_ipv4 && !empty($record['hwaddr'])) {
+                $key = strtolower($record['hwaddr']);
+                if (in_array($key, $reservedKeys, true)) {
+                    $is_reserved = true;
+                }
+            }
+
+            $record['is_reserved'] = $is_reserved ? '1' : '0';
         }
 
         $response = $this->searchRecordsetBase(


sopex:
I want to preface this by saying that I don't particularly enjoy the current UI presentation of dnsmasq. 

But why should client-id be the default? It does not really provide an advantage over MAC addresses.

Monviech:
@Curly060 

Thanks for the suggestion. At this point I rather want to know what caused the reservation rather than guessing with the boolean more.

https://github.com/opnsense/core/pull/9651/changes

I will fix this like in KEA, thank you for pointing out a possible direction.

Curly060:
@sopex 
> But why should client-id be the default? It does not really provide an advantage over MAC addresses.

To clarify: this change does not make client identifiers the default for IPv4.

The existing behavior remains unchanged for all MAC-based configurations. The only difference is that if a client identifier is explicitly configured in a static reservation, it will now be taken into account for IPv4 leases as well (which was previously not the case).

Regarding the usefulness of client identifiers:

They can provide advantages over MAC-based identification in certain scenarios, for example:
- stability across interface changes (e.g. replacing a NIC or using a bridge/bond)
- compatibility with modern DHCP clients (e.g. systemd-networkd) that prefer RFC4361-style identifiers

That said, MAC-based reservations remain fully supported and are perfectly fine for many setups.

@Monviech 
That makes sense - exposing what actually caused the reservation match (e.g. hwaddr vs client_id) would definitely be more transparent than a simple boolean.

My intention with this change was mainly to fix the current inconsistency where client_id-based reservations for IPv4 are not recognized at all, even though the data is already available.

I agree that a more explicit representation (similar to what KEA does) would be the better long-term solution. My patch would then just ensure that client_id-based matches are correctly detected in the first place.

Happy to adjust or rework this if you have a specific direction in mind.

Monviech:
@Curly060 don't worry, I will fix this issue myself. I'll treat this PR verbatim to an issue.

Monviech:
@Curly060 

Can you check if this solves your issue?
https://github.com/opnsense/core/pull/10003

PR #9995, opened 2026-03-18
Services: Kea: DHCPv4/6: Add type system with validation and encoding for supported DHCP options

Monviech:
For: https://github.com/opnsense/core/issues/9942

This adds a type framework using maps extracted from data inside the KEA documentation.

Most flat options have a validator and encoder added. Users are guided via validation messages which types are correct choices for each code selection. Advanced options (structured, nested, encapsulated) do not have an own validator or encoder yet, they fall back to hex. Unknown options that are valid in the IANA lists, but maybe not inside KEA yet, also fallback to hex.

The user can always bail out of the type system and always use hex instead for any option.

### DIFF
--- a/plist
+++ b/plist
@@ -1100,6 +1100,7 @@
 /usr/local/opnsense/mvc/tests/app/models/OPNsense/Base/FieldTypes/VirtualIPFieldTest.php
 /usr/local/opnsense/mvc/tests/app/models/OPNsense/Base/FieldTypes/VirtualIPFieldTest/config.xml
 /usr/local/opnsense/mvc/tests/app/models/OPNsense/Dnsmasq/FieldTypes/HostnameFieldTest.php
+/usr/local/opnsense/mvc/tests/app/models/OPNsense/Kea/FieldTypes/KeaOptionDataFieldTest.php
 /usr/local/opnsense/mvc/tests/phpunit.xml
 /usr/local/opnsense/mvc/tests/setup.php
 /usr/local/opnsense/scripts/auth/add_user.php
--- a/src/opnsense/mvc/tests/app/models/OPNsense/Kea/FieldTypes/KeaOptionDataFieldTest.php
+++ b/src/opnsense/mvc/tests/app/models/OPNsense/Kea/FieldTypes/KeaOptionDataFieldTest.php
@@ -1,0 +1,145 @@
+<?php
+
+/*
+ * Copyright (C) 2026 Deciso B.V.
+ * All rights reserved.
+ *
+ * Redistribution and use in source and binary forms, with or without
+ * modification, are permitted provided that the following conditions are met:
+ *
+ * 1. Redistributions of source code must retain the above copyright notice,
+ *    this list of conditions and the following disclaimer.
+ *
+ * 2. Redistributions in binary form must reproduce the above copyright
+ *    notice, this list of conditions and the following disclaimer in the
+ *    documentation and/or other materials provided with the distribution.
+ *
+ * THIS SOFTWARE IS PROVIDED ``AS IS'' AND ANY EXPRESS OR IMPLIED WARRANTIES,
+ * INCLUDING, BUT NOT LIMITED TO, THE IMPLIED WARRANTIES OF MERCHANTABILITY
+ * AND FITNESS FOR A PARTICULAR PURPOSE ARE DISCLAIMED. IN NO EVENT SHALL THE
+ * AUTHOR BE LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY,
+ * OR CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF
+ * SUBSTITUTE GOODS OR SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS
+ * INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN
+ * CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE)
+ * ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
+ * POSSIBILITY OF SUCH DAMAGE.
+ */
+
+namespace tests\OPNsense\Kea\FieldTypes;
+
+// @CodingStandardsIgnoreStart
+require_once __DIR__ . '/../../Base/FieldTypes/Field_Framework_TestCase.php';
+// @CodingStandardsIgnoreEnd
+
+use tests\OPNsense\Base\FieldTypes\Field_Framework_TestCase;
+use OPNsense\Kea\FieldTypes\KeaOptionDataField;
+
+class KeaOptionDataFieldTest extends Field_Framework_TestCase
+{
+    private function createField(string $encoding, string $value, int $code = 0): KeaOptionDataField
+    {
+        $node = function ($value) {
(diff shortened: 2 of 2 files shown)


src/opnsense/mvc/app/models/OPNsense/Kea/FieldTypes/KeaOptionDataField.php
```diff
@@ -45,32 +276,261 @@ public function setEncodingSource($value): void
...
+        $map = $this->getOptionTypeMap();
+        if (!isset($map[$code])) {
+            return true; // unknown/vendor options
+        }
+        return in_array($encoding->value, $map[$code], true);
+    }
+
+    public function getEncodedValue(): string
+    {
+        $data = trim($this->getValue());
```
fichtner: not sure if we're going to hide errors here more than this is suppsed to help :)

src/opnsense/mvc/app/models/OPNsense/Kea/KeaDhcpv4.php
```diff
@@ -252,7 +252,7 @@ private function getConfigClientClasses()
             }
             $result[] = [
                 'name' => $uuid,
-                'test' => sprintf('option[%d].hex == 0x%s', $option->match_code->asInt(), $option->match_data->getValue()),
+                'test' => sprintf('option[%d].hex == 0x%s', $option->match_code->asInt(), $option->match_data->getEncodedValue()),
```
fichtner: historically __toString() would be the one that offers an "encoded" value although this goes back to explicit casting in the code, but I wanted to mention it

swhite2:
Since the fieldtype is aware of the possible option<->encoding combinations, a separate metadata endpoint could also be considered (or bundled with the existing API call) to allow the front-end to match each option to the right encoding type, updating the selectpicker as necessary. This way users don't have to run into validation errors first.

Not a priority though, more a nice to have.

Monviech:
I think I'm done here for now. @fichtner and @swhite2 tested this briefly already and it looks good.

@fichtner will review this tomorrow, I won't change anything anymore till then.

Monviech:
When reviewing, here is documentation explaining how things should theoretically work now

https://github.com/opnsense/docs/pull/860

src/opnsense/mvc/app/models/OPNsense/Kea/FieldTypes/KeaOptionDataField.php
```diff
@@ -45,32 +300,237 @@ public function setEncodingSource($value): void
...
+        }
+    }
+
+    public function encodeValue(): string
+    {
+        $encoding = $this->getEncoding();
+        if ($encoding === null) {
+            return '';
+        }
+        $method = self::ENCODER_MAP[$encoding->value] ?? null;
```
fichtner, line 318: still believe this null can kill $this->$method
```suggestion
        $method = self::ENCODER_MAP[$encoding->value] ?? null;
        if ($method ===  null) {
            return '';
        }
```

fichtner: (unless you want to force the hard error)

src/opnsense/mvc/app/models/OPNsense/Kea/FieldTypes/KeaOptionDataField.php
```diff
@@ -45,32 +300,237 @@ public function setEncodingSource($value): void
...
+        return $this->$method($this->getValue());
+    }
+
     public function getValidators()
     {
         $validators = parent::getValidators();
-        if (!empty($this->internalValue)) {
+        if ($this->internalValue !== null && $this->internalValue !== '') {
```
fichtner: this should be equivalent... handling internalValue is a bit too internal IMO

```suggestion
        if ($this->isSet()) {
```

src/opnsense/mvc/app/models/OPNsense/Kea/FieldTypes/KeaOptionDataField.php
```diff
@@ -45,32 +300,237 @@ public function setEncodingSource($value): void
...
+                                gettext("Encoding '%s' is not valid for option %d, use %s or hex."),
+                                $encoding->value,
+                                $code,
+                                implode(', ', $allowed)
+                            )
+                        ];
                     }
-
-                    return $messages;
+                    $method = self::VALIDATOR_MAP[$encoding->value];
```
fichtner: you're not saveguarding here, maybe we shouldn't safeguard the other one too. I'm simply unsure :D

src/opnsense/mvc/app/models/OPNsense/Kea/FieldTypes/KeaOptionDataField.php
```diff
@@ -45,32 +300,237 @@ public function setEncodingSource($value): void
...
+        return KeaEncoding::tryFrom($this->getParentNode()->{$this->internalEncodingSource}->getValue());
+    }
+
+    private function isEncodingAllowed(): bool
+    {
+        $encoding = $this->getEncoding();
+        if ($encoding === null || $encoding === KeaEncoding::HEX) {
+            return true; // configuring hex is always allowed as bailout
+        }
+        $code = $this->getParentNode()->{$this->internalCodeSource}->asInt();
```
fichtner, line 380: this can cause issues when integrity is not given, but we'll have to see in practice

src/opnsense/mvc/app/models/OPNsense/Kea/FieldTypes/KeaOptionDataField.php
```diff
@@ -45,32 +300,237 @@ public function setEncodingSource($value): void
...
+        $map = $this->getOptionTypeMap();
+        if (!isset($map[$code])) {
+            return false; // unknown options are hex-only
+        }
+        return in_array($encoding->value, $map[$code], true);
+    }
+
+    private function getListValues(string $data): array
```
fichtner:
```suggestion
    private function toList(string $data): array
```

to avoid more get***Value() ambiguity. I'm not sure about the trim but since it's burried here we can try

src/opnsense/mvc/app/models/OPNsense/Kea/FieldTypes/KeaOptionDataField.php
```diff
@@ -45,32 +300,237 @@ public function setEncodingSource($value): void
...
+    private function encodeHex(string $data): string
+    {
+        return strtoupper($data);
+    }
+
+    private function encodeIpv4(string $data): string
+    {
+        return implode('', array_map(function ($ip) {
+            return implode('', array_map(fn($o) => sprintf('%02X', (int)$o), explode('.', $ip)));
+        }, $this->getListValues($data)));
```
fichtner: a comment and a test case for this would be good to have because it's not very obvious what this does :)

src/opnsense/mvc/app/models/OPNsense/Kea/FieldTypes/KeaOptionDataField.php
Monviech: lets better be safe than sorry :)

src/opnsense/mvc/app/models/OPNsense/Kea/FieldTypes/KeaOptionDataField.php
Monviech: I'll add a few assertions at least for the encoders. The validations should be simple enough I think.

Monviech:
@fichtner I added a Unit Test for the encoders, please give this one a look too.

fichtner:
@Monviech super, looking good :)

Monviech:
Thanks, here we go then :)

PR #10003, opened 2026-03-19
Services: Dnsmasq DNS & DHCP: Since client-id is a valid IPv4 reservation type as well, ensure the lease view handles it correctly

Monviech:
The same is also true for MAC address as IPv6 reservation type.

Fixes: https://github.com/opnsense/core/pull/9993
Fixes: https://github.com/opnsense/core/issues/9972

This is not taking IP protocols into account right now, I might need to encance this again one more step. But its very hard to be truly determenistic here with also only having a single lookup button in the GUI.

Needs another look by me first to split between v4 and v6 keys too.

EDIT:

I just looked back into the actual host identity we allow, and I cannot group reservation keys any better, since we allow mixed IPv4/IPv6 reservations based on multiple mac addresses/client_id/DUIDs etc... nothing more we can do here heuristically.

image

### DIFF
--- a/src/opnsense/mvc/app/views/OPNsense/Dnsmasq/leases.volt
+++ b/src/opnsense/mvc/app/views/OPNsense/Dnsmasq/leases.volt
@@ -87,11 +87,6 @@
                             : "{{ lang._('dynamic') }}";
                     },
                     "commands": function (column, row) {
-                        const isIPv6 = row.address.includes(':');
-                        const queryParams = {
-                            ...(isIPv6 ? { client_id: row.client_id || '' } : { hwaddr: row.hwaddr || '' }),
-                        };
-
                         const baseUrl = `/ui/dnsmasq/settings#hosts`;
                         const reservedBy = row.is_reserved || [];
                         let searchKey = '';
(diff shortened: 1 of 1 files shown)


src/opnsense/mvc/app/views/OPNsense/Dnsmasq/leases.volt
Curly060: is_reserved has changed semantics, see line 85 where you did take this into account.

Curly060:
wow, that was quick, thanks!
I have applied the patch and tested this. While it does work indeed, it introduces a bug with the add reservation button: it is always showing now, see my comment. If that is fixed, I can test again.

Monviech:
@Curly060 Thanks for testing, I fixed the spot you found, and also removed some heuristics about the "Add Reservation" button. It should now always present both client_id and mac address when creating a reservation, and you can decide in the dialog which to keep or not, or both.

There cannot be more heuristics about IP protocol and such because everything is so mixed up. It's a many to many relationship, so the approach right now should be fully deterministic.

Curly060:
Awesome, these are great improvements - the workflow is much more usable now.

What I have successfully tested:

- Adding reservations
  - for a dynamic DHCPv4 lease:
    - MAC only
    - Client Identifier only
    - both combined
  - for a dynamic DHCPv6 lease:
    - reservation correctly prefilled with DUID in the Client Identifier field
  - modifying the DUID:
    - verified that matching behaves as expected (DHCPv4 reservations based on MAC, client_id, or both only match when intended)
- Lease lookup / navigation
  - searching via the Hosts view works well and reliably finds the correct entries

The only remaining minor issue is the naming of the “Client Identifier” field in the reservation dialog, which can still be a bit misleading. However, with the improved prefilling and overall workflow, this is much less of a problem now in practice.

Overall, I am very happy with these changes. Thank you!

Monviech:
Hey, thank you for testing. This is good enough for master then.

About terminology, if you have any ideas how it could be made clearer, feel free to offer suggestions, also to our docs.
Please keep in mind to follow the dnsmasq manual closely in terms of terminology since that is what we try to relay to the user without inventing too much.

PR #10011, opened 2026-03-20
bootgrid: automatic grid height calculation

swhite2:
Grids will now fill all available screen space (if the amount of data allows), minus a margin. The margin may also include a user-supplied element or selector, of which the dimensions are used to subtract the grid height to leave space for said element. By default this element is `.grid-bottom-reserve`, as on most pages the `base_apply_button` is used below the grid.

Note that the page itself still has a scrollbar, as the `base_apply_button` has a larger padding that expands beyond the visible container to make room for the "Don't forget to press apply" banner. A separate PR will propose a styling change for this.

### DIFF
--- a/src/opnsense/mvc/app/views/layout_partials/base_apply_button.volt
+++ b/src/opnsense/mvc/app/views/layout_partials/base_apply_button.volt
@@ -1,5 +1,5 @@
 <section class="page-content-main">
-    <div class="content-box">
+    <div class="content-box grid-bottom-reserve">
         <div class="col-md-12">
             <br/>
             <div id="{{ message_id|default('change_message_base_form') }}" class="alert alert-info" style="display: none" role="alert">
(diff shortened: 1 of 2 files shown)


AdSchellevis (approved):
@swhite2 nice!

fichtner:
🌈 how magical 🌈

Monviech:
Its very good :)

falkenau:
Hello! We are running "OPNsense 26.1.9". For example, if we go to "-> Firewall -> Rules [new]" the grid-bottom-reserve is quite large. How can we reduce it to zero?
We need to scroll a lot within the grid. I tested three browsers: Chrome, Firefox, Edge.

swhite2:
@falkenau Are you referring to:

afbeelding

Or disabling in-grid scrolling entirely?

falkenau:
@swhite2 YES! I would love that. And there is enough room for the Apply Button at the top.

fichtner:
> @swhite2 YES! I would love that. And there is enough room for the Apply Button at the top.

Wait what

swhite2:
Not sure what's going on here, but I guess https://github.com/opnsense/core/pull/10375 is related.

falkenau:
Screenshot 2026-06-10 000524

Just look at all this wasted space. Why do we need to scroll so much in a tiny grid? :)

Issue #9536, opened 2025-12-14
Kea DHCP GUI: add option to flush active DHCP leases (parity with ISC DHCP)


Tomipapa:
### **Description**

- When using Kea DHCP in OPNsense, there is currently no GUI option to flush / clear active DHCP leases.
- In operational scenarios (e.g. network renumbering, VLAN changes, MAC reuse, test environments), being able to clear active leases is essential.
- At the moment, this requires dropping to the CLI and interacting directly with Kea’s control socket or lease database, which is inconvenient and inconsistent with the GUI-first workflow of OPNsense.
- Notably, this functionality already exists in the GUI when using ISC DHCP, which makes the limitation more visible when migrating to Kea.
### **Observed behaviour**
1. Kea DHCP enabled
2. Active leases visible in GUI
3. No GUI control to:
- flush all leases
- flush leases per subnet
- flush a single lease
4. Operator must use CLI (kea-shell, lease DB manipulation, or service restart)
### **Expected behaviour**
Provide a GUI action similar to ISC DHCP, for example:
- “Flush all leases” (global)
- “Flush leases for this subnet”
- (optional) “Release selected lease”

The action should:
- call Kea’s supported control interface (e.g. lease4-del, lease6-del, or database cleanup)
- be explicit and guarded (confirmation dialog)
- be logged for audit purposes
### **Why this matters**
- Common operational requirement (testing, troubleshooting, renumbering)
- CLI access is not always desirable or available
- Breaks parity with existing ISC DHCP GUI
- Slows down routine operations
- Increases risk of human error when manipulating leases manually
### **Environment**
- OPNsense 25.7.9 with Kea DHCP plugin
- Feature comparison point: ISC DHCP GUI lease management
### **Additional notes**
- This request is about GUI parity and operator usability, not a Kea limitation.
- Kea already provides mechanisms to manage leases programmatically; exposing them in the GUI would significantly improve day-to-day operations.

Happy to help test or provide feedback on UX/behaviour if needed.

Monviech:
It's good that the GUI does not allow to flush leases. As soon as the DHCP process has completed for a client, the server and client have a mutual agreement that the lease is valid for as long as the valid lease lifetime.

If the server now deletes the lease, another client can claim it, and that causes IP address duplication in the worst case.

DHCP Release from the client is the correct safe tool for this.

[...]

Issue #9217, opened 2025-09-17
Provide GUI option to delete dynamic Kea leases


jwgn:
**Important notices**
Before you add a new report, we ask you kindly to acknowledge the following:

- [x] I have read the contributing guide lines at https://github.com/opnsense/plugins/blob/master/CONTRIBUTING.md
- [x] I have searched the existing issues, open and closed, and I'm convinced that mine is new.
- [x] When the request is meant for an existing plugin, I've added its name to the title.

**Is your feature request related to a problem? Please describe.**
There is no option in the WebUI to delete dynamic Kea DHCP leases.

**Describe the solution you'd like**
Dynamic ISC DHCP leases can be deleted with a button in the WebUI. There is no such option in the Kea leases section. Please implement this functionality for Kea within the WebUI.

**Describe alternatives you've considered**
none

**Additional context**
OPNsense 25.7.3_7-amd64

Monviech:
Same discussion as here should apply:

https://github.com/opnsense/core/pull/8899

zimmra:
> Same discussion as here should apply:
> 
> [#8899](https://github.com/opnsense/core/pull/8899)

I'm not familiar with dnsmasq, but with KEA it's as simple as the desired rows being deleted from `/var/db/kea/kea-leases4.csv` and `/var/db/kea/kea-leases4.csv.[#]` and restarting the service. No dbus or anything advanced required like what's mentioned in #8899 

e.g. if you wanted to release the lease for `192.168.1.195` you could run something like  `sed -i '/^192\.168\.1\.195,/d' /var/db/kea/kea-leases4.* && pluginctl -s kea-dhcp restart`

Monviech:
But while KEA is running it has authority over these file(s). That is the same issue as with dnsmasq. You cannot alter a lease database while the process is running because you can cause a race with the running process.

In the linked PR, I wanted to force the process to be stopped before write operations to the lease database.

The dbus discussion was that dnsmasq does not provide any standard interface to delete a lease while it is running. If KEA provides an interface for this it could be discussed.

zimmra:
> You cannot alter a lease database while the process is running because you can cause a race with the running process.

Unless I'm interpreting this wrong, that specifically is a recommended method per [Kea's own documentation](https://kea.readthedocs.io/en/kea-3.0.2/arm/lease-expiration.html#lease-reclamation) - `Update lease information in the lease database to indicate that the lease is now available for reassignment`

[...]

PR #10019, opened 2026-03-23
Services: Kea: DHCPv4/6: Add delete lease command, use socket for up to date lease collection

Monviech:
Fixes: https://github.com/opnsense/core/issues/9217
Fixes: https://github.com/opnsense/core/issues/9536

This shows the direction such a button could take. It works in my production KEA setup.

The idea is that leases you can see in the views now reflect the correct in-memory representation of KEA. Any command you do that mutates these leases through the available command sockets mutates the in-memory state.

This means, the csv lease files do not always represent the correct state at any given time during runtime, only the state collected by the socket is true.

When deleting a lease, it disappears from the in-memory database instantly and will not show up in the lease view. Though in the csv file it will be marked with a lifetime of 0, or be represented in csv.1 and csv.2 multiple times due to rotating journaling.

In essence this means we can only trust the control socket here, and never the files.

Docs:
https://kea.readthedocs.io/en/latest/api.html#lease4-del
https://kea.readthedocs.io/en/latest/api.html#lease6-del
https://kea.readthedocs.io/en/latest/api.html#config-get
https://kea.readthedocs.io/en/latest/api.html#lease4-get-all
https://kea.readthedocs.io/en/latest/api.html#lease6-get-all

### DIFF
--- a/src/opnsense/mvc/app/controllers/OPNsense/Kea/Api/LeasesController.php
+++ b/src/opnsense/mvc/app/controllers/OPNsense/Kea/Api/LeasesController.php
@@ -138,7 +138,7 @@ abstract class LeasesController extends ApiControllerBase
 
         if (!empty($results['failed'])) {
             throw new UserException(sprintf(
-                gettext('Failed to delete lease(s) for ip(s) %s. Check the logs for details.'),
+                gettext('Failed to delete lease(s): %s'),
                 implode(', ', $results['failed'])
             ));
         }
--- a/src/opnsense/scripts/kea/del_kea_leases.py
+++ b/src/opnsense/scripts/kea/del_kea_leases.py
@@ -48,7 +48,7 @@ if __name__ == '__main__':
 
         result = KeaCtrl.send_command(cmd, {"ip-address": ip}, service)
         if result.get("result", 1) > 0:
-            results["failed"].append(ip)
+            results["failed"].append(f"{ip}: {result.get("text", "unknown error")}")
         else:
             results["removed"].append(ip)
 
--- a/src/opnsense/scripts/kea/lib/kea_ctrl.py
+++ b/src/opnsense/scripts/kea/lib/kea_ctrl.py
@@ -29,8 +29,6 @@
 import ujson
 import os
 import socket
-import sys
-import syslog
 
 class KeaCtrl:
     @staticmethod
@@ -41,11 +39,8 @@ class KeaCtrl:
         elif service == "dhcp6":
             path = "/var/run/kea/kea6-ctrl-socket"
 
-        syslog.openlog("kea-%s" % service, facility=syslog.LOG_LOCAL4)
-
         if not os.path.exists(path):
-            syslog.syslog(syslog.LOG_ERR, f"kea_ctrl.py: kea-{service} control socket path \"{path}\" does not exist, exiting.")
-            sys.exit(1)
+            return {"result": 1, "text": f"kea-{service} control socket path \"{path}\" does not exist"}
 
         payload = {"command": command}
         if args is not None:
@@ -62,12 +57,6 @@ class KeaCtrl:
                     if not chunk:
                         break
                     buffer.append(chunk)
-                response = ujson.loads(b''.join(buffer).decode('utf-8'))
-                code = response.get("result", 1)
-                if code > 0:
-                    text = response.get("text", "unknown error")
-                    syslog.syslog(syslog.LOG_ERR, f"kea_ctrl.py: kea-{service} command \"{command}\" returned non-zero exit code {code}: {text}")
-                return response
+                return ujson.loads(b''.join(buffer).decode('utf-8'))
         except (OSError, ConnectionError, socket.timeout) as e:
-            syslog.syslog(syslog.LOG_ERR, f"kea_ctrl.py: kea-{service} command \"{command}\" failed: {e}, exiting.")
-            sys.exit(1)
+            return {"result": 1, "text": f"kea-{service} command \"{command}\" failed: {e}"}


src/opnsense/mvc/app/views/OPNsense/Kea/leases4.volt
```diff
@@ -121,9 +121,33 @@
...
+                                                $('#grid-leases').bootgrid('reload');
+                                            }
+                                        },
+                                        "POST"
+                                    );
+                                }
+                            );
+                        });
+
+                        return $('<span>').append(reservationBtn).append(deleteBtn)[0];
```
swhite2: We can tie this whole block to the default `delete` command by setting `datakey: 'address'` and `del: '/api/kea/leases4/delete_lease/'`.

If you want batched deletes (`delete-selected`), you can set `selection: true, multiSelect: true`, and adjust the controller and backend calls to account for an array (https://github.com/opnsense/core/blob/master/src/opnsense/mvc/app/controllers/OPNsense/Base/ApiMutableModelControllerBase.php#L517).

src/opnsense/mvc/app/views/OPNsense/Kea/leases6.volt
```diff
@@ -130,9 +121,33 @@
...
+                                                $('#grid-leases').bootgrid('reload');
+                                            }
+                                        },
+                                        "POST"
+                                    );
+                                }
+                            );
+                        });
+
+                        return $('<span>').append(reservationBtn).append(deleteBtn)[0];
```
swhite2: Same here

src/opnsense/scripts/kea/del_kea_leases.py
```diff
@@ -0,0 +1,49 @@
...
+    cmd = "lease6-del" if ":" in ip else "lease4-del"
+
+    if not os.path.exists(path):
+        print(ujson.dumps({"status": "error", "message": f"socket not found: {path}"}))
+        exit(1)
+
+    sock = socket.socket(socket.AF_UNIX, socket.SOCK_STREAM)
+    sock.connect(path)
+    sock.sendall(ujson.dumps({"command": cmd, "arguments": {"ip-address": ip}}).encode() + b"\n")
+    print(sock.recv(4096).decode())
```
swhite2: Seems to me the `send_command` from `get_kea_leases.py` can be reused in a lib file. Now I know this is a simple delete command and the 4096 buffer is more than plenty, but I'd expect a terminating symbol instead of having to guess (which is what `get_kea_leases` checks for). ~~Is there also no notion of "success" or "failure" here?~~

Edit: seems the latter is handled in the controller

src/opnsense/scripts/kea/del_kea_leases.py
```diff
@@ -0,0 +1,72 @@
...
+
+import argparse
+import os
+import ujson
+import socket
+
+
+def send_command(socket_path, payload):
+    sock = socket.socket(socket.AF_UNIX, socket.SOCK_STREAM)
+    sock.connect(socket_path)
```
swhite2: This should probably be surrounded with a try-except-finally (`finally` for sock.close())

src/opnsense/scripts/kea/del_kea_leases.py
```diff
@@ -0,0 +1,72 @@
...
+
+    for ip in ips:
+        path = "/var/run/kea/kea6-ctrl-socket" if ":" in ip else "/var/run/kea/kea4-ctrl-socket"
+        cmd = "lease6-del" if ":" in ip else "lease4-del"
+
+        if not os.path.exists(path):
+            results.append({"ip": ip, "status": "error", "message": f"socket not found: {path}"})
+            continue
+
+        result = send_command(path, {"command": cmd, "arguments": {"ip-address": ip}})
```
swhite2: this is calling `connect()` for each iteration, probably best to do the socket handling outside the loop

src/opnsense/mvc/app/controllers/OPNsense/Kea/Api/LeasesController.php
```diff
@@ -120,4 +121,33 @@ public function searchAction()
...
+        }
+
+        $results = json_decode((new Backend())->configdpRun("kea delete lease " . escapeshellarg($ips)), true);
+
+        if (!is_array($results) || empty($results)) {
+            throw new UserException(gettext('Invalid backend response'));
+        }
+
+        foreach ($results as $result) {
+            if (!is_array($result) || ($result['status'] ?? '') === 'error' || ($result['result'] ?? 0) === 2) {
```
swhite2: Purely for separation of responsibility, parsing the KEA output might be a job for the backend script itself, not the controller - though I don't have a strong preference.

src/opnsense/scripts/kea/del_kea_leases.py
Monviech: Right now I don't know which IP calls which socket (path), so I would need preprocess more here. How much of an issue is this with the amount of leases this endpoint will most likely process per request, can't we just connect to the socket for each one for now?

The cost per transaction should be extremely small right now.

src/opnsense/scripts/kea/del_kea_leases.py
Monviech: Yeah good eyes, I'll fix that.

src/opnsense/mvc/app/controllers/OPNsense/Kea/Api/LeasesController.php
Monviech: I think the result code parsing is okay here for now. The backend script supplies json with the codes and I just throw here with a translated error message. This keeps the backend script simpler and more generic imo.

src/opnsense/scripts/kea/del_kea_leases.py
swhite2: We won't know until we measure :)

But it should be rather simple to sort the ip list into ipv4/ipv6 and do two loops. With KEA being the way it is...

Monviech: Reusing the socket introduces additional complexity around response and stream handling. As discussed I'll keep it simple for now.

src/opnsense/mvc/app/controllers/OPNsense/Kea/Api/LeasesController.php
```diff
@@ -120,4 +121,33 @@ public function searchAction()
...
+                throw new UserException(gettext('KEA Control Agent is not running'));
```
swhite2: After a bit of testing, it seems the control agent isn't linked to the `ctrl-socket`. If we hook the libraries unconditionally in `KeaDhcpv4/6.php`, we can access the socket always.

Monviech: Yeah that is a great hint, thank you.

swhite2: As discussed, I'll open a separate PR for this, as HA can be cleaned up from there as well

swhite2:
Let's expand the scope of this PR a bit:

- Separate the `libdhcp_lease_cmds` lib hooks from the `control-agent`.
- Create a separate python wrapper that handles the commands/responses of the KEA unix domain socket.
- Hook the existing scripts to said lib.
- Remove CSV lease fetch, only show leases when KEA is enabled.

src/opnsense/scripts/kea/get_kea_leases.py
```diff
@@ -32,28 +33,99 @@
...
+                "if": lease_if,
+                "if_descr": "",
+                "is_reserved": ""
+            })
+
+        if records:
+            print(ujson.dumps({"records": records}))
+            exit(0)
+
+    # CSV fallback if control socket is unavailable or returns no records
```
Monviech: It would be great if we can ditch this entirely since the in-memory database is truthy, and when KEA is not running presenting stale entries is worse than not showing anything.

src/opnsense/service/conf/actions.d/actions_kea.conf
```diff
@@ -46,3 +46,9 @@ command:/usr/local/opnsense/scripts/kea/get_kea_leases.py
 parameters:--proto inet6
 type:script_output
 message:list kea inet dhcp leases
+
+[delete.lease]
+command:/usr/local/opnsense/scripts/kea/del_kea_leases.py
+parameters:%s
+type:script_output
```
fichtner, line 53: Are we using this output? If not „script“ is more suitable

Monviech: https://github.com/opnsense/core/blob/46945f865b5dfc379562db75a91b2975405e9bdf/src/opnsense/mvc/app/controllers/OPNsense/Kea/Api/LeasesController.php#L141-L145

fichtner: 👍

swhite2:
> Always surprising how much work the plumbing for this type of „trivial use case“ can be.

@fichtner FWIW, the total amount of time spent here is certainly not as great as the amount of time this PR has been up. With the additional effort here we at least have centralized error handling and proper code reuse, but I'll certainly admit to getting a bit carried away ;)

fichtner:
It’s for a good cause so that’s ok 😊

Monviech:
Just saying I started doing this for a reason so we know how to control KEA later for the dynamic PD stuff, this one is a nice isolated POC in a way.

src/opnsense/mvc/app/controllers/OPNsense/Kea/Api/LeasesController.php
```diff
@@ -120,4 +121,37 @@ public function searchAction()
...
         return $response;
     }
+
+    public function delLeaseAction($ips = null)
+    {
+        if (!$this->request->isPost()) {
+            return ['status' => 'error', 'message' => gettext('Invalid request method')];
+        }
+
+        if (empty($ips)) {
```
AdSchellevis:
```suggestion
        elseif (empty($ips)) {
```

I know it's a bit "taste sensitive", but by using an `elseif` you can shrink the number of code lines a bit.

src/opnsense/mvc/app/controllers/OPNsense/Kea/Api/LeasesController.php
```diff
@@ -120,4 +121,37 @@ public function searchAction()
...
+            return ['status' => 'error', 'message' => gettext('Missing lease IP parameter')];
+        }
+
+        $results = json_decode((new Backend())->configdpRun("kea delete lease " . escapeshellarg($ips)), true);
```
AdSchellevis:
```suggestion
        $results = json_decode((new Backend())->configdpRun("kea delete lease ", [$ips]), true);
```

shell escaping is part of the wrapper already:

https://github.com/opnsense/core/blob/67425d3b42496168e61988a3eedeaf5c6f20270e/src/opnsense/mvc/app/library/OPNsense/Core/Backend.php#L201

src/opnsense/mvc/app/controllers/OPNsense/Kea/Api/LeasesController.php
```diff
@@ -120,4 +121,37 @@ public function searchAction()
...
+
+        if (!is_array($results) || empty($results)) {
+            throw new UserException(gettext('Invalid backend response'));
+        }
+
+        $errors = [];
```
AdSchellevis: This error handling feels a bit much to be honest, if we just make `del_kea_leases.py` return a list of failed and successful dropped entries we can ditch quite some code. 

E.g.
```
$result = [
   'failed' => ['1','2',..],
   'removed' => ['3','4',..],
]
```

--> `implode(', ', $result['failed'])`

src/opnsense/scripts/kea/lib/kea_ctrl.py
```diff
@@ -0,0 +1,88 @@
...
+import os
+import socket
+import syslog
+
+class KeaCtrl:
+    @staticmethod
+    def _socket_path(service):
+        if service == "dhcp4":
+            return "/var/run/kea/kea4-ctrl-socket"
+        if service == "dhcp6":
```
AdSchellevis:
```suggestion
        elif service == "dhcp6":
```

it can't both be true ;)

src/opnsense/scripts/kea/lib/kea_ctrl.py
```diff
@@ -0,0 +1,88 @@
...
+
+        while True:
+            chunk = sock.recv(4096)
+            if not chunk:
+                raise ConnectionError("Socket closed before full JSON was received")
+
+            buffer += chunk.decode("utf-8")
+
+            try:
+                obj, _ = decoder.raw_decode(buffer)
```
AdSchellevis: Maybe more of a question, but why don't we just read all chunks and parse the response next?  A (highly) simplified version of the `send_command()` implementation only seems to require a couple of lines

```
import ujson
import socket

filename = '/var/run/kea/kea4-ctrl-socket'

payload = {"command": 'config-get'}
with socket.socket(socket.AF_UNIX, socket.SOCK_STREAM) as sock:
    sock.settimeout(5)
    sock.connect(filename)
    sock.sendall(ujson.dumps(payload).encode("utf-8") + b"\n")
    buffer = []
    while True:
        chunk = sock.recv(4096)
        if not chunk:
            break
        buffer.append(chunk)
    res = ujson.loads((b''.join(buffer).decode("utf-8")))
    print(res)
```

src/opnsense/scripts/kea/lib/kea_ctrl.py
```diff
@@ -0,0 +1,88 @@
...
+                return obj
+            except json.JSONDecodeError:
+                continue
+
+    @staticmethod
+    def send_command(command, args = None, service = "dhcp4", timeout = 5.0):
+        path = KeaCtrl._socket_path(service)
+        syslog.openlog("kea-%s" % service, facility=syslog.LOG_LOCAL4)
```
AdSchellevis: Although it looks nice to push logging inside the `send_command`, it might be better to move it to the implementation. For example the `control socket path ... does not exist` error is rather fatal and might yield multiple times now when  `del_kea_leases.py‎` is called with a list of addresses.

I believe the actual caller might know better how to interpret the output, either pass it on or log and exit.

src/opnsense/mvc/app/models/OPNsense/Kea/KeaDhcpv4.php
```diff
@@ -320,13 +324,6 @@ public function generateConfig($target = '/usr/local/etc/kea/kea-dhcp4.conf')
             $cnf['Dhcp4']['expired-leases-processing'] = $expiredLeasesConfig;
         }
         if (!(new KeaCtrlAgent())->general->enabled->isEmpty()) {
-            $cnf['Dhcp4']['hooks-libraries'] = [];
-            $cnf['Dhcp4']['hooks-libraries'][] = [
-                'library' => '/usr/local/lib/kea/hooks/libdhcp_lease_cmds.so'
-            ];
-            $cnf['Dhcp4']['hooks-libraries'][] = [
-                'library' => '/usr/local/lib/kea/hooks/libdhcp_host_cmds.so'
-            ];
```
AdSchellevis, line 329: I think @swhite2 mentioned the `KeaCtrlAgent` dependency doesn't exist for HA either, which would be nice as it would help deprecate it as well 👍

swhite2: PR upcoming as soon as this one is merged

src/opnsense/scripts/kea/lib/kea_ctrl.py
swhite2: the output is always the same from KEA - at least for the error case, hence logging in a single place as it trims quite some code

AdSchellevis: That part is clear, the question just is if this is the place to log. When fetching data, the log might be less relevant if the user sees the result instantly (could be a `UserException` for example), which might also be the case for the delete action given it's only "user driven" any way.

swhite2: I suppose the intent is to catch that which isn't expected from KEA in detailed form, which belongs in the log anyway, assisting debugging at a later point in time instead of what is presented directly to the user.

fwiw, https://github.com/opnsense/core/pull/10019/commits/a730f11e2a69d61ca4e0657c6a51e7c44857e027 does exactly that, logging non-zero results automatically (hard exiting on fatal errors) but still passing the response along so the implementation can do something specific with it. In this case, stating "check the logs for details" saves some code.

PR #10027, opened 2026-03-24
System: Routing: Changed disable option to enable

### LINKED ISSUE #8780: system: change "disable" routes configuration to a more intuitive "enable"
**Important notices**

No sure what all the implications are but just looking at it it would be nicer to align this with standard model layouts since we now have defaults as well to enforce (on by default).

TomWalraven:
Changed the disabled option to enabled in system routes configuration.

Closes: #8780


src/opnsense/mvc/app/models/OPNsense/Routes/Route.xml
```diff
@@ -15,10 +15,10 @@
                 <ValidationMessage>Specify a valid gateway from the list matching the networks ip protocol.</ValidationMessage>
             </gateway>
             <descr type="DescriptionField"/>
-            <disabled type="BooleanField">
-                <Default>0</Default>
+            <enabled type="BooleanField">
```
fichtner: ok here is where this gets interesting: you need a version bump of the model and a matching Migration/ file in order to move all currently disabled entries to the enabled flag. Otherwise you lose the previous user setting for each route.

src/opnsense/mvc/app/models/OPNsense/Routes/Migrations/M1_0_1.php
```diff
@@ -0,0 +1,26 @@
...
+{
+    public function run($model)
+    {
+        $cfgObj = Config::getInstance()->object();
+        if (isset($cfgObj->staticroutes->route)) {
+            $modelRoutes = iterator_to_array($model->route->iterateItems());
+            foreach ($cfgObj->staticroutes->route as $route) {
+                $uuid = (string)$route['uuid'];
+                if (!empty($uuid) && isset($modelRoutes[$uuid]) && isset($route->disabled)) {
+                    $modelRoutes[$uuid]->enabled = ((int)$route->disabled === 1) ? 0 : 1;
```
fichtner:
```suggestion
                    $modelRoutes[$uuid]->enabled = empty((string)$route->disabled) ? '1' : '0';
```

the !empty() is safer as previous migrations have taught us and the cast inside to avoid simplexml element from testing empty instead of the content of the node just for safety. model nodes are strings always so force them right away

src/etc/inc/system.inc
```diff
@@ -761,7 +761,8 @@ function system_routing_configure($verbose = false, $interface_map = null, $moni
 
             $cmd[] = exec_safe('%s', $rtent['network']);
 
-            if (!empty($rtent['disabled'])) {
+            $route_enabled = isset($rtent['enabled']) ? !empty($rtent['enabled']) : empty($rtent['disabled']);
```
fichtner: this looks a bit like @AdSchellevis worrying about the impact :D

AdSchellevis: it wasn't me, but certainly could have been :) A note above to remove this in a future version will help a smoother landing and signal us after a couple of major releases its safe to drop the workaround again.

TomWalraven: I've added a note in 988fc12 to address this😃

src/etc/inc/plugins.inc.d/pf.inc
```diff
@@ -133,7 +133,7 @@ function pf_firewall($fw)
         $GatewaysList = $fw->getGateways()->gatewaysIndexedByName(false, true);
 
         foreach (get_staticroutes() as $route) {
-            if (!empty($GatewaysList[$route['gateway']]['interface']) && empty($route['disabled'])) {
+            if (!empty($GatewaysList[$route['gateway']]['interface']) && !empty($route['enabled'])) {
```
fichtner: I was thinking for get_staticroutes() we could just emulate 'disabled' to ease this (and make both the old and new code work at the same time)?

src/etc/inc/system.inc
fichtner: in any case it's nice... as mentioned folding this into get_staticroutes() may be even better

src/opnsense/mvc/app/controllers/OPNsense/Routes/forms/dialogEditRoute.xml
```diff
@@ -1,9 +1,9 @@
 <form>
   <field>
-    <id>route.disabled</id>
-    <label>Disabled</label>
+    <id>route.enabled</id>
+    <label>Enabled</label>
     <type>checkbox</type>
-    <help>Set this option to disable this static route without removing it from the list.</help>
```
fichtner, line 6: very nice to clean up this spurious note :)

src/opnsense/mvc/app/models/OPNsense/Routes/Migrations/M1_0_1.php
```diff
@@ -0,0 +1,26 @@
+<?php
+
```
fichtner, line 2: there's a copyright header missing here

src/opnsense/mvc/app/models/OPNsense/Routes/Migrations/M1_0_1.php
```diff
@@ -0,0 +1,26 @@
...
+                if (!empty($uuid) && isset($modelRoutes[$uuid]) && isset($route->disabled)) {
+                    $modelRoutes[$uuid]->enabled = empty((string)$route->disabled) ? '1' : '0';
+                }
+            }
+        }
+    }
+}
+
+
+
```
fichtner:
```suggestion
```

fichtner:
I did a quick audit of where else this is used and cleaned this up faa7dab4190eb501142b0f63dc0a6e93bd650628 where necessary.. I think folding the compat into get_staticroutes() is a good way forward that minimizes impact :)

src/etc/inc/util.inc
```diff
@@ -1163,6 +1163,11 @@ function get_staticroutes($returnsubnetsonly = false)
 
     foreach (config_read_array('staticroutes', 'route', false) as $route) {
         if (is_subnet($route['network'])) {
+            /* XXX: fallback for legacy disabled route option. can be removed after a couple of major releases. */
+            $route_disabled = isset($route['enabled']) ? empty($route['enabled']) : !empty($route['disabled']);
+            $route['enabled'] = $route_disabled ? '0' : '1';
+            unset($route['disabled']);
```
fichtner: it would be nice to offer compat for both 'enabled' and 'disabled' because then we don't have to adjust system.inc and pf.inc at all

fichtner: (this also allows outside consumers of get_staticroutes() to keep working)

TomWalraven: I've implemented this in 1bba529 🙂

fichtner (approved):
this turned out better than I could hope for. very nicely done 👍

src/opnsense/mvc/app/models/OPNsense/Routes/Migrations/M1_0_1.php
```diff
@@ -0,0 +1,49 @@
...
+            $modelRoutes = iterator_to_array($model->route->iterateItems());
+            foreach ($cfgObj->staticroutes->route as $route) {
+                $uuid = (string)$route['uuid'];
+                if (!empty($uuid) && isset($modelRoutes[$uuid]) && isset($route->disabled)) {
+                    $modelRoutes[$uuid]->enabled = empty((string)$route->disabled) ? '1' : '0';
+                }
+            }
+        }
+    }
+}
```
fichtner:
```suggestion
}

```

github and line endings 🤷‍♂️

src/opnsense/mvc/app/controllers/OPNsense/Routes/Api/RoutesController.php
```diff
@@ -123,20 +123,20 @@ public function delrouteAction($uuid)
...
+                if ($enabled == '0' || $enabled == '1') {
+                    $node->enabled = (string)$enabled;
+                } elseif ((string)$node->enabled == '1') {
+                    $node->enabled = '0';
                 } else {
-                    $node->disabled = '1';
+                    $node->enabled = '1';
                 }
-                $result['result'] = (string)$node->disabled == '1' ? 'Disabled' : 'Enabled';
+                $result['result'] = (string)$node->enabled == '1' ? 'Enabled' : 'Disabled';
```
swhite2: I think this can use `toggleBase()` now. In fact, this allows the front-end to use multi-select as well (batched toggle):

https://github.com/opnsense/core/blob/master/src/opnsense/mvc/app/views/OPNsense/Routes/index.volt#L39

(`{selection: true}`)

Make sure to set `batchDelete: false` as well, as the `delrouteAction` doesn't account for multiple UUIDs.

src/opnsense/mvc/app/controllers/OPNsense/Routes/Api/RoutesController.php
swhite2: If `toggleBase()` is used, this comment can go as well.

swhite2 (changes requested):

src/opnsense/mvc/app/controllers/OPNsense/Routes/Api/RoutesController.php
TomWalraven: Applied in 0f041ae. Thanks for the suggestion 😄

swhite2 (approved):
Good stuff :)

fichtner:
Merged, thanks!

PR #10028, opened 2026-03-24
menu: add Favorites section

Greelan:
Implements a Favorites section:

- The section only appears if the user has at least one favorite selected. If all favorites are removed, the Favorites section disappears
- Favorites are added/removed simply by clicking on a star icon at the right hand side of the page heading on leaf menu item pages (the star will always appear furthest right, for consistency of appearance, even if other buttons are in the heading). The star icon changes appearance if the item is in the Favorites list
- The Favorites list is stored per-user, and is subject to their permissions
- If a favorited menu item is removed (eg an interface is removed or a plugin uninstalled), the Favorites entry is also removed
- Favorites entries appear in the same order as the corresponding menu items
- The PR caters for both MVC and legacy pages

Some sample screenshots attached.

Screenshot 5
Screenshot 6

Javascript / HTML is not my strong suit, so be kind xD

### DIFF
--- a/plist
+++ b/plist
@@ -654,6 +654,7 @@
 /usr/local/opnsense/mvc/app/models/OPNsense/Base/FieldTypes/UpdateOnlyTextField.php
 /usr/local/opnsense/mvc/app/models/OPNsense/Base/FieldTypes/UrlField.php
 /usr/local/opnsense/mvc/app/models/OPNsense/Base/FieldTypes/VirtualIPField.php
+/usr/local/opnsense/mvc/app/models/OPNsense/Base/Menu/MenuFavorites.php
 /usr/local/opnsense/mvc/app/models/OPNsense/Base/Menu/MenuInitException.php
 /usr/local/opnsense/mvc/app/models/OPNsense/Base/Menu/MenuItem.php
 /usr/local/opnsense/mvc/app/models/OPNsense/Base/Menu/MenuSystem.php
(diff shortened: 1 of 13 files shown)


sopex:
Very cool idea!

AdSchellevis:
The idea is certainly nice, but reviewing a PR of this size takes quite some time. Can't promise anything yet as this topic isn't on our roadmap.

Greelan:
I understand

In any event, I've now just realised that I've introduced user-awareness into the menu system, which seems to be contrary to https://docs.opnsense.org/development/components/menusystem.html. So I expect you will have an issue with that...

sopex:
@Greelan 

Keep this open. This is extremely useful, no matter what some 8 year old documentation says. The priorities were totally different then.

Maybe it should not be implemented exactly this way but at the very least, open an issue/ feature request on this.

Greelan:
My intent is to refactor the code so it is (hopefully) compliant, and create a new PR.

Code development frameworks serve a purpose, so I don't have an issue with that. I just should have RTFM more carefully the first time (so much detail xD)

AdSchellevis:
@Greelan less code is easier to review indeed, I'm just trying to manage expectations here, you can surely just keep the PR open, opening a new one with less code is obviously also fine.

Greelan:
I think it is cleaner in a new PR: #10033

Refactoring allowed me to shorten the PR by 15%, and hopefully achieve much better MVC/development framework compliance xD

Issue #10022, opened 2026-03-23
minor issue: content in dropdown FW -> rule -> source not sorted


jniggemann:
**Important notices**

Before you add a new report, we ask you kindly to acknowledge the following:

- [x] I have read the contributing guide lines at https://github.com/opnsense/core/blob/master/CONTRIBUTING.md
- [x] I am convinced that my issue is new after having checked both open and closed issues at https://github.com/opnsense/core/issues?q=is%3Aissue

**Describe the bug**
On opnsense 26.1.4, aliases and networks are not sorted, see image.
<img width="598" height="698" alt="Image" src="https://github.com/user-attachments/assets/eebe919b-54c5-41be-b318-342a0ffca480" />

sopex:
Have fun :)

`opnsense-patch -a sopex 2bdde785c4645a2970f5c9761aff0605fe12bf20`

fichtner:
dc041207e2 addresses this with some additional concerns we had in the past. Also thanks to @sopex for finding the right spot and doing a PR although I wasn't able to use it in this instance.

    # opnsense-patch dc041207e2


Cheers,
Franco

PR #10031, opened 2026-03-25
firewall: sort aliases/networks

sopex:
Closes: https://github.com/opnsense/core/issues/10022

Sorts the aliases and networks selection while respecting the defaults/ any/ This Firewall


fichtner:
Thanks for this, but I have a few more things to do here looking at the file. Closing in upcoming commit.

Issue #10029, opened 2026-03-24
[BUG] Kea Match DHCP Option Match Data cannot be "0"


sorano:
**Important notices**

Before you add a new report, we ask you kindly to acknowledge the following:

- [X] I have read the contributing guide lines at https://github.com/opnsense/core/blob/master/CONTRIBUTING.md
- [X] I am convinced that my issue is new after having checked both open and closed issues at https://github.com/opnsense/core/issues?q=is%3Aissue

**Describe the bug**

I was trying out the new DHCP Options in KEA and wanted to have specific boot files for BIOS and UEFI.

I looked at the documentation: https://docs.opnsense.org/manual/kea.html#dhcp-options

And that uses 0 as Match Data, but when I try to do the same it wont accept 0.

**Expected behavior**

Accept 0 if it's valid config.

**Screenshots**

<img width="1246" height="513" alt="Image" src="https://github.com/user-attachments/assets/1d14fac6-5cff-4c59-9f99-3fb70024de24" />

**Environment**

OPNsense 26.1.5 (amd64)

Monviech:
@fichtner Maybe the constraint thinks that 0 is an empty value?

sorano:
Update:
Entering 00 into the Match Data field is accepted and works as expected.

Just tested PXEbooting two VM's, one with SeaBIOS and the other with OVMF (UEFI) .

fichtner:
@Monviech yes there is a subtle difference between "set" and "empty"... most of the time we use empty() which for "0" is true for better or worse. But this is certainly fixable.

Relevant code line most likely: https://github.com/opnsense/core/blob/e0b813e6e0e0df084acbe912cfda7bc305b753f8/src/opnsense/mvc/app/models/OPNsense/Base/Constraints/DependConstraint.php#L55

Monviech:
@sorano Thank you for testing the boot file configuration, valuable information that it works as expected. We patched the 0 issue today.

PR #10035, opened 2026-03-25
Services: Kea: DHCPv4/v6: Use SetConstraint for match_data to allow 0 as valid value

Monviech:
Fixes: https://github.com/opnsense/core/issues/10029


Monviech:
yep no problem

Issue #10030, opened 2026-03-24
Kea DHCP DDNS DHCP_DDNS_NO_MATCH No DNS servers match FQDN


oliverjm32:
**Describe the bug**

After setting up DDNS in Kea DHCPv4 I noticed I am getting errors in the Kea logs (see below) and the dns records in my DNS Server are not being updated. 

**Relevant log files**
root@OPNsense:/var/log/kea # cat /usr/local/etc/kea/kea-dhcp4.conf
```json
{
    "Dhcp4": {
        "valid-lifetime": 4000,
        "interfaces-config": {
            "interfaces": [
                "vlan020",
                "vlan010",
                "vlan025",
                "vlan040",
                "ix0"
            ],
            "dhcp-socket-type": "raw"
        },
        "lease-database": {
            "type": "memfile",
            "persist": true
        },
        "control-socket": {
            "socket-type": "unix",
            "socket-name": "\/var\/run\/kea\/kea4-ctrl-socket"
        },
        "loggers": [
            {
                "name": "kea-dhcp4",
                "output_options": [
                    {
                        "output": "syslog"
                    }
                ],
                "severity": "INFO"
            }
        ],
        "subnet4": [
            {
                "id": 1,
                "subnet": "192.168.20.0\/24",
                "next-server": "",
                "match-client-id": true,
                "option-data": [
                    {
                        "name": "domain-name-servers",
                        "data": "192.168.20.254"
                    },
                    {
                        "name": "routers",
                        "data": "192.168.20.254"
                    },
                    {
                        "name": "domain-name",
                        "data": "example.domain"
                    },
                    {
                        "name": "ntp-servers",
                        "data": "192.168.20.254"
                    }
                ],
                "pools": [
                    {
                        "pool": "192.168.20.1-192.168.20.200"
                    }
                ],
                "reservations": []
            },
            {
                "id": 2,
                "subnet": "192.168.10.0\/24",
                "next-server": "",
                "match-client-id": true,
                "option-data": [
                    {
                        "name": "domain-name-servers",
                        "data": "192.168.10.254"
                    },
                    {
                        "name": "domain-search",
                        "data": "example.domain"
                    },
                    {
                        "name": "routers",
                        "data": "192.168.10.254"
                    },
                    {
                        "name": "domain-name",
                        "data": "example.domain"
                    },
                    {
                        "name": "ntp-servers",
                        "data": "192.168.10.254"
                    }
                ],
                "pools": [
                    {
[...]

PR #10038, opened 2026-03-25
Services: Kea: DDNS: Add subnet specific qualifying suffix and prevent updates if no server is set

Monviech:
Fixes: https://github.com/opnsense/core/issues/10030

Adds two parameters that can be set subnet specific, only appends them when ddns is enabled globally.

Example from the docs:

https://kea.readthedocs.io/en/latest/arm/dhcp4-srv.html#ddns-for-dhcpv4

```ddns-send-updates```
```ddns-qualifying-suffix```

```Kea 2.7.6 added support for these parameters to the pool level.```

Support "to" the pool level must mean they are valid in the context of the subnet as well. My generated config shows no errors, but better let users test this.


Monviech:
Multiple users confirmed this patch is working as intended.

PR #10040, opened 2026-03-26
allow pool definitions with same format used in KeaDhcpv4

Astranox:
Kea DHCPv4 allows Pool defintions both as "$ip_start-$ip_end" as well as "$ip_start - $ip_end" (help text suggests the second one)
Kea DHCPv6 only accepts "$ip-$ip" as the help text correctly suggests, but if coming from DHCPv4 the pool is likely to be defined with spaces, which leads to the confusing message "Pool $pool not in specified subnet".

This trims off spaces when checking for pool/subnet match, so both input formats are accepted.
The kea-dhcp6.conf file accepts both formats, although the Kea documentation also has the different input formats between v4 and v6.
It does explicitely state that both formats are accepted in both DHCPv4 and DHCPv6 documentation:
"White space in pool definitions is ignored, so spaces before and after the hyphen are optional. They can be used to improve readability."

### DIFF
--- a/src/opnsense/mvc/app/models/OPNsense/Kea/KeaDhcpv6.php
+++ b/src/opnsense/mvc/app/models/OPNsense/Kea/KeaDhcpv6.php
@@ -79,7 +79,7 @@ class KeaDhcpv6 extends BaseModel
                 if (Util::isSubnet($pool)) {
                     $range = Util::cidrToRange($pool);
                 } else {
-                    $range = explode('-', $pool);
+                    $range = array_map('trim', explode('-', $pool));
                 }
                 foreach (!empty($range) ? $range : [] as $addr) {
                     if (!Util::isIPInCIDR($addr, $subnet->subnet->getValue())) {


src/opnsense/mvc/app/models/OPNsense/Kea/KeaDhcpv6.php
```diff
@@ -79,7 +79,7 @@ public function performValidation($validateFullModel = false)
                 if (Util::isSubnet($pool)) {
                     $range = Util::cidrToRange($pool);
                 } else {
-                    $range = explode('-', $pool);
+                    $range = array_map('trim', explode('-', $pool));
```
fichtner, line 82: I don't want to derail this but I really dislike the magic trim() as a fix for a modelling issue. Normally we reject bad input rigorously. Output format changes are something else. We also have migrations to sanitise data. We don't have to do exactly this in 2026.

fichtner: talking to @AdSchellevis this duplicates code in KeaPoolsField where we should anchor such a helper function to minimise the code here. The commit that introduced this due to Kea being picky was 6a666e80

Astranox: Sure, feel free to change the pull request as you see fit. The incorrect error message / UX issue cost me a minute or so and I thought I'd fix it right away - I just tried to do it similarly as in KeaDhcpv4, which also users array_map and trim.
Personally I'd prefer being able to use the spaces around the hyphens for readability reasons.

fichtner:
@Astranox thanks for your submission! I've tested and pondered a bit and 9ed85133008 seems to be working while moving fragile internals to the KeaPoolsField..  I expect this could break with upstream again following whitespace policy changes, but in that case a better getValues() implementation may be favourable.

Issue #7337, opened 2024-03-18
mvc: htmlspecialchars() use


fichtner:
**Important notices**

Our forum is located at https://forum.opnsense.org , please consider joining discussions there in stead of using GitHub for these matters.

Before you ask a new question, we ask you kindly to acknowledge the following:

- [x] I have read the contributing guide lines at https://github.com/opnsense/core/blob/master/CONTRIBUTING.md
- [x] I am convinced that my issue is new after having checked both open and closed issues at https://github.com/opnsense/core/issues?q=is%3Aissue

Cleaning up a stray copy+paste `html_safe()` reimplementation I noticed that `htmlspecialchars()` is called in a few places either to be encoded in the volt template or through the new `view_html_safe()`. We tend to avoid the PHP function because of special encoding issues in JS and also PHP deprecation warnings causing issues...

```
# git grep htmlspecialchars src/opnsense/mvc
src/opnsense/mvc/app/controllers/OPNsense/Base/ApiControllerBase.php:                    $this->response->setContent(htmlspecialchars(json_encode($data), ENT_NOQUOTES));
src/opnsense/mvc/app/controllers/OPNsense/Core/Api/BackupController.php:                        $result['items'][] = htmlspecialchars($line, ENT_QUOTES | ENT_HTML401);
src/opnsense/mvc/app/controllers/OPNsense/Cron/ItemController.php:        $this->view->selected_uuid = htmlspecialchars($uuid, ENT_QUOTES | ENT_HTML401);
src/opnsense/mvc/app/controllers/OPNsense/Diagnostics/LogController.php:        $this->view->module = htmlspecialchars($module, ENT_QUOTES | ENT_HTML401);
src/opnsense/mvc/app/controllers/OPNsense/Diagnostics/LogController.php:        $this->view->scope = htmlspecialchars($scope, ENT_QUOTES | ENT_HTML401);
src/opnsense/mvc/app/controllers/OPNsense/Firewall/Api/AliasController.php:                    $message .= htmlspecialchars(sprintf("\n[%s] %s", $key, $value), ENT_NOQUOTES | ENT_HTML401);
src/opnsense/mvc/app/controllers/OPNsense/Firewall/Api/FilterBaseController.php:            $descr = htmlspecialchars(!empty($ifdetail->descr) ? $ifdetail->descr : strtoupper($ifname));
[...]

PR #10041, opened 2026-03-26
mvc: deprecate htmlspecialchars() use

Greelan:
Attempt at addressing #7337

### DIFF
--- a/src/opnsense/mvc/app/views/OPNsense/Firewall/filter_rule.volt
+++ b/src/opnsense/mvc/app/views/OPNsense/Firewall/filter_rule.volt
@@ -524,10 +524,11 @@
                         const aliasMetadataList = row["alias_meta_" + column.id] || [];
 
                         const renderedItems = aliasMetadataList.map(aliasInfo => {
+                            const safeText = $('<span>').text(aliasInfo.value || '').html();
                             if (aliasInfo.isAlias) {
-                                const tooltipHtml = aliasInfo.summary || aliasInfo.description || aliasInfo.value || "";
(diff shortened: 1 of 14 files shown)


src/opnsense/mvc/app/views/OPNsense/Firewall/filter_rule.volt
```diff
@@ -524,18 +524,19 @@
                         const aliasMetadataList = row["alias_meta_" + column.id] || [];
 
                         const renderedItems = aliasMetadataList.map(aliasInfo => {
+                            const safeText = $('<span>').text(aliasInfo.value || '').html();
```
fichtner, line 527: In volt we have `{{ var | safe }}` btw...

fichtner: (it doesn't account for API data returns, I know but I wanted to mention it)

fichtner:
Thanks, I'll look at this in two weeks. Busy schedule at the moment. If I don't please ping me again.

src/opnsense/mvc/app/views/OPNsense/Firewall/filter_rule.volt
Greelan: Right, that's why I was happy to rely on `{{ var }}` where relevant

Greelan: Actually, my bad, I'd understood that volt auto-escaped. I've now added explicit `|safe` attributes

Greelan:
@fichtner ping as requested

Issue #9226, opened 2025-09-19
mvc: setActionHook() return message is never shown, because saveFormToEndpoint() doesn't know what it returns


fichtner:
**Important notices**

Our forum is located at https://forum.opnsense.org , please consider joining discussions there in stead of using GitHub for these matters.

Before you ask a new question, we ask you kindly to acknowledge the following:

- [x] I have read the contributing guide lines at https://github.com/opnsense/core/blob/master/CONTRIBUTING.md
- [x] I am convinced that my issue is new after having checked both open and closed issues at https://github.com/opnsense/core/issues?q=is%3Aissue

For Netbird I tried to add "validate-on-set-only" behaviour in the controller and setActionHook() is really good for the purpose. However, instead of injecting custom validation it injects a single 'error' key with a message, but that message is not considered a failure for saveFormToEndpoint(). Make this a bit nicer for future use.

https://github.com/opnsense/core/blob/5eddbce452d36dad8883a785f1f2fa7f42820c6c/src/opnsense/mvc/app/controllers/OPNsense/Base/ApiMutableModelControllerBase.php#L375-L380

Greelan:
See https://github.com/opnsense/core/commit/7397ce01770435b26706c644b2cb20110056bbe6

AdSchellevis:
When needed, we can always try to glue some additional field based response in the hook, but cleaning up the unused looks like a good first step.

fichtner:
@Greelan thanks for pushing this forward -- is this fixed from your POV? I don't mind too much closing this for the already mentioned reasons.


Cheers,
Franco

Greelan:
@fichtner , I will defer to @AdSchellevis' judgement on that score xD

fichtner:
don't be shy :D but fair enough, thanks!

Greelan:
You guys are light years ahead of me in understanding this codebase and how it all fits together, so I know my place lol

AdSchellevis:
closing works for me, thanks!

fichtner:
ding ding ding :)

PR #10046, opened 2026-03-27
mvc: fix setActionHook() return message not shown in UI

Greelan:
Addresses https://github.com/opnsense/core/issues/9226

Hopefully this is the "nicer" approach envisaged xD

### DIFF
--- a/src/opnsense/mvc/app/controllers/OPNsense/Base/ApiMutableModelControllerBase.php
+++ b/src/opnsense/mvc/app/controllers/OPNsense/Base/ApiMutableModelControllerBase.php
@@ -341,8 +341,11 @@ abstract class ApiMutableModelControllerBase extends ApiControllerBase
      * Hook to be overridden if the controller is to take an action when
      * setAction is called. This hook is called after a model has been
      * constructed and validated but before it serialized to the configuration
-     * and written to disk
-     * @return array|string|null validation errors (field-keyed array), error message (string), or null on success
+     * and written to disk. May be used for save-time validation that cannot be
+     * expressed in the model; errors must be pinned to a specific field.
+     * For fatal issues not attributable to a specific field, throw a UserException.
+     * @return array field-keyed validation errors, or null on success
+     * @throws UserException when action is not possible (and save should be aborted)
      */
     protected function setActionHook()
     {
@@ -364,7 +367,7 @@ abstract class ApiMutableModelControllerBase extends ApiControllerBase
      * @return array status / validation errors
      * @throws \OPNsense\Base\ValidationException on validation issues
      * @throws \ReflectionException when binding to the model class fails
-     * @throws UserException when denied write access
+     * @throws UserException when denied write access or when setActionHook() rejects the action
      */
     public function setAction()
     {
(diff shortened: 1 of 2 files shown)


AdSchellevis:
for general exceptions, better raise a `UserException` , e.g. 

https://github.com/opnsense/core/blob/f5c3fb7c75dd0ddc9d2205e7616ee6b65bdb51b2/src/opnsense/mvc/app/controllers/OPNsense/Trust/Api/CaController.php#L137

Greelan:
Understood, though that just reinforces to me the need for clarity on what exceptions that occur via `setActionHook()` should be surfaced by the JS, and what should fall simply to be handled by `UserException`.

The current formulation of `SetAction()` suggests there is a subset of non-validation errors that are intended to be handled not just by `UserException`. Or is that simply an error, and `setActionHook()` should only be used for pre-save validation errors, and hence it should return a `validations` array (or null) and never an error string?

AdSchellevis:
@Greelan I agree about the inconsistency here, as far as I know, we're not use the `error` output for `setBaseHook` anywhere. It looks like a relic from a time long gone to me. 

Since there are no controllers using the functionality from our end, I would plee for a small cleanup here so it's more clear the note in the phpdoc is the way to signal the user of fatal issues:

https://github.com/opnsense/core/blob/b60581625e7bcf5692a3851478a5691bb53f7edb/src/opnsense/mvc/app/controllers/OPNsense/Base/ApiMutableModelControllerBase.php#L356

The distinction in my opinion between (validation) errors and (hard) exceptions should be that we can pinpoint an error to some sort of input (a field), which isn't the case for exceptions. 

If, for some reason, we would like the hook to be able to emit (validation) errors as well, we should add some wrapper ensuring these messages belong to some field at least. So far I don't have a use-case for this myself.

Keep in mind, throwing a `UserException` will show a clear visible error to the user already.

Greelan:
@AdSchellevis is that now what you had in mind?

(As an aside, I *knew* I should have been wary of trying to climb into Franco's mind xD)

AdSchellevis:
@Greelan not exactly, I was planning to remove the return and expect the `setBaseHook` to throw when needed, I think I have been working on this a decade ago. I'll look into this later.

Greelan:
OK. I was trying to address Franco's idea while also ensuring backwards compatibility for any plugins that might currently use the hook

AdSchellevis:
As far as I could find, there are no other implementations of `setBaseHook` in plugins, in which case I go for "spring cleaning here", I'll commit and close this ticket in the process.

Greelan:
Just to be clear, this relates to `setActionHook()`

AdSchellevis:
my mistake, doing two things at the same time doesn't help it seems :) 

Only plugin usage is maltrail which doesn't return anything:

https://github.com/opnsense/plugins/blob/d0d9a7ffb23d2660696f187d5d652aaecc6ff53f/security/maltrail/src/opnsense/mvc/app/controllers/OPNsense/Maltrail/Api/ServerController.php#L39-L64

So, it's only our own code we need to double check.

Issue #10042, opened 2026-03-26
"best effort" priority reflected as "keep current priority" in the new rules UI


kevinchalet:
Hey guys 👋🏻

**Important notices**

Before you add a new report, we ask you kindly to acknowledge the following:

- [x] I have read the contributing guide lines at https://github.com/opnsense/core/blob/master/CONTRIBUTING.md
- [x] I am convinced that my issue is new after having checked both open and closed issues at https://github.com/opnsense/core/issues?q=is%3Aissue

**Describe the bug**

It seems there's a tiny bug in the new rules UI as the "best effort" VLAN PCP value is always reflected as "keep current priority" after saving a rule. The rule itself is correctly created with the `<set-prio>0</set-prio>` node but when editing it, it always appears as "keep current priority", which makes editing rules a bit painful as you always need to re-assign the priority manually (other values don't seem to be affected).

**To Reproduce**

1. Create a firewall rule with the `0` VLAN PCP value.
2. Save the rule.
3. Reload the UI and try to edit the rule.

**Expected behavior**

The correct "best effort" priority should be displayed.

**Screenshots**

When saving:

<img width="612" height="238" alt="Image" src="https://github.com/user-attachments/assets/e84deee8-b5ec-48bc-acc3-bc4b62fdaaf8" />

Once saved:

<img width="611" height="217" alt="Image" src="https://github.com/user-attachments/assets/6b2c24ea-02a0-44e8-8a69-5a8cc4897d2a" />

Interestingly, the tooltip doesn't seem to be affected:

<img width="272" height="126" alt="Image" src="https://github.com/user-attachments/assets/559c6f8f-1f5c-4709-a32e-f02f155c0b60" />

**Environment**

OPNsense 26.1.5 (amd64).

Monviech:
My first guess would be since the values are 0 in the model, that the option field somehow interprets the value 0 as empty and then shows the default

```
            <prio>0</prio>
            <set-prio>0</set-prio>
            <set-prio-low>0</set-prio-low>
```

Monviech:
It looks like something weird is going on, both "" and "0" have the selected key on 1. (Just adding this as notes for me, no reaction needed)

```
        "prio": {
            "": {
                "value": "Any priority",
                "selected": 1
            },
            "1": {
                "value": "Background (1, lowest)",
                "selected": 0
            },
            "0": {
                "value": "Best Effort (0, default)",
                "selected": 1
            },
            "2": {
[...]

PR #10047, opened 2026-03-27
mvc: BaseListField replace empty check with isSet so a 0 value is considered a non empty selection

Monviech:
Fixes: https://github.com/opnsense/core/issues/10042


fichtner:
I assume the tests say this is ok? If so we should also have a test that covers this change.

Monviech:
I did not check the tests yet, I wanted to see what you say about this first :)

fichtner:
From a general perspective this is fine because https://github.com/opnsense/core/blob/f5c3fb7c75dd0ddc9d2205e7616ee6b65bdb51b2/src/opnsense/mvc/app/models/OPNsense/Base/FieldTypes/BaseField.php#L645-L647

Monviech:
It looks like BaseListField itself has no test here

https://github.com/opnsense/core/tree/master/src/opnsense/mvc/tests/app/models/OPNsense/Base/FieldTypes

fichtner:
Since it's a base class it doesn't need direct testing, but then again we also never test getNodeOptions() so it is what it is at the moment ;)

Issue #9840, opened 2026-02-21
Since updating to 26.1 a new Category does not appear in Outbound NAT list of categories


lmoorez:
**Important notices**

Before you add a new report, we ask you kindly to acknowledge the following:

- [Y] I have read the contributing guide lines at https://github.com/opnsense/core/blob/master/CONTRIBUTING.md
- [Y] I am convinced that my issue is new after having checked both open and closed issues at https://github.com/opnsense/core/issues?q=is%3Aissue

**Describe the bug**

I created a new Category, created the incoming and outgoing rules and was able to select the new category from the drop-down list.

Created the Outbound NAT rule and when selecting the Category, the new item does not appear in the list of categories.

This is the first Outbound NAT rule I've created since upgrading from OPNsense 25.7 to 26.1.

Typing in the beginning of the category name results in it appearing in the field and can be selected.

**To Reproduce**

1. Go to 'Firewall -> Categories'
2. Create new entry 
3. Click on 'Firewall -> NAT -> Outbound'
4. Add a new entry
5. Scroll down to 'Category'
6. Click in the Category field for the list of categories.
7. The list does not show the newly created category, or any others created since 26.1.
8. Start typing the name of the new Category and select it when it appears.
9. Make other changes as desired and save the configuration.
10. The category is saved in the new NAT rule.

Selecting an existing Outbound NAT rule and viewing the list of categories also omits the new categories.

Tried the category selection in a different browser and observed the same behaviour.

**Screenshots**

<img width="898" height="267" alt="Image" src="https://github.com/user-attachments/assets/e0836cbc-b49a-4d7c-a708-88c94ca5abf9" />

<img width="751" height="347" alt="Image" src="https://github.com/user-attachments/assets/9a9570d2-129a-4da0-bd2d-d37b76dfa859" />

<img width="783" height="129" alt="Image" src="https://github.com/user-attachments/assets/76a8d851-712e-4e6a-92a9-70167b160cef" />

<img width="1761" height="461" alt="Image" src="https://github.com/user-attachments/assets/397a50c0-8319-4ff1-9195-35ce7612493b" />

**Relevant log files**

None

**Additional context**

My NAT operating mode is "Manual outbound NAT rule generation"

**Environment**

Software version used and hardware type if relevant, e.g.:

OPNsense 26.1.2_5 (amd64).
CPU: Intel(R) Core(TM) i7-7700 CPU @ 3.60GHz (3600.00-MHz K8-class CPU)
[...]

PR #10049, opened 2026-03-27
Firewall: Remove tokenizer from categories and use selectpicker instead

Monviech:
Fixes: https://github.com/opnsense/core/issues/9840

Remove the tokenizer from categories to more easily display all available items and improve search.


AdSchellevis:
in the long run an alternative for the tokenizer would be nice indeed, this package isn't maintained anymore as far as I know and we needed some other workarounds earlier to smoothen some of the glitches.

Monviech:
Let's change the approach then so this does not have to be cleaned up later. I just remove the tokenizer and use the normal selectpicker:

image

PR #10052, opened 2026-03-27
mvc: Remove UIModelGrid imports in IDS, Monit, Syslog SettingsController, unused

Monviech:
Removes the last artifacts of direct UIModelGrid imports, looks like it's unused in those controllers and just a leftover.

There are more use statements that are probably leftover, maybe a linter could catch them?

For: https://github.com/opnsense/core/issues/9797


fichtner:
❤️

fichtner:
Made a few changes for more explicit matching  2c74e4e

PR #10062, opened 2026-03-30
IPsec: Adding legacy dhgroups and hash algorithms to proposals.

jakobsen-lrz:
Hello,

this adds the dhgroups modp768, modp1024, modp1536 and the SHA1 Hash Algorithm to the Proposals. These can be necessary for compatibility.

Thanks.

### DIFF
--- a/src/opnsense/mvc/app/models/OPNsense/IPsec/FieldTypes/IPsecProposalField.php
+++ b/src/opnsense/mvc/app/models/OPNsense/IPsec/FieldTypes/IPsecProposalField.php
@@ -129,11 +129,15 @@ class IPsecProposalField extends BaseListField
             ],
             gettext('Commonly used, but insecure cipher suites') => [
                 'aes256-sha1-modp2048' => 'aes256-sha1-modp2048 [DH14]',
+                'aes192-sha1-modp2048' => 'aes192-sha1-modp2048 [DH14]',
                 'aes128-sha1-modp2048' => 'aes128-sha1-modp2048 [DH14]',
                 'aes256-sha1-modp4096' => 'aes256-sha1-modp4096 [DH16]',
                 'aes256-sha1-ecp521' => 'aes256-sha1-ecp521 [DH21, NIST EC]',
                 'aes256-sha256-modp1024' => 'aes256-sha256-modp1024 [DH2]',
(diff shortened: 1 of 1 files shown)


AdSchellevis:
it's probably better to only add the missing insecure ciphers in the following section:

https://github.com/opnsense/core/blob/67425d3b42496168e61988a3eedeaf5c6f20270e/src/opnsense/mvc/app/models/OPNsense/IPsec/FieldTypes/IPsecProposalField.php#L131

We try to minimize the number of insecure options as much as possible and mark them insecure explicitly to avoid confusion for users.

jakobsen-lrz:
I have made a change, so that the newly added insecure proposals arent grouped together with "Miscellanous" proposals, and put into a seperate group "Miscellanous (Insecure)".

I hope this is better.  This looks good on my firewall. I dont like that the "Miscellanous (Insecure)" group is shown above the "Miscellanous", but i dont know how to fix it.

I think it wouldnt be a good idea to manually add proposals every time someone needs a new specific insecure combination.

AdSchellevis:
@jakobsen-lrz my intention was to only show the ciphers your missing instead of generating them all, it's quite a list we're dumping now which likely isn't relevant for most users. Which (exact) ciphers do you miss?

jakobsen-lrz:
I would need these ciphers:
```
aes256-sha256-modp1536
aes256-sha256-modp1024
aes128-sha1-modp2048
aes192-sha1-modp2048
aes256-sha1-modp2048
aes128-sha256-modp1536
aes256gcm16-sha256-modp1536
```
I could just add them.

jakobsen-lrz:
I didnt notice that some of the proposals were already in the list of insecure proposals. Only these 4 would need to be added:
```
aes192-sha1-modp2048
aes256-sha256-modp1536
aes128-sha256-modp1536
aes256gcm16-sha256-modp1536
```
I changed the PR so now only these 4 would be added to the list, and no other proposals would be generated additionally.

AdSchellevis:
@jakobsen-lrz merged, thanks!

jakobsen-lrz:
thanks for your time

AdSchellevis:
@jakobsen-lrz thank you for your patience :)

Issue #10075, opened 2026-04-01
configctl Drops Empty Positional Parameters, Causing Argument Misalignment


M4jx:
**Important notices**

Before you add a new report, we ask you kindly to acknowledge the following:

- [x] I have read the contributing guide lines at https://github.com/opnsense/core/blob/master/CONTRIBUTING.md
- [x] I am convinced that my issue is new after having checked both open and closed issues at https://github.com/opnsense/core/issues?q=is%3Aissue

### Problem
When configctl is invoked with empty string arguments (e.g., to skip optional positional parameters), the empty strings are silently discarded, causing all subsequent arguments to shift left into the wrong parameter positions.

### Root Cause
[https://github.com/opnsense/core/blob/56ea14fd0bdbc44f5f00f1b0b014761ad962f731/src/opnsense/service/configd_ctl.py#L114](https://github.com/opnsense/core/blob/56ea14fd0bdbc44f5f00f1b0b014761ad962f731/src/opnsense/service/configd_ctl.py#L114)
The `' '.join()` on `args.command` converts empty strings into zero-width gaps between spaces. When configd receives this string and parses it with `shlex.split()`, consecutive spaces are treated as a single delimiter, the empty parameters are lost entirely.

Example: `configctl filter list states '' 5 '' '' ''`

The value 5 lands in --filter instead of --limit.

This affects any action where callers need to skip leading positional parameters. The PHP equivalent (`Backend::configdpRun`) does not have this bug because it uses `escapeshellarg()` which preserves empty strings as '' tokens.

### Suggested Fix
Modify [core/src/opnsense/service/configd_ctl.py](https://github.com/opnsense/core/blob/56ea14fd0bdbc44f5f00f1b0b014761ad962f731/src/opnsense/service/configd_ctl.py#L114)

```
import shlex  # add import

# line 114: replace
exec_commands=[' '.join(args.command)]
# with
exec_commands=[' '.join(shlex.quote(c) for c in args.command)]
```
This mirrors what PHP's `Backend::configdpRun` already does with `escapeshellarg()`. Both preserve empty strings as quoted '' tokens. On the configd daemon side, `shlex.split()` in processhandler.py already correctly reconstructs '' back into empty strings, so no changes are needed there.

AdSchellevis:
@M4jx if you like, you can open a PR with the suggested change as well, it looks like a good addition.

AdSchellevis:
https://github.com/opnsense/core/pull/10076

[...]

PR #10076, opened 2026-04-01
Bugfix #10075

M4jx:
Bugfix for #10075

`shlex.quote` will now make configctl preserve empty positional parameters


AdSchellevis:
@M4jx thanks!

PR #10080, opened 2026-04-02
Services: Kea: DHCPv4/6: remove KeaCtrlAgent dependency on HA configuration

swhite2:
As discussed in https://github.com/opnsense/core/pull/10019

With this there are no core dependencies on the Control Agent left.


Monviech:
@swhite2 @AdSchellevis 

So can this whole model and all dependencies be purged now?

https://github.com/opnsense/core/blob/master/src/opnsense/mvc/app/models/OPNsense/Kea/KeaCtrlAgent.xml

swhite2:
There may still be setups relying on it at the moment. Perhaps purge it in the next major?

AdSchellevis:
> There may still be setups relying on it at the moment. Perhaps purge it in the next major?

Sounds like a good plan to me

PR #10086, opened 2026-04-04
widgets: enhance Live Log widget

### LINKED ISSUE #9684: Live Log widget improvements
**Important notices**

A have a few improvement requests for the GUI Live log widget:

1. Possibility to set number of log messages/rows shown by the widget (currently is hard set to 5)
2. Possibility to set filters in the widget (at least by severity if not custom)
3. Possibility to have a button that when clicked on the widget it will open for us the whole general log

As per above if its feasible and not stupid.

Live with it as is.

**Additional context**
N/A

Greelan:
Context and approach are detailed in #9684 

Closes #9684 

Depends on https://github.com/opnsense/core/commit/1216d603f9ce823d6a19088397a5e866744adfef and https://github.com/opnsense/core/commit/7cfb03be5d7ff77b3692a3ba4c7d532828b7f414

### DIFF
--- a/src/opnsense/mvc/app/controllers/OPNsense/Diagnostics/Api/LogController.php
+++ b/src/opnsense/mvc/app/controllers/OPNsense/Diagnostics/Api/LogController.php
@@ -100,7 +100,8 @@ class LogController extends ApiControllerBase
                     [$offset, $searchPhrase, $module, $scope, $severities],
                     [
                         'Content-Type: text/event-stream',
-                        'Cache-Control: no-cache'
+                        'Cache-Control: no-cache',
+                        'X-Accel-Buffering: no' /* disable nginx reverse proxy buffering for SSE streams to ensure real-time delivery */
                     ],
                     60 /* XXX */
                 );
--- a/src/opnsense/www/js/widgets/LiveLog.js
+++ b/src/opnsense/www/js/widgets/LiveLog.js
@@ -27,46 +27,229 @@
 export default class LiveLog extends BaseTableWidget {
     constructor(config) {
         super(config);
+        this.configurable = true;
+        this._hasData = false;
+        this._severityLevels = [
+            'Emergency', 'Alert', 'Critical', 'Error',
+            'Warning', 'Notice', 'Informational', 'Debug'
+        ];
+        this._sources = {};
+        this._logPrefix = '/ui/diagnostics/log/core/';
+    }
+
+    getGridOptions() {
+        return {
+            sizeToContent: 650
+        };
     }
 
     getMarkup() {
         let $container = $('<div></div>');
-        let $table = this.createTable('live-log-table', {
-            headerPosition: 'top',
-            rotation: 5,
-            headers: [
-                this.translations.time,
-                this.translations.severity,
-                this.translations.process,
-                this.translations.message
-            ]
-        });
-
+        let $table = this.createTable('live-log-table', {headerPosition: 'none'});
         $container.append($table);
         return $container;
     }
 
-    async onMarkupRendered() {
-        const params = new URLSearchParams({
-            offset: 10,
-            searchPhrase: '',
-            severity: ''
-        }).toString();
+    async _fetchSources() {
+        const sources = {};
+        const logFilesLabel = $('a[href$="_LogFiles"]').first().text().trim();
+        for (const item of [...new MenuSystem().walk()].filter((x) => x.href.startsWith(this._logPrefix))) {
+            const scope = item.href.slice(this._logPrefix.length).split('/')[0];
+            const parts = item.breadcrumb().split(' : ');
+            if (parts.length > 2 && parts[parts.length - 2] === logFilesLabel) {
+                sources[scope] = parts[0] + ': ' + parts[parts.length - 1];
+            } else {
+                sources[scope] = parts[0] + ': ' + parts[parts.length - 2];
+            }
+        }
+        this._sources = sources;
+        return sources;
+    }
 
-        super.openEventSource(`/api/diagnostics/log/core/system/${'live?' + params}`, (event) => {
-            if (!event) {
-                super.closeEventSource();
+    async onMarkupRendered() {
+        const [, config] = await Promise.all([
+            this._fetchSources(),
+            this.getWidgetConfig().then(() => this.config?.widget ?? {})
+        ]);
+        this._startLog(config);
+    }
+
+    _renderRow(entry) {
+        const meta = [entry.timestamp, entry.severity, entry.process_name]
+            .filter(v => v && v !== 'null')
+            .join(' \u00b7 ');
+        let $row = $('<span></span>');
+        $row.append($('<span style="font-size:0.85em;color:#888"></span>').text(meta + ' \u00b7 '));
+        $row.append($('<span></span>').html(entry.line));
+        return $row;
+    }
+
+    _defaultSource() {
+        if (this._sources['system']) return 'system';
+        const keys = Object.keys(this._sources);
+        return keys.length > 0 ? keys[0] : null;
+    }
+
+    _startLog(config) {
+        const source = (config.source && this._sources[config.source])
+            ? config.source
+            : this._defaultSource();
+        config.source = source;
+
+        if (source === null) {
+            $(`#link-handle-${this.id}`).hide();
+            super.updateTable('live-log-table', [[this.translations.noaccess]], 'error');
+            this.config.callbacks.updateGrid();
+            return;
+        }
+
+        const lineCount = parseInt(config.lineCount ?? '25', 10);
+        const severity = config.severity ?? 'Notice';
+        const searchFilter = config.searchFilter ?? '';
+        const severityFilter = this._severityLevels
+            .slice(0, this._severityLevels.indexOf(severity) + 1)
+            .join(',');
+
+        $(`#link-handle-${this.id}`).show().find('a').attr('href', `${this._logPrefix}${source}`);
+
+        const sourceLabel = this._sources[source];
+        const severityLabel = this.translations[`severity${severity.toLowerCase()}`] ?? severity;
+        const filterHtml = searchFilter
+            ? ` &nbsp;|&nbsp; <b>${this.translations.filterlabel}:</b> ${$('<span>').text(searchFilter).html()}`
+            : '';
+        const summary =
+            `<div style="text-align:center">` +
+            `<b>${this.translations.sourcelabel}:</b> ${$('<span>').text(sourceLabel).html()} &nbsp;|&nbsp; ` +
+            `<b>${this.translations.thresholdlabel}:</b> ${severityLabel} &nbsp;|&nbsp; ` +
+            `<b>${this.translations.lineslabel}:</b> ${lineCount}` +
+            filterHtml +
+            `</div>`;
+
+        super.updateTable('live-log-table', [[summary]], 'summary');
+        const $summary = $(`#${this.id}_summary`);
+
+        const onEntry = (event) => {
+            let data;
+            try {
+                data = JSON.parse(event.data);
+            } catch (e) {
+                this.closeEventSource();
+                $(`#link-handle-${this.id}`).hide();
+                $(`#${this.id}_summary`).remove();
+                super.updateTable('live-log-table', [[this.translations.noaccess]], 'error');
+                this.config.callbacks.updateGrid();
+                return;
             }
 
-            const data = JSON.parse(event.data);
-            super.updateTable('live-log-table', [
-                [
-                    data.timestamp,
-                    data.severity,
-                    data.process_name,
-                    data.line
-                ]
-            ]);
-        });
+            $(`#${this.id}_error`).remove();
+            $summary.after(
+                $('<div class="flextable-row"></div>')
+                    .append($('<div class="flex-cell"></div>').append(this._renderRow(data)))
+            );
+            const rows = $(`#live-log-table .flextable-row:not(#${this.id}_summary)`);
+            if (rows.length > lineCount) {
+                rows.last().remove();
+            } else {
+                this.config.callbacks.updateGrid();
+            }
+        };
+
+        this._onData = (event) => {
+            if (event.type !== 'message') {
+                this._hasData = true;
+            }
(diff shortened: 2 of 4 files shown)


src/opnsense/mvc/app/controllers/OPNsense/Diagnostics/Api/LogController.php
```diff
@@ -100,7 +100,8 @@ public function __call($name, $arguments)
                     [$offset, $searchPhrase, $module, $scope, $severities],
                     [
                         'Content-Type: text/event-stream',
-                        'Cache-Control: no-cache'
+                        'Cache-Control: no-cache',
+                        'X-Accel-Buffering: no'
```
AdSchellevis: might be good to add a note here this is specifically for users behind an nginx reverse proxy.

src/opnsense/www/js/widgets/LiveLog.js
```diff
@@ -27,46 +27,193 @@
 export default class LiveLog extends BaseTableWidget {
     constructor(config) {
         super(config);
+        this.configurable = true;
+        this._hasData = false;
+        this._severityLevels = [
+            'Emergency', 'Alert', 'Critical', 'Error',
+            'Warning', 'Notice', 'Informational', 'Debug'
+        ];
+        this._sources = [
```
AdSchellevis: It's probably better (more future proof) to wire this to configctl syslog list applications as this should list all registered log targets. Exposing additional log targets looks like a nice idea, but I already see the next request coming to be honest. How to clone the widget to show two at the same time.... (which isn't a reason not to do this, just a request we can't prioritize at the moment)

AdSchellevis (commented):
@Greelan the widget looks nice! I made some comments in the PR, but leave the Javascript code to review for @swhite2 next week.

src/opnsense/www/js/widgets/LiveLog.js
Greelan: I understand the sentiment, though the challenge with that approach is then tying that list to meaningful names in the dropdown (in my PR I have matched the menu names), plus also some endpoints are not listed (notably `boot` and `system`). Also `ppp` comes up as `ppps`?

src/opnsense/mvc/app/controllers/OPNsense/Diagnostics/Api/LogController.php
Greelan: Will do.

src/opnsense/www/js/widgets/LiveLog.js
AdSchellevis: boot is a rather static thing, not processed via syslog, hence the missing entry (likely not very relevant for a live-log either). the names and facilities can indeed be different, but are both stored at all times:

https://github.com/opnsense/core/blob/48eeef5b8203410863707e52a328edc31db66e3a/src/etc/inc/plugins.inc.d/core.inc#L307

Adding an endpoint to decorate the missing information likely helps to fill in the gaps later on when needed, making the log choices static, instantly cuts all plugins from being valid options.

Greelan: OK, I wasn't sure that adding an endpoint would be acceptable. But as always your suggestions will be a definite improvement! I will have a go at implementing.

Greelan: @AdSchellevis I've now made the sources list dynamic. I've adopted a different approach however. I initially started to look at using the syslog applications list but realised that it would need some finessing/filtering to tie it to the visible log pages (eg to avoid duplication where multiple facilities use the same log target, and to deal with missing targets). At the same time, to reduce maintenance overhead, I decided to base the visible label dynamically off the menu breadcrumbs, rather than adding additional data into the $logfacilities array. It then occurred to me that, if I was traversing the menu tree anyway, I may as well use that as the source of the available log endpoints. So that is what I have done. This ensures that the options list in the widget for the log sources reflects what is in the GUI menu list (including logs for plugins), is ordered in the same way for consistency, and is dynamically updated.

Note that the options list is *not* ACL gated - it just gets the full menu tree. This would have been the same if the syslog applications list was used, so I decided not to include ACL gating in `LogController.php`.

AdSchellevis: @Greelan it might be a bit expensive, but the idea is nice, would it help if you could just extract the menu endpoints from javascript using a specific classname? The menu, including the reachable items for the user, is already available when rendering the widget.

Greelan: I actually thought about that too! Certainly it would help in the sense it would only show log types that the user actually has access to. But I was concerned about the client-side information becoming stale if for example a plugin was installed or removed via the CLI. Maybe I am overthinking it. I didn't think getting the server-side information was too much overhead given the contexts in which that occurred

Greelan: I probably am overthinking it ... The menu itself would be stale in the case where a plugin was added or removed via the CLI without a reload, so the widget would just be the same. I assume client side is the preference?

Doing it client side would also mean I could remove the system endpoint gate in Core.xml, though then the JS would have to deal what happens when the log target selected is removed or is no longer accessible (what default would it switch to, eg first available in the list?), and what would happen if all targets were no longer accessible (the dropdown would have to show "No log sources available" and the summary in the widget would need to change). This is where I tied myself in knots before

AdSchellevis: @Greelan I've add this https://github.com/opnsense/core/commit/1216d603f9ce823d6a19088397a5e866744adfef for you, which offers an easy wrapper around the menu system (with currently a limited set of controls).

Your implementation to collect the log endpoints would look something like:

```
[...new MenuSystem().walk().filter((x) => x.href.startsWith('/ui/diagnostics/log/core/'))]
```

It's just a start, but may be practical for other parts eventually too.

Greelan: Thank you! That works great. I've now switched to client-side, so no new API endpoint needed, or ACL concerns.

My next and final step will be to grapple with removing the system log target endpoint gateway, and make the widget available for any level of log file permissions.

AdSchellevis: @Greelan you're welcome, good to hear this helps. A wrapper around the menu system is practical to have anyway, it was a fun thing to work on :)

Greelan: Final changes done and I've cleaned up the commit history given the flip-flopping along the way. Thanks for pushing me to make this more robust!

AdSchellevis: Nice! You're welcome, I just saw an opportunity to offer some useful input.

src/opnsense/www/js/widgets/Metadata/Core.xml
```diff
@@ -228,15 +228,30 @@
...
+            <severityemergency>Emergency</severityemergency>
+            <severityalert>Alert</severityalert>
+            <severitycritical>Critical</severitycritical>
+            <severityerror>Error</severityerror>
+            <severitywarning>Warning</severitywarning>
+            <severitynotice>Notice</severitynotice>
+            <severityinformational>Informational</severityinformational>
+            <severitydebug>Debug</severitydebug>
+            <searchfilter>Search filter</searchfilter>
+            <searchfilterplaceholder>dhcp error</searchfilterplaceholder>
```
swhite2: I see what you're trying to do here, but perhaps it looks a little out of place. Maybe just nothing? Not sure myself

Greelan: The aim was simply to show format of the filter text (space separated words). I'm not wedded to the specific content. But having nothing for a placeholder will not give any guidance. Afaik help text is not available in widget options. And I didn't want to clutter the heading with format guidance.

Greelan:
Any chance this will get some love?

src/opnsense/www/js/widgets/LiveLog.js
```diff
@@ -27,46 +27,229 @@
...
+            $(`#link-handle-${this.id}`).hide();
+            super.updateTable('live-log-table', [[this.translations.noaccess]], 'error');
+            this.config.callbacks.updateGrid();
+            return;
+        }
+
+        const lineCount = parseInt(config.lineCount ?? '25', 10);
+        const severity = config.severity ?? 'Notice';
+        const searchFilter = config.searchFilter ?? '';
+        const severityFilter = this._severityLevels
```
swhite2: These should already default to the right thing through config. If `config` is an empty object (`onMarkupRendered`) I can imagine this is an error condition.

src/opnsense/www/js/widgets/LiveLog.js
```diff
@@ -27,46 +27,229 @@
...
+            const parts = item.breadcrumb().split(': ');
+            if (parts.length > 2 && parts[parts.length - 2] === logFilesLabel) {
+                sources[scope] = parts[0] + ': ' + parts[parts.length - 1];
+            } else {
+                sources[scope] = parts[0] + ': ' + parts[parts.length - 2];
+            }
+        }
+        this._sources = sources;
+        return sources;
+    }
```
swhite2, line 67: This seems to produce the following:

Image

Notice the System log files aren't differentiated (they should be `Audit`, `Backend` etc.)

src/opnsense/www/js/widgets/LiveLog.js
```diff
@@ -27,46 +27,229 @@
...
+        this._startLog(config);
+    }
+
+    _renderRow(entry) {
+        const meta = [entry.timestamp, entry.severity, entry.process_name]
+            .filter(v => v && v !== 'null')
+            .join(' \u00b7 ');
+        let $row = $('<span></span>');
+        $row.append($('<span style="font-size:0.85em;color:#888"></span>').text(meta + ' \u00b7 '));
+        $row.append($('<span></span>').html(entry.line));
```
swhite2, line 83: I think we should avoid calling `html()` here as `entry.line` could theoretically be something user-controlled. Haven't tested it, but `.text()` should work as well

src/opnsense/www/js/widgets/LiveLog.js
```diff
@@ -27,46 +27,229 @@
...
+                }
+            });
+        }
+    }
+
+    onVisibilityChanged(visible) {
+        if (!visible) {
+            this._hasData = false;
+        }
+        super.onVisibilityChanged(visible);
```
swhite2: I think theoretically _hasData can be set to `true` by the `_onData` handler before the eventsource is closed, probably best to call `onVisibilityChanged` first.

swhite2 (changes requested):
@Greelan Sorry for the wait, this slipped through on my end..

Couple of points, but the widget looks very good, thanks :)

src/opnsense/www/js/widgets/Metadata/Core.xml
swhite2: Can we rename this to "e.g. dhcp error"?

src/opnsense/www/js/widgets/LiveLog.js
Greelan: You're correct, of course, I included it as belt and braces. Can remove it.

src/opnsense/www/js/widgets/LiveLog.js
Greelan: Looks like you didn't have https://github.com/opnsense/core/commit/7cfb03be5d7ff77b3692a3ba4c7d532828b7f414 applied xD

swhite2: Yep, the branch was missing a rebase/merge :D

src/opnsense/www/js/widgets/LiveLog.js
Greelan: I did that because one of the logs (can't recall which, will need to check when I am back at my desktop) had html formatting that was rendered improperly without it. In any event, afaik the data is already properly escaped via Response.php?

src/opnsense/www/js/widgets/LiveLog.js
Greelan: Agreed, will make the change

src/opnsense/www/js/widgets/Metadata/Core.xml
Greelan: Can do

src/opnsense/www/js/widgets/LiveLog.js
swhite2: Fair enough. Indeed the proper safeguards are in place.

src/opnsense/www/js/widgets/LiveLog.js
```diff
@@ -27,46 +27,229 @@
...
-            searchPhrase: '',
-            severity: ''
-        }).toString();
-
-        super.openEventSource(`/api/diagnostics/log/core/system/${'live?' + params}`, (event) => {
-            if (!event) {
-                super.closeEventSource();
+        const [, config] = await Promise.all([
+            this._fetchSources(),
+            this.getWidgetConfig().then(() => this.config?.widget ?? {})
```
swhite2:
```suggestion
            this.getWidgetConfig()
```

It always returned an empty object on first widget load (instead of the defaults). `getWidgetConfig()` already handles defaults vs. stored values so you only need the returned object

swhite2:
Merged, thanks!

src/opnsense/mvc/app/controllers/OPNsense/Diagnostics/Api/LogController.php
```diff
@@ -100,7 +100,8 @@ public function __call($name, $arguments)
                     [$offset, $searchPhrase, $module, $scope, $severities],
                     [
                         'Content-Type: text/event-stream',
-                        'Cache-Control: no-cache'
+                        'Cache-Control: no-cache',
+                        'X-Accel-Buffering: no' /* disable nginx reverse proxy buffering for SSE streams to ensure real-time delivery */
```
swhite2, line 104: One more question @Greelan, any particular reason this needs to be solved here instead of on the proxy itself?

EDIT: I missed that you already mentioned it in the ticket. This likely needs ironing out on the other stream endpoints then as well.

PR #10089, opened 2026-04-04
Added NAS-IP-Address attribute to radius access server calls

metmayhem:
Added NAS-IP-Address attribute to radius access server calls
Changed NAS-Port-Type to virtual for radius access server calls

No open issues I could find regarding this, but I needed NAS-IP-Address to be set for Microsoft NPS to work. I also found that the NAS-Port-Type was hard coded to RADIUS_ETHERNET which didn't work with NPS either. I changed it to be hard coded RADIUS_VIRTUAL for access servers and kept the default of RADIUS_ETHERNET for everything else.

### DIFF
--- a/src/opnsense/mvc/app/library/OPNsense/Auth/Radius.php
+++ b/src/opnsense/mvc/app/library/OPNsense/Auth/Radius.php
@@ -172,8 +172,8 @@ class Radius extends Base implements IAuthConnector
             'radius_acct_port' => 'acctPort',
             'radius_protocol' => 'protocol',
             'radius_stationid' => 'calledStationId',
-	    'radius_nasipaddress' => 'nasIpAddress',
-	    'radius_nasporttype' => 'nasPortType',
+            'radius_nasipaddress' => 'nasIpAddress',
+            'radius_nasporttype' => 'nasPortType',
             'refid' => 'nasIdentifier'
         );
 
--- a/src/opnsense/mvc/app/library/OPNsense/Auth/Radius.php
+++ b/src/opnsense/mvc/app/library/OPNsense/Auth/Radius.php
@@ -70,6 +70,11 @@ class Radius extends Base implements IAuthConnector
     private $callingStationId = null;
 
     /**
+     * @var string ip addess to use for NAS-IP-Address attribute
+     */
+    private $nasIpAddress = null;
+
+    /**
      * @var int timeout to use
      */
     private $timeout = 10;
@@ -108,6 +113,8 @@ class Radius extends Base implements IAuthConnector
      * @var array list of groups to add by default
      */
     private $syncDefaultGroups = [];
+
+    private $nasPortType = RADIUS_ETHERNET;
 
     private function mapTerminateCause($cause)
     {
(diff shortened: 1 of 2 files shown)


src/opnsense/mvc/app/library/OPNsense/Auth/Radius.php
```diff
@@ -109,6 +114,8 @@ class Radius extends Base implements IAuthConnector
      */
     private $syncDefaultGroups = [];
 
+    private $nasPortType = RADIUS_ETHERNET;
```
AdSchellevis: better move this to the implementation as it appears to be static anyway now. e.g.

```
if(!radius_put_int($radius, RADIUS_NAS_PORT_TYPE, RADIUS_VIRTUAL))....
```

src/www/system_authservers.php
```diff
@@ -875,6 +885,15 @@
                     </div>
                   </td>
                 </tr>
+                <tr class="auth_radius auth_options hidden">
```
AdSchellevis: you should be able to add the field in `getConfigurationOptions()` including some basic validations. 

For an example, see:

https://github.com/opnsense/core/blob/a57e59d7106c46f7aaf627b4476f33b7f340cc09/src/opnsense/mvc/app/library/OPNsense/Auth/Voucher.php#L423-L434

AdSchellevis (changes requested):
@metmayhem it needs some modifications to ease future maintenance, but adding the field shouldn't be an issue.

src/opnsense/mvc/app/library/OPNsense/Auth/Radius.php
metmayhem: The reasoning I had behind doing it this way was because I wasn't sure what else used the Radius class. This preserves the original result while allowing me to modify it for the access server section. I can change it, though.

I was also debating on changing the access server page to have a drop down with all the available flags for port type, in case someone has filtering in place on the radius server and need a specific port type. The current fix handles my use-case, but the drop down might make it more future proof.

metmayhem: I went ahead and hard-coded it as requested.

src/www/system_authservers.php
metmayhem: I've moved it to getConfigurationOptions. I also changed it to a dropdown with all the available interface IPs listed.

metmayhem (commented):
Requested changes have been made. Also changed NAS IP Address field to a dropdown that contains all available interface IPs.

src/opnsense/mvc/app/library/OPNsense/Auth/Radius.php
```diff
@@ -27,6 +27,7 @@
  */
 
 namespace OPNsense\Auth;
+require_once("interfaces.inc");
```
AdSchellevis, line 30: it's not allowed to insert legacy files inside new MVC based code, but I also don't think we need it to be honest. If we just validate the input for validity, we don't have to select a static list of addresses (and make sure the user can also use virtual addresses as well).

I'll have a look at the code and see if I can fix the concerns.

AdSchellevis:
@metmayhem can you try https://github.com/opnsense/core/commit/95483e5048e3b2ebad3609886c4647f7a3feb762 ? It should implement what you need.

metmayhem:
It looks to be working. I did kind of like my addition of the dropdown (prevents human error during setup), but this gets the job done. Thanks

Issue #7316, opened 2024-03-10
OpenVPN: Add More Advanced Options to Instances


jasker5183:
**Important notices**

Before you add a new report, we ask you kindly to acknowledge the following:

- [x] I have read the contributing guide lines at https://github.com/opnsense/core/blob/master/CONTRIBUTING.md
- [x] I am convinced that my issue is new after having checked both open and closed issues at https://github.com/opnsense/core/issues?q=is%3Aissue

**Is your feature request related to a problem? Please describe.**

My VPN provider uses a few advanced options that with Instances I am unable to use. These include:

````
auth-nocache
remote-cert-tls server ## Getting warnings in my log because I don't have this ##
nobind
resolv-retry infinite
explicit-exit-notify 5
push-peer-info
comp-lzo no
data-ciphers CHACHA20-POLY1305:AES-256-GCM:AES-256-CBC:AES-192-GCM:AES-192-CBC:AES-128-GCM:AES-128-CBC
data-ciphers-fallback AES-256-CBC
````

I would also include:

````
mlock
mute-replay-warnings ## Too many AEAD Decrypt error: bad packet ID (may be a replay) in my log ##
replay-window 128 ## Might fix the AEAD errors ##
````

**Describe the solution you like**

Would just like the ability to use these advanced options but since they have been removed in Instances I cannot.

**Describe alternatives you considered**

I can add these options to the .conf files in /var/etc/openvpn but they will be removed if I ever make changes in the WebGUI.

**Additional context**

[OpenVPN 2.6 Reference Manual](https://openvpn.net/community-resources/reference-manual-for-openvpn-2-6/)

AdSchellevis:
Options have better change of getting including when explaining why they should be in there (which problem their aimed to solve). We have seen quite some options that have very little value or sometimes just do not reach their set goals under FreeBSD. Some of the mentioned options are already available under advanced by the way...

jasker5183:
These three options I would consider the most useful:

````
auth-nocache - Don't cache --askpass or --auth-user-pass username/passwords in virtual memory.

remote-cert-tls server - WARNING: No server certificate verification method has been enabled. See http://openvpn.net/howto.html#mitm for more info. Is the warning I'm getting in my logs, should be self-explanatory.

mlock - Using this option ensures that key material and tunnel data are never written to disk due to virtual memory paging operations which occur under most modern operating systems.
````

[...]

Issue #6758, opened 2023-08-19
VPN: OpenVPN: Instances missing a few options


tomatotoast:
**Important notices**

Before you add a new report, we ask you kindly to acknowledge the following:

- [x] I have read the contributing guide lines at https://github.com/opnsense/core/blob/master/CONTRIBUTING.md
- [x] I am convinced that my issue is new after having checked both open and closed issues at https://github.com/opnsense/core/issues?q=is%3Aissue

**Is your feature request related to a problem? Please describe.**

While migrating to the openvpn instances i noticed that the advanced section is missing. This means that some options in the gui are missing.

**The options are missing:**
_port-share
tcp-queue-limit
bcast-buffers_

to configure.

For other users the options are:
_tun-mtu
mssfix_

may also be interesting.

**Describe the solution you like**

Please add the options mentioned above to the new instances menu.

**Describe alternatives you considered**
alternative: add the advances section again

**Additional context**

OPNsense 23.7.1_3-amd64
FreeBSD 13.2-RELEASE-p2
OpenSSL 1.1.1v 1 Aug 2023

AdSchellevis:
We aim to keep the gui clean and as simple as possible, adding options needs stronger arguments to avoid ending up with a toggles nobody uses or should use (quite some toggles in OpenVPN are not very useful and people only set them because historically they had a use, but nowadays they don't). The legacy option will be around for quite some time.

Maybe we will offer a custom config file hook at some point to allow all options being used, a bit like we offer for other services as well, but no final decisions have been made on that topic.

tomatotoast:
Im not the only one needing openvpn port-share. i found this forum post:
https://forum.opnsense.org/index.php?topic=35438.msg172002#msg172002

I have a few setups running on a single IP with openvpn running on tcp 443 (so that users are always able to connect even when in an restriced like hotel network).
unfortunately since nginx or haproxy is also running i need to share the port.

My advanced config looks like this:

port-share 127.0.0.1 8443
tcp-queue-limit 4096
bcast-buffers 4096

The other options are used for performance tuning.

AdSchellevis:
out of curiosity, what does the rest of your config look like? I don't mind keeping a ticket open for discussion, but I would advise changing it to only include the port-share feature as that's what you seem to be looking for.

tomatotoast:
Apart from the three advanced options, my configuration is actually normal client vpn.

[...]

PR #10100, opened 2026-04-07
openvpn: add ADAPTER_DOMAIN_SUFFIX support to Instances module

berzdev:
## What does this PR do?

Adds an "Adapter Domain Suffix" field to the OpenVPN Instances server configuration. When set, the server pushes `dhcp-option ADAPTER_DOMAIN_SUFFIX <domain>` to clients.

## Why is this needed?

Windows clients using OpenVPN with DCO (Data Channel Offload) cannot resolve DNS short names (e.g. `server`) because the DCO driver does not support DHCP-based domain search list configuration. The only reliable workaround is pushing `ADAPTER_DOMAIN_SUFFIX`, which sets the connection-specific DNS suffix via wmic on the Windows client.

This has been an open upstream issue for 3+ years ([OpenVPN/openvpn#326](https://github.com/OpenVPN/openvpn/issues/326)) with no fix in sight. The DCO maintainer has explicitly declined to add DHCP support to the driver.

The legacy OpenVPN Server module allowed Custom Options as a workaround. The Instances module (which replaces it) does not, leaving administrators with no GUI-based solution.

## Changes

- **Model** (`OpenVPN.xml`): Added `adapter_domain_suffix` field of type `HostnameField` with list support
- **Config generator** (`OpenVPN.php`): Emit `push "dhcp-option ADAPTER_DOMAIN_SUFFIX <domain>"` for each configured value
- **Form** (`dialogInstance.xml`): Added tokenized multi-select field in the DNS options section, visible only for server instances

## How to test

1. Configure an OpenVPN server instance with the new "Adapter Domain Suffix" field set to `example.local`
2. Connect a Windows client with OpenVPN Connect (DCO enabled)
3. Run `ipconfig /all` – verify "Connection-specific DNS Suffix" shows `example.local` on the DCO adapter
4. Run `ping server` – verify it resolves to `server.example.local`

## Related issues

- Partially addresses https://github.com/opnsense/core/issues/7316
- Partially addresses https://github.com/opnsense/core/issues/6758
- Upstream: https://github.com/OpenVPN/openvpn/issues/326, https://github.com/OpenVPN/openvpn/issues/306

### DIFF
--- a/src/opnsense/mvc/app/controllers/OPNsense/OpenVPN/forms/dialogInstance.xml
+++ b/src/opnsense/mvc/app/controllers/OPNsense/OpenVPN/forms/dialogInstance.xml
@@ -645,6 +645,20 @@ Set to 0 to disable, remember to change your client as well.
         </grid_view>
     </field>
     <field>
+        <id>instance.adapter_domain_suffix</id>
+        <label>Adapter Domain Suffix</label>
+        <type>select_multiple</type>
+        <style>tokenize role role_server</style>
+        <allownew>true</allownew>
+        <help>
+            Push ADAPTER_DOMAIN_SUFFIX to clients. Sets the connection-specific DNS suffix on the VPN adapter,
+            required for reliable DNS short name resolution on Windows clients using the DCO (Data Channel Offload) driver.
+        </help>
+        <grid_view>
+            <visible>false</visible>
+        </grid_view>
+    </field>
+    <field>
         <id>instance.tun_mtu</id>
         <label>TUN device MTU</label>
         <type>text</type>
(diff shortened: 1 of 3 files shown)


AdSchellevis:
`ADAPTER_DOMAIN_SUFFIX` is an alias for `DOMAIN` which is already available (https://openvpn.net/community-docs/community-articles/openvpn-2-6-manual.html):

https://github.com/opnsense/core/blob/d092f74bb1a5205ea0ff25398123f5b54b1698fc/src/opnsense/mvc/app/models/OPNsense/OpenVPN/OpenVPN.php#L686

swhite2:
After testing this on a customer end yesterday, neither `DOMAIN` nor `DOMAIN-SEARCH` works with OpenVPN Connect (using DCO). One should use the community client instead. Even if `ADAPTER_DOMAIN_SUFFIX` would magically solve this, it wouldn't fix the `DOMAIN-SEARCH` case.

berzdev:
Thank you @swhite2 ! I didn't know before that there is a community client for OpenVPN. That works without problems.

PR #10103, opened 2026-04-08
ui: use space in apply box for the apply reminder

fichtner:
We could extend this a bit if we want to go this way. It's the best way to use existing space and move the message to where it matters most -- the apply button itself.

I'm not entirely sure about coloring.  We could instead make the whole box change color similar to what alerts are doing.

wide:

Screenshot 2026-04-08 at 12 15 35

squashed:

Screenshot 2026-04-08 at 12 15 43

Based on discussions with @Monviech and more recently @swhite2 since the alert box moved for related reasons.

### DIFF
--- a/src/opnsense/www/js/opnsense_bootgrid.js
+++ b/src/opnsense/www/js/opnsense_bootgrid.js
@@ -2056,25 +2056,7 @@ class UIBootgrid {
     showSaveAlert(event) {
         let editAlert = this.$compatElement.attr('data-editAlert');
         if (editAlert !== undefined) {
-            const $el = $("#" + editAlert);
-            const rect = $(".page-content-head").first()[0].getBoundingClientRect();
-            const top = $('.navbar').first().outerHeight() + 5;
-            const centerX = rect.left + (rect.width / 2);
-
-            $el.css({
-                position: "fixed",
-                top: top + "px",
-                left: centerX + "px",
-                right: "auto",
-                zIndex: 9999,
-                transform: "translateX(-50%)"
-            });
-
-            $el.slideDown(1000, function () {
-                setTimeout(function () {
-                    $el.not(":animated").slideUp(2000);
-                }, 2000);
-            });
+            $("#" + editAlert).show();
         }
     }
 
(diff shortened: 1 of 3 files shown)


src/opnsense/www/js/opnsense_ui.js
```diff
@@ -684,7 +684,9 @@ $.fn.SimpleActionButton = function (params) {
                         setIcon(icon, 'fa fa-check fa-spinner fa-pulse', 'fa fa-spinner fa-pulse');
                     } else {
                         setIcon(icon, 'fa fa-spinner fa-pulse', 'fa fa-check');
+                        $("#" + this_button.data('message-id')).hide();
 
+                        // XXX instead of infamous check mark we could show a success or even failure message
```
Monviech: plz don't touch my checkmark I love it ;_;

fichtner: advocating a bit for moving the checkmark outside the button to an appropriate success message, but both are fine in the end... it depends a little how far we want to take this concept

src/opnsense/www/js/opnsense_bootgrid.js
```diff
@@ -2056,25 +2056,7 @@ class UIBootgrid {
...
-                zIndex: 9999,
-                transform: "translateX(-50%)"
-            });
-
-            $el.slideDown(1000, function () {
-                setTimeout(function () {
-                    $el.not(":animated").slideUp(2000);
-                }, 2000);
-            });
+            $("#" + editAlert).show();
```
Monviech: Just driving by this quickly.

I like the idea of getting rid of the aggressive sliding like this until the button is clicked.

Keep in mind that some pages trigger the sliding in their own volt pages though for bootgrid unrelated reasons.

fichtner: sure, looks manageable.. just need to make a decision here first if that's the way to go or not.. but first.. vacation ;)

fichtner:
Did a second commit just to try the following style...

Screenshot 2026-04-09 at 09 00 58

swhite2:
> Did a second commit just to try the following style...

I like it :)

swhite2 (approved):
No objections code-wise, also happy with the latest result, works just fine on mobile as well. If anyone disagrees, let them speak now..

Monviech (approved):
I feel like this is a good idea and direction, thank you.

src/opnsense/www/js/opnsense_ui.js
fichtner:
```suggestion
```

PR #10111, opened 2026-04-09
Finish GHSA-jpm7-f59c-mp54 remediation

mandreko:
During the publishing of [GHSA-jpm7-f59c-mp54](https://github.com/opnsense/core/security/advisories/GHSA-jpm7-f59c-mp54), the temporary private fork from GitHub was not used. This change was lost from the changeset that I provided. This should help to finish closing that issue out entirely.

### DIFF
--- a/src/opnsense/mvc/app/library/OPNsense/Auth/LDAP.php
+++ b/src/opnsense/mvc/app/library/OPNsense/Auth/LDAP.php
@@ -471,26 +471,6 @@ class LDAP extends Base implements IAuthConnector
     }
 
     /**
-     * unused
-     * @return array mixed named list of authentication properties
-     */
-    public function getLastAuthProperties()
-    {
-        return $this->lastAuthProperties;
-    }
-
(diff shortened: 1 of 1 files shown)


fichtner:
One of the reasons this was not immediately considered is that f271c6a3fc4c seems to make a very related change an intentional bugfix so this requires more testing.

AdSchellevis:
I just tried to remove the `authenticate()` on my test machine and run successful validations against LDAP and LDAP+TOTP, so it looks like we'er good now.

When I added https://github.com/opnsense/core/commit/f271c6a3fc4cb1f1656eaa46303cfb2669ab7249 a couple of years ago, it was more a safety precaution as we didn't want to risk more breakage if I'm not mistaken. The required fix was renaming `authenticate` to `_authenticate`, which is what we achieve with this change. 

So, I'm ok with merging this now.

fichtner:
Merged, thanks!

PR #10033, opened 2026-03-25
menu: add Favorites section

Greelan:
Implements a Favorites section:

- The section only appears if the user has at least one favorite selected. If all favorites are removed, the Favorites section disappears
- Favorites are added/removed simply by clicking on a star icon at the right hand side of the page heading on leaf menu item pages (the star will always appear furthest right, for consistency of appearance, even if other buttons are in the heading). The star icon changes appearance if the item is in the Favorites list
- The Favorites list is stored per-user, and is subject to their permissions
- If a favorited menu item is removed (eg an interface is removed or a plugin uninstalled), the Favorites entry is also removed
- Favorites entries appear in the same order as the corresponding menu items
- The PR caters for both MVC and legacy pages

Some sample screenshots attached.

Screenshot 5
Screenshot 6

Supersedes #10028

Depends on https://github.com/opnsense/core/commit/1c2405e48076daae925d92a9109306500fe45835

Closes #9543

PR #10113, opened 2026-04-09
MVC: add support for pluggable dynamic menu items and move some existing parts out of the MenuSystem class

AdSchellevis:
In most cases we use static menu registrations, but there are exceptions which depend on interfaces for example. While looking at https://github.com/opnsense/core/pull/10033, a longer standing wish came up again, which is the reason to add this support right now. It also helps in removing some legacy components for good via plugins.

To register new menu items, the following pattern may be used:

* In your model, derive a Menu class from MenuContainer
* implement a method collect() which should add new menu items via the appendItem() {bound to appendItem in MenuSystem}

Always try to minimize the amount of code inside these plugins as this code will be executed on each page load.

To explain the concept, both interface and firewall menu items have been moved to their new classes in this PR.

### DIFF
--- a/src/opnsense/mvc/app/models/OPNsense/Base/Menu/MenuContainer.php
+++ b/src/opnsense/mvc/app/models/OPNsense/Base/Menu/MenuContainer.php
@@ -1,0 +1,49 @@
+<?php
+
+/*
+ * Copyright (C) 2026 Deciso B.V.
+ * All rights reserved.
+ *
+ * Redistribution and use in source and binary forms, with or without
+ * modification, are permitted provided that the following conditions are met:
+ *
+ * 1. Redistributions of source code must retain the above copyright notice,
+ *    this list of conditions and the following disclaimer.
+ *
+ * 2. Redistributions in binary form must reproduce the above copyright
+ *    notice, this list of conditions and the following disclaimer in the
+ *    documentation and/or other materials provided with the distribution.
+ *
+ * THIS SOFTWARE IS PROVIDED ``AS IS'' AND ANY EXPRESS OR IMPLIED WARRANTIES,
+ * INCLUDING, BUT NOT LIMITED TO, THE IMPLIED WARRANTIES OF MERCHANTABILITY
+ * AND FITNESS FOR A PARTICULAR PURPOSE ARE DISCLAIMED. IN NO EVENT SHALL THE
+ * AUTHOR BE LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY,
+ * OR CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF
+ * SUBSTITUTE GOODS OR SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS
+ * INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN
+ * CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE)
+ * ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
+ * POSSIBILITY OF SUCH DAMAGE.
+ */
+
+namespace OPNsense\Base\Menu;
+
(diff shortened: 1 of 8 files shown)


src/opnsense/mvc/app/models/OPNsense/Base/Menu/MenuContainer.php
```diff
@@ -0,0 +1,49 @@
...
+ * CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE)
+ * ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
+ * POSSIBILITY OF SUCH DAMAGE.
+ */
+
+namespace OPNsense\Base\Menu;
+
+abstract class MenuContainer
+{
+    private ?MenuSystem $menusystem = null;
```
Greelan, line 33: Given this is private, how will the favorites code access `getItem()` to build the breadcrumb labels for the Favorites menu?

AdSchellevis: @Greelan see `appendItem()` examples in Firewall and Interfaces

Greelan: Thanks, I've looked at those. I think the favorites scenario is different, because the labels can't just be read from the config, instead they need to be built from the existing menu tree (eg a favorite pointing to `/ui/firewall/filter` should show as `Firewall: Rules [new]`). Perhaps I'm missing something?

AdSchellevis: It might need some additional logic to prevent the favorite item needing the same uri as the actual target, but we'll see it when we'll get there. One of the options might be to just store a (md5) hash of the link and reflect the click action for example. 

The favorite description itself could be part of the stored content to prevent the need to render the favorites if everything else is already there (since other dynamic content may have been set to favorite as well).

Greelan: Agree it will need further thinking down the track. I wanted to avoid storing the label in case it became stale (menu item breadcrumbs changed).

Saying that though made me realise that there is a similar issue - though less likely - for the endpoint itself, eg when a MVC migration occurs. It's a bit different since the pruning logic will clean it up, but the consequence is that the menu item will be un-favourited. Not fatal but not ideal from a UX perspective.

AdSchellevis: @Greelan this https://github.com/opnsense/core/pull/10113/commits/f1ba1bdac397d2bb9363d86213ddf4a983199bdf is likely the missing part for the favorites. The commit message contains some context, with this you can easily reference menu items and should be able to capture clicks on the favorite menu items (and pass them along)

fichtner (approved):
Does LinkClass also work for the migration assistant icon?

src/opnsense/mvc/app/models/OPNsense/Base/Menu/MenuContainer.php
Greelan: Thanks! I will have a closer look at it when I get a chance.

AdSchellevis:
> Does LinkClass also work for the migration assistant icon?

Nope, it's a class on the link, not on its content. I was doubting refactoring the whole menu, but since I wanted to limit the impact, only pushed the concepts in there that are currently missing. Eventually it would be practical to render this from Javascript and offer an endpoint which contains all relevant items, but that's quite some work I'm afraid.

src/opnsense/mvc/app/models/OPNsense/Base/Menu/MenuContainer.php
Greelan: OK, I've now spent some time with this and have done some re-work to use the pluggable system, link classes, and JS wrapper. Will need to wait until the next OPNsense release for testing as it is too challenging to patch an existing installation with the needed upstream commits, given they build on other commits as well

fichtner: Just FYI I don't think this will land in 26.1.7 next week but 26.1.8 is very likely.

PR #10118, opened 2026-04-10
Don't select invalid gateway as active one

ASunc:
OPNsense GUI displays nicely that a gateway has invalid settings when it's IP address is outside interfaces address range (checked agains netmask). However, the backend logic does not work similarly. If multiple upstream gateways have been configured
and the topmost one (having highest priority) is not valid, system still tries to use it which renders a valid gateway configuration having lower priority useless.

Added code (with my *very* limited PHP skills) to Gateways.php to perform similar check as done in SettingsController.php already. It marks gateway as disabled if it's address falls outside of interfaces netmask and it has not been marked as a far gateway.

My use case for this scenario is a HA pair in datacenter where master and backup nodes have different upstream gateway address. This allows me to configure upstream gateways in master (with master's gateway as priority 10 and backup's gateway as priority 11). When configuration is synced to backup, this change detects that priority 10 gateway is not valid, causing system to select next gateway - which is the correct one for backup node.

As said, I'm not very skilled with PHP, so I'm not sure if this code is usable as-is. But I hope that the idea is.

This is tested on 25.7.11.

### DIFF
--- a/src/opnsense/mvc/app/models/OPNsense/Routing/Gateways.php
+++ b/src/opnsense/mvc/app/models/OPNsense/Routing/Gateways.php
@@ -33,6 +33,7 @@ use OPNsense\Core\Config;
 use OPNsense\Firewall\Util;
 use OPNsense\Interface\Autoconf;
 use OPNsense\Base\Messages\Message;
+use OPNsense\Core\Backend;
 
 class Gateways extends BaseModel
 {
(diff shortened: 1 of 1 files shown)


swhite2:
https://github.com/opnsense/core/commit/8f8a4c8c35de3a1212d5942385a041d0cf1a8273 solves this specific scenario as well, thanks for the report!

PR #10123, opened 2026-04-10
Kea: add explicit reverse DDNS zones

XtraLarge:
## Summary

Add an explicit `DNS reverse zone` field to Kea subnets and generate `reverse-ddns.ddns-domains` from that configured value.

Ref: #10122

## Why

The original proposal derived reverse zones automatically from configured subnets.

Based on maintainer feedback, this was changed to avoid implicit reverse-zone inference because the authoritative DNS reverse zone size and delegation boundaries cannot be assumed safely.

## What this changes

- keeps the existing forward DDNS generation unchanged
- adds a `DNS reverse zone` field to Kea DHCPv4 and DHCPv6 subnet dialogs
- generates `reverse-ddns.ddns-domains` only from explicitly configured reverse zones
- reuses the same DNS target and TSIG key configuration already used for forward DDNS

## Validation

Tested on my live OPNsense box on 2026-04-11:

- PHP syntax check for the installed `KeaDdns.php`
- XML validation for the updated model and form files
- applied explicit `ddns_reverse_zone` values to the active Kea subnet config
- regenerated Kea config via `kea_configure_do(true)`
- verified `/usr/local/etc/kea/kea-dhcp-ddns.conf` now contains the explicit reverse zones and still points to the configured DDNS target `127.0.0.1:53530`
- verified `kea-dhcp4` and `kea-dhcp-ddns` remain active after reload

Note: the live box also carries a separate local DDNS target port patch that is not part of this PR, because BIND in this setup listens on port `53530`.

PR #10117, opened 2026-04-10
Kea: allow configuring DDNS target port per subnet

XtraLarge:
## What
Add a configurable DDNS target port to Kea DHCPv4 and DHCPv6 subnets.

## Why
The current DDNS generator always writes `dns-servers[].port` as `53`, even though the authoritative DNS server may intentionally listen on a different port. This makes otherwise valid local resolver/authoritative split setups impossible to express in the GUI.

## How
- add a `ddns_dns_server_port` model field for DHCPv4 and DHCPv6 subnets
- expose the field in the subnet dialogs
- use the configured port in `KeaDdns::generateConfig()`
- keep the default at `53` for backward compatibility

## Notes
This keeps the existing behavior unchanged for current installations and only extends the generated `kea-dhcp-ddns.conf` when a non-default port is configured.

Ref: opnsense/core#10116

PR #10132, opened 2026-04-11
kea: add FQDN override flags for server-side PTR updates

### LINKED ISSUE #7768: Support for DDNS in Kea
**Important notices**

I am currently using ISC DHCPD which has support in OPNsense for dynamically updating my DNS servers with assigned IP addresses.

I would like for support for kea-dhcp-ddns to be added for the purposes of feature parity with the existing ISC DHCPD.

Looking at the documentation relating to this part, the required configuration interface is
1. Able to enter one or more TSIG keys, identified by TSIG key name and containing algorithm and secret.
2. Able to enter one or more forward domain names, along with specifying the DNS servers to use and the TSIG key name.
3. The exact same as (2), but for reverse domain names.

The only alternative for now is to stay with ISC DHCPD. Or to move to a separately hosted DHCP server.

**Additional context**
See relevant Kea documentation: https://kea.readthedocs.io/en/latest/arm/ddns.html

### LINKED ISSUE #9359: dhcp/kea:  Improve DDNS support

mbedworth:
## Summary

- **KeaDhcpv4.php**: Add `ddns-override-no-update`, `ddns-override-client-update`, and `ddns-update-on-renew` to the DHCPv4 config when DDNS is enabled.
- **KeaDhcpv6.php**: Apply the same three flags to the DHCPv6 config generator (previously missing entirely).

## Problem

When DDNS is enabled via the Kea DHCP GUI, PTR (reverse) records are often not created. Many DHCP clients (macOS, Windows, iOS) set FQDN option flags (`N=1` or `S=0`) instructing the server not to perform reverse updates. Without `ddns-override-no-update` and `ddns-override-client-update`, Kea respects these flags and skips PTR creation entirely.

This affects both DHCPv4 and DHCPv6. On the IPv6 side the problem is more severe: no AAAA or PTR records are registered for any dynamic DHCPv6 lease, even when `fqdn_fwd=1` is set on the lease, because the three override flags were absent from `KeaDhcpv6.php::generateConfig()` entirely.

## Solution

When DDNS is enabled, add three global settings to both `kea-dhcp4.conf` and `kea-dhcp6.conf`:

- `ddns-override-no-update: true` — override client N flag (RFC 4702 §3.1)
- `ddns-override-client-update: true` — override client S flag (RFC 4702 §3.2)
- `ddns-update-on-renew: true` — re-send DDNS updates on lease renewal

These flags are already supported by the upstream Kea model fields (`ddns_override_no_update`, etc.) — this change wires them into both config generators so they take effect when DDNS is active.

Note: this PR does not add automatic reverse zone derivation. Reverse zone names are inherently site-specific; the correct approach is for operators to configure them explicitly via the existing forward zone fields once #10123 or equivalent reverse zone UI support lands.

## Test plan

- [x] Configure DDNS on a DHCPv4 subnet with BIND9 as the DNS server
- [x] Verify `kea-dhcp4.conf` contains all three override flags
- [x] Verify `kea-dhcp6.conf` contains all three override flags
- [x] Force a DHCP renewal from a macOS client: `sudo ipconfig set en0 DHCP`
- [x] Verify forward lookup works: `dig @localhost <hostname>.<domain> A`
- [x] Verify forward lookup works: `dig @localhost <hostname>.<domain> AAAA`
- [x] Verify reverse lookup works: `dig @localhost -x <ipv4>`
- [x] Verify reverse lookup works: `dig @localhost -x <ipv6>`
- [x] Verify configs survive `configctl kea restart`

Tested on OPNsense 26.1.6 with Kea 3.0.2 and BIND 9.20.

Fixes #7768
Fixes #9359

### DIFF
--- a/src/opnsense/mvc/app/models/OPNsense/Kea/KeaDhcpv4.php
+++ b/src/opnsense/mvc/app/models/OPNsense/Kea/KeaDhcpv4.php
@@ -362,6 +362,14 @@ class KeaDhcpv4 extends BaseModel
                 'server-ip' => $ddns->general->server_ip->getValue(),
                 'server-port' => $ddns->general->server_port->asInt(),
             ];
+            /* Override client FQDN flags to ensure the server performs both
+             * forward and reverse DNS updates, regardless of client preferences.
+             * Without these, many clients (macOS, Windows) set the N or S flags
+             * in the FQDN option to skip reverse updates, leaving PTR records
+             * empty. See RFC 4702 sections 3.1-3.2. */
+            $cnf['Dhcp4']['ddns-override-no-update'] = true;
+            $cnf['Dhcp4']['ddns-override-client-update'] = true;
+            $cnf['Dhcp4']['ddns-update-on-renew'] = true;
         }
         File::file_put_contents($target, json_encode($cnf, JSON_PRETTY_PRINT), 0600);
     }
--- a/src/opnsense/mvc/app/models/OPNsense/Kea/KeaDdns.php
+++ b/src/opnsense/mvc/app/models/OPNsense/Kea/KeaDdns.php
@@ -101,6 +101,28 @@ class KeaDdns extends BaseModel
                             $reverse_domains[$rev_zone] = $reverse_domain;
                         }
                     }
+                } elseif (!empty($subnet_cidr) && str_contains($subnet_cidr, ':')) {
+                    /* IPv6: derive ip6.arpa zone from subnet prefix.
+                     * Use /48 boundary for the reverse zone since that is the
+                     * standard delegation size for ULA and most ISP allocations.
+                     * Multiple /64 subnets under the same /48 share one reverse zone. */
+                    $parts = explode('/', $subnet_cidr);
+                    $expanded = @inet_pton($parts[0]);
+                    if ($expanded !== false) {
+                        $hex = bin2hex($expanded);
+                        /* Use /48 (12 nibbles) as the reverse zone boundary */
+                        $nibbles = 12;
+                        $prefix_nibbles = substr($hex, 0, $nibbles);
+                        $rev_zone = implode('.', array_reverse(str_split($prefix_nibbles))) . '.ip6.arpa.';
+                        if (!isset($reverse_domains[$rev_zone])) {
+                            $reverse_domain = ['name' => $rev_zone, 'dns-servers' => []];
+                            if ($keyname) {
+                                $reverse_domain['key-name'] = $keyname;
+                            }
+                            $reverse_domain['dns-servers'][] = $server_entry;
+                            $reverse_domains[$rev_zone] = $reverse_domain;
+                        }
+                    }
                 }
             }
         }
--- a/src/opnsense/mvc/app/models/OPNsense/Kea/KeaDdns.php
+++ b/src/opnsense/mvc/app/models/OPNsense/Kea/KeaDdns.php
@@ -39,6 +39,7 @@ class KeaDdns extends BaseModel
             return;
         }
         $domains = [];
+        $reverse_domains = [];
         $keys = [];
         foreach ([(new KeaDhcpv4())->subnets->subnet4, (new KeaDhcpv6())->subnets->subnet6] as $subnets) {
             foreach ($subnets->iterateItems() as $subnet) {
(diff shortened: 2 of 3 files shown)


AdSchellevis:
same as https://github.com/opnsense/core/pull/10117 ?

mbedworth:
These are different features — #10117 adds a configurable DNS target port per subnet, while this PR adds automatic reverse-DDNS zone generation and FQDN override flags. Both happen to touch `KeaDdns.php` so there will be a merge conflict once either lands. Happy to rebase this on top of #10117 if that one merges first.

AdSchellevis:
it looks like I dropped the wrong PR, it's this one that overlaps https://github.com/opnsense/core/pull/10123

mbedworth:
Thanks for the pointer to #10123 — I reviewed it and agree there's overlap.

I've reworked this PR to drop the automatic reverse zone derivation entirely. Auto-deriving  zones from subnet CIDRs was the right idea locally but is too opinionated for the general case (split-horizon DNS, delegated sub-/24s, etc.).

What remains is only the FQDN override flags in `KeaDhcpv4.php`:
- `ddns-override-no-update`
- `ddns-override-client-update`
- `ddns-update-on-renew`

These are a separate concern from reverse zone config — they control whether Kea respects client FQDN option flags (RFC 4702) or overrides them server-side. Without them, PTR updates silently fail for most Apple/Windows clients regardless of how reverse zones are configured. The model fields already exist; this change just wires them into the config generator when DDNS is active.

No overlap with #10123 remains — this PR no longer touches `KeaDdns.php` at all.

src/opnsense/mvc/app/models/OPNsense/Kea/KeaDhcpv4.php
```diff
@@ -362,6 +362,14 @@ public function generateConfig($target = '/usr/local/etc/kea/kea-dhcp4.conf')
                 'server-ip' => $ddns->general->server_ip->getValue(),
                 'server-port' => $ddns->general->server_port->asInt(),
             ];
+            /* Override client FQDN flags to ensure the server performs both
+             * forward and reverse DNS updates, regardless of client preferences.
+             * Without these, many clients (macOS, Windows) set the N or S flags
+             * in the FQDN option to skip reverse updates, leaving PTR records
+             * empty. See RFC 4702 sections 3.1-3.2. */
+            $cnf['Dhcp4']['ddns-override-no-update'] = true;
```
Monviech, line 370: I would prefer if these configuration values could be added here:

https://github.com/opnsense/core/blob/master/src/opnsense/mvc/app/controllers/OPNsense/Kea/forms/ddnsSettings.xml

I want these user configurable (globally only). You can emit them via "ddns->general->" ... just like the server_ip and server_port.

I don't need the code comment, these configuration options are well documented in the KEA documentation.

Monviech (changes requested):

Monviech:
Since it can only be true or false the config generations can remain a single line per entry. Even if false they can always be printed. Here an example how I want it:

https://github.com/opnsense/core/blob/22f6b063c441f22b68646987973630fc64f29aad/src/opnsense/mvc/app/models/OPNsense/Kea/KeaDhcpv4.php#L209

Monviech:
I implemented it for you in here https://github.com/opnsense/core/commit/a27a2c7901614d26bd911e276f752450169ef9a2

mbedworth:
Hi Monviech,

Thank you for implementing this — the per-subnet checkbox approach is a well-structured solution that clarifies user options effectively.

One thing worth flagging: operators unfamiliar with this new configuration might inadvertently leave the per-subnet checkboxes unchecked, leading to silent PTR registration failures with macOS/Windows/iOS clients. These clients send FQDN option flags instructing the server not to perform reverse updates, and without the overrides explicitly enabled, PTR records simply won't be created.

It might be worth documenting this on the DDNS settings page, or exploring whether these flags should default to on when a DDNS server is configured per-subnet — a follow-up PR could address that if the idea has traction.

Monviech:
I'll add it to the documentation tomorrow. I dont think we should do other defaults than KEA, which is what I implemented (KEAs defaults).

Also these checkboxes unhide with advanced mode at the same time as the reverse zone, so that will peak the operators interest :)

mbedworth:
Makes sense.

PR #10134, opened 2026-04-12
ui: apply button

sopex:
Hides apply buttons that leak to subtabs and small UI fixes around these buttons to make the UI more homogenous.

### DIFF
--- a/src/opnsense/mvc/app/views/OPNsense/Firewall/alias.volt
+++ b/src/opnsense/mvc/app/views/OPNsense/Firewall/alias.volt
@@ -630,6 +630,14 @@
 
         $('.nav-tabs a').on('shown.bs.tab', function (e) {
             history.pushState(null, null, e.target.hash);
+            switch (e.target.hash) {
+                case '#actions':
+                    $('#reconfigureAct').closest('.content-box').hide();
+                    break;
+                default:
+                    $('#reconfigureAct').closest('.content-box').show();
+                    break;
+            }
         });
 
         // move filter into action header
--- a/src/opnsense/mvc/app/views/OPNsense/Hostdiscovery/settings.volt
+++ b/src/opnsense/mvc/app/views/OPNsense/Hostdiscovery/settings.volt
@@ -45,7 +45,11 @@
 
         $('a[data-toggle="tab"]').on('shown.bs.tab', function (e) {
             switch (e.target.hash) {
+                case '#settings':
+                    $('#reconfigureAct').closest('.content-box').show();
+                    break;
                 case '#hosts':
+                    $('#reconfigureAct').closest('.content-box').hide();
                     if (!$("#grid-hosts").hasClass('tabulator')) {
                         $("#grid-hosts").UIBootgrid({
                         search:'/api/hostdiscovery/service/search',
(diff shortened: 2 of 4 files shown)


fichtner:
This will break with https://github.com/opnsense/core/pull/10103 and I think if we want this we need to make it generic. Similar concern was voiced before but in general I’m not sure if it’s worth the trouble implementing. Hiding could always introduce subtle bugs that make the apply disappear more permanently when targeted.

sopex:
> This will break with #10103 and I think if we want this we need to make it generic. Similar concern was voiced before but in general I’m not sure if it’s worth the trouble implementing. Hiding could always introduce subtle bugs that make the apply disappear more permanently when targeted.

I think it's worth adding an id to #10103, people don' t like buttons that do nothing :)

fichtner:
a class would be more flexible, but yes

Monviech:
It's important (for me) that the Apply button is always visible and clickable. I don't want hiding or disabling logic.

Apply can be necessary even outside of config changes to regenerate templates and reload the service (the service buttons dont exist on all pages and what they do is more opaque imo).

I spent quite some effort to make sure the button is at the same spot everywhere and behaves the same, I don't want to shoot this down at the moment.

sopex:
> It's important (for me) that the Apply button is always visible and clickable. I don't want hiding or disabling logic.
> 
> Apply can be necessary even outside of config changes to regenerate templates and reload the service (the service buttons dont exist on all pages and what they do is more opaque imo).
> 
> I spent quite some effort to make sure the button is at the same spot everywhere and behaves the same, I don't want to shoot this down at the moment.

Everyone is working on Sunday :(

I am not sure why we are not trusting the MVC model, the whole page is just a show/hide div.
Plus, 3/4 pages have a restart button.

Maybe only IMO but this just seems like broken UI, not the thoughtful Apply button I am sure you wanted to add:

image

Monviech:
It's just a slippery slope, I want the apply button to be as dumb and predictable as possible.

When we start to hook it conditionally again it's going to be a slippery slope (next would be why not hide it on more pages, eg when config is not dirty etc...), even if the intentions are good.

Just wanted to add my two cents without blocking anything.

sopex:
I understand what you are saying, I agree with you, and I like that the apply button is in its predictable place with its nice box.

I just believe some cases, like these, and like the "Log file" pages of the services (where apply does not exist). Don't need it, and its presence becomes confusing. 
Maybe this is where the line should be drawn on not needed + visually confusing, otherwise uniformity needs to be kept.

Obviously, fair enough to have an opinion, you have done way too much work to stay idle :)

Monviech:
If its pages with two tabs just put the apply button template into the scope of the tab it should show. I think I've already done it like this in some spots. Its a one line pragmatic code change with no side effects.

sopex:
> If its pages with two tabs just put the apply button template into the scope of the tab it should show. I think I've already done it like this in some spots. Its a one line pragmatic code change with no side effects.

I tried that, it kills your nice detached box, and it's like appending it to the existing box. 
Plus, there are 3 tab pages.
I know I am not helping :)

Monviech:
I was under the assumption it worked when done correctly, just dropping the apply button template into the div of the tab.

Also the apply button template lets you define button IDs, so you can reuse it in multiple tabs without any overlap.

I can give this a look at some point if there are challenged here. I think just dropping the template in where it should be shown is more pragmatic than JS show/hide from a maintainability standpoint, its more declarative.

fichtner:
Thinking about this more and looking the diff it would be more beneficial to fully unify apply button use first before trying to micro optimise single pages for cosmetic preference.

fichtner:
See https://github.com/opnsense/core/pull/10328

sopex:
Closed for #10331

Issue #2038, opened 2017-12-31
Intrusion Detection: possibility to choose eve log as log output with several types (alert, http, dns, tls, files, ssh)


ifzenc:
Hi,
It could be nice to be able to select the log output as eve.json instead of fast.log and be able to select syslog instead of file. Additionally, the types should be selectable.
A potential output in suricata.yaml would be: 

```
 - eve-log:
      enabled: yes
      type: syslog #file|syslog|unix_dgram|unix_stream
      filename: eve.json
      # the following are valid when type: syslog above
      identity: "suricata"
      facility: local5
      level: Info ## possible levels: Emergency, Alert, Critical,
                   ## Error, Warning, Notice, Info, Debug
      types:
        - alert:
             payload: yes
             payload-buffer-size: 100kb
             payload-printable: yes
             http: yes
             tls: yes
        - http:
            extended: yes     # enable this for extended logging information
            # custom allows additional http fields to be included in eve-log
            # the example below adds three additional fields when uncommented
            #custom: [Accept-Encoding, Accept-Language, Authorization]
        - dns
        - tls:
            extended: yes     # enable this for extended logging information
        - files:
            force-magic: no   # force logging magic on all logged files
            force-md5: no     # force logging of md5 checksums
        #- drop
        - ssh
```

This is very useful if sending logs to a SIEM like Splunk as json is automatically parsed.

Thank you.

Additional info:
Version OPNSense 17.7.11
URL: ui/ids/

AdSchellevis:
Files are already in eve format (/var/log/suricata/eve.json), which is used by our log viewer. syslog currently uses fast.log (as mentioned in the help text). 

I'm not sure you can select the syslog output and still keep the file output we have now, which is needed for our product. Another option is to ship the log files using filebeat or similar from disk to the SIEM product.

ifzenc:
Thanks for your reply.
You're probably right it doesn't seem to be possible to have simultaneously syslog and file for eve format. I understand that you use the file for the log viewer, but it could be great to have the choice in the UI as well as the type of events to log.

In my configuration, I don't want to write logs on disk as it uses flash disk, that's why I chose syslog for eve.

AdSchellevis:
timeout

PR #10136, opened 2026-04-12
ids: add DNS, flow, stats EVE output types and community-id toggle

mbedworth:
## Summary

Extends the EVE log configuration with additional output types, following the pattern established for HTTP and TLS in #7775.

**New configurable EVE output types** (all disabled by default, under Advanced):

- **DNS** — logs DNS transactions with optional query/answer sub-toggles
- **Flow** — logs network flow records (duration, byte counts, protocol)
- **Stats** — logs Suricata engine statistics (capture counters, decoder stats, memory usage)

**Community ID** — makes the `community-id` field configurable via the GUI (was hardcoded to `false`). [Community ID](https://github.com/corelight/community-id-spec) is a standardised flow hash that enables cross-tool correlation between Suricata, Zeek, and other network security tools.

## Changes

| File | Change |
|------|--------|
| `IDS.xml` | Add `dns`, `flow`, `stats`, `communityId` fields under `eveLog`; bump version 1.1.2 → 1.1.3 |
| `suricata.yaml` | Add conditional Jinja2 blocks for new EVE types; make `community-id` configurable |
| `generalSettings.xml` | Add 6 new form fields under Advanced toggle |
| `M1_1_3.php` | No-op migration — new fields use model defaults |

## Motivation

SIEM integrations (Wazuh, Elastic, Splunk) benefit from richer EVE output beyond alerts. DNS logging enables threat hunting (tunneling, DGA detection), flow records provide network visibility, and stats enable performance monitoring. Community ID is increasingly expected by correlation tools.

The existing `custom.yaml` escape hatch cannot be used for this because Suricata's YAML merge fully overrides the `outputs` key rather than merging lists, so any `custom.yaml` approach would lose the default alert/anomaly/drop/ssh types.

This was originally raised in #2038 (2018, closed as timeout). PR #7775 solved it for HTTP and TLS — this PR extends the same pattern to the remaining commonly-requested types.

## Testing

Verified on OPNsense 26.1.6 / Suricata 8.0.4:

- Template renders correct YAML with all combinations of toggles enabled/disabled
- New model fields accessible via the API (`/api/ids/settings/get`)
- Migration runs cleanly on upgrade from 1.1.2
- **Live traffic test** (dual 1Gbps WAN, ~1400 events captured):
  - `flow`: 1,357 events with `community_id` hashes
  - `dns`: 53 events with query/answer and `community_id`
  - `stats`: engine statistics logged to EVE
  - `community_id` present in 1,412/1,413 events (all network events; stats correctly excluded)
  - Existing types (alert, anomaly, drop, ssh) unaffected


src/opnsense/mvc/app/models/OPNsense/IDS/Migrations/M1_1_3.php
```diff
@@ -0,0 +1,43 @@
+<?php
```
AdSchellevis: this file doesn't do anything here, default migrations are always triggered

mbedworth: Good catch, removed — the new fields default correctly via the model and the parent migration handles version tracking automatically. Thanks.

PR #10140, opened 2026-04-12
services: migrate opendns page to MVC/API

Greelan:
Replicates the current logic, with two enhancements:
- consolidates the registration logic into the one Apply button (with error handling)
- saves/restores previous DNS settings (assuming the user didn't enable standalone mode when enabling OpenDNS)

PR being opened for feedback on approach before full testing.

Closes #10108


src/opnsense/mvc/app/controllers/OPNsense/OpenDNS/Api/SettingsController.php
```diff
@@ -0,0 +1,68 @@
...
+     * runs after model validation; throws to abort the save when registration fails
+     */
+    protected function setActionHook()
+    {
+        $mdl = $this->getModel();
+        if ((string)$mdl->general->enabled !== '1') {
+            return;
+        }
+
+        require_once('plugins.inc.d/opendns.inc');
```
AdSchellevis: legacy imports are not allowed, if we need to signal legacy or plugin code, we need to pipe them via configd for consistency.

src/opnsense/mvc/app/controllers/OPNsense/OpenDNS/Api/SettingsController.php
```diff
@@ -0,0 +1,68 @@
...
+        $errors = [];
+        foreach (explode("\n", $result) as $line) {
+            $line = trim($line);
+            if ($line === '' || strpos($line, 'good') === 0 || $line === 'noop') {
+                continue;
+            }
+            $errors[] = $line;
+        }
+        if (!empty($errors)) {
+            throw new UserException(implode("\n", $errors), gettext('Error registering with OpenDNS.com using the credentials and network name shown above. Data not saved.'));
```
AdSchellevis: I would opt for our default workflow, persist and validate, similar to other forms like interfaces settings

https://github.com/opnsense/core/blob/master/src/opnsense/mvc/app/controllers/OPNsense/Interfaces/Api/SettingsController.php

https://github.com/opnsense/core/blob/f1a3150fa9b2401d515ab651affa125469decc1e/src/opnsense/mvc/app/views/OPNsense/Interface/settings.volt#L58-L64

src/opnsense/mvc/app/models/OPNsense/OpenDNS/OpenDNS.xml
```diff
@@ -0,0 +1,30 @@
+<model>
+    <mount>//OPNsense/OpenDNS</mount>
```
AdSchellevis: I think I would leave the data in it's original spot, probably saves you a migration too.

AdSchellevis (changes requested):
@Greelan just did a quick review to add some notes, no rush. Thanks for looking into this, it seems like you're having fun here :)

src/opnsense/mvc/app/controllers/OPNsense/OpenDNS/Api/SettingsController.php
Greelan: Thanks, I didn't want to persist the config until the external opendns.com registration succeeded, that's why the different approach. Didn't seem a good idea to save a config that didn't work.

src/opnsense/mvc/app/controllers/OPNsense/OpenDNS/Api/SettingsController.php
Greelan: Ah, that one slipped through when I transitioned to `setActionHook()` and got rid of a separate register script

src/opnsense/mvc/app/models/OPNsense/OpenDNS/OpenDNS.xml
Greelan: Still will need a smaller migration to deal with the legacy handling of booleans

PR #10149, opened 2026-04-12
unbound: blocklist improvements

### LINKED ISSUE #10133: unbound: blocklist improvements
**Important notices**

The way the default blocklists are presented to users is chaotic and causes option paralysis to both new and experienced users.

I have some ideas.

sopex:
**Important notices**

*Organizes DNSBLs by provider/category.
*Adds the Social Network blocklist by hegizi.
~~*Fixes: /ui/unbound/dnsbl/index#blocklist_tester having an extra apply button :)~~
*The tester now gives you the DNSBL name and provider instead of its shortcode.

**Related issue**
Closes: #10133

### DIFF
--- a/src/opnsense/mvc/app/controllers/OPNsense/Unbound/Api/DiagnosticsController.php
+++ b/src/opnsense/mvc/app/controllers/OPNsense/Unbound/Api/DiagnosticsController.php
@@ -128,6 +128,9 @@ class DiagnosticsController extends ApiControllerBase
                 $types = $mdl->dnsbl->blocklist->getTemplateNode()->type->getNodeData();
                 foreach ($response as $key => $value) {
                     if (isset($value['bl'])) {
+                        if (isset($types[$value['bl']]['optgroup'])) {
+                            $response[$key]['provider'] = $types[$value['bl']]['optgroup'];
+                        }
                         $response[$key]['bl'] = $types[$value['bl']]['value'] ?? $value['bl'];
                     }
                     if (isset($value['uuid'])) {
--- a/src/opnsense/mvc/app/controllers/OPNsense/Unbound/Api/DiagnosticsController.php
+++ b/src/opnsense/mvc/app/controllers/OPNsense/Unbound/Api/DiagnosticsController.php
@@ -127,13 +127,11 @@ class DiagnosticsController extends ApiControllerBase
                 $policies = $mdl->dnsbl->blocklist->getNodeContent();
                 $types = $mdl->dnsbl->blocklist->getTemplateNode()->type->getNodeData();
                 foreach ($response as $key => $value) {
-                    if (is_array($value)) {
-                        if (isset($value['bl'])) {
-                            $response[$key]['bl'] = $types[$value['bl']]['value'] ?? $value['bl'];
-                        }
-                        if (isset($value['uuid'])) {
-                            $response[$key]['policy'] = $policies[$value['uuid']]['description'] ?? '';
-                        }
+                    if (isset($value['bl'])) {
+                        $response[$key]['bl'] = $types[$value['bl']]['value'] ?? $value['bl'];
+                    }
+                    if (isset($value['uuid'])) {
+                        $response[$key]['policy'] = $policies[$value['uuid']]['description'] ?? '';
                     }
                 }
                 return $response;
--- a/src/opnsense/mvc/app/controllers/OPNsense/Unbound/Api/DiagnosticsController.php
+++ b/src/opnsense/mvc/app/controllers/OPNsense/Unbound/Api/DiagnosticsController.php
@@ -129,7 +129,7 @@ class DiagnosticsController extends ApiControllerBase
                 foreach ($response as $key => $value) {
                     if (isset($value['bl'])) {
                         if (isset($types[$value['bl']]['optgroup'])) {
-                            $response[$key]['provider'] = $types[$value['bl']]['optgroup'];
+                            $response[$key]['category'] = $types[$value['bl']]['optgroup'];
                         }
                         $response[$key]['bl'] = $types[$value['bl']]['value'] ?? $value['bl'];
                     }
--- a/src/opnsense/mvc/app/controllers/OPNsense/Unbound/Api/DiagnosticsController.php
+++ b/src/opnsense/mvc/app/controllers/OPNsense/Unbound/Api/DiagnosticsController.php
@@ -123,6 +123,19 @@ class DiagnosticsController extends ApiControllerBase
             ]), true);
 
             if (!empty($response)) {
+                $mdl = new \OPNsense\Unbound\Unbound();
+                $policies = $mdl->dnsbl->blocklist->getNodeContent();
+                $types = $mdl->dnsbl->blocklist->getTemplateNode()->type->getNodeData();
+                foreach ($response as $key => $value) {
+                    if (is_array($value)) {
+                        if (isset($value['bl'])) {
+                            $response[$key]['bl'] = $types[$value['bl']]['value'] ?? $value['bl'];
+                        }
+                        if (isset($value['uuid'])) {
+                            $response[$key]['policy'] = $policies[$value['uuid']]['description'] ?? '';
+                        }
+                    }
+                }
                 return $response;
             }
         }
--- a/src/opnsense/mvc/app/controllers/OPNsense/Unbound/Api/OverviewController.php
+++ b/src/opnsense/mvc/app/controllers/OPNsense/Unbound/Api/OverviewController.php
@@ -101,7 +101,7 @@ class OverviewController extends ApiControllerBase
 
         foreach ($parsed as $idx => $query) {
             if (isset($types[$query['blocklist']]['optgroup'])) {
-                $parsed[$idx]['category'] = $types[$query['blocklist']]['optgroup'];
+                $parsed[$idx]['provider'] = $types[$query['blocklist']]['optgroup'];
             }
             $parsed[$idx]['blocklist'] = $types[$query['blocklist']]['value'] ?? $query['blocklist'];
             $parsed[$idx]['policy'] = $policies[$query['uuid']]['description'] ?? '';
--- a/src/opnsense/mvc/app/controllers/OPNsense/Unbound/Api/OverviewController.php
+++ b/src/opnsense/mvc/app/controllers/OPNsense/Unbound/Api/OverviewController.php
@@ -72,8 +72,12 @@ class OverviewController extends ApiControllerBase
             return [];
         }
 
+        $types = $this->mdl->dnsbl->blocklist->getTemplateNode()->type->getNodeData();
         foreach ($parsed['top_blocked'] as $domain => $props) {
-            $parsed['top_blocked'][$domain]['blocklist'] ??= $this->getBlocklistDescription($props['blocklist']);
+            if (isset($types[$props['blocklist']]['optgroup'])) {
+                $parsed['top_blocked'][$domain]['category'] = $types[$props['blocklist']]['optgroup'];
+            }
+            $parsed['top_blocked'][$domain]['blocklist'] = $types[$props['blocklist']]['value'] ?? $props['blocklist'];
         }
 
         return $parsed;
--- a/src/opnsense/mvc/app/controllers/OPNsense/Unbound/Api/OverviewController.php
+++ b/src/opnsense/mvc/app/controllers/OPNsense/Unbound/Api/OverviewController.php
@@ -100,6 +100,9 @@ class OverviewController extends ApiControllerBase
         $types = $this->mdl->dnsbl->blocklist->getTemplateNode()->type->getNodeData();
 
         foreach ($parsed as $idx => $query) {
+            if (isset($types[$query['blocklist']]['optgroup'])) {
+                $parsed[$idx]['provider'] = $types[$query['blocklist']]['optgroup'];
+            }
             $parsed[$idx]['blocklist'] = $types[$query['blocklist']]['value'] ?? $query['blocklist'];
             $parsed[$idx]['policy'] = $policies[$query['uuid']]['description'] ?? '';
             /* Handle front-end color status mapping, start off with OK */
--- a/src/opnsense/mvc/app/controllers/OPNsense/Unbound/Api/OverviewController.php
+++ b/src/opnsense/mvc/app/controllers/OPNsense/Unbound/Api/OverviewController.php
@@ -101,7 +101,7 @@ class OverviewController extends ApiControllerBase
 
         foreach ($parsed as $idx => $query) {
             if (isset($types[$query['blocklist']]['optgroup'])) {
-                $parsed[$idx]['provider'] = $types[$query['blocklist']]['optgroup'];
+                $parsed[$idx]['category'] = $types[$query['blocklist']]['optgroup'];
             }
             $parsed[$idx]['blocklist'] = $types[$query['blocklist']]['value'] ?? $query['blocklist'];
             $parsed[$idx]['policy'] = $policies[$query['uuid']]['description'] ?? '';
--- a/src/opnsense/mvc/app/models/OPNsense/Unbound/Unbound.xml
+++ b/src/opnsense/mvc/app/models/OPNsense/Unbound/Unbound.xml
@@ -451,4 +451,3 @@
         </aliases>
     </items>
 </model>
-
--- a/src/opnsense/mvc/app/models/OPNsense/Unbound/Unbound.xml
+++ b/src/opnsense/mvc/app/models/OPNsense/Unbound/Unbound.xml
@@ -184,36 +184,47 @@
                 <type type="OptionField">
                     <Multiple>Y</Multiple>
                     <OptionValues>
-                        <atf>Abuse.ch - ThreatFox IOC database</atf>
-                        <ag>AdGuard List</ag>
-                        <el>EasyList</el>
-                        <ep>EasyPrivacy</ep>
-                        <hgz001>[hagezi] Multi LIGHT - Basic protection</hgz001>
-                        <hgz002>[hagezi] Multi NORMAL - All-round protection</hgz002>
-                        <hgz003>[hagezi] Multi PRO - Extended protection</hgz003>
-                        <hgz004>[hagezi] Multi PRO mini </hgz004>
-                        <hgz005>[hagezi] Multi PRO++ - Maximum protection</hgz005>
-                        <hgz006>[hagezi] Multi PRO++ mini</hgz006>
-                        <hgz007>[hagezi] Multi ULTIMATE - Aggressive protection</hgz007>
-                        <hgz008>[hagezi] Multi ULTIMATE mini</hgz008>
-                        <hgz009>[hagezi] Fake - scams / fakes</hgz009>
-                        <hgz010>[hagezi] Pop-Up Ads</hgz010>
-                        <hgz011>[hagezi] Threat Intelligence Feeds</hgz011>
-                        <hgz012>[hagezi] Threat Intelligence Feeds - Medium</hgz012>
-                        <hgz013>[hagezi] Threat Intelligence Feeds - Mini</hgz013>
-                        <hgz014>[hagezi] DoH/VPN/TOR/Proxy Bypass</hgz014>
-                        <hgz015>[hagezi] Safesearch not supported</hgz015>
-                        <hgz016>[hagezi] Dynamic DNS blocking</hgz016>
-                        <hgz017>[hagezi] Badware Hoster blocking</hgz017>
-                        <hgz018>[hagezi] Anti Piracy</hgz018>
-                        <hgz019>[hagezi] Gambling</hgz019>
-                        <hgz020>[hagezi] Gambling - Medium</hgz020>
-                        <hgz021>[hagezi] Gambling - Mini</hgz021>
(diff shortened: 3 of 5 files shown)


src/opnsense/scripts/unbound/unbound_dnsbl_options.py
```diff
@@ -0,0 +1,79 @@
...
+            "el": "EasyList",
+            "ep": "EasyPrivacy"
+        },
+        "Hagezi Multi - Clean the Internet": {
+            "hgz001": "LIGHT - Basic protection",
+            "hgz002": "NORMAL - All-round protection",
+            "hgz003": "PRO - Extended protection",
+            "hgz005": "PRO++ - Maximum protection",
+            "hgz007": "ULTIMATE - Aggressive protection",
+        },
```
fichtner: I feel like we're delving into marketing for blocklists here... it's horrible optics

src/opnsense/mvc/app/models/OPNsense/Unbound/Unbound.xml
```diff
@@ -181,40 +181,10 @@
                     <Default>1</Default>
                     <Required>Y</Required>
                 </enabled>
-                <type type="OptionField">
+                <type type="JsonKeyValueStoreField">
                     <Multiple>Y</Multiple>
-                    <OptionValues>
```
fichtner, line 186: you can do categories with option field...

src/opnsense/scripts/unbound/unbound_dnsbl_options.py
sopex: These are the existing names in opnsense... I don't like them. Do you want me to name them as I feel fit?

src/opnsense/mvc/app/models/OPNsense/Unbound/Unbound.xml
sopex: Yes. I know that. https://docs.opnsense.org/development/frontend/models_fieldtypes.html#optionfield

They didn't play well. I will give it another try.

src/opnsense/scripts/unbound/unbound_dnsbl_options.py
fichtner: If @swhite2 agrees we should simplify the existing wording

it’s just my opinion though… these terms do not offer much user guidance and categories can help too so this is a good idea

sopex: We can use the blocking type:

image

Relaxed Blocking
...

src/opnsense/mvc/app/models/OPNsense/Unbound/Unbound.xml
sopex: I am not sure how to make the table and tester get the descriptive names without flattening it, which makes it not worth it. Let me know if I am missing something.

src/opnsense/mvc/app/models/OPNsense/Unbound/Unbound.xml
```diff
@@ -181,40 +181,10 @@
                     <Default>1</Default>
                     <Required>Y</Required>
                 </enabled>
-                <type type="OptionField">
+                <type type="JsonKeyValueStoreField">
```
AdSchellevis: I'm pretty sure we don't need to use an external handler to offer a static list here as the OptionField itself supports sublevels 

https://github.com/opnsense/core/blob/913863a72e5dc0da1fcbffe2cd562bb6e903a454/src/opnsense/mvc/app/models/OPNsense/Base/FieldTypes/OptionField.php#L52

fichtner: As far as I understand this is about having no support for `<opt1 label="Actual category label"/>` at the moment

AdSchellevis: could be the case, haven't looked at it, but if that is the issue it's likely a one line fix...

src/opnsense/scripts/unbound/unbound_dnsbl_options.py
swhite2: We didn't come up with these names, upstream providers did. Why make it harder to find these lists publically?

src/opnsense/mvc/app/models/OPNsense/Unbound/Unbound.xml
swhite2: image

image

Works without issues

src/opnsense/mvc/app/models/OPNsense/Unbound/Unbound.xml
swhite2: @sopex See https://github.com/opnsense/core/pull/10149#discussion_r3071499848

Rendering the descriptive name in the table should be the default already, the fact that it isn't is a different issue. The API response should include `%type`, but it doesn't at the moment.

The `get_policies` endpoint in `overview.volt` does return the descriptive name.

src/opnsense/mvc/app/models/OPNsense/Unbound/Unbound.xml
fichtner: Thanks for the correction. It would be good to add more of these uses to core OptionFields. Apparently this was also fixed in translations not too long ago https://github.com/opnsense/lang/commit/ae1f1a6632e

sopex: @swhite2 My issue is not that table. That works fine. Let me know if I am missing something here:
image
Because this table doesn't like that they are not flat.

swhite2: ok, on master it seems the descriptive names aren't rendered anymore - not sure what changed there

Edit: seems to be https://github.com/opnsense/core/commit/c81417f26747a9e4e46f608c2791bbae805e79fd, unrelated.

sopex: Indeed + I am not sure how dnsbl_match.py‎ can get the OptionFields for the tester output.

swhite2: > Indeed + I am not sure how dnsbl_match.py‎ can get the OptionFields for the tester output.

It can't, we can only try to match in the front-end. You can get the policies from the `get_policies` endpoint, (i.e. https://github.com/opnsense/core/blob/master/src/opnsense/mvc/app/views/OPNsense/Unbound/overview.volt#L383)

sopex: Yeah, so this implementation is probably just fine? 
I get that this could be a static list, but we would have to jump through too many hoops for no reason.
A little Python script never hurt anyone :)

swhite2: > Yeah, so this implementation is probably just fine? I get that this could be a static list, but we would have to jump through too many hoops for no reason. A little Python script never hurt anyone :)

No, as discussed there's no need for a `JsonKeyValueStoreField`, there's also no need for descriptive names on the backend, so fetching the policies once in the front-end and mapping it there is just as much effort and likely just as many lines of code.

sopex: I apologize, but I got a bit confused. 
Just to make sure I understand: 
You suggest to reverse https://github.com/opnsense/core/commit/c81417f26747a9e4e46f608c2791bbae805e79fd, so the main table works again, and then you want me to hijack the output of ajaxCall('/api/unbound/diagnostics/testBlocklist' in the frontend to replace the "bl" with the descriptive name? 
The lines of code will be more than just using the script, but I get that that is not necessarily important.
Thanks for bearing with me. Just a bit confused.

swhite2: > I apologize, but I got a bit confused. Just to make sure I understand: You suggest to reverse [c81417f](https://github.com/opnsense/core/commit/c81417f26747a9e4e46f608c2791bbae805e79fd),

No sorry, that particular issue is being worked on - ignore that, it's only cosmetic.


> and then you want me to hijack the output of ajaxCall('/api/unbound/diagnostics/testBlocklist' in the frontend to replace the "bl" with the descriptive name?

Yes, this is a possiblity - but come to think of it you may also opt to modify `testBlocklistAction()` in the DiagnosticsController to match it there. You would only need `$this->mdl->dnsbl->blocklist->getNodeContent();` like `getPoliciesAction()` does. This should prevent more front-end code.

Thanks for bearing with me as well ;)

sopex:
Thanks for this https://github.com/opnsense/core/commit/ae326511d16ff6f35fee7f7dd22bf1d761b48f96 
Works great!

AdSchellevis:
@sopex thanks for confirming, missed that part yesterday it seems.

src/opnsense/mvc/app/controllers/OPNsense/Unbound/Api/DiagnosticsController.php
```diff
@@ -123,6 +123,19 @@ public function testBlocklistAction()
             ]), true);
 
             if (!empty($response)) {
+                $mdl = new \OPNsense\Unbound\Unbound();
+                $policies = $mdl->dnsbl->blocklist->getNodeContent();
+                $types = $mdl->dnsbl->blocklist->getTemplateNode()->type->getNodeData();
+                foreach ($response as $key => $value) {
+                    if (is_array($value)) {
```
swhite2: `isset($value['bl'])` already implicitly ensures `$value` is an array, so

```
if (isset($value['bl'])) {
    $response[$key]['bl'] = $types[$value['bl']]['value'] ?? $value['bl'];
}

```

should be enough

swhite2 (changes requested):

src/opnsense/mvc/app/models/OPNsense/Unbound/Unbound.xml
```diff
@@ -184,36 +184,47 @@
...
+                        <opt1 value="General Blocklists">
+                            <atf>Abuse.ch - ThreatFox IOC database</atf>
+                            <ag>AdGuard List</ag>
+                            <el>EasyList</el>
+                            <ep>EasyPrivacy</ep>
+                            <sb>Steven Black List</sb>
+                            <yy>YoYo List</yy>
+                        </opt1>
+                        <opt2 value="Hagezi Multi - Clean the Internet">
+                            <hgz001>LIGHT - Relaxed Blocking</hgz001>
```
fichtner, line 196: one thing that is annoying here for later lookup is that the vendor is no longer attached to the entry but moved to the category. the categories absolutely make sense but it would be nicer per use case, not vendor specific moving into the pickle of not having the vendor name in the label. Then it still has use cases in the string but in part that's due to vendor funkiness.

src/opnsense/mvc/app/models/OPNsense/Unbound/Unbound.xml
```diff
@@ -184,36 +184,47 @@
...
+                            <hgz010>Pop-Up Ads</hgz010>
+                            <hgz014>DoH/VPN/TOR/Proxy Bypass</hgz014>
+                            <hgz015>Safesearch not supported</hgz015>
+                            <hgz016>Dynamic DNS blocking</hgz016>
+                            <hgz017>Badware Hoster blocking</hgz017>
+                            <hgz018>Anti Piracy</hgz018>
+                            <hgz019>Gambling</hgz019>
+                            <hgz022>Social Networks</hgz022>
+                        </opt4>
+                        <opt5 value="Smaller versions of blocklists for underpowered systems">
```
fichtner: this is a bit too long as it tries to explain instead of using the documentation or a clearer label. Or simply use "minimal blocklists". comparatives are always problematic when trying to explain something... compared to what exactly? These are things that wouldn't prevail in a technical documentation.

sopex: Yes, this needs to be fixed. It was more of a personal note. These probably don't need to exist in opnsense either way.

They are designed mostly for phones, plus the way opnsense handles blocklists nowadays is very efficient.

Its compared to the ones above it but yes

src/opnsense/mvc/app/models/OPNsense/Unbound/Unbound.xml
sopex: I thought about this very early on, but  I don't think this is a problem after all.
You get the basic info from looking at it and then you can click edit for more info.

image

The per-category blocking will make very little sense...

src/opnsense/mvc/app/models/OPNsense/Unbound/Unbound.xml
sopex: Would "Underpowered systems" be a bad idea? I want something that discourages their usage.

src/opnsense/mvc/app/models/OPNsense/Unbound/Unbound.xml
fichtner: To me it looks like the grid was clearer before as now it's missing more context but the names are currently too long / inconsistent in length. the tradeoff in the PR is not well balanced at a quick glance. But it's rather difficult in this inhomogeneous set of data.

My main concern is that categories are a very good tool here but making vendor categories and general categories and use-case categories at the same time will not get you a clearer picture/overview of the blocklists which I think you goal was. The categories should categorize one thing and the use-case seems to be the most reasonable as it is also self-explanatory. And when you only have use-case categories then you need the vendor in the actual listing again.

In contrast it would also be ok to update the documentation which can hold more data and info and add several layers of categorization away from complicating the code/data representation in the model.

src/opnsense/mvc/app/models/OPNsense/Unbound/Unbound.xml
fichtner: As a concept yes also a bit imprecise. AI suggested "Minimal hardware requirements" amongst other things which is pretty spot on.

fichtner: For the record:

> Here are some alternative category labels with slightly different tones and clarity:
> 
> Neutral / Technical
> 
> Low-Resource Systems
> Resource-Constrained Systems
> Minimal Hardware Requirements
> Lightweight Performance
> Reduced System Load
> 
> User-Friendly
> 
> Runs on Older Devices
> Works on Basic Hardware
> Optimized for Simpler PCs
> For Entry-Level Systems
> Smooth on Low-End Devices
> 
> Marketing / Positive Spin
> 
> Performance-Friendly
> Efficiency Optimized
> Lightweight & Fast
> Built for Efficiency
> Streamlined Performance
>
> More Direct / Honest
>
> For Low-End Systems
> Limited Hardware Support
> Basic Performance Tier
> Reduced Performance Mode

sopex: Most of these would be equally good.
I like this "Resource Constrained Systems"
but "Minimal Hardware Requirements" is equally great

fichtner: One of the few use case where AI shines having been trained on marketing and technical docs alike ;)

src/opnsense/mvc/app/models/OPNsense/Unbound/Unbound.xml
sopex: There is zero context missing. On the contrary, IMO the context is front and center now. 

image&nbsp; &nbsp; &nbsp;&nbsp; &nbsp; &nbsp;&nbsp; &nbsp; &nbsp;image

The problem obviously stems from the blocklists themselves.

But I completely disagree with changing the grouping to usecase in this instance. 
A) Because the usecases with more than 1 option (Minimal Hardware Requirements excluded) are two:

- General Usage (Includes all except the targeted hegazi and 2 OIDC)

- Targeted Usage (Includes targeted hegazi and 2 OIDC)

B) To make this work, we present to the user "competitive" options for the same thing, which is not helpful...

Why would someone choose the hegazi gambling vs the XX gambling blocklist?
The way it is in this OR is like a shopping list that you pass through and get what you need

In the documentation, we can add anything you want, but the UI should stand on its own... Because people refuse to read the documentation even after searching for the solution for 3 hours 😞 https://github.com/opnsense/core/issues/10151 (Completely random point, it could just as easily have been me)

fichtner: I guess I’m missing the point here then. It’s difficult to be productive in these non-technical PRs and I only meant to provide input into what is good and bad about the current and the proposed way. This is a tricky problem domain with multi vendor presentation and always has been.

sopex: Your feedback is always valuable, technical or not. 

Let's leave it open to see if any good ideas for usecase grouping show up. 

I currently don't have any, but no reason to move forward here if I am the only one happy with the result. I am definitely not infallible 😂

sopex: @swhite2 Any UI ideas here?

fichtner (approved):
I'm missing the blocklist provider in the labels now as not everything evolves around the dropdowns, but I also don't want to block this.

swhite2:
@sopex I think this is the right approach but @fichtner's concern is valid, we're losing the provider context now.

The way I see it there are 3 places these labels can show up:

- In the grid cell. We don't need provider context here as the dropdown is available for lookup, though we can just format it in there if we want to.
- The blocklist tester.
- The DNS overview.

The last 2 can be solved with `$types = $mdl->dnsbl->blocklist->getTemplateNode()->type->getNodeData();`, which provides the `optgroup` next to `value` which stores the "provider" separately here. Either we add this as a separate field in the tester output or in the case of the DNS overview page format it nicely somewhere.

image

This approach also isn't tied to "provider" specifically, we can categorize it any way we see fit at a later point in time, it's not tied to any configuration property nor does code need to be overhauled.

sopex:
@swhite2 Thank you for your help, insightful as always.

swhite2:
@sopex Happy to help, one last thing: https://github.com/opnsense/core/blob/master/src/opnsense/mvc/app/views/OPNsense/Unbound/overview.volt#L633 also shows the blocklist

swhite2 (approved):
Ok minus the comment above

src/opnsense/mvc/app/controllers/OPNsense/Unbound/Api/OverviewController.php
```diff
@@ -100,6 +100,9 @@ public function searchQueriesAction()
         $types = $this->mdl->dnsbl->blocklist->getTemplateNode()->type->getNodeData();
 
         foreach ($parsed as $idx => $query) {
+            if (isset($types[$query['blocklist']]['optgroup'])) {
+                $parsed[$idx]['provider'] = $types[$query['blocklist']]['optgroup'];
```
swhite2: Oh, just noticed this, maybe we should just call this `category`

sopex:
> @sopex Happy to help, one last thing: https://github.com/opnsense/core/blob/master/src/opnsense/mvc/app/views/OPNsense/Unbound/overview.volt#L633 also shows the blocklist

Great! I think it's ok now.

src/opnsense/mvc/app/controllers/OPNsense/Unbound/Api/OverviewController.php
swhite2:
```suggestion
                $parsed[$idx]['category'] = $types[$query['blocklist']]['optgroup'];
```

src/opnsense/mvc/app/views/OPNsense/Unbound/overview.volt
```diff
@@ -627,14 +627,6 @@
                     $tbody.append($tr);
                 }
 
-                if (blocklist != "") {
-                    $container = $(`
-                        <div>
-                            {{ lang._('Blocklist match:')}} ${blocklist}
```
swhite2, line 633: Why remove this? This shows the matched blocklist in the policy quick edit dialog

src/opnsense/mvc/app/views/OPNsense/Unbound/overview.volt
```diff
@@ -725,6 +717,12 @@
                             },
                             "domain": function (column, row) {
                                 return row.domain;
+                            },
+                            "blocklist": function (column, row) {
+                                if (row.provider) {
+                                    return row.blocklist + ' - ' + row.provider;
+                                }
+                                return row.blocklist;
```
swhite2:
```suggestion
                                if (row.category) {
                                    return row.blocklist + ' - ' + row.category;
                                }
                                return row.blocklist;
```

swhite2 (changes requested):

src/opnsense/mvc/app/views/OPNsense/Unbound/overview.volt
sopex: Give me some time on this

swhite2: of course, just let me know if you think it's ready

sopex: @swhite2

swhite2 (approved):

swhite2:
@sopex Thanks!

PR #10156, opened 2026-04-15
Add cmd timeout, error detection and logging to wireguard re-resolver script

deajan:
Refactor DNS re-resolution script to improve logging and command execution, make it generally more reliable, add proper exit code, and make it portable for use in other projects (tested in OPNSense 26.1 and RHEL9 so far)

**Important notices**

- [X] I opened an issue first for non-trivial changes and linked it below (not applicable)

**Describe the problem**

I'm currently investigating issues with wireguard where after a WAN drop, wireguard never reconnects (handshake becomes stale) unless I restart the service on local or remote OPNsense side.

This PR adds stderr capture to the commands, adds a timeout in case a subprocess command gets stuck, and adds generic  (rotated) logging for optional debugging purposes, catches all possible errors and logs them.
It is a first step into diagonsis of what actually happens with wireguard not reconnecting.

@AdSchellevis Please let me know if I'm out of line with this.
I would also like to add an optional "restart service" action when handshake is stale and updating resolved FQDN don't fix the issue, but that's too broad for a diagnostic right now.

### DIFF
--- a/src/opnsense/scripts/wireguard/reresolve-dns.py
+++ b/src/opnsense/scripts/wireguard/reresolve-dns.py
@@ -1,7 +1,7 @@
-#!/usr/local/bin/python3
+#!/usr/bin/env python3
 
 """
-    Copyright (c) 2023 Ad Schellevis <ad@opnsense.org>
+    Copyright (c) 2023-2026 Ad Schellevis <ad@opnsense.org>
     All rights reserved.
 
     Redistribution and use in source and binary forms, with or without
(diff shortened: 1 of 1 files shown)


AdSchellevis:
@deajan just try to keep it simple, all processes can emit to syslog, if messages are missing, I have no objections to add them, but do have a desire to keep the script as simple as possible. Starting with a ticket describing the problem that is aimed to be solved is always a good starting point.

deajan:
@AdSchellevis I see your point, but since I'm trying to resolve the same issues on OPNsense that I encounter on my AlmaLinux setups when using Wireguard, I wanted to make a "can use it everywhere" solution based on your work. The result is working on both platforms currently.

Do you suggest me to remove the logfile support so logs are catched from stdout by the caller (eg cron) ?
I'm willing to adapt what to what you suggest, but I honestly don't understand it right now.
Can you give me some insight ?

Apart from the logging, I wanted to make the script "error proof(TM)", meaning that I want at least to catch any possible error and log it, and of course have a one minute timemout enforced so execution cannot be frozen by some unresponsive shell command.
Is that idea set okay for you ?

AdSchellevis:
@deajan I'm ok with catching and logging errors, certainly, just want to make sure changes are small and focused

deajan:
@AdSchellevis Sure, so what do you want me to remove / rework ? i can make the logging part optional, but it's convenient to log stuff and use the same script when running on different platforms.
Sorry if I'm bugging you but I genuinly don't understand what you want me to modify.

AdSchellevis:
I'm just asking for a minimal change to my script using syslog as output, looking at the request, it feels that this is about 5 to 10 lines of code, which is currently not the case. So, keep it simple, starting with the goals we try to achieve.

deajan:
@AdSchellevis Sorry for the delayed response.

I've updated the script to use syslog as a default log destination, while keeping the possibility to log to a file.
Let me know if you want me to change anything else.

Since I'm running said script for a month or so, I also noticed that on reboots sometimes the `wg set <ifname> peer <pubkey> endpoint <fqdn>` may fail because wireguard service couldn't start in the first place, because dns resolution wasn't working at the time the system started and wireguard service was started.
I've added a `--failure-command` optional argument that allows to restart the wireguard service in such cases.

So far, this made my reboots smooth on both a couple of OPNSenses and RHEL machines using wireguard.

AdSchellevis:
@deajan this doesn't really feel like the 5-10 lines fix requested to be honest, the change seems to have only grown over time which makes it even harder to review. For dns issues, there's the "Renew DNS for WireGuard on stale connections" cron job which can be scheduled to update the peers... (see also https://docs.opnsense.org/manual/how-tos/wireguard-s2s.html)

deajan:
@AdSchellevis I tried to keep the same script for both OPNSense and other systems, hence a new feature on top of the requested modification.

The point here is that wireguard service won't even start on some systems if DNS cannot be resolved, hence trying to re-resolve them won't work since service isn't started. This obviously happens on non OPNsense systems.

Anyway, Do you want me to remove the `--failure-command` option for this PR ?
I don't like it since it will fork my script into a opnsense and a non opnsense version, but I totally get your point.

fichtner:
I don’t fully understand the premise here. If DNS doesn’t work how does restarting WireGuard fix anything? The only thing that fixes this is retrying later once DNS is up (or not using host names). 

Unless something tunnels DNS over WireGuard which is less than ideal.

Cheers,
Franco

AdSchellevis:
> @AdSchellevis I tried to keep the same script for both OPNSense and other systems, hence a new feature on top of the requested modification.
> 
> The point here is that wireguard service won't even start on some systems if DNS cannot be resolved, hence trying to re-resolve them won't work since service isn't started. This obviously happens on non OPNsense systems.
> 
> Anyway, Do you want me to remove the `--failure-command` option for this PR ? I don't like it since it will fork my script into a opnsense and a non opnsense version, but I totally get your point.

If the PR is about missing logging, that's the only thing I expect in this PR, plain and simple. For the rest I concur with what @fichtner mentioned here.

deajan:
@fichtner This script applies to OPNSense for the DNS resolution part, and to other OSes when it comes to restart wiregaurd service. The point here is that wireguard service won't start if dns resolution isn't working on some OSes.

Anyway, I get it. I guess I'll close this PR in favor for one that does only add logging, and will keep a more complete version of the script that is platform agnostic and resolves the issue above.

deajan:
New pull request open at https://github.com/opnsense/core/pull/10319

PR #10158, opened 2026-04-15
Auth/LDAP: restore timing normalization on failed logins

TristanInSec:
The LDAP class overrides Base::authenticate() and calls _authenticate() directly, bypassing the constant-time timing normalization that Base provides on failed logins. This allows distinguishing "user not found" (fast) from "wrong password" (slow LDAP bind round-trip) via response timing, enabling remote user enumeration through the login form, Captive Portal, or OpenVPN authentication endpoints.

This is related to GHSA-f683-wg4m-rhf2.

Remove the override so LDAP/LDAPTOTP inherit Base's 2-second timing floor on failures, consistent with Local and LocalTOTP.

### DIFF
--- a/src/opnsense/mvc/app/library/OPNsense/Auth/LDAP.php
+++ b/src/opnsense/mvc/app/library/OPNsense/Auth/LDAP.php
@@ -479,17 +479,6 @@ class LDAP extends Base implements IAuthConnector
     }
 
     /**
-     * authenticate user against ldap server without Base's timer
-     * @param string $username username to authenticate
-     * @param string $password user password
-     * @return bool authentication status
-     */
-    public function authenticate($username, $password)
-    {
-        return $this->_authenticate($username, $password);
-    }
-
-    /**
      * authenticate user against ldap server, implementation as described in Base class
      * @param string $username username to authenticate
      * @param string $password user password


AdSchellevis:
@TristanInSec in theory that should work just fine, but with all the AI reports these days that seem to overlap, just a small control question, did you try this?

fichtner:
It would be really really nice if whatever tools are being used would bother checking for existing patches and PRs... https://github.com/opnsense/core/pull/10111

TristanInSec:
Duplicate of #10111, apologies for the noise.

TristanInSec:
> It would be really really nice if whatever tools are being used would bother checking for existing patches and PRs... #10111

You're right, I missed https://github.com/opnsense/core/pull/10111, apologies for the duplicate PR.

That said, https://github.com/opnsense/core/pull/10111 has been open since April 9 and the timing issue is not covered in CVE-2026-34578 or the 26.1.6 patch. The authenticate() override is still in production, so user enumeration via response timing is still possible without any LDAP injection.

Could the timing component be added to the existing advisory, or should it be tracked separately?

fichtner:
https://github.com/opnsense/core/security/advisories/GHSA-jpm7-f59c-mp54 point 2 covers timing and the follow up patch exists via https://github.com/opnsense/core/pull/10111

There's little testing for the PR for now and it is still there for risk of a discussed regression we've seen before. Just very little time at the moment in between minor and major releases and Easter.


Cheers,
Franco

AdSchellevis:
The story about timing attacks is a bit different, the fact that we haven't seen the user yet isn't the same as it doesn't exist. For local authentication (the original reason for the timing wrapper), that's different, for external authenticators there's always a lag in between.

TristanInSec:
Thanks for the clarification. Good to know point 2 already covers the timing component. Understood on the regression risk with #10111, no rush.

If you don't mind, as I discovered the same 2 findings independently, happy to be added to the report from @mandreko for credit.

I will keep you posted if I find new issues, and will check pending commits before submitting.

Cheers.

Issue #10152, opened 2026-04-13
Captive Portal RADIUS accounting crashes when IPv6 client address is sent as Framed-IP-Address


Lukas290507:
**Important notices**

Before you add a new report, we ask you kindly to acknowledge the following:

- [x] I have read the contributing guidelines at https://github.com/opnsense/core/blob/master/CONTRIBUTING.md
- [x] I am convinced that my issue is new after having checked both open and closed issues at https://github.com/opnsense/core/issues?q=is%3Aissue

**Describe the bug**
There appears to be a captive portal RADIUS accounting bug in 26.1.6 related to IPv6 client addresses.

`src/opnsense/scripts/captiveportal/process_accounting_messages.php` passes client ip addresses directly into `src/opnsense/mvc/app/library/OPNsense/Auth/Radius.php` ` stopAccounting()` / `updateAccounting()`.

```php
radius_put_addr($radius, RADIUS_FRAMED_IP_ADDRESS, $ip_address)
```

However, client ip addresses can contain IPv6 addresses, and `radius_put_addr()`  is IPv4-only. This causes repeated PHP fatals:

`PHP Fatal error: Error converting Address in /usr/local/opnsense/mvc/app/library/OPNsense/Auth/Radius.php`

**To Reproduce**

1. Captiveportal with Radius Server where accounting is enabled
2. Clients with ipv6 addresses
3. Error will appear 

**Expected behavior**

Captive Portal RADIUS accounting should handle sessions with IPv6 client addresses without crashing.

If the client address is IPv4, it can continue using `RADIUS_FRAMED_IP_ADDRESS`.
If the client address is IPv6, it should not try to encode it as `RADIUS_FRAMED_IP_ADDRESS`, its IPv4-only. It should either skip ipv6 addresses in accounting or use the appropriate IPv6 RADIUS attributes instead.

**Describe alternatives you considered**

""

**Screenshots**

<img width="1510" height="353" alt="Image" src="https://github.com/user-attachments/assets/49b2dbc5-acec-4d39-ad8d-336450cbf973" />

**Relevant log files**
Debuging Logs from testing:
``` Radius debug: username=xxxxxxxxx sessionid=xxxxxxxxxxxx== ip_address='192.168.199.247' 
    Radius debug: username=xxxxxxxxx sessionid=xxxxxxxxxxxxx== ip_address='fe80::xxxx:xxxx:xxxx:xxxx'
```

```
[...]

PR #10171, opened 2026-04-18
Avoid sending Framed-IP-Address for IPv6 accounting clients

Snowcap8141:
This patch avoids sending RADIUS Framed-IP-Address when the client address is IPv6.

A targeted unit test was added for the IPv4/IPv6 guard logic.

I was not able to validate this end-to-end in a full Captive Portal + RADIUS + IPv6 setup yet, so this should not be considered fully integration-tested.

After applying the patch on a test firewall, I did not observe visible regressions:
- patched file confirmed in place
- Web UI remained functional after reboot
- no immediate obvious runtime regressions were observed

**Important notices**

- Model used: OpenAI Codex / GPT-5-based assistant
- Extent of AI involvement: AI was used to help analyze the issue, draft the guard logic, draft the unit test, and help prepare this pull request text. Final review and submission were done manually.

---

**Describe the problem**

`Framed-IP-Address` appears to be sent in captive portal RADIUS accounting even when the client address is IPv6.

Since this path expects an IPv4 address, accounting may fail for IPv6 clients.

---

this pull request keeps the existing IPv4 behavior for `Framed-IP-Address`, but handles IPv6 client addresses through an IPv6 RADIUS attribute path instead of the IPv4-only handler.

The change is applied in both:
- `stopAccounting()`
- `updateAccounting()`

For IPv4 clients, the code still uses the existing `Framed-IP-Address` path.
For IPv6 clients, the address is encoded as an IPv6 RADIUS attribute payload instead of being omitted or passed through the IPv4-only `radius_put_addr()` path.

A targeted unit test was added and extended to verify:
- IPv4 still uses the IPv4 address handler
- IPv6 uses the IPv6 attribute path
- the generated IPv6 payload is encoded in the expected format

I was not able to validate this operationally in a full Captive Portal + RADIUS + IPv6 setup yet, so this should not be considered fully integration-tested.

After applying the patch on a test firewall, I did not observe visible regressions:
- patched file confirmed in place
- Web UI remained functional after reboot
- no immediate obvious runtime regressions were observed
---

**Related issue**

Ref: #10152

### DIFF
--- a/src/opnsense/mvc/app/library/OPNsense/Auth/Radius.php
+++ b/src/opnsense/mvc/app/library/OPNsense/Auth/Radius.php
@@ -140,6 +140,11 @@ class Radius extends Base implements IAuthConnector
         return [$wraps, $lower32];
     }
 
+    private function shouldSendFramedIpAddress($ipAddress)
+    {
+        return filter_var(trim((string)$ipAddress), FILTER_VALIDATE_IP, FILTER_FLAG_IPV4) !== false;
+    }
+
     /**
      * type name in configuration
      * @return string
@@ -370,7 +375,7 @@ class Radius extends Base implements IAuthConnector
                 $error = radius_strerror($radius);
             } elseif (!radius_put_int($radius, 53, $wraps_out)) { /* Acct-Output-Gigawords */
                 $error = radius_strerror($radius);
-            } elseif (!radius_put_addr($radius, RADIUS_FRAMED_IP_ADDRESS, $ip_address)) {
+            } elseif ($this->shouldSendFramedIpAddress($ip_address) && !radius_put_addr($radius, RADIUS_FRAMED_IP_ADDRESS, $ip_address)) {
                 $error = radius_strerror($radius);
             } elseif (!radius_put_int($radius, RADIUS_ACCT_TERMINATE_CAUSE, $this->mapTerminateCause($cause))) {
                 $error = radius_strerror($radius);
@@ -458,7 +463,7 @@ class Radius extends Base implements IAuthConnector
                 $error = radius_strerror($radius);
             } elseif (!radius_put_int($radius, 53, $wraps_out)) { /* Acct-Output-Gigawords */
                 $error = radius_strerror($radius);
-            } elseif (!radius_put_addr($radius, RADIUS_FRAMED_IP_ADDRESS, $ip_address)) {
+            } elseif ($this->shouldSendFramedIpAddress($ip_address) && !radius_put_addr($radius, RADIUS_FRAMED_IP_ADDRESS, $ip_address)) {
                 $error = radius_strerror($radius);
             }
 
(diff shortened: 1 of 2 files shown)


AdSchellevis:
I don't believe a proper fix here is simply not sending the address anymore, it should be offered with the proper type logically. Which either requires a patch on the radius library or a workaround using other handles available in the lib.

Snowcap8141:
Thanks, that makes sense.

My intention here was to avoid the failing IPv4-only path for IPv6 clients with a minimal guard, but I understand the concern that simply omitting the address is not the right long-term fix.

I will take a closer look at whether this should instead use the correct attribute/type for IPv6, or whether the radius library exposes a better handle for this case.

Snowcap8141:
> I don't believe a proper fix here is simply not sending the address anymore, it should be offered with the proper type logically. Which either requires a patch on the radius library or a workaround using other handles available in the lib.
hi


I updated the patch in the direction you suggested. The new version no longer omits the IPv6 client address, but uses an IPv6 RADIUS attribute path instead of the IPv4-only handler.

I also updated the unit test coverage accordingly.

I was not able to validate this end-to-end in a full Captive Portal + RADIUS + IPv6 setup, so I would appreciate it if you could verify whether this matches the expected behavior on your side.

swhite2:
@Snowcap8141 https://github.com/opnsense/core/commit/4da7b63381e1f5255603ee3b945ddd9ee07bb8b0 should fix the issue as well, limiting it to `Framed-IPv6-Address` to keep it simple.

I'm closing this PR, but if further discussion is required we can always do so in the original ticket.

Snowcap8141:
> @Snowcap8141 [4da7b63](https://github.com/opnsense/core/commit/4da7b63381e1f5255603ee3b945ddd9ee07bb8b0) should fix the issue as well, limiting it to `Framed-IPv6-Address` to keep it simple.
> 
> I'm closing this PR, but if further discussion is required we can always do so in the original ticket.


Hi, @swhite  thanks for the update and for implementing the IPv6 handling.

I noticed that the final solution uses Framed-IPv6-Address (attr 168), which aligns with the original issue I was trying to address in my PR.

I appreciate the simpler and cleaner approach you applied here.

Just one question for clarity: since this change is related to the initial report/PR, would it be possible to reference it for traceability?

No problem either way, I just wanted to better understand how contributions like this are usually handled.

Thanks again for your work.

swhite2:
@Snowcap8141 Both the commit and this PR reference the original ticket which should aid traceability.

Thanks for your initial proposal!

Snowcap8141:
> @Snowcap8141 Both the commit and this PR reference the original ticket which should aid traceability.
> 
> 
> 
> Thanks for your initial proposal!

@swhite2 Thanks for clarifying, appreciate it!

Glad the issue is resolved and aligned with the original direction 👍

PR #10178, opened 2026-04-20
system: distill service widget essence based on #9608

fichtner:
**Important notices**

- [x] I opened an issue first for non-trivial changes and linked it below.  #7887

---

**Describe the problem**

See ticket.

---

Less wasted space, move process status to background color.

Screenshot 2026-04-20 at 10 10 09

---

**Related issue**

#7887

### DIFF
--- a/src/opnsense/www/js/widgets/Services.js
+++ b/src/opnsense/www/js/widgets/Services.js
@@ -38,17 +38,16 @@ export default class Services extends BaseTableWidget {
     }
 
     getMarkup() {
-        return $(`<div id="services-container" style="display: grid; grid-template-columns: repeat(auto-fit, minmax(140px, 1fr)); padding: 5px; gap: 5px;"></div>`);
+        return $(`<div id="services-container" style="display: grid; grid-template-columns: repeat(auto-fit, minmax(140px, 1fr)); padding: 1px; gap: 1px;"></div>`);
     }
 
     serviceControl(actions) {
         return actions.map(({ action, id, title, icon }) => `
-            <button data-service_action="${action}" data-service="${id}"
-                  class="btn btn-xs btn-default srv_status_act2"
-                  style="font-size: 10px; padding: 2px 6px;"
+            <span data-service_action="${action}" data-service="${id}"
+                  class="srv_status_act2"
                   title="${title}" data-toggle="tooltip">
                 <i class="fa fa-fw fa-${icon}"></i>
-            </button>
+            </span>
         `).join('');
     }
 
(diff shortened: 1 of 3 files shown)


src/opnsense/www/js/widgets/Services.js
```diff
@@ -38,17 +38,16 @@ export default class Services extends BaseTableWidget {
...
-        return $(`<div id="services-container" style="display: grid; grid-template-columns: repeat(auto-fit, minmax(140px, 1fr)); padding: 5px; gap: 5px;"></div>`);
+        return $(`<div id="services-container" style="display: grid; grid-template-columns: repeat(auto-fit, minmax(140px, 1fr)); padding: 1px; gap: 1px;"></div>`);
     }
 
     serviceControl(actions) {
         return actions.map(({ action, id, title, icon }) => `
-            <button data-service_action="${action}" data-service="${id}"
-                  class="btn btn-xs btn-default srv_status_act2"
-                  style="font-size: 10px; padding: 2px 6px;"
+            <span data-service_action="${action}" data-service="${id}"
```
swhite2, line 46: one could argue `cursor: pointer` here

src/opnsense/www/js/widgets/Services.js
```diff
@@ -38,17 +38,16 @@ export default class Services extends BaseTableWidget {
...
+                  class="srv_status_act2"
                   title="${title}" data-toggle="tooltip">
                 <i class="fa fa-fw fa-${icon}"></i>
```
swhite2, line 49: The icons seem a bit transparent / lower in opacity / muted to me, or is this an optical illusion?

src/opnsense/www/js/widgets/Services.js
fichtner: sure why not

src/opnsense/www/js/widgets/Services.js
fichtner: after fighting button styles this was a happy little accident removing the button, but we can improve visibility further I think

swhite2: I for one will not stand in the way of Bob Rossin' the UI.

src/opnsense/www/js/widgets/Services.js
```diff
@@ -38,17 +38,16 @@ export default class Services extends BaseTableWidget {
     }
 
     getMarkup() {
-        return $(`<div id="services-container" style="display: grid; grid-template-columns: repeat(auto-fit, minmax(140px, 1fr)); padding: 5px; gap: 5px;"></div>`);
+        return $(`<div id="services-container" style="display: grid; grid-template-columns: repeat(auto-fit, minmax(140px, 1fr)); padding: 1px; gap: 1px;"></div>`);
```
sopex, line 41: My review is solely visual.
I think it is a bit foreign to the opnsense design language and also makes it more difficult to read the services due to being crammed together.
But my eyes don't like this shade of green at larger dosages, so this might be part of the problem :)

Screenshot 2026-04-20 113943

src/opnsense/www/js/widgets/Services.js
sopex: A happy tree is all that is missing.

src/opnsense/www/js/widgets/Services.js
fichtner: but if one is stopped it sticks out really well.. I still have to see if another color makes more sense and pinning it for all themes instead of trying to softcode the bootgrid defaults.. I theory the background color should be lighter but I don't think we've followed it as closely with out theming, see https://getbootstrap.com/docs/4.0/components/alerts/

as far as scaling and grid goes these are the previous defaults but I've made the font bigger again. I don't like it being smashed as much horizontally but it simply needs more tweaking.

sopex: The original has more vertical space and horizontally is night and day.

I believe the correct green would actually be darker and not lighter.
OPNsense orange is dark, and the interfaces/ traffic/ firewall graphs use very dark/bold colors, this and lighter greens feel washed up here.

Also, I still believe this feels more opnsense and not because I did it :)

image

5 services become 9

Monviech:
Just a different idea, how about sorting services by state (stopped, then started) first, and afterwards alphabetically. I do it in a few widgets to quickly show connected clients first (e.g. wireguard, ipsec, openvpn, etc...) or the certificate widget (expired show first).

That way saving space isn't too important, stopped services will always migrate to the first spots in the view.

swhite2:
> Just a different idea, how about sorting services by state (stopped, then started) first, and afterwards alphabetically.

I think in this instance predictability wins as users may be used to a particular ordering. Having elements jump around serves no particular purpose IMO.

Monviech:
> > Just a different idea, how about sorting services by state (stopped, then started) first, and afterwards alphabetically.
> 
> I think in this instance predictability wins as users may be used to a particular ordering. Having elements jump around serves no particular purpose IMO.

That makes sense, its just a personal preference of mine to see the most important things first in the list. Others might have different preferences. It's why GUI related things are always hard to distill to one size fits all.

I guess what I want could be destilled into a new sort of widget that is a meta status of all things that are currently not okay and displays them in a list or something. "Show me everything that is currently degraded" -> though scope creep would immediately kill such a "god" widget.

src/opnsense/www/js/widgets/Services.js
Monviech: I'd generally not lean too much into red and green color coding, and more into distinct icons, as there are a lot of visually impairments regarding red green or similiar shapes.

As reference I had to clean up this:
https://github.com/opnsense/core/pull/9704

And affected users will always report quickly.

ronin3510:
I’m all for not falling in the Material Design trap, one of the poorest design choices implemented at scale in what can only be described as anti-consumer move (almost white text over a white background model is what I’m referring to, with all the possible variations).

It already looks like the green background status idea with slightly darker green controls has been tossed but I felt necessary to comment about the similarity with Material Design.


One last thing I want to add is that for consistency I’d would evaluate color changes not only on how good they look on the white theme but on the dark one as well

PR #10184, opened 2026-04-21
Firewall: DNAT: "disabled" and "ipprotocol" migrations

### LINKED ISSUE #9859: firewall: dnat model "disabled", "ipprotocol" and "protocol" migrations
**Important notices**

For mid-term consideration:

- [ ] flip `disabled` field to `enabled`
- [ ] `ipprotocol` has "any" and "inet46" options.. removing "inet46" may be beneficial long term as the construct is weird (creating two rules or one without the setting depending on how it is handled underneath)
- [ ] `protocol` is optional and if `ipprotocol` stays optional remove the XXX from the field definition

TomWalraven:
**Important notices**

- Model used: GPT-5.3-Codex (Copilot)
- Extent of AI involvement: Copilot was used to assist with debugging the actionPostLoadingEvent method in DNatEnabledField. 
---

**Describe the problem**

The DNat page has a disabled flag while the other nat rule pages have enabled flags. This should be changed to enabled for better consistency. IPProtocol field has inet46 option which is redundent because the "any" option already exists.

---

Inline migrations for flipping the disabled flag to enabled and removing the inet46 option. Enabled is added as a volatile field and gets mapped from existing disabled values and rules that were created with inet46 option set get set to "any" now.

---

**Related issue**

Closes: #9859

### DIFF
--- a/src/opnsense/mvc/app/models/OPNsense/Firewall/FieldTypes/DNatEnabledField.php
+++ b/src/opnsense/mvc/app/models/OPNsense/Firewall/FieldTypes/DNatEnabledField.php
@@ -46,16 +46,17 @@ class DNatEnabledField extends BooleanField
         return parent::setValue($value);
     }
 
-    protected function actionPostLoadingEvent()
+    public function getValue(): string
     {
-        // Only use disabled flag if enabled does not have a value set.
-        if ($this->internalInitialValue === false || $this->getValue() === '') {
+        $value = parent::getValue();
+        $isNotInitialized = $this->internalInitialValue === false;
+        // Get value from disabled flag if enabled does not have a value set.
+        if ($isNotInitialized || $value === '') {
             $parent = $this->getParentNode();
             if ($parent !== null && isset($parent->disabled)) {
-                $legacyDisabled = (string)$parent->disabled;
-                $this->setValue($legacyDisabled === '1' ? '0' : '1');
+                return (string)$parent->disabled === '1' ? '0' : '1';
             }
         }
-        return parent::actionPostLoadingEvent();
+        return $value;
     }
 }
--- a/src/opnsense/mvc/app/models/OPNsense/Firewall/FieldTypes/DNatEnabledField.php
+++ b/src/opnsense/mvc/app/models/OPNsense/Firewall/FieldTypes/DNatEnabledField.php
@@ -1,0 +1,65 @@
+<?php
+
+/*
+ * Copyright (C) 2026 Deciso B.V.
+ * All rights reserved.
+ *
+ * Redistribution and use in source and binary forms, with or without
+ * modification, are permitted provided that the following conditions are met:
+ *
+ * 1. Redistributions of source code must retain the above copyright notice,
+ *    this list of conditions and the following disclaimer.
+ *
+ * 2. Redistributions in binary form must reproduce the above copyright
+ *    notice, this list of conditions and the following disclaimer in the
+ *    documentation and/or other materials provided with the distribution.
+ *
+ * THIS SOFTWARE IS PROVIDED ``AS IS'' AND ANY EXPRESS OR IMPLIED WARRANTIES,
+ * INCLUDING, BUT NOT LIMITED TO, THE IMPLIED WARRANTIES OF MERCHANTABILITY
+ * AND FITNESS FOR A PARTICULAR PURPOSE ARE DISCLAIMED. IN NO EVENT SHALL THE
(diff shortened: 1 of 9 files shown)


src/opnsense/mvc/app/models/OPNsense/Firewall/FieldTypes/DNatEnabledField.php
```diff
@@ -0,0 +1,61 @@
...
+        $value = (string)$value;
+        $parent = $this->getParentNode();
+        // Set disabled value opposite of enabled value for legacy compatibility.
+        if ($parent !== null && isset($parent->disabled) && $value !== '') {
+            $parent->disabled = $value === '1' ? '0' : '1';
+        }
+        return parent::setValue($value);
+    }
+
+    protected function actionPostLoadingEvent()
```
AdSchellevis: technically, I don't think we need a post loading event if we just overwrite getValue() and setValue(), the intent of your class is to act as a reflector, so it doesn't really have to know the value in order to use it.

src/opnsense/mvc/app/models/OPNsense/Firewall/FieldTypes/DNatIPProtocolField.php
```diff
@@ -0,0 +1,51 @@
...
+ * CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE)
+ * ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
+ * POSSIBILITY OF SUCH DAMAGE.
+ */
+
+namespace OPNsense\Firewall\FieldTypes;
+
+use OPNsense\Base\FieldTypes\OptionField;
+
+class DNatIPProtocolField extends OptionField
```
AdSchellevis: in this case you can just remove the option and use a migration te replace the value,

AdSchellevis (changes requested):

src/opnsense/mvc/app/models/OPNsense/Firewall/FieldTypes/DNatEnabledField.php
TomWalraven: that makes sense, i've applied this in 2bdf2db

src/opnsense/mvc/app/models/OPNsense/Firewall/FieldTypes/DNatIPProtocolField.php
TomWalraven: I used the Filter model for the migration since DNat model doesn't support migrations. Applied in 5d3e5c0

src/opnsense/mvc/app/models/OPNsense/Firewall/FieldTypes/DNatEnabledField.php
```diff
@@ -0,0 +1,62 @@
...
+{
+    public function setValue($value)
+    {
+        if ($value === null) {
+            return;
+        }
+        $value = (string)$value;
+        $parent = $this->getParentNode();
+        // Set disabled value opposite of enabled value for legacy compatibility.
+        if ($parent !== null && isset($parent->disabled) && $value !== '') {
```
AdSchellevis: I guess this was causing issue due to the setter being called during initialization, if that's the case, you move "enabled" below `disabled` in the model, so you can replace the block with a two line fix: 

```
$this->getParentNode()->disabled = $value === '1' ? '0' : '1';
return parent::setValue($value);
```

src/opnsense/mvc/app/models/OPNsense/Firewall/FieldTypes/DNatEnabledField.php
```diff
@@ -0,0 +1,62 @@
...
+            $parent->disabled = $value === '1' ? '0' : '1';
+        }
+        return parent::setValue($value);
+    }
+
+    public function getValue(): string
+    {
```
AdSchellevis, line 47: here we should always be able to "reflect" the inverse disabled flag, e.g.:

```
return $this->getParentNode()->disabled === '1' ? '0' : '1';
```

AdSchellevis (changes requested):

src/opnsense/mvc/app/models/OPNsense/Firewall/FieldTypes/DNatEnabledField.php
TomWalraven: Applied in 467bcf9. Let me know if there's anything else that should be changed😄

Issue #10181, opened 2026-04-21
KEA: Plug socket into dynamic pd route installation script


Monviech:
**Important notices**

Before you add a new report, we ask you kindly to acknowledge the following:

- [x] I have read the contributing guidelines at https://github.com/opnsense/core/blob/master/CONTRIBUTING.md
- [x] I am convinced that my issue is new after having checked both open and closed issues at https://github.com/opnsense/core/issues?q=is%3Aissue

**Is your feature request related to a problem? Please describe.**

The socket is better than the lease files. It will make the route installation better. The lease files on disk are not always the truth, and there is still the problem about lease files not appearing until a while which makes the daemon fail to start.

https://github.com/opnsense/core/blob/master/src/opnsense/scripts/kea/kea_prefix_watcher.py

**Describe the solution you like**

We can use methods established here:
https://github.com/opnsense/core/pull/10019

Poll the socket, act on new leases, or vanished leases. We try to keep the installed routes in sync with the running configuration best effort.
The running lease configuration is the same on master and backup in a HA.

**Describe alternatives you considered**

None.

**Additional context**

Potentially required to make this roadmap item more stable:
https://github.com/opnsense/core/issues/9941

Monviech:
Only issue is there is no reliable event mechanism still. Events are only emitted on the current primary KEA instance.

This means we have to poll, should be quite cheap though when using the domain socket.

Monviech:
Test script:

```
#!/usr/local/bin/python3

import ujson
from lib.kea_ctrl import KeaCtrl

resp = KeaCtrl.send_command(
    "lease6-get-page",
    {"from": "start", "limit": 10},
    "dhcp6"
)
print(ujson.dumps(resp, indent=2))
```

Response:

```
{
  "arguments": {
    "count": 1,
    "leases": [
      {
        "cltt": 1776929426,
        "duid": "00:01:00:01:31:3d:5f:3e:00:15:5d:00:ad:3c",
        "fqdn-fwd": false,
        "fqdn-rev": false,
        "hostname": "",
        "hw-address": "00:15:5d:00:ad:3c",
        "iaid": 3,
        "ip-address": "fd10:0:0:1100::",
        "preferred-lft": 2500,
        "prefix-len": 56,
        "state": 0,
        "subnet-id": 1,
        "type": "IA_PD",
        "valid-lft": 4000
      }
    ]
  },
  "result": 0,
  "text": "1 IPv6 lease(s) found."
}
```

[...]

PR #10197, opened 2026-04-23
Services: Kea DHCPv6: kea_prefix_watcher: Plug socket into dynamic pd route installation script

Monviech:
**Important notices**

- Model used: ChatGPT 5.4
- Extent of AI involvement: Crawled the KEA API docs, some back and forth brainstorming about polling vs. event based approach, evaluated some different socket outputs

---

**Describe the problem**

Problem is described in issue.

---

Use KeaCtrl python lib for kea_prefix_watcher to switch to polling the socket instead of streaming the csv lease files.

The running configuration has structured output, and we stay in memory without needing to read and parse any files.

The tradeoff is that we always have to poll the full set of leases, yet by using lease6-get-page which is especially made for efficient bulk operations, this should be rather cheap.

The main benefit is we always work with structured json data, can filter for ID_PD better, and reprocess the full truth on every poll interval to ensure the whole setup is self-healing

This also fixes that the script cannot start if no lease files exist (since they are irrelevant now)

For structured output check:
https://github.com/opnsense/core/issues/10181#issuecomment-4302537723

---

**Related issue**

https://github.com/opnsense/core/issues/10181

### DIFF
--- a/src/etc/inc/plugins.inc.d/kea.inc
+++ b/src/etc/inc/plugins.inc.d/kea.inc
@@ -173,7 +173,7 @@ function kea_configure_do($verbose = false)
             $keaDdns->generateConfig();
         }
         (new \OPNsense\Kea\KeaCtrlAgent())->generateConfig();
-        if ($keaDhcpv6->isEnabled() && $keaDhcpv6->hasPdPools()) {
+        if ($keaDhcpv6->isEnabled() && $keaDhcpv6->pd_pools->pd_pool->count() > 0) {
             mwexecfb(
                 '/usr/local/opnsense/scripts/kea/kea_prefix_watcher.py',
                 [],
--- a/src/etc/inc/plugins.inc.d/kea.inc
+++ b/src/etc/inc/plugins.inc.d/kea.inc
@@ -175,8 +175,7 @@ function kea_configure_do($verbose = false)
         (new \OPNsense\Kea\KeaCtrlAgent())->generateConfig();
         if ($keaDhcpv6->isEnabled() && $keaDhcpv6->hasPdPools()) {
             mwexecfb(
-                '/usr/local/opnsense/scripts/kea/kea_prefix_watcher.py --interval %s',
-                ['10'],
+                '/usr/local/opnsense/scripts/kea/kea_prefix_watcher.py',
                 [
                     'pidfile' => '/var/run/kea_prefix_watcher.pid',
                     'logfile' => '/var/log/kea/kea_prefix_watcher.log',
--- a/src/etc/inc/plugins.inc.d/kea.inc
+++ b/src/etc/inc/plugins.inc.d/kea.inc
@@ -173,10 +173,10 @@ function kea_configure_do($verbose = false)
             $keaDdns->generateConfig();
         }
         (new \OPNsense\Kea\KeaCtrlAgent())->generateConfig();
-        if ($keaDhcpv6->isEnabled()) {
+        if ($keaDhcpv6->isEnabled() && $keaDhcpv6->hasPdPools()) {
             mwexecfb(
                 '/usr/local/opnsense/scripts/kea/kea_prefix_watcher.py %s',
-                ['/var/db/kea/kea-leases6.csv'],
+                ['10'],
                 [
                     'pidfile' => '/var/run/kea_prefix_watcher.pid',
                     'logfile' => '/var/log/kea/kea_prefix_watcher.log',
--- a/src/etc/inc/plugins.inc.d/kea.inc
+++ b/src/etc/inc/plugins.inc.d/kea.inc
@@ -175,7 +175,7 @@ function kea_configure_do($verbose = false)
         (new \OPNsense\Kea\KeaCtrlAgent())->generateConfig();
         if ($keaDhcpv6->isEnabled() && $keaDhcpv6->hasPdPools()) {
             mwexecfb(
-                '/usr/local/opnsense/scripts/kea/kea_prefix_watcher.py %s',
+                '/usr/local/opnsense/scripts/kea/kea_prefix_watcher.py --interval %s',
                 ['10'],
                 [
                     'pidfile' => '/var/run/kea_prefix_watcher.pid',
--- a/src/etc/inc/plugins.inc.d/kea.inc
+++ b/src/etc/inc/plugins.inc.d/kea.inc
@@ -176,6 +176,7 @@ function kea_configure_do($verbose = false)
         if ($keaDhcpv6->isEnabled() && $keaDhcpv6->hasPdPools()) {
             mwexecfb(
                 '/usr/local/opnsense/scripts/kea/kea_prefix_watcher.py',
+                [],
                 [
                     'pidfile' => '/var/run/kea_prefix_watcher.pid',
                     'logfile' => '/var/log/kea/kea_prefix_watcher.log',
--- a/src/opnsense/mvc/app/models/OPNsense/Kea/KeaDhcpv6.php
+++ b/src/opnsense/mvc/app/models/OPNsense/Kea/KeaDhcpv6.php
@@ -119,15 +119,6 @@ class KeaDhcpv6 extends BaseModel
     }
 
     /**
-     * only start the kea_prefix_watcher if pd pools are actually configured
-     * @return bool
-     */
-    public function hasPdPools()
-    {
-        return $this->pd_pools->pd_pool->count() > 0;
-    }
-
-    /**
      * should filter rules be enabled
      * @return bool
      */
--- a/src/opnsense/mvc/app/models/OPNsense/Kea/KeaDhcpv6.php
+++ b/src/opnsense/mvc/app/models/OPNsense/Kea/KeaDhcpv6.php
@@ -124,10 +124,7 @@ class KeaDhcpv6 extends BaseModel
      */
     public function hasPdPools()
     {
-        foreach ($this->pd_pools->pd_pool->iterateItems() as $_) {
-            return true;  // found one
-        }
-        return false;
+        return $this->pd_pools->pd_pool->count() > 0;
     }
 
     /**
--- a/src/opnsense/mvc/app/models/OPNsense/Kea/KeaDhcpv6.php
+++ b/src/opnsense/mvc/app/models/OPNsense/Kea/KeaDhcpv6.php
@@ -119,6 +119,18 @@ class KeaDhcpv6 extends BaseModel
     }
 
     /**
+     * only start the kea_prefix_watcher if pd pools are actually configured
+     * @return bool
+     */
+    public function hasPdPools()
+    {
+        foreach ($this->pd_pools->pd_pool->iterateItems() as $_) {
+            return true;  // found one
+        }
+        return false;
+    }
+
+    /**
      * should filter rules be enabled
      * @return bool
      */
--- a/src/opnsense/scripts/kea/kea_prefix_watcher.py
+++ b/src/opnsense/scripts/kea/kea_prefix_watcher.py
@@ -28,6 +28,7 @@
 """
 import argparse
 import ipaddress
+import time
 import ujson
 import subprocess
 import syslog
--- a/src/opnsense/scripts/kea/kea_prefix_watcher.py
+++ b/src/opnsense/scripts/kea/kea_prefix_watcher.py
@@ -29,7 +29,6 @@
 import argparse
 import ipaddress
 import ujson
-import time
 import subprocess
 import syslog
 
@@ -37,7 +36,7 @@ from lib.kea_ctrl import KeaCtrl
 
 
 def yield_lease_records(service='dhcp6', limit=256, poll_interval=10):
-    """ 
+    """
     Poll KEA lease pages and yield records from domain socket
     https://kea.readthedocs.io/en/latest/api.html#lease6-get-page
     Since amount of leases can be large, using get-page is recommended.
@@ -56,8 +55,7 @@ def yield_lease_records(service='dhcp6', limit=256, poll_interval=10):
                         "address": lease.get("ip-address", ""),
                         "prefix_len": lease.get("prefix-len", 128),
                         "hwaddr": lease.get("hw-address", ""),
-                        "valid_lifetime": lease.get("valid-lft", 0),
-                        "expire": lease.get("cltt", 0) + lease.get("valid-lft", 0)
+                        "state": lease.get("state", 0)  # State 0 means active
                     }
 
             last_addr = leases[-1].get("ip-address")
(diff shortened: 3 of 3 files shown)


src/opnsense/scripts/kea/kea_prefix_watcher.py
```diff
@@ -101,40 +98,32 @@ def get(self, mac):
...
-                    # only add when still valid
-                    if subprocess.run(['/sbin/route', 'add', '-inet6', prefix, ll_addr], capture_output=True).returncode:
-                        syslog.syslog(syslog.LOG_ERR, "failed adding route %s -> %s" % (prefix, ll_addr))
-                    else:
-                        syslog.syslog(syslog.LOG_NOTICE, "add route %s -> %s" % (prefix, ll_addr))
+    for record in yield_lease_records(poll_interval=inputargs.interval):
+        # IA_PD: guaranteed via "type = IA_PD"
+        prefix = "%(address)s/%(prefix_len)d" %  record
+        if (prefix not in prefixes or prefixes[prefix].get('hwaddr') != record.get('hwaddr')) \
+                and record.get('state', 0) == 0:
```
Monviech, line 109: Since the diff is annoying here due to indentation change, here the most relevant changes highlighted.

src/opnsense/scripts/kea/kea_prefix_watcher.py
```diff
@@ -101,40 +98,32 @@ def get(self, mac):
...
+            ll_addr = hostwatch.get(record.get('hwaddr'))
+            if not ll_addr:
+                syslog.syslog(
+                    syslog.LOG_WARNING,
+                    "no LLA found for %s, skipping route %s" % (record.get('hwaddr'), prefix)
+                )
+                continue
+            # lazy drop
+            subprocess.run(['/sbin/route', 'delete', '-inet6', prefix], capture_output=True)
+            if record.get('state', 0) == 0:
```
Monviech, line 122: And here.

src/opnsense/scripts/kea/kea_prefix_watcher.py
```diff
@@ -101,40 +98,32 @@ def get(self, mac):
...
-                # lazy drop
-                subprocess.run(['/sbin/route', 'delete', '-inet6', prefix], capture_output=True)
-                if record.get('valid_lifetime', 0) > 0:
-                    # only add when still valid
-                    if subprocess.run(['/sbin/route', 'add', '-inet6', prefix, ll_addr], capture_output=True).returncode:
-                        syslog.syslog(syslog.LOG_ERR, "failed adding route %s -> %s" % (prefix, ll_addr))
-                    else:
-                        syslog.syslog(syslog.LOG_NOTICE, "add route %s -> %s" % (prefix, ll_addr))
+    for record in yield_lease_records(poll_interval=inputargs.interval):
+        # IA_PD: guaranteed via "type = IA_PD"
```
Monviech, line 106: Removed this check

`if record.get('lease_type', 0) == 2 and record.get('prefix_len', 128) <= 64:`

we are always safe due to earlier:

`if lease.get("type") == "IA_PD":`

src/etc/inc/plugins.inc.d/kea.inc
```diff
@@ -173,10 +173,10 @@ function kea_configure_do($verbose = false)
             $keaDdns->generateConfig();
         }
         (new \OPNsense\Kea\KeaCtrlAgent())->generateConfig();
-        if ($keaDhcpv6->isEnabled()) {
+        if ($keaDhcpv6->isEnabled() && $keaDhcpv6->hasPdPools()) {
```
Monviech: Since the polling behavior changed, we only start when PD pools exist, to only affect users who actually use the feature.

src/opnsense/scripts/kea/kea_prefix_watcher.py
```diff
@@ -26,49 +27,45 @@
...
-            else:
-                time.sleep(poll_interval)
+
+            for lease in leases:
+                if lease.get("type") == "IA_PD":
+                    yield {
+                        "address": lease.get("ip-address", ""),
+                        "prefix_len": lease.get("prefix-len", 128),
+                        "hwaddr": lease.get("hw-address", ""),
+                        "state": lease.get("state", 0)  # State 0 means active
```
Monviech: Time based logic has been removed. It looks like the state is enough to evaluate if a lease is valid or not.

src/opnsense/scripts/kea/kea_prefix_watcher.py
```diff
@@ -26,49 +27,45 @@
...
+            response = KeaCtrl.send_command('lease6-get-page', {"from": from_addr, "limit": limit}, service)
+            leases = response.get("arguments", {}).get("leases", [])
+            if not leases:
                 break
-            else:
-                time.sleep(poll_interval)
+
+            for lease in leases:
+                if lease.get("type") == "IA_PD":
+                    yield {
```
Monviech, line 54: For an example of the API structure please check:
https://github.com/opnsense/core/issues/10181#issuecomment-4302537723

src/etc/inc/plugins.inc.d/kea.inc
fichtner: when do we stop the service?

src/etc/inc/plugins.inc.d/kea.inc
```diff
@@ -173,10 +173,10 @@ function kea_configure_do($verbose = false)
...
             mwexecfb(
-                '/usr/local/opnsense/scripts/kea/kea_prefix_watcher.py %s',
-                ['/var/db/kea/kea-leases6.csv'],
+                '/usr/local/opnsense/scripts/kea/kea_prefix_watcher.py --interval %s',
```
fichtner: I think it's interesting you move the lease file to the daemon this but don't add a default for --interval which makes this invoke shorter :)

src/opnsense/mvc/app/models/OPNsense/Kea/KeaDhcpv6.php
```diff
@@ -118,6 +118,18 @@ public function isEnabled()
         return $this->general->enabled->isEqual('1') && !$this->general->interfaces->isEmpty();
     }
 
+    /**
+     * only start the kea_prefix_watcher if pd pools are actually configured
+     * @return bool
+     */
+    public function hasPdPools()
+    {
+        foreach ($this->pd_pools->pd_pool->iterateItems() as $_) {
```
fichtner: can't we just count the values here? count(items()) > 0 or !!count() if you will ;)

src/opnsense/scripts/kea/kea_prefix_watcher.py
```diff
@@ -101,40 +98,32 @@ def get(self, mac):
...
 
 
-
 if __name__ == '__main__':
     parser = argparse.ArgumentParser()
-    parser.add_argument('filename',
-                        help='filename to watch (including .1,.2, .completed)',
-                        default='/var/db/kea/kea-leases6.csv'
-    )
+    parser.add_argument('--interval', help='poll interval in seconds', default=10, type=int)
```
fichtner: you even set the default here :)

src/opnsense/scripts/kea/kea_prefix_watcher.py
fichtner: hide whitespaces makes this obvious, yep

src/etc/inc/plugins.inc.d/kea.inc
Monviech: on kea_configure_do
https://github.com/opnsense/core/blob/a27a2c7901614d26bd911e276f752450169ef9a2/src/etc/inc/plugins.inc.d/kea.inc#L159

But its a bit unfortunate since the running script has in memory records it will use to drop routes that do not exist anymore. If Kea is (re)configured that reference gets lost and any old route that has already been installed will stay.

Shouldn't be a big deal but its to keep in mind.

src/opnsense/mvc/app/models/OPNsense/Kea/KeaDhcpv6.php
Monviech: I'm interested in a Base function for this if you are interested in providing one, some of these pattern is used more often.

src/etc/inc/plugins.inc.d/kea.inc
Monviech: I removed the whole argparse as it did not make sense anymore, everything is inside the script as defaults now.

src/opnsense/scripts/kea/kea_prefix_watcher.py
```diff
@@ -25,50 +26,45 @@
...
-                if idx == 0 and lstpos is None:
-                    header = item
-                elif len(item) == len(header):
-                    result_rec = {}
-                    for i, key in enumerate(header):
-                        result_rec[key] = int(item[i]) if item[i].isdigit() else item[i]
-                    yield result_rec
-            lstpos = fhandle.tell()
-            if fn != filename:
+            response = KeaCtrl.send_command('lease6-get-page', {"from": from_addr, "limit": limit}, service)
```
swhite2, line 47: For reference, we can further limit the output through `subnets`, matching only on PD pools, but this would require passing the config along. I like the simplicity as is.

Monviech: We can keep this certainly in mind if the current approach is somehow not ideal for some reason. I have decided not to include pulling the config, parsing it, and then using it for simplicity reasons as you stated.

Thanks :)

src/opnsense/scripts/kea/kea_prefix_watcher.py
Monviech: Potentially this guard could be removed, it's duplicate now.

src/etc/inc/plugins.inc.d/kea.inc
```diff
@@ -173,10 +173,9 @@ function kea_configure_do($verbose = false)
             $keaDdns->generateConfig();
         }
         (new \OPNsense\Kea\KeaCtrlAgent())->generateConfig();
-        if ($keaDhcpv6->isEnabled()) {
+        if ($keaDhcpv6->isEnabled() && $keaDhcpv6->hasPdPools()) {
             mwexecfb(
-                '/usr/local/opnsense/scripts/kea/kea_prefix_watcher.py %s',
-                ['/var/db/kea/kea-leases6.csv'],
+                '/usr/local/opnsense/scripts/kea/kea_prefix_watcher.py',
```
fichtner, line 178: I think this breaks it, as you need to pass empty argument list [] still due to how mwexecfb now handles extra backgrounding options

Monviech: I have readded the empty argument list.

src/opnsense/scripts/kea/kea_prefix_watcher.py
Monviech: @fichtner

https://kea.readthedocs.io/en/latest/arm/hooks.html#the-lease4-get-by-lease6-get-by-commands

`the state parameter also accepts a string (vs. a number) for usual state names (default (or assigned) (0), declined (1), expired-reclaimed (2), released (3), and registered (4)).`

Probably needs `!= 3` instead if I understand this correctly, only going by state `0` is not sufficient.

Monviech: Hmm nope, I think only going with state `0` is the best choice here.

src/opnsense/mvc/app/models/OPNsense/Kea/KeaDhcpv6.php
Monviech: @fichtner I implemented your new base method here now

src/opnsense/mvc/app/models/OPNsense/Kea/KeaDhcpv6.php
```diff
@@ -118,6 +118,15 @@ public function isEnabled()
         return $this->general->enabled->isEqual('1') && !$this->general->interfaces->isEmpty();
     }
 
+    /**
+     * only start the kea_prefix_watcher if pd pools are actually configured
+     * @return bool
+     */
+    public function hasPdPools()
+    {
+        return $this->pd_pools->pd_pool->count() > 0;
```
fichtner: IMO this is canonical now so you could use it above and not adjust the model with an extra function

Monviech: good idea, saves some code, changed

fichtner (approved):

Monviech:
I retested the script and startup.

- startup works when pd pools exist
- no startup when no pd pools exist (service remains stopped after an apply)
- route for leases get installed:

```
{
  "arguments": {
    "count": 2,
    "leases": [
      {
        "cltt": 1777278908,
        "duid": "00:01:00:01:31:3d:5f:3e:00:15:5d:00:ad:3c",
        "fqdn-fwd": false,
        "fqdn-rev": false,
        "hostname": "",
        "hw-address": "00:15:5d:00:ad:3c",
        "iaid": 3,
        "ip-address": "fd10::1",
        "preferred-lft": 2500,
        "state": 0,
        "subnet-id": 1,
        "type": "IA_NA",
        "valid-lft": 4000
      },
      {
        "cltt": 1777277657,
        "duid": "00:01:00:01:31:3d:5f:3e:00:15:5d:00:ad:3c",
        "fqdn-fwd": false,
        "fqdn-rev": false,
        "hostname": "",
        "hw-address": "00:15:5d:00:ad:3c",
        "iaid": 3,
        "ip-address": "fd10:0:0:1100::",
        "preferred-lft": 2500,
        "prefix-len": 56,
        "state": 0,
        "subnet-id": 1,
        "type": "IA_PD",
        "valid-lft": 4000
      }
    ]
  },
  "result": 0,
  "text": "2 IPv6 lease(s) found."
}
```

```
/usr/local/opnsense/scripts/interfaces/list_hosts.py --proto inet6 --ndp
{"source":"arp-ndp","rows":[["hn0","00:15:5d:00:ad:40","fd42::1"],["hn0","00:15:5d:00:ad:40","fe80::215:5dff:fe00:ad40"],["hn2","00:15:5d:00:ad:42","fd98::1"],["hn2","00:15:5d:00:ad:3e","fe80::215:5dff:fe00:ad3e"],["hn2","00:15:5d:00:ad:3e","fe80::215:5dff:fe00:ad3e"],["hn2","00:15:5d:00:ad:3c","fe80::215:5dff:fe00:ad3c"],["hn2","00:15:5d:00:ad:42","fe80::215:5dff:fe00:ad42"]]}
```

`2026-04-27T08:41:05Noticekea-dhcp6add route fd10:0:0:1100::/56 -> fe80::215:5dff:fe00:ad3c%hn2`

```
Internet6:
Destination                       Gateway                       Flags         Netif Expire
::1                               link#1                        UHS             lo0
fd10:0:0:1100::/56                fe80::215:5dff:fe00:ad3c%hn2  UGS             hn2
```

Monviech:
Should be fine, I also tested what happens if you reserve the IA_PD network, the `state` does not change to something else than 0 (I have assumed it might go to "registered (4)", but it did not. Just triple checked assumptions.)

PR #10199, opened 2026-04-23
mvc: BaseField: add count()

fichtner:
**Important notices**

---

**Describe the problem**

@Monviech was using `iterateItems()` to see if any configuration exists for the field and a helper for this task appears to make sense.

---

Count similar to iterateItems, but only count to 1 if the field does not have child nodes.

---

**Related issue**

#10197

### DIFF
--- a/src/opnsense/mvc/app/models/OPNsense/Base/FieldTypes/BaseField.php
+++ b/src/opnsense/mvc/app/models/OPNsense/Base/FieldTypes/BaseField.php
@@ -345,6 +345,7 @@ abstract class BaseField
      */
     public function count(): int
     {
+        /* XXX could use count($this->getValues()) if we fix BaseField's version to reject empty */
         return $this->isContainer() ? iterator_count($this->iterateItems()) : 1;
     }
 
--- a/src/opnsense/mvc/app/models/OPNsense/Base/FieldTypes/BaseField.php
+++ b/src/opnsense/mvc/app/models/OPNsense/Base/FieldTypes/BaseField.php
@@ -340,6 +340,15 @@ abstract class BaseField
     }
 
     /**
+     * count the number of (non virtual) child nodes or 1 for the current node
+     * @return int count
+     */
+    public function count(): int
+    {
+        return $this->isContainer() ? iterator_count($this->iterateItems()) : 1;
+    }
+
+    /**
      * iterate all nodes recursively.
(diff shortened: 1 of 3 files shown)


src/opnsense/mvc/app/models/OPNsense/Base/FieldTypes/BaseField.php
```diff
@@ -339,6 +339,15 @@ public function iterateItems()
         }
     }
 
+    /**
+     * count the number of (non virtual) child nodes or 1 for the current node
+     * @return int count
+     */
+    public function count(): int
+    {
+        return count($this->getChildren()) ? iterator_count($this->iterateItems()) : 1;
```
fichtner: One idea here was to overload the else case with `count($this->getValues())` which could be practical as well although in reality we haven't needed it specifically yet. May be practical for the GUI when we want to hide the content from view but still show context. The downside is that the concept is a bit overloaded.

src/opnsense/mvc/app/models/OPNsense/Base/FieldTypes/BaseField.php
```diff
@@ -339,6 +339,15 @@ public function iterateItems()
         }
     }
 
+    /**
+     * count the number of (non virtual) child nodes or 1 for the current node
+     * @return int count
+     */
+    public function count(): int
+    {
+        return $this->isContainer() ? iterator_count($this->iterateItems()) : 1;
```
fichtner, line 349: One additional idea here is to overload the else case to count the number of values in the value field which could be practical as well although in reality we haven't needed it specifically yet. May be nice for the GUI when we want to hide the content from view but still show context. The downside is that the concept is a bit overloaded.

```suggestion
        return $this->isContainer() ? iterator_count($this->iterateItems()) : count($this->getValues());
```

swhite2 (approved):
Looks good as is IMO

src/opnsense/mvc/app/models/OPNsense/Base/FieldTypes/BaseField.php
fichtner, line 349:
```suggestion
        /* XXX could use count($this->getValues()) if we fix BaseField's version to reject empty */
        return $this->isContainer() ? iterator_count($this->iterateItems()) : 1;
```

PR #10201, opened 2026-04-24
Firewall: Use save method from ApiMutableModelControllerBase for log command, move rule command and savepoint action

### LINKED ISSUE #10198: Firewall: Rules [new] - remove direct "Config::getInstance()->save()"
**Important notices**

**Describe the bug**

When having a user with "user-config-readonly", it can safe the config by e.g. changing logging or moving a rule in the new firewall rules.

*Tip*: to validate your setup was working with the previous version, use opnsense-revert (https://docs.opnsense.org/manual/opnsense_tools.html#opnsense-revert)

**To Reproduce**

1. Go to 'System -> Access -> Users'
2. Create a new user:
    - Username: test
    - Group membership: admins
    - Privileges: System: Deny config write
4. Go to "Firewall - Rules [new]"
5. Toggle Logging of a rule
6. Config is saved

**Expected behavior**

Config is not saved

Most likely this: https://github.com/opnsense/core/blob/3c5ae0f3b3d72e15efd852061a77d85c43bc3d74/src/opnsense/mvc/app/controllers/OPNsense/Firewall/Api/FilterBaseController.php#L446

**Environment**

OPNsense 26.1.6

Monviech:
**Important notices**

- Model used:
- Extent of AI involvement:

---

**Related issue**

Fixes: https://github.com/opnsense/core/issues/10198

### DIFF
--- a/src/opnsense/mvc/app/controllers/OPNsense/Firewall/Api/FilterBaseController.php
+++ b/src/opnsense/mvc/app/controllers/OPNsense/Firewall/Api/FilterBaseController.php
@@ -31,6 +31,7 @@ namespace OPNsense\Firewall\Api;
 use OPNsense\Base\ApiMutableModelControllerBase;
 use OPNsense\Base\FieldTypes\PortField;
 use OPNsense\Base\UserException;
+use OPNsense\Core\ACL;
 use OPNsense\Core\Backend;
 use OPNsense\Core\Config;
 use OPNsense\Firewall\Alias;
@@ -301,7 +302,8 @@ abstract class FilterBaseController extends ApiMutableModelControllerBase
 
     public function applyAction($rollback_revision = null)
     {
-        if ($this->request->isPost()) {
+        // XXX: Privilege check is a workaround here
+        if ($this->request->isPost() && !(new ACL())->hasPrivilege($this->getUserName(), 'user-config-readonly')) {
             if ($rollback_revision != null) {
                 // background rollback timer
                 (new Backend())->configdpRun('filter rollback_timer', [$rollback_revision], true);
@@ -314,7 +316,8 @@ abstract class FilterBaseController extends ApiMutableModelControllerBase
 
     public function cancelRollbackAction($rollback_revision)
     {
-        if ($this->request->isPost()) {
+        // XXX: Privilege check is a workaround here
+        if ($this->request->isPost() && !(new ACL())->hasPrivilege($this->getUserName(), 'user-config-readonly')) {
             return array(
                 "status" => (new Backend())->configdpRun('filter cancel_rollback', [$rollback_revision])
             );
@@ -325,7 +328,8 @@ abstract class FilterBaseController extends ApiMutableModelControllerBase
 
     public function savepointAction()
     {
-        if ($this->request->isPost()) {
+        // XXX: Privilege check is a workaround here
+        if ($this->request->isPost() && !(new ACL())->hasPrivilege($this->getUserName(), 'user-config-readonly')) {
             // trigger a save, so we know revision->time matches our running config
             $this->save();
             return array(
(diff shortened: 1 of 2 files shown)


src/opnsense/mvc/app/controllers/OPNsense/Firewall/Api/FilterController.php
```diff
@@ -338,8 +338,7 @@ public function moveRuleBeforeAction($selected_uuid, $target_uuid)
         }
         if ($new_key !== null) {
             $selected_node->sequence = (string)$new_key;
-            $this->getModel()->serializeToConfig(false, true); /* we're only changing sequences, forcefully save */
-            Config::getInstance()->save();
+            $this->save(false, true);  /* only changing sequences, no need to validate */
```
Monviech: For reference, the firewall page has its own moveRuleBeforeAction as it is pretty different from the FilterBaseController method that gets used by the NAT pages, partly due to the priority group logic inside (floating, group...). NAT does not have those.

src/opnsense/mvc/app/controllers/OPNsense/Firewall/Api/FilterBaseController.php
```diff
@@ -327,7 +327,7 @@ public function savepointAction()
     {
         if ($this->request->isPost()) {
             // trigger a save, so we know revision->time matches our running config
-            Config::getInstance()->save();
+            $this->save();
```
Monviech: This is the safepoint feature.

On rollback action this gets saved in the model:
https://github.com/opnsense/core/blob/ea6f7103f1e27126abb3552cf30819c032b4549e/src/opnsense/mvc/app/models/OPNsense/Firewall/Filter.php#L376

I'm not sure how to approach this, or if we still need the savepoint/rollback Actions at all anymore.

AdSchellevis: I'm also doubting a bit if we should keep the  savepoint/rollback option, I don't know anyone using it myself (it was added ages ago). Maybe start with ensuring proper save now, open a ticket and plan for removal in 27.1?

AdSchellevis (approved):

src/opnsense/mvc/app/controllers/OPNsense/Firewall/Api/FilterBaseController.php
Monviech: The issue is the revertAction

https://github.com/opnsense/core/blob/ea6f7103f1e27126abb3552cf30819c032b4549e/src/opnsense/mvc/app/controllers/OPNsense/Firewall/Api/FilterBaseController.php#L341

It calls this

https://github.com/opnsense/core/blob/ea6f7103f1e27126abb3552cf30819c032b4549e/src/opnsense/mvc/app/models/OPNsense/Firewall/Filter.php#L364

Meaning the save isn't triggered directly from the controller.

Can I just add the readonly guard directly to the revertAction if it's removed later anyway, or skip this for now?

src/opnsense/mvc/app/controllers/OPNsense/Firewall/Api/FilterBaseController.php
```diff
@@ -442,8 +441,7 @@ protected function toggleRuleLogBase($uuid, $log, $node_reference)
         }
 
         $node->log = $log;
-        $mdl->serializeToConfig();
-        Config::getInstance()->save();
+        $this->save(false, true);  /* only changing log, no need to validate */
```
fichtner: is validating log/sequence an expensive operation?

Monviech: What happens if a rule has a different field that would trigger the validation. I think this does not take something like "isFieldChanged" into account when it validates so it can only target the exact log or sequence field?

fichtner: Normally it only checks if isFieldChanged unless full validation is requested

Monviech: I feel like it makes sense to enable validation then.

src/opnsense/mvc/app/controllers/OPNsense/Firewall/Api/FilterBaseController.php
Monviech: I feel like this could be a workaround?
https://github.com/opnsense/core/pull/10201/changes/79286a7561f6618a2f9881a179a127c21a3be91e

Issue #10202, opened 2026-04-24
Kea: allow reserving prefixes for IPv6 (for prefix delegation)


Monviech:
**Important notices**

Before you add a new report, we ask you kindly to acknowledge the following:

- [x] I have read the contributing guidelines at https://github.com/opnsense/core/blob/master/CONTRIBUTING.md
- [x] I am convinced that my issue is new after having checked both open and closed issues at https://github.com/opnsense/core/issues?q=is%3Aissue

**Is your feature request related to a problem? Please describe.**

Check if reserving a prefix delegation for a mac address is possible, right now input validation prevents it.

**Describe the solution you like**

Creating reservations for hosts for a prefix delegation

**Describe alternatives you considered**

If its not possible retain current behavior.

Monviech:
Judging from the KEA documentation, a reservation can contain both ip-addresses AND prefixes.

https://kb.isc.org/docs/what-are-host-reservations-how-to-use-them

```
"reservations": [
    {
        "duid": "01:02:03:04:05:06:07:08:09:0A",

        "ip-addresses": [
            "2001:db8:1:cafe::1"
        ],
        "prefixes": [
            "2001:db8:2:abcd::/64"
        ],
....
```

This means another field would have to be introduced in the current reservations that allows setting a prefix in addition to an ip address.

Monviech:
Fixed via: https://github.com/opnsense/core/pull/10206

w00000dy:
Thanks! 😄

Monviech:
@w00000dy No problem. Just for your information this was also something I fixed on the way:
https://github.com/opnsense/core/pull/10231

w00000dy:
Hi @Monviech,
I updated to version 26.1.7_1. Now, I can create reservations from the `/ui/kea/dhcp/leases6` page. However, the prefix delegation does not work when I add a reservation. It does work without a reservation.
It still looks a bit weird because it looks like I'm doing an IP address reservation and not a prefix delegation reservation.

Monviech:
Hello @w00000dy 

as far as I can tell from the changelog this here is not in 26.1.7. You might have to wait for a later release.
https://forum.opnsense.org/index.php?topic=51757.0

w00000dy:
All right! Then I'll wait for the next releases.

PR #10206, opened 2026-04-24
Services: Kea DHCPv6: Add prefix to reservations to allow for static PD allocations based on DUID/MAC

Monviech:
…

**Important notices**

- Model used: ChatGPT 5.4
- Extent of AI involvement: Help with the new Util function ``isIPv6PrefixInPrefix()``

---

**Describe the problem**

Reservations could only reserve addresses, not prefixes.

---

Adds prefixes to reservations as per KEA documentation.
https://kb.isc.org/docs/what-are-host-reservations-how-to-use-them

Allows to reserve a PD allocated prefix to a MAC address or DUID so upstream client/router always gets the same.
Fixes that prefix length was not shown in lease view.
Lease reservation button has some new magic to detect prefixes, to prefill correct field.

---

**Related issue**

https://github.com/opnsense/core/issues/10202

### DIFF
--- a/src/opnsense/mvc/app/models/OPNsense/Kea/KeaDhcpv6.php
+++ b/src/opnsense/mvc/app/models/OPNsense/Kea/KeaDhcpv6.php
@@ -225,10 +225,10 @@ class KeaDhcpv6 extends BaseModel
                     }
                 }
                 if (!$reservation->ip_address->isEmpty()) {
-                    $res['ip-addresses'] = explode(',', $reservation->ip_address->getValue());
+                    $res['ip-addresses'] = $reservation->ip_address->getValues();
                 }
                 if (!$reservation->prefix->isEmpty()) {
-                    $res['prefixes'] = explode(',', $reservation->prefix->getValue());
+                    $res['prefixes'] = $reservation->prefix->getValues();
                 }
                 if (!$reservation->domain_search->isEmpty()) {
                     $res['option-data'][] = [
--- a/src/opnsense/mvc/app/models/OPNsense/Kea/KeaDhcpv6.php
+++ b/src/opnsense/mvc/app/models/OPNsense/Kea/KeaDhcpv6.php
@@ -54,8 +54,14 @@ class KeaDhcpv6 extends BaseModel
             if ($subnet_node) {
                 $subnet = $subnet_node->subnet->getValue();
             }
-            if (!Util::isIPInCIDR($reservation->ip_address->getValue(), $subnet)) {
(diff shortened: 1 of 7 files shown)


src/opnsense/mvc/app/models/OPNsense/Kea/KeaDhcpv6.php
```diff
@@ -218,7 +224,12 @@ private function getConfigSubnets($ddns_enabled = false)
                         $res[str_replace('_', '-', $key)] = $reservation->$key->getValue();
                     }
                 }
-                $res['ip-addresses'] = explode(',', $reservation->ip_address->getValue());
+                if (!$reservation->ip_address->isEmpty()) {
+                    $res['ip-addresses'] = explode(',', $reservation->ip_address->getValue());
```
fichtner: getValues()? ;(

src/opnsense/mvc/app/models/OPNsense/Kea/KeaDhcpv6.php
```diff
@@ -218,7 +224,12 @@ private function getConfigSubnets($ddns_enabled = false)
...
+                }
+                if (!$reservation->prefix->isEmpty()) {
+                    $res['prefixes'] = explode(',', $reservation->prefix->getValue());
```
fichtner: same same

fichtner (approved):
thanks

src/opnsense/mvc/app/models/OPNsense/Kea/KeaDhcpv6.php
Monviech: I always forget getValues(), it's not my favorite somehow. All the other base field fellows have it way better with me somehow since they are more primitive I guess. xD

fichtner: that's ok, but my main concern is hardcoding "," which doesn't have to be the separator char

Monviech: I fixed it, its better with getValues(), you are correct.

Monviech:
I just noticed I still have to check what happens to static reservations that unbound receives.

Since the IP address is now optional if a Prefix has been added, not that something blows up.

I might choose to become less permissive and enforce IA_NA being present as IP address registration, always, combined with IA_PD. I don't want to go down the same exception rabbithole as dnsmasq.

Monviech:
Looks like static reservations are fine with empty address, it's skipped:

https://github.com/opnsense/core/blob/4c796dc4652d2db31f1a1683a2850b8e5e9efb6f/src/etc/inc/plugins.inc.d/kea.inc#L104-L105

PR #10220, opened 2026-04-27
Services: Kea DHCPv6: Allow customizing mac_sources and change default to ipv6-link-local

### LINKED ISSUE #10217: Incorrect LLA used in route to delegated prefix due to kea-dhcp6 extracting hardware address from DUID-LL[T]
**Important notices**

**Describe the bug**

Executive summary:

By default, kea-dhcp6 extracts the "hardware address" from DUID-LLT and DUID-LL types when creating leases. For delegated prefixes, opnsense picks up this hardware address and uses it to add a route. By design, the address used in the DUID-LL[T] is not necessarily that of the interface requesting the lease, so the route can get created with an incorrect and unreachable next-hop.

Fortunately there is a way to configure kea-dhcp6 to not use this extraction method. If opnsense is going to rely on the hardware address in the lease for routing, it seems it must apply this configuration.

Detail:

I've been experimenting a bit with Kea DHCP, and have run into an interesting issue. 

My "production" opnsense instance gets a /56 IA_PD from my ISP. I did get Kea basically working for IPv4 and v6, but wanted to experiment a bit further, so I created a sandbox opnsense instance to do this on. I wanted functional (internet routable) IPv6 in the sandbox. Prefix delegation appeared to work, and my sandbox LAN clients get appropriate addresses, but internet access was not working. On closer inspection, the route for IA_PD on the upstream firewall was pointing the LLA of the downstream firewall's LAN interface, but that's not reachable. It needs to point to the LLA of the WAN interface.

Upstream (production) firewall has a VLAN interface on a network in common with the WAN interface of the downstream firewall. Upstream firewall runs kea-dhcp6 on that interface, with an address pool and a PD pool:

```
            {
                "id": 2,
                "subnet": "2a07:aa:bb:cc06::\/64",
                "pools": [
                    {
                        "pool": "2a07:aa:bb:cc06::1:0-2a07:aa:bb:cc06::1:ffff"
                    }
                ],
                "pd-pools": [
                    {
                        "prefix": "2a07:aa:bb:ccf0::",
                        "prefix-len": 60,
                        "delegated-len": 62
                    }
                ],
                "reservations": [],
                "interface": "vlan0.6"
            }
        ],
```

Downstream (sandbox) firewall has two interfaces, igc0 and igc1. By default, opnsense uses the first interface (igc0) for LAN and the second (igc1) for WAN, and they are plumbed accordingly. The MAC addresses are 00:d0:b4:01:a5:a0 and 00:d0:b4:01:a5:a1 respectively.

dhcp6c defaults to generating a DUID-LLT using the LLA of the first interface it finds with a valid LLA (i.e. igc0, the LAN interface). It (correctly) uses the same DUID-LLT for DHCPv6 requests on ALL interfaces, including WAN (RFC 8415: "The choice of network interface can be completely arbitrary, as long as that interface provides a globally unique link-layer address for the link type; the same DUID-LLT SHOULD be used in configuring all network interfaces connected to the device, regardless of which interface's link-layer address was used to generate the DUID-LLT."). A packet capture of a DHCP request looks like:

```
Frame 5: Packet, 159 bytes on wire (1272 bits), 159 bytes captured (1272 bits)
Ethernet II, Src: Katsujima_01:a5:a1 (00:d0:b4:01:a5:a1), Dst: IPv6mcast_01:00:02 (33:33:00:01:00:02)
Internet Protocol Version 6, Src: fe80::2d0:b4ff:fe01:a5a1, Dst: ff02::1:2
User Datagram Protocol, Src Port: 546, Dst Port: 547
DHCPv6
    Message type: Solicit (1)
    Transaction ID: 0xa11308
    Client Identifier
        Option: Client Identifier (1)
        Length: 14
        DUID: 00010001317f500b00d0b401a5a0
        DUID Type: link-layer address plus time (1)
        Hardware type: Ethernet (1)
        DUID Time: Apr 25, 2026 11:12:27.000000000 GMT Daylight Time
        Link-layer address: 00:d0:b4:01:a5:a0
        Link-layer address (Ethernet): Katsujima_01:a5:a0 (00:d0:b4:01:a5:a0)
    Identity Association for Non-temporary Address
    Elapsed time
    Option Request
    Identity Association for Prefix Delegation

```

Note the MAC addresses in the packet source (WAN) and in the DUID (LAN).

kea-dhcp6 on the upstream firewall receives the DHCPc6 request for IA_NA and IA_PD, and grants leases accordingly. For DUID-LLT or DUID-LL, it "cleverly" extracts the "hardware-address" from the DUID (https://kea.readthedocs.io/en/kea-2.2.0/arm/dhcp6-srv.html#mac-hardware-addresses-in-dhcpv6 , https://github.com/isc-projects/kea/blob/b51e765d4a5356f49d0eee3075f855a037f16c96/src/lib/dhcp/pkt.cc#L217-L227), and records this as the "hwaddr" (`/var/db/kea/kea-leases6.csv*`) / "hw-address" (API).

```
root@opnsense:~ # echo '{ "command": "lease6-get-all", "service": [ "dhcp6" ] }' | socat - UNIX-CONNECT:/var/run/kea/kea6-ctrl-socket | jq '.arguments.leases[] | select(."type" == "IA_PD" and ."state" == 0) | {"ip-address": ."ip-address", state: .state, duid: .duid, "hw-address": ."hw-address"}'
{
  "ip-address": "2a07:aa:bb:ccfc::",
  "state": 0,
  "duid": "00:01:00:01:31:7f:50:0b:00:d0:b4:01:a5:a0",
  "hw-address": "00:d0:b4:01:a5:a0"
}
root@opnsense:~ #
```

Kea even admits (https://kea.readthedocs.io/en/kea-2.2.0/arm/dhcp6-srv.html#mac-hardware-addresses-in-dhcpv6) that this is forbidden by RFC 8415 ("Clients and servers MUST treat DUIDs as opaque values and MUST only compare DUIDs for equality. Clients and servers SHOULD NOT in any other way interpret DUIDs. [...] It should be noted that an attempt to parse a DUID to obtain a client's link-layer address is unreliable, as there is no guarantee that the client is still using the same link-layer address as when it generated its DUID.") but they do it (by default) anyway.

opnsense's "kea_prefix_watcher" then scans the `kea-leases6.csv*` files, picks up the IA_PD with its "hwaddr", and looks that MAC address up in "hostwatch" to determine the LLA to use in the downstream route to the IA_PD. Curiously, hostwatch (observed through `/usr/local/opnsense/scripts/interfaces/list_hosts.py --proto inet6 --ndp`) shows entries for both WAN *and* LAN interfaces of the downstream firewall, so the (wrong) LLA is found.

```
root@opnsense:~ # /usr/local/opnsense/scripts/interfaces/list_hosts.py --proto inet6 --ndp | jq '.rows | map(select(.[2] | startswith("fe80::2d0:b4ff:fe01:a5a")))'
[
  [
    "igc0",
    "00:d0:b4:01:a5:a0",
    "fe80::2d0:b4ff:fe01:a5a0"
  ],
  [
    "igc0",
    "00:d0:b4:01:a5:a2",
    "fe80::2d0:b4ff:fe01:a5a2"
  ],
  [
    "vlan0.6",
    "00:d0:b4:01:a5:a0",
    "fe80::2d0:b4ff:fe01:a5a0"
  ],
  [
    "vlan0.6",
    "00:d0:b4:01:a5:a1",
    "fe80::2d0:b4ff:fe01:a5a1"
  ]
]
root@opnsense:~ #
```

Therefore the upstream firewall ends up with a route to the IA_PD pointing to the LLA of the downstream firewall's LAN interface, which is unreachable:

```
root@opnsense:~ # route -6 get 2a07:aa:bb:ccfc::
   route to: 2a07:aa:bb:ccfc::
destination: 2a07:aa:bb:ccfc::
       mask: ffff:ffff:ffff:fffc::
    gateway: fe80::2d0:b4ff:fe01:a5a0%vlan0.6
        fib: 0
  interface: vlan0.6
      flags: <UP,GATEWAY,DONE,STATIC>
 recvpipe  sendpipe  ssthresh  rtt,msec    mtu        weight    expire
       0         0         0         0      1500         1         0
root@opnsense:~ #
```

and so return packets for attempts to ping google have nowhere to go:

```
root@opnsense:~ # tcpdump -nni vlan0.6 ip6
tcpdump: verbose output suppressed, use -v[v]... for full protocol decode
listening on vlan0.6, link-type EN10MB (Ethernet), snapshot length 262144 bytes
07:55:18.054908 IP6 fe80::7e83:34ff:feb8:dcc2 > ff02::1:ff01:a5a0: ICMP6, neighbor solicitation, who has fe80::2d0:b4ff:fe01:a5a0, length 32
07:55:18.060953 IP6 2a07:aa:bb:ccfc:6a7c:624f:630c:e4c8 > 2a00:1450:4009:c13::71: ICMP6, echo request, id 17, seq 8, length 64
07:55:19.054886 IP6 fe80::7e83:34ff:feb8:dcc2 > ff02::1:ff01:a5a0: ICMP6, neighbor solicitation, who has fe80::2d0:b4ff:fe01:a5a0, length 32
07:55:19.084820 IP6 2a07:aa:bb:ccfc:6a7c:624f:630c:e4c8 > 2a00:1450:4009:c13::71: ICMP6, echo request, id 17, seq 9, length 64
07:55:20.108888 IP6 2a07:aa:bb:ccfc:6a7c:624f:630c:e4c8 > 2a00:1450:4009:c13::71: ICMP6, echo request, id 17, seq 10, length 64
07:55:20.126340 IP6 fe80::7e83:34ff:feb8:dcc2 > ff02::1:ff01:a5a0: ICMP6, neighbor solicitation, who has fe80::2d0:b4ff:fe01:a5a0, length 32
07:55:21.125832 IP6 fe80::7e83:34ff:feb8:dcc2 > ff02::1:ff01:a5a0: ICMP6, neighbor solicitation, who has fe80::2d0:b4ff:fe01:a5a0, length 32
07:55:21.132967 IP6 2a07:aa:bb:ccfc:6a7c:624f:630c:e4c8 > 2a00:1450:4009:c13::71: ICMP6, echo request, id 17, seq 11, length 64
```

If I change the downstream firewall's dhcp6c to use a DUID-UUID instead of DUID-LLT, kea-dhcp6 grants a new lease, and it (apparently) uses the packet's source MAC address as the "hwaddr" (that of the WAN interface) in lieu of being able to (inappropriately) extract it from the DUID, and kea_prefix_watcher maps this to the correct LLA (the WAN one), and everything works. Explicitly setting to DUID-LLT or DUID-LL with the MAC address of the WAN interface also works.

Forunately it is possible to configure the `mac-sources` in kea-dhcp6 (https://kea.readthedocs.io/en/kea-2.2.0/arm/dhcp6-srv.html#mac-hardware-addresses-in-dhcpv6). It would be nice if there was a way to configure exclusions ("I want everything *except* the DUID method"), but it doesn't appear that there is, so we're stuck with:

```
{
    "Dhcp6": {
        "mac-sources": [
            "raw",
            "client-link-addr-option",
            "ipv6-link-local",
            "remote-id",
            "subscriber-id",
            "docsis-cmts",
            "docsis-modem"
        ]
     }
}
```

With that in place, the correct hardware address appears in the lease:

```
root@opnsense:~ # echo '{ "command": "lease6-get-all", "service": [ "dhcp6" ] }' | socat - UNIX-CONNECT:/var/run/kea/kea6-ctrl-socket | jq '.arguments.leases[] | select(."type" == "IA_PD" and ."state" == 0) | {"ip-address": ."ip-address", state: .state, duid: .duid, "hw-address": ."hw-address"}'
{
  "ip-address": "2a07:aa:bb:ccf0::",
  "state": 0,
  "duid": "00:01:00:01:31:7f:50:0b:00:d0:b4:01:a5:a0",
  "hw-address": "00:d0:b4:01:a5:a1"
}
root@opnsense:~ #
```

and the route is created correctly:

```
root@opnsense:~ # route -6 get 2a07:aa:bb:ccf0::
   route to: 2a07:aa:bb:ccf0::
destination: 2a07:aa:bb:ccf0::
       mask: ffff:ffff:ffff:fffc::
    gateway: fe80::2d0:b4ff:fe01:a5a1%vlan0.6
        fib: 0
  interface: vlan0.6
      flags: <UP,GATEWAY,DONE,STATIC>
 recvpipe  sendpipe  ssthresh  rtt,msec    mtu        weight    expire
       0         0         0         0      1500         1         0
root@opnsense:~ #
```

If opnsense is going to use the hardware address to form the routes for delegated prefixes, it appears it must apply this fix.

Side-note; scanning `/var/db/kea/kea-leases6.csv` seems potentially fragile. Shouldn't opnsense really be using the Kea API (i.e. a published interface) to get current lease details?

Further side notes; this my investigation turned up a couple of tangential bugs, which I'll report separately...

The opnsense UI (Interfaces -> Settings -> System DHCP Unique Identifier) actually generates an incorrectly formatted DUID-UUID. The "type" part of a DUID is always 2 bytes - per RFC 8415 "A DUID consists of a 2-octet type code represented in network byte order, followed by a variable number of octets that make up the actual identifier." and section 11.5 confirms that the type should be 16 bits (value of 4), so the DUID-UUID should start with "00:04:", not "00:00:00:04:". Curiously, though, kea-dhcp6 doesn't seem to care about the malformed DUID, and grants leases for it anyway.

Changing the "System DHCP Unique Identifier" doesn't seem to take effect unless/until I manually kill the `dhcp6c` process and then refresh (Interfaces -> Overview) the WAN interface. Just doing the refresh without killing the `dhcp6c` process results in the old DUID still being used.

**Environment**

OPNsense 26.1.6-amd64

Monviech:
**Important notices**

- Model used:
- Extent of AI involvement:

---

**Describe the problem**

Check issue.

---

Allow customizing mac_sources and change default to ipv6-link-local since it seems to align best with the expectations of our setup, especially taking PD route installation via kea_prefix_watcher.py into account.

Per default KEA would derive MAC addresses of clients from the DUID (using a precedence of methods via "any", but duid always comes first), but these do not take multiple interfaces into account. This means, the route target could be the wrong MAC address. The new default ipv6-link-local takes the EUI-64 assumption of the link-local address, this seems to be better suited as our default.

Two methods have been skipped since they are not implemented by KEA upstream, raw and subscriber-id.

---

**Related issue**

Fixes: https://github.com/opnsense/core/issues/10217

### DIFF
--- a/src/opnsense/mvc/app/controllers/OPNsense/Kea/forms/generalSettings6.xml
+++ b/src/opnsense/mvc/app/controllers/OPNsense/Kea/forms/generalSettings6.xml
@@ -52,8 +52,7 @@
         <id>dhcpv6.general.mac_sources</id>
         <label>MAC sources</label>
         <type>select_multiple</type>
-        <advanced>true</advanced>
-        <help>The DHCPv6 protocol does not provide any completely reliable way to retrieve hardware addresses of clients. To mitigate that issue, a number of mechanisms are available. Each of these mechanisms works in certain cases, but may not in others. Whether the mechanism works in a particular deployment is somewhat dependent on the network topology and the technologies used. Please note that this influences PD route installation, since the source MAC address of the client is required to target the link-local route.</help>
+        <help>The DHCPv6 protocol does not provide any completely reliable way to retrieve hardware addresses of clients. To mitigate that issue, a number of mechanisms are available. Each of these mechanisms works in certain cases, but may not in others. Whether the mechanism works in a particular deployment is somewhat dependent on the network topology and the technologies used. Please note that this influences PD route installation, since the source MAC address of the client is required to target the link-local route. It also influences MAC based reservations.</help>
     </field>
     <field>
         <id>dhcpv6.general.fwrules</id>
--- a/src/opnsense/mvc/app/controllers/OPNsense/Kea/forms/generalSettings6.xml
+++ b/src/opnsense/mvc/app/controllers/OPNsense/Kea/forms/generalSettings6.xml
@@ -49,6 +49,13 @@
         <help>The valid life time of the leases given out by the server in seconds.</help>
     </field>
     <field>
+        <id>dhcpv6.general.mac_sources</id>
+        <label>MAC sources</label>
+        <type>select_multiple</type>
+        <advanced>true</advanced>
+        <help>The DHCPv6 protocol does not provide any completely reliable way to retrieve hardware addresses of clients. To mitigate that issue, a number of mechanisms are available. Each of these mechanisms works in certain cases, but may not in others. Whether the mechanism works in a particular deployment is somewhat dependent on the network topology and the technologies used. Please note that this influences PD route installation, since the source MAC address of the client is required to target the link-local route.</help>
+    </field>
+    <field>
         <id>dhcpv6.general.fwrules</id>
         <label>Firewall rules</label>
         <type>checkbox</type>
--- a/src/opnsense/mvc/app/models/OPNsense/Kea/KeaDhcpv6.php
+++ b/src/opnsense/mvc/app/models/OPNsense/Kea/KeaDhcpv6.php
@@ -324,6 +324,7 @@ class KeaDhcpv6 extends BaseModel
         $cnf = [
             'Dhcp6' => [
                 'valid-lifetime' => $this->general->valid_lifetime->asInt(),
+                'mac-sources' => $this->general->mac_sources->getValues(),
                 'interfaces-config' => [
                     'interfaces' => $this->getConfigPhysicalInterfaces(),
                     /* socket retries are on a per-interface basis, failing to open one won't affect others */
--- a/src/opnsense/mvc/app/models/OPNsense/Kea/KeaDhcpv6.xml
+++ b/src/opnsense/mvc/app/models/OPNsense/Kea/KeaDhcpv6.xml
@@ -1,6 +1,6 @@
 <model>
     <mount>//OPNsense/Kea/dhcp6</mount>
-    <version>1.0.0</version>
+    <version>1.0.1</version>
     <description>Kea DHCPv6 configuration</description>
     <items>
         <general>
@@ -18,6 +18,19 @@
                 <Default>4000</Default>
                 <Required>Y</Required>
             </valid_lifetime>
+            <mac_sources type="OptionField">
+                <Default>ipv6-link-local</Default>
+                <Required>Y</Required>
+                <OptionValues>
+                    <duid>duid</duid>
+                    <ipv6-link-local>ipv6-link-local</ipv6-link-local>
+                    <client-link-addr-option>client-link-addr-option</client-link-addr-option>
+                    <remote-id>remote-id</remote-id>
+                    <docsis-cmts>docsis-cmts</docsis-cmts>
+                    <docsis-modem>docsis-modem</docsis-modem>
+                </OptionValues>
+                <Multiple>Y</Multiple>
+            </mac_sources>
             <fwrules type="BooleanField">
                 <Required>Y</Required>
                 <Default>1</Default>


Monviech:
@dseven As far as I understand, this might be the best default to ship:
https://github.com/opnsense/core/blob/388cf994965ab7d791dbaf93a4d49d501f6a27f3/src/opnsense/mvc/app/models/OPNsense/Kea/KeaDhcpv6.xml#L22

What is your opinion?

dseven:
Can you have multiple options enabled by default? My thinking was to make the change as minimally invasive as possible by making the sequence look the same as what `any` does, except with `duid` skipped (by default).

Monviech:
I think default also supports multiple comma separated values.

My intention with a single value though is that it streamlines the assumptions to one in default setups.

Any assumption is kinda bad here to be honest, a client could also not have an EUI-64 derived link local, and things worked with DUID before. I am kinda hesistant if we should maybe not implement our own default but just provide "any" as default, which would just be all values selected essentially.

EDIT: Though it looks like it's only a global option anyway so whatever we do some client will hit one or the other edge case lol.

dseven:
I wonder how likely it is that there'd be a case where DUID would be the only method that works. I tend to suspect that it would be much less likely than the cases where DUID just gets it plain wrong.

I'd be more concerned about turning off some of the other options by default impacting other use cases, such as a relay scenario (where I suppose the source address of the packet could belong to the relay, not the client).

Monviech:
The current PD route installation requires L2 adjacency though. Only "ipv6-link-local" and less so "duid" provide a MAC address of a client that's actually inside the same L2 broadcast domain.

Installing a PD route for a relayed client would currently also not give a correct result (it does not exist in the NDP table or hostwatch).

dseven:
Fair point. I suppose the lease hardware address is not meaningful for any other purpose (besides PD routing) in the opnsense context?

Monviech:
I think KEA also decides MAC based IPv6 reservations based on this derived assumption. That would also explain correlation with other tickets where user said "its not working for me (mac based reservation)", and it sounds plausible.

https://kea.readthedocs.io/en/latest/arm/dhcp6-srv.html#host-reservations-in-dhcpv6

Documentation clearly references the other spot here.

dseven:
Ahh, yes! If someone was trying to do hw-address reservations for a client that had generated a DUID-LL[T] based on some interface other than that one needing the lease, similar confusion could occur.

fichtner (approved):

Monviech:
@dseven Thank you for the discussion and your initial report. It helped to understand some more of these edges.

I know we might disagree here, but I will go with the single default as an initial step. Since the option is not advanced anymore, users that might be affected can choose what they desire.

The single default also helps documenting what we consider "best" for mac based reservations and PD route installation.

I'll make sure to also update the docs.

dseven:
No worries. I think it's a solid step. I was just trying to think though potential implications. Enjoyed the discussion too!

Issue #10224, opened 2026-04-27
Kea: Cannot delete prefix lease with lease delete button


Monviech:
**Important notices**

Before you add a new report, we ask you kindly to acknowledge the following:

- [x] I have read the contributing guidelines at https://github.com/opnsense/core/blob/master/CONTRIBUTING.md
- [x] I am convinced that my issue is new after having checked both open and closed issues at https://github.com/opnsense/core/issues?q=is%3Aissue

**Is your feature request related to a problem? Please describe.**

Find out why title happens.

Monviech:
Looks like this is the way, from the KEA docs, the type is needed.

The IPv6 address leases are deleted the same way, but using [lease6-del](https://kea.readthedocs.io/en/latest/arm/hooks.html#std-isccmd-lease6-del). The IPv6 prefix leases are also deleted using [lease6-del](https://kea.readthedocs.io/en/latest/arm/hooks.html#std-isccmd-lease6-del), but with some extra steps. The prefix should be referenced by its address and prefix length should be ignored. For example, to delete 2001:db8:1::/48, the following command can be used. The subnet-id parameter is optional.

```
{
    "command": "lease6-del",
    "arguments": {
        "ip-address": "2001:db8:1::",
        "type": "IA_PD",
        "subnet-id": 1
    }
}
```

PR #10227, opened 2026-04-28
Services: Kea DHCPv4/v6: Improve lease delete command for IPv6 so it can actually delete IA_NA and IA_PD leases

Monviech:
**Important notices**

- Model used:
- Extent of AI involvement:

---

**Describe the problem**

This was a rabbithole, but there was no other way to fix this than to spawn all those overrides for the commands and the additional parameter for the lease type.

---

Pass lease type for IPv6 so IA_NA and IA_PD leases are correctly deleted.

---

**Related issue**
https://github.com/opnsense/core/issues/10224

### DIFF
--- a/src/opnsense/scripts/kea/del_kea_leases.py
+++ b/src/opnsense/scripts/kea/del_kea_leases.py
@@ -33,8 +33,8 @@ from lib.kea_ctrl import KeaCtrl
 
 if __name__ == '__main__':
     parser = argparse.ArgumentParser()
-    parser.add_argument("--ip", required=True, help="IP address to delete")
-    parser.add_argument("--type", dest="lease_type", help="Lease type of IPv6 lease (e.g. IA_NA, IA_PD)", default=None)
+    parser.add_argument("--ip", required=True, help="IP address to delete (192.168.1.1 or fe80::1)")
+    parser.add_argument("--type", dest="lease_type", help="Lease type of IPv6 lease (IA_NA or IA_PD)", default=None)
 
     args = parser.parse_args()
     ip = args.ip.strip()
--- a/src/opnsense/scripts/kea/del_kea_leases.py
+++ b/src/opnsense/scripts/kea/del_kea_leases.py
@@ -33,8 +33,8 @@ from lib.kea_ctrl import KeaCtrl
 
 if __name__ == '__main__':
     parser = argparse.ArgumentParser()
-    parser.add_argument("ip", help="IP address(es) to delete, comma separated")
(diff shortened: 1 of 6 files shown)


src/opnsense/scripts/kea/del_kea_leases.py
```diff
@@ -33,8 +33,12 @@
 
 if __name__ == '__main__':
     parser = argparse.ArgumentParser()
-    parser.add_argument("ip", help="IP address(es) to delete, comma separated")
-    ips = [ip.strip() for ip in parser.parse_args().ip.split(',') if ip.strip()]
+    parser.add_argument("--ip", required=True, help="IP address(es) to delete, comma separated")
```
Monviech: gah I just noticed its not yet over. If multiple IP addresses are accepted, only a single type would be used for all.

Time to go in deeper...

fichtner: isn't there a way to disable batch delete :>

swhite2:
```
# git grep 'batchDelete'
src/opnsense/mvc/app/views/OPNsense/Routes/index.volt:            'options':{selection:true, multiSelect:true, batchDelete:false}
...
```

Monviech: Well currently the batch delete is working since it just loops the same API call for each address, always calling the script here with a single IP+Type combination.

So its hammering the script now, and not passing a single comma separated IP list like before.

So technically it works but its very hammer and nail.

swhite2: which, ironically, was the default behavior before `batchDelete` for any delete-selected operation

Monviech: I have ensured the whole lease delete pipeline is a single item now.

I can either keep the delete_selected behavior since it works (albeit less efficent than before), or remove it too.

Monviech:
After discussing a bit more with @swhite2 and weighing options, I will change the approach of this completely:

- I will use the lease get page endpoint to collect all IPv6 leases and build a small record database inside the delete lease script if there is a single IPv6 address present in the set
- I will then check all types for this IP address, it's very unlikely it exists as IA_NA and IA_PD at the same time. Not 0, but very very very unlikely.
- I will delete all types for that IP address

-> Main benefit, only script changes a bit, the rest of the controller, UI, etc stay untouched.
-> Theoretical drawback, in very very very unlikely cases, two IPv6 addresses might be deleted with one command.

PR #10237, opened 2026-04-30
Firewall: DNat: Display effective port when local-port is omitted

### LINKED ISSUE #9677: Firewall: NAT: Destination NAT: Should local-port be required?
**Important notices**

- Open Firewall - NAT - Destination NAT
- Create a new NAT rule with "Single port or range" selected in "local-port" field
- Leave empty and save

**Questions to answer** 

- Is the empty input allowed here? Or should the port be required?

**Additional context**

In the old Port Forward page the "local-port" was required in some circumstances:

Reference:
https://github.com/opnsense/core/blob/2e9ac2defb5e6df55b4055179bdea8606cbef235/src/www/firewall_nat_edit.php#L177-L179

TomWalraven:
**Important notices**

---

**Describe the problem**

Currently when selecting any/skipping local-port field, it's unclear which port is effective.

---

Display the selected destination port in the local-port field. Local-port will only display "*" if both destination port and local-port are omitted.

---

**Related issue**
Closes: #9677

### DIFF
--- a/src/opnsense/mvc/app/views/OPNsense/Firewall/nat_rule.volt
+++ b/src/opnsense/mvc/app/views/OPNsense/Firewall/nat_rule.volt
@@ -297,6 +297,15 @@
                         if (row.isGroup) {
                             return "";
                         }
+
+                        if (column.id === "local-port") {
+                            const localPort = row["local-port"] || "";
+                            const destinationPort = row["destination.port"] || "";
+                            const isOmitted = !localPort || localPort === "any";
+                            // Display "*" in case localPort and destinationPort are both omitted
+                            return (isOmitted ? destinationPort : localPort) || "*";
+                        }
+
                         const value = row[column.id] || "";
                         // DNAT uses network, SNAT and ONAT uses net
                         const isNegated = (row[column.id.replace(/network|net/, 'not')] == 1) ? "! " : "";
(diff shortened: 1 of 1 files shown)


Monviech:
@TomWalraven

I think we can take the chance and wrap this into elseif since the number of conditions grow.

I imagine the formatter like this:

```
                    alias: function(column, row) {
                        if (row.isGroup) {
                            return "";
                        }
                        const value = row[column.id] || "";
                        // DNAT uses network, SNAT and ONAT uses net
                        const isNegated = (row[column.id.replace(/network|net/, 'not')] == 1) ? "! " : "";

                        if (typeof value !== 'string') {
                            return '';
                        } else if (value === "local-port") {
                            // DNAT: mirror destination port into local-port for better visibility
                            return (!row["local-port"] ? row["destination.port"] : row["local-port"]) || "*";
                        } else if (!value || value === "any") {
                            return isNegated + '*';
                        }

                        const aliasMetadataList = row["alias_meta_" + column.id] || [];

                        ...
```

TomWalraven:
I've wrapped the code into else-if as you suggested, but i changed value === "local-port" to column.id === "local-port". Thank you! 96d5ac7ef11e82496a157573960a1eba498a73aa

Monviech:
Looks good, thank you :)

PR #10239, opened 2026-04-30
landing page redirect

### LINKED ISSUE #10238: users: landing page
**Important notices**

**Describe the bug**

- The landing page setting can be used for external redirects by utilizing the default normalization of browsers.
- If the user does not have access to the landing page, then they lose access to all pages

sopex:
**Important notices**

**Related issue**

Closes #10238

### DIFF
--- a/src/opnsense/mvc/app/controllers/OPNsense/Auth/forms/dialogUser.xml
+++ b/src/opnsense/mvc/app/controllers/OPNsense/Auth/forms/dialogUser.xml
@@ -86,6 +86,7 @@
         <label>Preferred landing page</label>
         <type>text</type>
         <help>Preferred landing page after login or authentication failure</help>
+        <hint>/ui/core/firmware#updates</hint>
         <grid_view>
             <visible>false</visible>
         </grid_view>
--- a/src/opnsense/mvc/app/controllers/OPNsense/Auth/forms/dialogUser.xml
+++ b/src/opnsense/mvc/app/controllers/OPNsense/Auth/forms/dialogUser.xml
@@ -86,7 +86,7 @@
         <label>Preferred landing page</label>
         <type>text</type>
         <help>Preferred landing page after login or authentication failure</help>
-        <hint>/ui/core/firmware#updates</hint>
+        <hint>/ui/component/controller#subpage</hint>
         <grid_view>
             <visible>false</visible>
         </grid_view>
--- a/src/opnsense/mvc/app/models/OPNsense/Auth/User.xml
+++ b/src/opnsense/mvc/app/models/OPNsense/Auth/User.xml
@@ -34,7 +34,7 @@
             <!-- password policy constraints store last pwd change timestamp -->
             <pwd_changed_at type="TextField"/>
             <landing_page type="TextField">
-                <Mask>^\/[a-zA-Z0-9_\-\.~\/?=&amp;#+,@!;]*$</Mask>
+                <Mask>/^\/[a-zA-Z0-9_\-\.~\/?=&amp;#+,@!;]*$/</Mask>
                 <ValidationMessage>Must be a valid relative path without backslashes or URL schemes.</ValidationMessage>
             </landing_page>
             <comment type="TextField"/>
--- a/src/opnsense/mvc/app/models/OPNsense/Auth/User.xml
+++ b/src/opnsense/mvc/app/models/OPNsense/Auth/User.xml
@@ -33,7 +33,10 @@
             <scrambled_password type="BooleanField" volatile="true"/>
             <!-- password policy constraints store last pwd change timestamp -->
             <pwd_changed_at type="TextField"/>
-            <landing_page type="TextField"/>
+            <landing_page type="TextField">
+                <Mask>^\/[a-zA-Z0-9_\-\.~\/?=&amp;#+,@!;]*$</Mask>
+                <ValidationMessage>Must be a valid relative path without backslashes or URL schemes.</ValidationMessage>
+            </landing_page>
             <comment type="TextField"/>
             <email type="TextField"/>
             <apikeys type=".\ApiKeyField"/>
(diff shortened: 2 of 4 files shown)


AdSchellevis:
@sopex it feels like we could simplify this, can you share your test examples which shouldn't be accepted?

sopex:
> @sopex it feels like we could simplify this, can you share your test examples which shouldn't be accepted?

Yes, sorry, I was absent minded.

`\\example.com`
`/\\example.com`
`\\\\example.com`

AdSchellevis:
wouldn't `ltrim($value, '/\\');` fix your problem?

sopex:
encoded slashes?

sopex:
You are probably correct. Give me a minute.

sopex:
%5C%5C Seems to pass
So maybe i decode it before the ltrim and loose the parse_url ?
I will try it later today @AdSchellevis

AdSchellevis:
@sopexno rush, thanks for looking at it.

sopex:
@AdSchellevis Seems fine to me. Also, it fixes a bug where, if the user didn't have the privileges to access the assigned landing page, no page was accessible.

src/opnsense/mvc/app/models/OPNsense/Core/ACL.php
```diff
@@ -412,10 +412,12 @@ public function getLandingPage($username)
         if (!empty($_SESSION['user_shouldChangePassword'])) {
             // ACL lock, may only access password page
             return "ui/user_portal";
-        } elseif (!empty($this->userDatabase[$username]['landing_page'])) {
-            // remove leading slash, which would result in redirection to //page (without host) after login or auth failure.
-            return ltrim($this->userDatabase[$username]['landing_page'], '/');
-        } elseif (!empty($this->userDatabase[$username])) {
+        } if (!empty($this->userDatabase[$username]['landing_page'])) {
+            $landing = ltrim(rawurldecode(trim($this->userDatabase[$username]['landing_page'])), '/\\');
+            if (!empty($landing) && $this->isPageAccessible($username, '/' . $landing)) {
```
AdSchellevis, line 417: ok, nice catch

src/www/authgui.inc
```diff
@@ -206,7 +206,8 @@ function session_auth()
                     $url = $acl->getLandingPage($_SESSION['Username']);
                     header(url_safe("Location: /{$url}"));
                 } else {
-                    header(url_safe("Location: {$_SERVER['REQUEST_URI']}"));
+                    $req_uri = '/' . ltrim($_SERVER['REQUEST_URI'], '/\\');
```
AdSchellevis: this I'm not sure about, why does this suddenly popup here?

src/opnsense/mvc/app/controllers/OPNsense/Auth/forms/dialogUser.xml
```diff
@@ -86,6 +86,7 @@
         <label>Preferred landing page</label>
         <type>text</type>
         <help>Preferred landing page after login or authentication failure</help>
+        <hint>/ui/core/firmware#updates</hint>
```
fichtner: I'm not sure this adds value. Maybe it should be explained generally in the help text, but random links that can't be copied are not helpful.

src/opnsense/mvc/app/models/OPNsense/Auth/User.xml
```diff
@@ -33,7 +33,10 @@
             <scrambled_password type="BooleanField" volatile="true"/>
             <!-- password policy constraints store last pwd change timestamp -->
             <pwd_changed_at type="TextField"/>
-            <landing_page type="TextField"/>
+            <landing_page type="TextField">
+                <Mask>/^\/[a-zA-Z0-9_\-\.~\/?=&amp;#+,@!;]*$/</Mask>
```
fichtner, line 37: would extending UrlField for this make sense? For one there is a regex here and also the &amp; sticks out since it's XML but then not really

AdSchellevis: I'm not sure we do have similar other use-cases to be honest, in which case it might be better to isolate the issue here. The `UrlField` is basically just a `FILTER_VALIDATE_URL`  in php, which is much broader than the allowed patterns here

src/opnsense/mvc/app/controllers/OPNsense/Auth/forms/dialogUser.xml
AdSchellevis: let's drop this then, nobody ever asked questions about this anyway

sopex: I believe it will increase usage of the feature.
Also, why make it copyable? The goal is for the user to understand, at a glance, what the feature is and how to use it.

src/www/authgui.inc
sopex: I cant recreate it. Clicking x.x.x.x/\\\example.com as a link seemed to work before. But its probably not the case.

AdSchellevis (approved):
I don't have a very strong opinion about the `hint` tag, so I'll leave this to @fichtner to decide on next steps. The rest in my opinion looks good.

src/opnsense/mvc/app/controllers/OPNsense/Auth/forms/dialogUser.xml
sopex: @fichtner Shall I delete it, or can we move forward as is?

fichtner: My specific issue is that it shows the firmware page which is neither generic nor common.

Options:

1. drop
2. use pseudo page hint `/ui/component/controller`
3. use some other debatable page from core, but it could lead back to the initial concern.

sopex: > My specific issue is that it shows the firmware page which is neither generic nor common.
> 
> Options:
> 
> 1. drop
> 2. use pseudo page hint `/ui/component/controller`
> 3. use some other debatable page from core, but it could lead back to the initial concern.

I chose that because it uses the # subpage, so it is a full example but I will just use a pseudo page

fichtner (approved):

fichtner:
Merged, thanks!

PR #10252, opened 2026-05-04
Services: Kea DHCPv6: Dynamic prefix delegation

### LINKED ISSUE #9941: kea: dynamic PD support for DHCPv6 WAN
**Important notices**

For feature parity with isc-dhcp and dnsmasq we need automatic PD support from a WAN interface.

A small lightweight approach would be nice. Kea doesn't have tools for this like Dnsmasq, but we want to avoid complications that isc-dhcp has had.

Ask Kea for better support? ;)

**Additional context**

N/A

Monviech:
**Important notices**

- Model used: ChatGPT 5.4
- Extent of AI involvement: Brainstorming about this, rubberducking, help with IPv6 splitting functions and stuff like that that handles prefix calculations

---

**Describe the problem**

Fixes: https://github.com/opnsense/core/issues/9941

---

The scope is intentionally limited. Validations prevent interactions we do not want. Things that can make KEA crash are auto generated. The user cannot decide much to protect from hard to troubleshoot issues.

The design revolves around two steps:
1. Create a fully functional KEA config, prevent the user from having too much freedom by constraining every input possible (this can be relaxed later). Keep the GUI as lean as possible, no dynamic hints, dynamic grid values etc.
To allow KEA to always start, even without any valid prefixes on interfaces, placeholder prefixes are emitted to the subnet, and a client-classes test prevents clients from getting leases from such subnets.
2. A hook script that will take care of prefix changes and subnet lease whipe by using user-context keys from the running KEA config and writing them to the KEA socket. (Read write). We do not need a post apply action this way.
The hook script is indempodent and very cheap to execute. A failure or a wrong configuration file cannot crash the running KEA daemon.

A concept for 1 is in POC phase and finished
A concept for 2 is in POC phase and finished

---

**Generated Configuration file example**
Placeholder prefix:
```
        "subnet6": [
            {
                "id": 1,
                "subnet": "dead:beef::\/58",
                "option-data": [],
                "pools": [
                    {
                        "pool": "dead:beef::\/64"
                    }
                ],
                "pd-pools": [
                    {
                        "prefix": "dead:beef:0:20::",
                        "prefix-len": 59,
                        "delegated-len": 64,
                        "user-context": {
                            "uuid": "c6dabbe0-a0b2-4688-b198-a8402b9b2723"
                        }
                    }
                ],
                "reservations": [],
                "interface": "hn2",
                "user-context": {
                    "uuid": "ab5bdee9-6381-487e-966c-a1cee1bd0a39",
                    "dynamic_prefix": true,
                    "prefix_valid": false,
                    "prefix_source": "wan"
                },
                "require-client-classes": [
                    "NO_LEASES_PLEASE"
                ],
            },
            {
                "id": 2,
                "subnet": "dead:beef:0:68::\/61",
                "option-data": [],
                "pools": [
                    {
                        "pool": "dead:beef:0:68::\/64"
                    }
                ],
                "pd-pools": [
                    {
                        "prefix": "dead:beef:0:6c::",
                        "prefix-len": 62,
                        "delegated-len": 64,
                        "user-context": {
                            "uuid": "fa0a7323-dad8-47ea-92cc-b43bd6e61d37"
                        }
                    }
                ],
                "reservations": [],
                "interface": "hn3",
                "user-context": {
                    "uuid": "057de7ee-bc6e-4677-b6df-8b0ccdedf084",
                    "dynamic_prefix": true,
                    "prefix_valid": false,
                    "prefix_source": "wan"
                },
                "require-client-classes": [
                    "NO_LEASES_PLEASE"
                ],
            },
        ],
        "client-classes": [
            {
                "name": "NO_LEASES_PLEASE",
                "test": "not member('ALL')",
                "only-if-required": true
            }
    }
}
```

"Real" prefix:
```
# cat /tmp/hn1_prefixv6
2001:db8:1234::/56
```
```
        "subnet6": [
            {
                "id": 1,
                "subnet": "2001:db8:1234::\/58",
                "option-data": [],
                "pools": [
                    {
                        "pool": "2001:db8:1234::\/64"
                    }
                ],
                "pd-pools": [
                    {
                        "prefix": "2001:db8:1234:20::",
                        "prefix-len": 59,
                        "delegated-len": 64,
                        "user-context": {
                            "uuid": "c6dabbe0-a0b2-4688-b198-a8402b9b2723"
                        }
                    }
                ],
                "reservations": [],
                "interface": "hn2",
                "user-context": {
                    "uuid": "ab5bdee9-6381-487e-966c-a1cee1bd0a39",
                    "dynamic_prefix": true,
                    "prefix_valid": true,
                    "prefix_source": "wan"
                },
            },
            {
                "id": 2,
                "subnet": "2001:db8:1234:68::\/61",
                "option-data": [],
                "pools": [
                    {
                        "pool": "2001:db8:1234:68::\/64"
                    }
                ],
                "pd-pools": [
                    {
                        "prefix": "2001:db8:1234:6c::",
                        "prefix-len": 62,
                        "delegated-len": 64,
                        "user-context": {
                            "uuid": "fa0a7323-dad8-47ea-92cc-b43bd6e61d37"
                        }
                    }
                ],
                "reservations": [],
                "interface": "hn3",
                "user-context": {
                    "uuid": "057de7ee-bc6e-4677-b6df-8b0ccdedf084",
                    "dynamic_prefix": true,
                    "prefix_valid": true,
                    "prefix_source": "wan"
                },
            },
```


src/opnsense/mvc/app/views/OPNsense/Kea/dhcpv6.volt
```diff
@@ -120,14 +136,38 @@
         /**
          *
          */
-        $("#subnet4\\.option_data_autocollect").change(function(){
+        $("#subnet6\\.option_data_autocollect").change(function(){
```
fichtner, line 146: is this a bug on stable that needs fixing separately?

fichtner (commented):
this looks nice! my only worries here are:

1. Validation of "track interface" relationship between subnet and prefix is not well-defined. In the old days you'd select a PD to be on a specific interface in which case it would already know which prefix to use (by looking up LAN -> WAN).
2. Only one dynamic PD with the current enforcement, but need to check the old code and the GUI workflow in order to judge.

src/opnsense/mvc/app/views/OPNsense/Kea/dhcpv6.volt
Monviech: yeah I just spotted it randomly so I changed it before I forget it

Monviech:
The main constraint here is that a subnet in kea must be unique. So you cannot attach the same prefix to multiple interfaces.

If we want to have that we would need more GUI magic that allows the WAN prefix to be splitted into smaller prefixes, or implement some partial address logic. Both would be fragile.

The current state that allows at least one network to host dynamic PD is a good start I think.

Im open for ideas here.

fichtner:
If that's the constraint we want to go forward with I don't mind. It also limits the scope and one PD is better than none :)

Monviech:
I try really hard to limit the scope with this one. Expanding scope later should be doable if users have needs that cannot be met with this model.

EG adding a new prefix ID field or something the like.

But since providers also like playing switcheroo with delegated prefix sizes for single customers all flexibility will bite someone at some point.

EDIT: All of these concerns to not apply anymore to the new model we use now that allows full flexibility.

src/opnsense/mvc/app/models/OPNsense/Kea/KeaDhcpv6.php
```diff
@@ -204,24 +311,54 @@ private function getConfigSubnets($ddns_enabled = false)
...
-                $record['pools'][] = ['pool' => $pool];
+            if (!$subnet->prefix_source->isEmpty()) {
+                // We auto generate a default IA_NA pool for dynamic prefixes.
+                // Allowing configuration in the GUI would be too brittle since KEAs configuration model is too static.
+                // It has no idea about something like "::1000-::2000" or a concept like partial addresses.
+                // This here assumes that Autoconf::getPrefix() always returns something like "2001:db8:1234:1::/64".
+                // XXX: It does not know about prefixes that might already be assigned via identity association by
+                // inside interface configurations (used by firewall itself)
+                $prefix = explode('/', $record['subnet'], 2)[0];
+                $record['pools'][] = ['pool' => "{$prefix}1000-{$prefix}2000"];
```
Monviech: @fichtner

If the provider delegates a /56 prefix, and the code splits it into two /57 prefixes, the second /57 can be used entirely by the prefix delegation IA_PD.

This here would use the first /64 prefix via this range, but thats essentially a bit bad because it would most likely overlap with prefixes the firewall uses itself on some interfaces.

It would be better to calculate the last available /64 prefix inside the first /57 prefix and set it here verbatim, the pool can be XXX/64 as well, not only a range.

Hmm no it should be the prefix that exists on the parent (LAN) interface already, otherwise it wont be something on-link.

src/opnsense/mvc/app/models/OPNsense/Kea/KeaDhcpv6.xml
```diff
@@ -128,6 +127,14 @@
                 <interface type="InterfaceField">
                     <Required>Y</Required>
                 </interface>
+                <prefix_source type="InterfaceField">
+                    <Constraints>
+                        <check001>
+                            <type>UniqueConstraint</type>
+                            <ValidationMessage>Prefix source interface must be unique per subnet.</ValidationMessage>
```
Monviech: A side effect here is that one interface can set the WAN prefix, but other interfaces can set the LAN/OPTX prefixes (if they exist as files as well for ifctl to read).

i'm pretty sure KEA will scream if a larger prefix overlaps with a carved out smaller prefix so this might need additional validation to prevent.

What I don't want to prevent is blocking users who have multiple WAN interfaces with different prefixes.

src/opnsense/mvc/app/models/OPNsense/Kea/KeaDhcpv6.php
Monviech: @fichtner 

So to configure the correct /64 pool here we would need pull the prefix of the LAN interface as well.
Does it also create the ifctl prefix file when its configured via identity_association?

Though funnily we don't need IA_NA for our prefix delegation since we use link local addresses as target, so the pool could be skipped entirely for our purpose. Its allowed to be empty.

fichtner: according to the v6 code we already have "interface" kea in IPv6 subnet but I suppose you want to know what was assigned there via ifctl which we currently don't have. But we do know the prefix ID from the WAN (and its address). It kind of circles back to not offering a selector for the tracking interface since we know where the prefix comes from with a bit of magic that NPT is using:

https://github.com/opnsense/core/blob/dd63dd1a8e506658c0d4742c7b3790e28ee601e3/src/opnsense/mvc/app/models/OPNsense/Firewall/Filter.php#L289-L299

Monviech: I think something like that would create some sort of expectation mismatch.

If multiple "subnet6" in KEA could be made "dynamic", and they all circle back to the prefix of the WAN interface, multiple subnets would then have the same subnet prefix, which is invalid.

From a very strict operational view, I think we can skip any pools for now, since I do not need to know it. Our kea_prefix_watcher script discovers MAC and installs routes via link-local address. IA_NA is not needed as any operational target.

If multiple pools should exist that can all do some sort of PD, we would need to calculate better and decrease the size of the subnet prefix per subnet in KEA.

Though that would most likely clash with how users usually configure their track interface or interface association, where they count up 1 id per subnet. There is nothing inbetween a prefix subnet that is larger than /64 could be calculated out of to also allow an IA_PD pool.

**Example with IP addresses:**

If the provider delegates this to WAN:

`WAN delegated prefix: 2001:db8:1234::/56
`
and multiple Kea subnet6 entries could be marked as dynamic and all point back to the WAN prefix via their tracked interface, they would all resolve to the same generated subnet:
```

LAN  dynamic subnet6 -> 2001:db8:1234::/56
OPT1 dynamic subnet6 -> 2001:db8:1234::/56
OPT2 dynamic subnet6 -> 2001:db8:1234::/56
```

That is invalid in Kea because the subnet6 entries overlap completely. Kea needs each subnet6 to be unique.

A dynamic PD-only configuration could look like this:

`WAN delegated prefix: 2001:db8:1234::/56`

```
Kea subnet6:
  subnet: 2001:db8:1234::/56
  pools:  []
  pd-pools:
    prefix:        2001:db8:1234::
    prefix-len:    56
    delegated-len: 64
```

That avoids inventing an IA_NA pool like:

`2001:db8:1234::1000 - 2001:db8:1234::2000`

If multiple dynamic PD-capable subnets should exist later, we would need an explicit slicing model. For example, from a provider /56 we could carve out larger-than-/64 Kea subnet containers:

`WAN delegated prefix: 2001:db8:1234::/56`

```
Kea dynamic subnet A: 2001:db8:1234:00::/58
Kea dynamic subnet B: 2001:db8:1234:40::/58
Kea dynamic subnet C: 2001:db8:1234:80::/58
Kea dynamic subnet D: 2001:db8:1234:c0::/58
```

Each of those could then contain its own PD pool, for example delegating /64s.

But that would likely clash with how users normally configure IPv6 tracking / identity association in OPNsense. Users usually think in terms of one prefix ID per interface:

```
LAN  prefix ID 0 -> 2001:db8:1234:00::/64
OPT1 prefix ID 1 -> 2001:db8:1234:01::/64
OPT2 prefix ID 2 -> 2001:db8:1234:02::/64
```

This means the model we would need for dynamic PD really belongs in the interface tracking configuration itself. Today, prefix IDs usually allocate adjacent /64s. If we wanted every tracked interface to also have room for delegated prefixes, the interface configuration would need to reserve larger spaced blocks instead, for example a /62 or /61 per interface, and then place the on-link /64 inside that block.

Monviech: @fichtner What I think I need is a helper that collects all prefix ids users have configured on their tracked interfaces, and returns the subnets they are contained in.

E.g. if a user configured:
```
WAN delegated prefix: 2001:db8:1234::/56
LAN 2001:db8:1234:00::/64 -> on link
OPT1 2001:db8:1234:40::/64 -> on link
OPT2 2001:db8:1234:60::/64 -> on link
```

It returns maximum prefix sizes for each that are free so KEA can directly consume them as subnets:

```
LAN  -> 2001:db8:1234:00::/58
OPT1 -> 2001:db8:1234:40::/59
OPT2 -> 2001:db8:1234:60::/59
```

If users then don't leave any space, it can be validated easily as well and returned that there is no free space for any PD in there.

fichtner: this overcomplicates it again I think. Most of the time users will use sequential IDs so the use of the second half of the actual pool makes the most sense as that would most likely be free given we have at least a /62 or /61

Monviech: It depends who has the authority over the prefixes. From my view its the interface configuration who has this authority, and should return which prefixes (calculated from the user configuration) are available to be configured as pools.

But you are right, an assumption like splitting the prefix is a good short term win that spawns no complexity at all, it just has to be documented cleanly, and it would only allow a single interface that has (dynamic) PD delegations.

src/opnsense/mvc/app/views/OPNsense/Kea/dhcpv6.volt
fichtner: please commit this one to master tomorrow :)

fichtner (approved):
tentative approval for the draft. looks snappy.

Monviech:
@fichtner Now the hook script will be easy. It doesnt need to know or calculate anything :)

- Regenerate template to get new dhcpv6 config
- Use config-reload (no need to post a file to any socket) https://kea.readthedocs.io/en/latest/arm/ctrl-channel.html#the-config-reload-command
- Wipe old leases in the dynamic subnet ids https://kea.readthedocs.io/en/latest/api.html#lease6-wipe

src/etc/inc/plugins.inc.d/kea.inc
```diff
@@ -146,10 +146,29 @@ function kea_staticmap($proto = null, $valid_addresses = true, $ifconfig_details
...
+function kea_generate_dhcpv6_do($verbose = false)
+{
+    kea_generate_dhcpv6();
+}
+
+function kea_generate_dhcpv6($keaDhcpv6 = null)
+{
+    if ($keaDhcpv6 === null) {
+        $keaDhcpv6 = new \OPNsense\Kea\KeaDhcpv6();
+    }
```
fichtner:
```suggestion
    $keaDhcpv6 ??= new \OPNsense\Kea\KeaDhcpv6();
```

auto-approve by @swhite2

src/etc/inc/plugins.inc.d/kea.inc
```diff
@@ -187,6 +203,21 @@ function kea_configure_do($verbose = false)
     }
 }
 
+function kea_newwanip($interfaces = [], $family = null)
```
fichtner:
```suggestion
function kea_newwanip($interfaces, $family)
```

these are ensured, no defaults

src/etc/inc/plugins.inc.d/kea.inc
```diff
@@ -187,6 +203,21 @@ function kea_configure_do($verbose = false)
...
+        return;
+    }
+    $keaDhcpv6 = new \OPNsense\Kea\KeaDhcpv6();
+    if (!$keaDhcpv6->isEnabled() || $keaDhcpv6->general->manual_config->isEmpty() === false) {
+        return;
+    }
+    if ($keaDhcpv6->pd_pools->pd_pool->count() == 0) {
+        return;
+    }
+    mwexec('/usr/local/opnsense/scripts/kea/kea_prefix_renew.py');
```
fichtner: I'd prefer nested ifs here instead of a number of returns

src/etc/inc/plugins.inc.d/kea.inc
```diff
@@ -146,10 +146,27 @@ function kea_staticmap($proto = null, $valid_addresses = true, $ifconfig_details
...
 
+// used by kea_prefix_renew hook script to regenerate config when dynamic IPv6 prefix changes
+function kea_generate_dhcpv6_do($verbose = false)
+{
+    kea_generate_dhcpv6();
+}
+
+function kea_generate_dhcpv6($keaDhcpv6 = null)
+{
+    $keaDhcpv6 ??= new \OPNsense\Kea\KeaDhcpv6();
```
fichtner:
```suggestion
    kea_generate_dhcpv6(new \OPNsense\Kea\KeaDhcpv6());
}

function kea_generate_dhcpv6($keaDhcpv6 )
{
```

on second glance this is what you wanted to do?

src/etc/inc/plugins.inc.d/kea.inc
Monviech: faiir, changed that

src/etc/inc/plugins.inc.d/kea.inc
```diff
@@ -187,6 +201,20 @@ function kea_configure_do($verbose = false)
...
+function kea_newwanip($interfaces, $family)
+{
+    if ($family === 'inet6') {
+        $keaDhcpv6 = new \OPNsense\Kea\KeaDhcpv6();
+        if (
+            $keaDhcpv6->isEnabled() &&
+            $keaDhcpv6->general->manual_config->isEmpty() &&
+            $keaDhcpv6->pd_pools->pd_pool->count() > 0
+        ) {
+            mwexec('/usr/local/opnsense/scripts/kea/kea_prefix_renew.py');
```
fichtner: in which parallel universe does mwexec() still exist? (:

Monviech: Oh yeah right...

src/etc/inc/plugins.inc.d/kea.inc
Monviech: I dont always want to instatiate the model two times.

It is already instantiated here:

https://github.com/opnsense/core/blob/8ed46cd3f5f1a78905280659ce69ed1425aca2f3/src/etc/inc/plugins.inc.d/kea.inc#L173

So I pass it into here:

https://github.com/opnsense/core/blob/8ed46cd3f5f1a78905280659ce69ed1425aca2f3/src/etc/inc/plugins.inc.d/kea.inc#L184

Which means I only need to instantiate it in kea_generate_dhcpv6 if it has not been passed into it yet, aka an outside caller that cannot pass the model.

fichtner: it's not. there are only two callers. if both provide the model we don't need to try and load it

src/opnsense/mvc/app/models/OPNsense/Kea/KeaDhcpv6.xml
```diff
@@ -220,7 +223,9 @@
                         <subnets>
                             <source>OPNsense.Kea.KeaDhcpv6</source>
                             <items>subnets.subnet6</items>
-                            <display>subnet</display>
+                            <display>interface,subnet</display>
+                            <!-- XXX: interface does not display its description for some reason -->
```
Monviech: @fichtner 

Marking a few where I need your help.

src/opnsense/mvc/app/models/OPNsense/Kea/KeaDhcpv6.xml
```diff
@@ -404,7 +409,9 @@
...
```
Monviech: Same

src/opnsense/mvc/app/views/OPNsense/Kea/dhcpv6.volt
```diff
@@ -82,8 +82,24 @@
                             },
                             options: {
                                 triggerEditFor: getUrlHash('edit'),
-                                initialSearchPhrase: getUrlHash('search')
-                            }
+                                initialSearchPhrase: getUrlHash('search'),
+                                formatters: {
+                                    subnet: function(column, row) {
+                                        if ((row.subnet || '') === '') {
+                                            // XXX: Could somehow dynamically insert current values from running KEA config but thats more glue than this
```
Monviech: Should we show what has been calculated or is this enough as first step?

I would need help how to pass the idassoc and kea config generated values into the grid and into the dialog.

src/opnsense/mvc/app/library/OPNsense/Interface/Idassoc.php
```diff
@@ -0,0 +1,172 @@
...
+            }
+        }
+
+        return $result;
+    }
+
+    /**
+     * Return configured IPv6 identity association prefix information.
+     */
+    public static function prefix($ifname = null): array
```
Monviech, line 173: If nothing exists here because nothing exists e.g., in  /tmp/igc0_prefixv6, KEA will fail to start.

But there isn't much we can do, it would be cooler though if the prefix file is not in tmp to help after reboots, not sure how feasible this is though or how much of a problem that is.

I just noticed it mainly because I manually created that file for testing.

fichtner: doesn't make any sense. in that case we need to defer emitting the dynamic PD

Monviech: Just to document it, only the **initial** start of KEA needs a valid config.

After the daemon has started once, using the socket reloads just results in an error during runtime, but the daemon continues to run and only the reconfiguring fails, so the old config is still in use.

```
2026-05-07T13:15:23

Critical
kea-dhcp6
FATAL [kea-dhcp6.dhcp6.0x49181605c008] DHCP6_DYNAMIC_RECONFIGURATION_FAIL dynamic server reconfiguration failed with file: /usr/local/etc/kea/kea-dhcp6.conf

2026-05-07T13:15:23
Error
kea-dhcp6
ERROR [kea-dhcp6.dhcp6.0x49181605c008] DHCP6_CONFIG_LOAD_FAIL configuration error using file: /usr/local/etc/kea/kea-dhcp6.conf, reason: subnet configuration failed: Invalid subnet syntax (prefix/len expected): (/usr/local/etc/kea/kea-dhcp6.conf:37:27)

root@opn-dev-02:/src/git/core # cat /var/run/kea/kea-dhcp6.kea-dhcp6.pid
10338
root@opn-dev-02:/src/git/core # /usr/local/opnsense/scripts/kea/kea_prefix_renew.py
{"wiped":[1,2],"generated":true,"reloaded":true}
root@opn-dev-02:/src/git/core # cat /var/run/kea/kea-dhcp6.kea-dhcp6.pid
10338
```

That means we create a temporary prefix if we do not get one from /tmp/ due to race conditions/slow WAN, etc... and use that as a placeholder so that the initial config generation succeeds.

Afterwards, we can always trigger the hook script on WAN events and at some point we have the correct prefixes in there.

Monviech: Fixed by always offering a temporary prefix, and adding a client_class that prevents clients from getting leases from that temporary prefix as long as its in the config.

src/opnsense/mvc/app/views/OPNsense/Kea/dhcpv6.volt
fichtner: if you emit %subnet from the controller it should be there without the need for a formatter?

src/opnsense/mvc/app/models/OPNsense/Kea/KeaDhcpv6.xml
fichtner: This is probably an issue with BaseListField use. It uses the value not the label.

src/opnsense/mvc/app/views/OPNsense/Kea/dhcpv6.volt
Monviech: would probably inflate the POC too much, lets go there later, the current choice is pragmatic.

src/etc/inc/plugins.inc.d/kea.inc
```diff
@@ -187,6 +200,20 @@ function kea_configure_do($verbose = false)
...
 }
 
+function kea_newwanip($interfaces, $family)
+{
+    if ($family === 'inet6') {
+        $keaDhcpv6 = new \OPNsense\Kea\KeaDhcpv6();
+        if (
+            $keaDhcpv6->isEnabled() &&
+            $keaDhcpv6->general->manual_config->isEmpty() &&
+            $keaDhcpv6->pd_pools->pd_pool->count() > 0
```
Monviech: This condition is too much since the subnet owns the dynamic prefix, and it can exist without a pd pool configured as well.

Monviech:
Please be gentle Copilot

src/opnsense/mvc/app/models/OPNsense/Kea/KeaDhcpv6.php
```diff
@@ -188,6 +308,16 @@ private function getConfigSubnets($ddns_enabled = false)
...
             }
+            if (!$subnet->dynamic_prefix->isEmpty()) {
+                // Used by hook script to know which subnets have a dynamic prefix, it reads the running conf from socket
+                $record['user-context']['dynamic_prefix'] = true;
+                $record['user-context']['prefix_valid'] = $idassoc['prefix_valid'] ?? false;
+                $record['user-context']['prefix_source'] = $idassoc['prefix_source'] ?? $if;
+                // If the prefix is temporary placeholder, we will not send leases to any client
+                if (empty($idassoc['prefix_valid'])) {
+                    $record['evaluate-client-classes'] = ['NO_LEASES_PLEASE'];
+                }
```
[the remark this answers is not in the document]
Monviech: @copilot This aligns to a deprecation message sent by the daemon itself. The configuration is valid.

2026-05-08T08:09:42Warningkea-dhcp6WARN [kea-dhcp6.dhcpsrv.0x43cacee5c008] DHCPSRV_ONLY_IF_REQUIRED_DEPRECATED The parameter 'only-if-required' is deprecated. Use 'only-in-additional-list' instead.

It has been changed due to this fact.

src/opnsense/mvc/app/models/OPNsense/Kea/KeaDhcpv6.php
```diff
@@ -389,6 +538,15 @@ public function generateConfig($target = '/usr/local/etc/kea/kea-dhcp6.conf')
...
         ];
         $client_classes = $this->getConfigClientClasses();
+
+        // Used by temporary dynamic-prefix placeholder subnets.
+        // The class can never match, and is only evaluated when required by such a subnet.
+        $client_classes[] = [
+            'name' => 'NO_LEASES_PLEASE',
+            'test' => "not member('ALL')",
+            'only-in-additional-list' => true,
+        ];
```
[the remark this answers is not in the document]
Monviech: @copilot  This aligns to a deprecation message sent by the daemon itself. The configuration is valid.

2026-05-08T08:09:42Warningkea-dhcp6WARN [kea-dhcp6.dhcpsrv.0x43cacee5c008] DHCPSRV_ONLY_IF_REQUIRED_DEPRECATED The parameter 'only-if-required' is deprecated. Use 'only-in-additional-list' instead.

It has been changed due to this fact.

Monviech:
No, bad copilot, don't commit changes! Now I have to clean up. Meh.

src/opnsense/scripts/kea/kea_prefix_renew.py
```diff
@@ -0,0 +1,70 @@
...
+    if failed:
+        result["status"] = "failed"
+
+    try:
+        KeaCtrl.send_command("config-reload", None, "dhcp6")
+
+    except Exception as e:
+        result["status"] = "failed"
+        syslog.syslog(syslog.LOG_ERR, "failed config-reload of Kea DHCPv6 configuration: %s" % e)
+
```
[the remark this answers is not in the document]
Monviech: If the config generation fails it does not mean the config that exists on disk is wrong. It will just be the one that has existed before reloading the config.

src/opnsense/scripts/kea/kea_prefix_renew.py
```diff
@@ -0,0 +1,70 @@
...
+            if subnet.get("id") is not None and subnet.get("user-context", {}).get("dynamic_prefix") is True:
+                subnet_id = int(subnet.get("id"))
+                KeaCtrl.send_command("lease6-wipe", {"subnet-id": subnet_id}, "dhcp6")
+
+    except Exception as e:
+        result["status"] = "failed"
+        syslog.syslog(syslog.LOG_ERR, "failed wiping dynamic prefix leases: %s" % e)
+
+    try:
+        failed = subprocess.run(["pluginctl", "kea_generate_dhcpv6"], check=False, capture_output=True).returncode != 0
```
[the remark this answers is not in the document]
Monviech: Syslog is enough in my opinion.

src/opnsense/mvc/app/models/OPNsense/Kea/KeaDhcpv6.php
```diff
@@ -161,9 +273,17 @@ private function getConfigSubnets($ddns_enabled = false)
...
         $subnet_id = 1;
         foreach ($this->subnets->subnet6->iterateItems() as $subnet_uuid => $subnet) {
+            // If subnet is dynamic, seed an initial subnet value so KEA can start
+            $if = $subnet->interface->getValue();
+            $subnet_value = $subnet->subnet->getValue();
+            $idassoc = [];
+            if (!$subnet->dynamic_prefix->isEmpty()) {
+                $idassoc = Idassoc::prefix($if);
+                $subnet_value = $idassoc['prefix_allocated'] ?? '';
+            }
```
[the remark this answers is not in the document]
Monviech: This is the only one that makes sense, but the validation of this would be tricky. I don't want to emit fake interfaces, that is a hard boundary.

Since this cannot happen during start-up or boot, it's less critical. I would take this as a known shortcoming for now.

src/opnsense/mvc/app/models/OPNsense/Kea/KeaDhcpv6.php
```diff
@@ -389,6 +538,15 @@ public function generateConfig($target = '/usr/local/etc/kea/kea-dhcp6.conf')
             ]
         ];
         $client_classes = $this->getConfigClientClasses();
+
+        // Used by temporary dynamic-prefix placeholder subnets.
+        // The class can never match, and is only evaluated when required by such a subnet.
+        $client_classes[] = [
+            'name' => 'NO_LEASES_PLEASE',
+            'test' => "not member('ALL')",
+            'only-in-additional-list' => true,
```
Monviech: This is wrong here after all. KEA documentation:

The only-in-additional-list flag is not mandatory; when its value is set to false (the default), membership is determined during classification and is available for subnet selection, for instance. When the value is set to true, membership is evaluated only if the class appears in an evaluate-additional-classes list and is usable only for option configuration.

src/opnsense/mvc/app/models/OPNsense/Kea/KeaDhcpv6.php
```diff
@@ -188,6 +308,16 @@ private function getConfigSubnets($ddns_enabled = false)
...
                 $record['user-context']['description'] = $subnet->description->getValue();
             }
+            if (!$subnet->dynamic_prefix->isEmpty()) {
+                // Used by hook script to know which subnets have a dynamic prefix, it reads the running conf from socket
+                $record['user-context']['dynamic_prefix'] = true;
+                $record['user-context']['prefix_valid'] = $idassoc['prefix_valid'] ?? false;
+                $record['user-context']['prefix_source'] = $idassoc['prefix_source'] ?? $if;
+                // If the prefix is temporary placeholder, we will not send leases to any client
+                if (empty($idassoc['prefix_valid'])) {
+                    $record['evaluate-client-classes'] = ['NO_LEASES_PLEASE'];
```
Monviech: this should also just become client-classes

src/opnsense/mvc/app/views/OPNsense/Kea/dhcpv6.volt
Monviech: Interestingly the code surrounding it didn't exist in dhcpv6. I implemented an option_data_autocollect inside this PR now so users can at least have an automatically tracked DHCPv6 DNS server if enabled.

fichtner (approved):
for me ok to be on master, great work 👍

Monviech:
@fichtner Thank you for your help with this, nice teamwork.

Monviech:
@fichtner I just tested this branch with the same patches we offered in the forum.

It seems to just workTM, I would release this.

ISP -> Fritzbox (delegated /56) -> OPNsense A (KEA) (delegated /60) -> OPNsense B (delegated /62)

Fritzbox:
image

OPNsense A:
image

image

image

image

image

image

OPNsense B:

image

image

fichtner:
Ok sure, let’s do it then. Thanks for testing!

src/opnsense/mvc/app/library/OPNsense/Interface/Idassoc.php
```diff
@@ -0,0 +1,183 @@
...
+    /**
+     * Calculate the on-link /64 prefix selected by a hexadecimal identity association prefix ID.
+     */
+    private static function calculatePrefix($source_prefix, $prefix_id, $prefix_len = 64): string
+    {
+        [$address, $source_prefix_len] = explode('/', $source_prefix, 2);
+        $bytes = array_values(unpack('C*', inet_pton($address)));
+
+        $source_prefix_len = (int)$source_prefix_len;
+        $prefix_id = hexdec((string)$prefix_id);
```
schreibubi, line 48: @Monviech @fichtner prefix_id here is already in decimal, no need to do a hexdec conversion here. This currently leads to issues with the kea-dhcp6.conf file containing incorrect "subnet" prefixes.

src/opnsense/mvc/app/library/OPNsense/Interface/Idassoc.php
```diff
@@ -0,0 +1,183 @@
...
+    /**
+     * Return the largest aligned prefix length starting at this track6-prefix-id
+     * within the configured track6_prefix_range.
+     *
+     * If track6_prefix_range is empty, assume one /64 slot.
+     */
+    private static function calculateUsablePrefixLength($source_prefix_len, $track6_prefix_id, $track6_prefix_range = ''): int
+    {
+        $source_prefix_len = (int)$source_prefix_len;
+        $track6_prefix_id = hexdec((string)$track6_prefix_id);
```
schreibubi, line 72: @Monviech @fichtner Same issue as above.

src/opnsense/mvc/app/library/OPNsense/Interface/Idassoc.php
Monviech: @schreibubi Please create a new issue that shows some of your configuration, the expected result, and the current result.

Relevant information is the interface configuration including the stored track6-prefix-id value, the delegated prefix, the expected prefix, and the actual prefix generated in kea-dhcp6.conf.

Let's not discuss this in a merged PR, thank you.

Monviech: @schreibubi You seem to be right about this, we missed that. I will patch this and also create a unit test for this file.

Monviech: https://github.com/opnsense/core/commit/b0e8dfcc54ee5f4e9eb1f8a2591fd92f28d1197b

Issue #9902, opened 2026-03-06
Idle timeout behaves like hard timeout; sessions dropped despite active traffic


derhelge:
**Important notices**

Before you add a new report, we ask you kindly to acknowledge the following:

- [x] I have read the contributing guide lines at https://github.com/opnsense/core/blob/master/CONTRIBUTING.md
- [x] I am convinced that my issue is new after having checked both open and closed issues at https://github.com/opnsense/core/issues?q=is%3Aissue

**Describe the bug**

Captive Portal users are disconnected/logged out when the Idle Timeout is reached, even while they are actively using the network. Behavior is as if Idle Timeout acts like a hard timeout. This affects all users.

It looks like user activity is not being recognized/registered by the captive portal session tracking.

**To Reproduce**

1.Enable Captive Portal on interface opt28 (vlan024).
2. Authenticate a WLAN client in the portal.
3. Generate continuous traffic (e.g. web browsing / persistent connections).
4. Wait until Idle Timeout is reached.
5. Observe the client session gets disconnected/logged out.

Running pfctl commands against the captive portal anchor fails; cannot list rules for the zone anchor:

```
# pfctl -a captiveportal_zone_0 -sr -v
No ALTQ support in kernel
ALTQ related functions disabled
pfctl: DIOCGETRULES: Invalid argument

# pfctl -a captiveportal_zone_0 -sn -v
No ALTQ support in kernel
ALTQ related functions disabled
pfctl: DIOCGETRULES: Invalid argument
```
Attempting to inspect ether rules for that anchor also fails:

```
# pfctl -vs ether -a captiveportal_zone_0
No ALTQ support in kernel
ALTQ related functions disabled
pfctl: DIOCGETETHRULES: No such file or directory
```

There are active PF states visible:
```
# pfctl -a captiveportal_zone_0 -ss | head
all tcp 142.xx.168.188:5228 <- 10.4.xxx.216:60884       ESTABLISHED:ESTABLISHED
all tcp xx.37.5.5:33159 (10.4.xxx.216:60884) -> 142.251.168.188:5228       ESTABLISHED:ESTABLISHED
all tcp xx.37.5.11:443 <- 176.5.85.xx:51227       FIN_WAIT_2:ESTABLISHED
#...
```

**Environment**
- OPNsense: 25.10.2 (business)
-  Captive Portal zone: 0
-  Interface:
 -  Device: vlan024
 -  Name: WLANTHGA
 -  Identifier: opt28
[...]

PR #10253, opened 2026-05-04
Captive Portal: reload ipfw on captive portal reconfigure

swhite2:
**Important notices**

- Model used:
- Extent of AI involvement:

---

**Describe the problem**

When having no shaper rules/pipes configured, reloading captive portal doesn't enable/start IPFW when configured through the GUI. If the device is rebooted afterwards or fully reconfigured, or if there are any pipes defines, this issues goes unnoticed. The only symptom is missing byte counters through IPFW.

Disabling the Captive portal also does not disable ipfw, a single reconfigure on the shaper fixes this.

---

Reload the IPFW template + reconfigure ipfw in the captive portal service controller

---

**Related issue**

https://github.com/opnsense/core/issues/9902

### DIFF
--- a/src/opnsense/mvc/app/controllers/OPNsense/CaptivePortal/Api/ServiceController.php
+++ b/src/opnsense/mvc/app/controllers/OPNsense/CaptivePortal/Api/ServiceController.php
@@ -54,7 +54,6 @@ class ServiceController extends ApiMutableServiceControllerBase
             }
 
             $status = parent::reconfigureAction();
-
         }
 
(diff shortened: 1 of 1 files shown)


fichtner (approved):
doesn't seem to make sense extending ApiMutableServiceControllerBase for this use case so it looks good

AdSchellevis (approved):
`template reload OPNsense/IPFW` was missing in the mix, sounds reasonable to solve that here indeed.

PR #10256, opened 2026-05-04
openvpn: check group membership from current config

### LINKED ISSUE #10255: OpenVPN local group constraint does not see LDAP group sync on first login
**Important notices**

- [ x ] I have read the contributing guidelines at https://github.com/opnsense/core/blob/master/CONTRIBUTING.md
- [ x ] I am convinced that my issue is new after having checked both open and closed issues at https://github.com/opnsense/core/issues?q=is%3Aissue

**Describe the bug**

When OpenVPN user authentication uses an LDAP backend with group synchronization, a newly synchronized user may be denied by the OpenVPN local group constraint on the first login attempt.

The authentication itself succeeds and LDAP group membership is synchronized, but the OpenVPN group check reads group membership through the legacy `getUserGroups()` path. That helper uses the global config state loaded before authentication, so group membership synchronized during the same login flow may not be visible until a later attempt.

**To Reproduce**

1. Configure an LDAP authentication server with user/group synchronization enabled.
2. Configure an OpenVPN server/instance using that LDAP backend.
3. Set a local group constraint matching a group synchronized from LDAP.
4. Try to connect with an LDAP user that has not yet been synchronized locally but is a member of the required LDAP group.

**Expected behavior**

After successful LDAP authentication and group synchronization, the OpenVPN local group constraint should evaluate the updated group membership in the same login flow.

**Observed behavior**

The first login attempt may be denied because the local group check does not see the group membership synchronized by the authenticator. A second login attempt can succeed because the synchronized group membership is then already present in the loaded config state.

**Additional context**

This affects the OpenVPN `user_pass_verify.php` authentication path.

A proposed fix is to evaluate the local group constraint against the current OPNsense config object after authentication succeeds, instead of using the legacy `getUserGroups()` helper.

**Environment**

OPNsense 26.1.7 (amd64).
Intel(R) Xeon(R) CPU E5-2643 v4

matdal5423:
LDAP authentication can update local group membership during a successful OpenVPN login. The local group constraint used the legacy getUserGroups() helper, which reads the global config state loaded before authentication.

Check membership against the current config object instead so group constraints can see memberships synchronized by the authenticator in the same login flow.

**Important notices**

- [ x ] I have read the contributing guidelines at https://github.com/opnsense/core/blob/master/CONTRIBUTING.md
- [ x ] I opened an issue first for non-trivial changes and linked it below.
- [ x ] AI tools were used to create at least part of the code submitted herewith.

- ChatGPT 5.5:
- I was aware of the issue; the AI then wrote the entire patch and the Pull Request and the Issue text (my english is quite bad) :)

---

**Describe the problem**

OpenVPN user authentication can synchronize LDAP group membership during a successful login. The local group constraint currently checks membership through the legacy `getUserGroups()` helper, which reads the global config state loaded before authentication.

As a result, group membership synchronized by the authenticator during the same login flow may not be visible until a second login attempt.

Fixes: #10255

---

Check OpenVPN local group membership against the current OPNsense config object after authentication succeeds. This lets the group constraint see memberships synchronized by LDAP/RADIUS-style authenticators in the same login flow, without changing the authentication order or the existing CSO/provisioning logic.

---

**Testing**

- `php -l src/opnsense/scripts/openvpn/user_pass_verify.php`
- `git diff --check -- src/opnsense/scripts/openvpn/user_pass_verify.php`
- `bmake lint`

`bmake lint` completed successfully with:

```text
Checked 2567 files in 5.9 seconds
No syntax error found

### DIFF
--- a/src/opnsense/scripts/openvpn/user_pass_verify.php
+++ b/src/opnsense/scripts/openvpn/user_pass_verify.php
@@ -53,6 +53,44 @@ function parse_auth_properties($props)
 }
 
 /**
+ * Check local group membership using the current configuration state
+ * @param string $username username to find
+ * @param string $groupname group name to test
+ * @return bool user is a member of the requested group
+ */
+function user_in_local_group($username, $groupname)
+{
+    $cnf = OPNsense\Core\Config::getInstance()->object();
+    $uid = null;
+
+    if (isset($cnf->system->user)) {
+        foreach ($cnf->system->user as $user) {
+            if ((string)$user->name == $username) {
+                $uid = (string)$user->uid;
+                break;
+            }
+        }
+    }
+
+    if ($uid === null || !isset($cnf->system->group)) {
+        return false;
+    }
+
+    foreach ($cnf->system->group as $group) {
+        if ((string)$group->name == $groupname) {
+            foreach ($group->member as $member) {
+                if (in_array($uid, explode(',', (string)$member))) {
+                    return true;
+                }
+            }
+            break;
+        }
+    }
+
+    return false;
+}
+
+/**
  * perform authentication
  * @param string $common_name certificate common name for this connection
  * @param string $serverid server identifier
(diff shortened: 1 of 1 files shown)


AdSchellevis:
https://github.com/opnsense/core/commit/f3b944d33480a5a2bcdc3b7c7d1f9c983e4b235a

PR #10264, opened 2026-05-06
firmware: ability to shutdown after upgrade

### LINKED ISSUE #10245: [FR] Add option to update and shutdown instead of restart when a restart is needed
**Important notices**

Sometimes, it's nice to schedule physical maintenance at the same time as update that require shutdowns, since there will be a disruption anyhow. However, at the moment, if we need to shut down the system to swap out hardware, and there's an update pending, we need to shut it down twice, which feels a bit redundant. It would be nice to have the option to just shutdown after the update.

When the system prepares for the restart, there's currently a popup informing us that there will be a restart. That popup could instead ask if we want to restart or shutdown the system. Alternatively, a toggle on the update page next to "This update requires a reboot" text that lets us choose shutdown instead.

The default should remain restart, to permit unattended/start and forget updates. The popup can incorporate this by giving a timeout for the question, and choosing restart once the timeout is up.

sopex:
**Important notices**

**Details**

This PR uses the dialog box that lets a user know that the upgrade will initiate a restart, to present the user with the ability to power off instead. If no choice is made after 30 seconds, it automatically restarts. 

Maybe the countdown should be removed and just wait for an answer, but it seems like ample time, and the dialog box shows up immediately after a user action, so no chance to be forgotten.

---

**Related issue**

Closes: https://github.com/opnsense/core/issues/10245

### DIFF
--- a/src/opnsense/mvc/app/controllers/OPNsense/Core/Api/FirmwareController.php
+++ b/src/opnsense/mvc/app/controllers/OPNsense/Core/Api/FirmwareController.php
@@ -323,7 +323,7 @@ class FirmwareController extends ApiMutableModelControllerBase
                     $response['status_msg'] = sprintf(
                         '%s %s',
                         $response['status_msg'],
-                        gettext('This update requires a reboot.')
+                        gettext('This update requires a reboot/ power off.')
                     );
                 }
                 $response['status_reboot'] = $active_reboot;
--- a/src/opnsense/mvc/app/controllers/OPNsense/Core/Api/FirmwareController.php
+++ b/src/opnsense/mvc/app/controllers/OPNsense/Core/Api/FirmwareController.php
@@ -769,6 +769,8 @@ class FirmwareController extends ApiMutableModelControllerBase
             $result['status'] = 'done';
         } elseif (strpos($cmd_result, '***REBOOT***') !== false) {
             $result['status'] = 'reboot';
+        } elseif (strpos($cmd_result, '***SHUTDOWN***') !== false) {
+            $result['status'] = 'shutdown';
         }
 
         return $result;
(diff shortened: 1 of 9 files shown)


fichtner:
I was thinking something similar but without a persistent option using a checkbox in the reboot dialog. Probably going to write this from scratch while testing which is in this case easier than doing review cycles.

sopex:
> I was thinking something similar but without a persistent option using a checkbox in the reboot dialog. Probably going to write this from scratch while testing which is in this case easier than doing review cycles.

Sure, rewrite it but this is not persistent and in the dialog box :) 
Review, what is that? My code needs review? :)
In all seriousness, this works as is. I did multiple updates and worked as expected.

fichtner:
well it is persistent: `<shutdown type="BooleanField"/>`

it would be more flexible to have this as an optional parameter for update/upgrade actions so it can be used for API-based maintenance too

sopex:
> well it is persistent: `<shutdown type="BooleanField"/>`
> 
> it would be more flexible to have this as an optional parameter for update/upgrade actions so it can be used for API-based maintenance too

Yeah ok. It flips back immediately after being used. It's a persistent variable, not a persistent option.

src/opnsense/mvc/app/controllers/OPNsense/Core/Api/FirmwareController.php
fichtner: bascially you add the parameter here and you don't need state

src/opnsense/mvc/app/controllers/OPNsense/Core/Api/FirmwareController.php
```diff
@@ -568,6 +572,8 @@ public function auditAction()
      */
     public function reinstallAction($pkg_name)
     {
+        if ($this->request->isPost() && $this->request->getPost('shutdown') === '1') {
```
fichtner: only makes sense for upgrade/update

fichtner:
As a general note the Power menu has "Power Off" and we should follow that terminology.

sopex:
> As a general note the Power menu has "Power Off" and we should follow that terminology.

It seemed to me that only user facing power off is used, so I just did the things the user sees and kept shutdown for the variables.

Issue #9355, opened 2025-11-04
Captive Portal: performance issue


Tawmu:
**Important notices**

Before you add a new report, we ask you kindly to acknowledge the following:

- [X] I have read the contributing guide lines at https://github.com/opnsense/core/blob/master/CONTRIBUTING.md
- [X] I am convinced that my issue is new after having checked both open and closed issues at https://github.com/opnsense/core/issues?q=is%3Aissue

**Describe the bug**

We run LAN gaming events and two events in a row now (July/October) we've seen OPNsense begin to hang and drop packets once our captive portal entries hit somewhere around 650 - 700 user entries. The issue is resolved when we are eventually able to login to the Web UI and disable the captive portal.

Symptoms are 100% system CPU usage and 5-15% packet loss, seemingly on processes kernel{if_io_tqg_0}, kernel{if_io_tqg_0}, kernel{if_io_tqg_2}, and kernel{if_io_tqg_3}. Roughly 650-700 captive portal entries and 2.5Gbps of traffic flowing between SFP+ interfaces.

The Captive Portal on OPNsense simply serves a HTML redirect page - we handle authentication in an external app that makes an API call to OPNsense to add the user's session in.

**To Reproduce**

Run many captive portal sessions with good amounts of traffic.

**Expected behavior**

OPNsense continues to route efficiently as more captive portal entries are created.

**Describe alternatives you considered**

We temporarily disabled captive portal and performance returned to normal.

**Environment**

Software version used and hardware type if relevant, e.g.:

OPNsense 25.7 (amd64).
Intel® Xeon™ E-2314
Network Intel® X520-DA2

Tawmu:
Just trying to investigate this myself and it appears that there was a migration from IPFW to PF in a semi recent release earlier this year - we've been using the OPNsense captive portal in this way for several years without issue and the first time we noticed this performance issue was at our July event (where the last time it would have performed properly is at our February event)

construct4261:
We have the exact same issue with a similar number of connections, and have not found a solution. We tried switching out the portal we built for the default but it had no impact.

swhite2:
While touching the subject, I took the liberty of reproducing this with ~500 virtualized clients. There seems to be bottleneck in the way ethernet rules are handled in pf (`pf_test_eth()`). Since these rules aren't part of any table, simple linked-list traversal is used for every relevant packet, which scales up execution time with the amount of clients connected to the portal.

There are options to address this, but all of them have downsides as well.

[...]

PR #10275, opened 2026-05-08
Captive Portal: re-introduce hash lookup for accounting purposes

swhite2:
**Important notices**

- Model used:
- Extent of AI involvement:

---

**Describe the problem**

Table redirection allowed for constant time lookups, with the migration to pf this was changed to a linear time lookup.

While here, fix a small edge case that kills states for ips flipping primary IPs according to hostwatch. Also make sure to include the set of ipfw keys to "registered addresses" to make sure theyre properly cleaned up from the table

---

Reuse table lookups.

The `add_accounting` logic now expects the full set of IP addresses associated with a single client to make sure they use the same rulenumber, this allows us to prevent wasting space in the ruleset. `del_accounting` is unaffected.

---

**Related issue**

https://github.com/opnsense/core/issues/9355

### DIFF
--- a/src/opnsense/scripts/captiveportal/lib/ipfw.py
+++ b/src/opnsense/scripts/captiveportal/lib/ipfw.py
@@ -101,14 +101,8 @@ class IPFW(object):
         ipfw_tbl = IPFW.list_table(table_number)
         acc_info = IPFW.list_accounting_info()
 
-        a_present = set()
-        a_missing = set()
-        for address in addresses:
-            if address not in acc_info:
-                a_missing.add(address)
-            else:
-                a_present.add(address)
-
+        a_present = set(addresses) & acc_info.keys()
+        a_missing = set(addresses) - acc_info.keys()
         present_rules = [(acc_info[address]['rule'], address) for address in sorted(a_present)]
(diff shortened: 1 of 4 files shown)


src/opnsense/scripts/captiveportal/lib/ipfw.py
```diff
@@ -1,5 +1,6 @@
 """
-    Copyright (c) 2025 Deciso B.V.
+    Copyright (c) 2025-2026 Deciso B.V.
+    Copyright (c) 2015-2019 Ad Schellevis <ad@opnsense.org>
```
fichtner, line 3: just for clarity: what's going on here?

swhite2: file (https://github.com/opnsense/core/blob/stable/24.7/src/opnsense/scripts/OPNsense/CaptivePortal/lib/ipfw.py#L2) got purged during transition to pf, but brought back for accounting purposes. Included the original copyright here

fichtner: ah that's why git showed no history. thanks 👍

AdSchellevis (approved):

Issue #6527, opened 2023-04-26
Feature Request:  Option to specify the Y axis value maximum on the Reporting\Traffic Graphs


NGD-Skier:
**Before you add a new report, we ask you kindly to acknowledge the following:**

- [X] I have read the contributing guide lines at https://github.com/opnsense/core/blob/master/CONTRIBUTING.md
- [X] I am convinced that my issue is new after having checked both open and closed issues at https://github.com/opnsense/core/issues?q=is%3Aissue

**Is your feature request related to a problem? Please describe.**
Problem?  Depends on how you define a problem...  Could also be considered a preference, but if I were to express it as a problem, I'd say it like this:  When I look at "Reporting\Traffic Graph," I see a beautiful graph going up and down while scrolling from right to left.  I **might** immediately think "wow, my bandwidth is pegged" or maybe (less common) "oh, not much traffic at the moment..."  But I might not be right...  (e.g. maybe the graph is near the top, but the scale is only 1Mb, whereas I have 25Mb).  To know what is actually going on, I have to take a look at the legend to see what the dynamic graph's Y axis scale is currently set to, so I can actually get meaning from the graph.

**A clear and concise description of what the problem is including your motivation for the request,**
For the purpose of determining whether any given slowness I'm experiencing is related to my bandwidth being at full capacity (or not), I am missing a solution that will allow me to instantly recognize a pegged interface."
Problem Summary:  I can't just glance at the traffic graph and instantly know how much my WAN (or other) interface is used.

**Describe the solution you like**
**A clear and concise description of what you want to happen.**
I would like two Radio Buttons (or check boxes) on the `/ui/Reporting/Traffic` page, labelled "Auto Max Rate In" and "Auto Max Rate Out" which when set to the Auto position, would maintain current behavior of the traffic graphs, but which when NOT set to Auto, would enable two text fields that accept numbers (e.g. bps or kbps), specifying the maximum value of the Y axis, such that the graph becomes a visual representation of just how full "my pipe" is ("pipe," in the general sense, not the "traffic shaper" sense).  i.e. When I specify a Max Rate (for either or both In and Out graphs), the Y axis will never change, so if I look at the graph and see all blue or all orange near the top of the graph, I'll instantly know that someone (probably my youngest daughter) is sucking up all our bandwidth.

**Describe alternatives you considered**
**A clear and concise description of any alternative solutions or features you considered.**
[...]

PR #10277, opened 2026-05-08
Option to Disable dynamic bps scale and set a static bps value on the Traffic Report page

NGD-Skier:
The intent of this branch is to allow a user to set a static bps value on the Traffic Report page, so they can, for example, set the max upload and max download rates, and have the traffic graph just use that for scale, so at a glance, they can see a visual representation of how much of their available bandwidth they're using.

This changes the default behavior where the graph sometimes suggests that bandwidth is maxed out, even when very little bandwidth is being used, as the graph auto-scales its vertical axis down to much smaller values when bandwidth usage is low.  This original behavior is retained when the "Auto Max In" and "Auto Max Out" check boxes are check, but when you uncheck them, you can specify both your max in/out scale (convenient for people with asymmetric connection speeds).

i.e. With the new, optional behavior, if you look and see that the graph is maxed out, then you know your bandwidth is being fully used, if the graph shows traffic at half the height of the graph, then half your bandwidth is being used, and if you're using minimal to no bandwidth, the graph will be near the bottom instead of auto-scaling and "zooming" in, making it look like your connection is experiencing high usage when its not.

The only files changed to enable this additional functionality is traffic.volt.  Nothing else has been changed.

This is related to the discussion of issue #6527 found at https://github.com/opnsense/core/issues/6527

Thanks to user AdSchellevis for suggesting where this change would be made.

Before submitting this PR, I fully updated my local OpnSense system to the latest version (as of today, May 8th, 2026), rebooted, and then tested the modified file to verify that both original functionality and updated functionality work correctly.

  -Mike

**Important notices**

- [X] AI tools were used to create at least part of the code submitted herewith.

- Model used:  GPT 5.5
- Extent of AI involvement:  Highly involved.  I only reviewed the code changes via a file diff using Beyond Compare.

---

**Describe the problem**

The dynamic Y-axis in the "Reporting:  Traffic" page makes utilization hard to judge at a glance, and this adds optional fixed max rates for In/Out, so at a glance you can tell roughly how much of your bandwidth capacity it being used (instead of seeing a graph that looks maxed out, but then looking at the auto-scaled Y axis and realizing the scale is 1Mb, and you're barely using any of your 100Mb connection's potential).

---

It changes the file "traffic.volt," to affect the preferred behavior mentioned above.  The original behavior is retained when the "Auto Max In" and "Auto Max Out" checkboxes are checked.

---

**Related issue**

https://github.com/opnsense/core/issues/6527

### DIFF
--- a/src/opnsense/mvc/app/views/OPNsense/Diagnostics/traffic.volt
+++ b/src/opnsense/mvc/app/views/OPNsense/Diagnostics/traffic.volt
@@ -67,6 +67,107 @@
             }
         }
 
+        function parse_rate(value, unit) {
+            let matches = (value || '').toString().trim().match(/^([0-9]*\.?[0-9]+)\s*([kmg]?)(?:b(?:ps)?)?$/i);
+            if (matches === null) {
+                return null;
+            }
+
+            let suffix = matches[2].toLowerCase();
+            if (suffix === '') {
+                suffix = (unit || 'bps').replace(/bps$/i, '').toLowerCase();
+            }
+
+            let multipliers = {
+                '': 1,
+                'k': 1000,
+                'm': 1000 * 1000,
+                'g': 1000 * 1000 * 1000
+            };
+
+            return Math.round(parseFloat(matches[1]) * multipliers[suffix]);
+        }
+
+        function toggle_rate_controls(direction) {
+            let is_auto = $("#traffic_scale_auto_" + direction).prop("checked");
+            $("#traffic_scale_max_" + direction).prop("disabled", is_auto);
+            $("#traffic_scale_unit_" + direction).prop("disabled", is_auto).selectpicker('refresh');
+            if (is_auto) {
+                $("#traffic_scale_max_" + direction).closest(".traffic-scale-group").removeClass("has-error");
+            }
+        }
+
+        function load_rate_controls() {
+            ['in', 'out'].forEach(function(direction) {
+                let auto_value = window.localStorage ? window.localStorage.getItem("api.diagnostics.traffic.max_rate." + direction + ".auto") : null;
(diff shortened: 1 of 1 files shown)


NGD-Skier:
Here's an example of what the Traffic Graph looks like with this update:

image

NGD-Skier:
And here an example of the original behavior (which is what you get when you check the Auto Max In/Out boxes), showing what I mean by the auto-scaling making it look like maybe you're hitting peak bandwidth, with a spike that rises near the top of the graph, but when you look at the scale on the left (Y Axis), you see you're only using ~2.5Mb IN of the 40Mb possible on this connection.

image

NGD-Skier:
... but when you are using the static values that represent your In/Out bandwidth capacities, you get a graph that instantly/visually shows you you're not using much bandwidth.  In this picture, I've set the manual Max In/Out values, to roughly the max bandwidth on my connection, so when my bandwidth usage has spikes to only ~2.5Mbps, the graph represent that as low bandwidth usage (in contrast with the previous picture with 2.5Mbps spiking the graph near the top.

image

NGD-Skier:
And here's showing just the WAN interface during a Speed Test via SpeedTest.net:

image

I wasn't able to capture the entire speedtest, which start with download and then does upload, so I captured the screenshot from the start of the download test, through the end of the download test, and then just the start of the upload speedtest before the start of the download speedtest went off the screen.

I adjusted my Max In/Out, so I could see a little more of the spikes before the throttling of my ISP kicks in and bandwidth usage stabilizes.

AdSchellevis:
@NGD-Skier https://github.com/opnsense/core/commit/6d94603bb92dc6fb25ed9038230b53bacad153b8 should do roughly what you are looking for, also explains a bit why I'm not a huge fan of AI generated code these days. There's one thing AI is very good in and that is collecting technical debt I'm afraid....

Issue #10281, opened 2026-05-09
Unbound DNS Blocklist does not trust user-defined Certificate Authorities


thereiam:
# Unbound DNS Blocklist does not trust user-defined Certificate Authorities

## Describe the bug

Unbound DNS blocklist downloads fail when the blocklist source is hosted on an internal HTTPS server using a certificate signed by a custom/user-defined Certificate Authority (CA), even when that CA has been added to OPNsense under the trust store.

The Unbound blocklist downloader appears to ignore the configured trusted CAs and only trusts the system/default CA bundle.

## To Reproduce

1. Create an internal HTTPS server using a TLS certificate signed by a custom/internal CA.
2. Add the custom CA to OPNsense:

   * `System -> Trust -> Authorities`
3. Configure an Unbound DNS blocklist:

   * `Services -> Unbound DNS -> Blocklist`
4. Enable advanced mode and add a blocklist URL pointing to the internal HTTPS server.
5. Save and apply changes.
6. Trigger a blocklist update/reload.

Example blocklist URL:

```text
https://internal-git.example.local/blocklists/ads
```

## Expected behavior

Unbound should trust certificates issued by user-defined CAs that are configured in the OPNsense trust store, and the blocklist should download successfully.

## Actual behavior

The download fails with an SSL certificate verification error indicating that the local issuer certificate cannot be found.

## Relevant log output

```text
Error    unbound    unable to download blocklist from https://internal-git.example.local/blocklists/ads and no cache available

Error    unbound    blocklist download : error reading file from https://internal-git.example.local/blocklists/ads
(error : HTTPSConnectionPool(host='internal-git.example.local', port=443):
Max retries exceeded with url: /blocklists/ads
(Caused by SSLError(SSLCertVerificationError(1,
'[SSL: CERTIFICATE_VERIFY_FAILED] certificate verify failed:
unable to get local issuer certificate (_ssl.c:1032)')))
```

## Environment

* OPNsense version: OPNsense 26.1.7_3-amd64

## Additional context

The issue appears specific to the Unbound blocklist downloader component and not to the general OPNsense trust store itself, since the CA is correctly imported and visible under the Trust Authorities section.

sopex:
Give this a try
`opnsense-patch -a sopex 38dd79e7bd1a673e7569ee4f32cdf083cb6e9ebb`

thereiam:
This worked, thanks!

sopex:
> Trigger a blocklist update/reload.

[...]

PR #10283, opened 2026-05-10
unbound: blocklists apply button

sopex:
**Important notices**

**Purpose**
Renames the apply button to let the user know that a blocklist update will occur

**Related issue**
https://github.com/opnsense/core/issues/10281

### DIFF
--- a/src/opnsense/mvc/app/views/OPNsense/Unbound/dnsbl.volt
+++ b/src/opnsense/mvc/app/views/OPNsense/Unbound/dnsbl.volt
@@ -29,19 +29,21 @@
    $(document).ready(function() {
         let gridLoaded = false;
         $('a[data-toggle="tab"]').on('shown.bs.tab', function (e) {
-            $('#reconfigureAct').closest('content-box').show();
-            if (e.target.id === 'blocklists_tab' && !gridLoaded) {
-                $("#{{formGridDnsbl['table_id']}}").UIBootgrid({
-                    search:'/api/unbound/settings/searchDnsbl/',
-                    get:'/api/unbound/settings/getDnsbl/',
-                    set:'/api/unbound/settings/setDnsbl/',
-                    add:'/api/unbound/settings/addDnsbl/',
-                    del:'/api/unbound/settings/delDnsbl/',
-                    toggle:'/api/unbound/settings/toggleDnsbl/',
-                });
-                gridLoaded = true;
-            } else if (e.target.id === 'blocklist_tester_tab') {
-                $('#reconfigureAct').closest('content-box').hide();
+            if (e.target.id === 'blocklists_tab') {
+                $('#reconfigureAct').closest('.content-box').show();
+                if (!gridLoaded) {
+                    $("#{{formGridDnsbl['table_id']}}").UIBootgrid({
+                        search:'/api/unbound/settings/searchDnsbl/',
+                        get:'/api/unbound/settings/getDnsbl/',
+                        set:'/api/unbound/settings/setDnsbl/',
+                        add:'/api/unbound/settings/addDnsbl/',
+                        del:'/api/unbound/settings/delDnsbl/',
+                        toggle:'/api/unbound/settings/toggleDnsbl/',
+                    });
+                    gridLoaded = true;
+                }
+            } else {
+                $('#reconfigureAct').closest('.content-box').hide();
             }
         });
 
@@ -124,5 +126,5 @@
     </div>
 </div>
 
-{{ partial('layout_partials/base_apply_button', {'data_endpoint': '/api/unbound/service/dnsbl', 'data_label': 'Apply & Update Blocklists'}) }}
+{{ partial('layout_partials/base_apply_button', {'data_endpoint': '/api/unbound/service/dnsbl', 'data_change_message_content': 'After changing settings, please remember to apply in order to update the blocklists.'}) }}
 {{ partial("layout_partials/base_dialog",['fields':formDialogDnsbl,'id':formGridDnsbl['edit_dialog_id'],'label':lang._('Edit Blocklist')])}}
--- a/src/opnsense/mvc/app/views/OPNsense/Unbound/dnsbl.volt
+++ b/src/opnsense/mvc/app/views/OPNsense/Unbound/dnsbl.volt
@@ -124,5 +124,5 @@
     </div>
 </div>
 
-{{ partial('layout_partials/base_apply_button', {'data_endpoint': '/api/unbound/service/dnsbl'}) }}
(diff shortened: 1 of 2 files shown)


src/opnsense/scripts/unbound/blocklists/__init__.py
```diff
@@ -106,6 +106,8 @@ def _uri_reader(self, uri):
             'timeout': 30,
             'stream': True
         }
+        if os.path.exists('/usr/local/etc/ssl/cert.pem'):
+            req_opts['verify'] = '/usr/local/etc/ssl/cert.pem'
```
fichtner: what I don't get is that `REQUESTS_CA_BUNDLE` should be set:

https://github.com/opnsense/core/blob/9b93f84c24eadcafbbcd87328f8a1fab4d562467/src/opnsense/service/conf/configd.conf#L7

And this should work for other scripting already.

sopex: requests.get() seems to need a verify argument to include CAs other the ones it is bundled with.
configd doesn't seem up to the task.

AdSchellevis: that's pretty weird, I'm quite sure this used to work earlier as also specified in https://requests.readthedocs.io/en/latest/user/advanced/ 

```
This list of trusted CAs can also be specified through the REQUESTS_CA_BUNDLE environment variable. I
```

sopex: ``` When you are using the prepared request flow, keep in mind that it does not take into account the environment. This can cause problems if you are using environment variables to change the behaviour of requests. For example: Self-signed SSL certificates specified in REQUESTS_CA_BUNDLE will not be taken into account. As a result an SSL: CERTIFICATE_VERIFY_FAILED is thrown. ```

fichtner: In either case I don't like hardcoding this everywhere it is deemed necessary.

AdSchellevis: > `When you are using the prepared request flow, keep in mind that it does not take into account the environment. This can cause problems if you are using environment variables to change the behaviour of requests. For example: Self-signed SSL certificates specified in REQUESTS_CA_BUNDLE will not be taken into account. As a result an SSL: CERTIFICATE_VERIFY_FAILED is thrown.`

sure, but we're not talking about a self-signed cert, but a cert signed by a self created CA, which is basically the same as any other CA in the list, unless we're trying to create a self-signed cert which does not belong to a CA.....

just to be sure, you did try this yourself?

fichtner: Self signed certs also work when uploading them to Authorities.

AdSchellevis: I could imagine that to be some sort of difference, but indeed it isn't env and parameter should be the same according the source. Just leaving this here for reference.

https://github.com/psf/requests/blob/e8d2c015eecda8273612dd4562425e00cd164ba5/src/requests/sessions.py#L849-L853

sopex: Upon further reflection, I believe they just tried to manually update the list without engaging configd...
Sorry for the wasted time.
Maybe the button should be renamed to "Apply & Update Blocklists" similarly to how Suricata displays it.

sopex: @fichtner Is there merit to this, or should we close it?
>Maybe the button should be renamed to "Apply & Update Blocklists" similarly to how Suricata rules display it.

fichtner: In practice the best course of action is to write a library extension for all scripts to use (and/or detect config issues) but that’s too much work to review and test to make sense.

it would be nice if requests had a better awareness of (Free)BSD, but then again a conscience bug report would also lead to quick help. Not everything needs to be fixed in code. Especially if people circumvent GUI or backend command dispatch (in edge cases).

swhite2: > Maybe the button should be renamed to "Apply & Update Blocklists" similarly to how Suricata displays it.

FWIW, I'm not opposed to this or something similar.. The "Apply" on this page serves a special purpose after all.

swhite2:
@sopex Let's do this on `data_change_message_content` instead. Better to keep it consistent.

sopex:
@swhite2 Instead or on top? Consistency is one thing, but usability is also important and I am sure these messages all look the same. When you read a few, you assume that all are the same.

fichtner:
It's in the apply bar on master after saving. Don't want to argue about people. Technically the message is much better than having individual apply buttons that differ in labelling.

sopex:
> It's in the apply bar on master after saving. Don't want to argue about people. Technically the message is much better than having individual apply buttons that differ in labelling.

I have seen your bar, it's very cool :)

IMO we probably leave it as is then.

This is more confusing, and people will wonder how to update:
image
<br>
This is explanatory, which is not great for this kind of bar/ breaks unification making it a help text:
image
<br>
And this breaks the unification:
image

fichtner:
I like the first one for what it is although the message needs tweaking. In most cases the eye will catch the message as long as it's clear (and includes the phrase "updating the blocklist"). I'm not a fan of "&" as it generally breaks the reading flow in technical situations and pulls the eye away from the words left and right.

fichtner:
"After changing settings, please remember to apply in order to update the blocklists."

sopex:
Will you do me the favor of letting me fix this? ;)

image

fichtner:
Sure, I'm just here to offer grumpy advice.

sopex:
> Sure, I'm just here to offer grumpy advice.

All good, 2 months ago asked me to wait. That's why I am asking again.

sopex:
@fichtner You are not grumpy

src/opnsense/mvc/app/views/OPNsense/Unbound/dnsbl.volt
```diff
@@ -29,19 +29,21 @@
...
+                        get:'/api/unbound/settings/getDnsbl/',
+                        set:'/api/unbound/settings/setDnsbl/',
+                        add:'/api/unbound/settings/addDnsbl/',
+                        del:'/api/unbound/settings/delDnsbl/',
+                        toggle:'/api/unbound/settings/toggleDnsbl/',
+                    });
+                    gridLoaded = true;
+                }
+            } else {
+                $('#reconfigureAct').closest('.content-box').hide();
```
fichtner, line 46: @swhite2 any idea why this is/was needed?

swhite2: It attempted to hide the apply button on the tester tab, and show it on the blocklists one, but failed to do so because of the missing `.` on `content-box`. Proposed code should work.

fichtner:
Merged, thanks!

Issue #10282, opened 2026-05-10
radvd: DNSSL can't be disabled without disabling RDNSS


max-foss:
**Important notices**

Before you add a new report, we ask you kindly to acknowledge the following:

- [x] I have read the contributing guidelines at https://github.com/opnsense/core/blob/master/CONTRIBUTING.md
- [x] I am convinced that my issue is new after having checked both open and closed issues at https://github.com/opnsense/core/issues?q=is%3Aissue

**Describe the bug**

A clear and concise description of what the bug is, including the last known working version (if any).

*Tip*: to validate your setup was working with the previous version, use opnsense-revert (https://docs.opnsense.org/manual/opnsense_tools.html#opnsense-revert)

**To Reproduce**

Go to Services -> Router Advertisements.
Open an Advanced entry in Advanced Mode.

**Expected behavior**

There should be a way to disable _DNSSL_ entirely, for example by entering `-1 `into  **DNS Search List Lifetime**, without needing to uncheck **Enable DNS** also disabling _RDNSS_ which I need.

**Describe alternatives you considered**

Putting a dummy value in with a long lifetime, still has pollution in radvd.conf and every RA packet and devices may act on it in bad ways.

**Screenshots**

**Relevant log files**

If applicable, information from log files supporting your claim.

**Additional context**

It seems to be a factor in some Android devices misbehaving and losing connectivity, especially with a short lifetime.

**Environment**

Software version used and hardware type if relevant, e.g.:
OPNsense 26.1.7_3-amd64

fichtner:
System: Settings: General uses "." for DNS search domain:

> Enter additional domains to add to the local list of search domains. Use "." to disable passing any other automatic search domain for resolving.

Only giving a single dot there should disable the functionality. Do you agree we can do the same in Radvd?


Cheers,
Franco

max-foss:
> System: Settings: General uses "." for DNS search domain:
> 
> > Enter additional domains to add to the local list of search domains. Use "." to disable passing any other automatic search domain for resolving.
> 
> Only giving a single dot there should disable the functionality. Do you agree we can do the same in Radvd?
> 
> Cheers, Franco

Why not? I could also build that soon, if this your preferred shape. It should omit the DNSSL-section of that interface from radvd.conf entirely.

fichtner:
Ok, waiting for your PR then. 👍 


Thanks,
Franco

max-foss:
> Ok, waiting for your PR then. 👍
> 
[...]

PR #10284, opened 2026-05-10
radvd: allow disabling DNSSL per interface #10282

max-foss:
**Important notices**

- Models used: Opus 4.6 and GPT-5.5
- Extent of AI involvement: Prototyping and validation, not a single word on the Issue or PR text was touched by an LLM.

---

**Describe the problem**

The **DNS Search List (DNSSL)** can't be disabled without also disabling **Recursive DNS Servers (RDNSS)** which may lead to adverse effects with certain devices, I only observed weirdness with Android phones.

---

Entering `.` into the **DNS Search List (DNSSL)** field will omit the relevant section for that interface in radvd.conf entirely per https://github.com/opnsense/core/issues/10282#issuecomment-4414880741.

---

**Related issue**

#10282

---

I already tried the patch for a brief period in isolation, it works as expected.

### DIFF
--- a/src/opnsense/mvc/app/models/OPNsense/Radvd/Radvd.xml
+++ b/src/opnsense/mvc/app/models/OPNsense/Radvd/Radvd.xml
@@ -60,11 +60,7 @@
                 <AddressFamily>ipv6</AddressFamily>
                 <AsList>Y</AsList>
             </RDNSS>
-            <DNSSL type="HostnameField">
-                <IsDNSName>Y</IsDNSName>
-                <IpAllowed>N</IpAllowed>
-                <AsList>Y</AsList>
-            </DNSSL>
+            <DNSSL type="CSVListField"/>
             <dns type="BooleanField">
                 <Required>Y</Required>
                 <Default>1</Default>
--- a/src/etc/inc/plugins.inc.d/radvd.inc
+++ b/src/etc/inc/plugins.inc.d/radvd.inc
@@ -340,7 +340,9 @@ function radvd_configure_do($verbose = false, $ignorelist = [])
                 $rdnss[] = $server;
             }
 
-            if (count($searchlist_tmp)) {
+            if (count($searchlist_tmp) == 1 && $searchlist_tmp[0] == '.') {
+                /* explicitly disabled */
+            } elseif (count($searchlist_tmp)) {
                 $dnssl = $searchlist_tmp;
             } elseif (!empty($config['system']['domain'])) {
                 $dnssl = [$config['system']['domain']];
--- a/src/opnsense/mvc/app/controllers/OPNsense/Radvd/forms/dialogEntry.xml
+++ b/src/opnsense/mvc/app/controllers/OPNsense/Radvd/forms/dialogEntry.xml
@@ -192,7 +192,7 @@
         <type>select_multiple</type>
         <style>tokenize</style>
         <allownew>true</allownew>
-        <help>The default is to use the domain name of this system as the DNSSL option. You may specify explicit domains here instead.</help>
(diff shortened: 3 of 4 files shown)


src/opnsense/mvc/app/models/OPNsense/Radvd/Radvd.xml
```diff
@@ -60,11 +60,7 @@
                 <AddressFamily>ipv6</AddressFamily>
                 <AsList>Y</AsList>
             </RDNSS>
-            <DNSSL type="HostnameField">
-                <IsDNSName>Y</IsDNSName>
-                <IpAllowed>N</IpAllowed>
-                <AsList>Y</AsList>
-            </DNSSL>
+            <DNSSL type="CSVListField"/>
```
fichtner, line 63: I see your issue and we probably shouldn't demote this to CSVListField which is more of a last resort for rapid implementation than it is good for integrity especially replacing it it blatantly like this. Would be nicer if `IsDNSName` recognises "." as a valid input. In theory it's a valid DNS name ... or make room for optionally allowing it as well.

I'm happy to do this PR. It's difficult to steer people steering AI.

Monviech: IsDNSName has been made more strict a few times if I remember correctly.

Would be better to create a custom hostnamefield like dnsmasq has for an issue like this.

fichtner: Yep, fair point.

max-foss: > Would be nicer if IsDNSName recognises "." as a valid input.

Well, to my knowledge **IsDNSName** is supposed to be based on the implementation of **FILTER_VALIDATE_DOMAIN** which does not consider `.` a valid domain:

```
% php -r "var_dump(filter_var('.', FILTER_VALIDATE_DOMAIN));"
bool(false)`
```

> It's difficult to steer people steering AI.

I'm not a vibe coder, in the before times I would have used Stack Overflow, PHP documentation and copy and paste of some existing code snippet making it work.

It was my own opinion that treating this as a **CSVListField** was the cleanest way to build it as-in the most local scope, least LOC touched and no surprising side-effects as I tend to prefer for projects I am not intricately familiar with.

Monviech: @fichtner wouldnt a simple boolean field do the job here as well? Imo its better than magic values, but its your choice here. Just wanted to give my two cents xD

fichtner: I have to say for the record if you think you didn’t degrade the code you definitely did. We value review from peers and we are often quite picky amongst ourselves as well which produces better quality code overall. It’s our process and I want you to respect it and avoid defending yourself for things you couldn’t know.

max-foss: > I have to say for the record if you think you didn’t degrade the code you definitely did. We value review from peers and we are often quite picky amongst ourselves as well which produces better quality code overall. It’s our process and I want you to respect it and avoid defending yourself for things you couldn’t know.

Would you like to take over from here or give me a hint on how you prefer it being done and I will do it 100% by hand.

My point regarding **FILTER_VALIDATE_DOMAIN** stands, I would avoid adding an unexpected layer on top just for this behaviour.

My original preferred shape was setting **DNS Search List Lifetime** to _-1_ because it is pretty minimal and less of a validation nightmare.

Thank you for your efforts.

fichtner:
Looking at this again I don't think we should proceed with this. I still don't oppose the idea, but I don't see the necessary human effort being put in.

PR #10290, opened 2026-05-11
Captive Portal: adjust redirection for modern portal support

swhite2:
**Important notices**

- Model used:
- Extent of AI involvement:

---

**Describe the problem**

Currently, the pf rules managing portal access force redirection for unauthenticated clients. There are however two issues related to the modern portal support documented here: https://docs.opnsense.org/manual/captiveportal.html#modern-portal-support, and only specifically when the administrator has set `https://<firewall_fqdn>/api/captiveportal/access/api`.

1. Clients connecting to `https://<firewall_fqdn>/api/captiveportal/access/api` will set the `Host` header to <firewall_fqdn>. If this doesn't contain `8000 + zoneid` as a manual header, the redirect rule still applies and the client gets a portal login page instead of a response to the `api` endpoint. This can be solved by exluding the redirect match in the lighttpd config:

```
$HTTP["url"] !~ "^/api/captiveportal/access/api$"
```

Which means that redirection for requests going to `/api/captiveportal/access/api` *exactly* must always be skipped, so that the request can be properly proxied as configured further down in the lighttpd configuration.

2. Once a client becomes authenticated, no HTTP(S) requests to the firewall are redirected anymore, meaning the client cannot access `/api/captiveportal/access/api`, as it's now sent to the normal OPNsense WebGUI. It will receive a response, but the endpoint cannot determine `zoneid`, nor `host`, since the request wasn't proxied. This response is therefore invalid.

---

Possible solutions:

1. Restricting all authenticated clients to the proxy as well, which is what this PR proposes (but only for discussion purposes). This means no client in a portal network can access the normal GUI anymore.
2. Create a special clause in the OPNsense WebGUI, redirecting requests to `/api/captiveportal/access/api` to localhost.
3. Do nothing and instead document that administrators should always specify the `8000 + zoneid` port. This has the drawback that it could break existing setups, though it is unlikely this would have worked *after* authentication to begin with.

---

**Related issue**

### DIFF
--- a/src/etc/inc/plugins.inc.d/captiveportal.inc
+++ b/src/etc/inc/plugins.inc.d/captiveportal.inc
@@ -156,44 +156,6 @@ function captiveportal_firewall($fw)
                         ]
                     );
 
-                    // Restrict authenticated clients to proxy only (IPv4)
-                    $fw->registerForwardRule(
-                        2,
-                        [
-                            'interface' => $intf,
-                            'nordr' => false,
-                            'ipprotocol' => 'inet',
-                            'protocol' => 'tcp',
-                            'from' => "<__captiveportal_zone_{$zoneid}>",
-                            'to' => '(self)',
-                            'to_port' => $to_port,
-                            'target' => '127.0.0.1',
-                            'localport' => $rdr_port,
-                            'natreflection' => 'disable',
-                            'descr' => "Redirect to Captive Portal (zone {$zoneid})",
-                            '#ref' => "ui/captiveportal#edit={$uuid}"
-                        ]
-                    );
-
-                    // Restrict authenticated clients to proxy only (IPv6)
-                    $fw->registerForwardRule(
-                        2,
-                        [
-                            'interface' => $intf,
-                            'nordr' => false,
-                            'ipprotocol' => 'inet6',
-                            'protocol' => 'tcp',
-                            'from' => "<__captiveportal_zone_{$zoneid}>",
-                            'to' => '(self)',
-                            'to_port' => $to_port,
-                            'target' => $intf . 'ip',
-                            'localport' => $rdr_port,
-                            'natreflection' => 'disable',
-                            'descr' => "Redirect to Captive Portal (zone {$zoneid})",
-                            '#ref' => "ui/captiveportal#edit={$uuid}"
-                        ]
-                    );
-
                     $fw->registerFilterRule(
                         2,
                         [
--- a/src/etc/inc/plugins.inc.d/captiveportal.inc
+++ b/src/etc/inc/plugins.inc.d/captiveportal.inc
@@ -107,51 +107,92 @@ function captiveportal_firewall($fw)
                     ]
                 );
 
-                // forward to localhost if not authenticated (IPv4)
-                $fw->registerForwardRule(
-                    2,
-                    [
-                        'interface' => $intf,
-                        'nordr' => false,
-                        'ipprotocol' => 'inet',
-                        'protocol' => 'tcp',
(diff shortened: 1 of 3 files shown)


src/opnsense/service/templates/OPNsense/Captiveportal/lighttpd-zone.conf
```diff
@@ -85,16 +85,20 @@ server.port              = {{ cp_zone_item.zoneid|int + 8000  }}
 url.redirect-code        = 302
 
 $HTTP["host"] !~ "(.*{{ cp_zone_item.redirect_host_match }}.*)" {
-	$HTTP["host"] =~ "([^:/]+)" {
-		url.redirect = ( "^(.*)$" => "{{ cp_zone_item.redirect_host }}?redirurl=%1$1")
-	}
+    $HTTP["url"] !~ "^/api/captiveportal/access/api$" {
```
AdSchellevis:
```suggestion
    $HTTP["url"] !~ "^/api/captiveportal/access" {
```

I think we can simply align this with `proxy.server = ....` as we want to prevent these from entering our default redirect.

src/opnsense/service/templates/OPNsense/Captiveportal/lighttpd-zone.conf
```diff
@@ -85,16 +85,20 @@ server.port              = {{ cp_zone_item.zoneid|int + 8000  }}
...
+    }
 }
 
 $SERVER["socket"] == "[::]:{{ cp_zone_item.zoneid|int + 8000 }}" {
-	$HTTP["host"] !~ "(.*{{cp_zone_item.redirect_host_match}}.*)" {
-		$HTTP["host"] =~ "(\[[^\]]+\]|[^:/]+)" {
-			url.redirect = ( "^(.*)$" => "{{cp_zone_item.redirect_host}}?redirurl=%1$1")
-		}
+    $HTTP["url"] !~ "^/api/captiveportal/access/api$" {
+        $HTTP["host"] !~ "(.*{{cp_zone_item.redirect_host_match}}.*)" {
```
AdSchellevis, line 97: same here

src/etc/inc/plugins.inc.d/captiveportal.inc
```diff
@@ -107,52 +107,93 @@ function captiveportal_firewall($fw)
...
+                            'to_port' => $to_port,
+                            'target' => $intf . 'ip',
+                            'localport' => $rdr_port,
+                            'natreflection' => 'disable',
+                            'descr' => "Redirect to Captive Portal (zone {$zoneid})",
+                            '#ref' => "ui/captiveportal#edit={$uuid}"
+                        ]
+                    );
+
+                    // Restrict authenticated clients to proxy only (IPv4)
```
AdSchellevis: let's leave this out and document behavior for authenticated clients.

src/etc/inc/plugins.inc.d/captiveportal.inc
```diff
@@ -107,52 +107,93 @@ function captiveportal_firewall($fw)
...
+                            'target' => '127.0.0.1',
+                            'localport' => $rdr_port,
+                            'natreflection' => 'disable',
+                            'descr' => "Redirect to Captive Portal (zone {$zoneid})",
+                            '#ref' => "ui/captiveportal#edit={$uuid}"
+                        ]
+                    );
+
+                    // Restrict authenticated clients to proxy only (IPv6)
+                    $fw->registerForwardRule(
```
AdSchellevis: same

AdSchellevis (changes requested):

swhite2:
Docs staged in https://github.com/opnsense/docs/pull/876

PR #10297, opened 2026-05-13
Services: Kea DHCPv4/6: When serializing the config to json, use JSON_UNESCAPED_UNICODE

### LINKED ISSUE #10296: Kea DHCPv4 fails to start when reservation hostname/description contains Unicode bidi characters (e.g. U+200E)
**Important notices**

**Describe the bug**

When a Kea DHCPv4 reservation's hostname or description field contains a Unicode bidi/format control character, OPNsense renders that character into `/usr/local/etc/kea/kea-dhcp4.conf` as a JSON `\uXXXX` escape. Kea's JSON parser rejects `\u` escapes for control/format code points and refuses to load the configuration, so `kea-dhcp4` will not start. The offending character (U+200E LRM, in my case) is invisible in the GUI, and Apply completes without any error indication while DHCP silently stays down. `configctl kea start` returns silently with no stderr surfaced to the user.

Last known working version: **26.1.7_3**. The defect is not specific to 26.1.8; any prior version with the same data in `config.xml` would fail to start Kea. It surfaced for me on the upgrade to 26.1.8 because Kea was restarted as part of the upgrade.

`opnsense-revert` was not run. The defect lives in data persisted in `config.xml`, not in code, so reverting versions would not change the outcome — Kea would fail to start with the same `\u200e` escape on any version.

**To Reproduce**

1. Go to Services → KEA DHCP → KEA DHCPv4 → Reservations
2. Add or edit a reservation. In the Description (or Hostname) field, paste a string containing a hidden Unicode bidi character. Retail product names from Amazon and manufacturer catalogs frequently embed these. Example value: `Aqara Hub M2 (‎HM2-G01 8782)` — there is a U+200E LRM between the `(` and `H`, invisible in most renderings. Confirm the bytes are present:

   ```
   # echo 'Aqara Hub M2 (‎HM2-G01 8782)' | hexdump -C
   00000000  41 71 61 72 61 20 48 75  62 20 4d 32 20 28 e2 80
   00000010  8e 48 4d 32 2d 47 30 31  20 38 37 38 32 29 0a
   ```

   The `e2 80 8e` sequence at offset 0x0E is the UTF-8 encoding of U+200E.
3. Save, then Apply.
4. Kea DHCPv4 fails to start. The GUI shows no error; `configctl kea start` returns silently. On the CLI:

   ```
   # kea-dhcp4 -t /usr/local/etc/kea/kea-dhcp4.conf
   Syntax check failed with: /usr/local/etc/kea/kea-dhcp4.conf:1146.44-78
   (near 17): Unsupported unicode escape in "Aqara Hub M2 (\u200eHM2-G01 8782)"
   ```

**Expected behavior**

Apply should not be able to leave Kea in a non-starting state without surfacing an error. Preferred outcomes, in order:

1. The Kea model validates reservation fields and rejects Unicode bidi/format control characters (at minimum U+200E, U+200F, U+202A–U+202E, U+2066–U+2069, U+FEFF) at save time, with a clear error pointing at the offending field and code point.
2. Failing that, the Kea template render strips those code points before writing `kea-dhcp4.conf`.
3. As a fallback, surface `kea-dhcp4`'s syntax-check stderr to the GUI when Apply fails to bring the service up, so users have something actionable instead of a silent failure.

Workaround: delete the offending field value and retype it by hand (not paste). After that, `kea-dhcp4 -t` validates cleanly and Kea starts. This requires the user to know that `kea-dhcp4 -t` exists and to inspect its output, since the GUI gives no signal that an invisible character is present.

To identify any other occurrences in `config.xml`:

```sh
sh
python3 << 'EOF'
import re
with open("/conf/config.xml", encoding="utf-8", errors="replace") as f:
    for i, line in enumerate(f, 1):
        for m in re.finditer(r"[\u200e\u200f\u202a-\u202e\u2066-\u2069\ufeff]", line):
            print(f"line {i}, col {m.start()+1}: U+{ord(m.group()):04X} in: {line.strip()[:120]}")
EOF
exit
```

(BSD grep lacks `-P`, hence Python; the `sh` wrapper is for users on the default tcsh root shell.)

In my case the scan returned exactly one hit — the single Aqara reservation description. One pasted product name was sufficient to prevent DHCP from starting.

**Screenshots**

N/A — the character is invisible in the GUI, which is part of the problem.

**Relevant log files**

`kea-dhcp4 -t` output above. `/var/log/dhcpd/latest.log` showed no entries from the failed start attempts in my case (the most recent entries were from a prior ISC-DHCP run). `configctl kea start` produced no output on stdout or stderr. The validation failure appears to occur before Kea initializes its logging subsystem, so users have no in-product signal of what went wrong.

**Additional context**

U+200E (Left-to-Right Mark) is routinely embedded in product names on retail sites (Amazon, manufacturer catalogs, Aliexpress) to help mixed-script text render correctly across locales. It is invisible in nearly every UI. Users copy-pasting device names into reservation descriptions — a common workflow for IoT-heavy networks — will hit this with no way to see what they did wrong.

This sits in a known problematic class: invisible/format characters in admin-editable fields. Beyond the immediate DHCP outage, the same input path is worth considering for general robustness (e.g. RTL override characters in display fields, BOMs from copy-paste).

Suggested remediation, narrowest to broadest:

1. Sanitize at the Kea template render step: strip Unicode bidi/format control code points before emitting the JSON. This contains the fix to the component that is actually breaking and avoids side effects on other services that may legitimately accept these characters in user-facing text.
2. Add a Kea-specific model validator that rejects those code points on save, with a field-level error message naming the offending character.
3. Independently of the above, capture and surface `kea-dhcp4 -t` stderr in the configd Apply path so any future Kea config error is visible in the GUI instead of failing silently.

**Environment**

- OPNsense 26.1.8 (amd64), upgraded from 26.1.7_3
- FreeBSD 14.3-RELEASE-p12
- Kea 3.0.3
- CPU: Intel Xeon E5-1650 (Sandy Bridge-E, 6C/12T, 3.20 GHz)
- Chipset: Intel C600/X79
- NICs: 2× Intel 82599ES 10 GbE SFI/SFP+ (`ix0`, `ix1`), 4× Intel I350 1 GbE (`igb0`–`igb3`)

Monviech:
**Important notices**

- Model used:
- Extent of AI involvement:

---

**Describe the problem**

KEA uses an input validation that checks the configuration for escaped unicode characters starting with 00, these are considered valid Any other escaped unicode character will refuse the whole config to be loaded.

Ref: https://github.com/isc-projects/kea/blob/476ff110e90b3412fcd1d089b276e86709df5c7f/src/lib/cc/data.cc#L471-L484

---

Using JSON_UNESCAPED_UNICODE will print these characters without escaping.

---

**Related issue**

Fixes: https://github.com/opnsense/core/issues/10296

### DIFF
--- a/src/opnsense/mvc/app/models/OPNsense/Kea/KeaCtrlAgent.php
+++ b/src/opnsense/mvc/app/models/OPNsense/Kea/KeaCtrlAgent.php
@@ -68,6 +68,6 @@ class KeaCtrlAgent extends BaseModel
             ]
         ];
 
-        File::file_put_contents($target, json_encode($cnf, JSON_PRETTY_PRINT), 0600);
+        File::file_put_contents($target, json_encode($cnf, JSON_PRETTY_PRINT, JSON_UNESCAPED_UNICODE), 0600);
     }
 }
--- a/src/opnsense/mvc/app/models/OPNsense/Kea/KeaDdns.php
+++ b/src/opnsense/mvc/app/models/OPNsense/Kea/KeaDdns.php
@@ -116,6 +116,6 @@ class KeaDdns extends BaseModel
             ]
         ];
 
-        File::file_put_contents($target, json_encode($cnf, JSON_PRETTY_PRINT), 0600);
+        File::file_put_contents($target, json_encode($cnf, JSON_PRETTY_PRINT, JSON_UNESCAPED_UNICODE), 0600);
     }
 }
--- a/src/opnsense/mvc/app/models/OPNsense/Kea/KeaDhcpv4.php
+++ b/src/opnsense/mvc/app/models/OPNsense/Kea/KeaDhcpv4.php
@@ -400,6 +400,6 @@ class KeaDhcpv4 extends BaseModel
                 'server-port' => $ddns->general->server_port->asInt(),
             ];
         }
-        File::file_put_contents($target, json_encode($cnf, JSON_PRETTY_PRINT), 0600);
+        File::file_put_contents($target, json_encode($cnf, JSON_PRETTY_PRINT | JSON_UNESCAPED_UNICODE), 0600);
     }
 }
--- a/src/opnsense/mvc/app/models/OPNsense/Kea/KeaDhcpv6.php
+++ b/src/opnsense/mvc/app/models/OPNsense/Kea/KeaDhcpv6.php
@@ -433,6 +433,6 @@ class KeaDhcpv6 extends BaseModel
                 'server-port' => $ddns->general->server_port->asInt(),
             ];
         }
-        File::file_put_contents($target, json_encode($cnf, JSON_PRETTY_PRINT), 0600);
+        File::file_put_contents($target, json_encode($cnf, JSON_PRETTY_PRINT | JSON_UNESCAPED_UNICODE), 0600);
     }
 }


Monviech:
Testing it:

Valid (notice the ü, ä and ”):

```
                "reservations": [
                    {
                        "ip-address": "192.168.30.1",
                        "hostname": "München",
                        "hw-address": "fe:fe:fe:fe:fe:fe",
                        "user-context": {
                            "uuid": "6d5349a0-84f2-4874-8590-93ec7e241d7d",
                            "description": "München"
                        }
                    },
                    {
                        "ip-address": "192.168.30.2",
                        "hostname": "”Käse",
                        "hw-address": "fe:fe:fe:fe:fe:f0",
                        "user-context": {
                            "uuid": "205a38ad-de90-40b8-ba98-d2a060c530a4",
                            "description": "”Käse"
                        }
                    }
                ],
```

Invalid:

```
                "reservations": [
                    {
                        "ip-address": "192.168.30.1",
                        "hostname": "M\u00fcnchen",
                        "hw-address": "fe:fe:fe:fe:fe:fe",
                        "user-context": {
                            "uuid": "6d5349a0-84f2-4874-8590-93ec7e241d7d",
                            "description": "M\u00fcnchen"
                        }
                    },
                    {
                        "ip-address": "192.168.30.2",
                        "hostname": "\u201dK\u00e4se",
                        "hw-address": "fe:fe:fe:fe:fe:f0",
                        "user-context": {
                            "uuid": "205a38ad-de90-40b8-ba98-d2a060c530a4",
                            "description": "\u201dK\u00e4se"
                        }
                    }
                ],
```

To be specific it is not about the description in user-context, it can happen with hostname too, and any other field, as its a global validation of the whole file.

fichtner (approved):
ok for me but @AdSchellevis should ack this too

AdSchellevis:
ack :)

PR #10299, opened 2026-05-13
Firewall: Rules [new]: Fix action, ipprotocol and protocol translations (legacy rules)

### LINKED ISSUE #10249: Existing legacy rules are not displayed correctly in the new rules (allow/block symbol). German and other translations.
**Important notices**

**Describe the bug**

After updating to version 26.1, a display error was noticed in the new rules. This only affects translated versions, in my case, German.

The legacy rules were transferred to the new rules using the wizard. The existing automatic rules were apparently not exported or imported, but rather loaded in the code via MVC. This is where the display error occurred. The Allow/Deny symbol is always set to Allow, which is obviously incorrect for rules that are set to Default Deny. See screenshot.

**To Reproduce**

1. Use a version prior to 26.1
2. Set the language to German
3. Update to 26.1
4. Export the existing rules using the Migration Assistant
5. Import the exported rules using the Migration Assistant
6. Open "New" under Firewall Rules
7. Inspect the automatic rules.
8. You will see, for example, that the "Default deny / state violation rule" is marked with the "Allow" symbol.

**Expected behavior**

The symbol should be the Deny symbol.

When querying `api/firewall/filter/search_rule/`, there appears to be an unintended translation of the `action` field. With legacy rules, the field is translated into German, which I believe is a bug. With MVC rules, no translation is performed; instead, the `%action` field is added as a translated version.

Perhaps this is a possible solution? Not tested!
https://github.com/hartmann-daniel/opnsense-core/commit/d94fabb285a5da41a30a02972e65ee4ad30c2fc3

**Screenshots**

English Version (no Problem):
Image

German Version (Problem in First Line)
Image

**Relevant log files**

No

**Additional context**

No

**Environment**

OPNsense 26.1.7_1-amd64

Monviech:
**Important notices**

- Model used:
- Extent of AI involvement:

---

**Describe the problem**

- Add % fields to separate translated and raw value for the frontend.

---

**Related issue**

Fixes: https://github.com/opnsense/core/issues/10249


src/opnsense/mvc/app/controllers/OPNsense/Firewall/Api/FilterController.php
```diff
@@ -57,17 +57,18 @@ private function getLegacyFieldMap()
                 $descr = !empty($ifdetail->descr) ? $ifdetail->descr : strtoupper($key);
                 $this->legacy_fieldmap['interface'][$key] = $descr;
             }
-            $this->legacy_fieldmap['action'] = [
+            $this->legacy_fieldmap['%action'] = [
```
fichtner, line 60: I'm not entirely sure but this could ask for trouble last time I went over the code to try and fix some of this

src/opnsense/mvc/app/controllers/OPNsense/Firewall/Api/FilterController.php
```diff
@@ -57,17 +57,18 @@ private function getLegacyFieldMap()
...
                 'pass' => gettext('Pass'),
                 'reject' => gettext('Reject'),
             ];
 
-            $this->legacy_fieldmap['ipprotocol'] = [
+            $this->legacy_fieldmap['%ipprotocol'] = [
                 'inet' => gettext('IPv4'),
                 'inet6' => gettext('IPv6'),
-                'inet46' => '', /* XXX remove when filter.lib.inc use is removed */
+                'inet46' => gettext('IPv4+IPv6'), /* XXX remove when filter.lib.inc use is removed */
```
fichtner: this was so "*" is displayed, which is a bit cleaner than cluttering IPv4+IPv6

src/opnsense/scripts/filter/list_non_mvc_rules.php
```diff
@@ -75,6 +75,9 @@
             }
         }
         $rule['action'] = $rule['action'] ?? 'pass';
+        $rule['%action'] = $rule['action'] ?? 'pass';
```
fichtner: what's the benefit of this if we use the same value that we would read anyway from the field above?

src/opnsense/scripts/filter/list_non_mvc_rules.php
```diff
@@ -75,6 +75,9 @@
...
+        $rule['%ipprotocol'] = $rule['ipprotocol'] ?? '';
+        $rule['%protocol'] = strtoupper($rule['protocol'] ?? '');
```
fichtner: this one at least makes some sense :)

src/opnsense/mvc/app/controllers/OPNsense/Firewall/Api/FilterController.php
Monviech: `%field` holds the translated value, `field` holds the raw value.

I need both.

`%action` is decorated here, and `action` is passed through raw.

Should we do something else?

src/opnsense/mvc/app/controllers/OPNsense/Firewall/Api/FilterController.php
Monviech: But then it's not the same as the MVC rules display inet46:

https://github.com/opnsense/core/blob/1aa4254289cdf018a09e59c3c89798e9435dc459/src/opnsense/mvc/app/models/OPNsense/Firewall/Filter.xml#L79

src/opnsense/scripts/filter/list_non_mvc_rules.php
Monviech: Yeah it was pragmatic to make it look like the MVC "translations" of this field.

src/opnsense/scripts/filter/list_non_mvc_rules.php
Monviech: `%action` is decorated, and `action` is passed through raw. If I miss something obvious please tell me.

src/opnsense/mvc/app/controllers/OPNsense/Firewall/Api/FilterController.php
fichtner: it doesn't have to be. `inet46` is an atrocity

src/opnsense/scripts/filter/list_non_mvc_rules.php
fichtner: where is it decorated?

src/opnsense/mvc/app/controllers/OPNsense/Firewall/Api/FilterController.php
fichtner: is this the spot where the decoration takes place? or at least where the information is provided?

fichtner: I mean it makes some sense but we're decorating in the JS then which is a bit pointless IMO

Monviech: As discussed we can use %action and %direction etc... inside here now instead of hardcoding the translations. 

https://github.com/opnsense/core/blob/1aa4254289cdf018a09e59c3c89798e9435dc459/src/opnsense/mvc/app/views/OPNsense/Firewall/filter_rule.volt#L439

A while ago there was no simpler way to do this because %field and field did not exist in the same way yet.

src/opnsense/mvc/app/controllers/OPNsense/Firewall/Api/FilterController.php
```diff
@@ -165,6 +169,9 @@ public function searchRuleAction()
                 $label = gettext('Automatically generated rules');
                 $record['categories'] = $label;  // Grouping key for tree view
                 $record['category_colors'] = [['name'  => $label]];  // Category formatter metadata
+            } else {
+                /* frontend can format categories with colors */
+                $record['category_colors'] = $this->getCategoryColors($r_categories);
```
Monviech, line 180: This one was moved out to a different PR since unrelated to this one:
https://github.com/opnsense/core/pull/10310

src/opnsense/mvc/app/models/OPNsense/Firewall/Filter.xml
```diff
@@ -76,7 +76,7 @@
                     <OptionValues>
                         <inet>IPv4</inet>
                         <inet6>IPv6</inet6>
-                        <inet46>IPv4+IPv6</inet46>
+                        <inet46>Any</inet46>
```
fichtner:
```suggestion
                        <inet46>any</inet46>
```

for consistency

src/opnsense/scripts/filter/list_non_mvc_rules.php
```diff
@@ -99,13 +97,23 @@
...
         }
 
+        // The %fields need translations in the controller
+        $rule['direction'] = $rule['direction'] ?? 'in';
+        $rule['%direction'] = $rule['direction'];
+        $rule['action'] = $rule['action'] ?? 'pass';
+        $rule['%action'] = $rule['action'];
+        $rule['ipprotocol'] = $rule['ipprotocol'] ?? '';
+        $rule['%ipprotocol'] = $rule['ipprotocol'];
+        // This one is "translated" via uppercase
```
fichtner:
```suggestion
```

this is obvious ;)

fichtner (approved):

src/opnsense/mvc/app/models/OPNsense/Firewall/Filter.xml
fichtner: DNat.xml also needs this fix

PR #10319, opened 2026-05-17
Add cmd timeout, error detection and syslog logging to wireguard re-r…

deajan:
Refactor DNS re-resolution script to improve logging and command execution, make it generally more reliable, add proper exit code, and make it portable for use in other projects (tested in OPNSense 26.1 and RHEL9 so far)

**Important notices**

- [X] I opened an issue first for non-trivial changes and linked it below (not applicable)

**Describe the problem**

I'm currently investigating issues with wireguard where after a WAN drop, wireguard never reconnects (handshake becomes stale) unless I restart the service on local or remote OPNsense side.

This PR adds stderr capture to the commands, adds a timeout in case a subprocess command gets stuck, and adds generic  (rotated) logging for optional debugging purposes, catches all possible errors and logs them.
It is a first step into diagonsis of what actually happens with wireguard not reconnecting.

@AdSchellevis Please let me know if I'm out of line with this.
~~I would also like to add an optional "restart service" action when handshake is stale and updating resolved FQDN don't fix the issue, but that's too broad for a diagnostic right now.~~

### DIFF
--- a/src/opnsense/scripts/wireguard/reresolve-dns.py
+++ b/src/opnsense/scripts/wireguard/reresolve-dns.py
@@ -1,31 +1,30 @@
 #!/usr/local/bin/python3
 
 """
-Copyright (c) 2023-2026 Ad Schellevis <ad@opnsense.org>
-All rights reserved.
+    Copyright (c) 2023-2026 Ad Schellevis <ad@opnsense.org>
+    All rights reserved.
 
-Redistribution and use in source and binary forms, with or without
-modification, are permitted provided that the following conditions are met:
+    Redistribution and use in source and binary forms, with or without
+    modification, are permitted provided that the following conditions are met:
 
-1. Redistributions of source code must retain the above copyright notice,
- this list of conditions and the following disclaimer.
+    1. Redistributions of source code must retain the above copyright notice,
+     this list of conditions and the following disclaimer.
 
-2. Redistributions in binary form must reproduce the above copyright
- notice, this list of conditions and the following disclaimer in the
- documentation and/or other materials provided with the distribution.
+    2. Redistributions in binary form must reproduce the above copyright
+     notice, this list of conditions and the following disclaimer in the
+     documentation and/or other materials provided with the distribution.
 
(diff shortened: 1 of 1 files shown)


AdSchellevis:
I don't mind adding some logging as mentioned earlier, but do mind using our production codebase to hunt some unspecified issue. 

Adding logging usually costs a couple of lines of code, easy to review without further impact.

deajan:
In this PR I did remove everythnig not logging/error catching related.

Just adding some log lines somewhere won't make the code bullet proof, since the whole script would exit without any logging on first error.

The only reason why there are many modified lines is indentation from the fact that previous code wasn't wrapped into functions. The wrapping makes it easier to implement error detection and logging IMO, and is generally considered a best practice.

The only actual code part I refactored here is the subprocess call, which I made into a separate function in order to avoid duplicating error catching code if subprocess call fails.

The full logic of your initial code is kept the same.
Isn't that good enough ?

AdSchellevis:
It's probably best to start with a ticket briefly describing the problem we do want to solve, this whole PR feels like sending a rocket to the moon..... 

If the script errors and exits without sending the message out, that's probably a parameter on `subprocess` to set, if the underlaying command has issues (which causes it not to exit), we need to understand the criteria under which that happens so we can assess what need fixing. Reading your code leads to a lot of questions, for which we simply have no supporting tickets pointing into a specific direction. 

I do understand your journey trying to fix your issue, but at the end of the day simplicity almost always wins over complexity, which makes it even more important to understand why changes are needed (and which bells and whistles are simply irrelevant).

Wrapping a function around the usual `if __name__ == '__main__':` isn't a problem, but adding more glue than functional code certainly is (and by all means has nothing to do with best practices....).

Long story short, let's start with a problem instead of a solution.....

deajan:
Basically I'm trying to solve why wireguard does not reconnect at times, hence I need the logging and error handling.

But okay, I see your point. I will leave this PR as is for now, manually keep my version of the script on my OPNsense instances until I get some logs, so I get to open a ticket with a description better than "it doesn't work sometimes" ;)

PR #10333, opened 2026-05-20
ui: improve form validation error append

fichtner:
Since this iterates over a lot of irrelevant IDs and then mismatches with the target change this by safeguarding against fields that are likely not going to work without help_block_<id> and switch target to a suffix match.

### DIFF
--- a/src/opnsense/www/js/opnsense.js
+++ b/src/opnsense/www/js/opnsense.js
@@ -236,16 +236,22 @@ function setFormData(parent,data) {
 function handleFormValidation(parent, validationErrors)
 {
     $("#" + parent).find("[id]").each(function () {
-        let target = $("*[id*='" + $(this).prop('id') + "']");
+        let help_block = $("span[id='help_block_" + $(this).prop('id') + "']");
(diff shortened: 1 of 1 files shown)


fichtner:
There's a visual drawback with the tokenizer, but something we already discussed because it is in need of a replacement anyway.

src/opnsense/www/js/opnsense.js
```diff
@@ -236,16 +236,22 @@ function setFormData(parent,data) {
 function handleFormValidation(parent, validationErrors)
 {
     $("#" + parent).find("[id]").each(function () {
-        let target = $("*[id*='" + $(this).prop('id') + "']");
+        let help_block = $("span[id='help_block_" + $(this).prop('id') + "']");
+        if (!help_block.length) {
+            return true;
```
swhite2, line 241: what's the `true` about here?

fichtner: true continues, false breaks the loop

Ari4ka (approved):

PR #10341, opened 2026-05-22
Firewall: Rules [new]: Add hint banner about Inspect button to explain it unhides automatic rules

Monviech:
**Important notices**

- Model used:
- Extent of AI involvement:

---

**Describe the problem**

If the Inspect button has never been pressed before in a browser, the banner will show to advertise that it unhides automatic firewall rules, hopefully helping first time users better who are not as adventurous to click the few available buttons nor read the documentation (but hopefully read the banner)

### DIFF
--- a/src/opnsense/mvc/app/views/OPNsense/Firewall/filter_rule.volt
+++ b/src/opnsense/mvc/app/views/OPNsense/Firewall/filter_rule.volt
@@ -1153,8 +1153,7 @@
     </div>
     <!-- grid -->
     <div id="inspect_hint_banner" class="alert alert-info" style="display: none;">
-        <i class="fa fa-info-circle"></i>
-        {{ lang._('Automatic firewall rules are hidden by default. To show them, press the Inspect button.') }}
+        {{ lang._('Automatic firewall rules are hidden by default. Press the %s button to reveal them.') | format('<i class="fa fa-fw fa-eye"></i>&nbsp;' ~ lang._('Inspect')) }}
     </div>
     {{ partial('layout_partials/base_bootgrid_table', formGridFilterRule + {'command_width': '180'}) }}
 </div>
--- a/src/opnsense/mvc/app/views/OPNsense/Firewall/filter_rule.volt
+++ b/src/opnsense/mvc/app/views/OPNsense/Firewall/filter_rule.volt
@@ -57,8 +57,11 @@
         let treeViewEnabled = localStorage.getItem("firewall_rule_tree") === "1";
         $('#toggle_tree_button').toggleClass('active btn-primary', treeViewEnabled);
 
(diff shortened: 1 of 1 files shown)


fichtner:
Thanks for proposing this. I’m not sure this helps. A new visible button may be much easier. Or a second tab for automatic rules? That wouldn’t help searching but it would separate concerns at least. Still thinking about it.

Monviech:
A new button would be fine, I think this banner here is very pragmatic too. Once somebody learned that Inspect unhides automatic rules there shouldn't be an issue anymore.

I don't like a second tab, as all rules should be shown in the correct sort_order at all times.

Personally I would go with this banner first as a test to see if the reports go away with it.

AdSchellevis:
I'm open to try it out, but don't have a very strong opinion on the matter.

Monviech:
@fichtner I could also offer "Inspect" and "Tree" enabled by default on first page load. If there are no rules yet, or when its after a migration, that shouldn't be a (big) problem. It just takes 1 second longer to load the initial page or so (depending on hardware, if its a potato then it can take longer)

Theoretically the banner should do the job just nicely with little glue.

image

fichtner:
@AdSchellevis @Monviech suggested a third button, maybe that's the nicest way if we can condense the toggle presentation a bit while still keeping it clear and visible even for newcomers

AdSchellevis:
I don't have a very strong preference, but when adding another button, auto rules should always be visible when "inspect" is enabled.... otherwise it will be confusing

Monviech:
My personal preference is keeping conditional glue and more buttons that nobody would intentionally press at a minimum.

A new button with a long explanation so that its very obvious what it does also takes a lot of screen real estate and increases decision fatigue and cognitive load.

From my perspective the priority would be:
1. banner
2. inspect default enabled
3. new button that doesnt overlap with Inspect's inherent functionality

I dont want to decide here I'll go the direction thats suggested.

fichtner:
It looks like an impasse.

4. rename the inspect button

Monviech:
Renaming sounds good to me too. Any ideas?

The button's  current functionality is "Show the full impact and statistics of the current ruleset".

AdSchellevis:
"detailed view" 🙃

Monviech:
Since inspect is already well known by now, I would not change the name.

I could offer this approach maybe?

- 1. Extract the rule statistics functionality to its own button called "Statistics". This would ease the performance impact of the normal "Inspect" button
- 2. Enable Inspect by default, which solves the discoverability issue of new users

fichtner:
With the given constraints it’s impossible to change this in a meaningful way. The gamble is either people get to accept it or will keep asking in which case the constraints need to be loosened. 😉

Monviech:
Let's move this discussion out of this PR then. I will create a new issue and we can see how we can approach this. I'm open for any change that helps users, but does not inflate the code too much.

Issue #10343, opened 2026-05-22
Kea per-reservation next-server option does nothing


DarkStarSword:
**Important notices**

Before you add a new report, we ask you kindly to acknowledge the following:

- [ Y ] I have read the contributing guidelines at https://github.com/opnsense/core/blob/master/CONTRIBUTING.md
- [ Y ] I am convinced that my issue is new after having checked both open and closed issues at https://github.com/opnsense/core/issues?q=is%3Aissue

**Describe the bug**

The "Next server" field in a reservation form has no effect - it does not generate the expected "next-server" field in the Kea config, and therefore the host attempts to pull the netboot image from the OPNSense IP instead of the expected IP.

Note that the "Next server" field in the subnet form does work as expected - this issue is only in regards to the one in the reservation form.

Note: This particular host does not seem to make use of option 66 "TFTP server" (which does appear to generate the correct config), it relies on "next-server" being set.

**To Reproduce**

1. Configure a host for netboot
2. Create a reservation in Kea matching this host, filling out "Next server" and "TFTP bootfile name"
3. Attempt to netboot the host, observe that it attempts to pull the image from the OPNSense IP instead of the next-server IP.

**Expected behavior**

The host the reservation matches retrieves the netboot image from the server specified in next-server.

**Describe alternatives you considered**

Setting Next Server in the subnet form does work, however this is limited flexibility, and an option presented in the UI should be functional.

**Screenshots**

<img width="857" height="350" alt="Image" src="https://github.com/user-attachments/assets/7d8119ed-9e99-4c4e-81c9-bbaafec9233e" />

<img width="1310" height="669" alt="Image" src="https://github.com/user-attachments/assets/9182f10b-5476-4202-8909-590854ee653d" />

<img width="921" height="452" alt="Image" src="https://github.com/user-attachments/assets/de6d48dc-aaa4-4380-be30-36d532650a01" />

**Relevant log files**

**Additional context**

Pull request incoming.

**Environment**

OPNsense 26.1.8_5-amd64

AdSchellevis:
@DarkStarSword  seems like I missed this in reviewing https://github.com/opnsense/core/pull/8890, thanks!

AdSchellevis:
closed with https://github.com/opnsense/core/pull/10344

PR #8890, opened 2025-06-27
Add DNS field to Kea DHCP4 Reservations

Gtt1229:
Added DNS field to Kea DHCP4 Reservations.

Utilizes masking instead of tokeniation for UI consistency.

Tested on 25.1.9_2

Kea reference for options: https://kea.readthedocs.io/en/kea-2.0.0/arm/dhcp4-srv.html#dhcp4-std-options

PR #10344, opened 2026-05-22
Kea: Hook up reservation.next_server

DarkStarSword:
**Important notices**

- [ Y ] I have read the contributing guidelines at https://github.com/opnsense/core/blob/master/CONTRIBUTING.md
- [ Y ] I opened an issue first for non-trivial changes and linked it below.
- [ No ] AI tools were used to create at least part of the code submitted herewith.

---

**Describe the problem**

This field was added to the Kea dialogReservation4.xml form in https://github.com/opnsense/core/pull/8890, however this specific option was not properly hooked up and did not generate the expected config, preventing netboot scenarios that rely on next-server from being set on a per-reservation basis (the per-subnet variant of this option does work fine).

---

This commit hooks it up to generate the expected "next-server" Kea config entry on a per-reservation basis, allowing netboot to work on a per-reservation basis.

---

**Related issue**

https://github.com/opnsense/core/issues/10343

### DIFF
--- a/src/opnsense/mvc/app/models/OPNsense/Kea/KeaDhcpv4.xml
+++ b/src/opnsense/mvc/app/models/OPNsense/Kea/KeaDhcpv4.xml
@@ -1,6 +1,6 @@
 <model>
     <mount>//OPNsense/Kea/dhcp4</mount>
-    <version>1.0.4</version>
+    <version>1.0.5</version>
     <description>Kea DHCPv4 configuration</description>
     <items>
         <general>
@@ -318,6 +318,10 @@
                     <IsDNSName>Y</IsDNSName>
                 </hostname>
                 <description type="DescriptionField"/>
+                <next_server type="NetworkField">
+                    <NetMaskAllowed>N</NetMaskAllowed>
+                    <AddressFamily>ipv4</AddressFamily>
+                </next_server>
                 <option_data>
                     <domain_name_servers type="NetworkField">
                         <NetMaskAllowed>N</NetMaskAllowed>
--- a/src/opnsense/mvc/app/models/OPNsense/Kea/KeaDhcpv4.php
+++ b/src/opnsense/mvc/app/models/OPNsense/Kea/KeaDhcpv4.php
@@ -211,6 +211,10 @@ class KeaDhcpv4 extends BaseModel
                     $res['client-id'] = $reservation->client_id->getValue();
                 }
 
+                if (!$reservation->next_server->isEmpty()) {
+                    $res['next-server'] = $reservation->next_server->getValue();
+                }
+
                 // Add DHCP option-data elements for reservations
                 $optdata = $this->collectOptionData($reservation->option_data);
                 /* append raw options */


src/opnsense/mvc/app/models/OPNsense/Kea/KeaDhcpv4.xml
```diff
@@ -1,6 +1,6 @@
 <model>
     <mount>//OPNsense/Kea/dhcp4</mount>
-    <version>1.0.4</version>
+    <version>1.0.5</version>
```
Monviech, line 3: @AdSchellevis Model bump is not needed I think.

AdSchellevis: @Monviech technically not required indeed (we're not expecting any defaults rendered in a template), doesn't hurt much either to be honest.

fichtner:
@Monviech likely unrelated to this chabge but I'm seeing:

```
[OPNsense\Kea\KeaDhcpv4:reservations.reservation.d8b781cf-2a89-4289-b87e-f0c2ba4bdefe.ip_address] Address not in specified subnet{10.2.0.1}
Model OPNsense\Kea\KeaDhcpv4 can't be saved, skip ( OPNsense\Base\ValidationException: [OPNsense\Kea\KeaDhcpv4:reservations.reservation.d8b781cf-2a89-4289-b87e-f0c2ba4bdefe.ip_address] Address not in specified subnet{10.2.0.1}
 in /usr/local/opnsense/mvc/app/models/OPNsense/Base/BaseModel.php:822
Stack trace:
#0 /usr/local/opnsense/mvc/app/models/OPNsense/Base/BaseModel.php(947): OPNsense\Base\BaseModel->serializeToConfig()
#1 /usr/local/opnsense/mvc/script/run_migrations.php(69): OPNsense\Base\BaseModel->runMigrations()
#2 {main} )
*** OPNsense\Kea\KeaDhcpv4 migration failed from 1.0.4 to 1.0.5, check log for details
```

Anything I screwed up or something we need to look at?

Monviech:
@fichtner 

Do you have a reservation with the IP address:
10.2.0.1

Is it part of the subnet its configured in?
(10.2.0.0/24)

Otherwise this validation:
https://github.com/opnsense/core/blob/8adcc434a35a231092209908ebe47d6667a6d986/src/opnsense/mvc/app/models/OPNsense/Kea/KeaDhcpv4.php#L88

Even if your config is invalid shouldn't it bail out here?
https://github.com/opnsense/core/blob/8adcc434a35a231092209908ebe47d6667a6d986/src/opnsense/mvc/app/models/OPNsense/Kea/KeaDhcpv4.php#L79

fichtner:
Strangely it's part of a 10.0.0.0/16 subnet. Looks like this was allowed at some point? Happy to drop the faulty entry but we should be wary this could be happening in the field too if we made it stricter. It doesn't matter until a migration kicks in...

Monviech:
Isn't 10.0.0.0/16:

10.0.0.0 through 10.0.255.255?

So 10.2.0.1 is not in there.

Also KEA doesn't start:
2026-06-01T08:15:42Errorkea-dhcp4ERROR [kea-dhcp4.dhcp4.0x25f91665c008] DHCP4_INIT_FAIL failed to initialize Kea server: configuration error using file '/usr/local/etc/kea/kea-dhcp4.conf': specified reservation '192.168.30.1' is not within the IPv4 subnet '192.168.80.0/24'

If you:
- Create a subnet
- Create a reservation
- Change the subnet IP afterwards -> No validation to prevent this

But as KEA will fail to start anyway I wonder how long such a config would stick around if the user actually wants to use KEA.

Issue #10094, opened 2026-04-06
WireGuard Instance tunnel address applied as 0.0.0.X/8 instead of correct IP


polumish:
## Environment

- OPNsense version: **26.1** (stable/26.1-n272033-b4ddb3e0f150)
- FreeBSD: **14.3-RELEASE-p9** (amd64)
- WireGuard: built-in (core, not os-wireguard plugin)
- Reproducible on: 2 out of 3 OPNsense boxes with the same firmware

## Description

When creating a WireGuard Instance via the GUI (VPN → WireGuard → Instances), the **Tunnel address** field is stored correctly in `config.xml` but the kernel interface receives a wrong IP address.

### Expected

- config.xml: `<tunneladdress>10.200.0.2/24</tunneladdress>`
- Interface: `inet 10.200.0.2 netmask 0xffffff00`

### Actual

- config.xml: `<tunneladdress>10.200.0.2/24</tunneladdress>` ✅ (correct)
- Interface: `inet 0.0.0.2 netmask 0xff000000` ❌ (wrong — `0.0.0.2/8`)

The pattern: the **last octet** of the real address (`.2`, `.4`) appears as the full address (`0.0.0.2`, `0.0.0.4`), and the netmask becomes `/8` instead of `/24`.

## Steps to Reproduce

1. Go to VPN → WireGuard → Instances → Add
2. Set **Tunnel address**: `10.200.0.2/24`
3. Set a valid private key, listen port, and optionally a peer
4. **Save** → **Apply**
5. SSH into OPNsense, run `ifconfig wg1` (or whichever interface was created)
6. Observe: `inet 0.0.0.2 netmask 0xff000000` instead of `inet 10.200.0.2 netmask 0xffffff00`

## Verification

```
# config.xml stores the correct value:
grep tunneladdress /conf/config.xml
# Output: <tunneladdress>10.200.0.2/24</tunneladdress>

# But the interface has wrong address:
ifconfig wg1
# Output: inet 0.0.0.2 netmask 0xff000000
```

## Workaround

Manually fix the address after Apply:
```
ifconfig wg1 inet 0.0.0.2/8 delete
ifconfig wg1 inet 10.200.0.2/24 alias
```

Or: edit the Instance again → Save → Apply (sometimes re-applying fixes it).

## Additional Notes

- Tested on **3 separate OPNsense boxes** with the same firmware (26.1). Bug reproduced on **2 out of 3**.
[...]

PR #10345, opened 2026-05-23
wireguard: heal interface when configured tunneladdress is missing

C00kiekiller:
Ref: #10106
Ref: #10094

**Important notices**

- Model used: Claude (Opus 4.7) via Claude Code CLI
- Extent of AI involvement: AI assistance was used to analyze the OPNsense source paths (interfaces.inc, plugins.inc.d/wireguard.inc, wg-service-control.php), draft the patch iterations, and structure the reproducer scripts. The bug was reproduced and the patch verified manually on a pair of OPNsense 26.1.8_5 test VMs (one patched, one unpatched as control), and the patch deployed and observed on a production HA pair before submission. All code was reviewed and refined across multiple iterations before submitting.

---

**Describe the problem**

The wg interface can end up with only the if_wg(4) default `0.0.0.2/8` when something destroys+recreates the device without a follow-up `wg_start` (e.g. `interface_configure()` running before `plugins_configure('vpn')`). Routing to peers breaks and `configctl wireguard configure` doesn't heal it because the stat-hash check short-circuits when the cnf-file md5 is unchanged. The only workaround today is toggling the instance via the GUI (disable+apply, enable+apply) to force the hash to differ, which isn't discoverable.

---

Two changes to `wg-service-control.php`:

1. `wg_start()` flushes addresses on the device before re-applying the tunneladdress aliases. Same `interfaces_addresses_flush()` helper that the existing "can not reconfigure without stopping it first" branch already uses, just applied unconditionally so any kernel-assigned default or stale alias is cleared.
2. In the `configure` action, a new `wg_has_tunneladdress()` predicate is added to the outer condition. When the device exists but the configured tunneladdress is missing from it, a full reconfigure is forced regardless of the stat-hash, with an explicit syslog notice.

Tested on 26.1.8_5. The bug reproduces deterministically with `ifconfig wgN destroy && ifconfig wg create name wgN && ifconfig wgN group wireguard` followed by `configctl wireguard configure` (no-op without the patch, restores tunneladdress with it).

---

**Related issue**

#10106, #10094

### DIFF
--- a/src/opnsense/scripts/wireguard/wg-service-control.php
+++ b/src/opnsense/scripts/wireguard/wg-service-control.php
@@ -72,6 +72,11 @@ function wg_start($server, $fhandle, $ifcfgflag = 'up', $reload = false)
 
     mwexecf('/usr/bin/wg syncconf %s %s', [$server->interface, $server->cnfFilename]);
 
+    /* drop leftover/default addresses before re-applying aliases */
+    $details = legacy_interfaces_details((string)$server->interface);
+    interfaces_addresses_flush((string)$server->interface, 4, $details);
+    interfaces_addresses_flush((string)$server->interface, 6, $details);
+
     foreach ($server->tunneladdress->getValues() as $alias) {
         $proto = strpos($alias, ':') === false ? "inet" : "inet6";
         mwexecf('/sbin/ifconfig %s %s %s alias', [$server->interface, $proto, $alias]);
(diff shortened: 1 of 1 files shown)


AdSchellevis:
Without a clear path to reproduce an issue, this certainly isn't going to meet the bar. It's likely some (custom) script or application mangled with the interface, but if there's a reproducible path causing the issue with a default install, there's something we can look into, but I think my colleague already stated something similar in the ticket (https://github.com/opnsense/core/issues/10094#issuecomment-4437440803).

fichtner (changes requested):
Not going to happen. Find the root cause please.

C00kiekiller:
understood, try to find the root cause.

PR #10355, opened 2026-05-26
Firewall: Rules [new]: Always show automatic and legacy rules, and a few more UI tweaks

### LINKED ISSUE #10352: Firewall Rules [new]: Better visibility for automatic rules
**Important notices**

Automatic rules are currently hidden behind the "Inspect" button, which has a performance impact if enabled by default because it also collects statistics.

It seems like some users have a discoverability issue finding automatic rules.

Documentation does not reliably help here.

Making automatic rules better discoverable by default.

Not changing the current approach.

**Additional context**

Initial discussion was around a banner, but it's not a sufficient solution:
https://github.com/opnsense/core/pull/10341

Monviech:
**Important notices**

- Model used:
- Extent of AI involvement:

---

**Related issue**

Fixes: https://github.com/opnsense/core/issues/10352

This change will always show a top level tree of the Automatic Rules, but all other rules will not be grouped by Category -  unless Tree view is also activated.

Side effect is that also all legacy rules that still exist will be mixed into the current ruleset, though in my opinion that change helps even more with the general visibility, and demystifies the inspect button.

Performance wise this will be a tiny bit slower I assume, since the additional fetch of the internal rules is now always active (without an opt out). Though after the first call the result is cached by configd.
https://github.com/opnsense/core/blob/09654ca796589e613affee99a86190e8b815ad08/src/opnsense/service/conf/actions.d/actions_filter.conf#L197

Diff is a bit larger since I also renamed some short variables into longer ones for better readability (r -> row, resp -> response et al)


swhite2 (dismissed):
Haven't tested it yet, but looks good code-wise

swhite2:
When selecting "All", the autogenerated rules include rule with a higher sort order than the user defined rules:

image

When limiting with e.g. 2 pages, the auto generated rules at the end of the set are properly rendered on the second page only.

Further discussion:

- The auto generated rules count as part of the pagination. To some extent this is weird, but the only way to solve it is to render all rules by default (disable pagination). The question is if this is acceptable performance-wise, I think so in general.
- We can also group the "user defined rules", but by doing so, we'd have to disable pagination as well to make sure all relevant rules appear at all times

Monviech:
Note for me it could also be a good idea to always show the "Collapse All" button if there will always be groups visible

swhite2:
Since we're here.. maybe some default widths make sense:

- Version: 80px
- port: 60px
- port: 60px
- protocol: 80px

swhite2:
As discussed, https://github.com/opnsense/core/commit/3507ecffc4ad7ae30932a189c89d5646d7b0be5e should address the possible edge case of deselecting categories while groups are collapsed

src/opnsense/mvc/app/controllers/OPNsense/Firewall/Api/FilterController.php
```diff
@@ -190,18 +183,16 @@ public function searchRuleAction()
...
+        $allrules = array_merge(
+            $allrules,
+            json_decode((new Backend())->configdRun('filter list non_mvc_rules'), true) ?? []
+        );
 
         $search_clauses = [];
         $backend = new Backend();
         foreach (preg_split('/\s+/', (string)$this->request->getPost('searchPhrase', null, '')) as $token) {
-            if ($show_all && Util::isIpAddress($token)) {
+            if (Util::isIpAddress($token)) {
```
swhite2: I think this should still be behind the `$show_all`, it's quite expensive

swhite2:
Last point, likely to be addressed in a separate PR, is If a specific interface is selected, all groupings should be collapsed, except for the `interface` rules, or `group` rules for "group" type interfaces such as IPsec.

On "All rules" however, alle groupings should be expanded by default.

It's likely best to disable any form of persistence for the grouping here to ease this implementation, and we should consider the expand/collapse behavior for searching as well.

src/opnsense/mvc/app/controllers/OPNsense/Firewall/Api/FilterController.php
Monviech: It would rather fit into an "advanced search" button next to the search bar, I want to keep the inspect button with one job only now, showing statistics.

Monviech: For now I will hide it behind the inspect button again. Since the search functionalities future is still in the works, a dedicated button can be added later if needed.

PR #10359, opened 2026-05-29
Firewall: NAT: Add the same UI design to the NAT pages as the firewall page

Monviech:
**Important notices**

- Model used:
- Extent of AI involvement:

---

**Describe the solution**

Add the same UI design to the NAT pages as the firewall page from 18533b4, but slightly adjusted as NAT rules do not have sort_order or priority groups, so all grouping happens either by category, or to a synthetic automatic category if the rules contain an is_automatic true boolean.

---

**Related issue**

https://github.com/opnsense/core/issues/10355


Monviech:
It would probably be a good idea to implement a sort_order here too, since then I could just use the same javascript as in the firewall rules page.

I think we have to draw some sort of distinction if automatic rules are added first, or last, e.g. in SNAT the automatic rules will be last, but in DNAT they will be first. So a sort_order seems necessary just like the firewall page implements.

This here is just so I don't forget that these pages also need the same love.

Monviech:
@AdSchellevis

I want your approval for the sort_order implementation.

DNAT implements it only in controller
SNAT, OneToOneNAT and NPTv6 implemts it as volatile model fields, same as filter rules already do

The javascript is now the same as here, which means we have the exact same logic: https://github.com/opnsense/core/commit/18533b4c60c27253f8c2ebbdfd9bc11d6450da1d

src/opnsense/mvc/app/controllers/OPNsense/Firewall/Api/DNatController.php
```diff
@@ -75,6 +75,12 @@ public function searchRuleAction()
                     $record["alias_meta_{$field}"] = $this->getNetworks($record[$field]);
                 }
             }
+            // Normal DNAT rule priority should be same as firewall interface rules
+            // This is only used for visualization to ensure the tabulator tree renders
+            // rules in the correct order, similar to firewall rules.
+            // It does not influence the processing order of the ruleset by sequence.
+            $record['sort_order'] = sprintf('%d.0%06d', 400000, (int)($record['sequence'] ?? 0));
```
AdSchellevis, line 82: It's probably the most pragmatic solution at the moment, if we would like to move this to the model (like the firewall rules), more glue code is needed and so-far this only has a meaning for the user-interface anyway.

src/opnsense/mvc/app/views/OPNsense/Firewall/nat_rule.volt
```diff
@@ -50,35 +50,110 @@
...
-                    current.category_colors = r.category_colors || [];
-
-                    buckets.push(current);
+        const ruleTypeMap = {
+            '1': { label: "{{ lang._('Automatically generated rules') }}", icon: "fa-magic", tooltip: "{{ lang._('Automatically generated rules') }}", color: "text-secondary" },
+            '4': { label: "{{ lang._('Interface rules') }}", icon: "fa-ethernet", tooltip: "{{ lang._('Interface rule') }}", color: "text-info" },
+            '5': { label: "{{ lang._('Automatically generated rules') }}", icon: "fa-magic", tooltip: "{{ lang._('Automatically generated rules') }}", color: "text-secondary" },
+        };
+
+        const getRuleTypeDigit = function(row) {
```
AdSchellevis: This feels a bit much to te honest, if we require each record to provide what it is, we may as well always ship it in the api result in a readable form (or is `prio_group` actually static here and we can map on `100000`, `400000` ?).

AdSchellevis (approved):
@Monviech some small remarks, no deal breakers. thanks!

src/opnsense/mvc/app/views/OPNsense/Firewall/nat_rule.volt
Monviech: Yeah we can map to prio_group instead, its static. I somehow missed that, thanks. Just have to expose it to the grid as column.

PR #10366, opened 2026-05-31
Interfaces: Assignments - refactor to MVC

AdSchellevis:
**Important notices**

- Model used: none
- Extent of AI involvement: none

---

**Describe the problem**

MVC migration of interface assignments page.
---

In order to migrate the interface assignments, we need to think of a way to use the differently named xml nodes for interfaces (wan, lan, ..) into something that closely resembles a standard model implementation. Since we can't match these nodes in our statically defined model xmls, the main idea is to flush all via an in-memory model with a separate load [construct] and save hook [serializeToConfig].

The next challenge is to "stash" updates and wait for "apply" in certain cases, for this we add some temporary attributes to the configuration which are synced after the actual system change has happend (pending_action, pending_if). When succesfully applied, the apply function cleans up the final stage of the configuration to make everything consistent again.

---

**Related issue**

https://github.com/opnsense/core/issues/9945


src/opnsense/mvc/app/models/OPNsense/Interfaces/Assignment.php
```diff
@@ -0,0 +1,97 @@
...
+ * POSSIBILITY OF SUCH DAMAGE.
+ */
+
+namespace OPNsense\Interfaces;
+
+use OPNsense\Base\BaseModel;
+use OPNsense\Core\Config;
+
+
+class Assignment extends BaseModel
```
fichtner: Model "Assignment" may be in need of a rename rather sooner than later. Maybe aim for the final name here?

src/opnsense/mvc/app/models/OPNsense/Interfaces/NetInterface.php
```diff
@@ -0,0 +1,97 @@
...
+                $node->if = (string)$intf->if;
+            }
+        }
+    }
+
+
+    /**
+     *  Account changes in config.xml when persisting, return "true" so callers know to flush to the configuration
+     */
+    public function serializeToConfig($validateFullModel = false, $disable_validation = false)
```
fichtner, line 115: if I recall correctly the legacy code did some more things like clearing dhcp settings and so on.. if it's no longer relevant that's good.

Also this might be a missing piece here for InterfaceField use in the future... if we drop an interface we'd ideally have to call normalize() which could break other validation. Being able to figure out a plan here (even if just something like "in use") might be good.

fichtner (approved):
I like the direction, thanks :)

src/opnsense/mvc/app/models/OPNsense/Interfaces/Assignment.php
AdSchellevis: good point, probably better to choose "Interface" here

src/opnsense/mvc/app/models/OPNsense/Interfaces/Assignment.php
```diff
@@ -0,0 +1,97 @@
...
+{
+    /**
+     * Merge configuration data in "in memory" model on construction
+     */
+    public function __construct($lazyload = false)
+    {
+        parent::__construct();
```
swhite2: Not sure if not passing `$lazyload` is intentional?

src/opnsense/mvc/app/models/OPNsense/Interfaces/NetInterface.php
AdSchellevis: > if I recall correctly the legacy code did some more things like clearing dhcp settings and so on.. if it's no longer relevant that's good.

looked at that, but given that stuff moved to plugins, I think we should just ditch it.

> Also this might be a missing piece here for InterfaceField use in the future... if we drop an interface we'd ideally have to call normalize() which could break other validation. Being able to figure out a plan here (even if just something like "in use") might be good.

I have to think about that, some logic may shift around a bit, I'm also not fully happy with the temp storage in config.xml yet, may move that out as well to a locked temp file to prevent a reboot of the box still waiting for an "apply".

src/opnsense/mvc/app/models/OPNsense/Interfaces/Assignment.php
AdSchellevis: method fingerprint, yes

src/opnsense/mvc/app/models/OPNsense/Interfaces/NetInterface.php
```diff
@@ -0,0 +1,97 @@
...
+            }
+            $existing_ifnames[] = $key;
+        }
+        $next_if = 1;
+        for ($i=1; $i < count($existing_ifnames); ++$i) {
+            if (!in_array('opt'.$i, $existing_ifnames)) {
+                $next_if = $i;
+                break;
+            }
+        }
```
sopex, line 135: Let's say we have lan, opt1, opt2 and opt3. And we delete opt1 and opt2. 
So we have lan and opt3, $existing_ifnames is 2, so this will check opt1, not find it, and break and fail, right?

```suggestion
        $next_if = 1;
        while (in_array('opt' . $next_if, $existing_ifnames)) {
            $next_if++;
        }
```

AdSchellevis: a bit shorter indeed, should do the same, why not.

src/opnsense/mvc/app/models/OPNsense/Interfaces/NetInterface.php
AdSchellevis: The cleanup is partially possibly here:

https://github.com/opnsense/core/blob/d10b62f71a5891f1987dfbd545607de801e8a62c/src/opnsense/mvc/app/controllers/OPNsense/Interfaces/Api/AssignmentController.php#L100

But there are some caveats, we could try to locate other "interface[s]" fields, but if removal of the interface in question leads to an empty value, we can't set it empty as that might have unintended side affects. As long as at least one interface remains (or we can add the search path properly), normalizing it would be possible.

src/opnsense/mvc/app/models/OPNsense/Interfaces/NetInterface.xml
```diff
@@ -0,0 +1,23 @@
...
+    <version>0.0.1</version>
+    <description>Interface assignments</description>
+    <items>
+        <interface type="ArrayField">
+            <descr type="DescriptionField"/>
+            <identifier type="TextField"/>
+            <!-- backref from if field -->
+            <icon type="TextField" volatile="true"/>
+            <optgroup type="TextField" volatile="true"/>
+            <if type=".\AssignmentInterfaceField">
```
sopex:
```suggestion
            <if type=".\NetInterfaceField">
```

src/opnsense/mvc/app/controllers/OPNsense/Interfaces/forms/dialogAssignment.xml
```diff
@@ -0,0 +1,39 @@
...
+        <help>Technical identifier of the interface, used by hasync for example.</help>
+    </field>
+-->
+    <field>
+        <id>interface.if</id>
+        <label>Interface</label>
+        <type>dropdown</type>
+        <help>Device name to connect this interface to.</help>
+    </field>
+</form>
```
sopex, line 39:
```
<form>
    <field>
        <id>interface.icon</id>
        <label>Status</label>
        <type>hidden</type>
        <grid_view>
            <width>5em</width>
            <formatter>statusformatter</formatter>
        </grid_view>
    </field>
    <field>
        <id>interface.identifier</id>
        <label>Identifier</label>
        <type>text</type>
        <type>info</type>
        <help>Technical identifier of the interface, used by hasync for example.</help>
        <grid_view>
            <width>5em</width>
        </grid_view>
    </field>
    <field>
        <id>interface.optgroup</id>
        <label>Type</label>
        <type>info</type>
        <grid_view>
            <width>5em</width>
        </grid_view>
    </field>
    <field>
        <id>interface.descr</id>
        <label>Description</label>
        <type>text</type>
        <help>You may enter a description here for your reference (not parsed).</help>
    </field>
    <field>
        <id>interface.if</id>
        <label>Interface</label>
        <type>dropdown</type>
        <help>Device name to connect this interface to.</help>
    </field>
</form>
```

sopex (commented):
My two cents on how the table/ edit box is best displayed.
Also, a minor rename :)

Best,
Konstantinos

src/opnsense/mvc/app/models/OPNsense/Interfaces/NetInterface.xml
AdSchellevis: @sopex thanks, missed a spot there indeed after the rename

fichtner: I don’t think this one is unclear and “net” makes no sense in our code base. Never ever is an interface assignment referred to as a network. If anything “Device” is a better substitute for “Assignment” but it’s not needed either.

src/opnsense/mvc/app/controllers/OPNsense/Interfaces/forms/dialogAssignment.xml
AdSchellevis: @sopex something seems off with the indents here making it almost impossible to track the suggested change. What was your suggestion for the form?

src/opnsense/mvc/app/models/OPNsense/Interfaces/NetInterface.xml
AdSchellevis: @fichtner let's go for DeviceField here, probably better indeed.

src/opnsense/mvc/app/controllers/OPNsense/Interfaces/forms/dialogAssignment.xml
sopex: @AdSchellevis Indentation is fine :( Github is just being difficult, maybe copypaste it locally? I am not sure how to better share it.

AdSchellevis: ok, I just couldn't find the suggested change, which makes it difficult then.

sopex: I am not sure why GitHub doesn't render it here.
Have a look here, if it helps:  https://github.com/sopex/core/commit/2ff695e9d6f8fecb7f3a00ac428a3ad6d52b181c

src/opnsense/mvc/app/views/OPNsense/Interface/assignment.volt
```diff
@@ -0,0 +1,54 @@
...
+            {   search:'/api/interfaces/assignment/search_item/',
+                get:'/api/interfaces/assignment/get_item/',
+                set:'/api/interfaces/assignment/set_item/',
+                add:'/api/interfaces/assignment/add_item/',
+                del:'/api/interfaces/assignment/del_item/',
+                options: {
+                    formatters: {
+                        statusformatter: function (column, row) {
+                            console.log(row);
+                            console.log(column);
```
swhite2: :)

AdSchellevis: 🙃

src/opnsense/mvc/app/models/OPNsense/Interfaces/FieldTypes/DeviceField.php
```diff
@@ -0,0 +1,73 @@
...
+        return parent::actionPostLoadingEvent();
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function setValue($value)
+    {
+        $this->getParentNode()->icon = 'fa fa-plug text-muted';
+        if (isset(static::$interfaces[$value])) {
```
swhite2, line 65: This is never set as far as I can see, all interfaces default to `text-muted`

fichtner:
Just some post-merge observations:

* the controller should fill a missing description, my `WAN` and `LAN` labels are not showing in the grid.
* not sure if we still have the auto-fill logic of the description of e.g. VLANs
* "Interface" column is the device column ;)
* "Prevent interface removal" is not honoured anymore, at least not from the Grid (trash button shows, but I don't want to test it from here)
* Maybe we should remove the clone feature? At least for pure assignments it's difficult to make it useful

sopex:
Points 1 and 3 were suggested in a previous comment by me, and can also be seen here:
https://github.com/sopex/core/commit/2ff695e9d6f8fecb7f3a00ac428a3ad6d52b181c 

Feel free to just use it.

From a UI perspective, the trash icon would be better if it remained and produced an error message in case the interface is locked.

AdSchellevis:
> Just some post-merge observations:
> 
> * the controller should fill a missing description, my `WAN` and `LAN` labels are not showing in the grid.

I do need to merge these defaults on load indeed, good point, thanks

> * not sure if we still have the auto-fill logic of the description of e.g. VLANs

nope, but also a bit difficult given the dropdown select nature

> * "Interface" column is the device column ;)

check, I'll replace that, forgot about that one it seems

> * "Prevent interface removal" is not honoured anymore, at least not from the Grid (trash button shows, but I don't want to test it from here)

Should work, there's an exception. you can safely drop and `/tmp/.interfaces.todo` as this isn't persistent until apply is hit.

> * Maybe we should remove the clone feature? At least for pure assignments it's difficult to make it useful

good point

fichtner:
Opening PRs is ok now that it's in master. Showing the identifier would help (and I think the old page also did that).

> Points 1 and 3 were suggested in a previous comment by me

No, point 1 is a controller thing since the framework always extrapolated a description from the strtoupper(identifier) if not set. Point 3 is a different one also as it's about the column name which is wrong.

> From a UI perspective, the trash icon would be better if it remained and produced an error message in case the interface is locked.

Sure, works for me too.

fichtner:
> > * "Prevent interface removal" is not honoured anymore, at least not from the Grid (trash button shows, but I don't want to test it from here)
> 
> Should work, there's an exception. you can safely drop and `/tmp/.interfaces.todo` as this isn't persistent until apply is hit.

Ok, good. Small side question: shouldn't we use `/var/lib/php/tmp` or find some other way to transport the information to the backend for priv separation?

AdSchellevis:
> Maybe we should remove the clone feature? At least for pure assignments it's difficult to make it useful

As I'm planning to move configuration settings as well, I expect the button would be useful again soon, so for consistency it might be better to keep it for now.

> Ok, good. Small side question: shouldn't we use /var/lib/php/tmp or find some other way to transport the information to the backend for priv separation?

Was doubting about that, but since backend only reads, it's less problematic to keep it in `/tmp`, but also don't have a very strong feeling about the location.

Issue #10196, opened 2026-04-23
Interfaces: Neighbors: Automatic Discovery: Disabled Banner shows when search yields no results


Monviech:
**Important notices**

Before you add a new report, we ask you kindly to acknowledge the following:

- [x] I have read the contributing guidelines at https://github.com/opnsense/core/blob/master/CONTRIBUTING.md
- [x] I am convinced that my issue is new after having checked both open and closed issues at https://github.com/opnsense/core/issues?q=is%3Aissue

**Describe the bug**

"Host discovery service is disabled, below the hosts currently known by this firewall via ARP and NDP"
is shown if no valid search result is found.

**To Reproduce**

1. Go to 'Interfaces: Neighbors: Automatic Discovery'
2. Type in a string into "Search" that yields no results
3. "Host discovery service is disabled, below the hosts currently known by this firewall via ARP and NDP" will appear

**Expected behavior**

The banner should not appear for no results.

**Describe alternatives you considered**

A clear and concise description of any alternative solutions or workaround you considered.

**Screenshots**

**Relevant log files**

If applicable, information from log files supporting your claim.

**Additional context**

**Environment**

Software version used and hardware type if relevant, e.g.:

26.7.a_475-amd64

fichtner:
Thanks for putting this in writing. I remember seeing this too.

PR #10368, opened 2026-06-01
hostwatch: pin warning banner to enabled flag

swhite2:
**Important notices**

- Model used:
- Extent of AI involvement:

---

**Describe the problem**

Banner stating ARP or NDP is used to fetch hosts also shows when no results are found (e.g. due to a search query). This is because at least one row is required to determine discovery source.

---

Easiest solution is to pin it on the `enabled` flag as is done in this PR. Alternatively, we can add a column showing the source, pinning it on the data as was the intent before.

---

**Related issue**

https://github.com/opnsense/core/issues/10196

### DIFF
--- a/src/opnsense/mvc/app/controllers/OPNsense/Hostdiscovery/Api/ServiceController.php
+++ b/src/opnsense/mvc/app/controllers/OPNsense/Hostdiscovery/Api/ServiceController.php
@@ -59,8 +59,8 @@ class ServiceController extends ApiMutableServiceControllerBase
                 'ether_address' => $rec[1],
                 'ip_address' => $rec[2],
                 'organization_name' => $rec[3],
-                'first_seen' => $rec[4] ?? '',
-                'last_seen' => $rec[5] ?? ''
+                'first_seen' => $rec[4],
+                'last_seen' => $rec[5]
             ];
         }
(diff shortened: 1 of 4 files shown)


src/opnsense/mvc/app/controllers/OPNsense/Hostdiscovery/Api/ServiceController.php
```diff
@@ -59,8 +59,8 @@ public function searchAction()
                 'ether_address' => $rec[1],
                 'ip_address' => $rec[2],
                 'organization_name' => $rec[3],
-                'first_seen' => $rec[4],
-                'last_seen' => $rec[5]
```
AdSchellevis, line 63: it's probably better to fix this in the script that collects the output:

https://github.com/opnsense/core/blob/d0f954ff3988e8855591da48cafe00aabe40d5a9/src/opnsense/scripts/interfaces/list_hosts.py#L109

https://github.com/opnsense/core/blob/d0f954ff3988e8855591da48cafe00aabe40d5a9/src/opnsense/scripts/interfaces/list_hosts.py#L118

I think I just missed that there earlier and it makes sense to always output the same structure no matter the underlaying source system.

PR #10372, opened 2026-06-01
Firewall: Rules: initiate tree expansion based on interface selection value

### LINKED ISSUE #10358: Firewall: Rules [new]: Evaluate tree collapse/expand and persistence
**Important notices**

Follow up to this PR:
https://github.com/opnsense/core/pull/10355

This comment defines different expected behavior for collapse/expand and persistence of tree rows:
https://github.com/opnsense/core/pull/10355#issuecomment-4564447453

Not changing persistence handling, and not automatically open and collapse tree rows.

Right now there is a nifty "Expand/Collapse All" button that handles this, and persistence ensures that during pagination or reloads of the page, colums are expanded or collapsed consistent to their persisted browser storage state.

E.g., if the interface column has been expanded, it will currently stay expanded even if you change the interface filter.

**Additional context**

swhite2:
**Important notices**

- Model used:
- Extent of AI involvement:

---

**Describe the problem**

See discussion in https://github.com/opnsense/core/issues/10358

---

If you now select an interface, the `interface rules` category tree will automatically open, if you select a group, the `group` tree will open. First-level groups (`interface rules`, `groups`, `floating`, `automatically generated`) are not persisted, while categories (user-defined) are.

This means that if you expand your own category, reloading the page will keep the expansion, but the first-level group where this category lives in may or may not render expanded, depending on interface type selection.

If you select "All rules", all first-level categories will expand.

---

**Related issue**

https://github.com/opnsense/core/issues/10358

### DIFF
--- a/src/etc/inc/filter.lib.inc
+++ b/src/etc/inc/filter.lib.inc
@@ -623,7 +623,7 @@ function filter_core_rules_system($fw, $defaults)
                                 // only try to add gateway rules for traffic leaving this interface
                                 // when the correct protocol is assigned to the interface
                                 $fw->registerFilterRule(
-                                    100000,
+                                    10000,
                                     ['from' => "({$ifcfg['if']})", 'direction' => 'out', 'gateway' => $gwname,
                                     'destination' => ['network' => $ifdescr, "not" => true],
                                     'statetype' => 'keep',
--- a/src/opnsense/mvc/app/views/OPNsense/Firewall/filter_rule.volt
+++ b/src/opnsense/mvc/app/views/OPNsense/Firewall/filter_rule.volt
@@ -66,14 +66,23 @@
 
         // read interface from URL hash once, for the first grid load
         let pendingUrlInterface = getUrlHash('interface') || null;
+        let previousGroupType = null;
+        let currentGroupType = null;
 
-        const ruleTypeMap = {
-            '0': { uuid: "auto0", label: "{{ lang._('Automatically generated rules') }}", icon: "fa-magic", tooltip: "{{ lang._('Automatically generated rules') }}", color: "text-secondary", groupType: null },
-            '2': { uuid: "floating", label: "{{ lang._('Floating rules') }}", icon: "fa-layer-group", tooltip: "{{ lang._('Floating rule') }}", color: "text-primary", groupType: "floating" },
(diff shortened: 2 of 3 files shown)


fichtner (commented):
not qualified to review but I'll try it once it is merged

swhite2:
One side effect of this change is that the automatically generated rule category at the top now appears twice:

afbeelding

Previously these two groups were merged in the frontend, which we can still do here, but the question is if it should be fixed on the backend instead: https://github.com/opnsense/core/blob/master/src/opnsense/scripts/filter/list_non_mvc_rules.php#L120 and merge the two into a single prefix (0 or 1).

fichtner:
Aren’t these the “tailing” automatic rules? Looks visually problematic, but when ordered by sequence it makes sense.

AdSchellevis:
I think your problem originates from:

https://github.com/opnsense/core/blob/99697111d694fdc470fcd3375b0019d3d1743e94/src/etc/inc/filter.lib.inc#L625-L626

Looking at the code, I expect the chosen number is just a bit too high, when changing this to `10000` it should still be at exactly the same spot, but inside the container. 


Between `10000` and `100000` nothing seems to register at the moment anyway.

swhite2:
> I think your problem originates from:
> 
> https://github.com/opnsense/core/blob/99697111d694fdc470fcd3375b0019d3d1743e94/src/etc/inc/filter.lib.inc#L625-L626
> 
> Looking at the code, I expect the chosen number is just a bit too high, when changing this to `10000` it should still be at exactly the same spot, but inside the container.
> 
> Between `10000` and `100000` nothing seems to register at the moment anyway.

Nice, exactly what was needed

fichtner (approved):
looks good to me, but I would suggest having automatic rules always collapsed... it flushes the actual user rules out of the screen for no immediate benefit

swhite2:
> looks good to me, but I would suggest having automatic rules always collapsed... it flushes the actual user rules out of the screen for no immediate benefit

Currently it's always collapsed except for the `All Rules` case, but I'll adjust to this :)

Monviech (approved):
For me this works and looks good, approved, nice work

AdSchellevis (approved):
@swhite2 as discussed, eventually it would be nice if we could embed an optional bucket function in `searchRecordsetBase()` to ease some of the magic, but let's try to finish this version first.

PR #10371, opened 2026-06-01
Firewall: NAT: download/upload rules as csv

Monviech:
**Important notices**

Before you submit a pull request, we ask you kindly to acknowledge the following:

- [x] I have read the contributing guidelines at https://github.com/opnsense/core/blob/master/CONTRIBUTING.md
- [x] I opened an issue first for non-trivial changes and linked it below.
- [ ] AI tools were used to create at least part of the code submitted herewith.

If AI was used, please disclose:

- Model used:
- Extent of AI involvement:

---

**Describe the problem**

NAT rules should have the same download/upload actions available as firewall rules currently do.

---

**Describe the proposed solution**

Move downloadRules and uploadRules into FilterBaseController and use them in all MVC Filter/NAT controllers.

---

**Related issue**

Required for: https://github.com/opnsense/core/issues/9515

PR #10373, opened 2026-06-02
Firewall: merge outbound NAT into SNAT page

### LINKED ISSUE #9515: firewall: merge outbound NAT into source NAT page
**Important notices**

Related to: https://github.com/opnsense/core/pull/9507

For the Source NAT a few things would be needed:

- [ ] The controller could emit the automatic rules inside the searchRuleAction() so they are visible in the frontend
- [ ] Some endpoint to change the Source NAT modes: https://github.com/opnsense/core/blob/559701cbc0df0c39abaf85102eca098164fe575d/src/www/firewall_nat_out.php#L303
- [ ] Should outbound NAT entries be migrated into source NAT or the other way around?

The goal is to remove the more or less duplicated static PHP page.

Monviech:
**Important notices**

- Model used: ChatGPT 5.5
- Extent of AI involvement: Code autocompletion usage via Codex

---

**Describe the problem**

Fixes: https://github.com/opnsense/core/issues/9515

For the Source NAT a few things would be needed:

- [x] The controller could emit the automatic rules inside the searchRuleAction() so they are visible in the frontend
    - [x] Hide either automatic, or manual rules from what searchRuleAction returns, depending on snat_mode
    - [x] when NAT mode is disabled hide grid completely 
- [x] Some endpoint to change the Source NAT modes
- [x] No automatic migration, offer export/import of rules
    - [x] Import/Export of rules in MVC NAT pages: https://github.com/opnsense/core/pull/10371
    - [x] Export of rules in Legacy Outbound page
    - [x] Reuse existing Migration Assistant
- [x] Hide legacy Outbound page when no rules exist in it, or outbound NAT is disabled/automatic


src/opnsense/mvc/app/controllers/OPNsense/Firewall/Api/MigrationController.php
```diff
@@ -44,12 +46,32 @@ public function downloadRulesAction()
 
     public function flushAction()
     {
-        if ($this->request->isPost()) {
+        if ($this->request->isPost() && !(new ACL())->hasPrivilege($this->getUserName(), 'user-config-readonly')) {
```
AdSchellevis: @Monviech nice catch, let's uniform this in https://github.com/opnsense/core/commit/ce423ee29079cff9bba5064a50df3d811c163ed6

src/opnsense/mvc/app/library/OPNsense/Core/ConfigMaintenance.php
```diff
@@ -41,6 +41,8 @@ public function __construct()
             'dhcpd' => gettext('Services: ISC DHCPv4 [legacy]'),
             'dhcpdv6' => gettext('Services: ISC DHCPv6 [legacy]'),
             'filter.rule' => gettext('Firewall: Rules [legacy]'),
+            // XXX: NAT Outbound is currently noop, the nested node is never reached
```
AdSchellevis: @Monviech not sure what you mean by this, it was missing or is still not fully correct?

src/opnsense/mvc/app/models/OPNsense/Firewall/Filter.php
```diff
@@ -395,4 +396,22 @@ public function hasSchedule()
...
     }
+
+    /**
+     * Persist volatile SNAT model fields into legacy configuration nodes.
+     */
+    public function serializeToConfig($validateFullModel = false, $disable_validation = false)
+    {
+        $result = parent::serializeToConfig($validateFullModel, $disable_validation);
+        $mode = $this->general->snat_mode->getValue();
+        if (!SNatModeField::isValidMode($mode)) {
```
AdSchellevis: ok, this is pretty weird, when requesting the model to sync, we should be able to trust our data is properly validated already. 

Testing this assumption that we can inject arbitrary data does seem to offer the expected results:

In:
```
{"filter":{"general":{"snat_mode":"xadvanced"}}}
```

Out:
```
{
    "result": "failed",
    "validations": {
        "filter.general.snat_mode": "Option [] not in list.",
        "filter.npt.rule.b15006a5-68c4-4838-a15e-ac8112477a56.trackif": "This interface is not tracking the current rule interface."
    }
}
```

Going to ditch this

src/opnsense/mvc/app/library/OPNsense/Core/ConfigMaintenance.php
Monviech: @AdSchellevis I added this here and it doesnt work. When triggering it the "rule" nodes are not deleted.

AdSchellevis (approved):
@Monviech I've added some small fixes on top, but I think it's complete now. Thanks! 

Only missing point is the validation part on the apply button, which might trigger on other validation issues in the model than the one field we are trying to set here. 

If you could double check the end result, I'm fine with merging this to master.

src/opnsense/mvc/app/library/OPNsense/Core/ConfigMaintenance.php
AdSchellevis: @Monviech ok, thanks, so the reset is currently non-functional, haven't tried that yet, but can put this on my todo list.

src/opnsense/mvc/app/models/OPNsense/Firewall/Filter.php
Monviech: You are correct, its a left over assumption from when I saved the instance directly in the controller in an earlier version of this branch. Sorry :)

src/opnsense/mvc/app/library/OPNsense/Core/ConfigMaintenance.php
Monviech: yeah, thanks for looking into it

src/opnsense/mvc/app/models/OPNsense/Firewall/Filter.php
AdSchellevis: don't be sorry :) we're reviewing each others work for good reasons and all make mistakes

src/opnsense/mvc/app/library/OPNsense/Core/ConfigMaintenance.php
AdSchellevis: ok, that's odd, if I flush the section in System/Configuration/Defaults I do seem to endup with the expected output:

```
    <outbound>
      <mode>hybrid</mode>
    </outbound>
```

Monviech:
@AdSchellevis I added a scoped setGeneralAction, is that what you had in mind?

AdSchellevis:
@Monviech almost, the validation part is missing, but conceptually this would be it indeed :)

src/opnsense/mvc/app/library/OPNsense/Core/ConfigMaintenance.php
Monviech: I just tried it as well, and it also works for me now for some reason. Maybe my code was in an inconsistent state.

AdSchellevis: ok, great, another one off the list :)

Monviech:
I used the same pattern as here now:
https://github.com/opnsense/core/blob/1e4ffcbc0f49ec0254efd3f05d0a47b264a9a31d/src/opnsense/mvc/app/controllers/OPNsense/Base/ApiMutableModelControllerBase.php#L369

Validation works:
image

PR #10391, opened 2026-06-08
Firewall: NAT: Source NAT: Hide command footer if snat_mode is automatic

Monviech:
**Important notices**

- Model used:
- Extent of AI involvement:

---

**Describe the problem**

When snat_mode is automatic, users are not expected to add rules or use any of the footer commands.

---

Hide them via css.

Theoretically it could also be done with the command filters in the grid, but the grid loads asynchronous, so the endpoint we use to decide this might not have fired yet and we would have to wrap everything into promises.

---

**Related issue**


swhite2 (approved):
Normally one would dictate this from the controller, but the footer is rendered during data load, so we can't know the state at the right time without bogging down the UI. This is the better approach indeed

PR #10396, opened 2026-06-09
Firewall: Rules: add banner if no rules defined

swhite2:
**Important notices**

- Model used:
- Extent of AI involvement:

---

**Describe the problem**

In the current firewall rules page, it isn't immediately obvious there are no firewall rules defined, as the floating and automatically generated sections are always rendered.

---

Show a message indicating all traffic will be blocked on this interface until a pass rule is applied, same as the legacy firewall pages:

image

---

**Related issue**

### DIFF
--- a/src/opnsense/mvc/app/views/OPNsense/Firewall/filter_rule.volt
+++ b/src/opnsense/mvc/app/views/OPNsense/Firewall/filter_rule.volt
@@ -105,9 +105,6 @@
         }
 
         function responseHandler(response) {
-            // remove empty ruleset message banner if applicable
-            $('#{{formGridFilterRule["table_id"]}}-banner').remove();
-
             // recursively clear children but keep buckets intact
             const clear = (buckets) => {
                 for (const bucket of buckets) {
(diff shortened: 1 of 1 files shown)


src/opnsense/mvc/app/views/OPNsense/Firewall/filter_rule.volt
```diff
@@ -866,6 +869,26 @@
...
+                let ifLabel = '';
+                for (const group of Object.values($('#interface_select').data().store)) {
+                    const item = group.items.find(item => item.value === $('#interface_select').val());
+                    if (item) {
+                        ifLabel = item.label;
+                        break;
+                    }
+                }
+                $("#{{formGridFilterRule['table_id']}}-header")
+                    .after($('<div id="{{formGridFilterRule["table_id"]}}-banner" class="alert alert-info text-center __ml __mr" style="font-weight: 400;">')
```
Monviech, line 884: Would it be better to add a hidden class here, and remove that class if the banner should be shown?

I'm unsure if the banner could flash into existence and vanish afterwards on slower hardware.

swhite2: The banner only renders when the grid has fully loaded and it matches the criteria of missing rules on either a group or an interface.

With my last commit I've put the banner removal and creation closer together, which should prevent any flashing at all (and only when going from interface without rules to another interface without rules).

Monviech (approved):
Looks good to me, that's something that definitely helps users. Thank you.

Issue #10208, opened 2026-04-25
Kea: permit reservation to be created without "ip address" field


whezzel:
**Important notices**

Before you add a new report, we ask you kindly to acknowledge the following:

- [x] I have read the contributing guidelines at https://github.com/opnsense/core/blob/master/CONTRIBUTING.md
- [x] I am convinced that my issue is new after having checked both open and closed issues at https://github.com/opnsense/core/issues?q=is%3Aissue

**Is your feature request related to a problem? Please describe.**

The DHCP reservation form (Services -> Kea DHCP -> Kea DHCPv4 -> Reservations; add new reservation) prevents the creation of new reservations if the IP address field is not set. In ISC DHCP, the IP address field is optional, and since Kea is poised to replace ISC DHCP, it should match the functionality where possible.

Per Kea documentation, this is a supported configuration for reservations (https://kb.isc.org/docs/what-are-host-reservations-how-to-use-them#example-addresses-prefixes)

>Every host reservation must reserve at least one resource; if not, Kea complains at configure time. Resources include address, prefixes, options, hostname, and client classes.

**Describe the solution you like**

I would like to be able to create reservations without being required to set an IP address. This would primarily be used to set a specific hostname on a device that does not return the correct hostname, without setting a specific IP on that device.

**Describe alternatives you considered**

- Create a DHCP reservation with an IP address configured.
- Manually configure Kea and set reservations directly in the config files without using the Web UI.

**Additional context**

Add any other context or screenshots about the feature request here or links to relevant forum thread or similar

Monviech:
Why not reserve both hostname and IP? Essentially that should not matter much.

I'm just going to push back a little at making every field optional because I did the host reservation in dnsmasq and there anything is very optional, in the end confusing the user more than it helps and inflating the complexity of the template, model and validations.

Right now I have a PR open where you can decide between IP address and prefix for a reservation, though having some IP address as primitive makes sense here.

Also the lease reservation logic compares duid and mac address to existing reservations, if one of them only has a hostname it would also show as static right now even though no IP address exists.

All in all it's a bit challenging not make this bad from a GUI perspective I'm afraid.

From a model perspective a hostname is also no resource a DHCP server can allocate, even though KEA allows it its more of an override. A DHCP server should allocate resources as IP addresses.

Please give one or two real usecases (not just festure parity with ISC) that is not possible without this functionality.

whezzel:
[...]

PR #10401, opened 2026-06-10
Implement deny unknown and permit reservation without IP.

alexwbaule:
Add a "know" class, when the ip is given by the reservation, we can accept only know class client.
Permit add reservation without fixed IP. So you can reserve only "MACADDR" to be used with "deny unknow class".
It can be used together or not.

**Important notices**

---

**Describe the problem**

There are two related missing features in the Kea DHCPv4 implementation:

1. There is no way to restrict a pool to only serve clients that have a reservation (deny unknown clients). This is a common requirement for networks where only pre-approved devices should receive an IP from a specific pool.

2. Reservations currently require a fixed IP address. This prevents creating MAC-only reservations, which are needed when using "deny unknown clients" — you want to allow a client into the pool without pinning it to a specific address.

---

**Feature 1: Deny unknown clients per subnet**

Added a `deny_unknown_clients` boolean field to the subnet4 model. When enabled, all pools in that subnet are generated with `"client-class": "KNOWN"` in the Kea configuration. The `KNOWN` class is a Kea built-in that matches only clients with an existing host reservation, effectively denying unknown clients from obtaining a lease from the pool.

**Feature 2: Reservations without fixed IP**

Removed the requirement for `ip_address` on DHCPv4 reservations. The `UniqueConstraint` is preserved but naturally skips validation when the field is empty (built-in framework behavior for non-required fields). The validation in `performValidation()` now only checks subnet membership when an IP is actually provided. Config generation already handles this correctly since it only emits `ip-address` when the field is non-empty.

These two features work together: you can enable "deny unknown clients" on a subnet and create MAC-only reservations to whitelist devices without assigning them a static IP — they will receive a dynamic address from the pool instead.

---

**Related issue**
 #10208

### DIFF
--- a/src/opnsense/mvc/app/controllers/OPNsense/Kea/forms/dialogReservation4.xml
+++ b/src/opnsense/mvc/app/controllers/OPNsense/Kea/forms/dialogReservation4.xml
@@ -9,7 +9,7 @@
         <id>reservation.ip_address</id>
         <label>IP address</label>
         <type>text</type>
-        <help>IP address to offer to the client</help>
+        <help>IP address to offer to the client. Leave empty to create a reservation without a fixed IP (useful for deny unknown clients).</help>
     </field>
     <field>
         <id>reservation.hw_address</id>
--- a/src/opnsense/mvc/app/controllers/OPNsense/Kea/forms/dialogSubnet4.xml
+++ b/src/opnsense/mvc/app/controllers/OPNsense/Kea/forms/dialogSubnet4.xml
@@ -28,6 +28,17 @@
         <help>List of pools, one per line in range or subnet format (e.g. 192.168.0.100 - 192.168.0.200 , 192.0.2.64/26</help>
     </field>
     <field>
+        <id>subnet4.deny_unknown_clients</id>
+        <label>Deny unknown clients</label>
+        <type>checkbox</type>
+        <help>When enabled, only clients with a reservation will receive an IP from the pool. Unknown clients will be denied.</help>
+        <grid_view>
+            <type>boolean</type>
+            <formatter>boolean</formatter>
+            <visible>false</visible>
+        </grid_view>
+    </field>
+    <field>
         <id>subnet4.valid_lifetime</id>
         <label>Valid lifetime</label>
         <type>text</type>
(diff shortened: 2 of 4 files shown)


Monviech:
I don't like removing the IP address as required primitive from the reservation.

Deny unknown clients already works when leaving the pool in a subnet empty without a new option added. Since you have to add reservations too for clients to get an IP address its essentially the same thing.

Adding an IP address and Mac address to a reservation added to a subnet with an empty pool should be a reasonable GUI requirement.

alexwbaule:
hi @Monviech 

I think we have 2 distinct behaviors regarding "denying unknown clients". One, if I create only ONE subnet, with empty configs, it's ok. However, we have configurations with more than one subnet, which will not work.
Regarding the requirement to set an IP when creating a reservation, in OPNSense with ISCP DHCPD this was possible, and I use it a lot. I don't want a fixed IP for all devices, I only want them to be known and have an IP.
With this I can use "denying unknown clients", and take advantage of DNS to resolve IP vs Name.
Kea allows, without any problem, setting only the macaddr, without IP. So much so that if I manually change /conf/config.xml and remove the IPs, it works.
I believe that many people would use both configurations from this PR.
We already have some people needing this in #10208, and I, including myself, a long time ago, when the migration from DHCPD to KEA was done, questioned this.

I think both implementations are valid and very useful.

I understand your point of view regarding the GUI, but in relation to what we had with DHCPD, it's a regression.

Monviech:
Hello,

I understand your position and that there are usecases for it.

I know that KEA allows a reservation to be based on any parameter, but what we expose in the GUI is modeled around "Host Reservations"

https://kea.readthedocs.io/en/stable/arm/dhcp4-srv.html#host-reservations-in-dhcpv4

KEA defines these as a combination of hw-address/duid/curcuit-id/client-id, ip-address and optionally hostname etc...

I understand that there are different models available, e.g. the "Reserving a Hostname" model:
https://kea.readthedocs.io/en/stable/arm/dhcp4-srv.html#reserving-a-hostname

Which does not require an ip address.

In your own personal usecase, you want to use it for a MAC whitelisting model, combined with a client-class.

Overall it's not about what KEA allows or not, it's which model primitive we offer here for most users, and right now it's host reservations.

I leave this open, I'm not blocking this, this just needs more opinions.

alexwbaule:
it's a good idea.

we can ask more people to get more opinions.

L1ghtn1ng:
I definitely would like this ability, it's an additional layer of security for defense in depth

Monviech:
Not right now, Kea is complex enough already in its config generation. And now reservations and client-classes got a bit more intricate via the flex hook framework here that was a requirement to get some container DHCP options working.
https://github.com/opnsense/core/commit/1c2b98a92b1e07dc822a45b47afbf9be1e401fd2

I'll close this, thank you for the suggestion.

PR #10407, opened 2026-06-11
ipsec: connections small ui improvment

sopex:
**Important notices**

**Describe the problem**
@swhite2 

Screenshot 2026-06-11 124318
Screenshot 2026-06-11 124236

**Related issue**

https://forum.opnsense.org/index.php?topic=52111.0

### DIFF
--- a/src/opnsense/mvc/app/views/OPNsense/IPsec/connections.volt
+++ b/src/opnsense/mvc/app/views/OPNsense/IPsec/connections.volt
@@ -59,6 +59,7 @@
             toggle:'/api/ipsec/connections/toggle_' + obj_type + '/',
             options:{
                 static: true,
+                disableScroll: true,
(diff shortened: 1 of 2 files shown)


fichtner:
I'll leave this for @swhite2 to assess. The 2 row issue pinch is the same problem on mobile landscape view for every grid and we can't set each table to disable scroll then

swhite2:
I think this change fits for the ipsec page, but the issue @fichtner pointed out remains. It's possible a minimum set height may fix this, but it has to be dynamic as we don't want dead space in the grid space.

Will test this later today

sopex:
I think this is fine here for multiple reasons and maybe minVisibleRows = 3 or 4 should be enforced on all other ocasions

swhite2 (approved):
Sure, I don't have objections here

PR #10408, opened 2026-06-11
ui: force tables to display 10 rows

sopex:
**Important notices**

**Describe the problem**
Phones and 4K screens make it very difficult for pixel-based minimums to work. A minimum amount of rows displayed is the way to go.

@fichtner Tagging you as the perfect tester :)
Expanding from here:
https://github.com/opnsense/core/pull/10407

### DIFF
--- a/src/opnsense/www/js/opnsense_bootgrid.js
+++ b/src/opnsense/www/js/opnsense_bootgrid.js
@@ -455,6 +455,7 @@ class UIBootgrid {
 
         if (bootGridOptions?.disableScroll ?? false) {
             this.options.disableScroll = true;
+            this.compatOptions['height'] = false;
         }
 
         this.tabulatorOptions = compatOptions.tabulatorOptions ??= {};
--- a/src/opnsense/www/js/opnsense_bootgrid.js
+++ b/src/opnsense/www/js/opnsense_bootgrid.js
@@ -128,6 +128,7 @@ class UIBootgrid {
         this.rememberedTreeIds = new Set(JSON.parse(localStorage.getItem(this.treeStorageKey) || '[]'));
         this.isVisible = false;
         this.scrollPos = 0;
+        this._suppressScrollPosUpdate = false;
         this.addButton = false;
         this.deleteSelectedButton = false;
 
@@ -652,10 +653,10 @@ class UIBootgrid {
         }
     }
 
-    _onDimensionChange() {
-        if (this._isScrolling) {
+    _onDimensionChange(force = false) {
+        if (!force && this._isScrolling) {
             clearTimeout(this._resizeAfterScrollTimer);
-            this._resizeAfterScrollTimer = setTimeout(() => this._onDimensionChange(), 150);
+            this._resizeAfterScrollTimer = setTimeout(() => this._onDimensionChange(force), 150);
             return;
         }
 
@@ -700,9 +701,13 @@ class UIBootgrid {
 
         if (Math.abs(currentTotalHeight - nextHeight) > 1) {
             this._suppressTableObserver = true;
+            this._suppressScrollPosUpdate = true;
+            const prevScrollPos = this.scrollPos;
             this.table.setHeight(nextHeight);
             this.table.redraw();
+            this._maintainScrollPosition(prevScrollPos);
             this._suppressTableObserver = false;
+            this._suppressScrollPosUpdate = false;
         }
     }
 
@@ -868,7 +873,7 @@ class UIBootgrid {
                         if (this.options.bottomReserveElement) {
                             this.pageHeight -= this.options.bottomReserveElement.getBoundingClientRect().height;
                         }
-                        this._onDimensionChange();
+                        this._onDimensionChange(true);
                     }
                 }));
                 pageObserver.observe(pageTarget);
@@ -902,6 +907,7 @@ class UIBootgrid {
         });
 
         this.table.on('scrollVertical', (top) => {
+            if (this._suppressScrollPosUpdate) return;
             this.scrollPos = top;
             this._isScrolling = true;
             clearTimeout(this._scrollEndTimer);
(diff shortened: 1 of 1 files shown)


fichtner:
It would be nicer to let @swhite2 take the lead. And 3 is as bad as 2 IMO.

sopex:
Of course.

In any case, in practice the jump is from 1.5 to 3.5 rows which is probably enough.

fichtner:
I'm leaning towards 5 or 10 if enough entries exist. As I said on mobile landscape it's rather hard to get a good overview and hitting the scroll position too then actually scroll again (still without better overview). The grids can have hundreds of entries so scrolling 2 of them at a time is weird.

src/opnsense/www/js/opnsense_bootgrid.js
```diff
@@ -665,12 +665,30 @@ class UIBootgrid {
 
         let nextContentHeight;
 
+        let minHeightLimit = defaultHeight;
+        if (this.tableInitialized) {
```
swhite2: `tableInitialized` can be true before any data is loaded, so there's no need for it here. `table.getRows()` will either be an empty array if the data hasn't loaded yet, or contain all rows if loaded - it doesn't affect the height calculation, rather the height calculation should be triggered at the right time.

src/opnsense/www/js/opnsense_bootgrid.js
```diff
@@ -665,12 +665,30 @@ class UIBootgrid {
...
+            const rows = this.table.getRows();
+            if (rows.length > 0) {
+                let rowsHeight = 0;
+                const limit = Math.min(rows.length, 3);
```
swhite2: Let's do 10 rows by default

sopex: Sure

src/opnsense/www/js/opnsense_bootgrid.js
sopex: Like this?

swhite2: Yep

swhite2:
@sopex testing this I see 2 issues:

- getRows() should be recursive, as data rows may be nested on the `dataTreeChildField` property, see filter_rule.volt.
- I see scrolling issues if there is too little space. Go to Aliases -> Diagnostics -> Bogons (just a large table), drag the developer tools all the way up until there is a scrollbar in both the page and the grid, then scroll the grid, it seems the page scrollbar is reset to the top with each scroll attempt.

sopex:
@swhite2 Done

fichtner:
When zooming in or reducing vertical space I still end up with the 2 entries.

sopex:
> When zooming in or reducing vertical space I still end up with the 2 entries.

On the phone?

fichtner:
> On the phone?

Tested only in browser for now but this should apply to all things.

Now it works, but the irony is on larger lists it's harder to find a scrollable background to get to the apply button hidden below ;)

sopex:
@fichtner My bad. I was being overzealous in trying to keep the page from scrolling on 1080p computers. But in combination with https://github.com/opnsense/core/pull/10409, this is a not an issue anymore.

Try it with #10409

sopex:
> Now it works, but the irony is on larger lists it's harder to find a scrollable background to get to the apply button hidden below ;)

Maybe some weird aspect ratio? You can always use the footer, the head bar, or the big menu bar :)

fichtner:
Good point. I avoided the footer thinking it doesn't work, but it does. Ok for me.

swhite2:
@sopex One glitch I see now is that when a grid is squashed (as always simulated on my end though dragging the developer tools up), and you scroll the grid down a bit, then create more space for the grid by increasing page height again, the grid doesn't scale with it anymore. As far as I can see the grid prioritizes resetting the grid scroll to the top, which shouldn't happen in this case. A second initiated resize will then scale the grid properly (because the scroll position is at the top).

Sorry for being annoying :)

sopex:
@swhite2 You need to try harder to annoy me. It's a bit of a unique problem, but all good :)

sopex:
Edit: I need your help afterall, this is unstable. It doesn't work consistently.

@swhite2 If you have a better idea, otherwise this seems good.

https://github.com/user-attachments/assets/2221b9a8-8667-4941-8987-dd0cacad32a9

fichtner:
Maybe it's already taken care of but I want to mention it -- on master the current resize still accounts for `page-content-main`?

Screenshot 2026-06-12 at 12 28 13

swhite2:
@sopex what is the unstable part? It works rather well now

sopex:
> @sopex what is the unstable part? It works rather well now

If it's fine on your system. Maybe it's time to kill my VM :)
Or maybe I have some weird browser caching.

swhite2:
can you rebase this branch on master (and force push)? some git weirdness here

sopex:
> can you rebase this branch on master (and force push)? some git weirdness here

Absolutely

swhite2:
For some reason the ipsec change from yesterday doesn't apply anymore?

sopex:
> For some reason the ipsec change from yesterday doesn't apply anymore?

Rebase issue, fixed

src/opnsense/www/js/opnsense_bootgrid.js
```diff
@@ -666,19 +673,41 @@ class UIBootgrid {
 
         let nextContentHeight;
 
+        let minHeightLimit = defaultHeight;
+        const rows = this.table.getRows("visible");
```
swhite2, line 677: Clever

swhite2:
@sopex There seem to be some functional issues remaining, but they are getting more complex to explain. I like the end result but not 100% sure about the code. If you don't mind I'll fork off this PR to keep your contribution intact but not merge the current state into master

src/opnsense/www/js/opnsense_bootgrid.js
sopex: Thanks :)

sopex:
@swhite2 Do what you have to do, no worries. Interested in seeing your implementation.

swhite2:
@sopex I've added https://github.com/opnsense/core/commit/87cb2d055dd0cb7e07ab30e8656c020a82803a91 which should to the same. Unfortunately the rebase would've been a bit messy so I didn't base it off of this PR - sorry about that.

sopex:
@swhite2 

I am here for the love of the game and because I like the project. I am not trying to impress anyone with my commits 😂 So don't worry about it.

Issue #9955, opened 2026-03-13
ui: form modal search field


fichtner:
**Important notices**

Before you add a new report, we ask you kindly to acknowledge the following:

- [x] I have read the contributing guide lines at https://github.com/opnsense/core/blob/master/CONTRIBUTING.md
- [x] I am convinced that my issue is new after having checked both open and closed issues at https://github.com/opnsense/core/issues?q=is%3Aissue

**Is your feature request related to a problem? Please describe.**

OpenVPN and firewall forms are long and sometimes it's hard to find the right option also because they may be hidden under advanced.

**Describe the solution you like**

A form search would be nice to have in these cases hiding irrelevant items.

**Describe alternatives you considered**

N/A

**Additional context**

N/A

swhite2:
Implemented in https://github.com/opnsense/core/commit/e5e5c86114f319e480981a64caffe3f95359ea09

PR #10410, opened 2026-06-11
MVC:ui - refactor base_dialog and parseFormNode

AdSchellevis:
**Important notices**

- Model used:
- Extent of AI involvement:

---

**Describe the problem**

Simplify the form templates.

Move defaults to `parseFormNode()` and introduce "sections" with children to avoid some more magic in the volt templates.

Set default title on multi selectpickers as these sometimes seem to miss the "Nothing selected" phrase for some odd reason.

Functionally this should be a backward compatible change.

---

In order to improve the modal forms at some point, we need to get rid of some complexity to improve overview of places to hook new functionality in.

---

**Related issue**

https://github.com/opnsense/core/issues/9955

### DIFF
--- a/src/opnsense/mvc/app/views/layout_partials/base_form.volt
+++ b/src/opnsense/mvc/app/views/layout_partials/base_form.volt
@@ -1,5 +1,5 @@
 {#
- # Copyright (c) 2014-2025 Deciso B.V.
+ # Copyright (c) 2014-2026 Deciso B.V.
  # All rights reserved.
  #
  # Redistribution and use in source and binary forms, with or without modification,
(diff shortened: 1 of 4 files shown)


src/opnsense/mvc/app/views/layout_partials/base_form.volt
```diff
@@ -35,90 +35,79 @@
...
+                <col class="col-md-{{ msgzone_width }}"/>
             {% else %}
-              {{ partial("layout_partials/form_input_tr",field)}}
+                <col style="width: 25%;" />
+                <col style="width: 40%;" />
+                <col style="width: 35%;" />
+            {% endif %}
+            </colgroup>
+            {% if section['type'] %}
+            <thead {% if section['static']|default('false')=='false' %} style="cursor: pointer;"{% endif %}>
```
fichtner, line 55: can we unify the header templating while we're here? noticed yesterday that there are two spots that do the same

src/opnsense/mvc/app/views/layout_partials/base_form.volt
```diff
@@ -35,90 +35,79 @@
...
+            {% endfor %}
+            </tbody>
+        </table>
+        </div>
+    {% endfor %}
 </form>
 
+
+<div class="tab-content content-box" style="padding: 10px;">
+    {% if apply_btn_id|default('') != '' %}
```
fichtner: I'd like to remove this or at least move the users of apply_btn_id to the new "button" form field if you agree, but can do that later

AdSchellevis: @fichtner was thinking the same, but was aiming for a first step that does "the same as it did", we only use this in a couple of spots anyway.

src/opnsense/mvc/app/views/layout_partials/base_form.volt
AdSchellevis: both templates are practically the same, I think we can merge them at some point, which should even ease this further (but requires modifications on the callers)

fichtner: sure, no rush. and, yes, I'd prefer unification of this layer for forms and dialogs too

src/opnsense/mvc/app/views/layout_partials/base_form.volt
AdSchellevis: I think we should merge the current state first, `apply_btn_id` is quite hard to remove at the moment, the direct callers are only a couple (including plugins), but the `base_tabs_content` is used in more, mostly, plugins.

I'll try to refactor the direct callers from core in the master branch in the meantime.

fichtner (approved):
ok let's go :)

AdSchellevis:
for reference https://github.com/opnsense/core/commit/c822d0efa3c8ddc24be955942dc602d82b26de2d

fichtner:
> PHP Warning:  Undefined variable $base_dialog_id in /var/lib/php/cache/_usr_local_opnsense_mvc_app_views_layout_partials_base_form.volt.php on line 47

AdSchellevis:
@fichtner thanks, should be fixed with https://github.com/opnsense/core/commit/25521f0a90c42bfc4d2d8b710d3a9c094a3c8d55

PR #10417, opened 2026-06-14
Firewall: fix 500 (TypeError) on alias getItem with unknown UUID

haxorton:
**Important notices**

- [ ] I opened an issue first for non-trivial changes and linked it below. <- problem is trivial

- Model used: Opus 4.8
- Extent of AI involvement: found the error while testing and helped locate the right class

---

**Describe the problem**

getItemAction() called array_keys($response['alias']['content']); for a UUID that does not resolve getBase() returns [], so the argument is null and on PHP 8 array_keys(null) raises a TypeError, which is unhandled and surfaces as HTTP 500. Guard the augmentation so an unknown UUID returns the empty base response (HTTP 200), consistent with the generic getItemAction behaviour.

---

Just an empty guard around the alias handler

---

**Related issue**

No related issue found

### DIFF
--- a/src/opnsense/mvc/app/controllers/OPNsense/Firewall/Api/AliasController.php
+++ b/src/opnsense/mvc/app/controllers/OPNsense/Firewall/Api/AliasController.php
@@ -168,17 +168,15 @@ class AliasController extends ApiMutableModelControllerBase
     public function getItemAction($uuid = null)
     {
         $response = $this->getBase("alias", "aliases.alias", $uuid);
-        if (!empty($response['alias'])) {
-            $selected_aliases = array_keys($response['alias']['content']);
-            foreach ($this->getModel()->aliasIterator() as $alias) {
-                if (!in_array($alias['name'], $selected_aliases)) {
-                    $response['alias']['content'][$alias['name']] = [
-                      "selected" => 0, "value" => $alias['name']
-                    ];
-                }
-                // append descriptions
-                $response['alias']['content'][$alias['name']]['description'] = $alias['description'];
+        $selected_aliases = array_keys($response['alias']['content'] ?? []);
+        foreach ($this->getModel()->aliasIterator() as $alias) {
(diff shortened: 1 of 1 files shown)


src/opnsense/mvc/app/controllers/OPNsense/Firewall/Api/AliasController.php
```diff
@@ -168,15 +168,17 @@ public function addItemAction()
...
     {
         $response = $this->getBase("alias", "aliases.alias", $uuid);
-        $selected_aliases = array_keys($response['alias']['content']);
-        foreach ($this->getModel()->aliasIterator() as $alias) {
-            if (!in_array($alias['name'], $selected_aliases)) {
-                $response['alias']['content'][$alias['name']] = [
-                  "selected" => 0, "value" => $alias['name']
-                ];
+        if (!empty($response['alias'])) {
+            $selected_aliases = array_keys($response['alias']['content']);
```
fichtner:
```suggestion
            $selected_aliases = array_keys($response['alias']['content'] ?? []);
```

doesn't this do the right thing? I doubt we do the right thing by omitting the look through unrelated aliasIterator()

haxorton: Both work. Matter of taste imho. I can live with any solution that is feasible and does not throw a 500 :)

fichtner: care to change your PR to retain authorship for you? otherwise I will commit. just let me know

haxorton: changed the fix to use null coalescing instead. Thanks for your feedback :)

fichtner (approved):

fichtner:
Merged, thanks!

PR #10428, opened 2026-06-16
fix: pattern for carp_status using wildcard

Vezpi:
**Important notices**

---

**Describe the problem**

When using the API endpoint `/api/diagnostics/interface/carp_status/maintenance` using `POST` to set the node into CARP maintenance mode, I get a 403 Forbidden, even if the user has granted access to all pages.

I have the following error in the backend logs:
```
 uri /api/diagnostics/interface/carp_status/maintenance not accessible for user <user> using api key <key>
```` 

---

The problem seems to be related to ACL.
I have added a `/*` in the pattern of the API endpoint in the related ACL file to make it work with the corresponding privilege.


fichtner:
Looks good to me. I added 68d9da1 since that seemed odd but it should work the same way as before. Merged, thanks!

PR #10429, opened 2026-06-17
routing: fix HTTP 500 when deleting a non-existent gateway

haxorton:
**Important notices**

- Model used: Opus 4.8
- Extent of AI involvement: testing, surfacing the issue and writing a nice commit message instead of my gibberish

---

**Describe the problem**

delGatewayAction() dereferenced the result of getNodeByReference() without a null check, so an unknown uuid reached "(string)$gateway->name" on null and raised an error, which the API renders as HTTP 500 ("Unexpected error, check log for details").

This is a super minor issue and only surfaces while using the API extensively.

---

Guard the lookup and return the already-initialised {"result":"failed"} instead, matching the inherited del* verbs and the adjacent toggleGatewayAction(), which already null-check getNodeByReference().

---

**Related issue**

none found

### DIFF
--- a/src/opnsense/mvc/app/controllers/OPNsense/Routing/Api/SettingsController.php
+++ b/src/opnsense/mvc/app/controllers/OPNsense/Routing/Api/SettingsController.php
@@ -200,6 +200,9 @@ class SettingsController extends ApiMutableModelControllerBase
         if ($this->request->isPost()) {
             if ($uuid != null) {
                 $gateway = $this->getModel()->getNodeByReference('gateway_item.' . $uuid);
+                if ($gateway === null) {
+                    return $result; // unknown uuid -> {"result":"failed"}, mirrors toggleGatewayAction()
+                }
                 $cfg = Config::getInstance()->object();
                 foreach ($cfg->interfaces->children() as $tag => $interface) {
                     if ((string)$interface->gateway == (string)$gateway->name) {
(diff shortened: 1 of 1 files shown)


src/opnsense/mvc/app/controllers/OPNsense/Routing/Api/SettingsController.php
```diff
@@ -200,6 +200,9 @@ public function delGatewayAction($uuid)
         if ($this->request->isPost()) {
             if ($uuid != null) {
                 $gateway = $this->getModel()->getNodeByReference('gateway_item.' . $uuid);
+                if ($gateway === null) {
+                    return $result; // unknown uuid -> {"result":"failed"}, mirrors toggleGatewayAction()
```
fichtner:
```suggestion
                    return $result;
```

ok but no need to make a comment here

fichtner:
Merged, thanks!

haxorton:
@fichtner with your comment on my other PR, shouldn't then this also be more like a UserException instead? Is there some guideline how to gracefully handle errors in general?
And sorry for the comment at the end of the line, I forgot to remove that one :(

fichtner:
I think there's no error needed if the UUID to be deleted is already deleted. Only if the cannot be deleted the user needs to know why.

PR #10430, opened 2026-06-17
firewall: return graceful failure instead of 500 on rule not-found

haxorton:
**Important notices**

- Model used: Opus 4.8
- Extent of AI involvement: testing, surfacing the code location and writing a nice commit message

---

**Describe the problem**

moveRuleBefore (FilterController and FilterBaseController::moveRuleBeforeBase) and toggleRuleLogBase threw a UserException when the referenced rule uuid did not exist.

Because api.php renders every thrown exception as HTTP 500, a non-existent uuid produced a 500, whereas the inherited del/set/toggle verbs return {"result":"failed"} and these very methods already return ["status" => "error", ...] for an invalid request method.

---

Return that same ["status" => "error", ...] shape for the not-found case so the outcome is consistent and no longer surfaces as a server error

Again, not something breaking but more a OCD topic to make it more streamlined.

---

**Related issue**

None found

### DIFF
--- a/src/opnsense/mvc/app/controllers/OPNsense/Firewall/Api/FilterBaseController.php
+++ b/src/opnsense/mvc/app/controllers/OPNsense/Firewall/Api/FilterBaseController.php
@@ -391,9 +391,8 @@ abstract class FilterBaseController extends ApiMutableModelControllerBase
         $selected_node = $this->getModel()->getNodeByReference($node_reference . '.' . $selected_uuid);
 
         if ($target_node === null || $selected_node === null) {
-            throw new UserException(
-                gettext("Either source or destination is not a rule managed with this component")
-            );
+            return ["status" => "error",
+                    "message" => gettext("Either source or destination is not a rule managed with this component")];
         }
 
         $step_size = 50;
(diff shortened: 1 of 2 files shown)


fichtner:
I'm not sure this is the way to go here. We were planning to increase the use of UserException() because it works on all GUI pages. Deferring to @AdSchellevis when he is back.

haxorton:
Ah if that is the plan, then I am fine with that too :)

AdSchellevis:
The user exceptions are intentional indeed, but we might need to improve the error model a bit for consistency indeed. Let's keep this open for further discussion.

A lot of these standard errors are silently ignored, which in most cases isn't an issue as these are hardly triggered anyway. We could consider adding some different exceptions and trapping them in the UI like the user-exception at some point for clarity.

haxorton:
Anything I can support with? I really like opnsense and love to have it, so when I can do something please say so.

AdSchellevis:
@haxorton not yet, but thanks for the offer.

AdSchellevis:
added https://github.com/opnsense/core/commit/99e49c574ff4e3c80b7e133c7204fec73d244f18 to offer some typing for exceptions and replaced the ones in the firewall code with warnings. For consistency we might consider replacing some other callers as well.

PR #10434, opened 2026-06-18
mvc: guard BaseField::setNodes() against a list given for a scalar leaf

haxorton:
**Important notices**

- Model used: Opus 4.8
- Extent of AI involvement: testing, finding the right spot and wording the commit message

---

**Describe the problem**

Posting a JSON array for a scalar (or AsList) model field through the API produces an uncontrolled HTTP 500 — {"errorMessage":"Unexpected error, check log for details"} — with a fatal PHP TypeError in /var/log/lighttpd/error.log:

TypeError: strtolower(): Argument #1 ($string) must be of type string, array given in .../Base/FieldTypes/NetworkField.php  (setValue: trim(strtolower($value)))

Root cause is in BaseField::setNodes(), not the field type. The dispatch is asymmetric:

```
if ($node->isContainer()) {
    if (is_array($data[$key])) { $node->setNodes($data[$key]); }
    else { throw new Exception("Invalid input type for {$key} ..."); } // container guarded
} else {
    $node->setValue($data[$key]);   // leaf: NO type guard — an array slips straight through
}
```

A container that receives a non-array is already rejected, but a leaf that receives an array is passed verbatim to setValue(). Any leaf field whose setter assumes a string then crashes. NetworkField is where it surfaces loudly (it does trim(strtolower($value))), reproducible on stock endpoints such as unbound/settings/addAcl, dhcrelay/settings/addDest, bind/acl/addAcl, and caddy/reverse_proxy/addAccessList (all expose an AsList NetworkField), e.g.:

POST /api/unbound/settings/addAcl {"acl":{"enabled":"1","action":"allow","networks":["10.0.0.0/8","192.168.0.0/16"]}}

---

Complete the type-symmetry the method already started: in the leaf branch, throw a UserException when the value is an array, mirroring the existing container-branch guard. This converts the uncontrolled TypeError into a controlled, message-bearing error at the dispatch point, so it covers every leaf field that assumes a string — not just NetworkField — and fits the move toward UserException for user-facing errors.

*Narrower alternative:* if you'd prefer to keep the change inside the field type, the same guard can live in NetworkField::setValue() (reject non-string input there). That fixes the reported crash but leaves other leaf field types that assume a string exposed, which is why I put the guard at the setNodes() dispatch instead. Happy to scope it down to NetworkField if that's the preference.

---

**Related issue**

None

### DIFF
--- a/src/opnsense/mvc/app/models/OPNsense/Base/FieldTypes/BaseField.php
+++ b/src/opnsense/mvc/app/models/OPNsense/Base/FieldTypes/BaseField.php
@@ -794,12 +794,14 @@ abstract class BaseField
         foreach ($this->iterateItems() as $key => $node) {
             if ($data != null && isset($data[$key])) {
                 if ($node->isContainer()) {
-                    if (is_array($data[$key])) {
-                        $node->setNodes($data[$key]);
-                    } else {
-                        throw new Exception("Invalid  input type for {$key} (configuration error?)");
+                    if (!is_array($data[$key])) {
+                        throw new Exception("Invalid input type for {$key} (configuration error?)");
                     }
+                    $node->setNodes($data[$key]);
                 } else {
+                    if (is_array($data[$key])) {
+                        throw new Exception("Invalid input for \"{$key}\": expected a single value, not a list.");
+                    }
                     $node->setValue($data[$key]);
                 }
             }


src/opnsense/mvc/app/models/OPNsense/Base/FieldTypes/BaseField.php
```diff
@@ -800,6 +801,11 @@ public function setNodes($data)
                         throw new Exception("Invalid  input type for {$key} (configuration error?)");
                     }
                 } else {
+                    if (is_array($data[$key])) {
+                        throw new UserException(
```
fichtner: I agree with the idea here, but I doubt the asymmetry between the Exception/UserException. Since these are field types and not necessarily operated by a "user" it's probably better to use Exception here too and translate it to a UserException if that's needed. Though, for a controller setting, the UserException makes more sense. We're in a model here.

src/opnsense/mvc/app/models/OPNsense/Base/FieldTypes/BaseField.php
```diff
@@ -800,6 +801,11 @@ public function setNodes($data)
...
+                            sprintf(gettext('Invalid input for "%s": expected a single value, not a list.'), $key)
```
fichtner: It's also a bit futile to try and translate errors on this level. One of the biggest challenges in tracing messages is a translation for this supplied by a user in a different language

src/opnsense/mvc/app/models/OPNsense/Base/FieldTypes/BaseField.php
```diff
@@ -800,6 +801,11 @@ public function setNodes($data)
                         throw new Exception("Invalid  input type for {$key} (configuration error?)");
                     }
                 } else {
+                    if (is_array($data[$key])) {
```
fichtner, line 802: Probably also best to use the opportunity to do an if (invalid combinations) { exception }; $node->setValue($data[$key]); instead of calling it twice conditionally inside the scope

src/opnsense/mvc/app/models/OPNsense/Base/FieldTypes/BaseField.php
haxorton: Fair enough, I just stuck to the pattern I saw so far. If it is some "GUI" shown error then yes, as translation would be nice, if it is API I am also kind of a fan to just stick to English.

src/opnsense/mvc/app/models/OPNsense/Base/FieldTypes/BaseField.php
haxorton: Exception is fine too as long as it is meaningful :) Want me to change the PR?

src/opnsense/mvc/app/models/OPNsense/Base/FieldTypes/BaseField.php
haxorton: I adapted your other comments but I am not sure what you meant with this one. Do you mean something like a "validate if with throw" and a set afterwards? That kind of doesn't work due to the different functions calls, wouldn't it?
Otherwise could you elaborate a bit more please?

fichtner: Sorry, I missed the fact that one calls setNodes and the other one calls setValue

fichtner (approved):
I'll likely adjust the messages a bit but this looks better indeed

fichtner:
Merged, thanks!

haxorton:
You are very welcome :)

PR #10436, opened 2026-06-21
system: don't lock the Administration page on an in-use web GUI cert that fails the server-purpose check

murosorg:
The web GUI certificate is re-validated on every save of System: Settings: Administration. If `cert_get_purpose()` reports the certificate is not usable as a server certificate, an input error is added, and since one error rejects the whole form, the page can no longer be saved. Older self-signed certificates created before the serverAuth purpose was enforced fail this check permanently, which also blocks saving unrelated SSH/HTTP settings on the same page.

The box is in fact already serving TLS with that very certificate, so failing the save is counter-productive and can lock the administrator out of the page.

Change: only block when the user selects a different certificate that is not server-capable. If the failing certificate is the one already in use, accept the save and force a web GUI reconfigure, which regenerates the self-signed certificate via `webgui_create_selfsigned()` and lets the other settings apply.

How to reproduce:
- have a web GUI self-signed certificate without the serverAuth purpose in use
- open System: Settings: Administration, change any SSH or HTTP option, Save
- before: "Certificate ... is not intended for server use." and nothing saves
- after: settings save, the GUI reconfigures and the certificate is regenerated

### DIFF
--- a/src/www/system_advanced_admin.php
+++ b/src/www/system_advanced_admin.php
@@ -134,15 +134,30 @@ if ($_SERVER['REQUEST_METHOD'] === 'GET') {
         }
     }
 
+    $webgui_cert_stale = false;
     if (!empty($pconfig['ssl-certref'])) {
+        $current_certref = $config['system']['webgui']['ssl-certref'] ?? '';
         foreach (config_read_array('cert', false) as $cert) {
             if ($cert['refid'] == $pconfig['ssl-certref']) {
                 if (cert_get_purpose($cert['crt'])['server'] == 'No') {
-                    $input_errors[] = gettext(
-                        sprintf('Certificate %s is not intended for server use.', $cert['descr'])
-                    );
-                    break;
+                    if ($pconfig['ssl-certref'] != $current_certref) {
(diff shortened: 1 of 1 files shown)


AdSchellevis:
I don’t think this is worth the effort, faulty certificates can easily trip depending services like lighttpd on minor upgrades which makes this a valuable unconditional error anyway.

fichtner:
So what added the cert and how was the webgui set to use it? Basic first questions.

murosorg:
Thanks for the quick look. To be upfront: the box that hit this had a
self-signed certificate without the serverAuth EKU set as the web GUI
certificate. It was not produced by the current webgui_create_selfsigned()
path, which correctly uses the server_cert profile, so a fresh install is not
affected; in our case the certificate predated that and was carried over.

So I agree this is not a defect in certificate generation. The only part that
felt off was that a single failing field (the certificate purpose) rejects the
whole Administration form, including unrelated SSH/HTTP settings, with no way
to recover from that page. If the intended behavior is that an unsuitable web
GUI certificate must be fixed in the certificate manager first.

I apologize for this waste of time and close this PR.

PR #10444, opened 2026-06-24
firewall: skip alias on rules GUI reload

fichtner:
Also align the alias load path in the controller with how !skip_alias serializes the sequence after rules reload inside filter_configure_sync().

**Important notices**

---

**Describe the problem**

Firewall reload is more expensive than necessary.

---

Skip alias on apply on new rules/nat GUI.

---

**Related issue**

N/A


Monviech (approved):
I think from what we have seen when testing this in a large ruleset it should be fine? It cuts reconfigure down by a lot especially when large tables are used.

Issue #10443, opened 2026-06-23
suricata: custom rules


sopex:
**Important notices**

Before you add a new report, we ask you kindly to acknowledge the following:

- [X] I have read the contributing guidelines at https://github.com/opnsense/core/blob/master/CONTRIBUTING.md
- [X] I am convinced that my issue is new after having checked both open and closed issues at https://github.com/opnsense/core/issues?q=is%3Aissue

**Is your feature request related to a problem? Please describe.**

Custom rules cannot be edited by "Policy" since they don't show up on the rulesets.

sopex:
@Monviech Hi, is this support because it should work?

Monviech:
The support label also means "Awaiting triage".

<img width="617" height="196" alt="Image" src="https://github.com/user-attachments/assets/a19e3e82-9023-4ff3-b6a5-9be9fa3a1081" />

PR #10455, opened 2026-06-26
suricata: policy for custom rules

sopex:
**Important notices**

**Describe the problem**

Custom rules cannot be edited by "Policy" since they don't show up on the rulesets. 
While these can be edited offline, I believe it's a useful tool.

**Related issue**

https://github.com/opnsense/core/issues/10443

### DIFF
--- a/src/opnsense/mvc/app/models/OPNsense/IDS/FieldTypes/PolicyRulesetsField.php
+++ b/src/opnsense/mvc/app/models/OPNsense/IDS/FieldTypes/PolicyRulesetsField.php
@@ -1,0 +1,82 @@
+<?php
+
+/*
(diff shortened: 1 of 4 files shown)


AdSchellevis:
@sopex I'm not sure we should do this, if you're manually dumping rules on the machine, why not set the proper policies in that case? It's rather simple to offer an xml to fetch these files and have the metadata available.

sopex:
@AdSchellevis Would you entertain a way to upload rule files through the gui and have the xml file created on the fly based on user choices?

AdSchellevis:
@sopex not planned, there is a small user defined rules tab, but it doesn't allow many options at the moment. But realistically, if one can dump rule files manually, the same would be the case for that xml file.

sopex:
@AdSchellevis Ok, just asking to not waste your time with PRs :)

PR #10460, opened 2026-06-26
dashboard: fix JavaScript bugs in dashboard widget code

Zenutrix:
**Important notices**

---

**Describe the problem**

Multiple JavaScript bugs in the dashboard widget code cause silent failures and crashes:

1. `BaseWidget.js`: `typeof(widget_config[key] === 'array')` passes a boolean into `typeof` instead of the value itself, always returning `'boolean'`. Type checks for empty arrays, objects, and strings never work — widget config options never fall back to their defaults.

2. `SystemInformation.js`, `Interfaces.js`, `opnsense_widget_manager.js`: `!x in obj` is parsed as `(!x) in obj` due to operator precedence — the guards never trigger.

3. `Firewall.js` (line 122): Closing `</div>` written as `<div>`, producing broken DOM structure.

4. `Firewall.js` (lines 187–188): `style` attribute missing closing quote — `data-toggle` and `title` are swallowed into the style value, breaking tooltips.

5. `Wireguard.js`: Error message interpolated directly into HTML instead of using jQuery `.text()`, risking unescaped content.

6. `KeaLeases.js`: `parseInt(config.leasesToShow)` without a fallback returns `NaN` when the config is empty, breaking the API request.

7. `OpenVPNClients.js`: When a server session appears before its client sessions in the API response, the `clients` array is overwritten with `null`, causing a crash on the next `.push()` call.

---

Each fix is minimal and targeted:

- `BaseWidget.js`: Replace broken `typeof()` checks with `Array.isArray()` and proper `typeof` comparisons.
- `SystemInformation.js`, `Interfaces.js`, `opnsense_widget_manager.js`: Add parentheses to correctly express `!(x in obj)`.
- `Firewall.js`: Fix closing tag and style attribute quote.
- `Wireguard.js`: Use `$('<a>').text(message)` instead of template literal interpolation.
- `KeaLeases.js`: Add null-coalescing fallback consistent with `DnsmasqLeases.js`.
- `OpenVPNClients.js`: Preserve the existing `clients` array when overwriting the server entry.

---

**Related issue**

None — self-contained bug fixes.

### DIFF
--- a/src/opnsense/www/js/widgets/KeaLeases.js
+++ b/src/opnsense/www/js/widgets/KeaLeases.js
@@ -60,7 +60,7 @@ export default class KeaLeases extends BaseTableWidget {
         }
 
         const config = await this.getWidgetConfig();
-        const limit = parseInt(config.leasesToShow);
+        const limit = parseInt(config.leasesToShow ?? '2', 10) || 2;
 
         const leasesRequestBody = JSON.stringify({
             rowCount: limit,
--- a/src/opnsense/www/js/opnsense_widget_manager.js
+++ b/src/opnsense/www/js/opnsense_widget_manager.js
@@ -194,7 +194,7 @@ class WidgetManager  {
             this.widgetClasses[id].onVisibilityChanged(!document.hidden);
         });
 
-        if (!id in this.widgetTranslations) {
+        if (!(id in this.widgetTranslations)) {
             console.error('Missing translations for widget', id);
         }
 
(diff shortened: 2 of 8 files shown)


src/opnsense/www/js/widgets/KeaLeases.js
```diff
@@ -60,7 +60,7 @@ export default class KeaLeases extends BaseTableWidget {
         }
 
         const config = await this.getWidgetConfig();
-        const limit = parseInt(config.leasesToShow);
+        const limit = parseInt(config.leasesToShow ?? '2', 10) || 2;
```
swhite2, line 63: The intent of `getWidgetConfig()` is to not have to provide a fallback. If the fallback wasn't provided because of a broken `getWidgetConfig()`, I'd rather we remove the fallback in `DnsmasqLeases.js` as well and instead rely on the fixed version.

Then again, I can't seem to reproduce this with an empty config.

swhite2 (commented):
Just one question, other fixes look good.

AdSchellevis:
@Zenutrix thanks!

PR #10474, opened 2026-06-30
captive portal: adjust accounting interval to Acct-Interim-Interval

### LINKED ISSUE #10453: Captive Portal ignores RADIUS `Acct-Interim-Interval` (attr 85); interim accounting sent every ~5s regardless
**Important notices**

**Describe the bug**

The Captive Portal does not honour the RADIUS `Acct-Interim-Interval` reply attribute (RADIUS attribute 85). When the RADIUS server returns `Acct-Interim-Interval` in the Access-Accept (e.g. 600 seconds), the portal still emits `Acct-Status-Type = Interim-Update` accounting packets on its internal background worker tick — approximately **every 5 seconds per active session** — instead of at the requested interval. Removing the attribute entirely does not change the cadence (interims remain ~5s), so there is currently no way to control interim frequency. The result is excessive, continuous accounting traffic toward the RADIUS server.

No specific last-known-working version — this appears longstanding rather than a regression. Relevant context: the captive portal moved from IPFW to PF in 25.1.4, so the accounting/worker code path changed around then.

**To Reproduce**

1. Configure a Captive Portal zone using a RADIUS auth server with accounting enabled.
2. Have the RADIUS server return `Acct-Interim-Interval = 600` in the Access-Accept (optionally also `Session-Timeout`).
3. Authenticate a client and capture RADIUS accounting traffic to the server.
4. Observe: `Interim-Update` packets are sent approximately every 5 seconds per session; the 600s `Acct-Interim-Interval` is not applied. (Removing the attribute → still ~5s.)

**Expected behavior**

Interim updates are sent at the interval indicated by `Acct-Interim-Interval` (and/or a configurable per-zone interim interval as a fallback when the server supplies none), rather than on the internal worker tick. Start and Stop accounting continue to be sent regardless of interim cadence.

- Setting `Acct-Interim-Interval` to 600s in the RADIUS reply — ignored (this is the core bug).
- Toggling the zone's "always send accounting requests" — no effect on cadence; it only governs whether sessions without a server-imposed limit are tracked at all.
- Removing `Session-Timeout` / `Acct-Interim-Interval` from the reply — with "always send" off this stops accounting **entirely** (no Start, no interim, no Stop), which is unacceptable when a Stop record is required. So accounting is effectively all-or-nothing per zone, with no "Start/Stop + interim at interval N" mode.
- A small local UDP shim to throttle interims per `Acct-Session-Id` — proved too fragile: the portal retransmits any interim it does not receive an Accounting-Response for, so suppressing interims caused retransmits that stalled sessions and produced unreliable accounting records. Abandoned. This operational cost is the reason honouring the attribute natively (or exposing a per-zone interim interval) is needed.

**Screenshots**

N/A.

**Relevant log files**

Packet capture on the portal's loopback at authentication (`tcpdump -i lo0 -vvnn udp port 1812`) shows the Access-Accept **does** carry the attribute, yet interims still arrive every ~5s:

```
Access-Accept (2) ...
  Acct-Interim-Interval Attribute (85), length: 6, Value: 10:00 min      (0x00000258 = 600)
```

Subsequent `Accounting-Request` packets (`Acct-Status-Type = Interim-Update`) then arrive ~5 seconds apart, each with an incrementing `Acct-Session-Time` and a distinct RADIUS id, each acknowledged by the server — i.e. genuine distinct interims, not retransmits.

**Additional context**

- The RADIUS server is reached via radsecproxy (RadSec/TLS), but this is **incidental** — radsecproxy only transports the packets, the `Acct-Interim-Interval` reaches the portal intact (see the capture above), and the cadence is decided entirely by the captive portal. The issue reproduces with a plain (non-RadSec) RADIUS server.
- At scale (thousands of concurrent sessions) ~5s-per-session interim traffic is significant unnecessary load on the NAS, any proxies, and the RADIUS server/database.
- This supersedes a previously-filed, non-template issue which is being closed in favour of this one.

**Environment**

OPNsense 26.1.10 (amd64), Captive Portal (PF backend, post-25.1.4 IPFW→PF migration). RADIUS auth/accounting via radsecproxy 1.11.2 (RadSec/TLS) to an external server — incidental, see Additional context.

swhite2:
This changes the default interval to 600, which is the recommended value according to RFC 2869.

**Important notices**

- Model used:
- Extent of AI involvement:

---

**Describe the problem**

See https://github.com/opnsense/core/issues/10453 for a description of the issue.

---

Change the default accounting interum-update to 600 seconds, but if the RADIUS backend sends back an `Acct-Interim-Interval` in the `Access-Accept`, use that value instead (if valid).

Adds 2 columns to the captive portal database, `accounting_interval` in `session_restrictions` and `last_update_time` in `accounting_state`.

---

**Related issue**

Closes https://github.com/opnsense/core/issues/10453

### DIFF
--- a/src/opnsense/scripts/captiveportal/lib/db.py
+++ b/src/opnsense/scripts/captiveportal/lib/db.py
@@ -101,13 +101,13 @@ class DB(object):
         # migration: add "accounting_interval" column to session_restrictions
         cur.execute("PRAGMA table_info(session_restrictions)")
         if not any([row[1] == 'accounting_interval' for row in cur.fetchall()]):
-            cur.execute("ALTER TABLE session_restrictions ADD COLUMN accounting_interval INT")
+            cur.execute("ALTER TABLE session_restrictions ADD COLUMN accounting_interval INT DEFAULT 600")
             self._connection.commit()
 
         # migration: add "last_update_time" column to accounting_state
         cur.execute("PRAGMA table_info(accounting_state)")
         if not any([row[1] == 'last_update_time' for row in cur.fetchall()]):
-            cur.execute("ALTER TABLE accounting_state ADD COLUMN last_update_time NUMBER")
+            cur.execute("ALTER TABLE accounting_state ADD COLUMN last_update_time NUMBER DEFAULT 0")
             self._connection.commit()
 
         cur.close()
@@ -638,7 +638,7 @@ class DB(object):
 
         self._connection.commit()
 
-    def update_session_restrictions(self, zoneid, sessionid, session_timeout, accounting_interval=None):
+    def update_session_restrictions(self, zoneid, sessionid, session_timeout, accounting_interval):
         """ upsert session restrictions
         :param zoneid: zone id
         :param sessionid: session id
@@ -647,6 +647,7 @@ class DB(object):
         :return: string "add"/"update" to signal the performed action to the client
         """
         cur = self._connection.cursor()
+        accounting_interval = 600 if not accounting_interval else accounting_interval
         qry_params = {'zoneid': zoneid, 'sessionid': sessionid, 'session_timeout': session_timeout, 'accounting_interval': accounting_interval}
         sql_update = """update session_restrictions
                         set session_timeout = :session_timeout,
--- a/src/opnsense/scripts/captiveportal/lib/db.py
+++ b/src/opnsense/scripts/captiveportal/lib/db.py
@@ -647,23 +647,46 @@ class DB(object):
         :return: string "add"/"update" to signal the performed action to the client
         """
         cur = self._connection.cursor()
-        accounting_interval = 600 if not accounting_interval else accounting_interval
-        qry_params = {'zoneid': zoneid, 'sessionid': sessionid, 'session_timeout': session_timeout, 'accounting_interval': accounting_interval}
-        sql_update = """update session_restrictions
-                        set session_timeout = :session_timeout,
-                        accounting_interval = :accounting_interval
-                        where zoneid = :zoneid and sessionid = :sessionid"""
+
+        has_accounting_interval = True if accounting_interval else False
+
+        qry_params = {
+            "zoneid": zoneid,
+            "sessionid": sessionid,
+            "session_timeout": session_timeout,
+            **({"accounting_interval": accounting_interval} if has_accounting_interval else {}),
+        }
+
+        sql_update = f"""
+            UPDATE session_restrictions
+            SET session_timeout = :session_timeout
+                {", accounting_interval = :accounting_interval" if has_accounting_interval else ""}
+            WHERE zoneid = :zoneid AND sessionid = :sessionid
+        """
 
         cur.execute(sql_update, qry_params)
+
         if cur.rowcount == 0:
-            sql_insert = """insert into session_restrictions(zoneid, sessionid, session_timeout, accounting_interval)
-                            values (:zoneid, :sessionid, :session_timeout, :accounting_interval)"""
+            sql_insert = f"""
+                INSERT INTO session_restrictions (
+                    zoneid,
+                    sessionid,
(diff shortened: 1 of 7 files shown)


src/opnsense/scripts/captiveportal/lib/db.py
```diff
@@ -98,6 +98,18 @@ def create(self, force_recreate=False):
...
+        if not any([row[1] == 'accounting_interval' for row in cur.fetchall()]):
+            cur.execute("ALTER TABLE session_restrictions ADD COLUMN accounting_interval INT")
+            self._connection.commit()
+
+        # migration: add "last_update_time" column to accounting_state
+        cur.execute("PRAGMA table_info(accounting_state)")
+        if not any([row[1] == 'last_update_time' for row in cur.fetchall()]):
+            cur.execute("ALTER TABLE accounting_state ADD COLUMN last_update_time NUMBER")
+            self._connection.commit()
+
```
AdSchellevis, line 112: it's a bummer sqlite doesn't support alter `table ... if not exists`

src/opnsense/scripts/captiveportal/lib/db.py
```diff
@@ -98,6 +98,18 @@ def create(self, force_recreate=False):
             """)
             self._connection.commit()
 
+        # migration: add "accounting_interval" column to session_restrictions
+        cur.execute("PRAGMA table_info(session_restrictions)")
+        if not any([row[1] == 'accounting_interval' for row in cur.fetchall()]):
+            cur.execute("ALTER TABLE session_restrictions ADD COLUMN accounting_interval INT")
```
AdSchellevis: I think I would move the default here so you don't have to cope with null later, it also ensures existing records have something configured.

src/opnsense/scripts/captiveportal/lib/db.py
```diff
@@ -98,6 +98,18 @@ def create(self, force_recreate=False):
...
+            self._connection.commit()
+
+        # migration: add "last_update_time" column to accounting_state
+        cur.execute("PRAGMA table_info(accounting_state)")
+        if not any([row[1] == 'last_update_time' for row in cur.fetchall()]):
+            cur.execute("ALTER TABLE accounting_state ADD COLUMN last_update_time NUMBER")
```
AdSchellevis: If I'm not misreading, it's assumed `last_update_time` has a number we can calculate with later, which means this should also have a default.

src/opnsense/scripts/captiveportal/process_accounting_messages.php
```diff
@@ -64,10 +66,12 @@
...
             if ($row['state'] == null) {
                 // new accounting state, send start event (if applicable)
-                $stmt = $db->prepare('insert into accounting_state(zoneid, sessionid, state)
-                                      values (:zoneid, :sessionid, \'RUNNING\')');
+                $stmt = $db->prepare('insert into accounting_state(zoneid, sessionid, state, last_update_time)
+                                      values (:zoneid, :sessionid, \'RUNNING\', :time)');
                 $stmt->bindParam(':zoneid', $row['zoneid']);
                 $stmt->bindParam(':sessionid', $row['sessionid']);
+                $now = time();
+                $stmt->bindParam(':time', $now);
```
AdSchellevis, line 74:
```suggestion
                $stmt->bindParam(':time', time());
```

AdSchellevis (changes requested):
@swhite2 some small remarks, but looks good. thanks!

src/opnsense/scripts/captiveportal/lib/db.py
swhite2: @AdSchellevis updated code should fix this, but it has the drawback of bloating the db code a bit. The obvious advantage is that we don't have to maintain the default in 3 places.

`None` or `''` doesn't translate to `insert the default value` unfortunately..

PR #10487, opened 2026-07-03
Services: Kea DHCP: DHCPv4 - add optional per-subnet interface assignment

WMP:
**Important notices**

- **Model used:** Claude Opus 4.8 (via Claude Code)
- **Extent of AI involvement:** The patch was drafted by the model, closely mirroring
  the existing DHCPv6 `interface` implementation. XML/model lint and PHP were verified,
  and the change was applied and tested by the submitter on a live 26.1.11 install
  (see Testing below).

---

**Describe the problem**

A DHCPv4 subnet in the Kea GUI can only be tied to an interface *implicitly*, by the
interface owning an address inside the subnet's CIDR (Kea's subnet-selection for
directly connected clients). There is no per-subnet `interface` field for DHCPv4 —
it exists only for DHCPv6.

As a consequence it is impossible to serve a subnet that is not addressed on any
interface, e.g. an additional routed block delivered over the same L2. Kea itself
supports a per-subnet `interface` parameter for exactly this case, and with the default
`dhcp-socket-type: raw` it can then answer clients on the selected interface without
owning an address in that subnet.

---

Add an optional per-subnet `interface` field to DHCPv4, mirroring the existing DHCPv6
implementation:

- **model** (`KeaDhcpv4.xml`): new optional `InterfaceField` on `subnet4`. Unlike DHCPv6
  it is *not* required, so leaving it empty preserves the current address-matching
  behaviour and existing configurations are unaffected. Model version bumped to `1.0.6`.
- **config generation** (`KeaDhcpv4.php`): emit `"interface"` into the `subnet4` record
  when set, via `Util::getRealInterface(..., 'inet')` — same pattern as DHCPv6.
- **validation** (`KeaDhcpv4.php`): warn when the chosen interface is not selected in the
  general settings (only checked when the field is set).
- **form** (`dialogSubnet4.xml`): expose it as a dropdown, placed as the first field so it
  shows as the leading column in the subnets grid, matching the DHCPv6 layout. (No
  grouping by interface, since unlike DHCPv6 the field is optional and often empty.)

Guarded by `if (!empty($device))`, so an empty field produces byte-for-byte the same
config as before — purely additive, no migration required.

---

**Testing**

Applied on a live OPNsense **26.1.11** install and exercised end-to-end:

- The subnets grid now shows an **Interface** column as the first column (parity with
  DHCPv6), populated with the interface description or left blank when unset.
- Subnet **with** an interface selected renders the expected key in the generated
  `/usr/local/etc/kea/kea-dhcp4.conf`:
```
root@OPNsense:~ # opnsense-version -v
26.1.11_5
root@OPNsense:~ # opnsense-patch -f -a WMP 859cf8f
Fetched 859cf8f via https://github.com/WMP/core
Hmm...  Looks like a unified diff to me...
The text leading up to this was:
--------------------------
|From 859cf8fbfaaf5a6f0ade27f72d41139c389c3280 Mon Sep 17 00:00:00 2001
|From: Marcin Janowski <marcin.janowski@assecobs.pl>
|Date: Fri, 3 Jul 2026 12:41:48 +0200
|Subject: [PATCH] Services: Kea DHCP: DHCPv4 - add optional per-subnet
| interface assignment
|
|DHCPv4 subnets could only be tied to an interface implicitly, by the
|interface owning an address inside the subnet's CIDR. This made it
|impossible to hand out an additional routed block that is not addressed
|on any interface, even though Kea itself supports a per-subnet
|"interface" parameter for exactly this case (already exposed for
|DHCPv6).
|
|Add an optional "interface" field to subnet4, mirroring the existing
|DHCPv6 implementation:
|- model: new optional InterfaceField (empty keeps the current
|  address-matching behaviour, so existing configs are unaffected)
|- config generation: emit "interface" into the subnet4 record when set
|- validation: warn when the chosen interface is not selected in the
|  general settings
|- form: expose it as a dropdown
|
|Co-Authored-By: Claude Opus 4.8 (1M context) <noreply@anthropic.com>
|---
| .../OPNsense/Kea/forms/dialogSubnet4.xml        |  6 ++++++
| .../mvc/app/models/OPNsense/Kea/KeaDhcpv4.php   | 17 +++++++++++++++++
| .../mvc/app/models/OPNsense/Kea/KeaDhcpv4.xml   |  3 ++-
| 3 files changed, 25 insertions(+), 1 deletion(-)
|
|diff --git a/src/opnsense/mvc/app/controllers/OPNsense/Kea/forms/dialogSubnet4.xml b/src/opnsense/mvc/app/controllers/OPNsense/Kea/forms/dialogSubnet4.xml
|index 6579802bb7e..926442b92cd 100644
|--- a/src/opnsense/mvc/app/controllers/OPNsense/Kea/forms/dialogSubnet4.xml
|+++ b/src/opnsense/mvc/app/controllers/OPNsense/Kea/forms/dialogSubnet4.xml
--------------------------
Patching file opnsense/mvc/app/controllers/OPNsense/Kea/forms/dialogSubnet4.xml using Plan A...
Hunk #1 succeeded at 1.
Hmm...  The next patch looks like a unified diff to me...
The text leading up to this was:
--------------------------
|diff --git a/src/opnsense/mvc/app/models/OPNsense/Kea/KeaDhcpv4.php b/src/opnsense/mvc/app/models/OPNsense/Kea/KeaDhcpv4.php
|index 63edd97f9d3..cd79953dcaf 100644
|--- a/src/opnsense/mvc/app/models/OPNsense/Kea/KeaDhcpv4.php
|+++ b/src/opnsense/mvc/app/models/OPNsense/Kea/KeaDhcpv4.php
--------------------------
Patching file opnsense/mvc/app/models/OPNsense/Kea/KeaDhcpv4.php using Plan A...
Hunk #1 succeeded at 94.
Hunk #2 succeeded at 193.
Hmm...  The next patch looks like a unified diff to me...
The text leading up to this was:
--------------------------
|diff --git a/src/opnsense/mvc/app/models/OPNsense/Kea/KeaDhcpv4.xml b/src/opnsense/mvc/app/models/OPNsense/Kea/KeaDhcpv4.xml
|index b5139814d20..338e5e715bb 100644
|--- a/src/opnsense/mvc/app/models/OPNsense/Kea/KeaDhcpv4.xml
|+++ b/src/opnsense/mvc/app/models/OPNsense/Kea/KeaDhcpv4.xml
--------------------------
Patching file opnsense/mvc/app/models/OPNsense/Kea/KeaDhcpv4.xml using Plan A...
Hunk #1 succeeded at 1.
Hunk #2 succeeded at 85.
done
All patches have been applied successfully.  Have a nice day.
root@OPNsense:~ # cat /usr/local/etc/kea/kea-dhcp4.conf
{
    "Dhcp4": {
        "valid-lifetime": 4000,
        "decline-probation-period": 600,
        "interfaces-config": {
            "interfaces": [
                "vtnet0"
            ],
            "dhcp-socket-type": "raw",
            "service-sockets-max-retries": 5,
            "service-sockets-retry-wait-time": 5000
        },
        "lease-database": {
            "type": "memfile",
            "persist": true
        },
        "control-socket": {
            "socket-type": "unix",
            "socket-name": "\/var\/run\/kea\/kea4-ctrl-socket"
        },
        "loggers": [
            {
                "name": "kea-dhcp4",
                "output_options": [
                    {
                        "output": "syslog"
                    }
                ],
                "severity": "INFO"
            }
        ],
        "subnet4": [
            {
                "id": 1,
                "subnet": "10.10.0.0\/24",
                "next-server": "",
                "match-client-id": true,
                "option-data": [
                    {
                        "name": "domain-name",
                        "data": "internal"
                    }
                ],
                "pools": [
                    {
                        "pool": "10.10.0.10-10.10.0.100"
                    }
                ],
                "reservations": [],
                "user-context": {
                    "uuid": "14023052-13f8-461d-a522-0501d793b528",
                    "description": "test"
                }
            },
            {
                "id": 2,
                "subnet": "10.11.0.0\/24",
                "next-server": "",
                "match-client-id": true,
                "option-data": [
                    {
                        "name": "domain-name",
                        "data": "internal"
                    }
                ],
                "pools": [
                    {
                        "pool": "10.11.0.10-10.11.0.100"
                    }
                ],
                "reservations": [],
                "interface": "vtnet0",
                "user-context": {
                    "uuid": "bb3a0f57-61f3-4689-a80b-523150cf80ce",
                    "description": "test with set interface to WAN"
                }
            }
        ],
        "hooks-libraries": [
            {
                "library": "\/usr\/local\/lib\/kea\/hooks\/libdhcp_lease_cmds.so"
            },
            {
                "library": "\/usr\/local\/lib\/kea\/hooks\/libdhcp_host_cmds.so"
            }
        ]
    }
}root@OPNsense:~ # 
```
Screenshot_20260703_132637
Screenshot_20260703_132440

image

### DIFF
--- a/src/opnsense/mvc/app/controllers/OPNsense/Kea/forms/dialogSubnet4.xml
+++ b/src/opnsense/mvc/app/controllers/OPNsense/Kea/forms/dialogSubnet4.xml
@@ -1,5 +1,11 @@
 <form>
     <field>
+        <id>subnet4.interface</id>
+        <label>Interface</label>
+        <type>dropdown</type>
+        <help>Optionally bind this subnet to a specific interface. This is only needed when the interface has no address within this subnet (e.g. when handing out an additional routed block); when left empty Kea selects the subnet by matching the interface address. The interface must also be selected in the general settings.</help>
+    </field>
+    <field>
         <id>subnet4.subnet</id>
         <label>Subnet</label>
         <type>text</type>
--- a/src/opnsense/mvc/app/models/OPNsense/Kea/KeaDhcpv4.php
+++ b/src/opnsense/mvc/app/models/OPNsense/Kea/KeaDhcpv4.php
@@ -94,6 +94,18 @@ class KeaDhcpv4 extends BaseModel
                 $messages->appendMessage(new Message(gettext("Either a client ID or a MAC address should be specified."), $key . ".hw_address"));
             }
         }
+        // validate changed subnets
+        $this_interfaces = $this->general->interfaces->getValues();
+        foreach ($this->subnets->subnet4->iterateItems() as $subnet) {
+            if (!$validateFullModel && !$subnet->isFieldChanged()) {
+                continue;
+            }
+            if (!$subnet->interface->isEmpty() && !in_array($subnet->interface->getValue(), $this_interfaces)) {
+                $messages->appendMessage(
+                    new Message(gettext('Interface is not selected in the general settings.'), $subnet->__reference . ".interface")
+                );
+            }
+        }
 
         return $messages;
     }
@@ -180,6 +192,11 @@ class KeaDhcpv4 extends BaseModel
             /* add valid-lifetime at this level if given */
             if ($subnet->valid_lifetime->isSet()) {
                 $record['valid-lifetime'] = $subnet->valid_lifetime->asInt();
+            }
+            /* optionally bind this subnet to a specific interface (e.g. an additional routed block) */
+            $device = Util::getRealInterface($subnet->interface->getValue(), 'inet');
+            if (!empty($device)) {
+                $record['interface'] = $device;
             }
             /* add allocator if selected */
             if (!$subnet->allocator->isEmpty()) {
--- a/src/opnsense/mvc/app/models/OPNsense/Kea/KeaDhcpv4.xml
+++ b/src/opnsense/mvc/app/models/OPNsense/Kea/KeaDhcpv4.xml
@@ -1,6 +1,6 @@
 <model>
     <mount>//OPNsense/Kea/dhcp4</mount>
-    <version>1.0.5</version>
+    <version>1.0.6</version>
     <description>Kea DHCPv4 configuration</description>
     <items>
         <general>
@@ -85,6 +85,7 @@
         </ha>
         <subnets>
             <subnet4 type="ArrayField">
+                <interface type="InterfaceField"/>
                 <subnet type="NetworkField">
                     <NetMaskRequired>Y</NetMaskRequired>
                     <AddressFamily>ipv4</AddressFamily>


Monviech:
You could run KEA with a UDP socket, and listen on a loopback interface you created in Devices -> Loopback.

Then with Services -> DHCRelay, listen on the real interfaces where your clients are, and relay messages to KEA on the loopback interface.

That way it can use the relay information to decide the subnet, which means we don't have to add anything new to KEAs DHCPv4 model right now for this very specific usecase.

I know this should (potentially) work, I recently configured and tested (something similiar, but a bit different) for a customer.

EDIT: Though now that I think about it, the giaddr in the relayed message would contain the original interface IP address, which would also not contain the off link IP address. 

Potentially the messages should be relayed by the router who has this routed IP range on-link.

Or putting a VIP on the interface of the off-link router could also solve it.

An off link DHCP range sounds a bit weird to be honest.

WMP:
Thanks @Monviech - that's fair, let me give the concrete use case so it's clear what I'm after.

I run OPNsense on OVH. There you can order several additional public IP blocks that get routed over the vRack and land on the **same native-VLAN L2 segment** where the firewall and Kea sit. The firewall's own address on that segment is private (e.g. `192.168.1.0/24`); the public blocks are **not addressed on the interface**.

What I want is a static `hw-address → IP` map - no dynamic pools at all. Each known client should get a fixed public address, drawn from **several separate off-link blocks, all served on that one interface**.

I understand the VIP approach for a single block: putting a VIP from the block on the interface makes CIDR-based subnet selection work, and it also gives a reachable `server-identifier` for unicast renewals. But with several blocks that becomes one VIP per block purely to satisfy subnet selection, on ranges I'd otherwise not need to address on the box at all.

One more practical angle specific to OVH: these blocks are small and paid per block (a /29 is 6 usable, and OVH already consumes the gateway inside the block, leaving ~5). Because the gateway lives in the block on the OVH side, my firewall needs no address from the block to route - clients use the OVH gateway directly. So the only reason to add a VIP would be to satisfy Kea's subnet selection, which burns one of those scarce usable addresses per block. With a per-subnet interface binding the box holds no address in the block at all, and every usable address stays available for clients. Across several blocks that's a concrete, recurring saving that the VIP workaround can't match.

Since the reserved addresses live in different subnets, I realize those subnets have to be attached to the receiving interface somehow, and that multiple subnets on one link is normally what shared-networks are for. So my question is really about the recommended shape:

- reservation-only subnets (empty pools, just `hw-address`/`ip-address`) grouped in a **shared-network**, attached to the interface, or
- individual subnets each pinned via the per-subnet `interface` parameter?

Either way the missing piece for me is being able to attach a subnet to an interface that holds no address from that subnet  - which DHCPv6 already exposes and DHCPv4 does not. That parity gap is what this PR tries to close; I'm happy to reshape it (e.g. wire it through shared-networks) toward whatever fits the Kea model best.

Monviech:
I do not understand the network layout. I would expect this:

OVH 10.10.10.1 ----> 10.10.10.2 WAN OPNSENSE LAN 201.0.0.1/29 ---->  201.0.0.2/29 Server1

Route 201.0.0.0/29 via 10.10.10.2/32

I don't understand why the servers would be on the same network that is used as transition (transfer) network (10.10.10.0/30). It doesn't make much sense to me, it has a certain smell of asymmetric routing.

If the OPNsense is not the default gateway for 201.0.0.2/29 Server1, what does it protect?

WMP:
Let me describe the OVH setup properly, because it doesn't map to a classic transit-router topology — there is no transfer network, and OPNsense is **not** the L3 gateway for the public block.

In OVH, every server (bare metal, a VM on bare metal, or a VPS) is connected to a virtual switch called the **vRack**. The vRack can carry VLANs. You can buy a block of public IPs and attach it to the vRack — but such a public block can only live on the **native (untagged) VLAN**. Each node then addresses one IP from the block directly on its interface, per OVH's guide: [Configuring an Additional IP block in a vRack](https://docs.ovhcloud.com/en/guides/bare-metal-cloud/dedicated-servers/configuring-an-ip-block-in-a-vrack#gnulinux-configurations). Importantly, **the gateway for that block is provided by OVH, inside the block itself** (in a /29 it's the second-to-last address) — not by any of our machines.

So the native vRack is a flat L2 where every node — our OPNsense firewalls and the customer VMs — is a **peer** holding a public IP, all using the **OVH gateway**. My OPNsense already gets its public IP exactly this way.

```
 Public block 203.0.113.0/29 (OVH layout):
   .0 network · .1–.5 usable · .6 OVH gateway · .7 broadcast   → 5 usable
   
                 OVH gateway 203.0.113.6   (provided by OVH, inside the block)
                          │
  ════════════════════════╪═══════════════════════════════════  vRack · NATIVE (untagged) — public blocks
        │           │            │            │
 ┌────────────┐ ┌────────────┐ ┌───────────┐ ┌───────────┐
 │ OPNsense A │ │ OPNsense B │ │   VM 1    │ │   VM 2    │
 │203.0.113.1 │ │203.0.113.2 │ │203.0.113.3│ │203.0.113.4│
 └─────┬──────┘ └─────┬──────┘ └───────────┘ └───────────┘
       │              │      FW pair — also the Kea DHCP server for the public blocks
       └──────┬───────┘
              │   OPNsense = default gateway + firewall for the VLAN below
  ════════════╪═══════════════════════════════════════════════  vRack · VLAN 30 (tagged) — private, no public IP
        │             │
   ┌──────────┐  ┌──────────┐
   │  host A  │  │  host B  │   default gw = OPNsense   (192.168.30.0/24)
   └──────────┘  └──────────┘
```

OPNsense's actual firewalling job is the **private tagged VLANs behind it** (e.g. VLAN 30, `192.168.30.0/24`, no public addressing) — there it *is* the default gateway and firewall for the hosts. On the native vRack it plays a second, separate role: the **DHCP server** for the public blocks. So it does protect real hosts; it just isn't the gateway for the public L2 (OVH is).

There is **no OVH-provided DHCP** on the vRack — the guide above configures the block statically, per-OS. But these are ordinary addresses on the native L2 (with the OVH gateway), so nothing stops me from handing them out with my **own** DHCP server instead of static config on every machine. That's what I want Kea for. OPNsense A and B each already hold one address from this block (their own public IPs), so for this block Kea can select it by CIDR-matching that address.

The real motivation is **additional blocks**. If I buy another pool, e.g.:

```
 Second block 198.51.100.0/29 (same OVH layout):
   .0 network · .1–.5 usable · .6 OVH gateway · .7 broadcast   → 5 usable
```

and I want to assign those to specific VMs as a controlled `hw-address → IP` map — instead of static config on each VM — I'd run my own Kea on the native vRack. Here's the crux: that Kea box (my OPNsense) holds **no address in the second block**, and shouldn't need one — it isn't the gateway (OVH is), it's purely the DHCP server. So Kea has to serve that subnet on the native-vRack interface even though the interface carries no address from it — which is exactly the per-subnet `interface` binding that DHCPv6 already exposes and DHCPv4 does not.

To your question — "if OPNsense isn't the default gateway, what does it protect?": it protects the private VLANs behind it (VLAN 30 above), where it *is* the gateway. Its role on the native vRack is separate — a DHCP server handing out public blocks — and that's exactly the case where the DHCP server legitimately has no address in the subnet it serves.

Monviech:
I understand the concept now but I think a static IP configuration is the best fit in the network with the shared public space.

I don't think your PR fixes a generic problem that most users have, it mostly adjusts the KEA model OPNsense provides to solve your problem in this specific environment, which in turn makes it more complex for the majority of other users. (More complex meaning they have more decisions to make, and we have to maintain this for years)

So I'm not really sure this will make it in, the usecase is very specific and constructed in my opinion.

WMP:
Fair enough — I agree. For this shared public L2 a static configuration is the pragmatic fit, and this is really a DHCP-server-on-a-flat-L2 role that has little to do with OPNsense's core firewalling. The use case is specific; I won't argue that.

Your main concern — added complexity for the majority — is the right thing to weigh, so let me offer a minimal-footprint version:

- mark the subnet `interface` field as **advanced**, so it doesn't show up in the normal add/edit flow, and
- keep it **hidden from the subnets grid by default**, so users who never touch it don't see it at all.

That way the default experience for everyone who doesn't need it stays exactly as it is today — the field only surfaces if someone deliberately goes looking for it under advanced.

On the maintenance side: it's a near-verbatim mirror of the DHCPv6 subnet `interface` handling you already maintain (same `InterfaceField`, same `getRealInterface()` emit, plus one optional validation), so it closes a DHCPv4/DHCPv6 parity gap rather than introducing a new concept — arguably less divergence to reason about, not more.

Totally your call and your long-term burden, though — if you'd still rather not carry it, I'm completely fine closing this. Happy to push the advanced/hidden-by-default version first if you'd like to see how small it looks that way.

Monviech:
I don't think its a good idea to add it but it's also not entirely my decision so I tag @AdSchellevis for a second opinion.

AdSchellevis:
At the moment I also don't think we should do this, OPNsense has no active role in the 203.0.113.X network and adding the interface option, which is generally not required, will likely only lead to confusion for others.

Issue #10489, opened 2026-07-03
Reporting: Health - filter selection resets on every page reload


issmirnov:
**Important notices**

Before you add a new report, we ask you kindly to acknowledge the following:

- [x] I have read the contributing guidelines at https://github.com/opnsense/core/blob/master/CONTRIBUTING.md
- [x] I am convinced that my issue is new after having checked both open and closed issues at https://github.com/opnsense/core/issues?q=is%3Aissue

**Is your feature request related to a problem? Please describe.**

The Reporting > Health page's Category, Subject, and Granularity selectors always reset to their defaults on every page load. For the purpose of bookmarking or sharing a specific graph view (e.g. linking a colleague directly to a particular interface's quality graph over the last 24 hours), or just not having to reselect the same filters after every reload, I am missing a way for the current selection to survive a refresh.

**Describe the solution you'd like**

Encode the current Category/Subject/Granularity selection into the page's URL query string on change, and restore it on load, falling back to today's defaults when a value is absent or no longer valid (e.g. a removed interface). I have a small patch for this and will open a pull request referencing this issue.

**Describe alternatives you considered**

None - this is a small, self-contained addition with no real alternative approach worth calling out.

**Additional context**

N/A

issmirnov:
PR is up at #10490

PR #10490, opened 2026-07-03
reporting: health: persist category/subject/granularity selection in the URL

issmirnov:
**Important notices**

- Model used: Claude Sonnet 5, via Claude Code
- Extent of AI involvement: I described the bug and asked Claude Code to investigate, implement, and verify a fix. It located the relevant controller/view/JS, implemented the change, and verified it by provisioning a throwaway OPNsense 26.1.6 VM and driving the actual UI in a browser rather than relying on code review alone. That live test caught two real bugs on the first attempt (a granularity restore that silently used the wrong detail level, and a bootstrap-select button that didn't visually refresh after a restore), which were then fixed and re-verified the same way. I reviewed the diff and the test evidence before submitting this PR.

---

**Describe the problem**

Reloading the Reporting > Health page always resets the Category, Subject, and Granularity selectors back to their defaults, discarding whatever the user was looking at and making it impossible to bookmark or share a specific view.

---

Read the three selections from the URL query string on load, falling back to today's defaults when a value is absent or no longer valid (e.g. a removed interface), and keep the query string in sync via `history.replaceState()` as the user changes any of them.

Two implementation details worth flagging for review:
- Restoring a saved detail level needs to seed `HealthGraph.currentDetailLevel` directly, since setting the select's value alone doesn't fire its change handler.
- Restoring a saved category/subject needs an explicit `selectpicker('refresh')` call, since bootstrap-select only redraws its button label on that, not on a re-triggered `changed.bs.select` event. This was invisible before because the only value ever selected here programmatically was whatever the widget already displayed by default.

Verified on a throwaway OPNsense 26.1.6 VM: set a non-default selection, confirmed the URL updates, then hard-refreshed and confirmed both the visual dropdown state and the actual data fetch (network request) use the restored values, including a graceful fallback when a saved subject is no longer valid.

Screenshots/recording to follow in a comment.

---

**Related issue**

Ref: #10489

### DIFF
--- a/src/opnsense/mvc/app/views/OPNsense/Diagnostics/health.volt
+++ b/src/opnsense/mvc/app/views/OPNsense/Diagnostics/health.volt
@@ -121,6 +121,14 @@
 
         const healthGraph = new HealthGraph('health-chart');
 
+        const syncUrlState = () => {
+            const params = new URLSearchParams(location.search);
+            params.set('category', $('#health-category-select').val());
+            params.set('subject', $('#health-subcategory-select').val());
+            params.set('granularity', $('#detail-select').val());
+            history.replaceState(null, '', `${location.pathname}?${params.toString()}`);
+        };
+
         const reset = () => {
             $('#info-disabled').hide();
             $('#main').show();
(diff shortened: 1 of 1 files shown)


issmirnov:
Demo:
opnsense-health-url-persist-demo

swhite2:
The more common pattern for persistence in core is to use localStorage. This doesn't cover the link-sharing use case, but then again that particular use case has rarely/never popped up over the years.

I'm not sure we should go for URL params, but if we do, better to sync it with localStorage persistence

src/opnsense/mvc/app/views/OPNsense/Diagnostics/health.volt
```diff
@@ -121,13 +121,34 @@
...
+                const savedCategory = urlParams.get('category');
+                let savedSubject = urlParams.get('subject');
+                const savedGranularity = urlParams.get('granularity');
+
+                if (savedGranularity !== null && ['0', '1', '2', '3'].includes(savedGranularity)) {
+                    $('#detail-select').val(savedGranularity);
+                    // .val() alone doesn't fire the change handler, so seed the graph's
+                    // state directly or the first fetch below would use the old default
+                    healthGraph.currentDetailLevel = savedGranularity;
+                }
```
swhite2, line 150: This block of logic is the only thing that executes on page load in tandem with the code below:

```
                // trigger event for first category, restoring a saved selection when still valid
                const categoryToSelect = (savedCategory && rrdOptions.data.hasOwnProperty(savedCategory))
                    ? savedCategory
                    : Object.keys(rrdOptions.data)[0];
                $('#health-category-select').val(categoryToSelect);
                $('#health-category-select').selectpicker('refresh');
                $('#health-category-select').trigger('changed.bs.select');
```

Therefore, and since touching the healthGraph internals bypassing the API is a big no-no, why not combine the two blocks and call 

```
$('#detail-select').val(savedGranularity).trigger('change');
```

so the proper `update()` is used.

Issue #10541, opened 2026-07-16
Netflow interface names missing due to index mismatch


Top-Cat:
**Important notices**

Before you add a new report, we ask you kindly to acknowledge the following:

- [x] I have read the contributing guidelines at https://github.com/opnsense/core/blob/master/CONTRIBUTING.md
- [x] I am convinced that my issue is new after having checked both open and closed issues at https://github.com/opnsense/core/issues?q=is%3Aissue

**Describe the bug**

When interfaces have been removed the insight dashboard can show graphs but all the interface names are blank. Also when filtering on the details page no flows will be shown.

**To Reproduce**

Add some random interfaces to bulk up the interface indexes. In my case these were vlan interfaces.
Add interfaces you want to monitor with netflow. I made new vlan interfaces on a different parent device.
Remove the original interfaces. (ie. delete the vlan devices)
Try to use netflow.

**Expected behavior**

The netflow graphs show flows, with the names of the interfaces correctly. The details tab allows you to filter and view flows.

AdSchellevis:
closed https://github.com/opnsense/core/pull/10542

PR #10542, opened 2026-07-16
Correctly map netflow indexes when there are gaps

Top-Cat:
**Important notices**

---

**Describe the problem**

When interfaces have been removed from the system there can be gaps in the indexes which breaks the mapping in the netflow aggregator.

---

Instead of assuming interfaces are returned in index order actually parse the output to gather the index values.

---

**Related issue**

#10541

### DIFF
--- a/src/opnsense/scripts/netflow/lib/parse.py
+++ b/src/opnsense/scripts/netflow/lib/parse.py
@@ -40,10 +40,27 @@ class Interfaces(object):
         """
         self._if_index = dict()
         sp = subprocess.run(['/usr/local/sbin/ifinfo'], capture_output=True, text=True)
-        interfaces = re.findall(r"Interface ([^ ]+)", sp.stdout)
+        interfaces = re.split(r'(?=^Interface )', sp.stdout, flags=re.MULTILINE)
 
-        for index, interface in enumerate(interfaces, start=1):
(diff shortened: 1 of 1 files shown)


AdSchellevis:
both sources use the same index number as far as I know, which means a mismatch would still occur when locating the name in one of the sources.

Top-Cat:
They use the same index numbers but this code used to just assume the first result was 1, second was 2. Except my interfaces do this:

```
# /usr/local/sbin/ifinfo | grep -E "Interface|index"
Interface ixl0 (ixl0):
	index: 1
Interface ixl1 (ixl1):
	index: 2
Interface igc0 (igc0):
	index: 3
Interface igc1 (igc1):
	index: 4
Interface ix0 (ix0):
	index: 5
Interface ix1 (ix1):
	index: 6
Interface lo0 (lo0):
	index: 7
Interface enc0 (enc0):
	index: 8
Interface pfsync0 (pfsync0):
	index: 9
Interface pflog0 (pflog0):
	index: 10
Interface pppoe0 (ng0):
	index: 11
Interface wg0 (wg0):
	index: 17
```

So the old code here maps `12 = wg0` and `17 = 17` which is wrong

AdSchellevis:
@Top-Cat ok, that makes sense, thanks, I'll have a look

PR #10563, opened 2026-07-17
Add options to dhcrelay for circuit_id and remote_id

Top-Cat:
**Important notices**

---

**Describe the problem**

https://forum.opnsense.org/index.php?topic=40110.0

Post the change to openbsd's dhcrelay the circuit/remote id values changed from interface names to interface indexes which do not have obvious mappings and could theoretically change with configuration changes and reboots.

---

This adds two new fields to each relay entry allowing the user to pick what data should be presented in these fields.
I've included the most common values that vendors populate these fields with but I'm open to others.

Screenshot 2026-07-17 at 20 11 39

### DIFF
--- a/src/etc/inc/plugins.inc.d/dhcrelay.inc
+++ b/src/etc/inc/plugins.inc.d/dhcrelay.inc
@@ -90,6 +90,8 @@ function dhcrelay_xmlrpc_sync()
 
 function dhcrelay_configure_id($verbose = false, $id_map = null)
 {
+    global $config;
+
     if (!plugins_argument_map($id_map)) {
         return;
     }
@@ -186,6 +188,30 @@ function dhcrelay_configure_id($verbose = false, $id_map = null)
 
         $cmd_frmt[] = '-i %s';
         $cmd_args[] = $device;
+
+        $circuit_id_values = [
+            'index' => '', // This is the default behaviour
+            'name' => $device,
+            'vlan' => !empty($ifconfig_details[$device]['vlan']['tag']) ? $ifconfig_details[$device]['vlan']['tag'] : '',
+        ];
+
+        $circuit_id = $circuit_id_values[$relay->circuit_id->getValue()];
+        if (!empty($circuit_id)) {
+            $cmd_frmt[] = '-C %s';
+            $cmd_args[] = $circuit_id;
+        }
+
+        $remote_id_values = [
+            'addr' => '', // This is the default behaviour
+            'mac' => $ifconfig_details[$device]['macaddr'],
+            'host' => $config['system']['hostname'],
+        ];
+
+        $remote_id = $remote_id_values[$relay->remote_id->getValue()];
+        if (!empty($remote_id)) {
+            $cmd_frmt[] = '-R %s';
+            $cmd_args[] = $remote_id;
+        }
 
         $has_servers = false;
 
--- a/src/etc/inc/plugins.inc.d/dhcrelay.inc
+++ b/src/etc/inc/plugins.inc.d/dhcrelay.inc
@@ -189,25 +189,13 @@ function dhcrelay_configure_id($verbose = false, $id_map = null)
         $cmd_frmt[] = '-i %s';
         $cmd_args[] = $device;
 
-        $circuit_id_values = [
-            'index' => '', // This is the default behaviour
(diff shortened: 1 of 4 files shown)


AdSchellevis:
This will break on ipv6 as `dhcrelay6` doesn't know `-C` and I prefer to add the options as regular value input as the used may decide for its own what to use for circuit and remote, the relay simply passes these values to the other end anyway.

Top-Cat:
I've added a check to make sure `-C` is only applied for dhcrelay and made both fields free text.
When empty they will default back to interface number and destination address as they do now.

src/opnsense/mvc/app/models/OPNsense/DHCRelay/DHCRelay.xml
```diff
@@ -23,6 +23,8 @@
             <agent_info type="BooleanField">
                 <Required>Y</Required>
             </agent_info>
+            <circuit_id type="TextField" />
```
AdSchellevis:
```suggestion
            <circuit_id type="StrictTextField" />
```

src/opnsense/mvc/app/models/OPNsense/DHCRelay/DHCRelay.xml
```diff
@@ -23,6 +23,8 @@
...
+            <remote_id type="TextField" />
```
AdSchellevis:
```suggestion
            <remote_id type="StrictTextField" />
```

src/etc/inc/plugins.inc.d/dhcrelay.inc
```diff
@@ -187,6 +189,18 @@ function dhcrelay_configure_id($verbose = false, $id_map = null)
         $cmd_frmt[] = '-i %s';
         $cmd_args[] = $device;
 
+        $circuit_id = $relay->circuit_id->getValue();
+        if ($family == 'inet' && !empty($circuit_id)) {
```
AdSchellevis:
```suggestion
        if ($family == 'inet' && !$relay->circuit_id->isEmpty()) {
```

src/etc/inc/plugins.inc.d/dhcrelay.inc
```diff
@@ -187,6 +189,18 @@ function dhcrelay_configure_id($verbose = false, $id_map = null)
...
+            $cmd_frmt[] = '-C %s';
+            $cmd_args[] = $circuit_id;
```
AdSchellevis:
```suggestion
            $cmd_args[] = $relay->circuit_id->getValue();
```

src/etc/inc/plugins.inc.d/dhcrelay.inc
```diff
@@ -90,6 +90,8 @@ function dhcrelay_xmlrpc_sync()
 
 function dhcrelay_configure_id($verbose = false, $id_map = null)
 {
+    global $config;
```
AdSchellevis: unused?

src/etc/inc/plugins.inc.d/dhcrelay.inc
```diff
@@ -187,6 +189,18 @@ function dhcrelay_configure_id($verbose = false, $id_map = null)
         $cmd_frmt[] = '-i %s';
         $cmd_args[] = $device;
 
+        $circuit_id = $relay->circuit_id->getValue();
```
AdSchellevis: it's easier to just use the objects handles, see below.

src/etc/inc/plugins.inc.d/dhcrelay.inc
```diff
@@ -187,6 +189,18 @@ function dhcrelay_configure_id($verbose = false, $id_map = null)
...
+        if ($family == 'inet' && !empty($circuit_id)) {
+            $cmd_frmt[] = '-C %s';
+            $cmd_args[] = $circuit_id;
+        }
+
+        $remote_id = $relay->remote_id->getValue();
```
AdSchellevis: same as above

AdSchellevis (changes requested):
@Top-Cat add some small review comments, nothing very difficult, but saves some code.

AdSchellevis:
@Top-Cat you're quick :) thanks! merged

Issue #4611, opened 2021-01-17
20.7.7 bug No check for $data throws error on reporting -> Traffic


Z0t:
Since 20.7.7 I get the following error randomly when monitoring traffic flow.  This resets the page and can be impactful when tracking live traffic issues.

The error reported:

`/usr/local/opnsense/mvc/app/controllers/OPNsense/Diagnostics/Api/TrafficController.php:74 - Invalid argument supplied for foreach() (errno=2)`

This seem to happen several times an hour which puts a pop up for each time the error happens.

https://github.com/opnsense/core/blob/de291e62660996b04d589381cad3396db5e9eb80/src/opnsense/mvc/app/controllers/OPNsense/Diagnostics/Api/TrafficController.php#L74

        if (count($iflist) > 0) {
            $data = (new Backend())->configdpRun('interface show top', [implode(",", $iflist)]);
            $data = json_decode($data, true);
            foreach ($data as $if => $content) {
                if (isset($ifmap[$if])) {
                    $response[$ifmap[$if]] = $content;
                }
            }
        }

It would appear that $data is not returning valid structures and there is not check for valid structure before the foreach so it fails.

This has been reported in the forums at least twice, and I've submitted "Crash" reports but the UI says it is not a crash (after I submit of course).

https://forum.opnsense.org/index.php?topic=20802.msg96877#msg96877

https://forum.opnsense.org/index.php?topic=20521.msg95452#msg95452

@AdSchellevis  this appears to be from your [commit ](https://github.com/opnsense/core/commit/4186f2bbe5dfd28dc7bf9f47585121e3273297f1)23 days ago

AdSchellevis:
My commit is exactly the fix for this issue and is scheduled for 20.7.8 if I'm not mistaken (note the line 74 is now an is_array()).
```
Error at /usr/local/opnsense/mvc/app/controllers/OPNsense/Diagnostics/Api/TrafficController.php:74 - Invalid argument supplied for foreach() (errno=2)
```

Z0t:
Ahh sorry I misread git blame.  I am not on the forums; it'd be good if someone could reply to those messages. Also if a maintainer could tag this with the release ID (rather than assume it is 20.7.8).

Thanks!

fichtner:
There is nothing here to do but wait for 20.7.8 for 2-3 more days now.


Cheers,
Franco

AdSchellevis:
sounds like a solid plan, lets close the issue

PR #10575, opened 2026-07-18
Reporting: Traffic - guard traffic_top.py against malformed/truncated iftop lines

### LINKED ISSUE #10574: traffic_top.py crashes with IndexError on short iftop lines (configd "interface show top" exit 1)
**Important notices**

- Model used: Anthropic Claude (Opus 4.8) via Claude Code
- Extent of AI involvement: The AI performed the diagnosis on the live firewall (reproducing the crash, isolating the failing line, checking the source in `master`, and searching for duplicate issues) and drafted this entire report. A human reviewed it before submission.

**Describe the bug**

`traffic_top.py` — which backs **Reporting → Traffic → Top talkers** and the **Diagnostics: Traffic** page — can crash with an unhandled `IndexError` while parsing `iftop` output. This makes the configd `interface show top` action exit non-zero, which surfaces in the log as:

```
Script action failed with Command '/usr/local/opnsense/scripts/interfaces/traffic_top.py
--interfaces '...'' returned non-zero exit status 1.
```

The parser treats every `iftop` line containing `=>` or `<=` as a full host row and indexes `parts[2]` / `parts[5]` without checking how many fields the line actually split into. When `iftop` emits a short/truncated line that still contains an arrow token, `line.split()` yields fewer than 3 elements and the script aborts. The script's own traceback (normally swallowed by configd) is:

```
  File "/usr/local/opnsense/scripts/interfaces/traffic_top.py", line 142, in <module>
    'rate': parts[2],
            ~~~~~^^^
IndexError: list index out of range
```

This is **not a regression** — the parsing code has been present since the "Top talkers" feature was introduced (2020) and reworked for aggregation (2021), so there is no previous working version to revert to. It reproduces on current 26.1.11 and is present unchanged in `master`.

**To Reproduce**

1. Have several interfaces carrying live traffic.
2. Go to **Diagnostics: Traffic** (`/ui/diagnostics/traffic`) and open it in **3 browser windows at the same time**, so the `diagnostics/traffic/top` endpoint is polled by overlapping/concurrent requests.
3. Leave them polling for a few minutes.
4. See intermittent `Script action failed ... traffic_top.py ... returned non-zero exit status 1` entries in **System → Log Files** / `/var/log/configd/latest.log`. (The graph in the browser also drops out on each failed poll.)

Deterministic CLI reproduction (fires within a minute or two on a busy box — the concurrency is what matters, since each invocation already launches one `iftop` per interface in parallel):

```sh
IF='<your interface list>'
for r in $(seq 1 60); do
  for c in 1 2 3 4; do
    ( /usr/local/opnsense/scripts/interfaces/traffic_top.py --interfaces "$IF" >/dev/null \
        || echo "FAIL round=$r c=$c" ) &
  done
  wait
done
```

**Expected behavior**

`traffic_top.py` should tolerate malformed/partial `iftop` lines — skip any line that doesn't have the expected number of columns — and never exit non-zero because of them. A single bad line from one interface should not fail the whole request.

- Workaround: only keep the Traffic/Diagnostics graph open in a single tab/window, which greatly reduces the concurrent polling that provokes the truncated `iftop` output. This only lowers the probability; it does not fix the underlying missing length check.

**Screenshots**

N/A (backend script crash; reproduced from the CLI).

**Relevant log files**

The crash is in the arrow-line parsing loop of `src/opnsense/scripts/interfaces/traffic_top.py`:

```python
135:        for line in iftop_data[intf].split('\n'):
136:            if line.find('=>') > -1 or line.find('<=') > -1:
137:                parts = line.split()
138:                if parts[0].find('.') == -1 and parts[0].find(':') == -1:
139:                    parts.pop(0)
140:                item = {
141:                    'address': parts[0],
142:                    'rate': parts[2],          # <-- IndexError here
143:                    'rate_bits': int(from_bformat(parts[2])),
144:                    'cumulative': parts[5],
145:                    'cumulative_bytes': int(from_bformat(parts[5])),
```

There is no `len(parts)` guard before accessing indices `0`, `2`, and `5`.

**Additional context**

Related, previously-closed issue: #4611 ("No check for `$data` throws error on reporting → Traffic", closed 2021) was the same event seen from the PHP side — `TrafficController.php` ran `foreach()` over the empty/invalid output of the failed `interface show top` action. That was fixed by adding an `if (is_array($data))` guard (still present today), which suppresses the GUI popup but only masks the symptom: `traffic_top.py` itself still crashes and the configd action still exits non-zero (the `Script action failed ... returned non-zero exit status 1` log entry, and a dropped graph poll). This report is about that unaddressed root cause.

Root cause: the **Diagnostics: Traffic** page polls `diagnostics/traffic/top`, which runs `traffic_top.py`. That script starts one `iftop -nNb -i <if> -s 2 -t` per interface concurrently via `ThreadPoolExecutor`. With several browser tabs polling at once, dozens of `iftop` processes run simultaneously; under that contention `iftop` occasionally writes a truncated line, and the unguarded parser throws. This is why it is intermittent and load-dependent, and why the number of monitored interfaces raises the odds (more concurrent `iftop` per invocation).

**Update:** while verifying a fix, a **second, distinct crash** in the same loop surfaced under the same concurrent load. `iftop` can also emit a non-host line that still contains `=>`/`<=` where `parts[0]` is not an address (e.g. a mangled footer line). The code then calls `ipaddress.ip_address(parts[0])` inside a `try/except subprocess.TimeoutExpired`, which does **not** catch the resulting `ValueError`, so the script still aborts:

```
  File "/usr/local/opnsense/scripts/interfaces/traffic_top.py", line 154, in <module>
    ip = ipaddress.ip_address(parts[0])
ValueError: 'receive' does not appear to be an IPv4 or IPv6 address
```

Suggested fix — require the expected column count **and** validate `parts[0]` as an IP before using the row (covering both modes), guard the leading-index `pop()` against an empty split, and drop the incorrect `except subprocess.TimeoutExpired` now that the address is validated up front:

```python
                parts = line.split()
                if parts and parts[0].find('.') == -1 and parts[0].find(':') == -1:
                    parts.pop(0)
                if len(parts) < 6:
                    continue
                try:
                    ip = ipaddress.ip_address(parts[0])
                except ValueError:
                    continue
                item = { ... }
                # attach tags using the validated `ip` directly (no try/except)
                if ip in all_local_addresses:
                    item['tags'].append('local')
                if ip.is_private:
                    item['tags'].append('private')
```

A complete patch addressing **both** failure modes is proposed in PR #10575.

**Environment**

Software version used and hardware type if relevant:

OPNsense 26.1.11_10 (amd64) — also reproduces on `master`
FreeBSD 14.3-RELEASE-p16
Python 3.13.14, iftop 1.0.p4_1
Affected GUI page: `/ui/diagnostics/traffic` (also `Reporting → Traffic → Top talkers`); API endpoint `diagnostics/traffic/top`; configd action `interface show top`

brendanbank:
Closes #10574.

### Problem

`traffic_top.py` (backing **Reporting → Traffic → Top talkers** and the **Diagnostics: Traffic** page) parses every `iftop` line containing `=>`/`<=` as a full host row.

Under load — e.g. the Traffic page polled by several browser tabs at once, each invocation spawning one `iftop` per interface via `ThreadPoolExecutor` — `iftop` can emit malformed output and the parser aborts, failing the `interface show top` configd action with exit status 1. Two distinct crashes were observed on a live system:

1. **Truncated row** (fewer columns) → `IndexError` on `parts[2]`/`parts[5]`:
   ```
   File "/usr/local/opnsense/scripts/interfaces/traffic_top.py", line 142, in <module>
       'rate': parts[2],
   IndexError: list index out of range
   ```
2. **Non-host output** that still contains `=>`/`<=` (e.g. a mangled footer line) → `parts[0]` is not an address, so `ipaddress.ip_address()` raises `ValueError`, which the surrounding `except subprocess.TimeoutExpired` does not catch:
   ```
   ip = ipaddress.ip_address(parts[0])
   ValueError: 'receive' does not appear to be an IPv4 or IPv6 address
   ```

Both are intermittent and load-dependent; the number of monitored interfaces raises the odds. This is the root cause behind the symptom previously worked around at the PHP layer in #4611 (which only added an `is_array($data)` guard around the `foreach`).

### Fix

Require the expected column count and validate `parts[0]` as an IP before using the row; skip anything else. Also guard the leading-index `pop()` against an empty split, and drop the incorrect `TimeoutExpired` handler now that the address is validated up front.

### Testing

- Reproduced **both** crashes on a live firewall under concurrent load (multiple Diagnostics/Traffic tabs / concurrent `interface show top` invocations).
- Unit-tested the parsing against representative `iftop` output: valid IPv4 and IPv6 rows parse correctly with the right `local`/`private` tags, while truncated rows and non-host lines are skipped with no exception. The pre-patch parser raises `IndexError`/`ValueError` on the same input.

### DIFF
--- a/src/opnsense/scripts/interfaces/traffic_top.py
+++ b/src/opnsense/scripts/interfaces/traffic_top.py
@@ -135,8 +135,18 @@ if __name__ == '__main__':
         for line in iftop_data[intf].split('\n'):
             if line.find('=>') > -1 or line.find('<=') > -1:
                 parts = line.split()
-                if parts[0].find('.') == -1 and parts[0].find(':') == -1:
+                if parts and parts[0].find('.') == -1 and parts[0].find(':') == -1:
                     parts.pop(0)
+                # skip malformed/truncated iftop lines; under load iftop may
+                # emit partial rows (fewer columns) or non-host output that
+                # still contains "=>"/"<=". Require the expected column count
+                # and a valid host address before using the row.
+                if len(parts) < 6:
+                    continue
+                try:
+                    ip = ipaddress.ip_address(parts[0])
+                except ValueError:
+                    continue
                 item = {
                     'address': parts[0],
                     'rate': parts[2],
@@ -146,14 +156,10 @@ if __name__ == '__main__':
                     'tags': []
                 }
                 # attach tags (type of address)
-                try:
-                    ip = ipaddress.ip_address(parts[0])
-                    if ip in all_local_addresses:
-                        item['tags'].append('local')
-                    if ip.is_private:
-                        item['tags'].append('private')
-                except subprocess.TimeoutExpired:
-                    pass
+                if ip in all_local_addresses:
+                    item['tags'].append('local')
+                if ip.is_private:
+                    item['tags'].append('private')
 
                 if line.find('=>') > -1:
                     other_end = item


AdSchellevis:
@brendanbank thanks, looks like a safe addition.

PR #10579, opened 2026-07-19
mvc: translate backend system status messages

Opnwall:
**Important notices**

- Model used: OpenAI Codex (GPT-5)
- Extent of AI involvement: Assisted with root-cause analysis, preparing the minimal patch, and drafting the pull request description. The contributor reviewed the change and operationally tested it on OPNsense 26.7.

---

**Describe the problem**

System status messages are generated by a configd backend script that does
not use the language selected by the current GUI user. Consequently,
messages available in the gettext catalog are returned to the GUI in English.

---

Translate the backend-provided status message and title in the API
controller, where the current user's locale is available.

This preserves the backend status format and allows static status messages,
such as the Unbound manual override warning, to use the selected GUI language.

Tested with:

- `git diff --check`
- PHP syntax validation
- Operational verification on OPNsense 26.7 using the `zh_CN` locale

---

**Related issue**

None; this is a small localization fix discovered during operational testing.

### DIFF
--- a/src/opnsense/mvc/app/controllers/OPNsense/Core/Api/SystemController.php
+++ b/src/opnsense/mvc/app/controllers/OPNsense/Core/Api/SystemController.php
@@ -86,6 +86,10 @@ class SystemController extends ApiControllerBase
             $acl = new ACL();
 
             foreach ($statuses as $subsystem => $status) {
+                $statuses[$subsystem]['message'] = gettext($status['message']);
+                if (!empty($status['title'])) {
+                    $statuses[$subsystem]['title'] = gettext($status['title']);
+                }
                 unset($statuses[$subsystem]['scope']);
                 $statuses[$subsystem]['status'] = $order[$status['statusCode']];
                 if (!empty($status['location'])) {


AdSchellevis:
looks like a similar case as https://github.com/opnsense/core/pull/8935, I'll leave it up to @swhite2 as he worked on this part earlier.

swhite2:
Given the discussion in #8935 and https://github.com/opnsense/core/pull/8935#issuecomment-3196557612 which would have been the most obvious fix, this looks like the right fix indeed

src/opnsense/mvc/app/controllers/OPNsense/Core/Api/SystemController.php
```diff
@@ -86,6 +86,10 @@ public function statusAction()
             $acl = new ACL();
 
             foreach ($statuses as $subsystem => $status) {
+                $statuses[$subsystem]['message'] = gettext($status['message']);
+                if (!empty($status['title'])) {
```
swhite2, line 90: I think I'd rather set a default empty string

```
    protected $internalTitle = '';
```

in `AbstractStatus.php` so we can skip the empty check. Both title and message should always be set anyway.

AdSchellevis:
@Opnwall thanks, merged in https://github.com/opnsense/core/commit/2df0ac3478a43a5ef8e2b75584856f87d8001926

Issue #10539, opened 2026-07-16
firewall: common collector for source NAT automatic rule display


smart7324:
**Important notices**

Before you add a new report, we ask you kindly to acknowledge the following:

- [X] I have read the contributing guidelines at https://github.com/opnsense/core/blob/master/CONTRIBUTING.md
- [X] I am convinced that my issue is new after having checked both open and closed issues at https://github.com/opnsense/core/issues?q=is%3Aissue
- [X] AI tools were used to create at least part of the text submitted herewith.

If AI was used, please disclose:

- Model used: ChatGPT 5.5
- Extent of AI involvement: Text generation out of voice transcription

**Describe the bug**

After upgrading from OPNsense 26.1.11 to OPNsense 26.7, the former **Outbound NAT** section is now shown as **Source NAT**.

I do not have any manually configured NAT rules. I use both WireGuard and OpenVPN, and both VPNs are assigned as interfaces according to the [WireGuard Road Warrior Setup documentation](https://docs.opnsense.org/manual/how-tos/wireguard-client.html#step-4-a-assign-an-interface-to-wireguard-recommended).

The assigned interfaces are:

- `NET_VPN`
- `NET_VPN_TCP`

The documentation states that assigning the VPN as an interface automatically adds the required IPv4 Outbound NAT rule, so no manual IPv4 NAT rule is required.

IPv4 connectivity through both VPNs works correctly.

However, in the previous Outbound NAT view, the automatically generated rules showed the assigned VPN interfaces/networks. In OPNsense 26.7, these two VPN interfaces (`NET_VPN` and `NET_VPN_TCP`) are no longer displayed in the corresponding automatically generated Source NAT rule information.

This appears to be a possible UI/display issue. The NAT functionality itself appears to work correctly.

**To Reproduce**

1. Configure WireGuard and OpenVPN.
2. Assign both VPNs as interfaces.
3. Use the assigned interfaces `NET_VPN` and `NET_VPN_TCP`.
4. Do not create any manual IPv4 NAT rules.
5. Go to **Firewall → NAT → Source NAT** in OPNsense 26.7.
6. View the automatically generated Source NAT rules.
7. Observe that the assigned VPN interfaces/networks are not displayed in the corresponding automatically generated rule information.

**Expected behavior**

The automatically generated Source NAT rule information should display the assigned VPN interfaces/networks (`NET_VPN` and `NET_VPN_TCP`) if they are covered by the automatically generated rules.

[...]

PR #10590, opened 2026-07-21
firewall: unify automatic source NAT networks #10539

fichtner:
Bring a bit of structure into this legacy code: move the "plugin" parts to its own "oubound_net" run target, collect it with the new filter_auto_source_nat() and just iterate over it from the code that needs this.  We do all of this to provide a possible configd target to expose the actual automatic rules skeleton to the MVC source NAT GUI.

```
      # pluginctl -r outbound_nat
      {
        "pf": {
          "wan": {
            "lan": "LAN networks",
            "lo0": "Loopback networks",
            "127.0.0.0/8": "127.0.0.0/8"
          }
        }
      }
```

### DIFF
--- a/src/etc/inc/filter.inc
+++ b/src/etc/inc/filter.inc
@@ -206,38 +206,21 @@ function filter_configure_sync($verbose = false, $load_aliases = true)
     }
 
     /* automatic outbound rules */
-    if (
-        empty($config['nat']['outbound']['mode']) ||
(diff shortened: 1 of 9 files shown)


AdSchellevis:
Initially I was thinking to add this into the `plugins_firewall` structure, but realistically I don't expect a lot of magic to appear in these hooks, in which case the run construct should be fine too.

PR #10594, opened 2026-07-22
mvc: create runInterfaceRegistation and use it from all required spots

fichtner:
1. Fix the stale cache after applying new interfaces/devices in existing paths
2. Allow to call it from controllers who do not use invokeInterfaceRegistration()
3. Add device applies to the mix

### DIFF
--- a/src/etc/inc/config.inc
+++ b/src/etc/inc/config.inc
@@ -211,14 +211,7 @@ function write_config($desc = '', $backup = true)
         }
     }
 
-    /**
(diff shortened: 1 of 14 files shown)


Monviech:
I think I noticed that while testing bridges. When creating one it took a while to turn up as selectable in the assignment page.

I guess this PR fixes that.

fichtner:
Correct. Still need to take this for a test drive this afternoon but I don't expect surprises.

Issue #10599, opened 2026-07-23
Assignment add_item endpoint returns the temporary internal UUID as uuid instead of the final identifier


kemoycampbell:
**Important notices**

Before you add a new report, we ask you kindly to acknowledge the following:

- [ X] I have read the contributing guidelines at https://github.com/opnsense/core/blob/master/CONTRIBUTING.md
- [ X] I am convinced that my issue is new after having checked both open and closed issues at https://github.com/opnsense/core/issues?q=is%3Aissue
- [ ] AI tools were used to create at least part of the text submitted herewith.

If AI was used, please disclose:

- Model used:
- Extent of AI involvement:

**Describe the bug**

Upon creating an interface the using the assignment API, the endpoint returns
`{
    "result": "saved",
    "uuid": "a1ecb292-243d-43c8-b09b-268921dad206"
}`

However, the remaining Assignment API endpoints expect the interface name (e.g. `opt1`) rather than the generated UUID returned above.

Upon investigating, it was found that `ArrayField` generates a temporary UUID when creating the interface assignment. Later, during `NetworkInterface::serializeToConfig()`, the final interface identifier (`opt1`, `opt2`, etc.) is generated and written to the configuration. However, the in-memory UUID attribute is not updated to reflect the final interface identifier.

In addition, `search_item` shows that the expected UUIDs for interface assignments are actually the generated interface names rather than the temporary internal UUID.

```
{
    "rows": [
        {
            "uuid": "wan",
            "descr": "",
            "%descr": "WAN",
            "identifier": "wan",
            "icon": "fa fa-plug text-success",
            "optgroup": "hardware",
            "if": "vtnet0",
            "%if": "vtnet0 (52:54:00:12:34:56)",
            "lock": "0"
        },
        {
            "uuid": "opt1",
            "descr": "",
            "%descr": "OPT1",
            "identifier": "opt1",
            "icon": "fa fa-plug text-success",
            "optgroup": "vlan",
            "if": "vlan0.1",
            "%if": "vlan0.1  (Parent: vtnet0, Tag: 100)",
            "lock": "0"
        }
    ],
    "rowCount": 2,
    "total": 2,
    "current": 1
}
```

This bug was introduced in the release of the `Assignment API` in version `26.7`

**To Reproduce**

1. Go to 'Interfaces'
2. Click on 'Assignments'
3. Open the network tab
4. Click on the `+` to create a new interface
5. Create a new assignment and click on save
6. Observe the JSON response with the uuid

[...]

PR #10600, opened 2026-07-23
modify the serializeToConfig to set the uuid of the new network ident…

kemoycampbell:
**Important notices**

- Model used:
- Extent of AI involvement:

---

**Describe the problem**
Upon creating an interface the using the assignment API, the endpoint returns
`{
    "result": "saved",
    "uuid": "a1ecb292-243d-43c8-b09b-268921dad206"
}`

However, the remaining Assignment API endpoints expect the interface name (e.g. `opt1`) rather than the generated UUID returned above.

Upon investigating, it was found that `ArrayField` generates a temporary UUID when creating the interface assignment. Later, during `NetworkInterface::serializeToConfig()`, the final interface identifier (`opt1`, `opt2`, etc.) is generated and written to the configuration. However, the in-memory UUID attribute is not updated to reflect the final interface identifier.

In addition, `search_item` shows that the expected UUIDs for interface assignments are actually the generated interface names rather than the temporary internal UUID.

```
{
    "rows": [
        {
            "uuid": "wan",
            "descr": "",
            "%descr": "WAN",
            "identifier": "wan",
            "icon": "fa fa-plug text-success",
            "optgroup": "hardware",
            "if": "vtnet0",
            "%if": "vtnet0 (52:54:00:12:34:56)",
            "lock": "0"
        },
        {
            "uuid": "opt1",
            "descr": "",
            "%descr": "OPT1",
            "identifier": "opt1",
            "icon": "fa fa-plug text-success",
            "optgroup": "vlan",
            "if": "vlan0.1",
            "%if": "vlan0.1  (Parent: vtnet0, Tag: 100)",
            "lock": "0"
        }
    ],
    "rowCount": 2,
    "total": 2,
    "current": 1
}
```

---

The fix updates the in-memory UUID attribute when the final interface identifier is generated in `NetworkInterface::serializeToConfig()`. This allows the `addBase` to return the final interface identifier as the  UUID instead of the temporarily internally generated UUID.

---

**Related issue**

Related to #10599

### DIFF
--- a/src/opnsense/mvc/app/models/OPNsense/Interfaces/NetworkInterface.php
+++ b/src/opnsense/mvc/app/models/OPNsense/Interfaces/NetworkInterface.php
@@ -149,7 +149,7 @@ class NetworkInterface extends BaseModel
                 /* We want the node to return the new network identifier as the uuid not
                 the internal random generated uuid by the ArrayField.
                 Assignments use identifier as the uuid in the config.xml */
-                $node = $this->getNodeByReference('interface.' . $key);
+                $this->getNodeByReference('interface.' . $key)?->setAttributeValue('uuid', $newIdentifier);
                 if ($node === null) {
                     throw new ModelException('Failed to locate interface node ' . $key . '
                     while updating uuid to the new network identifier');
--- a/src/opnsense/mvc/app/models/OPNsense/Interfaces/NetworkInterface.php
+++ b/src/opnsense/mvc/app/models/OPNsense/Interfaces/NetworkInterface.php
@@ -30,6 +30,7 @@ namespace OPNsense\Interfaces;
 
 use OPNsense\Base\BaseModel;
 use OPNsense\Base\Messages\Message;
+use OPNsense\Base\ModelException;
 use OPNsense\Core\Config;
 use OPNsense\Core\FileObject;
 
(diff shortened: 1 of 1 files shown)


src/opnsense/mvc/app/models/OPNsense/Interfaces/NetworkInterface.php
```diff
@@ -137,12 +138,23 @@ public function serializeToConfig($validateFullModel = false, $disable_validatio
...
+                $newif = Config::getInstance()->object()->interfaces->addChild($newIdentifier);
                 $newif->if = $intf['if'];
                 $newif->descr = $intf['descr'];
                 $newif->lock = $intf['lock'];
                 $next_if++;
+
+                /* We want the node to return the new network identifier as the uuid not
+                the internal random generated uuid by the ArrayField.
+                Assignments use identifier as the uuid in the config.xml */
+                $node = $this->getNodeByReference('interface.' . $key);
```
AdSchellevis:
```suggestion
                $this->getNodeByReference('interface.' . $key)?->setAttributeValue('uuid', $newIdentifier);
```

I'm not a huge fan of this hack, but if we simplify it I don't mind that much. If it can't map it, it's rather irrelevant, we can't handle the exception in any useful way here.

kemoycampbell: I agree that this is a bit of a hack. Ideally, I was thinking about having some kind of protected identifier hook in the base model that child models could override. The parent implementation could continue generating UUIDs as normal unless a child model provides its own identifier generation logic.

However, that felt more invasive and closer to a refactor-level change rather than something appropriate for this fix.

I agree that this can be simplified, and the exception can be removed since there is no useful recovery path at this point in the process.

AdSchellevis: @kemoycampbell This is mainly caused by the fact that we can't really change the identifiers for the interfaces, so a small hack to serve the response is not a huge issue and likely a one-off anyway. I'll merge the current state, thanks for taking the effort.

fichtner:
do we still need model exception in 68c0174?

src/opnsense/mvc/app/models/OPNsense/Interfaces/NetworkInterface.php
kemoycampbell: Sorry for the confusion with the word "identifier". In my previous remark, I used it poorly or rather interchange with 'uuid' as in "unless a child provides its own uuid generation logic". But thank you for the review, merging and for all the works you folks do on opnsene.

kemoycampbell:
@fichtner, i think we can live without it. The simpify code should be able to find the node reference and assign the identifier for the interface as the uuid. The only time I can think it may fail silence is when there is typo or incorrect look up, mostly on us as developers not the user part.

AdSchellevis:
@fichtner missed it, thanks!

fichtner:
too late 221f639176 :)

PR #10617, opened 2026-07-26
firewall: remove 2a10::/12 from bogonsv6.sample since it was de-bogonized in 2020

Belgarion:
Spent some time troubleshooting why incoming traffic was being blocked by the bogon rules after upgrade of OPNsense, and found that the system was using the list from bogonsv6.sample which contains entries that are no longer bogons.

2a10::/12 started being used for real network in 2020: https://www.ripe.net/about-us/news/de-bogonising-2a10-12/


fichtner:
Merged, thanks! The bogons samples are a bit archaic in nature. I don't have a good plan to replace these yet but eventually they should disappear from this code.


Cheers,
Franco

PR #10657, opened 2026-08-04
interfaces: actually release the IPv4 DHCP lease on teardown

### LINKED ISSUE #3628: DHCP Release on the Interfaces Overview page does not send DHCPRELEASE
**Important notices**

[x] I have read the contributing guide lines at https://github.com/opnsense/core/blob/master/CONTRIBUTING.md

[x] I have searched the existing issues and I'm convinced that mine is new.

**Describe the bug**
When clicking on "Release" on the Interfaces: Overview Page DHCP Clients are stopped but the IP is not returned to the server.

**To Reproduce**
1. Go to Interfaces: Overview
2. Click on Release on DHCP: from the WAN interface
3. Do a packet capture on the DHCP server
4. No DHCPRELEASE is sent on either DHCPv4 or DHCPv6

**Expected behavior**
DHCPRELEASE should be sent for DHCPv4 and DHCPv6 leases and PDs

**Relevant log files**
Aug 5 21:31:56	opnsense: /status_interfaces.php: Clearing states for stale wan route on vtnet1
Aug 5 21:31:56	dhcp6c[91668]: exiting
Aug 5 21:31:55	dhclient[87566]: exiting.
Aug 5 21:31:55	dhclient[87566]: connection closed

**Additional context**
See PCAP in the attachments (Actually do not, there is nothing to see except a bit of NTP and DNS traffic).
When renewing the leases I can see the traffic in a PCAP so I am confident that I am capturing correctly.

**Environment**
OPNsense 19.7.2 on Proxmox 6

SilentandUnknown:
FreeBSD base /sbin/dhclient has no release mode (getopt string is "bc:dl:np:qu" -- confirmed against its source, DHCPRELEASE does not appear anywhere in it). interface_reset() currently just kills the dhclient/daemon(8) supervisor pair via SIGTERM, so the ISP-side lease is never actually freed -- a subsequent renew often just hands back the same address. See opnsense/core#3628 (2019, still open on current master).

A prior attempt to fix this, PR #3275, stalled and was closed as "timeout" rather than on a clean technical rejection, but the thread did surface one real, legitimate objection: it shipped a checked-in binary (a swapped-in ISC dhclient), which is a genuine maintenance/ security liability for a firewall OS. It also carried ~10 lines of unrelated theme/CSS changes for new GUI buttons, which drew separate maintainer friction.

This patch avoids both: no binary, no package dependency, and no www/theme files touched at all. Instead of swapping clients, it hand- builds and sends a correct DHCPRELEASE (RFC 2131 4.4.4) from a small in-tree Python helper, using the last lease recorded in /var/db/dhclient.leases.<if>.

This is a real correctness fix to existing behavior, not a new feature -- interface_reset() is already called from real interactive paths (disabling an interface, deleting one, console re-addressing; see src/www/interfaces.php, apply_pending_if_changes.php, setaddr.php, console.inc), and $suspend already correctly distinguishes those genuine-release actions from routine reconfigures. No new UI is added or required.

Verified via live packet capture in an isolated VM lab (OPNsense 26.7 nano image in QEMU, WAN on a private socket network, dnsmasq + tcpdump standing in for the ISP, zero contact with any real network): interface_reset('wan') emits a correctly-formed, RFC-2131-compliant DHCPRELEASE, unicast from the leased address to the recorded server, padded to the real dhclient's 300-byte BOOTP minimum.

All three client-identifier (option 61) forms interfaces.inc can generate were exercised live in that lab, each confirmed both on the wire and by the server's own reaction:

  config                          option 61 sent        dnsmasq
  ------------------------------- ---------------------- -----------
  (none)                          01:52:54:00:00:00:01   lease freed
  dhcphostname "opnlab-cid-7"     raw string, 12 bytes    lease freed
  adv send 01:aa:bb:cc:dd:ee:ff   01:aa:bb:cc:dd:ee:ff    lease freed

As a negative control, a release deliberately sent with the wrong (MAC-derived) identifier while the lease was bound to a custom one was logged by dnsmasq as "unknown lease" and the binding was NOT released -- confirming the identifier handling is load-bearing, not cosmetic.

Implementation notes:
- The release must run after dhclient's supervisor is confirmed gone (it holds port 68) but before any address flush, since the release must be sent from the still-assigned leased address per RFC 2131.
- Failures during the release attempt are not muted: real errors (bad args, exec failures, timeouts) reach syslog, while the specific benign cases -- address already gone after an expired lease or a downed link (EADDRNOTAVAIL, ENETUNREACH, ENETDOWN, EHOSTUNREACH) -- are treated as a normal no-op rather than logged as an error.
- The release call is wrapped in a hard timeout, since RFC 2131 releases are fire-and-forget with no defined reply, and must not be able to stall the caller on a dead link.
- client_id() parses both value forms dhclient's own parser accepts for a byte-string option, since interfaces.inc can emit either: a quoted string (from dhcphostname, no RFC 2132 type byte) or a colon-hex byte list (reachable via adv_dhcp_send_options). Anything it can't confidently parse falls back to the MAC-derived type-1 form, which is what dhclient sends when told nothing.

Fixes: #3628

**Important notices**

- Model used:
- Extent of AI involvement:
Model used: Claude (Anthropic), via Claude Code — Sonnet and Opus across different phases of the work.
  - Extent of AI involvement: All code, the VM-lab test harness, and live packet-capture verification were
  performed by Claude Code agents. A human directed the work throughout — scoping the fix, deciding what to
  test, reviewing each verification pass, and deciding whether/how to submit — but did not hand-write the
  patch or test scripts.

### DIFF
--- a/src/etc/inc/interfaces.inc
+++ b/src/etc/inc/interfaces.inc
@@ -883,7 +883,19 @@ function interface_reset($interface, $ifacecfg = false, $suspend = false)
 
     switch ($ifcfg['ipaddr'] ?? 'none') {
         case 'dhcp':
+            /*
+             * Stop the supervisor first: it owns the restart timer, so
+             * signalling dhclient directly would only get it respawned.
+             * daemon(8) forwards the TERM to dhclient and exits.
+             */
             killbypid("/var/run/dhclient.{$device}.pid");
+            /*
+             * Only a real teardown gives up the lease; a suspend is a
+             * temporary reconfigure and must keep the address claimed.
+             */
+            if (!$suspend) {
+                interface_dhcp_release($device);
+            }
             break;
         default:
             break;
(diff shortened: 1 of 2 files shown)


Monviech:
This is the wrong abstraction layer. If you want dhclient(8) to do something it currently can't, you should go upstream.

https://bugs.freebsd.org/bugzilla/

PR #10672, opened 2026-08-06
Firewall: Add missing TLS ports to well-known ports

Self-Hosting-Group:
Add some missing TLS ports to well-known ports using https://www.iana.org/assignments/service-names-port-numbers


fichtner:
Removing ports is a bad idea as it breaks rules loading for no visible benefit. Adding new ones should be avoided too, especially if not in /etc/services. The whole construct of well known into pf is too fragile as FreeBSD likes to maintain /etc/services which breaks backward compat much like this file here is edited now.

fichtner (changes requested):
See other comment.

Self-Hosting-Group:
> Removing ports is a bad idea as it breaks rules loading for no visible benefit. Adding new ones should be avoided too, especially if not in /etc/services. The whole construct of well known into pf is too fragile as FreeBSD likes to maintain /etc/services which breaks backward compat much like this file here is edited now.

`nat-stun-port` was renamed to `stun`, the registered name, and `urd` to `submissions`, and no port numbers where removed. If wished entries with the old and current names could be inserted. Some are only in the IANA list and not yet in `/etc/services` on FreeBSD.

fichtner:
This file is to retain the old names in the config.xml since lost by such well-intentioned renames in /etc/services. The proper way to do this is rendering all ports as their port number but assisting with the services names in the GUI. It’s a much larger effort. Just getting the proper historic compat correct here cost some significant investigation time and editing of the list at hand. You can see from the git history of this file.

Self-Hosting-Group:
I like the idea of displaying the service name next to the port, but I also like the current lookup by name in the UI.

Now, no service names are renamed, but rather duplicated (`urd`/`submissions`), and only entries that are present in FreeBSD's `/etc/services` file are added with FreeBSD's names (`sips` -> `sip-tls`). Would that be a good intermediate step, until such an extended UI exists, to merge this newly pushed commit? Otherwise, let's close this PR.

Self-Hosting-Group:
Note: According to history, `nat-stun-port` was removed, then readded later, to the list, but is not in `/etc/services`

fichtner:
I’ll check next week when I’m back.

fichtner (approved):
All of these are in /etc/services so we can merge this as is. Thanks!

Self-Hosting-Group:
Thank you for merging. The newly added service-name `sip-tls` is the old name, since 2014 registered at IANA as `sips`. I think it's probably better to introduce the current name, as I'm also trying to change it in `/etc/services` upstream for FreeBSD.

fichtner:
It's really not worth the effort since this will break the soon-to-be consumers of sip-tls. As I said this needs a larger overhaul.

Self-Hosting-Group:
OK. However, I have already submitted this change to FreeBSD, along with further corrections, as I have already done for Debian. Note: An existing entry for, e.g., `nat-stun-port`, is missing from `/etc/services` and seems to not break.

Or remove the newly added (unreleased) service-name (`sip-tls`) completely?

fichtner:
26.7.3 is more or less finished for today already and /etc/services usually won't be picked from stable/15 backports -- only if we upgrade the OS and even then only if it was actually backported by FreeBSD.

Self-Hosting-Group:
Can the daemon security update (longer open ports PR) be included in version 26.7.3?

fichtner:
No, there's a ton of other non-community work around at the moment.

Issue #10677, opened 2026-08-06
Kea DHCPv6: PD route injection fails on bridge interfaces due to Hostwatch conflict


ybayart:
**Important notices**

Before you add a new report, we ask you kindly to acknowledge the following:

- [x] I have read the contributing guidelines at https://github.com/opnsense/core/blob/master/CONTRIBUTING.md
- [x] I am convinced that my issue is new after having checked both open and closed issues at https://github.com/opnsense/core/issues?q=is%3Aissue
- [ ] AI tools were used to create at least part of the text submitted herewith.

**Describe the bug**

The usage of Hostwatch instead of direct NDP parsing, introduced in https://github.com/opnsense/core/pull/9648, breaks PD route injection when the lease is on a bridged interface.

In my setup, `bridge0` contains `ixl0` and `ixl2`. In the Kea logs, I can see this error message:
```
failed adding route 2001:db8:1234:5678::/58 -> fe80::xxxx:yyyy:zzzz:ba01%ixl0
```
This fails because the physical member interface (ixl0) is used instead of the logical bridge interface (bridge0). Considering the logic here: https://github.com/opnsense/core/blob/704351e7d8cda7f40cf40272c2e2102c9ab52725/src/opnsense/scripts/kea/kea_prefix_watcher.py#L88-L90

Because Hostwatch detects traffic on the physical member, list_hosts.py returns multiple entries for the same MAC address. The loop simply overwrites _def_local_db with the last matching entry, which happens to be the wrong interface (the physical one) and sometimes the wrong link-local address.

Output of list_hosts.py using Hostwatch:
```
# /usr/local/opnsense/scripts/interfaces/list_hosts.py --proto inet6 --ndp | jq | grep -C1 'xx:xx:yy:yy:b9:ff'
      "bridge0",
      "xx:xx:yy:yy:b9:ff",
      "2001:db8:1234:5678::" #> not link-local
--
      "bridge0",
      "xx:xx:yy:yy:b9:ff",
      "2001:db8:1234:5678:xxxx:yyyy:zzzz:b9ff" #> not link-local
--
      "bridge0",
      "xx:xx:yy:yy:b9:ff",
      "fe80::xxxx:yyyy:zzzz:b9ff" #> saves this in _def_local_db (the correct one)
--
      "bridge0",
      "xx:xx:yy:yy:b9:ff",
      "fe80::xxxx:yyyy:zzzz:ba01" #> overwrites _def_local_db with a wrong address
--
      "ixl0",
[...]

PR #10678, opened 2026-08-06
feat: Kea DHCPv6 - Hostwatch can be disabled

ybayart:
**Important notices**

---

**Describe the problem**

Kea DHCPv6 Prefix Delegation route injection fails when operating on a bridged interface. The kea_prefix_watcher.py script relies on list_hosts.py, which uses Hostwatch data by default.

On a bridge setup, Hostwatch associates traffic with the underlying physical member interface rather than the logical bridge interface. As a result, list_hosts.py returns multiple conflicting entries, causing the prefix watcher to attempt injecting the PD route onto the physical interface instead of the bridge, which fails at the OS level.

---

Add an advanced configuration option in the GUI to bypass Hostwatch data and force reliance on NDP tables for prefix routing.

Changes:
1. **Model & Form**: Added a prefix_watcher_discover boolean field and exposed it as an advanced checkbox (Bypass Hostwatch) in the Kea DHCPv6 form.
2. **Script Logic**: Updated kea_prefix_watcher.py to read this new setting. If enabled, it appends the --discover flag when calling list_hosts.py.

---

**Related issue**

https://github.com/opnsense/core/issues/10677

### DIFF
--- a/src/opnsense/mvc/app/controllers/OPNsense/Kea/forms/generalSettings6.xml
+++ b/src/opnsense/mvc/app/controllers/OPNsense/Kea/forms/generalSettings6.xml
@@ -69,6 +69,13 @@
         <help>Automatically add a basic set of firewall rules to allow dhcp traffic, more fine grained controls can be offered manually when disabling this option.</help>
     </field>
     <field>
+        <id>dhcpv6.general.prefix_watcher_discover</id>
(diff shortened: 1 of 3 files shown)


Monviech:
I don't think it's a good idea, let's discuss this first in your issue.

A fallback already exists when hostwatch is disabled.

PR #10687, opened 2026-08-07
fix (radvd): pref64 lifetime too low, causing CLAT dropouts

BPplays:
**Important notices**

- Model used:
- Extent of AI involvement:

---

**Describe the problem**

the PREF64 lifetime doesn't follow `Valid Lifetime`, and with my configuration was only 16s (a scaled lifetime of 2) way lower then my valid or even preferred lifetime, and was causing the CLAT to drop out on my laptop

---

add `AdvValidLifetime` to the `nat64prefix` block
```
nat64prefix * {
    AdvValidLifetime 15000;
};
```

this fixes the scaled lifetime as verified by wireshark and despite being hard to find it's documented [here](https://github.com/radvd-project/radvd/blob/730e1c959b90f0c2f70e2854ac4e871f63822e03/radvd.conf.5.man#L832)

### DIFF
--- a/src/etc/inc/plugins.inc.d/radvd.inc
+++ b/src/etc/inc/plugins.inc.d/radvd.inc
@@ -306,6 +306,9 @@ function radvd_configure_do($verbose = false, $ignorelist = [])
 
         if (!$entry->nat64prefix->isEmpty()) {
             $radvdconf .= "    nat64prefix {$entry->nat64prefix->getValue()} {\n";
+            if (!$entry->AdvValidLifetime->isEmpty()) {
+                $radvdconf .= "        AdvValidLifetime {$entry->AdvValidLifetime->getValue()};\n";
+            }
             $radvdconf .= "    };\n";
         }
 


fichtner:
> with my configuration was only 16s

I'm doubting this a bit:

https://github.com/radvd-project/radvd/blob/730e1c959b90f0c2f70e2854ac4e871f63822e03/radvd.conf.5.man#L695-L702

The default computed value should be 600 * 3 in our case, which matches the documentation's default value.


Cheers,
Franco

BPplays:
@fichtner i set `MaxRtrAdvInterval` = 5s to see if some network problems i was having were caused by unreliable multicast, and 5 * 3 would round up to 16 here;
but in any case i don't see the point in the lifetime for PREF64 being fixed to 3 RAs, it essentially makes it pointless to adjust the ratio between the RA interval and the lifetime, normally raising the amount of RAs sent before the lifetime expires would make things more resilient against loss but here it has no effect if you're using PREF64

fichtner:
@BPplays so that's exactly how it is intended to work upstream. If I recall correctly pfSense added a separate `AdvValidLifetime` field for PREF64, but I'm still not convinced that is a good idea seeing the numbers in radvd add up correctly, but it might be the best way to deal with your configuration quirk. if you insist on it.

BPplays:
@fichtner 
> but it might be the best way to deal with your configuration quirk. if you insist on it.

the problem is more that this affects any environment with high loss (weak or crowded WI-FI for example), my problem isn't that the lifetime is short with my config, but that it only accounts for 3 RAs, this is a problem in any environment where losing 3 RAs in a row is somewhat probable

> If I recall correctly pfSense added a separate `AdvValidLifetime` field for PREF64

i didn't add a separate field here because i can't think of a reason you'd want this set differently to the base `Valid Lifetime`, setting it lower would just cause a CLAT to drop out before IPv6 in any case where the difference mattered so i can't see a use case

PR #10695, opened 2026-08-09
menu: fix for Favorites entries navigation when sidebar collapsed

Greelan:
**Important notices**

- Model used:
- Extent of AI involvement:

---

**Describe the problem**

With the menu sidebar collapsed to icons, hovering a Favorites entry instantly navigates to it, so you can't scroll the list. The collapsed sidebar simulates a click on hover (to open the submenu). Ordinary links ignore this because jQuery won't follow an href from a simulated click, but Favorites (given its different link/click handling) initiates a real native click on the underlying item. 

---

Ensure the handler ignores simulated hover-clicks and only acts on genuine clicks.

---

**Related issue**

### DIFF
--- a/src/opnsense/www/js/opnsense_favorites.js
+++ b/src/opnsense/www/js/opnsense_favorites.js
@@ -68,6 +68,7 @@ $(document).ready(function () {
 
     $favPanel.on('click', '.list-group-item', function (e) {
         e.preventDefault();
+        if (!e.originalEvent) return;
         const cls = getRefClass($(this));
         if (cls) {
             $favPanel.collapse('hide');


Greelan:
@AdSchellevis a fix for a slight anomaly I noticed (I don't normally have the menu sidebar collapsed so missed this behaviour).

AdSchellevis:
I think there's something off with how we render these items at the moment, which results in the sidebar assuming these are containers, couldn't find it yet, but this should fix the main issue as these events are passed through which we can detect like this indeed.

Greelan:
> I think there's something off with how we render these items at the moment, which results in the sidebar assuming these are containers

This? 
https://github.com/opnsense/core/blob/ae0088be989720b45c64e359d569a321ddc6454c/src/opnsense/www/js/opnsense_theme.js#L38

Maybe it should filter by collapse toggle so it doesn't pick up leaf items

AdSchellevis:
could be the issue, when looking at the dom for the favorites section it looks like this on my end:
```
<div class="collapse in" id="Favorites" aria-expanded="true" style="">
       <a href="#" class="list-group-item menu_ref_7e46272fe380827861cbaf5b484c43c9" data-parent="#Favorites" aria-expanded="false">Interfaces: [LANWAN]: [LAN]</a>
        <a href="#" class="list-group-item menu_ref_b62a8bd502070e3930a5e25cec618d75">Interfaces: [LANWAN]: [WAN]</a>
</div>
````

Similar construct for Lobby looks like:
```
<div class="collapse " id="Lobby" aria-expanded="false">
          <a href="/ui/core/dashboard" class="list-group-item menu_ref_0bba82bc88435834cd2ac6db1b8fa813 ">
            <div style="display: table;width: 100%;">
              <div style="display: table-row">
                <div style="display: table-cell">Dashboard</div>
                  <div style="display: table-cell; text-align:right; vertical-align:middle;"><span class="fa fa-dashboard fa-fw"></span></div>
              </div>
            </div>
          </a>

          <a href="/ui/user_portal" class="list-group-item menu_ref_ce03aec1a34649276e52e765f2bfeae0 ">
            <div style="display: table;width: 100%;">
              <div style="display: table-row">
                <div style="display: table-cell">Portal</div>
                  <div style="display: table-cell; text-align:right; vertical-align:middle;"><span class="fa fa-key fa-fw"></span></div>
              </div>
            </div>
          </a>

...
                                
</div>
````

My best guess is the sidebar tries to point to the inner diffs, but while fiddling with it I had some other issues, making sure both use the same layout is likely a good starting point to debug further.

Greelan:
I don't think the layout difference is the cause, and anyway there are reasons for the difference (eg not replicating icons on Favorites entries since not all leaf menu items have them, so it would look weirdly inconsistent)

fichtner:
@Greelan I'm late to the party... nice work on the favourites. I have one nitpick that I'd like to get off my chest:

feels like the star is misplaced at the far right side behind other action buttons. it would feels much more natural alongside the page name title (likely trailing it directly instead of in front but both could be options at first glance).

Not sure if this was discussed before...


Cheers,
Franco

Greelan:
I tried it at one early point (after first having it actually in the menu itself) to the far left but felt it got in the way of reading the page name as the primary piece of information. And I felt that having it right after the name meant it would "jump" around depending on the text length so wouldn't be as intuitively findable. 

It felt right to have it in a consistent place alongside other icons (where they exist) - the mouse always goes to the same spot. 

Not really discussed, no.

Greelan:
Well, some discussion on the forum:

https://forum.opnsense.org/index.php?msg=263392
https://forum.opnsense.org/index.php?msg=263433
https://forum.opnsense.org/index.php?msg=263458
https://forum.opnsense.org/index.php?msg=263463

And at least 2 people endorsed the option I ended up with xD

Greelan:
Another rationale for not having the star icon in the user's face is that once the Favorites list is created, it would generally be pretty static. That's been my experience generally. So it's not needed to have each leaf page screaming out its Favorites status. The star at the far right is less obtrusive (plus the other benefits I mentioned).

fichtner:
Technically and psychologically I have to disagree. The dissociative state between the page name and the favourites button behind e.g. a service widget seems impractical. Yes it's always on the right but it also pushes content to the left that it doesn't belong to if you ignore the dashboard page for example. A clearer separation of concerns would be nice.

Greelan:
@fichtner do I take it, given Favorites was skipped for 26.7.2, that this is now a blocker unless I make the star placement change?

fichtner:
@Greelan Standard procedure is shipping new code at least once in a devel release so others can potentially review it. The likely target for your changes are .3 and I'll try to propose something for it and we can discuss in a PR?

Greelan:
OK. If all you are after is the star after the heading, I can (begrudgingly!) do that (it was one of the approaches I originally tried, as noted above). The CSS can be simplified in that case too.

fichtner:
It would certainly light up my heart. ❤️

Greelan:
I will refer all complaints to you! 🤣

fichtner:
After the service widget incident nothing will scare me. :)

Greelan:
I'd assumed that it would have put you off getting into style arguments lol

Question is: should the star be the same size as currently (see screenshot), or smaller:

Screenshot 2026-08-13 at 4 45 17 pm

fichtner:
looks great as is

PR #10703, opened 2026-08-11
Firewall: NAT: Destination NAT: expose automatic rules similar to Source NAT

### LINKED ISSUE #10700: Collect automatic RDR rules
**Important notices**

- Model used:
- Extent of AI involvement:

Right now automatic RDR rules are only statically defined here:

https://github.com/opnsense/core/blob/ae388c4b6c16a4bce298fc5a0b605fe95f3cb2c4/src/opnsense/mvc/app/controllers/OPNsense/Firewall/Api/DNatController.php#L118-L122

Facilities like Captive Portal also inject RDR rules, it would be nice if those would appear as well.

Recently there has been this:
https://github.com/opnsense/core/pull/10590

Maybe it can be extended for
 ```# pluginctl -r destination_nat```

Monviech:
**Important notices**

- Model used:
- Extent of AI involvement:

---

**Describe the problem**

Fixes: https://github.com/opnsense/core/issues/10700

### DIFF
--- a/src/etc/inc/plugins.inc.d/captiveportal.inc
+++ b/src/etc/inc/plugins.inc.d/captiveportal.inc
@@ -191,10 +191,12 @@ function captiveportal_firewall($fw)
     $auto = captiveportal_auto_rules();
 
     foreach ($auto['destination'] as $rule) {
+        /* XXX #priority passed, but use inside $rule is not unified */
         $fw->registerDestinationNatRule($rule['#priority'], $rule);
     }
 
     foreach ($auto['filter'] as $rule) {
+        /* XXX #priority passed, but use inside $rule is not unified */
         $fw->registerFilterRule($rule['#priority'], $rule);
     }
 }
--- a/src/etc/inc/plugins.inc.d/captiveportal.inc
+++ b/src/etc/inc/plugins.inc.d/captiveportal.inc
@@ -135,7 +135,7 @@ function captiveportal_firewall($fw)
     global $config;
 
     foreach (captiveportal_auto_destination_nat() as $rule) {
-        $fw->registerForwardRule(2, $rule);
+        $fw->registerDestinationNatRule(2, $rule);
     }
 
     $cp = new \OPNsense\CaptivePortal\CaptivePortal();
--- a/src/etc/inc/plugins.inc.d/captiveportal.inc
+++ b/src/etc/inc/plugins.inc.d/captiveportal.inc
@@ -55,9 +55,88 @@ function captiveportal_syslog()
     return $logfacilities;
 }
 
+function captiveportal_auto_destination_nat()
+{
+    $rules = [];
+
+    $cp = new \OPNsense\CaptivePortal\CaptivePortal();
+    if ($cp->isEnabled()) {
+        foreach ($cp->zones->zone->iterateItems() as $zone) {
+            if ($zone->enabled->isEmpty()) {
+                continue;
+            }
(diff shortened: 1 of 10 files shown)


src/opnsense/mvc/app/controllers/OPNsense/Firewall/Api/DNatController.php
```diff
@@ -67,28 +67,64 @@ protected function setBaseHook($node)
...
-            foreach (['source.network','source.port','destination.network','destination.port', 'target', 'local-port'] as $field) {
-                if (!empty($record[$field])) {
-                    $record["alias_meta_{$field}"] = $this->getNetworks($record[$field]);
-                }
+        foreach ($this->getModel()->rule->iterateItems() as $uuid => $node) {
+            $record = ['uuid' => $uuid];
+
+            /* flatten nested source/destination containers */
+            $reflen = strlen($node->__reference) + 1;
+            foreach ($node->getFlatNodes() as $key => $value) {
```
Monviech: @fichtner 

The %field is missing here, so descriptions (aka translations) do not appear in the grid anymore for configured (model) records.

searchBase() would handle that automatically, but it also paginates.

Since I want to merge automatic and model records later, searchRecordsetBase() is better, but there is nothing to fill the gap (other than using UIModelGrid directly) that can generate the %field.

So we either need something new, or use UIModelGrid directly (which I don't want).

EDIT:

Duplicated a little bit of code to call getDescription here.

src/etc/inc/plugins.inc.d/captiveportal.inc
```diff
@@ -92,53 +171,8 @@ function captiveportal_firewall($fw)
 
                 foreach (['80', '443'] as $to_port) {
                     $rdr_port = $to_port === '443' ? (8000 + (int)$zoneid) : (9000 + (int)$zoneid);
-                    $to = $to_port === '443' ? "(self)" : "<__captiveportal_zone_{$zoneid}>";
```
fichtner, line 95: what I don't like here is breaking out the logic, which makes this harder to maintain in the future:

ideally we'd need a shared emitter and then fill registerForwardRule/registerFilterRule accordingly.

Also, what's the difference of registerDNatRule and registerForwardRule ?

src/etc/inc/plugins.inc.d/captiveportal.inc
```diff
@@ -55,10 +55,89 @@ function captiveportal_syslog()
     return $logfacilities;
 }
 
+function captiveportal_auto_destination_nat()
+{
+    $rules = [];
+
+    $cp = new \OPNsense\CaptivePortal\CaptivePortal();
```
fichtner, line 63: this will be more performant using the ??= trick and a passable model argument like the other code uses

src/opnsense/mvc/app/controllers/OPNsense/Firewall/Api/DNatController.php
fichtner: I think this is fine for now

fichtner:
@Monviech can you retest? one thing I see is that the feeds for anti-lockout are very different from the captive portal feeds. This should be unified so the GUI only gets presented the actual rules format as accepted by the registerXxxRule() functions.

fichtner (approved):

Monviech:
@fichtner I tested it and fixed rule references. Now it looks like its good to go.

Issue #10711, opened 2026-08-13
DNAT firewall registration issue with hyphenated known ports


Top-Cat:
**Important notices**

Before you add a new report, we ask you kindly to acknowledge the following:

- [x] I have read the contributing guidelines at https://github.com/opnsense/core/blob/master/CONTRIBUTING.md
- [x] I am convinced that my issue is new after having checked both open and closed issues at https://github.com/opnsense/core/issues?q=is%3Aissue

**Describe the bug**

DNAT fails to apply when altering destination port and original destination port is a well known port containing a hypen AND rule is set to register a firewall rule.

The well know port name is treated as if it were a port range and the code tries to do arithmetic on it:

```php
$parts = explode('-', $dnatrule['destination']['port']);
if (count($parts) == 2) {
    $range_end = $parts[1] - $parts[0] + $dnatrule['local-port']; // Error: "NS" - "NETBIOS" + "NETBIOS-DGM"
    $tmprule['destination']['port'] .= sprintf('-%s', $range_end);
}
```

```
Fatal error: Uncaught TypeError: Unsupported operand types: string - string in /usr/local/etc/inc/filter.lib.inc:663
Stack trace:
#0 /usr/local/etc/inc/filter.lib.inc(46): filter_core_rules_system(Object(OPNsense\Firewall\Plugin), Array)
#1 /usr/local/etc/inc/filter.inc(187): filter_core_bootstrap(Object(OPNsense\Firewall\Plugin))
#2 /usr/local/etc/rc.filter_configure(40): filter_configure_sync(true, false)
#3 {main}
  thrown in /usr/local/etc/inc/filter.lib.inc on line 663
```

**To Reproduce**

- Create a DNAT rule
- Set `Destination Port` to a well known port with a hypen in it. eg `MICROSOFT-DS` or `RADIUS-ACCT`
- Set `Redirect Target Port` to anything other than `any`
- Set `Firewall rule` to `Register rule`
- Click `Apply`

**Expected behavior**

Firewall rule is applied and redirects traffic while changing the destination port.

**Screenshots**

<img width="905" height="964" alt="Image" src="https://github.com/user-attachments/assets/e59dfa13-8f3d-4968-91f5-e4cebbde7a01" />

<img width="610" height="80" alt="Image" src="https://github.com/user-attachments/assets/a656037d-c1ae-4210-8c8d-63c9df6f83fa" />

PR #10712, opened 2026-08-13
Check all items are numeric before apply range to dnat automatic firewall rule

Top-Cat:
**Important notices**

---

**Describe the problem**

DNAT fails to apply when altering destination port and original destination port is a well known port containing a hypen AND rule is set to register a firewall rule.

---

Checks parts of the port range are all numbers before trying to do arithmetic.

---

**Related issue**

#10711

### DIFF
--- a/src/etc/inc/filter.lib.inc
+++ b/src/etc/inc/filter.lib.inc
@@ -653,8 +653,13 @@ function filter_core_rules_system($fw, $defaults)
             if (!empty($dnatrule['destination']['port']) && !empty($dnatrule['local-port'])) {
                 $tmprule['destination']['port'] = $dnatrule['local-port'];
                 $parts = explode('-', $dnatrule['destination']['port']);
-                if (count($parts) == 2 && is_numeric($parts[0]) && is_numeric($parts[1]) && is_numeric($dnatrule['local-port'])) {
-                    $range_end = $parts[1] - $parts[0] + $dnatrule['local-port'];
+                if (count($parts) == 2 && is_numeric($parts[0]) && is_numeric($parts[1])) {
+                    $known = \OPNsense\Base\FieldTypes\PortField::getWellKnown($dnatrule['local-port']);
+                    if (!empty($known)) {
+                        $tmprule['destination']['port'] = array_shift($known);
+                    }
+
+                    $range_end = $parts[1] - $parts[0] + $tmprule['destination']['port'];
                     $tmprule['destination']['port'] .= sprintf('-%s', $range_end);
                 }
             } elseif (!empty($dnatrule['local-port'])) {
--- a/src/etc/inc/filter.lib.inc
+++ b/src/etc/inc/filter.lib.inc
@@ -653,7 +653,7 @@ function filter_core_rules_system($fw, $defaults)
             if (!empty($dnatrule['destination']['port']) && !empty($dnatrule['local-port'])) {
                 $tmprule['destination']['port'] = $dnatrule['local-port'];
                 $parts = explode('-', $dnatrule['destination']['port']);
-                if (count($parts) == 2) {
+                if (count($parts) == 2 && is_numeric($parts[0]) && is_numeric($parts[1]) && is_numeric($dnatrule['local-port'])) {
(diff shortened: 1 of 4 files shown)


src/etc/inc/filter.lib.inc
```diff
@@ -653,7 +653,7 @@ function filter_core_rules_system($fw, $defaults)
             if (!empty($dnatrule['destination']['port']) && !empty($dnatrule['local-port'])) {
                 $tmprule['destination']['port'] = $dnatrule['local-port'];
                 $parts = explode('-', $dnatrule['destination']['port']);
-                if (count($parts) == 2) {
+                if (count($parts) == 2 && is_numeric($parts[0]) && is_numeric($parts[1]) && is_numeric($dnatrule['local-port'])) {
```
fichtner: it looks like this may need stricter validation. I'm not sure we want to do range calculations with strings so omitting this block is also wrong if actual input was given.

Top-Cat: I'm not sure I understand. In the example I gave in the issue #10711 omitting this block results in the correct firewall rule being created.

The only case where this fix doesn't work is when a range is mapped to a well known port (doesn't matter in that case if the port name has a hyphen or not). Happy to add a lookup to the actual port number to accommodate this case if you want.

Top-Cat:
I think this now covers all cases of destination and target ports / ranges / well known names.

> omitting this block is also wrong if actual input was given

Happy to investigate more if you can give me a setup that you think won't work 😄

AdSchellevis:
How about if we add a new method in `PortField` to return a port number (or range) depending on user input, e.g. `http` would return `80`, `80` would do the same and a range stays a range.

It would keep the implementation a bit cleaner as we're interested in the number for both cases.

Not sure about the name, but something like `normalizedPort()` might work. 

The first case could access the port direct ($rule)

https://github.com/opnsense/core/blob/ac14c29ff9f78687766e2feff4217804106bf56c/src/etc/inc/filter.lib.inc#L640

For the second, we need to map the source of the data, which would be the following if I'm not mistaken:

https://github.com/opnsense/core/blob/a698c1d94a75d0f3389257bcd21278801bf4d0e7/src/etc/inc/filter.inc#L242

It's not really problematic to replace unconditionally in my opinion, pf should understand the numbers anyway.

Top-Cat:
@AdSchellevis Is this what you were going for?

src/opnsense/mvc/app/models/OPNsense/Base/FieldTypes/PortField.php
```diff
@@ -262,4 +262,17 @@ public function getValidators()
...
+     * @return string
+     */
+    function normalizedPort() {
+        $known = self::getWellKnown((string)$this);
+
+        if (!empty($known)) {
+            return array_shift($known);
+        }
+
+        return (string)$this;
```
fichtner: for future robustness this should be:


```suggestion
        return $this->getValue();
```

src/opnsense/mvc/app/models/OPNsense/Base/FieldTypes/PortField.php
```diff
@@ -262,4 +262,17 @@ public function getValidators()
         }
         return parent::getValidators();
     }
+
+    /**
+     * @return string
+     */
+    function normalizedPort() {
+        $known = self::getWellKnown((string)$this);
```
fichtner: same here, __toString() may have other purposes

AdSchellevis:
> @AdSchellevis Is this what you were going for?

yep, very nice! thanks!

AdSchellevis (approved):

src/opnsense/mvc/app/models/OPNsense/Base/FieldTypes/PortField.php
```diff
@@ -262,4 +262,17 @@ public function getValidators()
         }
         return parent::getValidators();
     }
+
+    /**
+     * @return string
+     */
+    function normalizedPort() {
```
AdSchellevis:
```suggestion
    public function normalizedPort() {
```

PR #10718, opened 2026-08-15
menu: hide Favorites star unless title area hovered/tapped

Greelan:
**Important notices**

- Model used:
- Extent of AI involvement:

---

**Describe the problem**

Some concern has been expressed that moving the Favorites star on each leaf page to just after the title, rather than at the right-hand side, means it is too prominent and potentially annoying for some users.

---

Hide the star unless the title/star area is hovered over (or tapped on touch devices).

Desktop:

https://github.com/user-attachments/assets/f48a9e6a-e892-4cea-bf8a-c6e1f361ee90

Mobile:

https://github.com/user-attachments/assets/05772c53-fa30-4783-8120-0a669cc7644e

---

**Related issue**

N/A

### DIFF
--- a/src/opnsense/mvc/app/views/layouts/default.volt
+++ b/src/opnsense/mvc/app/views/layouts/default.volt
@@ -284,7 +284,7 @@
         <header class="page-content-head">
           <div class="container-fluid">
             <ul class="list-inline">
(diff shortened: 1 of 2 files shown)


Monviech:
Tested this on my iPad and Macbook, works correctly on both. The css is also a good choice here, it's very simple.

I think it's good, it might make initial discovery a bit harder, but it will annoy less users who do not want it. I think it's the best tradeoff.

If we still decide we want better discoverability, we can always revert this.

For now I will merge it. Thank you for your work :)

Greelan:
Thanks for the feedback :)

Hopefully that's also the consensus at today's dev meeting!

Monviech:
I think it's fine, nobody complained. So thank you, should be ready now.

fichtner:
Yep, I like it. Great work overall @Greelan

PR #10737, opened 2026-08-19
Clean QR codes

Roy-Orbison:
**Important notices**

---

**Describe the problem**

The version of the QR code generation library by Kazuhiko Arase is outdated, as is the jQuery plugin that uses it. The latter generates `<canvas>` images with inconsistently-sized & rectangular 'pixels'. An example (view image at full size):  
qrcode-janky

---

This patch updates the library to the current version, and replaces the existing jQuery plugin with a much simpler one, which uses the former's built-in image generation. It produces images as big (or as small) as they need to be, like so:  
qrcode-clear

---

**Related issue**

n.a.


AdSchellevis:
@Roy-Orbison thanks!

Issue #10751, opened 2026-08-22
Wireless - migrate to MVC


AdSchellevis:
**Important notices**

Before you add a new report, we ask you kindly to acknowledge the following:

- [x] I have read the contributing guidelines at https://github.com/opnsense/core/blob/master/CONTRIBUTING.md
- [x] I am convinced that my issue is new after having checked both open and closed issues at https://github.com/opnsense/core/issues?q=is%3Aissue
- [ ] AI tools were used to create at least part of the text submitted herewith.

If AI was used, please disclose:

- Model used:
- Extent of AI involvement:

**Is your feature request related to a problem? Please describe.**

Wireless is a bit of a mess, although support on the operating system level isn't great, there are still people using it.
In practice this means we can't easily deprecate `interfaces.php` because it hosts the settings for wireless at the moment. Removing wireless from core in full will likely be problematic because of the number of people using it, this basically leaves migration the only viable option.

**Describe the solution you like**

Migrate the current wireless page (`Interfaces: Wireless: Devices`) to MVC and move the settings currently in the interfaces/wireless configuration section to wireless/clone.

**Describe alternatives you considered**

Remove wireless support, which would be my personal preference (due to its limited OS support), but will leave current wireless users without much of a choice. 

**Additional context**

This is required to help making https://github.com/opnsense/core/issues/10568 the replacement for the current interface configuration.

PR #10782, opened 2026-08-23
Interfaces: Wireless: Devices - Migrate to MVC code

AdSchellevis:
**Important notices**

- Model used:
- Extent of AI involvement:

---

**Describe the problem**

Wireless is a bit of a mess, although support on the operating system level isn't great, there are still people using it.
In practice this means we can't easily deprecate interfaces.php because it hosts the settings for wireless at the moment. Removing wireless from core in full will likely be problematic because of the number of people using it, this basically leaves migration the only viable option.

---

first phase for https://github.com/opnsense/core/issues/10751

This commit just replaces the existing (rather minimal) wireless clone configuration section with an mvc implementation, so we have a landing spot for the actual settings later. Impact of this change should be relatively low and easy to merge to production as it doesn't change configuration settings.

To not impact the weirdness with "Persist common settings", we currently only overlay the clone property in the wireless container.

---

**Related issue**

https://github.com/opnsense/core/issues/10751


fichtner (approved):
looks like a good way forward, thanks!

PR #10790, opened 2026-08-24
dnsmasq: leases sorting fixes

### LINKED ISSUE #10821: UI bug: Ordering by Lease type in the dnsmasq UI does not work.
**Important notices**

**Describe the bug**

The table at `/ui/dnsmasq/leases` offers "click on the column title" sorting options, and it generally seems to work. However, sorting by Lease Type does *not* work, no matter which sort direction is picked, `static` and `dynamic` leases are intermixed. 

**To Reproduce**

1. Set up dnsmasq for DHCP
2. Have both static and dynamic leases in the list
3. Attempt to sort by "Lease Type"

**Screenshots**

I hid a few columns just to make it less noisy, but it reproduces with the default column set (I did reset local storage to test):

Image

**Environment**

OPNsense 26.7.3_8-amd64
Deciso DEC750 (AMD Ryzen Embedded V1500B)

Greelan:
**Important notices**

- Model used: Claude Opus 4.8
- Extent of AI involvement: Identified the source of the issues and suggested solutions.

---

**Describe the problem**

Ordering of various Dnsmasq lease fields (IP Address, MAC Address, DUID, Lease Type) currently does not work properly since the existing code does not take proper account of the underlying data structure (array) or type (hex numbers).

---

Ensure proper ordering.

(Note that the same issues likely affect the KEA leases sorting, however I am focusing this PR on Dnsmasq for the time being.)

---

**Related issue**

https://forum.opnsense.org/index.php?topic=52794.0

### DIFF
--- a/src/opnsense/mvc/app/controllers/OPNsense/Base/ApiControllerBase.php
+++ b/src/opnsense/mvc/app/controllers/OPNsense/Base/ApiControllerBase.php
@@ -88,7 +88,8 @@ class ApiControllerBase extends ControllerRoot
         $defaultSort = null,
         $filter_funct = null,
         $sort_flags = SORT_NATURAL | SORT_FLAG_CASE,
-        $search_clauses = null
+        $search_clauses = null,
+        $sort_transforms = []
     ) {
         $records = is_array($records) ? $records : []; // safeguard input, we are only able to search arrays.
         $itemsPerPage = intval($this->request->getPost('rowCount', 'int', 9999));
@@ -121,6 +122,10 @@ class ApiControllerBase extends ControllerRoot
                 }
             }
             $keys = array_column($records, $sortKey);
+            if (isset($sort_transforms[$sortKey])) {
+                $keys = array_map($sort_transforms[$sortKey], $keys);
+                $sort_flags = SORT_STRING; // a transform yields a directly comparable key
+            }
             array_multisort($keys, $sortOrder, $sort_flags, $records);
         }
 
--- a/src/opnsense/mvc/app/controllers/OPNsense/Base/ApiControllerBase.php
+++ b/src/opnsense/mvc/app/controllers/OPNsense/Base/ApiControllerBase.php
@@ -88,8 +88,7 @@ class ApiControllerBase extends ControllerRoot
         $defaultSort = null,
         $filter_funct = null,
         $sort_flags = SORT_NATURAL | SORT_FLAG_CASE,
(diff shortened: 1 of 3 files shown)


AdSchellevis:
it's probably better to change this in the template / form, eg use a sortable field (`sort_address`) in:

https://github.com/opnsense/core/blob/be882cccd6402a777abf1a90f46700811243b878/src/opnsense/mvc/app/views/OPNsense/Dnsmasq/leases.volt#L170

and add a formatter to show the `address` field instead of the `sort_address` one.

Greelan:
Thanks for the feedback. I take it the preference is to keep this out of `ApiControllerBase.php` and confine all the changes to the module?

Noting that there will need to be four sortable fields because this doesn't only impact `address`.

AdSchellevis:
given the values we do want to sort are already in the dataset received by `searchRecordsetBase()`, I would prefer using actual fieldnames indeed (instead of adding a mapper in between).

Greelan:
Thinking about it further, is there a case for a shared helper or helpers, such as in `OPNsense\Firewall\Util`, that sort IPs and hex codes? Then it wouldn't need to be replicated in each spot where it is needed (currently at least two, Dnsmasq and Kea).

AdSchellevis:
I had the idea some time ago to add types on data in the search function (https://github.com/opnsense/core/commit/dad956e3143e7aa26f6be140e4c0ff48d1758e22), but haven't found time to work on it further I'm afraid.

Ideally you should be able to user filter logic based on the type of data that's inside the container, same would logically apply for sorting in that case too. 

The main reason for wanting this is the filter page, which would be very practical to have a simple "impact assessment" for clauses provided (e.g. searching which rules impact `192.168.1.1` should logically also return `192.168.1.0/24`)
Currently this only works partially (locate aliases which match the search criteria)

This is more of a longer term solution I'm afraid.

Greelan:
@AdSchellevis not sure whether you noticed but I have refactored the PR as requested. Please let me know if you have any other feedback.

AdSchellevis:
@Greelan missed it, sorry about that, this should work, thanks!

Greelan:
Thanks. I'm happy to do the same for Kea if that would help?

AdSchellevis:
I think that would be practical, cc @Monviech

Monviech:
@Greelan If you have time, I would appreciate it. The same issues (e.g. the array in is_reserved) also applies to KEA. Thank you :)

PR #10795, opened 2026-08-25
mvc: advanced marker for fields POC

fichtner:
Also adds tooltip for it and help button.

**Important notices**

- Model used:
- Extent of AI involvement:

---

**Describe the problem**

Advanced fields are hard to spot in larger forms.

---

Show a marker next to them. Might even help people treating them with a bit more respect. ;)

Left:

Screenshot 2026-08-25 at 15 00 17

Right:

Screenshot 2026-08-25 at 14 59 53

Tooltip:

Screenshot 2026-08-25 at 15 00 27

I kinda like the right one, but left works also. Tooltips might be a bit annoying for general help but I included it for discussion.

### DIFF
--- a/src/opnsense/mvc/app/views/layout_partials/base_dialog.volt
+++ b/src/opnsense/mvc/app/views/layout_partials/base_dialog.volt
@@ -83,7 +83,7 @@
                         <tr>
                             <td colspan="3">
                                 <span>
-                                    {% if fields['advanced']|default(false) %}<a href="#"><i class="fa fa-toggle-off fa-fw text-danger" id="show_advanced_{{base_dialog_id}}"></i></a> <small>{{ lang._('advanced mode') }}</small> <i class="fa fa-cog fa-sm fa-fw text-info"></i>{% endif %}</td>
+                                    {% if fields['advanced']|default(false) %}<a href="#"><i class="fa fa-toggle-off fa-fw text-danger" id="show_advanced_{{base_dialog_id}}"></i></a> <small>{{ lang._('advanced mode') }}</small> <i class="fa fa-cog fa-sm fa-fw text-info"></i>{% endif %}
                                 </span>
                                 <span class="pull-right">
                                     {% if fields['help']|default(false) %}<i class="fa fa-info-circle fa-sm fa-fw text-primary"></i> <small>{{ lang._('full help') }}</small> <a href="#"><i class="fa fa-toggle-off fa-fw text-danger" id="show_all_help_{{base_dialog_id}}"></i></a>{% endif %}
--- a/src/opnsense/mvc/app/views/layout_partials/base_dialog.volt
+++ b/src/opnsense/mvc/app/views/layout_partials/base_dialog.volt
@@ -81,9 +81,13 @@
                         <tbody {%if not (section['static']|default(false)) %}class="collapsible" {% if section['collapse']|default(false) %}style="display: none;"{%endif%}{%endif%}>
                         {%  if not section['type'] and (fields['advanced']|default(false) or fields['help']|default(false)) %}
                         <tr>
-                            <td>{% if fields['advanced']|default(false) %}<a href="#"><i class="fa fa-toggle-off text-danger" id="show_advanced_{{base_dialog_id}}"></i></a> <small>{{ lang._('advanced mode') }}</small>{% endif %}</td>
-                            <td colspan="2" style="text-align:right;">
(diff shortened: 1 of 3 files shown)


Monviech:
The idea is great, I would prefer the same green toggle next to the names that is used to unhide the advanced mode.

EDIT: Well its also used for help, but the exclamation mark looks a little evil.

Might be a good idea to add the same icon to the actual advanced mode toggle so that it makes "sense"

fichtner:
Not attached to anything. Just want the visibility :D

Monviech:
I think a trailing fa-asterisk as you suggested might be the least invasive one. That also keeps all "i" fa icons at the same level (I need everything aligned, please, or I die, not literally but mentally)

Monviech:
Wait I just remembered, in the firewall rules I used fa-cog for advanced mode enabled as icon

https://github.com/opnsense/core/blob/dc9ac3196d5ba3cd17f00a3bd0b0f656edd7025e/src/opnsense/mvc/app/views/OPNsense/Firewall/filter_rule.volt#L508-L516

swhite2:
> The idea is great, I would prefer the same green toggle next to the names that is used to unhide the advanced mode.
> 
> EDIT: Well its also used for help, but the exclamation mark looks a little evil.
> 
> Might be a good idea to add the same icon to the actual advanced mode toggle so that it makes "sense"

Agreed, for me it suggests a misconfiguration or incompatible setting. The cog is a nice idea I think

fichtner:
Screenshot 2026-08-25 at 15 40 52

fichtner:
slight alternative to make it less heavy is `fa-cogs`

Screenshot 2026-08-25 at 15 42 30

Monviech:
I would use fa-cog and also add it here:

image

And maybe adjust full help to get the fa-info too?

fichtner:
I can try but in that case we need to flip the order for the full help display.... or maybe not, will see

fichtner:
Screenshot 2026-08-25 at 19 20 45

I have the strange feeling we should switch the help and the advanced toggle position in the header ;)

fichtner:
Screenshot 2026-08-25 at 19 23 09

single cog

SeimusS:
I love this idea!

But I would avoid using triangles as it makes the impression its a mandatory field to be set.
The cog is a better option cause it goes hand with hand with the FW rule advanced settings indicator :)

Regards,
S.

Monviech:
Single cog is great.

swhite2:
Nice!

swhite2 (approved):
Ok for me, can always be refined further on master if needed

Monviech (approved):
I don't mind if its one or the other but please change it accordingly in the firewall rules page as well.

fichtner:
Alright... so single or double? Leaning towards double myself but I don't mind too much and @Monviech already prefers the other one ;)

fichtner:
~~Oh and can we ditch the tooltips now that we have symbols in the toggles?~~ @AdSchellevis says omit then

fichtner:
Last but not least... blue vs. yellow.. fight!?

Screenshot 2026-08-26 at 11 08 21

Screenshot 2026-08-26 at 11 08 10

swhite2:
Team blue :)

Monviech:
Firewall does it like this:

image

image

Whatever you do please keep it consistent, that's all I want :)

fichtner:
There's no enable/disable concept here since we also have the toggle and muted help indicates it's not advanced mode... I'd like to avoid showing a muted advanced mode icon for every other field?

But in any case yellow seems to win here.

Monviech:
It would be nice if advanced mode could remember that a form has advanced options set and activate it or hint it via the cog symbols color.

fichtner:
Advanced functions may have defaults, which makes it unclear to the user if the setting came from them or is intended.

sopex:
Blue is more compatible with the interface. opnsense orange and yellow don't mix well :(

fichtner:
Well, nothing seems perfect here. These are just info/warning colors as good as the theme makes them. In most places I don’t like the info (blue) in opnsense light theme, but changing it now will trigger a chain of more requests for changes.


Cheers,
Franco

sopex:
True, even under the best conditions... The perfect UI doesn't exist. 

Just wanted to give my two cents since it seemed like a tossup :)

fichtner:
Thanks, it's appreciated. ❤️

PR #10799, opened 2026-08-26
Firewall: Rules: emit gateway debug message in block rules only when gateway was not empty

Monviech:
**Important notices**

- Model used:
- Extent of AI involvement:

---

**Describe the problem**

I wondered why my rules.debug was full of these messages for each of my block rules even though they didn't have a gateway or reply-to set.

---

Only emit these when something was removed from a block rule.

---

**Related issue**

None

### DIFF
--- a/src/opnsense/mvc/app/library/OPNsense/Firewall/FilterRule.php
+++ b/src/opnsense/mvc/app/library/OPNsense/Firewall/FilterRule.php
@@ -212,8 +212,10 @@ class FilterRule extends Rule
                 $this->log("Gateway protocol mismatch");
             }
             if (!empty($rule['type']) && $rule['type'] != 'pass') {
+                if (!empty($rule['gateway']) || !empty($rule['reply'])) {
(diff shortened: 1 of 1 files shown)


AdSchellevis (approved):

src/opnsense/mvc/app/library/OPNsense/Firewall/FilterRule.php
```diff
@@ -212,8 +212,10 @@ private function parseFilterRules()
                 $this->log("Gateway protocol mismatch");
             }
             if (!empty($rule['type']) && $rule['type'] != 'pass') {
+                if (!empty($rule['gateway']) || !empty($rule['reply'])) {
+                    $this->log("Gateway not allowed for block rules");
+                }
                 unset($rule['gateway'], $rule['reply']);
```
fichtner, line 218: theoretically this could move into the if condition for proper flow reasons, but I'm not 100% sure if there is a empty vs. isset issue at play

Issue #9946, opened 2026-03-13
Firewall: Scrub rules scope/removal/match rule-check scope


fichtner:
**Important notices**

Our forum is located at https://forum.opnsense.org , please consider joining discussions there in stead of using GitHub for these matters.

Before you ask a new question, we ask you kindly to acknowledge the following:

- [x] I have read the contributing guide lines at https://github.com/opnsense/core/blob/master/CONTRIBUTING.md
- [x] I am convinced that my issue is new after having checked both open and closed issues at https://github.com/opnsense/core/issues?q=is%3Aissue

As one of the few remaining legacy pages the question is how to proceed with this page:

- Migrate away from it by using match rules?
- Rewrite the page in MVC?

For this reason check out what pf.conf says about the features also WRT FreeBSD 15.

The consensus is add scrub support to firewall rules using the `match` keyword and perhaps migrate the global settings to a model in order to offer the same features in the new firewall rules.

kevinchalet:
> Migrate away from it by using match rules?

Independently of the outcome of this thread, being able to set the IP DSCP/ToS directly in regular firewall rules without using separate scrub rules would be really awesome! Because currently, if you want to enforce a specific VLAN PCP + a specific ToS, you need both a regular rule and a scrub rule, which can quickly become a PITA when you do that for a lot of rules.

Happy to open a separate ticket if you're interested in making that possible 😃

https://man.openbsd.org/pf.conf#set~4

Monviech:
I've used ChatGPT Codex to start something here, please note the entire code is AI generated and untested.
See it as impact analysis, but since it looks rather promising we might be able to salvage the concepts from it:

https://github.com/opnsense/core/tree/mvc_scrub

The main issue is that the old scrub rule concept (FreeBSD-14 and earlier) do not translate completely to the new scrub rule concept (FreeBSD-15 and later).

That means we will probably use the Migration Assistant, and only activate the new scrub rules that use "match" keywords when none of the old ones remain.

mimugmail:
I had countless tickets with slow transfers in vpn tunnels or cg nat where scrubbing mss was the solution. If there wont be an automatic migration I predict countless issues here and in forums

Monviech:
@mimugmail There will be an automatic migration for these rules. Thanks for the feedback. Not sure if it can cover /all/ cases, but the most usual ones will be migrated.

Here's an example: https://github.com/opnsense/core/pull/10826#issuecomment-5494086782

[...]

PR #10812, opened 2026-08-28
Firewall: Rules: Add match keyword and normalization options

### LINKED ISSUE #10801: Consider supporting `set-tos` in firewall rules (relates to https://github.com/opnsense/core/issues/9946)
**Important notices**

Since you are guys are in the process of adding very cool features to the new firewall rules (❤️ `received-on` ❤️), here's one I'd love to see supported: being able to set the IP DSCP/ToS - aka `set-tos` - directly in regular firewall rules without using separate scrub rules. The `tos` match modifier is already supported so it would be nice if `set-tos` was also supported by OPNsense.

Setting a specific ToS is already supported via scrub rules, but if you want to enforce a specific VLAN PCP + a specific ToS, you need both a regular rule and a scrub rule, which can quickly become a real PITA when you have to do that for a lot of rules... 🤣

**Additional context**

https://man.freebsd.org/cgi/man.cgi?query=pf.conf

Forked from https://github.com/opnsense/core/issues/9946.

Monviech:
**Important notices**

- Model used: ChatGPT 5.6 Sol
- Extent of AI involvement: Assisting as coding agent

---

**Describe the problem**

FreeBSD 15 supports scrubbing as an option on filter rules. The MVC firewall rules do not currently support the ```match``` action or ```scrub (...)``` options.

---

No automatic or manual migration is included here yet, this stage is only to add capability.

* Add ```match``` as a regular firewall rule action using the existing rule priority.
* Allow ```scrub (...)``` options on both ```pass``` and ```match``` rules.
* Generate global ```set reassembly```, random-ID and interface MSS rules using the modern PF syntax.
* Move the existing global scrub controls to Firewall Advanced without changing their configuration paths.
* Validate options that PF does not support on match rules (e.g. routing and state).

This PR does not migrate or remove legacy scrub rules, nor add compatibility handling for mixed rulesets, nor give user-created match rules special priority.
-> If the user removes all legacy scrub rules, the new ```set reassembly``` will appear.

---

**Related issue**

For: https://github.com/opnsense/core/issues/9946
For: https://github.com/opnsense/core/issues/10801

### DIFF
--- a/src/opnsense/mvc/app/library/OPNsense/Firewall/FilterRule.php
+++ b/src/opnsense/mvc/app/library/OPNsense/Firewall/FilterRule.php
@@ -54,6 +54,7 @@ class FilterRule extends Rule
         'os' => 'parsePlain, os {","}',
         'to' => 'parsePlainCurly,to ',
         'to_port' => 'parsePlainCurly, port ',
+        'scrub' => 'parseScrub',
         'received-on' => 'parseReceivedOn',
         'icmp-type' => 'parseReplaceSimple,skip:"skip",icmp-type {,}',
         'icmp6-type' => 'parsePlain,icmp6-type {,}',
@@ -101,6 +102,17 @@ class FilterRule extends Rule
         } else {
             return "";
         }
+    }
+
+    /**
+     * Render FreeBSD 15 style scrub options on a filter rule.
+     *
+     * @param string $value space-separated scrub options
+     * @return string
+     */
+    protected function parseScrub($value)
+    {
+        return empty($value) ? '' : "scrub ( {$value} ) ";
     }
 
     /**
@@ -211,7 +223,7 @@ class FilterRule extends Rule
                 $rule['disabled'] = true;
                 $this->log("Gateway protocol mismatch");
             }
-            if (!empty($rule['type']) && $rule['type'] != 'pass') {
+            if (in_array($rule['type'] ?? '', ['block', 'reject'], true)) {
                 if (!empty($rule['gateway']) || !empty($rule['reply'])) {
                     $this->log("Gateway not allowed for block rules");
                 }
--- a/src/opnsense/mvc/app/library/OPNsense/Firewall/FilterRule.php
+++ b/src/opnsense/mvc/app/library/OPNsense/Firewall/FilterRule.php
@@ -155,6 +155,9 @@ class FilterRule extends Rule
      */
     protected function convertReplyTo(&$rule)
     {
+        if (($rule['type'] ?? '') == 'match') {
+            return;
+        }
         if (!empty($rule['reply-to'])) {
             // reply-to gateway set, when found map to reply attribute, otherwise skip keyword
             if (!empty($this->gatewayMapping[$rule['reply-to']])) {
@@ -223,9 +226,10 @@ class FilterRule extends Rule
                 $rule['disabled'] = true;
                 $this->log("Gateway protocol mismatch");
             }
-            if (in_array($rule['type'] ?? '', ['block', 'reject'], true)) {
+            /* PF does not support routing on match or block rules. */
+            if (!empty($rule['type']) && $rule['type'] != 'pass') {
                 if (!empty($rule['gateway']) || !empty($rule['reply'])) {
-                    $this->log("Gateway not allowed for block rules");
+                    $this->log("Gateway not allowed for {$rule['type']} rules");
                 }
                 unset($rule['gateway'], $rule['reply']);
             }
--- a/src/opnsense/mvc/app/library/OPNsense/Firewall/FilterRule.php
+++ b/src/opnsense/mvc/app/library/OPNsense/Firewall/FilterRule.php
@@ -155,9 +155,6 @@ class FilterRule extends Rule
      */
     protected function convertReplyTo(&$rule)
     {
-        if (($rule['type'] ?? '') == 'match') {
-            return;
-        }
         if (!empty($rule['reply-to'])) {
             // reply-to gateway set, when found map to reply attribute, otherwise skip keyword
             if (!empty($this->gatewayMapping[$rule['reply-to']])) {
@@ -226,10 +223,9 @@ class FilterRule extends Rule
                 $rule['disabled'] = true;
                 $this->log("Gateway protocol mismatch");
             }
-            /* PF does not support routing on match or block rules. */
             if (!empty($rule['type']) && $rule['type'] != 'pass') {
                 if (!empty($rule['gateway']) || !empty($rule['reply'])) {
-                    $this->log("Gateway not allowed for {$rule['type']} rules");
+                    $this->log("Gateway not allowed for non-pass rules");
                 }
                 unset($rule['gateway'], $rule['reply']);
             }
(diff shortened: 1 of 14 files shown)


src/opnsense/mvc/app/library/OPNsense/Firewall/FilterRule.php
```diff
@@ -211,9 +226,10 @@ private function parseFilterRules()
                 $rule['disabled'] = true;
                 $this->log("Gateway protocol mismatch");
             }
+            /* PF does not support routing on match or block rules. */
             if (!empty($rule['type']) && $rule['type'] != 'pass') {
                 if (!empty($rule['gateway']) || !empty($rule['reply'])) {
-                    $this->log("Gateway not allowed for block rules");
+                    $this->log("Gateway not allowed for {$rule['type']} rules");
```
fichtner:
```suggestion
                    $this->log("Gateway not allowed for non-pass rules");
```

you can avoid the comment here -- and and I'm not sure about the early return. this code tries to make sure to adhere to validation and print an error but the new return simply ignores all (w maithout a log message). maybe rewriting the whole block is better?

src/opnsense/mvc/app/models/OPNsense/Firewall/FieldTypes/FilterRuleField.php
```diff
@@ -103,6 +105,26 @@ public function serialize()
...
+                'scrub_random_id' => 'random-id',
+                'scrub_max_mss' => 'max-mss',
+                'scrub_min_ttl' => 'min-ttl',
+                /* XXX: pf maps this legacy scrub option to "set tos", so pfctl may display both forms. */
+                'scrub_set_tos' => 'set-tos',
+            ] as $fieldname => $keyword
+        ) {
+            if (!$this->$fieldname->isEmpty()) {
+                $scrub_options[] = is_a($this->$fieldname, BooleanField::class) ?
+                    $keyword : $keyword . ' ' . (string)$this->$fieldname;
```
fichtner:
```suggestion
                    $keyword : $keyword . ' ' . $this->$fieldname->getValue();
```

src/opnsense/mvc/app/models/OPNsense/Firewall/Filter.php
```diff
@@ -188,6 +206,30 @@ public function performValidation($validateFullModel = false)
                         }
                     } else {
                         // Additional filter validations
+                        if ($rule->action->isEqual('match')) {
+                            if (!$rule->statetype->isEqual('none')) {
```
fichtner: Although I like the validations UX may suffer. We should consider JS helpers for preselecting the correct values on a change of the action.

src/www/system_advanced_firewall.php
```diff
@@ -36,6 +36,9 @@
 if ($_SERVER['REQUEST_METHOD'] === 'GET') {
     $pconfig = array();
     $pconfig['disablefilter'] = !empty($config['system']['disablefilter']);
+    $pconfig['scrubnodf'] = !empty($config['system']['scrubnodf']);
+    $pconfig['scrubrnid'] = !empty($config['system']['scrubrnid']);
+    $pconfig['scrub_interface_disable'] = !empty($config['system']['scrub_interface_disable']);
```
fichtner: I think this will cause nightmares for @AdSchellevis -- how about we start grouping this in a model right away?

src/opnsense/mvc/app/views/OPNsense/Firewall/filter_rule.volt
```diff
@@ -464,7 +464,9 @@
                         }
 
                         // Action
-                        if (row.action === "block") {
+                        if (row.action === "match") {
```
fichtner, line 467: note that when adding match you also need to adjust the firewall live log to distinguish matches

src/www/system_advanced_firewall.php
AdSchellevis: sounds like a plan to me, given the rules page is already pretty packed, best add a normalization settings which hosts these

Monviech:
As discussed with @fichtner I'll reduce scope here a bit more, leaving legacy options where they have been without moving anything.

src/www/system_advanced_firewall.php
fichtner: in a firewall model it would all be settings, but grouped by sections/headers I guess. like a bit of a preparation for the final move of that page?

src/opnsense/mvc/app/models/OPNsense/Firewall/Filter.php
Monviech: Since the same could be said about most of the combinations possible here, without dependency metadata I would have to replicate all of this in javascript.

I think the validation triggering also "teaches" the user about the why better than silently changing some fields.

Also what happens if somebody creates a match rule, and clones it later into a pass rule, the state-type would stay on none. Same when just selecting different values in the selectpicker, I don't know if the user wants a keep-state or no-state pass rule when changing the match back to pass etc...

src/www/system_advanced_firewall.php
Monviech: Should I reflect these in the model like I did for the SNAT mode? I already made a general firewall container here:

https://github.com/opnsense/core/blob/8834c626c163dbf1134be611364400fb4e6eac5f/src/opnsense/mvc/app/models/OPNsense/Firewall/Filter.xml#L7-L18

fichtner: I think so but unsure about scoping CC @AdSchellevis

src/opnsense/mvc/app/models/OPNsense/Firewall/Filter.php
AdSchellevis: It's a bit inconsistent with how block rules currently work, which are also stateless. 

https://github.com/opnsense/core/blob/b6a700290925eb91cadc4156d17066368a43a983/src/opnsense/mvc/app/library/OPNsense/Firewall/FilterRule.php#L247

Ideally I would rather have these fields connected to a type, but to limit the diff, it might be better to skip the validations as these are ignored anyway now.

AdSchellevis (changes requested):
@Monviech one small remark, which might simplify this further, but other than that this looks good. probably better to add the options here first and care for the migration in another PR.

src/opnsense/mvc/app/models/OPNsense/Firewall/Filter.php
fichtner: I hinted at a preselect of options through a listener on the action (similar to what Aliases or VIPs are doing). It could help, but I also see @Monviech's argument that this will just have to be hardcoded in JS. Since the overall changes needed or intended are unclear it's difficult to say how much code that is and if it needs constant maintenance (my gut feeling says no, but I also don't have a solution ready).

Monviech: I removed the validations for now, the backend normalizes.

I added a small JS handler that automatically empties/changes invalid combinations when the rule type is changed.

This should at least serialize the correct combinations from the front end (most of the time if the user doesn't change it back). The API can still serialize "invalid" combinations though that will then be normalized during ruleset generation.

When the legacy stuff is gone at some point this can probably be moved to validations.

PR #10813, opened 2026-08-28
rc: add watchdog to shutdown, reboot and reload_all cases

### LINKED ISSUE #10567: rc(8): improve shutdown behavior
**Important notices**

- [-] AI tools were used to create at least part of the text submitted herewith.

- Model used:
- Extent of AI involvement:

In order to shutdown or reboot the system, all running applications are being shutdown before the fact, unfortunately this doesn't seem to be happening to all applications as some ignore the signal, in which case the system hangs during the event.

Although not entirely sure, but logically there should be some timer which forces the application to exit even when it doesn't appear to be listening to our gentle requests.

None

**Additional context**

This should be rather easy to emulate by having a daemon looping at 100%, but that's likely the first thing we should do here, generate a reproducible case.

swhite2:
**Important notices**

- Model used: GPT 5.6-Sol
- Extent of AI involvement:

Generating initial code changes based, adjusted as necessary.

---

**Describe the problem**

See https://github.com/opnsense/core/issues/10567

---

Add a simple watchdog that checks service pid validity over the waiting period and kills accordingly. A simple

```
sleep 15
kill -KILL "${service_pid}"
```

is also possible and would simplify it greatly, but obviously has the race of possible reclaimed pid.

Then there's also the default timeout value, which is set to 15 here.

---

**Related issue**

https://github.com/opnsense/core/issues/10567

### DIFF
--- a/src/etc/rc.freebsd
+++ b/src/etc/rc.freebsd
@@ -79,10 +79,34 @@ rc_enabled()
 	esac
 }
 
+get_service_pid()
+{
+	rc_filename="$1"
+	service_name="$2"
+
+	status="$(${rc_filename} status 2>/dev/null)" || return 1
+
+	pid="$(printf '%s\n' "${status}" |
+	    sed -n "s/^${service_name} is running as pid \([0-9][0-9]*\)\.$/\1/p")"
+
+	case "${pid}" in
+		''|*[!0-9]*)
+			return 1
+			;;
+	esac
+
+	# Make sure it actually exists right now.
+	kill -0 "${pid}" 2>/dev/null || return 1
+
+	printf '%s\n' "${pid}"
+	return 0
+}
+
 rc_filenames="$(${RCORDER} /etc/rc.d/* /usr/local/etc/rc.d/* 2> /dev/null)"
 rc_filenames_defer=
 rc_filenames_reverse=
 rc_filenames_skip=
+service_stop_timeout="${service_stop_timeout:-15}"
 
 for rc_filename in ${rc_filenames}; do
 	eval "$(grep "^name[[:blank:]]*=" ${rc_filename})"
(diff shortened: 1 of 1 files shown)


src/etc/rc.freebsd
```diff
@@ -121,5 +145,42 @@ for rc_filename in ${rc_filenames}; do
...
+	if [ "${1}" != "stop" ]; then
+		${rc_filename} ${1}
+		continue
+	fi
+
+	service_pid="$(get_service_pid "${rc_filename}" "${name}")"
+
+	if [ -z "${service_pid}" ]; then
+		# no valid pid through status, stop without watchdog
+		${rc_filename} stop
```
fichtner: if we want to be thorough perhaps we should not assume stop will ever return and have to start the watchdog beforehand? at least this could make it return unless things are overly wedged, but then there's also no chance to a graceful shutdown beyond the timeout

fichtner: sorry I misread this earlier

src/etc/rc.freebsd
```diff
@@ -79,10 +79,34 @@ rc_enabled()
...
+	    sed -n "s/^${service_name} is running as pid \([0-9][0-9]*\)\.$/\1/p")"
+
+	case "${pid}" in
+		''|*[!0-9]*)
+			return 1
+			;;
+	esac
+
+	# Make sure it actually exists right now.
+	kill -0 "${pid}" 2>/dev/null || return 1
```
fichtner: I don't think this matters much. If the watchdog encounters the PID is should kill it and log, otherwise just exit

src/etc/rc.freebsd
```diff
@@ -79,10 +79,34 @@ rc_enabled()
...
+
+	printf '%s\n' "${pid}"
+	return 0
+}
+
 rc_filenames="$(${RCORDER} /etc/rc.d/* /usr/local/etc/rc.d/* 2> /dev/null)"
 rc_filenames_defer=
 rc_filenames_reverse=
 rc_filenames_skip=
+service_stop_timeout="${service_stop_timeout:-15}"
```
fichtner: shouldn't this be 90 seconds then? :)

fichtner: PS: doesn't really need this to be passed from environment

src/etc/rc.freebsd
```diff
@@ -121,5 +145,32 @@ for rc_filename in ${rc_filenames}; do
 		continue
 	fi
 
-	${rc_filename} ${1}
```
fichtner, line 124: I'd like to retain this line, first you make a if for stop, start the watchdog if needed, underneath make sure to kill the watchdog if watchdog_pid was set?

src/etc/rc.freebsd
```diff
@@ -79,10 +79,34 @@ rc_enabled()
 	esac
 }
 
+get_service_pid()
+{
+	rc_filename="$1"
+	service_name="$2"
+
+	status="$(${rc_filename} status 2>/dev/null)" || return 1
```
fichtner: how about:

pid=$(${rc_filename} status 2> /dev/null | awk '{gsub(/[^0-9]/, "", $NF); print $NF}')

src/etc/rc.freebsd
swhite2: > shouldn't this be 90 seconds then? :)

I'd like to keep this on the short side (per service). At the end of the day we're not running critical database software. The freebsd value of 90 was global for all services

fichtner: understand, but: this pertains to maybe 1-2 services only when it actually matters unlike the freebsd approach. 30-60 seconds doesn't hurt.

src/etc/rc.freebsd
```diff
@@ -115,11 +121,36 @@ fi
...
+					kill -KILL "${current_pid}" 2>/dev/null
+				else
+					echo " Service ${name} stop timed out, but pid disappeared"
+				fi
+			) &
+			watchdog_pid=$!
+		fi
+	fi
+
+	${rc_filename} "${1}"
```
fichtner:
```suggestion
	${rc_filename} ${1}
```

keep this as is, no need to change the behaviour if someone calls this wrong it may fail to work

fichtner (approved):
neat, thanks!

PR #10815, opened 2026-08-28
Firewall: NAT: Source NAT: add pool options and source hash key

### LINKED ISSUE #10814: Missing Outbound/Source NAT with Pool Option 'Round Robin with Sticky Address'
**Important notices**

No AI was used.

**Describe the bug**

After updating from 26.7.2 to 26.7.3_2, the ability to define legacy outbound NAT rules is missing. Source NAT does not yet support mapping to a NAT pool while using "Round Robin with Sticky Address". Currently, it appears to be a true round-robin, in which the same client may appear externally with different multiple IP addresses from the defined NAT Pool.

**To Reproduce**

Fresh installed 26.7 with Update to 26.7.2 shows missing legacy Outbound NAT after Update to 26.7.3_2

**Expected behavior**

At least least Outbound NAT should be visible until Source NAT is feature complete.

Created a Source NAT rule with translation to a set of virtual IPs (IP Alias). The translation works but round robin is not sticky. The same client appears with all IP aliases in the NAT pool on the outside.

**Additional context**

Since we have several fairly large private IP address ranges in use, we need to translate the private addresses to a NAT pool where each client gets translated to the same ip address (sticky) during its session.

**Environment**

OPNsense VM running on vSphere ESX with 26.7.3_2.

Monviech:
**Important notices**

- Model used:
- Extent of AI involvement:

---

**Describe the problem**

Adds missing pool options and source hash key to SNAT

---

Add them. Also to the SNAT migration export.

---

**Related issue**

Fixes: https://github.com/opnsense/core/issues/10814

### DIFF
--- a/src/opnsense/mvc/app/models/OPNsense/Firewall/Filter.xml
+++ b/src/opnsense/mvc/app/models/OPNsense/Firewall/Filter.xml
@@ -531,6 +531,17 @@
                     <EnableAlias>Y</EnableAlias>
                     <ValidationMessage>Please specify a valid port number, well-known name, alias or range.</ValidationMessage>
                 </target_port>
+                <poolopts type="OptionField">
+                    <BlankDesc>Default</BlankDesc>
+                    <OptionValues>
+                        <rr value="round-robin">Round Robin</rr>
+                        <rrsa value="round-robin sticky-address">Round Robin with Sticky Address</rrsa>
+                        <random>Random</random>
+                        <randomsa value="random sticky-address">Random with Sticky Address</randomsa>
+                        <source-hash>Source Hash</source-hash>
+                        <bitmask>Bitmask</bitmask>
+                    </OptionValues>
+                </poolopts>
                 <staticnatport type="BooleanField"/>
                 <log type="BooleanField">
                     <Default>0</Default>
--- a/src/opnsense/mvc/app/models/OPNsense/Firewall/Filter.xml
+++ b/src/opnsense/mvc/app/models/OPNsense/Firewall/Filter.xml
@@ -534,12 +534,12 @@
                 <poolopts type="OptionField">
                     <BlankDesc>Default</BlankDesc>
                     <OptionValues>
-                        <rr value="round-robin">Round Robin</rr>
-                        <rrsa value="round-robin sticky-address">Round Robin with Sticky Address</rrsa>
-                        <random>Random</random>
-                        <randomsa value="random sticky-address">Random with Sticky Address</randomsa>
-                        <source-hash>Source Hash</source-hash>
-                        <bitmask>Bitmask</bitmask>
+                        <p1 value="round-robin">Round Robin</p1>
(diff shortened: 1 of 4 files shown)


src/opnsense/mvc/app/models/OPNsense/Firewall/Filter.xml
```diff
@@ -531,6 +531,21 @@
...
                 </target_port>
+                <poolopts type="OptionField">
+                    <BlankDesc>Default</BlankDesc>
+                    <OptionValues>
+                        <rr value="round-robin">Round Robin</rr>
+                        <rrsa value="round-robin sticky-address">Round Robin with Sticky Address</rrsa>
+                        <random>Random</random>
+                        <randomsa value="random sticky-address">Random with Sticky Address</randomsa>
+                        <source-hash>Source Hash</source-hash>
+                        <bitmask>Bitmask</bitmask>
```
fichtner: this stuff is hard to process visually. it would be getter to use a mock element name p1.. p6 and use value= consistenly

Monviech: Sure no problem

fichtner: much better, thanks!

src/opnsense/mvc/app/models/OPNsense/Firewall/Filter.xml
```diff
@@ -531,6 +531,21 @@
                     <EnableAlias>Y</EnableAlias>
                     <ValidationMessage>Please specify a valid port number, well-known name, alias or range.</ValidationMessage>
                 </target_port>
+                <poolopts type="OptionField">
+                    <BlankDesc>Default</BlankDesc>
```
fichtner: not entirely sure about this but can tweak it later also

fichtner (approved):

src/opnsense/mvc/app/models/OPNsense/Firewall/Filter.xml
Monviech: The empty default also becomes round-robin when checking via pfctl

fichtner: true, maybe something for the help text instead of the label here. just for clarity.

Monviech: It's implicit:

**rule.debug**
nat on hn0 inet proto tcp from $red port $ports_80_443 to $dnsmasq port $ports_80_443 -> $alias_a port 53 # port_alias_1

**pfctl**
nat on hn0 inet proto tcp from <red> port 80:443 to <dnsmasq> port 80:443 -> <alias_a> port 53 round-robin

fichtner: I know, but the user may not know

Monviech: how about the helptext change now?

PR #10818, opened 2026-08-28
diagnostics: disable nginx reverse proxy buffering on configd streams

Greelan:
**Important notices**

- Model used:
- Extent of AI involvement:

---

**Describe the problem**

The dashboard traffic and CPU widgets and the live firewall log stutter behind a buffering nginx reverse proxy.

---

These event streams follow the same pattern as the live log widget, so adding the same `X-Accel-Buffering` header as was added for the live log widget in #10086.

The alternative is to do it centrally in `ApiControllerBase.php`.

---

**Related issue**

N/A

### DIFF
--- a/src/opnsense/mvc/app/controllers/OPNsense/Base/ApiControllerBase.php
+++ b/src/opnsense/mvc/app/controllers/OPNsense/Base/ApiControllerBase.php
@@ -227,10 +227,8 @@ class ApiControllerBase extends ControllerRoot
     ) {
         $response = (new Backend())->configdpStream($action, $params, $poll_timeout);
 
-        if (in_array('Content-Type: text/event-stream', $headers)) {
-            /* disable nginx reverse proxy buffering so SSE streams are delivered in real time */
-            $headers[] = 'X-Accel-Buffering: no';
-        }
+        /* disable nginx reverse proxy buffering so streamed configd output is delivered in real time */
+        $headers[] = 'X-Accel-Buffering: no';
         foreach ($headers as $header) {
             $parts = explode(':', $header, 2);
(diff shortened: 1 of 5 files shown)


AdSchellevis:
Maybe its better to merge a header before 

https://github.com/opnsense/core/blob/633a313c80dfa63fa697882bd8ddef44af970ff3/src/opnsense/mvc/app/controllers/OPNsense/Base/ApiControllerBase.php#L230

Personally I do find their header a bit obscure as we're already asking to not cache this stream with standard headers, in practice other webservers will usually just drop this anyway.

Greelan:
Right, that was the alternative I mentioned. Wasn't sure you would want Base touched. Anyway, done now.

AdSchellevis:
in this case it might be better, but is `X-Accel-Buffering` really connected to a certain type? I would expect streams should never be cached in between..

Greelan:
I was focused on SSE. I think the only other use is for CSV exports, where buffering may actually help?

AdSchellevis:
I think all current callers use the same type, but if at some point it seems to make sense to filter these, we can have a look then. For now I would prefer a single strategy and wait for feedback. When there is a valid reason to not do this, we should probably add the tag with a `yes` and filter on the header itself.

Greelan:
This is what I was referring to:
https://github.com/opnsense/core/blob/02b7e4cb5ef4885aca048ec9ed48de95178499e4/src/opnsense/mvc/app/controllers/OPNsense/Diagnostics/Api/LogController.php#L82-L91

But I can remove the type test if that is still your preference.

AdSchellevis:
let's remove the test, if someone complains, we'll see

Greelan:
OK, done.

Issue #10829, opened 2026-09-02
Support for pfsync version 1500


jakobsen-lrz:
**Important notices**

Before you add a new report, we ask you kindly to acknowledge the following:

- [ x] I have read the contributing guidelines at https://github.com/opnsense/core/blob/master/CONTRIBUTING.md
- [x ] I am convinced that my issue is new after having checked both open and closed issues at https://github.com/opnsense/core/issues?q=is%3Aissue

**Is your feature request related to a problem? Please describe.**
It is not possible to use the pfsync version 1500, which is the standard version for FreeBSD 15.

**Describe the solution you like**
I would like to be able to select pfsync version 1500 under the Hight Availibility settings.

fichtner:
[6264c38](https://github.com/opnsense/core/commit/6264c381207d15b474c3431fd7e3afa75d1dd8fd)

PR #10830, opened 2026-09-02
System: High Availability: Settings - Add pfsync version 1500

jakobsen-lrz:
**Important notices**

- [x ] I have read the contributing guidelines at https://github.com/opnsense/core/blob/master/CONTRIBUTING.md
- [x ] I opened an issue first for non-trivial changes and linked it below.

This adds the pfsync version 1500 to the model for HA Settings.

---

**Related issue**

#10829

### DIFF
--- a/src/opnsense/mvc/app/models/OPNsense/Core/Hasync.xml
+++ b/src/opnsense/mvc/app/models/OPNsense/Core/Hasync.xml
@@ -24,10 +24,11 @@
         </pfsyncpeerip>
         <pfsyncversion type="OptionField">
             <Required>Y</Required>
-            <Default>1400</Default>
+            <Default>1500</Default>
             <OptionValues>
                 <v1301 value="1301">OPNsense 24.1 or below</v1301>
-                <v1400 value="1400">OPNsense 24.7 or above</v1400>
+                <v1400 value="1400">OPNsense 24.7 - 26.1</v1400>
+                <v1500 value="1500">OPNsense 26.7 or above</v1500>
             </OptionValues>
         </pfsyncversion>
         <pfsyncdefer type="BooleanField"/>


src/opnsense/mvc/app/models/OPNsense/Core/Hasync.xml
```diff
@@ -24,10 +24,11 @@
         </pfsyncpeerip>
         <pfsyncversion type="OptionField">
             <Required>Y</Required>
-            <Default>1400</Default>
+            <Default>1500</Default>
```
fichtner, line 27: @AdSchellevis do we want to migrate here?

fichtner:
@jakobsen-lrz indeed, good find!

src/opnsense/mvc/app/models/OPNsense/Core/Hasync.xml
AdSchellevis: @fichtner changing the default should be enough, I don't expect the option to be backwards compatible.

fichtner: this not... just that nobody is going to benefit from new defaults except new installs if I remember this correctly. if that's a plan then fine

AdSchellevis: I'm afraid I can't come up with a better plan without breaking setups during the transition, you logically need a maintenance window to swap the sync on both nodes. Theoretically, one should change the versions on both nodes after both have been fully upgraded, if we would like to offer an upgrade, we likely should have a dialog in the hasync page signaling there's a newer version available.

fichtner: fair let's get this in and discuss later, thanks!

fichtner:
Merged, thanks!

PR #10848, opened 2026-09-06
suricata: rulesets: description fix

sopex:
**Important notices**

**Describe the problem**

Some XSS hardening killed the ruleset description links.

image

### DIFF
--- a/src/opnsense/www/js/opnsense.js
+++ b/src/opnsense/www/js/opnsense.js
@@ -229,17 +229,13 @@ function setFormData(parent,data) {
                             const value = node[keypart];
                             targetNode.empty();
                             if (targetNode.data('is_link')) {
-                                if (/^https?:\/\//i.test(value)) {
-                                    targetNode.append(
-                                        $("<a/>", {
-                                            href: value,
-                                            target: '_blank',
-                                            rel: 'noreferrer noopener'
-                                        }).text(value)
-                                    );
-                                } else {
-                                    targetNode.text(value);
-                                }
+                                targetNode.append(
+                                    $("<a/>", {
+                                        href: value,
+                                        target: '_blank',
+                                        rel: 'noreferrer noopener'
+                                    }).text(value)
+                                );
                             } else {
                                 targetNode.append(value);
                             }
--- a/src/opnsense/www/js/opnsense.js
+++ b/src/opnsense/www/js/opnsense.js
@@ -226,8 +226,23 @@ function setFormData(parent,data) {
                         targetNode.prop("checked", node[keypart] != 0);
                     } else if (targetNode.is("span")) {
                         if (node[keypart] != null) {
-                            targetNode.text("");
-                            targetNode.append(node[keypart]);
+                            const value = node[keypart];
+                            targetNode.empty();
+                            if (targetNode.data('is_link')) {
+                                if (/^https?:\/\//i.test(value)) {
+                                    targetNode.append(
+                                        $("<a/>", {
+                                            href: value,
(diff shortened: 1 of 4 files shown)


AdSchellevis:
I know, but this unfortunately opens the same gap as was closed with https://github.com/opnsense/core/security/advisories/GHSA-jm2f-ch62-rpxr, it's probably better to just replace the `<a href..` for the text that it encapsulates in 

https://github.com/opnsense/core/blob/b33d23ee9a71684d35be697575ee2aa89a26fc5e/src/opnsense/mvc/app/controllers/OPNsense/IDS/Api/SettingsController.php#L167-L178

sopex:
I don’t think this reopens [GHSA-jm2f-ch62-rpxr](https://github.com/opnsense/core/security/advisories/GHSA-jm2f-ch62-rpxr). In that case, data was passed as an HTML string to `.html()`, causing it to be reparsed as markup. Here, the `<a>` element is created as a DOM object, `href` is set via `.attr()`, and the visible value via `.text()`, so the URL is never parsed as HTML.

sopex:
This is even cleaner, but in practice, the same:

AdSchellevis:
It looks like I misread this indeed and also looked at the wrong grid. 

I'm not against fixing the link here, but I do think we should fix this more generic to improve maintainability.

My proposal would be to change the dialog to use url instead and change the type to `info_link` (new).

https://github.com/opnsense/core/blob/b33d23ee9a71684d35be697575ee2aa89a26fc5e/src/opnsense/mvc/app/controllers/OPNsense/IDS/forms/dialogRuleset.xml#L13-L17

Add the new `info_link` type in 

https://github.com/opnsense/core/blob/b33d23ee9a71684d35be697575ee2aa89a26fc5e/src/opnsense/mvc/app/views/layout_partials/form_input_tr.volt#L148-L149

Which adds some simple marker to know how to handle it, my suggestion would be `data-is_link=true`, next we implement a variation on your code in the following to wrap the value as a link:

https://github.com/opnsense/core/blob/b33d23ee9a71684d35be697575ee2aa89a26fc5e/src/opnsense/www/js/opnsense.js#L228-L231

Finally we need to check our documentation to see if we need to amend the new type (not sure if these are in there yet, but it would be practical at least).

sopex:
Something like this?

src/opnsense/www/js/opnsense.js
```diff
@@ -226,8 +226,23 @@ function setFormData(parent,data) {
                         targetNode.prop("checked", node[keypart] != 0);
                     } else if (targetNode.is("span")) {
                         if (node[keypart] != null) {
-                            targetNode.text("");
-                            targetNode.append(node[keypart]);
+                            const value = node[keypart];
+                            targetNode.empty();
+                            if (targetNode.data('is_link')) {
+                                if (/^https?:\/\//i.test(value)) {
```
AdSchellevis: when advertised as link, I think we can assume it's a link, right?

AdSchellevis (changes requested):
@sopex one small remark, but this looks good indeed

src/opnsense/www/js/opnsense.js
sopex: Indeed, it's probably a safe bet :)

sopex:
LGTM. I also opened a PR on the docs.

AdSchellevis:
@sopex looks good indeed, thanks a lot for doing this.

sopex:
@AdSchellevis Of course. I particularly like small improvements like these, and they always seem to come up :)

PR #10875, opened 2026-09-14
system: diag.disk rework

### LINKED ISSUE #9837: add  machine readable values in API system_disk
**Important notices**

When I monitor OPNsense via API and ask the /api/diagnostics/system/system_disk endpoint I got human readable output, but for monitoring (or in general) it would be great if we get machinereadable output.

the current situation looks like that:
```
[...]
    {
      "device": "zroot/ROOT/default",
      "type": "zfs",
      "blocks": "199G",
      "used": "1.3G",
      "available": "198G",
      "used_pct": 1,
      "mountpoint": "/"
    },
    {
      "device": "/dev/ada0p1",
      "type": "msdosfs",
      "blocks": "260M",
      "used": "1.7M",
      "available": "258M",
      "used_pct": 1,
      "mountpoint": "/boot/efi"
    },
[...]
```

great would be:
```
[...]
    {
      "device": "zroot/ROOT/default",
      "type": "zfs",
      "blocks": "208938564",
      "used": "1385252",
      "available": "207553312",
      "used_pct": 1,
      "mountpoint": "/"
    },
    {
      "device": "/dev/ada0p1",
      "type": "msdosfs",
      "blocks": "266144",
      "used": "1776",
      "available": "264368",
      "used_pct": 1,
      "mountpoint": "/boot/efi"
    },
[...]
```

also "block" is a little bit confusing, better would be "total"

**Additional context**

want to monitor opnsense with zabbix and API

Jorio2702:
**Important notices**

- [X] AI tools were used to create at least part of the code and text submitted herewith.

- Model used: GPT-5.6 Luna
- Extent of AI involvement: 
The formatBlocks() function was originally made by AI and then edited to fit my needs thereafter.

**Describe the problem**

Previously when calling the API endpoint system_disk (or systemDisk) it returned only the formatted storage values. For more detail this endpoint now returns those values and an unformatted amount of storage in bytes.
---

This pull request changes diag.disk to return raw blocks instead of formatted bytes so the API returns more info but keeps functionality the same. It also adds a new function, formatBlocks() that can be used to format the default blocks that diag.disk returns to the previously preformatted values. This function still needs a place where it can be used in both places it is currently being utilised (and possibly more where necessary).

---

**Related issue**

[#9837](https://github.com/opnsense/core/issues/9837)


src/opnsense/www/js/widgets/Disk.js
```diff
@@ -173,7 +173,7 @@ export default class Disk extends BaseGaugeWidget {
             let totals = [];
             for (const device of data.devices) {
                 let used = this._convertToBytes(device.used);
-                let total = this._convertToBytes(device.blocks);
+                let total = this._convertToBytes(device.total);
```
swhite2: I guess we can start using `device.used_bytes` and `device.total_bytes` and ditch `_convertToBytes` now

Jorio2702: Good call!

src/opnsense/mvc/app/controllers/OPNsense/Diagnostics/Api/SystemController.php
```diff
@@ -217,6 +225,35 @@ public function systemDiskAction()
         return $result;
     }
 
+    public function formatBlocks($blocks, $showUnit = true)
```
swhite2: Thinking about it, given the fact that a "block" is rather `df`-specific (including the `blocksize` of 512), it may be better to move this functionality to a backend python script, in it:

- fetch the output of df in json format
- format as this function would
- return all values as `systemDiskAction()` now currently does
- Shorten `systemDiskAction()` to return the result of `system diag disk` directly

At least in this way there's a single version of the truth for all relevant callers and we don't have to format in multiple places anymore.

Jorio2702: Great idea, will start work on this

src/opnsense/mvc/app/controllers/OPNsense/Diagnostics/Api/SystemController.php
```diff
@@ -192,29 +193,26 @@ public function systemResourcesAction()
...
+            if (!in_array(trim($fs['type']), ['cd9660', 'msdosfs', 'tmpfs', 'ufs', 'zfs'])) {
+                continue;
             }
+
+            // Blocks is deprecated and should not be used in future applications.
+            $devices[] = ['blocks' => $fs['total']] + $fs;
         }
 
-        return $result;
+        $disk_info['devices'] = $devices;
```
swhite2: likely best to move this logic to the backend script as well. You can then simply return something like

`return json_decode((new Backend())->configdRun('system diag disk'), true) ?? [];`

src/opnsense/mvc/app/controllers/OPNsense/Diagnostics/Api/SystemController.php
```diff
@@ -66,6 +66,7 @@ public function memoryAction()
         if (empty($data) || !is_array($data)) {
             return [];
         }
+
```
swhite2:
```suggestion
```

Best to keep the scope of the PR and resulting diff focused

src/opnsense/scripts/system/disk_info.py
```diff
@@ -0,0 +1,95 @@
...
+                "available_bytes": fs["available-blocks"] * 512,
+                "used_pct": fs["used-percent"],
+                "mountpoint": fs["mounted-on"],
+            })
+
+    return result
+
+if __name__ == "__main__":
+    result = disk_info()
+    print(json.dumps(result, indent=4))
```
swhite2: No need to pretty print :)

src/opnsense/scripts/system/disk_info.py
```diff
@@ -0,0 +1,95 @@
...
+        unit_index += 1
+
+    if value >= 100:
+        formatted = f"{value:.0f}"
+    elif value >= 10:
+        formatted = f"{value:.1f}"
+    else:
+        formatted = f"{value:.2f}"
+
+    if show_unit:
```
swhite2: Can likely drop this, there are no users of `show_unit`

swhite2 (changes requested):
Couple of little nitpicks, but otherwise looks good.

src/opnsense/scripts/system/disk_info.py
```diff
@@ -0,0 +1,99 @@
...
+
+            result["devices"].append({
+                "device": fs["name"],
+                "type": fs_type,
+
+                # "blocks" is deprecated and should not be used in future applications.
+                "blocks": format_blocks(fs["total-blocks"]),
+                # Use "total" instead:
+                "total": format_blocks(fs["total-blocks"]),
+                "total_bytes": fs["total-blocks"] * 512,
```
swhite2, line 85: One more concern I have here - if we're going to hardcode 512 and everywehre else in the script, we should probably use the `-b` option for df as well to ensure this remains the same.

src/opnsense/scripts/system/disk_info.py
```diff
@@ -0,0 +1,99 @@
...
+            fs_type = fs["type"].strip()
+
+            if fs_type not in ['cd9660', 'msdosfs', 'tmpfs', 'ufs', 'zfs']:
+                continue
+
+            result["devices"].append({
+                "device": fs["name"],
+                "type": fs_type,
+
+                # "blocks" is deprecated and should not be used in future applications.
```
swhite2:
```suggestion
                # XXX "blocks" provided for backwards compatibility
```

src/opnsense/scripts/system/disk_info.py
```diff
@@ -0,0 +1,99 @@
...
+                "blocks": format_blocks(fs["total-blocks"]),
+                # Use "total" instead:
```
swhite2:
```suggestion
```

swhite2 (changes requested):

src/opnsense/scripts/system/disk_info.py
```diff
@@ -0,0 +1,99 @@
...
+        formatted = f"{value:.2f}"
+
+    return formatted + units[unit_index]
+
+
+def disk_info():
+    result = {}
+
+    process = subprocess.run(
+        ["/bin/df", "-aT", "--libxo", "json"],
```
Jorio2702:
```suggestion
        ["/bin/df", "-abT", "--libxo", "json"],
```

swhite2 (approved):
Nice!

PR #10802, opened 2026-08-27
Interfaces: Assignments - next round of attributes,  closes https://github.com/opnsense/core/issues/10568

AdSchellevis:
o add media type and options
o most common ipv6 dhcp advanced options
o support "raw", but validated, send and receive options for both dhcp and dhcp6

This should complete the interfaces configuration section.

PR #10740, opened 2026-08-19
Interfaces: Assignments - add interface configuration settings in new assignments page

AdSchellevis:
**Important notices**

Before you submit a pull request, we ask you kindly to acknowledge the following:

- [x] I have read the contributing guidelines at https://github.com/opnsense/core/blob/master/CONTRIBUTING.md
- [x] I opened an issue first for non-trivial changes and linked it below.
- [ ] AI tools were used to create at least part of the code and text submitted herewith.

If AI was used, please disclose:

- Model used:
- Extent of AI involvement:

---

**Describe the problem**

This is part of a larger challenge, started with https://github.com/opnsense/core/issues/9945, the assignment of interfaces. Eventually we plan to merge the configuration pages with the assignments, which overs a more uniform user experience, API support and better support for larger sets of interfaces (due to not polluting the menu).

---

**Describe the proposed solution**

Refactor NetworkInterface model to reuse existing property names as much as possible, move from/to legacy logic into a custom fieldtype and store all legacy settings in a container named "pending" to ease reconfiguration and updating legacy configurations.

Skip wireless and other advanced settings for now, only implement a minimal set of validations.

---

**Related issue**

https://github.com/opnsense/core/issues/10568

PR #10389, opened 2026-06-08
mvc: Fix idassoc.php converting already decimal stored prefix_id via hexdec(), add unit test for idassoc.php

Monviech:
**Important notices**

Before you submit a pull request, we ask you kindly to acknowledge the following:

- [x] I have read the contributing guidelines at https://github.com/opnsense/core/blob/master/CONTRIBUTING.md
- [x] I opened an issue first for non-trivial changes and linked it below.
- [x] AI tools were used to create at least part of the code submitted herewith.

If AI was used, please disclose:

- Model used: ChatGPT 5.5
- Extent of AI involvement: Code autocompletion while writing the unit test

---

**Describe the problem**

https://github.com/opnsense/core/pull/10252#discussion_r3369553138

---

**Describe the proposed solution**

Remove hexdec() and add a unit test so regressions in that part can be caught early.

---

**Related issue**

https://github.com/opnsense/core/pull/10252

PR #10879, opened 2026-09-15
interfaces: round-trip the DHCPv6/track6 hex ID fields

### LINKED ISSUE #10878: Interfaces: Assignments does not round-trip the DHCPv6 / track6 hex ID fields
**Important notices**

- Model used: Claude Opus 5 (Claude Code)
- Extent of AI involvement: AI-assisted source analysis, reproduction scripting, duplicate search and drafting. The reproduction was run in a lab VM I control — a fresh `opnsense-devel` install, no patches — and I read the resulting `config.xml`, `dhcp6c.conf` and model state directly rather than relying on a summary. I reviewed this text before submitting.

**Describe the bug**

> **TL;DR** — Type `0x9` into *Optional interface ID* on a DHCPv6 WAN and save. It comes back as `0x0`. Two keys belonging to a mode the interface isn't even using (`track6-prefix-id`, `track6_ifid`) are created on **every** interface as the string `0x0`, which `is_numeric()` rejects downstream — silently dropping the `prefix-interface` block from `dhcp6c.conf`.

`NetworkInterfaceContainerField::toLegacy()` emits the four hex ID fields only when they are **not** set:

```php
foreach (['dhcp6-prefix-id', 'dhcp6_ifid', 'track6-prefix-id', 'track6_ifid'] as $fld) {
    if (!$this->$fld->isSet()) {
        $result[$fld] = intval($this->$fld->getValue(), 16);
    }
}
```

`BaseField::isSet()` is `getValue() !== ''`, so the conversion runs only for fields left empty. `dhcp6-prefix-id` and `dhcp6_ifid` are also in the `$skiplist` above, so when one of them *is* set it reaches the legacy array by no path at all. `track6-prefix-id` and `track6_ifid` are not in the skiplist, so the generic copy loop writes the model's own `0x…` representation into the legacy tree verbatim.

This has two halves, from two commits. 92cb28a669 (#10740) put `dhcp6-prefix-id` and `dhcp6_ifid` in the `$skiplist` but wrote the condition correctly, as `if (!$this->$fld->isEmpty())` — emit when the field *has* a value. 6e4dab1042 (#10802) then changed it to `if (!$this->$fld->isSet())`, keeping the `!` while swapping to the opposite predicate:

```diff
-            if (!$this->$fld->isEmpty()) {     /* 92cb28a669 */
+            if (!$this->$fld->isSet()) {       /* 6e4dab1042 */
                 $result[$fld] = intval($this->$fld->getValue(), 16);
             }
```

The same commit relaxed `fromLegacy()` from `!empty($data[$fld])` to `isset($data[$fld]) && $data[$fld] !== ''`, which integer `0` passes — that is what manufactures the `0x0` values.

Instrumenting the live model after the save described below:

```
MODEL  dhcp6_ifid         value='0x0' isSet=true  isEmpty=false
LEGACY dhcp6_ifid         <<ABSENT>>
MODEL  track6_ifid        value='0x0' isSet=true  isEmpty=false
LEGACY track6_ifid        '0x0'
```

The legacy side treats these keys as decimal integers — `src/etc/inc/interfaces.inc:2919` guards with `is_numeric()`, and `is_numeric('0x0')` is `false` on PHP 7+.

Last known working version: not applicable — this code path has never shipped in a release; it is master-only.

**To Reproduce**

Fresh install, WAN = DHCPv6, LAN = static. All four ID keys are absent from `config.xml` to begin with.

1. Go to Interfaces → Assignments and edit **wan**.
2. Type `0x9` into **Optional interface ID** (`interface.dhcp6_ifid`). Change nothing else.
3. Click Save, then Apply.
4. Reopen the same dialog — *Optional interface ID* now reads `0x0`.
5. Diff `config.xml` against a copy taken before step 2:

```diff
@@ <interfaces><wan> @@
       <dhcp6_norequest_dns>0</dhcp6_norequest_dns>
+      <track6-prefix-id>0x0</track6-prefix-id>
+      <track6_ifid>0x0</track6_ifid>
+      <dhcp6_ifid>0</dhcp6_ifid>
     </wan>
@@ <interfaces><lan> @@
       <dhcp6_norequest_dns>0</dhcp6_norequest_dns>
+      <track6-prefix-id>0x0</track6-prefix-id>
+      <track6_ifid>0x0</track6_ifid>
     </lan>
```

Three things go wrong in that one save: the typed `0x9` is not stored, `dhcp6_ifid` is written as `0`, and two track6 keys are created on both interfaces although neither is in Track Interface mode and those inputs are not rendered in the dialog.

**Downstream effect.** Single-variable test, LAN genuinely in Track Interface mode tracking WAN. Only `track6-prefix-id` differs between the two runs, with `configctl interface reconfigure` in between:

```
##### track6-prefix-id = '0'   (what src/www/interfaces.php writes) #####
id-assoc pd 0 {
  prefix ::/64 infinity;
  prefix-interface vtnet0 {
    sla-id 0;
    sla-len 0;
    ifid 0x0;
  };
};

##### track6-prefix-id = '0x0' (what the Assignments dialog writes) #####
id-assoc pd 0 {
  prefix ::/64 infinity;
};
```

The entire `prefix-interface` block is gone — prefix delegation to the tracked interface is silently not configured. Note also `ifid 0x0;` in the first run: `track6_ifid` is interpolated raw at `src/etc/inc/interfaces.inc:2925`, and dhcp6c expects a decimal number there.

**Expected behavior**

- A value typed into one of the four hex ID fields is stored, and reads back unchanged.
- Fields the operator did not touch are not created.
- The legacy tree receives decimal integers, as `src/www/interfaces.php` writes them (`intval($pconfig[...], 16)`, lines 126 and 129).

Leaving the `0x` form in the legacy tree and teaching the consumers to accept it. That looked worse: `src/www/interfaces.php` and the Assignments dialog would still disagree about the stored format, and every `is_numeric()` / `sprintf("%x", …)` / `str_pad()` site in `interfaces.inc` would need auditing. Making `toLegacy()` write what the legacy page already writes keeps one format in `config.xml`.

**Screenshots**

Attached: the same dialog before and after Save + Apply, with the affected field called out.

Optional interface ID: 0x9 typed, 0x0 stored

**Relevant log files**

No log entries are produced — nothing errors. The evidence is `config.xml` and `/var/etc/dhcp6c.conf`, both quoted above.

**Additional context**

`src/www/interfaces.php` and the Assignments dialog write the same keys in different formats, so the two pages disagree about what is stored.

The decimal form is the documented contract. `src/opnsense/mvc/tests/app/library/OPNsense/Interface/IdassocTest.php:48-51`, added in #10389, states that `track6-prefix-id` is a decimal string in the config even though the GUI shows it as hex. #10389 fixed the mirror image of this bug — a `hexdec()` applied to an already-decimal value — and added a unit test to catch regressions. The Assignments dialog now writes the `0x…` form that test says should not be there.

I have a patch and will open a PR referencing this issue.

Not claimed: I have not run dhcp6c against a real ISP with these files — the breakage is demonstrated at the generated-config level.

**Environment**

OPNsense 27.1.a_263 (amd64), `opnsense-devel` flavor, FreeBSD kernel 15.1-RELEASE-p1.
Reproduced on a freshly installed disposable VM (qemu, 2 NICs: WAN DHCP/DHCPv6, LAN static), switched to the development flavor with `opnsense-update -t opnsense-devel`. No patches applied; `pkg check -s opnsense-devel` is clean.
Hardware: virtual machine.

IvanTheGeek:

Model used: Claude Opus 5 (Claude Code)
Extent of AI involvement: AI-assisted source analysis, reproduction scripting and drafting. The
reproduction and the patched build were run in a lab VM I control — a fresh `opnsense-devel`
install — and I read the resulting `config.xml`, `dhcp6c.conf` and model state directly. I
reviewed this patch and text before submitting.

---

**Describe the problem**

Tracked in #10878.

`NetworkInterfaceContainerField::toLegacy()` guards the hex-to-decimal conversion for
`dhcp6-prefix-id`, `dhcp6_ifid`, `track6-prefix-id` and `track6_ifid` with
`if (!$this->$fld->isSet())` — the legacy value is emitted only for fields the operator left
**empty**, and a field they actually filled in is never emitted at all.

This is a polarity inversion introduced by 6e4dab1042 (#10802). The skiplist entries come from
92cb28a669 (#10740), which wrote the condition correctly:

```php
-            if (!$this->$fld->isEmpty()) {     /* 92cb28a669 — emit when the field HAS a value */
+            if (!$this->$fld->isSet()) {       /* 6e4dab1042 — emits only when it has NONE */
                 $result[$fld] = intval($this->$fld->getValue(), 16);
             }
```

`isEmpty()` is `empty($value)` and `isSet()` is `$value !== ''`
(`src/opnsense/mvc/app/models/OPNsense/Base/FieldTypes/BaseField.php:442-454`). Swapping one for
the other while keeping the `!` reverses the condition for every non-empty value.

`dhcp6-prefix-id` and `dhcp6_ifid` are also in the `$skiplist` above, so for those two a set
value reaches `$result` by no path at all. `track6-prefix-id` and `track6_ifid` are **not** in
the skiplist, so the generic copy loop writes the model's own `0x…` string into the legacy tree
instead. The same commit also relaxed `fromLegacy()` from `!empty($data[$fld])` to
`isset($data[$fld]) && $data[$fld] !== ''`, which integer `0` now passes — that is what
manufactures the `0x0` values in the first place.

The legacy side treats these keys as decimal integers: `src/etc/inc/interfaces.inc:2919` guards
with `is_numeric()`, and `is_numeric('0x0')` is `false` on PHP 7+, which silently drops the whole
`prefix-interface` block from `dhcp6c.conf`.

---

Skiplist the two track6 fields alongside the two dhcp6 ones, and use `isSet()` without the `!`.

I chose `isSet()` rather than restoring the literal `!isEmpty()` because `empty('0')` is true, so
the original would silently drop a legitimate interface ID of `0`. Measured on PHP 8.5.8, over
every value the field can hold:

```
value    | !isEmpty()  [92cb28a669] | !isSet()    [6e4dab1042] | isSet()     [proposed]
---------+--------------------------+--------------------------+------------------------
''       | omit                     | emit 0                   | omit
'0'      | omit                     | omit                     | emit 0
'0x0'    | emit 0                   | omit                     | emit 0
'0x9'    | emit 9                   | omit                     | emit 9
'0xfff'  | emit 4095                | omit                     | emit 4095
```

An unset field stays absent from the legacy array, matching a fresh install. Writing `''` instead
would be wrong: `src/etc/inc/interfaces.inc:2909` tests these keys with `isset()` and would then
emit a bare `ifid ;`.

```diff
--- a/src/opnsense/mvc/app/models/OPNsense/Interfaces/FieldTypes/NetworkInterfaceField.php
+++ b/src/opnsense/mvc/app/models/OPNsense/Interfaces/FieldTypes/NetworkInterfaceField.php
@@ -66,6 +66,8 @@ class NetworkInterfaceContainerField extends ContainerField
             'dhcp6_request_dns',
             'dhcp6-prefix-id',
             'dhcp6_ifid',
+            'track6-prefix-id',
+            'track6_ifid',
             'media',
         ];
         foreach ($this->iterateItems() as $key => $node) {
@@ -76,7 +78,7 @@ class NetworkInterfaceContainerField extends ContainerField
         }
         $result['dhcp6_norequest_dns'] = $this->dhcp6_request_dns->isEmpty() ? '1' : '0';
         foreach (['dhcp6-prefix-id', 'dhcp6_ifid', 'track6-prefix-id', 'track6_ifid'] as $fld) {
-            if (!$this->$fld->isSet()) {
+            if ($this->$fld->isSet()) {
                 $result[$fld] = intval($this->$fld->getValue(), 16);
             }
         }
```

---

**Tested**

A regression test is included: `tests/app/models/OPNsense/Interfaces/FieldTypes/NetworkInterfaceFieldTest.php`,
following the pattern #10389 set for `IdassocTest`. It fails on unpatched master and passes with
this change.

```
# unpatched master
FAILURES!  Tests: 6, Assertions: 7, Failures: 6.
    dhcp6-prefix-id should not be written
      Failed asserting that an array does not have the key 'dhcp6-prefix-id'.
    dhcp6_ifid must satisfy is_numeric()
      Failed asserting that false is true.

# with this patch
OK (6 tests, 16 assertions)
```

Full suite, FreeBSD, PHP 8.5.8, PHPUnit 9.6.36:

```
master:                    OK (308 tests, 736 assertions)
this patch:                OK (308 tests, 736 assertions)
this patch + new test:     OK (314 tests, 752 assertions)
```

`make lint` exits 0 with the same 404 findings as unpatched master, and does not mention the new
file. The new test needed a `plist` entry — `make plist-fix` adds it, and it is included here;
without it `lint-plist` fails.

OPNsense 27.1.a_263-amd64 (`opnsense-devel`), fresh install, all four ID keys absent to start
with. Patch applied to the installed tree, `pkill -x php-cgi` to recycle the workers.

Model-level probe:

```
--- after setting dhcp6_ifid = 0x9 ---
  dhcp6_ifid         value='0x9' isSet=true
--- toLegacy() ---
  dhcp6-prefix-id    <<ABSENT>>
  dhcp6_ifid         9
  track6-prefix-id   <<ABSENT>>
  track6_ifid        <<ABSENT>>
```

GUI round-trip — Assignments ▸ wan ▸ `0x9` into Optional interface ID ▸ Save ▸ Apply:

```diff
@@ <interfaces><wan> @@
+      <dhcp6_ifid>9</dhcp6_ifid>
```

Decimal `9`, the same representation `src/www/interfaces.php` writes
(`intval($pconfig[...], 16)`, lines 126/129). No track6 keys created on either interface.
Reopening the dialog shows `0x9`.

`make lint` and `make style` were run natively on FreeBSD against unpatched master and against
this patch, and the output compared:

```
make lint  : master 404 findings   patched 404 findings   IDENTICAL
make style : master 2259 findings  patched 2259 findings  IDENTICAL
```

(compared after normalizing line numbers, which shift by the lines this patch adds). The two
phpcs errors on `NetworkInterfaceField.php` — `$pppDevices` visibility and "Each class must be in
a file by itself" — are present on unpatched master; the second is the two-classes-per-file
pattern this codebase uses throughout (`FilterRuleField`, `SourceNatRuleField`, `CAsField`).

The new test file carries one `PSR1.Files.SideEffects` warning for its `require_once` — the
container class shares a file with `NetworkInterfaceField`, so the autoloader cannot reach it by
name. `KeaOptionDataFieldTest` and `HostnameFieldTest` carry the same warning for the same reason.

⚠️ If you verify this by editing the installed files rather than via `opnsense-patch`, recycle the
PHP workers afterwards (`pkill -x php-cgi`). `opcache.enable` is On and the GUI keeps serving the
pre-patch code otherwise — the first save after patching then changes nothing and looks like the
fix failed, while PHP CLI shows the patched behavior immediately.

⚠️ **This patch and the one for the boolean defect conflict textually.** That one's hunk carries
`if (!$this->$fld->isSet()) {` as trailing context, which this patch rewrites, so whichever lands
second needs a rebase. They touch adjacent lines, not the same one, so the resolution is trivial.

---

**Related issue**

Fixes #10878
Ref: the condition was inverted by 6e4dab1042 (#10802); the `$skiplist` entries that make a set
value unreachable come from 92cb28a669 (#10740), which had the condition right. Only three commits
have ever touched this file — the third is a style sweep.
Ref: #10389 fixed the mirror image of this (a `hexdec()` on an already-decimal value) and its
test documents the contract these keys are supposed to satisfy —
`tests/app/library/OPNsense/Interface/IdassocTest.php:48-51` says `track6-prefix-id` is a decimal
string in the config even though the GUI shows it as hex. This patch makes the Assignments dialog
honor that.


fichtner:
Your changes seem correct so thanks for them. At the moment I don't want to complicate the system with a unit test. It would create too much maintenance concerns while still under development.

Merged, thanks!

PR #10883, opened 2026-09-15
interfaces: remove stale VIP address after update

### LINKED ISSUE #10882: changing a VIP address or interface leaves the old address on the interface
**Important notices**

**Describe the bug**

Changing the address or interface of an existing Virtual IP does not remove the previous address from the interface.
The configuration is updated correctly, but the old address remains configured in the running system.
OPNsense version: **26.7.3_11**

**To Reproduce**

1. Create an IP Alias or CARP VIP, for example `10.252.1.1/24`.
2. Click **Apply** and confirm that the address is present on the interface with `ifconfig`.
3. Edit the existing VIP and change its address, for example from `10.252.1.1` to `192.0.2.77`.
   Alternatively, move the VIP to another interface.
4. Save the change.
5. Click **Apply**.
6. Check `/conf/config.xml` and confirm that only the new VIP address/interface is present.
7. Run `ifconfig <interface>`.

**Expected behavior**

Old address should be removed from interface.

ExpRam:
**Important notices**

---

**Describe the problem**

When an existing Virtual IP is edited, changing its address or interface can leave the previous address configured on the interface.

$node contains the current subnet and interface values. However, this call:
```
$result = $this->handleFormValidations($this->setBase('vip', 'vip', $uuid, $this->getVipOverlay()));
```
overwrites old node data with new POST data and all future checks are useless.

---

Store the current VIP subnet and interface before calling setBase().
Use these preserved values for change detection and for queuing the previous address for removal.

---

**Related issue**

Fixes: #10882


AdSchellevis:
I’m quite sure I tested changing addresses at some point, but looking at the code, I also think $node is a reference here which would be overwritten by the  setBase in the same block. This probably needs an extra test on our end to be sure.

PR #10884, opened 2026-09-15
IPsec: point strongSwan TPM plugin at the tpm(4) device

rkgfk:
Companion to the opnsense/tools change enabling `TPM TSS2` for `security/strongswan` (https://github.com/opnsense/tools/pull/519).

## Summary

When strongSwan is built with TPM support, the `tpm` plugin selects its
TCTI by checking for `/dev/tpmrm0`. FreeBSD has no in-kernel TPM resource
manager device (tpm(4) only creates `/dev/tpm0`), so charon falls back to
the D-Bus based `tabrmd` TCTI, which does not exist on OPNsense, and the
plugin refuses to load on every start.

This change adds a default `tpm { tcti { name = device  opts = /dev/tpm0 } }`
block to the generated `strongswan.conf` in `ipsec_write_strongswan_conf()`,
next to the existing legacy overwrites. A value already provided through
the model is left untouched.

## Backwards compatibility

* Systems without a TPM: the plugin loads (only the TCTI shared library is
  `dlopen()`ed) and never touches `/dev/tpm0` unless a TPM-bound key or
  certificate is referenced. Verified on a FreeBSD 15.1 VM without a TPM
  using the strongswan 6.1.0 package built with `TPM TSS2`:
  `plugin 'tpm': loaded successfully`, charon runs, and referencing a TPM
  key via `swanctl --load-creds` fails cleanly with `no TPM 2.0 found`
  while charon keeps running. Without this change the same setup logs
  `could not load "libtss2-tcti-tabrmd.so.0"` and `plugin 'tpm': failed to
  load` (charon still runs, but TPM keys can never be used).
* With an emulated TPM attached to the VM the same configuration loads a
  persisted ECDSA key from the TPM (`loaded ECDSA private key from token`)
  and charon signs with it.
* Systems running a strongswan package built without TPM support (e.g. the
  current 26.7 packages): charon ignores settings for plugins that are not
  present, so the additional block is harmless.

## Resulting strongswan.conf fragment

```
charon {
    ...
    plugins {
        ...
        tpm {
            tcti {
                name = device
                opts = /dev/tpm0
            }
        }
    }
}
```


AdSchellevis:
Best start with a ticket to explain the requested functionality, preferably including documentation on how one should use tpm on FreeBSD in these cases. We generally try to avoid adding features without ui implementation, but if it looks useful to add at some point, we can make exceptions.

rkgfk:
Thanks. I have opened a feature request that describes the use case, how a TPM 2.0 is used with strongSwan on FreeBSD (kernel module, TCTI, key creation with tpm2-tools, swanctl token secrets) and a proposal for a UI implementation in the key pair dialog: https://github.com/opnsense/core/issues/10886. The build side is in opnsense/tools#519; this PR only adds the FreeBSD-specific TCTI default so the plugin can load at all. Happy to work on the UI part if you consider the feature worth adding.

PR #10885, opened 2026-09-15
interfaces: omit false BooleanField values from the legacy config

### LINKED ISSUE #10881: Interfaces: Assignments writes unchecked options as "0", which isset() reads as enabled
**Important notices**

- Model used: Claude Opus 5 (Claude Code)
- Extent of AI involvement: AI-assisted source analysis, reproduction scripting, duplicate search and drafting. The reproduction was run in a lab VM I control — a fresh `opnsense-devel` install, no patches — and I read the resulting `config.xml`, `dhcp6c.conf` and model state directly rather than relying on a summary. I reviewed this text before submitting.

**Describe the bug**

> **TL;DR** — Open the Interfaces: Assignments dialog on a fresh install, change nothing, press Save. Interface options that are **off** get written to `config.xml` as the string `"0"` instead of being omitted. Large parts of the legacy tree presence-test these keys with `isset()`, which is true for `"0"` — so they read as **enabled**. The visible consequence: the WAN stops requesting an IPv6 address while "Request prefix only" is unticked in the GUI.

On a freshly installed system, opening the Interfaces: Assignments dialog and clicking Save
without changing anything writes interface options that are **off** into `config.xml` as `"0"`
— nine `BooleanField`s on WAN in the capture below, plus the derived `dhcp6_norequest_dns`.
Legacy code tests these keys with `isset()`, so a stored `"0"` reads as **enabled**, and the WAN
stops requesting a non-temporary IPv6 address.

`NetworkInterfaceContainerField::toLegacy()` emits `BooleanField` values via `getValue()`, which
returns the string `"0"` for false. Both writers then keep any value `!== ''`:

```php
foreach ($pending as $akey => $avalue) {
    if ($avalue !== '') {
        ... ->$akey = $avalue;              // "0" is written
    } elseif (isset(...)) {
        unset(...);
    }
}
foreach ($legacybools as $legacybool) {     // retroactive fix-up, 5 hardcoded keys
    if (empty($pending[$legacybool])) { unset(...); }
}
```

`$legacybools` covers `enable`, `lock` and the three `disable*offloading` keys — **5 of the 17**
`BooleanField`s in `NetworkInterface.xml`. The other 12 keep their `"0"`:

```
blockbogons  blockpriv  dhcp6-ia-pd-send-hint  dhcp6-information-only  dhcp6_rapid_commit
dhcp6_request_dns  dhcp6prefixonly  dhcpd6track6allowoverride  dhcphonourmtu
gateway_interface  hw_settings_overwrite  promisc
```

The compensation has always trailed the field set. 92cb28a669 (#10740) introduced the copy loop
and, in the same change, hand-unset two keys (`enable`, `lock`) in **both** writers — so the
mechanism was understood at the time. 6e4dab1042 (#10802) then named the list `$legacybools` and
grew the controller's copy to five, while adding five new `BooleanField`s to the model and
covering only three of them: `hw_settings_overwrite` and `dhcp6-information-only` arrived already
uncovered. `apply_pending_if_changes.php` still carries the original two-entry list, unchanged.

A hand-maintained list that has to track an XML-declared field set, in two places, with no test,
is the reason I would rather fix the transform.

There are **two** writers, each with its own hardcoded list, so a fix in one place is incomplete:
- `mvc/app/controllers/OPNsense/Interfaces/Api/AssignmentController.php:142-148` (5 entries)
- `scripts/interfaces/apply_pending_if_changes.php:74` (2 entries: `enable`, `lock`), reached via
  `configctl interface apply` before `config.xml` is written.

Note: I am **not** reporting the `adv_dhcp*` removal in the same method — the comment there
states it is deliberate.

**To Reproduce**

On a freshly installed system, WAN set to DHCP / DHCPv6 (the installer default):

1. Confirm the starting state — none of these keys exists in `/conf/config.xml`, and
   `/var/etc/dhcp6c.conf` contains both `send ia-na 0;` and `send ia-pd 0;`.
2. Go to Interfaces: Assignments, open the edit dialog for WAN.
3. **Change nothing.** Click Save, then Apply.
4. `diff` `/conf/config.xml` against a copy taken before step 2. Thirteen keys appear on WAN
   that were not there:

   ```diff
      <interfaces>
        <wan>
          ...
          <dhcp6-ia-pd-len>0</dhcp6-ia-pd-len>
   +      <promisc>0</promisc>
   +      <hw_settings_overwrite>0</hw_settings_overwrite>
   +      <disablevlanhwfilter>0</disablevlanhwfilter>
   +      <gateway_interface>0</gateway_interface>
   +      <dhcphonourmtu>0</dhcphonourmtu>
   +      <dhcp6prefixonly>0</dhcp6prefixonly>
   +      <dhcp6-information-only>0</dhcp6-information-only>
   +      <dhcp6-ia-pd-send-hint>0</dhcp6-ia-pd-send-hint>
   +      <dhcp6_rapid_commit>0</dhcp6_rapid_commit>
   +      <track6-prefix-id>0x0</track6-prefix-id>
   +      <track6_ifid>0x0</track6_ifid>
   +      <dhcpd6track6allowoverride>0</dhcpd6track6allowoverride>
   +      <dhcp6_norequest_dns>0</dhcp6_norequest_dns>
        </wan>
   ```

   `<lan>` gets the same thirteen, plus `<blockpriv>0</blockpriv>` and
   `<blockbogons>0</blockbogons>` — WAN already had those two as a genuine `1` from the
   installer, and they are correctly left alone.

   Thirteen keys, but they are not all this defect. Breaking them down by what the model
   declares them as:

   * **nine `BooleanField`s** — `promisc`, `hw_settings_overwrite`, `gateway_interface`,
     `dhcphonourmtu`, `dhcp6prefixonly`, `dhcp6-information-only`, `dhcp6-ia-pd-send-hint`,
     `dhcp6_rapid_commit`, `dhcpd6track6allowoverride`. **This is the defect.**
   * **`dhcp6_norequest_dns`** — not a model field at all; written by the explicit line
     `$result['dhcp6_norequest_dns'] = $this->dhcp6_request_dns->isEmpty() ? '1' : '0';`.
     Same defect, same one-character fix.
   * **`disablevlanhwfilter`** — an `OptionField` with `<Required>Y</Required>` and
     `<Default>0</Default>`, where `0` is a real choice ("Enable VLAN Hardware Filtering"), and
     `interfaces.inc:641,653` reads it with `!= 2` / `empty()`. Persisting its default is
     defensible. **I am not reporting this one.**
   * **`track6-prefix-id`, `track6_ifid`** — #10878's defect, not this one.

5. `/var/etc/dhcp6c.conf` has changed: `send ia-na 0;` and `id-assoc na 0 { };` are **gone**.

   Isolated to a single variable — only `<dhcp6prefixonly>` differs between these two runs,
   `configctl interface reconfigure wan` in between, nothing else touched:

   ```
   ##### dhcp6prefixonly = absent  (what src/www/interfaces.php leaves behind) #####
   5:  send ia-na 0;
   6:  send ia-pd 0;

   ##### dhcp6prefixonly = '0'     (what the Assignments dialog writes) #####
   5:  send ia-pd 0;
   ```

   The guard is `src/etc/inc/interfaces.inc:2948`, `if (!isset($wancfg['dhcp6prefixonly']))`.
   The WAN stops requesting an IPv6 address while **"Request prefix only" is unticked in the
   GUI**.

6. Go to Interfaces → [WAN] (the legacy page). Five options now display as **enabled**:
   "Dynamic gateway policy", "Promiscuous mode", "Send prefix hint", "Request prefix only" and
   "Manual configuration" — none of which was ever enabled. The next save of that page writes
   them as genuine `1`s.

7. A second consumer, independent of the first: the bogons table. `src/etc/inc/filter.inc:48`
   decides whether to load `bogonsv6` at all with `isset($ifacedata['blockbogons'])`, while the
   blocking rule itself at `src/etc/inc/filter.lib.inc:340` correctly uses `empty()`. Only the
   former sees a stored `"0"`, so a 75-entry table is loaded that no rule uses. Single variable,
   `configctl filter reload` between runs:

   ```
   wan='0'    lan='0'     -> bogonsv6 entries: 75    (what the dialog leaves behind)
   wan=absent lan=absent  -> bogonsv6 entries: 0
   wan='1'    lan=absent  -> bogonsv6 entries: 75    (control — genuinely enabled still works)
   ```

   Two consumers of the same key, tested two different ways, is what makes me think this belongs
   at the transform rather than in a per-key list.

**Expected behavior**

An option that is off is absent from `config.xml`, as `src/www/interfaces.php` writes it, so the
presence-testing consumers read it as disabled. A Save with no changes does not alter behavior.

Adding the affected keys to `$legacybools`. I do not think that is the right fix: the list is
hand-maintained against a 17-entry set declared in XML, with no test; it has drifted
before; and it would have to be duplicated into `apply_pending_if_changes.php`. I would
rather have `toLegacy()` emit `''` for a false `BooleanField` so both writers delete the node
through the branch they already have — but that is your design call, and I will rework the PR to
whichever layer you prefer.

**Screenshots**

The legacy [WAN] page after step 3. Three adjacent rows make the point on their own — all three
keys are stored identically as `"0"`, and only the read idiom differs:

Three adjacent rows on the legacy WAN page: Request prefix only and Send prefix hint render ticked, Send rapid commit renders unticked, all three stored as 0

| row | key | read at | renders |
|---|---|---|---|
| Request prefix only | `dhcp6prefixonly` | `interfaces.php:262` `isset()` | **on** ✗ |
| Send prefix hint | `dhcp6-ia-pd-send-hint` | `interfaces.php:261` `isset()` | **on** ✗ |
| Send rapid commit | `dhcp6_rapid_commit` | `interfaces.php:751` `!empty()` | off ✓ |

The full set of presence-tested reads on that page is lines 258 `gateway_interface`,
259 `promisc`, 261 `dhcp6-ia-pd-send-hint`, 262 `dhcp6prefixonly` and
267 `dhcpd6track6allowoverride`. Lines 256 `blockpriv` and 257 `blockbogons` use `!empty()` and
are correct, so the file is inconsistent with itself rather than uniformly wrong.

Excluded as a false positive: "Override MTU" also displays as enabled, but that is **not** this
defect — line 260 derives `dhcpoverridemtu` from an inverted `dhcphonourmtu` and `empty()` is
true for an absent key as well, so it looks the same on an untouched install.

Full pages for context — Interfaces → [WAN] after step 3:

Legacy WAN page, generic configuration section

Legacy WAN page, DHCPv6 client configuration section

**Relevant log files**

The `dhcp6c.conf` comparison is quoted inline under step 5 — the loss of `send ia-na 0;` is the
functional impact. Nothing errors, so there is no log entry to attach; the evidence is the
generated config.

**Additional context**

Related: #10878 / #10879 — a separate defect in the same method, which the same reproduction
triggers. That one is the hex ID round trip; this one is the booleans. They are independent fixes
and neither depends on the other.

The two patches also conflict textually: #10879's hunk rewrites
`if (!$this->$fld->isSet()) {`, which a fix for this issue carries as trailing context. Adjacent
lines, trivial to resolve, but whichever lands second needs a rebase.

I will open a PR for this alongside the report. The reason I would rather fix the transform than
extend `$legacybools` is that it removes the list entirely — there is then nothing left to fall
behind `NetworkInterface.xml`, which has already happened twice. The test that comes with it reads
every `BooleanField` declared in the model at test time, so a field added later is covered the day
it is added, with no edit to the test.

If you would rather solve it in the writers, or extend the lists, say so and I will rework it —
the diagnosis stands either way.

**Environment**

OPNsense 27.1.a_263 (amd64), `opnsense-devel` flavor, FreeBSD kernel 15.1-RELEASE-p1.
Reproduced on a **freshly installed** disposable VM (qemu, 2 NICs: WAN DHCP/DHCPv6, LAN static),
switched to the development flavor with `opnsense-update -t opnsense-devel`. No patches applied;
`pkg check -s opnsense-devel` is clean.
Released 26.7.x is **not** affected — this code is not present there.
Hardware: virtual machine.
Last known working version: not applicable — this code path has never shipped in a release.
It was introduced on master by 92cb28a669 (#10740), which added the generic copy loop
`$result[$key] = $node->getValue();`. Verified with
`git log -S '$result[$key] = $node->getValue();' -- <file>`, which returns that commit alone.
That same change added the two-key `['enable', 'lock']` unset to both writers, so the effect on
legacy consumers was understood from the start — just handled per key.

IvanTheGeek:

Model used: Claude Opus 5 (Claude Code)

Extent of AI involvement: AI-assisted source analysis, reproduction scripting and drafting. The
reproduction and the patched build were run in a lab VM I control — a fresh `opnsense-devel`
install — and I read the resulting `config.xml`, `dhcp6c.conf` and model state directly. I
reviewed this patch and text before submitting.

---

**Describe the problem**

Tracked in #10881. `toLegacy()` emits a false `BooleanField` as `"0"`, both writers keep any
value `!== ''`, and legacy consumers test presence — so unchecking an option enables it. Twelve of
the seventeen `BooleanField`s declared in `NetworkInterface.xml` are uncovered by the existing
rescue lists, and there are two writers each with its own hardcoded list.

---

Emit `''` for a false `BooleanField` in `toLegacy()`, so both writers delete the node via the
`elseif (isset(...)) unset(...)` branch they already have. Both rescue lists then become dead
code — `$legacybools` in `Api/AssignmentController.php` and the inline `['enable', 'lock']` in
`scripts/interfaces/apply_pending_if_changes.php`. This patch deliberately leaves both in place,
so it is a no-op for anything that still reads them; removing them is a separate, safe follow-up.

```diff
--- a/src/opnsense/mvc/app/models/OPNsense/Interfaces/FieldTypes/NetworkInterfaceField.php
+++ b/src/opnsense/mvc/app/models/OPNsense/Interfaces/FieldTypes/NetworkInterfaceField.php
@@ -72,9 +72,11 @@ class NetworkInterfaceContainerField extends ContainerField
             if (in_array($key, $skiplist)) {
                 continue; /* skip otherwise processed */
             }
-            $result[$key] = $node->getValue();
+            $result[$key] = ($node instanceof BooleanField && empty($node->getValue()))
+                ? '' : $node->getValue();
         }
-        $result['dhcp6_norequest_dns'] = $this->dhcp6_request_dns->isEmpty() ? '1' : '0';
+        /* no model field of its own, it is the inverse of dhcp6_request_dns */
+        $result['dhcp6_norequest_dns'] = $this->dhcp6_request_dns->isEmpty() ? '1' : '';
         foreach (['dhcp6-prefix-id', 'dhcp6_ifid', 'track6-prefix-id', 'track6_ifid'] as $fld) {
             if (!$this->$fld->isSet()) {
                 $result[$fld] = intval($this->$fld->getValue(), 16);

--- a/src/opnsense/mvc/app/models/OPNsense/Interfaces/NetworkInterface.php
+++ b/src/opnsense/mvc/app/models/OPNsense/Interfaces/NetworkInterface.php
@@ -29,7 +29,6 @@
 namespace OPNsense\Interfaces;
 
 use OPNsense\Base\BaseModel;
-use OPNsense\Base\FieldTypes\BooleanField;
 use OPNsense\Base\Messages\Message;
 use OPNsense\Core\Backend;
 use OPNsense\Core\Config;
@@ -138,10 +137,8 @@ class NetworkInterface extends BaseModel
                     $todo['pending_action'] = 'relink';
                 }
                 foreach ($todo['pending'] as $prop => $value) {
-                    if ($prop === 'dhcp6_norequest_dns' && !isset($intf->$prop)) {
-                        $curval = '0'; /* actually stored as dhcp6_request_dns in our model */
-                    } elseif (!isset($intf->$prop)) {
-                        $curval = $this->interface->$key->$prop instanceof BooleanField  ? '0' : '';
+                    if (!isset($intf->$prop)) {
+                        $curval = '';
                     } else {
                         $curval = $intf->$prop;
                     }
```

The `BooleanField` import must go with it: this patch removes its last use in that file, and
`Scripts/class-import.sh` (run by `make lint` as `lint-import`) otherwise goes from 48 stale-import
warnings to 49.

The `NetworkInterface.php` change follows from that rather than being incidental. Its special case
for `dhcp6_norequest_dns` existed only to force `'0'` where the generic branch would give `''` —
that key has no model field of its own, so `instanceof BooleanField` is false for it. Once
`toLegacy()` emits `''`, the special case and the `instanceof` fallback are both redundant, and the
comparison collapses to *absent means `''`* — the second hunk above.

The note that removed comment carried — that `dhcp6_norequest_dns` is stored as
`dhcp6_request_dns` — moves to the line in `toLegacy()` that performs the inversion, which is
where it applies.

The second hunk is not cosmetic, it is required. Left as it is, `serializeToConfig()` substitutes
`'0'` for an absent `BooleanField`, while the new `toLegacy()` now emits `''` for that same false
boolean — and `'0' != ''`, so every interface carrying any unticked option would be queued as
changed on every save. Collapsing the branches to *absent means `''`* is what keeps the two sides
of that comparison agreeing.

Based on master `14ce4ca943`.

---

**Tested**

Fresh `opnsense-devel` 27.1.a_263 lab VM, built from the stock image for this run and confirmed
unmodified with `pkg check -s opnsense-devel` (clean) before anything was patched.
Topology: WAN = DHCPv6, LAN = Track Interface via WAN, RA mode `unmanaged`.

The no-op save was driven twice: once through the GUI (Interfaces ▸ Assignments ▸ edit **wan** ▸
Save ▸ Apply) and once by calling `interfaces/assignment` directly. Both land on the same
`set_item` and `reconfigure` endpoints — the Assignments page is an MVC view that posts to them
(`assignment.volt:32`, `:78`) — so this is not a comparison of two controllers. What differs is
the payload. The browser posts an explicit `"0"` for every unticked checkbox — `getFormData()`
in `src/opnsense/www/js/opnsense.js:104` — while my API call sent a partial JSON body naming only
the description and never mentioning the boolean keys at all.

The stored result was the same either way, damaged and healed alike. What I compared was all
twelve of the `BooleanField`s the rescue lists miss — eleven of them directly, and
`dhcp6_request_dns` through the derived `dhcp6_norequest_dns` that `toLegacy()` emits in its place
— together with the `ia-na`/`ia-pd`/`id-assoc` lines of the generated `dhcp6c.conf`, radvd state,
the `bogonsv6` count and the `PROMISC` flag on the wire. Every one of them matched.

So the stored `"0"` is not an artifact of what the browser posts: it comes out the same when the
client sends `"0"` and when the client sends nothing at all for that key. Either way the value is
produced downstream, by the transform this patch changes. (Not the config file byte for byte,
though: `save()` stamps a fresh `<revision>` on every write.)

Ticking and unticking Promiscuous mode was also done both ways. There the two observables
captured were the stored `wan.promisc` value and the interface's `PROMISC` flag; both matched.

The remaining measurements below were taken at the config layer, once each — they are noted where
that matters.

*Unticked options stop being written, and a genuinely enabled one is left alone:*

```
wan.blockbogons=1          <- genuinely enabled, preserved
wan.blockpriv=1            <- genuinely enabled, preserved
wan.promisc=ABSENT
wan.gateway_interface=ABSENT
wan.dhcp6prefixonly=ABSENT
wan.dhcpd6track6allowoverride=ABSENT
wan.hw_settings_overwrite=ABSENT
lan.<same boolean keys>=ABSENT
```

The same holds for the rest of the twelve, on both interfaces. Two non-boolean keys,
`track6-prefix-id` and `track6_ifid`, are still written as `'0x0'` on both — that belongs to a
separate defect in the same `toLegacy()`, filed as #10878 and patched in #10879 (the hex ID
fields round-trip through `intval(..., 16)`). This patch does not touch it, and #10879 is
referred to a few more times below where the two interact.

The fix is self-healing: the strays the unpatched code already wrote are removed by the next save.

*`dhcp6c.conf` regains the address association, and stops sending a hint nobody asked for.*
Three lines move, and each one is gated by a presence test on a `BooleanField`:

```
gained:  send ia-na 0;          interfaces.inc:2938  if (!isset($wancfg['dhcp6prefixonly']))
gained:  id-assoc na 0 { };     interfaces.inc:2973  same gate
lost:    prefix ::/64 infinity; interfaces.inc:2979  if (isset($wancfg['dhcp6-ia-pd-send-hint']))
```

`dhcp6prefixonly` and `dhcp6-ia-pd-send-hint` are both `BooleanField`s
(`NetworkInterface.xml:175` and `:178`) and both are unticked in the GUI. Stored as `"0"` they
read as *set*, so the WAN never requests a non-temporary address, and it sends a prefix-length
hint the operator never asked for. Removing the keys restores both. Everything else in the
`id-assoc pd` block is unchanged on both sides.

*radvd is no longer killed on a tracking LAN:*

```
stock:   radvd.conf absent, process stopped
patched: radvd.conf present, process RUNNING
```

*The bogons table stops being loaded for an option nobody enabled.* The key is tested two ways —
`src/etc/inc/filter.inc:48` uses `isset()` to decide whether to load the table, while
`src/etc/inc/filter.lib.inc:346` uses `empty()` for the `<bogonsv6>` rule itself, so only the
former sees a stored `"0"`. Measured at the config layer — the key is set directly and
`configctl filter reload` run between rows, so that it is the single variable:

```
wan='0'    lan='0'     -> bogonsv6 entries: 75      (what stock leaves behind)
wan=absent lan=absent  -> bogonsv6 entries: 0       (nobody enabled it: table no longer loaded)
wan='1'    lan=absent  -> bogonsv6 entries: 75      (control — genuinely enabled still works)
```

To be clear about the middle row: on the test box `wan.blockbogons` is genuinely enabled, so the
patch leaves 75 entries there and should. The row shows what changes when *no* interface has the
option ticked — which under stock is indistinguishable from having ticked and unticked it.

*Ticking a box still works, both ways:*

```
tick Promiscuous mode, Save+Apply   -> wan.promisc=1        wire: PROMISC
untick, Save+Apply                  -> wan.promisc=ABSENT   wire: not promisc
```

Note the stored value is *absent*, not `"0"` — which is the whole point.

*Change detection is unaffected.* This patch touches `serializeToConfig()`'s comparison, so it is
worth confirming it does not disturb it. Measured from an identical pristine `config.xml` each
time, counting the entries in `/tmp/.interfaces.todo` — the queue `serializeToConfig()` writes
itself, so the count is the product's own decision and not something this test models:

```
                     first save   settled no-op save
stock                     2               2
this patch                2               2      <- identical to stock
#10879 only               2               0
this patch + #10879       2               0
```

Stock and this patch are indistinguishable: before the patch a false boolean compares `"0"`
against the `"0"` that `serializeToConfig()` substitutes for an absent node, and after it, `''`
against `''`. Both are self-consistent, so no `BooleanField` flags an interface either way.

The interfaces that stay queued on the first two rows are flagged by `dhcp6-prefix-id` and
`dhcp6_ifid` — `'' != 0` under PHP 8 — which is #10879's defect, not this one. Fixing that is
what takes the settled count to 0, with or without this patch.

*Survives a reboot.* Full reboot of the VM, then re-checked:

```
every false boolean:   still absent
wan.blockpriv=1        still present
wan.blockbogons=1      still present
dhcp6c.conf:           send ia-na 0;  send ia-pd 0;  id-assoc na 0 { };
radvd:                 RUNNING
```

The `dhcp6c.conf` line is the point: with the stock code there is no `send ia-na 0;` in the
generated config, so the client never asks for an IA_NA at all. I cannot show the lease that
would follow — this lab's WAN is a qemu SLIRP link, which *sends* router advertisements (on a
timer, and in reply to a solicitation) but runs no DHCPv6 address server. Its DHCPv4 server is
what leases the interface its `10.0.2.15/24`; on the v6 side, apart from the link-local, the only
address the interface ever holds is a SLAAC-derived one out of SLIRP's `fec0::/64`.

`make lint`, `make style` and `make test` were run natively on FreeBSD 15.1 with PHP 8.5, against
an unpatched `14ce4ca943` control in the same tree with the same tools:

```
make lint  : findings identical (see below)
make style : master   421 warnings / 196 errors
             patched  422 warnings / 196 errors      -> +1 warning, no new errors
make test  : master   OK (308 tests, 736 assertions)
             patched  OK (330 tests, 759 assertions) -> +22 tests, +23 assertions
```

For `make lint`: with parallel-lint's progress output stripped, the only line that differs
between the two runs is parallel-lint's own summary, `Checked 2615 files` -> `Checked 2616 files`
— the file this patch adds. No finding changes, and nothing new is reported.

For `make style`: phpcs reports exactly one file it did not report on master.

```
FILE: ...els/OPNsense/Interfaces/FieldTypes/NetworkInterfaceFieldTest.php
FOUND 0 ERRORS AND 1 WARNING AFFECTING 1 LINE

 1 | WARNING | A file should declare new symbols (classes, functions,
   |         | constants, etc.) and cause no other side effects, or
   |         | it should execute logic with side effects, but should
   |         | not do both. The first symbol is defined on line 47
   |         | and the first side effect is on line 34.
```

That is the `require_once` the test needs, because `NetworkInterfaceContainerField` shares a file
with `NetworkInterfaceField` and PSR-4 cannot reach it by name. **Twenty-six existing test files
carry the same warning for the same reason** — `TextFieldTest`, `PortFieldTest`,
`HostnameFieldTest`, `ProtocolFieldTest` and twenty-two others (52 files across the whole styled
tree). I have followed the existing pattern rather than invent a way around it, but say the word
if you would rather I did.

*Regression test.* Seven cases added to
`src/opnsense/mvc/tests/app/models/OPNsense/Interfaces/FieldTypes/NetworkInterfaceFieldTest.php`,
following the pattern #10389 set for `IdassocTest`. Three fail on unpatched master and pass
with this change; three are deliberate guards that must pass either way — a true boolean is
still written as `"1"`,
a non-boolean field whose value happens to be `"0"` is still written, and `dhcp6_norequest_dns`
is still `'1'` when DNS is not requested. (Its mirror case, the one asserting `''` when DNS *is*
requested, is one of the three that fail on stock — stock emits `'0'` there.)

The seventh is the one that matters most for keeping this fixed. The whole argument for fixing
the transform rather than extending `$legacybools` is that there is then no list left to fall
behind the model — so the test asserts that directly, by reading every `BooleanField` declared in
`NetworkInterface.xml` at test time via a data provider:

```php
foreach ($xml->xpath('//*[@type="BooleanField"]') as $node) { ... }
```

A `BooleanField` added to the model next year is covered the day it is added, with no edit to the
test file. On the current model that generates 16 cases (17 declared, less the deliberately
skiplisted `dhcp6_request_dns`, which is covered by the two derived-field tests instead).

Specificity matrix, measured on this branch alone (22 cases: 6 named plus 16 generated from the
model):

```
this patch (fix + test)      OK (22 tests, 23 assertions)
same test on stock master    Tests: 22, Assertions: 22, Failures: 19
```

Nineteen of the twenty-two fail without the change. The three that pass either way are the
deliberate guards — a true boolean is still written as `"1"`, a non-boolean whose value happens
to be `"0"` is still written, and the derived `dhcp6_norequest_dns` still inverts.

⚠️ **These two patches overlap.** They edit the same few lines of `toLegacy()` from both sides —
this one carries `if (!$this->$fld->isSet()) {` as trailing context, which #10879 rewrites, and
#10879 carries the `dhcp6_norequest_dns` assignment as leading context, which this one rewrites —
and they add the same `plist` line and the same new test file. So plain `git apply` fails for
whichever lands second, in either order.
`git apply --3way` handles it: I tried both landing orders against `14ce4ca943` and in each one
the two PHP files and the `plist` line merge cleanly and correctly, leaving both fixes intact.

The one thing no merge strategy can resolve is the test file: both PRs create
`.../FieldTypes/NetworkInterfaceFieldTest.php` from scratch — #10879 with its 6 cases, this one
with 7 — so it comes out as an add/add conflict, and whichever lands second should fold its cases
into the existing class rather than add the file.

I am happy to move this to a different layer if you would rather solve it in the writers, or to
extend `$legacybools` instead — say which and I will re-push.

⚠️ If you verify this by editing the files directly rather than via `opnsense-patch`, recycle the
PHP workers afterwards — `pkill -x php-cgi`, then `configctl webgui restart` if they do not come
back on their own — and confirm with `pgrep -x php-cgi` that the PIDs actually changed.
`opcache.enable` is On and the GUI keeps serving the pre-patch code otherwise, so the first save
after patching changes nothing and looks like the fix failed. PHP CLI shows the patched behavior
immediately, which makes it easy to misread.

---

**Related issue**

Fixes #10881

Ref: introduced by 92cb28a669 (#10740) — `git log -S '$result[$key] = $node->getValue();' --
src/opnsense/mvc/app/models/OPNsense/Interfaces/FieldTypes/NetworkInterfaceField.php` returns
that commit alone. 6e4dab1042 (#10802) then added the `$legacybools` list, which unsets five of
the seventeen BooleanFields after the fact.


AdSchellevis:
These walls of AI generated texts are impossible to proces and unfortunately often contain hallucinations. If I’m not mistaken, dhcp6_norequest_dns isn’t used with isset().

if there is an actual reproducible issue with a converted value, best open a ticket explaining how to reproduce the issue so we can assess if we should fix the model or the implementation, isset usage is discouraged from our end, but unfortunately some specific properties are difficult to change.

AdSchellevis:
It looks like I missed your other issue, since we discourage the use of isset(), $legacybools is likely the best option to keep the problematic items visible so we know which parts of the code to clean up in a later stage.

fichtner:
@AdSchellevis thanks for the tip, added in [d6a1f01](https://github.com/opnsense/core/commit/d6a1f0118c94fa5c133d54e8cd5dafa6aaca98bb) and follow ups

PR #10894, opened 2026-09-18
Shaper: delete pipes/queues before reconfiguring dnctl, fixes #10893

### LINKED ISSUE #10893: Shaper: Queues configured with CoDel + ECN run droptail instead (flowset busy)
**Important notices**

- [x] I have read the contributing guidelines *(tick once you have)*
- [x] I am convinced that my issue is new after having checked both open and closed issues

*Prior art checked: #1279 (2016, closed), #9298 (closed as timed-out; OPNsense-bot advised "the easiest option to gain traction is to close this ticket and open a new one using one of our templates" — this is that issue), #9875, #9906.*

- **Model used:** Claude Opus 5 (Anthropic), via Claude Code
- **Extent of AI involvement:** All investigation, live testing on the affected hosts, and drafting of this report. Every factual claim was verified by direct command output on running systems, not inferred from documentation. Reviewed by the submitter before filing.

## Describe the bug

Queues configured with CoDel + ECN run **droptail**. The setting never reaches a packet. The kernel says why; nothing surfaces it:

```
kernel: [1932] config_aqm Unable to configure flowset, flowset busy!   (x6, one per queue)
```

`dnctl` exits 0 and prints nothing, so OPNsense believes the apply succeeded. GUI, `config.xml` and the generated file all agree CoDel is on:

```
# grep ^queue /usr/local/etc/dnctl.conf
queue 10002 config pipe 10000 mask dst-ip /32  weight 100 codel  ecn
```

`mask` and `weight` from that exact line **are** applied. Only `codel` is dropped:

```
# dnctl queue show
q10002  50 sl. 44 flows (256 buckets) sched 10000 weight 100 lmax 1500 pri 0
```

A healthy flowset ends in `AQM CoDel …` or `droptail`. An affected one prints **neither** — the same signature visible in #1279's 2016 paste.

Last known working version: unknown; #1279 is from 2016 and #9298 reports it on 25.7.5, so this looks long-standing rather than a regression.

## To Reproduce

1. Firewall → Shaper → Pipes: create a pipe, scheduler QFQ.
2. Firewall → Shaper → Queues: create a queue on that pipe with **CoDel** and **ECN** enabled.
3. Firewall → Shaper → Rules: add a rule targeting that queue, so it carries traffic.
4. Reboot, or run `configctl shaper reload`.
5. `dnctl queue show` — no AQM on the queue. `dmesg | grep config_aqm` shows *flowset busy*.

Reproduced on two independent machines (a CARP pair), across reboots.

## Expected behavior

The queue runs CoDel as configured — or the apply fails loudly instead of succeeding silently.

## Cause, and the fix upstream already prescribed

This is not an upstream defect. The kernel refusal is **deliberate**. RasoolAlSaadi, who wrote the AQM code, explained it on #1279 in 2016:

> *"It seems that the pipe/queue is currently busy (a traffic is passing through pipe/queue ipfw rule) while you are trying to reconfigure AQM. Due to the fact that the internal variables of each AQM instance is allocated and freed dynamically, re-configuring AQM … can lead the AQM code to access unallocated memory … and kernel panic could happen. Therefore, I tried to avoid any potential kernel panic by preventing re-configuring AQM for busy pipe/queue."*

and gave the remedy:

> *"Delete the old pipes/queues that need to be re-configured with AQM and create them again with all the desired settings."*

OPNsense does not do that. `configctl shaper reload` → `scripts/shaper/start.sh dnctl` → `/etc/rc.d/dnctl start` → `dnctl /usr/local/etc/dnctl.conf`, with **no delete anywhere**, so the file is applied over flowsets that already exist and are passing traffic — precisely the case the kernel refuses. Every later reload hits the same wall and a reboot reproduces it, so no supported action in the product repairs the state.

The sibling template already gets this right:

```
# head -5 /usr/local/etc/ipfw.rules
flush                                                  <-- rebuilt from clean

# head -4 /usr/local/etc/dnctl.conf
pipe 10000 config bw 930Mbit/s type qfq codel ecn      <-- no delete anywhere
```

**Suggested fix:** emit `queue N delete` / `pipe N delete` at the top of the `dnctl.conf` template, mirroring the `flush` that `ipfw.rules` already emits. `dnctl` has no `flush` command and no `-f` option (`dnctl [pipe|queue|sched] {delete|list|show}`), so deletion must be by number. A recreated flowset is not busy, so the AQM applies.

That costs a sub-second window where rules reference flowsets that no longer exist. If that is unacceptable, a post-apply check that every configured queue reports an AQM would at least stop the failure being silent.

## Describe alternatives you considered

Verified on the affected host, same session:

| Action | Result |
| --- | --- |
| Fresh pipe + queues created together, before traffic | AQM on all of them |
| Same queue line against the in-use pipe | no AQM, **exit 0, no error** |
| `dnctl queue N delete` + recreate, pipe untouched | still no AQM |
| Reattach an existing queue to a fresh pipe | AQM, first flowset on that scheduler only |
| **Delete the pipes, re-apply the unchanged file** | **AQM on every queue** |

Ruled out by direct test: pipe-level codel conflicting with queue-level codel; a global AQM instance cap; reconfiguring a pipe stripping its queues' AQM.

## Relevant log files

```
# dmesg | grep config_aqm
[1932] config_aqm Unable to configure flowset, flowset busy!    (x6 at boot, one per queue)

# dnctl queue show     (after boot — no AQM on any)
q10000  50 sl. 46 flows (256 buckets) sched 10000 weight 100 lmax 1500 pri 0
q10002  50 sl. 44 flows (256 buckets) sched 10000 weight 100 lmax 1500 pri 0

# after deleting the pipes and re-applying the same unchanged file
q10000  50 sl. 17 flows (256 buckets) sched 10000 weight 100 lmax 1500 pri 0  AQM CoDel target 5ms interval 100ms ECN
q10002  50 sl. 10 flows (256 buckets) sched 10000 weight 100 lmax 1500 pri 0  AQM CoDel target 5ms interval 100ms ECN
```

## Additional context

Practical impact: after any reboot, the CoDel/PIE and ECN controls on shaper **queues** affect nothing. The only AQM instance left is on the pipes' own default flowset, which carries no packets when every rule targets a queue.

Measured cost here — droptail at 50 slots (~75KB) against a ~900KB BDP at 362 Mbit/s produced ~1700 drops in 27s (~0.4%) on a single host's flow during bulk downloads.

Corroborated by @MilesBHuff on #9298 (2026-03-13): *"I am also seeing this. It started in the v25 releases and is still happening in v26.1.4."*

## Environment

OPNsense 26.7.4_1 (amd64), FreeBSD 15.1-RELEASE-p3
Guest VM on Proxmox VE, 8 vCPU, virtio (vtnet) NICs
Two hosts in a CARP pair, both affected identically.

MikeBishop:
**Important notices**

- **Model used:** Claude Sonnet 5 (Anthropic), via Claude Code
- **Extent of AI involvement:** Investigation, live testing on the affected host, template fix, and drafting of this description. Every factual claim verified by direct command output on a running system. Reviewed by the submitter before filing.

---

**Describe the problem**

Queues configured with CoDel/PIE (+ ECN) run droptail instead. Dummynet's config_aqm refuses to (re)configure AQM on a pipe or queue that is still carrying traffic ("flowset busy" in dmesg), to avoid a use-after-free/kernel panic risk in AQM's dynamic instance allocation. `dnctl` exits 0 and prints nothing when this happens, so OPNsense believes the apply succeeded, and GUI/config.xml/generated dnctl.conf all show CoDel enabled while the kernel silently runs droptail. Reproduced live: `dnctl queue show` shows no AQM line on any queue, `dmesg | grep config_aqm` shows repeated "flowset busy" (one per queue) on every reload and reboot.

Root cause: `configctl shaper reload` → `scripts/shaper/start.sh dnctl` → `/etc/rc.d/dnctl start` → `dnctl dnctl.conf` reapplies the generated config directly over pipes/queues that already exist and are passing traffic — exactly the case the kernel refuses. Nothing deletes the old flowsets first, so every later reload and every reboot hits the same wall and no supported action repairs the state. The sibling ipfw.rules template already handles this correctly by starting with `flush`; dnctl.conf did not.

---

Emit `queue N delete` / `pipe N delete` for every configured queue and pipe at the top of the generated dnctl.conf, before the existing pipe/queue config blocks, mirroring the `flush` already present in ipfw.rules. A freshly deleted-and-recreated flowset is not busy, so AQM applies. `dnctl` exits 0 and continues processing subsequent lines when a delete target doesn't exist, so this is also safe on first boot / fresh config.

Verified live on an affected host:
- Before fix: all 6 configured queues show no AQM line in `dnctl queue show`; dmesg shows 6x `config_aqm ... flowset busy` per reload.
- Applied fixed template output directly: all 6 queues now show `AQM CoDel target 5ms interval 100ms ECN`; no new config_aqm errors in dmesg afterward.

---

**Related issue**

Closes https://github.com/opnsense/core/issues/10893


AdSchellevis:
The need to drop pipes and queues on boot sounds rather odd as none of these exist during boot  (AI hallucination?).

If switching types in some cases is not supported, we likely better add a validation to prevent that from happening. Always dropping references which are also used by states in pf could open a whole new can of worms…

MikeBishop:
Not at boot, no; this matters only on reload. You're right, it will print a warning on deleting the queue that doesn't yet exist. But this isn't just about a setting change that's not possible -- applying exactly the same settings a second time still leaves it broken.

The pf rules should be okay -- they reference the queues by number, so once the queues are back, they function just fine. But we can definitely move this to be only on the reload path rather than at boot and narrow the window where they drop traffic.

Issue #10896, opened 2026-09-18
Gateway monitor (dpinger) killed and not relaunched when SLAAC WAN address is deprecated during a loss event


jamesdalemccall:
**Important notices**

- [x] I have read the contributing guidelines at https://github.com/opnsense/core/blob/master/CONTRIBUTING.md
- [x] I am convinced that my issue is new after having checked both open and closed issues at https://github.com/opnsense/core/issues?q=is%3Aissue (closest are #7400 and #6913, which are boot / link down-up races; this happens on a link that never drops)
- [x] AI tools were used to create at least part of the text submitted herewith.

If AI was used, please disclose:

- Model used: Claude (Anthropic), via Claude Code Fable 5.1 Extra High
- Extent of AI involvement: log collection, code reading and drafting of this report, reviewed by me before submission. The diagnosis was verified against the logs and code on the live system.

**Describe the bug**

After a packet-loss burst on a WAN whose IPv6 address is SLAAC with a short preferred lifetime, the dpinger instance for that gateway is killed by the monitor recover hook and never relaunched. The gateway then shows `down` with RTT/loss `~` on the dashboard indefinitely while the link itself is healthy. Reproduced twice on 26.7.4 (2026-09-09 and 2026-09-18). Not tested on earlier versions.

Mechanism, from `plugins.inc.d/dpinger.inc` (`dpinger_configure_do`) and `interfaces.inc` (`_interfaces_primary_address6`):

1. The running instance for the gateway is killed via `killbypid()` *before* a bind address is resolved.
2. The bind address lookup (`interfaces_primary_address6()` / `interfaces_routed_address6()`) skips any address flagged `deprecated`.
3. Starlink advertises the WAN address with `pltime 150 vltime 300`. During a loss burst longer than 150 s, RAs are lost and the SLAAC address transitions to *deprecated*, so the lookup returns nothing, the function logs "The required STARLINK_SLAAC IPv6 interface address could not be found, skipping." and `continue`s.
4. Nothing re-triggers the monitor hook when the address becomes preferred again (SLAAC re-preferring does not fire `rc.newwanipv6`), so the monitor stays dead until a manual `pluginctl -c monitor STARLINK_SLAAC` or a gateway save.

**To Reproduce**

[...]

PR #10900, opened 2026-09-19
dpinger: keep the running instance when no bind address can be resolved

jamesdalemccall:
**Important notices**

- Model used: Claude (Anthropic), via Claude Code Fable 5.1
- Extent of AI involvement: drafted the patch from the analysis in #10896; I reviewed the diff and tested it operationally on the affected system before submitting.

---

**Describe the problem**

`dpinger_configure_do()` stops every running instance for the requested gateways before it resolves a bind address for the replacement. When the address lookup then fails and the gateway is skipped, the instance that was just stopped is never restarted, and the gateway stays unmonitored until some other event happens to reconfigure it (another gateway alarming, a gateway save, or a manual `pluginctl -c monitor`).

Seen on 26.7.4 with a SLAAC-only WAN whose address was flagged deprecated during a loss event: the `loss -> down` alert ran the recover hook, the hook killed the monitor, the lookup skipped the deprecated address, and the gateway showed `down` with `~` until an unrelated fiber flap ran the hook again (4 minutes later in one case, 2 h 49 min in another). On a single-WAN system nothing would have run it again.

---

Resolve the bind address first, and only stop the previous instance once the replacement is ready to start:

- The up-front loop still stops instances for gateways that are no longer configured (deleted, or monitoring disabled).
- Instances for configured gateways are left alone until the main loop has built the replacement's command line, then stopped immediately before `mwexecfb()`.
- On the "could not be found" skip path the log line now says `keeping previous instance` or `skipping`, so it is clear which case occurred.

For any gateway whose lookup succeeds this is the same stop-and-start as before, a few milliseconds later. It does not bind anything new to a deprecated address and needs no SLAAC state or events.

Tested on 26.7.4_1 with the change applied:

1. `pluginctl -c monitor WAN_GW` on a healthy gateway: old pid stopped, new pid started in the same second, gateway reporting normally.
2. Full `pluginctl -c monitor` with a decoy instance present (`dpinger -i FAKE -p /var/run/dpinger_FAKE.pid ...`): decoy stopped, all configured gateways restarted.
3. SLAAC address flagged with `ifconfig igc1 inet6 <addr> deprecated`, then `pluginctl -c monitor STARLINK_SLAAC`: log reads `The required STARLINK_SLAAC IPv6 interface address could not be found, keeping previous instance.`, pid unchanged, gateway still reporting RTT and loss. (Note for anyone reproducing: the `deprecated` keyword also clears `autoconf`; restore with `ifconfig igc1 inet6 <addr> prefixlen 64 autoconf pltime 150 vltime 300` afterwards.)

The gateway-monitor respawn cron the issue mentions was disabled for the duration of the tests.

---

**Related issue**

Ref: #10896


fichtner (changes requested):
Can we try not to use AI to jump to conclusions here as mentioned in the ticket already? The monitor is requested to be restarted, so the problem isn't the restart... the problem is the overly broad request. Whether or not the Starlink case can be recovered safely is a  different question, but glossing over this will likely cause regressions with people's working setup. This wouldn't be the first time.

jamesdalemccall:
You're right, the problem is the request, not the restart. `20-recover` collects every gateway whose status is `down`, so a monitor that is reporting normally on a legitimately down link gets killed along with genuinely stuck ones.

Would you accept narrowing it there to gateways with no data from the monitor (`$status['loss'] == '~'` — no process, no socket, or nothing reported yet), which is what "stuck in down state" in the comment describes? Closing #10900 regardless.

